PR #6550, opened 2020-09-11
[MU3] Fix #277279: Lack of Capabilities in the Custom Time Signature Creator

elerouxx:
Resolves: *https://musescore.org/en/node/277279* 
Lack of Capabilities in the Custom Time Signature Creator
*https://musescore.org/en/node/279745* 
Allow periods in custom time signatures
*https://musescore.org/en/node/277665*
Cannot use a decimal dot or vulgar fractures like ½ for time signatures

See other requests:
*https://musescore.org/en/node/312343* Time SIgnature Palette: Parenthesis
*https://musescore.org/en/node/312343* 3/4 Time Signature with Secondary 6/8 in parenthesis
*https://musescore.org/en/node/305902* Dual Time Signatures

This PR replaces the _appearance_ numerator/denominator fields with a single parsed text that allows more capabilites to the appearance of time signatures. These changes were made in the Time Signature creator (timedialog), Time Signature properties, and additionally added to the time signature Inspector directly, which is the faster way to enter text and change the appearance of a time signature (can be done locally if a single t/s is selected pressing CTRL).

This single field is properly parsed to identify "blocks" of time signatures (numerator/denominator) as well as complementary glyphs such as Operators +=-x and Parentheses ( ). In the absence of a denominator, the glyphs are drawn vertically centered, as current, and the Plus and Parentheses glyphs are replaced by the large versions.

Examples:
"2/4"
"3/4(6/8)"
"3+2/8"
"2/8+5/8"
"C"
"4/4=C"

image

**(NOTE: The 'Large Time Signature' feature from the inspector and at the end of the video below was removed from this PR, for a future work)**

In action:
parser_ts_inspector2

**Bracketed special characters:**

- To enter fractional glyphs, use brackets: ex. 3[1/2]/4
-  Note value glyphs like quarter, dotted quarter, can be entered in denominators (Carl Off style time signatures) by using numbers and dots within brackets: [2] [4] [8] and [16] where [8] is an eight note, [4.] is a dotted quarter note. Ex: 3/[8.]
- Other (unrecognized) texts between brackets will be discarded.

- Other glyphs include  Open Penderecki time signature ( type  '~' ) Open X time signature (type uppercase 'X') Comma ',' and period or augmentation dot '.'
- Spaces can be added to improve visual and will force separation of 'blocks'.

Additional notes:
- All Time Signature 'official' SMuFL glyphs have been enabled (see https://w3c.github.io/smufl/gitbook/tables/time-signatures.html)
- Big parentheses ( ) (SymId::timeSigParensLeft / Right) are not well implemented in Emmentaler and Gonville fonts. Emmentaler has bad bboxes/metrics, and in Gonville these glyphs are identical to the smaller SymId::timeSigParensLeftSmall. (both cases might need special handling or fontworks)
- The 'Narrow' glyphs from SMuFL have not been activated because they are just the common font time signatures being condensed horizontally, which can be achieved via the time signature Scaling in the inspector.

- [NA] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/libmscore/edit.cpp
+++ b/libmscore/edit.cpp
@@ -821,6 +821,7 @@ void Score::cmdAddTimeSig(Measure* fm, int staffIdx, TimeSig* ts, bool local)
                         nsig->undoChangeProperty(Pid::TIMESIG_TYPE, int(ts->timeSigType()));
                         nsig->undoChangeProperty(Pid::NUMERATOR_STRING, ts->numeratorString());
                         nsig->undoChangeProperty(Pid::DENOMINATOR_STRING, ts->denominatorString());
+                        nsig->undoChangeProperty(Pid::PARSER_STRING, ts->parserString());
                         nsig->undoChangeProperty(Pid::TIMESIG_STRETCH, QVariant::fromValue(ts->stretch()));
                         nsig->undoChangeProperty(Pid::GROUPS, QVariant::fromValue(ts->groups()));
                         nsig->setSelected(false);
@@ -894,6 +895,7 @@ void Score::cmdAddTimeSig(Measure* fm, int staffIdx, TimeSig* ts, bool local)
                               nsig->undoChangeProperty(Pid::TIMESIG, QVariant::fromValue(ts->sig()));
                               nsig->undoChangeProperty(Pid::NUMERATOR_STRING, ts->numeratorString());
                               nsig->undoChangeProperty(Pid::DENOMINATOR_STRING, ts->denominatorString());
+                              nsig->undoChangeProperty(Pid::PARSER_STRING, ts->parserString());
 
                               // HACK do it twice to accommodate undo
                               nsig->undoChangeProperty(Pid::TIMESIG_TYPE, int(ts->timeSigType()));
--- a/libmscore/property.cpp
+++ b/libmscore/property.cpp
@@ -138,6 +138,7 @@ static constexpr PropertyMetaData propertyList[] = {
       { Pid::ACCIDENTAL_TYPE,         true,  "subtype",               P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "type")             },
       { Pid::NUMERATOR_STRING,        false, "textN",                 P_TYPE::STRING,              DUMMY_QT_TRANSLATE_NOOP("propertyName", "numerator string") },
       { Pid::DENOMINATOR_STRING,      false, "textD",                 P_TYPE::STRING,              DUMMY_QT_TRANSLATE_NOOP("propertyName", "denominator string") },
+      { Pid::PARSER_STRING,           false, "textP",                 P_TYPE::STRING,              DUMMY_QT_TRANSLATE_NOOP("propertyName", "parser string") },
       { Pid::FBPREFIX,                false, "prefix",                P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "prefix")           },
       { Pid::FBDIGIT,                 false, "digit",                 P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "digit")            },
       { Pid::FBSUFFIX,                false, "suffix",                P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "suffix")           },
--- a/libmscore/property.h
+++ b/libmscore/property.h
@@ -144,6 +144,7 @@ enum class Pid {
       ACCIDENTAL_TYPE,
       NUMERATOR_STRING,
       DENOMINATOR_STRING,
+      PARSER_STRING,
       FBPREFIX,             // used for FiguredBassItem
       FBDIGIT,              //    "           "
       FBSUFFIX,             //    "           "
--- a/libmscore/sym.cpp
+++ b/libmscore/sym.cpp
@@ -2865,7 +2865,39 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symNames = { {
       "braceLarge",
       "braceLarger",
 
-//    MuseScore-local symbols, precomposed symbols to mimic some Emmentaler glyphs
+//    SMuFL stylistic alternates for Large Time Signatures
+
+      "timeSig0Large",
+      "timeSig1Large",
+      "timeSig2Large",
+      "timeSig3Large",
+      "timeSig4Large",
+      "timeSig5Large",
+      "timeSig6Large",
+      "timeSig7Large",
+      "timeSig8Large",
+      "timeSig9Large",
+      "timeSigCommonLarge",
+      "timeSigCutCommonLarge",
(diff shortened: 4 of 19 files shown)


Jojo-Schmitz:
Any chance to get this ready for 3.6?

shoogle:
I think this is the best solution that could make it into 3.6 rather than having to wait for 4.0.

I recommend:

1. Remove the separate input fields for numerator and denominator (i.e. just keep the parser text field).
2. Add some funky examples to the Time Signatures palette in the Advanced workspace. This should help alleviate the discoverability problem.

Jojo-Schmitz:
BTW: that `collect_artifacts` is no longer needed, those are now always generated

Marr11317:
Just tested the PR, and it works like a charm.

It also comes with a bonus: It's fairly common in jazz Charts to show a closing bracked after the time sig, like in this example: 
image

This was possible already with Musejazz, that provides a closing parenthesis by default in its common and cut time sigs, but can now be done with any. More comments in [this thread](https://musescore.org/en/node/313219).

Marr11317:
haha tumbao 3-3-2 in a time sig:
image

However, here's the text I used to make that work: `[1]/[[4.]+[4.]+[4]` 
As you can see, there's a closing bracket missing. It doesn't work if I add. Not a merge blocker though.

Marr11317:
looking at the vtest diff, it looks like this pr solves the mordent issues?
image

elerouxx:
> looking at the vtest diff, it looks like this pr solves the mordent issues?

Hey! Thanks for your comments. I was just updating the PR and preparing it for merge. Nice that you are having fun :D

Yes, this includes a small fix with the 'ADVANCE' value not being properly added to some glyphs when fallback is needed (for example the plus sign on a  3+2/4 time signature in gootville font in 3.5.2) and the side effect is that the ornaments were fixed.

elerouxx:
> haha tumbao 3-3-2 in a time sig:
> image
> 
> However, here's the text I used to make that work: `[1]/[[4.]+[4.]+[4]`
> As you can see, there's a closing bracket missing. It doesn't work if I add it. Not a merge blocker though.

Actually, plus signs aren't intended to work on a denominator, but you hacked it :D - but if you close that bracket, everything inside becomes a code and is not recognized.

Tantacrul:
I have a few comments.

First, when you click on a time signature and select 'large time signature' the result is a smaller, squashed time signature. I know what's happening here, but we can't allow UX like that in this app. You'd need to have a good understanding of SMuFL standards and quite a bit of development insight to understand what is going on. I don't think that option should exist unless it produces a desirable result by default. The user should not have to go messing around with scale settings. Most users will just consider it to be a new broken feature.

image

Jojo-Schmitz:
> looking at the vtest diff, it looks like this pr solves the mordent issues?
> image

So those vtest diffs are a Good Thing (TM)

Marr11317:
@tantacrul Dorico has a lot of these latex inputs, and even though it requires some documentation reading, in the end it is super fast to type and powerful. 

These custom time sigs are for advanced users. I wouldn't see someone inputting `4/8=[4.]+[8]` without a deep knowledge of what he is doing anyway.

Tantacrul:
> @Tantacrul Dorico has a lot of these latex inputs, and even though it requires some documentation reading, in the end it is super fast to type and powerful.
> 
> These custom time sigs are for advanced users. I wouldn't see someone inputting `4/8=[4.]+[8]` without a deep knowledge of what he is doing anyway.

How do these advanced users figure this out? I'm an advanced user and I would never have guessed.

As a general rule, do not put options in front of people that can not be deduced. This is a fairly common design principle.

Another point I'd make is that it is confusing too. It could easily be misconstrued by a less advanced user as something that allows you to actually change your time signature. I wouldn't be surprised to hear feedback from people trying this out and typing '3/8' into that box and thinking the result is broken.

elerouxx:
Just added a bunch of links to issues and discussions about time signature needs I tried to keep in mind. Basically, people having to hide time signatures to work around graphic needs, which are common engraving practices. 

The visual needs of time signatures can be complex and very diverse (as they are in real case examples) even if the real values are simpler. Dorico has a nice implementation but even so there will be limitations by design, from what I see.
 
I'd like to see working real _Numerator_/_Denominator_ fields above the Appearance Text field but that was not possible with current limitations, both mine and from the current code. Maybe a better label would help clarify this is meant for appearance only? ( _Alternate Text_, _Appearance Override_?) 

I opted for the smart text among other ideas that were discussed, because it doesn't block any future development: ultimately this property of the timeSig will most probably be saved on a string anyway, even if the interface is changed i.e. to a graphic draggable interface or whatever else.

As for the bracket 'codes', most users won't never use them, as they cover really specific needs. So that user will probably have to find out 'how to' in the documentation for the first time he needs fractions, or the very rare Carl Orff signatures (the note values in denominators). Other than that, the use of the text field is pretty straightforward, and it reflects instantly on the selected time signature.
Other options were discussed  (like using 'q' or 'w', or '8th' and '4th' instead of bracketed codes,  and some non practical ideas for fractional symbols) but none of the alternatives seemed better.

shoogle:
> How do these advanced users figure this out?

I suggested adding some funky examples to the Time Signatures palette in the Advanced workspace. I don't think it has been done yet, but if it was then it would enable users to add one of those examples and right click on it to see how it was constructed.

> It could easily be misconstrued by a less advanced user as something that allows you to actually change your time signature.

The existing dialog had that problem already, so I don't think that should count against this PR.

Ultimately I think this PR is a great improvement over what was available before, with no downsides that were not already present (except the Large Time Signature issue that is going to be removed/resolved anyway). Clearly a better interface is desirable at some point, but I see no reason to make advanced users wait for it when this is ready now. Even when that interface comes along, we will in all likelihood end up storing the digits in the MSCX file using the same LaTeX-style code as is implemented here.

elerouxx:
I think it's important to understand that the note value symbols, as well as the vulgar fractions, are here because of specific, uncommon but known styles of time signatures several people requested, and whoever needs them is because they know what to do with them. It's not like they are common available 'choices' for common use, and that users will have to figure out how to arbitrarily use in their music or otherwise will have problems creating time signatures. It's an advanced feature, just that.

elerouxx:
Since I removed the 'Large' checkbox, I will not link or otherwise modify the Scale X and Y fields - Although I agree with the suggestion, that feature was implemented long time ago (not by me), it is not related to this PR anymore, and most importantly, I have no clue how to do that.

Jojo-Schmitz:
rebase needed

elerouxx:
> rebase needed

Rebased. Also fixed 'Other' radio button and combo not working with the new single string implementation. 
TODO: This PR fixes refreshing of the notegroups in timeSigProperties, but this refresh only works with the text field. Other appearance buttons (Cut, other...) don't refresh the notegroups in properties dialog.

vpereverzev:
3.x merges are closed now, I suggest to create a new one after the discussion with Tantacrul about the UX part of this feature

PR #6552, opened 2020-09-11
Fix for rendermidi's bug. NoteOff event was missed.

psmokotnin:
Fix for a bug in rendermidi. NoteOff event was missed. NoteOn with zero velocity was sent instead.

### DIFF
--- a/libmscore/rendermidi.cpp
+++ b/libmscore/rendermidi.cpp
@@ -256,16 +256,18 @@ static void playNote(EventMap* events, const Note* note, int channel, int pitch,
         return;
     }
     velo = note->customizeVelocity(velo);
-    NPlayEvent ev(ME_NOTEON, channel, pitch, velo);
-    ev.setOriginatingStaff(staffIdx);
-    ev.setTuning(note->tuning());
-    ev.setNote(note);
+    NPlayEvent noteOnEvent(ME_NOTEON, channel, pitch, velo);
+    noteOnEvent.setOriginatingStaff(staffIdx);
+    noteOnEvent.setTuning(note->tuning());
+    noteOnEvent.setNote(note);
     if (offTime < onTime) {
         offTime = onTime;
     }
-    events->insert(std::pair<int, NPlayEvent>(onTime, ev));
-    ev.setVelo(0);
-    events->insert(std::pair<int, NPlayEvent>(offTime, ev));
+    events->insert(std::pair<int, NPlayEvent>(onTime, noteOnEvent));
+    NPlayEvent noteOffEvent = noteOnEvent;
+    noteOffEvent.setType(ME_NOTEOFF);
+    noteOffEvent.setVelo(0);
+    events->insert(std::pair<int, NPlayEvent>(offTime, noteOffEvent));
 }
 
 //---------------------------------------------------------


dmitrio95:
I am not against the change proposed here, but this is definitely not a bug. Using zero-velocity note on events to turn a note off is a rather common practice, and it is well-understood by all (or most of) MIDI-capable software. Moreover, proper handling of this way of turning notes off seems to be required by MIDI specification itself. So it is not a mistake that turning notes off was implemented in MuseScore that way — which doesn't mean though that this cannot be changed for a good reason.

MarcSabatella:
I agree with @dmitrio95 .  But, there may well be a bug in how we interpret MIDI *input*.  We should always accept either form this message.  Every so often we get bug reports about notes not being accepted on input, but only accepted after entering one note using another method, and then they stack up to form a chord.  The logical explanation is that we are not seeing the note-off / note-on-zero-velocity message.  Sometimes it may be the keyboard not sending it, but I wonder if we don't accept both forms?

psmokotnin:
MIDI1.0 standard says: "any Note-On command **must** have a corresponding specific Note-Off command"

psmokotnin:
@dmitrio95 can you, please, give some examples, what software doesn't support NoteOff command?

dmitrio95:
> MIDI1.0 standard says: "any Note-On command **must** have a corresponding specific Note-Off command"

The Complete MIDI 1.0 Detailed Specification has a "Note-off" section which says the following:

> > MIDI provides two roughly equivalent means of turning off a note (voice). A note may be turned off either by sending a Note-Off message for the same note number and channel, or by sending a Note-On message for that note and channel with a velocity value of zero.

and

> > A receiver must be capable of recognizing either method of turning off a note, and should treat them identically.

The sentence you have quoted appears in the section dedicated to treatment of All Notes Off messages, and it is likely not meant to be a statement that using Note-Off messages is the only correct method to turn notes off, especially that it would contradict what is said in the same document above.

> @dmitrio95 can you, please, give some examples, what software doesn't support NoteOff command?

I don't think that such notable examples exist, I just wanted to say that the current behavior is also correct and should not be considered a bug.

psmokotnin:
> The sentence you have quoted appears in the section dedicated to treatment of All Notes Off messages, and it is likely not meant to be a statement that using Note-Off messages

Everything said in standard is applicable.

> MIDI provides two roughly equivalent means of turning off a note (voice).

You've missed here the last sentence in this paragraph:

> The advantage to using "Note-On at zero velocity" is that it can avoid sending additional status bytes when Running Status is employed.

In the resume of the standard:

- NoteOn command must be followed by NoteOff
- In the running mode, NoteOff command can be missed if NoteOn with zero velocity is sent
- NoteOn with zero velocity is a patch for protocol efficiency
- Receivers must recognize both types

Additional note for the future of MIDI:
In the MIDI2.0 standard, this misconception was removed. NoteOn with zero velocity won't work like NoteOff.

dmitrio95:
> * In the running mode, NoteOff command can be missed if NoteOn with zero velocity is sent
> 
> * NoteOn with zero velocity is a patch for protocol efficiency

Yes, NoteOn with zero velocity was apparently allowed because of optimization which MuseScore doesn't seem to use. Still using that way of turning notes off should work regardless of making use of running status.

> In the MIDI2.0 standard, this misconception was removed. NoteOn with zero velocity won't work like NoteOff.

If MIDI2.0 makes NoteOff the only supported way to turn note off then I agree that it is better to change it now in MuseScore. I was just confused by the description of this PR which doesn't mention this motivation for the change. Anyway, thanks for clarification!

igorkorsukov (approved):

MarcSabatella:
Agreed, but I still hope we *accept* either form from external devices for MIDI input, because right now I’m not convinced we always do.

psmokotnin:
@MarcSabatella this PR is about event generation only.
For the MIDI input, we accept both types

Jojo-Schmitz:
the midi mtest in Travis fails quite badly

PR #6554, opened 2020-09-15
Fix #310248 - Correct Actual Duration of Split Measures +collect_arti…

worldwideweary:
Resolves: https://musescore.org/en/node/310248

Split measures did a blind reduction of fractions. This should make them more intelligible. For example, 4/4 split in half will be 2/4 instead of 1/2. Another test I ran was 12/8: split so there was [1/8] + [11/8] after and then split that [11/8] at [4/8] and the remainder came out fine as [7/8]. As it was originally, this would've resulted in [11/8] and [1/2] again

I tested this and it works fine, but I didn't extensively test it, so I hope someone else is willing to spend some time
verifying there's no problems. I added +collect_artifacts to the title for this purpose so maybe someone can just download the appImage and check it out

Update: mtests/split "-ref" files needed to be updated, so I manually ran the split actions on the original files and saved over the -ref.mscx files.

### DIFF
--- a/libmscore/splitMeasure.cpp
+++ b/libmscore/splitMeasure.cpp
@@ -104,8 +104,27 @@ void Score::splitMeasure(Segment* segment)
       Fraction ticks2 = measure->ticks() - ticks1;
       m1->setTimesig(measure->timesig());
       m2->setTimesig(measure->timesig());
-      m1->adjustToLen(ticks1.reduced(), false);
-      m2->adjustToLen(ticks2.reduced(), false);
+      ticks1.reduce();
+      ticks2.reduce();
+      // Now make sure this reduction doesn't go 'beyond' the original measure's
+      // actual denominator for both resultant measures.
+      if (ticks1.denominator() < measure->ticks().denominator()) {
+            if (measure->ticks().denominator() % m1->timesig().denominator() == 0) {
+                  int mult = measure->ticks().denominator() / ticks1.denominator();
+                  // *= operator audomatically reduces via GCD, so rather do literal multiplication:
+                  ticks1.setDenominator(ticks1.denominator() * mult);
+                  ticks1.setNumerator(ticks1.numerator() * mult);
+                  }
+            }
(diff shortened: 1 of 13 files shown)


libmscore/splitMeasure.cpp
```diff
@@ -104,8 +104,27 @@ void Score::splitMeasure(Segment* segment)
       Fraction ticks2 = measure->ticks() - ticks1;
       m1->setTimesig(measure->timesig());
       m2->setTimesig(measure->timesig());
-      m1->adjustToLen(ticks1.reduced(), false);
-      m2->adjustToLen(ticks2.reduced(), false);
+      ticks1.reduce();
+      ticks2.reduce();
+      // Now make sure this reduction doesn't go 'beyond' the original measure's
+      // actual denominator for both resultant measures.
```
MarcSabatella, line 110: Haven't tested yet, but was thinking about this last night and at least I finally realize why it was the way it was, and in principle this is the right fix.  You can't always preserve the original denominator, like when splitting 4/4 into 3/8 and 5/8, so the original code was needing to build the durations bottom-up.  Then it made sense to reduce, but it shouldn't have reduced all the way.

Presumably a similar thing happens on join, like combining 4/4 and 3/8, the result cannot be expressed in a denominator less than 8.  But, combining 3/8 and 5/8 will be problematic, an algorithm like this will say, keep the 8, when somehow it might be nice to know 4/4 is probably the desired answer.  Except when it isn't.  Hmm, in that case I guess you can still look at the nominal timesig.

worldwideweary: Hey Marc. Yeah the join functions will be a little bit more complicated in fixing up. I'd hate to leave them alone while fixing this, so I'm willing to take a stab at it, but I also realize it will require a little bit more reasoning behind the decisions here and I'm not sure I'm clear as to what should be done. 

P.S. With this PR, if you were to test Select Similar by Beat filter, it will clearly show the flaw in the old Similar Beat when using the first note of split measures related to regular measures, so these two will go "hand-in-hand" if you will.

PR #6555, opened 2020-09-15
[MU4] Plugins page

RomanPudashkin:


mu4/plugins/view/pluginsmodel.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+int PluginsModel::itemIndexByCodeKey(const QString& codeKey) const
+{
+    for (int i = 0; i < m_plugins.count(); ++i) {
+        if (m_plugins[i].codeKey == codeKey) {
+            return i;
+        }
+    }
+
+    return 0;
```
Eism, line 203: return -1

mu4/plugins/internal/pluginsservice.h
```diff
@@ -0,0 +1,68 @@
...
+#include "modularity/ioc.h"
+#include "io/path.h"
+#include "async/asyncable.h"
+
+namespace mu::plugins {
+class PluginsService : public IPluginsService, public async::Asyncable
+{
+    INJECT(plugins, IPluginsConfiguration, configuration)
+    INJECT(plugins, framework::IFileSystem, fileSystem)
+    INJECT(plugins, framework::INetworkManagerCreator, networkManagerCreator)
```
Eism, line 37: You don't use it.

mu4/plugins/internal/pluginsservice.h
```diff
@@ -0,0 +1,68 @@
...
+
+    async::Channel<PluginInfo> pluginChanged() const override;
+
+private:
+    bool isAccepted(const CodeKey& codeKey, PluginsStatus status) const;
+
+    CodeKeyList installedPlugins() const;
+    void setInstalledPlugins(const CodeKeyList& codeKeyList);
+
+    void startPlugins(const CodeKeyList& codeKeyList);
```
Eism, line 56: Can't find where the implementation of this method is.

framework/global/val.h
```diff
@@ -47,29 +48,31 @@ class Val
...
+
     static Val fromQVariant(const QVariant& var);
 #endif
 
     inline bool operator ==(const Val& v) const { return v.m_val == m_val && v.m_type == m_type; }
 
 private:
-
-    std::string m_val;    //! NOTE In C++17 can be replaced by std::any or std::variant
+    QVariant m_val; //! NOTE In C++17 can be replaced by std::any or std::variant
```
igorkorsukov, line 75: Not a good idea.
This means that there is a strong dependence on Qt.
The very meaning of the `Val` class is that there is no dependence on Qt. Otherwise, instead of `Val`, one could simply use `QVariant`

mu4/plugins/view/pluginsmodel.cpp
RomanPudashkin: fixed

mu4/plugins/internal/pluginsservice.h
RomanPudashkin: fixed

mu4/plugins/internal/pluginsservice.h
RomanPudashkin: fixed

framework/global/val.h
RomanPudashkin: I discussed this issue with Igor in Slack. We decided to leave my current solution as is because in some cases Val must storage some complex types (like QVariantMap) and currently there is no better ability to do this

Eism (approved):

PR #6556, opened 2020-09-15
[MU4] updated CI scripts

igorkorsukov:


mscore/musescore.cpp
```diff
@@ -8333,15 +8333,9 @@ static void initZitaResources()
 
 static void initResources()
 {
-#ifdef Q_OS_MAC
-    Q_INIT_RESOURCE(musescore);
-    Q_INIT_RESOURCE(qml);
-    Q_INIT_RESOURCE(shortcut_Mac);
```
Marr11317, line 8339: no more mac-specific shortcuts?

RobFog: Must be mistake?

igorkorsukov: Separate shortcuts for maс will remain but will be implemented differently.
The current implementation can be confusing.
In the code, we read the shortcuts.xml file, but under mac, we actually read the shortcuts-Mac.xml file, this can be misleading, it is not immediately clear what is happening.
In the new module, shortcuts will be done differently to solve this problem.

.github/workflows/ci_linux.yml
```diff
@@ -0,0 +1,33 @@
+name: CI_Linux
+
+on:
+  pull_request:
```
vpereverzev, line 4: Let's add also a "workflow_dispatch" trigger here, it'll allow us to run builds manually

mscore/musescore.cpp
Marr11317: But this has nothing to do with this PR, you should maybe revert it and take care of it in another PR.

igorkorsukov: You're right. These changes were on a separate branch. They got here only because at the moment the master is not building for Mac without these changes.
The master was broken other PR.
And the lack of CI builds for Mac did not allow it to be revealed in time.
I decided to combine these changes into one PR for practical reasons. I think while the active phase of development is in progress, this is permissible, although I agree that it is not desirable to do so.

vpereverzev (approved):

PR #6558, opened 2020-09-16
updated ci scripts

igorkorsukov:


igorkorsukov:
not there, it's too early

PR #6560, opened 2020-09-18
add extended fermata symbols to MusicXML I/O

rettinghaus:
This PR adds import/export for fermata symbols that were added to MusicXML 3.1 here: https://github.com/w3c/musicxml/pull/152/files

### DIFF
--- a/importexport/musicxml/exportxml.cpp
+++ b/importexport/musicxml/exportxml.cpp
@@ -1722,18 +1722,24 @@ static void fermata(const Fermata* const a, XmlWriter& xml)
       SymId id = a->symId();
       if (id == SymId::fermataAbove || id == SymId::fermataBelow)
             xml.tagE(tagName);
-      // MusicXML does not support the very short fermata nor short fermata (Henze),
-      // export as short fermata (better than not exporting at all)
-      else if (id == SymId::fermataShortAbove || id == SymId::fermataShortBelow
-               || id == SymId::fermataShortHenzeAbove || id == SymId::fermataShortHenzeBelow
-               || id == SymId::fermataVeryShortAbove || id == SymId::fermataVeryShortBelow)
+      else if (id == SymId::fermataShortAbove || id == SymId::fermataShortBelow) {
             xml.tag(tagName, "angled");
-      // MusicXML does not support the very long fermata  nor long fermata (Henze),
-      // export as long fermata (better than not exporting at all)
-      else if (id == SymId::fermataLongAbove || id == SymId::fermataLongBelow
-               || id == SymId::fermataLongHenzeAbove || id == SymId::fermataLongHenzeBelow
-               || id == SymId::fermataVeryLongAbove || id == SymId::fermataVeryLongBelow)
+      }
+      else if (id == SymId::fermataLongAbove || id == SymId::fermataLongBelow) {
             xml.tag(tagName, "square");
+      }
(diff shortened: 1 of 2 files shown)


RobFog:
cc @lvinken

importexport/musicxml/exportxml.cpp
```diff
@@ -2329,21 +2329,27 @@ static void fermatas(const QVector<Element*>& cra, XmlWriter& xml, Notations& no
...
+            else if (id == SymId::fermataVeryLongAbove || id == SymId::fermataVeryLongBelow) {
+                  xml.tag(tagName, "double-square");
+            }
+            else if (id == SymId::fermataLongHenzeAbove || id == SymId::fermataLongHenzeBelow) {
+                  xml.tag(tagName, "double-dot");
+            }
+            else if (id == SymId::fermataShortHenzeAbove || id == SymId::fermataShortHenzeBelow) {
+                  xml.tag(tagName, "half-curve");
+            }
       }
```
cbjeukendrup, line 2353: A `}` is missing here, therefore the build fails.
I guess it was the strange code style that confused you...

You could change it to the following (this way it also fits the best with the code style):
```cpp
static void fermatas(const QVector<Element*>& cra, XmlWriter& xml, Notations& notations)
      {
      for (const Element* e : cra) {
            if (!e->isFermata())
                  continue;
            const Fermata* a = toFermata(e);
            notations.tag(xml);
            QString tagName = "fermata";
            tagName += QString(" type=\"%1\"").arg(a->placement() == Placement::ABOVE ? "upright" : "inverted");
            tagName += color2xml(a);
            SymId id = a->symId();
            if (id == SymId::fermataAbove || id == SymId::fermataBelow)
                  xml.tagE(tagName);
            else if (id == SymId::fermataShortAbove || id == SymId::fermataShortBelow)
                  xml.tag(tagName, "angled");
            else if (id == SymId::fermataLongAbove || id == SymId::fermataLongBelow)
                  xml.tag(tagName, "square");
            else if (id == SymId::fermataVeryShortAbove || id == SymId::fermataVeryShortBelow)
                  xml.tag(tagName, "double-angled");
            else if (id == SymId::fermataVeryLongAbove || id == SymId::fermataVeryLongBelow)
                  xml.tag(tagName, "double-square");
            else if (id == SymId::fermataLongHenzeAbove || id == SymId::fermataLongHenzeBelow)
                  xml.tag(tagName, "double-dot");
            else if (id == SymId::fermataShortHenzeAbove || id == SymId::fermataShortHenzeBelow)
                  xml.tag(tagName, "half-curve");
            }
      }
```

cbjeukendrup (changes requested):
Looks good to me! The only thing is that one `}` is missing in `exportxml.cpp`; see comment at line 2353.

importexport/musicxml/exportxml.cpp
rettinghaus: I got confused, because I introduced this change originally to the master branch, but it should be 3.x. So I just copy-pasted the lines without thinking …

cbjeukendrup: Ah, I see. Now it's OK!

Jojo-Schmitz (approved):

Jojo-Schmitz:
Might be good enough for 3.5.2, but certainly is for 3.6

vpereverzev (approved):

PR #6561, opened 2020-09-18
Fix #12552: Slash in filename on Mac appears as colon

### LINKED ISSUE #12552: Metronome is sending as regular note all the time via MIDI Out device
** Description**
In Mu4 metronome is configured as a regular instrument. But due to using Midi channel 1 and note midi events it is sent via Midi output as a regular midi notes. Metronome should have own midi channel to be independent.
Also reproducible when MuseScore is opened at the same time with Guitar pro (with opened tab)

**To Reproduce**
1. Create new empty score
2. Select Midi output to your device (audio interface, synth etc.)
3. Play the score 
4. Pay attention to the A3 sounding all the time as a metronome

**Expected behavior**
To be discussed

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

cbjeukendrup:
Resolves: https://musescore.org/en/node/12552

On macOS, you can type a slash in a filename. Internally, macOS handles it as a colon, but to the users it normally shows it as a slash. 
However, in MuseScore, this wasn't the case; MuseScore consequently showed a colon. In this PR, I solved this.

There where two things about which I have tried to be very careful:
- I only changed the code which shows filenames to the user, for example in the window title. Code that handles filenames internally should not be touched. Otherwise, it would give problems with file paths of course.
- I made sure that my changes only affect macOS, not other OSs, by using `#ifdef Q_OS_MAC`.

Slash/colon issue fixed, in several places

---

### DIFF
--- a/libmscore/score.cpp
+++ b/libmscore/score.cpp
@@ -4770,7 +4770,13 @@ void MasterScore::setName(const QString& ss)
 
 QString MasterScore::title() const
       {
+#ifdef Q_OS_MAC
+      // macOS has an issue with slashes and colons
+      // https://musescore.org/en/node/12552#comment-1027472
+      return fileInfo()->completeBaseName().replace(':', '/');
+#else
       return fileInfo()->completeBaseName();
+#endif
       }
 
 QString Score::title() const
--- a/libmscore/scorefile.cpp
+++ b/libmscore/scorefile.cpp
@@ -410,8 +410,8 @@ bool MasterScore::saveFile(bool generateBackup)
             }
 #endif
       else {
-           QString fileName = info.completeBaseName() + ".mscx";
-           rv = Score::saveCompressedFile(&temp, fileName, false);
+            QString fileName = info.completeBaseName() + ".mscx";
+            rv = Score::saveCompressedFile(&temp, fileName, false);
             }
 
       if (!rv) {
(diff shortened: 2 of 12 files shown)


Eism:
I don’t think it’s a good way to fix this bug in so many places, unfortunately it’s poorly supported in the future.
Did I understand correctly that the problem appears if the user enters '/' while saving? Can we prohibit '/' input when saving?

cbjeukendrup:
@Eism True, this doesn't make the code easier to understand and to maintain... I considered writing a function for it, but I was not sure *where* I would do that.

After thinking about it another bit, I believe it's actually a problem of macOS. If you would send a file with '/' in the filename to a non-macOS computer, you would just see a colon. I think macOS should stop showing colons as slashes (because that's actually the problem). So I think I'm going to send some feedback about that to Apple...
We can think about it for another moment, but maybe it's the best to close this PR, isn't it?

vpereverzev:
I'd close it because it's a potentially problematic PR for other platforms, however I'd be interested to hear Apple's answer @cbjeukendrup

cbjeukendrup:
@vpereverzev It's okay that you closed it. I sent a message about this to Apple, using their Feedback Assistant, but I didn't get any reaction (I would almost say "as usually"...). I do consider this as a kind of bug from Apple, so I hope they will fix it once.

PR #6563, opened 2020-09-19
[MU4] Build in the MU3 and MU4 modes on CI

igorkorsukov:
and fixed build MU4 mode on Windows with msvc


RomanPudashkin (approved):

build/ci/macos/build.sh
```diff
@@ -2,4 +2,11 @@
 
 echo "Build MuseScore"
 
-make -f Makefile.osx ci  
\ No newline at end of file
+BUILD_MODE=$1
```
vpereverzev, line 5: Do we really need the separate arg here?
Isn't it enough to have only BUILD_UI_MU4 option?

Eism (approved):

build/ci/macos/build.sh
igorkorsukov: This is arg for bash script and BUILD_UI_MU4 option for cmake.
I can't just pass the cmake options through the bash script.
Now the meaning is this, in the bash script is set by the argument in which mode to build. In the bash script, this is implemented by substituting the necessary cmake options.
In general, there may be a more when one option for some modes.

PR #6565, opened 2020-09-21
[MU4] Creating new score from the new score dialog

RomanPudashkin:
- implemented parsing of pages result 

- implemented creating parts

### DIFF
--- a/mu4/notation/internal/masternotation.cpp
+++ b/mu4/notation/internal/masternotation.cpp
@@ -87,7 +87,9 @@ mu::Ret MasterNotation::load(const io::path& path, const INotationReaderPtr& rea
     return ret;
 }
 
-mu::Ret MasterNotation::doLoadScore(Ms::MasterScore* score, const io::path& path, const INotationReaderPtr& reader) const
+mu::Ret MasterNotation::doLoadScore(Ms::MasterScore* score,
+                                    const io::path& path,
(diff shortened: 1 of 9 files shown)


mu4/notation/internal/masternotation.cpp
```diff
@@ -87,7 +87,124 @@ mu::Ret MasterNotation::load(const io::path& path, const INotationReaderPtr& rea
...
+    }
+    instrument.setStringData(instrumentTemplate.stringData);
+    instrument.setSingleNoteDynamics(instrumentTemplate.singleNoteDynamics);
+    return instrument;
+}
+
+void MasterNotation::numberInstrumentNames(Ms::MasterScore* score)
+{
+    std::vector<QString> names;
+    std::vector<QString> firsts;
```
vpereverzev, line 176: It's not very clear what is the purpose of this vector by name

vpereverzev (approved):

mu4/notation/internal/masternotation.cpp
IsaacWeiss: I wrote this code—I guess it's just being moved here? Could call it `firstInstancesOfInstruments` or something similar.

vpereverzev: Yeah, much better

PR #6579, opened 2020-09-25
fix #310869: crash on DPI change in preferences

MarcSabatella:
Resolves: https://musescore.org/en/node/310869

The crash is a null pointer dereference.
We are processing a spin box but referencing a combo box.
This appear to me to have been a copy/paste error.
Comparing against similar code to handle "int" spin boxes,
it seems this line shouldn't be needed at all,
so I have removed it.

### DIFF
--- a/mscore/preferenceslistwidget.cpp
+++ b/mscore/preferenceslistwidget.cpp
@@ -403,12 +403,12 @@ void DoublePreferenceItem::apply()
                   double newValue = _editorComboBox->currentData().toDouble();
                   _initialValue = newValue;
                   PreferenceItem::apply(newValue);
+                  _initialEditorIndex = _editorComboBox->currentIndex();
                   }
             else if (_editorSpinBox) {
                   double newValue = _editorSpinBox->value();
                   _initialValue = newValue;
                   PreferenceItem::apply(newValue);
-                  _initialEditorIndex = _editorComboBox->currentIndex();
                   }
             }
       }


mscore/preferenceslistwidget.cpp
```diff
@@ -408,7 +408,7 @@ void DoublePreferenceItem::apply()
                   double newValue = _editorSpinBox->value();
                   _initialValue = newValue;
                   PreferenceItem::apply(newValue);
-                  _initialEditorIndex = _editorComboBox->currentIndex();
+                  //_initialEditorIndex = _editorComboBox->currentIndex();
```
Jojo-Schmitz, line 411: Why not just removing the line?

SKefalidis: Remove it, it's a copy-paste mistake.

SKefalidis:
The fix is correct :+1: :^)

SKefalidis:
Actually, you should move it over to the combo box part of apply.

SKefalidis (changes requested):
Move it here.
```
else if (_editorComboBox) {
            double newValue = _editorComboBox->currentData().toDouble();
            _initialValue = newValue;
            PreferenceItem::apply(newValue);
        }
```

Otherwise `isModified` will not work properly.

MarcSabatella:
OK, done.  Probably not testable even by hand, unless there really is a double combobox somewhere?

SKefalidis:
> OK, done. Probably not testable even by hand, unless there really is a double combobox somewhere?

You are right @MarcSabatella. I don't think that there is a double combobox somewhere, I added it just in case we need it in the future.

Jojo-Schmitz (approved):

PR #6580, opened 2020-09-25
[MU4] Fixed compiler warnings and move to Qt 5.15

RomanPudashkin:


mu4/instruments/view/instrumentlistmodel.cpp
```diff
@@ -204,7 +204,7 @@ void InstrumentListModel::unselectInstrument(const QString& id)
 
 void InstrumentListModel::swapSelectedInstruments(int firstIndex, int secondIndex)
 {
-    m_selectedInstruments.swap(firstIndex, secondIndex);
+    m_selectedInstruments.swapItemsAt(firstIndex, secondIndex);
```
Jojo-Schmitz, line 207: Requires Qt 5.13 or later, so better protect this with a corresponding `#if'

mu4/plugins/view/pluginsmodel.cpp
```diff
@@ -176,7 +176,7 @@ QStringList PluginsModel::categories() const
         result << plugin.category;
     }
 
-    return result.toList();
+    return result.values();
```
Jojo-Schmitz, line 179: requires Qt 5.15 I believe, , so better protect this with a corresponding `#if`

```
@@ -176,7 +176,11 @@ QStringList PluginsModel::categories() const
        result << plugin.category;
    }

#if (QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)) // ??
    return result.values();
#else
    return result.toList();
#endif
}

void PluginsModel::updatePlugin(const PluginInfo& plugin)
```

mu4/userscores/view/templatesmodel.cpp
```diff
@@ -31,7 +31,7 @@ void TemplatesModel::load()
 
 QStringList TemplatesModel::categoriesTitles() const
 {
-    return m_visibleCategoriesTitles.toList();
+    return m_visibleCategoriesTitles.values();
```
Jojo-Schmitz, line 34: see above

mu4/inspector/models/inspectorlistmodel.cpp
```diff
@@ -141,13 +141,11 @@ void InspectorListModel::removeUnusedModels(const QSet<Ms::ElementType>& newElem
             continue;
         }
 
-        // ToDo for Qt 5.15: QList<Ms::ElementType>::toSet vs. QSet<T>(list.begin(), list.end()) ??
-        QSet<Ms::ElementType> supportedElementTypes
-            = AbstractInspectorModel::supportedElementTypesBySectionType(model->sectionType()).toSet();
+        auto supportedElementTypes = AbstractInspectorModel::supportedElementTypesBySectionType(model->sectionType());
+        QSet<Ms::ElementType> supportedElementTypesSet(supportedElementTypes.begin(), supportedElementTypes.end());
+        supportedElementTypesSet.intersect(newElementTypeSet);
```
Jojo-Schmitz, line 146: Does this for Qt 5.12 too? Or would in deeded to get protected by some `#if`?

mu4/inspector/models/inspectorlistmodel.cpp
```diff
@@ -214,10 +212,13 @@ void InspectorListModel::subscribeOnSelectionChanges()
         }
 
         m_notation->interaction()->selectionChanged().onNotify(this, [this]() {
-            // ToDo for Qt 5.15: QVector<Note*>::fromStdVector() vs. QVector<T>(vector.begin(), vector.end()) ??
-            QVector<Ms::Element*> elements = QVector<Ms::Element*>::fromStdVector(m_notation->interaction()->selection()->elements());
+            QList<Ms::Element*> elements;
+
+            for (Ms::Element* element: m_notation->interaction()->selection()->elements()) {
+                elements << element;
+            }
```
Jojo-Schmitz, line 219: same here

mu4/instruments/view/instrumentlistmodel.cpp
RomanPudashkin: MU4 will use Qt 5.15. This is the final decision :)

igorkorsukov: @Jojo-Schmitz Could you tell me if you think it makes sense to support builds for Qt 5.12, 5.13?

Jojo-Schmitz: 5.13 may not make much sense, 5.12 might

igorkorsukov: And what could be the reasons for maintaining the Qr 5.12 build?

Jojo-Schmitz: > MU4 will use Qt 5.15. This is the final decision :)

Would have been nice it that had been discissed/published in the developers' chat

igorkorsukov: > macOS 10.10 and 10.11
Yes, it was a limiting factor for transferring MuseScore 3 to Qt higher than 5.9

But Apple itself does not support these platforms.
We leave for them MuseScore 3

And we think that they should not hinder the development of MuseScore 4

Jojo-Schmitz: It doesn't hinder so far, if done properly. So far we don't really need any features newer than Qt 5.9 as far as I can tell?

igorkorsukov: Qt 5.15 supports macOS 10.13 (https://doc.qt.io/qt-5/supported-platforms.html), it is still supported by Apple.
MacOS 10.12 is not supported by Apple since October 2019 (https://en.wikipedia.org/wiki/MacOS_Sierra)

Jojo-Schmitz: So we're loosing 3 macOS versions on the way. Mac users won't be amused...

igorkorsukov: > It doesn't hinder so far, if done properly. So far we don't really need any features newer than Qt 5.9 as far as I can tell?
> 

Our main motivation for moving to Qt 5.15 now is that there are a lot of changes in Qt 5.15 compared to Qt 5.9, especially with relationship to Qml.
And this is preparatory, the latest version for Qt 6.0

Now there is an active development (and it will still last at least six months). We can detect and fix all the problems associated with the transition to Qt 5.15 (and there can be many because there are many changes).
If we decide to switch to Qt 5.15 after the release of MuseScore 4, then our users will face all these problems.

Jojo-Schmitz: Yes, sure, using Qt 5.15 for MuseScore 4 is fine, but if and as long as we can keep compatibility with Qt 5.9 with a reasonably effort, that'd be even better. And so far I don't see any reason not to.

igorkorsukov: Unfortunately, the effort is more than desirable.
There will be a lot of `#ifdef` in the code
But even worse, as for Qml, there are many syntax changes.
And support for Qt 5.9 means that we cannot use any of the new features Qt 5.15

Jojo-Schmitz: Quite a few of those `#ifdef`s are in the code already.
What new featured are we using then?

igorkorsukov: image
https://gs.statcounter.com/macos-version-market-share/desktop/worldwide
These are not MuseScore statistics, but statistics on the use of the OS in general

MacOS 10.10 - 2.43%
MacOS 10.11 - 3.82%
MacOS 10.12 - 4.8%

When MuseScore 4 is released (not earlier than half a year), the use of these versions will be even less.

Jojo-Schmitz: Loosing ~11% of the Mac uses is quite a number

igorkorsukov: But we don't lose them :)
They will have MuseScore 3.6

Jojo-Schmitz: Tell them and take the heat ;-)

Jojo-Schmitz:
See also #6588

PR #6581, opened 2020-09-26
Separation of the playback and the audio driver callbacks.

psmokotnin:
In this PR notation playback was separated from audio driver's callback. Synthesizers generate sound directly to the audio and receive messages from playback.

Also:
- MacOs audio driver was changed from signed int 16 to float (32 bit)
- Fixed midi::Event deprecation warnings in NotationPlayback
- added mu::Timer to the framework
- removed MidiSource (midi as audio source) from audio engine


Jojo-Schmitz:
a rebase should fix the Travis mtest fail

psmokotnin:
> a rebase should fix the Travis mtest fail

Thanks! Anyway, this PR won't merge in that state and will be continued.

PR #6583, opened 2020-09-26
debug builds crash on scores with drum staff

Jojo-Schmitz:
Not needed for the master branch, which doesn't have that `IF_ASSERT_FAILED`

### DIFF
--- a/audio/midi/fluid/voice.cpp
+++ b/audio/midi/fluid/voice.cpp
@@ -1816,7 +1816,7 @@ void Sample::optimize()
             s->loopstart = s->start;
             }
 
-      IF_ASSERT_FAILED(s->loopend <= s->end) {
+      if (s->loopend > s->end) {
             s->loopend = s->end;
             }
 


Jojo-Schmitz:
The Travis CO fail seems to be due to the PR trying to upload nightly builds, which of course fails

audio/midi/fluid/voice.cpp
```diff
@@ -1816,7 +1816,7 @@ void Sample::optimize()
             s->loopstart = s->start;
             }
 
-      IF_ASSERT_FAILED(s->loopend <= s->end) {
+      if (s->loopend <= s->end) {
```
RomanPudashkin, line 1819: This is not an equal replacement

Condition
IF_ASSERT_FAILED(s->loopend <= s->end)

Equals to
if (!(s->loopend <= s->end))

Or just
if (s->loopend > s->end)

Jojo-Schmitz: oops, thanks for catching

PR #6584, opened 2020-09-26
[MU3] Fix #310821: Chord Symbol on drum stave sounds as percussion

Jojo-Schmitz:
This is just a (very) partial fix, it does playback chord symbols on drum staves as Piano, but doesn't show the correct sound in Mixer (shows "Standard" rather then "Grand Piano"), doesn't allow changes to the sound to another instrument, on trying to change it reverts the sound to Percussion and (probably) doesn't fix existing scores.

See https://musescore.org/en/node/310821

### DIFF
--- a/libmscore/instrument.h
+++ b/libmscore/instrument.h
@@ -189,6 +189,8 @@ public:
       bool userBankController() const           { return _userBankController; }
       void setUserBankController(bool val);
 
+      bool isHarmonyChannel() const { return _name == Channel::HARMONY_NAME; }
+
       QList<NamedEventList> midiActions;
       QList<MidiArticulation> articulation;
 
--- a/libmscore/part.cpp
+++ b/libmscore/part.cpp
@@ -606,6 +606,8 @@ void Part::updateHarmonyChannels(bool isDoOnInstrumentChanged, bool checkRemoval
             Channel* c = new Channel(*instr->channel(0));
             // default to program 0, which is piano in General MIDI
             c->setProgram(0);
+            if (c->bank() == 128) // drumset?
+                  c->setBank(0);
             c->setName(Channel::HARMONY_NAME);
             instr->appendChannel(c);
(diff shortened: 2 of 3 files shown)


MarcSabatella:
Unfortunately, I don't really know enough how these things are organized - in the soundfont or in Fluid - to be confident of how to fix it better.  I might guess, though, that if the Mixer isn't displaying it correctly, that's the Mixer's fault, not the chord symbol channel code's fault.  That is, probably the Mixer never was designed to handle channels on different banks.  And that could also be why changing it doesn't really work either.

Jojo-Schmitz:
The Travis Ci fail seems to be due to the PR trying to upload nightly builds, which of course fails

dmitrio95:
As far as I can tell, the Mixer issues you describe are caused by Mixer itself, not by something being wrong in the internal channels' structure. Mixer seems to assume (at least [here](https://github.com/musescore/MuseScore/blob/4983887d71bbab2197e00717a2eec179a009d2c1/mscore/mixer/mixerdetails.cpp#L166)) that a percussion instrument cannot contain non-percussion channels. This assumption is not correct anymore after harmony channels have been introduced. So this seems like something to be corrected in Mixer as it just doesn't handle this new channels configuration well.

Jojo-Schmitz:
OK, just how/where exactly?

dmitrio95:
Well, the easiest option I see is making harmony channels always non-percussive:
https://github.com/dmitrio95/MuseScore/commit/1ef5315a2df314d329aecd223457e7c257e74acb

This would disallow using percussion sounds in them but this doesn't seem to be needed anyway.

A more generic solution which would allow separate handling of percussive and non-percussive channels in one instrument seems to require more changes: `useDrumset` is now a property of instrument rather than individual channel.

Jojo-Schmitz:
Works perfectly, thanks!

PR #6586, opened 2020-09-28
Fix #309857 - Divider not removed after undo of deleting a vertical frame

njvdberg:
Resolves: https://musescore.org/en/node/309857

Remove SystemDividers on a System and the previous System is the system is a VBox.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -1536,8 +1536,15 @@ static void layoutPage(Page* page, qreal restHeight)
             System* s1 = page->systems().at(i);
             System* s2 = page->systems().at(i+1);
             s1->setDistance(s2->y() - s1->y());
-            if (s1->vbox() || s2->vbox() || s1->hasFixedDownDistance())
+            if (s1->vbox() || s2->vbox() || s1->hasFixedDownDistance()) {
+                  if (s2->vbox()) {
+                        checkDivider(true, s1, 0.0, true);      // remove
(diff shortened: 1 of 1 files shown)


libmscore/layout.cpp
```diff
@@ -1536,8 +1536,15 @@ static void layoutPage(Page* page, qreal restHeight)
             System* s1 = page->systems().at(i);
             System* s2 = page->systems().at(i+1);
             s1->setDistance(s2->y() - s1->y());
-            if (s1->vbox() || s2->vbox() || s1->hasFixedDownDistance())
+            if (s1->vbox() || s2->vbox() || s1->hasFixedDownDistance()) {
+                  if (s2->vbox()) {
+                        checkDivider(true, s1, 0.0, true);      // remove
```
vpereverzev, line 1541: Well...I'm not quite confident about this piece of code

What is happening here?

njvdberg: When layout a page, <code>layoutPage()</code> looks for systems which can have some extra space below the system, to spread the system over the page. So, only systems which
 -1 are **not** a vertical frame
 -2 doesn't have a vertical frame below.
 -3 doesn't have a fixed spacing.
These systems are not included for further processing, including adding/removing system dividers.

However, under some conditions, a systems is excluded from further processing **but** does have a system divider, e.g. when undoing a delete of a vertical box.

This patch makes sure the system dividers are removed from a system if the **next** system is a vertical frame. Also any divider on the vertical box are removed. too.

PR #6588, opened 2020-09-28
fix compiler warnings (master)

Jojo-Schmitz:
as seen in MinGW and/or MSVC


Jojo-Schmitz:
Travis CI fails since about a week, not this PR's fault

framework/midi/midievent.h
```diff
@@ -84,13 +84,13 @@ struct Event {
         setOpcode(opcode);
     }
 
-    [[deprecated]] Event(channel_t ch, EventType type)
+    /*[[deprecated]]*/ Event(channel_t ch, EventType type)
```
psmokotnin, line 87: This constructor will be deleted in MU4. That's why it's marked as deprecated and must not be used.

framework/midi/midievent.h
```diff
@@ -231,15 +231,15 @@ struct Event {
         return 0;
     }
 
-    [[deprecated]] EventType type() const
+    /*[[deprecated]]*/ EventType type() const
```
psmokotnin, line 234: This constructor will be deleted in MU4. That's why it's marked as deprecated and must not be used.

framework/midi/midievent.h
```diff
@@ -231,15 +231,15 @@ struct Event {
...
     {
         if (isChannelVoice()) {
             return static_cast<EventType>(opcode() << 4);
         }
         return EventType::ME_INVALID;
     }
 
-    [[deprecated]] void setType(EventType type)
+    /*[[deprecated]]*/ void setType(EventType type)
```
psmokotnin, line 242: This method will be deleted in MU4. That's why it's marked as deprecated and must not be used.

framework/midi/midievent.h
```diff
@@ -636,15 +636,15 @@ struct Event {
...
+        m_data[0] |= uint32_t(value);
     }
 
     void setPerNoteReset(bool value)
     {
         assertOpcode({ PerNoteManagement });
         assertMessageType({ ChannelVoice20 });
         m_data[0] &= 0xFFFFFFFE;
-        m_data[0] |= value;
+        m_data[0] |= uint32_t(value);
```
psmokotnin, line 647: use static_cast instead of old C-style cast

framework/midi/midievent.h
```diff
@@ -985,13 +985,13 @@ struct Event {
         // simple bit shift
         size_t scaleBits = dstBits - srcBits;
         uint32_t bitShiftedValue = srcVal << scaleBits;
-        uint32_t srcCenter = 2 ^ (srcBits - 1);
+        uint32_t srcCenter = uint32_t(2 ^ (srcBits - 1));
```
psmokotnin, line 988: use static_cast instead of old C-style cast

framework/midi/midievent.h
```diff
@@ -985,13 +985,13 @@ struct Event {
...
         if (srcVal <= srcCenter) {
             return bitShiftedValue;
         }
         // expanded bit repeat scheme
         size_t repeatBits = srcBits - 1;
-        uint32_t repeatMask = (2 ^ repeatBits) - 1;
+        uint32_t repeatMask = uint32_t((2 ^ repeatBits) - 1);
```
psmokotnin, line 994: use static_cast instead of old C-style cast

psmokotnin (changes requested):

framework/midi/midievent.h
Jojo-Schmitz: It currently is though, and as such leads to compiler warnings.
Marking this as a ToDo is fine, causing compiler warnings is not

framework/midi/midievent.h
Jojo-Schmitz: dito

framework/midi/midievent.h
Jojo-Schmitz: dito

framework/midi/midievent.h
Jojo-Schmitz: OK, just need to figure the exact syntax ;-)

framework/midi/midievent.h
Jojo-Schmitz: dito

framework/midi/midievent.h
Jojo-Schmitz: dito

framework/midi/midievent.h
Jojo-Schmitz: Furthermore it seems the build on macOS doesn't like that `[[deprecated]]` syntax at all and gets thrown off track by them quite miserably, causing the development builds not to succeed?

Jojo-Schmitz: And indeed `[[deprecated]]` is a C++14 feature (see https://en.cppreference.com/w/cpp/language/attributes/deprecated), but we're not using that (yet), esp. on Mac (but also Linux and MINGW) we're using `-std=c++11`

Jojo-Schmitz: Maybe using `QT_DEPRECATED`, `QT_DEPRECATED_X(sometext)`, `QT_DEPRECATED_VARIABLE` or `QT_DEPRECATED_CONSTRUCTOR` for this instead? At least then it should be working for any compiler?

psmokotnin: > so many warnings

You can disable deprecation warning in your build settings. Now we are working on updating the code with future remove these deprecated methods. 

> build on macOS

I'm working on macOS, everything is fine

> C++14 feature 

We are moving to C++17 soon. C++11 just ignores it as unknown attribute

> QT_DEPRECATED

This class does not use Qt at all.

Jojo-Schmitz: I know I could disable warnings, but would miss new ones (like with Qt 5.15) that just got introduced recently.
I know that this class doesn't use Qt, but it could use that macro. I tried, it works (but of course still emits a warning)

RomanPudashkin: I already fixed most of those warnings, but I have some troubles with building tests. Later I will return to my PR and do something with tests
https://github.com/musescore/MuseScore/pull/6580

Jojo-Schmitz:
No idea why that CI_Linux / build_mu4 fails (in generating a checksum of the AppImage)? Can't really imaging this being my fault.

framework/midi/midievent.h
Jojo-Schmitz: Reverted to `[[deprecated]]`

thirdparty/google_analytics/ganalytics.cpp
```diff
@@ -247,11 +247,18 @@ QString GAnalytics::Private::getSystemInfo()
...
-#endif
+    case QSysInfo::MV_10_13:
+        os = "Macintosh; Mac OS 10.13";
+        break;
+    case QSysInfo::MV_10_14:
+        os = "Macintosh; Mac OS 10.14";
+        break;
+    case QSysInfo::MV_10_15:
+        os = "Macintosh; Mac OS 10.15";
+        break;
```
mattmcclinch, line 261: Nice try, but these constants do not exist, and so this fails to build on macOS. In fact, the existing `QSysInfo` constants are deprecated. Better to rewrite the function like this:
```
#ifdef Q_OS_MAC
/**
 * Only on Mac OS X
 * Get the Operating system name and version.
 * @return os   The operating system name and version in a string.
 */
QString GAnalytics::Private::getSystemInfo()
{
    auto osType = QOperatingSystemVersion::currentType();
    auto version = QOperatingSystemVersion::current();
    QString os;
    switch (osType)
    {
    case QOperatingSystemVersion::MacOS:
        os = QString("Macintosh; Mac OS ");
        break;
    case QOperatingSystemVersion::IOS:
        os = QString("iPhone; iOS ");
        break;
    default:
        break;
    }
    os += QString("%1.%2").arg(version.majorVersion()).arg(version.minorVersion());
    return os;
}
#endif
```

Jojo-Schmitz: Hmm, I'm pretty sure I looked them up in the Qt 5.15.1 header files?
Will look again...

Jojo-Schmitz: But feel free to PR your proposal ;-)

Jojo-Schmitz: OK, you're right, those don't seem to exist. But your proposal doesn't build either, there seems to be no such thing as a `class QOperatingSystemVersion`.

mattmcclinch: It exists. You just need to `#include <QOperatingSystemVersion>`.

mattmcclinch: As far as I know, there are no CI tests to make sure that MuseScore builds on macOS. Otherwise, it would have failed.

Jojo-Schmitz: Ops, `#include <QOperatingSystemVersion>` helps there...

Jojo-Schmitz: Check #6647

PR #6589, opened 2020-09-29
fix #311036 Mtest fix

Eism:
Resolves: https://musescore.org/en/node/311036

### DIFF
--- a/mtest/CMakeLists.txt
+++ b/mtest/CMakeLists.txt
@@ -43,6 +43,7 @@ if (MSVC)
 endif (MSVC)
 
 QT5_ADD_RESOURCES(qrc_files ${PROJECT_SOURCE_DIR}/mtest/mtest.qrc
+      ${PROJECT_SOURCE_DIR}/framework/fonts/fonts_Leland.qrc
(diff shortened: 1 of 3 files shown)


RomanPudashkin (approved):

Jojo-Schmitz:
Sorry, no cigar. While this fix certainly is good and needed, it does not fix the mtests, they still fail the very same way.

mattmcclinch:
The problem can be solved by moving [this line](https://github.com/musescore/MuseScore/blob/master/mu4/plugins/api/qmlplugin.cpp#L28) inside the body of the `if (!MuseScoreCore::mscoreCore) { ... }` block. The stub `MuseScoreCore` object should only be constructed if `MuseScoreCore::mscoreCore` is null, because the `MuseScoreCore` constructor overwrites the value of `MuseScoreCore::mscoreCore` no matter what. This means that a subsequent call to `MuseScoreCore::mscoreCore->currentScore()` will now return `nullptr`, which is why the test (and plugins in general) are failing.

Eism:
@mattmcclinch Great job. That's right!

Jojo-Schmitz:
That seems to be it, please merge ASAP!

PR #6602, opened 2020-10-02
Fix #308589 - Tuplet number

njvdberg:
Resolves: https://musescore.org/en/node/308589

Set the scaling of the text element of the tuplet to <code>smallNoteMag</code> if all <code>ChordRest</code>'s of the tuplet are small.
This is the same approach as used for the beam of the tuplet.

### DIFF
--- a/libmscore/textbase.cpp
+++ b/libmscore/textbase.cpp
@@ -881,7 +881,7 @@ QFont TextFragment::font(const TextBase* t) const
       font.setItalic(format.italic());
       Q_ASSERT(m > 0.0);
 
-      font.setPointSizeF(m);
+      font.setPointSizeF(m * t->mag());
       return font;
       }
 
--- a/libmscore/tuplet.cpp
+++ b/libmscore/tuplet.cpp
@@ -192,6 +192,16 @@ void Tuplet::layout()
                   _number->setXmlText(QString("%1").arg(_ratio.numerator()));
             else
                   _number->setXmlText(QString("%1:%2").arg(_ratio.numerator()).arg(_ratio.denominator()));
+
+            bool small = true;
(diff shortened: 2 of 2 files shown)


libmscore/textbase.cpp
```diff
@@ -881,7 +881,7 @@ QFont TextFragment::font(const TextBase* t) const
       font.setItalic(format.italic());
       Q_ASSERT(m > 0.0);
 
-      font.setPointSizeF(m);
+      font.setPointSizeF(m * t->mag());
```
MarcSabatella, line 884: It seems this has the effect of also scaling other text that has a "mag" associated with it.  This could be good, could be bad,  See https://musescore.org/en/node/312109

njvdberg: Before this change <code>TextBase</code> simply ignored the scaling at all. The only way to the size of the text was changing the font size, which was the original "complaint" in [the original issue](https://musescore.org/en/node/308589). Now other text annotations are scaled with the note indeed. However, solving this is pretty simple (1 hour? I made already a simple prove of concept) but to decide which elements are supposed to scale with the notes and which element note, I'm not that sure that will be that easy.

MarcSabatella: Sure, that's why I'm hoping for more discussion in the forum, to arrive at a consensus.  Then, I think, all we need to do is make sure we have appropriate "mag" functions defined for the element types.  For instance, for fingering, it could simply return the staff mag.

njvdberg: That was indeed what I tried and that works fine 😃

Marr11317: Did you ask @octophonie? He might have some thought on that.

PR #6510, opened 2020-08-28
Fix #298849: The 2nd note in a passage of grace notes will not tie to the main chord

mattmcclinch:
Resolves: https://musescore.org/en/node/298849.

At this point in `searchTieNote()`, `chord`is not pointing to `note->chord()`, but rather to `note->chord()->parent()`. What we are really interested in is the grace index of `note->chord()`, and not that of the parent chord.

PR #6610, opened 2020-10-03
Fix #310602 - Adding a tie from grace note to note goes wrong way at times

njvdberg:
Resolves: https://musescore.org/en/node/310602

use the <code>graceIndex</code> of the selected grace note instead of the parent chord.


mattmcclinch:
This does basically the same thing as #6510. One or the other should be merged, but not both.

njvdberg:
Indeed, I missed PR #6510 when I was looking into this issue. Apparently, the two issue were duplicates. I'll close this one.

PR #6615, opened 2020-10-05
MIDI UMP fixes

psmokotnin:
Small fixes for MIDI UMP event class

### DIFF
--- a/framework/midi/midievent.h
+++ b/framework/midi/midievent.h
@@ -23,6 +23,10 @@
 #include <set>
 #include "audio/midi/event.h"
 
+#ifndef UNUSED
+#define UNUSED(x) (void)x;
+#endif
+
 namespace mu {
 namespace midi {
 using channel_t = uint8_t;
@@ -277,6 +281,7 @@ struct Event {
             break;
         default: assert(false);
         }
+        return 0;
     }
 
     void setNote(uint8_t value)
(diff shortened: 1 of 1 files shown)


framework/midi/midievent.h
```diff
@@ -23,6 +23,10 @@
 #include <set>
 #include "audio/midi/event.h"
 
+#ifndef UNUSED
+#define UNUSED(x) (void)x;
```
Jojo-Schmitz, line 27: Why not using `Q_UNUSED()`?

igorkorsukov: We try not to use Qt where it is not used.

Jojo-Schmitz: and reinventing the wheel when not needed?

igorkorsukov: We really want to limit the use of Qt

Jojo-Schmitz: Why?

igorkorsukov: We want to create a platform that is Qt independent and has an interface built using Qt.
This will give us the freedom to make decisions and use this platform.
For example, maybe we will build for the web, but Qt is not very suitable for the web because it is too large.

Jojo-Schmitz: OK, I see. I've also amended my #6588 to not use `QT_DEPRECATED` but reverted it to using `[[deprecated]]`, also because we now switched to C++17 anyway

PR #6628, opened 2020-10-06
[MU3] Added set build config from CI

igorkorsukov:
Added set build config from CI


MarcSabatella:
The effect of this seems to be that if I build MuseScore myself, I get a "release" build config now.  Which is actually kind of cool as far as I am concerned, but was unexpected by me.  If this was unintentional and you change it back, could you let me know how to circumvent it and keep release?  I guess customizing MUSESCORE_BUILD_CONFIG in my cmake settings within QtCreator (for builds done that way, obviously)?

igorkorsukov:
> The effect of this seems to be that if I build MuseScore myself, I get a "release" build config now. Which is actually kind of cool as far as I am concerned, but was unexpected by me. If this was unintentional and you change it back, could you let me know how to circumvent it and keep release? I guess customizing MUSESCORE_BUILD_CONFIG in my cmake settings within QtCreator (for builds done that way, obviously)?

oops this is a bug, I tested and miss this change.
Thank you, I will fix it.

igorkorsukov:
> I guess customizing MUSESCORE_BUILD_CONFIG in my cmake settings within QtCreator (for builds done that way, obviously)?

yes, you can customize this setting in the QtCreator

PR #6634, opened 2020-10-08
fix #311405 Workspace reset to Basic

Eism:
Resolves: https://musescore.org/en/node/311405

Translations for the installed languages are available for default workspaces(Basic, Advanced). Translations are not available for user-created workspaces. Therefore, you need to check the availability of translations. If there is no translation, then you need to search by name, otherwise by translation.

### DIFF
--- a/mscore/workspace.cpp
+++ b/mscore/workspace.cpp
@@ -230,7 +230,7 @@ void MuseScore::changeWorkspace(Workspace* p, bool first)
 
       connect(getPaletteWorkspace(), &PaletteWorkspace::userPaletteChanged, WorkspacesManager::currentWorkspace(), QOverload<>::of(&Workspace::setDirty), Qt::UniqueConnection);
 
-      preferences.setPreference(PREF_APP_WORKSPACE, p->translatableName());
+      preferences.setPreference(PREF_APP_WORKSPACE, p->id());
       emit mscore->workspacesChanged();
       }
 
@@ -272,10 +272,16 @@ bool WorkspacesManager::isDefaultEditedWorkspace(Workspace* workspace)
 void WorkspacesManager::initCurrentWorkspace()
       {
       initWorkspaces();
-      m_currentWorkspace = findByName(preferences.getString(PREF_APP_WORKSPACE));
+
+      QString workspaceName = preferences.getString(PREF_APP_WORKSPACE);
+      m_currentWorkspace = findByName(workspaceName);
       Q_ASSERT(!workspaces().empty());
-      if (m_currentWorkspace == 0)
-            m_currentWorkspace = workspaces().at(0);
+      if (!m_currentWorkspace) {
+          m_currentWorkspace = findByTranslatableName(workspaceName);
+          if (!m_currentWorkspace) {
+              m_currentWorkspace = workspaces().at(0);
+          }
+      }
       }
 
 void WorkspacesManager::remove(Workspace* workspace)
@@ -315,6 +321,11 @@ Workspace::Workspace()
       _saveTimer.setInterval(0);
       _saveTimer.setSingleShot(true);
       connect(&_saveTimer, &QTimer::timeout, this, &Workspace::ensureWorkspaceSaved);
+      }
+
+QString Workspace::id() const
+      {
+      return !_translatableName.isEmpty() ? _translatableName : _name;
       }
 
 //---------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Jojo-Schmitz:
I'm afraid this still doesn't work, in my self-buil it still reverts from Erweitert (Advanced) to Einfach (Basic), regardless of MuseScoreDevelopment3.ini having `workspace=Advanced`

Jojo-Schmitz:
Maybe we need the same mechanism on reading the workspace setting?

Jojo-Schmitz:
Testing...

Jojo-Schmitz:
This seems to work for me!

mscore/workspace.cpp
```diff
@@ -230,7 +230,8 @@ void MuseScore::changeWorkspace(Workspace* p, bool first)
 
       connect(getPaletteWorkspace(), &PaletteWorkspace::userPaletteChanged, WorkspacesManager::currentWorkspace(), QOverload<>::of(&Workspace::setDirty), Qt::UniqueConnection);
 
-      preferences.setPreference(PREF_APP_WORKSPACE, p->translatableName());
+      QString workspaceName = !p->translatableName().isEmpty() ? p->translatableName() : p->name();
```
Jojo-Schmitz, line 233: would it make sense to set `_translatableName` to `_name` if it is empty?

mscore/workspace.cpp
```diff
@@ -272,10 +273,16 @@ bool WorkspacesManager::isDefaultEditedWorkspace(Workspace* workspace)
...
+      m_currentWorkspace = findByName(workspaceName);
       Q_ASSERT(!workspaces().empty());
-      if (m_currentWorkspace == 0)
-            m_currentWorkspace = workspaces().at(0);
+      if (!m_currentWorkspace) {
+          m_currentWorkspace = findByTranslatableName(workspaceName);
+          if (!m_currentWorkspace) {
+              m_currentWorkspace = workspaces().at(0);
+          }
+      }
```
Jojo-Schmitz, line 285: For the benefit of an easier cherry-pick into master I'll let this coding style pass ;-)

Jojo-Schmitz:
the bad news is that the vtests fail, not sure why, but none of the images have been created apparently

mscore/workspace.cpp
Eism: I don't think this is a good idea because in the initialization of workspaces it is explicitly stated:
`const bool translate = isDefault || isEditedDefault;`
`if (translate)
     p->setTranslatableName(name);`

Let's not change the logic in the release. Just fix this bug

Jojo-Schmitz:
Travis failed too, a SIGSEGV in an mtest, see https://travis-ci.org/github/musescore/MuseScore/jobs/733958956#L4382

mscore/workspace.cpp
Jojo-Schmitz: OK

Jojo-Schmitz:
#6637 might be better

Jojo-Schmitz:
(removed bogus comment)

Jojo-Schmitz:
vtest and mtests fixed ;-)

Jojo-Schmitz:
But still that issue when using English (US) and creating and modifying a custom workspace :-(

Eism:
> But still that issue when using English (US) and creating and modifying a custom workspace :-(

But nothing has changed since the last build. Only the correct display of the current workspace in the ComboBox.

Jojo-Schmitz:
Yes, noticed. That for some reason seems to have fixed the vtest/mtest issues

Jojo-Schmitz:
What about https://github.com/musescore/MuseScore/blob/6339f829c794a6c13b4883c7e2c1ce0aeed33a20/mscore/musescore.cpp#L520 ?

Jojo-Schmitz:
Yes!, That seems to be it!

Eism:
> What about
> 
> https://github.com/musescore/MuseScore/blob/6339f829c794a6c13b4883c7e2c1ce0aeed33a20/mscore/musescore.cpp#L520
> 
> 
> and/or
> https://github.com/musescore/MuseScore/blob/6339f829c794a6c13b4883c7e2c1ce0aeed33a20/mscore/workspace.cpp#L136

I really don't want to change this functionality in a release. This way we can completely break up workspaces.

PR #6646, opened 2020-10-09
fix compiler warning reg. deprecation in a `+` operator

Jojo-Schmitz:
in a better way than my attempt in #6548, which then got reverted in #6631, actually in the same way as used for the corresponding `=` operator.

Eliminates 115 warnings ;-)


Jojo-Schmitz:
Hmm, the mtests are not happy with this, crash in tst_palettes and tst_workspace and report a bogus `free()`  :-(

igorkorsukov:
> Hmm, the mtests are not happy with this, crash in tst_palette and tst_workspace and report a `free()` on an invalid pointer :-(
> 
> Ah, relief. those are not my fault ;-)

I don't know why there was an error, but I am going to look and fix it tomorrow

PR #6647, opened 2020-10-09
Revert a change to google analytics that causes issues on Mac

Jojo-Schmitz:
to fix an error I introduced in #6588. See also https://musescore.org/en/node/311571

Also give credit where credit is due (including a rename of the directory).
Disable deprecation warnings for Apple too.


Jojo-Schmitz:
Mac build fails, but for an entirely different reason, not related to this PR as far as I can tell

mattmcclinch:
We probably want this function to return the same values as it did before, since it is used to format the HTTP header for sending Google Analytics results. The changes to the macOS side accomplish this; the changes to the Windows side do not.

Jojo-Schmitz:
I'll have another look tomorrow evening or Sunday. Unless are getting ahead, which I wouldn't mind

Jojo-Schmitz:
I've removed the changes for Windows

Jojo-Schmitz:
Alternative approach would be to revert that part of #6588, as this is third party code we probably should not be modifying (without an urgent need) anyway

Jojo-Schmitz:
Guess I'd better squash them. Unless we want those changes from the 1st commit, in which case I'd need to amend the 2nd. Just let me know

mattmcclinch:
The only reason to modify this third party code would be to get rid of the deprecation warnings, and if we wanted to do that properly, it would probably involve changes to more than just this macOS-specific code. And if we were thinking about sending changes upstream, we would need to make sure only to use QOperatingSystemVersion `#if (QT_VERSION >= QT_VERSION_CHECK(5, 9, 0))`.

Jojo-Schmitz:
The deprecation warnings had been taken case of long ago, by disabling them in CMakeLists.txt

mattmcclinch:
I don't know which version of Xcode you use on your Mac, but I get 33 deprecation warnings for google_analytics.

Jojo-Schmitz:
I don't use Xcode at all (als this problem hadn't happened). And indeed the warnings are disabled only for MSVC and MinGW (in thirdparty/google_analytics/CmakeLists.txt). Feel free to (tell me how to) add ignoring those for Xcode too.

mattmcclinch:
Where it says
```
    if (MINGW)
        set_target_properties( google_analytics PROPERTIES
          COMPILE_FLAGS "-Wno-deprecated-declarations"
        )
    endif (MINGW)
```
changing `if (MINGW)` to `if (MINGW OR APPLE)` seems to do the trick.

Jojo-Schmitz:
Easy enough :-), I'll add it.

mattmcclinch:
Great! But don't you want to squash the 2 commits?

Jojo-Schmitz:
Patience ;-)

igorkorsukov (approved):

PR #6651, opened 2020-10-11
Fixed unexpected geometry stretching in "Edit style" dialog

vpereverzev:
The sizing of "Header/Footer", "Fretboard diagramm" and "Measure" pages were weird. it turned out that when trying to adjust the page size, the actual size of the content was not explicit, resulting in unexpected behaviour.

Before:
изображение
изображение
изображение

After:
Peek 2020-10-11 16-08

### DIFF
--- a/mscore/editstyle.ui
+++ b/mscore/editstyle.ui
@@ -1670,7 +1670,7 @@
        </layout>
       </widget>
       <widget class="QWidget" name="PageHeaderFooter">
-       <layout class="QGridLayout" name="gridLayout_10">
+       <layout class="QGridLayout" name="gridLayout_10" rowstretch="1,1,1" columnstretch="0">
         <property name="leftMargin">
          <number>8</number>
         </property>
@@ -1690,6 +1690,9 @@
          <widget class="QGroupBox" name="showFooter">
           <property name="title">
            <string>Footer Text</string>
+          </property>
+          <property name="flat">
+           <bool>false</bool>
           </property>
           <property name="checkable">
            <bool>true</bool>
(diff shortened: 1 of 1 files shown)


MarcSabatella:
Do we know this does the right thing on high and low DPI displays?

Jojo-Schmitz:
I've seen this happening, and that on a low DPI screen (~100DPI)

MarcSabatella:
FWIW, 100 DPI is not "low" DPI, it's the standard.  It's more the 72 DPI TV sets I am concerned about.

On my relatively high DPI system, this "works", but it looks pretty odd, the boxes are on the small side, and they don't expand with the dialog.  Instead, the empty space around them expands:

Screenshot 2020-10-11 at 11 59 07 AM

Also, if I scroll through the list of the styles at least, the widget gets narrower and wider on its own which looks *really* odd.  That is, with "Score" selected, the list widget is narrow, but with "Measure Numbers" selected, it gets wider.  Then narrower again if I select "System".  But, it's that way in the current 3.5.1 release as well.

Jojo-Schmitz:
Same for me, that not expanding. Not the reggitting wider/narrower though

njvdberg:
I can live with the odd behavior with the list of styles getting wider or narrower but the strange scaling of the "Header, Footer" page is more annoying.

vpereverzev:
It may look strange, but nevertheless, it is the lesser of evils. The scaling of the input fields is not so critical considering that the input fields have built-in scroll bars

Peek 2020-10-12 20-32

Anyway, this is the only know working option that I know at the moment

Jojo-Schmitz:
It may not be perfect, but certainly is good enough

PR #6652, opened 2020-10-11
Fix #290946, #309133: macOS Dark Mode improvements [MU3]

cbjeukendrup:
Resolves: [#290946](https://musescore.org/en/node/290946) and [#309133](https://musescore.org/en/node/309133)

This PR actually solves two issues. 

1.   When you chose the Dark theme in _MuseScore_ (no matter whether the _system_ is in Dark mode or not), the window title bars remained light. Now, they will be light with the light theme and dark with the dark theme. 
    Window Title Bar Before-After

2.   Once, I suggested to let MuseScore automatically switch its theme when the system theme is changed. Now, I got this working, without breaking things on older/other operating systems. 
    macOS Dark Mode On-Off

### DIFF
--- a/mscore/macos/cocoabridge.mm
+++ b/mscore/macos/cocoabridge.mm
@@ -31,3 +31,9 @@ bool  CocoaBridge::isSystemDarkTheme()
     NSString *osxMode = [[NSUserDefaults standardUserDefaults] stringForKey:@"AppleInterfaceStyle"];
     return ([osxMode isEqualToString:@"Dark"]);
 }
+
+void  CocoaBridge::setWindowAppearanceIsDark(bool flag)
+{
+    if (@available(macOS 10.14, *))
+        [NSApp setAppearance:[NSAppearance appearanceNamed:flag ? NSAppearanceNameDarkAqua : NSAppearanceNameAqua]];
+}
--- a/mscore/macos/cocoabridge.mm
+++ b/mscore/macos/cocoabridge.mm
@@ -36,20 +36,27 @@ void CocoaBridge::removeObservers()
             [[NSDistributedNotificationCenter defaultCenter] removeObserver:darkModeObserverToken];
       }
 
-void CocoaBridge::setAllowsAutomaticWindowTabbing(bool flag)
-      {
-      if ([NSWindow respondsToSelector:@selector(allowsAutomaticWindowTabbing)])
-            [NSWindow setAllowsAutomaticWindowTabbing: flag];
-      }
-
 bool CocoaBridge::isSystemDarkTheme()
       {
       NSString *osxMode = [[NSUserDefaults standardUserDefaults] stringForKey:@"AppleInterfaceStyle"];
       return ([osxMode isEqualToString:@"Dark"]);
       }
 
+bool CocoaBridge::isSystemDarkModeSupported()
+      {
+      if (@available(macOS 10.14, *))
+            return true;
+      return false;
+      }
+
 void CocoaBridge::setWindowAppearanceIsDark(bool flag)
       {
       if (@available(macOS 10.14, *))
             [NSApp setAppearance:[NSAppearance appearanceNamed:flag ? NSAppearanceNameDarkAqua : NSAppearanceNameAqua]];
+      }
+
+void CocoaBridge::setAllowsAutomaticWindowTabbing(bool flag)
+      {
+      if ([NSWindow respondsToSelector:@selector(allowsAutomaticWindowTabbing)])
+            [NSWindow setAllowsAutomaticWindowTabbing: flag];
       }
--- a/mscore/macos/cocoabridge.mm
+++ b/mscore/macos/cocoabridge.mm
@@ -20,20 +20,36 @@
 #include "cocoabridge.h"
 #import <Cocoa/Cocoa.h>
 
-void  CocoaBridge::setAllowsAutomaticWindowTabbing(bool flag)
-{
-    if ([NSWindow respondsToSelector:@selector(allowsAutomaticWindowTabbing)])
-        [NSWindow setAllowsAutomaticWindowTabbing: flag];
-}
+id<NSObject> darkModeObserverToken;
 
-bool  CocoaBridge::isSystemDarkTheme()
-{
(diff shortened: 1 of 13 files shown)


Jojo-Schmitz:
See https://musescore.org/en/node/290946 which your PR seems to fix?

cbjeukendrup:
@Jojo-Schmitz I saw those, I will include them in the commit messages! (I thought PR title was enough)

Jojo-Schmitz:
And I see those being mentioned in the PR title only now, guess I need new glasses.

mscore/macos/cocoabridge.mm
```diff
@@ -20,14 +20,36 @@
...
+            observerToken = [[NSDistributedNotificationCenter defaultCenter]
+                             addObserverForName:@"AppleInterfaceThemeChangedNotification" object:nil
+                             queue:nil usingBlock:^(NSNotification * _Nonnull note) {   f();   }];
+      }
+
+CocoaBridge::~CocoaBridge()
+      {
+      if (@available(macOS 10.14, *))
+            [[NSDistributedNotificationCenter defaultCenter] removeObserver:observerToken];
+      }
```
mattmcclinch, line 37: This destructor is never called, because the `CocoaBridge` class is never instantiated. For that matter, we do not even need the constructor. This function should probably be renamed to something like `CocoaBridge::removeObservers()`, and called manually during the destruction of the `MuseScore` class. Sure, there is only one observer right now, but it is possible that we may want to add more in the future for other notifications. If and when that happens, it will not do to have an object called `observerToken`. We would need either a list of observer tokens or a more descriptive name for each one. I guess it depends on how much control we need over which observers are active at any given time.

cbjeukendrup: Oops! You're right. First, I referred to this StackOverflow post: https://stackoverflow.com/a/1008289/11432719. But then I changed my mind and decided not to use any instance of the CocoaBridge, but I forgot to do something with the destructor. Indeed, this way it doesn't make any sense. I will change this.

cbjeukendrup: Should be better now!

mscore/macos/cocoabridge.mm
```diff
@@ -20,14 +20,43 @@
...
-    if ([NSWindow respondsToSelector:@selector(allowsAutomaticWindowTabbing)])
-        [NSWindow setAllowsAutomaticWindowTabbing: flag];
-}
-
-bool  CocoaBridge::isSystemDarkTheme()
-{
-    NSString *osxMode = [[NSUserDefaults standardUserDefaults] stringForKey:@"AppleInterfaceStyle"];
-    return ([osxMode isEqualToString:@"Dark"]);
-}
+id<NSObject> darkModeObserverToken;
```
Marr11317, line 23: I've never seen such a syntax. Would you mind explaining what it means?

mattmcclinch: This is Objective-C code, which may explain why you are not familiar with the syntax. The purpose of the CocoaBridge class is to allow MuseScore to interact with the Cocoa subsystem on macOS, which is accomplished by using Objective-C. Since `cocoabridge.h` is valid C++ syntax as well as valid Objective-C syntax, the same header file can be #included by it’s own implementation file as well as any C++ sources that want to use the CocoaBridge class. If you have spent some time trying to learn Obective-C, and still have no idea what is going on, please let us know.

Marr11317: I have not spent time learning Objective-C:) And I had no idea it was possible to use non-C++ classes in a c++ project. That's actually kinda mindblowing. I'll have to dig on stackoverflow...

mattmcclinch: The macOS implementation of the Qt libraries is full of Objective-C code. This is no different. And StackOverflow is great, but it is not the first place I would turn to for a tutorial on Objective-C.

Marr11317: It wasn't about a tutorial on objective-C, but rather about combining c++ and objective-C in the same codebase. IMHO, SO is not so great apart for this kind of *wild* questions.

mattmcclinch: No need to search on StackOverflow. All you need to know is that Xcode can compile Objective-C code as well as C++ code and link them together.

cbjeukendrup: Yes, as @mattmcclinch said, it's Objective-C, which is the programming language Apple mainly uses for their frameworks. So CocoaBridge.mm is quite literally a bridge between our/Qt's C++ code and Apple's Objective-C code. 
I must admit that I don't have a very deep knowledge of Objective-C, but it is my most unfavourite language, primarily because of it's weird syntax with those square brackets. In C++, you would just write something like:
```cpp
object.method(val);
```
But with Objective-C objects, it would be:
```objc
[object method:val];
```
In my opinion, it looks bad, and it is inconvenient to type. 

If you're interested in native macOS/iOS developing, I recommend that you have a look at the Swift programming language. Apple has made this as a replacement for Objective-C, and I think it's a really nice and powerful language. 

Btw: This is the Objc-course I used: https://www.youtube.com/watch?v=5esQqZIJ83g but there are many others.

mscore/prefsdialog.cpp
```diff
@@ -559,6 +569,22 @@ PreferenceDialog::~PreferenceDialog()
...
+//---------------------------------------------------------
+
+void PreferenceDialog::retranslate()
+      {
+      retranslateUi(this);
+#ifdef Q_OS_MAC // On Mac, we have an extra theme option,
+      // namely, to follow the system's Dark Mode.
+      // Of course, we need to translate that too :)
+      if (CocoaBridge::isSystemDarkModeSupported())
+            styleName->setItemText(2, QCoreApplication::translate("PrefsDialogBase", "System", nullptr));
```
vpereverzev, line 583: the "disambiguation" parameter of QCoreApplication::translate method is optional, you can ignore it

cbjeukendrup: Fixed!

vpereverzev (approved):

vpereverzev:
@cbjeukendrup nice idea BTW, Martin appreciated it too

PR #6653, opened 2020-10-11
fix #311520: harmony playback preferences not honored until toggled

MarcSabatella:
Resolves: https://musescore.org/en/node/311520

The new preferences for disabling playback of chord symbols work fine,
only ater you've toggled them at least once, which forces them
to be written to the ".ini" file.
Otherwise, they are not being initialized correctly.
This is apparently from the use of QSettings to access the preferences
from code within the libmscore folder.
This technique is used at least one other place,
but it is apparently not reliable with respect to default values.

This commit adds new members of MScore class
so mscore can write them and libmscore can read them.
That is the approach used by most other preferences
that need to be accessed from libmscore.

In addition, there was a logic error
when creating a new score from a template.
There was code to handle new scores created from *old* templates,
so they wouldn't have chord symbols forced off
by the compatibility setting.
This code was erroneously being triggered for new templates as well,
making it impossible to disable chord symbol playback via a template.
That is fixed by performing the same version check used
when reading the template.

### DIFF
--- a/libmscore/mscore.cpp
+++ b/libmscore/mscore.cpp
@@ -96,6 +96,8 @@ QColor  MScore::dropColor;
 bool    MScore::warnPitchRange;
 int     MScore::pedalEventsMinTicks;
 
+bool    MScore::harmonyPlayDisableCompatibility;
+bool    MScore::harmonyPlayDisableNew;
 bool    MScore::playRepeats;
 bool    MScore::panPlayback;
 int     MScore::playbackSpeedIncrement;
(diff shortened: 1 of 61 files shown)


Jojo-Schmitz:
there's an mtest failure that does look relevant

MarcSabatella:
Not that I'm complaining, but I don't actually see that the mtests ran?  There is no travis line.  But anyhow, the code was good already, all I did was update the test references.  Only reason they passed previously is that the disable-playback-for-older-scores wasn't working correctly - the very problem this PR is fixing.

Jojo-Schmitz:
mtests are back, if you rebase ;-)

PR #6655, opened 2020-10-12
[MU4] Fixed build MacOS on CI

igorkorsukov:
I'll deal with error handling separately later.
Instead of my function (exit_if_err), we can `set -e` or `trap 'do_something' ERR` (https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script)
But now I can't just do it like that, because there is some code that returns an error as expected.
For example, in `package_mac`, an attempt is made to remove an old artifact, but it is never on the CI, so the expected error "file not found"

### DIFF
--- a/.github/workflows/ci_macos_mu4.yml
+++ b/.github/workflows/ci_macos_mu4.yml
@@ -68,7 +68,7 @@ jobs:
         bash ./build/ci/macos/checkcodestyle.sh  
     - name: Setup environment
       run: |
-        sudo bash ./build/ci/macos/setup.sh
+        bash ./build/ci/macos/setup.sh
     - name: Build          
       run: |
         T_ID=${{ secrets.TELEMETRY_TRACK_ID }}; if [ -z "$T_ID" ]; then T_ID="-"; fi
--- a/build/ci/linux/setup.sh
+++ b/build/ci/linux/setup.sh
@@ -130,8 +130,7 @@ cmake_dir="cmake/${cmake_version}"
 if [[ ! -d "${cmake_dir}" ]]; then
   mkdir -p "${cmake_dir}"
   cmake_url="https://cmake.org/files/v${cmake_version%.*}/cmake-${cmake_version}-Linux-x86_64.tar.gz"
-  wget -q --show-progress --no-check-certificate -O - "${cmake_url}" \
-    | tar --strip-components=1 -xz -C "${cmake_dir}"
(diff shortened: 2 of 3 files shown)


RomanPudashkin (approved):

Jojo-Schmitz:
This fixes the Mac builds in so far that it now reports failure if it fails, (which it still does) rather than failing silently.

BTW; the CI_LInux / build_m3 job's only remaining purpose (for the master branch) seems to be to run the mtest, right? So thoise would better get moved to the CI_Linux MU4 / build_m4 job? Or maybe combine with the CI_vtests job?

Jojo-Schmitz:
BTW, the `set -e` works on the last command of a pipeline, so `expected_error || true`  might fix that use case

Jojo-Schmitz:
Tested, it does:
```
#!/usr/bin/env bash
set -e
echo before
false || true
echo after
false
echo not here
```
Doesn't get to that last `echo`. But needs bash, a plain POSIX (k)sh doesn't seem to do it

igorkorsukov:
> BTW; the CI_LInux / build_m3 job's only remaining purpose (for the master branch) seems to be to run the mtests, right? So those would better get moved to the CI_Linux MU4 / build_m4 job I guess? Or maybe combine with the CI_vtests job? One job less to worry about and/or wait for.

Not just the launch of the mtest. This job also checks that the application is being built in MU3 mode.

Jojo-Schmitz:
> Not just the launch of the mtest. This job also checks that the application is being built in MU3 mode.

Which has been abandoned for the other platforms though?

igorkorsukov:
> BTW, the set -e works on the last command of a pipeline, so expected_error || true might fix that use case. 

Thank you for your solution. 
I want in this PR just fix MacOS build, and separately PR made error handling.

igorkorsukov:
> Which has been abandoned for the other platforms though?

At the moment, the following build configurations are left:
1. Linux MU4 - the main goal is to check that the application in MU4 mode is built under Linux and manual testing
2. Linux (MU3) - the main goal is to check that the application in MU3 mode is going to run under Linux and to perform the mtests
3. Windows MU4 - the main goal is to check that the application in MU4 mode is going to be built under Windows and manual testing
4. MacOS MU4 - the main goal is to check that the application in MU4 mode is going to be built under Windows and manual testing
5. vtests - performing visual tests

PR #6663, opened 2020-10-13
[MU4] Base of movements dialog

Eism:
Peek 2020-10-13 14-42


mu4/notation/qml/MuseScore/NotationScene/internal/NotationToolBarButton.qml
```diff
@@ -0,0 +1,89 @@
+import QtQuick 2.8
```
RomanPudashkin, line 1: Can we just improve FlatButton to add an ability add icon to right side of a text?

PR #6668, opened 2020-10-13
Fix assert's failed when clicking on a note

psmokotnin:
- Fix midi::Event assert's failed when clicking on a note

- fix playing a note on click

- fix CMake C++17 settings for macOS

### DIFF
--- a/mu4/notation/internal/notationplayback.cpp
+++ b/mu4/notation/internal/notationplayback.cpp
@@ -453,12 +453,26 @@ MidiData NotationPlayback::playNoteMidiData(const Ms::Note* note) const
     Ms::Instrument* instr = masterNote->part()->instrument(tick);
     channel_t channel = instr->channel(masterNote->subchannel())->channel();
 
+    Track t;
+    t.num = 0;
+    t.channels.push_back(channel);
+    midiData.tracks.push_back(t);
+
     Chunk chunk;
     chunk.beginTick = 0;
     chunk.endTick = Ms::MScore::defaultPlayDuration * 2;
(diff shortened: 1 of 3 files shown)


RobFog:
That should be the plural of "assert", so "asserts", right?

psmokotnin:
@RobFog I meant here this assert: https://github.com/musescore/MuseScore/blob/72ca0c374e7de0531ff65ced4f8f3aa66809421d/framework/midi/midievent.h#L252

RobFog:
Ah, just one assert then? I read the pull request title like there were multiple.

mu4/notation/internal/notationplayback.cpp
```diff
@@ -453,12 +453,26 @@ MidiData NotationPlayback::playNoteMidiData(const Ms::Note* note) const
...
     Chunk chunk;
     chunk.beginTick = 0;
     chunk.endTick = Ms::MScore::defaultPlayDuration * 2;
-    chunk.events.insert({ chunk.beginTick, Event(channel, EventType::ME_NOTEON, pitch, 80) });
-    chunk.events.insert({ Ms::MScore::defaultPlayDuration, Event(channel, EventType::ME_NOTEOFF, pitch, 0) });
-    chunk.events.insert({ chunk.endTick, Event(channel, EventType::ME_EOT, 0, 0) });
+    auto event = Event(Event::Opcode::NoteOn);
+    event.setChannel(channel);
+    event.setNote(pitch);
+    event.setVelocityFraction(0.63); //as 80 for 127 scale
```
RomanPudashkin, line 467: This code is in 3 places. I think better to place it in one function and reuse it

PR #6673, opened 2020-10-14
[MU3] restructure nightly builds

igorkorsukov:
restructure nightly builds

I changed ftp dirs structure, see https://ftp.osuosl.org/pub/musescore-nightlies/  

In the chat, they write that the default file sorting order is not convenient
I am researching how to solve this.


shoogle:
Just leave a note on the MuseScore.org downloads page (under nightlies) that tells people to scroll to the bottom of the OSUOSL page to find the latest nightly (or they can press the End key to go straight there, or click "Name" to reverse the sort order).

PR #6680, opened 2020-10-16
Fix #309322: Glissandos collide with double-digit start fretmarks

worldwideweary:
Resolves: https://musescore.org/en/node/309322 - Overlapping of double-digit tablature frets is fixed here, but this also recalibrates glissandos upon single-digit frets to be less "far out" from the number so that it is equivalent as with the fixed double-digit as presented here: https://github.com/musescore/MuseScore/pull/6680#issuecomment-710373369

**Updated**: [vtest/tablature-1.mscx] + [vtest/tablature-1-ref.png], but having vtests pass will require a moderator or whomever with the privilege of tagging "VTEST" to accept the reference .png file. Updated the .png file to be exported in 130DPI according to vtests/gen-ref specifications. 

Before: 
incorrectOffset
 

After:
glissOffsetCorrected

- [ x ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)


libmscore/glissando.cpp
```diff
@@ -323,6 +323,10 @@ void Glissando::layout()
 
       // initial note dots / ledger line / notehead
       offs1 *= -1.0;          // discount changes already applied
+
+      // Accommodate for TAB numerals for x1 position
+      offs1.rx() += (cr1->staff()->isTabStaff(cr1->tick())) ? anchor1->tabHeadWidth() * 0.50 : 0;
```
worldwideweary, line 328: I could check here for making sure the fret numnber is above 9 first like
if (cr1->staff()->isTabStaff(cr1->tick()))
            offs1.rx() += anchor1->fret() > 9 ? anchor1->tabHeadWidth() * 0.50 : 0;

and I think all would be okay, but upon further looking into the default as it is already for 0-9, the offset looks like it could be a little closer to the fret number like with the noteheads... so I think this should be fixed up at the same time rather than merely offsetting for a double-digit fret number only.

Jojo-Schmitz:
One of the vtests isn't happy (i.e. shows a difference), before:
tablature-1-ref
after:
tablature-1-1
diff:
tablature-1-diff
but that seems to be expected here, so could someone tag this PR as "vtests"?

worldwideweary:
At the moment it is to be expected. I've updated the code (will update this PR shortly) so that a single-digit is better than how it currently is in conjunction with the overlap fix of double-digits. Let me demonstrate: 

Old version (3.5) gives this for single digit: 
Old

The extra-space in the gap isn't too noticeable until you compare it with this, which shows [single and double digits] as equivalent in gap and also having no overlap of 10 fret:
SingleDigit

These changes will also trigger a VTEST fail because it changes how a single digit gliss looks on tab. Hopefully will soon update that also if it's figured out.

vtest/tablature-1.mscx
```diff
@@ -1,28 +1,75 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<museScore version="2.07">
```
worldwideweary, line 2: These changes are the result of running mscore with "-t" and allowing "Automatic Placement" updates (moving from version 2) with the addition of the code update for glissandos as in previous commits

Jojo-Schmitz:
Looks much better and in more than one way, before:
tablature-1-ref
After:
tablature-1-1
The diff though looks quite horri
tablature-1-diff
ble:

worldwideweary:
@Jojo-Schmitz Yeah it did the "auto-placement" update since it was previously generated in version 2.x. But I don't know what to do with the reference file. I updated it on a commit here (it came out much bigger than the original since I didn't change DPI at at all on my dev build... 360DPI seems bigger than what it wants), but I don't know what i can do else, and it's still using some older test file instead of my replacement. Any ideas of what I can do, or is it something someone with higher privilege must perform?

Jojo-Schmitz:
I don't know, sorry

worldwideweary:
> I don't know, sorry
Hopefully someone in the dev-chat can help clarify what needs to be done. Just don't want the PR to go dormant so that especially the double-digit overlap of tab gliss. can be fixed up and implemented into a near future build of MS.

vpereverzev:
> @Jojo-Schmitz Yeah it did the "auto-placement" update since it was previously generated in version 2.x, so from that there's additional better layout. But I don't know what to do with the reference file. I updated it on a commit here (it came out much bigger than the original since I didn't change DPI at at all on my dev build... 360DPI seems bigger than the original), but I don't know what i can do else, and it's still using some older test file instead of my replacement [_tablature-1-ref.png_]. Any ideas of what I can do, or is it something someone with higher privilege must perform? In which case they'd need to make sure the resolution was okay.

vtests job is quite strange, but we had no chance to rework it yet. This job is building MU3 twice (1. your branch, 2. reference branch - 3.x in your case). **So, reference images come from the current 3.x branch**

I'll just add **vtests** label here, so this job will ignore that difference.

vpereverzev:
looking for rebase BTW

Jojo-Schmitz:
a) this PR needs  rebase
b) the vtest label doesn't seem to work any longer

worldwideweary:
> Seems the vtest label doesn't seem to work any longer

Out of curiosity, how would a rebase help if it just says "conflicting file" with the .png that's supposed to replace the current 3.x branch file?

vpereverzev:
> Seems the vtest label doesn't seem to work any longer

it works, but should be set before the run of this job

vpereverzev:
> > Seems the vtest label doesn't seem to work any longer
> 
> Out of curiosity, how would a rebase help if it just says "conflicting file" with the .png that's supposed to replace the current 3.x branch file?

Looks like someone changed it recently

worldwideweary:
> Looks like someone changed it recently
Thanks: just rebased

On the side, it was a curiousity having to use "git checkout --theirs -- vtest/tablature-1-ref.png" to select the changes while rebasing since the .png file certainly can't "merge"

Jojo-Schmitz:
> > Seems the vtest label doesn't seem to work any longer
> 
> it works, but should be set before the run of this job

Setting that label used to trigger a rebuild (just to ignore its result and mark it green)

AntonioBL:
> 
> vtests job is quite strange, but we had no chance to rework it yet. This job is building MU3 twice (1. your branch, 2. reference branch - 3.x in your case). **So, reference images come from the current 3.x branch**

That's exactly what I had in mind when writing the CI vtest script. In this way the images are always up-to-date and the vtests shows *only* the differences introduced by the current PR. This would in principle save from having to update the ref images (and thus increasing the repository size), since the ref images are generated "on-the-fly", and also helps in confronting the current images with the ref images, since small changes can be caused by different PC configuration (it ensures that all images are generated on the same machine, with the same configurations).

In order to automatically start when a label is applied, the `labeled` case should be added to the `types:` trigger in `pull_request:` field. This is for example how the old code read:
```
  pull_request:
    branches:
    - master
    types: [opened, synchronize, reopened, labeled, unlabeled]
```
See also: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

PR #6683, opened 2020-10-16
[MU4] Language fix

Eism:
- Fixed default language initialization. From this https://musescore.org/en/node/311755
  In the configurator, if the default language is not selected, then select the system language. But unfortunately, the request to the configurator to get the default language was earlier than setting the default language.
  First set the default language(m_languagesConfiguration->init()), then apply the language(m_languagesController->init()).
- Fixed network requests in controllers
  The network requests must be run in a different thread than the main thread, GUI thread


igorkorsukov (approved):

RomanPudashkin (approved):

Jojo-Schmitz:
Those to merge commits really should not be there

PR #6687, opened 2020-10-16
[MU4] Added print free space for Linux CI

igorkorsukov:
Not building AppImage 
An error appears that there is no left space.
I'm going to investigation

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -56,6 +56,11 @@ jobs:
         sudo bash ./build/ci/linux/build.sh -n ${{ github.run_id }} --telemetry $T_ID
     - name: Run mtests 
       run: |
+        # At moment build and run mtests consume 13.4 Gb of the disk.
+        # After left too little free space.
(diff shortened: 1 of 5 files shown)


AntonioBL:
I think they reduced the size of the disk of the virtual machine.
You could add in the workflow steps:
(sudo) docker system prune -a -f
rm -rf /usr/local/lib/android

since no docker images are used and we can get rid of the cached ones, and we are not using Android compilers and SDK. Maybe we could delete or uninstall other packages.

igorkorsukov:
I see what:
* At the start (after checkout), we have available - 20572220 (19.6 Gb)
* After setup (install dependencies included Qt) - 19860900 (18.8 Gb) - 0.8 Gb
* After build source - 14552508 (13.8 Gb) - 5 Gb
* After build and run mtests - 402376 ( 393 Mb) - 13.4 Gb
* 393 Mb - Too little space to build AppImage

igorkorsukov:
> (sudo) docker system prune -a -f
> rm -rf /usr/local/lib/android

Thank you for your advice.

PR #6690, opened 2020-10-17
[MU4] fixed clicking on notation on HIDPI screens

RomanPudashkin:


mu4/notation/view/notationpaintview.cpp
```diff
@@ -360,19 +346,22 @@ void NotationPaintView::dropEvent(QDropEvent* ev)
...
+    QPoint scaledPoint(point.x() * scale, point.y() * scale);
+
+    return m_matrix.inverted().map(scaledPoint);
 }
 
-QRect NotationPaintView::toLogical(const QRect& r) const
+double NotationPaintView::guiScale() const
 {
-    return m_matrix.inverted().mapRect(r);
+    return window()->screen()->devicePixelRatio();
```
igorkorsukov, line 359: we have `UiConfiguration::guiScaling`, I think better use it (and move this code there)

Jojo-Schmitz:
rebase needed

mu4/notation/view/notationpaintview.cpp
RomanPudashkin: This is a problem. Here need to get the scale of the current screen (on which a view is placed). I could not find any solution without using window() of QQuickPaintedItem

igorkorsukov: there is `IMainWindow` interface, which can be used in the UiConfiguration by injection

RomanPudashkin: fixed

igorkorsukov (approved):

PR #6693, opened 2020-10-18
[MU3] Fix #311792: Override alter tag with note tuning value if present

infojunkie:
Resolves: https://musescore.org/en/node/311792

This patch modifies function `writePitch()` in exportxml.cpp to override the ["alter" MusicXML element](https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-alter.htm) with the value found in the `note->tuning()` if present. This allows the manual note tuning to be preserved during MusicXML export.

A test `TestMxmlIO::noteTuning()` is added to verify the desired behaviour.


Jojo-Schmitz:
You ticked the 2nd to last box of that Pp comments template bit in the commit forgot the #

Jojo-Schmitz:
@lvinken mind to check?

infojunkie:
> You ticked the 2nd to last box of that PR comment template, but in the commit title forgot the # ;-)

Yes terribly sorry! I read the guidelines after pushing the commit... Do you have a suggestion to fix that?

Jojo-Schmitz:
Quite simple: `git commit --amend`, add the #, `git push -f`

infojunkie:
Thanks. I used the recipe at https://www.educative.io/edpresso/how-to-change-a-git-commit-message-after-a-push, because it was the second-to-last commit that needed a change. Out of curiosity, do you use the squash option during PR merge?

Jojo-Schmitz:
No, usually not, we'd rather have the submitter to do it ;-)

lvinken:
First remarks:
I'm glad to see a test case, will have to check if it coverage is sufficient.
- please check if the code generates correct output for:
  - a G# without tuning
  - a G# with a small tuning (e.g. 10)
  - a G with a large tuning (e.g. 110)
The last two, I believe, should produce the same alter value.
Also, comparing a float with 0 is not a good idea due to inaccuracies in the floating point representation. Typically a small epsilon value is used instead. Similarly using a float as a boolean is not OK.
More to follow later, have to go to work now.

importexport/musicxml/exportxml.cpp
```diff
@@ -3211,6 +3211,11 @@ static void writePitch(XmlWriter& xml, const Note* const note, const bool useDru
                   default:                                             break;
                   }
             }
+      // Override accidental with explicit note tuning
+      auto tun = note->tuning();
+      if (fabs(tun) > 0.0) {
```
Jojo-Schmitz: As MuseScore allows tuning on cents only, and as comparing `float` to `0.0` and using it as a 'bool` isn't a good idea anyway, maybe better use `if (fabs(tun) > 0.01) {`, similar to how it is used in https://github.com/musescore/MuseScore/blob/090b8ca0bd9342c8e0b7fdd4b340018af9da1f5a/mscore/musescore.cpp#L6114

infojunkie: Fixed the comparison

Jojo-Schmitz:
Now please squash those 3 commits into one. `git rebase -i HEAD~3` (last 3 commits), replace the 2nd and 3rd `pick` with `fixup`, save, `git push -f`

infojunkie:
> please check if the code generates correct output for:
> a G# without tuning
> a G# with a small tuning (e.g. 10)
> a G with a large tuning (e.g. 110)
> The last two, I believe, should produce the same alter value.

Thanks for the useful test case. I have a few questions and comments:

- My understanding is that the Musescore note "tuning" value, as well as the MusicXML "alter" value, are not meant to take enharmonicity in consideration. This means that G# with a small tuning _t_ vs G with the same equivalent tuning (100+_t_) will NOT result in a same "tuning" value. I verified this by creating a sheet with your suggested case, and examined the [resulting Musescore file XML](https://github.com/musescore/MuseScore/files/5417731/testNoteTuning.zip). The explicitly given tunings are preserved. Is my understanding wrong?

- Currently exporting the file above to MusicXML seems to miss the G# + 10c tuning. Will fix this as part of my PR.

infojunkie:
> Now please squash those 3 commits into one. `git rebase -i HEAD~3` (last 3 commits), replace the 2nd and 3rd `pick` with `fixup`, save, `git push -f`

Thanks, I'll do that once I'm done with all changes here if you don't mind. Just a note, though: GitHub gives the option to "squash and merge" automatically, without using `git rebase` and `git push -f`. Happy to comply with your process, though.

Jojo-Schmitz:
Sure, take your time. As said, to the best of my knowledge the folks doing the merges do not squash.

infojunkie:
One more comment to @lvinken : there's a TODO in the `writePitch()` saying:
> // TODO what if both alter and alter2 are present? For Example: playing with transposing instruments

In my opinion, `alter + alter2` would make sense, since `alter` seems to deal with transposing instruments, while `alter2` deals with accidentals and note tunings. 

If you agree, can I include this fix too in this PR?

lvinken:
On this one I cannot immediately provide a full answer to and I do not have the time available to work out the details. In retrospect, I am not sure transposition is the main issue here. The current implementation uses a special case for a few microtonal accidentals, ignores other and also ignores the tuning values. This implies the output may be incorrect in several corner cases.

All I can add at this time is: please try to make sure whatever the MuseScore GUI can do(in terms of creating internal data structures and visualisation on screen), leads to the expected MusicXML output.

infojunkie:
> All I can add at this time is: please try to make sure whatever the MuseScore GUI can do(in terms of creating internal data structures and visualisation on screen), leads to the expected MusicXML output.

OK, thanks. I'll do my best to enumerate the possible combinations and cases.

Jojo-Schmitz:
Something severly went wrong with this, I guess you used `git merge` rather than `git pull --rebase`

infojunkie:
> Something severely went wrong with this, I guess you used `git merge` rather than `git pull --rebase`

Thanks, I think I fixed.

infojunkie:
I added code + updated tests to ensure that the 2 `alter` values can be combined to produce expected results. Next, I will add more comprehensive tests before I submit this PR for review.

importexport/musicxml/exportxml.cpp
```diff
@@ -3234,11 +3234,12 @@ static void writePitch(XmlWriter& xml, const Note* const note, const bool useDru
...
-            xml.tag("alter", alter);
-      if (!alter && alter2)
-            xml.tag("alter", alter2);
-      // TODO what if both alter and alter2 are present? For Example: playing with transposing instruments
+      // `alter` represents the "regular" (Western) pitch which can be 
+      // 0 (natural), 1 (sharp), -1 (flap) or some other integer depending on transposing instruments.
+      // `alter2` represents a microtone or manually-adjusted note tuning.
+      // In MusicXML, These two values are merged in the same "alter" tag.
+      if (alter || alter2)
+            xml.tag("alter", (double)alter + alter2);
```
Jojo-Schmitz, line 3242: Not `double(alter+alter2)`, i.e. casting the result of the addition rather than just one of the 2 and in a more C++ way?

infojunkie: Not sure, one of them is int and the other double.

Jojo-Schmitz: OK, I see, but the result is a double in any case.

importexport/musicxml/exportxml.cpp
```diff
@@ -3239,7 +3239,7 @@ static void writePitch(XmlWriter& xml, const Note* const note, const bool useDru
       // `alter2` represents a microtone or manually-adjusted note tuning.
       // In MusicXML, These two values are merged in the same "alter" tag.
       if (alter || alter2)
-            xml.tag("alter", (double)alter + alter2);
+            xml.tag("alter", double(alter + alter2));
```
Jojo-Schmitz, line 3242: What happens if you drop that cast? If that doesn't result in a compiler warning, just drop it.
The default result of any arithmetic involving a `double` and an `int` is to convert the result into a `double` anyhow.
See also https://docs.microsoft.com/en-us/cpp/cpp/standard-conversions?view=msvc-160#arithmetic-conversions

infojunkie: Thanks, dropping the cast also seems to pass the tests.

infojunkie:
I've added more comprehensive tests to cover the different cases in the code, as per the attached screenshot.
 
testNoteTuning

These include manual tunings over regular and extended accidentals, as well as ensuring that the hard-coded accidentals in the export function are correctly handled.

infojunkie:
I've squashed my commits and rebased to the latest 3.x. Waiting for tests then it's ready from my side.

PR #6701, opened 2020-10-19
[MU4] Added run utests on CI

igorkorsukov:
Added run utests (google tests) on CI


.github/workflows/ci_windows.yml
```diff
@@ -5,8 +5,8 @@ on:
   # pull_request:
   #   branches:
   #   - master
-  schedule:               # Schedule 3.x nightly build! (not master MU3)  
-    - cron: '0 4 */1 * *' # At 04:00 on every day-of-month
+  schedule:                 # Schedule 3.x nightly build! (not master MU3)  
+    - cron: '0 4 */1 */1 *' # At 04:00 on every day-of-month
```
Jojo-Schmitz, line 9: IMHO those `/1` are redundant. Unless cron on GitHub works different from all other cron implementations

igorkorsukov: Thanks.
I will know, but let it be.

PR #5934, opened 2020-04-13
[MU3] Update Bravura to 1.39

Jojo-Schmitz:
As taken fresh from https://github.com/steinbergmedia/bravura
Follow up to #4879, which updated it to 1.32

PR #4879, opened 2019-04-07
Update Bravura and BravuraText to latest Steinberg version 1.32

Jojo-Schmitz:
Plus the newly added LICENSE.txt and README.md.
As taken from https://github.com/steinbergmedia/bravura/

PR #6704, opened 2020-10-20
[MU4] Update Bravura to 1.39

Jojo-Schmitz:
As taken fresh from https://github.com/steinbergmedia/bravura
Follow up to #4879, which updated it to 1.32

Counterpart to #5934 (which is in 3,x meanwhile)


Jojo-Schmitz:
failing vtests expected

vpereverzev:
@Jojo-Schmitz I'm not against of this particular update, but we probably need to update the vtests ref-images in this PR?

vpereverzev (approved):

Jojo-Schmitz:
Maybe. But there's a lot of those that need updating and not just for Bravura as far as I can tell? And many/most of the scores used for that are still from 2.x

Jojo-Schmitz:
IMHO updated vtests is a whole project of its own

Jojo-Schmitz:
I've tried, but can't get MuseScore 4 to run in converter mode (on Windows) :-(

vpereverzev:
@Jojo-Schmitz Oh, I think it's much easier. In CI_vtests job details you can find an artifact "compare" - https://github.com/musescore/MuseScore/suites/1367961822/artifacts/22385020. It contains the "new" images and the existing "refs".

Jojo-Schmitz:
So just grabbing them from there might be the solution?

vpereverzev:
Yep, just grab them and upload the new ones here

Jojo-Schmitz:
That would cover only those this PR changes, but not the ones changed by other already merges ones, like from #6705.
Unless I'll collect those artifacts too (which in this case II could, but might miss others), should I?

vpereverzev:
No, I think all you need here is to rebase your branch on the latest master (if you didn't it before). And result should cover all the difference considering the previous PRs.

Jojo-Schmitz:
Ah! Of course...

Jojo-Schmitz:
Well, no, it does not. So I'll add the ledgerlines ones too.

PR #6359, opened 2020-07-21
[MU3] Allow ledgerlines exactly as wide as the noteheads

Jojo-Schmitz:
following a discussion in the MuseScore developers' chat on Telegram, see https://t.me/musescoreeditorchat/81326 ff.

Before this change the ledgerlines stick out the notedheads by 1/2 the line thickness either side, if their length is set to 0:
image
due to `QPen()` using `Qt::SquareCap` by default, which adds that half a line width either side.

Issue applies to 3.x (maybe 3.6, which is planned to improve layout?) and master

PR #6705, opened 2020-10-20
[MU4] Allow ledgerlines exactly as wide as the noteheads

Jojo-Schmitz:
and fix the vtest bat files to use MuseScore4

Counterpart to #6359

### DIFF
--- a/vtest/gen-ref.bat
+++ b/vtest/gen-ref.bat
@@ -1,7 +1,7 @@
 
 rem create reference
 
-set MSCORE=..\msvc.install_x64\bin\musescore.exe
+set MSCORE=..\msvc.install_x64\bin\musescore4.exe
 set DPI=130
 
 %MSCORE% %1.mscz -r %DPI% -o %1.png
--- a/libmscore/ledgerline.cpp
+++ b/libmscore/ledgerline.cpp
@@ -67,9 +67,9 @@ void LedgerLine::layout()
     }
     qreal w2 = _width * .5;
     if (vertical) {
-        bbox().setRect(-w2, -w2, _width, _len + _width);
+        bbox().setRect(-w2, 0, w2, _len);
     } else {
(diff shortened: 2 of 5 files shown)


Jojo-Schmitz:
vtest failures expected

vtest/gen-ref.bat
```diff
@@ -1,7 +1,7 @@
 
 rem create reference
 
-set MSCORE=..\msvc.install_x64\bin\musescore.exe
```
vpereverzev, line 4: @igorkorsukov take a look on that

Jojo-Schmitz: oops

Jojo-Schmitz: Was wrong before, and still ids ;-)

Jojo-Schmitz: It is a much simpler script and not meant to be run automatically, and so %INSTALL_PATH% is not set, so I guess we can let this pass for now?

igorkorsukov: I think yes, we can pass this 
This is files not used on CI, as I understand, they for local testing on Windows?

vpereverzev (approved):

vtest/gen-ref.bat
Jojo-Schmitz: Yes

vtest/gen-ref.bat
```diff
@@ -1,7 +1,7 @@
...
+set MSCORE=..\msvc.install_x64\bin\musescore4.exe
 set DPI=130
 
 %MSCORE% %1.mscz -r %DPI% -o %1.png
```
Jojo-Schmitz, line 7: Also here it needs to be `%MSCORE% %1.mscx -r %DPI% -o %1.png`, mscx vs. mscz, as I just found out

PR #6263, opened 2020-06-26
[MusicXML] Add missing Gould arrow quartertone accidentals

rettinghaus:
This PR is a simple addition to allow import/export of the added Gould arrow quartertone accidentals in MusicXML 3.1.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #6182, opened 2020-06-08
[MU3] Add more accidentals

Jojo-Schmitz:
see https://musescore.org/en/node/306457 and https://musescore.org/en/node/306456

* add missing Gould arrow down/up accidentals
* add more extended Helmholtz-Ellis accidentals
* add the most important (Spartan) Sagittal accidentals
* add curly braces to accidentals
* (almost) complete MusicXML im- and export

See also #6002 (Add support for triple sharp and triple flat, pending), #5737 (adding Wyschnegradsky accidentals, merged) and #6263 ([MusicXML] Add missing Gould arrow quartertone accidentals, merged)

PR #6002, opened 2020-04-28
Add support for triple sharp and triple flat

mattmcclinch:
SMuFL provides glyphs for triple sharp and triple flat, so they might as well be available to use in MuseScore. This involves adding 14 new TPCs and expanding the range of the AccidentalVal enum class.

This was requested on the forum topic [Triple sharps and triple flats](https://musescore.org/en/node/304503).

See https://seanyeh.com/pages/alkans_triple_sharp/ for an example of a triple sharp being used in a piece of music from the Romantic period.

PR #5737, opened 2020-02-20
fix #301387: add Wyschnegradsky accidentals

Jojo-Schmitz:
Fixes https://musescore.org/en/node/301387, discussed in https://musescore.org/en/node/301256

PR #6708, opened 2020-10-20
[MU4] Add more accidentals

Jojo-Schmitz:
see https://musescore.org/en/node/306457 and https://musescore.org/en/node/306456

* add missing Gould arrow down/up accidentals
* add more extended Helmholtz-Ellis accidentals
* add the most important (Spartan) Sagittal accidentals
* add curly braces to accidentals
* (almost) complete MusicXML im- and export

See also #6002 (Add support for triple sharp and triple flat, pending), #5737 (adding Wyschnegradsky accidentals, merged) and #6263 ([MusicXML] Add missing Gould arrow quartertone accidentals, merged)

Counterpart to #6182, which is in 3.x now.


Jojo-Schmitz:
Boy, what a royal pain to 'cherry-pick' this from 3.x for master :-(

libmscore/sym.cpp
```diff
@@ -2682,10 +2682,12 @@ const std::array<const char*, int(SymId::lastSym) + 1> Sym::symUserNames = { {
     QT_TRANSLATE_NOOP("symUserNames", "No symbol"),
     "4-string tab clef",
     "6-string tab clef",
-    "11 large diesis down, 3\u00b0 down [46 EDO]",
-    "11 large diesis up, (11L), (sharp less 11M), 3\u00b0 up [46 EDO]",
-    "11 medium diesis down, 1\u00b0[17 31] 2\u00b046 down, 1/4-tone down",
-    "11 medium diesis up, (11M), 1\u00b0[17 31] 2\u00b046 up, 1/4-tone up",
+    QT_TRANSLATE_NOOP("symUserNames", "11 large diesis down, 3\u00b0 down [46 EDO]"),
+    QT_TRANSLATE_NOOP("symUserNames", "11 large diesis up, (11L), (sharp less 11M), 3\u00b0 up [46 EDO]"),
+    //QT_TRANSLATE_NOOP("symUserNames", "11 medium diesis down, 1\u00b0[17 31] 2\u00b046 down, 1/4-tone down"),
```
vpereverzev, line 2687: Do we still need this?

Jojo-Schmitz: That is the original string as taken from SMuFL, but as such doesn't work with the translations system. So I'd rather keep this (and commented out), as a hint should this string ever change

Jojo-Schmitz: Hmm, maybe not...
We'll see once translations got enabled. I do remember there are problems with that `\u` notation and the translations system, if it is not really clear where the number stops and the next 'normal' letter shows

vpereverzev (approved):

PR #6714, opened 2020-10-20
[MU4] Staff/Part properties

Eism:
Moved staff properties to MU4.
Implemented setup the theme for widgets.

Peek 2020-10-19 21-22


Marr11317:
also, is it going to be rewritten in qml?

mu4/notation/view/widgets/editpitch.cpp
```diff
@@ -0,0 +1,70 @@
+//=============================================================================
+//  MusE Score
```
Marr11317, line 2: This is still the old copyright. I don't know if it will endup being changed. There's many more places though, and its not related to this PR specifically.

Eism:
> What is that color dialog? Is it some kind of platform default?

Yes, it is the default qt dialog for select colors: https://doc.qt.io/qt-5/qcolordialog.html

Eism:
> also, is it going to be rewritten in qml?

Yes, in the future

mu4/notation/view/widgets/editpitch.cpp
Eism: Thanks, fixed

mu4/notation/inotationinteraction.h
```diff
@@ -52,6 +52,7 @@ class INotationInteraction
 
     // Select
     virtual Element* hitElement(const QPointF& pos, float width) const = 0;
+    virtual int hitStaffIndex(const QPointF& pos) const = 0;
```
vpereverzev, line 55: I don't like to have this method in the interface. We've a pretty abstract method "hitElement" which should provide you a proper element which can provide you all necessary information. Otherwise, there will be A LOT of similar methods with different types of elements. I guess your "hitStaff" should be a private one in notationinteraction

mu4/notation/internal/notationinteraction.cpp
```diff
@@ -388,11 +388,16 @@ Element* NotationInteraction::hitElement(const QPointF& pos, float width) const
 {
     QList<Ms::Element*> ll = hitElements(pos, width);
     if (ll.isEmpty()) {
-        return nullptr;
+        return hitMeasure(pos).measure;
```
vpereverzev, line 391: why it's not a part of "hitElements" method?

mu4/notation/internal/notationinteraction.h
```diff
@@ -135,6 +136,14 @@ class NotationInteraction : public INotationInteraction
     void cmdAddSlur(const Ms::Slur* slurTemplate = nullptr);
     void addSlur(Ms::ChordRest* cr1, Ms::ChordRest* cr2, const Ms::Slur* slurTemplate);
 
+    struct HitMeasureData
```
vpereverzev, line 139: Let's avoid this struct, it may force others to create another structs like "HitSegmentData", "HitSystemData", etc.
All you need here is to add "staffIndex" field to Ms::Measure (maybe it already has one)

mu4/notation/view/resources/data/std_sample.mscx
```diff
@@ -0,0 +1,192 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vpereverzev, line 1: Where do we use this file?

vpereverzev: What is the purpose?

framework/ui/view/theme.cpp
```diff
@@ -243,3 +247,22 @@ void Theme::initMusicalFont()
         update();
     });
 }
+
+void Theme::setupWidgetTheme()
```
RomanPudashkin, line 251: I think this function needs to be in some other class. Theme must simply teel to us which of color/fonts used in the current moment. It must not configuring all system

mu4/notation/inotationselection.h
```diff
@@ -36,6 +36,8 @@ class INotationSelection
     virtual Element* element() const = 0;
     virtual std::vector<Element*> elements() const = 0;
 
+    virtual int firstStaffIndex() const = 0;
```
RomanPudashkin, line 39: Better not add that method to this interface. This is an abstract class for a selection. it shouldn't know about staves

mu4/notation/internal/notationactioncontroller.cpp
```diff
@@ -217,3 +218,14 @@ void NotationActionController::openPageStyle()
...
+
+void NotationActionController::openStaffProperties()
+{
+    auto interaction = currentNotationInteraction();
+    if (!interaction) {
+        return;
+    }
+
+    int staffIdx = interaction->selection()->firstStaffIndex();
+    interactive()->open("musescore://notation/staffproperties?sync=false&staffIdx=" + QString::number(staffIdx).toStdString());
```
RomanPudashkin, line 230: sync=false

Is this not-modal dialog?

mu4/notation/internal/notationparts.cpp
```diff
@@ -199,6 +199,36 @@ void NotationParts::setPartName(const ID& partId, const QString& name)
     m_partsChanged.notify();
 }
 
+void NotationParts::setPartSharpFlat(const ID& partId, const Ms::PreferSharpFlat& sharpFlat)
```
RomanPudashkin, line 202: PreferSharpFlat

I don't like this name. Better simply SharpFlat

mu4/notation/view/resources/data/std_sample.mscx
Eism: This file(and tab_sample.mscx) is needed to view changes in notation in real-time on test data, not on real data.
A new score is generated from this file and loaded to ExampleView.

mu4/notation/internal/notationactioncontroller.cpp
Eism: to close the context menu before opening the dialog

mu4/notation/inotationselection.h
Eism: But ... it's not that simple. In this case, staffIndex is the index of staff inside System. And libmscore expects this index to work correctly. The selection can be started from the first staff or from any other.

mu4/notation/internal/notationparts.cpp
Eism: fixed

mu4/notation/internal/notationactioncontroller.cpp
Eism: fixed

mu4/notation/internal/notationinteraction.cpp
Eism: fixed

mu4/notation/inotationinteraction.h
Eism: see https://github.com/musescore/MuseScore/pull/6714#discussion_r512454129

mu4/notation/inotationselection.h
RomanPudashkin: please rename this function to something like "startStaffIndex"

mu4/notation/notationtypes.h
```diff
@@ -129,6 +132,28 @@ struct ViewModeOption {
     QString actionString;
     ViewMode viewMode;
 };
+
+struct StaffUpdateOptions {
```
RomanPudashkin, line 136: Better rename to StaffConfig

RomanPudashkin (approved):

mu4/notation/internal/notationinteraction.h
Eism: decided to leave it as it is

framework/ui/view/theme.cpp
Eism: decided to leave it as it is

PR #6723, opened 2020-10-22
[MU3] Added tx2s3 workflow

igorkorsukov:
added tx2s3 workflow


Jojo-Schmitz:
Shouldn't this workflow rather get added to the [tx2s3 repository](https://github.com/musescore/tx2s3) rather then the MuseScore one?
(Similar question for the [handbook2PDF repository](https://github.com/musescore/handbook2PDF), resp. its daily cron job)

igorkorsukov:
> Shouldn't this workflow rather get added to the [tx2s3 repository](https://github.com/musescore/tx2s3) rather then the MuseScore one?

I am going to move this script to the main repository (MuseScore) and delete tx2s3 repository.
Because this script tightly bound with code MuseScore (translation structure, lang list, formats, version and etc)
Moreover, I think we will have different scripts for 3.x and master.
So, better put this script in the main repo and corresponding branches.

> Similar question for the handbook2PDF repository, resp. its daily cron job

It not related MuseScore code, so it will be in the separate repo, as is.

PR #6728, opened 2020-10-22
Re-enable workspace and palette mtests

AntonioBL:
This PR splits the mtest job from the Linux build job, and re-enables the "workspace" and "palette" mtests, as well as the use of AddressSanitizer in the tests.

- [N/A] I made sure the commit message title starts with "fix #424242:" if there is a related issue


.github/workflows/ci_linux.yml
```diff
@@ -70,3 +67,35 @@ jobs:
...
+        DO_PUBLISH='false'
+        echo "DO_PUBLISH=$DO_PUBLISH" >> $GITHUB_ENV
+        echo "DO_PUBLISH: $DO_PUBLISH"
+    - name: Setup environment
+      run: |
+        sudo bash ./build/ci/linux/setup.sh
+    - name: Build
+      run: |
+        T_ID=${{ secrets.TELEMETRY_TRACK_ID }}; if [ -z "$T_ID" ]; then T_ID="''"; fi
+        sudo bash ./build/ci/linux/build.sh -n "${{ github.run_id }}" --telemetry "$T_ID" --buildtype mtests
```
igorkorsukov, line 98: I think better add the option "build_mode" to `build.sh` or set "build_mode" by `buildtype` and remove "Configure workflow" step

AntonioBL: In principle build_mode just translates to "dev", "testing", "release", if I understood correctly. Each of them can be compiled as a debug build for tests. So in principle the "type" of build and the "mode" of build are two different things.
Or maybe I misunderstood what you are saying :sweat_smile:

AntonioBL:
I think an "install" step was missing in the first submission. Now (hopefully) it should work.

.github/workflows/ci_linux.yml
igorkorsukov: Correct, build mode - this is just the set config ("dev", "testing", "release"). For tests needs only "dev"

I think better to avoid "Configure workflow" to make it easy.

I see variants: 
1. Add --build_mode option to build.sh script and set it explicit
2. Set BUILD_MODE inside build.sh script used passed buildtype

And a new variant:
remove option `--buildtype`
add option `--build_mode`
set `--build_mode mtest`
and use `BUILD_MODE == mtest` for configure build 

that is, in fact, you can expand the values ​​of BUILD_MODE :)

AntonioBL: Ah, ok. :+1: 
I will also amend the "install" step (now it is failing because the environment is not set outside of the build script).

AntonioBL:
At the moment, the tests run without problem, but "workspace" mtest fails.
By looking at the log, it says:
```
FAIL!  : Ms::TestWorkspaces::testEditPalette(Basic) Compared values are not the same
Actual   (WorkspacesManager::currentWorkspace()->translatableName()): "Basic"
Expected (editedWorkspaceName)                                      : "Basic edited"
```
and similarly for other tests. Is it a problem of writing permissions?

igorkorsukov:
> At the moment, the tests run without problem, but "workspace" mtest fails.
> By looking at the log, it says:
> 
> ```
> FAIL!  : Ms::TestWorkspaces::testEditPalette(Basic) Compared values are not the same
> Actual   (WorkspacesManager::currentWorkspace()->translatableName()): "Basic"
> Expected (editedWorkspaceName)                                      : "Basic edited"
> ```
> 
> and similarly for other tests. Is it a problem of writing permissions?

@Eism Maybe you changed behavior?

PR #6733, opened 2020-10-23

Jojo-Schmitz:
and correct a small error in a bat file


mu4/importexport/internal/musicxml/importmxmlpass1.cpp
```diff
@@ -605,7 +605,7 @@ static Tid tidForCreditWords(const CreditWords* const word, std::vector<const Cr
         return Tid::TITLE;                      // no better title candidate found
     }
 
-    return Tid::DEFAULT;                        // not reached
+    //return Tid::DEFAULT;                        // not reached
```
vpereverzev, line 608: I'd remove this line actually

Jojo-Schmitz: OK

PR #6735, opened 2020-10-23
[MU3] Added dir pack for windows devel build on CI

igorkorsukov:
Added dir pack for windows devel build on CI for conveniences

### DIFF
--- a/.github/workflows/ci_tx2s3.yml
+++ b/.github/workflows/ci_tx2s3.yml
@@ -3,7 +3,7 @@ name: CI_Run_tx2s3
 on:
   pull_request:
     branches:
-    - 3.x
+    - 3.x_off
   schedule:
     - cron:  '0 */6 * * *' # Every 6 hour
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_tx2s3.yml
```diff
@@ -3,7 +3,7 @@ name: CI_Run_tx2s3
 on:
   pull_request:
     branches:
-    - 3.x
+    - 3.x_off
   schedule:
     - cron:  '0 */6 * * *' # Every 6 hour
```
Jojo-Schmitz, line 8: Every 6 hours is way to seldom, please make it every 15 minutes, like it was before on Travis, this helps the translators a lot in testing their changes/additions, and is what has been documented too.

`- cron:  '*15 * * * *' # Every 15 minutes`

igorkorsukov: ok, I will change it

PR #6736, opened 2020-10-23
[MU3] Changed pack type for Windows build on CI 

igorkorsukov:
Changed pack type for Windows build on CI for conveniences

* ~~for devel build (pull requests) - just dir~~
* for nightly build - 7z


igorkorsukov:
I returned 7z for devel builds 
because artifact size:
* with out 7z  - 291 Mb
* with 7z - 93 Mb

will take longer to download than unpack

PR #6738, opened 2020-10-23
[MU3] Restructured linux package script for CI

igorkorsukov:
I wanted make 7z pack for pullrequests builds Linux (instead AppImage), I think this is more convenient. But this is no convenient.

So, keeped AppImage. 
And made package script restructure

### DIFF
--- a/build/ci/linux/package.sh
+++ b/build/ci/linux/package.sh
@@ -3,232 +3,56 @@
 echo "Package MuseScore"
 trap 'echo Package failed; exit 1' ERR
 
-ARTIFACTS_DIR="build.artifacts"
-
-echo "=== ENVIRONMENT === "
-
-ENV_FILE=./../musescore_environment.sh
-cat ${ENV_FILE}
-. ${ENV_FILE}
-
-echo " "
-appimagetool --version
-echo " "
-linuxdeploy --list-plugins
-echo "===================="
-
-
-##########################################################################
-# BUNDLE DEPENDENCIES INTO APPDIR
-##########################################################################
-
-prefix="$(cat build.release/PREFIX.txt)" # MuseScore was installed here
-cd "$(dirname "${prefix}")"
-appdir="$(basename "${prefix}")" # directory that will become the AppImage
-
-# Prevent linuxdeploy setting RUNPATH in binaries that shouldn't have it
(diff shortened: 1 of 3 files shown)


shoogle:
@igorkorsukov, why did you originally think 7z was more convenient than AppImage, and what changed your mind? (Just curious)

igorkorsukov:
> @igorkorsukov, why did you originally think 7z was more convenient than AppImage, and what changed your mind? (Just curious)

This is not a simple story and quietly stupid  :)
It all started with Windows, for it used 7z for pull requests builds - this is not convenient, because it comes to unpack 2 archives, the zip that GitHub makes and 7z in which the application is the package.

I had two goals:
1. avoid unpacking the second 7z archive
2. do similar for all platforms

I did windows build without the second 7z archive, but artifact size became 291Mb (instead 92 Mb) - will take longer to download than unpacking, so I returned 7z.

But I wanted to do similar for all platform. So I tried to 7z pack and for Linux.
I did it, but saw, that this is not convenient, because needs unpack and more difficult launch app (need think before doing :) )

The second thing, why I wanted to do it, is I thinking, that better for devel (and maybe nightly) builds use local settings, which stored in the current dir (like Windows Portable). But this is the only idea.

shoogle:
@igorkorsukov, makes sense! I [sent feedback](https://support.github.com/contact/feedback) to GitHub asking for the Zip compression to be removed. (FYI, the space saving with 7z is because it looks for similarities between different files in the archive, whereas Zip compresses each file individually.)

You probably know already, but MuseScore's dev builds already save settings to a different folder than release builds. It's not "local" though since it is shared by all non-release builds.

However, the AppImage *does* have the ability to save preferences locally. If you have `MuseScore-XY.Z-arch.AppImage` then you can create folders in the same directory:

- `MuseScore-XY.Z-arch.AppImage.home` ← documents are stored here
- `MuseScore-XY.Z-arch.AppImage.config` ← settings are stored here

This is an automatic feature [built-in to the AppImage format](https://discourse.appimage.org/t/portable-configuration-data/105/7). Basically the AppImage runtime checks for those directories on startup and sets the enviroment variables `${HOME}` and `${XDG_CONFIG_HOME}` to the appropriate values. This makes it easy to carry the AppImage around on a USB stick along with all its documents and settings.

PR #6740, opened 2020-10-23
[MU4] Parts dialog

RomanPudashkin:
parts_dialog2


framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
```diff
@@ -54,6 +54,8 @@ FocusableItem {
             id: buttonIcon
 
             anchors.horizontalCenter: parent.horizontalCenter
+
+            font.pixelSize: root.iconPixelSize
```
vpereverzev, line 58: What is the motivation of this change? I believe it should depend on ui.theme

libmscore/score.cpp
```diff
@@ -392,24 +392,25 @@ Score::~Score()
 
 Score* Score::clone()
 {
```
vpereverzev, line 394: Could you explain literallly in few sentences what was wrong with the previous implementation?

libmscore/score.cpp
```diff
@@ -392,24 +392,25 @@ Score::~Score()
...
+    Score* copy = new Score(masterScore());
+    excerpt.setPartScore(copy);
+    copy->_style = _style;
+    masterScore()->addExcerpt(&excerpt);
+    Excerpt::createExcerpt(&excerpt);
+    masterScore()->removeExcerpt(&excerpt);
+    excerpt.setPartScore(nullptr);
 
-    score->addLayoutFlags(LayoutFlag::FIX_PITCH_VELO);
-    score->doLayout();
```
vpereverzev, line 411: I'm wondering if whether we still need this

libmscore/staff.h
```diff
@@ -279,6 +279,7 @@ class Staff final : public Element
     void setPlaybackVoice(int voice, bool val) { _playbackVoice[voice] = val; }
 
     std::array<bool, VOICES> visibilityVoices() const { return _visibilityVoices; }
+    bool voiceVisible(int voice) const { return _visibilityVoices[voice]; }
```
vpereverzev, line 282: I'd propose "isVoiceVisible" name

vpereverzev: also we need to gradually move implementation of methods into .cpp files

mu4/context/internal/globalcontext.cpp
```diff
@@ -79,6 +79,11 @@ void GlobalContext::setCurrentNotation(const INotationPtr& notation)
     }
 
     m_currentNotation = notation;
+
+    if (m_currentNotation) {
+        m_currentNotation->setOpened(true);
```
vpereverzev, line 84: The question arises, where will it become false?

libmscore/score.cpp
RomanPudashkin: The previous implementation was fully copied from MasterScore::clone. It created MasterScore instance and returned pointer to it. Hence Notation stored pointer to MasterScore and used overrided methods from MasterScore, not from Score

framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
RomanPudashkin: You added this property by yourself, but it not worked :)
In some cases need to change the size of an icon. In my case, this is X button. Icon of this button is smaller than other icons
изображение

mu4/context/internal/globalcontext.cpp
RomanPudashkin: After closing notation via X button in this panel
изображение

libmscore/score.cpp
RomanPudashkin: They are called from Excerpt::createExcerpt

libmscore/staff.h
RomanPudashkin: fixed

mu4/notation/internal/excerptnotation.cpp
```diff
@@ -41,7 +45,35 @@ ExcerptNotation::~ExcerptNotation()
...
+void ExcerptNotation::setExcerpt(Ms::Excerpt* excerpt)
+{
+    m_excerpt = excerpt;
+    setScore(m_excerpt->partScore());
+    setMetaInfo(m_metaInfo);
+}
+
+Meta ExcerptNotation::metaInfo() const
+{
+    return isInited() ? Notation::metaInfo() : m_metaInfo;
```
Eism, line 57: Strange condition. If initialized then should we return local data?

RomanPudashkin: If notation was inited with default constructor then it has no score. We use local data in this case

Eism (approved):

PR #4883, opened 2019-04-08
[MU4] Update to latest SMuFL version 1.3

Jojo-Schmitz:
As taken from https://github.com/w3c/smufl

Counterpart #6543 is in 3.x now

PR #6742, opened 2020-10-24
[MU3] reduce useless whitespace in JSON files; better way to update libmscore/sym.* from SMuFL (including bugfixes)

mirabilos:
Split-off PR, this one now only deals with the JSON and SMuFL topics; related MU4 PR is https://github.com/musescore/MuseScore/pull/4883

----

These are the first fruits from [my attempt to replace the Emmentaler accidentals with those from Parnassus in MScore](https://github.com/mirabilos/MuseScore/tree/parnassus-accidentals). These changes _do **not**_ yet change the font itself.

First off, I found over 12½ KB of useless whitespace at end-of-line in the font and SMuFL JSON files which getting rid of reduces the whole application size (by about 37K because Qt interns both latin1 and Unicode forms).

[…]

We could save even more space in the JSONs by re-indenting them to e.g. two spaces (as ECMA-262 suggests) or even one tab per level. Do tell me if you wish for me to update the PR to do just that.


Jojo-Schmitz:
We do take the SMuFL directly and as is from https://github.com/w3c/smufl, so better get these changes in there, else we''ll have merge conflicts when updating, like with #6543 (merged into 3.x) and #4883 (pending merge into master).

Jojo-Schmitz:
Hmm, maybe for 3.x this SMuFL change is OK, considering that SMuFL 1.4 most probably will never make it into 3.x, as 3.6 will be the last of that series. This change should not make it into master though until it comes from SMuFL directly.

mirabilos:
@Jojo-Schmitz good point, but the whitespace change (even the more daring one I suggested) can be automated; I’m assuming we don’t change those files ourselves.

I’d suggest doing so at build time, but considering we also build on Windows, that’s going to be tricky since it doesn’t even have `sed`.

I guess I’ll leave this in the 3.x PR as-is for now, and will just omit that part from the master PR then.

Jojo-Schmitz:
How much space would the _more daring one_ buy us? If it is 37k for those single (?) trailing spaces, it got to be 2 times more for a 2 space indent and 4 times more for a tab indent, right? So up to 5*37k, quite a lot!

Jojo-Schmitz:
And if we manage to do that at build time, it'd be even more, as we could eliminate all indentation.

mirabilos:
Yes, it’s a good 19.7% in total. I filed a PR against SMuFL with this: https://github.com/w3c/smufl/pull/135

This would save about 346K in the mscore binary. Granted, not *that* much compared to the total size, but the parser overhead would be even less.

(We could even strip the indent fully for the embedded version, saving a further 3*34½ kB.)

Jojo-Schmitz:
As https://github.com/w3c/smufl/pull/135 git merged into SMuFL, please adjust your's here to match that one ;-)

mirabilos:
Adjusted accordingly @Jojo-Schmitz ☻

Jojo-Schmitz:
Does this include the changes of all 3 commits of that SMuFL PR?

Jojo-Schmitz:
And please don't do this for master, I'll do it as part of #4883, and take it from SMuFL directly

mirabilos:
@Jojo-Schmitz it includes all these changes, but not (yet) the “follwed” fix. Should I add that, too?

Yes, I will not do that for master, only submit the code correction there.

Jojo-Schmitz:
Yes, the full package of the SMuFL PR please

mirabilos:
OK, updated with the SMuFL spelling fix, so these are now bitwise identical to the latest SMuFL

Jojo-Schmitz (approved):

Jojo-Schmitz (changes requested):

libmscore/sym.cpp
```diff
@@ -3473,14 +3470,15 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symUserNames = { {
...
-      QT_TRANSLATE_NOOP("symUserNames", "Lour\u00e9 (tenuto-staccato) below"),
+      QT_TRANSLATE_NOOP("symUserNames", "Louré (tenuto-staccato) above"),
+      QT_TRANSLATE_NOOP("symUserNames", "Louré (tenuto-staccato) below"),
       QT_TRANSLATE_NOOP("symUserNames", "Unstress above"),
       QT_TRANSLATE_NOOP("symUserNames", "Unstress below"),
       "Augmentation dot",
       QT_TRANSLATE_NOOP("symUserNames", "Dashed barline"),
       QT_TRANSLATE_NOOP("symUserNames", "Dotted barline"),
       QT_TRANSLATE_NOOP("symUserNames", "Double barline"),
+      "Final barline",
```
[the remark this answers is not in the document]
Jojo-Schmitz: Yes, please these:
```
    QT_TRANSLATE_NOOP("symUserNames", "Final barline"),
    QT_TRANSLATE_NOOP("symUserNames", "Heavy barline"),
    QT_TRANSLATE_NOOP("symUserNames", "Heavy double barline"),
    QT_TRANSLATE_NOOP("symUserNames", "Reverse final barline"),
    QT_TRANSLATE_NOOP("symUserNames", "Short barline"),
    QT_TRANSLATE_NOOP("symUserNames", "Single barline"),
    QT_TRANSLATE_NOOP("symUserNames", "Tick barline"),
```

fonttools/smufl2sym-in-trans.jsn
```diff
@@ -0,0 +1,341 @@
+{
+  "accSagittal11LargeDiesisDown": null,
```
Jojo-Schmitz, line 2: What is the purpose if this file? Why .jsn and not .json?

Jojo-Schmitz:
Wouldn't the `metadata.json` files (for at least 'our' fonts, for Bravura and Petaluma it might be worth a PR against those, maybe?) also benefit from the whitespace cleanup?

Jojo-Schmitz:
We should rather not touch `Petaluma/Metadat.json` and `Bravura/metada.json` though, not without Steinberg's buy-in

libmscore/sym.cpp
```diff
@@ -2842,6 +2844,8 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symNames = { {
       "windTrillKey",
       "windVeryTightEmbouchure",
       "windWeakAirPressure",
+      // SMuFL standard symbol IDs }}}
+
```
Jojo-Schmitz, line 2849: one empty line too much for my taste

libmscore/sym.cpp
```diff
@@ -3173,19 +3173,17 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symUserNames = { {
...
       QT_TRANSLATE_NOOP("symUserNames", "Combining lower by one 17-limit schisma"),
       QT_TRANSLATE_NOOP("symUserNames", "Combining lower by one 19-limit schisma"),
       QT_TRANSLATE_NOOP("symUserNames", "Combining lower by one 23-limit comma or 29-limit comma"),
       QT_TRANSLATE_NOOP("symUserNames", "Combining lower by one 31-limit schisma"),
       QT_TRANSLATE_NOOP("symUserNames", "Combining lower by one 53-limit comma"),
       "Combining open curly brace",
-      QT_TRANSLATE_NOOP("symUserNames", "Combining raise by one 17-limit schisma)"),
-      QT_TRANSLATE_NOOP("symUserNames", "Combining raise by one 19-limit schisma)"),
+      QT_TRANSLATE_NOOP("symUserNames", "Combining raise by one 17-limit schisma"),
+      QT_TRANSLATE_NOOP("symUserNames", "Combining raise by one 19-limit schisma"),
```
Jojo-Schmitz, line 3186: These two are my brown bag bugs :-/

libmscore/sym.cpp
```diff
@@ -4157,8 +4155,8 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symUserNames = { {
       "Toe",
       "Toe-click",
       "Toe-drop",
-      "Toe-dtep",
-      "Toe-dap",
+      "Toe-step",
+      "Toe-tap",
```
Jojo-Schmitz, line 4159: These two are likely my brown bag bugs :-/

libmscore/sym.cpp
```diff
@@ -4947,9 +4945,8 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symUserNames = { {
       "Oblique straight line SW-NE",
       "Oblique straight line tilted NW-SE",
       "Oblique straight line tilted SW-NE",
-      QT_TRANSLATE_NOOP("symUserNames", "Short trill"),
```
Jojo-Schmitz, line 4950: This is my brown bag bug :-/

libmscore/sym.cpp
```diff
@@ -4985,12 +4982,13 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symUserNames = { {
       "Shake (Muffat)",
       "Short oblique straight line NW-SE",
       "Short oblique straight line SW-NE",
+      QT_TRANSLATE_NOOP("symUserNames", "Short trill"),
```
Jojo-Schmitz, line 4985: This is my brown bag bug :-/

Jojo-Schmitz:
See also https://gist.github.com/lasconic/be08234b3dd074e99edc1fb060f0abcb for a Python script doing those conversions. Not as  complete as your script, but maybe more portable?

libmscore/sym.cpp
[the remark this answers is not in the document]
Jojo-Schmitz: I'll add those missing `QT_TRANSLATE_NOOP()`s in another PR, dealing with those additional barlines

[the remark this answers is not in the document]
mirabilos: We’ll need to add them to the JSON file as well, so better do it here and let #6752 depend on this one.

fonttools/smufl2sym-in-trans.jsn
mirabilos: As I documented in the commit message, this file contains all keys for which a QT_TRANSLATE_NOOP is to be generated.

mirabilos: And .jsn is because I grew up on MS-DOS with 8.3 filenames and prefer them… but if this is a sticking point, I can rename it.

libmscore/sym.cpp
mirabilos: The empty line makes it easier to see where the SMuFL-derived stuff begins and ends. (Yes, we also have the comment, but…)

If you insist, I can remove it.

mirabilos: Oh wait, I see. Sorry. Fixed.

mirabilos:
The Python script, unfortunately, not only barely does anything, it also doesn’t do that right (e.g. the Unicode characters, as you saw). It also doesn’t do the QT_TRANSLATE_NOOP thing. jq is pretty portable, and other than that it just needs sh (even the one from the Weihenstephan tools should suffice).

I even wrote the things in a way to depend on _only_ jq, not use sed and other tools, so we can even add a .bat file doing the same thing with jq for Windows…

mirabilos:
@Jojo-Schmitz I updated the PR to:

- not touch Pataluma (I already had kept fingers off Bravura)
- add those barlines as QT_TRANSLATE_NOOPs

Yes, the font JSON files would also benefit from the whitespace changes I submitted to SMuFL. But I think the other approach you suggested (to drop all optional whitespace at build time) is the generally better one.

Jojo-Schmitz:
Thanks
Considering how fast your SMuFL PR got accepted, it might be worthwhile to send a PR to Steinberg for Petaluma and Bravura too

mirabilos:
Joachim Schmitz dixit:

>Considering how fast your SMuFL PR got accepted, it might be worthwhile
>to send a PR to Steinberg for Petaluma and Bravura too

OK will do

mirabilos:
Updated to
- rename the `.jsn` file to `.json`
- split the [jq](https://stedolan.github.io/jq/) scripts into `.jq` files
  - make `smufl2sym.sh` just call those (GNU/Linux, BSD, Mac OSX)
  - add `smufl2sym.bat` doing the same on Windows, as requested by @Jojo-Schmitz 
    - depends on jq (see above) and [dos2unix](https://waterlan.home.xs4all.nl/dos2unix.html)
    - dependencies are documented in a comment at the top of the batch script
- removed the dependency on column(1)

Output is bytewise identical on GNU/Linux and Windows.

mirabilos:
Bravura PR: https://github.com/steinbergmedia/bravura/pull/36

Petaluma PR: https://github.com/steinbergmedia/petaluma/pull/6

mirabilos:
Updated with the new QT_TRANSLATE_NOOPS from commit b69e2a75d3639ea7aa9b31f1d9d0281d630fbf8b and commit 1f5ec3fc1d920bc4617efd9b81ff6df38051341f (no other change; rebasing on top of latest 3.x shows the resulting sym.cpp identical modulo a whitespace fix)

fonttools/smufl2sym-in-trans.jsn
Jojo-Schmitz: > As I documented in the commit message, this file contains all keys for which a QT_TRANSLATE_NOOP is to be generated.

Ah, sorry missed this reply... something to be very careful about, we'd need to remember adding those here **and** modifying in sym.cpp, or add just here **and** run the script, rather than adding **just** modifing in sym.cpp (as later run the script would revert that change).
Probably worth a comment in sym.cpp (and maybe even that comment added via this script, along the lines of "don't update the strings here manually, these are auto generated by ..."

mirabilos:
Joachim Schmitz dixit:

>Ah, sorry missed this reply... something to be very careful about, we'd
>need to remember adding those here **and** modifying in sym.cpp, or add
>just here **and** run the script, rather than adding **just** modifing
>in sym.cpp (as later run the script would revert that change).

I could move the generated parts out of sym.{cpp,h} and into separate
files that indicate they are generated and then #include these.

We could conceivably generate them at every build, but that’s overkill
plus requires jq, so, better to check them in I guess.

>Probably worth a comment in sym.cpp (and maybe even that comment added
>via this script, along the lines of "don't update the strings here
>manually, these are auto generated by ..."

Like that what I wrote above?

Jojo-Schmitz:
I'd leave sym.{cpp,h} as they are, just add that one line comment to the start of those, next to that `SMuFL standard symbol IDs`, making it `SMuFL standard symbol IDs, generated automagigally, so don't edit manually but use the script at ... instead`

mirabilos:
Joachim Schmitz dixit:

>I'd leave sym.{cpp,h} as they are, just add that one line comment to
>the start of those, next to that `SMuFL standard symbol IDs`, making it
>`SMuFL standard symbol IDs, generated automagigally, so don't edit
>manually but use the script at ... instead`

Hm, okay. I guess you can just put that comment in the three places
in these files manually, if it’s above the “SMuFL standard symbol IDs”
it’ll be outside of what the script generates and thus kept.

mirabilos:
@Jojo-Schmitz https://github.com/musescore/MuseScore/compare/ff733f2b050d70f650c816746c40d81c79a7a76f..c27d0317b8f9970d7a8f353f2bdace1a9c383bed what do you think?

Can you do the same for the MU4 PR #4883 ?

Jojo-Schmitz:
Yes. Tomorrow...

libmscore/sym.cpp
```diff
@@ -51,6 +51,12 @@ std::array<uint, size_t(SymId::lastSym)+1> ScoreFont::_mainSymCodeTable { {0} };
 QHash<QString, SymId> Sym::lnhash;
 const std::array<const char*, int(SymId::lastSym)+1> Sym::symNames = { {
       "noSym",
+
+      // DO NOT edit the SMuFL standard symbol IDs (see below) manually!
```
Jojo-Schmitz, line 55: `symbols IDs` vs. `symbol names`?

libmscore/sym.cpp
```diff
@@ -2879,14 +2886,19 @@ const std::array<const char*, int(SymId::lastSym)+1> Sym::symNames = { {
 
 const std::array<const char*, int(SymId::lastSym)+1> Sym::symUserNames = { {
       QT_TRANSLATE_NOOP("symUserNames", "No symbol"),
+
+      // DO NOT edit the SMuFL standard symbol IDs (see below) manually!
```
Jojo-Schmitz, line 2890: `symbols IDs` vs. `symbol user names`?

mirabilos:
Joachim Schmitz dixit:

>+      // DO NOT edit the SMuFL standard symbol IDs (see below) manually!
>
>`symbols IDs` vs. `symbol names`?

This was what the original had… okay, the original had them
only in sym.h… maybe “the order of these items is from the
SMuFL IDs” is a justification? ;-)

Jojo-Schmitz:
Check https://github.com/musescore/MuseScore/pull/4883/commits/8d56d717ed84cb1582cad856c3166391fe5b7471 and https://github.com/musescore/MuseScore/pull/4883/commits/b736fc0decf690c083d5d61ef092598f806680d9

mirabilos:
Updated this one accordingly (modulo the “namess” typo).

Jojo-Schmitz:
Typo fixed, check https://github.com/musescore/MuseScore/pull/4883/commits/7021bc9d9f38c49e45bf0ac2461019644ddd7382

PR #6743, opened 2020-10-24
Fix #257581, Fix #64636  - Changes to Measure properties are not propagated between score and existing linked parts

njvdberg:
Resolves: https://musescore.org/en/node/257581
Resolves: https://musescore.org/en/node/64636

When reading a mscx file containing parts, link all measures of the part to the corresponding measure in the MasterScore.

### DIFF
--- a/libmscore/read302.cpp
+++ b/libmscore/read302.cpp
@@ -24,6 +24,7 @@
 #include "excerpt.h"
 #include "spanner.h"
 #include "scoreOrder.h"
+#include "measurebase.h"
 
 #ifdef OMR
 #include "omr/omr.h"
@@ -192,6 +193,7 @@ bool Score::read(XmlReader& e)
                         ex->setPartScore(s);
                         e.setLastMeasure(nullptr);
                         s->read(e);
+                        s->linkMeasures(m);
                         ex->setTracks(e.tracks());
                         m->addExcerpt(ex);
                         }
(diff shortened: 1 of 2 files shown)


libmscore/read301.cpp
```diff
@@ -289,6 +291,27 @@ bool Score::read(XmlReader& e)
...
+void Score::linkMeasures(Score* score)
+      {
+      MeasureBase *mbMaster = score->first();
+      for (MeasureBase* mb = first(); mb; mb = mb->next()) {
+            if (!mb->isMeasure())
+                  continue;
+            while (mbMaster && !mbMaster->isMeasure())
+                  mbMaster = mbMaster->next();
+            if (!mbMaster) {
+                  qDebug("Measuses in MasterScoare and Score are not in sync.");
```
Jojo-Schmitz, line 307: typo

njvdberg: Oops

Jojo-Schmitz:
I wonder whether this is needed in read206.cpp too?

njvdberg:
That part of the code looks very similar in both read206.cpp and read301.cpp indeed. Will check with a MuseScore 2.x file.

njvdberg:
> I wonder whether this is needed in read206.cpp too? (certainly not in read114.cpp though)
Just checked, same change is **not** required in <code>read206.cpp</code>. As a matter of fact, this issue is a regression since 3.0 (educated guess, not checked). In the 2.06 version of the file, the file number was included in the <code>Measure</code> tag as an attribute <code>number</code>. This attribute is removed in file version 3.01, introducing the issue.

Jojo-Schmitz:
OK, thanks for checking!

libmscore/read301.cpp
```diff
@@ -289,6 +291,27 @@ bool Score::read(XmlReader& e)
       return true;
       }
 
+//---------------------------------------------------------
+// linkMeasures
+//---------------------------------------------------------
+
+void Score::linkMeasures(Score* score)
+      {
+      MeasureBase *mbMaster = score->first();
```
vpereverzev, line 300: I think MeasureBaseList is a good candidate for an implementation of this method

So here you can leave something like that:
`if (!score || _measure.size() == 0)
      return;

_measures->link(score->measures());
`

Also I'd suggest to add a very simple method to MeasureBaseList - "bool hasNext() const;"
It'll slightly improve your loops.

Jojo-Schmitz:
rebase needed

njvdberg:
Rebase done

Jojo-Schmitz (approved):

PR #6744, opened 2020-10-24
[MU3] Add support for triple sharp and triple flat

Jojo-Schmitz:
SMuFL provides glyphs for triple sharp and triple flat, so they might as well be available to use in MuseScore. This involves adding 14 new TPCs and expanding the range of the AccidentalVal enum class.

This was requested on the forum topic in [Triple sharps and triple flats](https://musescore.org/en/node/304503)  _(sort of, ignore the digressions into politics and key signatures)_.

See https://seanyeh.com/pages/alkans_triple_sharp/ for an example of a triple sharp being used in a piece of music from the Romantic period.

Rebased version of #6002 (@mattmcclinch, please review).

"If the mountain won't come to Mohammed, Mohammed must go to the mountain."  ;-)


Jojo-Schmitz:
I'm puzzled about those mtest failures in this PR here, they didn't happen in #6002

libmscore/key.cpp
```diff
@@ -188,7 +188,7 @@ AccidentalVal AccidentalState::accidentalVal(int line, bool &error) const
             error = true;
             return AccidentalVal::NATURAL;
             }
-      return AccidentalVal((state[line] & 0x0f) - 2);
+      return AccidentalVal((state[line] & 0x0f) - + int(AccidentalVal::MIN));
```
mattmcclinch, line 191: There is an extraneous `-` on this line. This is the cause of some and perhaps all of the mtest failures. It should be
```
      return AccidentalVal((state[line] & 0x0f) + int(AccidentalVal::MIN));
```

libmscore/key.cpp
```diff
@@ -198,15 +198,15 @@ AccidentalVal AccidentalState::accidentalVal(int line, bool &error) const
 
 void AccidentalState::init(Key key)
       {
-      memset(state, 2, MAX_ACC_STATE);
+      memset(state, + int(AccidentalVal::MAX), MAX_ACC_STATE);
```
mattmcclinch, line 201: There is an extraneous `+` here.

libmscore/key.cpp
```diff
@@ -230,7 +230,7 @@ void AccidentalState::init(Key key)
 void AccidentalState::init(const KeySigEvent& keySig, ClefType clef)
       {
       if (keySig.custom()) {
-            memset(state, 2, MAX_ACC_STATE);
+            memset(state, + int(AccidentalVal::MAX), MAX_ACC_STATE);
```
mattmcclinch, line 233: There is an extraneous `+` here.

libmscore/key.cpp
Jojo-Schmitz: Outch,  copy&paste error, thanks!

libmscore/key.cpp
Jojo-Schmitz: Dito

libmscore/key.cpp
Jojo-Schmitz: Dito

libmscore/key.cpp
```diff
@@ -198,15 +198,15 @@ AccidentalVal AccidentalState::accidentalVal(int line, bool &error) const
 
 void AccidentalState::init(Key key)
```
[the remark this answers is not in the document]
Jojo-Schmitz: Hmm,  the changes of this PR don't seem complicate at all, they are just adjusted to the change number of accidentals (that affect playback in semitones).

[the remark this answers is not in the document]
mattmcclinch: See https://github.com/mattmcclinch/MuseScore/commit/345042bf8f1d313c35d79c6b437044e70bca4e8e. I think it adds a bit more clarity. The variable `octave` is perhaps a bit misleading. `octaveStart` might be better.

PR #6751, opened 2020-10-25
remove debugging output “reading ticks <%s>”

mirabilos:
(if it’s not going to be used/fixed, it’s annoying on the console)

### DIFF
--- a/libmscore/xmlreader.cpp
+++ b/libmscore/xmlreader.cpp
@@ -185,7 +185,6 @@ Fraction XmlReader::readFraction()
       if (!s.isEmpty()) {
             int i = s.indexOf('/');
             if (i == -1) {
-                  qDebug("reading ticks <%s>", qPrintable(s));
                   return Fraction::fromTicks(s.toInt());
                   }
             else {


libmscore/xmlreader.cpp
```diff
@@ -185,7 +185,6 @@ Fraction XmlReader::readFraction()
       if (!s.isEmpty()) {
             int i = s.indexOf('/');
             if (i == -1) {
-                  qDebug("reading ticks <%s>", qPrintable(s));
```
Jojo-Schmitz, line 188: It is seen in Debug builds only though, isn't it?

mirabilos: [“This function does nothing if `QT_NO_DEBUG_OUTPUT` was defined during compilation.”](https://doc.qt.io/Qt-5/qtglobal.html#qDebug)

I see. `QT_NO_DEBUG_OUTPUT` is defined for RELEASE builds. For some reason, Debian builds are neither DEBUG nor RELEASE but "" (empty string).

Feel free to disregard this or merge it, I don’t mind either way.

Jojo-Schmitz: I guess Debian should do "RELEASE" builds then?

mirabilos:
Joachim Schmitz dixit:

>I guess Debian should do "RELEASE" builds then?

Not in my hands, unfortunately. Debian provides a wrapper
that calls cmake “correctly” and handles probably dozens
of things automatically which I don’t even know of, maybe
even cross compiling, but that wrapper also selects this
build type, and when I asked this seems to be normal.

But, as I said, I can live with that. Feel free to close this PR.

Jojo-Schmitz:
Maybe we should set `QT_NO_DEBUG_OUTPUT` in non-DEBUG builds then?

Jojo-Schmitz:
Does Debian builds disable the asserts?

mirabilos:
Joachim Schmitz dixit:

>Maybe we should set `QT_NO_DEBUG_OUTPUT` in non-DEBUG builds then?

Unsure. The messages themselves can help. If someone doesn’t want
them, they can add it to CPPFLAGS themselves. Better to not force
that.

>Does Debian builds disable the asserts?

Yes.

vpereverzev:
The very annoying one and also it doesn't provide any necessary information, even in debug mode.
Merged

PR #6754, opened 2020-10-25
Fix #312146 - "Flatten all Beams" setting ignored on reload

njvdberg:
Resolves: https://musescore.org/en/node/312146

No longer overwrite <code>Beam::_noSlope</code> after setting the style value.

Is issue is a regression caused by [fixed some issues from PVS-Studio report (step1)](https://github.com/musescore/MuseScore/commit/68b3415c475805ad49839ea65609dfbda796f602?branch=68b3415c475805ad49839ea65609dfbda796f602) which set in the <code>Beam</code>  the variable <code>_noSlope</code> back to <code>false</code>, overwriting the style value.

### DIFF
--- a/libmscore/beam.cpp
+++ b/libmscore/beam.cpp
@@ -58,7 +58,6 @@ struct BeamFragment {
 Beam::Beam(Score* s)
    : Element(s)
       {
-      initElementStyle(&beamStyle);
       _direction       = Direction::AUTO;
       _up              = true;
       _distribute      = false;
@@ -73,6 +72,7 @@ Beam::Beam(Score* s)
       maxMove          = 0;
       _isGrace         = false;
       _cross           = false;
+      initElementStyle(&beamStyle);
       }
 
 //---------------------------------------------------------


libmscore/beam.cpp
```diff
@@ -62,7 +62,6 @@ Beam::Beam(Score* s)
       _direction       = Direction::AUTO;
       _up              = true;
       _distribute      = false;
-      _noSlope         = false;
```
vpereverzev, line 65: Where does the "overwriting" happen? I'm curious because here I see normal value initialization, perhaps we need to find out the "root" of this issue somwhere

njvdberg: <code>_noSlopen</code> is a style parameters and set in <code>setProperty()</code> which is called by <code>initElementStyle()</code>, just a few line before this one. So, <code>_noSlope</code> is set as a style parameter and set to <code>false</code> here.
But make sure, <code>_noSlope</code> is always initialised, it might be better to move <code>initElementStyle()</code> to **after** the initialisation of the parameters.

njvdberg: BTW, it seems more elements might suffer this same issue, style parameter overwritten by initialisation. Is it better to check and extend this PR or create a new one?

igorkorsukov: PVS Studio reported that member value not initialized by default. So I added this code, like others.
I agree, better move initElementStyle() to after the initialisation of the parameters.

vpereverzev: Let's add the write call ordering here and then I'll merge it
If there is the same problem in other elements, then I'd solve it in a separate PR

njvdberg: @igorkorsukov Thanks for the explanation.
Update in on its way and after that, I'll check the other elements and create a separate PR for that.

vpereverzev (approved):

PR #6757, opened 2020-10-25
fix #308726: Avoid rest offsets with slash notation

tobik:
Resolves: https://musescore.org/en/node/308726

Do not offset rests if slash notation is active anywhere in the chord 

### DIFF
--- a/libmscore/rest.cpp
+++ b/libmscore/rest.cpp
@@ -532,6 +532,18 @@ int Rest::computeLineOffset(int lines)
                   }
             }
 
+      if (offsetVoices) {
+            // if the staff contains slash notation then don't offset voices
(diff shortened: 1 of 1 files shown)


MarcSabatella:
See https://musescore.org/en/node/318800 - I think this probably is what broke rests within the rhythmic slash notation.  If I'm following this correctly, the original problem was that rests *not* in slash notation were being offset unnecessarily.  Now, unfortunately, the rests that *are* in slash are not being offset either, and they definitely need to be (and were previously).

Probably the code added here needs to be skipped if the current rest is itself in slash notation.  Well, at least if in voice 3 or 4, which are the ones that need to be offset.

PR #6758, opened 2020-10-25
[MU3→4] Parnassus; indicate pitch, on/off time; debugging

mirabilos:
MU4 PR for the following PRs that were merged into MU3:
- https://github.com/musescore/MuseScore/pull/6751
- https://github.com/musescore/MuseScore/pull/6750
- https://github.com/musescore/MuseScore/pull/6747


fonts/mscore/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "fontName": "Emmentaler",
-    "fontVersion": 0.1,
+    "fontVersion": 2.002,
```
Marr11317, line 3: wow, big step, is this normal?

mirabilos: It never stepped. The font itself also has version numbers, which didn’t even follow normal font versioning schemes.

I’ve increased this to major version 2 to be sure it will register as newer than any old ones, and the minor 2 is from the F8vb/F15vb clef changes.

Jojo-Schmitz:
I guess the vtest differences are expected here

PR #6766, opened 2020-10-26
[MU4] Drawing multiple pages

RomanPudashkin:
- implemented drawing multiple pages of notation

### DIFF
--- a/mu4/notation/internal/notationconfiguration.cpp
+++ b/mu4/notation/internal/notationconfiguration.cpp
@@ -105,6 +105,21 @@ Channel<QColor> NotationConfiguration::backgroundColorChanged() const
     return m_backgroundColorChanged;
 }
 
+QColor NotationConfiguration::pageColor() const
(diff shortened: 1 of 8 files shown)


vpereverzev (approved):

mu4/notation/internal/notationconfiguration.cpp
```diff
@@ -105,6 +105,21 @@ Channel<QColor> NotationConfiguration::backgroundColorChanged() const
     return m_backgroundColorChanged;
 }
 
+QColor NotationConfiguration::pageColor() const
+{
+    return QColor("#ffffff");
```
vpereverzev, line 110: I don't remember if whether these properties are depending on some "style" or "preferences" values. Have you ckecked it already?

RomanPudashkin: At this moment they don't depend on style or preferences

PR #6767, opened 2020-10-26
[MU4] Implemented notation search

Eism:
Implemented search in notation
Ps: this is not the final design

Peek 2020-10-26 17-14

### DIFF
--- a/mu4/notation/view/searchpopupmodel.h
+++ b/mu4/notation/view/searchpopupmodel.h
@@ -1,3 +1,5 @@
+//=============================================================================
+//  MuseScore
 //  Music Composition & Notation
 //
 //  Copyright (C) 2020 MuseScore BVBA and others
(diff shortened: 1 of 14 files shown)


Marr11317:
In your gif, it looks weird when the C rehearsal mark is shown: it's in the top left corner, even though we would've expected the view to scroll so it would be in the bottom right corner.

mu4/notation/view/searchpopupmodel.h
```diff
@@ -0,0 +1,46 @@
...
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#ifndef MU_NOTATION_SEARCHPOPUPMODEL_H
+#define MU_NOTATION_SEARCHPOPUPMODEL_H
+
+#include <QtQml>
```
vpereverzev, line 22: Do you really need this guy? :)

mu4/notation/internal/notationsearchservice.cpp
```diff
@@ -0,0 +1,175 @@
...
+NotationSearchService::SearchData NotationSearchService::parseRehearsalMarkCommand(const std::string& text) const
+{
+    SearchData data;
+
+    QString qtext = QString::fromStdString(text).toLower();
+
+    int rehearsalMarkCodeLength = static_cast<int>(REHEARSAL_MARK_CODE.length());
+    if (qtext.startsWith(REHEARSAL_MARK_CODE.c_str()) && qtext.size() > rehearsalMarkCodeLength) {
+        data.elementType = ElementType::REHEARSAL_MARK;
+        data.value = qtext.mid(rehearsalMarkCodeLength);
```
vpereverzev, line 167: Probably need some explanation here
Did you moved it from MU3?

Eism: I didn't copy it(almost) :)
I am getting command and value from string here.
if the string starts with a specific command then the value will start from commandLenght to the end of the string.

mu4/notation/view/searchpopupmodel.h
Eism: fixed

RomanPudashkin (approved):

PR #5853, opened 2020-03-20
[MU3] Fix #127341: Implements portamento for fluid

Jojo-Schmitz:
to replace/rebase @MichaelFroelich's #4824

Resolves https://musescore.org/en/node/127341

PR #4824, opened 2019-03-23
fix #127341: implements portamento for fluid

MichaelFroelich:

PR #6768, opened 2020-10-26
[MU4] Fix #127341: Implements portamento for fluid

Jojo-Schmitz:
Resolves https://musescore.org/en/node/127341

to replace/rebase @MichaelFroelich's #4824 resp. #5853, but for master.
It uses the old fluidsynth, not the new one though, porting it to that is beyond my paygrade...,


vpereverzev:
@psmokotnin please, take a look on this

vpereverzev:
@psmokotnin is working on the new version of our powerful audio-engine. So I don't think that we'll merge this PR, but I'm wondering if whether we can extract some useful changes here. @psmokotnin

PR #6772, opened 2020-10-27
[MU3] Fix #310741: Always update Shadow Note

MLK97:
Resolves: https://musescore.org/en/node/310741

Changing the duration via keyboard shortcut of the note while in note-input mode didn't change the appearance of the shadow note or make it disappear.
Removing the visibility-check removes that issue.


MLK97:
I just got aware, that it probably needs some discussion, whether this is actually a wished behavior.
While in the mentioned issue the shadow note is a wished feature during keyboard input, other users like to have the shadow note hidden.

PR #6776, opened 2020-10-28
Resizing vbox accordingly to the content size

vpereverzev:
Now the height of the vbox depends on the size of its content.

Earlier, on some scores, the title page might look like this:

изображение

And here is what we have now:

Peek 2020-10-22 22-35

### DIFF
--- a/libmscore/box.cpp
+++ b/libmscore/box.cpp
@@ -321,6 +321,16 @@ void Box::add(Element* e)
       MeasureBase::add(e);
       }
 
+QRectF Box::contentRect() const
+      {
+      QRectF result;
+
+      for (const Element* element : el())
+            result = result.united(element->bbox());
+
+      return result;
+      }
+
 //---------------------------------------------------------
 //   getProperty
 //---------------------------------------------------------
@@ -724,6 +734,16 @@ VBox::VBox(Score* score)
       setLineBreak(true);
       }
 
+qreal VBox::minHeight() const
+      {
+      return point(Spatium(10));
+      }
+
+qreal VBox::maxHeight() const
+      {
+      return point(Spatium(30));
+      }
+
 //---------------------------------------------------------
 //   layout
 //---------------------------------------------------------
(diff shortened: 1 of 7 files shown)


MarcSabatella:
As logical as this seems, there are definitely, some, actually probably very many, scores that *rely* on this behavior, to created free-floating text and other various special effects.

So, I would propose any change like this not affect existing scores, only new ones.  One way to do this might be to make this auto-resize a property of the frame that has a default value of "off", but make sure the command to add a new frame always explicitly turns it on.  But also one could leverage the same hack I used for chord symbol playback, checking version on read.  Probably there would be other ways too.

Jojo-Schmitz:
Yes, I agree with @MarcSabatella. Also it'd better use a skyline, here "Attention" and "Words...) wouldn't collide it the latter were higher up.

vpereverzev:
@MarcSabatella Previously the height was hardcoded (10sp). Now I've introduced it as a minimum height, so if the content is not bigger than 10sp then nothing will be changed. So these boxes on the older scores will not be suddenly reduced

Jojo-Schmitz:
But they will get suddenly enlarged, which may destroy a carefully created layout

MarcSabatella:
The *default* height may have been hardcoded, but people can and did resize these to be bigger or smaller as they see fit.  Many people use completely empty frames, in fact, in places where they could have used spacers.  So if someone set a frame explicitly to be 4sp tall, or 0sp, or 20sp, they may well have done so for very specific reasons that may or may not relate to the contents.

Again, I love the fact that going forward vertical frames might start resizing by default.  But people will still want the ability to size them manually when desired, and they absolutely will want existing scores to not be messed up by this.  So somehow there needs to be a control for this autosizing, probably a property and checkbox in Inspector, and the autosizing needs to default to off for existing frames in existing scores.

MarcSabatella:
Of course, I realize there are presumably plans to offer people a choice upon import of pre-3.6 scores - do you want to take advantage of the new defaults.  And if someone elects to do so, then the autosizing would be appropriate.  I do like the idea of providing an easy "fix" for scores with badly formatted titles like the one shown.  It just can't come at the expense of people who carefully sized their frames for specific purposes.  I myself probably have hundreds of such scores.

vpereverzev:
@MarcSabatella I didn't expect this to be so important for existing users, because initially I started from the problems of users when importing scores. Thank you for explaining in detail the user cases with the old scores.

So I definitely agree with your point, it makes sense for me. I'd add a proper setting for this property and will try to consider all known issues

MarcSabatella:
By import, you mean from MusicXML or some such?  In that case, it probably makes sense to enable by default, since most likely that kind of formatting info won't have survived the conversion anyhow.

vpereverzev:
> By import, you mean from MusicXML or some such? In that case, it probably makes sense to enable by default, since most likely that kind of formatting info won't have survived the conversion anyhow.

Yes, mostly MusicXML
I've turned out this only for "new" scores, when you're importing some scores they are considered as "new" as well. And also now we have a separate option in inspector for that case.

PR #6777, opened 2020-10-28
Removed an accidental ledge from stem

vpereverzev:
Inspired by the request from @Tantacrul and @oktophonie

Before:

Снимок экрана от 2020-10-23 17-06-14

After:

Снимок экрана от 2020-10-23 17-04-11

### DIFF
--- a/libmscore/chord.cpp
+++ b/libmscore/chord.cpp
@@ -1425,7 +1425,7 @@ qreal Chord::defaultStemLength() const
             stemLen *= lineDistance;
 
       const qreal sgn = up() ? -1.0 : 1.0;
-      qreal stemLenPoints = stemLen * _spatium;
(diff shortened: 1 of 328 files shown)


cbjeukendrup:
> Wouldn't then the Flag need that extra round cap?

Good question; I would be interested to see how it looks if the top of the flag does not coincide with a staff line, so, if the note is one line higher.

Jojo-Schmitz:
Just flat.

Jojo-Schmitz:
Funny, regardless the vtests label the vtests get reported as failing?

Jojo-Schmitz:
I'm pretty sure the mtest failure is not this PR's fault, got to be the one just before mine (from PR #6776), @vpereverzev ?

PR #6782, opened 2020-10-29
[MU4] Sync CI scripts from 3.x to master

igorkorsukov:
Sync CI scripts from 3.x to master

### DIFF
--- a/build/ci/linux/runmtests.sh
+++ b/build/ci/linux/runmtests.sh
@@ -3,21 +3,22 @@
 echo "Run MuseScore mtest"
 trap 'echo Run tests failed; exit 1' ERR
 
-df -k .
+df -h .
 
 . ./../musescore_environment.sh
 
-cd build.release/mtest
+cd build.debug/mtest
 
-# vnc is the only tested platform plugin that allows to run
-# mscore executable in the used Travis environment.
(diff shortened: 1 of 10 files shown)


build/ci/linux/runmtests.sh
```diff
@@ -7,17 +7,18 @@ df -k .
...
 make -j2
 
 df -k .
 
-xvfb-run -a ctest -j2 --output-on-failure
+ctest -j2 --output-on-failure
 
-df -k .
\ No newline at end of file
+df -k .
```
Jojo-Schmitz, line 24: Not quite sure what the purpose of these is (I guess some debugging output), but a) IIRC `-k` is the default as per POSIX, and b) Linux' df's `-h' might be more helpful, 'human readable'?

igorkorsukov: We had a problem with free space.
I added this for monitoring
I will change to `-h`, thanks.

igorkorsukov: https://github.com/musescore/MuseScore/pull/6687

Jojo-Schmitz: Didn't notice those back then ;-)

PR #6783, opened 2020-10-29
[MU3] Sync CI scripts from master to 3.x

igorkorsukov:
Sync CI scripts from master to 3.x


.github/workflows/ci_windows.yml
```diff
@@ -89,7 +86,7 @@ jobs:
       if: env.DO_APPCAST == 'true'   
       shell: bash
       run: |
-        bash ./build/ci/tools/sparkle_appcast_gen.sh -p windows   
+        bash ./build/ci/tools/sparkle_appcast_gen.sh -p windows
```
Jojo-Schmitz, line 89: still a trailing space left here

build/ci/linux/setup.sh
```diff
@@ -136,12 +137,13 @@ cmake --version
 # POST INSTALL
 ##########################################################################
 
-chmod +x "${ENV_FILE}"
+chmod +x "$ENV_FILE"
```
Jojo-Schmitz, line 140: I believe the curly braces should rather stay, for it to not evailuate `$ENV` and append `"_FILE"`to it?

.github/workflows/ci_macos.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         echo "BUILD_MODE=$BUILD_MODE" >> $GITHUB_ENV
 
         DO_PUBLISH='false'
-        if [ "${{ github.event.inputs.publish }}" == "on" ]; then 
+        if [[ "${{ github.event.inputs.publish }}" == "on" || "$BUILD_MODE" == "nightly_build" ]]; then
```
Jojo-Schmitz, line 42: Probably doesn't matter, but this construct requires bash

build/ci/linux/setup.sh
igorkorsukov: > I believe the curly braces should rather stay, for it to not evailuate `$ENV` and append `"_FILE"`to it?

I agree, I will return it

.github/workflows/ci_macos.yml
igorkorsukov: On MacOS by default bash as shell

Jojo-Schmitz: Yes, I know. Just mentioning it, it is not a POSIX shell feature.

.github/workflows/ci_windows.yml
igorkorsukov: removed

build/ci/linux/setup.sh
igorkorsukov: returned

PR #6785, opened 2020-10-29
[MU4] Require Qt 5.15 as the minimum

Jojo-Schmitz:
for real this time, the previous method of only setting `QT_MIN_VERSION` just worked for Qt 4...

See also 484f8dc (PR #6588)
Counterpart to 3.x's #6786

### DIFF
--- a/build/FindQt5.cmake
+++ b/build/FindQt5.cmake
@@ -36,6 +36,8 @@ if (WIN32)
       )
 endif(WIN32)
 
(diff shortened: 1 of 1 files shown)


build/FindQt5.cmake
```diff
@@ -1,5 +1,4 @@
 set(_components
-    Core
```
Jojo-Schmitz, line 2: This needs to stay, to get added to the lists

PR #6789, opened 2020-10-29
fix #312083: tempofield in new score wizard not accessible

MarcSabatella:
Resolves: https://musescore.org/en/node/312083

A quirk in Qt prevents screen readers from recognizing UI elements
that are part of groups that were disabled in the UI file.
Fix is to enable the group in the UI file
but then disable it in the code.


mscore/newwizard.cpp
```diff
@@ -57,6 +57,7 @@ TimesigWizard::TimesigWizard(QWidget* parent)
       connect(tsCutTime,    SIGNAL(toggled(bool)), SLOT(cutTimeToggled(bool)));
       connect(tsFraction,   SIGNAL(toggled(bool)), SLOT(fractionToggled(bool)));
       pickupMeasure->setChecked(false); // checked in the UI file to enable screen reader on pickup duration controls
+      tempoGroup->setChecked(false);
```
Jojo-Schmitz, line 60: Might be worth a comment in the code?

MarcSabatella: Even though it’s the same comment as the line above?

Jojo-Schmitz: Ah, missed that

PR #6538, opened 2020-09-04
Algorithm for spreading staves/systems over a page

njvdberg:
This PR implements [automatic vertical justification of staves](https://trello.com/c/feV4aI2I/6-implement-automatic-vertical-justification-of-staves) of the [MuseScore Engraving Push](https://trello.com/b/JVGhCNm6/musescore-engraving-push)

The algorithm
-------------
After the normal layout of a page, a new algorithm tries the spread the staves in such way the space between the lowest staff and the bottom margin of the page is used too by adding extra space between the staves.
Not all gaps between the staves will get the same amount of space but the algorithm distinguishes 3 types of gaps:

1) Gaps between the systems.

2) Gaps between the staves within a system, grouped by a [ bracket.

3) Gaps between the staves witinh a system, but outside a [ bracket.

The space between the lowest staff and the bottom margin of the page is divided over these gaps using a weight factor for each of these types of gaps.

For an example of how this algorithm works, see these 2 pages of a score:

before

This pages contains 12 staves in 2 systems.
The applied weight factors were the defaults and the number of gaps to which the factors apply:

 1. Between systems: 6<br>1 gap (between the systems).
 2. Within systems, grouped by a [ bracket: 1.5<br>2 * 1 gap:
  between <code>Vln.</code> and <code>Vc.</code>
 3. Within systems, not grouped by a [ bracket: 3<br>2 * 4 gaps
  between <code>A. Sax.</code> and <code>Fl.</code>
  between <code>Fl.</code> and <code>Vln.</code>
  between <code>Vc.</code> and <code>Pno</code>
  between the two staves of the <code>Pno.</code>

Please note that the gap above the first staff of the system is not taken into account, this space remains unchanged.
At the bottom of the page, there is quite some space left between the lowest staff of the second system and the bottom music margin. The algorithm is designed to reduce this empty space by adding extra space between the gaps:
<code>"space left" * "weight factor for gap" / "sum of all weight factors on page"</code>

So, the between each <code>A. Sax.</code> and <code>Fl.</code>, <code>3/33 * "space left"</code>, so roughly 9% of the space left is added. Between the systems, 18% is added.

While adding this extra space, the maximum distance is obeyed. As a result, the algorithm can run in multiple passes. The algorithm stops:

 - When there is no space left,
 - All gaps reached their maximum distance or
 - After 10 passes (preventing an endless loop).

When calculating the <code>"sum of all weight factors on page"</code>, gaps which are already at their maximum space are ignored.

[...]

PR #6479, opened 2020-08-20
Automatic score ordering and bracketing

njvdberg:
This PR implements a few tasks of the [MuseScore Engraving Push](https://trello.com/b/JVGhCNm6/musescore-engraving-push):
 * [Implement default instrument ordering for newly created scores](https://trello.com/c/aqREuwol/3-implement-default-instrument-ordering-for-newly-created-scores)
  * [Implement default bracketing and bracing for newly created scores](https://trello.com/c/VoIML3kb/12-implement-default-bracketing-and-bracing-for-newly-created-scores)
  * [Create 'Instrument order' dropdown for the Instruments list](https://trello.com/c/gaQANcTa/8-create-instrument-order-dropdown-for-the-instruments-list)

The ordering uses to concept of `instrument families` and a score order defines the order of the families. `family`'s are defined just like `genre`'s and in the `<instrument>` section a family for a instrument is defined, e.g.:
```
<Family id="orchestral-strings">
    <name>Orchestral Strings</name>
</Family>
    .
    .
<Instrument id="violin">
    <family>orchestral-strings</family>
      .
      .
</Instrument>
```
A new file, `orders.xml` defines the available score orders. In the most simple form a score order just defines an order of families:
```
<Order id="an-example">
   <name>An Example</name>
   <family>clarinets<family>
   <family>flutes</family>
</Order>
```
An instrument belonging to the group `clarinets`, e.g. a Bass Clarinet, will be placed in the score above any instruments of the `flutes` family.

In this example, instruments which doesn't belong to the families `clarinets` or `flutes` will be placed in an `unsorted` family. If no `unsorted` family is defined an implicit `unsorted` family is defined.
```
<Order id="an-other-example">
   <name>An Other Example</name>
   <unsorted/>
   <family>clarinets<family>
   <family>flutes</family>
</Order>
```
Now all instruments, not being a member of `flutes` or `clarinets` will be place **before** the clarinets.
The `<unsorted>` tag has an optional attribute `group`.
```
<Order id="an-unsorted-example">
   <name>An Unsorted Example</name>
   <unsorted group="brass"/>
   <family>clarinets<family>
   <unsorted group="pitched-percussion"/>
   <family>flutes</family>
   <unsorted/>
</Order>
```
This places any unpitched percussion instrument between the `clarinets` and `flutes`.
Another special tag is `<soloists>`. Any instrument which marked as a soloist will go here.

Instrument families can be grouped in sections:
```
<Order id="bigband">
   <name>Big Band</name>
[...]

PR #6255, opened 2020-06-24
[MU3] Partial fix for #306911

Jojo-Schmitz:
See https://musescore.org/en/node/306911

This PRs fixes:

* clarify the "don't ask again"
* mention the current MuseScore version (that way the same dialog still
works in MuseScore 4)
* don't mark the score dirty if the reset is taken (it it is not taken, it is not marked dirty anyway)

It does not resolve the core issue of showing a Save dialog on close of any imported score if no changes have been done to the score (other than the reset).

Its counterpart for master had been merged meanwhile

PR #6791, opened 2020-10-29
[MU3] Engraving improvements dialog

vpereverzev:
Peek 2020-10-29 23-54

Design provided by @Tantacrul, although it's might be not a final version

This dialog will also include more handlers soon. For example: "Automatic score ordering and bracketing" - #6479 and "Algorithm for spreading staves/systems over a page" - #6538 (@njvdberg )

Also this PR includes the "First system indentation" handler (https://trello.com/c/kj5W4AQP)
Peek 2020-09-24 21-57

mtests failure is expected


libmscore/scorefile.cpp
```diff
@@ -253,7 +253,7 @@ void Score::writeMovement(XmlWriter& xml, bool selectionOnly)
       xml.etag();
 
       if (unhide)
-            endCmd(true);
+          endCmd(true);
```
Jojo-Schmitz, line 256: Was correct, for 3.x ;-)

libmscore/mscore.h
```diff
@@ -18,8 +18,8 @@
 
 namespace Ms {
 
-#define MSC_VERSION     "3.01"
-static constexpr int MSCVERSION = 301;
+#define MSC_VERSION     "3.02"
+static constexpr int MSCVERSION = 302;
```
Jojo-Schmitz, line 22: Dies this impose a compatibility issue for score created with this version when opening them with older 3.x versions?

Jojo-Schmitz:
No Leland yet?

Jojo-Schmitz:
The mtests don't like that `MSC_VERSION` change at all

libmscore/mscore.h
njvdberg: Yes, it does. And I'm not sure that's what we want.

libmscore/style.cpp
```diff
@@ -1323,7 +1326,7 @@ static const StyleType styleTypes[] {
       { Sid::tupletMinDistance,             "tupletMinDistance",             Spatium(0.5)  },
 
       { Sid::autoplaceEnabled,              "autoplaceEnabled",              true },
-
+      { Sid::qualityUpgradeAllowed,         "qualityUpgradeAllowed",         true }
```
njvdberg, line 1329: Is it necessary to make a style option <code>qualityUpgradeAllowed</code>. If it should be a style option, I think we need a kind of version identification since we might need such a option later for other upgrades in the future?

libmscore/layout.cpp
```diff
@@ -3727,13 +3727,11 @@ System* Score::collectSystem(LayoutContext& lc)
                   break;
                   }
             }
-      if (!allShown) {
```
njvdberg, line 3730: Will this change reintroduce issues #278057 and #277666? 
If not, also the few lines above should be removed (3724-3729).

libmscore/mscore.h
Jojo-Schmitz: I think we explicitly do **not** want this

njvdberg: My idea too. Also, in the 3.x series we already introduced new tag which will be lost when opening/saving the score in an older 3.x version. It seems inconsistent to change that behavior now.

Jojo-Schmitz: Loosing a tag when opening/saving with a downrev version is the lesser evil here IMHO

Jojo-Schmitz:
Well, along with the (still to be added) new fonts Leland, Leland Text and the Edwin family, we're gonna have backwards compatibility problem anyway I guess, so maybe we need to live with this then?

libmscore/layout.cpp
vshalkevich: Checked #278057 and #277666
No issues found

mscore/migration/iscoremigrationhandler.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "libmscore/score.h"
+
+class IScoreMigrationHandler
+{
+public:
+    IScoreMigrationHandler() = default;
+    virtual ~IScoreMigrationHandler() = default;
+
+    virtual bool handle(Ms::Score* score) = 0;
```
njvdberg, line 12: At this moment, the migration handlers have one argument only <code>Ms::Score*</code>. However, the Instrument Ordering handler will required another argument <code>Ms::MuseScore* mscore</code> (required to open the Instrument form which is part of the <code>MuseScore</code> class).

Do you prefer a new PR for this change or shall I add a remark at all files to be modified?

njvdberg (changes requested):

Jojo-Schmitz:
rebase needed (sorry, I think at least partly my #6255's fault)

vpereverzev:
> Well, along with the (still to be added) new fonts Leland, Leland Text and the Edwin family, we're gonna have backwards compatibility problem anyway I guess, so maybe we need to live with this then?

Yes, that was the main motivation for 3.02 introduction

mscore/migration/iscoremigrationhandler.h
vpereverzev: Looks like it's only needed for instrument handlers, so I suggest to use Ms::mscore directly, it's an external variable

All you need here is to include "mscore/musescore.h"

libmscore/style.cpp
vpereverzev: 3.6 will be the latest release of MU3, in MU4 we'll prepare another solution

libmscore/scorefile.cpp
vpereverzev: hah, fixed :)

Jojo-Schmitz:
That backwards compatibility issue is a very bad thing! And as explained in the Telegram chats, not needed at all, if you just change the internally coded defaults, i.e. change those to Leland and __not__ record that in the score.

libmscore/barline.cpp
```diff
@@ -475,7 +475,7 @@ void BarLine::drawDots(QPainter* painter, qreal x) const
             const StaffType* st = staffType();
 
             //workaround to make new Bravura and Petaluma font work correctly with repeatDots
```
Jojo-Schmitz, line 477: Comment needs to get adjusted too ;-)

libmscore/sym.cpp
```diff
@@ -6274,11 +6275,13 @@ void initScoreFonts()
...
-      QFont::insertSubstitution("ScoreFont",      "Bravura Text");
-      QFont::insertSubstitution("MuseJazz Text",  "Bravura Text");
-      QFont::insertSubstitution("Petaluma Text",  "Bravura Text");
+      QFont::insertSubstitution("Leland Text",    "Bravura Text");
+      QFont::insertSubstitution("Bravura Text",   "Leland Text");
+      QFont::insertSubstitution("MScore Text",    "Leland Text");
+      QFont::insertSubstitution("Gootville Text", "Leland Text");
+      QFont::insertSubstitution("ScoreFont",      "Leland Text");
+      QFont::insertSubstitution("MuseJazz Text",  "Leland Text");
+      QFont::insertSubstitution("Petaluma Text",  "Leland Text");
```
Jojo-Schmitz, line 6284: Pataluma should better fallback to MuseJazz (first, that then to Lelland and that to Bravura)

mscore/musescore.cpp
```diff
@@ -7646,7 +7646,7 @@ MuseScoreApplication::CommandLineParseResult MuseScoreApplication::parseCommandL
       parser.addOption(QCommandLineOption({"M", "midi-operations"}, "Specify MIDI import operations file", "file"));
       parser.addOption(QCommandLineOption({"w", "no-webview"}, "No web view in start center"));
       parser.addOption(QCommandLineOption({"P", "export-score-parts"}, "Use with '-o <file>.pdf', export score and parts"));
-      parser.addOption(QCommandLineOption(      "no-fallback-font", "Don't use Bravura as fallback musical font"));
+      parser.addOption(QCommandLineOption(      "no-fallback-font", "Don't use Leland as fallback musical font"));
```
Jojo-Schmitz, line 7649: Not sure what to make out of this, in connection with my proposal for Petaluma to fall back to MuseJazz first

vpereverzev:
> That backwards compatibility issue is a very bad thing! And as explained in the Telegram chats, not needed at all, if you just change the internally coded defaults, i.e. change those to Leland and **not** record that in the score.

I think this is not really an option, because it means that once the user changes the font from the standard Leland to, say, Bravura, they can never change it back and save again if we'll not write it in the score. Therefore, every time he opens his score, he will have to go into the styles dialog and temporarily return Leland :)

From my perspective, upgrading the format version is quite natural here, because we will greatly upgrade the engraving rules. It would be wrong if such drastic changes did not affect the format version. Perhaps @Tantacrul might add a few words here.

Jojo-Schmitz:
> I think this is not really an option, because it means that once the user changes the font from the standard Leland to, say, Bravura, they can never change it back and save again if we'll not write it in the score. Therefore, every time he opens his score, he will have to go into the styles dialog and temporarily return Leland :)

Huh? Bravura is not the default font and as such does get written into the score file. And is available in pre-3.6, so no issue there at all!
Only pre-3.6 scores using Emmentaler would 'automatically' (after allowing for this via that dialog) update to showing Leland. Or, if denied, store "Emmentaler" into the score. And 3.6 scores using Leland or (explicitly) Emmentaler, would just show Emmentaler when opened in any pre-3.6 version.

Jojo-Schmitz:
Or are there other and more things that "greatly upgrade the engraving rules" which I'm not aware of and that do affect compatibility? All the other I'm aware of, are just simplifying and/or automating things that are possible in pre-3.6 already, like instrument ordering and bracketing and indenting first system?

Jojo-Schmitz:
That dialog needs to get madse working for the dark theme!
foo

Jojo-Schmitz:
Also #6255 changed (or rather: was supposed to change)  that "Never ask this again" to the clearer "Remember my choice and don't ask again". In a different dialog (importing pre 3.x scores), but IMHO the same string makes sense here too.

njvdberg:
> Yes, that was the one of the main motivations for 3.02 introduction along with new engraving improvements, because it's a literally new version of the format anyway

For ScoreOrdering and VerticalStavesAdjustment this is not the case.
ScoreOrdering simply reshuffles the order and stores the ScoreOrder into the score file. When loading such a score in an older 3.x version of MuseScore, the score order itself remains but  the ScoreOrder definition tags are ignored.
For the VerticalStavesAdjustment, a several new tags are introduced for controlling the vertical spread. These will be ignored when load a score in an older version of MuseScore.

In the past, several new tags were introduced (Vertical Chord Aligment, Chord Play Back (?)) without the need of changing the file version number. I still have problems to see why it is really needed now.

vpereverzev:
> That dialog needs to get made working for the dark theme! Currently it leaves the user in the dark (pun intended):
> dialog
> Also #6255 changed (or rather: was supposed to change) that "Never ask this again" to the clearer "Remember my choice and don't ask again". In a different dialog (importing pre-3.x scores), but IMHO the same string makes sense here too.

THX, fixed

libmscore/sym.cpp
vpereverzev: @Tantacrul and @oktophonie approved that proposal, thx

fonts/leland/Leland.mss
```diff
@@ -1,18 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <museScore version="3.01">
```
Jojo-Schmitz, line 2: not `3.02`?

vpereverzev: yep, outdated

Jojo-Schmitz:
Leland and Edwin should get an 'honorable mention' in fonts/Readme.md

mtest/CMakeLists.txt
```diff
@@ -40,6 +40,8 @@ if (MSVC)
 endif (MSVC)
 
 QT5_ADD_RESOURCES(qrc_files ${PROJECT_SOURCE_DIR}/mtest/mtest.qrc
+            ${PROJECT_SOURCE_DIR}/mscore/musescorefonts-Leland.qrc
+            ${PROJECT_SOURCE_DIR}/mscore/musescorefonts-Edwin.qrc
```
Jojo-Schmitz, line 44: For some reason `${PROJECT_SOURCE_DIR}/mscore/musescorefonts-Campania.qrc` is missing in this file

mscore/editstyle.ui
```diff
@@ -303,6 +303,11 @@
                  <verstretch>0</verstretch>
                 </sizepolicy>
                </property>
+               <item>
+                <property name="text">
+                 <string notr="true">Leland</string>
+                </property>
+               </item>
                <item>
```
Jojo-Schmitz, line 311: "Leland Text" missing (in the next pull down) (even if it is just cosmetical)

libmscore/barline.cpp
```diff
@@ -474,8 +474,8 @@ void BarLine::drawDots(QPainter* painter, qreal x) const
       else {
             const StaffType* st = staffType();
 
-            //workaround to make new Bravura and Petaluma font work correctly with repeatDots
-            qreal offset = (score()->scoreFont()->name() == "Bravura" || score()->scoreFont()->name() == "Petaluma") ? 0 : 0.5 * score()->spatium() * mag();
+            //workaround to make new Bravura, Petaluma and Leland font work correctly with repeatDots
+            qreal offset = (score()->scoreFont()->name() == "Bravura" || score()->scoreFont()->name() == "Petaluma" || score()->scoreFont()->name() == "Leland") ? 0 : 0.5 * score()->spatium() * mag();
```
Jojo-Schmitz, line 478: Maybe we can change the order, Leland (the default) first, Pelatuma (the least likely) last? Or would that be 'micro-optimazation'?

Jojo-Schmitz:
Leland.mss needs to get installed into the styles folder (next to MuseJazz.mss), here in the GitHub repo only it doesn't help anybody ;-)

vpereverzev:
> Leland.mss needs to get installed into the styles folder (next to MuseJazz.mss), here in the GitHub repo only it doesn't help anybody ;-)

ah, very useful, TY
I've tried to find something related to Bravura

libmscore/barline.cpp
vpereverzev: yeah, why not

AntonioBL:
Ciao.
There is apparently a problem with vtest emmentaler-4, here is how it appears with this PR:
emmentaler-4-1

I didn't check, but are all "Emmentaler-*" vtests affected by this PR (not only their text, but also their musical symbols) just because they rely on Emmentaler being the (previous) default font (and thus they do not explicitly mention this font in the file structure)?

Additional note: isn't the percussion clef a little decentered? See for example vtest fmrest-1.
fmrest-1-1

Jojo-Schmitz:
I guess we'd need a whole new set of Leland tests, and make those Emmentaler test files explicitly using Emmentaler.
Or just rename them all to Leland* (and create a set of Emmentaler vtests, but maybe that is optional?)

mtest/CMakeLists.txt
Jojo-Schmitz: Thanks, solved ;-)

share/styles/Leland.mss
```diff
@@ -0,0 +1,1097 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Jojo-Schmitz, line 1: That's fine, but we now have a duplicate, that other in fonts/Leland?

mscore/editstyle.ui
Jojo-Schmitz: Thanks. solved

mscore/musescore.cpp
Jojo-Schmitz: Maybe just "Don't use a fallback musical font"?

Jojo-Schmitz: Or leave it Bravura, as even Leland falls back to Bravura in the end

oktophonie:
> Additional note: isn't the percussion clef a little decentered? See for example vtest fmrest-1.
Clefs are all left-justified, and SMuFL recommends setting zero sidebearings for most symbols, so that's why it looks like this (as it does in Bravura).
Emmentaler/MScore sets a wideish left sidebearing to quasi-centre the percussion clef. We might consider it as an option.

vpereverzev:
> I guess we'd need a whole new set of Leland vtests, and make those Emmentaler test files explicitly using Emmentaler.
> Or just rename them all to Leland* (and create a set of Emmentaler vtests, but maybe that is optional?)

Yes, that is exact plan for us

MarcSabatella:
The handling of clefs (up through 3.5) involves some special-casing in the code, in order to support the way clefs had previously been justified within the font.  I forget the details, but it will be good to review how clefs are positioned in all the different places they occur - beginning of score, mid-measure changes (especially if there are notes on other staves that might overlap), clef changes before barline, courtesy clefs, etc.

mscore/musescore.cpp
Jojo-Schmitz: See #6856

PR #6793, opened 2020-10-30
[MU4] Removed revision h file

igorkorsukov:
During development and during the release of release, the `revision.h`  file provides a lot of inconveniences.
We decided to get rid of it, replace it with the cmake option, which is set during build on CI.


Jojo-Schmitz:
Finally!

PR #6799, opened 2020-10-30
Fix #311340: Save online via command line (--save-online)

shoogle:
Resolves: https://musescore.org/en/node/311340

Adds `--save-online` option for command line:

```bash
mscore --save-online score1.mscz score2.mscx
```

Feature is useful for updating scores that are stored in a shared Git repository, or other location where they are editable by multiple people, but where only one person has permission to upload and publish to Musescore.com.

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -215,6 +215,7 @@ bool noWebView = false;
 bool exportScoreParts = false;
 bool saveScoreParts = false;
 bool ignoreWarnings = false;
+bool cliSaveOnline = false;
 bool exportScoreMedia = false;
 bool exportScoreMeta = false;
 bool exportScoreMp3 = false;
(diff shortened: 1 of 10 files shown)


Jojo-Schmitz:
A few coding style errors, nothing really serious

Jojo-Schmitz:
Wouldn't we need this for 3.x (3.6 actually) too?

mscore/musescore.cpp
```diff
@@ -8083,6 +8092,7 @@ MuseScoreApplication::CommandLineParseResult MuseScoreApplication::parseCommandL
     parser.addOption(QCommandLineOption({ "E", "install-extension" },
                                         "Install an extension, load soundfont as default unless -e is passed too",
                                         "extension file"));
+    parser.addOption(QCommandLineOption("save-online", "Upload score(s) to their source URL. Replaces existing online score(s)."));
```
Jojo-Schmitz, line 8095: Maybe add that this is not doing 1st time uploads, just updating existing ones and even that only for the same user?

Jojo-Schmitz: And please also update build/Linux+BSD/mscore.1.in

shoogle: Not sure there's room here, but I added a longer explanation in the manpage.

Jojo-Schmitz (approved):

Jojo-Schmitz:
Are you going to port this to 3.x too?

shoogle:
I updated the date. I will look at 3.x, but if it is very different then I won't bother. FYI, this seems to work for uploading 3.x scores.

Jojo-Schmitz:
"Save online" 3.x score with master builds may work now, but won't as soon as master changes to a newer file format, which I believe will happen sooner or later.

Jojo-Schmitz:
The difficult part of porting this from master to 3.x is the m4/cloud changes

PR #5616, opened 2020-01-16
Fix Deprecation warnings with Qt 5.15 (and 5.14)

Jojo-Schmitz:
at least those that are backwards compatible to at least Qt 5.9.
There are several hundred such (MSVC) warnings (557 C4996 in MSVC 2017, 680 in MSVC 2019, plus 5 'empty' ones) resp. 716 warnings (-Wdeprecated-declarations in MinGW, more than MSVC, due to being build in DEBUG mode), this PR fixes all but 5 (MSVC 2017), 16 (MSVC 2019), 17 (MinGW) of those.
The remaining few seem to require the use of replacements that are not properly documented,  possibly due to a Qt 5.15 documentation not being available yet (and/or don't work with Qt 5.9), at least I'm at a loss about what to do with them.

Here are the remaining ones (marked with ToDo comments for now):

Description |  File | Line 
 -- | -- | --
'QDesktopWidget::availableGeometry': Use QGuiApplication::screens() | framework\telemetry\widgets\telemetrypermissiondialog.cpp | 38 
'QJsonDocument::fromBinaryData': Use CBOR format instead | mscore\cloud\loginmanager.cpp | 83, 202 
'QJsonDocument::toBinaryData': Use CBOR format instead | mscore\cloud\loginmanager.cpp | 95, 187
'Qt::DefaultLocaleShortDate': Use QLocale | libmscore\page.cpp | 334, 342, 348, 354, 357, 362, 365
'QHelpEngineCore::linksForIdentifier': Use documentsForIdentifier() instead | mscore\help.cpp | 97, 143
'QButtonGroup::butonClicked': Use QButtonGroup::idClicked(int) instead | mscore\prefsdialog.cpp | 223
'QList<Ms::ElementType>::toSet': Use QSet<T>(list.begin(), list.end()) instead. | inspectors\models\inspectorlistmodel.cpp | 122

Hints how/whether to deal with those would be very welcome...

PR #6802, opened 2020-11-02
Fix #312442: Fix palette search shortcut functionality

worldwideweary:
Resolves: https://musescore.org/en/node/312442

From 3.5.1 to 3.5.2, there was a mishap for palette search functionality coming from fixing warnings for Qt 5.15 - https://github.com/musescore/MuseScore/pull/6627/

Update: As per conversation, this PR reverts the changes as mentioned in #6627 related to the function() keyword, but also in separate commit reforms the palette search functionality to make use of the function() keyword. 

Update: "While I'm up," I've decided to revisit palette search functionality since when the palette search widget is "floating", the keyboard shortcut doesn't work (wasn't a reversion... it just never worked appropriately). After trial/error, I learned that somehow the main paletteWidget "steals" focus after the focus upon the searchText in QML while floating, but it doesn't when docked. Adding a timer takes care of this. There may be a better way, but for now to have this actually working is what counts.  There's the additional problem that while floating, the ESCAPE key doesn't actually exit from the palette widget and resume focus back upon the score! These two problems I have resolved and am adding one additional/separate commit within this PR. 

- [ x ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [  ] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -4032,8 +4032,10 @@ bool MuseScoreApplication::event(QEvent* event)
 
 void MuseScore::focusScoreView()
       {
-      if (currentScoreView())
+      if (currentScoreView()) {
+            currentScoreView()->activateWindow();
             currentScoreView()->setFocus();
+            }
       else
             mscore->setFocus();
       }
--- a/mscore/palette/palettewidget.cpp
+++ b/mscore/palette/palettewidget.cpp
@@ -26,6 +26,7 @@
 #include "qml/nativetooltip.h"
 
 #include <QQmlContext>
+#include <QTimer>
 
 namespace Ms {
 
@@ -119,10 +120,13 @@ void PaletteWidget::setupStyle()
 
 void PaletteWidget::activateSearchBox()
       {
+      int delay = isFloating() ? 300 : 0;
       ensureQmlViewFocused();
-      const bool invoked = QMetaObject::invokeMethod(rootObject(), "requestPaletteSearch");
-      Q_UNUSED(invoked);
-      Q_ASSERT(invoked);
+      QTimer::singleShot(delay, this, [&](){
+          const bool invoked = QMetaObject::invokeMethod(rootObject(), "requestPaletteSearch");
+          Q_UNUSED(invoked);
+          Q_ASSERT(invoked);
+          });
       }
 
 //---------------------------------------------------------
--- a/mscore/palette/palettewidget.cpp
+++ b/mscore/palette/palettewidget.cpp
@@ -120,7 +120,9 @@ void PaletteWidget::setupStyle()
 void PaletteWidget::activateSearchBox()
       {
       ensureQmlViewFocused();
-      qmlInterface->requestPaletteSearch();
+      const bool invoked = QMetaObject::invokeMethod(rootObject(), "requestPaletteSearch");
+      Q_UNUSED(invoked);
+      Q_ASSERT(invoked);
       }
 
 //---------------------------------------------------------
--- a/mscore/qmldockwidget.cpp
+++ b/mscore/qmldockwidget.cpp
@@ -211,8 +211,10 @@ QQuickView* QmlDockWidget::getView()
 
 void QmlDockWidget::ensureQmlViewFocused()
       {
-      if (_view && !_view->activeFocusItem())
+      if (_view && !_view->activeFocusItem()) {
+            widget()->activateWindow();
             widget()->setFocus();
+            }
       }
 
 //---------------------------------------------------------
--- a/mscore/palette/palettewidget.h
+++ b/mscore/palette/palettewidget.h
@@ -69,8 +69,6 @@ class PaletteQmlInterface : public QObject
 
       void notifyElementDraggedToScoreView() { emit elementDraggedToScoreView(); }
 
-      void requestPaletteSearch() { emit paletteSearchRequested(); }
-
       Q_INVOKABLE Qt::KeyboardModifiers keyboardModifiers() const { return QGuiApplication::keyboardModifiers(); }
       };
 
--- a/mscore/qml/palettes/Palette.qml
+++ b/mscore/qml/palettes/Palette.qml
@@ -767,7 +767,7 @@ GridView {
                 // force not hiding palette cell if it is being dragged to a score
                 enabled: paletteCell.paletteDrag
                 target: mscore
-                function onElementDraggedToScoreView() { paletteCell.paletteDrag = false }
+                onElementDraggedToScoreView: paletteCell.paletteDrag = false
             }
         } // end ItemDelegate
     } // end DelegateModel
--- a/mscore/qml/palettes/PaletteTree.qml
+++ b/mscore/qml/palettes/PaletteTree.qml
@@ -772,7 +772,7 @@ ListView {
 
     Connections {
         target: palettesWidget
-        function onHasFocusChanged() {
+        onHasFocusChanged: {
             if (!palettesWidget.hasFocus) {
                 paletteSelectionModel.clearSelection();
                 expandedPopupIndex = null;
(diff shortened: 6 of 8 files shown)


Jojo-Schmitz (approved):

dmitrio95:
#6627 contains several such changes of signal handlers to `function`s in QML code. Should they also be reverted?

worldwideweary:
> #6627 contains several such changes of signal handlers to `function`s in QML code. Should they also be reverted?

I merely looked this up last night, but my guess is that those which have emit signals should be without the `function` keyword, but those that have `QMetaObject::invokeMethod()`  code  can have the function keyword without a problem (e.g. `applyCurrentPaletteElement()`).  

All the changes in #6627 have "emit" signals so they should probably be reverted to not using the function keyword, supposing there aren't any warnings or whatever. I haven't personally verified whether the code is actually occurring correctly or not for the dragging/focus changing. 

paletteWidget.h's void notifyElementDraggedToScoreView() { emit elementDraggedToScoreView(); }
onHasFocusChanged in paletteWidgetHeader.qml doesn't have a corresponding emission as I can find,

Jojo-Schmitz:
To be honest I don't even know why I changed those. Building now with these changes reverted, let's see...

Jojo-Schmitz:
I guess they came in here because I cherry-picked from #5616. But I don't remember why those changes made it into that one either... we may need to revert them there too?

Jojo-Schmitz:
Actually in b7b3564's commit message I stated: `Fix qml runtime warnings exactly following the advice given by that warning, but without really know(ing) what I'm doing here or whether that is still backwards compatible to Qt 5.9.`

Jojo-Schmitz:
I don't get any warnings in 3.x (and when builing that against Qt 5.15, in MinGW and MSVC)  when reverting those changes, so please go ahead and revert them

Jojo-Schmitz:
Whether or not to revert them in master too might still need to get determined...

Jojo-Schmitz:
Hmm, actually I do get warnings, at runtime and tons of them:
```
Warning: qrc:/qml/palettes/PaletteTree.qml:773:5: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... } (qrc:/qml/palettes/PaletteTree.qml:773, )
Warning: qrc:/qml/palettes/Palette.qml:766:13: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... } (qrc:/qml/palettes/Palette.qml:766, )
Warning: qrc:/qml/palettes/PalettesWidgetHeader.qml:186:5: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... } (qrc:/qml/palettes/PalettesWidgetHeader.qml:186, )
Warning: qrc:/qml/palettes/PalettesWidgetHeader.qml:174:5: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... } (qrc:/qml/palettes/PalettesWidgetHeader.qml:174, )
Warning: file:///C:/Users/Jojo/Documents/GitHub/MuseScore/win64install/qml/QtQuick/Controls/TableViewColumn.qml:66:1: QML TableViewColumn: Accessible must be attached to an Item (file:///C:/Users/Jojo/Documents/GitHub/MuseScore/win64install/qml/QtQuick/Controls/TableViewColumn.qml:66, )
Warning: qrc:/qml/palettes/PaletteTree.qml:516: TypeError: Cannot read property 'width' of null (qrc:/qml/palettes/PaletteTree.qml:516, )
Warning: qrc:/qml/palettes/PaletteTree.qml:353:19: QML ItemDelegate: Binding loop detected for property "width" (qrc:/qml/palettes/PaletteTree.qml:353, )
```

worldwideweary:
> Hmm, actually I do get warnings, at runtime and tons of them (here only one of each):

... Sounds like we're going to need to change something else with the way the code is structured or something in C++ since this isn't working with the simple `function ()` syntax. . .  since Qt 6 will probably no longer support it

dmitrio95:
Most probably `function()` syntax for signal handlers is not yet working in Qt 5.9, and those deprecation warnings are relevant only for newer versions of Qt. If so, we need to revert these changes in `3.x` branch but leave them in `master`.

Jojo-Schmitz:
Yes, exactly

worldwideweary:
> Yes, exactly

It looks like onElementDraggedToScoreView is failing also. To test this, drag a palette item (treble clef for instance) to a score view area without dropping it and notice that the palette area goes blank. 3.5.0 doesn't.

I've figured out a way to get this to be a function for the palette searching and have it work, but not sure if i should just leave it all alone in this reversion or actually update this to use the function() mechanism with appropriate code changes to have it work properly. It involves using similar code to `void PaletteWidget::applyCurrentPaletteElement()` and then redirecting a function in PaletteWidget.qml  into PaletteWidgetHeader.qml without using emit or the qml "Connections"

Jojo-Schmitz:
Either that or just revert all 4 of those

Jojo-Schmitz:
Either that or just revert all 4 of those, the easy way out. But fixing it your way may help esp. Linux distributions that use a new Qt version.

worldwideweary:
Ok. What I think I'll do is revert them all in this PR and update the title, but then also have a separate commit within this PR that "converts" the palette searching mechanism into `function()` form. That way if anyone in the future wants to convert the others, they may have a lead here as one possible solution. So long as everything is working fine for users, that's of main concern.

Jojo-Schmitz:
You could add those changes as a 3rd commit

worldwideweary:
> You could add those changes as a 3rd commit

Should be good to go.

Jojo-Schmitz:
I meant the convert to a proper `function()` form

Jojo-Schmitz:
Oh, I see. I though you meant to convert those other 4 too

worldwideweary:
I think there are only 4 in total, including the palette search. I've only "converted" the palette search here and "reverted" the other three.

Jojo-Schmitz:
Got it

mscore/palette/palettewidget.cpp
```diff
@@ -119,10 +120,13 @@ void PaletteWidget::setupStyle()
 
 void PaletteWidget::activateSearchBox()
       {
+      int delay = isFloating() ? 300 : 0;
```
worldwideweary, line 123: May seem strange, but when palette is floating, the focus in event ends up occuring "after" the focus in event of the searchText and ends up then removing the preparation for search string. Adding a slight delay to the request _only_ when floating clears this up

mscore/qmldockwidget.cpp
```diff
@@ -211,8 +211,10 @@ QQuickView* QmlDockWidget::getView()
 
 void QmlDockWidget::ensureQmlViewFocused()
       {
-      if (_view && !_view->activeFocusItem())
+      if (_view && !_view->activeFocusItem()) {
+            widget()->activateWindow();
```
worldwideweary, line 215: This ends up allowing using the search shortcut while floating. Unfortunately took quite some time for me to figure this out.

mscore/musescore.cpp
```diff
@@ -4032,8 +4032,10 @@ bool MuseScoreApplication::event(QEvent* event)
 
 void MuseScore::focusScoreView()
       {
-      if (currentScoreView())
+      if (currentScoreView()) {
+            currentScoreView()->activateWindow();
```
worldwideweary, line 4036: Likewise, this allows for pressing ESCAPE and from within the search string while floating and have the score resume focus. Unfortunately doesn't work without this PR/Commit.

worldwideweary:
Alright. Also got floating palettes working with search shortcut + escaping working. This wasn't working ever since like 3.2.3 when searching was first implemented I think it was.

PR #6804, opened 2020-11-02
[MU3] Removed revision h file

igorkorsukov:
Removed revision h file


.gitignore
```diff
@@ -27,7 +27,6 @@ DocLib
 build.qt5
 *.sfd~
 /.project
-mscore/revision.h
```
Jojo-Schmitz, line 30: It might even stay here, for those that still use that hook?

Jojo-Schmitz:
With this I guess we can finally also get rid if stull like thts in the release notes:`
'''
Important note!

Revision number (in mscore/revision.h) is incorrect in the assets automatically generated by GitHub. Use MuseScore-x.y.z.zip instead.
``

.gitignore
igorkorsukov: returned

Jojo-Schmitz:
The Mac build fail, `/bin/sh:  : command not found` seems to indicate some shell/environmane variable not expanding properly?

igorkorsukov:
> The Mac build fail, `/bin/sh: : command not found` seems to indicate some shell/environment variable not expanding properly?
> 
> Maybe that `$(FORMATTER_START)`?

There was waste `\` after `$(FORMATTER_END);  `
now all right for MacOS, builded and packed right.

PR #6806, opened 2020-11-03
[MU3] Fix #312513: Add Sicilian translation

Jojo-Schmitz:
also fix the instructions how to add a new language.

Resolves https://musescore.org/en/node/312513.

Does `git cherry-pick` cleanly to master (and is needed there too)


Jojo-Schmitz:
Does need that language to get enabled on Transifex though...

PR #6808, opened 2020-11-03
[MU4] Voices exchanging

RomanPudashkin:
- implemented voices exchanging

### DIFF
--- a/mu4/notation/internal/notationinteraction.cpp
+++ b/mu4/notation/internal/notationinteraction.cpp
@@ -1939,3 +1939,25 @@ void NotationInteraction::transpose(const TransposeOptions& options)
 
     m_notation->notifyAboutNotationChanged();
 }
+
+void NotationInteraction::swapVoices(int voiceIndex1, int voiceIndex2)
+{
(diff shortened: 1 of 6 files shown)


mu4/notation/internal/notationinteraction.cpp
```diff
@@ -1939,3 +1939,16 @@ void NotationInteraction::transpose(const TransposeOptions& options)
 
     m_notation->notifyAboutNotationChanged();
 }
+
+void NotationInteraction::swapVoices(int voiceIndex1, int voiceIndex2)
+{
+    if (voiceIndex1 == voiceIndex2) {
+        return;
+    }
```
igorkorsukov, line 1947: maybe add a check that voice is valid?

vpereverzev (approved):

mu4/notation/internal/notationinteraction.cpp
vpereverzev: is there any INVALID_VOICE_INDEX?

RomanPudashkin: fixed

PR #6811, opened 2020-11-04
Fix #299196  Fix #312393 - Missing space between key signature and first note if time signature is hidden

njvdberg:
Resolves: https://musescore.org/en/node/299196
Resolves: https://musescore.org/en/node/312393

When calculating the minimum measure width in <code>Measure::computeMinWidth()</code>, skip segments when all elements in that are invisible or have a zero width too.

### DIFF
--- a/libmscore/segment.cpp
+++ b/libmscore/segment.cpp
@@ -1063,6 +1063,24 @@ bool Segment::hasElements(int minTrack, int maxTrack) const
       }
 
 //---------------------------------------------------------
+//   allElementsInvisible
+///  return true if all elements in the segment are invisible
+//---------------------------------------------------------
+
+bool Segment::allElementsInvisible() const
+      {
+      if (isType(SegmentType::BarLineType | SegmentType::ChordRest | SegmentType::Breath))
+            return false;
+
+      for (Element* e : _elist) {
+            if (e && e->visible())
+                  return false;
+            }
+
+      return true;
+      }
+
+//---------------------------------------------------------
 //   hasAnnotationOrElement
 ///  return true if an annotation of type type or and element is found in the track range
 //---------------------------------------------------------
--- a/libmscore/segment.cpp
+++ b/libmscore/segment.cpp
@@ -1073,7 +1073,7 @@ bool Segment::allElementsInvisible() const
             return false;
 
       for (Element* e : _elist) {
-            if (e && e->visible())
+            if (e && e->visible() && !qFuzzyCompare(e->width(), 0.0))
                   return false;
             }
 
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -4283,12 +4283,14 @@ void Measure::computeMinWidth(Segment* s, qreal x, bool isSystemHeader)
 
       while (s) {
             s->rxpos() = x;
-            if (!s->enabled() || !s->visible()) {
+            if (!s->enabled() || !s->visible() || s->allElementsInvisible()) {
                   s->setWidth(0);
                   s = s->next();
                   continue;
                   }
             Segment* ns = s->nextActive();
+            while (ns && ns->allElementsInvisible())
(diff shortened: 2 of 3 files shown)


libmscore/segment.cpp
```diff
@@ -1062,6 +1062,20 @@ bool Segment::hasElements(int minTrack, int maxTrack) const
       return false;
       }
 
+//---------------------------------------------------------
+//   allElementsInvisible
+///  return true if all elements in the segment are invisible
+//---------------------------------------------------------
+
+bool Segment::allElementsInvisible() const
```
MarcSabatella, line 1070: Would there be a way to only do this once during layout, like at the same time we are marking the segment itself visible?  I guess maybe this will turn out to be only once, but still, it's an extra loop through the elements.

libmscore/segment.cpp
```diff
@@ -1062,6 +1062,20 @@ bool Segment::hasElements(int minTrack, int maxTrack) const
...
+      {
+      for (Element* e : _elist) {
+            if (e && e->visible())
```
MarcSabatella, line 1073: Without thinking this totally through or testing it, it seems this would cause invisible rests to now be ignored?  If so, this seems like it would break existing scores that rely on this to achieve particular spacing effects (eg, direct proportional spacing, equal width measures, etc).

Jojo-Schmitz:
I guess the vtest differences are OK
Before:
harmony-1-ref voice-2-ref
After
harmony-1-1 voice-2-1

MarcSabatella:
The chord symbol one is not OK.  That's probably another problem with simply skipping segments with no visible elements - it breaks the scheme that is used to allow chord symbols to space themselves out.  We could fix that by keeping the segment if it has annotations.  But maybe smarter is to just skip this whole business (always returning false for allElementsInvisible) for chordrest segments?

njvdberg:
Skipping <code>ChordRest</code> elements will solve both violations but I'm not sure whether using invisible notes to tune the layout is supported. If this is **not** supported, <code>allElementsInvisible</code> has to check for both invisible elements and  annotations. If using invisible notes for tuning the layout **is** supported (or we won't take the risk was used already),  <code>allElementsInvisible</code> can simply skip all <code>ChordRest</code>s.
It seems to me, for now the last option is the best. It solves an obvious issue and hasn't the risk breaking existing scores. If tuning the layout using invisible notes is not wanted, MU4 might be a better moment to do this.

Jojo-Schmitz:
>  I'm not sure whether using invisible notes to tune the layout is supported. 

It may not be officially supported, but it is certainly used in quite many scores.
@MarcSabatella remark from furter up:

> this seems like it would break existing scores that rely on this to achieve particular spacing effects (eg, direct proportional spacing, equal width measures, etc).

njvdberg:
Oops, I was that fixated on this comment I forgot the previous one :disappointed:. So skipping <code>ChordRest</code> (and all kind of barline segments) is the best option and that's what it will be.

njvdberg:
This PR now also solves [issue #312393 - Key signature change to C major adds extra space after the clef in error](https://musescore.org/en/node/312393)

Jojo-Schmitz:
vtest diff before:
accidental-10-ref
after:
accidental-10-1

MarcSabatella:
That to me is the correct/expected result.

Jojo-Schmitz:
Maybe the ref image should get updated then

njvdberg:
This is indeed expected result. When looking at the shapes you will see a zero width key signature:

Screenshot_20201105_165339

This PR now also skips segments with elements having all a zero width. There are still some other shapes visible which doesn;t to seem related to any element, but that's something for later.

To my knowledge, updating the reference is not necessary, vtest always runs against the previous commit. So, when this PR is merged, it becomes the reference (but correct my if I'm wrong).

Jojo-Schmitz:
The vtests on CI would indeed not be affected

PR #6818, opened 2020-11-05
[MU4] Porting 3.5.2_backend fixes to master

Jojo-Schmitz:


RomanPudashkin (approved):

libmscore/synthesizerstate.h
```diff
@@ -86,8 +86,14 @@ class SynthesizerState : public std::list<SynthesizerGroup>
...
+          {
+              0, "Zita1"
+          },
+#endif
+          {
+              2, "0.1"
+          },
           { 3, "440" },
           { 4, "1" },
           { 5, "1" }
```
Howard-C, line 99: Why only change the first two and not all others?

Jojo-Schmitz: Because the code style check only wanted those. IMHO bad enough...

Howard-C: This is the problem of introducing automatic check. Do we still get to vote for its expulsion from MuseScore? :)

Jojo-Schmitz: It comes with a price tag for sure...

PR #6826, opened 2020-11-05
fix #312608: stems too short on chords outside staff

MarcSabatella:
Resolves: https://musescore.org/en/node/312608

The stem shortening code (for notes outside the staff
whose stems point outwards) tries to enforce a minimum length.
However, this is applied too literally - it checks the full stem length.
So for *chords* as opposed to single notes, it often doesn't apply -
the portion of the the stem inside the chord already exceeds that.

Fix is to adjust the shortening code to check the "exposed" portion
of the stem only - the part extending further from the staff.
If *that* part is less than the "shortest stem" setting,
then we should extend the stem by the difference.

### DIFF
--- a/libmscore/chord.cpp
+++ b/libmscore/chord.cpp
@@ -1400,8 +1400,11 @@ qreal Chord::defaultStemLength() const
                   if (sel > staffHlfHgt)                    // if stem ends below ('>') staff mid position,
                         sel = staffHlfHgt;                  // stretch it to mid position
                   stemLen = sel - dy;                       // actual stem length
-                  if (-stemLen < shortest)                  // is stem too short,
-                        stemLen = -shortest;                // lengthen it to shortest possible length
+                  qreal exposedLen = sel - ul * .5;         // portion extending above top note of chord
+                  if (-exposedLen < shortest) {             // if stem too short,
+                        qreal diff = shortest + exposedLen;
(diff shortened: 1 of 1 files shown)


MarcSabatella:
vtest changes are good.  In 3.5.2, those stems were being shortened slightly in cases where they shouldn't have.  Easiest way to verify is to remove all but the outermost notes from each affected chord, and you'll see the stem lengthen.

vpereverzev:
@MarcSabatella could you rebase it on the latest 3.x so we can test it properly. I'm afraid these changes might not mix well with the latest fixes in 3.x

MarcSabatella:
Rebased.  Offhand I don't see any reason it shouldn't still work the same.  I guess maybe you are thinking of changes made regarding tweaking stem lengths for rounded ends or to get the right exact ending point relative to a flag?

MarcSabatella:
These appear to be the same expected vtest "failures" as before the rebase.  Yes, some stems outside the staff are a little longer now, that's the point.

MarcSabatella:
The mtest failure is due to a Spatium different in one of the MusicXML tests.  I didn't touch that here, and had the same failure in PR #7026, so I'm kind of thinking it's just a broken test that needs fixing after the spatium default change recently.  Unless somehow my rebase is missing something?  I can't really see how, though.

Jojo-Schmitz:
vtests fail in a rather strange way now

PR #6831, opened 2020-11-06
Zita Reverb. Prevent NaN in params.

psmokotnin:
Zita reverb could receive params out of float type range.

inhouse team.


BernardGreenberg:
Is it known that this actually fixes the "no more reverb in uploaded scores" problem?  If so, why did it just manifest now. Thanks.

PR #6832, opened 2020-11-06
Zita. Prevent NaN in params.

psmokotnin:
Zita reverb could receive params out of float type range.

inhouse team.


Jojo-Schmitz:
It seems some elements of `static const std::vector<ParDescr> pd` could use an `f` suffix.

PR #6836, opened 2020-11-07
[MU4] Elements select

Eism:
- Transferred select note dialog to MU4
- Transferred select element dialog to MU4

Peek 2020-11-07 16-06


mu4/notation/inotationelements.h
```diff
@@ -18,6 +19,7 @@ class INotationElements
     virtual ~INotationElements() = default;
 
     virtual Ms::Element* search(const std::string& searchCommand) const = 0;
```
vpereverzev, line 21: From that point it's not really obvious which "select" should be used. Is there any chance to merge these two methods and leave a single way to search?

igorkorsukov:
needs rebase

mu4/notation/inotationelements.h
RomanPudashkin: I think, better to rename this method to something like "elements" and also rename SearchElementOptions to FilterElementsOptions. By default, this method should return all score elements

RomanPudashkin, line 21: Please, rename searchCommand to searchText

mu4/notation/inotationinteraction.h
```diff
@@ -53,7 +53,8 @@ class INotationInteraction
     // Select
     virtual Element* hitElement(const QPointF& pos, float width) const = 0;
     virtual int hitStaffIndex(const QPointF& pos) const = 0;
-    virtual void select(Element* element, SelectType type, int staffIndex = 0) = 0;
+    virtual void select(Element* e, SelectType type, int staffIdx = 0) = 0;
+    virtual void select(std::vector<Element*> elements, SelectType type, int staffIndex = 0) = 0;
```
RomanPudashkin, line 57: Better to unite select(Element* e, ...) and select(std::vector<Element*> elements, ...)

mu4/notation/inotationelements.h
Eism: fixed. Thanks!

RomanPudashkin (approved):

mu4/notation/internal/notationactioncontroller.cpp
```diff
@@ -478,6 +570,23 @@ void NotationActionController::openPartsDialog()
     interactive()->open("musescore://notation/parts");
 }
 
+FilterElementsOptions NotationActionController::elementsFilterOptions(const Element* element) const
+{
+    FilterElementsOptions options;
+    options.elementType = element->type();
+
+    if (element->type() == ElementType::NOTE) {
+        const Ms::Note* note = dynamic_cast<const Ms::Note*>(element);
```
igorkorsukov, line 579: Better use method `toNote`, it have extra checks

igorkorsukov (approved):

PR #6837, opened 2020-11-07
fix #307271 : [Musicxml Export] - Wrong tuplet calculation in two-note tremolo

AntonioBL:
Resolves: https://musescore.org/en/node/307271

This fixes a bug in tuplet import (tick was not set, so when deleting a tuplet in a file imported from xml there was corruption) and in xml import of a two-note tremolo at the beginning of a tuplet. It adds the optional tuplet-actual and tuples-normal fields in MusicXML export for compatibility with some Braille exporting softwares.


AntonioBL:
The case of nested tuplet import is not dealt with, since import of MusicXML nested tuplets is not implemented yet. (But export works ok)

Jojo-Schmitz:
Are you planing to port this to the 3.x branch too?

PR #6838, opened 2020-11-08
fix #259771 : make tremolo between notes play correct rhythm when bet…

AntonioBL:
…ween tuplets

Resolves: https://musescore.org/en/node/259771

By using tick instead of actual ticks, the number of note divisions (tticks/t) is correctly computed (while if actualTicks are used in tticks, the number of ticks per division -t- must be corrected by the tuplet ratio).
Single note tremolos already use ticks instead of actualTicks.


Jojo-Schmitz:
Are you planing to port this to the 3.x branch too?

PR #6839, opened 2020-11-09
Speed up MSVC builds and tidy msvc_build.bat script

shoogle:
Speed improvements:

- CMake configuration is only done on the first build.

- Avoid building if configuration was unsuccessful (fail fast).

- Preserve %ERRORLEVEL% on exit (avoid running subsequent commands).

Other improvements:

- Remove check for "nul" directory as this doesn't work on modern
  Windows. It is now possible to check for the directory itself.

- Use K&R braces/parentheses everywhere (i.e. not banner style).

- Strip trailing whitespace.


Jojo-Schmitz:
1. Are you planning to port this to the 3.x branch too? Might be a simple `git cherry-pick`
2. What in this context is a "modern Windows"? Isn't it more a question of the file system uses, NTFS vs FAT?

Jojo-Schmitz (approved):

PR #6841, opened 2020-11-09
[MU3] Fix warnings reported by Clang-Tidy and Clazy

Jojo-Schmitz:
Clang-Tidy and Clazy are available as part of QtCreator.
There are lots and lots of warnings reported by them.
This PR fixes those with obvious changes, those where the reason is understood and a good fix is known.
Lots more are still there and not yet understood nor fully investigated, quite a few looking pretty terrifying.


mscore/plugin/api/cursor.h
```diff
@@ -69,24 +69,24 @@ class Cursor : public QObject {
       Q_PROPERTY(int filter     READ filter    WRITE setFilter)
 
       /** MIDI tick position, read only */
-      Q_PROPERTY(int tick         READ tick) // FIXME: fraction transition
+      Q_PROPERTY(int tick         READ tick CONSTANT) // FIXME: fraction transition
```
dmitrio95, line 72: Most of properties in plugin API are not constant, it would be incorrect to add `CONSTANT` attribute to all of them.

Jojo-Schmitz: Hmm, wouldn't all then don't have a WRITE method be CONSTANT?

Jojo-Schmitz: Which would, which would need a NOTIFY method?

Jojo-Schmitz: Need to revert them all?

Jojo-Schmitz:
Any idea why it crashes at the tst_runscripts mtest?

mscore/plugin/api/cursor.h
dmitrio95: According to the documentation [here](https://doc.qt.io/qt-5/properties.html), `CONSTANT` attribute means that reading this property for the same object will always return the same value. As most of objects in plugin API just expose properties of the underlying libmscore objects, and those objects may change its state while score is being edited (by user or by plugin), such properties are not `CONSTANT` in the sense defined in the documentation.

Having `NOTIFY` signals for those properties would be ideal, but that would require adding a mechanism that would allow objects in libmscore report their properties changes. That would be a non-trivial and potentially resource-consuming change (as some properties like positions of elements change for many objects after every layout), and the benefit of this is unclear. `NOTIFY` signals would enable those properties being used in property bindings, but this seems to be not quite useful for plugin API objects as those are not normally used as usual QML components but are rather being handled in JavaScript code of functions or signal handlers.

For those objects that are intended to be used in QML markup (like [`PluginAPI`](https://github.com/musescore/MuseScore/blob/3.x/mscore/plugin/api/qmlpluginapi.h) object which is a root component of any plugin or objects from [`util`](https://github.com/musescore/MuseScore/blob/3.x/mscore/plugin/api/util.h) part of the API) it would indeed be good to mark those properties that do not change as `CONSTANT` and possibly add `NOTIFY` signals for other properties. However I doubt that adding those notifier signals to objects that cannot be created directly in QML code worth the effort.

mscore/scoretab.h
```diff
@@ -71,7 +71,7 @@ class ScoreTab : public QWidget {
       const TabScoreView* tabScoreView(int idx) const;
 
    signals:
-      void currentScoreViewChanged(ScoreView*);
+      void currentScoreViewChanged(Ms::ScoreView*);
```
dmitrio95, line 74: > Any idea why it crashes at the tst_runscripts mtest?

It looks like all `connect()` calls that use the old string-based connection syntax (with `SIGNAL()` and `SLOT()` macros) fail as they are unable to find signals by their old names. For example, for this signal there is the following warning in mtest log:
```
unknown:unknown: QObject::connect: No such signal Ms::ScoreTab::currentScoreViewChanged(ScoreView*) in /home/runner/work/MuseScore/MuseScore/mscore/musescore.cpp:6242
```

This may be the reason for failing script tests, as the log contains a lot of warnings which may result from signal-slot connections not working as expected.

Possible fixes are:
1) Move all `connect()` statements to the new syntax based on function pointers (https://wiki.qt.io/New_Signal_Slot_Syntax);
2) Change names of signals and slots in `connect()` statements to match these changes;
3) Revert these changes (but then the warnings would return back?)

mscore/plugin/api/cursor.h
Jojo-Schmitz: OK, guess I'll revert that part of the changes then

mscore/scoretab.h
Jojo-Schmitz: I'll revert those too, for now

Jojo-Schmitz:
I'm pretty sure the mtest failure is not my (PR's) fault, got to be the one just before mine (from PR #6776), @vpereverzev ?

vpereverzev:
looking forward for rebase

vpereverzev (approved):

Jojo-Schmitz:
> looking forward for rebase

On my todo list for today

Jojo-Schmitz:
done, will run another analysis.

Jojo-Schmitz:
ran another round (takes more than ione hour), added a few more changes.
There are still almost 800 diagnostics left...

Jojo-Schmitz:
Let me know whether you'd want a similar one for the master branch

PR #6847, opened 2020-11-10
[MU3] Fix #312254: Add alternative choice of “To Coda”

Jojo-Schmitz:
Add a "To `<sym>coda</sym>`" to master "Repeats and Jumps" palette .
The "To " part of it is sort of needed to tell it from the plain Coda in the palettes (so to not have to rely on the tool tip only), but is likely to not be wanted in score, so needs to get deleted there manually.
Seems we need to hardcode the size of the Coda symbol.

Removing some rather rare symbols from the Basis workspace's "Repeats
and Jumps" palette, namely "Segno variation" and "Varied coda".

Resolves https://musescore.org/en/node/312254

Port to MuseScore 4 needed too I guess, I'll take care of it once this PR passed muster ;-)

### DIFF
--- a/libmscore/marker.cpp
+++ b/libmscore/marker.cpp
@@ -29,14 +29,15 @@ static const ElementStyle markerStyle {
 
 //must be in sync with Marker::Type enum
 const MarkerTypeItem markerTypeTable[] = {
-      { Marker::Type::SEGNO   , QT_TRANSLATE_NOOP("markerType", "Segno")          },
-      { Marker::Type::VARSEGNO, QT_TRANSLATE_NOOP("markerType", "Segno variation")},
-      { Marker::Type::CODA    , QT_TRANSLATE_NOOP("markerType", "Coda")           },
-      { Marker::Type::VARCODA , QT_TRANSLATE_NOOP("markerType", "Varied coda")    },
-      { Marker::Type::CODETTA , QT_TRANSLATE_NOOP("markerType", "Codetta")        },
-      { Marker::Type::FINE    , QT_TRANSLATE_NOOP("markerType", "Fine")           },
-      { Marker::Type::TOCODA  , QT_TRANSLATE_NOOP("markerType", "To Coda")        },
-      { Marker::Type::USER    , QT_TRANSLATE_NOOP("markerType", "Custom")         }
+      { Marker::Type::SEGNO    , QT_TRANSLATE_NOOP("markerType", "Segno")           },
+      { Marker::Type::VARSEGNO , QT_TRANSLATE_NOOP("markerType", "Segno variation") },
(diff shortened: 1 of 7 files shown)


Jojo-Schmitz:
mtest fails in tst_palette, but without any decent error message, so I'm pretty much in the dark here. Probably due to using the identical tooltip.

libmscore/marker.cpp
```diff
@@ -107,6 +108,12 @@ void Marker::setMarkerType(Type t)
                   setLabel("coda");
                   break;
 
+            case Type::TOCODASYM:
+                  txt = "To <font size=\"20\"/><sym>coda</sym>";
```
Jojo-Schmitz, line 112: Unfortunately we have to hard code the size of that symbol (here to the default size of the 'plain' Coda)

share/workspaces/Advanced.xml
```diff
@@ -2211,6 +2211,13 @@
             <label>coda</label>
             </Marker>
           </Cell>
+        <Cell name="To Coda">
+          <Marker>
+            <style>Repeat Text Left</style>
+            <text>To <font size="20"/><sym>coda</sym></text>
```
Jojo-Schmitz, line 2217: Unfortunately we have to hard code the size of that symbol (here to the default size of the 'plain' Coda, 20 pt), it'd be way too small otherwise (12 pt per default).

libmscore/marker.cpp
```diff
@@ -29,15 +29,15 @@ static const ElementStyle markerStyle {
...
-      { Marker::Type::TOCODASYM, QT_TRANSLATE_NOOP("markerType", "To Coda")        },
-      { Marker::Type::USER     , QT_TRANSLATE_NOOP("markerType", "Custom")         }
+      { Marker::Type::SEGNO    , QT_TRANSLATE_NOOP("markerType", "Segno")           },
+      { Marker::Type::VARSEGNO , QT_TRANSLATE_NOOP("markerType", "Segno variation") },
+      { Marker::Type::CODA     , QT_TRANSLATE_NOOP("markerType", "Coda")            },
+      { Marker::Type::VARCODA  , QT_TRANSLATE_NOOP("markerType", "Varied coda")     },
+      { Marker::Type::CODETTA  , QT_TRANSLATE_NOOP("markerType", "Codetta")         },
+      { Marker::Type::FINE     , QT_TRANSLATE_NOOP("markerType", "Fine")            },
+      { Marker::Type::TOCODA   , QT_TRANSLATE_NOOP("markerType", "To Coda")         },
+      { Marker::Type::TOCODASYM, QT_TRANSLATE_NOOP("markerType", "To Coda (Symbol)")},
```
Jojo-Schmitz, line 39: To be able to distinguish the 2 Codas, esp. for visually impaired users (and to appease the tst_palette mtest)

PR #6852, opened 2020-11-12
Implement adjustable ease-in and ease-out interpolation curve for all glissando styles.

ypoissant:
Resolves: [Portamento interpolation curve](https://github.com/musescore/MuseScore/pull/4824#issuecomment-613501871)

A portamento with ease-in and ease-out feels more natural than a linearly interpolated portamento. The easeInOut class allows to adjust the ease-in and ease-out from a linear interpolation to an exponential-like or inverse-exponential-like, to an S curve. This is achieved by using a Bezier curve where the second and third control points are adjusted according to two sliders.


libmscore/EaseInOut.h
```diff
@@ -0,0 +1,50 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2010-2011 Werner Schweer
```
Jojo-Schmitz, line 5: `-2020 MuseScore BVBA` please

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2002-2012 Werner Schweer
```
Jojo-Schmitz, line 5: `-2020 MuseScore BVBA` please

libmscore/property.cpp
```diff
@@ -178,10 +178,13 @@ static constexpr PropertyMetaData propertyList[] = {
...
       { Pid::MEASURE_NUMBER_MODE,     false, "measureNumberMode",     P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "measure number mode") },
+
       { Pid::GLISS_TYPE,              false, "subtype",               P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "subtype")          },
       { Pid::GLISS_TEXT,              false, 0,                       P_TYPE::STRING,              DUMMY_QT_TRANSLATE_NOOP("propertyName", "text")             },
-
       { Pid::GLISS_SHOW_TEXT,         false, 0,                       P_TYPE::BOOL,                DUMMY_QT_TRANSLATE_NOOP("propertyName", "showing text")     },
+      { Pid::GLISS_STYLE,             true,  "glissandoStyle",        P_TYPE::GLISS_STYLE,         DUMMY_QT_TRANSLATE_NOOP("propertyName", "glissando style") },
+      { Pid::GLISS_EASEIN,            false, "easeInSpin",          P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "ease in")          },
+      { Pid::GLISS_EASEOUT,           false, "easeOutSpin",         P_TYPE::INT,                 DUMMY_QT_TRANSLATE_NOOP("propertyName", "ease out")         },
```
Jojo-Schmitz, line 187: formatting

Jojo-Schmitz:
No idea why the linking fails on Linux (build, mtest, vtest) but work on Windows and Mac??

libmscore/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ add_library (
 
       types.h accidental.h ambitus.h arpeggio.h articulation.h audio.h bagpembell.h barline.h beam.h bend.h
       box.h bracket.h bracketItem.h breath.h bsp.h bsymbol.h changeMap.h chord.h chordline.h chordlist.h chordrest.h clef.h
-      cleflist.h connector.h drumset.h dsp.h duration.h durationtype.h dynamic.h element.h
+      cleflist.h connector.h drumset.h dsp.h duration.h durationtype.h dynamic.h easeInOut.h element.h
```
Jojo-Schmitz, line 48: Ah!!! It needs to be `EaseInoOut.h", uppercsase `E`, case sensitive (not sure though why it works on Mac, this has a case sentitife file system too)
Or better rename the files to start with a lower case`e`

libmscore/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ add_library (
       bracket.cpp breath.cpp bsp.cpp changeMap.cpp chord.cpp chordline.cpp
       chordlist.cpp chordrest.cpp clef.cpp cleflist.cpp
       drumset.cpp durationtype.cpp dynamic.cpp dynamichairpingroup.cpp edit.cpp noteentry.cpp
-      element.cpp elementgroup.cpp excerpt.cpp
+      easeInOut.cpp element.cpp elementgroup.cpp excerpt.cpp
```
Jojo-Schmitz, line 66: See above

libmscore/EaseInOut.h
ypoissant: Done. Thanks

libmscore/property.cpp
ypoissant: Done. Thanks

libmscore/easeInOut.cpp
ypoissant: Done. Thanks

libmscore/CMakeLists.txt
ypoissant: Done. I decided on renaming the file with a lowercase 'e'. Thanks

libmscore/CMakeLists.txt
ypoissant: Done. I decided on renaming the file with a lowercase 'e'

Jojo-Schmitz: Nope, not yet, just for the .h so far ;-)

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+    } else if (discr > 0.0) {
+        // Case that happens about 25% of the time --> 1 root.
+        qreal q2 = q / 2.0;
+        qreal sd = std::sqrt(discr);
+        qreal u = std::pow(-q2 + sd, 1.0 / 3.0);
+        qreal v = std::pow(q2 + sd, 1.0 / 3.0);
+        t = u - v - a / 3.0;
+    } else {
```
njvdberg, line 69: This is an implied float <code>==</code> operations which [generally doesn't make sense](https://floating-point-gui.de/errors/comparison/). Better to use something like:
<code>
if (qFuzzyCompare(discr, 0.0) {

    // Case that happens extremely rarely --> 2 roots.

    ...

} else if  (discr < 0.0) {

    // Case that happens about 75% of the time --> 3 roots.

    ...

} else {

    // Case that happens about 25% of the time --> 1 root.

    ...

}

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+//-------------------------------------------------------------------------------------------------
+// This function is even more simplified and optimized because all the Bezier control points are
+// constant. Thus, when computing the Y root there is only one case and the math simplifies to the
+// following simple expression.
+//-------------------------------------------------------------------------------------------------
+qreal EaseInOut::tFromY(const qreal y) const
+{
+      if (y == 0.0 || y == 1.0)
```
njvdberg, line 89: Same here, don't use floating <code>==</code> but use <code>qFuzzyCompare</code> instead.

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+void EaseInOut::timeList(const int nbNotes, const int duration, QList<int>* times) const
+{
+      qreal nNotes = qreal(nbNotes);
+      qreal space = qreal(duration);
+      if (_easeIn == 0.0 && _easeOut == 0.0) {
+            for (qreal n = 0; n <= nNotes; n++) {
+                  times->push_back(int((n / nNotes) * space));
+                  }
+      } else {
+            for (qreal n = 0; n <= nNotes; n++) {
```
njvdberg, line 107: Better to keep the loop variable integer.

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+//-------------------------------------------------------------------------------------------------
+// Given a number of note to place within the given duration, return the list of on-times for each
+// note given the current ease-in and ease-out parameters. The first note is at time 0 while the 
+//-------------------------------------------------------------------------------------------------
+void EaseInOut::timeList(const int nbNotes, const int duration, QList<int>* times) const
+{
+      qreal nNotes = qreal(nbNotes);
+      qreal space = qreal(duration);
+      if (_easeIn == 0.0 && _easeOut == 0.0) {
+            for (qreal n = 0; n <= nNotes; n++) {
```
njvdberg, line 103: Better to keep the loop variable integer.

libmscore/CMakeLists.txt
Jojo-Schmitz: I think you'd still need to rename the file `EaseInOut.cpp` to `easeInOut.cpp`

njvdberg (changes requested):

libmscore/easeInOut.cpp
Jojo-Schmitz: Didn't even know that a loop counter doesn't __have__ to be integer

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+      return 0.5 + std::cos((4.0 * M_PI + std::acos(1.0 - 2.0 * y)) / 3.0);
+      }
+
+//-------------------------------------------------------------------------------------------------
+// Given a number of note to place within the given duration, return the list of on-times for each
+// note given the current ease-in and ease-out parameters. The first note is at time 0 while the 
+//-------------------------------------------------------------------------------------------------
+void EaseInOut::timeList(const int nbNotes, const int duration, QList<int>* times) const
+{
+      qreal nNotes = qreal(nbNotes);
```
Jojo-Schmitz, line 100: why not leaving it an int?

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+      qreal space = qreal(duration);
+      if (_easeIn == 0.0 && _easeOut == 0.0) {
+            for (qreal n = 0; n <= nNotes; n++) {
+                  times->push_back(int((n / nNotes) * space));
```
Jojo-Schmitz, line 104: just to cast it back into an int here?

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,116 @@
...
+                  }
+      } else {
+            for (qreal n = 0; n <= nNotes; n++) {
+                  //times->push_back(int(XfromY(n / nNotes) * space));
+                  qreal y = n / nNotes;
```
Jojo-Schmitz, line 109: here though at least one of the operands needs to be (casted into) a float

libmscore/CMakeLists.txt
Jojo-Schmitz: And the Linux build errors prove me right it seems

Jojo-Schmitz: Hmm, no, mtest and vtest failed for different reasons apparently

Jojo-Schmitz: It just crashes all over the place

libmscore/easeInOut.cpp
ypoissant: Thanks for the comment. I analyzed the cases and the only place where discr is actually used is where discr < 0.0. Then discr is square rooted, which in the worst case will result in sd being denormalized or zeroed. Then sd is added to q2 and -q2. This is not going to result in math error. In the two other cases, discr is not used any further. Beside, the tests are against zero, which in floating point representation is precise. Not against an arbitrary value, which can suffer from rounding errors. So because this function needs to be quick rather than precise, I will leave the simple float compares there.

libmscore/easeInOut.cpp
ypoissant: Thanks for the comment. 'n' is used, inside the loop to compute a ratio. Using an int there would require that n be cast to qreal before the ratio calculation or result in quatization errors when nNotes is small. I agree that loop counters are better as int but in this case, it is not used as a counter. So I'll leave it this way

libmscore/easeInOut.cpp
ypoissant: The cast back to int is for the ratio times space. The ratio must be float to work.

libmscore/easeInOut.cpp
ypoissant: Oups!. That one is a debugging leftover. I'll fix it. Thanks.

libmscore/easeInOut.cpp
ypoissant: See reply below.

libmscore/easeInOut.cpp
ypoissant: I actually made a change to 'for (qreal n = 0.0; n <= nNotes; n += 1.0) {' to make the intention more specific.

libmscore/easeInOut.cpp
ypoissant: 0.0 and 1.0 are represented precisely in float format. The trouble is when there is a fractional part on the number. In this case, anyway, those are precise end conditions. Any other y values will require a Bezier interpolation.

Jojo-Schmitz: Maybe, but equality is not really. Is 0.0000000001 == 0.0?

ypoissant: I don't understand. 0.00000000000001 is not equal to 0.0. Do you suggest that I truncate small values to zero?

Jojo-Schmitz: yes

Jojo-Schmitz: That is what `qFuzzyCompare()`does

ypoissant:
I don't know how to handle the errors to vtests and mtests. Any hint would be appreciated. Thanks.

libmscore/easeInOut.cpp
njvdberg: This issue is not whether <code>discr</code> is used or not but the whether the software will enter the correct branch. Most likely, you'll never end-up in the <code>else</code> branch (which is actually the <code>==</code> branch).

Also, even a float comparison against is not save if at least one of the two values is the result of a calculation. E.g.
<code>((2.3 - 1.0)/2.0 - 0.65) == 0.0</code> will return <code>false</code>! On my computer, this evaluates to <code>-1.11022e-16 == 0.0</code>.
It it really a bad idea to use a float comparison.

njvdberg (changes requested):

libmscore/easeInOut.cpp
njvdberg: When using float (or double) calculation, the result is almost always rounded because the the actual result doesn't "fit" in the float/double format. So, rounding errors of +/- 1 bit are most likely.
<code>qFuzzCompare</code> indeed tries to take this "one bit errors" into account. See [here](https://floating-point-gui.de/errors/comparison/) for a nice explanation.

libmscore/easeInOut.cpp
njvdberg: You might be lucky here but if the step is 1.3 and <code>nNotes</code> is 7.8 you might have one iteration too little because if cumulative errors in the add operations. The common way to handle is to have an integer loop variable and calculate the real value inside the loop. E.g.
```
for (int i = 0; i < nNotes; ++i) {
     real n = static_cast<qreal>(i);
     ...
}
````

Also, it might be my Fortran background (oops!), but a variable <code>n</code> is for me an integer, not a float :wink: .

File: libmscore/easeInOut.cpp (hunk 8 above)
Comment (njvdberg) [reply libmscore/easeInOut.cpp:107]: To prevent cumulative errors, it is still better to use an integer loop variable and cast the integer to a float once you need it in an expression.

File: libmscore/easeInOut.cpp (hunk 11 above)
Comment (njvdberg) [reply libmscore/easeInOut.cpp:104]: A better solution is having <code>n</code> an integer loop variable and use
```
times->push_back(static_cast<int>(static_cast<qreal>(n) / nNotes) * space));
```

**Comment by Jojo-Schmitz** [comment]
> I don't know how to handle the errors to vtests and mtests. Any hint would be appreciated. Thanks.

Here's one hint at where it crashes:
/home/runner/work/MuseScore/MuseScore/libmscore/style.cpp:unknown: ASSERT: "t.idx() == idx" in file /home/runner/work/MuseScore/MuseScore/libmscore/style.cpp, line 2889

File: libmscore/style.h [hunk 13]
```diff
@@ -968,6 +968,8 @@ enum class Sid {
       glissandoFrameBgColor,
       glissandoLineWidth,
       glissandoText,
+      glissandoEaseIn,
+      glissandoEaseOut,
```
Jojo-Schmitz, line 972: You added 2 here, but not in style.cpp, this causes the crash resp. assertion failure

Jojo-Schmitz: Delete those 2, you don't use them anywhere

ypoissant: I see. I added those two styles in the beginning of my exploration but they caused the "apply as style" button to appear in the inspector so I removed it. But obviously, I did not remove it everywhere. Thanks for pointing to it. I'll fix that.

ypoissant:
@njvdberg I agree with the principle that doing comparison on float calculations is not optimal precisionwise. And indeed, the final else condition will most probably never be fired (thus the "extremely rarely" comment). This case is only there to make sure that if this condition ever happen, the function will not crash.

Personally I'm OK with the fact that with close enough to zero, the execution may take one of the two other branches and the computed t will still be close enough to the true value given the application for which the function is used, which is computing warped MIDI event times that, in the end, will be quantized to values with immensely less precision than the possible float errors that this may introduce. 

This function is called several times for a single portamento note. I expect to use it for bends too. So I took the approach of prefering execution speed against precision. qFuzzyCompare is quite expensive if you look at the implementation.

This said, I will use qFuzzyCompare if this is a MuseScore development requirement.

libmscore/easeInOut.cpp
ypoissant: Well. OK. No more arguments from me. I will just do that.

Jojo-Schmitz:
Something went wrong with your rebase

ypoissant:
I just did a new rebase. I hope this one is good.

libmscore/easeInOut.cpp
ypoissant: BTW, my preferred reference concerning floating point error management is from the book "[Physically Based Rendering: From Theory to Practice, Chapter 3.9 - Managing Rounding Errors](http://www.pbr-book.org/3ed-2018/Shapes/Managing_Rounding_Error.html)". I invite you to read this chapter for an in-depth explanation.

njvdberg: The first link is about floating comparisons, while your link is about rounding errors. So, different subjects
 Nevertheless, thanks, very interesting. I will certainly go over it. Reminds me to my old study years, back in the seventies (but without a nice book!) :smile: .

njvdberg:
Still something wrong, this commit contains too many commit from other PR's. Rebase from a wrong reference? Or a merge instead of a rebase?

Jojo-Schmitz:
Try `git pull --rebase musescore 3.x` (if `musescore` is the name of your upstream repo).
And indeed it is better to create your own topic branch rather than working off of a branch name that exists upstream too, but for this PR that bird has flown I guess.

ypoissant:
To be honest, I have next to no experience with Git and GitHub. I followed the handbook instructions and created a fork. I thought this was a branch. Then from this fork, I created a repository on my computer and worked from this repository. I understand that the fork is known as 'origin' on my computer and I also created a reference to the main musescore repository, the one forked from, named 'upstream'. I understand that this was wrong but I really don't know how to fix that.  I will try with "git pull -- rebase upstream 3.x". I think that is what you instruct me to do.

Jojo-Schmitz:
Yes it is. And a rebase is now needed anyway, to resolve conflicts

ypoissant:
Done.

Jojo-Schmitz:
Nope

njvdberg:
> To be honest, I have next to no experience with Git and GitHub. I followed the handbook instructions and created a fork. I thought this was a branch.

Unfortunately, a fork is not a branch. The fork is just your "private" copy on which you can create branches and change/add code. I think it must be possible to retrieve your changes into a branch but I'm no git guru either and have to try a few things first. I'm willing to help you with this, somewhere this weekend.

libmscore/easeInOut.cpp
ypoissant: @njvdberg To me, this is the same subject. Rather than focusing on the application of a float comparison rule, I focus on the consequences of using a float comparison vs letting the error go through. This is something I learned in 3D rendering where performance is of paramount importance and the heavy arsenal of error management (of which qFuzzyCompare is a member) is avoided by making the code robust to errors. Error management code is used only in the very few places where it is absolutely required.

So I spent one month of leisure time analyzing the boundary conditions of the Bezier functions using Mathematica before I started looking at MuseScore code. I tried to argue, here, that there are no consequences in breaking this rule in this function because even if the zero discriminant is only very close to zero due to float errors, the path that will be followed will still work correctly. This is even more so given the constraints in which the cubic Bezier is used as transfer function.

What turned me off, however, is the requirement to use an int in a for loop. I fully accept the example where the for loop will loop through one too few. That is a good example and a very possible outcome. A blatant example of why code review is good. However, IMO, the right solution is to add one half of the increment step to the stop condition. Not to use int. Using int in interpolation functions results in, for example, the MuseScore bend function not reaching its true final note most of the time.

I understand you are a 50 years experienced Fortran and C++ (and probably more) programmer. I respect you for this. I'm also an experienced programmer. 40 years with Assembler, Pascal, Lisp, Mathematica, C/C++, Javascript (a little bit of Fortran as student and a few other arcane languages). My specialty is in 3D computer graphics and rendering, of which the last 10 years in physically-based rendering.

ypoissant:
I accept your offer to help this weekend. At the worst, I can save a copy of my changes. revert everything in my fork, create a branch and merge my changes back into the branch.

Jojo-Schmitz:
You did a `git pull --rebase upstream 3.x`, followed by `git push -f`?

ypoissant:
Yes. I did a "git pull --rebase upstream 3.x" then "git push --progress origin 3.x:3.x"

ypoissant:
I'll try "git push -f"

ypoissant:
I just did "git push -f" and it replied "Everything is up-to-date"

Jojo-Schmitz:
No idea what `git push --progress origin 3.x:3.x"` does

ypoissant:
I just did

Jojo-Schmitz:
Ah, yes!

libmscore/easeInOut.cpp
njvdberg: @ypoissant I started with Fortran (and Algol) during my studies and used Fortran in my first job for scientific programs. Last 20 years I worked as a CAD engineer (semiconductors) and used mainly SKILL/SKILL++ (which is a Lisp dialect). There I  spent days on programs showing a very strange result (or even didn't come to an end) and often it was because in the algorithm there was a floating (not) equal test. E.g. in one case I found the string "1n" and "1e-9" (which is the same) give a different result. it turns out "1n" was internally replaced by an expression 1.0*1e-9 which which changes in the last bit from 1e-9. This kind of issues I've seen several times over the years. And sometimes in code which worked fine for years. That's the reason I'm always triggered when I see this kind of test. No intent to offend you, and if I did, my apologies.

My worry is still, in this particular case, depending on the calculation, <code>y</code> is 0.0 but due to the limited number of bits for the mantissa the last bit flips so the result is not exactly 0.0 but something like 1.11022e-16 (on my computer). then the test <code>y == 0.0</code> will fail but <code>qFuzzyCompare(y, 0.0)</code> not since this functions tries to take this last mantissa bit into account.

njvdberg:
> At the worst, I can save a copy of my changes. revert everything in my fork, create a branch and merge my changes back into the branch.

Let's see what we can do here. At least it sounds as an interesting issue :smiley:.

njvdberg:
Just cloned your repository and it isn't that difficult to "save" your changes. Main issue I see is the fact the changes are made on 3.x instead of a branch.
What I did was was
 - Checkout your branch to <code>easeInOut</code> (but almost any name is fine).
 - Squash the all your commits into one single commit. Any PR is just one commit, unless it solves different issues or the commit are really unrelated.
- Rebase the branch <code>easeInOut</code> on <code>upstream/3.x</code>. <code>upstream</code> is the here name of the main repository <code>git://github.com/musescore/MuseScore.git</code>.

Now you can push this <code>easeInOut</code> and create a new PR for that branch and close this one. Then everything should be fine.

ypoissant:
OK. Thanks for the directions. I'm not sure how to do most of that though.

I'll look for the 'checkout' command and how to use it. I understand that this is used to create the branch on my local repository.

I'll look at how to squash the commits into a single commit. My guess is it's done manually or is there a git command to do that?

I'll do that later today.

ypoissant:
Is there anything I have to do on my side to fix the mtests and vtests errors? I don't see how those errors may be related to my changes since none of the vtests and mtests pass.

njvdberg:
If you are not familiar with git, I found [this GIT manual/tutorial](https://git-scm.com/book/en/v2) very helpful.

For creating branches, I always use <code>git checkout -b "branch-name" "start-commit"</code>. This creates a branch <code>branch-name</code> from <code>start-commit</code>, e.g. <code>git checkout -b easeInOut upstream3.x"</code>.

For rebase, use <code>git rebase -i HEAD~"n"</code> where <code>n</code> is the number of commit you want to squash. This command will show a command file you can edit. For each commit specify what you want to do. Please note the commit are in reversed order. If you want to squash all commits into one, leave the first line unchanged (<code>pick</code> are replace <code>pick</code> at the following lines by <code>squash</code> or <code>fixup</code>. Using <code>squash</code> you can edit the commit message too, when using <code>fixup</code> the commits of the "fixed up" commit is regarded.

ypoissant:
Thanks for the detailed instructions. I have the book GitPro and also search in the git tutorials. But being used to SVN and TFS, which are very different paradigms, I find it usually difficult to map concepts from one to the other especially given that the two types of systems use the same keywords to designate slightly different concepts. I think I get the general concepts but the effect of each command variations is beyond me.

njvdberg:
Regarding the vtests and mtests it seems the program crashes while trying to get a style option. If I remember correctly, you added style options? In that case the lists in both <code>style.cpp</code> and <code>style.h</code> are no longer in sync. Better to check this carefully and make sure in both <code>enum class Sid</code> in <code>libmscore/style.h</code> and <code>static const StyleType styleTypes[]</code> in <code>libmscore/style.cpp</code> the Sid's are in the same order.

njvdberg:
> But being used to SVN and TFS

I can remember that Linus Tovalds, when started git, said that should everything what SVN isn't. So no surprise it is a "little" different :wink:

Jojo-Schmitz:
> Is there anything I have to do on my side to fix the mtests and vtests errors? I don't see how those errors may be related to my changes since none of the vtests and mtests pass.

Same as https://github.com/musescore/MuseScore/pull/6852#pullrequestreview-529193843 ?

ypoissant:
No. I did fixed it and at some point all the tests passed. But it seems some of the multiple push, fetch and rebase trashed something. That's scary. Lucky that I did not trust Git and saved a copy of all my changes locally.

ypoissant:
Revision history
Here is the revision history. Indeed, the first revision I did was re-commited and trashed all the 7 or 8 commits I did in between.

ypoissant:
Closing this PR because I created a branch and I'm about to call a PR from it.

Jojo-Schmitz:
Continued in #6868

PR #6854, opened 2020-11-12
[MU3] Disabled qml cache for all platforms

igorkorsukov:
I tested run MuseScore on MacOS BigSur on an ARM processor (macmini) - and found out, that MuseScore crashed on start. 
I investigated the problem and found out if set `QML_DISABLE_DISK_CACHE=true` MuseScore starts and works fine.  (except WebView (QWebEngine) - it doesn't work)

There were also crash problems on some Linux platforms when upgrading with qml cache.
In general, the Qml cache doesn't really matter for MuseScore 3
Therefore, it is better to turn it off.

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -521,7 +521,7 @@ void MuseScore::preferencesChanged(bool fromWorkspace, bool changeUI)
       // change workspace
       if (!fromWorkspace && preferences.getString(PREF_APP_WORKSPACE) != WorkspacesManager::currentWorkspace()->name()) {
             Workspace* workspace = WorkspacesManager::findByName(preferences.getString(PREF_APP_WORKSPACE));
-            
+
             if (workspace)
                   mscore->changeWorkspace(workspace);
             }
@@ -793,7 +793,7 @@ bool MuseScore::importExtension(QString path)
 
                   if (!sfzs.isEmpty())
                         synti->storeState();
-                  
+
                   s->gui()->synthesizerChanged();
                   }
 
@@ -816,7 +816,7 @@ bool MuseScore::importExtension(QString path)
                         }
                   if (!sfs.isEmpty())
                         synti->storeState();
-                  
+
                   s->gui()->synthesizerChanged();
                   }
             };
@@ -883,7 +883,7 @@ bool MuseScore::uninstallExtension(QString extensionId)
                   }
             if (found)
                   synti->storeState();
-            
+
             s->gui()->synthesizerChanged();
             }
       bool refreshWorkspaces = false;
@@ -913,7 +913,7 @@ bool MuseScore::uninstallExtension(QString extensionId)
             bool curWorkspaceDisappeared = false;
             if (WorkspacesManager::findByName(curWorkspaceName))
                   curWorkspaceDisappeared = true;
-            
+
             if (curWorkspaceDisappeared)
                   changeWorkspace(WorkspacesManager::workspaces().last());
             }
@@ -1288,7 +1288,7 @@ MuseScore::MuseScore()
       {
       feedbackTools = addToolBar("");
       feedbackTools->setObjectName("feedback-tools");
-	  // Forbid to move or undock the toolbar...
+      // Forbid to move or undock the toolbar...
       feedbackTools->setMovable(false);
       feedbackTools->setFloatable(false);
       // Add a spacer to align the buttons to the right side.
@@ -2119,7 +2119,7 @@ void MuseScore::retranslate()
       if (paletteWorkspace)
           paletteWorkspace->retranslate();
       }
-      
+
 //---------------------------------------------------------
 //   setMenuTitles
 //---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


igorkorsukov:
opps, changed code style, I am going to return it (auto format)

Jojo-Schmitz:
But wouldn't without QWebEngine not working the startcenter also not work?

igorkorsukov:
> But wouldn't without QWebEngine not working the startcenter also not work?

Starcenter working, but not showing web content (just white blank)

Jojo-Schmitz:
Which means that it is not working, ot the same as it the `-w` option had been used. So not a solution we should be using, IMHO

igorkorsukov:
> Which means that it is not working, or the same as if the `-w` option had been used. So not a solution we should be using, IMHO

I think, better now release version without working startcenter, instead of not having any working version on MacOS BigSur on ARM

Jojo-Schmitz:
Then why not excluding for Mac BugSur on Arm only?

igorkorsukov:
> Then why not excluding for Mac BugSur on Arm only?

I didn't quite understand the question, why not turn off the qml cache only for BugSur on Arm?

Jojo-Schmitz:
yes, why off for all, even those that don#t have an issue with it?

igorkorsukov:
> yes, why off for all, even those that don#t have an issue with it?

There were already gaps with Qml cache on other platforms.
I believe qml cache is not important for MuseScore 3.
Therefore, I believe that it is better to just turn it off for everyone.
Instead of wasting time looking for how to reliably determine that MuseScore is running on MacOS BigSur on a Arm processor in compatibility mode (under Rosetta emulator) :)

Jojo-Schmitz:
Any way to get the startcenter fully working regardless?
How's performance affected, like for the now qml based palettes?

Jojo-Schmitz:
And what about the "Save online" dialog?

igorkorsukov:
> Any way to get the startcenter fully working regardless?
Perhaps we can make patches of Chromium, rebuild it, and use "self" QWebEngine
https://bugreports.qt.io/browse/QTBUG-86405

> How's performance affected, like for the now qml based palettes?
I don't notice the difference

Jojo-Schmitz:
Checked "Save online"?

igorkorsukov:
> Checked "Save online"?

Checked it out. Unfortunately, it doesn't work either.

Jojo-Schmitz:
So we have an absolute show stopper for this change

igorkorsukov:
> So we have an absolute show stopper for this change

Maybe I didn't understand you, but you think it's better not to have any working version - a splash screen is shown and crashes on start. Instead of having a version that launches and works the main functionality, except for the start center and saving online? (and maybe something else)

Ximich:
mac/arm: Update docs -- everything builds and runs now!

Startup is pretty slow at the moment for reasons not yet known,
but other than that things seem to work.

We haven't tried running any of the test binaries yet.

Ximich:
https://bugs.chromium.org/p/chromium/issues/detail?id=1098899&q=%22apple%20silicon%22&can=2

Jojo-Schmitz:
It is indeed better to not have a version for Mac BigSur on ARM, if that means no other version can use "Save online"

Jojo-Schmitz:
Strange enough the PortableApp for 3.5.2 can use "Save online", but is (currently, before your change) the only version with a disabled qml cache?

Ximich:
>>> mac/arm: Update docs -- everything builds and runs now!
>>> https://ci.chromium.org/p/chromium/builders/ci/mac-arm64 is green and the binary even starts.
>>> So I think we can mark this one fixed and have targeted bugs (under label Mac-Arm64) for stuff that's left to do.

https://bugs.chromium.org/p/chromium/issues/detail?id=1098899&q=%22apple%20silicon%22&can=2

igorkorsukov:
> It is indeed better to not have a version for Mac BigSur on ARM, if that means no other version can use "Save online"

:) There was a misunderstanding.

Disabling Qml cache does not affect QWebEngine at all (startcenter, save online).
This disabling the cache does not affect anything at all, except for a small load optimization (which is not even noticeable).

The point is only that QWebEngine does not work on MacOS BigSur on ARM processor, so the start center does not work, save online, only this platform.
Of course, everything works on other platforms.

Jojo-Schmitz:
Whatever, don't sacrifice "Save online" (or the web part of the start center) for Windows, almost all Macs, and Linux for the brandnew Mac BigSur on ARM

Jojo-Schmitz:
Oh, a misunderstanding then. Yes, I think we can live with the restriction that macOS BigSur on ARM cannot use "Save online" or the web part of the start center. We'd just need to make that clean in the release notes.

igorkorsukov:
> Oh, indeed a misunderstanding then. Yes, I think we can live with the restriction that macOS BigSur on ARM cannot use "Save online" or the web part of the start center. We'd just need to make that clean in the release notes.

Yippee!, we have resolved the misunderstanding :) 
Now we are investigating how to solve the problem with the web engine for BigSur on ARM

Jojo-Schmitz:
;-)

PR #6856, opened 2020-11-13
[MU3] Various changes for Leland

Jojo-Schmitz:
* Add Leland and Edwin fonts to Readme.md
* Rename some edwin/*.txt to *.md
* Fix compiler warning
* Fixing a usage message (to not mention any fallback font. man page too)
* Unify a dialog string (using the same as for importing pre-3.x scores) ___(actually that old dialog seems to have gotten removed entirely, removing the changes from PR #6255, to (partially) fix issue [#306911](https://musescore.org/en/node/306911))___
* Maintaining alphabetical order of fonts
or
* Leland on top, rest alphabetical

Let me know whether you want "Maintaining alphabetical order of fonts" or "Leland on top, rest alphabetical", I'd prefer the former (is used for Text fonts too, the former list was alphabetical and I'm not 100% sure whether Bravura as the fallback font needs to stay on top), but see some logic behind the latter (default on top).


libmscore/layout.cpp
```diff
@@ -3720,13 +3720,15 @@ System* Score::collectSystem(LayoutContext& lc)
             }
 
       hideEmptyStaves(system, lc.firstSystem);
+#if 0
```
Jojo-Schmitz, line 3723: Looks like dead code to me?

fonts/README.md
```diff
@@ -8,7 +8,9 @@ This directory contains the fonts used internally by MuseScore and the sources u
 
 * **bravura** The [_Bravura_](https://github.com/steinbergmedia/bravura/) fonts and the documentation for them. These fonts are **not** maintained by the MuseScore community, please **DO NOT EDIT** these fonts. 
 * **campania** The [_Campania_](https://github.com/MarcSabatella/Campania) font for Roman numeral analysis.  This font is not maintained by the MuseScore project, but it is open source and contributions from the MuseScore community are encouraged at the main site for the font.
+* **edwin** The [_Edwin_] text font family. This font is maintained by the MuseScore project.
```
cbjeukendrup, line 11: Shouldn't the [ ] be left out, because there is no hyperlink?

Jojo-Schmitz: Yes, thanks

Jojo-Schmitz:
I'm pretty sure the mtest failure is not my (/PR's) fault, got to be the one just before mine (from PR #6776)

libmscore/layout.cpp
njvdberg: Not anymore, this code is restore in latest <code>3.x</code> commits.

Jojo-Schmitz: Nope, I'm on the latest

Jojo-Schmitz: 8c7af0b2b removes an `if (!allShown) `

Jojo-Schmitz: So exactly the last commit before my PR

vpereverzev:
@oktophonie take a look please

PR #6857, opened 2020-11-13
Fix #312489 Harmony duration interpretation

jeetee:
Resolves: https://musescore.org/en/node/312489

Makes the Harmony realization follow the repeatlist order when determining the next Harmony element

- [N/A] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)

### DIFF
--- a/libmscore/cmd.cpp
+++ b/libmscore/cmd.cpp
@@ -3195,8 +3195,8 @@ void Score::cmdRealizeChordSymbols(bool literal, Voicing voicing, HDuration dura
                   continue;
             RealizedHarmony r = h->getRealizedHarmony();
             Segment* seg = h->parent()->isSegment() ? toSegment(h->parent()) : toSegment(h->parent()->parent());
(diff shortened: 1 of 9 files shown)


jeetee:
Currently Harmony is followed even across RepeatSegments, this might be unwanted. Reviewing with @MarcSabatella now

MarcSabatella:
I approve!  We discussed different cases, some it's questionable as to what user might prefer by default, but the implementation here makes it easy to stop playback with an "NC" or by explicitly setting duration to measure.  The alternative of stopping playback at the end of the repeat segment would not have an equally good way to continue it.  In any case, it would be easy to tweak the part of the code that moves on to the next segment to not do so, if we get complaints.

vpereverzev (approved):

vpereverzev:
looking forward for rebase and then I'll be able to merge this

PR #6858, opened 2020-11-14
Fix #312959: Implement New Navigation Commands (Frames/Sections/Playback Cursor)

worldwideweary:
Resolves: https://musescore.org/en/node/312959

Included herein are the following commands:

**[Next/previous section]**
This includes frames so that if a traversed section has a frame, that frame will be the new position: else, the first chord-rest of traversed section will be.

**[Next/previous frame]**
This only traverses frames [horizontal/vertical/text], having no regard of section breaks. This also updates the [next/previous measure] commands to allow getting out of a frame and into the next or previous measure, as  nothing would happen before these commands existed.

**[Go to recent playback cursor position]**
This takes into consideration current track (voice/instrument), but if a ChordRest exists in a given segment where the initial track doesn't, it will take the user there. I decided to not have this take the user to a non-musical sounding element since the playback cursor only traverses chord/rest elements. In that sense, this won't do exactly what was desired by @MarcSabatella, i.e., take the user to the beginning barline of a next measure for score accessibility read-out purposes after a section break is met with. For that, if a frame exists, a [next section] command would work and then [alt+right] to traverse elements, and if not, [next section] + [alt-left] to read the elements before the first chordrest of the new section. If this command is issued at the end of a section after stopping playback, current selection will be at the next measure's first chord-rest rather than first element. Hopefully it gets some use. Still open to suggestions, but this currently seems, imho, to be the most "generically acceptable" way for this command.
Included is a demonstration .gif of **[Go to recent playback cursor position]**:

GotoPlayCursor

### DIFF
--- a/libmscore/cmd.cpp
+++ b/libmscore/cmd.cpp
@@ -2315,8 +2315,8 @@ Element* Score::move(const QString& cmd)
                                     }
                         break;
                         }
-                  case ElementType::HBOX: Q_FALLTHROUGH();
-                  case ElementType::VBOX: Q_FALLTHROUGH();
+                  case ElementType::HBOX: // fallthrough
+                  case ElementType::VBOX: // fallthrough
                   case ElementType::TBOX:
                         box = toBox(el);
                         break;
--- a/libmscore/cmd.cpp
+++ b/libmscore/cmd.cpp
@@ -2443,6 +2443,16 @@ Element* Score::move(const QString& cmd)
             auto measureBase = cr ? cr->measure()->findMeasureBase() : box->findMeasureBase();
             el = measureBase ? cmdNextPrevFrame(measureBase, false) : nullptr;
             }
+      else if (cmd == "next-section") {
+            if (!(el = box))
+                  el = cr;
+            el = cmdNextPrevSection(el, true);
+            }
+      else if (cmd == "prev-section") {
+            if (!(el = box))
+                  el = cr;
+            el = cmdNextPrevSection(el, false);
+            }
       else if (cmd == "next-track" && cr) {
             el = nextTrack(cr);
             if (noteEntryMode())
--- a/libmscore/cmd.cpp
+++ b/libmscore/cmd.cpp
@@ -2243,7 +2243,8 @@ bool Score::processMidiInput()
 
 Element* Score::move(const QString& cmd)
       {
-      ChordRest* cr;
+      ChordRest* cr { nullptr };
+      Box* box { nullptr };
       if (noteEntryMode()) {
             // if selection exists and is grace note, use it
             // otherwise use chord/rest at input position
@@ -2274,7 +2275,8 @@ Element* Score::move(const QString& cmd)
             // trg is the element to select on "next-chord" cmd
             // cr is the ChordRest to move from on other cmd's
             int track = el->track();            // keep note of element track
-            el = el->parent();
+            if (!el->isBox())
+                  el = el->parent();
             // element with no parent (eg, a newly-added line) - no way to find context
             if (!el)
                   return 0;
@@ -2313,6 +2315,11 @@ Element* Score::move(const QString& cmd)
                                     }
                         break;
                         }
+                  case ElementType::HBOX: Q_FALLTHROUGH();
+                  case ElementType::VBOX: Q_FALLTHROUGH();
+                  case ElementType::TBOX:
+                        box = toBox(el);
+                        break;
                   default:                      // on anything else, return failure
                         return 0;
                   }
@@ -2326,8 +2333,8 @@ Element* Score::move(const QString& cmd)
                   select(trg, SelectType::SINGLE, 0);
                   return trg;
                   }
-            // if no chordrest found, do nothing
-            if (!cr)
+            // if no chordrest and no box (frame) found, do nothing
+            if (!cr && !box)
                   return 0;
             // if some chordrest found, continue with default processing
             }
@@ -2336,7 +2343,7 @@ Element* Score::move(const QString& cmd)
       Segment* ois = noteEntryMode() ? _is.segment() : nullptr;
       Measure* oim = ois ? ois->measure() : nullptr;
 
-      if (cmd == "next-chord") {
+      if (cmd == "next-chord" && cr) {
             // note input cursor
             if (noteEntryMode())
                   _is.moveToNextInputPos();
@@ -2362,7 +2369,7 @@ Element* Score::move(const QString& cmd)
             else if (!el)
                   el = cr;
             }
-      else if (cmd == "prev-chord") {
+      else if (cmd == "prev-chord" && cr) {
             // note input cursor
             if (noteEntryMode() && _is.segment()) {
                   Measure* m = _is.segment()->measure();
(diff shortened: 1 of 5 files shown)


libmscore/cmd.cpp
```diff
@@ -2391,31 +2398,45 @@ Element* Score::move(const QString& cmd)
...
-            if (noteEntryMode())
+            if (box && box->prevMeasure() && box->prevMeasure()->first())
+                  el = box->prevMeasure()->first()->nextChordRest(0, false);
+            if (cr)
+                  el = prevMeasure(cr);
+            if (el && noteEntryMode())
                   _is.moveInputPos(el);
             }
-      else if (cmd == "next-system") {
+      else if (cmd == "next-system" && cr) {
```
Marr11317, line 2416: what about "next-system" when `cr` is falsy?

libmscore/score.cpp
```diff
@@ -4028,6 +4028,22 @@ ChordRest* Score::cmdNextPrevSystem(ChordRest* cr, bool next)
...
 
+//---------------------------------------------------------
+//   cmdNextPrevFrame
+//   Return next/previous [Vertical/Horizontal/Text] frame
+//   to be used as a navigation command
+//---------------------------------------------------------
+
+Box* Score::cmdNextPrevFrame(MeasureBase* currentMeasureBase, bool next) const
+      {
+            Box* selectedBox { nullptr };
```
Marr11317, line 4039: I remember seeing a discussion about list initialization somewhere, but don't remember the outcome. Didn't we use `selectedBox = nullptr` in the rest of the codebase?

libmscore/cmd.cpp
```diff
@@ -2301,6 +2303,11 @@ Element* Score::move(const QString& cmd)
                                     }
                         break;
                         }
+                  case ElementType::HBOX: Q_FALLTHROUGH();
```
Marr11317, line 2306: If we're trying to get rid of Qt in libmscore, maybe just use a comment?
                      case ElementType::HBOX: // fallthrough

libmscore/cmd.cpp
worldwideweary: No need

libmscore/score.cpp
worldwideweary: Not sure what they used.

libmscore/cmd.cpp
worldwideweary: Are we?

Marr11317: that's a question for @igorkorsukov

libmscore/cmd.cpp
Marr11317: what did it do before this PR?

worldwideweary: Although..., allowing _prev-system_ to execute when a valid frame is selected would add a little functionality for when going backwards. Otherwise, _next-measure_ would do what would be imagined _next-system_ would do when a valid frame is selected.

worldwideweary: Nothing. A valid ChordRest* was and is required.

MarcSabatella:
Thanks for this!  Personally I'd be fine with next-measure traversing frames too, I guess the main thing to think about would be how that affected Ctrl+Shift+Right for extending selection by measures.

Anyhow, being able to traverse frames Is nice, but I fear it doesn't really fully address the use case I was concerned with, which was how a blind user could easily reach the next section - whether it starts with measure or frame - after playing back to the end of one section.  For that purpose a next-section command really would be preferable.  Or, perhaps at least as useful and even more so to the average user, a "go to playback position".  That way you could stop playback *anywhere* - not just at a section break - and quickly resume navigation.  Probably most people might want it to select the first chord/rest after the playback cursor, but I'd need that be first *element*, including potentially a frame.

worldwideweary:
> Thanks for this! Personally I'd be fine with next-measure traversing frames too

Shouldn't be too difficult, but the problem is that it would require reshaping the next/prev measure command for measures since it returns and takes only chordrests*. Could probably deal with frames without the main next/prev measure function and do some things in the command handling area instead

But then you'd have to check for playback activity because one can use per-measure navigation while playing. To jump to a frame in the middle of playing wouldn't make sense and would have to be safe guarded against.

As for section commands, I might be able to add a commit to this PR including those soon. I was under the impression that most use-cases of section breaks are accompanied by a following frame of some sort for the next section, so first frame traversal became priority before doing anything else (plus adding getting to a measure from a frame by updating the measure navigation shortcuts). In your case, I imagined having an introductory frame be read first before continuing playback or exploring the musical information. Wouldn't want a blind user to miss a frame between one section and the next if there were one.

MarcSabatella:
That's a good point too about frames *within* the section.  I don't do that personally, but it definitely speaks to the value of the next-frame command!

In my case, my theory handouts general go frame - music - frame - music - frame - music etc.  There is a section break at the end of each block of "music".  So next frame and next section end up being the same usually.  But *some* of the music blicks, there are section breaks *within* the block.  Like, multiple four-bar examples for one text frame.  So if a person started playback, it would pause after the first example, and they wouldn't realize there is another example before the frame.  Of course, if they continued playback they would hear it, but they wouldn't know if the example went with the same block or the next block, until they went through everything to figure it out.

So really for my own purposes, the "goto to playback position" is actually the best solution, it handles all these cases.  Not sure exactly where the playback position is in this case, but I'd guess it's the end tick of the measure with the break, which is also the start tick of the next measure or frame, and it's that measure or frame I'd want,  Sighted users would probably want only notes or rests, but blind users would want the first element of the measure, probably the header clef.

Nice thing, though, about next-section is that it allows someone to quickly read the text only and skip the examples, which no doubt has lots of value too, so I'm glad you did this!

libmscore/cmd.cpp
```diff
@@ -2443,6 +2443,55 @@ Element* Score::move(const QString& cmd)
             auto measureBase = cr ? cr->measure()->findMeasureBase() : box->findMeasureBase();
             el = measureBase ? cmdNextPrevFrame(measureBase, false) : nullptr;
             }
+      else if (cmd == "next-section") {
```
worldwideweary, line 2446: These haven't been tested, will download artifact tomorrow and try it out, and if it all works fine, could easily fix this up to be more modular looking. Something's up with "make revision" in 3.x for me currently.

libmscore/score.cpp
```diff
@@ -4080,12 +4080,124 @@ ChordRest* Score::cmdNextPrevSystem(ChordRest* cr, bool next)
 
 Box* Score::cmdNextPrevFrame(MeasureBase* currentMeasureBase, bool next) const
       {
-            Box* selectedBox { nullptr };
```
worldwideweary: Snuck in a banner-style fix.

libmscore/score.cpp
```diff
@@ -4080,12 +4080,124 @@ ChordRest* Score::cmdNextPrevSystem(ChordRest* cr, bool next)
...
+                        }
                   }
-            return selectedBox;
+            // -------------------------
+            // Previous Section of Score
+            // -------------------------
+            else {
+                  auto currentSegment = el->isChordRest() ? toChordRest(el)->segment() : nullptr;
+                  if ((destination = getNextPrevSectionBreak(currentMeasureBase, false))) {
+                        if (currentSegment) {
```
worldwideweary, line 4115: Dealt with segments here so that 1) Previous Section acts like Previous System in that the user will go to the beginning of the current section first if "in the middle" of a section (even if in its first measure but not at its beginning)

worldwideweary:
... Let's just say I won't express  how many times I had to compile and test-run in order to get all this to work :)

libmscore/cmd.cpp
igorkorsukov: Right, we want to reduce our dependency on Qt where possible.
But we want to do this evolutionarily, so now the rule is this: where Qt is already used, you can use Qt, where it does not use, you do not need to use Qt. 
So far, everything is simple.

worldwideweary: Ok, I changed these to` // fallthrough`
There's really only about 3 different Q_FALLTHROUGH(); statements in libmscore/ directory. I'm almost tempted to just convert them here to be done with it, but I'll leave it alone.

RobFog:
> ... Let's just say I won't express how many times I had to compile and test-run in order to get all this to work :)

Yet you forgot to include "fixes #312959"? :P

worldwideweary:
> Yet you forgot to include "fixes #312959" in your commit message?
Noticed your PR Merged messages in the issue trackers. Thanks! 
I was under the impression that the PR title was what mattered, but looks like i was incorrect. The first commit message apparently needs the Fix #xxxxxx? If that is true, I'll make sure to have the first commit message have such a title in the future if I do some MU4 activities. This is one of the first times I tried having what I thought to be clean commit messages separated rather than squashed since there are multiple functions going on here.

Later.

RobFog:
Yes, it's the commit messages that matter. I don't think it matters which of the commits has it. @Jojo-Schmitz?

Jojo-Schmitz:
Yes

Jojo-Schmitz:
Commit message title actually

PR #6862, opened 2020-11-14
Fix build on QtCreator/MinGW

Jojo-Schmitz:


libmscore/scoreOrder.cpp
```diff
@@ -364,7 +364,7 @@ QString ScoreOrder::getFamilyName(const InstrumentTemplate* instrTemplate, bool
 QString ScoreOrder::getId() const
       {
       if (_customised)
-            return QString("%1-%2").arg(_id).arg(reinterpret_cast<ulong>(this), 0, 16);
+            return QString("%1-%2").arg(_id).arg(reinterpret_cast<qulonglong>(this), 0, 16);
```
igorkorsukov, line 367: this is good but better use `quintptr`

Jojo-Schmitz: done

PR #6863, opened 2020-11-14
[MU3] Added make local build revision

igorkorsukov:
Added make local build revision

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ build.qt5
 *.sfd~
 /.project
 mscore/revision.h
+local_build_revision.env
 .DS_Store
 CMakeLists.txt.user.*
 *.qmlc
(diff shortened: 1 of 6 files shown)


worldwideweary:
Looking forward to the merge :)

shoogle:
How will the `MUSESCORE_REVISION` variable get set for incremental builds (i.e. builds where CMake is not run beforehand)?

igorkorsukov:
> How will the `MUSESCORE_REVISION` variable get set for incremental builds (i.e. builds where CMake is not run beforehand)?

During development, it is not expected to set a revision.
They need to complete the development, make a commit, and only then update the revision and build the application (with the run cmake). If someone needs all this.

shoogle:
Fair enough. I guess the revision is more important for nightlies than for personal builds, especially if it refers to a personal commit that's not in the main repository.

PR #6865, opened 2020-11-15
[MU3] Fix some compiler warnings

Jojo-Schmitz:


mscore/instrwidget.cpp
```diff
@@ -622,15 +622,15 @@ int InstrumentsWidget::findPrvItem(PartListItem* pli, int number)
       const int last = number ? number - 1 : partiturList->topLevelItemCount();
       QTreeWidgetItem* item = nullptr;
       for (int idx = 0; --number && (item = partiturList->topLevelItem(idx)); ++idx) {
-            PartListItem* pli = (PartListItem*)item;
```
njvdberg, line 625: What is the problem this rename is solving? Just curious.

Jojo-Schmitz: MSVC complains the variable name to be the same as the function's parameter

njvdberg: MSVC doesn't like shadowing then. Something to keep in mind.
By the way, it might be solved also by the new PR I'm creating for ScoreOrdering (is related to it).

PR #6868, opened 2020-11-15
Add Ease in and ease out to all glissandi styles

ypoissant:
Resolves: [Glissando portamento playback on 3.5-dev issues](https://musescore.org/en/node/302937)

Linearly interpolated glissandi don't sound natural. It does not replicate the starting and ending inertia of the glissando movement. It does not allow expressive glissandi execution. The ease-in and ease-out properties can be set to values from 0 to 100 for interpolation that go from linear to 'S' of different shapes. While implementing this, I fixed two issues.

### DIFF
--- a/libmscore/easeInOut.cpp
+++ b/libmscore/easeInOut.cpp
@@ -1,0 +1,111 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2020 MuseScore BVBA
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2
+//  as published by the Free Software Foundation and appearing in
+//  the file LICENCE.GPL
+//=============================================================================
+
+/**
+ \file
+ Implementation of classe easeInOut for implementing transfer curve with
+ parametrerable ease-In and ease-Out.
+*/
+
+#include "easeInOut.h"
+
+namespace Ms {
+
+//-------------------------------------------------------------------------------------------------
+// The following function is inspired by "A Primer on Bézier Curve"sections 17 and 23 by Pomax:
+// https://pomax.github.io/bezierinfo/
+// However, the fuction is greatly specialized, simplified and optimized for use as an ease-in and
+// ease-out transfer curve for bends, glissandi and portamenti in MuseScore. The function computes
+// Y from X by performing a cubic root finding to compute t from X on the X component of the
+// transfer curve and then a cubic Bezier evaluation to compute Y from t on the Y component of the
+// transfer curve. The code could be simplified because the Bezier curve is constrained as a well
+// defined transfer function. It is not suitable for general or arbitrary Bezier curve work. -YP
+//-------------------------------------------------------------------------------------------------
+qreal EaseInOut::tFromX(const qreal x) const {
+      const qreal pi = 3.14159265358979323846;
+      qreal c = -x;
+      qreal w1 = _easeIn - x;
+      qreal w2 = 1.0 - _easeOut - x;
+      qreal d = x + 3.0 * w1 - 3.0 * w2 + (1.0 - x);
+      qreal a = (-3.0 * x - 6.0 * w1 + 3 * w2) / d;
+      qreal b = (3.0 * x + 3.0 * w1) / d;
+      c = -x / d;
+      qreal a2 = a * a;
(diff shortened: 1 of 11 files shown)


Jojo-Schmitz:
What is the difference between the 1st and the 2nd commit? How can this happen??

The 3rd one doesn't belong here at all.

ypoissant:
Really I don't know. I actually don't know what I'm doing with Git and all these commands. I'm just applying the suggestion as is.  I just redid a "git pull --rebase upstream 3.x" followed by a "git push -f". Let's see what that will give.

ypoissant:
I applied those two git commands, as is, 3 times yet, on this branch and they all resulted in different sets of commits. I agree, the duplicate commit we saw previously doesn't make sense to me too.

libmscore/glissando.cpp
```diff
@@ -286,10 +292,9 @@ void Glissando::layout()
          // but ignore graces after, as they are not the first note of the system,
          // even if their segment is the first segment of the system
          && !(cr2->noteType() == NoteType::GRACE8_AFTER
-            || cr2->noteType() == NoteType::GRACE16_AFTER || cr2->noteType() == NoteType::GRACE32_AFTER)
+         || cr2->noteType() == NoteType::GRACE16_AFTER || cr2->noteType() == NoteType::GRACE32_AFTER)
```
Jojo-Schmitz, line 295: IMHO that indent should stay here, to repflect the parens

libmscore/easeInOut.cpp
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+//-------------------------------------------------------------------------------------------------
+// Given a number of note to place within the given duration, return the list of on-times for each
+// note given the current ease-in and ease-out parameters. The first note is at time 0 while the 
+//-------------------------------------------------------------------------------------------------
+void EaseInOut::timeList(const int nbNotes, const int duration, QList<int>* times) const {
+      qreal nNotes = qreal(nbNotes);
+      qreal space = qreal(duration);
+      if (_easeIn == 0.0 && _easeOut == 0.0) {
```
njvdberg, line 100: Is this <code>0.0</code> in this case the result of a calculation? If yes, there can be cases the calculation which should return <code>0.0</code> will be, due to the limited length of the mantissa, one bit off and switch to the wrong branch. Also here <code>qFuzzyCompare()</code> is better.

ypoissant: Not result of calculation. This coming from either the inspector slider or the spinbox, which are ints going from 0 to 100 and divided by 100. So 0 stays 0.0 and 100/100 becomes 1.0. Values in between, 0.1, 0.2, 0.3 may be slightly off though.

libmscore/glissando.cpp
ypoissant: Hm... That's the result of passing uncrustify with the config I found in the discussion about proposed new coding style. So I cannot trust this config. I'll fix it.

ypoissant: Something is wrong with the two commands you gave me for rebasing and pushing because every time I use it, it pushes the first commit and, thus, trashes the fix I just did.

ypoissant: I don't know how to undo that. GitKraken doesn't show this additional commit and my local source shows the fix is there.

ypoissant: I found this in Git push doc:
"-f
--force
Usually, the command refuses to update a remote ref that is not an ancestor of the local ref used to overwrite it. Also, when --force-with-lease option is used, the command refuses to update a remote ref whose current value does not match what is expected.

This flag disables these checks, and can cause the remote repository to lose commits; use it with care."

Jojo-Schmitz: Indeed after a rebase a `-push -f` is needed,

libmscore/easeInOut.cpp
njvdberg: Then let's hope the slider will return a "real" 0.0 then :wink:

njvdberg: The slider might fine, the <code>QDoubleSpinBox</code> isn't that trustworthy, is converts doubles to strings and back (always asking for problems :smiley: )

njvdberg (approved):

libmscore/easeInOut.cpp
ypoissant: I'm not using double spinbox but QSpinBox which returns an int.

PR #6870, opened 2020-11-15
fix 313004: don't reading sounding pitch for non-tranposing instruments

MarcSabatella:
Resolves: https://musescore.org/en/node/313004

A recent change to accessibility info (for status bar and
screen readers) caused the "sounding pitch" to be added
for all instruments, not just transposing ones.
It's important not to read this info when not needed,
because it slows down score reading by screen reader.

Fix is to simply check transposition before adding the info.


Marr11317:
there's a typo in the title...

vpereverzev (approved):

PR #6877, opened 2020-11-16
[MU3] Fulfilled the default style file field in preferences with Leland.mss

vpereverzev:
изображение


Marr11317:
what about style for part?

vpereverzev:
> what about style for part?

good question, actually

RobFog:
By "fulfil" you probably mean "fill out"?

vpereverzev:
@RobFog who knows :) ("fill out" much better actually)

vpereverzev:
@Marr11317 also updated that one, TY

Jojo-Schmitz:
That is not the solution IMHO. It should be applied if, and only if, being to so by that dialog that shws for Pre 3.6 scores, and should be the default settings for new scores,

mscore/prefsdialog.ui
```diff
@@ -90,7 +90,7 @@
       <string>Preferences Tab Manager</string>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>3</number>
```
Jojo-Schmitz, line 93: bad!!!

Jojo-Schmitz:
And please at least provide a 3.5 mss style file, as an option, next to the MuseJazz and Leland ones, like an Emmentaler.mss

PR #6885, opened 2020-11-17
New Export dialog

cbjeukendrup:
In this PR, I created a new Export dialog.
ezgif-4-723ab2f05a16

Design is made in collaboration with @Tantacrul. 

I made this dialog to make it easy for the user to select exactly which parts he/she wants to export. Formerly, it was rather unclear what the export options would do. Now, the behavior is more consistent and predictable. 

Furthermore, I added the export preferences to the dialog, so that they are more discoverable and easily accessible right when you need them. 

I also restructured the File menu a little; I put the options in a bit more sensible order. 

(On the page with MusicXML settings, a workaround was needed to make the text for the radio buttons wrap. I had to put the text in a separate label. I wanted keep the functionality of clicking the text to select the radio button, so I created a clickable label.)

Feedback is appreciated!
- Did I use the correct functions for translating?
- Haven't I confused `Score`/`MasterScore`/`Excerpt`/`Part`?

---


mscore/shortcut.cpp
```diff
@@ -114,17 +114,6 @@ Shortcut Shortcut::_sc[] = {
...
-         MsWidget::MAIN_WINDOW,
-         STATE_NORMAL | STATE_NOTE_ENTRY | STATE_EDIT | STATE_PLAY,
-         "file-part-export",
-         QT_TRANSLATE_NOOP("action","Export Parts…"),
-         QT_TRANSLATE_NOOP("action","Export parts"),
-         QT_TRANSLATE_NOOP("action","Save a copy of the score's parts in various formats"),
-         Icons::fileSave_ICON,
-         Qt::WindowShortcut,
-         ShortcutFlags::A_SCORE
-         },
```
Marr11317, line 127: If you remove it from here, it means it won't be assignable to a shortcut anymore.

cbjeukendrup: Yes, that is intended. The idea is that there will be a single action for export, which opens the Export dialog, where you can select whether you want the score and/or the parts.
(That is the action above the deleted one, so you don't see it directly in the GitHub diff.)

vpereverzev (approved):

PR #6888, opened 2020-11-17
Fix #313105: "No chord" MusicXML import

infojunkie:
Resolves: https://musescore.org/en/node/313105

This fix imports a MusicXML `harmony` element with `kind=none` to signify a "no chord" (invalid root) on the MuseScore sheet as per [forum discussion](https://musescore.org/en/node/313008).

### DIFF
--- a/importexport/musicxml/importmxmlpass2.cpp
+++ b/importexport/musicxml/importmxmlpass2.cpp
@@ -4944,14 +4944,11 @@ void MusicXMLParserPass2::harmony(const QString& partId, Measure* measure, const
                         if (_e.name() == "root-step") {
                               // attributes: print-style
                               step = _e.readElementText();
-                              /* TODO: check if this is required
-                              if (ee.hasAttribute("text")) {
-                                    QString rtext = ee.attribute("text");
-                                    if (rtext == "") {
+                              if (_e.attributes().hasAttribute("text")) {
+                                    if (_e.attributes().value("text").toString() == "") {
                                           invalidRoot = true;
+                                          }
                                     }
-                              }
-                               */
                               }
                         else if (_e.name() == "root-alter") {
                               // attributes: print-object, print-style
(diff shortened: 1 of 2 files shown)


importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -4940,15 +4940,12 @@ void MusicXMLParserPass2::harmony(const QString& partId, Measure* measure, const
                         if (_e.name() == "root-step") {
                               // attributes: print-style
                               step = _e.readElementText();
-                              /* TODO: check if this is required
-                              if (ee.hasAttribute("text")) {
-                                    QString rtext = ee.attribute("text");
-                                    if (rtext == "") {
+                              if (_e.attributes().hasAttribute("text")) {
```
MarcSabatella, line 4943: I suppose this means if the kind tag appears *before* the root-step tag, and there is no "none" on the root-step, we end up setting the root tpc after all.  That's probably a reasonable interpretation of a somewhat ambiguous spec anyhow.

infojunkie: Thinking more about it, I would have preferred to only rely on `kind=="none"` to signal an invalid root, since that much is clear. Happy to play it however you see fit though.

infojunkie: Anything more to do here on my side?

Jojo-Schmitz:
@lvinken Mind to have a look?

lvinken:
Will have a look.

importexport/musicxml/importmxmlpass2.cpp
lvinken: Kind before root-step would not be a valid MusicXML file, the schema does not allow that.
The root-step element is a child of the root element, which must be before the kind element.

lvinken:
Checked, looks fine to me.

PR #6442, opened 2020-08-12
fix #98816 - wrong tie direction in tabs

Git-Lior:
Resolves: https://musescore.org/en/node/98816
also related to: https://musescore.org/en/node/117351

the issues were caused by two problems:
1. the order of `ties` was based on the order of `notes` in the chord, which are ordered by `pitch` first and not `line`.
2. note in tabs need to be compared by `string` and not `line` (the value of which is `INVALID_LINE`).

the logic of deciding the tie direction stayed the same, but the variables changed like so:
* `tieNote` - the start note of the current tie
* `tieAbove` - the tie above the current tie (if null - the current tie is the top one)
* `tieBelow` - the tie below the current tie (if null - the current tie is the bottom one)

the procedure `compareNotesPos` (maybe needs to be renamed) decides the order of the notes,
such that `line` and `string` take priority over `pitch`, solving the case of ukulele in #98816.

before (ukulele tabs):

before-uke

the problem in the first bar was caused by the ukulele tuning as explained in #98816,
and the problem in the second bar was caused by note input as explained in the first comment of #117351.

after (ukulele tabs):

after-uke

I used the score in https://musescore.org/en/node/22459 to confirm that the normal cases of tie direction were
not affected by this change. result after the change:

after-tie-direction

for the CLA, my username in musescore.org is liorkl

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #5566, opened 2019-12-28
fix #277800: allow to preserve ties when replacing measures (by removing and inserting them)

Git-Lior:
Resolves: https://musescore.org/en/node/277800

This bug one of a set of bugs that are caused by a distinction that is not currently made between **"removing"** measures (deleting them and deleting everything connected to them) and **"detaching"** measures from the score (keeping ties and other spanners connected at one end, and only dissociating them from the detached measures). this distinction is important when performing an action like `rewriteMeasures`, which needs to only "detach" the old measures and insert new ones, preserving the outgoing ties and other spanners.

This solution is only for ties and not for other spanners, but it solves similar problems with ties described above (I also tested ties across split/join measures and it worked).

(for CLA, my musescore username is liorkl)

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #6889, opened 2020-11-19
[MU3] Porting some master changes/PRs to 3.x

Jojo-Schmitz:
* #6442 - [Fix #98816](https://musescore.org/en/node/98816) and [fix #117351](https://musescore.org/en/node/117351): Decouple tie direction from _notes order, and compare strings (not lines) in tabs
* #5566 - [Fix #277800](https://musescore.org/en/node/277800): Allow to preserve ties when replacing measures (by removing and inserting them)

### DIFF
--- a/libmscore/tie.cpp
+++ b/libmscore/tie.cpp
@@ -534,6 +534,17 @@ void Tie::write(XmlWriter& xml) const
 //   calculateDirection
 //---------------------------------------------------------
 
+static int compareNotesPos(const Note* n1, const Note* n2)
+{
+    if (n1->line() != n2->line()) {
+        return n2->line() - n1->line();
+    } else if (n1->string() != n2->string()) {
+        return n2->string() - n1->string();
+    } else {
+        return n1->pitch() - n2->pitch();
+    }
+}
+
 void Tie::calculateDirection()
       {
       Chord* c1   = startNote()->chord();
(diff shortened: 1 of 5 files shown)


vpereverzev:
vtest failure seems to be expected because of #6442

AntonioBL:
Actually, there are differences in the vtests due to this PR also in normal ties case. **In principle this should not have happened**.
Before:
tie-3-ref

And with this PR:
tie-3-1

vpereverzev:
> Actually, there are differences in the vtests due to this PR also in normal ties case. **In principle this should not have happened**.
> Before:
> tie-3-ref
> 
> And with this PR:
> tie-3-1

Oh, wow. It's a weird result, we need to fix this or even either revert these particular changes

libmscore/tie.cpp
```diff
@@ -563,31 +574,46 @@ void Tie::calculateDirection()
...
-                                    noteIdx = int(i);
+                              tiesCount++;
+                              int noteDiff = compareNotesPos(notes[i], tieNote);
+
+                              if (noteDiff > 0) {
+                                    if (!tieAbove)
+                                          tieAbove = notes[i];
+                                    else if (compareNotesPos(notes[i], tieAbove) < 0)
+                                          tieAbove = notes[i];
+                                    else if (noteDiff < 0) {
```
Git-Lior, line 592: `if (noteDiff > 0)` should close before the `else if (noteDiff < 0)` block starts

Git-Lior (commented):
@vpereverzev @AntonioBL I think the problem was caused by this error in porting

by the way @Jojo-Schmitz thank you for creating this PR, saved me some time 👍

vpereverzev:
@Git-Lior nice to see you around!

Are you able to prepare a quick-fix then?

libmscore/tie.cpp
Jojo-Schmitz: I might have removed a couple braces too much I guess?

Jojo-Schmitz:
Check #6892

PR #6459, opened 2020-08-16
[MU3] Fix #309192: Add advanced preference to remove letters from piano keyboard

Jojo-Schmitz:
plus a minor cleanup.

Resolves https://musescore.org/en/node/309192

PR #6890, opened 2020-11-19
[MU4] Fix #309192: Add advanced preference to remove letters from piano keyboard

Jojo-Schmitz:
plus a minor cleanup.

Resolves https://musescore.org/en/node/309192

master version of #6459


Jojo-Schmitz:
Apparently this has gotten removed from Mu4 again? Why?

PR #6891, opened 2020-11-19
Improves recognition of instrument when loading a pre-3.6 score.

njvdberg:
Resolves: https://trello.com/c/lXc1BqEB/18-imagepng

In case the instrument in the score file has <ocde>MusicXMLid</code> with value <code>strings.group</code>, the value of controller 132 is used to  find the correct strings group. This prevents all <code>Violons</code>, <code>Violas</code>, <code>Violoncellos</code> and <code>Contrabasses</code> are mapped onto <code>Strings</code> which causes wrong brackets to be generated.

### DIFF
--- a/libmscore/instrument.cpp
+++ b/libmscore/instrument.cpp
@@ -1520,14 +1520,44 @@ void Instrument::updateInstrumentId()
       if (!_id.isEmpty() || _instrumentId.isEmpty())
             return;
 
+      // When reading a score create with pre-3.6, instruments doesn't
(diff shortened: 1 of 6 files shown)


libmscore/instrument.cpp
```diff
@@ -1520,14 +1520,44 @@ void Instrument::updateInstrumentId()
       if (!_id.isEmpty() || _instrumentId.isEmpty())
             return;
 
+      // When reading a score create with pre-3.6, instruments doesn't
```
Marr11317, line 1523: created?

PR #6893, opened 2020-11-20
[MU4] Fixed settings init

RomanPudashkin:
Init of settings and init of modules depend on each other cyclically: for correctly init of settings need to call addItem from modules, but for correctly init of modules need to init settings before. I refactored settings to break that dependency. Now settings will be inited at first and after that will be inited modules using already ready settings

- moved init of settings before init of modules as was before
- refactored settings to remove dependency from init of modules

### DIFF
--- a/framework/global/settings.cpp
+++ b/framework/global/settings.cpp
@@ -18,10 +18,10 @@
 //=============================================================================
 
 #include "settings.h"
-#include <QSettings>
-
 #include "config.h"
 #include "log.h"
+
+#include <QSettings>
 
 using namespace mu;
 using namespace mu::framework;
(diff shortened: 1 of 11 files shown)


framework/global/settings.cpp
```diff
@@ -48,127 +48,122 @@ Settings::~Settings()
...
-        Item& item = it->second;
-        item.val = Val::fromQVariant(m_settings->value(key));
-        item.val.setType(item.defaultVal.type());
-    }
+    // TODO: implement writing/reading first part of key (module name)
+    m_settings->setValue(QString::fromStdString(key.key), value.toQVariant());
 }
 
-Val Settings::value(const Key& key) const
+void Settings::setDefaultValue(const Key& key, const Val& value)
```
Eism, line 123: needed to send notifications about change default value

RomanPudashkin: There is no channel like "defaultValueChanged", so I don't think that need send any notification from this method at the current moment

Eism (approved):

PR #6894, opened 2020-11-20
[MU3] Score symbols bounding boxes

vpereverzev:
resolves - https://trello.com/c/OxOtQn1j

изображение
each score symbols has his own bounding box, but this is not taken into account when rendering.

before:
изображение

after:
изображение


Jojo-Schmitz:
I'd guess we'd have to expect quite many vtest differences

Jojo-Schmitz:
This difference is a bug, before:
articulation-1-ref
after
articulation-1-1

vpereverzev:
> I'd guess we'd have to expect quite many vtest differences
> 
> And indeed thare are.
> 
> This difference certainly is a bug, before:
> articulation-1-ref
> after
> articulation-1-1
> Note that flag for the 5th 8th
> There are multiple examples for this, also this, before:
> bravura-1-ref
> after:
> bravura-1-1
> Which shows a 2nd bug even, way too long stems.

Yes, I've also found out that recently, working on fix

cbjeukendrup:
I think the stem lengths are fine, but it's just the flag that appears too high. That makes it look like the stem is too long, but I don't think that is the case. (If you look closely at the picture, there is a small edge where the stem ends, and that seems to be at the correct height.)

Jojo-Schmitz:
Yes, you're most probably right. I guess the bounding box of the innermost flag is taken rather than the one of the outermost?

libmscore/chord.cpp
```diff
@@ -1281,7 +1281,7 @@ qreal hookAdjustment(QString font, int hooks, bool up, bool small)
       {
       bool fallback = MScore::useFallbackFont && (hooks > 5);
 
-      if (font == "Emmentaler" && !fallback) {
+      if (font == "Leland" || font == "Emmentaler" && !fallback) {
```
Jojo-Schmitz, line 1284: Not sure about the order of evaluation here, would prefere parens to make it if clearer

vpereverzev: good catch, I've forgotten to remove that one, TY

PR #6902, opened 2020-11-21
[MU4] Disabled use std filesystem (fix compiling with MinGW)

igorkorsukov:
Disabled use std filesystem (fix compiling with MinGW)

### DIFF
--- a/thirdparty/haw_logger/logger/logdefdest.cpp
+++ b/thirdparty/haw_logger/logger/logdefdest.cpp
@@ -7,15 +7,16 @@
     #define USE_QT_DIR
     #include <QDir>
 #else
-    #if __has_include(<filesystem>)
-        #include <filesystem>
-        namespace fs = std::filesystem;
-    #elif __has_include(<experimental/filesystem>)
-        #include <experimental/filesystem>
-        namespace fs = std::experimental::filesystem;
-    #else
-        #error compiler must either support c++17
-    #endif
+//    #if __has_include(<filesystem>)
+//        #include <filesystem>
+//        namespace fs = std::filesystem;
+//    #elif __has_include(<experimental/filesystem>)
+//        #include <experimental/filesystem>
+//        namespace fs = std::experimental::filesystem;
+//    #else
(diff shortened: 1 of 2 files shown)


thirdparty/haw_logger/logger/logdefdest.cpp
```diff
@@ -7,15 +7,16 @@
...
+//    #if __has_include(<filesystem>)
+//        #include <filesystem>
+//        namespace fs = std::filesystem;
+//    #elif __has_include(<experimental/filesystem>)
+//        #include <experimental/filesystem>
+//        namespace fs = std::experimental::filesystem;
+//    #else
+//        #error compiler must either support c++17
+//    #endif
+#error At moment supported only with Qt
```
Jojo-Schmitz, line 19: With Qt?

igorkorsukov:
Known Issues

* MacOS 10.15 XCode 10.3 on CI not has `<filesystem>`
* MacOS 10.14 has `<filesystem>` (pass check `#if __has_include(<filesystem>)`) but doesn't have implementation (path, exists) - very strange 
* gcc 7 (on CI, default for ubuntu 18.10) has `<experimental/filesystem>` and needs link `stdc++fs` lib (work fine)
* mingw has `<filesystem>`, but doesn't have `stdc++fs` lib

see also https://stackoverflow.com/questions/54290254/problem-adding-stdfilesystem-to-cmake-project

thirdparty/haw_logger/logger/logdefdest.cpp
igorkorsukov: yes, at the moment using QDir   
(The original idea was not to depend on Qt)

igorkorsukov: needs to define HAW_LOGGER_QT_SUPPORT and have QDir 
(we have it)

Jojo-Schmitz: Ah, I see (now)

PR #6810, opened 2020-11-04
[MU3] Fix #311986: handle duplicate voltas

jeetee:
Resolves: https://musescore.org/en/node/311986

Implemented preprocessing of voltas to linearize overlaps. In case of overlaps the voltas are split into smaller ones with the overlap using cross-section of both repeatlists as its repeatlist.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [(N/A for 3.x)] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #6904, opened 2020-11-21
[MU4] Fix #311986: handle duplicate voltas

jeetee:
Resolves: https://musescore.org/en/node/311986

Cherry pick of PR #6810 from 3.x to master

### DIFF
--- a/libmscore/repeatlist.cpp
+++ b/libmscore/repeatlist.cpp
@@ -21,7 +21,7 @@
 #include "types.h"
 #include "volta.h"
 
-#include <stack>
+#include <list>
 #include <utility> // std::pair
 
 namespace Ms {
@@ -266,7 +266,7 @@ void RepeatList::flatten()
     do {
         s->addMeasure(m);
         m = m->nextMeasure();
-    }while (m);
+    } while (m);
     push_back(s);
 
     _expanded = false;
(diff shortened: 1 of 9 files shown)


vpereverzev:
https://github.com/musescore/MuseScore/pull/6904/checks?check_run_id=1436848735
code style check failure, should be easy to fix

jeetee:
I don't have uncrustify locally, and QtCreator seemed happy. Especially those multiline function arguments is what it does by default when using the Qt codestyle.
Do we have a style file for QtCreator as we did for 3.x ?

I'll apply those changes but FWIW I'm rather against the requirement to place the pointer indicator concatenated with the type name (being fully aware that this is somewhat a "holy war" point for many). My arguments against it are:
1. A T* itself isn't a type, the '*' itself is the type
2. It breaks style as soon as const comes into play, something that is very important to convey code intent
3. It completely break style when volatile comes into play (admittedly, as we're not doing hardware interfacing as much, this is probably a moot worry in the MS codebase).
4. (I might be biased here because of this) Coding safety standards such as the Barr standard call for the space to be there as a lexical safety rule.
Note that this only applies to the pointer type, not to the reference indicator, which should be concatenated.

jeetee:
@vpereverzev ready

PR #6912, opened 2020-11-23
[MU4] Fix compiler warnings

Jojo-Schmitz:
reg. unused variables, parameters and deprecations

### DIFF
--- a/libmscore/repeatlist.cpp
+++ b/libmscore/repeatlist.cpp
@@ -389,11 +389,11 @@ void RepeatList::collectRepeatListElements()
                         remainder = swap;
                     }
                     // Cross-section of the repeatList
-                    std::list<int> endings = remainder->endings().toStdList();
+                    std::list<int> endings(remainder->endings().begin(), remainder->endings().end());
                     endings.remove_if([&volta](const int& ending) {
                             return !(volta->hasEnding(ending));
                         });
-                    remainder->setEndings(QList<int>::fromStdList(endings));
+                    remainder->setEndings(QList<int>(endings.begin(), endings.end()));
                     // Split and merge done
                     preProcessedVoltas.push_back(remainder);
                     if (volta->endMeasure() != remainder->endMeasure()) {
--- a/mu4/importexport/internal/musicxml/importmxmlpass2.cpp
+++ b/mu4/importexport/internal/musicxml/importmxmlpass2.cpp
@@ -2197,7 +2197,7 @@ void MusicXMLParserPass2::attributes(const QString& partId, Measure* measure, co
         } else if (_e.name() == "key") {
             key(partId, measure, tick);
         } else if (_e.name() == "measure-style") {
-            measureStyle(partId, measure);
+            measureStyle(measure);
         } else if (_e.name() == "staff-details") {
             staffDetails(partId);
         } else if (_e.name() == "time") {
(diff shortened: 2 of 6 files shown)


libmscore/repeatlist.cpp
```diff
@@ -389,11 +389,11 @@ void RepeatList::collectRepeatListElements()
                         remainder = swap;
                     }
                     // Cross-section of the repeatList
-                    std::list<int> endings = remainder->endings().toStdList();
+                    std::list<int> endings(remainder->endings().begin(), remainder->endings().end());
```
Jojo-Schmitz, line 392: @jeetee please check this one here

libmscore/repeatlist.cpp
```diff
@@ -389,11 +389,11 @@ void RepeatList::collectRepeatListElements()
...
                     endings.remove_if([&volta](const int& ending) {
                             return !(volta->hasEnding(ending));
                         });
-                    remainder->setEndings(QList<int>::fromStdList(endings));
+                    remainder->setEndings(QList<int>(endings.begin(), endings.end()));
```
Jojo-Schmitz, line 396: and this too

mu4/importexport/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -2362,6 +2362,7 @@ void MusicXMLParserPass2::staffTuning(StringData* t)
 
 void MusicXMLParserPass2::measureStyle(const QString& partId, Measure* measure)
 {
+    Q_UNUSED(partId);
```
RomanPudashkin, line 2365: Please, remove this argument from the method signature

Jojo-Schmitz: That's the other solution, yes. I though that maybe @lvinken has further plans with this, so left it in for now.

libmscore/repeatlist.cpp
jeetee: Agreed, master only as per https://doc.qt.io/qt-5/qlist.html#fromStdList

libmscore/repeatlist.cpp
jeetee: OK as per https://doc.qt.io/qt-5/qlist.html#fromStdList

libmscore/repeatlist.cpp
jeetee: And more importantly, according to https://www.cplusplus.com/reference/list/list/list/ variant 3

mu4/importexport/internal/musicxml/importmxmlpass2.cpp
RomanPudashkin: Also, you should use our UNUSED macro (defined in log.h) in mu4. And please, don't add ; because macro already contains ; inside it and this can be a reason for a new warning :)

Jojo-Schmitz: Well, QtCreator warns about the missing semicolon IIRC

lvinken: I'm OK with removing the argument. As of now, I have no recollection of any plan to use it and if I need it in future I'll simply add it back myself.

PR #6917, opened 2020-11-23
[MU3] Paren time sig

Marr11317:
Resolves: https://musescore.org/en/node/313219#comment-1042033

NOTE: This PR needs elerouxx's PR to be merged first (#6550 )

MuseJazz includes parentheses by default in its time signatures. With elerouxx's PR #6550 it is now possible to append them manually. As such, new scores created with 3.6 will not have this trailing parenthesis by default. 

The change to MuseJazz hasn't been done yet, but because I have no development machine right now, I am testing with the build here and the artifacts generated while I remove the MuseJazz Parentheses.


Marr11317:
@elerouxx do you have any clue why [these changes](https://github.com/musescore/MuseScore/pull/6917/commits/ac39238619f1f55d6f48ffe27a63e586170854cc#diff-99314dfd537ef914e998d85f1ad9531863797db00952c3c70aceedfc672ecd97R261) do not set the parser string?

elerouxx:
> @elerouxx do you have any clue why [these changes](https://github.com/musescore/MuseScore/pull/6917/commits/ac39238619f1f55d6f48ffe27a63e586170854cc#diff-99314dfd537ef914e998d85f1ad9531863797db00952c3c70aceedfc672ecd97R261) do not set the parser string?

You mean timeSig.cpp @ 261? I don't think the (old) variable is getting set true for these types. This variable seems to be for a very old time signature implementation that had a sequence of nominators. (additive?).

Marr11317:
okay, this is now ready. As soon as elerouxx's PR is merged this one can be too. (or both can be merged at the same time with this PR)

So here's how it works: When older scores are loaded, if these scores are using MuseJazz, then the symbols are corrected.

Jojo-Schmitz:
The vtest fails are actually fixes, before:
musejazz-4-ref
emmentaler-4-ref
after:
emmentaler-4-1
musejazz-4-1

vpereverzev:
3.x merges are closed

PR #6918, opened 2020-11-23
[MU4] Fix #312489 Harmony duration interpretation

jeetee:
Resolves: https://musescore.org/en/node/312489

Cherry pick of PR #6857 from 3.x to master


jeetee:
Here is a perfect example of the lexical wrongness of pointer notation combined with const:
Original code= `Measure const * abc` (= a pointer to a const Measure)
"Corrected"= `Measure const* abc` (= quickly misgrouped and misread as a const* to a Measure, which it isn't)

PR #6932, opened 2020-11-24
[MU4] Fixed build logger

igorkorsukov:
My error.
But I'm confused, what not one compiler  (gcc 7, msvc 2019, clang 10, mingw 8)  notified me about this error.


Jojo-Schmitz:
New compilers -> new warnings and errors.... Quite often it is more difficult to port code from one compiler to another than from one platform (Operating system) to another (still using the same compiler)

PR #6939, opened 2020-11-25
Fix #313613: Note value shortcuts on text input

Asmatzaile:
Resolves: https://musescore.org/en/node/313613

Shortcuts for note durations would allow changing metronome marks more easily.

- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)

### DIFF
--- a/libmscore/textedit.cpp
+++ b/libmscore/textedit.cpp
@@ -425,7 +425,37 @@ bool TextBase::edit(EditData& ed)
                         case Qt::Key_H:
                               s = "\u266e"; // Unicode natural
                               break;
-                         case Qt::Key_Space:
+                        case Qt::Key_1:
+                              s = "\u1d163"; // Unicode 64th note
+                              break;
+                        case Qt::Key_2:
+                              s = "\u1d162"; // Unicode 32th note
+                              break;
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
I'd rather have those shortcuts configurable rather than hardcoded. IMHO we should restrict the hardcoded shortcuts down to the absolute minimum.

njvdberg:
I agree having configurable shortcuts is much better, these hardcoded shortcuts might conflict with other defined shortcuts.

Jojo-Schmitz:
as @mike320 pointed out in the forum and issue tracker, these here are valid and usud only during text editing, so should not conflict with other shortcuts

Jojo-Schmitz:
It is not quite clear though whether it it is wise to use the same numbers here as being used by default for note durations at note input, as those are changeable, whereas these here are not?

Asmatzaile:
I agree that those shortcuts, although maybe predefined, should be configurable. I am thus closing the PR because idk how to make the textedit.cpp file know about the shortcut.h file.

PR #6942, opened 2020-11-26
[MU4] Added gen dump symbols for MacOS on CI

igorkorsukov:
Added gen dump symbols for MacOS on CI

### DIFF
--- a/build/ci/macos/dumpsyms.sh
+++ b/build/ci/macos/dumpsyms.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+echo "Generate dump symbols"
+
+GEN_SCRIPT=tools/crashdump/generate_syms.sh
+DUMPSYMS_BIN=$HOME/breakpad/dump_syms
+ARTIFACTS_DIR=build.artifacts
+BUILD_DIR=applebuild
+SYMBOLS_DIR=$ARTIFACTS_DIR/symbols
+MSCORE_BIN=$BUILD_DIR/mscore.app/Contents/MacOS/mscore
+
+echo "GEN_SCRIPT: $GEN_SCRIPT"
+echo "DUMPSYMS_BIN: $DUMPSYMS_BIN"
+echo "BUILD_DIR: $BUILD_DIR"
+echo "SYMBOLS_DIR: $SYMBOLS_DIR"
+echo "MSCORE_BIN: $MSCORE_BIN"
+
+$GEN_SCRIPT --dumpsyms-bin $DUMPSYMS_BIN --build-dir $BUILD_DIR --symbols-dir $SYMBOLS_DIR --mscore-bin $MSCORE_BIN
+
+echo "-----"
+ls $SYMBOLS_DIR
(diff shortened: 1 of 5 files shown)


build/ci/macos/dumpsyms.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "GEN_SCRIPT: $GEN_SCRIPT"
+echo "DUMPSYMS_BIN: $DUMPSYMS_BIN"
+echo "BUILD_DIR: $BUILD_DIR"
+echo "SYMBOLS_DIR: $SYMBOLS_DIR"
+echo "MSCORE_BIN: $MSCORE_BIN"
+
+$GEN_SCRIPT --dumpsyms-bin $DUMPSYMS_BIN --build-dir $BUILD_DIR --symbols-dir $SYMBOLS_DIR --mscore-bin $MSCORE_BIN
+
+echo "-----"
+ls $SYMBOLS_DIR
```
Jojo-Schmitz, line 21: Seems this is where the CI errors out? If that is not intended an 'exit 0` after the 'ls` should help

igorkorsukov: error with the wrong path to binary mscore
I added debug code, to find the error

build/ci/macos/dumpsyms.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+echo "Generate dump symbols"
+
+GEN_SCRIPT=tools/crashdump/generate_syms.sh
+DUMPSYMS_BIN=$HOME/breakpad/dump_syms
+ARTIFACTS_DIR=build.artifacts
+BUILD_DIR=build.release
+SYMBOLS_DIR=$ARTIFACTS_DIR/symbols
+MSCORE_BIN=$BUILD_DIR/main/mscore.app/Contents/MacOS/mscore
```
Jojo-Schmitz, line 10: Shouldn't that be `MSCORE_BIN=$BUILD_DIR/main/mscore.app/Contents/MacOS/bin/mscore`?

build/ci/macos/dumpsyms.sh
Jojo-Schmitz: It still fails though at that `ls`, causing the script to end with a non-zero exit status

build/ci/macos/dumpsyms.sh
Jojo-Schmitz: Instead of all those `echo`s you could use `set -x` and see every command and every variable being printed

igorkorsukov: The fact is that I need to find out not where the error is (where the error I know is not the right path). I need to find the correct path to the compiled binary file (locally on my machine, I have it different, not the same as on the server).

I changed the path to the path to the binary file in the installation directory (and not in the build directory, as I originally wanted), this will also work.

igorkorsukov: I know that: 
* build.release/main/mscore.app/Contents/MacOS/mscore - path in the build dir to mscore binary on my local machine   
* applebuild/mscore.app/Contents/MacOS/mscore - path in the install dir to mscore binary on CI   

But I do not know what is the path to the ьscore binary in the build dir on CI.
It looks like this -  build.release/main/mscore.build/Release/mscore.app/Contents/MacOS/mscore 
But until the end I did not find out, I decided to specify the path from the install dir, this will also work.

igorkorsukov: I understood your advice about set -x.  Thanks :)

PR #6593, opened 2020-10-01
Restructure Palettes buttons and add "Expand/Collapse All" buttons

cbjeukendrup:
Palettes Expand-Collapse All - by cbjeukendrup

This is what I did:
- Moved the "Add Palettes" button to the right of the search field;
- Replaced the button text with a '+' icon;
- Made the search field a bit wider;
- Added an extra '...' button at the right;
- Created a popup for the new button;
- Filled the popup with "Expand/Collapse All" buttons;
- Moved the "Single Palette" option to the new popup;
- Removed the context menu formerly holding "Single Palette";
- Tested thoroughly on my Mac.

A MuseScore user and friend of mine requested this feature, so I started looking if I could implement it. I also had some correspondence about this with Tantacrul in the MuseScore Developer Chat, and he agreed with the changes. 
By the way: this is the very first time that I'm using Qml (and sort of my first real contribution for MuseScore), so I'll be welcoming any feedback!

Maybe this PR is something for MuseScore 3.6, because it's not really a breaking change, but it involves some string translations. 

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [N/A] I created the test (mtest, vtest, script test) to verify the changes I made

PR #6944, opened 2020-11-26
Make 'Add Palettes' button more prominent

cbjeukendrup:
Resolves: https://trello.com/c/vncptyKX/1-this-change-to-palettes-needs-to-be-reverted

In PR #6593, I made the 'Add Palettes' button smaller to give the Search field more space. 
The 'Add Palettes' button is very important though and should be more prominent, so in this PR I give it the space it deserves. I created a small Search button, and when you click it, the other buttons are hidden and the Search field appears. This is inspired by the design for MuseScore 4.

Here you see it in action:
image

---


MarcSabatella:
How does this affect the palette search keyboard command?  Hopefully, it still transfers keyboard focus to the search box and automatically displays it, so the existing workflow for keyboard users (including blind users) is not adversely affected.

cbjeukendrup:
When you press F9 to open the Palettes, the search field is shown and focused. 
If the search field is focussed and you press Tab, the 'X' button ('Stop search') gets focused, and after another Tab the list of palettes gets focused. 
If the search field is not shown, the Tab cycle is as follows:
Add Palettes -> Search button (magnifying glass) -> Options button ('...') -> Palettes list.

MarcSabatella:
That much sounds good, but what the search palette keyboard command?  Not F9 to open/close the palette, but Ctrl+F9 to transfer control directly to the search box.  It is important this doesn’t take additional keystrokes and that it doesn’t behave differently depending on whether the search box is currently open or not (a blind user won’t be aware of this).

cbjeukendrup:
Ah, didn't know that command, but it also works: the search field is shown and focused. In fact, I updated the function which was responsible for focusing the search field, so that it also makes it visible when it was hidden. 

One thing that I'm not sure about, is the following: 
Formerly, when the search field was focussed and you pressed Tab, the 'X' button would be skipped and the focus would move directly to the Palettes list. So in fact, the 'X' button would never get focused.
In the current situation, the 'X' button is essential to close the search field, so it needs to get focused at some time. So I decided to make the order as I mentioned in my previous comment. Do you think this is problematic?

MarcSabatella:
At one time, I thought we had tuned the down arrow to go directly to the first search result, bypassing the palette name etc.  I think maybe that went away with the QML rewrite, but also maybe it had to do with organizing the search results?  My memory is very fuzzy.  Anyhow, the current behavior of down arrow is pretty decent, it goes first to the palette name, then the first element, skipping the "..." menu for the palette.  So that would be nice to preserve.  Tab hits everything (including, now, the "X"), but Down hits only the essentials.

MarcSabatella:
Now that I use this, I'm happy with most things except for one: the "X" within the search box shouldn't close the box but merely clear the search field.  There needs to be another way to close the box, I think.  See https://musescore.org/en/node/313776 - it's nice that these changes fixed a bug with the "X" in previous versions, but the current behavior is not a step forward for that particular use case.

Tantacrul:
@MarcSabatella - I'd like to see this released in beta. I suspect this might not be an issue but if it is, we'll be told about it soon enough.

MarcSabatella:
That's fine.  I don't personally use that X so much, but I know there are people who have come to rely on the palette search as a really efficient way of adding symbols using keyboard only.  I'm not talking blind users so much as sighted power users.  And based on the amount of comments I've seen about the exact semantics of the "X", I'm guessing the vast majority won't care but there will be 2-3 who complain bitterly and say they can't update because of this.

Hmm, also, FWIW, it's possible that everything about this new flow might break Philip Rothman's "Notation Express" gizmo.  Not that this is a deal-breaker, but it could be worth a heads up to him.

henkdegroot:
@MarcSabatella @Tantacrul 
I have noticed this change in the 3.x branch and compiled on my windows setup. I click the magnify button and the search box shows with the X. However I can continue to work in my score and other function, while the search box remains open.
The 'Add Palettes' button only reveals itself again after clicking the X. CTRL+F9 does not seem to work to set focus to the search box.
Perhaps 'ESC' should be an active key to exit/clear the search box? Not sure if that will cause other issues, just a thought.

Tantacrul:
Definitely agree on the 'esc.' button idea. I was quite rushed, so never
thought to test it with shortcuts.

cbjeukendrup:
OK, I will make the Esc button do the same as the X. I was planning some minor improvements anyway, so I will do this too in one run. 

I'm not experiencing problems with Ctrl F9 (or actually the Mac equivalent Cmd+Fn+F9). For me, it does set focus to the search box and I can start to type. 

That you can continue working in the score when the search field is open, is actually by design (at least, how I meant it to work). @Tantacrul what do you think, should that be different?

Tantacrul:
If we prevented you using the score when searching, then I don't think it
would be possible to apply anything from the palettes menu, making it a bit
redundant.

So, yes. By design. Interested to get user feedback about it. It will be
MUCH better in MS4, when it searches all palettes, not just the ones you
have opened.

M

henkdegroot:
@Tantacrul  not sure what you mean with the search in MS4 being better. Currently it already searches through all palettes. Or perhaps you mean it will be searching also through palettes not enabled in the view?

@cbjeukendrup. Ctrl F9 does not work on my windows system. Not in the self compiled 3,x and also not in 3.5.2.

henkdegroot:
@cbjeukendrup , Ctrl + F9 is working after doing a factory reset. When I checked my shortcuts list, I did not see an entry for the Search in Palets and now there is. Not sure why it was not in the list, maybe it was in the list and I did not see it or a translation issue (I am using Dutch text) and some items still show in English, even items that have been translated a while ago. Of course new features are English only as translation has not started yet. Bottom line, ignore the Ctrl+F9.

Tantacrul:
> @Tantacrul not sure what you mean with the search in MS4 being better. Currently it already searches through all palettes. Or perhaps you mean it will be searching also through palettes not enabled in the view?

-Yes, that's what I mean.

MarcSabatella:
> @cbjeukendrup , Ctrl + F9 is working after doing a factory reset. When I checked my shortcuts list, I did not see an entry for the Search in Palets and now there is.

That's normal, I think, if you have previously customized your shortcuts - newly added shortcuts won't show up because we use your customized shortcut instead.  In the future, instead of a full factory reset, just reset that one shortcut to its default.  The Cltr+F9 was added around 3.3 I think, presumably your custom shortcuts predated that and were missing other newly added shortcuts as well.

Probably we should figure out a way to merge new additions with saved custom shortcuts.

henkdegroot:
"Probably we should figure out a way to merge new additions with saved custom shortcuts."

Sounds like an excellent feature to me. As a normal user would not be aware of new standard shortcuts introduced in the software and therefore will most likely not use them or report a potential bug/defect. Perhaps highlighting custom shorts in the list could help as well, so you can easily review what custom shortcuts you have setup and/or were is does not match the standard (if needed).

PR #6947, opened 2020-11-26
New default canvas background color

cbjeukendrup:
Resolves: https://trello.com/c/yhs0S2ZT/8-propose-new-slightly-different-default-colours-for-canvas-and-bg-to-reduce-the-harshness-of-the-score-appearance

Implements the new default canvas background color which @Tantacrul suggested, to reduce the harshness of the score appearance. 

(et cetera)


IsaacWeiss:
We have options other than solid colors... to me the solid blue looks extremely harsh.
image
image

Tantacrul:
> We have options other than solid colors... to me the solid blue looks extremely harsh.
> image
> image

There's always going to be variance in taste.
 
Sit with it for a while. I had it this way for a few days and it's quite nice to work with. BTW: textured backgrounds are a little outdated, and I really think the app needs a bit of colour!

I'm sure I've not heard the last of this.... :)

PR #6948, opened 2020-11-26
[MU3] Make AutoSize default for new scores but disable it for older scores.

njvdberg:
Resolves: https://trello.com/c/11WgLsL1/23-disabling-vertical-frame-enable-auto-size-is-not-saved-in-score-file

In the <code>VBox::VBox()</code> constructor, the variable <code>_isAutoSizeEnabled</code> was always set to <code>true</code> for score with a version number >= 3.02. However, the default for that property is <code>false</code>. So, when this option was switched of, the property has the default value and isn't written into the score file. Next, when reading the file, there no such property but if the score version was >= 3.02, the property was set to <code>true</code> again.

This is solved by setting the default of <code>_isAutoSizeEnabled</code> to <code>true</code>. But when reading an score < 3.02, this property in set to <code>false</code> after reading the properties in <code>Box::read()</code>. Also, <code>Box::getProperty()</code> will always return <code>false</code> for older scores.

### DIFF
--- a/libmscore/box.cpp
+++ b/libmscore/box.cpp
@@ -224,6 +224,8 @@ void Box::read(XmlReader& e)
       _boxHeight       = Spatium(0);     // override default set in constructor
       _boxWidth        = Spatium(0);
       MeasureBase::read(e);
+      if (score()->mscVersion() < 302)
+            _isAutoSizeEnabled = false; // disable auto-size for older scores by default.
       }
 
 //---------------------------------------------------------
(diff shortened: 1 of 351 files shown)


RobFog:
"disables" should "disable"?

Jojo-Schmitz:
On vtest diff, before:
mmrest-6-ref
after
mmrest-6-1

Jojo-Schmitz:
That vtest fail might be easily fixed by updating the scorefile it stems from I guess?

njvdberg:
The vtest violation was caused by reading a 2.x file. In both <code>read114.cpp</code> and <code>read206.cpp</code> the auto resize had to be disabled explicitly.

Jojo-Schmitz:
Now, on top of the vtest fails, mtests fail too

njvdberg:
> Now, mtests fail (and macos uploading artifacts)

The mtest was my mistake, the files got a new tag.
But why uploading artifact fails every now and then puzzles me. It seems to something in the system, not in the sources.

Jojo-Schmitz:
yes, likely not your fault

PR #6955, opened 2020-11-27
fix #292387 : [MusicXML] Drumset of MusicXML sounds pitch

AntonioBL:
Resolves: https://musescore.org/en/node/292387

This PR sets the channel bank to 128 before setting the first instrument, which is then added to the masterScore midiMapping (and the channel is cloned to midiMapping articulation channel).
Bank 128 is thus set for both the masterChannel and the articulation channel inside midiMapping

### DIFF
--- a/importexport/musicxml/importmxmlpass2.cpp
+++ b/importexport/musicxml/importmxmlpass2.cpp
@@ -473,13 +473,17 @@ static Instrument createInstrument(const MusicXMLInstrument& mxmlInstr, const In
 //   updatePartWithInstrument
 //---------------------------------------------------------
 
-static void updatePartWithInstrument(Part* const part, const MusicXMLInstrument& mxmlInstr, const Interval interval)
+static void updatePartWithInstrument(Part* const part, const MusicXMLInstrument& mxmlInstr, const Interval interval, const bool hasDrumset = false)
       {
-      const Instrument instr = createInstrument(mxmlInstr, interval);
+      Instrument instr = createInstrument(mxmlInstr, interval);
+      if (hasDrumset)
+            instr.channel(0)->setBank(128);
       part->setInstrument(instr);
-      if (mxmlInstr.midiChannel >= 0) part->setMidiChannel(mxmlInstr.midiChannel, mxmlInstr.midiPort);
+      if (mxmlInstr.midiChannel >= 0)
+            part->setMidiChannel(mxmlInstr.midiChannel, mxmlInstr.midiPort);
       // note: setMidiProgram() does more than simply setting the MIDI program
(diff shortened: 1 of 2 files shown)


AntonioBL:
@dmitrio95 : Sorry to disturb you, but you are the expert of the midiMapping; can you please have a look at the related comment https://musescore.org/en/node/292387#comment-1042795 ?

Additional note: I think the line https://github.com/musescore/MuseScore/blob/28f7e1e28e72a4d9fea8ffa5c7f47a7c13378e68/mu4/importexport/internal/musicxml/importmxmlpass2.cpp#L513 becomes redundant with the change of this PR, but I am not 100% sure.

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -473,13 +473,17 @@ static Instrument createInstrument(const MusicXMLInstrument& mxmlInstr, const In
...
+      if (hasDrumset)
+            instr.channel(0)->setBank(128);
       part->setInstrument(instr);
-      if (mxmlInstr.midiChannel >= 0) part->setMidiChannel(mxmlInstr.midiChannel, mxmlInstr.midiPort);
+      if (mxmlInstr.midiChannel >= 0)
+            part->setMidiChannel(mxmlInstr.midiChannel, mxmlInstr.midiPort);
       // note: setMidiProgram() does more than simply setting the MIDI program
-      if (mxmlInstr.midiProgram >= 0) part->setMidiProgram(mxmlInstr.midiProgram);
+      if (mxmlInstr.midiProgram >= 0)
+            part->setMidiProgram(mxmlInstr.midiProgram);
```
dmitrio95, line 486: This doesn't seem to be an issue for this pull request, but `part->setMidiProgram()` also sets bank to a value which is 0 by default:
https://github.com/musescore/MuseScore/blob/4983887d71bbab2197e00717a2eec179a009d2c1/libmscore/part.h#L108

So with this PR `updatePartWithInstrument` effectively sets the channel's bank to 128, adds the channel to MIDI mapping (via `setMidiChannel()`) and then sets the channel's bank value back to 0. Here it doesn't seem to cause any issues (although it doesn't seem right anyway),  but in other places (for example, in [OVE import](https://github.com/musescore/MuseScore/blob/4983887d71bbab2197e00717a2eec179a009d2c1/importexport/ove/importove.cpp#L623-L624)) it might cause undesirable results. So maybe `setMidiProgram()` behavior should be somehow reconsidered, although this certainly doesn't belong to this PR.

AntonioBL: Thank you very much for your comment. 

So line `part->instrument()->channel(0)->setBank(128); ` is not actually redundant, in the end.

PR #6967, opened 2020-11-29
 Fix #86376, fix #313736: Additions to F2 Special Characters > Common Symbols

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/86376 and https://musescore.org/en/node/313736
On top add the most common currency symbols, $, €, ¢, £ and ¥

Except for indenting changes this should apply to master also.


Jojo-Schmitz:
wrong branch

PR #6970, opened 2020-11-29
[MU3] Fix AppImage nightly autoupdates and build scripts

shoogle:
@igorkorsukov, I tried updating with [AppImageUpdate](https://github.com/AppImage/AppImageUpdate/releases) but it didn't work so I downloaded the latest zsync file and examined it:

```
$ wget https://ftp.osuosl.org/pub/musescore-nightlies/linux/3x/nightly/MuseScoreNightly-latest-x86_64.AppImage.zsync
$ head MuseScoreNightly-latest-x86_64.AppImage.zsync
zsync: 0.6.2
Filename: MuseScore-x86_64.AppImage
MTime: Sun, 29 Nov 2020 04:34:26 +0000
Blocksize: 4096
Length: 154656744
Hash-Lengths: 2,2,5
URL: MuseScore-x86_64.AppImage
SHA-1: 915cc5b7410607ecc2fc43631861ba184fabc24d

[start of binary data...]
```

Notice that `MuseScore-x86_64.AppImage` (default AppImage name) was hardcoded as both Filename and (relative) URL at the time the zsync was created. When we subsequently renamed the AppImage file to `MuseScore-[build_info]-x86_64.AppImage` the link would stop working. I have now updated the script so that the AppImage is created with its final name, so the zsync metadata should be valid.

Other improvements:

- Substituted `3x` string for a variable in $UPDATE_INFORMATION so the scripts can be safely copied to the master branch.

- Latest "AppImage" symlink was being created for [Windows](https://ftp.osuosl.org/pub/musescore-nightlies/windows/3x/nightly/?C=M;O=D) and [macOS](https://ftp.osuosl.org/pub/musescore-nightlies/macos/3x/nightly/?C=M;O=D) as well as Linux. The symlink might be useful so I kept it, but the file extension is now `.7z` or `.dmg` on those platforms.

### DIFF
--- a/build/ci/tools/osuosl/publish.sh
+++ b/build/ci/tools/osuosl/publish.sh
@@ -10,7 +10,7 @@ while [[ "$#" -gt 0 ]]; do
         -s|--secret) OSUOSL_SSH_ENCRYPT_SECRET="$2"; shift ;;
         --os) OS="$2"; shift ;;
         -m|--mode) BUILD_MODE="$2"; shift ;;
-        -v|--version) MAJOR_VERSION="$2"; shift ;;
+        -v|--version) BUILD_VERSION="$2"; shift ;;
         *) echo "Unknown parameter passed: $1"; exit 1 ;;
     esac
     shift
@@ -18,11 +18,15 @@ done
 
 if [ -z "$ARTIFACT_NAME" ]; then ARTIFACT_NAME="$(cat $ARTIFACTS_DIR/env/artifact_name.env)"; fi
 if [ -z "$BUILD_MODE" ]; then BUILD_MODE=$(cat $ARTIFACTS_DIR/env/build_mode.env); fi
+if [ -z "$BUILD_VERSION" ]; then BUILD_VERSION=$(cat $ARTIFACTS_DIR/env/build_version.env); fi
+
+MAJOR_VERSION="${BUILD_VERSION%%.*}"
 
 echo "ARTIFACT_NAME: $ARTIFACT_NAME"
 echo "SECRET: $OSUOSL_SSH_ENCRYPT_SECRET"
 echo "OS: $OS"
 echo "BUILD_MODE: $BUILD_MODE"
+echo "BUILD_VERSION: $BUILD_VERSION"
 echo "MAJOR_VERSION: $MAJOR_VERSION"
 
 OS_IS_VALID=0
@@ -43,7 +47,7 @@ if [ -z "$OSUOSL_SSH_ENCRYPT_SECRET" ]; then echo "error: not set OSUOSL_SSH_ENC
 
 SSH_KEY=build/ci/tools/osuosl/osuosl_nighlies_rsa
 
-#if [ "$OS" == "windows" ]; then 
+#if [ "$OS" == "windows" ]; then
 #Icacls $SSH_KEY
 #Icacls $SSH_KEY /Inheritance:r
 #Icacls $SSH_KEY /Grant:r "$(whoami)":'(R,W)'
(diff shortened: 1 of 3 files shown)


build/ci/tools/osuosl/publish.sh
```diff
@@ -54,16 +58,20 @@ chmod 600 $SSH_KEY
...
 echo "Copy ${ARTIFACTS_DIR}/${ARTIFACT_NAME} to $FTP_PATH"
 scp -oStrictHostKeyChecking=no -C -i $SSH_KEY $ARTIFACTS_DIR/$ARTIFACT_NAME musescore-nightlies@ftp-osl.osuosl.org:~/ftp/$FTP_PATH
 
 # For Linux, we also need to send a .zsync file, if exists
-ZSYNC_EXISTS=0
-if [ "$OS" == "linux" ]; then 
+if [ "$OS" == "linux" ]; then
     if [ -f "$ARTIFACTS_DIR/${ARTIFACT_NAME}.zsync" ]; then
         echo "Copy ${ARTIFACTS_DIR}/${ARTIFACT_NAME}.zsync to $FTP_PATH"
+        sleep 1m # For fine sort order on FTP
```
igorkorsukov, line 71: I looked, it is still not sorted correctly, all 4 files have the same time. I think I need to try to set 2 minutes.
If that doesn't work, we might need to call the trigger, then delay, create links, and call the trigger again.

build/ci/tools/osuosl/publish.sh
```diff
@@ -72,33 +80,20 @@ echo $PUBLISH_URL > $ARTIFACTS_DIR/env/publish_url.env
...
+if [ "$BUILD_MODE" == "nightly_build" ]; then
     echo "Delete old MuseScoreNightly files"
 
-    if [ "$OS" == "linux" ]; then 
-        ssh -i $SSH_KEY musescore-nightlies@ftp-osl.osuosl.org "cd ~/ftp/$FTP_PATH; ls MuseScoreNightly*.AppImage       -t | tail -n +41 | xargs rm -f"
-        ssh -i $SSH_KEY musescore-nightlies@ftp-osl.osuosl.org "cd ~/ftp/$FTP_PATH; ls MuseScoreNightly*.AppImage.zsync -t | tail -n +41 | xargs rm -f"
-    else 
-        ssh -i $SSH_KEY musescore-nightlies@ftp-osl.osuosl.org "cd ~/ftp/$FTP_PATH; ls MuseScoreNightly* -t | tail -n +41 | xargs rm -f"
-    fi
+    ssh -i $SSH_KEY musescore-nightlies@ftp-osl.osuosl.org "cd ~/ftp/$FTP_PATH; ls MuseScoreNightly* -t | tail -n +42 | xargs rm -f"
```
igorkorsukov, line 86: I think it's not important, but if we don't do a separate logic for Linux, there will be 2 times less available files in the history of days.

build/ci/tools/osuosl/publish.sh
```diff
@@ -72,33 +80,20 @@ echo $PUBLISH_URL > $ARTIFACTS_DIR/env/publish_url.env
...
 fi
 
 # Create link to latest
-if [ "$BUILD_MODE" == "nightly_build" ]; then 
+if [ "$BUILD_MODE" == "nightly_build" ]; then
 
     echo "Create link to latest"
 
     sleep 1m # For fine sort order on FTP
```
igorkorsukov, line 94: I looked, it is still not sorted correctly, all 4 files have the same time. I think I need to try to set 2 minutes.
If that doesn't work, we might need to call the trigger, then delay, create links, and call the trigger again.

build/ci/tools/osuosl/publish.sh
shoogle: @igorkorsukov, the latest zsync is a real file now, not a symlink, and I'm not keeping historic zsync files. So there will only be one less nightly on Linux, not half.

igorkorsukov: I got it!

build/ci/tools/osuosl/publish.sh
shoogle: @igorkorsukov, did you check via SSH or just looking at the [webpage](https://ftp.osuosl.org/pub/musescore-nightlies/linux/3x/nightly/?C=M;O=D)? The files have the same timestamp on the webpage (to the minute), but the sorting is done on the server rather than through JavaScript. Notice that:

- If you reverse the sort order on the webpage then the list reverses.
- If you switch to ordering by name then the files are ordered differently.

 So it appears that the server has access to the full timestamp (to the milli/micro second) rather than just to the nearest minute.

Regarding both zsync files appearing above both AppImages, I think that the system is using the timestamp of the original file rather than the symlink. Using hard links won't help here, unfortunately, but I could replace the symlink with a physical copy if the order matters more than disk space?

igorkorsukov: I was looking at a web page.
If the link uses the timestamp of the file, then I suggest keep it as it is, you can even remove `sleep 1m`.
Let's see if someone complains

PR #6972, opened 2020-11-30
[Mu4] volta jump regressions 3.5.1

jeetee:
Resolves: https://musescore.org/en/node/313751 and https://musescore.org/en/node/313772

master version of #6971

### DIFF
--- a/mtest/libmscore/repeat/repeat65.mscx
+++ b/mtest/libmscore/repeat/repeat65.mscx
@@ -1,0 +1,241 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<museScore version="3.02">
+  <Score>
+    <LayerTag id="0" tag="default"></LayerTag>
+    <currentLayer>0</currentLayer>
+    <Division>480</Division>
+    <Style>
+      <showMeasureNumberOne>1</showMeasureNumberOne>
+      <measureNumberInterval>1</measureNumberInterval>
+      <measureNumberSystem>0</measureNumberSystem>
+      <Spatium>1.76389</Spatium>
+      </Style>
+    <showInvisible>1</showInvisible>
+    <showUnprintable>1</showUnprintable>
+    <showFrames>1</showFrames>
+    <showMargins>0</showMargins>
+    <metaTag name="arranger"></metaTag>
+    <metaTag name="composer"></metaTag>
+    <metaTag name="copyright"></metaTag>
+    <metaTag name="lyricist"></metaTag>
+    <metaTag name="movementNumber"></metaTag>
+    <metaTag name="movementTitle"></metaTag>
+    <metaTag name="poet"></metaTag>
+    <metaTag name="source"></metaTag>
+    <metaTag name="translator"></metaTag>
+    <metaTag name="workNumber"></metaTag>
+    <metaTag name="workTitle">Repeat Test</metaTag>
+    <Part>
+      <Staff id="1">
+        <StaffType group="pitched">
+          <name>stdNormal</name>
+          </StaffType>
+        </Staff>
+      <trackName>Piano</trackName>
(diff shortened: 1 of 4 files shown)


libmscore/repeatlist.cpp
```diff
@@ -764,7 +769,7 @@ void RepeatList::unwind()
     std::pair<QList<QList<RepeatListElement*>*>::const_iterator, QList<RepeatListElement*>::const_iterator> continueAt = std::make_pair(
         _rlElements.cend(), _rlElements[0]->cend());
     Jump const* activeJump = nullptr;
-    bool forceFinalRepeat = false;   // Used during jump processing
+    bool forceFinalRepeat;   // Used during jump processing
```
Eism, line 772: better to save first initialization

jeetee: Why? It only slows down.

jeetee: OK, I see now it is part of the guidelines.

At least I'd rather keep initialization to an unused value here instead of moving declaration inside the loop which would be even more wasted instructions.

jeetee: Force pushed it

Eism: I'll just explain why this is so :)
it is an undefined state. After this line it is not clear whether it is true or false.
It is even worse if someone decides to use this flag before initialization. Just safer, it's hard to reproduce bugs like this :)

jeetee: Yes, and usually I'd agree that initialization is a must from a safety pov. in this situation however, the comment clarifies the intended use of the variable, which "logically" should only be declared/scoped within the loop.
The only reason for not doing so, is to prevent continuous stack allocations which would slow down processing.

Similarly, you now take a penalty for explicitly initializing something to a value that is irrelevant (and should be at that stage). If at some future point in time someone decides indeed to use that flag, it would be part of their modification and responsibility to ensure it is initialized correctly for their use case.


I know that the above statement perhaps comes across a bit harsh; but I am misformed by years of embedded safety programming, where there are quite some rules about when initialization is required (indeed, as a ground rule, always) and when it should be omitted as part of the intent of the declaration.

With regards to this PR specifically, it was a mere port of the one for Mu3, so I try to keep most things as similar as possible. I'll gladly perform a code review session once 3.x is over and out (somewhere after 3.6.1 likely) to bring the whole file up to Mu4 standards and guidelines.

jeetee: (as an aside, note the similar approach for the playCount variable just above it; we probably should make that follow the rules in a future update then as well ;-) )

Eism (approved):

jeetee:
Can we see a closer review of the failing mtest reason within the github actions?
Same mtest for mu3 were happy though..

Eism:
@jeetee 
TestRepeat::repeat65() readScore: cannot load </mtest/libmscore/repeat/repeat65.mscx> type <mscx>

with error: Score::FileError::FILE_TOO_NEW
description: This score was saved using a newer version of MuseScore. Visit the %1MuseScore website%2 to obtain the latest version.

Eism:
@jeetee please add error code output with description to the console. will be useful in the future
/mtest/testutils.cpp:140

jeetee:
I'll look into that, but mtest is currently missing as part of the project on MinGW/QtCreator by default for me; haven't looked into that yet.

So likely something changed here between build systems and or mu3 vs mu4 as well; as in mu3 this output at least was generated at one point. Or perhaps the mtests used to run under debug and now don't?

vpereverzev:
Looking for rebase and then I'll merge it

jeetee:
happy times

Eism (approved):

PR #6983, opened 2020-12-01
[MU3] Center score when switching LayoutMode

cbjeukendrup:
In Page view and Single Page view:
If the width of the entire score is smaller than the width of the Score View, then the score will be centered horizontally; otherwise, the score will be left-aligned, with some padding.

In Page view, the same goes for height / vertical alignment. 

Small width, large height:
Small width, large height

Large width and height:
Schermafbeelding 2020-12-01 om 11 14 25

### DIFF
--- a/mscore/prefsdialog.cpp
+++ b/mscore/prefsdialog.cpp
@@ -1235,9 +1235,7 @@ void PreferenceDialog::applyPageVertical()
                   ss->doLayout();
             }
       if (cv)
-            cv->setOffset(0.0, 0.0);
-      if (mscore->currentScoreView())
-            mscore->currentScoreView()->setOffset(0.0, 0.0);
+            cv->pageTop();
       mscore->scorePageLayoutChanged();
       mscore->update();
       }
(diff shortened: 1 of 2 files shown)


mscore/prefsdialog.cpp
```diff
@@ -1235,9 +1235,9 @@ void PreferenceDialog::applyPageVertical()
                   ss->doLayout();
             }
       if (cv)
-            cv->setOffset(0.0, 0.0);
+            cv->pageTop();
       if (mscore->currentScoreView())
-            mscore->currentScoreView()->setOffset(0.0, 0.0);
+            mscore->currentScoreView()->pageTop();
```
cbjeukendrup, line 1240: Do line 1239 and 1240 make sense, or can they be deleted? They seem redundant to me, since `cv` _is_ `mscore->currentScoreView()` (see some lines above).

vpereverzev: mm, it's quite strange actually, looks like a duplication
let's remove it

cbjeukendrup: I did that, and it doesn't seem to cause any problems. So let's indeed remove it!

PR #6935, opened 2020-11-24
[mu3] Fix #214996: Add a style option to show measure number range at mmrests

Marr11317:
The MMRestRange is a new class derived from MeasureNumber.
There is 3 possible bracketing around the ranges: parentheses, brackets (the default), or none.

rebased version of #6009 (there's a gif demonstrating the thing in the other PR)

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [ ] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #6987, opened 2020-12-01
Followup for PR #6935, that included a bug.

Marr11317:
`MeasureNumber`s and `MMRestRange`s were `delete`d in `measure::remove`,
which caused a crash when undoing the deletion of these elements.

Although this may look as a big refactoring for the small issue it solves, it includes a refactoring of the measure number class into a measureNumberBase class, which MMRestRange, as well as MeasureNumber, inherit. So in fact there's only code moving here.

```
Previously:   TextBase->MeasureNumber->MMRestRange
WIth this PR: TextBase->MeasureNumberBase->MeasureNumber
              							 ->MMRestRange
```

This refactoring should lead to a less errorprone implementation.

### DIFF
--- a/libmscore/measurenumberbase.cpp
+++ b/libmscore/measurenumberbase.cpp
@@ -1,0 +1,188 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2020 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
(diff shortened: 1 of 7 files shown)


libmscore/measurenumberbase.cpp
```diff
@@ -0,0 +1,182 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2011-2014 Werner Schweer
```
vpereverzev, line 5: Could you update this licence, along the way? It's very outdated

libmscore/measurenumberbase.cpp
```diff
@@ -0,0 +1,182 @@
...
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2
+//  as published by the Free Software Foundation and appearing in
+//  the file LICENCE.GPL
+//=============================================================================
+
+#include "score.h"
+#include "measurenumberbase.h"
+// #include "xml.h"
```
vpereverzev, line 15: looks like we don't need it anymore

Marr11317:
updated license and removed unused includes

PR #6989, opened 2020-12-02
Clean-up Page page of Styles and fix tabstops

Howard-C:
**There're two sets of options, only one of which can be enabled at the same time, so I put them in two group boxes, and enabling one group box will disable the other.**

Before:
image

After:
image

### DIFF
--- a/mscore/editstyle.cpp
+++ b/mscore/editstyle.cpp
@@ -557,16 +557,17 @@ EditStyle::EditStyle(Score* s, QWidget* parent)
       showFooter->setToolTip(toolTipHeaderFooter);
       showFooter->setToolTipDuration(5000);
 
-      connect(buttonBox,           SIGNAL(clicked(QAbstractButton*)), SLOT(buttonClicked(QAbstractButton*)));
-      connect(enableVerticalSpread,SIGNAL(toggled(bool)),             SLOT(toggleVerticalJustificationStaves(bool)));
-      connect(headerOddEven,       SIGNAL(toggled(bool)),             SLOT(toggleHeaderOddEven(bool)));
-      connect(footerOddEven,       SIGNAL(toggled(bool)),             SLOT(toggleFooterOddEven(bool)));
-      connect(chordDescriptionFileButton, SIGNAL(clicked()),          SLOT(selectChordDescriptionFile()));
-      connect(chordsStandard,      SIGNAL(toggled(bool)),             SLOT(setChordStyle(bool)));
-      connect(chordsJazz,          SIGNAL(toggled(bool)),             SLOT(setChordStyle(bool)));
-      connect(chordsCustom,        SIGNAL(toggled(bool)),             SLOT(setChordStyle(bool)));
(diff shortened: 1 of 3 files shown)


mscore/editstyle.ui
```diff
@@ -250,7 +250,7 @@
        </size>
       </property>
       <property name="currentIndex">
-       <number>0</number>
+       <number>1</number>
```
Jojo-Schmitz, line 253: Wrong...

mscore/editstyle.cpp
```diff
@@ -1363,25 +1364,24 @@ void EditStyle::enableStyleWidget(const Sid idx, bool enable)
...
-      enableStyleWidget(Sid::spreadSquareBracket, checked);
-      enableStyleWidget(Sid::spreadCurlyBracket, checked);
-      enableStyleWidget(Sid::minSystemSpread, checked);
-      enableStyleWidget(Sid::maxSystemSpread, checked);
-      enableStyleWidget(Sid::minStaffSpread, checked);
-      enableStyleWidget(Sid::maxStaffSpread, checked);
-      enableStyleWidget(Sid::maxAkkoladeDistance, checked);
-      enableStyleWidget(Sid::maxPageFillSpread, checked);
+      disableVerticalSpread->setChecked(!checked);
+      cs->setLayoutAll();
```
Howard-C, line 1373: Not sure why I need to add this sentence to make the score keep up with the change of the group boxes. Without this commit the score will update immediately.

mscore/editstyle.ui
Howard-C: Oops

mscore/editstyle.ui
```diff
@@ -910,39 +910,48 @@
...
+             <property name="toolTip">
+              <string>Reset to default</string>
+             </property>
              <property name="accessibleName">
-              <string>Reset &quot;Max. page fill distance&quot; value</string>
+              <string>Reset 'Last system fill threshold' value</string>
              </property>
              <property name="text">
-              <string>...</string>
+              <string notr="true"/>
```
Jojo-Schmitz, line 945: lost the ellipsis?

Howard-C: I think the git diff makes it appear as though these text fields got swapped, so there's this change. The text it applies to is different

Howard-C: Actually it's a reset button, it doesn't even need the ellipsis

Jojo-Schmitz: it may loose that ` notr="true"` then. (a minor nitpick though)

Jojo-Schmitz (approved):

Jojo-Schmitz:
Next stop is cleaning up the overly crowded and too high "Score" page ;-)

mscore/editstyle.ui
Howard-C: I didn't change it, it was done by Qt Designer

vpereverzev:
Welcome back, @Howard-C

Howard-C:
@vpereverzev Actually, no :) I freed my time to do this because Beta is coming shortly. I'll still be gone for at least a month, but may pop back to do some other small emergency fixes

PR #6990, opened 2020-12-02
[MU4] Removed old code (in the master only keeps MU4)

igorkorsukov:
Removed old code (in the master only keeps MU4)


main/main.cpp
```diff
@@ -18,7 +18,6 @@
 //=============================================================================
 
 #include "config.h"
-#include "mscore/musescore.h"
 #include "modulessetup.h"
 
 #ifdef BUILD_UI_MU4
```
RomanPudashkin, line 23: You missed this

igorkorsukov: I plan to do it little by little. One of the next steps will be to remove this option. Thanks

mu4/commonscene/exampleview.cpp
```diff
@@ -11,7 +11,7 @@
 //=============================================================================
 
 #include "exampleview.h"
-#include "mscore/preferences.h"
+//#include "mscore/preferences.h"
```
RomanPudashkin, line 14: Can we just remove this header?

main/main.cpp
igorkorsukov: "Rough strokes" are now done :)

mu4/commonscene/exampleview.cpp
igorkorsukov: removed

PR #6993, opened 2020-12-02
[MU4] Removed Aeolus

igorkorsukov:
Removed Aeolus


Jojo-Schmitz:
A really pity, I'd rather see this enabled again and updated, esp. in connection with moving to GPL3 this should be possible

igorkorsukov:
> A really pity, I'd rather see this enabled again and updated, esp. in connection with moving to GPL3 this should be possible

I'm also sorry to remove it, but I think it's reasonable. If we want and can use this synthesizer future, then it's better to add it again, like other synthesizers, and not cling to this code.

RomanPudashkin (approved):

Eism (approved):

PR #7000, opened 2020-12-02
[MU3] Fix #313857: Fractional font sizes can't be input using the "Preferences" dialog

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/313857


jthistle:
Congratulations! Another one to add to the list ;)

PR #7003, opened 2020-12-02
[MU3] Fix 310158: Don't null terminate MIDI sequence track name messages

drebelsky:
Resolves: https://musescore.org/en/node/310158

This PR makes it so that the null terminator from a part's name is not copied.


Jojo-Schmitz:
the mtests are not happy, which I guess is understandable, but needs fixing anyway?

Jojo-Schmitz:
I don't quite understand why the vtests failed, but can't imagine this to be your (PR's) fault.

drebelsky:
It looks like the layout code for 3.x has been changed since I opened the PR, but since my commits don't change these, I believe they could be merged without problems. I could also pull in the 3.x branch if that makes more sense.

Jojo-Schmitz:
As far as I can tell the build in CI is always done against the then latest commit.
But try a rebase to see if it helps here

audio/exports/exportmidi.cpp
```diff
@@ -90,7 +90,7 @@ void ExportMidi::writeHeader()
             Staff* staff  = cs->staff(staffIdx);
 
             QByteArray partName = staff->partName().toUtf8();
-            int len = partName.length() + 1;
+            int len = partName.length();
```
psmokotnin, line 93: This line was correct. An increasing of the length must be here.

See Qt's documentation: https://doc.qt.io/qt-5/qbytearray.html#data

psmokotnin (changes requested):

psmokotnin:
I would fix it here.
https://github.com/musescore/MuseScore/blob/8c98dedc06f7931f9a583c64217a92512207caa1/audio/midi/midifile.cpp#L138-L139

audio/exports/exportmidi.cpp
drebelsky: Since we're doing a `memcpy` instead of a `strcpy`, I don't think the `+ 1` is necessary.

drebelsky:
I feel like it makes sense to store the non-null-terminated string as the data for the event since it's the actual data. Fixing it in 
https://github.com/musescore/MuseScore/blob/8c98dedc06f7931f9a583c64217a92512207caa1/audio/midi/midifile.cpp#L138-L139
probably works, but then we have to do a special check for a track name message. We could also change
https://github.com/musescore/MuseScore/blob/8c98dedc06f7931f9a583c64217a92512207caa1/audio/exports/exportmidi.cpp#L102, but then the actual data length and nominal data lengths would be different.

psmokotnin:
But creation of char string without a null termination is very crashable way.
https://github.com/musescore/MuseScore/blob/6e287410b52e4173aa21109e1ca11bd563978feb/audio/midi/event.h#L205

It's good to fix it where file are saved instead of creation of broken MIDI events. For an example, midi import adds null termination while reading a file.

Jojo-Schmitz:
Please squash the commits `git rebase -i HEAD~3`.

audio/midi/midifile.cpp
```diff
@@ -135,8 +135,14 @@ void MidiFile::writeEvent(const MidiEvent& event)
             case ME_META:
                   put(ME_META);
                   put(event.metaType());
-                  putvl(event.len());
-                  write(event.edata(), event.len());
+                  if (event.metaType() == META_TRACK_NAME) {
```
psmokotnin, line 138: Why track name type only? What about the other text types (0x01 - 0x07)?

Jojo-Schmitz: Because the issue is about track name only I guess, "Sequence Track Name MIDI messages wrongly null-terminated"

psmokotnin: Obviously, we have to check out what the MIDI standard says here.

drebelsky: In https://www.midi.org/component/edocman/rp-001-v1-0-standard-midi-files-specification-96-1-4-pdf/fdocument?Itemid=9999, it looks like types 0x01–0x07 are text types, and https://www.midi.org/component/edocman/smf-device-name-and-program-name-meta-events/fdocument?Itemid=9999 adds 0x08 and 0x09. In https://github.com/musescore/MuseScore/blob/6e287410b52e4173aa21109e1ca11bd563978feb/audio/midi/event.h#L77-L81,  we define some "mscore extensions" which are also (I believe) text types. Should we null terminate these extensions? Also, tangentially related, would it make sense if we changed the (nonstandard) extensions to be sent as System Exclusive messages or sequencer specific meta messages, instead?

AndroYD84:
This problem resurfaced in the current stable and nightly build.
If I export a midi score from MuseScore and run this simple Python script:
```
import os,sys
import muspy

midi = muspy.read_midi('midi_file.mid')
print([track.name for track in midi.tracks])
```
It outputs something like `['Piano\x00', 'Strings\x00']` instead of `['Piano', 'Strings']`, happens only if the MIDI file was exported from MuseScore.

Jojo-Schmitz:
Of course it does, the changes from this PR haven't been merged into 3.x nor been ported/merged to master

Jojo-Schmitz:
The logs for the failing tests (vtests and mtest) have expired, are no longer available, so impossible to check what when wrong there, and so equally impossible to tell why they went wrong.

Please squash the commits, `git rebase -i HEAD~4` and then `git push -f` the result.

Jojo-Schmitz:
Also better rebase the entire thing on top of the master branch, as there won't be any furter 3.x release

Jojo-Schmitz:
The mtest failures are most probably due to the *-ref.mid files only having been changed after the 1st commit which then lated got reverted, and that after the 4th commit all text meta events are not null-terminated anymore.

Jojo-Schmitz:
OK, I bit the bullit and ported it over to master, was pretty easy and straight forward, esp. as none of the tests needed to get ported ;-)

Jojo-Schmitz:
I'd like to include this PR in my PR #9000, but can't seem to get the mtests to work, how would these *-ref.mid files need to get generated?

Jojo-Schmitz:
OK, I managed it, it is now part of PR #9000, As no further 3.x is planned and the changed are ion master already too, I believe this PR here could/should get closed.

RomanPudashkin:
3.x is closed for any changes

PR #7005, opened 2020-12-03
Fix #313926: Correct focus-in problem when a dialog box is open

worldwideweary:
Resolves: https://musescore.org/en/node/313926 (bug-fix)

**Explanation**:
MS steals the focus here as mentioned in the issue. This is due to a clash between the code found in 
`MuseScore::onFocusWindowChanged(QWindow* w)` 
mixed with a recent change (from #6802) with `MuseScore::focusScoreView `to allow for the QML dock widget for palettes to float and honor [`escape key functionality`] to bring score back into focus. Apparently `onFocusWindowChanged` is making a dummy widget and calling focus on it, and that focus was switching focus to the main application if a dialog box was open, even when accessing OS/WindowManager-level windows. 

Instead of attempting to change the `onFocusWindowChanged` code (what seems like a hack based upon QML dock window limitations?), I made sure that the activation only occurs when specifically triggering Escape from within the Palette QMLwidget and all seems well.

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -4023,7 +4023,6 @@ bool MuseScoreApplication::event(QEvent* event)
 void MuseScore::focusScoreView()
       {
       if (currentScoreView()) {
-            currentScoreView()->activateWindow();
             currentScoreView()->setFocus();
             }
       else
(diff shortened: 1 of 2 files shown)


njvdberg (approved):
Applied the PR and it works as advertised. Thanks!

PR #7006, opened 2020-12-03
Allow windows build with vscode

Marr11317:
Resolves: no issue in tracker

using msvc build.bat, this allows to build musescore with vscode.


Marr11317:
Note that the other option to have a viable vscode config would be to add .vscode to .gitignore, so it that users could manage that themselves these settings themselves. It would however require more manual setup.

.vscode/c_cpp_properties.json
```diff
@@ -28,26 +28,42 @@
...
                 ]
             }
+        },
+        {
+            "name": "Win32",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "C:/Qt/5.12.10/msvc2017_64/bin",
+                "C:/Qt/5.12.10/msvc2017_64/include/**",
+                "C:/Qt/5.12.10/msvc2017_64/include/QtGui"
```
Jojo-Schmitz, line 52: So we really need to hardcode these values?

Marr11317: Maybe not. These are for not for building anyway. I can just remove them from there and let people choose what they include.

Jojo-Schmitz: Something like `${QT_INSTALL_PREFIX}/`maybe?

Marr11317: only available with cmake-initialized environments. I'll remove these include paths for the time being. Individuals will add them if they wish.

PR #7007, opened 2020-12-03
fix margins

Marr11317:
Resolves: https://trello.com/c/lju5Hxmz/43-right-page-margins-are-much-too-thin-on-all-scores-new-and-old

There is probably a cleaner fix for this, but I'll leave it for the dev team.

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -8020,7 +8020,7 @@ void MuseScore::init(QStringList& argv)
                   QRectF psf = p.paperRect(QPrinter::Inch);
                   MScore::defaultStyle().set(Sid::pageWidth,  psf.width());
                   MScore::defaultStyle().set(Sid::pageHeight, psf.height());
-                  MScore::defaultStyle().set(Sid::pagePrintableWidth, psf.width()-20.0/INCH);
+                  MScore::defaultStyle().set(Sid::pagePrintableWidth, psf.width() - 30.0/INCH);
                   MScore::defaultStyle().set(Sid::enableVerticalSpread, true);
                   }
             }


mscore/musescore.cpp
```diff
@@ -8020,7 +8020,7 @@ void MuseScore::init(QStringList& argv)
                   QRectF psf = p.paperRect(QPrinter::Inch);
                   MScore::defaultStyle().set(Sid::pageWidth,  psf.width());
                   MScore::defaultStyle().set(Sid::pageHeight, psf.height());
-                  MScore::defaultStyle().set(Sid::pagePrintableWidth, psf.width()-20.0/INCH);
+                  MScore::defaultStyle().set(Sid::pagePrintableWidth, psf.width() - 30.0/INCH);
```
njvdberg, line 8023: Where does this magic number comes from? Because it was a magic number the issue happened in the first place :wink:
Better to replace the <code>30.0</code> by an expression or a constant with a descriptive name.

Marr11317: It comes from the new margins of 15/inch.
Yes, it should be a constant somewhere.

IsaacWeiss: Actually I think the constant should be `MARGIN = 15.0/INCH`, and this should be `2 * MARGIN`, and both magic numbers would be gotten rid of.

njvdberg: Total <code>MARGIN</code> isn't a fixed number, that's the reason this issue exists :wink: 
So, better <code>MARGIN  = "Sid::pageOddLeftMargin" - "Sid::pageOddRightMargin"</code>. It doesn't matter whether the margins for the odd or even pages are used because in the UI the even and odd pages are always symmetrical.

PR #6992, opened 2020-12-02
[MU4] Fix -Wdeprecated-copy warnings

minerscale:
Resolves: https://musescore.org/en/node/313872

Replaced some copy constructors with the defualt copy constructor, for the most part this appears okay, except for the change with event.cpp/event.h which I'm very unsure about whether or not it works. It compiles on my machine and doesn't appear to introduce any new bugs, so I guess we'll see what Travis CI thinks.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7012, opened 2020-12-03
[MU4] Followup to PR #6992, suppressed remaining -Wdeprecated-copy

minerscale:
Resolves: https://musescore.org/en/node/313872

Suppressed the last -Wdeprecated-copy warnings

### DIFF
--- a/src/framework/midi_old/CMakeLists.txt
+++ b/src/framework/midi_old/CMakeLists.txt
@@ -38,4 +38,6 @@ set (MODULE_DEF
 
 include(${PROJECT_SOURCE_DIR}/build/module.cmake)
 
-
+if (NOT MSVC)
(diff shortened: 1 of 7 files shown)


Eism:
how many warnings in the master?
maybe we should fix these warnings and not ignore them?

igorkorsukov:
> how many warnings in the master?
> maybe we should fix these warnings and not ignore them?

We tried to fix it, but some cannot be fixed.
https://github.com/musescore/MuseScore/pull/6992

minerscale:
> how many warnings in the master?
> maybe we should fix these warnings and not ignore them?

Yeah I already fixed like, 90% of them, the rest are warnings emitted by a qt header file, and there seems to be no way to change the copy constructor in such a way GCC doesn't complain about them, hence we just ignore them.

PR #7015, opened 2020-12-03
Fix #313896 Portamento playdown not working

ypoissant:
Resolves: [Glissando down playback not working when set to "Portamento"](https://musescore.org/en/node/313896)

Portamento style glissando going from a higher note to a lower note did not play like a portamento but played the two begin and end note.

### DIFF
--- a/libmscore/rendermidi.cpp
+++ b/libmscore/rendermidi.cpp
@@ -268,7 +268,7 @@ static void playNote(EventMap* events, const Note* note, int channel, int pitch,
                         Note* nextNote = toNote(spanner->endElement());
                         double pitchDelta = (static_cast<double>(nextNote->pitch()) - pitch) * 50.0;
                         double timeDelta = static_cast<double>(offTime - onTime);
-                        double timeStep = timeDelta / pitchDelta * 20.0;
+                        double timeStep = std::abs(timeDelta / pitchDelta * 20.0);
                         double t = 0.0;
                         QList<int> onTimes;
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
I believe the commit title should be `Fix #313896: Portamento playdown not working` for the aute marking as fixed to work

Jojo-Schmitz:
And you left out the direct link to the issue in the PR message

ypoissant:
I just clicked on the wrong button prematurely. I edited the PR message. But I'm not sure if the direct link to the fixed issue is formatted correctly.

Jojo-Schmitz:
That's OK, it leads to it

libmscore/rendermidi.cpp
```diff
@@ -268,7 +268,7 @@ static void playNote(EventMap* events, const Note* note, int channel, int pitch,
                         Note* nextNote = toNote(spanner->endElement());
                         double pitchDelta = (static_cast<double>(nextNote->pitch()) - pitch) * 50.0;
                         double timeDelta = static_cast<double>(offTime - onTime);
-                        double timeStep = timeDelta / pitchDelta * 20.0;
+                        double timeStep = std::abs(timeDelta / pitchDelta * 20.0);
```
vpereverzev, line 271: Do we prevent 0 values of pitchDelta here?

ypoissant: I'll make sure this condition is trapped in my next update to PR #7021. And for some reason, the std::abs I added here is not present in PR #7021 so I'll add it again.

PR #7017, opened 2020-12-03
[MU4] Fix #313857: Fractional font sizes can't be input using the "Preferences" dialog

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/313857

This is the #7000 version for master, albeit pretty much untested, as the UI for preferences is... rather rough at the edges to put it very mildly

### DIFF
--- a/src/framework/ui/internal/uiconfiguration.cpp
+++ b/src/framework/ui/internal/uiconfiguration.cpp
@@ -34,7 +34,7 @@ void UiConfiguration::init()
     });
 
     settings()->valueChanged(FONT_SIZE_KEY).onReceive(nullptr, [this](const Val& val) {
-        m_currentFontSizeChannel.send(val.toInt());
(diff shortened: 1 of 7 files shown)


RomanPudashkin:
I think, better to delay this fix. Recently I have made the fonts theme by the design of the design team
https://github.com/musescore/MuseScore/pull/6984

It will help to fix our "fonts hell" in qml. Also, we want to find the best way to make fonts the same on all platforms. Probably we will switch the fonts theme to using points, but we need some time to find a suitable configuration for this

Jojo-Schmitz:
Yes, I don't think this one is really ready for merge, just wanted to prevent it getting forgotten

RobFog:
Maybe this should be made a draft then?

Jojo-Schmitz:
done

Jojo-Schmitz:
note to self: rebase needed

PR #7019, opened 2020-12-03
Fix #313886: Streamline user experience regarding Export

cbjeukendrup:
Resolves: https://musescore.org/en/node/313886

- Smarter suggestions for filenames.
- If only one file will be created during export, the filename will be exactly as the user types in the SaveDialog, so therefore, we can have the SaveDialog responsible for asking whether the user wants to replace any existing files.
- Otherwise, we will ask the user ourselves.
- If the user has clicked "Skip all" or "Replace all", that choice will be synced between the ExportDialog and savePng and saveSvg (by having them use a pointer to the same SaveReplacePolicy), so that the user won't be asked again the same question.
 - Removed the "Export was successful" message (was redundant).

There is a known issue in Qt/macOS: it seems since macOS 10.15, the `DontConfirmOverwrite` option is ignored with `QFileDialog::getSaveFileName`. This annoying in cases like when a file "Somefile.pdf" exists, and the user is exporting multiple scores and types "Somefile.pdf" in the QFileDialog. macOS thinks that the file "Somefile.pdf" will be overwritten, so it gives a false alarm to the user; in fact, files like "Somefile - Full Score.pdf" and "Somefile - Violin 1.pdf" are created, and "Somefile.pdf" will be untouched. 
In the case that "Somefile - Full Score.pdf" already exists, MuseScore will still ask whether it should be overwritten. So there is completely no risk of data loss, only the user is confused with a false warning by macOS. We can't do much about that, and it only occurs in quite specific circumstances, so I think it's not a huge problem for us.

### DIFF
--- a/libmscore/musescoreCore.h
+++ b/libmscore/musescoreCore.h
@@ -17,6 +17,7 @@ namespace Ms {
 
 class MasterScore;
 class Score;
+enum class SaveReplacePolicy;
 
 //---------------------------------------------------------
 //   class MuseScoreCore
@@ -34,7 +35,7 @@ class MuseScoreCore
       Score* currentScore() const        { return cs;     }
       void setCurrentScore(Score* score) { cs = score;    }
 
-      virtual bool saveAs(Score*, bool /*saveCopy*/, const QString& /*path*/, const QString& /*ext*/) { return false; }
+      virtual bool saveAs(Score*, bool /*saveCopy*/, const QString& /*path*/, const QString& /*ext*/, SaveReplacePolicy* /*replacePolicy*/ = nullptr) { return false; }
       virtual void closeScore(Score*) {}
       virtual void cmd(QAction* /*a*/) {}
       virtual void setCurrentView(int /*tabIdx*/, int /*idx*/) {}
@@ -46,7 +47,5 @@ class MuseScoreCore
       virtual void updateInspector() {}
       };
 
-
 } // namespace Ms
 #endif
-
--- a/mscore/exportdialog.cpp
+++ b/mscore/exportdialog.cpp
@@ -348,8 +348,21 @@ void MuseScore::showExportDialog(const QString& type)
 
 void ExportDialog::accept()
       {
-      QString saveDialogTitle = tr("Export");
-
+      // Collect which scores to export
+      QList<Score*> scores;
+      bool containsMasterScore = false;
+      for (int i = 0; i < listWidget->count(); i++) {
+            ExportScoreItem* item = static_cast<ExportScoreItem*>(listWidget->item(i));
+            if (item->isChecked()) {
+                  scores.append(item->score());
+                  if (item->score()->isMaster())
+                        containsMasterScore = true;
+                  }
+            }
+      if (scores.isEmpty())
+            return; // Should never happen
+      
+      // Ask for save name and location
       QString saveDirectory;
       if (cs->masterScore()->fileInfo()->exists())
             saveDirectory = cs->masterScore()->fileInfo()->dir().path();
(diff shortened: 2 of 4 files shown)


libmscore/musescoreCore.h
```diff
@@ -34,7 +34,7 @@ class MuseScoreCore
       Score* currentScore() const        { return cs;     }
       void setCurrentScore(Score* score) { cs = score;    }
```
vpereverzev, line 36: what is the reason of bool* here?

I suggest to use simple `bool noToAll = false`

cbjeukendrup: The `savePng` and `saveSvg` functions ask the user about overwriting any existing files. If the user chose "Replace All" or "Skip All", that choice would formerly only be remembered inside that single call to the `saveSvg` or `savePng` function; if the function would be called the next time (for the next selected score) the user's choice will be forgotten, and the user will be asked again. 
To solve that, I use a pointer and pass it (via `saveAs`) to `savePng` and `saveSvg`, to make sure the `ExportDialog` and all calls to `savePng` or `saveSvg` are using the same bool, so if the user chooses "Replace All" or "Skip All", he won't be asked again. 
(Hope this explanation is clear ;-) )

cbjeukendrup: The `savePng` and `saveSvg` functions ask the user about overwriting any existing files. If the user chose "Replace All" or "Skip All", that choice would formerly only be remembered inside that single call to the `saveSvg` or `savePng` function; if the function would be called the next time (for the next selected score) the user's choice will be forgotten, and the user will be asked again. 
To solve that, I use a pointer and pass it (via `saveAs`) to `savePng` and `saveSvg`, to make sure the `ExportDialog` and all calls to `savePng` or `saveSvg` are using the same bool, so if the user chooses "Replace All" or "Skip All", he won't be asked again. 
(Hope this explanation is clear ;-) )

vpereverzev: ah, I see now, thx :)

It's quite confusing actually, perhaps would be bettter to create enum here and return it's values as a result

`virtual SavePolicy saveAs`

cbjeukendrup: Yes, that sounds like a good idea! I will try that soon (I'm quite busy these days!).
The saveAs function also returns a bool to indicate whether the export was successful. If that bool is never used, I can just change the function to return something like `SavePolicy`. But if the bool _is_ needed, shall I return a struct like this?
```
struct SaveAsResult {
    bool success;
    SavePolicy savePolicy
}
```

vpereverzev: yes, struct would make sense in this case

cbjeukendrup: I decided to create an `enum SaveReplacePolicy`, but still use pointers, because if I would pass the SaveReplacePolicy via the return value in a struct, quite many changes are necessary in other places, where only the `bool` about success is useful. I think that would cause many complications. 
Is that okay?

cbjeukendrup:
Would somebody know what is going on with the test and test failures? I don't think they are related to this PR. I did a rebase, but that hasn't helped yet.

MarcSabatella:
> Would somebody know what is going on with the test and test failures? I don't think they are related to this PR. I did a rebase, but that hasn't helped yet.

If you mean one involving MusicXML, that was added after the Spatium change a couple of days ago and needs to be updated.  I submitted a PR to do that yesterday.

cbjeukendrup:
@MarcSabatella Ah, your PR isn't merged yet (I somehow thought it was). So that will be all right then.

libmscore/musescoreCore.h
vpereverzev: Yeah, it's ok

PR #7021, opened 2020-12-03
Fix #313960 Portamento starts on wrong pitch

ypoissant:
Resolves: [Portamento style glissando starts on wrong note](https://musescore.org/en/node/313960)

Portamento style glissando start pitch was offset by twice the start note

### DIFF
--- a/libmscore/rendermidi.cpp
+++ b/libmscore/rendermidi.cpp
@@ -276,7 +276,7 @@ static void playNote(EventMap* events, const Note* note, int channel, int pitch,
                         easeInOut.timeList(static_cast<int>((timeDelta + timeStep * 0.5) / timeStep), int(timeDelta), &onTimes);
                         double nTimes = static_cast<double>(onTimes.size() - 1);
                         for (double time : onTimes) {
-                              int p = int(pitch + (t / nTimes) * pitchDelta);
+                              int p = int((t / nTimes) * pitchDelta);
                               int timeStamp = std::min(onTime + int(time), offTime - 1);
                               int midiPitch = (p * 16384) / 1200 + 8192;
                               NPlayEvent evb(ME_PITCHBEND, channel, midiPitch % 128, midiPitch / 128);
--- a/libmscore/rendermidi.cpp
+++ b/libmscore/rendermidi.cpp
@@ -268,28 +268,30 @@ static void playNote(EventMap* events, const Note* note, int channel, int pitch,
                         Note* nextNote = toNote(spanner->endElement());
                         double pitchDelta = (static_cast<double>(nextNote->pitch()) - pitch) * 50.0;
                         double timeDelta = static_cast<double>(offTime - onTime);
-                        double timeStep = std::abs(timeDelta / pitchDelta * 20.0);
-                        double t = 0.0;
-                        QList<int> onTimes;
-                        EaseInOut easeInOut(static_cast<qreal>(glissando->easeIn()) / 100.0,
-                              static_cast<qreal>(glissando->easeOut()) / 100.0);
-                        easeInOut.timeList(static_cast<int>((timeDelta + timeStep * 0.5) / timeStep), int(timeDelta), &onTimes);
-                        double nTimes = static_cast<double>(onTimes.size() - 1);
-                        for (double time : onTimes) {
-                              int p = int((t / nTimes) * pitchDelta);
-                              int timeStamp = std::min(onTime + int(time), offTime - 1);
-                              int midiPitch = (p * 16384) / 1200 + 8192;
-                              NPlayEvent evb(ME_PITCHBEND, channel, midiPitch % 128, midiPitch / 128);
+                        if (pitchDelta != 0.0 && timeDelta != 0.0) {
+                              double timeStep = std::abs(timeDelta / pitchDelta * 20.0);
+                              double t = 0.0;
+                              QList<int> onTimes;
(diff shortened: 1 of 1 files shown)


drebelsky:
I wonder if it makes sense to fix some adjacent problems in this PR. For example, https://musescore.org/en/node/313896 can be fixed by ensuring that `timeDelta` is positive. We could also consider changing `pitchDelta` to be an integral value (in which case we just multiply `timeStep` by `0.4` instead of `20.0` and divide by `24` instead of `1200` when calculating `midiPitch`). Then, changing `p` to be a double instead of an int has the benefit of perhaps allowing a little more granularity to the pitch bends.

drebelsky:
I hadn't seen https://github.com/musescore/MuseScore/pull/7015/, which resolves issue 313896.

ypoissant:
> I hadn't seen #7015, which resolves issue 313896.

I agree with your observations concerning use of float for p. But for MS3.6 it is going to stay the way it is for two reasons: 1) it is not really a bug and 2) p is used in line 281, which is copied from the bend function (line 502) and should really be a function that will be rewritten to work with different pitch wheel sensitivities, which, right now, is hard coded for a +/- 2 octaves.

Eventually, I want to rewrite the pitch bending function (lines 497 to 502) to have the ease-in, ease-out curve too and, while rewriting that, fix the fact that p don't reach the end pitch most of the time. But that will be for MS4.

libmscore/rendermidi.cpp
```diff
@@ -276,7 +276,7 @@ static void playNote(EventMap* events, const Note* note, int channel, int pitch,
                         easeInOut.timeList(static_cast<int>((timeDelta + timeStep * 0.5) / timeStep), int(timeDelta), &onTimes);
                         double nTimes = static_cast<double>(onTimes.size() - 1);
                         for (double time : onTimes) {
-                              int p = int(pitch + (t / nTimes) * pitchDelta);
+                              int p = int((t / nTimes) * pitchDelta);
```
vpereverzev, line 279: I am wondering what is the likelihood that the nTimes variable will be 0 here

ypoissant: Good point. I will make sure this is robust.

ypoissant: Actually, EaseInOut::timeList(...) will always produce at least one time so the likelyhood that nTime will be 0 is 0.

drebelsky: Isn't `nTimes` one less than than the length of the time list?

ypoissant: Oh yes. You're right. I was going to commit a change that did trap both conditions anyways.

ypoissant: And add the std::abs I added in PR #7015 that somehow disapeared here.

ypoissant: I checked and nTimes should always be at least 3 whatever the timeDelta and assuming pitchDelta cannot be less than one half step.

ypoissant:
I think the vtests and mtests fail were because I did not rebase before pushing my changes. Does that makes sense? Because my changes did not change any visuals and the mtest that failed doesn't seems related to my changes.
I just did a rebase and pushed again but the tests don't restart. What am I doing wrong?

vpereverzev:
> I think the vtests and mtests fail were because I did not rebase before pushing my changes. Does that makes sense? Because my changes did not change any visuals and the mtest that failed doesn't seems related to my changes.
> I just did a rebase and pushed again but the tests don't restart. What am I doing wrong?

Yes, just rebase it on the current 3.x

Jojo-Schmitz:
Rebase needed (again)
vtest fail in a rather strange (and most probably unrelated) way

ypoissant:
I tried a rebase using the following command and got this:
```
D:\Programming_Projects\MuseScore>git pull --rebase upstream 3.x
From https://github.com/musescore/MuseScore
 * branch                  3.x        -> FETCH_HEAD
Auto packing the repository in background for optimum performance.
See "git help gc" for manual housekeeping.
Enumerating objects: 195111, done.
Counting objects: 100% (195111/195111), done.
Delta compression using up to 12 threads
Compressing objects: 100% (38621/38621), done.
Writing objects: 100% (195111/195111), done.
Total 195111 (delta 156236), reused 194095 (delta 155468), pack-reused 0
Unlink of file '.git/objects/pack/pack-03fa22e76aaef2f7e9f4ba2402e54fa81f69efac.pack' failed. Should I try again? (y/n) y
Unlink of file '.git/objects/pack/pack-03fa22e76aaef2f7e9f4ba2402e54fa81f69efac.pack' failed. Should I try again? (y/n) y
Unlink of file '.git/objects/pack/pack-03fa22e76aaef2f7e9f4ba2402e54fa81f69efac.pack' failed. Should I try again? (y/n) n
Unlink of file '.git/objects/pack/pack-03fa22e76aaef2f7e9f4ba2402e54fa81f69efac.idx' failed. Should I try again? (y/n) n
Unlink of file '.git/objects/pack/pack-0f64fddd8558fad5e70d7a6d63349d630518dcdb.pack' failed. Should I try again? (y/n) n
```
I have a ton of those. What does that means and what can I do?

ypoissant:
I did a rebase of my 'FixPitchBend' branch. Still vtest fails. Then I did rebase of my 3.x branch and pushed. That doesn't seem to have any effect of the vtests.

PR #7024, opened 2020-12-04
Clean and fix ZitaReverb

psmokotnin:
Fixes for ZitaReverb:
- params check

- default initialized values

### DIFF
--- a/effects/zita1/zita.h
+++ b/effects/zita1/zita.h
@@ -20,7 +20,7 @@
 #define __ZITA_H__
 
 #include "effects/effect.h"
-
+#include <atomic>
 namespace Ms {
 
 class EffectGui;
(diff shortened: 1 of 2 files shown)


RobFog:
I reckon this should address https://trello.com/c/oAUqTd41/48-strange-effects-stereo-issue-causes-poor-playback-quality

RomanPudashkin:
Could you please explain the reason why we can't just revert the previous fix
https://github.com/musescore/MuseScore/pull/6832

I'm afraid that this fix will be a reason for a new bunch of hard reproduced bugs

psmokotnin:
@RomanPudashkin then you'll still have a possibility of receiving the wrong numbers in reverb's parameters.

RomanPudashkin:
> @RomanPudashkin then you'll still have a possibility of receiving the wrong numbers in reverb's parameters.

Can you reproduce this situation or this is an assumption?

psmokotnin:
@RomanPudashkin when bad value (out of range) stored in the score file:
image

effects/zita1/zita.h
```diff
@@ -201,28 +218,27 @@ class ZitaReverb : public Effect
       {
       Q_OBJECT
 
-      float   _fsamp = 0.f;
+      float   _fsamp = 1.f;
```
psmokotnin, line 221: @RomanPudashkin another interesting moment. If we keep `0.f;` and object won't be initialized, than will catch division by zero;

PR #7016, opened 2020-12-03
[MU3] Leland style values update

vpereverzev:
- Updated default style values from newest Leland
- Resolved https://trello.com/c/Jef5H0PR
- Resolved https://trello.com/c/lju5Hxmz

PR #7026, opened 2020-12-04
fix #288098, #307301, #305958, #295629: interactions between breaks a…

MarcSabatella:
…nd frames

Resolves: https://musescore.org/en/node/288098
Resolves: https://musescore.org/en/node/307301
Resolves: https://musescore.org/en/node/305958
Resolves: https://musescore.org/en/node/295629

A number of bugs all turn out to be related to a single root cause,
where we don't correctly handle checking for section breaks
in the presence of frames.
In the case of the long instrument name showing inappropriately,
it's because in the case where the previous systems is just a frame,
we don't see any measure,
so we fall back to relying on the layout of the previous system
having already set up the context for us.
This isn't necessarily valid when doing partial layout.
So we need to check more thoroughly here,
accepting a frame if it has a section break itself,
but otherwise searching backwards looking for frame with a break
or a measure.
That is, we need to skip break-less frames when looking to see
if we are starting a section.
I have therefore implemented such a function.

This turns out to also fix an issue with horizontal frames,
where a break on a horizontal frame ending a system
was being ignored for the purpose of setting long names.

The same applies to the new first system indent facility,
so while these bugs are not recent regressions,
they take on new importance in 3.6.

The new function is also used to fix a related by with measure numbers,
where a frame could get in the way of resetting them after a section break.
It is used again to solve the corresponding issue on MusicXML export.

Along the way I discovered a bug leading to an assertion failure
when deleting breaks attached to frames,
so I needed to fix that by making sure LayoutBreak::measure()
doesn't assume its parent is in fact an actual Measure
as opposed to a MeasureBase.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -3708,7 +3708,9 @@ System* Score::collectSystem(LayoutContext& lc)
       {
       if (!lc.curMeasure)
             return 0;
-      Measure* measure  = _systems.empty() ? 0 : _systems.back()->lastMeasure();
+      const MeasureBase* measure  = _systems.empty() ? 0 : _systems.back()->measures().back();
+      if (measure)
+            measure = measure->findPotentialSectionBreak();
       if (measure) {
             lc.firstSystem        = measure->sectionBreak() && _layoutMode != LayoutMode::FLOAT;
             lc.firstSystemIndent  = lc.firstSystem && measure->sectionBreakElement()->firstSystemIdentation() && styleB(Sid::enableIndentationOnFirstSystem);
@@ -3807,6 +3809,10 @@ System* Score::collectSystem(LayoutContext& lc)
                   // measure in the system and we finally can create the end barline for it
 
                   Measure* m = toMeasure(lc.prevMeasure);
+                  // TODO: if lc.curMeasure is a frame, removing the trailer may be premature
+                  // but merely skipping this code isn't good enough,
+                  // we need to find the right time to re-enable the trailer,
+                  // since it seems to be disabled somewhere else
                   if (m->trailer()) {
                         qreal ow = m->width();
                         m->removeSystemTrailer();
@@ -3827,6 +3833,12 @@ System* Score::collectSystem(LayoutContext& lc)
                                     }
                               }
                         }
+                  // TODO: we actually still don't know for sure
+                  // if this will be the last true measure of the system or not
+                  // since the lc.curMeasure may be a frame
+                  // but at this point we have no choice but to assume it isn't
+                  // since we don't know yet if another true measure will fit
+                  // worst that happens is we don't get the automatic double bar before a courtesy key signature
                   minWidth += m->createEndBarLines(false);    // create final barLine
                   }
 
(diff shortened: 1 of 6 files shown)


MarcSabatella:
Lots of interrelated bugs fixed here, mostly by applying one new function that searches backward from a MeasureBase trying to find the one that will determine if this is the end of a section or lot.  Too many places in code where we looked at either prev() or prevMeasure() to decide this, but neither is reliable.  So one new member function allows solving quite a few bugs.

There is another bug I looked at but didn't immediately see a solution for - see https://musescore.org/en/node/251706. This involves courtesy key signatures *not* appearing before a horizontal frame when they should - if there is no section break.  Originally when submitted, the behavior was different, but also wrong.  The current issue is not actually connected to section breaks.  I left some TODO's in the code regarding my investigation there, but that will have to wait.

MarcSabatella:
I *think* the mtest failures here are actually correct.  I actually changed two things in the MusicXMl export - getting the measure number reset right, but also, getting the breaks right (?) in the "print" element in the presence.  But one of the failures was a spatium difference.  I didn't change anything there, though, did I?  I do recall seeing a recent change to the default spatium value, maybe actually that broke tests for everyone?

MarcSabatella:
Never mind about the break, that is my fault, good job, mtest!  I need to only use my new section break finder for actual section breaks, not line breaks (or page breaks).

But I'm still mystified by the Spatium change.

MarcSabatella:
Re-pushed the commit and that should fix the part of the mtest failure I think is my fault.  I expect to still see the Spatium failure, because at this point I'm thinking it's just broken.

njvdberg:
> I expect to still see the Spatium failure, because at this point I'm thinking it's just broken.

Due to PR #7016 which changed the <code>spatium</code> and this PR is based on a commit before the merge of that PR?

MarcSabatella:
> > I expect to still see the Spatium failure, because at this point I'm thinking it's just broken.
> 
> Due to PR #7016 which changed the `spatium` and this PR is based on a commit before the merge of that PR?

I guess that could be, but a) I did t touch that file, b) there were hundreds of files that were changed by that PR, and yet this is the only failure, and c) if I am based to old version of that file, I am also based to an old version of everything else so it should be consistent.  So something still seems wrong to me.

Nonetheless, I will try updating and rebasing again.

Jojo-Schmitz:
Why the vtest failure?

libmscore/layout.cpp
```diff
@@ -3882,6 +3894,7 @@ System* Score::collectSystem(LayoutContext& lc)
                               if (!s->enabled())
                                     s->setEnabled(true);
                               }
+                        // TODO: use findPotentialSectionBreak here to handle breaks on frames correctly?
```
MarcSabatella, line 3897: Note to self: yes, I was able to break this setting a section break on a vertical frame then editing something on the following section, the measure number reappeared at the start of the system.

libmscore/layout.cpp
```diff
@@ -4907,7 +4925,12 @@ void Score::doLayoutRange(const Fraction& st, const Fraction& et)
                   lc.tick      = Fraction(0,1);
                   }
             else {
-                  LayoutBreak* sectionBreak = lc.nextMeasure->prevMeasure()->sectionBreakElement();
+                  const MeasureBase* mb = lc.nextMeasure->prev();
+                  if (mb)
+                        mb->findPotentialSectionBreak();
```
MarcSabatella, line 4930: Note to self: works better if you actually assign the return value!

PR #7027, opened 2020-12-04
Fix #302427: Export breaths after rests in MusicXML

drebelsky:
Resolves: https://musescore.org/en/node/302427

### DIFF
--- a/importexport/musicxml/exportxml.cpp
+++ b/importexport/musicxml/exportxml.cpp
@@ -2390,24 +2390,6 @@ void ExportMusicXml::wavyLineStartStop(const ChordRest* const cr, Notations& not
       }
 
 //---------------------------------------------------------
-//   hasBreathMark - determine if chord has breath-mark
-//---------------------------------------------------------
-
-static Breath* hasBreathMark(Chord* ch)
-      {
-      Fraction tick = ch->tick() + ch->actualTicks();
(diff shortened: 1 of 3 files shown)


importexport/musicxml/exportxml.cpp
```diff
@@ -2400,6 +2400,13 @@ static Breath* hasBreathMark(Chord* ch)
       return s ? toBreath(s->element(ch->track())) : 0;
       }
 
+static Breath* hasBreathMark(Rest* rest)
```
vpereverzev, line 2403: I guess it makes sense to move this method into Rest class

vpereverzev:
A rebase on the latest 3.x should fix these mtest issues

Jojo-Schmitz:
> A rebase on the latest 3.x should fix these mtest issues

So you can as well just merge it

importexport/musicxml/exportxml.cpp
drebelsky: Since we also use this for `Chord`s (`static Breath* hasBreathMark(Chord* ch)`, line 2396), would it make sense to move this method into the `ChordRest` class?

vpereverzev: Makes sense for me

Jojo-Schmitz:
Why the vtest failuree with lots of differences?

PR #7029, opened 2020-12-05
[MU3] Update track mapping in excerpts when order of staves is changed.

njvdberg:
Resolves: https://musescore.org/en/node/313664

When (re-)sorting staves after changing the order of instruments, the track mapping in the excerpts wasn't updated. Since linking elements are based on this track mapping, the linking was not possible (or wrong),
With this PR, 


Jojo-Schmitz:
A single minor fault in mtests: https://github.com/musescore/MuseScore/pull/7029/checks?check_run_id=1503118388#step:5:770

PR #7030, opened 2020-12-05
[MU4] Fix-Woverloaded-virtual warnings and linux specific -Wclass-memaccess warning

minerscale:
Note: I made sure no data in _alsaData was left undefined before removing the memset.


minerscale (commented):
```if (NOT MSVC AND NOT MINGW AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9.0)```

Do you know if we can get rid of `AND NOT MINGW` now that I've added a check for the compiler version? MINGW uses GCC, no? So if the GCC version is > 9, we suppress the warning.

Nevermind, it doesn't seem to even work >:(

minerscale:
Oops, I was in python mode and forgot the endif :facepalm:

minerscale:
Okay can I stop messing up already? :P

igorkorsukov:
> Okay can I stop messing up already? :P

Glory СI !! ;)

PR #7001, opened 2020-12-02
fix #313420 [MusicXML] Instrument brackets are all imported at column 0

AntonioBL:
Resolves: https://musescore.org/en/node/313420

System bracket implementation was changed between MuseScore 2.x and 3.x, but it was never updated in the MusicXML import.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7032, opened 2020-12-05
update/fix mtest after PR 7001

MarcSabatella:
An mtest had been added to PR #7001 before the recent spatium change, but the PR was not merged until afterwards.  Thus, the ref file was not updated.  This updates the spatium value in the mtest ref.


MarcSabatella:
I think you did this backwards.  The Inspector list is nice and logical, it's the one in Format / Style that is practically random.  It should be changed to match the Inspector list.

AntonioBL:
@igorkorsukov or @vpereverzev please merge this PR which fixes the mtests.
As Marc said, the mtest of my PR was based on the old spacing value, so it must be updated for the mtest to pass now.

PR #7035, opened 2020-12-06
[MU3] Fix #314029: Order of Text styles in Inspector dropdown and Styles dialog is different

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/314029


MarcSabatella:
I think this is backwards.  We went out of our way to make the Inspector list nice and logical, grouping related types that are likely to be used together.  It's the list in Format / Style that is practically random and should be fixed to match the Inspector list.  For example, there is no logical reason Lyricist and Translator should be so far apart, or why the Fingerings styles should be so far from the rest of the note-specific types, or why Instrument Changes or Sticking go at the end.  That's probably just because those types were added recently and stuck at the end.

Not that the Inspector list is perfect, and certainly there could be tweaks there.  For instance, Lyricist and Lyrics both start with the same letters but aside from that, no one ever swaps these two styles.  On the other hand, changing Staff Text into Lyrics style is very common.  So moving them to be closer together in the Inspector could certainly make sense.

Jojo-Schmitz:
Somehow I manged to screw this branch up... confinined in #7036

PR #7041, opened 2020-12-07
fix #229546 [MusicXML] Clef change export/import in MusicXML switches…

AntonioBL:
… side of barline

Resolves: https://musescore.org/en/node/229546
This adds the "after-clef" attribute when exporting a MusicXML file, and reads it when importing the file.
In case of a clef with also a courtesy clef, it makes sure to export the clef only once.

### DIFF
--- a/importexport/musicxml/exportxml.cpp
+++ b/importexport/musicxml/exportxml.cpp
@@ -5696,7 +5696,7 @@ void ExportMusicXml::findAndExportClef(const Measure* const m, const int staves,
                         clefDebug("exportxml: clef at start measure ti=%d ct=%d gen=%d", tick, int(cle->clefType()), cle->generated());
                         // output only clef changes, not generated clefs at line beginning
                         // exception: at tick=0, export clef anyway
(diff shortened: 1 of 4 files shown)


AntonioBL:
I think it should work as expected. I could not find corner cases where it did not work.
But I would prefer someone else to check that this PR behaves properly.

I couldn't find how a clef change at the very first note can be written in MusciXML (without it becoming the actual main clef).

Jojo-Schmitz:
Maybe @lvinken could check on it?

rettinghaus (approved):

lvinken:
Was unable to do so yesterday and obviously missed my opportunity. Real life interfering. On first glance, the change seems OK to me.

Jojo-Schmitz:
So instead of a pre-merge blessing it is now an after-merge "Ego te absolvo" ;-)
Thanks for checking.

AntonioBL:
> Was unable to do so yesterday and obviously missed my opportunity. Real life interfering. On first glance, the change seems OK to me.

Thank you very much for checking. :+1:

PR #6956, opened 2020-11-28
Acc padding

Marr11317:
Resolves: https://trello.com/c/4jEoQVsR/22-add-padding-inside-parentheses-for-bracketed-accidentals

Add padding around bracketed accidentals

New: 
image

Old:
image

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)
- [ ] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7043, opened 2020-12-08
fix for accidental / ledger line collisions

MarcSabatella:
Check for ledger lines on left notes within the loop and factor that in as well

Not meant to be merged as is, but to incorporate into @Marr11317 's work

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -767,23 +767,42 @@ static QPair<qreal, qreal> layoutAccidental(AcEl* me, AcEl* above, AcEl* below, 
       pnd *= mag;
       pd *= mag;
 
+      Chord* chord = me->note->chord();
+      Staff* staff = chord->staff();
+      Fraction tick = chord->tick();
+
       // extra space for ledger lines
-      if (me->line <= -2 || me->line >= me->note->staff()->lines(me->note->chord()->tick()) * 2)
-            lx = qMin(lx, -acc->staff()->spatium(acc->tick()) * acc->score()->styleS(Sid::ledgerLineLength).val());
+      qreal ledgerAdjust = 0.0;
+      bool ledgerAbove = chord->upNote()->line() <= -2;
+      bool ledgerBelow = chord->downNote()->line() >= staff->lines(tick) * 2;
+      if (ledgerAbove || ledgerBelow) {
+            // ledger lines are present
+            // check for collision with lines above & below staff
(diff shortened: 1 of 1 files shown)


libmscore/layout.cpp
```diff
@@ -764,26 +764,30 @@ static qreal layoutAccidental(AcEl* me, AcEl* above, AcEl* below, qreal colOffse
       qreal lx = colOffset;
       Accidental* acc = me->note->accidental();
       qreal mag = acc->mag();
+      const qreal ledgerAdjust = -(acc->staff()->spatium(acc->tick()) * acc->score()->styleS(Sid::ledgerLineLength).val());
```
Marr11317, line 767: this will now be calculated by every single accidental. Could that be calculated only if needed?

MarcSabatella: Sure, with some effort.  But maybe not worth it.  Normally we don't need to layout out every single accidental in the whole score on every layout (unlike the case when this code was written).  We do far more expensive calculations regularly.

MarcSabatella:
The vtest differences look good to me.  Note accidental-9 looks strange on purpose - it sets the accidental-to-note distance to 2 sp.

libmscore/layout.cpp
MarcSabatella: One quick check would be, only do the calculation is the chord actually has ledger lines - that much should be simple to determine up front.

MarcSabatella:
Since PR #6956 was merged before @Marr11317 had a chance to incorporate this code, now I *would* like this PR to be considered.  I have also improved it from the original version.

Here's the commit message that otherwise would have been the description:

The current algorithm for avoiding collision with ledger lines
only checks for a ledger line on the same note as the accidental,
and it does not account for "left notes" (notes on left of downsteam).

This commit improves the algorithm to check not just the line
of the accidental, but its top and bottom, and does the adjustment
if the accidental extends anywhere into ledger lines.
It also accounts for left notes.

Marr11317 (approved):
looks good! Now no `style()->get(Sid::...)` call is done if it's not needed.

MarcSabatella:
I have rebased and pushed.  I'm expecting the same vtest differences (improvements) as before, maybe a few additional ones because I've since also enhanced the code to resolve more collisions.

RobFog:
This addresses https://trello.com/c/Uj9B5F3m/50-ledger-lines-are-colliding-with-accidentals, correct?

PR #7044, opened 2020-12-08
[MU3] Fix #305555: add cut2 and cut3 time signatures to master palette

Jojo-Schmitz:
resolves https://musescore.org/en/node/305555

Should go into master too, will create separate PR if and when needed.


Jojo-Schmitz:
oops, forgot to update fonttools/smufl2sym-in-trans.json ... will post a follow-up PR

Jojo-Schmitz:
See #20196 and #20197, that part needs to get reverted

PR #7048, opened 2020-12-08
fix vtests 

Marr11317:
after https://github.com/musescore/MuseScore/commit/1f0c7ee0f134cd538c93d05db6d7052da89eb484

export for page number was changed from `"-%pageNumber"` to `" - %pageNumber"`

### DIFF
--- a/vtest/gen
+++ b/vtest/gen
@@ -156,8 +156,8 @@ echo "   <link rel=\"stylesheet\" type=\"text/css\" href=\"style.css\">" >> $F
 echo "  </head>" >> $F
 echo "  <body>" >> $F
 echo "    <div id=\"topbar\">" >> $F
-echo "      <span>Current</span>" >> $F
-echo "      <span>Reference</span>" >> $F
+echo "      <span>After PR</span>" >> $F
+echo "      <span>Before PR</span>" >> $F
 echo "      <span>Comparison</span>" >> $F
 echo "    </div>" >> $F
 echo "    <div id=\"topmargin\"></div>" >> $F
--- a/vtest/gen.bat
+++ b/vtest/gen.bat
@@ -104,8 +104,8 @@ echo     ^<link rel="stylesheet" type="text/css" href="style.css"^> >> %F%
 echo   ^<head^> >> %F%
 echo   ^<body^> >> %F%
 echo     ^<div id="topbar"^> >> %F%
-echo       ^<span^>Current^</span^> >> %F%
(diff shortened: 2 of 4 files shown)


cbjeukendrup:
Oh oooops... This was my fault, with PR #7019. I changed it, with the idea to make it more convenient for the user, but didn't notice that it would cause problems here. Thanks for fixing it, @Marr11317 !

vtest/gen.bat
```diff
@@ -104,8 +104,8 @@ echo     ^<link rel="stylesheet" type="text/css" href="style.css"^> >> %F%
 echo   ^<head^> >> %F%
 echo   ^<body^> >> %F%
 echo     ^<div id="topbar"^> >> %F%
-echo       ^<span^>Current^</span^> >> %F%
-echo       ^<span^>Reference^</span^> >> %F%
+echo       ^<span^>After PR^</span^> >> %F%
+echo       ^<span^>Before PR^</span^> >> %F%
```
cbjeukendrup, line 108: Shouldn't these be updated also in the files "gen" and "gen-compare"?

Marr11317: oh right, forgot that.

Marr11317: done

Jojo-Schmitz (approved):

Marr11317:
> 
> 
> Oh oooops... This was my fault, with PR #7019. I changed it, with the idea to make it more convenient for the user, but didn't notice that it would cause problems here. Thanks for fixing it, @Marr11317 !

I could have changed it directly in the vtests, but thought that this change might annoy users that are regularly using this export function, since they would then have two different exported files for the same thing. e.g. `myscore-1.png` and `myscore - 1.png`

Jojo-Schmitz:
Spaces in filenames are bad, if created by default

vtest/gen
```diff
@@ -156,8 +156,8 @@ echo "   <link rel=\"stylesheet\" type=\"text/css\" href=\"style.css\">" >> $F
 echo "  </head>" >> $F
 echo "  <body>" >> $F
 echo "    <div id=\"topbar\">" >> $F
-echo "      <span>Current</span>" >> $F
-echo "      <span>Reference</span>" >> $F
+echo "      <span>After PR</span>" >> $F
+echo "      <span>Before PR</span>" >> $F
```
Jojo-Schmitz, line 160: Maybe just "Before" and "After" are better? or ""change" rather than "PR"? Not sure whether vtests are (or should be) used only on PRs.

PR #7050, opened 2020-12-08
Fix stem length of notes with mirrored notehead

cbjeukendrup:
Resolves: https://trello.com/c/EpRJtt7w/51-square-caps-adds-gap-between-stem-and-notehead-when-notehead-is-reversed

Fixed by cleverly reading the anchor points for noteheads. (In contrast to what was suggested in the Trello issue, this was not related to replacing the round caps by flat ones, but became only more visible.)

Before: 
Mirrored Noteheads before
After:
Mirrored Noteheads after
The issue with the ledger line being too long for that last note already existed in 3.5.2. I might solve that in another PR. 

A (still pretty theoretical) use case for 'mirrored noteheads' could be, when one doesn't like how the notes are oriented in a chord. Or, when one wants to show on an educational worksheet: "This is wrong". 
Mirrored Noteheads Use Case

---

### DIFF
--- a/libmscore/stem.cpp
+++ b/libmscore/stem.cpp
@@ -107,7 +107,10 @@ void Stem::layout()
             else {                              // non-TAB
                   // move stem start to note attach point
                   Note* n  = up() ? chord()->downNote() : chord()->upNote();
-                  y1      += (up() ? n->stemUpSE().y() : n->stemDownNW().y());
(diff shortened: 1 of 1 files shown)


Marr11317:
does this also fix the x notehead issue?

cbjeukendrup:
@Marr11317 No, unfortunately not. I don't really see yet what is the cause of that problem...

vpereverzev (approved):

PR #7051, opened 2020-12-08
Make maximum zoom limit bigger

Marr11317:
Resolves: no issue in tracker

When taking precise screenshots, I am often discouraged by this max zoom level. Now the zoom limit is way higher. Not too high (>1000) because when it's too high the scoreview can become unresponsive.

a beautiful screenshot which can now be taken of a stem, a notehead and a ledger line:
image


Asmatzaile:
I do like to be constrained by the zoom limit. What if an option was added to preferences>canvas>zoom, to be able to change the maximum zoom limit, with the default set to the previous limit?

Jojo-Schmitz (approved):

PR #7054, opened 2020-12-09
Temporarily disable Save to Cloud option for Beta release

cbjeukendrup:
Resolves: https://trello.com/c/TiatpJIG/57-disable-save-to-cloud-option-for-beta-release

> We need to prevent users trying to upload scores, since the version on MuseScore.com will still be using 3.5.2.
> It doesn't make sense to update the MuseScore.com version until we are satisfied that 3.6 is stable.

(I hope this is the desired way to do this. Obviously it should be reverted again, at some point.)

### DIFF
--- a/mscore/cloud/uploadscoredialog.cpp
+++ b/mscore/cloud/uploadscoredialog.cpp
@@ -28,6 +28,10 @@ namespace Ms {
 
 void MuseScore::showUploadScoreDialog()
       {
+      if (MuseScore::unstable()) {
+            QMessageBox::warning(this, tr("Save online"), tr("Saving scores online is disabled in this unstable prerelease version of MuseScore."));
+            return;
+      }
       if (!currentScore())
             return;
       if (!currentScore()->sanityCheck(QString())) {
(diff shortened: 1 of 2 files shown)


Jojo-Schmitz:
Maybe allowing for this in experimental mode? So we can test whether Save Online works at all?

cbjeukendrup:
@Jojo-Schmitz You should actually ask @vpereverzev, but I thought the point was that things could go wrong on musescore.com if 'too new' files would be uploaded.

vpereverzev:
We just can't upload anything since backend of musescore.com uses older version than ours. That's why we shouldn't confuse people.

vpereverzev (approved):

Jojo-Schmitz:
If it were hidden behind the -x (experimental) option, there would be no confusion, but those knowing this could test, lika (also) the translations

vpereverzev:
no matter of unstable or experimental mode, you'll get problems with "upload score" feature, because of the backend of musescore.com

Jojo-Schmitz:
You get that only after the attempts, after having seen the dialog(s) and its translation. No harm on the server, just a test drill

PR #7058, opened 2020-12-10
Don't discard events with positive velocity (Note On) during an export

psmokotnin:
Resolves: 
https://musescore.com/groups/improving-musescore-com/discuss/5084153
https://musescore.com/groups/improving-musescore-com/discuss/5083064

Fix a discarding of repeated events when they should be refired.


MarcSabatella:
So in particular, this also resolves https://musescore.org/en/node/303561?

I haven't followed all the discussions of this over the years, but I know there has been a lot of back and forth about the best way to handle this, and a year or so ago a solution was decided upon and implemented for the native playback within MuseScore.  Is the solution implemented here the same basic approach, so the exported audio will actually sound the same as the internal playback?

psmokotnin:
@MarcSabatella yes

PR #7063, opened 2020-12-10
Symbols fixes

vpereverzev:
Resolves: 
- https://trello.com/c/YPvZyZY5
- https://trello.com/c/fUKGB3te
- https://trello.com/c/knmrYs5s
- https://trello.com/c/q4EWcZhz


Jojo-Schmitz:
some mtests fail

libmscore/sym.h
```diff
@@ -2913,7 +2913,7 @@ class Sym {
    public:
       Sym() { }
 
-      bool isValid() const                       { return _code != -1; }
+      bool isValid() const                       { return _code != -1 && _bbox.isValid(); }
```
MarcSabatella, line 2916: This check seems to have broken the ability to add the space character from the Special Characters dialog.  Not sure if this check was necessary, or if there is perhaps a way to give space a valid bbox, or if we could just exempt space from the check

PR #7067, opened 2020-12-11
further improvements to accidental / ledger line avoidance

MarcSabatella:
Resolves: https://trello.com/c/Uj9B5F3m/50-ledger-lines-are-colliding-with-accidentals

A recent improvement to this algorithm allowed accidentals
to avoid ledger lines in a number of cases where it didn't before.
However, it was a little overly aggressive and avoided ledger lines
in a few case where it wasn't necessary.
In particular, in cases where the accidental is actually
further from the staff than thre last "left note"
(note on the left side of a downsteam)
and could potentially have been tucked in closer.

This change detects that situation -
where the bottom of an accidental above the staff
is above the top of highest extended ledger line
(above the staff; conversely for below the staff).
In these cases, we do *not* add extra horizontal space.

This fixes the slight regression a few of the vtests
after the previous improvement -
allowing the accidentals to move closer to the stem
as they wetre before that recent change.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -773,6 +773,7 @@ static QPair<qreal, qreal> layoutAccidental(AcEl* me, AcEl* above, AcEl* below, 
 
       // extra space for ledger lines
       qreal ledgerAdjust = 0.0;
+      qreal ledgerVerticalClear = 0.0;
       bool ledgerAbove = chord->upNote()->line() <= -2;
       bool ledgerBelow = chord->downNote()->line() >= staff->lines(tick) * 2;
       if (ledgerAbove || ledgerBelow) {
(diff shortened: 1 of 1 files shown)


libmscore/layout.cpp
```diff
@@ -792,12 +794,22 @@ static QPair<qreal, qreal> layoutAccidental(AcEl* me, AcEl* above, AcEl* below,
       for (int i = 0; i < lns; ++i) {
             Note* ln = leftNotes[i];
             int lnLine = ln->line();
-            qreal lnLedgerAdjust = 0.0;
-            if (lnLine <= -2 || lnLine >= staff->lines(tick) * 2)
-                  lnLedgerAdjust = ledgerAdjust;
```
MarcSabatella, line 797: This is where I was adding space to avoid the ledger lines on "left notes", but it did this for all left notes on ledger lines.  In this PR, I refine this by avoiding setting lnLegderAdjust in cases where it isn't needed.

libmscore/layout.cpp
```diff
@@ -792,12 +794,22 @@ static QPair<qreal, qreal> layoutAccidental(AcEl* me, AcEl* above, AcEl* below,
...
             qreal lnBottom = lnTop + sp;
             if (me->top - lnBottom <= pnd && lnTop - me->bottom <= pnd) {
+                  qreal lnLedgerAdjust = 0.0;
+                  if (lnLine <= -2 || lnLine >= staff->lines(tick) * 2) {
+                        // left note has a ledger line we probably need to clear horizontally as well
+                        // except for accidentals that clear the last extended ledger line vertically
+                        // in these cases, the accidental may tuck closer
+                        Note* lastLnNote = lnLine < 0  ? leftNotes[0] : leftNotes[lns - 1];
+                        int lastLnLine = lastLnNote->line();
+                        qreal ledgerY = (lastLnLine / 2) * sp;
```
MarcSabatella, line 807: Here is where I still lnLedgerAdjust, but only if the accidental does not in fact vertically clear the last extended ledger line.  That allows this case to tuck the accidental over the ledger line, as shown in the topmost accidental of the the second and third chords here:

image

PR #7071, opened 2020-12-11
[MU4] Fix compiler warnings

Jojo-Schmitz:


src/libmscore/score.cpp
```diff
@@ -3166,9 +3166,9 @@ void Score::padToggle(Pad p, const EditData& ed)
         } else {
             changeCRlen(cr, _is.duration());
 
-            for (const SymId& articulation: _is.articulationIds()) {
+            for (const SymId& articId: _is.articulationIds()) {
```
RomanPudashkin, line 3169: Please rename this to articulationId

Jojo-Schmitz: done

PR #7074, opened 2020-12-12
[MU4] Warnings fix

Eism:
- Fixed warnings if select an element in notation
- Fixed some warnings in c++


Jojo-Schmitz (approved):
I was working on that myself and came to basically the same results ;-)

RomanPudashkin (approved):

PR #7076, opened 2020-12-12
[MU3] Don't show empty Inspector

Jojo-Schmitz:
for tremolos between notes if on tablature or not on half notes.
Also disable Inspector for system- and page breaks.

Resolves: issue discussed in the developers' chat on Telegram.

### DIFF
--- a/mscore/inspector/inspector.cpp
+++ b/mscore/inspector/inspector.cpp
@@ -320,14 +320,19 @@ void Inspector::update(Score* s)
                         case ElementType::LAYOUT_BREAK:
                               if (toLayoutBreak(element())->layoutBreakType() == LayoutBreak::Type::SECTION)
                                     ie = new InspectorSectionBreak(this);
+#if 0 // currently empty and such not needed
                               else
                                     ie = new InspectorBreak(this);
+#endif
                               break;
                         case ElementType::BEND:
                               ie = new InspectorBend(this);
                               break;
                         case ElementType::TREMOLO:
-                              ie = new InspectorTremolo(this);
+                              if (toTremolo(element())->customStyleApplicable())
+                                    ie = new InspectorTremolo(this);
+                              else
+                                    ie = new InspectorElement(this);
                               break;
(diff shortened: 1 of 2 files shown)


Howard-C:
Just do a `for` loop for all tremolos

Jojo-Schmitz:
> Just do a `for` loop for all tremolos

?

Jojo-Schmitz:
That's exactly the complication we agreed on to not be worth the effort, as far as understood, isn't is?

mscore/inspector/inspector.cpp
```diff
@@ -320,14 +320,17 @@ void Inspector::update(Score* s)
                         case ElementType::LAYOUT_BREAK:
                               if (toLayoutBreak(element())->layoutBreakType() == LayoutBreak::Type::SECTION)
                                     ie = new InspectorSectionBreak(this);
-                              else
-                                    ie = new InspectorBreak(this);
+                              //else
+                              //      ie = new InspectorBreak(this);  // currently enpty and such not needed
```
Asmatzaile, line 324: typo?

Jojo-Schmitz: yes

Jojo-Schmitz: fixed

Howard-C:
> That's exactly the added complexity we agreed on to not be worth the effort, as far as understood, isn't is?
> 
> Getting it to work on a list selection containing 'proper' and 'improper' elements would be relatively complex

Without the complex check it's still possible to get an empty section sometimes, unless you delete the same kind of check for the specific option in `InspectorTremolo::setElement()`

Jojo-Schmitz:
> it's still possible to get an empty section sometimes

How?

Howard-C:
If the first element is applicable but the second is not, the section would be shown, but due to the check in `InspectorTremolo::setElement()`, the style option is invisible, which is the only option

Jojo-Schmitz:
I see...

PR #6393, opened 2020-07-30
Manually merged #6090 to master

anatoly-os:
@dmitrio95 could you please take a look at the compilation error? That was pretty complex merging process that broke compilation, unfortunately.

Screenshot 2020-07-30 at 15 48 40

PR #6090, opened 2020-05-16
Instruments and channels API for plugins

dmitrio95:
There is no separate issue for this but there are numerous requests for functionality provided here across the forum (the list may be incomplete):
- https://musescore.org/en/node/294183
- https://musescore.org/en/node/295638 (partially)
- https://musescore.org/en/node/294858
- https://musescore.org/en/node/56916

Ability to append instruments to a score and manipulate some playback-related properties were present in MuseScore 2.X but have not yet been implemented in MuseScore 3. This PR implements a somewhat more complete version of that, providing an access to the entire instruments and channels structure which is currently used inside MuseScore. In addition to that, both to provide a link between elements and parts and to expose an ability to mute voices, `Staff` objects are also exposed to plugins in this pull request.

PR #7081, opened 2020-12-12
[MU4] Synchronize plugin API changes from 3.x branch

dmitrio95:
This PR contains most of plugin API changes that are present in 3.x branch but have not yet been merged to master. This includes changes from the following pull requests:
 - #6090
 - #6205
 - #6722

When rebasing those changes I had to fix the conflict mentioned in #6393. The solution in this PR simply ignores the fact that `Staff` has changed its place in the inheritance hierarchy. However, as I have mentioned in [this comment](https://github.com/musescore/MuseScore/pull/6393#issuecomment-666585959), this may need to be reconsidered at some point.

There are still yet not merged changes from #6868 and #6935 (and now potentially #7079). I didn't include them to this PR as those PRs are mostly not about plugin API changes and are better to be ported to MU4 separately.

I didn't manage to launch any plugins in a build from current master neither before nor after the changes in this PR so I wasn't able to test these changes. However this PR should make it easier to revive the plugins functionality available in 3.6 version if/when the time comes to do it.


dmitrio95:
Windows build seems to fail due to to the presence of [`small` definition](https://stackoverflow.com/a/27794577/6072926) in Windows headers. Not sure what exactly caused them to appear in the scope of libmscore headers, but I'll try to undefine them for now.

src/libmscore/sym.h
```diff
@@ -21,13 +21,30 @@
 #include "ft2build.h"
 #include FT_FREETYPE_H
 
+// Needs to be duplicated here and in style.h since moc doesn't handle macros from #include'd files
+#ifdef SCRIPT_INTERFACE
+#define BEGIN_QT_REGISTERED_ENUM(Name) \
```
igorkorsukov, line 26: It might be better to declare these definitions in one place (separate h file) and use them where necessary instead of duplicating them.

dmitrio95: I have tried including it from a separate file when working on #6205 but that didn't work for me. Therefore I have duplicated the macro and add the comment (two lines above) explaining that by `moc` not handling included files.

However I have just checked it again, and it turns out that I was wrong: `moc` does process included files, and the real issue here is just that `moc` doesn't get invoked for `style.h` and `sym.h` if this macro is moved out of these files. The reason is that `moc` invocations are handled by CMake automoc mechanism which detects the files to be processed by certain criteria including presence of macros like `Q_OBJECT` and `Q_GADGET`. As `style.h` and `sym.h` don't contain it except for this macro, automoc doesn't launch moc on those files if the macro is moved out of them.

With CMake versions newer than 3.10 this can be fixed by listing the name of this macro in [AUTOMOC_MACRO_NAMES](https://cmake.org/cmake/help/latest/prop_tgt/AUTOMOC_MACRO_NAMES.html) property. This indeed works but this solution seems to be applicable only for master branch (which currently requires CMake 3.16) but not for 3.x (which sets the minimum CMake version to 3.3). I have added a commit with `AUTOMOC_MACRO_NAMES` to this PR so these definitions are no longer duplicated here.

With older versions of CMake launching moc on these files seems to require either creating dummy classes with macros recognized by automoc (like it is done in [`types.h`](https://github.com/musescore/MuseScore/blob/1304cfd1309ed98c9516a8a5697ab80f19c699af/libmscore/types.h#L630-L634)) or manually listing the relevant headers in [`qt5-wrap-cpp`](https://doc.qt.io/qt-5/qtcore-cmake-qt5-wrap-cpp.html) macro, so I guess for the current 3.x branch there is no solution that would be much better than duplicating this macro.

PR #6065, opened 2020-05-09
Plugin API: add basic access to score style settings

dmitrio95:
Provides a basic access to score style settings to plugins. As discussed in [this issue](https://musescore.org/en/node/289326#comment-963145), this would be needed to implement some basic subset of Album feature which might be of use while the full implementation of Albums is not available. Still an access to score style settings should certainly be useful for many other applications.

As for the technical side, to avoid large changes involving moving or rewriting the entire `Sid` enum I added something similar to what has been used for some enumerations in 2.X: see https://github.com/musescore/MuseScore/commit/456c26e461651979623ca4ea4e6876b239c49559. Unlike the solution in the linked commit, the solution proposed here does not require massive formatting changes to the enumeration but only requires to wrap it inside two `BEGIN_QT_REGISTERED_ENUM` and `END_QT_REGISTERED_ENUM` macros. The downside of this implementation is that it doesn't really play well with forward declarations of enumerations but this doesn't seem to be a real issue in this case. Perhaps the same technique could be used in the future to work with some other large enumerations like `SymId` with meta-object system in a similar way.

PR #7082, opened 2020-12-13
[MU3] add transpose function to pluginAPI::score

Marr11317:
Resolves: No issue in tracker

Adds ability to transpose a score from plugin.

NOTE: I still cannot build Musescore on my PC, so I will test the generated artifacts when the CI outputs them.

### DIFF
--- a/libmscore/mscore.h
+++ b/libmscore/mscore.h
@@ -15,6 +15,7 @@
 
 #include "config.h"
 #include "style.h"
+#include "types.h"
 
 namespace Ms {
 
@@ -119,22 +120,6 @@ enum class BracketType : signed char {
 
 enum class PlaceText : char {
       AUTO, ABOVE, BELOW, LEFT
-      };
-
-//---------------------------------------------------------
-//   TransposeDirection
-//---------------------------------------------------------
-
-enum class TransposeDirection : char {
-      UP, DOWN, CLOSEST
-      };
-
-//---------------------------------------------------------
-//   TransposeMode
-//---------------------------------------------------------
-
-enum class TransposeMode : char {
-      TO_KEY, BY_INTERVAL, DIATONICALLY
       };
 
 //---------------------------------------------------------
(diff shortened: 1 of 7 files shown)


Jojo-Schmitz (changes requested):

dmitrio95:
The build fails because the enumerations used in `DECLARE_API_ENUM` declarations (`Ms::TransposeMode`, `Ms::TransposeDirection` and `Ms::Key`) are not registered to the Qt meta-object system and thus there is no `QMetaEnum` for them which can be used to obtain names of the corresponding enumerators. This can be fixed using one of the following approaches:
1. Move the enumeration to `types.h` file and add `Q_ENUM_NS` declaration for this enumeration (see, for example, 54c64e1603e32c4cb16cfd59a5d46e108a7aac47). Moving to `types.h` is needed since `Q_ENUM_NAMESPACE` needs to be in the file which contains `Q_NAMESPACE` declaration which cannot be added to more than one file in the same namespace, see [QTBUG-68611](https://bugreports.qt.io/browse/QTBUG-68611).
2. Put the enumeration to some class and register it to the meta-type system with `Q_ENUM` macro. This is something which I have implemented with `BEGIN_QT_REGISTERED_ENUM` and `END_QT_REGISTERED_ENUM` macros in #6065 and #6205. Due to Qt moc limitations, those macros need to be duplicated across files which doesn't make it a good solution either. I used it in those cases since the relevant enumerations contain large number of elements and are relatively often changed in the process of MuseScore development (so moving the enumerations would make it more difficult to track changes history and do a rebase when necessary), but it may not worth it for smaller enumerations like those touched in this PR.

Of course, there may possibly be other solutions, I just listed those which are currently being used for that purpose.

Jojo-Schmitz:
Seems the `Ms::Key` is lacking that
```
      ///.\{
      ...
      ///.\}
```
'decorations', might that be the cause for the build failure?

dmitrio95:
Yes, `Key` also needs the same treatment as the other two enums.

The `///.\{`-`///.\}` thing is needed to make Doxygen display all enumeration's members in the generated documentation without having to document them all manually. Doxygen somehow doesn't display enum members which don't have documentation, but writing documentation for all of them often doesn't make sense (especially if such documentation would be trivial as enumerators names are already self-describing). So this thing exploits [`DISTRIBUTE_GROUP_DOC`](https://www.doxygen.nl/manual/config.html#cfg_distribute_group_doc) Doxygen option to add an empty documentation to all members of the enumeration so they all get listed in tables like [this one](https://musescore.github.io/MuseScore_PluginAPI_Docs/plugins/html/namespace_ms.html#a16b11be27a8e9362dd122c4d879e01ae). So this is needed just to get enumerators displayed in the documentation correctly, and this can be achieved without that by adding documentation comments to all of the enumerators.

Jojo-Schmitz:
why?

Marr11317:
I'll wait for the plugin api to be ported to the 4.x codebase before reworking on it.

PR #7091, opened 2020-12-14
fix #277756: no kerning for MuseJazzText

MarcSabatella:
Resolves: https://musescore.org/en/node/277756

When MuseJazzText was first introduced (as MuseJazz),
there was quite a bit of good kerning info.
This was lost when MuseJazzText was split of from MuseJazz -
the kern tables never made the transition.

Although the old kern information is still available in GitHub,
it turns out to no longer work well - too much has changed in the font
(in particu,ar, it seems glyph bearings have been narrowed).
So this commit creates a brand new kern table to MuseJazzText.otf,
using FontForge's autokern feature to generate some reasonable values
for a hand-selected set of character classes.


MarcSabatella:
I expect this will show some vtest diffs, which I will review.

MarcSabatella:
I approve the changes in the vtests.  They are basically all just because of the kerning between the "Ja" in the name MuseJazz :-)

PR #7092, opened 2020-12-14
fix systems jumping down page on every edit

MarcSabatella:
A recent change improved much about the staff spreading algorithm,
and one important part of that was supposed to be
saving and restoring the default system height.
However, it seems somehow the restore never happened -
a new function restoreLayout2() is provided but never called.
The result is that in each edit, other systems on the page
would get taller, adding more and more space with each edit,
eventually pushing all other systems off the page entirely.

Comments in chat indicate the restoring of default system height
was meant to happen just before distributing staves,
so I have added a call to restoreLayout2()
at the top of the existing loop through systems
at the beginning of distributeStaves().

I do still see some layout shifts where system distance varies
as you enter notes into the last measure of a system
(and then back again as you enter notes into a measure
other than the first of the next system).
Somehow the algorithm computing the system spread
seems to sometimes get a wrong idea about how far to fill the page.
But this issue was probably there before my change here,
just masked because the ever-increasing system heights
were masking this.

The problem I am fixing here is the more important one for now,
as it was pushing systems right off the page,
so you'd end up with only one per page when three or four would fit.

- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -1588,6 +1588,8 @@ static void distributeStaves(Page* page)
                   transferCurlyBracket  = false;
                   }
             else {
+                  // reset staff and system heights
+                  system->restoreLayout2();
                   bool newSystem       { true };
                   bool addSpaceAroundNormalBracket { false };
                   bool addSpaceAroundCurlyBracket  { false };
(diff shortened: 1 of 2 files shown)


njvdberg:
This actually the same PR as #7090

vpereverzev:
duplicate

MarcSabatella:
Part of what I did here is duplicated in #7090 indeed, but the additional resets I did at the top of collectPage are still needed.  I'll submit a new PR for that.

PR #7094, opened 2020-12-15
[MU3] Remove (the commented out) mtest for cut2 and cut3 timesigs

Jojo-Schmitz:
and update updateReference.sh

Yet another follow up for #7044


Jojo-Schmitz:
@Howard-C mind to double check? I don't quite get what you did (and why) with this test in 72c6b76

Howard-C:
Please don't touch that test, add a new test instead. It was for a specific bug fix as I explained in the chat.

PR #7095, opened 2020-12-15
[MU4] Added unity build

igorkorsukov:
on my machine build speed up from 04:05 to 02:18 (without ccache)

Gave a lot of problems `using namespace Ms`, I strongly discourage using it anyway.


src/context/internal/globalcontext.cpp
```diff
@@ -23,7 +23,7 @@ using namespace mu::notation;
 using namespace mu::shortcuts;
 using namespace mu::async;
 
-static const mu::Uri NOTAION_PAGE("musescore://notation");
+static const mu::Uri NOTAION_PAGE_URI("musescore://notation");
```
Jojo-Schmitz, line 26: Really NOTAION, not NOTA**T**ION?

src/context/internal/globalcontext.cpp
```diff
@@ -101,7 +101,7 @@ ShortcutContext GlobalContext::currentShortcutContext() const
 {
     if (playbackController()->isPlaying()) {
         return ShortcutContext::Playing;
-    } else if (interactive()->currentUri().val == NOTAION_PAGE) {
+    } else if (interactive()->currentUri().val == NOTAION_PAGE_URI) {
```
Jojo-Schmitz, line 104: here too

src/context/internal/globalcontext.cpp
igorkorsukov: Thanks, I will fix in next PR

Jojo-Schmitz: really a minor nitpick ;-)

PR #7096, opened 2020-12-15
[MU4] Fixed loading shortcuts

Eism:


RomanPudashkin (approved):

src/framework/shortcuts/internal/shortcutsconfiguration.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+static const std::string SHORTCUTS_FILE_NAME("shortcuts.xml");
+static const std::string SHORTCUTS_QRC_FILE_PATH(":/data/" + SHORTCUTS_FILE_NAME);
+
+mu::io::path ShortcutsConfiguration::shortcutsPath() const
+{
+    io::path path = globalConfiguration()->dataPath() + "/" + SHORTCUTS_FILE_NAME;
+    if (!fileSystem()->exists(path)) {
+        path = SHORTCUTS_QRC_FILE_PATH;
+    }
```
igorkorsukov, line 31: And if the file is there, but corrupted. Shouldn't we load the default file in this case?

Eism: thanks, fixed

igorkorsukov (approved):

PR #7099, opened 2020-12-16
[MU3] Enable Ctrl+Shift+# to add a sharp for some more non-US keyboards

Jojo-Schmitz:
like Spanish QWERTZ, Belgian and French AZERTY. Follow up for #7093 

Resolves: https://musescore.org/en/node/314375 (forum topic)

Ctrl+Shift+3 (for AZERTY) is also the shortcut for switching to voice 3, but there should not be a collision, as this one here is only used in text edit mode, while the other is not. @jeetee , please test ;-)
@Asmatzaile please test your Spanish QWERTZY too...

### DIFF
--- a/libmscore/textedit.cpp
+++ b/libmscore/textedit.cpp
@@ -423,6 +423,8 @@ bool TextBase::edit(EditData& ed)
                         case Qt::Key_NumberSign: // e.g. QWERTY (US)
                         case Qt::Key_AsciiTilde: // e.g. QWERTY (GB)
                         case Qt::Key_Apostrophe: // e.g. QWERTZ (DE)
+                        case Qt::Key_periodcentered: // e.g. QWERTY (ES)
+                        case Qt::Key_3: // e.g. AZERTY (FR, BE)
                               s = "\u266f"; // Unicode sharp
                               break;
                         case Qt::Key_H:


jeetee:
Change voice is Ctrl+Alt+3, which only works using the 3 form the numerical pad for Belgian AZERTY as that keystroke using top row keys (Ctrl+Alt+") is seen as '#'

Using this PR, I can enter the sharp into lyrics using either Ctrl+Shift+" or using Ctrl+Alt+Shift+". Using Ctrl+Shift+3(numpad) doesn't seem to work. Side remark is that on my laptop there is no real numpad, so this might be due to the Fn key as well.
Ctrl+Alt+" still enters the normal hash (as expected).

So from my end, this seems okayish enough.

Interestingly enough, here's what is being reported by MuseScore when trying to define those shortcuts:
```
Ctrl+Shift+"           →     Ctrl+3 ("conflict with triplet")
Ctrl+Alt+Shift+"    →     Ctrl+Alt+3 ("conflict with voice3", but doesn't work in practice to change to voice3 😕 )
Ctrl+Alt+3             →     NumPad+Ctrl+Alt+3 (no conflict definition, but in practice this does works for voice3 😕 )
Ctrl+Alt+Shift+3   →     NumPad+Ctrl+Alt+PgDown
Ctrl+Shift+3          →     NumPad+Ctrl+PgDown
```
God knows where the PgDown comes from, likely from full numpad functions, but that's not even labeled on my laptop 🤷

Jojo-Schmitz:
OK, thanks for checking. So from your side this change is OK, right?

jeetee:
Yes, this change is good.

Asmatzaile:
Thank you! Now it works as expected 🎉

PR #6928, opened 2020-11-24
Fix #313195: Add Synalepha shortcut

Asmatzaile:
Resolves: *https://musescore.org/en/node/313195*

Synalephas are common enough to have their own dedicated shortcut. ~~As suggested by @jeetee , I added an action to the shortcut.cpp listing.~~ I changed the textedit.cpp file.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [ ] I made sure the code compiles on my machine
- [ ] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7100, opened 2020-12-16
[MU4] Enable Ctrl+Shift+# to add a sharp for some non-US keyboards, add Synalepha shortcut

Jojo-Schmitz:
* Enable Ctrl+Shift+# to add a sharp for some non-US keyboards like German QWERTY, British and Spanish QWERTZ, Belgian and French AZERTY
* Fix #313195: Add Synalepha shortcut (so far only in 'normal text edit mode, but not for Lyrics, watch this space...)

Resolves: https://musescore.org/en/node/314375 (forum topic) and https://musescore.org/en/node/313195

This is for master what #6928, #7093. #7099, #7105 are for 3.x

### DIFF
--- a/src/libmscore/textedit.cpp
+++ b/src/libmscore/textedit.cpp
@@ -448,7 +448,11 @@ bool TextBase::edit(EditData& ed)
             case Qt::Key_B:
                 s = "\u266d";               // Unicode flat
                 break;
-            case Qt::Key_NumberSign:
+            case Qt::Key_NumberSign: // e.g. QWERTY (US)
+            case Qt::Key_AsciiTilde: // e.g. QWERTY (GB)
+            case Qt::Key_Apostrophe: // e.g. QWERTZ (DE)
+            case Qt::Key_periodcentered: // e.g. QWERTY (ES)
+            case Qt::Key_3: // e.g. AZERTY (FR, BE)
                 s = "\u266f";               // Unicode sharp
                 break;
             case Qt::Key_H:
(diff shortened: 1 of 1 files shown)


Asmatzaile:
editlyrics.cpp could be updated with your fix :)

Jojo-Schmitz:
I will.

Jojo-Schmitz:
Hmm, I don't find the place to add this, there's no 'editlyrics.cpp' anymore it seems?!?

Asmatzaile:
I'm not sure but I think some relevant code could be at libmscore/lyrics.cpp, lines 454-460

Jojo-Schmitz:
Hmm, I don't think https://github.com/musescore/MuseScore/blob/dfa0d89ee388a7a229c405984f5d4449cd277580/src/libmscore/lyrics.cpp#L454-L461 is the right place. Similar code is in 3.x (https://github.com/musescore/MuseScore/blob/e5396e2ec81414cabe29bd4934e50d6dddbb64cd/libmscore/lyrics.cpp#L460-L465) and there it wasn't thig right plase

Asmatzaile:
Hm, maybe lyric editing hasn't been added to mu4 yet. I tried the ctrl+L shortcut in the latest nightly and it didn't work.

Jojo-Schmitz:
What does Ctrl+L do?

Asmatzaile:
Oh, that's true.
The functionality that makes a dash and goes to the next note when hitting the minus key doesn't work however.

PR #7102, opened 2020-12-16
[MU3] Updated url for tx2s3

igorkorsukov:
Updated url for tx2s3


Jojo-Schmitz:
On top you'd also need to fill that new location with live, i.e. at least coppy accross avwerything from the prevois location (translations and extensions)

Jojo-Schmitz:
And the mechanism that pulls updates from Transifex needs to get fixed (for 3.5 and 3.6, althpough it might be too late for the former, depending on the next step, see below)

And finally the updates strings need to get pushed up to Transifex

PR #7103, opened 2020-12-16
[MU4] Fixed tx2s3 workflow

igorkorsukov:
Fixed tx2s3 workflow

### DIFF
--- a/.github/workflows/ci_tx2s3.yml
+++ b/.github/workflows/ci_tx2s3.yml
@@ -62,10 +62,12 @@ jobs:
           echo "not set credentials, not possible run"
         fi 
         
-        if [ ${{ github.event_name != 'schedule' }} ]; then 
-          echo "This is job only for schedule run tx2s3 for 3.x branch"
-          DO_RUN='false'; 
-        fi
+        echo "github.event_name: ${{ github.event_name }}"
+
+        # if [ ${{ github.event_name != 'schedule' }} ]; then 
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Strange that fixing something on master will work for updating translations for 3.x... but parently it does!

Jojo-Schmitz:
Hmm, no, it apparently does not, it does not grab recently updated translations from Tnasifex and does not makes them available to the tranmslations updater in MuseScore

igorkorsukov:
> Strange that fixing something on master will work for updating translations for 3.x... but apparently it does!

scheduled workflows run only in the master, 
so tx2s3 workflow runs in the master, checkout on 3.x branch and run scripts from the 3.x branch. 
(same work nightly builds for 3.x)

igorkorsukov:
> Hmm, no, it apparently does not, it does not grab recently updated translations from Tansifex and/or does not make them available to the translations updater in MuseScore, not for 3.5.2, 3.6.0 Beta nor the laetest self built

Now tx2s3 uploading to 3.6 version, but URL for 3.5 version in the source code. Vasiliy fixed it. But I don't know what url for you. 
Anyway, we'll figure it out tomorrow

Jojo-Schmitz:
Ah, I see, resourcemanager.h still uses 3.5: `static inline QString baseAddr() { return "http://extensions.musescore.org/3.6/"; }`

Jojo-Schmitz:
But not even when fixing that it finds updated translations...

PR #7110, opened 2020-12-17
fix https://trello.com/c/P9b373re/73-atonal-key-signature-causes-spacing-issues

MarcSabatella:
Resolves: https://trello.com/c/P9b373re/73-atonal-key-signature-causes-spacing-issues

A recent change to how invisible key and time signatures are handled
exposed some issues in other parts of the code that should have been
ignoring invisible signatures but weren't.
As a result, things likes starts of ties, slurs, and melisma lines
that were continued from the previous system
were being incorrectly calculated, causing them to be too short.
They should start right after the clef or last visible header element
but were instead starting after the invisible key signature,
which are a result of the recent change,
now coincides with the first note.

It turns out all these elements used different calculations
to determine the start point, and while they all used similar code,
there were also discrepancies that make no sense we can see.
So, I have taken the code that was used for slurs only
and made ties and melisma lines reference that as well.
I also added some enhancements to that code
that had been previously made for ties only
(better handling elements that are not left-aligned in their segments).

Centering of measure rests, multimeasure rests, and measure repeats
was also affected, but here the code was different enough
that I could not simply reuse the same function.
Instead I made the relevant change directly to that calculation.
That being, skipping segments that have all elements invisible.
Luckily a nice convenient allElementsInvisible function was provided
by the same PR that changed the position of these invisible elements.

vtests added.


MarcSabatella:
system-10

MarcSabatella:
system-11

MarcSabatella:
Hmm, I didn't actually enable the first system indent on those, but the command-line generate of the reference did it for me.  Not sure if that will mess anything up.

MarcSabatella:
The vtest failure here is expected, on @oktophonie 's recommendation, I changed the algorithm to add a margin after the clef & key signature when continuing a tie.  One of the existing vtests was expecting the old result.

Meanwhile, I've pushed a version rebased on current 3.x.  It seems command-line conversion is still not correct with respect to styles - my vtest files are explicitly set to *not* indent, but I'm still getting the intent.

PR #7116, opened 2020-12-19
Styles application issues

vpereverzev:
resolves - https://trello.com/c/sabysJAY

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -3560,15 +3560,11 @@ static void loadScores(const QStringList& argv)
             foreach(const QString& name, argv) {
                   if (name.isEmpty())
                         continue;
-                  MasterScore* score = mscore->readScore(name);
-                  if (score) {
-                        mscore->appendScore(score);
+
+                  Score* score = mscore->openScore(name);
+
+                  if (score)
                         scoresOnCommandline = true;
-                        if(!MScore::noGui) {
-                              mscore->addRecentScore(score);
-                              mscore->writeSessionFile(false);
-                              }
-                        }
                   }
             }
 
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -3557,7 +3557,7 @@ static void loadScores(const QStringList& argv)
                   }
             }
       else {
-            foreach(const QString& name, argv) {
+            for(const QString& name : argv) {
                   if (name.isEmpty())
                         continue;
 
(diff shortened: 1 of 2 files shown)


mscore/musescore.cpp
```diff
@@ -3560,15 +3560,11 @@ static void loadScores(const QStringList& argv)
...
                         continue;
-                  MasterScore* score = mscore->readScore(name);
-                  if (score) {
-                        mscore->appendScore(score);
+
+                  Score* score = mscore->openScore(name);
+
+                  if (score)
                         scoresOnCommandline = true;
-                        if(!MScore::noGui) {
```
Jojo-Schmitz, line 3567: Isn't double-clicking a score also affected by this? It is on the command line in that case, but in GUI mode, as far as I can  tell

vpereverzev: Yep, I didn't mention it in the commit message, but it also affect such a case.

Jojo-Schmitz: But then it should add the score to the list of recently used ones, shouldn't it? Or does it still?

Jojo-Schmitz: Also while at it, you could change that `foreach()` ;-)

Jojo-Schmitz:
Strange Mac build fail, doesn't seem to be related to this PR but rather a general issue?

mscore/musescore.cpp
vpereverzev: the openscore method already does all this, so I just cleaned it up here as unnecessary

Jojo-Schmitz: I see, sorry for the noise then

vpereverzev:
> Strange Mac build fail, doesn't seem to be related to this PR but rather a general issue? See https://github.com/musescore/MuseScore/pull/7116/checks#step:5:43

there is a known issue with python installation via brew, we're trying to figure out how to resolve this

MarcSabatella:
Currently it seems the style settings don't work correctly on the command line.  If I explicitly set no first system indent in a 3.6 score, command line PNG conversion nevertheless adds an indent.  This show up in the new vtests I tried adding for PR #7110.  I can't tell if this fix will address that also.  But I think so?

vpereverzev:
> 
> 
> Currently it seems the style settings don't work correctly on the command line. If I explicitly set no first system indent in a 3.6 score, command line PNG conversion nevertheless adds an indent. This show up in the new vtests I tried adding for PR #7110. I can't tell if this fix will address that also. But I think so?

Yes, I think this fix would help you

PR #7117, opened 2020-12-19
[MU3] Fixed install python 3.9 error on MacOS CI

igorkorsukov:
Fixed install python 3.9 error on MacOS CI

### DIFF
--- a/build/ci/macos/setup.sh
+++ b/build/ci/macos/setup.sh
@@ -16,6 +16,9 @@ rm bottles/freetype* | $SKIP_ERR_FLAG
 
 brew update >/dev/null | $SKIP_ERR_FLAG
 
+# fixing install python 3.9 error (it is a dependency for JACK)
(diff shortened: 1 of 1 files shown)


build/ci/macos/setup.sh
```diff
@@ -16,6 +16,9 @@ rm bottles/freetype* | $SKIP_ERR_FLAG
 
 brew update >/dev/null | $SKIP_ERR_FLAG
 
+# fixing install python 3.9 error (it is dependency for jack)
```
RobFog, line 19: → "a dependency"

RobFog (changes requested):

build/ci/macos/setup.sh
igorkorsukov: > → "a dependency", "JACK"?

Thanks, fixed.

PR #6914, opened 2020-11-23
[MU3] Fix #311340: Save online via command line (--save-online)

shoogle:
Backport PR #6799 and PR #6908 to the 3.x branch. Special treat for @Jojo-Schmitz.

PR #7118, opened 2020-12-19
[MU3] Fix printing user errors in musescore.cpp

orivej:
Resolves: https://github.com/musescore/MuseScore/pull/6914#discussion_r546256995

v3.6_public_beta fails to build with Nixpkgs because by default it enables `-Wformat -Wformat-security -Werror=format-security`, which does not accept `qFatal(qUtf8Printable(tr("…")))`. This fixes it by replacing qFatal with fprintf. (See also commit description.)

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -2001,7 +2001,8 @@ MuseScore::MuseScore()
             const bool loadSuccess = _loginManager->load();
 
             if (cliSaveOnline && !loadSuccess) {
-                  qFatal(qUtf8Printable(tr("No login credentials stored. Please sign in via the GUI.")));
+                  fprintf(stderr, "%s\n", qPrintable(tr("No login credentials stored. Please sign in via the GUI.")));
+                  ::exit(EXIT_FAILURE);
                   }
             }
 
@@ -7768,7 +7769,8 @@ MuseScoreApplication::CommandLineParseResult MuseScoreApplication::parseCommandL
             MScore::noGui = true;
 
             if (parser.positionalArguments().isEmpty()) {
-                  qFatal("Must specify at least one score to save online.");
+                  fprintf(stderr, "%s\n", qPrintable(tr("Must specify at least one score to save online.")));
+                  ::exit(EXIT_FAILURE);
             }
       }
 
@@ -7822,8 +7824,10 @@ MuseScoreApplication::CommandLineParseResult MuseScoreApplication::parseCommandL
             diffMode = true;
             }
       if (parser.isSet("run-test-script")) {
-            if (rawDiffMode || diffMode)
-                  qFatal("incompatible options");
+            if (rawDiffMode || diffMode) {
+                  fprintf(stderr, "%s\n", qPrintable(tr("--run-test-script is incompatible with --diff and --raw-diff")));
+                  ::exit(EXIT_FAILURE);
+                  }
             MScore::noGui = true;
             scriptTestMode = true;
             }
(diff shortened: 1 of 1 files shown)


mscore/musescore.cpp
```diff
@@ -2001,7 +2001,7 @@ MuseScore::MuseScore()
             const bool loadSuccess = _loginManager->load();
 
             if (cliSaveOnline && !loadSuccess) {
-                  qFatal(qUtf8Printable(tr("No login credentials stored. Please sign in via the GUI.")));
+                  qFatal("No login credentials stored. Please sign in via the GUI.");
```
Jojo-Schmitz, line 2004: Veto. This way that text is not translatable

orivej: No other qFatal is translatable. (I don't think that qFatal, qWarning, and other developer messages need translation.)

Jojo-Schmitz: IMHO they do, all of them

Jojo-Schmitz: Not the qDebug() ones though, not sure about qWarning().

Jojo-Schmitz: If at that place the translations engine is available, which for example it is not at the place wherethe Usage message is printed

Jojo-Schmitz: At this place translations are avaibable though, as far as I can tell.

Jojo-Schmitz: How about `qFatal("%s", qUtf8Printable(tr("No login credentials stored. Please sign in via the GUI.")));`

shoogle: @orivej, does this work:

```c++
QString msg = tr("No login credentials stored. Please sign in via the GUI.");
qFatal(qUtf8Printable(msg));
```
Or
```c++
auto msg = qUtf8Printable(tr("No login credentials stored. Please sign in via the GUI."));
qFatal(msg);
```
?

Jojo-Schmitz: Or that, yes

orivej: I meant that no other qFatal (or qWarning, or qDebug) uses `tr` and therefore is not translatable. Why does this specific qFatal have to be the only one with `tr`?

Jojo-Schmitz: Because the other `qFatal()`s should too

Jojo-Schmitz: Esp. those that are related to running MuseScore from the commandline (does `qFatal()` bring up a dialog in GUI mode?)

orivej: `qFatal(qUtf8Printable(msg))` and `qFatal(msg)` do not pass `-Wformat-security`: it protects against the case where the translated msg contains `%s` or other special formatting directives.

`qFatal("%s", …)` works. However, qFatal should not be used for user error reporting (which seems to be the case here), it should be used for violated developer assumptions (like `abort()` in C) because it creates a core dump: https://doc.qt.io/qt-5/qtglobal.html#qFatal . This is also why it should not be translated: it is easier to grep the code for the error message rather than to correlate the translated error message to the original.

Jojo-Schmitz: Ok, so here we might better be using `fprintf(stderr, "...")` or `std::cerr << ...`

shoogle: @Jojo-Schmitz, what about `qWarning() << "..."` followed by `exit(EXIT_FAILURE)`?

Jojo-Schmitz: Would work too I guess. Not sure though whether either of these functions prefixes the (error) message with "Fatal:", "Warning:" or some such

Jojo-Schmitz: and whether and how it works on Windows commandline (but there I'm not sure about `cerr?  and the likes either)

orivej: OK. I had to use `::exit` because `exit` resolves into `QCoreApplication::exit()` (exit from the main loop), which does nothing when the main loop is not running.

Jojo-Schmitz (approved):

PR #7121, opened 2020-12-19
Fix #314149: Order of buttons in ScoreMigrationDialog platform specific

cbjeukendrup:
Resolves: https://musescore.org/en/node/314149 = https://trello.com/c/FAckEYaZ/60-buttons-are-backwards-on-apply-new-styles-dialog-on-windows-see-https-musescoreorg-en-node-314149

On some platforms, among which Windows, the order of the buttons in the Score Migration Dialog was not like usually on those platforms. This is (should be) solved by using QML's DialogButtonBox, which automatically takes the platform into account.

Ironically, I'm on macOS, and can't test it on Windows. Maybe I can do that tomorrow, but if somebody else would have time to try it, that would be great!

### DIFF
--- a/mscore/qml.qrc
+++ b/mscore/qml.qrc
@@ -1,13 +1,24 @@
 <RCC>
     <qresource prefix="/">
+        <file>qml/migration/CheckBoxControl.qml</file>
+        <file>qml/migration/placeholder.png</file>
+        <file>qml/migration/ScoreMigrationDialog.qml</file>
+        <file>qml/migration/StyledDialogButtonBox.qml</file>
+        <file>qml/migration/Tick_Icon.png</file>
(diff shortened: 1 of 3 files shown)


Jojo-Schmitz:
Tested the artifact, the dialog (on Windows) looks like this with this PR:
image
Before it looked like this:
image

Jojo-Schmitz (approved):

Jojo-Schmitz:
What I'm missing in that dialog:

1.  naming the fonts, Leland and Edwin, respectivly.
2. having a default button, "Keep".
    a) changing to "Apply" once any of those check boxes in the top section are checked

cbjeukendrup:
@Jojo-Schmitz Both sound like very good ideas, IMO. I will have a look.

PR #7125, opened 2020-12-20
Avoid deprecated assignment past the end of QString

orivej:
https://doc.qt.io/qt-5/qstring.html#operator-5b-5d says this about
QCharRef QString::operator[](int position):

"Before Qt 5.14 it was possible to use this operator to access a
character at an out-of-bounds position in the string, and then assign
to such a position, causing the string to be automatically
resized. These behaviors are deprecated, and will be changed in a
future version of Qt."

Corresponding v3 PR: #7126

### DIFF
--- a/src/libmscore/chordlist.cpp
+++ b/src/libmscore/chordlist.cpp
@@ -515,7 +515,7 @@ bool ParsedChord::parse(const QString& s, const ChordList* cl, bool syntaxOnly, 
 #endif
     int lastLeadingToken;
     int len = s.size();
-    int i, j;
+    int i;
     int thirdKey = 0, seventhKey = 0;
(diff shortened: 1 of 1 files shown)


src/libmscore/chordlist.cpp
```diff
@@ -872,14 +872,14 @@ bool ParsedChord::parse(const QString& s, const ChordList* cl, bool syntaxOnly,
             ++i;
         }
         // get second token - a number <= 13
```
orivej, line 874: Actually this code accepts any number than starts with 10, 11, 12, or 13, such as 12345. I did not attempt to fix it in case it breaks some usage.

Jojo-Schmitz (approved):

PR #7126, opened 2020-12-20
[MU3] Avoid deprecated assignment past the end of QString

orivej:
https://doc.qt.io/qt-5/qstring.html#operator-5b-5d says this about
QCharRef QString::operator[](int position):

"Before Qt 5.14 it was possible to use this operator to access a
character at an out-of-bounds position in the string, and then assign
to such a position, causing the string to be automatically
resized. These behaviors are deprecated, and will be changed in a
future version of Qt."

Corresponding v4 PR: #7125

### DIFF
--- a/libmscore/chordlist.cpp
+++ b/libmscore/chordlist.cpp
@@ -487,7 +487,7 @@ bool ParsedChord::parse(const QString& s, const ChordList* cl, bool syntaxOnly, 
 #endif
       int lastLeadingToken;
       int len = s.size();
-      int i, j;
+      int i;
       int thirdKey = 0, seventhKey = 0;
       bool susChord = false;
       QList<HDegree> hdl;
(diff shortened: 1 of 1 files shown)


libmscore/chordlist.cpp
```diff
@@ -617,11 +618,11 @@ bool ParsedChord::parse(const QString& s, const ChordList* cl, bool syntaxOnly,
...
       lastLeadingToken = _tokenList.size();
       // get extension - up to first non-digit other than comma or slash
-      for (j = 0, tok1 = ""; i < len; ++i, ++j) {
+      for (tok1 = ""; i < len; ++i) {
             if (!s[i].isDigit() && s[i] != ',' && s[i] != '/')
                   break;
-            tok1[j] = s[i];
-            }
+            tok1.push_back(s[i]);
+      }
```
Jojo-Schmitz, line 625: this is changeing the old 3.x code style. Not a big deal though, more a nitpick...

Jojo-Schmitz (approved):

libmscore/chordlist.cpp
orivej: Oops, I've brought the old style back.

PR #7133, opened 2020-12-21
[MU4] Fix #314612: Erroneous duplicate check (actually 4 times) of an identical condition

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/314612

@psmokotnin : please review


RobFog:
Errnous → Erronous

Jojo-Schmitz:
I buy the first, but the second

psmokotnin (approved):

psmokotnin:
@Jojo-Schmitz Thanks for the fix

Jojo-Schmitz:
Guess we'd rather have to thank [maddin200](https://musescore.org/en/user/1622771) for finding this

PR #7141, opened 2020-12-21
fix #314633: inconsistent stem directions on middle-line beamed notes

MarcSabatella:
Resolves: https://musescore.org/en/node/314633

This code has been around a long time and does what it sets out to do,
but there seems to be agreement it isn't desired.
When setting the stem direction for two note beam groups
where both notes are on or equidistant from the middle line,
we look to a previous beam group to set the stem direction,
rather than just always pointing down like we do for
single notes on the middle line or larger beamed groups.

Fix consists of removing that code.

vtest added.


MarcSabatella:
Turns out there *were* a few tests that already covered this case.  I believe I have fixed the refs for the mtests that failed, hopefully that passes this time.  But the vtest beams-16 is expected to fail now, the beamed pair on the third beat of the second measure is one of the cases this code was causing to have stems up for not really a good reason.

PR #7142, opened 2020-12-22
[MU4] Fix compiler warning reg. `size_t` vs. `int`

Jojo-Schmitz:
@Eism please review ;-)

### DIFF
--- a/src/notation/view/noteinputbarcustomisemodel.cpp
+++ b/src/notation/view/noteinputbarcustomisemodel.cpp
@@ -111,7 +111,7 @@ static size_t indexOf(const ActionList& actions, const ActionName& actionName)
         }
     }
 
-    return -1;
+    return static_cast<size_t>(-1);
 }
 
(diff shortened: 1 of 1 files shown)


Eism:
Hm, static_cast<size_t>(-1) it also not a valid number.
it's better to avoid using size_t to get the index of an element 
just replace it with static int indexOf
and 
return static_cast<size_t>(i);

how I love std :)

Jojo-Schmitz:
Well, an index is always positive, isn't it?

Jojo-Schmitz:
`static_cast<size_t>(-1)` as far as I can tell is FFFF(FFFF), on any two's complement machine?

Eism:
Yes, if an element is found. size_t is needed for dimension, not for indexing if I understood correctly.
I did as in standard Qt containers (QList).

igorkorsukov:
replaced by https://github.com/musescore/MuseScore/pull/7164

PR #7143, opened 2020-12-22
[MU3] Fix #314655: Avoid spaces and more when exporting files

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/314655

Also avoids that "Success" dialog, which is an annoying extra click with no value IHMO

### DIFF
--- a/mscore/exportdialog.cpp
+++ b/mscore/exportdialog.cpp
@@ -446,10 +446,10 @@ void ExportDialog::accept()
       QString name = QString("%1/%2").arg(saveDirectory, cs->masterScore()->fileInfo()->completeBaseName());
       if (oneScore) {
             Score* score = scores.first();
-            name.append(QString(" - %1").arg(score->isMaster() ? tr("Full Score") : score->title()));
+            name.append(QString("%1").arg(score->isMaster() ? "" : "-" + score->title()));
             }
       else if (singlePDF)
-            name.append(QString(" - %1").arg(containsMasterScore ? tr("Score and Parts") : tr("Parts")));
+            name.append(QString("-%1").arg(containsMasterScore ? tr("Score_and_Parts") : tr("Parts")));
 
 #ifdef Q_OS_WIN
       if (QOperatingSystemVersion::current() > QOperatingSystemVersion(QOperatingSystemVersion::Windows, 5, 1))    // XP
@@ -485,10 +485,10 @@ void ExportDialog::accept()
             SaveReplacePolicy replacePolicy = SaveReplacePolicy::NO_CHOICE;
 
             for (Score* score : scores) {
-                  QString definitiveFilename = QString("%1/%2 - %3.%4")
+                  QString definitiveFilename = QString("%1/%2%3.%4")
                         .arg(fileinfo.absolutePath(),
                              fileinfo.completeBaseName(),
-                             score->isMaster() ? tr("Full Score") : score->title(),
+                             score->isMaster() ? "" : "-" + score->title(),
                              suffix);
                   if (saveFormat != "png" && saveFormat != "svg" && QFileInfo(definitiveFilename).exists()) {
                         // Png and Svg export functions change the filename, so they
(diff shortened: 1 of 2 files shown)


mscore/exportdialog.cpp
```diff
@@ -444,12 +444,12 @@ void ExportDialog::accept()
...
-      if (oneScore) {
+      /*if (oneScore) {
             Score* score = scores.first();
-            name.append(QString(" - %1").arg(score->isMaster() ? tr("Full Score") : score->title()));
+            name.append(QString("%1").arg(score->isMaster() ? "" : "-" + score->title()));
             }
-      else if (singlePDF)
-            name.append(QString(" - %1").arg(containsMasterScore ? tr("Score and Parts") : tr("Parts")));
+      else*/ if (singlePDF)
+            name.append(QString("-%1").arg(containsMasterScore ? tr("Score_and_Parts") : tr("Parts")));
```
Jojo-Schmitz, line 452: This is the name that was used for this pre 3.6.

mscore/file.cpp
```diff
@@ -2023,7 +2023,7 @@ bool MuseScore::savePdf(Score* cs_, QPrinter& printer)
             QString partname = cs_->metaTag("partName");
             if (partname.isEmpty()) // partName unset?
                   partname = cs_->title(); // fall back to excerpt's tab title
-            title += " - " + partname;
+            title += "-" + partname;
```
Jojo-Schmitz, line 2026: Not too sure here, this might not be used as a filename.

mscore/file.cpp
```diff
@@ -2791,7 +2791,7 @@ bool MuseScore::saveSvg(Score* score, const QString& name, const NotesColors& no
             QString fileName(name);
             if (fileName.endsWith(".svg"))
                   fileName = fileName.left(fileName.size() - 4);
-            fileName += QString(" - %1.svg").arg(pageNumber+1, padding, 10, QLatin1Char('0'));
+            fileName += QString("-%1.svg").arg(pageNumber+1, padding, 10, QLatin1Char('0'));
```
Jojo-Schmitz, line 2794: this looks like an old leftover, those spaces arounf the - had been removed for PNG a long time ago IIRC

mscore/exportdialog.cpp
```diff
@@ -519,8 +519,8 @@ void ExportDialog::accept()
                         success = false;
                   }
             }
-      if (success)
-            QMessageBox::information(this, tr("Export"), tr("Export was successful."));
+      if (!success)
+            QMessageBox::critical(this, tr("Export"), tr("Export was not successful."));
```
Jojo-Schmitz, line 523: No news are good news...

mscore/exportdialog.cpp
```diff
@@ -446,10 +446,10 @@ void ExportDialog::accept()
       QString name = QString("%1/%2").arg(saveDirectory, cs->masterScore()->fileInfo()->completeBaseName());
       if (oneScore) {
             Score* score = scores.first();
-            name.append(QString(" - %1").arg(score->isMaster() ? tr("Full Score") : score->title()));
+            name.append(QString("%1").arg(score->isMaster() ? "" : "-" + score->title()));
```
Jojo-Schmitz, line 449: Not sure when this part is used, didn't notice a different behavoir when I had this part commented out?

mscore/exportdialog.cpp
cbjeukendrup, line 523: If the export is *not* successful, `mscore->saveAs` already presents a warning to the user. So no need to do it here another time. 

@Tantacrul told me that he preferred to keep the "Export successful" message.

Tantacrul: Yes. I used it a few times recently and found that the dialog can hang for a little while (If I'm exporting a long audio file for example). At that point, the 'successful' dialog, although a bit dated, was reassuring. Otherwise the user may be led to believe the process failed. In future, it would be ideal to create a progress bar-type solution.

Jojo-Schmitz: For longer exports like MP3 we do have that progres bar. Having to dismiss this success dialog is highly annoying

Jojo-Schmitz: For longer exports like MP3 and WAV we do have that progres bar (and ever had)
Having to dismiss this success dialog is highly annoying

Jojo-Schmitz: reverted and commented out...

mscore/file.cpp
Jojo-Schmitz: So leaving it in

PR #7144, opened 2020-12-22
FIxed an issue with styles applying for "recent scores"

vpereverzev:
Resolves: https://trello.com/c/ALyDotX0
The same situation as in #7116 , I wonder how many such "duplications" we still have


shoogle:
> I wonder how many such "duplications" we still have

Maybe make the `readScore()` method private and put it somewhere that people can't use it accidentally when they should use `openScore()` instead?

PR #7145, opened 2020-12-22
[MU3] Fix #286153: Check box "Show MIDI controls in mixer" in Preferences/Score page has no effect

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/286153

### DIFF
--- a/mscore/mixer/mixer.cpp
+++ b/mscore/mixer/mixer.cpp
@@ -76,7 +76,7 @@ char userRangeToReverb(double v) { return (char)qBound(0, static_cast<int>(v / 1
 
 Mixer::Mixer(QWidget* parent)
     : QDockWidget(qApp->translate("Mixer", "Mixer"), parent),
-      showDetails(true),
+      showDetails(preferences.getBool(PREF_IO_MIDI_SHOWCONTROLSINMIXER)),
       trackHolder(nullptr)
       {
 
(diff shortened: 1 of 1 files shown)


mscore/mixer/mixer.cpp
```diff
@@ -76,7 +76,7 @@ char userRangeToReverb(double v) { return (char)qBound(0, static_cast<int>(v / 1
 
 Mixer::Mixer(QWidget* parent)
     : QDockWidget(qApp->translate("Mixer", "Mixer"), parent),
-      showDetails(true),
+      showDetails(preferences.getBool(PREF_IO_MIDI_SHOWCONTROLSINMIXER)),
```
Jojo-Schmitz, line 79: Somehow this doesn't seem to result in the prefs being read and set when opening the Mixer for the 1st time

mscore/mixer/mixer.cpp
```diff
@@ -503,6 +504,7 @@ void MuseScore::showMixer(bool visible)
 
       if (mixer == 0) {
             mixer = new Mixer(this);
+            mixer->showDetailsToggled(preferences.getBool(PREF_IO_MIDI_SHOWCONTROLSINMIXER));
```
Jojo-Schmitz, line 507: This is what brings it to senses on the 1st open.

PR #7147, opened 2020-12-22
fix https://trello.com/c/P9b373re - bad start point for continued lines

MarcSabatella:
Resolves: https://trello.com/c/P9b373re

And earlier PR fixed this issue for ties, slurs, and melisma lines.
This fixes it for other lines types,
using the same utility function System::firstNoteRestSegmentX()
to ensure a consistent start point for all such lines.

I expect this change will affect a few tests, such as line-2 and line-3.  These changes should be improvements - a consistent margin is now added after the header before the continuation line, rather than being sometimes too big and sometimes too small.

### DIFF
--- a/libmscore/glissando.cpp
+++ b/libmscore/glissando.cpp
@@ -296,8 +296,12 @@ void Glissando::layout()
             || cr2->noteType() == NoteType::GRACE16_AFTER || cr2->noteType() == NoteType::GRACE32_AFTER)
          // also ignore if cr1 is a child of cr2, which means cr1 is a grace-before of cr2
          && !(cr1->parent() == cr2)) {
-            segm2->rxpos() -= GLISS_STARTOFSYSTEM_WIDTH * _spatium;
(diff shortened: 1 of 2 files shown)


MarcSabatella:
The line-2 and line-3 vtest failures were expected, but not the gliss-1.  Glissando continuation has long been off, actually overlapping the header.  But now it's worse somehow.  Will investigate.

MarcSabatella:
I understand the glissando issue, it's really a direct result of the same change here.  I will push an update tomorrow

MarcSabatella:
The situation with glissandi is that there was once code to actually add more space before the first note of the system to accommodate them.  This worked in 2.3.2 but hasn't worked for years since, probably due to so much else changing for MuseScore 3.  End result is that continue glissandi were in fact overlapping the header (clef an/or key signature).  Not good at all.

Figuring out how to re-enable the code adding extra space is not trivial, I think.

So instead, I've compromised by making the glissando continuation start immediately after the header, without the 1 sp margin.

PR #7148, opened 2020-12-22
[MU3] fix #314665: don't force "long" slurs above the staff

MarcSabatella:
Resolves: https://musescore.org/en/node/314665

NOTE: I am confident the code itself is good,
not as confident we want to make this change for 3.6,
at least not without adding more compatibility handling.
But I wanted to make this available for testing and discussion.

We have code to force "long" slurs above the staff,
which isn't a terrible idea, but we're way too heavy-handed about it.
Anything more than one measure in duration is considered long.

Since this determination is best made subjectively by the user,
the code here removes this check.
However, it will doubtless have an efrect on existing scores.
In most cases, a good one, as slurs that were erroneously forced up
will now be where they really belong, below the staff.
In some cases, though - especially truly long slurs
like ones extending for most of the system -
it could cause too much extra space below the staff.

Ultimately we could consider adding a style setting to control this.
And having the default different for older versus newer scores.

### DIFF
--- a/libmscore/slur.cpp
+++ b/libmscore/slur.cpp
@@ -1049,11 +1049,13 @@ SpannerSegment* Slur::layoutSystem(System* system)
                         _up = !(startCR()->up());
 
                         Measure* m1 = startCR()->measure();
+#if 0
+                        // the following code was in place until 3.6,
+                        // to force "long" slurs (duration > one measure) above
+                        // but it's much too aggressive - one measure isn't necessarily long
                         if ((endCR()->tick() - startCR()->tick()) > m1->ticks()) // long slurs are always above
                               _up = true;
-                        else
-                              _up = !startCR()->up();
-
+#endif
                         if (c1 && c2 && isDirectionMixture(c1, c2) && !c1->isGrace()) {
                               // slurs go above if start and end note have different stem directions,
                               // but grace notes are exceptions


Jojo-Schmitz:

MarcSabatella:
Well, I assume the same change should apply there.  Although at the time, the belief was that the slur algorithms were going to be rewritten more completely.

BTW, this PR probably should not have been merged for 3.x.  Or at least, it's one that should not be included in any hypothetical 3.6.3, as it has too great a potential effect on existing scores.

Jojo-Schmitz:
It doesn't case any vtest issues as far as I can tell

MarcSabatella:
Not surprising there isn't a vtest for this particular case, it only affects "long" slurs.  But it very definitely changes the default position of slurs in many cases, that is the whole point.

Right now, this the default for the following cases:

image

The reason the first slur is below the staff is that normally one keeps slurs on the notehead side.  But there is code to enforce a "rule" that says to place "long" slurs above the staff regardless of stem direction, and the second one qualifies as long while the first doesn't.  it's quite arbitrary, really, and @oktophone had expressed the opinion this rule is not really worth implementing.

So this PR changes the second slur to be below the staff,  This will affect a *lot* of scores.  it's a good change, I think, but not one to consider for any 3.x release, were there to ever be such a thing.

PR #7154, opened 2020-12-23
[MU3] Style Dialog improvements

cbjeukendrup:
Resolves: https://musescore.org/en/node/314646 = https://trello.com/c/TyNjUPAH/79-format-style-page-and-format-style-score-dialogs-broken

- Fix #314646: Fix sizing of some pages in the Edit Style Dialog
  _And make sure it will also work on smaller screens by using a Scroll Area for the most critical pages._

- Style dialog: reduce width of some pages and add some more scroll areas
  _To make sure it also looks and works nice on smaller screens and to prevent the side bar from becoming wider and smaller for each page. 
  And the Tab order is cleaner than ever before!_

- Show Style Dialog always non-modally and bring it to front if already open (Follow up for PR #6845)
  _In Format > Style…, the dialog was shown non-modally, but when you opened it from a context menu or by double-clicking the header or footer, it was shown modally. Now, it will always be shown non-modally.
  When the dialog was already open but in the background, and you did something to open it again, nothing would happen. Now, the dialog will be brought to front in that situation._

- Fix warning about icon file not being found
  _At some places in the Style dialog, the name for the reset icon was specified without the `:/` prefix, like `data/icons/edit-reset.svg` instead of `:/data/icons/edit-reset.svg`. That produced a runtime warning (but remarkably, the icon did show up correctly)._

Here is a GIF showing all the pages of the style dialog:
Style Dialog

And here is a capture of the dialog on my 21,5" screen (4096 × 2304), to get a better idea of the size:
Style Dialog on 21,5 inch screen

If necessary, the user can resize the dialog even further down. 

### DIFF
--- a/mscore/editstyle.cpp
+++ b/mscore/editstyle.cpp
@@ -780,120 +780,173 @@ EditStyle::EditStyle(Score* s, QWidget* parent)
       MuseScore::restoreGeometry(this);
       }
 
+//---------------------------------------------------------
+//   adjustPagesStackSize
+//---------------------------------------------------------
+
 void EditStyle::adjustPagesStackSize(int currentPageIndex)
-{
-    QSize preferredSize = pageStack->widget(currentPageIndex)->sizeHint();
-    pageStack->setMinimumSize(preferredSize);
+      {
+      QSize preferredSize = pageStack->widget(currentPageIndex)->sizeHint();
+      pageStack->setMinimumSize(preferredSize);
 
-    connect(pageStack, &QStackedWidget::currentChanged, [this](int currentIndex) {
-        QWidget* currentPage = pageStack->widget(currentIndex);
-        if (!currentPage) {
-            return;
-        }
+      connect(pageStack, &QStackedWidget::currentChanged, [this](int currentIndex) {
+            QWidget* currentPage = pageStack->widget(currentIndex);
+            if (!currentPage)
+                  return;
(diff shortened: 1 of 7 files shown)


Jojo-Schmitz:
I hope you left the tab order intact (or fixed it again)?

cbjeukendrup:
> I hope you left the tab order intact (or fixed it again)?
> Edit: I see you mention it in the initial PR message.

Yep, don't worry!

mscore/editstyle.cpp
```diff
@@ -780,24 +780,57 @@ EditStyle::EditStyle(Score* s, QWidget* parent)
...
+      
+      if (_styleDlg->isVisible()) {
+            _styleDlg->raise();
+            _styleDlg->activateWindow();
+            }
+      else
+            _styleDlg->show();
+      }
+
+void MuseScore::showStyleDialogHeaderFooter()
```
IsaacWeiss, line 801: Since this function is identical to the one above except for one line, why not just check if `e` is a header or footer in `showStyleDialog()`?

cbjeukendrup: Good point. I felt indeed a little bad about this, but the problem was that there is no separate ElementType for Header/Footer. Your comment encouraged me to do a little more research, and I found a solution that is pretty okay. So thank you! 😊

PR #7155, opened 2020-12-24
[MU3] Fix #314696: Ties end at wrong note

mattmcclinch:
Resolves: https://musescore.org/en/node/314696.

When reading the properties of a tied note from a 2.x score, check to see if the specified pitch matches the pitch of the note at the beginning of the tie. If the pitches do not match, then print a debug message to the console, and set the pitch of the new note to match the pitch of the starting note.

### DIFF
--- a/libmscore/read206.cpp
+++ b/libmscore/read206.cpp
@@ -1336,8 +1336,18 @@ bool readNoteProperties206(Note* note, XmlReader& e)
       {
       const QStringRef& tag(e.name());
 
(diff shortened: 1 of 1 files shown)


mattmcclinch:
This worked to fix the original example, but it does not fix the latest example posted to the issue thread. In fact, it makes it worse. I may take another look at what might be done to solve this issue of importing ties from certain 2.x scores, but I think the first step is to revert this change.

PR #7156, opened 2020-12-24
[MU4] Fix #314696: Ties end at wrong note

mattmcclinch:
Resolves: https://musescore.org/en/node/314696.

When reading the properties of a tied note from a 2.x score, check to see if the specified pitch matches the pitch of the note at the beginning of the tie. If the pitches do not match, then print a debug message to the console, and set the pitch of the new note to match the pitch of the starting note.


mattmcclinch:
This was not the right fix, as evidenced by the latest example posted to the issue thread.

PR #7160, opened 2020-12-25
[MU4] Notation scrollbar

Eism:
_(sorry for the bad quality of the gif, in fact, nothing lags)_

Peek 2020-12-25 13-23

### DIFF
--- a/src/notation/internal/notationconfiguration.cpp
+++ b/src/notation/internal/notationconfiguration.cpp
@@ -65,15 +65,20 @@ void NotationConfiguration::init()
 {
     settings()->setDefaultValue(ANCHORLINE_COLOR, Val(QColor("#C31989")));
 
-    settings()->setDefaultValue(BACKGROUND_COLOR, Val(QColor("#142433")));
+    settings()->setDefaultValue(BACKGROUND_COLOR, Val(theme()->backgroundSecondaryColor()));
     settings()->valueChanged(BACKGROUND_COLOR).onReceive(nullptr, [this](const Val& val) {
         LOGD() << "BACKGROUND_COLOR changed: " << val.toString();
         m_backgroundColorChanged.send(val.toQColor());
     });
 
-    settings()->setDefaultValue(FOREGROUND_COLOR, Val(QColor("#f9f9f9")));
+    settings()->setDefaultValue(FOREGROUND_COLOR, Val(theme()->backgroundPrimaryColor()));
     settings()->valueChanged(FOREGROUND_COLOR).onReceive(nullptr, [this](const Val& val) {
         LOGD() << "FOREGROUND_COLOR changed: " << val.toString();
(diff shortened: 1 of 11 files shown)


src/notation/internal/notationconfiguration.cpp
```diff
@@ -130,17 +141,19 @@ bool NotationConfiguration::foregroundUseColor() const
...
-
 QColor NotationConfiguration::foregroundColor() const
 {
     if (settings()->value(FOREGROUND_USE_USER_COLOR).toBool()) {
         return settings()->value(FOREGROUND_COLOR).toQColor();
     }
-    return defaultForegroundColor();
+
+    QColor color = settings()->value(FOREGROUND_COLOR).toQColor();
+    QColor defaultColor = settings()->defaultValue(FOREGROUND_COLOR).toQColor();
```
RomanPudashkin, line 151: This code was duplicated. Please, fix it

src/notation/view/notationpaintview.cpp
```diff
@@ -78,6 +90,30 @@ NotationPaintView::~NotationPaintView()
     delete m_noteInputCursor;
 }
 
+void NotationPaintView::requestScrollHorizontal(qreal position)
```
RomanPudashkin, line 93: Please, rename this method to scrollHorizontal (along with the method placed bellow)

src/notation/view/notationpaintview.cpp
```diff
@@ -271,6 +307,90 @@ QRect NotationPaintView::viewport() const
...
+qreal NotationPaintView::horizontalScrollableAreaSize() const
+{
+    if (viewport().left() < notationContentRect().left()
+        && viewport().right() > notationContentRect().right()) {
+        return 0;
+    }
+
+    qreal scrollableWidht = horizontalScrollableSize();
+
+    return viewport().width() / scrollableWidht;
```
RomanPudashkin, line 345: Typo:
scrollableWi**dht**

RomanPudashkin: Can we get division on zero here?

src/notation/view/notationpaintview.cpp
```diff
@@ -26,6 +26,11 @@
 using namespace mu::notation;
 using namespace mu::framework;
 
+static constexpr qreal MIN_SCROLL_SIZE = 0.2;
+static constexpr qreal MAX_SCROLL_SIZE = 1.0;
+
+static constexpr qreal CANVAS_SIDE_MARGIN = 8000;
```
RomanPudashkin, line 32: Please, use some proportional value instead of this magic number (that depends on the current value of width/height). You also should use guiScale() when you will calculate this value

RomanPudashkin: I will fix this issue later

RomanPudashkin (approved):

PR #7161, opened 2020-12-25
[MU3] Cleanup translatable texts

Jojo-Schmitz:
* Fix time signature names translation in palettes
* Fix compiler warnings
* Fix translatable strings
   * 'Untranslate' some button texts
   * British vs. American English
   * Reset texts mismatch with settings they refer to
   * Make text consistent with neighbours
   * Some fixes that where pending since 3.5 string freeze
   * Add missing strings from order.xml and make sure "Custom" gets translated too
   * Use same string as the corresponding UI file does
   * Remove trailing space
   * Remove superfluous colon


cbjeukendrup:
> Will most probably cause merge conflicts with #7154

What we can do if we want to avoid conflicts, is that I add the fixes you did in `editstyle.ui` to my PR (of course with credits to you 😉). Otherwise, I guess that it would be favorable if my PR would be merged first.

Jojo-Schmitz:
Yes, I agree. Credit is not really needed though ;-)

cbjeukendrup:
@Jojo-Schmitz Do you think you are ready with `editstyle.ui`? Then I will add the changes to my PR.

And maybe you also want to do for 3.x, what I did for master here: https://github.com/musescore/MuseScore/pull/7169/commits/c93dbd5095fc4635c9abeda2fa2a0473ee353873
Seems quite translation-related to me.

Jojo-Schmitz:
I left my comment there, that %n is perfectly correct there.
I hope to be do with edidstyle.ui

cbjeukendrup:
Unfortunately I don't see your comment (yet)... could you maybe post a link to it?

If you find something else in editstyle.ui, please let me know and I will add it!

Jojo-Schmitz:
https://github.com/musescore/MuseScore/pull/7169/commits/c93dbd5095fc4635c9abeda2fa2a0473ee353873#r549134457

importexport/musicxml/exportxml.cpp
```diff
@@ -4977,8 +4977,9 @@ static void annotations(ExportMusicXml* exp, int strack, int etrack, int track,
                   wtrack = findTrackForAnnotations(e->track(), seg);
 
             if (track == wtrack) {
-                  if (commonAnnotations(exp, e, sstaff))
+                  if (commonAnnotations(exp, e, sstaff)) {
                         ;  // already handled
```
vpereverzev, line 4981: why do we need this 'IF' at all?

Jojo-Schmitz: Good question, I guess mainly for documenting?

Jojo-Schmitz: 427a5396

Jojo-Schmitz: @thibault ?

PR #7164, opened 2020-12-26
[MU4] Fixed compiler warning

Eism:
#7142
@Jojo-Schmitz


Jojo-Schmitz:
Interesting. Thanks for taking care of it

PR #7165, opened 2020-12-26
[MU4] Updated used XCode on CI

igorkorsukov:
Updated used XCode on CI

### DIFF
--- a/build/cmake/SetupBuildEnvironment.cmake
+++ b/build/cmake/SetupBuildEnvironment.cmake
@@ -106,45 +106,7 @@ endif()
 # APPLE specific
 if (OS_IS_MAC)
       set(CMAKE_OSX_ARCHITECTURES x86_64)
-      set(MACOSX_DEPLOYMENT_TARGET 10.10)
-      set(CMAKE_OSX_DEPLOYMENT_TARGET 10.10)
+      set(MACOSX_DEPLOYMENT_TARGET 10.14)
+      set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14)
 endif(OS_IS_MAC)
 
-
-#   string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE)
-#   # Windows: Add -mconsole to LINK_FLAGS to get a console window for debug output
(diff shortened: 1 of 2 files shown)


build/cmake/SetupBuildEnvironment.cmake
```diff
@@ -106,45 +106,7 @@ endif()
 # APPLE specific
 if (OS_IS_MAC)
       set(CMAKE_OSX_ARCHITECTURES x86_64)
-      set(MACOSX_DEPLOYMENT_TARGET 10.10)
-      set(CMAKE_OSX_DEPLOYMENT_TARGET 10.10)
+      set(MACOSX_DEPLOYMENT_TARGET 10.14)
+      set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14)
```
cbjeukendrup, line 110: What I have often wondered about is, is the deployment target just the minimum required system version, or not necessarily? If it is, shouldn't it be 10.13 in this case? According to the docs, Qt 5.15 supports macOS 10.13-10.15 and Xcode 11 or 12 with the macOS 10.15 SDK (https://doc.qt.io/qt-5/supported-platforms.html#macos), but I'm not sure if we need to use exactly those settings.

igorkorsukov: > What I have often wondered about is, is the deployment target just the minimum required system version, or not necessarily? If it is, shouldn't it be 10.13 in this case? According to the docs, Qt 5.15 supports macOS 10.13-10.15 and Xcode 11 or 12 with the macOS 10.15 SDK (https://doc.qt.io/qt-5/supported-platforms.html#macos), but I'm not sure if we need to use exactly those settings.

In the future, we plan to switch to Qt 6.0
For Qt 6.0 minimally supports 10.14
I think it's better to immediately make a limitation for MuseScore 4.0 minimally supports 10.14, instead of, for example, releasing MuseScore 4.0 with a minimum supported 10.13, and then releasing MuseScore 4.1 with a minimum supported 10.14 - this will introduce a lot of confusion and difficulties.

cbjeukendrup: Yes, that's a good idea. Exiting to hear about plans to move to Qt 6!

PR #7166, opened 2020-12-26
[MU4] Converter implementation

igorkorsukov:
Converter implementation


src/appshell/commandlinecontroller.cpp
```diff
@@ -40,18 +41,35 @@ void CommandLineController::parse(const QStringList& args)
...
+            LOGE() << "Option: -D not recognized DPI value: " << m_parser.value("D");
         }
     }
 
     // Converter mode
+    if (m_parser.isSet("r")) {
+        std::optional<float> val = floatValue("r");
+        if (val) {
+            importexportConfiguration()->setExportPngDpiResolution(val);
+            LOGE() << "Option: -r not recognized DPI value: " << m_parser.value("r");
```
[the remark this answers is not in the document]
igorkorsukov: > Isn't here a line missing?
> 
> ```
>         } else {
> ```

If we were unable to recognize the value, then we ignore the option. I don't know what could be else.

src/appshell/commandlinecontroller.cpp
```diff
@@ -40,18 +41,35 @@ void CommandLineController::parse(const QStringList& args)
...
 {
-    // Common
-    if (m_parser.isSet("D")) {
+    auto floatValue = [this](const QString& name) -> std::optional<float> {
         bool ok;
-        float val = m_parser.value("D").toFloat(&ok);
+        float val = m_parser.value(name).toFloat(&ok);
         if (ok) {
+            return val;
+        } else {
```
RomanPudashkin, line 49: Better return result without this else

src/appshell/commandlinecontroller.cpp
```diff
@@ -40,18 +41,35 @@ void CommandLineController::parse(const QStringList& args)
...
+            return std::nullopt;
+        }
+    };
+
+    // Common
+    if (m_parser.isSet("D")) {
```
RomanPudashkin, line 55: Maybe better to make named constants for all options? Like this:

static const QString CHANGE_DPI_OPTION("D")

src/appshell/commandlinecontroller.h
```diff
@@ -25,12 +25,14 @@
...
 #include "global/iapplication.h"
 #include "ui/iuiconfiguration.h"
+#include "importexport/iimportexportconfiguration.h"
 
 namespace mu::appshell {
 class CommandLineController
 {
     INJECT(appshell, framework::IApplication, application)
     INJECT(appshell, framework::IUiConfiguration, uiConfiguration)
+    INJECT(appshell, importexport::IImportexportConfiguration, importexportConfiguration)
```
RomanPudashkin, line 35: Can we make appshellconfiguration and use it instead of uiConfiguration and importexportConfiguration?

igorkorsukov: Interesting idea.
But I think this is an overcomplication, I do not see any benefit, and there is a lack - now we can immediately see in the application of commands what they affect (and this may not only be a change in configurations), we will not see it with appshellconfiguration, we will need to additionally go to appshellconfiguration ...
The CommandLineController itself is intended to be a controller - to determine what needs to be changed

src/importexport/internal/flacwriter.cpp
```diff
@@ -23,15 +23,14 @@
 
 using namespace mu::importexport;
 using namespace mu::framework;
-using namespace Ms;
 
-mu::Ret FlacWriter::write(const Score& score, IODevice& destinationDevice, const Options& options)
+mu::Ret FlacWriter::write(const notation::INotation* notation, IODevice& destinationDevice, const Options& options)
```
RomanPudashkin, line 27: I think, better to use INotationPtr instead of raw pointer

igorkorsukov: > I think, better to use INotationPtr instead of raw pointer

Yes, but it's inconvenient here
https://github.com/musescore/MuseScore/pull/7166/files#diff-bc2452ffd86e64620f2974b5e7fa0a8c1d3bdf00b5adee80b1dec460877c92b0R539

src/importexport/internal/svgwriter.cpp
```diff
@@ -36,13 +36,22 @@ using namespace mu::importexport;
 using namespace mu::framework;
 using namespace Ms;
```
RomanPudashkin, line 37: Please, remove this using

src/importexport/internal/flacwriter.cpp
igorkorsukov: I will fix it

src/appshell/commandlinecontroller.cpp
igorkorsukov: I think it will not be convenient.
Looking for where the option is processed ...
If you make constants, you must first find the name of the constant, and then the place where it is processed.

src/appshell/commandlinecontroller.cpp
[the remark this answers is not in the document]
igorkorsukov: opps, I get it, you're right !! :)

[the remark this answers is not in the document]
igorkorsukov: fixed

src/appshell/commandlinecontroller.cpp
igorkorsukov: changed

src/importexport/internal/flacwriter.cpp
igorkorsukov: fixed

src/importexport/internal/svgwriter.cpp
igorkorsukov: fixed

RomanPudashkin (approved):

PR #7169, opened 2020-12-27
[MU4] Fix compiler/analyzer warnings found using Xcode

cbjeukendrup:
See the commit messages for details. 

### DIFF
--- a/src/libmscore/fraction.h
+++ b/src/libmscore/fraction.h
@@ -194,7 +194,7 @@ public:
     Fraction& operator*=(const Fraction& val)
     {
         _numerator *= val._numerator;
-        _denominator  *= val._denominator;
+        _denominator *= val._denominator;
         if (val._denominator != 1) {
             reduce();                            // We should be free to fully reduce here
         }
--- a/src/libmscore/read206.cpp
+++ b/src/libmscore/read206.cpp
@@ -1501,8 +1501,8 @@ bool readNoteProperties206(Note* note, XmlReader& e)
             sp->setTick(e.tick());
             note->addSpannerFor(sp);
             sp->setParent(note);
+            adjustPlacement(sp);
         }
-        adjustPlacement(sp);
     } else if (tag == "offset") {
         note->Element::readProperties(e);
     } else if (note->Element::readProperties(e)) {
--- a/src/libmscore/segment.cpp
+++ b/src/libmscore/segment.cpp
@@ -1715,6 +1715,9 @@ Element* Segment::nextElement(int activeStaff)
     if (!e && !score()->selection().elements().isEmpty()) {
         e = score()->selection().elements().first();
     }
+    if (!e) {
+        return nullptr;
+    }
     switch (e->type()) {
     case ElementType::DYNAMIC:
     case ElementType::HARMONY:
@@ -1852,6 +1855,9 @@ Element* Segment::prevElement(int activeStaff)
     Element* e = score()->selection().element();
     if (!e && !score()->selection().elements().isEmpty()) {
         e = score()->selection().elements().last();
+    }
+    if (!e) {
+        return nullptr;
     }
     switch (e->type()) {
     case ElementType::DYNAMIC:
--- a/src/palette/qml/MuseScore/Palette/PaletteTree.qml
+++ b/src/palette/qml/MuseScore/Palette/PaletteTree.qml
@@ -29,7 +29,7 @@ import "utils.js" as Utils
 
 ListView {
     id: paletteTree
-    Accessible.name: qsTrc("palette", "Palettes Tree, contains %n palette(s)").arg(count)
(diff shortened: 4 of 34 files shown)


cbjeukendrup:
The code style check fails... but I'm really convinced that the change is correct. Or am I missing something?

src/palette/qml/MuseScore/Palette/PaletteTree.qml
```diff
@@ -29,7 +29,7 @@ import "utils.js" as Utils
 
 ListView {
     id: paletteTree
-    Accessible.name: qsTrc("palette", "Palettes Tree, contains %n palette(s)").arg(count)
+    Accessible.name: qsTrc("palette", "Palettes Tree, contains %1 palette(s)").arg(count)
```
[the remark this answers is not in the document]
cbjeukendrup: Ah, that is what it means...!
I thought it was a mistake because there was a warning and I could not find any documentation about `%n`. Now I have finally found [some documentation](https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals), so I will have a closer look at it.

src/libmscore/read206.cpp
```diff
@@ -1501,8 +1501,8 @@ bool readNoteProperties206(Note* note, XmlReader& e)
             sp->setTick(e.tick());
             note->addSpannerFor(sp);
             sp->setParent(note);
+            adjustPlacement(sp);
         }
-        adjustPlacement(sp);
```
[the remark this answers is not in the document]
cbjeukendrup: Yes, that seems to be the case. Probably with many other things too. I am planning to do also a 3.x version, probably next week.

src/palette/qml/MuseScore/Palette/PaletteTree.qml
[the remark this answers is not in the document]
cbjeukendrup: I found out what was the problem! Thanks Jojo, for making me aware of this!
In 3.x, this problem does not occur, so no need to change anything there.

src/libmscore/fraction.h
```diff
@@ -245,15 +245,15 @@ class Fraction
         if (ticks == -1) {
             return Fraction(-1,1);        // HACK
         }
-        return Fraction(ticks, MScore::division* 4).reduced();
+        return Fraction(ticks, MScore::division * 4).reduced();
```
Jojo-Schmitz, line 248: Believe it or not: the current code style wants it that way...

src/libmscore/fraction.h
```diff
@@ -245,15 +245,15 @@ class Fraction
...
     }
 
     //---------------------------------------------------------
     //   eps
     ///   A very small fraction, corresponds to 1 MIDI tick
     //---------------------------------------------------------
 
-    static Fraction eps() { return Fraction(1, MScore::division* 4); }
+    static Fraction eps() { return Fraction(1, MScore::division * 4); }
```
Jojo-Schmitz, line 256: same here

igorkorsukov:
> The code style check fails... but I'm really convinced that the change is correct. Or am I missing something?

Your changes are correct, the code style checker is not working correctly.
So that we can push your changes, you need to change the config of the code style checker so that your code passes the check.
https://github.com/musescore/MuseScore/tree/master/tools/codestyle

src/libmscore/segment.cpp
```diff
@@ -986,7 +986,7 @@ bool Segment::setProperty(Pid propertyId, const QVariant& v)
 qreal Segment::widthInStaff(int staffIdx, SegmentType t) const
 {
     const qreal segX = x();
-    qreal nextSegX = segX;
+    qreal nextSegX;
```
igorkorsukov, line 989: Would you tell, why this change was made?

cbjeukendrup: In the lines here below, `nextSegX` will be set to another value anyway, so Xcode gave a warning that the initial value is never read. Therefore, I removed that initial value.

RomanPudashkin: Please, return initial values to all variables with primitive types. Variables with undefined values are very dangerous. After some time implementation of this function may change and this initial value can save from hard reproduced bug

cbjeukendrup: I see, I returned it!

cbjeukendrup:
@igorkorsukov I'm not sure what I should change in the Uncrustify config. The config seems fine to me, but uncrustify thinks that that star is a pointer star instead of an arithmetic operator. 
If the PR would be merged, would then on all subsequent PRs the code style check fail? Or will uncrustify take it for granted then?
Otherwise, I think this small cosmetic issue is not worth the trouble.

Jojo-Schmitz:
Just adjust your code and let @igorkorsukov amd his team deal with uncrustify

cbjeukendrup:
> Just adjust your code and let @igorkorsukov amd his team deal with uncrustify

That seems indeed the best for now. I reverted the 'controversial' changes, so we can deal with it at some other moment.

PR #7171, opened 2020-12-28
fix #137091: [MusicXML Import] should not create visual tempo element…

lvinken:
… for <sound tempo="num"/> element

Resolves: https://musescore.org/en/node/137091

To comply with the MusicXML spec, import tempo changes without explicit text to display as invisible. Also prevent duplicate tempo changes in case these are present in multiple parts. Requested by the Music21 project.

Note that this PR uses the smallest set of changes to achieve the goal. Some duplicated code left in place to minimise risk.

### DIFF
--- a/importexport/musicxml/importmxmlpass2.cpp
+++ b/importexport/musicxml/importmxmlpass2.cpp
@@ -52,6 +52,7 @@
 #include "libmscore/staff.h"
 #include "libmscore/stafftext.h"
 #include "libmscore/sym.h"
+#include "libmscore/tempo.h"
 #include "libmscore/tempotext.h"
 #include "libmscore/tie.h"
 #include "libmscore/timesig.h"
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
Amazing, thank you! cc/ @mscuthbert @GeorgeTouros

matangover:
@lvinken I think the detection of duplicate tempo changes can be a bit more specific to actually only detect duplicates. Right now it removes tempo changes even if they are not duplicate, i.e. a score with simultaneous tempo changes does not properly survive a MusicXML export-import roundtrip.

Example -- this score:
image
After export to xml and import from xml, will look like this:
image
... and its playing tempo will be different from the original tempo.

In general maybe duplicate tempo elimination should be avoided - if the MusicXML included duplicate tempos (for whatever reason..?) then MuseScore should import them. What do you think?

lvinken:
@matangover obviously whenever possible an exact representation of the score in MusicXML should be generated by MuseScore, on import the MusicXML data should be converted as accurately as possible. In case you have found a bug I suggest you submit an issue including the relevant data. I could not quickly reproduce the symptom.

Note: a closed pull request would not be the best place to discuss new issues.

PR #7176, opened 2020-12-28
Improvements to line dragging behavior

MarcSabatella:
This PR contains a number of separate but related improvements to line dragging functionality.  It addresses some long-standing bugs (the behavior where aligned pedal lines can jump wildly on drag has been around since 3.1, and the problem where reset does not restore the offset to styled status has been around since 3.0), as well as some more recent regressions (the inability to manually adjust end handles without being forced into an anchor change was new with 3.5).

The commits here are mostly separate and could standalone for cherry-picking if need be.  However, the fix to the wild jumps of dragged pedal lines does rely in part on the fix to the styling of the offset property.  And they are all related to dragging behavior and are bested tested together.

Here are the direct links to the issues fixed (also found within the individual commit messages):

1) https://musescore.org/en/node/314854
2) https://trello.com/c/QRquMj6w/84-pedal-markings-seem-broken-when-trying-to-move-them-vertically
3) https://musescore.org/en/node/313721
4) https://musescore.org/en/node/309368

### DIFF
--- a/libmscore/element.cpp
+++ b/libmscore/element.cpp
@@ -2099,9 +2099,9 @@ QVector<QLineF> Element::genericDragAnchorLines() const
       if (parent()->isSegment()) {
             System* system = toSegment(parent())->measure()->system();
             const int stIdx = staffIdx();
-            yp = system->staffCanvasYpage(stIdx);
+            yp = system ? system->staffCanvasYpage(stIdx) : 0.0;
             if (placement() == Placement::BELOW)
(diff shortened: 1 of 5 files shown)


MarcSabatella:
I believe I understand the vtest failures, will update.

MarcSabatella:
The remaining vtest failure is expected.  The fix for the wildly jumping pedal lines was to disable automatic alignment for manually adjusted lines, and these are manually adjusted.

PR #7179, opened 2020-12-29
[MU3] Fix #314899: Identical expression

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/314899

Same fix needed in master, see #7180, or #7627, both closed by now.
And see #8762, which brought this up again.

### DIFF
--- a/importexport/midiimport/importmidi_simplify.cpp
+++ b/importexport/midiimport/importmidi_simplify.cpp
@@ -32,23 +32,6 @@ bool hasComplexBeamedDurations(const QList<std::pair<ReducedFraction, TDuration>
       return false;
       }
 
-
-#ifdef QT_DEBUG
-
-bool areDurationsEqual(
-            const QList<std::pair<ReducedFraction, TDuration> > &durations,
-            const ReducedFraction &desiredLen)
-      {
-      ReducedFraction sum(0, 1);
-      for (const auto &d: durations)
-            sum += ReducedFraction(d.second.fraction()) / d.first;
-
-      return desiredLen == desiredLen;
-      }
-
-#endif
-
-
 void lengthenNote(
             MidiNote &note,
             int voice,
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
The mtest fail (https://github.com/musescore/MuseScore/pull/7179/checks?check_run_id=1622482883#step:5:762) seems to be the result of this PR's change, and reveal a real bug for a change.

AntonioBL:
The crash is caused by an assert at line 104 of importexport/midiimport/importmidi_simplify.cpp
` QFATAL : TestImportMidi::voiceSeparationTuplet() unknown: ASSERT failure in Simplify::lengthenNote: "Too short note durations remaining", file /media/antonio/Mint0/MuseScore/importexport/midiimport/importmidi_simplify.cpp, line 104`

AntonioBL:
And I think that in master importmidi mtests are not being tested at the moment.

Jojo-Schmitz:
It had to be one of those 4 Q_ASSERT_X, but why?

AntonioBL:
By using a break at that X_ASSERT (line 104 in 3.x and line 102 in master), the noteDurations list is:
first    second
2/2     1/16 dots 0
2/2     1/128 dots 0
2/2     1/512 dots 0
which gives sum = 37/512, while the desiredLen (offset - DurationStart) is 7/96.

Jojo-Schmitz:
Close but not the same

Jojo-Schmitz:
Guess a case of "never check for error conditions you can't handle"

Jojo-Schmitz:
Need to introduce an epsilon for this apparent floating point comparison?

Jojo-Schmitz:
The tuplet_3_5_7_tuplets test produces a corrupt MuseScore file (in 3.5.2)

AntonioBL:
tuplet_3_5_7_tuplets.mid is imported exactly as the reference if "Split staff" is unchecked. And indeed in the mtest there is `data.trackOpers.doStaffSplit.setDefaultValue(false, false);`

Jojo-Schmitz:
For me that .mid file crashes with an assertion failure on load (in a debug build). Only when importing in a non-debug build I couls save it as mscz (but without using any of those options) and got that corrupt score.
Anyway; I think this crash might be a red herring, as it crashes in an entirely different place, one that my change can't possibly have caused.

RomanPudashkin (approved):

Jojo-Schmitz:
Want me to port it over to master?

Jojo-Schmitz:
That "Thumbs up" doesn't notify...
Anyway, see #8849

PR #7180, opened 2020-12-29
[MU4] Fix #314899: Identical expression

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/314899

This #7179 but for master


Jojo-Schmitz:
As this here needs a rebase and is handel also in #7627, I'm closing it

PR #7184, opened 2020-12-29
Fix #269740: Retranslate when current language is updated

cbjeukendrup:
Previously, when you updated the current language using the Resource Manager, you would only be able to see the updated strings after restarting MuseScore. This was, because the condition to determine whether the *current* language had been updated was wrong, so the `retranslate` function would not be called. Now, the condition is correct, and the function is correctly called.
I also renamed some variables to make the code clearer to understand, and I fixed a problem with the variable `localeName` not being updated on a language switch, which caused that the wrong language appeared at the top of the list in the Resource Manager, and might have caused other problems too.

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -224,7 +224,6 @@ static bool scoresOnCommandline { false };
 
 static QList<QTranslator*> translatorList;
 
-QString localeName;
 bool useFactorySettings = false;
 bool deletePreferences = false;
 QString styleName;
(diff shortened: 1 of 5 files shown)


Jojo-Schmitz (approved):

Jojo-Schmitz:
Resolves https://musescore.org/en/node/269740

cbjeukendrup:
> Resolves https://musescore.org/en/node/269740

Ah, thanks, I will include that in the commit message!

PR #7192, opened 2020-12-30
fix measure number colliding with brackets

MarcSabatella:
When we changed measure numbers to not scale with staff size,
this caused the offset to not scalke either,
and this in turn causes the units to be interpreted as mm not sp.
The old default value of -2 is good for sp, but too small in mm.

A bandaid fix would be to increase the default value,
but the approach taken here is I think better.
The idea is that converting the offset units to mm
only makes sense for elements like titles that are placed in frames,
as opposed to elements lime measure numbers that are on staff.

So I have introduced a new function offsetIsSpatiumDependent(),
and I use it instead of sizeIsSpatiumDependent() in the relevant places.
The offset is considered spatium dependent if the size is,
*or* if the object is not attached to a frame.
There might be a better way to make the determination,
but this gets the job done.

### DIFF
--- a/libmscore/element.cpp
+++ b/libmscore/element.cpp
@@ -110,7 +110,7 @@ namespace Ms {
 
 void Element::spatiumChanged(qreal oldValue, qreal newValue)
       {
-      if (sizeIsSpatiumDependent())
+      if (offsetIsSpatiumDependent())
             _offset *= (newValue / oldValue);
       }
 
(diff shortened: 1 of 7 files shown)


MarcSabatella:
The vtest diffs are all tp be expected - measure numbers with this change are back to their 3.5.2 position, higher above the staff.

Jojo-Schmitz:
Some though are quite a lot higher than others?

MarcSabatella:
They should be the same as 3.5.2, are there some where you see a difference?

Jojo-Schmitz:
No, not looked into 3.5 at all

MarcSabatella:
OK.  Well, I don't see anything that looks different to me.  The one where the measure numbers seem most strangely high is measure-number-2, but that's because it specifically set a larger value for the offset.

Jojo-Schmitz:
OK. Guess I meant that ;-)

PR #7193, opened 2020-12-30
Fix #287858 and Retranslate several dialogs on language switch

cbjeukendrup:
Resolves: https://musescore.org/en/node/287858

- Retranslate Style Dialog on language switch
- Retranslate Instruments Dialog on language switch
- Fix #287858: Make the Edit Toolbars Dialog fully (re)translatable
- Retranslate Text Tools on language switch
- Retranslate Timeline on language switch
- Retranslate Mixer on language switch

---

### DIFF
--- a/mscore/abstractdialog.cpp
+++ b/mscore/abstractdialog.cpp
@@ -26,11 +26,6 @@ AbstractDialog::AbstractDialog(QWidget * parent, Qt::WindowFlags f)
       {
       }
 
-AbstractDialog::~AbstractDialog()
-      {
-
-      }
-
 void AbstractDialog::changeEvent(QEvent *event)
       {
       QDialog::changeEvent(event);
--- a/mscore/toolbarEditor.cpp
+++ b/mscore/toolbarEditor.cpp
@@ -41,7 +41,7 @@ void MuseScore::showToolbarEditor()
 //---------------------------------------------------------
 
 ToolbarEditor::ToolbarEditor(QWidget* parent)
-   : QDialog(parent)
+   : AbstractDialog(parent)
       {
       setObjectName("ToolbarEditor");
       setupUi(this);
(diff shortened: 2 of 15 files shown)


mscore/toolbarEditor.cpp
```diff
@@ -18,9 +18,9 @@
 namespace Ms {
 
 static const char* toolbars[] = {
-      QT_TRANSLATE_NOOP("toolbar", "Note Input"),
-      QT_TRANSLATE_NOOP("toolbar", "File Operations"),
-      QT_TRANSLATE_NOOP("toolbar", "Playback Controls")
+      "Note Input",
+      "File Operations",
+      "Playback Controls"
```
Jojo-Schmitz, line 23: Isn't this rather untranslating these strings?

cbjeukendrup: Yes, I'm untranslating them here, because they are already translated in line 50. So they were actually translated twice.

Jojo-Schmitz: They are not translated **here**, just extracet by `lupdate`. The translation happens at `qApp->translate("toolbar",...)`

mscore/abstractdialog.cpp
```diff
@@ -28,7 +28,6 @@ AbstractDialog::AbstractDialog(QWidget * parent, Qt::WindowFlags f)
 
 AbstractDialog::~AbstractDialog()
       {
-
```
Jojo-Schmitz, line 31: More a general question: is an empty destructor needed at all?

mscore/toolbarEditor.cpp
Jojo-Schmitz: `QT_TRANSLATE_NOOP()` is just for `lupdate` and on compile (or rather on pre-process) just evaluates to the 2nd arg.

Jojo-Schmitz: `qApp->translated(context, string)` then pick up `string` in ist translated form.

cbjeukendrup: Ah, I see, I will have another look at it!

mscore/abstractdialog.cpp
cbjeukendrup: I would not know why... so tried removing it. Seems that doesn't cause problems.

Jojo-Schmitz (approved):

Jojo-Schmitz:
c831488 just got reverted

vpereverzev:
See https://trello.com/c/YzDXvdJ7

PR #7194, opened 2020-12-31
Reduce score orders

njvdberg:
Resolves: https://trello.com/c/0AzAkQpX/52-check-over-the-new-orderings-that-have-been-added
Resolves: https://trello.com/c/PlM5cRqi/91-title-should-be-orchestral-not-orchestra

This PR contains 2, related, commits:
1)  If a loaded score or template contains a score order *aandA* this score order doesn't exists yet. added it to the list of score orders as a not customised score order. This enables having score orders in templates without having defined in <code>orders.xml</code>
2) Adapted all templates to new engraving standard and removed <code>Small Concert Band</code> and <code>Marching Band</code> from <code>orders.xml</code>. These score orders are now in the templates only. Also score order <code>Orchestra</code> is renamed to <code>Orchestral</code>.
 


Jojo-Schmitz:
You'd need to run that Python script in share/instruments to update (and commit) instrumentxml.h with the new/changes strings

njvdberg:
> You'd need to run that Python script in share/instruments to update (and commit) instrumentxml.h with the new/changes strings

Done. Thanks!

libmscore/read302.cpp
```diff
@@ -299,6 +299,17 @@ bool Score::read(XmlReader& e)
                         setScoreOrder(order);
                         }
                   }
+            else {
```
vpereverzev, line 302: I think we don't need this duplication here

You could just put following lines between 290-291:

`ScoreOrder* defined = scoreOrders.findByName(order->getName());
if (!defined)
      defined = scoreOrders.findById(order->getId());
....`

vpereverzev (changes requested):

Jojo-Schmitz:
needs a rebase too

libmscore/read302.cpp
njvdberg: This won't work because, although the branches of the two nested <code>if</code> statements are similar, the condition isn't.
The idea here is:
- 1 Is there already a score order defined in the score order list (as defined by <code>orders.xml</code>
- 2 If there is already a order with the same name defined, is it actually the **same** score (all instruments in the score are in the order as defined by the score order). If the instruments are in the same order, use the score order as defined in <code>orders.xml</code>, otherwise use the order as found in the score file as a customised version.
- 3 If there is no order with the same name found, check whether the same score is read already before (so, with same <code>ID</code>). If such order is found, use that score, otherwise use the the score order as defined in the score file.

This makes is possible to have templates with their own, specialised, score orders, without the necessity the have them defined in any <code>orders.xml</code> file.

In generally I agree and try to prevent these kind of duplications but in this case I didn't see how without making it more difficult to understand what is meant here. Nevertheless, some comments might a good idea.

njvdberg:
Rebase done.

PR #6461, opened 2020-08-17
Fix #281639: Better handling of Clefs, Barlines on ossias and cutaway

elerouxx:
Partially Resolves: *https://musescore.org/en/node/281639*
(Brackets out of scope of this PR)

**Fixes visibility of Bar Lines for Ossias or Cutaway scores.**
The EndBarLine of an invisible measure will automatically show if next measure is visible and in the same system.
Span works as expected, no manual insert of bar line needed.
(Barlines can always be manually set to invisible, if desired)

**Fixes showing courtesy clefs on invisible measures**
On a cutaway staff, If a courtesy clef is added to the end of an empty measure, a short staff "trailer" will be created to display the clef. This allows clefs to be displayed on the first visible measure or after an invisible section of a cutaway staff.
  (Clefs must be created manually by adding the clef to a  whole measure, as always).

Before
image

After
image

As always, feedback and review will be appreciated.

*Some additional small fixes:
- Dragging manually-added bar lines won't set the Global span anymore (which was probably the intention of the existing code)
- Measure numbers will show even if measure is invisible or cut away from the system.
- Spacers will always show, even on invisible measures, to avoid losing track of them on a cutaway score.*

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7196, opened 2020-12-31
Fix #314910: Cutaway courtesy clefs in 3.6 don't show before time sig…

elerouxx:
…natures

Resolves: *https://musescore.org/en/node/315014*

*Fixes cutaway courtesy clefs not showing when the EndBarLine segment is not the last segment in the measure. (This happens when there is a time signature or a key signature in next measure, generating timesig/keysig Courtesy segments)*

This is an improvement to functionality introduced in PR #6461.

- [NA] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -2879,7 +2879,7 @@ bool Measure::isEmpty(int staffIdx) const
 //---------------------------------------------------------
 //   isCutawayClef
 ///    Check for empty measure with only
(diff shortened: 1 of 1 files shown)


MarcSabatella:
I don't know that it is this change in particular, but somehow I do suspect the new code for handling courtesy clefs with respect to cutaway might be responsible for https://musescore.org/en/node/315861.  See my comment there.

PR #7197, opened 2020-12-31
badly formed -> malformed

riaanvn:
String correction; discussed in MuseScore Translation Telegram group


riaanvn:
Realised my mistake. Will re-submit against 3.x

PR #7200, opened 2021-01-01
fix #314642 [MusicXML] big 32nd tuplet exports wrong durations

AntonioBL:
Resolves: https://musescore.org/en/node/314642

The bug originated when commit https://github.com/musescore/MuseScore/commit/31c68c31ad807c11ea3acc196f8991008a43babf was merged, because of a rounding error for the short notes involved in the tuplet (it produced 59/1920 instead of 60/1920). This PR forces the duration type calculation to use the actual duration fraction of the note instead of its (integer) ticks. The original tick-based calculations is a residual of the time when the internal code of MuseScore used integer ticks (not actual fractions) for note durations.

### DIFF
--- a/importexport/musicxml/exportxml.cpp
+++ b/importexport/musicxml/exportxml.cpp
@@ -1530,9 +1530,12 @@ static void unpitch2xml(const Note* note, QString& s, int& octave)
 
 static QString tick2xml(const Fraction& ticks, int* dots)
       {
-      TDuration t;
-      t.setVal(ticks.ticks());
+      TDuration t(ticks);
       *dots = t.dots();
+      if (ticks == Fraction(0,1)) {
+            t.setType("measure");
+            *dots = 0;
+            }
       return t.name();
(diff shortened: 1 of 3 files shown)


lvinken:
Hi Antonio, thanks for looking into this. I tested your fix with quintuplets, septuplets and nonuplets consisting of notes from 1024ths up to quarters, all worked OK.

Note that function tick2xml() still uses ticks internally, but apparently this causes no problems. A structural solution would be to remove tick2xml() and use the same duration handling for notes as is currently used for rests. See "type" in ExportMusicXml::rest().

AntonioBL:
Thank you Leon for testing the PR.
I modified tick2xml(), which is used other two times in the file, so that it converts a fraction to the corresponding duration without going through its relative integer ticks. It seems to work ok and does not break any of the tests.

lvinken:
Commit  fc37c01 also tested OK by me.

PR #5660, opened 2020-01-31
fix #281503 : enable compilation with make on macOS

AntonioBL:
Resolves: https://musescore.org/en/node/281503
This PR enables compilation of MuseScore under MacOS without installing the full XCode IDE, but only the command line tools.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ N/A ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7206, opened 2021-01-06
[MU4] Added macOS module and implement two dark mode improvements

cbjeukendrup:
In this PR, I add a 'macOS' module for Mac-specific things. This way, other systems will be affected by it as little as possible. 
@igorkorsukov I would be interested to hear your opinion about this approach!

This makes it possible to:
- observe system notifications;
- set the appearance of the app (= the window title bars) to light or dark.

And of course more things can be added in the future. 

For now, I used this to make the following possible:
- Make the appearance of the window title bars follow MuseScore's theme
- Give the user the possibility to let MuseScore switch it's theme along with the system's theme. That seems a controversial subject: I am personally a great fan of this, but Tantacrul didn't really like it. Anyway, every user can choose what they like. I think it would be good to ask users which theme they want when they launch MuseScore for the first time.

https://user-images.githubusercontent.com/48658420/103769733-1fa74500-5025-11eb-9713-38c9cfe8f5b9.mov

(I also fixed a typo in theme.cpp: currentThemeProperites -> currentThemeProperties)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -42,6 +42,10 @@ if (BUILD_UNIT_TESTS)
     add_subdirectory(libmscore/tests)
 endif(BUILD_UNIT_TESTS)
 
+if (OS_IS_MAC)
+    add_subdirectory(macos)
+endif()
+
 if (OS_IS_WASM)
     add_subdirectory(wasmtest)
 endif()
--- a/src/framework/ui/internal/uiconfiguration.cpp
+++ b/src/framework/ui/internal/uiconfiguration.cpp
@@ -27,8 +27,15 @@ void UiConfiguration::init()
     settings()->setDefaultValue(UI_MUSICAL_FONT_SIZE_KEY, Val(12));
 
     settings()->valueChanged(UI_THEME_TYPE_KEY).onReceive(nullptr, [this](const Val& val) {
-        m_currentThemeTypeChannel.send(static_cast<ThemeType>(val.toInt()));
+        m_currentPreferredThemeTypeChannel.send(static_cast<ThemeType>(val.toInt()));
+        m_currentActualThemeTypeChannel.send(actualThemeType());
     });
+
+#ifdef Q_OS_MAC
+    macos()->systemDarkModeSwitched().onReceive(nullptr, [this](const bool) {
+        m_currentActualThemeTypeChannel.send(actualThemeType());
+    });
+#endif
 
     settings()->valueChanged(UI_FONT_FAMILY_KEY).onReceive(nullptr, [this](const Val&) {
         m_fontChanged.notify();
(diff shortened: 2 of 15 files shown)


cbjeukendrup:
The Mac build is failing... but on my own Mac it worked flawlessly!
It says:
```
error: Objective-C was disabled in PCH file but is currently enabled
```
I think it is because CI uses Ninja and I use Xcode. The problem is similar to what Antonio solved in PR #5660. I will have a look, but @igorkorsukov, maybe you could guide me a bit...

igorkorsukov:
Great job!

But I believe there is no need to add a `macos` module.
If we add the `macos` module, it will introduce confusion and some complications.

The first problem is that we have to add all macOS-specific functions to this module, or add some and some not.
For example, should we add an audio driver to this module? Or file storage paths?

Now the architecture of the project is such that we can exclude some modules (at the stage of compilation and / or execution) and in this case the code of the excluded module will be absent in the application and nothing associated with this module will be executed.
For example, for the "converter" mode of operation, we can exclude the UI module, because in this mode the application is launched in the console mode without GUI. And we expect that nothing will be executed, which is related to UI.
For example, the `observeDistributedNotifications("AppleInterfaceThemeChangedNotification"...` code will not be executed.
If there is a separate `macos` module, it will be difficult to make this behavior, it will be necessary to check or what conditions (and duplicate logic) or check which modules are added and they need functions, and which are not added and they do not need functions, or always execute everything, regardless of whether you need it or not.

Better make how in the audio module with audio drivers - in each module that has platform-specific functions, add a platform directory and in it add implementations for each platform of the functions necessary for this module. 
https://github.com/musescore/MuseScore/tree/master/src/framework/audio/internal/platform

igorkorsukov:
> The Mac build is failing... but on my own Mac it worked flawlessly!
> It says:
> 
> ```
> error: Objective-C was disabled in PCH file but is currently enabled
> ```
> 
> I think it is because CI uses Ninja and I use Xcode. The problem is similar to what Antonio solved in PR #5660. I will have a look, but @igorkorsukov, maybe you could guide me a bit...

I'll see what I can do about it later. Thank you.

igorkorsukov:
And the third thing I want to draw your attention to.
Better to have a common interface and implement it for each platform.
Then there will be no platform checks in the common code (not platform specific)
```
#ifdef Q_OS_MAC
    macos()->systemDarkModeSwitched().onReceive(nullptr, [this](const bool) {
        m_currentActualThemeTypeChannel.send(actualThemeType());
    });
#endif
```

For example, switching to the dark theme is also available on other platforms.
i.e. in this case it is better to do something like this

1. Add an interface like `IPlatformTheme`
```
class IPlatformTheme ...
{
...
    virtual bool isDarkMode() const = 0;
    virtual async::Notification systemDarkModeSwitched() const = 0;
}
```

2. Add `ui/internal/platform/macos/MacOsPlatformTheme` implementation
3. Add `ui/internal/platform/stub/StubPlatformTheme` implementation - stub for other plarforms
4. Add registration of the required implementation (like registration audio driver)
```
void Ui::registerExports()
{
...
#ifdef Q_OS_MACOS
    framework::ioc()->registerExport<IPlatformTheme>(moduleName(), new MacOsPlatformTheme());
#else  
    framework::ioc()->registerExport<IPlatformTheme>(moduleName(), new StubPlatformTheme());
#endif
...
}
```

5. Use in common code without platform check
```
    platformTheme()->systemDarkModeSwitched().onNotify(this, [this]() {
        ...
    });
```

cbjeukendrup:
Thank you! That looks indeed like a better idea. I will definitely use that!

cbjeukendrup:
One question, shall I put the platform-specific files in `ui/view/platform/...` (there is also `theme.cpp`) or shall I put it in `ui/internal/platform/...`?

igorkorsukov:
> One question, shall I put the platform-specific files in `ui/view/platform/...` (there is also `theme.cpp`) or shall I put it in `ui/internal/platform/...`?

Better ui/internal/platform

cbjeukendrup:
@igorkorsukov I did a second version with your suggestions in PR #7207!

PR #7207, opened 2021-01-07
Implement PlatformTheme and macOS Dark Mode improvements

cbjeukendrup:
Replaces PR #7206

Through PlatformTheme, we can now:
- Check whether the current platform has a Dark Mode that we can follow;
- Check whether that Dark Mode is currently enabled;
- Get notifications when the platform's Dark Mode is switched, so that MuseScore can switch along;
- Set the 'appearance' of the app; on macOS that includes the style of the window title bar, light or dark. This is used to make the window title bar color follow MuseScore's theme.

The platform-specific code is entirely 'hidden' behind `IPlatformTheme`, so that there are as little platform checks as possible. 
For platforms other than macOS, there is no real implementation (yet), but with this structure, it will be very easy to add that if we want.

https://user-images.githubusercontent.com/48658420/103769733-1fa74500-5025-11eb-9713-38c9cfe8f5b9.mov

### DIFF
--- a/src/framework/ui/CMakeLists.txt
+++ b/src/framework/ui/CMakeLists.txt
@@ -22,6 +22,27 @@ set(MODULE ui)
 set(MODULE_QRC ui.qrc)
 set(MODULE_QML_IMPORT ${CMAKE_CURRENT_LIST_DIR}/qml)
 
+include(GetPlatformInfo)
+if (OS_IS_MAC)
+    set(PLATFORM_THEME_SRC
+        ${CMAKE_CURRENT_LIST_DIR}/internal/platform/macos/macosplatformtheme.mm
+        ${CMAKE_CURRENT_LIST_DIR}/internal/platform/macos/macosplatformtheme.h
+    )
+    # Don't mix C++ and Objective-C++ in Unity Build
+    set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}/internal/platform/macos/macosplatformtheme.mm
+                                PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
+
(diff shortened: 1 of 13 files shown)


cbjeukendrup:
Hm, some errors... I will have a look.

AntonioBL:
Ciao.
For me the solution was to separate the cocoabridge.mm as an independent target and set its compile flags different from the global flags (because it is objective-C), see https://github.com/musescore/MuseScore/pull/5660/files#diff-d0bd26050b66c7303012694a65be2bee30494e77f927d7f14a30b9012b16a8ffR147

igorkorsukov:
> Edit: Maybe you can disable the use of PCH for that file by following this approach: https://stackoverflow.com/questions/40886738/disable-precompiled-header-for-a-single-file/59803632#59803632

This looks like a good variant

cbjeukendrup:
> Edit: Maybe you can disable the use of PCH for that file by following this approach: https://stackoverflow.com/questions/40886738/disable-precompiled-header-for-a-single-file/59803632#59803632

Thanks @AntonioBL ! I have force-pushed, so we will see what the build does :)

cbjeukendrup:
> Edit: Maybe you can disable the use of PCH for that file by following this approach: https://stackoverflow.com/questions/40886738/disable-precompiled-header-for-a-single-file/59803632#59803632

Thanks @AntonioBL ! I have force-pushed, so we will see what the build does :)

igorkorsukov (approved):

igorkorsukov:
@cbjeukendrup Great job!

cbjeukendrup:
@igorkorsukov Thank you!

PR #7210, opened 2021-01-09
Draw Ambitus with Flat Caps instead of Round Caps

cbjeukendrup:
Draw Ambitus with Flat Caps instead of Round Caps, to match normal ledger lines and stems.

Schermafbeelding 2021-01-09 om 14 53 32

This is of course needed for the master branch too. Should I make a separate PR for that or can it be cherry-picked?

### DIFF
--- a/libmscore/ambitus.cpp
+++ b/libmscore/ambitus.cpp
@@ -454,7 +454,7 @@ void Ambitus::draw(QPainter* p) const
       {
       qreal _spatium = spatium();
       qreal lw = lineWidth().val() * _spatium;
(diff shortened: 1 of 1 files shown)


cbjeukendrup:
The test failure is as expected. Should I do anything to update the tests?

Marr11317:
the vtest compare against master, so you shouldn`t have to do anything.

cbjeukendrup:
Interesting that the vtests still fail despite the label. Apparently the condition [here](https://github.com/musescore/MuseScore/actions/runs/474115439/workflow#L83) is not working for some reason...

PR #7220, opened 2021-01-11
3.5.2_backend_porting

vpereverzev:
Ported the latest changes from 3.5.2_backend for musescore.com


Jojo-Schmitz:
That new commandline option needs to get added to the man-page too, https://github.com/musescore/MuseScore/blob/3.x/build/Linux%2BBSD/mscore.1.in

Jojo-Schmitz:
Ignoring the vtest failures are not the right thing, if those are due to crashes!

vpereverzev:
> Ignoring the vtest failures are not the right thing, if those are due to crashes!

it's not related to this PR

Jojo-Schmitz:
Hmm, but to what then?

vpereverzev:
> Hmm, but to what then?

dunno, but it's already in 3.x, I'm trying to fix it now

Jojo-Schmitz:
Some stuff from #7218 ?

PR #7223, opened 2021-01-12
Enable Sicilianu again and update man-page

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/312513 and fixes an aftermath of PR #7220


Jojo-Schmitz:
Segmentation vaults during vtests. Certainly not my fault !?!

PR #7233, opened 2021-01-13
page_margins_fix

vpereverzev:
Resolves https://trello.com/c/5oMkjkGm


MarcSabatella:
I trust this works somehow, but I'm not understanding what it is really doing.

MarcSabatella:
I still don't understand what this does, but FWIW, it doesn't seem to have any effect on margins in new scores I create - they still show that 15.06 mm.  That's after removing my custom MSS style forcing letter as the default.

PR #7235, opened 2021-01-13
[MU4] remove MIDI2.0 UMP deprecated methods

psmokotnin:
- Added constructor for midi::Event from Ms::MidiCoreEvent
- Add convertor from MIDI 1.0 Channel Voice to Ms::MidiCoreEvent
- Remove deprecated methods
- Fix an assert if notation playback needs more than 16 channels

### DIFF
--- a/src/framework/midi/internal/fluidsynth.cpp
+++ b/src/framework/midi/internal/fluidsynth.cpp
@@ -230,28 +230,32 @@ bool FluidSynth::handleEvent(const Event& e)
     }
 
     int ret = FLUID_OK;
-    switch (e.type()) {
-    case EventType::ME_NOTEON: {
-        ret = fluid_synth_noteon(m_fluid->synth, e.channel(), e.note(), e.velocity());
+    int channel = 0;
+    if (e.isChannelVoice()) {
+        channel = e.channel() + e.group() * 16;
+    }
+    switch (e.opcode()) {
+    case Event::Opcode::NoteOn: {
+        ret = fluid_synth_noteon(m_fluid->synth, channel, e.note(), e.velocity());
     } break;
-    case EventType::ME_NOTEOFF: {
-        ret = fluid_synth_noteoff(m_fluid->synth, e.channel(), e.note());
+    case Event::Opcode::NoteOff: {
+        ret = fluid_synth_noteoff(m_fluid->synth, channel, e.note());
     } break;
-    case EventType::ME_CONTROLLER: {
+    case Event::Opcode::ControlChange: {
         if (e.index() == CntrType::CTRL_PROGRAM) {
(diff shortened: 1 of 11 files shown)


src/framework/midi/internal/fluidsynth.cpp
```diff
@@ -230,28 +230,32 @@ bool FluidSynth::handleEvent(const Event& e)
     }
 
     int ret = FLUID_OK;
-    switch (e.type()) {
-    case EventType::ME_NOTEON: {
-        ret = fluid_synth_noteon(m_fluid->synth, e.channel(), e.note(), e.velocity());
+    int channel = 0;
+    if (e.isChannelVoice()) {
+        channel = e.channel() + e.group() * 16;
```
igorkorsukov, line 235: This is magic, and this logic is duplicated. Better to do something about it.

src/framework/midi/internal/fluidsynth.cpp
```diff
@@ -230,28 +230,32 @@ bool FluidSynth::handleEvent(const Event& e)
     }
 
     int ret = FLUID_OK;
-    switch (e.type()) {
-    case EventType::ME_NOTEON: {
-        ret = fluid_synth_noteon(m_fluid->synth, e.channel(), e.note(), e.velocity());
+    int channel = 0;
+    if (e.isChannelVoice()) {
```
igorkorsukov, line 234: If the event is not `isChannelVoice`, this is an exception? or in this case, the channel must be 0?

src/framework/midi/internal/zerberussynth.cpp
```diff
@@ -134,27 +134,31 @@ bool ZerberusSynth::handleEvent(const Event& e)
...
         NOT_IMPLEMENTED;
     } break;
-    case EventType::ME_PITCHBEND: {
+    case Event::Opcode::PitchBend: {
         ret = false;
         NOT_IMPLEMENTED;
     } break;
     default: {
-        NOT_SUPPORTED << " event type: " << static_cast<int>(e.type());
+        NOT_SUPPORTED << "not supported event opcode: " << e.opcodeString();
```
igorkorsukov, line 161: the `NOT_SUPPORTED` macro output `not supported`, that is, there will be a duplicate message.

src/framework/midi/midievent.h
```diff
@@ -95,17 +89,22 @@ struct Event {
         setOpcode(opcode);
     }
 
-    ME_DEPRECATED Event(channel_t ch, EventType type)
+    Event(Ms::MidiCoreEvent event)
```
igorkorsukov, line 92: This constructor and the `toMSEvent` method are equivalent to methods that have been marked deprecated and intended to get rid of.

PR #7238, opened 2021-01-14
[MU4] List view delegate

RomanPudashkin:


src/framework/uicomponents/qml/MuseScore/UiComponents/qmldir
```diff
@@ -30,3 +30,4 @@ InstallationPanel 1.0 InstallationPanel.qml
 StyledScrollBar 1.0 StyledScrollBar.qml
 GradientTabButton 1.0 GradientTabButton.qml
 GridViewSectional 1.0 GridViewSectional.qml
+ListItemBlank 1.0 ListItemBlank.qml
```
igorkorsukov, line 33: maybe just `ListItem`?

Eism: in my opinion, blank describes the use of this component better

igorkorsukov: It confuses me that it is often used as a standalone element, and not just as a base class for other elements. that is, it is not abstract, but can be used independently.

Eism: But the component is not used as a standalone element. In each use, Roman adds content(
for example StyledTextLabel). If a component is used as a standalone element, then there will be nothing on it, only a background.

It may be better to add a content property to the component, and explicitly specify when using the component, then it looks more like an abstract component.

src/framework/uicomponents/qml/MuseScore/UiComponents/ListItemBlank.qml
```diff
@@ -0,0 +1,83 @@
...
+    Accessible.selectable: true
+    Accessible.selected: isSelected
+
+    Rectangle {
+        id: background
+
+        anchors.fill: parent
+
+        color: "transparent"
+        opacity: root.enabled ? 1 : ui.theme.itemOpacityDisabled
```
Eism, line 26: better to move this condition to a separate state

src/framework/uicomponents/qml/MuseScore/UiComponents/qmldir
igorkorsukov: I see, I agree with `ListItemBlank` name :)

Eism (approved):

PR #7240, opened 2021-01-14
[MU4] Inspector actualization

RomanPudashkin:


src/inspector/models/general/playback/internal_models/arpeggioplaybackmodel.cpp
```diff
@@ -2,10 +2,12 @@
 
 #include "dataformatter.h"
 
+using namespace mu::inspector;
+
 ArpeggioPlaybackModel::ArpeggioPlaybackModel(QObject* parent, IElementRepositoryService* repository)
     : AbstractInspectorModel(parent, repository)
 {
-    setTitle(tr("Arpeggio"));
+    setTitle(translate("Arpeggio"));
```
igorkorsukov, line 10: This is the solution not working.
When running `lupdate` it parses source and do search pattern `tr(...` extract data and generate `*.ts` file. Possible changes the pattern or add a custom pattern. Our custom pattern will be `trc(..` and `qtrc(..`. So you need to use them.

RomanPudashkin: This method uses qtrc inside it (see inspector/internal/translation.h)

igorkorsukov: `lupdate` just a text parser, not a compiler, so it can't call `translate` function and get context

RomanPudashkin: fixed

Eism (approved):

igorkorsukov (approved):

PR #7241, opened 2021-01-14
fix #315529: crash on load of score with repeat

MarcSabatella:
Resolves: https://musescore.org/en/node/315529

Crash happens trying to find the last element before an end barline,
since in the case of start repeats, there may be no end barline.
Fix is to provide an initialization to the variable used.


Jojo-Schmitz:
MSVC even emits a warning about this...

Jojo-Schmitz (approved):

igorkorsukov (approved):

PR #7246, opened 2021-01-15
[MU3] Try fix windows msi sign

igorkorsukov:
try fix windows msi sign

### DIFF
--- a/build/ci/windows/package.bat
+++ b/build/ci/windows/package.bat
@@ -133,12 +133,12 @@ ECHO "Start msi packing..."
 :: sign dlls and exe files
 IF %DO_SIGN% == ON (
     where /q secure-file
-    IF ERRORLEVEL 1 ( choco install -y choco install -y --ignore-checksums secure-file )
+    IF ERRORLEVEL 1 ( choco install -y --ignore-checksums secure-file )
     secure-file -decrypt build\ci\windows\resources\musescore.p12.enc -secret %SIGN_CERTIFICATE_ENCRYPT_SECRET%
 
     for /f "delims=" %%f in ('dir /a-d /b /s "%INSTALL_DIR%\*.dll" "%INSTALL_DIR%\*.exe"') do (
         ECHO "Signing %%f"
-        %SIGNTOOL% sign /f "build\ci\windows\resources\musescore.p12" /t http://timestamp.verisign.com/scripts/timstamp.dll /p %SIGN_CERTIFICATE_PASSWORD% "%%f"
+        %SIGNTOOL% sign /f "build\ci\windows\resources\musescore.p12" /t http://timestamp.globalsign.com/scripts/timstamp.dll /p %SIGN_CERTIFICATE_PASSWORD% "%%f"
     )
 
 ) ELSE (


Jojo-Schmitz:
It is not just the MSI, the PortabelApp is affected too as far as I can tell

build/ci/windows/package.bat
```diff
@@ -47,7 +47,7 @@ IF %BUILD_WIN_PORTABLE% == ON (
 
 :: Setup package type
 IF %BUILD_WIN_PORTABLE% == ON    ( SET PACKAGE_TYPE="portable") ELSE (
-IF %BUILD_MODE% == devel_build   ( SET PACKAGE_TYPE="7z") ELSE (
+IF %BUILD_MODE% == devel_build   ( SET PACKAGE_TYPE="msi") ELSE (
```
igorkorsukov, line 50: temporary for the test will be back

build/ci/windows/package.bat
```diff
@@ -133,12 +133,12 @@ ECHO "Start msi packing..."
...
 IF %DO_SIGN% == ON (
     where /q secure-file
-    IF ERRORLEVEL 1 ( choco install -y choco install -y --ignore-checksums secure-file )
+    IF ERRORLEVEL 1 ( choco install -y --ignore-checksums secure-file )
     secure-file -decrypt build\ci\windows\resources\musescore.p12.enc -secret %SIGN_CERTIFICATE_ENCRYPT_SECRET%
 
     for /f "delims=" %%f in ('dir /a-d /b /s "%INSTALL_DIR%\*.dll" "%INSTALL_DIR%\*.exe"') do (
         ECHO "Signing %%f"
-        %SIGNTOOL% sign /f "build\ci\windows\resources\musescore.p12" /t http://timestamp.verisign.com/scripts/timstamp.dll /p %SIGN_CERTIFICATE_PASSWORD% "%%f"
+        %SIGNTOOL% sign /f "build\ci\windows\resources\musescore.p12" /t http://timestamp.globalsign.com/scripts/timstamp.dll /p %SIGN_CERTIFICATE_PASSWORD% "%%f"
```
igorkorsukov, line 141: Now http://timestamp.verisign.com/scripts/timstamp.dll, no response 
So try to use alternative
https://stackoverflow.com/questions/9714798/http-timestamp-verisign-com-scripts-timstamp-dll-not-available

igorkorsukov:
> It is not just the MSI, the PortabelApp is affected too as far as I can tell

hmm, PortabelApp we don't sign

build/ci/windows/package.bat
igorkorsukov: For testing in my repo, where available secrets
https://github.com/igorkorsukov/MuseScore/pull/81

igorkorsukov:
> > It is not just the MSI, the PortabelApp is affected too as far as I can tell
> 
> hmm, PortabelApp we didn't sign any

It looks like we have two problems:
1. MSI is not signed, so a warning is shown in the Windows
2. No publisher set

This PR fix problem 1, the next PR will fix issue 2

igorkorsukov:
I tested in my repo, msi signature works. Windows does not show a warning message.

cbjeukendrup:
I also found this solution yesterday, but when I tested it, the signing still didn't work. Probably that has to do with GitHub Secrets.
I found this yesterday about VeriSign: https://stackoverflow.com/questions/65541786/is-the-verisign-timestamp-server-down
In line 198 is also something about VeriSign, maybe that needs to be fixed too.

igorkorsukov:
> In line 198 of package.bat is also something about VeriSign, maybe that needs to be fixed too.

Yes, I noticed late.
Fixed now https://github.com/musescore/MuseScore/pull/7247

igorkorsukov:
> It seems that VeriSign went down quite recently, so that explains why we didn't see problems before.

It looks like it breaks down from time to time, Google search shows problems in 2012, 14, 18, 19, etc.

igorkorsukov:
Later I will try to make it so that several servers are used, if one is not available, the other is used ...

igorkorsukov:
> I found this yesterday about VeriSign: https://stackoverflow.com/questions/65541786/is-the-verisign-timestamp-server-down  

"So the http://timestamp.verisign.com timestamp server is no more." - ok :)

PR #7249, opened 2021-01-15
[MU3] Fix MSVC compiler warnings

Jojo-Schmitz:


libmscore/measure.cpp
```diff
@@ -2897,7 +2897,7 @@ bool Measure::isCutawayClef(int staffIdx) const
             etrack = strack + VOICES;
             }
       // find segment before EndBarLine
-      Segment* s;
+      Segment* s = nullptr;
```
Jojo-Schmitz, line 2900: same as in #7241

mscore/musescore.h
```diff
@@ -684,7 +684,7 @@ class MuseScore : public QMainWindow, public MuseScoreCore {
       void loadFile(const QUrl&);
       QTemporaryFile* getTemporaryScoreFileCopy(const QFileInfo& info, const QString& baseNameTemplate);
       QNetworkAccessManager* networkManager();
-      virtual Score* openScore(const QString& fn, bool switchTab = true, const bool appendToExistingTabs = true);
+      virtual Score* openScore(const QString& fn, bool switchTab = true, const bool appendToExistingTabs = true) override;
```
Jojo-Schmitz, line 687: Review needed, it does fix a compiler warning but I'm not sure whether we really want to override `Ms::MuseScoreCore::openScore()` from libmscore/musescoreCore.h line 45 here?

Howard-C: Does it matter? I think it is going to be overridden with or without the "override" keyword

Jojo-Schmitz: It was implicitly overridden, is now explcitly overridden, that apeases the compiler. Question is whether this really wanted.

Howard-C: I'm just saying even if that wasn't what we want, it still would not be solved (returned to the behaviour of parent class) just by not adding `override`. Am I wrong?

Jojo-Schmitz: No. And that's why I started this discussion here ;-)

PR #7250, opened 2021-01-15
[MU3] Don't disable Qt 5.15 for 3.x

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/315557 (forum topic)

Disabling Qt 5.15 seems no longer needed (not with 5.15.2 at least), it doesn't crash any longer on Linux nor on Windows (where that disabling hadn't worked anyhow).

### DIFF
--- a/build/FindQt5.cmake
+++ b/build/FindQt5.cmake
@@ -36,14 +36,6 @@ if (WIN32)
       )
 endif(WIN32)
 
-# For Windows, because of these lines, for some unknown reason, the build of the .msi package fails.
-if(NOT ${CMAKE_HOST_SYSTEM_NAME} MATCHES "Windows")
-    find_package(Qt5Core 5.15.0 QUIET)
-    if (Qt5Core_FOUND)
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Could someone on a Mac try this out? Building a Mac version with Qt 5.15.2 and load a pre 3.6 score there, to check whather that causes a crash?

cbjeukendrup:
> Could someone on a Mac try this out? Building a Mac version with Qt 5.15.2 and load a pre-3.6 score there, to check whether that causes a crash?

Will do!

cbjeukendrup:
Result: not any problem! 
(only a bunch of deprecation warnings, but that's understandable.)

Jojo-Schmitz:
Thanks! Getting rid of this stuff makes if far easier to switch between 3.x and master, and most probably would have led **me** to detect and fix the warnings from #7249 earlier and hence avoided the crash from [#315529](https://musescore.org/en/node/315529) to make it into the 3.6 release...

PR #7254, opened 2021-01-15
Fix #315553 - MuseScore 3.6 Symphony Orchestra Template Violin Errors

njvdberg:
Resolves: https://musescore.org/en/node/315553

Correct the order of Violins 1 and 2.

### DIFF
--- a/share/templates/08-Orchestral/02-Symphony_Orchestra.mscx
+++ b/share/templates/08-Orchestral/02-Symphony_Orchestra.mscx
@@ -1339,11 +1339,13 @@
           <name>stdNormal</name>
           </StaffType>
         <bracket type="0" span="5" col="0"/>
+        <bracket type="2" span="2" col="1"/>
+        <barLineSpan>1</barLineSpan>
         </Staff>
-      <trackName>Violins 2</trackName>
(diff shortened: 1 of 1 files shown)


share/templates/08-Orchestral/02-Symphony_Orchestra.mscx
```diff
@@ -1339,11 +1339,13 @@
           <name>stdNormal</name>
           </StaffType>
         <bracket type="0" span="5" col="0"/>
+        <bracket type="2" span="2" col="1"/>
```
vpereverzev, line 1342: I'm pretty confident about all the changes, but not sure about this particular line
Do we really need this?

njvdberg: Yes, this line is needed. It connects <code>Violins 1</code> and <code>Violins 2</code> with a thin bracket, similar to e.g. <code>Flute 1</code> and <code>Flute 2</code>.
Of course we could rely on the automatic bracketing to add this thin bracket but then we should remove similar line at other places too. However, the advantage of having these lines here is, when somebody simple open this template instead of creating a new score using this file as a template, all bracketing is as it should be.

PR #7256, opened 2021-01-16
[MU3] Export hidden tempo text to MusicXML

matangover:
Hidden tempo text is now exported as a `<sound tempo="XX"/>` element without a container `<direction>` element.

This is now in line with the way MuseScore imports `<sound>` elements, see: https://github.com/musescore/MuseScore/pull/7171

I changed testSound1 from a test with a reference to a roundtrip import-export test, because it is now fully supported. Some other test reference files had to be changed because they contain hidden tempo markings as well.


Jojo-Schmitz:
rebase needed.

matangover:
rebase done

mtest/musicxml/io/tst_mxml_io.cpp
```diff
@@ -171,7 +171,7 @@ private slots:
       void slurTieLineStyle() { mxmlIoTest("testSlurTieLineStyle"); }
       void slurs() { mxmlIoTest("testSlurs"); }
       void slurs2() { mxmlIoTest("testSlurs2"); }
-      void sound1() { mxmlIoTestRef("testSound1"); }
+      void sound1() { mxmlIoTest("testSound1"); }
```
Jojo-Schmitz, line 174: Why no longer test against the ref file, but at the ssame tuime updating that in the same PR?

Jojo-Schmitz:

PR #7257, opened 2021-01-16
[MU3] Added several timeservers for windows sign on CI

igorkorsukov:
Added several timeservers for windows sign on CI

### DIFF
--- a/build/ci/windows/sign.bat
+++ b/build/ci/windows/sign.bat
@@ -1,0 +1,96 @@
+@echo off
+setlocal ENABLEDELAYEDEXPANSION
+ECHO "MuseScore sign"
+
+SET CERT_ENCRYPT_SECRET=""
+SET CERT_PASSWORD=""
+SET FILES_DIR=""
+SET FILE_PATH=""
+SET FILE_NAME=""
+
+SET SIGNTOOL="C:\Program Files (x86)\Windows Kits\10\bin\x64\signtool.exe"
+SET SIGNCERT="build\ci\windows\resources\musescore.p12"
+SET SIGNCERT_ENCRYPTED="build\ci\windows\resources\musescore.p12.enc"
+
+SET TIMESERVER1="http://timestamp.globalsign.com/scripts/timstamp.dll"
+SET TIMESERVER2="http://timestamp.sectigo.com"
+SET TIMESERVER3="http://tsa.starfieldtech.com"
+
+:GETOPTS
+IF /I "%1" == "--dir" SET FILES_DIR=%2& SHIFT
+IF /I "%1" == "--file" SET FILE_PATH=%2& SHIFT
+IF /I "%1" == "--name" SET FILE_NAME=%2& SHIFT
+IF /I "%1" == "--secret" SET CERT_ENCRYPT_SECRET=%2& SHIFT
+IF /I "%1" == "--pass" SET CERT_PASSWORD=%2& SHIFT
+SHIFT
+IF NOT "%1" == "" GOTO GETOPTS
+
+IF %FILES_DIR% == "" ( 
+    SET MODE="file"
+) ELSE (
+    SET MODE="dir"
+)
+
+ECHO "MODE: %MODE%"
+ECHO "FILES_DIR: %FILES_DIR%"
+ECHO "FILE_PATH: %FILE_PATH%"
+ECHO "FILE_NAME: %FILE_NAME%"
+
+ECHO "Install secure-file"
+WHERE /q secure-file
+IF ERRORLEVEL 1 ( choco install -y --ignore-checksums secure-file )
+
+ECHO "Decrypt certificate"
+secure-file -decrypt %SIGNCERT_ENCRYPTED% -secret %CERT_ENCRYPT_SECRET%
+
+:: Sign files in dir
+IF %MODE% == "dir" (
+    ECHO "Sign files in dir: %FILES_DIR%"
+    FOR /f "delims=" %%f in ('dir /a-d /b /s "%FILES_DIR%\*.dll" "%FILES_DIR%\*.exe"') do (
+        ECHO.
+        ECHO "======================="
+        ECHO "Signing %%f"
+        %SIGNTOOL% sign /f %SIGNCERT% /t %TIMESERVER1% /p %CERT_PASSWORD% "%%f"
+        IF ERRORLEVEL 1 (
+            ECHO "signtool return error level: %ERRORLEVEL%, try use TIMESERVER2"
+            CALL %SIGNTOOL% sign /f %SIGNCERT% /t %TIMESERVER2% /p %CERT_PASSWORD% "%%f"
+        )
+        IF ERRORLEVEL 1 (
+            ECHO "signtool return error level: %ERRORLEVEL%, try use TIMESERVER3"
+            %SIGNTOOL% sign /f %SIGNCERT% /t %TIMESERVER3% /p %CERT_PASSWORD% "%%f"
+        )
+        IF ERRORLEVEL 1 (
+            ECHO "signtool return error level: %ERRORLEVEL%, sign failed"
+            exit /b 1 
+        )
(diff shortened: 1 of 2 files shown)


build/ci/windows/sign.bat
```diff
@@ -0,0 +1,66 @@
...
+SET CERT_PASSWORD=""
+SET FILES_DIR=""
+SET FILE_PATH=""
+SET FILE_NAME=""
+
+SET SIGNTOOL="C:\Program Files (x86)\Windows Kits\10\bin\x64\signtool.exe"
+SET SIGNCERT="build\ci\windows\resources\musescore.p12"
+
+::SET TIMESERVER1="http://timestamp.globalsign.com/scripts/timstamp.dll"
+SET TIMESERVER1="http://timestamp.verisign.com/scripts/timstamp.dll"
```
Jojo-Schmitz, line 14: This one doesn't seem to work

build/ci/windows/sign.bat
```diff
@@ -0,0 +1,66 @@
...
+SET TIMESERVER2="http://timestamp.sectigo.com"
```
Jojo-Schmitz, line 15: This one redirects to an https page, https://sectigo.com/resource-library/time-stamping-server

build/ci/windows/sign.bat
```diff
@@ -0,0 +1,66 @@
...
+SET CERT_ENCRYPT_SECRET=""
+SET CERT_PASSWORD=""
+SET FILES_DIR=""
+SET FILE_PATH=""
+SET FILE_NAME=""
+
+SET SIGNTOOL="C:\Program Files (x86)\Windows Kits\10\bin\x64\signtool.exe"
+SET SIGNCERT="build\ci\windows\resources\musescore.p12"
+
+::SET TIMESERVER1="http://timestamp.globalsign.com/scripts/timstamp.dll"
```
Jojo-Schmitz, line 13: This one redirectes to an https page, https://www.globalsign.com/en/timestamp-service

build/ci/windows/sign.bat
```diff
@@ -0,0 +1,66 @@
...
+SET TIMESERVER1="http://timestamp.verisign.com/scripts/timstamp.dll"
+SET TIMESERVER2="http://timestamp.sectigo.com"
+SET TIMESERVER3="http://tsa.starfieldtech.com"
```
Jojo-Schmitz, line 16: This one does not provide an https page, do we really want to pass username/password across such a connection?

build/ci/windows/sign.bat
igorkorsukov: This is for test :)

build/ci/windows/sign.bat
igorkorsukov: I will test, I think it works without any username and password, because timestamp works on other port, not default 80 or 443

build/ci/windows/sign.bat
Jojo-Schmitz: I know...

build/ci/windows/sign.bat
igorkorsukov: I did not understand the question first. As far as I see, through this connection, nothing is transmitted, only get a timestamp

Jojo-Schmitz: Not if that port isn't getting explicity specified on the URL, like http://tsa.starfieldtech.com:13 (daytime) or http://tsa.starfieldtech.com:37 (time) or http://tsa.starfieldtech.com:123 (ntp). No need to try: neither of these work ;-)

igorkorsukov: I think the signtool.exe utility can open the TCP connection to the right port on this url. ie the port determines the specification of obtaining a timestamp

igorkorsukov: Explicit specify the port needs for the browser, because it does not know what we want from it :)

Jojo-Schmitz: The signtool probably needs just hostnames, not the protocol, so the entire "http://" stuff should not be needed

Jojo-Schmitz: Hmm, not sure, it claims that the `/t` option takes a URL as an additional parameter, not just a plain hostname?

igorkorsukov: Yes, I agree. just so used to work and so publish the urls

Jojo-Schmitz: Ther example at https://docs.microsoft.com/de-de/windows/win32/seccrypto/adding-time-stamps-to-previously-signed-files says: `igntool timestamp -t https://timestamp.digicert.com MyControl.exe` so a) really a URL and b) using https ;-)

igorkorsukov: In any case, I will check the work of each address :)

Jojo-Schmitz:
https://github.com/musescore/MuseScore/pull/7257/checks?check_run_id=1720521734#step:7:57  "Encrupt sert"???

build/ci/windows/sign.bat
```diff
@@ -0,0 +1,95 @@
...
+ECHO "MODE: %MODE%"
+ECHO "FILES_DIR: %FILES_DIR%"
+ECHO "FILE_PATH: %FILE_PATH%"
+ECHO "FILE_NAME: %FILE_NAME%"
+
+ECHO "Install secure-file"
+WHERE /q secure-file
+IF ERRORLEVEL 1 ( choco install -y --ignore-checksums secure-file )
+
+ECHO "Encrupt sert"
```
Jojo-Schmitz, line 42: Typo... should be "Encrypt cert"  I guess ;-)

Jojo-Schmitz: Or rather "Decrypt cert", as that's what it seems to be doing

igorkorsukov: Fixed, thanks!

igorkorsukov:
Finished this task.
I tested it, all timeservers are working.
If the first does not work, then the 2nd will be used, if the second does not work, then the 3rd will be used.
Also, if we fail to sign for some reason, then the entire build will fail.

build/ci/windows/sign.bat
```diff
@@ -0,0 +1,95 @@
...
+ECHO "FILES_DIR: %FILES_DIR%"
+ECHO "FILE_PATH: %FILE_PATH%"
+ECHO "FILE_NAME: %FILE_NAME%"
+
+ECHO "Install secure-file"
+WHERE /q secure-file
+IF ERRORLEVEL 1 ( choco install -y --ignore-checksums secure-file )
+
+ECHO "Decrypt sert"
+secure-file -decrypt build\ci\windows\resources\musescore.p12.enc -secret %CERT_ENCRYPT_SECRET%
```
Jojo-Schmitz, line 43: Shouldn#t/couldn't this better be 

   secure-file -decrypt %CERT_ENCRYPT_SECRET%.enc -secret %CERT_ENCRYPT_SECRET%

build/ci/windows/sign.bat
```diff
@@ -0,0 +1,95 @@
...
+ECHO "MODE: %MODE%"
+ECHO "FILES_DIR: %FILES_DIR%"
+ECHO "FILE_PATH: %FILE_PATH%"
+ECHO "FILE_NAME: %FILE_NAME%"
+
+ECHO "Install secure-file"
+WHERE /q secure-file
+IF ERRORLEVEL 1 ( choco install -y --ignore-checksums secure-file )
+
+ECHO "Decrypt sert"
```
Jojo-Schmitz, line 42: still a typo, "sert", should be "cert"

build/ci/windows/sign.bat
igorkorsukov: changed

build/ci/windows/sign.bat
igorkorsukov: ops, fixed, thanks!

PR #7258, opened 2021-01-16
Fix #315627 - Musescore 3.6 crashes when opening large orchestration created in Musescore 3.4.2

njvdberg:
Resolves: https://musescore.org/en/node/315627

Solve a crash in case <code>nextMeasure()</code> returns a <code>nullptr</code> instead of a pointer to a measure.

### DIFF
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -2908,8 +2908,10 @@ bool Measure::isCutawayClef(int staffIdx) const
             return false;
       for (int track = strack; track < etrack; ++track) {
             Element* e = s->element(track);
-            if (e && e->isClef() && (nextMeasure()->system() == system() || toClef(e)->showCourtesy()))
-                return true;
+            if (!e || !e->isClef())
+                  continue;
+            if ((nextMeasure() && (nextMeasure()->system() == system())) || toClef(e)->showCourtesy())
+                  return true;
             }            
       return false;
       }


MarcSabatella:
Seems nextMeasure() will always return nullptr for the last measure.  So, why doesn't this crash on every score?  What are the conditions that lead to this?

njvdberg:
The modified check is in <code>Measure::isCutawayClef()</code> and this method returns <code>true</code> if
- 1 An <code>EndBarLine</code> segment is found and this segment  **and**
- 2 This segment has a <code>Clef</code> element **and**
- 3 The next measure is on the same system.

The original check was:
```
e && e->isClef() && (nextMeasure()->system() == system()
```

So, in most cases the <code>nextMeasure()</code> is never called, only when the element is a <code>Clef</code>.

As a matter of fact, the full original check was
```
e && e->isClef() && (nextMeasure()->system() == system() || toClef(e)->showCourtesy()
````
This potentially also could cause a crash on <code>toClef(e)->showCourtesy()</code> if the part before the <code>||</code> evaluates to <code>false</code> **and** <code>e</code> doesn't happens to be an <code>Clef</code>. This also solved by first check whether <code>e</code> is not zero **and** is a Clef.

MarcSabatella:
Answering my own question - this happens only if there is a staff whose last measure contains a clef as the last element before the barline.  Not an easy thing to make happen, but one way is to add a clef change somewhere then delete all measures after the change.  This results in the clef still being present but not visible.  I can also get one to be visible by entering a mid-measure clef between two half notes in 4/4 then changing the first to a whole note.

PR #7260, opened 2021-01-17
fix #290356: Allow copy-pasting of the LetRing, PalmMute and Vibrato elements.

dnovikov:
Resolves: https://musescore.org/en/node/290356

This code converts HairPin, LetRing, PalmMute and Vibrato to spanners, to utilize **setTick()** and **setTrack2()** methods. This prepares the ground to be used for any line element.

### DIFF
--- a/libmscore/paste.cpp
+++ b/libmscore/paste.cpp
@@ -534,6 +534,9 @@ void Score::readAddConnector(ConnectorInfoReader* info, bool pasteMode)
             case ElementType::TRILL:
             case ElementType::TEXTLINE:
             case ElementType::VOLTA:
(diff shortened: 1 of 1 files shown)


MarcSabatella:
That sounds like an ambitious goal, but surely there is a much simpler and a safer / more backwards-compatible fix for the short term?  Shouldn't it just be to add the appropriate handler to a switch statement?

dnovikov:
@MarcSabatella thank you for pointing out the right direction. I compiled the code with the switch statement only, and it worked fine.

MarcSabatella:
Nice!  I actually had no idea where the switch was :-), but I figured it had be something like this.

PR #7261, opened 2021-01-18
[MU3] Allow beaming across crotchet rests.

mattmcclinch:
Currently it is possible to create beams across rests that are less than a quarter note in duration by setting the beaming property of the rest to "Beam middle". This change allows us to honor the beaming properties of all rests, no matter the duration. See https://musescore.org/en/node/315677.


igorkorsukov:

PR #7263, opened 2021-01-18
Fix #315237: Don't switch tab when opening a score when a modal is open

cbjeukendrup:
Resolves: https://musescore.org/en/node/315237

- Don't switch tab when opening a score when a modal is open
- Make ScoreMigrationDialog QQuickWidget, so that it gets recognized by QApplication::activeModalWidget()
- (This caused an issue that the QML content wasn't focussed when the dialog was opened, which I fixed too.)

Result:
- When you open one score, it is appended to the tab bar, and its tab is activated
- The ScoreMigrationDialog for that score is shown
- When you open a second score, it is appended to the tab bar, but its tab is not activated, because a modal dialog is open (namely the ScoreMigrationDialog for score 1)
- When you close the dialog for score 1 and then switch to score 2 's tab, the ScoreMigrationDialog is shown for score 2.

(Maybe an even better solution would be to check not if there is *any* modal dialog open in the whole application, but to check whether there is a modal dialog *related to the current score* (i.e. when the style dialog is open, don't switch, but when the preferences dialog is open, do switch. But that's a bit of an edge case and IMHO not worth the complication.)

### DIFF
--- a/mscore/scoretab.cpp
+++ b/mscore/scoretab.cpp
@@ -332,6 +332,8 @@ void ScoreTab::setCurrent(int n)
             tab2->setVisible(false);
             }
       emit currentScoreViewChanged(v);
+
+      initScoreView(n);
       }
 
 //---------------------------------------------------------
@@ -555,26 +557,26 @@ void ScoreTab::removeTab(int idx, bool noCurrentChangedSignal)
 //   initScoreView
 //---------------------------------------------------------
 
-void ScoreTab::initScoreView(const int idx, const qreal logicalZoomLevel, const ZoomIndex zoomIndex, const qreal xoffset, const qreal yoffset)
+void ScoreTab::initScoreView(int idx)
       {
-      ScoreView* v = view(idx);
-      if (!v)  {
-            v = new ScoreView;
-            Score* sc = scoreList->value(idx);
-            if( sc != 0 )
-                  v->setScore(sc);
-            else {
-                  delete v;
-                  return;
+      auto it = scoreViewsToInit.find(idx);
+      if (it != scoreViewsToInit.end()) {
+            ScoreViewState s = it->second;
+            if (ScoreView* v = view(idx)) {
+                  v->setLogicalZoom(s.zoomIndex, s.logicalZoomLevel);
+                  v->setOffset(s.xOffset, s.yOffset);
                   }
-            QSplitter* vs = new QSplitter;
-            vs->setObjectName("score");
-            vs->setAccessibleName("");
-            vs->addWidget(v);
-            stack->addWidget(vs);
+            scoreViewsToInit.erase(it); // Init only one time
             }
-      v->setLogicalZoom(zoomIndex, logicalZoomLevel);
-      v->setOffset(xoffset, yoffset);
       }
-}
 
+//---------------------------------------------------------
+//   queueInitScoreView
+//---------------------------------------------------------
+
+void ScoreTab::queueInitScoreView(const int idx, const qreal logicalZoomLevel, const ZoomIndex zoomIndex, const qreal xoffset, const qreal yoffset)
+      {
+      scoreViewsToInit[idx] = ScoreViewState(logicalZoomLevel, zoomIndex, xoffset, yoffset);
+      }
+
+}
--- a/mscore/scoreview.h
+++ b/mscore/scoreview.h
@@ -130,6 +130,26 @@ enum class ViewState {
       };
 
 //---------------------------------------------------------
+//   ScoreViewState
+//---------------------------------------------------------
+
+struct ScoreViewState {
+      qreal logicalZoomLevel = 1.0;
+      ZoomIndex zoomIndex = ZoomIndex::ZOOM_FREE;
+      qreal xOffset = 0.0;
+      qreal yOffset = 0.0;
+
+      ScoreViewState() {}
+      ScoreViewState(qreal zl, ZoomIndex zi, qreal x, qreal y)
+            {
+            logicalZoomLevel = zl;
+            zoomIndex = zi;
+            xOffset = x;
+            yOffset = y;
+            }
+};
+
+//---------------------------------------------------------
 //   ScoreView
 //---------------------------------------------------------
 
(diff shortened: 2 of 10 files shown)


mscore/startcenter.cpp
```diff
@@ -124,8 +124,8 @@ void Startcenter::loadScore(QString s)
             newScore();
             }
       else {
-            mscore->openScore(s);
```
vpereverzev, line 127: Not sure about this change, what is the intent here?

oktophonie:
Just been testing this. Let's say I have a session with three scores open: one 3.6 score, and two from 3.5 where I've chosen "keep old style". These are open in three tabs which I'll call (from left to right) 3.6, 3.5a and 3.5b. I have the 'Continue last session' option chosen in Edit -> Preferences -> Program start.

If I close and re-open the session, the three scores open again and whichever was the active/visible tab when I closed the session remains the active tab. That seems like logical behaviour, fine. However:

- If 3.6 is the active tab, I still get the score migration dialog for 3.5a; surely it shouldn't appear until I switch to that tab?
- If 3.5a is the active tab, I get the score migration dialog for 3.5a; that makes sense. If I switch to 3.5b, I am then shown the migration dialog for that score, which also makes sense
- If 3.5b is the active tab, I get _two_ score migration dialogs, one for 3.5b (on top) and one for 3.5a (underneath).

So in short, it seems like if a 3.6 score is active, we get the dialog for the first 3.5 score regardless; and if a 3.5 score is active, we get the dialog for that score, but also one for the first (leftmost) other 3.5 score if there is one open.

In even shorter: the dialog for the first 3.5 score in a session always opens, in many cases when we don't yet want it to.

mscore/startcenter.cpp
cbjeukendrup: I thought, the start center would be seen as a modal dialog, so if it is still open when openScore is called, the tab won't be switched, while actually it should be. So, first close the start center, then open the score, and the tab will be switched.

cbjeukendrup:
@oktophonie Thanks! I now also notice this. It seems that the "Restore session" logic somehow circumvents the checks I added in openScore, so I will have a second look at that.

vpereverzev:
I think we've an easier option to resolve this issue. All we need is to handle switchTab variable properly. Generally speaking, "switchTab" should be "true" only if we're:
- case 1: trying to open a single score (our default behaviour already)
- case 2: trying to open list of scores, so it'll be true only for the last score in this list

Considering, that there are only a couple of places where we need to open a LIST of scores, I'd suggest to create a new method "openScores(const QStringList& pathList)" where we'll cover the "case 2"

cbjeukendrup:
@vpereverzev But I'm afraid it is not that simple. What if the user opens, say, Score A, and immediately after that Score B, without interacting with MuseScore. That should be the same as opening a list of scores, but I think it would be pretty difficult to 'recognize' that. So I think that the idea of this PR is okay, but that the "Restore session" logic isn't perfect yet. I saw that `loadScores(const QStringList& argv)` still uses readScore instead of openScore, so that might be the problem. I'm looking at that now.

mscore/scoreview.h
```diff
@@ -129,6 +129,17 @@ enum class ViewState {
       FOTO_LASSO,
       };
 
+//---------------------------------------------------------
+//   ScoreViewState
+//---------------------------------------------------------
+
+struct ScoreViewState {
+      qreal logicalZoomLevel;
```
vpereverzev, line 137: it's a bit safer to pre-initialize values here

mscore/scoretab.cpp
```diff
@@ -555,26 +557,26 @@ void ScoreTab::removeTab(int idx, bool noCurrentChangedSignal)
...
-                  delete v;
-                  return;
+      auto it = scoreViewsToInit.find(idx);
+      if (it != scoreViewsToInit.end()) {
+            ScoreViewState s = it->second;
+            if (ScoreView* v = view(idx)) {
+                  v->setLogicalZoom(s.zoomIndex, s.logicalZoomLevel);
+                  v->setOffset(s.xOffset, s.yOffset);
                   }
-            QSplitter* vs = new QSplitter;
```
vpereverzev, line 570: mm, what about this part of code? Why has it gone entirely?

mscore/scoretab.cpp
```diff
@@ -555,26 +557,26 @@ void ScoreTab::removeTab(int idx, bool noCurrentChangedSignal)
...
 
-void ScoreTab::initScoreView(const int idx, const qreal logicalZoomLevel, const ZoomIndex zoomIndex, const qreal xoffset, const qreal yoffset)
+void ScoreTab::initScoreView(int idx)
       {
-      ScoreView* v = view(idx);
-      if (!v)  {
-            v = new ScoreView;
-            Score* sc = scoreList->value(idx);
-            if( sc != 0 )
-                  v->setScore(sc);
```
vpereverzev, line 565: two notes here:
- I see like previously we were creating a new ScoreView if we couldn't get it from view(idx)
- Also I can't find a place where we're setting a Score for ScoreView

I assume it's happening somewhere else, but it's not really obvious now how it works

cbjeukendrup: I believe it works like this:
- The tab is added in `restoreSession` / `openScore` / `appendScore` (but the score view not yet initialized)
- When the user switches to a tab, via signals/slots `ScoreTab::setCurrent` gets called, where the ScoreView is initialized and the score is set. 
- This function, `ScoreTab::initScoreView` is called at the bottom of `ScoreTab::setCurrent`, where we have guarantee that the scoreView is initialized.

vpereverzev:
Peek 2021-01-26 20-43

meh, there is something wrong with ScoreView titles when I'm trying to restore last session. They are becoming valid only after a score update.

mscore/scoreview.h
cbjeukendrup: Fixed!

cbjeukendrup:
> meh, there is something wrong with ScoreView titles when I'm trying to restore last session. They are becoming valid only after a score update.

Hm, I will have another look...

vpereverzev:
> > meh, there is something wrong with ScoreView titles when I'm trying to restore last session. They are becoming valid only after a score update.
> 
> Hm, I am having another look...

One more note - looks like it happens when you've newly created scores in your session

cbjeukendrup:
@vpereverzev It should be fixed now!

PR #7265, opened 2021-01-19
Link to Bilibili video if using Chinese

Howard-C:


Howard-C:
If you don't know about what's going on: I've been speaking to @Tantacrul about changing the links of the banners on musescore.org and musescore.com to Bilibili videos if the user is in China mainland, and the musescore.org link has already been changed. Bilibili is the largest video platform in China mainland, where we usually cannot view YouTube. So wouldn't hurt to change the link in MuseScore itself too.

PR #6848, opened 2020-11-10
[MU4] audioengine

psmokotnin:
New audio engine.

- Full refactoring of framework/audio

- framework/rpc module was added

- SoLoud was removed.

- dr_wav, dr_mp3 and vorbis_std was updated and moved to thirdparty folder

AudioEngine2

inhouse team

PR #7266, opened 2021-01-19
[MU4] Audio. Step 1. Rework audio rpc system

igorkorsukov:
What is the problem?
Calls to methods of an instance of the Sequencer class are made from different threads. This is very dangerous, such a solution is very fragile, not reliable, and often contains many potential crashes or deadlocks due to the thread-unsafe code.

This solution fixes threading problems. 
Now an instance of the Sequencer class works in only one thread.

List of potential issues found in general during the review: 
* Crash on play/stop - https://github.com/musescore/MuseScore/pull/6848#discussion_r528023079
* Crash on change status - https://github.com/musescore/MuseScore/pull/6848#discussion_r528036364
* Crash on play note - https://github.com/musescore/MuseScore/pull/6848#discussion_r528173460
* Crash on play note introduced after fix prev crash - https://github.com/musescore/MuseScore/pull/6848#discussion_r544112474
* Crash with use loop variable - https://github.com/musescore/MuseScore/pull/6848#discussion_r528029073
* Deadlock with use loop variable introduced after fix prev crash - https://github.com/musescore/MuseScore/pull/6848/files#r560118419
* Crash on player->stop() - https://github.com/musescore/MuseScore/pull/6848#discussion_r558445611

Reasons for reworking the RPC system:
* Interaction should be in both directions, not just one.
* It is necessary to transfer not only simple types but also objects as arguments. In this case, objects can be serialized or not, depends on the platform and the implementation of the RPC system for this platform. (https://github.com/musescore/MuseScore/pull/6848#discussion_r559679858)
* It is necessary to ensure the operation of notifications from targets (channels, notifications) (https://github.com/musescore/MuseScore/pull/6848#discussion_r559687099)

Also in the previous RPC system, there was no explicit RPC controller, and its responsibility was spread over several classes:
* RPCServer - has mapping message -> target
* IRPCTarget - has mapping message -> method
* Sequencer(and Mixer) - has mapping message args -> target args

Now an explicitly added RPC controller, which implements all three responsibilities.


src/framework/audio/internal/audioengine.cpp
```diff
@@ -151,7 +136,10 @@ void AudioEngine::setBuffer(IAudioBufferPtr buffer)
 {
     m_buffer = buffer;
     m_buffer->setSource(m_mixer);
-    m_worker->setAudioBuffer(m_buffer);
+
+    //! TODO It doesn't look obvious.
+    //! Requires detailed research
+    //m_worker->setAudioBuffer(m_buffer);
```
vpereverzev, line 142: I see like we're initializing audio buffer in modulesetup. So I'm wondering how often we'll need to call AudioEngine::setBuffer. For me it seems like we'll need it just to setup it at the very beginning (at the initialization step) and that's all. Am I missing something?

src/framework/audio/internal/midiplayer.cpp
```diff
@@ -142,9 +142,18 @@ void MIDIPlayer::forwardTime(unsigned long miliseconds)
         toTick = maxValidTick;
     }
 
-    while (m_streamState.requested) {
-        //wait NotationPlayback send data
+    //! TODO Research in more detail whether we can simply ignore, or we  need to wait,
```
vpereverzev, line 145: not sure about the problem here, let's discuss it directly

src/framework/audio/isequencer.h
```diff
@@ -47,22 +46,21 @@ class ISequencer : MODULE_EXPORT_INTERFACE
...
     virtual Status status() const = 0;
     virtual async::Channel<Status> statusChanged() const = 0;
-    virtual async::Channel<audio_track_t> audioTrackAdded() const = 0;
 
     virtual void initMIDITrack(track_id id) = 0;
     virtual void initAudioTrack(track_id id) = 0;
     virtual void setMIDITrack(track_id id, const std::shared_ptr<midi::MidiStream>& stream) = 0;
     virtual void setAudioTrack(track_id id, const std::shared_ptr<audio::IAudioStream>& stream) = 0;
 
-    virtual bool play() = 0;
```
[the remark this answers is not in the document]
igorkorsukov: There is no easy way to return a boolean (or Ret) from a method unless it is called in the caller's thread, which I avoided in this PR.
that is, the caller (PlaybackController) is in the main thread.
and we need to call this method in another thread, in the audio worker thread.

There are three options for solving the problem:
1. Adding the ability to block RPC calls (like `QMetaObject::invokeMethod` with `Qt::BlockingQueuedConnection`) - is not easy.
2. Adding a single error signal for Sequencer - is easy and can be convenient.
3. Adding a return promise from the method, and the promise will report success or failure - pretty easy.

I will think about how to handle errors later.

In the origin PR, this is `bool` was never used.

src/framework/audio/isequencer.h
```diff
@@ -47,22 +46,21 @@ class ISequencer : MODULE_EXPORT_INTERFACE
...
     virtual void stop() = 0;
-    virtual void seek(unsigned long sec) = 0;
+    virtual void seek(uint64_t sec) = 0;
     virtual void rewind() = 0;
-    virtual void setLoop(unsigned int from, unsigned int to) = 0;
+    virtual void setLoop(uint64_t from, uint64_t to) = 0;
     virtual void unsetLoop() = 0;
 
-    virtual async::Channel<mu::midi::tick_t> midiTickPlayed(track_id id) const = 0;
+    virtual async::Channel<midi::tick_t> midiTickPlayed(track_id id) const = 0;
```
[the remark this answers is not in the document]
igorkorsukov: It used in the PlaybackController for move play cursor 
```
    switch (m_cursorType) {
    case SMOOTH:
        sequencer()->positionChanged().onNotify(this, [this]() {
            if (m_notation) {
                auto seconds = sequencer()->playbackPosition();
                auto ticks = m_notation->playback()->secToTick(seconds);
                m_tickPlayed.send(ticks);
            }
        });
        break;

    case STEPPED:
        sequencer()->midiTickPlayed(MIDI_TRACK).onReceive(this, [this](midi::tick_t tick) {
            m_tickPlayed.send(tick);
```
https://github.com/musescore/MuseScore/pull/6848/files#diff-e15947d1f0fec491c6db314864cf4aa785c75c070bad012d37baa82f01ac09a4R58

src/framework/audio/internal/audioengine.cpp
igorkorsukov: In generally you right. 
Only maybe we can set buffer for write to file or any other destination, but I'm not sure at the moment.

src/framework/audio/internal/sequencer.h
```diff
@@ -26,23 +26,22 @@
...
 
-    bool play() override;
+    void play() override;
     void pause() override;
     void stop() override;
-    void seek(unsigned long miliseconds) override;
+    void seek(uint64_t miliseconds) override;
     void rewind() override;
-    void setLoop(unsigned int from, unsigned int to) override;
+    void setLoop(uint64_t from, uint64_t to) override;
```
RomanPudashkin, line 44: Better to add some unit to arguments names (like fromMilisecond, toMilisecond)

src/framework/audio/internal/rpc/rpccontroller.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    if (calls.empty()) {
+        // Bind to sequencer
+
+        bindMethod(calls, "initMIDITrack", [this](const Args& args) {
+            sequencer()->initMIDITrack(args.arg<ISequencer::track_id>(0));
+        });
+
+        bindMethod(calls, "initAudioTrack", [this](const Args& args) {
+            sequencer()->initMIDITrack(args.arg<ISequencer::track_id>(0));
```
Eism, line 73: maybe initAudioTrack instead of initMIDITrack?

src/framework/audio/internal/audiothread.h
```diff
@@ -45,9 +44,13 @@ class AudioThread
     //! use if you don't want to use internal thread
     void loopBody();
 
+    std::shared_ptr<rpc::QueuedRpcChannel> channel() const;
```
RomanPudashkin, line 47: Better to add an alias to this:

using QueuedRpcChannelPtr = std::shared_ptr<rpc::QueuedRpcChannel>

src/framework/audio/internal/rpc/rpctypes.h
```diff
@@ -0,0 +1,144 @@
...
+
+using Method = std::string;
+
+//! NOTE Depending on the implementation of the RPC channel,
+//! the data may or may not need to be serialized.
+//! For example, if we communicate with Web Worker or a separate process,
+//! then in this case it is impossible to transfer objects, structures, pointers and etc - the data must be serialized.
+//! If we communicate between threads in the same application, then the data does not need to be serialized.
+
+class Args
```
Eism, line 51: it looks like ActionData :)
can we combine and put it in the global module?

src/framework/audio/internal/rpc/queuedrpcchannel.h
```diff
@@ -0,0 +1,68 @@
...
+    bool isWorkerThread() const;
+    void setupWorkerThread(); //! NOTE Must called from worker thread
+
+    void setupMainThread(); //! NOTE Must called from main thread
+
+    void process();
+
+private:
+
+    using MQ = std::queue<Msg>;
```
RomanPudashkin, line 50: Better to rename to MessageQueue

src/framework/audio/internal/sequencer.h
igorkorsukov: renamed

src/framework/audio/internal/rpc/rpccontroller.cpp
igorkorsukov: sure, fixed, thanks!!

src/framework/audio/internal/audiothread.h
igorkorsukov: added

src/framework/audio/internal/rpc/rpctypes.h
igorkorsukov: I'm not sure, yes, the current implementation same, but these are different types and may be changed by new requirements.

src/framework/audio/internal/rpc/queuedrpcchannel.h
igorkorsukov: I would prefer to leave it this way, it is a private type, used only in a couple of places inside the internal implementation of the class.

PR #7272, opened 2021-01-20
[MU4] Audio. Step 2. Fixed audio classes inheritance 

igorkorsukov:
I and the MSVC compiler are not fans of duplicate inheritance.
MSVC compiler output a lot of warnings related to duplicate inheritance, like `inherits ... via dominance ...`
(see https://github.com/musescore/MuseScore/pull/6848#issuecomment-762827401) 

**AudioPlayer**
*Before* 
```
class AudioPlayer : public AbstractPlayer, public AbstractAudioSource, public IAudioPlayer
{

class IAudioPlayer : public virtual IPlayer, public virtual IAudioSource
{

class AbstractPlayer : public virtual IPlayer
{

class AbstractAudioSource : public virtual IAudioSource
{
```
image

*After*
```
class IAudioPlayer : public virtual IPlayer
{
...
    virtual IAudioSource* audioSource() const = 0; 

class AudioPlayer : public IAudioPlayer, public AbstractAudioSource
{
    IAudioSource* audioSource() const { return this; }
```

image

Relationships have changed.
Before that, it was "the audio player is the audio source"
Now "audio player returns audio source"
(by the way, the midi player "is not and was not an audio source")

**MIDIPlayer**
*Before* 
```
class MIDIPlayer : public AbstractPlayer, public IMIDIPlayer, ...
{

class IMIDIPlayer : public virtual IPlayer
{

class AbstractPlayer : public virtual IPlayer
{
```
image

*After*
```
class MIDIPlayer : public IMIDIPlayer, ...
{

class IMIDIPlayer : public IPlayer
{
```
image

Perhaps a more complex one could be done, but I just decided to get rid of `AbstractPlayer`. It is simple and contains almost no logic, I decided that it is an overengine, it only complicates inheritance.

**Mixer**
*Before*
```
class Mixer : public IMixer, public AbstractAudioSource
{

class IMixer : public virtual IAudioSource
{
```
image

*After*
```
class Mixer : public IMixer, public AbstractAudioSource
{

class IMixer 
{
...
     virtual IAudioSource mixedSource() = 0;
```
image

Relationships have changed.
Before that, it was "the mixer is the audio source"
Now "mixer returns mixed audio source"


Eism (approved):

src/framework/audio/iaudioplayer.h
```diff
@@ -27,13 +27,15 @@
...
-class IAudioPlayer : public virtual IPlayer, public virtual IAudioSource
+class IAudioPlayer : public IPlayer
 {
 public:
     ~IAudioPlayer() = default;
 
     virtual void unload() = 0;
     virtual Ret load(const std::shared_ptr<audio::IAudioStream>& stream) = 0;
+
+    virtual std::shared_ptr<IAudioSource> audioSource() = 0;
```
RomanPudashkin, line 38: Please, add an alias to this pointer:

using IAudioSourcePtr = std::shared_ptr<IAudioSource>

And add const to this method

src/framework/audio/imixer.h
```diff
@@ -58,14 +55,17 @@ class IMixer : virtual public IAudioSource
     virtual unsigned int addChannel(std::shared_ptr<IAudioSource> source) = 0;
     virtual void removeChannel(unsigned int channelId) = 0;
 
+    //! mixed source
+    virtual std::shared_ptr<IAudioSource> mixedSource() = 0;
```
RomanPudashkin, line 59: Please, add const to this method

src/framework/audio/iaudioplayer.h
igorkorsukov: > using IAudioSourcePtr = std::shared_ptr
I'm not a fan of such aliases, but ok, added

> And add const to this method
unfortunately impossible, does not compile

src/framework/audio/imixer.h
igorkorsukov: unfortunately impossible, does not compile

PR #7274, opened 2021-01-20
[MU4] Audio. Step 3. Fixed different issues in the audio module

igorkorsukov:
Looked through the uncorrected and unanswered comments in https://github.com/musescore/MuseScore/pull/6848 and fixed a lot.

### DIFF
--- a/src/framework/audio/isequencer.h
+++ b/src/framework/audio/isequencer.h
@@ -34,10 +34,10 @@ class ISequencer : MODULE_EXPORT_INTERFACE
 public:
     virtual ~ISequencer() = default;
 
-    using midi_track_t = std::shared_ptr<IMIDIPlayer>;
-    using audio_track_t = std::shared_ptr<IAudioPlayer>;
-    using track_t = std::shared_ptr<IPlayer>;
-    using track_id = unsigned int;
+    using MidiTrack = std::shared_ptr<IMIDIPlayer>;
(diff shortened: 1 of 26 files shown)


src/framework/audio/isequencer.h
```diff
@@ -34,10 +34,10 @@ class ISequencer : MODULE_EXPORT_INTERFACE
 public:
     virtual ~ISequencer() = default;
 
-    using midi_track_t = std::shared_ptr<IMIDIPlayer>;
-    using audio_track_t = std::shared_ptr<IAudioPlayer>;
-    using track_t = std::shared_ptr<IPlayer>;
-    using track_id = unsigned int;
+    using MidiTrack = std::shared_ptr<IMIDIPlayer>;
+    using AudioTrack = std::shared_ptr<IAudioPlayer>;
```
RomanPudashkin, line 38: I think, better to rename those aliases. They should have the same name as interfaces:

MidiTrack -> IMidiPlayerPtr
AudioTrack -> IAudioPlayerPtr
etc

igorkorsukov: Not sure.
These types are not aliases to `std :: shared_ptr <IMIDIPlayer>`, they are types belonging to the mixer, that is, we may decide in the future to do so
```
using MidiTrack = std::shared_ptr<MyMidiTrack>;
```
for the mixer, it will still be `MidiTrack`, but it will no longer be `IMidiPlayerPtr`

PR #7276, opened 2021-01-20
[MU4] Stubs system

Eism:
Some modules may be missing in the project. For example, someone doesn't need a network module.
If the developer decides to get rid of the module or just wants to save build time and not build a module that he does not need, then all the dependencies that are given outside the module will not be available.
Then other modules that use this dependency may crash because module dependency not exported. Then the developer will have to add another nullptr checker to all dependency uses. If not, then this is a potential place to crash, and the code just becomes ugly

We decided to add a stub module for each module, which will register stubs instead of implementation.
The stab module repeats the structure of the original module.

Example for network module:
Exports INetworkManagerCreator(see NetworkModule::registerExports()).

To work correctly without the network module (the exported interfaces and types must be in the path of the original module), we must do:
1. create a module along with the path corresponding to the location of the original module
src/framework/network -> src/stubs/framework/network
The standard CMakeLists.txt template, a simple NetworkStubModule class with an overridden registerExports method, in which we register a stub instead of a real class. The module name is better call as module_stub
2. create stubs for the exported interfaces
Stubs should be as simple as possible, ideally, they should return default values of each value
3. register the BUILD_NETWORK_MODULE compilation option in
/CMakeLists.txt
/build/config.h.in
4. replace the stub module in main.cpp with the original module if the module build option is disabled

### DIFF
--- a/src/stubs/framework/network/networkmanagerstub.cpp
+++ b/src/stubs/framework/network/networkmanagerstub.cpp
@@ -1,0 +1,55 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2020 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#include "networkmanagerstub.h"
+
+using namespace mu::network;
+
+mu::Ret NetworkManagerStub::get(const QUrl&, mu::system::IODevice*)
+{
+    return make_ret(Ret::Code::NotSupported);
+}
+
+mu::Ret NetworkManagerStub::head(const QUrl&)
+{
+    return make_ret(Ret::Code::NotSupported);
+}
+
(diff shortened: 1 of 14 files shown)


RomanPudashkin (approved):

src/stubs/framework/network/networkmanagerstub.cpp
```diff
@@ -0,0 +1,55 @@
...
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#include "networkmanagerstub.h"
+
+using namespace mu::network;
+
+mu::Ret NetworkManagerStub::get(const QUrl&, mu::system::IODevice*)
+{
+    return make_ret(Ret::Code::UnknownError);
```
igorkorsukov, line 25: I think it better to return  `Ret::Code::NotSupported`

src/stubs/framework/network/networkmodule.cpp
```diff
@@ -0,0 +1,36 @@
...
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#include "networkmodule.h"
```
igorkorsukov, line 19: I think better to rename the file too

src/stubs/framework/network/networkmanagerstub.cpp
Eism: thanks, fixed

src/stubs/framework/network/networkmodule.cpp
Eism: thanks, fixed

igorkorsukov (approved):

PR #7282, opened 2021-01-20
[MU4] Add option to open popup upwards

cbjeukendrup:
- DevTools tab improvements: 
  - Select "UI Gallery" when you first open DevTools tab (it is the topmost tab, but the "Interactive" tab would be selected)
  - Same nice-looking sidebar as in home tab (not strictly necessary, but still nice)

- Option to open popup upwards
  Schermafbeelding 2021-01-20 om 23 41 48
  Might be needed for example if we want a popup in the status bar. 
  (I would like to implement a zoom popup in the status bar, but the popup won't grow higher than the status bar, which makes it useless. Like this: 
  Schermafbeelding 2021-01-20 om 23 38 32

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopup.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopup.qml
@@ -28,16 +28,22 @@ Popup {
     property color borderColor: ui.theme.strokeColor
     property color fillColor: ui.theme.backgroundPrimaryColor
     property bool isOpened: false
+    property bool opensUpward: false
 
     property var arrowX: width / 2
     property alias arrowHeight: arrow.height
 
     readonly property int borderWidth: 1
 
-    topPadding: bottomPadding + arrow.height
-    bottomPadding: 12
-    leftPadding: bottomPadding
-    rightPadding: bottomPadding
+    property Item anchorItem: null
+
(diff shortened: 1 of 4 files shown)


RomanPudashkin:
That is a great improvement! Thank you very much! We already have some cases when we should use this popup (with the arrow at the bottom) and it really helps

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopup.qml
```diff
@@ -88,9 +103,10 @@ Popup {
                 border { width: root.borderWidth; color: root.borderColor }
 
                 anchors {
-                    top: arrow.bottom
-                    topMargin: -1
-                    bottom: parent.bottom
+                    top: opensUpward ? parent.top : arrow.bottom
+                    topMargin: opensUpward ? undefined : -1
+                    bottom: opensUpward ? arrow.top : parent.bottom
+                    bottomMargin: opensUpward ? -1 : undefined
```
RomanPudashkin, line 109: Better to use 0 instead of undefined

cbjeukendrup: Fixed!

RomanPudashkin (approved):

PR #7284, opened 2021-01-21
[MU4] Extensions stubs

Eism:


src/stubs/extensions/qml/MuseScore/Extensions/ExtensionsPage.qml
```diff
@@ -0,0 +1,7 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+Item {
+    property string search: ""
+    property string backgroundColor: ""
+}
```
igorkorsukov, line 7: I think better to add Text like "Extensions Page Stub"

Eism: added

PR #7285, opened 2021-01-21
[MU4] Languages stubs

Eism:


src/stubs/languages/qml/MuseScore/Languages/LanguagesPage.qml
```diff
@@ -0,0 +1,6 @@
+import QtQuick 2.15
+
+Item {
+    property string search: ""
+    property string backgroundColor: ""
+}
```
igorkorsukov, line 6: I think it better to add Text like "Languages Page Stub"

Eism: added

igorkorsukov (approved):

PR #7293, opened 2021-01-22
Various fixes for 3.6.1

vpereverzev:
- https://trello.com/c/llFAEqi1
- https://trello.com/c/OxplS29w
- updates for Edwin.mss


Jojo-Schmitz:
Some fommit should mentione "Fix #315671" so that https://musescore.org/en/node/315671 gets marked fixed...

share/styles/Edwin.mss
```diff
@@ -62,6 +62,8 @@
     <staffFontFace>Edwin</staffFontFace>
     <rehearsalMarkFontFace>Edwin</rehearsalMarkFontFace>
     <rehearsalMarkFontSize>14</rehearsalMarkFontSize>
+    <rehearsalMarkFrameWidth>0.16</rehearsalMarkFrameWidth>
+    <rehearsalMarkFrameRound>0</rehearsalMarkFrameRound>
```
Jojo-Schmitz, line 66: Why? it is the new default anyhow, wouldn't it need to go to one of the pre3.6 mss files instead?

Jojo-Schmitz: That seems to cause at least one of the mtest failures

oktophonie: Edwin.mss is the subset of the new defaults that's applied if you tick the 'use Edwin' box in the score migration dialog; so it contains all settings related to updated text styles. (Or am I misunderstanding you?)

Jojo-Schmitz: OK. So it is not the one takes when "Reset all setting to default".
Still it might be good to add those former settings to the pre 3.6 mss files?

mtest/libmscore/compat206/lidemptytext-ref.mscx
```diff
@@ -241,6 +241,7 @@
         <barNoteDistance>1.2</barNoteDistance>
         <pedalPosBelow x="0" y="0"/>
         <trillPosAbove x="0" y="0"/>
+        <createMultiMeasureRests>1</createMultiMeasureRests>
```
Jojo-Schmitz, line 244: Why is this needed now? I don't see anything in the rest of this PR that might explain this

share/styles/Edwin.mss
oktophonie: It looks to me like they are already in there (legacy-style-defaults-v1.mss, legacy-style-defaults-v2.mss and legacy-style-defaults-v3.mss), with the former values (0.2 and 20).

Jojo-Schmitz: OK, then I rest my case, thanks for checking

mtest/libmscore/compat206/lidemptytext-ref.mscx
Jojo-Schmitz: It isn't needed anmore it seems...

RobFog:
For reference, the Trello issue for the Edwin changes is here: https://trello.com/c/hihxtdrm/111-fix-edwin-martin

libmscore/read302.cpp
```diff
@@ -187,7 +187,7 @@ bool Score::read(XmlReader& e)
                   else {
                         e.tracks().clear();     // ???
                         MasterScore* m = masterScore();
-                        Score* s       = new Score(m, MScore::baseStyle());
+                        Score* s       = new Score(m, m->style());
```
MarcSabatella, line 190: This seems to be what is responsible for https://musescore.org/en/node/316559.

If I understand what is going on here, we are about to read the part.  It may its own style settings we read from the file, but for everything else, we should be taking the defaults from the "regular" defaults.  Instead, here we are taking them from the parent score, leading to exactly the problem described.

This change was made, I assume, to make sure that whatever "keep old style" setting we applied to the score also make their way into the part.  Better, then, to do this directly.  That is, don't look to the parent score, but the what set of style defaults the parent score was itself drawing from.  We've got a whole bunch of functions available, I assume one of them does the trick.

MarcSabatella: Seems to me the same code used in setting style for the main score should be used here.  So, I stole/adapted this from MasterScore::read1() and it *seems* to work, replacing the line above with the following:

                        Score* s       = new Score(m, MScore::baseStyle());
                        int defaultsVersion = m->style().defaultStyleVersion();
                        s->setStyle(*MStyle::resolveStyleDefaults(defaultsVersion));
                        s->style().setDefaultStyleVersion(defaultsVersion);

Could be there is some detail I have wrong, or something more elegant.  Anyhow, this is, I think, the basic right idea.  The excerpt should have it's style defaults set in the same way the master score's style defaults were set - so the same "keep old style" settings apply.  As opposed to what is happening now, which is actually copy the master score's current style settings.

Preusmably the same sort of change would happening in read302.

PR #7305, opened 2021-01-22
fix #315841: indent can cause measures to not fit on system

MarcSabatella:
Resolves: https://musescore.org/en/node/315841

The first system indent setting does it what it says,
but it doens't actually add the space until *after*
we've collected the measures to fit on the system.
This can result in one too many measures being added to the system
(actually could be more than one if you have a crazy big indent).
The result is the measures will overflow into the margin.

This commit fixes the issue by accounting for the indent
on the first layout of the system when laying out its first measure.
This is the same time we allocate space for the instrument names,
so it makes sense to handle indent here as well.
In fact, that already happens on the subsequent system layout
that happens *after* measures are collected.
There is even code to do a "max" betwene indent and instrument names.
So it seems natural to incorporate the indent during that first layout.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -3784,7 +3784,7 @@ System* Score::collectSystem(LayoutContext& lc)
                   Measure* m = toMeasure(lc.curMeasure);
                   if (firstMeasure) {
                         layoutSystemMinWidth = minWidth;
-                        system->layoutSystem(minWidth);
+                        system->layoutSystem(minWidth, lc.firstSystem, lc.firstSystemIndent);
                         minWidth += system->leftMargin();
                         if (m->repeatStart()) {
                               Segment* s = m->findSegmentR(SegmentType::StartRepeatBarLine, Fraction(0,1));
--- a/vtest/accidental-12.mscx
+++ b/vtest/accidental-12.mscx
@@ -26,6 +26,7 @@
       <harmonyFretDist>0.5</harmonyFretDist>
       <harmonyPlay>0</harmonyPlay>
       <showMeasureNumber>0</showMeasureNumber>
+      <enableIndentationOnFirstSystem>0</enableIndentationOnFirstSystem>
       <showFooter>0</showFooter>
       <defaultFramePadding>0.5</defaultFramePadding>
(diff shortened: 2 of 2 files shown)


MarcSabatella:
I see why the one vtest failed, and it is expected, even if strange at first.  It truly is the case that accounting properly for the 5 sp indent means you can't *really* fit two measures here.  You can verify this by resetting the indent to 0 but then adding a 5 sp horizontal frame - same exact result.

So probably the vtest should be updated.  I can do that tomorrow if need be, but I would hope to have someone review the code itself.

MarcSabatella:
vtest updated to remove the indent.  I still will not match, but this is now good to be the new ref - it should show both measures as it did originally.

vtest/accidental-12.mscx
```diff
@@ -104,13 +105,14 @@
         <defaultClef>F</defaultClef>
         </Staff>
       <trackName>Piano</trackName>
-      <Instrument>
+      <Instrument id="piano">
```
njvdberg, line 108: This is correct. I just wondering why it isn't added before :open_mouth: since it should be here since the score ordering is merged.

libmscore/layout.cpp
```diff
@@ -3784,7 +3784,7 @@ System* Score::collectSystem(LayoutContext& lc)
                   Measure* m = toMeasure(lc.curMeasure);
                   if (firstMeasure) {
                         layoutSystemMinWidth = minWidth;
-                        system->layoutSystem(minWidth);
+                        system->layoutSystem(minWidth, lc.firstSystem, lc.firstSystemIndent);
```
njvdberg, line 3787: These two added parameter should be added indeed. Only in <code>layoutlinear</code> they don't make sense.

njvdberg (approved):

vtest/accidental-12.mscx
MarcSabatella: It's possible this had been saved in 3.6 before that merge.

MarcSabatella:
Just confirming that the new vtest diff is indeed correct - I remove the indent to allow both measures to fit.

PR #7306, opened 2021-01-22
fix #315861: don't show clef changes on hidden staves

MarcSabatella:
Resolves: https://musescore.org/en/node/315861

Issue is that a clef change at the end of a system
on a staff that is hidden
will still generate the clef and it will be displayed
on the top staff of the score.

The issue stems from the new isCutawayClef() function
returning true in many places where it does not need to.
Most importantly to this issue,
it's returning true for the hidden staff,
We could check for staff visibility everything this function is called,
but just checking within the function seems simpler.

Hoewever, in my simple test case added to the issue,
cutaway staves were not even invovled.
It seems this function should not be doing any of the things it does
*unless* it's a cutaway staff.
Checking for that, as well as staff visibility,
at the start of this function should help protect
against regressions that might otherwise appear
if this function doens't do what makes sense in the usual case.

### DIFF
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -2884,7 +2884,10 @@ bool Measure::isEmpty(int staffIdx) const
 
 bool Measure::isCutawayClef(int staffIdx) const
       {
-      if (!isEmpty(staffIdx))
+      if (!score()->staff(staffIdx) || !_mstaves[staffIdx])
+            return false;
+      bool empty = (score()->staff(staffIdx)->cutaway() && isEmpty(staffIdx)) || !_mstaves[staffIdx]->visible();
+      if (!empty)
             return false;
       int strack;
       int etrack;
--- a/libmscore/segment.cpp
+++ b/libmscore/segment.cpp
@@ -1165,7 +1165,7 @@ void Segment::scanElements(void* data, void (*func)(void*, Element*), bool all)
       {
       for (int track = 0; track < score()->nstaves() * VOICES; ++track) {
             int staffIdx = track/VOICES;
-            if (!all && !(score()->staff(staffIdx)->show())) {
+            if (!all && !(score()->staff(staffIdx)->show() && system() && !system()->staves()->empty() && system()->staff(staffIdx)->show())) {
                   track += VOICES - 1;
                   continue;
                   }


MarcSabatella:
Just a note that @mattmcclinch has thoughts about making a change instead in Segment::scanElements().  Also I think @elerouxx and/or @njvdberg should probably weigh in.

Some further observations:

- I believe the code as I have it here is may be having the effect of preventing the courtesy clef from even being generated on invisible/hidden staves.  If so, that's probably OK, but something to think through a bit more.
- isCutawayClef() is called multiple places, so if we do change something about the call in scanElements(), it would be good to verify the results are still consistent.
- If we elect to go with a scanElements() solution, I'd *still* recommend adding the check for cutaway() here in isCutawayClef().  There is no way this function should even be looking at segments or clefs for ordinary measures of ordinary staves - this measure is not a cutaway clef if this is not a cutaway staff.  I initially thought we should also be checking visible(), but apparently that's not actually the case.
- One other glitch in the way things work now is that the staff lines actually get drawn separately in the cases where we are thinking we are drawing a cutaway clef.  You can actually see this in the original report, the staff lines under the spurious clef are too thick.  It would also be good to make sure we never try to draw these staff lines if there are already present.

I actually do think my change here probably accomplishes what it needs to with minimal risk, but I acknowledge I don't really understand the full context of what is going on here.  That is why I would defer to what others think.  I just wanted to bring these issues up as points to consider.

mattmcclinch:
For the record, [this](https://github.com/mattmcclinch/MuseScore/commit/f20008c) is the change that I have suggested. I also made the note that we may be able to get away with simply checking `system()->staff(staffIdx)->show()` instead of (rather than in addition to) `score()->staff(staffIdx)->show()`, but in terms of risk avoidance for 3.6.1, I think it may be safest to check both.

The problem was introduced in #6461, when the check for `measure()->visible(staffIdx)` was removed from this very same line (libmscore/segment.cpp, line 1168) to allow for the drawing of clefs and barlines in some situations on measures that are otherwise invisible. The problem is that `measure()->visible(staffIdx)` checks a number of factors, one of them being the value of `system()->staff(staffIdx)->show()`, and the removal of the explicit check for the former removed the implicit check for the latter.

MarcSabatella:
After further discussion on Telegram and further testing, I have updated this PR to reflect my best understanding:

- @mattmcclinch is right that the check for whether or not this staff is visible on this system - the call to show() - belongs in scanElements
- My concern that isCutawayClef() was still working too hard - looking for clefs in measures that are not invisible and not on cutaway staves - led me to the realization that in one case, it actually wasn't working hard enough.  It missed the case of an invisible measure (set explicitly in measure properties) that happened to not be empty.  So, we'd display the courtesy clef in an invisible measure that is empty, but the clef would disappear if you tried to enter notes into it.

So, the current version fixes the original issue as described with the change in scanElements(), but also the issue I mentioned above with non-empty invisible measures via the change to isCutawayClef().  The latter change *also* prevents the function from doing anything else for "normal" measures and thus hopefully preventing further regressions in scores that don't use these features.

MarcSabatella:
Bad typo in previous version fixed, and commit message updated too.

MarcSabatella:
add nullptr check to avoid potential crash

libmscore/measure.cpp
```diff
@@ -2884,7 +2884,8 @@ bool Measure::isEmpty(int staffIdx) const
 
 bool Measure::isCutawayClef(int staffIdx) const
       {
-      if (!isEmpty(staffIdx))
+      bool empty = (score()->staff(staffIdx)->cutaway() && isEmpty(staffIdx)) || !_mstaves[staffIdx]->visible();
```
vpereverzev, line 2887: Would be perfect to have nullptr checks for score()->staff(staffIdx) and _mstaves[staffIdx]

MarcSabatella: agreed, coming soon

PR #7308, opened 2021-01-23
Fix #316071: Add option to Copy SMuFL Symbol Code for symbols in Master Palette

cbjeukendrup:
Resolves: https://musescore.org/en/node/316071

ezgif-7-6c7cc96908b9

This is an extended version of PR #7301.

### DIFF
--- a/mscore/palette.cpp
+++ b/mscore/palette.cpp
@@ -263,7 +263,7 @@ void Palette::contextMenuEvent(QContextMenuEvent* event)
       int i = idx(event->pos());
       if (i == -1) {
             // palette context menu
-            if (!_moreElements)
+            if (_isSymbolsPaletteInMasterPalette || !_moreElements)
                   return;
             QMenu menu;
             QAction* moreAction = menu.addAction(tr("More Elements…"));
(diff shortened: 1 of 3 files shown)


mscore/palette.cpp
```diff
@@ -274,6 +274,20 @@ void Palette::contextMenuEvent(QContextMenuEvent* event)
...
+                        QString symSmuflName = QString("<sym>%1</sym>").arg(regex.match(cell->name).captured(1));
+                        QApplication::clipboard()->setText(symSmuflName);
+                        }
+                  }
+            return;
+            }
+
       QMenu menu;
       QAction* deleteCellAction   = menu.addAction(tr("Delete"));
       QAction* contextAction = menu.addAction(tr("Properties…"));
```
Jojo-Schmitz, line 293: Can't we get rid of these two options in master symbols palette, where neither does anything anyway?

mscore/symboldialog.cpp
```diff
@@ -57,7 +58,7 @@ void SymbolDialog::createSymbols()
                || Sym::id2userName(id).contains(search->text(), Qt::CaseInsensitive)) {
                   Symbol* s = new Symbol(gscore);
                   s->setSym(SymId(id), f);
-                  sp->append(s, Sym::id2userName(SymId(id)));
+                  sp->append(s, qApp->translate("symUserNames", Sym::id2userName(SymId(id)).toUtf8()) + " \"<sym>" + Sym::id2name(SymId(id)) + "</sym>\"");
```
Jojo-Schmitz, line 61: Shamelessly stolen from #7301  ;-)

Jojo-Schmitz (approved):

mscore/palette.cpp
cbjeukendrup: For the symbols section, I removed them with this PR, and for the 'normal' sections, there is somehow no right-click menu at all (right click does for me the same as left click).

Jojo-Schmitz: Ah, right, I missed the `return` statement above

PR #7309, opened 2021-01-23
Fix tempo rounding issue in MusicXML export

matangover:
This is a long PR description for a seemingly simple issue :)

Since MuseScore internally stores tempo in units of 'beats per seconds' and truncates it to a fixed precision, the exported MusicXML tempo was sometimes incorrectly rounded. For example, 92 BPM in MuseScore was exported as 91.9998 and 88 BPM was exported as 88.0002.

To eliminate this error, tempo is now rounded to up to 2 decimal places when exporting (that is, up to 2 digits to the right of the decimal separator). Since the MuseScore interface itself only allows up to 2 decimal places anyway (in the Inspector), this is not expected to limit users.

Disclaimer: if files with fractional tempos are imported (with more than 2 decimal places), their tempo will be rounded to 2 decimal places on export, but this is presumably very rare (compared to the very common integer tempi which this PR fixes).

**To reproduce** the bug: Open MuseScore 3.6 and create a new score. Drag a 'quarter = 80' tempo marking to the score. Export the score to MusicXML – see the tempo is exported incorrectly (`sound tempo="79.9998"`). After applying this PR, it would be exported correctly.

**Note**: the underlying cause to this bug is actually not related to MusicXML export, it is due to the fact that tempo is stored imprecisely in mscx/mscz files. To illustrate this, try the following: Open MuseScore 3.6 and create a new score. Drag a 'quarter = 80' tempo marking to the score. In the Inspector, untick 'Follow text', and then re-tick it (this causes re-computation of the tempo internally). Export the score to MusicXML – see the tempo is exported correctly (`sound tempo="80"`). Save the score to mscx/mscz, close the score, and re-open the mscx/mscz. Export again to MusicXML – now the tempo is exported incorrectly (`sound tempo="79.9998"`).

However, this PR doesn't fix the underlying issue because I'm not sure that this fix would be desirable to you. I submitted that change in #7310. In any case, even if the other PR _is_ applied, the current PR is still needed, because of two reasons: 1) old mscx/mscz files that users have lying around, 2) the user's stored palette configurations will already have the tempo imprecisely stored.

### DIFF
--- a/importexport/musicxml/exportxml.cpp
+++ b/importexport/musicxml/exportxml.cpp
@@ -3959,7 +3959,11 @@ void ExportMusicXml::tempoText(TempoText const* const text, int staff)
 
       if (staff)
             _xml.tag("staff", staff);
-      _xml.tagE(QString("sound tempo=\"%1\"").arg(QString::number(text->tempo()*60.0)));
+      // Format tempo with maximum 2 decimal places, because in some MuseScore files tempo is stored
+      // imprecisely and this could cause rounding errors (e.g. 92 BPM would be saved as 91.9998).
+      qreal bpm = text->tempo() * 60.0;
+      qreal bpmRounded = round(bpm * 100) / 100;
+      _xml.tagE(QString("sound tempo=\"%1\"").arg(QString::number(bpmRounded)));
       _xml.etag();
       }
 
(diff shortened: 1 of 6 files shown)


vpereverzev:
I encourage you to create this PR for master, not for 3.x

matangover:
Is this the general guidance for all prs? Are they being cherry picked later?

vpereverzev:
I'm currently working on 3.6.1 patch release and it'll include only fixes for critical regressions. There will be no "good to have" changes, because otherwise we'll go into infinite loop of fixing issues.

matangover:
Ah ok, will there be further 3.x releases?

vpereverzev:
Never mind, we'll port it into master directly

PR #7310, opened 2021-01-23
Store tempo more precisely in mscx/mscz files

matangover:
Tempo in mscx/mscz files was previously stored with up to 6 significant decimal digits (the default precision for number formatting). Since the unit for tempo is 'beats per second', this caused rounding issues with
common tempo values. For example, the default palette presets 80 BPM, 92 BPM (Andante) and 116 BPM (Allegretto) would actually be saved as 79.9998 BPM, 91.9998 BPM, and 115.9998 BPM respectively. This affected playback tempo as well as exporting to audio and MusicXML.

Now tempo is stored with up to 17 significant digits (the maximum number of digits needed to uniquely represent any `double`) to eliminate the rounding error.

Note that the main change of this PR is in `tempotext.cpp`. The rest is just updated tests. The added test fails if the fix is not applied.

Note on the chosen precision: a `double` is used internally to store the tempo, and 17 decimal digits is enough to represent any value of `double` uniquely (see [documentation](https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10)). This means that no information is lost in the serialization to text.


Jojo-Schmitz:
As we only allow 2 decimals in setting them, wha nor rounding at that?

matangover:
Because they are set as BPM but stored as beats per second. So many integers then become infinite fractions.

njvdberg:
For me it is not clear which problem is solved here. If a piece is played with 79.9998 BPM instead od 80.0000 BPM, I'm not sure this is noticeable. Also for MusicXML, causes the current implementation problem when imported into a program?

matangover:
Yes, some apps round down tempo in musicxml import (e.g. Verovio currently imports 79.9998 as 79). This is a bug in those apps but nevertheless it's also a bug in musescore. 

Generally the bug is not noticeable in playback. However in extreme cases it is -- e.g. for 92bpm, after about 400k beats, musescore will be one beat behind the correct time.

In any case I don't see any reason not to fix it, since the fix is simple and has no adverse effects.

Jojo-Schmitz:
> Because they are set as BPM but stored as beats per second. So many integers then become infinite fractions.

Ah, yes, I see

igorkorsukov:
@matangover Could you please port the changes to the master

PR #7312, opened 2021-01-24
Fix #316025 - Fixed spacers do not function with vertical justification enabled

njvdberg:
Resolves: https://musescore.org/en/node/316025

When a system contains a fixed spacer, the gap above that system will become a gap with a fixed gap and no extra space will be added to this gap.
To implement this, <code>System::minDistance()</code> will now, instead of just saving the fixed distance, save the fixed spacer instead. <code>distributeStaves()</code> will use this to find out whether a system has a fixed spacer or not.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -1570,15 +1570,16 @@ static void distributeStaves(Page* page)
       VerticalGapDataList vgdl;
 
       // Find and classify all gaps between staves.
-      int ngaps { 0 };
-      qreal prevYBottom { page->tm() };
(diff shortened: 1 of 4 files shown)


MarcSabatella:
Thanks for this!  I had taken a quick look hoping there was a simple two-line fix somewhere, but soon realized it would likely mean a change to the VerticalGapData, and at that point realized I was in a little my head as far as feeling confident I'd get it right.  So I can just say, this *looks* like the right sort of fix.  Will do a test of the result and report back.

MarcSabatella:
Seems to work beautifully!

PR #7316, opened 2021-01-24
[MU4] SMuFL Symbol Anchors, ledger lines, flag placing, Shadow Note refactoring

cbjeukendrup:
- **Implement new way to get and set SMuFL Anchors for symbols**
  There is a new enum for the currently supported anchors, and the anchors are now stored in a std::map.

- **Fix ledger line length for Chord and Ambitus**
  That is: 
  a) draw with FlatCaps instead of RoundCaps
  b) append the _whole_ ledgerLineLength at both sides of the note, instead of half on each side.

- **Correctly handle note flag placing, respecting bbox**
  A MU4 version for PR #6894

- **Shadow note refactoring**
  In a way a MU4 version of PR #7064
  Give the Shadow Note more information and make better use of its inheritance from Element, so that is can be drawn much more 'realistically'.
  Also implements the corrections made to flag placement / stem height from the previous commit.

- **Fix stem length of notes with mirrored notehead**
  A MU4 version for PR #7050
  At least, the issue is fixed only after a re-layout is triggered. That could suggest two things:
  - It is broken in another place too (but very unclear where)
  - n->mirror() doesn't return the right value at first.

### DIFF
--- a/src/libmscore/element.h
+++ b/src/libmscore/element.h
@@ -382,11 +382,11 @@ public:
     int z() const;
     void setZ(int val) { _z = val; }
 
-    int staffIdx() const { return _track >> 2; }
-    void setStaffIdx(int val) { _track = val << 2 | voice(); }
+    int staffIdx() const { return _track / VOICES; }
+    void setStaffIdx(int val) { _track = val * VOICES + voice(); }
     virtual int vStaffIdx() const { return staffIdx(); }
-    int voice() const { return _track & 3; }
-    void setVoice(int v) { _track = (_track & ~3) | v; }
+    int voice() const { return _track % VOICES; }
+    void setVoice(int v) { _track = (_track / VOICES) * VOICES + v; }
     Staff* staff() const;
     const StaffType* staffType() const;
     bool onTabStaff() const;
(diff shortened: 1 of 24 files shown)


src/libmscore/element.h
```diff
@@ -383,9 +383,10 @@ class Element : public ScoreElement
     void setZ(int val) { _z = val; }
 
     int staffIdx() const { return _track >> 2; }
+    void setStaffIdx(int val) { _track = val << 2 | voice(); }
     virtual int vStaffIdx() const { return staffIdx(); }
     int voice() const { return _track & 3; }
-    void setVoice(int v) { _track = (_track / VOICES) * VOICES + v; }
+    void setVoice(int v) { _track = (_track & ~3) | v; }
```
RomanPudashkin, line 389: Better to use some named constant (as before) instead of magic number

cbjeukendrup: Fixed! (in other places too)

RomanPudashkin (approved):

igorkorsukov:
Please, do rebase

cbjeukendrup:
> Please could you do rebase

Done!

PR #7328, opened 2021-01-25
fix #315636: no space between barline and note with invisible key or time signature

MarcSabatella:
Resolves: https://musescore.org/en/node/315636

When code was added to better handle invisible key or time signatures
at the beginning of a score,
this inadvertently affected spacing between barline and first note
if there were invisible key or time signature changes later.
The spacing was never actually *correct* here -
it was controlled by the key or time signature margin -
but it more or less worked and scores depended on it.

The problem is that the changes made involved
setting the width of these invisible segments to 0
and skipping them.
Whereas previously, they got processed enough
to yield some space between the barline and first note.
The firx here is just to limit the effect of the previous change
so not *all* segments with all elements invisible,
but only those in system header measures,
or segments after the beginning of the measure.
Thus we continue to handle beginnings of scores and systems well,
and also courtesies at the end of measures.
But we revert to the previous behavior for key or time signatures
at the beginning of measures that are *not* headers.
Again, it wasn't ideal, but it basically worked,
nand now works again the same way.

I left TODO's in the code to indicate something
of what would need to happen to fix this "for real",
with the space controlled by barNoteDistance as the user expects,
rather than depending on the key or time signature margins.
A quick attempt to implement this led to problems
that convinced me to put it off until a larger scale rewrite
of the spacing algorithms as a whole.

### DIFF
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -4429,13 +4429,21 @@ void Measure::computeMinWidth(Segment* s, qreal x, bool isSystemHeader)
 
       while (s) {
             s->rxpos() = x;
-            if (!s->enabled() || !s->visible() || s->allElementsInvisible()) {
+            // skip disabled / invisible segments
+            // segments with all elements invisible are skipped,
+            // but only for headers or segments later in the measure -
+            // invisible key or time signatures at the beginning of non-header measures are treated normally here
+            // otherwise we would not allocate enough space for the first note
+            // as it is, this isn't quite right as the space will be given by key or time margins,
+            // not the bar to note distance
+            // TODO: skip these segments entirely and get the correct bar to note distance
+            if (!s->enabled() || !s->visible() || ((header() || s->rtick().isNotZero()) && s->allElementsInvisible())) {
                   s->setWidth(0);
                   s = s->next();
                   continue;
                   }
             Segment* ns = s->nextActive();
-            while (ns && ns->allElementsInvisible())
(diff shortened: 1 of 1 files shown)


njvdberg:
With this PR there is an larger distance between the first note in a measure and the left barline:
after
This distance also depends whether it is preceded by a invisible key or time signature.
Without this PR this distance is more or less the same:
after
Is this an expected result of the PR?

MarcSabatella:
Yes.  See discussion on Telegram.  My POR restored 3.5.2 behavior that people are already used and have adjusted for where needed.  The above clearly shows too little space after the barline in 3.6.  It is barely sufficient in that example, but consider what happens if there is an accidental or second:

Screenshot 2021-01-26 at 1 54 30 PM

versus with my PR:

Screenshot 2021-01-26 at 1 54 46 PM

It *is* true that there is slightly too much space in the case of the invisible time signature as compared to key signature, but that's inconsequential in comparison.  And again it's the same in 3.5.2.  That's because the space here is not barNoteDistance as it should be (but never has been), instead it's controlled by the margin settings for the key and time signatures respectively, which happen to differ slightly.

MarcSabatella:
Also, to be clear: the amount of space in 3.6 here might be similar between invisible key and time signatures, but it's definitely *way* less than it should be - less than what it would be with no key or time signature at all.  Here is 3.6:

image

and here is 3.5.2 (same as my PR):

image

PR #7337, opened 2021-01-26
Fix #316373: Show part name in timeline rather than instrument name

Jojo-Schmitz:
Show part name in timeline rather than instrument name, if that exists, but long intrument name does not

So Long Instrument Name, falling back to Part Name, falling back to Instrument name. The middle part being the change proposed here

Resolves: https://musescore.org/en/node/316373


Jojo-Schmitz:
I don't see whether, how or where this would need to go to master?

PR #7345, opened 2021-01-27
WiX QtWebEngineProcess related crash after update

jeetee:
Attempt to force overwrite files in hope that this will prevent missing files if QtWebEngineProcess is left running.

**_Current issue_**: When QWEP is running, the uninstall "old" version fails to close the process and thus also fails to remove its supporting files. At that point, the delete action becomes deferred to next reboot.
After that, the "install" step happens, which skips those files now, as they are not changed between the previous 3.x package and this one; and they are already present.
End result is that those files are removed next reboot cycle, leaving the installation crippled.

My hope is that by forcing file overwrites during install to "always", that the install step will also fail to update them and thus also defer that to the next reboot (and after their deletion).

It would be great if you could make a stable package build out of this PR so I (or you :) ) can test out this theory.

- [N/A] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)
- [N/A] I made sure the commit message title starts with "fix #424242:" if there is a related issue

### DIFF
--- a/build/packaging/WIX.template.in
+++ b/build/packaging/WIX.template.in
@@ -33,7 +33,7 @@
         <Property Id="WIXUI_INSTALLDIR" Value="INSTALL_ROOT"/>
 
         <!-- See https://docs.microsoft.com/ru-ru/windows/win32/msi/reinstallmode that explains different modes -->
-        <Property Id='REINSTALLMODE' Value='dmus'/>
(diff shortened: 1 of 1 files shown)


jeetee:
related to https://musescore.org/en/node/275837 but I was too tired to look that up yesterday

PR #7346, opened 2021-01-28
Update Petaluma to 1.065

Jojo-Schmitz:
Update Petaluma to 1.065, which got released just yesterday evening. This fixes:

* Fix [#314722](https://musescore.org/en/node/314722): Dynamics text is tiny in Petaluma font
* Fix [#315238](https://musescore.org/en/node/315238): Petaluma wings on End-Repeat barline misplaced
* Reduces size of petaluma_metadata.json by using TABs rather than multiple spaces for formatting (similar changes had been done to the other fonts recently)
* Added handwritten-style glyphs for Figured Bass and Multi-Segment Lines ranges.
* Fixed rotation of downward arpeggio swash and arrow.

Should go to master too, should be a simple cherry-pick


MarcSabatella:
Nice!  This will be one of the things that itself makes 3.6.2 worthwhile, was hoping we'd see these fixes in time.

PR #7353, opened 2021-01-28
[MU4] Improvements in toolbars and panels

RomanPudashkin:


Eism (approved):

src/appshell/view/dockwindow/dockview.cpp
```diff
@@ -56,69 +52,84 @@ void DockView::componentComplete()
...
 
-    if (_content) {
+    if (m_content) {
         QQmlEngine* engine = nullptr;
 #if QT_VERSION < QT_VERSION_CHECK(5, 12, 0)
         engine = framework::ioc()->resolve<ui::IUiEngine>("appshell")->qmlEngine();
 #else
-        engine = _content->engine();
+        engine = m_content->engine();
 #endif
```
cbjeukendrup, line 61: Is this version check still needed? I don't think so, as Qt 5.15 is our minimum.

RomanPudashkin: removed, thanks

PR #7356, opened 2021-01-28
[MU3] fix #316536: Add missing baritone-horns and euphoniums to orders.xm…

bmarwell:
…l and templates

Resolves: https://musescore.org/en/node/316536

Adds euphoniums especially to Brass Ensembles and European Concert Bands.
Just for completeness, although not too common, also to orchestral.

- [X] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)

### DIFF
--- a/share/instruments/orders.xml
+++ b/share/instruments/orders.xml
@@ -16,6 +16,8 @@
             <family>cornets</family>
             <family>flugelhorns</family>
             <family>trombones</family>
+            <family>baritone-horns</family>
+            <family>euphoniums</family>
             <family>tubas</family>
         </section>
         <section id="timpani">
@@ -278,6 +280,8 @@
             <family>cornets</family>
             <family>flugelhorns</family>
             <family>horns</family>
+            <family>baritone-horns</family>
+            <family>euphoniums</family>
             <family>trombones</family>
             <family>tubas</family>
         </section>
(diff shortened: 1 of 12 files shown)


share/instruments/orders.xml
```diff
@@ -278,6 +280,8 @@
             <family>cornets</family>
             <family>flugelhorns</family>
             <family>horns</family>
+            <family>baritone-horns</family>
+            <family>euphoniums</family>
```
njvdberg, line 284: I'm slightly surprised by the fact baritones/euphoniums are here **above** the trombones while in all other orders they are **below** the trombones. Is this order correct?

njvdberg (changes requested):

share/instruments/orders.xml
bmarwell: That is correct for brass bands, yup.

Jojo-Schmitz:
Issze and PR title is wrong, should be #316536

Jojo-Schmitz:
Also you have marked the conversations resolved but not submitted the requested changes

bmarwell:
> Commit- and PR title is wrong, should be #316536
> Also you have marked the conversations resolved but not submitted the requested changes

Sorry, something went wrong when pushing the update (which I did not see). But as I had the commit ready (and thought I pushed it), I already resolved the conversations.

Anyway, PR is updated now.

njvdberg (approved):

Jojo-Schmitz:

PR #7358, opened 2021-01-28
Bugfix/287598 fix deleted spanner range

WcRnm:
Resolves: *https://musescore.org/en/node/287598

Spanners in the measure following a range of measures were deleted during a rewrite.

This PR includes a modification to the third party IntervalTree library to allow soft limits (LT, GT) when comparing/finding elements in a range. Only one of these cases is used to fix this issue, but it seemed irresponsible not to include the full set of limit options.


WcRnm:
I fully understand not wanting to modify the third party library. I'm sure I can come up with another option, perhaps one that is much simpler as well.

WcRnm:
Closing this PR - Simpler solution is in process.

PR #7360, opened 2021-01-28
[MU3] Fix macOS dmg name, fix macOS dmg window height, update version number

cbjeukendrup:
- **Fixed macOS dmg name**
  The name of the _downloaded_ dmg was okay, but the name of the _volume_ that is mounted when you open the dmg was wrong:
  Schermafbeelding 2021-01-28 om 21 56 21
  I fixed it in such way that it will match what was intended previously. If we ever want to change it, that can be done very easily now. 

- **Fixed macOS dmg window height**
  Make the full background image visible, by taking the height of the window title bar into account.
  

- **Update version number**
  To 3.6.2. (I assumed that would be desired, or isn't it?)


igorkorsukov (approved):

Jojo-Schmitz:
Would a change equivalent to https://github.com/musescore/MuseScore/commit/0ef97b5c517a1a6564e4e2d32c4d03151d1c7645#diff-9e7722ca57674a1f6b12e1b419c1cd4918ab641dd58b60127de7427417ade066L3-R3 be needed here too?

PR #7362, opened 2021-01-28
Fix #315779: Disable auto-size of vertical frame when dragging the height handle

cbjeukendrup:
Resolves: https://musescore.org/en/node/315779

### DIFF
--- a/libmscore/box.cpp
+++ b/libmscore/box.cpp
@@ -796,6 +796,20 @@ void VBox::layout()
       }
 
 //---------------------------------------------------------
+//   startEditDrag
+//---------------------------------------------------------
+
+void VBox::startEditDrag(EditData& ed)
+      {
+      if (isAutoSizeEnabled()) {
+            setAutoSizeEnabled(false);
+            setBoxHeight(Spatium(height() / spatium()));
+      }
+      Box::startEditDrag(ed);
+      }
+
+
+//---------------------------------------------------------
 //   layout
 //---------------------------------------------------------
 
--- a/libmscore/box.h
+++ b/libmscore/box.h
@@ -153,6 +153,8 @@ class VBox : public Box {
       QVariant getProperty(Pid propertyId) const override;
       void layout() override;
 
+      void startEditDrag(EditData&) override;
+
       std::vector<QPointF> gripsPositions(const EditData&) const override;
       };
 


Jojo-Schmitz:
There a few mtests that fail, seems those insist on 10 being the height of vertical frames. I don't quite get why and how thios is related to disabling the Autimatic site on dragging that size handle?

libmscore/box.cpp
```diff
@@ -790,11 +790,24 @@ void VBox::layout()
                   contentHeight = minHeight();
 
             setHeight(contentHeight);
+            setBoxHeight(Spatium(contentHeight / spatium()));
```
Jojo-Schmitz, line 793: Is this here the cause for the mtest fails?

jeetee: Seems likely, as those now all get that default size set explicitly due to auto-size instead of keeping the old size value.

cbjeukendrup: I think I have found a way of doing it without this line. I will try that soon.

MarcSabatella:
That certainly looks a lot safer now.  Just thinking it through - what is the expected behavior if you turn autosize off for a box that has been expanded already buy this - should the box snap to the height given in the Inspector? And then let's say you add more content to the box, and now it's not big enough - should turning it on automatically snap the box to the required size?  Currently, both happen, and I think that's probably as expected.  It looks to me like this change won;t affect that, which is good, but also works as I expect with respect to dragging, where starting a drag on an autosize-expanded box disables autosize but does *not* cause the box instantly shrink.  In that sense it's very much like how Alt+drag of an autoplaced element disables autoplace but doesn't cause the position to suddenly reset.

So on paper all is good as far as I am concerned.

Once thing that has bugged me about the autosize design all along is that magic number 10 sp that basically serves as a minimum box height with autosize enabled.  I'd have preferred this to have been a style setting, assuming we could work out the compatibility implications.  FWIW, it's technically possible to implement that without much of a UI change, just add the style setting and mark the height property as styled in the boxStyle definition in box.cpp.  This should automatically result in a "set as style" button being added to the Inspector.  But, I'm not really recommending that right now - something to think about though.

libmscore/box.h
```diff
@@ -153,6 +153,8 @@ class VBox : public Box {
       QVariant getProperty(Pid propertyId) const override;
       void layout() override;
 
+      virtual void startEditDrag(EditData&) override;
```
vpereverzev, line 156: just curious, is there any reason to mark it virtual here?

cbjeukendrup:
No, not really. It was just copied from `Box`, line 54. Should I remove it?

Op 3 feb. 2021, om 18:25 heeft pereverzev+v <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:



@vpereverzev commented on this pull request.

In libmscore/box.h<https://github.com/musescore/MuseScore/pull/7362#discussion_r569601030>:

> @@ -153,6 +153,8 @@ class VBox : public Box {
       QVariant getProperty(Pid propertyId) const override;
       void layout() override;

+      virtual void startEditDrag(EditData&) override;


just curious, is there any reason to mark it virtual here?

libmscore/box.h
cbjeukendrup: No, not really. It was just copied from `Box`, line 54. Should I remove it?

vpereverzev: Yeah, it's better to remove it, because it'll ported in master soon

cbjeukendrup: Done!

vpereverzev: cool, thanks

igorkorsukov:
@cbjeukendrup Could you please port the changes to the master

cbjeukendrup:
@igorkorsukov Yep, coming soon!

PR #7363, opened 2021-01-29
fix #316571 updated close icon file reference for MIDI import panel

jeetee:
Opted for the toolbar- version over the dialog- version due to better matching size with the arrows

Resolves: https://musescore.org/en/node/316571

got overlooked in #7072 

### DIFF
--- a/mscore/importmidi_ui/importmidi_panel.ui
+++ b/mscore/importmidi_ui/importmidi_panel.ui
@@ -83,7 +83,7 @@
           </property>
           <property name="icon">
            <iconset resource="../musescore.qrc">
-            <normaloff>:/data/icons/png/window-close.png</normaloff>:/data/icons/png/window-close.png</iconset>
+            <normaloff>:/data/icons/png/toolbar-close.png</normaloff>:/data/icons/png/toolbar-close.png</iconset>
           </property>
           <property name="autoRaise">
            <bool>true</bool>


mscore/importmidi_ui/importmidi_panel.ui
```diff
@@ -83,7 +83,7 @@
           </property>
           <property name="icon">
            <iconset resource="../musescore.qrc">
-            <normaloff>:/data/icons/png/window-close.png</normaloff>:/data/icons/png/window-close.png</iconset>
+            <normaloff>:/data/icons/png/toolbar-close.png</normaloff>:/data/icons/png/toolbar-close.png</iconset>
```
cbjeukendrup, line 86: But where is `toolbar-close.png`? It doesn't seem to exist, and in the artifact the icon doesn't show up for me. I think it would be the best to take `:/data/icons/window-close.svg`. 

Futhermore, in `mscore/importmidi_ui/importmidi_panel.cpp` you need two add these lines:

```diff
+      _ui->toolButtonHideMidiPanel->setIcon(*icons[int(Icons::close_ICON)]);
+      _ui->pushButtonUp->setIcon(*icons[int(Icons::arrowUp_ICON)]);
       _ui->pushButtonDown->setIcon(*icons[int(Icons::arrowDown_ICON)]);
```

And in `mscore/data/icons/window-close.svg`:

```diff
- <svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48">
+ <svg fill="#3B3F45" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48">
```

This is needed to get it to work correctly with dark mode. However, _switching_ between light and dark mode won't work, like in very many places in the program. That would be quite complicated to solve, I think.

cbjeukendrup: (And yes, it's true that you need to hard-code the color into the svg, so that it is replaced in `MIconEnginePrivate::loadDataForModeAndState`.)

(I'm looking forward to MuseScore4's icons system, which is a lot less complicated...)

jeetee:
@cbjeukendrup will PR a better suited fix.

PR #7364, opened 2021-01-29
[MU3] Updated *.ts files

igorkorsukov:
Updated *.ts files


Jojo-Schmitz (commented):
Looks good, I can confirm that (in the Windows artifact) all translations are at their latest version.

Jojo-Schmitz:
Question though is whether this update shouldn't better be (one of) the last thing(s) done before building a release, to grab the latest available translations at that time

igorkorsukov:
> Question though is whether this update shouldn't better be (one of) the last thing(s) done before building a release, to grab the latest available translations at that time

I believe you are right. Now I'm just checking that everything works as it should, the translations are being updated.

Jojo-Schmitz:
I see, so this is just a test PR. Might still be good enough for 3,x though now, just needs a new one for 3.6.2

igorkorsukov:
> I see, so this is just a test PR. Might still be good enough for 3,x though now, just needs a new one for 3.6.2
We decided to add it now and update 3.6.1, and then update it again for 3.6.2 (there shouldn't be so many changes :))

Jojo-Schmitz:
Good, next such update should indeed be pretty small, unless our translators get really busy.

BTW: the Russian translations would need quite a bit of love ;-)  583 strings missing currently.

PR #7365, opened 2021-01-29
[MU4] Grips drawing

Eism:
https://user-images.githubusercontent.com/10116828/106258791-fc884380-6226-11eb-933e-4764b1d6ce94.mp4

### DIFF
--- a/src/notation/inotationinteraction.h
+++ b/src/notation/inotationinteraction.h
@@ -84,6 +84,12 @@ public:
     virtual async::Notification textEditingStarted() const = 0;
     virtual async::Notification textEditingChanged() const = 0;
 
+    // Grip edit
+    virtual bool isGripEditStarted() const = 0;
+    virtual bool isHitGrip(const QPointF& pos) const = 0;
+    virtual void startEditGrip(const QPointF& pos) = 0;
+    virtual void endEditGrip() = 0;
+
     virtual void splitSelectedMeasure() = 0;
     virtual void joinSelectedMeasures() = 0;
 
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -132,10 +132,9 @@ void NotationInteraction::paint(QPainter* painter)
     m_shadowNote->draw(painter);
 
     drawAnchorLines(painter);
-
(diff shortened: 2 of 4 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -548,6 +563,14 @@ void NotationInteraction::drag(const QPointF& fromPos, const QPointF& toPos, Dra
 
 void NotationInteraction::endDrag()
 {
+    if (isGripEditStarted()) {
+        m_gripEditData.element->endEditDrag(m_gripEditData);
+    } else {
+        for (auto& g : m_dragData.dragGroups) {
```
RomanPudashkin, line 569: Please, rename this variable to group

src/notation/internal/notationinteraction.cpp
```diff
@@ -1569,6 +1592,25 @@ void NotationInteraction::drawSelectionRange(QPainter* painter)
...
 
+void NotationInteraction::drawGripPoints(QPainter* painter)
+{
+    if (!selection()->element() || !m_gripEditData.element) {
+        return;
+    }
+
+    m_gripEditData.grip.resize(m_gripEditData.grips);
+
+    qreal gripWidth = 8.0 / painter->worldTransform().m11();
```
RomanPudashkin, line 1603: Please, add named constant instead of magic number

src/notation/inotationinteraction.h
```diff
@@ -84,6 +84,12 @@ class INotationInteraction
     virtual async::Notification textEditingStarted() const = 0;
     virtual async::Notification textEditingChanged() const = 0;
 
+    // Grip edit
+    virtual bool isGripEditStarted() const = 0;
+    virtual bool isClickOnGrip(const QPointF& clickPos) = 0;
```
RomanPudashkin, line 89: I think, better to rename those arguments to logicPos or just pos (as above)

RomanPudashkin: isClickOnGrip should be const

Eism: fixed

RomanPudashkin (approved):

PR #7366, opened 2021-01-29
Fix #316507: Application now supports using the system theme option i…

sidharth-anand:
Resolves: https://musescore.org/en/node/316507

This patch allows Windows users to choose the system theme option and changes the theme accordingly. (Works on versions later than Windows 10 build 1809, where dark theme was first introduced. On older builds, defaults to light theme which is the system default)

### DIFF
--- a/mscore/preferences.cpp
+++ b/mscore/preferences.cpp
@@ -25,6 +25,12 @@
 #include "macos/cocoabridge.h"
 #endif
 
+#ifdef Q_OS_WIN
+#include <Windows.h>
+#include <Winreg.h>
+#include <string>
+#endif
+
 namespace Ms {
 //! Note: see musescore.h getSharePath(). This copy is needed for build without MuseScore
 QString sharePath()
@@ -511,6 +517,19 @@ MuseScoreEffectiveStyleType Preferences::effectiveGlobalStyle() const
 #ifdef Q_OS_MAC // On Mac, follow system theme if desired
       if (s == MuseScorePreferredStyleType::FOLLOW_SYSTEM)
             return CocoaBridge::isSystemDarkTheme() ? MuseScoreEffectiveStyleType::DARK_FUSION : MuseScoreEffectiveStyleType::LIGHT_FUSION;
+#endif
+#ifdef Q_OS_WIN
+      if (s == MuseScorePreferredStyleType::FOLLOW_SYSTEM) {
+            std::wstring subKey = L"Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize";
+            std::wstring keyVal = L"AppsUseLightTheme";
+            DWORD data{};
+            DWORD datasize = sizeof(data);
+            LONG retCode = RegGetValue(HKEY_CURRENT_USER, subKey.c_str(), keyVal.c_str(), RRF_RT_REG_DWORD, nullptr, &data, &datasize);
+            if (retCode == ERROR_SUCCESS)
+                return data == 0 ? MuseScoreEffectiveStyleType::DARK_FUSION : MuseScoreEffectiveStyleType::LIGHT_FUSION;
+            else
+                return MuseScoreEffectiveStyleType::LIGHT_FUSION;
+            }
 #endif
       return MuseScoreEffectiveStyleType(s);
       }
--- a/mscore/prefsdialog.cpp
+++ b/mscore/prefsdialog.cpp
@@ -46,6 +46,34 @@
 
 namespace Ms {
 
+#ifdef Q_OS_WIN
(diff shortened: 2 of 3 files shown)


mscore/preferences.cpp
```diff
@@ -511,6 +517,20 @@ MuseScoreEffectiveStyleType Preferences::effectiveGlobalStyle() const
...
+          std::wstring subKey = L"Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize";
+          std::wstring keyVal = L"AppsUseLightTheme";
+          DWORD data{};
+          DWORD datasize = sizeof(data);
+          LONG retCode = RegGetValue(HKEY_CURRENT_USER, subKey.c_str(), keyVal.c_str(), RRF_RT_REG_DWORD, nullptr, &data, &datasize);
+          if (retCode == ERROR_SUCCESS)
+              return data == 0 ? MuseScoreEffectiveStyleType::DARK_FUSION : MuseScoreEffectiveStyleType::LIGHT_FUSION;
+          else
+              return MuseScoreEffectiveStyleType::LIGHT_FUSION;
+      }
```
Jojo-Schmitz, line 533: Nitpick: this uses the new indent style from master, not the old (banner, 6 spaces) 3.x one.

mscore/prefsdialog.cpp
```diff
@@ -82,6 +82,9 @@ PreferenceDialog::PreferenceDialog(QWidget* parent)
       if (CocoaBridge::isSystemDarkModeSupported())
             styleName->addItem(tr("System"));
 #endif
+#ifdef Q_OS_WIN
+      styleName->addItem(tr("System"));
```
Jojo-Schmitz, line 86: Note: this looks like an added string, but due to same context should reuse the existing translation, so really is not something that needs to get translated again

mscore/prefsdialog.cpp
```diff
@@ -654,6 +657,9 @@ void PreferenceDialog::updateValues(bool useDefaultValues, bool setup)
...
```
Jojo-Schmitz, line 661: dito

Jojo-Schmitz (approved):

cbjeukendrup:
This is very nice! However, will it also automatically update when MuseScore is already running and then the user switches the theme in their Windows settings?

Jojo-Schmitz:
I don#t think so, It should on next start though?

cbjeukendrup:
https://stackoverflow.com/a/41763234/11432719 might be helpful for implementing this.

RobFog:
Is it realistic this will go into 3.x?

sidharth-anand:
Nope it doesn't automatically update when MuseScore is running. It updates on the next start or if you switch to dark/light then switch back. Windows doesn't really provide a mechanism for listening to settings changes in c++ without using the native HWND window. It looks like it might be possible only for C# or UWP apps, and the link @cbjeukendrup suggests this too.

cbjeukendrup:
>Is it realistic this will go into 3.x?

In a way, this is pretty safe. It only affects the case that the user sets the theme explicitly to Follow System, which is on Windows not the default (which I think is good). So I think it could well go into 3.6.2. And I can very easily port this to 4.x. 

By the way, on Mac, it does automatically switch (see CocoaBridge.mm). 

I still have hope that it is possible to implement automatic switching on Windows, but it may be tricky to figure it out. 

Another question: isn't it a good idea to show the "Follow System" option only on versions of Windows that actually support dark theme? Something like this:
```cpp
#ifdef Q_OS_WIN
      if (windows build >= 1809)
            styleName->addItem(tr("System"));
```
(in two places in prefsdialog.cpp)

sidharth-anand:
> isn't it a good idea to show the "Follow System" option only on versions of Windows that actually support dark theme?

Getting the build number seems to be simpler than i thought it would be on Windows. I'll submit another patch doing that soon

Jojo-Schmitz:
Add it to this PR here

sidharth-anand:
Sure I'll update this PR then

vpereverzev:
I don't like the idea of fixing this for 3.x, but I admit it might be useful to port it to master. It is not known what pitfalls are hidden here.

cbjeukendrup:
FWIW, in the meantime, I already created the master version, but am still puzzling about the automatically switching, so haven't created a PR yet.

vpereverzev:
> 
> 
> FWIW, in the meantime, I already created the master version, but am still puzzling about the automatically switching, so haven't created a PR yet.
> 
> And I'm also looking into a Linux/Ubuntu version. That seems also possible, but also a bit of a puzzle.

OK, cool. Then I can freely close this PR

PR #7373, opened 2021-01-31
Fix #316441: Crash when changing time signature in front of a corrupted measure

mattmcclinch:
Resolves: https://musescore.org/en/node/316441.

Modified checkRest() so that it returns true on success or false on failure, instead of causing the program to crash if the end of the measure list is reached prematurely. This allows us to safely back out of a track list rewrite in the case of a time signature change in front of a corrupted measure.


igorkorsukov:

PR #7378, opened 2021-02-01
[MU4] Update to Bravura 1.392 and SMuFL 1.4 Beta

Jojo-Schmitz:
Let me know whether this is wanted for 3.x too, as it is still in Beta, I guess it'd rather not


Jojo-Schmitz:
SMuFL 1.4 has just been released and the part of it that had been merged with this PR, the metadata files, hasn't change since that PR.

PR #7384, opened 2021-02-02
Fix #316797: Deleting a breath/caesura selects the wrong note.

mattmcclinch:
Resolves: https://musescore.org/en/node/316797.

Since caesuras are placed after the selected note/rest, it makes sense that when the caesura is deleted, we should select the note/rest that precedes the caesura (in the same track).

### DIFF
--- a/libmscore/edit.cpp
+++ b/libmscore/edit.cpp
@@ -2499,6 +2499,15 @@ void Score::cmdDeleteSelection()
                               tick = toRest(e)->tick();
                         else if (e->isSpannerSegment())
                               tick = toSpannerSegment(e)->spanner()->tick();
+                        else if (e->isBreath()) {
+                              // we want the tick of the ChordRest that precedes the breath mark (in the same track)
+                              for (Segment* s = toBreath(e)->segment()->prev(); s; s = s->prev()) {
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz (approved):

Howard-C (approved):

libmscore/edit.cpp
```diff
@@ -2499,6 +2499,15 @@ void Score::cmdDeleteSelection()
                               tick = toRest(e)->tick();
                         else if (e->isSpannerSegment())
                               tick = toSpannerSegment(e)->spanner()->tick();
+                        else if (e->isBreath()) {
+                              // we want the tick of the ChordRest that precedes the breath mark (in the same track)
+                              for (Segment* s = toBreath(e)->segment()->prev(); s; s = s->prev()) {
+                                    if (s->isChordRestType() && s->element(e->track())) {
```
vpereverzev, line 2505: I'd prefer to also see nullptr check here

mattmcclinch: A nullptr check for what? s? That is handled by the loop condition.

igorkorsukov:

PR #7393, opened 2021-02-02
[MU4] Scroll improvements

cbjeukendrup:
- **Take horizontal scrolling into account**
Most mouses can't scroll horizontally, but some can. Now, MuseScore will also take horizontal scrolling into account.
Furthermore, the scrolling speed will now feel more comfortable and will feel the same at all zoom states and screen resolutions.

- **Update scrollbars when window is resized**

- **Fix minimum/maximum zoom**
And remove unused property.

- **Fix #312574: Always dragging with middle mouse button**
Resolves: https://musescore.org/en/node/312574
Like in many apps, the middle mouse button will now always allow dragging (_only_ dragging).
A very useful aspect of this is that you can now move the canvas using the MiddleButton, without having to leave Note Input mode. And you can now move the canvas without worrying about accidentally selecting items, as the middle mouse button doesn't select things. 

---

### DIFF
--- a/src/notation/view/notationpaintview.cpp
+++ b/src/notation/view/notationpaintview.cpp
@@ -203,10 +203,10 @@ void NotationPaintView::onViewSizeChanged()
         return;
     }
 
-    QPoint topLeft = toLogical(QPoint(0, 0));
-    QPoint bottomRight = toLogical(QPoint(width(), height()));
-    notation()->setViewSize(QSizeF(bottomRight.x() - topLeft.x(), bottomRight.y() - topLeft.y()));
+    notation()->setViewSize(viewport().size());
 
+    emit horizontalScrollChanged();
+    emit verticalScrollChanged();
     emit viewportChanged(viewport());
 }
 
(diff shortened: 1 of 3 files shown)


RomanPudashkin (approved):

IsaacWeiss:
Question for @Tantacrul: with new scroll bars and mouse wheel both in the picture, do we still need click+drag to move the view? Isn't spawning a selection rectangle the more common design pattern (as Shift+click+drag currently does in MuseScore)?

cbjeukendrup:
@IsaacWeiss I would personally prefer keeping the drag-to-move functionality. This way, each user can use the option of their choice. 

About the selection rectangle: it would be nice if users could easily toggle the behavior, maybe just with a button in the status bar. Using shift-drag or middleButton-drag should "invert" the behavior. So if you choose "select" as the default behavior, you can still quickly use "move" by using shift or middleButton. (That happens to be exactly what Dorico does.)

Tantacrul:
I would say 'no'. I think clicking to drag is fundamental for 'local'
movement (next/previous note, next/previous system, etc.). The scrollbars
are mainly for navigating between pages - often which are quite far apart.

I worked on two apps that went down the 'selection' rabbit hole. It always
causes enormous confusion and complication. It is certainly more a visual
app thing, where you need to do things like pixel selection.

If we were to change this model, it would need to be carefully prototyped
and tested.

M

Tantacrul:
We should first determine if it is a problem users are actually
encountering. If not (if they find selection to be relatively easy) then we
should not consider changing what we have now.

PR #5881, opened 2020-03-27
fix  #292652: Empty lines become smaller.

SKefalidis:
Resolves: https://musescore.org/en/node/292652

After the user is done editing a text object endEdit is called. This function unifies all previous commands (edits) into a single one so that the user can undo all changes with a single action. The problem is that UndoChangeProperty leads to a call of setXmlText, and in turn that leads to a call to createLayout(). 

In createLayout() a new line character adds a new TextBlock, but if no letters are inserted into that TextBlock (that is, if the line is empty) the TextBlock has no TextFragments (which hold information about the format, fontSize included). That leads to the TextBlock using the default font, which leads to smaller empty lines. I add an empty TextFragment to every TextBlock. I probably can improve on this, by deleting empty TextBlocks when they are not needed.

This PR also seems to take care of a problem while editing, where the cursor could "forget" it's font options and reset to the defaults. If in fact, this PR does not fix that issue, I'll upload a related PR that fixes it.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [n\a] I created the test (mtest, vtest, script test) to verify the changes I made

PR #4364, opened 2018-12-10
278068 newline in new text

tobik:
This pull request fixes the discartion of the first newline in a new text.

In addition to https://github.com/musescore/MuseScore/pull/4359 the EOL marker of the first line is false, if the text has been new inserted and edit mode is still active. In that case also the first line must be set to EOL true.

PR #7396, opened 2021-02-03
Fix #316754: Empty rehearsal mark not deleted after entering a line break

mattmcclinch:
Resolves: https://musescore.org/en/node/316754.

When #4359 attempted to fix https://musescore.org/en/node/278068 (and was shortly followed up by #4364), the end result was that when a newline is inserted into a text block (causing it to split into two text blocks), the second text block would always end with a newline, even if the original text block did not. This is fixed here by setting the second text block's EOL flag to that of the original text block.

This is what is attempted by the function that joins two text blocks when a newline is deleted, but that function wasn't getting it right either. So that has been corrected here as well.

Taken together, these two errors meant that there would be a blank line at the end of any text element that had ever contained a newline. This blank line would be small, but it would always be present, even if you tried to delete it.

Because this blank was always present in multiline text elements, and because it was small enough to not attract much attention, #5881 only added an empty text fragment before a newline, and not after. But now that we do not have to have unwanted newline characters, in order to preserve the font size on a blank line, an empty text fragment may be needed after a newline also.

### DIFF
--- a/libmscore/textbase.cpp
+++ b/libmscore/textbase.cpp
@@ -1656,6 +1656,8 @@ void TextBase::createLayout()
                         cursor.setColumn(0);
                         if (rows() <= cursor.row())
                               _layout.append(TextBlock());
+                        if (_layout[cursor.row()].fragments().size() == 0)
+                              _layout[cursor.row()].insertEmptyFragmentIfNeeded(&cursor); // an empty fragment may be needed on either side of the newline
                         }
                   else {
                         if (symState)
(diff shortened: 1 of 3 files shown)


Jojo-Schmitz:
This may fix it for more than just rehearasl marks, right?

Jojo-Schmitz (approved):

mattmcclinch:
Yes, this fixes the issue for all text elements, not just rehearsal marks. And thanks for pointing out the typo in the link, by the way.

Of course, the thing to be aware of is that text elements in existing scores that used to have a tiny trailing blank line will now have a full-sized trailing blank line. But at least now it can be deleted, whereas up until now it could not.

igorkorsukov:

PR #7400, opened 2021-02-03
Fix #315408: Mixer panel not fully shown when opened

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/315408

Before #7145, to fix https://musescore.org/en/node/286153, that setting got ignored, so it always showed the full mixer.


Jojo-Schmitz:
I don't see whether, how or where this would need to go to master?

PR #7402, opened 2021-02-03
[MU4] Ported #6479, #6880:  Automatic score ordering and bracketing

igorkorsukov:
Ported #6479, #6880:  Automatic score ordering and bracketing

### DIFF
--- a/src/libmscore/instrtemplate.cpp
+++ b/src/libmscore/instrtemplate.cpp
@@ -17,6 +17,7 @@
 #include "style.h"
 #include "sym.h"
 #include "stringdata.h"
+#include "scoreOrder.h"
 #include "utils.h"
 #include "xml.h"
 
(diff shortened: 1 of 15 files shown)


src/libmscore/instrtemplate.cpp
```diff
@@ -334,6 +352,9 @@ void InstrumentTemplate::write(XmlWriter& xml) const
             ma.write(xml);
         }
     }
+    if (!family) {
+        xml.tag("family", family->id);
+    }
```
igorkorsukov, line 357: see https://github.com/musescore/MuseScore/pull/6479#pullrequestreview-582320169 

needs sync after fix in the 3.x

njvdberg: Solved in PR #7405.

PR #7403, opened 2021-02-03
Fix #305777: Make applying tremolo a toggle operation

mattmcclinch:
Resolves: https://musescore.org/en/node/305777.

Currently, when a tremolo is added from the palette, any existing tremolo is removed and then the new one is added. With this change, the new tremolo will not be added if it is the same type of tremolo, thus allowing tremolos to be toggled via the palette. This works for one- and two-note tremolos alike.


igorkorsukov:

PR #7406, opened 2021-02-03
[MU4] New components for playback panel

RomanPudashkin:
components

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
@@ -1,0 +1,117 @@
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+import MuseScore.Ui 1.0
+
+Item {
+    id: root
+
+    property int minValue: 0
+    property int maxValue: 999
+    property int value: 0
+
+    implicitWidth: textField.contentWidth
+    implicitHeight: 30
+
+    QtObject {
+        id: privateProperties
+
+        property int maxNumberLength: root.maxValue.toString().length
+
+        function pad(value) {
+            var str = value.toString()
+
+            while (str.length < maxNumberLength) {
+                str = "0" + str
+            }
+
+            return str
+        }
+    }
+
+    TextField {
+        id: textField
+
+        anchors.centerIn: parent
+
+        text: privateProperties.pad(root.value)
+
+        onTextEdited: {
+            var currentValue = parseInt(text)
+            var str = currentValue.toString()
+
+            if (str.length > privateProperties.maxNumberLength || currentValue > root.maxValue) {
+                var lastDigit = str.charAt(str.length - 1)
+                root.value = parseInt(lastDigit)
+            } else {
+                root.value = currentValue
+            }
+
+            text = privateProperties.pad(root.value)
+        }
+
+        background: Rectangle {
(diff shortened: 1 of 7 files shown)


igorkorsukov (approved):

src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
```diff
@@ -0,0 +1,121 @@
...
+
+            while (str.length < maxNumberLength) {
+                str = "0" + str
+            }
+
+            return str
+        }
+    }
+
+    TextField {
```
vpereverzev, line 36: Isn't it better to use TextInputField here? 

It's already handling all the focus behaviours, key presses, etc.

src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            text = privateProperties.pad(root.value)
+        }
+
+        background: Rectangle {
+            id: textFieldBackground
+
+            color: "transparent"
+            opacity: 1
```
vpereverzev, line 61: I don't think it's necessary

src/framework/uicomponents/qml/MuseScore/UiComponents/TimeInputField.qml
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    StyledTextLabel {
+        anchors.verticalCenter: parent.verticalCenter
+        text: ":"
+        font: hoursField.font
+    }
+
+    NumberInputField {
```
vpereverzev, line 70: I believe we should also pay attention to validation here

like seconds value shouldn't be bigger than 60 and I don't see any validation for now, is it just me?

vpereverzev: That's why I suggest to create a new C++ validator for time  (see uicomponents/view/validators) - so you don't need this new components OR reuse IntInputValidator

src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
RomanPudashkin: I do not think it's a good idea, because the behavior of this component is very different from TextInputField. Otherwise, we will overcomplicate the TextInputField implementation

src/framework/uicomponents/qml/MuseScore/UiComponents/TimeInputField.qml
RomanPudashkin: There is a validation for the maximum value (maxValue: 60)

src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
RomanPudashkin: removed

src/framework/uicomponents/qml/MuseScore/UiComponents/NumberInputField.qml
vpereverzev: tell me briefly what is the difference

vpereverzev (approved):

PR #7407, opened 2021-02-03
Fix #316679: Segmentation Fault when opening a file with a missing section break element

mattmcclinch:
Resolves: https://musescore.org/en/node/316679.

The LayoutBreak class failed to override Element::subtype(), causing a LayoutBreak of type LINE to be considered a match for a LayoutBreak of type SECTION when determining which elements to reuse from a previous incarnation of an MMRest.


igorkorsukov:

PR #7412, opened 2021-02-04
Fix #316555: Invisible breath shouldn't impact layout.

mattmcclinch:
Resolves: https://musescore.org/en/node/316555.

Allows Segment::allElementsInvisible() to return true for breath segments if all elements in the segment are invisible.


Jojo-Schmitz (approved):

njvdberg (approved):

igorkorsukov:

PR #7423, opened 2021-02-05
[MU4] Ported #6715: Lines: Dragging a handle causes anchor to shoot off in …

igorkorsukov:
Ported #6715: Lines: Dragging a handle causes anchor to shoot off in opposite direction

### DIFF
--- a/src/libmscore/measure.h
+++ b/src/libmscore/measure.h
@@ -186,10 +186,12 @@ public:
     bool isIrregular() const { return m_timesig != _len; }
 
     int size() const { return m_segments.size(); }
-    Segment* first() const { return m_segments.first(); }
+    Ms::Segment* first() const { return m_segments.first(); }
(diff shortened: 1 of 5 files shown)


src/libmscore/measure.h
```diff
@@ -186,10 +186,12 @@ class Measure final : public MeasureBase
     bool isIrregular() const { return m_timesig != _len; }
 
     int size() const { return m_segments.size(); }
-    Segment* first() const { return m_segments.first(); }
+    Ms::Segment* first() const { return m_segments.first(); }
```
vpereverzev, line 189: It should be already in Ms namespace

igorkorsukov: I want just port, without any changes, see https://github.com/musescore/MuseScore/blob/3.x/libmscore/measure.h#L149

vpereverzev (approved):

PR #7435, opened 2021-02-05
Fix #305387: Add spinner for controlling glissando line thickness in inspector

sidharth-anand:
Resolves: https://musescore.org/en/node/305387

This patch allows the user to directly change the thickness of the glissando line from the inspector

### DIFF
--- a/mscore/inspector/inspectorGlissando.cpp
+++ b/mscore/inspector/inspectorGlissando.cpp
@@ -37,6 +37,7 @@ InspectorGlissando::InspectorGlissando(QWidget* parent)
             { Pid::FONT_FACE,       0, g.fontFace,       g.resetFontFace       },
             { Pid::FONT_SIZE,       0, g.fontSize,       g.resetFontSize       },
             { Pid::FONT_STYLE,      0, g.fontStyle,      g.resetFontStyle      },
+            { Pid::LINE_WIDTH,      0, g.lineWidthSpin,  g.resetLineWidth      },
             };
       const std::vector<InspectorPanel> ppList = {
             { g.title, g.panel }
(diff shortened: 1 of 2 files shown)


cbjeukendrup:
I tested it and noticed two things:
- If you select a glissando that is already wavy, the line thickness control is (erroneously) shown. 
However, if you select a glissando that is a line and then change it to "wavy", the line thickness control is correctly hidden (and comes back again if you change it back to "line").
- The line thickness control shows no suffix / measurement unit. It would be nice if it would say `0.15sp` instead of just `0.15`. I believe this is supported out-of-the-box in Qt Creator, so you could just look at the properties of other spinboxes that do show `sp`.

sidharth-anand:
I'll update the pr with these fixes

cbjeukendrup:
Ah, for the first one, it looks like you need to add this: 
```diff
void InspectorGlissando::setElement()
      {
      InspectorElementBase::setElement();
      if (!g.playGlissando->isChecked()) {
            g.labelGlissandoStyle->setEnabled(false);
            g.glissandoStyle->setEnabled(false);
            g.resetGlissandoStyle->setEnabled(false);
            }
      if (!g.showText->isChecked())
            g.textWidget->setVisible(false);
+     g.lineWidthWidget->setHidden(g.type->currentIndex() == 1);
      g.easeInSlider->setSliderPosition(g.easeInSpin->value());
      g.easeOutSlider->setSliderPosition(g.easeOutSpin->value());
      g.easeInOutCanvas->setEaseInOut(g.easeInSpin->value(), g.easeOutSpin->value());

      updateEvents();
      }
```

sidharth-anand:
Updated with both the fixes

Jojo-Schmitz (approved):

Jojo-Schmitz:
Needs #7704

Jojo-Schmitz:

PR #7450, opened 2021-02-08
[MU4] Implement automatic theme switching for Windows

cbjeukendrup:
Implement automatic theme switching for Windows:

Windows automatic theme switching
(Just like on macOS, MuseScore switches much more fluent to dark theme than the OS itself. That must be an advantage of QML 😊)

Note: The window title bar doesn't yet switch. That would be a nice next step for the future, but for now, this was really enough Windows for me.

It was the first time that I wrote code with multiple threads. I hope I didn't do dangerous things, and would really appreciate a review :)

Furthermore, I don't have a computer available with Windows 7, so if somebody would want to test if this PR builds and runs correctly on Windows 7 too, that would be great. 

If this PR gets merged earlier than PR #7447, that PR has to be updated, or the other way round. 

Follow-up for PR #7366 (kind of).

### DIFF
--- a/src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
+++ b/src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
@@ -1,0 +1,157 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2021 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+
+#include "windowsplatformtheme.h"
+#include "log.h"
+#include <Windows.h>
+
+using namespace mu::ui;
+using namespace mu::async;
+
+using ThemeType = IUiConfiguration::ThemeType;
+
+static const std::wstring windowsThemeKey = L"Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize";
+static const std::wstring windowsThemeSubkey = L"AppsUseLightTheme";
+
+HKEY hKey = nullptr;
+
+HANDLE hThemeChangeEvent = nullptr;
+HANDLE hStopListeningEvent = nullptr;
+
+WindowsPlatformTheme::WindowsPlatformTheme()
+{
+    using fnRtlGetNtVersionNumbers = void(WINAPI*)(LPDWORD major, LPDWORD minor, LPDWORD build);
+    auto rtlGetNtVersionNumbers
+        = reinterpret_cast<fnRtlGetNtVersionNumbers>(GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetNtVersionNumbers"));
+    if (rtlGetNtVersionNumbers) {
+        DWORD major, minor, buildNumber;
+        rtlGetNtVersionNumbers(&major, &minor, &buildNumber);
+        m_buildNumber = buildNumber & ~0xF0000000;
+    }
+}
+
+WindowsPlatformTheme::~WindowsPlatformTheme()
+{
+    stopListening();
+}
+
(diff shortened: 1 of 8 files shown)


src/framework/ui/internal/platform/windows/windowsplatformtheme.h
```diff
@@ -0,0 +1,58 @@
...
+
+    bool isDarkMode() const override;
+    async::Channel<bool> darkModeSwitched() const override;
+
+    void setAppThemeDark(bool) override;
+
+private:
+    async::Channel<bool> m_darkModeSwitched;
+    int m_buildNumber = 0;
+    bool isListening = false;
```
igorkorsukov, line 50: needs rename to `m_isListening` and better use `std::atomic<bool>` in this case

src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
```diff
@@ -0,0 +1,139 @@
...
+                     KEY_NOTIFY | KEY_CREATE_SUB_KEY | KEY_ENUMERATE_SUB_KEYS | KEY_QUERY_VALUE | KEY_WOW64_64KEY,
+                     &hKey) == ERROR_SUCCESS) {
+        std::thread listenThread([this]() { listen(); });
+        listenThread.detach();
+    } else {
+        LOGD() << "Failed opening key for listening dark theme changes.";
+    }
+}
+
+void WindowsPlatformTheme::listen()
```
igorkorsukov, line 85: better rename to `th_listen` (this is not rule, but this helpful for easy understand)

src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
```diff
@@ -0,0 +1,139 @@
...
+WindowsPlatformTheme::~WindowsPlatformTheme()
+{
+    stopListening();
+}
+
+void WindowsPlatformTheme::init()
+{
+    updateListeningStatus(configuration()->preferredThemeType());
+
+    configuration()->preferredThemeTypeChanged().onReceive(nullptr, [this](const ThemeType& newThemeType) {
```
igorkorsukov, line 55: better add in the `windowsplatformtheme.h `

```
#include "async/asyncable.h"

class WindowsPlatformTheme : public IPlatformTheme, public async::Asyncable 
```

and here 

```
configuration()->preferredThemeTypeChanged().onReceive(this, [this](const ThemeType& newThemeType) {
```

it is an automatic disconnect mechanism

cbjeukendrup:
Thank you @igorkorsukov ! I updated the PR with your feedback.

RomanPudashkin (approved):

src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+    if (m_isListening) {
+        return;
+    }
+    m_isListening = true;
+    if (RegOpenKeyEx(HKEY_CURRENT_USER, windowsThemeKey.c_str(), 0,
+                     KEY_NOTIFY | KEY_CREATE_SUB_KEY | KEY_ENUMERATE_SUB_KEYS | KEY_QUERY_VALUE | KEY_WOW64_64KEY,
+                     &hKey) == ERROR_SUCCESS) {
+        std::thread listenThread([this]() { th_listen(); });
+        listenThread.detach();
```
igorkorsukov, line 79: This moment can lead to a potential crash - if, at the time of closing the application when the WindowsPlatformTheme object is destroyed, a call to its class member (`m_darkModeSwitched`). In this case, the probability of failure is extremely small.
But in general, it is better to do this in such cases:
1. Make the thread a member of the class
2. In the `stopListening` method, do this

```
void WindowsPlatformTheme::stopListening()
{
    m_isListening = false;
    m_listenThread.join();
}
```

cbjeukendrup: But if I would put m_listenThread.join() in stopListening, then it blocks the app each time when you disable automatic theme switching, for at most 4000 ms (or whatever time we set in WaitForSingleObject). Maybe another option would be to call it only in the destructor, but I think that could theoretically still cause problems.

But, I just found a potentially very useful StackOverflow post (https://stackoverflow.com/a/6033171/11432719), so I will try that.

cbjeukendrup: My plan worked: now a "Stop" event is sent, so that `WaitForMultipleObjects` will return directly when it receives the event and the thread finishes (almost) directly, so calling `join()` won't cause waiting any more. Now it should be a bit safer!

igorkorsukov: Well done!

igorkorsukov (approved):

PR #7453, opened 2021-02-08
fix #317163: properties lost on save/reload

MarcSabatella:
Resolves: https://musescore.org/en/node/317163

The recent changes to style handling
result in an extra styleChanged() call on load,
which is not a problem in itself
but exposes some existing problems in various read() functions.
If styled properties were not marked as unstyled during the read,
they get reset by this styleChanged() call,
meaning any custom settings are lost on load.

In fact, they would mostly have been lost after a *second* save/reload
even in older versions, since the properties are not marked as unstyled
and thus won't get written.
But now they are lost immediately upon the first save/reload.

The fix is to to be sure we mark the property unstyled on read,
which basically means inserting calls to readStyledProperty()
in the read() or readProperties() functions of various element types.

This was originally reported for volta line style,
and I quickly discovered it also appplies to line width,
and to pedal line width.
I then scanned the code looking for read() functions
that did not handle styledProperties.
I can't guarantee I foudn them all,
but I did find the mmrest number offset and tuplet number text.
The latter actually *seemed* to be handling the styled properties,
but only for the tuplet itself, not the number.
The font property flags needed to be copied to the number.

### DIFF
--- a/libmscore/pedal.cpp
+++ b/libmscore/pedal.cpp
@@ -107,7 +107,10 @@ void Pedal::read(XmlReader& e)
       if (score()->mscVersion() < 301)
             e.addSpanner(e.intAttribute("id", -1), this);
       while (e.readNextStartElement()) {
-            if (!TextLineBase::readProperties(e))
+            const QStringRef& tag(e.name());
+            if (readStyledProperty(e, tag))
+                  ;
+            else if (!TextLineBase::readProperties(e))
                   e.unknown();
             }
       }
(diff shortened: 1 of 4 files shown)


Jojo-Schmitz:
counterpart got merged to master

vpereverzev:
Closed, since it's already in master

Jojo-Schmitz:
Still needed in 3.x though

Jojo-Schmitz:
Please reopen and merge into 3.x

MarcSabatella:
Several reports of this lately, FWIW.

RobFog:
Please reopen and merge.

PR #7455, opened 2021-02-09
Fix #316504: Update MusicXML Import of Fretboard Diagrams

worldwideweary:
Resolves: https://musescore.org/en/node/316504

After updating Fret Diagrams to allow for [Open markers + additional fret dots] from within MuseScore's fretboard diagram editor, MusicXML import needed updating to properly form a diagram with potentially both open markers and fret marks on the same string. There was also an issue with extraneous mute markings even with the presence of a fret-dot while being imported (they are implicitly formed during initialization rather than being explicit data like with fret markers). This takes care of issue linked above.

Aside: even though FretDiagram::init() isn't involved in importing from xml, it was erroneously setting _maxFrets rather than _frets from the corresponding "getter" frets() function, so changed that also. 

### DIFF
--- a/importexport/musicxml/importmxmlpass2.cpp
+++ b/importexport/musicxml/importmxmlpass2.cpp
@@ -4935,8 +4935,11 @@ FretDiagram* MusicXMLParserPass2::frame()
                   if (string > 0) {
                         if (fret == 0)
                               fd->setMarker(actualString, FretMarkerType::CIRCLE);
-                        else if (fret > 0)
+                        else if (fret > 0) {
+                              if (fd->marker(actualString).mtype == FretMarkerType::CROSS)
+                                    fd->setMarker(actualString, FretMarkerType::NONE);
                               fd->setDot(actualString, fret, true);
+                              }
                         }
                   else
                         _logger->logError(QString("FretDiagram::readMusicXML: illegal frame-note string %1").arg(string), &_e);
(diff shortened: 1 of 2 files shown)


RobFog:
Perhaps this feature should be added to MuseScore 4 instead of 3.x?

worldwideweary:
@RobFog, any ideas?

RobFog:
I'm not a programmer but from the description it sounds like this is a larger change. The maintainers don't seem to want to introduce new features and major changes in 3.x.

RobFog:
@jthistle What do you think?

jthistle:
This is a 25 line change that doesn't really introduce any new features, but rather makes existing features compatible with MusicXML. That is, it's a fix rather than a feature. I'm all for this PR.

jthistle:
That said, would it be possible to add an automated test (or modify an existing one) to check that this works and stays working?

lvinken:
Code reviewed and tested, all OK. The current MusicXML regression tests (read MusicXML, write MusicXML, compare) do not catch this issue, as it affects fretboard state that cannot be exported to MusicXML. Reading MusicXML and writing in MuseScore format does catch the issue.

Jojo-Schmitz:
Interestingly large parts of this, most parts of the changes to libmscore/fret.cpp, are in master already.
For the rest see #8014

PR #5848, opened 2020-03-20
fix #275483: Added the ability to change the color of the playbackCursor

SKefalidis:
Resolves: https://musescore.org/en/node/275483

It gives the user the ability to change the color of the playback cursor (can't be made fully opaque). 
Because having a fully opaque cursor does not have an obvious use if you choose a fully opaque color you get the default transparency level, so the user can change color with even fewer clicks.

To solve the original issue you make the cursor transparent and the voice color black.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [n/a] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7458, opened 2021-02-10
Fix #275483: Give [playback cursor] a distinct color preference

worldwideweary:
Resolves: https://musescore.org/en/node/275483
Reopens the closed PR: https://github.com/musescore/MuseScore/pull/5848

+ Makes a slight adjustment to the previous code using a hex value rather than string to be in conformity with the surrounding code as it now currently is.

Description: Adds a distinct color for the playback cursor: defaults to 50 alpha level if user attempts full opacity. And, as mentioned in the original PR, 100% transparency is allowed to have no cursor showing upon playback.

P.S. I consider this a bug-fix in the sense that there was a previous closed issue for specifically changing the highlight color as if it was implemented totally, but in reality it bound "Voice-1" color with "Playback cursor" color. As such, their separation could be seen as a bug fix. Point of mentioning this is that this should, in my opinion, be included into 3.6.3 if released and not sat around only for version 4 later, especially since this was available months back (the closed PR mentioned above) but not included due to a false alarm of someone else stating that they would be providing something better which never came to pass.


Jojo-Schmitz:

PR #7461, opened 2021-02-10
[MU4] Implemented missing file operations

Eism:


src/framework/uicomponents/uicomponentstypes.h
```diff
@@ -57,7 +57,8 @@ struct MenuItem : public actions::ActionItem
             { "icon", static_cast<int>(iconCode) },
             { "enabled", enabled },
             { "checked", checked },
-            { "subitems", _subitems }
+            { "subitems", _subitems },
+            { "data", data }
```
RomanPudashkin, line 61: I think, better to rename data to params or args

src/notation/imasternotation.h
```diff
@@ -41,7 +41,8 @@ class IMasterNotation
     virtual Ret createNew(const ScoreCreateOptions& scoreInfo) = 0;
     virtual RetVal<bool> created() const = 0;
 
-    virtual Ret save(const io::path& path = io::path()) = 0;
+    virtual Ret save(const io::path& path = io::path(), SaveMode saveMode = SaveMode::Unknown) = 0;
+    virtual Ret saveSelection(const io::path& path = io::path()) = 0;
```
RomanPudashkin, line 45: I think, better to implement this using one method. We can create a special enum to separate two different modes

src/userscores/internal/filescorecontroller.cpp
```diff
@@ -121,12 +127,47 @@ void FileScoreController::saveScore()
...
-    io::path selectedFilePath = selectScoreSavingFile(defaultFilePath);
+    io::path selectedFilePath = selectScoreSavingFile(defaultFilePath, qtrc("userscores", "Save Score"));
+    if (selectedFilePath.empty()) {
+        return;
+    }
+
+    doSaveScore(selectedFilePath, SaveMode::SaveAs);
+}
+
+void FileScoreController::saveScoreACopy()
```
RomanPudashkin, line 138: Better to rename to saveScoreCopy

src/userscores/internal/userscoresservice.cpp
```diff
@@ -0,0 +1,58 @@
...
+    return result;
+}
+
+std::vector<Meta> UserScoresService::parseRecentList(const QStringList& recentScoresPathList) const
+{
+    std::vector<Meta> result;
+    for (const QString& path : recentScoresPathList) {
+        RetVal<Meta> meta = msczMetaReader()->readMeta(path);
+        if (!meta.ret) {
+            LOGW() << "Score reader error" << path;
```
RomanPudashkin, line 50: Better to use LOGE in this case

src/appshell/view/appmenumodel.cpp
```diff
@@ -42,23 +42,18 @@ void AppMenuModel::load()
...
-    });
+    setupConnections();
 
     emit itemsChanged();
 }
 
-void AppMenuModel::handleAction(const QString& actionCode)
+void AppMenuModel::handleAction(const QString& actionCode, const QVariant& actionData)
 {
+    if (actionCode == "file-open") {
```
RomanPudashkin, line 52: I don't like the idea to write a logic that depends on concrete codes here. This approach has poor scalability (you should add else-if for all actions that have arguments). Better write an abstract code for handling actions:

args = action.args

if (action.args.isEmpty) {
dispatch(actionCode)
} else {
dispatch(actionCode, action.args)
}

igorkorsukov: or just `dispatch(actionCode, action.args)`

Eism: args is QVariant, not ActionData
need to restore the types of arguments, and this is possible only by code

src/framework/uicomponents/uicomponentstypes.h
Eism: thanks, renamed to args

src/notation/imasternotation.h
Eism: thanks, fixed

src/userscores/internal/filescorecontroller.cpp
Eism: thanks, renamed

src/userscores/internal/userscoresservice.cpp
Eism: fixed

src/appshell/view/appmenumodel.cpp
Eism: thanks 
I am not passing arguments to qml now. The index of the action in the group comes from qml, by which I find the action locally

src/appshell/view/appmenumodel.cpp
```diff
@@ -42,24 +43,17 @@ void AppMenuModel::load()
...
+    setupConnections();
 
     emit itemsChanged();
 }
 
-void AppMenuModel::handleAction(const QString& actionCode)
+void AppMenuModel::handleAction(const QString& actionCode, int actionIndex)
 {
-    actionsDispatcher()->dispatch(codeFromQString(actionCode));
+    ActionCode _actionCode = codeFromQString(actionCode);
```
igorkorsukov, line 53: I believe, better in such a case  use `val_` for args

igorkorsukov (approved):

src/appshell/view/appmenumodel.cpp
Eism: renamed

RomanPudashkin (approved):

PR #7464, opened 2021-02-10
[MU3] Fix #317319: Remove unneeded skyline extension for time- and key signatures

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/317319

Should go to master too, so see #7498 (merged there already)

### DIFF
--- a/libmscore/keysig.cpp
+++ b/libmscore/keysig.cpp
@@ -281,26 +281,6 @@ void KeySig::layout()
       }
 
 //---------------------------------------------------------
-//   shape
-//---------------------------------------------------------
-
-Shape KeySig::shape() const
-      {
-      QRectF box(bbox());
-      const Staff* st = staff();
-      if (st && addToSkyline()) {
-            // Extend key signature shape up and down to
-            // the first ledger line height to ensure that
-            // no notes will be too close to the keysig.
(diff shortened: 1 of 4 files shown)


Jojo-Schmitz:
The failed vtests are showing improvements, not failures actually. In general less space between time signature and 1st note (if there is no key signature) e.g. before:
accidental-23-ref
after
accidental-23-1
But also this, before:
slashed_grace-3-ref
after:
slashed_grace-3-1

MarcSabatella:
Right, and as I observed on Telegram, the overlapping grace notes can happen even without this PR, if the grace note is a little higher above the staff.  So I wouldn't worry about that with respect to this PR - to fix that there needs to be other changes to the spacing calculations.

In the other cases here, right now there is clearly *too much* space, and this PR is a definitely improvement.  Like, the first image above with the Ab and Abb, we're applying the "system header distance" to determine the distance to the first note, but inappropriately considering these fake extensions as something we need to clear, so we're measuring from there to the double flat on the higher note, rather than from the time signature itself to the lower note.

This extra space was actually noted in a Facebook thread recently but I never got around to investigating.

Anyhow, I approve these changes.

Jojo-Schmitz:
See #9000

PR #7465, opened 2021-02-11
[MU4] Play panel infrastructure

RomanPudashkin:
Various improvements and fixes needed for the play panel


src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopupView.qml
```diff
@@ -29,20 +30,31 @@ PopupView {
...
     readonly property int borderWidth: 1
 
+    function toggleOpened() {
+        if (isOpened) {
+            close()
+        } else {
+            open()
+        }
+    }
+
```
cbjeukendrup, line 46: Couldn't this be done in "PopupView.cpp"? Or has it advantages to do it in QML?

src/framework/ui/view/iconcodes.h
```diff
@@ -252,6 +253,7 @@ class IconCode
         NOTE_DOTTED_2 = 0xF395,
         NOTE_DOTTED_3 = 0xF396,
         NOTE_DOTTED_4 = 0xF397,
+        MIDI_INPUT = 0xEF98,
```
Eism, line 256: enums must be sorted by value

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopupView.qml
RomanPudashkin: Yes, you are right. Moved to cpp

src/framework/ui/view/iconcodes.h
RomanPudashkin: fixed, thanks

Eism (approved):

igorkorsukov (approved):

PR #7480, opened 2021-02-11
[MU3] fix #317368: fix Score.staves property

dmitrio95:
Adds Staff type registration that was missed in 4ea5007b76251658cee2cd6a012b7896aade682a.
Changes \since statement for Score.staves property because the property
appears to be broken in earlier versions. The other relevant property
(Element.staff) works fine in the previous versions, therefore no changes for it.

Resolves: https://musescore.org/en/node/317368


igorkorsukov (approved):

Jojo-Schmitz:
Wouldn't this be needed for master too?

dmitrio95:
Yes, it is indeed needed for master as well, I have just opened #7522 for this.

PR #7481, opened 2021-02-12
[MU3] Fix #314060: Onscreen Keyboard to show actual note events rather than only score selection during playback

worldwideweary:
…an score selection during playback

Resolves: https://musescore.org/en/node/314060

Onscreen keyboard didn't update correctly with note events (like trills, tremolos between notes, mordents etc). This in my opinion could actually be considered a major bug rather than minor depending on the stress attached to the onscreen keyboard.

I don't know if the .COM side uses the same code for the production of its SEND-TO-YOUTUBE ability, but even there it also suffers from incorrect keyboard updating. It would be cool if this fixed the .COM side also, but that depends on whatever mechanisms are being utilized, and I don't have any information on that at the moment.

As an example, here is how it looks in 3.6.2 and below:
no-ornamentarticulation

And here is how it is with these changes
with-ornamentarticulation

### DIFF
--- a/mscore/events.cpp
+++ b/mscore/events.cpp
@@ -15,6 +15,7 @@
 #include "musescore.h"
 #include "seq.h"
 #include "texttools.h"
+#include "pianotools.h"
 #include "fotomode.h"
 #include "tourhandler.h"
 #include "scoreaccessibility.h"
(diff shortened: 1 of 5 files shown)


Jojo-Schmitz:

PR #7514, opened 2021-02-12
fix #317272: page settings dialog shows wrong values in pre-3.6 parts

MarcSabatella:
Resolves: https://musescore.org/en/node/317272

When displaying the page settings dialog,
we create a preview by cloning the score,
and the dialog actually manipulates the preview's values.
However, the clone is always forced to current style defaults,
so for pre-3.6 scores that, the preview and the spinboxes show
new 3.6 defaults as opposed to the old.
This causes the preview to display completely unlike the actual score,
and it also means the default values for the margins are wrong.

A full fix for this would involve re-examining the logic used
in the clone() methods for both Score and MasterScore.
But there could be any numebr of side effects to this.
So the fix here is very conservative:
it just overrides the clone's style with the original style
after the fact.
Thus it can only affect the page settings dialog.

It should also be noted this bug only affects parts, not parent scores.
That is because the process of reading/writing the parent score
resets the style again based on version tags encountered.
So probably there is another fix here involving that logic.
This too would almost certain have side effects.

### DIFF
--- a/libmscore/score.cpp
+++ b/libmscore/score.cpp
@@ -358,6 +358,13 @@ Score* Score::clone()
       XmlWriter xml(this, &buffer);
       xml.header();
 
+      // TODO: this code to set MSC_VERSION explicitly causes the preview score in page settings dialog
+      // to always use current style defaults rather than those of the source score
+      // this clearly wrong for the page settings dialog and causes https://musescore.org/en/node/317272
+      // it's possible this code should be replaced by code that sets version based on the source score
+      // but it's also possible other code relies on the current behavior
+      // it's also possible MasterScore::clone() should contain the same change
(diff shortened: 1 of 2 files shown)


MarcSabatella:
This one I cannot see how to adapt for master - Score::clone() is entirely different, and pagesettings.cpp no longer exists.  It's very possibly the issue no longer relevant anyhow as probably we don't intend to continue to special case these imports.

But, this is definitely something that should be high priority for 3.x, so I'd recommend merging this even without a corresponding PR for master.

Jojo-Schmitz:

MarcSabatella:
Perhaps, but since it deals with dialog code that I assume has been or is being completely rewritten, I will leave it to someone else.

PR #7516, opened 2021-02-12
Fix #317424: Missing tie in part after paste

mattmcclinch:
Resolves: https://musescore.org/en/node/317424.

When a tie is created as a result of pasting a note that is too long to fit in the measure, ties must be added to the corresponding notes in any and all linked scores. This is accomplished by using the undoAddElement() function to add the tie, rather than simply using setTieFor() and setTieBack() on notes in the current score.


Jojo-Schmitz:

PR #7517, opened 2021-02-12
Fix #317425: Corruption/crash after repeating a 256th note or smaller.

mattmcclinch:
Resolves: https://musescore.org/en/node/317425.

In the MIME data for a StaffList, the starting tick and tick length are encoded as integers, which is not sufficient when dealing with durations shorter than 1/128th. But there is really no reason for this, since ticks are stored internally as Fractions, and Fractions can be converted to and from strings without any loss in precision.


Jojo-Schmitz:
mtests are not happy

mattmcclinch:
I’ll just have to update the tst_selectionfilter refs to expect the new encoding.

Jojo-Schmitz:

PR #7518, opened 2021-02-12
Fix #317426: Rests grouped incorrectly after adding 256th note or smaller

mattmcclinch:
Resolves: https://musescore.org/en/node/317426.

The functions that deal with rhythm grouping operate on integer ticks, but they are smart enough to realize that integers can only be split in half so many times. If and when this limit is exceeded, populateRhythmicList gives up trying to group rhythms based on subbeats, and simply appends the result of toDurationList() to the current duration list that is being built. But toDurationList() returns a list that is ordered from largest to smallest, so it really ought to be appended in the reverse order.


Jojo-Schmitz:

PR #7519, opened 2021-02-12
Fix #317188: Allow [Q] and [W] to work on a range selection

mattmcclinch:
Resolves: https://musescore.org/en/node/317188.


Jojo-Schmitz:

Jojo-Schmitz:
I just did it, see #24385

PR #5730, opened 2020-02-18
Timeline optimizations

dmitrio95:
As full timeline refactoring (#4542) has not been finished as for this moment, this PR tries to fix the most prominent performance issues of the old Timeline implementation. In general, the following optimizations are implemented:
1) Optimize the procedure of highlighting currently visible measures by avoiding iterating over all timeline items. This mostly affects score scrolling which could previously be noticeably slowed down on large scores if Timeline was open.
2) Implement partial updating of measure rectangles in Timeline. Top panel with metadata (tempo changes, time signatures etc.) is still fully updated on each action but the overall procedure of Timeline updating is still much faster than it was before.
3) Don't update data of invisible Timeline when loading or switching a score. This case has been missed from the previous optimizations for not updating invisible Timeline.

Although Timeline performance is still not ideal with this PR and could be optimized further, this PR should still hopefully make Timeline more usable for people working on relatively large scores. Users who don't use Timeline should also benefit from this PR due to the last optimization for invisible Timeline.

PR #7524, opened 2021-02-14
[MU3] fix #317511: fix a crash on Undo after adding measures with timeline open

dmitrio95:
Resolves: https://musescore.org/en/node/317511

The issue got introduced in #5730. This pull request fixes the issue by adding the missing null pointer checks after searching for measures by tick values found in `CmdState`.


Jojo-Schmitz:

PR #7529, opened 2021-02-15
[MU3] Reimplement spacers in vertical staves adjustment to make sure all spacers are taken into account.

njvdberg:
Solves the ignored spacer between last staff and bottom of the page.
Also it was found some combination were not handled correctly. To solve this <code>distributeStaves()</code> now implements it own code for finding the correct <code>Spacer</code> for every spacing.

A dedicated PR for <code>master</code> will be created shortly.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -1575,7 +1575,7 @@ static void distributeStaves(Page* page)
       qreal  prevYBottom  { page->tm() };
       qreal  yBottom      { 0.0        };
       bool   vbox         { false      };
-      Spacer* activeSpacer { nullptr    };
+      Spacer* nextSpacer  { nullptr    };
       bool transferNormalBracket { false };
       bool transferCurlyBracket  { false };
       for (System* system : page->systems()) {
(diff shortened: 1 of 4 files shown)


vpereverzev:
Closed, since it's already in master

Jojo-Schmitz:
Still needed for 3.x though

libmscore/layout.cpp
```diff
@@ -1614,8 +1614,32 @@ static void distributeStaves(Page* page)
                         if (!sysStaff->show())
                               continue;
 
+                        Spacer* activeSpacer { nextSpacer };
```
vpereverzev, line 1617: @njvdberg can you please exract this new logic into a new method? layout.cpp is already overcomplicated

njvdberg: I'll make it in 3.x and will create a dedicated PR for master myself (I'm used to it now :smile: )

njvdberg: Done.
See [PR #7722](https://github.com/musescore/MuseScore/pull/7722) for `master`.

For now, only `distrubuteStaves()` will use the new methods but I will replace other places for finding the active spacers by these new methods. However, this will be two separate PR's (for `3.x` and `master`).

PR #7531, opened 2021-02-15
ONE BUILD SCRIPT TO RULE THEM ALL!

shoogle:
Adds a cross-platform script to automate the process of creating a build directory and compiling MuseScore. The single file `build.cmake` will eventually replace the functionality of all these files:

- [Makefile](https://github.com/musescore/MuseScore/blob/master/Makefile)
- [Makefile.mingw](https://github.com/musescore/MuseScore/blob/master/Makefile.mingw)
- [Makefile.osx](https://github.com/musescore/MuseScore/blob/master/Makefile.mingw)
- [msvc_build.bat](https://github.com/musescore/MuseScore/blob/master/msvc_build.bat)
- [ninja_build.bat](https://github.com/musescore/MuseScore/blob/master/ninja_build.bat)
- [ninja_build.sh](https://github.com/musescore/MuseScore/blob/master/ninja_build.sh)

__Usage:__

Use this command to run the `build.cmake` script on any platform:

```bash
cmake -P build.cmake [args...]
```

Developers with a Unix-like environment (including Git Bash on Windows) can also use:

```bash
./build.cmake [args...]
```

__Build step arguments:__

```
    clean       Delete the build directory.
    configure   Create a build directory and run CMake inside it.
    build       Compile code using the native build tool.
    install     Copy compiled files to final destinations.
    run         Run the installed program.
```

Each step implies all previous steps except 'clean' and 'run', which are only performed when explictly requested. If no steps are given then the configure, build, and install steps are performed by default.

All other arguments are passed to CMake during configuration.

__Example command:__

```bash
cmake -P build.cmake -GNinja -DCMAKE_BUILD_TYPE=Debug
cmake -P build.cmake -GNinja -DCMAKE_BUILD_TYPE=Debug run -F
```
This creates and installs a Debug build using the Ninja generator, and then runs the compiled program with the -F option to perform a factory reset, thereby ensuring it is in the initial state.

__Overriding the default settings:__

If you frequently need to build with non-default settings then you can create a file build_overrides.cmake with all your personal settings to avoid having to pass them in on the command line each time.

```cmake
# build_overrides.cmake example file
set(ENV{QTDIR} "$ENV{HOME}/Qt/5.15.2/gcc_64")
list(APPEND CONFIGURE_ARGS -GNinja -DCMAKE_BUILD_TYPE=Debug)
```

This file is ignored by Git to prevent it being shared with other developers, but you could always copy another file into its place (e.g. during a CI build on GitHub Actions).


igorkorsukov:
And what is the advantage over `ninja_build.sh` ? `ninja_build.sh` is also a single cross-platform build script

shoogle:
@igorkorsukov, `ninja_build.sh` works on multiple operating systems but only with a single build tool, whereas `build.cmake` will work with all build tools. Of course we could adapt `ninja_build.sh` to work with other tools besides Ninja. It's written in Bash so that does have certain advantages in terms of familiarity to developers. I actually have been using [my own cross-platform Bash script](https://gist.github.com/shoogle/3c61f7fcd7a086947034ca90167d857c) to build MuseScore myself for a while now. However, I decided to switch to CMake before contributing it because we already use CMake for the other build scripts so we know it will be installed and available in `PATH` on all machines.

igorkorsukov:
> @igorkorsukov, `ninja_build.sh` works on multiple operating systems but only with a single build tool, whereas `build.cmake` will work with all build tools. Of course we could adapt `ninja_build.sh` to work with other tools besides Ninja. It's written in Bash so that does have certain advantages in terms of familiarity to developers. I actually have been using [my own cross-platform Bash script](https://gist.github.com/shoogle/3c61f7fcd7a086947034ca90167d857c) to build MuseScore myself for a while now. However, I decided to switch to CMake before contributing it because we already use CMake for the other build scripts so we know it will be installed and available in `PATH` on all machines.

Do you think we need other builds tools? (except for ninja)
It may be better, on the contrary, to eliminate the builds tools zoo and only support the ninja.
Then it will be easier to write documentation, teach others, solve build problems, and so on.

Using `cmake` instead of `sh` is an interesting idea!

build.cmake
```diff
@@ -0,0 +1,191 @@
...
+    include(ProcessorCount)
+    ProcessorCount(CPUS)
+endif()
+
+# Qt Creator
+if(DEFINED ENV{QTDIR})
+    set(ENV{PATH} "$ENV{QTDIR}/bin:$ENV{PATH}") # Add selected build kit to PATH
+endif()
+
+# MSVC / Visual Studio
```
Jojo-Schmitz, line 57: What about QtCreator/MinGW, and QtCreator/MSVC?

build/gen-qt-projectfile.bat
```diff
@@ -1,25 +1,25 @@
...
-echo.
-echo.
-
-echo RESOURCES = \
-echo     ./mu4/scenes/palette/palette.qrc \
-echo     ./telemetry/telemetry_resources.qrc \
-echo.
-
-cd /d %OLD_DIR%
+@echo off
```
Jojo-Schmitz, line 1: Is that (I assume it to be cr vs. cr/lf) the reason why `lupdate` doesn't really work on Windows?

build.cmake
shoogle: Not got that far yet

shoogle:
> Do you think we need other builds tools? (except for ninja) It may be better, on the contrary, to eliminate the builds tools zoo and only support the ninja.

I think it's a good idea to use Ninja on all platforms. I don't think we should drop support for the other tools altogether, but we can certainly focus the documentation around Ninja. It's much easier to say *"use CMake and Ninja"* rather than *"use CMake and whatever your platform's native build tool happens to be"*.

shoogle:
The main disadvantage of using CMake over a native shell is that it's not possible to `source` (or `CALL`) a native script file to set environment variables in the current process. Of course you can include other CMake files that get and set environment variables within CMake, but this doesn't help with existing scripts that are outside our control (e.g. `vcvars*.bat` on Windows). I had to create a [wrapper script for CMake](https://github.com/musescore/MuseScore/pull/7531/files#diff-f8d8b1f2a6f114e47480cc4c22393f263fe69afa0c1d9962e2f338ebccd9954a) to get around this issue, though it's not an ideal solution.

igorkorsukov:
> The main disadvantage of using CMake over a native shell is that it's not possible to `source` (or `CALL`) a native script file to set environment variables in the current process. Of course you can include other CMake files that get and set environment variables within CMake, but this doesn't help with existing scripts that are outside our control (e.g. `vcvars*.bat` on Windows). I had to create a [wrapper script for CMake](https://github.com/musescore/MuseScore/pull/7531/files#diff-f8d8b1f2a6f114e47480cc4c22393f263fe69afa0c1d9962e2f338ebccd9954a) to get around this issue, though it's not an ideal solution.

I came to the same solution when I was doing a common build script using ninja :)
https://github.com/musescore/MuseScore/blob/master/ninja_build.bat

shoogle:
@igorkorsukov, I saw, and I even nicked that script to create mine :grin:. It's unfortunate though because CMD screws around with the arguments (`-DFOO=BAR` becomes 2 separate arguments: `-DFOO` and `BAR`). Fortunately CMake can cope with either form so it doesn't matter here, but would matter in some situations (e.g. file path contains `=`), so it's a shame we can't avoid CMD entirely.

shoogle:
I think this is good enough for people to start using. We can add the `package` step later.

The configuration works with all generators on all platforms. The build fails with MinGW and Ninja on Windows, but works with MSVC. I get the same issue when I try to build using `Makefile.mingw` or `ninja_build.bat` so I don't think it is anything to do with this script.

igorkorsukov (approved):

Jojo-Schmitz:
> The build fails with MinGW and Ninja on Windows, but works with MSVC. I

You mean MinGW builds don't work any more, or just not when using that build script?

shoogle:
@Jojo-Schmitz, MinGW is not working for me any more even with `Makefile.mingw`. It's been that way for over a year now, which I why I switched to using MSVC some time ago.

Jojo-Schmitz:
I see. As long as it continues to work with QtCreator I couldn't care less ;-)

MarcSabatella:
Same, I use QtCreator/MinGW when building on Windows, and hope that continues to work going forward.

shoogle:
@Jojo-Schmitz, @MarcSabatella, you won't be affected by this. In fact you've given me an idea: I can probably get MinGW and Ninja working with `build.cmake` simply by copying whatever settings Qt Creator uses with those generators.

PR #7532, opened 2021-02-15
Fix drawing of border on RoundedRectangle

cbjeukendrup:
Before: RoundedRectangle was not used in many places, and nowhere with a border. I tried using it for a ComboBox with a TextInputField, and then discovered that the border was drawn wrong, namely along the edge of the rect, causing the line thickness to be half of the given value, except at the rounded corners. 

After: the border is now drawn inside the rectangle, just like with a normal QML Rectangle.
I added borders to the test on the devtools page, it now looks like this:
Schermafbeelding 2021-02-15 om 19 55 53

I also added a Border QML Type for convenience, and a radius property to easily set a radius for all corners but change it for one corner, for example.

### DIFF
--- a/src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
+++ b/src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
@@ -480,6 +480,8 @@ Rectangle {
                 topLeftRadius: 40
 
                 color: "gray"
+                border.color: "red"
+                border.width: 3
             }
 
             RoundedRectangle {
@@ -489,6 +491,8 @@ Rectangle {
                 topRightRadius: 30
 
                 color: "gray"
+                border.color: "yellow"
+                border.width: 10
             }
 
             RoundedRectangle {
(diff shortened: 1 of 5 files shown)


src/framework/ui/uitypes.h
```diff
@@ -54,6 +54,18 @@ struct ContainerMeta
     ContainerMeta(const ContainerType::Type& type, int widgetMetaTypeId)
         : type(type), widgetMetaTypeId(widgetMetaTypeId) {}
 };
+
+class Border : public QObject
```
RomanPudashkin, line 58: I believe uicomponents module is a more suitable place for this

Eism: I'm more interested in why you decided to move the border to c++?

cbjeukendrup: Fixed, thanks!
Actually, it just didn't came to my mind to put it in QML 😊, but that's indeed a good idea!

RomanPudashkin (approved):

Eism (approved):

PR #6546, opened 2020-09-08
Crash during the initialization of start center, fix 275837

vpereverzev:
Resolves: https://musescore.org/en/node/278817

If there is no internet connection or an editor is blocked by firewall then we'll hang on download extensions step because of QEventLoop::exec(). The problem is that we can quit from event loop only when QNetworkReply::finished() signal will be emited.

PR #7533, opened 2021-02-16
fix #314950: fix inability to download extensions

dmitrio95:
Resolves: https://musescore.org/en/node/314950

The following issues in the tracker are also probably describing the same issue:
 - https://musescore.org/en/node/315198
 - https://musescore.org/en/node/312737

The reason of those download issues is a 20 seconds download timeout introduced in #6546. For most applications within MuseScore it works fine but this timeout is too small for downloading extensions of tens of megabytes in size (which, of course, depends on the Internet connection speed). This pull requests increases the timeout up to 2 hours for downloading extensions, I am not sure though whether a timeout makes sense in this context.

### DIFF
--- a/mscore/resourceManager.cpp
+++ b/mscore/resourceManager.cpp
@@ -22,6 +22,8 @@ namespace Ms {
 extern QString dataPath;
 extern QString mscoreGlobalShare;
 
+static constexpr int extensionDownloadTimeoutMs = 20 * 60 * 1000;
+
 ResourceManager::ResourceManager(QWidget *parent) :
       QDialog(parent)
       {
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
While 20 seconds surely is too showr, 2 hours seems too long, maybe 20 minutes?

dmitrio95:
Actually I don't like a hard timeout here at all because it forces us to make assumptions on the Internet connection speed on the user's side. 20 minutes timeout would work only if download goes at a rate of at least approximately 70 KB/s — which is probably the case for most users but it is unclear why a download should fail if it goes somewhat slower. 2 hours don't make it much better anyway so I changed it to 20 minutes for now. But maybe a better solution would be to disable the timeout for extensions download and leave the timeout only for the other cases of using `DownloadUtils`.

Jojo-Schmitz:
If a download takes more than 2 hours, it is quite likely you don't want it to happen at all, I'd guess

dmitrio95:
Yes, I agree that a 20 minutes timeout is probably better than 2 hours. The question is whether we want this timeout at all.

PR #7534, opened 2021-02-16
[MU4] Application state

Eism:


src/workspace/internal/workspacesettingsstream.cpp
```diff
@@ -78,7 +78,7 @@ void WorkspaceSettingsStream::write(const AbstractDataPtrList& settingsList, IOD
 void WorkspaceSettingsStream::writeSettings(XmlWriter& writer, const AbstractDataPtr& data) const
 {
     SettingsDataPtr settings = std::dynamic_pointer_cast<SettingsData>(data);
-    IF_ASSERT_FAILED(settings) {
+    if (!settings) {
```
RomanPudashkin, line 81: Can you please explain the reason for this change?

src/framework/ui/internal/uiconfiguration.cpp
```diff
@@ -173,3 +179,58 @@ float UiConfiguration::physicalDotsPerInch() const
     }
     return 100;
 }
+
+QByteArray UiConfiguration::state(const std::string& objectName) const
```
RomanPudashkin, line 183: I think, it is better to rename this method to windowState(pageName) or pageState(pageName)

src/workspace/workspacetypes.h
```diff
@@ -61,6 +61,13 @@ struct ToolbarData : public AbstractData
 
 using ToolbarDataPtr = std::shared_ptr<ToolbarData>;
 
+struct UiArrangmentData : public AbstractData
+{
+    std::map<std::string /*key*/, Val> values;
```
RomanPudashkin, line 66: I think this is a too abstract way to store a window state. Better to store state directly in QByteArray. Maybe it's better to create a special struct:

struct PageState {
std::string pageName;
QByteArray state;
}

src/workspace/internal/workspacesettingsstream.cpp
Eism: because I created stubs for all data and we need to write them to a file for recovery after restart

src/workspace/workspacetypes.h
Eism: there may be other data in the future

RomanPudashkin: If we want to store a new property for a window then we can just add a new member to PageState. Val is a too abstract way to store data in this case, because you only need to store an array of bytes. Val can store anything and it can be confusing

src/workspace/internal/workspacesettingsstream.cpp
RomanPudashkin: Seems there is a bug in NewWorkspaceData::addStubData. You should add an object of a concrete type. At the moment you add only objects of UiArrangementData type

src/workspace/view/newworkspacemodel.cpp
```diff
@@ -143,10 +143,35 @@ QVariant NewWorkspaceModel::createWorkspace()
...
+        addStubData(newWorkspace, tag);
         for (AbstractDataPtr data : currentWorkspace->dataList(tag)) {
             newWorkspace->addData(data);
         }
     }
 
     return QVariant::fromValue(newWorkspace);
 }
+
+void NewWorkspaceModel::addStubData(IWorkspacePtr workspace, WorkspaceTag tag) const
```
RomanPudashkin, line 155: I think, better to place this code to the workspace implementation, not in this model

src/workspace/workspacetypes.h
Eism: If you look at the implementation of UiConfiguration::uiValue, then you will understand that UiArrangment is similar to settings. Key - value. We don't know what keys are stored and what types of values are there. And it's better not to complicate with this knowledge

src/workspace/internal/workspacesettingsstream.cpp
Eism: reverted, fixed bug

src/workspace/view/newworkspacemodel.cpp
Eism: decided to save tags instead of creating stubs

src/workspace/internal/workspaceuiarrangmentstream.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+using namespace mu::workspace;
+using namespace mu::framework;
+using namespace mu::system;
+
+static const std::string UI_ARRANGMENT_TAG("UiArrangment");
+static const std::string UI_SETTINGS_TAG("Settings");
+static const std::string UI_SETTINGS_NAME_TAG("name");
+
+AbstractDataPtrList WorkspaceUiArrangmentStream::read(IODevice& sourceDevice) const
```
RomanPudashkin, line 34: Seems that this class is a full copy of WorkspaceSettingsStream (the only difference is in UI_SETTINGS_TAG). I suggest using WorkspaceSettingsStream to read UI arrangements from a workspace. You can add WorkspaceTag (or some other enum that contains Settings and UiArrangement values)  to WorkspaceSettingsStream constructor and resolve the required XML tag inside that class

src/framework/global/iworkspaceuiarrangment.h
```diff
@@ -0,0 +1,40 @@
...
+//=============================================================================
+#ifndef MU_FRAMEWORK_IWORKSPACEUIARRANGMENT_H
+#define MU_FRAMEWORK_IWORKSPACEUIARRANGMENT_H
+
+#include "modularity/imoduleexport.h"
+#include "val.h"
+#include "async/notification.h"
+
+namespace mu::framework {
+class IWorkspaceUiArrangment : MODULE_EXPORT_INTERFACE
```
RomanPudashkin, line 27: Better to rename this class to IWorkspaceSettings

src/framework/ui/internal/uiconfiguration.cpp
```diff
@@ -173,3 +179,58 @@ float UiConfiguration::physicalDotsPerInch() const
...
+std::string UiConfiguration::byteArrayToString(const QByteArray& byteArray) const
+{
+    QByteArray byteArray64 = byteArray.toBase64();
+    return byteArray64.toStdString();
+}
+
+mu::Val UiConfiguration::uiValue(const std::string& key) const
+{
+    Val val = workspaceUiArrangment()->value(key);
+    if (!val.isNull()) {
```
RomanPudashkin, line 219: Better to add contains(key) to an interface of this class and hide this logic inside it

src/workspace/internal/workspacesettingsstream.h
```diff
@@ -38,6 +40,10 @@ class WorkspaceSettingsStream : public IWorkspaceDataStream
 private:
     SettingsDataPtr readSettings(framework::XmlReader& reader) const;
     void writeSettings(framework::XmlWriter& writer, const AbstractDataPtr& data) const;
+
+    std::string_view tagName() const;
+
+    WorkspaceTag m_tag;
```
RomanPudashkin, line 46: Please, add some default value

src/workspace/internal/workspacesettings.cpp
```diff
@@ -0,0 +1,83 @@
...
+    if (settingsData && settingsData->values.find(key) != settingsData->values.end()) {
+        return settingsData->values[key];
+    }
+
+    return Val();
+}
+
+bool WorkspaceSettings::setValue(WorkspaceTag tag, const std::string& key, const mu::Val& value) const
+{
+    AbstractDataPtr abstractData = currentWorkspace()->data(tag);
```
RomanPudashkin, line 59: Please, add a check on null here

src/workspace/internal/workspacesettings.cpp
```diff
@@ -0,0 +1,83 @@
...
+    if (!currentWorkspace) {
+        return false;
+    }
+
+    return containsTag(currentWorkspace->tags(), tag);
+}
+
+mu::Val WorkspaceSettings::value(WorkspaceTag tag, const std::string& key) const
+{
+    AbstractDataPtr abstractData = currentWorkspace()->data(tag);
```
RomanPudashkin, line 48: Please, add a check on null here

src/workspace/internal/workspace.h
```diff
@@ -38,6 +38,9 @@ class Workspace : public IWorkspace
     std::string name() const override;
     std::string title() const override;
 
+    std::vector<WorkspaceTag> tags() const override;
```
RomanPudashkin, line 41: It would be great to add an alias:

WorkspaceTagList = std::vector<WorkspaceTag>

RomanPudashkin (approved):

PR #7544, opened 2021-02-16
[MU3] Fix #317264 - instruments window staves column doesn't show changed part names

njvdberg:
Resolves: https://musescore.org/en/node/317264

<code>PartListItem::PartListItem(Part* p, QTreeWidget* lv)</code> now takes the part name for the name, as it used to be, instead of the instrument name. Unless the instrument name is an empty string, in which case the part name is used.

**NOTE  1**
This is a <code>3.x</code> only PR and **not** required for <code>master</code> because this issue doesn't apply to <code>master</code>. 

**NOTE 2**
PR #7571 implements preventing an empty string in the instrument list for <code>master</code>.

### DIFF
--- a/mscore/instrwidget.cpp
+++ b/mscore/instrwidget.cpp
@@ -426,7 +426,7 @@ PartListItem::PartListItem(Part* p, QTreeWidget* lv)
       part = p;
       it   = 0;
       op   = ListItemOp::KEEP;
-      _name = QString(p->instrument()->trackName());
+      _name = QString(p->partName().isEmpty() ? p->instrument()->trackName() : p->partName());
       setSoloist(false); //TODO, must be taken from part.
       setFlags(flags() | Qt::ItemIsUserCheckable);
       }


mscore/instrwidget.cpp
```diff
@@ -426,7 +426,7 @@ PartListItem::PartListItem(Part* p, QTreeWidget* lv)
       part = p;
       it   = 0;
       op   = ListItemOp::KEEP;
-      _name = QString(p->instrument()->trackName());
+      _name = QString(p->partName());
```
Jojo-Schmitz, line 429: Mayby use `instrument->trackname()` as the fallback it `partname()` is empty?

njvdberg: It crossed my mind but I reverted to what it used to be in 3.5.
Nevertheless a good idea if you don't mind to deviate from what it is used to be :wink:

Jojo-Schmitz:
The better is good's enemy ;-)

mscore/instrwidget.cpp
Jojo-Schmitz: The better is good's enemy ;-)

njvdberg: It is by the way now exactly as it is in MU4 already. If we change it, then it should be changed in MU4 too. Which could be a nice challenge of course...

Jojo-Schmitz: Yep

Jojo-Schmitz: Better than potentially an empty sting there

njvdberg: Done.

Jojo-Schmitz: thanks

vpereverzev:
closed because the necessary parts have already been transferred to the master

Jojo-Schmitz:
still needed for 3.x thoiugh

RobFog:
> closed because the necessary parts have already been transferred to the master

As @njvdberg's description states this pull request apparently fixes a 3.x-only regression.

vpereverzev:
ah, @njvdberg sorry, didn't recognized that it's not just a duplicate. I'll reopen this one

PR #7545, opened 2021-02-17
Fix #317631: Deleting all text in edit mode breaks Undo

mattmcclinch:
Resolves: https://musescore.org/en/node/317631.

If all text is deleted from a text element, the element itself is deleted. In order for undo to restore the text that was deleted, we must reset the element's text to what it was before editing, just prior to removing the element.


Jojo-Schmitz:

PR #7546, opened 2021-02-17
[MU3] fix REMEMBERLASTCONNECTIONS & TIMEBASEMASTER preferences not affecting jackParametersChanged

SKefalidis:
Resolves: https://musescore.org/en/node/317665

These 2 preferences were being treated as normal preferences. That way their values were set before jackParametersChanged was set, so they didn't affect it.

- [n\a] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/mscore/prefsdialog.cpp
+++ b/mscore/prefsdialog.cpp
@@ -332,8 +332,6 @@ void PreferenceDialog::start()
                   new BoolPreferenceItem(PREF_UI_APP_STARTUP_SHOWSTARTCENTER, showStartcenter),
                   new BoolPreferenceItem(PREF_UI_APP_STARTUP_SHOWTOURS, showTours),
                   new BoolPreferenceItem(PREF_APP_TELEMETRY_ALLOWED, collectTelemetry),
-                  new BoolPreferenceItem(PREF_IO_JACK_TIMEBASEMASTER, becomeTimebaseMaster),
-                  new BoolPreferenceItem(PREF_IO_JACK_REMEMBERLASTCONNECTIONS, rememberLastMidiConnections),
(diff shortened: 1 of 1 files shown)


ericfont:
I'm verifying that this commit has fixed that issue I reported.  I'm now able to turn on and off MuseScore's JackTransport setting, and so can use (or not use) external Jack programs like Ardour in sync with MuseScore.  Thanks!

mscore/prefsdialog.cpp
```diff
@@ -1380,8 +1380,9 @@ void PreferenceDialog::apply()
                         || preferences.getBool(PREF_IO_JACK_REMEMBERLASTCONNECTIONS) != rememberLastMidiConnections->isChecked()
                         || preferences.getBool(PREF_IO_JACK_TIMEBASEMASTER) != becomeTimebaseMaster->isChecked())
                         && (wasJack && nowJack);
-            //till this
 
+            preferences.setPreference(PREF_IO_JACK_TIMEBASEMASTER, becomeTimebaseMaster->isChecked());
+            preferences.setPreference(PREF_IO_JACK_REMEMBERLASTCONNECTIONS, rememberLastMidiConnections->isChecked());
             preferences.setPreference(PREF_IO_JACK_USEJACKTRANSPORT, jackDriver->isChecked() && useJackTransport->isChecked()); //this
```
ericfont, line 1386: I would suggest you might as well remove this meaningless " //this" comment at the end of this line while you are making this commit since it is an adjacent line.

ericfont:
and the remember last connection box also seems to be working to selectively or not remember an extra midi connection I make.

PR #7566, opened 2021-02-18
[MU3] Fix #317703: Disabling "Enable Auto-Size" for Title text in template not respected

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/317703

Also fixed a typo.

Seems for all templates (all but 00-Blank.mscx, which doesn't have a frame at all, and Create_New_Score.mscz and My_First_score.mscx) this `boxAutoSize` is explictly set to off, on purpose? Apparently not, shows just now, with the bug fixed...
Similar for `useStandardNoteNames` (resulting in have neither this nor any of the German modes set).

So I fixed the templates too
Version for master in #7584 (and got merged there)


Jojo-Schmitz:
Seems for all templates (all but 00-Blank.mscx and not in Create_New_Score and My_First_score) this is explictly set to off, on purpose?

share/templates/01-General/00-Blank.mscx
```diff
@@ -5,8 +5,6 @@
     <currentLayer>0</currentLayer>
     <Division>480</Division>
     <Style>
-      <enableVerticalSpread>1</enableVerticalSpread>
```
njvdberg, line 8: Why is this removed. This disables the vertical staves adjustment while is is intended to be the default

share/templates/01-General/01-Treble_Clef.mscx
```diff
@@ -3,12 +3,8 @@
   <Score>
     <LayerTag id="0" tag="default"></LayerTag>
     <currentLayer>0</currentLayer>
-    <Synthesizer>
-      </Synthesizer>
     <Division>480</Division>
     <Style>
-      <enableVerticalSpread>1</enableVerticalSpread>
```
njvdberg, line 10: Same here (and on also same for the other templates).

share/templates/02-Choral/03-SATB_+_Piano.mscx
```diff
@@ -1,14 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <museScore version="3.02">
-  <programVersion>3.6.0</programVersion>
```
njvdberg, line 3: Should be removed indeed.

share/templates/01-General/00-Blank.mscx
Jojo-Schmitz: If is is the default, it doesn need to be in the style settings, won't harm, but serves no purpose either

share/templates/01-General/01-Treble_Clef.mscx
Jojo-Schmitz: same here

share/templates/01-General/00-Blank.mscx
Jojo-Schmitz: But indeed ist is not, it does default to false.:
https://github.com/musescore/MuseScore/blob/2ba915f6bb080a80d8fabf97a374a811171c9e17/libmscore/style.cpp#L84
To me that seems a bug, doesn't it?

Jojo-Schmitz: It should be on foe every new score. not just for those created from templates, doesn't it?

Jojo-Schmitz: It is now...

Jojo-Schmitz: Hmm, no, bad idea. I'll revert that part of the change

njvdberg: Indeed, bad idea :wink: I made this value default to make sure pre-3.6 score will not get vertical staves adjustment enabled unintentionally. That's why the templates were changed.

Jojo-Schmitz: Thanks for brining this up ;-) Saved me a brown bag I guess

njvdberg (approved):

share/templates/01-General/00-Blank.mscx
njvdberg: At least one :wink:

MarcSabatella: Really, though, couldn't it have been the default for 3.6?  We explicitly keep a separate set of pre-3.6 defaults we apply explicitly upon load for older scores, that mechanism should have worked for ensuring compatibility.

Of course, I think I know the answer here: that mechanism probably didn't exist, or didn't work well when you started working on the spread algorithm.  Sopbox alert: this to me is probably the biggest mistake we made in 3.6 development.  We should have decided on and implemented our compatibility strategy first, before embarking on any further changes.  We would have saved us endless headaches as it would have been fully tested and debugged long before release, and we'd also have a more consistent story regarding these defaults.  See also the chord symbol playback defaults - it was as we approached 3.5 knowing full well 3.6 was also going to be a thing that I started sounding the alarm on this.  Oh well...

njvdberg: When starting the work on vertical staves adjustment this mechanism didn't exist, wasn't even discussed. There was one compatibility issue, the chord symbol playback which gave a lot of discussion. 
Based on experiences in my previous life, the easiest way to keep compatibility was to set the default to false and update the templates. At least at the moment I was working on it. Which of course doesn't mean the new pre-3.6 defaults can't be used for vertical staves adjustment. In fact, I think it it pretty straightforward, it is only one parameter. There are no "shared" parameters.

With hindsight, we made some mistakes when starting 3.6 (this compatibility and not using the new code style for 3.6 too, causing quite some migration problems to master) but hopefully we learned from it :smiley:

Jojo-Schmitz: Don't get me started on that code style thing...

share/templates/06-Popular/02-Bluegrass_Band.mscx
```diff
@@ -469,11 +468,9 @@
     <Staff id="1">
       <VBox>
         <height>10</height>
-        <boxAutoSize>0</boxAutoSize>
         <Text>
           <style>Title</style>
-          <text>Bluegrass Band
-</text>
+          <text>Bluegrass Band</text>
```
Jojo-Schmitz, line 473: Found when porting this PR to master...

share/templates/01-General/00-Blank.mscx
Jojo-Schmitz: Question is whether to change `enableVerticalSpread` to `true` (and remove it from the templates) for master?

PR #7568, opened 2021-02-18
Fix #317747: Measure number appears after edit in first measure after section break with intervening frame

mattmcclinch:
Resolves: https://musescore.org/en/node/317747.

This was just a simple case of forgetting to update `mb` with the result of `mb->findPotentialSectionBreak()`.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -4981,8 +4981,8 @@ void Score::doLayoutRange(const Fraction& st, const Fraction& et)
             else {
                   const MeasureBase* mb = lc.nextMeasure->prev();
                   if (mb)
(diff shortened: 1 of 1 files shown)


mattmcclinch:
This is handled in #7569.

MarcSabatella:
Hah, I see you submitted yours first, nice job!  Really, I'm not sure the second change I made is strictly needed, certainly it isn't just for this exact case.  But it could be in other cases, and certainly shouldn't hurt.

mattmcclinch:
I included a `nullptr` check in mine. You might consider doing the same in yours in both places, since `findPotentialSectionBreak()` can potentially return `nullptr`.

PR #7571, opened 2021-02-19
[MU4] Port of PR #7544 to master.

njvdberg:
Prevents empty part name. If name part is empty the instrument panel will show the instrument name instead of an empty string.

### DIFF
--- a/src/instruments/view/instrumentpaneltreemodel.cpp
+++ b/src/instruments/view/instrumentpaneltreemodel.cpp
@@ -568,7 +568,7 @@ AbstractInstrumentPanelTreeItem* InstrumentPanelTreeModel::buildPartItem(const P
 void InstrumentPanelTreeModel::updatePartItem(PartTreeItem* item, const Part* part)
 {
     item->setId(part->id());
-    item->setTitle(part->partName());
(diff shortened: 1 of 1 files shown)


src/instruments/view/instrumentpaneltreemodel.cpp
```diff
@@ -567,8 +567,9 @@ AbstractInstrumentPanelTreeItem* InstrumentPanelTreeModel::buildPartItem(const P
 
 void InstrumentPanelTreeModel::updatePartItem(PartTreeItem* item, const Part* part)
 {
+    std::cout << "InstrumentPanelTreeModel::updatePartItem: title = \"" << part->partName().toStdString() << '"' << std::endl;
```
igorkorsukov, line 570: better use `LOGD` or `qDebug()`

RomanPudashkin: I think it's better to just remove this log, because other developers don't need to see this information often

njvdberg: Better not use at all! Was a left-over of some testing, removed.

PR #7583, opened 2021-02-20
[MU4] General menu. View

Eism:
Implemented:
Toggle Palettes
Toggle Instruments
Toggle Inspector
Toggle StatusBar
Toggle Note Input toolbar
Toggle Notation toolbar
Toggle Undo/Redo toolbar
Toggle Playback toolbar
Toggle Master Palette
Toggle Full Screen

Select workspace
Open workspaces dialog from menubar

Show invisible on the score 
Show Unprintable on the score 
Show Frames on the score 
Show Page Margins on the score 
Mark Irregular Measures on the score 

Fixed some bugs in restore application state


src/appshell/iapplicationactioncontroller.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "modularity/imoduleexport.h"
+#include "retval.h"
+
+namespace mu::appshell {
+class IApplicationActionController : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IApplicationActionController)
+public:
+    ~IApplicationActionController() = default;
```
RomanPudashkin, line 30: You missed virtual

src/appshell/internal/appshellconfiguration.cpp
```diff
@@ -22,7 +22,16 @@ using namespace mu::appshell;
...
+    return false;
+#endif
+}
+
+bool AppShellConfiguration::isFullScreenAvailable() const
+{
+#ifndef Q_OS_MAC
     return true;
 #else
     return false;
```
RomanPudashkin, line 37: But full screen is also available on over platfroms...

src/appshell/iappshellconfiguration.h
```diff
@@ -31,8 +31,19 @@ class IAppShellConfiguration : MODULE_EXPORT_INTERFACE
     virtual ~IAppShellConfiguration() = default;
 
     virtual bool isAppUpdatable() const = 0;
+    virtual bool isFullScreenAvailable() const = 0;
 
     virtual ValCh<QStringList> recentScoreList() const = 0;
+
+    virtual ValCh<bool> isPalettePanelVisible() const = 0;
```
RomanPudashkin, line 38: I think, better to implement this using only two method:

Channel<PanelKey> pannelVisibilityChanged()
bool isPannelVisible(PanelKey key)

PanelKey {
InstrumentsPanel,
StatusBar,
Navigator,
....
}

src/appshell/internal/applicationactioncontroller.cpp
```diff
@@ -19,16 +19,43 @@
...
     dispatcher()->reg(this, "quit", this, &ApplicationActionController::quit);
+
+    dispatcher()->reg(this, "fullscreen", this, &ApplicationActionController::toggleFullScreen);
+}
+
+mu::ValCh<bool> ApplicationActionController::isFullScreen() const
+{
+    ValCh<bool> result;
+    result.ch = m_fullScreenChannel;
+    result.val = mainWindow()->qMainWindow() ? mainWindow()->qMainWindow()->isFullScreen() : false;
```
RomanPudashkin, line 38: Better to move this code to a separate method because you use it in 2 different places:

bool isFullScreen() {
return mainWindow()->qMainWindow() ? mainWindow()->qMainWindow()->isFullScreen() : false;
}

src/appshell/internal/applicationactions.cpp
```diff
@@ -28,6 +28,10 @@ const ActionList ApplicationActions::m_actions = {
     ActionItem("quit",
                ShortcutContext::Any,
                QT_TRANSLATE_NOOP("action", "Quit")
+               ),
+    ActionItem("fullscreen",
+               ShortcutContext::Any,
+               QT_TRANSLATE_NOOP("action", "Full Screen")
```
RomanPudashkin, line 34: Is there any description for this action?

src/appshell/view/dockwindow/dockpanel.cpp
```diff
@@ -158,6 +175,16 @@ void DockPanel::setClosable(bool closable)
     emit closableChanged(closable);
 }
 
+void DockPanel::setVisible(bool visible)
```
RomanPudashkin, line 178: Better to move this method (and all related ones) to DockView to prevent duplication of code

src/notation/internal/notationactions.cpp
```diff
@@ -677,6 +677,66 @@ const ActionList NotationActions::m_actions = {
...
+               ShortcutContext::NotationActive,
+               QT_TRANSLATE_NOOP("action", "Status Bar"),
+               QT_TRANSLATE_NOOP("action", "Toggle 'Status Bar'")
+               ),
+    ActionItem("toggle-noteinput",
+               ShortcutContext::NotationActive,
+               QT_TRANSLATE_NOOP("action", "Note Input"),
+               QT_TRANSLATE_NOOP("action", "Toggle 'Note Input' toolbar")
+               ),
+    ActionItem("toogle-notationtoolbar",
```
RomanPudashkin, line 706: Typo:

toggle

You should check in other places too

src/appshell/view/appmenumodel.cpp
```diff
@@ -199,35 +239,41 @@ MenuItem AppMenuModel::editItem()
...
-        makeAction("inspector", selectedElementOnScore()), // need implement
-        makeAction("toggle-playpanel", selectedElementOnScore()), // need implement
-        makeAction("toggle-navigator", selectedElementOnScore()), // need implement
-        makeAction("toggle-timeline", selectedElementOnScore()), // need implement
-        makeAction("toggle-mixer", selectedElementOnScore()), // need implement
-        makeAction("synth-control", selectedElementOnScore()), // need implement
-        makeAction("toggle-selection-window", selectedElementOnScore()), // need implement
-        makeAction("toggle-piano", selectedElementOnScore()), // need implement
-        makeAction("toggle-scorecmp-tool", scoreOpened()), // need implement
+        makeAction("toggle-palette", isNotationPage(), configuration()->isPalettePanelVisible().val),
```
RomanPudashkin, line 242: I believe we should do all those checks in concrete modules. For example, palettes module register "toggle-palette" action in PaletteActionsController. Then PaletteActionsController should answer that it is possible to use this action now. We already have an interface for this - Actionable::canReceiveAction

@igorkorsukov what do you think?

src/notation/internal/notationconfiguration.h
```diff
@@ -89,11 +91,32 @@ class NotationConfiguration : public INotationConfiguration, public async::Async
     void setToolbarActions(const std::string& toolbarName, const std::vector<std::string>& actions) override;
 
     ValCh<bool> isNavigatorVisible() const override;
-    void setNavigatorVisible(bool visible) override;
+    void setIsNavigatorVisible(bool visible) override;
 
     ValCh<framework::Orientation> canvasOrientation() const override;
     void setCanvasOrientation(framework::Orientation orientation) override;
 
+    ValCh<bool> isPalettePanelVisible() const override;
```
RomanPudashkin, line 99: I believe you should merge all those isSomethingVisible methods in one

src/appshell/internal/appshellconfiguration.cpp
Marr11317: You misread this: it's `#ifndef`, not `#ifdef`. This means fullscreen is available on every platform *except* macos. 
I didn't know macos had no fullscreen mode though...

cbjeukendrup: But macOS _does_ have full screen! 
image
The default keyboard shortcut is ⌃⌘F.

Marr11317: I don't know why it's there then... Maybe an old qt bug? Maybe something can be found in git history?

IsaacWeiss: I added that a while ago. Whatever version of macOS was current then automatically appended the option to the View menu even if it was already present, so we hid the first one in order not to duplicate the macOS-provided one.

cbjeukendrup: @IsaacWeiss Thanks, that's good to know. For me on macOS Big Sur, no Full Screen option is added to the menu, neither in MuseScore 3 nor in 4, so now, there is no such option in the menu at all. However, the keyboard shortcut magically works. 
But indeed, on macOS 10.15, a full screen item is added. So we would need a version check, which is doable but not really nice. Another disadvantage of the macOS provided item is that it seems to use the system language instead of MuseScore's language. 

But maybe there is good news: there seems to be a way to stop macOS from adding that item, see https://stackoverflow.com/questions/52154977/how-to-get-rid-of-enter-full-screen-menu-item. We could put it in the `init` method of MacOSPlatformTheme. This gives us full control, but also gives us the responsibility to manage full screen correctly. We would need to see how that works out in practice.

IsaacWeiss: Not magic, see #2522. :-) It is managed correctly—the two menu entries trigger(ed) the exact same action.

cbjeukendrup: Ah, nice! I didn't realize for a moment that keyboard shortcuts of course also work when they are not in a menu item. In that case, we can indeed be confident about abandoning macOS's full screen item and use our own, because in fact that is already kind of what happens. I could create a PR soon.
@Eism If we decide to do it like that, you could actually remove the whole `isFullScreenAvailable` method, as it should be always true.

We could think a bit about "Enter" and "Exit". It would be nice if actions could have multiple titles based on the current state. For example, "Show Palettes" and "Hide Palettes" instead of only "Toggle Palettes". It seems that other macOS apps also prefer that. With our current actions system, that isn't possible, and it would maybe involve many changes.

Eism: Thanks guys for getting involved in this issue
I really didn't know why fullscreen didn't work on macOS and decided not to change the logic. A little later, I wanted to investigate this problem
Ok, now I understand(thanks to your help) that I can safely remove this code

src/appshell/iapplicationactioncontroller.h
Eism: thanks, fixed

src/appshell/internal/appshellconfiguration.cpp
Eism: thanks, removed

src/appshell/iappshellconfiguration.h
Eism: thanks, fixed

src/appshell/internal/applicationactions.cpp
Eism: thanks, added description

src/appshell/view/dockwindow/dockpanel.cpp
Eism: thanks, fixed

src/notation/internal/notationactions.cpp
Eism: thanks, fixed

src/notation/internal/notationconfiguration.h
Eism: thanks, fixed

src/appshell/view/appmenumodel.cpp
Eism: I will fix it in next PR

RomanPudashkin (approved):

PR #7585, opened 2021-02-20
[MU4] General menu. Help

Eism:
Implemented:
Online Handbook
Ask for Help
Report a Bug
Leave Feedback
Revert to Factory Settings
About Qt

Fixed updating page of languages

### DIFF
--- a/src/appshell/internal/applicationactioncontroller.cpp
+++ b/src/appshell/internal/applicationactioncontroller.cpp
@@ -29,6 +29,8 @@ void ApplicationActionController::init()
 {
     dispatcher()->reg(this, "quit", this, &ApplicationActionController::quit);
 
+    dispatcher()->reg(this, "about-qt", this, &ApplicationActionController::openAboutQtDialog);
+
     dispatcher()->reg(this, "online-handbook", this, &ApplicationActionController::openOnlineHandbookPage);
     dispatcher()->reg(this, "ask-help", this, &ApplicationActionController::openAskForHelpPage);
     dispatcher()->reg(this, "report-bug", this, &ApplicationActionController::openBugReportPage);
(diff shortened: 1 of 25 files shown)


src/appshell/internal/applicationactioncontroller.cpp
```diff
@@ -21,14 +21,72 @@
...
 }
+
+void ApplicationActionController::openAboutQtDialog()
+{
+    QApplication::aboutQt();
+}
+
+void ApplicationActionController::openOnlineHandbookPage()
+{
+    QString currentLanguageCode = languagesService()->currentLanguage().val.code;
```
RomanPudashkin, line 52: 1. better to implement a separate method for getting current language code:

currentLanguageCode() {
return languagesService()->currentLanguage().val.code
}

2. I think it is better to remove languageCode from configuration methods and get it dirrectly in their implementation. I doubt that somewhere we will use any of these methods with non-current language code

src/appshell/internal/applicationactioncontroller.cpp
```diff
@@ -21,14 +21,72 @@
...
+
+void ApplicationActionController::openLeaveFeedbackPage()
+{
+    std::string leaveFeedbackUrl = configuration()->leaveFeedbackUrl();
+    interactive()->openUrl(leaveFeedbackUrl);
+}
+
+void ApplicationActionController::revertToFactorySettings()
+{
+    std::string question = "This will reset all your preferences.\n"
```
RomanPudashkin, line 79: It seems that you missed a translation

src/appshell/internal/applicationactioncontroller.cpp
Eism: moved getting language code to configuration

RomanPudashkin (approved):

PR #7587, opened 2021-02-20
Fixed wrong bitwise operator: & instead of |

L0uisc:
Resolves: *(direct link to the issue)*

In `globals.h`, the following enum is declared:

```cpp

enum TelemetryDataCollectionType : unsigned char {
      COLLECT_NO_DATA = 0,
      COLLECT_CRASH_FREE_DATA = 1,
      COLLECT_INSPECTOR_DATA = 1 << 1,
      COLLECT_SHORTCUT_AND_MENU_DATA = 1 << 2,
      COLLECT_ALL_DATA = COLLECT_CRASH_FREE_DATA & COLLECT_INSPECTOR_DATA & COLLECT_SHORTCUT_AND_MENU_DATA
};
```
The fifth member (`COLLECT_ALL_DATA`) will evaluate to 0, duplicating the first member (`COLLECT_NO_DATA`). This is most probably not what was intended. This fix changes the bitwise operator to `|` in order to make the last member a bit field with all bits on.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7588, opened 2021-02-21
[MU4] Update pull_request_template.md

Jojo-Schmitz:
Resolves a 404 error, apparently needed for master only (3.x has it correct, still even on 3.x PRs the  master template comes up, see #7587, probably because master is the default branch)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@ Resolves: *(direct link to the issue)*
 <!-- Use "x" to fill the checkboxes below like [x] -->
 
 - [ ] I signed [CLA](https://musescore.org/en/cla)
-- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/CodeGuidelines.md)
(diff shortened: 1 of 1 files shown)


shoogle:
For MuseScore links it's probably best to use the permanent numeric version: https://musescore.org/node/15435

You can find this by clicking 'Edit' on the page in question, or by inspecting the page source (press Ctrl+U [macOS: Cmd+U] in most browsers), which is necessary if you don't have permission to edit the page in question.

Jojo-Schmitz:
Well, that's not the perma-link though, that'd be something like https://musescore.org/en/node/15435/revisions/474525/view

Jojo-Schmitz:
Also I just took that link from the 3.x branch

Jojo-Schmitz:
Hmm, it is the wrong link anyhow, the one for the old 3.x code style.

shoogle:
Looks like new guidelines are at https://github.com/musescore/MuseScore/wiki/CodeGuidelines

Jojo-Schmitz:
Thanks, corrected. 
Hmm maybe better just correcting https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules to point to https://github.com/musescore/MuseScore/wiki/CodeGuidelines

PR #5474, opened 2019-11-15
fix #297040: Palette keyboard navigation

shoogle:
Implements standard keyboard navigation for an item view / tree view as demonstrated by QAbstractItemView and QTreeView classes. Also changes how elements are displayed as necessary to indicate the different kinds of selection (e.g. when current item is not selected).

Resolves: https://musescore.org/en/node/297040

## Features

- Adds blue outline for current item (selected items have blue fill).
- Changing item also selects it (unless Ctrl is held).
- More button behaves like a normal item for navigation purposes

__New keyboard shortcuts:__

- Asterisk (*) key expands or collapses all items
- Menu key (or [Ctrl+]Shift+F10) opens context menus
- Press any letter key to jump to palette starting with that letter

__Various improvements to screen reader output:__

- Palette names end in "palette" to distinguish from elements
- Says when palettes are expanded
- Says how many elements in palette, and palettes in view
- Says when palette elements are selected

__Selection behavior__

Slightly changes behaviour for __mouse users__ to match that described in [QAbstractItemView::ExtendedSelection](https://doc.qt.io/qt-5/qabstractitemview.html#SelectionMode-enum):

> When the user selects an item in the usual way, the selection is cleared and the new item selected. However, if the user presses the Ctrl key when clicking on an item, the clicked item gets toggled and all other items are left untouched. If the user presses the Shift key while clicking on an item, all items between the current item and the clicked item are selected or unselected, depending on the state of the clicked item. Multiple items can be selected by dragging the mouse over them.

The behaviour for __keyboard users__ is almost the same, with point and click replaced by arrow keys and spacebar. Keyboard users also have the ability to change current item without changing the selection, which is done by holding Ctrl when using the arrow keys, as described [here](https://doc.qt.io/qt-5/qabstractitemview.html#details).

This behaviour for keyboard and mouse matches that of File Explorer in Windows 10.

## Checklist

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
[...]

PR #7590, opened 2021-02-21
fix #313977: avoid adding a palette element twice on double-click

dmitrio95:
Resolves: https://musescore.org/en/node/313977

The issue doesn't happen in MuseScore 3.4.2 but happens in the 3.5 version. The possible reason is that #5474 has occasionally reverted the previous commit which had fixed this issue, 77e4347f10d7c520048165f01ce0b01fce6ff9b8. This pull request returns the empty double-click handler back. Removing the handler entirely causes two `onClicked` events being triggered and results in identical elements being added as well, so we have to put an empty handler there to prevent this.

### DIFF
--- a/mscore/qml/palettes/Palette.qml
+++ b/mscore/qml/palettes/Palette.qml
@@ -691,9 +691,7 @@ GridView {
                 }
 
                 onDoubleClicked: {
-                    const index = paletteCell.modelIndex;
-                    paletteView.selectionModel.setCurrentIndex(index, ItemSelectionModel.Current);
-                    paletteView.paletteController.applyPaletteElement(index, mouse.modifiers);
+                    // Empty handler to avoid onClicked being triggered twice on double-click.
                 }
             }
 


mscore/qml/palettes/Palette.qml
```diff
@@ -691,9 +691,7 @@ GridView {
                 }
 
                 onDoubleClicked: {
-                    const index = paletteCell.modelIndex;
-                    paletteView.selectionModel.setCurrentIndex(index, ItemSelectionModel.Current);
-                    paletteView.paletteController.applyPaletteElement(index, mouse.modifiers);
+                    // Empty handler to avoid onClicked being triggered twice on double-click.
```
Jojo-Schmitz, line 694: I love issues getting fixed by removing code ;-)

Jojo-Schmitz:

Jojo-Schmitz:
As far as I can tell the culprit code is still in master, see https://github.com/musescore/MuseScore/blob/bb068491ffd14a1cb7cf724c8d34531ca5a5f209/src/palette/qml/MuseScore/Palette/internal/Palette.qml#L543-L547

Jojo-Schmitz:
See #11462

PR #7591, opened 2021-02-21
[MU4] Correct wrong creating of score by the NewScore dialog.

njvdberg:
When creating a new score without a pick measure the score looks like:

Screenshot_20210221_171852

This score shows 2 issues:
 -1) There is a pickup measure (measure 2) of 1 quarter note. This is solved in <code>MasterNotation::createNew()</code> which used a wrong check for the pick measure.

 -2) The number of measures in the score is 1 more as specified in the dialog. The is causes by creating a measure in <code>NotationParts::setInstruments()</code> when the score doesn't contain any measures.

With this PR the created score is correct:

Screenshot_20210221_172737

Or with a pickup measure:

Screenshot_20210221_172828

### DIFF
--- a/src/notation/internal/notationparts.cpp
+++ b/src/notation/internal/notationparts.cpp
@@ -136,10 +136,6 @@ void NotationParts::setInstruments(const mu::instruments::InstrumentList& instru
     removeMissingInstruments(instruments);
     appendNewInstruments(instruments);
 
-    if (score()->measures()->empty()) {
-        score()->insertMeasure(ElementType::MEASURE, 0, false);
-    }
-
     sortParts(instruments);
 
     updateScore();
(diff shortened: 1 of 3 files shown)


RomanPudashkin (approved):

src/notation/internal/notationparts.cpp
```diff
@@ -136,10 +136,6 @@ void NotationParts::setInstruments(const mu::instruments::InstrumentList& instru
     removeMissingInstruments(instruments);
     appendNewInstruments(instruments);
 
-    if (score()->measures()->empty()) {
```
Eism, line 139: the application crashes if all instruments are removed from the score in the instruments dialog
I found in MU3 this insurance instrdialog.cpp: 571

I suggest disabling the Ok button in the instruments dialog when no instruments selected like in the new score wizard.
Then we can remove this code safely

RomanPudashkin: изображение

This is what Elnur means

njvdberg: @Eism , @RomanPudashkin Thanks for both hints! I will update the PR.

Eism (approved):

PR #7600, opened 2021-02-24
[MU4] Update BravuraText to 1.393

Jojo-Schmitz:
Follow-up for #7378

Taken straight from https://github.com/steinbergmedia/bravura


Eism:
Just thoughts on third-party library updates
- We should consider using only release versions of the libraries because alpha and beta versions may contain bugs
- To switch to a new release in special cases, when, for example, a bug that we found has been fixed in the new version

Jojo-Schmitz:
Well, the SMuFL 1.4 Beta and the Bravura that goes along with it is in master already. And as we're pretty far away from releasing MuseScore 4, this is a good testbed, isn't it?

Jojo-Schmitz:
But we may leave this PR as a Draft for the time being, and as a reminder to update to SMuFL 1.4 + Bravura, once that gets released. I know that there are changes pending for either.

Jojo-Schmitz:
I believe though that the Bravura font is pretty much independant from the SMuFL version.

Jojo-Schmitz:
See https://github.com/steinbergmedia/bravura/commit/301087ca0b0d30b65d81bc3e718ff64b613e2a9a#commitcomment-48261819 which confirms this

Jojo-Schmitz:
SMuFL 1.4 has just been released and the part of it that had been merged with #7378, the metadata files, hasn't change since that PR.

PR #7601, opened 2021-02-24
[MU4] Fixed win core audio driver

igorkorsukov:


src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
```diff
@@ -45,93 +59,99 @@ std::string CoreAudioDriver::name() const
...
             logError(hr);
 
             auto bufferSize = bufferFrameCount - bufferPading;
-            hr = m_renderClient->GetBuffer(bufferSize, &pData);
+            hr = s_data->renderClient->GetBuffer(bufferSize, &pData);
             logError(hr);
             if (!pData || hr != S_OK) {
                 continue;
             }
-            activeSpec->callback(nullptr, reinterpret_cast<uint8_t*>(pData),
```
igorkorsukov, line 120: This is the first error - accessing this pointer in the thread, but this structure is an output argument and is a local variable, and after opening the driver, it will be destroyed, so an invalid pointer is being accessed in the thread.

vpereverzev (approved):

src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
```diff
@@ -45,93 +59,99 @@ std::string CoreAudioDriver::name() const
...
 
-    hr = m_audioClient->GetService(IID_IAudioRenderClient, (void**)&m_renderClient);
+    hr = s_data->audioClient->GetService(MU_IID_IAudioRenderClient, (void**)&s_data->renderClient);
     CHECK_HRESULT(hr);
 
-    auto hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+    s_data->hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
     CHECK_HRESULT(hr);
-    hr = m_audioClient->SetEventHandle(hEvent);
+    hr = s_data->audioClient->SetEventHandle(s_data->hEvent);
```
cbjeukendrup, line 121: I have a feeling that line 120 and 121 are in the wrong order, or isn't that the case?

igorkorsukov: Exactly, thanks. I'll fix it in some other PR.

PR #7602, opened 2021-02-25
[MU3] Fix #318035: Prevent (potential) crash on MusicXML import

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/318035

Prevents the crash reported in that forum topic and another potential one, by not continuing when there is no `measure` to work on.

This is for 3.x, for master see #7603  (it got merged there meanwhile)


Jojo-Schmitz:
@lvinken: mind to review?

lvinken:
Looks OK to me. Thanks for fixing this.

PR #7609, opened 2021-02-26
[MU4] Autotest tool prototype

igorkorsukov:
At the moment, it can open the score file for the list of scores, change the zoom, close the score.
So far, these are experiments to understand whether we should move in this direction.

https://user-images.githubusercontent.com/3818029/109325601-ca3a2800-785e-11eb-87ba-c7fc83f88bdd.mp4

### DIFF
--- a/src/appshell/qml/DevTools/Autobot/AutobotPanel.qml
+++ b/src/appshell/qml/DevTools/Autobot/AutobotPanel.qml
@@ -1,0 +1,30 @@
+import QtQuick 2.7
+import MuseScore.UiComponents 1.0
+import MuseScore.Autobot 1.0
+
+Rectangle {
+
+    color: "#edd400"
+
+    AutobotModel {
+        id: autobot
+    }
+
+    Row {
+        anchors.top: parent.top
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.margins: 32
+        height: contentRect.height
+
+        FlatButton {
+            text: "Run"
+            onClicked: autobot.run()
+        }
+
+        FlatButton {
+            text: "Stop"
+            onClicked: autobot.stop()
+        }
+    }
+}
--- a/src/autobot/internal/autobot.cpp
+++ b/src/autobot/internal/autobot.cpp
@@ -1,0 +1,80 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2021 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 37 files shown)


src/userscores/internal/userscoresservice.cpp
```diff
@@ -43,6 +43,7 @@ mu::ValCh<std::vector<Meta> > UserScoresService::recentScoreList() const
 
 std::vector<Meta> UserScoresService::parseRecentList(const QStringList& recentScoresPathList) const
 {
+    TRACEFUNC;
```
Marr11317, line 46: what is this for?

igorkorsukov: This is a macro for the built-in profiler.
Measures the execution time of functions
And we can display the report at any time like this

```
Main thread. Top 150 by sum time (total count: 6)
Function                                                    Call time         Call count        Sum time          
UserScoresService::recentScoreList                          7.656 ms          11                84.216 ms         
UserScoresService::parseRecentList                          7.626 ms          11                83.886 ms         
UserScoresConfiguration::recentScoreList                    0.018 ms          12                0.221 ms          
UiConfiguration::setPageState                               0.032 ms          4                 0.130 ms          
NotationPaintView::load                                     0.027 ms          2                 0.055 ms          
UiConfiguration::pageState                                  0.018 ms          1                 0.018 ms   
```

We can see which functions are called too often (for example, now there is some problem with `recentScoreList`, called too often). And we can see how long the functions take to execute, in order to decide whether to optimize them or not.

src/autobot/internal/autobot.cpp
```diff
@@ -0,0 +1,80 @@
...
+    m_currentIndex = -1;
+    m_scores = scores.val;
+    m_running = true;
+
+    nextScore();
+}
+
+void Autobot::stop()
+{
+    m_running = false;
```
Eism, line 64: I think you missed mutex for this variable

src/autobot/qml/MuseScore/Autobot/Autobot.qml
```diff
@@ -0,0 +1,8 @@
+import QtQuick 2.7
```
Eism, line 1: this file not needed more?

src/appshell/qml/DevTools/Autobot/AutobotPanel.qml
```diff
@@ -0,0 +1,30 @@
+import QtQuick 2.7
```
Eism, line 1: it is better to move this file to autobot module

src/autobot/internal/autobot.cpp
igorkorsukov: At the moment no threads here :)

src/autobot/qml/MuseScore/Autobot/Autobot.qml
igorkorsukov: Yes, but I don't know yet how everything will develop. Therefore, let it be, it will be seen further.

PR #7613, opened 2021-03-01
[MU4] Implemented about dialogs

Eism:
image

image


src/appshell/qml/AboutDialog.qml
```diff
@@ -0,0 +1,119 @@
...
+                            onClicked: {
+                                aboutModel.copyRevisionToClipboard()
+                            }
+                        }
+                    }
+                }
+
+                StyledTextLabel {
+                    Layout.fillWidth: true
+                    text: {
```
RomanPudashkin, line 78: I think it is better to get this text from cpp model

src/appshell/qml/AboutDialog.qml
```diff
@@ -0,0 +1,119 @@
...
+                Layout.bottomMargin: 16
+
+                spacing: 12
+
+                FlatButton {
+                    text: qsTrc("global", "OK")
+
+                    onClicked: {
+                        root.ret = { errcode: 0 }
+                        root.hide()
```
RomanPudashkin, line 113: You can use root.hide without this line
root.ret = { errcode: 0 }

src/appshell/qml/AboutDialog.qml
```diff
@@ -0,0 +1,119 @@
...
+                        font: ui.theme.bodyBoldFont
+                    }
+
+                    Row {
+                        spacing: 4
+                        anchors.horizontalCenter: parent.horizontalCenter
+
+                        StyledTextLabel {
+                            anchors.verticalCenter: parent.verticalCenter
+                            text: qsTrc("appshell", "Revision: ") + aboutModel.museScoreRevision()
```
RomanPudashkin, line 62: I think it is better to get this text from cpp model

RomanPudashkin (approved):

PR #7616, opened 2021-03-01
[MU3] Fix #317227, fix #317185: Disable centering score when opening score or switching LayoutMode

cbjeukendrup:
Resolves: https://musescore.org/en/node/317227, https://musescore.org/en/node/317185

Centering the score was maybe esthetically a nice idea, but in practice, it turns out to be less ergonomic / efficient. Therefore, it's now disabled; the score will be loaded with a bit of padding in the top left corner, like it was in 3.5.2 when you switched to another LayoutMode.
The way in which centering the score on opening was implemented also caused https://musescore.org/en/node/317185.

Schermafbeelding 2021-03-01 om 16 23 15

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -2661,10 +2661,6 @@ void MuseScore::setCurrentScoreView(ScoreView* view)
                   }
             cs = cv->score();
             cv->setFocusRect();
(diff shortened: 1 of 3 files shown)


Jojo-Schmitz:

PR #7617, opened 2021-03-01
[DAISY] Fix #298147 MusicXML: Handle measure number offsets

shoogle:
Resolves: https://musescore.org/en/node/298147

Previously it was assumed that measure numbers always started at 1 and increased by 1 for each regular measure. Now the actual measure numbers given in the MusicXML are used within MuseScore during import to calculate measure number offsets. During export, these offsets are converted back into absolute numbers for use in the MusicXML.

To-do:

- [x] Check export works as expected
- [x] Check import works as expected
- [x] Add tests and reference files

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -6621,6 +6621,7 @@ void MeasureNumberStateHandler::updateForMeasure(const Measure* const m)
     }
 
     // update measure numbers and cache result
+    _measureNo += m->noOffset();
     _cachedAttributes = " number=";
     if ((_irregularMeasureNo + _measureNo) == 2 && m->irregular()) {
         _cachedAttributes += "\"0\" implicit=\"yes\"";
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -1708,7 +1708,8 @@ void MusicXMLParserPass2::part()
 #endif
 
     // read the measures
-    int nr = 0;   // current measure sequence number
+    int nr = 0; // current measure sequence number (always increments by one for each measure)
+    _measureNumber = 0; // written measure number (doesn't always increment by 1)
     while (_e.readNextStartElement()) {
         if (_e.name() == "measure") {
             Fraction t = _pass1.getMeasureStart(nr);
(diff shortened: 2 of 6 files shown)


src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -2017,7 +2017,7 @@ static bool hasTempoTextAtTick(const TempoMap* const tempoMap, const int tick)
 void MusicXMLParserPass2::measure(const QString& partId, const Fraction time)
 {
     Q_ASSERT(_e.isStartElement() && _e.name() == "measure");
-    QString number = _e.attributes().value("number").toString();
+    int number = _e.attributes().value("number").toInt();
     //qDebug("measure %s start", qPrintable(number));
```
Jojo-Schmitz, line 2021: Now the `%s`in the `qDebug()` needs toi get changed to a `%d` I guess

RobFog:
What's "DAISY"?

Jojo-Schmitz:
Referring to daisy-chaining?

shoogle:
> What's "DAISY"?

The [DAISY Music Braille Project](https://daisy.org/activities/projects/music-braille/). I'm working with them on some accessibility and Braille features in MuseScore. One of those features is better MusicXML support for use with external Braille convertion tools.

Jojo-Schmitz:
I'd have a port of this for 3.x ready.

Jojo-Schmitz:
The backport to master in #7870 fails some mtests, esp. with measure numbers, so likely related.

Jojo-Schmitz:
The mtests here don't show it, but in my port to 3.x they do, see https://github.com/musescore/MuseScore/pull/7870/checks?check_run_id=2253440919, lots if issue with measure numbers (but also some more). Not sure though whether the tests need to get adjusted or the code?

Jojo-Schmitz:
@shoogle: Any plans to continue on this PR (and your other 2 DAISY ones)?

shoogle:
@Jojo-Schmitz, yes, thanks for the reminder. It looks like the MusicXML tests have been re-enabled now so I'll take another look at this next week.

Jojo-Schmitz:
any progress on this?

Jojo-Schmitz:
Now indeed a couple MusicXML import tests fail, in a similar way as mentioned [futher up](#issuecomment-828194813), but now also with some negative measure numbers.

src/importexport/musicxml/internal/musicxml/exportxml.cpp
```diff
@@ -6621,6 +6621,7 @@ void MeasureNumberStateHandler::updateForMeasure(const Measure* const m)
     }
 
     // update measure numbers and cache result
+    _measureNo += m->noOffset();
```
Jojo-Schmitz, line 6624: This seems to cause https://musescore.org/en/node/343581

Jojo-Schmitz, line 6624: This seems to cause https://musescore.org/en/node/343581, a pretty bad issue with screwed up actual measure durations

PR #7618, opened 2021-03-02
[MU3] Fix #318251: Compacting Section Break Inspector

Jojo-Schmitz:
It is way too wide (depending on translation too), and so easy to miss the input field for the Pause property

Resolves: https://musescore.org/en/node/318251
Does require 7 changed/new strings to get a) pushed up to Transifex and b) translated there. These won't mix and match nicely with the strings and translations of 3.6-3.6.2, unfortunately, but it also fixes a wrong accessibility text for one of the reset buttons.

Doesn't apply to master


Jojo-Schmitz:
Needs #7704

PR #7619, opened 2021-03-02
[MU4] Added qpainter provider

igorkorsukov:
see https://github.com/musescore/MuseScore/pull/7614

### DIFF
--- a/src/libmscore/draw/painter.cpp
+++ b/src/libmscore/draw/painter.cpp
@@ -20,247 +20,234 @@
 
 using namespace mu::draw;
 
-Painter::Painter(QPainter* painter)
-    : m_painter(painter)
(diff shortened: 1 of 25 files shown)


src/libmscore/draw/painter.cpp
```diff
@@ -20,274 +20,234 @@
 
 using namespace mu::draw;
 
-Painter::Painter(QPainter* painter)
-    : m_painter(painter)
+Painter::Painter(IPaintProviderPtr painter)
+    : m_p(painter)
```
RomanPudashkin, line 24: I think it is better to rename this member to m_provider

igorkorsukov: renamed

RomanPudashkin (approved):

PR #7620, opened 2021-03-02
[MU3] Add "countin" tool  to the toolbar in Advanced workspace

Jojo-Schmitz:
because there is no reason not to... (found this by accident in the MDL workspace)

As far as I can tell not needed in master

### DIFF
--- a/share/workspaces/Advanced.xml
+++ b/share/workspaces/Advanced.xml
@@ -3355,6 +3355,7 @@
       <action>repeat</action>
       <action>pan</action>
       <action>metronome</action>
+      <action>countin</action>
       </Toolbar>
     </Workspace>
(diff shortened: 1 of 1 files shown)


chenlung:
I would say “count-in” or “count in”.

Jojo-Schmitz:
That's not the name of that tool though, so just won't work

Jojo-Schmitz:
That's not the (internal!) name of that tool though, so just won't work. This PR is not adding any functionality, just making an existing one available in the toolbar (of the Advanced Workspace)

Jojo-Schmitz:
There are 3 more possible candidates: Note entry for 256th, 512th and 1024th notes.

PR #7621, opened 2021-03-04
[DAISY] Fix #269926 MusicXML: Fermatas on notes with grace notes

shoogle:
Resolves: https://musescore.org/en/node/269926

Previously, if a note had both a fermata and a grace note then the grace note would gain a fermata during export.

To-do:

- [x] Check export works as expected
- [ ] Check import works as expected (can't do this yet because import is currently disabled in MU4)
- [ ] Add tests and reference files


Jojo-Schmitz:
I'd have a port of this for 3.x ready.

Jojo-Schmitz:
The backport to master in #7870 fails some mtests, esp. with measure numbers, so probably not related to this particular PR, but those may serve as the lacking import test here in master

Jojo-Schmitz (approved):

shoogle:
Closed in favour of #13488

PR #7626, opened 2021-03-05
bug fixes in 2 files

zanaviska:
Resolves: always true bug and always odd bug

I have editted error with 2 files, which should make wrong behaviour

### DIFF
--- a/src/libmscore/lyrics.h
+++ b/src/libmscore/lyrics.h
@@ -82,7 +82,7 @@ public:
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return _no % 2; }
+    bool isEven() const { return !(_no % 2); }
     void setSyllabic(Syllabic s) { _syllabic = s; }
     Syllabic syllabic() const { return _syllabic; }
     void add(Element*) override;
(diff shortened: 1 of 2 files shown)


src/libmscore/lyrics.h
```diff
@@ -82,7 +82,7 @@ class Lyrics final : public TextBase
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return _no % 1; }
+    bool isEven() const { return _no % 2; }
```
Howard-C, line 85: I think it should be `!(_no % 2)`

zanaviska: oh, yes, or ```_no&1```, it's my error too

Howard-C:
Now please squash the two commits into one using `git rebase -i HEAD^^` :)

Howard-C:
And sign the [CLA](https://musescore.org/cla) please

src/libmscore/lyrics.h
Howard-C: You can use `!(_no & 1)` too, it's faster, but the speed probably doesn't matter...

zanaviska:
> Now please squash the two commits into one using `git rebase -i HEAD^^` :)

So I have to create new pull request?

Howard-C:
> > Now please squash the two commits into one using `git rebase -i HEAD^^` :)
> 
> So I have to create new pull request?

No need, just squash in this branch and use `git push -f` to force-push

Howard-C:
But in this case there're only a few lines of change, so sure, you can create another branch and another PR, it's probably easier.

PR #6021, opened 2020-05-02
[MU3] fix #303422: lyrics color is preserved when importing mxml + small bugfix and cleanup for isEven

SKefalidis:
Relates to: https://musescore.org/en/node/303422

I removed the `_even` variable which does not do anything other than creating a bug where there are lyrics that are neither odd nor even (as far as `Lyrics::layout` is concerned). Even `Layout::isEven` uses `_no` instead of `_even`.

This further breaks the behavior of the program when importing lyrics (since now no lyrics have the correct color when imported), but I would think that it is a step in the right direction :^).

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/musescore-coding-rules)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #7627, opened 2021-03-05
bug fixes

zanaviska:
Resolves always true and always odd bug

### DIFF
--- a/src/importexport/midiimport/internal/midiimport/importmidi_simplify.cpp
+++ b/src/importexport/midiimport/internal/midiimport/importmidi_simplify.cpp
@@ -41,7 +41,7 @@ bool areDurationsEqual(
         sum += ReducedFraction(d.second.fraction()) / d.first;
     }
 
-    return desiredLen == desiredLen;
+    return sum == desiredLen;
 }
 
 #endif
--- a/src/libmscore/lyrics.h
+++ b/src/libmscore/lyrics.h
@@ -82,7 +82,7 @@ public:
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return _no&1; }// _no is 0-based
+    bool isEven() const { return _no & 1; } // _no is 0-based
     void setSyllabic(Syllabic s) { _syllabic = s; }
     Syllabic syllabic() const { return _syllabic; }
     void add(Element*) override;
--- a/src/libmscore/lyrics.h
+++ b/src/libmscore/lyrics.h
@@ -82,7 +82,7 @@ public:
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return _no % 1; }
+    bool isEven() const { return !(_no % 2); }
     void setSyllabic(Syllabic s) { _syllabic = s; }
     Syllabic syllabic() const { return _syllabic; }
     void add(Element*) override;
--- a/src/libmscore/lyrics.h
+++ b/src/libmscore/lyrics.h
@@ -82,7 +82,7 @@ public:
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return !(_no % 2); }
+    bool isEven() const { return _no&1; }// _no is 0-based
     void setSyllabic(Syllabic s) { _syllabic = s; }
     Syllabic syllabic() const { return _syllabic; }
     void add(Element*) override;


src/importexport/midiimport/internal/midiimport/importmidi_simplify.cpp
```diff
@@ -41,7 +41,7 @@ bool areDurationsEqual(
         sum += ReducedFraction(d.second.fraction()) / d.first;
     }
 
-    return desiredLen == desiredLen;
+    return sum == desiredLen;
```
Jojo-Schmitz, line 44: See #7180, which does the exact same thing

Jojo-Schmitz: but needs a rebase, so I'll close mine.

src/libmscore/lyrics.h
```diff
@@ -82,7 +82,7 @@ class Lyrics final : public TextBase
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return _no % 1; }
+    bool isEven() const { return !(_no % 2); }
```
Jojo-Schmitz, line 85: Outch! That Mega Hertz!

Jojo-Schmitz (approved):

src/libmscore/lyrics.h
Jojo-Schmitz: This bug is in 3.x too BTW

src/importexport/midiimport/internal/midiimport/importmidi_simplify.cpp
Jojo-Schmitz: See also #7179 for 3.x, and the problems that went along with it, 2 mtests crashing

src/libmscore/lyrics.h
Jojo-Schmitz: See #7628

Jojo-Schmitz: Might (!?) resolve:

* https://musescore.org/en/node/9304
* https://musescore.org/en/node/285495
* https://musescore.org/en/node/288604

Jojo-Schmitz: That fix seems wrong though, `_no` is zero-based. and `0 % 2` is `0` and `!0` is true, but verse 1 is odd. So just drop that `!` or add a '  + 1` to `_no` (prior to the modulo operation).

Jojo-Schmitz: Another possible fix is to just replace the `%` with a `&`, that's a method used in other places.

Jojo-Schmitz, line 85: See above, changes needed

Jojo-Schmitz (changes requested):

src/importexport/midiimport/internal/midiimport/importmidi_simplify.cpp
Jojo-Schmitz, line 44: See also #7179 for 3.x, and the problems that went along with it, 2 mtests crashing (tests that are not enabled yet in master IIRC)

Jojo-Schmitz: See also https://musescore.org/en/node/314899, where this came up before

src/libmscore/lyrics.h
Howard-C: I don't understand why the fix is wrong, the function is `isEven()` so if `_no` is `0` it should return `true`, right?

Howard-C: I don't understand why the fix is wrong, the function is `isEven()` so if `_no` is `0` it should return `true`, right?

Jojo-Schmitz: No0 if `_no` is 0, that is verse 1 and 1 is odd

Jojo-Schmitz: No, if `_no` is 0, that is verse 1 and 1 is odd. So for 0 and 2 that should return false, for 1 and 3 true. Yes, that is quite off (pun intended)

Jojo-Schmitz: Hence my proposed fix with changing the `%` to `&`.

zanaviska: LMAO, than my first pull request was [correct](https://github.com/musescore/MuseScore/pull/7626#discussion_r588095924)

Howard-C: Oh I understand now. Yes, `_no & 1` is correct. And I believe it would've been the intention of the person who wrote this part of the code.

Howard-C: But I'd suggest adding a comment explaining why it should be the opposite of the apparent meaning.

Jojo-Schmitz: `// _no is 0-based`

src/libmscore/lyrics.h
```diff
@@ -82,7 +82,7 @@ class Lyrics final : public TextBase
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return !(_no % 2); }
+    bool isEven() const { return _no&1; }// _no is 0-based
```
Jojo-Schmitz, line 85: spaces around the `&` please (that's why the codestyle check failed)

src/importexport/midiimport/internal/midiimport/importmidi_simplify.cpp
Jojo-Schmitz: Try whether ther MIDI mtests / imports "voice_tuplet.mid" and "voice_3_5_7_tuplets.mid" work with this change (they crash with the corresponding change in 3.x, see #7179)

src/libmscore/lyrics.h
```diff
@@ -82,7 +82,7 @@ class Lyrics final : public TextBase
     QString subtypeName() const override { return QObject::tr("Verse %1").arg(_no + 1); }
     void setNo(int n) { _no = n; }
     int no() const { return _no; }
-    bool isEven() const { return _no&1; }// _no is 0-based
+    bool isEven() const { return _no & 1; } // _no is 0-based
```
Jojo-Schmitz, line 85: See also #6021 (for 3.x)

RomanPudashkin:
closed due to inactivity

PR #7630, opened 2021-03-05
[MU4] Solve a crash when dropping an element.

njvdberg:
Member <code>view</code> of <code>dropData</code> in <code>NotationInteraction::applyDropPaletteElement()</code> wasn't initialized. With this PR it is set to a pointer to <code>ScoreCallbacks()</code>.

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -1238,6 +1238,7 @@ void NotationInteraction::applyDropPaletteElement(Ms::Score* score, Ms::Element*
                                                   QPointF pt, bool pasteMode)
 {
     Ms::EditData dropData;
+    dropData.view        = new ScoreCallbacks();
     dropData.pos         = pt.isNull() ? target->pagePos() : pt;
     dropData.dragOffset  = QPointF();
     dropData.modifiers   = modifiers;


src/notation/internal/notationinteraction.cpp
```diff
@@ -1238,6 +1238,7 @@ void NotationInteraction::applyDropPaletteElement(Ms::Score* score, Ms::Element*
                                                   QPointF pt, bool pasteMode)
 {
     Ms::EditData dropData;
+    dropData.view        = new ScoreCallbacks();
```
RomanPudashkin, line 1241: I can't find a place where we delete this resource... Maybe you know where is this place located?

njvdberg: I didn't see it either, I just used the same construct as seen at several other places. So, if this is "wrong" it is wrong at other places too. But maybe a good opportunity to look into that too.
PR will be updated.

njvdberg: Due to the nature of <code>ScoreCallbacks</code> (multiple inheritance of <code>MuseScoreView</code> and <code>MuseScoreCore</code>) it is not possible to delete it because <code>MuseScoreCore</code> is a singleton.
In fact, the concept of having a class like <code>ScoreCallbacks</code> seems problematic too me.
@igorkorsukov Can you explain what the purpose of <code>ScoreCallbacks</code> and how a multiple inheritance of both a "normal" class and a singleton is supposed to work?

RomanPudashkin: Thanks for explaining. I will merge this PR (because this crash can infer to development) and create an issue about those memory leaks

RomanPudashkin (approved):

src/notation/internal/notationinteraction.cpp
RomanPudashkin: https://github.com/musescore/MuseScore/issues/7706

igorkorsukov: > Due to the nature of `ScoreCallbacks` (multiple inheritance of `MuseScoreView` and `MuseScoreCore`) it is not possible to delete it because `MuseScoreCore` is a singleton.
> In fact, the concept of having a class like `ScoreCallbacks` seems problematic too me.
> @igorkorsukov Can you explain what the purpose of `ScoreCallbacks` and how a multiple inheritance of both a "normal" class and a singleton is supposed to work?

Previously, in the MU3, instead of `ScoreCallbacks`, a pointer to the current view was passed.
In the MU4 architecture, `libmscore` does not need to "know" about the view, so I created this class as a temporary solution.
We will investigate in more detail for what purposes it is used and decide how best to do it.

PR #7635, opened 2021-03-05
[MU4] General menu optimization

Eism:


src/appshell/view/appmenumodel.cpp
```diff
@@ -229,36 +226,52 @@ MenuItem AppMenuModel::editItem()
...
+                   [this]() { return m_viewMenuController->isPanelAvailable(PanelType::Palette); }),
+        makeAction("toggle-instruments", isPanelVisible(PanelType::Instruments),
+                   [this]() { return m_viewMenuController->isPanelAvailable(PanelType::Instruments); }),
+        makeAction("masterpalette", paletteController()->isMasterPaletteOpened().val,
+                   [this]() { return m_viewMenuController->isMasterPaletteAvailable(); }),
+        makeAction("inspector", isPanelVisible(PanelType::Inspector),
+                   [this]() { return m_viewMenuController->isPanelAvailable(PanelType::Inspector); }),
+        makeAction("toggle-navigator", isPanelVisible(PanelType::NotationNavigator),
+                   [this]() { return m_viewMenuController->isPanelAvailable(PanelType::NotationNavigator); }),
+        makeAction("toggle-timeline", isPanelVisible(PanelType::TimeLine),
```
RomanPudashkin, line 239: I think, we can simplify getting of availability of a menu item. You already have an abstract interface for this - IActionsAvailability. We can move this check inside makeAction:

makeAction(actionCode) {
...
item.enabled = resolveController(actionCode)->isActionAvailable(actionCode)
....
}

Eism: We can not resolve controller in makeAction by action code because makeAction do not know about parent of action(File, View, etc)

src/appshell/view/editmenucontroller.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+mu::async::Channel<std::vector<mu::actions::ActionCode> > EditMenuController::actionsAvailableChanged() const
+{
+    return m_actionsReceiveAvailableChanged;
+}
+
+std::map<mu::actions::ActionCode, EditMenuController::AvailableCallback> EditMenuController::actions() const
+{
+    static std::map<ActionCode, AvailableCallback > _actions = {
```
RomanPudashkin, line 58: Seems that you can simplify this method (same for similar methods in other controllers). It can return just a list of action codes and EditMenuController::actionAvailable will looks like:

actionAvailable(actionCode) {
return controller()->actionAvailable(actionCode);
}

src/framework/uicomponents/view/abstractmenumodel.h
```diff
@@ -0,0 +1,78 @@
...
+    MenuItem makeMenu(const std::string& title, const MenuItemList& actions, bool enabled = true,
+                      const actions::ActionCode& menuActionCode = "") const;
+    MenuItem makeAction(const actions::ActionCode& actionCode) const;
+    MenuItem makeAction(const actions::ActionCode& actionCode, bool checked) const;
+    MenuItem makeSeparator() const;
+
+    void updateItemsEnabled(const std::vector<actions::ActionCode>& actionCodes);
+
+private:
+    static bool enabledTrue() { return true; }
```
RomanPudashkin, line 69: Seems that you don't use this method

src/notation/internal/addmenucontroller.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+std::map<ActionCode, AddMenuController::AvailableCallback > AddMenuController::actions() const
+{
+    static std::map<ActionCode, AvailableCallback > _actions = {
+        { "note-input", std::bind(&AddMenuController::isNoteInputAvailable, this) },
+        { "note-c", std::bind(&AddMenuController::isNoteAvailable, this, NoteName::C, NoteAddingMode::NextChord) },
+        { "note-d", std::bind(&AddMenuController::isNoteAvailable, this, NoteName::D, NoteAddingMode::NextChord) },
+        { "note-e", std::bind(&AddMenuController::isNoteAvailable, this, NoteName::E, NoteAddingMode::NextChord) },
+        { "note-f", std::bind(&AddMenuController::isNoteAvailable, this, NoteName::F, NoteAddingMode::NextChord) },
+        { "note-g", std::bind(&AddMenuController::isNoteAvailable, this, NoteName::G, NoteAddingMode::NextChord) },
```
RomanPudashkin, line 61: You can use cycles for similar actions like note-* and interval-*

src/appshell/ieditmenucontroller.h
```diff
@@ -0,0 +1,32 @@
...
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#ifndef MU_APPSHELL_IEDITMENUCONTROLLER_H
+#define MU_APPSHELL_IEDITMENUCONTROLLER_H
+
+#include "modularity/imoduleexport.h"
+#include "uicomponents/imenucontroller.h"
+
+namespace mu::appshell {
+class IEditMenuController : public uicomponents::IMenuController, MODULE_EXPORT_INTERFACE
```
RomanPudashkin, line 26: You can create a register for IMenuController instead of creating many empty interfaces and use it:
изображение

src/appshell/view/appmenumodel.cpp
```diff
@@ -86,185 +108,179 @@ INotationInteractionPtr AppMenuModel::notationInteraction() const
...
-        makeAction("toggle-mixer", isNotationPage()), // need implement
-        makeAction("synth-control", isNotationPage()), // need implement
-        makeAction("toggle-selection-window", isNotationPage()), // need implement
-        makeAction("toggle-piano", isNotationPage()), // need implement
-        makeAction("toggle-scorecmp-tool", isNotationPage()), // need implement
+        makeAction("toggle-palette", isPanelVisible(PanelType::Palette)),
+        makeAction("toggle-instruments", isPanelVisible(PanelType::Instruments)),
+        makeAction("masterpalette", paletteController()->isMasterPaletteOpened().val),
+        makeAction("inspector", isPanelVisible(PanelType::Inspector)),
+        makeAction("toggle-navigator", isPanelVisible(PanelType::NotationNavigator)),
```
RomanPudashkin, line 253: Seems that you can add a method to IMenuController:

actionActivated(actionCode)

The goal of this method to tell us that user activated this action (you can pick a more proper name for it). So, makeAction will look like this:

makeAction(actionCode) {
...
item.enabled = resolveController(actionCode)->actionEnabled(actionCode)
item.checked = resolveController(actionCode)->actionActivated(actionCode)
}

src/appshell/view/editmenucontroller.cpp
Eism: Each method can contain additional logic to determine the availability of an action

src/notation/notationtypes.h
```diff
@@ -427,6 +468,123 @@ inline bool isVoiceIndexValid(int voiceIndex)
 {
     return 0 <= voiceIndex && voiceIndex < VOICES;
 }
+
+inline std::string noteNameToString(NoteName noteName)
+{
+    std::map<NoteName, std::string> noteNameStrings {
```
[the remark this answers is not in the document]
Eism: why better?

src/notation/internal/addmenucontroller.cpp
Eism: it will be difficult to pass parameters in cycles

src/appshell/view/editmenucontroller.cpp
RomanPudashkin: There are currently no checks that contain any additional logic. If we will want to add those checks, we can do this without using any callbacks, just like this:

actionAvailable(actionCode) {
if (actionCode == "codeWithSomeCustomLogic") {
return customCheck()
}

return controller()->actionAvailable(actionCode);
}

Eism: what happens if half have one logic and the other another? do we list all the code here through if?

RomanPudashkin: To be honest, I still doubt that we will ever need a custom check for any code. If such a situation arises, then most likely there will be a small number of such codes. But at the moment we should use std::bind and callbacks for any new action code (in all menu controllers) that we plan to add to the action list. This code is hard to read and it requires many manual operations when you want to add some new action

src/notation/notationtypes.h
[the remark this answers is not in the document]
RomanPudashkin: If we will add some new value to NoteName then we will get a warning and we will not forget to add this code here

src/appshell/view/editmenucontroller.cpp
Eism: maybe, but it necessary
Next example
Now some menu controllers work with two module controllers(see FileMenuController work with controllers from userscores and notation)

RomanPudashkin: You can create two separate lists for those actions (or create an enumerator and use it instead of a large bunch of private methods that are just delegated calls to the controller)

RomanPudashkin (approved):

PR #7698, opened 2021-03-06
Fix #318388: Crash when trying to attach an image to key signature

mattmcclinch:
Resolves: https://musescore.org/en/node/318388.

`target->drop(data)` should never be called unless `target->acceptDrop(data)` returns `true`.


vpereverzev (approved):

Jojo-Schmitz:
I believe I see my screw-up (but that #7699 is not part of #9000, it is part of the 33.x branch which #9000 uses as its basis).

(If it isn't https://github.com/musescore/MuseScore/pull/7699/files#r822351432, let me know)

Jojo-Schmitz:
(If it isn't https://github.com/musescore/MuseScore/pull/7699/files#r822351432, let me know)

PR #7699, opened 2021-03-06
[MU3] Image paste crashes

Jojo-Schmitz:
Resolves: 

* [Fix #318388](https://musescore.org/en/node/318388): Crash when trying to attach an image to key signature (3.x counterpart to #7698)
* [Fix #318370](https://musescore.org/en/node/318370): crash when pasting an image onto a (title) text (may not be needed after the above fix, but won't harm)

### DIFF
--- a/libmscore/paste.cpp
+++ b/libmscore/paste.cpp
@@ -1125,10 +1125,12 @@ void Score::cmdPaste(const QMimeData* ms, MuseScoreView* view, Fraction scale)
                   addRefresh(target->abbox());   // layout() ?!
                   EditData ddata(view);
                   ddata.view       = view;
-                  ddata.dropElement    = nel;
-                  target->drop(ddata);
-                  if (_selection.element())
-                        addRefresh(_selection.element()->abbox());
+                  if (target->acceptDrop(ddata)) {
+                        ddata.dropElement    = nel;
+                        target->drop(ddata);
+                        if (_selection.element())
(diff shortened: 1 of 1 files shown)


libmscore/textbase.cpp
```diff
@@ -10,6 +10,8 @@
 //  the file LICENCE.GPL
 //=============================================================================
 
+#include "global/log.h"
```
vpereverzev, line 13: I suggest to not include it here, otherwise we can easily miss this during the port of the changes

in short, we're going to reduce external dependencies from "libmscore" in master

Jojo-Schmitz: OK, removed this. Reduces this PR to a port of @mattmcclinch's fix for master to 3.x

libmscore/paste.cpp
```diff
@@ -1125,10 +1125,12 @@ void Score::cmdPaste(const QMimeData* ms, MuseScoreView* view, Fraction scale)
                   addRefresh(target->abbox());   // layout() ?!
                   EditData ddata(view);
                   ddata.view       = view;
-                  ddata.dropElement    = nel;
-                  target->drop(ddata);
-                  if (_selection.element())
-                        addRefresh(_selection.element()->abbox());
+                  if (target->acceptDrop(ddata)) {
+                        ddata.dropElement    = nel;
```
Jojo-Schmitz, line 1129: Dang, this should have stayed before the `if`

PR #7705, opened 2021-03-08
[MU3] Fix #318534: Missing muted channels in several templates' Horns and Trombones

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/318534

That update reveals that range and articulations were not matching too.

For master see #9102


Jojo-Schmitz:
@njvdberg please review

Jojo-Schmitz:
Is a counterpart PR for master wanted/needed?

Jojo-Schmitz:
Here esp. @njvdberg is asked ;-)

PR #7707, opened 2021-03-09
[MU4] Added buffered paint provider

igorkorsukov:
see also https://github.com/musescore/MuseScore/pull/7614

### DIFF
--- a/src/libmscore/draw/bufferedpaintprovider.cpp
+++ b/src/libmscore/draw/bufferedpaintprovider.cpp
@@ -1,0 +1,292 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2021 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#include "bufferedpaintprovider.h"
+
+using namespace mu::draw;
+
+BufferedPaintProvider::BufferedPaintProvider()
+{
+    //! NOTE Make data with default state
+    m_buf.datas.push_back(DrawBuffer::Data());
+
+    m_isActive = true;
+}
+
+QPaintDevice* BufferedPaintProvider::device() const
+{
+    return nullptr;
+}
+
+QPainter* BufferedPaintProvider::qpainter() const
+{
+    return nullptr;
+}
+
+bool BufferedPaintProvider::end()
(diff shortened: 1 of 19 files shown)


src/libmscore/draw/bufferedpaintprovider.h
```diff
@@ -0,0 +1,154 @@
...
+
+#include <QPen>
+#include <QBrush>
+#include <QFont>
+
+#include "ipaintprovider.h"
+#include "drawtypes.h"
+
+namespace mu::draw {
+struct DrawBuffer
```
RomanPudashkin, line 32: I think we can move this struct to a private part of BufferedPaintProvider

src/libmscore/draw/bufferedpaintprovider.h
```diff
@@ -0,0 +1,154 @@
...
+
+    void drawText(const QPointF& p, const QString& s) override;
+    void drawText(const QRectF& r, int flags, const QString& text) override;
+
+    void drawGlyphRun(const QPointF& position, const QGlyphRun& glyphRun) override;
+
+    void fillRect(const QRectF& r, const QBrush& brush) override;
+
+    void drawPixmap(const QPointF& p, const QPixmap& pm) override;
+    void drawTiledPixmap(const QRectF& rect, const QPixmap& pm, const QPointF& offset = QPointF()) override;
```
RomanPudashkin, line 140: We use different namings in different methods for rects (r vs rect). I think it is better to use "rect" in all places

src/libmscore/draw/painter.cpp
```diff
@@ -163,91 +227,158 @@ QRect Painter::viewport() const
...
 
-void Painter::drawArc(const QRectF& rect, int a, int alen)
+void Painter::drawArc(const QRectF& r, int a, int alen)
 {
-    m_provider->drawArc(rect, a, alen);
+    QRectF rect = r.normalized();
+
+    QPainterPath path;
+    path.arcMoveTo(rect, a / 16.0);
+    path.arcTo(rect, a / 16.0, alen / 16.0);
```
RomanPudashkin, line 322: Can you explain please what this number means?

src/libmscore/draw/bufferedpaintprovider.cpp
```diff
@@ -0,0 +1,286 @@
...
+    for (int i = 1; i < pointCount; ++i) {
+        QPointF p1 = points[i - 1];
+        QPointF p2 = points[i];
+        d.lines.push_back(QLineF(p1, p2));
+    }
+}
+
+void BufferedPaintProvider::drawPolygon(const QPointF* points, int pointCount, Qt::FillRule fillRule)
+{
+    QVector<QPointF> vec(pointCount);
```
RomanPudashkin, line 242: We can use QPolygonF instead of this

RomanPudashkin (approved):

src/libmscore/draw/bufferedpaintprovider.h
igorkorsukov: fixed

src/libmscore/draw/painter.cpp
igorkorsukov: This is copied from the QPainter source code.
Added a comment about it.

src/libmscore/draw/bufferedpaintprovider.cpp
igorkorsukov: fixed

PR #7710, opened 2021-03-09
[MU4] Added draw objects trace

igorkorsukov:
See also https://github.com/musescore/MuseScore/pull/7614

### DIFF
--- a/src/libmscore/draw/painter.cpp
+++ b/src/libmscore/draw/painter.cpp
@@ -60,7 +60,6 @@ Painter::Painter(QPainter* qp, const std::string& name, bool overship)
 
 Painter::~Painter()
 {
-    end();
 }
 
 QPaintDevice* Painter::device() const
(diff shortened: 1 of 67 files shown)


src/libmscore/draw/painter.h
```diff
@@ -229,6 +233,37 @@ inline void Painter::drawText(int x, int y, const QString& text)
...
+
+#ifdef TRACE_OBJ_DRAW_ENABLED
+    #define TRACE_OBJ_DRAW \
+    mu::draw::PainterObjMarker __drawObjMarker(painter, name(), pagePos())
+
+    #define TRACE_OBJ_DRAW_C(painter, objName, objPagePos) \
+    mu::draw::PainterObjMarker __drawObjMarker(painter, objName, objPagePos)
+#else
+    #define TRACE_OBJ_DRAW
+    #define TRACE_OBJ_DRAW_C(painter, objName)
```
cbjeukendrup, line 265: Isn't `, objPagePos` missing here?

src/libmscore/draw/painter.h
```diff
@@ -229,6 +233,37 @@ inline void Painter::drawText(int x, int y, const QString& text)
...
     drawText(QPointF(x, y), text);
 }
+
+class PainterObjMarker
+{
+public:
+    PainterObjMarker(Painter* p, const std::string& name, const QPointF& objPagePos)
+        : m_painter(p), m_name(name), m_objPagePos(objPagePos)
+    {
+        p->beginObject(name, m_objPagePos);
```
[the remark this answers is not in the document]
igorkorsukov: There is no difference :) I did it consistently.

src/libmscore/draw/painter.h
igorkorsukov: fixed

src/libmscore/draw/painter.cpp
```diff
@@ -75,17 +74,34 @@ QPainter* Painter::qpainter() const
 
 bool Painter::end()
 {
-    return m_provider->end(m_name);
+    bool ok = m_provider->end(m_name);
     if (extended) {
         extended->end(m_name);
```
RomanPudashkin, line 79: Should we check the result here too?

igorkorsukov: > Should we check the result here too?

There is no need. "Extended" is intended primarily for automated testing. In the near future, I will show what I have planned to do :)

PR #7711, opened 2021-03-09
[MU3] Fix typo in a the name of a style setting

Jojo-Schmitz:
For master see #7712

### DIFF
--- a/libmscore/style.cpp
+++ b/libmscore/style.cpp
@@ -434,7 +434,7 @@ static const StyleType styleTypes[] {
       { Sid::mmRestNumberPos,         "mmRestNumberPos",         Spatium(-1.5) },
       { Sid::hideEmptyStaves,         "hideEmptyStaves",         QVariant(false) },
       { Sid::dontHideStavesInFirstSystem,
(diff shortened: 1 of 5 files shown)


Jojo-Schmitz:
I'm slightly puzzled that 3.x needs these changes to vtests, but for master it does not?

libmscore/style.cpp
```diff
@@ -434,7 +434,7 @@ static const StyleType styleTypes[] {
       { Sid::mmRestNumberPos,         "mmRestNumberPos",         Spatium(-1.5) },
       { Sid::hideEmptyStaves,         "hideEmptyStaves",         QVariant(false) },
       { Sid::dontHideStavesInFirstSystem,
-                                 "dontHidStavesInFirstSystm",    QVariant(true) },
+                                 "dontHideStavesInFirstSystem",  QVariant(true) },
```
njvdberg, line 437: Does this name change cause compatibility issues with existing scores?

Jojo-Schmitz: It may (read: will) cause downrev issues though.

njvdberg: Ah, I missed that one.

PR #7713, opened 2021-03-10
[MU4] Enable vertical staves adjustment

njvdberg:
This PR does actually 2 things:

 1. In both <code>EditStyle::enableVerticalSpreadClicked()</code> and >code>EditStyle::disableVerticalSpreadClicked()</code> the line <code> cs->setLayoutAll(); // TODO</code> is removed since this isn't necessary. <code>setLayoutAll()</code> is already called in <code>setStyleValue()</code>.
 
 2. Solves a potential crash because the test <code>if (!mb->isMeasure())</code> was at the wrong place. 

### DIFF
--- a/src/libmscore/layout.cpp
+++ b/src/libmscore/layout.cpp
@@ -1700,10 +1700,10 @@ static void distributeStaves(Page* page)
                 Spacer* activeSpacer { nextSpacer };
                 nextSpacer = nullptr;
                 for (MeasureBase* mb : system->measures()) {
-                    Measure* m = toMeasure(mb);
-                    if (!mb->isMeasure()) {
+                    if (!(mb && mb->isMeasure())) {
                         continue;
                     }
+                    Measure* m = toMeasure(mb);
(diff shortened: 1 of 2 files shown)


src/libmscore/layout.cpp
```diff
@@ -1700,10 +1700,10 @@ static void distributeStaves(Page* page)
                 Spacer* activeSpacer { nextSpacer };
                 nextSpacer = nullptr;
                 for (MeasureBase* mb : system->measures()) {
-                    Measure* m = toMeasure(mb);
-                    if (!mb->isMeasure()) {
+                    if (!(mb && mb->isMeasure())) {
                         continue;
                     }
+                    Measure* m = toMeasure(mb);
```
RomanPudashkin, line 1706: Maybe is it better to add a check on null?

njvdberg: That's done already in line 1703. `toMeasure()` is actually just a cast from `Element` to `Measure` and the test two lines ahead make sure `m` isn't a `nullptr`.

PR #7721, opened 2021-03-11
[MU4] Reduced paint provider interface

igorkorsukov:
See also https://github.com/musescore/MuseScore/pull/7614


src/autobot/autobotmodule.cpp
```diff
@@ -37,6 +40,8 @@ std::string AutobotModule::moduleName() const
 void AutobotModule::registerExports()
 {
     framework::ioc()->registerExport<IAutobot>(moduleName(), s_autobot);
+
+    //draw::Painter::extended = AbPaintProvider::instance();
```
RomanPudashkin, line 44: Better to remove this code or add some TODO

src/autobot/internal/draw/abpaintprovider.cpp
```diff
@@ -0,0 +1,50 @@
...
+//=============================================================================
+#include "abpaintprovider.h"
+
+#include "log.h"
+
+using namespace mu::autobot;
+
+const std::shared_ptr<AbPaintProvider>& AbPaintProvider::instance()
+{
+    static std::shared_ptr<AbPaintProvider> p = std::shared_ptr<AbPaintProvider>(new AbPaintProvider());
```
RomanPudashkin, line 27: You can use std::make_shared instead of new

src/autobot/internal/draw/abpaintprovider.h
```diff
@@ -0,0 +1,42 @@
...
+
+    static const std::shared_ptr<AbPaintProvider>& instance();
+
+    void beginTarget(const std::string& name) override;
+    bool endTarget(const std::string& name, bool endDraw = false) override;
+
+    void serialize(std::string& out);
+
+private:
+    AbPaintProvider();
```
RomanPudashkin, line 38: You can use = default operator:
AbPaintProvider() = default;

src/autobot/internal/draw/abpaintprovider.h
```diff
@@ -0,0 +1,42 @@
...
+class AbPaintProvider : public draw::BufferedPaintProvider
+{
+public:
+
+    static const std::shared_ptr<AbPaintProvider>& instance();
+
+    void beginTarget(const std::string& name) override;
+    bool endTarget(const std::string& name, bool endDraw = false) override;
+
+    void serialize(std::string& out);
```
RomanPudashkin, line 35: Maybe is it better to replace void with std::string instead of using reference in arguments?

std::string serialize()

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -42,7 +42,7 @@ NotationViewInputController::NotationViewInputController(IControlledView* view)
         dispatcher()->reg(this, "zoomin", this, &NotationViewInputController::zoomIn);
         dispatcher()->reg(this, "zoomout", this, &NotationViewInputController::zoomOut);
         dispatcher()->reg(this, "zoom-page-width", this, &NotationViewInputController::zoomToPageWidth);
-        dispatcher()->reg(this, "zoom100", [this]() { setZoom(100); });
```
RomanPudashkin, line 45: I transferred this action from MU3 so probably we should keep it

src/libmscore/draw/painter.h
```diff
@@ -158,23 +154,36 @@ class Painter
     static IPaintProviderPtr extended;
 
 private:
+
+    struct State {
+        QRect window;
+        QRect viewport;
+        bool isVxF = false;
```
RomanPudashkin, line 161: Can you explain please what this flag does?

src/autobot/autobotmodule.cpp
igorkorsukov: In the next PR I will continue to work on this code.

src/autobot/internal/draw/abpaintprovider.cpp
igorkorsukov: Not in this case, the `AbPaintProvider` constructor is private, so `make_shared` does not work.

src/autobot/internal/draw/abpaintprovider.h
igorkorsukov: I want to avoid copying a very large string

src/libmscore/draw/qpainterprovider.h
```diff
@@ -61,45 +61,18 @@ class QPainterProvider : public IPaintProvider
     void save() override;
     void restore() override;
 
-    void setWorldTransform(const QTransform& matrix, bool combine = false) override;
-    const QTransform& worldTransform() const override;
-
-    void setTransform(const QTransform& transform, bool combine = false) override;
+    void setTransform(const QTransform& matrix) override;
```
RomanPudashkin, line 64: Can you rename matrix to transform, please?

src/notation/view/notationviewinputcontroller.cpp
igorkorsukov: I have not found a place where it would be sent. If we need it, we can send `zoom-x-percent` with the argument 100

src/libmscore/draw/painter.h
igorkorsukov: This means that `viewTransform` is installed

src/libmscore/draw/qpainterprovider.h
igorkorsukov: renamed
(although transform and matrix are essentially the same :))

src/notation/view/notationviewinputcontroller.cpp
RomanPudashkin: It can be sent via shortcuts
image

src/libmscore/draw/painter.h
RomanPudashkin: Then it would be great to rename this flag to "isViewTransformInstalled"

src/autobot/internal/draw/abpaintprovider.h
igorkorsukov: I can do it, I can not do it :)

src/notation/view/notationviewinputcontroller.cpp
igorkorsukov: Returned action `zoom100`. Thanks!

src/libmscore/draw/painter.h
igorkorsukov: I took this solution from the Qt sources.
I had an idea not to stray too far from them.

RomanPudashkin (approved):

PR #7724, opened 2021-03-12
[MU4] Added draw buffer serialization to json

igorkorsukov:
see also https://github.com/musescore/MuseScore/pull/7614

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -31,6 +31,8 @@ DockPage {
         pageModel.isNoteInputBarVisible = notationNoteInputBar.visible
         notationNoteInputBar.visible = Qt.binding(function() { return pageModel.isNoteInputBarVisible })
 
+        autobotPanel.visible = Qt.binding(function() { return pageModel.isInspectorPanelVisible })
+
         pageModel.init()
     }
 
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -31,8 +31,6 @@ DockPage {
         pageModel.isNoteInputBarVisible = notationNoteInputBar.visible
         notationNoteInputBar.visible = Qt.binding(function() { return pageModel.isNoteInputBarVisible })
 
-        autobotPanel.visible = Qt.binding(function() { return pageModel.isInspectorPanelVisible })
-
         pageModel.init()
     }
 
@@ -149,6 +147,7 @@ DockPage {
             area: Qt.RightDockWidgetArea
             floatable: true
             closable: true
+            visible: false
 
             Loader {
                 anchors.fill: parent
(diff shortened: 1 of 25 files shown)


src/appshell/qml/NotationPage/NotationPage.qml
```diff
@@ -31,6 +31,8 @@ DockPage {
         pageModel.isNoteInputBarVisible = notationNoteInputBar.visible
         notationNoteInputBar.visible = Qt.binding(function() { return pageModel.isNoteInputBarVisible })
 
+        autobotPanel.visible = Qt.binding(function() { return pageModel.isInspectorPanelVisible })
```
RomanPudashkin, line 34: >isInspectorPanelVisible

Seems this is a typo

src/autobot/internal/steps/abdrawserializationstep.cpp
```diff
@@ -18,15 +18,37 @@
...
+            LOGE() << "failed make path: " << drawDataPath;
+            doFinish(ctx);
+            return;
+        }
+    }
+
+    QByteArray data = AbPaintProvider::instance()->serialize();
+
+    io::path scorePath = ctx.val<io::path>(AbContext::Key::ScoreFile);
+    io::path filePath = drawDataPath + "/" + io::basename(scorePath) + ".json";
```
RomanPudashkin, line 43: I think it is better to move this code to the configuration

src/autobot/internal/draw/abpaintprovider.cpp
```diff
@@ -38,14 +39,15 @@ void AbPaintProvider::beginTarget(const std::string& name)
...
 }
 
-bool AbPaintProvider::endTarget(const std::string& name, bool endDraw)
+bool AbPaintProvider::endTarget(bool endDraw)
 {
-    UNUSED(endDraw);
-    LOGI() << name;
-    return BufferedPaintProvider::endTarget(name, endDraw);
+    bool ok = BufferedPaintProvider::endTarget(endDraw);
+    if (ok && buffer().name == "notationview") {
```
RomanPudashkin, line 45: Please explain why you need this check?

src/autobot/internal/steps/abdrawserializationstep.cpp
igorkorsukov: I don't think so, this is the path for the current score data, depends on the current context

src/autobot/internal/draw/abpaintprovider.cpp
igorkorsukov: In normal, one paint provider instance for one painter.
In this test, one paint provider instance for all painters (that is, in fact, it is a singleton). This name contains what is currently being painted (there may also be palettes, export, etc.)
In this case, we need only notation view painting

src/appshell/qml/NotationPage/NotationPage.qml
igorkorsukov: removed

RomanPudashkin (approved):

PR #7728, opened 2021-03-14
[MU3] fix #89216: multiple possible causes of crashes or audible artefacts

mirabilos:
Resolves: https://bugs.debian.org/985129 and https://musescore.org/en/node/89216 and probably others

 - Track sample name so we can issue proper warning messages, show filename
 - On read errors, issue an error message and mark sample as invalid
 - Mark sample as invalid if Ogg Vorbis decompression (SF3) fails
 - Do all sanity checks on {,loop}{start,end} with SF2 semantics for end;
   only switch end to point to the last sample afterwards in only one place
 - Add sanity check provided by the SoundFont spec as extra warning
 - Do not crash if there is no data[]
 - Issue diagnostics if disabling a sample
 - Swap two members to improve structure packing/alignment while there
 - Use unsigned integers for SoundFont element sizes properly

I’ve tested this patch so far against MuseScore 2.3.2 and am currently testing it against 3.2.3 as well. I’ve added `+ collect_artifacts` so @Jojo-Schmitz can test 3.x HEAD ☺

With this patch enabled, multiple issues in MuseScore_General_HQ are found (and also in the nōn-HQ soundfont). I’ve already sent @mrbumpy409 an eMail regarding these, so you can ignore them.

This change _basically_ reverts commit 5aea0ab5b50b9037ba8ed9dd100dcf06396a4c9d as I found the root cause**s** for that. I did not do a full revert because I’m unsure if the `log.h` stuff is used elsewhere; if not, please feel free to g/c _that_ as well.

### DIFF
--- a/audio/midi/fluid/sfont.cpp
+++ b/audio/midi/fluid/sfont.cpp
@@ -614,6 +614,7 @@ Sample::Sample(SFont* s)
       data        = 0;
       amplitude_that_reaches_noise_floor_is_valid = false;
       amplitude_that_reaches_noise_floor = 0.0;
+      name[0]     = 0;
       }
 
 //---------------------------------------------------------
@@ -651,18 +652,22 @@ void Sample::load()
             std::vector<char> p;
             p.resize(size);
             if (fd.read(p.data(), size) != size) {
-                  qDebug("read %d failed", size);
+                  qWarning("SoundFont(%s) Sample(%s) read %u bytes failed", qPrintable(sf->get_name()), name, size);
+                  setValid(false);
                   return;
                   }
-            decompressOggVorbis(p.data(), size);
+            setValid(decompressOggVorbis(p.data(), size));
 #endif
             }
       else {
             data = new short[size];
             size *= sizeof(short);
 
-            if (fd.read((char*)data, size) != size)
+            if (fd.read((char*)data, size) != size) {
+                  qWarning("SoundFont(%s) Sample(%s) read %u bytes failed", qPrintable(sf->get_name()), name, size);
+                  setValid(false);
                   return;
+                  }
 
             if (QSysInfo::ByteOrder == QSysInfo::BigEndian) {
                   unsigned char hi, lo;
@@ -676,11 +681,34 @@ void Sample::load()
                         data[i] = s;
                         }
                   }
-            end       -= (start + 1);       // marks last sample, contrary to SF spec.
+            end       -= start;
             loopstart -= start;
             loopend   -= start;
             start      = 0;
             }
+
+      // sanity checks:
+      // - start < end in SFont::load_shdr(int)
+      // - start < loopstart < loopend < end for !SF3 ibidem
+      // - … for SF3 in Sample::decompressOggVorbis(char *, unsigned int) in sfont3.cpp
+      // - most importantly, they are done *before* start is normalised to 0, which it is at this point
+      //
+      // now add some extra sanity checks from the SF2 spec (biased so they work with unsigned int);
+      // just a warning, they probably work with Fluid, possibly with audible artefacts though...
+      if (!(start + 7 < loopstart) || !(loopstart + 31 < loopend) || !(loopend + 7 < end))
+            qDebug("SoundFont(%s) Sample(%s) start(%u) startloop(%u) endloop(%u) end(%u) smaller than SoundFont 2.04 spec chapter 7.10 recommendation",
+                     qPrintable(sf->get_name()), name, start, loopstart, loopend, end);
+
+      // this used to cause a crash
+      if (!data) {
+            qWarning("SoundFont(%s) Sample(%s) data is nil", qPrintable(sf->get_name()), name);
+            setValid(false);
+            }
+
+      // from here, end marks the last sample, not one past as in the SF2 spec
+      if (end > 0)
+            end -= 1;
+
       optimize();
       }
 
@@ -749,6 +777,8 @@ bool SFont::load()
             }
       SFChunk chunk;
 
+      // so that any subsequent errors can be attributed to the right file
+      qDebug("Loading soundfont: %s", qPrintable(f.fileName()));
       try {
             readchunk(&chunk);
             if (chunkid(chunk.id) != RIFF_ID)
@@ -1628,9 +1658,7 @@ void SFont::load_shdr (int size)
       for (int i = 0; i < size; i++) {
             Sample* p = new Sample(this);
             sample.append(p);
-            char buffer[21];
-            READSTR (buffer);
-            // READSTR (p->name);
+            READSTR (p->name);
             READD (p->start);
             READD (p->end);	      /* - end, loopstart and loopend */
             READD (p->loopstart);	/* - will be checked and turned into */
@@ -1645,28 +1673,47 @@ void SFont::load_shdr (int size)
                   continue;
                   }
             if ((p->end > getSamplesize()) || (p->start > (p->end - 4))) {
-                  qWarning("Sample start/end file positions are invalid, disabling");
+                  qWarning("Sample(%s) start/end file positions are invalid, disabling", p->name);
                   p->setValid(false);
                   continue;
                   }
             p->setValid(true);
             if (p->sampletype & FLUID_SAMPLETYPE_OGG_VORBIS) {
+                  // done in Sample::decompressOggVorbis in sfont3.cpp
                   }
             else {
-                  // loop is fowled?? (cluck cluck :)
-                  if (p->loopend > p->end || p->loopstart >= p->loopend || p->loopstart <= p->start) {
-                        /* can pad loop by 8 samples and ensure at least 4 for loop (2*8+4) */
-                        if ((p->end - p->start) >= 20) {
-                              p->loopstart = p->start + 8;
-                              p->loopend   = p->end - 8;
-                              }
-                        else {      // loop is fowled, sample is tiny (can't pad 8 samples)
-                              p->loopstart = p->start + 1;
-                              p->loopend   = p->end - 1;
-                              }
+                  // try to fixup start<loopstart<loopend<end similar to recent FluidSynth;
+                  // some of these are technically invalid but exist in the field; see the
+                  // SoundFont 2.04 spec-related diagnostic about this in sfont.cpp
+                  if (p->loopstart == p->loopend) {
+                        // normalise to ensure they are within the sample, even non-looped
+                        p->loopstart = p->loopend = p->start;
                         }
-                  if ((p->end - p->start) < 8)
+                  else if (p->loopstart > p->loopend) {
+                        unsigned int looptmp;
+
+                        qWarning("Sample(%s) swapping reversed loopstart<=>loopend",
+                                 p->name);
+                        looptmp = p->loopstart;
+                        p->loopstart = p->loopend;
+                        p->loopend = looptmp;
+                        }
+                  // ensure they are at least within the sample
+                  if (p->loopstart < p->start || p->loopstart > p->end) {
+                        qWarning("Sample(%s) loopstart(%u) out of bounds [start=%u end=%u], setting to start",
+                                 p->name, p->loopstart, p->start, p->end);
+                        p->loopstart = p->start;
+                        }
+                  if (p->loopend < p->start || p->loopend > p->end) {
+                        qWarning("Sample(%s) loopend(%u) out of bounds [start=%u end=%u], setting to end",
+                                 p->name, p->loopend, p->start, p->end);
+                        p->loopstart = p->end;
+                        }
+                  // ensure it can even play
+                  if ((p->end - p->start) < 8) {
+                        qWarning("Sample(%s) too small, disabling", p->name);
                         p->setValid(false);
+                        }
                   }
             }
       FSKIP (SFSHDRSIZE);	/* skip terminal shdr */
--- a/audio/midi/fluid/sfont3.cpp
+++ b/audio/midi/fluid/sfont3.cpp
@@ -10,7 +10,7 @@ namespace FluidS {
 //   decompressOggVorbis
 //---------------------------------------------------------
 
-bool Sample::decompressOggVorbis(char* src, int size)
+bool Sample::decompressOggVorbis(char* src, unsigned int size)
       {
       AudioFile af;
       QByteArray ba(src, size);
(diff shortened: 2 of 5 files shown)


Jojo-Schmitz:
Just FYI: that `+ collect_artifacts` is no longer needed

Jojo-Schmitz:
What has commit 5aea0ab got to do with it? I don't see anything in your PR that reverts anything from that, except for the last chunk,  the changes to audio/midi/fluid/voice.cpp

mirabilos:
commit 5aea0ab was “try fixed crash on sample optimize”, which basically *is* the two `if`s in `fluid/voice.cpp`; the rest was just adding those macros.

As I wrote, if these macros are now unused they can be g/c’d, but the commit was about, and this reverts, these two `if`s.

audio/midi/fluid/sfont3.cpp
```diff
@@ -10,40 +10,45 @@ namespace FluidS {
...
+            qWarning("SoundFont(%s) Sample(%s) start(%u) startloop(%u) endloop(%u) end(%u) fowled (broken soundfont?), fixing up",
+                     qPrintable(sf->get_name()), name, start, loopstart, loopend, end);
             /* can pad loop by 8 samples and ensure at least 4 for loop (2*8+4) */
             if ((end - start) >= 20) {
                   loopstart = start + 8;
-                  loopend = end - 8;
+                  loopend   = end - 8;
                   }
-            else { // loop is fowled, sample is tiny (can't pad 8 samples)
+            else {      // loop is fowled, sample is tiny (can't pad 8 samples)
```
Jojo-Schmitz, line 44: better drop these and the following whitespace changes, they probably won't work for master's style anyhow

mirabilos: These are here because they are *precise* copy/paste from `sfont.cpp` replacing what was in `sfont3.cpp` before; I’ll handle master similarily. But here, I want to make them identical.

Jojo-Schmitz: Ah, I see.

audio/midi/fluid/sfont.cpp
```diff
@@ -1645,16 +1673,19 @@ void SFont::load_shdr (int size)
                   continue;
                   }
             if ((p->end > getSamplesize()) || (p->start > (p->end - 4))) {
-                  qWarning("Sample start/end file positions are invalid, disabling");
+                  qWarning("Sample(%s) start/end file positions are invalid, disabling", p->name);
```
Jojo-Schmitz, line 1676: shouldn't there be a space before the `(%s)`?

audio/midi/fluid/sfont3.cpp
```diff
@@ -10,40 +10,45 @@ namespace FluidS {
...
+bool Sample::decompressOggVorbis(char* src, unsigned int size)
       {
       AudioFile af;
       QByteArray ba(src, size);
 
       start = 0;
       end   = 0;
       if (!af.open(ba)) {
-            qDebug("Sample::decompressOggVorbis: open failed: %s", af.error());
+            qDebug("SoundFont(%s) Sample(%s) decompressOggVorbis: open failed: %s", qPrintable(sf->get_name()), name, af.error());
```
Jojo-Schmitz, line 21: same here?

audio/midi/fluid/sfont.cpp
mirabilos: I find this easier to read (the string directly relates to Sample), and the alternatives would be…

    qWarning("Sample %s start/end…

… which is bad for strings with arbitrary contents, and…

    qWarning("Sample \"%s\" start/end…

… which is just uglier, and the way I used has something more of a programmer style.

Sorry, I have a headache and can barely think ,let alone formulate right.

mirabilos:
I’m in discussion with @mrbumpy409 (regarding how the MuseScore_General soundfont relates to these) and have [feedback from FluidSynth upstream](https://github.com/FluidSynth/fluidsynth/issues/814) and am especially looking at FluidSynth’s current set of start/end/loopstart/loopend checks:

https://github.com/FluidSynth/fluidsynth/blob/b8fb6c81e1ca27c0bba2f6a0168832214f91d497/src/sfloader/fluid_sfont.c#L773..L832

Considering both, I’ll probably change this:

- demote the `qWarning` about SoundFont spec noncompliance (regarding the 8/32/8 minimum sizes) into a `qDebug`
- make `loopstart == start` valid (need to test this, of course)
- make `loopend == loopstart` to avoid looping valid (MuseScore_General apparently has samples that “are not supposed to be looped”, which the SoundFont spec does not even consider)

This is more conservative than doing what FluidSynth does, but FluidSynth has tons of other changes which may or may not impact this, so I’d prefer that.

Comments?

Jojo-Schmitz:
I'd say: go for it.

audio/midi/fluid/sfont.cpp
Jojo-Schmitz: Probably not woth loosing much sweat on, as most users would never see those anyhow

Jojo-Schmitz:
Not sure whether it is related, but MuseScore 4 gives a few errors on the MuseScore General.sf3 soundfont:

    10:34:53.965 | ERROR | 2 | FluidSynth | init: requested set value for 'synth.sample-rate' out of range
    10:34:53.965 | ERROR | 2 | FluidSynth | init: requested set value for 'synth.sample-rate' out of range

Jojo-Schmitz:
I've played wiht this PR and found out that it crashes when on a drum staff switching to Note Input mode !?!

mirabilos:
Joachim Schmitz dixit:

>I've played wiht this PR and found out that it crashes when on a drum
>staff switching to Note Input mode !?!

Oh, ouch. That we can’t have, of course. I’ll try to reproduce this,
will look at it (when doing the other discussed changes).

Thanks,
//mirabilos
-- 
<ch> you introduced a merge commit        │<mika> % g rebase -i HEAD^^
<mika> sorry, no idea and rebasing just fscked │<mika> Segmentation
<ch> should have cloned into a clean repo      │  fault (core dumped)
<ch> if I rebase that now, it's really ugh     │<mika:#grml> wuahhhhhh

Jojo-Schmitz:
Unless I screwed up that is. Reproducing should be easy: new score, add Congas, select first rest. press N, KABOOM

Jojo-Schmitz:
But none of the files you modified mentioned in the stack trace, very strange

Jojo-Schmitz:
1   std::_Rb_tree<int, std::pair<int const, Ms::StaffType>, std::_Select1st<std::pair<int const, Ms::StaffType>>, std::less<int>, std::allocator<std::pair<int const, Ms::StaffType>>>::empty                                  stl_tree.h        1009 0x10ad72c 
2   std::map<int, Ms::StaffType>::empty                                                                                                                                                                                        stl_map.h         464  0x1095988 
3   Ms::StaffTypeList::staffType                                                                                                                                                                                               stafftypelist.cpp 29   0xbbad4e  
4   Ms::Staff::staffType                                                                                                                                                                                                       staff.cpp         1012 0xaa8afa  
5   Ms::Staff::isTabStaff                                                                                                                                                                                                      staff.cpp         1518 0xaaaf26  
6   Ms::Chord::add                                                                                                                                                                                                             chord.cpp         521  0x9347d5  
7   Ms::DrumTools::updateDrumset                                                                                                                                                                                               drumtools.cpp     155  0x4a2814  
8   Ms::DrumTools::setDrumset                                                                                                                                                                                                  drumtools.cpp     175  0x4a2a42  
9   Ms::MuseScore::showDrumTools                                                                                                                                                                                               musescore.cpp     6827 0x4346e4  
10  Ms::MuseScore::changeState                                                                                                                                                                                                 musescore.cpp     4456 0x424d5b  
11  Ms::MuseScore::updateInputState                                                                                                                                                                                            keyb.cpp          92   0x4d7c62  
12  Ms::ScoreView::startNoteEntry                                                                                                                                                                                              scoreview.cpp     3238 0x64aa0d  
13  Ms::ScoreView::changeState                                                                                                                                                                                                 events.cpp        1186 0x67511e  
14  Ms::ScoreView::<lambda(Ms::ScoreView *, const QByteArray&)>::operator()(Ms::ScoreView *, const QByteArray &) const                                                                                                         scoreview.cpp     2112 0x63fea2  
15  std::_Function_handler<void(Ms::ScoreView *, const QByteArray&), Ms::ScoreView::cmd(const char *)::<lambda(Ms::ScoreView *, const QByteArray&)>>::_M_invoke(const std::_Any_data &, Ms::ScoreView *&&, const QByteArray &) std_function.h    297  0x65ac7a  
16  std::function<void (Ms::ScoreView *, QByteArray const&)>::operator()(Ms::ScoreView *, QByteArray const&) const                                                                                                             std_function.h    687  0x10afbad 
17  Ms::ScoreView::cmd                                                                                                                                                                                                         scoreview.cpp     2890 0x648719  
18  Ms::ScoreView::cmd                                                                                                                                                                                                         scoreview.cpp     2089 0x63fdff  
19  Ms::MuseScore::cmd                                                                                                                                                                                                         musescore.cpp     6601 0x4338b4  
20  Ms::MuseScore::cmd                                                                                                                                                                                                         musescore.cpp     6032 0x43024a  
... <Mehr>

Jojo-Schmitz:
Does look like some kind of stack corruption to me

Jojo-Schmitz:
The CI artifact (for Windoes) does **not** crash ... Might a MinGW issue, or a Qt 5.15.2 one? Or  Debug vs. RelWithDebInfo?

mirabilos:
Joachim Schmitz dixit:

>Unless I screwed up that is. Reproducing should be easy: new score, add
>Congas, select first rest. press N, KABOOM

I built that very same commit, and I cannot reproduce this.
From the backtrace you’ve given…

>1   std::_Rb_tree<int, std::pair<int const, Ms::StaffType>, std::_Select1st<std::pair<int const, Ms::StaffType>>, std::less<int>, std::allocator<std::pair<int const, Ms::StaffType>>>::empty                                  stl_tree.h        1009 0x10ad72c
>2   std::map<int, Ms::StaffType>::empty                                                                                                                                                                                        stl_map.h         464  0x1095988
>3   Ms::StaffTypeList::staffType                                                                                                                                                                                               stafftypelist.cpp 29   0xbbad4e

   29       if (staffTypeChanges.empty())
   30             return firstStaffType;

>4   Ms::Staff::staffType                                                                                                                                                                                                       staff.cpp         1012 0xaa8afa
[…]

… I’m assuming this to be unrelated. Look at libmscore/stafftypelist.h:

   28 class StaffTypeList {

   30       std::map<int, StaffType> staffTypeChanges;

   33       StaffTypeList() {}

That std::map is never instantiated as far as I can tell.
Perhaps it only works by accident but if the stack/heap
contains something else in that place, it explodes.

bye,
//mirabilos
-- 
<ch> you introduced a merge commit        │<mika> % g rebase -i HEAD^^
<mika> sorry, no idea and rebasing just fscked │<mika> Segmentation
<ch> should have cloned into a clean repo      │  fault (core dumped)
<ch> if I rebase that now, it's really ugh     │<mika:#grml> wuahhhhhh

Jojo-Schmitz:
Yes, maybe your change is just the straw that breaks the camel's back, under some cicumstancesne and some platforms.
But it happens only with that change of your's, not with any other.

Jojo-Schmitz:
Where/how to place such a `new`?

Jojo-Schmitz:
There is a(n empty) constructor, `StaffTypeList() {}`

Jojo-Schmitz:
shouldn't that instantiate all (private) members?

Jojo-Schmitz:
`std::map` is used in multiple places, but I can't see a `new` on any of them

Jojo-Schmitz:
Happens on MSVC (and with RelWithDebInfo) too.
So might still be a Qt 5.15.2 issue (which I'm using currently), or a plain Windows one?

mirabilos:
Joachim Schmitz dixit:

>`std::map` is used in multiple places, but I can't see a `new` on any of them

Oh okay, so this is an embedded object, not a pointer?

Hrm. Then I’m fresh out of ideas.

mirabilos:
Joachim Schmitz dixit:

>Happens on MSVC (and with RelWithDebInfo) too.
>So might still be a Qt 5.15.2 issue (which I'm using currently), or a plain Windows one?

Huh.

I have no way to test the latter. I’m both compiling against and
running with 5.15.2+dfsg-5 so either they fixed something in Qt
in the Debian patches or the latter or… dunno.

I’ve also no idea how to debug that under Windows. I’d try to
inspect the various objects involved and their members, even
internal ones, and maybe set a breakpoint to just before the
crash (or, if the debugger has a “time machine”, rewind a bit).

AntonioBL:
@mirabilos : if it is a hard-to-find bug, under Linux you could try to compile (in debug mode) with Address Sanitizer turned on (and then run from command line without the need for a debugger). It then halts whenever a memory/address violation is found.
Just clean the build directory and build after having defined a couple of additional env variables, see: https://github.com/musescore/MuseScore/blob/6fe71f9ed129e502b0ab4030d4672833efbbd6de/.github/workflows/ci_linux.yml#L121-L122
Maybe it can be enabled also with CMake, but it has been a while since I last compiled MuseScore (and my laptop broke so I cannot check at the moment).

If it is a Windows-only bug then it is harder to deal with. I don't know if it is possible to (compile and) run under Wine.

Ciao,
ABL

Jojo-Schmitz:
Maybe it only happens if you rebase to the latest 3.x? So that some of the commits between 5c548c3bf7fbf95189221b6e68eb1cd4c4b8027c (which your PR is based on) and 7d1c732a18f76410dff7514a775930bc39d4ee39 is triggigering this?

mirabilos:
Joachim Schmitz dixit:

>Maybe it only happens if you rebase to the latest 3.x? So that some of
>the commits between 5c548c3bf7fbf95189221b6e68eb1cd4c4b8027c (which
>your PR is based on) and 7d1c732a18f76410dff7514a775930bc39d4ee39 is
>triggigering this?

Oh, I thought you tested this commit, not a rebased version of it?

Jojo-Schmitz:
I tested this commit, built myself and on top of the latest 3.x.

Jojo-Schmitz:
Sorry, I should have made that clear earlier...

mirabilos:
Joachim Schmitz dixit:

>I tested this commit, built myself and on top of the latest 3.x.

No, then you did not test this commit, you tested the commitdiff
between this commit and its parent applied on top of a different
commit. A commit is always the state of the source tree (in git),
not a change (like in other version control systems).

So… on top of which precise commit in the 3.x branch (which *is*
a moving target) did you test this?

Jojo-Schmitz:
The latest. Last time yesterday. It is if at all, a slow moving target currently

Jojo-Schmitz:
Very strange, I can **not** reproduce the issue any longer, not even when rebasing to the latest 3.x.

Jojo-Schmitz:
Sorry for the mess and waste of time (your's and mind), the fault was entirely on my side, had another code change there that I though to have had excluded but apparently did not. Bug is finally once and forever sorted now, your PR seems just doing fine as far as I can tell.

mirabilos:
Joachim Schmitz dixit:

>Sorry for the mess and waste of time (your's and mind), the fault was
>entirely on my side, had another code change there that I though to
>have had excluded but apparently did not. Bug is finally once and
>forever sorted now, your PR seems just doing fine as far as I can tell.

Phew, thanks for confirming.

I’ll continue with doing some of the changes discussed here and/or
requested by s.chriscollins then, when I have time again (just got
an urgent “please make me a PDF from this score but with this other
text and these changes” request)…

mirabilos:
I’ve updated this as follows:

- change “SoundFont 2.04 spec chapter 7.10 recommendation” warning
  from `qWarning` to `qDebug` so it doesn’t show up in upstream mu͒ 3.x
  release builds
- change “loop is fowled??” check-and-fixup code to match recent
  FluidSynth and real-existing soundfonts better:
  - permit `loopstart`==`loopend` to avoid looping
    (fix them up as `loopstart`=`loopend`=`start`, so it’s within the
    sample always) which FluidSynth can do, even here:
        `fluid/voice.cpp:#define FLUID_MIN_LOOP_SIZE 2`
        `fluid/voice.cpp:#define FLUID_MIN_LOOP_PAD  0`
  - permit `loopstart`==`start` since `FLUID_MIN_LOOP_PAD` is 0 and
    this is used but do ensure loopstart and loopend are both
    within the sample, at least (and warn)
  - if `loopstart > loopend`, swap them around, with warning
  - stop forcibly padding the loop when not necessary, fixes
    playback of some bad/old samples
- more `unsigned int` (XXX should be `size_t`?) for sample offsets

This works well for me, in both 2.3.2 and 3.2.3 with which I’ve locally tested this. Tests and further review (mostly by [diff from the previous installment](https://github.com/musescore/MuseScore/compare/5f4b41eaa7f21703281455b51b6e5bcd868d0bb5..ca0456559a35aa653a20876da1bac273502d1932)? though only the `audio/midi/fluid/` changes, the rest is a rebase on top of latest `3.x`…) welcome!

I plan on trying to get this into the **really**-soon-to-be-released Debian 11, so timely OK or veto appreciated.

mirabilos:
@Jojo-Schmitz did you have a chance to look at this again?

Jojo-Schmitz:
It works fine for me, but I don't recall ever having had the problem this PRs trying to fix

RomanPudashkin:
3.x is closed for any changes

PR #7731, opened 2021-03-15
[MU4] Added draw deserialization step

igorkorsukov:
see also 
https://github.com/musescore/MuseScore/pull/7724
https://github.com/musescore/MuseScore/pull/7614

### DIFF
--- a/src/libmscore/draw/drawbufferjson.cpp
+++ b/src/libmscore/draw/drawbufferjson.cpp
@@ -16,21 +16,15 @@
 //  along with this program; if not, write to the Free Software
 //  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 //=============================================================================
-
-//! HACK to access QPainterPath private data
-#undef QPAINTERPATH_H
-#undef QPAINTERPATH_P_H
-#define private public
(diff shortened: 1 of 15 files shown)


src/libmscore/draw/drawbufferjson.cpp
```diff
@@ -39,31 +40,57 @@ static int rtoi(qreal v)
...
+
 static QJsonObject toObj(const QPen& pen)
 {
-    QJsonObject o;
-    o["style"] = static_cast<int>(pen.style());
-    o["color"] = pen.color().name();
-    o["width"] = pen.width();
-    return o;
+    QJsonObject obj;
+    obj["style"] = static_cast<int>(pen.style());
```
RomanPudashkin, line 51: I think it is better to add static strings at the top of this file and use them as keys for serialization/deserialization

RomanPudashkin (approved):

src/libmscore/draw/drawbufferjson.cpp
igorkorsukov: Not a bad idea, but I think this is superfluously in this case.

PR #7735, opened 2021-03-16
[MU4] Added draw data compare step

igorkorsukov:
see also
#7724
#7614
#7731

### DIFF
--- a/src/libmscore/draw/drawcomp.cpp
+++ b/src/libmscore/draw/drawcomp.cpp
@@ -1,0 +1,443 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2021 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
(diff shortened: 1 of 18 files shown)


src/libmscore/draw/drawcomp.cpp
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    if (sz1.height() != sz2.height()) {
+        return false;
+    }
+
+    return true;
+}
+
+static bool isEqual(const QPen& p1, const QPen& p2)
```
RomanPudashkin, line 89: Can we just use == instead of the current implementation?

 static bool isEqual(const QPen& p1, const QPen& p2) {
 return p1 == p2;
 }

src/libmscore/draw/drawcomp.cpp
```diff
@@ -0,0 +1,443 @@
...
+    if (f1.italic() != f2.italic()) {
+        return false;
+    }
+
+    return true;
+}
+
+static bool isEqual(const QTransform& t1, const QTransform& t2)
+{
+    if (t1 != t2) {
```
RomanPudashkin, line 142: You can simplify this function:

 static bool isEqual(const QTransform& t1, const QTransform& t2) {
 return t1 == t2;
 }

igorkorsukov: I want all methods to be consistent

src/libmscore/draw/drawcomp.cpp
igorkorsukov: There are a number of properties that I do not compare, for example brush, because I do not save them in json, respectively, I do not read them json, because they are not important to us, but because of them the comparison maybe failing.

RomanPudashkin (approved):

PR #7737, opened 2021-03-17
[MU4] Added diff painting step

igorkorsukov:
see also
#7724
#7614
#7731
#7735


src/autobot/internal/steps/abdiffdrawstep.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+using namespace mu::autobot;
+
+void AbDiffDrawStep::doRun(AbContext ctx)
+{
+    draw::Diff diff = ctx.val<draw::Diff>(AbContext::Key::DiffDrawData);
+    AbPaintProvider::instance()->setDiff(diff);
+    AbPaintProvider::instance()->setIsDiffDrawEnabled(true);
+
+    dispatcher()->dispatch("dev-notationview-redraw");
```
RomanPudashkin, line 35: Can you please explain the purpose of this?

igorkorsukov: I need to call the notation redrawing from the test.
I thought how best to do this, variants:
1. Add view notations to the global context - but there can be several of them at the same time (one in the test)
2. We can do redrawn notation view indirectly, change something in the notation itself, or change the zoom and return it back, something like that - but I don't like this variant.
3. Add dev actions - I selected this variant.

RomanPudashkin (approved):

PR #7739, opened 2021-03-17
[MU3] Fix #297613: (Linux) File save can't determine file type without extension

Jojo-Schmitz:
This basically reverts 3390aab, so for Mac and Windows nothing changes, but for Linux it now defaults to use the Qt file dialogs rather than the native ones. Can still get changed in the advanced properties, by ticking `ui/Application/useNativeDialogs`.

Resolves: https://musescore.org/en/node/297613

As far as I can tell this is not yet needed/feasible for master

### DIFF
--- a/mscore/preferences.cpp
+++ b/mscore/preferences.cpp
@@ -239,7 +239,11 @@ void Preferences::init(bool storeInMemoryOnly)
             {PREF_UI_APP_RASTER_HORIZONTAL,                        new IntPreference(2)},
             {PREF_UI_APP_RASTER_VERTICAL,                          new IntPreference(2)},
             {PREF_UI_APP_SHOWSTATUSBAR,                            new BoolPreference(true)},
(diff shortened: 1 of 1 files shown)


RobFog:
cc @mirabilos

MarcSabatella:
I have mixed feelings on this, but not *strong* ones either way.  To me neither option is exactly perfect.

mirabilos:
Erk, this is… not good.

On the other hand, the nōn-native file **open** dialogue is so slow (because it must generate preview pictures for `.mscx` files) it’s choosing between a rock and a hard place. I have quite a strong opinion against the nōn-native open dialogue, and in the save dialogue you can easily work around this by manually adding the required file extension instead of letting the software add it automatically :/

I’d research into why it doesn’t do that instead.

Jojo-Schmitz:
@mirabilos any outcome of your further research?

Jojo-Schmitz:
See #9000

mirabilos:
I said I’d, not I’ll… I’ve got few detail knowledge of GUI things personally.

PR #7740, opened 2021-03-17
[MU4] Do not save unused images in the MSCZ archive.

mattmcclinch:
Resolves: https://musescore.org/en/node/318772.

When checking whether an image is used in a given score, also check whether the image's parent actually contains the image. If not, this means that the image has been removed, but is being kept alive in the undo stack, and should not be saved in the MSCZ archive.

### DIFF
--- a/src/libmscore/imageStore.cpp
+++ b/src/libmscore/imageStore.cpp
@@ -55,7 +55,7 @@ void ImageStoreItem::reference(Image* image)
 bool ImageStoreItem::isUsed(Score* score) const
 {
     foreach (Image* image, _references) {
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
For 3.x this won't work, as that branch is apparently lacking a `treeChildIdx()` method, but your proposal in https://github.com/mattmcclinch/MuseScore/commit/f2c3cbc should do there.

Jojo-Schmitz (approved):

Jojo-Schmitz:
See #7871 for a 3.x version

PR #7744, opened 2021-03-18
Fix compilation with MinGW on Windows

shoogle:
Errors fixed:

__Unicode__

```
src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp: In member function 'void mu::ui::WindowsPlatformTheme::startListening()':
src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp:78:62: error: cannot convert 'const wchar_t' to 'LPCSTR' {aka 'const char'}
     if (RegOpenKeyEx(HKEY_CURRENT_USER, windowsThemeKey.c_str(), 0,
                                                              ^
```

__Type__

```
src/libmscore/scorefile.cpp: In member function 'bool Ms::MasterScore::saveFile(bool)':
src/libmscore/scorefile.cpp:460:72: error: invalid cast from type 'QByteArray' to type 'LPCTSTR' {aka 'const wchar_t*'}
             SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
                                                                        ^
```

### DIFF
--- a/build/cmake/SetupBuildEnvironment.cmake
+++ b/build/cmake/SetupBuildEnvironment.cmake
@@ -53,6 +53,9 @@ elseif(CC_IS_MINGW)
         set(CMAKE_EXE_LINKER_FLAGS "-Wl,--large-address-aware")
     endif (NOT BUILD_64)
 
+    add_definitions(-D_UNICODE)
+    add_definitions(-DUNICODE)
+
 elseif(CC_IS_CLANG)
     message(STATUS "Using Compiler CLANG ${CMAKE_CXX_COMPILER_VERSION}")
 
--- a/src/libmscore/scorefile.cpp
+++ b/src/libmscore/scorefile.cpp
@@ -449,16 +449,7 @@ bool MasterScore::saveFile(bool generateBackup)
             dir.mkdir(backupSubdirString);
 #ifdef Q_OS_WIN
             const QString backupDirNativePath = QDir::toNativeSeparators(backupDirString);
-#if (defined (_MSCVER) || defined (_MSC_VER))
-   #if (defined (UNICODE))
-            SetFileAttributes((LPCTSTR)backupDirNativePath.unicode(), FILE_ATTRIBUTE_HIDDEN);
-   #else
-            // Use byte-based Windows function
-            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
-   #endif
-#else
-            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
-#endif
+            SetFileAttributesW(reinterpret_cast<LPCWSTR>(backupDirNativePath.utf16()), FILE_ATTRIBUTE_HIDDEN);
 #endif
         }
         const QString backupName = QString(".") + info.fileName() + QString(",");
--- a/src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
+++ b/src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp
@@ -75,9 +75,9 @@ void WindowsPlatformTheme::startListening()
         return;
     }
     m_isListening = true;
-    if (RegOpenKeyEx(HKEY_CURRENT_USER, windowsThemeKey.c_str(), 0,
-                     KEY_NOTIFY | KEY_CREATE_SUB_KEY | KEY_ENUMERATE_SUB_KEYS | KEY_QUERY_VALUE | KEY_WOW64_64KEY,
-                     &hKey) == ERROR_SUCCESS) {
+    if (RegOpenKeyExW(HKEY_CURRENT_USER, windowsThemeKey.c_str(), 0,
+                      KEY_NOTIFY | KEY_CREATE_SUB_KEY | KEY_ENUMERATE_SUB_KEYS | KEY_QUERY_VALUE | KEY_WOW64_64KEY,
+                      &hKey) == ERROR_SUCCESS) {
         m_listenThread = std::thread([this]() { th_listen(); });
     } else {
         LOGD() << "Failed opening key for listening dark theme changes.";


Jojo-Schmitz:
Does it also solve the warning?

    src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp:41:126: warning: cast between incompatible function types from 'FARPROC' {aka 'int (__attribute__((stdcall)) *)()'} to 'fnRtlGetNtVersionNumbers' {aka 'void (__attribute__((stdcall)) *)(long unsigned int*, long unsigned int*, long unsigned int*)'} [-Wcast-function-type]
         = reinterpret_cast<fnRtlGetNtVersionNumbers>(GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetNtVersionNumbers"));

Jojo-Schmitz:
No it does not. Not the scope of this PR I guess.

src/libmscore/scorefile.cpp
```diff
@@ -457,7 +457,7 @@ bool MasterScore::saveFile(bool generateBackup)
             SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
    #endif
 #else
-            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
+            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit().constData(), FILE_ATTRIBUTE_HIDDEN);
```
Jojo-Schmitz, line 460: replacing `toLocal8Bit().constData()` with `unicode()` works too, which means we could just drop the entire `#if (defined (_MSCVER) || defined (_MSC_VER))...#else...#endif` stuff, simplifying that preprocessor "mess" quite a bit?

Jojo-Schmitz: Actually the `toLocal8Bit().constData()` is needed in the above `#else` branche too anyway., So it should be:

   #ifdef Q_OS_WIN
            const QString backupDirNativePath = QDir::toNativeSeparators(backupDirString);
       #if (defined (UNICODE))
            SetFileAttributes((LPCTSTR)backupDirNativePath.unicode(), FILE_ATTRIBUTE_HIDDEN);
       #else
            // Use byte-based Windows function
            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit().constData(), FILE_ATTRIBUTE_HIDDEN);
       #endif
   #endif

igorkorsukov: I think we can assume that Unicode is always on. So, we can simply do:
```
#ifdef Q_OS_WIN
        const QString backupDirNativePath = QDir::toNativeSeparators(backupDirString);
        SetFileAttributes((LPCTSTR)backupDirNativePath.unicode()->unicode(), FILE_ATTRIBUTE_HIDDEN);
#endif
```

Jojo-Schmitz: Works (compiles and links and runs without errors) in MinGW and MSVC, I checked both, with and without UNICODE being #define'd and even in 3.x (which doesn't `#define UNICODE` for MinGW at all, so I'm not sure whether it should be done here, that then would require a different fix for the other error though.)

Jojo-Schmitz: Well, there are several places in the code that switch UNICODE off explicitly, to get the char based wWindows functions.

IIRC the UNICODE thing was causing https://musescore.org/en/node/287955 ?

Jojo-Schmitz: As mentioned, in 3.x MinGW does not use/set UNICODE, I guess for a reason?

build/cmake/SetupBuildEnvironment.cmake
```diff
@@ -53,6 +53,9 @@ elseif(CC_IS_MINGW)
         set(CMAKE_EXE_LINKER_FLAGS "-Wl,--large-address-aware")
     endif (NOT BUILD_64)
 
+    add_definitions(-D_UNICODE)
+    add_definitions(-DUNICODE)
+
```
Jojo-Schmitz, line 58: Anodher appoarch is to make the use of `std::wstring` in src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp dependant on whether `UNICODE` is `#define`'d or not:

    #if defined(UNICODE)
    static const std::wstring windowsThemeKey = L"Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize";
    static const std::wstring windowsThemeSubkey = L"AppsUseLightTheme";
    #else
    static const std::string windowsThemeKey = "Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize";
    static const std::string windowsThemeSubkey = "AppsUseLightTheme";
    #endif

src/libmscore/scorefile.cpp
```diff
@@ -448,17 +448,13 @@ bool MasterScore::saveFile(bool generateBackup)
...
 #ifdef Q_OS_WIN
-            const QString backupDirNativePath = QDir::toNativeSeparators(backupDirString);
-#if (defined (_MSCVER) || defined (_MSC_VER))
+        const QString backupDirNativePath = QDir::toNativeSeparators(backupDirString);
    #if (defined (UNICODE))
-            SetFileAttributes((LPCTSTR)backupDirNativePath.unicode(), FILE_ATTRIBUTE_HIDDEN);
+        SetFileAttributes((LPCTSTR)backupDirNativePath.unicode(), FILE_ATTRIBUTE_HIDDEN);
    #else
-            // Use byte-based Windows function
-            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
```
Jojo-Schmitz, line 457: those indents are still needed

src/libmscore/scorefile.cpp
```diff
@@ -448,17 +448,13 @@ bool MasterScore::saveFile(bool generateBackup)
...
+        // Use byte-based Windows function
+        SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit().constData(), FILE_ATTRIBUTE_HIDDEN);
```
Jojo-Schmitz, line 456: this is not needed, woks without that `.constData()`

build/cmake/SetupBuildEnvironment.cmake
Jojo-Schmitz: So this and the 'cleanup' change (dropping the `#if (defined (_MSCVER) || defined (_MSC_VER))...#else...#endif` stuff including the else branch) I proposed should fix the MinGW compile issue. No need to change whether to use UNICODE or not.

Jojo-Schmitz: @cbjeukendrup, your opinion (your code from ad8ad3f0c)?

And idea in the compiler warning?

src/libmscore/scorefile.cpp
shoogle: @igorkorsukov, is the double unicode (`unicode()->unicode()`) in your [last comment](https://github.com/musescore/MuseScore/pull/7744#discussion_r596616234) intentional?

What's the consensus on whether the `#else` clause is still needed?

igorkorsukov: > @igorkorsukov, is the double unicode (unicode()->unicode()) in your last comment intentional?

Yes, intentionally, I think it is more correct.
```
const QChar *QString::unicode() const
```
Cast `const QChar *`  to `LPCTSTR` works only on the basis of the internal implementation of `QChar`

Jojo-Schmitz: As far as I can tell the entire `#else` clause is not needed.

igorkorsukov: > What's the consensus on whether the #else clause is still needed?

I see no reason for it, I believe that we can assume that `UNICODE` is always on by default.
If we need to turn it off somewhere, then turn it off only in these special places.

build/cmake/SetupBuildEnvironment.cmake
igorkorsukov: I see no reason for it, I believe that we can assume that UNICODE is always on by default.
If we need to turn it off somewhere, then turn it off only in these special places.

src/libmscore/scorefile.cpp
Jojo-Schmitz: I mean that matching the `#if (defined (_MSCVER) || defined (_MSC_VER))`. I'd rather keep the `UNICODE` part, for now, we may revisit that later, but as said, some (Windows related) partx in tthe code explicitly disable `UNICODE`

build/cmake/SetupBuildEnvironment.cmake
Jojo-Schmitz: It is on for MSVC, off for MinGW, and has been that way ever since @anatoly-os ported it to MSVC

src/libmscore/scorefile.cpp
igorkorsukov: > as said, some (Windows related) parts in the code explicitly disable UNICODE

Let those specific parts turn off Unicode if they need to (by way, PortMidi, propper not used in the MU4)
But why do we need to build the entire project with unicode disabled?
For example, in this case, if we disable unicode, it will stop working with any not ANSI (English) paths and file names.

build/cmake/SetupBuildEnvironment.cmake
Jojo-Schmitz: But if the decision is to require `UNICODE`for both fro now on, that should be fine.

Jojo-Schmitz: It does certainly simplify the code

Jojo-Schmitz: And we don't officially build with MinGW anyway, so this change would never affect anyone on any release

src/libmscore/scorefile.cpp
```diff
@@ -449,16 +449,7 @@ bool MasterScore::saveFile(bool generateBackup)
...
-   #if (defined (UNICODE))
-            SetFileAttributes((LPCTSTR)backupDirNativePath.unicode(), FILE_ATTRIBUTE_HIDDEN);
-   #else
-            // Use byte-based Windows function
-            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
-   #endif
-#else
-            SetFileAttributes((LPCTSTR)backupDirNativePath.toLocal8Bit(), FILE_ATTRIBUTE_HIDDEN);
-#endif
+            SetFileAttributes(reinterpret_cast<LPCTSTR>(backupDirNativePath.unicode()->unicode()), FILE_ATTRIBUTE_HIDDEN);
```
Jojo-Schmitz, line 452: that `->unicode()` is, at best, superfluos. If hasn't been need ever before, I don't see why it would now.

shoogle, line 452: @igorkorsukov, `reinterpret_cast` was necessary to stop Qt Creator complaining.

It still seems a bit wierd to do [`QString::unicode()`](https://doc.qt.io/qt-5/qstring.html#unicode), which returns `const QChar *` (i.e. a pointer to a sequence of QChars), and then call [`QChar::unicode()`](https://doc.qt.io/qt-5/qchar.html#unicode) on just the first `QChar` in the sequence, or am I missing something here? Surely we need to end up with a pointer, but doesn't `QChar::unicode()` just return a codepoint?

Jojo-Schmitz: Yet another sign that this `->unicode()` isn't needed, and now even does harm.

igorkorsukov: As I said, the cast `QChar*` to `LPCTSTR` is not very correct, it only works thanks to the current implementation of  `QChar*`, it works because `sizeof( QChar) == sizeof(uint16_t)`. If the internal implementation of `QChar` changes, some field will be added, for example `QChar` will be like this:
```
class QChar {
chort* data;
bool someInternalField;
}
```
then everything will break.
It is unlikely that there will be a change in the implementation of `QChar`, that is, in this case, we are talking about the formal correctness of the work.

Jojo-Schmitz: So let's cross that bridge whe we get there. Currently the code just works (and did all the way up to now), and does not need that `->unicode()`. I don't think Qt will change their implementation, I also don't think we're going to move do any Qt version higher than 5.15.2, as none of those are a) LTS and b) free to use anymore.

Whether we change from a C-style cast to a C++ style one is a matter of taste here

igorkorsukov: https://github.com/musescore/MuseScore/pull/7744#discussion_r597465131

Jojo-Schmitz: > @igorkorsukov, `reinterpret_cast` was necessary to stop Qt Creator complaining.

`(LPCTSTR)` works just fine too.

igorkorsukov: `(LPCTSTR) == reinterpret_cast` almost   :)
`(LPCTSTR)` - C style
`reinterpret_cast` - C++ stylr

Jojo-Schmitz: I got that (now)

igorkorsukov: > I also don't think we're going to move do any Qt version higher than 5.15.2, as none of those are a) LTS and b) free to use anymore.

But we are creating an application for long-term use and development, in a few years there will be Qt 6 and Qt 7 ... :)

Jojo-Schmitz: As said, let's rather cross that brigdge when we get there.

igorkorsukov: let's cross that bridge

igorkorsukov:
```
reinterpret_cast<LPCTSTR>(backupDirNativePath.unicode()->unicode())
```
Is not correct :(
Correct
```
reinterpret_cast<LPCTSTR>(&backupDirNativePath.unicode()->unicode())
```

or I think it better, just 

```
backupDirNativePath.utf16()
```

Jojo-Schmitz: No cast needed? Then I like it better ;-)

Jojo-Schmitz: Nope, needs a cast

igorkorsukov: I thought again
```
reinterpret_cast<LPCTSTR>(&backupDirNativePath.unicode()->unicode())
```
and 
```
&backupDirNativePath.unicode()->unicode()
```
not correct too, sorry for misleading

So
```
reinterpret_cast<LPCTSTR>(backupDirNativePath.unicode())
```
not correct, but work by the internal implementation 

```
backupDirNativePath.utf16()
``
correct and work

Comment (Jojo-Schmitz) [reply src/libmscore/scorefile.cpp:452]: No, it doesn't even compile:

    libmscore\scorefile.cpp:450:62: error: invalid conversion from 'const ushort*' {aka 'const short unsigned int*'} to 'LPCWSTR' {aka 'const wchar_t*'} [-fpermissive]
                   SetFileAttributes(backupDirNativePath.utf16(), FILE_ATTRIBUTE_HIDDEN);

Comment (igorkorsukov) [reply src/libmscore/scorefile.cpp:452]: Eh, it means we need a cast, but it's still the most correct option.
```
reinterpret_cast<LPCTSTR>(backupDirNativePath.utf16())
```
The specification is guaranteed that: 
ushort - 16 bit
wchar_t - 16 bit for Windows 

https://en.cppreference.com/w/cpp/language/types

Comment (Jojo-Schmitz) [reply src/libmscore/scorefile.cpp:452]: That does compile at least ;-), but isn't utf16 also unicode? And a method called `unicode()` pretty well matches the other part of the PR, `-DUNICODE` and `-D_UNICODE`

Comment (igorkorsukov) [reply src/libmscore/scorefile.cpp:452]: > And a method called unicode() pretty well matches the other part of the PR, -DUNICODE and -D_UNICODE

Nope, this is just an embarrassment, because the unicode() method just returns a pointer to one of the Qt classes :)

Comment (igorkorsukov) [reply src/libmscore/scorefile.cpp:452]: > but isn't utf16 also unicode? 

yes! 
"Unicode can be implemented by different character encodings. The Unicode standard defines Unicode Transformation Formats (UTF) UTF-8, UTF-16, and UTF-32, and several other encodings. The most commonly used encodings are UTF-8, UTF-16, and UCS-2 (a precursor of UTF-16 without full support for Unicode); GB18030 is standardized in China and implements Unicode fully, while not an official Unicode standard."
https://en.wikipedia.org/wiki/Unicode

In this case, we need exactly UTF-16
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/76f10dd8-699d-45e6-a53c-5aefc586da20

Comment (igorkorsukov) [reply src/libmscore/scorefile.cpp:452]: by way

#ifdef UNICODE
 typedef LPCWSTR LPCTSTR; 
#else
 typedef LPCSTR LPCTSTR;
#endif

#ifdef UNICODE
  #define SetFileAttributes SetFileAttributesW; 
#else
  #defineSetFileAttributes SetFileAttributesA;
#endif

Comment (Jojo-Schmitz) [reply src/libmscore/scorefile.cpp:452]: So `reinterpret_cast<LPCTSTR>(backupDirNativePath.utf16())` is the winner, I guess

Comment (shoogle) [reply src/libmscore/scorefile.cpp:452]: > and might even work independant of UNICODE being #define'd

Not quite. If `UNICODE` is not defined then `SetFileAttributes` and `LPCTSTR` expect an 8-bit string, not a 16-bit string like `QString::utf16()` provides. We could use `QString::toLocal8Bit().constData()` in that case, but there is a better way:

```c++
SetFileAttributesW(reinterpret_cast<LPCWSTR>(backupDirNativePath.utf16()), FILE_ATTRIBUTE_HIDDEN);
```

Notice explicit use of the `SetFileAttributesW` and `LPCWSTR` rather than the aliases `SetFileAttributes` and `LPCTSTR`, so this should work on modern Windows regardless of whether `UNICODE` is defined.

As I understand it, undefining `UNICODE` only changes the aliases; it doesn't actually disable Unicode support via the explicit `W` functions. The aliases are necessary for compatibility with code written prior to Windows NT, but I don't think we need to go there!

Jojo-Schmitz: But that other error still requires UNICODE, right? Or is there a method to make it waterproof too?

igorkorsukov: yes, there we can also use the method with the `W` suffix

igorkorsukov: There is just a choice, either uses Windows API methods with the `W` suffix everywhere or define the default `UNICODE` by default.

Jojo-Schmitz: Also for the warning generated a few lines later?

shoogle: I've updated the PR to use the `W` suffix in both files. The warning is different and I don't have time to investigate further unfortunately. I recommend merging this as-is to fix the error for MinGW users. The warning can wait for another time.

igorkorsukov (approved):

PR #7757, opened 2021-03-19
[MU4] Solve issue of fixed spacer between last staff and bottom of page.

njvdberg:
Solve issue of fixed spacer between last staff and bottom of page.
A negative value for a fixed spacer on the last staff will no longer push the last staff below the `Music Bottom Margin`.

Also the behavior is similar for both pre-3.6 staff spacing and vertical staves adjustment:

 1. When creating a fixed spacer on the last staff the gap is the distance between the bottom of the last staff and `Page Bottom Margin`.
 
 2. When the gap is increased the last staff is pushed up.

 3. When the gap is decreased the last staff is pushed down but it will **never** pushed **below** the `Music Bottom Margin`

 4. Because both pre-3.6 staff spacing and vertical staves adjustment are based on dividing remaining space between the last staff and `Music Bottom Margin`, when maximum distances between staves and/or systems are too small, it is not possible to honor the fixed spacer on the last staff with violating these maximum distances. In these situation, the maximum distances take precedence and the fixed gap between the last staff and  `Music Bottom Margin`. is honored.
In this, a fixed spacer on the last staff shows a slightly different behavior as fixed spacers on other staves.

This PR for `master` is similar to #7756 for `3.x`.

### DIFF
--- a/src/libmscore/layout.cpp
+++ b/src/libmscore/layout.cpp
@@ -1658,6 +1658,7 @@ static void distributeStaves(Page* page)
     int ngaps { 0 };
     qreal prevYBottom  { page->tm() };
     qreal yBottom      { 0.0 };
+    qreal spacerOffset { 0.0 };
     bool vbox          { false };
     Spacer* nextSpacer { nullptr };
     bool transferNormalBracket { false };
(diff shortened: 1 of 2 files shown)


RobFog:
Rebase needed.

njvdberg:
> Rebase needed.

Done

PR #7759, opened 2021-03-19
[MU4] Preferences. Update

RomanPudashkin:


src/appshell/internal/appshellconfiguration.h
```diff
@@ -41,6 +41,9 @@ class AppShellConfiguration : public IAppShellConfiguration
 
     bool isAppUpdatable() const override;
 
+    bool needCheckForUpdate() const override;
```
Eism, line 44: better to return ValCh for notification

PR #7777, opened 2021-03-21
[MU4] Fix 317951 - Horizontal Frame on system by itself at end of page causes previous system to disappear

njvdberg:
Resolves: https://musescore.org/en/node/317951

Will return the correct bottom y coordinate in case a `SysStaff` contains only a horizontal frame.
In this case the `SysStaff` doens't have a valid skyline and the initial (far too large) maximum value is returned.
This is solved by returning the height of the `SysStaff` in case the skyline is invalid (has no segments).

PR [#7776](https://github.com/musescore/MuseScore/pull/7776) is similar for `3.x`.


src/libmscore/skyline.cpp
```diff
@@ -284,6 +284,12 @@ void SkylineLine::paint(QPainter& p) const
         }
         x1 = x2;
     }
+
```
Jojo-Schmitz: still a line too much for my taste ;-)

Jojo-Schmitz (approved):

src/libmscore/skyline.cpp
njvdberg: `uncrustify` had the same comment :wink:

Jojo-Schmitz: @jojo-bot at work ;-)

njvdberg: And I thought giving the bot some extra work :smile:

PR #7778, opened 2021-03-21
[MU3] Solve a crash when score contains a vbox which is not on top of a page.

njvdberg:
When a `VBox` found, a `VerticalGapData` object is created but if this `VBox` wasn't the first element on a page, the application crashed because the constructor was looking for a `Spacer` on a non-existing staff.

PR [#7779](https://github.com/musescore/MuseScore/pull/7779) is similar for `master`.

### DIFF
--- a/libmscore/layout.cpp
+++ b/libmscore/layout.cpp
@@ -5120,7 +5120,7 @@ VerticalGapData::VerticalGapData(bool first, System *sys, Staff *st, SysStaff *s
             _normalisedSpacing = system->y() + (sysStaff ? sysStaff->y() : 0.0) - y;
             _maxActualSpacing = system->score()->styleP(Sid::maxStaffSpread);
 
-            Spacer* spacer { sys->upSpacer(st->idx(), nextSpacer) };
+            Spacer* spacer { staff ? system->upSpacer(staff->idx(), nextSpacer) : nullptr };
 
             if (spacer) {
                   _fixedSpacer = spacer->spacerType() == SpacerType::FIXED;


libmscore/layout.cpp
```diff
@@ -5120,7 +5120,7 @@ VerticalGapData::VerticalGapData(bool first, System *sys, Staff *st, SysStaff *s
             _normalisedSpacing = system->y() + (sysStaff ? sysStaff->y() : 0.0) - y;
             _maxActualSpacing = system->score()->styleP(Sid::maxStaffSpread);
 
-            Spacer* spacer { sys->upSpacer(st->idx(), nextSpacer) };
+            Spacer* spacer { st ? sys->upSpacer(st->idx(), nextSpacer) : nullptr };
```
Jojo-Schmitz, line 5123: maybe `(sys && st)`?

njvdberg: No, there is always a system, a `VBox` is also a system. However, a `VBox` doesn't have any `SysStaff`'s and that caused the crash (if the `VBox` was somewhere in the middle of the page).

Jojo-Schmitz: So maybe it'd never happen, but that check won't harm either

njvdberg: If a test is required, this will be a wrong place since `system` is already dereferenced several time times before (lines 5540, 5543 and 5544). Also before calling the constructor the `system` argument is dereferenced already.

Jojo-Schmitz: `system` is used before, not `sys`.

Jojo-Schmitz: but yes, probably not need to check, not there at least

Jojo-Schmitz: anyway: how to reproduce the issue prior to this change?

njvdberg: This is in the constructor and with `system(sys)` the member `system` is initialized with the argument `sys`, so both are the same.

The issue can be reproduced creating a `VBox` somewhere after the first system.

Jojo-Schmitz: We, no, I can't reproduce the issue.
And a mix of using `system` and `sys` looks wrong to me

njvdberg: You tried to reproduce with 3.6.2 of the 3.x branch? The crash occurs in the 3.x branch (and master) only, not in 3.6.2 .

Strictly speaking, mix of `sys` and `system` is not "wrong" but it certainly is confusing and is corrected.

Jojo-Schmitz: Ah, I tried to reproduce in 3.6.2, so that crash is a later regression

Jojo-Schmitz: and yes, 'wrong' is a bit strong, but certainly confusing

njvdberg: Yes, a later regression due to PR #7529

PR #7781, opened 2021-03-22
[MU4] Preferences. Folders

Eism:
image

### DIFF
--- a/src/appshell/view/folderspreferencesmodel.cpp
+++ b/src/appshell/view/folderspreferencesmodel.cpp
@@ -1,0 +1,193 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2021 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program; if not, write to the Free Software
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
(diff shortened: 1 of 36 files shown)


src/appshell/view/folderspreferencesmodel.cpp
```diff
@@ -0,0 +1,193 @@
...
+void FoldersPreferencesModel::load()
+{
+    beginResetModel();
+
+    m_folders = {
+        { FolderType::Scores, qtrc("appshell", "Scores"), scoresPath() },
+        { FolderType::Styles, qtrc("appshell", "Styles"), stylesPath() },
+        { FolderType::Templates, qtrc("appshell", "Templates"), templatesPath() },
+        { FolderType::Plugins, qtrc("appshell", "Plugins"), pluginsPath() },
+        { FolderType::SoundFonts, qtrc("appshell", "Soundfonts"), "" }, // todo: need implement
```
RomanPudashkin, line 85: Sound**F**onts?

src/extensions/internal/extensionsconfiguration.cpp
```diff
@@ -246,7 +245,30 @@ io::paths ExtensionsConfiguration::templatesPaths() const
...
 }
 
+ValCh<io::path> ExtensionsConfiguration::extensionsPath() const
+{
+    ValCh<io::path> result;
+    result.ch = m_extensionsPathChanged;
+
+    result.val = settings()->value(USER_EXTENSIONS_PATH).toString();
+    if (result.val.empty()) {
+        result.val = settings()->defaultValue(USER_EXTENSIONS_PATH).toString();
```
RomanPudashkin, line 255: You can call setDefaultValue in ExntesionsConfiguration::init and use Settings::value directly:

 ValCh<io::path> ExtensionsConfiguration::extensionsPath() const
 {
 ValCh<io::path> result;
 result.ch = m_extensionsPathChanged;
 result.val = settings()->value(USER_EXTENSIONS_PATH);
 
 return result;
 }

Eism: thanks, fixed

src/appshell/view/folderspreferencesmodel.cpp
Eism: thanks, fixed

RomanPudashkin (approved):

Issue #7766, opened 2021-03-20
[MU4 Issue] Upper Menu functions dont work


vshalkevich:
**Describe the bug**
Upper Menu functions dont work

**To Reproduce**
1. Create a score
2. Click on File > New from the upper menu (or any ther option)

**Expected behavior**
New Score Wizard opens (and other functions from upper menu works)

**Desktop (please complete the following information):**
MacOS

Issue #7691, opened 2021-03-06
[MU4 Issue] Upper Menu bars disappeared after score was created


vshalkevich:
**Describe the bug**
Upper Menu bars disappeared after score was created

**To Reproduce**
1. Create a score
2. Try to find upper Menu bar

**Expected behavior**
upper Menu bar should be visible after new score created

**Desktop (please complete the following information):**
MacOS

PR #7782, opened 2021-03-22
Get menus to work on macOS.

mattmcclinch:
This addresses MU4 issues #7691 and #7766. It is not ideal, since there is no apparent reason why this should be necessary, and it seems like a pointless thing to do. But it works.

### DIFF
--- a/src/appshell/view/dockwindow/dockmenubar.cpp
+++ b/src/appshell/view/dockwindow/dockmenubar.cpp
@@ -100,6 +100,7 @@ QMenu* DockMenuBar::makeMenu(const QVariantMap& menuItem) const
         }
     }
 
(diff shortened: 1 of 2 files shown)


Eism:
Thanks, @mattmcclinch

Eism (approved):

RobFog:
Is this meant to be marked "In progress"?

mattmcclinch:
>Is this meant to be marked as "In progress"?

It's a little late for that😉. But no, while for some reason I do not have the ability to add/remove labels on my own PRs, I could have marked it as a Draft if I had wanted to. But I think I was as happy with it as I was ever going to be.

vshalkevich:
You may write labels in the comments and moderator will add them :D

PR #7790, opened 2021-03-23
[MU4] Added test case select and files list

igorkorsukov:
https://user-images.githubusercontent.com/3818029/112158293-b16c2a80-8bf0-11eb-89ca-75456f3ee3b5.mp4


src/notation/view/notationpaintview.cpp
```diff
@@ -567,7 +567,11 @@ void NotationPaintView::scale(qreal scaling, const QPoint& pos)
     int dx = pointAfterScaling.x() - pointBeforeScaling.x();
     int dy = pointAfterScaling.y() - pointBeforeScaling.y();
 
-    moveCanvas(dx, dy);
+    if (dx != 0 || dy != 0) {
```
vpereverzev, line 570: maybe we just have to put "update" at the beginning of the "moveCanvas", right before this condition?

```
if (dx == 0 && dy == 0) {
        return;
    }
```

 Otherwise, we can forget this logic the next time we need to use the moveCanvas method.

vpereverzev:
just curious about these two files, because they're empty and I don't see their names in .qrc, so probably you just forgot to delete them?
https://github.com/musescore/MuseScore/pull/7790/files#diff-c5f3d3f4efb6c8fa80b3b981df666a6af9f9cc453e349a69260234e8bd2d38a1

https://github.com/musescore/MuseScore/pull/7790/files#diff-a180fcd537c06a96f2631467c74c375874707bb612573878dc91bb0edf915c87

src/autobot/iautobot.h
```diff
@@ -28,8 +34,22 @@ class IAutobot : MODULE_EXPORT_INTERFACE
...
+        Stoped = 0,
+        RunningAll,
+        RunningFile
+    };
+
+    virtual std::vector<ITestCasePtr> testCases() const = 0;
+    virtual ITestCasePtr testCase(const std::string& name) const = 0;
+
+    virtual void runAll(const std::string& testCaseName) = 0;
+    virtual void runFile(const std::string& testCaseName, int fileIndex) = 0;
```
vpereverzev, line 47: Based on the name of the method, I'd expect to have only one argument like filePath

`runFile(const std::string& filePath)`

could you explain the meaning of the second argument?

src/notation/view/notationpaintview.cpp
igorkorsukov: I believe `moveCanvas` is correct if `dx` and `dy` are 0 then no need to move the canvas - no call update needed

src/autobot/iautobot.h
igorkorsukov: first arg - test case, which needs run
second arg - file as file index in the file list (just for optimization, avoiding search file by path and etc)

igorkorsukov:
> just curious about these two files, because they're empty and I don't see their names in .qrc, so probably you just forgot to delete them?
> https://github.com/musescore/MuseScore/pull/7790/files#diff-c5f3d3f4efb6c8fa80b3b981df666a6af9f9cc453e349a69260234e8bd2d38a1
> 
> https://github.com/musescore/MuseScore/pull/7790/files#diff-a180fcd537c06a96f2631467c74c375874707bb612573878dc91bb0edf915c87

removed
(I had some troubles with a fix merge conflict, probably because of this they keep)

RomanPudashkin (approved):

PR #7793, opened 2021-03-24
[MU4] Fix #319079: Shift Selecting Last Word

apricot2012:
Resolves: *https://musescore.org/en/node/319079*

Original code only attempts to find the next ChordRest segment as the end marker for selecting a range, this fix adds the detection for the EndBarLine segment as a potential end marker, which resolves the issue.

### DIFF
--- a/src/libmscore/score.cpp
+++ b/src/libmscore/score.cpp
@@ -3421,7 +3421,7 @@ void Score::selectRange(Element* e, int staffIdx)
                     s2 = s2->next1MM(SegmentType::ChordRest);
                 }
 
-                if (s1 && s2) {
+                if (s1) {
                     _selection.setRange(s1, s2, idx1, idx2 + 1);
                     selectSimilarInRange(e);
                     if (selectedElement->track() == e->track()) {
--- a/src/libmscore/score.cpp
+++ b/src/libmscore/score.cpp
@@ -3407,6 +3407,12 @@ void Score::selectRange(Element* e, int staffIdx)
         if (selectedElement && e->type() == selectedElement->type()) {
             int idx1 = selectedElement->staffIdx();
             int idx2 = e->staffIdx();
+            if (idx2 < idx1) {
+                int temp = idx1;
+                idx1 = idx2;
+                idx2 = temp;
+            }
+
             if (idx1 >= 0 && idx2 >= 0) {
                 Fraction t1 = selectedElement->tick();
                 Fraction t2 = e->tick();


Jojo-Schmitz (approved):

src/libmscore/score.cpp
```diff
@@ -3418,7 +3418,7 @@ void Score::selectRange(Element* e, int staffIdx)
                 Segment* s1 = tick2segmentMM(t1, true, SegmentType::ChordRest);
                 Segment* s2 = tick2segmentMM(t2, true, SegmentType::ChordRest);
                 if (s2) {
-                    s2 = s2->next1MM(SegmentType::ChordRest);
+                    s2 = s2->next1MM(SegmentType::ChordRest | SegmentType::EndBarLine);
```
MarcSabatella, line 3421: Nice job tracking down where this goes wrong!  But I wonder, maybe this isn't the best solution.  Elsewhere we use the convention of passing nullptr as the end segment when creating a range, as the way of saying "select to end of score".  Maybe that's a better answer here - just let s2 be nullptr if that's what next1MM returns, and change the next line to simply "if (s1)"?

My only hesitation about recommending this is that elsewhere there are some odd inconsistencies with the end of selection where sometimes the end barline is selected and sometimes not (this is true for selections within the score too, not just at the end).  So maybe for MuseScore 4 we want to consider the best way to handle this.

Jojo-Schmitz: Tested:

                        if (s2)
                              s2 = s2->next1MM(SegmentType::ChordRest);
                        if (s1) {
                              _selection.setRange(s1, s2, idx1, idx2 + 1);

Indeed works too. And deleting code (here: the `& s2`) to fix a bug seems nicer than adding some ;-)

apricot2012: Thanks for the detailed feedback :). I've seen nullptr being used in other conditionals of score.cpp but didn't notice that it can also be an end marker itself. Indeed this looks like a more elegant fix to me -- made a new commit with the suggested changes.

Jojo-Schmitz (approved):

Jojo-Schmitz (approved):

Jojo-Schmitz:
Now just the commit message (the body) doesn't fit anymore

Jojo-Schmitz:
Unfortunatly this change leads to a crash, if the start and end syllable are in different staves (and the end one is at the end of the score), if the end syllable is in a higher staff:

https://github.com/musescore/MuseScore/blob/19a05847a5dcd2e4cce84a197f29652497d718cc/src/libmscore/select.cpp#L709

Jojo-Schmitz:
Independant of whether we'd use the 1st or the 2nd proposed fix.
Without it that selecting in the 'wrong' order doesn't work at all (not even if not including the last sylllable of a score), so this fix just uncovered another bug.

Jojo-Schmitz:
Maybe we should jusr do away with that assert?

Jojo-Schmitz:
I see nothing strictly illegal about `staffEnd > staffStart` not being the case

Jojo-Schmitz:
But while removing that condition from the Q_ASSERT fixes the crash, it still doesn't make such a selection work :-(

Jojo-Schmitz:
Anyway, this hasn't got anything to do with the issue this PR is fixing.

apricot2012:
> Now just the commit message (the body) doesn't fit anymore.
> 
> Unfortunatly this change leads to a crash, if the start and end syllable are in different staves (and the end one is at the end of the score and in a higher staff), it dies of a failed assertion:
> 
> https://github.com/musescore/MuseScore/blob/19a05847a5dcd2e4cce84a197f29652497d718cc/src/libmscore/select.cpp#L709
> 
> Independant of whether we'd use the 1st or the 2nd proposed fix, or none at all.
> Without it that selecting in the 'wrong' order doesn't work at all (not even if not including the last sylllable of a score), so this fix just uncovered another bug (but crashes on a debug build only).
> Maybe we should remove or modify that assert?
> I see nothing strictly illegal about `staffEnd > staffStart` not being the case.
> But while removing that condition from the `Q_ASSERT()` fixes the crash, it still doesn't make such a selection work :-(
> 
> Anyway, this hasn't got anything to do with the issue this PR is fixing.

Anything that I should do to correct the commit message? The updated commit message is only associated with my second commit.

Also did some tests around the issue mentioned above. Took a crack at a fix by adding this comparison in the `Score::selectRange(Element* e, int staffIdx)` function that reverses the two indexes if they don't satisfy idx2 > idx1:
```           
            int idx1 = selectedElement->staffIdx();
            int idx2 = e->staffIdx();
            if (idx2 < idx1) {
                int temp = idx1;
                idx1 = idx2;
                idx2 = temp;
            }
```

Surprisingly, with some elementary testing, this seems to resolve the issue.

Jojo-Schmitz:
Your commit message is fine, but you should squash those 2 commits into just one (and keep the 2nd message).
Fixing that other issue might be something for a separate commit or even separate PR?

Jojo-Schmitz:
But yers, that additional fix indeed seems to work fine

src/libmscore/score.cpp
Jojo-Schmitz, line 3421: Now you've changed the commit message to that of the 2nd fix, but actually kept the 1st fix ;-)

Jojo-Schmitz (approved):

apricot2012 (commented):
Oops, wrong button.

apricot2012:
> Your commit message is fine, but you should squash those 2 commits into just one (and keep the 2nd message).
> Fixing that other issue might be something for a separate commit or even separate PR?
> But yes, that additional fix indeed seems to work fine.

Committing the second fix now :D

Jojo-Schmitz (approved):

shoogle (approved):
Works for me!

PR #7794, opened 2021-03-24
[MU4] Added files statuses

igorkorsukov:
Added files statuses

### DIFF
--- a/src/autobot/autobottypes.h
+++ b/src/autobot/autobottypes.h
@@ -26,16 +26,9 @@
 #include "iteststep.h"
 
 namespace mu::autobot {
-enum class FileStatus {
-    Undefined = 0,
-    None,
-    Success,
-    Failed
-};
-
 struct File {
     io::path path;
-    FileStatus status = FileStatus::Undefined;
+    Ret ret;
 };
 
 using Files = std::vector<File>;
--- a/src/autobot/internal/autobot.h
+++ b/src/autobot/internal/autobot.h
@@ -36,24 +36,30 @@ public:
     std::vector<ITestCasePtr> testCases() const override;
     ITestCasePtr testCase(const std::string& name) const override;
 
-    void runAll(const std::string& testCaseName) override;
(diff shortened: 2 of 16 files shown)


src/autobot/autobottypes.h
```diff
@@ -26,16 +26,9 @@
...
-    Undefined = 0,
-    None,
-    Success,
-    Failed
-};
-
 struct File {
     io::path path;
-    FileStatus status = FileStatus::Undefined;
+    Ret ret;
```
RomanPudashkin, line 31: Could you explain why you need ret here?

src/autobot/internal/autobot.h
```diff
@@ -36,25 +36,31 @@ class Autobot : public IAutobot, public async::Asyncable
     std::vector<ITestCasePtr> testCases() const override;
     ITestCasePtr testCase(const std::string& name) const override;
 
-    void runAll(const std::string& testCaseName) override;
-    void runFile(const std::string& testCaseName, int fileIndex) override;
+    void setCurrentTestCase(const std::string& name) override;
+    const ValCh<ITestCasePtr>& currentTestCase() const override;
+
+    void runAll() override;
```
RomanPudashkin, line 42: Maybe is it better to rename this method to runAllFiles?

src/autobot/autobottypes.h
igorkorsukov: this is the result of the file test (in the future, there may be code and text message)

src/autobot/internal/autobot.h
igorkorsukov: renamed

RomanPudashkin (approved):

src/autobot/autobottypes.h
RomanPudashkin: I would prefer to rename this to "testCaseRet" then :) Or rename this struct to FileTestResult (or something similar on your choice :))

PR #7799, opened 2021-03-25
[MU3] Fix #319210: MusicXML import corrupted when xml is stripped from whitespace

Jojo-Schmitz:
Culprit seems an (empty) `<mordent/>` tag if that is not followed by a newline.

Resolves: https://musescore.org/en/node/319210

A corresponding PR for master is to follow, as soon as this PR here got @lvinken's official blessing ;-)

Edit: see #7809 for master


lvinken:
Fix is correct, thanks for looking into this. I am rather time-limited myself at the moment.

Jojo-Schmitz:
Thanks for the review.

PR #7801, opened 2021-03-26
[MU4] Fixed twice triggering of actions on menu bar

Eism:
Note from the documentation: This signal is emitted for the main parent menu in a hierarchy. Hence, only the parent menu needs to be connected to a slot; sub-menus need not be connected

### DIFF
--- a/src/appshell/view/dockwindow/dockmenubar.cpp
+++ b/src/appshell/view/dockwindow/dockmenubar.cpp
@@ -74,7 +74,9 @@ void DockMenuBar::updateMenus()
 {
     QList<QMenu*> menus;
     for (const QVariant& item: m_items) {
-        menus << makeMenu(item.toMap());
+        QMenu* menu = makeMenu(item.toMap());
+        connect(menu, &QMenu::triggered, this, &DockMenuBar::onActionTriggered);
+        menus << menu;
     }
 
     emit changed(menus);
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

vshalkevich:
Still reproduce crash on Windows 10

Eism:
Crash on libmscore, does not depend on the menu. It will be fixed in another PR

mattmcclinch:
For some reason, it is not behaving on macOS according to the documentation. Meaning that if the sub-menu is not connected to a slot, then nothing happens when the sub-menu is triggered. It turns out that `QMenu::addMenu()` does not set the parent of the child menu, which for some reason is only a problem on macOS. This can be fixed by making sure to explicitly set the parent of each menu. Then we can go back to simply connecting the menubar to the slot (as it was before #7782), instead of having to connect each menu. See https://github.com/mattmcclinch/MuseScore/commit/eaad17630a8de163cb6e1b36f9bdb427a3ee84ff.

shoogle:
I can confirm that the actions in the Add menu are not doing anything on macOS. @mattmcclinch, would you be able to submit a PR to fix the parenting issue? I think the devs are mostly on Linux so it's difficult for them to test whether that fixes it.

vshalkevich:
Good work

PR #7804, opened 2021-03-26
[MU3] Fix #319242, fix #319020:  Volta/Ottava/Pedal lines disappear when switching language

Jojo-Schmitz:
and while at it also remove a duplicate line and an unused variable

Resolves: https://musescore.org/en/node/319242 and https://musescore.org/en/node/319020

Unfortunately I wasn't able to reproduce the issue in my local build, without this change (but was able to in 3.6.2), so I'm not 100% sure it fixes the issue. User feedback and testing with the GitHub artifact myself indicates that it does not fix the issue :-(
Also whether https://musescore.org/en/node/319020 really is the same issue is unclear.

To me the changes do make perfect sense though and are worth being done in any case, as the current code obviously is just wrong resp. superfluous.

This does not apply to master (which correctly starts with `idx = 0` in both cases)

### DIFF
--- a/mscore/editstyle.cpp
+++ b/mscore/editstyle.cpp
@@ -80,7 +80,7 @@ EditStyle::EditStyle(Score* s, QWidget* parent)
       fbStyle->addButton(radioFBModern, 0);
       fbStyle->addButton(radioFBHistoric, 1);
 
-      int dta = 1;
+      int dta = 0;
       voltaLineStyle->clear();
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
As per user feedback this doesn't seem to fix this issue :-( (but is still a fix worth doing, as the currently code it really wrong resp. superfluous)

PR #7805, opened 2021-03-26
Fix #7774: Decreased the scrolling sensitivity in customize toolbar d…

### LINKED ISSUE #7774: [MU4 Issue] Scrolling too sensitive in customize toolbar dialog
**Describe the bug**
If user scroll down it takes on the very bottom of the list. There should be less sensitive scrolling

**To Reproduce**
1. Create a score
2. Click on "Customize toolbar" icon
3. Scroll down 

**Expected behavior**
Scrolling should be less sensitive. One scroll click down for one instrument down

**Screenshots**
Scrolling too sensitive in customize toolbar dialog

MacOS
Windows 10

Mystic-Slice:
…ialog

Resolves: https://github.com/musescore/MuseScore/issues/7774

An additional attribute - maximumFlickVelocity is added to control the scroll speed

### DIFF
--- a/src/notation/qml/MuseScore/NotationScene/NoteInputBarCustomiseDialog.qml
+++ b/src/notation/qml/MuseScore/NotationScene/NoteInputBarCustomiseDialog.qml
@@ -116,6 +116,8 @@ QmlDialog {
 
                 model: customiseModel
 
+                maximumFlickVelocity: 1000
+
                 boundsBehavior: Flickable.StopAtBounds
                 clip: true
 


vshalkevich:
I think we need to make common solution for a whole Scrolling not only for that dialog menu

vshalkevich:
Scrolling seems fine for dialog menu

vpereverzev:
> 
> 
> I think we need to make common solution for a whole Scrolling not only for that dialog menu

Then you should write it directly in the ticket, because this PR is fixing the exact case described in issue. We shouldn't expect more than we asking for.

src/notation/qml/MuseScore/NotationScene/NoteInputBarCustomiseDialog.qml
```diff
@@ -116,6 +116,8 @@ QmlDialog {
 
                 model: customiseModel
 
+                maximumFlickVelocity: 1000
```
vpereverzev, line 119: I guess we can specify some default value for "ui.theme" or even for StyledScrollBar component and use it here

@Eism since it's ur code, what would you prefer? :)

Jojo-Schmitz:
rebase needed

bkunda:
@vpereverzev Could we please have an update on this issue? I have found that context menus throughout the program are still scrolling with too much sensitivity. 

out-of-control-scroll

Happy to raise a new issue if needed. Let me know what you would prefer 🙂

cbjeukendrup:
Fixed by #10162

Issue #7772, opened 2021-03-21
[MU4 Issue] Can't scroll in popup (need to if it doesn't fit on the screen)


vshalkevich:
**Describe the bug**
Can't scroll in popup (need to if it doesn't fit on the screen)

**To Reproduce**
1. Open MuseScore with the lowest possible screen resolution
2. Create a score
3. Add a note
4. Select note
5. Open "Inspector" tab
6. Click on Note Settings
7. Click on head
8. Try to scroll down "netehead" menu

**Expected behavior**
Menues should be able to scroll up and down if they don`t fit in the screen

**Screenshots**
![2021-03-21 05_19_57-Window](https://user-images.githubusercontent.com/71372310/111891576-46aacb80-8a05-11eb-8fe0-2ecd5f5aac4f.jpg)

**Desktop (please complete the following information):**
MacOS
Windows 10

Mystic-Slice:
I am working on it.

PR #7807, opened 2021-03-27
[MU4] Added Scrollbar to increase accessibility for Head Settings

ArjunBoi:
Resolves: #7772 

Added a ScrollBar that allows the user to easily access previously inaccessible/clipped items in the following menu:
_*Inspector*_ -> _*Note Settings*_ (This setting appears when you have a note selected in the ScoreView) -> _*Head*_

[Previous Behaviour](https://github.com/ArjunBoi/ImagesAndFilesForLinks/blob/master/MuseScoreEdits/Old_HeadSettings.gif "Previous Behaviour") (Leads to a Gif i made, describing the issue visually)

[Edited Behaviour](https://github.com/ArjunBoi/ImagesAndFilesForLinks/blob/master/MuseScoreEdits/New_HeadSettings.gif "Edited Behaviour")

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/notation/notes/HeadSettings.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/notation/notes/HeadSettings.qml
@@ -1,5 +1,5 @@
 import QtQuick 2.0
-import QtQuick.Controls 2.0
+import QtQuick.Controls 2.15
 import MuseScore.Inspector 1.0
 import MuseScore.UiComponents 1.0
 import MuseScore.Ui 1.0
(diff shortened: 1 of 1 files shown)


ArjunBoi:
### Another Issue that i have run into with the code in this D-PR:

Having implemented the scrollbar, the new view of the drop-down (Show More) menu in the _Head_ Settings is clipped ever so slightly from the left side (only a few pixels).

I can't seem to figure out what i'm doing wrong or what could be the solution to this problem. I have already tried messing with the alignment, anchor(s) and clip settings, but there doesn't seem to be any perfect solution to it. 
Setting _clip_ to `false` does seem to fix the dropdown view, but it breaks the rest of the alignment completely, so that is definitely not the way to go...

Any kind of help/insight is welcome 🙌🏻 

Old, Properly Clipped View: (The drop-down triangle is fully visible)

image



New, Awkwardly Clipped View: (Notice that only half of the drop-down triangle is visible)

image

src/inspector/view/qml/MuseScore/Inspector/notation/notes/HeadSettings.qml
```diff
@@ -11,200 +11,206 @@ FocusableItem {
...
+        width: parent.width + 10
+        clip: true
 
-        width: parent.width
+        Column {
+            id: contentColumn
+            width: root.width - (2 * anchors.leftMargin)
+            spacing: 16
+            anchors.left: parent.left
+            anchors.leftMargin: 5
```
Eism, line 27: we use values that are multiples of 4 (4, 8, 12, 16 ...) in the margins, spacings, width, etc.

Eism (commented):
The main thing is that we have three such tabs, but the fix was made only in one of them. I like this change :) But we also need to make the same behavior for all tabs. @Tantacrul What do you think about this?

ArjunBoi:
> The main thing is that we have three such tabs, but the fix was made only in one of them. I like this change :) But we also need to make the same behavior for all tabs.

Understood!
I have changed the status of this PR again to a Draft PR because this solution may need a complete rework, especially if we are seeking a more general solution that will (ideally by default) apply the ScrollView to all 3 tabs.

I do fully intend to see this solution through!, but I have got my midterm exams in a couple of days, after which i will be focussing on my GSoC Proposal, so progress and updates for this PR will be significantly slowed.

PR #7810, opened 2021-03-27
[MU4] vst_ressurection_part_1

vpereverzev:
Unfortunately, it's turned out that our VST module is not working at all. We're even unable to properly launch plugin's editor.

This PR introduces a new VST module implementation, which is much more simple and uses a full capacity of VST3 SDK.

Basicly, the current implementation gives us an ability to launch a 'native' plugin editor. In the further PRs I'm going to "ressurect" audio handling via VST plugins.

Here is an example of launching "Helm" VST3 plugin, tested on Win/Mac/Lin:

vst_demo_new


Marr11317:
why did you close it?

PR #7812, opened 2021-03-28
[MU4] Fixed shortcuts activating when editing notation text

aaaeka:
Previously, when editing notation text and pressing a shortcut key, for example "I", the instruments panel would open. That would make it impossible to write the letter "i".

### DIFF
--- a/src/instruments/view/instrumentpaneltreemodel.cpp
+++ b/src/instruments/view/instrumentpaneltreemodel.cpp
@@ -146,6 +146,11 @@ void InstrumentPanelTreeModel::selectRow(const QModelIndex& rowIndex)
 
 void InstrumentPanelTreeModel::addInstruments()
 {
+    notation::INotationInteractionPtr interaction = context()->currentNotation()->interaction();
+    if (interaction->isTextEditingStarted()) {
+        return;
+    }
+
     auto mode = ISelectInstrumentsScenario::SelectInstrumentsMode::ShowCurrentInstruments;
     RetVal<InstrumentList> selectedInstruments = selectInstrumentsScenario()->selectInstruments(mode);
     if (!selectedInstruments.ret) {


src/framework/shortcuts/internal/shortcutscontroller.h
```diff
@@ -24,13 +24,15 @@
...
 #include "actions/iactionsregister.h"
+#include "context/iglobalcontext.h"
 
 namespace mu::shortcuts {
 class ShortcutsController : public IShortcutsController
 {
     INJECT(shortcuts, IShortcutsRegister, shortcutsRegister)
     INJECT(shortcuts, actions::IActionsDispatcher, dispatcher)
     INJECT(shortcuts, actions::IActionsRegister, aregister)
+    INJECT(shortcuts, context::IGlobalContext, globalContext)
```
RomanPudashkin, line 35: The framework will depend on an external module after this change. Probably the issue should be fixed in the notation module

aaaeka: Ok, will fix it.

src/instruments/view/instrumentpaneltreemodel.cpp
```diff
@@ -146,6 +146,11 @@ void InstrumentPanelTreeModel::selectRow(const QModelIndex& rowIndex)
 
 void InstrumentPanelTreeModel::addInstruments()
 {
+    notation::INotationInteractionPtr interaction = context()->currentNotation()->interaction();
+    if (interaction->isTextEditingStarted()) {
+        return;
+    }
```
igorkorsukov, line 152: I believe that we should check for the current `ShortcutContext` value (and extend and develop it), not add dependencies

`ShortcutContext GlobalContext::currentShortcutContext() const`

aaaeka: Assigning a value like ```NotationTextEditing``` only works for shortcuts that need to work only when editing text. What we need here is the opposite of that? Not sure how I should go about implementing this.

igorkorsukov: I am now writing documentation on shortcuts, and I found that, in fact, this system is broken. I ask you to pause work on this task for now. We (the in-house team) will fix the shortcut system, and then we will fix this issue

igorkorsukov: Thank you for noticing this issue.

aaaeka: No problem.

igorkorsukov:
not needs now, fixed by https://github.com/musescore/MuseScore/pull/7858 :)

PR #7814, opened 2021-03-29
Export Dialog Solution B

cbjeukendrup:
Prototype of Export Dialog, with the UI implemented "manually".
This solution might require a bit more maintenance, but it is much easier and, most importantly, gives all flexibility regarding how we want it to look. This solution has therefore my strong preference. 

~Note: most things already work, however the code quality is not optimal yet, and there are still some things that need to be added.~

Status: Ready for review!


cbjeukendrup:
@Eism Could you please do a review?

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+{
+    return context()->currentMasterNotation();
+}
+
+bool ExportDialogModel::isMainNotation(INotationPtr notation) const
+{
+    return notation == masterNotation()->notation();
+}
+
+bool ExportDialogModel::isNotationIndexValid(int index) const
```
Eism, line 148: just isIndexValid

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+void ExportDialogModel::setExportSuffix(QString suffix)
+{
+    if (m_selectedExportSuffix == suffix) {
+        return;
+    }
+
+    m_selectedExportSuffix = suffix;
+    emit selectedExportSuffixChanged(suffix);
+
+    setUnitType(static_cast<int>(exportScoreService()->supportedUnitTypes(suffix.toStdString())[0]));
```
Eism, line 172: need to check for empty of types

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+}
+
+QList<int> ExportDialogModel::availableUnitTypes() const
+{
+    QList<int> result;
+    for (notation::INotationWriter::UnitType t : exportScoreService()->supportedUnitTypes(m_selectedExportSuffix.toStdString())) {
```
Eism, line 178: better rename to type

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+        result << static_cast<int>(t);
+    }
```
Eism, line 180: please add a new line here

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+bool ExportDialogModel::exportScores()
+{
+    INotationPtrList notations;
+
+    for (const QModelIndex& index: m_selectionModel->selectedIndexes()) {
+        notations.push_back(m_notations[index.row()]);
+    }
+
+    IF_ASSERT_FAILED(!notations.empty()) {
```
Eism, line 212: we can call the export score method from qml if we did not select any scores?

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+        return false;
+    }
+
+    /// If only one file will be created, the filename will be exactly what the user
+    /// types in the save dialog and therefore we can put the file dialog in charge of
+    /// asking the user whether an existing file should be overridden. Otherwise, we
+    /// will take care of that ourselves.
+    bool willCreateOneFile = exportScoreService()->willCreateOnlyOneFile(
```
Eism, line 220: I think, we can move all this code to exportService and call just exportScoreService()->exportScores(notations,  m_selectedUnitType)

src/userscores/internal/exportscoreservice.h
```diff
@@ -0,0 +1,72 @@
...
+#include "iuserscoresconfiguration.h"
+#include "iinteractive.h"
+#include "notation/inotationwritersregister.h"
+#include "importexport/imagesexport/iimagesexportconfiguration.h"
+#include "context/iglobalcontext.h"
+#include "notation/inotation.h"
+#include "system/ifilesystem.h"
+
+namespace mu::userscores {
+class ExportScoreService : public IExportScoreService
```
Eism, line 35: better rename this class to ExportScoreScenario because in this class you interact with the user, not only get set some data

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+    ExportOptionItem {
+        firstColumnWidth: root.firstColumnWidth
+        text: qsTrc("userscores", "Format:")
+
+        StyledComboBox {
+            id: fileTypeComboBox
+            Layout.fillWidth: true
+
+            model: [
+                { textRole: qsTrc("userscores", "PDF File"), valueRole: "pdf" },
```
Eism, line 24: better to move it to the model

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+                { textRole: qsTrc("userscores", "MusicXML"), valueRole: "xml" }
+            ]
+
+            maxVisibleItemCount: count
+
+            textRoleName: "textRole"
+            valueRoleName: "valueRole"
+
+            currentIndex: {
+                if (exportModel.selectedExportSuffix === "musicxml" || exportModel.selectedExportSuffix === "mxl") {
```
Eism, line 41: Can you explain to me,  why you need this condition? and can we move it to the model?

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+                    exportModel.selectedExportSuffix = "mxl";
+                } else {
+                    exportModel.selectedExportSuffix = value
+                }
+            }
+        }
+    }
+
+    ExportOptionItem {
+        visible: exportModel.selectedExportSuffix === "pdf"
```
Eism, line 59: I think we need the enum on the model for formats

Eism: Also better to add a loader instead of creating all pages for formats and setting visible by changing the format

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+    ColumnLayout {
+        id: audioOptions
+        spacing: 12
+
+        visible: ["mp3", "wav", "ogg", "flac"].includes(exportModel.selectedExportSuffix)
+
+        CheckBox {
+            text: qsTrc("userscores", "Normalize")
+            checked: exportModel.normalizeAudio
+            onClicked: exportModel.normalizeAudio = !checked
```
Eism, line 126: Please add braces

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
Eism: And move all pages to separate files

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+
+        ExportOptionItem {
+            text: qsTrc("userscores", "File type:")
+            firstColumnWidth: root.firstColumnWidth
+
+            StyledComboBox {
+                id: xmlTypeComboBox
+                Layout.fillWidth: true
+
+                model: [
```
Eism, line 213: better to move it to the model

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+                currentIndex: indexOfValue(exportModel.selectedExportSuffix)
+                onValueChanged: exportModel.selectedExportSuffix = value
+            }
+        }
+
+        RadioButtonGroup {
+            spacing: 12
+            orientation: Qt.Vertical
+            Layout.fillWidth: true
+            model: [
```
Eism, line 231: better to move it to the model

src/userscores/qml/MuseScore/UserScores/internal/ExportOptionsView.qml
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+
+    RadioButtonGroup {
+        Layout.fillWidth: true
+        visible: model.length > 1
+        spacing: 12
+        orientation: Qt.Vertical
+
+        model: {
```
Eism, line 253: better to move it to the model

src/userscores/view/exportdialogmodel.h
```diff
@@ -0,0 +1,176 @@
...
+    bool isNotationIndexValid(int index) const;
+
+    notation::IMasterNotationPtr masterNotation() const;
+
+    uicomponents::ItemMultiSelectionModel* m_selectionModel = nullptr;
+    QList<notation::INotationPtr> m_notations;
+
+    QString exportFilter() const;
+
+    QString m_selectedExportSuffix = "pdf";
```
Eism, line 171: Please remove the hardcoded value

src/userscores/view/exportdialogmodel.h
```diff
@@ -0,0 +1,176 @@
...
+    bool isMainNotation(notation::INotationPtr notation) const;
+
+    bool isNotationIndexValid(int index) const;
+
+    notation::IMasterNotationPtr masterNotation() const;
+
+    uicomponents::ItemMultiSelectionModel* m_selectionModel = nullptr;
+    QList<notation::INotationPtr> m_notations;
+
+    QString exportFilter() const;
```
Eism, line 169: Please move methods above members

src/userscores/view/exportdialogmodel.cpp
cbjeukendrup: The button is (should be) disabled when no scores are selected. But I will change into a normal `if`.

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+        return;
+    }
+
+    QModelIndex modelIndex = index(scoreIndex);
+    m_selectionModel->QItemSelectionModel::select(modelIndex, selected ? QItemSelectionModel::Select : QItemSelectionModel::Deselect);
+
+    emit dataChanged(modelIndex, modelIndex);
+}
+
+void ExportDialogModel::toggleSelected(int scoreIndex)
```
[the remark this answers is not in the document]
cbjeukendrup: Toggling: "flips" the state (selected becomes unselected, unselected becomes selected)
Setting: selects or unselects the item as specified, no matter if it _was_ selected or not.

`toggleSelected(idx)` == `setSelected(idx, !isSelected(idx));`

src/userscores/view/exportdialogmodel.cpp
```diff
@@ -0,0 +1,473 @@
...
+}
+
+void ExportDialogModel::setSelected(int scoreIndex, bool selected)
+{
+    if (!isNotationIndexValid(scoreIndex)) {
+        return;
+    }
+
+    QModelIndex modelIndex = index(scoreIndex);
+    m_selectionModel->QItemSelectionModel::select(modelIndex, selected ? QItemSelectionModel::Select : QItemSelectionModel::Deselect);
```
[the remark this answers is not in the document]
cbjeukendrup: We can't, and actually we should not use ItemMultiSelectionModel at all. It is meant for a different situation, namely a list view where the user can use keyboard modifiers to control what will happen. In this case, keyboard modifiers should not be considered, but it's a matter of checkboxes.

cbjeukendrup:
@Eism Thank you for the comprehensive review! I have fixed the trivial things, and I will now continue working on a better way of handling the export types: maybe an enum, maybe a struct containing things like allowed suffixes, user-readable name, file filter etc.

src/userscores/view/exportdialogmodel.cpp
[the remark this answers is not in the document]
Eism: I think toggleSelected is an unnecessary method. Just call setSelected(index, selected) from qml

src/userscores/view/exportdialogmodel.cpp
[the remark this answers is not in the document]
Eism: looks better now

cbjeukendrup:
I've finished fixing the code review issues!

src/importexport/imagesexport/internal/pdfwriter.h
```diff
@@ -31,10 +31,8 @@ class PdfWriter : public notation::AbstractNotationWriter
     INJECT(iex_imagesexport, IImagesExportConfiguration, configuration)
 
 public:
-    std::vector<INotationWriter::UnitType> supportedUnitTypes() const override;
-    Ret write(const notation::INotationPtr notation, system::IODevice& destinationDevice, const Options& options = Options()) override;
-    Ret writeList(const notation::INotationPtrList& notations, system::IODevice& destinationDevice,
-                  const Options& options = Options()) override;
+    std::vector<notation::WriterUnitType> supportedUnitTypes() const override;
+    Ret write(const notation::INotationPtrList& notations, system::IODevice& destinationDevice,const Options& options = Options()) override;
```
cbjeukendrup, line 35: Uncrustify bug: if I write on one line
```
Ret write(const notation::INotationPtrList& notations, system::IODevice& destinationDevice, const Options& options = Options()) override;
```
it wants me to change it into 
```
Ret write(const notation::INotationPtrList& notations, system::IODevice& destinationDevice, 
          const Options& options = Options()) override;
```
And if I do that, it wants me to change it into 
```
Ret write(const notation::INotationPtrList& notations, system::IODevice& destinationDevice,const Options& options = Options()) override;
```
So that is what I did, although it's not beautiful without a space after that comma.

igorkorsukov: I think, we may rename `destinationDevice` to like `destination` or `dest` or `destDevice` to correspond the line width constraint :)

cbjeukendrup: We might also change this line:
https://github.com/musescore/MuseScore/blob/c111b1d1f482dea0cc8166d072b967eb07cefa38/tools/codestyle/uncrustify_musescore.cfg#L429
I think it would be okay to have a space after all commas.

igorkorsukov: Yes, I agree with you

cbjeukendrup: Should I go for "add" or "force"?

igorkorsukov: I think better "force", but need testing

Eism: Let's make a separate PR if there are many fixes

cbjeukendrup: I think "add" appears to be better. "force" would screw up things like https://github.com/musescore/MuseScore/blob/c111b1d1f482dea0cc8166d072b967eb07cefa38/src/importexport/guitarpro/internal/importgtp-gp6.cpp#L69 and https://github.com/musescore/MuseScore/blob/c111b1d1f482dea0cc8166d072b967eb07cefa38/src/notation/view/widgets/editstyle.cpp#L95
I don't think we really want that :)

igorkorsukov: Let's use "add"

cbjeukendrup: Here's the PR: https://github.com/musescore/MuseScore/pull/7961

cbjeukendrup:
There is a problem with the CheckBoxes in ListItemBlank in the scores list view. The mouse area from ListItemBlank (=FocusableControl) seems to be on top of the mouse area of the checkboxes, causing the checkboxes to be never hovered or triggered. 

We could solve it by changing the contentItem of FocusableControl like this:
```
    Item {
        id: contentItem
        anchors.fill: focusRectItem
        anchors.margins: 2 //! NOTE margin needed to show focus border
        z: mouseAreaItem.z + 1
    }
```
This makes sure that any child mouse areas are prioritized over the one provided by FocusableControl.

Maybe the following would also work: Don’t “capture” the mouse event. 
Like this:
```
    MouseArea {
        id: mouseAreaItem
        anchors.fill: parent

        onClicked: {
            root.insureActiveFocus()
            mouse.accepted = false
        }
    }
```
But this way, `mouse.accepted = false` will have to be copied to all places where `mousearea.onClicked` is overridden, which is a bit tedious…

igorkorsukov:
> ```qml
> MouseArea
> ```

Needs set `propagateComposedEvents = true` 
```
MouseArea {
        id: mouseAreaItem
        propagateComposedEvents: true
```

cbjeukendrup:
@igorkorsukov Thanks, the combination of `propagateComposedEvents: true` and `mouse.accepted = false` works for clicking. However, hovering does not work...

igorkorsukov:
> @igorkorsukov Thanks, the combination of `propagateComposedEvents: true` and `mouse.accepted = false` works for clicking. However, hovering does not work...

Let's make a working сlick, I'll be researching this issue later.

igorkorsukov:
> @igorkorsukov Thanks, the combination of `propagateComposedEvents: true` and `mouse.accepted = false` works for clicking. However, hovering does not work...

I thought we just need to move the `MouseArea` above the content to `FocusableControl`. And we should make sure that  `MouseArea` is not created in the heirs for the entire area to implement the click for all item (we need to use `MouseArea` from `FocusableControl`) and then everything will be fine.

cbjeukendrup:
@igorkorsukov I rebased and indeed since commit https://github.com/musescore/MuseScore/commit/26fdf4d66cd8cc241638ab107f183117fbe2b1ca the checkboxes are working perfectly. So this PR is ready now!

src/notation/notationtypes.h
```diff
@@ -166,6 +166,12 @@ enum class SaveMode
     SaveOnline
 };
 
+enum class WriterUnitType {
```
Eism, line 169: It is better to move it to INotationWriter

Eism (approved):
Good job Casper!
I am going to merge this pull request, please fix review issues in a separate PR

cbjeukendrup:
Thanks Elnur! I will do the fixes soon.

src/userscores/qml/MuseScore/UserScores/internal/Export/AudioSettingsPage.qml
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+import QtQuick 2
```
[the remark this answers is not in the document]
cbjeukendrup: Yes! I watched https://www.youtube.com/watch?v=JbSNobpE_sM, and they told that using minor versions will be discouraged from Qt 5.15 and will be unsupported from Qt 6 / QML 3.

Issue #7828, opened 2021-03-30
[MU4 Issue] Font size causes crash when manually edited


vshalkevich:
**Describe the bug**
Font size causes crash when manually edited

**To Reproduce**
1. Click on Edit from the upper Menu Bar
2. Click on 'Preferences
3. Click on Appearance from the left panel
4. Try manyally change Body Text Size field

**Expected behavior**
No crash. Body Text Size changes

**Screenshots**
https://user-images.githubusercontent.com/71372310/112928033-9f254b80-911e-11eb-86c2-423fd3c63946.mp4

**Desktop (please complete the following information):**
MacOS
Windows 10

**Additional context**

PR #7831, opened 2021-03-30
Fix GH #7828: Fix input validator in IncrementalPropertyControl

cbjeukendrup:
Resolves: #7828

Use IntInputValidator instead of DoubleInputValidator when `decimals` is zero.
(This commit was part of PR #7814, but I decided to make a separate PR to have this issue fixed quickly.)

Also fixes some Clang compiler warnings.

### DIFF
--- a/src/libmscore/draw/drawjson.cpp
+++ b/src/libmscore/draw/drawjson.cpp
@@ -197,7 +197,7 @@ static void fromObj(const QJsonObject& obj, QPainterPath& path)
 
     QJsonArray elsArr = obj["elements"].toArray();
     std::vector<QPainterPath::Element> curveEls;
-    for (const QJsonValue& elVal : elsArr) {
+    for (const QJsonValue elVal : elsArr) {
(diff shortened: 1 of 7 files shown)


src/libmscore/draw/drawjson.cpp
```diff
@@ -197,7 +197,7 @@ static void fromObj(const QJsonObject& obj, QPainterPath& path)
 
     QJsonArray elsArr = obj["elements"].toArray();
     std::vector<QPainterPath::Element> curveEls;
-    for (const QJsonValue& elVal : elsArr) {
+    for (const QJsonValue elVal : elsArr) {
```
RomanPudashkin, line 200: Can you please explain why you removed references?

cbjeukendrup: Clang says: "Loop variable 'elVal' is always a copy because the range of type 'QJsonArray' does not return a reference"

RomanPudashkin (approved):

PR #7833, opened 2021-03-30
[MU4] vst_ressurection_part1

vpereverzev:
Unfortunately, it's turned out that our VST module is not working at all. We've been unable to properly launch plugin's editor for a while.

This PR introduces a new VST module implementation, which is much more simple and uses a full capacity of VST3 SDK.

Basicly, the current implementation gives us an ability to launch a 'native' plugin editor. In the further PRs I'm going to "ressurect" audio handling via VST plugins.

Here is an example of launching "Helm" VST3 plugin, tested on Win/Mac/Lin:

112736001-1957a380-8f58-11eb-9d95-6e8e4ae227d6

The next step is to properly handle audio using VST3 SDK


src/framework/vst/internal/vstconfiguration.cpp
```diff
@@ -1,43 +1,44 @@
...
+{
+#ifdef Q_OS_MAC
+    return Steinberg::kPlatformTypeNSView;
+#elif defined(Q_OS_IOS)
+    return Steinberg::kPlatformTypeUIView;
+#elif defined(Q_OS_WIN)
+    return Steinberg::kPlatformTypeHWND;
+#elif defined(Q_OS_LINUX)
+    return Steinberg::kPlatformTypeX11EmbedWindowID;
+#endif
```
RomanPudashkin, line 43: It would be great to add some default return here

RomanPudashkin (approved):

Eism (approved):

src/framework/vst/ivstconfiguration.h
```diff
@@ -1,43 +1,41 @@
...
+namespace mu::vst {
+class IVstConfiguration : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IVstConfiguration)
+
+public:
+    virtual ~IVstConfiguration() = default;
+
+    virtual mu::io::path customSearchPath() const = 0;
+    virtual Steinberg::FIDString currentPlatformType() const = 0;
```
igorkorsukov, line 37: I would prefer that third-party lib headers do not appear in `.h` files, this often making problems during compilation. In this case, it would be good to add an enumerator with the platform and a method to get it to the `IApplication` interface and use it directly or through this configurator interface, i.e. it would return `IApplication::Platform`

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+using namespace mu;
+using namespace mu::vst;
+
+VstPlugin::VstPlugin()
+    : m_factory(nullptr)
+{
+}
+
+Ret VstPlugin::load(const std::string& pluginPath)
```
igorkorsukov, line 32: better use `io::path`

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+    PluginContextFactory::instance().setPluginContext(&m_pluginContext);
+
+    std::string errorString;
+
+    m_module = PluginModule::create(pluginPath, errorString);
+
+    if (!m_module) {
+        LOGE() << errorString;
+        return make_ret(Err::NoPluginModule);
```
igorkorsukov, line 42: better use `make_ret(Err::NoPluginModule, errorString)` (create it, if not present :) )

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    m_factory = m_module->getFactory();
+
+    if (!m_factory.get()) {
+        return make_ret(Err::NoPluginFactory);
+    }
+
+    for (const ClassInfo& classInfo : m_factory.classInfos()) {
+        if (classInfo.category() != kVstAudioEffectClass) {
+            continue;
```
igorkorsukov, line 53: I think would be like add LOGI message, about that audio not supported now

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -0,0 +1,123 @@
...
+    m_pluginController = owned(m_pluginProvider->getController());
+
+    if (!m_pluginController) {
+        return make_ret(Err::NoPluginController);
+    }
+
+    return Ret(true);
+}
+
+std::string VstPlugin::id() const
```
igorkorsukov, line 73: I believe, better add separate type for ID (using, typedef)

src/framework/vst/internal/vstpluginrepository.cpp
```diff
@@ -0,0 +1,136 @@
...
+using namespace vst;
+
+//@see https://developer.steinberg.help/pages/viewpage.action?pageId=9798275
+static const std::string VST3_PACKAGE_EXTENSION = "*.vst3";
+
+Ret VstPluginRepository::loadAvailablePlugins()
+{
+    Ret result(true);
+
+    for (const std::string& pluginPath : pluginPathsFromDefaultLocation()) {
```
igorkorsukov, line 36: better use `io::path`

src/framework/vst/internal/vstpluginrepository.h
```diff
@@ -0,0 +1,61 @@
...
+
+private:
+    RetVal<io::paths> pluginPathsFromCustomLocation(const io::path& customPath) const;
+    PluginModule::PathList pluginPathsFromDefaultLocation() const;
+
+    Ret instantiatePlugin(const std::string& path);
+
+    VstPluginMetaList availablePluginsMetaList() const;
+
+    std::unordered_map<std::string, VstPluginPtr> m_pluginsMap;
```
igorkorsukov, line 55: definitely need a separate type for ID

src/framework/vst/sdk/CMakeLists.txt
```diff
@@ -17,42 +17,146 @@
...
 
-set(MODULE_SRC
-    ${VST3_SDK_PATH}/base/source/fobject.cpp
+include(GetPlatformInfo)
+
+if (MUSESCORE_BUILD_CONFIG EQUAL "release")
+    set(MODULE_DEF -DRELEASE)
+else()
+    set(MODULE_DEF -DDEVELOPMENT)
+endif()
```
igorkorsukov, line 32: better use 
```
include(GetBuildType)
if (BUILD_IS_DEBUG) 
...
```

src/framework/vst/internal/vstpluginrepository.cpp
vpereverzev: std::string is coming from Steinberg::Vst::PathList here, I'd leave it as it is since it's a private function

src/framework/vst/ivstconfiguration.h
vpereverzev: I'm actually thinking that it was a bad idea to put it here (in configuration) and the name of the method is not really good, since it's not really a "platform type", but a UI api. I'd just move it closer to a customer of this method, because the use case is really rare

src/framework/vst/ivstpluginrepository.h
```diff
@@ -1,49 +1,42 @@
...
+namespace mu::vst {
+class IVstPluginRepository : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IVstPluginRepository)
+
+public:
+    virtual ~IVstPluginRepository() = default;
+
+    virtual Ret loadAvailablePlugins() = 0;
+    virtual RetVal<VstPluginPtr> findPluginById(const std::string& id) const = 0;
```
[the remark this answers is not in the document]
vpereverzev: I'd leave it like this

src/framework/vst/sdk/CMakeLists.txt
vpereverzev: cool, much better

src/framework/vst/internal/vstconfiguration.cpp
vpereverzev: I think it was a bad idea to put this method here, anyway it worth to add default value. Thx

src/framework/vst/internal/vstplugin.cpp
vpereverzev: yep, replaced

src/framework/vst/internal/vstplugin.cpp
vpereverzev: meh, I'm too lazy :)
I'll leave it that way, but will add an additional output to the olg below

src/framework/vst/internal/vstplugin.cpp
vpereverzev: agreed

src/framework/vst/internal/vstplugin.cpp
vpereverzev: agreed, thx

PR #7837, opened 2021-03-30
[MU4] Fix submenus on macOS

mattmcclinch:
Resolves: https://github.com/musescore/MuseScore/pull/7801#issuecomment-808331900

QMenu::addMenu() does not actually set the parent of the child menu, which on macOS means that the parent menu is not triggered when an action on the child menu is triggered. This PR makes sure to set the parent of each menu, so that the menubar will be triggered when any menu or submenu inside it is triggered.

### DIFF
--- a/src/appshell/view/dockwindow/dockmenubar.cpp
+++ b/src/appshell/view/dockwindow/dockmenubar.cpp
@@ -75,7 +75,6 @@ void DockMenuBar::updateMenus()
     QList<QMenu*> menus;
     for (const QVariant& item: m_items) {
         QMenu* menu = makeMenu(item.toMap());
-        connect(menu, &QMenu::triggered, this, &DockMenuBar::onActionTriggered);
         menus << menu;
     }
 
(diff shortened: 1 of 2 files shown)


shoogle (approved):
Works for me on macOS. Actions from the Add menu are performed exactly once.

vshalkevich:
Works fine for me

Shivansh-Bakshi:
This breaks it for me on Windows. I reverted the changes from this commit locally and it works but with these changes, sub menu options don't show up.

shoogle:
@Shivansh-Bakshi, thanks. @vshalkevich, which platform did you test on? We should probably try to coordinate in future.

I'm quite busy right now, but I'll have a look at this in a few days if nobody else gets there first. I have access to Windows, macOS and Linux so I should be able to find something that works on all three even if it involves some conditional coding based on platform.

vshalkevich:
I checked on MacOS

mattmcclinch:
I knew I should have tested this on Windows and Linux also, since I have access to those as well. I am going to try to investigate why we have to jump through these hoops for macOS, but in the meantime, we can surround the new `setParent()` calls with 
`#ifdef Q_OS_MAC`

`#endif`
See https://github.com/mattmcclinch/MuseScore/commit/57317c7439f4218057f95f49b1a37342bb4d362b.

Issue #7656, opened 2021-03-06
[MU4 Issue] Inspector panel doesnt reflects on changes from View panel


vshalkevich:
**Describe the bug**
"Page Margins" checkbox without tick

**To Reproduce**
1. Create a score
2. Click View
3. Tick "Show page margins"
4. Open Inspector tab from instuments panel
5. See "Page Margins" checkbox

**Expected behavior**
"Page Margins" checkbox without tick

**Screenshots**
![image](https://user-images.githubusercontent.com/71372310/110194558-d66b4a00-7e49-11eb-8c16-7b0d66c1710d.png)
![image (1)](https://user-images.githubusercontent.com/71372310/110194559-d703e080-7e49-11eb-90de-093c8f027e5e.png)

**Desktop (please complete the following information):**
Windows

**Additional context**

Howard-C:
Dibs!

Howard-C:
Dibs!

Howard-C:
> Dibs!

Well, turns out the show invisible/unprintable/... are not `Sid`s, it should be a whole new system other than `NotationStyle`. I'll leave the in-house developers to that :)

vshalkevich:
Page Margins from upper menu connected with Page Margin tick from the Inspector (Properties tab for now). That seems good 
Need to double-check Page Margin function.

PR #7841, opened 2021-03-30
[MU4] Update display options in inspector from the view menu panel

aaaeka:
Resolves: #7656

Also adds functionality to the display checkboxes in inspector when nothing is selected.

### DIFF
--- a/src/inspector/internal/compatibility/mu4inspectoradapter.h
+++ b/src/inspector/internal/compatibility/mu4inspectoradapter.h
@@ -27,13 +27,16 @@
 #include "modularity/ioc.h"
 #include "context/iglobalcontext.h"
 #include "notation/inotation.h"
+#include "notation/notationtypes.h"
 #include "iinteractive.h"
+#include "actions/iactionsdispatcher.h"
 
 namespace mu::inspector {
 class MU4InspectorAdapter : public IInspectorAdapter
 {
     INJECT(inspector, mu::context::IGlobalContext, context)
     INJECT(inspector, mu::framework::IInteractive, interactive)
+    INJECT(inspector, mu::actions::IActionsDispatcher, dispatcher)
 
 public:
     MU4InspectorAdapter() = default;
@@ -41,6 +44,8 @@ public:
     bool isNotationExisting() const override;
     bool isTextEditingStarted() const override;
     async::Notification isTextEditingChanged() const override;
+    notation::ScoreConfig scoreConfig() const override;
+    async::Channel<notation::ScoreConfigType> scoreConfigChanged() const override;
 
     // notation commands
     void beginCommand() override;
@@ -60,15 +65,16 @@ public:
     void showGridConfigurationDialog() override;
 
     // actions
-    void updatePageMarginsVisibility(const bool isVisible) override;
-    void updateFramesVisibility(const bool isVisible) override;
     void updateHorizontalGridSnapping(const bool isSnapped) override;
     void updateVerticalGridSnapping(const bool isSnapped) override;
-    void updateUnprintableElementsVisibility(const bool isVisible) override;
-    void updateInvisibleElementsDisplaying(const bool isVisible) override;
+    void toggleInvisibleElementsDisplaying() override;
+    void toggleUnprintableElementsVisibility() override;
+    void toggleFramesVisibility() override;
+    void togglePageMarginsVisibility() override;
 
     // notation layout
     void updateNotation() override;
+    async::Notification currentNotationChanged() const override;
 
 private:
     mu::notation::INotationUndoStackPtr undoStack() const;
--- a/src/inspector/models/score/scoredisplaysettingsmodel.cpp
+++ b/src/inspector/models/score/scoredisplaysettingsmodel.cpp
@@ -29,6 +29,13 @@ ScoreSettingsModel::ScoreSettingsModel(QObject* parent, IElementRepositoryServic
     setSectionType(InspectorSectionType::SECTION_SCORE_DISPLAY);
     setTitle(qtrc("inspector", "Show"));
     createProperties();
+
+    m_shouldShowInvisible = adapter()->isNotationExisting() ? adapter()->scoreConfig().isShowInvisibleElements : true;
+    m_shouldShowUnprintable = adapter()->isNotationExisting() ? adapter()->scoreConfig().isShowUnprintableElements : true;
+    m_shouldShowFrames = adapter()->isNotationExisting() ? adapter()->scoreConfig().isShowFrames : true;
+    m_shouldShowPageMargins = adapter()->isNotationExisting() ? adapter()->scoreConfig().isShowPageMargins : false;
+
+    setupConnections();
 }
 
 void ScoreSettingsModel::createProperties()
@@ -60,9 +67,9 @@ void ScoreSettingsModel::loadProperties()
 void ScoreSettingsModel::resetProperties()
 {
     setShouldShowInvisible(false);
+    setShouldShowUnprintable(false);
     setShouldShowFrames(false);
     setShouldShowPageMargins(false);
-    setShouldShowUnprintable(false);
 }
 
 bool ScoreSettingsModel::shouldShowInvisible() const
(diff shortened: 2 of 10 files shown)


aaaeka:
Also resolves: #7847

RobFog:
Typo in commit message: "inpector"

src/inspector/models/score/scoredisplaysettingsmodel.cpp
```diff
@@ -106,8 +105,82 @@ void ScoreSettingsModel::setShouldShowPageMargins(bool shouldShowPageMargins)
...
 
-    adapter()->updatePageMarginsVisibility(shouldShowPageMargins);
+    adapter()->updatePageMarginsVisibility();
+
+    updateShouldShowPageMargins(shouldShowPageMargins);
+}
+
+void ScoreSettingsModel::updateShouldShowInvisible(bool isVisible)
+{
+    m_shouldShowInvisible = isVisible;
```
RomanPudashkin, line 115: Better to add a check here (same for methods below) that the value is actually changed:

if (isVisible == m_shouldShowInvisible) {
return;
}

m_shouldShowInvisible = isVisible;
emit shouldShowInvisibleChanged(isVisible);

src/inspector/models/score/scoredisplaysettingsmodel.h
```diff
@@ -40,10 +42,19 @@ public slots:
...
+    void updateShouldShowUnprintable(bool isVisible);
+    void updateShouldShowFrames(bool isVisible);
+    void updateShouldShowPageMargins(bool isVisible);
+
+    void updateFromConfig(mu::notation::ScoreConfigType configType);
+    void updateAll();
+    void setupConnections();
+
+    bool m_shouldShowInvisible = adapter()->isNotationExisting() ? adapter()->scoreConfig().isShowInvisibleElements : true;
+    bool m_shouldShowUnprintable = adapter()->isNotationExisting() ? adapter()->scoreConfig().isShowUnprintableElements : true;
```
RomanPudashkin, line 55: Better to put this code to the constructor of this class and here you should use only default values:

bool m_shouldShowInvisible = false;
bool m_shouldShowUnprintable = false;
...

src/notation/internal/notationuiactions.cpp
```diff
@@ -1181,6 +1181,18 @@ void NotationUiActions::init()
             alist.push_back(a.code);
         }
         m_actionCheckedChanged.send(alist);
+
+        m_controller->currentNotationInteraction()->scoreConfigChanged().onReceive(this, [this](ScoreConfigType configType) {
+            std::unordered_map<ScoreConfigType, std::string> configActions = {
```
RomanPudashkin, line 1186: Better to mark this collection as static const:

static const std::unordered_map<ScoreConfigType, std::string> configActions = {
...

RomanPudashkin:
@vshalkevich please, check this

src/notation/internal/notationuiactions.cpp
aaaeka: It can't compile with ```static const``` so I left it with just ```static```.

vshalkevich:
@aaaeka 
Please, clarify what this new functions do:

updateShouldShowUnprintable
updateShouldShowFrames
updateShouldShowPageMargins
updateShouldShowInvisible

src/inspector/internal/compatibility/mu4inspectoradapter.h
```diff
@@ -60,15 +65,16 @@ class MU4InspectorAdapter : public IInspectorAdapter
...
-    void updatePageMarginsVisibility(const bool isVisible) override;
-    void updateFramesVisibility(const bool isVisible) override;
     void updateHorizontalGridSnapping(const bool isSnapped) override;
     void updateVerticalGridSnapping(const bool isSnapped) override;
-    void updateUnprintableElementsVisibility(const bool isVisible) override;
-    void updateInvisibleElementsDisplaying(const bool isVisible) override;
+    void updateInvisibleElementsDisplaying() override;
+    void updateUnprintableElementsVisibility() override;
+    void updateFramesVisibility() override;
+    void updatePageMarginsVisibility() override;
```
cbjeukendrup, line 73: How about renaming these to `toggle` instead of `update`?

src/notation/internal/notationuiactions.cpp
cbjeukendrup: If you use [`at()`](https://en.cppreference.com/w/cpp/container/unordered_map/at) instead of [`operator []`](https://en.cppreference.com/w/cpp/container/unordered_map/operator_at), you can make `configActions` const. 

Because [] performs an insertion if the key is not found, that one is not const; at() throws an exception when the key is not found. But as long as all ScoreConfigType enum values are in the map, that won't be a problem. 

@RomanPudashkin Do you think it would be good to use at()?

aaaeka:
> @aaaeka
> Please, clarify what this new functions do:
> 
> updateShouldShowUnprintable
> updateShouldShowFrames
> updateShouldShowPageMargins
> updateShouldShowInvisible

They change the private variables and send a change signal to the model.

src/inspector/internal/compatibility/mu4inspectoradapter.h
aaaeka: But they don't toggle anything, just send an action.

cbjeukendrup:
But I thought, the actions that they send are toggling something rather than updating something.

src/notation/internal/notationuiactions.cpp
aaaeka: Sure, I can change it to ```at()```.

aaaeka:
> But I thought, the actions that they send are toggling something rather than updating something.

Do you mean ```shouldShowInvisibleChanged``` in ```scoredisplaysettingsmodel.h``` or ```updateInvisibleElementsDisplaying``` in the inspector adapter?

cbjeukendrup:
I mean in the Adapter. (Sorry, I typed my comment in the wrong box)

aaaeka:
> I mean in the Adapter. (Sorry, I typed my comment in the wrong box)

I guess it makes sense since later on in ```notationactioncontroller``` it gets toggled, so ok I'll change it.

RomanPudashkin (approved):

vshalkevich:
Can`t check for now because Instruments panel is empty for me (Windows 10). 
It seems like Master branch issue.

shoogle (approved):
Works for me. To be clear, the "Page Margins" checkbox is not doing anything at the moment, but that appears to be the case even in `master` (via "View > Show Page Margins") so it's not the fault of this PR. The other checkboxes are all working as expected with this PR.

njvdberg:
When running a debug version, with this PR MU4 crashes on an assert whenever a score is closed (just click on the `x` on the tab)

The assert 
```
IF_ASSERT_FAILED(context() && context()->currentNotation()) {
```
in `mu::notation::ScoreConfig MU4InspectorAdapter::scoreConfig()` (mu4inspectoradapter.cpp) fails because `context()->currentNotation()` return a `nullptr`.
This is understandable because `scoreConfig()` is called **because** the current notation was set to `nullptr` (in `NotationSwitchListModel::closeNotation(int index)`).

It seems to me, lines
```
IF_ASSERT_FAILED(context() && context()->currentNotation()) {
```
Should be replaced by something like
```
Q_ASSERT(context())
if (context() && context()->currentNotation()) {
```

aaaeka:
> When running a debug version, with this PR MU4 crashes on an assert whenever a score is closed (just click on the `x` on the tab)
> 
> The assert
> 
> ```
> IF_ASSERT_FAILED(context() && context()->currentNotation()) {
> ```
> 
> in `mu::notation::ScoreConfig MU4InspectorAdapter::scoreConfig()` (mu4inspectoradapter.cpp) fails because `context()->currentNotation()` return a `nullptr`.
> This is understandable because `scoreConfig()` is called **because** the current notation was set to `nullptr` (in `NotationSwitchListModel::closeNotation(int index)`).
> 
> It seems to me, lines
> 
> ```
> IF_ASSERT_FAILED(context() && context()->currentNotation()) {
> ```
> 
> Should be replaced by something like
> 
> ```
> Q_ASSERT(context())
> if (context() && context()->currentNotation()) {
> ```

Should I open a seperate PR to fix this?

cbjeukendrup:
@aaaeka I believe you can't push to a PR that is already merged. But I'm currently preparing a PR about InspectorAdapter, so I'll do this fix in the same PR.

njvdberg:
@aaaeka Just my 2 cents, but whether it possible or not, once a PR is merged it is a bad idea to change afterwards. 

@cbjeukendrup Depends on how "complex" your PR is, stacking too many issues into one PR is also not such good idea (just 2 other cents :wink: ) My first guess would be to solve this in a separate PR.

Issue #7839, opened 2021-03-30
[MU4 Issue] set Body Text Size min to 8 and max to 24 


vshalkevich:
**Describe the bug**
There is no Min and Max Body Text Size set in Preferences. It may occur to unreadable text size

**To Reproduce**
1. Click on Edit from the upper Menu Bar
2. Click on 'Preferences
3. Click on Appearance from the left panel
4. Need to make manually changing Body Text Size field to min 8 and max 24

**Expected behavior**
set Body Text Size min to 8 and max to 24 

**Screenshots**
![2021-03-30 17_55_58-bin](https://user-images.githubusercontent.com/71372310/113015973-b435c500-9186-11eb-8120-c865efeb91b3.jpg)

**Desktop (please complete the following information):**
MacOS
Windows 10

**Additional context**

PR #7842, opened 2021-03-30
[MU4] Clamp body text size

aaaeka:
Resolves: #7839

### DIFF
--- a/src/appshell/qml/Preferences/internal/UiFontSection.qml
+++ b/src/appshell/qml/Preferences/internal/UiFontSection.qml
@@ -65,8 +65,8 @@ Column {
 
             implicitWidth: 112
 
-            minValue: 0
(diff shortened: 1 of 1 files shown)


aaaeka:
Perhaps the max text size should be 18? That's the last value where the text still fits in all the boxes.

cbjeukendrup (commented):
I think it would be good to additionally / instead set the minimum/maximum values in UiFontSection.qml:
Schermafbeelding 2021-03-30 om 22 27 01

If you do that, it should not be possible any more to set a "wrong" value, so that would make it not really necessary to put another guard in AppearancePreferencesModel.

cbjeukendrup (approved):

RomanPudashkin (approved):

PR #7843, opened 2021-03-30

cbjeukendrup:
Fix place of "About" actions in macOS menu bar, by giving them the correct `QAction::MenuRole`.


cbjeukendrup:
Maybe the "menuRoleForActionCode" function should be somewhere else, in some class. @Eism what would you advice?

Eism:
@cbjeukendrup @mattmcclinch thanks!
Guys, I like your solution!
@mattmcclinch, let's do the PR and @vshalkevich will check on different platforms

cbjeukendrup:
@mattmcclinch That was indeed the solution I was thinking of!
If you make a new PR, I will close this one.

PR #7855, opened 2021-03-31

mattmcclinch:
Fix place of "About" actions in macOS menu bar, by giving them the correct QAction::MenuRole. This replaces #7843.

Resolves: https://github.com/musescore/MuseScore/pull/7843#issuecomment-810742404.


shoogle:
@mattmcclinch, I tested this on macOS and found "About MuseScore" and "About Qt" were under the MuseScore menu as expected. but "About MusicXML" was still under the Help menu. Is this intentional?

mattmcclinch:
Yes, it is intentional. Otherwise, there would be two separators in the help menu with nothing in between, which looks really bad.

shoogle:
Maybe we should keep "About Qt" in the Help menu too, and just have "About MuseScore" in the MuseScore menu?

mattmcclinch:
>Maybe we should keep "About Qt" in the Help menu too, and just have "About MuseScore" in the MuseScore menu?

We could do that, simply by not specifying `MenuRole::AboutQtRole` when creating the `"about-qt"` menu item. I agree that it might look a little bit better that way. But I think we should have a _really_ good reason for not letting Qt place this menu item where it thinks it ought to go. The way I have it now is the way it was in MuseScore 3. @cbjeukendrup, do you have any strong feelings about the matter?

cbjeukendrup:
I don't have a strong opinion about this. Some thoughts:
- The "About MusicXML" is quite unrelated to the other ones, so I wouldn't mind if it's in a different place. 
- Personally, I'm not a great fan of the "About Qt" item, but it's mandatory to have it. I'm not sure if it is also mandatory to put it in the MuseScore menu, or if it would also be okay to hide it in the Help menu. 
- However, I think it would not make sense to put the "About Qt" item just after of before the "About MusicXML" item, because in a way, they are very different. 

So I think I would conclude to leave it as it is now (MuseScore and Qt in the MuseScore menu, MusicXML in the Help menu).

mattmcclinch:
The "About Qt" item is nice because it tells you which version of Qt was used to build the application. The "About MusicXML" item, on the other hand, seems to serve no purpose whatsoever.

shoogle:
> But I think we should have a really good reason for not letting Qt place this menu item where it thinks it ought to go.

Right. I didn't realise it was putting itself in that menu rather than something we had done ourselves.

igorkorsukov:
would you please do rebase?

DmitryArefiev:
Hi @mattmcclinch , can you do rebase please?

I logged #9073 which is probably fixed with your PR

RomanPudashkin:
Fixed in
https://github.com/musescore/MuseScore/pull/9410

PR #7860, opened 2021-03-31
Fix menus on Windows and Linux

mattmcclinch:
#7837 fixed submenus on macOS, but it broke menus on Windows and Linux. This makes sure to only call `QMenu::setParent()` on macOS, since it is problematic to do so otherwise. This also provides a better solution to the disappearing menu problem on macOS that was addressed in #7782.


shoogle:
Tested on macOS and the Add menu works as it should. @vshalkevich, could you test on Windows?

vshalkevich:
Tested on Windows 10 and it seems fine.

shoogle (approved):
Works on Linux too.

PR #7861, opened 2021-03-31
[MU4] Preferences. MIDI device mapping

RomanPudashkin:
midi-device-mappings


src/framework/ui/internal/uiconfiguration.cpp
```diff
@@ -266,15 +266,44 @@ void UiConfiguration::writeThemes(const ThemeList& themes)
...
     return db.families();
 }
 
-ThemeList UiConfiguration::themes() const
-{
-    return m_themes;
+QStringList UiConfiguration::possibleAccentColors() const
+{
+    static const QStringList lightAccentColors {
+        "#F36565",
```
cbjeukendrup, line 283: Couldn't these be part of the `ThemeInfo` struct? (either directly in the struct or in the `QMap<ThemeStyleKey, QVariant> values`)

igorkorsukov:
@Tantacrul  by the way, interesting, why this names "MIDI Device Mapping", but not, for example, "MIDI Control" or "Control by MIDI"

Tantacrul:
> @Tantacrul by the way, interesting, why this names "MIDI Device Mapping", but not, for example, "MIDI Control" or "Control by MIDI"

It could be called 'MIDI Control'. In fact that could be a better name.

MIDI Device Mapping is more precise... but I'm also worried it might sound a bit more advanced than it really is.

Tantacrul:
BTW: It would be 'Mapping', not 'Mappings'

src/framework/ui/internal/uiconfiguration.cpp
RomanPudashkin: I think ThemeInfo should contain only values that affect the UI. Those values are used only in the accent colors list and they don't affect to color of any UI element, so better don't store them in the theme

Eism (approved):

Eism (approved):

PR #5802, opened 2020-03-11
[MU4] fix #302104: slur and grace note (acciaccatura) fails

mattmcclinch:
Resolves: https://musescore.org/en/node/302104.

This commit does the following:
- Allows a slur to be added from a grace note to its main note if a range selection is used.
- Changes the behavior of applying the Slur palette element to a list selection of more than two notes so that it is consistent with how the "S" keyboard shortcut already behaves.
- Does not start edit mode for a newly created slur if the selection contains more than one element.
- Fixes a mathematical error in ChordRest::isBefore() that can cause a slur to be added with an incorrect span if grace notes are involved.

PR #5777, opened 2020-03-02
[MU4] fix #301768: Unexpected tieing behaviour

mattmcclinch:
Resolves: https://musescore.org/en/node/301768.

The `cmdAddTie()` function currently skips over notes that already have an existing tie. This is fine when handling the "chord-tie" command, for which `addToChord` is `true`. But when `addToChord` is `false`, as is the case for the regular "tie" command, what we really want to do is remove the existing tie, and then treat the note as if it never had a tie in the first place.

PR #7871, opened 2021-04-02
[MU3] Backporting some master PRs to 3.x

Jojo-Schmitz:
Backporting @mattmcclinch's PRs for master to 3.x:

* [Fix #307704](https://musescore.org/en/node/307704): Capo settings applied via staff text do not affect chord symbols _(#7700, merged)_
* [Fix #302104](https://musescore.org/en/node/302104): Slur and grace note (acciaccatura) fails _(#5802, this one actually needs to get "forward ported" to master, at least rebased)_
* [Fix #301768](https://musescore.org/en/node/301768): Unexpected tieing behaviour _(#5777 this one actually needs to get forward "ported to master", at least rebased)_
* [Fix #301016](https://musescore.org/en/node/301016): Copying a partial measure leaves rest(s) missing in voices 2-4. _(#5702, this one actually needs to get "forward ported" to master, at least rebased)_
* [Fix #318772](https://musescore.org/en/node/318772): Don't save unused images in MSCZ file _(not quite the same, but similar to #7740, merged)_

### DIFF
--- a/libmscore/check.cpp
+++ b/libmscore/check.cpp
@@ -319,7 +319,7 @@ bool Score::checkClefs()
 //   fillGap
 //---------------------------------------------------------
 
-void Measure::fillGap(const Fraction& pos, const Fraction& len, int track, const Fraction& stretch)
+void Measure::fillGap(const Fraction& pos, const Fraction& len, int track, const Fraction& stretch, bool useGapRests)
       {
       qDebug("measure %6d pos %d, len %d/%d, stretch %d/%d track %d",
          tick().ticks(),
@@ -334,7 +334,7 @@ void Measure::fillGap(const Fraction& pos, const Fraction& len, int track, const
             rest->setTicks(len);
             rest->setDurationType(d);
             rest->setTrack(track);
-            rest->setGap(true);
+            rest->setGap(useGapRests);
             score()->undoAddCR(rest, this, (pos / stretch) + tick());
             }
       }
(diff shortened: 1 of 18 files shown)


Jojo-Schmitz:
Not sure what the mtest failurs is trying to tell me?

mattmcclinch:
`mtest/libmscore/copypaste/copypaste_tremolo-ref.mscx` was added after #5702 was submitted, so it needs to be updated with the new expected result.

Jojo-Schmitz:
OK, thanks, will do. The rest looks OK to you?

worldwideweary:
There's a problem with 3.x ("3.7") and the https://github.com/musescore/MuseScore/pull/7740 PR:

Create a score with an image attached. Save. 
Close score, reload, then do a save (Ctrl+S) - the resulting file will have lost the image. 

Reverting Fix #318772 (not sure if with MU4 how this is) fixes the problem.

For now, possibly having extra images outweighs the total problem of losing images. Should probably double check MU4 to see if it's okay there, but that's not my concern for now.

Jojo-Schmitz:
I can confirm the issue with 3.x. I've tested just with 4.0.2: it seems not affected. Now need to understand why...

Jojo-Schmitz:
I believe the fix is to set `_used` to `true` in both constructors, Check https://github.com/Jojo-Schmitz/MuseScore/pull/96

PR #7891, opened 2021-04-04
[MU3] Fix #319584: Many percussions elements/sounds are not getting translated in status bar and Edit Drumset dialog

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/319584

We'd need to mix and match https://github.com/musescore/MuseScore/blob/6fe71f9ed129e502b0ab4030d4672833efbbd6de/src/libmscore/drumset.cpp#L256-L312 and the more complete https://github.com/musescore/MuseScore/blob/6fe71f9ed129e502b0ab4030d4672833efbbd6de/src/importexport/guitarpro/internal/importgtp.cpp#L303-L373
with instruments.xml

* Fix missing translations for Hi Congas
* Fix missing translation for Hi Bongo
* Fix missing translation for Hi Woodblock
* Fix missing translation for Vibraslap
* Fix wrong translation/name for Sleigh Bell
* Fix missing translations for Snare drums
* Fix missing translations for Bell Tree by changing it to the apparently more correct "Mark Tree", requires updated translations though
* Fix missing translations for Toms
* Fix missing/wrong translation/name and sound for Slap
* Fix/add wrong translation/missing name and sound for Castanets/Hand Clap
* Fix missing translations for Bass Drums
* Fix missing translations for Ride Cymbal

To do: Look after the marching percussion stuff, it seems to be using an entirely different number-to-name mapping? Here we may need to add strings (and subsequently translate them), and by other means than done in ths PR so far. I'd rather not though this here and now...

For master see #7893


vpereverzev:
merged in master, 3.x merges are closed

PR #7897, opened 2021-04-05
[MU4] Autobot configure options

igorkorsukov:


src/autobot/internal/autobotconfiguration.cpp
```diff
@@ -27,14 +27,19 @@ mu::io::path AutobotConfiguration::dataPath() const
...
 mu::io::path AutobotConfiguration::drawDataPath() const
 {
     return dataPath() + "/draw_data";
 }
 
-mu::io::path AutobotConfiguration::scoreDrawData(const io::path& scorePath) const
+mu::io::path AutobotConfiguration::fileDrawDataPath(const io::path& filePath) const
 {
-    return drawDataPath() + "/" + io::basename(scorePath) + ".json";
+    return drawDataPath() + "/" + io::basename(filePath) + ".json";
```
vpereverzev, line 42: is it also working for windows with their "\" separators?

src/autobot/internal/autobotconfiguration.h
```diff
@@ -28,9 +28,11 @@ class AutobotConfiguration : public IAutobotConfiguration
     AutobotConfiguration() = default;
 
     io::path dataPath() const override;
-    io::path drawDataPath() const override;
 
-    io::path scoreDrawData(const io::path& scorePath) const override;
+    io::path filesPath() const override;
+
+    io::path drawDataPath() const override;
+    io::path fileDrawDataPath(const io::path& filePath) const override;
```
vpereverzev, line 35: the name is really confusing

Probably you'd prefer something like (if I understood you correctly):
`io::path drawDataDir() const
io::path drawDataFilePath(const io::path& fileName) const`

src/autobot/internal/autobotconfiguration.cpp
igorkorsukov: yes it is provided by the filesystem implementation

src/autobot/internal/autobotconfiguration.h
igorkorsukov: The purpose of this method is to create a file path with the drawing data for a given score file.

RomanPudashkin (approved):

Issue #7763, opened 2021-03-19
[MU4 Issue] [enhancement] The width of The grabber on the very left hand side on Note Input Bar should be 32px


vshalkevich:
**Describe the bug**
The width of The grabber on the very left hand side on Note Input Bar is 35px
The width of The grabber on the very left hand side on Note Input Bar should be 32px

**To Reproduce**
1. Create a score
2. Look at Note Input Bar

**Expected behavior**
The width of The grabber on the very left hand side on Note Input Bar should be 32px

**Screenshots**
![image (1)](https://user-images.githubusercontent.com/71372310/111849282-4388e000-891e-11eb-91f4-efebc31fe594.png)

**Desktop (please complete the following information):**
MacOS

cbjeukendrup:
I believe that technically the width of the grabber is okay, but there is an erroneous 3px padding left to the buttons, produced by the GridView containing them.

Issue #7762, opened 2021-03-19
[MU4 Issue] [enhancement] The separator line on Note Input Bar, the spacing both sides should be 6px


vshalkevich:
**Describe the bug**
The separator line on Note Input Bar the spacing both sides is 2 px
The separator line on Note Input Bar, the spacing both sides should be 6px

**To Reproduce**
1. Create a score
2. Look at separator line on Note Input Bar

**Expected behavior**
The separator line on Note Input Bar, the spacing both sides should be 6px

**Screenshots**
![image](https://user-images.githubusercontent.com/71372310/111849053-c65d6b00-891d-11eb-9d92-b469022d6ab4.png)

**Desktop (please complete the following information):**
MacOS
Windows 10

PR #7899, opened 2021-04-05
Fix spacing of items in NoteInputBar

cbjeukendrup:
Resolves: #7762 
Resolves: #7763

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml
@@ -7,6 +7,7 @@ Item {
 
     property var model: null
     property int orientation: Qt.Horizontal
+    readonly property bool isHorizontal: orientation === Qt.Horizontal
 
     property Component sectionDelegate: Item {}
     property Component itemDelegate: Item {}
@@ -27,6 +28,9 @@ Item {
     QtObject {
         id: privateProperties
 
+        property int spacingBeforeSection: isHorizontal ? columnSpacing : rowSpacing
+        property int spacingAfterSection: spacingBeforeSection
+
         function modelSections() {
             var _sections = []
 
@@ -45,22 +49,25 @@ Item {
 
     Loader {
         anchors.fill: parent
-        sourceComponent: orientation === Qt.Horizontal ? horizontalView : verticalView
+        sourceComponent: isHorizontal ? horizontalView : verticalView
     }
 
     Component {
         id: horizontalView
 
         Row {
-            spacing: 0
+            spacing: privateProperties.spacingBeforeSection
 
             Repeater {
                 model: Boolean(root.model) ? privateProperties.modelSections() : []
 
                 Row {
-                    spacing: 0
+                    spacing: privateProperties.spacingAfterSection
+                    height: parent.height
 
                     GridViewSection {
+                        anchors.verticalCenter: parent.verticalCenter
+
                         width: root.sectionWidth
                         height: root.sectionHeight
 
@@ -78,9 +85,6 @@ Item {
                         cellWidth: root.cellWidth
                         cellHeight: root.cellHeight
 
-                        sectionWidth: root.sectionWidth
-                        sectionHeight: root.sectionHeight
-
                         rows: root.rows
                         rowSpacing: root.rowSpacing
                         columns: root.columns
@@ -95,17 +99,17 @@ Item {
         id: verticalView
 
         Column {
-            spacing: 0
+            spacing: privateProperties.spacingBeforeSection
 
             Repeater {
                 model: Boolean(root.model) ? privateProperties.modelSections() : []
 
                 Column {
-                    spacing: 0
+                    spacing: privateProperties.spacingAfterSection
+                    width: parent.width
 
                     GridViewSection {
-                        anchors.left: parent.left
-                        anchors.leftMargin: root.columnSpacing
+                        anchors.horizontalCenter: parent.horizontalCenter
 
                         width: root.sectionWidth
                         height: root.sectionHeight
@@ -123,9 +127,6 @@ Item {
 
                         cellWidth: root.cellWidth
                         cellHeight: root.cellHeight
-
-                        sectionWidth: root.sectionWidth
-                        sectionHeight: root.sectionHeight
 
                         rows: root.rows
                         rowSpacing: root.rowSpacing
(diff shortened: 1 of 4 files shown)


src/framework/uicomponents/qml/MuseScore/UiComponents/internal/GridViewDelegate.qml
```diff
@@ -41,14 +41,17 @@ Item {
     GridView {
         id: gridView
 
-        property int columns: root.columns !== -1 ? root.columns : gridView.count
+        property int columns: root.columns !== -1 ? root.columns : Math.ceil(gridView.count / gridView.rows)
```
Eism, line 44: I think you have a binding loop warning here because columns depend on rows and rows are also column dependent

src/framework/uicomponents/qml/MuseScore/UiComponents/internal/GridViewDelegate.qml
```diff
@@ -41,14 +41,17 @@ Item {
...
         property int rows: root.rows !== -1 ? root.rows : Math.ceil(gridView.count / gridView.columns)
 
+        x: -1/2 * root.columnSpacing
```
Eism, line 47: I think you can replace this on anchors with margins

src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml
```diff
@@ -24,6 +25,9 @@ Item {
     property int columns: noLimit
     property int columnSpacing: 2
 
+    property int spacingBeforeSection: isHorizontal ? columnSpacing : rowSpacing
```
Eism, line 28: better move it to private properties

src/framework/uicomponents/qml/MuseScore/UiComponents/internal/GridViewDelegate.qml
cbjeukendrup: Ah, good point, that will happen if root.rows and root.columns are both -1. However, isn't that a bad situation anyway?

Eism: What problem did you solve here? Because the old solution seems logical to me

Eism: if both are -1 then 
columns = gridView.count 
rows = Math.ceil(gridView.count / gridView.count) = 1
in the old version

cbjeukendrup: But what if rows = 2 (given) and columns = -1?
In the old version, `columns` would then become just `gridView.count`, but it should be `Math.ceil(gridView.count / gridView.rows)`. (Otherwise, there would be 2x too much cells).
This case is not used anywhere in MuseScore yet, but it seems better to make that work too, in case we need it in the future.

Or am I now mistaken? 🤯

I think I should add a special case for when rows and columns are both -1, so that rows defaults to +1 in that case.

src/framework/uicomponents/qml/MuseScore/UiComponents/internal/GridViewDelegate.qml
cbjeukendrup: Currently that is not possible, because `root.width` depends on `gridview.width`. 
We could do the following: 
It is more verbose, but maybe also "feels" more correct. Do you think it's worth it?
```
Item {
    id: root
    ...
    width: gridView.columns * gridView.cellWidth - root.columnSpacing
    height: gridView.rows * gridView.cellHeight - root.rowSpacing

    GridView {
        id: gridView
        ...
        anchors.fill: parent
        anchors.topMargin: -root.rowSpacing / 2
        anchors.bottomMargin: -root.rowSpacing / 2
        anchors.leftMargin: -root.columnSpacing / 2
        anchors.rightMargin: -root.columnSpacing / 2

        cellWidth: ...
    }
}

File: src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml (hunk 3 above)
Comment (cbjeukendrup) [reply src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml:28]: I put it here deliberately, thinking we might want to override it at some point. Shall I move it to private for now? If we somewhen need it, we can move it back.

**Review by Eism — APPROVED** [review]

File: src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml (hunk 3 above)
Comment (Eism) [reply src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml]: The second sentence is correct :)

**Comment by vshalkevich** [comment]
Need to get the latest fixes for Windows and Linux platforms. 
This PR still unable to launch for me

**Comment by vshalkevich** [comment]
Need to get the latest fixes for Windows and Linux platforms into this PR.
This PR still unable to launch for me at this moment

Please, rebase to the current master

**Comment by cbjeukendrup** [comment]
@vshalkevich Done!

**Comment by vshalkevich** [comment]
Works fine for MacOS

**Comment by vshalkevich** [comment]
Seems fine for me

File: src/framework/uicomponents/qml/MuseScore/UiComponents/internal/GridViewSection.qml [hunk 4]
```diff
@@ -1,20 +1,20 @@
...
 
-        onStatusChanged: {
-            if (status === Loader.Ready) {
-                itemModel = Qt.binding( function() { return Boolean(modelData) ? modelData : null });
-                itemIndex = index
-            }
+        onLoaded: {
+            itemModel = Qt.binding( function() { return Boolean(modelData) ? modelData : null });
+            itemIndex = index
+            root.visible = Qt.binding( function() { return Boolean(item) ? item.visible : false })
```
Eism, line 17: @cbjeukendrup 
Can you explain why you added this code?

Sections are not displayed on the master because of this code

cbjeukendrup: Without this code, a GridViewSection was visible even when it's content (the separator in the NoteInputBar) is not visible, which causes that an empty space is shown, and an extra `spacing` is added by the Column or Row containing the GridViewSection (see GridViewSectional.qml, line 65).

Also, I'm not sure whether this code is the cause that sections are not shown, because in the sample grid view on the DevTools page, they are shown correctly. 
I think the cause is that the `state` in NoteInputBar.qml is not correctly set initially; because when you trigger a state change by dragging the toolbar to a location where it is vertical, the separators are shown correctly. I also believe they are still shown correctly with the old docking system, but not in the new one. So I think initially a signal about `orientationChanged` is somehow not being emitted.

PR #7900, opened 2021-04-05
Fix Edwin on macOS and add other missing fonts to install

cbjeukendrup:
The following fonts were in the fonts folder, but were not installed on macOS:
- `${PROJECT_SOURCE_DIR}/fonts/bravura/Bravura.otf`
- `${PROJECT_SOURCE_DIR}/fonts/edwin/Edwin-*`
- `${PROJECT_SOURCE_DIR}/fonts/gootville/Gootville.otf`
- `${PROJECT_SOURCE_DIR}/fonts/leland/LelandText.otf`
- `${PROJECT_SOURCE_DIR}/fonts/musejazz/MuseJazz.otf`
- `${PROJECT_SOURCE_DIR}/fonts/petaluma/Petaluma.otf`

For this reason, Edwin didn't work on Mac. Now, all fonts seem to work correctly in MU4. 

About the other fonts that were not installed: this has also been the case with MuseScore 3, and I have not really noticed problems with it. Does somebody know whether they were left out on purpose??

Anyway, it seems strange to me that for Leland the "Text" variant was left out, and for the other fonts the "normal" version. This might be the cause of some issues.

### DIFF
--- a/src/main/CMakeLists.txt
+++ b/src/main/CMakeLists.txt
@@ -456,24 +456,29 @@ elseif(OS_IS_MAC)
     install(FILES "${PROJECT_SOURCE_DIR}/src/main/res/musescoreDocument.icns" DESTINATION ${Mscore_SHARE_NAME}${Mscore_INSTALL_NAME})
 
     install(FILES
-            ${PROJECT_SOURCE_DIR}/fonts/gootville/GootvilleText.otf
-            ${PROJECT_SOURCE_DIR}/fonts/mscore/MScoreText.ttf
-            ${PROJECT_SOURCE_DIR}/fonts/mscore/MusescoreIcon.ttf
-            ${PROJECT_SOURCE_DIR}/fonts/musejazz/MuseJazzText.otf
+            ${PROJECT_SOURCE_DIR}/fonts/bravura/BravuraText.otf
             ${PROJECT_SOURCE_DIR}/fonts/campania/Campania.otf
+            ${PROJECT_SOURCE_DIR}/fonts/edwin/Edwin-BdIta.otf
+            ${PROJECT_SOURCE_DIR}/fonts/edwin/Edwin-Bold.otf
+            ${PROJECT_SOURCE_DIR}/fonts/edwin/Edwin-Italic.otf
+            ${PROJECT_SOURCE_DIR}/fonts/edwin/Edwin-Roman.otf
+            ${PROJECT_SOURCE_DIR}/fonts/firasans/FiraSans-Regular.ttf
+            ${PROJECT_SOURCE_DIR}/fonts/firasans/FiraSans-SemiBold.ttf
+            ${PROJECT_SOURCE_DIR}/fonts/gootville/GootvilleText.otf
+            ${PROJECT_SOURCE_DIR}/fonts/FreeSans.ttf
             ${PROJECT_SOURCE_DIR}/fonts/FreeSerif.ttf
             ${PROJECT_SOURCE_DIR}/fonts/FreeSerifBold.ttf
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Might that be somehow related to the 3.x issue https://musescore.org/en/node/316152

Jojo-Schmitz:
I don't think the Musical Fonts need to get installed, Just bthe Text fonts and maybe the Musical Text Fonts too?
But if  Musical Fonts need to get installed, then mscore.ttf too.

cbjeukendrup:
> But if Musical Fonts need to get installed, then mscore.ttf (AKA Emmentaler) too.

Good catch, thanks. Think I'll wait if somebody knows whether the Musical Text fonts should be installed, and then I will update the PR.

Jojo-Schmitz:
I think they don't. They were not in 3.x either

cbjeukendrup:
Things are getting a bit clearer to me. 

On other OSs, the fonts are just in the qrc file and get loaded in fontssetup.cpp, `int loadStatusCode = QFontDatabase::addApplicationFont(font);`. 

But on macOS, the fonts from the qrc are _not_ registered using QFontDatabase, but they _are_ used within MuseScore, being loaded from the qrc, in sym.cpp. 
So even when no musical fonts are installed, the musical fonts will work because they are in the qrc so they are compiled into the executable. (I tested this by deleting the fonts from the install directory of MuseScore 3.)

Edwin however, is apparently not taken from the qrc but from the installed fonts, because when it is deleted from the install directory, it doesn't work (macOS takes some awful fallback font). 

It's still quite unclear how the Musical Text Fonts work.

It appears that in MU3, both Leland and Leland Text were installed, which makes the printer issue even stranger, I think.

Jojo-Schmitz:
Indeed Leland should not neet to get installed in 3.x either. Maybe it is the cause for https://musescore.org/en/node/316152, maybe it doesn't harm to install it, mbut it should not be needed.

Jojo-Schmitz:
Where exactly is Leland.otf getting installed to on a Mac, what's the exact file location?

Jojo-Schmitz (approved):

cbjeukendrup:
Apparently Leland (not Leland Text) is also used in the UI, and then it is not taken from the qrc, so we have to install that one too.

(In MU4, there is no separate `fonts-Mac` qrc, so all fonts will (also) be built in, but I think that's no problem.)

Jojo-Schmitz:
Where exactly is it getting used in the UI?

cbjeukendrup:
Al least in the New Score dialog and in the play panel for the tempo

Jojo-Schmitz:
Play panel for tempo should be using a Musical Text font, I think.

You meant this here?
image

The palettes use(d to use) Bravura, are they using Leland now?

Jojo-Schmitz:
Which play panel BTW, the thing on the toolbar, showing note=value? That on doesn't work at all for 8th notes, shows 16th instead.

cbjeukendrup:
Those key signatures are "icons" from the MuseScore Icon Font, but the time signature and the quarter note for the tempo (also used in the play panel) are from Leland. (Yes, with play panel I meant playback toolbar.)

Good point about the Palettes, I haven't seen those for a while, but after some factory resets I got them back again.
It seems they use Leland, but from the qrc. Which seems logical to me because they work with that `gscore`, which is a kind of normal score and uses the ScoreFonts from libmscore, which are loaded from qrc.

Jojo-Schmitz:
Those should be using a scoreview, and as such render just like normal scores, don't they?

cbjeukendrup:
Not really a score view, but indeed a normal `Score`.

RomanPudashkin (approved):

PR #7902, opened 2021-04-06
[MU3] On Mac don't 'install' Leland and don't 'build in' (musical) text fonts

Jojo-Schmitz:
For Leland rather the build in version should get used and for Leland Text and the Edwin fonts the installed version should get used.

See also #7900 (for master, merged already)

May resolve (testing/testers needed): https://musescore.org/en/node/316152  (actually that is being fxed by #8578, as confirmed by several users)


Jojo-Schmitz:
See #9000

PR #7906, opened 2021-04-06
[MU4] Concept: keyboard navigation by subsections

igorkorsukov:
including https://github.com/musescore/MuseScore/pull/7904

https://user-images.githubusercontent.com/3818029/113729132-0468d600-96f7-11eb-944e-8c77af54323c.mp4

### DIFF
--- a/src/framework/ui/ikeynavigationsection.h
+++ b/src/framework/ui/ikeynavigationsection.h
@@ -20,8 +20,21 @@
 #define MU_UI_IKEYNAVIGATIONSECTION_H
 
 #include <QString>
+#include <QList>
 
 namespace mu::ui {
+class IKeyNavigationSubSection
+{
+public:
+    virtual ~IKeyNavigationSubSection() = default;
+
+    virtual QString name() const = 0;
+    virtual int order() const = 0;
+    virtual bool enabled() const = 0;
+    virtual bool active() const = 0;
+    virtual void setActive(bool arg) = 0;
+};
+
 class IKeyNavigationSection
 {
 public:
@@ -32,6 +45,8 @@ public:
     virtual bool enabled() const = 0;
     virtual bool active() const = 0;
     virtual void setActive(bool arg) = 0;
+
+    virtual const QList<IKeyNavigationSubSection*>& subsections() const = 0;
 };
 }
 
--- a/src/framework/ui/ikeynavigationsection.h
+++ b/src/framework/ui/ikeynavigationsection.h
@@ -1,0 +1,38 @@
+//=============================================================================
+//  MuseScore
+//  Music Composition & Notation
+//
+//  Copyright (C) 2021 MuseScore BVBA and others
+//
+//  This program is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License version 2.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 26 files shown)


src/framework/ui/ikeynavigationsection.h
```diff
@@ -0,0 +1,53 @@
...
+//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+//=============================================================================
+#ifndef MU_UI_IKEYNAVIGATIONSECTION_H
+#define MU_UI_IKEYNAVIGATIONSECTION_H
+
+#include <QString>
+#include <QList>
+
+namespace mu::ui {
+class IKeyNavigationSubSection
```
RomanPudashkin, line 26: Do we really need this interface? It seems that we only need IKeyNavigationSection. Technically a situation may happen when we need sub-sub sections and I afraid that we'll create IKeyNavigationSubSubSection :)

igorkorsukov: I thought :) And I decided to add a separate interface, for two purposes:
1. Avoid `section-> section-> section...` tree. We only have a `section-> subsection-> control` and that's it.
2. For cleaning type

src/framework/ui/uitypes.h
```diff
@@ -162,14 +162,16 @@ struct UiAction
     std::string shortcut;
 
     UiAction() = default;
-    UiAction(const actions::ActionCode& code, UiContext ctx, const QString& title, Checkable ch = Checkable::No)
+    UiAction(const actions::ActionCode& code, UiContext ctx, Checkable ch = Checkable::No)
+        : code(code), context(ctx), checkable(ch) {}
+    UiAction(const actions::ActionCode& code, UiContext ctx, const char* title, Checkable ch = Checkable::No)
         : code(code), context(ctx), title(title), checkable(ch) {}
-    UiAction(const actions::ActionCode& code, UiContext ctx, const QString& title, const QString& desc, Checkable ch = Checkable::No)
+    UiAction(const actions::ActionCode& code, UiContext ctx, const char* title, const char* desc, Checkable ch = Checkable::No)
```
RomanPudashkin, line 169: Better to use std::string_view instead of const char*

igorkorsukov: no, not in this case :)
typically we do
```
const UiActionList ApplicationUiActions::m_actions = {
    UiAction("quit",
             mu::context::UiCtxAny,
             QT_TRANSLATE_NOOP("action", "Quit")
             ),
    UiAction("fullscreen",
             mu::context::UiCtxAny,
             QT_TRANSLATE_NOOP("action", "Full Screen"),
             QT_TRANSLATE_NOOP("action", "Full screen"),
             Checkable::Yes
             ),
```
here `QT_TRANSLATE_NOOP("action", "Esc")` just `const char*`

src/framework/ui/ikeynavigationsection.h
RomanPudashkin: Make sense, but I don't like that we have two almost identical interfaces. Maybe is it better to create one base interface and inherit IKeyNavigationSection from it?

igorkorsukov: I think this is overengineering :) [KISS](https://en.wikipedia.org/wiki/KISS_principle)

igorkorsukov: In the next PR I going to add common interface

RomanPudashkin (approved):

Issue #7693, opened 2021-03-06
[MU4 Issue] Able to click and hold on number spinners but value changes too slowly


vshalkevich:
**Describe the bug**
Able to click and hold on number spinners but value changes too slowly
Speed could accelerate the longer you hold (5 seconds to do entire range)
If no acceleration then it should take longer

**To Reproduce**
1. Start New Score Wizard
2. Click "Next"
3. Click "tempo"
4. Tick checkbox "show tempo marking on my score"
5. click and hold on tempo number spinner

**Expected behavior**

**Desktop (please complete the following information):**
MacOS
Windows

Mystic-Slice:
I am working on it

PR #7914, opened 2021-04-08
Increased the speed of value change for press and hold

Mystic-Slice:
Resolves: https://github.com/musescore/MuseScore/issues/7693

The number changes faster now (currently set to 10 counts per second). The speed can be controlled by changing the value of timerTimeInterval. No acceleration of change in value for now. The constant speed set now must be comfortable.

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/ValueAdjustControl.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/ValueAdjustControl.qml
@@ -9,6 +9,8 @@ Item {
     signal increaseButtonClicked
     signal decreaseButtonClicked
 
+    property int timerTimeInterval: 300
+
     height: childrenRect.height
     width: childrenRect.width
 
@@ -50,7 +52,7 @@ Item {
                 Timer {
                     id: continiousIncreaseTimer
 
-                    interval: 300
+                    interval: timerTimeInterval
 
                     repeat: true
 
@@ -89,7 +91,7 @@ Item {
                 Timer {
                     id: continiousDecreaseTimer
 
-                    interval: 300
+                    interval: timerTimeInterval
                     repeat: true
 
                     running: decreaseMouseArea.pressed
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/ValueAdjustControl.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/ValueAdjustControl.qml
@@ -9,8 +9,6 @@ Item {
     signal increaseButtonClicked
     signal decreaseButtonClicked
 
-    property int timerTimeInterval: 300
-
     height: childrenRect.height
     width: childrenRect.width
 
@@ -52,7 +50,7 @@ Item {
                 Timer {
                     id: continuousIncreaseTimer
 
-                    interval: timerTimeInterval
+                    interval: 100
 
                     repeat: true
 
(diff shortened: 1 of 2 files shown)


cbjeukendrup:
Very nice! Another issue with this is, that it takes quite long before the number _starts_ changing. This can be adjusted using the [`pressAndHoldDuration`](https://doc.qt.io/qt-5/qml-qtquick-mousearea.html#pressAndHoldInterval-prop) property of the MouseArea.

src/framework/uicomponents/qml/MuseScore/UiComponents/ValueAdjustControl.qml
```diff
@@ -9,6 +9,8 @@ Item {
     signal increaseButtonClicked
     signal decreaseButtonClicked
 
+    property int timerTimeInterval: 300
```
cbjeukendrup, line 12: Because ValueAdjustControl is used in only one place, I think it is better to make this directly the correct value (100), instead of setting it to a "wrong" default value that is never used and setting it separately in IncrementalPropertyControl.

Mystic-Slice:
I will make the changes

src/framework/uicomponents/qml/MuseScore/UiComponents/ValueAdjustControl.qml
Mystic-Slice: Yes, this makes more sense.

Mystic-Slice:
> Very nice! Another issue with this is, that it takes quite long before the number _starts_ changing. This can be adjusted using the [`pressAndHoldInterval`](https://doc.qt.io/qt-5/qml-qtquick-mousearea.html#pressAndHoldInterval-prop) property of the MouseArea.

I am unable to add 'pressAndHoldInterval' property, I get a warning and the app won't start. But the property is valid and I was able to find it even in the documentation. I don't know why it shows an error.

cbjeukendrup:
I found out that it has to do with the QtQuick version: you need to change the first line of the file to `import QtQuick 2.15`.

By the way, I saw that there is some discrepancy between the two mouse areas. Now that you're working in this file, could you please solve those?

Mystic-Slice:
> I found out that it has to do with the QtQuick version: you need to change the first line of the file to `import QtQuick 2.15`.

Oh ok I will change the version and try it.

Mystic-Slice:
> By the way, I saw that there is some discrepancy between the two mouse areas and timers. Now that you're working in this file, could you please solve those?

Can you please explain the issue you found a little more? I don't understand.

cbjeukendrup:
- In the `increaseMouseArea`, there is a line `preventStealing: true` that is missing in the `decreaseMouseArea`
- In the `continuousDecreaseTimer`, the line `running: decreaseMouseArea.pressed` should be removed

And there are some differences with whitespace; just nice if that's consistent.

Mystic-Slice:
> * In the `increaseMouseArea`, there is a line `preventStealing: true` that is missing in the `decreaseMouseArea`
> * In the `continuousDecreaseTimer`, the line `running: decreaseMouseArea.pressed` should be removed
> 
> And there are some differences with whitespace; just nice if that's consistent.

Ohh okok. Sure, I will make the changes.

Mystic-Slice:
@cbjeukendrup I have made the changes and thnkx for your help :)

cbjeukendrup (approved):

vshalkevich:
Seems fine, thanks

vpereverzev (approved):

vpereverzev:
Looks OK. There are only a few notes there - I suggest to add a few properties and use them along the component e.g. 
`QtObject {
    id: privateProperties

    property int pressDelay: 200
...`

OFC it's not preventing me to merge this PR, so you can polish it in the another one using information above

PR #7921, opened 2021-04-09
[MU4] Added key navigation for instruments panel

igorkorsukov:
not completed, without popups

https://user-images.githubusercontent.com/3818029/114197219-683b0b00-9952-11eb-93b2-c453c853eb30.mp4

### DIFF
--- a/src/framework/ui/ikeynavigation.h
+++ b/src/framework/ui/ikeynavigation.h
@@ -38,8 +38,17 @@ class IKeyNavigation
 public:
     virtual ~IKeyNavigation() = default;
 
+    struct Index
+    {
+        int column = -1;
+        int row = -1;
+
+        void setOrder(int n) { column = n; }
+        int order() const { return column; }
+    };
+
     virtual QString name() const = 0;
-    virtual int order() const = 0;
+    virtual const Index& index() const = 0;
     virtual bool enabled() const = 0;
     virtual bool active() const = 0;
     virtual void setActive(bool arg) = 0;
@@ -60,6 +69,14 @@ class IKeyNavigationSubSection : public IKeyNavigation
 public:
     virtual ~IKeyNavigationSubSection() = default;
 
+    //! NOTE Please sync with view/KeyNavigationSubSection::Direction
+    enum class Direction {
+        Horizontal = 0,
+        Vertical,
+        Both
+    };
+
+    virtual Direction direction() const = 0;
     virtual const QSet<IKeyNavigationControl*>& controls() const = 0;
     virtual async::Channel<SubSectionControl> forceActiveRequested() const = 0;
 };
--- a/src/framework/ui/view/keynavigationsubsection.h
+++ b/src/framework/ui/view/keynavigationsubsection.h
@@ -32,18 +32,30 @@ class KeyNavigationControl;
 class KeyNavigationSubSection : public AbstractKeyNavigation, public IKeyNavigationSubSection, public async::Asyncable
 {
     Q_OBJECT
-    Q_PROPERTY(KeyNavigationSection * section READ section WRITE setSection)
+    Q_PROPERTY(KeyNavigationSection * section READ section WRITE setSection NOTIFY sectionChanged)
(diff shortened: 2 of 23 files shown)


RobFog:
Typo: "instrumens" should be "instruments

igorkorsukov:
> Typo: "instrumens" should be "instruments

Thanks, fixed

src/framework/ui/ikeynavigation.h
```diff
@@ -60,6 +69,13 @@ class IKeyNavigationSubSection : public IKeyNavigation
 public:
     virtual ~IKeyNavigationSubSection() = default;
 
+    enum class Direction {
+        Horizontal = 0,
+        Vertical,
+        Grid
```
RomanPudashkin, line 75: I think Layout is a better name for this enum because Grid isn't a direction

src/framework/ui/view/keynavigationsubsection.h
```diff
@@ -32,18 +32,30 @@ class KeyNavigationControl;
...
     Q_OBJECT
-    Q_PROPERTY(KeyNavigationSection * section READ section WRITE setSection)
+    Q_PROPERTY(KeyNavigationSection * section READ section WRITE setSection NOTIFY sectionChanged)
+    Q_PROPERTY(QmlDirection direction READ direction_property WRITE setDirection NOTIFY directionChanged)
 
 public:
     explicit KeyNavigationSubSection(QObject* parent = nullptr);
     ~KeyNavigationSubSection() override;
 
+    //! NOTE Must be sync with IKeyNavigationSubSection::Direction
```
RomanPudashkin, line 42: I afraid that we'll forget to add new values here

src/framework/ui/ikeynavigation.h
igorkorsukov: I do not think it's a good idea.
This enum is about directions of navigation, not about location.
Perhaps the `Grid` can be renamed, for example to `Table`, or `Both`.

src/instruments/view/instrumentpaneltreemodel.cpp
```diff
@@ -344,6 +346,7 @@ int InstrumentPanelTreeModel::columnCount(const QModelIndex&) const
 
 QVariant InstrumentPanelTreeModel::data(const QModelIndex& index, int role) const
 {
+    TRACEFUNC;
```
RomanPudashkin, line 349: I think that better to remove TRACEFUNC from data and isSelected because those methods are trivial

src/framework/ui/ikeynavigation.h
RomanPudashkin: I like the idea to rename Grid to Both

igorkorsukov: renamed

src/framework/ui/view/keynavigationsubsection.h
igorkorsukov: added comment in both place

src/instruments/view/instrumentpaneltreemodel.cpp
igorkorsukov: By the way, I wanted to see how many times this method is called. Two instruments, one (piano) to expanded, two more elements appear - 59 calls to this method :)

igorkorsukov: removed

src/framework/ui/ikeynavigation.h
vpereverzev: It's still "Grid" in keynavigationsubsection.h

If it's a direction of navigation, I'd propose to name it like "LeftToRight" or "FromLeftToRight". Because even in the horizontal direction you have 2 possible ways to change focus (from left to right, from right to left).

igorkorsukov: > It's still "Grid" in keynavigationsubsection.h

Thanks, fixed

vpereverzev (approved):

PR #7934, opened 2021-04-11
Refactor Playback Toolbar to use AbstractMenuModel and StyledMenu

cbjeukendrup:
Screenshot


src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenuItem.qml
```diff
@@ -16,6 +16,7 @@ ListItemBlank {
 
     property var modelData
     property bool reserveSpaceForInvisibleItems: true
+    property bool unifyCheckMarksAndIcons: true
```
RomanPudashkin, line 19: I don't quite understand what this flag does. Can you explain me, please?

src/playback/view/playbacktoolbarmodel.cpp
```diff
@@ -59,119 +58,170 @@ QVariant PlaybackToolBarModel::data(const QModelIndex& index, int role) const
...
     playbackController()->isPlayAllowedChanged().onNotify(this, [this]() {
-        updateState();
+        emit maxPlayTimeChanged();
         updatePlayTime();
         emit isPlayAllowedChanged();
     });
 
     playbackController()->isPlayingChanged().onNotify(this, [this]() {
-        updateState();
+        onActionsStateChanges({ "play" });
```
RomanPudashkin, line 110: Better to add a static string at the top of the file for this code because I see it in several places

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenuItem.qml
cbjeukendrup: When this is true, the checkmark of a checked action appears _instead of_ the icon.
When it is false, the checkmark appears _on the left of_ the icon. 
| unifyCheckMarksAndIcons = true | | unifyCheckMarksAndIcons = false |
|-|-|-|
|Schermafbeelding 2021-04-11 om 16 03 36|vs|Schermafbeelding 2021-04-11 om 16 00 31|

src/playback/view/playbacktoolbarmodel.cpp
cbjeukendrup: Thanks, fixed!

src/playback/view/playbacktoolbarmodel.h
```diff
@@ -40,6 +40,7 @@ class PlaybackToolBarModel : public QAbstractListModel, public async::Asyncable
     INJECT(playback, IPlaybackController, playbackController)
     INJECT(playback, workspace::IWorkspaceManager, workspaceManager)
 
+    Q_PROPERTY(bool floating READ floating WRITE setFloating NOTIFY floatingChanged)
```
RomanPudashkin, line 43: I'm not sure "floating" is the best name for this property but at least you should rename it to "isFloating" to provide uniformity with "isPlayAllowed" :)

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenuItem.qml
RomanPudashkin: Oh, I see. But do we really need to display the checkmark instead of the icon somewhere?

src/playback/view/playbacktoolbarmodel.h
cbjeukendrup: I renamed it to "isToolbarFloating" :)

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenuItem.qml
RomanPudashkin: It seems that we should always display both the checkmark and the icon (correct me if I'm wrong)

cbjeukendrup: Maybe that's a question for @Tantacrul. I did this to keep the existing behaviour in other places, but maybe the existing behaviour was not correct.

RomanPudashkin: I think that better to remove this flag and always display both the checkmark and the icon. We can change this behaviour later if need be

cbjeukendrup: Oops, messages crossed :) 
I will do it like this:
- if any item in a menu has a checkmark, all items have a "column" for checkmarks
- if any item has an icon, all items will have a "column" for icons
- but if all items that are checkable don't have an icon, the checkmarks and icons are in the same "column" because that saves space.

Tantacrul: > Oops, messages crossed :)
> I will do it like this:
> 
> * if any item in a menu has a checkmark, all items have a "column" for checkmarks
> * if any item has an icon, all items will have a "column" for icons
> * but if all items that are checkable don't have an icon, the checkmarks and icons are in the same "column" because that saves space.

That makes sense to me. This kind of flexibility in context menus can be seen on a few other apps and OS's too.

cbjeukendrup:
@Tantacrul Here are some screenshots of how it looks now, is it okay? You could also test the build to be completely sure :)
Schermafbeelding 2021-04-11 om 21 48 35
Schermafbeelding 2021-04-11 om 21 49 02

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledTextLabel.qml
```diff
@@ -3,6 +3,8 @@ import QtQuick 2.12
 Text {
     id: root
 
+    readonly property bool isEmpty: Boolean(text)
```
RomanPudashkin, line 6: You have a bug here. Should be:

readonly property bool isEmpty: !Boolean(text)

or just:
readonly property bool isEmpty: text.length === 0

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
```diff
@@ -21,11 +21,45 @@ StyledPopupView {
     property alias model: view.model
     property int itemWidth: 300
     property bool reserveSpaceForInvisibleItems: true
+    property bool unifyCheckMarksAndIcons: true
```
RomanPudashkin, line 24: It seems that this flag is not needed anymore

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
```diff
@@ -21,11 +21,45 @@ StyledPopupView {
...
+                       && (modelData.checkable || (modelData.icon && modelData.icon !== IconCode.NONE))) {
+                privateProperties.hasItemsWithIconOrCheckable = true
+            }
+
+            if (!privateProperties.hasItemsWithSubmenu
+                    && Boolean(modelData.subitems) && modelData.subitems.length > 0) {
+                privateProperties.hasItemsWithSubmenu = true
+            }
+
+            if (!privateProperties.hasItemsWithShortcut
```
RomanPudashkin, line 51: Can we simplify this code a bit?

if (Boolean(modelData.shortcut)) {
privateProperties.hasItemsWithShortcut = true
}

Same for all conditions in if above

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
```diff
@@ -21,11 +21,45 @@ StyledPopupView {
...
+
+        for (let i = 0; i < model.length; i++) {
+            let modelData = model[i]
+            if (!privateProperties.hasItemsWithIconAndCheckable
+                    && modelData.checkable
+                    && (Boolean(modelData.icon) && modelData.icon !== IconCode.NONE)) {
+                privateProperties.hasItemsWithIconAndCheckable = true
+                privateProperties.hasItemsWithIconOrCheckable = true
+            } else if (!privateProperties.hasItemsWithIconOrCheckable
+                       && (modelData.checkable || (modelData.icon && modelData.icon !== IconCode.NONE))) {
```
RomanPudashkin, line 42: You have this code in several places:
modelData.icon && modelData.icon !== IconCode.NONE

It is better to add some temp variable for this condition after this line
let modelData = model[i]

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
```diff
@@ -21,11 +21,45 @@ StyledPopupView {
...
     signal handleAction(string actionCode, int actionIndex)
 
+    onModelChanged: {
+        privateProperties.hasItemsWithIconAndCheckable = false
+        privateProperties.hasItemsWithIconOrCheckable = false
+        privateProperties.hasItemsWithSubmenu = false
+        privateProperties.hasItemsWithShortcut = false
+
+        for (let i = 0; i < model.length; i++) {
+            let modelData = model[i]
```
RomanPudashkin, line 35: Better to add an empty line here

igorkorsukov:
@cbjeukendrup You solved this problem in a very timely! Thank you! 
Could you please fix the conflict? We would like to merge your work (even perhaps without correcting comments, they can be corrected separately). because the task of keyboard navigation depends on this work, which is now in active development.

cbjeukendrup:
@igorkorsukov I'm currently working on the rebase, and I have already fixed the review comments along the way. Force-push coming very soon!

src/framework/uicomponents/qml/MuseScore/UiComponents/ListItemBlank.qml
```diff
@@ -24,7 +24,9 @@ FocusableControl {
     Accessible.selectable: true
     Accessible.selected: isSelected
 
-    mouseArea.hoverEnabled: Boolean(root.hint) && root.visible
+    background.color: normalStateColor
+
+    mouseArea.hoverEnabled: root.visible
```
cbjeukendrup, line 29: Hovering is not only needed when there is a tooltip, but also for the background color.

igorkorsukov: Exactly, thanks for noticing and correcting

Tantacrul:
Quick one:

If you install the font on your machine, it (for some reason) gets
prioritised over the one packaged with MuseScore. Then when new icons are
added, they don't appear in MS4. Just install the font from your machine in
that instance. I'm guessing this isn't your issue, but just in case it is,
I thought I'd mention it.

cbjeukendrup:
@Tantacrul Ah, that was it!! Thanks!

Tantacrul:
He he he

PR #7936, opened 2021-04-11
Fix handling actions from NotationPaintView context menu

cbjeukendrup:
The `handleAction` method got lost in commit 42ce82d0e5e58d90f552f47c92abf6bebe609222. Now it's back.

@Eism Did you remove it on purpose or had you just missed that it was still used in NotationView.qml?


Eism:
Yes, it is my fault :)

PR #7938, opened 2021-04-12
318949 notationcontextinspector

Rice-Puddn:
Resolves: https://musescore.org/en/node/318949

Added the "inspector" item to the notation right-click context menu to make it more rapidly accessible. 


RomanPudashkin:
@Tantacrul please, check this

Tantacrul:
I don't think this is a solid enough solution. If we are going to do that, we should be consistent.

So: in the context menu, there should be an option called 'View' which contains all the panels (Palettes, Inspector, Instruments, etc.). There should be a 'tick' next to ones that are currently open.

Issue #7654, opened 2021-03-06
[MU4 Issue] If multiple instruments selected and click to hide them, only one item hidden


vshalkevich:
**Describe the bug**
Only clicked instrument hidden

**To Reproduce**
1. Create a score with multiple instruments
2. Select several instruments from the left side panel
3. Click on an eye near instrument to hide it

**Expected behavior**
All selected instruments should be hidden when we click on an eye near any of selected instruments

**Desktop (please complete the following information):**
MacOS

Tantacrul:
Fixed

HemantAntony:
@Tantacrul this issue still occurs

Tantacrul:
It's not an issue.

Tantacrul:
OK, reopening, since @HemantAntony has a PR for the request ready to be reviewed.

RomanPudashkin:
@Tantacrul could you please check it? I can merge it right now

HemantAntony:
@RomanPudashkin There are two. This is mine: #10203

Tantacrul:
@RomanPudashkin Will do now.

@HemantAntony - actually, you can send any PR's you want tested, which you small to either me, @bkunda or @DmitryArefiev if the engineers are busy. Do recommend this to others also!

RomanPudashkin:
> @RomanPudashkin There are two. This is mine: #10203

@Tantacrul 
Oops, sorry, I meant this PR:
https://github.com/musescore/MuseScore/pull/10656

Tantacrul:
@HemantAntony - the artefact has expired :(

<img width="710" alt="image" src="https://user-images.githubusercontent.com/21022311/166104627-9a053b81-ac1b-4a0c-9024-13915a037140.png">

HemantAntony:
I'll rebase and it should be there soon

Tantacrul:
Tested. Looking good to me.

PR #7940, opened 2021-04-12
[MU4 Issue] If multiple instruments selected and click to hide them, only one item hidden

FergalOK:
Resolves: https://github.com/musescore/MuseScore/issues/7654

 All selected instruments should be hidden when we click on an eye near any of selected instruments


src/instruments/qml/MuseScore/Instruments/internal/InstrumentsTreeItemDelegate.qml
```diff
@@ -245,7 +245,12 @@ Item {
                     return
                 }
 
-                model.itemRole.isVisible = !model.itemRole.isVisible
+                if (!root.isSelected) {
+                    model.itemRole.isVisible = !model.itemRole.isVisible
+                }
+                else {
+                    instrumentTreeModel.setIsVisibleSelectedRows(!model.itemRole.isVisible)
```
Eism, line 252: The delegate does not know about the model it is in. It is an independent component that we can use with another model. Another model may not have a setIsVisibleSelectedRows method and we will have an error.
It is better to make the signal visibilityChangeRequested(isVisible) and process the signal in the InstrumentsPanel

PS: The editor highlights properties(like instrumentTreeModel) that are declared by the parent but not in the current file. Pay attention to it

src/instruments/qml/MuseScore/Instruments/internal/InstrumentsTreeItemDelegate.qml
```diff
@@ -245,7 +245,12 @@ Item {
                     return
                 }
 
-                model.itemRole.isVisible = !model.itemRole.isVisible
+                if (!root.isSelected) {
+                    model.itemRole.isVisible = !model.itemRole.isVisible
+                }
+                else {
```
Eism, line 251: Please do not wrap else on a new line
https://github.com/musescore/MuseScore/wiki/CodeGuidelines#braces

src/instruments/qml/MuseScore/Instruments/internal/InstrumentsTreeItemDelegate.qml
FergalOK: Thanks for the feedback, I believe I did what you suggested
I would appreciate any other feedback you have to give since this is the first time I'm using QT/QML

PR #7944, opened 2021-04-12
[MU4] vst ressurection part2

vpereverzev:
This PR introduces VST audio processing infrastructure. Currently we're able to work with VSTi plugins. We will also need to implement support for regular VST audio effects soon.

Here is an example of work with BBC Orchestral plugin:

https://user-images.githubusercontent.com/57620349/114443997-a9d7ea00-9bce-11eb-94e8-969b567be1e1.mp4

Important note: this PR should be followed by - https://github.com/musescore/MuseScore/pull/7945

### DIFF
--- a/src/framework/audio/internal/synthesizers/synthesizersregister.cpp
+++ b/src/framework/audio/internal/synthesizers/synthesizersregister.cpp
@@ -29,7 +29,7 @@ void SynthesizersRegister::registerSynthesizer(const SynthName& name, ISynthesiz
     std::lock_guard<std::mutex> lock(m_mutex);
     m_synths[name] = std::make_shared<SanitySynthesizer>(s);
 
-    m_synthNotificationChannel.send(m_synths[name]);
+    m_synthAdded.send(m_synths[name]);
 }
 
 std::shared_ptr<ISynthesizer> SynthesizersRegister::synthesizer(const SynthName& name) const
@@ -54,9 +54,9 @@ std::vector<std::shared_ptr<ISynthesizer> > SynthesizersRegister::synthesizers()
     return synths;
 }
 
-async::Channel<ISynthesizerPtr> SynthesizersRegister::synthesizerAddedNotify() const
+async::Channel<ISynthesizerPtr> SynthesizersRegister::synthesizerAdded() const
 {
-    return m_synthNotificationChannel;
+    return m_synthAdded;
 }
 
 void SynthesizersRegister::setDefaultSynthesizer(const SynthName& name)
--- a/src/framework/audio/internal/synthesizers/synthesizersregister.cpp
+++ b/src/framework/audio/internal/synthesizers/synthesizersregister.cpp
@@ -20,6 +20,7 @@
 #include "internal/audiosanitizer.h"
 #include "sanitysynthesizer.h"
 
+using namespace mu;
 using namespace mu::audio::synth;
 
 void SynthesizersRegister::registerSynthesizer(const SynthName& name, ISynthesizerPtr s)
@@ -27,6 +28,8 @@ void SynthesizersRegister::registerSynthesizer(const SynthName& name, ISynthesiz
     ONLY_AUDIO_MAIN_OR_WORKER_THREAD;
     std::lock_guard<std::mutex> lock(m_mutex);
     m_synths[name] = std::make_shared<SanitySynthesizer>(s);
+
+    m_synthNotificationChannel.send(m_synths[name]);
 }
 
 std::shared_ptr<ISynthesizer> SynthesizersRegister::synthesizer(const SynthName& name) const
@@ -49,6 +52,11 @@ std::vector<std::shared_ptr<ISynthesizer> > SynthesizersRegister::synthesizers()
         synths.push_back(it->second);
     }
     return synths;
+}
+
+async::Channel<ISynthesizerPtr> SynthesizersRegister::synthesizerAddedNotify() const
+{
+    return m_synthNotificationChannel;
 }
 
 void SynthesizersRegister::setDefaultSynthesizer(const SynthName& name)
--- a/src/framework/audio/internal/worker/audioengine.cpp
+++ b/src/framework/audio/internal/worker/audioengine.cpp
@@ -99,11 +99,14 @@ void AudioEngine::onDriverOpened(unsigned int sampleRate, uint16_t readBufferSiz
     setSampleRate(sampleRate);
     setReadBufferSize(readBufferSize);
 
-    //! TODO Add a subscription to add or remove synthesizers
     std::vector<ISynthesizerPtr> synths = synthesizersRegister()->synthesizers();
     for (const ISynthesizerPtr& synth : synths) {
         m_mixer->addChannel(synth);
     }
+
+    synthesizersRegister()->synthesizerAddedNotify().onReceive(this, [this](const ISynthesizerPtr& synth) {
+       m_mixer->addChannel(synth);
+    });
 }
 
 void AudioEngine::setSampleRate(unsigned int sampleRate)
--- a/src/framework/audio/internal/worker/audioengine.cpp
+++ b/src/framework/audio/internal/worker/audioengine.cpp
@@ -104,8 +104,8 @@ void AudioEngine::onDriverOpened(unsigned int sampleRate, uint16_t readBufferSiz
         m_mixer->addChannel(synth);
     }
 
-    synthesizersRegister()->synthesizerAddedNotify().onReceive(this, [this](const ISynthesizerPtr& synth) {
-       m_mixer->addChannel(synth);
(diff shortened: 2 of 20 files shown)


src/framework/audio/internal/synthesizers/synthesizersregister.cpp
```diff
@@ -51,6 +54,11 @@ std::vector<std::shared_ptr<ISynthesizer> > SynthesizersRegister::synthesizers()
     return synths;
 }
 
+async::Channel<ISynthesizerPtr> SynthesizersRegister::synthesizerAddedNotify() const
+{
+    return m_synthNotificationChannel;
```
igorkorsukov, line 59: I think the `Notify` in the method name is superfluous, and it is better for the channel name to match the name of the method, i.e.
```
async::Channel<ISynthesizerPtr> SynthesizersRegister::synthesizerAdded() const
{
    return m_synthesizerAdded;
}
```

src/framework/audio/internal/worker/audioengine.cpp
```diff
@@ -99,11 +99,14 @@ void AudioEngine::onDriverOpened(unsigned int sampleRate, uint16_t readBufferSiz
...
 
-    //! TODO Add a subscription to add or remove synthesizers
     std::vector<ISynthesizerPtr> synths = synthesizersRegister()->synthesizers();
     for (const ISynthesizerPtr& synth : synths) {
         m_mixer->addChannel(synth);
     }
+
+    synthesizersRegister()->synthesizerAddedNotify().onReceive(this, [this](const ISynthesizerPtr& synth) {
+       m_mixer->addChannel(synth);
+    });
```
igorkorsukov, line 109: also needs removing

src/framework/midi/midievent.h
```diff
@@ -982,12 +982,12 @@ struct Event {
     void assertMessageType(const std::set<MessageType>& supportedTypes) const
     {
         UNUSED(supportedTypes);
-        assert(isMessageTypeIn(supportedTypes));
+        //assert(isMessageTypeIn(supportedTypes));
```
igorkorsukov, line 985: need a comment why disabled

src/framework/midi/midievent.h
```diff
@@ -982,12 +982,12 @@ struct Event {
...
     }
 
-    void assertChannelVoice() const { assert(isChannelVoice()); }
+    void assertChannelVoice() const { /*assert(isChannelVoice());*/ }
```
igorkorsukov, line 988: need a comment why disabled

src/framework/vst/internal/synth/vstaudioclient.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+VstAudioClient::~VstAudioClient()
+{
+    IAudioProcessorPtr processor = m_pluginComponent;
+    if (!processor) {
+        return;
+    }
+
+    processor->setProcessing(false);
+    m_pluginComponent->setActive(false);
```
igorkorsukov, line 16: A bit strange and confusing code. Better or everywhere to use a class member or variable

src/framework/vst/internal/synth/vstaudioclient.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+void VstAudioClient::init(PluginComponentPtr component)
+{
+    if (!component) {
+        return;
+    }
```
igorkorsukov, line 23: maybe better `if assert`?

src/framework/vst/internal/synth/vstaudioclient.cpp
```diff
@@ -0,0 +1,153 @@
...
+    if (m_eventList.addEvent(vstFromMidi(e)) == Steinberg::kResultTrue) {
+        return true;
+    }
+
+    return false;
+}
+
+void VstAudioClient::process(float* output, unsigned int samples)
+{
+    IAudioProcessorPtr processor = m_pluginComponent;
```
igorkorsukov, line 39: maybe better to add a method to get a processor

src/framework/vst/internal/synth/vstaudioclient.cpp
```diff
@@ -0,0 +1,153 @@
...
+{
+    for (unsigned int i = 0; i < samples; ++i) {
+        for (unsigned int s = 0; s < audio::synth::AUDIO_CHANNELS; ++s) {
+            auto getFromChannel = std::min<unsigned int>(s, m_processData.outputs[0].numChannels - 1);
+            output[i * audio::synth::AUDIO_CHANNELS + s] = m_processData.outputs[0].channelBuffers32[getFromChannel][i];
+        }
+    }
+}
+
+VstEvent VstAudioClient::vstFromMidi(const mu::midi::Event& event)
```
igorkorsukov, line 116: I think better to change the method signature 
```
bool VstAudioClient::vstFromMidi(const mu::midi::Event& event, VstEvent& out)
{
...
}
```
to avoid copying

src/framework/vst/internal/synth/vstsynthesiser.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+bool VstSynthesiser::isActive() const
+{
+    return m_isActive;
+}
+
+void VstSynthesiser::setIsActive(bool arg)
+{
+    m_isActive = arg;
+}
```
igorkorsukov, line 47: m_pluginComponent (in the VstAudioClient) has method isActive, I think better use it
```
void VstSynthesiser::setIsActive(bool arg)
{
     m_vstAudioClient->setIsActive(arg);
```

src/framework/vst/internal/synth/vstsynthesiser.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+std::string VstSynthesiser::name() const
+{
+    return m_pluginMeta.name;
+}
+
+audio::synth::SoundFontFormats VstSynthesiser::soundFontFormats() const
+{
+    NOT_SUPPORTED;
+    return { audio::synth::SoundFontFormat::Undefined };
```
igorkorsukov, line 57: Maybe better add new value, like `SoundFontFormat::Embeded` or `SoundFontFormat::External`

src/framework/vst/internal/synth/vstsynthesiser.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+void VstSynthesiser::flushSound()
+{
+    NOT_IMPLEMENTED;
+}
+
+Ret VstSynthesiser::setupChannels(const std::vector<midi::Event>& /*events*/)
+{
+    NOT_IMPLEMENTED;
+    return Ret(Ret::Code::Ok);
```
igorkorsukov, line 99: I think need just send events to m_vstAudioClient (call m_vstAudioClient->handleEvent(e))

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -112,6 +112,17 @@ PluginView VstPlugin::view()
     return m_pluginView;
 }
```
igorkorsukov, line 113: I think, better method make `const` and for m_pluginView add `mutable`

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -112,6 +112,17 @@ PluginView VstPlugin::view()
...
+PluginComponentPtr VstPlugin::component()
+{
+    if (m_pluginComponent) {
+        return m_pluginComponent;
+    }
+
+    m_pluginComponent = m_pluginProvider->getComponent();
+
+    return m_pluginComponent;
+}
```
igorkorsukov, line 124: I think, better method make `const` and for m_pluginComponent add `mutable`

vpereverzev: yep, also makes sense for view() method

src/framework/vst/internal/synth/vstsynthesiser.cpp
vpereverzev: Embeded sounds good to me

src/framework/vst/internal/synth/vstsynthesiser.cpp
vpereverzev: yeah, I'm going to care about "channels" in the next PRs, maybe we can also add functionality for channel management (volume, balance, etc.). I just need to dive deeper first

src/framework/vst/internal/synth/vstaudioclient.cpp
vpereverzev: There was no reason for that previously since it was the same amount of symbols. But now I'm using static_cast there, so private getter would makes sense

src/framework/vst/internal/synth/vstaudioclient.cpp
vpereverzev: Actually, you're right, since the method is going to be called really often. Good proposal

PR #7947, opened 2021-04-13
[MU3] Fix #319079: Shift Selecting Last Word

Jojo-Schmitz:
Resolves: *https://musescore.org/en/node/319079*

Original code only attempts to find the next ChordRest segment as the end marker for selecting a range, this fix adds the detection for the EndBarLine segment as a potential end marker, which resolves the issue.

Backport of #7793 to the 3.x branch


vpereverzev:
merges to 3.x are closed now

PR #7948, opened 2021-04-13
Fixed build on msvc

Eism:


src/appshell/internal/notationpagestate.cpp
```diff
@@ -46,7 +46,13 @@ bool NotationPageState::isPanelVisible(PanelType type) const
...
     case PanelType::NotationStatusBar:
         return configuration()->isNotationStatusBarVisible().val;
-    default:
+    case PanelType::Mixer:
+    case PanelType::TimeLine:
+    case PanelType::Synthesizer:
+    case PanelType::SelectionFilter:
+    case PanelType::Piano:
+    case PanelType::ComparisonTool:
+    case PanelType::Undefined:
```
Jojo-Schmitz, line 55: with that you could add the `return false` outside the switch, so the code still works as intended if more types get added, even if the warning that there is an unhandled case got ignored or overseen

Eism: `return false` is already outside the switch

Jojo-Schmitz: Ah, right, you removed it in an eralier commit, but reinstated it in this. Sorry for the noise...

Jojo-Schmitz (approved):

RomanPudashkin (approved):

PR #7949, opened 2021-04-13
[MU4] Added key navigation for palette panel 

igorkorsukov:
Just prototype 

Navigation pattern

```
     c1   c2        c3      c4
r1   [    [e]       [d]     [m] ]       
r2        [p1]
r3        [p2]
r4        [pn]
r5        [mo]
r6   [    [e]       [d]     [m] ]       
r7        [p1]
r8        [p2]
r9        [pn]
r10       [mo]
r11  [    [e]       [d]     [m] ]    
r12  [    [e]       [d]     [m] ]   
```

https://user-images.githubusercontent.com/3818029/114538190-61164480-9c53-11eb-903a-8d3ca4d2f1a5.mp4

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
@@ -33,17 +33,8 @@ FocusableControl {
 
     opacity: root.enabled ? 1.0 : ui.theme.itemOpacityDisabled
 
-    onInternalClicked: {
-        root.clicked()
-    }
-
-    onInternalPressAndHold: {
-        root.pressAndHold()
-    }
-
-    onInternalTriggered: {
-        root.clicked()
-    }
+    mouseArea.onClicked: root.clicked()
+    mouseArea.onPressAndHold: root.pressAndHold()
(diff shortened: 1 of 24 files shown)


src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
```diff
@@ -33,30 +33,10 @@ FocusableControl {
...
+    mouseArea.onPressAndHold: root.pressAndHold()
 
-    mouseArea.hoverEnabled: true
-    mouseArea.onContainsMouseChanged: {
-        if (!Boolean(root.hint)) {
-            return
-        }
-
-        if (mouseArea.containsMouse) {
-            ui.tooltip.show(this, root.hint)
```
Eism, line 55: do we no longer show the tooltip for buttons on hover?

igorkorsukov: Oops, this is a mistake, possibly an incorrectly resolved conflict. 
Fixed (returned), thanks!

Eism (approved):

PR #7952, opened 2021-04-13
Fix #318860: Restore rest offset with slash notation in voices 3 and 4

tobik:
Resolves: *https://musescore.org/en/node/318800#comment-1074094*

*Reintroduce offsets for rests in voice 3 and 4 (1-based) in connection with slash notation* 

According to https://musescore.org/en/node/318800#comment-1074063 slash notation in voice 1 and 2 are full size and not offset outside the stuff, voice 3 and 4 are small and outside the stuff.


Jojo-Schmitz (approved):

libmscore/rest.cpp
```diff
@@ -532,7 +532,7 @@ int Rest::computeLineOffset(int lines)
                   }
             }
 
-      if (offsetVoices) {
+      if (offsetVoices && voice() < 2) {
```
Jojo-Schmitz, line 535: This means voice 1 and 2, right? IIRC `voice()` is 0 based.

tobik: Yes, don't offset voice 1 and voice 2 in combination with slash notation, but do offset voice 3 and 4

Jojo-Schmitz: Might be worth a comment explaining this?

tobik: For detailed description please see https://musescore.org/en/node/318800#comment-1074063

Jojo-Schmitz: Yes, that's where I found it. Might be good though to summarize this in a short comment in the code

Jojo-Schmitz:
The 1st vtest fail shows an expected difference, and indeed needs fixing of the ref file.
The 2nd vtest fail is expected too, but for a different reason: it does not take the new/change ref file into account, but takes the one from the previous commit. So it is OK to have this vtest fail.

tobik:
Then vtest will be OK with the next commit? One more reason to add the suggested remark :-)

Jojo-Schmitz:
Not sure whether the vtests pick on the ref files of the previous commit, or the one the entire PR is based on.
But when in doubt, try it out ;-) and add a 3rd commit to add such a comment.
If it helps getting past the vtest fail, fine, if it doesn't, fine too, in any case I guess this PR should have only 1 commit in the end, and then would fail the vtest again. As we know about it, and verfied that the vtest fail is OK and has been dealt with properly, this won't be a problem as far as getting this PR merged. Except that it may need to get ported over to master (too).

Jojo-Schmitz (approved):

Jojo-Schmitz:
OK, apparently the vtests use the last commit before the entire PR for comparison

MarcSabatella (approved):
Looks good to me!

Jojo-Schmitz:
I guess you should squash these 3 commits into just one. I've ported it over to master meanwhile. see #8017

vpereverzev:
merged into master, 3.x merges are closed

Issue #7925, opened 2021-04-10
[MU4 Issue] No playback when entering notes (plays when clicking on them though)


vshalkevich:
**Describe the bug**
No playback when entering notes (plays when clicking on them though)

**To Reproduce**
1. Create a score
2. Click on a Pencil icon (or CTRL + N shortcut) to enter Note Input Mode
3. Put a note on the Score

**Expected behavior**
Note should make sound when it`s added on a score (like in MuseScore 3)

**Desktop (please complete the following information):**
MacOS
Windows 10

PR #7954, opened 2021-04-14
[MU4] Implement playback when entering notes

aaaeka:
Resolves: #7925 

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -425,6 +425,8 @@ void NotationActionController::addNote(NoteName note, NoteAddingMode addingMode)
     }
 
     noteInput->addNote(note, addingMode);
+
+    playSelectedElement();
 }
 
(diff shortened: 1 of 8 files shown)


src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1386,6 +1392,25 @@ void NotationActionController::toggleScoreConfig(ScoreConfigType configType)
...
+    Element* element = interaction->selection()->element();
+    if (!element || !element->isNote()) {
+        return;
+    }
+
+    if (playbackConfiguration()->playChordWhenEditing()) {
+        element = element->elementBase();
+    }
+
+    playbackController()->playElementOnClick(element);
```
igorkorsukov, line 1411: I think now it is better to rename the method `playElementOnClick` to like `playElement` :)

aaaeka: Renamed

PR #7956, opened 2021-04-14
[MU4] Fix next\previous chord action in multiple element selection

aaaeka:
Before, when there was a multiple element selection, actions ```next-chord``` and ```prev-chord``` would not work.

A simple 1 line fix.

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -578,11 +578,12 @@ void NotationActionController::moveAction(const actions::ActionCode& actionCode)
         return;
     }
 
-    Element* element = interaction->selection()->element();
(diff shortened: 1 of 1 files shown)


src/notation/internal/notationactioncontroller.cpp
```diff
@@ -578,7 +578,7 @@ void NotationActionController::moveAction(const actions::ActionCode& actionCode)
         return;
     }
 
-    Element* element = interaction->selection()->element();
+    Element* element = interaction->selection()->elements().back();
```
igorkorsukov, line 581: I think, needs to check on `elements` not empty

aaaeka: Fixed!

PR #7958, opened 2021-04-15
[MU4] license upgrade

vpereverzev:
Migration on GPLv3 version


shoogle:
Any particular reason for closing? BTW, it's not really necessary to ignore this commit for the blame because it only affects the comments at the top of the file, not actual code. Also, the changes are important (not just whitespace) so we probably don't want to ignore them anyway.

PR #7960, opened 2021-04-15
[MU4] license update

vpereverzev:
gplv3 license migration


CMakeLists.txt
```diff
@@ -1,21 +1,24 @@
...
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License
-#  along with this program; if not, write to the Free Software
-#  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
-#=============================================================================
+#[[
+// SPDX-License-Identifier: GPL-3.0-only
```
shoogle, line 2: Why not just replace `//` with `#` on all lines for CMakeLists.txt, *.cmake, shell scripts, and Python files?

build/Linux+BSD/mscore.1.preview.sh
```diff
@@ -1,4 +1,25 @@
 #!/bin/sh
+##
```
shoogle, line 2: I'd say this line above is unnecessary. In the Linux kernel, the identifier always starts straight after the sheband line.

CMakeLists.txt
vpereverzev: This is basically a result of automated work from external utility, so I don't really care about the specific formatting cases

build/pch/pch.h
```diff
@@ -1,21 +1,24 @@
...
-//  but WITHOUT ANY WARRANTY; without even the implied warranty of
-//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-//  GNU General Public License for more details.
-//
-//  You should have received a copy of the GNU General Public License
-//  along with this program; if not, write to the Free Software
-//  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
-//=============================================================================
+/**
+ * // SPDX-License-Identifier: GPL-3.0-only
```
shoogle, line 2: Something has gone wrong here. The formatter has added a second comment level when `//` is already sufficient.

CMakeLists.txt
shoogle: I can only think of two cases though:

1. Files that use `//` as comments (`.c`, `.h`, `.cpp`, `.hpp`)
2. Files that use `#` as comments (`.sh`, `.bash`, `.py`, `.cmake`, `CMakeLists.txt`)

So it might worth writing a simple shell script to add the license header based on extention. I could do this if you don't have time.

But if you want to use the existing formatter then all you need to do is remove the `//` from each line and pass in the license text on its own:

```
SPDX-License-Identifier: GPL-3.0-only
MuseScore-CLA-applies
=============================================================================
MuseScore
Music Composition & Notation

Copyright (C) 2021 MuseScore BVBA and others

This program is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License version 3 as
published by the Free Software Foundation.

This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU General Public License
along with this program.  If not, see <https://www.gnu.org/licenses/>.
=============================================================================
```

The problem currently is that each line starts with `//` so the formatter things that is supposed to be *part of* the comment rather than the escape sequence to introduce the comment. This is why the `.cpp` files are getting `/**` added before the comment.

build/pch/pch.h
shoogle: Ok, I realise what the problem is. The formatter thinks `//` is part of the comment, so you simply need to remove that and pass the license text in on its own. See my [comment above](https://github.com/musescore/MuseScore/pull/7960/files#discussion_r614445858).

build/ci/linux/build_utest.sh
```diff
@@ -1,4 +1,25 @@
 #!/usr/bin/env bash
+##
+## // SPDX-License-Identifier: GPL-3.0-only
+## // MuseScore-CLA-applies
+## //=============================================================================
+## //  MuseScore
+## //  Music Composition & Notation
+## //
```
igorkorsukov, line 8: a little redundant, I would like to see just `#` :)

src/appshell/view/dockwindow/eventswatcher.h
```diff
@@ -1,21 +1,24 @@
...
-//=============================================================================
+/**
+ * // SPDX-License-Identifier: GPL-3.0-only
+ * // MuseScore-CLA-applies
+ * //=============================================================================
+ * //  MuseScore
+ * //  Music Composition & Notation
+ * //
+ * //  Copyright (C) 2021 MuseScore BVBA and others
+ * //
```
igorkorsukov, line 9: a little redundant, I would like to see just 
```
/**

*/
```
or 
```
//
//
```

build/ci/linux/build.sh
```diff
@@ -1,4 +1,25 @@
 #!/usr/bin/env bash
+##
+## // SPDX-License-Identifier: GPL-3.0-only
+## // MuseScore-CLA-applies
```
igorkorsukov, line 4: Do we have to indicate these two lines? Should they be above the line? I would like to see them below the line, if they are required, it's easier to perceive

src/appshell/view/dockwindow/eventswatcher.h
igorkorsukov: or better just `/*` not `/**`
Because it is better for QtCreator to highlight it as a simple comment - dull gray, and not as a doxygen comment - bright green (for me, but I think so for others :))

build/ci/linux/build.sh
shoogle: These two lines are not necessary if the main text is present, but equally the main text is not necessary if these lines are present. So my preference would be to remove the main text and keep just the following:

```
#!/usr/bin/env shell
# SPDX-License-Identifier: GPL-3.0-only
# MuseScore-CLA-applies
#
# MuseScore
# Music Composition & Notation
#
# Copyright (C) 2021 MuseScore BVBA and others
```

The SPDX license identifier is an industry standard shorthand for the license in question (see https://spdx.org/licenses/).

This is how the [Linux kernel does it](https://github.com/torvalds/linux/blob/master/kernel/audit.c#L1-L9) for example. The kernel licensing rules page says (just above [here](https://www.kernel.org/doc/html/latest/process/license-rules.html#license-identifier-syntax)):

> SPDX license identifiers are machine parsable and precise shorthands for the license under which the content of the file is contributed. SPDX license identifiers are managed by the SPDX Workgroup at the Linux Foundation and **have been agreed on by partners throughout the industry, tool vendors, and legal teams**.

There is no need to state the full disclaimer because that is already part of the GPL itself. Even the link to the GPL is not needed because these days it is a widely recognised license and an internet search for "GPL 3" gives the correct result (for example, it has been [confirmed in a US court](https://en.wikipedia.org/wiki/Creative_Commons_license#Drauglis_v._Kappa_Map_Group,_LLC) that "CC BY-SA 2.0" without a link is sufficient text to identify the [Creative Commons Attribution-ShareAlike 2.0 license](https://creativecommons.org/licenses/by-sa/2.0/)).

.git-blame-ignore-revs
```diff
@@ -1,3 +1,6 @@
 # Conversion to new code style
 45b9887603f08523f9a663b164bd9f10b85f8b25
 9114d609168b327f00ef47b9891733cdb119b512
+
+#License update
+b6b6d463c0b5452014f46e088a50ec3aca5e5392
```
shoogle, line 6: It is not necessary or desirable to exclude the license update from the `git blame` output. The license update does not affect blame output for any lines of actual code, and excluding it simply makes Git work harder to [find a previous commit](https://www.git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt) where a nearby line was changed.

PR #7963, opened 2021-04-16
[MU4] Updated license headers for .cpp and .h files

vpereverzev:
Unfortunately, the configuration of the formatting utility does not allow you to customize something for one file type and not break for everything else. Therefore, I'm going to make separate pull requests for each file type, and besides, it should be convenient for reviewing.


build/pch/pch.h
```diff
@@ -1,21 +1,24 @@
...
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * ============================================================================= */
```
shoogle, line 21: I recommend adding a newline after the `====` to keep to the 80 character limit. The comment can be closed on the next line.

shoogle (commented):
Is there a reason for keeping the full disclaimer etc. rather than [just using the identifiers](https://github.com/musescore/MuseScore/pull/7960#discussion_r614673498)?

```
/* SPDX-License-Identifier: GPL-3.0-only
 * MuseScore-CLA-applies
 *
 * MuseScore
 * Music Composition & Notation
 *
 * Copyright (C) 2021 MuseScore BVBA and others
 */
```

If that's good enough for the Linux Kernel then it should be good enough for us.

Jojo-Schmitz:
rebase needed

igorkorsukov (approved):

PR #7981, opened 2021-04-18
Dynamically calculate width of StyledMenu

cbjeukendrup:
StyledMenu AutoWidth

For each item of the menu, the width of a temporary StyledMenuItem is calculated. The maximum of those widths is used to determine the final width.
The calculation is done in two parts: left part and right part. This means that it will look like the left version in the picture below. I think that looks cleaner, and it seems macOS itself also uses this technique. 
There is also a minimum width of 178.
Menu auto-width two options

@Tantacrul Does this makes sense?


Tantacrul:
Yeah, it does make sense!

One thing I notice here, which would be great if you could adjust: these distances I've marked in Pink appear to be more than twice as wide as they need to be. I would suggest a gutter distance of around 12px (if there are no icons to the left of the text).

Those arrows on the right can totally be moved closer to the edge. I think that was an oversight in the design.

Thanks Casper!

image

cbjeukendrup:
@Tantacrul thanks! I made the left margin 12px, and the right margin also 12px if there is a shortcut or 6px if there is a submenu indicator. To test if it's well aligned, I put some actions with a shortcut below the actions with a submenu. It looks like this:
Schermafbeelding 2021-04-18 om 23 54 18

Tantacrul:
Smokin'

RobFog:
Another rebase is necessary.

shoogle:
@cbjeukendrup, here's what we spotted on the build review earlier: the Notes menu has a large empty region on the left side due to the top item requiring space for both an icon and a checkbox.

image

It looks even more extreme when the checkbox is unchecked:

image

@Tantacrul, what do you think about ditching the checkbox column altogether and instead doing something to the icon to indicate it is checked/enabled? We could make it blue for example, like in the toolbar:

image image

Tantacrul:
> @cbjeukendrup, here's what we spotted on the build review earlier: the Notes menu has a large empty region on the left side due to the top item requiring space for both an icon and a checkbox.


I would be inclined to just remove the Note Input toggle from that context menu. I don't think it really needs to go there. It already has a representation in the UI and a shortcut too.

cbjeukendrup:
> I would be inclined to just remove the Note Input toggle from that context menu. I don't think it really needs to go there. It already has a representation in the UI and a shortcut too.

That seems a good decision to me. Users can use that action with one click directly from the toolbar, so why would they want to have it so far away in a menu too. I'll do a rebase and remove the button from the menu.

Eism (approved):

PR #7984, opened 2021-04-19
[MU4] Update .git-blame-ignore-revs to include one more rev to ignore

Jojo-Schmitz:
Namely:

* 4a0c9722961b0c1ad0c185955cb31b52e3ff5d85

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,5 @@
 # Conversion to new code style
 45b9887603f08523f9a663b164bd9f10b85f8b25
 9114d609168b327f00ef47b9891733cdb119b512
+# Change of code style
+4a0c9722961b0c1ad0c185955cb31b52e3ff5d85


vpereverzev:
изображение

It's not a code style change and I don't think we need to ignore any of these commits

Jojo-Schmitz:
Well, 4a0c9722961b0c1ad0c185955cb31b52e3ff5d85 (and a typo in my PR, will ammend)
Got the others (GPLv3) I saw a corresponding change in your initial PR.

vpereverzev:
> 
> 
> Well, [4a0c972](https://github.com/musescore/MuseScore/commit/4a0c9722961b0c1ad0c185955cb31b52e3ff5d85) (and a typo or rather cut/paste error in my PR, will amend).
> For the others (GPLv3) I saw a corresponding change in your initial PR.

Yes, it was. But we decided to not ignore them.

Regarding https://github.com/musescore/MuseScore/commit/4a0c9722961b0c1ad0c185955cb31b52e3ff5d85
I'm not against, let's include it here

Jojo-Schmitz:
I see. Done

.git-blame-ignore-revs
```diff
@@ -1,3 +1,5 @@
 # Conversion to new code style
 45b9887603f08523f9a663b164bd9f10b85f8b25
 9114d609168b327f00ef47b9891733cdb119b512
+# Change of code style
```
shoogle, line 4: I don't think we need the extra comment. Whenever we change the style, whatever we change to becomes the "new style", so the previous comment is just as applicable. If anything you could edit the first comment to read "Change of code style" and remove this one.

Jojo-Schmitz:
Well, the 1st 2 commits where the initial setup, the last is a change of that, further changes would just get appended, no new headline needed

.git-blame-ignore-revs
Jojo-Schmitz: Well, the 1st 2 commits where the initial setup, the last is a change of that, further changes would just get appended, no new headline needed

PR #7988, opened 2021-04-20
[MU4] Fix #7970: Help > About... Replace link to donate page, change text, say GPL version, say build etc.

### LINKED ISSUE #7970: [MU4 Issue] Help > About... Replace link to donate page, change text, say GPL version, say build etc.
**Describe the bug**
About MuseScore pop-up content is missed

**To Reproduce**
1. Click on "Help" (from the Upper Menu)
2. Click on 'About...'
3. Check About MuseScore pop-up links and text

**Expected behavior**
1) Replace link to donate page with link to contribute page and rephrase to "Help with MuseScore's development".
2) musescore.org link should take you to that page, not musescore.org/en-GB etc. (it should take to the user's language page)
3) Should say GPL version (soon to be GPL 3)
4) should say "development build", "nightly build" or similar (and build version)

MacOS
Windows 10

Jojo-Schmitz:
Resolves: #7970

### DIFF
--- a/src/appshell/internal/appshellconfiguration.cpp
+++ b/src/appshell/internal/appshellconfiguration.cpp
@@ -35,7 +35,7 @@ static const std::string ONLINE_HANDBOOK_URL("https://musescore.org/redirect/hel
 static const std::string ASK_FOR_HELP_URL("https://musescore.org/redirect/post/question?locale=");
 static const std::string BUG_REPORT_URL("https://musescore.org/redirect/post/bug-report?locale=");
 static const std::string LEAVE_FEEDBACK_URL("https://musescore.com/content/editor-feedback?");
-static const std::string MUSESCORE_URL("http://www.musescore.org/");
+static const std::string MUSESCORE_URL("https://www.musescore.org/");
 static const std::string MUSICXML_LICENSE_URL("https://www.w3.org/community/about/process/final/");
 static const std::string MUSICXML_LICENSE_DEED_URL("https://www.w3.org/community/about/process/fsa-deed/");
 
@@ -147,9 +147,14 @@ std::string AppShellConfiguration::museScoreUrl() const
     return MUSESCORE_URL + languageCode;
 }
 
+std::string AppShellConfiguration::museScoreForumUrl() const
+{
+    return MUSESCORE_URL + "forum";
+}
+
 std::string AppShellConfiguration::museScoreContributionUrl() const
 {
-    return museScoreUrl() + "/donate";
+    return MUSESCORE_URL + "contribute";
 }
 
 std::string AppShellConfiguration::musicXMLLicenseUrl() const
--- a/src/appshell/qml/AboutDialog.qml
+++ b/src/appshell/qml/AboutDialog.qml
@@ -113,8 +113,13 @@ QmlDialog {
 
                 StyledTextLabel {
                     Layout.fillWidth: true
-                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License version 3.")
-                    enabled: false
+                    text: {
+                        var message = qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the %1GNU General Public License version 3%2.")
+                        return message
+                        .arg("<a href='https://www.gnu.org/licenses/gpl-3.0.html'>")
+                        .arg("</a>")
+                        .replace("\n", "<br>")
+                    }
                     wrapMode: Text.WordWrap
                     maximumLineCount: 3
                 }
--- a/src/appshell/qml/AboutDialog.qml
+++ b/src/appshell/qml/AboutDialog.qml
@@ -97,14 +97,15 @@ QmlDialog {
                 StyledTextLabel {
                     Layout.fillWidth: true
                     text: {
-                        var message = qsTrc("appshell", "Visit <a href='%1'>%2</a> for new versions and more information.<br>Support MuseScore with your <a href='%3'>%4</a>.")
+                        var message = qsTrc("appshell", "Visit %1 for new versions and more information.\nGet %2 with the program or %3 to its development.")
                         var museScoreUrl = aboutModel.museScoreUrl()
+                        var museScoreForumUrl = aboutModel.museScoreForumUrl()
                         var museScoreContributionUrl = aboutModel.museScoreContributionUrl()
                         return message
-                        .arg(museScoreUrl.url)
-                        .arg(museScoreUrl.displayName)
-                        .arg(museScoreContributionUrl.url)
-                        .arg(museScoreContributionUrl.displayName)
+                        .arg("<a href='" + museScoreUrl.url + "'>" + museScoreUrl.displayName + "</a>")
+                        .arg("<a href='" + museScoreForumUrl.url + "'>" + museScoreForumUrl.displayName + "</a>")
+                        .arg("<a href='" + museScoreContributionUrl.url + "'>" + museScoreContributionUrl.displayName + "</a>")
+                        .replace("\n", "<br>")
                     }
                     wrapMode: Text.WordWrap
                     maximumLineCount: 3
@@ -112,7 +113,7 @@ QmlDialog {
 
                 StyledTextLabel {
                     Layout.fillWidth: true
-                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License.")
+                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License version 3.")
                     enabled: false
                     wrapMode: Text.WordWrap
                     maximumLineCount: 3
--- a/src/appshell/view/aboutmodel.cpp
+++ b/src/appshell/view/aboutmodel.cpp
@@ -22,6 +22,7 @@
 #include "aboutmodel.h"
 
 #include "translation.h"
+#include "version.h"
 
 #include <QClipboard>
 #include <QUrl>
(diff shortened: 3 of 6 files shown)


src/appshell/qml/AboutDialog.qml
```diff
@@ -112,8 +112,13 @@ QmlDialog {
...
                     Layout.fillWidth: true
-                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License version 3.")
-                    enabled: false
+                    text: {
+                        var message = qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.<br>Published under the %1GNU General Public License version 3%2.")
+                        return message
+                        .arg("<a href='https://www.gnu.org/licenses/gpl-3.0.html'>")
+                        .arg("</a>")
+                    }
+                    //enabled: false
```
Jojo-Schmitz, line 121: if disabled the hypelking doesn't work

src/appshell/qml/AboutDialog.qml
```diff
@@ -97,23 +97,27 @@ QmlDialog {
...
+                        .arg("<a href='" + museScoreContributionUrl.url + "'>" + museScoreContributionUrl.displayName + "</a>")
                     }
                     wrapMode: Text.WordWrap
                     maximumLineCount: 3
                 }
 
                 StyledTextLabel {
                     Layout.fillWidth: true
-                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License.")
-                    enabled: false
```
Jojo-Schmitz, line 116: If disabled the hyperlink doesn't work and the entire text is only very barely readable.

src/appshell/internal/appshellconfiguration.cpp
```diff
@@ -35,7 +35,7 @@ static const std::string ONLINE_HANDBOOK_URL("https://musescore.org/redirect/hel
 static const std::string ASK_FOR_HELP_URL("https://musescore.org/redirect/post/question?locale=");
 static const std::string BUG_REPORT_URL("https://musescore.org/redirect/post/bug-report?locale=");
 static const std::string LEAVE_FEEDBACK_URL("https://musescore.com/content/editor-feedback?");
-static const std::string MUSESCORE_URL("http://www.musescore.org/");
+static const std::string MUSESCORE_URL("https://www.musescore.org/");
```
Jojo-Schmitz, line 38: it redirects to that anyway, but better be explicit

src/appshell/internal/appshellconfiguration.cpp
```diff
@@ -147,9 +147,14 @@ std::string AppShellConfiguration::museScoreUrl() const
...
 
+std::string AppShellConfiguration::museScoreForumUrl() const
+{
+    return MUSESCORE_URL + "forum";
+}
+
 std::string AppShellConfiguration::museScoreContributionUrl() const
 {
-    return museScoreUrl() + "/donate";
+    return MUSESCORE_URL + "contribute";
```
Jojo-Schmitz, line 157: The leading slash is not needed, as `MUSESCORE_URL` has a trailing slash already. And it actually even harms, as it'd prevernt redirection to work (once that is put in place)

src/appshell/qml/AboutDialog.qml
```diff
@@ -97,23 +97,27 @@ QmlDialog {
                 StyledTextLabel {
                     Layout.fillWidth: true
                     text: {
-                        var message = qsTrc("appshell", "Visit <a href='%1'>%2</a> for new versions and more information.<br>Support MuseScore with your <a href='%3'>%4</a>.")
+                        var message = qsTrc("appshell", "Visit %1 for new versions and more information.<br>Get %2 with the program or %3 to is development.")
```
Jojo-Schmitz, line 100: Trying to avoid as much HTML code as possible, to make translations easier and more secure too.
The `<br>` though is needed here, not sure how well that works with translations, in 3.x there was a `\n` that for replaces by a `<br>` after translation did its job.

src/appshell/qml/AboutDialog.qml
```diff
@@ -97,23 +97,27 @@ QmlDialog {
...
                     wrapMode: Text.WordWrap
                     maximumLineCount: 3
                 }
 
                 StyledTextLabel {
                     Layout.fillWidth: true
-                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License.")
-                    enabled: false
+                    text: {
+                        var message = qsTrc("appshell", "Copyright &copy; 1999-2021 MuseScore BVBA and others.<br>Published under the %1GNU General Public License version 3%2.")
```
Jojo-Schmitz, line 116: See above comment reg. HTML code in translateble text

src/appshell/internal/appshellconfiguration.cpp
```diff
@@ -147,9 +147,14 @@ std::string AppShellConfiguration::museScoreUrl() const
     return MUSESCORE_URL + languageCode;
 }
 
+std::string AppShellConfiguration::museScoreForumUrl() const
+{
+    return MUSESCORE_URL + "forum";
```
Jojo-Schmitz, line 152: Or should we be using `ASK_FOR_HELP_URL` here?

Jojo-Schmitz: Ah, no, that goes directly to create a forum topic.

src/appshell/qml/AboutDialog.qml
Jojo-Schmitz: Hmm, the `&copy;` doesn't seem to work

Jojo-Schmitz: Actually the &copy; doesn't work either!?!

src/appshell/qml/AboutDialog.qml
```diff
@@ -112,8 +112,12 @@ QmlDialog {
 
                 StyledTextLabel {
                     Layout.fillWidth: true
-                    text: qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.\nPublished under the GNU General Public License version 3.")
-                    enabled: false
+                    text: {
+                        var message = qsTrc("appshell", "Copyright © 1999-2021 MuseScore BVBA and others.<br>Published under the %1GNU General Public License version 3%2.")
```
Jojo-Schmitz, line 116: See above comment reg. HTML code in translatable text

Jojo-Schmitz: For some reason we need to keep the $copy, and can't replace it with its HLML equivalent `&copy;`? But do have to replace `\n` with `<br>`, pretty strange.

Jojo-Schmitz:
This is how it now looks:
image

src/appshell/qml/AboutDialog.qml
Jojo-Schmitz: I just found the solution for the `<br>` issue. It is actually identical to what 3.x used, just in qml rather than c++.

src/appshell/qml/AboutDialog.qml
Jojo-Schmitz: I just found the solution for the `<br>` issue. It is actually identical to what 3.x used, just in qml rather than c++.

src/appshell/view/aboutmodel.cpp
```diff
@@ -36,7 +38,7 @@ AboutModel::AboutModel(QObject* parent)
 
 QString AboutModel::museScoreVersion() const
 {
-    return QString::fromStdString(configuration()->museScoreVersion());
+    return QString::fromStdString(configuration()->museScoreVersion()) + (mu::framework::Version::unstable() ? " development build" : "");
```
shoogle, line 41: We normally try to make a distinction between "development build" (PR build) and "nightly build" (master branch). Is it possible to access the BUILD_CONFIG string here? If not then I think we should just say "unstable prerelease" or something like that here.

Jojo-Schmitz: Well, we don't, never did as far as I can tell, in 3.x we used "Unstable prerelease of  3..." if and when `Version::unstable()` returned `true` . We can use that wording for master too of course, but that's not really what you originally requested?

Jojo-Schmitz: "development build" is the new name for "nighly build" (and since quite long), so the same thing really. 
I'm noit sure whether we can distinguish a PR artifact from a build of a merged PR.
`BUILD_CONFIG' would be "dev" in either case, as far as I can tell

shoogle: TLDR: Ok, let's stick with "developer build" then.

I originally requested "developer build" or "nightly build" (i.e. use the one that is most appropriate for a given build), but you are correct that the config string doesn't make this distinction anymore. Nightly still appears [in the CI scripts](https://github.com/musescore/MuseScore/blob/e927e90b00066addd3bd63d2461e012377b23222/build/ci/tools/make_build_mode_env.sh#L47) but you won't have access to it here.

Jojo-Schmitz: Should it be translatable? The old term "Unstable Prerelease for Version: %1" was translatable.
It would be actually only translated after the 4.0 translation round starts, so around a Beta.

shoogle: There's no harm in translating the string, but I don't think it really matters. It doesn't appear in release builds, and anybody using a development build would have had to navigate through the [nightly builds page](https://ftp.osuosl.org/pub/musescore-nightlies/) or one of the compilation guides, all of which are only available in English.

Jojo-Schmitz: I though so too. So will leave it as is.

Jojo-Schmitz: What the heck... why not just do it the same way as in 3.x. Or at least similar:
image
Let me know.

Jojo-Schmitz: Ah, just let's do it

shoogle (approved):

wizofaus:
Revision is still not right though...

PR #7989, opened 2021-04-20
[MU4] Added navigation for palette panel popups

igorkorsukov:
Not very convenient requires discussion about directions of navigation (later)

### DIFF
--- a/src/instruments/qml/MuseScore/Instruments/InstrumentsPanel.qml
+++ b/src/instruments/qml/MuseScore/Instruments/InstrumentsPanel.qml
@@ -242,6 +242,24 @@ Item {
                                 onVisibleChanged: {
                                     treeItemDelegateLoader.updateIsSelected()
                                 }
+
+                                property var contentYBackup: 0
+
+                                onPopupOpened: {
+                                    contentYBackup = flickable.contentY
+                                    var mappedPopupY = mapToItem(flickable, popupX, popupY).y
+
+                                    if (mappedPopupY + popupHeight < flickable.height - contentColumn.sideMargin) {
+                                        return
+                                    }
+
+                                    var hiddenPopupPartHeight = Math.abs(flickable.height - (mappedPopupY + popupHeight))
+                                    flickable.contentY += hiddenPopupPartHeight + contentColumn.sideMargin
+                                }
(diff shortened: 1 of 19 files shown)


src/instruments/qml/MuseScore/Instruments/InstrumentsPanel.qml
```diff
@@ -242,6 +242,24 @@ Item {
                                 onVisibleChanged: {
                                     treeItemDelegateLoader.updateIsSelected()
                                 }
+
+                                property var contentYBackup: 0
+
+                                onPopupOpened: {
```
igorkorsukov, line 248: Accidentally removed by prev commit, now restored

src/palette/qml/MuseScore/Palette/internal/MoreElementsPopup.qml
```diff
@@ -53,13 +53,21 @@ StyledPopup {
     property bool drawGrid
 
     property int maxHeight: 400
-    implicitHeight: column.implicitHeight + topPadding + bottomPadding
-    width: parent.width
+    contentHeight: column.implicitHeight// + topPadding + bottomPadding
```
vpereverzev, line 56: any reason of this being commented?

igorkorsukov: removed

src/palette/qml/MuseScore/Palette/internal/TreePaletteHeader.qml
```diff
@@ -147,78 +137,39 @@ Item {
         acceptedButtons: Qt.RightButton
 
         onClicked: {
-            toggleContextMenu()
+            menuLoader.toggleContextMenu(null)
```
vpereverzev, line 140: do we really need to call it if navigationParentControl is null? I'm not quite sure what would we expect in this case

igorkorsukov: this is clicked on RightButton, does not participate in navigation 
I generally think this action is not necessary

vpereverzev (approved):

PR #7993, opened 2021-04-22
Use QStyle to make QWidgets look like QML

cbjeukendrup:
Use QStyle to make QWidgets look like QML. 

I made UiTheme a subclass of QProxyStyle, override some methods to make QWidgets look like our QML components, and set it as the style for the app.

Why using UiTheme for this? This turned out to be very convenient because of the easy access to colors and theme information. 

I did not style _all_ widgets, just the most used ones, because this is of course just a temporary solution until we have time to rewrite all those dialogs. Of course anyone may feel free to style the remaining controls. 

QStyle could be interesting for some other things too; for example in docktoolbar.cpp I replaced most of the usage of stylesheets with QStyle. I think this solution is much better than those hardcoded SVGs for the grip icons. However, for an unclear reason, making the grips hoverable still won't work, even after setting `WA_Hover` or `mouseTracking`.

(There is still other work to do with the look of dialogs... they deserve some tidying.)

Here are two screenshots:
Style dialog and Breaks dialog
Toolbars and menus

Status:
- [x] Tested on macOS
- [x] Tested on Windows
- [x] Tested on Linux (Ubuntu)

### DIFF
--- a/src/framework/ui/view/uitheme.cpp
+++ b/src/framework/ui/view/uitheme.cpp
@@ -357,6 +357,9 @@ void UiTheme::setupWidgetTheme()
     QColor fontPrimaryColorDisabled = fontPrimaryColor();
     fontPrimaryColorDisabled.setAlphaF(itemOpacityDisabled());
 
+    QColor linkColorDisabled = linkColor();
+    linkColorDisabled.setAlphaF(itemOpacityDisabled());
+
     QColor backgroundPrimaryColorDisabled = backgroundPrimaryColor();
     backgroundPrimaryColorDisabled.setAlphaF(itemOpacityDisabled());
 
@@ -376,6 +379,9 @@ void UiTheme::setupWidgetTheme()
 
     palette.setColor(QPalette::Text, fontPrimaryColor());
     palette.setColor(QPalette::Disabled, QPalette::Text, fontPrimaryColorDisabled);
+
+    palette.setColor(QPalette::Link, linkColor());
+    palette.setColor(QPalette::Disabled, QPalette::Link, linkColorDisabled);
 
     palette.setColor(QPalette::Button, backgroundSecondaryColor());
     palette.setColor(QPalette::Disabled, QPalette::Button, backgroundSecondaryColorDisabled);
--- a/src/framework/ui/view/uitheme.cpp
+++ b/src/framework/ui/view/uitheme.cpp
@@ -448,163 +448,67 @@ void UiTheme::drawPrimitive(QStyle::PrimitiveElement element, const QStyleOption
 
     //! NOTE This drawing code is based on the implementation in QML.
     switch (element) {
+    // Buttons (and ComboBoxes)
     case QStyle::PE_PanelButtonCommand: {
-        const QStyleOptionButton* buttonOption = qstyleoption_cast<const QStyleOptionButton*>(option);
+        auto buttonOption = qstyleoption_cast<const QStyleOptionButton*>(option);
         const bool accentButton = buttonOption && buttonOption->features & QStyleOptionButton::DefaultButton;
         const bool flat = buttonOption && buttonOption->features & QStyleOptionButton::Flat;
 
-        QColor backgroundColor(accentButton ? accentColor()
-                               : flat ? Qt::transparent
-                               : buttonColor());
-
-        backgroundColor.setAlphaF(!enabled ? buttonOpacityNormal() * itemOpacityDisabled()
-                                  : pressed ? buttonOpacityHit()
-                                  : hovered ? buttonOpacityHover()
-                                  : !flat ? buttonOpacityNormal()
-                                  : 0);
-
-        drawRoundedRect(painter, option->rect, 3, backgroundColor);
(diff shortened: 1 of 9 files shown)


cbjeukendrup:
@RomanPudashkin It's quite some time ago, but what was the problem you were solving with commit 83d07651749bc248a9cd448a66d3c9d6c3ac689b? On my Windows and Linux machines, it makes all text in Widgets much bigger than in QML. Without that commit, the texts are exactly as big as in QML. This is the case on multiple screen resolutions. 
On macOS, there seems to be no problem anyway.

RomanPudashkin:
> @RomanPudashkin It's quite some time ago, but what was the problem you were solving with commit [83d0765](https://github.com/musescore/MuseScore/commit/83d07651749bc248a9cd448a66d3c9d6c3ac689b)? On my Windows and Linux machines, it makes all text in Widgets much bigger than in QML. Without that commit, the texts are exactly as big as in QML. This is the case on multiple screen resolutions.
> On macOS, there seems to be no problem anyway.

Fonts were too small in all widgets on windows/linux. You can revert this fix if the problem doens't reproduce anymore because it was a temporary solution anyway :)

cbjeukendrup:
> Fonts were too small in all widgets on windows/linux. You can revert this fix if the problem doens't reproduce anymore because it was a temporary solution anyway :)

Thanks :)
I made another change, and now it looks good, and also the existing widgets get updated when the font is changed in Preferences. 
On Windows, it now looks like the screenshot from macOS I posted above.

src/framework/ui/view/uitheme.cpp
```diff
@@ -340,14 +397,371 @@ void UiTheme::setupWidgetTheme()
...
+    QProxyStyle::unpolish(widget);
+
+    if (qobject_cast<QAbstractItemView*>(widget)
+        || qobject_cast<QGroupBox*>(widget)) {
+        widget->setMouseTracking(false);
+        widget->setAttribute(Qt::WA_Hover, false);
+    }
+}
+
+void UiTheme::drawPrimitive(QStyle::PrimitiveElement element, const QStyleOption* option, QPainter* painter, const QWidget* widget) const
```
RomanPudashkin, line 440: This method seems over-complicated for me. I think you can divide it into several small methods

cbjeukendrup: Thanks, done!

cbjeukendrup:
What is going on with the UTests?

src/framework/ui/view/uitheme.cpp
```diff
@@ -22,23 +22,74 @@
...
 #include <QApplication>
+#include <QMenu>
 #include <QPalette>
+#include <QStyleFactory>
+#include <QStyleOption>
+#include <QToolBar>
 #include <QVariant>
 
+#include "log.h"
+#include "notation/view/widgets/radiobuttongroupbox.h"
```
Eism, line 35: Please, remove the dependence of the framework on the notation

vshalkevich:
CI_utests / run_tests (pull_request) Failing after 3m — run_tests 
Why is it failed ?

RomanPudashkin:
> CI_utests / run_tests (pull_request) Failing after 3m — run_tests
> Why is it failed ?

Probably due to the problem that has been mentioned by Elnur

src/framework/ui/view/uitheme.cpp
cbjeukendrup: That would indeed solve the linking problem in the UTests. But it is needed to check whether we have a normal group box or a RadioButtonGroupBox, because they require different metrics. But maybe I can find a different way to do that.
We could also move RadioButtonGroupBox from the notation module to something else, like uicomponents. In fact it has nothing to do with notation.

RomanPudashkin: >We could also move RadioButtonGroupBox from the notation module to something else, like uicomponents. In fact it has nothing to do with notation.

Yes, probably this is the best solution at the moment

cbjeukendrup: Found another solution. Do you like it? If not, I will just do the first idea (moving RadioButtonGroupBox).

RomanPudashkin: >Do you like it?

No :) Your last solution is less clear than the idea to move RadioButtonGroupBox to uicomponents. I don't like that we will add a dependency in the ui module on uicomponents, but at least this is a more clear solution

cbjeukendrup: Done! 
Oh, but now the test is failing again... 
I also thought of a third solution, namely checking the class name from the metaObject of the QGroupBox. Like this:
```cpp
groupBox->metaObject()->className();
```
That should work without any dependencies on RadioButtonGroupBox.

cbjeukendrup: I applied that third solution now, I guess it is the only way.

RomanPudashkin (approved):

cbjeukendrup:
Oops, that last commit is incomplete. It does not fix the whole problem. One more iteration...

Issue #7881, opened 2021-04-03
[MU4 Issue] Can't copy or paste text to or from another program


vshalkevich:
**Describe the bug**
Can't copy or paste text to or from another program

**To Reproduce**
1. Create a score with title  
2. Copy text from another app
3. Double click on score title to change it
4. Try to paste copied text from another program

**Expected behavior**
User can copy and paste text in MuseScore from another apps

**Screenshots**

**Desktop (please complete the following information):**
Linux Ubunty

**Additional context**

wizofaus:
Should be fixed by https://github.com/musescore/MuseScore/pull/8306

PR #7999, opened 2021-04-24
[MU4] Implement copy, paste, cut in TextBase; fix escape in TextBase

aaaeka:
Resolves: #7881 

Implements copy, paste and cut in text boxes. Esc now should also work as intended in text boxes.


wizofaus:
This looks like a duplicate of https://github.com/musescore/MuseScore/pull/8306 - you've basically done more or less what I did but I'm pretty sure yours will have some refresh/undo stack issues that I encountered along the way.
Also I implemented copy/paste for formatted text, which was something sorely missing in MU3.

aaaeka:
Your solution looks like a more robust one than mine so I will close this PR.

PR #8002, opened 2021-04-25
Implement Main Window Title

cbjeukendrup:
Currently, I implemented the following behaviour: 
When no score is opened, the title is just "MuseScore 4".
When a score is opened, the title will be the filename of the score. 

@Tantacrul Could you tell me what the behaviour should be exactly?

When the current score is saved, the `windowFilePath` will be set, which has some benefits on macOS. Additionally, when the score is modified, that black dot in the close button on macOS is shown, by setting `windowModified`.

Schermafbeelding 2021-04-25 om 14 03 07
Schermafbeelding 2021-04-25 om 14 07 55

Note: because currently `currentMasterNotation` is not always correctly set, the title is not updated when a score is closed.


Tantacrul:
That's cool! Nice catch.

I would add (if it doesn't do this already) that it reflects the title of the **currently active** project. I know multiple projects are still not working in MS4 yet but that would be the desired behaviour when they are.

cbjeukendrup:
It is indeed the title of the active project - probably, but I can't really test it...
When later turns out that it is wrong, that's because `context()->currentMasterNotation()` is not correctly set. 

(I hope multi-project support gets implemented soon. And we need to come up with a good solution for split view (which I think is very useful, but about which I haven't seen anything in MuseScore 4). SplitView sometimes caused difficulties in MuseScore 3, because there is a score that _is_ visible, but not the _active_ score.)

Tantacrul:
We're not going to use that vertical stacking tab thing from 3. Now that we have the ability to close and reopen parts, it's not necessary anymore. There's a design for the new system, which is pretty straight forward.

RomanPudashkin:
This solution is great! We need to adapt it for the new dock system

cbjeukendrup:
> This solution is great! We need to adapt it for the new dock system

Thanks! I started doing that when the KDAB stuff was merged, but I encountered a problem: with the new solution, a QWindow is used instead of the QWidget-based QMainWindow. Bizarrely, QWindow has no API to set the window "modified". That can only be done with a QWidget. I looked at the source code for QWidget, and it turns out QWidget accesses the QPlatformWindow of its QWindow. We can also access the QPlatformWindow of the QML window (I have a trick for getting its QWindow), but we cannot include qplatformwindow.h (it's private), so we can't call the necessary method of the QPlatformWindow. 

But I can continue without the `windowModified` stuff (which only has an effect on macOS anyway), so that we have at least the window title and filepath.

cbjeukendrup:
@RomanPudashkin I updated this solution, _with_ the `fileModified` stuff! I just went around Qt and used the macOS APIs directly. 

In case you would like to compare this with the old solution, it's here: https://github.com/musescore/MuseScore/commit/ecafd7b9e7f191ce30c8b3da800041fe1daa3624

Needs some testing on Windows and Linux.

cbjeukendrup:
Rebased & resolved merge conflicts.

RomanPudashkin (approved):

RomanPudashkin:
@vshalkevich please check this on windows and linux

vshalkevich:
As I see this modifications are for MacOS. On Linux and Windows I can`t see them but application works fine on Windows 10 and Linux.

PR #8004, opened 2021-04-25
Concept: Dynamically calculate (minimum) size of NoteInputBar

cbjeukendrup:
This PR introduces an algorithm for automatically calculating the minimum width and height of the NoteInputBar. 

Schermafbeelding 2021-04-25 om 18 31 09

However, in practice this turns out to be useless:
- When the minimum size of the toolbar is bigger than what fits in the window, it is not possible to re-dock the toolbar;
- Sometimes, when dragging the toolbar around, the app randomly crashes, seemingly by the fault of Qt;
- When moving from the horizontal toolbar area to the vertical one, only the height of the toolbar is recalculated, and the width stays the same, causing it to look like this:
  Schermafbeelding 2021-04-25 om 18 39 13 kopie
  Also happens when going from vertical to horizontal. The reason for this is unclear to me, maybe also a Qt thing. Might be related to the second issue. 

I still create the PR, hoping that somebody can still use the algorithm at some point.

A counterpart could be: calculate how many items fit in a given width and how many should be hidden in a '...' menu.

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -114,18 +114,27 @@ DockPage {
         id: notationNoteInputBar
         objectName: "notationNoteInputBar"
 
-        minimumWidth: orientation == Qt.Horizontal ? 900 : 96
-        minimumHeight: orientation == Qt.Horizontal ? 48 : 0
+        property int idealWidth: 0
+        property int idealHeight: 0
+
+        minimumWidth: orientation == Qt.Horizontal ? idealWidth : 90
+        minimumHeight: orientation == Qt.Horizontal ? 48 : idealHeight
 
         color: notationPage.color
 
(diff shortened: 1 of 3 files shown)


vpereverzev:
@cbjeukendrup I don't want to waste your time on the old dock-system, because at the moment we're about to merge the new one (based on KDAB dock widgets)

Also there is a note for the future PRs - try to avoid putting such logic in QML, general rule - keep QML as simple as possible and pull all the "business" logic into C++ models and services

cbjeukendrup:
@vpereverzev I don't really feel like it's wasted time, because most of the logic can also be used with the new docking system -- even better than with the old system, hopefully. 

About your second point, I generally try to do this. However, in this particular case, I doubt whether it is efficient to create a separate C++ model for calculating the size of a GridView, then pass all the necessary data into that model and pass the result back. Feels a bit cumbersome to me. But what would you say about this case?

PR #8007, opened 2021-04-26
[MU4] KDAB dockwidgets integration. Docks

RomanPudashkin:
- added all types of docks based on KDAB dockwidgets

### DIFF
--- a/src/appshell/view/dockwindow_kdab/dockpage.cpp
+++ b/src/appshell/view/dockwindow_kdab/dockpage.cpp
@@ -1,0 +1,139 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "dockpage.h"
+
+#include "docktoolbar.h"
+#include "dockcentral.h"
+#include "dockpanel.h"
+#include "dockstatusbar.h"
+
+#include "log.h"
+
+using namespace mu::dock;
+
+DockPage::DockPage(QQuickItem* parent)
+    : QQuickItem(parent),
+    m_mainToolBars(this),
+    m_toolBars(this),
+    m_panels(this),
+    m_statusBars(this)
+{
+}
+
+QString DockPage::uri() const
+{
+    return m_uri;
+}
+
+QQmlListProperty<DockToolBar> DockPage::mainToolBarsProperty()
+{
+    return m_mainToolBars.property();
+}
+
+QQmlListProperty<DockToolBar> DockPage::toolBarsProperty()
+{
+    return m_toolBars.property();
+}
+
+QQmlListProperty<DockPanel> DockPage::panelsProperty()
+{
+    return m_panels.property();
+}
+
+QQmlListProperty<DockStatusBar> DockPage::statusBarsProperty()
+{
+    return m_statusBars.property();
+}
+
+QList<DockToolBar*> DockPage::mainToolBars() const
+{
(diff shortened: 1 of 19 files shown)


src/appshell/view/dockwindow_kdab/dockpage.cpp
```diff
@@ -0,0 +1,153 @@
...
+    : QQuickItem(parent),
+    m_toolBars(this),
+    m_panels(this),
+    m_statusBars(this)
+{
+}
+
+QString DockPage::uniqueName() const
+{
+    return m_uniqueName;
```
igorkorsukov, line 45: I think it better to use `objectName`

src/appshell/view/dockwindow_kdab/dockpage.cpp
```diff
@@ -0,0 +1,153 @@
...
+    window.addDock(m_central, KDDockWidgets::Location_OnRight);
+
+    for (DockPanel* panel : m_panels.list()) {
+        window.addDock(panel, KDDockWidgets::Location_OnLeft);
+    }
+
+    DockToolBar* prevToolBar = nullptr;
+    for (DockToolBar* toolBar : m_toolBars.list()) {
+        auto location = prevToolBar ? KDDockWidgets::Location_OnRight : KDDockWidgets::Location_OnTop;
+        window.addDock(toolBar, location, prevToolBar);
```
igorkorsukov, line 116: Why is there such a location? I believe we need an explanation (comment)
And it seems that location should be a property of the toolbar

src/appshell/view/dockwindow_kdab/dockpage.cpp
```diff
@@ -0,0 +1,153 @@
...
+        prevToolBar = toolBar;
+    }
+
+    DockStatusBar* prevStatusBar = nullptr;
+    for (DockStatusBar* statusBar : m_statusBars.list()) {
+        auto location = prevStatusBar ? KDDockWidgets::Location_OnRight : KDDockWidgets::Location_OnBottom;
+        window.addDock(statusBar, location, prevStatusBar);
```
igorkorsukov, line 123: Why is there such a location? I believe we need an explanation (comment)
And it seems that location should be a property of the toolbar

src/appshell/view/dockwindow_kdab/docktoolbar.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+Qt::Orientation DockToolBar::orientation() const
+{
+    return m_orientation;
+}
+
+QStringList DockToolBar::allowedPagesUriList() const
+{
+    return m_allowedPagesUriList;
```
igorkorsukov, line 51: What is it for?

src/appshell/view/dockwindow_kdab/dockwindow.cpp
```diff
@@ -0,0 +1,136 @@
...
+    newPage->init(*this);
+
+    DockToolBar* prevToolBar = nullptr;
+    for (DockToolBar* toolBar : m_toolBars.list()) {
+        auto location = prevToolBar ? KDDockWidgets::Location_OnRight : KDDockWidgets::Location_OnTop;
+        addDock(toolBar, location, prevToolBar);
+
+        if (!toolBar->allowedPagesUriList().empty() && !toolBar->allowedPagesUriList().contains(uri)) {
+            toolBar->close();
+        }
```
igorkorsukov, line 107: Confusing.
As far as I can see from the code (and as it was before), we have:
1. Toolbars of the window - always shown
2. Page toolbars - shown if the page is shown

It is strange to see the window toolbar that needs to be hidden for the given pages. Maybe then such a toolbar can simply be move to the page?

src/appshell/view/dockwindow_kdab/internal/dockbase.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+QString DockBase::title() const
+{
+    return m_title;
+}
+
+QString DockBase::uniqueName() const
+{
+    return m_uniqueName;
+}
```
igorkorsukov, line 46: Better to use `objectName`

src/appshell/view/dockwindow_kdab/dockpage.cpp
RomanPudashkin: fixed

src/appshell/view/dockwindow_kdab/dockpage.cpp
RomanPudashkin: added a comment

src/appshell/view/dockwindow_kdab/docktoolbar.cpp
RomanPudashkin: removed

src/appshell/view/dockwindow_kdab/dockwindow.cpp
RomanPudashkin: fixed

igorkorsukov (approved):

PR #8011, opened 2021-04-27
[MU3] Fix #320517: Voice 2 note stems face wrong way when noteheads made invisible

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/320517

PR for master: #8012 and #8967 (fixing a regression of #8012) (both merged already)

### DIFF
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -60,6 +60,7 @@
 #include "staff.h"
 #include "stafftext.h"
 #include "stafftype.h"
+#include "stem.h"
 #include "stringdata.h"
 #include "style.h"
 #include "sym.h"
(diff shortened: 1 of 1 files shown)


libmscore/measure.cpp
```diff
@@ -2802,6 +2802,9 @@ bool Measure::hasVoices(int staffIdx, Fraction stick, Fraction len) const
                                           break;
                                           }
                                     }
+                              // same if stem or hook(s) are visible
+                              if ((c->stem() || c->hook()) && c->visible())
+                                    v = true;
```
Jojo-Schmitz, line 2807: Maybe moving this to above the `notes()` check and also add a `break;`, to handle the easy and most probable cases first?
Or would that be considered 'micro-optimization'?

Jojo-Schmitz:
Unfortunatly this seems to cause a regession: a (visible) voice 1 is all up-stem in presence of an invisible voice 2!

Jojo-Schmitz:
See #9000

PR #8012, opened 2021-04-27
[MU4] Fix #320517: Voice 2 note stems face wrong way when noteheads made invisible

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/320517

Counterpart to #8011, but for master


Jojo-Schmitz:
Unfortunatly this seems to cause a regession: a (visible) voice 1 is all up-stem in presence of an invisible voice 2!

PR #8013, opened 2021-04-27
[MU4] Added navigation for Inspector Playback Popup

igorkorsukov:


src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopupView.qml
```diff
@@ -123,8 +131,11 @@ PopupView {
                 id: contentBody
                 x: root.margins
                 y: root.margins
-                width: childrenRect.width
-                height: childrenRect.height
+                width: root.contentWidth
```
vpereverzev, line 134: should be enough to set only implicit size
if width or height are not set then implicit size would be used by default

vpereverzev (approved):

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopupView.qml
igorkorsukov: Explicit is always better than not explicit :)

Issue #7686, opened 2021-03-06
[MU4 Issue] Cannot import non-native file formats (MusicXML, MIDI, etc.)


vshalkevich:
**Describe the bug**
Can open only MSCX and MSCZ formats
Other music formats can not be opened (like MusicXML etc.)

**To Reproduce**
1. Click "File" from the upper Menu Bar
2. Click "Open"
3. Select MusicXML format file

**Expected behavior**
MusicXML and other supported in MU3 formats should be opened

**Desktop (please complete the following information):**
MacOS

vshalkevich:
![2021-06-23 17_02_55-MuseScore 4](https://user-images.githubusercontent.com/71372310/123122712-ec2f6a00-d44e-11eb-9bd9-da365ab41751.jpg)

Can open MusicXML but error message appears

PR #8014, opened 2021-04-27
[MU4] Fix #316504: Update MusicXML Import of Fretboard Diagrams

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/316504

Port of #7455 to master

### DIFF
--- a/src/libmscore/fret.cpp
+++ b/src/libmscore/fret.cpp
@@ -294,7 +294,7 @@ void FretDiagram::init(StringData* stringData, Chord* chord)
                 setDot(string, fret);
             }
         }
-        _maxFrets = stringData->frets();
+        _frets = stringData->frets();
     } else {
         _maxFrets = 6;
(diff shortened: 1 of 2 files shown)


src/libmscore/fret.cpp
```diff
@@ -1333,6 +1333,7 @@ void FretDiagram::writeMusicXML(XmlWriter& xml) const
             xml.tag("fret", "0");
             xml.etag();
         } else {
```
worldwideweary, line 1335: Careful, I explicitly removed this else statement for a reason in original PR :)

Jojo-Schmitz: oops

Jojo-Schmitz: Better now?

Jojo-Schmitz:
Currently no way to really test it, due to issue #7686

PR #8015, opened 2021-04-27
[MU4] Added navigation for AppearancePopup in the Inspector

igorkorsukov:


src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopupView.qml
```diff
@@ -39,7 +39,7 @@ PopupView {
     property int padding: 16
     property int margins: 16
 
-    property int contentWidth: contentBody.childrenRect.width
+    property int contentWidth: 240
```
vpereverzev, line 42: Curious about this change, why this value is strictly defined now

src/inspector/view/qml/MuseScore/Inspector/general/appearance/internal/VerticalSpacingSection.qml
```diff
@@ -19,27 +19,38 @@
...
+import MuseScore.Inspector 1.0
 import "../../../common"
 
 InspectorPropertyView {
     id: root
 
     property QtObject minimumDistance: undefined
 
+    property NavigationPanel navigationPanel: null
+    property int navigationRowOffset: 0
```
vpereverzev, line 34: how do we set navigationRowOffset? Is it always 0 or not?

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopupView.qml
igorkorsukov: Previously, I set this value in each popup, that is, it was duplicated in many places.
Now I decided to move it to the base class so as not to duplicate it.

src/inspector/view/qml/MuseScore/Inspector/general/appearance/internal/VerticalSpacingSection.qml
igorkorsukov: yes, we setting it 
https://github.com/musescore/MuseScore/pull/8015/files#diff-d21e999f7e896c07228450ed7a42657cf9623c69aebe4309bab921747c5de1a0R63

vpereverzev (approved):

Issue #7969, opened 2021-04-17
[MU4 Issue] Changes in the style dialog should update live in the score


vshalkevich:
**Describe the bug**
Text style doesn't update live in the score

**To Reproduce**
1. Create a score
2. Add text to the score
3. Open style dialog (Upper Menu: Format -> Style...)
4. Change text style 

**Expected behavior**
Changes in the style dialog should update live in the score

**Desktop (please complete the following information):**
MacOS

PR #8031, opened 2021-04-29
[MU4] Update notation styles live when changing them in style dialog

aaaeka:
Resolves: #7969 

A very simple fix.

### DIFF
--- a/src/notation/internal/notationstyle.cpp
+++ b/src/notation/internal/notationstyle.cpp
@@ -52,6 +52,7 @@ void NotationStyle::setStyleValue(const StyleId& styleId, const QVariant& newVal
         m_getScore->score()->cmdConcertPitchChanged(newValue.toBool());
     } else {
         m_getScore->score()->undoChangeStyleVal(styleId, newValue);
+        m_getScore->score()->update();
     }
 
     m_styleChanged.notify();
--- a/src/notation/view/widgets/editstyle.cpp
+++ b/src/notation/view/widgets/editstyle.cpp
@@ -1214,6 +1214,8 @@ void EditStyle::buttonClicked(QAbstractButton* b)
         }
         break;
     }
+
+    globalContext()->currentNotation()->style()->styleChanged().notify();
 }
 
 //---------------------------------------------------------


cbjeukendrup:
But isn't the issue that the notation should update directly when making changes and not only when clicking OK?

aaaeka:
> But isn't the issue that the notation should update directly when making changes and not only when clicking OK?

True, will fix.

vshalkevich:
Style applies only when I click OK button. 
It should be updated when I select another style at the same time (before user click OK)

cbjeukendrup:
Maybe it's as simple as copying or moving the line you added, to `setStyleValue`.

aaaeka:
I was searching on and off to find a way to fix this problem for a week now. I think I already tried the solution you suggested, but I can't remember for sure so I'll try it again.

aaaeka:
Turns out it was simply because ```score()->update()``` was not called.

cbjeukendrup:
Ah, nice spot! It works perfectly now!

cbjeukendrup (approved):

vshalkevich:
Checked. Works good

vshalkevich:
Also works fine on Windows 10 without this changes

vshalkevich:
On Windows 10 feature works fine without fixes (without this PR)

cbjeukendrup:
> On Windows 10 feature works fine without fixes (without this PR)

Are you sure? For me it does not, on Windows 10.

Eism:
> > On Windows 10 feature works fine without fixes (without this PR)
> 
> Are you sure? For me it does not, on Windows 10.

it works on Windows 10 for me too

vshalkevich:
> 
> 
> > On Windows 10 feature works fine without fixes (without this PR)
> 
> Are you sure? For me it does not, on Windows 10.

Yes, strange. 
@aaaeka What this code do? I can`t understand why it works withour this PR in some cases and doesn`t work for others?

cbjeukendrup:
I mean, _with_ the PR it works perfectly for me on all platforms, but _without_ the PR it does not work, on Windows 10 just like on macOS.
So I am convinced this PR is okay and is needed.

Eism:
> I mean, _with_ the PR it works perfectly for me on all platforms, but _without_ the PR it does not work, on Windows 10 just like on macOS.
> So I am convinced this PR is okay and is needed.

it is necessary to understand the problem. Please answer these questions
1. steps to reproduce the bug
2. Is the bug reproduced on the latest master?

cbjeukendrup:
On the latest master, the following happens:
1. Create a score
2. Open the style dialog
3. Change (for example) the musical font from Leland to Emmentaler
4. You see _some_ things updating in the score view, but most things not.
5. When you press OK and drag the notation (to trigger redrawing), everything updates.

With this PR, number 4 is like this:
4. As soon as you make any change in the style dialog, you see the whole score updating correctly.

Eism:
> 4. You see some things updating in the score view, but most things not.
can you give an example of non-updatable elements?

cbjeukendrup:
This video shows what happens when changing the font. In fact, only the brackets are changing somewhat. 
When I click OK and move the notation a bit, you see the font changing (the G-clef is the most obvious). 

https://user-images.githubusercontent.com/48658420/118023912-5e943100-b35e-11eb-9ed6-228accecae6f.mov

RomanPudashkin (approved):

PR #8034, opened 2021-04-30
[MU3] Fix #303422: Importing colored lyrics from MusicXML

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/303422

Up to 3.5 it worked only for first line (number="1"), in 3.6 and master it stopped working all together

This is for 3.x what #8033 is for master, as on master MusicXML import can't get tested currently...

Image of importing the test file, `mtest/musicxml/io/testLyricColor.xml`:
image


Jojo-Schmitz:
@iveshenry18 this one might be something for your 3.6.2_backend fixes too

PR #8036, opened 2021-04-30
[MU3] Fix #320573: [MusicXML import] fretboard frets property not saved in MuseScore format

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/320573

PR for master in #8037 (but there MusicXML import is currently not working)


vpereverzev:
3.x merges are closed

PR #8045, opened 2021-05-04
[MU4] Reimplementation qml dialogs base class 

igorkorsukov:


vpereverzev (approved):

src/inspector/view/qml/MuseScore/Inspector/general/GeneralInspectorView.qml
```diff
@@ -143,7 +143,7 @@ InspectorSectionView {
 
                 PlaybackPopup {
                     id: playbackPopup
-                    navigation.parentControl: playbackButton.navigation
+                    navigationParentControl: playbackButton.navigation
```
Eism, line 146: Is it necessary to set parental control here? Because you set the previous control as a parent for the popup by default in PopupView::open

igorkorsukov: You right, not necessary, but I prefer to set explicit.

Eism (approved):

PR #8047, opened 2021-05-05
[MU4] Added navigation for NewScore Dialog (first screen)

igorkorsukov:
Now I implemented it stupidly, I think later we can do it more conveniently.

https://user-images.githubusercontent.com/3818029/117138217-ffa54980-adaa-11eb-8e90-f7e67473a2be.mp4

### DIFF
--- a/src/framework/ui/view/navigationsection.h
+++ b/src/framework/ui/view/navigationsection.h
@@ -36,12 +36,23 @@ namespace mu::ui {
 class NavigationSection : public AbstractNavigation, public INavigationSection, public async::Asyncable
 {
     Q_OBJECT
+    Q_PROPERTY(QmlType type READ type_property WRITE setType NOTIFY typeChanged)
+
     INJECT(ui, INavigationController, navigationController)
 
 public:
     explicit NavigationSection(QObject* parent = nullptr);
     ~NavigationSection() override;
 
+    //! NOTE Please sync with INavigationSection::Type
+    enum QmlType {
+        Regular = 0,
+        Exclusive
+    };
+    Q_ENUM(QmlType)
+
+    QmlType type_property() const;
+    INavigationSection::Type type() const override;
     QString name() const override;
 
     const Index& index() const override;
@@ -66,11 +77,18 @@ public:
     void addPanel(NavigationPanel* panel);
     void removePanel(NavigationPanel* panel);
 
+public slots:
+    void setType(QmlType type);
+
+signals:
+    void typeChanged(QmlType type);
+
 private:
 
     std::set<INavigationPanel*> m_panels;
     async::Notification m_panelsListChanged;
     async::Channel<SectionPanelControl> m_forceActiveRequested;
+    QmlType m_type = Regular;
 };
 }
 
--- a/src/instruments/qml/MuseScore/Instruments/internal/FamilyView.qml
+++ b/src/instruments/qml/MuseScore/Instruments/internal/FamilyView.qml
@@ -33,17 +33,19 @@ Item {
     property alias families: familiesBox.model
     property var groups: null
 
+    property alias navigation: navPanel
+
     signal familySelected(string familyId)
     signal groupSelected(string groupId)
 
     QtObject {
-        id: privateProperties
+        id: prv
 
         property int currentGroupIndex: -1
     }
 
     function clearSelection() {
-        privateProperties.currentGroupIndex = -1
+        prv.currentGroupIndex = -1
     }
 
     function setFamily(familyId) {
@@ -53,11 +55,18 @@ Item {
     function focusGroup(groupId) {
         for (var i in root.groups) {
             if (root.groups[i].id === groupId) {
-                privateProperties.currentGroupIndex = i
+                prv.currentGroupIndex = i
                 groupsView.positionViewAtIndex(groupsView.currentGroupIndex, ListView.Beginning)
                 return
             }
         }
+    }
+
+    NavigationPanel {
+        id: navPanel
+        name: "FamilyView"
+        direction: NavigationPanel.Vertical
+        enabled: root.visible
     }
 
     StyledTextLabel {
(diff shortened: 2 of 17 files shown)


src/instruments/qml/MuseScore/Instruments/internal/FamilyView.qml
```diff
@@ -119,7 +140,8 @@ Item {
             }
 
             onClicked: {
-                privateProperties.currentGroupIndex = index
+                console.log("onClicked")
```
RomanPudashkin, line 143: Please, remove this log

src/userscores/qml/MuseScore/UserScores/NewScoreDialog.qml
```diff
@@ -89,10 +88,23 @@ StyledDialogView {
             readonly property int buttonHeight: 30
             readonly property int buttonWidth: 132
 
+            NavigationPanel {
+                id: navBottomPanel
+
+                name: "BottomPanel"
+                section: root.navigationSection
+                order: 100
```
RomanPudashkin, line 96: Please, explain what this number means

src/userscores/qml/MuseScore/UserScores/internal/TitleListView.qml
```diff
@@ -66,8 +79,15 @@ Item {
         currentIndex: 0
 
         delegate: ListItemBlank {
+            id: item
```
RomanPudashkin, line 82: It seems that you don't need explicitly set id for delegates

src/userscores/qml/MuseScore/UserScores/internal/TitleListView.qml
```diff
@@ -19,8 +19,8 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-import QtQuick 2.7
-
+import QtQuick 2.15
```
RomanPudashkin, line 22: Please, add an empty line after Qt import

src/userscores/qml/MuseScore/UserScores/internal/ChooseTemplatePage.qml
```diff
@@ -53,6 +58,10 @@ Item {
             Layout.fillHeight: true
             Layout.preferredWidth: parent.width / 4
 
+            navigation.section: root.navigationSection
+            navigation.name: "Category"
+            navigation.order: 2
```
RomanPudashkin, line 63: Why does it start from 2 and not from 1?

src/instruments/qml/MuseScore/Instruments/internal/FamilyView.qml
```diff
@@ -107,7 +120,15 @@ Item {
...
 
         delegate: ListItemBlank {
-            isSelected: privateProperties.currentGroupIndex === index
+            id: item
+
+            isSelected: prv.currentGroupIndex === index
+
+            navigation.name: modelData.name
+            navigation.panel: navPanel
+            navigation.row: 2 + model.index
```
RomanPudashkin, line 129: Please, explain why you use 2 here

src/framework/ui/view/navigationsection.h
```diff
@@ -36,12 +36,23 @@ namespace mu::ui {
...
     Q_OBJECT
+    Q_PROPERTY(QmlType type READ type_property WRITE setType NOTIFY typeChanged)
+
     INJECT(ui, INavigationController, navigationController)
 
 public:
     explicit NavigationSection(QObject* parent = nullptr);
     ~NavigationSection() override;
 
+    //! NOTE Please sync with INavigationSection::Type
```
RomanPudashkin, line 47: Can we have just have one enum?

src/userscores/qml/MuseScore/UserScores/NewScoreDialog.qml
igorkorsukov: Knowingly big to be at the very end

src/instruments/qml/MuseScore/Instruments/internal/FamilyView.qml
igorkorsukov: 2 this is next row after row 1 (navigation.name: "familiesBox")

src/userscores/qml/MuseScore/UserScores/internal/ChooseTemplatePage.qml
igorkorsukov: we have one panel before 

image

src/userscores/qml/MuseScore/UserScores/internal/TitleListView.qml
igorkorsukov: This is for `onNavigationActived: item.clicked()`

src/instruments/qml/MuseScore/Instruments/internal/FamilyView.qml
igorkorsukov: Thanks, removed

src/userscores/qml/MuseScore/UserScores/internal/TitleListView.qml
igorkorsukov: done

src/framework/ui/view/navigationsection.h
igorkorsukov: I thought about this but didn't find a good solution.
We can add Q_ENUM for `INavigationSection::Type`, but we must register `INavigationSection` in qml, for this we must add Q_GADGET, and in Qml we must write `type: INavigationSection.Type` (maybe we can use another name into qml) I think this is not the best solution.
Alternatively, we can move all emuns to a separate namespace, which is not good for C ++ code, but is possible.
(Also we have to remember what we want to avoid with Qt in libmscore).
Perhaps we can find a tricky solution for dynamic reinstance enums from one namespace to another, but I think this is too much. The current solution is simple and works.

RomanPudashkin (approved):

PR #8048, opened 2021-05-05
[MU4] Show part name in  instrument form instead of instrument name.

njvdberg:
If opening `ChooseInstrumentsPage`, display the part name if available and use the instrument names otherwise.

The behavior is now similar as it was in 3.x.

The PR is required for the upcoming score ordering PR which needs part information for the ordering.

### DIFF
--- a/src/instruments/internal/selectinstrumentscenario.cpp
+++ b/src/instruments/internal/selectinstrumentscenario.cpp
@@ -28,7 +28,7 @@ mu::RetVal<InstrumentList> SelectInstrumentsScenario::selectInstruments(SelectIn
 {
     QStringList params;
     if (mode == SelectInstrumentsMode::ShowCurrentInstruments) {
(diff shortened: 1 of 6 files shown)


Jojo-Schmitz:
Is this the for for https://musescore.org/en/node/317264 (and for master) and so the counterpart to #7544?

njvdberg:
@Jojo-Schmitz "kind of" and "no" :wink: 
Compared with `3.x` the new `Instruments` form is 100% rewrite and also the underlying C++ code is completely new. But by chance it showed the same problem, instrument names instead of part names. In this sense it is a kind of counterpart of #7544.
And no, #7571 wasn't enough. As a matter of fact it is doing something completely different. #7571 is changing the `Instruments` **Panel** while this PR is changing the `Instruments` **form**.

njvdberg:
This PR is included in PR #8099 .

PR #8050, opened 2021-05-05
[MU4] Added some utests for navigation controller

igorkorsukov:
The main purpose of this PR is to create an environment for navigation unit tests.

image

### DIFF
--- a/src/framework/ui/internal/navigationcontroller.h
+++ b/src/framework/ui/internal/navigationcontroller.h
@@ -76,7 +76,7 @@ private:
     void goToNextRowControl();
     void goToPrevRowControl();
 
-    void goToControl(MoveDirection direction, INavigationPanel* activeSubSec = nullptr);
+    void goToControl(MoveDirection direction, INavigationPanel* activePanel = nullptr);
 
     void onLeft();
     void onRight();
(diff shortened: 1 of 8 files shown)


src/framework/ui/internal/navigationcontroller.h
```diff
@@ -90,7 +90,7 @@ class NavigationController : public INavigationController, public actions::Actio
     void doActivateSection(INavigationSection* s, bool isActivateLastPanel = false);
     void doDeactivateSection(INavigationSection* s);
     void doActivatePanel(INavigationPanel* s);
-    void doDeactivatePanel(INavigationPanel* s);
+    void doDeactivatePanel(INavigationPanel* panel);
```
RomanPudashkin, line 93: Better to fix args names in all functions

src/framework/ui/tests/navigationcontroller_tests.cpp
```diff
@@ -0,0 +1,295 @@
...
+    //! GIVEN Two section, not active
+    Section* sect1 = make_section(env, 1, 2, 3);
+    Section* sect2 = make_section(env, 2, 2, 3);
+
+    env.controller.reg(sect1->section);
+    env.controller.reg(sect2->section);
+
+    //! CHECK The first section, the first panel, the first control must be activated.
+    EXPECT_CALL(*sect1->section, setActive(true));
+    EXPECT_CALL(*sect1->panels[0]->panel, setActive(true));
```
RomanPudashkin, line 200: I think it is better to add a check for panels/controls size:

ASSERT_FALSE(sect1->panels.isEmpty())
ASSERT_FALSE(sect1->panels[0].controls.isEmpty())

RomanPudashkin (approved):

src/framework/ui/tests/navigationcontroller_tests.cpp
igorkorsukov: > I think it is better to add a check for panels/controls size:
> 
> ASSERT_FALSE(sect1->panels.isEmpty())
> ASSERT_FALSE(sect1->panels[0].controls.isEmpty())

I think this is redundant because these panels and sections are created not by the code under test, but by the test itself. There are no conditions when lists can be empty, unless, of course, you explicitly set the count to zero, but in this case, I hope the developer understands what he is doing :)

src/framework/ui/internal/navigationcontroller.h
igorkorsukov: done

PR #8054, opened 2021-05-06
[MU4] Fixed PopupView on Windows

igorkorsukov:
I reimplemented PopupView again.. 
On Windows not shown popups 
I can fix it with the old implementation, but! 
I found, what on Windows impossible do transparent background for QQuickView (not top-level), so our popup with the arrow shown with black frame. 
Now used another implementation work with Qml View

I tested on Windows and Linux, I can't test on MacOS at the momemt. 

I believe that we must re-implement PopupMenu, which needs to make that the submenus do not create new windows, but everything is in one window.

### DIFF
--- a/src/appshell/appshell.cpp
+++ b/src/appshell/appshell.cpp
@@ -186,7 +186,7 @@ int AppShell::run(int argc, char** argv)
 
         //! Needs to be set because we use transparent windows for PopupView.
         //! Needs to be called before any QQuickWindows are shown.
-        QQuickWindow::setDefaultAlphaBuffer(true);
+        // QQuickWindow::setDefaultAlphaBuffer(true);
 
         engine->load(url);
     }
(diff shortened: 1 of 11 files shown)


cbjeukendrup:
Bad news: it does not work well on macOS...
When you initially open the app and then try to open a popup window or popup, the popup does not appear.
However, when you then click the app icon in the Dock while the app is running (to bring all windows to front?), the popups that should have opened earlier appear all at once, and from that moment popups work correctly.

igorkorsukov:
> Bad news: it does not work well on macOS...
> When you initially open the app and then try to open a popup window or popup, the popup does not appear.
> However, when you then click the app icon in the Dock while the app is running (to bring all windows to front?), the popups that should have opened earlier appear all at once, and from that moment popups work correctly.

I again change implementation... prev implementation too slow on some computers. So, as I will finish, will be nice, if we all tested it on different platforms

Eism:
Works well for me on Windows 10

cbjeukendrup:
Still not...
Now the popups appear inside the main window, not as separate windows. It seems to use the coordinates of the screen, but applied relative to the main window. So in full screen, they appear in the correctly, because then window coordinates and screen coordinates are the same. 

Here's a recording:

https://user-images.githubusercontent.com/48658420/117450186-31501900-af41-11eb-8e87-19a32a394216.mov

vshalkevich:
Windows seems fine. Better than it was

vpereverzev (approved):

PR #8065, opened 2021-05-07
[MU4] Added Async and Promise

igorkorsukov:
Async - called method on next event loop (same thread, not another), like Qt.callLater

Use 
```
LOGI() << "before call";
async::Async::call(this, []() {
    LOGI() << "call";
});
LOGI() << "after call";
```

Will be output:
before call
after call
call

Promise - primitive to return a value asynchronously

Use
```
mu::async::Promise<int> SomeClass::calc()
{
    static int counter = 0;
    int count = ++counter;
    return async::Promise<int>([count](mu::async::Promise<int>::Resolve resolve, mu::async::Promise<int>::Reject reject) {
        QTimer::singleShot(1000, [count, resolve, reject]() {
            if (count % 5 == 0) {
                reject(1, std::string("count % 5 == 0"));
            } else {
                resolve(count);
            }
        });
    });
}

void SomeClass::someMethod()
{
    calc().onResolve(this, [](int res) {
        LOGI() << "success calc: " << res;
    }).onReject(this, [](int code, const std::string& msg) {
        LOGE() << "failed calc, err: [" << code << "] " << msg;
    });
}
```


thirdparty/deto_async/async/internal/abstractinvoker.h
```diff
@@ -103,7 +103,7 @@ inline void processEvents()
     AbstractInvoker::processEvents();
 }
 
-inline void onMainThreadInvoke(const std::function<void(const std::function<void()>&)>& f)
+inline void onMainThreadInvoke(const std::function<void(const std::function<void()>&, bool)>& f)
```
igorkorsukov, line 106: I know, this is ugly, I will do better later, not now.

vpereverzev (approved):

Issue #7986, opened 2021-04-19
[MU4 Issue] Pressing the Tab button should not automatically move you through hidden panels


vshalkevich:
**Describe the bug**
Pressing the Tab button should not automatically move you through hidden panels

**To Reproduce**
1. Create a score
2. Add notes to the score
3. Add text to the score
4. Select text
5. Press "tab" button from your keyboard

**Expected behavior**
Pressing the Tab button should not automatically move you through hidden panels. However, a user should be able to access the actual panel tabs and change to a different panel by pressing **Spacebar** or **Enter**

Tantacrul:
@vshalkevich - you should probably assign @igorkorsukov to these bugs since they relate to the work he is doing.

PR #8069, opened 2021-05-08
[MU4] Fixed navigation to hidden controls

igorkorsukov:
https://github.com/musescore/MuseScore/issues/7986

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/GradientTabButton.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/GradientTabButton.qml
@@ -38,6 +38,7 @@ RadioDelegate {
     property var normalStateFont: ui.theme.tabFont
     property var selectedStateFont: ui.theme.tabBoldFont
 
+    property bool isDoToggleOnNavigationActive: true
     property alias navigation: keynavCtrl
 
     height: 48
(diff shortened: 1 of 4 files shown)


src/framework/uicomponents/qml/MuseScore/UiComponents/GradientTabButton.qml
```diff
@@ -38,6 +38,7 @@ RadioDelegate {
     property var normalStateFont: ui.theme.tabFont
     property var selectedStateFont: ui.theme.tabBoldFont
 
+    property bool isDoToggleOnNavigationActive: true
```
vpereverzev, line 41: maybe "isTogglingEnabled/Available" would sound better?

igorkorsukov: It is important here that when "navigation is activated" (focus transition), that is, the toggled signal occurs when the button is clicked, and in this case, it can be sent or not sent when the focus is moved

igorkorsukov: It doesn't matter anymore, in the next PR I will remove this functionality

vpereverzev (approved):

PR #8080, opened 2021-05-11
[MU4] Move focus to notation view on note input

igorkorsukov:
At the moment, the method` UINavigationController::activateByName` not necessary, but I decided to keep it, I think it will help in the future.

### DIFF
--- a/src/notation/view/notationpaintview.h
+++ b/src/notation/view/notationpaintview.h
@@ -109,7 +109,7 @@ signals:
     void backgroundColorChanged(QColor color);
     void viewportChanged(QRect viewport);
 
-    void mousePressed();
+    void activeFocusRequested();
 
 protected:
     void setNotation(INotationPtr notation);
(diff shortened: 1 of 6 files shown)


src/notation/view/notationpaintview.h
```diff
@@ -109,7 +109,7 @@ class NotationPaintView : public QQuickPaintedItem, public IControlledView, publ
     void backgroundColorChanged(QColor color);
     void viewportChanged(QRect viewport);
 
-    void mousePressed();
+    void requestActiveFocus();
```
Eism, line 112: may be activeFocusRequested?

src/notation/internal/notationactioncontroller.h
```diff
@@ -44,6 +45,7 @@ class NotationActionController : public actions::Actionable, public async::Async
     INJECT(notation, playback::IPlaybackController, playbackController)
     INJECT(notation, playback::IPlaybackConfiguration, playbackConfiguration)
     INJECT(notation, INotationConfiguration, configuration)
+    INJECT(notation, ui::INavigationController, navigationController)
```
Eism, line 48: You are not using this class

igorkorsukov: removed, thanks!

Eism (approved):

PR #7995, opened 2021-04-23
[MU4] Change instrument id from musicXMLid to an unique instrument id

njvdberg:
Resolves: https://musescore.org/en/node/320404

Change instrument id from `musicXMLid` to an unique instrument id as defined by the `id` attribute of the `instrument` tag in `instruments.xml`.

Main reason for this change are `musicXMLid` isn't unique for instruments. Also not all instruments have a `musicXMLid` defined.

A small check of the existing `instruments.xml`:
1) 584 instruments defined.
2) 561 instruments don't have an `musicXMLid`.
3) There are only 335 unique  `musicXMLid`'s defined.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8099, opened 2021-05-13
[MU4] Preparation ScoreOrdering: part names in instruments form

njvdberg:
This PR is the first of a number of PR's required to
- implement score ordering.
- solve several issues with `Parts` and `Excerpts` 
- independent sorting of staves/parts in both main score and excepts.

The changes in this PR are:

-1 Replace *MusicXMLid* as an instrument id by the `id` as given in `instruments.xml`
This solves the issue an *Oboe* suddenly becomes a *Baroque Oboe* when the instruments form is opened again. The reason for this name change is *MusicxXMLid* isn't unique, several instrument share the same *MusicXMLid*. Both *Oboe* and *Baroque Oboe* share the same *MusicXMLid*, *wind.reed.oboe* but there also 3 *Cornet*s, 4 *Bariton Horn*s and even 10 *Horn*s share the same id. This similar to PR #7995 but since it was required here too it is include in this PR and PR #7995 is closed.

-2 *Your score* in instrument form now shows `Parts` rather than `Instruments`
In the list under *Your score*, "Instruments" should refer to `Parts`, not `Instruments`. Please note `Part` is a class as used internally, not "Part" as seen on the UI, which is actually an `Excerpt`. And to complicate things even more, the word "Instrument" on the UI sometimes refers to a `Part`-class, sometime to an `Instrument`-class. The column *Your score* actually show the `Parts` for an `Instrument`. 
By this change 2 issues are solved
  - When changing the part name *Your score* now shows this part name.
  - Without this PR, deleting in instrument in a part with an instrument change removes the wrong instruments. E.g. in a part starting with an *Alto Saxophone*, switching to *Clarinet*, switching to *Tenor Saxophone* and back to *Alto Saxophone*, removes all instrument changes!
With this PR the `Part` is removed, what actually is supposed to happen. PR #8048 is partly doing the same but this PR is more elaborated and replaces PR #8048.

-3 *Soloist* function implemented
When a part is made *Soloist* the part is made soloists. The form now also reflect the actual status of the part, soloist not no soloist. This is first step to implementing score ordering.

-4 `Make soloist` button moved
Original the *Make soloist* button was in the header of the part list under *Your score*. To make more space for the score order name is was decided to move this button to the instrument line. For the discussion, see the discussion with @Tantacrul on Discord (`design` channel, April 15th, 2021).

-5 Some minor typos corrected.
When encountered some minor typos are corrected.

### DIFF
--- a/src/instruments/instrumentstypes.h
+++ b/src/instruments/instrumentstypes.h
@@ -162,7 +162,14 @@ struct Instrument
     QString abbreviature() const { return !shortNames.isEmpty() ? shortNames.first().name() : QString(); }
 };
 
-using InstrumentList = QList<Instrument>;
+struct PartInstrument {
+    bool part;
+    bool soloist;
+    QString partId;
+    Instrument instrument;
+};
+
+using PartInstrumentList = QList<PartInstrument>;
 
 struct InstrumentTemplate
 {
--- a/src/instruments/instrumentstypes.h
+++ b/src/instruments/instrumentstypes.h
@@ -163,8 +163,8 @@ struct Instrument
 };
 
 struct PartInstrument {
-    bool part;
-    bool soloist;
+    bool isExistingPart;
+    bool isSoloist;
     QString partId;
     Instrument instrument;
 };
--- a/src/instruments/instrumentstypes.h
+++ b/src/instruments/instrumentstypes.h
@@ -122,6 +122,7 @@ struct Instrument
     StaffNameList longNames;
     StaffNameList shortNames;
     QString name;
+    QString musicXMLid;
     QString description;
 
     bool extended = false;
--- a/src/instruments/internal/instrumentsreader.cpp
+++ b/src/instruments/internal/instrumentsreader.cpp
@@ -155,7 +155,8 @@ InstrumentTemplate InstrumentsReader::readInstrumentTemplate(Ms::XmlReader& read
     InstrumentTemplate instrumentTemplate;
     Instrument& instrument = instrumentTemplate.instrument;
 
-    instrumentTemplate.id = reader.attributes().value("id").toString();
+    instrument.id = reader.attributes().value("id").toString();
+    instrumentTemplate.id = instrument.id;
 
     bool customDrumset = false;
 
@@ -242,7 +243,7 @@ InstrumentTemplate InstrumentsReader::readInstrumentTemplate(Ms::XmlReader& read
         } else if (reader.name() == "transposeDiatonic") {
             instrument.transpose.diatonic = reader.readElementText().toInt();
         } else if (reader.name() == "instrumentId") {
-            instrument.id = reader.readElementText();
+            instrument.musicXMLid = reader.readElementText();
         } else if (reader.name() == "StringData") {
             instrument.stringData = readStringData(reader);
         } else if (reader.name() == "useDrumset") {
@@ -296,7 +297,7 @@ InstrumentTemplate InstrumentsReader::readInstrumentTemplate(Ms::XmlReader& read
             QString templateId = reader.readElementText();
             initInstrument(instrument, generalMeta.instrumentTemplates[templateId].instrument);
         } else if (reader.name() == "musicXMLid") {
-            instrument.id = reader.readElementText();
+            instrument.musicXMLid = reader.readElementText();
         } else if (reader.name() == "genre") {
             instrument.genreIds << reader.readElementText();
         } else if (reader.name() == "singleNoteDynamics") {
@@ -417,6 +418,7 @@ void InstrumentsReader::fillByDefault(Instrument& instrument) const
 void InstrumentsReader::initInstrument(Instrument& sourceInstrument, const Instrument& destinationInstrument) const
 {
     sourceInstrument.id = destinationInstrument.id;
+    sourceInstrument.musicXMLid = destinationInstrument.musicXMLid;
     sourceInstrument.longNames = destinationInstrument.longNames;
     sourceInstrument.shortNames = destinationInstrument.shortNames;
     sourceInstrument.staves = destinationInstrument.staves;
--- a/src/instruments/internal/instrumentsreader.cpp
+++ b/src/instruments/internal/instrumentsreader.cpp
@@ -306,7 +306,7 @@ InstrumentTemplate InstrumentsReader::readInstrumentTemplate(Ms::XmlReader& read
         }
     }
 
-    fillByDeffault(instrument);
+    fillByDefault(instrument);
 
     return instrumentTemplate;
 }
@@ -385,7 +385,7 @@ StringData InstrumentsReader::readStringData(Ms::XmlReader& reader) const
     return StringData(frets, strings);
 }
 
-void InstrumentsReader::fillByDeffault(Instrument& instrument) const
+void InstrumentsReader::fillByDefault(Instrument& instrument) const
 {
     if (instrument.channels.empty()) {
         Channel a;
--- a/src/instruments/view/instrumentlistmodel.cpp
+++ b/src/instruments/view/instrumentlistmodel.cpp
@@ -36,6 +36,8 @@ static const QString NAME_KEY("name");
 static const QString TRANSPOSITIONS_KEY("transpositions");
 static const QString GROUP_ID("groupId");
 static const QString CONFIG_KEY("config");
+static const QString SOLOIST_KEY("isSoloist");
+static const QString PART_KEY("isExistingPart");
 
 InstrumentListModel::InstrumentListModel(QObject* parent)
     : QObject(parent)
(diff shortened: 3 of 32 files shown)


Tantacrul:
The new position of the soloist button is working well, I think.

We'll need to do a bit of UI cleanup on the button. One quick thing: I think we should reduce the distance between the edges of the text and the edge of the button

image

I also think we need to create a special button style just for this. @njvdberg - are you comfortable implementing that or should I ask @cbjeukendrup or @Eism ?

njvdberg:
@Tantacrul Depends on how difficult the specs of this new button will be and on the urgency. If it isn't the most urgent job, I'm willing to give it try, always nice to explore new things. I suppose I can fallback on @cbjeukendrup and/or @Eism in case I'm running into problems. Nice MSoC project (Musescore Summer of Code) :smiley:

cbjeukendrup:
@njvdberg I would recommend to extend the existing FlatButton component, rather than creating a new one. You could add a property called something like `narrowMargins`, and then take that into account for the width. 

Currently, the 'calculation' for the width is on one line, so when taking the new property into account, I think it's a good idea to write it out like
```
    width: {
        if (...) {
            return ...
        }

        return ...
    }

**Comment by njvdberg** [comment]
PR #8147 requires this PR, so please merge this PR before PR #8147.

File: src/libmscore/undo.h [hunk 1]
```diff
@@ -256,6 +256,22 @@ class RemovePart : public UndoCommand
     UNDO_NAME("RemovePart")
 };
 
+//---------------------------------------------------------
+//   SetSoloist
+//---------------------------------------------------------
+
+class SetSoloist : public UndoCommand
+{
+    Part* part;
```
RomanPudashkin, line 265: Please, add default values here:

Part* part = nullptr;
bool soloist = false;

src/instruments/view/instrumentlistmodel.cpp
```diff
@@ -259,9 +289,16 @@ void InstrumentListModel::selectInstrument(const QString& instrumentId, const QS
     emit selectedInstrumentsChanged();
 }
 
-void InstrumentListModel::makeSoloist(const QString&)
+void InstrumentListModel::toggleSoloist(const QString& instrumentId)
 {
-    NOT_IMPLEMENTED;
+    for (int i = 0; i < m_selectedInstruments.count(); ++i) {
+        if (m_selectedInstruments[i].id == instrumentId) {
```
RomanPudashkin, line 295: I think, we need to add a new method for getting an index of a selected instrument by id:

int selectedInstrumentIndex(instrumentId)

Because the code for getting the selected instrument can be found in two different places (toggleSoloist and unselectInstrument)

src/instruments/instrumentstypes.h
```diff
@@ -161,7 +162,14 @@ struct Instrument
     QString abbreviature() const { return !shortNames.isEmpty() ? shortNames.first().name() : QString(); }
 };
 
-using InstrumentList = QList<Instrument>;
+struct PartInstrument {
+    bool part;
```
RomanPudashkin, line 166: Please, add the word "is" (as well as a default value) to each of bool members here:

bool isPart = false;
bool isSoloist = false;

RomanPudashkin, line 166: I don't quite get the meaning of the flag "part". Can you please explain it to me?

RomanPudashkin:
@vshalkevich please, check this PR

src/instruments/instrumentstypes.h
njvdberg: Apparently a wrong name then :wink: 
The Instrument Form returns a list of `PartInstrument` structure, representing the new list, and order, of `Parts`. For processing it is important whether an element of this list represents a newly added `Part` or an already existing `Part`.
E.g. when setting part Soloist, for new parts the flag can be set straight away, for existing `Parts` the undo/redo has to be used.

src/libmscore/undo.h
njvdberg: Good point. However, it seems none of the classes in this file supplied a default value. Might be an idea to go over this file and supply default values for all members of all other classes.

njvdberg: Done.

src/instruments/instrumentstypes.h
njvdberg: Done.

src/instruments/view/instrumentlistmodel.cpp
njvdberg: Made a private method `InstrumentListModel::findInstrumentIndex(const QString& instrumentId) const` which returns the index in `m_selectedInstruments` of the specified `instrumentId` and this method is used in both `toggleSoloist()` and `unselectInstrument()`.

src/instruments/instrumentstypes.h
njvdberg: Name wasn't indeed not that clear, changed into `isExistingPart`.

vshalkevich:
@njvdberg Could you rebuild failed MU4 file for Window 10 x64 ? I would like to check also on it

RomanPudashkin (approved):

cbjeukendrup:
> Could you rebuild failed MU4 file for Window 10 x64 ? I would like to check also on it

It must have failed due to the Chocolatey package manager server being down due to maintenance. Now, it should be working again, so when triggering the workflow again, it will build.

njvdberg:
> It must have failed due to the Chocolatey package manager server being down due to maintenance.

I didn't realize that thing was down to whatever reason but I reckoned the cause was somewhere outside my scope :wink:

src/instruments/internal/instrumentsreader.cpp
```diff
@@ -242,7 +243,7 @@ InstrumentTemplate InstrumentsReader::readInstrumentTemplate(Ms::XmlReader& read
         } else if (reader.name() == "transposeDiatonic") {
             instrument.transpose.diatonic = reader.readElementText().toInt();
         } else if (reader.name() == "instrumentId") {
-            instrument.id = reader.readElementText();
+            instrument.musicXMLid = reader.readElementText();
```
cbjeukendrup, line 246: Why are both `"instrumentId"` and `"musicXMLid"` going to `instrument.musicXMLid`?

njvdberg: "`instrumentId`" is the unique identification, that's the one which we really need. "`muscXMLid`" is still used somewhere, e.g. when reading a score which doesn't have a `id` specified for an instrument. Then `musicXMLid` is used for guess what instrument it should be. It is also used for reading MU1 and MU2 scores.
So it is best keep this parameter (and up-to-date).

cbjeukendrup: Shouldn't `"instrumentId"` then go to `instrument.id` instead of `instrument.musicXMLid`?

njvdberg: No, in the `Instrument` section of the score file, the MusicXMLid is called `instrumentId` (because was used as an id in older version of MuseScore. The same MusicXMLid is defined in `instruments.xml` with the `musicXMLid` tag.
An example:
In `instruments.xml` the Flute is defined as
```
<Instrument id="flute">
    <family>flutes</family>
    <longName>Flute</longName>
    <shortName>Fl.</shortName>
    <description>Standard Concert Flute</description>
    <musicXMLid>wind.flutes.flute</musicXMLid>
           .
           .
```
In the score file the same Flute appears as 
```
   <Instrument id="flute">
    <longName>Flute</longName>
    <shortName>Fl.</shortName>
           .
           .
    <instrumentId>wind.flutes.flute</instrumentId>
           .
           .
```
But welcome in the wonder world of instrument id's :wink: Here is some history....

cbjeukendrup: Ah, now I understand! Thanks!

njvdberg: My pleasure :smiley: 

This drove me crazy last year, all these id's and in the meantime changing names.

njvdberg:
Just another rebase.

src/instruments/internal/instrumentsreader.cpp
shoogle: For MU4, I think we would be justified in changing `<instrumentId>` to `<MusicXMLid>` in the MSCX file format and using compatibility code to deal with old scores during import. There will already be other changes in the file format (e.g. for the Smart Tempos project), so it might be worth tidying up the syntax here and potentially [a few other places](https://musescore.org/en/node/296654) too.

njvdberg: Changing the tag `<instrumentId>` into '<MusicXMLid>` would be nice and makes thing more clear :smile:

PR #8103, opened 2021-05-14
[MU4] Fix #321174: Organ sound for "SATB Closed Score + Organ" template is pan flute

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/321174

Counterpart of #8102, but for master


njvdberg:
Not sure what sound better :wink:

Jojo-Schmitz:
True, but now it is at least consistent with the full SATB+Organ template

PR #8104, opened 2021-05-14
[MU4] Fixed view mode availability

igorkorsukov:


src/notation/qml/MuseScore/NotationScene/ViewModeControl.qml
```diff
@@ -62,6 +62,8 @@ Item {
                 return
             }
 
+            menu.model = 0
+            menu.model = model.items
```
Eism, line 66: binding model.items to the model in the StyledMenu is no longer required

src/notation/qml/MuseScore/NotationScene/ViewModeControl.qml
```diff
@@ -62,6 +62,8 @@ Item {
                 return
             }
 
+            menu.model = 0
```
Eism, line 65: better to use the clear function for model

igorkorsukov: Did not understand the suggestion. Are you suggesting adding a function?
```
function clear() {
    menu.model = 0
}
```

Eism: No, I mean menu.model.clear()

Eism: No, I mean menu.model.clear()

PR #8107, opened 2021-05-14
[MU4] Docksystem toolbars

Eism:


src/appshell/qml/kdab/WindowContent.qml
```diff
@@ -23,7 +23,7 @@ import QtQuick 2.15
 import QtQuick.Controls 2.15
 
 import MuseScore.AppShell 1.0
-import MuseScore.Dock 1.0
+import MuseScore.Dock 1.0 as Dock
```
igorkorsukov, line 26: Why did you decide to add an alias and use it?

igorkorsukov (approved):

src/appshell/qml/kdab/WindowContent.qml
RomanPudashkin: removed

PR #8100, opened 2021-05-14
[MU3 Backend] ENG-27: Set XML import preferences and styling

iveshenry18:
Resolves: [ENG-27](https://mu--se.atlassian.net/browse/ENG-27) Determine import process and options

Change default preferences, reset style on MusicXML import, and apply new PVGMusicXML.mss.

- [X] I signed [CLA](https://musescore.org/en/cla)
- [X] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [X] I made sure the code compiles on my machine
- [X] I made sure there are no unnecessary changes in the code
- [X] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [X] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [X] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [X] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8108, opened 2021-05-15
[MU3 Backend] ENG-17: multi-voice beam fix

iveshenry18:
Resolves: [ENG-17](https://mu--se.atlassian.net/browse/ENG-17) Beams become unjoined when there are two voices

The presence of a second voice would prematurely break
beams upon MusicXML import in some cases, as only one beam was
stored at a time as the "current beam", and this would get thrown away
if a note in another voice/track was handled before the end of a
beam. This commit adds a QMap to store one beam *per voice* at a time,
avoiding the conflict while retaining the existing logic.

__NOTE:__ this branch currently includes the commits from PR #8100 as well.

- [X] I created the test (mtest, vtest, script test) to verify the changes I made


Jojo-Schmitz:
Is any of this needed for master too?

iveshenry18:
Yes, I believe the plan is to port this to master later. The previous PR (#8100) will not apply because it is specific to the conversion job I mentioned previously, but this one is a general beaming fix that can apply to all XML imports.

Jojo-Schmitz:
So that 3rd commit especially, but not the 1st and 2nd, right?

PR #8111, opened 2021-05-15
[MU3] Fix #321227: Problems with removing empty trailing measures

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/321227

My approach is instead of `Measure::isFullMeasureRest()` to use `Measure::isEmpty(-1)` which is documented to check a measure accross all staves.
Problem is, that this doesn't work (anymore?), but crashes, so that needed to get fixed too. Also in that special case annotations with the system flag being set, fermatas and invisible annotations should should mean a measure is not empty.

Corresponding PR for master is #8112 (merged already)

### DIFF
--- a/libmscore/measure.cpp
+++ b/libmscore/measure.cpp
@@ -2841,7 +2841,6 @@ bool Measure::isEmpty(int staffIdx) const
       {
       int strack;
       int etrack;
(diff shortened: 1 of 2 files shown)


Jojo-Schmitz:
See #9000

PR #8114, opened 2021-05-18
[MU4] Fix #8098: Text-only buttons have not the same height as icon b…

### LINKED ISSUE #8098: [MU4 Issue] Text-only buttons have not the same height as icon buttons
**Describe the bug**
Text-only buttons have not the same height as icon buttons. Text-only buttons are 29px high instead of 30px. 

**To Reproduce**
1. Find a place where a text-only button and an icon button are next to each other. 
    For example:
    1. Header of the palettes panel (see also: https://github.com/musescore/MuseScore/issues/7634)
    2. Header of the instruments panel
    3. Header of the parts dialog
2. The text-only button is 1px smaller than the icon button. 

**Expected behavior**
Probably we should give text-only buttons a minimum height of 30px.

**Screenshots**
(This screenshot is from a high resolution display with a scaling factor of 2.)
Schermafbeelding 2021-05-13 om 16 19 11

 - OS: N/A

JRYamasaki:
…uttons

    -Text is 10 pixels high (even) while the icons are 15 pixels high (odd).
    As a result we require a one pixel offset for the height of buttons that
    contain only text.

Resolves: https://github.com/musescore/MuseScore/issues/8098

**The Bug**
The text is 10 pixels high (even) while the icons are 15 pixels high (odd). As a result we require a one pixel offset for the height of buttons that contain only text.

**The Fix**
If we have a button with only text, add one pixel to the height of the button to compensate for the even-pixel height of 10.

Alternatively, the text could be increased to 11 pixels and then we would not require this change.

**What the fix looks like**
palette
instrument
parts

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
@@ -86,11 +86,12 @@ FocusableControl {
         id: contentWrapper
 
         property int spacing: Boolean(!buttonIcon.isEmpty) && Boolean(textLabel.text) ? 4 : 0
+        property int extraSpacing: buttonIcon.isEmpty && textLabel.text ? 1 : 0 // Required because font is an even-pixel height (10px), #8098
 
         anchors.verticalCenter: parent ? parent.verticalCenter : undefined
         anchors.horizontalCenter: parent ? parent.horizontalCenter : undefined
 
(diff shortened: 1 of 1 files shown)


RobFog:
@cbjeukendrup, you filed #8098. Would you be able to review this pull request?

@JRYamasaki, it looks like the pull request requires a rebase.

cbjeukendrup:
This fix will break as soon as the user chooses a different font size or family. I think something more thorough is needed. I'll take a look myself probably tonight.

PR #8119, opened 2021-05-19
[MU4] Accessibility concept implementation  

igorkorsukov:
Qt has a low-level cross-platform interface for implementing accessibility on different platforms (Windows, macOS, and Linux), as well as a high-level one for QWidget and QQuick.

We want to implement accessibility for the notation (libmscore) as well, but we will not be able to use high-level interfaces QWidget or QQuick there is.

I also believe that if we use the high-level interface in QQuick, then this is a quick start, but in the long term, it will cause more problems when we want to implement our behaviors, comply with the navigation system, etc.

For these reasons, I decided to use only the low-level Qt Accessibility interface. And will develop your own high-level interface, which can be applied anywhere in the application, including the notation (libmscore)

This is a prototype solution.
Not everything is done yet, but something is already working :)

https://user-images.githubusercontent.com/3818029/118850866-7d03aa80-b8d1-11eb-8eec-bebe6439ac06.mp4

### DIFF
--- a/src/framework/accessibility/iaccessibilitycontroller.h
+++ b/src/framework/accessibility/iaccessibilitycontroller.h
@@ -20,23 +20,24 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-#ifndef MU_ACCESSIBILITY_ACCESSIBILITYMODULE_H
-#define MU_ACCESSIBILITY_ACCESSIBILITYMODULE_H
+#ifndef MU_ACCESSIBILITY_IACCESSIBILITYCONTROLLER_H
+#define MU_ACCESSIBILITY_IACCESSIBILITYCONTROLLER_H
 
-#include "framework/global/modularity/imodulesetup.h"
+#include "modularity/imoduleexport.h"
+#include "iaccessibility.h"
 
 namespace mu::accessibility {
-class AccessibilityModule : public framework::IModuleSetup
+class IAccessibilityController : MODULE_EXPORT_INTERFACE
 {
+    INTERFACE_ID(IAccessibilityController)
 public:
-    std::string moduleName() const override;
+    virtual ~IAccessibilityController() = default;
 
-    void registerExports() override;
-    void resolveImports() override;
-    void registerResources() override;
-    void registerUiTypes() override;
-    void onInit(const framework::IApplication::RunMode& mode) override;
+    virtual void created(IAccessibility* parent, IAccessibility* item) = 0;
+    virtual void destroyed(IAccessibility* item) = 0;
+    virtual void actived(IAccessibility* item, bool isActive) = 0;
+    virtual void focused(IAccessibility* item) = 0;
 };
 }
 
-#endif // MU_ACCESSIBILITY_ACCESSIBILITYMODULE_H
+#endif // MU_ACCESSIBILITY_IACCESSIBILITYCONTROLLER_H
--- a/src/framework/accessibility/internal/accessibilitycontroller.cpp
+++ b/src/framework/accessibility/internal/accessibilitycontroller.cpp
@@ -1,0 +1,205 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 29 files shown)


src/framework/accessibility/internal/accessibilitycontroller.cpp
```diff
@@ -0,0 +1,205 @@
...
+        return;
+    }
+    Item item = m_items.takeAt(idx);
+    QAccessibleEvent ev(item.object, QAccessible::ObjectDestroyed);
+    sendEvent(&ev);
+
+    delete item.object;
+}
+
+void AccessibilityController::actived(IAccessibility* aitem, bool isActive)
```
vpereverzev, line 111: activated?

src/framework/accessibility/internal/accessibilitycontroller.cpp
```diff
@@ -0,0 +1,205 @@
...
+    LOGI() << aitem->accessibleName();
+    int idx = indexBy(aitem);
+    IF_ASSERT_FAILED(idx >= 0) {
+        return;
+    }
+    Item item = m_items.takeAt(idx);
+    QAccessibleEvent ev(item.object, QAccessible::ObjectDestroyed);
+    sendEvent(&ev);
+
+    delete item.object;
```
vpereverzev, line 108: do we really need to delete object here? The point is that it's a bit unclear when this function would be called (before an object will be destroyed or after)

Perhaps make sense to rename the function

src/framework/accessibility/internal/accessibilitycontroller.cpp
```diff
@@ -0,0 +1,205 @@
...
+    QAccessibleEvent ev(item.object, QAccessible::Focus);
+    sendEvent(&ev);
+}
+
+void AccessibilityController::sendEvent(QAccessibleEvent* ev)
+{
+    QAccessible::installUpdateHandler(0);
+    QAccessible::updateAccessibility(ev);
+    //! NOTE Disabled any events from Qt
+    QAccessible::installUpdateHandler(updateHandlerNoop);
```
vpereverzev, line 142: isn't it enough to mock it at "init"?

src/framework/accessibility/internal/accessibilitycontroller.cpp
vpereverzev: I'm also thinking about the existing QObject destruction mechanism. How likely it would cross-over with this functionality

src/framework/accessibility/iaccessibilitycontroller.h
```diff
@@ -0,0 +1,43 @@
...
+namespace mu::accessibility {
+class IAccessibilityController : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IAccessibilityController)
+public:
+    virtual ~IAccessibilityController() = default;
+
+    virtual void created(IAccessibility* parent, IAccessibility* item) = 0;
+    virtual void destroyed(IAccessibility* item) = 0;
+    virtual void actived(IAccessibility* item, bool isActive) = 0;
```
vpereverzev, line 38: I just realised that we have both "activated" and "focused" methods at the same time, could you explain when it'd be called and what does it mean for an item that it's active/inactive?

src/framework/accessibility/internal/accessibilitycontroller.cpp
igorkorsukov: This is a service object, which the controller created itself, so the controller deleted this object then it not need more.

igorkorsukov: > I'm also thinking about the existing QObject destruction mechanism. How likely it would cross-over with this functionality

For this, need set parent to object, in this case, controller creating this object, so it is parent, but! controller live all time, so all objects will be live all time.

src/framework/accessibility/internal/accessibilitycontroller.cpp
igorkorsukov: In Qt such code
```
void QAccessible::updateAccessibility(QAccessibleEvent* event)
{
        ...
        if (updateHandler) {
            updateHandler(event);
            return;
        }
        ...
}
```

So, I set a noop handler to prevent any events.
To send my event, I reset the handler, call the update event, and set the noop handler again.

src/framework/accessibility/internal/accessibilitycontroller.cpp
vpereverzev: got it

src/framework/accessibility/iaccessibilitycontroller.h
igorkorsukov: The answer to this question not fully clear at the moment. At the moment, for example, "panel" may be active, and "control" may be focused. 

I try to do this clarity later

vpereverzev (approved):

PR #8131, opened 2021-05-20
[MU4] Implemented accessible tree

igorkorsukov:
Now Screen Reader "reads" panels as well :)

https://user-images.githubusercontent.com/3818029/119001350-2f984380-b98c-11eb-93f1-3f412d217bd6.mp4


Jojo-Schmitz:
Just in time for the 10th aniversary of the [Global Accessibility Awareness Day (GAAD)](https://en.wikipedia.org/wiki/Global_Accessibility_Awareness_Day)

Tantacrul:
Guys, it is really wonderful that we are getting these details in NOW. Excellent progress!

MS4 is going to be an absolute beast! And better yet, we're making it better for everyone.

MarcSabatella:
Wow, crazy, I had no idea GAAD was a thing.  Coincidentally, I'm being interviewed for an accessibility-related podcast today too.

I'm also glad to see this work being integrated.  From a less technical level, can someone give me a brief overview of what this will mean?  I have the sense that keyboard navigation of UI elements is already improved in master compared to 3.6.  The screen reader experience is already pretty good in MuseScore with some annoying exceptions here and there - but also, only with those screen readers we actually support.  Will this implementation allow MuseScore 4 to work with VoiceOver, for instance?

Jojo-Schmitz:
I had no idea either, but that just came up in an email at my dayjob, so I had to share it ;-)

igorkorsukov:
> Wow, crazy, I had no idea GAAD was a thing. Coincidentally, I'm being interviewed for an accessibility-related podcast today too.
> 
> I'm also glad to see this work being integrated. From a less technical level, can someone give me a brief overview of what this will mean? I have the sense that keyboard navigation of UI elements is already improved in master compared to 3.6. The screen reader experience is already pretty good in MuseScore with some annoying exceptions here and there - but also, only with those screen readers we actually support. Will this implementation allow MuseScore 4 to work with VoiceOver, for instance?

We use Qt Accessibility.
Qt supports UI Automation on Windows, macOS Accessibility on macOS, and AT-SPI via DBus on Unix/X11. 
https://doc.qt.io/qt-5/accessible.html
So, I think VoiceOver will be work too

Now the system under development will allow adding accessibility for any place of the application, including the notation view

vpereverzev (approved):

MarcSabatella:
Yes, in theory the QAccessible system supports macOS, but for reasons we don't really understand, the methods we were using to trigger the screen reader to read the score as you navigate (firing "value changed" or other events on the ScoreView object) just wouldn't get VoiceOver's attention,  The event gets sent and the info is there butVoiceOver just doesn't read it unless you then do whatever the special "read description of current object" command in VoiceOver is.  @shoogle did a prototype of an alternate implementation of ScoreView that *should* work.  but I don't know how that fits in with the work here.

shoogle:
I didn't know the devs were working on this so I'm as surprised and impressed as anyone to see it included now!

@MarcSabatella, I did indeed create an [example Qt application](https://github.com/shoogle/accessibility-qt-chart-example#accessibility-for-qt-chart-example) that demos accessibility for a custom widget. I based it on the [Qt Chart Example](https://doc.qt.io/qt-5/qtwidgets-itemviews-chart-example.html) rather than our ScoreView, but of course the ScoreView is what I had in mind when I created it. The implementation looks very similar to what @igorkorsukov has done here, so the notation stuff should slot in nicely.

Interestingly, I found that it is possible to have accessibility that is "too good". For example, the screen reader might say:

- _"Note C4 quarter **button**"_ when we would prefer it just said _"Note C4 quarter"_ (at least in the Score View).

- _"Rest **1 of 685**, selected"_ when we would prefer it just said _"Rest, selected"_.

To avoid getting this unwanted information, I found it was necessary to [derive](https://github.com/shoogle/accessibility-qt-chart-example/blob/872e2d03afcb824cd7ba3003bb81ebebaec47e72/accessiblepieview.h#L42) directly from the base class [QAccessibleInterface](https://doc.qt.io/qt-5/qaccessibleinterface.html) instead of the specialised [QAccessibleTableInterface](https://doc.qt.io/qt-5/qaccessibletableinterface.html) class. I also found it was necessary to [return a different Role](https://github.com/shoogle/accessibility-qt-chart-example/blob/872e2d03afcb824cd7ba3003bb81ebebaec47e72/accessiblepieview.cpp#L168-L185) on macOS compared to the other platforms.

Of course those findings apply specifically to the ScoreView. In other places we would probably want to know that something is a button, or how many items there are in total.

shoogle:
When it comes to the ScoreView, I am toying with the idea that, instead of giving every element in the score it's own accessible object (possibly thousands), we just create 2 dummy accessible objects and then switch focus from one to the other any time we want the screen reader to say something:

1. User selects a note.
    - We rename `Object1` to _"Note C4 quarter"_ and tell the screen reader that it has focus.
2. User selects a rest.
    - We rename `Object2` to _"Rest"_ and give it focus.
3. User selects a tie.
    - We rename `Object1` to _"Tie"_ and give it focus again.

It's a bit hacky of course, but it should work just as well as doing it properly, at least in terms of keyboard accessibility. It might not work for mouse users who want to be able to hover the cursor over something and have the screen reader say what it is, but that might not be feasible anyway for performance reasons.

In fact, we might be _forced_ to use a hacky solution to deal with things like range selections and other state changes:

4. User starts a range selection.
    - We rename `Object2` to _"Range selection: start beat 1, end beat 3"_ and give it focus again.
5. User adds staccato dots to the range selection.
    - We rename `Object1` to _"Staccato, added 3, removed 0"_ and give it focus again.
6. User enters note input mode.
    - We rename `Object2` to _"Note input mode"_ and give it focus again.

I don't know how we can trigger speech output for those events other than by creating a dummy item with the appropriate name and tricking the screen reader into thinking it has focus.

MarcSabatella:
That's pretty much the exact hack I have long had a notion to try.

PR #8132, opened 2021-05-20
[MU4] Standard dialogs interface

Eism:


src/framework/global/iinteractive.h
```diff
@@ -88,21 +88,41 @@ class IInteractive : MODULE_EXPORT_INTERFACE
...
+    {
+        T1 button = T1(Button::NoButton);
+        bool showAgain = true;
+
+        Result() = default;
+        Result(const T1& button)
+            : button(button) {}
+        Result(const T1& button, bool showAgain)
+            : button(button), showAgain(showAgain) {}
+    };
```
igorkorsukov, line 102:
```
struct Result
{
        Result() = default;
        Result(const T1& button)
            : m_button(button) {}
        Result(const T1& button, bool showAgain)
            : m_button(button), showAgain(showAgain) {}

Button standartButton() const { return static_cast<Button>(button); }
int button() const { return m_button; } 

private:
        int m_button = int(Button::NoButton);
        bool showAgain = true;
};
```

src/framework/global/iinteractive.h
```diff
@@ -88,21 +88,41 @@ class IInteractive : MODULE_EXPORT_INTERFACE
...
+
+    enum class CustomizeType {
+        WithIcon,
+        WithShowAgainOption
+    };
+
+    virtual Result<Button> question(const std::string& title, const std::string& text, const Buttons& buttons,
+                                    const Button& def = Button::NoButton, const QFlags<CustomizeType>& customize = {}) const = 0;
```
igorkorsukov, line 110: I think it better to rename CustomizeType to Option

src/framework/global/iinteractive.h
Eism: thanks, fixed

igorkorsukov (approved):

PR #8142, opened 2021-05-23
[MU4] Notenames plugins: For hidden notes skip notenames

Jojo-Schmitz:
This PR fixes:

* If a note(head) is invisble, the notenames plugins still put a visible name to it. 
* If there's a list selection, only the 1st note of that gets named.

### DIFF
--- a/share/plugins/notenames-interactive.qml
+++ b/share/plugins/notenames-interactive.qml
@@ -25,7 +25,7 @@ import QtQuick.Controls 2.0
 import MuseScore 3.0
 
 MuseScore {
-    version: "3.4"
+    version: "3.5"
     description: qsTr("This plugin names notes as per your language setting")
     menuPath: "Plugins.Notes." + qsTr("Note Names (Interactive)")
     pluginType: "dock"
@@ -64,9 +64,11 @@ MuseScore {
     function getChordName(chord) {
         var text = "";
         var notes = chord.notes;
+        var sep = "\n";   // change to "," if you want them horizontally (anybody?)
         for (var i = 0; i < notes.length; i++) {
-            var sep = "\n";   // change to "," if you want them horizontally (anybody?)
-            if ( i > 0 )
+            if (!notes[i].visible)
+                continue // skip notes a that are not selected or invisible
+            if (text) // only if text isn't empty
                 text = sep + text; // any but top note
             if (typeof notes[i].tpc === "undefined") // like for grace notes ?!?
                 return;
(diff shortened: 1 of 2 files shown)


share/plugins/notenames.qml
```diff
@@ -34,8 +34,8 @@ MuseScore {
    function nameChord (notes, text, small) {
       var sep = "\n";   // change to "," if you want them horizontally (anybody?)
       for (var i = 0; i < notes.length; i++) {
-         if (!notes[i].visible)
-            continue // skip invisible notes
+         if ((curScore.selection.elements.length && !notes[i].selected) || !notes[i].visible)
+            continue // skip notes that are not selected or invisible
```
Jojo-Schmitz, line 38: @wizofaus: please double-check.

wizofaus: There are two issues as I pointed out on Discord - 1) if the first note in a chord is invisible or unselected but the 2nd isn't, it'll add a separator before the note name. 2) if there's no text to add for a particular chord (all notes invisible/unselected), it still creates an empty text element

Jojo-Schmitz: I don't see the 2nd issue. Maybe because an empty text is not visible?
And idea how to solve the 1st issue? Special case for 1st note?

wizofaus: I don't know how you'd "see" that a empty text element is created, but programmatically it clearly is still happening (accepted, Musescore might have some special logic to ignore a request to add an empty text element, but that seems unlikely).
I suggested two possible ways to solve the leading separator issue on Discord (either only add if there's already text, or always add and strip off leading separators at the end).

share/plugins/notenames-interactive.qml
```diff
@@ -64,9 +64,11 @@ MuseScore {
     function getChordName(chord) {
         var text = "";
         var notes = chord.notes;
+        var sep = "\n";   // change to "," if you want them horizontally (anybody?)
         for (var i = 0; i < notes.length; i++) {
-            var sep = "\n";   // change to "," if you want them horizontally (anybody?)
-            if ( i > 0 )
+            if ((curScore.selection.elements.length && !notes[i].selected) || !notes[i].visible)
+                continue // skip notes a that are not selected or invisible
+            if (text.text) // only if text isn't empty
```
wizofaus, line 71: Just `if (text)` here

Jojo-Schmitz: OK, this here works too though

Jojo-Schmitz: Ah, only for this plugin, right, thanks for catching!

wizofaus (approved):

RomanPudashkin (approved):

PR #8147, opened 2021-05-24
[MU4] Handling of Parts and Excerpt in instrument panel

njvdberg:
This PR handles `Parts` and `Excerpts` in the instrument panel.

In this PR a new algorithm for `Staff` sorting is implemented. This algorithm now uses the existing `Score::sortStaves()` method. 
If `Parts` are moved the staves of the part are set in the right order, then `Score::sortStaves()` will sort the `Parts` accordingly.

`Parts` can be moved using the Instruments Form or by dragging the `Parts` in the Instrument Panel. Both option now use the same algorithm or ordering the staves and parts.

Also undo/redo is now fully functional for inserting/removing/moves `Parts`. If undoing a change, the Instrument Panel is now updated to reflect the situation after the the undo (or redo).

Next, this PR have some changes in `Excerpt` handling.
 -1) If reading a score and the score has an `Excerpt` defined for a `Part`, no new `Excerpt` is created for that `Part`.
 -2) If a `Part` is removed, it is it removed from all `Excerpts` as well.
 -3) If removing a `Part` results in an empty `Excerpt`, this `Excerpt` is removed too.
 -4) All `Excerpt` handling is done via undo/redo. Undo removing a `Part` which led to removing of an `Excerpt` will insert both `Part` and `Excerpt`.
 -5) Any change in `Part` order will update the track mapping between tracks in the master score and the corresponding tracks in the `Excerpts`.

**Remark**
Please note this PR builds on top of PR #8099, this PR is required for this PR.
Also this PR is required for (the upcoming) score ordering in MU4.

### DIFF
--- a/src/engraving/libmscore/excerpt.cpp
+++ b/src/engraving/libmscore/excerpt.cpp
@@ -109,6 +109,22 @@ bool Excerpt::isEmpty() const
     return partScore() ? partScore()->parts().empty() : true;
 }
 
+void Excerpt::removePart(const QString& id)
+{
+    int index = 0;
+    for (Part* part: parts()) {
+        if (part->id() == id) {
+            break;
+        }
+        ++index;
+    }
+    if (index >= _parts.size()) {
+        return;
+    }
+
+    partScore()->undoRemovePart(partScore()->parts().at(index));
+}
+
 //---------------------------------------------------------
 //   read
 //---------------------------------------------------------
@@ -207,7 +223,6 @@ void Excerpt::createExcerpt(Excerpt* excerpt)
     // Set instruments and create linked staves
     for (const Part* part : parts) {
         Part* p = new Part(score);
-        p->setId(part->id());
         p->setInstrument(*part->instrument());
         p->setPartName(part->partName());
 
@@ -412,13 +427,17 @@ void MasterScore::deleteExcerpt(Excerpt* excerpt)
     undo(new RemoveExcerpt(excerpt));
 }
 
-void MasterScore::initExcerpt(Excerpt* excerpt)
+void MasterScore::initExcerpt(Excerpt* excerpt, bool fakeUndo)
 {
     Score* score = new Score(masterScore());
     excerpt->setPartScore(score);
     score->style().set(Sid::createMultiMeasureRests, true);
-    auto excerptCmdFake = new AddExcerpt(excerpt);
-    excerptCmdFake->redo(nullptr);
+    auto excerptCmd = new AddExcerpt(excerpt);
+    if (fakeUndo) {
+        excerptCmd->redo(nullptr);
+    } else {
+        score->undo(excerptCmd);
+    }
     Excerpt::createExcerpt(excerpt);
 }
 
--- a/src/engraving/libmscore/undo.cpp
+++ b/src/engraving/libmscore/undo.cpp
@@ -1171,9 +1171,18 @@ MapExcerptTracks::MapExcerptTracks(Score* s, QList<int> l)
      *    For the "undo" all staves which value -1 are *not* remapped since
      *    it is assumed this staves are removed later.
      */
+    int maxIndex = 0;
+    for (int i: l) {
+        maxIndex = std::max(i, maxIndex);
+    }
+
+    for (int i = 0; i <= maxIndex; ++i) {
+        rlist.append(-1);
+    }
+
     for (int i = 0; i < l.size(); ++i) {
         if (l[i] >= 0) {
-            rlist.insert(l[i], i);
+            rlist.replace(l[i], i);
         }
     }
     list = l;
@@ -2133,12 +2142,22 @@ void AddExcerpt::redo(EditData*)
 }
 
 //---------------------------------------------------------
+//   RemoveExcerpt
+//---------------------------------------------------------
+
+RemoveExcerpt::RemoveExcerpt(Excerpt* ex)
+    : excerpt(ex)
+{
+    index = excerpt->oscore()->excerpts().indexOf(excerpt);
+}
+
+//---------------------------------------------------------
 //   RemoveExcerpt::undo()
 //---------------------------------------------------------
 
 void RemoveExcerpt::undo(EditData*)
 {
-    excerpt->oscore()->addExcerpt(excerpt);
+    excerpt->oscore()->addExcerpt(excerpt, index);
 }
 
 //---------------------------------------------------------
--- a/src/instruments/view/instrumentspaneltreemodel.cpp
+++ b/src/instruments/view/instrumentspaneltreemodel.cpp
@@ -42,7 +42,17 @@ InstrumentsPanelTreeModel::InstrumentsPanelTreeModel(QObject* parent)
 {
     context()->currentMasterNotationChanged().onNotify(this, [this]() {
         IMasterNotationPtr masterNotation = context()->currentMasterNotation();
-        m_masterNotationParts = masterNotation ? masterNotation->parts() : nullptr;
+
+        m_masterNotationParts = nullptr;
+
+        if (masterNotation) {
+            m_masterNotationParts = masterNotation->parts();
+            m_masterNotationParts->partsChanged().onNotify(this, [this]() {
+                if (m_notationParts) {
+                    load();
+                }
+            });
+        }
     });
 
     context()->currentNotationChanged().onNotify(this, [this]() {
@@ -98,6 +108,10 @@ void InstrumentsPanelTreeModel::deleteItems()
 
 void InstrumentsPanelTreeModel::load()
 {
+    if (m_isLoadingBlocked) {
+        return;
+    }
+
     TRACEFUNC;
     beginResetModel();
 
@@ -219,11 +233,13 @@ bool InstrumentsPanelTreeModel::removeRows(int row, int count, const QModelIndex
         parentItem = m_rootItem;
     }
 
+    m_isLoadingBlocked = true;
     beginRemoveRows(parent, row, row + count - 1);
 
     parentItem->removeChildren(row, count, true);
 
     endRemoveRows();
+    m_isLoadingBlocked = false;
 
     emit isEmptyChanged();
 
@@ -238,6 +254,8 @@ AbstractInstrumentsPanelTreeItem* InstrumentsPanelTreeModel::modelIndexToItem(co
 bool InstrumentsPanelTreeModel::moveRows(const QModelIndex& sourceParent, int sourceRow, int count, const QModelIndex& destinationParent,
                                          int destinationChild)
 {
+    m_isLoadingBlocked = true;
+
     AbstractInstrumentsPanelTreeItem* sourceParentItem = modelIndexToItem(sourceParent);
     AbstractInstrumentsPanelTreeItem* destinationParentItem = modelIndexToItem(destinationParent);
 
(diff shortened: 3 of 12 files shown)


src/libmscore/undo.h
```diff
@@ -256,6 +256,22 @@ class RemovePart : public UndoCommand
     UNDO_NAME("RemovePart")
 };
 
+//---------------------------------------------------------
+//   SetSoloist
+//---------------------------------------------------------
+
+class SetSoloist : public UndoCommand
+{
+    Part* part;
```
RomanPudashkin, line 265: Please, add default values here:

Part* part = nullptr;
bool soloist = false;

njvdberg:
Solved a last minute crash, PR is now finished.

Tantacrul:
I take it I won't be able to see any functional difference yet, correct?

If not, is there anything you'd like me to test with this?

njvdberg:
> I take it I won't be able to see any functional difference yet, correct?

Correct, it's mainly the groundwork to make score ordering work.

njvdberg:
Just another rebase.

njvdberg:
Yet another rebase.

src/engraving/libmscore/excerpt.cpp
```diff
@@ -207,7 +223,6 @@ void Excerpt::createExcerpt(Excerpt* excerpt)
     // Set instruments and create linked staves
     for (const Part* part : parts) {
         Part* p = new Part(score);
-        p->setId(part->id());
```
RomanPudashkin, line 210: Can you please explain the reason for removing this line?

src/engraving/libmscore/excerpt.cpp
```diff
@@ -412,13 +427,17 @@ void MasterScore::deleteExcerpt(Excerpt* excerpt)
     undo(new RemoveExcerpt(excerpt));
 }
 
-void MasterScore::initExcerpt(Excerpt* excerpt)
+void MasterScore::initExcerpt(Excerpt* excerpt, bool fakeUndo)
```
RomanPudashkin, line 430: I don't quite understand the meaning of this flag. Can you please explain it to me?

src/engraving/libmscore/undo.cpp
```diff
@@ -2132,13 +2141,23 @@ void AddExcerpt::redo(EditData*)
...
 }
 
+//---------------------------------------------------------
+//   RemoveExcerpt
+//---------------------------------------------------------
+
+RemoveExcerpt::RemoveExcerpt(Excerpt* ex)
+    : excerpt(ex)
+{
+    index = excerpt->oscore()->excerpts().indexOf(excerpt);
```
RomanPudashkin, line 2151: It would be great to add a check on null here:

if (excerpt && execrpt->oscore()) {
index = excerpt->oscore()->excerpts().indexOf(excerpt);
}

src/instruments/view/instrumentspaneltreemodel.cpp
```diff
@@ -42,7 +42,16 @@ InstrumentsPanelTreeModel::InstrumentsPanelTreeModel(QObject* parent)
 {
     context()->currentMasterNotationChanged().onNotify(this, [this]() {
         IMasterNotationPtr masterNotation = context()->currentMasterNotation();
-        m_masterNotationParts = masterNotation ? masterNotation->parts() : nullptr;
+        if (masterNotation) {
```
RomanPudashkin, line 45: We can slightly simplify this condition:

m_masterNotationParts = nullptr;

if (masterNotation) {
m_masterNotationParts = masterNotation->parts();
m_masterNotationParts->partsChanged().onNotify(this, [this]() { ... }); 
}

It helps to remove the nested else block

src/instruments/view/instrumentspaneltreemodel.cpp
```diff
@@ -219,11 +232,13 @@ bool InstrumentsPanelTreeModel::removeRows(int row, int count, const QModelIndex
         parentItem = m_rootItem;
     }
 
+    m_isLoadingBlocked = true;
```
RomanPudashkin, line 235: Please explain the reason for this flag. InstrumentsPanelTreeModel::load() should be called before any functions in this model

src/notation/internal/notationparts.cpp
```diff
@@ -1263,26 +1320,33 @@ void NotationParts::updateSoloist(const PartInstrumentList& parts)
...
+void NotationParts::sortParts(const PartInstrumentList& parts, const QList<Ms::Staff*>& originalStaves)
 {
     Q_ASSERT(score()->parts().size() == static_cast<int>(parts.size()));
 
-    for (int i = 0; i < parts.size(); ++i) {
-        const Part* currentPart = score()->parts().at(i);
+    QList<int> staffMapping;
+    QList<int> trackMapping;
+    int idx2 = 0;
+    bool sort = false;
```
RomanPudashkin, line 1330: I think a better name is
needSort

src/notation/internal/notationparts.cpp
```diff
@@ -1263,26 +1320,33 @@ void NotationParts::updateSoloist(const PartInstrumentList& parts)
...
 
-        if (currentPart->id() == parts.at(i).partId) {
-            continue;
+    int partIndex = 0;
+    for (const PartInstrument& pi: parts) {
+        Ms::Part* currentPart = pi.isExistingPart ? part(pi.partId) : masterScore()->parts()[partIndex];
```
RomanPudashkin, line 1334: I think you need to check that masterScore()->parts() contains partIndex, because we can get SIGSEVG here

Like this:

QList<Part*> masterParts = masterScore()->parts()
for (const PartInstrument& pi: parts) {
if (partIndex >= masterParts.size()) {
...
}
...
}

src/notation/internal/notationparts.cpp
```diff
@@ -1263,26 +1320,33 @@ void NotationParts::updateSoloist(const PartInstrumentList& parts)
...
-void NotationParts::sortParts(const PartInstrumentList& parts)
+void NotationParts::sortParts(const PartInstrumentList& parts, const QList<Ms::Staff*>& originalStaves)
 {
     Q_ASSERT(score()->parts().size() == static_cast<int>(parts.size()));
 
-    for (int i = 0; i < parts.size(); ++i) {
-        const Part* currentPart = score()->parts().at(i);
+    QList<int> staffMapping;
+    QList<int> trackMapping;
+    int idx2 = 0;
```
RomanPudashkin, line 1329: Probably we can rename this variable. The current name doesn't look too obviously

src/engraving/libmscore/excerpt.cpp
njvdberg: `Excerpt::parts()` returns a list of pointers to `Part`s in the master score and for each part an new `Part` is created in the `Excerpt` score. This is **not** the same `Part` as the one in the master score, hence it should have its own, unique, id.

src/engraving/libmscore/excerpt.cpp
njvdberg: In the original version of `MasterScore::initExcerpt()` it was not possible to undo creating an `Excerpt` but undo/redo creating an `Excerpt` is required in several cases.

src/engraving/libmscore/undo.cpp
njvdberg: In general, none of the classes in `undo.cpp` have this kind of checks, it's the responsibility of the calling classes to make sure the arguments are correct. Which makes sense because these classes are used to do something via undo/redo instead of doing it straight away. In this case, the code come to the conclusion an `Excerpt` has to be removed but in stead of simply calling `masterScore->removeExcerpt(excerpt)` it uses `masterScore->undo(new RemoveExcerpt(excerpt)`. So sanity tests on arguments are expected to done **before** a new `RemoveExcerpt` object is created.

src/instruments/view/instrumentspaneltreemodel.cpp
njvdberg: This is a relic of my past where every `if` should have an `else` :-).
But I will change it.

src/notation/internal/notationparts.cpp
njvdberg: It isn't obvious at all but just copy of original 3.x code where apparently people suffered from find a decent name :smiley: 
I'll give it another try,

src/notation/internal/notationparts.cpp
njvdberg: Good catch :+1:
This is supposed to be checked by the assert in line 1325 but there `score()` is used instead of `masterScore()`. Until now this is not a problem since `sortParts()` is called for main scores only. But when this method is used for sorting parts in an excerpt, then `score()` and `masterScore()` are no longer the same. Also the score in `sortParts()` should be same as the score of the passed staves.
So, instead of using `score()` or `masterScore()` and testing whether this score is the same as (the first stave of) `originalStaves`, `originalstaves[0]->score()` could be used and check on empty `origialStaves` only.

src/notation/internal/notationparts.cpp
njvdberg: Agreed.

src/instruments/view/instrumentspaneltreemodel.cpp
njvdberg: Problem here is `InstrumentsPanelTreeModel::load()` will be called everything parts are changed, so also when parts are removed or moved.
This flag prevents the model to be reset while rows are moved/removed.

src/instruments/view/instrumentspaneltreemodel.cpp
njvdberg: Done

src/notation/internal/notationparts.cpp
njvdberg: Done.

src/notation/internal/notationparts.cpp
njvdberg: Done.

src/notation/internal/notationparts.cpp
njvdberg: Done.

src/engraving/libmscore/excerpt.cpp
```diff
@@ -109,6 +109,22 @@ bool Excerpt::isEmpty() const
...
+    int index = 0;
+    for (Part* part: parts()) {
+        if (part->id() == id) {
+            break;
+        }
+        ++index;
+    }
+    if (index >= _parts.size()) {
+        return;
+    }
```
wizofaus, line 123: Maybe cleaner as.
```
auto iter= std::find_if(parts().begin(), parts().end(), [id](auto p) { return p->id() == id; }); 
if (iter != parts.end())
  partScore()->undoRemovePart(partScore()->parts().at(std::distance(parts.begin(), iter)));
```
?
But can't you walk the parts in partScore()->parts() to find the one with the matching id?  Relying on indices matching up between two collections is always fraught!

wizofaus:
It doesn't appear this addresses issues in the panel itself where expanded tree nodes automatically collapse again after any refresh (which is pretty annoying)?  That was one of the reasons I gave up working on it!

wizofaus:
I'm still seeing that when you delete an instrument in the instrument panel, it just moves down to the bottom of the list (I believe because it's still internally referred to by an excerpt)

RomanPudashkin (approved):

Issue #8044, opened 2021-05-04
[MU4 Issue] Wrong key for transposing instruments when adding to the score later


njvdberg:
**Describe the bug**
When adding transposing instruments at an existing score the newly added instruments don't show the correct key.

**To Reproduce**
1. Create a new score with one Clarinet
2. Finish the score

The score look like:
![Screenshot_20210504_144112](https://user-images.githubusercontent.com/48352523/117004800-c7d4ce00-ace6-11eb-9718-a4551b464633.png)
Since score is in C major, the Clarinet will show 2 sharps (which is correct).

3. Open the `Instruments` form by pressing `I`.
4. Add a new Clarinet and close the form.

The score will now looks like:
![Screenshot_20210504_144331](https://user-images.githubusercontent.com/48352523/117005063-1a15ef00-ace7-11eb-9882-4a7b025cbb25.png)
The new Clarinet is missing the 2 sharps.

**Expected behavior**
The score should looks like
![Screenshot_20210504_144457](https://user-images.githubusercontent.com/48352523/117005244-4d587e00-ace7-11eb-9afa-a419e886158e.png)
with both Clarinets the same key.

**Desktop (please complete the following information):**
 - OS: Linux (openSUSE Tumbleweed)

**Additional context**
Observe in MuseScoreNightly-202105030415-master-58e2cbd-x86_64.AppImage

njvdberg:
This should be part of [MU4 - INSTRUMENTS_PANEL](https://github.com/musescore/MuseScore/projects/8).

Tantacrul:
This has been bugging me for ages.

njvdberg:
Will be solved with PR #8149.

PR #8149, opened 2021-05-24
[MU4] Correct key signature when adding transposing instruments.

njvdberg:
Resolves: https://github.com/musescore/MuseScore/issues/8044

When adding a new Part, adjust the key signature for transposing instruments.

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -2765,6 +2765,46 @@ void Score::adjustKeySigs(int sidx, int eidx, KeyList km)
 }
 
 //---------------------------------------------------------
+//   getKeyList
+//      This is taken from MuseScore::editInstrList()
+//---------------------------------------------------------
+
+KeyList Score::keyList() const
+{
+    // find the keylist of the first pitched staff
+    KeyList tmpKeymap;
+    Staff* firstStaff = nullptr;
+    for (Staff* s : masterScore()->staves()) {
+        if (!s->isDrumStaff(Fraction(0, 1))) {
+            KeyList* km = s->keyList();
+            tmpKeymap.insert(km->begin(), km->end());
+            firstStaff = s;
+            break;
+        }
+    }
+
+    Key normalizedC = Key::C;
+    // normalize the keyevents to concert pitch if necessary
+    if (firstStaff && !masterScore()->styleB(Ms::Sid::concertPitch) && firstStaff->part()->instrument()->transpose().chromatic) {
+        int interval = firstStaff->part()->instrument()->transpose().chromatic;
+        normalizedC = transposeKey(normalizedC, interval);
+        for (auto i = tmpKeymap.begin(); i != tmpKeymap.end(); ++i) {
+            int tick = i->first;
+            Key oKey = i->second.key();
+            tmpKeymap[tick].setKey(transposeKey(oKey, interval));
+        }
+    }
+
+    // create initial keyevent for transposing instrument if necessary
+    auto i = tmpKeymap.begin();
+    if (i == tmpKeymap.end() || i->first != 0) {
+        tmpKeymap[0].setKey(normalizedC);
+    }
+
+    return tmpKeymap;
+}
+
+//---------------------------------------------------------
 //   cmdRemoveStaff
 //---------------------------------------------------------
 
--- a/src/notation/internal/notationparts.cpp
+++ b/src/notation/internal/notationparts.cpp
@@ -1217,6 +1217,13 @@ void NotationParts::appendStaves(Part* part, const mu::instruments::Instrument& 
 
         insertStaff(staff, staffIndex);
     }
+
+    if (!part->nstaves()) {
+        return;
+    }
+    int firstStaffIndex = part->staff(0)->idx();
+    int endStaffIndex = firstStaffIndex + part->nstaves();
+    masterScore()->adjustKeySigs(firstStaffIndex, endStaffIndex, masterScore()->keyList());
 }
 
 void NotationParts::removeMissingInstruments(const InstrumentList& instruments)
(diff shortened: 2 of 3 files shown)


src/libmscore/score.cpp
```diff
@@ -2750,6 +2750,46 @@ void Score::adjustKeySigs(int sidx, int eidx, KeyList km)
     }
 }
 
+//---------------------------------------------------------
+//   getKeyList
+//      This is taken from MuseScore::editInstrList()
```
wizofaus, line 2755: What does this mean, there is no such function?

src/libmscore/score.cpp
```diff
@@ -2750,6 +2750,46 @@ void Score::adjustKeySigs(int sidx, int eidx, KeyList km)
...
+//---------------------------------------------------------
+
+KeyList Score::getKeyList() const
+{
+    // find the keylist of the first pitched staff
+    KeyList tmpKeymap;
```
wizofaus, line 2761: If it's a KeyList does it make sense to call the variable "tmpKeymap"?  Even "tmp" doesn't seem quite right either, given it's the returned result of the function (even if that's by creating a copy).

src/notation/internal/notationparts.cpp
```diff
@@ -22,6 +22,7 @@
 #include "notationparts.h"
 
 #include "libmscore/score.h"
+#include "libmscore/keylist.h"
```
wizofaus, line 25: Needed?

wizofaus:
Sorry, new to this, don't mean to be presumptuous but just trying to help out - seems like something it might be worth adding a unit test for if possible?

src/libmscore/score.cpp
njvdberg: Not anymore, but there was such code in `instrdialog.cpp` in MuseScore 3.x. This dialog is replaced by a new one but this code is still required.

src/libmscore/score.cpp
njvdberg: True, but as the comment said, it is just copied from 3.x code. Changing too much makes it unclear to track where this code original came from.

njvdberg:
| Sorry, new to this, don't mean to be presumptuous

Don't worry, it is certainly to taken as such. All comments are welcome!
Regarding the unit test, this code is part of the UI. I don't know whether we have unit tests for the UI or how to create them...

wizofaus:
> > Sorry, new to this, don't mean to be presumptuous
> 
> Don't worry, it is certainly to taken as such. All comments are welcome!
> Regarding the unit test, this code is part of the UI. I don't know whether we have unit tests for the UI or how to create them...

Wouldn't (shouldn't?) these changes be triggered even when programmatically modifying a score?  I.e. they'd execute and affect the construction of the score even if there was no UI as such...

src/notation/internal/notationparts.cpp
njvdberg: Not anymore, good catch.
In the first version the code which is now in `Score::getKeyList()` was in `NotationParts::appendStaves()` and then this line was necessary. When moved to a dedicated method this line wasn't removed.

njvdberg:
I don't know. On one hand is good to have as much as possible covered by unit tests but on the other hand I think you can also have too many tests. It is difficult to prevent an overlap between the tests. Also, when creating such a test, how do we know it is not testing a kind of "academic" situation while real life behavior isn't tested?

I don't have a real opinion on this and am open for any suggestions on that is the best thing to do here.

njvdberg:
Just another rebase.

src/engraving/libmscore/score.cpp
```diff
@@ -2764,6 +2764,46 @@ void Score::adjustKeySigs(int sidx, int eidx, KeyList km)
     }
 }
 
+//---------------------------------------------------------
+//   getKeyList
+//      This is taken from MuseScore::editInstrList()
+//---------------------------------------------------------
+
+KeyList Score::getKeyList() const
```
RomanPudashkin, line 2772: Better to rename this function to keyList() (code style is the reason)

src/engraving/libmscore/score.cpp
```diff
@@ -2764,6 +2764,46 @@ void Score::adjustKeySigs(int sidx, int eidx, KeyList km)
...
+{
+    // find the keylist of the first pitched staff
+    KeyList tmpKeymap;
```
RomanPudashkin, line 2775: Probably "result" will be a better name for this but not important

src/engraving/libmscore/score.cpp
```diff
@@ -2764,6 +2764,46 @@ void Score::adjustKeySigs(int sidx, int eidx, KeyList km)
...
+    Staff* firstStaff = 0;
```
RomanPudashkin, line 2776: Please, use nullptr instead of 0

src/notation/internal/notationparts.cpp
```diff
@@ -1217,6 +1217,10 @@ void NotationParts::appendStaves(Part* part, const mu::instruments::Instrument&
 
         insertStaff(staff, staffIndex);
     }
+
+    int firstStaffIndex = part->staff(0)->idx();
```
RomanPudashkin, line 1221: Probably you need to add a check for emptiness here:

if (part->staves().isEmpty()) {
return;
}
int firstStaffIndex = ...
...

src/engraving/libmscore/score.cpp
njvdberg: Done.

src/engraving/libmscore/score.cpp
njvdberg: Done.

src/notation/internal/notationparts.cpp
njvdberg: This will only happen if the `Instrument` has no staves which is wrong in the first place (instruments have at least one staff).
However since methods in `NotationParts` are a kind of "high level" methods, a tests might do no harm here.
Done.

RomanPudashkin (approved):

Issue #8148, opened 2021-05-24
[MU4 Task] Lasso selection not hooked up


wizofaus:
The Lasso selection (shift+click & drag with mouse) isn't hooked up in MU4.
Should be relatively simple to do so (the Lasso class etc. is still in place), and needed in order to add a proposed enhancement to allow "list/distinct" selection by using CTRL+drag (create a selection with just the distinct elements fully inside the lasso rectangle).

wizofaus:
https://github.com/musescore/MuseScore/pull/8157

wizofaus:
This is broken again now...not sure if it should be re-opened? It some times sort of works but never draws the blue rectangle around the final selection, and most of the time doesn't seem to work at all.

wizofaus:
So I think *most* of the issues I'm seeing are to do with the mouse position offset (that happens for me on every build because of the transparent frame problem), but even so it seems far more difficult to use the lasso than it should be.

PR #8157, opened 2021-05-25
[MU4] Enabled lasso selection

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8148

Lasso selection functionality from MU3 wasn't implemented in MU4.

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -49,6 +49,7 @@
 #include "libmscore/navigate.h"
 #include "libmscore/keysig.h"
 #include "libmscore/instrchange.h"
+#include "libmscore/lasso.h"
 
 #include "masternotation.h"
 #include "scorecallbacks.h"
@@ -60,7 +61,7 @@
 using namespace mu::notation;
 
 NotationInteraction::NotationInteraction(Notation* notation, INotationUndoStackPtr undoStack)
-    : m_notation(notation), m_undoStack(undoStack)
+    : m_notation(notation), m_undoStack(undoStack), m_lasso(new Ms::Lasso(notation->score()))
 {
     m_noteInput = std::make_shared<NotationNoteInput>(notation, this, m_undoStack);
     m_selection = std::make_shared<NotationSelection>(notation);
@@ -140,6 +141,9 @@ void NotationInteraction::paint(mu::draw::Painter* painter)
     drawTextEditMode(painter);
     drawSelectionRange(painter);
     drawGripPoints(painter);
+    if (!m_lasso->bbox().isEmpty()) {
+        m_lasso->draw(painter);
+    }
 }
 
 INotationNoteInputPtr NotationInteraction::noteInput() const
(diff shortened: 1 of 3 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -581,6 +584,26 @@ void NotationInteraction::startDrag(const std::vector<Element*>& elems,
     }
 }
 
+void NotationInteraction::doDragLasso(const QPointF& pt)
+{
+    score()->addRefresh(m_lasso->canvasBoundingRect());
+    QRectF r;
+    r.setCoords(m_dragData.beginMove.x(), m_dragData.beginMove.y(), pt.x(), pt.y());
+    m_lasso->setbbox(r.normalized());
+    QRectF _lassoRect(m_lasso->bbox());
```
wizofaus, line 593: can remove this line

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -431,7 +437,8 @@ void NotationViewInputController::startDragElements(ElementType elementsType, co
 
 void NotationViewInputController::mouseReleaseEvent(QMouseEvent*)
 {
-    if (!m_interactData.hitElement && !m_isCanvasDragged && !m_view->notationInteraction()->isGripEditStarted()) {
+    if (!m_interactData.hitElement && !m_isCanvasDragged && !m_view->notationInteraction()->isGripEditStarted() && 
+        !m_view->notationInteraction()->isDragStarted()) {
```
wizofaus, line 441: Can't seem to find any case where adding this causes unexpected behaviour and it's definitely necessary for lasso selection to work.

igorkorsukov:
To understand why this functionality is needed?
Why not just use a lot of different tools to create screenshots?
In this case, a screenshot can be taken from the Publish page, where service symbols will not be visible (when this page will be implemented)

wizofaus:
> To understand why this functionality is needed?
> Why not just use a lot of different tools to create screenshots?
> In this case, a screenshot can be taken from the Publish page, where service symbols will not be visible (when this page will be implemented)

It's in MuseScore 3 and is far and away the most intuitive way of selecting a passage of music, and I'm pretty sure all the other popular notation packages out there have the same sort of functionality.  I actually have plans to improve it further so it can do "list"/distinct selections too, not just ranges.
I don't know what you mean by screenshots, this is to do with live selection?

igorkorsukov:
@Tantacrul What do you think of this functionality?

cbjeukendrup:
@igorkorsukov You might be confusing the "image capture mode" (which is making screenshots, and will indeed be in the publish page, as far as I know) 
and the lasso selection, which is the ability to select music in an area by dragging a box with the mouse. 

(Here below I show them both)

https://user-images.githubusercontent.com/48658420/119507425-21bd3680-bd6f-11eb-876e-fe585d2ddceb.mov

igorkorsukov:
> @igorkorsukov You might be confusing the "image capture mode" (which is making screenshots, and will indeed be in the publish page, as far as I know)
> and the lasso selection, which is the ability to select music in an area by dragging a box with the mouse.
> 
> (Here below I show them both)
> 
>  Schermopname.2021-05-25.om.15.36.47.mov

@cbjeukendrup @wizofaus Oh, ah, I really went confusing, thanks !!

RobFog:
I can't do without this functionality.

Tantacrul:
This is so cool. I actually realised the other day this was missing :)

Issue #8084, opened 2021-05-11
[MU4 Issue] Pencil icon dropdown has two ticks instead of one


vshalkevich:
**Describe the bug**
Pencil icon dropdown always has two ticks
In this dropdown only one item can have a tick

**To Reproduce**
1. Create a score
2. Click on 'Pencil' icon from the Note Input Bar and hold until dropdown appears
3. Click on any item from the dropdown
4.  Click on 'Pencil' icon from the Note Input Bar and hold until dropdown appears again
5. observe what item ticked

**Expected behavior**
In this dropdown only one item should have a tick

**Screenshots**
![Double tick](https://user-images.githubusercontent.com/71372310/117880495-db68d180-b2b0-11eb-862f-d2d4be0df3bb.jpg)

**Desktop (please complete the following information):**
Windows 10

**Additional context**
In some cases dropdown items don’t aligned. Can't reproduce now

cbjeukendrup:
Working on this!

PR #8173, opened 2021-05-25
Allow right-clicking as alternative to long-clicking in NoteInputBar

cbjeukendrup:
Allow right-clicking as alternative to long-clicking in NoteInputBar on buttons that have both an action and a menu.
Also renamed `showSubitemsByPressAndHold` to `isMenuSecondary`, because the property means that triggering the action is the primary function of the button and opening the menu the secondary function.
Also enhanced (accent) color behaviour.
Oh -- and also avoid some "unqualified lookups".

Here's a screen recording of the new behaviour, but you can't judge very well from this because you can't know which mouse buttons I'm using. 

https://user-images.githubusercontent.com/48658420/119560446-42ec4a00-bda4-11eb-83a7-e6a888a3f10e.mov

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
@@ -50,8 +50,8 @@ FocusScope {
 
     property bool isClickOnKeyNavTriggered: true
 
-    signal clicked()
-    signal pressAndHold()
+    signal clicked(var mouse)
+    signal pressAndHold(var mouse)
 
     objectName: root.text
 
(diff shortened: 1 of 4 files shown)


Eism:
@cbjeukendrup do we really need this feature?
Who did the request for this feature come from?

cbjeukendrup:
I brought up this topic a while ago, here: https://discord.com/channels/818804595450445834/818848561047273544/829749779893583903
I believe @shoogle reacted positively on this, and personally I think it feels more streamlined.

(By the way, I'll solve the merge conflict later today anyway.)

vpereverzev:
@cbjeukendrup can't see attached video

cbjeukendrup:
> @cbjeukendrup can't see attached video

That's strange. Let's try a new one: 

https://user-images.githubusercontent.com/48658420/120219048-e7eea300-c23a-11eb-8a11-35ff4b9b4a34.mov

Tantacrul:
I don't have any issues with the idea:

Question: is this a known bug: I select a different option and the 'Step Time' option remains selected (or at least still has a tick next to it)

image


Second: when long pressing or right-clicking on this button (and triggering the popup), it would be good to disable the tooltip, which ends up overlapping it.

Eism:
@Tantacrul second idea is already implemented on #8194

cbjeukendrup:
The issue about the double tick mark in the menu is reported already: https://github.com/musescore/MuseScore/issues/8084

Tantacrul:
Thought so. Thanks!

Eism (approved):

PR #8174, opened 2021-05-25
lasso list selection

wizofaus:
Resolves: Adding new feature to support lasso selection without converting to range

Current MU3 behaviour of always converting the list selection shown while drawing the lasso to a range is not always desirable or expected, e.g. if you need to select all notes between a certain pitch range.
This includes the changes in PR https://github.com/musescore/MuseScore/pull/8157 which are required.

### DIFF
--- a/src/notation/internal/notationinteraction.h
+++ b/src/notation/internal/notationinteraction.h
@@ -248,6 +248,7 @@ private:
         Ms::EditData ed;
         std::vector<Element*> elements;
         std::vector<std::unique_ptr<Ms::ElementGroup> > dragGroups;
+        DragMode mode { DragMode::BothXY };
         void reset();
     };
 
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -399,11 +399,12 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
 
         m_view->notationInteraction()->drag(m_interactData.beginPoint, logicPos, mode);
         return;
-    } else if (m_interactData.hitElement == nullptr && (event->modifiers() & Qt::ShiftModifier)) {
+    } else if (m_interactData.hitElement == nullptr && (keyState & (Qt::ShiftModifier | Qt::ControlModifier))) {
         if (!m_view->notationInteraction()->isDragStarted()) {
             m_view->notationInteraction()->startDrag(std::vector<Element*>(), QPoint(), [](const Element*) { return false; });
         }
-        m_view->notationInteraction()->drag(m_interactData.beginPoint, logicPos, DragMode::BothXY);
+        m_view->notationInteraction()->drag(m_interactData.beginPoint, logicPos,
+                                            keyState & Qt::ControlModifier ? DragMode::LassoList : DragMode::BothXY);
         return;
     }
 
--- a/src/libmscore/score.cpp
+++ b/src/libmscore/score.cpp
@@ -3719,7 +3719,7 @@ void Score::lassoSelect(const QRectF& bbox)
 //   lassoSelectEnd
 //---------------------------------------------------------
 
-void Score::lassoSelectEnd()
+void Score::lassoSelectEnd(bool convertToRange)
 {
     int noteRestCount     = 0;
     Segment* startSegment = 0;
@@ -3734,6 +3734,11 @@ void Score::lassoSelectEnd()
         return;
     }
     _selection.setState(SelState::LIST);
+
+    if (!convertToRange) {
+        setUpdateAll();
+        return;
+    }
 
     foreach (const Element* e, _selection.elements()) {
         if (e->type() != ElementType::NOTE && e->type() != ElementType::REST) {
--- a/src/libmscore/score.h
+++ b/src/libmscore/score.h
@@ -1028,7 +1028,7 @@ public:
     void setScoreOrder(ScoreOrder* order) { _scoreOrder = order; }
 
     void lassoSelect(const QRectF&);
-    void lassoSelectEnd();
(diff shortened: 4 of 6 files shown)


src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -399,6 +399,13 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
 
         m_view->notationInteraction()->drag(m_interactData.beginPoint, logicPos, mode);
         return;
+    } else if (m_interactData.hitElement == nullptr && (keyState & (Qt::ShiftModifier | Qt::ControlModifier))) {
+        if (!m_view->notationInteraction()->isDragStarted()) {
+            m_view->notationInteraction()->startDrag(std::vector<Element*>(), QPoint(), [](const Element*) { return false; });
+        }
+        m_view->notationInteraction()->drag(m_interactData.beginPoint, logicPos, 
+            (keyState & Qt::ControlModifier) ? DragMode::LassoList : DragMode::BothXY);
```
wizofaus, line 407: NB this means that holding both shift and control down results in a list selection too.

Jojo-Schmitz: (and is a minor code style issue, see the check that now ran)

Jojo-Schmitz:
Rebase needed (on top of a code style fix)

Jojo-Schmitz:
Still a rebase is needed to fix the merge conflicts

wizofaus:
> Still a rebase is needed to fix the merge conflicts

All the conflicts are being caused by commit squashing, as far as I can tell.  Git can't work out what individual commits were used in each branch.

src/notation/view/notationviewinputcontroller.cpp
wizofaus: > (and is a minor code style issue, see the check that now ran)

Seems uncrustify can't handle that syntax properly unless I remove the parenthesis.

src/notation/internal/notationinteraction.h
```diff
@@ -249,6 +249,7 @@ class NotationInteraction : public INotationInteraction, public async::Asyncable
         std::vector<Element*> elements;
         std::vector<std::unique_ptr<Ms::ElementGroup> > dragGroups;
         void reset();
+        DragMode mode;
```
igorkorsukov, line 252: It would be nice to initialize with a default value

wizofaus: I don't mind adding it but then it might look odd, all the other fields in the struct assume they're auto initialised with the default value...as it happens the field is meaningless until some of the other fields get set too.
But I do note that I added this field *after* a method declaration (`void reset()`), it's probably better to move it above it along with the others?

igorkorsukov: Other fields have (or should have) constructors, so they have default initialization.
Enums have no constructors and are accordingly initialized with a random value.
The code is changing all the time ... and, accordingly, a situation may accidentally arise when they will use it before the assignment of any value, and since the initialized value is random, this can lead to hard-to-reproduce errors.
Therefore, we always strive to initialize all fields with default values, for example, you can add an undefined value to the enum and initialize it.

wizofaus: Sorry, keep having to mentally adjust from C#-land, anyway seems there's a new syntax for declaring initializer values, which is used elsewhere so have done that (thought I'm not sure how it differs from just using = ).

src/notation/internal/notationinteraction.h
```diff
@@ -248,6 +248,7 @@ class NotationInteraction : public INotationInteraction, public async::Asyncable
         Ms::EditData ed;
         std::vector<Element*> elements;
         std::vector<std::unique_ptr<Ms::ElementGroup> > dragGroups;
+        DragMode mode { DragMode::BothXY };
```
igorkorsukov, line 251: We usually use initialization with `=` everywhere. It seems to be easier to read :) But it doesn't matter, initialization with `{}` is more universal.

igorkorsukov, line 251: We usually use initialization with `=` everywhere. It seems to be easier to read :) But it doesn't matter, initialization with `{}` is more universal.

wizofaus: The first file I checked was element.h  which uses brace initialization for everything...apparently the only difference is that assignment initialization can cause auto-conversions (e.g. -1 into an unsigned int), whereas for { } / brace initialization that's an error...

igorkorsukov: > The first file I checked was element.h which uses brace initialization for everything...apparently the only difference is that assignment initialization can cause auto-conversions (e.g. -1 into an unsigned int), whereas for { } / brace initialization that's an error...

Unfortunately, there are a lot of different styles of code in the `libmscore` dir now, because this is code from the old version.
Therefore, it is better to look elsewhere, with new code :)
Also we have code style documentation https://github.com/musescore/MuseScore/wiki/CodeGuidelines

wizofaus:
Thanks for merging, but note that this is a new feature that didn't exist in MU3, do we need an issue to track updating the documentation?

shoogle:
@wizofaus, feel free to create that issue over at https://musescore.org/en/project/issues/musescore using type "Wording/Translation" (not a perfect match but we don't really have need for documentation issues often enough to create a dedicated type for them).

wizofaus:
> @wizofaus, feel free to create that issue over at https://musescore.org/en/project/issues/musescore using type "Wording/Translation" (not a perfect match but we don't really have need for documentation issues often enough to create a dedicated type for them).

Done!

Jojo-Schmitz:
https://musescore.org/en/node/321719

PR #8176, opened 2021-05-26
[MU3 Backend] ENG-34: Fix erroneous space created for melismas

iveshenry18:
(Partially) Resolves: [ENG-34](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-34): Improve lyrics spacing

Previously, melismas were treated the same as normal Lyrics in terms of
horizontal spacing—they would cause a spacer to be added to the
ChordRest to which they were attached, causing an unnecessary gap after
the first note of a melisma.

This commit prevents a melisma from adding a right spacer to
their first ChordRest, and rather creates a right
spacer on the last ChordRest of a lyric if necessary.

In pursuit of this, this commit refactors ChordRest::_melismaEnd(s) from
a bool to a std::set<*Lyrics>, giving a ChordRest a way to access the
melismatic Lyrics that end on it. This change also solves the problem of
said bool being untrustworthy (i.e. erroneously being changed to false
when one of multiple melismas was removed).

### DIFF
--- a/libmscore/chordrest.cpp
+++ b/libmscore/chordrest.cpp
@@ -1243,19 +1243,9 @@ bool ChordRest::isMelismaEnd() const
 //    No-op and return false if Lyrics not present in _melismaEnds.
 //---------------------------------------------------------
 
-bool ChordRest::removeMelismaEnd(const Lyrics* l)
+void ChordRest::removeMelismaEnd(Lyrics* const l)
       {
-      bool found = false;
-      for (std::set<Lyrics*>::iterator it = _melismaEnds.begin(); it != _melismaEnds.end(); ) {
-            if (*it == l) {
-                  it = _melismaEnds.erase(it); // Returns next iterator
-                  found = true;
-                  }
-            else {
-                  ++it;
-                  }
-            }
-      return found;
+      _melismaEnds.erase(l);
       }
 
 //---------------------------------------------------------
--- a/libmscore/chordrest.h
+++ b/libmscore/chordrest.h
@@ -140,7 +140,7 @@ class ChordRest : public DurationElement {
       int lastVerse(Placement) const;
       const std::set<Lyrics*>& melismaEnds() const { return _melismaEnds; }
       std::set<Lyrics*>& melismaEnds()             { return _melismaEnds; }
-      bool removeMelismaEnd(const Lyrics* l);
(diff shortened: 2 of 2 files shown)


RobFog:
And another for @lvinken 😉

oktophonie (approved):

libmscore/chordrest.cpp
```diff
@@ -1232,20 +1231,31 @@ QString ChordRest::accessibleExtraInfo() const
...
+//   removeMelismaEnd()
+//    removes a Lyrics object from the _melismaEnds set
+//    if present by iterating through the set.
+//    Returns a boolean representing whether the Lyrics
+//    was found.
+//    No-op and return false if Lyrics not present in _melismaEnds.
 //---------------------------------------------------------
 
-void ChordRest::setMelismaEnd(bool v)
+bool ChordRest::removeMelismaEnd(const Lyrics* l)
```
vpereverzev, line 1246: just use the combination of "std::set<Key,Compare,Allocator>::find" and "std::set<Key,Compare,Allocator>::erase"

I'm also not sure if whether it makes sense to store it in std::set, what was the point?

iveshenry18: Good call; I honestly can't recall why I made this so overly complex.
The reason I'm storing the values is this: previously there was a _melismaEnd boolean that stored whether a ChordRest had the end of a melisma, but this actually was unreliable if there were multiple lyrics. Additionally, this spacing fix required that a ChordRest have pointers to any lyrics/melismas that end on it in order to create a properly-sized spacer. Since I needed to store those pointers and there already was an existing melisma-end data member, I basically made a drop-in replacement for the existing _melismaEnd functions that stored all pointers (realistically, this will be ~0–3 values) rather than just setting a boolean. I chose a std::set to avoid redundancy (since I didn't write the original code where they get added); although possibly an unordered_set would be quicker, or some other Qt container. Does that make sense? Let me know if I'm missing something

vpereverzev: ah, sorry, missed that one in my emails

prefer to use std::containers wherever it's possible within the libmscore
regarding your case, I think it's enough to stick with std::set then

vpereverzev (approved):

PR #8181, opened 2021-05-26
[MU3 Backend] Eng-35: MusicXML import Small Notes vs. Small Chord 

iveshenry18:
Resolves: [ENG-35](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-35): All elements of small notes should be reduced in size

Since MusicXML only stores note-level information about smallness
(\<type size='cue'\> and \<cue\>), chords that should be small were being
read as normal-sized chords of small notes. This caused issues with
the sizes of flags, ledger lines, and articulations.

This commit adds handleSmallness to ensure that a chord with all small
notes is imported as a small chord with all non-small notes, as well
as a test for this particular case.

For future discussion: it may be worth considering making the _small attribute of a Chord fully derived from its member Notes; leaving the two independent can create redundancy and other errors such as this one.

### DIFF
--- a/importexport/musicxml/importmxmlpass2.cpp
+++ b/importexport/musicxml/importmxmlpass2.cpp
@@ -3999,7 +3999,7 @@ static TDuration determineDuration(const bool rest, const QString& type, const i
 static Chord* findOrCreateChord(Score* score, Measure* m,
                                 const Fraction& tick, const int track, const int move,
                                 const TDuration duration, const Fraction dura,
-                                Beam::Mode bm)
+                                Beam::Mode bm, bool small)
       {
       //qDebug("findOrCreateChord tick %d track %d dur ticks %d ticks %s bm %hhd",
       //       tick, track, duration.ticks(), qPrintable(dura.print()), bm);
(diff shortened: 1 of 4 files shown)


oktophonie (approved):

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -4087,6 +4090,38 @@ static void handleDisplayStep(ChordRest* cr, int step, int octave, const Fractio
...
+
+/**
+ * Handle the distinction between small notes and a small 
+ * chord, to ensure a chord with all small notes is small.
+ * This also handles the fact that a small note being added
+ * to a small chord should not itself be small.
+ * I.e. a chord is "small until proven otherwise".
+ */
+
+static void handleSmallness(bool cueOrSmall, Note* note, Chord* c)
```
vpereverzev, line 4105: I think it makes sense to handle this logic directly in Chord class. Just because it might be useful not only for import functionality, but for the whole domain model

We can consider it in the further PRs

vpereverzev (approved):

importexport/musicxml/importmxmlpass2.cpp
iveshenry18: I agree! Taking it a step further, we could make the _small attribute of a Chord fully dependent on its member Notes (i.e. Chord::setSmall() just sets all the Notes to small, and Chord::small() reads whether all the Notes are small; Chord doesn't actually store a _small flag of its own). Not sure if that would introduce any regressions though. Just something to consider!

Jojo-Schmitz: Better leave such a change for the master branch only I guess.

Issue #8184, opened 2021-05-27
[MU4 Issue] Toggling concert pitch in Style dialog messes up key signatures


cbjeukendrup:
**Describe the bug**
Toggling concert pitch in Style dialog messes up key signatures. 

**To Reproduce**
1. Open the Format > Style dialog. 
2. Toggle "Display in concert pitch".
3. Press OK.

The key signatures on the score are totally wrong; when sharps or flats should appear, they sometimes disappear instead, or the other way round, or too many appear or disappear. 

**Expected behavior**
The result of the steps above should be exactly the same as toggling the concert pitch button in the status bar (which works fine). 

**Screenshots**

https://user-images.githubusercontent.com/48658420/119822368-dda76e80-bef3-11eb-83af-17e445eb41a5.mov

**Desktop (please complete the following information):**
 - OS: reported on macOS, probably other OSs too

njvdberg:
Solved with PR #8186.
Problem is the style editor fires the command to change to pitch twice so the transposing instruments are transposed twice.

PR #8186, opened 2021-05-27
Prevents switch to concert pitch/normal when score is already in concert/normal pitch.

njvdberg:
Resolves: https://github.com/musescore/MuseScore/issues/8184

When switching to concert/normal pitch, transposing instruments are transposed by `Score::cmdConcertPitchChanged()`.
However for some reason when switch the concert pitch within the style editor, the operation is fired twice. So also `Score::cmdConcertPitchChanged()` is called twice and a s result the transposing instruments are transposed twice.

E.g. for a piece in C, the Bb Clarinet is in D. Switching to concert pitch the Bb Clarinet is transposed a major second down, to C. But because the command is fired twice the Bb clarinet is transposed a major second down again resulting in Bb.
When next the tuning fork on the main GUI is used to switch back to normal pitch the Bb Clarinet is transposed a major second up, to C. But now the command is fired only once (as it should be) and the Bb Clarinet now stays into C, like any other non-transposing instrument. Which is wrong. 

With this PR `Score::cmdConcertPitchChanged()` checks first whether the new state is different as the existing state and will transpose transposing instruments when the state really changes.

### DIFF
--- a/src/libmscore/score.cpp
+++ b/src/libmscore/score.cpp
@@ -2864,6 +2864,10 @@ void Score::mapExcerptTracks(QList<int>& dst)
 
 void Score::cmdConcertPitchChanged(bool flag)
 {
+    if (flag == styleB(Ms::Sid::concertPitch)) {
+        return;
+    }
+
     undoChangeStyleVal(Sid::concertPitch, flag);         // change style flag
 
     for (Staff* staff : qAsConst(_staves)) {


cbjeukendrup (approved):
Nice, Niek, this works perfectly! That's quick!

RomanPudashkin (approved):

njvdberg:
I just had a look at transposing instruments and though this couldn't be that hard to solve. And that turned put to be a lucky guess :wink:

PR #8188, opened 2021-05-27
[MU4] Fix #321376: Musescore cannot open MusicXML file it created if filename contains an ampersand

Jojo-Schmitz:
Resolves: Resolves: https://musescore.org/en/node/321376

This is for master what #8187 is for 3.x


vpereverzev:
@iveshenry18 FYI

PR #8194, opened 2021-05-28
[MU4] Popups restrictions

Eism:
https://user-images.githubusercontent.com/10116828/119996524-241dcb80-bfcf-11eb-9c50-55f6106dba2c.mp4

### DIFF
--- a/src/framework/uicomponents/view/popupview.cpp
+++ b/src/framework/uicomponents/view/popupview.cpp
@@ -467,10 +467,10 @@ QRect PopupView::currentScreenGeometry() const
 {
     QScreen* currentScreen = QGuiApplication::screenAt(m_globalPos.toPoint());
     if (!currentScreen) {
-        return QRect();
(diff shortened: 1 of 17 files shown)


src/framework/uicomponents/view/popupview.cpp
```diff
@@ -405,14 +406,144 @@ void PopupView::setRet(QVariantMap ret)
...
 }
+
+QRect PopupView::currentScreenGeometry() const
+{
+    QScreen* currentScreen = QGuiApplication::screenAt(m_globalPos.toPoint());
+    if (!currentScreen) {
+        return QRect();
+    }
+
+    return currentScreen->geometry();
```
cbjeukendrup, line 473: How about using <code>**available**Geometry()</code> here instead, to avoid collision with the system's menu bar and task bar etc?

Eism: thanks! all right, fixed

Issue #8198, opened 2021-05-29
[MU4 Issue] [Docking] Shouldn't be able to click/drag empty space beside tabs in title bar / tab bar


cbjeukendrup:
**Describe the bug**
Clicking or dragging the empty space next to the tabs in the title bar of tabbed dock widgets should not be possible. 

**To Reproduce**
1. Have two dock widgets docked with tabs in the same location.
2. Click on the empty space on the right of the tabs.
**BUG 1:** An empty panel is shown.
3. Drag that empty space.
**BUG 2:** The whole frame containing all tabbed dock widgets is undocked. 
4. Drag one of the two dock widgets, so that it is moved to a new floating window. 
5. Try to drag the 'old' floating window (which contains now only one tab). 
**BUG 3:** assert failed in thirdparty/KDDockWidgets/src/private/TitleBar.cpp, line 206

**Expected behavior**
Clicking or dragging that space should do nothing, and there should be no assert failing.

**Screenshots**

https://user-images.githubusercontent.com/48658420/120068156-8d5d1780-c07f-11eb-903e-0951358cc1e0.mov

**Desktop (please complete the following information):**
 - OS: all (exact same situation)

cbjeukendrup:
**BUG 4:** The empty space next to the rightmost tab has no bottom border.
<img width="300" alt="Schermafbeelding 2021-05-30 om 22 08 43" src="https://user-images.githubusercontent.com/48658420/120118582-9a6b2b00-c193-11eb-8d22-b3bcd259b245.png">

PR #8211, opened 2021-05-31
[MU4] Fixes for tabs

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/8198

### DIFF
--- a/src/appshell/qml/dockwindow/DockFrame.qml
+++ b/src/appshell/qml/dockwindow/DockFrame.qml
@@ -34,7 +34,7 @@ Rectangle {
     //! NOTE: please, don't rename those properties because they are used in c++
     property QtObject frameCpp
     readonly property QtObject titleBarCpp: Boolean(frameCpp) ? frameCpp.actualTitleBar : null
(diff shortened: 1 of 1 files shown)


RomanPudashkin:
@cbjeukendrup please check this PR

cbjeukendrup:
It's much better now! 
However, I still found an "edge case" (quite literally 😉): when you drag the current tab at the height where the bottom border would be if it is not the selected tab, then still all dock widgets get undocked. Maybe the mouse area is just one pixel too tall?

https://user-images.githubusercontent.com/48658420/120192434-75ba9600-c21b-11eb-9ae4-d79388299fc1.mov

Also, when you hover a non-selected tab, the borders become lighter. I think they shouldn't. 
Schermafbeelding 2021-05-31 om 14 23 16

RomanPudashkin:
@cbjeukendrup thank you for testing! I have fixed the "edge case" but I'll fix separators later because this PR contains really important fixes and I want to merge it first. We  probably need to create a separate issue for the separators bug

cbjeukendrup:
@RomanPudashkin Thanks, it's fixed now! (I tested on macOS.)
I will create a new issue for the separators color bug.

Eism (approved):

PR #8219, opened 2021-06-01
[MU4] Added engraving module

igorkorsukov:
And moved `libmscore` to `engraving/libmscore` 
I think in the future it will be renamed to something like an `engine`, or a `model` ... or will separate into parts, but all this will be later.


vpereverzev (approved):

Jojo-Schmitz:
This move also needs adjustments to tools/fonttools/smufl2sym*

igorkorsukov:
> This move also needs adjustments to tools/fonttools/smufl2sym*

Thanks, I will do it

igorkorsukov:
> This move also needs adjustments to tools/fonttools/smufl2sym*

I looked, there is no dependency on libmscore path (it generate new output files, not changing current)
But I figure out other bug, and fixed it :)
https://github.com/musescore/MuseScore/pull/8245

Jojo-Schmitz:
Oops...

PR #8237, opened 2021-06-03
[MU3 Backend] ENG-6: Add partial default-y support

iveshenry18:
Resolves: [ENG-6](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-6): default-y attribute not respected

Due to inconsistencies in the MusicXML spec and its implementation
with regards to default-y and relative-y, precise implementation is
not currently reasonable. Instead, this commit adds two essential
functions which provide a sufficient amount of support for the vertical
layout of direction elements:

- In the case that the placement (above/below) isn't specified,
this is inferred by the value of default-y + relative-y (here called
totalY).
- Rather than being added to their Segment upon read, direction elements
with a default-y or relative-y specified are collected, sorted by
totalY, then added to their Segment in ascending order. Autoplace places
elements in the order they are added, so this allows at least the
ordering of default-y and relative-y to be respected


oktophonie (approved):

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -2127,6 +2140,13 @@ void MusicXMLParserPass2::measure(const QString& partId,
                   removeBeam(beam);
       }
 
+      // Sort and add delayed directions
+      std::sort(delayedDirections.begin(), delayedDirections.end(), delayedDirectionCompare);
```
vpereverzev, line 2144: Just use a lambda here, there is no need in static method :)

importexport/musicxml/importmxmlpass2.h
```diff
@@ -179,7 +179,9 @@ class Glissando;
 class Pedal;
 class Trill;
 class MxmlLogger;
+class MusicXMLDelayedDirectionElement;
 
+using DelayedDirectionsList = QVector<MusicXMLDelayedDirectionElement*>;
```
vpereverzev, line 184: It's better to use QList for your case instead of QVector

iveshenry18:
@vpereverzev thank you again for the comments! I've applied them and pushed the amended commit.

vpereverzev (approved):

PR #8239, opened 2021-06-03
ScoreFont and Sym cleanup

cbjeukendrup:
- Moved `ScoreFont` from sym.h to separate file scorefont.h
- Cleaned the code in `ScoreFont` somewhat
- Reduced unnecessary includes of sym.h
  (I remarked that when making changes to sym.h, Ninja had to recompile more than 200 files, which seemed unwanted to me. Now, it's only about 32.)

Maybe we should move (some of) the static methods of the `Sym` class to `ScoreFont`, and `SmuflAnchorId` to scorefont.h, so that sym.h will be needed only in scorefont.cpp (or if we want, we could even move the `Sym` class to scorefont.cpp).

(This was actually part of a bigger PR I was planning to make, but I was not yet sure about everything in that bigger PR, so I'll do this first. I hope it is not inconvenient that I make this PR now, considering the current changes to the fonts/symbols system.)


src/engraving/libmscore/mscore.cpp
```diff
@@ -314,7 +314,7 @@ void MScore::init()
     //
     _baseStyle.precomputeValues();
 
-    initScoreFonts();
+    ScoreFont::initScoreFonts();
```
igorkorsukov, line 317: cool!

src/engraving/libmscore/scorefont.cpp
```diff
@@ -0,0 +1,769 @@
...
+
+#include "draw/painter.h"
+#include "log.h"
+#include "mscore.h"
+#include "sym.h"
+
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QJsonParseError>
```
igorkorsukov, line 32: Better to move these headers above our headers

cbjeukendrup: I will do that. But what is our "official" policy with headers? Is it like the following?
1. Corresponding header for cpp file
2. Qt / STL / other "foreign" headers
3. our own headers

Or should 1 and 2 be the other way round?

And I assume that it is preferred to keep them sorted alphabetically when there are many includes.

src/engraving/libmscore/textbase.cpp
```diff
@@ -1855,7 +1856,7 @@ void TextBase::createLayout()
                     SymId id = Sym::name2id(sym);
                     if (id != SymId::noSym) {
                         CharFormat fmt = *cursor.format();              // save format
-                        // uint code = score()->scoreFont()->sym(id).code();
+                        // uint code = score()->scoreFont()->sym(id).codepoint();
```
igorkorsukov, line 1859: unnecessary change

src/engraving/libmscore/scorefont.cpp
Eism: I like this order
1. Corresponding header for cpp file
2. STL
3. Qt
4. 3rd party
5. libmscore
6. MU4 classes
   a. global
   c. other modules
   d. internal module

Please only not alphabetically

igorkorsukov:
```
Corresponding header for cpp file
Qt / STL / other "foreign" headers
our own headers
```
Right!
You can make an alphabetical order within the group :)

On the good, we need to swap 1 and 2, but historically it turned out as you wrote.
It is convenient to group STL, Qt, ours by modules ...
This order is also related to the fact that various definitions can be defined, which we do not even know about, but can affect the behavior of "foreign" headers. Better the way "foreign" definitions affect us, we will find out and fix it faster. For example https://github.com/musescore/MuseScore/blob/master/src/framework/ui/view/iconcodes.h#L42

cbjeukendrup: Thanks! I'll keep this in mind in the future :)
Maybe we should mention this in the Wiki (https://github.com/musescore/MuseScore/wiki/CodeGuidelines#header-and-implementation).

igorkorsukov (approved):

PR #8240, opened 2021-06-03
[MU4] Midi Mapping. Configuration of midi devices

Eism:


cbjeukendrup:
@Eism I implemented the `devicesChanged` notification for macOS using the CoreMIDI API. 
I think this way it is less power-consuming, because it makes use of the system notification instead of checking for devices every 500 ms.
You can find it here: https://github.com/musescore/MuseScore/commit/602260bd4afad3287ea95d0e617c3aac8613abeb

Eism:
@cbjeukendrup I like your solution. I did not find such a solution for Linux and Windows. For macOS, I wasn't hoping to find such a solution after two fail :(

Ok, let's merge your solution with a separate PR. Need to test and remove extra code about my solution

cbjeukendrup:
Good plan! I'll wait until this PR is merged, and then create a PR for the macOS solution.

RomanPudashkin (approved):

src/framework/midi/internal/midideviceslistener.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    m_actualDevicesCallback = callback;
+}
+
+mu::async::Notification MidiDevicesListener::devicesChanged() const
+{
+    return m_devicesChanged;
+}
+
+void MidiDevicesListener::updateDevices(MidiDevicesListener* self)
```
igorkorsukov, line 53: Would be nice add `th_` prefix

src/framework/midi/internal/midideviceslistener.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    self->doUpdateDevices();
+}
+
+void MidiDevicesListener::doUpdateDevices()
```
igorkorsukov, line 58: Would be nice add `th_` prefix

src/framework/midi/internal/midideviceslistener.cpp
```diff
@@ -0,0 +1,77 @@
...
+    while (true) {
+        MidiDeviceList devices = m_actualDevicesCallback();
+
+        setDevices(devices);
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
+    }
+}
+
+void MidiDevicesListener::setDevices(const MidiDeviceList& devices)
```
igorkorsukov, line 69: Would be nice add `th_` prefix

src/framework/midi/internal/midideviceslistener.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void MidiDevicesListener::updateDevices(MidiDevicesListener* self)
+{
+    self->doUpdateDevices();
+}
+
+void MidiDevicesListener::doUpdateDevices()
+{
+    while (true) {
```
igorkorsukov, line 60: We need a flag like `m_running` (atomic), which must be set in false in the destructor, before the join, otherwise the join will never happen.

src/framework/midi/internal/platform/lin/alsamidiinport.cpp
```diff
@@ -53,12 +72,12 @@ AlsaMidiInPort::~AlsaMidiInPort()
     }
 }
 
-std::vector<MidiDevice> AlsaMidiInPort::devices() const
+MidiDeviceList AlsaMidiInPort::devices() const
 {
```
igorkorsukov, line 76: We need to add a mutex, now this method is called from two threads. Or you need to make sure that this method is not called from two threads

igorkorsukov: Also for other platforms

cbjeukendrup:
I am working on a follow-up for this PR, fixing the review issues, and I'll also fix a crash because m_actualDevicesCallback is called before it is set (`libc++abi: terminating with uncaught exception of type std::__1::bad_function_call: std::exception`).
Or is somebody else already doing this?

src/framework/midi/internal/midideviceslistener.cpp
cbjeukendrup: Should `m_actualDevicesCallback` also be atomic? (Or I could just prohibit to set the callback while already running.)

Eism:
@cbjeukendrup hmm
okay, if you have already started, then continue

src/framework/midi/internal/midideviceslistener.cpp
igorkorsukov: m_actualDevicesCallback not posible to be atomic, so we should add mutex for it, or better yes, just prohibit to set the callback while already running.

PR #8247, opened 2021-06-04
[MU4] Replaced Qt geometry primitives to MU geometry primitives. Step1

igorkorsukov:


RomanPudashkin (approved):

src/engraving/draw/qpainterprovider.cpp
```diff
@@ -186,35 +186,41 @@ void QPainterProvider::drawPath(const QPainterPath& path)
     m_painter->drawPath(path);
 }
 
-void QPainterProvider::drawPolygon(const QPointF* points, int pointCount, PolygonMode mode)
+void QPainterProvider::drawPolygon(const PointF* points, int pointCount, PolygonMode mode)
 {
+    IF_ASSERT_FAILED(sizeof(QPointF) == sizeof(PointF)) {
```
cbjeukendrup, line 191: Isn't this a situation where you could also use `static_assert`?

igorkorsukov: Good point! Changed to use static_assert

src/engraving/draw/painter.cpp
```diff
@@ -315,55 +315,55 @@ void Painter::drawPath(const QPainterPath& path)
...
         }
         drawPolyline(pts, 2);
     }
 }
 
-void Painter::drawLines(const QPointF* pointPairs, int lineCount)
+void Painter::drawLines(const PointF* pointPairs, int lineCount)
 {
-    Q_ASSERT(sizeof(QLineF) == 2 * sizeof(QPointF));
+    Q_ASSERT(sizeof(LineF) == 2 * sizeof(PointF));
```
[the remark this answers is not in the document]
igorkorsukov: You are right! I will fix it in a next PR. Thanks!

PR #8257, opened 2021-06-06
Temporary solution for horizontal DockPanels

cbjeukendrup:
Needed to start working on the Piano Roll (and Mixer).

Later I will probably dive deeper into the dock widgets stuff and make a better solution. But first I would like to make some progress with the Piano Roll. 

TODO: 
- The design is not correct yet
- There is quite some code duplication for toolbars and panels
- The panel holders appear in the wrong place and don't work (I couldn't find out what is the problem here... 😞)
- Ultimately, we need nested toolbars inside the Piano Roll Panel. This might add more complexity the dock widgets system. 

Screenshot of current situation:

Schermafbeelding 2021-06-06 om 23 57 04

### DIFF
--- a/src/appshell/view/dockwindow/dockwindow.cpp
+++ b/src/appshell/view/dockwindow/dockwindow.cpp
@@ -468,15 +468,17 @@ DockToolBarHolder* DockWindow::resolveToolbarDockingHolder(const QPoint& localPo
     }
 
     QRect centralFrameGeometry = centralDock->frameGeometry();
-    centralFrameGeometry.setTopLeft(m_mainWindow->mapFromGlobal(centralDock->mapToGlobal({ centralDock->x(), centralDock->y() })));
+    centralFrameGeometry.moveTopLeft(m_mainWindow->mapFromGlobal(centralDock->mapToGlobal({ centralDock->x(), centralDock->y() })));
 
     QRect mainFrameGeometry = m_mainWindow->rect();
     DockToolBarHolder* newHolder = nullptr;
 
     if (localPos.y() < MAX_DISTANCE_TO_HOLDER) { // main toolbar holder
         newHolder = m_mainToolBarDockingHolder;
-    } else if (localPos.y() > centralFrameGeometry.top()
-               && localPos.y() < centralFrameGeometry.top() + MAX_DISTANCE_TO_HOLDER) { // page top toolbar holder
+    }
+    // TODO: Need to take any panels docked at top into account
+    else if (localPos.y() > centralFrameGeometry.top()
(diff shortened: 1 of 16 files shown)


src/appshell/view/dockwindow/dockwindow.cpp
```diff
@@ -403,15 +477,49 @@ DockToolBarHolder* DockWindow::resolveToolbarDockingHolder(const QPoint& localPo
...
+    centralFrameGeometry.setTopLeft(m_mainWindow->mapFromGlobal(centralDock->mapToGlobal({ centralDock->x(), centralDock->y() })));
+
+    QRect mainFrameGeometry = m_mainWindow->rect();
+    DockPanelHolder* newHolder = nullptr;
+
+    if (localPos.y() > centralFrameGeometry.top()
+        && localPos.y() < centralFrameGeometry.top() + MAX_DISTANCE_TO_HOLDER) { // page top panel holder
+        newHolder = page->panelHolderByLocation(DockBase::DockLocation::Top);
+    } else if (localPos.y() < centralFrameGeometry.bottom()) { // page left panel holder
+        if (localPos.x() < MAX_DISTANCE_TO_HOLDER) {
```
Eism, line 516: It won't work if there is a toolbar attached to the left.
I think it needs a binding to centralFrameGeometry instead of mainFrameGeometry like for top

cbjeukendrup:
Now it all works a lot better, also thanks to Elnur's comment. 
The only thing that does not work, is docking a non-main toolbar at the top while there are also panels at the top. It is complicated to check whether the holder needs to be shown, because it is between the main toolbars and the panels, so we can't use the whole window rect nor the central view rect. 

Current situation:
Schermafbeelding 2021-06-12 om 20 45 28

Eism:
The problems I found
1. I can dock the Piano Roll and the Mixer to left and right even though the allowedAreas flag is set: Qt.TopDockWidgetArea | Qt.BottomDockWidgetArea
2. I can not dock the Piano Roll and the Mixer to the bottom after undocking them

TODO. Maybe not related to this PR, but became noticeable after adding new panels
1. Need to remove drop indicators for panels like for toolbars
2. It is necessary to limit the area of the panels
    a. We should only dock the panel to the center of another panel
    b. We shouldn't show the panel holder if there is already a panel docking
    c. We should only dock the panel to the panel

Eism (approved):

cbjeukendrup:
> a. We should only dock the panel to the center of another panel

Martin said that he likes the ability to have multiple palettes docked into one area. (https://discord.com/channels/818804595450445834/818848561047273544/854812000348930048)
Personally, I also think it's nice to have.

PR #8258, opened 2021-06-06
Add the possibility to have async::Channels that take more than one parameter

cbjeukendrup:
Now we can do like

```
async::Channel<QString, framework::Orientation> m_dockOrientationChanged;

m_dockOrientationChanged.send(toolBarName, orientation);

m_dockOrientationChanged.onReceive(this, [this](QString name, mu::framework::Orientation orientation) {
    ...
})
```

I did this actually because I was fascinated by the possibilities with templates in C++, but I'm not sure if we should really use this. I once heard that such template magic is quite difficult for compilers (although on my Mac with Clang, I didn't notice much difference), and in those few cases where we actually do need to send multiple parameters, using `std::pair` (or `std::tuple`) is not problematic at all. So, it's a costs and benefits consideration.

Also, I'm in fact modifying third party code in this PR, which is... questionable.

### DIFF
--- a/src/appshell/view/dockwindow/dockwindow.cpp
+++ b/src/appshell/view/dockwindow/dockwindow.cpp
@@ -59,12 +59,12 @@ void DockWindow::componentComplete()
         resetWindowState();
     });
 
-    mainWindow()->changeToolBarOrientationRequested().onReceive(this, [this](std::pair<QString, mu::framework::Orientation> orientation) {
+    mainWindow()->changeToolBarOrientationRequested().onReceive(this, [this](QString name, mu::framework::Orientation orientation) {
         const DockPage* page = currentPage();
-        DockToolBar* toolBar = page ? dynamic_cast<DockToolBar*>(page->dockByName(orientation.first)) : nullptr;
+        DockToolBar* toolBar = page ? dynamic_cast<DockToolBar*>(page->dockByName(name)) : nullptr;
 
         if (toolBar) {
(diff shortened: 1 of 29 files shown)


igorkorsukov:
The changes are very cool. Let me think about it!

> Also, I'm in fact modifying third party code in this PR, which is... questionable.
In this case, don't worry about it.
This code is mine, I created it a long time ago 
We will think about how best to proceed.

RomanPudashkin (approved):

igorkorsukov:
@cbjeukendrup These channels were inspired by the channels from Golang.
Golang's philosophy is to keep it simple so that developers of all levels can efficiently design large programs together.
In Golang, the channels have only one argument, I think this was done on purpose, so the complication looks disquieting.
On the other hand, for the consumer, the change looks simple and convenient, temptingly convenient :) And the fact that the implementation is difficult inside is not so important if it is hidden and the consumer does not know anything about it.
In general, I think you did a cool thing and it needs to be merged.

cbjeukendrup:
@igorkorsukov Thanks, that's clear! I will solve the conflicts now, so that the PR can be merged.

igorkorsukov (approved):

vpereverzev:
@cbjeukendrup @igorkorsukov 
nice changes! Well done!

Unfortunately you've forgotten to apply these changes to promise.h (probably because no one used it before), but it's quite important piece of async, at least for my rework of the audio-engine

cbjeukendrup:
@vpereverzev Oops, apologies! Fix is already on the way.

PR #8267, opened 2021-06-07
[MU4] Score ordering

njvdberg:
This PR implements:
 -1 Score ordering as it was introduced in 3.6.
 -2 Automatic bracketing and bar line span as was introduced in 3.6.
 -3 ~~[Suggested changes](https://github.com/musescore/MuseScore/pull/8099#issuecomment-841179574) regarding the `Make/Undo Soloist` button.~~ ( moved to separate PR #8304)
 -4 Solves a minor problem in bar line spanning for `Soloists`. When there are multiple `soloists` defined **and** the `soloists` group is **not** in a `section`, the bar lines of all soloist instruments were connect, this shouldn't because bar line spanning is a `section` property. Outside a section bar lines are never connected.

Please note this PR is built on top of PR #8147, PR #8147 has to be merged first.


RobFog:
Shouldn't "this PR has to be merged first." rather "that PR has to be merged first."? 😉

njvdberg:
@RobFog You might be right. But to be sure, I just mentioned the actual PR again :wink:

njvdberg:
Due to rebase conflicts the [suggested changes](https://github.com/musescore/MuseScore/pull/8099#issuecomment-841179574) regarding the `Make/Undo Soloist` button are moved to separate PR #8304.

src/engraving/libmscore/score.h
```diff
@@ -1028,8 +1028,9 @@ class Score : public QObject, public ScoreElement
     void setEnableVerticalSpread(bool val);
     qreal maxSystemDistance() const;
 
-    ScoreOrder* scoreOrder() const { return _scoreOrder; }
-    void setScoreOrder(ScoreOrder* order) { _scoreOrder = order; }
+    ScoreOrder scoreOrder() const { return _scoreOrder; }
+    void setScoreOrder(ScoreOrder order) { _scoreOrder = order; }
```
RomanPudashkin, line 1032: Please, move the implementation of those methods to cpp

src/engraving/libmscore/scoreorder.cpp
```diff
@@ -0,0 +1,437 @@
...
+{
+    if (!reader.hasAttribute(name)) {
+        return defvalue;
+    }
+    QString attr { reader.attribute(name) };
+    if (attr.toLower() == "false") {
+        return false;
+    } else if (attr.toLower() == "true") {
+        return true;
+    } else {
```
RomanPudashkin, line 52: You can remove this else:

} else if (attr.toLower() == "true") {
...
}

qDebug("invalid value \"%s\" for attribute \"%s\", using default \"%d\"", qPrintable(attr), qPrintable(name), defvalue);
return defvalue;
}

src/engraving/libmscore/scoreorder.cpp
```diff
@@ -0,0 +1,437 @@
...
+        return QString("<unsorted>");
+    }
+
+    if (soloist) {
+        return QString("<soloists>");
+    } else if (instrumentMap.contains(instrTemplate->id)) {
+        return instrumentMap[instrTemplate->id].id;
+    } else if (instrTemplate->family) {
+        return instrTemplate->family->id;
+    } else {
```
RomanPudashkin, line 181: You can remove this else

src/instruments/instrumentstypes.h
```diff
@@ -171,6 +180,37 @@ struct PartInstrument {
 
 using PartInstrumentList = QList<PartInstrument>;
 
+struct ScoreOrderGroup {
+    int index;
+    QString family;
+    QString section { QString() };
+    QString unsorted { QString() };
+    bool bracket { false };
```
RomanPudashkin, line 188: Better to use = here

bracket = false;
showSystemMarkings = false;
...

src/engraving/libmscore/score.h
njvdberg: Done

src/engraving/libmscore/scoreorder.cpp
njvdberg: Done

src/engraving/libmscore/scoreorder.cpp
njvdberg: Done

src/instruments/instrumentstypes.h
njvdberg: Done

src/instruments/instrumentstypes.h
```diff
@@ -171,6 +180,37 @@ struct PartInstrument {
...
+    QString family;
+    QString section { QString() };
+    QString unsorted { QString() };
+    bool bracket = false;
+    bool showSystemMarkings = false;
+    bool barLineSpan = false;
+    bool thinBracket = false;
+};
+
+struct InstrumentOverwrite {
```
RomanPudashkin, line 194: Seems that you can just make an alias instead of copying the whole structure

using InstrumentOverwrite = Ms::InstrumentOverwrite;

src/notation/internal/notation.h
```diff
@@ -50,6 +50,8 @@ class Notation : virtual public INotation, public IGetScore, public async::Async
     Meta metaInfo() const override;
     void setMetaInfo(const Meta& meta) override;
 
+    instruments::ScoreOrder scoreOrder() const override;
```
RomanPudashkin, line 53: Seems that a better place for this method is INotationPartsPtr. What do you think?

src/instruments/view/instrumentlistmodel.h
```diff
@@ -78,9 +83,36 @@ class InstrumentListModel : public QObject, public async::Asyncable
...
+
+    void scoreOrdersChanged();
+
 private:
+    struct SelectedInstrumentInfo
+    {
+        QString id;
+        QString name;
+        QString familyId;
+        bool isSoloist;
```
RomanPudashkin, line 96: Please, add default values for all bool here (and for ScoreOrderInfo also)

src/instruments/instrumentstypes.h
```diff
@@ -171,6 +180,37 @@ struct PartInstrument {
 
 using PartInstrumentList = QList<PartInstrument>;
 
+struct ScoreOrderGroup {
+    int index;
```
RomanPudashkin, line 184: Please, add a default value

src/instruments/instrumentstypes.h
```diff
@@ -171,6 +180,37 @@ struct PartInstrument {
...
+    QString family;
+    QString section { QString() };
```
RomanPudashkin, line 186: section will be empty by default, you don't need to call QString() explicitly

src/instruments/view/instrumentlistmodel.cpp
```diff
@@ -336,18 +375,171 @@ void InstrumentListModel::setSearchText(const QString& text)
...
+    }
 
     return result;
 }
 
-void InstrumentListModel::selectOrderType(const QString&)
+QVariant InstrumentListModel::selectedScoreOrderIndex() const
 {
-    NOT_IMPLEMENTED;
+    QString id = m_selectedScoreOrderIndex <= 0 ? QString("---") : m_scoreOrders[m_selectedScoreOrderIndex].id;
```
RomanPudashkin, line 402: Probably you can just return an invalid QVariant() in a case when m_selectedScoreOrderIndex is invalid

src/instruments/view/instrumentlistmodel.cpp
```diff
@@ -336,18 +375,171 @@ void InstrumentListModel::setSearchText(const QString& text)
...
+    return QVariant(m_selectedScoreOrderIndex);
+}
+
+void InstrumentListModel::setSelectedScoreOrderIndex(const QVariant& index)
+{
+    m_selectedScoreOrderIndex = index.toInt();
```
RomanPudashkin, line 408: Did you miss emit scoreOrdersChanged here? Because this function is a WRITE method
Q_PROPERTY(QVariant selectedScoreOrderIndex READ selectedScoreOrderIndex WRITE setSelectedScoreOrderIndex NOTIFY scoreOrdersChanged)

src/instruments/instrumentstypes.h
```diff
@@ -171,6 +180,37 @@ struct PartInstrument {
...
+    int index;
+    QString id;
+    QString name;
+    QMap<QString, InstrumentOverwrite> instrumentMap;
+    QList<ScoreOrderGroup> groups;
+
+    bool isValid() { return !groups.empty(); }
+};
+
+struct PartInstrumentListScoreOrder {
```
RomanPudashkin, line 209: I would prefer to rename this to OrderedPartInstrumentList or to PartInstrumentListWithOrder. What do you think?

src/instruments/view/instrumentlistmodel.cpp
```diff
@@ -38,14 +38,16 @@ static const QString GROUP_ID("groupId");
 static const QString CONFIG_KEY("config");
 static const QString SOLOIST_KEY("isSoloist");
 static const QString PART_KEY("isExistingPart");
+static const QString SOLOISTS_SCOREORDERGROUP("<soloists>");
+static const QString UNSORTED_SCOREORDERGROUP("<unsorted>");
```
RomanPudashkin, line 42: Probably those strings are unused

RomanPudashkin:
I left several new issues in your PR but they are not critical. But we need to do some testing before merging

@vshalkevich please check this PR

src/notation/internal/notation.h
njvdberg: The reason to put `scoreOrder()` in `notation.h` instead of `inotationparts.h` was `inotationparts.h` seems have have "setter" methods only, no "retrieval" methods. I didn't want to unnecessary mess up the interface.

vshalkevich:
@njvdberg Please, make rebase with current master build

src/instruments/instrumentstypes.h
njvdberg: The structure contains a list of `PartInstrument`s **and** a `ScoreOrder`. `OrderedPartInstrumentList` suggests the `PartInstruments` are ordered. That might be true but it ignores the fact the structure also contains the score order itself.
`PartInstrumentListAndOrder` also suggests an order which is not necessarily true. A better name should be then `PartInstrumentListAndScoreOrder` but personally I think it too long.
The best name might be `PassedFormData`. It make it clear it is data from the form but doesn't try to explain **what** data is passed.

src/instruments/instrumentstypes.h
njvdberg: Good catch, I wasn't even aware this was possible.
Done.

src/instruments/view/instrumentlistmodel.h
njvdberg: Done.

src/instruments/instrumentstypes.h
njvdberg: Done.

src/instruments/instrumentstypes.h
njvdberg: Done.

src/instruments/view/instrumentlistmodel.cpp
njvdberg: As a matter of fact, the whole was unnecessary because `id` was never used :smiley: Apparently the compiler didn't catch that, bad compiler...

src/instruments/view/instrumentlistmodel.cpp
njvdberg: Indeed, that one was missed.

src/instruments/view/instrumentlistmodel.cpp
njvdberg: Indeed a relic of previous version. Removed.

njvdberg:
@RomanPudashkin 

> I left several new issues in your PR but they are not critical. We need to do some testing before merging

Nevertheless I'm curious to know, if it is not too much for of course, Always willing to learn :wink:

njvdberg:
@vshalkevich 

> @njvdberg Please, make rebase with current master build

Done

vshalkevich:
@njvdberg 
2021-06-20 21_00_48-MuseScore 4

Changing Instruments order via Instruments panel doesnt take effect in Parts dialog 
Instruments In Parts dialog keep their positions (but they shoud be changed according to Instruments panel)

vshalkevich:
2021-06-20 21_27_14-MuseScore 4
Deleting instruments from Parts dialog doesnt make effect on Instruments panel.

njvdberg:
@vshalkevich Both reported issues are as "designed" and part of the coming PR implementing [Instruments_2020_V07_Second_Amendment](https://www.dropbox.com/s/rhs2ietylrbzrjt/Instruments_2020_V07_Second_Amendment_V01.pdf?dl=0)

vshalkevich:
Okay. That`s fine.

RomanPudashkin (approved):

Issue #7882, opened 2021-04-03
[MU4 Issue] Can't add text (staff text, lyrics, tempos, frames, etc.)


vshalkevich:
**Describe the bug**
A short and clear description of what the bug is.

**To Reproduce**
1. Try to create a score with text, lyrics, tempos)

**Expected behavior**
It is possible to create text, lyrics, etc.

**Screenshots**

**Desktop (please complete the following information):**
Linux Ubunty

**Additional context**

wizofaus:
I suspect you're seeing what I did which is that palette commands that add elements as children of/anchored to other elements don't trigger a proper refresh so you don't see the newly added elements straight away. Should be fixed by https://github.com/musescore/MuseScore/pull/8274

wizofaus:
@vshalkevich can you still see a problem here?

PR #8274, opened 2021-06-08
[MU4] Fix #8263 ensure score refreshed after adding attached elements from palette

### LINKED ISSUE #8263: [MU4 Issue] Accidentals added to trill line don't show immediately
**Describe the bug**
After adding a trill line to a note, then selecting the trill line and adding an accidental, it doesn't appear immediately, only after some other change that triggers a refresh.
NB same also happens adding accidentals to notes.

**To Reproduce**
1. Add a trill line to a note
2. Select the trill line
3. Add an accidental from the palette
4. No accidental appears
5. Change the note in some other way (or save/close/re-open etc.)
6. Accidental appears as expected

Window 10 MSVC 2019 build.

**Additional context**
Has been reported by users since 3.2.2, and on Mac etc.

Doesn't *always* happen in 3.2.6, but I can't figure out when it does and doesn't.

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8263

Accidentals weren't appearing after adding them from the palette to notes or trill lines

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -26,6 +26,7 @@
 #include <QRectF>
 #include <QPainter>
 #include <QClipboard>
+#include <QApplication>
 
 #include "ptrutils.h"
 
@@ -95,12 +96,16 @@ Ms::Score* NotationInteraction::score() const
 
 void NotationInteraction::startEdit()
 {
+    m_notifyAboutDropChanged = false;
     m_undoStack->prepareChanges();
 }
 
 void NotationInteraction::apply()
 {
     m_undoStack->commitChanges();
+    if (m_notifyAboutDropChanged) {
+        notifyAboutDropChanged();
+    }
 }
 
 void NotationInteraction::notifyAboutDragChanged()
@@ -1392,7 +1397,7 @@ void NotationInteraction::applyDropPaletteElement(Ms::Score* score, Ms::Element*
         }
         dropData.dropElement = 0;
 
-        notifyAboutDropChanged();
+        m_notifyAboutDropChanged = true;
     }
 }
 
@@ -1599,7 +1604,7 @@ bool NotationInteraction::dragMeasureAnchorElement(const PointF& pos)
         m_dropData.ed.dropElement->score()->addRefresh(m_dropData.ed.dropElement->canvasBoundingRect());
         m_dropData.ed.dropElement->setTrack(track);
         m_dropData.ed.dropElement->score()->addRefresh(m_dropData.ed.dropElement->canvasBoundingRect());
-        notifyAboutDropChanged();
+        m_notifyAboutDropChanged = true;
         return true;
     }
     m_dropData.ed.dropElement->score()->addRefresh(m_dropData.ed.dropElement->canvasBoundingRect());
--- a/src/notation/internal/notationinteraction.h
+++ b/src/notation/internal/notationinteraction.h
@@ -283,6 +283,8 @@ private:
     async::Channel<ScoreConfigType> m_scoreConfigChanged;
 
     Ms::Lasso* m_lasso = nullptr;
+
+    bool m_notifyAboutDropChanged = false;
 };
 }
 


wizofaus:
This took me far too long to figure out (at least 2 hours, including build issues)   The  whole way accidentals and trill lines are handled really threw me - specifically the Trill::treeChild( ) and Accidental::treeParent( ) overrides that are in scoretree.cpp (and not trill.cpp!).  There seems to be a lot of "special case" logic here that I'm not sure I understand the reason for.  Also I found that the button in the Inspector to bring up the "accidentals" properties dialog doesn't work (for adding brackets etc.) - this is true in general, not just trill line accidentals. It seems somewhat similar to the Change Time Signature issue.

MarcSabatella:
I don’t know the changes made for 4.0, but to me there shouldn’t be a need to modify a basic function like this that is called on every drop - should be more a matter of making sure the operation that actually adds the accidental triggers a layout.  But maybe things have changed.  Just be sure this doesn’t adversely affect performance of other drops - that they don’t trigger full score relayout  unnecessarily.

wizofaus:
> I don’t know the changes made for 4.0, but to me there shouldn’t be a need to modify a basic function like this that is called on every drop - should be more a matter of making sure the operation that actually adds the accidental triggers a layout. But maybe things have changed. Just be sure this doesn’t adversely affect performance of other drops - that they don’t trigger full score relayout unnecessarily.

This function specifically only supports drops when new elements are added to the score, but the problem was nothing was triggering the bsp tree to be re-built in that case.  All drop actions that add new elements to the score need to do this, and I don't know all the other cases this gets triggered (though it does get triggered adding an accidental to a note via the palette, and that didn't have the same refresh issue, so I can look at how it works in that case).  I certainly can't notice any performance issues but I haven't tried with a really really big score.

wizofaus:
> > I don’t know the changes made for 4.0, but to me there shouldn’t be a need to modify a basic function like this that is called on every drop - should be more a matter of making sure the operation that actually adds the accidental triggers a layout. But maybe things have changed. Just be sure this doesn’t adversely affect performance of other drops - that they don’t trigger full score relayout unnecessarily.
> 
> This function specifically only supports drops when new elements are added to the score, but the problem was nothing was triggering the bsp tree to be re-built in that case. All drop actions that add new elements to the score need to do this, and I don't know all the other cases this gets triggered (though it does get triggered adding an accidental to a note via the palette, and that DOES have the same problem). I certainly can't notice any performance issues but I haven't tried with a really really big score.
> 
> The only thing I'd still like to check is why it sometimes _does_ work in 3.6.2 without this change. There is no trigger of a bsp tree rebuild when adding items from the palette, and yet it still repaints correctly which doesn't seem possible.

Ok so in 3.6.2 the tree gets rebuilt because of:
```
        page->rebuildBspTree()

 	MuseScore3.exe!Ms::LayoutContext::collectPage() Line 4875	C++
 	MuseScore3.exe!Ms::LayoutContext::layout() Line 5049	C++
 	MuseScore3.exe!Ms::Score::doLayoutRange(const Ms::Fraction & st, const Ms::Fraction & et) Line 5035	C++
 	MuseScore3.exe!Ms::Score::update(bool resetCmdState) Line 302	C++
 	MuseScore3.exe!Ms::Score::endCmd(bool rollback) Line 262	C++
>	MuseScore3.exe!Ms::Palette::applyPaletteElement(Ms::Element * element, QFlags<enum Qt::KeyboardModifier> modifiers) Line 797	C++
```

now to figure out why it doesn't do that in 4.0

wizofaus:
> > > I don’t know the changes made for 4.0, but to me there shouldn’t be a need to modify a basic function like this that is called on every drop - should be more a matter of making sure the operation that actually adds the accidental triggers a layout. But maybe things have changed. Just be sure this doesn’t adversely affect performance of other drops - that they don’t trigger full score relayout unnecessarily.
> > 
> > 
> > This function specifically only supports drops when new elements are added to the score, but the problem was nothing was triggering the bsp tree to be re-built in that case. All drop actions that add new elements to the score need to do this, and I don't know all the other cases this gets triggered (though it does get triggered adding an accidental to a note via the palette, and that DOES have the same problem). I certainly can't notice any performance issues but I haven't tried with a really really big score.
> > The only thing I'd still like to check is why it sometimes _does_ work in 3.6.2 without this change. There is no trigger of a bsp tree rebuild when adding items from the palette, and yet it still repaints correctly which doesn't seem possible.
> 
> Ok so in 3.6.2 the tree gets rebuilt because of:
> 
> ```
>         page->rebuildBspTree()
> 
>  	MuseScore3.exe!Ms::LayoutContext::collectPage() Line 4875	C++
>  	MuseScore3.exe!Ms::LayoutContext::layout() Line 5049	C++
>  	MuseScore3.exe!Ms::Score::doLayoutRange(const Ms::Fraction & st, const Ms::Fraction & et) Line 5035	C++
>  	MuseScore3.exe!Ms::Score::update(bool resetCmdState) Line 302	C++
>  	MuseScore3.exe!Ms::Score::endCmd(bool rollback) Line 262	C++
> >	MuseScore3.exe!Ms::Palette::applyPaletteElement(Ms::Element * element, QFlags<enum Qt::KeyboardModifier> modifiers) Line 797	C++
> ```
> 
> now to figure out why it doesn't do that in 4.0

The plot thickens - in 4.0 Score.update( ) is called when notifyAboutDropChanged() is called, but by that time it seems the CmdState has lots its layoutRange so nothing actually happens.

wizofaus:
> > > > I don’t know the changes made for 4.0, but to me there shouldn’t be a need to modify a basic function like this that is called on every drop - should be more a matter of making sure the operation that actually adds the accidental triggers a layout. But maybe things have changed. Just be sure this doesn’t adversely affect performance of other drops - that they don’t trigger full score relayout unnecessarily.
> > > 
> > > 
> > > This function specifically only supports drops when new elements are added to the score, but the problem was nothing was triggering the bsp tree to be re-built in that case. All drop actions that add new elements to the score need to do this, and I don't know all the other cases this gets triggered (though it does get triggered adding an accidental to a note via the palette, and that DOES have the same problem). I certainly can't notice any performance issues but I haven't tried with a really really big score.
> > > The only thing I'd still like to check is why it sometimes _does_ work in 3.6.2 without this change. There is no trigger of a bsp tree rebuild when adding items from the palette, and yet it still repaints correctly which doesn't seem possible.
> > 
> > 
> > Ok so in 3.6.2 the tree gets rebuilt because of:
> > ```
> >         page->rebuildBspTree()
> > 
> >  	MuseScore3.exe!Ms::LayoutContext::collectPage() Line 4875	C++
> >  	MuseScore3.exe!Ms::LayoutContext::layout() Line 5049	C++
> >  	MuseScore3.exe!Ms::Score::doLayoutRange(const Ms::Fraction & st, const Ms::Fraction & et) Line 5035	C++
> >  	MuseScore3.exe!Ms::Score::update(bool resetCmdState) Line 302	C++
> >  	MuseScore3.exe!Ms::Score::endCmd(bool rollback) Line 262	C++
> > >	MuseScore3.exe!Ms::Palette::applyPaletteElement(Ms::Element * element, QFlags<enum Qt::KeyboardModifier> modifiers) Line 797	C++
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > now to figure out why it doesn't do that in 4.0
> 
> In 4.0 notifyAboutDropChanged() is calling NotationParts::updatePartTitles( ) and in that case the CmdState is reset, and I believe it's because of this that the layout is not refreshing properly.

Ok I have an alternative fix that instead of calling `notifyAboutDropChanged()` directly just sets a flag to call it after `apply()` is called, which seems to work fine in the cases I've tested too.

src/notation/internal/notationinteraction.cpp
```diff
@@ -1389,7 +1394,7 @@ void NotationInteraction::applyDropPaletteElement(Ms::Score* score, Ms::Element*
         }
         dropData.dropElement = 0;
 
-        notifyAboutDropChanged();
```
wizofaus, line 1392: This was the main culprit - calling this before apply() means the cmdState gets reset and hence endCmd() doesn't update the score as required.
I'm still not convinced this is quite right though as now there are two serialized calls to startCmd()/endCmd() for each add-from-palette operation (previously they were nested), which might cause an issue with undo - but the undo/redo stack is already unstable in MU4 so hard to know if it's caused an issue.

wizofaus:
Is there a hold up getting this merged?  Without this fix adding almost anything from the palettes appears to do nothing (because the score isn't updating properly).

wizofaus:
Should also fix https://github.com/musescore/MuseScore/issues/7882

wizofaus:
BTW this also fixes the issue where the score becomes corrupted and MuseScore crashes after changing the time signature.

src/notation/internal/notationinteraction.cpp
```diff
@@ -26,6 +26,7 @@
 #include <QRectF>
 #include <QPainter>
 #include <QClipboard>
+#include <QApplication>
```
wizofaus, line 29: Without this you can't compile just this file on its own and you get errors highlighted in Visual Studio.  Not quite sure why it compiles as part of the full build!

PR #8275, opened 2021-06-08
[MU3 Backend] ENG-8: Pedal Change Element

iveshenry18:
Resolves: [ENG-8](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-8): Pedal elements are ignored

This PR implements the pedal type="change" on MusicXML import and export, including a slightly (though minimally) hacky fix to accommodate the difference between the MusicXML representation of the pedal change (a single event), and the MuseScore representation (a HookType::HOOK_45 to end a line and a new line with a HookType::HOOK_45).

Additionally, this PR changes the behavior when a pedal type="start" that comes before the previous pedal has been stopped (a case which shouldn't occur, but nevertheless seems to in many scores). The previous behavior was to essentially ignore this new "start" and just continue the existing line; this change elects to delete the line that was started but never stopped, which is observed to be the more common intention.

Finally, this PR changes the default behavior of the sign attribute of pedals; where the MusicXML specification dictates an omitted sign attribute should be "yes" if line is "no" and vice versa, Dolet 6.6 for Sibelius erroneously omits this attribute when the sign is present (even when the line "yes"), so I have added an override set the default to "yes". Since this is technically against the spec, this portion (commit 148c9a6) may be omitted (or amended) when eventually porting to master.


RobFog:
cc @lvinken

lvinken:
Thanks for reminding me. I am unable to do an extensive review right now, but a quick glance at the conversation and the code changes suggests this is OK. Wrt the incorrect Dolet export, I'm afraid that due to lack of clear specification and/or formal test suite many applications will produce (somewhat) incorrect MusicXML in the foreseeable future. We'll have to try to make the best out of that, the only discussion IMHO is whether to do that automatically or involve the end user in some way (such as setting import options).

oktophonie (approved):

RomanPudashkin (approved):

importexport/musicxml/exportxml.cpp
```diff
@@ -4330,13 +4330,39 @@ void ExportMusicXml::ottava(Ottava const* const ot, int staff, const Fraction& t
...
+      if (pd->tick() == tick) {
+            switch (pd->beginHookType()) {
+                  case HookType::HOOK_45:
+                        pedalType = "change";
+                        break;
+                  case HookType::NONE:
+                        if (pd->beginText() == "") {
+                              pedalType = "resume";
+                              break;
+                              }
```
Jojo-Schmitz, line 4349: Is this supposed to fall though in the else case?

iveshenry18: Yes; a pedal with a NONE HookType but with beginText != "" is of pedalType "start"

Jojo-Schmitz: OK, then a `// FALLTHROUGH` is in order, to prevent a compiler warning and make the intention clear to human readers too

iveshenry18: #8481

PR #8280, opened 2021-06-08
High Contrast Overhaul

ArjunBoi:
The aim is to introduce two new High Contrast themes (Black and White) ~under a new "Accessibility" page under the "Appearance" tab in "Preferences" menu.~

That design has since been scrapped. The new design requires me to have the Accessibility settings under the "Appearance" tab itself, accessible through an "Enable High-Contrast" CheckBox. This design makes more sense since it also shares a lot of settings between itself and the "Appearance" page.

Find the changes at: Edit -> Preferences -> Appearance ~[-> Accessibility]~

Putting up in Draft form because this PR will require multiple commits. This will also allow mentors and maintainers to follow along with my progress and advise/review accordingly.


ArjunBoi:
I see that my build failed for MacOS.
I have discovered and solved the issue and will push the commit soon with other changes.

ArjunBoi:
### In [this commit:](https://github.com/musescore/MuseScore/pull/8280/commits/f9ab792f4cd6dc4cc2e1e2361a96942cfc1f9da3)

### **Things Added:**

- MacOS Build should now work.
- Appearance Tab has a new child page named "Accessibility".
- Appearance Tab now only shows General Themes.
- High Contrast Themes now moved to the new Accessibility Tab.
- Accessibility Tab also displays colours of various current properties (Accent Colour/Font Colour etc.), and allows the user to select a colour from the entire colour spectrum and use that colour for that property instead. (unfinished)
- Modified ColorPicker QML element to suit current usecase's needs (No fundamental changes; just added a bool property that allows the developer to send and use a new colour instead (currently only implemented for border colour). (See ColorPicker.qml) This change won't affect other places where colorPicker is being used.


### **Unfinished/Missing Features:**

- Colour of Disabled Text property has been set to "#000000" because there is no mention of this in the figma design. I couldn't find any instance of this being used in any of the other general themes either. @Tantacrul, @jessjwilliamson 
- Accessibility tab allows user to pick a new colour from the colour spectrum for the concerned property, but the colour won't be saved/updated, working on the back-end logic for this.
- Inappropriate spacing between the properties in Accessibility Page (Supposed to be 12 pixels). Current spacing is whatever the Repeater{} element sees fit. I tried implementing ListView but that kept breaking things. Will come up with a proper solution in the next commit.
- Changing themes in a specific tab i.e either _only_ Accessibility or _only_ Appearance selects one theme at a time, like it should. However, when the user is switching and checking different themes out between both the tabs, a theme from each of the tab remains highlighted (i.e. there is no communication between Accessibility and Appearance). @Eism has helped me out a lot with how the solution should work, and i am currently working on this. 
**Desired Behaviour:** Selecting a theme should de-select all other themes.

### **To Do ( ideally when everything mentioned above is working properly):**

- Score Page colour Inversion
- Implement "Reset-to-Default" button that sets HC themes to their default values.
- Inset Stroke for all buttons in the UI for better visibility.

### **Other Notes:**

Had to use a "Magic Number" (AccessibilityPreferencesPage.qml line 90) because the orientation just wasn't lining up without it.

src/appshell/view/preferences/accessibilitypreferencesmodel.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+#ifndef MU_APPSHELL_ACCESSIBILITYPREFERENCESMODEL_H
+#define MU_APPSHELL_ACCESSIBILITYPREFERENCESMODEL_H
+
+#include <QObject>
+//#include "appearancepreferencesmodel.h"
+
+#include "modularity/ioc.h"
+#include "ui/iuiconfiguration.h"
+//#include "async/asyncable.h"
```
cbjeukendrup, line 30: Because there are now two theme pages, the theme can change while the change is not initiated by the model. That is perfectly fine, but it does mean that you need to subscribe to theme changes and emit the `themesChanged` signal also when the theme is changed "externally". 

This can be done by using `uiConfiguration()->currentThemeChanged().onNotify(...)`. (To see examples, just search for `onNotify`.)
You need to call this in a new `load()` method (example: `CanvasPreferencesModel`). 

In order to be able to subscribe for `onNotify`, the model class need to be `async::Asyncable`, and therefore you need to include this header.
(All this goes for both the AppearancePreferencesModel and the AccessibilityPreferencesModel.)

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
```diff
@@ -0,0 +1,99 @@
...
+                anchors.right: parent.right
+
+                StyledTextLabel {
+                    text: modelData["textRole"]
+                    width: content.firstColumnWidth
+                    horizontalAlignment: Text.AlignLeft
+                }
+
+                ColorPicker {
+                    y: -8
```
cbjeukendrup, line 90: I don't really like changing the x and y properties of an item... and the `-8` is a "magic number".
I guess the problem you were solving is that the label and the color picker were not aligned. You can fix that by doing
```
    StyledTextLabel {
        ...
    }
```
(Bonus: a <code>Column*Layout*</code> centers its items automatically. But when there are no other big advantages of using a `Layout`, using a plain `Column` is better.)

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
```diff
@@ -35,6 +35,7 @@ PreferencesPage {
         id: appearanceModel
     }
 
+
```
cbjeukendrup, line 38: Unnecessary change

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
```diff
@@ -0,0 +1,99 @@
...
+
+            onThemeChangeRequested: {
+                accessibilityModel.currentThemeIndex = newThemeIndex
+            }
+        }
+
+        SeparatorLine {}
+
+        StyledTextLabel {
+            text: qsTrc("appshell", "UI Colours")
```
cbjeukendrup, line 63: American vs British: in code, we tend to use American, so in this case `Colors`. For the British users, it gets "translated" into British by `qsTrc`.

cbjeukendrup (commented):
This looks very good overall! 
Sorry for this massive review 😅 Most comments are about just tiny things.

cbjeukendrup:
About the Linux checks: no idea why its failing for this PR specifically. Maybe "bad luck" with the unity build. The problem seems that in dockage.cpp, before line 26, a `#include "docktoolbarholder.h"` is missing.

src/appshell/appshell.qrc
```diff
@@ -75,5 +75,6 @@
         <file>qml/platform/linux/Main.qml</file>
         <file>qml/platform/mac/Main.qml</file>
         <file>qml/platform/win/Main.qml</file>
+        <file>qml/Preferences/AccessibilityPreferencesPage.qml</file>
```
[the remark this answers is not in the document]
ArjunBoi: Done!
I knew this was very out of place, and I couldn't figure out how to re-order the files (the only way I found was to sort them alphabetically, and that would've messed up the ordering even more) when I had it open in Qt Creator.
I ended up modifying it using VS Code instead, as that showed me the raw code.

src/appshell/qml/Preferences/internal/ThemesSection.qml
```diff
@@ -26,15 +26,15 @@ import MuseScore.UiComponents 1.0
...
     property alias currentThemeIndex: view.currentIndex
     property alias themes: view.model
-
     signal themeChangeRequested(var newThemeIndex)
 
     spacing: 18
 
     StyledTextLabel {
-        text: qsTrc("appshell", "Themes")
+        id: styledtextlabel
```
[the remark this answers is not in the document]
ArjunBoi: Done!

src/appshell/qml/Preferences/internal/ThemesSection.qml
```diff
@@ -70,9 +70,11 @@ Column {
             }
 
             RoundedRadioButton {
+                id: roundedradiobutton
                 width: parent.width
 
                 checked: view.currentIndex === model.index
+
```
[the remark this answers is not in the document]
ArjunBoi: These are indeed very unnecessary.
These changes were made to go hand in hand with others that I had made while trying out some things, and I forgot to clean up after my little experiments 😅 
Fixed!

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
```diff
@@ -0,0 +1,87 @@
...
+        if (theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE || theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE) {
+            hcThemes.push_back(theme);
+        }
+    }
+
+    for (int i = 0; i < static_cast<int>(hcThemes.size()); ++i) {
+        if (hcThemes[i].codeKey == (uiConfiguration()->currentTheme()).codeKey) {
+            return i;
+        }
+    }
```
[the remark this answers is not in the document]
ArjunBoi: Done!
I copied this part of the code from `appearancepreferencesmodel.cpp`
Should I make this change over there too?

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
```diff
@@ -0,0 +1,87 @@
...
+    ThemeList hcThemes;                                        //creation of hcThemes ensures that we only search in
+                                                               //a list of High Contrast Themes and not the entire list of Themes
+    for (const ThemeInfo& theme : uiConfiguration()->themes()) {
+        if (theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE || theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE) {
+            hcThemes.push_back(theme);
+        }
+    }
+
+    for (int i = 0; i < static_cast<int>(hcThemes.size()); ++i) {
+        if (hcThemes[i].codeKey == (uiConfiguration()->currentTheme()).codeKey) {
```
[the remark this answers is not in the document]
ArjunBoi: Fixed!

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+void AccessibilityPreferencesModel::setCurrentThemeIndex(int index)
+{
+    ThemeList hcThemes;
+
+    for (const ThemeInfo& theme : uiConfiguration()->themes()) {
+        if (theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE || theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE) {
+            hcThemes.push_back(theme);
+        }
+    }
```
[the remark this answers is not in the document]
ArjunBoi: I don't think adding those inline functions in `uitypes.h` would be of much help...
There aren't many moments in the logic where I need just the ThemeCodes of the themes, and I'm afraid that introducing more and more variables with the word "theme" will decrease code comprehension (it was already hard keeping track of all the "Theme" related variables and what they do 😅 )

> ```using ThemeCodeList = std::vector<ThemeCode>;```

From what I understand, adding this bit of code will replace my `if (theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE || theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE)` statement, and allow me to directly use a for loop to iterate over the desired themes based on their CodeKey inside this vector.

However, I still want to raise the same point that creating more variables will be confusing later on. I think it is more expressive and direct the way it currently is. I have used that loop twice in the code, but if you feel that it is still redundant, or if I find that I need to use it more than twice, I can shift that logic to a private function in the same file and call that wherever I need for a bit more tidiness :)

(Also, something very similar to what you suggested already exists in uitypes.h i.e. `using ThemeList = std::vector<ThemeInfo>;`
This is logically very different of course, but very similar in naming)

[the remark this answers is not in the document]
ArjunBoi: That said though, I see what you mean when you say that this would optimize certain areas of the code. 🤔

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    return result;
+}
+
+int AccessibilityPreferencesModel::currentThemeIndex() const
+{
+    ThemeList hcThemes;                                        //creation of hcThemes ensures that we only search in
+                                                               //a list of High Contrast Themes and not the entire list of Themes
```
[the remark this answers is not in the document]
ArjunBoi: Undestood!

src/appshell/view/preferences/accessibilitypreferencesmodel.h
```diff
@@ -0,0 +1,58 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#ifndef MU_APPSHELL_ACCESSIBILITYPREFERENCESMODEL_H
+#define MU_APPSHELL_ACCESSIBILITYPREFERENCESMODEL_H
+
+#include <QObject>
+//#include "appearancepreferencesmodel.h"
```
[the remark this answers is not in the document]
ArjunBoi: Done.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -72,10 +74,16 @@ QString AppearancePreferencesModel::wallpapersDir() const
 
 int AppearancePreferencesModel::currentThemeIndex() const
 {
-    ThemeList themes = allThemes();
+    ThemeList genThemes;                                        //creation of genThemes ensures that we only search in
+                                                                //a list of General Themes and not the entire list of Themes
```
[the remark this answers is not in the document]
ArjunBoi: Also Fixed.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -38,16 +38,18 @@ AppearancePreferencesModel::AppearancePreferencesModel(QObject* parent)
...
-        result << ThemeConverter::toMap(theme);
+        if (theme.codeKey == LIGHT_THEME_CODE || theme.codeKey == DARK_THEME_CODE) {
+            result << ThemeConverter::toMap(theme);
+        }
     }
 
-    return result;
-}
+    return result;                                                                      //only returns General Themes
+}                                                                                       //i.e. excludes HC Themes
```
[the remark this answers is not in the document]
ArjunBoi: Removed the comment entirely.
The name of the function is self-explanatory.

src/appshell/view/preferences/preferencesmodel.cpp
```diff
@@ -145,20 +145,25 @@ void PreferencesModel::load(const QString& currentPageId)
 
     m_rootItem = new PreferencePageItem();
 
-    QList<PreferencePageItem*> generalItems {
+    QList<PreferencePageItem*> generalItems {                                                               //Child items for General Tab
```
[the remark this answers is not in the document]
ArjunBoi: Understood! 😅 
Fixed.

src/framework/ui/internal/platform/macos/macosplatformtheme.mm
```diff
@@ -77,7 +77,7 @@
 {
     if (isSystemDarkMode()) {
         if (isSystemHighContrast()) {
-            return HIGH_CONTRAST_THEME_CODE;
+            return HIGH_CONTRAST_BLACK_THEME_CODE;
```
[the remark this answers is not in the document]
ArjunBoi: Fixed and Replaced!

src/framework/uicomponents/qml/MuseScore/UiComponents/ColorPicker.qml
```diff
@@ -83,7 +85,7 @@ Rectangle {
             name: "NORMAL"
             when: !clickableArea.containsMouse && !colorDialog.visible
 
-            PropertyChanges { target: root; border.color: ui.theme.buttonColor }
+            PropertyChanges { target: root; border.color: overrideDefaultColors? setBorderColor : ui.theme.buttonColor }
```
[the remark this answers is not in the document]
ArjunBoi: Understood!
Removed the changes I had to make :)

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
```diff
@@ -0,0 +1,99 @@
...
+            font: ui.theme.bodyBoldFont
+        }
+
+        Repeater {
+            width: parent.width
+            //spacing: 12
+            //orientation: Qt.Vertical
```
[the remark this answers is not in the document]
ArjunBoi: I am going to be experimenting more between ListView and Repeater because the spacing still isn't exactly what I want on that page...

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
ArjunBoi: Understood!
This fixed the spacing right up, no need for the "Magic Number" anymore :)

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: Again, a result of my experiments and me forgetting to clean up afterwards. :(
Fixed.

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
ArjunBoi: Ahhh I see...
Also fixed.

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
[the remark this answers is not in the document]
cbjeukendrup: Yeah, good idea!

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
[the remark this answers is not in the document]
cbjeukendrup: Good to know: Repeater does not manage the layout of its items. It only fills its parent, with one item for each entry in its model. So you need to put it inside (for example) a column to make it work. 
And it's very okay to use nested columns to work with different spacings!

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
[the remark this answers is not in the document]
ArjunBoi: Done!

src/appshell/qml/Preferences/AccessibilityPreferencesPage.qml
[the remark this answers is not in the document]
ArjunBoi: Thanks for this Casper!
I tried this and it fixed the spacing issue right up :)

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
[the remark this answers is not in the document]
cbjeukendrup: If we use consistent naming, then I think it should not be too complicated. a `Theme` is a struct with all the information for a theme, a `ThemeCode` is only a key for a theme. That's doable, isn't it? 
But I agree that things become unclear when the naming is not consistent. For example, `uiConfiguration()->setCurrentTheme(...)` takes a `ThemeCode`, so it is named wrongly. Feel free to adjust that. 

With a function for getting `highContrastThemeCodes`, you could simplify for example `AccessibilityPreferencesModel::setCurrentThemeIndex` like this:
```cpp
void AccessibilityPreferencesModel::setCurrentThemeIndex(int index)
{
    ThemeCodeList hcThemeCodes = highContrastThemeCodes();
    if (index < 0 || index >= static_cast<int>(hcThemeCodes.size())) {
        return;
    }

    if (hcThemeCodes[index] == uiConfiguration()->currentThemeCodeKey) {
        return;
    }

    uiConfiguration()->setCurrentTheme(hcThemes[index].codeKey);
    emit themesChanged();
}
```
Also `AccessibilityPreferencesModel::currentThemeIndex()` and the corresponding methods of AppearancePreferencesModel can be simplified.

src/appshell/view/preferences/accessibilitypreferencesmodel.h
ArjunBoi: Thanks for the help Casper!
I've got this working now :)

src/appshell/view/preferences/accessibilitypreferencesmodel.cpp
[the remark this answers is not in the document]
ArjunBoi: I have implemented something akin to this in this form:
```c++
bool UiTheme::isCurrentThemeHighContrast() const
{
    return configuration()->currentTheme().codeKey == HIGH_CONTRAST_BLACK_THEME_CODE
           || configuration()->currentTheme().codeKey == HIGH_CONTRAST_WHITE_THEME_CODE;
}

bool UiTheme::isCurrentThemeGeneral() const
{
    return configuration()->currentTheme().codeKey == LIGHT_THEME_CODE || configuration()->currentTheme().codeKey == DARK_THEME_CODE;
}
```

This is different from what you were suggesting, since the use-cases are different, but this still saves me a lot of time whenever i need a particular feature active only for certain themes. :)

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
```diff
@@ -49,26 +49,85 @@ PreferencesPage {
...
-            themes: appearanceModel.themes
-            currentThemeIndex: appearanceModel.currentThemeIndex
+        CheckBox {
+            id: highContrastEnable
+            width: 200
+            checked: appearanceModel.enableHighContrastChecked()
 
-            onThemeChangeRequested: {
-                appearanceModel.currentThemeIndex = newThemeIndex
+            text: "Enable High-Contrast"
```
RomanPudashkin, line 62: You missed qsTrc here

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
```diff
@@ -142,5 +201,14 @@ PreferencesPage {
...
             }
         }
+
+        FlatButton {
+            text: qsTrc("appshell", "Reset Theme to Default")
+
+            onClicked: {
+                appearanceModel.resetThemeToDefault()
+            }
+
```
RomanPudashkin, line 211: Please, remove the empty line

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
```diff
@@ -49,26 +49,85 @@ PreferencesPage {
...
+        }
 
-            colors: appearanceModel.accentColors
-            currentColorIndex: appearanceModel.currentAccentColorIndex
-            firstColumnWidth: parent.firstColumnWidth
+        Component {
+            id: generalThemesModel
+
+            Column {
+                spacing: 24
```
RomanPudashkin, line 77: Better to create a private readonly property and reuse it:

QObject {
id: prv

readonly property int spacing: 24
}

...

Column {
spacing: prv.spacing
}

src/appshell/qml/Preferences/internal/UiColorsSection.qml
```diff
@@ -0,0 +1,74 @@
...
+    property int firstColumnWidth: 0
+
+    signal colorChangeRequested(var newColor, var propertyName)
+
+    StyledTextLabel {
+        text: qsTrc("appshell", "UI Colors")
+        font: ui.theme.bodyBoldFont
+    }
+
+
```
RomanPudashkin, line 40: Please, remove the empty line

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -78,6 +93,51 @@ QStringList AppearancePreferencesModel::accentColors() const
...
+    uiConfiguration()->loadLastUsedHighContrastTheme();
+}
+
+void AppearancePreferencesModel::setNewColor(const QColor& newColor, const QString& propertyName)
+{
+    //! NOTE: Considered using a "switch()" statement here, but it would require a type conversion
+    //! from std::string to some form of a primitive literal. This has a workaround by implementing
+    //! a hash function, but I went for the "if/else if" ladder instead since we don't have that many cases anyway.
+
+    if (propertyName.toStdString() == "Accent Color:") {
```
RomanPudashkin, line 129: Better to create some enum and use it instead of text from UI

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -131,6 +183,17 @@ ThemeList AppearancePreferencesModel::allThemes() const
     return uiConfiguration()->themes();
 }
 
+bool AppearancePreferencesModel::isCurrentThemeHighContrast() const
+{
+    return uiConfiguration()->currentTheme().codeKey == HIGH_CONTRAST_BLACK_THEME_CODE
```
RomanPudashkin, line 188: Better to simplify this method like this (the same for isCurrentThemeGeneral):

static const QStringList contrastThemes {
HIGH_CONTRAST_BLACK_THEME_CODE,
HIGH_CONTRAST_WHITE_THEME_CODE
};

return contrastThemes.contais(currentThemeCode());

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -172,19 +235,17 @@ QString AppearancePreferencesModel::foregroundWallpaperPath() const
...
-    if (index < 0 || index >= static_cast<int>(themes.size())) {
+    if (themeCode == currentThemeCode()) {
         return;
     }
 
-    if (index == currentThemeIndex()) {
-        return;
+    for (const ThemeInfo& theme : allThemes()) {
+        if (themeCode == QString::fromStdString(theme.codeKey)) {
+            uiConfiguration()->setCurrentTheme(theme.codeKey);
```
RomanPudashkin, line 246: It seems you missed break (or return) here

src/appshell/view/preferences/appearancepreferencesmodel.h
```diff
@@ -78,12 +81,17 @@ class AppearancePreferencesModel : public QObject, public async::Asyncable
     QColor foregroundColor() const;
     QString foregroundWallpaperPath() const;
 
+    Q_INVOKABLE void resetThemeToDefault();
+    Q_INVOKABLE bool enableHighContrastChecked();
+    Q_INVOKABLE void loadLastUsedGeneralTheme();
```
RomanPudashkin, line 86: Could you please explain the purpose of this method (same for loadLastUsedHighContrastTheme)?

src/appshell/view/preferences/appearancepreferencesmodel.cpp
RomanPudashkin: UPD: seems like isCurrentThemeHighContrast and isCurrentThemeGeneral were copy-pasted in several places. Better to use UiConfiguration::isCurrentThemeHighContrast() (and UiConfiguration::isCurrentThemeGeneral()) and apply my suggested fixes there

src/notation/internal/notationconfiguration.cpp
```diff
@@ -101,8 +105,33 @@ void NotationConfiguration::init()
         m_backgroundChanged.notify();
     });
 
-    settings()->setDefaultValue(BACKGROUND_COLOR, Val(QColor("#385f94")));
-    settings()->valueChanged(BACKGROUND_COLOR).onReceive(nullptr, [this](const Val&) {
+//    settings()->setDefaultValue(BACKGROUND_COLOR, Val(QColor("#385f94")));
+//    settings()->valueChanged(BACKGROUND_COLOR).onReceive(nullptr, [this](const Val&) {
```
RomanPudashkin, line 109: Better to remove all commented code

src/notation/view/notationpaintview.cpp
```diff
@@ -375,6 +375,12 @@ void NotationPaintView::paintBackground(const RectF& rect, draw::Painter* painte
 {
     QString wallpaperPath = configuration()->backgroundWallpaperPath().toQString();
 
+    qDebug("NotationPaintView::paintBackground ACCESSED!");
+    qDebug("NotationPaintView::paintBackground ACCESSED!");
+    qDebug("NotationPaintView::paintBackground ACCESSED!");
```
RomanPudashkin, line 380: Please, remove this

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
```diff
@@ -184,7 +185,7 @@ Item {
 
         contentWidth: root.popupWidth
         contentHeight: root.height * Math.min(root.count, root.popupItemsCount)
-        padding: 0
+        padding: 1
```
RomanPudashkin, line 188: Could you explain the reason for this change?

src/appshell/qml/Preferences/internal/UiColorsSection.qml
```diff
@@ -0,0 +1,74 @@
...
+    GridLayout {
+        columnSpacing: parent.width/8
+        rowSpacing: 20
+        columns: 2
+        Repeater {
+
+            model: [
+                { textRole: "Accent Color:", colorRole: ui.theme.accentColor},
+                { textRole: "Text and Icons:", colorRole: ui.theme.fontPrimaryColor},
+                { textRole: "Disabled Text:", colorRole: "#000000"},
```
RomanPudashkin, line 50: I don't quite understand why you don't use ui.theme here. Could you please explain?

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: I did!
Fixed :)

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: Missed that one!
Fixed :)

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: Understood!
Should i also make this change for `firstColumnWidth` declared in the same file? i.e. declare it inside the QtObject and refer to it as `prv.firstColumnWidth`?

src/appshell/qml/Preferences/internal/UiColorsSection.qml
ArjunBoi: Fixed :)

src/notation/view/notationpaintview.cpp
ArjunBoi: Ah, Apologies. 
Fixed :)

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
ArjunBoi: Ah, this.
The background rectangle for the Dropdowns (specifically for High-Contrast themes) needed an encompassing border, which is only visible with padding.

I think a more appropriate value would be `ui.theme.borderWidth`, so that it only shows up for High-Contrast themes and not the General Light/Dark ones.

src/notation/internal/notationconfiguration.cpp
ArjunBoi: Fixed :)

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: Apologies, rookie mistake.
Fixed :)

src/appshell/view/preferences/appearancepreferencesmodel.h
ArjunBoi: These methods are called upon checking/unchecking the "Enable High-Contrast" checkbox.
Whenever i switch between the modes (General/High-Contrast) (by checking/unchecking the checkbox), the active theme would be whatever theme i last used in that category. I think this adds a little touch of convenience when switching between themes, both for the user who may have customized a theme in both categories perhaps, and for the developer who may be testing features/differences between a general and HC theme.

for example,
1. I have the Dark theme active
2. I then check the high-contrast checkbox, and i will get defaulted to the HC-Black theme (only when switching to HC-Mode for the *first* time)
3. I switch to the HC-White theme instead, because it looks cool
4. I uncheck the checkbox, and am taken back to whichever general theme I used last, which in this case was Dark theme
5. Upon checking the box again, I would have the HC-White theme active again, since that is the theme i used last in the HC Category

The last used General/HC Theme values are stored in `UI_PREVIOUS_GENERAL_THEME` and `UI_PREVIOUS_HIGH_CONTRAST_THEME` variables respectively, in `uiconfiguration.cpp`.

src/appshell/qml/Preferences/internal/UiColorsSection.qml
ArjunBoi: From what i understand, you're asking about the "Disabled Text" property, right?

As of right now, the Disabled Text property isn't tied to anything i.e. changing the colors of that property doesn't really do anything.
From what i've been able to decipher, we don't have a standard way of referencing "Disabled" items. However, Each Item in QML has an `enabled` property and QWidget has an `enabled` member. Perhaps i need to create a new Theme variable (much like the ones in the `ThemeStyleKey` enum in `uitypes.h`, and then connect it across all UI elements to have a color for disabled properties 🤔 .

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
RomanPudashkin: "Should i also make this change for firstColumnWidth declared in the same file?"

Yes, you can move this property to prv also

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
RomanPudashkin: Yes, better to replace this magic number with ui.theme.borderWidth

src/appshell/qml/Preferences/internal/UiColorsSection.qml
RomanPudashkin: >From what i understand, you're asking about the "Disabled Text" property, right?

Yes, right

>Perhaps i need to create a new Theme variable (much like the ones in the ThemeStyleKey enum in uitypes.h)

It would be great

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -78,6 +93,51 @@ QStringList AppearancePreferencesModel::accentColors() const
...
 
+void AppearancePreferencesModel::resetThemeToDefault()
+{
+    uiConfiguration()->resetCurrentThemeToDefault(currentTheme().codeKey);
+    notationConfiguration()->resetCurrentBackgroundColorToDefault();
+    emit backgroundColorChanged();
+    emit themesChanged();
+}
+
+bool AppearancePreferencesModel::enableHighContrastChecked()
```
RomanPudashkin, line 104: I believe you can use isCurrentThemeHighContrast directly in qml (after marking it as Q_INVOKABLE). You can remove  enableHighContrastChecked after those changes

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: Fixed :)

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
ArjunBoi: Fixed :)

src/appshell/qml/Preferences/internal/UiColorsSection.qml
ArjunBoi: Understood!
As of now, I'm working on the navCtrl borders (nearly done, will post an update within the next few hours). 
This isn't as high of a priority as score-page color inversion is, so i've decided to hold off on fixing this for a little longer, or until i can finish the score inversion feature.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: I apologize, for i don't know how to properly implement an enum in this case. But i understand why using text from UI is just bad practice.

Should i create a new `propertyKey` variable in `UiColorsSection.qml` like so:
```
model: [
                { textRole: "Accent Color:", colorRole: ui.theme.accentColor, propertyKey: 0 },
                { textRole: "Text and Icons:", colorRole: ui.theme.fontPrimaryColor, propertyKey: 1},
                { textRole: "Disabled Text:", colorRole: "#000000", propertyKey: 2},
                { textRole: "Border Color:", colorRole: ui.theme.strokeColor, propertyKey: 3}
            ]
```

and then change the if/else-if ladder in `appearancepreferencesmodel.cpp` to a `switch()` statement like so:
```
switch(propertyKey)
{
    case 0: uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::ACCENT_COLOR, Val(newColor)); 
        break;
    
    case 1: uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::FONT_PRIMARY_COLOR, Val(newColor)); 
        break;
    
    case 2: return; //because a common way to refer to disabled items/text isn't fully implemented yet
    
    case 3: uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::STROKE_COLOR, Val(newColor)); 
        break;

    default: qDebug("Setting not Found!");    
```


Again, I'm suggesting this because i do not fully understand the enum approach.

RomanPudashkin: You can declare an enum in appearancepreferencesmodel.h (see navigationpanel.h for example):

Screenshot 2021-08-03 at 18 15 30

Like this:

ColorType {
AccentColor,
TextAndIconsColor,
...
}
Q_ENUM(ColorType)

You can use this enum in both qml and cpp

RomanPudashkin: So, your function will be like this:

Q_INVOKABLE void setNewColor(const QColor& newColor, ColorType colorType)

cbjeukendrup: And in QML you access the enum values like 
```qml
AppearancePreferencesModel.AccentColor
AppearancePreferencesModel.TextAndIconsColor
```
You need to use AppearancePreferencesModel instead of the name of the enum in C++. This is by the design of Qt.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: > Better to simplify this method like this

This is how the code was turning out to be. 

```c++
bool UiConfiguration::isCurrentThemeHighContrast() const
{
    static const QStringList highContrastThemes {
        QString::fromStdString(HIGH_CONTRAST_BLACK_THEME_CODE),
        QString::fromStdString(HIGH_CONTRAST_WHITE_THEME_CODE)
    };

    return highContrastThemes.contains(QString::fromStdString(currentThemeCodeKey()));
}
```

To try to reduce the overheads and conversions, i tried to implement it in a way that used `std::string` itself, and came up with this:
```c++
static const std::vector<std::string> highContrastThemes {
        HIGH_CONTRAST_BLACK_THEME_CODE,
        HIGH_CONTRAST_WHITE_THEME_CODE
    };
    
    if(std::find(highContrastThemes.begin(), highContrastThemes.end(), currentThemeCodeKey()) != highContrastThemes.end())
    {
        return true;
    } else {
        return false;
    }
```

Should i go with this instead?

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: Understood!
Enum implemented, and is working properly :).
Thank you for the tip Casper!

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: I have removed `isCurrentThemeHighContrast` from `appearancepreferencesmodel.cpp` completely, as suggested. `isCurrentThemeHighContrast` and `isCurrentThemeGeneral` now lie in `uiconfiguration.cpp`. This is what `enableHighContrastChecked()` looks like now:
```c++
bool AppearancePreferencesModel::enableHighContrastChecked()
{
    return uiConfiguration()->isCurrentThemeHighContrast();
}
```
I need the Q_INVOKABLE function (i.e. `enableHighContrastChecked()`) inside AppearancePreferencesModel class for it to work, which is why i can't remove it from here. (unless there is a way to use Q_INVOKABLE functions from different files?)

From what I understand, `uiconfiguration.cpp` file represents the lowest "ModelData" layer, and it sends data up to `uitheme.cpp` and other files i.e. the "ViewModel" layer. I think using a function directly in QML from `uiconfiguration.cpp` will mess up this layering system.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
RomanPudashkin: In this case, I more like the first implementation for this method :) We can just slightly improve it:

ThemeCode currentThemeCode = uiConfiguration()->currentTheme().codeKey;
return currentThemeCode == HIGH_CONTRAST_BLACK_THEME_CODE || currentThemeCode == HIGH_CONTRAST_WHITE_THEME_CODE;

src/appshell/view/preferences/appearancepreferencesmodel.cpp
RomanPudashkin: >From what I understand, uiconfiguration.cpp file represents the lowest "ModelData" layer, and it sends data up to uitheme.cpp and other files i.e. the "ViewModel" layer. I think using a function directly in QML from uiconfiguration.cpp will mess up this layering system.


Yes, you are absolutely right :) 

There were 2 different isCurrentThemeHighContrast (one in the model, and another in the configuration) and separately enableHighContrastChecked that just called isCurrentThemeHighContrast of the model

Your suggested solution is good: we will have Q_INVOKABLE enableHighContrastChecked in the model and isCurrentThemeHighContrast in the configuration

RomanPudashkin:
I want to merge this PR. Please, push all review fixes. You can continue your work in a separate PR

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: > In this case, I more like the first implementation for this method :)
You mean the original implementation, I believe?

Since this function has now found a new home in `uiconfiguration.cpp`, this is what it looks like now, after all suggested changes:
```c++
bool UiConfiguration::isCurrentThemeGeneral() const
{
    ThemeCode currentThemeCode = currentThemeCodeKey();

    return currentThemeCode == DARK_THEME_CODE || currentThemeCode == LIGHT_THEME_CODE;
}
```

and

```
bool UiConfiguration::isCurrentThemeHighContrast() const
{
    ThemeCode currentThemeCode = currentThemeCodeKey();

    return currentThemeCode == HIGH_CONTRAST_BLACK_THEME_CODE || currentThemeCode == HIGH_CONTRAST_WHITE_THEME_CODE;
}
```

ArjunBoi:
I have some changes from the navCtrl border issue that collides with the reviewed files...

Should i just comment them out so that this branch compiles? This means some of the commented code will make it into the master branch with this PR. I will un-comment it in the new PR so I'm back at the same point of progress.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: Understood!
I've made the appropriate changes :)

ArjunBoi:
Huh, It compiles fine for me.
The failing step is the upload of the artifact to GitHub.
I have an unreliable internet connection, so maybe a connection break at that time caused the Linux and MacOS tests to fail? 🤔

Jojo-Schmitz:
I'd say this is pure and plain GitHub problem, not your's

     Error: We are currently unable to download the log. Please try again later.

Jojo-Schmitz:
Not related at all to your internet connection.

Other PRs have the same issue

ArjunBoi:
Ah, I see
Thanks for the clarification 😄

src/appshell/view/preferences/preferencesmodel.cpp
```diff
@@ -146,7 +146,7 @@ void PreferencesModel::load(const QString& currentPageId)
     m_rootItem = new PreferencePageItem();
 
     QList<PreferencePageItem*> generalItems {
-        makeItem("general-start", qtrc("appshell", "Programme start"), IconCode::Code::NONE,
+        makeItem("general-start", qtrc("appshell", "Programme Start"), IconCode::Code::NONE,
```
Jojo-Schmitz, line 149: `"Program start"`, US spelling and title case

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
```diff
@@ -42,33 +42,97 @@ PreferencesPage {
...
+        }
 
-        ThemesSection {
-            width: parent.width
+        CheckBox {
+            id: highContrastEnable
+            width: 200
+            checked: appearanceModel.enableHighContrastChecked()
+
+            text: qsTrc("appshell", "Enable High-Contrast")
```
Jojo-Schmitz, line 67: Sentence case, not Title Case

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
```diff
@@ -142,5 +206,13 @@ PreferencesPage {
                 appearanceModel.foregroundWallpaperPath = newWallpaperPath
             }
         }
+
+        FlatButton {
+            text: qsTrc("appshell", "Reset Theme to Default")
```
Jojo-Schmitz, line 211: same

src/appshell/qml/Preferences/internal/UiColorsSection.qml
```diff
@@ -0,0 +1,74 @@
...
+Column {
+    id: root
+    spacing: 18
+
+    property int firstColumnWidth: 0
+
+    signal colorChangeRequested(var newColor, var propertyType)
+
+    StyledTextLabel {
+        text: qsTrc("appshell", "UI Colors")
```
Jojo-Schmitz, line 37: same

src/appshell/qml/Preferences/internal/UiColorsSection.qml
```diff
@@ -0,0 +1,74 @@
...
+        columnSpacing: parent.width/8
+        rowSpacing: 20
+        columns: 2
+        Repeater {
+
+            model: [
+                { textRole: "Accent Color:", colorRole: ui.theme.accentColor, typeRole: AppearancePreferencesModel.AccentColor},
+                { textRole: "Text and Icons:", colorRole: ui.theme.fontPrimaryColor, typeRole: AppearancePreferencesModel.TextAndIconsColor},
+                { textRole: "Disabled Text:", colorRole: "#000000", typeRole: AppearancePreferencesModel.DisabledColor},
+                { textRole: "Border Color:", colorRole: ui.theme.strokeColor, typeRole: AppearancePreferencesModel.BorderColor}
```
Jojo-Schmitz, line 51: same for these 4

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -78,6 +93,48 @@ QStringList AppearancePreferencesModel::accentColors() const
...
+
+    case TextAndIconsColor: uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::FONT_PRIMARY_COLOR, Val(newColor));
+        break;
+
+    case DisabledColor: return;
+
+    case BorderColor: uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::STROKE_COLOR, Val(newColor));
+        break;
+
+    default: qDebug("INVALID COLORTYPE RECEIVED!");
```
Jojo-Schmitz, line 133: No need to shout ;-), a normal sentence works too here

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
Jojo-Schmitz: (But yet, our policy there just changed recently)

src/appshell/view/preferences/appearancepreferencesmodel.cpp
Eism: just remove the default case

Jojo-Schmitz: True, has the added benefit of a compiler warning if new cases get added but not dealt with here

src/appshell/view/preferences/preferencesmodel.cpp
ArjunBoi: Converted to `"Program start"`
Fixed :)

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: Converted to `"Enable high-contrast"`
Fixed.

src/appshell/qml/Preferences/AppearancePreferencesPage.qml
ArjunBoi: Converted to `"Reset theme to default"`
Fixed.

src/appshell/qml/Preferences/internal/UiColorsSection.qml
ArjunBoi: Converted to `"UI colors"`
Fixed.

src/appshell/qml/Preferences/internal/UiColorsSection.qml
ArjunBoi: Converted to `"Accent color:"`, `"Text and icons:"`, `"Disabled text:"`, `"Border color:"`
Fixed :)

src/appshell/view/preferences/appearancepreferencesmodel.cpp
ArjunBoi: Ah, apologies 😅 
Converted to `"Invalid ColorType Received."`
Fixed

cbjeukendrup: I'd still say it's even better to remove the `default:` entirely. At the moment, it is not needed, and without it, as Jojo says, the compiler will warn when somebody adds a new case to the enum and forgets to handle that case here.

ArjunBoi: Ah understood
I will remove the default case entirely.

RomanPudashkin (approved):

PR #8281, opened 2021-06-08
[MU4] Midi mapping

Eism:


src/framework/global/utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include "utils.h"
+
+#include "translation.h"
+
+using namespace mu;
+
+static const std::vector<std::string> noteHeadNamesLower = {
+    trc("global", "c"),
```
cbjeukendrup, line 29: I believe these static variables won't work well with translations. They are translated when they are needed the first time, but not updated when the user changes the app language. Instead you would need to use QT_TRANSLATE_NOOP and then call trc or qtrc every time the string is needed.

src/framework/shortcuts/view/editmidimappingmodel.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "editmidimappingmodel.h"
+
+#include "log.h"
+#include "utils.h"
+#include "translation.h"
+
+using namespace mu::shortcuts;
+using namespace mu::midi;
+
+static const QString WAITING_STATE = mu::qtrc("shortcuts", "Waiting...");
```
cbjeukendrup, line 32: Same about the translations. And isn't it too much to use a named constant here? It's used only one time.

src/framework/shortcuts/view/mididevicemappingmodel.cpp
```diff
@@ -23,20 +23,24 @@
...
 
 static const QString TITLE_KEY("title");
 static const QString ICON_KEY("icon");
 static const QString STATUS_KEY("status");
 static const QString ENABLED_KEY("enabled");
+static const QString MAPPED_TYPE_KEY("mappedType");
 static const QString MAPPED_VALUE_KEY("mappedValue");
 
-inline std::vector<std::string> allMidiActions()
+inline std::vector<ActionCode> allMidiActions()
```
cbjeukendrup, line 43: There's even a ActionCodeList type alias for this

src/framework/shortcuts/view/mididevicemappingmodel.cpp
```diff
@@ -51,11 +55,12 @@ inline std::vector<std::string> allMidiActions()
         "pad-note-16",
         "pad-note-32",
         "pad-note-64",
+        "undo",
```
cbjeukendrup, line 58: Isn't it a bit of an illogical order to have undo between the note input related actions?

src/framework/global/utils.cpp
Eism: thanks, fixed

src/framework/shortcuts/view/mididevicemappingmodel.cpp
Eism: good question, @jessjwilliamson FYI

RomanPudashkin (approved):

src/framework/shortcuts/view/mididevicemappingmodel.cpp
jessjwilliamson: @Eism This was just reflection from MS3. Also, I realised there are some actions missing from the list - Stop and Real time advance

Eism: Real-time advance has not yet been ported from MU3

Eism: I will add stop action
@jessjwilliamson what about the ordering?

jessjwilliamson: Hi, really sorry, but the 'Stop' action is no use as I remember we replaced this with having a play and pause button instead. The stop button is not even in the play panel.

PR #8285, opened 2021-06-08
[MU3 Backend] ENG-11: Articulations initially drawn incorrectly

iveshenry18:
Resolves: [ENG-11](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-11): Articulation marks initially drawn in wrong position

Specifically when dragging-and-dropping an XML file into MuseScore, the
articulations didn't layout correctly. This was due to a difference
between `openScore` (called when opening recent or from files) and
`dropEvent` (called when dropping).

~~This commit adds the style and layout-related lines to make sure `score->layout` is called after loading, which fixes the issue.~~

~~There appear to be further differences regarding migration, which may be worth further investigation.~~
 
This commit removes the partially-duplicated code in `dropEvent` and
rather calls `openScore` to handle loading, preventing discrepancies
between the two approaches.

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -2920,20 +2920,11 @@ void MuseScore::dropEvent(QDropEvent* event)
       {
       const QMimeData* dta = event->mimeData();
       if (dta->hasUrls()) {
-            int view = -1;
             foreach(const QUrl& u, event->mimeData()->urls()) {
                   if (u.scheme() == "file") {
                         QString file = u.toLocalFile();
-                        MasterScore* score = readScore(file);
-                        if (score) {
-                              view = appendScore(score);
-                              addRecentScore(score);
-                              }
+                        openScore(file);
                         }
-                  }
-            if (view != -1) {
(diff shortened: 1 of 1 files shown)


oktophonie (approved):

cbjeukendrup:
Wouldn't it be better to just call `openScore` from `dropEvent`? That would avoid code duplication and discrepancies between the different ways to open a score.

iveshenry18:
> Wouldn't it be better to just call `openScore` from `dropEvent`?

I agree that this would be much cleaner (I winced as I copy-pasted code from on method to another). I was avoiding it because I'm not sure all the implications of the migration code, and I wasn't sure whether there was a great way to test for regressions, but I'll look into it!

iveshenry18:
It didn't seem to introduce any regressions as far as I could tell, so I updated the commit to take that approach. Good call!

oktophonie (approved):

vpereverzev (approved):

PR #8288, opened 2021-06-09
[MU4] Fixed showing popups

Eism:


igorkorsukov (approved):

wizofaus:
I assume though this won't fix the "Change time signature" popup? (that you currently can't access the button for at all, but I think your PR here might fix - but the button is then hooked up to a NOT_IMPLEMENTED function...)

Eism:
@wizofaus I only fixed the popup opening because it is a high priority bug

PR #8292, opened 2021-06-10
[MU4] Midi mapping stabilization

Eism:


src/playback/internal/playbackcontroller.cpp
```diff
@@ -46,6 +47,7 @@ static const ActionCode REPEAT_CODE("repeat");
 void PlaybackController::init()
 {
     dispatcher()->reg(this, PLAY_CODE, this, &PlaybackController::togglePlay);
+    dispatcher()->reg(this, STOP_CODE, this, &PlaybackController::pause);
```
igorkorsukov, line 50: Is it better to add a stop method?

vpereverzev (approved):

PR #8293, opened 2021-06-10
[MU4] Various improvements in the notation status bar

RomanPudashkin:


cbjeukendrup:
@RomanPudashkin There is a crash at src/appshell/view/notationstatusbarmodel.cpp:169 when closing a score. I think it would be great it you could fix that too in this PR.

RomanPudashkin:
> @RomanPudashkin There is a crash at src/appshell/view/notationstatusbarmodel.cpp:169 when closing a score (missing nullptr check). I think it would be great it you could fix that too in this PR.

Thank you! Fixed

Eism (approved):

PR #8295, opened 2021-06-10
Fix #8084: two checkmarks in Note Input menu in NoteInputBar

### LINKED ISSUE #8084: [MU4 Issue] Pencil icon dropdown has two ticks instead of one

cbjeukendrup:
Resolves: #8084

This PR solves several issues with the Note Input menu. 

First, the order of the items was wrong, because the items were added by iterating over a QMap, which does not maintain the order (it sorts by key). Now, I changed this into a `std::vector<std::pair<...>>`, so we get the order we want.

Then, I fixed the checkmarks issue. I added a separate action for just toggling note input and setting the note input method to "Steptime (default)". This is also how it was in MuseScore 3.

However, in the original design, there is no checkmark at all; the selected method is just highlighted, so the checkmark is somewhat redundant. So I removed it in the third commit. Here's the two versions, so you can choose:
|With checkmark|Without checkmark|
|-|-|
|Schermafbeelding 2021-06-10 om 23 03 28|Schermafbeelding 2021-06-10 om 23 18 53|

### DIFF
--- a/src/notation/view/noteinputbarmodel.cpp
+++ b/src/notation/view/noteinputbarmodel.cpp
@@ -43,7 +43,7 @@ static const ActionCode TUPLET_ACTION_CODE("tuplet");
 static int INVALID_INDEX = -1;
 
 static const std::vector<std::pair<ActionCode, NoteInputMethod> > noteInputModeActions = {
-    { "note-input", NoteInputMethod::STEPTIME },
+    { "note-input-steptime", NoteInputMethod::STEPTIME },
     { "note-input-rhythm", NoteInputMethod::RHYTHM },
     { "note-input-repitch", NoteInputMethod::REPITCH },
     { "note-input-realtime-auto", NoteInputMethod::REALTIME_AUTO },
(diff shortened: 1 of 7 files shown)


src/notation/view/noteinputbarmodel.cpp
```diff
@@ -834,18 +826,9 @@ std::vector<std::string> NoteInputBarModel::currentWorkspaceActions() const
     return toolbarData->actions;
 }
 
-void NoteInputBarModel::handleAction(const QString& action, int actionIndex)
+void NoteInputBarModel::handleAction(const QString& action)
 {
     ActionCode actionCode = codeFromQString(action);
-    if (isNoteInputModeAction(actionCode)) {
```
RomanPudashkin, line 840: Thank you for this fix!

RomanPudashkin (approved):

PR #8299, opened 2021-06-12
Implement registering recent files with the operating system

cbjeukendrup:
In macOS apps that are built entirely with Apple's SDK, and in some special other apps, the most recently used files appear in the context menu of the app icon in the Dock (among other places). I implemented this also for MuseScore. 
This will enable users to get back to what they were doing even quicker, using the functionality they know from other apps.<sup>1</sup>
RecentFilesMacOS
(normally, there is a separator line between the recent files and the other items in the dock menu, but it's missing due to a bug in macOS, which I reported to Apple.)

Similar functionality appeared to exist on Windows. So I implemented it for Windows too.
Schermafbeelding 2021-06-12 om 13 34 02
NOTE: On Windows, it will only work if the app is [registered](https://docs.microsoft.com/en-us/windows/win32/shell/app-registration) for the correct file types. So in development versions of MuseScore 4 it probably won't work (unless you register it manually in the Register editor, as I did), but in the released version it will, because the .msi installer registers the app (that's at least the case with MuseScore 3).

Also note: clearing the recent files does work on macOS, but does not seem to work on Windows. I did what the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shaddtorecentdocs) seemed to want me to do, but without success. I'm not sure if it's even possible at all.

<sup>1</sup> I've probably watched too much WWDC 😅.

### DIFF
--- a/src/userscores/CMakeLists.txt
+++ b/src/userscores/CMakeLists.txt
@@ -35,6 +35,11 @@ if (OS_IS_MAC)
 
     set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}/internal/platform/macos/macosrecentfilescontroller.mm
                                 PROPERTIES SKIP_PRECOMPILE_HEADERS ON)
+elseif(OS_IS_WIN)
+    set(PLATFORM_SRC
+        ${CMAKE_CURRENT_LIST_DIR}/internal/platform/windows/windowsrecentfilescontroller.cpp
+        ${CMAKE_CURRENT_LIST_DIR}/internal/platform/windows/windowsrecentfilescontroller.h
+    )
 else()
     set(PLATFORM_SRC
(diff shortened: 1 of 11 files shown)


RomanPudashkin (approved):

RomanPudashkin:
@cbjeukendrup 

I have found one small issue. Those files have been deleted but they are displaying in this context menu. Probably we need to implement some actualization for this menu

Screenshot 2021-06-17 at 12 15 54

cbjeukendrup:
@RomanPudashkin Unfortunately, this seems a general issue in macOS. All apps (even built-in apps) still display removed files in their recent list. I will report this to Apple using the Feedback Assistant.

We can't do much about this; we can only bump a file to the top of the list, or clear the whole list.
Here's the documentation: https://developer.apple.com/documentation/appkit/nsdocumentcontroller/1514967-notenewrecentdocumenturl?language=objc

PR #8301, opened 2021-06-13
[MU3] fix #296212 - formatting lost when copying/pasting text

wizofaus:
Resolves: https://musescore.org/en/node/296212#comment-1084020

Copying & pasting text from within text elements on a score doesn't behave as expected - formatting is lost and sometimes apparently randomly re-applied. There's also a bug with deleting all text and using "undo" to restore (formatting not properly restored if musical symbols used).

However a bigger problem is that copy/paste etc. doesn't seem to work at all currently in MU4 which is blocking the ability to test these changes (they were tested on the 3.6.2 branch only). I'm just creating this so the work doesn't get lost.

### DIFF
--- a/src/engraving/libmscore/textbase.h
+++ b/src/engraving/libmscore/textbase.h
@@ -158,8 +158,8 @@ public:
     QString currentWord() const;
     QString currentLine() const;
     bool set(const mu::PointF& p, QTextCursor::MoveMode mode = QTextCursor::MoveAnchor);
-    QString selectedText() const;
-    QString extractText(int r1, int c1, int r2, int c2) const;
+    QString selectedText(bool withFormat = false) const;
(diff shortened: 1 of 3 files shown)


src/engraving/libmscore/textbase.h
```diff
@@ -157,7 +157,7 @@ class TextCursor
     QChar currentCharacter() const;
     QString currentWord() const;
     QString currentLine() const;
-    bool set(const mu::PointF& p, QTextCursor::MoveMode mode = QTextCursor::MoveAnchor);
+    bool set(const QPointF& p, QTextCursor::MoveMode mode = QTextCursor::MoveAnchor);
```
wizofaus, line 160: fix

Jojo-Schmitz:
Is this now obsolete with #8306?

wizofaus:
Yep, surprised github doesn't pick up that it's a NO-OP. Closing.

PR #8317, opened 2021-06-15
[MU4] [Multiprojects] Added check and activation opened score  

igorkorsukov:
Including https://github.com/musescore/MuseScore/pull/8316

https://user-images.githubusercontent.com/3818029/122050643-f1a00b00-cde3-11eb-98f1-f5bf62a98952.mp4

### DIFF
--- a/src/multiinstances/internal/ipc/ipc.h
+++ b/src/multiinstances/internal/ipc/ipc.h
@@ -48,7 +48,7 @@ enum class Code {
     Undefined = -1,
     Success = 0,
     Timeout,
-    AllRecevied
+    AllAnswered
 };
 
 enum class MsgType {
@@ -66,11 +66,7 @@ struct Msg
     QString method;
     QStringList args;
 
-        << << << < HEAD
-        bool isValid() const { return type != MsgType::Undefined && !method.isEmpty(); }
-        ==
-        == ===bool isValid() const { return !method.isEmpty(); }
-        >> >> >> > moved ipc server to separated thread
+    bool isValid() const { return type != MsgType::Undefined && !method.isEmpty(); }
 };
 
 static void serialize(const Msg& msg, QByteArray& data);
--- a/src/multiinstances/internal/ipc/ipc.h
+++ b/src/multiinstances/internal/ipc/ipc.h
@@ -58,7 +58,11 @@ struct Msg
     QString method;
     QStringList args;
 
-    bool isValid() const { return type != MsgType::Undefined && !method.isEmpty(); }
+        << << << < HEAD
+        bool isValid() const { return type != MsgType::Undefined && !method.isEmpty(); }
+        ==
+        == ===bool isValid() const { return !method.isEmpty(); }
+        >> >> >> > moved ipc server to separated thread
 };
 
 struct Meta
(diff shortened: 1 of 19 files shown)


src/multiinstances/internal/ipc/ipcloop.cpp
```diff
@@ -0,0 +1,43 @@
...
+IpcLoop::IpcLoop()
+{
+    m_timer.setSingleShot(true);
+    QObject::connect(&m_timer, &QTimer::timeout, [this]() {
+        exit(Code::Timeout);
+    });
+}
+
+Code IpcLoop::exec(int timeout)
+{
```
vpereverzev, line 35: might make sense to protect it with m_timer.isActive()

src/multiinstances/internal/ipc/ipcchannel.cpp
```diff
@@ -57,9 +61,83 @@ bool IpcChannel::send(const Msg& msg)
...
+{
+    Msg res;
+    res.destID = BROADCAST_ID;
+    res.type = MsgType::Notify;
+    res.method = method;
+    res.args = args;
+    send(res);
+}
+
+Code IpcChannel::syncRequestToAll(const QString& method, const QStringList& args, const OnReceived& onReceived)
```
vpereverzev, line 84: might make sense to call it syncBroadcast

I'm a bit curious to know what is the use case of this method? Kinda to broadcast message and make sure that everyone has received it?

src/multiinstances/internal/ipc/ipc.h
```diff
@@ -44,6 +44,13 @@ static const QString IPC_WHOIS("ipc_whois");
 static const QString IPC_METAINFO("ipc_metainfo");
 static const QString IPC_PING("ipc_ping");
 
+enum class Code {
+    Undefined = -1,
+    Success = 0,
+    Timeout,
+    AllRecevied
```
vpereverzev, line 51: it's a bit confusing, let's say I've sent a message via syncBroadcast, so which one should I expect as the result - Success or AllReceived (btw there is typo here). I mean they look similiar to me

src/multiinstances/internal/ipc/ipcchannel.cpp
igorkorsukov: Just broadcast not suitable for this case, because the term broadcast often using just for notify. In this case, we asking all (do request for all) and waiting for an answer (response) from all.

Like, has anyone opened this one score?
not
not
yes
not

src/multiinstances/internal/ipc/ipc.h
igorkorsukov: > it's a bit confusing, let's say I've sent a message via syncBroadcast, so which one should I expect as the result - Success or AllReceived (btw there is typo here). I mean they look similiar to me

The idea here is that we do request for all, and put callback, that checking responses. If someone responses success for us, we break wait others, and will be returned Success.
If no one answers us with success, and there is no one else to wait (everyone answered), then the result will be returned AllRecevied (maybe better to rename AllAnswered)

Like, 
1. Has anyone opened this one score?
not
not
yes -> Success for us 
-- break --
not

2. Has anyone opened this one score?
not
not
not  
not
-> AllReceived

vpereverzev: Yeah, I see now. Perhaps it would make sense to have the "retval" semantic, but I'm not quite sure how to put it in here

igorkorsukov: renamed to `AllAnswered`

src/multiinstances/internal/ipc/ipcloop.cpp
igorkorsukov: added

vpereverzev (approved):

PR #8318, opened 2021-06-15
[MU4] fix #7706 fix drag & drop memory leaks

### LINKED ISSUE #7706: [MU4 Issue] Memory leaks after drag & drop elements to a score
**Describe the bug**
There are several memory allocations (see _new ScoreCallbacks()_) in NotationInteraction. But we never delete those resources

More details:
https://github.com/musescore/MuseScore/pull/7630

**To Reproduce**

**Expected behavior**
We should free the allocated memory

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/7706

Lots of unnecessary new-ing (and no deleting!) when dragging & dropping

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -1814,7 +1814,7 @@ void Score::toggleAccidental(AccidentalType at, const EditData& ed)
         _is.setRest(false);
     } else {
         if (selection().isNone()) {
-            ed.view->startNoteEntryMode();
+            ed.view()->startNoteEntryMode();
             _is.setAccidentalType(at);
             _is.setDuration(TDuration::DurationType::V_QUARTER);
             _is.setRest(false);
@@ -3894,8 +3894,7 @@ void Score::cmdPitchDown()
     if (el && el->isLyrics()) {
         cmdMoveLyrics(toLyrics(el), Direction::DOWN);
     } else if (el && (el->isArticulation() || el->isTextBase())) {
-        el->undoChangeProperty(Pid::OFFSET, QVariant::fromValue(el->offset() + PointF(0.0,
-                                                                                      MScore::nudgeStep* el->spatium())),
+        el->undoChangeProperty(Pid::OFFSET, QVariant::fromValue(el->offset() + PointF(0.0, MScore::nudgeStep* el->spatium())),
                                PropertyFlags::UNSTYLED);
     } else if (el && el->isRest()) {
         cmdMoveRest(toRest(el), Direction::DOWN);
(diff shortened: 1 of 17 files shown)


src/engraving/libmscore/element.h
```diff
@@ -135,9 +135,10 @@ class ElementEditData;
 class EditData
 {
     QList<ElementEditData*> data;
+    MuseScoreView* view_ { 0 };
 
 public:
-    MuseScoreView* view              { 0 };
```
wizofaus, line 140: Not sure why uncrustify doesn't complain about publicly settable pointer fields!

src/notation/internal/notationinteraction.cpp
```diff
@@ -72,9 +73,8 @@ NotationInteraction::NotationInteraction(Notation* notation, INotationUndoStackP
         }
     });
 
-    m_dragData.ed.view = new ScoreCallbacks();
-    m_dropData.ed.view = new ScoreCallbacks();
-    m_gripEditData.view = new ScoreCallbacks();
+    m_dragData.ed = Ms::EditData(&m_scoreCallbacks);
+    m_dropData.ed = Ms::EditData(&m_scoreCallbacks);
```
wizofaus, line 77: Tiny bit inefficient but this only gets created rarely

src/engraving/libmscore/cmd.cpp
```diff
@@ -3876,7 +3876,7 @@ void Score::cmdPitchUp()
     if (el && el->isLyrics()) {
         cmdMoveLyrics(toLyrics(el), Direction::UP);
     } else if (el && (el->isArticulation() || el->isTextBase())) {
-        el->undoChangeProperty(Pid::OFFSET, el->offset() + PointF(0.0, -MScore::nudgeStep* el->spatium()), PropertyFlags::UNSTYLED);
+        el->undoChangeProperty(Pid::OFFSET, el->offset() + PointF(0.0, -MScore::nudgeStep * el->spatium()), PropertyFlags::UNSTYLED);
```
wizofaus, line 3879: this came from uncrustify...

vshalkevich:
@wizofaus 
CI_utests / run_tests (pull_request)
Test failed. Maybe there are any critical issues

src/notation/internal/notationinteraction.cpp
igorkorsukov: A bit unreliable, potential crash is possible if we pass a reference to a member of one class to another class. (for example, if in the future `m_dragData` and `m_scoreCallbacks` are swapped in the .h file and there will be some call to `m_scoreCallbacks` in the `EditData` destructor)
Maybe it's better to do a `shared_ptr`?
It will also be more correct semantically - the semantics of the code will "say" that `m_scoreCallbacks` is a shared object.

wizofaus:
This is just following the pattern used elsewhere.  The objects all have the same lifetime/owner, so there's no risk of a crash unless an EditData object is copied and allowed to live longer than its parent - making the copy constructor private would probably be a good idea.

igorkorsukov (approved):

vshalkevich:
seems fine for me

PR #8324, opened 2021-06-15
[MU4] Scores uploading

RomanPudashkin:


src/cloud/internal/cloudservice.cpp
```diff
@@ -41,6 +42,17 @@ static const QString REFRESH_TOKEN_KEY("refresh_token");
 static const QString DEVICE_ID_KEY("device_id");
 
 constexpr int USER_UNAUTHORIZED_ERR_CODE = 401;
+constexpr int INVALID_SCORE_ID = 0;
+
+int scoreIdFromSourceUrl(const QUrl& sourceUrl)
```
igorkorsukov, line 47: Better add `static`, else this function will be global

src/cloud/internal/cloudservice.cpp
```diff
@@ -165,7 +169,21 @@ void CloudService::onUserAuthorized()
     m_refreshToken = m_oauth2->refreshToken();
 
     saveTokens();
-    downloadUserInfo();
+
+    if (downloadUserInfo() == RequestStatus::Ok) {
+        m_onUserAuthorizedCallback();
```
igorkorsukov, line 174: I believe need to add `IF_ASSERT_FAILED(m_onUserAuthorizedCallback)`

src/cloud/internal/cloudservice.h
```diff
@@ -55,7 +57,10 @@ class CloudService : public QObject, public IAuthorizationService, public IUploa
     ValCh<bool> userAuthorized() const override;
     ValCh<AccountInfo> accountInfo() const override;
 
-    framework::ProgressChannel uploadScore(const QByteArray& scoreData, const std::string& title, const QUrl& sourceUrl = QUrl()) override;
+    void uploadScore(system::IODevice& scoreSourceDevice, const QString& title, const QUrl& sourceUrl = QUrl()) override;
+
+    async::Channel<QUrl> sourceUrlReceived() const override;
+    framework::ProgressChannel progressChannel() const override;
```
igorkorsukov, line 63: It's strange that `ProgressChannel` in the namespace is a framework, not an namespace async  :)

src/cloud/internal/cloudservice.h
```diff
@@ -76,8 +88,10 @@ private slots:
     network::RequestHeaders headers() const;
 
     RequestStatus downloadUserInfo();
+    RequestStatus doUploadScore(system::IODevice& scoreSourceDevice, const QString& title, const QUrl& sourceUrl = QUrl());
```
igorkorsukov, line 91: It is strange that we have `IODevice` in the system namespace, not io.

src/cloud/internal/cloudservice.h
```diff
@@ -76,8 +88,10 @@ private slots:
...
 
-    void setAccountInfo(const AccountInfo& info);
+    using RequestCallback = std::function<RequestStatus()>;
+    void executeRequest(const RequestCallback& requestCallback);
```
igorkorsukov, line 94: I think it better to use async::Promise

src/cloud/internal/cloudservice.cpp
RomanPudashkin: m_onUserAuthorizedCallback can be invalid (see CloudService::signIn()). This callback is needed when you need something to do right after authorization (for example, uploading of the score will continue if a user wasn't authorized when they pressed on "Upload to MuseScore.com")

igorkorsukov: Then you just need to check `if (m_onUserAuthorizedCallback)`, otherwise it will crash :)

RomanPudashkin: fixed

src/cloud/internal/cloudservice.cpp
RomanPudashkin: fixed

src/cloud/internal/cloudservice.h
RomanPudashkin: Progress is stored in mu::global. ProgressChannel just an alias. We can put ProgressChannel into mu::async, but it seems a little strange for me

src/cloud/internal/cloudservice.h
RomanPudashkin: moved IODevice into mu::io

src/cloud/internal/cloudservice.h
```diff
@@ -66,6 +71,13 @@ private slots:
     bool updateTokens();
     void clearTokens();
 
+    using OnUserAuthorizedCallback = std::function<void ()>;
+    void authorize(const OnUserAuthorizedCallback& onUserAuthorizedCallback = OnUserAuthorizedCallback());
```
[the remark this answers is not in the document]
RomanPudashkin: Discussed this issue with Igor, decided to leave it as it is

PR #8298, opened 2021-06-12
[MU3 Backend] ENG-52: Infer incorrectly exported fingerings

iveshenry18:
Resolves: [ENG-52](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-52): Correct for incorrectly exported fingerings

A common error in exporters is misinterpreting fingerings as word-type
directions. This commit adds the `MusicXMLInferredFingering` class and the
associated code necessary to detect these directions, find the correct
notes to add them to, then add them if such notes are found.

The current iteration is able to distribute the fingerings across voices
if necessary, as well as round the tick value of a direction to the
nearest greatest common denominator of the measure.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8325, opened 2021-06-15
[MU3 Backend] ENG-25: Improve placement defaults

iveshenry18:
Resolves: [ENG-25](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-25): Dynamics should appear above vocal staves

This adds some context-aware placement defaults for text and
dynamics (in the case that neither "placement" nor "default-y" nor
"relative-y" are specified). Namely, both expression text and dynamics
are placed below the staff unless it is a vocal staff.

To this end, there are some simple checks added to discern whether a
part is a vocal part, which is currently a simple check against the
name of the staff and a check for the presence of lyrics. Additionally,
the "expression text" is currently defined by the presence of italics in
the `_wordsText`, and an absence of `_tempoText`.

This PR is built on commits from #8298.


oktophonie (approved):

Jojo-Schmitz:
rebase needed

iveshenry18:
Rebased!

importexport/musicxml/importxmlfirstpass.cpp
```diff
@@ -92,6 +92,19 @@ void MusicXmlPart::calcOctaveShifts()
             }
       }
 
+bool MusicXmlPart::isVocalStaff() const
+      {
+      QList<QString> vocalInstrumentNames({"Voice",
```
vpereverzev, line 97: I'd suggest to use an std::vector here and also make it static, because you don't really need to allocate it on every call

it also makes sense to put it at the top of .cpp as a static constant variable

iveshenry18: Good call!

vpereverzev (approved):

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -2580,16 +2586,22 @@ void MusicXMLParserDirection::direction(const QString& partId,
                         t->setFrameType(FrameType::SQUARE);
                         t->setFrameRound(0);
                         }
+
+                  QString wordsPlacement = placement();
+                  // Case-based defaults
+                  if (wordsPlacement.isEmpty())
+                        if (isVocalStaff) wordsPlacement = "above";
+                        else if (isExpressionText) wordsPlacement = "below";
```
Jojo-Schmitz, line 2594: Here a warning reg. ambigues `else` is generated and additional curly braces are suggested.

PR #8330, opened 2021-06-16
[MU4] Multiprojects. Added sync changes settings between instances 

igorkorsukov:
https://user-images.githubusercontent.com/3818029/122218369-abfc4480-ceae-11eb-9582-0dad303e9589.mp4

### DIFF
--- a/src/framework/global/settings.cpp
+++ b/src/framework/global/settings.cpp
@@ -146,6 +146,8 @@ void Settings::setValue(const Key& key, const Val& value)
         async::Channel<Val> channel = it->second;
         channel.send(value);
     }
+
+    multiInstancesProvider()->settingsSetValue(key.key, value);
 }
 
 void Settings::writeValue(const Key& key, const Val& value)
@@ -207,6 +209,8 @@ void Settings::beginTransaction()
 
     m_localSettings = m_items;
     m_isTransactionStarted = true;
+
+    multiInstancesProvider()->settingsBeginTransaction();
 }
 
 void Settings::commitTransaction()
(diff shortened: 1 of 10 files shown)


src/multiinstances/internal/multiinstancesprovider.cpp
```diff
@@ -73,6 +83,28 @@ void MultiInstancesProvider::onMsg(const Msg& msg)
             mainWindow()->requestShowOnFront();
         }
     }
+    // Settings
+    else if (msg.type == MsgType::Request && msg.method == METHOD_PREFERENCES_IS_OPENED) {
+        bool isOpened = interactive()->isOpened(PREFERENCES_URI).val;
+        m_ipcChannel->response(METHOD_PREFERENCES_IS_OPENED, { QString::number(isOpened) }, msg.srcID);
+    } else if (msg.method == METHOD_SETTINGS_BEGIN_TRANSACTION) {
```
cbjeukendrup, line 90: Wrong key? Should be METHOD_ACTIVATE_WINDOW_WITH_OPENED_PREFERENCES, I think.

igorkorsukov: big thanks, fixed, hmm, before making PR, I tested..., hmm

vpereverzev (approved):

src/framework/global/settings.cpp
```diff
@@ -110,6 +116,10 @@ Settings::Items Settings::readItems() const
         result[item.key] = item;
     }
 
+    if (multiInstancesProvider()) {
```
vpereverzev, line 119: BTW it might make sense to put IF_ASSERT here and below

igorkorsukov: No, just ifs are needed here, and I think it's better to keep them. They are needed in tests, in tests, there is a call to the settings.

Issue #8182, opened 2021-05-26
[MU4 Issue] Palettes icons gone after docking it to Notation


vshalkevich:
**Describe the bug**
Palettes icons gone after docking it to score

**To Reproduce**
1. Create a score
2. Open Palettes tab in the left Instruments Panel
3. Dock any palette to the score

**Expected behavior**
Palettes icons should be presented always

**Screenshots**
https://user-images.githubusercontent.com/71372310/119739586-1e1ed200-be8b-11eb-911b-8c79e5a46470.mp4

**Desktop (please complete the following information):**
Windows 10
MacOS

**Additional context**
Icon returns after switching Instrument Panels

cbjeukendrup:
"docking" → "dragging"?

I think this is related to (if not a duplicate of) https://github.com/musescore/MuseScore/issues/8035.
It seems it just depends on how quickly you drag how far the item moves off, and when you drag quickly enough, it moves outside the visible area of the palette.

By the way, confirmed on macOS too.

Also, it would be great if we could modify the Palettes code in such way that it is possible to have one "stand-alone" palette without a palettes tree. We will need this when rewriting the Master palette (or some replacement that contains only the functionality that's not elsewhere in the UI) and the Key Signature Creator (which contains an Accidentals Palette) from QWidgets to QML.

vshalkevich:
@cbjeukendrup 
Thanks for reply.
Bugs seems similar but I think they are different. Need to check both of them after fixing one of them.

Issue #8035, opened 2021-04-30
[MU4 Issue] Palettes item shifts after addinng on a Score. 


vshalkevich:
**Describe the bug**
If add Palette from Instruments panel to a score, Palette item shifts in Instruments panel. 

**To Reproduce**
1. Create a score
2. Add notes
3. Open Palette tab from the Instruments Panel
4. Expand any of Palette 
5. Add Palette item on the score (stick it to a note)

**Expected behavior**
Palette items should keep their position all the time

**Screenshots**

https://user-images.githubusercontent.com/71372310/116705453-3a873600-a9d5-11eb-919e-cff96421b097.mp4

**Desktop (please complete the following information):**
 Windows 10
 Linux

PR #8333, opened 2021-06-16
[MU4] Various fixes for the palettes panel

RomanPudashkin:
Resolves:
https://github.com/musescore/MuseScore/issues/8182
https://github.com/musescore/MuseScore/issues/8035

And several bugs without any issue

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/MenuButton.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/MenuButton.qml
@@ -27,11 +27,19 @@ FlatButton {
     id: root
 
     property var menuModel
+    property alias isMenuOpened: menuLoader.isMenuOpened
 
     signal handleAction(string actionCode, int actionIndex)
 
+    function toggleMenu(item, x, y) {
+        menuLoader.parent = item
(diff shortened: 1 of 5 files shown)


src/notation/internal/notationparts.cpp
```diff
@@ -1329,7 +1329,9 @@ void NotationParts::updateSoloist(const PartInstrumentList& parts)
 
 void NotationParts::sortParts(const PartInstrumentList& parts, const QList<Ms::Staff*>& originalStaves)
 {
-    Q_ASSERT(!originalStaves.empty());
+    if (originalStaves.empty()) {
+        return;
+    }
 
     Ms::Score* score = originalStaves[0]->score();
```
RomanPudashkin, line 1336: @njvdberg app crashes here for some unclear reasons (after creating a new score that contains several instruments). I have added a check on emptiness, but I'm not sure if this is ok because there was an assert

Eism (approved):

src/notation/internal/notationparts.cpp
RomanPudashkin: Fixed in https://github.com/musescore/MuseScore/pull/8341

vshalkevich:
All is OK. Bugs fixed

PR #8335, opened 2021-06-16
[MU4] fix #8154 - crash on exit (~UiEngine)

### LINKED ISSUE #8154: [MU4 Issue] Access Violation on shutdown (Windows)
**Describe the bug**
Unhandled Access Violation occurs when quitting MuseScore 4 if a score has been opened

notationuiactions.cpp, line 1255
`m_controller->currentNotationInteraction()->scoreConfigChanged().onReceive(this, [this](ScoreConfigType configType)`

**To Reproduce**

1. Build/Run MuseScore4 from VS 2019/Windows 2010
2. Open or create any score
3. Quit

**Expected behavior**
Probably shouldn't crash.

**Screenshots**

 - OS: Windows 10

**Additional context**

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8154

Crash in ~UiEngine

### DIFF
--- a/src/engraving/libmscore/mscore.cpp
+++ b/src/engraving/libmscore/mscore.cpp
@@ -138,7 +138,6 @@ Sequencer* MScore::seq = 0;
 MuseScoreCore* MuseScoreCore::mscoreCore;
 
 extern void initDrumset();
-extern QString mscoreGlobalShare;
 
 std::vector<MScoreError> MScore::errorList {
     { MsError::MS_NO_ERROR,                     0,    0 },
--- a/src/framework/ui/internal/uiengine.cpp
+++ b/src/framework/ui/internal/uiengine.cpp
@@ -31,10 +31,6 @@
 #include "config.h"
 #include "log.h"
 
-namespace Ms {
-extern QString mscoreGlobalShare; //! FIXME Need to remove global variable
-}
-
 using namespace mu::ui;
 
 UiEngine* UiEngine::instance()
@@ -67,6 +63,8 @@ void UiEngine::quit()
     m_engine->quit();
     delete m_engine;
     m_engine = nullptr;
+    delete m_theme;
+    m_theme = nullptr;
 }
 
 QQuickItem* UiEngine::rootItem() const


src/engraving/libmscore/mscore.cpp
```diff
@@ -138,7 +138,6 @@ Sequencer* MScore::seq = 0;
 MuseScoreCore* MuseScoreCore::mscoreCore;
 
 extern void initDrumset();
-extern QString mscoreGlobalShare;
```
wizofaus, line 141: no longer used

src/framework/ui/internal/uiengine.cpp
```diff
@@ -67,6 +63,8 @@ void UiEngine::quit()
     m_engine->quit();
     delete m_engine;
     m_engine = nullptr;
+    delete m_theme;
+    m_theme = nullptr;
```
wizofaus, line 67: this feels a bit more like a workaround than determining the root cause, as it wasn't happening with yesterday's code, but given it's crashing inside QtCore itself I'm not sure it's worth digging too much deeper

RomanPudashkin: There is already deleting of m_theme in UiEngine::~UiEngine(). I afraid this fix will hide some serious problems

wizofaus: I left that in case a UIEngine is created and destroyed before quit() is called, but in normal circumstances it won't do anything.  Deleting m_theme in the destructor definitely causes a crash, I assume because something else in Qt has already been destroyed and made invalid.  But I don't know why it wasn't happening before yesterday.

AntonioBL:
If it may be of help, here is a stack trace from AddressSanitizer under Linux when exiting MuseScore (with or without an open score). A use-after-free happens inside UiEngine::quit() at instruction "delete m_engine".
I tested with a debug build, commit 6b501b834c6e40d8e3cc08122f
[crash_when_exit.txt](https://github.com/musescore/MuseScore/files/6677684/crash_when_exit.txt)

RomanPudashkin:
This crash was fixed here
https://github.com/musescore/MuseScore/pull/8356

PR #8337, opened 2021-06-17
[MU4] fix #8302 - ensure we handle keyPress (via shortcutOverride) etc.

### LINKED ISSUE #8302: [MU4 Issue] Copy/cut/paste don't work in text elements
**Describe the bug**
When editing text inside a text element none of the clipboard operations work at all.
There appears to be no code to even hook these operations up.

Nor do any of these operations work at all via the `cmd( )` function used by plugins.

**Expected behavior**
Should work as per MU3.6

**Screenshots**

 - Windows 10, built with MSVC 2019

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8302

  Text entry was working via keyReleaseEvent which isn't really right as when editing text you expect to see it appear on key *down*/press.  But Qt's shortcut key mechanism was interfering was this, so it needs to be handled via ShortcutOverride events instead. Also fixed a bug where hitting backspace deleted the previous word, because Ctrl had previously been used and the modifier flag was still set. 

### DIFF
--- a/src/notation/view/notationpaintview.cpp
+++ b/src/notation/view/notationpaintview.cpp
@@ -688,7 +688,7 @@ void NotationPaintView::hoverMoveEvent(QHoverEvent* event)
     }
 }
 
-void NotationPaintView::keyReleaseEvent(QKeyEvent* event)
+void NotationPaintView::shortcutOverride(QKeyEvent* event)
 {
     if (isInited()) {
         std::string sequence = QKeySequence(event->key()).toString().toStdString();
@@ -712,6 +712,14 @@ void NotationPaintView::keyReleaseEvent(QKeyEvent* event)
         }
         m_inputController->keyReleaseEvent(event);
     }
+}
+
+bool NotationPaintView::event(QEvent* ev)
+{
+    if (ev->type() == QEvent::Type::ShortcutOverride) {
+        shortcutOverride(static_cast<QKeyEvent*>(ev));
+    }
+    return QQuickPaintedItem::event(ev);
 }
 
 void NotationPaintView::dragEnterEvent(QDragEnterEvent* event)
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -2230,6 +2230,7 @@ void NotationInteraction::deleteSelection()
         auto textBase = toTextBase(m_textEditData.element);
         if (!textBase->deleteSelectedText(m_textEditData)) {
             m_textEditData.key = Qt::Key_Backspace;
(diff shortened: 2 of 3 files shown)


src/notation/view/notationpaintview.cpp
```diff
@@ -714,6 +714,14 @@ void NotationPaintView::keyReleaseEvent(QKeyEvent* event)
     }
 }
 
+bool NotationPaintView::event(QEvent* ev)
+{
+    if (ev->type() == QEvent::Type::ShortcutOverride) {
+        shortcutOverride(static_cast<QKeyEvent*>(ev));
```
wizofaus, line 720: I suppose I could have called it handleShortcutOverride or shortcutOverrideEvent but this isn't a standard virtual override.

wizofaus:
While this is OK to merge in its current state, and certainly fixes some key issues, I do think we need a bit of a rethink as to how shortcut keys are handled.
It seems in MU3 the individual elements get "first bite" as it were, and only if the key isn't handled does the shortcut key lookup kick in. But that means there's a lot of logic inside, e.g. Slur and TextBase etc. to handle what particular keystrokes should do, and for instance, if you remove the shortcut mapping for X (defaults to "flip"), X still works while editing a slur.
In some cases this logic would definitely be better handled outside the element (and in fact the "flip" action does work on a slur, even if mapped to another key, in MU3, but not in MU4 currently), but in other cases that's not trivial, particularly inserting hexcodes into a text element (as it needs to hold state), and would require the list of possible actions that you could define shortcuts for (and would need to be handled via additional code) to be quite a lot bigger than it is currently.
In MU4 the shortcut mechanism kicks in *first*, and keyPress events don't even make it to the NotationPaintView (and hence to anything else) - hence the use of the ShortcutOverride event.  My initial implementation was to have this only do something if there was no enabled action for any shortcut defined for the key combination, and that can work, but it does mean we need to ensure that actions are properly disabled/enabled everywhere (and they're not currently). I'm looking at what needs to be done to ensure we DO disable actions properly now, but it's a fairly big job, and there doesn't seem to be a clean way of doing it - i.e. I can't easily define for each action "this is the handler, and this the `isEnabled()` function for it" which is how I've done it in the past with old-school MFC applications etc.
The other alternative, essentially mimicking the MU3 behaviour, is to allow all shortcut-override events through to the NotationActionController/NotationInteraction and have it determine depending on what's currently selected whether it should pass the keystroke through or not. As long as we can determine whether the action did anything (e.g. whether Element::edit( ) returns true or false) we should then be able to mark the event as accepted to prevent the shortcut key doing anything.

Eism:
duplicate of #8348

PR #8345, opened 2021-06-17
Chord Symbol Style Editor - Inspector Pop-up

Mystic-Slice:
The Chord Symbol Style Editor has two parts: Inspector pop-up and Styles Dialog.

The Inspector pop-up is implemented in this PR. There are few more things to do like matching the highlight in the grid view to the style currently selected and replacing text in the grid view items with previews of each style. 

The chord description file search is also implemented partially. The items listed in the pop-up are from the file search. This is handled by the class `ChordSymbolStyleManager`. It is going to be the source of chord style data for both the pop-up and Styles Dialog. It will also read and write style data. For `ChordSymbolStyleManager` to recognize the chord description file, the file must have the attribute `type` set to "chordStyle".


src/notation/inotation.h
```diff
@@ -72,6 +72,9 @@ class INotation
     // styles
     virtual INotationStylePtr style() const = 0;
 
+    // score
+    virtual Ms::Score* score() const = 0;
+
```
cbjeukendrup, line 77: This change seems unnecessary

Mystic-Slice: Removing this code causes my build to crash. Is it the same for you too?

cbjeukendrup: No, that sounds strange... I'll investigate tomorrow.

src/inspector/models/notation/chordsymbols/chordsymbolsettingsmodel.cpp
```diff
@@ -59,6 +60,20 @@ void ChordSymbolSettingsModel::resetProperties()
     m_durationType->resetToDefault();
 }
 
+void ChordSymbolSettingsModel::setChordStylesModel(ChordSymbolStylesModel* chordSymbolStylesModel)
+{
+    if (m_chordStylesModel == chordSymbolStylesModel) {
```
wizofaus, line 65: Seems a bit unnecessary unless you're triggering an event, also I'd generally write setters/getters as inline functions in the header file unless there's a particular reason not to.

src/inspector/models/notation/chordsymbols/chordsymbolstylesmodel.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include "chordsymbolstylesmodel.h"
+
+using namespace mu::inspector;
+using namespace mu::notation;
+ChordSymbolStylesModel::ChordSymbolStylesModel(QObject* parent)
```
wizofaus, line 26: would expect a blank line before this unless uncrustify doesn't like it!

src/inspector/models/notation/chordsymbols/chordsymbolstylesmodel.cpp
```diff
@@ -0,0 +1,80 @@
...
+    : QAbstractListModel(parent)
+{
+    styleManager = new ChordSymbolStyleManager();
+    m_styles = styleManager->getChordStyles();
+}
+
+int ChordSymbolStylesModel::rowCount(const QModelIndex&) const
+{
+    return m_styles.count();
```
wizofaus, line 35: what's the parameter for?

src/inspector/models/notation/chordsymbols/chordsymbolstylesmodel.cpp
```diff
@@ -0,0 +1,80 @@
...
+    default:
+        break;
+    }
+
+    return QVariant();
+}
+
+void ChordSymbolStylesModel::setChordStyle(QString styleName) const
+{
+    QString f = "chords_std.xml"; // Fall back
```
wizofaus, line 70: Would try to think of a better variable name here!

src/inspector/models/notation/chordsymbols/chordsymbolstylesmodel.cpp
cbjeukendrup: This method is overridden from QAbstractItemModel.

src/inspector/models/notation/chordsymbols/chordsymbolsettingsmodel.cpp
cbjeukendrup: @wizofaus It's generally our style to put nothing of function definitions in header files.

But I agree the setter is not necessary. Setters like this with such a check are necessary for use with QML, to avoid binding loops. But when a setter is only used in C++, the check is not necessary. In this case, the setter is only used inside the class itself, and only one time (if I'm not mistaken), so I think it should be removed and replaced by just `m_chordStylesModel = new ChordSymbolStylesModel(this);`.

wizofaus: Does it need to be alloc'ed on the heap at all? I.e. couldn't m_chordStylesModel just be a ChordSymbolStylesModel instance rather than a pointer?

cbjeukendrup: Good point. Also, currently a `delete` is missing, but with just an instance, that would not be necessary. I believe QML can only work with pointers to Object-derived items, so you would need to pass a pointer to the instance in the getter method.

wizofaus: Except the current object is passed in as "parent" to the QObject, I'm not sure if that's necessary or what it will mean - if Qt automatically deletes children when you delete the parent it would be an issue!

src/inspector/models/notation/chordsymbols/chordsymbolstylesmodel.cpp
Mystic-Slice: I had it temporarily. Forgot to change it.

src/inspector/models/notation/chordsymbols/chordsymbolsettingsmodel.cpp
Mystic-Slice: I added the setter method at first thinking I might use it from somewhere else in the code. But now it does seem unnecessary. I will remove it now and add it if I needed it again.
And regarding `ChordSymbolStylesModel` being a pointer or instance, a pointer is fine right? Is there any reason to change it?

wizofaus: You just need to check if it automatically gets deleted when the parent does, if not, you'll need to delete it yourself or use an instance.

wizofaus: From reading the Qt documentation it seems it will get automatically deleted as you're passing in the current object as the parent.

Mystic-Slice: > From reading the Qt documentation it seems it will get automatically deleted as you're passing in the current object as the parent.

Oh Ok.

src/notation/view/widgets/ChordSymbolStyleEditor.qml
```diff
@@ -0,0 +1,148 @@
...
+
+        FlatButton {
+            id: button
+
+            width: grid.cellWidth-5
+            height: grid.cellHeight-5
+
+            Text {
+                id: buttontext
+                text: styleName
```
cbjeukendrup, line 51: When you want to display text on a button, it is better to use the `text:` property of FlatButton. 
Also, it is preferred to use our own components (StyledTextLabel) instead of QML's default ones, because our own ones have the correct styling.

Mystic-Slice: Ok. I will change it finally(if it is still there). Because anyway I will be changing it to a canvas or smth.

cbjeukendrup:
Something strange is going on with styledef.cpp. I think it has to do with line endings (Windows uses different line endings from other OSs, and that sometimes causes trouble with Git). Could you try to fix that?

Mystic-Slice:
> Something strange is going on with styledef.cpp. I think it has to do with line endings (Windows uses different line endings from other OSs, and that sometimes causes trouble with Git). Could you try to fix that?

Should I just make sure that each line does not have trailing spaces?

cbjeukendrup:
No, it's something different. A line break is a hidden character, but there are multiple kinds of those characters: "carriage return" (`CR`), "line feed" (`LF`), and both in succession (`CRLF`). Different OSs use different characters. See https://stackoverflow.com/questions/1552749/difference-between-cr-lf-lf-and-cr-line-break-types.
Normally, Git makes sure that the correct line breaks are used when you checkout or commit, but for some reason that didn't happen here. Or, it _did_ happen but the file was wrong _before_ you edited it (in that case, the change is okay). 
The problem is that you can't very easily tell which type is used in a file.

Jojo-Schmitz:
And funny enough LFCR would have been the only real and correct way, mimiking the old mechanical typewriters, where the lever doing CR/LF does the Line Feed first and then the Carriage Return.

Jojo-Schmitz:
I don't see any issue with styledef.cpp though?

Mystic-Slice:
> I don't see any issue with styledef.cpp though?

Problem is actually in `styledef.h`. Probably a typo,

Jojo-Schmitz:
Ah, yes!

src/inspector/view/qml/MuseScore/Inspector/notation/chordsymbols/ChordSymbolSettings.qml
```diff
@@ -36,13 +36,22 @@ Column {
...
+        id: styleSettings
+        width: parent.width
+        chordStylesModel: root.model ? root.model.chordStylesModel : null
+    }
+
+    FlatButton {
+        width: parent.width
+        height: 30
+
+        text: qsTrc("inspector", "Edit styles NW")
```
Jojo-Schmitz, line 53: What does NW stand for?

Mystic-Slice: "Not Working" :) 
I have not implemented the functionality for that button yet. That's why.

Jojo-Schmitz: ;-)

src/notation/view/widgets/editstaff.cpp
```diff
@@ -156,6 +156,7 @@ void EditStaff::setStaff(Staff* s, const Fraction& tick)
     invisible->setChecked(m_staff->invisible(Fraction(0, 1)));
     isSmallCheckbox->setChecked(stt->isSmall());
     color->setColor(stt->color().toQColor());
+
```
Jojo-Schmitz, line 159: this change seems useless and may just lead to merge conflicts for no good reason

Mystic-Slice: Ahh yes. I missed that during rebase. I will fix it. Thnkx.

src/engraving/style/styledef.h
```diff
@@ -1,1500 +1,1521 @@
...
+        const char* valueType() const { return _defaultValue.typeName(); }
+        const char* name() const { return _name; }
+        const QVariant& defaultValue() const { return _defaultValue; }
+    };
+
+    static const StyleValue styleValues[int(Sid::STYLES)];
+};
+}
+
+#endif // MU_ENGRAVING_STYLEDEF_H
```
Jojo-Schmitz, line 1521: What happened here, maybe just a carriage return / linefeed change? If so: please revert.

src/notation/view/widgets/editstyle.ui
```diff
@@ -235,7 +240,7 @@
      </widget>
      <widget class="QStackedWidget" name="pageStack">
       <property name="currentIndex">
-       <number>0</number>
+       <number>35</number>
```
Jojo-Schmitz, line 243: This change is not needed. Probably harmlsee too though, IIRC it gets set to 1 programatically anyway

src/notation/view/widgets/editstyle.ui
```diff
@@ -222,6 +222,11 @@
         <string>Chord Symbols</string>
        </property>
       </item>
+      <item>
+       <property name="text">
+        <string>Chord Symbols New</string>
```
Jojo-Schmitz, line 227: Strange naming... I guess this is only temporary?

src/engraving/style/styledef.h
Mystic-Slice: I am sorry I do not understand. If you are asking if I have made any change to this file that pertains to the project, then yes. But the difference should be few lines only. I do not know what is happening here.

src/notation/view/widgets/editstyle.ui
Mystic-Slice: Yes. In the end, I will take out the old Chord Symbols page and rename this.

src/engraving/style/styledef.h
Jojo-Schmitz: Then please double check in those changes. I don't see a need to change the entire file, this will cause merge conflicts later

Jojo-Schmitz:
Rebasxe needed

Jojo-Schmitz:
Rebase is still (or again?) needed

src/notation/view/widgets/editstyle.ui
Mystic-Slice: Resolved

src/engraving/style/styledef.h
Mystic-Slice: Done

src/notation/view/widgets/editstaff.cpp
Mystic-Slice: Done

Mystic-Slice:
> Rebase is still (or again?) needed

Done

src/notation/view/widgets/editstyle.ui
Mystic-Slice: Done

bkunda:
This project is ongoing as it still has a number of functionalities that need to be implemented and/or resolved. Nevertheless, I'm happy that it's generally working well enough to be merged if it also passes all other relevant checks. @vpereverzev @shoogle

src/engraving/libmscore/chordlist.cpp
```diff
@@ -1432,6 +1434,799 @@ qreal ChordList::position(const QStringList& names, ChordTokenClass ctc) const
     }
 }
 
+//---------------------------------------------------------
+//   findModifierStartIndices
+//---------------------------------------------------------
+void ParsedChord::findModifierStackIndices()
+{
+    QStringList alterationsList = { "b", "bb", "#", "##", "natural" };
+    QStringList addOmitList = { "omit", "no", "add" };
```
RomanPudashkin, line 1443: Better to declare those lists as static const at the top of the file and use them wherever needed

src/engraving/libmscore/chordlist.cpp
```diff
@@ -1432,6 +1434,799 @@ qreal ChordList::position(const QStringList& names, ChordTokenClass ctc) const
     }
 }
 
+//---------------------------------------------------------
+//   findModifierStartIndices
+//---------------------------------------------------------
+void ParsedChord::findModifierStackIndices()
```
RomanPudashkin, line 1440: This method has a confusing signature. If something has the word "find" in its name, it should return some value

src/engraving/libmscore/chordlist.cpp
```diff
@@ -1432,6 +1434,799 @@ qreal ChordList::position(const QStringList& names, ChordTokenClass ctc) const
...
+{
+    QStringList alterationsList = { "b", "bb", "#", "##", "natural" };
+    QStringList addOmitList = { "omit", "no", "add" };
+    for (int index = 0; index < _tokenList.size(); index++) {
+        const ChordToken& tok = _tokenList.at(index);
+        if (tok.names.size() > 0) {
```
RomanPudashkin, line 1446: You can slightly simplify this block of code:

if (tock.names.isEmpty()) {
continue;
}

if (alterationsList.contains(tok.names.first())) {
..

src/engraving/libmscore/chordlist.cpp
```diff
@@ -1432,6 +1434,799 @@ qreal ChordList::position(const QStringList& names, ChordTokenClass ctc) const
...
+}
+
+//---------------------------------------------------------
+//   checkQualitySymbolsLetterCase
+//---------------------------------------------------------
+void ParsedChord::checkQualitySymbolsLetterCase(const ChordList* cl)
+{
+    // Changes the letter case of the first letter in quality symbols
+    // Leaves untouched if there is no other replacement
+    if (cl->autoCapitalization) {
```
RomanPudashkin, line 1482: You can simplify the implementation of this method:

if (!cl->autoCapitalization) {
return;
}

...

src/engraving/libmscore/chordlist.cpp
```diff
@@ -1432,6 +1434,799 @@ qreal ChordList::position(const QStringList& names, ChordTokenClass ctc) const
...
+        // It is important to have the lists synced
+        QStringList majorUpperCase = { "Maj", "Ma" };
+        QStringList majorLowerCase = { "maj", "ma" };
```
RomanPudashkin, line 1485: Better to declare those lists as static const

Jojo-Schmitz:
Another rebase is needed...

Mystic-Slice:
@RomanPudashkin Hi! Your reviews seem to have disappeared. Was that intentional?

RomanPudashkin:
> @RomanPudashkin Hi! Your reviews seem to have disappeared. Was that intentional?

Yes, it is. I decided to spend more time analyzing your PR because it has more complex problems than I expected

Jojo-Schmitz:
Rebase needed

bkunda:
Unfortunately we're going to have to close this one now since it is particularly old, and chord symbol styles were given a separate overhaul in the 4.6 release (meaning anything still relevant from here would have to be re-written from scratch).

PR #8347, opened 2021-06-17
[MU3 Backend] ENG-40 dynamics inference

iveshenry18:
Resolves: [ENG-40](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-40): Dynamics are sometimes rendered as plain text

A common XML exporter error is to mistake dynamics for direction text.
This commit adds two functions for correcting these mistakes.
1. In the case that a whole direction is a dynamic, this converts
that whole direction to the specified dynamic, including handling the
velocity change.
2. In the case that a direction contains a dynamic in addition to other
text (i.e. *mf cresc.*), this converts the dynamic portion to the proper
MuseScore dynamic symbol (with no velocity change).
In pursuit of this, the `dynList` of `Dyn` structs is moved from `dynamic.cpp`
to `dynamic.h`, so it can be used as the list of symbols to match.


oktophonie (approved):

Jojo-Schmitz:
rebase needed

iveshenry18:
Rebased!

vpereverzev (approved):

libmscore/dynamic.h
```diff
@@ -21,6 +21,59 @@ namespace Ms {
...
+      int velocity;      ///< associated midi velocity (0-127, -1 = none)
+      bool accent;       ///< if true add velocity to current chord velocity
+      const char* tag;   // name of dynamics, eg. "fff"
+      const char* text;  // utf8 text of dynamic
+      int changeInVelocity;
+      };
+
+// variant with ligatures, works for both emmentaler and bravura:
+
+static Dyn dynList[] = {
```
Jojo-Schmitz, line 39: This resulte in a ton )>120) of warnings reg. a defined but unused variable

Jojo-Schmitz: Variables should never ever get **defined** in header files...

Jojo-Schmitz: Hmm, making that `static const Dyn dynList[] = {` fixes the compiler warnings

iveshenry18: Hmm, fair point. I couldn't figure out a simple way to make this list available to the XML importer without declaring it in the header file, but I understand defining it there is a bit dirty. Perhaps a vector rather than array, to define it without having to know the size?

Jojo-Schmitz: Just adding the `const` appeases the compiler

iveshenry18: I'll add a PR with that then—thanks!

iveshenry18: #8478

Jojo-Schmitz: Maybe you can add fixes for the other warnings, from your other PRs?

Issue #7687, opened 2021-03-06
[MU4 Issue] Keyboard shortcuts not implemented


vshalkevich:
**Describe the bug**
Shortcuts not working:
1. Change line or slur length with arrow keys
2. Range selection with arrow keys + shift
3. N to toggle Note Input (can enter but not exit currently)
4. Home, End keys to jump to beginning or end of score
5. Copy (Cmd+C) and Paste (Cmd+V) shortcuts not working (works via right-click though)
6. No keyboard control for number spinboxes (IncrementalPropertyControl) (Checked on Linux Ubunty)
7. No keyboard control for sliders (Checked on Linux Ubunty)
8. No keyboard navigation in ValueList (Checked on Linux Ubunty)

**To Reproduce**
1. Try to use shortcuts

**Expected behavior**
Shortcuts work

**Desktop (please complete the following information):**
MacOS
Windows 10

iwoithe:
Not sure if it's connected or not, but when I tried compiling MuseScore (from [this](https://github.com/musescore/MuseScore/commit/5c1a1bb5812ff0be1064834bd844a6f20aa8d2d1) commit), it failed (OS: Windows 10, Compiler: MSYS Makefiles). Output of the error is below.
```
In file included from C:/iwoithe/Software/Source_Code/MuseScore/builds/Win-Qt5.15.0-mingw73_64-Make-RelWithDebInfo/src/framework/shortcuts/CMakeFiles/shortcuts.dir/Unity/unity_0_cxx.cxx:7:
C:/iwoithe/Software/Source_Code/MuseScore/src/framework/shortcuts/internal/shortcutsinstancemodel.cpp: In member function 'void mu::shortcuts::ShortcutsInstanceModel::load()':
C:/iwoithe/Software/Source_Code/MuseScore/src/framework/shortcuts/internal/shortcutsinstancemodel.cpp:59:39: error: no matching function for call to 'QShortcut::QShortcut(QString, QWidget*&, mu::shortcuts::ShortcutsInstanceModel::load()::<lambda()>, Qt::ShortcutContext)'
   59 |             }, Qt::ApplicationShortcut);
      |                                       ^
In file included from C:/Qt/Qt5.15.0/5.15.0/mingw73_64/include/QtWidgets/QShortcut:1,
                 from C:/iwoithe/Software/Source_Code/MuseScore/src/framework/shortcuts/internal/shortcutsinstancemodel.cpp:22,
[...]

PR #8348, opened 2021-06-18
Better fix for element-specific shortcut keys

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/7687 etc.

For now, this reverts to MU3's mechanism of letting the current selected element have "first bite" at any keypress, and only if it doesn't process it, allow any actions that the shortcut key maps to to fire.

- [  ] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -2230,6 +2230,7 @@ void NotationInteraction::deleteSelection()
         auto textBase = toTextBase(m_textEditData.element);
         if (!textBase->deleteSelectedText(m_textEditData)) {
             m_textEditData.key = Qt::Key_Backspace;
+            m_textEditData.modifiers = 0;
             textBase->edit(m_textEditData);
         }
     } else {
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -1901,7 +1901,7 @@ void NotationInteraction::moveText(MoveDirection d, bool quickly)
 
 bool NotationInteraction::isTextEditingStarted() const
 {
-    return m_textEditData.element != nullptr;
+    return m_textEditData.element && m_textEditData.element->isTextBase();
 }
 
 void NotationInteraction::startEditText(Element* element, const PointF& cursorPos)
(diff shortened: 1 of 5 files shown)


src/notation/view/notationpaintview.cpp
```diff
@@ -652,6 +652,7 @@ void NotationPaintView::mousePressEvent(QMouseEvent* event)
 {
     setFocus(true);
     emit activeFocusRequested();
+    forceActiveFocus();
```
wizofaus, line 655: Necessary or NotationPaintView won't receive ShortcutOverride events at all. Arguably it should forceActiveFocus the moment it's first displayed too?

src/notation/view/notationpaintview.cpp
```diff
@@ -688,30 +689,19 @@ void NotationPaintView::hoverMoveEvent(QHoverEvent* event)
...
-            }
-            if (event->modifiers() & Qt::KeyboardModifier::ControlModifier) {
-                sequence = "Ctrl+" + sequence;
-            }
-            const auto& shortcuts = shortcutsRegister()->shortcutsForSequence(sequence);
-            const std::shared_ptr<IUiActionsRegister> theRegister = actionsRegister();
-            if (std::find_if(shortcuts.begin(), shortcuts.end(), [theRegister](auto s) {
-                return theRegister->actionState(s.action).enabled;
-            }) != shortcuts.end()) {
-                return;
```
wizofaus, line 710: there's still an argument for keeping this logic if we decide it makes more sense to have the shortcut-mapped actions to kick in first, but that requires a lot more work to ensure they're only enabled when they make sense (e.g. "next-chord" shouldn't be enabled when a slur is selected etc. etc.)

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -473,11 +473,9 @@ void NotationViewInputController::hoverMoveEvent(QHoverEvent* event)
     }
 }
 
-void NotationViewInputController::keyReleaseEvent(QKeyEvent* event)
+void NotationViewInputController::keyPressEvent(QKeyEvent* event)
 {
-    if (m_view->notationInteraction()->isTextEditingStarted()) {
-        m_view->notationInteraction()->editText(event);
-    }
+    m_view->notationInteraction()->editText(event);
```
wizofaus, line 478: 'editText' is a bit of a misnomer now, probably should rename to "keyPressEvent".

src/notation/internal/notationinteraction.cpp
```diff
@@ -1937,18 +1937,26 @@ void NotationInteraction::startEditText(Element* element, const PointF& cursorPo
 
 void NotationInteraction::editText(QKeyEvent* event)
 {
-    IF_ASSERT_FAILED(m_textEditData.element) {
+    if (!m_textEditData.element && selection()->element()) {
```
wizofaus, line 1940: should probably rename to just m_editData now we're using it for other types of elements too

igorkorsukov (approved):

PR #8349, opened 2021-06-18
Prevent need for QtWidgets include directory

wizofaus:
Just to confirm what the full set of dependencies for engraving.lib were I created a console app that just instantiated an Ms::Score. Currently including score.h requires having the QtWidgets library include path, which shouldn't technically be necessary, so this change obviates the need for that.
I was able to successfully link the app just using:
engraving.lib
qzip.lib
Qt5Core.lib
Qt5Widgets.lib
Qt5Gui.lib
Qt5Svg.lib
zlibstat.lib

Proving that engraving.lib doesn't have any concrete dependencies on other MuseScore libraries.

### DIFF
--- a/src/engraving/draw/fontmetrics.h
+++ b/src/engraving/draw/fontmetrics.h
@@ -24,7 +24,7 @@
 
 #include "font.h"
 
-#include "modularity/ioc.h"
+#include "framework/global/modularity/ioc.h"
 #include "ifontprovider.h"
 #include "geometry.h"
 
--- a/src/engraving/libmscore/types.h
+++ b/src/engraving/libmscore/types.h
@@ -23,7 +23,9 @@
 #ifndef __TYPES_H__
 #define __TYPES_H__
 
-#include <QComboBox>
+class QComboBox;
+class QAction;
+class QMimeData;
 
 #include "config.h"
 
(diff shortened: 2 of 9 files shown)


src/engraving/libmscore/types.h
```diff
@@ -23,7 +23,9 @@
 #ifndef __TYPES_H__
 #define __TYPES_H__
 
-#include <QComboBox>
```
wizofaus, line 26: the key change!

src/engraving/draw/fontmetrics.h
```diff
@@ -24,7 +24,7 @@
 
 #include "font.h"
 
-#include "modularity/ioc.h"
+#include "framework/global/modularity/ioc.h"
```
wizofaus, line 27: also avoid need for src/framework/global include directory when including text.h etc.

igorkorsukov:
> Proving that engraving.lib doesn't have any concrete dependencies on other MuseScore libraries.

BTW We are currently also removing dependency on Qt5Gui and Qt5Widgets, step by step.

igorkorsukov (approved):

src/engraving/libmscore/types.h
```diff
@@ -23,7 +23,9 @@
...
+class QComboBox;
+class QAction;
+class QMimeData;
```
cbjeukendrup, line 28: Why were QAction and QMimeData added here? For me it also works without them.

wizofaus:
The tests failed to compile without this

PR #8353, opened 2021-06-18
[MU4] Fixed multiinstances locking

igorkorsukov:


src/multiinstances/resourcelockguard.h
```diff
@@ -0,0 +1,54 @@
...
+
+    ~ResourceLockGuard()
+    {
+        if (m_provider) {
+            m_provider->unlockResource(m_name);
+        }
+    }
+
+private:
+    std::shared_ptr<IMultiInstancesProvider> m_provider;
```
vpereverzev, line 49: would be nice to initialize it with nullptr

vpereverzev (approved):

src/multiinstances/resourcelockguard.h
igorkorsukov: added

PR #8355, opened 2021-06-18
Renamed global paths getters

igorkorsukov:
And added page about resources to wiki: https://github.com/musescore/MuseScore/wiki/Resources

### DIFF
--- a/src/extensions/internal/extensionsconfiguration.cpp
+++ b/src/extensions/internal/extensionsconfiguration.cpp
@@ -53,7 +53,7 @@ void ExtensionsConfiguration::init()
 {
     settings()->setDefaultValue(CHECK_FOR_UPDATE, Val(true));
 
-    settings()->setDefaultValue(USER_EXTENSIONS_PATH, Val(globalConfiguration()->sharePath().toStdString() + EXTENSIONS_DIR.toStdString()));
+    settings()->setDefaultValue(USER_EXTENSIONS_PATH, Val((globalConfiguration()->appDataPath() + EXTENSIONS_DIR).toStdString()));
     settings()->valueChanged(USER_EXTENSIONS_PATH).onReceive(nullptr, [this](const Val& val) {
         m_extensionsPathChanged.send(val.toString());
     });
@@ -269,5 +269,5 @@ io::path ExtensionsConfiguration::extensionTemplatesPath(const QString& extensio
 
 io::path ExtensionsConfiguration::extensionsDataPath() const
 {
-    return globalConfiguration()->dataPath() + EXTENSIONS_DIR;
+    return globalConfiguration()->userDataPath() + EXTENSIONS_DIR;
 }
--- a/src/instruments/internal/instrumentsconfiguration.cpp
+++ b/src/instruments/internal/instrumentsconfiguration.cpp
@@ -37,7 +37,7 @@ static const Settings::Key SECOND_SCORE_ORDER_LIST_KEY(module_name, "application
 void InstrumentsConfiguration::init()
 {
     settings()->setDefaultValue(FIRST_INSTRUMENT_LIST_KEY,
-                                Val(globalConfiguration()->sharePath().toStdString() + "instruments/instruments.xml"));
+                                Val(globalConfiguration()->appDataPath().toStdString() + "instruments/instruments.xml"));
     settings()->valueChanged(FIRST_INSTRUMENT_LIST_KEY).onReceive(nullptr, [this](const Val&) {
         m_instrumentListPathsChanged.notify();
     });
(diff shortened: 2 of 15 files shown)


src/extensions/internal/extensionsconfiguration.cpp
```diff
@@ -53,7 +53,7 @@ void ExtensionsConfiguration::init()
 {
     settings()->setDefaultValue(CHECK_FOR_UPDATE, Val(true));
 
-    settings()->setDefaultValue(USER_EXTENSIONS_PATH, Val(globalConfiguration()->sharePath().toStdString() + EXTENSIONS_DIR.toStdString()));
+    settings()->setDefaultValue(USER_EXTENSIONS_PATH, Val((globalConfiguration()->appDataPath() + EXTENSIONS_DIR).toStdString()));
```
igorkorsukov, line 56: @Eism It looks like there is a mistake. The key is called `user...` and the application data folder is used (not the user data folder)

cbjeukendrup:
For MuseScore 3, on macOS it was like this: 
* `mscore.app/Contents/Resources/`
    * (app data)
    * NOTE: is Read-only, on macOS!!
* `~/Documents/MuseScore[Version]/`
    * Extensions/
    * Images/
    * Plugins/
    * Scores/
    * SoundFonts/
    * Styles/
    * Templates/
* `~/Library/Application Support/MuseScore/MuseScore[Version]/`
    * locale/ = downloaded translations
    * plugins/ = ??
    * QtWebEngine/ = ??
    * tours/ 
        * completedTours.list
    * workspaces/
    * session
    * shortcuts.xml
* `~/Library/Preferences/org.musescore.MuseScore[Version].plist`

I think it's okay to have the preferences file in that place (that's very common for macOS apps). 

About the other user data paths:
I think there are two things. 
A path like `~/Documents/MuseScore[Version]/` for "visible" user data, like templates, style files, SoundFonts; things the user installs or edits.
And a path like `~/Library/Application Support/MuseScore/MuseScore[Version]/` for "hidden" data, e.g. data that will/should not be displayed in its raw form to the user, like which tours are completed, session files, and auto-save files.

In MuseScore 3, shortcuts.xml is also in the "hidden" user data folder. I think this is good, because it's bad if an uneducated user edits this file and corrupts it, and because the values inside it can also be edited via the GUI. 

When the user wants to share this file, it can be "exported" in the Preferences dialog, so that the user can save a copy of the file somewhere they choose, and then also imported.

src/extensions/internal/extensionsconfiguration.cpp
cbjeukendrup: I think it should indeed be "userDataPath", and also a `+ "/" ` is missing.

igorkorsukov:
> For MuseScore 3, on macOS it was like this:
> 
> * `mscore.app/Contents/Resources/`
>   
>   * (app data)
>   * NOTE: is Read-only, on macOS!!
> * `~/Documents/MuseScore[Version]/`
>   
>   * Extensions/
>   * Images/
>   * Plugins/
>   * Scores/
>   * SoundFonts/
>   * Styles/
>   * Templates/
> * `~/Library/Application Support/MuseScore/MuseScore[Version]/`
>   
>   * locale/ = downloaded translations
>   * plugins/ = ??
>   * QtWebEngine/ = ??
>   * tours/
>     
>     * completedTours.list
>   * workspaces/
>   * session
>   * shortcuts.xml
> * `~/Library/Preferences/org.musescore.MuseScore[Version].plist`
> 
> I think it's okay to have the preferences file in that place (that's very common for macOS apps).
> 
> About the other user data paths:
> I think there are two things.
> A path like `~/Documents/MuseScore[Version]/` for "visible" user data, like templates, style files, SoundFonts; things the user installs or edits.
> And a path like `~/Library/Application Support/MuseScore/MuseScore[Version]/` for "hidden" data, e.g. data that will/should not be displayed in its raw form to the user, like which tours are completed, session files, and auto-save files.
> 
> In MuseScore 3, shortcuts.xml is also in the "hidden" user data folder. I think this is good, because it's bad if an uneducated user edits this file and corrupts it, and because the values inside it can also be edited via the GUI.
> 
> When the user wants to share this file, it can be "exported" in the Preferences dialog, so that the user can save a copy of the file somewhere they choose, and then also imported.

Thanks for the info, I started trying to figure out which path to what.

src/instruments/internal/instrumentsconfiguration.cpp
```diff
@@ -48,7 +48,7 @@ void InstrumentsConfiguration::init()
     });
 
     settings()->setDefaultValue(FIRST_SCORE_ORDER_LIST_KEY,
-                                Val(globalConfiguration()->sharePath().toStdString() + "instruments/orders.xml"));
+                                Val(globalConfiguration()->appDataPath().toStdString() + "instruments/orders.xml"));
```
igorkorsukov, line 51: Looks like there is missing `/`

vpereverzev (approved):

PR #8356, opened 2021-06-18
Fixed theme double deleting

igorkorsukov:
And added comment to code

image


RomanPudashkin (approved):

igorkorsukov:
A good solution would be:
* Make `UiEngine` not singleton (`IoC` is enough)
* Make it so that `UiEngine` is deleted before `QApplication` (we need to call `Ioc` clear, by the way now it's interesting it's calling ...)
* And then we can keep `delete m_theme` in the `UiEngine` destructor for clarity and semantic correctness

Sounds like tech debt :)

PR #8359, opened 2021-06-19
[MU4] fix #7674 - playback doesn't update after changing notation

### LINKED ISSUE #7674: [MU4 Issue] Playback cursor doesn't work on rests
**Describe the bug**
You can not see dynamically blue play bar on empty spaces. It moves dynamically only on notes

**To Reproduce**
1. Create an empty score
2. Add notes in the middle of the score
3. Start playing score from the beggining

**Expected behavior**
You should see that blue playing bar moving throw the score dynamically

Windows 10

wizofaus:
And cursor doesn't advance through tacet passages.

Resolves: https://github.com/musescore/MuseScore/issues/7674 and possibly others

The sequencer was not getting correctly stopped on changing notes in the score, meaning after doing so, it just keep playing from the same cached "playback" data. 
Further the way stepped time works means that no events were triggering the cursor to advance when no notes are playing.

### DIFF
--- a/src/notation/internal/notationplayback.cpp
+++ b/src/notation/internal/notationplayback.cpp
@@ -258,7 +258,7 @@ void NotationPlayback::makeChunk(midi::Chunk& chunk, tick_t fromTick) const
 
         midi::EventType etype = static_cast<midi::EventType>(ev.type());
         static const std::set<EventType> SKIP_EVENTS
-            = { EventType::ME_INVALID, EventType::ME_EOT, EventType::ME_TICK1, EventType::ME_TICK2 };
+            = { EventType::ME_INVALID, EventType::ME_EOT };
         if (SKIP_EVENTS.find(etype) != SKIP_EVENTS.end()) {
             continue;
         }
@@ -716,7 +716,7 @@ QRect NotationPlayback::loopBoundaryRectByTick(LoopBoundaryType boundaryType, in
 
     for (int i = 0; i < score()->nstaves(); ++i) {
         Ms::SysStaff* ss = system->staff(i);
-        if (!ss->show() || !score()->staff(i)->show()) {
+        if (!ss || !ss->show() || !score()->staff(i)->show()) {
             continue;
         }
         y2 = ss->y() + ss->bbox().height();
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -61,6 +61,14 @@ void PlaybackController::init()
     onNotationChanged();
     globalContext()->currentNotationChanged().onNotify(this, [this]() {
         onNotationChanged();
+        auto notation = globalContext()->currentNotation();
+        if (notation) {
+            notation->notationChanged().onNotify(this, [this, notation]() {
+                if (notation == m_notation) {
+                    onNotationChanged();
+                }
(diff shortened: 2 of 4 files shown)


src/playback/internal/playbackcontroller.cpp
```diff
@@ -72,6 +80,8 @@ void PlaybackController::init()
     sequencer()->positionChanged().onNotify(this, [this]() {
         if (configuration()->cursorType() == PlaybackCursorType::SMOOTH) {
             m_tickPlayed.send(currentTick());
+        } else if (currentTick() % 480 < 24) {
+            m_tickPlayed.send(currentTick());
```
wizofaus, line 84: Yes, this is a kludgy fix, but I don't know what else to do, because to do properly requires knowing the current time signature (to find out the beat length), and that's not available at this point.
The < 24 is required because the ticks received don't reliably fall on 1/4 note boundaries (multiples of 480), but in all the tests I did they were within 24 ticks after such boundaries.

wizofaus: I think I have a better idea now...the metronome track should fix this issue, just need to emit silent ticks...but the metronome track doesn't seem to be working either!

src/notation/internal/notationplayback.cpp
```diff
@@ -716,7 +716,7 @@ QRect NotationPlayback::loopBoundaryRectByTick(LoopBoundaryType boundaryType, in
 
     for (int i = 0; i < score()->nstaves(); ++i) {
         Ms::SysStaff* ss = system->staff(i);
-        if (!ss->show() || !score()->staff(i)->show()) {
+        if (!ss || !ss->show() || !score()->staff(i)->show()) {
```
wizofaus, line 719: this was to prevent a crash but is almost certainly hiding a bigger problem, will report separately

src/notation/internal/notationplayback.cpp
```diff
@@ -258,7 +258,7 @@ void NotationPlayback::makeChunk(midi::Chunk& chunk, tick_t fromTick) const
 
         midi::EventType etype = static_cast<midi::EventType>(ev.type());
         static const std::set<EventType> SKIP_EVENTS
-            = { EventType::ME_INVALID, EventType::ME_EOT, EventType::ME_TICK1, EventType::ME_TICK2 };
+            = { EventType::ME_INVALID, EventType::ME_EOT };
```
wizofaus, line 261: Even though ME_TICK1/ME_TICK2 aren't standard MIDI events, we still need to know about them - currently there's no playback for them supported at all, but at least it will trigger tickPlayed events which triggers the playback cursor movement etc.

vpereverzev:
I'm working on the rework of the entire playback engine at the moment, so I strongly discourage to spend efforts on such issues.
But anyway, thanks for your work, I'll put this PR into archive, perhaps the first part of your changes might be useful for the future

wizofaus:
Hmm, then I can suggest it would be good to update any open unassigned issues related to playback and assign them to yourself,  as I'm basically trying to fix as many as possible and I don't have any interest into who's working on what. I did spend quite a bit of time in this and I don't mind if it helps my understanding of the code but if the code is being largely rewritten then I'm not sure I learned anything useful!

PR #8412, opened 2021-06-22
[MU3 Backend] ENG-55: Infer tempo text

iveshenry18:
Resolves: [ENG-55](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-55): Tempi incorrectly rendered as Stave Text

This PR includes commits for inferring tempo text that is incorrectly exported as staff text.

Sometimes tempo text is incorrectly exported as staff text, often in the
form "q = 60". This commit adds regex-based inferencing for this case,
converting the left side of the equals sign to the proper symbol, and
calculating the resulting tempo from the resulting symbol and the value
of the right side.

Sometimes tempo text with additional instructions is exported as two
separate directions. Previously, this was imported as a Tempo Text
element and a separate Staff Text element. This commit adds a function
to infer these cases and merge the text from the Staff Text element
into the Tempo Text.

This now adds inferring based on common tempo text words such as "rit"
and "accel". In the case that one of these is inferred, the underlying
tempo remains the same.


oktophonie (approved):

iveshenry18:
Rebased!

vpereverzev:
one more time, pls :)

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -2442,6 +2443,20 @@ static void preventNegativeTick(const Fraction& tick, Fraction& offset, MxmlLogg
...
 
+//---------------------------------------------------------
+//   isTempoOrphanCandidate
+//---------------------------------------------------------
+
+bool MusicXMLDelayedDirectionElement::isTempoOrphanCandidate() const
+      {
+      return _element->isStaffText()
+            && _placement == "above"
+            && toStaffText(_element)->xmlText().contains("<b>");
```
wizofaus, line 2454: seems a fragile way of determining whether text might be intended as a tempo marking - esp. the `contains("<b>")` bit, even if it were true that using bold styling was the right measure to use, wouldn't it be better to explicitly query the formatting information to check for bold? And will this work if the bold comes from the Style?

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -2455,6 +2470,38 @@ QString MusicXMLParserDirection::placement() const
...
+                        ddi2Next = std::next(ddi2);
+                        // Combine with tempo text if present
+                        if (ddi1 != ddi2
+                            && (*ddi2)->tick() == (*ddi1)->tick()
+                            && (*ddi2)->element()->isTempoText()) {
+                              TempoText* tt = toTempoText((*ddi2)->element());
+                              StaffText* st = toStaffText((*ddi1)->element());
+                              QString sep = tt->plainText().endsWith(' ') || st->plainText().startsWith(' ') ? "" : " ";
+                              tt->setXmlText(tt->xmlText() + sep + st->xmlText());
+                              delete st;
```
wizofaus, line 2496: is just deleting the StaffText like that safe?  I'd expect to have to call "removeElement" on its parent...

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -3030,6 +3081,65 @@ MusicXMLDelayedDirectionElement* MusicXMLInferredFingering::toDelayedDirection()
...
+//   attemptTempoTextCoercion
+//---------------------------------------------------------
+/**
+ Infers if a direction is likely tempo text, possibly changing
+ the _wordsText to the appropriate note symbol and inferring the _tpoSound.
+ */
+
+bool MusicXMLParserDirection::attemptTempoTextCoercion(const Fraction& tick)
+      {
+      QList<QString> tempoWords{"rit", "rall", "accel", "tempo", "allegr", "poco", "molto", "più", "meno", "mosso", "rubato"};
```
wizofaus, line 3128: There's at least few terms in there that might not be tempo at all (e.g. poco _**f**_), but it seems a kinda arbitrary list, what percentage of cases do you expect it to cover?  Why "allegr" but not, say, "moderato", "andante" or "lento"?
And obviously a lot of music these days doesn't use Italian directions at all...but I guess I'm not fully clued in with what you're trying to solve here.

importexport/musicxml/importmxmlpass2.cpp
iveshenry18: Indeed this is a fairly wide net for candidates; however, it only actually does anything if there is an existing tempo marking at the exact same staff and tick. This does risk finding a false positive if there is a score that happens to have (inexplicably) bolded direction text on the same beat as a tempo marking (and above the staff). This case is likely quite rare (bolded text above the staff is almost always a tempo direction), but nonetheless we may put this behind an importing option when porting this to master. 

As far as the `contains("<b>")`, it's true that it's not the prettiest, but MusicXML doesn't have styles as such. All the formatting is part of the <text> element, which gets translated to MuseScore's text XML (i.e. a  `font-weight="bold"` gets translated to `<b>` tags) and applied to the element with `setXmlText()`. This is because sometimes a direction will include multiple different text elements, with some portions bold and some not. For these reasons, directly checking the style of the element is always false, so checking the XML for `<b>` is most closely correlated with the MusicXML import. I'll change it from `contains("<b>")` to a simple regex to make it a bit more robust.

importexport/musicxml/importmxmlpass2.cpp
iveshenry18: Safe! At this point in the code the elements have not yet been added to the score, so they don't yet have parents.

importexport/musicxml/importmxmlpass2.cpp
iveshenry18: Yes, at this point it's certainly non-exhaustive list. We currently have some XML scores that we are testing with, so these are pulled from there. The list can certainly grow over time.

As far as the words that may be part of things other than tempo markings (poco, molto), note that this list is only used _after_ it is checked that the text is above the staff and bold (I'll also use the slightly-more-robust regex version of bold-checking I mention above). It would be very rare (and likely incorrect) to have a fully-bold direction above the staff with the word poco that _isn't_ tempo text.

That said, there was a missing bracket in the code that uses this list. Glad you said something!

importexport/musicxml/importmxmlpass2.cpp
wizofaus: Yes, I understand how xmlTest() works currently, and it's likely that in its current form checking for `<b>` is reliable enough, but in future it mightn't work if there's a switch to using, say,  `<strong>` or `<b style="font-size:large">` etc. etc
Did you consider checking it at the point of import from the MusicXML format (which is defined by a standard that's unlikely to change any time soon) ? Understand that would be a bigger change but just seems more robust.

iveshenry18: From a future-proofing standpoint that would make sense. I'm a bit wary that the complexity added in checking it at the point of import will itself be rigid/error-prone (partly because the existing direction import classes—`MusicXMLParserDirection` and `MusicXMLDelayedDirectionElement`—could use a good refactor). Also, the regression tests I've added would in theory catch the error if such a refactor were to occur, and certainly one would have to be checking the regression tests actively if they were to make a refactor so big.

...but I'll give it a try!

iveshenry18: Done! A little cumbersome, since `isBold` has to be passed down from `MusicXMLParserDirection` to `MusicXMLDelayedDirectionElement`, but I agree that this is more future-proof.

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -2453,6 +2454,20 @@ static void preventNegativeTick(const Fraction& tick, Fraction& offset, MxmlLogg
...
 
+//---------------------------------------------------------
+//   isTempoOrphanCandidate
+//---------------------------------------------------------
+
+bool MusicXMLDelayedDirectionElement::isTempoOrphanCandidate() const
+      {
+      return _element->isStaffText()
+            && _placement == "above"
+            && isBold();
```
wizofaus, line 2465: looks much cleaner! Perhaps _placement shouldn't be a string, but I gather that's as per the MusicXML standard.

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -2659,15 +2709,15 @@ void MusicXMLParserDirection::direction(const QString& partId,
 
             // Add element to score later, after collecting all the others and sorting by default-y
             // This allows default-y to be at least respected by the order of elements
-            MusicXMLDelayedDirectionElement* delayedDirection = new MusicXMLDelayedDirectionElement(hasTotalY() ? totalY() : 100, dyn, track, dynamicsPlacement, measure, tick + _offset);
+            MusicXMLDelayedDirectionElement* delayedDirection = new MusicXMLDelayedDirectionElement(hasTotalY() ? totalY() : 100, dyn, track, dynamicsPlacement, measure, tick + _offset, _isBold);
             delayedDirections.push_back(delayedDirection);
             }
 
       // handle the elems
       foreach( auto elem, _elems) {
```
Jojo-Schmitz, line 2717: while near by: change to `for (auto elem: _elems) {`?

iveshenry18: Might as well!

importexport/musicxml/importmxmlpass2.cpp
wizofaus: I think you should at least support "Egyptically": https://pbs.twimg.com/media/D1H2tmKU8AAtK2w?format=jpg&name=small

vpereverzev (approved):

oktophonie:
[ENG-55: Infer tempo text](https://trello.com/c/LsBHM5JA/24-eng-55-infer-tempo-text)

PR #8419, opened 2021-06-23
[MU4] Fixed updating instruments panel

Eism:
Need reset list model only when the list of parts has been changed


RomanPudashkin (approved):

vshalkevich:
Buttons aligned but file path looks cutted (if long file path)
2021-06-24 19_38_41-MuseScore 4

vshalkevich:
All is OK

PR #8422, opened 2021-06-23
Rework workspaces

igorkorsukov:
Before: there was one xml file in which all the data was stored - this is very inconvenient for use.
Now: each data type will be stored in a separate file, and the workspace is just a container for those files.

Also, the use of workspaces has also been redesigned.

In some places, the logic is sketchy (perhaps something was not done) or very complex. So, there is still work to be done.


src/appshell/internal/appshellconfiguration.h
```diff
@@ -91,7 +95,6 @@ class AppShellConfiguration : public IAppShellConfiguration
 
     std::string currentLanguageCode() const;
 
-    async::Channel<bool> m_notationStatusBarVisibleChanged;
     async::Channel<bool> m_notationNavigatorVisibleChanged;
```
vpereverzev, line 98: looks like it could be delete either

src/workspace/workspacetypes.h
```diff
@@ -22,56 +22,28 @@
...
-    Unknown,
-    UiArrangement,
-    Settings,
+static const std::string DEFAULT_WORKSPACE_NAME("Default");
+
+enum class DataKey {
+    Undefined = 0,
+    UiSettings,
+    UiStates,
+    UiToolConfigs,
```
Eism, line 32: Better use toolbar instead of tool

igorkorsukov: I am not sure, maybe `tool` is not only `bar`? :)

igorkorsukov: for example, some context menu?

src/appshell/internal/appshellconfiguration.h
igorkorsukov: Thanks, removed

src/framework/ui/internal/uiarrangement.cpp
```diff
@@ -0,0 +1,168 @@
...
+mu::async::Notification UiArrangement::stateChanged(const QString& key) const
+{
+    return m_statesNotifications[key];
+}
+
+ToolConfig UiArrangement::toolConfig(const QString& toolName) const
+{
+    TRACEFUNC;
+
+    //! NOTE Maybe we need to cache?
```
vpereverzev, line 121: makes sense to me, but not in this PR I guess

src/workspace/workspacetypes.h
Eism: or add a new type like 'UIContextMenu'

igorkorsukov: I think not, I see no reason to split.
The meaning of the separation is something like this:
* Palettes - can be customized by the teacher for the students, in accordance with the topic that they are currently going through.
* Tools - or the same as the Palettes, the teacher sets up for the students, or for yourself you set up only those controls that you use (and it doesn't matter where they are located, the point is that you set up the controllers)
* The geometry and arrangement of the panels  - are probably customizable for yourself.
* UI style - is also probably customizable for yourself as convenient.

src/framework/ui/internal/uiarrangement.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    ToolConfig config;
+
+    config.items.reserve(itemsArr.size());
+    for (const QJsonValue& v : itemsArr) {
+        QJsonObject itemObj = v.toObject();
+
+        ToolConfig::Item item;
+        item.action = itemObj.value("action").toString().toStdString();
+        item.show = itemObj.value("show").toInt(1);
```
Eism, line 134: toInt(1) - Is it the setting of the default value? Then it is wrong

src/workspace/workspacetypes.h
igorkorsukov: Based on this logic, I think that it makes sense to join palettеs and controls (consider that pallets are controls) to one option.
And join the arrangement of the panels and the UI style - consider that this is a UI style

src/framework/ui/internal/uiarrangement.cpp
igorkorsukov: Yes, this is the default.
Why are you thinking wrong?
This should not be the case, the value should always be set, but if there was some kind of software error, then I prefer that all the controls are shown, instead of all are hidden

Eism: Sorry, my mistake. I looked at the description of the QVariant::toInt method instead of QJsonObject::toInt

vpereverzev (approved):

src/framework/ui/internal/uiarrangement.cpp
igorkorsukov: I added `TRACEFUNC`, we need watch for it values

src/workspace/workspacetypes.h
Tantacrul: I'm not 100% certain I understand. Perhaps let's jump on a Discord chat.

Eism (approved):

PR #8423, opened 2021-06-23
Fix #8364 - remove dangerous and unneeded museScoreCore global variable

### LINKED ISSUE #8364: [MU4 Issue] Crash in Score::endCmd due to unsafe use of global variable
**Describe the bug**
Slightly random, haven't been able to reproduce 100% consistently, but certain actions (in this case trying to change the duration of a note) can cause a crash in Score::endCmd, specifically at the line `MuseScoreCore::mscoreCore->endCmd(isCmdFromInspector, rollback)`.
This seems like quite dangerous and unnecessary use of a global variable. At the point it crashed `MuseScoreCore::mscoreCore` appeared to refer to something already deleted.

**To Reproduce**
1. Create a new score
2. Open the new score dialog (Ctrl+N)
3. Cancel/close dialog
4. Add a note
5. MuseScore sometimes crashes (sometimes)

**Expected behavior**
Preferably, applicaton should not crash!

**Screenshots**

**Additional context**

read access violation
>	MuseScore4.exe!Ms::Score::endCmd(const bool isCmdFromInspector, bool rollback) Line 303	C++
 	MuseScore4.exe!mu::notation::NotationUndoStack::commitChanges() Line 110	C++
 	MuseScore4.exe!mu::notation::NotationNoteInput::padNote(const Ms::Pad & pad) Line 202	C++
 	MuseScore4.exe!mu::notation::NotationActionController::padNote(const Ms::Pad & pad) Line 497	C++

Virtual table pointer for MuseScoreCore::mscoreCore was invalid.

Basically ~MuseScoreCore can occasionally be called for the object currently referred to by MuseScoreCore::mscoreCore but I haven't managed to work out why/when it happens.

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8364

Dangerous use of museScoreCore global variable was causing random crashes

### DIFF
--- a/src/plugins/api/qmlpluginapi.cpp
+++ b/src/plugins/api/qmlpluginapi.cpp
@@ -374,5 +374,16 @@ void PluginAPI::registerQmlTypes()
 
     qmlTypesRegistered = true;
 }
+
+MuseScoreCore* PluginAPI::msc() const
+{
+    static MuseScoreCore mscStatic;
+    if (this->context() && this->context()->currentNotation()) {
+        mscStatic.setCurrentScore(this->context()->currentNotation()->elements()->msScore());
+    } else {
+        mscStatic.setCurrentScore(0);
+    }
+    return &mscStatic;
+}
 }
 }
--- a/src/plugins/api/qmlpluginapi.h
+++ b/src/plugins/api/qmlpluginapi.h
@@ -33,11 +33,13 @@
 #include "libmscore/score.h"
 #include "libmscore/spanner.h"
 #include "framework/shortcuts/ishortcutscontroller.h"
+#include "context/iglobalcontext.h"
 #include "modularity/ioc.h"
 
 namespace Ms {
 class Element;
 class MScore;
+class MuseScoreCore;
 
 /**
  * \namespace Ms::PluginAPI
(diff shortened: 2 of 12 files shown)


src/plugins/api/qmlpluginapi.cpp
```diff
@@ -374,5 +374,16 @@ void PluginAPI::registerQmlTypes()
 
     qmlTypesRegistered = true;
 }
+
+MuseScoreCore* PluginAPI::msc() const
+{
+    static MuseScoreCore mscStatic;
+    if (this->context() && this->context()->currentNotation()) {
+        mscStatic.setCurrentScore(this->context()->currentNotation()->elements()->msScore());
```
wizofaus, line 382: This is still not ideal, as we're storing the pointer to the current Score potentially beyond its lifetime. Probably should change to shared_ptr, or better still get rid of MuseScoreCore completely if possible and just use Score?  Did try making MuseScoreCore injectable but that didn't seem to work.
But until the plugin mechanism is implemented in MU4 it doesn't really matter, none of this is used for anything.

src/plugins/api/qmlpluginapi.h
```diff
@@ -74,6 +76,7 @@ class PluginAPI : public Ms::QmlPlugin
     Q_OBJECT
 
     INJECT(plugins, mu::shortcuts::IShortcutsController, shortcuts)
+    INJECT(userscores, mu::context::IGlobalContext, context)
```
igorkorsukov, line 79: should be change `userscore` to `plugins`

igorkorsukov (approved):

PR #8424, opened 2021-06-23
[MU3] Fix #322347 - ensure hidden rests in V2-4 don't move back to default position

wizofaus:
Resolves: https://musescore.org/en/node/322347

Hiding V2-4 rests currently causes them to move back to the default/vertical center position on a stave, causing them to interfere with visible V1 elements. In general hiding items shouldn't cause the hidden item to move.

NB this is only a partial fix - the other improvement would be to ensure that hidden elements are drawn first (in a layer underneath everything else) so that visible ones can't be obscured by them.

### DIFF
--- a/src/engraving/libmscore/rest.cpp
+++ b/src/engraving/libmscore/rest.cpp
@@ -444,7 +444,7 @@ int Rest::getDotline(TDuration::DurationType durationType)
 int Rest::computeLineOffset(int lines)
 {
     Segment* s = segment();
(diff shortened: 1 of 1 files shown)


wizofaus:
Fixing the z-order so that invisible elements are painted first, under visible elements, is easy with a simple change to `Element::z()` to always return 0 for invisible elements, or a relatively small change to `paintElements( )` so that `!visible()` overrides `z()` when sorting.
Not sure which is best.

PR #8425, opened 2021-06-23
Partial fix #298623: export Symbol elements to MusicXML

dmitrio95:
Resolves: https://musescore.org/en/node/298623 (partially)

Exports `Symbol` elements to MusicXML as `<other-direction>` or `<other-notation>`
tags, depending on a type of the symbol's parent element.

A test case can be added if MusicXML export tests get re-enabled, see my comment in https://github.com/musescore/MuseScore/pull/8273#issuecomment-867199710.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -3160,6 +3160,16 @@ static QString instrId(int partNr, int instrNr)
 }
 
 //---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


vpereverzev (approved):

Jojo-Schmitz:
Just in case this ever gets backported to 3.x or 3.6.2_backend (@iveshenry18?): it needs #8273 getting backported first

iveshenry18:
I'll keep it in mind! 3.6.2_backend is really prioritizing the importing.

PR #8427, opened 2021-06-24
[MU4] Restored ability to close docks

RomanPudashkin:
The ability was lost after switching to the new dock system


Eism:
You also need to remove the configuration of the status bar through a separate property in the configurator IAppShellConfiguration::isNotationStatusBarVisible
This was a solution to the limitation of the old docking system, now we do not need it

vshalkevich:
Works fine for me

igorkorsukov (approved):

PR #8282, opened 2021-06-08
Musicxml duration 3.x

lvinken:
Resolves:

- https://musescore.org/en/node/321751
- https://musescore.org/en/node/321753

321751:
Currently, the importer uses MusicXML note duration instead of note length as calculated from the MusicXML note
type, dots, time-modification etcetera to determine how long a note takes and where to place the next note. When
the MusicXML note duration does not match the calculated value, various types of corruption occur.
    
When using the calculated duration, none of these problems occur. This change does not affect correctly encoded
MusicXML files.

321753:
See discussion in https://musescore.org/en/node/313339. This change enables correct import of Overture files containing
changed note durations (while leaving the note type unchanged) by using the duration to change the note's off time.
Strictly speaking Overture does not adhere to the MusicMXL spec, but this change does not seem to introduce any regression. If necessary it can be made optional in future (either as a user setting or by making it Overture specific).

Providing a 3.x PR to benefit from the better auto test coverage and in case another 3.x release will be made.
Will also syncing to master shortly.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8429, opened 2021-06-24
Musicxml duration master

lvinken:
Resolves:

- https://musescore.org/en/node/321751
- https://musescore.org/en/node/321753

321751:
Currently, the importer uses MusicXML note duration instead of note length as calculated from the MusicXML note
type, dots, time-modification etcetera to determine how long a note takes and where to place the next note. When
the MusicXML note duration does not match the calculated value, various types of corruption occur.

When using the calculated duration, none of these problems occur. This change does not affect correctly encoded
MusicXML files.

321753:
See discussion in https://musescore.org/en/node/313339. This change enables correct import of Overture files containing
changed note durations (while leaving the note type unchanged) by using the duration to change the note's off time.
Strictly speaking Overture does not adhere to the MusicMXL spec, but this change does not seem to introduce any regression. If necessary it can be made optional in future (either as a user setting or by making it Overture specific).

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -4547,7 +4547,7 @@ Note* MusicXMLParserPass2::note(const QString& partId,
     // check for timing error(s) and set dura
     // keep in this order as checkTiming() might change dura
     auto errorStr = mnd.checkTiming(type, rest, grace);
-    dura = mnd.dura();
+    dura = mnd.duration();
     if (errorStr != "") {
         _logger->logError(errorStr, &_e);
     }
@@ -4583,7 +4583,7 @@ Note* MusicXMLParserPass2::note(const QString& partId,
     if (!chord && !grace) {
         auto& tuplet = tuplets[voice];
         auto& tupletState = tupletStates[voice];
-        tupletAction = tupletState.determineTupletAction(mnd.dura(), timeMod, notations.tupletDesc().type,
+        tupletAction = tupletState.determineTupletAction(mnd.duration(), timeMod, notations.tupletDesc().type,
                                                          mnd.normalType(), missingPrev, missingCurr);
         if (tupletAction & MxmlTupletFlag::STOP_PREVIOUS) {
             // tuplet start while already in tuplet
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -4704,6 +4704,25 @@ Note* MusicXMLParserPass2::note(const QString& partId,
         setNoteHead(note, noteheadColor, noteheadParentheses, noteheadFilled);
         note->setVisible(hasHead && printObject);     // TODO also set the stem to invisible
 
+        if (mnd.calculatedDuration().isValid()
+            && mnd.specifiedDuration().isValid()
+            && mnd.calculatedDuration().isNotZero()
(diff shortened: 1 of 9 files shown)


lvinken:
Noticed the build failures, will fix, cannot commit to a delivery date yet.

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -4754,6 +4754,25 @@ Note* MusicXMLParserPass2::note(const QString& partId,
...
+            NoteEventList nel;
+            NoteEvent ne;
+            ne.setLen(noteLen);
+            nel.append(ne);
+            note->setPlayEvents(nel);
+            if (c) {
+                c->setPlayEventType(PlayEventType::User);
+            }
+        }
+
```
Jojo-Schmitz, line 4775: See https://github.com/musescore/MuseScore/pull/8282#pullrequestreview-721270042

lvinken:
Will do.

lvinken:
Rebased.

Jojo-Schmitz:
One if the unit tests is unhappy, and it is a related one. It does seem to like the beam positions.

lvinken:
The unit test fails on a new testfile (introduced with #321753), which I failed to update for the current master. Will do.

Jojo-Schmitz:
Rebase needed

lvinken:
Will have a look. Note that this seems to have been ignored for a long time without any feedback. I am unsure whether this will be merged at all :-(.

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -4704,6 +4704,25 @@ Note* MusicXMLParserPass2::note(const QString& partId,
...
+            && mnd.calculatedDuration() != mnd.specifiedDuration()) {
+            // convert duration into note length
+            Fraction durationMult { (mnd.specifiedDuration() / mnd.calculatedDuration()).reduced() };
+            durationMult = (1000 * durationMult).reduced();
+            const int noteLen = durationMult.numerator() / durationMult.denominator();
+
+            NoteEventList nel;
+            NoteEvent ne;
+            ne.setLen(noteLen);
+            nel.append(ne);
```
Jojo-Schmitz, line 4719: That method doesn't exist (anymore? The move from `QVector` to `std::vector`?, if so it'd be 'nel.push_back(ne)` here I guess).

src/importexport/musicxml/tests/data/testDurationLargeError.xml
```diff
@@ -0,0 +1,265 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
+<score-partwise version="3.1">
```
Jojo-Schmitz, line 3: Probably needs to be 4.0

src/importexport/musicxml/tests/data/testDurationLargeError_ref.xml
```diff
@@ -0,0 +1,237 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
+<score-partwise version="3.1">
```
Jojo-Schmitz, line 3: Probably needs to be 4.0

RomanPudashkin (approved):

PR #8430, opened 2021-06-25
[MU3 Backend] ENG-54: Staff Visibility

iveshenry18:
Resolves: [ENG-54](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-54): Show both staves of grand staff even when one is empty

Since HideMode::AUTO behaves in uncommon ways for most scores (hiding
empty staves in a grand staff), this changes the hideWhenEmpty to
INSTRUMENT when creating the Staffs upon import. In the case that a
score part has a variable number of staves, it uses this to infer that
we do truly want HideMode::AUTO.

Previously, a staff's emptiness was calculated based on its note and
annotation content. Since this is evaluated before the system was laid
out, the system did not have pointers to its SpannerSegments. This
commit adds a way to include whether a staff has any spanners
(particularly pedal markings and hanging slurs) in assessing emptiness,
creating more consistent/expected hiding behavior in HideMode::AUTO.


oktophonie (approved):

vpereverzev (approved):

libmscore/layout.cpp
```diff
@@ -1339,6 +1343,12 @@ void Score::hideEmptyStaves(System* system, bool isFirstSystem)
                     && !(isFirstSystem && styleB(Sid::dontHideStavesInFirstSystem))
                     && hideMode != Staff::HideMode::NEVER)) {
                   bool hideStaff = true;
+                  for (auto spanner : spanners) {
+                        if (spanner.value->staff() == staff) {
```
Jojo-Schmitz, line 1347: Simlpe fix seems to be to add ` && !spanner.value->systemFlag()` to the condition here.

iveshenry18: Good catch! I'll go ahead and make that tweak today, as well as add something to the test case to catch future regressions of that specific variety.

Jojo-Schmitz: On the danger to repeat myself: For getting regressions like this caught better and quicker, we'd really need 'nightly builds', so have the PRs on top of the 3.x or master branch.
Having them on top of 3.x and they'd easily could get `git cherry-pick`ed (I did the opposite and cherry-picxked them from 3.6.2_backend to 3,x, but I'm just one and can't kill'em all ;-)), having them on master (too) takes care of them not getting forgotten.

PR #8431, opened 2021-06-25
Update instruments.xml to add families and improve sorting

shoogle:
Run share/instruments/update_instruments_xml.py to fetch the latest version of the online spreadsheet. In this version, every instrument has been assigned a family. Also, instruments are now sorted based on their group, family and minimum professional pitch, among other things. This ensures they appear roughly in score order according to standard orchestral layout. Finally, descriptions have been improved and a few small fixed made to fix incorrect or missing data.

In addition, some compatibility code was changed to improve instrument recognition in older MSCX score files that lack the appropriate MuseScore or MusicXML instrument IDs. The old code returned the *first* instrument that matched a given criteria (e.g. same trackName), whereas the new code returns the instrument that gives the *best* match according to multiple criteria. This means the return value is less dependent on the order in which instruments are defined within instruments.xml.

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -34,32 +34,128 @@
       <Genre id="classroom">
             <name>Classroom</name>
       </Genre>
+      <Family id="winds">
+            <name>Winds</name>
+      </Family>
       <Family id="flutes">
             <name>Flutes</name>
       </Family>
-      <Family id="oboes">
-            <name>Oboes</name>
+      <Family id="dizis">
+            <name>Dizis</name>
       </Family>
-      <Family id="clarinets">
-            <name>Clarinets</name>
+      <Family id="fifes">
+            <name>Fifes</name>
       </Family>
-      <Family id="bassoons">
-            <name>Bassoons</name>
+      <Family id="whistles">
+            <name>Whistles</name>
       </Family>
-      <Family id="saxophones">
-            <name>Saxophones</name>
+      <Family id="flageolets">
+            <name>Flageolets</name>
       </Family>
       <Family id="recorders">
             <name>Recorders</name>
       </Family>
+      <Family id="ocarinas">
+            <name>Ocarinas</name>
+      </Family>
+      <Family id="gemshorns">
+            <name>Gemshorns</name>
+      </Family>
+      <Family id="pan-flutes">
+            <name>Pan Flutes</name>
+      </Family>
+      <Family id="quenas">
+            <name>Quenas</name>
+      </Family>
+      <Family id="oboes">
+            <name>Oboes</name>
+      </Family>
+      <Family id="shawms">
+            <name>Shawms</name>
+      </Family>
+      <Family id="cromornes">
+            <name>Cromornes</name>
+      </Family>
+      <Family id="crumhorns">
+            <name>Crumhorns</name>
+      </Family>
+      <Family id="cornamuses">
+            <name>Cornamuses</name>
+      </Family>
+      <Family id="kelhorns">
+            <name>Kelhorns</name>
+      </Family>
+      <Family id="rauschpfeifes">
+            <name>Rauschpfeifes</name>
+      </Family>
+      <Family id="duduks">
+            <name>Duduks</name>
+      </Family>
+      <Family id="shenais">
+            <name>Shenais</name>
+      </Family>
+      <Family id="clarinets">
+            <name>Clarinets</name>
+      </Family>
+      <Family id="chalumeaus">
+            <name>Chalumeaus</name>
+      </Family>
+      <Family id="xaphoons">
+            <name>Xaphoons</name>
+      </Family>
+      <Family id="tarogatos">
+            <name>Tarogatos</name>
+      </Family>
+      <Family id="octavins">
+            <name>Octavins</name>
+      </Family>
+      <Family id="saxophones">
+            <name>Saxophones</name>
+      </Family>
+      <Family id="bassoons">
+            <name>Bassoons</name>
+      </Family>
+      <Family id="reed-contrabasses">
+            <name>Reed Contrabasses</name>
+      </Family>
+      <Family id="dulcians">
+            <name>Dulcians</name>
+      </Family>
+      <Family id="racketts">
+            <name>Racketts</name>
+      </Family>
+      <Family id="sarrusophones">
+            <name>Sarrusophones</name>
+      </Family>
+      <Family id="bagpipes">
+            <name>Bagpipes</name>
+      </Family>
+      <Family id="accordions">
+            <name>Accordions</name>
+      </Family>
+      <Family id="harmonicas">
+            <name>Harmonicas</name>
+      </Family>
+      <Family id="melodicas">
+            <name>Melodicas</name>
+      </Family>
+      <Family id="shengs">
+            <name>Shengs</name>
+      </Family>
+      <Family id="brass">
+            <name>Brass</name>
+      </Family>
       <Family id="horns">
             <name>Horns</name>
+      </Family>
+      <Family id="wagner-tubas">
+            <name>Wagner Tubas</name>
       </Family>
       <Family id="cornets">
             <name>Cornets</name>
       </Family>
-      <Family id="flugelhorns">
-            <name>Flugelhorns</name>
+      <Family id="saxhorns">
+            <name>Saxhorns</name>
       </Family>
       <Family id="alto-horns">
             <name>Alto Horns</name>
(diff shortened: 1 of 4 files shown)


share/instruments/update_instruments_xml.py
```diff
@@ -159,12 +160,20 @@ def to_list(str):
...
         to_subelement(el, instrument, 'shortName')
-        to_subelement(el, instrument, 'init')
+        # if instrument["ddName"] != '[hide]':
+        #     to_subelement(el, instrument, 'trackName')
+        #     to_subelement(el, instrument, 'longName')
+        #     to_subelement(el, instrument, 'shortName')
+        #     if instrument["ddName"]:
+        #         dd_el = ET.SubElement(el, 'dropdownName')
+        #         dd_el.text = instrument["ddName"]
+        #         to_attribute(dd_el, instrument, 'ddMeaning', 'meaning')
```
shoogle, line 176: @RomanPudashkin, once this is merged, you can uncomment these 8 lines and delete the 3 lines immediately above (that set `trackName`, `longName` and `shortName` before the `if` statement), then rerun the script to generate the new style of instruments.xml that you will need to fix the dropdown in the New Score Wizard. It will fetch some new data from the online spreadsheet as well as adding the `<dropdownName>` element.

njvdberg:
@RomanPudashkin Score Ordering uses the order of instruments in `instruments.xml` to find the order of the instruments in the **same** family. Is this taken into account with this PR? This might a possible cause of incompatibility with existing scores. Since families can be overwritten in score orders, it is not enough to keep the order of instruments within the families intact. This has to be checked with existing score orders in `orders.xml` as well in the templates in `share/templates`.

shoogle:
@njvdberg, the new ordering system has been devised by myself and @oktophonie. We know that changing the order of instruments in `instruments.xml` will affect score ordering for instruments within a family. That much is intensional, but we would be interested to learn about other problems that we may not have foreseen, or places where you think our new ordering (which you can see in the [online spreadsheet](https://docs.google.com/spreadsheets/d/1SwqZb8lq5rfv5regPSA10drWjUAoi65EuMoYtG-4k5s/edit#gid=516529997)) is incorrect.

### New scores vs. old scores

We want new scores created in MuseScore 4 to use our new ordering, but we want old scores created in MuseScore 3 to display in MuseScore 4 just as they did in MuseScore 3. When you open an old file the ordering should be unchanged, but if you try to add or remove instruments then things could start to move around. Possible solutions are:

1. Do nothing and accept that adding instruments will cause the order to change (hopefully not by much).
2. Lock the instruments dialog for imported scores until the user opts-in to the new ordering system.
2. Create a new custom ordering for imported scores. Newly added instruments must be placed manually.

### Algorithmic sorting

If all else fails then we may consider reverting to the previous ordering of `instruments.xml`, but the old ordering was done manually and we would prefer to order instruments based on an algorithm if possible as this will make it easier to add new instruments in the future. I will explain how the algorithm works in a future build review.

We don't think our algorithm has changed the relative order of any popular combinations of instruments, except perhaps for the unpitched percussion instruments (which have no "correct" ordering anyway). If people have used some of the more exotic instruments then they may find they are ordered differently to before, but the ordering for the vast majority of scores should not have changed.

njvdberg:
I had a quick check on the differences between the old and new `instruments.xml` and see a few possible conflicts but this mainly concerns "exotic" instruments which I guess won't be used with score ordering in the first place. So for the "common" instruments I don't expect any issues. Maybe except for `unsorted woodwinds` because the complete saxophone section is moved above the bassoon. But on the other hand, it was unsorted already.

Also, in MU4 score ordering I made a small change compared to MU3. When opening the instruments form the existing order of the instruments is checked against the specified score order. If it doesn't match, the score order is change to `<score order> (customized)` and the "wrong" order is kept as it is (at least that's the intention :smiley:).

shoogle:
@njvdberg, please take a look at my changes to [instrtemplate.cpp](https://github.com/musescore/MuseScore/pull/8431/files?file-filters%5B%5D=.cpp#diff-b1c09164aa01fb825af5f4b3ab1468053d7086638d802e4d4a34b2ccbbf0a289) and [instrument.cpp](https://github.com/musescore/MuseScore/pull/8431/files?file-filters%5B%5D=.cpp#diff-ac87c1cf46d5c22019a09ef85703b4976e26a329ad5b019f7fdce27d0e2d3487) and let me know what you think.

src/engraving/libmscore/instrument.cpp
```diff
@@ -1739,31 +1739,48 @@ void Instrument::updateInstrumentId()
...
-                } else {
-                    _id = it->id;
-                    return;
+                }
+            } else {
+                int matchStrength = 0
+                                    + (minPitchP() == it->minPitchP)
+                                    + (minPitchA() == it->minPitchA)
+                                    + (maxPitchA() == it->maxPitchA)
+                                    + (maxPitchP() == it->maxPitchP);
```
njvdberg, line 1774: Using pitches is a little tricky. E.g. the lowest pitch for a baritone saxophone is a Bb (written) for amateur and A (written) for professional. This isn't 100% true, the lowest note in this case is defined by the instrument, not the capability of the player (for the low A the instrument needs an extension). Same for Bb clarinets which can have an extension to Eb (instead of having an E as lowest note).
When users decide to change the range this test will fail (BTW, when I have to write for instruments I'm not that familiar with, I actually change the ranges, amateur to what is comfortable to play, professional to what is is possible for the particular player. Now MuseScore gives an indication whether the part is playable for the player I'm writing for).
So, if pitches are used, the `==` isn't that great choice. But whether `<=` or `>=` should be used, no idea.
Maybe the MIDI channel can be used too?

src/engraving/libmscore/instrtemplate.cpp
```diff
@@ -707,18 +707,23 @@ InstrumentTemplate* searchTemplateForMusicXmlId(const QString& mxmlId)
...
+    int bestMatchStrength = 0; // higher for better matches
     for (InstrumentGroup* g : qAsConst(instrumentGroups)) {
         for (InstrumentTemplate* it : qAsConst(g->instrumentTemplates)) {
             for (const QString& name : nameList) {
-                if (it->trackName == name
-                    || it->longNames.contains(StaffName(name))
-                    || it->shortNames.contains(StaffName(name))) {
-                    return it;
+                int matchStrength = 0
+                                    + (4 * (it->trackName == name)) // most weight to track name since there are fewer duplicates
```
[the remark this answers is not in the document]
shoogle: The "Piano", "Grand Piano", and "Upright Piano" all have the same short name: "Pno.". There may be other cases, either now or in the future as we add more instruments.

src/engraving/libmscore/instrument.cpp
shoogle: We can check against the MIDI program and MIDI bank number for the default channel (i.e. the first channel in the channels list).

> When users decide to change the range this test will fail

The ranges are being used as a weighting to choose between multiple instruments with the same MusicXML ID. The ranges are not used as a strict condition, so if no instrument matches the given ranges then the code simply returns the first instrument that matches the MusicXML ID (actually, it returns the instrument with the shortest ID that matches the MusicXML ID).

But if we wanted to be really fancy we could do this:

```c++
int matchStrength = 100
    - qAbs(minPitchP() - it->minPitchP)
    - qAbs(minPitchA() - it->minPitchA)
    - qAbs(maxPitchA() - it->maxPitchA)
    - qAbs(maxPitchP() - it->maxPitchP);
```

But this may not work so well for unpitched instruments or instruments with missing pitch data, and I think it's probably not worth the effort to work around those cases.

share/instruments/instruments.xml
```diff
@@ -121,21 +292,75 @@
...
+            <name>Tamburicas</name>
+      </Family>
+      <Family id="bandurrias">
+            <name>Bandurrias</name>
+      </Family>
+      <Family id="lauds">
+            <name>Lauds</name>
+      </Family>
+      <Family id="strings">
+            <name>Strings</name>
```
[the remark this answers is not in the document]
shoogle: I don't know what problems he faced in the past, but these families were all suggested by Simon.

src/engraving/libmscore/instrtemplate.cpp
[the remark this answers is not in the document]
shoogle: I have added a quick exit if there is an exact match.

src/engraving/libmscore/instrument.cpp
shoogle: Not checking against MIDI program and bank number for now since the current code works without it. I looked at adding it but it's not elegant unless we get rid of the special handling for strings, so I'm leaving it as it is for now.

RomanPudashkin (approved):

Issue #8434, opened 2021-06-26
[MU4 Issue] Extra "F"s appear after each note when adding notes in step-input mode


wizofaus:
**Describe the bug**
Every time a note is added via the keyboard (A-G) in step-input mode an additional F is added afterwards.

**To Reproduce**
1. Create new score
2. Type A B C
3. Note that the notes A F B F C F are added

**Expected behavior**
F's should not be added

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**
Started after recent commit 88e35a16e247bda288062089308f60b3aea41f9f merged ("Implemented add fret actions")

PR #8436, opened 2021-06-26
Ensure actions enabled per staff type (prevent shortcuts firing multiple actions)

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8434

Better fix for this - ensure that actions like note-a ... note-g and fret-0 ... fret-7 which share shortcuts by default are only enabled based on current staff type (TAB or standard)


wizofaus:
@cbjeukendrup - I'd be keen to discuss how this concept could be extended further for all actions handled by the notationactioncontroller, it should mean we can simplify the uicontext stuff.  I'd also like to remove the hard-coded special cases from canReceiveAction( ).

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -68,13 +68,13 @@ void NotationActionController::init()
...
-    dispatcher()->reg(this, "note-a", [this]() { addNote(NoteName::A, NoteAddingMode::NextChord); });
-    dispatcher()->reg(this, "note-b", [this]() { addNote(NoteName::B, NoteAddingMode::NextChord); });
+    registerAction("note-c", [this]() { addNote(NoteName::C, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
+    registerAction("note-d", [this]() { addNote(NoteName::D, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
+    registerAction("note-e", [this]() { addNote(NoteName::E, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
+    registerAction("note-f", [this]() { addNote(NoteName::F, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
+    registerAction("note-g", [this]() { addNote(NoteName::G, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
+    registerAction("note-a", [this]() { addNote(NoteName::A, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
+    registerAction("note-b", [this]() { addNote(NoteName::B, NoteAddingMode::NextChord); }, &NotationActionController::isStandardStaff);
```
wizofaus, line 77: Could probably be done in a loop too - except that the NoteName enum increases from C-B whereas the char codes for c - b don't, and C++ doesn't allow indexing of enum values that I'm aware of.  But there is a lot of repetition here...

cbjeukendrup:
@wizofaus The latest version of the actions system was implemented by @igorkorsukov, so we should also ask him. 

I spent some time looking at how the current system for determining an action's state works exactly, and now that I fully understand it, I think it's quite good.

However, the problem of those UiContexts is that there is in fact not _one_ Ui state at one time. There are many aspects: the state of the undo stack, the state of the note input, the state of the selection in the notation, et cetera. 
And some actions work in multiple contexts, which requires us to create contexts like `UiCtxNotationOpenedOrTextEditing`. I think this is not ideal, and really complicates the logic for matching contexts. So I think we need to do two things:
- Create the possibility to have actions that work in multiple contexts (maybe somewhat like the flags system that was used in MuseScore 3)
- Keep the contexts general (and maybe mutually exclusive, like in MuseSore 3?), and handle specific cases in methods like NotationUiActions::actionEnabled. 

Especially for the second point, I think this PR is a good step. The implementation of `isStandardStaff` might be incorrect because there are also percussion staves, and those may also require special handling, but that's just a detail.

wizofaus:
That compound state (notation  opened or text editing) was something I introduced to allow clipboard actions to work while editing text but I'd rather get rid of the "text editing" context mode and use this new method instead. 
Yes I know about percussion staves and orginally i was going to call it "isNotTablatureStaff", it just seemed a bit of a mouthful!

wizofaus:
This is a minimal fix to prevent shortcut keys firing both note-* and fret-* actions, but https://github.com/musescore/MuseScore/pull/8437 is a much more complete fix that addresses a number of other issues (and I've spent a lot more time testing!).

igorkorsukov:
> @wizofaus The latest version of the actions system was implemented by @igorkorsukov, so we should also ask him.
> 
> I spent some time looking at how the current system for determining an action's state works exactly, and now that I fully understand it, I think it's quite good.
> 
> However, the problem of those UiContexts is that there is in fact not _one_ Ui state at one time. There are many aspects: the state of the undo stack, the state of the note input, the state of the selection in the notation, et cetera.
> And some actions work in multiple contexts, which requires us to create contexts like `UiCtxNotationOpenedOrTextEditing`. I think this is not ideal, and really complicates the logic for matching contexts. So I think we need to do two things:
> 
> * Create the possibility to have actions that work in multiple contexts (maybe somewhat like the flags system that was used in MuseScore 3)
> * Keep the contexts general (and maybe mutually exclusive, like in MuseSore 3?), and handle specific cases in methods like NotationUiActions::actionEnabled.
> 
> Especially for the second point, I think this PR is a good step. The implementation of `isStandardStaff` might be incorrect because there are also percussion staves, and those may also require special handling, but that's just a detail.

Yes, multi contexts have been asking for a long time, but I wanted to postpone the moment of added multi contents as much as possible to be sure that they are really needed (their adding can introduce a lot of complications and confusion).

@wizofaus It's great that you are actively developing the system. It is a good idea to immediately specify the availability condition when registering an action.

src/notation/internal/notationactioncontroller.h
```diff
@@ -186,6 +190,9 @@ class NotationActionController : public actions::Actionable, public async::Async
     bool canUndo() const;
     bool canRedo() const;
     bool isNotationPage() const;
+    bool isStandardStaff() const;
+    bool isTablatureStaff() const;
+    void registerAction(const mu::actions::ActionCode&, std::function<void()>, bool (NotationActionController::*)() const);
```
igorkorsukov, line 195: Maybe it makes sense, instead of `bool (NotationActionController::*)() const`, to make a more general solution, like `std::function<bool()>`

wizofaus: They're both there...but I'd suggest you review my other PR first, it uses this idea but expands it considerably and simplifies the UiContext stuff too (no more "text editing" context for a start!).

src/notation/internal/notationactioncontroller.cpp
igorkorsukov: > registerNoteAction('a', NoteName::A);
Good idea, we can just add a lambda ...

src/notation/internal/notationactioncontroller.h
igorkorsukov: I looked, good solution!

PR #8437, opened 2021-06-27
Fixed undo stack and simplified/improved action enabling

wizofaus:
Resolves: Quite a few issues!

Supersedes https://github.com/musescore/MuseScore/pull/8436

### DIFF
--- a/src/context/internal/uicontextresolver.cpp
+++ b/src/context/internal/uicontextresolver.cpp
@@ -51,6 +51,10 @@ void UiContextResolver::init()
             notation->interaction()->textEditingStarted().onNotify(this, [this]() {
                 notifyAboutContextChanged();
             });
+
+            notation->undoStack()->stackChanged().onNotify(this, [this]() {
+                notifyAboutContextChanged();
+            });
         }
         notifyAboutContextChanged();
     });
@@ -82,10 +86,6 @@ UiContext UiContextResolver::currentUiContext() const
             return context::UiCtxUnknown;
         }
 
-        if (notation->interaction()->isTextEditingStarted()) {
-            return context::UiCtxNotationTextEditing;
-        }
-
         ui::INavigationSection* activeSection = navigationController()->activeSection();
         if (!activeSection || activeSection->name() == NOTATION_NAVIGATION_SECTION) {
             return context::UiCtxNotationFocused;
@@ -99,22 +99,13 @@ UiContext UiContextResolver::currentUiContext() const
 
 bool UiContextResolver::match(const ui::UiContext& currentCtx, const ui::UiContext& actCtx) const
 {
-    //! NOTE If now editing the notation text, then we allow actions only with the context `UiCtxNotationTextEditing`,
-    //! all others, even `UiCtxAny`, are forbidden
-    if (currentCtx == context::UiCtxNotationTextEditing) {
-        return actCtx == context::UiCtxNotationTextEditing || actCtx == context::UiCtxNotationOpenedOrTextEditing;
-    }
-
     if (actCtx == context::UiCtxAny) {
         return true;
     }
 
     //! NOTE If the current context is `UiCtxNotationFocused`, then we allow `UiCtxNotationOpened` too
-    if (currentCtx == context::UiCtxNotationFocused) {
-        if (actCtx == context::UiCtxNotationOpened || actCtx == context::UiCtxNotationOpenedOrTextEditing) {
-            return true;
-        }
-        return actCtx == context::UiCtxNotationFocused;
+    if (currentCtx == context::UiCtxNotationFocused && actCtx == context::UiCtxNotationOpened) {
+        return true;
     }
 
     return currentCtx == actCtx;
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -4412,6 +4412,7 @@ void Score::cmdAddFret(int fret)
         qDebug("cannot enter notes here (no chord rest at current position)");
         return;
     }
+    is.setRest(false);
     Position pos;
     pos.segment   = is.segment();
     pos.staffIdx  = is.track() / VOICES;
--- a/src/engraving/libmscore/input.cpp
+++ b/src/engraving/libmscore/input.cpp
@@ -212,6 +212,8 @@ void InputState::update(Selection& selection)
 
     Element* e = selection.element();
     if (e == 0) {
+        setTrack(selection.activeTrack());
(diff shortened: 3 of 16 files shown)


src/context/internal/uicontextresolver.cpp
```diff
@@ -51,6 +51,10 @@ void UiContextResolver::init()
             notation->interaction()->textEditingStarted().onNotify(this, [this]() {
                 notifyAboutContextChanged();
             });
+
+            notation->undoStack()->stackChanged().onNotify(this, [this]() {
+                notifyAboutContextChanged();
```
wizofaus, line 56: This appears to be the most reliable trigger for ensuring all action's enabled status are updated as needed

src/engraving/libmscore/textedit.cpp
```diff
@@ -248,12 +248,18 @@ void TextBase::insertText(EditData& ed, const QString& s)
...
+        return false; 
+    }
     TextCursor* cursor = ted->cursor();
 
     // do nothing on Shift, it messes up IME on Windows. See #64046
     if (ed.key == Qt::Key_Shift) {
         return false;
     }
+    if (ed.key == Qt::Key_Escape) {
+        return false;
```
wizofaus, line 261: Seems MU3 was behaving quite differently here, the Text elements were only getting keystrokes depending on what widget group and its related "ActionGroup" was active, but that concept no longer exists in MU4, so all keystrokes are arriving here, and the element must return false for any it doesn't want.

src/inspector/models/abstractinspectormodel.cpp
```diff
@@ -235,7 +235,7 @@ Ms::Sid AbstractInspectorModel::styleIdByPropertyId(const Ms::Pid pid) const
 
 void AbstractInspectorModel::updateStyleValue(const Ms::Sid& sid, const QVariant& newValue)
 {
-    if (style()) {
+    if (style() && style()->styleValue(sid) != newValue) {
         beginCommand();
```
wizofaus, line 239: this was causing some truly bizarre and random behaviour - basically the inspector was triggering changes to the undo stack even if nothing had changed, just by clicking on empty parts of the score

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -954,22 +946,20 @@ void NotationActionController::addFret(int fretIndex)
...
 {
-    auto notation = currentNotation();
-    if (!notation) {
+    auto interaction = currentNotationInteraction();
+    if (!interaction) {
         return;
     }
-
-    notation->undoStack()->undo();
+    interaction->undo();
```
wizofaus, line 953: Needed to move this to the next layer down because the EditData structure is needed for undo/redo to work properly when editing text elements

src/notation/internal/notationinteraction.cpp
```diff
@@ -1949,12 +1949,14 @@ void NotationInteraction::editText(QKeyEvent* event)
     m_textEditData.key = event->key();
     m_textEditData.modifiers = event->modifiers();
     m_textEditData.s = event->text();
+    startEdit();
```
wizofaus, line 1952: necessary to ensure undo stack works properly while editing text elements

src/notation/internal/notationuiactions.cpp
```diff
@@ -157,15 +157,15 @@ const UiActionList NotationUiActions::m_actions = {
              QT_TRANSLATE_NOOP("action", "Decrease duration dotted")
              ),
     UiAction("cut",
-             mu::context::UiCtxNotationOpenedOrTextEditing,
+             mu::context::UiCtxNotationFocused,
```
wizofaus, line 160: not sure if it should be Opened or Focused...

src/engraving/libmscore/cmd.cpp
```diff
@@ -4412,6 +4412,7 @@ void Score::cmdAddFret(int fret)
         qDebug("cannot enter notes here (no chord rest at current position)");
         return;
     }
+    is.setRest(false);
```
wizofaus, line 4415: To match cmdAddPitch, necessary if you've selected an empty measure.

src/engraving/libmscore/input.cpp
```diff
@@ -212,6 +212,8 @@ void InputState::update(Selection& selection)
 
     Element* e = selection.element();
     if (e == 0) {
+        setTrack(selection.activeTrack());
+        setSegment(selection.startSegment());
```
wizofaus, line 216: ensures that when an empty measure selected the active track/segment are set correctly.
Note that this was broken in MU3 too.

wizofaus:
Lots of changes here because I kept discovering lots of bugs/issues trying to test the original changes I'd made simply to solve the problem of how shortcuts fired between Tablature and non-Tablature staves.
In fact even MU3 has bugs in this area - if you're on a standard staff and select an empty measure in a TAB staff, it doesn't correctly disable the "add note" commands and/or enable the "add fret" commands - e.g. if you hit a number key (0-9), it actually changes the duration in the standard staff.
The undo stack was horribly broken in general, I'd been noticing that quite a bit - I won't say it's perfect now but it does largely work and I haven't triggered any random crashes for a while (which happened all the time before).

src/notation/internal/notationinteraction.cpp
```diff
@@ -1949,12 +1950,16 @@ void NotationInteraction::editText(QKeyEvent* event)
...
     if (m_textEditData.element->edit(m_textEditData)) {
         event->accept();
+        apply();
+    } else {
+        m_undoStack->rollbackChanges();
+    }
+    if (!wasEditingText) {
+        m_textEditData.element = nullptr;
```
wizofaus, line 1961: Fixes a bug I'd introduced a while back, where `m_textEditData.element` could be left referring to a deleted element after undo etc., caused very weird app crashes that MSVC debugger couldn't even catch.

igorkorsukov (approved):

src/notation/internal/notationuiactions.cpp
```diff
@@ -349,13 +349,13 @@ const UiActionList NotationUiActions::m_actions = {
              QT_TRANSLATE_NOOP("action", "Edit score properties")
              ),
     UiAction("undo",
-             mu::context::UiCtxNotationOpenedOrTextEditing,
+             mu::context::UiCtxNotationFocused,
```
Eism, line 352: @wizofaus 
Can you explain why you chose UiCtxNotationFocused over UiCtxNotationOpened here?
This is a global shortcut for score and doesn't need the notation to have focus.

PR #8449, opened 2021-06-29
Added mi sync for cloud token

igorkorsukov:


Eism:
Need to add resource guard to the CloudService::signOut method, because the file is deleted there

igorkorsukov:
> Need to add resource guard to the CloudService::signOut method, because the file is deleted there

Thanks! Added.

Eism (approved):

PR #8459, opened 2021-06-29
Changed namespace for modularity

igorkorsukov:


src/userscores/userscoresmodule.cpp
```diff
@@ -123,9 +123,9 @@ void UserScoresModule::registerUiTypes()
     ioc()->resolve<ui::IUiEngine>(moduleName())->addSourceImportPath(userscores_QML_IMPORT);
 }
 
-void UserScoresModule::onInit(const IApplication::RunMode& mode)
+void UserScoresModule::onInit(const framework::IApplication::RunMode& mode)
 {
-    if (IApplication::RunMode::Converter == mode) {
+    if (framework::IApplication::RunMode::Converter == mode) {
```
vpereverzev, line 128: probably you have to put "using namespace mu", because it's expecting mu::framework::IApplication here

igorkorsukov: Nope, we are in the context of the `mu` namespace (mu::userscores)

PR #8463, opened 2021-06-30
qpen removed from engraving

alexpavlov96:
*removed QPen from engraving (as a part of the task - removing all Qt dependencies from engraving)*


src/engraving/draw/drawtypes.h
```diff
@@ -38,6 +38,33 @@ enum class PolygonMode {
     Convex,
     Polyline
 };
+
+enum PenStyle { // pen style
```
igorkorsukov, line 42: better add `enum class ...` and use `PenStyle::NoPen` and etc

igorkorsukov: Also for other enums

src/engraving/draw/pen.h
```diff
@@ -0,0 +1,147 @@
...
+        setDefaults();
+        _style = style;
+    }
+    
+    Pen(const QColor &color) {
+        setDefaults();
+        _brush = color;
+    }
+    
+    Pen(const QBrush &brush, double width, PenStyle s = SolidLine,
```
igorkorsukov, line 46: I believe that `QBrush` is not needed, this constructor can be replaced with `Pen(const QColor &color, double width,...)`

src/engraving/draw/pen.h
```diff
@@ -0,0 +1,147 @@
...
+    void setWidthF(double width) { _width = width; }
+
+
+    int width() const { return qRound(_width); }
+    void setWidth(int width) { _width = width; }
+
+    QColor color() const { return _brush.color(); }
+    void setColor(const QColor &color) { _brush = QBrush(color); }
+    QBrush brush() const { return _brush; }
+    void setBrush(const QBrush &brush) { _brush = brush; }
```
igorkorsukov, line 112: I believe that QBrush is not needed, can be removed

src/engraving/draw/pen.h
```diff
@@ -0,0 +1,147 @@
...
+            _dashPattern << 1;
+        }
+    }
+    
+    double widthF() const { return _width; }
+    void setWidthF(double width) { _width = width; }
+
+
+    int width() const { return qRound(_width); }
+    void setWidth(int width) { _width = width; }
```
igorkorsukov, line 107: I believe that we can only keep `widthF`

src/engraving/draw/pen.h
```diff
@@ -0,0 +1,147 @@
...
+#endif
+
+private:
+    void setDefaults() {
+        _brush = Qt::black;
+        _width = _defaultWidth;
+        _style = _penDefaultStyle;
+        _capStyle = _penDefaultCap;
+        _joinStyle = _penDefaultJoin;
+    }
```
igorkorsukov, line 130: I believe that this method is unnecessary and members like `defaultX` are not needed
It is sufficient (and necessary) to simply initialize the fields with their default values.
```
double _width = 1.0;
PenStyle _style = SolidLine;
...
```

src/engraving/draw/qpainterprovider.cpp
```diff
@@ -132,19 +132,21 @@ const Font& QPainterProvider::font() const
     return m_font;
 }
 
-void QPainterProvider::setPen(const QPen& pen)
+void QPainterProvider::setPen(const Pen& pen)
 {
-    m_painter->setPen(pen);
+    m_pen = pen;
+    m_painter->setPen(m_pen.toQPen());
```
igorkorsukov, line 138: If we introduce a member `m_pen`, then it needs to be initialized with a pen from the painter, as is done for the font.
But I think in this case we don't need to add a class member, it's enough just to do in the getter
```
const Pen& QPainterProvider::pen() const
{
    return Pen::fromQPen(m_painter->pen());
}
```

src/engraving/libmscore/mmrest.cpp
```diff
@@ -95,8 +96,9 @@ void MMRest::draw(mu::draw::Painter* painter) const
             x += symBbox(sym).width() + spacing;
         }
     } else {
-        QPen pen(painter->pen());
-        pen.setCapStyle(Qt::FlatCap);
+        using namespace mu::draw;
```
igorkorsukov, line 99: I believe that in such simple cases, ` using namespace mu::draw;`  is not needed, or it should be added uniformly in all places, i.e., for example, at the beginning of each `draw` method. (rather than sticking in the middle of the method)

src/engraving/draw/painter.h
```diff
@@ -25,7 +25,7 @@
 #include <list>
 #include <stack>
 
-#include <QPen>
+#include "pen.h"
```
igorkorsukov, line 28: should be moved below

igorkorsukov:
@alexpavlov96 Please change the code in accordance with our accepted code style https://github.com/musescore/MuseScore/wiki/CodeGuidelines

src/engraving/draw/pen.h
```diff
@@ -0,0 +1,147 @@
...
+    
+    QBrush _brush;
+    double _width;
+    PenStyle _style;
+    PenCapStyle _capStyle;
+    PenJoinStyle _joinStyle;
+    mutable QVector<double> _dashPattern; s
```
Jojo-Schmitz, line 137: That stray `s` is screwing up the entire build

Jojo-Schmitz: (it results in `_defaultWidth` not to be seen)

src/engraving/draw/pen.h
```diff
@@ -0,0 +1,147 @@
...
+    
+    PenStyle style() const { return _style; }
+    void setStyle(PenStyle style) {
+        if (_style == style)
+            return;
+        _style = style;
+        _dashPattern.clear();
+      }
+    
+    QVector<double> dashPattern() const
```
igorkorsukov, line 58: Better use `std::vector`

src/engraving/draw/qpainterprovider.cpp
alexpavlov96: @igorkorsukov then we have a reference to a local object, i will initialize m_pen from the painter

src/engraving/draw/drawtypes.h
alexpavlov96: @igorkorsukov is it fine to leave like this until we don't have Qt? Otherwise I have conflicts with Qt namespace and I don't want to rename or add extra namespace

igorkorsukov: Did not understand you. enum class, on the contrary, excludes conflicts

src/engraving/draw/utils/drawcomp.cpp
```diff
@@ -89,13 +89,15 @@ static bool isEqual(const QSize& sz1, const QSize& sz2)
...
+static bool isEqual(const Pen& p1, const Pen& p2)
 {
     if (p1.style() != p2.style()) {
         return false;
     }
+    
+    double epsilon = 0.0001;
 
-    if (p1.width() != p2.width()) {
+    if (fabs(p1.widthF() - p2.widthF()) > epsilon) {
```
igorkorsukov, line 100: better use `!RealIsEqual(...)`

Jojo-Schmitz:
Still codestyle issues (at least)

src/engraving/libmscore/box.cpp
```diff
@@ -111,13 +111,10 @@ void Box::draw(mu::draw::Painter* painter) const
         qreal w = spatium() * .15;
         QPainterPathStroker stroker;
         stroker.setWidth(w);
-        stroker.setJoinStyle(Qt::MiterJoin);
-        stroker.setCapStyle(Qt::SquareCap);
+        stroker.setJoinStyle(Qt::PenJoinStyle::MiterJoin);
+        stroker.setCapStyle(Qt::PenCapStyle::SquareCap);
```
igorkorsukov, line 115: shouldn't be `Qt::`

Jojo-Schmitz:
Wouldn't a rebase get the CI_vtest to run?

PR #8466, opened 2021-06-30
[MU4] The base of Learn page

Eism:
image


src/main/main.cpp
```diff
@@ -212,6 +218,12 @@ int main(int argc, char** argv)
     app.addModule(new mu::userscores::UserScoresStubModule());
 #endif
 
+#ifdef BUILD_LEARN_MODULE
+    app.addModule(new mu::learn::LearnModule());
+#else
+    app.addModule(new mu::learn::LearnStubModule());
+#endif
```
igorkorsukov, line 225: I believe that we can give the stub module the same name as the real one (do not add a stub) and then there will be no need to do ifdef here - the code is cleaner and more convenient.

igorkorsukov (approved):

src/main/main.cpp
Eism: thanks, fixed

PR #8467, opened 2021-06-30
[MU4] audio_refactoring_part_1

vpereverzev:
This PR introduces the rework of the audio-engine in MU4.

Each part in the score has its own track in the audio engine. Each track is combined into a sequence track and has its own personal audio source and mixer channel, so each track can be set independently.

Instead of using the internals of the audio engine directly and bumping into various thread-safety issues, a high-level IPlayback facade with asynchronous API was introduced. IPlayback is the only entry point into the audio engine from a consumer point of view. 

To simplify the understanding of facade pattern you can also take a look on INotation interface, which is also a facade basically.

It's also worth to mention that all of these changes are an important part of the future of the mixer panel.

image

image

image

[full scheme available here](https://drive.google.com/file/d/1aeWXkifwIyjXgtcHVYC4yj9Nd1iBA-Sc/view?usp=sharing)

Important notes:

- These PRs are targeting for my fork, just because I want to divide the entire work into smaller PRs, so that it'll be a bit easier for review. But I assume that these PRs aren't going to  pass checks of our CI, just because they have to be merged into master simultaneously via the main PR.

- There is one important issue in the current master that remains unresolved - the playback loudness of some of the instruments is so low. It's happening because of the conversions from midiOld format into a new one. This particular issue would be fixed later


src/framework/audio/audiotypes.h
```diff
@@ -0,0 +1,62 @@
...
+using TrackId = int32_t;
+using TrackIdList = std::vector<TrackId>;
+using TrackName = std::string;
+
+using MixerChannelId = int32_t;
+
+using AudioSourceName = std::string;
+
+using FxProcessorId = std::string;
+using FxProcessorIdList =  std::vector<std::string>;
```
igorkorsukov, line 32: better `using FxProcessorIdList =  std::vector<FxProcessorId>;`

src/framework/audio/audiotypes.h
```diff
@@ -0,0 +1,62 @@
...
+
+struct AudioOutputParams {
+    FxProcessorIdList fxProcessors;
+    volume_db_t volume = 1.f;
+    balance_t balance = 0.f;
+    bool isMuted = false;
+
+    bool operator ==(const AudioOutputParams& other)
```
igorkorsukov, line 40: better `bool operator ==(const AudioOutputParams& other) const`

src/framework/audio/internal/worker/abstractaudiosource.h
```diff
@@ -32,11 +32,15 @@ class AbstractAudioSource : public IAudioSource
 
     virtual void setSampleRate(unsigned int sampleRate) override;
 
+    virtual bool isActive() const override;
+    virtual void setIsActive(bool arg) override;
```
igorkorsukov, line 36: better without `virtual`

src/framework/audio/internal/worker/audioengine.cpp
```diff
@@ -137,22 +119,20 @@ unsigned int AudioEngine::sampleRate() const
...
+IAudioBufferPtr AudioEngine::buffer() const
 {
     ONLY_AUDIO_WORKER_THREAD;
     return m_buffer;
 }
 
-std::shared_ptr<IMixer> AudioEngine::mixer() const
+IMixerPtr AudioEngine::mixer() const
 {
-    ONLY_AUDIO_WORKER_THREAD;
```
igorkorsukov, line 148: Why is this marker removed?
I would recommend creating the necessary markers and placing them as much as possible to identify errors

src/framework/audio/internal/worker/audioiohandler.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        if (s) {
+            s->audioIO()->setOutputParams(trackId, params);
+        }
+    }, AudioThread::ID);
+}
+
+Channel<TrackSequenceId, TrackId, AudioOutputParams> AudioIOHandler::outputParamsChanged() const
+{
+    return m_outputParamsChanged;
```
igorkorsukov, line 95: I believe we need to add a thread validation/indication marker:
* ONLY_AUDIO_WORKER_THREAD
* ONLY_AUDIO_MAIN_THREAD
* ONLY_AUDIO_MAIN_OR_WORKER_THREAD
* ANY_AUDIO__THREAD

src/framework/audio/internal/worker/itracksequence.h
```diff
@@ -0,0 +1,38 @@
+#ifndef MU_AUDIO_ITRACKSEQUENCE_H
```
igorkorsukov, line 1: You need to add a license, and in other files too

src/framework/audio/iplayback.h
```diff
@@ -19,36 +19,35 @@
...
-    virtual IMixerPtr mixer() const = 0;
-    virtual IAudioBufferPtr buffer() const = 0;
-    virtual void setAudioBuffer(IAudioBufferPtr) = 0;
+    virtual async::Promise<TrackSequenceId> addSequence() = 0;
+    virtual async::Promise<TrackSequenceIdList> sequenceIdList() const = 0;
+    virtual void removeSequence(const TrackSequenceId id) = 0;
+
+    virtual ITracksPtr tracks() const = 0;
+    virtual IPlayersPtr players() const = 0;
+    virtual IAudioIOPtr audioIO() const = 0;
```
igorkorsukov, line 48: I looked at how to use it, I believe this is not convenient and not clear. May be better to make one interface for all?

igorkorsukov: It is difficult to understand what needs to be done to use this interface and in what order
Is it true that I can add a track with the given input and output parameters - for this I need to use the `ITracks::addTrack` interface?
And I can change the in and out parameters - for this I need to use the `IAudioIO::setInputParams`,  `IAudioIO::setOutputParams` interface?

src/framework/audio/internal/worker/itracksequence.h
vpereverzev: damn, I've completely forgotten it, thanks

fixed in - https://github.com/musescore/MuseScore/pull/8471/commits/1077547adede978049c6c39844395683f679e9af

src/framework/audio/internal/worker/audioiohandler.cpp
vpereverzev: yeah, makes sense

added in https://github.com/musescore/MuseScore/pull/8471/commits/1f4270c0615b7641b18b9f43c8f0026b35f7392c

src/framework/audio/internal/worker/audioengine.cpp
vpereverzev: I don't actually recall that, maybe it was just accidentally removed

Anyway, nice catch. Fixed in https://github.com/musescore/MuseScore/pull/8471/commits/1f4270c0615b7641b18b9f43c8f0026b35f7392c

src/framework/audio/internal/worker/abstractaudiosource.h
vpereverzev: yep, thanks

fixed in https://github.com/musescore/MuseScore/pull/8471/commits/4b672c8c28daa3a027ee891548c119f9bc981bf6

src/framework/audio/audiotypes.h
vpereverzev: yeah, missed that one, thanks

fixed in https://github.com/musescore/MuseScore/pull/8471/commits/b2c24697aea9187e4f2ea98259412409cd09e325

src/framework/audio/audiotypes.h
vpereverzev: agreed

https://github.com/musescore/MuseScore/pull/8471/commits/b2c24697aea9187e4f2ea98259412409cd09e325

PR #8471, opened 2021-06-30
[MU4] audio_refactoring_main

vpereverzev:
The main PR with the audio refactoring

see an explanation at:
- https://github.com/musescore/MuseScore/pull/8467
- https://github.com/musescore/MuseScore/pull/8469
- https://github.com/musescore/MuseScore/pull/8470


src/framework/audio/iaudioio.h
```diff
@@ -0,0 +1,58 @@
...
+    virtual void setInputParams(const TrackSequenceId sequenceId, const TrackId trackId, const AudioInputParams& params) = 0;
+    virtual async::Channel<TrackSequenceId, TrackId, AudioInputParams> inputParamsChanged() const = 0;
+
+    virtual async::Promise<AudioOutputParams> outputParams(const TrackSequenceId sequenceId, const TrackId trackId) const = 0;
+    virtual void setOutputParams(const TrackSequenceId sequenceId, const TrackId trackId, const AudioOutputParams& params) = 0;
+    virtual async::Channel<TrackSequenceId, TrackId, AudioOutputParams> outputParamsChanged() const = 0;
+
+    virtual async::Promise<AudioOutputParams> globalOutputParams() const = 0;
+    virtual void setGlobalOutputParams(const AudioOutputParams& params) = 0;
+    virtual async::Channel<AudioOutputParams> globalOutputParamsChanged() const = 0;
```
igorkorsukov, line 49: better rename global to master

src/framework/audio/iplayback.h
```diff
@@ -19,36 +19,35 @@
...
-    virtual IAudioBufferPtr buffer() const = 0;
-    virtual void setAudioBuffer(IAudioBufferPtr) = 0;
+    virtual async::Promise<TrackSequenceId> addSequence() = 0;
+    virtual async::Promise<TrackSequenceIdList> sequenceIdList() const = 0;
+    virtual void removeSequence(const TrackSequenceId id) = 0;
+
+    virtual ITracksPtr tracks() const = 0;
+    virtual IPlayersPtr players() const = 0;
+    virtual IAudioIOPtr audioIO() const = 0;
 };
```
igorkorsukov, line 49: I think needs comments for this interface, like:
```
    // 1. Add Sequence
    virtual async::Promise<TrackSequenceId> addSequence() = 0;
    virtual async::Promise<TrackSequenceIdList> sequenceIdList() const = 0;
    virtual void removeSequence(const TrackSequenceId id) = 0;

    // 2. Setup tracks for Sequence
    virtual ITracksPtr tracks() const = 0;

    // 3. Play Sequence
    virtual IPlayersPtr players() const = 0;

    // 4. Output params
    virtual IAudioIOPtr audioIO() const = 0;
```

src/framework/audio/iplayers.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <memory>
+
+#include "async/promise.h"
+#include "async/channel.h"
+
+#include "audiotypes.h"
+
+namespace mu::audio {
+class IPlayers
```
igorkorsukov, line 34: I believe better rename to just `IPlayer`

src/playback/internal/playbackcontroller.cpp
```diff
@@ -215,58 +224,71 @@ void PlaybackController::togglePlay()
...
 }
 
 void PlaybackController::resume()
 {
-    sequencer()->play();
+    IF_ASSERT_FAILED(playback()) {
+        return;
+    }
+
+    playback()->players()->play(m_currentSequenceId);
```
igorkorsukov, line 288: I think we need some decisions to update midi data, like:
* reset buffer into midiaudiosource and request new data
* notify about new data through midistream
* explicit set new data for track here

src/framework/audio/iaudioio.h
vpereverzev: fixed

src/framework/audio/iplayback.h
vpereverzev: yep, done

src/framework/audio/iplayers.h
vpereverzev: renamed

src/playback/internal/playbackcontroller.cpp
vpereverzev: yes, I just realized that this is a little more complicated than I originally thought. We need to make sure that we don't accidentally re-create synthesizers / samplers because of this, and also we should only do this when we "resume" playback so that we don't do it twice on the first playback.

I will take care of this in the next PR

vpereverzev: I'm personally thinking about 1 and 2 options as the preferable

vpereverzev: turns out that it's actually not so scary, but I've found even more tiny issues during this fix :)

Anyway, I'd like to include them in another PRs, since this one is already gigantic

src/framework/audio/devtools/waveformmodel.h
```diff
@@ -0,0 +1,87 @@
...
+#include "async/asyncable.h"
+
+#include "iplayback.h"
+
+namespace mu::audio {
+class WaveFormModel : public QObject, public async::Asyncable
+{
+    Q_OBJECT
+
+    INJECT(audio, IPlayback, playback)
```
igorkorsukov, line 38: It may be worth adding to `IAudioIO` and IoC as well.
Otherwise, it is not clear what this class really depends on and what it can affect (for example, can it control play or add new tracks ...)

vpereverzev: Not sure, I think that would overlap with the idea of a single entry point. It might be confusing which one to use

vpereverzev: Instead, we might do that trick

image

so that you'd have to include "IPlayer" interface directly in the consumer class, in order to use it

vpereverzev: and obviously you'd need to include only what you actually need

src/framework/audio/iaudiosource.h
```diff
@@ -25,14 +25,22 @@
...
 namespace mu::audio {
 class IAudioSource
 {
 public:
     virtual ~IAudioSource() = default;
 
+    virtual bool isActive() const = 0;
+    virtual void setIsActive(bool arg) = 0;
+
+    virtual void seek(const msecs_t newPositionMsecs) { UNUSED(newPositionMsecs) }
```
igorkorsukov, line 42: I think better remove it, bad primer

vpereverzev: primer? What do you mean? Пример?))

src/framework/audio/internal/worker/audioengine.cpp
```diff
@@ -137,22 +119,24 @@ unsigned int AudioEngine::sampleRate() const
...
 }
 
-std::shared_ptr<ISequencer> AudioEngine::sequencer() const
+void AudioEngine::setMixer(IMixerPtr mixerPtr)
 {
-    ONLY_AUDIO_WORKER_THREAD;
-    return m_sequencer;
+    ONLY_AUDIO_MAIN_OR_WORKER_THREAD;
+
+    m_mixer = mixerPtr;
```
igorkorsukov, line 139: If it can be called from more than one thread, then synchronization is needed.
I would recommend making it so that it is called from only one thread. For example, objects were created and initialized in the thread in which they should be used

src/framework/audio/iaudiosource.h
igorkorsukov: image
:)

src/framework/audio/internal/worker/audioengine.cpp
vpereverzev: the only reason of that is the registration of IMixer for IOC, so that it'll be able in other audio classes via INJECT

vpereverzev: hold on, maybe I was wrong here with ONLY_AUDIO_MAIN...

vpereverzev: yeah, it could be just ONLY_AUDIO_WORKER_THREAD

I don't actually recall why I've not put it here

src/framework/audio/internal/worker/mixer.cpp
```diff
@@ -20,19 +20,27 @@
...
+using namespace mu::async;
 
 Mixer::Mixer()
 {
-    ONLY_AUDIO_WORKER_THREAD;
 }
 
 Mixer::~Mixer()
 {
-    ONLY_AUDIO_WORKER_THREAD;
```
igorkorsukov, line 35: I recommend keeping this and making it so that the mixer is created in the thread in which it should be used.

src/framework/audio/iaudiosource.h
vpereverzev: а я думаю, что еще за плохая грунтовка такая

src/framework/audio/internal/worker/mixer.h
```diff
@@ -24,60 +24,62 @@
...
+#include "iaudioconfiguration.h"
 #include "abstractaudiosource.h"
 #include "mixerchannel.h"
 #include "clock.h"
 
 namespace mu::audio {
-class Mixer : public IMixer, public AbstractAudioSource, public std::enable_shared_from_this<Mixer>
+class Mixer : public IMixer, public AbstractAudioSource, public std::enable_shared_from_this<Mixer>, public async::Asyncable
 {
+    INJECT(audio, IAudioConfiguration, config)
```
igorkorsukov, line 40: for a mixer and other similar classes that should be in an audio worker, I would recommend setting options explicitly, rather than using dependencies.
Otherwise, we need to make sure that AudioConfiguration (and other dependencies) are thread-safe.

src/framework/audio/internal/worker/playback.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+    m_playerHandlersPtr = std::make_shared<PlayerHandler>(this);
+    m_trackHandlersPtr = std::make_shared<TracksHandler>(this);
+    m_audioOutputPtr = std::make_shared<AudioOutputHandler>(this);
+}
+
+Promise<TrackSequenceId> Playback::addSequence()
+{
+    return Promise<TrackSequenceId>([this](Promise<TrackSequenceId>::Resolve resolve, Promise<TrackSequenceId>::Reject /*reject*/) {
+        TrackSequenceId newId = m_sequences.size();
```
igorkorsukov, line 51: I think it better to add check tread marker

vpereverzev: Ah, completely forgot this one, thanks

src/framework/midi/miditypes.h
```diff
@@ -32,111 +32,87 @@
...
-        }
-        return chunks.rbegin()->second.endTick;
-    }
+struct MidiStream {
+    tick_t lastTick = 0;
 
-    std::string dump(bool withEvents = false)
+    ValCh<std::vector<Event> > controlEventsStream;
+    async::Channel<Events, tick_t /*endTick*/> mainStream;
+    async::Channel<Events, tick_t /*endTick*/> backgroundStream;
```
igorkorsukov, line 92: what's the difference between `mainStream` and `backgroundStream`?

src/notation/internal/inotationmidievents.h
```diff
@@ -0,0 +1,50 @@
...
+{
+public:
+    virtual ~INotationMidiEvents() = default;
+
+    virtual void init() = 0;
+    virtual midi::Events retrieveEvents(const midi::channel_t midiChannel, const midi::tick_t fromTick,
+                                        const midi::tick_t toTick) const = 0;
+    virtual midi::Events retrieveEventsForElement(const Element* element, const midi::channel_t midiChannel) const = 0;
+    virtual std::vector<midi::Event> retrieveSetupEvents(const std::list<InstrumentChannel*> instrChannel) const = 0;
+};
```
igorkorsukov, line 45: Wouldn't it be better to put this in `INotationPlayback` instead of adding another interface?

igorkorsukov (approved):

src/framework/midi/miditypes.h
vpereverzev: ah, it was so hard to name it. And I'm still not happy with that

mainStream - is aimed to work with the general stream of events from track
the only purpose of the "background" one is to serve for "on-click" playback


I think in the nearest future I'll rename it into "auxillaryStream" or even remove it entirely, because it was added before the idea with "EventsBuffer", it might turn out that we can achieve the same result using the single stream of events in the latest solution.

src/notation/internal/inotationmidievents.h
vpereverzev: My personal goal was to separate all the stuff with midi-rendering into a single class, since it's going to be expanded in the nearest future

src/framework/audio/devtools/waveformmodel.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+    playback()->audioOutput()->masterSignalAmplitudeChanged().onReceive(this, [this](const audioch_t, const float amplitude) {
+        setCurrentSignalAmplitude(amplitude);
+    });
+
+    playback()->audioOutput()->masterVolumePressureChanged().onReceive(this, [this](const audioch_t, const volume_dbfs_t pressure) {
+        if (pressure < MIN_DISPLAYED_DBFS) {
+            setCurrentVolumePressure(MIN_DISPLAYED_DBFS);
+        } else if (pressure > MAX_DISPLAYED_DBFS) {
+            setCurrentVolumePressure(MIN_DISPLAYED_DBFS);
```
cbjeukendrup, line 41: Should this maybe be <code>setCurrentVolumePressure(**MAX**_DISPLAYED_DBFS);</code>?

vpereverzev: yep, I've also noticed this yesterday

It has been fixed in another branch

igorkorsukov:
This PR will still be viewed and it is possible there will be some more comments if something needs to be corrected - this will be corrected in other PRs

PR #8475, opened 2021-06-30
Fix some compiler warnings

cbjeukendrup:
NOTE: please review if the changes in the third commit will work correctly together with scorefile.cpp:279. 
I have a feeling that these changes _fix_ a bug there, but I'm not 100% sure. In MU3, there was no `isCmdFromInspector` parameter, and the code in scorefile.cpp is the same as in MU4, so that makes me assume that it was a bug in MU4.


cbjeukendrup:
VTests are failing because master branch can't build at the moment (but it's fixed with in this PR)

PR #8488, opened 2021-07-01
[MU3] Fix #322674: Change String Orchestra template to use multiple of each string instrument rather than being a string quintet

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/322674

This is for 3.x what #8489 is for master (merged already)


Jojo-Schmitz:
See #9000

PR #8491, opened 2021-07-01
Get rid of QWidgets from engraving

igorkorsukov:
QSvg depends on QWidgets, so I have hidden QSvgRender.

Also, added a sandbox project to check the dependencies for the engraving build.

### DIFF
--- a/src/engraving/interactive/messagebox.h
+++ b/src/engraving/interactive/messagebox.h
@@ -19,37 +19,31 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-#include "drawlogger.h"
+#ifndef MU_ENGRAVING_MESSAGEBOX_H
+#define MU_ENGRAVING_MESSAGEBOX_H
 
-#include "log.h"
+#include <string>
 
-using namespace mu::draw;
-
-static const std::string DRAW_OBJ_TAG("DRAW_OBJ");
-
-void DrawObjectsLogger::beginObject(const std::string& name, const PointF& pagePos)
-{
-    m_objects.push(name);
-    std::string gap;
-    gap.resize(m_objects.size());
-#ifdef LOG_STREAM
-    LOG_STREAM(haw::logger::Logger::DEBG, DRAW_OBJ_TAG) << "Begin: " << gap << name << "{" << pagePos.x() << "," << pagePos.y() << "}";
-#else
-    UNUSED(pagePos);
-#endif
-}
-
-void DrawObjectsLogger::endObject()
-{
-    IF_ASSERT_FAILED(!m_objects.empty()) {
-        return;
-    }
-
-    std::string gap;
-    gap.resize(m_objects.size());
-#ifdef LOG_STREAM
-    LOG_STREAM(haw::logger::Logger::DEBG, DRAW_OBJ_TAG) << "End:   " << gap << m_objects.top();
+#ifndef NO_ENGRAVING_INTERACTIVE
+#include "modularity/ioc.h"
+#include "iinteractive.h"
 #endif
 
-    m_objects.pop();
+namespace mu::engraving {
+class MessageBox
+{
+#ifndef NO_ENGRAVING_INTERACTIVE
+    INJECT_STATIC(engraving, framework::IInteractive, interactive)
+#endif
+public:
+
+    enum Button {
+        Ok,
+        Cancel
+    };
+
+    static Button warning(const std::string& title, const std::string& text);
+};
 }
+
+#endif // MU_ENGRAVING_MESSAGEBOX_H
--- a/src/engraving/libmscore/clef.cpp
+++ b/src/engraving/libmscore/clef.cpp
@@ -25,6 +25,8 @@
  Implementation of classes Clef (partial) and ClefList (complete).
 */
 
+#include <QCoreApplication>
+
 #include "clef.h"
 #include "measure.h"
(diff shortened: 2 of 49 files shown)


src/engraving/libmscore/vibrato.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include <cmath>
 
+#include <QGuiApplication>
```
vpereverzev, line 25: what is this for?

maybe we can include necessary missing types directly?

src/engraving/libmscore/textbase.cpp
```diff
@@ -21,6 +21,8 @@
  */
 
 #include <cmath>
+
+#include <QGuiApplication>
```
vpereverzev, line 25: Perhaps it might make sense to include it only here, I assume that this was needed in the rest of the files just because of include of "textbase"?

Most of them are classes derived/related of "textbase"

src/engraving/libmscore/vibrato.cpp
igorkorsukov: this is for `qApp->translation`
In the future should be replaced to use our `trc` function

src/engraving/libmscore/textbase.cpp
igorkorsukov: no, this is for qApp->translation
In the future should be replaced to use our `trc` function

src/engraving/libmscore/image.cpp
```diff
@@ -536,7 +543,7 @@ void Image::layout()
     setPos(0.0, 0.0);
     if (imageType == ImageType::SVG && !svgDoc) {
         if (_storeItem) {
-            svgDoc = new QSvgRenderer(_storeItem->buffer());
+            svgDoc = new SvgRenderer(_storeItem->buffer());
```
vpereverzev, line 546: doesn't it make sense to just allocate him once?

src/engraving/libmscore/image.cpp
```diff
@@ -77,7 +79,7 @@ Image::Image(const Image& img)
     if (imageType == ImageType::RASTER) {
         rasterDoc = img.rasterDoc ? new QImage(*img.rasterDoc) : 0;
     } else if (imageType == ImageType::SVG) {
-        svgDoc = img.svgDoc ? new QSvgRenderer(_storeItem->buffer()) : 0;
+        svgDoc = img.svgDoc ? new SvgRenderer(_storeItem->buffer()) : 0;
```
vpereverzev, line 82: so maybe we just need to add "setBuffer" method there, so we don't need to recreate it in "layout()" method every time

src/engraving/libmscore/clef.cpp
```diff
@@ -25,6 +25,8 @@
  Implementation of classes Clef (partial) and ClefList (complete).
 */
 
+#include <QCoreApplication>
```
vpereverzev, line 28: curious, why QCoreApplication is needed here

src/engraving/interactive/messagebox.h
```diff
@@ -0,0 +1,49 @@
...
+    INJECT_STATIC(engraving, framework::IInteractive, interactive)
+#endif
+public:
+
+    enum Button {
+        Ok,
+        Cancel
+    };
+
+    static Button warning(const std::string& title, const std::string& text);
```
vpereverzev, line 45: I think it makes sense to add an opportunity of configuring the amount of button roles here

it might be warning only with "ok" or whatever

src/engraving/libmscore/clef.cpp
igorkorsukov: in fact, in most places at the moment, we need `QCoreApplication`, because of translation.
But I did not waste time on this and just replaced `QApplication` with `QGuiApplication` with auto-replace. (because `QApplication` is `QWidget`)
Because in any case, we have to use our translation function and all this will be removed.

src/engraving/interactive/messagebox.h
igorkorsukov: I added this for backward compatibility because it's difficult to just remove it.
I really hope that we will not add new places where there will be interactivity with the user. And take away the ones that are (two places)

src/engraving/libmscore/score.h
```diff
@@ -837,7 +837,6 @@ class Score : public QObject, public ScoreElement
 
     void changeSelectedNotesVoice(int);
 
-    void colorItem(Element*);
```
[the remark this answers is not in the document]
igorkorsukov: yes, not using

src/engraving/libmscore/image.cpp
igorkorsukov: I think yes, needs refactoring for this class (by way, there is a memory leak too).
But I would not like to mix different tasks in one PR

src/engraving/libmscore/image.cpp
igorkorsukov: I think yes, needs refactoring for this class (by way, there is a memory leak too).
But I would not like to mix different tasks in one PR

src/engraving/interactive/messagebox.h
vpereverzev: gotchya

src/engraving/libmscore/clef.cpp
vpereverzev: OK, let's create a separate task then

vpereverzev: see #8500

PR #8493, opened 2021-07-02
Fix #270709: fix MusicXML export of Jump and Marker elements with custom labels

dmitrio95:
Resolves: https://musescore.org/en/node/270709

This pull request allows MusicXML export of Jump and Marker elements (Segno, To Coda etc.) with custom jump labels, except for the case when one "To Coda" marker should eventually jump to two different places (see also [my comment](https://musescore.org/en/node/270709#comment-1086598) in the issue). Jump labels in the exported MusicXML file now correspond to labels assigned to the respective Coda elements, which is why one test reference file got changed.

Since marker types with custom labels do not get saved to MSCX files (which is partially described in [this issue](https://musescore.org/en/node/291349), in the part concerning "To Coda" marker having changed to "Custom" type), MusicXML exporter has to guess a marker type from the actual usage of the marker's jump label. This logic is also implemented in this pull request so Marker elements with custom labels don't get ignored (if their type can be deduced).

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -344,6 +344,7 @@ class ExportMusicXml
     Hairpin const* hairpins[MAX_NUMBER_LEVEL];
     Ottava const* ottavas[MAX_NUMBER_LEVEL];
     Trill const* trills[MAX_NUMBER_LEVEL];
+    std::vector<const Jump*> _jumpElements;
     int div;
     double millimeters;
     int tenths;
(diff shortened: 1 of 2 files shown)


Jojo-Schmitz:
@iveshenry18 wouldn't that be something for 3.6.2_backend too?

RobFog:
cc @lvinken

iveshenry18:
> @iveshenry18 wouldn't that be something for 3.6.2_backend too?

Possibly! I'll see about cherry-picking it next week.

vpereverzev (approved):

Issue #8440, opened 2021-06-28
[MU4 Issue] UI slow to respond when selecting notes etc.


wizofaus:
**Describe the bug**
When selecting any part of the score containing notes, Mu4 is quite unresponsive, taking up to 700 or 800msec to repaint and show what's just been selected. Worse, when using palette to add items like "Staff Text" to multiple notes, it can take 10+ seconds to update.

**To Reproduce**
1. Create a new score
2. Add a note
3. Click on an empty measure - notice it shows as selected immediately
4. Click on a note - there's a noticeable delay before it shows as selected
5. Add 7 or 8 notes, select them and use palette to add "Staff Text" - can take 10+ seconds to update!

**Expected behavior**
UI should update immediately after selecting (or adding elements etc.)

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**
This is largely due to the selectionChanged callback installed by the InspectorListModel, and specifically the time it takes to do endInsertRows(), which can be over 700 ms.  I haven't figured out any way just marking the model as "dirty" at this point and only rebuilding it when necessary, but presumably it's possible.
Some of the other callbacks triggered when selecting a note a bit slow too, possibly including the one installed by NotationNoteInput but I didn't really investigate much further than that.

Visual Studio profiling output also confirms the slowness is in QAbstractItemModel::endInsertRows, and in particular when any of the sub-models are NotationSettingsProxyModel.

wizofaus:
BTW this can be improved at least for the case of adding to a large number of selected elements from the palette by ensuring we don't trigger the selectionChanged event for each added element. Essentially we can just trigger it once at the end, similar to how I changed it for dropChanged.

igorkorsukov:
@DmitryArefiev 
* Add 7 or 8 notes, select them and use palette to add "Staff Text" - can take 10+ seconds to update!  - This is fixed in the PR https://github.com/musescore/MuseScore/pull/9854

Issue about selection needs actualization, I can't reproduce it.

DmitryArefiev:
Tested on master build (Win10, Mac11, LinuxMint 20) - CAN'T REPRODUCE NOW

PR #8495, opened 2021-07-02
[MU4] part fix for #8440 - speed up adding from palette to multiple notes/rests

wizofaus:
Resolves: Only partly resolves https://github.com/musescore/MuseScore/issues/8440

Updating the inspector panel is very slow (don't know how to fix that sorry, it's a Qt/Qml issue), but it's far worse that it tries to update it for every single element created as a result of applying a palette element to multiple selected notes/rests.


src/notation/internal/notationinteraction.cpp
```diff
@@ -492,7 +493,11 @@ void NotationInteraction::select(const std::vector<Element*>& elements, SelectTy
     for (Element* element: elements) {
         score()->select(element, type, staffIndex);
     }
+}
 
+void NotationInteraction::select(const std::vector<Element*>& elements, SelectType type, int staffIndex)
```
wizofaus, line 498: Need to preserve this for external calls to public INotationInteraction::select.

src/notation/internal/notationinteraction.cpp
```diff
@@ -1394,7 +1399,7 @@ void NotationInteraction::applyDropPaletteElement(Ms::Score* score, Ms::Element*
         }
 
         if (el && !score->inputState().noteEntryMode()) {
-            select({ el }, Ms::SelectType::SINGLE, 0);
+            selectInternal({ el }, Ms::SelectType::SINGLE, 0);
```
wizofaus, line 1402: Having this fire the selectionChanged event was the main culprit

src/notation/internal/notationinteraction.cpp
```diff
@@ -481,7 +482,7 @@ void NotationInteraction::moveChordNoteSelection(MoveDirection d)
     notifyAboutSelectionChanged();
 }
 
-void NotationInteraction::select(const std::vector<Element*>& elements, SelectType type, int staffIndex)
+void NotationInteraction::selectInternal(const std::vector<Element*>& elements, SelectType type, int staffIndex)
```
igorkorsukov, line 485: Just a suggestion. I usually in such cases call `doSomeThing` - that is, direct action. In this case `doSelect`. Like this in many places in the code.

RomanPudashkin (approved):

PR #8501, opened 2021-07-02
Removed use QClipboard in the engraving

igorkorsukov:


src/engraving/libmscore/lyrics.cpp
```diff
@@ -461,9 +454,7 @@ void Lyrics::paste(EditData& ed)
     }
 
     score()->endCmd();
-    txt = sl.join(" ");
```
vpereverzev, line 464: do you have any ideas what is happening here?

I mean we've read from the clipboard and then have set text into clipboard again? :)

igorkorsukov: so I also thought, did not come up with the point of this, so I decided to just remove, and not figure out how to do it differently.
if we need something like that, then we'll do it (although i doubt)

igorkorsukov: I also thought that if I do: 
1. Ctrl+V 
2. Ctrl+C 
3. Ctrl+C -> different result

This is not what the user expects, it convinced me to remove this code

vpereverzev (approved):

PR #8503, opened 2021-07-02
Replaced qApp->translation to qtrc

igorkorsukov:
I would be clean up translations contexts, but not easy, so if it doing, then need separate pr

### DIFF
--- a/src/engraving/libmscore/barline.h
+++ b/src/engraving/libmscore/barline.h
@@ -23,8 +23,6 @@
 #ifndef __BARLINE_H__
 #define __BARLINE_H__
 
-#include <QGuiApplication>
-
 #include "element.h"
 #include "mscore.h"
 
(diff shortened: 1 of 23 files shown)


src/engraving/libmscore/cmd.cpp
```diff
@@ -26,6 +26,10 @@
 */
 
 #include <assert.h>
+#include <QGuiApplication>
```
igorkorsukov, line 29: needs for `qApp->keyboardModifiers()`

src/engraving/libmscore/mscore.cpp
```diff
@@ -21,8 +21,11 @@
  */
 
 #include <QDir>
+#include <QCoreApplication>
```
igorkorsukov, line 24: needs for `QCoreApplication::applicationDirPath()`

src/engraving/libmscore/barline.h
```diff
@@ -144,7 +142,6 @@ class BarLine final : public Element
     static BarLineType barLineType(const QString&);
 
     int subtype() const override { return int(_barLineType); }
-    QString subtypeName() const override { return qApp->translate("barline", barLineTypeName().toUtf8()); }
```
vpereverzev, line 147: Has it not been used?

vpereverzev (approved):

src/engraving/libmscore/barline.h
igorkorsukov: yep

Issue #8498, opened 2021-07-02
Fretboard Diagrams are using the Wrong UI Colour & Do Not Refresh on the score


Tantacrul:
**Describe the bug**
Fretboard diagrams are not using the Font Primary colour, which causes them to look incorrect in Dark mode / HC mode.

Changes made in the inspector do not refresh on the score immediately

See video
https://www.dropbox.com/s/cn1efmngpaz1iv4/Fretboard_Diagrams_Wrong_Colour_%26_Do_Not_Refresh_on_The_Score.mov?dl=0

Tantacrul:
@Howard-C - I thought you might like this one.

Howard-C:
I created #8504, but I only (partly) solved the first issue, plus another issue that isn't mentioned here. The part I didn't solve is that when I change the theme, it's gonna need a click on the diagram to change its colour (but after that it will be fine), not sure why it needs that though.

Eism:
fixed on #8504

PR #8504, opened 2021-07-02
Fix some issues of the fret diagram inspector

Howard-C:
1) Draw the fret diagram on the inspector using `fontPrimaryColor` (from #8498)
2) Make the hover event on the fret diagram work again

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/notation/fretdiagrams/FretDiagramPopup.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/notation/fretdiagrams/FretDiagramPopup.qml
@@ -72,6 +72,7 @@ StyledPopupView {
                 isMultipleDotsModeOn: root.model ? root.model.isMultipleDotsModeOn : false
                 currentFretDotType: root.model ? root.model.currentFretDotType : false
                 visible: root.model ? root.model.areSettingsAvailable : false
+                color: ui.theme.fontPrimaryColor
 
                 width: parent.width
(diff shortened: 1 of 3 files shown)


Tantacrul:
Excellent work on the UI.

The second problem is clearly an interaction issue. Observing the changes you make in realtime, help you to recognise how the interaction works, and removes all doubt,

If you do not get an immediate update, the natural assumption is that something isn't working. All Inspector interactions should take effect immediately unless their is an 'apply' button (which there isn't at present)

Howard-C:
I understand, but I'm afraid the issue is not with the inspector, because this part of the code is largely copied from 3.x, where it works fine, and my test does show a change to how scores are _supposed_ to be layout, so I guess it's a problem with how or when the score is refreshed, from other parts of the new 4.0 code. I'll try more later.

Eism:
@Howard-C 
1. To update the data immediately in the notation, you should notify about data changing in the notation(globalContext()->currentNotation()->notationChanged().notify();  in mousePressEvent method)
Also please replace the calling m_diagram->score()->startCmd(); to calling globalContext()->currentNotation()->undoStack()->prepareChanges();
It needs for undo/redo changing

2. To update the color immediately, add an update method to setColor method
The color is set correctly, but QQuickPaintedItem does not know about these changes for repainting

Howard-C:
@Eism Done! Thanks for the pointers.

Eism (approved):

PR #8510, opened 2021-07-03
Allow pasting all spanners 

wizofaus:
Resolves: https://musescore.org/en/node/2861

The behavior when cutting/copying lines (spanners) to the clipboard and pasting was very inconsistent for no good reason - basically only hairpins worked as expected. There are still some points for discussion here and some things that don't quite work as well as they could but it's a vast improvement on the current arbitrary limitations. Largely this involved getting rid of a lot of special case logic, though some had to be re-instated for slurs and note-anchored lines.

### DIFF
--- a/src/engraving/libmscore/chordrest.cpp
+++ b/src/engraving/libmscore/chordrest.cpp
@@ -639,17 +639,15 @@ Element* ChordRest::drop(EditData& data)
     }
     break;
 
-    case ElementType::HAIRPIN:
-    {
-        Hairpin* hairpin = toHairpin(e);
-        hairpin->setTick(tick());
-        hairpin->setTrack(track());
-        hairpin->setTrack2(track());
-        score()->undoAddElement(hairpin);
-    }
-        return e;
-
     default:
+        if (e->isSpanner()) {
+            Spanner* spanner = toSpanner(e);
+            spanner->setTick(tick());
+            spanner->setTrack(track());
+            spanner->setTrack2(track());
+            score()->undoAddElement(spanner);
+            return e;
+        }
         qDebug("cannot drop %s", e->name());
         delete e;
         return 0;
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -1981,10 +1981,10 @@ void Score::addNoteLine()
     TextLine* line = new TextLine(this);
     line->setParent(firstNote);
     line->setStartElement(firstNote);
-    line->setEndElement(lastNote);
     line->setDiagonal(true);
     line->setAnchor(Spanner::Anchor::NOTE);
     line->setTick(firstNote->chord()->tick());
+    line->setEndElement(lastNote);
 
     undoAddElement(line);
 }
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -1854,8 +1854,6 @@ bool Note::acceptDrop(EditData& data) const
            || (type == ElementType::KEYSIG)
            || (type == ElementType::TIMESIG)
            || (type == ElementType::BAR_LINE)
-           || (type == ElementType::SLUR)
-           || (type == ElementType::HAIRPIN)
            || (type == ElementType::STAFF_TEXT)
            || (type == ElementType::SYSTEM_TEXT)
            || (type == ElementType::STICKING)
@@ -1864,7 +1862,8 @@ bool Note::acceptDrop(EditData& data) const
            || (type == ElementType::TREMOLOBAR)
            || (type == ElementType::FRET_DIAGRAM)
            || (type == ElementType::FIGURED_BASS)
-           || (type == ElementType::LYRICS);
+           || (type == ElementType::LYRICS)
+           || e->isSpanner();
 }
 
 //---------------------------------------------------------
@@ -1899,15 +1898,6 @@ Element* Note::drop(EditData& data)
             delete e;
         }
         return 0;
-
-    case ElementType::SLUR:
-        data.view()->addSlur(chord(), nullptr, toSlur(e));
-        delete e;
-        return 0;
-
-    case ElementType::HAIRPIN:
-        // forward this event to a chord
-        return chord()->drop(data);
 
     case ElementType::LYRICS:
         e->setParent(ch);
@@ -2089,6 +2079,17 @@ Element* Note::drop(EditData& data)
     break;
 
     default:
+        Spanner* spanner;
+        if (e->isSpanner() && (spanner = toSpanner(e))->anchor() == Spanner::Anchor::NOTE) {
+            spanner->setParent(this);
+            spanner->setStartElement(this);
+            spanner->setTick(tick());
+            spanner->setTrack(track());
+            spanner->setTrack2(track());
+            spanner->computeEndElement();
+            score()->undoAddElement(spanner);
+            return e;
+        }
         return ch->drop(data);
     }
     return 0;
--- a/src/engraving/libmscore/rest.cpp
+++ b/src/engraving/libmscore/rest.cpp
@@ -176,7 +176,6 @@ bool Rest::acceptDrop(EditData& data) const
         || (type == ElementType::STAFF_STATE)
         || (type == ElementType::INSTRUMENT_CHANGE)
         || (type == ElementType::DYNAMIC)
-        || (type == ElementType::HAIRPIN)
         || (type == ElementType::HARMONY)
         || (type == ElementType::TEMPO_TEXT)
         || (type == ElementType::REHEARSAL_MARK)
@@ -188,7 +187,7 @@ bool Rest::acceptDrop(EditData& data) const
         ) {
         return true;
     }
-    return false;
+    return type != ElementType::SLUR && e->isSpanner();
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/slur.cpp
+++ b/src/engraving/libmscore/slur.cpp
@@ -120,7 +120,7 @@ bool SlurSegment::edit(EditData& ed)
 {
     Slur* sl = slur();
 
-    if (ed.key == Qt::Key_X) {
+    if (ed.key == Qt::Key_X && !ed.modifiers) {
         sl->undoChangeProperty(Pid::SLUR_DIRECTION, QVariant::fromValue<Direction>(sl->up() ? Direction::DOWN : Direction::UP));
         sl->layout();
         return true;
@@ -494,6 +494,12 @@ bool SlurSegment::isEdited() const
         }
     }
     return false;
+}
+
+Slur::Slur(const Slur& s)
+    : SlurTie(s)
+{
+    _sourceStemArrangement = s._sourceStemArrangement;
 }
 
 //---------------------------------------------------------
@@ -979,6 +985,16 @@ Slur::Slur(Score* s)
 }
 
 //---------------------------------------------------------
+//   calcStemArrangement
+//---------------------------------------------------------
+
+int calcStemArrangement(Element* start, Element* end)
+{
+    return (start && toChord(start)->stem() && toChord(start)->stem()->up() ? 2 : 0)
+           + (end && end->isChord() && toChord(end)->stem() && toChord(end)->stem()->up() ? 4 : 0);
+}
+
+//---------------------------------------------------------
 //   write
 //---------------------------------------------------------
 
@@ -992,8 +1008,25 @@ void Slur::write(XmlWriter& xml) const
         return;
     }
     xml.stag(this);
+    if (xml.clipboardmode()) {
+        xml.tag("stemArr", calcStemArrangement(startElement(), endElement()));
+    }
     SlurTie::writeProperties(xml);
     xml.etag();
+}
+
+//---------------------------------------------------------
+//   readProperties
+//---------------------------------------------------------
+
+bool Slur::readProperties(XmlReader& e)
+{
+    const QStringRef& tag(e.name());
+    if (tag == "stemArr") {
+        _sourceStemArrangement = e.readInt();
+        return true;
+    }
+    return SlurTie::readProperties(e);
 }
 
 //---------------------------------------------------------
@@ -1090,6 +1123,15 @@ SpannerSegment* Slur::layoutSystem(System* system)
             }
             Chord* c1 = startCR()->isChord() ? toChord(startCR()) : 0;
             Chord* c2 = endCR()->isChord() ? toChord(endCR()) : 0;
+
+            if (_sourceStemArrangement != -1) {
+                if (_sourceStemArrangement != calcStemArrangement(c1, c2)) {
+                    // copy & paste from incompatible stem arrangement, so reset bezier points
+                    for (int g = 0; g < (int)Ms::Grip::GRIPS; ++g) {
+                        slurSegment->ups((Ms::Grip)g) = UP();
+                    }
+                }
+            }
 
             if (c1 && c1->beam() && c1->beam()->cross()) {
                 // TODO: stem direction is not finalized, so we cannot use it here
--- a/src/engraving/libmscore/slur.h
+++ b/src/engraving/libmscore/slur.h
@@ -65,14 +65,17 @@ public:
 class Slur final : public SlurTie
 {
     void slurPosChord(SlurPos*);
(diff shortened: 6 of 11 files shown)


src/engraving/libmscore/chordrest.cpp
```diff
@@ -408,6 +409,25 @@ void ChordRest::undoSetSmall(bool val)
     undoChangeProperty(Pid::SMALL, val);
 }
 
+//---------------------------------------------------------
+//   trimLength - shorten a spanner so it ends on a note/chord
+//---------------------------------------------------------
```
wizofaus, line 414: need to update comment to reflect function name

src/engraving/libmscore/chordrest.cpp
```diff
@@ -408,6 +409,25 @@ void ChordRest::undoSetSmall(bool val)
...
+    while (chordRest = Ms::nextChordRest(chordRest, true)) {
+        auto crTicks = chordRest->ticks();
+        if (!chordRest->isRest()) {
+            trimTo += crTicks;
+        }
+        if ((length -= crTicks) <= Fraction()) {
+            break;
+        };
+    }
+    spanner->setTicks(trimTo);
```
wizofaus, line 428: should probably look further to see what existing functions (similar to findCR etc.) might do this

src/engraving/libmscore/chordrest.cpp
```diff
@@ -639,17 +659,18 @@ Element* ChordRest::drop(EditData& data)
     }
     break;
 
-    case ElementType::HAIRPIN:
-    {
-        Hairpin* hairpin = toHairpin(e);
-        hairpin->setTick(tick());
-        hairpin->setTrack(track());
-        hairpin->setTrack2(track());
```
wizofaus, line 647: basically this behaviour makes sense for ALL spanners, no need for it to be a special case for hairpins

src/engraving/libmscore/note.cpp
```diff
@@ -1900,15 +1899,12 @@ Element* Note::drop(EditData& data)
...
-    case ElementType::HAIRPIN:
-        // forward this event to a chord
-        return chord()->drop(data);
-
+    /*
+case ElementType::SLUR:
+    data.view()->addSlur(chord(), nullptr, toSlur(e));
+    delete e;
+    return 0;
+    */
```
wizofaus, line 1907: remove

src/engraving/libmscore/note.cpp
```diff
@@ -2089,6 +2085,17 @@ Element* Note::drop(EditData& data)
     break;
 
     default:
+        Spanner* spanner;
+        if (e->isSpanner() && (spanner = toSpanner(e))->anchor() == Spanner::Anchor::NOTE) {
+            spanner->setParent(this);
```
wizofaus, line 2090: couldn't find a nice way to avoid this one special case - note-anchored lines need to know exactly which note they're being pasted on to, and computeEndElement() needs to be called after setTrack2().
However it could be simplified if setTrack2( ) just triggered the call to computeEndElement() automatically...

src/engraving/libmscore/rest.cpp
```diff
@@ -188,7 +187,7 @@ bool Rest::acceptDrop(EditData& data) const
         ) {
         return true;
     }
-    return false;
+    return type != ElementType::SLUR && e->isSpanner();
```
wizofaus, line 190: for now, keep existing behaviour that we don't allow dropping slurs on rests, even though it actually works fine...

src/engraving/libmscore/slur.cpp
```diff
@@ -120,7 +120,7 @@ bool SlurSegment::edit(EditData& ed)
 {
     Slur* sl = slur();
 
-    if (ed.key == Qt::Key_X) {
+    if (ed.key == Qt::Key_X && !ed.modifiers) {
```
wizofaus, line 123: (this fix is already in another PR but was making testing a pain!)

src/engraving/libmscore/slurtie.cpp
```diff
@@ -460,7 +460,7 @@ SlurTie::~SlurTie()
 
 void SlurTie::writeProperties(XmlWriter& xml) const
 {
-    Element::writeProperties(xml);
+    Spanner::writeProperties(xml);
```
wizofaus, line 463: critical - ensures that the "ticks" property is correctly preserved when creating mimeData

src/engraving/libmscore/spanner.cpp
```diff
@@ -757,6 +757,14 @@ void Spanner::computeEndElement()
 
     case Anchor::CHORD:
     case Anchor::NOTE:
+        if (!_endElement) {
+            ChordRest* cr = score()->findCR(tick2(), track2() / VOICES);
+            if (cr && _anchor == Anchor::NOTE && cr->isChord()) {
+                _endElement = toChord(cr)->upNote();
```
wizofaus, line 763: when pasting a note-anchored line, find the "most" likely candidate for attaching the other end to, though this could definitely be improved

src/engraving/libmscore/spanner.cpp
```diff
@@ -1069,6 +1077,9 @@ void Spanner::setEndElement(Element* e)
     }
 #endif
     _endElement = e;
+    if (e && _tick > Fraction() && ticks() == Fraction()) {
+        setTicks(e->tick() - _tick);
```
wizofaus, line 1081: only needed for note anchored lines, which never get their `ticks` property updated properly when creating or when loading

src/engraving/libmscore/edit.cpp
```diff
@@ -1981,10 +1981,10 @@ void Score::addNoteLine()
     TextLine* line = new TextLine(this);
     line->setParent(firstNote);
     line->setStartElement(firstNote);
-    line->setEndElement(lastNote);
     line->setDiagonal(true);
     line->setAnchor(Spanner::Anchor::NOTE);
     line->setTick(firstNote->chord()->tick());
+    line->setEndElement(lastNote);
```
wizofaus, line 1987: makes this consistent with the order used when loading from a file

src/engraving/libmscore/slur.cpp
```diff
@@ -978,6 +987,16 @@ Slur::Slur(Score* s)
     setAnchor(Anchor::CHORD);
 }
 
+//---------------------------------------------------------
+//   calcStemArrangement
+//---------------------------------------------------------
```
wizofaus, line 992: when copying/pasting slurs around the score, we need to know if the stem arrangement is compatible, or the bezier points don't make sense

src/engraving/libmscore/slur.cpp
```diff
@@ -1091,6 +1127,15 @@ SpannerSegment* Slur::layoutSystem(System* system)
             Chord* c1 = startCR()->isChord() ? toChord(startCR()) : 0;
             Chord* c2 = endCR()->isChord() ? toChord(endCR()) : 0;
 
+            if (_sourceStemArrangement != -1) {
+                if (_sourceStemArrangement != calcStemArrangement(c1, c2)) {
+                    // copy & paste from incompatible stem arrangement, so reset bezier points
+                    for (int g = 0; g < (int)Ms::Grip::GRIPS; ++g) {
+                        slurSegment->ups((Ms::Grip)g) = UP();
+                    }
```
wizofaus, line 1135: BTW currently this is only done for slur direction=AUTO (vs UP/DOWN), but might make sense to do in all cases.

src/engraving/libmscore/slur.h
```diff
@@ -65,14 +65,17 @@ class SlurSegment final : public SlurTieSegment
 class Slur final : public SlurTie
 {
     void slurPosChord(SlurPos*);
+    int _sourceStemArrangement = -1;
```
wizofaus, line 68: transient property needed to remember stem arrangement of a slur copied to the clipboad

src/engraving/libmscore/slurtie.cpp
```diff
@@ -501,11 +501,7 @@ bool SlurTie::readProperties(XmlReader& e)
 
 void SlurTie::read(XmlReader& e)
 {
-    while (e.readNextStartElement()) {
-        if (!SlurTie::readProperties(e)) {
-            e.unknown();
-        }
-    }
+    Spanner::read(e);
```
wizofaus, line 504: Probably should just remove this override entirely, there was no need for it and it meant `Slur::readProperties` was never getting called.

wizofaus:
The work done here will probably impact what needs to be done to restore the CTRL+SHIFT+drag functionality too (https://github.com/musescore/MuseScore/issues/8512). It still needs more testing, unfortunately because the Inspector is still not fully implemented for many element types, it's hard to test quite a lot of things thoroughly.

src/engraving/libmscore/spanner.cpp
```diff
@@ -385,6 +385,11 @@ Spanner::Spanner(const Spanner& s)
     _tick         = s._tick;
     _ticks        = s._ticks;
     _track2       = s._track2;
+    if (!s.startElement() && !spannerSegments().size()) {
```
wizofaus, line 388: logically I'd expect spanner segments to always be cloned, but it was causing some issues for me, hence for now I've only put it in for when the element is cloned when pasting from clipboard

src/engraving/libmscore/textbase.cpp
```diff
@@ -1668,11 +1668,10 @@ TextBase::TextBase(Score* s, ElementFlags f)
 TextBase::TextBase(const TextBase& st)
     : Element(st)
 {
-    _cursor                      = st._cursor;
+    _cursor                      = new TextCursor(this);
     _text                        = st._text;
-    _layout                      = st._layout;
     textInvalid                  = st.textInvalid;
```
wizofaus, line 1673: copying _layout was a big issue, as it meant copying a lot of raw pointers

src/engraving/libmscore/textbase.cpp
```diff
@@ -1694,6 +1693,12 @@ TextBase::TextBase(const TextBase& st)
         _propertyFlagsList[i] = st._propertyFlagsList[i];
     }
     _links = 0;
+    createLayout();
+}
+
+TextBase::~TextBase()
+{
+    delete _cursor;
```
wizofaus, line 1701: existing memory leak

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -56,8 +56,8 @@ TextLineBaseSegment::TextLineBaseSegment(Spanner* sp, Score* score, ElementFlags
 TextLineBaseSegment::TextLineBaseSegment(const TextLineBaseSegment& seg)
     : LineSegment(seg)
 {
-    _text    = new Text(*seg._text);
-    _endText = new Text(*seg._endText);
+    _text    = seg._text->clone();
+    _endText = seg._endText->clone();
```
wizofaus, line 60: doesn't really change anything, but in future might if _text/_endText could point to objects derived from Text, and more obvious what's being done

src/engraving/libmscore/textbase.cpp
```diff
@@ -1668,10 +1668,10 @@ TextBase::TextBase(Score* s, ElementFlags f)
 TextBase::TextBase(const TextBase& st)
     : Element(st)
 {
-    _cursor                      = st._cursor;
+    _cursor                      = new TextCursor(this);
```
wizofaus, line 1671: Fixes a critical bug where the TextCursor._text pointer can be left pointed to deleted memory (happens now when copying text elements)

src/engraving/libmscore/slur.cpp
```diff
@@ -978,6 +984,16 @@ Slur::Slur(Score* s)
     setAnchor(Anchor::CHORD);
 }
 
+//---------------------------------------------------------
+//   calcStemArrangement
+//---------------------------------------------------------
+
+int calcStemArrangement(Element* start, Element* end)
```
igorkorsukov, line 991: better to add static (or anonymous namespace)

igorkorsukov (approved):

Jojo-Schmitz:
Not for MU3 but for MU4. Might get backported to MU3 though

Jojo-Schmitz:
Just FYI: this PR backported to 3.x cause an mtest failure, see https://github.com/Jojo-Schmitz/MuseScore/runs/3171043582?check_suite_focus=true#step:5:747
```
35/62 Test #34: tst_selectionfilter ..............***Failed    4.28 sec
--- /home/runner/work/MuseScore/MuseScore/mtest/libmscore/selectionfilter/selectionfilter6-base-ref.xml	2021-07-27 11:19:38.597387743 +0000
+++ selectionfilter6-base.xml	2021-07-27 11:47:21.765384310 +0000
@@ -8,6 +8,8 @@
       <durationType>quarter</durationType>
       <Spanner type="Slur">
         <Slur>
+          <stemArr>0</stemArr>
+          <ticks_f>1/4</ticks_f>
           </Slur>
         <next>
           <location>
@@ -38,6 +40,8 @@
       <durationType>quarter</durationType>
       <Spanner type="Slur">
         <Slur>
+          <stemArr>0</stemArr>
+          <ticks_f>1/4</ticks_f>
           </Slur>
         <next>
           <location>
   <diff -u /home/runner/work/MuseScore/MuseScore/mtest/libmscore/selectionfilter/selectionfilter6-base-ref.xml selectionfilter6-base.xml failed

```
Niot sure why it doesn't in master though?

wizofaus:
Not sure but I will say stemArr isn't supposed to be serialised into an mscx file, only in clipboard mode. The ticks_f is probably right though,  does the slur *look* right?

Jojo-Schmitz:
It is not in the mscx, just in the clipboard file. 
And yes, it look good im the score

wizofaus:
Good to hear, well I would say that getting those tests working again is pretty important, I'm not happy that there's no workable way for me to run/modify the unit tests currently, at least as far as I've tried.

wizofaus:
The test (TestSelectionFilter::filterSlur) is definitely part of the test-suite in the master branch, I just don't see much evidence it's being run at all.
But I can't even seem to generate a .vcxproj file for these tests, even with BUILD_UNIT_TESTS ON in CMakeLists.txt.
I'm thinking of trying to come up with another method of running these tests so I can easily/quick  launch & debug them in MSVC, but it may require manually adding calls to the tests to another source file (given you can't readily use reflection in C++ to find the test functions and run them automatically, which I guess is why it uses Qt slots)

wizofaus:
Actually I just force-regenerated the sln/vcxproj files and now I do have a engraving_tests project, and it even builds and runs!
There's a lot of errors running the tests though, it seems to be writing read the xml files from the wrong location.

wizofaus:
However when you specify a specific test function to run as a command line argument, QTest::exec barfs because it's not in the first test object in the suite.  Have got a fix for that, however we probably want to think about having a way of specifying all tests in a suite. Also having qtestsuite.h as all inline functions that are then included by all the test source files seems odd, it just means heaps of rebuilding every time I need to adjust something, and there's no need for that sort of optimisation running unit tests.  Surely we can have a .cpp file or lib that's used by all the test projects?  Or have a cut down  .h file that has just what the individual test classes need (to register themselves with the suite).  Anyway, the filterSlur() test runs now and does indeed generate the diff @Jojo-Schmitz pointed out, should be easy to fix.

Jojo-Schmitz:
As far as I can tell this PR causes a cerash when pasting a tie onto a rest. Easy fix:
```
index 31adff8cf..78c2c65f3 100644
--- a/libmscore/rest.cpp
+++ b/libmscore/rest.cpp
@@ -232,7 +232,7 @@ bool Rest::acceptDrop(EditData& data) const
          ) {
             return true;
             }
-      return type != ElementType::SLUR && e->isSpanner();
+      return type != ElementType::SLUR && type != ElementType::TIE && e->isSpanner();
       }
```

wizofaus:
Good catch (though slurs can usefully start on rests and MuseScore supports it, just change the note to a rest after adding slur). Need to start adding more unit tests for that sort of thing.

Jojo-Schmitz:
As said, ties are even more useless here.
Would we want to allow pastting/dropping slurs on rests then?

wizofaus:
Not sure what consequences it would have if you have a list/range selection that includes both notes/rests. The existing behaviour of only slurring from the first note to the last (per staff) should be preserved. Either way your fix needs to go into MU4.

Jojo-Schmitz:
pasting slurs onto rests indeed does work! Question is whether we want to allow for it

PR #8190, opened 2021-05-28
[MU3 Backend] ENG-28: Combine articulations on import

iveshenry18:
Resolves: [ENG-28](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-28): Articulations should be combined where possible

This commit adds handling of combined articulations such as
accent staccatos and marcato tenutos on both XML import and export.
On import, pairs of articulations are combined where possible; on
export, these combined articulation symbols are separated into their
component MusicXML elements. This also adds a test for these cases,
and updates a test that previously didn't take this effect into account.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8514, opened 2021-07-04
Fix #302826: MusicXML export of compound articulations

dmitrio95:
Resolves: https://musescore.org/en/node/302826

Allows MusicXML export of those compound articulations that do not have direct MusicXML equivalent. This is done by exporting them as multiple single articulations, like it is proposed in the issue description.

Since there are no changes in import of such articulations (multiple articulations are not merged to compound ones here), a test case can be added after other MusicXML export tests are re-enabled (see #8513).

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -2620,6 +2620,48 @@ static void fermatas(const QVector<Element*>& cra, XmlWriter& xml, Notations& no
 }
 
 //---------------------------------------------------------
+//   splitCompoundArticulations
+//---------------------------------------------------------
+
+static std::vector<SymId> splitCompoundArticulations(const SymId sid)
+{
+#define ARTIC_2(compound, single1, single2) \
+case SymId::artic##compound##Above: { \
+    return { SymId::artic##single1##Above, SymId::artic##single2##Above }; \
+} \
+case SymId::artic##compound##Below: { \
+    return { SymId::artic##single1##Below, SymId::artic##single2##Below }; \
+}
+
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
@iveshenry18 wouldn't that be something for 3.6.2_backend too?

Jojo-Schmitz:
It does collide with  #8190 though

dmitrio95:
Yes, indeed it does, and that fix also covers merging articulations on import. Maybe it's better to cherry-pick that fix to master then? It would just be needed to add few more compound articulation types (types combined with `SoftAccent` are not present in #8190).

iveshenry18:
@Jojo-Schmitz, we may not need this for 3.6.2_backend, since that branch is primarily concerned with importing. However, for porting to master, I do think cherry-picking #8190 and adding any compound articulations I may have missed would be a good strategy

Jojo-Schmitz:
Like adding these to your PR?

            case SymId::articStaccatoSoftAccentAbove:
            case SymId::articStaccatoSoftAccentBelow:
                  return { "staccato", "soft-accent" };
                  break;
                  
            case SymId::articSoftAccentTenutoAbove:
            case SymId::articSoftAccentTenutoBelow:
                  return { "soft-accent", "tenuto" };
                  break;


            case SymId::articSoftAccentTenutoStaccatoAbove:
            case SymId::articSoftAccentTenutoStaccatoBelow:
                  return { "staccato", "soft-accent", "tenuto" };
                  break;

iveshenry18:
Yep!

dmitrio95:
OK, then I guess this pull request is not needed. I didn't know this had already been fixed in another branch.

Jojo-Schmitz:
Would now also fix #16748

PR #8515, opened 2021-07-04
[MU3] Fix #322759: Inputing Ottava Bassa via menu or shortcut shows it above staff

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/322759

PR for master in #8516 (merged already)

PR #8516, opened 2021-07-04
[MU4] Fix #322759: Inputing Ottava Bassa via menu or shortcut shows it above staff

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/322759

This is for master what #8515 is for 3.x

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -1869,6 +1869,10 @@ void Score::cmdAddOttava(OttavaType type)
             }
             Ottava* ottava = new Ottava(this);
             ottava->setOttavaType(type);
+            if (type == OttavaType::OTTAVA_8VB /*|| type == OttavaType::OTTAVA_15MB || type == OttavaType::OTTAVA_22MB*/) {
+                ottava->setPlacement(Placement::BELOW);
+                ottava->styleChanged();
+            }
             ottava->setTrack(cr1->track());
             ottava->setTrack2(cr1->track());
             ottava->setTick(cr1->tick());
@@ -1888,7 +1892,10 @@ void Score::cmdAddOttava(OttavaType type)
 
         Ottava* ottava = new Ottava(this);
         ottava->setOttavaType(type);
-
+        if (type == OttavaType::OTTAVA_8VB /*|| type == OttavaType::OTTAVA_15MB || type == OttavaType::OTTAVA_22MB*/) {
+            ottava->setPlacement(Placement::BELOW);
+            ottava->styleChanged();
+        }
         ottava->setTrack(cr1->track());
         ottava->setTrack2(cr1->track());
         ottava->setTick(cr1->tick());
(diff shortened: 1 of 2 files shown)


src/engraving/libmscore/edit.cpp
```diff
@@ -1869,6 +1869,10 @@ void Score::cmdAddOttava(OttavaType type)
             }
             Ottava* ottava = new Ottava(this);
             ottava->setOttavaType(type);
+            if (type == OttavaType::OTTAVA_8VB/* || type == OttavaType::OTTAVA_15MB || type == OttavaType::OTTAVA_22MB*/) {
```
Jojo-Schmitz, line 1872: No need to check for `OttavaType::OTTAVA_15MB` and `OttavaType::OTTAVA_22MB`, as there's no shortcut for those

src/engraving/libmscore/edit.cpp
```diff
@@ -1888,7 +1892,10 @@ void Score::cmdAddOttava(OttavaType type)
 
         Ottava* ottava = new Ottava(this);
         ottava->setOttavaType(type);
-
+        if (type == OttavaType::OTTAVA_8VB/* || type == OttavaType::OTTAVA_15MB || type == OttavaType::OTTAVA_22MB*/) {
```
Jojo-Schmitz, line 1895: same

src/palette/internal/palette/palettecreator.cpp
```diff
@@ -1423,6 +1423,7 @@ PalettePanel* PaletteCreator::newLinesPalettePanel()
 
     ottava = makeElement<Ottava>(gscore);
     ottava->setOttavaType(OttavaType::OTTAVA_22MB);
+    ottava->setPlacement(Placement::BELOW);
```
Jojo-Schmitz, line 1426: May not be needed, certainly not for the issue at hand, but I found this to be pretty inconsistent, at least.

src/engraving/libmscore/edit.cpp
wizofaus: But logically the user should be able to define such shortcuts ( and you probably can by editing files directly), so I'd think we should support that anyway?
More importantly, shouldn't this logic belong inside Ottava itself? (i.e. setOttavaType( ) should set the "placement" property as appropriate)

src/engraving/libmscore/edit.cpp
wizofaus: Code duplication, again, would make sense to me to keep inside Ottava::setOttavaType

Jojo-Schmitz: That duplication is not my fault though, and I don't plan to fix that, not for this issue.

src/engraving/libmscore/edit.cpp
Jojo-Schmitz: I don't think you can just invent new shortcuts without code changes. Directly editing the shortcuts.xml won't work, as far as I can tell.
Yes. that decision, above/below, might better be taken elsewhere, something for another PT maybe, here I just tried to keep the changes as minimal as possible

Jojo-Schmitz: Hmm, maybe that `cmdAddOttava()` can get used from whithin plugins, and then with arguments other than just `OttavaType::OTTAVA_8VA` and `OttavaType::OTTAVA_8VB`.

Jojo-Schmitz: Edit: no, It can't, only the  "add-8va" and "add-8vb" from cmd.cpp

PR #8523, opened 2021-07-05
Added workspace name check

igorkorsukov:


src/framework/global/io/path.cpp
```diff
@@ -115,6 +115,58 @@ mu::io::path mu::io::dirpath(const mu::io::path& path)
...
+bool mu::io::isAllowedFileName(const path& fn_)
+{
+    QString fn = basename(fn_).toQString();
+
+    // Windows filenames are not case sensitive.
+    fn = fn.toUpper();
+
+    static const QString illegal="<>:\"|?*";
+
+    foreach (const QChar& c, fn) {
```
Eism, line 127: better to use for instead of foreach

PR #8308, opened 2021-06-14
Allow popups to be opened for editable notation elements

sidharth-anand:
image

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [ ] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8524, opened 2021-07-05
Tempo Popup

sidharth-anand:
## Features

* The popup is fully functional and supports user input through all 3 tabs.
* The tempo marking is also modified to prevent user input inside the equation.

https://user-images.githubusercontent.com/55060749/124446441-0b5abf80-dd9e-11eb-8dab-32e27afdd125.mp4

Note: This PR depends on #8308 


src/engraving/libmscore/tempotext.cpp
```diff
@@ -473,6 +603,88 @@ void TempoText::layout()
...
+    float baseDuration = 1.0f;
+
+    for (auto duration : durationMap) {
+        if (marking[0] == duration.first) {
+            baseDuration = duration.second;
+            break;
+        }
+    }
+
+    const int dots = marking.count(".");
```
Jojo-Schmitz, line 683: either make this float

src/engraving/libmscore/tempotext.cpp
```diff
@@ -473,6 +603,88 @@ void TempoText::layout()
...
+
+    return baseDuration * std::powf(1.5, dots);
```
Jojo-Schmitz, line 685: or cast it to float here, in order to get it to cimpile on Linux, like in GitHib's CI

wizofaus: `std::pow(3, dots)/std::pow(2, dots)` would arguably be more efficient

Tantacrul:
I have a few thoughts for improvements:

### Insertion point

1. The first thing is the insertion point (for text editing). It looks weird as a rectangle that doesn't flash on an off. For that reason, it looks like an unintended artefact, like an uppercase 'i'. It would make a lot more sense if it would flash on and off.
image

2. When the user clicks on a text field in the popup, the insertion point in the tempo marking should disappear. Then if the user single-clicks back on the tempo marking again (because we're still in an 'edit' mode) the insertion should re-appear. This would help the user understand what is in focus.
image

Tantacrul:
### Visuals


1. If you trigger the popup and then move the score around a little, then open the popup again, you'll find that it gets misaligned pretty easily. It would be great if it would always be centred to the position of the tempo marking.
image

2. The vertical positioning is a bit weird too. If the tempo marking appears in the middle of the screen, then the popup looks OK. However, if the tempo marking is towards the top of my screen, the popup appears to be too far away. If I scroll so the tempo marking appears to be at the bottom of my screen, then the popup appears to the right of it, which looks weirder.

Top of my screen:
image


Bottom of screen:
image

I would suggest the popup is a good bit closer, like this:
image

And if there is not enough visible room under the tempo marking, then I think the idea of keeping it to the right makes sense and I would remove the upwards pointing arrow and keep it closer to the marking itself.
image

Tantacrul:
### Spacing of popup

Just a few little things here and there:

1. There is too little padding between the tabs and the content underneath. Generally, there is a 12px gap between sections, like this:
image
2. This is particularly bad with the 'Code' tab, where the title directly under it is overlapping the same space. Here's a visual correction I've made to give you an idea of how it should look
image

Tantacrul:
### Other visual issues with Popup

1. I've noticed that whole wholenote and double-wholenote symbols are weird looking in the context menu.
image

I've created two new symbols in the MuseScoreIcon font (not Leland), which you should use instead for the context menu items instead. These are specifically for when these values are displayed vertically. The unicodes are F3D3 & F3D4 respectively.

image

Tantacrul:
### Wrong icons used for Whole Note and Double Whole note in metronome marking.

image


These look to be too small and are pointing at Bravura. The correct symbols are found in LelandText: ECA2 and ECA0.
image

@oktophonie can confirm

Tantacrul:
### Wrong visuals for selected tab state

Currently, the text in a selected / unselected tabs are inverted. A selected tab should be bolder, less opaque. Basically, you just need to swap the stylings around.

This is how it looks now:
image

This is how it should look (see Figma for reference)
image

Tantacrul:
### Bug

'Visibility' sometimes requires being toggled twice for it to work.

1. Create tempo marking
2. Double click to get the popup
3. Type something in the main tempo text field (like Andante, etc.)
4. Press the 'visibility' toggle in the popup (to off)
5. Observe it does not work
6. Turn it on and off again
7. Now it works

Tantacrul:
Last comment: I don't think a user should be able to press the backspace button to delete portions of the text that are controlled by the popup.

sidharth-anand:
>     1. If you trigger the popup and then move the score around a little, then open the popup again, you'll find that it gets misaligned pretty easily. It would be great if it would always be centred to the position of the tempo marking.

I'm not able to reproduce this. The popup seems to be centered to the element irrespective of whether i move the score, or scroll

I've fixed most of the other issues and will be updating the PR soon

Jojo-Schmitz:
A rebase is needed too

Tantacrul:
Wow. This is a big improvement!

One bug I noticed about 'visibility': If you put text before and after a tempo marking (see image) and then turn the visibility off and back on again, it forgets the position of the tempo marking and puts it at the end.

Before:
image

After:
image

Tantacrul:
I'm also curious about what problems you are facing with the whole / double whole note symbols.

image

Just in case, here's the latest version of the font (make sure to uninstall any versions on your machine, btw)
https://www.dropbox.com/s/ip59ren10u69hr7/MusescoreIcon.ttf?dl=0

Unicodes are F3D3 & F3F4

Tantacrul:
One last thing:

Some of the metronome marks look wrong (often far too small).

For example, compare this 32nd, which is correct:
image

With this 64th, which is far too small (and defaulting to Bravura, which is the backup, implying it is pointing at the wrong font):
image

The same goes for the whole note and double whole note:
image

I'm going to guess that you are telling it to look in 'Leland', whereas I believe they should be looking at 'Leland Text'.

Do make sure to use the latest version of both Leland and Leland text, incidentally.
https://github.com/MuseScoreFonts/Leland

The best person to speak to about all of the icon stuff is definitely @oktophonie

Jojo-Schmitz:
Small code style issue. Not sure why the unit tests fail.

src/engraving/libmscore/property.cpp
```diff
@@ -142,7 +142,7 @@ static constexpr PropertyMetaData propertyList[] = {
     { Pid::SCALE,                   false, "scale",                 P_TYPE::SCALE,          DUMMY_QT_TR_NOOP("propertyName", "scale") },
     { Pid::LOCK_ASPECT_RATIO,       false, "lockAspectRatio",       P_TYPE::BOOL,           DUMMY_QT_TR_NOOP("propertyName", "aspect ratio locked") },
     { Pid::SIZE_IS_SPATIUM,         false, "sizeIsSpatium",         P_TYPE::BOOL,           DUMMY_QT_TR_NOOP("propertyName", "size is spatium") },
-    { Pid::TEXT,                    true,  "text",                  P_TYPE::STRING,         DUMMY_QT_TR_NOOP("propertyName", "text") },
+    { Pid::TEXT,                    true,  "text",                  P_TYPE::STRING,         DUMMY_QT_TR_NOOP("propertyName", "text") },
```
wizofaus, line 145: did uncrustify pick this up?

src/engraving/libmscore/property.cpp
```diff
@@ -316,6 +318,7 @@ static constexpr PropertyMetaData propertyList[] = {
     { Pid::SIZE_SPATIUM_DEPENDENT,  false, "sizeIsSpatiumDependent",P_TYPE::BOOL,           DUMMY_QT_TR_NOOP("propertyName", "spatium dependent font") },
     { Pid::ALIGN,                   false, "align",                 P_TYPE::ALIGN,          DUMMY_QT_TR_NOOP("propertyName", "align") },
     { Pid::TEXT_SCRIPT_ALIGN,       false, "align",                 P_TYPE::INT,            DUMMY_QT_TR_NOOP("propertyName", "text script align") },
+    { Pid::TEXT_HIDE_CURSOR,        true,  "hideTextCursor",        P_TYPE::BOOL,           DUMMY_QT_TR_NOOP("propertyName", "hideTextCursor") },
```
wizofaus, line 321: Why would that be a property of an element?

src/engraving/libmscore/property.h
```diff
@@ -156,6 +156,8 @@ enum class Pid {
     SPACE,            // used for spacer
     TEMPO,
     TEMPO_FOLLOW_TEXT,
+    TEMPO_EQUATION,
+    TEMPO_EQUATION_VISIBLE,
```
wizofaus, line 160: I'd be concerned about inserting new values into this enum, in case there's any code anywhere that relies on the existing values (esp. with serialization). Adding at the end is always safer.

src/engraving/libmscore/textbase.cpp
```diff
@@ -3084,6 +3092,17 @@ Sid TextBase::offsetSid() const
     return Sid::NOSTYLE;
 }
 
+bool TextBase::moveCursor(TextCursor* cursor, int key, bool ctrlPressed, TextCursor::MoveMode moveMode) const
```
wizofaus, line 3095: But this already works? (using Ctrl+left/right to move words).  Any reason you needed to add this?

src/engraving/libmscore/textbase.cpp
```diff
@@ -3365,7 +3384,8 @@ void TextBase::drawEditMode(mu::draw::Painter* p, EditData& ed)
     p->setPen(pen);
 
     // Don't draw cursor if there is a selection
-    if (!cursor->hasSelection()) {
+    LOGD() << _hideCursor;
+    if (!cursor->hasSelection() && !_hideCursor) {
```
wizofaus, line 3388: don't think we need that log message - what's an example of why this would be wanted on a per-element basis?

src/engraving/libmscore/property.cpp
Jojo-Schmitz: no

src/engraving/libmscore/textedit.cpp
```diff
@@ -321,6 +321,10 @@ bool TextBase::edit(EditData& ed)
             return true;
 
         case Qt::Key_Delete:
+            if (!toTextBase(ed.element)->canDelete(cursor, ed.key)) {
```
wizofaus, line 324: toTextBase(ed.element) should always be the same as `this`, no?

src/engraving/libmscore/textedit.cpp
```diff
@@ -357,21 +365,14 @@ bool TextBase::edit(EditData& ed)
...
-                                       mm) && type() == ElementType::LYRICS) {
-                return false;
-            }
-            s.clear();
-            break;
-
         case Qt::Key_Right:
-            if (!_cursor->movePosition(ctrlPressed ? TextCursor::MoveOperation::NextWord : TextCursor::MoveOperation::Right,
-                                       mm) && type() == ElementType::LYRICS) {
+            if (!toTextBase(ed.element)->moveCursor(_cursor, ed.key, ctrlPressed, mm) && type() == ElementType::LYRICS) {
```
wizofaus, line 369: moveCursor( ) shouldn't be what's responsible for mapping keys to directions, IMO

src/engraving/libmscore/tempotext.cpp
```diff
@@ -413,12 +551,26 @@ bool TempoText::setProperty(Pid propertyId, const QVariant& v)
...
+        if (isEquationValid(v.toString())) {
+            _equation = v.toString();
+            parseEquation();
+        }
+        break;
+    case Pid::TEMPO_EQUATION_VISIBLE:
+        _isEquationVisible = v.toBool();
+
+        if (!_isEquationVisible)
+            _lastEquationIndex = equationIndices().first;
```
Jojo-Schmitz, line 570: Here the code style check fails. braces needed

fonts/leland/FONTLOG.txt
```diff
@@ -11,7 +11,7 @@ Basic font information
 
 Leland is a Unicode typeface designed by MuseScore BVBA for its music notation and scoring applications.
 
-It is compliant with version 1.2 of the Standard Music Font Layout (SMuFL), a community-driven standard for how music symbols should be laid out in the Unicode Private Use Area (PUA) in the Basic Multilingual Plane (BMP) for compatibility between different scoring applications.
+It is compliant with version 1.4 of the Standard Music Font Layout (SMuFL), a community-driven standard for how music symbols should be laid out in the Unicode Private Use Area (PUA) in the Basic Multilingual Plane (BMP) for compatibility between different scoring applications.
```
Jojo-Schmitz, line 14: Tthis is part of #8853

src/engraving/libmscore/tempotext.h
```diff
@@ -68,21 +81,40 @@ class TempoText final : public TextBase
...
     void updateRelative();
 
+    void setEquationFromTempo(int tempo);
+
+    void parseEquation();
+
+    bool moveCursor(TextCursor* cursor, int key, bool ctrlPressed, TextCursor::MoveMode moveMode) const override;
+    bool canDelete(TextCursor* cursor, int key) const override;
+
+    void dragTo(EditData& ed) override;
```
wizofaus, line 97: do all of these need to be public? Particularly moveCursor.

fonts/leland/FONTLOG.txt
```diff
@@ -32,6 +32,47 @@ https://musescore.com/contact
...
+    - Metronome mark double whole note, whole note, 32nd note (ECA0, ECA2, ECAB)
+    - Standard accidentals for chord symbols (ED60-ED66)
+  - Modified:
+    - Metronome mark notes (ECA3, ECA5, ECA7, ECA9) and dot (ECB7)
+- Added metadata for glyphsWithAlternates and ligatures
+- Added metadata for preferred text fonts, in engravingDefaults
+
+04 February 2021 (Martin Keary & Simon Smith) Leland Version 0.52
+- Packaged with MuseScore 3.6.2
+- Added kerning for fz pairs in dynamics
```
Jojo-Schmitz, line 75: this too is part of #8853

fonts/leland/metadata.json
```diff
@@ -22,23 +22,27 @@
         "stemThickness": 0.11,
         "subBracketThickness": 0.11,
         "textEnclosureThickness": 0.11,
+        "textFontFamily": [
+            "Edwin",
+            "serif"
+        ],
```
Jojo-Schmitz, line 28: this too is part of #8853, but rnamed leland_metadata.json (on sync with the leland repo)

src/notation/view/notationpaintview.cpp
```diff
@@ -324,6 +324,27 @@ void NotationPaintView::hideContextMenu()
...
+    }
+
+    const RectF& bbox = hitElement->canvasBoundingRect();
+    PointF canvasPos = hitElement->canvasPos();
+
+    QString type(hitElement->name());
+    PointF elementPos = fromLogical(canvasPos.toQPoint());
+    RectF size(fromLogical(QRect(0, 0, bbox.width(), bbox.height())));
+
+    emit openElementPopupRequested(type, elementPos.toQPoint(), QPoint(size.width(), size.height()));
```
wizofaus, line 340: The intention is for this to exactly cover the element?

fonts/leland/FONTLOG.txt
Jojo-Schmitz: Actually even newer, Leland 0.61

src/engraving/libmscore/tempotext.cpp
```diff
@@ -89,77 +97,85 @@ void TempoText::read(XmlReader& e)
...
-    TempoPattern("\uECAB",                     1.0 / 480.0,  TDuration::DurationType::V_32ND),       // 1/32
-    TempoPattern("\uECA1",                     1.0 / 7.5,    TDuration::DurationType::V_BREVE),      // longa
-    TempoPattern("\uECA0",                     1.0 / 7.5,    TDuration::DurationType::V_BREVE),      // double whole
-    TempoPattern("\uECAD",                     1.0 / 960.0,  TDuration::DurationType::V_64TH),       // 1/64
-    TempoPattern("\uECAF",                     1.0 / 1920.0, TDuration::DurationType::V_128TH),      // 1/128
-    TempoPattern("\uECB1",                     1.0 / 3840.0, TDuration::DurationType::V_256TH),      // 1/256
-    TempoPattern("\uECB3",                     1.0 / 7680.0, TDuration::DurationType::V_512TH),      // 1/512
-    TempoPattern("\uECB5",                     1.0 / 15360.0, TDuration::DurationType::V_1024TH),     // 1/1024
+static const EquationDescrption durationMap[] = {
+    { "d",  8.0f,   "\uECA0", "<sym>metNoteDoubleWhole</sym>",  TDuration::DurationType::V_BREVE },
```
wizofaus, line 147: why do you need the unicode code AND the xml - one can be determined from the other.
And it's not clear why you need <sym> </sym> in the table given they all have it and could be added programmatically as needed.

src/engraving/libmscore/tempotext.cpp
```diff
@@ -308,51 +446,41 @@ void TempoText::undoChangeProperty(Pid id, const QVariant& v, PropertyFlags ps)
...
-                if (match2.hasMatch()) {
-                    _relative = pa2.f / pa.f;
-                    _isRelative = true;
-                    updateRelative();
-                    updateScore();
-                    return;
-                }
+    static const float quarterNotePlayback = 1.f / 60.f;
+
+    static const QRegularExpression bpmExpression("^[\\[,\\(]?[ ]*(?<note>[dwhqest].*) *= *(?<bpm>\\d+)[\\],\\)]?$");
```
wizofaus, line 451: "dwhqest" occurs 8 times in this file and I would think is moderately likely to need modifying in future, better to use a named constant.

Jojo-Schmitz:
It does fail unit tests, 
```
The following tests FAILED:
	  5 - engraving_tests (Failed)
	 12 - iex_musicxml_tests (Failed)
```
Those need to get looked after

src/engraving/libmscore/property.cpp
Jojo-Schmitz: it probably should have though

src/engraving/libmscore/textbase.cpp
sidharth-anand: This is done so the individual elements can modify the movement behavior of the cursor if required. (TempoText does not allow the cursor to move inside the tempo marking itself)

wizofaus: Are you saying somebody can't, for example, edit the = symbol and replace it with ≈? That would be a serious regression on MU3 if so...

sidharth-anand: @Tantacrul should probably answer this

wizofaus: As long as I can have this: 
image

src/engraving/libmscore/property.cpp
Jojo-Schmitz: @sidharth-anand please remove that stzray space at the end of the line

src/importexport/ove/internal/importove.cpp
```diff
@@ -1044,11 +1093,8 @@ TDuration OveNoteType_To_Duration(ovebase::NoteType noteType)
 //      break;
 //  }
```
Jojo-Schmitz, line 1094: these 8 lines should get removed

src/palette/internal/palettecreator.cpp
```diff
@@ -1105,75 +1105,48 @@ PalettePtr PaletteCreator::newTempoPalette(bool defaultPalette)
...
-                     QT_TRANSLATE_NOOP("palette", "Eighth note = eighth note metric modulation"),
-                     1.0 / 1.0, true, false, true, false, false),
-        TempoPattern("<sym>metNoteQuarterUp</sym> = <sym>metNoteQuarterUp</sym>",
-                     QT_TRANSLATE_NOOP("palette", "Quarter note = quarter note metric modulation"),
-                     1.0 / 1.0, true, false, true, false, false),
-        TempoPattern("<sym>metNote8thUp</sym><sym>space</sym><sym>metAugmentationDot</sym> = <sym>metNoteQuarterUp</sym>",
-                     QT_TRANSLATE_NOOP("palette", "Dotted eighth note = quarter note metric modulation"),
-                     2.0 / 3.0, true, false, true, false, false),
+    static const TempoPattern tempoPatterns[] = {
+        TempoPattern("", "h = 80", "Half notes = 80 BPM"),
```
Jojo-Schmitz, line 1121: We're loosing ther translatability (of tooltips IIRC)

src/importexport/ove/internal/importove.cpp
Jojo-Schmitz: Or get changed to match the comments above, `//    return "w";`

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledPopup.qml
```diff
@@ -33,17 +33,18 @@ Popup {
 
     property Item anchorItem: null
     property real arrowX: Boolean(anchorItem) ? anchorItem.x - x + anchorItem.width / 2 : width / 2
+    property alias arrowHeight: arrow.height
 
     property alias navigation: navPanel
     property bool isDoActiveParentOnClose: true
 
+
```
Jojo-Schmitz, line 41: excessive whitespace

bkunda:
> > ```
> > 1. If you trigger the popup and then move the score around a little, then open the popup again, you'll find that it gets misaligned pretty easily. It would be great if it would always be centred to the position of the tempo marking.
> > ```
> 
> 
> I'm not able to reproduce this. The popup seems to be centered to the element irrespective of whether i move the score, or scroll
> 
> 
> I've fixed most of the other issues and will be updating the PR soon

bkunda:
And apologies - did not mean to close this issue! 🤦‍♂️

Jojo-Schmitz:
rebase needed

bkunda:
**Tempo text without a bpm value**
An interesting issue here!
This observation relates to tempo text containing no bpm value (e.g. "Allegretto" or "Vivace", as opposed to "Allegro♩=120").

See this video for a full explainer: https://www.dropbox.com/s/2regyildnhthwxm/no-bpm-text-issue.mov?dl=0

In summary: 
- If making any adjustments to the bpm value of the tempo text is to trigger visibility of the bpm value itself (as it currently does), then the visibility checkbox should automatically be selected (which currently doesn't happen)
- I did not find the relationship between the visibility checkbox and the tempo ("bpm") section to be sufficiently clear, and might suggest a slight tweaking of the checkbox's placement:

Screenshot 2021-09-06 at 21 52 35

(Note: I've also suggested a slightly more descriptive label (image on the right), which _might_ just make the function that bit clearer).

@Tantacrul let me know your thoughts on this. I realise it's your design baby! 😬

src/engraving/tests/keysig_data/keysig.mscx
```diff
@@ -69,7 +69,11 @@
             <sigD>4</sigD>
             </TimeSig>
           <Tempo>
-            <tempo>1.66667</tempo>
+            <playbackTempo>1.66667</playbackTempo>
+            <notatedTempo>100</notatedTempo>
+            <equation>q = 100</equation>
+            <equationVisible>1</equationVisible>
+            <followText>1</followText>
             <text>𝅘𝅥 = 100</text>
```
shoogle, line 77: Is there a reason for keeping the metronome marking in the `<text>` element now that you have the `<equation>` element?

Also, I originally suggested this for the MSCX syntax:

```xml
    <text>Some text <equation>q = 100</equation> some more text</text>
```

This allows text after the equation. Is there a reason you chose not to do it this way?

Tantacrul:
> @Tantacrul let me know your thoughts on this. I realise it's your design baby! 😬

Please! Go to town and don't worry about me. I agree with your suggestions

bkunda:
> > @Tantacrul let me know your thoughts on this. I realise it's your design baby! 😬
> 
> Please! Go to town and don't worry about me. I agree with your suggestions

All good then! 😌

bkunda:
**A (minor) spacing issue**

By default, a space should be inserted between the descriptive text ("Allegro") and the tempo/bpm text.

Current default spacing:
Screenshot 2021-09-06 at 22 51 00

Desired default spacing:
Screenshot 2021-09-06 at 22 51 11

Incidentally, I was able to create the second image above using the spacebar. Noting @Tantacrul's comment above re: the backspace key, I might suggest that, for consistency's sake, the spacebar also be rendered unusable.... 

...OR...

They **both** be rendered **usable**. The problem here is that, I can currently add spaces _between_ elements, but I cannot _remove_ them. If I accidentally, or even intentionally, add more spaces than I end up wanting, I can't remove them without deleting the tempo text and starting again. 

A further argument in favour of keeping keyboard keys usable: I currently can enter a line space using the return key, in order to generate this (often desirable) result: 
Screenshot 2021-09-06 at 23 02 59

It is a problem for me that some keys work while other keys don't. 

My vote would be to reinstate the backspace key functionality, particularly as it allows the user to more liberally modify the text imported from the palette (For example, adding the word "Molto" before something like "Vivace"; It feels intuitive to be able to make modifications like this directly in the tempo text itself, and having the standard keyboard keys like backspace and spacebar at your disposal is an important part of this experience).

bkunda:
> **A (minor) spacing issue**
> 
> By default, a space should be inserted between the descriptive text ("Allegro") and the tempo/bpm text.
> 
> Current default spacing:
> Screenshot 2021-09-06 at 22 51 00
> 
> Desired default spacing:
> Screenshot 2021-09-06 at 22 51 11
> 
> Incidentally, I was able to create the second image above using the spacebar. Noting @Tantacrul's comment above re: the backspace key, I might suggest that, for consistency's sake, the spacebar also be rendered unusable....
> 
> ...OR...
> 
> They **both** be rendered **usable**. The problem here is that, I can currently add spaces _between_ elements, but I cannot _remove_ them. If I accidentally, or even intentionally, add more spaces than I end up wanting, I can't remove them without deleting the tempo text and starting again.
> 
> A further argument in favour of keeping keyboard keys usable: I currently can enter a line space using the return key, in order to generate this (often desirable) result:
> Screenshot 2021-09-06 at 23 02 59
> 
> It is a problem for me that some keys work while other keys don't.
> 
> My vote would be to reinstate the backspace key functionality, particularly as it allows the user to more liberally modify the text imported from the palette (For example, adding the word "Molto" before something like "Vivace"; It feels intuitive to be able to make modifications like this directly in the tempo text itself, and having the standard keyboard keys like backspace and spacebar at your disposal is an important part of this experience).

Please also note a bug to fix: tempo text font is changing unexpectedly (as in the third image of this comment) after a few goes at making modifications in the popup.

bkunda:
Hi again 😀
Just some further (minor) visual fixes to do.
Alignment is currently off-centre for most elements (in MacOS build), and there are some incorrect sizes set for the text box and equal sign:
Screenshot 2021-09-07 at 18 18 00
I believe the Figma file should have the correct spacing, and @jessjwilliamson can also help if there are any questions about exactly how many pixels should be where.

Jojo-Schmitz:
rebase needed

bkunda:
This PR has been inactive for a long time and would need to be recreated from scratch to be mergeable today, so unfortunately it will now need to be closed.

PR #8532, opened 2021-07-06
[MU4] Fix #8522 - ensure when cloning textbase we don't clone raw pointers

### LINKED ISSUE #8522: [MU4 Issue] Crash after editing copied text element
**Describe the bug**
Trying to edit a text element created via copy/paste causes MU4 to crash

**To Reproduce**
1. Create a new score
2. Add a text element
3. Select just the text element (single element selection), copy to clipboard
4. Select a rest in a subsequent measure, paste it
5. Double-click on the newly pasted text element
6. MuseScore crashes

**Expected behavior**
Shouldn't crash!

**Screenshots**

Windows 10, built with MSVC

**Additional context**
Due to the TextBase copy constructor copying the TextCursor from a temporary element, such that the TextCursor still has a pointer to that element (which then gets destructed). Fixed in https://github.com/musescore/MuseScore/pull/8510 but that introduces many other changes and needs more testing.

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issue/8522

Copying TextBase elements was leaving TextCursor pointing to wrong object

### DIFF
--- a/src/engraving/libmscore/textbase.cpp
+++ b/src/engraving/libmscore/textbase.cpp
@@ -1668,7 +1668,7 @@ TextBase::TextBase(Score* s, ElementFlags f)
 TextBase::TextBase(const TextBase& st)
     : Element(st)
 {
-    _cursor                      = st._cursor;
+    _cursor                      = new TextCursor(this);
     _text                        = st._text;
     _layout                      = st._layout;
     textInvalid                  = st.textInvalid;


wizofaus:
My other PR that has this fix is getting pretty big and still has some outstanding issues, but this should be fixed either way.

src/engraving/libmscore/textbase.cpp
```diff
@@ -1668,7 +1668,7 @@ TextBase::TextBase(Score* s, ElementFlags f)
 TextBase::TextBase(const TextBase& st)
     : Element(st)
 {
-    _cursor                      = st._cursor;
+    _cursor                      = new TextCursor(this);
```
wizofaus, line 1671: NB this is actually a memory-leak, but one that already exists when creating new TextBase elements.  As far as I can tell it's completely safe to delete in the destructor, but perhaps we should use unique_ptr<> ?

igorkorsukov:
merged by https://github.com/musescore/MuseScore/pull/8510/files#r663629890

PR #8551, opened 2021-07-06
[MU4] Inspector Cleanup. Replaced the checkboxes with the ‘eye’ buttons

Eism:
image

image

### DIFF
--- a/src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
+++ b/src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
@@ -63,6 +63,7 @@ Rectangle {
                     { textRole: "StyledPopupView", componentRole: styledPopupViewComponent },
                     { textRole: "StyledMenu", componentRole: styledMenuComponent },
                     { textRole: "CheckBox", componentRole: checkBoxSample },
+                    { textRole: "VisibilityBox", componentRole: visibilityBoxSample },
                     { textRole: "ColorPicker", componentRole: colorPickerSample },
                     { textRole: "ExpandableBlank", componentRole: expandableBlankSample },
                     { textRole: "FlatButton", componentRole: flatButtonSample },
@@ -328,6 +329,24 @@ Rectangle {
             text: "Option"
             onClicked: {
                 checked = !checked
+            }
+        }
+    }
+
+    Component {
+        id: visibilityBoxSample
+
+        Column {
+            spacing: 8
+
+            width: 200
+
+            VisibilityBox {
+                text: "Visibility Box"
+
+                onVisibleToggled: {
+                    isVisible = !isVisible
+                }
             }
         }
     }
--- a/src/appshell/internal/applicationuiactions.cpp
+++ b/src/appshell/internal/applicationuiactions.cpp
@@ -102,8 +102,8 @@ const UiActionList ApplicationUiActions::m_actions = {
              ),
     UiAction("inspector",
              mu::context::UiCtxNotationOpened,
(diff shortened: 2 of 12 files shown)


src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
```diff
@@ -389,6 +390,43 @@ Rectangle {
...
+
+        Column {
+            spacing: 8
+
+            width: 200
+
+            VisibleButton {
+                text: "Visible Button"
+
+                onVisibleEdited: {
```
vpereverzev, line 404: not quite sure why would we need to define it every time, I believe it should happen internally and you could just receive a signal for an additional logic

vpereverzev (approved):

src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
```diff
@@ -389,6 +390,43 @@ Rectangle {
...
+                    isVisible = !isVisible
+                }
+            }
+
+            VisibleButton {
+                contentItem: RowLayout {
```
vpereverzev, line 410: I think it makes more sense to keep it simple

I mean if I'd want to attach another button to this one, I'd make that layout locally, otherwise the component seems to be overcomplicated

vpereverzev: Or it's just will be another component who will make that layout for me

src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
```diff
@@ -389,6 +390,43 @@ Rectangle {
...
 
+    Component {
+        id: visibleButtonSample
+
+        Column {
+            spacing: 8
+
+            width: 200
+
+            VisibleButton {
```
vpereverzev, line 401: it might make sense to rename it onto VisibilityButton or just an EyeButton

src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
Eism: But it is already signal) 
You can see how I use this signal in the inspector

src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
Eism: agree with you, I forgot to rename to VisibilityButton

src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
vpereverzev: I haven't got the point :)

1 - it's still makes sense to keep the notification signal
2 - I'd expect that "VisibilityButton" component would take care about isVisible property by itself

Eism: We usually bind the isVisible property to the model. When the user tries to change a property, the value is sent to the model. There may be logic in the model that prohibits changing the property, then the model will ignore and nothing will change.

The component should just show the data that was provided to it, and not manage something

src/appshell/qml/DevTools/Gallery/GeneralComponentsGallery.qml
Eism: It will be difficult to reuse in the instruments panel
there is another button between the eye and the text
image

PR #8553, opened 2021-07-06
Fixed Dropdown issues

igorkorsukov:
https://www.dropbox.com/s/3m9f362rapy6f5j/Dropdown%20Box%20v1.pdf?dl=0


src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
```diff
@@ -0,0 +1,224 @@
...
+    property int popupItemsCount: root.defaultPopupItemsCount()
+
+    property alias navigation: navCtrl
+
+    property alias dropIcon: dropIconItem
+
+    property string textRole: "text"
+    property string valueRole: "value"
+
+    text: root.displayText
```
vpereverzev, line 53: Is there any particular reason to have 3 properties like "displayedText", "currentText" and "text" simultaneously? It's a bit confusing which one to use

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
```diff
@@ -0,0 +1,224 @@
...
+
+            return root.model[index][roleName]
+        }
+
+        // Complex models (like QAbstractItemModel) with multiple role names
+        if (!(index < delegateModel.count)) {
+            return def
+        }
+
+        var item = delegateModel.items.get(index)
```
vpereverzev, line 95: maybe we can get it directly from the root.model or even from the ListView.model

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
igorkorsukov: hmm, I will remove displayedText

src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
igorkorsukov: unfortunately no, but I think maybe we need our own class for accessing to QAbstractListModel in Qml

igorkorsukov: I will do it differently in the next PR

vpereverzev (approved):

PR #8492, opened 2021-07-01
[MU3 Backend] ENG-59: Fix missing secondary beam breaks

iveshenry18:
Resolves: [ENG-59](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-59): Fix missing breaks in secondary beams

Previously, only the first beam was taken into account when importing
Beam::Mode. This commit adds a mechanism to collect the beam type
for all beams and make a more correct choice about Beam::Mode (including
BEGIN32 and BEGIN64). Additionally, it adds handling of these modes to
the export code. Now the round trip from .mscx -> .xml -> .mscx is
correct and complete *when beam modes are specified explicitly in the
.mscx*; however, more thorough handling of Beam::Mode::AUTO (where it
results in the equivalent of BEGIN32 or BEGIN64) on export needs to be
done.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8284, opened 2021-06-08
[MU3 Backend] ENG-41: Add space between tempo text and metronome marking

iveshenry18:
Resolves: [ENG-41](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-41): Add space between tempo and metronome mark

Adds a space between the words and the metronome marking upon MusicXML import,
if there isn't already a space present at the end of the words.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8554, opened 2021-07-06
[MU3 Backend] MusicXML compiler warnings

iveshenry18:
Resolves: Compiler warnings mentioned in #8492, #8325, and #8284

### DIFF
--- a/importexport/musicxml/importmxmlpass2.cpp
+++ b/importexport/musicxml/importmxmlpass2.cpp
@@ -2748,7 +2748,11 @@ void MusicXMLParserDirection::direction(const QString& partId,
                         t = new TempoText(_score);
                         QString rawWordsText = _wordsText;
                         rawWordsText.remove(QRegularExpression("(<.*?>)"));
-                        QString sep = _metroText != "" && _wordsText != "" && rawWordsText.back() != ' ' ? " " : "";
+                        #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+                              QString sep = _metroText != "" && _wordsText != "" && rawWordsText.back() != ' ' ? " " : "";
+                        #else
+                              QString sep = _metroText != "" && _wordsText != "" && rawWordsText.at(rawWordsText.size() - 1) != ' ' ? " " : "";
+                        #endif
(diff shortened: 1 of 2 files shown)


importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -4925,7 +4926,7 @@ Note* MusicXMLParserPass2::note(const QString& partId,
             else if (_e.name() == "stem")
                   stem(stemDir, noStem);
             else if (_e.name() == "type") {
-                  small = _e.attributes().value("size") == "cue" | _e.attributes().value("size") == "grace-cue";
+                  small = _e.attributes().value("size") == "cue" || _e.attributes().value("size") == "grace-cue";
```
Jojo-Schmitz, line 4929: Better also add parantheses to the expression?

iveshenry18: It's not necessary (equality operators have precedence over logical OR, and bitwise OR for that matter. But if it'll help everyone sleep at night I'm happy to!

Jojo-Schmitz: I added them first to stop the compiler warning, then detected the | vs || bug and fixed that too.

Jojo-Schmitz: I checked now thogh: it does. Still those parens won't harm. OTH it shows that compiler warnings sometimes even point at a bug ;-)

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -4926,7 +4926,7 @@ Note* MusicXMLParserPass2::note(const QString& partId,
             else if (_e.name() == "stem")
                   stem(stemDir, noStem);
             else if (_e.name() == "type") {
-                  small = _e.attributes().value("size") == "cue" | _e.attributes().value("size") == "grace-cue";
+                  small = (_e.attributes().value("size") == "cue") || (_e.attributes().value("size") == "grace-cue");
```
Jojo-Schmitz, line 4929: Hmm, I was suggesting the parenthezise the entire || expression, not their elements

PR #8555, opened 2021-07-06
Fix #8546: Fix uses of deprecated Qt classes and methods

### LINKED ISSUE #8546: [MU4 Task] [TECH_DEBTS] Provide an alternative solution for the Qt warning fixes
**Task description**
Replace https://github.com/musescore/MuseScore/commit/709fc493a45f925707eb4afe80758b0f6f29c2cc with an alternative solution

cbjeukendrup:
Resolves: #8546 
(and some other warnings and similar things)

### DIFF
--- a/build/pch/pch.h
+++ b/build/pch/pch.h
@@ -35,7 +35,6 @@ It is best to use a tool to determine the most commonly used inclusions.
 See https://github.com/musescore/musescore_devtools/tree/main/include-what-you-use
 */
 
-
 // Std includes
 #include <memory> //413
 #include <vector> //401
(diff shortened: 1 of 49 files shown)


wizofaus:
Wow that touches a lot of a files, would be good to get tested & merged ASAP to prevent conflicts. Do you need help getting it to build with msvc?

wizofaus:
`int __cdecl version(void)` is unresolved - not sure why it's expecting it to be __cdecl, but moving `extern int version();` out of the parseVersion function fixes this.
I'm just not sure what the justification for having this as a standalone function in util.cpp without a header file for it is?

cbjeukendrup:
> I'm just not sure what the justification for having this as a standalone function in util.cpp without a header file for it is?

I missed that it was also in the utils.h header. Now it should be better.

igorkorsukov (approved):

PR #8556, opened 2021-07-06
[MU3 Backend] ENG-38: Handle hanging ties

iveshenry18:
Resolves: [ENG-38](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-38): Allow hanging ties at end of score or before repeats

This commit keeps track of ties added to the score during 
MusicXML import, making sure they are
terminated before the end of the part. If not, the tie is deleted and
replaced with a l.v. articulation. 

Additionally, this include a fixup function for improving the placement of l.v. symbols on import (TODO: improve the automatic layout
of these articulations.)


wizofaus:
BTW, not directly related to this - but do you have any plans to improve how the musicxml import handles the duration element and the attack/release attributes? Seems the latter are being ignored currently (hence all information about how long each note should be played back for if not the full duration is lost), and the handling for when the "duration" element doesn't quite match the type & time-modification elements isn't complete (MusicXML seems to allow how a note should be notated to be specified independently of how much time it should take up, which I believe should be possible with MuseScore's native data structures but at least in the case of tuplets would cause some challenges).

oktophonie (approved):

iveshenry18:
We don't currently have plans for this, but I'll keep an eye out for any cases of it causing an issue! If it seems to negatively affect scores then I'll add it to our list.

wizofaus:
There's been a number of discussions in musescore.org about it, but the fact that the attack/release attributes are ignored entirely seems hard to justify, as it would seriously affect the playback of an imported score that did have those specified.

iveshenry18:
@wizofaus certainly it is a feature that should we should support, since it's in the spec.

That said, my current efforts are essentially prioritized according to a set of a couple hundred scores exported with the Dolet plugin for Sibelius (which is why many of my commits have to do with "inferring" incorrect exports from Dolet, even if those inferences are technically outside of the spec). It's not ideal, but it's a pragmatic approach in an ecosystem of imperfect implementations of a sufficiently complex spec. Among this set of (again, a couple hundred) scores, there is exactly one usage of the "release" attribute, and no uses of the "attack" attribute; furthermore, that one example of the "release" attribute doesn't really affect its score's playback in a noticeable way.

All this to say, I can't guarantee that I'll get to this issue, but I will if I get the chance! Of course, anyone else is always welcome to implement this and submit a PR.

wizofaus:
Understood, it's probably not a high priority fix overall, just figure if it was going to get done anywhere it would as part of what you've been working on.
Does Sibelius/Dolet allow notating something as a tuplet but overriding the internal note duration to be something else in such a way it affects XML export?

iveshenry18:
> Does Sibelius/Dolet allow notating something as a tuplet but overriding the internal note duration to be something else in such a way it affects XML export?

I'm not sure, and don't currently have a Sibelius license to test. The one example found in the set of scores I'm working with was on the last note of the piece, which happened to be an eighth note.

vpereverzev (approved):

PR #8581, opened 2021-07-08
[MU3 Backend] ENG-63: Properly handle cue grace notes

iveshenry18:
Resolves: [ENG-63](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-63): Grace notes within 'small' passages are not being reduced

Previously, our importer and exporter didn't account for notes that
were both grace notes and cue notes (therefore extra-small in size).
This commit amends that. Additionally, since Dolet also seems to fail
in this area, this commit adds a function to coerce the `small` and
`play` attributes of a grace chord to match that of its main chord.
Finally, it adds support for the "grace-cue" size to the exporter and
importer. In the importer, this is treated the same as the "cue" size, as
exporters tend to misinterpret the distinction between these.


oktophonie (approved):

vpereverzev (approved):

importexport/musicxml/importmxmlpass2.cpp
```diff
@@ -4842,7 +4871,7 @@ Note* MusicXMLParserPass2::note(const QString& partId,
             else if (_e.name() == "stem")
                   stem(stemDir, noStem);
             else if (_e.name() == "type") {
-                  small = _e.attributes().value("size") == "cue";
+                  small = _e.attributes().value("size") == "cue" | _e.attributes().value("size") == "grace-cue";
```
Jojo-Schmitz, line 4874: This causes a compiler warning due to mussing parentheses. Alsi IU guess it shpould be `||`` rather than `|`?

iveshenry18: #8554 (for posterity)

PR #8585, opened 2021-07-09
Added mscz reader and writer 

igorkorsukov:
These reader and writer encapsulate what and how we store in mscz file.


src/engraving/compat/mscxcompat.cpp
```diff
@@ -0,0 +1,67 @@
...
+            LOGE() << "failed open file: " << path;
+            return Ms::Score::FileError::FILE_OPEN_ERROR;
+        }
+
+        QByteArray mscxData = mscxFile.readAll();
+
+        QBuffer buf(&msczData);
+        MsczWriter writer(&buf);
+        filePath = path + ".mscz";
+        writer.setFilePath(filePath);
```
wizofaus, line 47: Is that assuming it's always OK to write to a file in the current directory that has the same name as the .mscx file but with .mscz appended to the end? Is that safe/sensible?
Is the a reason we only want Ms::Score( ) to be able to read from an .mscz file on disk, hence meaning we have to compress an mscx into a .mscz file first before it can then uncompress and parse it?

src/engraving/libmscore/score.h
```diff
@@ -1438,15 +1446,10 @@ class MasterScore : public Score
     void setTempomap(TempoMap* tm);
 
     bool saveFile(bool generateBackup = true);
-    FileError read1(XmlReader&, bool ignoreVersionError);
```
wizofaus, line 1441: good to see these gone from the public interface!  But it seems sensible to keep at least some support for loading uncompressed mscx files inside libmscore, unless I'm missing something...

src/userscores/internal/filescorecontroller.cpp
```diff
@@ -342,12 +342,12 @@ void FileScoreController::exportScore()
 
 io::path FileScoreController::selectScoreOpenningFile()
```
wizofaus, line 343: existing typo - should be Opening

src/engraving/io/msczreader.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+#include "log.h"
+
+//! NOTE The current implementation resolves files by extension.
+//! This will probably be changed in the future.
+
+using namespace mu::engraving;
+
+MsczReader::MsczReader(const QString& filePath)
+    : m_filePath(filePath), m_device(new QFile(filePath)), m_selfDeviceOwner(true)
```
wizofaus, line 38: Are we still using raw pointers for new code then?  Would seem to be some opportunities for using unique_ptr etc. here.

src/engraving/compat/mscxcompat.cpp
```diff
@@ -0,0 +1,67 @@
...
+        writer.open();
+        writer.writeScore(mscxData);
+    } else {
+        QFile msczFile(path);
+        if (!msczFile.open(QIODevice::ReadOnly)) {
```
vpereverzev, line 52: looks like we'll open and read the content of file even if it's not .mscz, makes sense to add an explicit "else if" or even make a validation at the beginning of the method

thirdparty/deto_async/async/internal/queuedinvoker.cpp
```diff
@@ -22,8 +22,12 @@ void QueuedInvoker::invoke(const std::thread::id& th, const Functor& f, bool isA
 
 void QueuedInvoker::processEvents()
 {
-    std::lock_guard<std::recursive_mutex> lock(m_mutex);
-    Queue& q = m_queues[std::this_thread::get_id()];
+    Queue q;
+    {
+        std::lock_guard<std::recursive_mutex> lock(m_mutex);
+        q = m_queues[std::this_thread::get_id()];
```
vpereverzev, line 28: might make sense to use extract here

vpereverzev: but I'm curious what was the problem?

igorkorsukov: Unfortunately, my stdlib implementation no has method `value` for `node_type`
```
auto n = m_map.extract(key);
T val = n.value(); <-  not compiled

output:
/home/igor/Dev/MuseScore/thirdparty/deto_async/async/internal/queuedinvoker.cpp:30:15: error: ‘class std::_Node_handle<std::thread::id, std::pair<const std::thread::id, std::queue<std::function<void()> > >, std::allocator<std::_Rb_tree_node<std::pair<const std::thread::id, std::queue<std::function<void()> > > > > >’ has no member named ‘value’
         q = n.value();
```

Ubuntu 18.04 gcc 7.5 (same as on CI)

igorkorsukov: > but I'm curious what was the problem?

I found, that when the application starts, after a few seconds, a freeze appears for about 5-10 seconds.
Freeze calls because of attempts to lock this mutex when call "invoke", that is, someone, another thread, holds this mutex in this place when processing messages.
that is, there is a very long handler that takes a long time to execute and, accordingly, holds the mutex.
In fact, we need to hold the mutex while we process the messages. We only need a mutex to pop the message queue.

src/engraving/compat/mscxcompat.cpp
igorkorsukov: We are planning to add new data storage to `mscz`, such as mixer settings, VST settings, etc.
We do not want to add this directly to the score xml file.
Moreover, we want to remove some data from score xml file, for example, styles, and store them separately.
Also, for example, we can separately store meta-information about the score - the author, title, etc., so as not to open and read the entire score to get it.
As a result, it turns out that the container will store many different files with different information, similar to how "open documents" are stored.
Accordingly, we will have to abandon the `mscx` format.
Therefore, I made it so that the main format is `mscz`, and `mscx` is opened in compatibility mode.

src/engraving/io/msczreader.cpp
igorkorsukov: `m_device` can be created here or passed from outside (another constructor or setter). We can require a shared point, but in practice, it is not convenient in this case. Because a typical use case is as follows:
```
QByteArray data = ...
QBuffer buf(&data);
MsczReader reader(&buf);
```

src/engraving/compat/mscxcompat.cpp
wizofaus: Yes I gathered that (and I agree it makes sense), but when reading an mscx file, creating a dummy mscz file with a particular filename first just doesn't seem like the best approach when Score can already read an mscx directly.

igorkorsukov: This file is created in memory (in a buffer, not in the filesystem).
I believe that we need to lighten `libmscore` as much as possible, it is already heavily overloaded with various code.
I think the opening of `mscx` will be only for tests, and very little is needed by users.

wizofaus: Ah ok, the use of "setFilePath" made it look like it was writing to a file, that was really my only concern (though sure, compressing just to decompress again seems a bit inefficient, but unlikely to be a real problem).

src/engraving/io/msczreader.cpp
wizofaus: Yes I understand that, but in the case it does need to create/manage its own device, was thinking unique_ptr or similar would be appropriate.

src/engraving/compat/mscxcompat.cpp
igorkorsukov: > Ah ok, the use of "setFilePath" made it look like it was writing to a file, that was really my only concern (though sure, compressing just to decompress again seems a bit inefficient, but unlikely to be a real problem).

Unfortunately, despite the fact that we are working with a buffer, we need to set the name of the files at the moment. Because the name of the score xml is determined by the name of the file.
I believe that we can change this and store the score xml with a constant name like `score.mscx`.

thirdparty/deto_async/async/internal/queuedinvoker.cpp
vpereverzev: it's confusing, but there is no method "value()" for a node_handle of std::map, it's called "mapped()". It's a value from <key, value> pair in node_handle from std::map

src/userscores/internal/filescorecontroller.cpp
igorkorsukov: thanks, fixed

src/engraving/compat/mscxcompat.cpp
igorkorsukov: added check

thirdparty/deto_async/async/internal/queuedinvoker.cpp
igorkorsukov: thanks, added use `extract`

vpereverzev (approved):

PR #8586, opened 2021-07-10
[MU4] Fix #7767 - ensure undo works properly for staff properties dialog

### LINKED ISSUE #7767: [MU4 Issue] Can't undo stave transposition (made via right-click > Stave Properties)
**Describe the bug**
Can't undo stave transposition (made via right-click > Stave Properties)
Also, can't undo text editing

**To Reproduce**
1. Create score
2. Select note
3. right-click > Stave Properties

**Expected behavior**
stave transposition functions work

MacOS

**Additional context**
About transposition
https://musescore.org/en/handbook/3/transposition

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/7767

Undo wasn't correctly working for changing various staff properties

### DIFF
--- a/src/notation/internal/notationparts.cpp
+++ b/src/notation/internal/notationparts.cpp
@@ -706,8 +706,8 @@ void NotationParts::replaceInstrument(const ID& instrumentId, const ID& fromPart
         return;
     }
 
-    part->setInstrument(InstrumentsConverter::convertInstrument(newInstrument), oldInstrumentInfo.fraction);
-    doSetPartName(part, formatPartName(part));
+    score()->undo(new Ms::ChangePart(part, new Ms::Instrument(InstrumentsConverter::convertInstrument(newInstrument)),
+                                     formatPartName(part)));
     updateScore();
 
     ChangedNotifier<mu::instruments::Instrument>* notifier = partNotifier(part->id());
--- a/src/notation/view/widgets/editstaff.cpp
+++ b/src/notation/view/widgets/editstaff.cpp
@@ -31,6 +31,7 @@
 #include "libmscore/stringdata.h"
 #include "libmscore/text.h"
 #include "libmscore/utils.h"
+#include "libmscore/undo.h"
 
 #include "log.h"
 #include "translation.h"
@@ -303,8 +304,10 @@ void EditStaff::bboxClicked(QAbstractButton* button)
 
 void EditStaff::apply()
 {
+    int index = m_staff->score()->undoStack()->getCurIdx();
     applyStaffProperties();
     applyPartProperties();
+    m_staff->score()->undoStack()->mergeCommands(index);
 }
 
 void EditStaff::minPitchAClicked()


src/notation/internal/notationparts.cpp
```diff
@@ -706,7 +706,7 @@ void NotationParts::replaceInstrument(const ID& instrumentId, const ID& fromPart
         return;
     }
 
-    part->setInstrument(InstrumentsConverter::convertInstrument(newInstrument), oldInstrumentInfo.fraction);
+    score()->undo(new Ms::ChangePart(part, new Ms::Instrument(InstrumentsConverter::convertInstrument(newInstrument)), part->name()));
     doSetPartName(part, formatPartName(part));
```
wizofaus, line 710: It does concern me that I'm now discarding oldInstrumentInfo.fraction, but I can't find a case that it's anything other than -1 - in fact that seems to be a bug, because in MU3 if you use "staff properties" at a point after a "Change Instrument" event, it shows the transposition etc. info as set by that "change instrument" event and that's what you can edit - whereas for MU4 "Change Instrument" events seem to be ignored in the staff properties dialog.
MU3 has a bunch of code in editstaff.cpp (lines 381-) to support this, but the MU4 code doesn't pass any info about the current score position when doing this.

src/notation/internal/notationparts.cpp
```diff
@@ -706,7 +706,7 @@ void NotationParts::replaceInstrument(const ID& instrumentId, const ID& fromPart
         return;
     }
 
-    part->setInstrument(InstrumentsConverter::convertInstrument(newInstrument), oldInstrumentInfo.fraction);
+    score()->undo(new Ms::ChangePart(part, new Ms::Instrument(InstrumentsConverter::convertInstrument(newInstrument)), part->name()));
```
RomanPudashkin, line 709: score()->undo(new Ms::ChangePart(...)) will be called again right after this line (see doSetPartName implementation). Better to use undo only once

wizofaus: Ah yes, missed that, agree it makes sense to use just put the one ChangePart command on the undo stack, thanks.

src/notation/internal/notationparts.cpp
```diff
@@ -706,8 +706,8 @@ void NotationParts::replaceInstrument(const ID& instrumentId, const ID& fromPart
         return;
     }
 
-    part->setInstrument(InstrumentsConverter::convertInstrument(newInstrument), oldInstrumentInfo.fraction);
-    doSetPartName(part, formatPartName(part));
+    score()->undo(new Ms::ChangePart(part, new Ms::Instrument(InstrumentsConverter::convertInstrument(newInstrument)),
+                                     formatPartName(part)));
```
wizofaus, line 710: for Undo to work properly we can't change the instrument on the part before creating the "ChangePart" command, as it needs access to the original value in order to revert to it when you undo. ChangePart can also change the name, so no need to call doSetPartName explicitly.

RomanPudashkin (approved):

PR #8593, opened 2021-07-12
[MU4] Disabled restoring of states

RomanPudashkin:


src/appshell/view/dockwindow/dockwindow.cpp
```diff
@@ -436,20 +433,26 @@ void DockWindow::restorePageState(const QString& pageName)
...
 
+bool DockWindow::restoreLayout(const QByteArray& layout, KDDockWidgets::RestoreOptions)
+{
+    if (layout.isEmpty()) {
+        return true;
+    }
+
+    //! TODO: restoring of layout is temporary disabled because it causes many troubles for developers
+    //KDDockWidgets::LayoutSaver layoutSaver(option);
+    //return layoutSaver.restoreLayout(state);
```
igorkorsukov, line 451: Maybe we need to add a message to the log so that we don't forget and it reminds us of this moment.

igorkorsukov (approved):

src/appshell/view/dockwindow/dockwindow.cpp
RomanPudashkin: Do you mean the NOT_IMPLEMENTED macro?

igorkorsukov: I think like:
```
LOGI() << "TODO: restoring of layout is temporary disabled because it troubles ";
```

RomanPudashkin: fixed

PR #8598, opened 2021-07-12
Added engraving and notation project concept

igorkorsukov:
In addition to the score itself, a mscz file also stores other data, for example, settings for a synthesizer, mixer, omr, etc.
So, we should talk not just about a score, but about a Project.
To some extent, the project functions are performed by MasterScore, but this is not its responsibility, this functionality only overloads it and confuses what, where and why.

I add explicitly the concept of a project.

### DIFF
--- a/src/importexport/guitarpro/internal/importgtp.cpp
+++ b/src/importexport/guitarpro/internal/importgtp.cpp
@@ -2975,7 +2975,7 @@ Score::FileError importGTP(MasterScore* score, const QString& name)
             continue;
         }
         QMultiMap<int, int> tracks;
-        Score* pscore = new Score(score);
+        Score* pscore = score->createScore();
         //TODO-ws		pscore->showLyrics = score->showLyrics;
         pscore->style().set(Sid::createMultiMeasureRests, false);
         pscore->style().set(Sid::ArpeggioHiddenInStdIfTab, true);
--- a/src/notation/internal/notationproject.cpp
+++ b/src/notation/internal/notationproject.cpp
@@ -1,0 +1,236 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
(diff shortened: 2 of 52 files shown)


src/importexport/guitarpro/internal/importgtp.cpp
```diff
@@ -2975,7 +2975,7 @@ Score::FileError importGTP(MasterScore* score, const QString& name)
             continue;
         }
         QMultiMap<int, int> tracks;
-        Score* pscore = new Score(score);
+        Score* pscore =  score->createScore();
```
worldwideweary, line 2978: Extra space? tsk tsk!

igorkorsukov: thanks, fixed

src/notation/internal/notationproject.cpp
```diff
@@ -0,0 +1,236 @@
...
+    Ret ret = reader->read(score, path, options);
+    if (!ret) {
+        return ret;
+    }
+
+    if (!Ms::MScore::lastError.isEmpty()) {
+        LOGE() << Ms::MScore::lastError;
+    }
+
+    project->setupMasterScore();
```
igorkorsukov, line 84: In general, the loading process is ugly right now, I will try to make it better in the next PR

Eism (approved):

src/userscores/view/templatepaintview.cpp
```diff
@@ -63,16 +63,16 @@ QString TemplatePaintView::zoomOutSequence() const
...
 
     if (!ret) {
         LOGE() << ret.toString();
     }
 
-    setNotation(masterNotation->notation());
+    setNotation(notationProject->masterNotation()->notation());
 
-    if (masterNotation) {
+    if (notationProject) {
```
wizofaus, line 75: Something seems wrong here (though it seemed wrong already), the notation project gets deleted when the shared_ptr goes out of scope, but there are still references to it (via the setNotation call), which is causing MuseScore to crash doing almost anything now. @igorkorsukov

Issue #8592, opened 2021-07-12
[MU4 Issue] Crash when adjusting grip handle for trill


wizofaus:
**Describe the bug**
MU4 (and MU3) crashes randomly when dragging trill grip handles around

**To Reproduce**
1. Add a trill line to the first measure of a new score (default treble clef only template)
2. Drag the end grip handle towards the top of the page
3. MuseScore crashes

**Expected behavior**
Probably should avoid actually crashing.
Note that this isn't specific to trills, it happens with basically any sort of line with grip handles that can be moved vertically.

**Screenshots**

**Desktop (please complete the following information):**
 - Windows 10, built with MSVC 2019
 - 
**Additional context**

Exception thrown at 0x0000000140299C91 in MuseScore4.exe: 0xC0000005: Access violation reading location 0x0000000000000060.

>	MuseScore4.exe!Ms::System::staffCanvasYpage(int staffIdx) Line 1335	C++
 	MuseScore4.exe!Ms::LineSegment::findSegmentForGrip(Ms::Grip grip, mu::PointX<double> pos) Line 440	C++
 	MuseScore4.exe!Ms::LineSegment::rebaseAnchors(Ms::EditData & ed, Ms::Grip grip) Line 594	C++
 	MuseScore4.exe!Ms::LineSegment::editDrag(Ms::EditData & ed) Line 695	C++
 	MuseScore4.exe!mu::notation::NotationInteraction::drag(const mu::PointX<double> & fromPos, const mu::PointX<double> & toPos, mu::notation::DragMode mode) Line 672	C++
 	MuseScore4.exe!mu::notation::NotationViewInputController::mouseMoveEvent(QMouseEvent * event) Line 406	C++
 	MuseScore4.exe!mu::notation::NotationPaintView::mouseMoveEvent(QMouseEvent * event) Line 667	C++

For some strange reason the system has 0 staves.

wizofaus:
Can easily fix by changing line.cpp, line 435 to

`    if (!foundSystems.empty() && !foundSystems.contains(sys) && foundSystems[0]->staves()->size()) {`

but I don't quite understand why the page does have an "empty" system with no staves on it and why this problem doesn't seem to exist in 3.6.

PR #8600, opened 2021-07-12
Fix issues discovered while moving grip handles around on trills

wizofaus:
Resolves: https://musescore.org/en/node/322957 https://github.com/musescore/MuseScore/issues/8592

Two unrelated issues, neither really to do with trill lines specifically at all but both discovered at the same time:

a) if you drag the grip handle from any line into a 0-stave system (e.g. for a V-frame, as new scores have by default at the top of the score), it crashes
b) a bug in the layout engine where it reused cached information from a previous layout caused erratic behaviour when dragging grip handles around, though it could be triggered in other ways,

### DIFF
--- a/src/engraving/libmscore/layout.cpp
+++ b/src/engraving/libmscore/layout.cpp
@@ -2156,6 +2156,9 @@ System* Score::getNextSystem(LayoutContext& lc)
     if (!isVBox) {
         int nstaves = Score::nstaves();
         system->adjustStavesNumber(nstaves);
+        for (int i = 0; i < nstaves; ++i) {
+            system->staff(i)->setShow(score()->staff(i)->show());
+        }
     }
     return system;
 }
--- a/src/engraving/libmscore/line.cpp
+++ b/src/engraving/libmscore/line.cpp
@@ -432,7 +432,7 @@ Segment* LineSegment::findSegmentForGrip(Grip grip, PointF pos) const
     System* sys = oldSeg->system();
     const QList<System*> foundSystems = score()->searchSystem(pos, sys, spacingFactor);
 
-    if (!foundSystems.empty() && !foundSystems.contains(sys)) {
+    if (!foundSystems.empty() && !foundSystems.contains(sys) && foundSystems[0]->staves()->size()) {
         sys = foundSystems[0];
     }
 


src/engraving/libmscore/layout.cpp
```diff
@@ -2156,6 +2156,9 @@ System* Score::getNextSystem(LayoutContext& lc)
     if (!isVBox) {
         int nstaves = Score::nstaves();
         system->adjustStavesNumber(nstaves);
+        for (int i = 0; i < nstaves; ++i) {
+            system->staff(i)->setShow(score()->staff(i)->show());
+        }
```
wizofaus, line 2161: without this, it "remembers" which staves were hidden for this system on the previous layout, but that changes whether it takes into account the instrument name when calculating the width, meaning you can get a different result depending on the state of the cache.

src/engraving/libmscore/line.cpp
```diff
@@ -432,7 +432,7 @@ Segment* LineSegment::findSegmentForGrip(Grip grip, PointF pos) const
     System* sys = oldSeg->system();
     const QList<System*> foundSystems = score()->searchSystem(pos, sys, spacingFactor);
 
-    if (!foundSystems.empty() && !foundSystems.contains(sys)) {
+    if (!foundSystems.empty() && !foundSystems.contains(sys) && foundSystems[0]->staves()->size()) {
```
wizofaus, line 435: ignore 0-staff systems (used for V-Boxes etc.) when trying to find system during grip handle dragging

igorkorsukov (approved):

src/engraving/libmscore/layout.cpp
Jojo-Schmitz: This change, backported to the 3.x branch, causes MuseScore(3.exe) to crash (in Debug builds, QtCreator, MinGW)), on  opening any score due to a failed assertion in Qt:
Fatal: ASSERT failure in QList<T>::operator[]: "index out of range", file C:/Qt/5.15.2/mingw81_64/include/QtCore/qlist.h, line 575

Not sure how relevant this is to the master branch though.

wizofaus: I worked this fix out in MU3 and if it'd crashing there (which it wasn't for me) it's wrong in MU4 too - I'll have a look when I can.

Jojo-Schmitz: It certainly crashes for me.

Jojo-Schmitz: Oops, I added these 3 lines a line too late (before the `system->adjustStavesNumber(nstaves);` rather than after), downside of an all-manual cherry-pick. Have a brown bag for me?
Sorry for the noise...

wizofaus: Yeah I was pretty sure it was OK!  Took me hours to work out what that problem was too BTW, surprised it was such a simple change in the end, and a bit surprised it doesn't cause more weird layout issues.

Jojo-Schmitz:
It also solves https://musescore.org/en/node/323559. Yet another good candidate for backporting to 3.x (I'd have it deary) and a 3.6.3!

PR #8603, opened 2021-07-13
Fixed crash with master score

igorkorsukov:


src/engraving/engravingproject.cpp
```diff
@@ -30,22 +30,26 @@ using namespace mu::engraving;
 
 std::shared_ptr<EngravingProject> EngravingProject::create()
 {
-    return std::shared_ptr<EngravingProject>(new EngravingProject(Ms::MScore::defaultStyle()));
+    std::shared_ptr<EngravingProject> p = std::shared_ptr<EngravingProject>(new EngravingProject());
+    p->init(Ms::MScore::defaultStyle());
```
wizofaus, line 34: not sure I understand why this is an improvement?

src/engraving/libmscore/score.cpp
```diff
@@ -5289,8 +5292,8 @@ MasterScore::MasterScore()
     metaTags().insert("creationDate", QDate::currentDate().toString(Qt::ISODate));
 }
 
-MasterScore::MasterScore(const MStyle& s)
-    : MasterScore{}
+MasterScore::MasterScore(const MStyle& s, std::shared_ptr<engraving::EngravingProject> project)
+    : MasterScore{project}
```
wizofaus, line 5296: so EngravingProject keeps a raw pointer to MasterScore, but MasterScore keeps a shared_ptr to EngravingProject? Feels a bit odd to me...but I'm not quite sure how an EngravingProject is supposed to behave.

src/engraving/engravingproject.cpp
igorkorsukov: At the moment the project is creating the master score and can be given to someone else as a raw pointer.
This leads to two problems:
1. The link to the project may not be saved and, accordingly, the project object can be deleted, and since it created the master score, it must delete it, BUT, someone at this moment can store a link to this master score and use it, respectively, a crash will occur (this crash was when opening templates)

To fix this, I added a link to the project in the master score so that the project will not be deleted while the score that it created is alive.
But the link to `shared_from_this` cannot be passed in the constructor, because at this moment the shared pointer has not yet been created. So, I had to add the init method. In this case, it is safe, because the project constructor is not public and the project can be created only by calling the create function.

2. The second problem is that there could be double deletions.
And the one we gave the master score deletes it and the project deletes it.

To solve this problem, I added the corresponding code to the destructor of the master score

In general, these are just difficulties of the transition period. In the future, we will fix all this, at least we should use the master score as a shared pointer

Eism (approved):

src/engraving/libmscore/score.cpp
igorkorsukov: Step by step we will lead everything to how it should be :)

src/engraving/engravingproject.cpp
wizofaus: Ok...I guess have to trust you! But I've seen a number of worrying crashes recently because of the mix of raw and shared_ptr's - I'm all for using shared_ptr's but it seems once you use them once you need to use them everywhere or the code becomes increasingly unstable.  I did try fixing by having the NotationPaintView *own* the NotationProject as a member, but while it prevented the original crash it still crashed later on (as NotationProject::createNew had the same lifetime issue, where the pointer to MasterScore was stored and used after the EngravingProject went out of scope...).

PR #8608, opened 2021-07-13
Added project audio settings prototype

igorkorsukov:


src/engraving/io/msczwriter.h
```diff
@@ -44,21 +44,32 @@ class MsczWriter
...
+    void writeAudioSettingsJsonFile(const QByteArray& data);
 
 private:
 
     struct Meta {
-        QString mscxFileName;
-        std::vector<QString> imageFilePaths;
-        QString audioFile;
-
+        std::vector<QString> files;
```
vpereverzev, line 56: might make sense to use io::paths

src/notation/internal/notationproject.cpp
```diff
@@ -36,6 +42,7 @@ NotationProject::NotationProject()
 {
     m_engravingProject = EngravingProject::create();
     m_masterNotation = std::shared_ptr<MasterNotation>(new MasterNotation());
+    m_projectAudioSettings = std::shared_ptr<ProjectAudioSettings>(new ProjectAudioSettings());
```
vpereverzev, line 45: would be slightly better to use std::make_shared

src/notation/internal/notationproject.cpp
```diff
@@ -48,31 +55,102 @@ mu::Ret NotationProject::load(const io::path& path, const io::path& stylePath, b
...
+    ProjectAudioSettingsPtr audioSettings = std::shared_ptr<ProjectAudioSettings>(new ProjectAudioSettings());
+    Ret ret = audioSettings->read(reader);
+    if (!ret) {
+        return ret;
+    }
+
+    // Set current if all success
+    m_engravingProject = project;
+
+    m_masterNotation = std::shared_ptr<MasterNotation>(new MasterNotation());
```
vpereverzev, line 123: It's a bit confusing, why would we need to create MasterNotation and ProjectAudioSettings twice.The first time in constructor of NotationProject, the second time here.

vpereverzev (approved):

src/engraving/io/msczwriter.h
igorkorsukov: not now.
At the moment engraving module, not dependent on anything, and this is good. `io:path` in another module, and it depends on a lot of things

src/notation/internal/notationproject.cpp
igorkorsukov: not this case, I did `ProjectAudioSettings` constructor private (for controlling who can create it), so we can't use `std::make_shared` in such cases

src/notation/internal/notationproject.cpp
igorkorsukov: In the constructor, I create these objects to ensure their existence, that is, I want to "guarantee" that the project always returns always valid objects, so as not to check for null in many places.

In loading, I create new objects to ensure that no state remains from the old object and to indicate that these are objects of the new object.

So, a double creation is obtained. It can be optimized, but I don't think it's necessary.

PR #8617, opened 2021-07-14
Added Project module

igorkorsukov:
So, now we should think and talk 

* Create not score (or notation), but create notation project 
* Save not score (or notation), but save notation project 
* Import not score (or notation), but import notation project 
* Export not score (or notation), but export notation project 

and so on


Eism (approved):

src/project/projecttypes.h
```diff
@@ -0,0 +1,54 @@
...
+#include "notation/notationtypes.h"
+
+namespace mu::project {
+struct ProjectCreateOptions
+{
+    QString title;
+    QString subtitle;
+    QString composer;
+    QString lyricist;
+    QString copyright;
```
vpereverzev, line 38: All that fields are called as "meta" even in MU3, might make sense to pull them into a separate ProjectMeta struct

vpereverzev (approved):

src/project/projecttypes.h
igorkorsukov: yes, we have a structure "meta", but now there are a lot of different things in it, so I decided not to use it here yet, but I also don’t want to introduce another structure "meta" :)

PR #8618, opened 2021-07-14
Engraving style. Step 1

igorkorsukov:
Moved style to separate dir
And separate MStyle class and Sid with default values


vpereverzev (approved):

src/engraving/libmscore/volta.cpp
```diff
@@ -22,18 +22,19 @@
 
 #include "volta.h"
 
+#include <algorithm>
+
+#include "style/style.h"
+
```
wizofaus, line 28: do we need all these extra blank lines?

wizofaus:
Hope this doesn't cause too many conflicts for others!

src/engraving/libmscore/volta.cpp
igorkorsukov: I usually try to group includes
but this is not necessary, optional

PR #8620, opened 2021-07-14
[MU4] Fixed instruments names in the instruments panel

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/pull/8432#pullrequestreview-702245624

### DIFF
--- a/src/instrumentsscene/view/instrumentlistmodel.cpp
+++ b/src/instrumentsscene/view/instrumentlistmodel.cpp
@@ -37,21 +37,6 @@ static const QString CONFIG_KEY("config");
 static const QString SOLOIST_KEY("isSoloist");
 static const QString IS_EXISTING_PART_KEY("isExistingPart");
 
-static QString formatSelectedInstrumentTitle(const Instrument& instrument)
-{
-    const QString& traitName = instrument.trait.name;
-    const QString& instrumentName = instrument.name;
-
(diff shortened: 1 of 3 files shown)


RomanPudashkin:
@shoogle please, check this

shoogle (changes requested):
@RomanPudashkin, almost! We'd like the part name to include information in parentheses that is displayed in the New Score Wizard, such as for "Horn (bass clef)", "Tenor Lute (5-course)", etc.

I suspect you will need to create a new function `formatPartName` that takes the output of [formatSelectedInstrumentTitle](https://github.com/musescore/MuseScore/blob/3c6ad5f14f82834abf302410dc05bcb3a98d77fb/src/instrumentsscene/view/instrumentlistmodel.cpp#L40-L53) and then adds a number on the end if necessary like in [formatInstrumentName](https://github.com/musescore/MuseScore/blob/3c6ad5f14f82834abf302410dc05bcb3a98d77fb/src/notation/internal/notationparts.cpp#L43-L52). The number should match the number in the score.

### Desired result

image


image

image

image

RomanPudashkin:
@shoogle I've updated my solution. Can you check it again, please?

shoogle (approved):
@RomanPudashkin, perfect, thanks!

Eism (approved):

Issue #8626, opened 2021-07-15
[MU4 Issue] New scores have no barlines/stems etc,


wizofaus:
**Describe the bug**
Creating a new score using the "Choose Instruments" dialog shows staves with no barlines and when you add notes the stems aren't visible.

**To Reproduce**
1. Home|New Score|Choose Instruments
2. Add at least 1 instrument, click Done
3. Observe score has no barlines (and clef is too close to left edge)
4. Add 1/4 or shorter length notes, observer notes have no stems

**Expected behavior**
Barlines and stems would be nice 

**Screenshots**

**Desktop (please complete the following information):**
 - Windows 10/built with MSVC

**Additional context**
Problem doesn't happen creating a score from a template or loading an existing score. Appears to be something messed up with the default score style values.

wizofaus:
Can be fixed by adding precomputeValues() to MStyle constructor or Score::setStyle(), line 5222 before the first return.
Former seems safer...

igorkorsukov:
It looks like we have long ago lost the call to `precomputeValues` of `defaultStyle`, it's strange what worked.

PR #8627, opened 2021-07-15
Clean up engraving style class

igorkorsukov:
Also fixed https://github.com/musescore/MuseScore/issues/8626


src/engraving/compat/pageformat.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace mu::engraving::compat {
+//---------------------------------------------------------
+//   @@ PageFormat
+//---------------------------------------------------------
+
+class PageFormat
+{
+    mu::SizeF _size;
+    qreal _printableWidth;          // _width - left margin - right margin
```
vpereverzev, line 40: might make sense to initialize values here

vpereverzev (approved):

src/engraving/compat/pageformat.h
igorkorsukov: as it was, so I keep, this is an obsolete class, it is better to change it less and it is not worth wasting time

Eism (approved):

PR #8634, opened 2021-07-15
Clean up Palettes code

cbjeukendrup:
- Remove and refactored some old code
- Add keyboard navigation for StyledPopup, CreateCustomPalettePopup, AddPalettesPopup
- Fix the "palette-search" action (Cmd/Ctrl + F9)


src/framework/global/uri.cpp
```diff
@@ -203,3 +203,10 @@ void UriQuery::addParam(const std::string& key, const Val& val)
 {
     m_params[key] = val;
 }
+
+UriQuery UriQuery::addingParam(const std::string& key, const Val& val) const
```
RomanPudashkin, line 207: I think it is better to use Uri Query::addParam directly where you need to add a param to URI. This method is a little bit unclear (at least for me :))

PR #8643, opened 2021-07-16
Cleanup style and default style classes

igorkorsukov:
Step 4
step by step...


vpereverzev (approved):

igorkorsukov:
Legacy styles were not loaded before. Now they are loaded, but it turned out that the test files are not actuale. So, I updated the test files

PR #8644, opened 2021-07-16
Moved chordlist from style to score

igorkorsukov:
Step 5
step by step...


igorkorsukov:
Legacy styles were not loaded before. Now they are loaded, but it turned out that the test files are not actuale. So, I updated the test files

vpereverzev (approved):

PR #8653, opened 2021-07-18
[MU4] fix #8652 - possible fix for ensuring text styles don't affect appearance of ScoreText symbols

### LINKED ISSUE #8652: [MU4 Issue] Dynamics shown in wrong size/style when added
**Describe the bug**
Dynamics shown in italics (dynamics codepoints are already styled as italicized so this looks wrong) and shown too small when added to scores.

**To Reproduce**
1. Create a new score
2. Add a dynamic to the first measure
3. Dynamic is shown too small and too slanted

**Expected behavior**
As per MU3 or as per text elements with dynamics symbols inserted.

**Screenshots**

 - Windows 10

**Additional context**
This is because when it applies the "Dynamic" text style to the element it overrides the existing formatting information for the dynamics symbol (using "ScoreText" etc.). MU3 doesn't do this, as `setFontStyle` is just a standard setter for an M_PROPERTY. I'm not sure why MU4 changed the behaviour for setFontStyle/setFamily/setSize but it's definitely quite broken as it is.
One option is to restore the MU3 behaviour of how those properties are treated, including to TextCursor::init(), the required call to cursor.init() in createLayout(), and setting layoutInvalid to true in TextBase::setProperty( ).

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8652

Restore MU3 text style behaviour for text-based elements, ensures that e.g. dynamics don't have their font style information overridden.


src/engraving/libmscore/textbase.cpp
```diff
@@ -1650,6 +1650,7 @@ TextBase::TextBase(Score* s, Tid tid, ElementFlags f)
     setFamily("Edwin");
     setSize(10.0);
     setFontStyle(FontStyle::Normal);
+    _cursor->init();
```
wizofaus, line 1653: MU3 doesn't have this, as doesn't own a TextCursor.

wizofaus:
Can't say I understand why the MusicXML export tests are failing - it's expecting `font-family="Edwin"` but that's being omitted

12: -          <words font-family="Edwin">After system break</words>
12: +          <words>After system break</words>

Actually I can't see why it should ever have been included, given that's the default. Will investigate later.

src/inspector/types/texttypes.h
```diff
@@ -55,52 +55,65 @@ class TextTypes
...
         TEXT_TYPE_SUBTITLE,
         TEXT_TYPE_COMPOSER,
         TEXT_TYPE_POET,
-        TEXT_TYPE_LYRICS_ODD,
-        TEXT_TYPE_LYRICS_EVEN,
-        TEXT_TYPE_FINGERING,
-        TEXT_TYPE_LH_GUITAR_FINGERING,
-        TEXT_TYPE_RH_GUITAR_FINGERING,
-        TEXT_TYPE_STRING_NUMBER,
+        TEXT_TYPE_TRANSLATOR,
```
wizofaus, line 65: technically this is to fix a different issue (https://github.com/musescore/MuseScore/issues/8654), but hard to test current issue without fixing this too

igorkorsukov:
> @igorkorsukov you actually changed testBreaksImplExpl_all_ref.xml recently to add "font-family"="Edwin", was this change expected?

We have legacy styles and some logic for applying them.
Prior to my styling work, these legacy styles were not loaded for tests.
I fixed it, now the legacy styles are loaded and are always applied, including when running tests.
But it turned out that the MusicXML tests are not actual.

Please update from the master, hopefully, the problem should be fixed.

wizofaus:
> Please update from the master, hopefully, the problem should be fixed.

I'm looking at the latest copy from the master branch.   I don't really understand why it would make sense for `"font-family"="Edwin"` to have been added just for this SystemText element (this was in your last commit). But to be honest nothing much about the way that styles are behaving is making sense to me...what *should* be the Font Family for "SystemText" when loading one of the reference scores like testBreaksImplExpl.mscx - Edwin or FreeSerif? (NB the file itself doesn't specify either). When I load it in MU3.6 it actually gives you an option whether to update the styles, if you don't select the option it stays as FreeSerif, if you do it actually appears to update it to "Edwin" internally. But in MU4 it always stays as "FreeSerif" internally (the property is Ms::Sid::systemTextFontFace).

igorkorsukov:
We need your help.
There is a method
```
MStyle* resolveStyleDefaults(const int defaultsVersion)
      {
      switch (defaultsVersion) {
            case LEGACY_MSC_VERSION_V3:
                 return styleDefaults301();
            case LEGACY_MSC_VERSION_V2:
                 return styleDefaults206();
            case LEGACY_MSC_VERSION_V1:
                 return styleDefaults114();
            default:
                 return &MScore::baseStyle();
          }
      }
```

And there is the logic behind applying these legacy styles.
In tests of these legacy styles were not loaded (just because the resources where the style files were located were not initialized).
I fixed it and the styles now loading.
But this required a change in the tests.
How should it be right?

wizofaus:
BTW copying and pasting text elements currently doesn't work properly, I believe due to the changes made in how style/formatting info is serialized/deserialized.

wizofaus:
BTW the alternative way of fixing the original bug (with dynamics being incorrectly styled) is to ensure that when the *style* is being changed, then calls to TextCursor::setFormat( ) *don't* change the formatting of text blocks that already have "ScoreText" as the font family.  But calls outside of this (when the style is not being changed, i.e in direct response to a user action) do need to apply this formatting.  One way this could be easily approximated it just to not only call TextCursor::changeSelectionFormat( ) if there is something selected.  But that's not *quite* right, as a user could select all the text in the element, then change the text style, and it would still (incorrectly) apply the text style's formatting to ScoreText symbols.  Interestingly also if you don't have any text selected and choose a new text style, the full text gets automatically selected - which seems to be due to the existing selectAll() call TextCursor.setFormat(). But instead of doing the selectAll() there we could do it in TextBase::undoChangeProperty, which you *know* is in response to a user event.  Needs a bit more thought/testing though.

wizofaus:
Confirming that removing the calls to selectAll() and changeSelectionFormat() in TextCusor::setFormat( ) work.  
However there's another bug that was introduced with the changes to use the TextCursor to store the format info, which is that when you put the cursor just *before* the ScoreText symbol at the beginning, it picks up the wrong font (it calls "TextCusor::init()" which just has some hard-coded font info).  Really in this case, where there's no TextBlock to get the info from, it needs to be able to get it from somewhere else, e.g. the TextBase owner.  That is in principle possible but requires a bit more mucking around.
The key thing though is to restore the MU3.6 behaviour - when you add a dynamic it's styled correctly (using Leland/right size/no italic), but when you edit it to type text before the dynamic symbol it starts inputting the text in the "Dynamics" text style (which is italic Edwin for new scores).  And if you DO change the text style for the element, it shouldn't affect the format of any musical symbols. Will have another look tomorrow hopefully.

wizofaus:
Ok I did some more experimentation in Mu 3.6:

* create dynamic
* select it, open inspector
* toggle bold: no change in appearance
* toggle italic: no change in appearance
* change font family: no change in appearance
* toggle underline: shows with underline
* change font size: dynamic gets bigger
* save, xml for <Dynamic> is

            <subtype>f</subtype>
            <velocity>96</velocity>
            <family>Courier</family>
            <size>14</size>
            <bold>1</bold>
            <underline>1</underline>

On re-opening dynamic renders as before, and in inspector all changed properties show correctly, i.e.
bold & italic selected despite not being applied.

Now double-click on dynamic, ensure all text is selected

In "Text edit" toolbar the bottom:

Note that Italic & Bold *are* selected, even though not applied to the dynamic symbol (I'd say this is actually a bug).
Deselect bold and *reselect*, and only now dynamic is truly bolded.
Save, xml for <Dynamic> is

            <subtype>other-dynamics</subtype>
            <velocity>64</velocity>
            <text><i><b></b></i><b><font face="ScoreText"/></b></text>
            </Dynamic>

Re-open score and dynamic is displayed as before, and inspector shows properties as before.

On that basis, I believe trying to support storing a text element's font family/size/style in the TextCursor is going to be problematic.
@vpereverzev. Thoughts? I'm happy to keep working on it either way.

vpereverzev:
@wizofaus thanks for your investigation

The intent of these changes in https://github.com/musescore/MuseScore/commit/70745ef6874a82a26adb283e79666ff420bb69d7 was to encapsulate the text formatting functionality because of the new inspector. Previously all that logic was distributed over the various dialogs and mscore.cpp, so we definitely don't want to lose that.

Speaking about the issue, I think that the root of this is that we don't realize that text is ALREADY italic OR we don't have necessary information about it. So I'd definitely encourage to move in that direction.

1) If we have all the necessary information, but we're ignoring it by accidental - we should recover that functionality

2) If we literally don't have any information about that, then we should find out a way of how to make up for it

wizofaus:
> 2. If we literally don't have any information about that, then we should find out a way of how to make up for it

It's specifically the changes you made to TextBase setFontStyle()/setFamily()/setSize() that store this info in the TextCursor, which really only represents "transient" state (i.e. what format is the text that's currently selected, or what format would be used for inserting new text at the cursor position) that's the problem. This PR actually reverses that change and restores it to having actual TextBase fields as per MU3.6, but I'm just concerned you made that change to solve a particular problem.

wizofaus:
Looks like I'm now able to run/debug tests locally quite easily which is awesome! So we need to decide what the correct approach is here...as I said I don't believe storing the whole text element's font properties in the TextCursor is going to work well, as that only represents a transient state. @vpereverzev

wizofaus:
BTW I have a simpler "fix" which works at least for the main problems, which is just to remove the call it _text->selectAll in TextCursor::setFormat, and in TextCursor::set( ), move updateCursorFormat up 2 lines, inside the preceding }.
That allows the TextCursor to "remember" what the font-properties should be for entering new text, without applying them to the existing musical symbols.  However there's still quite a number of issues, and it still "breaks" all the current musicxml tests, because I believe their reference scores are not correct (they contain lots of unnecessary `font-family="Edwin"` xml attributes). @vpereverzev

wizofaus:
BTW the next problem is that undo/redo for changing the whole element's font properties is broken in cases where a mixture of properties are used (e.g. the text is part italic and part non-italic, as is often the case for dynamics such as _poco_ **_f_** etc.
In fact it wasn't great in 3.6 either - where basically if you tried to change something like the "italic" property for such an element, it actually did nothing at all. But if you did change, say, "bold" for such an element, then undo it, it would work correctly - restoring it back to the original format.   In 4.0 this doesn't work, the "undo" logic first gets the original "font-style" property (which holds the bold/italic/underline settings) value, which is returned as -1, and tries to restore that on undo.
The problem is currently when TextBase::undoChangeProperty( ) is called from the properties panel (inspector), there's no way of knowing for sure what the user actually clicked, so no good way of actually determining how to undo that specific action.
Happy to try to come up with a solution but it seems the whole management of text element font styling really needs some more thought/design work.

Tantacrul:
I'd appreciate a suggestion for how we can improve this. We're not quite able to focus on new ideas for this now. As we've mentioned a few times, I'm unhappy with this entire system and I feel a more comprehensive solution would be ideal. For example, I've long been thinking of a visual interface that acts as a kind of 'container' which would hold different text elements with different stylings but would maintain relative positions, etc.

wizofaus:
The current functionality is broken and restoring MU3's functionality (which is far from ideal) is not possible.  So we have to do something...

wizofaus:
I've had a conversation with @Tantacrul about this and now have a pretty good idea of what behaviour is wanted in all cases and I've put my hand up to get it working as best I can (hopefully it won't require changing anything on the QML side of things).
To confirm:
* The correct default font for all styles should be "Edwin", not FreeSerif
* Bold/Italic will never be applied to musical symbols (even if that's how they'd been saved in an MU3 score)
* The properties panel should always show as best as possible what the current selected font face/size/style flags (bold/italic etc.) are, even if multiple elements are selected or text is selected within the element currently being edited (the font/size dropdowns should show -- if values are different, and the style indicators should show as "off" unless ALL selected text or selected elements has the same value).  However I may end up discarding this PR and doing it in a new branch.

Tantacrul:
To confirm, this is what we should see when two different fonts are selected (an indeterminate state)

image

Tantacrul:
@wizofaus - I notice when I add this text that the italics are working as expected... but if I change Bold / Italic / font, it is ignored both for the symbol and also the rest of the text too. Was this by design? For the symbol, the behaviour is correct for sure. However, if I change italics or switch to (say) Arial, I'd expect the 'regular' text to update accordingly, right?

(tested on Mac OS)

image

wizofaus:
Closing this, see https://github.com/musescore/MuseScore/pull/8815

Issue #8632, opened 2021-07-15
[MU4 Issue] Context menu doesn`t closed if click outside the menu on left mouse button


vshalkevich:
**Describe the bug**
Context menu doesn`t closed if click outside the menu on left mouse button
Context menu closes only on right mouse button outside the menu

**To Reproduce**
1. Create a score
2. Click right mouse button on the score 
3. Click Left mouse button on the score (outside the appeared context menu)

**Expected behavior**
Context menu should be closed 

**Screenshots**
https://user-images.githubusercontent.com/71372310/125829846-b0df8770-aa5f-40f6-b793-9736966b37f5.mp4

**Desktop (please complete the following information):**
Linux Mint

Issue #8657, opened 2021-07-19
Context menus aren't fully selecting objects


bkunda:
When an unselected object in the project/score editor is selected with a right-click, the context menu options that appear do not seem to affect the selected object. 

This is notwithstanding the fact that playback of the right-click selected object **does** work.

This video illustrates the issue in greater detail: https://www.dropbox.com/s/5wne1q0rqlp9j5v/Context-menu-issue-1.mp4?dl=0

**Expected behavior**
Right-clicking on any object should simultaneously perform 2 operations:
1. Select the object (i.e. it should become highlighted)
2. Trigger the relevant context menu

Apple MacBook Air M1 running iOS 11.4

wizofaus:
That behaviour is the same as per MU 3.6 though.  I agree it's probably not expected behaviour in most cases, and other apps do tend to behave as you suggest (both selecting the object and showing the context menu). I don't think this is P0 issue though!

bkunda:
I set it as P0 because, in my view, this issue must be fixed before release as the right-click command is a very fundamental system command that currently does not behave as expected (notwithstanding MuseScore 3.6's functionality, which we shouldn't be using as a benchmark in this case). Moreover, it currently does not behave consistently with other apps (and not just music notation apps, but even, for example, your average word processor). This is a matter of ensuring MuseScore feels as "polished" as possible and won't give rise to unpleasant user experiences, particularly at the basic level of ordinary system commands (like right-click). 

I am of course happy for someone to re-classify it as appropriate - I am still fairly new to Github so am very willing to be given some direction if my classifications are off from time to time! 🙂

wizofaus:
Have a look through the list of P1 and P2 bugs - most are far more serious than this,  which has an easy workaround and users have been dealing with for years.  P0 I'd reserve for a critical new problem that prevents users from accessing critical functionality.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

bkunda:
Thanks for this suggestion @wizofaus, I will check this out! 
I will hasten to add, though, that just because users have been managing a workaround for a problem, doesn't mean we shouldn't be trying to fix it! I hope you agree 🙂

wizofaus:
All bugs should be fixed, but we still have to prioritize them :) anyway, that's just my take on things as a volunteer contributor...

RomanPudashkin:
I'm working on this

bkunda:
Thank you @RomanPudashkin

PR #8665, opened 2021-07-19
[MU4] Various fixes for the context menu

RomanPudashkin:
Resolves:
https://github.com/musescore/MuseScore/issues/8632
https://github.com/musescore/MuseScore/issues/8657

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
@@ -136,8 +136,14 @@ StyledPopupView {
             }
         }
 
-        var itemHeight = (view.contentHeight - view.spacing * (model.length - 1)) / model.length
-        root.contentHeight = view.contentHeight - sepCount * (itemHeight - prv.separatorHeight) +
+        var itemHeight = 0
+        for(var child in view.contentItem.children) {
+            itemHeight = Math.max(itemHeight, view.contentItem.children[child].height)
+        }
+
+        var itemsCount = model.length - sepCount
+
+        root.contentHeight = itemHeight * itemsCount + sepCount * prv.separatorHeight +
                 prv.viewVerticalMargin * 2
     }
 
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -314,20 +314,16 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
         return;
     }
 
-    if (m_interactData.hitElement) {
-        if (!m_interactData.hitElement->selected()) {
-            if (event->button() != Qt::MouseButton::RightButton) {
-                SelectType selectType = SelectType::SINGLE;
-                if (keyState == Qt::NoModifier) {
-                    selectType = SelectType::SINGLE;
-                } else if (keyState & Qt::ShiftModifier) {
-                    selectType = SelectType::RANGE;
-                } else if (keyState & Qt::ControlModifier) {
-                    selectType = SelectType::ADD;
-                }
-                m_view->notationInteraction()->select({ m_interactData.hitElement }, selectType, m_interactData.hitStaffIndex);
-            }
+    if (m_interactData.hitElement && !m_interactData.hitElement->selected()) {
+        SelectType selectType = SelectType::SINGLE;
+        if (keyState == Qt::NoModifier) {
+            selectType = SelectType::SINGLE;
+        } else if (keyState & Qt::ShiftModifier) {
+            selectType = SelectType::RANGE;
+        } else if (keyState & Qt::ControlModifier) {
+            selectType = SelectType::ADD;
         }
+        m_view->notationInteraction()->select({ m_interactData.hitElement }, selectType, m_interactData.hitStaffIndex);
     }
 
     if (m_view->notationInteraction()->isHitGrip(logicPos)) {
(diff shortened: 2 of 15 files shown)


src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -314,20 +314,16 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
...
-                } else if (keyState & Qt::ControlModifier) {
-                    selectType = SelectType::ADD;
-                }
-                m_view->notationInteraction()->select({ m_interactData.hitElement }, selectType, m_interactData.hitStaffIndex);
-            }
+    if (m_interactData.hitElement && !m_interactData.hitElement->selected()) {
+        SelectType selectType = SelectType::SINGLE;
+        if (keyState == Qt::NoModifier) {
+            selectType = SelectType::SINGLE;
+        } else if (keyState & Qt::ShiftModifier) {
```
wizofaus, line 321: Does it make sense for shift+right click to do a range selection and ctrl+right click to add to a list selection? @Tantacrul 
That's not what happens in, say, File Explorer in windows - right click always cancel the current selection and just selects the single file you click on.

Tantacrul: I need to test this to get an idea of how this feels. Can't quite comment on what I think is best - despite @wizofaus's comments, which are compelling. I'm also not on top of the collection of related issues. Perhaps @bkunda can advise / test.

Will resume normal services soon. A bit snowed under at the mo.

bkunda: From what I have been able to test here, right-click now correctly selects a new item and automatically triggers the context menu ✅ and left-clicking outside the context menu closes the menu ✅. So far so good 👍🏻

I'm not sure I completely understand all of @wizofaus's comments, particularly around shift+right click v. ctrl+right click. But as far as this is concerned:

"the implementation done here means that if you carefully select a list of items, then right click on one of them to access the context menu, you lose your selection. I don't think that would be expected."

...I was successful in selecting a range of non-adjacent items using ctrl+left click, then right-clicking a selected item (or indeed, right clicking a _new_ item with the ctrl key still pressed) to trigger the context menu, which then affected all of my selected items. This to me is expected behaviour, and it is consistent with how the File Explorer works (actually it's better than how File Explorer works, because File Explorer won't let you add a new item to the list and trigger the context menu at the same time). 

In saying this, I was only able to achieve this operation on single items. Something we seem not to be able to do (in neither 3.6 nor 4), which would be really useful, is to be able to perform the same add-to-list/non-adjacent selection on non-consecutive _ranges_ (i.e. multiple _bars_ of music at a time, not located next to each other). Perhaps this lies outside the scope of this particular issue, but since we're now discussing multiple selections, I wonder if it can be done. @RomanPudashkin?

wizofaus: "Non-consecutive ranges" wouldn't really make sense to me, and I can't see how we could support them sensibly - definitely outside the scope of this PR anyway!
Did you test holding down CTRL (or SHIFT) to select multiple items, then releasing CTRL/SHIFT then right-clicking on one of the selected items?  From what I can see in the code, that will *deselect* all the existing items, which is not how other apps behave.

bkunda: This is what I have managed to doÖ
https://user-images.githubusercontent.com/86290556/126283619-5657f038-e188-48a8-a4d0-e9fc5b9e9096.mov

Of course the clef is not affected (this is consistent with MS3.6 at any rate).

wizofaus: Ah yep...I just re-read the code, you're right, if you right-click on an already selected note it won't change the selection, my bad.

RomanPudashkin: @wizofaus you were right, there is a problem :) The selected area (it is important to use SHIFT for selecting) will be deselected if a user right-clicks on some elements from this area. I prepared a fix in the last commit. @bkunda could you check how the context menu and the elements selecting feel now (especially in cases when CTRL and SHIFT are used for that)?

bkunda: Almost there @RomanPudashkin!. Right-clicking on a selected area (whether it was selected by shift or ctrl) does not deselect the selection anymore. This is now working as expected. 

What isn't now working is: crash when shift+selecting or ctrl+selecting a range, then _right clicking_ outside of that range.   

Interestingly, the crash did **not** occur if the right click was in a different frame (e.g. a vertical frame containing header text).

This crash was repeatable in exactly the same way in both MacOS and Windows builds:

https://user-images.githubusercontent.com/86290556/126325124-abbd4c0d-1fba-4235-8ccd-8cc7626a21da.mov


https://user-images.githubusercontent.com/86290556/126325139-e04b0a6b-a9d3-48e9-94a0-01d337c6f568.mov


https://user-images.githubusercontent.com/86290556/126325149-f254d48f-c0a5-4592-ba83-5a98310f574a.mov

RomanPudashkin: @bkunda thanks for testing! Fixed

Eism (approved):

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledMenu.qml
```diff
@@ -136,8 +136,14 @@ StyledPopupView {
             }
         }
 
-        var itemHeight = (view.contentHeight - view.spacing * (model.length - 1)) / model.length
-        root.contentHeight = view.contentHeight - sepCount * (itemHeight - prv.separatorHeight) +
+        var itemHeight = 0
+        for(var child in view.contentItem.children) {
```
Jojo-Schmitz, line 140: No space between `for` and `(` in qml files, only in cpp ones? Looks like an inconsistent code style to me...

PR #8666, opened 2021-07-19
The style is stored in a separate file (inside mscz)

igorkorsukov:
The ChordList is stored in a separate file (inside mscz), too


wizofaus:
BTW one concern with doing this is that it makes it harder to store "reference" mscx files in github that can be used for unit testing - I particularly like the way the unit current unit tests work where they show you the diff between the mscx output generated and the reference one.
We could partially address this by improving the unit test code to be able handle mscz files too, i.e. decompressing them and doing a file-by-file diff (for binary files you most likely wouldn't want to show every byte that had changed though!).
But you still wouldn't be able to git-diffs between versions of mscz files (at least, I'm not aware of git being able to automatically handle compressed files for this).
I'm wondering if it would be worthwhile to have the ability to have a folder of uncompressed files represent a "score" (or notation project as you've called it elsewhere).

jeetee:
Are there any plans to have a more complete xml-format as well (such as mscx was for mscz in 3.x)? Some people use versioning on their scores, and most versioning systems don't play nice with binary formats. But having to maintain several separate files for a single score would be a big step backwards for those use cases; not to mention the likely filename clashes between style files if one wishes to have multiple mscx files in the same folder.

wizofaus:
Actually looking at the structure of existing mscx and mss files I'm not sure I see the advantage of separating them out, given both have a toplevel `<museScore>` element, then the former has a everything under the `<Score>` element and the latter everything under the `<Style>` element. Is the intention that one mscz file could contain multiple mscx files that all refer to the same mss file?
Even if so, I'd think we'd still want to support cases where the two are stored in a single text-based format, even if it's primarily for development/troubleshooting/testing purposes.

wizofaus:
Also I notice that with the current MU4 build if you try to save a score as mscx, it just ends up as an empty file. Preserving this functionality is vitally important, even if it's not encouraged for regular usage. As a developer I nearly always save my scores in mscx format.

igorkorsukov:
We want to keep all data separate to simplify the system, maintain compatibility easier, and add new ones easier. Requirements and capabilities will only grow, keeping everything in one xml file, we are faced with great complexity.
For example, a function such as notes, comments to a score (and the score elements) can be added further, they will also need to be stored somewhere.
We believe that the best decision would be to do it in the same way as open documents are stored.

On the other hand, I certainly understand what you are talking about.
I can suggest making it so that we can save and write to a folder instead of a zip archive, I think this solves the described problems.

wizofaus:
Can we at least ensure saving to mscx still works,  it's a critical troubleshooting/ development tool.

igorkorsukov:
> Can we at least ensure saving to mscx still works, it's a critical troubleshooting/ development tool.

Unfortunately, we have to abandon the storage in mscx. Today I will make it possible to save to a folder, this solves the same problems as mscx, and even better.

wizofaus:
As long as that includes mscx files that can be compared with those generated by 3.6 that's fine. 
I don't think it needs to be a generally available/advertised feature,  as long as it's still there for those of us that need it (and understand any limitations thereof).

vpereverzev (approved):

Jojo-Schmitz:
> As long as that includes mscx files that can be compared with those generated by 3.6 that's fine.

It won't, not fully, as  parts of the 3.6 mscx are being moved out

wizofaus:
If you can see in a diff that's the only difference that's fine.

Jojo-Schmitz:
Not good for automated tests though

Jojo-Schmitz:
I guess forcing MuseScore 3 to read such a score (i.e. ignoring the warning thatit got created ny a newer version) would still work but loose all style settings, so fall back to their (3.x) defaults?

wizofaus:
> Not good for automated tests though

Having the automated tests be able to compare multiple output files shouldn't be that hard,  but knowing it's safe to update all the reference files to the new format is more of an issue - you could have a process to verify that when the new format file is loaded then output as musicxml the result is the same at least. But making such a big change to the file format at the same time as a large number of other changes is probably my biggest worry.

Jojo-Schmitz:
Then again at a major version bump this is the only time this can be done

Jojo-Schmitz:
Here's a collection of other issues that would requite a major version bump: https://musescore.org/en/node/296654

wizofaus:
Sure but ideally the file format change would have been made first without changing anything else inside libmscore, then you could generate a new set of reference files for the unit tests, and use those to have confidence that further fixes/changes to libmscore hadn't broken anything. But software development rarely follows the ideal path, for all sorts of reasons. As it is I don't even know how to run the tests locally, hence can't realistically write new ones to verify the various changes I've made haven't broken anything and help ensure future changes don't break what I've done.  On that basis most of my PRs should really never have even been approved!

igorkorsukov:
> Sure but ideally the file format change would have been made first without changing anything else inside libmscore, then you could generate a new set of reference files for the unit tests, and use those to have confidence that further fixes/changes to libmscore hadn't broken anything. But software development rarely follows the ideal path, for all sorts of reasons. As it is I don't even know how to run the tests locally, hence can't realistically write new ones to verify the various changes I've made haven't broken anything and help ensure future changes don't break what I've done. On that basis most of my PRs should really never have even been approved!

I believe that changing the file format, and just the method of storage, into separate files is a fairly safe thing, purely technical, just it works or not.
But on the whole, you are right.
This week I will try to pay attention to tests, write documentation on how they work (we have two types of them), and how to run them locally.
I also adapt tests for the new storage format.

wizofaus:
I think I have some idea how they *work*, I just need to know how to run them under the debugger in Visual Studio!  I did manage to get it working with the 3.6 branch, but on master it just wasn't working for me (in fact I couldn't even compile some of the tests due to references to files that don't exist any more etc.)

shoogle:
> it makes it harder to store "reference" mscx files in github

This is also a concern for score repositories like [OpenScore/Lieder](https://github.com/OpenScore/Lieder) where files are [stored as MSCX](https://github.com/OpenScore/Lieder/tree/main/scores/Beach,_Amy/4_Songs,_Op.51/3_Juni)

> I can suggest making it so that we can save and write to a folder instead of a zip archive

That would work for me, and would be great for when there are PNG / SVG images and other resources in the file; those could go in the same project directory (basically just MSCZ as a directory rather than ZIP file).

It might be useful to give the project directory a file extension like `.mscd` (MuseScore Directory), a bit like how application bundles on macOS use the `.app` extension even though they are really just folders.

jeetee:
> "Unfortunately, we have to abandon the storage in mscx. Today I will make it possible to save to a folder, *this solves the same problems as mscx, and even better*."

Not at all. Things a single mscx has in advantage over a folder with files:
1. It allows me to transfer a single file to someone else in a readable format that doesn't get filtered by overzealous spamfilters instead of being forced to send multiple files to someone else when sharing a score.
2. It allows me to save scores into a single folder as an easily browsable list (such as on a file/versioning server). This will likely no longer be possible as each file now has to be buddied with a style file, which likely leads to filename collisions.
3. It allows file-diffing _the entire score_ by external text-diff tools more easily

The advantage the folder approach has over a single mscx:
1. Easier sharing of a single style file between multiple scores

For me, that's a 3 to 1 score against splitting the mscx format.
I can kind of understand wanting to have the *option* to do so and do it as such inside the mscz; but do feel that the single mscx-export should remain an option. So I am curious as to which issues the folder structure is supposed to fix better (feel free to discuss here or on Discord as suits you best).
If all those extracted files have a different root element, then I guess it shouldn't be that hard to have them inlined and that inlining should come at virtually no cost at all?

Jojo-Schmitz:
Hmm, how about leaving mscx 'export' as it way, complete with style and chord symbols (but without images), and not only in test mode, and then only  in mscz format do the splitting? Wouldn't that combine the best of both worlds?

shoogle:
I agree with the points @jeetee raises, though they could potentially be solved by having an option to store the MSCZ zip file with zero compression (this is possible and is allowed by the ZIP standard) so that the plaintext is still readable.

I think there will come a time (MuseScore 5 probably) where styles will be allowed to change _midway_ through a score. This will be necessary to allow for pull-out pages, inserts, and temporarily reduced staff sizes when lots of instruments play at once (@oktophonie will confirm these are real things in notation that we should support in MuseScore). That would necessitate putting style information back into the main score file, but within a `<StyleChange>` element that could occur anywhere in the music, not just in a header region at the top of the file. We would probably need a whole new file format to support those things anyway, so having them separately in MSCX doesn't limit us right now.

igorkorsukov:
> It allows me to transfer a single file to someone else in a readable format that doesn't get filtered by overzealous spamfilters instead of being forced to send multiple files to someone else when sharing a score.

For this, it is better to use the `mscz` format

> It allows me to save scores into a single folder as an easily browsable list (such as on a file/versioning server). This will likely no longer be possible as each file now has to be buddied with a style file, which likely leads to filename collisions.

When saving to a folder, instead of a list of files, there will just be a list of folders, each of which has its own styles and etc.
It is similar to folders with projects (for example C ++ projects :)). We are working on a project, there are many files in the project, e.g. added pictures.

> It allows file-diffing the entire score by external text-diff tools more easily

This will be easier to do, the score file will become smaller, there will be no information that is not directly related to the notation.

> The advantage the folder approach has over a single mscx:

* We can store all files, even such as pictures, audio samples, etc.
* Different files can be in different readable formats (e.g. json)
* We can take some files from other scores, for example styles, chordlist, some settings, etc.
* In the future, we will add new functionality (for example, notes for a score), this data will also need to be stored. If we use one xml file, we will have to enlarge it even more, and then it will be even more difficult to work with it in plain text mode (read, compare, etc.). Or we have to lose this data.
If we keep everything separately, then it is easier to work with each part and nothing will be lost.

igorkorsukov:
> Hmm, how about leaving mscx 'export' as it way, complete with style and chord symbols (but still without images), and not only in test mode, and then only in mscz format do the splitting? Wouldn't that combine the best of both worlds?

I resist this because I'm afraid that it will greatly complicate the code, there will be a lot of combinatorics. Complex code means a lot of bugs and a long development time. This is bad for everyone, both users and developers.
I believe that we need to make it easier and more reliable and by all means, avoid combinatorics or we need to isolate it.

Also, a large number of combinatorics in the code greatly interferes with growth, such a system stops at one time.

Jojo-Schmitz:
rebase needed

igorkorsukov:
For testing purposes, I have added a cmake option ENGRAVING_COMPAT_WRITESTYLE_302 - if enabled, the score will be recorded as before, as 3.02 format .
This is a temporary option, we will support it as long as it makes sense and until it becomes too difficult to do.

PR #8667, opened 2021-07-20
Added read/write to dir

igorkorsukov:
Now we can specify the `mscx` format when saving, in this case a folder will be created in which all files will be written as in `mscz`.
We can also open `mscx`, in this case the system will try to get other files relative to the current one as from  `mscz`.

### DIFF
--- a/src/engraving/io/msczreader.cpp
+++ b/src/engraving/io/msczreader.cpp
@@ -24,6 +24,8 @@
 #include <QXmlStreamReader>
 #include <QFile>
 #include <QFileInfo>
+#include <QDir>
+#include <QDirIterator>
 
 #include "thirdparty/qzip/qzipreader_p.h"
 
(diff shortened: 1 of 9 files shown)


src/engraving/io/msczreader.cpp
```diff
@@ -129,32 +181,73 @@ const MsczReader::Meta& MsczReader::meta() const
...
+
+            for (const MQZipReader::FileInfo& fi : fileInfoList) {
+                if (fi.isFile) {
+                    files << fi.filePath;
+                }
             }
+        } break;
+        case Mode::Dir: {
+            QString root = rootPath();
+            QDirIterator::IteratorFlags flags = QDirIterator::Subdirectories;
```
vpereverzev, line 202: isn't it making sense to use framework::filesystem for that mode?

vpereverzev (approved):

src/engraving/io/msczreader.cpp
igorkorsukov: At the moment, the engraving module does not depend on anything and for now it is better to keep it that way.

PR #8673, opened 2021-07-20
[MU4] Element context

RomanPudashkin:


src/notation/internal/notationinteraction.cpp
```diff
@@ -556,6 +556,8 @@ void NotationInteraction::clearSelection()
     score()->deselectAll();
 
     notifyAboutSelectionChanged();
+
+    selection()->setHitElementContext(INotationSelection::ElementContext());
```
igorkorsukov, line 560: Maybe better add method like `resetElementContext`

src/notation/internal/notationselection.h
```diff
@@ -56,6 +59,7 @@ class NotationSelection : public INotationSelection
 
     IGetScore* m_getScore = nullptr;
     INotationSelectionRangePtr m_range;
+    ElementContext m_hitElementContext;
```
igorkorsukov, line 62: I think better rename `ElementContext` to `HitElementContext`

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -116,6 +116,11 @@ INotationInteractionPtr NotationViewInputController::viewInteraction() const
     return m_view->notationInteraction();
 }
 
+Element* NotationViewInputController::hitElement() const
+{
+    return m_view->notationInteraction()->selection()->hitElementContext().element;
```
igorkorsukov, line 121: Maybe better move `hitElementContext` to `notationInteraction` (not `selection`), because may be hit, but not "selection"

src/notation/internal/notationselection.h
RomanPudashkin: renamed

src/notation/view/notationviewinputcontroller.cpp
RomanPudashkin: moved

src/notation/internal/notationinteraction.cpp
RomanPudashkin: No more need, because setHitElementContext is a part of NotationInteraction now

Eism (approved):

PR #8675, opened 2021-07-20
Added xmlfile mode for mscreader/writer

igorkorsukov:
xml file like
```
<files>
<file name="ff2.mscx">
<![CDATA[
<museScore version="3.02">
  <programVersion>4.0.0</programVersion>
  <programRevision></programRevision>
  <Score>
...
    </Score>
  </museScore>

]]>
</file>
<file name="audiosettings.json">
<![CDATA[
{
    "example": {
        "someValue": 42
    }
}

]]>
</file>
</files>

### DIFF
--- a/src/engraving/io/mscreader.cpp
+++ b/src/engraving/io/mscreader.cpp
@@ -300,20 +300,80 @@ QByteArray MscReader::DirReader::fileData(const QString& fileName) const
 
 bool MscReader::XmlFileReader::open(QIODevice* device, const QString& filePath)
 {
+    m_device = device;
+    if (!m_device) {
+        m_device = new QFile(filePath);
+        m_selfDeviceOwner = true;
+    }
+
+    if (!m_device->isOpen()) {
+        if (!m_device->open(QIODevice::ReadOnly)) {
+            LOGE() << "failed open file: " << filePath;
+            return false;
+        }
+    }
+
+    return true;
 }
 
 void MscReader::XmlFileReader::close()
 {
+    if (m_device) {
+        m_device->close();
+    }
 }
 
 bool MscReader::XmlFileReader::isOpened() const
 {
+    return m_device ? m_device->isOpen() : false;
 }
 
 QStringList MscReader::XmlFileReader::fileList() const
 {
+    if (!m_device) {
+        return QStringList();
+    }
+
+    QStringList files;
+
+    m_device->seek(0);
+    QXmlStreamReader xml(m_device);
+    while (xml.readNextStartElement()) {
+        if ("file" != xml.name()) {
+            xml.skipCurrentElement();
+            continue;
+        }
+
+        QStringRef fileName = xml.attributes().value("name");
+        files << fileName.toString();
+        xml.skipCurrentElement();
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: src/engraving/io/mscreader.cpp [hunk 1]
```diff
@@ -0,0 +1,379 @@
...
+    close();
+}
+
+void MscReader::setParams(const Params& params)
+{
+    IF_ASSERT_FAILED(!isOpened()) {
+        return;
+    }
+
+    if (m_reader) {
```
vpereverzev, line 55: why would we need to delete m_reader here?

If it's really necessary, maybe it makes sense to call "close()" here?

src/engraving/io/mscreader.cpp
```diff
@@ -0,0 +1,379 @@
...
+#include "log.h"
+
+//! NOTE The current implementation resolves files by extension.
+//! This will probably be changed in the future.
+
+using namespace mu::engraving;
+
+MscReader::MscReader(const Params& params)
+    : m_params(params)
+{
```
vpereverzev, line 41: might make sense to just create a reader at this point

src/engraving/io/mscreader.cpp
```diff
@@ -0,0 +1,379 @@
...
+}
+
+MscReader::~MscReader()
+{
+    close();
+}
+
+void MscReader::setParams(const Params& params)
```
vpereverzev, line 49: Maybe it's better to not allow to change parameters of an existing MscReader instance and just leave an explicit constructor MscReader(const Params& params)?

src/engraving/io/mscreader.cpp
igorkorsukov: I think, call `close` is bad idea, and should be already closed at the moment. 
But, we should delete reader, because it may be created, but mode may be changed, so may be needs other reader implementation.

src/engraving/io/mscreader.cpp
```diff
@@ -0,0 +1,379 @@
...
+        m_reader->close();
+
+        delete m_reader;
+        m_reader = nullptr;
+    }
+}
+
+bool MscReader::isOpened() const
+{
+    return m_reader->isOpened();
```
vpereverzev, line 85: might be safer to use "reader()" here

igorkorsukov: I did just that to the writer ...
but I think differently
I'll fix it to:
```
return m_reader ? m_reader->isOpened() : false;
```

src/engraving/io/mscreader.cpp
igorkorsukov: I think no, We can be set params again, I wasn't created reader if not needs. But I check reader on null everywhere

src/engraving/io/mscreader.cpp
igorkorsukov: Maybe, but sometimes it may not be convenient. For example, if we want to make the reader a member of the class and use one instance, setting parameters to it.

vpereverzev (approved):

PR #8676, opened 2021-07-20
[MU4] synthesizers_pipeline_step_2

vpereverzev:
image


src/framework/audio/synthtypes.h
```diff
@@ -47,6 +48,52 @@ enum class SynthType {
...
+struct SynthUri : public UriQuery
+{
+    SynthType type = SynthType::Undefined;
+
+    SynthUri() = default;
+    SynthUri(const SynthType& type)
+        : UriQuery("musescore://audio/" + synthTypeToString(type)), type(type) {}
+
+private:
+    std::string synthTypeToString(const SynthType& type)
```
igorkorsukov, line 77: Strange, two functions `inline std::string synthTypeToString(const SynthType& type)`

vpereverzev: damn, no idea how I've missed this

thanks, fixed

igorkorsukov (approved):

PR #8706, opened 2021-07-23
[MU4] Fix more compiler warnings

Jojo-Schmitz:
as seen in MS VS2019

Now just 11 warnings remaining:
7 in third party code, stb_vorbis.c (6 times C4457, once C4701), probably best disable these warnings there?
One about en empty telemetry.qrc, delete that file?
One (C4505) in moc_DockWidgetQuick.cpp, also third party and on top a generated file, maybe disable that warning?
2 (C4573 about `QObject::disconnect` not being able to gather `this` for in dockframemodel.cpp, lines 99 and 100, no idea what this is about, way above my pay grade ;-)

### DIFF
--- a/src/engraving/libmscore/instrtemplate.cpp
+++ b/src/engraving/libmscore/instrtemplate.cpp
@@ -500,7 +500,7 @@ void InstrumentTemplate::read(XmlReader& e)
             QString traitName = qtrc("InstrumentsXML", e.readElementText().toUtf8().data());
             trait.isDefault = traitName.contains("*");
             trait.isHiddenOnScore = traitName.contains("(") && traitName.contains(")");
-            trait.name = traitName.remove("\*").remove("(").remove(")");
+            trait.name = traitName.remove("*").remove("(").remove(")");
         } else if (tag == "StringData") {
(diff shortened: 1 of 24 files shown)


src/engraving/libmscore/instrtemplate.cpp
```diff
@@ -500,7 +500,7 @@ void InstrumentTemplate::read(XmlReader& e)
             QString traitName = qtrc("InstrumentsXML", e.readElementText().toUtf8().data());
             trait.isDefault = traitName.contains("*");
             trait.isHiddenOnScore = traitName.contains("(") && traitName.contains(")");
-            trait.name = traitName.remove("\*").remove("(").remove(")");
+            trait.name = traitName.remove("*").remove("(").remove(")");
```
Jojo-Schmitz, line 503: I don't think a `\*` should get removed here, but if so it'd need to be `"\\*"`

RobFog:
"fix", not "fox". ;)

RobFog:
Oh, and "warnings", not "wanrings".

igorkorsukov (approved):

PR #8715, opened 2021-07-23
[MU4] Fix compiler warnings (again)

Jojo-Schmitz:
Fixed earlier today, but these 2 seem to have been reverted by a later change.

Also slightly different than before, pventing warnings on macOS/Clang too

### DIFF
--- a/src/palette/internal/palette/palettecreator.cpp
+++ b/src/palette/internal/palette/palettecreator.cpp
@@ -1076,7 +1076,7 @@ PalettePanel* PaletteCreator::newBracketsPalettePanel()
 
     for (size_t i = 0; i < types.size(); ++i) {
         auto b1 = makeElement<Bracket>(gscore);
-        auto bi1 = bracketItemOwner.brackets()[i];
+        auto bi1 = bracketItemOwner.brackets()[static_cast<int>(i)];
         const auto& type = types[i];
         bi1->setBracketType(type.first);
         b1->setBracketItem(bi1);
--- a/src/engraving/draw/painterpath.cpp
+++ b/src/engraving/draw/painterpath.cpp
@@ -147,7 +147,7 @@ size_t PainterPath::elementCount() const
     return m_elements.size();
 }
 
-PainterPath::Element PainterPath::elementAt(int i) const
+PainterPath::Element PainterPath::elementAt(size_t i) const
 {
     assert(i >= 0 && i < elementCount());
     return m_elements.at(i);
--- a/src/engraving/draw/painterpath.h
+++ b/src/engraving/draw/painterpath.h
@@ -95,7 +95,7 @@ public:
 
     bool isEmpty() const;
     size_t elementCount() const;
-    PainterPath::Element elementAt(int i) const;
+    PainterPath::Element elementAt(size_t i) const;
     void addRect(const RectF& r);
     inline void addRect(double x, double y, double w, double h)
     {


src/engraving/libmscore/chordline.cpp
```diff
@@ -230,7 +230,7 @@ void ChordLine::write(XmlWriter& xml) const
     if (modified) {
         size_t n = path.elementCount();
         xml.stag("Path");
-        for (size_t i = 0; i < n; ++i) {
+        for (int i = 0; i < n; ++i) {
```
cbjeukendrup, line 233: For me, this causes a compiler warning about comparing a signed number with an unsigned number (int vs size_t). I think we should rather change PainterPath::elementAt() to use size_t instead of int.

Jojo-Schmitz: Possible, it fixed the warning with MSVC, I didn't cross-check this with MinGW (gcc).

Jojo-Schmitz: Better now?

src/palette/internal/palette/palettecreator.cpp
```diff
@@ -1074,7 +1074,7 @@ PalettePanel* PaletteCreator::newBracketsPalettePanel()
     static Staff bracketItemOwner(gscore);
     bracketItemOwner.setBracketType(static_cast<int>(types.size()) - 1, BracketType::NORMAL);
 
-    for (size_t i = 0; i < types.size(); ++i) {
+    for (int i = 0; i < types.size(); ++i) {
         auto b1 = makeElement<Bracket>(gscore);
         auto bi1 = bracketItemOwner.brackets()[i];
```
Jojo-Schmitz, line 1079: Here we'd need a cast though, or overload the `QList`´s `[]`operator. Or does it go though without a warning for you?

src/engraving/libmscore/chordline.cpp
cbjeukendrup: Yes, this one seems perfect now!

src/palette/internal/palette/palettecreator.cpp
cbjeukendrup: It's problematic that QList etc. use `int` for size and subscripts...
The real solution would be to change the Staff::brackets() to a std::vector instead of a Qt container.
But using a cast would also be okay for now, I think.

Jojo-Schmitz: OK, hold a sec...

Jojo-Schmitz: Another other warnings you're getting?

cbjeukendrup: No, now there's nothing about int/size_t etc. any more for me!

(there are some totally different warnings on macOS, but most are not really about the code, there are some strange linker things...)

src/palette/internal/palette/palettecreator.cpp
```diff
@@ -1076,7 +1076,7 @@ PalettePanel* PaletteCreator::newBracketsPalettePanel()
 
     for (size_t i = 0; i < types.size(); ++i) {
         auto b1 = makeElement<Bracket>(gscore);
-        auto bi1 = bracketItemOwner.brackets()[i];
+        auto bi1 = bracketItemOwner.brackets()[static_cast<int(i)];
```
cbjeukendrup, line 1079: Typo (`>` is missing)

Jojo-Schmitz: I know, fixed

src/palette/internal/palette/palettecreator.cpp
Jojo-Schmitz: Great!

PR #8716, opened 2021-07-24
[MU4] Mixer crashes fix

Eism:
The mixer panel is subscribed to changes tracks in the playback. When adding/deleting a track, a track item is also added/deleted in the model.
The mixer panel is also subscribed to change the current sequence. Changing the current sequence will delete all track items in the model.
In the playback, a signal is sent for each deletion of a track when the sequence was changed. As a result, the track is deleted twice in the model: removeItem and load(clear method)

Removed reloading tracks when changing the sequencer

### DIFF
--- a/src/playback/view/mixerpanelmodel.cpp
+++ b/src/playback/view/mixerpanelmodel.cpp
@@ -31,23 +31,11 @@ using namespace mu::audio;
 MixerPanelModel::MixerPanelModel(QObject* parent)
     : QAbstractListModel(parent)
 {
-    controller()->currentTrackSequenceIdChanged().onNotify(this, [this]() {
-        load();
-    });
 }
 
 void MixerPanelModel::load()
 {
-    TrackSequenceId sequenceId = controller()->currentTrackSequenceId();
-
-    playback()->tracks()->trackIdList(sequenceId)
-    .onResolve(this, [this, sequenceId](const TrackIdList& trackIdList) {
-        loadItems(sequenceId, trackIdList);
-    })
-    .onReject(this, [sequenceId](int errCode, std::string text) {
-        LOGE() << "unable to find track sequence:" << sequenceId << ", error code: " << errCode
-               << ", " << text;
-    });
+    m_mixerChannelList.append(buildMasterChannelItem());
 
     playback()->tracks()->trackAdded().onReceive(this, [this](const TrackSequenceId sequenceId, const TrackId trackId) {
         addItem(sequenceId, trackId);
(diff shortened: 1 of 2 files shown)


src/playback/view/mixerpanelmodel.cpp
```diff
@@ -111,9 +83,11 @@ void MixerPanelModel::removeItem(const audio::TrackId trackId)
...
     beginResetModel();
 
-    (void)std::remove_if(m_mixerChannelList.begin(), m_mixerChannelList.end(), [&trackId](const MixerChannelItem* item) {
-        return item->id() == trackId;
-    });
+    for (int i = m_mixerChannelList.count() - 1; i >= 0; i--) {
+        if (m_mixerChannelList[i]->id() == trackId) {
+            m_mixerChannelList.removeAt(i);
+        }
+    }
```
wizofaus, line 90: Any reason not to use `remove_if / m_mixelChannel.erase( )`? (it works fine as far as I can tell). But this is definitely better than just using remove_if without erase.
Do we need to worry about deleting the removed elements though? Qt will clean them up when the MixelPanelModel is deleted, but I that only happens when you shut down, so there could still be a memory leak each time you change scores here.

src/playback/view/mixerpanelmodel.cpp
```diff
@@ -31,23 +31,11 @@ using namespace mu::audio;
 MixerPanelModel::MixerPanelModel(QObject* parent)
     : QAbstractListModel(parent)
 {
-    controller()->currentTrackSequenceIdChanged().onNotify(this, [this]() {
```
wizofaus, line 34: How does MixerPanelModel deal with switching between scores now?

wizofaus:
Tried this out and a) still crashes inside Qt when creating a new score from the Symphony Orchestra template
b) when switching between full score and parts I had a crash in Ms::paintElements (elements.cpp, line 1394)

`for (const Element* element : sortedElements)`

 	ucrtbased.dll!00007ffafa3fafec()	Unknown
 	ucrtbased.dll!00007ffafa3faf37()	Unknown
 	MuseScore4.exe!std::_Adl_verify_range<class QList<class Ms::Element *>::iterator,class QList<class Ms::Element *>::iterator>(class QList<class Ms::Element *>::iterator const &,class QList<class Ms::Element *>::iterator const &)	Unknown
 	MuseScore4.exe!std::_Get_unwrapped<class QList<class Ms::Element *>::iterator const &>(class QList<class Ms::Element *>::iterator const &)	Unknown
 	MuseScore4.exe!std::_Seek_wrapped<class std::_Vector_iterator<class std::_Vector_val<struct std::_Simple_types<class Ms::Element *> > >,class Ms::Element * *>(class std::_Vector_iterator<class std::_Vector_val<struct std::_Simple_types<class Ms::Element *> > > &,class Ms::Element * * &&)	Unknown
 	MuseScore4.exe!QtPrivate::qMakeForeachContainer<class QList<class Ms::Element *> &>(class QList<class Ms::Element *> &)	Unknown
 	**MuseScore4.exe!Ms::paintElements(class mu::draw::Painter &,class QList<class Ms::Element *> const &)	Unknown**
>	MuseScore4.exe!mu::notation::Notation::paintPages(mu::draw::Painter * painter, const mu::RectX<double> & frameRect, const QList<Ms::Page *> & pages, bool paintBorders) Line 329	C++
 	MuseScore4.exe!mu::notation::Notation::paint(mu::draw::Painter * painter, const mu::RectX<double> & frameRect) Line 300	C++
 	MuseScore4.exe!mu::notation::NotationPaintView::paint(QPainter * qp) Line 355	C++

wizofaus:
I was able to avoid that crash using a different part, but then it crashed in AbstractInvoker::invokeCallback - the callback.receiver was definitely corrupted/probably deleted, it appeared to be mu::audio::MidiAudioSource receiver.
But even when not crashing the mixer wasn't correctly reconfiguring itself as you switched between tabs, often just showing the master slider and nothing else.

Eism:
Merged another solution #8732

PR #8721, opened 2021-07-26
[MU3] Fix 304642#: inspector update on subtype change

lyrra:
- its not enough to only track element-type change when
  to consider if update inspector.
- use Qt deleteLater which is more safe if we have
  signals roaming to old deleted widgets.

Resolves: https://musescore.org/en/node/304642

The bug that, if you select a slur and tie and click in inspector leads to crash, Is fixed with this commit.

### DIFF
--- a/mscore/inspector/inspector.cpp
+++ b/mscore/inspector/inspector.cpp
@@ -120,6 +120,7 @@ Inspector::Inspector(QWidget* parent)
       ie             = 0;
       oe             = 0;
       oSameTypes     = true;
+      oSameSubtypes  = true;
       _score         = 0;
 //      retranslate();
       setWindowTitle(tr("Inspector"));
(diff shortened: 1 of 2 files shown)


wizofaus:
NB as far as I can tell this shouldn't be an issue in MU4 as all this code has been completely rewritten.  But if there is to be a 3.6.x bug fix release, it would make sense to have this in it.

Jojo-Schmitz:
For some strange reason the macOS build fails: https://github.com/musescore/MuseScore/pull/8721/checks?check_run_id=3164918161#step:5:48
Doesn't seem to be this PR's fault.

Jojo-Schmitz (approved):
Looks good to me and indeed does fix the crash.
Can't find any negative side effects.

lyrra:
Closing, included in PR #9000

mscore/inspector/inspector.cpp
```diff
@@ -324,6 +327,7 @@ void Inspector::update(Score* s)
                               else
                                     ie = new InspectorBreak(this);
 #endif
+                              //FIX: it looks as we might end up with ie = 0
```
Jojo-Schmitz, line 330: See below

mscore/inspector/inspector.cpp
```diff
@@ -399,8 +403,13 @@ void Inspector::update(Score* s)
                               break;
                         }
                   }
+            Q_ASSERT(ie);
```
Jojo-Schmitz, line 406: this indeed makes MuseScore crash on selecting any layoutbreak but section breaks, when it is a Debug build.
Simply doing
```
      if (ie)
            return;
```
should fix this

PR #8748, opened 2021-07-28
[MU4] Fix (MSVC) compiler warnings

Jojo-Schmitz:


src/engraving/libmscore/masterscore.cpp
```diff
@@ -529,7 +529,7 @@ Score::FileError MasterScore::read(XmlReader& e, bool ignoreVersionError, mu::en
                 ++counter;
 
                 if (style().defaultStyleVersion() == 400) {
-                    int k =1;
+                    //int k =1;
```
igorkorsukov, line 532: Oh oh (

Jojo-Schmitz: Well, commenting it out is really just band aid here ;-)

Jojo-Schmitz: It is all but 1 of 97 warnings, the other 96 stem from the `class` vs. `struct` thing

PR #8754, opened 2021-07-29
[MU4] fix beaming issue

wizofaus:
Resolved beaming issue that was breaking a unit test

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -74,19 +74,6 @@ struct BeamFragment {
 Beam::Beam(Score* s)
     : Element(s)
 {
-    _direction       = Direction::AUTO;
-    _up              = true;
-    _distribute      = false;
-    _userModified[0] = false;
-    _userModified[1] = false;
-    _grow1           = 1.0;
-    _grow2           = 1.0;
-    _beamDist        = 0.;
-    _id              = 0;
-    minMove          = 0;
-    maxMove          = 0;
-    _isGrace         = false;
-    _cross           = false;
     initElementStyle(&beamStyle);
 }
 
@@ -267,9 +254,7 @@ void Beam::draw(mu::draw::Painter* painter) const
 
 bool Beam::isNoSlope() const
 {
-    PointF currentBeamPos = beamPos();
-
-    return qFuzzyCompare(currentBeamPos.x(), currentBeamPos.y());
+    return _userModified[0] && qFuzzyCompare(beamPos().first, beamPos().second);
 }
 
 //---------------------------------------------------------
@@ -278,17 +263,17 @@ bool Beam::isNoSlope() const
 
 void Beam::alignBeamPosition()
 {
-    PointF currentBeamPos = beamPos();
+    BeamPos currentBeamPos = beamPos();
 
-    qreal currentX = currentBeamPos.x();
-    qreal currentY = currentBeamPos.y();
+    qreal currentY1 = currentBeamPos.first;
+    qreal currentY2 = currentBeamPos.second;
 
-    qreal maxValue = qMax(qAbs(currentX), qAbs(currentY));
+    qreal maxValue = qMax(qAbs(currentY1), qAbs(currentY2));
 
-    if (qFuzzyCompare(qAbs(currentX), maxValue)) {
-        setBeamPos(PointF(currentX, currentX));
+    if (qFuzzyCompare(qAbs(currentY1), maxValue)) {
+        setBeamPos(qMakePair(currentY1, currentY1));
     } else {
-        setBeamPos(PointF(currentY, currentY));
+        setBeamPos(qMakePair(currentY2, currentY2));
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/engraving/libmscore/beam.cpp
```diff
@@ -74,19 +74,6 @@ struct BeamFragment {
 Beam::Beam(Score* s)
     : Element(s)
 {
-    _direction       = Direction::AUTO;
```
wizofaus, line 77: initialized on declaration

src/engraving/libmscore/beam.cpp
```diff
@@ -267,9 +254,7 @@ void Beam::draw(mu::draw::Painter* painter) const
 
 bool Beam::isNoSlope() const
 {
-    PointF currentBeamPos = beamPos();
-
-    return qFuzzyCompare(currentBeamPos.x(), currentBeamPos.y());
+    return _userModified[0] && qFuzzyCompare(beamPos().first, beamPos().second);
```
wizofaus, line 257: the key fix - without checking _userModified[0] it was assuming beams must be always stay horizontal just because they had been previously

src/engraving/libmscore/beam.cpp
```diff
@@ -278,17 +263,17 @@ bool Beam::isNoSlope() const
 
 void Beam::alignBeamPosition()
 {
-    PointF currentBeamPos = beamPos();
+    BeamPos currentBeamPos = beamPos();
```
wizofaus, line 266: the use of PointF here was beyond comprehension - using `QPair<qreal, qreal>` isn't perfect but it needs to be a type that works with QVariant.

src/engraving/libmscore/element.cpp
```diff
@@ -1377,18 +1377,18 @@ void paintElements(mu::draw::Painter& painter, const QList<Element*>& elements)
...
                 return true;
-            } else if (!e2->visible()) {
-                return false;
             }
 
             return e1->track() > e2->track();
         }
 
-        return e1->z() <= e2->z();
+        return e1->z() < e2->z();
```
wizofaus, line 1391: unrelated change but without it was getting lots of asserts about the predicate being passed to std::sort( ) was invalid, as it didn't consistently return a < b when b > a.

src/engraving/tests/tst_exchangevoices.cpp
```diff
@@ -135,7 +135,6 @@ void TestExchangevoices::undoChangeVoice()
     score->startCmd();
     score->changeSelectedNotesVoice(1);
     score->endCmd();
-    QSKIP("TODO: Need to confirm beam length difference here");
     QVERIFY(saveCompareScore(score, writeFile1, reference1));
```
wizofaus, line 138: actually I'd prefer to add some more tests specifically around beaming, and I do have some prepared but they need a bit more work

src/engraving/libmscore/beam.h
```diff
@@ -48,28 +50,27 @@ class Beam final : public Element
...
-    bool _up;
-    bool _distribute;                     // equal spacing of elements
-    bool _noSlope;
+    bool _up                { true };
+    bool _distribute        { false };                    // equal spacing of elements
 
-    bool _userModified[2];                // 0: auto/down  1: up
-    bool _isGrace;
-    bool _cross;
+    bool _userModified[2]   { { false } };                // 0: auto/down  1: up
```
Jojo-Schmitz, line 58: dropping the extra pair of braces should fix the build issue on Linux.
It only sets the 1st element to `false`, but then implicitly initializes the 2nd to `0`, which incidentially is `false`;-)

src/engraving/libmscore/beam.cpp
Jojo-Schmitz: What is beyond comprehension there?

src/engraving/libmscore/beam.cpp
```diff
@@ -2451,25 +2436,24 @@ Element* Beam::drop(EditData& data)
...
 {
     if (fragments.empty()) {
-        return PointF(0.0, 0.0);
+        return BeamPos{ 0.0, 0.0 };
     }
     BeamFragment* f = fragments.back();
     int idx = (_direction == Direction::AUTO || _direction == Direction::DOWN) ? 0 : 1;
     qreal _spatium = spatium();
-    return PointF(f->py1[idx] / _spatium, f->py2[idx] / _spatium);
+    return BeamPos( f->py1[idx] / _spatium, f->py2[idx] / _spatium );
```
Jojo-Schmitz, line 2449: the leading and trailing space inside the parens is violating code style ;-)

src/engraving/libmscore/beam.h
```diff
@@ -48,28 +50,27 @@ class Beam final : public Element
     Q_GADGET
     QVector<ChordRest*> _elements;          // must be sorted by tick
     QVector<mu::LineF*> beamSegments;
-    Direction _direction;
+    Direction _direction    { Direction::AUTO};
```
Jojo-Schmitz, line 53: Here a trailing space is need to match the desired code style

src/engraving/libmscore/beam.cpp
wizofaus: Because it's two *y* values not x & y.

Jojo-Schmitz: Ah! Now the penny dropped.

src/engraving/libmscore/beam.h
Jojo-Schmitz: Before the brace

wizofaus:
Does this need someone else to review? @vpereverzev

RomanPudashkin (approved):

PR #8760, opened 2021-07-29
Added project auto save 

igorkorsukov:
I do not know if something is needed more complicated, I did it simply, further experience will show whether it is necessary to complicate

### DIFF
--- a/src/project/internal/projectautosaver.cpp
+++ b/src/project/internal/projectautosaver.cpp
@@ -1,0 +1,62 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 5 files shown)


src/project/internal/projectautosaver.cpp
```diff
@@ -0,0 +1,62 @@
...
+        return;
+    }
+
+    if (project->created().val) {
+        LOGD() << "[autosave] project just created";
+        return;
+    }
+
+    if (!project->needSave().val) {
+        LOGD() << "[autosave] project not need save";
```
Jojo-Schmitz, line 51: Typo: `does not` and probably `saving`

src/project/internal/projectautosaver.cpp
```diff
@@ -0,0 +1,62 @@
...
+        return;
+    }
+
+    Ret ret = project->save();
+    if (!ret) {
+        LOGE() << "[autosave] failed save project, err: " << ret.toString();
```
Jojo-Schmitz, line 57: Typoe: `to save`

src/project/internal/projectautosaver.cpp
```diff
@@ -0,0 +1,62 @@
...
+        return;
+    }
+
+    LOGD() << "[autosave] success project saved";
```
Jojo-Schmitz, line 61: Typoe: `successfully saved project` or `project saved successfully ` or `success saving project`

src/project/internal/projectautosaver.cpp
igorkorsukov: Thanks, fixed!

src/project/internal/projectautosaver.cpp
igorkorsukov: Thanks, fixed!

src/project/internal/projectautosaver.cpp
igorkorsukov: Thanks, fixed!

RomanPudashkin (approved):

PR #8447, opened 2021-06-28
[MU3 Backend] ENG-9: Limit copyright info to first page

iveshenry18:
Resolves: [ENG-9](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-9): Copyright info appears on every page

Currently, MuseScore displays copyright metadata in a centered footer by default.
This results in the information being displayed on every page, which
is inconsistent with many published scores.

This commit hides the footers on XML import and rather creates a VBox,
populates it with the copyright metadata, and places it at the bottom
of the first page (either below the first explicit pagebreak, or below
the last system of the first page after an initial layout).

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #8763, opened 2021-07-29
[MU3 Backend] ENG-91: Fix parts with excessive hidden staves (and related issues)

iveshenry18:
Resolves: [ENG-91](https://mu--se.atlassian.net/jira/software/projects/ENG/boards/21?selectedIssue=ENG-91): Excessive hidden staves cause note import to fail

Dolet sometimes exports scores with up to 10 staves, most of which are
hidden. Since MuseScore has a MAX_STAFF limit of 4, such scores are
corrupted (and many other issues are introduced when there are parts
with greater than 2 staves).

This commit adds a mechanism for discarding these hidden staves in such
cases. In doing so, it also must implement a mapping from MusicXML staff
number to index in the Part's Staff list. Therefore, this commit also
adds a function to MusicXmlPart that applies this mapping if present,
or defaults to decrementing the value, and uses this function wherever
the MusicXML staff number is read.

In addition to this mechanism, this PR contains a commit that increases the 
MAX_STAVES constant from 4 to 10. This particular commit probably ought
NOT be ported to master.

Also adds handling for staff-size, which is relevant to many scores with hidden staves
(where hidden staves are for ornament clarification).

Also, this fixes the handling of some corner cases involving part-group barline spanning and brackets/braces.
The spec is somewhat ambiguous about some of the defaults here, but these adjustments are
sensible and produce better results in many complicated scores.

Also, this adds a fix for default \<pedal\> handling, adding a Pedal element to the score
rather than Symbol elements, which fixes layout and playback issues.

Finally, this is built on the commit from #8447.


Jojo-Schmitz:
rebase needed

oktophonie (approved):

Jojo-Schmitz:
This one will most probably be particularly difficult to port to master.

iveshenry18:
> This one will most probably be particularly difficult to port to master.

It will certainly be a challenge, and there are commits that are really for this branch only. I tried to separate the commits such that they could somewhat be cherry-picked, but many of them are interconnected.

PR #8765, opened 2021-07-29
Enable Unity Build for Engraving module

cbjeukendrup:
It appeared that a `#ifndef MU_..._H` guard was missing, and for the rest, only the read114.cpp and read206.cpp files were conflicting with each other, so I enabled the Unity Build for the whole module and excluded only those critical files. 
So, hopefully we get a little faster builds now :)


cbjeukendrup:
Ah, the Windows build is failing. Again because of the same problem that Ashwath is having in his GSoC PR: in some Windows header, there is a `#define small char`. So when we have a function `small()`, it goes wrong. 

It might be better anyway to rename `small()` functions to `isSmall()` or something even more descriptive, but I will first look for a real solution to the problem.

Jojo-Schmitz:
IMHO the real solution is indeed to change `small()` to `isSmall()`
And that header guard is need in any case, it is the RightThing(tm), not just for Unity builds ;.)

Jojo-Schmitz:
Best if we could get rid of that `#undef STRING_NONE` too

Jojo-Schmitz:
Here we could get away with an `#undef small` in accidental.h.

Jojo-Schmitz:
I do get a warning:
```
Two or more files with the name of chordlist.cpp will produce outputs to the same location. This can lead to an incorrect build result.  The files involved are ...\src\engraving\libmscore\chordlist.cpp, ...\src\engraving\compat\chordlist.cpp.	
```

Jojo-Schmitz:
Exclsing those 2 from Unity results in compiler errors:
```
20>.../src/engraving/libmscore/textlinebase.cpp(469,38): error C2371: 'Ms::pids': redefinition; different basic types (compiling source file ...\msvc.build_x64\src\engraving\CMakeFiles\engraving.dir\Unity\unity_12_cxx.cxx)
20>.../src/engraving/libmscore/textbase.cpp(2396): message : see declaration of 'Ms::pids' (compiling source file ...\msvc.build_x64\src\engraving\CMakeFiles\engraving.dir\Unity\unity_12_cxx.cxx)
```

Jojo-Schmitz:
So renaming at least one of those is due. With that it finally builds

Jojo-Schmitz:
Complete (re)build in 1 minute on my machine ;-)
Without that change it takes 10 seconds longer.

Jojo-Schmitz:
That's an option too, but won't solve the warning I mentioned above, and, once fixed, the naming conflicts I also mentioned

cbjeukendrup:
I didn't like just putting `#undef small` in one place so that it just works (until it breaks again); I wanted to know what was going on exactly. So I put `#undef small` below every #include <windows.h>, in the whole project. And even that didn't solve the problem. It seems that it also comes via some Qt headers. 

So I decided not to pollute the code with `#undef small`, and renamed the `small` functions and vars. That's indeed quite a lot of changes, but not too many in my opinion. 

I did not encounter a warning or error about chordlist.cpp however, and can't seem to find anything about it in the CI build logs either. 

On my two-core (!) Windows laptop, a full build (including unit tests) takes still almost 10 minutes (I don't know how long it takes without unity). But it's great to hear that it gives good results for you; that's in fact a 14% reduction.

Jojo-Schmitz:
OK, let's see what it does on my machine once merged. If I then still see that warning about the 2 chordlist.cpp files, I know what to do ;-)

Jojo-Schmitz:
Oh, and I don't do the unit tests (haven't figured out how to, so I let GitHub CI do it for me ;-)

Jojo-Schmitz:
Remains the issue with `#undef STRING_NONE`

cbjeukendrup:
> Remains the issue with `#undef STRING_NONE`

True, I'll fix that.

cbjeukendrup:
By the way, while looking at the CI log, I saw that there was a mistake in build_ninja.sh, so I fixed that too.

src/plugins/api/elements.h
```diff
@@ -168,7 +168,7 @@ class Element : public Ms::PluginAPI::ScoreElement
     API_PROPERTY_T(bool,   visible,       VISIBLE)
     /** Stacking order of this element */
     API_PROPERTY_T(int,    z,             Z)
-    API_PROPERTY(small,                   SMALL)
+    API_PROPERTY(isSmall,                 SMALL)
```
Jojo-Schmitz, line 171: This seems not needed and might break plugins using `small`

igorkorsukov:
It seems, windows.h was in one place 
https://github.com/musescore/MuseScore/pull/8785

Jojo-Schmitz:
> It seems, windows.h was in one place
> #8785

Maybe, but I don't think this relates to this PR here. It is in main.cpp too (and a few other places, like filesystem.cpp)

igorkorsukov:
> > It seems, windows.h was in one place
> > #8785
> 
> Maybe, but I don't think this relates to this PR here. It is in main.cpp too (and a few other places, like filesystem.cpp)

unit build does not combine all project files, but only files of one target (module)

Jojo-Schmitz:
Still, as long as the conflicting `small` and `STRING_NONE` are not used (or `#undef`d), those shouldn't cause any trouble

igorkorsukov:
Another solution, we can put `fixed_windows.h` somewhere and only include it :)
fixed_windows.h
```
#ifndef MU_FIXED_WINDOWS_H
#define MU_FIXED_WINDOWS_H

#include <windows.h>
#undef small
#undef STRING_NONE
#undef WARNING
#undef INFO
#undef ERROR
#undef QUESTION
#undef FILE_OPEN
...

#endif // MU_FIXED_WINDOWS_H
```

src/plugins/api/elements.h
Jojo-Schmitz: It does cause an mtest failure (in 3.x):
```
52/62 Test #53: tst_scripting ....................***Failed    1.09 sec
This plugin does not support createPlatformOpenGLContext!
--- /home/runner/work/MuseScore/MuseScore/mtest/scripting/p1.log.ref	2021-08-03 11:40:32.524648773 +0000
+++ p1.log	2021-08-03 12:05:47.534379914 +0000
@@ -4,13 +4,13 @@
 found:TimeSig (24) at 0
 found:Chord (93) at 0
   beamMode:0
-  small:false
+  small:
   stemDirection:AUTO
   duration:1/4
 found:Rest (25) at 480
 found:Chord (93) at 960
   beamMode:0
-  small:false
+  small:
   stemDirection:AUTO
   duration:1/2
 found:BarLine (12) at 1920
   <diff -u /home/runner/work/MuseScore/MuseScore/mtest/scripting/p1.log.ref p1.log failed

```

cbjeukendrup: Okay, I'll undo this change! (and other plugin API related ones, if any)

Jojo-Schmitz:
Rebase needed too (meanwhile)

cbjeukendrup:
> Rebase needed too (meanwhile)

I'm working on it (but I'm having some technical trouble at the moment...)

src/engraving/libmscore/stringdata.cpp
```diff
@@ -427,11 +427,11 @@ int StringData::fret(int pitch, int string, int pitchOffset) const
 {
     int strings = stringTable.size();
     if (strings < 1) {                          // no strings at all!
-        return FRET_NONE;
+        return INVALID_FRET_INDEX;
```
Jojo-Schmitz, line 430: You forgot to addjust the comment above

Jojo-Schmitz: same for `StringData::fret()` further up

cbjeukendrup: Thanks, I'll fix it!

RomanPudashkin (approved):

Jojo-Schmitz:
As mentioned earlier this PR **does** result in an MSCV compiler warning, even after merge:

    MSB8027	Two or more files with the name of chordlist.cpp will produce outputs to the same location. This can lead to an incorrect build result.  The files involved are 
    C:\Users\Jojo\Documents\GitHub\MuseScore\src\engraving\libmscore\chordlist.cpp, 
    C:\Users\Jojo\Documents\GitHub\MuseScore\src\engraving\compat\chordlist.cpp.

Jojo-Schmitz:
There's anothewr compiler warning, from MinGW, unrelated to this PR, but related to another change from @cbjeukendrup 👍 

```
...\src\framework\ui\internal\platform\windows\windowsplatformtheme.cpp:44: Warnung: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'fnRtlGetNtVersionNumbers' {aka 'void (*)(long unsigned int*, long unsigned int*, long unsigned int*)'} [-Wcast-function-type]
In file included from ...\build.qtc\src\framework\ui\CMakeFiles\ui.dir\Unity\unity_0_cxx.cxx:13:
.../src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp: In constructor 'mu::ui::WindowsPlatformTheme::WindowsPlatformTheme()':
.../src/framework/ui/internal/platform/windows/windowsplatformtheme.cpp:44:126: warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'fnRtlGetNtVersionNumbers' {aka 'void (*)(long unsigned int*, long unsigned int*, long unsigned int*)'} [-Wcast-function-type]
         = reinterpret_cast<fnRtlGetNtVersionNumbers>(GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlGetNtVersionNumbers"));
                                                                                                                              ^
```
No idea what to do about this?

cbjeukendrup:
I wonder why I have never seen that chordlist warning. Which generator do you use, Ninja or Visual Studio? I use Ninja, so that might explain the difference.

I'll try to fix the MinGW warning.

Jojo-Schmitz:
VS not Ninja

PR #8767, opened 2021-07-30
[MU4] Fix (MSVC) compiler warning

Jojo-Schmitz:


src/palette/view/widgets/palettewidget.cpp
```diff
@@ -107,7 +107,7 @@ const std::vector<PaletteCellPtr>& PaletteWidget::actualCellsList() const
 
 int PaletteWidget::actualCellCount() const
 {
-    return m_isFilterActive ? m_filteredCells.size() : cells().size();
+    return static_cast<int>(m_isFilterActive ? m_filteredCells.size() : cells().size());
```
wizofaus, line 110: better to change actualCellCount() to return size_t?

Jojo-Schmitz: Maybe, but that requires a lot of changes in lots of other places (I tried it)

PR #8772, opened 2021-08-01
[MU4] fix MU3 issue #323402, accidentals not shown as expected for altered unisons

wizofaus:
Resolves: https://musescore.org/en/node/323402

Altered unisons where the same note occurs twice in one chord but with different accidentals need to ensure that the accidentals for both notes are always correctly shown (currently they appear and disappear somewhat randomly)

### DIFF
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -2322,6 +2322,16 @@ bool Note::dotIsUp() const
     }
 }
 
+static bool hasAlteredUnison(Note* note)
+{
+    const auto& chordNotes = note->chord()->notes();
+    AccidentalVal accVal = tpc2alter(note->tpc());
+    int relLine = absStep(note->tpc(), note->epitch());
+    return std::find_if(chordNotes.begin(), chordNotes.end(), [note, accVal, relLine](Note* n) {
+        return n != note && !n->hidden() && absStep(n->tpc(), n->epitch()) == relLine && tpc2alter(n->tpc()) != accVal;
+    }) != chordNotes.end();
+}
+
 //---------------------------------------------------------
 //   updateAccidental
 //    set _accidental and _line depending on tpc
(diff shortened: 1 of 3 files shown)


src/engraving/libmscore/note.cpp
```diff
@@ -2354,6 +2363,10 @@ void Note::updateAccidental(AccidentalState* as)
             } else if (acci == AccidentalType::NONE) {
                 acci = AccidentalType::NATURAL;
             }
+        } else if (chordHasAnotherAccidentalForSameNote(this)) {
```
wizofaus, line 2366: potentially this condition could be just be added to the `if` above, but hard to be confident that won't have other consequences

src/engraving/tests/tst_note.cpp
```diff
@@ -513,6 +514,18 @@ void TestNote::tpcDegrees()
     //QCOMPARE(tpc2degree(Tpc::TPC_B_S, Key::C_S), 7);
 }
 
+void TestNote::alteredUnison()
+{
+    MasterScore* score = readScore(NOTE_DATA_DIR + "altered-unison.mscx");
+    Measure* m = score->firstMeasure();
+    Chord* c = m->findChord(Fraction(0, 1), 0);
+    QVERIFY(c->downNote()->accidental() && c->downNote()->accidental()->accidentalType() == Ms::AccidentalType::FLAT);
+    QVERIFY(c->upNote()->accidental() && c->upNote()->accidental()->accidentalType() == Ms::AccidentalType::NATURAL);
```
wizofaus, line 523: Would have preferred to use QCOMPARE however this is crashing for me using MSVC debug C runtime, when it tries to `delete` the strings returned by QTest::toString( ). No idea why, it all looks fine stepping through it.

src/engraving/libmscore/note.cpp
```diff
@@ -2322,6 +2322,16 @@ bool Note::dotIsUp() const
...
 }
 
+static bool hasAlteredUnison(Note* note)
+{
+    const auto& chordNotes = note->chord()->notes();
+    AccidentalVal accVal = tpc2alter(note->tpc());
+    int relLine = absStep(note->tpc(), note->epitch());
+    return std::find_if(chordNotes.begin(), chordNotes.end(), [note, accVal, relLine](Note* n) {
+        return n != note && !n->hidden() && absStep(n->tpc(), n->epitch()) == relLine && tpc2alter(n->tpc()) != accVal;
+    }) != chordNotes.end();
```
wizofaus, line 2332: i.e. look for a note on the same A-G step (including octave), but that would require a different accidental.

wizofaus:
Does this need somebody else to review/approve @igorkorsukov

igorkorsukov (approved):

Issue #8782, opened 2021-08-02
[MU4 Issue] Opening mscx file opens wrong score


wizofaus:
**Describe the bug**
When you try to open an mscx file from a directory containing other mscx files, it opens the first of those alphabetically instead of the one you actually chose.

**To Reproduce**
1. Save/copy a.mscx and b.mscx (with different contents) into a directory
2. File|Open b.mscx
3. a.mscx is opened instead!

**Expected behavior**
Should open the file you select

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**

mcsIoModeBySuffix in mscio.h seems to assume if you open an mscx file it should use the whole directory as its source of files, even for the score itself.
Adding 
```
    if (params().filePath.endsWith(".mscx")) {
        return fileData(QFileInfo(params().filePath).fileName());
    }
```

to MscReader::readScoreFile() is the simplest fix I can find.

PR #8787, opened 2021-08-03
Fixed read .mscx

igorkorsukov:
Resolved https://github.com/musescore/MuseScore/issues/8782

### DIFF
--- a/src/engraving/io/mscreader.cpp
+++ b/src/engraving/io/mscreader.cpp
@@ -119,13 +119,16 @@ QByteArray MscReader::readStyleFile() const
 
 QByteArray MscReader::readScoreFile() const
 {
-    QString mscxFileName;
-    QStringList files = reader()->fileList();
-    for (const QString& name : files) {
-        // mscx file in the root dir
-        if (!name.contains("/") && name.endsWith(".mscx")) {
-            mscxFileName = name;
-            break;
+    QString mscxFileName = QFileInfo(m_params.filePath).completeBaseName() + ".mscx";
+    QByteArray data = fileData(mscxFileName);
+    if (data.isEmpty() && reader()->isContainer()) {
+        QStringList files = reader()->fileList();
+        for (const QString& name : files) {
+            // mscx file in the root dir
+            if (!name.contains("/") && name.endsWith(".mscx")) {
+                mscxFileName = name;
+                break;
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/engraving/io/mscreader.cpp
```diff
@@ -119,13 +119,16 @@ QByteArray MscReader::readStyleFile() const
...
-    QString mscxFileName;
-    QStringList files = reader()->fileList();
-    for (const QString& name : files) {
-        // mscx file in the root dir
-        if (!name.contains("/") && name.endsWith(".mscx")) {
-            mscxFileName = name;
-            break;
+    QString mscxFileName = QFileInfo(m_params.filePath).completeBaseName() + ".mscx";
+    QByteArray data = fileData(mscxFileName);
+    if (data.isEmpty()) {
```
wizofaus, line 124: when would this ever happen though? is the idea that you'd actually be able to select a directory name to open instead of an mscx file?

igorkorsukov: I would like the ыcore file in the `mscz` archive to have a fixed name `score.mscx`. But right now, this cannot be changed, and I'm not 100% sure that this is a good idea.
So, this code is just in case

> is the idea that you'd actually be able to select a directory name to open instead of an mscx file?
I thought about it, but there is a technical limitation, the API of Qt allows us to make either a file selection or a directory selection, but not both at once. I would not like to do two different menu items for opening.
So, now it is done so that we select the `mscx` file, but in fact, we select the project directory (similar to how we open the C++ project in IDE)

src/engraving/io/mscreader.cpp
```diff
@@ -119,13 +119,16 @@ QByteArray MscReader::readStyleFile() const
 
 QByteArray MscReader::readScoreFile() const
```
Eism, line 120: what about another methods?
excerptNames
imageFileNames

igorkorsukov: Good catch, I improved the solution

Eism (approved):

src/engraving/io/mscreader.cpp
wizofaus: Ok, as long as the facility to open/import mscx files when there are multiple ones in a directory works as expected - I do that all the time with the unit test files now (and it's somewhat irritating that you can't easily update them by saving, which slightly unexpectedly creates a whole new directory,  but it's workable)

PR #8788, opened 2021-08-03
Moved export scenario and model to project module

igorkorsukov:
qml will be moved separately

### DIFF
--- a/src/project/internal/exportprojectscenario.h
+++ b/src/project/internal/exportprojectscenario.h
@@ -19,28 +19,28 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-#ifndef MU_USERSCORES_EXPORTSCORESERVICE_H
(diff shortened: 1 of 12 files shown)


src/project/internal/exportprojectscenario.h
```diff
@@ -19,25 +19,25 @@
...
 #include "context/iglobalcontext.h"
 #include "system/ifilesystem.h"
 
-namespace mu::userscores {
-class ExportScoreScenario : public IExportScoreScenario
+namespace mu::project {
+class ExportProjectScenario : public IExportProjectScenario
 {
-    INJECT(userscores, project::IProjectConfiguration, configuration)
+    INJECT(userscores, IProjectConfiguration, configuration)
```
Eism, line 38: please rename userscores to project here

Eism (approved):

src/project/internal/exportprojectscenario.h
igorkorsukov: Thanks, fixed

PR #8795, opened 2021-08-04
[MU4] Fixed using of ID for parts/staves

RomanPudashkin:


Eism (approved):

src/engraving/libmscore/excerpt.cpp
```diff
@@ -114,7 +114,7 @@ bool Excerpt::isEmpty() const
     return partScore() ? partScore()->parts().empty() : true;
 }
 
-void Excerpt::removePart(const QString& id)
+void Excerpt::removePart(const ID& id)
```
wizofaus, line 117: Would it be better to call it StaffID or PartID to make it clear what sort of ID it is elsewhere? (I'd think the argument name `id` would be better if it was more explicit too)

src/engraving/libmscore/mscore.h
```diff
@@ -135,6 +135,9 @@ static const int VISUAL_STRING_NONE      = -100;      // no ordinal for the visu
 static const int STRING_NONE             = -1;        // no ordinal for a physical string (0 = topmost in instrument)
 static const int FRET_NONE               = -1;        // no ordinal for a fret
 
+using ID = uint64_t;
+static constexpr ID INVALID_ID = 0;
```
wizofaus, line 139: What sort of objects will have these sorts of IDs?

src/engraving/libmscore/part.cpp
```diff
@@ -241,12 +268,13 @@ void Part::setStaves(int n)
...
     }
+
     int staffIdx = score()->staffIdx(this) + ns;
     for (int i = ns; i < n; ++i) {
-        Staff* staff = new Staff(score());
-        staff->setPart(this);
+        Staff* staff = createStaff(score(), this);
         _staves.push_back(staff);
-        score()->staves().insert(staffIdx, staff);
+        const_cast<QList<Staff*>&>(score()->staves()).insert(staffIdx, staff);
```
wizofaus, line 276: Could we avoid a const_cast here and export a "insertStaff" function from Score? (much as I'd hate to make that interface any bigger!)

src/engraving/libmscore/part.h
```diff
@@ -70,21 +70,24 @@ class Part final : public ScoreElement
     QString _partName;              ///< used in tracklist (mixer)
     InstrumentList _instruments;
     QList<Staff*> _staves;
-    QString _id;                    ///< used for MusicXml import
-    bool _show;                     ///< show part in partitur if true
-    bool _soloist;                  ///< used in score ordering
+    ID _id = INVALID_ID;             ///< used for MusicXml import
```
wizofaus, line 73: We don't seem to have settled on a preference for using = vs { } for member initialization...I'd been using { } everywhere as that seemed to be the most common.

src/engraving/libmscore/staff.cpp
```diff
@@ -720,8 +710,8 @@ Fraction Staff::currentKeyTick(const Fraction& tick) const
 
 void Staff::write(XmlWriter& xml) const
 {
-    int idx = this->idx();
-    xml.stag(this, QString("id=\"%1\"").arg(idx + 1));
+    xml.stag(this, QString("id=\"%1\"").arg(idx() + 1));
```
wizofaus, line 713: I know it was like this already but it seems odd that we use `idx() + 1` as the id here, when there's already an _id property?

src/engraving/libmscore/mscore.h
RomanPudashkin: I don't want to create dependency on the framework in libmscore because libmscore can be used independently in other projects. Probably later we'll move some types (like ID or async primitives) to libmscore, but now isn't the right time for that :)

src/engraving/libmscore/staff.cpp
RomanPudashkin: This requires a lot of work. I'll change this in a separate PR

src/engraving/libmscore/part.h
RomanPudashkin: I like classical initialization (without any brackets, just = operator) better. It's more explicit

src/engraving/libmscore/part.cpp
RomanPudashkin: This is a hack. I have absolutely no idea what is going on in this method  :) It was really easy to apply my changes to all libmscore, but not in this method. For some reason, insertStaff breaks it, so I decided to left it as it is

src/engraving/libmscore/excerpt.cpp
RomanPudashkin: I don't like the idea to create aliases for all sorts of IDs (PartId, StaffId, MeasureId...). I think, the best solution here (and in the same situation) is just to rename the argument to partId):

void Excerpt::removePart(const ID& partId)

Eism (approved):

Issue #8442, opened 2021-06-28
[MU4 Issue] Tabbing outline stroke needs to sit on the outside of each element (currently it is inside)


Tantacrul:
I've created this visual to help explain the idea:

![image](https://user-images.githubusercontent.com/21022311/123637000-64ae7600-d815-11eb-9d22-9c99aca2bc0c.png)

The stroke for tabbing needs to sit on the outside of interface elements so that it does not clash with the strokes in High Contrast mode.

This was in the initial design document but seems to have been overlooked.

cbjeukendrup:
I discussed this some days ago with @ArjunBoi. I believe he is willing to do this in his accessibility PR (or maybe also just in a separate PR). 

So Arjun, would it be okay if you were assigned for this issue?

ArjunBoi:
Sure!
I will take up this issue 😄 
The current implementation has us choose between which border to display in a lot of elements `border.color : keyNavCtrl.active ? ui.theme.focusColor : ui.theme.borderColor`, with the same logic for the border's Width.

I believe the solution for this would be to have another Rectangle element (The size of this new element would be the same as the background Rectangle element + 2px + ui.theme.borderWidth) and have that as a dedicated Navigation Control border rectangle. The width of the border would be 2px, and the existing background Rectangle would hold the "normal" use case's borders. 
These are just my initial ideas of course 😅

shoogle:
@ArjunBoi, I suggest you implement it however you think it best and then ask for a review.

If it is going to take a lot of work to get it right everywhere then I suggest you just do it in one place to start with, get your method confirmed in a review, then extend it to work the same way in other place.

ArjunBoi:
@shoogle, @Tantacrul, @cbjeukendrup 

I've got the navCtrl borders working (the idea is working, and i have implemented it for FlatButtons only as of yet so that i can get it approved first before making large scale changes) 

The navCtrl borders, in decreasing order of visibility are:

* Light Theme [very good visibility]
* Dark Theme [moderate visibility]
* High Contrast White Theme [not that great visibility, since we already have black borders]
* High Contrast Black Theme [near invisible, unless you're REALLY looking out for it]

The color being used for these borders is `#75507b`, which isn't that great for darker themes. 
I think increasing the size of the navCtrl borders from 2px may not improve visibility at all, so we could try switching up the border color for different themes instead?

**Demonstration:**
Pay attention to the following buttons:
* "File", "Edit", "View" ... etc. buttons at the top 
[...]

PR #8806, opened 2021-08-06
Navigation Control Border Re-Implementation 

ArjunBoi:
Resolves: #8442

All elements supporting Navigation Control now draw the border rectangle on the outside. 
There were a total of 5 different elements i.e. 
* `Dropdown`: (it works for the main Dropdown item, but the items inside it are very close together. I have implemented this in such a way that the borders for the popup elements are drawn on the inside). 
* `DockPanelTab` : The borders are being drawn on the inside, again because of lack of space.
* `InstrumentsTreeItemDelegate` : Borders on the inside, same reason. 
* `PaletteTree` : This file uses `ListItemBlank` as a delegate to draw its children. I've done it in such a way such that only this instance of `ListItemBlank` 's borders get drawn on the inside.
* `Expandable Blank` : Same thing

The borders for these are drawn on the inside, in a way that it does not interfere with the High-Contrast borders; i.e. the HC 1px border + the 2px navCtrl border will combine to make a single 3px border on the inside (the innermost being the HC border, followed by the navCtrl one).

For a demonstration, Navigate to DevTools -> KeyNav. (Please make sure that you are in Light mode. The colors for the elements in that page are hard-coded and don't work well with other themes)

This PR does not modify the existing Navigation Control logic for elements, just the way the borders are drawn around them.


src/appshell/qml/DevTools/KeyNav/KeyNavSubSection.qml
```diff
@@ -34,8 +34,21 @@ Rectangle {
     height: 40
     width: btns.childrenRect.width
 
-    border.color: "#75507b"
-    border.width: keynavsec.active ? 4 : 0
+    Rectangle {
```
RomanPudashkin, line 37: I believe you need to create a separate component (NavigatioBorder) and reuse it instead of copy-pasting this code in all components

src/framework/ui/view/uitheme.cpp
```diff
@@ -125,7 +125,7 @@ void UiTheme::initThemeValues()
     m_focusColor = themeValues[FOCUS_COLOR].toString();
 
     m_borderWidth = themeValues[BORDER_WIDTH].toReal();
-    //m_navCtrlBorderWidth = 2;
+    m_navCtrlBorderWidth = 2.0;
```
RomanPudashkin, line 128: Better to get this value from the current theme to avoid the magic number

ArjunBoi:
There are a lot of bugs within the Navigation Control system that made testing very troublesome.
There were elements that were mentioned only in a few places in the code, which were out of the reach of NavigationControl and thus weren't test-able immediately.

In other places, the application would often crash trying out the different options. I also encountered behavior in which i would press space to select an option in a group of `FlatRadioButton`s, and the application wouldn't update the changes made because of having that button selected. The changes were only made once i clicked that button with my mouse. In the same element, pressing "Enter" instead of "Spacebar" led to a crash. (I found this behavior in the Text settings in the Inspector/Properties tab)

src/framework/ui/view/uitheme.cpp
ArjunBoi: Yes that
I'm going to make it into a theme variable.
That will allow us to have different navigation border widths for different themes, which might come in handy

src/appshell/qml/DevTools/KeyNav/KeyNavSubSection.qml
ArjunBoi: Understood!

src/framework/ui/view/uitheme.cpp
ArjunBoi: Done :)

src/appshell/qml/DevTools/KeyNav/KeyNavSubSection.qml
ArjunBoi: Fixed :)

RomanPudashkin (approved):

Issue #8769, opened 2021-07-30
Note input colour is incorrect


bkunda:
The note cursor in note input mode is currently using a very bright green instead of the correct shade of blue.

Incorrect colour in MS4:

https://user-images.githubusercontent.com/86290556/127651143-13e135ef-3874-475b-b6ce-09ae8f7f5c2e.mov

Correct colour in MS3.6

https://user-images.githubusercontent.com/86290556/127651179-587cf670-71aa-4935-83be-220f7dcc41ff.mov

**Expected behavior**
The note input colour should match the current colours used in MS3.6.

MacBook Air M1, MacOS11.4

cbjeukendrup:
I'm working on this!

Also, the note input colour should take the voice colour into account. In MU3, the colour was calculated by making the relevant voice colour 35% lighter. Of course I could just replicate this behaviour in MU4, but we could also just take the voice color with something like 0.7 opacity. How about that?

bkunda:
Thanks Caspar! That sounds like a logical solution to me 🙂

PR #8814, opened 2021-08-07
Fix color of shadow note

cbjeukendrup:
Resolves: #8769 

And simplified the list of colors in EngravingConfiguration and transferred the colors from MScore to EngravingConfiguration. 

---

@bkunda Now the shadow note color equals the voice selection color, but with 0.5 opacity. The result slightly disappoints me; a problem is that the shadow note exists of multiple components which are drawn separately, which becomes obvious with a non-1.0 opacity. It seems not trivial to change the way the note is drawn. 
I also doubt whether the shadow note is visible well enough, especially when you hover a black note that is already in the score. 

You can download the build here from GitHub to try it out.

Maybe it's better to just take the MU3 approach? 

Schermafbeelding 2021-08-07 om 17 21 13


RomanPudashkin (approved):

bkunda:
I have to say I quite like this solution @cbjeukendrup, but I think @jessjwilliamson and @Tantacrul should check it out as well because it is an especially noticeable and integral part of how the UI works. 
I'm glad all the colours are working well now, though (gone be the ultra highlight green!).

Tantacrul:
I think you should increase the opacity to 80% so it can pass the minimum contrast ratio.

Currently this is how it performs:
image


You can use this site to test:
https://webaim.org/resources/contrastchecker/

Tantacrul:
Hmmm. I almost wonder if we should remove the opacity entirely?

jessjwilliamson:
Hey just catching up on this thread..

I don't remember having an opacity on note input?.. just a slightly lighter version colour of selected voice 1 colour? I'm not a huge fan of the opacity idea, to be honest. 

"I also doubt whether the shadow note is visible well enough, especially when you hover a black note that is already in the score." I absolutely agree.

Tantacrul:
@jessjwilliamson - if we want to colour to be lighter, then we should create specific colours - along with their 'light' equivalents. Can you do this?

I don't recall if we made any changes to the V1, V2, V3 & V4 colours. They're a little ghastly in MS3 and could do with a tiny bit of refinement.

cbjeukendrup:
I'm not sure if creating another set of lighter colours is a good idea. The voices colors can be customised, so therefore I would prefer if the lighter color is calculated based on the normal voice colours. 

In the end I think the MU3 behaviour was totally fine. The reason for the opacity experiment was that the colors were broken recently, caused by the switch from the `QColor` representation (from Qt) to our own `Color`, and our own colours work a bit differently, so they don't have the `lighter` and `darker` functions that we used from `QColor`. But that can be easily solved by just converting back and forth to `QColor`.

cbjeukendrup:
I returned the MU3 behaviour in #8820.
If we still want to adjust the default selection colours, please let me know, so I can add the change to that PR. The lighter colors for the shadow note and drop target elements will be calculated based on those (35% lighter; the 35% can of course be adjusted too).

Tantacrul:
@cbjeukendrup - sure: some way to apply general ‘lightness’ would fine, although I’d opt for not going too light (again, for contrast reasons). Agree the MuseScore 3 way wasn’t bad.

@jessjwilliamson: in that case, let’s define slightly refined new colours for each voice.

Issue #8812, opened 2021-08-06
Text editing interaction regression: clicking on text is harder and text selection is semi-broken


Tantacrul:
### Bug
The ability to select different parts of a selected text box is significantly harder in MS4.

### Problems:
See video here. Please note that the first example (where it is working) is MS3. The second example is MS4.
https://www.dropbox.com/s/kqsncswr7t6lk83/Check_out_insertion_point_rubbishness.mov?dl=0

- Compared to MS3, when editing text, the user can't click to reposition the insertion point. It just deselects the text field.
- When editing text, you can't click and drag to select a range of text. Again, it just exits the editing mode

Mac OS - Catalina

wizofaus:
Line 366 of notationviewinputcontroller.cpp
`        if (!hitElement || !hitElement->isText()) {`

needs to be `isTextBase()`.

I have this fix on the branch I'm working on but it won't ready for a while.

vshalkevich:
Cant reproduce

PR #8815, opened 2021-08-08
[MU4] text editing fixes

wizofaus:
Resolves: So many issues...but originally https://github.com/musescore/MuseScore/issues/8652


src/engraving/io/xmlreader.cpp
```diff
@@ -390,8 +390,10 @@ void XmlReader::htmlToString(int level, QString* s)
             --level;
             return;
         case QXmlStreamReader::Characters:
-            if (!isWhitespace()) {
```
wizofaus, line 393: this is a bug in MU3, it causes space to go missing from text in some circumstances, but I don't know when this isWhitespace() ever *is* needed, probably need to check git history to determine why it was added

src/engraving/libmscore/textbase.cpp
```diff
@@ -968,11 +988,11 @@ mu::draw::Font TextFragment::font(const TextBase* t) const
...
     } else {
         family = format.fontFamily();
+        font.setBold(format.bold());
+        font.setItalic(format.italic());
     }
 
     font.setFamily(family);
-    font.setBold(format.bold());
-    font.setItalic(format.italic());
     Q_ASSERT(m > 0.0);
```
wizofaus, line 996: Much simpler and more reliable to ignore bold/italic flags here than when setting the CharFormat

src/engraving/libmscore/textbase.cpp
```diff
@@ -2593,6 +2612,13 @@ bool TextBase::acceptDrop(EditData& data) const
 
 void TextBase::setXmlText(const QString& s)
 {
+    // ensure cursor has correct format from element's text-style
+    _cursor->init();
+    for (const StyledProperty& spp : *textStyle(tid())) {
+        if (spp.pid == Pid::FONT_FACE || spp.pid == Pid::FONT_STYLE || spp.pid == Pid::FONT_SIZE || spp.pid == Pid::TEXT_SCRIPT_ALIGN) {
+            setProperty(spp.pid, styleValue(spp.pid, getPropertyStyle(spp.pid)));
```
wizofaus, line 2619: TODO: simplify

src/engraving/libmscore/textedit.h
```diff
@@ -60,13 +60,74 @@ struct TextEditData : public ElementEditData {
 class TextEditUndoCommand : public UndoCommand
 {
 protected:
-    TextCursor c;
+    TextCursor curs;
```
wizofaus, line 63: todo: rename to _cursor

src/engraving/libmscore/textedit.h
```diff
@@ -60,13 +60,74 @@ struct TextEditData : public ElementEditData {
...
+    }
+
+    virtual void redo(EditData*) override
+    {
+        xmlText = cursor().text()->xmlText();
+        restoreSelection();
+        if (propertyId == Pid::FONT_STYLE) {
+            FontStyle setStyle = static_cast<FontStyle>(propertyVal.toInt());
+            if ((setStyle& FontStyle::Bold) != (existingStyle & FontStyle::Bold)) {
+                cursor().setFormat(FormatId::Bold, setStyle & FontStyle::Bold);
```
wizofaus, line 120: all this logic is because bold/italic/underline aren't properties in their own right, but we need to know which one is being set, to ensure we only apply that and don't override the others

src/framework/shortcuts/data/shortcuts-Mac.xml
```diff
@@ -1093,7 +1093,7 @@
     </SC>
   <SC>
     <key>text-b</key>
-    <seq>Ctrl+b</seq>
+    <seq>Ctrl+B</seq>
```
wizofaus, line 1096: lowercase letters don't seem to work here - but actually even though MU3 shows these as customisable shortcut keys, they aren't - if you change the shortcut key combination for this command (likewise text-i & text-u) it's not honoured. Fixed in MU4, they're top-level actions now.

src/inspector/models/abstractinspectormodel.cpp
```diff
@@ -343,6 +343,9 @@ QVariant AbstractInspectorModel::valueToElementUnits(const Ms::Pid& pid, const Q
     }
 
     default:
+        if (value.type() == QVariant::Type::Color) {
+            return QVariant::fromValue(mu::draw::Color(value.value<QColor>()));
+        }
```
wizofaus, line 348: fixes bug where selecting color in inspector didn't work.

src/inspector/models/inspectorlistmodel.cpp
```diff
@@ -292,5 +292,12 @@ void InspectorListModel::subscribeOnSelectionChanges()
             auto elements = m_notation->interaction()->selection()->elements();
             setElementList(QList(elements.cbegin(), elements.cend()));
         });
+
+        m_notation->interaction()->textEditingChanged().onNotify(this, [this]() {
+            auto element = m_notation->interaction()->selection()->element();
+            if (element != nullptr) {
+                setElementList(QList { element });
```
wizofaus, line 299: necessary to ensure inspector updates as expected while editing text

src/inspector/types/texttypes.h
```diff
@@ -56,52 +55,65 @@ class TextTypes
         FRAME_TYPE_SQUARE,
         FRAME_TYPE_CIRCLE
     };
-
+    // must match Ms::Tid
```
wizofaus, line 58: i don't like this but not sure what else to do...

src/inspector/view/qml/MuseScore/Inspector/common/InspectorPropertyView.qml
```diff
@@ -84,7 +84,7 @@ Column {
                 menuAlign: Qt.AlignHCenter
 
                 onHandleMenuItem: {
-                    switch (id) {
+                    switch (itemId) {
```
wizofaus, line 87: due to these, "Reset" and "set as style" buttons were not working

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -363,7 +363,7 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
     }
 
     if (viewInteraction()->isTextEditingStarted()) {
-        if (!hitElement || !hitElement->isText()) {
+        if (!hitElement || !hitElement->isTextBase()) {
```
wizofaus, line 366: meant mouse navigation while editing most types of text elements wasn't working (https://github.com/musescore/MuseScore/issues/8812)

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -100,204 +100,208 @@ void NotationActionController::init()
...
-    dispatcher()->reg(this, "paste-double", [this]() { pasteSelection(PastingType::Double); });
-    dispatcher()->reg(this, "paste-special", [this]() { pasteSelection(PastingType::Special); });
-    dispatcher()->reg(this, "swap", this, &NotationActionController::swapSelection);
-    dispatcher()->reg(this, "delete", this, &NotationActionController::deleteSelection);
-    dispatcher()->reg(this, "flip", this, &NotationActionController::flipSelection);
-    dispatcher()->reg(this, "tie", this, &NotationActionController::addTie);
-    dispatcher()->reg(this, "chord-tie", this, &NotationActionController::chordTie);
-    dispatcher()->reg(this, "add-slur", this, &NotationActionController::addSlur);
+    registerAction("flat2", [this]() { toggleAccidental(AccidentalType::FLAT2); });
+    registerAction("flat", [this]() { toggleAccidental(AccidentalType::FLAT); });
```
wizofaus, line 104: changed so that none of these commands are available while editing text - "insert measure" was a particular problem because the default shortcut was Ctrl+B which is also for toggling bold styling. This is consistent with Mu3 and what makes logical sense anyway.

wizofaus:
Unit tests were expected to fail, next big piece of work is to fix up all the unit test reference files.

src/importexport/musicxml/tests/data/testBreaksImplExpl_all_ref.xml
```diff
@@ -134,7 +134,7 @@
       <print new-page="yes"/>
       <direction>
         <direction-type>
-          <words font-family="Edwin">After system break</words>
```
wizofaus, line 137: These had been added not long ago and are not in the MU3 reference set.
However I am slightly concerned there's actually no mention at all of "Edwin" in most of the xml files, whereas the MU3 reference set did have "words-font" and "lyrics-font" entries specifying the defaults. Will investigate.

src/notation/internal/notationinteraction.cpp
```diff
@@ -2016,6 +2016,7 @@ void NotationInteraction::endEditText()
     m_textEditData.clearData();
 
     notifyAboutTextEditingChanged();
+    notifyAboutSelectionChanged();
```
wizofaus, line 2019: necessary to ensure that the properties panel (inspector) and menus etc. update correctly when you finish editing a text element using the Esc key (it was working in other cases, such as clicking away, as that action already triggered this notification)

src/engraving/libmscore/textedit.h
```diff
@@ -60,13 +60,75 @@ struct TextEditData : public ElementEditData {
...
+
+    virtual void undo(EditData*) override
+    {
+        cursor().text()->resetFormatting();
+        cursor().text()->setXmlText(xmlText);
+        restoreSelection();
+        cursor().text()->layout();
+    }
+
+    virtual void redo(EditData*) override
```
RomanPudashkin, line 114: 1. better to move all implementation to cpp. We should leave only declarations in h, all implementation must be stored in cpp

2. you can remove virtual, because you already used override

wizofaus: agree, just that everything else in here was done this way

wizofaus: moved implementations to .cpp file

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -364,10 +364,16 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
...
             viewInteraction()->endEditText();
         } else {
-            viewInteraction()->changeTextCursorPosition(m_beginPoint);
+            const mu::RectF& bbox = hitElement->canvasBoundingRect();
+            mu::PointF constrainedPt = mu::PointF(
+                m_beginPoint.x() < bbox.left() ? bbox.left()
+                : m_beginPoint.x() >= bbox.right() ? bbox.right() - 1 : m_beginPoint.x(),
+                m_beginPoint.y() < bbox.top() ? bbox.top()
+                : m_beginPoint.y() >= bbox.bottom() ? bbox.bottom() - 1 : m_beginPoint.y());
+            viewInteraction()->changeTextCursorPosition(constrainedPt);
```
wizofaus, line 376: ensures that if you click just outside the element bounding box (close enough to be a 'hit'), it can still find the cursor position correctly

src/notation/internal/notationinteraction.cpp
```diff
@@ -325,6 +325,15 @@ QList<Ms::Element*> NotationInteraction::hitElements(const PointF& p_in, float w
 
     PointF p = p_in - page->pos();
 
+    if (isTextEditingStarted()) {
+        auto editW = w * 2;
+        RectF hitRect(p.x() - editW, p.y() - editW, 2.0 * editW, 2.0 * editW);
+        if (m_textEditData.element->intersects(hitRect)) {
```
wizofaus, line 331: doubling the "hit" tolerance might be excessive for higher values (it's a user preference), but certainly for the default value, it's surprisingly hard to ensure you can select text from the beginning or the end using the mouse. Even in Mu3 this is a bit of an issue.

vshalkevich:
Works fine for me

RomanPudashkin (approved):

PR #8816, opened 2021-08-08
[MU3] fix #321809: refuse dropping a MMRest (avoid crash)

lyrra:

### DIFF
--- a/libmscore/rest.cpp
+++ b/libmscore/rest.cpp
@@ -196,6 +196,9 @@ QRectF Rest::drag(EditData& ed)
 
 bool Rest::acceptDrop(EditData& data) const
       {
+      if (measure()->isMMRest()) { // avoid crash
+            return false;
+            }
       Element* e = data.dropElement;
       ElementType type = e->type();
       if (


Jojo-Schmitz:
Applies to master (AKA MuseScore 4) too (expect for indent style)

Jojo-Schmitz:
Copying (just) a tie onto a (normal) rest still causes a crash, even with that PR. Might be a different issue altogether though?

libmscore/rest.cpp
```diff
@@ -196,6 +196,9 @@ QRectF Rest::drag(EditData& ed)
 
 bool Rest::acceptDrop(EditData& data) const
       {
+      if (measure()->isMMRest()) { // avoid crash
+            return false;
+            }
```
wizofaus, line 201: Is this definitely right though? Seems you can safely add some elements to multi-measure rests?

Jojo-Schmitz: You could use the artifact from the corresponding 3.x PR to test this

Jojo-Schmitz: Question: is this about dropping an MM Rest (onto something), or about dropping (something) onto an MM rest, or both?

wizofaus:
Something on to a rest

wizofaus:
The element being dropped is in data.dropElement, it's not even checking that

Jojo-Schmitz:
I see, so my change of the commit title in #8818 is correct ;-)

wizofaus:
This definitely the wrong fix. You should be able to paste a note onto a MM-rest and in principle it will work (it should "break" the rest and add the note into first measure - as happens if you use step-entry with an MM-rest selected), but there's a bug in the logic I'm trying figure out.

Jojo-Schmitz:
How about this then:
```
index 4338d7f29..6879a6e17 100644
--- a/libmscore/rest.cpp
+++ b/libmscore/rest.cpp
@@ -242,6 +239,12 @@ bool Rest::acceptDrop(EditData& data) const
 Element* Rest::drop(EditData& data)
       {
       Element* e = data.dropElement;
+
+      bool mmRest = measure()->isMMRest();
+      int mmRestCount = measure()->mmRestCount();
+      if (mmRest)
+            measure()->setMMRestCount(1); // effectively disabling mmrests temporarily
+
       switch (e->type()) {
             case ElementType::ARTICULATION:
                   {
@@ -251,6 +254,8 @@ Element* Rest::drop(EditData& data)
                         e = 0;
                         }
                   }
+                  if (mmRest)
+                        measure()->setMMRestCount(mmRestCount);
                   return e;

             case ElementType::CHORD: {
@@ -284,11 +289,17 @@ Element* Rest::drop(EditData& data)
             case ElementType::IMAGE:
                   e->setParent(this);
                   score()->undoAddElement(e);
+                  if (mmRest)
+                        measure()->setMMRestCount(mmRestCount);
                   return e;

             default:
+                  if (mmRest)
+                        measure()->setMMRestCount(mmRestCount);
                   return ChordRest::drop(data);
             }
+      if (mmRest)
+            measure()->setMMRestCount(mmRestCount);
       return 0;
       }

```

Jojo-Schmitz:
Hmm, still crashes on pasting a single note(head), but works for a range selection. And that worked before too.
So pasting a note head onto an MMRest should ceate a tied series of notes as long the the entire mmrest? Nah, c'mon...

wizofaus:
Hmm, that doesn't look right to me.  I agree that given the MU3 behaviour of pasting a note-head preserving the duration of the destination element is an issue, which is why I suggested if we were to support it for MMRests, just set it to the length of the measure. But honestly for MU3 i'd go with the simplest/safest fix you can, and if that means preventing pasting notes onto MMRests then fine - my only issue was that the proposed fix stopped the use pasting anything at all on to an MMRest.

Jojo-Schmitz:
OK, how about this then:
```
index 4338d7f29..d93adb462 100644
--- a/libmscore/rest.cpp
+++ b/libmscore/rest.cpp
@@ -196,9 +196,6 @@ QRectF Rest::drag(EditData& ed)

 bool Rest::acceptDrop(EditData& data) const
       {
-      if (measure()->isMMRest()) { // avoid crash
-            return false;
-            }
       Element* e = data.dropElement;
       ElementType type = e->type();
       if (
@@ -217,7 +214,7 @@ bool Rest::acceptDrop(EditData& data) const
          || (type == ElementType::BAR_LINE)
          || (type == ElementType::BREATH)
          || (type == ElementType::CHORD)
-         || (type == ElementType::NOTE)
+         || (type == ElementType::NOTE && !measure()->isMMRest())
          || (type == ElementType::STAFF_STATE)
          || (type == ElementType::INSTRUMENT_CHANGE)
          || (type == ElementType::DYNAMIC)
```

wizofaus:
Better, but what happens if you try to paste a rest...or a chord (actually I don't even know how you select just a "chord" and paste it)?

Jojo-Schmitz:
`EementType::REST` is not part of that function, so not an issue. You select a chord by click/select), shift+right, shift+left.

wizofaus:
Nope, that selects a range.

Jojo-Schmitz:
That one-liner seems the cheapest fix possible and the most correct one too. I don't thing anyone reallly wants pasting a note head onto an mm rest to create a huge series of tied notes, not even @Tantacrul ;-)

Jojo-Schmitz:
Yes, a chord basically is a range. noteheads + duration.

wizofaus:
Yeah but pasting range-selections goes through completely different logic from what I've observed, anyway, as long as it's not crashing I think it's fine.

Jojo-Schmitz:
Possible. But it doesn't crash indeed. And that's the whole point of the PR.

Jojo-Schmitz:
Backporting #8825 would be the best option. 2nd best my one-liner above to disable pasting only note(heads) onto MM rests.

lyrra:
Closing this as a fix has been merged. Thanks for a good discussion!

Jojo-Schmitz:
Well, for 3.x none has been merged yet, but the issue is resolved for master

PR #8818, opened 2021-08-08
[MU4] Fix crash when pasting tie onto rest

Jojo-Schmitz:
Fixes a regression introduced with #8510

While at it (and being asked for it by @lyrra) also porting #8816 to master, to fix https://musescore.org/en/node/321809.
Edit: using a different and IMHO better fix now, see discussion in #8816., which only disables pasting a `NOTE`(head) onto a MMRest:
```
index 4338d7f29..d93adb462 100644
--- a/libmscore/rest.cpp
+++ b/libmscore/rest.cpp
@@ -217,7 +214,7 @@ bool Rest::acceptDrop(EditData& data) const
          || (type == ElementType::BAR_LINE)
          || (type == ElementType::BREATH)
          || (type == ElementType::CHORD)
-         || (type == ElementType::NOTE)
+         || (type == ElementType::NOTE && !measure()->isMMRest()) // avoid crash
          || (type == ElementType::STAFF_STATE)
          || (type == ElementType::INSTRUMENT_CHANGE)
          || (type == ElementType::DYNAMIC)
```
Edit 2: #8825 fixes that even better, by actually making it work, so I'm removing that above mentioned change from this PR

### DIFF
--- a/src/engraving/libmscore/rest.cpp
+++ b/src/engraving/libmscore/rest.cpp
@@ -188,7 +188,8 @@ bool Rest::acceptDrop(EditData& data) const
         ) {
         return true;
     }
-    return type != ElementType::SLUR && e->isSpanner();
+    // prevent 'hanging' slurs, avoid crash on tie
+    return type != ElementType::SLUR && type != ElementType::TIE && e->isSpanner();
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -1863,7 +1863,7 @@ bool Note::acceptDrop(EditData& data) const
            || (type == ElementType::FRET_DIAGRAM)
            || (type == ElementType::FIGURED_BASS)
            || (type == ElementType::LYRICS)
-           || e->isSpanner();
+           || (type != ElementType::TIE && e->isSpanner());
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -98,12 +98,12 @@ bool TieSegment::edit(EditData& ed)
 {
     SlurTie* sl = tie();
 
-    if (ed.key == Qt::Key_X) {
+    if (ed.key == Qt::Key_X && !ed.modifiers) {
         sl->setSlurDirection(sl->up() ? Direction::DOWN : Direction::UP);
         sl->layout();
         return true;
     }
-    if (ed.key == Qt::Key_Home) {
+    if (ed.key == Qt::Key_Home && !ed.modifiers) {
         ups(ed.curGrip).off = PointF();
         sl->layout();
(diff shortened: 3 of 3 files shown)


src/engraving/libmscore/rest.cpp
```diff
@@ -188,7 +188,7 @@ bool Rest::acceptDrop(EditData& data) const
         ) {
         return true;
     }
-    return type != ElementType::SLUR && e->isSpanner();
+    return type != ElementType::SLUR && type != ElementType::TIE && e->isSpanner();
```
Jojo-Schmitz, line 194: It seems possible to allow pasting/dropping slurs onto rests, only ties crash. Would we want to allow for that, @wizofaus?

wizofaus: I wouldn't change it at this stage, unless you're 100% certain what impacts it will have.
I would definitely write a unit test to ensure we catch cases like this that can cause a crash though!

Jojo-Schmitz: OK, I'm not 100% sure. Only a very basic test showed that it works. And even if it does, it is questionable how often this is reaslly wanted anyway, so let's better not take the risk, not now at least.

Jojo-Schmitz: Feel free to provide a unit test ;-)

wizofaus: I'm not in charge of approvals but if I were I'd require one to be submitted along with the fix!  Only reason I hadn't been is that I couldn't work out how to compile/run them previously, but that's working fine now.
But a simple test that creates a tie programmatically and confirms that it's not possible to drop it onto a rest (and cause a crash) shouldn't be hard to write. Key thing is that the test *should* have crashed before this fix.

Jojo-Schmitz: No idea how such a test would need to be made. Also having one after the fix seems pretty pointless?

wizofaus: The point of the test is ensure someone else can't make another change in the future that triggers the same behaviour.
This should do it, it crashes for me without your fix:

```
    Element* tie = new Tie(score);
    ChordRest* chordRest = score->firstSegment(SegmentType::ChordRest)->nextChordRest(4);
    ed.dropElement = tie;
    if (chordRest->acceptDrop(ed)) {
        chordRest->drop(ed);
    }
```

wizofaus:
I've already put up a PR that fixes it properly for MU4...

Jojo-Schmitz:
I guess you mean #8825

Jojo-Schmitz:
Which won't fix the "paste tie onto rest" crash issue, right?

wizofaus:
Not in its present form, no.

Jojo-Schmitz:
Feel free to integrate that one-line of mine

RomanPudashkin (approved):

Issue #8824, opened 2021-08-09
[MU4 Issue] Pasting notes onto rests not reliable


wizofaus:
MU4's behaviour was changed so that when copying and pasting notes, the length is preserved, however it's not reliable currently.  Especially when pasting onto a MM rest (which crashes).

**To Reproduce**
1. Create an empty score
2. Add a 1/4 note to the first measure
3. Copy it to the clipboard
4. Turn on multi-measure rests
5. Select the multi-measure rest and paste
6. MuseScore crashes

However there's a bunch of other problematic behaviours when copying/pasting notes/rests on to existing notes/rests

**Expected behavior**
a) Pasting a note on to a multi-measure rest should behave the same as pasting onto a whole rest, but with the measure pasted into no longer included in the multi-measure rest (as it's no longer empty)
b) Pasting a single note on to a single rest of the same duration: should just convert rest to note (works now, except attached articulations etc. are lost - what logically should be kept?)
c) Pasting a single note on to a single rest of a longer duration: should add note with copied duration and pad out rest (mostly works now, except for articulations, though I did see it misbehave once or twice)
d) Pasting a single note on to a single rest of a shorter duration: should add note with copied duration and remove whatever subsequent note/rests occur within the pasted duration period
e) Pasting a single note on to a single note of the same duration: should just replace note
f) Pasting a single note on to a single note of a longer duration: should replace note with copied note and pad out with rests (to confirm?) - not what's happening now, the note gets split up unexpectedly.
g) Pasting a single note on to a single note of a shorter duration: should replace note with copied note and remove whatever subsequent note/rests occur within the pasted duration period (similar to pasting a range onto a selected range of a shorter duration)
h) Pasting a single note on to multiple (list-selected) rests/notes - tricky one, currently this doesn't work properly unless the note is exactly the same duration as all the target rests.  Logically this needs to allow for cases where the selected targets get replaced - e.g. if you select two consecutive 1/8 rests (or notes) and paste a 1/4 note onto them, the only sensible thing to do is replace both with the single note. But if the 1/8 rests/notes are separated by more than an 1/8, you can sensibly convert both of them to the note being pasted.
i) Pasting a single rest on to a note/rest - not sure, currently this does nothing in MU3 or MU4.
j) pasting a single note/rest onto a selected *range* - not sure, currently this does nothing in MU3 or MU4.
k) pasting a list-selection of notes/rests on to anything - not sure, currently this does nothing in MU3 or MU4.
[...]

PR #8825, opened 2021-08-10
Paste note fixes

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8824

Ensure that pasting notes works as expected in all circumstances

### DIFF
--- a/src/engraving/libmscore/chordrest.cpp
+++ b/src/engraving/libmscore/chordrest.cpp
@@ -519,16 +519,17 @@ Element* ChordRest::drop(EditData& data)
 
     case ElementType::NOTE: {
         Note* note = toNote(e);
-        NoteVal nval;
-        nval.pitch = note->pitch();
-        nval.tpc1 = note->tpc1();
-        nval.headGroup = note->headGroup();
-        nval.fret = note->fret();
-        nval.string = note->string();
-        score()->setNoteRest(segment(), track(), nval, ticks(), Direction::AUTO);
-        delete e;
+        Segment* seg = segment();
+        score()->undoRemoveElement(this);
+        Chord* chord = new Chord(score());
+        chord->setTrack(track());
+        chord->setDurationType(durationType());
+        chord->setTicks(ticks());
+        chord->setTuplet(tuplet());
+        chord->add(note);
+        score()->undoAddCR(chord, seg->measure(), seg->tick());
+        return note;
     }
-    break;
 
     case ElementType::HARMONY:
     {
@@ -1433,7 +1434,7 @@ void ChordRest::removeMarkings(bool /* keepTremolo */)
 bool ChordRest::isBefore(const ChordRest* o) const
 {
     if (!o || this == o) {
-        return true;
+        return false;
     }
     int otick = o->tick().ticks();
     int t     = tick().ticks();
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -777,15 +777,19 @@ Segment* Score::setNoteRest(Segment* segment, int track, NoteVal nval, Fraction 
     Element* nr   = nullptr;
     Tie* tie      = nullptr;
     ChordRest* cr = toChordRest(segment->element(track));
-
+    Tuplet* tuplet = cr && cr->tuplet() && sd <= cr->tuplet()->ticks() ? cr->tuplet() : nullptr;
     Measure* measure = nullptr;
+    bool targetIsRest = cr && cr->isRest();
     for (;;) {
         if (track % VOICES) {
             expandVoice(segment, track);
         }
-
+        if (targetIsRest && !cr->isRest()) {
+            undoRemoveElement(cr);
+            segment = addRest(segment, track, cr->ticks(), cr->tuplet())->segment();
+        }
         // the returned gap ends at the measure boundary or at tuplet end
-        Fraction dd = makeGap(segment, track, sd, cr ? cr->tuplet() : 0);
+        Fraction dd = makeGap(segment, track, sd, tuplet);
 
         if (dd.isZero()) {
             qDebug("cannot get gap at %d type: %d/%d", tick.ticks(), sd.numerator(),
@@ -846,7 +850,10 @@ Segment* Score::setNoteRest(Segment* segment, int track, NoteVal nval, Fraction 
                     note->setTieFor(tie);
                 }
             }
-            ncr->setTuplet(cr ? cr->tuplet() : 0);
+            if (tuplet && sd <= tuplet->ticks()) {
+                ncr->setTuplet(tuplet);
+            }
+            tuplet = 0;
             undoAddCR(ncr, measure, tick);
             if (addTie) {
                 undoAddElement(addTie);
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -453,7 +453,7 @@ Rest* Score::setRest(const Fraction& _tick, int track, const Fraction& _l, bool 
             // compute list of durations which will fit l
             //
             std::vector<TDuration> dList;
-            if (tuplet || staff->isLocalTimeSignature(tick)) {
+            if (tuplet || staff->isLocalTimeSignature(tick) || f == Fraction(0, 1)) {
                 dList = toDurationList(l, useDots);
                 std::reverse(dList.begin(), dList.end());
             } else {
--- a/src/engraving/libmscore/paste.cpp
+++ b/src/engraving/libmscore/paste.cpp
@@ -1031,6 +1031,66 @@ void Score::pasteSymbols(XmlReader& e, ChordRest* dst)
     }                                     // inner while readNextstartElement()
 }                                         // pasteSymbolList()
 
+static ChordRest* replaceWithRest(ChordRest* target)
+{
+    target->score()->undoRemoveElement(target);
+    return target->score()->addRest(target->segment(), target->track(), target->ticks(), target->tuplet());
+}
+
+static Note* prepareTarget(ChordRest* target, Note* with, const Fraction& duration)
+{
+    if (!target->segment()->element(target->track())) {
+        return nullptr; // target was removed by previous operation, ignore this
+    }
+    if (target->isChord() && target->ticks() > duration) {
+        target = replaceWithRest(target); // prevent unexpected note splitting
+    }
+    Segment* segment = target->segment();
+    if (segment->measure()->isMMRest()) {
+        Measure* m = segment->measure()->mmRestFirst();
+        segment = m->findSegment(SegmentType::ChordRest, m->tick());
+    }
+    segment = target->score()->setNoteRest(segment, target->track(),
+                                           with->noteVal(), duration, Direction::AUTO, false, {}, false, &target->score()->inputState());
+    return toChord(segment->nextChordRest(target->track()))->upNote();
+}
+
+static Element* prepareTarget(Element* target, Note* with, const Fraction& duration)
+{
+    if (target->isNote() && toNote(target)->chord()->ticks() != duration) {
+        return prepareTarget(toNote(target)->chord(), with, duration);
+    }
+    if (target->isChordRest() && toChordRest(target)->ticks() != duration) {
+        return prepareTarget(toChordRest(target), with, duration);
+    }
+    return target;
+}
+
+static bool canPasteStaff(XmlReader& reader, const Fraction& scale)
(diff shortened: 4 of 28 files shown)


src/engraving/libmscore/chordrest.cpp
```diff
@@ -1433,7 +1433,7 @@ void ChordRest::removeMarkings(bool /* keepTremolo */)
 bool ChordRest::isBefore(const ChordRest* o) const
 {
     if (!o || this == o) {
-        return true;
+        return false;
```
wizofaus, line 1436: Prevent asserts due to unstable std::sort predicates.

src/engraving/libmscore/cmd.cpp
```diff
@@ -1049,6 +1053,13 @@ Fraction Score::makeGap(Segment* segment, int track, const Fraction& _sd, Tuplet
             tuplet = 0;
         } else {
             if (seg != firstSegment || !keepChord) {
+                if (cr->isChord()) {
+                    for (auto note : toChord(cr)->notes()) {
+                        _selection.remove(note);
+                    }
+                } else {
+                    _selection.remove(cr);
+                }
```
wizofaus, line 1062: This logic should probably be in Score::removeElement( ) but wanted to minimize impact

src/engraving/libmscore/cmd.cpp
```diff
@@ -3773,7 +3784,7 @@ void Score::cmdPitchUp()
     if (el && el->isLyrics()) {
         cmdMoveLyrics(toLyrics(el), Direction::UP);
     } else if (el && (el->isArticulation() || el->isTextBase())) {
-        el->undoChangeProperty(Pid::OFFSET, el->offset() + PointF(0.0, -MScore::nudgeStep* el->spatium()), PropertyFlags::UNSTYLED);
+        el->undoChangeProperty(Pid::OFFSET, el->offset() + PointF(0.0, -MScore::nudgeStep * el->spatium()), PropertyFlags::UNSTYLED);
```
wizofaus, line 3787: hmm, uncrustify version incompatibilities strike again

src/engraving/libmscore/edit.cpp
```diff
@@ -453,7 +453,7 @@ Rest* Score::setRest(const Fraction& _tick, int track, const Fraction& _l, bool
             // compute list of durations which will fit l
             //
             std::vector<TDuration> dList;
-            if (tuplet || staff->isLocalTimeSignature(tick)) {
+            if (tuplet || staff->isLocalTimeSignature(tick) || f == Fraction(0, 1)) {
```
wizofaus, line 456: was crashing without this when pasting, e.g. a whole note onto a multi-measure rest, the 'else' part of this test definitely doesn't work if `f` is 0 anyway

src/engraving/libmscore/paste.cpp
```diff
@@ -1059,48 +1071,37 @@ void Score::cmdPaste(const QMimeData* ms, MuseScoreView* view, Fraction scale)
...
-                    // (ChordRest::drop() will select it for us)
-                    if (targetType == ElementType::NOTE) {
-                        select(nel);
+                    // duration information needs to be provided separately in this case
+                    if (target->isNote() && toNote(target)->chord()->ticks() != duration) {
+                        newEl = replaceNote(toNote(target)->chord(), toNote(newEl), duration);
+                    } else if (target->isChordRest()) {
+                        newEl = replaceNote(toChordRest(target), toNote(newEl), duration);
+                    } else {
+                        target->drop(ddata);
```
wizofaus, line 1092: would prefer that above logic was all inside Note::drop( ) and ChordRest::drop( ) but would require adding even more info (duration) to the already-bloated EditData structure...

Jojo-Schmitz:
There's a merge commit that doesn't belong into this PR.
Is this supposed to fix https://musescore.org/en/node/321809 for master? Alternative to the 2nd commit for #8818?

wizofaus:
> There's a merge commit that doesn't belong into this PR.
> Is this supposed to fix https://musescore.org/en/node/321809 for master? Alternative to the 2nd commit for #8818?

That came from a rebase.  I tried to squash but it wouldn't play nicely. Need to fix the code-style thing though.

wizofaus:
And when I say "fix", I mean screw up my sensible formatting to suit the current version of uncrustify used by the git workflows.

Jojo-Schmitz:
As far as I can tell this PR, on dropping a notehead of a quarter note onto an MMRest, pastes a quarter note, not a note equal to a the full lenthg of the measure?

wizofaus:
It preserves the length of the note you copied, which is consistent with how it works when pasted onto anything else.
That's a new behaviour of MU4 (MU3 only ever transferred the pitch as the clipboard data didn't even include the note duration).

Jojo-Schmitz:
But that then is not what you describe as the desired outcome in section a) of 8824

wizofaus:
Yes it is, because pasting on to a whole measure rest in MU4 *does* preserve the length of the copied note (adding/removing rests as required to fit it on).

Jojo-Schmitz:
_Pasting a note on to a multi-measure rest should behave the same as pasting onto a whole rest,_
That's not what it does

Jojo-Schmitz:
Hmm, OK, not what it did in 3.x...

wizofaus:
BTW this is not ready for merge yet, still a number of issues I just discovered.

RobFog:
> BTW this is not ready for merge yet, still a number of issues I just discovered.

You can mark the pull request as a draft.

wizofaus:
I'd looked for that option (convert to draft), couldn't see it, it's not very obvious!

src/engraving/libmscore/chordrest.cpp
```diff
@@ -519,16 +519,17 @@ Element* ChordRest::drop(EditData& data)
...
-        delete e;
+        Segment* seg = segment();
+        score()->undoRemoveElement(this);
+        Chord* chord = new Chord(score());
+        chord->setTrack(track());
+        chord->setDurationType(durationType());
+        chord->setTicks(ticks());
+        chord->setTuplet(tuplet());
+        chord->add(note);
+        score()->undoAddCR(chord, seg->measure(), seg->tick());
```
wizofaus, line 530: existing method failed to transfer all properties and didn't work with tuplets etc.

src/engraving/libmscore/cmd.cpp
```diff
@@ -777,15 +777,19 @@ Segment* Score::setNoteRest(Segment* segment, int track, NoteVal nval, Fraction
...
             expandVoice(segment, track);
         }
-
+        if (targetIsRest && !cr->isRest()) {
+            undoRemoveElement(cr);
+            segment = addRest(segment, track, cr->ticks(), cr->tuplet())->segment();
+        }
         // the returned gap ends at the measure boundary or at tuplet end
-        Fraction dd = makeGap(segment, track, sd, cr ? cr->tuplet() : 0);
+        Fraction dd = makeGap(segment, track, sd, tuplet);
```
wizofaus, line 792: the changes here are about ensuring that a) if the element at the segment position is a rest, then any other notes that would be overlapped get removed instead of being "split" into component durations - actually I'd rather simplify this so that setNoteRest *always* does that but it may have unexpected side effects and b) only the *initial* tuplet is preserved where possible - the existing logic tried to preserve all tuplets that fell inside the duration, which didn't make sense for any case I could think of

src/engraving/libmscore/cmd.cpp
```diff
@@ -1026,7 +1033,7 @@ Fraction Score::makeGap(Segment* segment, int track, const Fraction& _sd, Tuplet
             }
         }
         Tuplet* ltuplet = cr->tuplet();
-        if (ltuplet != tuplet) {
+        if (ltuplet != tuplet || (tuplet && sd > tuplet->ticks())) {
```
wizofaus, line 1036: breaking change, investigating

src/engraving/libmscore/cmd.cpp.uncrustify
```diff
@@ -0,0 +1,4438 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
```
Jojo-Schmitz, line 2: I guess this file doesn't belong into the PR

Jojo-Schmitz:
Issue: select note(head), use "Paste double/half duration". It keeps the duration rather than doubling/halfing

wizofaus:
Good point, yes, I'd expect that to work.

src/engraving/libmscore/cmd.cpp.uncrustify
wizofaus: I'd removed it within minutes!

src/engraving/tests/tst_copypaste.cpp
```diff
@@ -82,12 +83,21 @@ private slots:
...
+    void copypasteTuplet02() { copypastetuplet("02"); }
+    void copypasteQtrNoteOntoWholeRest() { copypastenote("01"); }
+    void copypasteQtrNoteOntoWholeNote() { copypastenote("02"); }
+    void copypasteQtrNoteOntoQtrRest() { copypastenote("03"); }
+    void copypasteQtrNoteOntoQtrNote() { copypastenote("04"); }
+    void copypasteWholeNoteOntoQtrNote() { copypastenote("05"); }
+    void copypasteWholeNoteOntoQtrRest() { copypastenote("06"); }
+    void copypasteQtrNoteOntoTriplet() { copypastenote("07"); }
+    void copypasteWholeNoteOntoTriplet() { copypastenote("08"); }
+    void copypasteQtrNoteIntoChord() { copypastenote("09"); }
```
wizofaus, line 98: pretty minimal set of tests, there's quite a few more edge cases I'd like to have tests for, but for some them the "desired" behaviour isn't necessarily what we're getting now, I guess at least having a test to ensure it doesn't crash would be good!

src/engraving/libmscore/paste.cpp
```diff
@@ -1031,6 +1031,60 @@ void Score::pasteSymbols(XmlReader& e, ChordRest* dst)
...
+    if (target->isNote() && toNote(target)->chord()->ticks() != duration) {
+        return prepareTarget(toNote(target)->chord(), with, duration);
+    }
+    if (target->isChordRest() && toChordRest(target)->ticks() != duration) {
+        return prepareTarget(toChordRest(target), with, duration);
+    }
+    return target;
+}
+
+static bool canPasteStaff(XmlReader& reader, const Fraction& scale)
```
wizofaus, line 1069: to prevent crashing when using "paste half length" on a range with 1/1024th notes, a bit kludgy but I tried a few other ways and this was the more reliable and least code by far

src/engraving/libmscore/paste.cpp
```diff
@@ -1129,9 +1164,12 @@ void Score::cmdPaste(const QMimeData* ms, MuseScoreView* view, Fraction scale)
                 qDebug("paste <%s>", data.data());
             }
             XmlReader e(data);
-            e.setPasteMode(true);
-            if (!pasteStaff(e, cr->segment(), cr->staffIdx(), scale)) {
-                return;
+            if (canPasteStaff(e, scale)) {
+                XmlReader e(data);
```
wizofaus, line 1168: XmlReader doesn't have a rewind...should add one perhaps?

RomanPudashkin: You already declared XmlReader e(data) at line 1166. We will get a warning here

src/engraving/libmscore/cmd.cpp.uncrustify
Jojo-Schmitz: But not pushed it equally fast I guess? Anyway, issue is sorted now.

src/engraving/libmscore/paste.cpp
Jojo-Schmitz: Declared? Rather called. So it might get called twice now.

wizofaus: Correct,  you can't rewind it so you have to create a new one... happy to add a rewind function

wizofaus: They're in different scopes so the behavior is well defined, but I agree stylistically some may prefer avoiding reusing the name (actually I don't like single letter variable names generally!).

wizofaus: Took your suggestion of creating a separate method, creating a rewind method was a bit painful/dangerous due to inner workings of Qt.

RomanPudashkin (approved):

vshalkevich:
Works fine

PR #8830, opened 2021-08-10
[MU4] notation_tracks_workflow

### LINKED ISSUE #8457: MU4 crashes when creating parts
MU4 crashes when a part is selected and the "open" button is pressed. 

See these two videos for reproduction: 

https://www.dropbox.com/s/3zojyrv8y4jnt4s/Crashes_Parts-1.mov?dl=0
https://www.dropbox.com/s/ihnucl5liok7w94/Crashes_Parts-2.mov?dl=0

Note: In the first video, the crash only occurred when the "piano" part was opened; the "vocal" part opened correctly. In my second video, it was opening the "vocal" part that triggered the crash. 

MU4 should not crash when a part is opened. 

Apple MacBook Air M1 running OS11.4

vpereverzev:
closes #8457 

Also includes fixes for the MixerChannels processing and memory leaks in NotationParts


src/playback/internal/playbackcontroller.cpp
```diff
@@ -538,6 +544,29 @@ void PlaybackController::setCurrentSequence(const TrackSequenceId sequenceId)
...
+
+    for (const Part* part : partList) {
+        addTrack(part->id(), part->partName().toStdString());
+    }
+
+    partList.onItemAdded(this, [this](const Part* part) {
+        addTrack(part->id(), part->partName().toStdString());
+    });
+
+    partList.onItemAdded(this, [this](const Part* part) {
```
cbjeukendrup, line 565: Shouldn't this be `onItemRemoved`?

vpereverzev: oh, wow
nice catch, thx

src/notation/imasternotationmididata.h
```diff
@@ -29,22 +29,26 @@
...
                                         const midi::tick_t toTick) const = 0;
     virtual midi::Events retrieveEventsForElement(const Element* element, const midi::channel_t midiChannel) const = 0;
     virtual std::vector<midi::Event> retrieveSetupEvents(const std::list<InstrumentChannel*> instrChannel) const = 0;
 };
 
-using INotationMidiEventsPtr = std::shared_ptr<INotationMidiEvents>;
+using IMasterNotationMidiDataPtr = std::shared_ptr<IMasterNotationMidiData>;
 }
 
 #endif // MU_NOTATION_INOTATIONMIDIDATA_H
```
Eism, line 54: MU_NOTATION_IMASTERNOTATIONMIDIDATA_H

src/notation/notationtypes.h
```diff
@@ -136,6 +136,7 @@ using InstrumentGenreList = QList<const InstrumentGenre*>;
 using ScoreOrderList = QList<const ScoreOrder*>;
 using InstrumentGroupList = QList<const InstrumentGroup*>;
 using MidiArticulationList = QList<MidiArticulation>;
+using PartId = ID;
```
Eism, line 139: I do not agree with this. Making aliases for objects is not a good idea because we have many objects and will have many aliases. Just use an id with a named argument

src/notation/internal/masternotation.cpp
```diff
@@ -77,8 +78,8 @@ void MasterNotation::setMasterScore(Ms::MasterScore* score)
     }
 
     setScore(score);
-
     initExcerptNotations(masterScore()->excerpts());
+    m_notationMidiData->init(m_parts);
```
Eism, line 82: I think, better to pass m_parts to MasterNotationMidiData as a parameter in constructor like m_notationChanged
and there init what you need by parts changed notification

Then you do not need to call notationMidiData::init in several places

src/notation/internal/masternotationmididata.h
```diff
@@ -97,9 +113,14 @@ class NotationMidiEvents : public INotationMidiEvents, public async::Asyncable
...
+
     std::unique_ptr<Ms::MidiRenderer> m_midiRenderImpl = nullptr;
     IGetScore* m_getScore = nullptr;
+    INotationPartsPtr m_parts = nullptr;
 };
+
+using MasterNotationMidiDataPtr = std::shared_ptr<MasterNotationMidiData>;
 }
 
 #endif // MU_NOTATION_NOTATIONMIDEVENTS_H
```
Eism, line 126: MU_NOTATION_MASTERNOTATIONMIDIDATA_H

src/playback/internal/playbackcontroller.cpp
```diff
@@ -194,45 +202,53 @@ void PlaybackController::playElement(const notation::Element* element)
...
-    IF_ASSERT_FAILED(project) {
+    if (!m_notation || !m_masterNotation) {
         return;
     }
 
-    io::path notationKey = project->path();
+    INotationPartsPtr notationParts = m_notation->parts();
 
-    auto search = m_sequenceIdMap.find(notationKey.toStdString());
+    for (const Part* part : m_masterNotation->parts()->partList()) {
```
Eism, line 238: better to move this code to a separate method like muteNonexistentParts, because it is not clear what is happening here

src/notation/internal/masternotationmididata.h
```diff
@@ -31,18 +31,24 @@
...
 namespace mu::notation {
-class NotationMidiEvents : public INotationMidiEvents, public async::Asyncable
+class MasterNotationMidiData : public IMasterNotationMidiData, public async::Asyncable
 {
     INJECT(notation, INotationConfiguration, configuration)
 
 public:
-    explicit NotationMidiEvents(IGetScore* getScore, async::Notification notationChanged);
+    explicit MasterNotationMidiData(IGetScore* getScore, async::Notification notationChanged);
+    ~MasterNotationMidiData();
```
RomanPudashkin, line 45: You missed override here

src/notation/internal/masternotationmididata.cpp
```diff
@@ -0,0 +1,576 @@
...
+}
+
+void MasterNotationMidiData::init(INotationPartsPtr parts)
+{
+    IF_ASSERT_FAILED(score() && parts) {
+        return;
+    }
+
+    m_parts = std::move(parts);
+    m_midiRenderImpl = std::unique_ptr<Ms::MidiRenderer>(new Ms::MidiRenderer(score()));
```
RomanPudashkin, line 54: You can use make_unique here

src/notation/internal/masternotationmididata.cpp
```diff
@@ -0,0 +1,576 @@
...
+
+    return make_ret(Ret::Code::Ok);
+}
+
+Ret MasterNotationMidiData::playChordMidiData(const Ms::Chord* chord) const
+{
+    Ms::Part* part = chord->part();
+    Ms::Fraction tick = chord->segment() ? chord->segment()->tick() : Ms::Fraction(0, 1);
+    Ms::Instrument* instr = part->instrument(tick);
+    channel_t midiChannel = instr->channel(chord->notes()[0]->subchannel())->channel();
```
RomanPudashkin, line 313: Please, add a check on emptiness here

src/notation/internal/masternotation.cpp
vpereverzev: might make sense, thanks

src/notation/notationtypes.h
vpereverzev: There is a couple of cons of that approach, but in general I'm not against

src/notation/internal/masternotationmididata.h
vpereverzev: thx, fixed

src/notation/internal/masternotationmididata.cpp
vpereverzev: thx, fixed

src/notation/internal/masternotationmididata.h
vpereverzev: thx

src/notation/internal/masternotation.cpp
vpereverzev: Actually, no

I have to be sure that the pointer on score is valid at the moment of initialization. I could move parts subscriptions to the constructor, but it doesn't make sense to process them if I have no score at the moment.

I'll leave it as it was

src/notation/internal/masternotationmididata.cpp
vpereverzev: well, if there is a chord without any note, than it should be an exceptional case
added IF_ASSERT statement

RomanPudashkin (approved):

Issue #8823, opened 2021-08-09
[MU4 Task] Implement Selection Filter Panel


cbjeukendrup:
**Task description**
The Selection filter panel is not yet implemented. 

_The selection filter panel in MU3:_
<img width="537" alt="Schermafbeelding 2021-08-09 om 16 46 52" src="https://user-images.githubusercontent.com/48658420/128725885-4e7e0a58-ef19-4c34-ab25-15947b6ab7e2.png">

cbjeukendrup:
I would like to implement this.

wizofaus:
My only comment would be that while at this stage there's no point including extra element types, I'm planning to "fix" copy/paste to support ALL element types including time signatures etc.  So the ability to filter them out for a range selection will be needed at that point - just something to keep in mind.
Might need to consider what the reasoning behind the order of the element types listed there is?

PR #8832, opened 2021-08-10
Implement Selection Filter Panel

cbjeukendrup:
Resolves: #8823 

https://user-images.githubusercontent.com/48658420/128937045-f5b598e6-7815-4f02-9150-8075cd63af57.mov

When there is no notation, the panel becomes disabled.


src/notation/view/selectionfiltermodel.cpp
```diff
@@ -0,0 +1,233 @@
...
+{
+    if (!currentNotationInteraction()) {
+        return;
+    }
+
+    m_isNoneSelected = true;
+    m_isAllSelected = true;
+
+    // Skip "All" type
+    for (auto type : m_types.mid(1)) {
```
RomanPudashkin, line 69: I think, better to add more explicit skipping for this element:

for (SelectionFilterType type : m_types) {
if (type == All) {
continue;
}
...
}

Because this code will break if we change the position of the All type in the list

src/notation/view/selectionfiltermodel.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "translation.h"
+
+using namespace mu::notation;
+
+SelectionFilterModel::SelectionFilterModel(QObject* parent)
+    : QAbstractListModel(parent)
+{
+}
+
+void SelectionFilterModel::load()
```
RomanPudashkin, line 35: Please, add TRACEFUNC here

src/notation/internal/notationinteraction.cpp
```diff
@@ -576,6 +576,29 @@ mu::async::Notification NotationInteraction::selectionChanged() const
...
+void NotationInteraction::setSelectionTypeFiltered(SelectionFilterType type, bool filtered)
+{
+    score()->selectionFilter().setFiltered(type, filtered);
+    if (selection()->isRange()) {
+        score()->selection().updateSelectedElements();
+        notifyAboutSelectionChanged();
+    }
+}
+
+void NotationInteraction::setAllSelectionTypesFiltered(bool filtered)
```
RomanPudashkin, line 593: It seems that you can use the All type and setSelectionTypeFiltered instead of adding the separate method for this. Like this:

void NotationInteraction::setSelectionTypeFiltered(SelectionFilterType type, bool filtered)
{
if (type == All) {
    score()->selectionFilter().setAllFiltered(filtered);
} else {
    score()->selectionFilter().setFiltered(filtered);
}
...
}

Probably we can combine setAllFiltered and setFiltered too

cbjeukendrup: Thanks, good idea! Fixed.

RomanPudashkin (approved):

PR #8833, opened 2021-08-11
317098 musicxml white space master

lvinken:
Resolves:
https://musescore.org/en/node/317098
https://musescore.org/en/node/317099

#317098:
The MusicXML importer (incorrectly) uses QXmlStreamReader::readNext() to skip to the end of an element.
This works correctly only if the element is emtpy, which cannot always be guaranteed. The readNext() calls
are replaced by QXmlStreamReader::skipCurrentElement(), which will always work.

#317099:
The MusicXML parser contains several assertions checking start and end elements to verify the parser is still in sync.
For end users they have no effect (resulting in incomplete import without any error), for a debug build they crash
MuseScore when triggered.
The assertions have been replaced by a dialog providing meaningful information (expected element name and type versus actual) to the end user and the developer.


lvinken:
Used "push -f" to fix failed code style check.

RomanPudashkin (approved):

RomanPudashkin:
rebase needed

Issue #8847, opened 2021-08-12
[MU4 Issue] Pasting ties causes crash


wizofaus:
**Describe the bug**
Pasting a tie onto a note or rest causes MU4 to crash, and ctrl+x doesn't work when slur selected

**To Reproduce**
1. Add two notes of the same pitch and tie them.
2. Copy the tie to the clipboard
3. Try pasting it on a note or rest
4. MuseScore crashes

Also: ctrl+x when slur selected doesn't activate "cut", as slur segment is interpreting "X" as "flip".

**Expected behavior**
Pasting ties onto rests should do nothing, it makes no sense.
Pasting on to note might potentially be useful but MU3 doesn't support it so for now simplest just to ignore.

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**
Was introduced by my earlier changes to allow copying/pasting all spanners (ties will have to be the exception for now!)

wizofaus:
Fixed

PR #8848, opened 2021-08-12
Fix pasting ties

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8847

Some of my previous improvements had broken pasting ties and cutting via Ctrl+X

### DIFF
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -2009,11 +2009,6 @@ Element* Note::drop(EditData& data)
         n->setTpc2(Ms::transposeTpc(n->tpc1(), v, true));
         // replace this note with new note
         n->setParent(ch);
-        if (this->tieBack()) {
-            n->setTieBack(this->tieBack());
-            n->tieBack()->setEndNote(n);
-            this->setTieBack(nullptr);
-        }
         score()->undoRemoveElement(this);
         score()->undoAddElement(n);
     }
--- a/share/plugins/colornotes.qml
+++ b/share/plugins/colornotes.qml
@@ -78,9 +78,9 @@ MuseScore {
                                     note.dots[i].color = colors[note.pitch % 12];
                               else
                                     note.dots[i].color = black;
-                              }
                         }
                   }
+            }
       }
 
       onRun: {
--- a/share/plugins/colornotes.qml
+++ b/share/plugins/colornotes.qml
@@ -49,17 +49,13 @@ MuseScore {
 
       function applyToNotesInSelection(func) {
             var fullScore = !curScore.selection.elements.length
-            if (fullScore) {
+            if (fullScore)
                   cmd("select-all")
-                  curScore.startCmd()
-            }
             for (var i in curScore.selection.elements)
                   if (curScore.selection.elements[i].pitch)
                         func(curScore.selection.elements[i])
-            if (fullScore) {
-                  curScore.endCmd()
+            if (fullScore)
                   cmd("escape")
-            }
       }
 
       function colorNote(note) {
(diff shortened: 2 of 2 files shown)


src/engraving/libmscore/tie.cpp
```diff
@@ -98,12 +98,12 @@ bool TieSegment::edit(EditData& ed)
 {
     SlurTie* sl = tie();
 
-    if (ed.key == Qt::Key_X) {
+    if (ed.key == Qt::Key_X && !ed.modifiers) {
```
wizofaus, line 101: TBH I don't like this, I'd rather this be in response to the "flip placement" command which is normally mapped to "X", but this is a minimal fix.
Note that MU3 allows using, e.g. CTRL+ALT+X to flip a slur if CTRL+ALT+X isn't mapped to another command, but this isn't really expected/documented behaviour.

Jojo-Schmitz:
and contains multiple commits that don't belong here. Also vails tjhe vtests?

src/engraving/libmscore/rest.cpp
```diff
@@ -188,7 +188,7 @@ bool Rest::acceptDrop(EditData& data) const
         ) {
         return true;
     }
-    return type != ElementType::SLUR && e->isSpanner();
+    return type != ElementType::SLUR && type != ElementType::TIE && e->isSpanner();
```
Jojo-Schmitz, line 191: This is the same as my #8818

Jojo-Schmitz: Except for a comment:
`// prevent 'hanging' slurs, avoid crash on tie`

wizofaus: > It contains multiple commits that don't belong here. Also fails the vtests?

Something's really odd with one of my local repos, that happens quite a lot.
The vtests is something random, and there's no obvious way to re-run (other than pushing more commits)

RomanPudashkin: @Jojo-Schmitz

Your PR has been merged

wizofaus: But this diff doesn't show it...

wizofaus:
Very odd, somehow all these changes already got merged...

Jojo-Schmitz:
Not odd at all, I added them to my PR ;-)

src/engraving/libmscore/rest.cpp
Jojo-Schmitz: > Something's really odd with one of my local repos, that happens quite a lot.
May need to clean your repo, `git gc` or some such?

> The vtests is something random, and there's no obvious way to re-run (other than pushing more commits)
`git commit --amend; git push -f` is what I use in sich cases.

src/engraving/libmscore/note.cpp
```diff
@@ -2009,11 +2009,6 @@ Element* Note::drop(EditData& data)
         n->setTpc2(Ms::transposeTpc(n->tpc1(), v, true));
         // replace this note with new note
         n->setParent(ch);
-        if (this->tieBack()) {
-            n->setTieBack(this->tieBack());
-            n->tieBack()->setEndNote(n);
-            this->setTieBack(nullptr);
-        }
```
Jojo-Schmitz, line 2016: I'm confused now, is that the chage you really wanted?

Jojo-Schmitz: If so: this has not been part of my (meanwhile merged) PR

Jojo-Schmitz: It is however partt of your #8825, and as such in the current code base. Should it get removed again, ir is this just another tell-tale sign of a botched local repo?

PR #8850, opened 2021-08-13
[MU4] Fixed creating blank part

Eism:


Eism:
alternative solution #8887 8887

PR #8867, opened 2021-08-17
ScoreFont / Sym / SymId / SymNames cleanup 

cbjeukendrup:
- Moved `SymNames` from `Sym` class to a dedicated struct
- Replaced `Sym` class with a private struct inside `ScoreFont`
- Removed practically unused `scale` parameter from `ScoreFont::draw`


src/engraving/libmscore/textbase.cpp
```diff
@@ -2393,7 +2390,7 @@ void TextBase::writeProperties(XmlWriter& xml, bool writeText, bool /*writeStyle
     }
 }
 
-static constexpr std::array<Pid, 18> pids { {
+static constexpr std::array<Pid, 18> TextBasePropertyId { {
```
Jojo-Schmitz, line 2394: Part of #8798

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -466,7 +466,7 @@ void TextLineBaseSegment::spatiumChanged(qreal ov, qreal nv)
     _endText->spatiumChanged(ov, nv);
 }
 
-static constexpr std::array<Pid, 26> pids = { {
+static constexpr std::array<Pid, 26> TextLineBasePropertyId = { {
```
Jojo-Schmitz, line 469: Part of #8798

src/engraving/libmscore/textbase.cpp
```diff
@@ -2416,7 +2413,7 @@ static constexpr std::array<Pid, 18> pids { {
 bool TextBase::readProperties(XmlReader& e)
 {
     const QStringRef& tag(e.name());
-    for (Pid i :pids) {
+    for (Pid i : TextBasePropertyId) {
```
Jojo-Schmitz, line 2417: dito

src/engraving/libmscore/textbase.cpp
```diff
@@ -2468,7 +2465,7 @@ Pid TextBase::propertyId(const QStringRef& name) const
     if (name == "text") {
         return Pid::TEXT;
     }
-    for (Pid pid : pids) {
+    for (Pid pid : TextBasePropertyId) {
```
Jojo-Schmitz, line 2469: dito

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -501,7 +501,7 @@ static constexpr std::array<Pid, 26> pids = { {
 
 Element* TextLineBaseSegment::propertyDelegate(Pid pid)
 {
-    for (Pid id : pids) {
+    for (Pid id : TextLineBasePropertyId) {
```
Jojo-Schmitz, line 504: dito

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -568,7 +568,7 @@ void TextLineBase::spatiumChanged(qreal /*ov*/, qreal /*nv*/)
 
 void TextLineBase::writeProperties(XmlWriter& xml) const
 {
-    for (Pid pid : pids) {
+    for (Pid pid : TextLineBasePropertyId) {
```
Jojo-Schmitz, line 571: dito

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -583,7 +583,7 @@ void TextLineBase::writeProperties(XmlWriter& xml) const
 bool TextLineBase::readProperties(XmlReader& e)
 {
     const QStringRef& tag(e.name());
-    for (Pid i : pids) {
+    for (Pid i : TextLineBasePropertyId) {
```
Jojo-Schmitz, line 586: dito

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -598,7 +598,7 @@ bool TextLineBase::readProperties(XmlReader& e)
 
 Pid TextLineBase::propertyId(const QStringRef& name) const
 {
-    for (Pid pid : pids) {
+    for (Pid pid : TextLineBasePropertyId) {
```
Jojo-Schmitz, line 601: dito

src/engraving/libmscore/textlinebase.cpp
cbjeukendrup: True. It was needed to make this PR work. I did exactly the same as in that PR, so that it can be auto-merged and no conflict will occur (hopefully).

Jojo-Schmitz: I see

src/engraving/libmscore/textbase.cpp
```diff
@@ -1845,16 +1842,17 @@ void TextBase::createLayout()
...
                     symState = false;
-                    SymId id = Sym::name2id(sym);
+                    SymId id = SymNames::symIdByName(sym);
                     if (id != SymId::noSym) {
-                        CharFormat fmt = *cursor.format();  // save format
-                                                            // uint code = score()->scoreFont()->sym(id).code();
-                        uint code = ScoreFont::fallbackFont()->sym(id).code();
+                        CharFormat fmt = *cursor.format(); // save format
+
+                        //uint code = score()->scoreFont()->symCode(id);
```
wizofaus, line 1849: think this can be removed?

src/engraving/libmscore/textbase.cpp
```diff
@@ -1874,7 +1872,7 @@ void TextBase::createLayout()
                 } else if (token == "quot") {
                     insert(&cursor, '"');
                 } else {
-                    // TODO insert(&cursor, Sym::name2id(token));
+                    // TODO insert(&cursor, SymNames::symIdByName(token));
```
wizofaus, line 1875: indeed?

Jojo-Schmitz:
rebase needed

src/engraving/libmscore/scorefont.h
```diff
@@ -65,7 +68,6 @@ class ScoreFont
     static ScoreFont* fallbackFont();
     static const char* fallbackTextFont();
 
-    Sym sym(SymId id) const;
     uint symCode(SymId id) const;
     QString toString(SymId id) const;
```
wizofaus, line 72: master has an extra function in here (fromCode), I know because I put it there! Not sure why it's gone missing...

cbjeukendrup: Because I had not rebased yet :)
Now it's there.

Jojo-Schmitz:
Like... now?

cbjeukendrup:
> Like... now? The Unity issues are solved for me

Yes, thanks!

Jojo-Schmitz:
That empty src/engraving/infrastructure/draw/painterpath.cpp.uncrustify  doesn't belong into this PR

cbjeukendrup:
> That empty src/engraving/infrastructure/draw/painterpath.cpp.uncrustify doesn't belong into this PR

Oops, fixed!
And rebased.

Jojo-Schmitz:
I believe tools/fonttools/smufl2sym.sh and friends need to get adjusted to still work after this PR

cbjeukendrup:
> I believe tools/fonttools/smufl2sym.sh and friends need to get adjusted to still work after this PR

Done!

cbjeukendrup:
Rebased.

cbjeukendrup:
@RomanPudashkin it would be great if you or someone else from the team could review and merge this PR, so that other work can be built on top of it.

igorkorsukov:
@cbjeukendrup Please explain why you decided to move the table with symbol names from `.cpp` to `.h` file?

Will this not lead to such problems:
* Compilation takes longer because the `.h` file has become large (about 220 kB) and the preprocessor will need to insert it into many other files
* Possible memory overuse, if this file is included in different libraries (modules), probably indirectly, then several instances of this large table will probably be instantiated, each library (module) has its own.

cbjeukendrup:
The reason was that we can now make use of `constexpr`, so the table will be compiled into the executable, instead of that it has to be initialized at runtime. 
But your concerns are probably valid, especially the first one. About the second one, I'm not sure, and I haven't yet found documentation that gives clarity about it.

We have to make a consideration: either constexpr and a faster runtime (if that's true at all) or hiding it in a cpp file and a faster compile time... what do you think?

Jojo-Schmitz:
About how much time are we talking? In either case.

Jojo-Schmitz:
I tried it out, the build time (and on my quite powerfull maching): plain master 59 seconds, your PR, 1 minute. Hand-stopped so not 100% accurate. So I'd say the compile time is the same

igorkorsukov:
> The reason was that we can now make use of `constexpr`, so the table will be compiled into the executable, instead of that it has to be initialized at runtime.
> But your concerns are probably valid, especially the first one. About the second one, I'm not sure, and I haven't yet found documentation that gives clarity about it.
> 
> We have to make a consideration: either constexpr and a faster runtime (if that's true at all) or hiding it in a cpp file and a faster compile time... what do you think?

I build your branch, as far as I can see, in the engraving library this table is instantiated 5 times (present 5 times in `libengraving.a`). In `libnotation.a`, it is instantiated once.

It would be nice to measure a benchmark on the speed of both solutions.
In general, I think, yes, these functions are called frequently, but not so frequently that the difference in both solutions matters in our case.
For example, we can add the `TRACEFUNC` macro for the test, it will not measure such a small-time correctly, but it will be able to count the number of calls for some big score in order to understand what to measure in the benchmark.

I think we have much more potential for speeding up in algorithms (do not do too much, do not duplicate calls, etc.) :)

cbjeukendrup:
I decided to move the tables back to the cpp file, but added `constexpr` to the _definition_ instead of to the _declaration_. I guess that makes more sense.

RomanPudashkin (approved):

Issue #8856, opened 2021-08-15
[MU4 Issue] Load style command does nothing


wizofaus:
**Describe the bug**
Whether from the Format menu or context menu from right clicking on an empty part of the score, choosing "Load style" does nothing.

**To Reproduce**
1. Open a score
2. Choose Format|Load Style
3. ...?

**Expected behavior**
Dialog to allow loading a style file should appear as per MU3.
 
And "Save style" is missing entirely?

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**

wizofaus:
BTW I started looking at this and while it should be reasonably straightforward to implement:
a) MU3 always prompts the user with a warning if the .mss file version does not exactly match the current version, is that behaviour still what we want (the warning is quite stark, saying "the style file is not compatible with this version")?  Some of the .mss files that are being built into the install are still v3 so would cause this warning - I assume they need to be updated?
b) MU3 does nothing at all if you attempt to load an .mss file that can't be parsed or read (e.g. a random text file or one you don't have permission to read) - is that the right behaviour?
c) the "Save Style" command isn't even being shown currently, but presumably we want to keep this too?

@Tantacrul

Tantacrul:
This seems to have simply been missed. We would want to keep MS3 functionality here as a first step, so these options (and warnings) should be reinstated. I don’t have a strong opinion about what the default folder would be and if we think the messaging is a bit stark, we can reword it.

cheers!

wizofaus:
Happy to put up a PR but I need my text editing fixes merged first, as it involves adding new functions to the ends of the same files.

wizofaus:
How about this? @Tantacrul 
![image](https://user-images.githubusercontent.com/5364225/129625198-4b181e5a-cf4c-423d-bdc7-42579e113fd3.png)

Jojo-Schmitz:
'an older'` maybe? and/or 'may not be fully compatible with this version'

wizofaus:
The check is literally that for exactly the same version, I don't know why.
Happy to add "and may not be compatible".

Tantacrul:
Agreed ‘older version’ might be better. I also wonder whether we should explain the potential consequences too?

wizofaus:
I wouldn't expect loading an "older" version to give a warning at all, it's only loading a file from a newer version that's likely not to work as expected.

Tantacrul:
[...]

PR #8868, opened 2021-08-17
Load/save style commands

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8856

Load/save style commands not implemented

### DIFF
--- a/src/engraving/style/style.cpp
+++ b/src/engraving/style/style.cpp
@@ -284,6 +284,21 @@ bool MStyle::read(QIODevice* device, bool ign)
     return true;
 }
 
+bool MStyle::isValid(QIODevice* device)
+{
+    XmlReader e(device);
+    while (e.error() == XmlReader::Error::NoError && e.readNextStartElement()) {
+        if (e.name() == "museScore") {
+            while (e.readNextStartElement()) {
+                if (e.name() == "Style") {
+                    return true;
+                }
+            }
+        }
+    }
+    return false;
+}
+
 void MStyle::read(XmlReader& e, compat::ReadChordListHook* readChordListHook)
 {
     TRACEFUNC;
--- a/src/notation/internal/notationstyle.cpp
+++ b/src/notation/internal/notationstyle.cpp
@@ -33,8 +33,8 @@
 using namespace mu::notation;
 using namespace mu::async;
 
-NotationStyle::NotationStyle(IGetScore* getScore)
-    : m_getScore(getScore)
+NotationStyle::NotationStyle(IGetScore* getScore, INotationUndoStackPtr undoStack)
+    : m_getScore(getScore), m_undoStack(undoStack)
 {
 }
 
@@ -88,4 +88,20 @@ void NotationStyle::applyToAllParts()
 Notification NotationStyle::styleChanged() const
 {
     return m_styleChanged;
+}
+
+bool NotationStyle::loadStyle(const mu::io::path& path, bool allowAnyVersion)
+{
+    m_undoStack->prepareChanges();
+    bool result = m_getScore->score()->loadStyle(path.toQString(), allowAnyVersion);
+    m_undoStack->commitChanges();
+    if (result) {
+        styleChanged().notify();
+    }
+    return result;
+}
+
+bool NotationStyle::saveStyle(const mu::io::path& path)
+{
+    return m_getScore->score()->saveStyle(path.toQString());
 }
--- a/src/appshell/internal/applicationactioncontroller.cpp
+++ b/src/appshell/internal/applicationactioncontroller.cpp
@@ -147,7 +147,7 @@ void ApplicationActionController::revertToFactorySettings()
         IInteractive::Button::No
     });
 
-    if (result.standartButton() == IInteractive::Button::Yes) {
+    if (result.standardButton() == IInteractive::Button::Yes) {
         configuration()->revertToFactorySettings();
     }
 }
--- a/src/engraving/infrastructure/interactive/messagebox.cpp
+++ b/src/engraving/infrastructure/interactive/messagebox.cpp
@@ -30,7 +30,7 @@ MessageBox::Button MessageBox::warning(const std::string& title, const std::stri
 #ifndef NO_ENGRAVING_INTERACTIVE
     using namespace mu::framework;
     IInteractive::Result res = interactive()->warning(title, text, { IInteractive::Button::Ok, IInteractive::Button::Cancel });
-    if (res.standartButton() == IInteractive::Button::Ok) {
+    if (res.standardButton() == IInteractive::Button::Ok) {
         return MessageBox::Button::Ok;
     }
 
--- a/src/engraving/libmscore/scorefile.cpp
+++ b/src/engraving/libmscore/scorefile.cpp
@@ -459,7 +459,7 @@ bool Score::saveStyle(const QString& name)
     }
     QFile f(info.filePath());
     if (!f.open(QIODevice::WriteOnly)) {
-        MScore::lastError = tr("Open Style File\n%1\nfailed: %2").arg(info.filePath(), strerror(errno));
+        MScore::lastError = tr("Open Style File %1 failed: %2").arg(info.filePath(), f.errorString());
         return false;
     }
 
--- a/src/engraving/style/style.h
+++ b/src/engraving/style/style.h
@@ -72,6 +72,7 @@ public:
     bool read(QIODevice* device, bool ign = false);
     bool write(QIODevice* device);
     void save(XmlWriter& xml, bool optimize);
+    static bool isValid(QIODevice* device);
 
     void precomputeValues();
 
--- a/src/framework/global/iinteractive.h
+++ b/src/framework/global/iinteractive.h
@@ -99,7 +99,7 @@ public:
         Result(const int& button, bool showAgain)
             : m_button(button), m_showAgain(showAgain) {}
 
-        Button standartButton() const { return static_cast<Button>(m_button); }
+        Button standardButton() const { return static_cast<Button>(m_button); }
         int button() const { return m_button; }
 
         bool showAgain() const { return m_showAgain; }
--- a/src/framework/ui/dev/interactivetestsmodel.cpp
+++ b/src/framework/ui/dev/interactivetestsmodel.cpp
@@ -103,7 +103,7 @@ void InteractiveTestsModel::question()
           interactive()->buttonData(IInteractive::Button::Yes) }, 0,
         IInteractive::Option::WithIcon);
 
-    if (result.standartButton() == IInteractive::Button::Yes) {
+    if (result.standardButton() == IInteractive::Button::Yes) {
         LOGI() << "Yes!!";
     } else {
         LOGI() << "No!!";
@@ -129,7 +129,7 @@ void InteractiveTestsModel::information()
 {
     IInteractive::Result result = interactive()->info("Tuplet cannot cross barlines", "", {}, 0,
                                                       IInteractive::Option::WithIcon | IInteractive::Option::WithShowAgain);
-    LOGD() << interactive()->buttonData(result.standartButton()).text;
+    LOGD() << interactive()->buttonData(result.standardButton()).text;
 }
 
 void InteractiveTestsModel::warning()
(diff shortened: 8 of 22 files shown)


src/notation/view/widgets/editstyle.cpp
```diff
@@ -1587,16 +1587,23 @@ void EditStyle::setValues()
     disableVerticalSpread->setChecked(!styleValue(StyleId::enableVerticalSpread).toBool());
 }
 
+mu::io::path EditStyle::selectStyleFile(std::shared_ptr<mu::framework::IInteractive> interactive,
+                                        std::shared_ptr<INotationConfiguration> configuration, bool forLoad)
```
wizofaus, line 1591: I'd rather put this somewhere else but I'm not sure where...

src/notation/inotationinteraction.h
```diff
@@ -194,6 +194,9 @@ class INotationInteraction
     virtual void toggleBold() = 0;
     virtual void toggleItalic() = 0;
     virtual void toggleUnderline() = 0;
+
+    virtual void loadStyle(const mu::io::path&) = 0;
+    virtual void saveStyle(const mu::io::path&) = 0;
```
RomanPudashkin, line 199: Better to move those methods to INotationStyle

wizofaus: Are you sure? For loadStyle() need the startEdit()/apply()/notifyAbout... functions from NotationInteraction too.
I'd agree it might be a good place to put the "selectStyleFile" function that shows the load/save dialog for style files.

RomanPudashkin: You can pass NotationUndoStack to NotationStyle:

Screenshot 2021-08-18 at 09 41 28

But I don't understand why you need notifyAboutSelectionChanged in loadStyle. Could you explain please?

wizofaus: So the properties panel can update itself as necessary.

RomanPudashkin: Better to use styleChanged() in this case, because we can't guarantee that selection is really changed

src/notation/internal/notationinteraction.cpp
```diff
@@ -3504,3 +3505,35 @@ void NotationInteraction::toggleUnderline()
...
+        interactive()->error(trc("notation", "The style file could not be loaded."),
+                             f.errorString().toStdString(), { IInteractive::Button::Ok },
+                             IInteractive::Button::Ok, IInteractive::Option::WithIcon);
+        return;
+    }
+
+    startEdit();
+    if (!score()->loadStyle(stylePath.toQString()) && interactive()->warning(
+            trc("notation",
+                "Since this style file is from a different version of MuseScore, your score is not guaranteed to display correctly."),
```
RomanPudashkin, line 3522: It looks like we need to return some error code from loadStyle (not just bool). We can't guarantee that the "old version" is the reason for the failure

src/engraving/style/style.cpp
```diff
@@ -284,6 +284,21 @@ bool MStyle::read(QIODevice* device, bool ign)
...
 }
 
+bool MStyle::isValid(QIODevice* device)
+{
+    XmlReader e(device);
+    while (e.error() == XmlReader::Error::NoError && e.readNextStartElement()) {
+        if (e.name() == "museScore") {
+            while (e.readNextStartElement()) {
+                if (e.name() == "Style") {
+                    return true;
```
RomanPudashkin, line 294: I'm afraid that this method doesn't guarantee that a style is really valid. A style may contain the "Style" tag, but be corrupted elsewhere

src/notation/view/widgets/editstyle.cpp
RomanPudashkin: I think we need to create a separate interface for user interaction with styles. For example, like this: 

Screenshot 2021-08-18 at 11 08 05

RomanPudashkin: After that, we need to move all code from the screenshot to StyleScenario. Because NotationInteraction isn't a proper place for it

Screenshot 2021-08-18 at 09 59 17

src/notation/internal/notationinteraction.cpp
wizofaus: That's why I added the isValid check. MU3 didn't even have that!  Agree that returning "false" to mean version mismatch isn't ideal but  I'm not sure it's worth investing too much time improving that particular code right now.

src/notation/inotationinteraction.h
wizofaus: True, assuming the properties panel is listening to that! (obviously it should be)

src/engraving/style/style.cpp
wizofaus: It's enough to be confident we can actually extract useful info from it. And it's a significant improvement on MU3 which doesn't even tell you if the file isn't XML at all. Obviously there's always room for improvement but there are far more serious issues to address with very limited time/ resources.  And no spec other than "do what MU3 does".

src/notation/view/widgets/editstyle.cpp
wizofaus: I'm happy to move it to NotationStyle, though NotationInteraction seemed logical enough based on what else it does.

src/notation/inotationinteraction.h
wizofaus: Actually it wasn't, but have fixed.

src/notation/view/widgets/editstyle.cpp
wizofaus: Move to NotationStyle, seemed to make sense.

src/notation/inotationinteraction.h
wizofaus: Done

src/notation/view/widgets/editstyle.cpp
RomanPudashkin: I believe, NotationStyle isn't a proper place for the interaction logic (such as asking questions, showing errors, etc). We need to create a separate class for this (I suggested IStyleScenario, but you can use a different name for it). After that, NotationStyle::loadStyle will look like:

```
loadStyle(path) {
    m_undoStack->prepareChanges();
    bool ok = score()->loadStyle(path);

    if (ok) {
        m_undoStack->commitChanges();
    } else {
        m_undoStack->rollbackChanged();
    }

     return ok;
} 
```

And StyleScenarion::loadStyle will look like:

```
loadStyle(notation, path) {
    bool ok = notation->style()->loadStyle(path)
    if (!ok) {
        interactive()->error(...)
    }
...
} 
```

wizofaus: That's why I originally put it in NotationInteraction!
I'm not comfortable introducing a whole new class just for these two commands, and "Scenario" seems like the wrong word anyway. Feel free to hand it over to one of the core team if you'd prefer a different implementation, I only picked this because it seemed pretty unlikely to impact anyone else.

src/notation/internal/notationstyle.cpp
```diff
@@ -89,3 +89,53 @@ Notification NotationStyle::styleChanged() const
 {
     return m_styleChanged;
 }
+
+mu::io::path NotationStyle::selectStyleFile(bool forLoad)
```
igorkorsukov, line 93: This is not a good place for this code.
We adhere to the MVVM architectural pattern
https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel
User interaction must be at the ViewModel level
This place is Model, there should be no user interaction, ideally. Otherwise, this code (loading, saving styles) cannot be used in other cases, for example, during auto testing, or on the backend, or anywhere else.

src/notation/internal/notationstyle.cpp
```diff
@@ -89,3 +89,53 @@ Notification NotationStyle::styleChanged() const
...
+    m_undoStack->prepareChanges();
+    if (!m_getScore->score()->loadStyle(stylePath.toQString()) && interactive()->warning(
+            trc("notation",
+                "Since this style file is from a different version of MuseScore, your score is not guaranteed to display correctly."),
+            trc("notation", "Click OK to load anyway."), { IInteractive::Button::Ok, IInteractive::Button::Cancel },
+            IInteractive::Button::Ok, IInteractive::Option::WithIcon).standardButton()
+        == IInteractive::Button::Ok) {
+        m_getScore->score()->loadStyle(stylePath.toQString(), true);
+    }
+    m_undoStack->commitChanges();
```
igorkorsukov, line 126: This is not a good place for this code.
We adhere to the MVVM architectural pattern
https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel
User interaction must be at the ViewModel level
This place is Model, there should be no user interaction, ideally. Otherwise, this code (loading, saving styles) cannot be used in other cases, for example, during auto testing, or on the backend, or anywhere else.

igorkorsukov: ViewModel classes are in `modulename/view`

wizofaus: It was Roman's suggestion but I agree it's not the best place to put it, I had it in NotationActionController originally. Thankfully these are about the only two actions I've implemented recently (out of more than 60!) that do require user interaction with dialogs etc. But we need a discussion about how this is going to work with the plugin engine etc.

src/notation/view/widgets/editstyle.cpp
igorkorsukov: This is not a good place for this code.
We adhere to the MVVM architectural pattern
https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel
User interaction must be at the ViewModel level
This place is Model, there should be no user interaction, ideally. Otherwise, this code (loading, saving styles) cannot be used in other cases, for example, during auto testing, or on the backend, or anywhere else.

igorkorsukov: ViewModel classes are in modulename/view

wizofaus: Which is where NotationViewController is, and where I originally had the code to show the dialogs.  I then put the code to do the actual style load/save in NotationInteraction, but Roman didn't approve.  That part's mostly just plumbing anyway, the code that actually does the real work is all inside the engraving library anyway.

RomanPudashkin: > Which is where NotationViewController is, and where I originally had the code to show the dialogs. I then put the code to do the actual style load/save in NotationInteraction, but Roman didn't approve. That part's mostly just plumbing anyway, the code that actually does the real work is all inside the engraving library anyway.

Originally, your code for showing dialogs was in NotationInteraction (see my screenshot above) :) NotationInteraction isn't part of view models. So I suggested, that you can create a separate class and reuse it in NotationActionsController and in the EditStyle dialog

wizofaus: That's why I don't like squashing commits, I'm sure there was an earlier version I had that put showing the dialogs into NotationActionController but there's no trace of it now.

wizofaus: Actually you can see how it was originally here: https://github.com/musescore/MuseScore/pull/8868/commits/6f72b43180bd5c75a1909bfb28d7e2328b8358a4

I didn't really like that either (putting the static helper function in EditStyle), but it did at least adhere to the MVVM pattern.

Jojo-Schmitz: Because you didn't swuash nor rebase yet...

wizofaus: Rare occasion of my common sense kicking in then :-P

RomanPudashkin:
Please, rebase your PR. I want to merge it

src/notation/internal/notationinteraction.h
```diff
@@ -32,6 +32,7 @@
 #include "inotationconfiguration.h"
 #include "inotationundostack.h"
 #include "iselectinstrumentscenario.h"
+#include "iinteractive.h"
```
wizofaus, line 35: reverting this

Jojo-Schmitz:
that merge commit doesn't belong into this PR

wizofaus:
> that merge commit doesn't belong into this PR (and it didn't rebase either)

I tried rebasing and it showed me 100s of commits to rebase, had this problem before...no idea what's happened

Jojo-Schmitz:
So far `git pull --rebase upstream master` never failed on me

wizofaus:
> So far `git pull --rebase upstream master` never failed on me

well i've completed screwed it now - https://github.com/musescore/MuseScore/pull/8868/commits/019d77c1dbe319f2f10978a49713063c1a5b1b22 is showing the right versions of several of the files I need, checking out that specific commit doesn't give me the expected versions.
Did I mention that I'm not a fan of squashing commits...

wizofaus:
Think it's OK now. I'd tried to rebase in a separate repo clone to avoid stuffing up my regular development clone, but it seems that's what stuffed git up.  Did the exact same steps on my regular clone and everything worked fine.

RomanPudashkin (approved):

PR #8870, opened 2021-08-18
[MU4] fix #8694 grip dragging not working

### LINKED ISSUE #8694: Lines not resizing correctly
Line items from the palette can't be lengthened or shortened in the score/project editor.

This seems to be affecting ALL line objects (e.g. slurs, hairpins, crescendo lines, pedal lines, etc.... albeit with **some** inconsistency, as you will see in this demo video):
 
https://user-images.githubusercontent.com/86290556/126642964-de6b6bb1-79ae-4bd0-8aee-fabe50db18bf.mp4

**Expected behaviour**
User should be able to lengthen/shorten all line objects by adjusting left and right handles. 
Line objects should also snap to notes.
See for example: 
https://user-images.githubusercontent.com/86290556/126642985-aa6e2405-2620-4c8c-b741-184b400a2197.mov

Apple MacBook Air M1 MacOS 11.4
Reproduced in Windows 10 ARM (via Parallels)

wizofaus:
Resolves: https://github.com/musescore/issues/8694

Grip mouse operations not working

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -2094,56 +2094,52 @@ bool NotationInteraction::isGripEditStarted() const
     return m_gripEditData.element && m_gripEditData.curGrip != Ms::Grip::NO_GRIP;
 }
 
-bool NotationInteraction::isHitGrip(const PointF& pos) const
+static int findGrip(const QVector<mu::RectF>& grips, Ms::Page* page, const mu::PointF& canvasPos)
 {
-    if (!selection()->element() || m_gripEditData.grip.empty()) {
-        return false;
+    if (grips.empty() || page == nullptr) {
+        return -1;
     }
-
-    qreal align = m_gripEditData.grip[0].width() / 2;
-
-    for (int i = 0; i < m_gripEditData.grips; ++i) {
-        if (m_gripEditData.grip[i].adjusted(-align, -align, align, align).contains(pos)) {
-            return true;
+    mu::PointF pos = canvasPos - page->pos();
+    qreal align = grips[0].width() / 2;
+    for (int i = 0; i < grips.size(); ++i) {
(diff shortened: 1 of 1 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -2976,15 +2972,10 @@ void NotationInteraction::updateGripEdit(const std::vector<Element*>& elements)
...
         return;
     }
-
     m_gripEditData.grips = element->gripsCount();
     m_gripEditData.curGrip = Ms::Grip::NO_GRIP;
     m_gripEditData.element = element;
-    m_gripEditData.grip.resize(m_gripEditData.grips);
-
     m_gripEditData.element->startEdit(m_gripEditData);
-    m_gripEditData.element->updateGrips(m_gripEditData);
```
wizofaus, line 2986: this was producing meaningless grip rectangles that weren't adjusted for current world transform at all

src/notation/internal/notationinteraction.cpp
```diff
@@ -2094,56 +2094,52 @@ bool NotationInteraction::isGripEditStarted() const
...
+            return i;
         }
     }
+    return -1;
+}
 
-    return false;
+bool NotationInteraction::isHitGrip(const PointF& pos) const
+{
+    return selection()->element() && findGrip(m_gripEditData.grip, point2page(pos), pos) != -1;
```
RomanPudashkin, line 2114: You don't use selection()->element() here, so you can remove it from here

wizofaus: `isHitGrip` should return `false` if  `selection()->element()` is `nullptr`.
In principle `m_gripEditData.grip` should always be an empty array in that case, but the check was there already.

RomanPudashkin (approved):

RobFog:
Link above gives a 404. This should be the issue that was fixed: #8694.

vshalkevich:
Works fine

Issue #8270, opened 2021-06-07
[MU4 Issue] Transparent app border on Windows


shoogle:
A transparent frame is visible around the app window and buttons are visually offset so you have to click above and to the left of each button to activate it.

Occurs on Windows in the nightly builds as well as in self-compiled builds.

https://user-images.githubusercontent.com/11011881/129974508-7a22e869-fa00-43e2-9b18-3ca787cfe2bf.mp4

**Desktop:**
 - OS: Windows 10, when compiled outside Qt Creator (e.g. Nightly builds, or with the [build.cmake](https://github.com/musescore/MuseScore/wiki/Compile-on-the-command-line#Using-buildcmake-recommended) script)
 
**System specs:** (it's a MacBook Pro running Windows in a dual-boot setup)

![image](https://user-images.githubusercontent.com/11011881/122230147-49587800-ceb1-11eb-9341-a1cc659ea73f.png)

shoogle:
Report from a community member:
> The transparent frame only appears when I run the Musescore NightlyBuilds via the processor's own Intel graphics, but not via the NVidia graphics card...

wizofaus:
Any progress on this? It makes working with MU4 just to test my own changes seriously painful...and I can't figure out what's causing it either, something buried inside Qt itself it seems.

cbjeukendrup:
@Eism is working on Windows-titlebar related things (see https://github.com/musescore/MuseScore/issues/8221#issuecomment-860063040) so I guess he will fix this one too.

Eism:
@cbjeukendrup
I also hope so :)

RomanPudashkin:
@shoogle @wizofaus 

I can't reproduce this issue on my laptop with windows 10 (it has only integrated Intel UHD Graphics 620). It would be great to add to the description of the issue more details about your hardware

wizofaus:
How did you build it? My understanding is that it doesn't happen when built with QtCreator. @RomanPudashkin 
This is just an ASUS laptop with Intel HD Graphics 530 and NVIDIA GForce GTX as display adapters. I don't know how to run an app using a particular adapter.

RomanPudashkin:
> How did you build it? My understanding is that it doesn't happen when built with QtCreator

Yes, I have built it using QtCreator

wizofaus:
If you build just with MSVC or use the nightly Windows build does it happen?

RomanPudashkin:
My QtCreator already uses MSVC

[...]

PR #8872, opened 2021-08-18
[MU4] Added opengl buglist

Eism:
Resolves: #8270

### DIFF
--- a/src/appshell/appshell.cpp
+++ b/src/appshell/appshell.cpp
@@ -70,6 +70,12 @@ int AppShell::run(int argc, char** argv)
         appName  = "MuseScore4";
     }
 
+#ifdef Q_OS_WIN
+    if (!qEnvironmentVariableIsSet("QT_OPENGL_BUGLIST")) {
+        qputenv("QT_OPENGL_BUGLIST", ":/resources/win_opengl_buglist.json");
+    }
+#endif
+
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
     QApplication app(argc, argv);
     QCoreApplication::setApplicationName(appName);
--- a/src/appshell/appshell.cpp
+++ b/src/appshell/appshell.cpp
@@ -71,6 +71,10 @@ int AppShell::run(int argc, char** argv)
     }
 
 #ifdef Q_OS_WIN
+    // NOTE: There are some problems with rendering the application window on some integrated graphics processors
+    //       see https://github.com/musescore/MuseScore/issues/8270
+    QCoreApplication::setAttribute(Qt::AA_UseOpenGLES);
+
     if (!qEnvironmentVariableIsSet("QT_OPENGL_BUGLIST")) {
         qputenv("QT_OPENGL_BUGLIST", ":/resources/win_opengl_buglist.json");
     }
--- a/src/appshell/appshell.qrc
+++ b/src/appshell/appshell.qrc
@@ -76,5 +76,6 @@
         <file>qml/platform/linux/Main.qml</file>
         <file>qml/platform/mac/Main.qml</file>
         <file>qml/platform/win/Main.qml</file>
+        <file>resources/win_opengl_buglist.json</file>
     </qresource>
 </RCC>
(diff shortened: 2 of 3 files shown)


wizofaus:
Unfortunately this makes no difference to the transparent frame or mouse offset issue for me.

Eism:
sad, this hack helped me
I'll think about how to fix the bug

As a temporary solution, you can write in appshel.cpp
QCoreApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);

shoogle:
The PR didn't work for me either, I'm afraid.

> As a temporary solution, you can write in appshel.cpp
QCoreApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);

This works but the app runs slowly when switching tabs or opening dialogs, etc. so not ideal obviously.

wizofaus:
The change to us AA_UseOpenGLES seems to work fine, well done!  No noticeable performance issues either.

wizofaus:
Having said that, it's still different to the borders in other applications, e.g. Visual Studio

image

Whereas MuseScore now looks like

image

So missing that single-pixel darker line.

src/appshell/appshell.cpp
```diff
@@ -70,6 +70,16 @@ int AppShell::run(int argc, char** argv)
         appName  = "MuseScore4";
     }
 
+#ifdef Q_OS_WIN
+    // NOTE: There are some problems with rendering the application window on some integrated graphics processors
+    //       see https://github.com/musescore/MuseScore/issues/8270
+    QCoreApplication::setAttribute(Qt::AA_UseOpenGLES);
+
+    if (!qEnvironmentVariableIsSet("QT_OPENGL_BUGLIST")) {
+        qputenv("QT_OPENGL_BUGLIST", ":/resources/win_opengl_buglist.json");
```
wizofaus, line 79: BTW it works fine on my system without this, just setting AA_UseOpenGLES is sufficient.

Jojo-Schmitz:
Isn't this basically a resurrection of https://github.com/musescore/MuseScore/commit/ed4949c72d3f1d8f779561b29186ed59c10a766f ?
Begs the question why/how it got lost for master?

src/appshell/appshell.cpp
Eism: yes, but there may be problems with these processors and it is very difficult to understand what the problem is.
insure against them

Jojo-Schmitz: > BTW it works fine on my system without this, just setting AA_UseOpenGLES is sufficient.

See ed4949c as to why it was needed in 3.x, there's no reason to assume it isn't needed in master, as far as I can tell

wizofaus:
It's only the AA_UseOpenGLES that makes a difference for me, the other changes don't seem to be needed (though may well solve other problems).

Jojo-Schmitz:
They sure did in 3.x, and I see no reason to assume them to not be needed for master too?

wizofaus:
3.x used Qt Widgets though and that did a lot of things differently. But honestly I don't know.

Jojo-Schmitz:
It used QML (and therefor QtQuick) too (like for the palettes), just not as much.

wizofaus:
Apparently AA_UseOpenGLES does nothing with Qt 6, something to be aware of if there's any serious consideration of switching.

shoogle (approved):
Works for me with ` QCoreApplication::setAttribute(Qt::AA_UseOpenGLES);`. We might as well keep the `BUGLIST` stuff if that helped on @Eism's machine.

> it's still different to the borders in other applications, e.g. Visual Studio
> So missing that single-pixel darker line, and missing the app icon from where you can bring up the window menu

The app border on my machine is the same as File Explorer. The window menu appears when I right-click on the title bar so I don't think the app icon is a big loss.

image

I have dual monitors. The app menu is incorrectly scaled on the external non-HDPI monitor but this was already the case before this PR. It is always scaled correctly on the internal HDPI monitor, and also on the external monitor if I disable the internal one.

RomanPudashkin (approved):

vshalkevich:
Seems fine even with minor bugs

PR #8879, opened 2021-08-19
[MU4] fix #8159 (partial), implement move-up/move-down/up-chord/down-chord/top-chord/bottom-chord

### LINKED ISSUE #8159: [MU4 Issue] Close command, and many shortcut keys don't work
**Describe the bug**
Key functionality not working:
Close command (after opening score)
Esc key (to deselect etc., and in general to close dialogs etc.)
Cursor keys (including, e.g., Shift+Ctrl+up/down, to move notes to other staves for cross-staff beaming)
Del & backspace key 

**To Reproduce**
1. Build/run MU4 in VS 2019, open a score
2. Click on File|Close
3. Select a note, attempt to use various keys as above
4. Note nothing happens

**Expected behavior**
Functions should work as per MU3

**Screenshots**

 - OS: Windows 10
 - 
**Additional context**

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8159 (partially)

Various score manipulation/navigation commands not implemented

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -3142,7 +3142,11 @@ void Score::cmdDeleteSelection()
     deselectAll();
     // make new selection if appropriate
     if (noteEntryMode()) {
-        cr = _is.cr();
+        if (cr) {
+            _is.setSegment(cr->segment());
+        } else {
+            cr = _is.cr();
+        }
     }
     if (cr) {
         if (cr->isChord()) {
--- a/src/framework/shortcuts/internal/shortcutsregister.cpp
+++ b/src/framework/shortcuts/internal/shortcutsregister.cpp
@@ -69,7 +69,7 @@ void ShortcutsRegister::load()
         if (!ok) {
             m_shortcuts = m_defaultShortcuts;
         } else {
-            mergeSortcuts(m_shortcuts, m_defaultShortcuts);
+            mergeShortcuts(m_shortcuts, m_defaultShortcuts);
         }
         ok = true;
     }
@@ -80,7 +80,7 @@ void ShortcutsRegister::load()
     }
 }
 
-void ShortcutsRegister::mergeSortcuts(ShortcutList& shortcuts, const ShortcutList& defaultShortcuts) const
+void ShortcutsRegister::mergeShortcuts(ShortcutList& shortcuts, const ShortcutList& defaultShortcuts) const
 {
     ShortcutList needadd;
     for (const Shortcut& sh : defaultShortcuts) {
@@ -163,7 +163,15 @@ bool ShortcutsRegister::readFromFile(ShortcutList& shortcuts, const io::path& pa
 
         Shortcut shortcut = readShortcut(reader);
         if (shortcut.isValid()) {
-            shortcuts.push_back(shortcut);
+            if (shortcut.sequence.empty()) {
+                shortcuts.push_back(shortcut);
+            } else {
+                auto seqList = QString::fromStdString(shortcut.sequence).split("\n", Qt::SkipEmptyParts);
+                for (QString seq : seqList) {
+                    shortcut.sequence = seq.toStdString();
+                    shortcuts.push_back(shortcut);
+                }
+            }
         }
     }
 
@@ -186,7 +194,7 @@ Shortcut ShortcutsRegister::readShortcut(framework::XmlReader& reader) const
         } else if (tag == STANDARD_KEY_TAG) {
             shortcut.standardKey = QKeySequence::StandardKey(reader.readInt());
         } else if (tag == SEQUENCE_TAG) {
-            shortcut.sequence = reader.readString();
+            shortcut.sequence += reader.readString() + "\n";
         } else {
             reader.skipCurrentElement();
         }
@@ -210,7 +218,7 @@ mu::Ret ShortcutsRegister::setShortcuts(const ShortcutList& shortcuts)
 
     if (ok) {
         m_shortcuts = shortcuts;
-        mergeSortcuts(m_shortcuts, m_defaultShortcuts);
+        mergeShortcuts(m_shortcuts, m_defaultShortcuts);
         m_shortcutsChanged.notify();
     }
 
--- a/src/notation/inotationinteraction.h
+++ b/src/notation/inotationinteraction.h
@@ -71,6 +71,7 @@ public:
     virtual INotationSelectionPtr selection() const = 0;
     virtual void clearSelection() = 0;
     virtual async::Notification selectionChanged() const = 0;
+    virtual void selectTopOrBottomOfChord(MoveDirection d) = 0;
 
     // SelectionFilter
     virtual bool isSelectionTypeFiltered(SelectionFilterType type) const = 0;
@@ -101,6 +102,7 @@ public:
     virtual void moveSelection(MoveDirection d, MoveSelectionType type) = 0;
     virtual void movePitch(MoveDirection d, PitchMode mode) = 0;  //! NOTE Requires a note to be selected
     virtual void moveText(MoveDirection d, bool quickly) = 0;     //! NOTE Requires a text element to be selected
+    virtual void moveChordRestToStaff(MoveDirection d) = 0;
 
     // Text edit
     virtual bool isTextEditingStarted() const = 0;
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -143,8 +143,8 @@ void NotationActionController::init()
     registerMoveAction("pitch-down");
     registerMoveAction("pitch-up-octave");
     registerMoveAction("pitch-down-octave");
-    registerMoveAction("up-chord");
-    registerMoveAction("down-chord");
+    registerAction("move-up", [this]() { moveChordRestToStaff(MoveDirection::Up); }, &NotationActionController::hasSelection);
+    registerAction("move-down", [this]() { moveChordRestToStaff(MoveDirection::Down); }, &NotationActionController::hasSelection);
 
     registerAction("double-duration", [this]() { increaseDecreaseDuration(-1, /*stepByDots*/ false); });
     registerAction("half-duration", [this]() { increaseDecreaseDuration(1, false); });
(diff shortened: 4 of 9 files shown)


src/notation/internal/notationactioncontroller.cpp
```diff
@@ -143,8 +143,8 @@ void NotationActionController::init()
     registerMoveAction("pitch-down");
     registerMoveAction("pitch-up-octave");
     registerMoveAction("pitch-down-octave");
-    registerMoveAction("up-chord");
-    registerMoveAction("down-chord");
```
wizofaus, line 147: these weren't actually working

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -649,6 +653,24 @@ void NotationActionController::addBracketsToSelection(BracketsType type)
...
+}
+
+void NotationActionController::moveDown()
+{
+    auto interaction = currentNotationInteraction();
+    if (!interaction) {
+        return;
+    }
+    interaction->moveSelectionByStaff(MoveDirection::Down);
+}
```
wizofaus, line 672: probably should combine these two into one function...

src/notation/inotationinteraction.h
```diff
@@ -71,6 +71,7 @@ class INotationInteraction
     virtual INotationSelectionPtr selection() const = 0;
     virtual void clearSelection() = 0;
     virtual async::Notification selectionChanged() const = 0;
+    virtual void selectChordTopOrBottom(MoveDirection d) = 0;
```
wizofaus, line 74: rename to selectTopOrBottomOfChord

src/notation/internal/notationinteraction.cpp
```diff
@@ -1852,6 +1875,34 @@ void NotationInteraction::moveSelection(MoveDirection d, MoveSelectionType type)
...
+            if (dir == MoveDirection::Up) {
+                score()->moveUp(cr);
+            } else if (dir == MoveDirection::Down) {
+                score()->moveDown(cr);
+            }
+        }
+    }
+    apply();
+    notifyAboutNotationChanged();
+    notifyAboutSelectionChanged();
```
wizofaus, line 1903: probably unnecessary to send both...

src/engraving/libmscore/edit.cpp
```diff
@@ -3142,7 +3142,11 @@ void Score::cmdDeleteSelection()
     deselectAll();
     // make new selection if appropriate
     if (noteEntryMode()) {
-        cr = _is.cr();
+        if (cr) {
+            _is.setSegment(cr->segment());
+        } else {
+            cr = _is.cr();
+        }
```
wizofaus, line 3149: This emulates the MU3 "backspace" behaviour where the note entry cursor moves to the note just deleted. It will do this when using either "delete" or "backspace", but unlike MU3 adds a new action to the undo stack (which can then be undone etc.).

src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -80,12 +80,12 @@ void ShortcutsRegister::load()
...
 }
 
-void ShortcutsRegister::mergeSortcuts(ShortcutList& shortcuts, const ShortcutList& defaultShortcuts) const
+void ShortcutsRegister::mergeShortcuts(ShortcutList& shortcuts, const ShortcutList& defaultShortcuts) const
 {
     ShortcutList needadd;
     for (const Shortcut& sh : defaultShortcuts) {
         auto it = std::find_if(shortcuts.begin(), shortcuts.end(), [sh](const Shortcut& i) {
-            return i.action == sh.action;
+            return i.action == sh.action && i.sequence == sh.sequence;
```
wizofaus, line 88: actually i'm not sure what the right behaviour is here - is the intention that you can *replace* shortcuts rather than merge/add them?

src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -186,7 +190,7 @@ Shortcut ShortcutsRegister::readShortcut(framework::XmlReader& reader) const
         } else if (tag == STANDARD_KEY_TAG) {
             shortcut.standardKey = QKeySequence::StandardKey(reader.readInt());
         } else if (tag == SEQUENCE_TAG) {
-            shortcut.sequence = reader.readString();
+            shortcut.sequence += reader.readString() + ";";
```
wizofaus, line 193: quite a lot of commands have multiple alternative sequences defined, including importantly "delete" for which you can use both "Del" and "Backspace". Existing code was only ever taking the last.

src/framework/shortcuts/internal/shortcutsregister.cpp
wizofaus: Reset to factory settings doesn't actually work, you have to delete the local file (for me it was  C:/Users/dylan.nicholson/AppData/Local/MuseScore/MuseScore4Development/shortcuts.xml)
But I think requiring existing MU4 users (remembering it's still pre-alpha!) and restoring the existing logic is the only correct solution, or if you removed any of the default shortcut mappings, they'd just come back again next time you started up MuseScore.

wizofaus:
The fix for actions with multiple shortcut sequences (like "delete") will only work if you haven't customised the shortcuts, or if you delete the file that stored your shortcut customisations, which on windows is %LOCALAPPDATA%\MuseScore\MuseScore4Development\shortcuts.xml
This is because this would have been saved with ONLY the single sequence defined for each action, which will override what gets loaded from the default shortcuts definitions file.

src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -186,7 +194,7 @@ Shortcut ShortcutsRegister::readShortcut(framework::XmlReader& reader) const
         } else if (tag == STANDARD_KEY_TAG) {
             shortcut.standardKey = QKeySequence::StandardKey(reader.readInt());
         } else if (tag == SEQUENCE_TAG) {
-            shortcut.sequence = reader.readString();
+            shortcut.sequence += reader.readString() + "\n";
```
wizofaus, line 197: \n can never possibly be part of a shortcut sequence name so this always safe, even if it a very slight kludge

wizofaus:
There's a problem introduced here now that Ctrl+U now both activates "full screen" *and* underline mode when editing text!
I was aware that might be an issue when I implemented Ctrl+U for toggle underline, but didn't understand at the time why it wasn't working for "toggle full screen".  Not sure how to best handle.

wizofaus:
For those testing, what's been implemented here are the commands:

move-up/move-down:   Move chord/rest to staff above/below (ctrl+shift+up/down by default), NB these only make sense in a multi-staff instrument, e.g. using the grand staff template.

up-chord/down-chord:  Go to higher/lower pitched note in chord (alt+up/down by default), NB despite the name, you can actually jump to the previous or next staff too.

top-chord/bottom-chord: Go to top/bottom note in chord (ctrl+alt+up/down by default).  These commands only work within the current staff (or voice even).

delete:  del or backspace key. In note entry mode this now causes the input cursor to move back to where the deleted note was (which MU3 did for "backspace" but not "delete").  However unlike MU3 in "insert" mode backspace does not remove the time range occupied by the previously inserted note (you can use Ctrl+Z for that).

ANY command with multiple shortcut key sequences, e.g. "full screen", all shortcut key sequences should work.  

As noted, you'll need to delete your existing user shortcuts.xml file if you've previously customized shortcuts in MU4 for this to work.

Tantacrul:
> There's a problem introduced here now that Ctrl+U now both activates "full screen" _and_ underline mode when editing text!
> I was aware that might be an issue when I implemented Ctrl+U for toggle underline, but didn't understand at the time why it wasn't working for "toggle full screen".
> 
> Not sure how to best handle - what I've done is to inject globalContext into ApplicationActionController and have ApplicationActionController::toggleFullScreen query `globalContext()->currentNotation()->interaction()->isTextEditingStarted()` and it works but seems to be adding dependencies that shouldn't ideally be needed.
> 
> Actually the simplest change is to handle Ctrl+U as a special case inside NotationInteraction::editText( ) and swallow the shortcut key and force map it to toggleUnderline, but obviously not nice and really want something that will work sensibly even if the user chooses some other shortcut key for fullscreen that happens to have a different meaning in text editing mode.
> But note that's essentially what MU3 does now (you can remap a new shortcut to toggle-underline there but it doesn't actually work - the text element internally handles Ctrl+U as a special case).
> 
> I'd actually much rather we change ShortcutsController::activate( ) and ActionsDispatcher to use a "hierarchical" mechanism where the focused/active window (or "most specific" action) got first bite and was able to return for each mapped action whether any were handled, and if it not, they would be passed up to the next level. In other words, in the case of Ctrl+U, even though it maps to both "fullscreen" and "toggleUnderline", if the NotationActionController says it can receive "toggleUnderline", then it's dispatched there and no further processing for the shortcut key takes place.
> But currently all ShortcutsController::activate( ) knows is that ctrl+u is mapped to both actions, and "fullscreen" is _first_, and processes that before even checking if "toggleUnderline" is handled.
> Even if it we can just change the order of the shortcuts registry so that "toggleUnderline" is first, and if it's enabled currently, we dispatch that action and don't continue processing other actions, that would make more sense - as it is, I can't think of any case it would actually be useful for a single shortcut key sequence to fire off multiple actions.
> As an experiment I changed it to reverse the order of the actions and to stop looking after finding the first one that's enabled, which does work but it's a bit fragile, and will cause some shortcut keys not to work in certain cases where we leave commands enabled even though they don't do anything useful. Personally I wouldn't be shipping a factory default set of shortcut keys that map multiple actions to the same key sequence, but I guess existing users are used to it.
> 
> @igorkorsukov @Tantacrul

I like the idea of certain shortcuts firing only in specific contexts. I'd suggest right now that we should be a little careful about making any alterations if we think it might introduce new issues. Since we're trying to lock MS4 down, we should generally stick to MS3 behaviour and agree to revisit problematic things later. Really depends on risk.

wizofaus:
I'm trying to achieve MU3 behavior - I.e. Ctrl+U means toggle underline while editing text, fullscreen elsewhere. But mu3's implementation is buggy (doesn't work it you assign a different shortcut to toggle underline) and Mu4 handles shortcuts/ commands very differently to Mu3. "Fullscreen" is considered globally available currently, and what I'm trying to come up with is the best way to ensure that if a more "specific" action for the same shortcut like toggle underline is currently enabled that takes precedence.
Is there any scenario we want one shortcut key to fire multiple simultaneous actions? (Remembering they could happen in any order,  not like a macro/ defined sequence of actions,  which would definitively be nice to have one day).

src/framework/shortcuts/internal/shortcutsregister.cpp
RomanPudashkin: Could you please explain the purpose of this change?

wizofaus: Currently it's ignoring multiple sequences for the same action (e.g. Dell or backspace to mean "delete")

RomanPudashkin (approved):

wizofaus:
Oh...wasn't expecting this to get merged as is, the Ctrl+U issue never got addressed and these changes are part of a new PR already (that does address that issue).

RobFog:
If a pull request is not ready, you can mark it as a draft.

PR #8886, opened 2021-08-20
Score-Page Color Inversion

ArjunBoi:
Resolves: #8909 & adds additional features

This is an additional feature to the newly added High-Contrast themes of MuseScore (as part of my Google Summer of Code project) wherein the user is given the ability to invert the colors of the Score-Page from within the Edit -> Preferences -> Appearance tab.

From the Appearance tab, Toggle High-Contrast mode, and there will be a new "Invert score colors" checkbox under the "UI colors" Category.

Behavior:
* Checking the "Invert score colors" checkbox will set the color of the foreground (or paper color) to black, and the color of the notes/clefs/staff/text/time-signatures etc. to a very slightly dulled white (255,255,255,200).
* Un-checking the score inversion checkbox restores previous settings i.e. foreground color, note color etc.
* While Score Inversion is active, the ability to modify the page color is taken away from the user so as to maintain proper visibility and avoid very mismatched score/page color combinations. The Page Color settings stop receiving mouse and keyboard events and are rendered at 0.6 opacity, indicating that they are indeed disabled.
* While Score Inversion is active, If the user was to toggle between High-Contrast/General themes, the score-color/page-color would respond accordingly, wherein switching to a general theme will restore previously used "normal" score-color/paper-color values, since the inversion feature is only reserved for High-Contrast mode.

### DIFF
--- a/src/appshell/qml/Preferences/internal/UiColorsSection.qml
+++ b/src/appshell/qml/Preferences/internal/UiColorsSection.qml
@@ -34,7 +34,6 @@ BaseSection {
 
     property alias scoreInversionChecked: scoreInversionEnable.checked
 
-    signal changePageColorRequested()
     signal scoreInversionRequested(var newValue)
     signal colorChangeRequested(var newColor, var propertyType)
 
@@ -46,7 +45,6 @@ BaseSection {
         onClicked: {
             checked = !checked
             root.scoreInversionRequested(checked)
-            root.changePageColorRequested()
         }
     }
 
--- a/src/appshell/qml/Preferences/internal/UiColorsSection.qml
+++ b/src/appshell/qml/Preferences/internal/UiColorsSection.qml
@@ -32,21 +32,7 @@ BaseSection {
     title: qsTrc("appshell", "UI colors")
     navigation.direction: NavigationPanel.Both
 
-    property bool scoreInversionChecked: false
-
-    signal scoreInversionRequested(var newValue)
     signal colorChangeRequested(var newColor, var propertyType)
-
-    CheckBox {
-        id: scoreInversionEnable
-
-        checked: root.scoreInversionChecked
-        text: qsTrc("appshell", "Invert score colors")
-
-        onClicked: {
-            root.scoreInversionRequested(!checked)
-        }
-    }
 
     GridLayout {
         id: grid
--- a/src/appshell/qml/Preferences/internal/UiColorsSection.qml
+++ b/src/appshell/qml/Preferences/internal/UiColorsSection.qml
@@ -34,6 +34,7 @@ BaseSection {
 
     property alias scoreInversionChecked: scoreInversionEnable.checked
 
+    signal changePageColorRequested()
     signal scoreInversionRequested(var newValue)
     signal colorChangeRequested(var newColor, var propertyType)
 
@@ -44,7 +45,8 @@ BaseSection {
 
         onClicked: {
             checked = !checked
-            scoreInversionRequested(checked)
+            root.scoreInversionRequested(checked)
+            root.changePageColorRequested()
         }
     }
 
--- a/src/appshell/qml/Preferences/internal/UiColorsSection.qml
+++ b/src/appshell/qml/Preferences/internal/UiColorsSection.qml
@@ -32,7 +32,7 @@ BaseSection {
     title: qsTrc("appshell", "UI colors")
     navigation.direction: NavigationPanel.Both
 
-    property alias scoreInversionChecked: scoreInversionEnable.checked
+    property bool scoreInversionChecked: false
 
     signal scoreInversionRequested(var newValue)
     signal colorChangeRequested(var newColor, var propertyType)
@@ -40,11 +40,11 @@ BaseSection {
     CheckBox {
         id: scoreInversionEnable
 
+        checked: root.scoreInversionChecked
         text: qsTrc("appshell", "Invert score colors")
 
         onClicked: {
-            checked = !checked
-            root.scoreInversionRequested(checked)
+            root.scoreInversionRequested(!checked)
         }
     }
 
--- a/src/appshell/qml/Preferences/internal/UiColorsSection.qml
+++ b/src/appshell/qml/Preferences/internal/UiColorsSection.qml
@@ -32,7 +32,21 @@ BaseSection {
     title: qsTrc("appshell", "UI colors")
     navigation.direction: NavigationPanel.Both
 
+    property alias scoreInversionChecked: scoreInversionEnable.checked
+
+    signal scoreInversionRequested(var newValue)
     signal colorChangeRequested(var newColor, var propertyType)
+
+    CheckBox {
+        id: scoreInversionEnable
+
+        text: qsTrc("appshell", "Invert score colors")
+
+        onClicked: {
+            checked = !checked
+            scoreInversionRequested(checked)
+        }
+    }
 
     GridLayout {
         id: grid
--- a/src/appshell/view/preferences/appearancepreferencesmodel.h
+++ b/src/appshell/view/preferences/appearancepreferencesmodel.h
@@ -26,7 +26,7 @@
 
 #include "modularity/ioc.h"
 #include "ui/iuiconfiguration.h"
-#include "notation/view/notationpaintview.h"    //testing
+//#include "notation/view/notationpaintview.h"    //testing
 #include "notation/inotationconfiguration.h"
 #include "async/asyncable.h"
 
--- a/src/appshell/view/preferences/appearancepreferencesmodel.h
+++ b/src/appshell/view/preferences/appearancepreferencesmodel.h
@@ -36,6 +36,7 @@ class AppearancePreferencesModel : public QObject, public async::Asyncable
 
     INJECT(appshell, ui::IUiConfiguration, uiConfiguration)
     INJECT(appshell, notation::INotationConfiguration, notationConfiguration)
+    INJECT(appshell, engraving::IEngravingConfiguration, engravingConfiguration)
 
     Q_PROPERTY(bool highContrastEnabled READ highContrastEnabled WRITE setHighContrastEnabled NOTIFY highContrastEnabledChanged)
     Q_PROPERTY(QVariantList generalThemes READ generalThemes NOTIFY themesChanged)
--- a/src/appshell/view/preferences/appearancepreferencesmodel.h
+++ b/src/appshell/view/preferences/appearancepreferencesmodel.h
@@ -96,7 +96,6 @@ public:
     bool scoreInversionEnabled() const;
 
     Q_INVOKABLE void resetThemeToDefault();
-    Q_INVOKABLE bool enableHighContrastChecked();
     Q_INVOKABLE void setNewColor(const QColor& newColor, ColorType colorType);
     Q_INVOKABLE QStringList allFonts() const;
     Q_INVOKABLE QString wallpaperPathFilter() const;
--- a/src/appshell/view/preferences/appearancepreferencesmodel.h
+++ b/src/appshell/view/preferences/appearancepreferencesmodel.h
@@ -58,7 +58,7 @@ class AppearancePreferencesModel : public QObject, public async::Asyncable
     Q_PROPERTY(
         QString foregroundWallpaperPath READ foregroundWallpaperPath WRITE setForegroundWallpaperPath NOTIFY foregroundWallpaperPathChanged)
 
-    Q_PROPERTY(bool invertScoreColor READ invertScoreColor WRITE setInvertScoreColor NOTIFY invertScoreColorChanged)
+    Q_PROPERTY(bool scoreInversionEnabled READ scoreInversionEnabled WRITE setScoreInversionEnabled NOTIFY invertScoreColorChanged)
 
 public:
     explicit AppearancePreferencesModel(QObject* parent = nullptr);
@@ -93,7 +93,7 @@ public:
     QColor foregroundColor() const;
     QString foregroundWallpaperPath() const;
 
-    bool invertScoreColor() const;
+    bool scoreInversionEnabled() const;
 
     Q_INVOKABLE void resetThemeToDefault();
     Q_INVOKABLE bool enableHighContrastChecked();
@@ -114,7 +114,7 @@ public slots:
     void setForegroundUseColor(bool value);
     void setForegroundColor(const QColor& color);
     void setForegroundWallpaperPath(const QString& path);
-    void setInvertScoreColor(bool value);
+    void setScoreInversionEnabled(bool value);
 
 signals:
     void highContrastEnabledChanged();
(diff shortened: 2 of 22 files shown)


src/engraving/libmscore/element.cpp
```diff
@@ -1368,22 +1377,22 @@ void paintElements(mu::draw::Painter& painter, const QList<Element*>& elements)
...
 
-            return e1->track() > e2->track();
-        }
+                return e1->track() > e2->track();
+            }
 
-        return e1->z() < e2->z();
-    });
+            return e1->z() < e2->z();
+        });
```
Jojo-Schmitz, line 1395: these were all perfectly indented before

src/engraving/libmscore/stafftype.h
```diff
@@ -392,10 +392,10 @@ class StaffType
 
     // properties setters (setting some props invalidates metrics)
     void  setDurationFontName(const QString&);
-    void  setDurationFontSize(qreal);
+    void setDurationFontSize(qreal);
     void  setDurationFontUserY(qreal val) { _durationFontUserY = val; }
     void  setFretFontName(const QString&);
-    void  setFretFontSize(qreal);
+    void setFretFontSize(qreal);
```
Jojo-Schmitz, line 398: why onle those and not any of the others in that block?

src/engraving/libmscore/element.cpp
ArjunBoi: That is a problem with my Uncrustify file. (possibly caused by having an older version?)
I encounter the same problem in `uitypes.h` where i have to disable Uncrustify and manually reset it to whatever the codestyle approves. I'll do it here too :)

src/engraving/libmscore/element.cpp
```diff
@@ -127,7 +127,7 @@
 #endif
 
 #include "log.h"
-#define LOG_PROP() if (0) LOGD()
+#define LOG_PROP() if (0) LOGD ()
```
Jojo-Schmitz, line 130: code style issue

src/engraving/libmscore/element.cpp
```diff
@@ -483,7 +483,7 @@ Fraction Element::beat() const
     tsm->tickValues(tick().ticks(), &bar, &beat, &ticks);
     int ticksB = ticks_beat(tsm->timesig(tick().ticks()).timesig().denominator());
 
-    Fraction complexFraction((++beat * ticksB) + ticks, ticksB);
+    Fraction complexFraction((++beat* ticksB) + ticks, ticksB);
```
Jojo-Schmitz, line 486: code style issue

src/engraving/libmscore/stafftype.h
Jojo-Schmitz: Should be either all or none. I'M for none

src/engraving/libmscore/element.cpp
ArjunBoi: fixed

src/engraving/libmscore/element.cpp
ArjunBoi: fixed

src/engraving/libmscore/element.cpp
ArjunBoi: This function has been removed in the latest master i believe

src/engraving/libmscore/stafftype.h
ArjunBoi: fixed

src/engraving/libmscore/element.cpp
Jojo-Schmitz: A rebase would shoe, I guess 
It is needed anyhow

Jojo-Schmitz: and indeed element.cpp is one of the conflicting files

src/notation/internal/notationconfiguration.cpp
```diff
@@ -266,7 +273,11 @@ async::Notification NotationConfiguration::backgroundChanged() const
 
 QColor NotationConfiguration::foregroundColor() const
 {
-    return settings()->value(FOREGROUND_COLOR).toQColor();
+    if (scoreInversionEnabled() && uiConfiguration()->isCurrentThemeHighContrast()) {
+        return QColor("#000000");
```
Jojo-Schmitz, line 277: Wouldn't that just be `Color::black`?

src/notation/internal/notationconfiguration.cpp
```diff
@@ -299,19 +310,42 @@ async::Notification NotationConfiguration::foregroundChanged() const
...
 io::path NotationConfiguration::wallpapersDefaultDirPath() const
 {
     return globalConfiguration()->appDataPath() + "/wallpapers";
 }
 
 QColor NotationConfiguration::borderColor() const
 {
-    return QColor(0, 0, 0, 102);
+    if (uiConfiguration()->isCurrentThemeHighContrast()) {
+        return QColor(255, 255, 255);
```
Jojo-Schmitz, line 336: Wouldn't that just be `Color::white`?

Jojo-Schmitz:
Quite a few unit tests crash.

src/notation/internal/notationconfiguration.cpp
cbjeukendrup: This is QColor though, rather than mu::draw::Color.

Jojo-Schmitz: OK, then `QColor::black`

ArjunBoi: There is no such thing as `QColor::black`
There is only a `QColor::black()`, which returns the black part of the color from the Cmyk coloring system.

src/notation/internal/notationconfiguration.cpp
ArjunBoi: Used `QColorConstants::White` instead.

ArjunBoi:
How would I go about fixing the unit tests? @Jojo-Schmitz

Jojo-Schmitz:
Running them manually in a debugger

src/appshell/qml/Preferences/internal/UiColorsSection.qml
```diff
@@ -32,8 +32,22 @@ BaseSection {
...
+    signal scoreInversionRequested(var newValue)
     signal colorChangeRequested(var newColor, var propertyType)
 
+    CheckBox {
+        id: scoreInversionEnable
+
+        text: qsTrc("appshell", "Invert score colors")
+
+        onClicked: {
+            checked = !checked
```
Eism, line 46: after this line, your alias doesn't work anymore because you rewrite to a fixed value(just bool, no binding)
property alias scoreInversionChecked: scoreInversionEnable.checked

qml should request about value changed and no more. The model will change the local value and emit a signal about data changed, the view will update the state by alias. But the model can also cancel data changing, then the data of model and of view are different because the view already changed value(checked = !checked).

You can see how I fixed it for highContrastEnabled

src/appshell/view/preferences/appearancepreferencesmodel.h
```diff
@@ -58,6 +58,8 @@ class AppearancePreferencesModel : public QObject, public async::Asyncable
     Q_PROPERTY(
         QString foregroundWallpaperPath READ foregroundWallpaperPath WRITE setForegroundWallpaperPath NOTIFY foregroundWallpaperPathChanged)
 
+    Q_PROPERTY(bool invertScoreColor READ invertScoreColor WRITE setInvertScoreColor NOTIFY invertScoreColorChanged)
```
Eism, line 61: it is better to rename to scoreInversionEnabled

src/notation/internal/notationconfiguration.cpp
Jojo-Schmitz: IMHO yes.

src/notation/internal/notationconfiguration.cpp
Jojo-Schmitz: Yup

src/engraving/libmscore/engravingitem.cpp
```diff
@@ -546,6 +551,11 @@ mu::draw::Color EngravingItem::curColor(bool isVisible, mu::draw::Color normalCo
     if (!isVisible) {
         return engravingConfiguration()->invisibleColor();
     }
+    if (engravingConfiguration()->isCurrentThemeHighContrast() && engravingConfiguration()->scoreInversionEnabled()
+        && !score()->isPaletteScore()) {
+        return mu::draw::Color(255, 255, 255, 160); //slightly dulled white for less strain on the eyes
```
shoogle, line 556: It would be better to return a full opacity off-white colour:

```c++
return mu::draw::Color(160, 160, 160, 255);
```

Alpha transparency affects performance because the program has to calculate the colour of pixels behind each object instead of just ignoring those pixels. Using a full opacity colour also prevents the colour changing when you have multiple semi-transparent objects on top of each other.

src/framework/ui/view/uitheme.cpp
```diff
@@ -524,7 +524,7 @@ void UiTheme::drawPrimitive(QStyle::PrimitiveElement element, const QStyleOption
 
     // GroupBox
     case QStyle::PE_FrameGroupBox: {
-        drawRoundedRect(painter, option->rect, 3, QBrush("#03000000"), QPen(strokeColor(), borderWidth()));
+        drawRoundedRect(painter, option->rect, 3, QBrush("#03000000"), QPen(strokeColor(), 1));
```
shoogle, line 527: Is this intentional? Looks like you are replacing a calculated value with a "magic number".

src/notation/internal/notationconfiguration.cpp
```diff
@@ -299,19 +310,42 @@ async::Notification NotationConfiguration::foregroundChanged() const
...
     return globalConfiguration()->appDataPath() + "/wallpapers";
 }
 
 QColor NotationConfiguration::borderColor() const
 {
-    return QColor(0, 0, 0, 102);
+    if (uiConfiguration()->isHighContrast()) {
+        return QColorConstants::White;
+    } else {
+        return QColor(0, 0, 0, 102);
```
shoogle, line 338:
```c++
return QColor(102, 102, 102, 0);
```

shoogle (changes requested):
I think the option to invert score colours should be available in normal themes, not just the high contrast themes. I might want to invert colours in the normal Dark mode.

Also, I see `borderWidth` referred to in lots of places seemingly for different things. I'm not sure what it has to do with score inversion.

src/framework/ui/view/uitheme.cpp
ArjunBoi: Ah yes this is intentional, since borderWidth is 1 for HC themes and 0 for general ones.
The border width around group boxes in the styles dialog needs to be 1 for general themes as well and not just the high contrast ones. 
Perhaps a more appropriate line would be:
```c++
drawRoundedRect(painter, option->rect, 3, QBrush("#03000000"), QPen(strokeColor(), configuration()->isHighContrast() ? borderWidth() : 1));
```
This would also account for further changes to High-Contrast borderWidth should someone choose to modify those values...

src/appshell/qml/Preferences/internal/UiColorsSection.qml
ArjunBoi: Ah I see what you mean

I have removed the `checked = !checked` line, and made the alias a bool property. The `onClicked` function now looks like this:
```c++
onClicked: {
            root.scoreInversionRequested(!checked)
        }
```
The system is now almost identical to the High Contrast checkbox.

shoogle:
> How would I go about fixing the unit tests?

> Running them manually in a debugger

I've added a new page to the Wiki on how to [fix the unit tests](https://github.com/musescore/MuseScore/wiki/Fix-the-unit-tests).

src/appshell/view/preferences/appearancepreferencesmodel.h
ArjunBoi: Fixed

src/engraving/libmscore/engravingitem.cpp
ArjunBoi: Ahh I never would've thought of this!
Understood and fixed!

src/notation/internal/notationconfiguration.cpp
ArjunBoi: Also fixed

shoogle: > Wouldn't your suggested color (102, 102, 102, 0) be fully transparent?

Oops yes! I just swapped the values (i.e. did the reverse of [this](https://github.com/musescore/MuseScore/pull/8886#discussion_r703296369)) but of course the alpha always needs to be 255 to remain full opacity.

> Perhaps this should be left as is? as i assume this acts just as a "shadow" around the score-page?

Yes, if it is really just a shadow then it needs to use alpha transparency, so you can ignore my comment! I thought this might be a dark colour used for the score background or something.

src/framework/ui/view/uitheme.cpp
shoogle: If the intent is *"follow `borderWidth()` where possible but the element must always have a border"* then I think this would be preferable:

```qml
borderWidth() > 1 ? borderWidth() : 1
```

Or even better:

```qml
Math.max(borderWidth(), 1)
```

I think I saw a few other place where you might want to do something similar.

It's best not to use High Contrast as a way to check for a non-zero border because we might have more themes in the future with different border sizes, or make the border size configurable outside of the High Contrast theme, etc.

ArjunBoi: Fixed!

ArjunBoi:
> I think the option to invert score colours should be available in normal themes, not just the high contrast themes. I might want to invert colours in the normal Dark mode.

Understood!
Now that the option is supposed to be visible constantly on the "Appearance" page (irrespective of theme), where should i put the score-inversion checkbox? @shoogle @Tantacrul

src/framework/ui/view/uitheme.cpp
```diff
@@ -32,9 +32,11 @@
 #include <QToolBar>
 #include <QTextEdit>
 #include <QVariant>
+#include <math.h>
 
 #include "log.h"
 
+
```
Jojo-Schmitz, line 39: this extra line cause the code style check to fail

ArjunBoi: Yes, fixed already :)

Jojo-Schmitz: You were quicxker in fixing that I in reporting ;-)

ArjunBoi:
Still working on the unit tests

Tantacrul:
@ArjunBoi  
> Now that the option is supposed to be visible constantly on the "Appearance" page (irrespective of theme), where should i put the score-inversion checkbox? @shoogle @Tantacrul

I would say at the bottom. Under 'Paper' and above the option to 'Reset theme to default'
image

ArjunBoi:
The requested changes have already been implemented, and i do not know how to resolve/get rid of this big ol' "Changes Requested" message...

Jojo-Schmitz:
Guess that question goes to @shoogle

RomanPudashkin (approved):

PR #8894, opened 2021-08-20
Fix #8893: allow for 128th as the denominator for measures

### LINKED ISSUE #8893: [MU4 Task] Support for more denominators in time signature
**Task description**
Support for 128 and 256 denominator in custom time signature.

Jojo-Schmitz:
Partly resolves: #8893


src/notation/view/widgets/measureproperties.ui
```diff
@@ -214,6 +214,11 @@
               <string>64</string>
              </property>
             </item>
+            <item>
+             <property name="text">
+              <string>128</string>
+             </property>
+            </item>
```
wizofaus, line 221: Isn't this going to be replaced with a qml version?

Jojo-Schmitz: Maybe, it isn't yet though, as far as I can tell

src/engraving/libmscore/splitMeasure.cpp
```diff
@@ -145,6 +144,10 @@ void Score::splitMeasure(Segment* segment)
             ticks2.setNumerator(ticks2.numerator() * mult);
         }
     }
+    if (ticks1.denominator() > 128 || ticks2.denominator() > 128) {
+        MScore::setError(MsError::CANNOT_SPLIT_MEASURE_TOO_SHORT);
+        return;
+    }
```
Jojo-Schmitz, line 150: Difficult to test, because of #8895. It doesn't crash, it doesn't split the measure, but also doesn't give any error message

AntonioBL:
The use of 64th as the denominator limit was also present in the MusicXML importer.
See, for example, the merged PR (in 3.x): https://github.com/musescore/MuseScore/pull/7208
Sometimes, the conversion to ticks could cause rounding errors; in that PR a check was done to prevent rounding errors, by assuming 1/64 as the minimum allowed step.
I don't know if extending to 1/128 in that part of the code could cause probems when importing old exported xml files.

Jojo-Schmitz:
Well, creating such a short measure (1/128) was possible before, via the split measure tool (and even shorter, leading to a crash, fixed with this PR). Just not the 'official' way, not via the measure poperties UI.

But let's see what the mtests say about such a change. and/ what checking the artifacts does say.

Jojo-Schmitz:
Capella import does allow for it too. But it seems BB and GTP import does not. Not sure whether this is because of MuseScore or because of BB and GTP though?
It is also not allowed in a time signature or tempo text (and I don't thing it is needed there).

Jojo-Schmitz:
Indeed an mtest is failing: https://github.com/musescore/MuseScore/pull/8894/checks?check_run_id=3398580316#step:8:10264

Jojo-Schmitz:
A musicxml file with a 1/128th measure does import cleanly into this PR, importing it into 3.6.2 has a issue though, it imports as an 1/64th, but with 2 128th rests, one of which made invisible. That much might be unavoidable I guess, 3.6.2's code not being prepared for such a short MusicXML measure die to that 64th rounding going on?

Jojo-Schmitz:
So maybe extending MusicXML import to support 1/128th is not really needed? Or at least seems a pretty much unrelated issue?

Jojo-Schmitz:
Actually it seems that mtest is loosing exactly that **unprinted** 128th rest, which it should not have shown in the first place, and only does because before that change it can't have a measure with just a sinfge 128th duration in it. So simply that mtest needs to get adjusted.

PR #8897, opened 2021-08-22
Fix build problems that are found when Unity Build is disabled

cbjeukendrup:
Also necessary when unity build is enabled but files are added or moved.


thirdparty/KDDockWidgets/src/private/Utils_p.h
```diff
@@ -12,7 +12,7 @@
 #ifndef KD_UTILS_P_H
 #define KD_UTILS_P_H
 
-#include "Config.h"
```
Eism, line 15: We shouldn't make changes in third-party libraries in our repository, because when we update the library, we replace our version with the new version and we may lose our changes

Jojo-Schmitz: No code style fixes either I guess.

igorkorsukov: In general, yes, third party libraries do not need to be changed. But in this case, I think, the changes are not critical, they can be kept, they will not interfere with the update to the new version.

igorkorsukov (approved):

src/importexport/guitarpro/internal/importgtp.h
```diff
@@ -122,6 +110,8 @@ struct GpBar {
     GpBar();
 };
 
+inline Drumset* gpDrumset = nullptr;
```
Jojo-Schmitz, line 113: Note to self: this is a C++17 feature!

thirdparty/KDDockWidgets/src/private/Utils_p.h
Jojo-Schmitz: Hmm, @igorkorsukov merged it anyhow?!

src/framework/midi/internal/platform/osx/coremidioutport.h
```diff
@@ -49,7 +49,7 @@ class CoreMidiOutPort : public IMidiOutPort
     void initCore();
 
     struct Core;
-    std::unique_ptr<Core> m_core;
+    std::shared_ptr<Core> m_core;
```
Eism, line 52: please fix for other platforms: Windows, Linux

Jojo-Schmitz: As @cbjeukendrup works on a Mac, he can't vouch for those I guess, At least he won't see the corresponding issues in his builds

igorkorsukov: I fixed it for Linux

Jojo-Schmitz: Where? Not in alsamidiinport.cpp and alsamidioutport.cpp as far as I can tell, only in the .h files?

Jojo-Schmitz: Hmm, at least not using `make_shared` there

igorkorsukov: https://github.com/musescore/MuseScore/blob/master/src/framework/midi/internal/platform/lin/alsamidiinport.h#L61
https://github.com/musescore/MuseScore/blob/master/src/framework/midi/internal/platform/lin/alsamidiinport.cpp#L49

Jojo-Schmitz: As I wrote: not using `make_shared` in the cpp files

Jojo-Schmitz: See #8899

PR #8903, opened 2021-08-23
[MU4] ENG-97: Dashed lines that cross systems become continuous

asattely:
The problem was that MuseScore was attempting to draw the hook with the solid line style even when there was no hook to be drawn (because the line continued to the next system)


Jojo-Schmitz:
a) don't merge, but pull (and puhh -f), to avoid those merge commits
b) don't use your master branch, to not block your self from furter work

asattely:
> a) don't merge, but pull (and puhh -f), to avoid those merge commits
> b) don't use your master branch, to not block your self from furter work

yup, that's my mistake, sorry!

cbjeukendrup:
And shouldn't those "ENG-##" pull requests (also) go to the `3.6.2_backend` branch?

Jojo-Schmitz:
All previous were. Here though I'm not even sure which problem it is trying to solve?

asattely:
> All previous were. Here though I'm not even sure which problem it is trying to solve?
> 
> Like how to reproduce the issue, in master and/or 3.x

It's a libmscore issue, where a dashed line with a centered hook (such as let ring or palm mute) would display as solid when continuing to the next system.
image

Jojo-Schmitz:
Ah! And indeed 3.x has the same issue

Jojo-Schmitz:
Only that code change doesn't fix it there (not tested it on master yet)

Jojo-Schmitz:
Actually it does, I've just been too stupid to apply it properly...

Jojo-Schmitz (approved):

Jojo-Schmitz (approved):

Jojo-Schmitz:
Also fixes https://musescore.org/en/node/307982

oktophonie:
ENG-* changes from now on are for master only.

Jojo-Schmitz:
Why do we need a 3rd bug tracker? And why one that is not public? And why for a bug that has an entry in one of the public trackers?

PR #8904, opened 2021-08-23
Actualization diagnostics module

igorkorsukov:
Added diagnostics menu item
Moved dev code to diagnostics module
Actualization engraving accessibility code


src/appshell/view/appmenumodel.cpp
```diff
@@ -54,7 +54,8 @@ void AppMenuModel::load()
         addItem(),
         formatItem(),
         toolsItem(),
-        helpItem()
+        helpItem(),
+        diagnosticItem()
```
cbjeukendrup, line 58: Wouldn't it be better to hide the Diagnostics menu under the Help menu? I don't think most users will want it to be visible all the time.

igorkorsukov: I plan that later this item will be hidden behind some kind of define. But now, before the release of at least a beta version, I think it will be useful to have it always visible.

vpereverzev (approved):

PR #8907, opened 2021-08-24
Add retry on saving file if temp file cannot be opened for writing. …

manish-makwana:
…Saving will be retried up to 3 times, with a 10ms wait between each attempt. This may help with saving from cloud synced folders (dropbox etc).

Resolves: https://musescore.org/en/comment/1093332

Changes:
- Retry opening the temp file up to 3 times.
- 10 millisecond wait between each attempt.

Reason:
- Workaround for race conditions if other programs (such as dropbox) lock the temp file for access at the same time.

Potential issues:
- Wait might not be long enough to avoid conflicts with other programs.
- Unsure if this is thread safe. (Does MuseScore use multithreading?)
- Other issues (such as filename format) might be causing the original error, not race conditions.

### DIFF
--- a/libmscore/scorefile.cpp
+++ b/libmscore/scorefile.cpp
@@ -36,6 +36,8 @@
 #include "stafftype.h"
 #include "sym.h"
 #include "scoreOrder.h"
+#include <chrono>
+#include <thread>
 
 #include "mscore/preferences.h"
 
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
The 3.6.2 branch is a dead end road. better use 3.x (but even that may turn out as a dead end), best use master.

manish-makwana:
@Jojo-Schmitz Ohh I didn't realise 3.x was a separate branch and more up to date haha.

I'm confused by the code structure in master. E.g. I can't find the function that saves files - seems to have been renamed. Are there any docs about the code reorganisation in v4?

Jojo-Schmitz:
For master scorefile.cpp is now in src/engraving/libmscore
But that temp file stuff seems gone, so I guess you're PR is not relevant for master anymore

wizofaus:
Try NotationProject::doSave

manish-makwana:
Ok thanks, seems like the best thing will be to wait for v4 to be released then :)

Issue #8819, opened 2021-08-09
[MU4 Issue] commands not implemented (including "Toggle Create Multimeasure Rest")


wizofaus:
From cmd.cpp, most are not currently implemented:

         "mirror-note",
         "add-trill",
         "add-up-bow",
         "add-down-bow",
         "reset-style",
         "clef-violin",
         "clef-bass",
         "sharp2-post",
         "sharp-post",
         "nat-post",
         "flat-post",
         "flat2-post",
         "add-audio",
         "transpose-up",
         "transpose-down",
         "pitch-up-diatonic-alterations",
         "pitch-down-diatonic-alterations",
         "full-measure-rest",
         "toggle-insert-mode",
         "pitch-up",
         "pitch-down",
         "pitch-up-octave",
         "pitch-down-octave",
         "pad-note-increase",
         "pad-note-decrease",
         "pad-note-increase-TAB",
         "pad-note-decrease-TAB",
         "toggle-mmrest",
         "toggle-hide-empty",
         "set-visible",
         "unset-visible",
         "relayout",
         "toggle-autoplace",
         "autoplace-enabled",

Some additional unimplemented commands also discovered along the way.
These are implemented in https://github.com/musescore/MuseScore/pull/8908:

pitch-up-diatonic 
pitch-down-diatonic
top-staff
empty-trailing-measure 
select-similar-range
edit-element
next-segment-element
prev-segment-element

These are not...

fotomode
playback-position　
apply-current-palette-element
insert-fretframe　　（apparently this is "experimental"!)
fbeam1
fbeam2
toggle-midiimportpanel  (are we keeping the midi import panel?)
toggle-fileoperations (hide/show "file" toolbar, which MU4 doesn't have at this point, not sure if it should?)
toggle-imagecapture (this command exists in the toolbar shown while in the "Publish" tab, but does nothing. How is it different to fotomode?)
toggle-feedback - will this toolbar be included in MU4 - seems overkill to have a separate toolbar just for this one command.  Note that the "feedback" command itself is in the Help menu and does work.
toggle-workspaces-toolbar - again, will MU4 have this toolbar?
toggle-selection-window - selection filter panel, was implemented recently but hooked up via a new different named action "toggle-selection-filter", which doesn't have a shortcut key assigned to it.
create-new-workspace - currently can only do this from the "configure workspaces" dialog
mag - called "Zoom canvas" in the shortcuts dialog, it actually crashes MU3! (there's no code to implement it)
[...]

Issue #7985, opened 2021-04-19
[MU4 Issue] [enhancement] the slur button should not be a toggle


vshalkevich:
**Describe the bug**
the slur button should not be a toggle: it should do the exact same thing as pressing the 'S' shortcut in MS3.6

Tantacrul:
This is a high priority. In addition, the slur should work EXACTLY like it does in MS3. When the user adds a slur, it should remain selected. In addition, the shortcut (on Mac) cmd+Shift+Left (or right) should extend or contract the slur.

I would assign this to either @RomanPudashkin  or @elerouxx because these bugs have been introduced due to the new Slur button in the Note Input bar changes.

wizofaus:
Cmd+Shift+Left/Right is implemented in my branch (https://github.com/musescore/MuseScore/pull/8908/).
It no longer appears to be a toggle, but it doesn't select it by default, can add that fix too easily enough.

wizofaus:
Ok but in MuseScore 3 the slurs are *not* selected when you add them, but they are in edit mode anyway.
Are you happy with my proposal that adding a slur should deselect whatever note was selected and select the slur instead?
And what about the cases where slurs are added over a list or range of notes? (where in v3 you can't adjust the newly added slur).

Tantacrul:
Oh, yes, I see. That MS3 behaviour would actually be quite clever... if it didn't block the user from using the main note input UI :) 
(although elements from palettes still work though, which is useful)

I think your suggestion for MS4 sounds about right. If indeed we could be clever and try not to interrupt further note input, that would be ideal. I have suggestions for new 'special' types of selection we could try in later versions but def out of scope for this.

Regarding list / range selections.... I'm in two minds. I can see why there wouldn't be a need to select the slur in those cases since multi-selection effectively specifies the length, so users most likely wouldn't need to shorten / lengthen them. I actually suspect MS3 has this right (despite all other major apps **_not_** doing this)

Tantacrul:
About Palettes: we've deliberately not fiddled with this in MS4. I think there are some pretty simple and powerful UX changes we can make to how elements behave when applied / dragged. Won't burden this conversation with that detail though!

wizofaus:
Ok but now we're proposing that slurs have a different behaviour to all other lines/spanners, in that they *do* get selected when added.  It'd be just as easy to make that true for all lines rather than just slurs if it makes more sense from the user POV.

wizofaus:
[...]

Issue #7946, opened 2021-04-12
[MU4 Issue] Many shortcuts are not implemented


aaaeka:
**Describe the bug**
Many shortcuts from the ```shortcuts.xml``` file are not yet implemented in MU4

**List of unimplemented shortcuts by action name**

- [x] chord-tie - #7967
- [x] rest - #7967
- [x] rest-1 - #7967
- [x] rest-2 - #7967
- [x] rest-4 - #7967
- [x] rest-8 - #7967
- [x] up-chord - #7967
- [x] down-chord - #7967
- [x] first-element - #7967
- [x] last-element - #7967
- [x] palette-search - implemented, but  takes long time to complete
- [ ] get-location
- [ ] move-up
- [ ] move-down
- [ ] select-next-measure
- [ ] select-prev-measure
- [ ] select-begin-line
- [ ] select-end-line
- [ ] select-begin-score
- [ ] select-end-score
- [ ] select-staff-above
- [ ] select-staff-below
- [ ] scr-next
- [ ] scr-prev
- [ ] page-prev
- [ ] page-next
- [ ] page-top
- [ ] page-end
- [x] append-measure
- [ ] clef-violin
- [ ] clef-bass
- [ ] reset - implemented incorrectly
- [ ] help
- [ ] zoom100
- [ ] mirror-note
- [x] double-duration
- [x] half-duration
- [x] inc-duration-dotted
- [x] dec-duration-dotted
- [ ] repeat-sel
- [x] enh-both
- [x] enh-current
- [ ] fullscreen - only works with Meta+Ctrl+F, but not with Ctrl+U
- [ ] symbols
- [x] toggle-visible
- [ ] lock
- [ ] print

wizofaus:
I'm doing move-up/move-down (and some others not mentioned here in https://github.com/musescore/MuseScore/pull/8879).  I think quite a number of these (particularly the select- ones) are done already.

wizofaus:
Will start going through these and any others that don't appear to be have been ported from MU3 yet, happy for you to assign to me @Tantacrul.

wizofaus:
palette-search - works but at least in my debug build can sometimes take a *long* time to react, >2 seconds!

get-location - unimplemented, but I'm not 100% sure what it's supposed to do in MU3, do I need to install NVDA?  I can implement code that does mostly the same thing, but will be hard to verify.

[...]

Issue #7697, opened 2021-03-06
[MU4 Issue] Add slur (shortcut: S), should leave you in edit mode on right-most endpoint


vshalkevich:
**Describe the bug**
Slurs are not working correctly.

**To Reproduce**
- add two adjacent notes
- select the first
- click "S" / add slur command
- slur added, but not in edit mode as it is per MU3
- IMPORTANT - the shortcut to extend the slur (Shift+Alt+ Right arrow) is not working

**Desktop (please complete the following information):**
All

wizofaus:
Specifically the repro steps are:
1. add two adjacent notes
2. select the first
3. click "S" / add slur command
4. slur added, but not in edit mode as it is per MU3

Note that in MU3 the slur isn't in edit mode if you select multiple notes before adding the slur, not quite sure why the difference. Can you confirm we want to preserve/restore MU3 behaviour here @Tantacrul ?

Eism:
@wizofaus 
see https://github.com/musescore/MuseScore/issues/7985#issuecomment-833671601
'EXACTLY like it does in MS3'

I am working on it now

wizofaus:
Are you still working on this @Eism as I've made a lot of changes around edit mode for slurs and in fact I'm still working on it (specifically, detecting when the start/end handles are near noteheads etc., and that anchor lines are drawn correctly). But I haven't done anything with the specific list listed here.

wizofaus:
See my comments in https://github.com/musescore/MuseScore/issues/7985, I didn't actually realise it was possible for an element to be in edit mode without being selected until I looked into this, which is going to be tricky to support. But I have already done the bulk of what's necessary here.

Tantacrul:
Changed to P1 because this is one of the most important notated elements and the inability to extend it using a shortcut is a serious problem.

PR #8908, opened 2021-08-24
[MU4] port remaining actions from MU3 (WIP)

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8819, https://github.com/musescore/MuseScore/issues/7946 (partly), https://github.com/musescore/MuseScore/issues/7985, https://github.com/musescore/MuseScore/issues/7697 etc. etc.

0) added/fixed support for following actions:
    next-frame = Next frame
    prev-frame = Previous frame
    next-system = Next system
    prev-system = Previous system
    show-irregular  = Mark irregular measures
    toggle-insert-mode = Toggle 'insert mode'
    zoom100 = zoom to 100% 
    get-location = Accessibility: get location  (untested with screen-reader)
    edit-element = Edit element 
    select-prev-measure = Select to beginning of measure
    select-next-measure = Select to end of measure
    select-begin-line = Select to beginning of line
    select-end-line = Select to end of line
    select-begin-score = Select to beginning of score
    select-end-score = Select to end of score
    select-staff-above = Add staff above to selection
    select-staff-below = Add staff below to selection
    scr-prev = Screen: Previous
    scr-next = Screen: Next
    page-prev = Page: Previous
    page-next = Page: Next
    page-top = Page: Top of first
    page-end = Page: Bottom of last
    help = Context sensitive help
    repeat-sel = Repeat selection
    pitch-up-diatonic = Diatonic pitch up
    pitch-down-diatonic = Diatonic pitch down
    top-staff = Go to top staff
    empty-trailing-measure = Go to first empty trailing measure
    mirror-note    = Mirror notehead
    add-trill = Toggle trill
    add-up-bow = Toggle up bow
    add-down-bow = Toggle down bow
    clef-violin = Add treble clef
    clef-bass = Add bass clef
    sharp2-post = Double ♯ (non-toggle)
    sharp-post = ♯ (non-toggle)
    nat-post = ♮ (non-toggle)
    flat-post = ♭ (non-toggle)
    flat2-post = Double ♭ (non-toggle)
    transpose-up = Transpose up
    transpose-down = Transpose down
    pitch-up-diatonic-alterations = Diatonic pitch up (keep degree alterations)
    pitch-down-diatonic-alterations = Diatonic pitch down (keep degree alterations)
    full-measure-rest = Full measure rest
    toggle-mmrest = Toggle 'Create multimeasure rest'
    toggle-hide-empty = Toggle 'Hide empty staves'
    set-visible = Set visible
    unset-visible = Set invisible
    toggle-autoplace = Toggle 'automatic placement' for selected elements
    autoplace-enabled = Toggle 'automatic placement' (whole score)
    pad-note-increase = Note input: double duration
    pad-note-decrease = Note input: halve duration
    pad-note-increase-TAB = Note input: double duration (TAB)
    pad-note-decrease-TAB = Note input: halve duration (TAB)
    move-prev = Move chord/rest left (note entry mode only, actually swaps)
    move-next = Move chord/rest right (note entry mode only, actually swaps)
    next-segment-element = Accessibility: Next segment element
    prev-segment-element = Accessibility: Previous segment element

(I'd previously had "reset-style" = Reset all style values to default, but removed as undocumented, had no shortcut by default, and the underlying implementation had been removed)

1) hid redundant top grip handle for barlines
2) keyboard support for modifying barline extent (up/down cursor keys)
3) ensured you can re-adjust barline span w/o CTRL key if it had previously been adjusted using CTRL key
4) ensured that when barline is dragged with mouse down to another staff, the barline is only broken below that staff if it was already joined to the staff
5) fixed bug dragging beams with mouse (wasn't working)
6) added keyboard support to adjusting bracket span (up/down cursor keys)
7) changed "full-screen" shortcut key to be F11 (Ctrl+U already meant 'underline')
8) ensured that preferences|shortcuts dialog shows *all* known actions/commands, not just those that shortcuts are defined for
9) ensured that inspector/properties panel updates correctly even if selection hasn't changed (e.g. use 'v' key while an element is selected to toggle hide/show flag)
10) ensure "select similar elements in range" works
11) ensure the grip handles are selected in preference to elements underneath them (when trying to click on them)
12) ensure text edit mode is triggered when clicking on an already selected text element
13) ensures CTRL+click toggles selection of elements even when there's a range already selected
14) ensures ALL elements can be dragged in edit mode
15) enter edit mode when double-clicking on any element
16) fixed group dragging to work as per MU3 (e.g. select a range then drag a note, it only moves the notes, but if you select a *list* it moves everything)
17) edit mode now mostly works as per MU3:
    * all elements in edit mode (except clefs/instrument names which don't support it) can be freely dragged around, usually adjusting x/y offset
    * cursor keys in edit mode generally also adjust x/y offset, with CTRL/ALT/SHIFT modifiers acting as per MU3
    * no other commands available in edit mode, must hit "Esc" to return to regular mode
 18) various fixes for adjusting grip handles, particularly for slurs, such that a) when mouse is dragged over notehead the notehead is highlighted and the slur "snaps" to it, b) anchor lines drawn as expected
 19) fixed "add-slur" so that a) in note entry mode it toggles based on whether a slur is *active*, not whether one is present, as per v3, b) after adding a slur it immediately goes into edit mode and can be adjusted with the keyboard 


src/notation/internal/notationactioncontroller.cpp
```diff
@@ -36,6 +36,40 @@ static const ActionCode ESCAPE_ACTION_CODE = "escape";
...
+                                              P1 param1, bool (NotationActionController::* enabler)() const)
+{
+    registerAction(code, [this, handler, param1]()
+    {
+        auto interaction = currentNotationInteraction().get();
+        if (interaction) {
+            (interaction->*handler)(param1);
+        }
+    }, enabler);
+}
```
wizofaus, line 50: Will move this to the end of the file, I'm pretty sure modern compilers can cope with template definitions being anywhere in the same compilation unit (they didn't 20 years ago!)

src/notation/internal/notationinteraction.cpp
```diff
@@ -3495,3 +3650,55 @@ void NotationInteraction::toggleUnderline()
...
+}
+
+void NotationInteraction::insertClef(Ms::ClefType clef)
+{
+    execute(&Ms::Score::cmdInsertClef, clef);
+}
+
+void NotationInteraction::changeAccidental(Ms::AccidentalType accidental)
+{
+    execute(&Ms::Score::changeAccidental, accidental);
```
wizofaus, line 3680: TODO: need a way of indicating that this is a pitch-change and requires playing back the currently selected note

src/appshell/internal/applicationuiactions.cpp
```diff
@@ -42,15 +42,15 @@ const UiActionList ApplicationUiActions::m_actions = {
              ),
     UiAction("about",
              mu::context::UiCtxAny,
-             QT_TRANSLATE_NOOP("action", "About...")
+             QT_TRANSLATE_NOOP("action", "About�")
```
wizofaus, line 45: Just for you @Jojo-Schmitz

src/engraving/libmscore/cmd.cpp
```diff
@@ -4359,9 +4359,9 @@ void Score::cmd(const QString& cmd, EditData& ed)
     static const std::vector<ScoreCmd> cmdList {
         { "toggle-visible",             [](Score* cs, EditData&) { cs->cmdToggleVisible(); } },
```
wizofaus, line 4360: actually this function is going, just leaving it here for reference

src/engraving/libmscore/score.h
```diff
@@ -515,24 +515,14 @@ class Score : public QObject, public ScoreElement
     mu::engraving::Layout m_layout;
     mu::engraving::LayoutOptions m_layoutOptions;
 
-    ChordRest* nextMeasure(ChordRest* element, bool selectBehavior = false, bool mmRest = false);
-    ChordRest* prevMeasure(ChordRest* element, bool mmRest = false);
-    void cmdResetAllStyle();
-
     Note* getSelectedNote();
-    ChordRest* upStaff(ChordRest* cr);
-    ChordRest* downStaff(ChordRest* cr);
```
wizofaus, line 524: A number of these needed to be made public in order to replace `cmd()` and `selectMove()`

src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -256,9 +265,15 @@ Notification ShortcutsRegister::shortcutsChanged() const
     return m_shortcutsChanged;
 }
 
-const Shortcut& ShortcutsRegister::shortcut(const std::string& actionCode) const
+const Shortcut& ShortcutsRegister::shortcut(const std::string& actionCode)
 {
-    return findShortcut(m_shortcuts, actionCode);
+    const auto& shortCut = findShortcut(m_shortcuts, actionCode);
```
wizofaus, line 270: This is a bit of a kludge, but it ensures that the ShortcutsRegister knows about all actions, even those that aren't in the shipped shortcuts.xml files, as currently the shortcuts dialog uses this to build the list of actions/shortcut assignments (but perhaps it should get them from UiActionsRegister instead...)

src/appshell/internal/applicationuiactions.cpp
wizofaus: D'oh, except now it looks like this 
image

Jojo-Schmitz: Not good. I'm currently fighting with this too. Somehint about UTF8 vs. some Windowes codepage I guess

Jojo-Schmitz: It does work for editstyle.ui (in addition to make them untranslatable) and in notationuiactions.cpp though

wizofaus: It's exactly that, for some reason notationuiactions.cpp gets "correctly" guessed as being UTF-8 whereas applicationuiactions.cpp doesn't.  I honestly have no idea why yet. I can put a UTF-8 encoding marker at the beginning but I don't know if all compilers on all platforms will recognise that correctly?

Jojo-Schmitz: They should, this explict marker is in some/many(?) files already I believe

Jojo-Schmitz: Visual Studio seems to have some internal heuristic to automagically detect this, apparently not really reliable though

wizofaus: I think I've fixed it, even without the UTF-8 encoding marker, will push again

wizofaus: Well I saved them *with* the marker then removed them with a binary editor and they seem OK now - but yes, probably better to have the markers there as long as all tools on all platforms know how to read them (including uncrustify etc.)

Jojo-Schmitz: Mind to do the same for diagnosticsactions.cpp, extensionsservice.cpp, editmidimappingmodel.cpp, instrumentsuiactions.cpp, languagesservice.cpp, projectuiactions.cpp?

Jojo-Schmitz: editstyle.ui and notationsuiactions.cpp need the ... change too, but not the UTF one

wizofaus: I think I've changed quite enough for one PR! They're not really related and I don't have time to test them all.  notationuiactions.cpp should be fixed throughout.

src/framework/shortcuts/data/shortcuts-Mac.xml
```diff
@@ -833,7 +833,7 @@
     </SC>
   <SC>
     <key>fullscreen</key>
-    <seq>Ctrl+U</seq>
+    <seq>F11</seq>
```
wizofaus, line 836: F11 used to mean "toggle play panel" but that's gone in MU4, and F11 is a fairly common key in Windows apps at least to toggle full screen mode, and best of all doesn't conflict with Ctrl+U to mean "toggle underline".

src/appshell/internal/applicationuiactions.cpp
Jojo-Schmitz: OK. I'll see what I can do to the rest

src/notation/internal/notationinteraction.cpp
```diff
@@ -2845,7 +2995,52 @@ void NotationInteraction::replaceSelectedNotesWithSlashes()
     notifyAboutNotationChanged();
 }
 
-void mu::notation::NotationInteraction::changeEnharmonicSpelling(bool both)
+void NotationInteraction::repeatSelection()
+{
+    const Ms::Selection& selection = score()->selection();
```
wizofaus, line 3000: copied from ScoreView::cmdRepeatSelection, with some simplifications (actually I'm pretty sure it can be simplified even further)

src/appshell/internal/applicationuiactions.cpp
wizofaus: Interestingly that question-mark in a diamond is how it shows in the notification email too, but not on github.com. Go figure.  If you're confident that the UTF-8 BOM works then we should add it all files (or at least, any that have non-ASCII characters).

Jojo-Schmitz: See #8918

Jojo-Schmitz: The 'trick' was to use vi(m) rather than the builtin editor from Visual Studio

Jojo-Schmitz: It's been merged, maybe you need (or want) to rebase

Jojo-Schmitz:
rebase needed

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -246,6 +252,79 @@ void NotationViewInputController::setViewMode(const ViewMode& viewMode)
...
+    auto notation = currentNotation();
+    if (!notation) {
+        return;
+    }
+    auto scale = m_view->currentScaling();
+    if (notation->viewMode() == ViewMode::LINE) {
+        m_view->moveCanvasHorizontal(m_view->width() * direction * scrollStep / scale);
+    } else {
+        m_view->moveCanvasVertical(m_view->height() * direction * scrollStep / scale);
+    }
```
wizofaus, line 270: This implementation is actually quite a lot simpler than the one in MU3 (in ScoreView.cpp) - it has extra logic for frames that I couldn't quite figure out the need for.

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -246,6 +252,79 @@ void NotationViewInputController::setViewMode(const ViewMode& viewMode)
...
+}
+
+void NotationViewInputController::endOfScore()
+{
+    auto offset = m_view->toLogical(QPoint());
+    auto rect = m_view->notationContentRect();
+    auto scale = m_view->currentScaling();
+    m_view->moveCanvas(offset.x() + m_view->width() / scale - thickPadding - rect.width(),
+                       offset.y() + m_view->height() / scale - thickPadding - rect.height());
+}
```
wizofaus, line 326: Actually this doesn't work properly for the case where the last page isn't "full" - it always scrolls to the bottom of the final page.  Needs a bit more thought.

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -246,6 +252,85 @@ void NotationViewInputController::setViewMode(const ViewMode& viewMode)
...
+    if (!notation) {
+        return;
+    }
+    Ms::MeasureBase* lastMeasure = notation->elements()->msScore()->lastMeasureMM();
+    auto lmRect = lastMeasure->canvasBoundingRect();
+    auto scale = m_view->currentScaling();
+    qreal desiredX = std::max(-thickPadding, lmRect.right() + thickPadding - m_view->width() / scale);
+    qreal desiredY = std::max(-thickPadding, lmRect.bottom() + thickPadding - m_view->height() / scale);
+    auto offset = m_view->toLogical(QPoint());
+    m_view->moveCanvas(offset.x() - desiredX, offset.y() - desiredY);
```
wizofaus, line 331: Again, much simpler than the MU3 logic and doesn't behave precisely the same way (MU3 seems to scroll unnecessarily far, leaving lots of negative space to the right of the score), but in all cases I tested it ensured the final measure in the last staff (i.e. the bottom right corner of the final measure) was "just" scrolled into view with minimal negative space.

wizofaus:
NB as part of implementing edit-element I also made a number of fixes to how elements with grip-handles behave when selected, including when their anchor lines are drawn etc.  Some behaviours that are slightly different to MU3 (re specific "edit mode") to be confirmed with @Tantacrul.

src/notation/internal/notationinteraction.cpp
```diff
@@ -2012,115 +2116,201 @@ void NotationInteraction::startEditText(Element* element, const PointF& cursorPo
...
+        if (editData.modifiers & Qt::ControlModifier) {
+            return spatium;
+        } else if (editData.modifiers & Qt::AltModifier) {
+            return 4 * spatium;
+        } else {
+            return 0.25 * spatium;
+        }
+    } else {
+        if (editData.modifiers & Qt::ControlModifier) {
+            return Ms::MScore::nudgeStep10 * spatium;
```
Jojo-Schmitz, line 2156: For some bizarre reason uncrustify wants it to be `return Ms::MScore::nudgeStep10* spatium;`

src/notation/internal/notationinteraction.cpp
```diff
@@ -2012,115 +2116,201 @@ void NotationInteraction::startEditText(Element* element, const PointF& cursorPo
...
+        } else if (editData.modifiers & Qt::AltModifier) {
+            return Ms::MScore::nudgeStep50 * spatium;
```
Jojo-Schmitz, line 2158: here too

src/notation/internal/notationinteraction.cpp
```diff
@@ -2012,115 +2116,201 @@ void NotationInteraction::startEditText(Element* element, const PointF& cursorPo
...
+        } else {
+            return Ms::MScore::nudgeStep * spatium;
```
Jojo-Schmitz, line 2160: and here

wizofaus: Yep, logged it as an issue with whatever version of uncrustify is being used on github, as what it recommends makes no sense,  looks like it's parsing it as a pointer.

wizofaus:
@vpereverzev I'm not sure if you or Igor will have time to look through this before it gets too much bigger - there's still a few things I need to sort out with @Tantacrul re the expected behaviour but it could definitely use looking over from an architectural/code-design angle.  Will resolve the conflicts ASAP.

src/notation/internal/notationinteraction.cpp.uncrustify
```diff
@@ -0,0 +1,3918 @@
+/*
```
Jojo-Schmitz, line 1: That file doesn't blong into the repo at all, and causes the GitHub CI codestyle failure

src/engraving/libmscore/barline.cpp
```diff
@@ -1058,6 +1058,24 @@ void BarLine::startEdit(EditData& ed)
...
+            }
+        } else {
+            score()->staff(staffIdx())->undoChangeProperty(Pid::STAFF_BARLINE_SPAN, !spanStaff());
+        }
+        return true;
+    }
+    return Element::edit(ed);
+}
+
+
```
Jojo-Schmitz, line 1078: Now this one extra empty lined cause the codestyle check to fail.
But that seems to be the least problem of your latest changes, which don#t seem to build at all?

src/engraving/libmscore/barline.cpp
```diff
@@ -1040,7 +1040,7 @@ std::vector<PointF> BarLine::gripsPositions(const EditData& ed) const
     const PointF pp = pagePos();
 
     return {
-        PointF(lw * .5, y1 + bed->yoff1) + pp,
+        //PointF(lw * .5, y1 + bed->yoff1) + pp,
```
wizofaus, line 1043: removing redundant upper grip handle for Barline object, @Tantacrul was keen to test it like this before committing definitely (it's redundant as a grip handle but does at least provide visual feedback where the top of the barline is...)

src/engraving/libmscore/barline.cpp
```diff
@@ -1058,6 +1058,23 @@ void BarLine::startEdit(EditData& ed)
...
+            BarLine* b = toBarLine(segment()->element(staffIdx() * VOICES));
+            if (b) {
+                b->undoChangeProperty(Pid::BARLINE_SPAN, !spanStaff());
+            }
+        } else {
+            score()->staff(staffIdx())->undoChangeProperty(Pid::STAFF_BARLINE_SPAN, !spanStaff());
+        }
+        return true;
+    }
+    return EngravingItem::edit(ed);
```
wizofaus, line 1075: adding keyboard support to barline extend/break

src/engraving/libmscore/barline.cpp
```diff
@@ -1137,9 +1153,10 @@ void BarLine::endEditDrag(EditData& ed)
     int newSpanFrom = 0;
     int newSpanTo = 0;
 
-    bool localDrag = ed.control() || segment()->isBarLineType();
+    bool localDrag = ed.control() || segment()->isBarLineType() || spanStaff() != score()->staff(staffIdx())->barLineSpan();
```
wizofaus, line 1156: fixes a bug (also in MU3) where trying to adjust a barline previously adjusted using "CTRL" key modifier doesn't work

src/engraving/libmscore/barline.cpp
```diff
@@ -1150,17 +1167,24 @@ void BarLine::endEditDrag(EditData& ed)
...
         }
     } else {
+        bool breakLast = false;
         for (int staffIdx = staffIdx1; staffIdx < staffIdx2; ++staffIdx) {
+            breakLast = score()->staff(staffIdx)->barLineSpan();
             score()->staff(staffIdx)->undoChangeProperty(Pid::STAFF_BARLINE_SPAN, true);
         }
-        score()->staff(staffIdx2)->undoChangeProperty(Pid::STAFF_BARLINE_SPAN, false);
+        if (breakLast) {
+            score()->staff(staffIdx2)->undoChangeProperty(Pid::STAFF_BARLINE_SPAN, false);
```
wizofaus, line 1186: came out of discussions with @Tantacrul  and @MarcSabatella around when barlines should be broken after dragging handle down, essentially, they shouldn't be if barline was already broken before the staff you drag the barline onto.

src/engraving/libmscore/bracket.cpp
```diff
@@ -343,6 +343,7 @@ void Bracket::endEdit(EditData& ed)
     triggerLayoutAll();
     score()->update();
     ed.element = 0;           // score layout invalidates element
+    ed.clearData();
```
wizofaus, line 346: ed.data can be left containing reference to deleted Bracket element otherw ise

src/engraving/libmscore/bracket.cpp
```diff
@@ -389,9 +390,6 @@ void Bracket::endEditDrag(EditData&)
     qreal ey = system()->staff(staffIdx2)->y() + score()->staff(staffIdx2)->height();
     h2 = (ey - sy) * .5;
     bracketItem()->undoChangeProperty(Pid::BRACKET_SPAN, staffIdx2 - staffIdx1 + 1);
-    // brackets do not survive layout
-    // make sure layout is not called:
-    score()->cmdState()._setUpdateMode(UpdateMode::Update);
```
wizofaus, line 394: This is no longer needed, and in fact causes a problem that the barlines are not updated properly

src/engraving/libmscore/bracket.cpp
```diff
@@ -426,6 +424,15 @@ EngravingItem* Bracket::drop(EditData& data)
...
 bool Bracket::edit(EditData& ed)
 {
+    if (ed.key == Qt::Key_Up && span() > 1) {
+        bracketItem()->undoChangeProperty(Pid::BRACKET_SPAN, span() - 1);
+        return true;
+    }
+    if (ed.key == Qt::Key_Down && _lastStaff < system()->staves()->size() - 1) {
+        bracketItem()->undoChangeProperty(Pid::BRACKET_SPAN, span() + 1);
+        return true;
+    }
```
wizofaus, line 434: adding keyboard support for adjust bracket span

src/engraving/libmscore/cmd.cpp
```diff
@@ -4359,9 +4359,9 @@ void Score::cmd(const QString& cmd, EditData& ed)
...
         { "mirror-note",                [](Score* cs, EditData&) { cs->cmdMirrorNoteHead(); } },
-        { "add-trill",                  [](Score* cs, EditData&) { cs->addArticulation(SymId::ornamentTrill); } },
-        { "add-up-bow",                 [](Score* cs, EditData&) { cs->addArticulation(SymId::stringsUpBow); } },
-        { "add-down-bow",               [](Score* cs, EditData&) { cs->addArticulation(SymId::stringsDownBow); } },
+        { "add-trill",                  [](Score* cs, EditData&) { cs->toggleArticulation(SymId::ornamentTrill); } },
+        { "add-up-bow",                 [](Score* cs, EditData&) { cs->toggleArticulation(SymId::stringsUpBow); } },
+        { "add-down-bow",               [](Score* cs, EditData&) { cs->toggleArticulation(SymId::stringsDownBow); } },
```
wizofaus, line 4364: all these actions are actually toggles despite the name

src/engraving/libmscore/undo.cpp
```diff
@@ -92,7 +92,7 @@
 #include "masterscore.h"
 
 #include "log.h"
-#define LOG_UNDO() if (0) LOGD()
+#define LOG_UNDO() if (1) LOGD()
```
wizofaus, line 95: will revert this, just for personal diagnostic purposes!

src/inspector/models/inspectorlistmodel.cpp
```diff
@@ -293,6 +293,11 @@ void InspectorListModel::subscribeOnSelectionChanges()
             setElementList(QList(elements.cbegin(), elements.cend()));
         });
 
+        m_notation->notationChanged().onNotify(this, [this]() {
+            auto elements = m_notation->interaction()->selection()->elements();
+            setElementList(QList(elements.cbegin(), elements.cend()));
+        });
```
wizofaus, line 299: this is quite critical - even if what elements are selected hasn't changed, if properties of said elements have, the inspector needs to update itself

src/notation/inotationselection.h
```diff
@@ -49,6 +49,8 @@ class INotationSelection
     virtual QRectF canvasBoundingRect() const = 0;
 
     virtual INotationSelectionRangePtr range() const = 0;
+
+    virtual EngravingItem* lastElementHit() const = 0;
```
wizofaus, line 53: to facilitate "select similar elements in range"

src/notation/internal/notationinteraction.h
```diff
@@ -256,14 +280,17 @@ class NotationInteraction : public INotationInteraction, public async::Asyncable
...
+    void updateGripEdit();
     void resetGripEdit();
 
     void resetStretch();
     void resetTextStyleOverrides();
     void resetBeamMode();
     void resetShapesAndPosition();
 
+    template<typename P>
+    void execute(void (Ms::Score::* function)(P), P param);
```
wizofaus, line 292: "catch-all" function to allow executing Score:: methods without INotationInteraction needing to get too bloated

src/notation/view/notationpaintview.h
```diff
@@ -126,7 +124,7 @@ class NotationPaintView : public QQuickPaintedItem, public IControlledView, publ
     void moveCanvasToPosition(const QPoint& logicPos);
     double guiScaling() const;
 
-    QRectF notationContentRect() const;
+    QRectF notationContentRect() const override;
```
wizofaus, line 127: actually I didn't end up needing this, but could be useful

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -333,6 +419,13 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
         return;
     }
 
+    if (viewInteraction()->isHitGrip(logicPos)) {
+        viewInteraction()->startEditGrip(logicPos);
+        return;
```
wizofaus, line 424: ensures grips get priority over elements underneath them

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -333,6 +419,13 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
...
+    }
+
+    m_pressedOnSelected = hitElement && hitElement == viewInteraction()->selection()->element();
```
wizofaus, line 427: needed to allow clicking on an already selected text element to enter text edit mode (as per MU3)

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -352,18 +445,11 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
...
 
-    if (viewInteraction()->isHitGrip(logicPos)) {
-        viewInteraction()->startEditGrip(logicPos);
-        return;
-    }
-
     if (hitElement) {
         playbackController()->playElement(hitElement);
-    } else {
-        viewInteraction()->endEditGrip();
```
wizofaus, line 363: now handled by call to endEditElement() below

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -504,6 +592,8 @@ void NotationViewInputController::mouseReleaseEvent(QMouseEvent*)
 
     if (viewInteraction()->isDragStarted()) {
         viewInteraction()->endDrag();
+    } else if (m_pressedOnSelected && viewInteraction()->selection()->element()->isTextBase()) {
+        viewInteraction()->startEditText(viewInteraction()->selection()->element(), PointF(m_view->toLogical(event->pos())));
```
wizofaus, line 596: as per MU3, when clicking on an already selected text element, start text edit mode - but it only occurs on button *release* (whereas the selection happens on press).

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -517,6 +607,8 @@ void NotationViewInputController::mouseDoubleClickEvent(QMouseEvent* event)
 
     if (element->isTextBase()) {
         viewInteraction()->startEditText(element, m_view->toLogical(event->pos()));
+    } else {
+        viewInteraction()->startEditElement();
```
wizofaus, line 611: whether or not we need to support "element edit mode" is still under discussion - but for now, it is in order to enable using the keyboard to make offset/leading space adjustments

src/engraving/libmscore/beam.cpp
```diff
@@ -2249,7 +2249,7 @@ void Beam::editDrag(EditData& ed)
     // Because of the logic in Beam::setProperty(),
     // changing Pid::BEAM_POS only has an effect if Pid::USER_MODIFIED is true.
     undoChangeProperty(Pid::USER_MODIFIED, true);
-    undoChangeProperty(Pid::BEAM_POS, PointF(y1 / _spatium, y2 / _spatium));
+    undoChangeProperty(Pid::BEAM_POS, QVariant::fromValue(qMakePair(y1 / _spatium, y2 / _spatium)));
```
wizofaus, line 2252: this was preventing using the mouse from editing Beams (my fault, I'd changed this property to use a `QPair<float,float>` a while back but missed a few places)

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -436,7 +524,7 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
     const EngravingItem* hitElement = this->hitElement();
 
     // drag element
-    if (!middleButton && ((hitElement && hitElement->isMovable())
+    if (!middleButton && ((hitElement && (hitElement->isMovable() || viewInteraction()->isElementEditStarted()))
```
wizofaus, line 527: all elements are draggable in edit mode

src/notation/internal/notationinteraction.cpp
```diff
@@ -735,21 +735,25 @@ void NotationInteraction::drag(const PointF& fromPos, const PointF& toPos, DragM
...
-        m_dragData.ed.curGrip = m_gripEditData.curGrip;
+        m_dragData.ed.curGrip = m_editData.curGrip;
         m_dragData.ed.delta = m_dragData.ed.pos - m_dragData.ed.lastPos;
         m_dragData.ed.moveDelta = m_dragData.ed.delta - m_dragData.elementOffset;
-        m_dragData.ed.addData(m_gripEditData.getData(m_gripEditData.element));
-        m_gripEditData.element->editDrag(m_dragData.ed);
+        m_dragData.ed.addData(m_editData.getData(m_editData.element));
+        m_editData.element->editDrag(m_dragData.ed);
+    } else if (isElementEditStarted()) {
+        m_dragData.ed.delta = evtDelta;
```
wizofaus, line 755: weird, but the base implementation for Element::editDrag (sorry, EngravingItem::editDrag) requires this (for dragging in edit mode)

src/notation/internal/notationinteraction.cpp
```diff
@@ -779,35 +783,32 @@ void NotationInteraction::drag(const PointF& fromPos, const PointF& toPos, DragM
...
     if (isGripEditStarted()) {
-        m_gripEditData.element->endEditDrag(m_gripEditData);
+        m_editData.element->endEditDrag(m_editData);
+        m_editData.element->endEdit(m_editData);
     } else {
         for (auto& group : m_dragData.dragGroups) {
             group->endDrag(m_dragData.ed);
         }
-        if (!m_lasso->bbox().isEmpty()) {
+        if (m_lasso && !m_lasso->bbox().isEmpty()) {
```
wizofaus, line 797: prevents crash

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -343,6 +436,9 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
             selectType = SelectType::ADD;
         }
         viewInteraction()->select({ hitElement }, selectType, hitStaffIndex);
+    } else if (hitElement && event->button() == Qt::MouseButton::LeftButton
+               && event->modifiers() == Qt::KeyboardModifier::ControlModifier) {
+        viewInteraction()->select({ hitElement }, SelectType::ADD, hitStaffIndex);
```
wizofaus, line 441: restores MU3 behaviour of toggling selection for an element on left button+CTRL click

src/engraving/libmscore/mscoreview.cpp
```diff
@@ -89,4 +89,65 @@ const QList<EngravingItem*> MuseScoreView::elementsAt(const mu::PointF& p)
...
+EngravingItem* MuseScoreView::elementNear(const mu::PointF& pos) const
+{
+    QList<EngravingItem*> near = elementsNear(pos);
+    if (near.isEmpty()) {
+        return nullptr;
+    }
+    return near.front();
+}
+
+const QList<EngravingItem*> MuseScoreView::elementsNear(const mu::PointF& pos) const
```
wizofaus, line 102: ported from v3 (ScoreView::elementsNear)

src/notation/internal/notationinteraction.cpp
```diff
@@ -752,43 +771,32 @@ void NotationInteraction::drag(const PointF& fromPos, const PointF& toPos, DragM
...
-        QVector<LineF> elAnchorLines = e->dragAnchorLines();
-        const Ms::EngravingItem* page = e->findAncestor(ElementType::PAGE);
-        const PointF pageOffset((page ? page : e)->pos());
-
-        if (!elAnchorLines.isEmpty()) {
-            for (LineF& l : elAnchorLines) {
-                l.translate(pageOffset);
-                anchorLines.push_back(l);
+    if (isGripEditStarted()) {
+        updateAnchorLines();
```
wizofaus, line 775: this was definitely wrong, was causing anchor lines to disappear once you started dragging grip handles

src/notation/internal/notationinteraction.cpp
```diff
@@ -752,43 +771,32 @@ void NotationInteraction::drag(const PointF& fromPos, const PointF& toPos, DragM
...
+    } else {
+        std::vector<LineF> anchorLines;
+        for (const EngravingItem* e : selection()->elements()) {
+            QVector<LineF> elAnchorLines = e->dragAnchorLines();
+            if (!elAnchorLines.isEmpty()) {
+                for (LineF& l : elAnchorLines) {
+                    anchorLines.push_back(l);
```
wizofaus, line 782: anchor lines are already in canvas coordinates, no need to adjust for page

src/notation/internal/notationinteraction.cpp
```diff
@@ -2182,47 +2436,94 @@ static int findGrip(const QVector<mu::RectF>& grips, const mu::PointF& canvasPos
...
+    QVector<LineF> anchorLines = m_editData.element->gripAnchorLines(anchorLinesGrip);
 
-    EngravingItem* page = m_gripEditData.element->findAncestor(ElementType::PAGE);
-    const PointF pageOffset((page ? page : m_gripEditData.element)->pos());
+    EngravingItem* page = m_editData.element->findAncestor(ElementType::PAGE);
     if (!anchorLines.isEmpty()) {
         for (LineF& line : anchorLines) {
-            line.translate(pageOffset);
-            lines.push_back(line);
+            if (line.p1() != line.p2()) {
```
wizofaus, line 2460: anchor lines already canvas-relative, no need to adjust for page, but occasionally anchor lines are single points, which causes asserts if you try to render them

src/notation/internal/notationinteraction.cpp
```diff
@@ -1896,6 +2009,23 @@ void NotationInteraction::moveSelection(MoveDirection d, MoveSelectionType type)
...
+    if (score()->noteEntryMode()) {
+        score()->inputState().moveInputPos(el);
+    }
+    score()->select(el, SelectType::SINGLE, 0);
+    notifyAboutSelectionChanged();
+}
+
+void NotationInteraction::selectEmptyTrailingMeasure()
+{
+    // TODO, move implementation from Score::move
```
wizofaus, line 2024: I have this ready to push now, was pretty straightforward

src/framework/shortcuts/internal/shortcutsregister.cpp
wizofaus: BTW this can be removed from this PR easily enough, it was just to make it easier for me to test, but in the end it makes more sense to test by defining the shortcuts from MU3 and importing.  I would say the far better solution is to change the Shortcuts dialog to use the actions register to get the list of possible actions, then look up what any current shortcut assignments are using the shortcuts register.  I'm happy to start working on the preferences|shortcuts dialog as a separate PR if needed (there's quite a number of issues with it), that can be done independently of this one, but I've been tending to leave QML-related changes alone as I haven't managed to get qt creator working.

src/notation/internal/notationinteraction.cpp
wizofaus: I ended up swapping those operands just to avoid this...

vpereverzev:
@wizofaus I'd be happy to merge your fixes (1 - 19) via separate PRs (once they've tested) by @DmitryArefiev
Most of them are quite important, so I suggest you to cherry-pick them and open separate PRs.

Speaking about actions - I discourage you to just port them from MU3 since we're trying to establish a new system and revise the number of actions/short-cuts.

wizofaus:
Unfortunately cherry picking isn't really an option because I'd been told all commits needed to be squashed. Some of them may be in the reflog but I'm not sure how feasible it is to extract them. The initial point of this PR was to implement a large number of actions that are necessary to edit and navigate scores efficiently with the keyboard, though admittedly many of them I didn't even know of until systemically going through them. Certainly being able to use, say, PgUp and PgDn to scroll through the score, was something I spent considerable time getting right but those are in fact some of the many actions "scr-prev"/"scr-next" that were in the list.  
And most of the other fixes 1-19 came about as a result of implementing "edit-element" (Alt+shift+E,  also in the score context menu for v3). So can I get a list of exactly which actions you don't want in the PR, and I can take those out?

Jojo-Schmitz:
rebase needed

RomanPudashkin:
@wizofaus could you please rebase your PR? I want to merge it

wizofaus:
Hi @RomanPudashkin, sure, I'll look into it ASAP, not sure how big a job it is, I actually thought this PR was just going to get abandoned!

Tantacrul:
@wizofaus - it has always been on the schedule but we've only got around to it now. :)

RomanPudashkin:
Hi @wizofaus! Any updates on this PR? Do you need some help?

wizofaus:
Still working on it, when I can, there are quite a lot of conflicts that aren't straightforward to resolve!  Hopefully in the next day or so.

wizofaus:
Still chugging through this, it's pretty messy, and once there's a build it will need a fair bit of testing to make sure nothing's been messed up.

wizofaus:
Seems we have a build now, but I downloaded it and tried it - and while I could add notes fine etc. "esc" doesn't work for exiting note input mode, so I'll have to see what's happened there.  Most other stuff seems ok (e.g. the slur fixes etc.), but there's a lot to test!

wizofaus:
So I did a bit more testing - at first I thought i couldn't reproduce it at all with my local build, but it seems it *only* happens if the very first thing you do after starting up the app is to create a new score and start entering notes. In that case "ESC" simply doesn't work to quit note-enter mode. But after clicking on the score it starts working.
I believe it's because this action is only registered for UiCtxNotationFocused, but initially the current context is evaluating as UiCtxNotationOpened.  This is definitely not something I changed in this PR but was something I'd dealt with a long time back not long after I first started working with MuseScore4.
It's not clear to me what the right solution is here - in fact I would expect the navigation controller should consider the "notation" window to be the active section right after opening a score, and hence the current context would be UiCtxNotationFocused.
It even appears this might be something that was changed a few weeks ago - git log is showing it as being something you changed @igorkorsukov (in commit 8e178d334aca5b7d2c8b76dabd7c714ae5af29bd).  But I'm not sure if it had been working correctly before then.

igorkorsukov:
> It's not clear to me what the right solution is here - in fact I would expect the navigation controller should consider the "notation" window to be the active section right after opening a score, and hence the current context would be UiCtxNotationFocused.
It even appears this might be something that was changed a few weeks ago - git log is showing it as being something you changed @igorkorsukov (in commits 8e178d3 and 4d1cf11). But I'm not sure if it had been working correctly before then. Interestingly the m_notationViewFocusedCounter seems only ever go up, but it starts at 0, and never tests as > 0 (necessary for UiCtxNotationFocused to the current context) until the notation view gets a Qt "Focus In" event, which doesn't happen until you click somewhere on it.

At the moment, we did not make it so that the notation was immediately in focus when the notation page was opened (i.e. this is not a broken behavior, it was simply never intended to do it)

@Tantacrul  what do you think? Should the focus immediately "jump" into the score when opening the notation page? (and accordingly, shortcuts that work in the context of the notation focused can be triggered) or not, and the focus should remain where it is now, without jumping anywhere.
For example, if we opened the notation page using the keyboard (we are on the primari buttons panel), should the focus immediately jump to the notation? And the next press of the Left key will select the next score element, and not the next primari button.

Jojo-Schmitz:
rebase needed

wizofaus:
Either way, I've confirmed it's not a problem with this PR so probably should be raised as a separate issue.
And yes, I'm aware it needs rebasing, I've been rebasing almost daily for the last week, people keep changing stuff...

vpereverzev:
@wizofaus could you fix the build of this branch? @DmitryArefiev will test it once it ready and we'll be ready to merge this.

DmitryArefiev:
Hi @wizofaus , I've noticed your changes about shortcuts and other improvements. That's very useful features. I like that Repeat (R) command is back. Thanks!

But also found some other issues (mostly tested on Win and Mac) 

- Edit lyrics **doesn't work as it should** (using left/right doesn't move to the next syllable)

https://user-images.githubusercontent.com/90187801/144099553-61ad3b0d-bab2-4af0-8b4a-6d34ee3c65da.mp4

Next I will go through the list of fixes in the description:

2. keyboard support for modifying barline extent (up/down cursor keys) - **DOESN'T WORK**
5. fixed bug dragging beams with mouse - **STILL OCCURS**
6. added keyboard support to adjusting bracket span (up/down cursor keys) - **DOESN'T WORK**
7. changed "full-screen" shortcut key to be F11 (Ctrl+U already meant 'underline') - **it still works as in previous builds** and not like in MU3 (in MU3, Ctr-U shortcut works in two modes: 1) when editing text, it works as underline option 2) when in general mode it enters full screen)
9. ensured that inspector/properties panel updates correctly even if selection hasn't changed (e.g. use 'v' key while an element is selected to toggle hide/show flag) - **DOESN'T WORK** (tested on 'v' key)
10. ensure "select similar elements in range" works - **this not implemented in MU4** (not added to right click menu and absent in shortcuts list in Preferences)
14. ensures ALL elements can be dragged in edit mode - please clarify how it should work (sorry, I'm not MU advanced user yet)
15. enter edit mode when double-clicking on any element - it works **not like** in MU3 and not like in previous builds.. Looks like double click works like a triple click, please see video below:


https://user-images.githubusercontent.com/90187801/144102438-0ff455ff-7e35-48f4-9237-d52a45ae1806.mp4


17. edit mode now mostly works as per MU3:
    b) cursor keys in edit mode generally also adjust x/y offset, with CTRL/ALT/SHIFT modifiers acting as per MU3 - **DOESN'T WORK** with Lines (Slurs, Hairpins)
18. various fixes for adjusting grip handles, particularly for slurs, such that a) when mouse is dragged over notehead the notehead is highlighted and the slur "snaps" to it, b) anchor lines drawn as expected - **DOESN'T WORK** as expected.  When slur anchors to the next note, that note becomes selected forever. Please see video below:

 
  https://user-images.githubusercontent.com/90187801/144103539-30400928-7547-4b9a-a592-c6f6f50d1324.mp4

Tantacrul:
When this PR gets merged, it's going to be epic :)

wizofaus:
Thanks for testing @DmitryArefiev 
1) confirmed that left/right doesn't work on lyrics, that's a regression (had been working in previously nightly builds), but not sure yet whether it's due to changes in this PR, but basically it seems arrow keys don't work for a lot of operations where they did previously
2) confirmed that up/down not working with barlines - probably related to 1)
3) dragging beams with mouse working fine for me, what were the exact steps you used?
4) confirmed, suspect it's the same problem as 1/2
5) you may need to delete your *local* copy of shortcuts.xml for this to work (on Windows it's in %USERPROFILE%\AppData\Local\MuseScore\MuseScore4Development)
6) confirmed not working, may have been due to some changes I had to make while rebasing, but not a regression at least, it was already like this!
7) it is implemented, but you have to find "Select: similar in the range" in the preferences dialog and assign a shortcut, NB there was a bug where assigning shortcuts didn't work if you had a search filter in place, not sure if that's been fixed - unfortunately when I just tried now MuseScore crashed on trying to assign a shortcut (so you have to edit shortcuts.xml instead, and even how to do that isn't obvious).  But yes, I would assume it should go in the context menu too.
8) basically double-clicking on an element (or using Alt+Shift+E shortcut) should put it in edit mode, previously most element types couldn't be dragged.  But I also notice cursor keys are not working on edit mode, probably the same problem as 1/2/4 above.
9) double-clicking does seem to correctly set edit mode at least for notes and such where it works in MU3, the problem is there's no feedback that you're in "edit mode". The only way you can tell is checking, e.g. the tool menu to see if all commands are disabled, and dragging elements around which behaves quite differently between edit and non-edit mode. But note that I wrote up a long proposal about getting rid of edit mode (except for text elements), technically it's not really needed but it would require quite a few design decisions, and possibly scrapping certain functionality.

The main issues seem to be around cursor keys not working, so I'll check into that when I have time.

oktophonie:
3. may be related to #9792 (@Nick-Mazuk ?) but will be fixed separately, shortly

wizofaus:
It may be already reported but I couldn't find it so I logged a separate bug that trying to assign any shortcut a different key is crashing currently, again, unrelated to this PR, which makes some of the changes here hard to test, but I did managed to edit my shortcuts.xml to include

  <SC>
    <key>select-similar-range</key>
    <seq>Ctrl+Shift+R</seq>
    </SC>

And then confirm that Ctrl+Shift+R works for 'select similar in range', but the usage is a bit odd - you have to first select the range then *right* click on the element you want to select similar elements for - the context menu appears but then if you press Ctrl+Shift+R it will activate the "select similar elements in range" command as expected.  But there needs to be a separate issue raised to get it added to the context menu (as having it enable/disable as expected is a bit tricky).

So far though I have fixed the arrow keys not working for barlines/brackets/edit mode for notes/rests etc., not sure what caused that.
Left/right not working for navigating between lyrics was my fault, a bug I'd introduced with this PR, should be fixed with next push.
Updating the inspector when changing properties via keyboard shortcuts (e.g. v for visible) should be working too.

DmitryArefiev:
Hi @wizofaus , 

- Found a **crash** in your new PR build when **entering lyrics and adding "-"** symbol after syllable 

https://user-images.githubusercontent.com/90187801/144442625-aff21046-09a4-41e1-88e3-b29dc5a49112.mp4


- **Something wrong with Palettes** in your new build

https://user-images.githubusercontent.com/90187801/144447070-8773b0d3-98c9-4a7e-849f-3105d064c2ac.mp4



Other (previous) issues:

> 3.  dragging beams with mouse working fine for me, what were the exact steps you used?

https://user-images.githubusercontent.com/90187801/144443039-6fa73e46-3a94-4f9b-8e2c-2cc8f6dd92ee.mp4

If it's related to #9792 then I'll just log a new issue for that after PR is merged.

- **Issue about double click text** is still there. 
Double click should work as Alt+Shift+E shortcut: just enter Edit text mode and not select the text - but I can create a separate bug for that.

- Also, I have some doubts **about using F11 shortcut for Full Screen** command. On Mac you should use Fn modifier to trigger F11.
And MU3 has Ctrl+U shortcut. There is no problem for using Ctrl+U for underline text (when in edit mode) and using it for Full Screen when not in text edit mode. But this can discussed later with designer team and not a big problem.

> Also confirm the (cosmetic) issue with the notehead staying highlighted after adjusting a slur by dragging it onto a notehead. Will look at that when I can, it de-highlights correctly if you keep dragging so should be easy to fix. Not a regression though, as this "snap" functionality wasn't working previously.

I will create a separate issue for that.

Other issues with arrow keys are working fine now! Thanks!

wizofaus:
I can't see any way that palettes issue would be related to this PR but I'll investigate

wizofaus:
So I went back and built from master and:
a) the palette problem is still there, so presumably in the nightly build too, and unrelated to this PR
b) likewise the crash on hyphen when editing lyrics 
c) dragging beams not supported on master, but I notice not even "Force horizontal" or editing the beam height in the inspector/properties panel works, presumably for the same reason

The select-on-double-click-text doesn't happen though, so that seems to be a bug introduced in my PR.

DmitryArefiev:
> a) the palette problem is still there, so presumably in the nightly build too, and unrelated to this PR

You totally right! Can reproduce on the latest master build. Logged.

I'll check your new build today. Thank you for quick fixing!

DmitryArefiev:
@wizofaus The issue with mouse double click is fixed and the crash on hyphen no longer occurs in your build. Thanks! 
Let's promote it. Other issues I'll log separately.

DmitryArefiev:
Tested on Win10, Mac12 - **PASS**     

@RomanPudashkin  Please merge into master

vpereverzev:
Ta-daa :tada: 

@wizofaus Please, next time try to separate fixes - 1 issue = 1 PR. Otherwise, it takes so much time to review and to test it properly

Tantacrul:
> Tested on Win10, Mac12 - **PASS**

This is a momentous occasion 👯. Nice ending to the week!

wizofaus:
> Ta-daa 🎉
> 
> @wizofaus Please, next time try to separate fixes - 1 issue = 1 PR. Otherwise, it takes so much time to review and to test it properly

Absolutely, it definitely should never have grown that big.  But there were a lot of surprising interactions between many of the bugs that were addressed, and switching between branches is pretty painful still (takes forever to recompile etc.).  Anyway, I'm glad all that work didn't all go to waste!

src/engraving/libmscore/barline.cpp
```diff
@@ -1056,6 +1056,23 @@ void BarLine::startEdit(EditData& ed)
     ed.addData(bed);
 }
 
+bool BarLine::edit(EditData& ed)
+{
+    bool local = ed.control() || segment()->isBarLineType() || spanStaff() != score()->staff(staffIdx())->barLineSpan();
+    if (ed.key == Qt::Key_Up && spanStaff() || ed.key == Qt::Key_Down && !spanStaff()) {
```
Jojo-Schmitz, line 1062: Causes a compiler warning in MinGW, suggest to add some parens:
```
    if ((ed.key == Qt::Key_Up && spanStaff()) || (ed.key == Qt::Key_Down && !spanStaff())) {
```

src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -269,9 +270,15 @@ Notification ShortcutsRegister::shortcutsChanged() const
...
+    if (shortCut.action.empty()) {
+        Shortcut sc = { actionCode };
```
Jojo-Schmitz, line 277: Casues a compiler warning in MinGW reg. an uninitialized member `sequence`.

bkunda:
This is indeed a momentous occasion! 🎉 Bravo @wizofaus, @vpereverzev and our terrific development team! I'll look forward to testing all of these key commands in the coming days 😁

PR #8921, opened 2021-08-24
[MU4] Fix compiler warnings

Jojo-Schmitz:
One more remaining:
```
Warning	C4505	'checkNotifySignalValidity_mu__notation__NoteInputBarCustomiseModel': unreferenced function with internal linkage has been removed (compiling source file ...\src\notation\notation_autogen\mocs_compilation_RelWithDebInfo.cpp)
```
Last time we had a similar one, but in thirdparty and there just disabled it, see bee94112, here however it is 'our' code.

### DIFF
--- a/src/appshell/view/dockwindow/docktypes.h
+++ b/src/appshell/view/dockwindow/docktypes.h
@@ -27,8 +27,8 @@
 #include <QVariant>
 
 namespace mu::dock {
-static const char* CONTEXT_MENU_MODEL_PROPERTY("contextMenuModel");
-static const char* DOCK_PANEL_PROPERY("dockPanel");
+inline const char* CONTEXT_MENU_MODEL_PROPERTY("contextMenuModel");
+inline const char* DOCK_PANEL_PROPERY("dockPanel");
 
 enum class DockType {
     Undefined = -1,
--- a/src/engraving/libmscore/score.h
+++ b/src/engraving/libmscore/score.h
@@ -55,8 +55,6 @@ class QMimeData;
 
 namespace mu::engraving {
 class EngravingProject;
-class LayoutContext;
-class LayoutMeasure;
 class AccessibleScore;
 }
 
@@ -428,8 +426,6 @@ private:
     friend class mu::engraving::compat::ReadScoreHook;
     friend class mu::engraving::compat::Read302;
     friend class mu::engraving::Layout;
-    friend class mu::engraving::LayoutContext;
-    friend class mu::engraving::LayoutMeasure;
 
     static std::set<Score*> validScores;
     int _linkId { 0 };
--- a/src/framework/accessibility/internal/accessibilitycontroller.cpp
+++ b/src/framework/accessibility/internal/accessibilitycontroller.cpp
@@ -154,6 +154,8 @@ void AccessibilityController::propertyChanged(IAccessible* item, IAccessible::Pr
         break;
     case IAccessible::Property::Name: etype = QAccessible::NameChanged;
         break;
+    case IAccessible::Property::Description: etype = QAccessible::DescriptionChanged;
+        break;
     }
 
     QAccessibleEvent ev(it.object, etype);
--- a/src/notation/view/noteinputbarcustomisemodel.cpp
+++ b/src/notation/view/noteinputbarcustomisemodel.cpp
@@ -71,10 +71,10 @@ QVariant NoteInputBarCustomiseModel::data(const QModelIndex& index, int role) co
 
     switch (role) {
     case ItemRole: return QVariant::fromValue(item);
-    default: return SelectableItemListModel::data(index, role);
+    default: break;
     }
 
-    return QVariant();
+    return SelectableItemListModel::data(index, role);
 }
 
 QHash<int, QByteArray> NoteInputBarCustomiseModel::roleNames() const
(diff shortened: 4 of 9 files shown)


src/notation/view/noteinputbarcustomisemodel.cpp
```diff
@@ -73,8 +73,6 @@ QVariant NoteInputBarCustomiseModel::data(const QModelIndex& index, int role) co
     case ItemRole: return QVariant::fromValue(item);
     default: return SelectableItemListModel::data(index, role);
     }
-
-    return QVariant();
```
cbjeukendrup, line 77: I believe some compilers (gcc?) would warn that the function does not always have a return value (which is nonsense). The solution to make them all happy would be:
```cpp
    default: break;
    }

    return SelectableItemListModel::data(index, role);
}
```

Jojo-Schmitz: Here MSVC complained about statement without effect

Jojo-Schmitz: But your proposal works too, so let's use that.

cbjeukendrup: I see, but that would also be fixed with my proposed solution, without causing a (bogus) warning on Gcc.

src/engraving/libmscore/score.h
```diff
@@ -428,7 +428,7 @@ class Score : public QObject, public ScoreElement
     friend class mu::engraving::compat::ReadScoreHook;
     friend class mu::engraving::compat::Read302;
     friend class mu::engraving::Layout;
-    friend class mu::engraving::LayoutContext;
+    friend struct mu::engraving::LayoutContext;
```
cbjeukendrup, line 431: It seems that both LayoutContext and LayoutMeasure can be removed completely from this file (also the forward declarations at the top).

Jojo-Schmitz: the forware declaration of `class EngravingProject` too...

src/framework/accessibility/internal/accessibilitycontroller.cpp
```diff
@@ -154,6 +154,8 @@ void AccessibilityController::propertyChanged(IAccessible* item, IAccessible::Pr
         break;
     case IAccessible::Property::Name: etype = QAccessible::NameChanged;
         break;
+    case IAccessible::Property::Description: etype = QAccessible::DescriptionChanged;
+        break;
```
Jojo-Schmitz, line 158: @shoogle: is that something you just forgot to add?

src/appshell/view/dockwindow/docktypes.h
```diff
@@ -27,8 +27,8 @@
 #include <QVariant>
 
 namespace mu::dock {
-static const char* CONTEXT_MENU_MODEL_PROPERTY("contextMenuModel");
-static const char* DOCK_PANEL_PROPERY("dockPanel");
+inline const char* CONTEXT_MENU_MODEL_PROPERTY("contextMenuModel");
+inline const char* DOCK_PANEL_PROPERY("dockPanel");
```
Jojo-Schmitz, line 31: making use of a recently leared C++17 feature ;-)

cbjeukendrup: In this case, it could even be `constexpr` (which implies `inline`)

Jojo-Schmitz: Nope: "ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]"
                                                               ^

Jojo-Schmitz: Or `constexpr` instead of `inline` and not instead of `nline const`?

cbjeukendrup: Yes, I meant `constexpr const char*`. I believe `constexpr` implies const for the pointer, so you need another `const` for the char it points to.

src/framework/accessibility/internal/accessibilitycontroller.cpp
shoogle: Probably. In #8912 I just did the minimum necessary to get it to compile and run. I'll try to pay more attention to warnings next time as well as errors.

Jojo-Schmitz: This wasn't a complaint, more the request for an official blessing of that change ;-)

src/appshell/view/dockwindow/docktypes.h
Jojo-Schmitz: the ìnline` looks better to me than a `constexpr const`...

igorkorsukov: Just for fun, wouldn't inline create new copies of the string and thus use up more memory? Instead of using the same string.

Jojo-Schmitz: Not as var as I read the C++17 standard

Jojo-Schmitz: See <also https://github.com/musescore/MuseScore/pull/8897#pullrequestreview-735821234 which uses the same trick

Jojo-Schmitz: From [C++17 N4659 standard draft](https://github.com/cplusplus/draft/blob/master/papers/n4659.pdf), 10.1.6 "The inline specifier":
```
6 An inline function or variable with external linkage shall have the same address in all translation units.
```

igorkorsukov (approved):

PR #8928, opened 2021-08-25
Small cleanup in Libmscore

cbjeukendrup:
- Cleanup some #includes and forward declarations
- Use Color::black and Color::transparent etc. throughout the project

I would like to propose that we keep `#include`s sorted alphabetically per group (groups like Qt, libmscore, non-libmscore, log.h, etc.). It is a very small effort to put a new include at the correct line, and this makes maintenance and cleanup much easier. I even found a case where one header was included multiple times in one cpp file.


src/engraving/libmscore/textlinebase.h
```diff
@@ -76,16 +76,20 @@ enum class HookType : char {
     NONE, HOOK_90, HOOK_45, HOOK_90T
 };
 
+//---------------------------------------------------------
+//   PlaceText
+//---------------------------------------------------------
+
+enum class PlaceText : char {
```
Jojo-Schmitz, line 83: I wonder whether we'd need a `Q_DECLARE_METATYPE(Ms::PlaceText);` too?

src/engraving/libmscore/scorefile.cpp
```diff
@@ -395,13 +395,13 @@ std::shared_ptr<mu::draw::Pixmap> Score::createThumbnail()
 
     Page* page = pages().at(0);
     RectF fr  = page->abbox();
```
Jojo-Schmitz, line 397: if reducing whitespace, then here too

src/engraving/draw/color.cpp
```diff
@@ -18,7 +18,7 @@ static void insertHexComponent(int num, std::stringstream& ss);
 
 const Color Color::black { 0, 0, 0 };
 const Color Color::white { 255, 255, 255 };
-const Color Color::transparent { 0, 0, 0, 0 };
+const Color Color::transparent { 255, 255, 255, 0 };
```
Jojo-Schmitz, line 21: It seems this change, from transparent black to transparente white, breaks the mtests?

Jojo-Schmitz: See https://github.com/musescore/MuseScore/pull/8928/checks?check_run_id=3420975345#step:8:787

Jojo-Schmitz: strangly only for tuplets though?

Jojo-Schmitz: No, it seems to nt the the `Color::black` lacking the trailing `, 255`? Still, why only for tuplets?
It now wants `<color r="0" g="0" b="0" a="255"/>` being set explicitly.

cbjeukendrup: But even if I add it, the tests are still failing...

Jojo-Schmitz: Yeah, something very strange is goind on there. if it were the missinf 255, it should fail at much more places than just the tuplets

RomanPudashkin (approved):

PR #8942, opened 2021-08-26
Accessibility: Speech for all toolbar buttons

shoogle:
Give every toolbar button an accessible name and description for use with screen readers. By default, the name and description matches the tooltip title and description, but they can be different if required for a specific button.

Tooltips for accidentals in the Note Input toolbar have been updated to use words (e.g. 'sharp', 'flat', etc.) instead of the relevant
Unicode symbol (♯, ♭). This is necessary to enable screen readers to say the name correctly. It is also an improvement for sighted users who might not know the correct name for the symbol. The symbol itself is displayed as the button icon in the toolbar so there is no need to also show it in the tooltip.

### Before

image

### After

image


src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
```diff
@@ -84,7 +84,8 @@ FocusScope {
         enabled: root.enabled && root.visible
 
         accessible.role: MUAccessible.Button
-        accessible.name: root.text
+        accessible.name: root.toolTipTitle
```
Eism, line 87: it won't work for buttons that don't have a tooltip like OK/Cancel, etc.
I think we need to keep the master version
accessible.name: Boolean(root.text) ? root.text : root.toolTipTitle

shoogle: Good idea, thanks! I see you implemented it already.

Eism (approved):

Issue #8370, opened 2021-06-21
[MU4 Task] [Accessibility] Top-menu


vpereverzev:
**Task description**
Replicate an existing approach on top-menu popups

![2021-06-21 13_49_36-Screen Reader - Google Таблицы — Mozilla Firefox](https://user-images.githubusercontent.com/71372310/122750725-8c429300-d297-11eb-9a4a-9684d10b18ae.jpg)

See example in 37eaaec3789f0c9ec0b95d817a0634fc04c4a077

cbjeukendrup:
I think we should just create a generalized approach for StyledMenu. (see also #8373)

vpereverzev:
Yep, you're right from the implementation POV, from the management/testing point of view it's much more convenient to separate such things

Issue #8377, opened 2021-06-21
[MU4 Task] [Accessibility] Dock-panels - "three dots" popup


vpereverzev:
**Task description**
Replicate an existing approach on the "three dots" menu in dock-panels

![2021-06-21 13_31_41-Screen Reader - Google Таблицы — Mozilla Firefox](https://user-images.githubusercontent.com/71372310/122748607-12110f00-d295-11eb-87a1-32a3d61f3efb.jpg)

See example in 37eaaec3789f0c9ec0b95d817a0634fc04c4a077

Issue #8376, opened 2021-06-21
[MU4 Task] [Accessibility] Page scale popup


vpereverzev:
**Task description**
Replicate an existing approach on the "page scale" popup

![2021-06-21 13_33_15-Screen Reader - Google Таблицы — Mozilla Firefox](https://user-images.githubusercontent.com/71372310/122748809-48e72500-d295-11eb-94c8-5c13a32d69a9.jpg)

See example in 37eaaec3789f0c9ec0b95d817a0634fc04c4a077

Issue #8375, opened 2021-06-21
[MU4 Task] [Accessibility] Page view popup


vpereverzev:
**Task description**
Replicate an existing approach on the "page view" popup

![2021-06-21 13_34_12-Screen Reader - Google Таблицы — Mozilla Firefox](https://user-images.githubusercontent.com/71372310/122748920-66b48a00-d295-11eb-9802-ed48616a2818.jpg)

See example in 37eaaec3789f0c9ec0b95d817a0634fc04c4a077

PR #8950, opened 2021-08-27
[MU4] Accessibility for menu

Eism:
Resolves: 
#8370
#8373
#8375
#8376
#8377

### DIFF
--- a/src/appshell/qml/AppMenuBar.qml
+++ b/src/appshell/qml/AppMenuBar.qml
@@ -26,12 +26,15 @@ import MuseScore.UiComponents 1.0
 import MuseScore.AppShell 1.0
 
 ListView {
+    id: root
 
     property alias navigation: navPanel
 
     height: contentItem.childrenRect.height
     width: contentWidth
     orientation: Qt.Horizontal
+
+    interactive: false
 
     model: appMenuModel.items
 
@@ -52,6 +55,8 @@ ListView {
     NavigationPanel {
         id: navPanel
         name: "AppMenuBar"
+        enabled: root.enabled && root.visible
+        accessible.name: qsTrc("appshell", "App menu bar") + " " + navPanel.directionInfo
     }
 
     delegate: FlatButton {
@@ -65,8 +70,12 @@ ListView {
         hoveredStateColor: ui.theme.accentColor
         text: Boolean(item) ? item.title : ""
 
+        navigation.name: text
         navigation.panel: navPanel
         navigation.order: index
+        navigation.accessible.name: {
+            return text.replace('&', '')
+        }
 
         mouseArea.onContainsMouseChanged: {
             if (!mouseArea.containsMouse || !prv.showedMenu ||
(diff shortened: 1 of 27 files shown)


src/appshell/qml/AppMenuBar.qml
```diff
@@ -52,6 +53,8 @@ ListView {
     NavigationPanel {
         id: navPanel
         name: "AppMenuBar"
+        enabled: root.enabled && root.visible
+        accessible.name: qsTrc("appshell", "App menu bar") + " " + navPanel.directionInfo
```
shoogle, line 57: Including `navPanel.directionInfo` in the name is helpful at first but I think blind users will quickly get tired of hearing _"direction is horizontal"_ or _"direction is vertical"_ all the time. Instead, I think we should adopt the following convention:

- Toolbars and menu bars are always __horizontal__.
- Item views (e.g. tables and trees) are always __both directions__ unless there is only one column of data in the model.
- Everything else (e.g. menu items, lists, single-column tables and trees) are always __vertical__.

This should make it unnecessary to say the direction explicitly in most cases. If there are specific places where we need to break the convention then we can always say it in those places only.

@igorkorsukov, what do you think?

src/appshell/qml/AppTitleBar.qml
```diff
@@ -82,10 +88,23 @@ Rectangle {
             anchors.right: parent.right
             anchors.verticalCenter: parent.verticalCenter
 
+            NavigationPanel {
+                id: navAppControlPanel
+                name: "AppControl"
+                section: navSec
+                order: 1
+                accessible.name: qsTrc("appshell", "App control") + " " + navAppControlPanel.directionInfo
```
shoogle, line 96: Same

shoogle (commented):
Thanks for looking at this! It works for me on Windows (Narrator, not NVDA) and Linux (Orca).

Like all accessibility in MU4, the menus are not currently working on Windows with NVDA or macOS with VoiceOver. I have a few ideas for how we might fix that which I will try on another PR. It should be good to merge this one in the meantime.

Key navigation of the menus is working for me on all platforms; it's just the screen reader speech that works on some but not others.

src/appshell/qml/AppMenuBar.qml
igorkorsukov: Yes, I agree with that.
During development, I added direction information for demonstration and debugging purposes only.
Now we need to think and solve all the moments as they should be in production.

vshalkevich:
Works good!

src/appshell/qml/AppMenuBar.qml
Eism: thanks, removed

shoogle (approved):

RomanPudashkin (approved):

DmitryArefiev:
Checked #8370 #8373 #8375 #8376 #8377with Narrator on Win10 and Orca on Linux Mint 20.2 - FIXED

Accessibility doesn't work properly with NVDA (Win) and VoiceOver (Mac) - logged separately as #9155 , #9153
The issue with Orca logged as #9209

Issue #8952, opened 2021-08-27
[MU4 Issue] [Engraving] Note stems don't take the SMuFL anchors "stemUpNW" and "stemDownSW" into account


cbjeukendrup:
**Describe the bug**
The stems of notes and chords don't take the [`stemUpNW` and `stemDownSW` SMuFL anchors](https://w3c.github.io/smufl/latest/specification/glyph-registration-notes-flags.html) of the flag into account. This causes that the stem has an incorrect length. 
For normal flags, this is not very visible (but if you look closely, you might see an edge at some zoom levels when anti-aliasing is turned on). The problem becomes very obvious with straight flags*:

![Straight flags wrong stem length](https://user-images.githubusercontent.com/48658420/131194521-6775afde-b899-417e-b135-fd7f13356b6a.png)

\* Not yet possible, but I will implement this as soon as this issue is fixed and PR #8867 gets merged.

**Additional context**
For the Shadow Note, it already works correctly because I implemented this in PR #7316.

oktophonie:
Be my guest ;)

If you're in that bit of the code, is it worth investigating a fix for `stemUpSE` and `stemDownNW` on certain noteheads like the cross and triangle? They seem _almost_ right, but not quite:
![x-noteheads](https://user-images.githubusercontent.com/20604032/131210350-c4f1431d-3246-412b-8ab0-1af5fe62bdb0.png)

Jojo-Schmitz:
That 'edge' on normal flags:
![image](https://user-images.githubusercontent.com/1786669/131214554-bda188a6-ff44-4ef0-a5d0-8ca4016717dd.png)

DmitryArefiev:
Hi @cbjeukendrup , can you help with creating straight flags please? I'm new to MU and don't find a way how to create straight flags..

cbjeukendrup:
The problem is that there is no way to enable straight flags yet. I can implement that when PR #8867 is merged. (The screenshot at the top of this issue was taken from a development build that was made on top of that PR.) 

For testing PR #9237, you could also look at the position of the grip box to make the stem longer or shorter; that one always appears at the end of the stem line. 
<img width="585" alt="Schermafbeelding 2021-09-23 om 16 36 54" src="https://user-images.githubusercontent.com/48658420/134527634-9fac48ae-1d5c-4d23-b3c5-e26a61ad76f5.png">

DmitryArefiev:
> 
> 
[...]

PR #8853, opened 2021-08-13
[MU4] Fix #316152: Some characters do not print from Leland on certain printer drivers on Mac

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/316152

Fix got confirmed by several users (for a development build based on 3.x)

This it to master what #8578 is to 3.x

vtest errors are to be expected and got @oktophonie's [blessing](https://github.com/musescore/MuseScore/pull/8578#issuecomment-876370990) ;-)
Edit: interesting, no vtest failures? This is rather unexpected! See #8977...

PR #8958, opened 2021-08-29
Fixed SMuFL anchors

cbjeukendrup:
Resolves: 
- https://github.com/musescore/MuseScore/issues/8952#issuecomment-907587882 (only @oktophonie's comment, not yet the whole issue)
- https://trello.com/c/86Ks7zZX/58-stemdownnw-stemupse-not-honoured-for-x-notehead

While working on #8952, I wrote some extra drawing code for debugging bboxes, positioning, anchors etc.. This revealed that the SMuFL anchors were hardly ever in the correct place, so stems were not attached correctly to noteheads. They always seemed to be at 80% of their correct position from the origin. And exactly that was the case. Now it looks like this: 
Schermafbeelding 2021-08-29 om 18 02 08
(red `+` means origin of element; orange `+` is flag anchor point; yellow `+` is notehead anchor point; the green rect is the bounding box. Please don't mind the badly-placed and purple-coloured stems in this picture.)

Here is it without those extra things:
Schermafbeelding 2021-08-29 om 18 38 26

It also turned out that the metadata file for Petaluma was not added to the application, so for Petaluma the anchors did not work at all. That's fixed too with this PR.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importxmlfirstpass.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importxmlfirstpass.cpp
@@ -22,6 +22,8 @@
 
 #include "importxmlfirstpass.h"
 
+#include "log.h"
+
 namespace Ms {
 // TODO: move somewhere else
 
--- a/src/engraving/engravingmodule.h
+++ b/src/engraving/engravingmodule.h
@@ -35,7 +35,7 @@ public:
     void registerResources() override;
     void registerUiTypes() override;
     void onInit(const framework::IApplication::RunMode& mode) override;
-    void onDeinit();
+    void onDeinit() override;
 };
 }
 
--- a/src/engraving/libmscore/mscoreview.cpp
+++ b/src/engraving/libmscore/mscoreview.cpp
@@ -62,7 +62,7 @@ Element* MuseScoreView::elementAt(const mu::PointF& p)
 
 Page* MuseScoreView::point2page(const mu::PointF& p)
 {
-    if (score()->layoutMode() == LayoutMode::LINE) {
+    if (score()->layoutMode() == mu::engraving::LayoutMode::LINE) {
         return score()->pages().isEmpty() ? 0 : score()->pages().front();
     }
     foreach (Page* page, score()->pages()) {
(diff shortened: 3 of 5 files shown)


Jojo-Schmitz:
What puzzels me is that there are no vtest differences repoted? There should be some minor one (but for the better)

cbjeukendrup:
> What puzzles me is that there are no vtest differences reported? There should be some minor ones (but for the better)

That makes me a little worried too, also because I remember a PR of yours for which we also expected vtest differences but there wasn't. Makes me wonder whether the vtests are working at all...

Jojo-Schmitz:
Yes, #8853

cbjeukendrup:
It probably starts going wrong because the vtest input files are still old mscx files. It is unclear to me whether any output images are generated at all.

Then the compare step silently fails:
```
Run ./vtest/vtest.sh -c compare
MuseScore VTest
RUN_CMD: compare
SCORES_DIR: ./vtest/scores
ARTIFACTS_DIR: vtest.artifacts
MSCORE_BIN: build.debug/install/bin/mscore
DPI: 130
Compare PNG files and references
ls: cannot access 'vtest.artifacts/ref/*.ref.png': No such file or directory
```

Will try to fix it tomorrow.

Jojo-Schmitz:
It is a separate issue though, not the business of this PR here. Which is good as it is, and should get merged sooner rather than later ;-)

oktophonie (approved):

src/importexport/musicxml/internal/musicxml/importxmlfirstpass.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include "importxmlfirstpass.h"
 
+#include "log.h"
```
vpereverzev, line 25: is it really in use here?

vpereverzev (approved):

src/importexport/musicxml/internal/musicxml/importxmlfirstpass.cpp
cbjeukendrup: Yes, `IF_ASSERT_FAILED()`, in line 176.

PR #8966, opened 2021-08-30
Changed type virtual method to type field

igorkorsukov:
This is made for two reasons:
1. A virtual method cannot be called in the constructor and destructor, sometimes it is necessary (directly or indirectly)
2. It is difficult to debug with a virtual method in the debugger - the type of the element is not visible. With the type field, we see the type of the element in the debugger.

### DIFF
--- a/src/engraving/libmscore/box.cpp
+++ b/src/engraving/libmscore/box.cpp
@@ -58,8 +58,8 @@ static const ElementStyle hBoxStyle {
 //   Box
 //---------------------------------------------------------
 
-Box::Box(Score* score)
-    : MeasureBase(score)
+Box::Box(const ElementType& type, Score* score)
+    : MeasureBase(type, score)
 {
 }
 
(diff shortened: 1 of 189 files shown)


vpereverzev (approved):

src/engraving/libmscore/box.cpp
```diff
@@ -58,8 +58,8 @@ static const ElementStyle hBoxStyle {
 //   Box
 //---------------------------------------------------------
 
-Box::Box(Score* score)
-    : MeasureBase(score)
+Box::Box(const ElementType& type, Score* score)
```
wizofaus, line 61: is there a reason to bother passing an enum as a reference? (other than the possibility of it being changed to an expensive-to-copy type in the future, but that seems unlikely...)

igorkorsukov: there is no particular reason, just a habit, I think this is not good and not bad.

PR #8976, opened 2021-08-31
Gp6 gp7 reading

handrok:
Refactoring of reading files of Guitar Pro 6 and 7 versions.


src/importexport/guitarpro/internal/importgtp.cpp
```diff
@@ -1332,6 +1332,34 @@ bool GuitarPro1::read(QFile* fp)
     return true;
 }
 
+int GuitarPro::harmonicOvertone(Note* note, float harmonicValue, int harmonicType)
+{
+    int result{ 0 };
+
+    if (harmonicValue == 12.0f) {
```
Jojo-Schmitz, line 1339: Checking floating point numbers for equality is asking for trouble. Some sort of epsilon is needed or the corresponding method Qt provides for this: [qFuzzyCompare()](https://doc.qt.io/qt-5/qtglobal.html#qFuzzyCompare)

handrok: Fixed. I have founded mu::RealIsEqual utilite in global module

Jojo-Schmitz: That probably does use qFuzzyCompare under the covers?

handrok: mu::RealIsEqual use its own realisation. But it is the same actually.

Jojo-Schmitz: I see. Not sure though why reinventing the wheel there rather than taking an existing method of a library we use anyway.

Jojo-Schmitz: and qFuzzyCompare is used in a lot of other places in the source

src/importexport/guitarpro/internal/importgtp.cpp
```diff
@@ -30,6 +30,7 @@
 #include "importptb.h"
 
 #include "engraving/infrastructure/io/xml.h"
+#include "realfn.h"
```
Jojo-Schmitz, line 33: this is now no longer neeeded

handrok: sure

Jojo-Schmitz:
I'm puzzled about what the refactoring is, I see tons of new code, but so far no old code being replaced by it?

handrok:
I'm not ready to delete old code now. Old code leads to crash a lot of gp files. New code leads to the same crashes. To fix this crashes old code can be useful for me. There is a easy way to switch between old code and new code. I have a big dataset of guitar pro files. When most of the crashes have been fixed, I will delete old code.

Jojo-Schmitz:
I see, this apeases my curosity ;-), thanks for explaining

PR #8979, opened 2021-08-31
[MU4] Fix #324232: after a MIDI export without expanding repeats, those are not played in the current session anymore either

Jojo-Schmitz:
Resolves: https://musescore.org/de/node/324232

To reproduce: Have "Play Repeats" enabled, export a score to MIDI, disable "Expand repeats", then play the still open score: it won't repeat, but the Play Repeats button is still active.

Very same fix would apply to 3.x too

### DIFF
--- a/src/engraving/libmscore/masterscore.h
+++ b/src/engraving/libmscore/masterscore.h
@@ -151,6 +151,7 @@ public:
     void setPlaylistClean() { _playlistDirty = false; }
 
     void setExpandRepeats(bool expandRepeats);
+    bool expandRepeats() const { return _expandRepeats; }
     void updateRepeatListTempo();
     const RepeatList& repeatList() const override;
     const RepeatList& repeatList2() const override;
--- a/src/engraving/libmscore/rendermidi.cpp
+++ b/src/engraving/libmscore/rendermidi.cpp
@@ -2468,12 +2468,14 @@ void Score::renderMidi(EventMap* events, const SynthesizerState& synthState)
 
 void Score::renderMidi(EventMap* events, bool metronome, bool expandRepeats, const SynthesizerState& synthState)
 {
+    bool expandRepeatsBackup = masterScore()->expandRepeats();
     masterScore()->setExpandRepeats(expandRepeats);
     MidiRenderer::Context ctx;
     ctx.synthState = synthState;
     ctx.metronome = metronome;
     ctx.renderHarmony = true;
     MidiRenderer(this).renderScore(events, ctx);
+    masterScore()->setExpandRepeats(expandRepeatsBackup);
 }
 
 void MidiRenderer::renderScore(EventMap* events, const Context& ctx)


src/engraving/libmscore/rendermidi.cpp
```diff
@@ -2473,6 +2473,7 @@ void Score::renderMidi(EventMap* events, bool metronome, bool expandRepeats, con
     ctx.metronome = metronome;
     ctx.renderHarmony = true;
     MidiRenderer(this).renderScore(events, ctx);
+    masterScore()->setExpandRepeats(MScore::playRepeats);
```
RomanPudashkin, line 2476: But we already have a code like this above (line 2470)...

Jojo-Schmitz: No, there we set it as per the `expandRepeats` argument of that method, here as per the preferences setting, the one that is reflected by the Play Repeats button

Jojo-Schmitz: No, there we set it as per the `expandRepeats` argument of that method (from Edit > Preferences > Export > MIDI > Expand repeats) , here as per `MScore::playRepeats`, the setting of the Play Repeats button.

RomanPudashkin: The expandRepeats argument should always have the same value as the MScore::playRepeats variable. This method:

`void renderMidi(EventMap* events, bool metronome, bool expandRepeats, const SynthesizerState& synthState);
`

is used in two different places. In the first one:

```
void Score::renderMidi(EventMap* events, const SynthesizerState& synthState)
{
    renderMidi(events, true, MScore::playRepeats, synthState);
}
```

expandRepeats is set explicitly by the MScore::playRepeats value.

In the second one:

```
bool ExportMidi::write(QIODevice* device, bool midiExpandRepeats..)
{
    ...
    m_score->renderMidi(..., midiExpandRepeats, ...);
    ...
}
```

expandRepeats is set by the NotationConfiguration::isPlayRepeatsEnabled() value. The configuration must have the same value as MScore::playRepeats. And this is a bug if they have different values (I would prefer to remove MScore::playRepeats completely but this isn't related to this PR)

Jojo-Schmitz: No, one is a UI setting (the play repeats button in the toolbar), valid for playback, the other a MIDI  export preference setting.
Currentzly theyx interact badly, this PR fixes that and after the MIDI export (which may or may nor have it set) sets it back to the UI setting.

Eism: If we talk only about the problem and about this method, I think you just need to backup the expandRepeats value and restore after rendering
Will this fix the problem?

Jojo-Schmitz: comparing `expandRepeats` with `MScore::playRepeats` before, storing it (if different) before the `renderScore()` and restoring (if it was different) after should work too, but would be more lines of code for no real benefit.

Jojo-Schmitz: Setting it only if different might be a small optimization?
```
if (expandRepeats != MScore::playRepeats) {
        masterScore()->setExpandRepeats(MScore::playRepeats);
}
```

Jojo-Schmitz: Or maybe resetting it in `ExportMidi::write()`, after the `m_score->renderMidi(..., midiExpandRepeats, ...);`?

Jojo-Schmitz: I'd rather reset it in `renderMidi()` though, close to the set.

Eism: For example

    bool expandRepeatsBackup = masterScore()->expandRepeats();
    masterScore()->setExpandRepeats(expandRepeats);
    
    MidiRenderer::Context ctx;
    ctx.synthState = synthState;
    ctx.metronome = metronome;
    ctx.renderHarmony = true;
    MidiRenderer(this).renderScore(events, ctx);
    
    masterScore()->setExpandRepeats(expandRepeatsBackup);
    
If a new developer looks at this method, then he will understand what this method does.
I think Roman was confused by this moment. He didn't understand the relationship between expandRepeats and MScore::playRepeats(neither did I)

Jojo-Schmitz: OK, that's just on extra line, and looks OK to me.

Jojo-Schmitz: But... there is no `expandRepeats()`!

Jojo-Schmitz: Back to my `if` clause above...

Jojo-Schmitz: Maybe adding a comment `// like possible if called from ExportMidi::write()`?

Eism: I think it is better to add a getter for expandRepeats method, we already have the setter

Jojo-Schmitz: OK, done. I liked it better the small and simple way it was before though...

src/engraving/libmscore/masterscore.h
```diff
@@ -151,6 +151,7 @@ class MasterScore : public Score
     void setPlaylistClean() { _playlistDirty = false; }
 
     void setExpandRepeats(bool expandRepeats);
+    bool expandRepeats();
```
RomanPudashkin, line 154: Please, add const here

src/engraving/libmscore/rendermidi.cpp
RomanPudashkin: The new solution looks much better, thanks!

src/engraving/libmscore/masterscore.h
Jojo-Schmitz: done. no override?

RomanPudashkin: There is no need for this. This method isn't virtual

src/engraving/libmscore/rendermidi.cpp
Jojo-Schmitz: Yes, now it is 3 lines, that's bearable ;-)

RomanPudashkin (approved):

Issue #8381, opened 2021-06-21
[MU4 Task] [KEY_NAV] Workspaces dialog


vpereverzev:
**Task description**
Replicate an existing approach of key-navigation on the workspaces dialog

See example in #8025 

See related task #8371

PR #8983, opened 2021-09-01
8381 workspacesdialog

nightlight521:
Implemented: https://github.com/musescore/MuseScore/issues/8381?fbclid=IwAR3aipaI4TbcJUOgVy2HJTIj69DYep-8lgolKYUcesecLkXq3KrOzotKDKg

Summary: Key navigation added for the workspaces dialog, not including the "new workspaces" dialog that opens when you press "create new workspace."

### DIFF
--- a/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
+++ b/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
@@ -19,7 +19,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-import QtQuick 2.12
+import QtQuick 2.15
 import QtQuick.Controls 2.2
 
 import MuseScore.Ui 1.0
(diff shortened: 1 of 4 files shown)


Eism:
@nightlight521 
Navigation for workspaces dialog should have something like this structure(see picture)

Also, I don’t understand what the navigation sequence should be for this dialog
@Tantacrul can you help us?

ksnip_20210901-092806

src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
```diff
@@ -61,6 +70,9 @@ RadioButtonGroup {
         background: FlatRadioButton {
             anchors.fill: parent
 
+            navigation.panel: navPanel
+            keynavRow: model ? model.index : 0
```
Eism, line 74: you can use navigation.row instead of adding a new property(keynavRow)

Eism:
@nightlight521 you can see how I implemented navigation for preferences dialog(finished for menu and bottom panels)
#8984 

PS: you also need to add accessibility for dialog. Also can see how I did it for preferences

Tantacrul:
> Also, I don’t understand what the navigation sequence should be for this dialog
@Tantacrul can you help us?

If it's about tab navigation, I think your image looks about right to me. Perhaps @shoogle could give more perspective?

Eism:
Reimplemented on #9439

PR #8985, opened 2021-09-01
[MU4] Don't have strings in plugins translatable that [dw]on't work anyhow

Jojo-Schmitz:
The plugin description and -path are not shown translated even if they are (like for the helloqml plugin and German).
Showing as being translatable gives a wrong impression as examples for other plugins.
BTW: Those descriptions do translate in the Plugin Editor/Creator's console log, but only there.
While at it, make helloqml actually work, it apparently had never been ported from 2.x to 3.x (and later).

To avoid merge conflicts I've added another commit here: 
Fix [#325019](https://musescore.org/en/node/325019): Fix notenames plugin when adding octaves to chords and while at it make a few changes to it recommended by QtCreator's editor

Same changes would apply to 3.x too


Jojo-Schmitz:
@team Mind to review and merge this one?

PR #8987, opened 2021-09-01
[MU4] mixer_panel_implementation

vpereverzev:
This PR introduces the implementation of the new Mixer Panel in MU4.

It allows you to apply stackable effects (primarly VST FX), playback instruments via VSTi or Fluid(using the soundfonts), adjust volume level and panning.

In order to get an idea of how it works, please check the following demo attachments.

mixer_workflow

https://user-images.githubusercontent.com/57620349/131661709-6d895d10-33fb-497f-b37c-4ec6cf20e4cb.mp4

There is still a couple of remaining tasks and issues to polish (like a fallback mechanism for audio input sources), but they'd be resolved soon in the next PRs (see related tasks in the project - https://github.com/musescore/MuseScore/projects/38


src/framework/vst/vsttypes.h
```diff
@@ -63,14 +63,23 @@ enum class VstPluginType {
...
-    PluginId id;
-    VstPluginType type = VstPluginType::Undefined;
-    std::string name;
-    std::string path;
-};
-
-using VstPluginMetaList = std::vector<VstPluginMeta>;
+/// @see https://steinbergmedia.github.io/vst3_doc/vstinterfaces/namespaceSteinberg_1_1Vst_1_1PlugType.html
+namespace PluginCategory {
+static const std::string Analyzer = "Analyzer";
```
Jojo-Schmitz, line 68: Not translatable?

vpereverzev: it's not meant to be, as it's an internal stuff for filtering VST plugins

Jojo-Schmitz: OK

src/framework/vst/internal/vstmodulesrepository.cpp
```diff
@@ -128,6 +114,45 @@ void VstModulesRepository::addModule(const io::path& path)
     m_modules.emplace(io::basename(path).toStdString(), module);
 }
 
+audio::AudioResourceMetaList VstModulesRepository::modulesMetaList(const VstPluginType& type) const
+{
+    audio::AudioResourceMetaList result;
+
+    static auto hasNativeEditorSupport = []() {
```
RomanPudashkin, line 121: What do you think about transferring this lambda to VstConfiguration?

vpereverzev: Initially I was going to put it here, but I think we can avoid that restriction in the nearest future

src/framework/vst/vsttypes.h
RomanPudashkin, line 68: Can we use constexpr std::string_view here?

src/notation/internal/masternotationmididata.cpp
```diff
@@ -157,22 +160,29 @@ std::vector<Event> MasterNotationMidiData::retrieveSetupEvents(const std::list<I
     return result;
 }
 
-Events MasterNotationMidiData::eventsFromRange(const channel_t midiChannel, const tick_t fromTick, const tick_t toTick) const
+Events MasterNotationMidiData::eventsFromRange(const std::vector<channel_t>& midiChannels, const tick_t fromTick, const tick_t toTick) const
```
RomanPudashkin, line 163: You use std::vector<channel_t> in many places. Wouldn't it be better to create an alias?
MidiChannels = std::vector< channel_t>

src/playback/qml/MuseScore/Playback/internal/MixerBalanceSection.qml
```diff
@@ -0,0 +1,71 @@
...
+
+import QtQuick 2.15
+import MuseScore.Ui 1.0
+import MuseScore.UiComponents 1.0
+import MuseScore.Audio 1.0
+
+MixerPanelSection {
+    id: root
+
+    headerTitle: qsTr("Pan")
```
RomanPudashkin, line 31: You missed a translation context here

src/playback/qml/MuseScore/Playback/internal/MixerFxSection.qml
```diff
@@ -0,0 +1,63 @@
...
+
+import QtQuick 2.15
+import MuseScore.Ui 1.0
+import MuseScore.UiComponents 1.0
+import MuseScore.Audio 1.0
+
+MixerPanelSection {
+    id: root
+
+    headerTitle: qsTr("Audio Fx")
```
RomanPudashkin, line 31: You missed a translation context here

src/playback/qml/MuseScore/Playback/internal/MixerBalanceSection.qml
vpereverzev: thanks, nice catch

src/playback/qml/MuseScore/Playback/internal/MixerSoundSection.qml
```diff
@@ -0,0 +1,52 @@
...
+
+import QtQuick 2.15
+import MuseScore.Ui 1.0
+import MuseScore.UiComponents 1.0
+import MuseScore.Audio 1.0
+
+MixerPanelSection {
+    id: root
+
+    headerTitle: qsTr("Sound")
```
RomanPudashkin, line 31: You missed a translation context here

src/playback/qml/MuseScore/Playback/internal/MixerTitleSection.qml
```diff
@@ -0,0 +1,46 @@
...
+
+import QtQuick 2.15
+import MuseScore.Ui 1.0
+import MuseScore.UiComponents 1.0
+import MuseScore.Audio 1.0
+
+MixerPanelSection {
+    id: root
+
+    headerTitle: qsTr("Name")
```
RomanPudashkin, line 31: You missed a translation context here

src/playback/qml/MuseScore/Playback/internal/MixerVolumeSection.qml
```diff
@@ -0,0 +1,102 @@
...
+
+import QtQuick 2.15
+import MuseScore.Ui 1.0
+import MuseScore.UiComponents 1.0
+import MuseScore.Audio 1.0
+
+MixerPanelSection {
+    id: root
+
+    headerTitle: qsTr("Volume")
```
RomanPudashkin, line 31: You missed a translation context here

src/playback/view/internal/abstractaudioresourceitem.cpp
```diff
@@ -0,0 +1,59 @@
+#include "abstractaudioresourceitem.h"
+
+#include <QList>
+#include <QTimer>
+
+using namespace mu::playback;
+
+//!Note Some resources like VST plugins are not able to work in a couple of msecs
+//!     So we've to add explicit delay before the launching their 'native' editor views
+static const int EXPLICIT_DELAY_MSECS = 3000;
```
RomanPudashkin, line 10: You can use contexpr here

src/playback/qml/MuseScore/Playback/internal/MixerBalanceSection.qml
vpereverzev: it should be qsTrc actually

src/notation/internal/masternotationmididata.cpp
vpereverzev: Yeah, might make sense
I'd fix it in another PR since it's used quite often

src/playback/view/internal/outputresourceitem.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <QList>
+
+#include "log.h"
+#include "translation.h"
+
+using namespace mu::playback;
+using namespace mu::audio;
+
+static const QString NO_FX_MENU_ITEM_ID = QString::fromStdString(mu::trc("playback", "No effect"));
```
RomanPudashkin, line 11: There is a mistake. Translation will not be applied to this string because it is static

src/playback/qml/MuseScore/Playback/internal/MixerBalanceSection.qml
vpereverzev: fixed

src/playback/view/internal/outputresourceitem.cpp
vpereverzev: damn, yeah
probably better to replace it with a getter

Jojo-Schmitz: So a `QT_TRANSLATE_NOOP()` is needed here

src/framework/audio/audiotypes.h
```diff
@@ -53,31 +54,87 @@ using MixerChannelId = int32_t;
...
+
+    bool isValid() const
+    {
+        return !id.empty() && !vendor.empty();
+    }
+
+    bool operator==(const AudioResourceMeta& other) const
+    {
+        return id == other.id
+               && vendor == other.vendor
```
RomanPudashkin, line 80: What about comparing by type?

vpereverzev: I think I've missed that one when have been adding the type field

thanks

src/framework/vst/vsttypes.h
vpereverzev: it's not gonna be copied, but I'm not against
fixed

src/playback/view/internal/abstractaudioresourceitem.cpp
vpereverzev: I'm not against, fixed

RomanPudashkin (approved):

src/playback/view/internal/outputresourceitem.cpp
```diff
@@ -8,7 +8,12 @@
 using namespace mu::playback;
 using namespace mu::audio;
 
-static const QString NO_FX_MENU_ITEM_ID = QString::fromStdString(mu::trc("playback", "No effect"));
+static const QString& NO_FX_MENU_ITEM_ID()
+{
+    static std::string id = mu::trc("playback", "No effect");
```
cbjeukendrup, line 13: I think this still won't work because it's still a static string. So it will get the correct value when it is called for the first time, but it won't be updated until the app is restarted when the user changes the app language.

Also, I think it is better to use `mu::qtrc` directly, because `mu::trc` calls `QCoreApplication::translate()` just like `mu::qtrc` and then converts it to a std::string, and here the std::string is converted back to a QString. So, using `mu::qtrc` here would eliminate two conversions.

PR #8996, opened 2021-09-01
Engraving elements tree refactoring

igorkorsukov:
//! NOTE Before, element tree is made to be done like this
    //! class ScoreElement
    //! {
    //!     Score* _score;
    //!     ScoreElement(Score* score)...
    //! }
    //!
    //! class Element : public ScoreElement
    //! {
    //!    Element* _parent;
    //!    Element(Score* s) : ScoreElement(s)...
    //!
    //!    Element* parent() const { return _parent; }
    //!    void setElement(Element* e) { _parent = e; }
    //! }
    //! accordingly:
    //! * All elements have a ref to score which they are located.
    //! * The base element (ScoreElement) itself has no parent property
    //! * The parent of an element could be set or could not be set, so in general, it is impossible to build a tree of elements
    //! (to solve this problem, a `treeParent` method was added that tries to determine the parent)
    //! * There was also logic for some elements that if a parent is set, then an element in the tree,
    //! if not set, then no, but for other elements, it does not matter.
    //!
    //! Now the element tree is:
    //! class ScoreElement
    //! {
    //!     ScoreElement* m_parent;
    //!     ScoreElement(ScoreElement* parent)...
    //! }
    //! accordingly:
    //! * No more score property (score is searched for in the parent tree)
    //! * All objects must belong to someone
    //!
    //! For compatibility purposes, it has been done so that the new structure has the old behavior.

### DIFF
--- a/src/engraving/libmscore/bsymbol.cpp
+++ b/src/engraving/libmscore/bsymbol.cpp
@@ -39,8 +39,8 @@ namespace Ms {
 //   BSymbol
 //---------------------------------------------------------
 
-BSymbol::BSymbol(const Ms::ElementType& type, Score* s, ElementFlags f)
-    : Element(type, s, f)
+BSymbol::BSymbol(const Ms::ElementType& type, Ms::Element* parent, ElementFlags f)
+    : Element(type, parent, f)
 {
     _align = Align::LEFT | Align::BASELINE;
 }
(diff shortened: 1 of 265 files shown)


vpereverzev (approved):

src/engraving/libmscore/clef.cpp
```diff
@@ -311,7 +311,7 @@ Element* Clef::drop(EditData& data)
             if (segm->element(track())) {
                 score()->undoRemoveElement(segm->element(track()));
             }
-            Ambitus* r = new Ambitus(score());
+            Ambitus* r = new Ambitus(segm);
             r->setParent(segm);
```
cbjeukendrup, line 315: Is this still needed?

src/engraving/libmscore/bsymbol.cpp
```diff
@@ -105,7 +105,10 @@ bool BSymbol::readProperties(XmlReader& e)
 void BSymbol::add(Element* e)
 {
     if (e->isSymbol() || e->isImage()) {
-        e->setParent(this);
+        if (e->parent() != this) {
+            e->setParent(e);
```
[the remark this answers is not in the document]
igorkorsukov: ops, a lot of thanks

src/engraving/libmscore/clef.cpp
cbjeukendrup: Ah, yes, I indeed realized that after seeing the `isParentSetExplicitly` stuff. Thanks!

PR #9001, opened 2021-09-01
fix #8973 Change how tie directions for chords are calculated

### LINKED ISSUE #8973: [MU4 Issue] Improve tie direction in chords containing seconds
As far as I can deduce, MuseScore has two primary rules for determining the direction of ties in chords (when 'Tie direction' = 'Auto'):

1. the uppermost tie curves up, the bottommost curves down, and those in the middle go up or down following the usual rules (if the note is on the middle stave line, or higher, the tie curves up, otherwise it curves down).
2. when the interval of a second is encountered in a chord, the tie to the bottom note always curves down, and the tie to the top note curves up

Ideally, the tie direction should flip only once in the chord; rule 2 causes a problem where the interval of a second appears more than once in the chord, since this means the direction changes more than once.

I recommend the following solution:

- adjust rule 1 to say: for a chord with an odd number of notes, the lower half curve down, the upper half curve up. If there is an odd number, the middle tie curves away from the stem (i.e. curving up if the tie is down, and vice versa); this happens to be the rule recommended by Gould(!)
- UNLESS the chord contains _one_ (and only one) instance of the interval of a second (or a unison), in which case the tie direction changes at that point (the tie to the upper note of the second/unison curves up, as do the ties of any higher notes; the tie to the lower note of the second/unison curves down, as do the ties of any lower notes).

Thus:

chord-ties-1

At first this does not appear to be a graphical improvement as there are many collisions with noteheads and between adjacent ties. This is, however, already a problem in many situations and will receive proper treatment in a separate issue (#8980).

(Naturally, this is all moot where there is more than one voice on the stave.)

asattely:
Resolves: *https://github.com/musescore/MuseScore/issues/8973*

The rules of tie direction calculation from chords has been changed in accordance with the rules in the issue:
- for a chord with an even number of notes, the lower half curve down, the upper half curve up. If there is an odd number, the middle tie curves away from the stem (i.e. curving up if the tie is down, and vice versa); this happens to be the rule recommended by Gould(!)
- UNLESS the chord contains one (and only one) instance of the interval of a second (or a unison), in which case the tie direction changes at that point (the tie to the upper note of the second/unison curves up, as do the ties of any higher notes; the tie to the lower note of the second/unison curves down, as do the ties of any lower notes).

The following image demonstrates this fix's functionality:
image

### DIFF
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -610,48 +610,81 @@ void Tie::calculateDirection()
             //
             // chords
             //
-            int tiesCount = 0;
-            Note* tieNote = startNote();
-            Note* tieAbove = nullptr;
-            Note* tieBelow = nullptr;
-
-            for (size_t i = 0; i < n; ++i) {
-                if (notes[i]->tieFor()) {
-                    tiesCount++;
-                    int noteDiff = compareNotesPos(notes[i], tieNote);
-
-                    if (noteDiff > 0) {
-                        if (!tieAbove) {
-                            tieAbove = notes[i];
-                        } else if (compareNotesPos(notes[i], tieAbove) < 0) {
-                            tieAbove = notes[i];
-                        }
-                    } else if (noteDiff < 0) {
-                        if (!tieBelow) {
-                            tieBelow = notes[i];
-                        } else if (compareNotesPos(notes[i], tieBelow) > 0) {
-                            tieBelow = notes[i];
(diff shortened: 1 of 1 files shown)


asattely:
Had to amend the commit to account for codestyle problems

Jojo-Schmitz:
Here an image how it looks in 3.6.2:
image

oktophonie (approved):

RomanPudashkin (approved):

PR #9011, opened 2021-09-02
Renamed ScoreElement to EngravingObject, Element to EngravingItem

igorkorsukov:
Like QObject and QQuickItem accordingly

### DIFF
--- a/mtest/testutils.cpp
+++ b/mtest/testutils.cpp
@@ -61,7 +61,7 @@ namespace Ms {
 //    writes and element and reads it back
 //---------------------------------------------------------
 
-Element* MTest::writeReadElement(Element* element)
+EngravingItem* MTest::writeReadElement(EngravingItem* element)
 {
     //
     // write element
@@ -80,7 +80,7 @@ Element* MTest::writeReadElement(Element* element)
     XmlReader e(buffer.buffer());
     e.readNextStartElement();
     QString tag(e.name().toString());
-    element = Element::name2Element(e.name(), score);
+    element = EngravingItem::name2Element(e.name(), score);
     element->read(e);
     return element;
 }
--- a/mtest/testutils.h
+++ b/mtest/testutils.h
@@ -23,7 +23,7 @@
 #ifndef __TESTUTILS_H__
 #define __TESTUTILS_H__
 
-#include "libmscore/element.h"
+#include "libmscore/engravingitem.h"
 
 namespace Ms {
 class MScore;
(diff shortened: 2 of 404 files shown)


RomanPudashkin (approved):

wizofaus:
Since merging in this commit I'm getting a crash in Staff::~Staff on shut down, specifically inside EngravingObject::score( ), e->isScore( ) (line 805 of engravingobject.cpp), e appeared to an already deleted pointer.
This probably doesn't happen reliably unless you're using a debug runtime that deliberately fills deleted memory blocks with junk to help detect memory management issues.  I don't know which Staff it is, it seems to be a static variable somewhere.

wizofaus:
And now I'm seeing it fail on a QT assert

    if (type != ElementType::SCORE) {
        IF_ASSERT_FAILED(parent) {
        }
    }

In EngravingObject::EngravingObject, as part of `new Ms::Lasso` (from NotationInteraction)

 	Qt5Core.dll!qt_message_fatal(QtMsgType __formal, const QMessageLogContext & context, const QString & message) Line 1903	C++
 	Qt5Core.dll!QMessageLogger::fatal(const char * msg, ...) Line 893	C++
 	Qt5Core.dll!qt_assert(const char * assertion, const char * file, int line) Line 3359	C++
 	MuseScore4.exe!Ms::EngravingObject::EngravingObject(const Ms::ElementType & type, Ms::EngravingObject * parent) Line 63	C++
 	MuseScore4.exe!Ms::EngravingItem::EngravingItem(const Ms::ElementType & type, Ms::EngravingObject * se, QFlags<enum Ms::ElementFlag> f, mu::engraving::AccessibleElement * access) Line 216	C++
 	MuseScore4.exe!Ms::Lasso::Lasso(Ms::Score * s) Line 35	C++
>	MuseScore4.exe!mu::notation::NotationInteraction::NotationInteraction(mu::notation::Notation * notation, std::shared_ptr<mu::notation::INotationUndoStack> undoStack) Line 73	C++

wizofaus:
Actually ignore that latter one, seems to be the result of a merge issue, I can see you made the allocation m_lasso lazy (in NotationInteraction::doDragLasso).

igorkorsukov:
> Since merging in this commit I'm getting a crash in Staff::~Staff on shut down, specifically inside EngravingObject::score( ), e->isScore( ) (line 805 of engravingobject.cpp), e appeared to an already deleted pointer.
> This probably doesn't happen reliably unless you're using a debug runtime that deliberately fills deleted memory blocks with junk to help detect memory management issues. I believe the problem is when bracketItemOwner in palettecreator.cpp gets destroyed, it's quite likely that gpaletteScore is getting deleted before bracketItemOwner. If you make it a pointer instead it doesn't crash.
> 
> @igorkorsukov

Yes, now there is a crash when the application is closed. Hope it doesn't bother much.
This will be fixed later when I tidy up the creation and deletion of items.

PR #9014, opened 2021-09-02
Added engraving items factory 

igorkorsukov:


Eism (approved):

src/engraving/libmscore/engravingobject.cpp
```diff
@@ -102,6 +107,26 @@ EngravingObject::~EngravingObject()
...
 }
 
+void EngravingObject::doSetParent(EngravingObject* p)
+{
+    if (m_parent == p) {
+        return;
+    }
+
+    if (p) {
+        p->m_onDestroyed.resetOnReceive(this);
```
cbjeukendrup, line 117: Why is this needed? Shouldn't this be `m_parent` (the old parent) instead of `p` (the new one)?

igorkorsukov: oops, yeah, you're right.
in any case, this is a temporary hack, I will remove this code later.

PR #9000, opened 2021-09-01
[MU3] Fix GH#7449: Almost everything suitable for 3.6.3 or 3.7.0

Jojo-Schmitz:
Mostly resolves #7449 (i.e.: it includes the vast majority of the PRs listed there, but not all. Contains far more than listed there too)

Submitted mainly to get the development builds from GitHub CI (AKA artifacts, see the corresponding checks), for Windows (64-bit and 32-bit, no PortableApp unfortunately), Mac and Linux (64-bit AppImage) and also to get the vtests to run (which are expected to fail, but I'm curious in what way. Edit: As far as I can tell the vtests failures don't show any real failures, only changes for the better, but see for yourself, from the artifact of the corresponding check), and also for the mtests of course (which I expect it to pass with flying colors. Edit: indeed they do).

If you want to download and test, see https://github.com/musescore/MuseScore/wiki/Downloading-and-running-test-builds#Downloading-Builds-from-Pull-Requests, needs some adjustment for MuseScore 3 though.

Yes, it really is more than 360 commits (and still counting)...
All this on top of the 81 commits that the 3.x branch is ahead of 3.6.2 already.
So this indeed feels more like a 3.7(.0) rather than a 3.6.3, but as long as such a thing is not planned for, nor another PR for such a much smaller 3.6.3 exists...
Edit: I made up my mind and made it a 3.7.0 now ;-)
(Might also get named 3.9.0, to make clear that it really is the last minor release before/below 4.0)

### From PRs against the 3.x branch (not merged into it yet):

* #4853
* #5096 _(Despite the title this really is/was for 3.x, needs #7704)_
* #5752 _(Despite the title this really is/was for 3.x)_
* #5762 _(Despite its title it is rather based on 3.x, updated version)_
* #5805 _(Despite the title this really is/was for 3.x, needs #7704)_ 
* #5840 _(Despite the title this really is/was for 3.x, marked as WIP)_
* #5902 _(Needs #7704. Doesn't seem to work properly, @SKefalidis?)_ 
* #6095
* #6312 _(Might needs some compatibility switch. May cause 'denser' layouts, and so differences at where automatic system breaks are placed)_
* #6489
* #6693
* #7003
* #7388 _(Might be usefull to maintainers of the builds for Linux distributions)_
* #7443 _(Both parts)_
* #7453 / #7597
* #7455 / #8014
* #7463 / #7584 _(Partly)_
* #7464 / #7498
* #7500 _(Needs #7704)_
[...]

PR #9019, opened 2021-09-03
[MU3] Fix #274126 Crash when entering music on top of an unterminated slur by MIDI

lyrra:
Resolves: https://musescore.org/en/node/274126

Tiny fix, that enables different note-value entered from midi, on top of an slur.

Perhaps too late for 3.6.3, on the other hand, a very conservative change.

### DIFF
--- a/libmscore/noteentry.cpp
+++ b/libmscore/noteentry.cpp
@@ -256,17 +256,19 @@ Note* Score::addPitch(NoteVal& nval, bool addFlag, InputState* externalInputStat
             if (e) {
                   Fraction stick = Fraction(0, 1);
                   Element* ee = is.slur()->startElement();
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
I guess I'll juist add it to #9000, but won't it apply to master too?

lyrra:
> I guess I'll just add it to #9000 (Edit: done), but won't it apply to master too?
I guess it applies to master too. But I can't verify it in MU4, no audio/midi in gui.
MU3 is tested by launching two musescores and midi-connecting them in qjackctl graph.

Jojo-Schmitz:
Yes, 3.x is often still pretty usefull as a testbed

RomanPudashkin:
3.x is closed for any changes

PR #9033, opened 2021-09-06
Pianoroll2

blackears:


src/pianoroll/view/pianorollview.h
```diff
@@ -0,0 +1,254 @@
...
+    void cutChord(const QPointF& pos);
+    void eraseNote(const QPointF& pos);
+    Ms::Fraction roundDownToSubdivision(double wholeNote);
+
+    void pasteNotes(const QString& copiedNotes, Ms::Fraction pasteStartTick, Ms::Fraction lengthOffset, int pitchOffset, bool xIsOffset);
+    std::vector<Ms::Note*> addNote(Ms::Fraction startTick, Ms::Fraction duration, int pitch, int track);
+    bool cutChordRest(Ms::ChordRest* targetCr, int track, Ms::Fraction cutTick, Ms::ChordRest*& cr0, Ms::ChordRest*& cr1);
+    std::vector<Ms::Note*> getSegmentNotes(Ms::Segment* seg, int track);
+
+    static void append(std::vector<Ms::Note*>& a, std::vector<Ms::Note*>& b)
```
wizofaus, line 187: Questionable whether this function belongs here, but if you make the second parameter a const ref you should avoid most of those build errors.

src/pianoroll/view/pianorollview.h
```diff
@@ -0,0 +1,254 @@
...
+public:
+    enum class PianorollTool : char { SELECT, ADD, CUT, ERASE };
+    Q_ENUM(PianorollTool)
+
+    static const BarPattern barPatterns[];
+
+private:
+    INJECT(pianoroll, context::IGlobalContext, globalContext)
+    INJECT(pianoroll, IPianorollController, controller)
+    INJECT(playback, audio::IPlayback, playback)
```
wizofaus, line 83: believe it should be pianoroll for the 1st parameter here (not entirely sure what difference it makes though!)

src/pianoroll/view/pianorollview.h
blackears: I've rewritten the code to avoid this method entirely.

src/pianoroll/view/pianorollview.h
blackears: Fixed.

igorkorsukov: > not entirely sure what difference it makes though!

It's actually just for information. In the future, I want to make a diagnostic page on which to show what depends on what.

blackears:
Created new PR to replace this one. https://github.com/musescore/MuseScore/pull/9109/files

PR #9036, opened 2021-09-06
Moved creation an "accidental" to factory

igorkorsukov:
The problem is - when creating engraving elements, we need to do some initialization:
1. Initialize the style (now the virtual function is called in the element constructors, we should not call the virtual function in the constructors and destructors)
2. We need to register the item with the accessibility system. But the implementation of accessibility for some elements may be different, this creates difficulties in registration in the constructor of the base class (we cannot just call a virtual function)

To solve these and other tasks, the elements need to be created by a factory, and accordingly, then we can call the init virtual method, in which we can make the necessary initialization.

This is the start of the big task of bringing creation to a factory.
This PR is made in order to disable diagnostic paint. I forgot to comment on the appropriate definition. In order not to forget to do this anymore, I moved this define to the cmake option, which is disabled by default.


wizofaus:
I believe you also need to fix NotationInteraction::doEndDrag to be 

`if (m_lasso && !m_lasso->bbox().isEmpty())`

as the m_lasso pointer can now be nullptr when releasing the mouse after dragging.

igorkorsukov:
> I believe you also need to fix NotationInteraction::doEndDrag to be
> 
> `if (m_lasso && !m_lasso->bbox().isEmpty())`
> 
> as the m_lasso pointer can now be nullptr when releasing the mouse after dragging.

Thank you. I will fix it in next PR

Eism (approved):

PR #9052, opened 2021-09-07
[MU4] Fix #324413: Allow importing files with uppercase extensions

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/324413

Same method is used for GuitarPro .gtp, .pg and .gpx too, but broke at the "front door"

### DIFF
--- a/src/framework/global/io/path.cpp
+++ b/src/framework/global/io/path.cpp
@@ -85,7 +85,7 @@ paths path::pathsFromString(const std::string& str, const std::string& delim)
 std::string mu::io::suffix(const mu::io::path& path)
 {
     QFileInfo fi(path.toQString());
-    return fi.suffix().toStdString();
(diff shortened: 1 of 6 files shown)


src/framework/global/io/path.cpp
```diff
@@ -85,7 +85,7 @@ paths path::pathsFromString(const std::string& str, const std::string& delim)
 std::string mu::io::suffix(const mu::io::path& path)
 {
     QFileInfo fi(path.toQString());
-    return fi.suffix().toStdString();
+    return fi.suffix().toLower().toStdString();
```
Jojo-Schmitz, line 88: This seems the "front door", Downshifting the suffix here should solve it everywhere, I think.

RomanPudashkin (approved):

PR #9053, opened 2021-09-07
[MU4] Fix some compiler warnings

Jojo-Schmitz:
1. Actually seen while (accidentially) doing an `lupdate` (which works surprisingly fast now, in 3.x (and on Windows) it takes more than an hour, here it took just about a minute). Not really compiler warnings though...
2. Code style issues seen after fixing the 1st
3. May not affect master, but will 3.x. And won't harm here
4. Fixing one of quite many warnings seen on GitHub CI for Mac
5. Fixing compiler warnings seen after rebasing to include #8976, revisited after rebasing to include #9069, which resolved some of those warnings in a slightly different way
6. Fixing unhandled cases.

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   build_mu3:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     steps:
     - name: Cancel Previous Runs
       uses: styfle/cancel-workflow-action@0.5.0
@@ -116,7 +116,7 @@ jobs:
         path: ./build.artifacts/
 
   run_mtests:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     env:
         # Enable AddressSanitizer in the mtest build
         CFLAGS: "-fsanitize=address -fno-omit-frame-pointer"
--- a/src/engraving/infrastructure/draw/painterpath.cpp
+++ b/src/engraving/infrastructure/draw/painterpath.cpp
@@ -25,15 +25,12 @@
 #include "log.h"
 
 namespace mu {
-
 static constexpr double pi = 3.14159265358979323846;
 static constexpr double pathKappa = 0.5522847498;
 
-static void findEllipseCoords(const RectF &r, double angle, double length,
-                            PointF* startPoint, PointF *endPoint);
+static void findEllipseCoords(const RectF& r, double angle, double length, PointF* startPoint, PointF* endPoint);
 
-static PointF curvesForArc(const RectF &rect, double startAngle, double sweepLength,
-                          PointF *curves, int *point_count);
+static PointF curvesForArc(const RectF& rect, double startAngle, double sweepLength, PointF* curves, int* point_count);
 
 static double angleForArc(double angle);
 
@@ -62,19 +59,19 @@ void PainterPath::lineTo(const PointF& p)
 {
     if (!hasValidCoords(p)) {
 #ifdef TRACE_DRAW_OBJ_ENABLED
-    LOGW() << "PainterPath::lineTo: Adding point with invalid coordinates, ignoring call";
+        LOGW() << "PainterPath::lineTo: Adding point with invalid coordinates, ignoring call";
 #endif
         return;
     }
     ensureData();
     setDirty();
-    
+
     assert(!m_elements.empty());
     maybeMoveTo();
     if (p == PointF(m_elements.back())) {
         return;
     }
-    
+
     m_elements.push_back({ p.x(), p.y(), ElementType::LineToElement });
     m_convex = m_elements.size() == 3 || (m_elements.size() == 4 && isClosed());
 }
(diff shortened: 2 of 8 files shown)


src/framework/audio/internal/worker/equaliser.h
```diff
@@ -47,7 +47,7 @@ class Equaliser : public IFxProcessor
     unsigned int m_sampleRate = 0;
     bool m_active = true;
 
-    float m_gain = 0, m_frequency = 1'000.f, m_q = 1.f;
+    float m_gain = 0, m_frequency = 1000.f, m_q = 1.f;
```
Jojo-Schmitz, line 50: I really wonder why/how that ever compiled?

src/framework/audio/internal/worker/sinesource.h
```diff
@@ -36,7 +36,7 @@ class SineSource : public AbstractAudioSource
     void process(float* buffer, unsigned int sampleCount) override;
 
 private:
-    float m_frequency = 1'000.f;
+    float m_frequency = 1000.f;
```
Jojo-Schmitz, line 39: Same here

src/engraving/infrastructure/draw/painterpath.cpp
```diff
@@ -41,7 +41,7 @@ void PainterPath::moveTo(const PointF& p)
 {
     if (!hasValidCoords(p)) {
 #ifdef TRACE_DRAW_OBJ_ENABLED
-        LOGW() << "PainterPath::moveTo: Adding point with invalid coordinates, ignoring call");
+        LOGW() << "PainterPath::moveTo: Adding point with invalid coordinates, ignoring call";
```
Jojo-Schmitz, line 44: Probably TRACE_DRAW_OBJ_ENABLED` never got used so far?

Jojo-Schmitz:
Some code style check failures, which I deny any responsibility for ;-)

src/framework/audio/internal/worker/equaliser.h
lyrra: It seems to be some c++14 fanciness:
https://stackoverflow.com/questions/54447208/why-can-i-separate-digits-with-an-apostrophe-in-c-but-not-c

On a similar note, I was wondering why MU4 dropped -std= flag when compiling (MU3 has -std=gnu++11). It works by luck because gcc defaults to c++14 (version 6 up to 10 it seems), and as such you shouldn't get a warning?

src/engraving/infrastructure/draw/painterpath.cpp
Jojo-Schmitz: Seems rempving that stray paren cause the code style check to kick in?

src/framework/audio/internal/worker/equaliser.h
Jojo-Schmitz: At least `lupdate` stumbles accros this, so let's better avoid to use this particular feature

Jojo-Schmitz: MU4 switched to C++14, Guess that's why?

Jojo-Schmitz: But yes, better make this explicit

Jojo-Schmitz:
I've fixed the code style issues anyway. Not a good sign that these had been missed, just because of a stray paren in some `#ifdef`d out code?

src/framework/audio/internal/worker/equaliser.h
cbjeukendrup: From https://en.cppreference.com/w/cpp/language/integer_literal:

> Optional single quotes (') may be inserted between the digits as a separator. They are ignored by the compiler. (since C++14)

Jojo-Schmitz: Yes, that's what @lyrra said already, basically. So I learen a new C++14 feature today ;-)
And as per my son (student of computer science) C23 will have this feature too.
I don't think this is really needed here though, and apparently throws off `lupdate`, so we'd better avoid using it as long as there are no compelling reasons.

.github/workflows/ci_linux.yml
```diff
@@ -19,7 +19,7 @@ on:
 
 jobs:
   build_mu3:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
```
Jojo-Schmitz, line 22: Even if not used for MuseScore 4 currently

.github/workflows/ci_linux.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         path: ./build.artifacts/
 
   run_mtests:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
```
Jojo-Schmitz, line 119: Not sure whether this is used for MuseScore 4?
Still won#t harm to change it.

cbjeukendrup:
Wait... remember [that empty `painterpath.cpp.uncrustify` that got accidentally added to PR #8867](https://github.com/musescore/MuseScore/pull/8867#issuecomment-909486011)? 
Every time when I run uncrustify, a file like that gets created. So it got really upset by that stray parens... Should be fine after this PR.

lyrra (approved):

Jojo-Schmitz:
I guess `uncrustify` is similarly unaware of conditional compiles as `lupdate` Which would explain this quirk.

src/inspector/models/inspectormodelcreator.cpp
```diff
@@ -142,6 +142,9 @@ AbstractInspectorModel* InspectorModelCreator::newInspectorModel(AbstractInspect
         return new TremoloSettingsModel(parent, repository);
     case AbstractInspectorModel::InspectorModelType::TYPE_MEASURE_REPEAT:
         return new MeasureRepeatSettingsModel(parent, repository);
+    case AbstractInspectorModel::InspectorModelType::TYPE_BREATH:
+    case AbstractInspectorModel::InspectorModelType::TYPE_ARPEGGIO:
+    case AbstractInspectorModel::InspectorModelType::TYPE_DYNAMIC:
```
Jojo-Schmitz, line 147: Athough these seem to indicate that there currently is no Inspector for Breaths, Arpeggions and Dynamics

igorkorsukov (approved):

PR #9068, opened 2021-09-08
horizontal fixed layout added (each measure has the same size)

alexpavlov96:

### DIFF
--- a/src/engraving/libmscore/masterscore.h
+++ b/src/engraving/libmscore/masterscore.h
@@ -44,20 +44,20 @@ class MasterScore : public Score
     TempoMap* _tempomap;
     RepeatList* _repeatList;
     RepeatList* _repeatList2;
-    bool _expandRepeats     { MScore::playRepeats };
-    bool _playlistDirty     { true };
+    bool _expandRepeats = MScore::playRepeats;
+    bool _playlistDirty = true;
     QList<Excerpt*> _excerpts;
     std::vector<PartChannelSettingsLink> _playbackSettingsLinks;
     Score* _playbackScore = nullptr;
     Revisions* _revisions;
 
-    bool _readOnly          { false };
+    bool _readOnly = false;
 
     CmdState _cmdState;       // modified during cmd processing
 
     Fraction _pos[3];                      ///< 0 - current, 1 - left loop, 2 - right loop
 
-    int _midiPortCount      { 0 };                    // A count of ALSA midi out ports
+    int _midiPortCount = 0;                           // A count of ALSA midi out ports
     QQueue<MidiInputEvent> _midiInputQueue;           // MIDI events that have yet to be processed
     std::list<MidiInputEvent> _activeMidiPitches;     // MIDI keys currently being held down
     std::vector<MidiMapping> _midiMapping;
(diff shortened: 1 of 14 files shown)


Jojo-Schmitz:
Seems this would fix a pretty frequent request, see https://musescore.org/en/node/313555.
If so maybe change the commit title to start with "Fix #313555: "?

src/engraving/libmscore/measure.cpp
```diff
@@ -4823,4 +4824,238 @@ void Measure::computeMinWidth()
...
+    x += segment->extraLeadingSpace().val() * spatium();
+
+    return x;
+}
+
+static Segment* findFirstEnabledSegment(Measure* measure)
+{
+    Segment* current = nullptr;
+    for (current = measure->first(); current && !current->enabled(); current = current->next()) {
+    }
```
Jojo-Schmitz, line 4859: Empty `for` loop?

Jojo-Schmitz:
Isn't there any UI to switch this on/off?

alexpavlov96:
> Seems this would fix a pretty frequent request, see https://musescore.org/en/node/313555.
> If so maybe change the commit title to start with "Fix #313555: "?
> 
> Isn't there any UI to switch this on/off?

@Jojo-Schmitz there is no UI, I added the code that works on MU mobile so we have the same library. I didn't know about the issue. I tested with setting in layoutoptions.h:

struct LayoutOptions
{
    LayoutMode mode = LayoutMode::HORIZONTAL_FIXED;

src/engraving/libmscore/measure.cpp
alexpavlov96: it chanegs the current segment

Jojo-Schmitz: Yes, I see that it used the side effect, but that's confusing code, to me at least.

Jojo-Schmitz:
OK, so a UI for this is missing, but needed.

src/engraving/libmscore/measure.cpp
igorkorsukov: I agree with JoJo, I believe needs rewrite this code

src/engraving/libmscore/masterscore.h
```diff
@@ -66,6 +66,8 @@ class MasterScore : public Score
     QSet<int> occupiedMidiChannels;                   // each entry is port*16+channel, port range: 0-inf, channel: 0-15
     unsigned int searchMidiMappingFrom = 0;           // makes getting next free MIDI mapping faster
 
+    qreal m_widthOfSegmentCell { 3 };
```
igorkorsukov, line 69: needs `qreal m_widthOfSegmentCell = 3.0;`(https://github.com/musescore/MuseScore/wiki/CodeGuidelines#declaring-variables)

src/engraving/libmscore/measure.cpp
```diff
@@ -4823,4 +4824,238 @@ void Measure::computeMinWidth()
...
 }
+
+void Measure::layoutSegmentsInPracticeMode(const std::vector<int>& visibleParts)
+{
+    layoutSegmentsWithDuration(visibleParts);
+}
+
+qreal Measure::computeFirstSegmentXPosition(Segment* segment)
+{
+    qreal x{ 0 };
```
igorkorsukov, line 4835: needs `qreal x = 0.0;` 
https://github.com/musescore/MuseScore/wiki/CodeGuidelines#declaring-variables

src/engraving/libmscore/measure.cpp
Jojo-Schmitz: yes, that while loop is better

alexpavlov96:
> OK, so a UI for this is not part of this PR (yet), but needed.
> At least an advanced preference setting.

@Jojo-Schmitz, no UI yet, so I didn't change the name

igorkorsukov:
We plan to make the UI later

Issue #9072, opened 2021-09-08
[MU4 Issue] Closing staff properties dialog causes crash (debug build only?)


wizofaus:
**Describe the bug**
After accessing the staff properties dialog and clicking cancel or OK, MuseScore crashes. Only seems to happen using debug version of Qt library though.

**To Reproduce**
1. In any score, select a measure etc., right click and should "staff properties"
2. Click OK or Cancel
3. Crash

**Expected behavior**
Not that...

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**
Using debug version of Qt library

```
 	Qt5Qmld.dll!QFlagPointer<QQmlDelayedError>::isNull() Line 160	C++
 	Qt5Qmld.dll!QQmlJavaScriptExpression::hasError() Line 252	C++
 	Qt5Qmld.dll!QQmlBoundSignal_callback(QQmlNotifierEndpoint * e, void * * a) Line 362	C++
 	Qt5Qmld.dll!QQmlNotifier::emitNotify(QQmlNotifierEndpoint * endpoint, void * * a) Line 105	C++
 	Qt5Qmld.dll!QQmlData::signalEmitted(QAbstractDeclarativeData * __formal, QObject * object, int index, void * * a) Line 836	C++
>	Qt5Cored.dll!doActivate<0>(QObject * sender, int signal_index, void * * argv) Line 3782	C++
  ....
```
Note that a number of asserts occur when opening this dialog and this is printed to the debug output:

```
21:53:45.776 | ERROR | main_thread | Qt         | destroyed: Object 0x20341840 destroyed while one of its QML signal handlers is in progress.
Most likely the object was deleted synchronously (use QObject::deleteLater() instead), or the application is running a nested event loop.
This behavior is NOT supported!
qrc:/qml/MuseScore/UiComponents/StyledMenuLoader.qml:56: function() { [native code] }
Debug Error!

Program: ...usescore\MuseScore\msvc.build_x64\src\main\Debug\Qt5Cored.dll
Module: 5.15.2
qml\qqmlengine.cpp
Line: 1926

Object 0x20341840 destroyed while one of its QML signal handlers is in progress.
Most likely the object was deleted synchronously (use QObject::deleteLater() instead), or the application is running a nested event loop.
This behavior is NOT supported!
qrc:/qml/MuseScore/UiComponents/StyledMenuLoader.qml:56: function() { [native code] }

(Press Retry to debug the application)
[...]

PR #9080, opened 2021-09-08
[MU4] Fixed crash after opening dialogs via the context menu

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/9072


wizofaus:
Can confirm this resolves the problem (including all the asserts etc.)

PR #9086, opened 2021-09-08
fix #9020: Select correct brace symbol for each system independently

### LINKED ISSUE #9020: [MU4 Issue] Select correct brace symbol for each system independently
MuseScore has access to four alternative brace symbols which are to be used according to the number of staves that they span, currently defined as:

- 1 stave: `braceSmall`
- 2 staves: `brace` (the default glyph)
- 3 staves: `braceLarge`
- 4+ staves: `braceLarger`

(see: https://w3c.github.io/smufl/latest/tables/staff-brackets-and-dividers.html)

However, this is calculated using the number of staves belonging to the instrument. Instead, it should be calculated _for each system independently_, based on how many staves the brace needs to span on that system. In, for example, a piano piece, a system with two staves will use `brace`, a system with 3 will use `braceLarge`, etc.

**Should fix:**
https://musescore.org/en/node/322436
(the matter discussed in that issue of aligning the starts of systems to be fixed by #8968)

Nick-Mazuk:
Resolves: #9020

MuseScore uses different brace symbols depending on the number of staves the brace covers. Currently, this is is calculated once for the entire piece. This PR allows the brace size to be calculated on a system-by-system basis, such that staves hidden in a particular system don't affect the size of the brace.

### DIFF
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -598,12 +598,12 @@ Bracket* System::createBracket(Ms::BracketItem* bi, int column, int staffIdx, QL
     }
 
     for (; firstStaff <= lastStaff; ++firstStaff) {
-        if (score()->staff(firstStaff)->show()) {
+        if (staff(firstStaff)->show()) {
             break;
         }
     }
     for (; lastStaff >= firstStaff; --lastStaff) {
-        if (score()->staff(lastStaff)->show()) {
+        if (staff(lastStaff)->show()) {
             break;
         }
     }
@@ -612,7 +612,7 @@ Bracket* System::createBracket(Ms::BracketItem* bi, int column, int staffIdx, QL
     // do not show bracket if it only spans one
     // system due to some invisible staves
     //
-    if ((span > 1) || (bi->bracketSpan() == span)) {
+    if ((span > 1) || (bi->bracketSpan() == span) || (span == 1 && score()->styleB(Sid::alwaysShowBracketsWhenEmptyStavesAreHidden))) {
         //
         // this bracket is visible
         //
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -612,7 +612,9 @@ Bracket* System::createBracket(Ms::BracketItem* bi, int column, int staffIdx, QL
     // do not show bracket if it only spans one
     // system due to some invisible staves
     //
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
vtest error might be expected, but certainly not this one?

Jojo-Schmitz:
If you prefix the commit title with "Fix #322436:" that issue would (also) get auto-closed on merging this PR

Jojo-Schmitz (approved):
Before:
image
after:
image

Nick-Mazuk:
> vtest error might be expected, but certainly not this one?

It looks like the test failed when creating the reference images, not even during the actual tests. I'm not too knowledgable about the test suite, but I don't know why this change would cause the image creation to fail.



> If you prefix the commit title with "Fix #322436:" that issue would (also) get auto-closed on merging this PR

Good to know. Typically, GitHub auto-closes issues even if it's referenced in the PR message as "resolved" or "fixed". But if adding it to the commit message that way is the preferred way to do things here, I'll do that in the future.

Jojo-Schmitz:
It is for issues on the muescore.org tracker

Nick-Mazuk:
> Reg. vtest:
> The vtest error is something that happens at times, nothing to worry about.
> Except it'd be interesting to see whether there's a 'real' vtest a) testing these braces and b) failing on this (rightfully in this case, alöbeit a failure we'd have to accept).
> But checking the [vtests of my backport of this to 3.x](https://github.com/musescore/MuseScore/pull/9000/checks?check_run_id=3553238349) show no such test nor failure (but a lot of unrelated others).
> 
> Reg. commit title:
> It is for issues on the muescore.org tracker

I see, thanks for the explanations!

DmitryArefiev:
Hi @Nick-Mazuk  can you help with testing this. I don't see a difference between latest master (Win) and old build 20-days old.

And one more thing.. The brace size is different depending on what staves are hidden (in the middle of system or starting from the edge):
brace testing

Jojo-Schmitz:
That's what the PR is all about. The brace size is picked based on the number of __visible__ staves

Nick-Mazuk:
@DmitryArefiev The desired behavior is the one where the brace is smaller when you hide Flute 2, 3, 4. When you hide the middle two staves, the brace size should also become smaller (which it isn't, so that's a bug). I think I know what's going on and will submit a new PR with a fix hopefully today.

PR #9092, opened 2021-09-09
added option to always layout the title on the same page as beginning of the score

alexpavlov96:
when there's a lot of instruments, the first page may not fit, so title is layouted on the separate page

### DIFF
--- a/src/engraving/layout/layoutpage.cpp
+++ b/src/engraving/layout/layoutpage.cpp
@@ -206,7 +206,11 @@ void LayoutPage::collectPage(const LayoutOptions& options, LayoutContext& lc)
                 qreal margin = qMax(lc.curSystem->minBottom(), lc.curSystem->spacerDistance(false));
                 dist += qMax(margin, slb);
             }
-            breakPage = (y + dist) >= ey && breakPages;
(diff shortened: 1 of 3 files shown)


Jojo-Schmitz:
This one too is without a UI, without any way to set that option, or am I missing something?

alexpavlov96:
> This one too is without a UI, without any way to set that option, or am I missing something?

You're right, there is no UI

Jojo-Schmitz:
Also wouldn't that lead to the bottom part of a system to not fit the page anymore?

alexpavlov96:
> Also wouldn't that lead to the bottom part of a system to not fit the page anymore?

I'm checking that!

Jojo-Schmitz:
Apparantly I did not, yet ;-)
So sorry for the noise.

Jojo-Schmitz:
Ah, I see. I took your "I'm checking that" for "checking that is in the code alerady"

Jojo-Schmitz:
Actually the fact that MuseScore does automagically add a page break should already indicate that this just doesn't fit the page

Jojo-Schmitz:
A Vertical frame, on top of a page, needs its height plus its lower distance as extra space. So you'd need to make the staff distance smaller for it to fit, by `staff_distance = (number_of_staves - 1) / (frame-height + frem_bottom_distance)`. Or something along those lines...

PR #9095, opened 2021-09-09
fix #9013 Adjust dot placement for upward hooks

### LINKED ISSUE #9013: [MU4 Issue] Dots should be offset rightward to clear flag
In certain situations, where a note has both a flag and a dot, the flag and dot collide.
In these cases the dot should be offset rightward to clear the flag.

Currently, where a note is on a line and the dot is displaced into the space above, MuseScore attempts to avoid the collision by extending the stem length by 0.5sp. This is not a good solution (it creates inconsistent-looking stem lengths, and usually doesn't avoid the collision anyway.)

dot-flag

This adjustment will not be required in every case where a note has both a flag and a dot, but only:

- when the stem is up
- when the flag symbol actually extends vertically into the space where the dot is drawn; this depends on the stem length and the height of the specific flag glyph itself (which also varies in different fonts)

In the case of a chord, the dots should always align, so if one needs to be offset, they must all be offset.

**Should fix:**
https://musescore.org/en/node/323759

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/9013

Quick fix to move dots over if the stem is up and there is a hook on the chord and the chord's dot placement is left of the right bound of the hook:
image

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1365,6 +1365,7 @@ qreal Chord::defaultStemLength() const
     }
 
     qreal normalStemLen = isSmall() ? 2.5 : 3.5;
+    /* We no longer need to adjust stem hight for dot-hook collisions (Issue #9095)
     if (hookIdx && tab == 0) {
         if (up() && durationType().dots()) {
             //
@@ -1375,7 +1376,7 @@ qreal Chord::defaultStemLength() const
             }
             shortenStem = false;
         }
-    }
+    } */
 
     if (isGrace()) {
         // grace notes stems are not subject to normal
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -2269,7 +2269,7 @@ void Note::layout2()
             // the top dot in the chord, not the dot for this particular note:
             qreal dotY = chord()->notes().back()->y() + chord()->notes().back()->dots().first()->pos().y();
             if (chord()->dotPosX() < hookRight && dotY < hookBottom) {
-                d = chord()->hook()->width();
+                d = chord()->hook()->width() + (0.25 * spatium());
             }
         }
         // if TAB and stems through staff
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -2265,7 +2265,10 @@ void Note::layout2()
         // adjust dot distance for hooks
         if (chord()->hook() && chord()->up()) {
             qreal hookRight = chord()->hook()->width() + chord()->hook()->x() + chord()->pos().x();
-            if (chord()->dotPosX() < hookRight) {
+            qreal hookBottom = chord()->hook()->height() + chord()->hook()->y() + chord()->pos().y() + (0.25 * spatium());
+            // the top dot in the chord, not the dot for this particular note:
+            qreal dotY = chord()->notes().back()->y() + chord()->notes().back()->dots().first()->pos().y();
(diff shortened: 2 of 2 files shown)


asattely:
There are situations where the dot doesn't need to be adjusted, so this has been fixed with the most recent commit:
image

asattely:
image

src/engraving/libmscore/chord.cpp
```diff
@@ -1365,6 +1365,7 @@ qreal Chord::defaultStemLength() const
     }
 
     qreal normalStemLen = isSmall() ? 2.5 : 3.5;
+    /* We no longer need to adjust stem hight for dot-hook collisions (Issue #9095)
```
Jojo-Schmitz, line 1368: Better use #if 0... #endif. Here you're creating a compiler warning "comment inside comment"

Jojo-Schmitz:
Rebase needed

Jojo-Schmitz:
rebase needed

asattely:
rebase done did

src/engraving/libmscore/note.cpp
```diff
@@ -2261,6 +2262,16 @@ void Note::layout2()
...
         qreal dd = score()->point(score()->styleS(Sid::dotDotDistance)) * mag();
         qreal x  = chord()->dotPosX() - pos().x() - chord()->pos().x();
+        // adjust dot distance for hooks
+        if (chord()->hook() && chord()->up()) {
+            qreal hookRight = chord()->hook()->width() + chord()->hook()->x() + chord()->pos().x();
+            qreal hookBottom = chord()->hook()->height() + chord()->hook()->y() + chord()->pos().y() + (0.25 * spatium());
+            // the top dot in the chord, not the dot for this particular note:
+            qreal dotY = chord()->notes().back()->y() + chord()->notes().back()->dots().first()->pos().y();
+            if (chord()->dotPosX() < hookRight && dotY < hookBottom) {
+                d = chord()->hook()->width() + (0.25 * spatium());
```
Jojo-Schmitz, line 2272: are both these `0.25` the same number? If so can't we consolidate those into one variable?

asattely: They are not. The first one is adding a little bit of margin to the bottom of a hook for collision detection purposes (because otherwise they could be still right next to each other and not be technically colliding), and the second adds margin to the right side of the hook. Vertical vs horizontal positioning.

Jojo-Schmitz: Thanks for explaining.

DmitryArefiev:
Checked #9013 on Win10.

Hi @asattely , It's better than it was. But I think the dot can be too far from note in some cases and can collide with rests:




Even comparing with a desired result from #9013 : 


@Tantacrul @bkunda Please review

Jojo-Schmitz:
Which might mean that we'd need 3 different distances, no hook half hook and full hook?

Tantacrul:
This is very much an @oktophonie call

DmitryArefiev:
Found a minor bug about shadow note dot position and real dot position. Logged separately #9265

Dotted shadow note

DmitryArefiev:
Checked #9013 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

Please merge into master

src/engraving/libmscore/chord.cpp
RomanPudashkin: Better to remove this code completely

DmitryArefiev:
@igorkorsukov @Eism Please merge into master

PR #9097, opened 2021-09-09
fix #8970: remove space between accidentals and arpeggios

### LINKED ISSUE #8970: [MU4 Issue] Accidentals create too much space for arpeggio lines
The presence of an accidental creates too much extra space between an arpeggio line and the following chord. (This affects all vertical lines from the Arpeggios/Glissandos palette, not just the standard wiggle.)

arpeggios

**Should fix:**
https://musescore.org/en/node/292231 (points 1 and 2)

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/8970

Accidentals have a quarter space padding around them. This creates an unnecessarily large gap between accidentals and arpeggios. This PR decreases that gap by a quarter space, making the gap the same size as if the accidentals weren't there.

Before:

Screen Shot 2021-09-09 at 4 00 08 PM

After:

Screen Shot 2021-09-09 at 3 55 52 PM

Currently, the decrease of a quarter space is hardcoded in. Alternatively, we can use the same value that defines the spaces between accidentals. I have not yet found where this is defined, though.

### DIFF
--- a/src/engraving/libmscore/arpeggio.cpp
+++ b/src/engraving/libmscore/arpeggio.cpp
@@ -607,13 +607,16 @@ qreal Arpeggio::insetDistance(QVector<Accidental*>& accidentals, qreal mag_) con
     qreal top = furthestAccidental->note()->pos().y() * mag_ + bbox.top() * mag_;
     qreal bottom = furthestAccidental->note()->pos().y() * mag_ + bbox.bottom() * mag_;
 
-    // if it collides with the top arrow/hoo
-    if (hasTopArrow && top <= arpeggioTop) {
-        return accidentalCutOutX;
-    }
-
-    // if it collides with the bottom arrow/hook
-    if (hasBottomArrow && bottom >= arpeggioBottom) {
+    // if it collides with the top arrow/hook
+    if ((hasTopArrow && top <= arpeggioTop)
+        // if it collides with the bottom arrow/hook
+        || (hasBottomArrow && bottom >= arpeggioBottom)) {
+        // optical adjustment for one edge case
+        if (type == ArpeggioType::BRACKET
+            && (furthestAccidental->symbol() == SymId::accidentalNatural
+                || furthestAccidental->symbol() == SymId::accidentalFlat)) {
+            return accidentalCutOutX + score()->styleP(Sid::ArpeggioAccidentalDistance) * mag_ / 2;
+        }
         return accidentalCutOutX;
     }
 
--- a/src/engraving/libmscore/arpeggio.cpp
+++ b/src/engraving/libmscore/arpeggio.cpp
@@ -28,6 +28,7 @@
 #include "io/xml.h"
 
 #include "scorefont.h"
+#include "accidental.h"
 #include "chord.h"
 #include "note.h"
 #include "score.h"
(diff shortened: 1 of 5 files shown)


Nick-Mazuk:
I could go either way. I tried both and went with 0.25 spaces because then the gap size is identical with and without accidentals.

Jojo-Schmitz:
Fair enough. To me in the images from original issue's score the distance between barline and arpeggio is still too small with 0.25. But that's nothing this PR here touches on, not directly at least.

Nick-Mazuk:
Just added a commit that does a few things:

1. Shrinks gap to just 0.25 spaces
2. Removes the hard coded magic number for the accidental padding
3. Added a style setting for the gap size. It's currently not exposed to the user, but it can be in the future if we wish

Nick-Mazuk:
This is a complete rewrite of the feature, but I believe it's finalized.

1. Increases gap back to 0.5 spaces
2. Decreases the gap for ♯ by the [cutout](https://w3c.github.io/smufl/latest/specification/bbox-cut-outs.html) to make it visually consistent with ♭
3. The gap follows the contour of the arpeggio and accidentals (i.e., if an accidental can fit closer to the arpeggio than it's bounding box, it will be closer)

For instance, take a look at this image. The sharp now fits inside of the bracket because there's space!

image

All test cases:

accidental-test

Here's the test file I used to validate the changes.

[accidental-test.mscz.zip](https://github.com/musescore/MuseScore/files/7172554/accidental-test.mscz.zip)

Nick-Mazuk:
> ; missing

Fixed

Nick-Mazuk:
Assuming this looks good with everyone, I'm done working on it and it can be merged.

Jojo-Schmitz:
You may want to squash commits. Those last 3 commits should really be just one

src/engraving/libmscore/chord.cpp
```diff
@@ -1914,6 +1914,12 @@ void Chord::layoutPitched()
     //  process notes
     //-----------------------------------------
 
+    // Keeps track if there are any accidentals in this cord.
```
Jojo-Schmitz, line 1917: typo, "chord"

Nick-Mazuk:
Fixed typo and squashed the commits

Nick-Mazuk:
@Jojo-Schmitz, resolved your comments about the unreachable `breaks` and qreals.

Nick-Mazuk:
As requested by @oktophonie, the gap at measure 15 has been reduced to this:

Screen Shot 2021-09-19 at 8 45 46 PM

Nick-Mazuk:
These last few updates take care of every last edge case I can think of, including checking the output in multiple fonts. There are also a few minor tweaks as requested by Simon.

There's also some code refactoring to make sure it's easier to read.

Here's what the original test cases look like now:

accidental-test

And with Petaluma, a font with completely different metrics:

accidental-test

oktophonie (approved):

src/engraving/libmscore/arpeggio.cpp
```diff
@@ -23,14 +23,17 @@
 #include "arpeggio.h"
 
 #include <cmath>
+#include <QVector>
```
vpereverzev, line 26: At the moment we're trying to reduce dependencies from Qt in engraving module. Make sure that you're using STL containers here and in the further PRs

Nick-Mazuk: Ok, will do. If I notice that there's an unneeded, existing Qt dependency, I assume I should convert it to a STL container?

Jojo-Schmitz: In my backport of this PR to 3.x that include was not needed (Neither was the cmath) and 3.x's all.h doesn't either.

Jojo-Schmitz: You're actually using a `QVector` in chord.cpp too bit not the include

DmitryArefiev:
Assigning to @vshalkevich

vshalkevich:
Checked on Linux Ubunty, Windows 10, MacOS
Works fine

PR #9102, opened 2021-09-10
[MU4] Fix #318534, fix #327790: Add Missing muted channels in several templates' Horns and Trombones and add missing combined articulations to templates

Jojo-Schmitz:
* Fix #318534: Missing muted channels in several templates' Horns and Trombones
   That update reveals that range and combined articulations were not matching too.
   This is for master what #7705 is (or rather was) for 3.x
   Resolves: https://musescore.org/en/node/318534
* Fix #327790: Add missing combined articulations to templates
   Those overlooked in the earlier commit
   Resolves: https://musescore.org/en/node/327790

### DIFF
--- a/share/templates/02-Choral/01-SATB/01-SATB.mscx
+++ b/share/templates/02-Choral/01-SATB/01-SATB.mscx
@@ -146,6 +146,18 @@
           <gateTime>67</gateTime>
           </Articulation>
         <Articulation name="sforzato">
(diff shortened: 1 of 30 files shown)


Jojo-Schmitz:
Note to self: rebase needed.

Jojo-Schmitz:
May resolve #10336

PR #9103, opened 2021-09-10
Added engraving read context (Step 1)

igorkorsukov:
Step by step
Some improvements to the implementation of the engraving domain model

The ultimate goal is to remove the `EngravingObject::score` method, it is very bad, it allows us to wherever do anything.

A step towards this is adding read context.
In fact, the reading context is already there, but not explicitly, the Score and Xml reader was used as the read context.
There will now be an explicit read context
The implementation of the read context refers to Score, but this is optional and may be changed in the future.

Also, the read context shrinks the interface to the minimum, instead of providing a huge Score interface.


src/engraving/accessibility/accessiblescore.cpp
```diff
@@ -110,7 +110,7 @@ size_t AccessibleScore::accessibleChildCount() const
     for (const EngravingObject* obj : m_score->EngravingObject::children()) {
         if (obj->isEngravingItem()) {
             AccessibleItem* access = Ms::toEngravingItem(obj)->accessible();
-            if (access && access->isAvalaible()) {
+            if (access && access->registred()) {
```
Marr11317, line 113: registered

src/engraving/accessibility/accessiblescore.cpp
```diff
@@ -125,7 +125,7 @@ const IAccessible* AccessibleScore::accessibleChild(size_t i) const
...
```
Marr11317, line 128: dito

src/engraving/compat/read114.cpp
```diff
@@ -2828,7 +2829,7 @@ Score::FileError Read114::read114(MasterScore* masterScore, XmlReader& e)
                 if (e.name() == "tempo") {
                     int tick   = e.attribute("tick").toInt();
                     double tmp = e.readElementText().toDouble();
-                    tick       = (tick * MScore::division + masterScore->_fileDivision / 2) / masterScore->_fileDivision;
+                    tick       = ctx.fileDivision(tick);
```
Marr11317, line 2832: isn't there some calculations lost here?

src/engraving/libmscore/ledgerline.cpp
```diff
@@ -46,6 +46,10 @@ LedgerLine::LedgerLine(Score* s)
     _next       = 0;
 }
 
+LedgerLine::~LedgerLine()
```
Marr11317, line 49: are empty destructors recommended in the style guide?

igorkorsukov: no, the code style guide doesn't formalize it.
I added an empty destructor to put a breakpoint and see where these objects are being deleted, to see why so many of them are created and deleted during layout
https://github.com/musescore/MuseScore/pull/9100

src/engraving/compat/read114.cpp
igorkorsukov: no, it's ok, this method will be called https://github.com/musescore/MuseScore/blob/master/src/engraving/libmscore/score.h#L886

src/engraving/accessibility/accessiblescore.cpp
igorkorsukov: Thank you, fixed.

vpereverzev (approved):

PR #9104, opened 2021-09-10
fix #8971 Tie directions on multivoice chords

### LINKED ISSUE #8971: [MU4 Issue] Auto tie direction incorrect when splitting into two voices
This is admittedly rather niche, but:

When the note on the left side of a tie is in voice 1 and has its stem down, and there are no extra voices, and a second voice is introduced under the note on the right side of the tie, 'Tie direction: Auto' should result in a tie that slopes _up_, rather than down:

auto-tie

**Should fix:**
https://musescore.org/en/node/254426

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/8971

Originally, if one or both chords on either side of the tie were affected by other voices, the stem direction for the first chord was used. This has been changed to use the stem direction of the chord affected by other voices.
image

### DIFF
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -593,9 +593,11 @@ void Tie::calculateDirection()
     if (_slurDirection == Direction::AUTO) {
         std::vector<Note*> notes = c1->notes();
         size_t n = notes.size();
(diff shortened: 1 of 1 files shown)


DmitryArefiev:
Checked #8971 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

PR #9105, opened 2021-09-11
Fix #8986: WorkspacesDialog size was increased and it is resizable no…

### LINKED ISSUE #8986: [MU4 Issue] Workspaces dialog is too short, can't be resized and doesn't show a scrollbar
**Bug:**
Workspace dialog is too small. If you add a new workspace option, you can't see it and it is hard to navigate to it due to the lack of a scrollbar. (See image below)

**Desired behaviour**
- Workspaces should default to a taller size - similar to the parts dialog
- Scrollbars should be visible when there is a need for them
- Users should be able to resize panel (as with 'Preferences')

image

MacOS, Intel

Josempb9:
…w. Also, a scrollbar was added to the list of workspaces

### DIFF
--- a/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
+++ b/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
@@ -34,6 +34,10 @@ RadioButtonGroup {
     boundsBehavior: Flickable.StopAtBounds
     orientation: Qt.Vertical
 
+    ScrollBar.vertical: StyledScrollBar {
+        policy: ScrollBar.AsNeeded
+    }
+
     property int leftPadding: 0
 
     Connections {
--- a/src/workspace/qml/MuseScore/Workspace/WorkspacesDialog.qml
+++ b/src/workspace/qml/MuseScore/Workspace/WorkspacesDialog.qml
@@ -31,8 +31,9 @@ import "internal"
 StyledDialogView {
     id: root
 
-    contentWidth: 552
-    contentHeight: 286
+    contentWidth: 664
+    contentHeight: 558
+    resizable: true
 
     WorkspaceListModel {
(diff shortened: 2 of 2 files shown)


src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
```diff
@@ -34,6 +34,10 @@ RadioButtonGroup {
     boundsBehavior: Flickable.StopAtBounds
     orientation: Qt.Vertical
 
+    ScrollBar.vertical: ScrollBar {
```
cbjeukendrup, line 37: I think it's better to use our `StyledScrollBar` component instead, which will have the correct look and behaviour.

Josempb9: Hi @cbjeukendrup, thanks for pointing this out. 
I have updated the QML file accordingly and amended the commit with the new changes. I noticed that it appears with a different commit ID. If you want me remove this pull request and start from scratch that also fine for me.

RomanPudashkin (approved):

src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
cbjeukendrup: Thank you, it is perfect! We prefer amending existing PRs instead of closing and creating new ones, to keep the PR history clean. The different commit ID is no problem (the content of the commit is indeed different, so it will have a different ID). 

And congratulations on having your first PR merged!

Josempb9: Thanks @cbjeukendrup. I hope I can contribute with other things in the upcoming weeks.
Keep the great work  !

Issue #7640, opened 2021-03-06
[MU4 Issue] zoom in and out on trackpad doesn`t work


vshalkevich:
**Describe the bug**
Zoom in and zoom out doesn`t work with trackpad (screenshot attached)(Bug also reproduces on Score page)

**To Reproduce**
1. Start creating a new score
2. Select "Choose from template"
3. Try to zoom in or zoom out preview page with trackpad

**Expected behavior**
Zoom in and zoom out should work with trackpad

**Screenshots**
![5 - Actual result (The page with a bug)](https://user-images.githubusercontent.com/71372310/110192322-58557600-7e3e-11eb-96ff-5c67aa71634f.png)

**Desktop (please complete the following information):**
Windows 10 (works now)
MacOS (**DOESN'T WORK NOW**)

**Additional context**

vshalkevich:
Issue reproduced again

cbjeukendrup:
I'm working on the fix for macOS. On Windows, it should already work.

DmitryArefiev:
Confirm. Still doesn't work on Mac. 
(I've edited the description)

PR #9108, opened 2021-09-12
Implemented zooming the Notation View by Pinch Gestures

### LINKED ISSUE #7879: [MU4 Issue] Zooming and scrolling in one part affects all parts
**Describe the bug**
Zooming and scrolling in one part affects all parts

**To Reproduce**
1. Create a score
2. Create a part
3. Zoom in (or scroll) on a main score
4. Open previously created part

**Expected behavior**
Zoom in / out works separately for main score and every part

**Screenshots**

Linux Ubunty

**Additional context**

cbjeukendrup:
Resolves: #7640 
(for both the regular NotationPaintView and the TemplatePaintView)

We need to check if this does not break zooming on operating systems other that macOS.

Todo: when zooming using the trackpad, this is not reflected by the zoom field in the Status Bar. I will fix this in a next PR, where I will also move the information about the view state to (a struct inside) the Notation class. This way, each Notation will have its own view state, which would be a fix for #7879.

### DIFF
--- a/src/engraving/libmscore/mscore.cpp
+++ b/src/engraving/libmscore/mscore.cpp
@@ -107,7 +107,6 @@ int MScore::pedalEventsMinTicks;
 bool MScore::harmonyPlayDisableCompatibility;
 bool MScore::harmonyPlayDisableNew;
 bool MScore::playRepeats;
-bool MScore::panPlayback;
 int MScore::playbackSpeedIncrement;
 qreal MScore::nudgeStep;
 qreal MScore::nudgeStep10;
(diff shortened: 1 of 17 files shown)


RomanPudashkin (approved):

vshalkevich:
Need somebody to check on Linux with touchpad. On my Linux virtual machine touchpad doesn`t work

Tantacrul:
I'm so looking forward to this working on my mac!!!

DmitryArefiev:
> 

> Need somebody to check on Linux with touchpad. On my Linux virtual machine touchpad doesn`t work

The same thing with my Linux machine. Zoom in/out with touchpad  doesn't work in any application. As I found for now, Touchpad gestures should be installed separately. Investigating now..

-----------------------------
On Mac10.15.6 and Win10, zooming in/out and scrolling with 2 fingers work fine

ModernMozart:
> Need somebody to check on Linux with touchpad. On my Linux virtual machine touchpad doesn`t work

Most Linux distros don't have touchpad gestures, there are a couple exceptions, but the vast majority do not

PR #9109, opened 2021-09-13
Moving pianoroll changes into new PR.

blackears:

### DIFF
--- a/src/engraving/libmscore/engravingitem.cpp
+++ b/src/engraving/libmscore/engravingitem.cpp
@@ -844,7 +844,7 @@ void EngravingItem::removeAnimationTrack(const std::string& propertyName)
     }
 }
 
-bool EngravingItem::hitShapeContains(const PointF & p) const
+bool EngravingItem::hitShapeContains(const PointF& p) const
 {
     return hitShape().contains(p - pagePos());
 }


src/engraving/libmscore/automationvertex.h
```diff
@@ -0,0 +1,48 @@
...
+#include "engravingitem.h"
+#include "mscore.h"
+
+namespace Ms {
+class AutomationTrack;
+
+class AutomationVertex : public EngravingItem
+{
+    double _value;
+    double _ticks;  //Number of whole notes offset from start of score
```
wizofaus, line 35: how will this work if the user removes or inserts a time range (measure etc.) into the music - will all existing vertices be adjusted accordingly?
All ticks can just be integers no? (even _value might be better as a float or qreal, I haven't seen a lot of doubles being used so far)

blackears: I was planning for the vertices to be stored in a list sorted by _ticks.  _ticks could be an integer or something else, but if it is an integer then we will need to determine what unit one tick is.  We could use the 1/480 of a whole note MS3 used - is that unit still relevant?

wizofaus: Yeah I assumed that's what it was meant to be!
But what I meant is that if you had, say, vertices defined for ticks 480 to 4800 then the user removed the first (4/4) measure entirely, would that modify all extract vertices in all staffs to subtract 480 from the "tick" values?
What if the user removed measures at the end of the piece that meant the vertex tick value was 
now beyond the end of the piece? (I would think you couldn't just delete the vertex if it was defining a slope...)

blackears: If a measure is deleted, I would presume any coresponding vertices would be deleted too and it would be up to the user to fix things if it caused a problem. Probably something to ask Martin about, though.

src/pianoroll/view/pianorollview.cpp
```diff
@@ -523,6 +525,58 @@ void PianorollView::drawDraggedNotes(QPainter* painter)
         return;
     }
 
+    if (m_dragStyle == DragStyle::EVENT_LENGTH || m_dragStyle == DragStyle::EVENT_MOVE
```
Jojo-Schmitz: One trailing space too much (for the code style check to pass)

RobFog:
A rebase is needed.

blackears:
Looks like Github just upgrades its security.  I need to change all my keys first.

Jojo-Schmitz:
One unit test failed, see https://github.com/musescore/MuseScore/runs/5579586289?check_suite_focus=true#step:8:1231

src/engraving/libmscore/property.h
```diff
@@ -392,6 +392,9 @@ enum class Pid {
     TEMPO_CHANGE_TYPE,
     TEMPO_EASING_METHOD,
 
+    PAN,
```
Jojo-Schmitz, line 395: There's a mismatch between the place where these 2 lines got added and the corresponding lines in properties.cpp.
Moving these 2 here to above `TEMPO_CHANG_TYPE or(mo moving the corresponding lines in proerties.cpp down) might resolve the crash in the unit test

Jojo-Schmitz: And indeed it does ;-)

scorster:
Just adding a link to PRE bug reports and requests in the MuseScore discussion area.

https://github.com/orgs/musescore/discussions/16925

bkunda:
Closing this one as the piano roll will need to be developed internally.

PR #9112, opened 2021-09-13
Fixed scaling for 100% zoom

cbjeukendrup:
When zooming to 100% in an A4 score, the page size on screen is exactly as big as a physical sheet of A4 paper. Same as in MU3.

Tested on macOS on the builtin display, both when running in normal mode and in low resolution mode (which can be enabled by showing the app in Finder -> right click -> show info -> open in low resolution)
Schermafbeelding 2021-09-13 om 19 35 03


RomanPudashkin (approved):

vshalkevich:
Works fine for me

PR #9127, opened 2021-09-15
Improved engraving object tree

igorkorsukov:


RobFog:
Typo in commit message: "imporoved" should be "improved".

igorkorsukov:
> Typo in commit message: "imporoved" should be "improved".

Thank you, fixed.

src/engraving/libmscore/volta.cpp
```diff
@@ -65,8 +65,8 @@ static const ElementStyle voltaStyle {
 //   VoltaSegment
 //---------------------------------------------------------
 
-VoltaSegment::VoltaSegment(Spanner* sp, Score* s)
-    : TextLineBaseSegment(ElementType::VOLTA_SEGMENT, sp, s, ElementFlag::MOVABLE | ElementFlag::ON_STAFF | ElementFlag::SYSTEM)
+VoltaSegment::VoltaSegment(Spanner* sp, EngravingItem* parent)
```
vpereverzev, line 68: maybe it makes sense to require only one of these arguments?

like (Spanner* parent)

or (EngravingItem* parent) + explicit local check that parent is "Spanner"

igorkorsukov: You right, I will fix it

igorkorsukov: I will fix it in a next PR

vpereverzev (approved):

PR #9137, opened 2021-09-15
fix #9007 - Invisible key/time signatures having adverse effect on spacing

### LINKED ISSUE #9007: [MU4 Issue] Invisible key/time signatures cause spacing inconsistencies
Here the lower stave (with invisible items) should exactly match the upper (where those items are simply absent):

invisible-signatures

The invisible items seem unproblematic when at the start of a system, whereupon they align to the first note in the bar, but otherwise cause problems (singly and in combination).

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/9007

There were a lot of TODO comments surrounding this issue, so it feels good to finally start calculating this stuff the correct way.
image
(hidden elements invisible on the left, visible on the right)

There is also code that needs to be fixed that concerns practice mode, but that can be fixed in a future issue that will involve a giant refactor of how practice mode works (because right now there's a lot of copy and pasted code)

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -3695,12 +3695,12 @@ void Measure::stretchMeasure(qreal targetWidth)
     std::multimap<qreal, Segment*> springs;
 
     Segment* seg = first();
-    while (seg && !seg->enabled()) {
+    while (seg && (!seg->enabled() || seg->allElementsInvisible())) {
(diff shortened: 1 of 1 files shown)


DmitryArefiev:
Checked #9007 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

RomanPudashkin (approved):

PR #9140, opened 2021-09-15
fix edge case from #9086, inner hidden staves affect bracket span

Nick-Mazuk:
After [#9086](https://github.com/musescore/MuseScore/pull/9086) was merged, @DmitryArefiev pointed out a bug in the implimentation:

https://github.com/musescore/MuseScore/pull/9086#issuecomment-919983727

The original PR was to select correct brace symbol for each system independently. This mean that hidden staves should not affect bracket size. The original PR worked when outer staves were hidden, but not inner staves.

This PR resolves that bug such that hidden inner staves don't affect bracket size.

Tested this PR with the following file: [bracket-staves.mscz.zip](https://github.com/musescore/MuseScore/files/7173614/bracket-staves.mscz.zip)


DmitryArefiev:
Checked #9140 on Win10, Mac10.15.7, Linux Mint 20.2 - FIXED

Please promote to master

RomanPudashkin (approved):

PR #9156, opened 2021-09-16
Improvement engraving tree  

igorkorsukov:


thirdparty/KDDockWidgets/src/private/multisplitter/Item.cpp
```diff
@@ -714,6 +714,7 @@ void Item::setGeometry(QRect rect)
 
 void Item::dumpLayout(int level)
 {
+    return;
```
vpereverzev, line 717: have you forgotten that?

vpereverzev (approved):

thirdparty/KDDockWidgets/src/private/multisplitter/Item.cpp
igorkorsukov: no, I'm bored with the output of this dump when the application starts, the console is heavily littered

PR #9158, opened 2021-09-16
fix #8944 - Add right margin to system trailers

### LINKED ISSUE #8944: [MU4 Issue] System-final time/key signature should have padding to right
Time signatures and key signatures at the very end of a system, after the last barline, currently finish exactly flush with the stave, but there should be a small amount of padding to the right (1sp will do for now, but this will eventually be configurable).

Currently:
key-signature 
key-signature2
time-signature

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/8944

image

Added a new style setting (not currently exposed in the UI) and documented it in https://github.com/musescore/MuseScore/wiki/Style-Parameter-changes-between-3.6-and-4.0

### DIFF
--- a/src/engraving/libmscore/segment.cpp
+++ b/src/engraving/libmscore/segment.cpp
@@ -2273,6 +2273,9 @@ qreal Segment::minRight() const
     if (isClefType()) {
         distance += score()->styleP(Sid::clefBarlineDistance);
     }
+    if (trailer()) {
+        distance += score()->styleP(Sid::systemTrailerRightMargin);
+    }
     return distance;
 }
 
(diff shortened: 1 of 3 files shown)


Jojo-Schmitz:
IMHO 1.0 sp is quite a lot, maybe we'd better take the same default value as for `clfefLeftMargin?, `0.8`, `timesigLeftMargin, `, `0.63` or `keysigLeftMargin`, `0.5` (seems `timesigLeftMargin, `, `0.63` would be a nice compromise), which should make the courtesy clef, time- or key signature look nicely centered?

asattely:
going to reduce the default margin rq, don't merge quite yet!

oktophonie (approved):

DmitryArefiev:
Assigning to @vshalkevich

DmitryArefiev:
Checked #8944 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

FYI @Tantacrul @oktophonie Comparison old and new distance:


https://user-images.githubusercontent.com/90187801/135280154-b500b293-3469-4085-9153-a86b838cdd50.mp4

Issue #9139, opened 2021-09-15
[MU4 Issue] Ambitus items overlays in Properties tab


vshalkevich:
**Describe the bug**
Ambitus items overlays in Properties tab (screenshot with bug added)

**To Reproduce**
1. Create a score 
2. Add Ambitus on the score (Palettes tab > Ambitus > drag&drop Ambitus on the score)
3. Select Ambitus
4. Open Propertiest tab

**Expected behavior**
Ambitus items displayed in Properties tab

**Screenshots**
![Ambitus](https://user-images.githubusercontent.com/71372310/133508950-b58cf6c2-eff6-44cb-a245-9ee897410ef6.jpg)

**Desktop (please complete the following information):**
Windows 10
MacOS

cbjeukendrup:
If nobody else is working on this yet, I'm working on it :)

Issue #8497, opened 2021-07-02
[MU4 Issue] New Score Wizard - Family and Instrument columns should be re-sized


jessjwilliamson:
**Describe the bug**
As shown on Figma, the family column is 200 pixels wide. The builds are showing it's 246 wide.
Reducing the width back to 200 can also improve spacing for the Instruments family. So with the family column being 200 pixels, this results that the Instruments family will be 272 pixels wide which will be very beneficial as there are dropdown combo boxes that are struggling to fit in right now - for example, we have a combo dropdown that needs to fit space for an instrument that has '13-course' inside (see screenshot). 

**Expected behavior**
![Screenshot 2021-07-02 at 10 57 16](https://user-images.githubusercontent.com/71656189/124257493-52d61700-db24-11eb-9e70-d08aef9b9257.png)
![Screenshot 2021-07-02 at 11 10 56](https://user-images.githubusercontent.com/71656189/124259276-3fc44680-db26-11eb-94c2-9483341cd55e.png)

jessjwilliamson:
Hi, I just checked the nightly build, and the sizes are incorrect. The Family column is very close, it's showing 197 pixels wide instead of 200 pixels. The Instruments column is showing 246 instead of 272 as I requested. It's showing 'Your score' 3rd column to be extremely wide now because of this. Please check again and update this. thanks :)

*Expected behaviour*
<img width="985" alt="Screenshot 2021-07-13 at 13 19 57" src="https://user-images.githubusercontent.com/71656189/125450432-bf523cb5-fe53-45a8-9215-9b1dec53b902.png">

*Screenshots*
<img width="1332" alt="Screenshot 2021-07-13 at 13 17 16" src="https://user-images.githubusercontent.com/71656189/125450113-a8c30139-2f0b-4dfb-b108-5ba608629b3e.png">

jessjwilliamson:
Hi, @RomanPudashkin @Eism 
I want to bump this issue again, can we please make sure the Instruments column is wider, please. Like a selected state on an instrument would be 276 pixels wide. Additionally, this also solves long-worded instruments (as shown in the screenshot) without the '...' continuous issue and a dropbox can fit next to '12 Hole chromatic harmonica'. For the screenshot, I have re-sized the NSW window on purpose to demonstrate the desired sizing for the default window. 
[...]

PR #9160, opened 2021-09-16
Fix the width of the columns on the Choose Instruments Page (NSW) + other issues

cbjeukendrup:
Resolves: #8497 
Schermafbeelding 2021-09-17 om 00 48 38

Resolves: #9139 
Schermafbeelding 2021-09-17 om 00 43 38

And introduces some code quality improvements.

**To do:** 
- The icons to select the note head type are somewhat small (@Tantacrul?);
- Selecting the pitches for the ambits is very buggy.


RomanPudashkin (approved):

Tantacrul:
About those noteheads: I agree they're a hair small looking (both here and elsewhere). I'm not 100% sure how they are drawn. Could we bump them all up by around 20%? May take a little trial and error, since some are way bigger than others. To really get it right, we'd need to specify slightly different sizes for different noteheads.
image


Also, I notice that you never see scrollbars in those notehead options, so you can't tell that there are more noteheads out of view! That's another important one.

jessjwilliamson:
> About those noteheads: I agree they're a hair small looking (both here and elsewhere). I'm not 100% sure how they are drawn. Could we bump them all up by around 20%? May take a little trial and error, since some are way bigger than others. To really get it right, we'd need to specify slightly different sizes for different noteheads.
> image
> 
> Also, I notice that you never see scrollbars in those notehead options, so you can't tell that there are more noteheads out of view! That's another important one.

I'm seeing these icons look quite blurry.. are they imported SVGs or from font? Would like to investigate to make them look sharper :)

jessjwilliamson:
> Regarding the scrollbars: I just added them in this PR :)
> However, like all scrollbars in the app, they only become visible when the user already scrolls, or hovers the area where the scrollbar would be.
> 
> Regarding the notepad groups icons: they are indeed SVGs. For some reason, QML won't render them beautifully. We could try to use them directly from the music font (Leland).
> And I think the size of the note head _group_ icons is okay, but the note head _type_ icons are too small. Those are from the MusescoreIcon font.
> Schermafbeelding 2021-09-17 om 13 08 18

I think they were not blurry in the past.. this issue is also on Beam types. Do you think because they're SVGs the render is causing this way?

Tantacrul:
Ah. Regarding the notehead types, they should be set to 24pt (assuming they're 16pt now?)

Tantacrul:
And, yes, we should replace the SVG noteheads. I'll have to port them from Leland. It'll take a bit of time. Perhaps a job for @jessjwilliamson :)

jessjwilliamson:
> And, yes, we should replace the SVG noteheads. I'll have to port them from Leland. It'll take a bit of time. Perhaps a job for @jessjwilliamson :)

Sure thing!

vshalkevich:
This PR works fine for me

PR #9178, opened 2021-09-17
Removed old engraving tree implementation

igorkorsukov:


src/engraving/libmscore/beam.cpp
```diff
@@ -2728,7 +2728,7 @@ void Beam::startDrag(EditData& editData)
 
 void Beam::scanElements(void* data, void (* func)(void*, EngravingItem*), bool all)
 {
-    ChordRest* cr = toChordRest(treeParent());
+    ChordRest* cr = toChordRest(_elements[0]);
```
vpereverzev, line 2731: makes sense to check _elements emptiness

vpereverzev (approved):

Issue #9043, opened 2021-09-06
[MU4 Issue] [VJ] Music should automatically clear headers and footers


oktophonie:
A footer is positioned relative to the bottom page margin, according to the setting in 'Style -> Text Styles -> Footer -> Offset -> Y'. (The default bottom margin is 15mm, the default footer-y-offset is 5mm, so the footer appears 10mm above the bottom of the page.)

The value in 'Style -> Page -> Music bottom margin' determines the distance between the bottom page margin and the music on the bottom stave (measuring to the lowest point of the music on the stave, not to the stave itself).

When the footer is tall (for example if it has several lines) there is a risk of it colliding with the music on the bottom stave, unless the 'Music bottom margin' is increased accordingly.

There is a relatively simple solution to avoid these collisions automatically. If the footer itself extends inside the margin (i.e. if its height is, with default settings, more than 5mm) then the 'Music bottom margin' value should be measured from the top of the footer, not from the bottom page margin.

For consistency this should be applied to the header/'Music top margin' in the same way. (With the default styles, this means the measurement will always be being made from the bottom of the header (page number), since this sits just inside the top margin, i.e. with a y-offset of 0.)

DmitryArefiev:
Checked #9043 on Win10, Mac11.13.1, Linux Mint 20.2 - DONE

Issue #9159, opened 2021-09-16
[MU4 Issue] Footer text disappears for multi-line text


Nick-Mazuk:
**Describe the bug**

Multi-line footers cause text to disappear.

**To Reproduce**
1. Create a new document
2. Go to Format > Style > Header/Footer
3. Start typing some text with line breaks
4. Watch as text disappears on score

**Expected behavior**

Footer text should not disappear. It should be displayed exactly as it is in the settings.

**Screenshots**

https://user-images.githubusercontent.com/5659171/133680750-1b1cdb9f-b844-488a-9b7c-a155f9f68e33.mov

**Desktop (please complete the following information):**
 - OS: macOS 11.6.0

**Additional info:**

I have not checked if the same bug happens to other footers and headers.

Also, I'm working on https://github.com/musescore/MuseScore/issues/9043 which involves much of the same code. If someone works on this issue before a PR for https://github.com/musescore/MuseScore/issues/9043 is merged, there will be merge conflicts.

DmitryArefiev:
Checked  #9159 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

PR #9193, opened 2021-09-20
Header footer margin

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/9043

This PR removes collisions between headers/footers and music. When header or footer text becomes too tall, the music will automatically move out of the way.

**Note**: If when testing this header/footer text disappears, this is behavior existed before this PR and is documented in https://github.com/musescore/MuseScore/issues/9159

Header

Footer


Jojo-Schmitz:
Some test failures, crashes actually

Nick-Mazuk:
Fixed the failing tests

Jojo-Schmitz:
Nope

Jojo-Schmitz:
For me it jcrashes whan opening an existins score, and with failed assertion:

Fatal: ASSERT: "!strcmp(MStyle::valueType(idx),"Ms::Spatium")" in file .../libmscore/score.h, line 896

src/engraving/style/styledef.cpp
```diff
@@ -1096,7 +1096,7 @@ const StyleDef::StyleValue StyleDef::styleValues[int(Sid::STYLES)] = {
     { Sid::headerFontStyle,               "headerFontStyle",              1 },
     { Sid::headerColor,                   "headerColor",                  QVariant::fromValue(draw::Color::black) },
     { Sid::headerAlign,                   "headerAlign",                  QVariant::fromValue(Align::CENTER | Align::TOP) },
-    { Sid::headerOffset,                  "headerOffset",                 PointF() },
+    { Sid::headerOffset,                  "headerOffset",                 PointF(0.0, 0.0) },
```
Jojo-Schmitz, line 1099: This change should not be needed

src/engraving/layout/layoutpage.cpp
```diff
@@ -102,7 +102,8 @@ void LayoutPage::collectPage(const LayoutOptions& options, LayoutContext& ctx)
 {
     const qreal slb = ctx.score()->styleP(Sid::staffLowerBorder);
     bool breakPages = ctx.score()->layoutMode() != LayoutMode::SYSTEM;
-    qreal footerExtension = qMax(0.0, ctx.page->footerHeight() - ctx.score()->styleP(Sid::footerOffset)); // how much the footer extends above the margin
+    qreal footerExtension = qMax(0.0, ctx.page->footerHeight() - ctx.score()->styleV(Sid::footerOffset).value<PointF>().y()); // how much the footer extends above the margin
+    qreal headerExtension = qMax(0.0, ctx.page->headerHeight() - ctx.score()->styleV(Sid::headerOffset).value<PointF>().y()); // how much the header extends below the margin
```
Jojo-Schmitz, line 106: Yes, this make much more sense now.

Jojo-Schmitz:
Still fails engraving_tests.

src/engraving/style/styledef.cpp
Nick-Mazuk: This will be reverted in next commit.

DmitryArefiev:
Hi @Nick-Mazuk , should it work like this?
01  clear headers and footers

Nick-Mazuk:
> Hi @Nick-Mazuk , should it work like this?

No, that's a bug. Thanks for finding it, I'll fix it!

DmitryArefiev:
> 
> 
> > Hi @Nick-Mazuk , should it work like this?
> 
> No, that's a bug. Thanks for finding it, I'll fix it!

Cool! Then just assign to me when it's ready

Nick-Mazuk:
Going to potentially make further updates to clean up how everything works. Here's an issue in case anyone wants to comment: https://github.com/musescore/MuseScore/issues/9340

Nick-Mazuk:
A lot changed since the original version of the PR. Many bugs have been fixed, and behavior is slightly different. I've updated the original post on the PR with the new details (to keep things more organized).

@DmitryArefiev Fixed the bug you found. I can't seem to assign this to you, though. When testing, can you test many different places that text is rendered? I had to update the base text class, so there may be side effects on any text. I haven't found any yet, but something to be aware of.

Jojo-Schmitz:
As far as I can tell this works fine for foooters but does nothing for headers?

Jojo-Schmitz:
Actually it does work for headers too, but not if there's a vertical frame. Should it?

DmitryArefiev:
The same on my side with a vertical frame.
And another thing, it doesn't respect a page breaks:
Footers and headers

DmitryArefiev:
@Nick-Mazuk Do you want the two issues above to be logged separately? (we can't promote the changes then) 

(#9159 also looks good)

Nick-Mazuk:
@Jojo-Schmitz @DmitryArefiev, I've fixed the collision with the vertical frame.

> And another thing, it doesn't respect a page breaks:

@DmitryArefiev, this is the desired behavior. Under Format > Style… > Page, you'll notice the setting "Max System Difference" (set to 15sp by default). When systems are spaced out evenly, the distance between them will not exceed this value. So when you have tall systems and a tall footer (like in your example), the systems will not reach the top of the footer because the required space would exceed the max distance. I suspect if you increase this setting value, you'll get the results you were expecting.

Jojo-Schmitz:
Looks good

DmitryArefiev:
> @DmitryArefiev, this is the desired behavior. Under Format > Style… > Page, you'll notice the setting "Max System Difference" (set to 15sp by default). When systems are spaced out evenly, the distance between them will not exceed this value. So when you have tall systems and a tall footer (like in your example), the systems will not reach the top of the footer because the required space would exceed the max distance. I suspect if you increase this setting value, you'll get the results you were expecting.
@Nick-Mazuk  As I understand, the page break should lock the amount of systems on one particular page. So the footer should not move system with a page break to a next page. 
But I guess my case above with a such tall footer text is a very rare workflow so let's promote it now

DmitryArefiev:
Checked #9043, #9159 on Win10, Mac11.13.1, Linux Mint 20.2 - DONE

Ready to be promoted (if Code Review has been done)

Nick-Mazuk:
> As I understand, the page break should lock the amount of systems on one particular page. So the footer should not move system with a page break to a next page.
> But I guess my case above with a such tall footer text is a very rare workflow so let's promote it now

Oh, I see what you mean now. From the video, that is the behavior I'd expect, though I can see how someone would not expect that behavior. I'll double-check with Simon when I speak with him tomorrow to confirm the desired result.

Jojo-Schmitz:
The Merge commit doesn't belong into this (or any) PR

oktophonie:
The behaviour with regard to page breaks seems as I would expect: a page break simply means that the next system will be on the following page, but it can't force more systems to fit on the preceding page than there is space for.

There's scope for some sort of "keep systems together on page" functionality in the future (analogous to "keep bars together on system") but that would be an entirely separate issue from this.

DmitryArefiev:
Got it. Thanks for review Simon! 

Lets merge it into master

DmitryArefiev (approved):

oktophonie (approved):

Issue #9194, opened 2021-09-20
[MU4 Issue]  Back-up  never happens


wizofaus:
**Describe the bug**
If MuseScore crashes at any time while trying to save a file, even during AutoSave, the mscz is left unusable, and no recovery appears possible (no backup file created).

**To Reproduce**
1. No single good way to reproduce except using the debugger and setting breakpoints/terminating process part way through saving.  Though it might be possible by using "remove empty trailing measures" before saving.

**Expected behaviour**
Original file should remain untouched until saving has finished, then updated via a rename or copy. Or at least, a backup should be created before saving, so that any unexpected application abort during save doesn't mean losing already saved work.
Though obviously in an ideal world the application should never abort unexpectedly while saving!

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**
BTW while it appears to be coded to create a back-up every time it saves, in fact it never happens as `NotationProject::makeCurrentFileAsBackup()` exits early after determining that there's no valid "created()" property. But even after addressing this, the test `if (fileSystem()->exists(filePath))` after this causes another early exit - I'm pretty sure there's supposed to be a ! in there (i.e. if the file *doesn't* exist!).

wizofaus:
If it didn't crash then you can't have verified that a backup got created (or the file didn't get corrupted).
Unfortunately like I said it's hard to provide a reliable easy of having it crash while saving unless you use a debugger to stop/kill the process.

Tantacrul:
OK, so I've tested something related to this (I don't have the ability to force a save operation to crash)

1. Created a new score. Did not save it. Wrote a few notes. Used the 'Crash' button in Dev Tools to intentionally crash the app.
2. When reopening, I got this message:
<img width="349" alt="image" src="https://user-images.githubusercontent.com/21022311/166326356-6e00885d-1ff4-4cba-9036-405ebb522c1f.png">
3. I then reopened the file and continued writing. Crashed again. Reopened again
4. Since autosave only happens every few mins (I think), I often lost a bit of recently added information, which is fine / to be expected

@wizofaus - can you advise whether you think this should be reopened or whether there is another issue to be raised about this?

wizofaus:
The biggest problem is fixed - it now saves to a temporary file and only replaces the original file at the end, so even a crash during save won't corrupt the mscz file.

But the logic in  NotationProject::makeCurrentFileAsBackup() still looks wrong to me

    Ret ret = fileSystem()->exists(filePath);
[...]

PR #9200, opened 2021-09-20
Made save more safe

igorkorsukov:
https://github.com/musescore/MuseScore/issues/9194

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -328,8 +328,10 @@ mu::Ret NotationProject::saveScore(const io::path& path, SaveMode saveMode)
 
     Ret ret = doSave(true);
     if (!ret) {
-        ret.setText(Ms::MScore::lastError.toStdString());
-    } else if (saveMode != SaveMode::SaveCopy || oldFilePath == path) {
+        return ret;
+    }
+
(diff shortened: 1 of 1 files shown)


Eism (approved):

DmitryArefiev:
Just checked Auto Save feature. No crashes during testing (on Win10, Mac11.13.1, Linux Mint 20.2)

src/project/internal/notationproject.cpp
```diff
@@ -338,40 +340,60 @@ mu::Ret NotationProject::saveScore(const io::path& path, SaveMode saveMode)
...
 
-    MscWriter msczWriter(params);
-    Ret ret = writeProject(msczWriter, false);
-    if (!ret) {
-        LOGE() << "failed write project to buffer";
-        return ret;
+    // Step 3: create backup if need
+    {
+        if (generateBackup) {
+            makeCurrentFileAsBackup();
```
wizofaus, line 378: But the logic in `makeCurrentFileAsBackup` is still wrong, it doesn't appear to do anything?  Line 192 (effectively an early return if the current project file *does* exist) is surely not right...(but for me even the first test in that function `if (!created().val)` was succeeding, hence returning at that point before doing anything). @igorkorsukov

igorkorsukov: I renamed and reopened your issue

PR #9237, opened 2021-09-22
Fix #8952: Simplify/unify stem layout code and let it take SMuFL anchors of flag into account

### LINKED ISSUE #8952: [MU4 Issue] [Engraving] Note stems don't take the SMuFL anchors "stemUpNW" and "stemDownSW" into account

cbjeukendrup:
Resolves: #8952

Also fixes an issue that has existed for a longer time, where on a TAB staff, the width of the stem of the start chord of a beam influences the end position of the beam:
Schermafbeelding 2021-09-22 om 23 47 39

It would be nice if the vtests could be repaired... that would give a lot more confidence. But I tested it with all cases I could think of (normal staves, different kinds of TAB staffs; with and without beams; tremolos...)

### DIFF
--- a/src/engraving/layout/layoutchords.cpp
+++ b/src/engraving/layout/layoutchords.cpp
@@ -879,7 +879,7 @@ void LayoutChords::layoutChords3(const MStyle& style, std::vector<Note*>& notes,
             if (chord->stem()) {
                 chord->stem()->layout();
                 if (chord->hook()) {
-                    chord->hook()->rypos() = chord->stem()->hookPos().y();
+                    chord->hook()->rypos() = chord->stem()->flagPosition().y();
                 }
             }
         }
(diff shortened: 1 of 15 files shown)


DmitryArefiev:
Checked #8952 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED
(tested using the position of the grip box to make the stem longer or shorter)

RomanPudashkin (approved):

PR #9240, opened 2021-09-23
[MU3] Fixed linux schedule workflow on CI

igorkorsukov:


.github/workflows/ci_linux.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         T_ID=${{ secrets.TELEMETRY_TRACK_ID }}; if [ -z "$T_ID" ]; then T_ID="''"; fi
         YT_API_KEY=${{ secrets.YOUTUBE_API_KEY }}; if [ -z "$YT_API_KEY" ]; then YT_API_KEY="''"; fi
         sudo bash ./build/ci/linux/build.sh -n ${{ github.run_id }} --telemetry $T_ID --youtube_api_key $YT_API_KEY
-      name: Package
+    - name: Package
```
Jojo-Schmitz, line 98: What is that doing? I had no problems so far without that change

Jojo-Schmitz: Or is it because I didn't try to create a pagkage yet? (Well, I tried, but failed, maybe because of this?)

Jojo-Schmitz: Oh, that's the master branch, so sorry for the noise

igorkorsukov: I notice what Linux nightly build is broken for MU3
https://github.com/musescore/MuseScore/actions/runs/1264459893

I don't know when this happened, but here's a syntax error

Workflow for nightly builds MU3 is located in the master (only from the master, workflows can be started on a schedule). This workflow runs on a schedule and builds the `x.3` branch

Jojo-Schmitz: I see, thanks

Jojo-Schmitz: And whatever it was, it broke on July 8th, last Linux nightly is from July 7th

Jojo-Schmitz: It did for macOS too BTW, so that needs to get fixed too?

Jojo-Schmitz: Excact same error in .github/workflows/ci_macos.yml, line 96

Jojo-Schmitz: See #9256

PR #9280, opened 2021-09-24
re-add overshoot to arpeggios

Nick-Mazuk:
Fixes regression from earlier this week and re-adds the overshoot in arpeggios. Ultimately, this regression/fix is better than the previous solution as the code. The code is cleaner and we have more precise control over the overshoot amount. Before, a scaling factor was used to add the overshoot (e.g the arpeggios were X% larger). Now, the overshoot is precisely 0.25 spaces.

Current

image

Fixed

Screen Shot 2021-09-24 at 3 46 30 PM

### DIFF
--- a/src/engraving/libmscore/arpeggio.cpp
+++ b/src/engraving/libmscore/arpeggio.cpp
@@ -35,6 +35,7 @@
 #include "score.h"
 #include "symid.h"
 #include "staff.h"
+#include "stafflines.h"
 #include "part.h"
 #include "page.h"
 #include "segment.h"
@@ -141,9 +142,9 @@ void Arpeggio::read(XmlReader& e)
 
 void Arpeggio::symbolLine(SymId end, SymId fill)
 {
-    qreal y1 = -_userLen1;
-    qreal y2 = _height + _userLen2;
-    qreal w   = y2 - y1;
+    qreal top = calcTop();
+    qreal bottom = calcBottom();
+    qreal w   = bottom - top;
     qreal mag = magS();
     ScoreFont* f = score()->scoreFont();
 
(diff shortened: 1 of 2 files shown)


cbjeukendrup:
I don't think commit 809db11cedde0bb94c7c97507782759b0faeb690 was the problem, because in a commit way before it (a70cb6e9b27943213a48dc771d646e3995430541), it was already broken too.

Also, that scaling factor which I removed did not have any effect, because it was initialized as `painter->worldTransform().m11()`, which is the scaling that the painter already had anyway. 

Anyway, your solution looks good to me!

Nick-Mazuk:
> I don't think commit [809db11](https://github.com/musescore/MuseScore/commit/809db11cedde0bb94c7c97507782759b0faeb690) was the problem, because in a commit way before it ([a70cb6e](https://github.com/musescore/MuseScore/commit/a70cb6e9b27943213a48dc771d646e3995430541)), it was already broken too.

Thanks for the clarification!

Jojo-Schmitz:
Maybe it is rather "Add overshoot", or "Move overshoot up"? In 3.6.2 it is 'undershooting'.

Jojo-Schmitz:
Should it overshoot at the arrow or at the upper end?

Jojo-Schmitz:
I guess images are needed, before and after, for all arpeggions, to make this clear.

Jojo-Schmitz:
And again this PR shows that the vtests are not working...

Nick-Mazuk:
Going to make some tweaks to this later today. Not ready to merge.

Nick-Mazuk:
Updated PR to take care of many more edge cases. I also verified that the new vertical shifts don't cause any issues with accidentals (https://github.com/musescore/MuseScore/pull/9097).

arpeggios-ends

Test file:
[arpeggios-ends.mscz.zip](https://github.com/musescore/MuseScore/files/7240397/arpeggios-ends.mscz.zip)

oktophonie (approved):

Issue #9005, opened 2021-09-02
[MU4 Issue] Tie position changes when notes are flipped


oktophonie:
MuseScore draws ties 'outside' the notes when joining single notes (ties start and end at the top or bottom side of the notehead), but 'inside' the notes when joining notes in chords (ties start after the right side of the first note and end before the left side of the next).

However, if either the first note of a tied single-note pair is flipped, or the tie itself is flipped, such that the tie is no longer curving away from the stem, MuseScore switches from the 'outside' placement to the 'inside' placement. (The stem direction of the second note seems to have no bearing on the outcome.) This is an error.

![inside-outside-ties](https://user-images.githubusercontent.com/20604032/131821695-3ed58de3-3a5e-4552-8903-18e189829247.png)

Here the ties in _b_ (single notes flipped) are positioned the same as in _c_ (chords), but they should retain the same position as in _a_.

The current behaviour is perhaps because, when flipped, one end of the tie needs to be offset differently so as to not intersect a stem. This will be solved more properly in #8980.

Partial fix for: https://musescore.org/en/node/45576

PR #9281, opened 2021-09-25
Rewrite of inside tie endpoint code

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/9005

Previously, X adjustments to tie endpoints was being done before the Y adjustments, which could sometimes be fairly substantial, resulting in collisions that couldn't be accounted for. This PR moves all of the X adjustment code _after_ the Y adjustment, resulting in much nicer-looking ties.
image

They adjust better for collisions with accidentals:
image

They no longer adjust for hidden stems:
image

And now they also account for accidentals, noteheads, stems, and hooks in other voices in the same tick:
image

Barring the offset being parameterized down the line, this should be the last time we need to touch the tie endpoint horizontal adjustment code for inside-style ties. 

By virtue of applying this only to inside-style ties, https://github.com/musescore/MuseScore/issues/9005 has been fixed as a byproduct, as those are outside-style ties and have been left out.

### DIFF
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -37,6 +37,8 @@
 #include "chord.h"
 #include "hook.h"
 #include "ledgerline.h"
+#include "accidental.h"
+#include "stem.h"
 
 using namespace mu;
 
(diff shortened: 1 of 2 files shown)


asattely:
Whoops. I'll fix the utests on Monday.

asattely:
I lied, fixed it today

Jojo-Schmitz:
Those lies are accepted ;-)

src/engraving/libmscore/tie.h
```diff
@@ -82,6 +86,8 @@ class Tie final : public SlurTie
     Note* startNote() const;
     Note* endNote() const;
 
+    bool isInside() { return _isInside; }
```
Jojo-Schmitz, line 89: Better `bool isInside() conmst { return _isInside; }`?

src/engraving/libmscore/tie.cpp
```diff
@@ -501,7 +650,6 @@ void Tie::slurPos(SlurPos* sp)
     qreal x1, y1;
     qreal x2, y2;
     bool shortStart = false;
```
Jojo-Schmitz, line 652: As far as I can tell this variable is not used at all, except for being set here and a fe lines further down.

Jojo-Schmitz:
While playing with this (resp. its backport to 3.x, where it is well possible that I made mistakes) I got MuserScore to hang and repeatedly showing 
`Debug: ===SeqMsgFifo: overflow (...\mscore\seq.cpp:1547, void Ms::SeqMsgFifo::enqueue(const Ms::SeqMsg&))`

oktophonie (approved):

asattely:
Needs a bit more work, ties across system boundaries are currently breaking

Jojo-Schmitz:
Oops, indeed, 2nd example but with a system break:
image

asattely:
> Oops, indeed, 2nd example but with a system break:
> image
> If that is what you mean?
> Basically with ties over a system break we're still having the same problems as we had before this PR.
Yep, just forgot to adjust the second segment in the same way as the first. There is still a little bit of vertical weirdness for continued ties which will be addressed in a later issue

src/engraving/libmscore/tie.cpp
```diff
@@ -877,6 +964,7 @@ TieSegment* Tie::layoutBack(System* system)
 
     segment->layoutSegment(PointF(x, sPos.p2.y()), sPos.p2);
     segment->setSpannerSegmentType(SpannerSegmentType::END);
+    segment->adjustX();
```
Jojo-Schmitz, line 967: This line indeed fixes the issue with ties over system breaks

Jojo-Schmitz:
Reg. the build fail see #9299, sym,h is now symid.h

src/engraving/libmscore/tie.cpp
asattely: yep, easy fix. though it seems like we might need a rebase now because the build is failing, hold

Jojo-Schmitz: Yep, my PR #9299 should fix this

Jojo-Schmitz:
Now a rebase should fix the failures

asattely:
This one is now ready as far as I'm concerned

asattely:
added some top and bottom margin to left-hand notehead collision detection

worldwideweary:
Hey, is it possible you could check the situation with this code when a chord with a tie-in from a crushed/appogg exists? For example: CEG with an acciaccatura or appoggiatura as C tied into the C of CEG ... the tie is displaced incorrectly it looks like from this PR (found via using PR#9000 accumulation) This stuff is used for workarounds for repeat measures when tying into them etc.

Should be: 


But this is what comes out:




Found problem by reverting this along with vertical tie placement (that one should be fine) in PR#9000  If this could be fixed, it'd be much appreciated. Again, it doesn't happen when it's just one note as it seems, but only when the main anchored ChordRest is a chord with two or more notes. P.S. Thanks for the main premise of this PR

Issue #9264, opened 2021-09-24
[Crash] Close new score without saving


Tantacrul:
**Describe the bug**
If you create a new score, mark it dirty and then attempt to close it without saving, you get a persistent crash on Mac. See video:

https://user-images.githubusercontent.com/21022311/134659335-bdb0c5cf-3fc0-4c8a-ae4e-275e02f1cc71.mov

DmitryArefiev:
To reproduce this MuseScore should use **Dark Theme**

DmitryArefiev:
Checked #9264 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

PR #9295, opened 2021-09-27
Fixed crash on score close

igorkorsukov:
There is an implementation problem in the asynchronous communication system (channels, promises, etc.).
AbstractInvoker uses the CallBack structure by value. It may be that we put CallBack in the queue, but while CallBack is in the queue and has not yet been executed, we can delete it and, accordingly, we do not need to call it. Now, this moment has not been worked out properly.
I will work through this case later. The current solution solves the problem but is rather temporary.

### DIFF
--- a/thirdparty/deto_async/async/asyncable.h
+++ b/thirdparty/deto_async/async/asyncable.h
@@ -25,6 +25,8 @@ public:
         virtual void disconnectAsync(Asyncable* a) = 0;
     };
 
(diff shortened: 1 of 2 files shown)


vpereverzev (approved):

DmitryArefiev:
Checked #9264 on Win10, Mac11.13.1, Linux Mint 20.2 - FIXED

PR #9299, opened 2021-09-27
[MU4] Fix build

Jojo-Schmitz:
after #9097


Jojo-Schmitz:
vtests fail because the previous build fails, which is exactly what this PR is fixing...

Jojo-Schmitz:
Please merge this one ASAP

Issue #8367, opened 2021-06-21
[MU4 Task] [Accessibility] Note-input bar settings dialog


vpereverzev:
**Task description**
Replicate an existing approach on the note input bar settings dialog and the popup 

![2021-06-21 13_56_42-Screen Reader - Google Таблицы — Mozilla Firefox](https://user-images.githubusercontent.com/71372310/122751530-8ac59a80-d298-11eb-9c04-3e858285ec34.jpg)

![изображение](https://user-images.githubusercontent.com/71372310/122755365-949dcc80-d29d-11eb-8fea-9d2029cbc1fd.png)

See example in 37eaaec3789f0c9ec0b95d817a0634fc04c4a077

DmitryArefiev:
Checked with Narrator on Win10 - FIXED

Doesn't work properly with Orca on Linux (but can be related to my environment) - logged as #9353

PR #9304, opened 2021-09-27
[MU4] Accessibility. Note input customise dialog

Eism:
Resolves: #8367


igorkorsukov (approved):

DmitryArefiev:
Checked with Narrator on Win10 - FIXED

Doesn't work properly with Orca on Linux (but can be related to my environment) - logged as #9353

PR #9307, opened 2021-09-27
Added the possibility to use straight flags

cbjeukendrup:
https://user-images.githubusercontent.com/48658420/134992231-9a117989-0b8b-46c0-acdd-41b757583fca.mov

Note: currently, it is broken when using Leland, because the metadata file for Leland needs to be fixed, but @oktophonie is working on that. 

For fonts that do not support straight flags, Bravura is used as a fallback. (We may choose to implement a fallback chain, so that MuseScore first tries the font the user specified, then Leland, then Bravura.)


src/engraving/libmscore/symnames.cpp
```diff
@@ -6015,6 +6032,23 @@ constexpr const std::array<const char*, size_t(SymId::lastSym) + 1> SymNames::s_
     "Large brace",
     "Larger brace",
 
+    "Combining flag 8 (1024th) below straight",
```
Jojo-Schmitz, line 6035: Translatabilite of those whouel be needed, now that they became '1st class citizen'

cbjeukendrup:
Later today, I will implement that the Shadow Note follows the style of the score.

src/engraving/libmscore/symnames.cpp
cbjeukendrup: Why would that be needed? As far as I know, these names are never exposed in the UI. The only reason that I added them, is to make sure that the lists of names stay consistent. 

By the way, the names of the normal flags are not translated either. If the names for the straight flags need to be translatable, shouldn’t those for the normal ones be too?

Jojo-Schmitz: These names are exposed, at least in the Symbols palette of the master palette. But indeed if the normal flags ain't, these don't need it either, so sorry for the noise.

RomanPudashkin (approved):

Issue #8365, opened 2021-06-21
[MU4 Task] Instruments panel accessibility


vpereverzev:
**Task description**

Replicate an existing approach on "instruments panel"

See example in 37eaaec3789f0c9ec0b95d817a0634fc04c4a077

PR #9315, opened 2021-09-28
[MU4] Accessibility. Instruments panel

Eism:
Resolves: #8365


DmitryArefiev:
Checked #9315/#8365 with Narrator/NVDA on Win10 and Orca on Linux Mint 20.2 - FIXED

@Eism Please merge into master

Issue #9270, opened 2021-09-24
[MU4 Issue] Window doesn't match name of VST


jessjwilliamson:
**Describe the bug**
When open any VST it'll always say 'Musescore4Development' where it should be the name of the plugin.

**Screenshots**

![Window no VST name](https://user-images.githubusercontent.com/71656189/134674010-b9e9c499-ac3b-4da8-a417-c67bb0bfacdd.png)

**Desktop (please complete the following information):**
 - OS: Windows

Issue #9269, opened 2021-09-24
[MU4 Issue] Multiple VST windows open


jessjwilliamson:
**Describe the bug**
VST should only have one window per plugin. 
In this case, I have Kontakt, but 3 windows open because I am able to click on the Kontakt Sound button multiple times resulting in multiple windows. This must be only one window per plugin. 

![VST Multiple windows](https://user-images.githubusercontent.com/71656189/134670172-5f8307b2-78a7-467a-b457-be89832f057b.png)

Same for another plugin, to prove it's not just Kontakt.

![VST Multiple windows Super 8](https://user-images.githubusercontent.com/71656189/134670203-490445d7-1739-4332-9e0a-742d499fc6d8.png)

**Desktop (please complete the following information):**
 - OS: Windows

RobFog:
There's a typo in the issue title: "Mulitple".

Issue #9248, opened 2021-09-23
[MU4 Issue] Keep scrollbar in Mixer visible at all times


jessjwilliamson:
**Describe the bug**
Request from both me and Martin. We believe to keep the scrollbar perpetually visible in the mixer window. 

**Expected behavior**
![image](https://user-images.githubusercontent.com/71656189/134535007-62a3971b-a051-4697-86e5-1b279c60c21e.png)

**Desktop (please complete the following information):**
 - OS: MacOS

Issue #9242, opened 2021-09-23
[MU4 Issue] Volume numbering in Mixer should only have tenths after decimal point


jessjwilliamson:
**Describe the bug**
The numbering has 2 numbers after the decimal point which is not very necessary. This should be simplified.

**Expected behavior**

What would be easier is to only have the 'tenth' number or just one number after the decimal point in volume number text fields.
For example, the mixer volume currently are having numbers like 0.42, 1.34, 8.24 etc. 
What would be more simplified is to have number values such as 0.4, 1.3, 8.2 etc. Just only one number after the decimal point.
![image](https://user-images.githubusercontent.com/71656189/134498912-4e75d4a9-f6d9-437b-8016-e1b1858af3d6.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/71656189/134499155-83a9b704-f6c8-4bd4-88f8-75a979479330.png)

**Desktop (please complete the following information):**
 - OS: MacOS

*Added notes**
Also, because of too many numbers, it's causing the text to cut off like the number 2 here:
![image](https://user-images.githubusercontent.com/71656189/134499347-6350cd51-6b0f-4480-acc4-fe41f6912d1d.png)

PR #9346, opened 2021-09-30
[MU4] vst_mac_os

vpereverzev:
Fixed a build for VST3 SDK on MAC OS

Also resolved the following issues:
- #9269
- #9270
- #9242
- #9248

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -251,7 +251,7 @@ DockPage {
             allowedAreas: Qt.TopDockWidgetArea | Qt.BottomDockWidgetArea
 
             height: minimumHeight
-            minimumHeight: 180
+            minimumHeight: 260
             maximumHeight: 520
 
             tabifyPanel: pianoRollPanel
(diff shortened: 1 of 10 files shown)


mattblo:
Now there is a problem with opening VST plugins I think. I can in the NightlyBuilds from tonight, now actually open only one VST plugin: Even if they are different plugins! 
Just one VST Plugin open
When the BBC Symphonic Orchestra is open I cannot open the Aria Player and vice versa.

I think it should be rather per channel only one VST plugin window can be opened. That means if I have three channels with e.g. "BBC Sympponic Orchestra", then I should be able to open one plugin for each channel. But not for one channel by clicking several times 3 plugins. Do I understand this correctly? @jessjwilliamson

jessjwilliamson:
> Now there is a problem with opening VST plugins I think. I can in the NightlyBuilds from tonight, now actually open only one VST plugin: Even if they are different plugins! Just one VST Plugin open When the BBC Symphonic Orchestra is open I cannot open the Aria Player and vice versa.
> 
> I think it should be rather per channel only one VST plugin window can be opened. That means if I have three channels with e.g. "BBC Sympponic Orchestra", then I should be able to open one plugin for each channel. But not for one channel by clicking several times 3 plugins. Do I understand this correctly? @jessjwilliamson @vpereverzev

Yes! exactly, one plugin window PER channel, not several per channel. Maybe my request was not well explained properly but yes that's correct.

mattblo:
In addition, I think it would be useful if you try to open an already opened plugin again, that this is then brought to the foreground. Do I see that correctly?

PR #9354, opened 2021-10-01
Fix `PopupView`s with an anchor item going off screen

cbjeukendrup:
I found that popups that have an anchor item didn't take the bounds of the screen into account. 
Schermafbeelding 2021-09-21 om 20 53 46
 
Now they do.

Also, @jessjwilliamson contacted me because for many fonts other than Fira Sans, the texts that should have been semibold were just regular, even though those fonts do have a semibold version included. This was due to _how_ the font weight was set in UiTheme. Now, it works well with both Fira Sans and other fonts. For fonts that don't have a semibold version, but do have a bold version, Qt falls back to the bold variant, which is desired.


jessjwilliamson:
WOW! You fixed the typeface weights! Looks excellent! <3 Happy for this to be merged into master! and holy crap looks like the VSTs now load into MacOS now :D

src/framework/ui/view/uitheme.cpp
```diff
@@ -354,10 +352,7 @@ void UiTheme::setupUiFonts()
 
         font->setPixelSize(size);
         font->setFamily(QString::fromStdString(family));
-
-        if (weight == QFont::DemiBold) {
```
Eism, line 358: I implemented it in https://github.com/musescore/MuseScore/pull/7097 for fixing bold fonts on Windows and macOS
@DmitryArefiev FYI need to check

DmitryArefiev:
Hi @cbjeukendrup , I've compared PR build with master build and don't see a difference.. Can you explain where the fix is please? Thanks!
popup

cbjeukendrup:
You will only see the difference if you undock the mixer panel and move it so that the popup button is close to the border of the screen. In the old version, the popup would go off the screen, and in the new version, it would be moved like most popups so that it will remain visible.

DmitryArefiev:
Oh I see now.. Thank you!

DmitryArefiev:
Checked #9354 on Win10, Mac11, LinuxMint 20 - FIXED

Please merge into master (@Eism @RomanPudashkin )

DmitryArefiev:
I have a strange behaviour on **Windows**. For me, it still doesn't change font to bold using **Arial** or *Tahoma*. 
But on @jessjwilliamson side (Windows) it looks fine 

(Mac and Linux builds work as it should) 

@cbjeukendrup can you try the change on your side please?

Here is my video:

https://user-images.githubusercontent.com/90187801/136052075-9dfea8b4-bbcc-4b3e-8d5e-8930427143db.mp4

Issue #9027, opened 2021-09-03
[MU4 Issue] Use opticalCenter metadata to centre dynamics on notes


oktophonie:
SMuFL metadata provides an `opticalCenter` for dynamics glyphs that should be used for centering dynamics on noteheads; see:
https://w3c.github.io/smufl/latest/specification/aligning-dynamics.html

![optical-center](https://user-images.githubusercontent.com/20604032/132033619-8c7cc684-6aae-4ed1-8442-4321937c504c.png)

We should read this data and use it (at least optionally) for centering dynamics. (So far not even Dorico uses this info!)

Jojo-Schmitz:
But we're not using a **_pp_** glyph, instead two _**p**_ ones, creates the need of calulating the center of those 2 (ot 3 of 4...) `opticatCenter`s

oktophonie:
Two adjacent _p_s are actually combined into a _pp_ ligature when rendered, so it's still a single glyph.

oktophonie:
This will have to wait until we switch over to use the music font (e.g. Leland) rather than the musical text font (e.g. Leland Text) for dynamic symbols, so probably after MU4.

Nick-Mazuk:
Note to self: when https://github.com/musescore/MuseScore/pull/9359 is merged, I can continue work on this

oktophonie:
Stretch goal:

Where a dynamic is on the first note of a bar and it collides with a barline connecting an adjacent stave, offset it rightwards to clear the barline. (Generally this will only affect longer dynamics like _fff_ and _ppp_ etc).

![stretch](https://user-images.githubusercontent.com/20604032/136162293-e39c0406-2a23-4fc0-b199-7f7ad68ebb35.png)

DmitryArefiev:
Checked #9027 on Win10, Mac11, LinuxMint 20 - DONE (compared with previous builds)

Issue #9218, opened 2021-09-21
[MU4 Issue] Exported PNGs always have transparent background


Nick-Mazuk:
**Describe the bug**

Exported PNGs always have transparent background, even when "Transparent background" is unchecked.

**To Reproduce**
1. Create or open a score
2. Go to File > Export
3. For "Format", select "PNG images"
4. Make sure "Transparent background" is unchecked

**Expected behavior**

If the transparent background checkbox is unchecked, the background should not be transparent. I'd expect the background to be white by default.

**Screenshots**

https://user-images.githubusercontent.com/5659171/134218962-db0eb07d-8b7f-4278-9525-2cdce14b0f78.mov

If you go through this process with this file: 
[accidental-test.mscz.zip](https://github.com/musescore/MuseScore/files/7205082/accidental-test.mscz.zip)

here's the output:

![accidental-test-1](https://user-images.githubusercontent.com/5659171/134218583-74816602-d637-45a0-8066-833456019dac.png)

**Desktop (please complete the following information):**
 - OS: macOS 11.6

PR #9359, opened 2021-10-03
use symbol fonts for dynamics/tempo symbols

Nick-Mazuk:
Related (but does not resolve) https://github.com/musescore/MuseScore/issues/9027

Currently, dynamics and tempo markings use text fonts (e.g., Leland Text) and not symbol fonts (e.g., Leland). This prevented us from using SMuFL attributes to precisely position these elements on the page.

Features:

- Adds two new style parameters to adjust the font-size of the symbols independently from the text for both the dynamics and tempo markings
- Automatically scales dynamics to be roughly the same size between fonts. This is a "patch fix" until the underlying font files are adjusted

Known issues:

- Trying to edit dynamics (i.e., by double-clicking on them) crashes MuseScore. This existed before this PR.
- In the rare case MuseScore you can edit dynamics without it crashing, the vertical cursor is way too tall. This seems unrelated to this PR so I didn't fix it, but I can if desired


Nick-Mazuk:
The builds are successful, but the tests pass simply because the original title of the PR included a `/`.

Jojo-Schmitz:
Unless I screwed it up I do get some weird results:
Before:
image
After:
image
Seems too large and there doesn#t seem to ba any special hanling for MuseJazz
The corresponding check for Bravora shows larger notes too, but I think this is intended, before:
image
after:
image
Gonville and Emmentaler look OK, no big difference.

For Dynamics it is getting really wild though, esp. for Bravura, before:
image
after
image

MuseJazz dynamics get destroued entirely, before:
image
after:
image

For Gonville and Emmentaler they got a bit bigger, again probably on purpose.

oktophonie:
The discrepancy with MuseJazz is to be expected; the font needs to be updated to correct the scale of these symbols (as we just did with Leland).

Bravura (not Bravura Text!) works as expected on my machine.

Emmentaler and Gonville both seem to give me small (too-small) Bravura dynamics, for some reason, though.

Nick-Mazuk:
Just a heads up, Jojo, could you add images with opaque backgrounds next time? Those previous images do not have any backgrounds. Several of us use GitHub in dark mode, meaning it becomes really difficult to see the images (without digging through GitHub's settings and turning light mode on temporarily).

Screen Shot 2021-10-04 at 10 02 47 AM

On macOS, you can easily take screenshots by pressing ⌘⇧4 then click and drag over the area you wish to take the screenshot of. This will automatically include the white background for the music. I don't know if there's an equivalent feature in Windows or Linux.

Jojo-Schmitz:
Well, I used copy/paste from the 3.x vtests. and those are generated with transparent background apparently.

Nick-Mazuk:
> Well, I used copy/paste from the 3.x vtests. and those are generated with transparent background apparently.

Interesting. I do know that I've run into issues with exporting non-transparent backgrounds in MuseScore 4 (https://github.com/musescore/MuseScore/issues/9218). Wonder if they're related?

Jojo-Schmitz:
> The discrepancy with MuseJazz metronome mark symbols is to be expected; the font needs to be updated to correct the scale of these symbols (as we just did with Leland).

Or getting special-cased in `qreal TextBase::dynamicScalingRatio() const` just like it currently happens for Gonville, Emmentaler and Petaluma

> Bravura (not Bravura Text!) dynamics work as expected on my machine. (as does MuseJazz)

Not sure what the 3.x vtests uses, but those dynamics are way too big.

> Emmentaler and Gonville both seem to give me small (too-small) Bravura dynamics, for some reason, though.

Maybe those special-casing for those in `qreal TextBase::dynamicScalingRatio() const` needs some adjusting then?

Jojo-Schmitz:
A rebase should fix the build issue

Jojo-Schmitz:
That merge went horribly wrong I'm affraid.

Nick-Mazuk:
> That merge went horribly wrong I'm affraid.

No! That's not what a rebase should look like…

I'll fix it after lunch

Jojo-Schmitz:
I'm using ` git pull --rebase upstream master`

src/engraving/libmscore/textbase.cpp
```diff
@@ -961,7 +961,7 @@ mu::draw::Font TextFragment::font(const TextBase* t) const
     QString family;
     if (format.fontFamily() == "ScoreText") {
         if (t->parent() && t->isDynamic()) {
-            family = t->score()->styleSt(Sid::MusicalSymbolFont);
+            family = t->score()->scoreFont()->fontByName(t->score()->styleSt(Sid::MusicalSymbolFont))->family();
```
Jojo-Schmitz, line 964: Why this change?

Nick-Mazuk: This fixes the issue where Emmentaler and Gonville simply displayed Bravura instead. Internally, Emmentaler is called "mscore" and Gonville is called "Gootville". This change maps the fonts to the correct values.

You could alternatively hardcode in the mappings, but that would make it harder to maintain later as the `ScoreFont` class should be the single source of truth for the fonts.

Nick-Mazuk:
Ok, rebase issue should be fixed and several font files have been updated (thanks Simon for that!).

I expect vtests to fail since there are breaking visual changes.

Jojo-Schmitz:
The fact that the vests don't fail show how broken they are...

Nick-Mazuk:
> The fact that the vests don't fail show how broken they are...

We know…

src/engraving/libmscore/textbase.cpp
Jojo-Schmitz: OK, so as a hack for 3,x (which is apparently lacking that `ScoreFont::fontByName()` we could use this, right?
```
                  family = t->score()->styleSt(Sid::MusicalSymbolFont);
                  if (family == "Emmentaler")
                        family = "MScore";
                  else if (family == "Gonville")
                        family = "Gootville";
```

Nick-Mazuk: That should work.

Jojo-Schmitz: Well, unfortunaly it dosn't really, the vtest for Dynamics are just way too large and not in the ScoreFont style, most obvious for MuseJazz, but the same for Emmentaler, Gonville, Bravura.
I could do a plain PR of this one here on top of the latest 3.x, would that help?

oktophonie (approved):

Jojo-Schmitz:
OK, one problem seems to be that this PR assumes that `Sid::MusicalSymbolFont` and `Sid::MusicalTextFont` always match. That is not neccessarily the case, you can habe a score with Bravura glyphs, but MuseJazz text. So when before it used the MusicalTextFont, it should now use the __corresponding__ MusicalSymbolFont, not the configured one.

Jojo-Schmitz:
Could it be that in 3.x we can't use a Musical Symbol Font for text? Seems the only sensible explanation to my failed attempts to get this to work.

PR #9361, opened 2021-10-03
Fix code style with Uncrustify 0.73.0

cbjeukendrup:
I just noticed that Homebrew now offers Uncrustify 0.73.0. In this version, the behaviour of Uncrustify changed somewhat, so I updated the code style accordingly. 

For all other PRs, the code style check will fail, until this PR gets merged and those other PRs get rebased on top of this one.

I don't necessarily like all of the changes, but yeah... we don't have much choice.


cbjeukendrup:
Hm, apparently on our CI not yet...

Jojo-Schmitz:
Indeed it is still Uncrustify 0.72.0 in GitHub CI

Issue #9153, opened 2021-09-16
[MU4 Issue] [Accessibility] Voice Over doesn't read almost all elements (Mac)


DmitryArefiev:
**Describe the bug**
Voice Over doesn't read almost all elements (Mac)

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: Mac10.15.6

**Additional context**

DmitryArefiev:
Checked #9421/#9153 on Win10, **Mac11**, LinuxMint 20 - FIXED

PR #9364, opened 2021-10-04
Implemented keyboard navigation for Export dialog

cbjeukendrup:


src/project/qml/MuseScore/Project/ExportDialog.qml
```diff
@@ -35,32 +35,44 @@ StyledDialogView {
...
+    }
 
-        readonly property int sideMargin: 24
+    onOpened: {
+        exportScoresListView.focusOnFirst()
     }
 
-    Component.onCompleted: {
-        exportModel.load();
+    NavigationPanel {
```
Eism, line 53: Please move these panels to the appropriate rows, this will simplify the code

cbjeukendrup: Thanks, fixed!

src/project/qml/MuseScore/Project/internal/Export/AudioSettingsPage.qml
```diff
@@ -23,67 +23,73 @@ import QtQuick 2.15
...
     ExportOptionItem {
-        Layout.fillWidth: false
         text: qsTrc("project", "Sample rate:")
-        firstColumnWidth: root.firstColumnWidth
 
         Dropdown {
-            id: srates
             Layout.preferredWidth: 126
 
+            navigation.name: "SampleRatesDropdown"
```
Eism, line 54: You also need to add the option title(Format) for accessibility, because the reader will only say the value
and it will not be clear to the user

navigation.accessibility.name: option.text + " " + currentText

src/project/qml/MuseScore/Project/internal/Export/ExportOptionsView.qml
```diff
@@ -76,14 +83,18 @@ ColumnLayout {
...
         text: qsTrc("project", "File type:")
-        firstColumnWidth: privateProperties.firstColumnWidth
 
         Dropdown {
             id: subtypeComboBox
             Layout.fillWidth: true
 
+            navigation.name: "ExportSubtypeDropdown"
+            navigation.panel: navPanel
+            navigation.row: 2
```
Eism, line 96: you need to control the navigation(enable state) depending on the visibility of the parent
If the parent is invisible, then the navigation will still work, because it is just an object

navigation.enable: option.visible

src/project/qml/MuseScore/Project/internal/Export/AudioSettingsPage.qml
cbjeukendrup: It's difficult that I can't really test it, because our accessibility system does not work at all on macOS (#9153).
Anyway, I did now what seemed right to me, so I hope that will work.

src/project/qml/MuseScore/Project/internal/Export/ExportOptionsView.qml
cbjeukendrup: I think this is not necessary: this is already being done in DropdownItem.qml (which is used as the base for the Dropdown component).
And in my experience, when the parent QQuickItem is invisible, the child items will be invisible too.

Also, when I test it on my computer, it works fine.

Eism: All right. I described another case

Eism (approved):

PR #9371, opened 2021-10-04
Changed pipe-organ pedals clef from F8vb to F

FiskFan1999:
Resolves: "Pipe Organ" instrument has a F8vb clef in the pedals by default, which is not appropriate for organ music.

This pull (which is nothing more than an xml file adjustment, akin to a spelling fix) changes the default clef of the third stave of the pipe-organ instrument from a transposed bass clef to a non-transposed bass clef. The reason for this change is simply that no organ music in the world uses a bass clef that is transposed by one octave for pedals. Terminology of the pipe organ requires that the compass of the pedalboard reach from `C` to `f'` or `g'` (depending on the particular pedalboard), and not from `C,` to `f`, for example (these are the pitches which are played by an 8-foot stop).

I understand, however, that one reason to define the pipe organ clef in the way it is now is to be of help for those who are relying on the midi playback to hear 16-foot tone in the pedals. Although I would argue that having a transposed bass clef for the pedals is still inappropriate for printed organ music, using 16-foot tone in the pedals and 8-foot tone in the manuals is a common registration. Hopefully, someone with more experience with the Musescore source code than me can consider implementing stave-specific transposition similar to the already-existing transposeChromatic tag.

- [x ] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -10110,8 +10110,12 @@
                   <bracketSpan>2</bracketSpan>
                   <barlineSpan>2</barlineSpan>
                   <clef staff="2">F</clef>
-                  <clef staff="3">F8vb</clef>
+                  <clef staff="3">F</clef>
                   <barlineSpan staff="3">1</barlineSpan>
+                  <!--
+                  <transposeDiatonic>-7</transposeDiatonic>
+                  <transposeChromatic>-12</transposeChromatic>
+                  -->
                   <Channel>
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
As far as I know that 3rd staff of an organ, for pedal, and the pitches for those are an octave below those from the 2nd staff, played on the manual. And we can't transpose down just a single staff, hence the need to use the F8vb clef.
And that commented out setting for octave transposing shouldn't be in there, IMHO, as it would affect the entire instrument, once  made active.

Issue #9360, opened 2021-10-03
[MU4 Issue] CI build fails with PR title contains `/`


Nick-Mazuk:
**Describe the bug**

When a PR contains the `/` character, the GitHub Actions test build will fail. Specifically, the build artifacts cannot be uploaded. Here's a sample build failure.

https://github.com/musescore/MuseScore/pull/9359/checks?check_run_id=3783637639

**Expected behavior**

The title of a PR should have no impact to whether or not the build tests succeed.

Jojo-Schmitz:
See faff816 which changed how those  artifact names get generated, Intersitingly not for the (disabled) Windoes 32-bit and PortableApp builds.

Jojo-Schmitz:
See faff816 which changed how those  artifact names get generated, Interstingly not for the (currently disabled) Windows 32-bit and PortableApp builds, which seems another bug (and would probably need the same change).
That commit is now what broke it though, it just removed an `_` from the name.
Maybe `${{ github.event.pull_request.number }}`, i.e. picking on the PR number rather than the title would work here and be more reliable? @igorkorsukov ?

igorkorsukov:
We have added a PR title to the artifact name to make the work of the testers easier. They download hundreds of artifacts, they need to distinguish them from each other easily.
I believe that the `/` symbol is not often present in the title of the PR, but it is better just not to use it. It's about the title of the PR, if it is created automatically from a commit, then we can just edit it.
I believe that making the testers' job easier is more important than being able to use the `/` symbol in the PR title.
Also, I think we do not need to waste time and complicate the CI scripts to be able to specify the `/` symbol in the PR title because there is no need for it.

Jojo-Schmitz:
The PR **number** should make them distinguishable as easy, and even easier (think sorting them by name) than the **title**.

Jojo-Schmitz:
Also it is more than just the / that is forbidden here:
`Invalid artifact name characters include: ",:,<,>,|,*,?,\,/.`

cbjeukendrup:
Maybe even better if _both_ the number and the title are included. 

Anyway, at least on my computer, there is a relatively straightforward way to replace some special characters in the name:
```zsh
tr "/;:$" "_" <<< "\$a/vulnerable;name:"
``` 
gives
```
_a_vulnerable_name_
```

Jojo-Schmitz:
Strange enough `;` is allowed (see above)

igorkorsukov:
[...]

PR #9373, opened 2021-10-05
Fix uploading artifacts for pull requests with: <\special characters/> in the "title"

cbjeukendrup:
Resolves: #9360

Also bump version of styfle/cancel-workflow-action from 0.5.0 to 0.9.1.

### DIFF
--- a/.github/workflows/ci_linux_mu4.yml
+++ b/.github/workflows/ci_linux_mu4.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@0.5.0
+      uses: styfle/cancel-workflow-action@0.9.1
       with:
         access_token: ${{ github.token }}
     - name: Clone repository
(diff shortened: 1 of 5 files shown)


.github/workflows/ci_linux_mu4.yml
```diff
@@ -79,6 +79,8 @@ jobs:
           DO_PUBLISH='false'
         fi
 
+        ARTIFACT_NAME="$(tr "\":<>|*?\\/" "_" <<<"MU4_${{ github.run_id }}_Lin_${{ github.event.pull_request.number }}_${{ github.event.pull_request.title }}")"
```
igorkorsukov, line 82: The build scripts use a variable called `ARTIFACT_NAME`, this is the name of the executable file.
There may be a conflict with this name.
So, I propose to rename, for example, to `UPLOAD_ARTIFACT_NAME`

cbjeukendrup: Done!

.github/workflows/ci_windows_mu4.yml
```diff
@@ -138,7 +142,7 @@ jobs:
       if: ${{ always() }}
       uses: actions/upload-artifact@v2
       with:
-        name: MU4_${{ github.run_id }}_Win${{ github.event.pull_request.title }}
+        name: ${{ env.UPLOAD_ARTIFACT_NAME }}
```
Jojo-Schmitz, line 145: Why not changing this for the 32bit and the portableApp builds too, even with those being commented out at the moment?
Maybe adding a Win32 / PortableApp prefix?

cbjeukendrup: I thought the commented out ones were not really up to date anyway, so initially didn't touch them.
But I updated them too, now.

Jojo-Schmitz: They will get used for Release and probably Beta, maybe Alpha too

igorkorsukov (approved):

PR #9380, opened 2021-10-05
Added autobot zoom script test

igorkorsukov:
Added the ability to write scripts for test automation.
Now the goal is not to write automated tests (which do something and check the result), but to be able to write scripts to automate routine tester tasks so that manual testing is faster and easier.

this is just the beginnings of the system

https://user-images.githubusercontent.com/3818029/136062033-434021d1-7561-4bd0-8bdf-66eca82fe207.mp4

### DIFF
--- a/src/autobot/autobotmodule.cpp
+++ b/src/autobot/autobotmodule.cpp
@@ -38,7 +38,11 @@
 #include "internal/autobotactions.h"
 #include "internal/autobotscriptsrepository.h"
 
+#include "internal/api/apiregister.h"
+#include "internal/api/logapi.h"
+
 using namespace mu::autobot;
+using namespace mu::api;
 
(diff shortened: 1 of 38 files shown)


vpereverzev (approved):

src/autobot/autobotmodule.cpp
```diff
@@ -69,6 +75,11 @@ void AutobotModule::resolveImports()
     if (ar) {
         ar->reg(std::make_shared<AutobotActions>());
     }
+
+    auto api = modularity::ioc()->resolve<IApiRegister>(moduleName());
+    if (ar) {
```
cbjeukendrup, line 82: Oops, this should have been `if (api) {`, shouldn't it?

igorkorsukov: > Oops, this should have been `if (api) {`, shouldn't it?

Oops...
I will fix it in a next PR

PR #9387, opened 2021-10-06
[MU3] Use symbol fonts for dynamics or tempo symbols / Let dynamic/tempo symbols auto-resize to surrounding text 

Jojo-Schmitz:
Backport of #9359, #9395 and #942, and only/mainly to get the vtests to run (which are expected to fail)


Jojo-Schmitz:
No longer needed, now that a) the vtests on master work and b) those PRs are merged

Issue #9379, opened 2021-10-05
[MU4 Issue] Improve spacing of accidentals in key signatures


oktophonie:
MuseScore seems to space accidentals at the same interval regardless of the font. Already with Leland the sharps are slightly too close for comfort:

![leland-sharps](https://user-images.githubusercontent.com/20604032/136035803-37881058-0df2-49e5-a79a-b451ed7c0faf.png)

And with Petaluma they are colliding unpleasantly:

![petaluma-sharps](https://user-images.githubusercontent.com/20604032/136035867-30867021-1205-469d-bc62-eceb4a445a55.png)

The spacing should take into account the width/shape of the symbols in each respective fonts, as when accidentals appear on chords, e.g.:

![petaluma-fixed](https://user-images.githubusercontent.com/20604032/136036213-d7c2e589-d9ce-48fc-b49f-196f9cd18834.png)

PR #9389, opened 2021-10-06
use accidental width to space key signatures

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/9379

This PR uses the accidental width to calculate the spacing of key signatures. Before, all accidentals were spaced 1 space apart, regardless of font or symbol. Now, the key signature calculates each symbol's width to space them out appropriately, and the gap between them is `accidentalDistance`, the same style value used for accidentals on chords.

This will provide greater consistency and clarity for accidentals.

Notes
- This PR may not work for custom key signatures. As of today, it looks like it's not possible to create custom key signatures in v4, so I was not able to test it.
- Unlike for chords, there does not seem to be much space remove extra space with symbol cutouts. This is because key signatures have well defined patterns (like they move in 4ths or 5ths), and there's just not the vertical clearance in any of the fonts. _The only_ places I could see making the case for adding this is sharps in Peteluma or naturals cancelling out the sharps. If this is desired, I can add that in (though it would increase the scope of the PR quite a bit).

Screen Shot 2021-10-06 at 2 57 34 PM

Screen Shot 2021-10-06 at 2 57 55 PM

Screen Shot 2021-10-06 at 2 58 16 PM

Screen Shot 2021-10-06 at 3 03 43 PM

### DIFF
--- a/src/engraving/libmscore/keysig.cpp
+++ b/src/engraving/libmscore/keysig.cpp
@@ -100,13 +100,18 @@ void KeySig::addLayout(SymId sym, int line)
     if (_sig.keySymbols().size() > 0) {
         KeySym& previous = _sig.keySymbols().back();
         qreal accidentalGap = score()->styleS(Sid::keysigAccidentalDistance).val();
+        if (previous.sym != sym) {
+            accidentalGap *= 2;
+        } else if (previous.sym == SymId::accidentalNatural && sym == SymId::accidentalNatural) {
+            accidentalGap = score()->styleS(Sid::keysigNaturalDistance).val();
+        }
         // width is divided by mag() to get the staff-scaling-independent width of the symbols
         qreal previousWidth = symWidth(previous.sym) / score()->spatium() / mag();
         x = previous.spos.x() + previousWidth + accidentalGap;
         bool isAscending = y < previous.spos.y();
         SmuflAnchorId currentCutout = isAscending ? SmuflAnchorId::cutOutSW : SmuflAnchorId::cutOutNW;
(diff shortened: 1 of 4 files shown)


Jojo-Schmitz:
Compare with this
image
using Leland, it does seem to use cutouts and looks narrower than your 1st example. same with Bravura, Emmentaler, Gonville, MuseJazz and Petaluma too, very clearly (and possible a tad too much) for the latter two in 3.6.2, MuseJazz:
image
Petaluma:
image
Not with flats though.
So IMHO those cutouts should be used.

Nick-Mazuk:
Two updates to the PR:

1. The accidental distance is now defined by its own style parameter named `keysigAccidentalDistance` It currently has the same value as `accidentalDistance`.
2. Accidental cutouts are now honored

Screen Shot 2021-11-29 at 10 57 26 PM

Screen Shot 2021-11-29 at 10 57 09 PM

Nick-Mazuk:
Looks like the vtests are failing, but the exact failure is expected with this change.

Jojo-Schmitz:
Before:
image
After:
image
See a `vtests` label is in order ;-)

oktophonie:
The kerning-via-cutouts is excellent in principle, but the naturals in flat-formation are too close together. Indeed everything is a bit too tight so probably the basic spacing value should be increased slightly.

Nick-Mazuk:
Quick update, this PR is not ready for merge. For some reason, if the gap size style property for key signature is updated, the score does not visually update.

RobFog:
You can mark the pull request as a "draft".

Nick-Mazuk:
Ok, assuming the tests pass, I believe the PR is ready for merging.

oktophonie (approved):

RomanPudashkin (approved):

PR #9390, opened 2021-10-07
[MU3] Use accidental width to space key signatures

Jojo-Schmitz:
Backport of #9389 and only/mainly to get the vtests to run (which are expected to fail)


Jojo-Schmitz:
No longer needed, now that the vtests on master work

PR #9395, opened 2021-10-07
let dynamic/tempo symbols auto-resize to surrounding text

Nick-Mazuk:
Follow up to resolve some of the comments brought up in https://github.com/musescore/MuseScore/pull/9359 after it was merged.

Instead of having separate font sizes for dynamic/tempo symbols, the symbols will automatically resize given the surrounding context text sizes.

One other change: default dynamics font size is now 10pt instead of 11pt.


Jojo-Schmitz:
It also still does nothing to import MuseScore 2.x scores correctly. All dynamics (in Bravura, Emmentaler, Gonville, MuseJazz, other fonts didn't exist in 2.x) are are getting rendered way too big and identically in a fallback font (Leland or Bravura, I guess). Test scores attached (from 3.x vtests):
[vtest.zip](https://github.com/musescore/MuseScore/files/7309081/vtest.zip)

PR #9290, opened 2021-09-27
Show instrument descriptions in New Score Wizard

shoogle:
Instrument descriptions are displayed in the bottom left corner of the New Score Wizard dialog. The description text is taken from instruments.xml (i.e. the online instruments spreadsheet).

### Timpani

image

### G Dizi

image

PR #9405, opened 2021-10-09
Add treble clef version of B sousaphone

simonstuder:
Added the treble clef version of the b sousaphone to help beginners play multiple instruments with the same clef.
In our town so far multiple bands use a custom instrument.xml as second instrument list.

`generateTs.py` was run, but `lupdate.sh` step has unclear instructions and no ts file was generated.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [ ] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [ ] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #9406, opened 2021-10-09
[MU4] Make instruments descriptions translatable

Jojo-Schmitz:
and have the script adding the license header to intrumentsxml.h

Reveals how long that generateTS.py hasn't been run and that it probably should be running as part of the build process.

Needed after #9290 ("Show instrument descriptions in New Score Wizard")


Jojo-Schmitz:
Still that script should also add the GPL header, shouldn't it?
Please also check #9405

PR #9409, opened 2021-10-09
avoid crash when no telemetry

lyrra:
When entering preferences musescore crashes, due to telemetryConfiguration() returning NULL.


RomanPudashkin:
See https://github.com/musescore/MuseScore/pull/10068

Issue #9073, opened 2021-09-08
[MU4 Issue] About MuseScore opens About XML window on Mac


DmitryArefiev:
**Describe the bug**
About MuseScore opens About XML window on Mac

**To Reproduce**
Go to MuseScore>About MuseScore

**Expected behavior**
It should open About MuseScore window instead of About MusicXML

**Screenshots**
![About MuseScore_Mac](https://user-images.githubusercontent.com/90187801/132516787-ff77393d-e17b-41e5-a7d7-fb871b954a86.png)

**Desktop (please complete the following information):**
 - OS: Mac10.15.6 (Mac only)

cbjeukendrup:
Fixed in #7855.

Jojo-Schmitz:
Also fixed in #9410 now ;-)

PR #9410, opened 2021-10-09

cbjeukendrup:
Resolves: #9073

Port of PR #7855, which has gone stale and has needed a rebase for several months now.


cbjeukendrup:
Thanks! I guess "Co-authored-by:" is the best option in this case, considering that we came to this solution together.

Issue #8236, opened 2021-06-02
[MU4 Task] Add opening scores from File Explorer / Finder / other places outside MuseScore


cbjeukendrup:
**Task description**
Implement opening scores from File Explorer (Windows) or Finder (macOS) or equivalent sources outside MuseScore.

Currently when you tell for example Finder to open a mscx/mscz file with MuseScore 4, the app becomes active but does not yet actually open the score.
(See also MU3)

DmitryArefiev:
Tested  on latest master build on Mac11.6 - DONE

PR #9411, opened 2021-10-10
[macOS] Allow opening files from Finder, dock icon menu and more

cbjeukendrup:
Resolves: #8236

Before, in MU4 it was only possible to open files from within the app (for example, from the home tab, or the open recent menu).

Now, the app also reacts to File Open events from macOS, so now you can open files from Finder, from the context menu of MuseScore's Dock icon (which makes the work done in PR #8299 a lot more useful), and more (just like in MU3).

(If you have MU3 installed too, macOS will likely prefer using MU3, so you would have to choose "Open with > MU4".)

On Windows, it already works (sort of) due to the multi-instances system.

I also performed a very small cleanup in some CMakeLists.txt files.

### DIFF
--- a/src/project/internal/projectfilescontroller.cpp
+++ b/src/project/internal/projectfilescontroller.cpp
@@ -21,8 +21,8 @@
  */
 #include "projectfilescontroller.h"
 
-#include <QObject>
 #include <QBuffer>
+#include <QFileOpenEvent>
 
 #include "translation.h"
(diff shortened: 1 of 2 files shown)


DmitryArefiev:
Tested #8236 on latest master build on Mac11.6 - DONE

Now works on Mac. Thanks for fixing that!

Issue #9397, opened 2021-10-08
[MU4 Issue] Weird score scrolling (need to specify exactly how scrolling should work)


vshalkevich:
**Describe the bug**
Videos with weird scrolling attached: 
1) Using mouse scroll, it is possible to scroll under the manually scrolling bar
2) Scrollbar UI disappeared, scrolling is possible by using mouse scrollbar

**Expected behavior**
need to specify exactly how scrolling should work

**Screenshots**

https://user-images.githubusercontent.com/71372310/136481040-bd5903cd-19eb-4068-84ab-5bfc9ae7915a.mp4
https://user-images.githubusercontent.com/71372310/136481053-f5929641-813f-4f7e-9ea4-8a1042be5587.mp4

**Desktop (please complete the following information):**
Windows 10 
Linux
MacOS

bkunda:
Hi Slavovich! Good spot here.

I've tried to explain my preference for expected scrollbar behaviour in this video: https://www.dropbox.com/s/zzdqkd1i7mvlopk/Weird-scrollbar-behaviour.m4v?dl=0

Basically, click on scrollbar + drag should work in the same way as using the scroll wheel.

**Current behaviour**: click on scrollbar + drag locks scroll to page height/width.
**Expected behaviour**: both scrolling methods should scroll past the edge of the page to the x/y-axis limit currently in place (a limit that does not allow the user to scroll the score out of view). 

I was unable to reproduce the disappearing scrollbar issue on Windows or Mac, or perhaps I didn't understand it completely. 

Another issue I spotted in the meanwhile: click on scrollbar + scroll with scroll wheel produces an undesirable result. In my video, I advocate for disabling scroll with this behaviour (consistent with how Windows and MacOS works, as well as some apps like Adobe apps). I'm open to discussion on this though (Figma, for example, allows users to click on a scrollbar and use a scroll wheel); were this to be permitted, scrolling should work exactly as it does by using scroll wheel alone.

Hope this helps!

PR #9412, opened 2021-10-10
Make notation view scrollbars behave the same as scrolling with the mouse

cbjeukendrup:
Resolves: #9397 

Now, if we want to change how far the user can scroll (with the mouse or scrollbars or whatever), we can simply adjust the `NotationPaintView::scrollableAreaRect` method; the scrollbars and everything will behave accordingly.

It also works well together with the preferences option to limit the scrolling area to the page borders (see Edit > Preferences > Canvas > Limit scroll area to page borders).

I also added some extra checks to make sure the notation stays in the allowed area when the window is resized and when the user just toggles that preference option.

(Note: the issue with scrolling with the mouse wheel while keeping the scrollbar pressed, is _not_ supposed to be fixed by this PR.)

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledScrollBar.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledScrollBar.qml
@@ -22,8 +22,6 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 
-import MuseScore.Ui 1.0
-
 ScrollBar {
     id: root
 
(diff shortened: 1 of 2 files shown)


cbjeukendrup:
I restored the Figma-like scrolling, and made custom scrollbars, to allow creating over-scroll using the scrollbars. It works quite well! I also made some improvements to the existing StyledScrollBar, which is not used for the notation view anymore, but in all other places in the app. The work is slightly inspired by Qt's own implementation (see https://code.qt.io/cgit/qt/qtquickcontrols2.git/tree/src/quicktemplates2/qquickscrollbar.cpp?h=5.15.2 and https://code.qt.io/cgit/qt/qtquickcontrols2.git/tree/src/imports/controls/ScrollBar.qml?h=5.15.2).

cbjeukendrup:
It looks expected to me that the scrollbar is not visible at low zoom levels, because then the score is always fully visible in the view. I think it would not make much sense to show a scrollbar that has a height of 100% and therefore cannot be moved. 

At zoom levels where the height of the score is less than 25% of the view, the scroll limit of 75% prevents the score from going outside the view anyway, so there will never be any overscroll, so no scrollbar.

(I'll add the bottom/right padding to the scrollbars. I hadn't seen that Figma does this, but it looks quite good there, so let's do that too.)

DmitryArefiev:
Tested #9397 on Win10, Mac11, LinuxMint 20 - DONE

Please merge into master

Issue #9257, opened 2021-09-24
[MU4 Issue] Grip handles are too small in MacOS


bkunda:
Grip handles used to resize lines are appearing too small in MacOS.

Demo video: compares MU4 handle size with MU3 handle size. Desired behaviour: handles should appear as they do in MuseScore3. 

![grip-handles-too-small](https://user-images.githubusercontent.com/86290556/134617175-eaf5731c-5701-40d3-bed2-a523f67ec12b.gif)

MacBook Air M1, MacOS 11.6

cbjeukendrup:
I think this is not related to macOS, but to high-dpi screens (because if you run the app in low dpi mode on macOS, the problem is solved).

Also, I doubt whether we should exactly follow the behaviour of MU3. In MU3, grip handles have a fixed size, regardless of the zoom level. That way, they look okay-ish with common zoom levels, but they are relatively too big at smaller zoom levels and invisibly small at higher zoom levels (because they don't stand out from the black lines).

In Dorico, the grips follow the zoom level, which looks and works a lot better in my opinion. We could also do that, and maybe set a minimum and maximum size, so that they don't get too small at smaller zoom levels or too big at higher zoom levels.

DmitryArefiev:
Checked #9257 on Mac11 - FIXED

PR #9414, opened 2021-10-10
Fix grip size (replicate MU3 behaviour) and make it easier to select grips

cbjeukendrup:
Resolves: #9257 

Despite my comment there, I replicated the MU3 behaviour (which was in fact already done, but guiScaling was not taken into account).

Also, below a grip, there might be elements that do not belong to that grip. When clicking the grip at the location of such an element, that element would get selected rather than the grip. That was frustrating and made it difficult to grab the grip.
I changed that: now, when you click anywhere inside a grip's rectangle, you will be selecting the grip, instead of the unrelated element below it.
(When you still want to select the element below it, just click outside the grip, or click somewhere on the canvas to unselect the element to which the grip belongs, so that the grip disappears.)

### DIFF
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -316,6 +316,11 @@ void NotationViewInputController::mousePressEvent(QMouseEvent* event)
         hitStaffIndex = context.staff ? context.staff->idx() : -1;
     }
 
+    if (viewInteraction()->isHitGrip(logicPos)) {
+        viewInteraction()->startEditGrip(logicPos);
+        return;
+    }
+
     if (hitElement) {
         RetVal<midi::tick_t> tick = m_view->notationPlayback()->playPositionTickByElement(hitElement);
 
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
I take that something liks eht 1sdt commit might be needed for 3.x as well?
I've found 2 places in the 3.x code that might need this:
https://github.com/musescore/MuseScore/blob/2513676e512d29d554cb6c4d37d3efaf53ea2c5b/mscore/editelement.cpp#L39-L54
and
https://github.com/musescore/MuseScore/blob/2513676e512d29d554cb6c4d37d3efaf53ea2c5b/mscore/fotomode.cpp#L150-L155

cbjeukendrup:
@Jojo-Schmitz I don't think it's necessary in MU3, in the first place because it just works as is (don't fix it if it ain't broken 😉).

In the second place, I think that the guiScaling is somehow handled by the scaling of the painter in MU3. In MU4, there is a lot of multiplying and dividing with that guiScaling, and I suspect that this shouldn't be necessary. I will research that in the future.

Anyway, I see that you added it already to your 3.7 PR, so I will try how it works out in the artifact.

PR #9421, opened 2021-10-11
[MU4] Fixed accessibility on macOS

Eism:
Resolves: #9153

For macOS, installing a root object(QAccessible::setRootObject) doesn't make any sense(not implemented). Instead, the root object of the main window (QQuickWindow) is taken. Let's return our root object for QQuickWindow so that accessibility works with our custom tree, not the qt tree

Also, installFactory adds a factory to the queue. And whichever factory is installed last, it will be called first. To replace the qt tree with work, we need to set our factory after the gui is initialized(on the first request to register the gui element)

### DIFF
--- a/src/framework/accessibility/accessibilitymodule.cpp
+++ b/src/framework/accessibility/accessibilitymodule.cpp
@@ -32,8 +32,6 @@
 using namespace mu::accessibility;
 using namespace mu::modularity;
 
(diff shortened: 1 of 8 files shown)


DmitryArefiev:
Checked #9421/#9153 on Win10, **Mac11**, LinuxMint 20 - FIXED

@Eism  Please merge into master

PR #9427, opened 2021-10-11
[MU4] Fix dynamic and ottava symbols when Musical Text font does not match the Musical Symbols font

Jojo-Schmitz:
after #9359 and #9395 and as discussed there.
Also after #12354

Sample scores attached (from the 3.x vtests):
[vtest.zip](https://github.com/musescore/MuseScore/files/7327172/vtest.zip)

### DIFF
--- a/src/engraving/libmscore/textbase.cpp
+++ b/src/engraving/libmscore/textbase.cpp
@@ -796,7 +796,8 @@ mu::draw::Font TextFragment::font(const TextBase* t) const
     if (format.fontFamily() == "ScoreText") {
         if (t->isDynamic() || t->textStyleType() == TextStyleType::OTTAVA) {
             family
-                = String::fromStdString(engravingFonts()->fontByName(t->score()->styleSt(Sid::MusicalSymbolFont).toStdString())->family());
+                = String::fromStdString(engravingFonts()->fontByName(t->score()->styleSt(Sid::MusicalTextFont).replace(u" Text",
+                                                                                                                       u"").toStdString())->family());
             fontType = draw::Font::Type::MusicSymbol;
             // to keep desired size ratio (based on 20pt symbol size to 10pt text size)
             m *= 2;


vpereverzev:
@DmitryArefiev please check this one

RobFog:
Has this been investigated?

Jojo-Schmitz:
I believe the vtests differences are expected, 4 oof the 5 are rather small differences, and the big difference is due to the ref being wrong, before:
image
after:
image
Here clealy the after image is expected, it is MuseJazz after all.

Jojo-Schmitz:
The other smaller diffs most probably also are due to the dynamics taking the Musical fonts dynamics rather the Musical Text font ones, which is excatly what this PR is triying to fix, (only that the differences between Emmentaler Tetx, Bravura Text and Leland Text are much smaller than those to MuseJazz Text.

So I believe a vtests label is in order here., These diffs rather prove that this PR is needed

Jojo-Schmitz:
It is not just pre-3.x dysnamics actually, but pre 4.x dynamics. A score with dynamics from 3.6.2 and earlier where the Musical Symbol font (e.g. Leland) does not match the Musical Text font (e.g. MuseJazz Text), shows the dynamics in the Musical Text font (e.g. Leland Text) that matches the Musical Symbols font (e.g. Leland).

Jojo-Schmitz:
This PR fixes that by still using the dynamics from the Musical Symbols font, as per #9359, but deriving the name from the configured Musical Text font rather than taking that of the configured Musical Symbols font

Jojo-Schmitz:
It is not even pre 4.x scores, but even 4.x score directly

oktophonie:
This behaviour seems potentially misleading to me, obscuring the fact the the dynamic symbols always come from a Musical Symbols font, not a Musical Text font.

If we want to allow the font used for dynamics to not match that used for the rest of the musical symbols, it would be better to have it independently configurable ("Dynamics font" or similar, with the available options being the available Musical Symbols fonts). That would also avoid changes of font during migration.

Jojo-Schmitz:
Well, up to that PR I mentioned above dynamics had always been taken from the Musical Text font (just like ke for tempo text for example), and the change was IMHO only for the better centering of them, using their optical center.
But if MuseScore 4 is really supposed to use the Musical Symbols font now, we'd at least need a compatibility fix for MuseScore 3 and earlier

oktophonie:
The initial impetus for the change was to be able to use the `opticalCenter`, but more importantly is also to confirm with the concepts in SMuFL of what the Symbol vs Text fonts are for. Dynamics are to be thought of as music symbols rather than text symbols (somewhat counter-intuitively, admittedly), and as such their proportions in the Musical Symbols font are determined so that they appear the right size alongside all the other musical symbols.

So, yes, MU4 is definitely to use a Musical Symbols font for dynamics in all cases. That doesn't mean that the font absolutely _must_ match the font used for the rest of the music symbols (though I'd argue that in the vast majority of cases it _should_).

Jojo-Schmitz:
Understood and nothing this PR is changing, it still uses the Music Symbols font for dynamics, but derives the name of if from the Musical Text font in effect for that score.
It might by for the benefit of a minority of casaes, those whe these fonts are different, but still, those do exist.

oktophonie:
Except that if the user sets Musical symbols font to Bravura but Musical text font to Leland Text (for example), the dynamics _should_ come from Bravura, not from Leland.

Jojo-Schmitz:
Which it didn't do before that change in the above mentioned PR and therefore should not after.

At the very  least not for pre-4.x scores

oktophonie:
It should do it after. That's the point. If you want to accommodate the few scores where the dynamics come from a different font, it needs to be independently configurable, as I said.

Jojo-Schmitz:
No need for it to be configurable, just render pre 4.x scores the same as before, in this respect.

If not even that, the corresponding 3 vtests should get removed, as they now don't show what they are (or rather: were) intended to show. The test are emmentaler-text-3, gonville-text-3 and musejazz-text-3.

RobFog:
This seems to be stuck in milestone 45 which was closed a long time ago?

src/engraving/libmscore/textbase.cpp
```diff
@@ -792,10 +792,14 @@ mu::draw::Font TextFragment::font(const TextBase* t) const
 
     String family;
     if (format.fontFamily() == "ScoreText") {
-        if (t->isDynamic() || t->textStyleType() == TextStyleType::OTTAVA) {
+        if (t->textStyleType() == TextStyleType::OTTAVA) {
             family = SymbolFonts::fontByName(t->score()->styleSt(Sid::MusicalSymbolFont))->family();
             // to keep desired size ratio (based on 20pt symbol size to 10pt text size)
             m *= 2;
+        } else if (t->isDynamic()) {
+            family = t->score()->scoreFont()->fontByName(t->score()->styleSt(Sid::MusicalTextFont).replace(" Text", ""))->family();
```
Jojo-Schmitz, line 800: Any idea anyone what to change this into to make it compile again?

Jojo-Schmitz:
Compile error, same code used to build just fine before the (needed) rebase, not sure what changed under the covers and how to fix it?

Jojo-Schmitz:
musejazz-text-3-1 diff

Jojo-Schmitz:
note to self: rebase needed

oktophonie:
Closing as we're working on a holistic solution for this.

Jojo-Schmitz:
Oh well, after 1 1/2 years of basically having been ignored...

Jojo-Schmitz:
I do admit though, that my solution is a bit hacky.

PR #9452, opened 2021-10-12
use opticalCenter to align dynamics

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/9027

Previously, dynamics were horizontally aligned in respect to their bounding box. Now, they are aligned to the SMuFL `opticalCenter` property. If a dynamic does not have that property, it falls back to using its bounding box.

Note that the `sff` is not aligned correctly because the font needs to be updated.

@oktophonie, some of the dynamics in other fonts do not feel properly aligned when using `opticalCenter` (mainly Petaluma and Bravura). I double-checked the code is correctly using the values present in their `metadata.json`. Could you check the original font files to ensure the correct value for `opticalCenter` is present to begin with?

dynamics-fonts

Test file: [optical-center.mscz.zip](https://github.com/musescore/MuseScore/files/7333386/optical-center.mscz.zip)

### DIFF
--- a/src/engraving/libmscore/dynamic.cpp
+++ b/src/engraving/libmscore/dynamic.cpp
@@ -27,6 +27,7 @@
 #include "score.h"
 #include "measure.h"
 #include "system.h"
+#include "symid.h"
 #include "segment.h"
 #include "utils.h"
 #include "mscore.h"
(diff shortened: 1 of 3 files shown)


oktophonie:
This is great! Nobody panic about the unkerned _sff_! (already fixed in new Leland) :)

Now that I can finally see these in context I see some of the opticalCenter positions need to be tweaked a little too, which I will do forthwith.

oktophonie:
Bravura and Petaluma look fine to me - that is, the values in their metadata are present and are being used as expected. Some of the positionings are a bit different from Leland, but the whole point of this mechanism is that it allows precisely that, i.e. that the exact placement of dynamics relative to noteheads can be specific to the music font being used.

PR #9454, opened 2021-10-12
reformat code to match code style

Nick-Mazuk:
29 files were not properly formatted according to the code style bash script. This PR simply fixes the style problems found by that bash script.

Without this change, it can be a pain to check code style locally since running the bash script results in dozens of unwanted changes. With this PR merged, developers can use the bash script again to check code style before submitting a PR.

All changes were made by simply running.

```bash
bash ./build/ci/macos/checkcodestyle.sh
```

_P.S. We need to figure out how these style errors were allowed on the main branch to begin with. Simple things like this should not be allowed past our CI tests and code review._


cbjeukendrup:
See #9361.

The problem is that GitHub for some reason still uses Uncrustify 0.72.0, while the newest version available is 0.73.0, which is what you and I use locally. That causes the difference.

PR #9461, opened 2021-10-13
fix #9417 Nested tuplet bracket collisions

### LINKED ISSUE #9417: [MU4 Issue] Nested tuplets should clear other markings
Tuplets clear articulation marks that appear at the same side, normally.
In the case of _nested_ tuplets, the outside tuplet will clear the articulations, but the inner tuplet(s) does not:

tuplets

Furthermore, if there are no articulations, the outer tuplet does not clear the inner:

tuplets2

**Should fix:**
https://musescore.org/en/node/22905

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/9417

Nested tuplets weren't being added to the skyline, presumably because the infrastructure wasn't in place to add them in the right order (nested -> parent). I added that infrastructure, which ended up being a recursive layout function that lays out all nested tuplets before the containing one.
image
Now that they're properly added to the skyline, they respect collisions with accidentals and other lines.

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -784,21 +784,30 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
     // layout tuplets
     //-------------------------------------------------------------
 
+    std::map<int, Fraction> skipTo;
     for (Segment* s : sl) {
         for (EngravingItem* e : s->elist()) {
             if (!e || !e->isChordRest() || !score->staff(e->staffIdx())->show()) {
                 continue;
             }
-            ChordRest* cr = toChordRest(e);
-            if (!LayoutTuplets::isTopTuplet(cr)) {
+            int track = e->track();
+            if (skipTo.count(track) && e->tick() < skipTo[track]) {
+                continue; // don't lay out tuplets for this voice that have already been done
+            }
+            // find the top tuplet for this segment
+            DurationElement* de = toChordRest(e);
+            if (!de->tuplet()) {
                 continue;
             }
-            DurationElement* de = cr;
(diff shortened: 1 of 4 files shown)


asattely:
Ack, missed some utests. Will have that fixed by tomorrow.

Jojo-Schmitz:
As far as I can this does not (yet) solbe the rebndering of the score from https://musescore.org/en/node/22905:
image
even if there is an improvment vs 
image

Jojo-Schmitz:
> The only style setting I can see being referenced is `Sid::autoplaceEnabled` here:

But that is enabled, so can't be the culprit, or can it?

asattely:
Previously, the code didn't account for split staves--it dealt with detection of nested tuplets in the layout stage by classifying the tuplets by voice. On split staves, the voice can be the same but the tuplets could still not be nested. This has been fixed by considering the track rather than the voice.
image

Jojo-Schmitz:
What about the 2 failing tests? My suspicion is that those are not being tested for in master, maybe?

asattely:
I looked into the tests you mentioned and it seems that the problem is that the files are not able to be loaded in whatsoever, which will be a different issue down the line--there are currently a number of issues with loading old scores. I did test nested ratio-style tuplets with and without brackets and they all seem to be working fine, so I suspect it might be something to do with the loading of old scores with weird tuplets in them. When we start working on old score compatibility we might have to revisit this one but in the meantime I can't reproduce whatever problem is happening in a backport without the ability to load in the file in the first place.

Jojo-Schmitz:
The 2 scores from compat114/tuplets_[12], are indeed creported as being corrupt, maybe that is what master doesn't like?
The xml from musicxml/io/testTuplets5 is not though.

Issue #9431, opened 2021-10-12
[Inspector] Text should be able to go to two lines for all check boxes


Tantacrul:
**Describe the bug**
Currently, when text is too long, it simply falls off the page in the inspector. It was always intended that checkboxes should allow text to fall onto two lines to avoid this problem. This will also to help with localisation in future.

Here is one example, when you click on a time signature.
![image](https://user-images.githubusercontent.com/21022311/136945527-e04f0a62-3685-4a80-80f7-61234b1eff0d.png)

In addition, when we translate other check boxes (and the similar 'eye' toggles), it will be very useful to allow text to move to two lines to avoid hitting character limits - and to make it possible to horizontally resize panels.

![image](https://user-images.githubusercontent.com/21022311/136945923-98f8a933-5608-4b77-8695-22c6f2f08b36.png)

Text should be aligned to the centre of the checkbox so that when it moves to two lines, it is aligned correctly (See the UI_COMPONENTS_&_STYLES page in Figma). This should be possible for all checkboxes (and 'eye' toggles).

![image](https://user-images.githubusercontent.com/21022311/136947869-85aacf2e-9017-42e4-b2f8-f1d40319a2c0.png)

DmitryArefiev:
Tested on Win10, Mac11, LinuxMint 20 - FIXED

PR #9475, opened 2021-10-14
Fix width of CheckBoxes in Inspector so that they use two lines instead of being clipped

cbjeukendrup:
Resolves: #9431

In most cases, it was a matter of telling the checkboxes about the maximum width they were allowed to use, namely that of the parent.
However, there was also a mistake in the code that calculates the width of the text label.

Also, brought the implementations of CheckBox and VisibilityBox closer to each other.

And, changed spacing between checkbox and text from 8 to 6, because on Figma, it's 6 and the spacing for radio buttons is 6 too.


DmitryArefiev:
Tested #9431 on Win10, Mac11, LinuxMint 20 - FIXED

Issue #9396, opened 2021-10-08
[MU4 Issue] Playback button looks active after ending score playback


vshalkevich:
**Describe the bug**
Playback button looks active after ending score playback. It doesn't look like it has actually stopped.

**To Reproduce**
1. Create a score
2. Click Play button from Playpanel
3. Wait until score playback ends

**Expected behavior**
Playhead (blue bar) should disappear and the pause button should turn back to a play button. 

**Screenshots**
https://user-images.githubusercontent.com/71372310/136480717-a5d98d84-a1aa-47bc-9333-df02792c48f6.mp4

**Desktop (please complete the following information):**
MacOS
Windows 10
Linux

PR #9487, opened 2021-10-17
Fix #9362 Stop playback at the end of the score

### LINKED ISSUE #9362: [MU4 Issue] Playback mode is not stopped at the end of the score
**Describe the bug**
The playback mode is not automatically stopped at the end of the score, so the playback starts again automatically as soon as you click into any position of the score. In MS3 the playback was stopped at the end. If you clicked on another note afterwards, you were in NoteInput mode. It should be the same in MS4

**To Reproduce**
1. Open a Score and play to End
2. Click on an other Note
3. Playback starts from this position without the blue Cursor (Playback Cursor)

**Expected behavior**
End playback mode at the end of the score

**Screenshots**

https://user-images.githubusercontent.com/63284090/135806749-f9fdfbbb-0fc6-4f46-af64-43072305f76a.mp4

 - OS: Windows 11

martin-danhier:
Resolves: #9362, #9396 

Hello,

Currently, when the player arrives at the end of the score (when the time reaches the value of the ``timeDuration`` variable that defines when the end is), the player pauses. This prevents the playback from starting again from the beginning. However, it does not completely stop the playback, which means that if we move the player at another point of the score, it will start again.

To fix this issue, I added a notification that the player can send when the time reaches the ``timeDuration``. The playback controller can then listen to that notification and properly stop the playback when that happens.

### DIFF
--- a/src/framework/audio/internal/worker/playerhandler.cpp
+++ b/src/framework/audio/internal/worker/playerhandler.cpp
@@ -164,7 +164,7 @@ Channel<TrackSequenceId, PlaybackStatus> PlayerHandler::playbackStatusChanged() 
 Notification PlayerHandler::playbackDurationTimedOut() const
 {
     ONLY_AUDIO_MAIN_OR_WORKER_THREAD;
-    
+
     return m_playbackDurationTimedOut;
 }
 


vpereverzev:
A couple of notes here:
- in order to fix this issue we just have to handle notification about playback position change in PlaybackController (who knows about actual duration of a score) and if it's >= duration of the score, PlaybackController should explicitly call stop()
- I'd suggest to not work on any issue related to playback now, since we're working on the upgrade of our playback quality at the moment, this means that it'll be refactored

PR #9493, opened 2021-10-18
Don't let horizontal and vertical NotationScrollBars collide (and other improvements)

cbjeukendrup:
- When scrolling slowly at high zoom levels, the view would not move, because of rounding errors. Now, scrolling works super super smooth at any zoom level at any speed. To make this work, I fixed the rounding errors, and while at it, replaced some QPoint and QRect usages with our own PointF and RectF. I also replaced QTransform with our own Transform. 

- The Notation Navigator works once again (its scaling got somehow set to zero and was never recovered, because scaling happens by factor, and any factor times zero is zero. The "division by zero" crash you might expect did not occur though, because there was a check against it, but that check actually hid the problem).

- I fixed the size / line width of some zoom-independent elements, like the range selection rectangle and the lasso. Additionally, I eliminated a lot of multiplications and divisions by guiScaling, and thus simplified the code somewhat.

- Don't let horizontal and vertical notation scrollbars collide in the bottom left corner

- Rename PainterPath::EngravingItem back to PainterPath::Element (the Big Renaming turned out a little bit too big 😄)

### DIFF
--- a/src/engraving/infrastructure/draw/painterpath.cpp
+++ b/src/engraving/infrastructure/draw/painterpath.cpp
@@ -112,7 +112,7 @@ void PainterPath::translate(double dx, double dy)
     }
 
     setDirty();
-    PainterPath::EngravingItem* element = m_elements.data();
+    PainterPath::Element* element = m_elements.data();
     assert(element);
     while (m_elementsLeft--) {
         element->x += dx;
@@ -144,7 +144,7 @@ size_t PainterPath::elementCount() const
     return m_elements.size();
 }
 
-PainterPath::EngravingItem PainterPath::elementAt(size_t i) const
+PainterPath::Element PainterPath::elementAt(size_t i) const
 {
     assert(i < elementCount());
     return m_elements.at(i);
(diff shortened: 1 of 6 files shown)


DmitryArefiev:
Hello @cbjeukendrup 

1. How can I check the 3rd part of your fix?
> I fixed the size / line width of some zoom-independent elements, like the range selection rectangle and the lasso. Additionally, I eliminated a lot of multiplications and divisions by guiScaling, and thus simplified the code somewhat.

2. When checked the 4th part of you fix I found the **right** scrolling control (appears when zooming) is **too thin**
 Not sure it's related or not to your change..

https://user-images.githubusercontent.com/90187801/140132572-3609c16a-9300-4fbd-a4b8-3c472b2371cd.mp4

cbjeukendrup:
Hi @DmitryArefiev!

About point 1: you could 
- make a lasso selection (Ctrl/Cmd + drag) and verify that the line width of the lasso is the same regardless of the zoom level or screen resolution;
- select a whole measure and verify that the line width of the blue rectangle around the selected measure is the same regardless of the zoom level or screen resolution;
- select a barline or a tie or something that has grips, and verify that the size of the grips is the same regardless of the zoom level or screen resolution.

About point 2: this seems to be the fault of the docking system; the rightmost 10 pixels or so of the window are cut off. This is a relatively recent regression. You can dock the Note Input Bar to the right side of the window to see the effect more obviously, and to see that the width of the notation scrollbar is actually correct.

DmitryArefiev:
Thanks! @cbjeukendrup 

For point 2 I will log a new bug

For point 1 I see the line width is changing when zooming:

selection

cbjeukendrup:
That's a kind of optical illusion. On macOS, it's easy to take screenshots, open them in the Preview app, and zoom in on the screenshot so that you can select the individual pixels. Both when the score is zoomed in and when it's zoomed out, the line width is 3px.
Schermafbeelding 2021-11-03 om 23 15 36

Issue #9025, opened 2021-09-03
[MU4 Issue] Fractional beam on small staff too long


oktophonie:
On a small stave (or resized stave) fractional beams are scaled wrongly - actually getting longer, rather than shorter.

![fractional](https://user-images.githubusercontent.com/20604032/132011882-fe6ac30e-830b-41e7-9ce1-5c9f228b87c8.png)

**See / Should fix:**
https://musescore.org/en/node/319243

PR #9496, opened 2021-10-19
Take staff scaling into account when calculating length of fractional beams

cbjeukendrup:
Resolves: #9025 
Resolves: https://musescore.org/en/node/319243

Fractional beam length fixed


Nick-Mazuk:
I'm actually working on redoing much of the beaming code right now. The first commit which fixes the issue is great! It should be merged in my opinion.

The second one with adding the `const` values should not be merged. As part of the new beam specs, I'm completely redoing the slanting algorithm, so merging that commit will only lead to merge conflicts when my PR is done.

PR #9504, opened 2021-10-19
[From MU3 backend] ENG-30: Change non-end melismas to hyphens

asattely:
From https://github.com/musescore/MuseScore/pull/8133


asattely:
I'm going to close all of these forward-ports because a lot of them deal with the same files, so it makes sense to combine them in a more intelligent way.

PR #9506, opened 2021-10-19
[From MU3 backend] Eng-35: MusicXML import Small Notes vs. Small Chord

asattely:
from https://github.com/musescore/MuseScore/pull/8181


asattely:
I'm going to close all of these forward-ports because a lot of them deal with the same files, so it makes sense to combine them in a more intelligent way.

Issue #9458, opened 2021-10-13
[Crash] Opening a Guitar Pro file (.gpx) causes MU4 to crash


oktophonie:
**Describe the bug**
Opening any .gpx file causes MU4 to crash (after about a second).

**To Reproduce**
- File -> Open, select a .gpx file to open

**Desktop (please complete the following information):**
- Windows 10/11

Jojo-Schmitz:
I guess #9628 is for fixing that crash?

PR #9512, opened 2021-10-20
Returned old gp import implementation 

igorkorsukov:
To find a solution to the problem https://github.com/musescore/MuseScore/issues/9458


src/importexport/guitarpro/internal/importgtp.cpp
```diff
@@ -2994,6 +2994,9 @@ Score::FileError importGTP(MasterScore* score, const QString& name)
     //
     std::vector<Part*> infoParts;
     for (Part* part : score->parts()) {
+        //! HACK Temporary disabled, something not corrected, so crashed
+        continue;
+
```
Jojo-Schmitz, line 2999: Is this hack still needed? That "temporary" meanwhile lasts more than half a year...

PR #9574, opened 2021-10-26
[MU4] Fix #325825: Crash on playback with an open volta before a section break on a frame

Jojo-Schmitz:
Resolves: *https://musescore.org/en/node/325825*

Same issue with 3.x, so will add it to #9000 too


Jojo-Schmitz:
Please merge this (or rather its 3.x variant) also into the 3.6.2_backend branch, there just came up another (so now 2nd) issue about this, see https://musescore.org/en/node/331098, and it does make the backend crash (on MP3 export and/or playback)

Issue #9464, opened 2021-10-14
[MU4 Issue] 'Reset all styles to default' button has no effect


oktophonie:
**Describe the bug**
'Reset all styles to default' does nothing when clicked.

**To Reproduce**
- Open the Style window (Format -> Style from the menu, or right-click the score and choose 'Style')
- Click the 'Reset All Styles To Default' button at the bottom left.

**Expected behaviour**
All the default style settings from MU4 should be applied, overriding any customisations in the current file.

DmitryArefiev:
Tested on master build on Win10, Mac11, LinuxMint 20 - PASS

Issue #9187, opened 2021-09-17
Vertical justification should be on by default for all new scores (as with MS3.6)


Tantacrul:
https://user-images.githubusercontent.com/21022311/133820976-d5f575b9-f908-464b-9e31-1afe65e36a9f.mov

Jojo-Schmitz:
Probably those changes for 3.6 have not yet been migrated to master

Tantacrul:
Ah. I understand. I just saw the settings existed and thought something had gone wrong.

We are in the middle of porting over that entire converting old scores -> Leland+engraving improvements design right now.

DmitryArefiev:
Tested on master build on Win10, Mac11, LinuxMint 20 - PASS

PR #9602, opened 2021-10-28
Implemented Reset all styles to default button in EditStyle dialog

cbjeukendrup:
Resolves: #9464
Resolves: #9187

### DIFF
--- a/.github/workflows/ci_linux_mu4.yml
+++ b/.github/workflows/ci_linux_mu4.yml
@@ -41,6 +41,8 @@ jobs:
       run: |
         sudo bash ./build/ci/tools/setup_ccache.sh     
     - name: "Configure workflow"
+      env:
+        pull_request_title: ${{ github.event.pull_request.title }}
       run: |
         sudo bash ./build/ci/tools/make_build_mode_env.sh -e ${{ github.event_name }} -m ${{ github.event.inputs.build_mode }}
         BUILD_MODE=$(cat ./build.artifacts/env/build_mode.env)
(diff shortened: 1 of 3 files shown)


AntonioBL:
Apparently, the utests and vtests fail because of the commit "Enable vertical justification of staves by default".
Indeed, in the "failing" vtests the vertical space between staves has changed with respect to the reference build (and meybe it has changed for the better; note that vtests just point out differences with respect to the ref branch, but they cannot automatically judge what is better).
In the utests, the ref file contain an additional
      `<enableVerticalSpread>1</enableVerticalSpread>`
which does not appear in the files created by this PR.

The other build failures are maybe a temporary "hiccup" in the virtual machines. You can try to rebase so that the build restart.

Ciao,
ABL

Jojo-Schmitz:
So the utest ref files need to get adjusted, to not contain the new default setting.
While the vtest files might get 'corrected' to contain the old default setting. Or just ignore the failure 
Something is still wrong with the vtests though: I don't see the artifact.

cbjeukendrup:
I do see the artifact at https://github.com/musescore/MuseScore/actions/runs/1399562015. (But apparently GitHub removed the link in the top right corner in the log view?)

Jojo-Schmitz:
Ah, maybe that's just a change in the GutHub UI then?

cbjeukendrup:
The unit tests should be fixed now. For the vtests, we probably need the `vtests` label.

But I don't understand why the Linux/MacOS/Windows builds are failing...

Jojo-Schmitz:
Funny enough the vtests are fine now. I guess because they don't differ to the commit before. So as long as you don#t squash the cimmits the VTEST lable is not needed ;-)

cbjeukendrup:
No, the Vtests use the base branch as their reference, so in this case the current master.

The culprit may have been that I made a mistake in the legacy style files (I wrote `enableVerticalSpreading` instead of `enableVerticalSpread`, don't know why I did that 😅).

Jojo-Schmitz:
The build failures seem to be an issue for @igorkorsukov to investigate and solve?

igorkorsukov:
problem due to quotation marks in the title of the PR
`if [ "pull_request" == "pull_request" ]; then PR_INFO="_9602_Implemented "Reset all styles to default" button in EditStyle dialog"; fi`

Unfortunately, the renaming and restarting did not solve, the old title was remembered.
At the moment, needs to do `push` to restart the build with a new title.

And later I will try to fix this problem.

cbjeukendrup:
Thank you @igorkorsukov for investigating. I will fix it soon.

cbjeukendrup:
That last commit should fix the problem for future PRs that have quotes in the title.

DmitryArefiev:
Tested #9464 #9187 on master build on Win10, Mac11, LinuxMint 20 - PASS

PR #9608, opened 2021-10-29
[Experiment] Implemented sending and receiving MIDI 2.0 events for CoreMIDI (macOS)

cbjeukendrup:
In macOS 11.0 (autumn 2020), many CoreMIDI APIs were deprecated, in favor of new ones that support MIDI 2.0. Because MuseScore also works with MIDI 2.0, I thought this would be a perfect combination. So I updated the implementation of our CoreMidiInPort and CoreMidiOutPort, to let them send and receive MIDI 2.0 events. Because those new APIs are only available on macOS 11.0 and later, I added a version check, so that order macOS versions still use the MIDI 1.0 APIs.

@vpereverzev You might be interested in this PR.

The documentation about CoreMIDI is scarce, so I'm not completely sure whether this implementation is correct, but it seems to be in line with everything I could find.

I could only test it with MIDI 1.0 devices. That worked correctly (even better than the old implementation). So CoreMIDI automatically converts MIDI 2.0 events to 1.0 when sending to a 1.0 device.

It would be great if someone who has a MIDI 2.0 device could test whether this works.

(Also, I discovered some playback issues:
- When entering notes with mouse or keyboard, those notes are played on the connected MIDI keyboard, but not through MuseScore's playback / the computer's speakers. 
- When playing the score (using the play button in the playback toolbar) the score only plays through MuseScore's playback / the computer's speakers, but not through the connected MIDI keyboard.
- Entering notes with a MIDI keyboard has never worked for me and still doesn't, even though there is some log output indicating that MuseScore does receive the MIDI events.)

### DIFF
--- a/src/framework/midi/internal/platform/osx/coremidiinport.cpp
+++ b/src/framework/midi/internal/platform/osx/coremidiinport.cpp
@@ -94,26 +94,6 @@ async::Notification CoreMidiInPort::devicesChanged() const
     return m_devicesChanged;
 }
 
-static void proccess(const MIDIPacketList* list, void* readProc, void* srcConn)
-{
-    UNUSED(srcConn);
-
-    CoreMidiInPort* self = static_cast<CoreMidiInPort*>(readProc);
-    const MIDIPacket* packet = &list->packet[0];
-
-    for (UInt32 index = 0; index < list->numPackets; index++) {
-        if (packet->length != 0 && packet->length <= 4) {
-            uint32_t message(0);
-            memcpy(&message, packet->data, std::min(sizeof(message), sizeof(char) * packet->length));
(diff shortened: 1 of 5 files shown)


vpereverzev (approved):

vpereverzev:
Looks good to me. OS version check makes a lot of sense.

> So CoreMIDI automatically converts MIDI 2.0 events to 1.0 when sending to a 1.0 device.

In general, that's how MIDI2.0 UMP(Universal MIDI Package) works.

Although MIDI2.0 is still in development and will not be finally released very soon, I don't think the thirdparty API will change much over time. So it's better to move to the new version right away.

cbjeukendrup:
Thanks!

PR #9610, opened 2021-10-29
[MU4] muse_playback_events_introduction

vpereverzev:
This PR introduces our own playback event format called "MusePlaybackEvents". Goals and motivation are described here - https://musescore.org/en/node/325860

Generally speaking, this opens the way for us to improve the quality of playback closely to the real one. 

For community members:
There is an entire specification of this new format. But, unfortunately, I had no time to transfer it into our GitHub Wiki.
I'm going to transfer it later (probably after my short vacation). But for now you can get a basic idea from learning the source code and unit-tests in particular.

image

Example of concept:
image

### DIFF
--- a/src/framework/mpe/mpetypes.h
+++ b/src/framework/mpe/mpetypes.h
@@ -501,8 +501,8 @@ struct ExpressionCurve : public std::map<duration_percentage_t, dynamic_level_t>
     dynamic_level_t maxAmplitudeLevel() const
     {
         const auto& max = std::max_element(cbegin(), cend(), [](const auto& f, const auto& s) {
-                return f.second < s.second;
-            });
+            return f.second < s.second;
+        });
 
         if (max == cend()) {
             return 0.f;


src/framework/mpe/mpetypes.h
```diff
@@ -0,0 +1,516 @@
...
+
+namespace mu::mpe {
+// common
+using msecs_t = uint64_t;
+using percentage_t = float; // from 0.0 (0%) to 1.0 (100%)
+constexpr percentage_t PERCENTAGE_PRECISION_STEP = 0.1; // 10%
+
+// Arrangement
+using timestamp_t = msecs_t;
+using duration_t = msecs_t;
```
igorkorsukov, line 43: Maybe it is better to indicate this value in ticks? Then it will not depend on the tempo and, when the tempo changes, it will remain valid, there will be no need to recalculate and update it. For example, this can be important if we want to write the rendered events separately and synchronize them with the notation representation.

src/framework/mpe/events.h
```diff
@@ -0,0 +1,171 @@
...
+            m_expressionCtx.expressionCurve.emplace(pair.first, (pair.second * ratio) + appliedOffsetMap[pair.first]);
+        }
+    }
+
+    ArrangementContext m_arrangementCtx;
+    PitchContext m_pitchCtx;
+    ExpressionContext m_expressionCtx;
+};
+
+struct RestEvent
```
igorkorsukov, line 156: If `NoteEvent` and `RestEvent` do not have a base class, how are `NoteEvent` and `RestEvent` to be stored and transferred? Not on the same list one after the other? On two different lists?

src/framework/mpe/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+set(MODULE mpe)
+
+set(MODULE_SRC
+    ${CMAKE_CURRENT_LIST_DIR}/events.h
+    ${CMAKE_CURRENT_LIST_DIR}/mpetypes.h
+    )
+
+set(MODULE_QRC mpe.qrc)
```
igorkorsukov, line 28: I believe it's better to add `dummy.cpp` instead of an empty resource file :)

Jojo-Schmitz: Only that there is no such file?
Same is true BTW for src/framework/telemetry/CMakeLists.cpp and ...telemetry.qrc
Both empty qrc files lead to a build warning "No resources in ..."

vpereverzev: I just prepared this along the way. The actual resources would be introduced at on of the next PRs

src/framework/mpe/events.h
vpereverzev: It's going to be a single list of std::variant<NoteEvent, RestEvent>

src/framework/mpe/mpetypes.h
vpereverzev: I was thinking about this, but decided to not introduce any extra terms/conversions here. Because, in fact, I'll need to recalculate and update data anyway, even using an additional indexing. For example, if you change the duration of the note, you have to shift indexes of the rest of events. Also the data should be updated for any dynamic change as well.

So I decided to stick with milliseconds for now, since it allows to avoid unnecessary conversions, provides a more convenient way to handle complex articulations and it's a bit clearer to understand.

PR #9616, opened 2021-10-31
colornotes plugin - when only single note is selected, acts as nothing was selected

sammik:
If only single note is selected, act as nothing was selected.
When one edit score, ussually last entered, or edited note remains selected. 
In such cases plugin colors only that single note, but user wants to colors entire score.

### DIFF
--- a/share/plugins/colornotes.qml
+++ b/share/plugins/colornotes.qml
@@ -45,7 +45,7 @@ MuseScore {
       property string black : "#000000"
 
       // Apply the given function to all notes (elements with pitch) in selection
-      // or, if nothing is selected, in the entire score
+      // or, if nothing, or just signle note is selected, in the entire score
 
       function applyToNotesInSelection(func) {
             var fullScore = !(curScore.selection.elements.length > 1)
--- a/share/plugins/colornotes.qml
+++ b/share/plugins/colornotes.qml
@@ -48,7 +48,7 @@ MuseScore {
       // or, if nothing is selected, in the entire score
 
       function applyToNotesInSelection(func) {
-            var fullScore = !curScore.selection.elements.length
+            var fullScore = !(curScore.selection.elements.length > 1)
             if (fullScore) {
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
The convention so war was that when nothing is selected work on the entire score. Many plugins work that way, and also other finctions. like Tools > Transpose and Tools > Add/remove system breaks

Apart from that all these merge commits don#t belong hwere, you'd need to `git pull --rebase upstream master` to get rid of them (and than `git push -f`)

sammik:
Jojo, thank you for comment and tips.

I just changed 
"if nothing selected work on entire score" 
to 
"if nothing, or just single note selected work on entire score"

First makes sense as default convention / idea startpoint.
Second accept basic idea, but changes its adaptation for better user feeling, it makes sense in some cases, like, in my opinion, is coloring notes. 
User often accidentaly leaves selected single note (last created / edited) and in that cases plugin looks like it does nothing, as user doesnt see, it actually changed colour of that selected single note (which remains blue=selected) and in fact, user wanted to apply plugin to whole score.

May be better, next time, to discus it in forum first.

Jojo-Schmitz:
It is IMHO quite simple: if something/anything is selected, just work on that.
As a convenienece, to not have the user to explicitly select everything, if nothing is selected at all, work on the entire score.

You can make your own plugins behave differently (but I'd recommend against that) ,but for stuff thatb comes as part of MuseScore this should not be done, for conistency with other parts of the UI/UX, like the tools for transpose, switch voices 
 and add/remove system break.
I don't see at all why color notes should be different here. Or why you'd want it for colornotes but not for notenames and the tools mentioned above (esp. for the voice tool it'd be quite catastrophic)

sammik:
"if nothing is selected at all, work on the entire score"
sure, this is not changed

"I don't see at all why color notes should be different"
Answer is in comment above.

OK, I close it.

PR #9637, opened 2021-11-02
Added TC5: UsingInstruments

igorkorsukov:
It seems there are a couple of bugs...

https://user-images.githubusercontent.com/3818029/139877361-441d2642-ab03-4bd3-a515-8e664b2dff02.mp4

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/FocusableControl.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/FocusableControl.qml
@@ -43,6 +43,9 @@ FocusScope {
         }
     }
 
+    Component.onCompleted: navCtrl.enabled = true
+    Component.onDestruction: navCtrl.enabled = false
+
     NavigationControl {
         id: navCtrl
         name: root.objectName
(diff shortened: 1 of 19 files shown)


RomanPudashkin (approved):

src/framework/uicomponents/qml/MuseScore/UiComponents/FocusableControl.qml
```diff
@@ -43,6 +43,9 @@ FocusScope {
         }
     }
 
+    Component.onCompleted: navCtrl.enabled = true
```
cbjeukendrup, line 46: Shouldn't this have been
```qml
Component.onCompleted: navCtrl.enabled = Qt.binding(() => (root.enabled && root.visible))
```
?

igorkorsukov: Here I was solving another problem.
The fact is that the ListView can destroy delegates, but without physically deleting them as objects, that is, they remain in memory (as far as I know, this is done due to optimization to reuse delegate objects).
But in this case, objects of navigation controls also remain in memory and in the navigation tree. I found this when working with the list of instruments in the new score dialog or when adding a new instrument from the instruments panel. The list of navigation controls of the instruments was duplicated after several reloads of the model, because of this feature of the ListView

igorkorsukov: @cbjeukendrup I understand your point, you are right, I will correct it in the next PR. Thanks! :)

PR #9653, opened 2021-11-04
Fix #9352: Add ability to guess tie attachment point for different noteheads

### LINKED ISSUE #9352: [MU4 Issue] Properly determine 'midpoint' of noteheads for origin of ties
This is less critical for 'normal' noteheads, which are round, though as they are somewhat oblique the centre at the top is optically a little further right than the centre at the bottom.

However, it is important for slashes, for example:

slashes

The cutout metadata may be able to help with this; failing that, perhaps slashes might be special-cased.

**Should fix:**
https://musescore.org/en/node/325037

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/9352

Noteheads are now taken into account when determining where to attach ties. For noteheads with cutout information, this is the average of the cutouts. Also, now that noteheads have a single attachment point, ties adjust for each other when they share that point:
image

For slash noteheads, which are special, the tie attachment point takes the stem anchor into account:
image

For all other noteheads, the exact center is used.
image

### DIFF
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -1072,6 +1072,57 @@ qreal Note::headBodyWidth() const
     return headWidth() + 2 * bboxXShift();
 }
 
+//---------------------------------------------------------
+//   outsideTieAttachX
+//
+//    returns the X-position for tie attachment for this particular notehead
+//---------------------------------------------------------
+qreal Note::outsideTieAttachX(bool up) const
+{
+    qreal xo;
+
+    // special cases:
+    if (_headGroup == NoteHead::Group::HEAD_SLASH) {
+        // the anchors are really close to the stem attach points
+        xo = up ? symSmuflAnchor(noteHead(), SmuflAnchorId::stemUpSE).x() : symSmuflAnchor(noteHead(), SmuflAnchorId::stemDownNW).x();
+        xo += spatium() * 0.13 * (chord()->up() ? mag() : -mag());
+        return x() + xo;
+    }
+    if (_headGroup == NoteHead::Group::HEAD_SLASHED1 || _headGroup == NoteHead::Group::HEAD_SLASHED2) {
+        // just use the very center of the notehead
+        return x() + ((headBodyWidth() / 2) * mag());
+    }
+    /* Noteheads do not have optical centers at this time, but here's the code
+       to future-proof
+    xo = symSmuflAnchor(noteHead(), SmuflAnchorId::opticalCenter).x() * mag();
+    if (xo > 0) {
+        return x() + xo;
+    }
+    */
+    // try for average of cutouts
+    if (up) {
+        qreal xNE = symSmuflAnchor(noteHead(), SmuflAnchorId::cutOutNE).x();
+        qreal xNW = symSmuflAnchor(noteHead(), SmuflAnchorId::cutOutNW).x();
+        xo = ((xNE + xNW) / 2) * mag();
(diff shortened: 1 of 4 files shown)


Jojo-Schmitz:
3 vtests show differences, all with ties, unsurprisingly. 3 look OK, 3 don:t:
before:
image
after:
image
The tie in the 2nd measure is too high
before:
image
after:
image
In top staff the 1st tie in the 1st measure is too high, in bottom staff the last ties in the 1sxt meaure seem to ba to high/low
but the worst, before:
image
after
image
 the bottom tie starts way too late. It started to late before already, but now even later

Jojo-Schmitz:
Actually there's one more, before:
image
After:
image
See the tie for the 1st grace to note and  2 gracenotes at the end, they are too far to the right

asattely:
Yup, I noticed those as well. Thanks for looking through and getting the screenshots, I'll definitely be taking another whack at this once I finish something I'm working on tomorrow.

asattely:
So, the extra height on some of the ties were due to the fact that the vtests still use Emmentaler rather than Leland, for the record. The vtests will almost certainly still fail because the ties will look different, but now the ties from gracenotes will be a lot nicer with a little special handling. The wholenotes one was a weird oversight that should be fixed now. Let's see what the vtests say on this one.

Jojo-Schmitz:
Looks better, but this here does not, before:
image
after:
image
The 2st tie looks better, the other 3 don't, they go too far past the note they tie to

asattely:
Yep, very well aware. The problem is that headWidth (and, bafflingly, the bounding box as well) always returns the width of a regular-sized notehead instead of the grace-sized one, so I just need to figure out how to get the grace-sized head width and we're golden.

Jojo-Schmitz:
See (in 3.x) Format > Style > Sites > Grace note size. Defaults to 70%, so you'ds need to pick on that style setting

asattely:
The answer was `mag()` (or `magS()`) which holds information about the note's relative size

asattely:
Those are inside-style ties because they are going to or from a multinote chord. This is an issue in other inside-style tie situations as well, and in the future there will be a minimum tie width issue where that will be looked at in greater detail.

asattely:
Still messing around with different fonts and noteheads, there are so many combinations to go through..!

asattely:
Rebase finished. I was also requested to add a small left- or right-offset to outside tie endpoints, which I ended up making 1/8 sp

Jojo-Schmitz:
4 utests crash with a failed assertion:
ASSERT: "_anchor == Anchor::CHORD" in file .../src/engraving/libmscore/spanner.cpp, line 879

asattely:
Tests should be fixed now

Jojo-Schmitz:
Nope, exact same crash

asattely:
The weird thing is that these tests pass when I run them locally. I think I identified the problem, we'll see if the tests are good this time

AntonioBL:
The vtests are now better, except for "grace-5-1".
Before
grace-5-1 ref
After
grace-5-1
First, second and third ties are better than the reference, but there is something strange in the vertical position of the tie between the two grace notes (i.e. the fourth tie in the file).

Ciao

Nick-Mazuk:
Does this work as expected on chords where every note has a different notehead?

asattely:
Fixed a little issue with forced-direction outside ties colliding with hooks, and also added another special case for slashed noteheads, which are different from slash noteheads for the record.

oktophonie (approved):

PR #9662, opened 2021-11-07
Fix #325735: [MusicXML import] Final barlines wrongly and only partia…

lvinken:
…lly assumed from musicxml import lacking barlines

Resolves: https://musescore.org/en/node/325735

Ensures all staves end with a normal barline, if no barline is present in the MusicXML file. The current implementation handles the first staff only, incorrectly resulting in a normal barline in the first staff and a final barline in every other staff. 
This was not caught by the autotester, as the assumption in Measure::endBarLineType() that all barlines have the same type results in the expected MusicXML output anyway. Requested by the music21 project.


RomanPudashkin:
Please rebase your PR to the latest master

lvinken:
Rebase done

Issue #9654, opened 2021-11-04
[MU4 Task] Remove 'beam distance' setting and replace with toggle switches (Styles dialog)


bkunda:
**Task description**
We'd like to propose a minor redesign of the **Beams** settings in the **Styles** dialog.
This would see the removal of the text entry field for "Beam distance (relative to beam thickness)", replacing it with toggle switches for "Regular" and "Wide" distances. This would also entail some subtle tweaks to the UI.
 
**Current screen**
<img width="1120" alt="Screenshot 2021-11-04 at 8 36 55 pm" src="https://user-images.githubusercontent.com/86290556/140291360-42cc287e-a6b3-46fa-a1f8-fe23f04543e0.png">

**New design**
<img width="929" alt="Option2" src="https://user-images.githubusercontent.com/86290556/140291485-60c83927-819c-475b-b9de-e41bff7a7ea3.png">

The settings for these switches would be (as if they were entered in MS3.6): 

- Regular: 50%
- Wide: 100%

Figma design file: https://www.figma.com/file/l0TQUrNFZmEfrZnfHhFDfb/Musescore-Notation-App-MASTER?node-id=19024%3A632206

Please tag @oktophonie with any further questions about how this should function.

Thanks!
Bradley

cbjeukendrup:
That looks like another one I could work on (when my exams are done 😉)

I will probably build some general infrastructure to be able to write Style dialog pages in QML. I already wrote things for the Notes page in particular (for the Flag Styles), but now that we will rewrite more pages in QML, it is worth it to write reusable models.

bkunda:
Thanks @cbjeukendrup ☺️
We'd like to get on to this one pretty soon as I believe it's a necessary feature within a broader beams project @oktophonie is working on. 
But I know our devs are pretty stretched at the moment, so if you can lend a hand then we'd be grateful! I'm just not sure who will end up getting to it first 😉
FWIW, we would like to tackle the entire styles dialog once MS4 is released, and so it could end up being quite a different concept, with a completely different UX/UI in future. This being said, I expect your proposal to rewrite any updates to this dialog in QML makes a lot of sense. I just thought it was worth mentioning this (FYI 😁).

cbjeukendrup:
I'm working on it now, and hope to open a PR tonight!

PR #9683, opened 2021-11-09
Implemented redesigned Beams page of EditStyle dialog

cbjeukendrup:
Resolves: #9654
Schermafbeelding 2021-11-10 om 01 02 21


RomanPudashkin (approved):

bkunda:
Hi @cbjeukendrup!
Wonderful work, as always.
Just a couple of small points from my end:

1. The 'flatten all beams' checkbox no longer toggles correctly:
flatten-beams-doesn't-work

2. The 16th-note icons look a tad larger than I had in the Figma file (28pt). I'd prefer them to be at this size if possible, but if it's a question of uniformity with other similar buttons (e.g. with the straight flags toggle buttons), then I'd understand! 

Many thanks,
Bradley

cbjeukendrup:
Thanks! To be honest, for those radio buttons I just copied the code from the note flags type section, where the icons are 32px. But indeed, on the Beams page they look much better at 28px.
I also noticed that the title of the Beams distance section was wrong (it said "Beam thickness" instead of "Beam distance"), so I just created a small follow-up pull request. 

The "Flatten all beams" checkbox also didn't work before. I think that this is more task for @Nick-Mazuk, who is working on all kinds of beaming problems, if I'm not mistaken.

bkunda:
Good spot on that title discrepancy! I'm quietly ashamed I missed that one...🙄
Thanks for raising this follow-up PR 👍🏻
As for that dodgy checkbox, we'll leave it in @Nick-Mazuk's very capable hands (unless he suggests otherwise). 
Many thanks!
Bradley

Nick-Mazuk:
Yep, I'll make sure all these options work with the new beaming code!

Nick-Mazuk:
Once https://github.com/musescore/MuseScore/pull/9934 and https://github.com/musescore/MuseScore/pull/9937 are merged, the beams code will fully work with with these new options.

PR #9706, opened 2021-11-10
Fix #9444 key signature mode property update process

### LINKED ISSUE #9444: [Inspector] 'Mode' dropdown for key signatures non-functional (both in MS3.6 and MS4.0)
**Describe the bug**
It seems this feature has not been working for a while. Here is an old bug: https://musescore.org/en/node/304086

Currently if you choose any option and then click away from the key signature and then click back on it, the dropdown defaults back to 'unknown'.

image

### Desired to fix

1. This is intended to specify the key signature mode for XML import/export. Since it is not currently functioning - and does not appear to have functioned in a long time, I suggest hiding the UI for now.
2. Alternatively, if a community member wants to tackle it, we can assign the task to them. If fixing this issue, it would be wise to alter the description to make it clear that it is an XML-related feature only. Currently, its intended function is not obvious at all.

LovCAPONE:
Resolves: *https://github.com/musescore/MuseScore/issues/9444*

Update the staff's key signature associated with the key signature
selected in the Inspector whenever a property is modified.

The problem is that whenever the user changes the key signature mode
property in the Inspector, the value in the staff's key signature at
this tick is not updated. So, in the score update process when the key
signature mode is fetched the value returned is the default key
signature mode.


Jojo-Schmitz (approved):

LovCAPONE:
@Jojo-Schmitz This PR is ready for review.

Jojo-Schmitz (approved):

Jojo-Schmitz:
It is fine by me and does indeed fix the issue.

RomanPudashkin (approved):

PR #9731, opened 2021-11-13
Added the print icon

HemantAntony:
Resolves: It doesn't resolve any issue. There wasn't a print icon near the print button. I added it. I didn't sign the CLA because it just took 2 lines to fix. For the same reason, I didn't do any tests.

Before it used to be like this:

Before

After adding the code:

After


HemantAntony:
Also, are there supposed to be dots after "print"?

Jojo-Schmitz:
> Also, are there supposed to be dots after "print"?

Yes, as there's a dialog following (it is actually an ellipsis)

HemantAntony:
> Yes, as there's a dialog following (it is actually an ellipsis)
Oh ok

PR #9735, opened 2021-11-15
Added ability select TC in the list, and speed mode to runner

igorkorsukov:
image


src/autobot/view/autobotscriptsmodel.h
```diff
@@ -89,6 +93,7 @@ public slots:
     bool m_isRunAllTCMode = false;
     QMap<io::path, IAutobot::Status> m_statuses;
     QMap<int, bool> m_selected;
+    QString m_speedMode;
```
Eism, line 96: not used

igorkorsukov: removed, thanks!

Eism (approved):

Issue #9737, opened 2021-11-15
[Learn Section] Thumbnails for YouTube videos are the wrong size / ratio


Tantacrul:
**Describe the bug**
In the learn section, the thumbnails appear to be at a 5:3 ratio which is narrower that the standard for YouTube thumbnails: 16:9.

Below, I've compared a thumb from MS4 (above) and the same thumb from YouTube (below)

![image](https://user-images.githubusercontent.com/21022311/141764280-758b4980-887d-4700-b90b-1b3018630811.png)

This will cause our thumbnails to look distorted when we begin creating tutorial videos.

Essentially, the height of the thumbnails is roughly fine but they should be made wider to fit an exact 16:9 ratio (and it should be an exact integer too). Measured on my Mac (high res), the thumbs appear to be 500x300. Perhaps they could be changed to 512x288?

jessjwilliamson:
I agree absolutely fine to go ahead and fix the ratio. 256x144 looks good just did a test on Figma. That would also mean 50 pixels spacing in between instead of 56 if we want to keep the width of the overall row the same.
![image](https://user-images.githubusercontent.com/71656189/141770677-22851f17-fb5c-4cb1-b659-522c7407a66e.png)

Issue #9690, opened 2021-11-10
[MU4 Task] Reduce height of bottom bar


jessjwilliamson:
Recently it has been reduced (from 36 to 32 pixels), but decided to reduce it even further to 28 pixels due to vertical spacing issues with the mixer.

I've done some tidying up and made sure it works with hover buttons as well.
![image](https://user-images.githubusercontent.com/71656189/141109425-04c02ced-e7d9-48a1-9372-4be1f217b828.png)

All of this can be found in the 'BOTTOM INFO BAR' Figma page

Issue #9223, opened 2021-09-22
[MU4 Issue] Mixer - Volume number meters glitch when change font in Preferences


jessjwilliamson:
**Describe the bug**
There seems to be a strange bug where you change the font in Preferences, the numbers on the volume meter looks glitchy like the replaced font is overriding original. It's not changing as it should be. The right side of the volume meter labels behave fine, the left side not.

https://user-images.githubusercontent.com/71656189/134336945-c2044e71-4954-499a-a750-1696c7da8eeb.mov

**Desktop (please complete the following information):**
 - OS: MacOS

cbjeukendrup:
I will submit a fix soon!

PR #9740, opened 2021-11-15
Another round of UI finesse fixes

cbjeukendrup:
Resolves: #9737
Resolves: #9223 
Resolves: #9690 

NOTE: I couldn't really test the Learn page because I have no YouTube API key, so I see no thumbnails at all.

### DIFF
--- a/src/framework/audio/qml/MuseScore/Audio/VolumeSlider.qml
+++ b/src/framework/audio/qml/MuseScore/Audio/VolumeSlider.qml
@@ -228,7 +228,7 @@ Slider {
                 ctx.textAlign = "end"
             }
 
-            ctx.clearRect(0, 0, bgCanvas.width, bgCanvas.height)
+            ctx.clearRect(0, 0, bgCanvas.height, bgCanvas.width)
             ctx.font = prv.unitTextFont
 
             ctx.fillStyle = Utils.colorWithAlpha(ui.theme.fontPrimaryColor, 0.5)
--- a/src/notation/qml/MuseScore/NotationScene/ConcertPitchControl.qml
+++ b/src/notation/qml/MuseScore/NotationScene/ConcertPitchControl.qml
@@ -33,9 +33,12 @@ FlatButton {
 
     orientation: Qt.Horizontal
     transparent: true
+    margins: 4
 
     contentItem: Row {
         spacing: 6
+
+        width: implicitWidth + 4 // some extra right padding
 
         CheckBox {
             checked: root.checked
--- a/src/notation/qml/MuseScore/NotationScene/ZoomControl.qml
+++ b/src/notation/qml/MuseScore/NotationScene/ZoomControl.qml
@@ -49,6 +49,8 @@ RowLayout {
         icon: IconCode.ZOOM_IN
         iconFont: ui.theme.toolbarIconsFont
 
+        width: height
+        height: 28
         transparent: true
 
         navigation.panel: root.navigationPanel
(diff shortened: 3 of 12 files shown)


src/notation/qml/MuseScore/NotationScene/ConcertPitchControl.qml
```diff
@@ -33,10 +33,13 @@ FlatButton {
 
     orientation: Qt.Horizontal
     transparent: true
+    margins: 4
 
     contentItem: Row {
         spacing: 6
 
+        width: implicitWidth + 4 // some extra right padding
```
RomanPudashkin, line 41: As far as I understand, you can rewrite your code like this:

width: implicitWidth + root.margins

src/framework/audio/qml/MuseScore/Audio/VolumeSlider.qml
```diff
@@ -228,7 +228,7 @@ Slider {
                 ctx.textAlign = "end"
             }
 
-            ctx.clearRect(0, 0, bgCanvas.width, bgCanvas.height)
+            ctx.clearRect(0, 0, bgCanvas.height, bgCanvas.width)
```
RomanPudashkin, line 231: Could you please explain what is the purpose of this change?

src/notation/qml/MuseScore/NotationScene/ZoomControl.qml
```diff
@@ -49,6 +49,8 @@ RowLayout {
         icon: IconCode.ZOOM_IN
         iconFont: ui.theme.toolbarIconsFont
 
+        width: 28
+        height: 28
```
RomanPudashkin, line 53: Better to use bindings in such situations:

width: 28
height: width

src/framework/audio/qml/MuseScore/Audio/VolumeSlider.qml
cbjeukendrup: It's because we're in a rotated world here, so width and height are reversed. 
(In the lines before this, the `ctx` is rotated.)

jessjwilliamson:
Thumbnails have disappeared in Learn page!

cbjeukendrup:
> Thumbnails have disappeared in Learn page!

You mean the Learn page is completely empty? That's right, because test builds for pull requests don't have access to the Youtube API key, so therefore they can't download the playlist from Youtube. So we'll probably see the result only in nightlies, when this PR is merged.

jessjwilliamson:
> > Thumbnails have disappeared in Learn page!
> 
> You mean the Learn page is completely empty? That's right, because test builds for pull requests don't have access to the Youtube API key, so therefore they can't download the playlist from Youtube. So we'll probably see the result only in nightlies, when this PR is merged.

Ah! that's fine :) I can see changes on the plugins and that looks good!

src/notation/qml/MuseScore/NotationScene/ZoomControl.qml
cbjeukendrup: Done!

src/notation/qml/MuseScore/NotationScene/ConcertPitchControl.qml
cbjeukendrup: Semantically, not really, I think. That extra padding is not necessarily meant to be the same as the normal margin.

RomanPudashkin (approved):

Issue #9508, opened 2021-10-20
Staff/part properties has buttons with missing icons


Tantacrul:
**Describe the bug**
(Mac OS - Catalina)

Right click on a measure and choose the Staff/part properties option. Observe how some icons are missing:

![image](https://user-images.githubusercontent.com/21022311/138056188-d3849424-a7e3-471a-a478-4be8180965c2.png)

Here is the same dialog from MS3
![image](https://user-images.githubusercontent.com/21022311/138056287-4ae86885-5be8-4794-812e-00ca47b2aa80.png)

@jessjwilliamson - can you specify (and make if needed) the correct icons and unicodes here?

jessjwilliamson:
The 4 icons within the Usable pitch range should be our pencil icon - Unicode EF63

The bottom left ones are arrows- Unicode EF37 and EF36 (up then down)

Issue #9430, opened 2021-10-12
Time Signature Properties popup: Icons not displaying and one dropdown not functioning


Tantacrul:
**Describe the bug**
Two options from the 'Time Signature Properties' panel are not displaying / working in MS4 (see image)

1. The icons for 'Beam properties'
2. The options in the 'Other' dropdown

Please see MuseScore 3 to determine how these should look. If there is a need for the design team to create icons for 'Beam properties' in our icon font, then let us know and we will create them.

![image](https://user-images.githubusercontent.com/21022311/136944330-aec0fed5-4a40-41dc-b450-23b64cd49735.png)

cbjeukendrup:
I just had another idea: how about moving at least those "appearance" settings to the Inspector? I suddenly thought it might actually make sense to have them there, instead of hidden in a dialog.

PR #9741, opened 2021-11-15
Fix icons in all QWidget dialogs (on macOS)

cbjeukendrup:
Resolves: #9430
Resolves: #9508

The key was: use `setStyleSheet` instead of `setFont`. The latter does not work on macOS.


src/notation/view/widgets/editstyle.cpp
```diff
@@ -68,6 +63,17 @@ static void fillDirectionComboBox(QComboBox* comboBox)
     comboBox->addItem(Ms::toUserString(Ms::Direction::DOWN), QVariant::fromValue<Ms::Direction>(Ms::Direction::DOWN));
 }
 
+template<class T>
+void EditStyle::setWidgetIcon(T* widget, IconCode::Code code)
```
RomanPudashkin, line 67: This function looks quite useful. It seems that we can reuse it in other widget dialogs. I suggest creating a new class with this function as a static method.

cbjeukendrup: Yeah, that was what I was planning to do. Indeed, we also need it for #9508, which I will also fix.

RomanPudashkin (approved):

PR #9743, opened 2021-11-16
Fix #9175 Cannot open multiple score parts at once

### LINKED ISSUE #9175: Parts Dialog: cannot open multiple part scores at once
**Describe the bug**
The user can multiselect scores... but can only open one at a time, which can be quite painful.

**See video**
https://www.dropbox.com/s/18pvp22xxmjvu0g/Parts_Can%27t_Open_Multiple.mov?dl=0

HemantAntony:
Resolves: https://github.com/musescore/MuseScore/issues/9175

After selecting multiple parts in "parts" dialog, you can now open all of them. Previously, you could open only one. Therefore, you would have to open all of them one-by-one, causing an inconvenience. So, I created a loop which iterates through the selected rows and opens them. I have also sorted the selected rows in the order in which they appear, so that when you open them, it opens in the correct order.

### DIFF
--- a/src/notation/view/partlistmodel.cpp
+++ b/src/notation/view/partlistmodel.cpp
@@ -279,6 +279,8 @@ void PartListModel::openSelectedParts()
         return;
     }
 
+    std::sort(rows.begin(), rows.end());
+
     ExcerptNotationList newExcerpts;
     for (int index : rows) {
         newExcerpts.push_back(m_excerpts[index]);
@@ -286,6 +288,9 @@ void PartListModel::openSelectedParts()
 
     masterNotation()->addExcerpts(newExcerpts);
 
+    for (int index : rows) {
+        m_excerpts[index]->notation()->setOpened(true);
+    }
     context()->setCurrentNotation(m_excerpts[rows.last()]->notation());
 }
 


src/notation/view/partlistmodel.cpp
```diff
@@ -279,14 +279,18 @@ void PartListModel::openSelectedParts()
...
     ExcerptNotationList newExcerpts;
     for (int index : rows) {
         newExcerpts.push_back(m_excerpts[index]);
     }
 
     masterNotation()->addExcerpts(newExcerpts);
 
-    context()->setCurrentNotation(m_excerpts[rows.last()]->notation());
+    for (int index : rows) {
+        context()->setCurrentNotation(m_excerpts[i]->notation());
```
RomanPudashkin, line 292: Better to use INotation::setOpened here:

 for (int index : rows) {
 m_excerpts[i]->notation()->setOpened(true);
 }
 
setCurrentNotation should be used only once (after that cycle with the last notation in the selected list) because it is quite expensive:
 
 context()->setCurrentNotation(m_excerpts[rows.last()]->notation());

HemantAntony: Ok will do that. Thanks!

RomanPudashkin (approved):

DmitryArefiev:
Tested #9175 on latest master build (Win10, Mac12, LinuxMint 20.2) - FIXED

PR #9750, opened 2021-11-16
Replaced QVariant to PropertyValue for properties

igorkorsukov:


src/engraving/libmscore/beam.cpp
```diff
@@ -267,17 +267,17 @@ bool Beam::isNoSlope() const
 
 void Beam::alignBeamPosition()
 {
-    BeamPos currentBeamPos = beamPos();
+    PointF currentBeamPos = beamPos();
```
Jojo-Schmitz, line 270: As this is not really about x and y coordinates of one of a beams handles. but more about the left and right y-psotions, the `QPair` approach seems to make more sense, doesn't it?

src/engraving/libmscore/beam.cpp
```diff
@@ -267,17 +267,17 @@ bool Beam::isNoSlope() const
...
 
-    qreal currentY1 = currentBeamPos.first;
-    qreal currentY2 = currentBeamPos.second;
+    qreal currentY1 = currentBeamPos.x();
+    qreal currentY2 = currentBeamPos.y();
```
Jojo-Schmitz, line 273: the naming of the variables here make this particularly clear

src/engraving/libmscore/beam.cpp
```diff
@@ -2448,8 +2448,8 @@ void Beam::setBeamPos(const BeamPos& bp)
     _userModified[idx] = true;
     setGenerated(false);
     qreal _spatium = spatium();
-    f->py1[idx] = bp.first * _spatium;
-    f->py2[idx] = bp.second * _spatium;
+    f->py1[idx] = bp.x() * _spatium;
+    f->py2[idx] = bp.y() * _spatium;
```
Jojo-Schmitz, line 2452: here too

src/engraving/libmscore/dynamic.h
```diff
@@ -86,10 +54,8 @@ class Dynamic final : public TextBase
         const char* name;
     };
 
-    Q_ENUM(Type);
```
Jojo-Schmitz, line 89: Isn't thsat needed to make them available to plugins?

src/engraving/libmscore/textlinebase.h
```diff
@@ -68,14 +68,6 @@ class TextLineBaseSegment : public LineSegment
...
 };
 
-//---------------------------------------------------------
-//   HookType
-//---------------------------------------------------------
-
-enum class HookType : char {
-    NONE, HOOK_90, HOOK_45, HOOK_90T
-};
-
```
Jojo-Schmitz, line 78: Needed for plugins?

src/engraving/libmscore/textlinebase.h
```diff
@@ -126,17 +118,17 @@ class TextLineBase : public SLine
...
 
-    virtual QVariant getProperty(Pid id) const override;
-    virtual bool setProperty(Pid propertyId, const QVariant&) override;
-    virtual Pid propertyId(const QStringRef& xmlName) const override;
+    mu::engraving::PropertyValue getProperty(Pid id) const override;
+    bool setProperty(Pid propertyId, const QVariant&) override;
+    Pid propertyId(const QStringRef& xmlName) const override;
 };
 }     // namespace Ms
 Q_DECLARE_METATYPE(Ms::HookType);
```
Jojo-Schmitz, line 134: and referenced here?

src/engraving/libmscore/types.h
```diff
@@ -664,6 +664,46 @@ enum class TupletBracketType : char {
...
+    SFPP,
+    RFZ,
+    RF,
+    FZ,
+    M,
+    R,
+    S,
+    Z
+};
+
```
Jojo-Schmitz, line 706: Ah, I see, now here!

src/engraving/libmscore/types.h
```diff
@@ -678,6 +718,7 @@ Q_ENUM_NS(NoteType);
 Q_ENUM_NS(PlayEventType);
 Q_ENUM_NS(AccidentalType);
 Q_ENUM_NS(HarmonyType);
+Q_ENUM_NS(DynamicType);
```
Jojo-Schmitz, line 721: Missing for `HookType`?

src/engraving/libmscore/types.h
```diff
@@ -703,5 +744,6 @@ Q_DECLARE_METATYPE(Ms::PlayEventType);
 Q_DECLARE_METATYPE(Ms::AccidentalType);
 
 Q_DECLARE_METATYPE(Ms::HPlacement);
+Q_DECLARE_METATYPE(Ms::DynamicType)
```
Jojo-Schmitz, line 747: same?

src/engraving/libmscore/beam.cpp
Jojo-Schmitz: `(Q)Point` was the hack here, not QPair`

igorkorsukov: I agree. I will make it separate PR. There you don't just need to change the value type, but probably introduce a new type in `P_TYPE`, because now `Pid::BEAM_POS` is `P_TYPE::POINT`. 
https://github.com/musescore/MuseScore/blob/master/src/engraving/libmscore/property.cpp#L148

src/engraving/libmscore/dynamic.h
igorkorsukov: I'll take care of this in a separate PR

src/engraving/libmscore/textlinebase.h
igorkorsukov: I'll take care of this in a separate PR

src/engraving/libmscore/textlinebase.h
igorkorsukov: I'll take care of this in a separate PR

src/engraving/libmscore/types.h
igorkorsukov: I'll take care of this in a separate PR

src/engraving/libmscore/types.h
igorkorsukov: I'll take care of this in a separate PR

Jojo-Schmitz:
I don't believe the vtest failure is to be expected here, there are 6 tests failing, all in the same way, with the final barline, so just one example.
Before:
image
After:
image

It seems that both, before and after are unexpected, the 'before' images should show a proper end barline (thin-thick), the after images do show them but with too much distance between thin and thick?

igorkorsukov:
> I don't believe the vtest failure is to be expected here, there are 6 tests failing, all in the same way, with the final barline, so just one example. Before: image After: image
> 
> It seems that both, before and after are unexpected, the 'before' images should show a proper end barline (thin-thick), the after images do show them but with too much distance between thin and thick? Even stranger: only 6 vtests are affected, but many more to contain end barlines!?! Here's how that one looks in 3.x: harmony-10-ref

Thanks! I just started to investigate what is the problem

vpereverzev (approved):

src/engraving/libmscore/groups.h
```diff
@@ -24,7 +24,7 @@
 #define __GROUPS__
 
 #include "fraction.h"
-#include "mscore.h"
+//#include "mscore.h"
```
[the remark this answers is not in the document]
igorkorsukov: yes!

igorkorsukov:
> I don't believe the vtest failure is to be expected here, there are 6 tests failing, all in the same way, with the final barline, so just one example. Before: image After: image
> 
> It seems that both, before and after are unexpected, the 'before' images should show a proper end barline (thin-thick), the after images do show them but with too much distance between thin and thick? Even stranger: only 6 vtests are affected, but many more to contain end barlines!?! Here's how that one looks in 3.x: harmony-10-ref

I found, what happened:
* I broke end barline with my previous PR (https://github.com/musescore/MuseScore/pull/9738)
* vtest not triggered on this, because it crashed 
from vtest generate current log
```
17:05:26.025 | INFO  | main_thread | ConverterController | fileConvert: in: ./vtest/scores/harmony-10.mscx, out: vtest.artifacts/current/harmony-10.png
mscore: ../src/engraving/property/propertyvalue.h:65: T mu::engraving::PropertyValue::value() const [with T = double]: Assertion `pv' failed.
[11254:11254:20211115,170526.033316:ERROR elf_dynamic_array_reader.h:61] tag not found
```
PropertyValue had type `Spatium`, but tried get `double`
* Distance between end lines it was same, as this PR, I checked this on night build, builded 7 day ago 

So, we should: 
* fix vtest, on crash it should be fail 
* add issue about distance between end bar lines (@DmitryArefiev FYI)
* this PR can be merge

Jojo-Schmitz:
Are you going to create issues for the vtests to fail on a crash and the end barline?

igorkorsukov:
> Are you going to create issues for the vtests to fail on a crash and the end barline?

Added:
* https://github.com/musescore/MuseScore/issues/9777
* https://github.com/musescore/MuseScore/issues/9778

njvdberg:
This PR seems to cause a crash when opening the style editor, see also #9793.

Issue #9736, opened 2021-11-15
[MU4 Task] Enable the window state restoring


RomanPudashkin:
See: https://github.com/musescore/MuseScore/pull/8593

Should resolve:
https://github.com/musescore/MuseScore/issues/9550
https://github.com/musescore/MuseScore/issues/9652

Issue #9652, opened 2021-11-03
[MU4 Issue] Note input bar glitch when switching to vertical docking


jessjwilliamson:
**Describe the bug**
When docking note input bar from horizontal to vertical, it's fine at first, then switch to Home then back to Score, it's in 'vertical' view but taking up the whole horizontal dock space.

https://user-images.githubusercontent.com/71656189/140166183-fa300cb5-cf44-4533-a6b3-a38eb7e286af.mp4

**Desktop (please complete the following information):**
 - OS: Windows

Issue #9550, opened 2021-10-25
[MU4 Issue] Dock panels location reset after changing main tabs


Marr11317:
see https://youtu.be/0Y0X10yiOYg?t=598

PR #9764, opened 2021-11-17
[MU4] State restoring

RomanPudashkin:
Resolves:
https://github.com/musescore/MuseScore/issues/9736
https://github.com/musescore/MuseScore/issues/9550
https://github.com/musescore/MuseScore/issues/9652

### DIFF
--- a/src/appshell/internal/startupscenario.cpp
+++ b/src/appshell/internal/startupscenario.cpp
@@ -23,6 +23,7 @@
 #include "startupscenario.h"
 
 #include "async/async.h"
+#include "log.h"
 
 using namespace mu::appshell;
 using namespace mu::actions;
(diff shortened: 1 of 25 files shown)


src/appshell/internal/startupscenario.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "startupscenario.h"
 
+#include "log.h"
 #include "async/async.h"
```
igorkorsukov, line 26: I ask you to adding `log.h` at the very end.
This is due to the fact that many defines are defined in `log.h.
There is a possibility that they will override the definitions defined in other files, which we may not immediately notice and can lead to bad consequences.

RomanPudashkin: fixed

src/appshell/internal/startupscenario.cpp
```diff
@@ -64,39 +65,39 @@ void StartupScenario::setStartupScorePath(const io::path& path)
...
+    StartupSessionType sessionType;
+    if (!m_sessionType.isEmpty()) {
+        sessionType = sessionTypeTromString(m_sessionType);
+    } else {
+        sessionType = configuration()->startupSessionType();
+    }
+
+    interactive()->open(startupPageUri(sessionType));
+
+    switch (sessionType) {
```
igorkorsukov, line 84: It looks like these actions in the switch should be performed after the page is opened, that is, an excellent solution would be to add a signal (channel) to interactive, which will notify that the page has opened and make these actions to the handlers of this notification. Well, or at least make them asynchronously at the next event loop.

src/appshell/view/dockwindow/dockwindow.cpp
```diff
@@ -36,11 +36,48 @@
...
 
+namespace mu::dock {
+static const QList<Location> POSSIBLE_LOCATIONS {
+    Location::Left,
+    Location::Right,
+    Location::Top,
+    Location::Bottom
+};
+
+KDDockWidgets::Location locationToKLocation(Location location)
```
igorkorsukov, line 53: must be at least static

src/appshell/view/dockwindow/dockwindow.cpp
```diff
@@ -36,11 +36,48 @@
...
+    case Location::Top: return KDDockWidgets::Location_OnTop;
+    case Location::Bottom: return KDDockWidgets::Location_OnBottom;
+    case Location::Center: break;
+    case Location::Undefined: break;
+    }
+
+    return KDDockWidgets::Location_None;
+}
+
+void clearRegistry()
```
igorkorsukov, line 67: must be at least static

src/appshell/view/dockwindow/dockwindow.cpp
RomanPudashkin: fixed

src/appshell/view/dockwindow/dockwindow.cpp
RomanPudashkin: fixed

src/appshell/internal/startupscenario.cpp
RomanPudashkin: OK, I'll add a signal, but I want to do this in a separate PR

src/appshell/view/dockwindow/internal/dockwindowactionscontroller.cpp
```diff
@@ -37,6 +37,7 @@ void DockWindowActionsController::init()
     dispatcher()->reg(this, "set-dock-open", this, &DockWindowActionsController::setDockOpen);
     dispatcher()->reg(this, "toggle-dock", this, &DockWindowActionsController::toggleOpened);
     dispatcher()->reg(this, "toggle-floating", this, &DockWindowActionsController::toggleFloating);
+    dispatcher()->reg(this, "restore-default-layout", this, &DockWindowActionsController::restoreDefaultLayout);
```
igorkorsukov, line 40: It would be great if all these actions started with `dock-`. So it would be easier to see what these actions are and there would be less of conflict. For example, `restore-default-layout` can also be in the context of score.

RomanPudashkin: renamed

igorkorsukov (approved):

DmitryArefiev:
Tested on latest master build on Win10, Mac12, LinuxMint 20 - FIXED

PR #9766, opened 2021-11-17
[MU4] mpe_articulation_profiles_editor

vpereverzev:
This PR introduces a simple utility "MPE Articulation Profile Editor". For now it's supposed to be used internally, but in future we'll be able to provide users with this functionality.

image

Here is a small example of defining a "scoop" articulation:
scoop_sample


src/framework/mpe/internal/stringutils.h
```diff
@@ -0,0 +1,177 @@
+/*
```
igorkorsukov, line 1: I believe needs another name for this file (we already have a file with the name `stringutils.h`), like `articulationstrings.h`

igorkorsukov (approved):

src/framework/mpe/internal/stringutils.h
vpereverzev: makes sense to me, renamed

Issue #9023, opened 2021-09-03
[MU4 Issue] Rationalise stem shortening outside of stave


oktophonie:
We can already progressively shorten stems outside the stave using the style options:

![style](https://user-images.githubusercontent.com/20604032/132006950-5ae960a1-16c7-4744-a1e1-5dfc4dd64466.png)

We need to also be able to:

- specify at what point the shortening begins (after x amount of extension beyond the stave);
- cap the amount of shortening based on the number of flags present; additionally, current behaviour is a little inconsistent between 1 and 2+ flags.

**Should fix:**
https://musescore.org/en/node/17697 (if that still counts as open!)

Issue #9494, opened 2021-10-19
[MU4 Issue] Revise beam placement algorithm


oktophonie:
Within the stave, all beams should straddle a line, hang from it, or sit on it (not hover in the middle).

![beams1](https://user-images.githubusercontent.com/20604032/137855824-b90d4663-6ffe-49fa-8907-8f07f0d2f6da.png)

When slanting beams, avoid wedges between stem, stave line and the inside of beams:

![beams2](https://user-images.githubusercontent.com/20604032/137855844-f84c271c-2d27-4f49-91ad-ad071a965b46.png)

RobFog:
This should be fixed with #9792?

oktophonie:
Indeed so!

PR #9792, opened 2021-11-19
refactor beam code

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/9023, https://github.com/musescore/MuseScore/issues/9494

Screen Shot 2021-11-18 at 9 22 55 PM

This PR is a complete rewrite of the beaming code to match with @oktophonie's new beam specifications. This PR should receive his approval before merging.

Attached are a PDF of the desired results I used for testing as well as the original MuseScore file that generated those results.

[beams.pdf](https://github.com/musescore/MuseScore/files/7567925/beams.pdf)

[stems.mscz.zip](https://github.com/musescore/MuseScore/files/7567927/stems.mscz.zip)

There are a few features that used to work in the old beaming code but are not yet implemented yet. Expect these in a new PR very soon:

- Respecting all style preferences
- Cross-staff beams
- Beam breaks
- Feathered beams

As far as I know, the missing features do not produce crashes or other unintended side effects. Instead of waiting to submit this PR until those features are done, I'm submitting it to:

- Gather feedback earlier. This code was written from scratch, and while I did my best to eliminate all possible bugs, the lack of solid unit and integration testing made it difficult to guarantee it's free of bugs. So the more people that use it and report bugs, the better the beaming code will be before the alpha launch.
- Limit potential merge conflicts. This is a long-lived branch, and those can make collaboration difficult. As long as I'm working on the beam code in a long-lived branch, others cannot make edits or fix bugs. By submitting this PR before everything is done, it gives others the freedom to fix bugs and to use the new APIs freely (granted there are fewer public APIs).


Nick-Mazuk:
I'll look at the failed test tomorrow

MarcSabatella:
I just want to say, I am really happy to see this happening!

The cross-staff beaming is a pain currently because it can't laid out fully until we calculate system distances, but we can't do that until we do "most" beaming, so there are all sorts of dependencies and obscure bugs and glitches that result.  Similar story for cross-measure beams, and most especially ones that cross systems.  Would be great to be able to finally handle those, but even if this doesn't do the job right away, a major rewrite of the beam code was oong overdue, and hopefully it's a good first step to solving those other thorny problems.

RobFog:
Rebase needed.

Nick-Mazuk:
Pushed an update with a rebase and hopefully fixed the unit tests to match the new specs.

Admittedly, I did remove a few outdated unit tests and I do need to add many more unit tests in future PRs.

Nick-Mazuk:
The vtests are failing, but I'd expect them to fail given that this PR includes substantial visual changes.

oktophonie (approved):

wizofaus:
It would that this has broken the case where the user has manually adjusted the beam position (in which case the y positions are stored in a the property 'fragments', but then in layout2( ) this just gets overwritten? @Nick-Mazuk

src/engraving/libmscore/beam.cpp
```diff
@@ -611,1482 +506,539 @@ void Beam::layout()
...
+    endAnchor.setY(quarterSpace * (isStartDictator ? pointer : dictator));
+
+    qreal beamWidth = score()->styleMM(Sid::beamWidth).val() * mag();
+    _beamDist = beamWidth + 0.25 * spatium();
+
+    add8thSpaceSlant(isStartDictator ? startAnchor : endAnchor, dictator, pointer, beamCount, interval, middleLine, isFlat);
+
+    int fragmentIndex = (_direction == Direction::AUTO || _direction == Direction::DOWN) ? 0 : 1;
+    fragments[frag]->py1[fragmentIndex] = startAnchor.y();
+    fragments[frag]->py2[fragmentIndex] = endAnchor.y();
```
wizofaus, line 1037: I would think we need to check the _userModified flag at this point, and not overwrite y y coordinates for `fragments` where that's true? @Nick-Mazuk

PR #9794, opened 2021-11-19
Fix #9218: Fix transparent .png file export

### LINKED ISSUE #9218: [MU4 Issue] Exported PNGs always have transparent background

HemantAntony:
Before this commit, when exported in .png format, exported .png files always used to have
a transparent background whether transparent option was checked or not.
This commit has solved this bug. Also, the default value of transparent background has been changed from true to false

Resolves: https://github.com/musescore/MuseScore/issues/9218

### DIFF
--- a/src/importexport/imagesexport/internal/pngwriter.cpp
+++ b/src/importexport/imagesexport/internal/pngwriter.cpp
@@ -82,7 +82,7 @@ mu::Ret PngWriter::write(INotationPtr notation, Device& destinationDevice, const
     image.setDotsPerMeterY(std::lrint((CANVAS_DPI * 1000) / Ms::INCH));
 
     const bool TRANSPARENT_BACKGROUND = options.value(OptionKey::TRANSPARENT_BACKGROUND, Val(false)).toBool();
-    image.fill(TRANSPARENT_BACKGROUND ? 0 : Qt::white);
+    image.fill(TRANSPARENT_BACKGROUND ? Qt::transparent : Qt::white);
 
     double scaling = CANVAS_DPI / Ms::DPI;
     Ms::MScore::pixelRatio = 1.0 / scaling;
--- a/src/importexport/imagesexport/internal/imagesexportconfiguration.cpp
+++ b/src/importexport/imagesexport/internal/imagesexportconfiguration.cpp
@@ -35,7 +35,7 @@ static const Settings::Key EXPORT_PNG_USE_TRANSPARENCY_KEY("iex_imagesexport", "
 void ImagesExportConfiguration::init()
 {
     settings()->setDefaultValue(EXPORT_PNG_DPI_RESOLUTION_KEY, Val(Ms::DPI));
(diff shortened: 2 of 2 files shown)


HemantAntony:
Also, should the transparent option be unchecked by default? I would also like to point out that the png export isn't perfect. Look at the right side
TestImage-1

cbjeukendrup:
Personally, I would say, export with white background by default, so yes, the transparent option is unchecked by default. But doubtlessly others think different about that...

And about that image, it seems to crop the image to the topmost/leftmost/rightmost/bottommost elements, instead of keeping the margins and the shape of the page. I think that is definitely not what we want, thanks for pointing that out!

cbjeukendrup:
@igorkorsukov Could you please add the vtests label? The only difference is that the ref images have a transparent background and with this PR they get a white background. That seems desirable to me, isn't it?

Jojo-Schmitz:
IIRC it had been asked for, for the benefit of those using a dark mode when viewing those PNGs copy/pasted from the vtests into GitHub comments?

HemantAntony:
Please don't merge yet. I'm still working on this

src/importexport/imagesexport/internal/pngwriter.cpp
```diff
@@ -69,7 +69,7 @@ mu::Ret PngWriter::write(INotationPtr notation, Device& destinationDevice, const
     RectF pageRect = page->abbox();
 
     if (TRIM_MARGIN_SIZE >= 0) {
-        QMarginsF margins(TRIM_MARGIN_SIZE, TRIM_MARGIN_SIZE, TRIM_MARGIN_SIZE, TRIM_MARGIN_SIZE);
+        QMarginsF margins(TRIM_MARGIN_SIZE + 230, TRIM_MARGIN_SIZE + 420, TRIM_MARGIN_SIZE + 220, TRIM_MARGIN_SIZE + 340);
```
RomanPudashkin, line 72: These hardcoded margins look odd. They may cause some problems in the feature because this code will be used for MuseScore.com. Please, remove them and I'll merge your PR after that. It is better to create a separate issue for the margins problem and solve it in a separate PR

RomanPudashkin (approved):

Issue #9787, opened 2021-11-18
[MU4 Issue] Text on the bottom bar squeezes to two lines


jessjwilliamson:
**Describe the bug**
Just found this issue when I click on a tempo, the text at the bottom left becomes into two lines when it should be just 1.
This is may be related to the recent bottom bar vertical size reduction? https://github.com/musescore/MuseScore/issues/9690

![Screenshot 2021-11-18 144618](https://user-images.githubusercontent.com/71656189/142437766-d8f7e7c7-5745-4a8f-bc38-f6d395a25ce2.png)

**Desktop (please complete the following information):**
 - OS: Windows

jessjwilliamson:
@cbjeukendrup

cbjeukendrup:
It happens because the tempo text contains a newline, which will therefore also appear in the status bar. In the status bar, newlines and other whitespace should be replaced with just a space, which I has implemented in MuseScore 3, but has apparently not been ported. I'll take care of it.

Issue #9742, opened 2021-11-16
[MU4 Task] Implement "getting started" screens


bkunda:
**Task description**

This task concerns the dialog that appears when the user opens MuseScore for the first time (immediately after installation).

Spec document sent directly to @vpereverzev.

cbjeukendrup:
Isn't this duplicating #9468?

vpereverzev:
@cbjeukendrup do you want to tackle this? It's not such a big, but still interesting task. If so, let me know

bkunda:
Duplicate of #9468

cbjeukendrup:
> @cbjeukendrup do you want to tackle this? It's not such a big, but still interesting visual task. If so, let me know

Yes, sure! Indeed a nice task. Would be great if you or Bradley could send me the specification document.

Issue #9468, opened 2021-10-14
[MU4 Task] Implement new "Getting Started" user flow


bkunda:
Design spec in progress for new first time installation user flow

PR #9801, opened 2021-11-20
First Launch Setup Dialog

cbjeukendrup:
Resolves: #9468
Resolves: #9742 (duplicate)

Resolves: #9787 

Note to testers: by default, if you have completed the setup flow, the dialog won't appear anymore of course. To see it again, you can go to DevTools > Settings, and there uncheck `hasCompletedFirstLaunchSetup`, and then you will see the dialog again on the next launch.


src/appshell/view/dockwindow/dockpageview.cpp
```diff
@@ -56,6 +56,11 @@ QString DockPageView::uri() const
     return m_uri;
 }
 
+void DockPageView::open(const QVariantMap& params)
```
RomanPudashkin, line 59: What do you think about renaming this method to setParams?

src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
```diff
@@ -0,0 +1,126 @@
...
+    IInteractive::Result result
+        = interactive()->warning(trc("appshell", "Are you sure you want to cancel?"),
+                                 trc("appshell", "If you choose to cancel, then make "
+                                                 "sure to check out our free playback "
+                                                 "libraries in Home > Audio."),
+                                 buttons,
+                                 int(IInteractive::Button::Cancel),
+                                 IInteractive::Option::WithIcon);
+
+    return result.button() == int(IInteractive::Button::Cancel);
```
RomanPudashkin, line 120: You can use standardButton() in such sittuations

src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "firstlaunchsetupmodel.h"
+
+using namespace mu::appshell;
+
+FirstLaunchSetupModel::FirstLaunchSetupModel(QObject* parent)
+    : QObject(parent)
+{
+    m_pages = {
+        Page { "ThemesPage.qml", "musescore://notation" },
+        Page { "PlaybackPage.qml", "musescore://notation", true },
```
RomanPudashkin, line 31: It would be great to add a comment on what this bool means (or even replace it with a static constant)

RomanPudashkin (approved):

RomanPudashkin:
Looks amazing! Thanks for your work!

src/appshell/view/dockwindow/dockpageview.cpp
cbjeukendrup: Good idea. That way, it's a bit more general, so it is more natural to reuse it for other things in the future.

src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
cbjeukendrup: Thanks, done!

src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
cbjeukendrup: Added a comment.
(See the first commit of #9803)

PR #9803, opened 2021-11-20
Refinements for the Preferences dialog

cbjeukendrup:
Most notably, I added gradients indicating that pages are scrollable.

https://user-images.githubusercontent.com/48658420/142730136-473e5a0d-ce68-412f-9354-6a01755f0ef9.mov

Note: in this video, you see that the scrolling sometimes jumps. This must be a QML bug, because it happens with all scrollable views when scrolling with the mouse or trackpad. Scrolling by using the scrollbar works fine.

### DIFF
--- a/src/appshell/qml/HomePage/HomePage.qml
+++ b/src/appshell/qml/HomePage/HomePage.qml
@@ -40,7 +40,7 @@ DockPage {
     objectName: "Home"
     uri: "musescore://home"
 
-    onOpenRequested: function(params) {
+    onSetParamsRequested: function(params) {
         if (Boolean(params["section"])) {
             setCurrentCentral(params["section"])
 
--- a/src/appshell/qml/Preferences/AdvancedPreferencesPage.qml
+++ b/src/appshell/qml/Preferences/AdvancedPreferencesPage.qml
@@ -20,7 +20,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 import QtQuick 2.15
-import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 
 import MuseScore.Ui 1.0
@@ -31,6 +30,8 @@ import "internal"
 
 PreferencesPage {
     id: root
+
+    contentFillsAvailableHeight: true
 
     Component.onCompleted: {
         preferencesModel.load()
--- a/src/appshell/qml/Preferences/AppearancePreferencesPage.qml
+++ b/src/appshell/qml/Preferences/AppearancePreferencesPage.qml
@@ -30,8 +30,6 @@ import "internal"
 PreferencesPage {
     id: root
 
-    contentHeight: content.height
-
     AppearancePreferencesModel {
         id: appearanceModel
     }
@@ -41,13 +39,11 @@ PreferencesPage {
     }
 
     Column {
-        id: content
-
         width: parent.width
         spacing: root.sectionsSpacing
 
         ThemesSection {
-            width: content.width
+            width: parent.width
 
             themes: appearanceModel.highContrastEnabled ? appearanceModel.highContrastThemes : appearanceModel.generalThemes
             currentThemeCode: appearanceModel.currentThemeCode
@@ -88,7 +84,7 @@ PreferencesPage {
         UiColorsSection {
             id: uiColorsSection
 
-            width: content.width
+            width: parent.width
 
             visible: appearanceModel.highContrastEnabled
 
--- a/src/appshell/qml/Preferences/AppearancePreferencesPage.qml
+++ b/src/appshell/qml/Preferences/AppearancePreferencesPage.qml
@@ -58,15 +58,15 @@ PreferencesPage {
             navigation.section: root.navigationSection
             navigation.order: root.navigationOrderStart + 1
 
-            onThemeChangeRequested: {
+            onThemeChangeRequested: function(newThemeCode) {
                 appearanceModel.currentThemeCode = newThemeCode
             }
 
-            onHighContrastChangeRequested: {
+            onHighContrastChangeRequested: function(enabled) {
                 appearanceModel.highContrastEnabled = enabled
             }
 
-            onAccentColorChangeRequested: {
+            onAccentColorChangeRequested: function(newColorIndex) {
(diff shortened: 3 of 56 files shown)


cbjeukendrup:
I converted it to draft because I was going to remove the Telemetry settings, but I can better do that in a separate PR, because there's a bit more to do about Telemetry.

src/appshell/qml/Preferences/ScrollablePreferencesPage.qml
```diff
@@ -0,0 +1,112 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+import MuseScore.Ui 1.0
+import MuseScore.UiComponents 1.0
+
+PreferencesPage {
```
RomanPudashkin, line 28: It would be great to have only PreferencesPage without ScrollablePreferencesPage. It looks like PreferencePage can decide whether it can be scrolled or not (and show/hide gradients after that)

cbjeukendrup: In this context, "Scrollable" does not pre se mean that the content height is taller than the available. If that was the case, it would indeed be obvious that one PreferencesPage type is enough.

On "Scrollable" pages, the content height is determined by the contents. In practice, the page may or may not be scrollable, depending on whether the content height is taller than the available height. This applies to most pages, including the "General" and "Appearance" pages.

On "Non-Scrollable" pages, the content height is determined by the available height. So, the view will never be scrollable but will be stretched to fill the available height. This applies to the "MIDI device mappings" page, the "Shortcuts" page and the "Advanced" page.

Of course, it is _our choice_ whether the contents of a page should take its intrinsic size, or that it fills up the available height. So this can't be determined by the PreferencesPage.

However, I simplified it now, so that there is only one PreferencesPage component, with a property to control the behaviour. This seems like a more elegant solution.

cbjeukendrup:
Hm, this got automatically closed now, which was not planned! It seems I can't reopen it myself...

Eism:
@cbjeukendrup 
Please explain what problems did you solve with this PR? :)

Jojo-Schmitz:
Why do some vtests fail?

src/appshell/qml/Preferences/ScrollablePreferencesPage.qml
Eism: If we decide to add more items to the page, then we will need to control scrolling with this flag. Previously, this was done automatically(by contentHeight: content.height).

So, I agree with Roman
> It looks like PreferencePage can decide whether it can be scrolled or not (and show/hide gradients after that)

cbjeukendrup: Again, that is not the point of the `contentDeterminesContentHeight` flag. I think that we are agreeing all three, but that there is a misunderstanding.

If you add more content to a normal page, it will still automatically become scrollable if the size of the window is too small. So, in fact, nothing changes regarding that. Before, we would have to write `contentHeight: content.height`, but that is not even necessary anymore: it goes automatically.

But on some pages, we do _not_ want that the content height is determined based on the implicit height of the content, but we want that the content takes all available height, not more, not less. For example, the MIDI Device mappings page. Before, we would write `contentHeight: height` on those pages (where `height` is the height of the Flickable). Now, we can write `contentDeterminesContentHeight: false` instead, which is technically more characters, but its meaning is clearer.

So, in code, the idea is the following: 

```diff
 // GeneralPreferencesPage.qml and many others
 // content height determined by content itself
 PreferencesPage {
-    contentHeight: content.height
     
     Column {
         id: content
         ...
     }
 }
 
 // MidiDeviceMappingsPreferencesPage.qml, ShortcutsPreferencesPage.qml, AdvancedPreferencesPage.qml
 // content height follows available height
 PreferencesPage {
-    contentHeight: height // `height` is the height of the preferences page
+    contentDeterminesContentHeight: false
     
     PageContent {
         anchors.fill: parent
     }
 }
```

I hope this explanation is more understandable...

Maybe it would be clearer if I replace the `contentDeterminesContentHeight` property with a property called `contentFillsAvailableHeight`. So instead of `contentDeterminesContentHeight: false`, we would write `contentFillsAvailableHeight: true` on those exception pages.

cbjeukendrup: Indeed, I changed it to `contentFillsAvailableHeight: true`. I hope it is clear now.

cbjeukendrup:
> @cbjeukendrup
> Please explain what problems did you solve with this PR? :)

Per commit:

1. Some code review fixes for First Launch Setup Dialog (I created this PR a while ago, just after the First LaunchSetupDialog PR was merged, so that's why those fixes are in this PR)

2. In the Preferences dialog, when you press Tab the first time to activate keyboard navigation, the focus will always jump to the selected item in the pages menu, even when that item has been selected earlier by using the mouse.

3. I cleaned up the spacing, sizes and placement of all controls. I made columns the same width, made the spacing consistent between all different components, etc.. Less magic numbers, more controlled properties. In some cases, even Figma was not completely consistent, so the values might deviate a tiny bit from Figma, but I ensured everything looks beautiful.

4. I fixed/cleaned the navigation focus border for ValueList. I also removed some undesired padding that we added to prevent problems with the focus border, but those problems have been solved in a better way now, so we don't need the padding hacks anymore.

5. I added gradients to indicate that a page is scrollable. Before, the padding was added _outside_ the pages. Now, it will be added _inside_ the Flickable, so that the content will not be clipped at 30 px from the top and bottom, but instead it will extend under the gradient. This also fixes problems with focus borders being cut off at the borders of the page.

6. And I made some more improvements, mainly to the vertical spacing inside sections. It is very consistent now, and I love consistency like this :)
  (I didn't want to add this to the fourth commit, to keep the commits a bit concise...)

---

> Why do some vtests fail?

Apparently I had to rebase :)
I think that I fixed that in PR #10057, so that the tests won't fail if you didn't touch them but have not rebased yet.

Eism (approved):

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Issue #9705, opened 2021-11-10
[MU4 Task] Accessibility. Master Palette


Eism:
**Task description**
Actualize accessibility for widget dialogs

Issue #9704, opened 2021-11-10
[MU4 Task] Accessibility. Edit Drumset, Key Edit, Symbol, Time, Time Signature dialogs


Eism:
**Task description**
Actualize accessibility for widget dialogs

PR #9824, opened 2021-11-22
[MU4] Accessibility. Master palette

Eism:
Resolves: #9705 #9704

### DIFF
--- a/src/framework/accessibility/iaccessibleinterfaceregister.h
+++ b/src/framework/accessibility/iaccessibleinterfaceregister.h
@@ -19,21 +19,26 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
+#ifndef MU_UI_IACCESSIBLEINTERFACEREGISTER_H
+#define MU_UI_IACCESSIBLEINTERFACEREGISTER_H
 
-#ifndef MU_ACCESSIBILITY_ACCESSIBILITYMODULE_H
-#define MU_ACCESSIBILITY_ACCESSIBILITYMODULE_H
+#include "modularity/imoduleexport.h"
 
-#include "framework/global/modularity/imodulesetup.h"
+#include <QAccessibleInterface>
 
 namespace mu::accessibility {
-class AccessibilityModule : public modularity::IModuleSetup
+class IAccessibleInterfaceRegister : MODULE_EXPORT_INTERFACE
 {
-public:
-    std::string moduleName() const override;
+    INTERFACE_ID(IAccessibleInterfaceRegister)
 
-    void registerExports() override;
-    void resolveImports() override;
+public:
+    virtual ~IAccessibleInterfaceRegister() = default;
+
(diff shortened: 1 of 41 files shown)


src/framework/accessibility/iaccessibleinterfaceregister.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace mu::accessibility {
+class IAccessibleInterfaceRegister : MODULE_EXPORT_INTERFACE
+{
+    INTERFACE_ID(IAccessibleInterfaceRegister)
+
+public:
+    virtual ~IAccessibleInterfaceRegister() = default;
+
+    using InterfaceGetter = std::function<QAccessibleInterface* (QObject*)>;
```
igorkorsukov, line 37: Do we really need this interface?
I believe that there is no need for Qt interfaces to appear in our interfaces.
It is better if we make an implimentation from IAccessible who will be a proxy for QAccessibleInterface

Or at least you need to rename IAccessibleInterfaceRegister to IQAccessibleInterfaceRegister

src/framework/accessibility/internal/accessibilitycontroller.cpp
```diff
@@ -47,20 +47,21 @@ AccessibilityController::~AccessibilityController()
...
+    return static_cast<QAccessibleInterface*>(new AccessibleItemInterface(s_rootObject));
+}
+
 static QAccessibleInterface* muAccessibleFactory(const QString& classname, QObject* object)
 {
-#ifdef Q_OS_MAC
-    if (classname == QLatin1String("QQuickWindow")) {
-        return static_cast<QAccessibleInterface*>(new AccessibleItemInterface(s_rootObject));
+    auto accessibleInterfaceRegister = mu::modularity::ioc()->resolve<IAccessibleInterfaceRegister>("accessibility");
+    if (!accessibleInterfaceRegister) {
```
igorkorsukov, line 58: This function is called very often, and `ioc()->resolve` is quite an overhead call, it is better to do it once

src/framework/accessibility/iaccessibleinterfaceregister.h
```diff
@@ -0,0 +1,44 @@
...
+
+    virtual void registerInterfaceGetter(const std::string& className, const InterfaceGetter& interfaceGetter) = 0;
```
igorkorsukov, line 39: if it uses a Qt type that cannot be gotten rid of, then it makes no sense to use `std::string`

src/framework/accessibility/iaccessibleinterfaceregister.h
Eism: thanks, renamed

src/framework/accessibility/internal/accessibilitycontroller.cpp
Eism: fixed

src/framework/accessibility/iaccessibleinterfaceregister.h
Eism: fixed

igorkorsukov (approved):

PR #9828, opened 2021-11-22
Adapt code style for Uncrustify 0.74.0

cbjeukendrup:
Uncrustify 0.74.0 is out, and it has somewhat changed behaviour, so this PR adapts our code to the new behaviour. 

As with 0.73.0, not all changes are desired. For five of the 10 changed files it becomes worse than what it was, for two other files the changes are questionable. Anyway, we have little choice... 

Like with 0.73.0, it will probably take some time before the new version also gets used on GitHub CI. Therefore, I'm just creating this PR now, so that we can merge it right at the moment when it's needed.

### DIFF
--- a/src/cloud/internal/cloudservice.h
+++ b/src/cloud/internal/cloudservice.h
@@ -92,7 +92,7 @@ private:
     RequestStatus downloadUserInfo();
     RequestStatus doUploadScore(io::Device& scoreSourceDevice, const QString& title, const QUrl& sourceUrl = QUrl());
 
-    using RequestCallback = std::function<RequestStatus()>;
+    using RequestCallback = std::function<RequestStatus ()>;
     void executeRequest(const RequestCallback& requestCallback);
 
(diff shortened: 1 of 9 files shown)


Jojo-Schmitz:
I guess #9952 makes this PR here obsolete?

RobFog:
Why? 0.73 is currently used. This pull request is for 0.74.

Jojo-Schmitz:
Because I think this PR is based on the assumption that GitHub CI for macOS may switch to 0.74. We've now avoided getting caught by that by switching to Linx and use out own copy of Uncrustify

cbjeukendrup:
Indeed, this is not necessary anymore now...

igorkorsukov:
I think we will come back to this PR in the future.
At the moment, we decided that it is better to freeze the Uncrustify version, and we will update when we want, we are ready to do it, and not at the most unexpected moment :)

Jojo-Schmitz:
In that canse it might be feasible to keep this PR open (and in a Draft state)?

Issue #9157, opened 2021-09-16
[VST - MIXER] VST Windows should be prevented from 'going behind' the main window


Tantacrul:
**Describe the bug**
In many other DAW's, VST windows are forced to front of the main interface. If this isn't done, then they very easily get lost and it is quite painful for the user to relocate them.

**Expected behaviour**
All VST windows stay 'in front' of the main MS4 window.

DmitryArefiev:
Hi @Tantacrul , 
we talked with @vpereverzev and @RomanPudashkin about this.. and turned out it's hard to implement this feature.

Also that behavior with locked VST window usually appears only in DAW and not in notation software (e.g. Sibelius)

>If this isn't done, then they very easily get lost and it is quite painful for the user to relocate them
Now Roman did a fix by using button on the mixer and user can easily return VST window in front just but clicking on VSTs name in mixer.

Another argument to use the mixer button only is that the VST plugin can have a wide window and overlap score view (please see my video from comment above with Vital plugin)

So, if the initial goal/behavior is important I can leave the bug in Open state

Tantacrul:
Hmmm. It is quite a problem, I feel. I'm aware of how Sibelius performs here but it's quite bad. I've already had the experience of losing my VST windows a few times. I also don't think it matters that the windows are large and can overlap the score because that's just the nature of VST's. As you've said, most DAWs do it - and they do it for a reason.

How difficult are we talking? I'm reluctant for this issue not to get fixed.

cbjeukendrup:
I've looked into this a little bit, and I'm really stunned how difficult this is, while it seems a quite straightforward and reasonable task.
 
With one extra line of code, you can make the window stay on top of the whole OS, but I expect that is not what we want, right?

Keeping a window in front of just the _application_ seems impossible using Qt; there is just no API for it. So to achieve this, we would need to use native, OS dependent APIs.

On macOS, it is relatively easily doable, once you've got hold of the underlying `NSWindow*`. A complication though, is that you need Objective-C++ to be able to work with macOS APIs, so we would need to add an extra Objective-C++ source file to get just this simple thing done. In that extra file should be something like:
```objective-c++
    NSView* nsView = (__bridge NSView*)reinterpret_cast<void*>(window->winId());
    NSWindow* nsWindow = [nsView window];
    if (nsWindow) {
        [nsWindow setLevel:NSFloatingWindowLevel];
    }
```

[...]

PR #9830, opened 2021-11-23
Make VST windows stay on top when the application is active

cbjeukendrup:
Resolves: #9157

Also introduces a common base class for VstFxEditorView and VstiEditorView to reduce code duplication.

### DIFF
--- a/build/package_mac
+++ b/build/package_mac
@@ -2,7 +2,7 @@
 
 APPNAME=mscore
 LONG_NAME="MuseScore"
-LONGER_NAME="MuseScore 3"
+LONGER_NAME="MuseScore 4"
 VERSION=0
 
 while [[ "$#" -gt 0 ]]; do
@@ -15,7 +15,7 @@ while [[ "$#" -gt 0 ]]; do
     shift
 done
 
-if [ $VERSION -eq 0 ]; then 
+if [ "$VERSION" = 0 ]; then
     SCRIPT_DIR=$(cd "$(dirname ${BASH_SOURCE[0]})" && pwd)
     VERSION=$(cmake -P "$SCRIPT_DIR/../config.cmake" | sed -n -e "s/^.*VERSION  *//p")
 fi    
--- a/src/framework/vst/CMakeLists.txt
+++ b/src/framework/vst/CMakeLists.txt
@@ -63,6 +63,8 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/devtools/vstpluginlistmodelexample.cpp
     ${CMAKE_CURRENT_LIST_DIR}/devtools/vstpluginlistmodelexample.h
 
+    ${CMAKE_CURRENT_LIST_DIR}/view/abstractvsteditorview.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/view/abstractvsteditorview.h
     ${CMAKE_CURRENT_LIST_DIR}/view/vstieditorview.cpp
     ${CMAKE_CURRENT_LIST_DIR}/view/vstieditorview.h
     ${CMAKE_CURRENT_LIST_DIR}/view/vstfxeditorview.cpp
--- a/src/framework/vst/devtools/vstpluginlistmodelexample.cpp
+++ b/src/framework/vst/devtools/vstpluginlistmodelexample.cpp
@@ -88,7 +88,7 @@ void VstPluginListModelExample::load()
 
 void VstPluginListModelExample::showSynthPluginEditor()
 {
-    QString uri = QString("musescore://vsti/editor?sync=false&trackId=%1&resourceId=%2")
+    QString uri = QString("musescore://vsti/editor?sync=false&floating=true&trackId=%1&resourceId=%2")
                   .arg(m_currentTrackId)
                   .arg(m_currentSynthResource);
 
@@ -97,7 +97,7 @@ void VstPluginListModelExample::showSynthPluginEditor()
 
 void VstPluginListModelExample::showFxPluginEditor()
 {
-    QString uri = QString("musescore://vstfx/editor?sync=false&trackId=%1&resourceId=%2")
+    QString uri = QString("musescore://vstfx/editor?sync=false&floating=true&trackId=%1&resourceId=%2")
                   .arg(m_currentTrackId)
                   .arg(m_currentFxResource);
 
--- a/src/framework/vst/view/abstractvsteditorview.cpp
+++ b/src/framework/vst/view/abstractvsteditorview.cpp
@@ -1,0 +1,182 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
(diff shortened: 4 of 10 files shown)


vpereverzev:
@cbjeukendrup

- I suggest to just use a combination of Qt::WindowStaysOnTopHint and Qt::NativeWindow. I've tested this at the weekend and it works fine on Windows and MAC. There are some problems on Linux, but we don't support VST on Linux anyway
- I suggest to not inherit VstFx dialog from VSTi one. It's not really OK from semantic POV. If you want to reuse some code, it's better to introduce a base class for both of them, like "VstEditorDialog" or something like that. However, the amount of duplicating code wouldn't be huge, since you literally need to adjust 2 lines overall

cbjeukendrup:
@vpereverzev The problem of just using Qt::WindowStaysOnTopHint, is that the VST window will stay on top of _all_ other applications in the OS, and I believe this is not desired. So therefore I remove the Qt::WindowStaysOnTopHint when the application is not frontmost, so that VST windows are allowed to go behind other applications, but not behind the main window. 

But of course, if we do want the VST windows on top of all other applications, even if MuseScore is in the background, then it would be obvious to use just Qt::WindowStaysOnTopHint. If I had known that this was the desired behaviour, I would have done that weeks ago.

About the code duplication, it was not primarily because I had to write code in two places, but more because I saw some small discrepancies in the existing code that were probably not right. 
I originally created a common base class "AbstractVstEditorDialog", but then switched to this approach because the VSTi dialog would have almost everything in common with the AbstractVstEditorDialog. But I absolutely agree that a common base class makes more sense, so returned that approach.

vpereverzev:
> @vpereverzev The problem of just using Qt::WindowStaysOnTopHint, is that the VST window will stay on top of _all_ other applications in the OS, and I believe this is not desired. So therefore I remove the Qt::WindowStaysOnTopHint when the application is not frontmost, so that VST windows are allowed to go behind other applications, but not behind the main window.
> 
> But of course, if we do want the VST windows on top of all other applications, even if MuseScore is in the background, then it would be obvious to use just Qt::WindowStaysOnTopHint. If I had known that this was the desired behaviour, I would have done that weeks ago.
> 
> About the code duplication, it was not primarily because I had to write code in two places, but more because I saw some small discrepancies in the existing code that were probably not right. I originally created a common base class "AbstractVstEditorDialog", but then switched to this approach because the VSTi dialog would have almost everything in common with the AbstractVstEditorDialog. But I absolutely agree that a common base class makes more sense, so returned that approach.

BTW what were the problems with Qt::Tool attribute?

cbjeukendrup:
Yeah, I mentioned it here: https://github.com/musescore/MuseScore/issues/9157#issuecomment-943580456

vpereverzev:
> Yeah, I mentioned it here: [#9157 (comment)](https://github.com/musescore/MuseScore/issues/9157#issuecomment-943580456)

hmm, it's a lesser evil I'd say

Let's use Qt::Tool | Qt::NativeWindow for now and then ask @DmitryArefiev to test this on Win and Mac.

Also, I forgot to mention, it probably might make sense to add "&floating=true" argument in URI for this dialogs

P.S. I suggest you to install some VST plugins for your machine. There are lots of free VST plugins for MAC, however some of them are not implemented "correctly". Among others, I usually prefer to use free 'u-he" plugins for testing purposes

cbjeukendrup:
@vpereverzev I tried Qt::Tool on Windows, and it does not work. It does not keep the window in front, but only gives a strange title bar. 

My "hacky" solution seems to work well on Windows, like it does also on macOS. 

I also did some more research about what Qt::Tool actually does on macOS.

It appears that it works actually quite similar to my "hacky" solution. Normally, a Qt::Tool window would also float always on top of all applications, but it is hidden when the app is inactive, so you won't notice that in practice. The native macOS window has the an option called `setHidesOnDeactivate`, to enable or disable this behaviour. The only difference between my solution and Qt/macOS's solution, is that macOS hides the window altogether, but my solution only removes the "floating" window flag. 

So I get the feeling that my solution wasn't that evil at all. 

Another downside of using Qt::Tool is that such windows don't appear in macOS's "Mission control" view, so you can't find them in the overview of all windows. After playing a bit with the different solutions, I must conclude that I prefer the UX of my "hacky" solution. 

So, could you please reconsider it?

vpereverzev:
> @vpereverzev I tried Qt::Tool on Windows, and it does not work. It does not keep the window in front, but only gives a strange title bar.
> 
> My "hacky" solution seems to work well on Windows, like it does also on macOS.
> 
> I also did some more research about what Qt::Tool actually does on macOS.
> 
> It appears that it works actually quite similar to my "hacky" solution. Normally, a Qt::Tool window would also float always on top of all applications, but it is hidden when the app is inactive, so you won't notice that in practice. The native macOS window has the an option called `setHidesOnDeactivate`, to enable or disable this behaviour. The only difference between my solution and Qt/macOS's solution, is that macOS hides the window altogether, but my solution only removes the "floating" window flag.
> 
> So I get the feeling that my solution wasn't that evil at all.
> 
> Another downside of using Qt::Tool is that such windows don't appear in macOS's "Mission control" view, so you can't find them in the overview of all windows. After playing a bit with the different solutions, I must conclude that I prefer the UX of my "hacky" solution.
> 
> So, could you please reconsider it?

OK, got you. Let's stick with this approach then

vpereverzev (approved):

jessjwilliamson:
There is a problem with clicking inside the VST :( 

https://user-images.githubusercontent.com/71656189/143584340-28d418f7-da87-4a47-ad07-1ab3646f230a.mp4

cbjeukendrup:
Hm, I notice that some things are not going well on Windows. I'll investigate. The conclusion may be that it is the best to just hide the VST window when none of the MuseScore windows is active, to prevent bugs. @jessjwilliamson Would you be okay with that?

jessjwilliamson:
> Hm, I notice that some things are not going well on Windows. I'll investigate. The conclusion may be that it is the best to just hide the VST window when none of the MuseScore windows is active, to prevent bugs. @jessjwilliamson Would you be okay with that?

The point was the user to use the VST while able to edit their score so it's always on display. It's not too much needed when changing to another app, but would expect it to still be there when switching back to Musescore... Let's try that for now and see how it goes. I also noticed how some of the VST FX are experiencing bad reactions causing the whole app to freeze and unrespond, forcing it to close via task manager. :c

jessjwilliamson:
@cbjeukendrup Just tested this on Mac, no issues!

cbjeukendrup:
It turns out that on Windows, the app becomes "inactive" for a fraction of a second each time when you click inside the plugin view. Therefore, that makes the plugin window become not-on-top for a moment and then on-top again, which requires sneakily closing and opening it twice (you can't simply switch a window from on-top and not-on-top or vice versa, but it needs to be closed to do that). 

So my solution of hiding the window entirely when the app is inactive won't solve anything. And I'm afraid we can't do much to suppress the redundant inactivate/activate events that apparently occur when clicking inside the window. We'd need to find another solution. Maybe the only option is to use native Windows APIs... anyway, I'll continue investigating this as soon as I can.

Issue #9447, opened 2021-10-12
[Crash] Removing an accidental from an Ambitus causes MS4 to crash.


Tantacrul:
**Describe the bug**
See video:

https://user-images.githubusercontent.com/21022311/136996415-18a575de-8ae1-48f0-9148-02bf2aeac55a.mov

I'm making this a P0 on the assumption that it is part of a wider issue with crashing. If was was specific to the Ambitus, I'd probably mark it as a P1 or even P2.

igorkorsukov:
@DmitryArefiev Please check, PR https://github.com/musescore/MuseScore/pull/9831
I fixed the crash (triggering of the assertion in the development build). But it looks like the ambitus is not displayed correctly, if so, then I propose to have a separate issue with a different priority

Jojo-Schmitz:
Apparently affects 3.x too

DmitryArefiev:
Tested #9447 on Win10, Mac12, LinuxMint 20.2 - FIXED (should be in the next master)


Found another crash when using ambitus #9866 (not caused by the current fix)

PR #9831, opened 2021-11-23
Fixed trigger assert on remove acc from ambitus

igorkorsukov:


RomanPudashkin (approved):

src/engraving/libmscore/ambitus.cpp
```diff
@@ -679,6 +679,16 @@ Ambitus::Ranges Ambitus::estimateRanges() const
     return result;
 }
 
+void Ambitus::remove(EngravingItem* e)
+{
+    if (e->type() == ElementType::ACCIDENTAL) {
+        //! NOTE Do nothing (removing _topAccid or _bottomAccid)
```
Jojo-Schmitz, line 685: Why doing nothing? Via Inspector the Ambitus can get changed to a non-accidental note, why not via deleting the accidental here?

igorkorsukov: Because they are deleted in
```
Ambitus::~Ambitus()
{
    delete _topAccid;
    delete _bottomAccid;
}
```

They hide somehow differently (I did not investigate in detail)

Jojo-Schmitz: But, using you patch, they don't disapear, so don't delete at all, unless you delete the entire Ambitus (only that would call the destructor)

DmitryArefiev:
Tested #9447 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

@igorkorsukov Found another crash when using ambitus #9866 (not causes by the current PR)

src/engraving/libmscore/ambitus.cpp
igorkorsukov: I had no plans to change the implementation and logic of `Ambitus`.
I just avoided triggering the assertion, that is, I didn’t make worse than it was, I did it a little better.
Do you think it is necessary to reconsider the implementation of `Ambitus`?

Jojo-Schmitz: Well, doing nothing here is certainly better than a crash

PR #9850, opened 2021-11-23
Fix computation of smallness for tuplets

asattely:
The problem:
image

The issue was that a tuplet is set to small if all of its elements are small chordrests. More accurately, it is set to regular size if any of its elements are normal-sized chordrests. This fails for tuplets whose elements are only other tuplets, and the containing tuplet remains small.

Since tuplets are laid out from contained to containing, the fix adds an `isSmall` value which is set in the layout function, which layout checks in a way similar to how it checks the smallness of chordrests:
image
image

### DIFF
--- a/src/engraving/libmscore/tuplet.cpp
+++ b/src/engraving/libmscore/tuplet.cpp
@@ -206,14 +206,14 @@ void Tuplet::layout()
             _number->setXmlText(QString("%1:%2").arg(_ratio.numerator()).arg(_ratio.denominator()));
         }
 
-        bool isSmall = true;
+        _isSmall = true;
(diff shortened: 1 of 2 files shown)


asattely:
Aaaaaand the build fails.

cbjeukendrup:
Root cause seems that it fails to install wget, and therefore it fails to download the dependencies. Probably a temporary issue.

asattely:
Yeah, likely. None of the changes I've made should affect any of that so I'll just push an empty amend commit in a little bit and see if it fixes itself

oktophonie (approved):

igorkorsukov (approved):

Issue #9779, opened 2021-11-18
[Inspector] Unable to change stem direction in Properties after applying flip direction with X from keyboard to the group of notes


DmitryArefiev:
Unable to change stem direction in Properties after applying flip direction command with X from keyboard

https://user-images.githubusercontent.com/90187801/142410623-cb10c5eb-3e5b-4505-8629-d52da1935d46.mp4

**Desktop:**
Win10/Mac12/LinuxMint20

**Additional context**
Occurs in **MU3** as well

Jojo-Schmitz:
Isn't that a duplicate of #9650 and such fixed along with #9774?

DmitryArefiev:
> 
> 
> Isn't that a (more or less) duplicate of #9650 and such fixed along with #9774?

Don't think so. The steps are different and not fixed with #9774.
The current issue is about how flip command (from keyboard) locks stem direction from editing from Properties even if applied only to the first note of group

Jojo-Schmitz:
OK, I see

LovCAPONE:
Since this issue isn't assigned to anyone I'll fix it and make a PR.

DmitryArefiev:
Tested on latest master build on Win10, Mac12, LinuxMint 20 - FIXED

Thanks!

PR #9851, opened 2021-11-23
Fix beam stem direction flip (MU3)

LovCAPONE:
Update chord's set stem direction method to set the beam's direction
to Direction::AUTO if applicable.

Update the beam's set direction method to update its notes' stem
direction.

Resolves: *[(direct link to the issue)](https://github.com/musescore/MuseScore/issues/9779)*

When a beam's direction is set with the "x" keyboard key, the
"_direction" member of the beam class is no longer equal to
"Direction::AUTO". So, in the beam's layout code the final beam
direction is not taken from the first note in the note group
(of the beam) but from the "_direction" member.

The problem is that if the user tries to set the beam's direction
with a note stem direction (in the property panel), it will be
ignored in the beam's layout code.


LovCAPONE:
This PR is ready for review.

@Jojo-Schmitz The issue was that in the Beam::setBeamDirection method, there are 2 "if" blocks and the second block has to be nested. If it's not than there is a stack overflow crash.

Nick-Mazuk:
FYI, I'm in the middle of refactoring most of the beams code, and this PR conflicts with most of those changes.

https://github.com/musescore/MuseScore/pull/9792

LovCAPONE:
> FYI, I'm in the middle of refactoring most of the beams code, and this PR conflicts with most of those changes.
> 
> #9792

Ok thanks for the notification, I'll fix the PR.

Jojo-Schmitz:
Thanks, integrated into PR #9000

LovCAPONE:
> Thanks, integrated into PR #9000

You're welcome!

RomanPudashkin:
3.x is closed for any changes

Issue #9094, opened 2021-09-09
[MU4 Issue] Performance: MU4 is very slow or compleatly freezes when adding multiple articulations in a big score 


DmitryArefiev:
**Describe the bug**
MU4 is very slow or compleatly freezes when adding multiple articulations in a big score 

**To Reproduce**
1. Open [Big_Score.zip](https://github.com/musescore/MuseScore/files/7138238/Big_Score.zip)
2. Select top staff in the first system 
3. Add an accent (see videos below to compare MU4 and MU3)

**Expected behavior**
Performance should be the same as in MU3

**Desktop (please complete the following information):**

**MU4**
https://user-images.githubusercontent.com/90187801/132726179-2c3a0995-e2d7-454b-82ca-3c1020f3c9d4.mp4
**MU3**
https://user-images.githubusercontent.com/90187801/132726205-a3016b2a-d609-4245-a980-a8d60d1dfe6e.mp4

DmitryArefiev:
Can be related to #8440

igorkorsukov:
@DmitryArefiev Please check, PR https://github.com/musescore/MuseScore/pull/9854

RobFog:
compleatly → completely

DmitryArefiev:
Tested on master build (Win10, Mac11, LinuxMint 20) - FIXED

PR #9854, opened 2021-11-24
Added Profiler View and Removed update parts name on drop elements

igorkorsukov:
For view Profile data - go to Diagnostic->System->Show profiler...

image

About removing update parts name on drop elements
There are several problems:
* The parts names were updated every time any element was dropped from the Palette to Score
* Each time the names of all parts are updated ( only need for which we added an instrument change)
* Updating the names of parts, if the new name differs from the current one, is carried out slowly, respectively, when updating the names of all parts, this can take more than 30 seconds
* It is not correct to update the name of the part for a drop element from the Palette, because we can also edit an already added "instrument change". Accordingly, in this case, we should also update the name of the part. Accordingly, part name updates when adding a "instrument change" and editing a "instrument change" should be consistent.

In general, the зart name update needs to be re-implemented properly. @DmitryArefiev FYI

### DIFF
--- a/src/diagnostics/view/system/profilerviewmodel.cpp
+++ b/src/diagnostics/view/system/profilerviewmodel.cpp
@@ -1,0 +1,123 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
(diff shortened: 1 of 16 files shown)


src/diagnostics/view/system/profilerviewmodel.cpp
```diff
@@ -0,0 +1,123 @@
...
+        item.group = group;
+        item.data = data;
+
+        m_allList.append(item);
+    }
+
+    find(m_searchText);
+}
+
+void ProfilerViewModel::find(const QString& str)
```
RomanPudashkin, line 94: The name of this function is confusing. I would suggest renaming it to load (or something similar). You can also use SortFilterProxyModel in such situations

src/diagnostics/view/system/profilerviewmodel.cpp
```diff
@@ -0,0 +1,123 @@
...
+    QStringList list = str.split("\n");
+    foreach (const QString& data, list) {
+        Item item;
+        item.group = group;
+        item.data = data;
+
+        m_allList.append(item);
+    }
+
+    group = "Other thread";
```
RomanPudashkin, line 80: This block of code duplicates the block above. I think you can add a lambda with this code and reuse it twice

RomanPudashkin (approved):

src/diagnostics/view/system/profilerviewmodel.cpp
igorkorsukov: This method is find (filtering), not load data (data already loaded to `m_allList`)

I don't like QSortFilterProxyModel, if I can access and can edit model source.

DmitryArefiev:
Tested #9094 8440 on master build (Win10, Mac11, LinuxMint 20) - FIXED

PR #9863, opened 2021-11-24
Fix #9553 "Open Recent" not disabled

### LINKED ISSUE #9553: [MU4 Issue] File->Open Recent not disabled even if there is no recent files.
... and does nothing when clicked

see https://youtu.be/0Y0X10yiOYg?t=666

HemantAntony:
Resolves: https://github.com/musescore/MuseScore/issues/9553

This commit solved the bug that File->"Open Recent" is not disabled even if
there is no recent files. Now, it decides whether to enable "Open
Recent" on the basis of whether the recent scores list is empty.

Also, I have removed the comment: need implement of export since I think it has been implemented. Please tell me otherwise.

### DIFF
--- a/src/notation/internal/notationuiactions.cpp
+++ b/src/notation/internal/notationuiactions.cpp
@@ -184,7 +184,8 @@ const UiActionList NotationUiActions::m_actions = {
              ),
     UiAction("copy",
              mu::context::UiCtxNotationFocused,
-             QT_TRANSLATE_NOOP("action", "Copy")
+             QT_TRANSLATE_NOOP("action", "Copy"),
+             IconCode::Code::COPY
              ),
     UiAction("paste",
              mu::context::UiCtxNotationFocused,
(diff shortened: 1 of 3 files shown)


RobFog:
Does the second commit belong in this pull request?

cbjeukendrup:
Because both commits are about improvements to the actions and  menus, I think it makes sense to group them. (Otherwise, you would end up with a lot of small PRs...)

src/notation/internal/notationuiactions.cpp
```diff
@@ -213,7 +214,7 @@ const UiActionList NotationUiActions::m_actions = {
     UiAction("toggle-visible",
              mu::context::UiCtxNotationOpened,
              QT_TRANSLATE_NOOP("action", "Toggle visibility of elements"),
-             QT_TRANSLATE_NOOP("action", "Toggle visibility of elements")
+             QT_TRANSLATE_NOOP("action", "Toggle visibility of elements"),
```
cbjeukendrup, line 217: You've an extra comma here, that's why the test builds are failing! Please amend your commit and force-push, then the checks will be run again and everything will be all right.

For the rest, it looks great!

Jojo-Schmitz, line 217: That comma is not needed. Actually the whole line is not needed (and then the comma on the line abobe is not needed anomre either)

src/notation/internal/notationuiactions.cpp
```diff
@@ -184,7 +184,8 @@ const UiActionList NotationUiActions::m_actions = {
              ),
     UiAction("copy",
              mu::context::UiCtxNotationFocused,
-             QT_TRANSLATE_NOOP("action", "Copy")
+             QT_TRANSLATE_NOOP("action", "Copy"),
```
Jojo-Schmitz, line 187: Here I believe you'd need to duplicate that line

src/notation/internal/notationuiactions.cpp
Jojo-Schmitz: Sure this comma is the culprit?

cbjeukendrup: Yes I am. Strange discrepancy in C++: in an enum or list, a comma before a `}` is allowed, but before a `)` it gives an error.

Jojo-Schmitz: OK, so that's 2 build problems then, the other one needs to get fixed too

HemantAntony:
I knew that it would bring an error. I saw it in the last second, but I was too tired to fix it. Will do that soon!

RomanPudashkin (approved):

Issue #9869, opened 2021-11-25
[MU4 Issue] Popup background colour does not fully update when switching between dark and light modes


bkunda:
**Describe the bug**
When switching between dark and light modes, the fill colour of the triangular pointer in the upper border of certain popups (e.g. "Appearance" in **Properties** and "More" in **Palettes**) does not update with the GUI colour scheme.

![popup-colour-discrepancy](https://user-images.githubusercontent.com/86290556/143435006-1898d0de-725b-4305-b4ff-01bfe06c3d47.gif)

**Expected behavior**
Fill colour of this UI element should match background colour of popup, relative to the chosen UI mode (dark/light).

**Desktop (please complete the following information):**
 - MacOS 11.6
 - Windows 11

PR #9872, opened 2021-11-25
Two more UI finesse fixes

cbjeukendrup:
Resolves: #9869 
Resolves: #10045 

- Fix updating color of arrow in StyledPopupView
- Fix height of NoteInputBar (again)

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -182,6 +182,8 @@ DockPage {
                 root.toolBarRightDropDestination
             ]
 
+            thickness: orientation === Qt.Horizontal ? 40 : 76
+
             NoteInputBar {
                 orientation: noteInputBar.orientation
                 floating: noteInputBar.floating
--- a/src/appshell/qml/dockwindow/DockToolBar.qml
+++ b/src/appshell/qml/dockwindow/DockToolBar.qml
@@ -20,7 +20,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 import QtQuick 2.15
-import QtQuick.Layouts 1.15
 
 import MuseScore.Ui 1.0
 import MuseScore.UiComponents 1.0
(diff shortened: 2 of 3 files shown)


src/appshell/qml/NotationPage/NotationPage.qml
```diff
@@ -182,6 +182,8 @@ DockPage {
                 root.toolBarRightDropDestination
             ]
 
+            thickness: orientation === Qt.Horizontal ? 40 : 76
```
RomanPudashkin, line 185: Could you please explain what is the purpose of this change?

cbjeukendrup: By design, the Note Input Bar should be 40px high when horizontal, and 72 wide when vertical. 
The other toolbar should still be 36 high.

RomanPudashkin: We already have these size limits. The bar should use them
Screenshot 2021-11-25 at 14 16 19

cbjeukendrup: It is a bit inconsistent. In the other (non-floatable) toolbars, the content component depends on the fact that the height (thickness) of the toolbar is hardcoded to 36. Instead of hardcoding this value, I made a changeable property, because it is different for the NoteInputBar. 

For floating toolbars, the content height of the toolbars is used. This is mainly for the PlaybackToolbar, which should have a different height when floating, not the default 36px. Because of this, we also have to specify the desired height and width of the NoteInputBar, so that it gets the correct width when it's floating too.

Eventually we could make this consistent: 
- either make sure that for each toolbar the content item always specifies the correct size, and then use the size of that content item to determine the minimum and maximum size of the toolbar
- or use the `thickness` property consistently, so also for the PlaybackToolbar when it's floating.

The latter is doable but is not really nicer than what we have now, and the former is more work and has the disadvantage that we manually need to make sure that all four top toolbars have the same implicit content height of 36.

So what would you consider the best option?

cbjeukendrup: I tried various solutions, including the ones I mentioned in the previous comment, but also, using the maximum of the content height and the default thickness (36). See https://github.com/cbjeukendrup/MuseScore/commit/4e6cb60830f8fe9b7dd7891acd3070e785fd0890. But with everything I tried, it goes wrong when re-docking the Playback Toolbar:

https://user-images.githubusercontent.com/48658420/145690765-af7c9c0b-6184-4a0a-b79a-de78cfa3e275.mov

With the current situation in this PR, that problem does not occur. So maybe it's a good option to merge this as it is now, so that at least the visual with the note input bar, and then we can investigate later why other solutions don't work well with the Playback Toolbar.

jessjwilliamson:
Tested this on Windows, looks fixed to me! :)

RomanPudashkin (approved):

PR #9879, opened 2021-11-25
Clearing property value types (Fraction) [Step 7]

igorkorsukov:


src/engraving/types/constants.h
```diff
@@ -30,4 +30,9 @@ struct Constant
 };
 }
 
+//! NOTE compat
+namespace Ms {
+using Constant = mu::engraving::Constant;
```
cbjeukendrup, line 35: Wouldn't it be nicer to call this struct `Constants`? That sounds more natural to me (like, you take the "division" from the (list of) Constants).

igorkorsukov: ok, I agree

vpereverzev (approved):

PR #9883, opened 2021-11-25
Fix #9719 add/remove system breaks behavior

### LINKED ISSUE #9719: [MU4 Issue] Add/remove System Breaks dialog should should match MU3 behaviour
**Describe the bug**
**1.** If nothing is selected, when open **Add/remove System Breaks** dialog it should select the whole score to make clear to the user that System Breaks operation will be applied to the whole score (like **MU3** does)
Even so, MU4 makes System Breaks changes in that case (that's expected)

https://user-images.githubusercontent.com/90187801/141475455-2bc56c79-d0a5-4ab1-a9f4-fe165aca3d9a.mp4

**2.** If some object is selected (**NOT** range selection), when open **Add/remove System Breaks** dialog it should select the whole score and change System Breaks changes (like **MU3** does)
Now if rest or note is selected, MU4 doesn't make selection of whole score and **doesn't make System Breaks changes** (the 2nd part is worse)

https://user-images.githubusercontent.com/90187801/141477327-deffb1c0-d62c-4818-84b5-5e8cd22ce998.mp4

 

**Expected behavior**
**MU4** should behave like **MU3** 

**Desktop:**
Win10/Mac11/LinuxMint20

LovCAPONE:
Resolves: *https://github.com/musescore/MuseScore/issues/9719*

Implement BreaksDialog::showEvent method to show the user on which
range the breaks will be applied before the dialog is displayed.

Call Selection::isRange method to know if a range is already selected.

All of the code for the breaks dialog is in its accept method which is
called only when the dialog is closed. So, the problem is that there
is no way for the user to see the range on which the system breaks
will be applied.

Also, the code which sets the range on which the breaks will be applied
needs to first check if a range is selected. The problem is that in
the current implementation, it calls the Selection::isNone method.
The right method to call is Selection::isRange because we need to
know if a range is selected. Selection::isNone returns false even if
something other than a range is selected.

### DIFF
--- a/src/notation/view/widgets/breaksdialog.cpp
+++ b/src/notation/view/widgets/breaksdialog.cpp
@@ -56,14 +56,6 @@ void BreaksDialog::accept()
     }
 
     INotationInteractionPtr interaction = notation->interaction();
-    INotationSelectionPtr selection = interaction->selection();
-
-    bool allSelected = false;
-
-    if (selection->isNone()) {
-        interaction->selectAll();
-        allSelected = true;
-    }
 
     int interval = intervalButton->isChecked() ? intervalBox->value() : 0;
     BreaksSpawnIntervalType intervalType = BreaksSpawnIntervalType::MeasuresInterval;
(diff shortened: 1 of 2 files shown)


LovCAPONE:
This PR is ready for review.

RomanPudashkin (approved):

DmitryArefiev:
Tested #9719 on latest master build (Win10, Mac12, LinuxMint 20.2) - FIXED

Thanks!

Issue #9877, opened 2021-11-25
[MU4 Issue] Placement of bar repeat symbols is imperfect


oktophonie:
Either these symbols are a positioned tiny bit too low, or the scale of the symbol is a tiny bit too small. The dots should appear exactly in the middle of the spaces. Also the 2-bar/4-bar repeat symbols should be exactly centered on the middle of the barline (they are a bit too far to the right):

![image](https://user-images.githubusercontent.com/20604032/143466708-756c4579-e135-4fe2-9546-cf0e2f462c9d.png)

Dorico comparison (here the vertical placement is ideal, though the horizontal is not):

![image](https://user-images.githubusercontent.com/20604032/143466964-47c8394f-1aa0-4660-a201-69c5e9d16583.png)

(Font in these examples is Leland)

cbjeukendrup:
Working on it!

cbjeukendrup:
@oktophonie How should the bar repeat sign be positioned on staves with an even number of lines (for example, TAB staves)? Dorico just places it in the middle of the staff, which results in the dots colliding with the staff lines. Shouldn't the symbol be positioned on the line just above the middle? (So, for a 6 line staff, the third line from the top?)

oktophonie:
Not sure there's a standard for this, to be honest - but I'd prefer to do it as you suggest, i.e. where there's an even number of lines, centre the symbols on the line just above the middle.

PR #9886, opened 2021-11-25
Fix (vertical) positioning of Measure Repeat Signs

cbjeukendrup:
Resolves: #9877

Schermafbeelding 2021-11-25 om 22 15 25

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -363,7 +363,7 @@ void Beam::layout1()
             std::set<int> noteSet(_notes.begin(), _notes.end());
             notes.assign(noteSet.begin(), noteSet.end());
             _up = Chord::computeAutoStemDirection(&notes) > 0;
-            int middleStaffLine = firstNote->staff()->middleLine(firstNote->tick());
+            int middleStaffLine = firstNote->staffType()->middleLine();
             for (uint i = 0; i < _notes.size(); i++) {
                 _notes[i] += middleStaffLine;
             }
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -238,9 +238,10 @@ int Chord::downString() const
 
 std::vector<int> Chord::noteDistances() const
 {
-    Q_ASSERT(staff());
-    bool isTabStaff = staff()->staffTypeForElement(this)->isTabStaff();
-    int staffMiddleLine = staff()->middleLine(tick());
+    const StaffType* staffType = this->staffType();
+    Q_ASSERT(staffType);
+    bool isTabStaff = staffType->isTabStaff();
+    int staffMiddleLine = staffType->middleLine();
     int upStringAmount = 0;
     if (isTabStaff) {
         upStringAmount = upString() * 2;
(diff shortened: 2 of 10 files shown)


oktophonie (dismissed):

cbjeukendrup:
@oktophonie It turns out that this fix works perfectly with Leland, but in other fonts, the repeat sign seems to have a "built-in" offset, which seems to make no sense. Therefore, it now looks misplaced with fonts other than Leland. 

That offset _seems_ to be equal to half of the staff line width specified by the "style defaults" accompanying the font.

Before my fix, the position was adjusted by half of the _current_ staff line width of the score, so therefore, it was theoretically incorrect too when the user had modified the staff line width. 

So what we could do, is check whether the current ScoreFont is Leland, and otherwise query the default staff line width from the score font and adjust the symbol position accordingly. That will feel like a hack, but it'll be okay as long as we support a fixed set of six score fonts, but I wonder what will happen when I make it possible to use custom fonts (because we can't guess whether the offsetting is needed for those fonts). 

Anyway, maybe the biggest benefit of my fix, is that it now takes into account the number of staff lines (and the staff lines distance).

oktophonie:
Good point. Looking at the Bravura symbols they extend vertically from -250 to +265 (not -250 to +250, which would be from the second bottom line to the top bottom line exactly). That would explain the previous slightly-off scaling, etc.

I'm not sure what the reasoning for that is and the SMuFL spec doesn't enlighten me.

Is the simple solution not to vertically centre the bounding box (rather than the y=0) of these symbols on the middle of the staff line you want to position it on? Leland would then look just as it does now, and Bravura would similarly appear vertically centered.

cbjeukendrup:
> Is the simple solution not to vertically centre the bounding box (rather than the y=0) of these symbols on the middle of the staff line you want to position it on? Leland would then look just as it does now, and Bravura would similarly appear vertically centered.

Yes, that worked! Now it should look good with most fonts; only maybe with some asymmetric Jazz fonts it's not entirely how the font designer wants it. But that will be a minority.

I also realised that the positioning for the number above the repeat symbols was not positioned correctly on non-default staves. I fixed that, but this means that the interpretation for the value of the position has changed. So users who have changed that setting manually, will have to adjust it to fit the new behaviour when they open their scores in MuseScore 4. Or we could write compatibility code that does this automatically.

Jojo-Schmitz:
I'd think that these multimeasure repeats are used exceptionally frequently especially on Jazz leadsheets, wouldn't they?

RobFog:
And drum scores… :)

Jojo-Schmitz:
Which always (?) had an odd number of staff lines (1, 3 or 5), so shouldn't be an issue here

RobFog:
I write with five staff lines but I couldn't say if there are exceptions to the rule you mention.

cbjeukendrup:
Maybe in some cases where you have two woodblocks, or two cymbals, you might notate that on two lines. 

Anyway, the number of staff lines was not really an issue anymore; the problem was the centering of the symbol around the desired staff line. But that is solved now.

About Jazz fonts, yes, doubtlessly those bar repeats are not uncommon in Jazz scores, but for MuseJazz and Petaluma is looks very reasonable now, so there will probably be only few fonts with which it does not look good (which can't be used in MuseScore yet anyway).

oktophonie (approved):

chenlung:
> Which always (?) have an odd number of staff lines (1, 3 or 5), so shouldn't be an issue here?

Also two lines.

cbjeukendrup:
Rebased.

vpereverzev (approved):

PR #9888, opened 2021-11-26
Milimetre

### LINKED ISSUE #9840: [MU4 Issue] "Gap above/below" vertical frames displays incorrect values
**Describe the bug**
The **Gap above** and **Gap below** text edit fields within the **Vertical frames** Properties panel are showing incorrect values, causing unexpected layout behaviour when altered. 

https://user-images.githubusercontent.com/86290556/143016487-53a84a0f-9f72-4921-8710-e45e924f0ff9.mp4

These fields are also occasionally inclined to show an unnecessary number of decimal places: 
Screenshot 2021-11-23 at 10 28 31 pm

**Expected behavior**
In MS3.6, vertical frames appear to have a default padding of 7sp, although the **Gap above** doesn't seem to have any effect on the topmost vertical frame of a score (perhaps a bug in MS3.6?). At any rate, expected behaviour should include:

- 7sp default "Gap above/below" for vertical frames
- Incremental adjustments to the text edit field that map correctly to the actual repositioning of the corresponding frame in the score
- No more than the usual 2 decimal places in the text edit fields 

 - Windows 11
 - MacOS 11.6

**Additional context**
Does not seem to affect horizontal frames

### LINKED ISSUE #9844: [MU4 Issue] Line objects showing incorrect thickness in properties panel
**Describe the bug**
The default thickness of line objects (e.g. hairpins, cresc. and dim. lines) appears to be 2.98, which is much thicker than the actual line weight of the selected object. 
Moreover, changes made directly to the **Thickness** text edit field are replaced by the default value (2.98) once the corresponding object has been deselected and reselected.  

https://user-images.githubusercontent.com/86290556/143030363-e5135ed9-5512-40fe-bf2d-15ac01a0bb45.mp4

**Expected behaviour**

1. Default line thickness for these objects should be 0.12sp (@oktophonie please correct me if I'm wrong here).
2. The line weight of the selected object should correspond with the value shown in the **Thickness** text edit field in the **Properties** panel.

 - MacOS 11.6 (M1)
 - Windows 11

Howard-C:
Fixes #9840, fixes #9844


src/engraving/types/dimension.h
```diff
@@ -106,40 +106,40 @@ class Milimetre
...
+    r *= b;
+    return r;
+}
+
+inline Milimetre operator*(qreal a, const Milimetre& b)
+{
+    Milimetre r(b);
+    r *= a;
+    return r;
+}
```
Howard-C, line 142: @igorkorsukov Hi, may I ask why these operator functions were commented out? In this PR I needed to multiply a `Milimetre` with `qreal` so I brought these back to life, and I'm just wondering if there may be side effects.

igorkorsukov: I wanted to know if they were needed, so I commented out (I saw they were not needed and wanted to delete). If you need them, you can uncomment.
I just fear that when there are many such operators, then we can do whatever we want, however we want ... and in the code it may not be clear what is actually happening and how it all works.

Howard-C: Thanks for explaining. I am also curious as to why some of the operators aren't defined in the `Milimetre` class itself, like the first four that have a `Milimetre` variable as the first parameter. Is there some kind of tradition concerning what should be defined in the class and what should not be?

igorkorsukov: I believe that the operators with the first parameter are the target class, should be defined in the class declaration.
With one exception, there is no way to change the source code. For example, we cannot change the source code of QString, but we can add an operator outside the declaration of the QString class.
(at the moment these are just copy-paste from Spatium, I was planning to cleaning them)

Howard-C: Yes but couldn't we change for example
```
inline Milimetre operator+(const Milimetre& a, const Milimetre& b)
{
    Milimetre r(a);
    r += b;
    return r;
}
```
to
```
inline Milimetre Milimetre::operator+(const Milimetre& b)
{
    Milimetre r(*this);
    r += b;
    return r;
}
```
?

igorkorsukov: Yes, we can change to

Howard-C: Got it :)

igorkorsukov (approved):

Issue #9041, opened 2021-09-06
[MU4 Issue] [VJ] 'Max grand staff distance' behaves unpredictably


oktophonie:
Increasing 'Max. grand staff distance' works up to a point as expected, but stops at a certain value - possibly when limited by the 'Max staff distance' or 'Max page fill distance' settings? - and thereafter has no further effect on stave spacing, _but_ does seem to continue causing the curly brace to get thicker and thicker.

Possibly the brace is being scaled (horizontally) according to the theoretical distance that 'Max. grand staff distance' specifies, but not according to the actual spacing of the staves as they appear on the page.

<img width="2590" alt="bracket-fun" src="https://user-images.githubusercontent.com/20604032/132220249-3fbd2feb-78c1-4f76-8ab0-110b5ec0c337.png">

Jojo-Schmitz:
See also https://musescore.org/en/node/326130

njvdberg:
> We'd need so somehow detect and use the actual akkolade distance rather then using the max.

That's indeed what I'm trying to do :smile:

njvdberg:
See PR #9903

PR #9903, opened 2021-11-28
Always update _magx after changing the height of bracket.

njvdberg:
Resolves: https://github.com/musescore/MuseScore/issues/9041

This PR updates the `_magx`  of a bracket when the height of a bracket is changed. 

### DIFF
--- a/src/engraving/libmscore/bracket.cpp
+++ b/src/engraving/libmscore/bracket.cpp
@@ -144,10 +144,10 @@ void Bracket::setStaffSpan(int a, int b)
         if (score()->styleSt(Sid::MusicalSymbolFont) == "Leland") {
             v = qMin(4, v);
         }
-        // total default height of a system of n staves / height of a 5 line staff
-        qreal dist = score()->enableVerticalSpread() ? score()->styleS(Sid::maxAkkoladeDistance).val() : score()->styleS(
-            Sid::akkoladeDistance).val();
-        _magx = v + ((v - 1) * dist / 4.0);
+
+        // 1.625 is a "magic" number based on akkoladeDistance/4.0 (default value 6.5).
+        _magx = v + ((v - 1) * 1.625);
+
         if (v == 1) {
             _braceSymbol = SymId::braceSmall;
         } else if (v <= 2) {


DmitryArefiev:
Hi @njvdberg @oktophonie I've tested #9041 on Win10 for now and found some issues:

1. The default thickness of brace has changed
2. Gap between brace and staff changing after adding a note/rest
3. In Format>Style>Page> Max grand staff distance, Reset to default option is greyed out after after changes and after reopening style settings

https://user-images.githubusercontent.com/90187801/144258522-a29018ec-af76-4871-9bf9-753f6012942b.mp4

@oktophonie  Can you check this PR on your side as well please?

oktophonie:
I see the same issues a @DmitryArefiev. Is the wrong (too thick) brace symbol also being used to begin with?

njvdberg:
Ah, I missed that for some reason. But at least the results is better as without the PR :smile:

njvdberg:
@DmitryArefiev 
> 3. In Format>Style>Page> Max grand staff distance, Reset to default option is greyed out after making changes and reopening style settings

This is not related with this PR and solved with via PR #9904.

njvdberg:
@oktophonie When playing around with this issue I get more and more puzzled what the actual spec should be.
If `Vertical justification of staves` is disabled, the width of brace is calculate based on the number of staves spanned and the value of the parameter `Grand staff distance` (please note **not** the actual distance, which can be larger!!!!)
Here a number of examples:

`Grand staff distance` = 6.5sp
Screenshot_20211204_165506

`Grand staff distance` = 10.0sp
Screenshot_20211204_165600

`Grand staff distance` = 22.0sp
Screenshot_20211204_165701

Is this what is supposed to happen? Or should the width of the brace be depend on the number of spanned staves only?
The brace with `Grand staff distance` = 22.0sp then looks like:
Screenshot_20211204_165930

oktophonie:
I personally prefer your last example, i.e. using only the number of staves. The reason the stylistic alternates for the brace symbol exist is to provide different thicknesses for different contexts. Also, as the distance between staves grows larger, if we only scale the symbol vertically (and not horizontally), we also know it will have a consistent width, which is helpful for margin calculations and so on.

njvdberg:
I'm very happy with this answer :smile:. This make it straightforward to have a single solution for both `Vertical adjustment of staves` enabled and disabled.

Jojo-Schmitz:
What's up with the MacOS build here? Seems unrelated to this PR

src/engraving/libmscore/bracket.cpp
```diff
@@ -144,10 +144,10 @@ void Bracket::setStaffSpan(int a, int b)
         if (score()->styleSt(Sid::MusicalSymbolFont) == "Leland") {
             v = qMin(4, v);
         }
-        // total default height of a system of n staves / height of a 5 line staff
-        qreal dist = score()->enableVerticalSpread() ? score()->styleS(Sid::maxAkkoladeDistance).val() : score()->styleS(
-            Sid::akkoladeDistance).val();
-        _magx = v + ((v - 1) * dist / 4.0);
+
+        // 1.125 is a "magic" number based on akkoladeDistance/4.0 (default value 6.5).
+        _magx = v + ((v - 1) * 1.125);
```
Jojo-Schmitz, line 149: Any special reason not to take `score()->styleS(Sid::akkoladeDistance).val()` here then?

njvdberg: Yes, using this parameter is exactly what causes the issue this PR tries to solve :wink: 
Also, this parameter is a parameter used when vertical adjustment of staves is disabled, using it when this adjustment is **enabled** is very confusing (especially since the parameter is disabled).
The (new) idea of this PR is the make the scaling of the brace independent of the spacing, only the number of staves spanned is used. See also [this comment](https://github.com/musescore/MuseScore/pull/9903#issuecomment-986057266).

njvdberg:
> What's up with the MacOS build here? Seems unrelated to this PR

My idea too and I will push this PR again. But first I want to explain a strange behavior I see in my tests for `leland` only. It might be related to migration but I like to understand this first.

njvdberg:
Now the width of the brace scales with the number of spanned staves only, the value of any style parameter is no longer not taken into account.
The following images show the result of this change. These images show 3 different spacing between the two staves spanned by the brace. From left to right 650sp, 22.0sp and 6.5sp (default for vertical spread of staves disabled).

**Bravura, vertical spread of staves disabled**
Reference (master)
bravura-disabled-ref

For the larger values of `Grand staff distance` the wrong braces are clearly visible.
With the PR the braces will have the same width, independent of the value of `Grand staff distance`
bravura-disabled-cur
Comparing these both picture show there is no difference for the default value (6.5sp) of  `Grand staff distance`
diff-bravura-disabled

**Bravura, vertical spread of staves enabled**
Reference (master)
bravura-enabled-ref
Since the width of he brace is based on the the number of spanned staves **and** value of `Grand staff distance` (in all cases 65.0sp) the braces are clearly wrong in all cases.

The PR solves this by using the number of spanned staves only.
bravura-enabled-cur

**Leland, MuseJazz, Petaluma**
Similar images can be made for the fonts Leland, MuseJazz and Petaluma.

**Emmentaler, Gonville**
These font are an exception since these fonts don't use the symbol for the fonts. This PR doesn't touch these fonts.

oktophonie (approved):

PR #9905, opened 2021-11-28
Add shortcuts to navigate through notations

HemantAntony:
Resolves: No issue

This commit has added shortcuts to navigate through notations. If you
use the Ctrl+W shortcut, you can close a notation. You can switch to the
next and previous notations by using Ctrl+Tab and Ctrl+Shift+Tab
shortcuts. You can also set the current notation by using
Ctrl+Shift+#Number#.

### DIFF
--- a/src/framework/shortcuts/data/shortcuts-Mac.xml
+++ b/src/framework/shortcuts/data/shortcuts-Mac.xml
@@ -43,14 +43,6 @@
         <seq>Shift+Tab</seq>
     </SC>
     <SC>
-        <key>nav-next-tab</key>
-        <seq>Ctrl+Tab</seq>
-    </SC>
-    <SC>
-        <key>nav-prev-tab</key>
-        <seq>Ctrl+Shift+Tab</seq>
-    </SC>
-    <SC>
         <key>nav-right</key>
         <seq>Right</seq>
     </SC>
@@ -117,7 +109,7 @@
     <std>63</std>
     </SC>
   <SC>
-    <key>file-close</key>
+    <key>notation-close</key>
     <std>4</std>
     </SC>
   <SC>
@@ -841,12 +833,48 @@
     <seq>P</seq>
     </SC>
   <SC>
-    <key>next-score</key>
+    <key>notation-next</key>
     <std>20</std>
     </SC>
   <SC>
-    <key>previous-score</key>
-    <std>21</std>
+    <key>notation-previous</key>
+    <seq>Meta+Shift+Tab</seq>
+    </SC>
+  <SC>
+    <key>notation-change-1</key>
+    <seq>Ctrl+Shift+1</seq>
+    </SC>
+  <SC>
+    <key>notation-change-2</key>
+    <seq>Ctrl+Shift+2</seq>
+    </SC>
+  <SC>
+    <key>notation-change-3</key>
+    <seq>Ctrl+Shift+3</seq>
+    </SC>
+  <SC>
+    <key>notation-change-4</key>
+    <seq>Ctrl+Shift+4</seq>
+    </SC>
+  <SC>
+    <key>notation-change-5</key>
+    <seq>Ctrl+Shift+5</seq>
+    </SC>
+  <SC>
+    <key>notation-change-6</key>
+    <seq>Ctrl+Shift+6</seq>
+    </SC>
+  <SC>
+    <key>notation-change-7</key>
+    <seq>Ctrl+Shift+7</seq>
+    </SC>
+  <SC>
+    <key>notation-change-8</key>
+    <seq>Ctrl+Shift+8</seq>
+    </SC>
+  <SC>
+    <key>notation-change-9</key>
+    <seq>Ctrl+Shift+9</seq>
     </SC>
   <SC>
     <key>inspector</key>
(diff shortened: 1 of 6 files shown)


HemantAntony:
For some unknown reason, shortcut for Ctrl+Shift+Tab does not work when the shortcut is written as <std>21</std>, but works when it is written as <seq>Ctrl+Shift+Tab</seq>.

src/framework/shortcuts/data/shortcuts-Mac.xml
```diff
@@ -841,12 +833,48 @@
...
-    <key>next-score</key>
+    <key>notation-next</key>
     <std>20</std>
     </SC>
   <SC>
-    <key>previous-score</key>
-    <std>21</std>
+    <key>notation-previous</key>
+    <seq>Ctrl+Shift+Tab</seq>
+    </SC>
```
cbjeukendrup, line 842: One thing you couldn't know: on macOS, `Ctrl` maps to the Cmd key, and Cmd+Tab is a shortcut reserved by the OS (it does something similar to Alt+Tab on Windows). Instead, it is common for macOS apps to use Ctrl+Tab. To achieve that, you should write `Meta+Tab` in this `shortcuts-Mac.xml` file.

But actually, I think you should just use `<std>21</std>`. That will be interpreted as `QKeySequence::StandardKey::PreviousChild`, which maps to the desired shortcut automatically based on the platform.

HemantAntony: What about the Ctrl+Shift+#Number#. Should that still be Ctrl or Meta?

HemantAntony:
I would also like to point that there are a bunch of warnings that get printed whenever I do anything related to the notation tabs like clicking on them. There are a lot of messages. They usually cover up the console. Should I create an issue for this?

HemantAntony:
Hey, this pr is ready for merge

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1907,6 +1920,110 @@ void NotationActionController::toggleStrike()
     currentNotationInteraction()->toggleStrike();
 }
 
+bool NotationActionController::isMasterNotation(const INotationPtr notation) const
+{
+    return globalContext()->currentMasterNotation()->notation() == notation;
```
igorkorsukov, line 1925: You need to check everywhere that the current notation is not null, otherwise if an action is sent when the notation is not yet open, a crash will occur

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1907,6 +1920,110 @@ void NotationActionController::toggleStrike()
...
+}
+
+QList<INotationPtr> NotationActionController::getOpenNotations()
```
igorkorsukov, line 1928: We give preference to `std` data types instead of `Qt` (unless it's directly related to the UI)

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1907,6 +1920,110 @@ void NotationActionController::toggleStrike()
...
+    return notations;
+}
+
+void NotationActionController::closeCurrentNotation()
+{
+    INotationPtr nextNotation = nullptr;
+    IMasterNotationPtr masterNotation = globalContext()->currentMasterNotation();
+    QList<INotationPtr> notations = getOpenNotations();
+    for (int i = 0; i < notations.size(); i++) {
+        INotationPtr notation = notations[i];
```
igorkorsukov, line 1947: If you do not want to modify the value in the list, then it is better to use the `at` method

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1907,6 +1920,110 @@ void NotationActionController::toggleStrike()
...
+        globalContext()->setCurrentNotation(nextNotation);
+    }
+}
+
+void NotationActionController::openPreviousNotation()
+{
+    QList<INotationPtr> notations = getOpenNotations();
+    INotationPtr previousNotation = nullptr;
+    if (globalContext()->currentMasterNotation()->notation() == currentNotation()) { //If the main project is opened, previous part is previous notation
+        previousNotation = notations[notations.size() - 1];
```
igorkorsukov, line 1976: `notations` can be empty, then a crash will occur

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1907,6 +1920,110 @@ void NotationActionController::toggleStrike()
...
+    } else {
+        for (int i = 0; i < notations.size(); i++) {
+            if (notations[i] == currentNotation()) {
+                if (i != 0) {//If the opened part is not first
+                    previousNotation = notations[i - 1];
+                } else { //If the opened part is first, the next notation is the main project notation
+                    previousNotation = globalContext()->currentMasterNotation()->notation();
+                }
+                break;
+            }
```
igorkorsukov, line 1986: You have duplicated this logic here and in closeCurrentNotation. Better to make separate methods for getting next / previous notation and use them in close / open methods.
Each method should have one responsibility (S from SOLID https://en.wikipedia.org/wiki/SOLID)
Also we need to implement in such a way that the implementation of the logic can be re-used elsewhere, without having to re-implement it.

Issue #9907, opened 2021-11-29
[MU4 Issue] Crash relating to unpitched percussion instruments


bkunda:
**Describe the bug**
MuseScore crashes in either of the following situations:
1. Create new score containing only an unpitched percussion instrument (crash occurs when clicking "Done" in new score wizard)
2. In a score containing unpitched percussion instruments and any other instrument, crash occurs when triggering note entry mode for the unpitched percussion instrument

Demo vide: https://www.dropbox.com/s/1rsuqdccnu3h3dd/crash-unpitched-perc-instruments.mp4?dl=0

**Expected behavior**
MuseScore should not crash
Activating note entry mode for an unpitched percussion instrument should trigger percussion panel

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1)
 - Windows 11

igorkorsukov:
@DmitryArefiev Please check, PR https://github.com/musescore/MuseScore/pull/9917

PR #9917, opened 2021-11-29
Fixed calc default stem length

igorkorsukov:


vpereverzev (approved):

src/engraving/libmscore/chord.cpp
```diff
@@ -1491,7 +1491,7 @@ qreal Chord::calcDefaultStemLength()
     _minStemLength = calcMinStemLength();
     qreal minStemLengthSpaces = _minStemLength / _spatium;
 
-    int staffLineCount = staffItem->lines(tick());
+    int staffLineCount = staffItem ? staffItem->lines(tick()) : 4;
```
cbjeukendrup, line 1494: After #9886 is merged, you can probably also write 
```c++
int staffLineCount = staffType ? staffType->lines() : 4;
```
which eliminates is more efficient.

I also think that a default value of `5` instead of `4` would make more sense musically.

Nick-Mazuk:
It looks like this PR is in response to https://github.com/musescore/MuseScore/issues/9907, where unpitched percussion staves will crash MuseScore. If that's the case, few comments:

1) Is there a reason why the `staff()` is not defined for unpitched percussion staves? It seems like `staff()` should be defined for every instrument that has a staff. If there isn't a reason, I'd argue that this seg fault is a symptom, not a root cause. We should aim to fix root causes, not symptoms.
2) I agree with @cbjeukendrup that the default value should be 5, not 4, since 5-line staves are the most common. However, this value is purely used to determine the stem length (for instance, does the stem end outside the staff?). If the note is not on a staff, this code path should ideally not run. In which case, there would be no seg fault to begin with (i.e., so we wouldn't need a default value).

cbjeukendrup:
> 1. Is there a reason why the `staff()` is not defined for unpitched percussion staves?

My guess is that chords on percussion staves on the score do have a valid staff, but that it is the Drumset Panel where they don't have a staff. The fact that the crash occurs at the moments where the drum set panel would normally appear, supports that guess.

igorkorsukov:
> I also think that a default value of 5 instead of 4 would make more sense musically.
I agree with 5, I just looked at the number before, but it seems to be wrong, now I see that there is "lines -1"
https://github.com/musescore/MuseScore/blob/3.6.2/libmscore/chord.cpp#L1389

DmitryArefiev:
Tested #9907 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Issue #9775, opened 2021-11-18
[MU4 Issue] "Pan score automatically" does not work


bkunda:
**Describe the bug**
Two parts to this issue: 

1. "Pan score automatically" does not follow the score during **playback**.
2. "Pan score automatically" does not follow the note entry point in **note entry** mode

In both cases, the score zooms in to a high magnification on an outermost corner of the document.

(Apologies for the lengthy video, but I thought it was worth demonstrating current behaviour and why it's important)

https://user-images.githubusercontent.com/86290556/142393675-55142bba-3585-4fbf-9683-3d8a90d6278f.mp4

**Expected behavior**
Replicating MS3.6 behaviour would be a good fix for now.
I.e. When "Pan score automatically" is switched ON, the current playback point should be followed during playback, and when the user is in note entry mode, the score should follow the active note entry point.

**Desktop (please complete the following information):**
 - MacOS 11.6
 - Windows 11

jessjwilliamson:
Was about to report this! Thanks @bkunda!

MarcSabatella:
BTW, speaking of behavior in note input mode: while I like it as is, it's worth mentioning that a significant number of people would prefer to see the score track note the *input cursor*, but the *current selection*.  This is an important distinction when you reach the last note of the last measure of a page while entering notes - does the view stay on the last measure you just completed, or jump to the new?  If you're entering music for a single voice instrument like flute with no chords, you like that we currently do move on to the next page, so you can see the measure you are about to enter notes into.  But in piano and guitar music, it's extremely common that after entering one note on the last beat of the measure, you actually intend to add more notes to form a chord.  For keyboard users, no big deal since Shift+letter still works.  For mouse users it feels extremely wrong to have to constantly scroll back to click in another note.

I don't have a solution - again, I *like* the current behavior.  But it's something to think about - is there a way to somehow have the best of both worlds?

bkunda:
Very good point @MarcSabatella. I'm adding this to my list of design improvements. We may need to think a few things through to achieve a "best of both worlds" scenario, so it will most likely be a post-MS4 improvement 🙂

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED (should be in the next master)

Issue #9912, opened 2021-11-29
[MU4 Issue] Score position changes when interacting with empty tablature staff


bkunda:
**Describe the bug**
Selecting an empty bar on a tablature staff, as well as entering note entry mode on the selected staff, changes the position of the score. 

![score-position-changes-tablature](https://user-images.githubusercontent.com/86290556/143854485-b04d5574-5212-4baf-bb27-6e0d277104eb.gif)

This is pretty crucial as it is currently almost impossible to enter notation into tablature staves with the current behaviour. 

**Expected behavior**
Score should not change position when user clicks on a tablature staff. 

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1) 
 - Windows 11

**Additional context**
May relate to #9775

igorkorsukov:
It makes sense to fix it after solving this issues:  https://github.com/musescore/MuseScore/issues/9812

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED (should be in the next master)

Issue #9812, opened 2021-11-22
[MU4 Issue] Score position changes in unpredictable ways when selecting any object that is partially out-of-view


bkunda:
**Describe the bug**
Selecting any object that is partially out-of-view (i.e. not 100% in the score window view) repositions the score in seemingly unpredictable ways.
This significantly impairs workflow (the score keeps moving away from the selection point).

https://user-images.githubusercontent.com/86290556/142843369-2e6dac26-e148-42ac-a9c7-4d60b97eb818.mp4

**Expected behavior**
Score should not reposition when objects are selected.

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1)
 - Windows 11

**Additional context**
Similar problem as reported in #9615.  
N.B. video show score repositioning to top right hand corner. Previous tests I have ran have seen the score move to the bottom right corner as well.

Issue #9615, opened 2021-10-29
[MU4 Issue] Score position is changing weirdly during playback or using Ctr+click selection


DmitryArefiev:
https://user-images.githubusercontent.com/90187801/139476853-1453ba97-5dad-42ad-93b3-aec3f65c807c.mp4

https://user-images.githubusercontent.com/90187801/139477420-561e1aab-cde3-404a-a8da-7ab133bd618a.mp4

**To Reproduce**
1. Open score with more than 3 pages
2. Start playback

or

1. Open score with more than 3 pages
2. Ctr (on Win) + click on note on 1st p. 
3. Ctr + click on note on 2nd p. 
4. Ctr + click on note on 3rd p. 

**Desktop**
 - OS: Win10/Mac11/LinuxMint20

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED (should be in the next master)

PR #9923, opened 2021-11-29
Fix panning score during playback and note input and when selecting elements

cbjeukendrup:
Resolves: #9775

By fixing NotationPaintView::canvasPos, which fixes in turn NotationPaintView::adjustCanvasPosition.

Also replaced some QPointF to PointF.

### DIFF
--- a/src/notation/view/notationnavigator.cpp
+++ b/src/notation/view/notationnavigator.cpp
@@ -86,7 +86,7 @@ void NotationNavigator::rescale()
         return;
     }
 
-    setScaling(scaling, QPointF());
+    setScaling(scaling, PointF());
 }
 
 void NotationNavigator::wheelEvent(QWheelEvent*)
--- a/src/notation/view/notationpaintview.cpp
+++ b/src/notation/view/notationpaintview.cpp
@@ -716,7 +716,7 @@ qreal NotationPaintView::currentScaling() const
     return m_matrix.m11();
 }
 
-void NotationPaintView::setScaling(qreal scaling, const QPointF& pos)
+void NotationPaintView::setScaling(qreal scaling, const PointF& pos)
 {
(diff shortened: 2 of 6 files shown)


DmitryArefiev:
Tested (#9775, #9912, #9812, #9615) on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Thanks!

bkunda:
This is great! Thanks so much @cbjeukendrup ☺️

My only request for this, to bring it more in-line with MS3.6 behaviour, is that panning during note entry mode should position the system higher than it currently does in this build. It mainly affects higher zoom levels when working on upper staves in a system. At the moment, the upper staff appears at the bottom of the score window, whereas in MS3.6 the entire system always seems to be visible (I dare not say "vertically centred", because it seems to sit slightly lower than the vertical centre point. But systems are certainly higher up than where this build repositions them while panning).

Just to give you a side-by-side of Ms3.6 v. this build's behaviour (both 4-staff systems at 252% zoom, entering notes on uppermost staff):

MS3.6
MS3 6-pan-252-highest-staff

This build
MS4-pan-252-highest-staff

Let me know if this needs to be raised as a new issue, or if it can be incorporated into this PR before merging. Thanks!

RomanPudashkin (approved):

DmitryArefiev:
Hello @bkunda, we decided to merge it because a lot of issues were fixed with this PR!
Can you log the issue above separately please (about matching with MU3.6 regarding note input) ? 

Thank you! :)

cbjeukendrup:
@DmitryArefiev I think that's a good idea! I'll try to tackle the discrepancies between mu3 and mu4 as soon as I can.

bkunda:
No problem @DmitryArefiev, and will do! Thank you :-)

PR #6312, opened 2020-07-08
[MU3] improve horizontal spacing & vertical alignment +collect_artifacts

MarcSabatella:
This is a trial implementation of a spacing improvement that seems likely to be good.

Further explanation to follow...

PR #5840, opened 2020-03-19
[MU4] exclude accidentals etc. from chord shape if autoplace disabled

MarcSabatella:
There are a number of spacing issues that stem (pun not intended)
from the inclusion of things other than the notes themselves
in chord shapes:
    - despite the best intentions of "smart layout" changes for 3.0,
      accidentals can't tuck under the previous note
    - accidentals in a measure result in increased note spacing,
      even if the measure ie wider enough to not need extra space
    - for the latter, similar for arpeggios, fingering, etc

There has been good discussion and proposals for ways of fixing this,
but they require relatively significant changes to the layout,
probably require new options to support, etc.
And the tucking issue might not be solvable at all for beamed notes
without changes to the beaming algorithm as well.

This commit provides a manual workaround:
allowing the user to disable autoplace for a note,
which results in the accidentals and other attached symbols
being excluded from the chord shape.
The note itself is still added to the chord shape,
so the note is spaced as it normally would be
if it didn't have accidentals etc.
That means you will get collisions if spacing is tight,
so don't disable autoplace unless there is definitely room.
And in that case, the spacing automatically improves,
plus accidentals will tuck under previous notes where appropriate.

I'm not calling this a fix because eventually this should be automatic.
But see:

https://musescore.org/en/node/289446
https://musescore.org/en/node/299741
https://musescore.org/en/node/285591

I have not yet created vtests because I'm not sure we're ready for that, I expect the possibility of further changes based on review.  I'm therefore marking this WIP, and not necessarily recommending this be included in 3.5 unless there is strong agreement it makes sense to do this now while we await a good autoamtic solution.

PR #9928, opened 2021-11-30
Fix #299741 and #289446 - Improved horizontal spacing algorithm 

mike-spa:
Fix #299741 and #289446

Ensures consistent spacing of same-duration notes within the same system.
Behaves much better with measures containing many accidentals.
Has customizable spacing formula (possibly to be chosen in Style settings). 
The default spacing formula now perfectly replicates the spacing suggested by Gould.
Provides smoother behaviour to the } and { stretch commands.

Please see attached PDF for full documentation.

- [x] I signed [CLA]
- [x] I made sure the code in the PR follows [the coding rules]
 
[An improved horizontal spacing algorithm for Musescore.pdf](https://github.com/musescore/MuseScore/files/7625701/An.improved.horizontal.spacing.algorithm.for.Musescore.pdf)

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -324,7 +324,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                     } else {
                         m->removeSystemTrailer();
                     }
-                    m->computeWidth(m->minSysTicks(), oldStretch);
+                    m->computeWidth(m->system()->minSysTicks(), oldStretch);
                     m->layoutMeasureElements();
                     LayoutBeams::restoreBeams(m);
                     if (m == nm || !m->noBreak()) {
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -378,8 +378,6 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             } else if (mb->isMeasure()) {
                 Measure* m  = toMeasure(mb);
                 mw          += m->width();                       // measures are stretched already with basicStretch()
-                //int weight   = m->layoutWeight();
-                //totalWeight += weight * m->basicStretch();
                 qreal weight = m->stretchWeight(); // The stretch is now assigned proportionally to the width
                 totalWeight += weight; // No need now to multiply again by m->basicStretch()
             }
@@ -417,8 +415,6 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             mb->setPos(pos);
             Measure* m = toMeasure(mb);
             qreal stretch = m->basicStretch();
-            //int weight = m->layoutWeight();
-            //ww  += rest * weight * stretch;
             qreal weight = m->stretchWeight(); // Updated stretching formula
             ww += rest * weight;
             m->stretchMeasure(ww);
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -86,7 +86,8 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     bool curTrailer = ctx.curMeasure->trailer();
     MeasureBase* breakMeasure = nullptr;
 
-    Fraction minSysTicks = Fraction(4, 1); // Inizialize variable which stores the shortest note of the system
+    Fraction minTicks = Fraction(1, 8); // Inizialize variable which stores the shortest note of the system
+    Fraction prevMinTicks = Fraction(1, 1);
     bool changeMinSysTicks = false;
 
     while (ctx.curMeasure) {      // collect measure for system
@@ -96,10 +97,12 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         qreal ww  = 0;          // width of current measure
 
         // After appending a new measure, the shortest note in the system may change, in which case
-        // we need to recompute the layout of the previous measures. By updating the width of these
+        // we need to recompute the layout of the previous measures. When updating the width of these
         // measures, minWidth must be updated accordingly.
         if (ctx.curMeasure->isMeasure()) {
-            if (toMeasure(ctx.curMeasure)->computeTicks() < minSysTicks) {
+            if (toMeasure(ctx.curMeasure)->computeTicks() < minTicks) {
+                prevMinTicks = minTicks; // We save the previous value in case we need to restore it (see later)
+                minTicks = toMeasure(ctx.curMeasure)->computeTicks();
                 changeMinSysTicks = true;
                 for (MeasureBase* mb : system->measures()) {
                     if (mb == ctx.curMeasure) {
@@ -107,7 +110,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                     }
                     if (mb->isMeasure()) {
                         qreal prevWidth = toMeasure(mb)->width();
-                        toMeasure(mb)->computeMinWidth();
+                        toMeasure(mb)->computeMinWidth(minTicks);
                         qreal newWidth = toMeasure(mb)->width();
                         minWidth += newWidth - prevWidth;
                     }
@@ -149,7 +152,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             } else {
                 m->addSystemTrailer(m->nextMeasure());
             }
-            m->computeMinWidth();
+            m->computeMinWidth(minTicks);
             ww = m->width();
         } else if (ctx.curMeasure->isHBox()) {
             ctx.curMeasure->computeMinWidth();
@@ -185,12 +188,13 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 ctx.curMeasure->setParent(oldSystem);
             }
             // If the last appended measure caused a re-layout of the previous measures, now that we are
-            // removing it we need to re-layout the previous measures again.
+            // removing it we need to re-layout the previous measures again. 
             if (changeMinSysTicks) {
+                minTicks = prevMinTicks; // If the last measure caused it to change, now we need to restore it!
                 for (MeasureBase* mb : system->measures()) {
                     if (mb->isMeasure()) {
                         qreal prevWidth = toMeasure(mb)->width();
-                        toMeasure(mb)->computeMinWidth();
+                        toMeasure(mb)->computeMinWidth(minTicks);
                         qreal newWidth = toMeasure(mb)->width();
                         minWidth += newWidth - prevWidth;
                     }
@@ -224,7 +228,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                     Segment* s = m1->findSegmentR(SegmentType::StartRepeatBarLine, Fraction(0, 1));
                     if (!s->enabled()) {
                         s->setEnabled(true);
-                        m1->computeMinWidth();
+                        m1->computeMinWidth(minTicks);
                         ww = m1->width();
                     }
                 }
@@ -317,7 +321,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                     } else {
                         m->removeSystemTrailer();
                     }
-                    m->computeMinWidth();
+                    m->computeMinWidth(m->minSysTicks());
                     m->stretchMeasure(m->oldWidth());
                     LayoutBeams::restoreBeams(m);
                     if (m == nm || !m->noBreak()) {
@@ -359,7 +363,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             qreal w = lm->width();
             lm->addSystemTrailer(nm);
             if (lm->trailer()) {
-                lm->computeMinWidth();
+                lm->computeMinWidth(minTicks);
             }
             minWidth += lm->width() - w;
         }
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -169,7 +169,10 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // check if lc.curMeasure fits, remove if not
         // collect at least one measure and the break
 
-        bool doBreak = (system->measures().size() > 1) && ((curSysWidth + ww) > systemWidth * 1.025);
+        static constexpr acceptanceRange = 1.025; // We allow the initial width of the system to be slightly
+        // larger than the target system width. This avoids having to put a measure in the next system just because
+        // it overshoots the width by a tiny amount.
+        bool doBreak = (system->measures().size() > 1) && ((curSysWidth + ww) > systemWidth * acceptanceRange);
         if (doBreak) {
             breakMeasure = ctx.curMeasure;
             system->removeLastMeasure();
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -112,8 +112,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                         minWidth += newWidth - prevWidth;
                     }
                 }
-            }
-            else {
+            } else {
                 changeMinSysTicks = false;
             }
         }
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -90,7 +90,6 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     bool changeMinSysTicks = false;
 
     while (ctx.curMeasure) {      // collect measure for system
-
         System* oldSystem = ctx.curMeasure->system();
         system->appendMeasure(ctx.curMeasure);
 
@@ -99,11 +98,13 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // After appending a new measure, the shortest note in the system may change, in which case
         // we need to recompute the layout of the previous measures. By updating the width of these
         // measures, minWidth must be updated accordingly.
-        if (ctx.curMeasure->isMeasure())
+        if (ctx.curMeasure->isMeasure()) {
             if (toMeasure(ctx.curMeasure)->computeTicks() < minSysTicks) {
                 changeMinSysTicks = true;
                 for (MeasureBase* mb : system->measures()) {
-                    if (mb == ctx.curMeasure) break; // Cause I want to change only previous measures, not current one
+                    if (mb == ctx.curMeasure) {
+                        break; // Cause I want to change only previous measures, not current one
+                    }
                     if (mb->isMeasure()) {
                         qreal prevWidth = toMeasure(mb)->width();
                         toMeasure(mb)->computeMinWidth();
@@ -112,7 +113,10 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                     }
                 }
             }
-            else changeMinSysTicks = false;
+            else {
+                changeMinSysTicks = false;
+            }
+        }
 
         if (ctx.curMeasure->isMeasure()) {
             Measure* m = toMeasure(ctx.curMeasure);
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -188,7 +188,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 ctx.curMeasure->setParent(oldSystem);
             }
             // If the last appended measure caused a re-layout of the previous measures, now that we are
-            // removing it we need to re-layout the previous measures again. 
+            // removing it we need to re-layout the previous measures again.
             if (changeMinSysTicks) {
                 minTicks = prevMinTicks; // If the last measure caused it to change, now we need to restore it!
                 for (MeasureBase* mb : system->measures()) {
@@ -421,7 +421,6 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             }
             mb->setPos(pos);
             Measure* m = toMeasure(mb);
-            qreal stretch = m->basicStretch();
             qreal weight = m->stretchWeight(); // Updated stretching formula
             ww += rest * weight;
             m->stretchMeasure(ww);
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -86,7 +86,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     bool curTrailer = ctx.curMeasure->trailer();
     MeasureBase* breakMeasure = nullptr;
 
-    Fraction minTicks = Fraction(10000, 1); // Initializing this variable at a random high value.
+    Fraction minTicks = Fraction::max(); // Initializing this variable at an arbitrary high value.
     // In principle, it just needs to be longer than any possible note.
     Fraction prevMinTicks = Fraction(1, 1);
     bool changeMinSysTicks = false;
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -74,7 +74,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     Fraction lcmTick = ctx.curMeasure->tick();
     system->setInstrumentNames(ctx, ctx.startWithLongNames, lcmTick);
 
-    qreal minWidth    = 0;
+    qreal curSysWidth    = 0;
     qreal layoutSystemMinWidth = 0;
     bool firstMeasure = true;
     bool createHeader = false;
@@ -99,7 +99,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
 
         // After appending a new measure, the shortest note in the system may change, in which case
         // we need to recompute the layout of the previous measures. When updating the width of these
-        // measures, minWidth must be updated accordingly.
+        // measures, curSysWidth must be updated accordingly.
         if (ctx.curMeasure->isMeasure()) {
             if (toMeasure(ctx.curMeasure)->computeTicks() < minTicks) {
                 prevMinTicks = minTicks; // We save the previous value in case we need to restore it (see later)
@@ -113,7 +113,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                         qreal prevWidth = toMeasure(mb)->width();
                         toMeasure(mb)->computeWidth(minTicks, 1);
                         qreal newWidth = toMeasure(mb)->width();
-                        minWidth += newWidth - prevWidth;
+                        curSysWidth += newWidth - prevWidth;
                     }
                 }
             } else {
@@ -124,9 +124,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         if (ctx.curMeasure->isMeasure()) {
             Measure* m = toMeasure(ctx.curMeasure);
             if (firstMeasure) {
-                layoutSystemMinWidth = minWidth;
-                system->layoutSystem(ctx, minWidth, ctx.firstSystem, ctx.firstSystemIndent);
-                minWidth += system->leftMargin();
+                layoutSystemMinWidth = curSysWidth;
+                system->layoutSystem(ctx, curSysWidth, ctx.firstSystem, ctx.firstSystemIndent);
+                curSysWidth += system->leftMargin();
                 if (m->repeatStart()) {
                     Segment* s = m->findSegmentR(SegmentType::StartRepeatBarLine, Fraction(0, 1));
                     if (!s->enabled()) {
@@ -168,7 +168,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // check if lc.curMeasure fits, remove if not
         // collect at least one measure and the break
 
-        bool doBreak = (system->measures().size() > 1) && ((minWidth + ww) > systemWidth * 1.035);
+        bool doBreak = (system->measures().size() > 1) && ((curSysWidth + ww) > systemWidth * 1.025);
         if (doBreak) {
             breakMeasure = ctx.curMeasure;
             system->removeLastMeasure();
@@ -184,7 +184,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 ctx.curMeasure  = ctx.prevMeasure;
                 ctx.prevMeasure = ctx.curMeasure->prevMeasure();
 
-                minWidth -= system->lastMeasure()->width();
+                curSysWidth -= system->lastMeasure()->width();
                 system->removeLastMeasure();
                 ctx.curMeasure->setParent(oldSystem);
             }
@@ -197,7 +197,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                         qreal prevWidth = toMeasure(mb)->width();
                         toMeasure(mb)->computeWidth(minTicks, 1);
                         qreal newWidth = toMeasure(mb)->width();
-                        minWidth += newWidth - prevWidth;
+                        curSysWidth += newWidth - prevWidth;
                     }
                 }
             }
@@ -217,7 +217,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             if (m->trailer()) {
                 qreal ow = m->width();
                 m->removeSystemTrailer();
-                minWidth += m->width() - ow;
+                curSysWidth += m->width() - ow;
             }
             // if the prev measure is an end repeat and the cur measure
             // is an repeat, the createEndBarLines() created an start-end repeat barline
@@ -240,7 +240,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             // but at this point we have no choice but to assume it isn't
             // since we don't know yet if another true measure will fit
             // worst that happens is we don't get the automatic double bar before a courtesy key signature
-            minWidth += m->createEndBarLines(false);          // create final barLine
+            curSysWidth += m->createEndBarLines(false);          // create final barLine
         }
 
         MeasureBase* mb = ctx.curMeasure;
@@ -283,11 +283,11 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
 
         LayoutMeasure::getNextMeasure(options, ctx);
 
-        minWidth += ww;
+        curSysWidth += ww;
 
         // ElementType nt = lc.curMeasure ? lc.curMeasure->type() : ElementType::INVALID;
         mb = ctx.curMeasure;
-        bool tooWide = false;     // minWidth + minMeasureWidth > systemWidth;  // TODO: noBreak
+        bool tooWide = false;     // curSysWidth + minMeasureWidth > systemWidth;  // TODO: noBreak
         if (lineBreak || !mb || mb->isVBox() || mb->isTBox() || mb->isFBox() || tooWide) {
             break;
         }
@@ -344,15 +344,15 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     if (ctx.prevMeasure && ctx.prevMeasure->isMeasure()) {
         LayoutBeams::breakCrossMeasureBeams(ctx, toMeasure(ctx.prevMeasure));
         qreal w = toMeasure(ctx.prevMeasure)->createEndBarLines(true);
-        minWidth += w;
+        curSysWidth += w;
     }
 
     hideEmptyStaves(score, system, ctx.firstSystem);
     // Relayout system decorations to reuse space properly for
     // hidden staves' instrument names or other hidden elements.
-    minWidth -= system->leftMargin();
+    curSysWidth -= system->leftMargin();
     system->layoutSystem(ctx, layoutSystemMinWidth, ctx.firstSystem, ctx.firstSystemIndent);
-    minWidth += system->leftMargin();
+    curSysWidth += system->leftMargin();
 
     //-------------------------------------------------------
     //    add system trailer if needed
@@ -368,7 +368,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             if (lm->trailer()) {
                 lm->computeWidth(minTicks, 1);
             }
-            minWidth += lm->width() - w;
+            curSysWidth += lm->width() - w;
         }
     }
 
@@ -376,9 +376,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     // Gradient descent method: calls the computeWidth() function with a stretch parameter
     // proportional to the difference between the current length and the target length.
     // After few iterations, the length will converge to the target length.
-    qreal newRest = systemWidth - minWidth;
-    if (ctx.curMeasure == 0 && ((minWidth / systemWidth) <= score->styleD(Sid::lastSystemFillLimit))) {
-        // We do not stretch last system if minWidth is <= lastSystemFillLimit
+    qreal newRest = systemWidth - curSysWidth;
+    if (ctx.curMeasure == 0 && ((curSysWidth / systemWidth) <= score->styleD(Sid::lastSystemFillLimit))) {
+        // We do not stretch last system if curSysWidth is <= lastSystemFillLimit
         newRest = 0;
     }
     if (MScore::noHorizontalStretch) { // Debug feature
@@ -388,19 +388,19 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     qreal prevWidth = 0;
     int iter = 0;
     while (abs(newRest) > score->spatium() * 0.08 && iter < 200) { // spatium*0.08 is about the width of a note stem
-        stretchCoeff += 1.5 * newRest / minWidth;
+        stretchCoeff += 1.5 * newRest / curSysWidth;
         for (MeasureBase* mb : system->measures()) {
             if (mb->isMeasure()) {
                 Measure* m = toMeasure(mb);
                 prevWidth = m->width();
                 m->computeWidth(minTicks, stretchCoeff);
-                minWidth += m->width() - prevWidth;
+                curSysWidth += m->width() - prevWidth;
             }
         }
-        newRest = systemWidth - minWidth;
+        newRest = systemWidth - curSysWidth;
         iter++;
     }
-    
+
     // LAYOUT MEASURES
     PointF pos;
     firstMeasure = true;
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -110,7 +110,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                     }
                     if (mb->isMeasure()) {
                         qreal prevWidth = toMeasure(mb)->width();
-                        toMeasure(mb)->computeMinWidth(minTicks);
+                        toMeasure(mb)->computeMinWidth(minTicks, 1);
                         qreal newWidth = toMeasure(mb)->width();
                         minWidth += newWidth - prevWidth;
                     }
@@ -152,7 +152,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             } else {
                 m->addSystemTrailer(m->nextMeasure());
             }
-            m->computeMinWidth(minTicks);
+            m->computeMinWidth(minTicks, 1);
             ww = m->width();
         } else if (ctx.curMeasure->isHBox()) {
             ctx.curMeasure->computeMinWidth();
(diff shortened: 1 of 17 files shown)


Jojo-Schmitz:
The vtests failures are probably to be expected, but the codestyle (pretty straight forward to fix) and utest (why failing on beams???) failures certainly are not

Jojo-Schmitz:
Seems a `vtests` label is in order (a quick eyeballing the diffs doesn't show any unwanted chnages there)

MarcSabatella:
Wow, thanks for tackling this!  A surprisingly small amount of code for what it is trying to accomplish.  It would be great to see some samples of it in action; some new vtests would be a good idea assuming we're still doing that.

mike-spa:
> The vtests failures are probably to be expected, but the failures on codestyle (pretty straight forward to fix) and utest (why failing on beams???) certainly are not.
> 
> Seems a `vtests` label is in order (a quick eyeballing the diffs doesn't show any unwanted changes there)

Forgive me guys, I'm a total noob. This is the first time I've ever contributed to an open source project so I'm a bit overwhelmed. I don't know what these tests mean and why the code is failing them, or how I should fix that. What should I do?

Jojo-Schmitz:
First check on the code style check, it does tell you exactly how it woans the code to be formatted like to pass. Low hanging fruit ;-)

Why the utests fail is beyond me. Not sure whether they do reveal a coding problem or it is just the tests that need to get adjusted.

The vtests are, as mentioned, supposed to fail, and now with that `vtetsts`lable on this PR will be ignored, not counted as failed

asattely:
> 

You've done a great job, don't worry about the weird github particulars. If you look at the 'Details' link next to the failing CI_codestyle check, you can see where it wants you to make changes. From looking at it myself, it seems like the majority of the cases are just adding brackets to conditionals and putting brackets for functions on the next line. Once you fix those, you can force push your changes to the same branch and it'll run the checks again.

Nick-Mazuk:
Welcome @mike-spa!

This is a great first commit. To answer a few of your questions about the tests:

### utests

The "utests" are our unit tests, and they help ensure things work properly.

One of the things it checks for are if the beams are at the correct vertical positions. If you click "details" next to the failing tests, you'll see some snippets that look like this:

```
7: --- copypastesymbollist-lyrics.mscx	2021-11-30 13:24:13.342782365 +0000
7: +++ /home/runner/work/MuseScore/MuseScore/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx	2021-11-30 13:06:57.277409640 +0000
7: @@ -287,7 +287,7 @@
7:        <Measure>
7:          <voice>
7:            <Beam>
7: -            <l1>30</l1>
7: +            <l1>24</l1>
7:              <l2>32</l2>
7:              </Beam>
7:            <Chord>
7:    <diff -u copypastesymbollist-lyrics.mscx 
```

This essentially tells you the difference between the previous xml output and the new xml output. The `<l1>` and `<l2>` tags simply show the vertical positions of the beams in 1/8 space units. And because the values are different than they were before, it tells us that the beams are at different vertical positions than they were previously.

Now why is it different when you didn't touch the beam code? A beam's slant is partially determined by the spacing between the first and last notes of the beam group. So when you adjust the horizontal spacing, I'd expect some beams to have different slants (and therefore change their vertical positions). I need to investigate further to know if this is the exact reason, but it's expected that these tests would fail. You probably don't need to worry about the utests failing right now.

### vtests

Vtests are just visual snapshot tests. Imagine exporting the score as a PNG from both the current version of MuseScore and the version with your proposed changes. Then, compare those PNGs pixel by pixel to see if there's a difference. That's essentially how the vtests work.

So anytime you make visual changes to the engraved layout, it's possible the vtests will fail. The next question to ask is "why did the vtests fail?" If the changes are the exact visual changes you're expecting, great! No need to worry about fixing them (the change is desired). If not, well, that's what the vtests are for.

I looked through the test output, and everything looks good. The only differences deal with horizontal spacing, and it looks like it's the kind of changes you intentionally made. I wouldn't worry about fixing them.

---

Hope this helps!

mike-spa:
> Welcome @mike-spa!
> 
> This is a great first commit. To answer a few of your questions about the tests:
> 
> ### utests
> The "utests" are our unit tests, and they help ensure things work properly.
> 
> One of the things it checks for are if the beams are at the correct vertical positions. If you click "details" next to the failing tests, you'll see some snippets that look like this:
> 
> ```
> 7: --- copypastesymbollist-lyrics.mscx	2021-11-30 13:24:13.342782365 +0000
> 7: +++ /home/runner/work/MuseScore/MuseScore/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx	2021-11-30 13:06:57.277409640 +0000
> 7: @@ -287,7 +287,7 @@
> 7:        <Measure>
> 7:          <voice>
> 7:            <Beam>
> 7: -            <l1>30</l1>
> 7: +            <l1>24</l1>
> 7:              <l2>32</l2>
> 7:              </Beam>
> 7:            <Chord>
> 7:    <diff -u copypastesymbollist-lyrics.mscx 
> ```
> 
> This essentially tells you the difference between the previous xml output and the new xml output. The `<l1>` and `<l2>` tags simply show the vertical positions of the beams in 1/8 space units. And because the values are different than they were before, it tells us that the beams are at different vertical positions than they were previously.
> 
> Now why is it different when you didn't touch the beam code? A beam's slant is partially determined by the spacing between the first and last notes of the beam group. So when you adjust the horizontal spacing, I'd expect some beams to have different slants (and therefore change their vertical positions). I need to investigate further to know if this is the exact reason, but it's expected that these tests would fail. You probably don't need to worry about the utests failing right now.
> 
> ### vtests
> Vtests are just visual snapshot tests. Imagine exporting the score as a PNG from both the current version of MuseScore and the version with your proposed changes. Then, compare those PNGs pixel by pixel to see if there's a difference. That's essentially how the vtests work.
> 
> So anytime you make visual changes to the engraved layout, it's possible the vtests will fail. The next question to ask is "why did the vtests fail?" If the changes are the exact visual changes you're expecting, great! No need to worry about fixing them (the change is desired). If not, well, that's what the vtests are for.
> 
> I looked through the test output, and everything looks good. The only differences deal with horizontal spacing, and it looks like it's the kind of changes you intentionally made. I wouldn't worry about fixing them.
> 
> Hope this helps!

Aha! That makes sense, thanks a lot! The vtest is supposed to fail, and you're probably right about the beams, that's a relief! I've re-committed the code correcting for the style errors, it should be ok now!

Jojo-Schmitz:
OK, so it seems (!) we'd then need to adjust those utest to the changed spacing. @Nick-Mazuk are you investigating into that?

Nick-Mazuk:
> OK, so it seems (!) we'd then need to adjust those utest to the changed spacing. @Nick-Mazuk are you investigating into that?

Hoping to get to that later today. If not, tomorrow.

MarcSabatella:
BTW, in Discord I mentioned my trial PR https://github.com/musescore/MuseScore/pull/6312 as something also worth revisiting in this context (like seeing how well it still applies to MuseScore 4; my code was only for 3).  It's a pretty simple change and in a different part of the code from this PR, but I think they kind of go hand in hand in terms of both addressing issues with inconsistent spacing.  This PR is about spacing caused by bad decisions about stretch; my PR is about allocating too much space between notes in the first place, in cases where notes overlap between staves.

Another relevant PR to think about is https://github.com/musescore/MuseScore/pull/5840.  This is about accidentals not tucking under/over earlier notes as was actually originally implemented by Werner for MuseScore 3 many years ago but was backed out before release.  As far as I can tell, the only issue preventing it from working was that we weren't sure about stem direction until too late, because beaming code didn't decide on stem directions until later on.  I have no idea if the new beaming code changes this, but anyhow, this seemed as good a place as any to point out that there is some existing work that could possibly be leveraged and to make sure the people working on engraving improvements know about it.  I am not in a position to take this back up right now myself.

Jojo-Schmitz:
Still code style issues, 2 are about trailing space, 2 others about the bracing of an `else`?

Nick-Mazuk:
So I've gone through the unit tests in more detail, and it was definitely the beams causing the tests to fail. Also, my theory was correct and the different spacing caused the vertical positions of the beams to change. Finally, I can confirm that all the beams look good.

@mike-spa, you'll need to update the test files to make sure all the tests pass. Unfortunately, I cannot update the PR to do this for you. Here's how you can do this.

First, you'll need to look through the test logs to find several sections that look like this. This will tell you exactly what you need to update.

```
7: --- copypastesymbollist-lyrics.mscx	2021-11-30 13:24:13.342782365 +0000
7: +++ /home/runner/work/MuseScore/MuseScore/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx	2021-11-30 13:06:57.277409640 +0000
7: @@ -287,7 +287,7 @@
7:        <Measure>
7:          <voice>
7:            <Beam>
7: -            <l1>30</l1>
7: +            <l1>24</l1>
7:              <l2>32</l2>
7:              </Beam>
7:            <Chord>
7:    <diff -u copypastesymbollist-lyrics.mscx 
```

There are three things to note:

1. The path to the file, in this case it's `src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx`.
2. The line number which this diff starts at, in this case it's 287.
3. The actual differences between the output and what's expected. In this case, `24` was the expected out, but instead the beam height was `30`.

So for this case, you'll just need to replace the `24` with the `30` on line 290.

In total, there are 8 files that need to be changed:

- Beam-A.mscx
- Beam-B.mscx
- Beam-C.mscx
- Beam-D.mscx
- Beam-E.mscx
- Beam-F.mscx
- Beam-G.mscx
- copypastesymbollist-lyrics-ref.mscx

Jojo-Schmitz:
@Nick-Mazuk: You could though create a PR on top of @mike-spa's branch for this PR, which he'd then just need to merge.

Jojo-Schmitz:
BTW: These unit tests seem to be the reversed, `diff new old` rather than `diff old new`, resulting in a somewhat confusing output.

mike-spa:
@Nick-Mazuk 
Just for my curiosity/understanding: we currently have the stem direction bug (I experienced stems randomly flipping even before touching any code). Isn't that itself an obvious cause for the beam test to fail?

Nick-Mazuk:
> @Nick-Mazuk Just for my curiosity/understanding: we currently have the stem direction bug (I experienced stems randomly flipping even before touching any code). Isn't that itself an obvious cause for the beam test to fail?

Would you mind creating an issue about that with more detail and potentially an example or two?

https://github.com/musescore/MuseScore/issues/new?assignees=&labels=&template=-mu4--development-issue.md&title=%5BMU4+Issue%5D

mike-spa:
> 

Aha! Yes, the flips I experienced were all about the middle line. Maybe it's just that I don't understand the rule behind the flips, but they did feel quite random. I'll crete the issue later, and if the behaviour is as intended I'll remove it.

Meanwhile, do you still want me to modify the utest file? Because my concern was: if I change the utest file to pass the test, but the test failure was caused by a possible bug, then I'm basically adapting the test to the bug, which wouldn't be good, right?

Nick-Mazuk:
> > 
> 
> Aha! Yes, the flips I experienced were all about the middle line. Maybe it's just that I don't understand the rule behind the flips, but they did feel quite random. I'll crete the issue later, and if the behaviour is as intended I'll remove it.
> 
> Meanwhile, do you still want me to modify the utest file? Because my concern was: if I change the utest file to pass the test, but the test failure was caused by a possible bug, then I'm basically adapting the test to the bug, which wouldn't be good, right?

Yes, go ahead and fix them. That flipping shouldn't affect any of the tests.

Saw your issue https://github.com/musescore/MuseScore/issues/9959, and yeah, the behavior is buggy and confusing. So I created PR https://github.com/musescore/MuseScore/pull/9969 to disable that feature. Not sure if we're going to fix it or scrap it altogether (probably the latter). When creating this PR, though, I didn't have to modify any of the tests, so the utests here shouldn't be failing because of the context flipping. Anyways, that flipping was not based on the horizontal spacing whatsoever.

src/engraving/layout/layoutsystem.cpp
```diff
@@ -159,6 +187,19 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 system->removeLastMeasure();
                 ctx.curMeasure->setParent(oldSystem);
             }
+            // If the last appended measure caused a re-layout of the previous measures, now that we are
+            // removing it we need to re-layout the previous measures again.
```
Jojo-Schmitz, line 191: the trailing space it causeing the code style check to fail (yes, it is pretty picky)

Jojo-Schmitz:
That merge is the wrong thing to do. Instead try `git pull --rebase upstream master`, followed by `git push -f`

Jojo-Schmitz:
Here's the needed changes to get the codestyle check and the unit tests pass (at least I believe this be the case):
```
diff --git a/src/engraving/layout/layoutsystem.cpp b/src/engraving/layout/layoutsystem.cpp
index 01375621b9..83741bd78c 100644
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -188,7 +188,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 ctx.curMeasure->setParent(oldSystem);
             }
             // If the last appended measure caused a re-layout of the previous measures, now that we are
-            // removing it we need to re-layout the previous measures again.
+            // removing it we need to re-layout the previous measures again.
             if (changeMinSysTicks) {
                 minTicks = prevMinTicks; // If the last measure caused it to change, now we need to restore it!
                 for (MeasureBase* mb : system->measures()) {
diff --git a/src/engraving/utests/beam_data/Beam-A.mscx b/src/engraving/utests/beam_data/Beam-A.mscx
index c9a4d48d0f..2764f04de5 100644
--- a/src/engraving/utests/beam_data/Beam-A.mscx
+++ b/src/engraving/utests/beam_data/Beam-A.mscx
@@ -712,8 +712,8 @@
               </Note>
             </Chord>
           <Beam>
-            <l1>38</l1>
-            <l2>40</l2>
+            <l1>32</l1>
+            <l2>38</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
diff --git a/src/engraving/utests/beam_data/Beam-B.mscx b/src/engraving/utests/beam_data/Beam-B.mscx
index 40215d7c82..37924672f6 100644
--- a/src/engraving/utests/beam_data/Beam-B.mscx
+++ b/src/engraving/utests/beam_data/Beam-B.mscx
@@ -297,8 +297,8 @@
             </Chord>
           <Beam>
             <StemDirection>up</StemDirection>
-            <l1>-6</l1>
-            <l2>-8</l2>
+            <l1>0</l1>
+            <l2>-6</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
@@ -1123,8 +1123,8 @@
             </Chord>
           <Beam>
             <StemDirection>down</StemDirection>
-            <l1>38</l1>
-            <l2>40</l2>
+            <l1>32</l1>
+            <l2>38</l2>
             </Beam>
           <Chord>
             <durationType>eighth</durationType>
diff --git a/src/engraving/utests/beam_data/Beam-C.mscx b/src/engraving/utests/beam_data/Beam-C.mscx
index 2f2ee3f5db..3631bf8149 100644
--- a/src/engraving/utests/beam_data/Beam-C.mscx
+++ b/src/engraving/utests/beam_data/Beam-C.mscx
@@ -255,8 +255,8 @@
               </Note>
             </Chord>
           <Beam>
-            <l1>-6</l1>
-            <l2>-8</l2>
+            <l1>0</l1>
+            <l2>-6</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
diff --git a/src/engraving/utests/beam_data/Beam-D.mscx b/src/engraving/utests/beam_data/Beam-D.mscx
index bce45604f8..73f9a3488b 100644
--- a/src/engraving/utests/beam_data/Beam-D.mscx
+++ b/src/engraving/utests/beam_data/Beam-D.mscx
@@ -814,7 +814,7 @@
         <voice>
           <Beam>
             <StemDirection>down</StemDirection>
-            <l1>44</l1>
+            <l1>42</l1>
             <l2>48</l2>
             </Beam>
           <Chord>
diff --git a/src/engraving/utests/beam_data/Beam-E.mscx b/src/engraving/utests/beam_data/Beam-E.mscx
index 706ec306ea..ca7fc12b93 100644
--- a/src/engraving/utests/beam_data/Beam-E.mscx
+++ b/src/engraving/utests/beam_data/Beam-E.mscx
@@ -990,7 +990,7 @@
         <voice>
           <Beam>
             <StemDirection>down</StemDirection>
-            <l1>52</l1>
+            <l1>46</l1>
             <l2>48</l2>
             </Beam>
           <Chord>
@@ -1010,7 +1010,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>52</l1>
-            <l2>48</l2>
+            <l2>46</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
@@ -1030,7 +1030,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>52</l1>
-            <l2>48</l2>
+            <l2>46</l2>
             </Beam>
           <Chord>
             <durationType>eighth</durationType>
diff --git a/src/engraving/utests/beam_data/Beam-F.mscx b/src/engraving/utests/beam_data/Beam-F.mscx
index b14e6db9ae..a68818fe6f 100644
--- a/src/engraving/utests/beam_data/Beam-F.mscx
+++ b/src/engraving/utests/beam_data/Beam-F.mscx
@@ -992,7 +992,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>48</l1>
-            <l2>44</l2>
+            <l2>42</l2>
             </Beam>
           <Chord>
             <durationType>eighth</durationType>
@@ -1011,7 +1011,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>48</l1>
-            <l2>44</l2>
+            <l2>42</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
@@ -1031,7 +1031,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>48</l1>
-            <l2>44</l2>
+            <l2>42</l2>
             </Beam>
           <Chord>
             <durationType>eighth</durationType>
diff --git a/src/engraving/utests/beam_data/Beam-G.mscx b/src/engraving/utests/beam_data/Beam-G.mscx
index 14ef4c6b86..9c820f7260 100644
--- a/src/engraving/utests/beam_data/Beam-G.mscx
+++ b/src/engraving/utests/beam_data/Beam-G.mscx
@@ -902,7 +902,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>44</l1>
-            <l2>40</l2>
+            <l2>38</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
@@ -922,7 +922,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>44</l1>
-            <l2>40</l2>
+            <l2>38</l2>
             </Beam>
           <Chord>
             <durationType>eighth</durationType>
@@ -941,7 +941,7 @@
           <Beam>
             <StemDirection>down</StemDirection>
             <l1>44</l1>
-            <l2>40</l2>
+            <l2>38</l2>
             </Beam>
           <Chord>
             <BeamMode>begin</BeamMode>
diff --git a/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx b/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx
index c57624da95..e8359bbdc7 100644
--- a/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx
+++ b/src/engraving/utests/copypastesymbollist_data/copypastesymbollist-lyrics-ref.mscx
@@ -287,7 +287,7 @@
       <Measure>
         <voice>
           <Beam>
-            <l1>24</l1>
+            <l1>30</l1>
             <l2>32</l2>
             </Beam>
           <Chord>
```

src/engraving/libmscore/measure.cpp
```diff
@@ -3160,6 +3160,10 @@ void Measure::stretchMeasure(qreal targetWidth)
 
     std::multimap<qreal, Segment*> springs;
 
+    //Fraction minTicks = computeTicks();
+    Fraction minTicks = minSysTicks(); // The stretching must be done according to the shortest note *of the system*!
+    qreal minNoteSpace = (score()->noteHeadWidth() + score()->styleMM(Sid::minNoteDistance));
```
Jojo-Schmitz, line 3165: This one is not needed, not used anywhere, leads to a compiler warning (in MSCV)

src/engraving/layout/layoutsystem.cpp
```diff
@@ -381,8 +422,8 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
             mb->setPos(pos);
             Measure* m = toMeasure(mb);
             qreal stretch = m->basicStretch();
```
Jojo-Schmitz, line 424: This one is now no longer needed (nor used) and leads to a compiler warning (in MSVC)

src/engraving/libmscore/measure.cpp
```diff
@@ -3160,6 +3160,10 @@ void Measure::stretchMeasure(qreal targetWidth)
 
     std::multimap<qreal, Segment*> springs;
 
+    //Fraction minTicks = computeTicks();
+    Fraction minTicks = minSysTicks(); // The stretching must be done according to the shortest note *of the system*!
```
Jojo-Schmitz, line 3164: This is overwrites the one a few lines up, causes a compiler warning. The commented out line above seems not needed

Jojo-Schmitz:
See [b7b0fac](https://github.com/Jojo-Schmitz/MuseScore/commit/b7b0facf097391f6831d84695be9a1fa04f87834) for a commit fixing all these issue (that one remaining code style isses, the 3 compiler warnings and those 8 unit tests)

Jojo-Schmitz:
Almost there, one last utest, for src/engraving/utests/beam_data/Beam-E.mscx
```
 --- Beam-E.mscx	2021-12-10 09:15:55.025776022 +0000
 +++ /home/runner/work/MuseScore/MuseScore/src/engraving/utests/beam_data/Beam-E.mscx	2021-12-10 08:58:51.526510812 +0000
 @@ -990,8 +990,8 @@
          <voice>
            <Beam>
              <StemDirection>down</StemDirection>
 -            <l1>52</l1>
 -            <l2>46</l2>
 +            <l1>46</l1>
 +            <l2>48</l2>
              </Beam>
            <Chord>
              <durationType>eighth</durationType>
```

worldwideweary:
P.S. Dare you to make a 3.x compatible version for self-compiling folks in-between releases ;-)

mike-spa:
> 

That wouldn't be a bad idea ;) I may do that if I happen to find the time.

Jojo-Schmitz:
It'd be much apreciated. Doesn't see too straight forward though

Jojo-Schmitz:
and we're only a tiny fix for the utests away from that as far as I can tell.

mike-spa:
Pushed some new code. To be revised by @oktophonie. More utest failures are possible/expected.

src/engraving/layout/layoutsystem.cpp
```diff
@@ -322,52 +366,42 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
...
-            } else {
-                rest = minWidth;
-            }
-        }
-        rest /= totalWeight;
+        newRest = systemWidth - minWidth;
+        iter++;
     }
-
+
```
Jojo-Schmitz, line 403: A stray tab here ? Code style issue

src/engraving/libmscore/measure.cpp
```diff
@@ -4115,12 +4077,71 @@ static bool hasAccidental(Segment* s)
...
+//      Stretch formula
+//      Computes the stretch for a given segment depending
+//      on its duration with respect to the shortest one.
+//      Three different options proposed (see documentation).
+//---------------------------------------------------------
+qreal Measure::stretchFormula(Fraction curTicks, Fraction minTicks, qreal stretchCoeff)
+{
+    qreal slope = userStretch() * score()->styleD(Sid::measureSpacing);
+
+    if (minTicks <= Fraction(1 , 32)) {
```
Jojo-Schmitz, line 4110: A space too much before the comma, code style issue

src/engraving/libmscore/measure.cpp
```diff
@@ -4173,6 +4183,15 @@ void Measure::computeMinWidth(Segment* s, qreal x, bool isSystemHeader)
                 isSystemHeader = false;
             } else {
                 w = s->minHorizontalDistance(ns, false);
+                // New spacing algorithm: we apply an additional spacing which depends on the duration of
+                // the note with respect to the shortest note *of the system*.
+                Fraction t = s->ticks();
+                if (t > m_timesig) { // Avoids long MM rests being super wide
+                    t = 2 * m_timesig;
```
Jojo-Schmitz, line 4190: Trailing space, code style issue

Jojo-Schmitz:
Indeed 2 more unit tests that are failing, in Beam-D.mscx and test7.cap-ref.mscx, in addition to the one in Beam-E.mscx mentioned further up.

Jojo-Schmitz:
The vtests differences are huge, due to a recent change in the master branch, A Rebase of this PR should reduce them to the ones we expect.

Your change to `minNoteDFistance` broke more (9) unit tests  and there's 2 others about beams.

Jojo-Schmitz:
Only one failing unit test left, for src/importexport/capella/tests/data/test7.cap-ref.mscx and its beam

Jojo-Schmitz:
All tests passed, vtest differences are still huge, but as far as I can tell none of them are unexpected, all due to the different (and better) spacing.

As far as I can tell this is ready to get merged, @oktophonie ?

src/engraving/libmscore/measure.cpp
```diff
@@ -4083,8 +4083,7 @@ static bool hasAccidental(Segment* s)
 
 Fraction Measure::minSysTicks()
 {
-    Fraction minTicks = Fraction(1, 8);
-    //System* sys = system();
+    Fraction minTicks = Fraction(10, 1);
```
Jojo-Schmitz, line 4086: That's pretty larg for a minimum, 10 whole notes?

mike-spa: That is just the initialization of the variable. It needs to start a random high value in order for the function to correcly return the shortest note. The 1/8 was a mistake.

Jojo-Schmitz: Might be worth a comment in the code

Jojo-Schmitz: And the largest possible fraction is a longa `fraction(4, 1) or `DurationType::maxtype.ticks()` (?), I believe, seems a good starting point?

Jojo-Schmitz: Hmm, that could be quatdouple dotted, so even longer, makes it 39/4 is my math isn't failing me?

Jojo-Schmitz: Surprisingly close to your 10/1 ;-)

Jojo-Schmitz: Actually my math was failing me (as usual :-( ): it is 29/4.

mike-spa: That starting point has *no effect whatsoever*. It just needs to be larger than any note you may encounter. It could just as well be 100 whole notes.

Jojo-Schmitz: True, still worth a comment to explain that IMHO

oktophonie (approved):

MarcSabatella:
I just wanted to bump this and make sure it's not being lost in the shuffle.  To me this would be one of the most important improvements MsueScore 4 could offer.

src/engraving/libmscore/measure.cpp
```diff
@@ -4115,12 +4077,77 @@ static bool hasAccidental(Segment* s)
...
+        qreal reduction = qMax((1 - 0.2 * log2(qreal(Fraction(1, 16).ticks()) / qreal(minTicks.ticks()))), 0.3);
+        slope = slope * reduction;
+    }
+
+    // Linear spacing
+    //qreal str = 1 - 0.112 * slope +  0.112 * slope * qreal(curTicks.ticks()) / qreal(minTicks.ticks());
+    // Logarithmic spacing (MS 3.6)
+    //qreal str = 1.0 + 0.721 * slope * log(qreal(curTicks.ticks()) / qreal(minTicks.ticks()));
+    // Quadratic spacing (Gould)
+    qreal str = 1 - 0.647 * slope + 0.647 * slope * sqrt(qreal(curTicks.ticks()) / qreal(minTicks.ticks()));
```
vpereverzev, line 4128: avoid the usage of 'magic numbers'. I suggest to declare a named variable here or at the top of this file, like:

static constexpr float STRETCH_FACTOR = 0.647;

src/engraving/libmscore/measure.cpp
```diff
@@ -4115,12 +4077,77 @@ static bool hasAccidental(Segment* s)
...
+
+    if (minTicks > Fraction(1, 16)) {
+        // Avoids long notes being too narrow in the absense of shorter notes.
+        str = str * (1 - 0.5 + 0.5 * sqrt(qreal(minTicks.ticks()) / qreal(Fraction(1, 16).ticks())));
+    }
+
+    str *= stretchCoeff;
```
vpereverzev, line 4135: I don't see any reason to receive this 'stretchCoeff' from outside since it always true and contradict with expected purpose of this method.

I suggest to:
- rename stretchFormula to 'float stretchFactor(const Fraction availableTicks, const Fraction minTicks) const'

src/engraving/libmscore/measure.cpp
```diff
@@ -4115,12 +4077,77 @@ static bool hasAccidental(Segment* s)
...
+    qreal slope = userStretch() * score()->styleD(Sid::measureSpacing);
+
+    if (curTicks > m_timesig) { // This is the case of MM rests
+        curTicks = curTicks - m_timesig; // 2 bars MM rests is spaced same as 1 bar
+        if (curTicks > m_timesig * 20) { // To avoid excessive
+            curTicks = m_timesig * 20;
+        }
+    }
+
+    if (minTicks <= Fraction(1, 32)) {
```
vpereverzev, line 4116: initialize Fraction(1,32) and Fraction(1, 16) at the beginning of the method, there is no need to construct them every time here and below

src/engraving/libmscore/measure.cpp
vpereverzev: the same here:
- make a named constexpr variable, make sure that you put a comment explaining these numbers

src/engraving/libmscore/measure.cpp
```diff
@@ -4115,12 +4077,77 @@ static bool hasAccidental(Segment* s)
...
-//    sets the minimum stretched width of segment list s
-//    set the width and x position for all segments
+//      minSysTicks
+//      returns the shortest note/rest in the system
+//---------------------------------------------------------
+
+Fraction Measure::minSysTicks()
+{
+    Fraction minTicks = Fraction(10, 1);
+    for (MeasureBase* mb : system()->measures()) {
```
vpereverzev, line 4087: if you're looping through the measures of the system here, I don't understand the reason why the methods is here.

I believe this method should be moved to the System class then

src/engraving/libmscore/measure.cpp
```diff
@@ -4231,10 +4255,33 @@ void Measure::computeMinWidth(Segment* s, qreal x, bool isSystemHeader)
         x += w;
         s = s->next();
     }
+
+    // This doesn't seem to be needed but I'm not sure yet.
```
vpereverzev, line 4259: then just delete it. If necessary you can always return this from git history

src/engraving/libmscore/measure.cpp
```diff
@@ -4115,12 +4077,77 @@ static bool hasAccidental(Segment* s)
...
+//      on its duration with respect to the shortest one.
+//      Three different options proposed (see documentation).
+//---------------------------------------------------------
+qreal Measure::stretchFormula(Fraction curTicks, Fraction minTicks, qreal stretchCoeff)
+{
+    qreal slope = userStretch() * score()->styleD(Sid::measureSpacing);
+
+    if (curTicks > m_timesig) { // This is the case of MM rests
+        curTicks = curTicks - m_timesig; // 2 bars MM rests is spaced same as 1 bar
+        if (curTicks > m_timesig * 20) { // To avoid excessive
```
vpereverzev, line 4111: it's not clear what 20 does here, probably requires a comment or, again, named constant variable

src/engraving/libmscore/measure.cpp
```diff
@@ -4074,6 +4033,9 @@ void Measure::setStretchedWidth(qreal w)
         w = minWidth;
     }
 
+    // New spacing algorithm: this stretching is not necessary anymore.
+
+#if 0
```
vpereverzev, line 4038: never use that

If there is something redundant, then just remove it, otherwise the code will turn into a garbage

src/engraving/libmscore/measure.cpp
```diff
@@ -3660,6 +3591,34 @@ qreal Measure::basicWidth() const
...
+//   Returns a weight parameter proportional to the
+//   current with of the measure.
+//---------------------------------------------------------
+qreal Measure::stretchWeight()
+{
+    if (isMMRest()) {
+        return width();
+    }
+    qreal w = 0;
+    qreal W = 0;
```
vpereverzev, line 3605: double declaration of **w**

src/engraving/libmscore/measure.cpp
```diff
@@ -3660,6 +3591,34 @@ qreal Measure::basicWidth() const
...
+    qreal weight = 0;
+    qreal minNoteSpace = score()->noteHeadWidth() * 1.05;
```
vpereverzev, line 3607: magic number - 1.05

src/engraving/libmscore/measure.cpp
```diff
@@ -3660,6 +3591,34 @@ qreal Measure::basicWidth() const
...
+    for (Segment* s = first(); s; s = s->next()) {
+        if (s->enabled() && s->visible() && !s->allElementsInvisible() && s->isChordRestType()) {
+            qreal str = stretchFormula(s->ticks(), minTicks, 1);
+            w += minNoteSpace * str * basicStretch();
+            W += s->width();
+            // You cannot simply do weight += s->width(), because the s->width() also includes accidentals
+        }
+        continue;
+    }
+    weight = w * (w / W); // Empirical correction factor (see documentation)
```
vpereverzev, line 3618: which documentation? You're talking about PDF?  If so, then put a relevant comment here with an explanation of what the 'Empirical correction factor' is

src/engraving/libmscore/measure.cpp
Jojo-Schmitz: Upper case W and lower case w...

src/engraving/libmscore/system.cpp
```diff
@@ -1824,4 +1824,24 @@ int System::lastVisibleSysStaffOfPart(const Part* part) const
...
 }
+
+//---------------------------------------------------------
+//      minSysTicks
+//      returns the shortest note/rest in the system
+//---------------------------------------------------------
+
+Fraction System::minSysTicks() const
+{
+    Fraction minTicks = Fraction(10, 1); // Initializing the variable at a random high value.
```
vpereverzev, line 1835: still no explanation here

other developers need to understand what you think (10, 1) is sufficient number here

src/engraving/libmscore/system.cpp
```diff
@@ -1824,4 +1824,24 @@ int System::lastVisibleSysStaffOfPart(const Part* part) const
...
+    for (MeasureBase* mb : measures()) {
+        if (mb->isMeasure()) {
+            Measure* m = toMeasure(mb);
+            Fraction curMinTicks = m->computeTicks();
+            if (curMinTicks < minTicks) {
```
vpereverzev, line 1840: minTicks = std::min(m->computeTicks(), minTicks)

src/engraving/libmscore/measure.cpp
```diff
@@ -4170,6 +4110,17 @@ void Measure::computeMinWidth(Segment* s, qreal x, bool isSystemHeader)
...
+                // New spacing algorithm: we apply an additional spacing which depends on the duration of
+                // the note with respect to the shortest note *of the system*.
+                if (s->isChordRestType()) {
+                    Fraction t = s->ticks();
+                    qreal durStretch = durationStretch(t, minTicks);
+                    qreal minStretchedWidth = minNoteSpace * durStretch * usrStretch * stretchCoeff;
+                    // NOTE: durStretch is the spacing factor purely determined by the duration of the note.
+                    // usrStretch is the spacing factor determined by user settings.
+                    // stretchCoeff is the spacing factor used to justify the systems, i.e. getting the systems to fill the page.
+                    w = qMax(w, minStretchedWidth);
```
vpereverzev, line 4122: prefer to use STL instead of Qt at libmscore classes

vpereverzev: std::max in this particular case

src/engraving/libmscore/measure.cpp
```diff
@@ -4126,19 +4077,8 @@ void Measure::computeMinWidth(Segment* s, qreal x, bool isSystemHeader)
...
-            MMRest* mmRest = toMMRest(seg->element(staff2track(st)));
-            if (mmRest) {
-                mmRest->rxpos() = 0;
-                mmRest->setWidth(score()->styleMM(Sid::minMMRestWidth) * mag());
-                mmRest->segment()->createShapes();
-            }
-        }
-    }
+    qreal minNoteSpace = score()->noteHeadWidth() * 1.05;
+    qreal usrStretch = qMax(userStretch() * score()->styleD(Sid::measureSpacing), qreal(0.1)); // The qMax avoids stretch going to zero
```
vpereverzev, line 4081: std::max

src/engraving/libmscore/measure.cpp
```diff
@@ -4112,12 +4008,67 @@ static bool hasAccidental(Segment* s)
 }
 
 //---------------------------------------------------------
-//   computeMinWidth
-//    sets the minimum stretched width of segment list s
-//    set the width and x position for all segments
+//      Stretch formula
```
vpereverzev, line 4011: don't forget to update method's description

src/engraving/libmscore/measure.cpp
```diff
@@ -4112,12 +4008,67 @@ static bool hasAccidental(Segment* s)
...
+//      Computes the stretch for a given segment depending
+//      on its duration with respect to the shortest one.
+//      Three different options proposed (see documentation).
 //---------------------------------------------------------
+float Measure::durationStretch(Fraction curTicks, const Fraction minTicks) const
+{
+    qreal userSlope = userStretch() * score()->styleD(Sid::measureSpacing);
+    static constexpr qreal baseSlope = 0.647;
```
vpereverzev, line 4019: explanation comment needed

stick to this rule every time you want to write hardcoded numbers

src/engraving/libmscore/measure.cpp
```diff
@@ -3140,85 +3140,16 @@ QString Measure::accessibleInfo() const
...
-            s = s->next();
-        }
-        qreal x = s ? s->pos().x() : 0.0;
-        while (s) {
-            s->rxpos() = x;
-            x += s->width();
-            s = s->nextEnabled();
-        }
-    }
+    //bbox().setWidth(targetWidth); // Doesn't seem to be needed but I keep it here just to be safe
```
vpereverzev, line 3152: as I said previously

always prefer to remove the code if you think it's redundant, if necessary we can recover logic from git history

src/engraving/layout/layoutsystem.cpp
```diff
@@ -322,52 +366,42 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
...
+        // We do not stretch last system if curSysWidth is <= lastSystemFillLimit
+        newRest = 0;
+    }
+    if (MScore::noHorizontalStretch) { // Debug feature
+        newRest = 0;
+    }
+    qreal stretchCoeff = 1;
+    qreal prevWidth = 0;
+    int iter = 0;
+    while (abs(newRest) > score->spatium() * 0.08 && iter < 200) { // spatium*0.08 is about the width of a note stem
```
vpereverzev, line 390: looks like if I'll change a width of a stem in "Properties" panel or in style we'll not layout this properly?

makes sense to use the value from the current style or make a constant variable at stem.cpp

src/engraving/libmscore/system.cpp
mike-spa: Wait, but the variable changes its value inside this method. And it needs to start at a high value every time the method is called. So I cannot declare it static constexpr. Or am I missing something?

Jojo-Schmitz: The longest note/rest MuseScore can possibly create is a quadruple dotted longa, that is one quarter short of 8/1.
So `Fraction(8, 1)` is more than enough already, no need to go to `Fraction (INT_MAX, 1)`.
Well, maybe, one day, we add suppoer for a maxima (8/1), which quadrouple dotted would be 15.5/1 , so `Fraction(16/1)` should be on the safest possible side ;-)

src/engraving/libmscore/measure.cpp
mike-spa: This one is explained in the longer comment that follows

src/engraving/libmscore/system.cpp
mike-spa: I have the feeling that we may be overthinking this. The fraction really just needs to be longer than any possible note/rest, but "how much longer" does not make any difference, neither conceptually nor computationally. Choosing Fraction (10, 1) was a bad idea because it looks like a "reasonable" value, so a reader would assume that it is supposed to "mean something". 
I should have started with Fraction(1000, 1) or something like that, so it's immediately obvious that it's a random value. This is what I'm doing now. And I'm adding a more verbose comment. We should be fine :)

src/engraving/layout/layoutsystem.cpp
mike-spa: Good one, thanks! :)

src/engraving/libmscore/measure.cpp
mike-spa: Sorry, forgot that one

src/engraving/libmscore/system.cpp
Jojo-Schmitz: Well, 1000 is a much a 'magic number' as 10. Or 16. The latter at least makes sense.

MarcSabatella:
FWIW, we use that same trick of choosing an arbitrary large default value a number of other places n the code as well, and my recollection is we usually use "10000" (ten thousand).  At least, for other values in a similar basic range, like for MIDI pitch numbers of staff line numbers or offsets in sp units.

vpereverzev:
> FWIW, we use that same trick of choosing an arbitrary large default value a number of other places n the code as well, and my recollection is we usually use "10000" (ten thousand). At least, for other values in a similar basic range, like for MIDI pitch numbers of staff line numbers or offsets in sp units.

@mike-spa 
I encourage to make `static constexpr Fraction maxFraction(10000, 1)` in constants.h and make sure that you're using this value in your calculations

mike-spa:
> > FWIW, we use that same trick of choosing an arbitrary large default value a number of other places n the code as well, and my recollection is we usually use "10000" (ten thousand). At least, for other values in a similar basic range, like for MIDI pitch numbers of staff line numbers or offsets in sp units.
> 
> @mike-spa I encourage to make `static constexpr Fraction maxFraction(10000, 1)` in constants.h and make sure that you're using this value in your calculations

The compiler doesn't let me do that. I guess it's because I need to do include Fraction.h into constants.h if I want to declare a Fraction, but Fraction.h also includes constant.h at its beginning. Or am I missing something?

vpereverzev:
> > > FWIW, we use that same trick of choosing an arbitrary large default value a number of other places n the code as well, and my recollection is we usually use "10000" (ten thousand). At least, for other values in a similar basic range, like for MIDI pitch numbers of staff line numbers or offsets in sp units.
> > 
> > 
> > @mike-spa I encourage to make `static constexpr Fraction maxFraction(10000, 1)` in constants.h and make sure that you're using this value in your calculations
> 
> The compiler doesn't let me do that. I guess it's because I need to do include Fraction.h into constants.h if I want to declare a Fraction, but Fraction.h also includes constant.h at its beginning. Or am I missing something?

you can also add this variable in fraction.h I think

class Fraction {
...
public:
    static constexpr Fraction MAX(10000, 1);
....

mike-spa:
@vpereverzev Thank for your hints, I now need 

> > > > FWIW, we use that same trick of choosing an arbitrary large default value a number of other places n the code as well, and my recollection is we usually use "10000" (ten thousand). At least, for other values in a similar basic range, like for MIDI pitch numbers of staff line numbers or offsets in sp units.
> > > 
> > > 
> > > @mike-spa I encourage to make `static constexpr Fraction maxFraction(10000, 1)` in constants.h and make sure that you're using this value in your calculations
> > 
> > 
> > The compiler doesn't let me do that. I guess it's because I need to do include Fraction.h into constants.h if I want to declare a Fraction, but Fraction.h also includes constant.h at its beginning. Or am I missing something?
> 
> you can also add this variable in fraction.h I think
> 
> class Fraction { ... public: static constexpr Fraction MAX(10000, 1); ....

@vpereverzev thank you, I'm gonna try that tomorrow. For now, I've pushed some new code with all the other fixes you suggested.

src/engraving/libmscore/system.cpp
```diff
@@ -1832,14 +1832,13 @@ int System::lastVisibleSysStaffOfPart(const Part* part) const
 
 Fraction System::minSysTicks() const
 {
-    Fraction minTicks = Fraction(10, 1); // Initializing the variable at a random high value.
+    Fraction minTicks = Fraction (10000, 1); // Initializing the variable at a random high value.
```
Jojo-Schmitz, line 1835: There's a space too much fort the code style check

MarcSabatella:
Fantastic!  Congratulations on the great work, @mike-spa!

PR #9934, opened 2021-11-30
respect beamNoSlope style parameter

Nick-Mazuk:
PR https://github.com/musescore/MuseScore/pull/9683 updated the style preferences for beams for MS4.

This PR adds support for the `beamNoSlope` style parameter with the new beam code implemented in https://github.com/musescore/MuseScore/pull/9792.


Asmatzaile:
Hi,
Not sure whether this is the right place to bring this up, but regarding the slope parameter of various score elements, the naming of its toggle function is inconsistent.
For beams, it's called "force horizontal", whereas for hairpins (and also for lines in MS 3.6, not sure why the option is no longer in the inspector in MS 4.0), it's called "allow diagonal".

oktophonie (approved):

Nick-Mazuk:
> Hi, Not sure whether this is the right place to bring this up, but regarding the slope parameter of various score elements, the naming of its toggle function is inconsistent. For beams, it's called "force horizontal", whereas for hairpins (and also for lines in MS 3.6, not sure why the option is no longer in the inspector in MS 4.0), it's called "allow diagonal".

So I was not apart of the decision on what to name each of these, but I do see how it might be confusing that they're opposites.

However, I think the rationale is that by default (and standard conventions), beams are slanted and hairpins are flat. And so each of these parameters essentially changes the behavior of MuseScore away from the default.

Asmatzaile:
> So I was not apart of the decision on what to name each of these, but I do see how it might be confusing that they're opposites.
> 
> However, I think the rationale is that by default (and standard conventions), beams are slanted and hairpins are flat. And so each of these parameters essentially changes the behavior of MuseScore away from the default.

I can understand how each parameter changes the behavior of MuseScore away from the default. Yet, in other parts of the application, the state of 'checked' has little to do with the default (e. g., in the general part of 'notes', 'cue size' is unchecked while 'visible' is checked). So I do find the different namings confusing.

RomanPudashkin (approved):

Issue #9634, opened 2021-11-02
[MU4 Issue] Mixer needs to have reduced vertical spacing


jessjwilliamson:
**Describe the bug**
The build has got some added space compared to the Figma version. I've done side by side comparison with pixel numbers labeled.

**Screenshots**

![Screenshot 2021-11-02 131609](https://user-images.githubusercontent.com/71656189/139854356-013cdca9-2830-4069-8910-bca474afcbd2.png)

**Desktop (please complete the following information):**
 - OS: Mac, Windows

**Additional context**
Please go to MIXER page on Figma for exact sizes

Issue #9246, opened 2021-09-23
[MU4 Issue] Mixer - Volume and pan functionality improvements.


jessjwilliamson:
Sound on for me talking over.
This is a video advising solutions for easier functionality for the volume fader and pan knob.
Using Ableton has reference.
I forgot to mention that double-clicking the Pan knob can reset it back to center/zero.

https://user-images.githubusercontent.com/71656189/134523114-259e42a4-b155-4976-88a9-67117d8253ba.mov

Issue #9149, opened 2021-09-16
[MU4 Issue] No strokes between Mixer faders


jessjwilliamson:
**Describe bug**
Currently, there are no strokes in between faders in Mixer. Red arrows on the screenshot shows where there are missing strokes.

**Expected behavior**
<img width="1531" alt="Screenshot 2021-09-16 at 10 32 48" src="https://user-images.githubusercontent.com/71656189/133588966-8cb7aae4-3016-43af-8f15-7c197216fef9.png">

**Screenshots**
![image](https://user-images.githubusercontent.com/71656189/133589183-723d48a9-b8c9-4b11-911f-4a1a1afc2404.png)

**Desktop (please complete the following information):**
 - OS: Mac OS

PR #9935, opened 2021-11-30
Mixer - Add separators, fine-tune spacing

cbjeukendrup:
Resolves: #9149
Resolves: #9634
Resolves: #9246

- Added separator lines between channel strips in Mixer
- Fine-tuned the spacing
- You can now also double-click the handle of the volume slider to reset the volume
- To change the value of the pan knob, you now need to drag up and down instead of drawing circles with the mouse

### DIFF
--- a/src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
+++ b/src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
@@ -165,6 +165,7 @@ Dial {
     }
 
     MouseArea {
+        id: mouseArea
         anchors.fill: parent
         onDoubleClicked: {
             root.newValueRequested(0)
@@ -196,6 +197,14 @@ Dial {
             let newValue = initialValue + dist * sgn
             let bounded = Math.max(root.from, Math.min(newValue, root.to))
             root.newValueRequested(bounded)
+        }
+
+        // We also listen for wheel events here, but for a different reason:
+        // Qml Dial has a bug that it doesn't emit moved() when the value is changed through a wheel event.
+        // So when we see a wheel event, we let the dial handle it, but we will account for emitting the signal.
+        onWheel: function(wheel) {
+            wheel.accepted = false
+            Qt.callLater(function() { root.newValueRequested(Math.round(value)) })
         }
     }
 }
--- a/src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
+++ b/src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
@@ -19,7 +19,6 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 
@@ -33,11 +32,11 @@ Dial {
 
     property alias navigation: navCtrl
 
-    width: prv.radius * 2
-    height: width
+    implicitWidth: prv.radius * 2
+    implicitHeight: implicitWidth
 
-    implicitHeight: height
-    implicitWidth: width
+    width: implicitWidth
+    height: width
 
     wheelEnabled: true
 
@@ -121,18 +120,18 @@ Dial {
 
             ctx.strokeStyle = prv.outerArcColor
             ctx.beginPath()
-            ctx.arc(width/2, height/2, prv.radius - prv.outerArcLineWidth, -140 * (Math.PI/180) - Math.PI/2, 140 * (Math.PI/180) - Math.PI/2, false)
+            ctx.arc(width/2, height/2, prv.radius - prv.outerArcLineWidth/2, -140 * (Math.PI/180) - Math.PI/2, 140 * (Math.PI/180) - Math.PI/2, false)
             ctx.stroke()
 
             ctx.strokeStyle = prv.valueArcColor
             ctx.beginPath()
-            ctx.arc(width/2, height/2, prv.radius - prv.outerArcLineWidth, -Math.PI/2, root.angle * (Math.PI/180) - Math.PI/2, prv.reversed)
+            ctx.arc(width/2, height/2, prv.radius - prv.outerArcLineWidth/2, -Math.PI/2, root.angle * (Math.PI/180) - Math.PI/2, prv.reversed)
             ctx.stroke()
 
             ctx.lineWidth = prv.innerArcLineWidth
             ctx.strokeStyle = prv.innerArcColor
             ctx.beginPath()
-            ctx.arc(width/2, height/2, prv.radius - (prv.outerArcLineWidth + prv.innerArcLineWidth), 0, Math.PI * 2, false)
+            ctx.arc(width/2, height/2, prv.radius - (prv.outerArcLineWidth + prv.innerArcLineWidth/2), 0, Math.PI * 2, false)
             ctx.stroke()
         }
     }
--- a/src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
+++ b/src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
@@ -44,6 +44,7 @@ Dial {
     to: root.valueScale
     value: 0
 
+    signal newValueRequested(int newValue)
     signal increaseRequested()
     signal decreaseRequested()
 
@@ -157,5 +158,44 @@ Dial {
 
     Component.onCompleted: {
         backgroundCanvas.requestPaint()
+    }
+
+    onMoved: {
+        newValueRequested(value)
+    }
+
+    MouseArea {
+        anchors.fill: parent
+        onDoubleClicked: {
+            root.newValueRequested(0)
+        }
+
+        // The MouseArea steals mouse press events from the slider.
+        // There is really no way to prevent that.
+        // (if you set mouse.accepted to false in the onPressed handler,
+        // the MouseArea won't receive doubleClick events).
+        // So we have to reimplement the dragging behaviour.
+        // That gives us the opportunity to do it better than Qt.
+        // We will allow both dragging vertically and horizontally.
+
+        preventStealing: true // Don't let a Flickable steal the mouse
+
+        property int initialValue: 0
+        property real dragStartX: 0
+        property real dragStartY: 0
+
+        onPressed: function(mouse) {
+            initialValue = root.value
+            dragStartX = mouse.x; dragStartY = mouse.y
+        }
+
+        onPositionChanged: function(mouse)  {
+            let dx = mouse.x - dragStartX; let dy = mouse.y - dragStartY
+            let dist = Math.sqrt(dx * dx + dy * dy)
+            let sgn = (dy < dx) ? 1 : -1
+            let newValue = initialValue + dist * sgn
+            let bounded = Math.max(root.from, Math.min(newValue, root.to))
+            root.newValueRequested(bounded)
+        }
     }
 }
--- a/src/playback/qml/MuseScore/Playback/MixerPanel.qml
+++ b/src/playback/qml/MuseScore/Playback/MixerPanel.qml
@@ -120,16 +120,15 @@ Rectangle {
 
         Column {
             id: contentColumn
-
             width: childrenRect.width
-
-            spacing: 8
+            spacing: 0
 
             MixerSoundSection {
                 id: soundSection
 
                 visible: contextMenuModel.soundSectionVisible
                 headerVisible: contextMenuModel.labelsSectionVisible
+                spacingAbove: 8
                 rootPanel: root
 
                 model: mixerPanelModel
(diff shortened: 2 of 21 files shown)


Asmatzaile:
Hi,
I don't know about Ableton but some DAWs, such as Reaper, allow left-right scrolling in addition to down-up scrolling to set the value of the pan knob* (I pan scrolling in the former way because it feels more natural to me).
I can't get to see the mixer in the Windows artifacts generated by this PR, but by looking at the code it seems that it only accepts the mouse y value as input. Could you make it so the mouse's displacement in the x axis is also taken into in account when setting the knob's value?
Many thanks

*and knobs in general

jessjwilliamson:
> Hi, I don't know about Ableton but some DAWs, such as Reaper, allow left-right scrolling in addition to down-up scrolling to set the value of the pan knob* (I pan scrolling in the former way because it feels more natural to me). I can't get to see the mixer in the Windows artifacts generated by this PR, but by looking at the code it seems that it only accepts the mouse y value as input. Could you make it so the mouse's displacement in the x axis is also taken into in account when setting the knob's value? Many thanks
> 
> *and knobs in general

Hi! It may not have been clear enough from my end, but the Y value was suppose to be an ADDITION to the already implemented value, so both types of panning was the desired way. Can see on my [Ableton demonstration](https://github.com/musescore/MuseScore/issues/9246) that it does both - going around the edge AS WELL as the up and down motion :)

jessjwilliamson:
Just a couple notes:
- I mentioned on a reply above about the panning, can it be possible to still keep the original function and have the up and down as well? See [Ableton for demonstration](https://github.com/musescore/MuseScore/issues/9246) 

> Hi, I don't know about Ableton but some DAWs, such as Reaper, allow left-right scrolling in addition to down-up scrolling to set the value of the pan knob* (I pan scrolling in the former way because it feels more natural to me). I can't get to see the mixer in the Windows artifacts generated by this PR, but by looking at the code it seems that it only accepts the mouse y value as input. Could you make it so the mouse's displacement in the x axis is also taken into in account when setting the knob's value? Many thanks
> 
> *and knobs in general

Hi, Sorry I made a comment then deleted it because I was mistaken - I thought Ableton does both Y  value and also around, but it looks like it's just the Y value. Just checked Ableton, Reaper and Logic, they both have this Y value, they don't really have a way to go 'around' and would prefer to keep it that way as it is the standard now.

jessjwilliamson:
@cbjeukendrup 
Hi! Thank you so much for implementing these! just a couple of things:
1. Can the panning also go back to 0 value when double-clicking on the pan knob as well please! Just like the volume fader, just panning as well :D 
2. Unfortunately, the strokes do not go all the way to the top when resizing the whole Mixer window vertically. Would be great that the strokes can go all the way up no matter how tall the Mixer window is
image
3. When adding Audio FX, there are missing stroke gaps on channels where there are no FX
Screenshot 2021-12-01 115201
That is all!

DmitryArefiev:
Hi @cbjeukendrup , I think volume fader shouldn't behave like this when dragging:

https://user-images.githubusercontent.com/90187801/144230672-b8c9b40c-288c-4bb4-89e0-ce2da044f47a.mp4

Asmatzaile:
> Hi, Sorry I made a comment then deleted it because I was mistaken - I thought Ableton does both Y value and also around, but it looks like it's just the Y value. Just checked Ableton, Reaper and Logic, they both have this Y value, they don't really have a way to go 'around' and would prefer to keep it that way as it is the standard now.

So yes, among the DAWs I checked and had knobs (Ableton, Logic, Reaper, LMMS, FL Studio), all accept the Y value and only Reaper accepts the X value also. Worth noting that other DAWs that don't have knobs (for example, Cubase, Ardour and Audacity, or Dorico's mixer) use X displacement for panning.
So, even though adding horizontal displacement would feel more natural for me and maybe some other users (after all, drag left to pan left, and drag right to pan right), and it wouldn't interfere with the workflow of people accustomed to dragging vertically, I understand that to resemble the knob's behavior in most other applications you'd want to ignore the movement in the X axis.

cbjeukendrup:
@jessjwilliamson @DmitryArefiev I fixed all these issues! And two more:

- Improved appearance of FX slots
- All channels will show an equal number of FX slots, so when some channels have less filled slots than others, those channels will have multiple empty slots. The maximum number of 4 slots still applies.

Next round of testing can begin!

jessjwilliamson:
> @jessjwilliamson @DmitryArefiev I fixed all these issues! And two more:
> 
> * Improved appearance of FX slots
> * All channels will show an equal number of FX slots, so when some channels have less filled slots than others, those channels will have multiple empty slots. The maximum number of 4 slots still applies.
> 
> Next round of testing can begin!

Check your discord messages about the empty slots please! :)

jessjwilliamson:
@cbjeukendrup Here are the next couple of UI steps to look into now :)
1. Sounds are 26 pixels high and Audio FX are 24 (if that becomes a height problem in the future then all should be 24)
2. Take a look into the hover states of each button of the Sounds and FX slots - bypass button, open VST and dropdown. At the moment it's acting that a whole button is 1 hover state, but hit states are so far correct. If you go the the 'MS4 Main Interface Panels' project I've highlighted with a green box to show where to find the state rules for you to look at :) 

Thank you!

cbjeukendrup:
@jessjwilliamson Fixed!

@vpereverzev I heard that you had deliberately not implemented "showing multiple empty fx slots when other channels have more filled slots" because that would cause problems. Could you tell me what problems that would cause?

I implemented it in this PR, but my implementation is probably naive and inefficient, so probably it's better if I remove it from this PR and either me or someone else creates a more decent solution. 

Of course it is undesired that there is an effect in the second slot but not one in the first slot. That situation could occur with this PR. But that situation can also occur now already, namely when the user adds two effects and then removes the first one but keeps the second one (although I have not been able to produce that situation, because the app crashes almost always when removing an effect).

So, what do you think I should do with the solution in this PR?

vpereverzev:
> @jessjwilliamson Fixed!
> 
> @vpereverzev I heard that you had deliberately not implemented "showing multiple empty fx slots when other channels have more filled slots" because that would cause problems. Could you tell me what problems that would cause?
> 
> I implemented it in this PR, but my implementation is probably naive and inefficient, so probably it's better if I remove it from this PR and either me or someone else creates a more decent solution.
> 
> Of course it is undesired that there is an effect in the second slot but not one in the first slot. That situation could occur with this PR. But that situation can also occur now already, namely when the user adds two effects and then removes the first one but keeps the second one (although I have not been able to produce that situation, because the app crashes almost always when removing an effect).
> 
> So, what do you think I should do with the solution in this PR?

Ye, it was an intentional thing. In short, it's all about playing around the performance and efficiency. And even the current state is not as performant, as I want it to be.

So I'd encourage you to exclude any changes to the models and limit yourself with QML changes

cbjeukendrup:
> Ye, it was an intentional thing. In short, it's all about playing around the performance and efficiency. And even the current state is not as performant, as I want it to be.
> 
> So I'd encourage you to exclude any changes to the models and limit yourself with QML changes

Thanks, I removed that commit. I only kept the changes to the model that were necessary for the QML things (and renamed `ItemRole` to `ChannelItemRole`, since that's slightly more descriptive and was necessary to avoid naming conflicts with QML `Loader`'s `item` property), but those are trivial.

So I think this PR is ready for review then! (It's probably the easiest to look at the commits in isolation.)

vpereverzev:
> > Ye, it was an intentional thing. In short, it's all about playing around the performance and efficiency. And even the current state is not as performant, as I want it to be.
> > So I'd encourage you to exclude any changes to the models and limit yourself with QML changes
> 
> Thanks, I removed that commit. I only kept the changes to the model that were necessary for the QML things (and renamed `ItemRole` to `ChannelItemRole`, since that's slightly more descriptive and was necessary to avoid naming conflicts with QML `Loader`'s `item` property), but those are trivial.
> 
> So I think this PR is ready for review then! (It's probably the easiest to look at the commits in isolation.)

OK, I'll take a look tomorrow

src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
```diff
@@ -241,17 +235,17 @@ Item {
...
             id: selectorLoader
 
-            Layout.preferredWidth: selectorLoader.active ? 24 : 0
+            Layout.fillWidth: !titleLoader.visible
+            Layout.preferredWidth: root.height
             Layout.preferredHeight: root.height
             Layout.alignment: Qt.AlignRight
 
-            active: (root.showAdditionalButtons || root.resourcePickingActive) && root.supportsMenu
+            visible: root.showAdditionalButtons && root.supportsMenu
```
vpereverzev, line 243: not quite sure about this change, seems redundant

src/playback/view/internal/abstractaudioresourceitem.cpp
```diff
@@ -31,6 +31,11 @@ bool AbstractAudioResourceItem::isBlank() const
     return true;
 }
 
+bool AbstractAudioResourceItem::isActive() const
+{
+    return true;
```
vpereverzev, line 36: it's better to make it 'false' by default

src/playback/view/mixerpanelmodel.cpp
```diff
@@ -34,6 +34,10 @@ MixerPanelModel::MixerPanelModel(QObject* parent)
     controller()->currentTrackSequenceIdChanged().onNotify(this, [this]() {
         load(QVariant::fromValue(m_itemsNavigationSection));
     });
+
+    connect(this, &QAbstractItemModel::modelReset, this, &MixerPanelModel::rowCountChanged);
```
vpereverzev, line 38: I think you could just emit countChanged at "add/removeItem" methods

src/playback/qml/MuseScore/Playback/MixerPanel.qml
```diff
@@ -75,8 +78,8 @@ Rectangle {
         clip: true
         boundsBehavior: Flickable.StopAtBounds
 
-        contentWidth: contentColumn.width
-        contentHeight: contentColumn.height
+        contentWidth: contentColumn.width + 1 // for trailing separator
```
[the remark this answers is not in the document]
cbjeukendrup: Sorry, what do you mean? I don't really see how the code here relates to rounding. Did you mean a different line?

src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
```diff
@@ -159,4 +159,52 @@ Dial {
...
+    }
+
+    MouseArea {
+        id: mouseArea
+        anchors.fill: parent
+        onDoubleClicked: {
+            root.newValueRequested(0)
+        }
+
+        // The MouseArea steals mouse press events from the slider.
```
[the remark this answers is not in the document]
cbjeukendrup: I agree we can do this in a future PR. Maybe we should do this via C++, like the QML components also do. It would be great if we could subclass QML components, but we can't because they're in private headers...

src/playback/qml/MuseScore/Playback/MixerPanel.qml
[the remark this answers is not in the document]
vpereverzev: Yeah, ignore this one. I was reviewing this at deep night.

src/framework/audio/qml/MuseScore/Audio/KnobControl.qml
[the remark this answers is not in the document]
vpereverzev: If you're speaking about QtQuick.Controls, then there is no need to deal with this library here. The logic around this "new" component is pretty simple.

However, if it'll be necessary, we can wrap our own implementation into Control qml type (which is a base (C++) component for all QtQuick.Controls)

src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
cbjeukendrup: Do you mean the change from `active` to `visible`? Mainly because I think it is semantically more correct, but also because Layout automatically excludes items that are invisible when it calculates the position of other items. So we don't need to set `Layout.preferredWidth` to zero manually (using property binding). That has the advantage that we can bind other properties to `Layout.preferredWidth`. But you could also argue that it is no problem when `Layout.preferredWidth` is set to zero when the loader is inactive, because then the items that would bind properties to it are not rendered anyway... 

Or if you mean that I removed `|| root.resourcePickingActive`, that is because I moved that inside the `root.showAdditionalButtons` property, because I found that we're using `root.showAdditionalButtons || root.resourcePickingActive` again and again.

jessjwilliamson:
Hi! @vpereverzev is this ready to be merged please? :)

src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
vpereverzev: Ha, haven't seen this "root.showAdditionalButtons" before, who made this change?

Looks fine to me anyway

vpereverzev:
@cbjeukendrup please, resolve the conflicts and I'll merge this one

vpereverzev (approved):

cbjeukendrup:
Done!

PR #9936, opened 2021-11-30
Small staff beams

Nick-Mazuk:
Previously, beams were not placed correctly on smaller staves:

image

Now, they are:

Screen Shot 2021-11-30 at 3 00 17 PM


oktophonie:
The stem length/attachment is now good, but the distance between the beams is a bit off (too wide).

Normal size:
image

Small size:
image

Nick-Mazuk:
The beam distance should be fixed.

We'll have to double-check this in PR https://github.com/musescore/MuseScore/pull/9937 after this one is merged. That PR fixes this problem in a slightly different manor to also allow for wide beam spacing.

oktophonie (approved):

RomanPudashkin (approved):

PR #9937, opened 2021-11-30
enable support for wide beams

Nick-Mazuk:
PR #9683 updated the style preferences for beams for MS4.

One of these style settings is for wide beams, spaced 1 space apart instead of the traditional 0.75 spaces. This PR adds support for this.

- [x] Needs feedback by @oktophonie 
- [x] Needs several tests


src/engraving/libmscore/tremolo.cpp
```diff
@@ -290,7 +290,7 @@ void Tremolo::layoutOneNoteTremolo(qreal x, qreal y, qreal h, qreal spatium)
         yOffset -= up ? 1.5 * spatium : 1 * spatium;
         yOffset -= beams >= 2 ? 0.5 * spatium : 0.0;
     } else if (beams) {
-        yOffset -= beams * 0.75 * spatium;
+        yOffset -= beams * (score()->styleD(Sid::beamDistance) == 1.0 ? 1.0 : 0.75) * spatium;
```
cbjeukendrup, line 293: Would it eventually make sense to change the style value to a bool like "useWideBeams", and do the conversion for old scores while importing those? 
Otherwise, it could be good to use `qFuzzyIsNull(score()->styleD(Sid::beamDistance))` instead of a plain comparison.

Nick-Mazuk: I would thing a `useWideBeams` style parameter would make much more sense—especially since I've made that comparison in several places in this PR.

I did notice that the `beamDistance` parameter is used in `src/engraving/libmscore/scorefont.cpp` on line 531. I'm not as familiar with that part of the codebase yet. Do you know if updating the style parameter to be a boolean would negatively affect the `scorefont.cpp` code?

If not, I'll go ahead and update the style parameter in this PR as well.

Nick-Mazuk: Ok, I believe I updated everything correctly including the UI code. Could you double-check it? I've not touched any UI code yet, so while it seems to be working, you probably have a better idea about all the edge cases.

Nick-Mazuk:
This PR is essentially ready and complete.

However, there will be a small merge conflict with https://github.com/musescore/MuseScore/pull/9936, so this PR will need to be updated before merging. https://github.com/musescore/MuseScore/pull/9936 should be merged first.

src/engraving/libmscore/tremolo.cpp
cbjeukendrup: Looks good to me!

Nick-Mazuk:
Will need to be rebased and refactored again once https://github.com/musescore/MuseScore/pull/9976 is merged as there will be merge conflicts

src/engraving/libmscore/beam.cpp
```diff
@@ -1001,11 +1006,9 @@ void Beam::layout2(std::vector<ChordRest*> chordRests, SpannerSegmentType, int f
         LayoutBeams::respace(&chordRests);
     }
 
-    // FIXME
-    if (!chordRests.front()->isChord() || !chordRests.back()->isChord()) {
-        NOT_IMPL_RETURN;
-    }
+    _beamSpacing = score()->styleB(Sid::useWideBeams) ? 4 : 3;
 
+    // todo: add edge case for when a beam starts or ends on a rest
```
cbjeukendrup, line 1011: The problem is that `toChord` will cause a crash (assertion failure) when the item is not a chord. The VTests contain that scenario, which is why I added a check for that.

Nick-Mazuk: Oh, yes. I removed that when testing because it was causing another issue with this specific PR and forgot to add that back in.

Nick-Mazuk:
The vtests failed in an expected way. Part of this PR includes a minor bugfix where small notes (either because the staff is small or the note is small on a full-size staff) now have the correct beam spacing.

oktophonie (approved):

RomanPudashkin (approved):

PR #9952, opened 2021-12-02
set specific version of uncrustify on CI

igorkorsukov:
0.73

### DIFF
--- a/.github/workflows/ci_codestyle.yml
+++ b/.github/workflows/ci_codestyle.yml
@@ -7,10 +7,10 @@ on:
 
 jobs:
   codestyle:    
-    runs-on: macos-10.15
+    runs-on: ubuntu-18.04
     steps:
     - name: Clone repository
       uses: actions/checkout@v2    
     - name: Test coding style
       run: |
-        bash ./build/ci/macos/checkcodestyle.sh
+        bash ./build/ci/linux/checkcodestyle.sh
--- a/build/ci/linux/checkcodestyle.sh
+++ b/build/ci/linux/checkcodestyle.sh
@@ -21,8 +21,10 @@
 echo "Check code style"
 trap 'echo Check code style failed; exit 1' ERR
 
-brew install uncrustify findutils
-export PATH="/usr/local/opt/findutils/libexec/gnubin:${PATH}"
+mkdir -p $HOME/build_tools/uncrustify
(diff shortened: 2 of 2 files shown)


cbjeukendrup:
I'm not a fan of this, because now I can't use Uncrustify anymore on my Macs (with Homebrew, you can't install a specific version). 

I will try if that version from the Amazon AWS works on my Mac...

igorkorsukov:
> I will try if that version from the Amazon AWS works on my Mac...

no, it won't work because it's Linux

> I'm not a fan of this, because now I can't use Uncrustify anymore on my Macs (with Homebrew, you can't install a specific version).

You can build it yourself, it's easy, we (in house) all do it :)
* You need to download the archive with the sources in the release of the desired tag
* Builds in several commands, according to the instructions on the main page


In general, I understand your inconvenience.
But on the other hand, Uncrustify is updated unexpectedly, maybe a year, or maybe a month. The update may not be in good time, it may break our priorities.
Another problem, not all changes are good, some changes raise questions, perhaps they require a revision of the config, this requires discussion.

Jojo-Schmitz:
Well, judging from how often the code style check fails even for code from the inhouse team, I have a slight doubt about whether you're using it locally ;-)
Anyway, the GitHub actition is what I personally use for doing these checks... I'm not even sure whether a port of Uncrustify to Windows does exist.

cbjeukendrup:
Linux is still Unix like macOS, so I'm still going to give it a try. But if it doesn't work, building from source will be easy enough. 

And I agree that it is inconvenient that Uncrustify is updated at unexpected moments (and what makes it even more unexpected, is that the new version becomes available on GitHub CI with some delay after it is released on Homebrew). So let's keep it this way for now :)

And Uncrustify has many bugs. I believe our config is fine, but with every new version, some bugs get introduced. Maybe I'm going to look into ClangFormat at some time, because it might be a bit smarter.

igorkorsukov:
> Well, judging from how often the code style check fails even for code from the inhouse team, I have a slight doubt about whether you're using it locally ;-)

I definitely use it, I have it configured to use it in the QtCreator on save the file, the style is automatically corrected. But it happens that with a massive renaming it may not work correctly (

Jojo-Schmitz:
No worries, I was just pulling legs ;-)

PR #9955, opened 2021-12-02
Add rest position in musicxml export

julienripoche:
Resolves: https://musescore.org/en/node/34346

Export rest position attributes (default-x, default-y) in the same way it has already been done for notes.


julienripoche:
Hi ! I added some code to export rest positional attributes in the same way as for the notes.
I originally tested it on the v3.6.2 branch and it worked as expected.
But I couldn't manage to install the latest version of Qt in order to build the master branch on my machine.

Jojo-Schmitz:
@lvinken Mind to double-check this?

src/importexport/musicxml/internal/musicxml/exportxml.cpp
```diff
@@ -3737,6 +3737,29 @@ void ExportMusicXml::chord(Chord* chord, int staff, const std::vector<Lyrics*>*
     }
 }
 
+//---------------------------------------------------------
+//   restPosition
+//---------------------------------------------------------
+
+QString ExportMusicXml::restPosition(const ExportMusicXml* const expMxml, const Rest* const rest)
```
Jojo-Schmitz, line 3744: Maybe better make it `static`, matching `color2xml()`?

src/importexport/musicxml/internal/musicxml/exportxml.cpp
```diff
@@ -3737,6 +3737,29 @@ void ExportMusicXml::chord(Chord* chord, int staff, const std::vector<Lyrics*>*
...
+        const double pageHeight  = expMxml->getTenthsFromInches(expMxml->score()->styleD(Sid::pageHeight));
+
+        double measureX = expMxml->getTenthsFromDots(rest->measure()->pagePos().x());
+        double measureY = pageHeight - expMxml->getTenthsFromDots(rest->measure()->pagePos().y());
+        double noteX = expMxml->getTenthsFromDots(rest->pagePos().x());
+        double noteY = pageHeight - expMxml->getTenthsFromDots(rest->pagePos().y());
+
+        res += QString(" default-x=\"%1\"").arg(QString::number(noteX - measureX, 'f', 2));
+        res += QString(" default-y=\"%1\"").arg(QString::number(noteY - measureY, 'f', 2));
+    }
```
Jojo-Schmitz, line 3758: Not sure, but maybe better to have
```
    } else {
        res = "";
    }
```

src/importexport/musicxml/internal/musicxml/exportxml.cpp
```diff
@@ -3757,6 +3781,7 @@ void ExportMusicXml::rest(Rest* rest, int staff)
 
     QString noteTag = QString("note");
     noteTag += color2xml(rest);
+    noteTag += restPosition(this, rest);
```
Jojo-Schmitz, line 3784: I wonder if we could simply use `notePosition(this, dynamic_cast<Note*>(rest))` here, i.e. fully reuse the existing code rather than duplicating it? Make the entire fix a one-liner.

julienripoche:
I actually thought it was possible but I'm not very familliar with c++ and thus did not know the correst syntax.
I tried what you proposed.

src/importexport/musicxml/internal/musicxml/exportxml.cpp
Jojo-Schmitz: Or, probably much cleaner, change notePosition to this:

```
//   noteRestPosition

QString ExportMusicXml::noteRestPosition(const ExportMusicXml* const expMxml, const ChordRest* const cr)
{
    QString res;

    IF_ASSERT_FAILED(!cr->isNote() && !cr->isRest()) {
        return "";
    }

    if (configuration()->musicxmlExportLayout()) {
        const double pageHeight  = expMxml->getTenthsFromInches(expMxml->score()->styleD(Sid::pageHeight));

        double measureX = expMxml->getTenthsFromDots(cr->measure()->pagePos().x());
        double measureY = pageHeight - expMxml->getTenthsFromDots(cr->measure()->pagePos().y());
        double noteX = expMxml->getTenthsFromDots(cr->pagePos().x());
        double noteY = pageHeight - expMxml->getTenthsFromDots(cr->pagePos().y());

        res += QString(" default-x=\"%1\"").arg(QString::number(noteX - measureX, 'f', 2));
        res += QString(" default-y=\"%1\"").arg(QString::number(noteY - measureY, 'f', 2));
    }

    return res;
}

**Comment by Jojo-Schmitz** [comment]
Well, yeah, I'm not sure about that casting trick either, it certainly does look quite hacky, but maybe my alternative approach is better, no casting needed and inside making sure the method is only used on notes and rests, not also on chords.

File: src/importexport/musicxml/internal/musicxml/exportxml.cpp [hunk 4]
```diff
@@ -3757,6 +3757,7 @@ void ExportMusicXml::rest(Rest* rest, int staff)
 
     QString noteTag = QString("note");
     noteTag += color2xml(rest);
+    noteTag += notePosition(this, dynamic_cast<Note*>(rest)),
```
Jojo-Schmitz, line 3760: semicolon, not colon ;-)

julienripoche:
Oups sorry...

julienripoche:
Should I try with your cleaner approach ?

Jojo-Schmitz:
I'd like input from others on that

src/importexport/musicxml/internal/musicxml/exportxml.cpp
Jojo-Schmitz: Maybe even call it `chordRestPosition()` and drop the `IF_ASSERT_FAILED` part entirely, as it should be legal for chord too.

Jojo-Schmitz: Actually just forget about it, it just doesn't work. And the code isn't 100% identical either, `notePosition()` picks on the chord a note belongs to, `restPosition()`(of course) does not.
So back to your initial approach, 2 methods for 2 types, which similar but not identical code.
Sorry for the mess...

julienripoche:
Oh yes I forgot to mention this slight difference between the notePosition and restPosition methods.
So I rolled back to the 2 methods version.

Jojo-Schmitz:
Actually the issue this fixes is https://musescore.org/en/node/34451, https://musescore.org/en/node/34346 is just a forum thread that triggered the creation of that issue.
You may want to start the PR and/or commit title with "Fix #34451: "

julienripoche:
Done for PR and commit titles

lvinken:
Checked, thanks for looking into this. In its current state (commit 30dd4e2) the code does exactly what it is supposed to do.

I do agree with Jojo that the duplication of code is not desired and also think the mechanism to generate default-x and default-y should be generalised (and extended to other elements).

Wrt to this change specifically, as the code (both for Note and Rest) depends the element's postion and its containing measure's position, in master both Note and Rest inherit from EngravingItem and EngravingItem defines the measure() function, the solution for now could be to create a single function supporting both Note and Rest with an EngravingItem as parameter.

Jojo-Schmitz:
So something like this?
```
//   noteRestPosition

QString ExportMusicXml::noteRestPosition(const ExportMusicXml* const expMxml, const EngravingItem* const ei)
{
    QString res;

   const auto chord = ei->chord();

    if (configuration()->musicxmlExportLayout()) {
        const double pageHeight  = expMxml->getTenthsFromInches(expMxml->score()->styleD(Sid::pageHeight));

        double measureX = expMxml->getTenthsFromDots(chord->measure()->pagePos().x());
        double measureY = pageHeight - expMxml->getTenthsFromDots(chord->measure()->pagePos().y());
        double noteX = expMxml->getTenthsFromDots(ei->pagePos().x());
        double noteY = pageHeight - expMxml->getTenthsFromDots(ei->pagePos().y());

        res += QString(" default-x=\"%1\"").arg(QString::number(noteX - measureX, 'f', 2));
        res += QString(" default-y=\"%1\"").arg(QString::number(noteY - measureY, 'f', 2));
    }

    return res;
}
```

(Entirely untested!)

Jojo-Schmitz:
For 3.x this here does seem work:
```
//   elementPosition

static QString elementPosition(const ExportMusicXml* const expMxml, const Element* const elm)
      {
      QString res;

      if (preferences.getBool(PREF_EXPORT_MUSICXML_EXPORTLAYOUT)) {
            const double pageHeight  = expMxml->getTenthsFromInches(expMxml->score()->styleD(Sid::pageHeight));

            const auto meas = elm->findMeasure();
            return res;

            double measureX = expMxml->getTenthsFromDots(meas->pagePos().x());
            double measureY = pageHeight - expMxml->getTenthsFromDots(meas->pagePos().y());
            double noteX = expMxml->getTenthsFromDots(elm->pagePos().x());
            double noteY = pageHeight - expMxml->getTenthsFromDots(elm->pagePos().y());

            res += QString(" default-x=\"%1\"").arg(QString::number(noteX - measureX,'f',2));
            res += QString(" default-y=\"%1\"").arg(QString::number(noteY - measureY,'f',2));
            }

      return res;
      }
```
In master it probably should be this instead
```
        IF_ASSERT_FAILED(meas) {
            return res;
        }
```

lvinken:
Both Jojo's 3.x and master solution look OK to me.

julienripoche:
Hi @Jojo-Schmitz and happy new year !
Should I commit your elementPosition method for use in the master branch ?
I understand you take care of the 3.x branch ?

Jojo-Schmitz:
Not sure, either way is apparently OK. Mine seems a bit more generic though

Jojo-Schmitz:
Note that the `Q_ASSERT(!meas)` im my version for 3.x (https://github.com/musescore/MuseScore/pull/9955#issuecomment-986015549) is just wrong and bad.
But then the `IF_ASSERT_FAILED(meas)` for master (https://github.com/musescore/MuseScore/pull/9955#issuecomment-986003042) is probably equelly bad

lvinken:
I think an assertion is kind of useful here: if the code is generalised it should be robust against being called for unexpected element types (such as elements that are not attached to measures). In that case for a production build simply returning an empty string should be OK. For a development build either a message or a failed assertion would be OK.

This could be achieved by
Q_ASSERT(meas); if (!meas) return res;

Jojo-Schmitz:
OK, so for master is should be `IF_ASSERT_FAILED(!meas)`

lvinken:
Yes, but (I assume) without the exclamation mark.

Jojo-Schmitz:
No, it is like a norlaf if, that abends when the condition is true:
```
#define IF_ASSERT_FAILED_X(cond, msg) if (!(cond)) { \
        LOGE() << "\"ASSERT FAILED!\":" << msg << __FILE__ << __LINE__; \
        Q_ASSERT(cond); \
} \
    if (!(cond)) \

#define IF_ASSERT_FAILED(cond) IF_ASSERT_FAILED_X(cond, #cond)
```

Jojo-Schmitz:
@julienripoche are you going to implement it that way, or should I?

Jojo-Schmitz:
I just did it, see PR #10456

cbjeukendrup:
Closing, since this has been replaced by #10456.

PR #9979, opened 2021-12-06
add support for beam breaks

Nick-Mazuk:
Adds back in support for beam breaks. Behavior should be identical to v3.6.

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -741,6 +741,7 @@ void Beam::createBeamSegments(std::vector<ChordRest*> chordRests)
         levelHasBeam = false;
         Chord* startChord = nullptr;
         Chord* endChord = nullptr;
+        bool breakBeam = false;
 
         for (uint i = 0; i < chordRests.size(); i++) {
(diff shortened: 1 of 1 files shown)


src/engraving/libmscore/beam.cpp
```diff
@@ -741,36 +741,40 @@ void Beam::createBeamSegments(std::vector<ChordRest*> chordRests)
...
             ChordRest* chordRest = chordRests[i];
             if (!chordRest->isChord()) {
                 continue;
             }
             Chord* chord = toChord(chordRest);
-            if (level < chord->beams()) {
+            Beam::Mode beamMode = chord->beamMode();
+            breakBeam = level < chord->beams()
+                        && ((level >= 1 && beamMode == Beam::Mode::BEGIN32) || (level >= 2 && beamMode == Beam::Mode::BEGIN64));
+            if (level < chord->beams() && !breakBeam) {
```
cbjeukendrup, line 755: I think you'll need to change `Beam::Mode` to `BeamMode`, because it was renamed recently.

Nick-Mazuk:
superseded by https://github.com/musescore/MuseScore/pull/10200

PR #9980, opened 2021-12-06
Fixed crash on getting shortcut

igorkorsukov:
The crash itself was added by this PR
https://github.com/musescore/MuseScore/pull/9973

But I also paid attention to the implementation of the `hortcutsRegister::shortcut` method.
I read the explanation from @wizofaus (https://github.com/musescore/MuseScore/pull/8908/files#r694720397), I think that the problem indicated in the explanation needs to be solved differently, we can, for example, add a separate method to create shortcuts that are absent in  `shortcuts.xml` or, as @wizofaus suggests, use the ui action resistor as source for build the form of shortcuts. While this is not meant to be, not all actions need to have shortcuts. I believe that we, as developers, must explicitly prescribe which actions can have shortcuts and which cannot (for example, we can explicitly prescribe these actions in `shortcuts.xml`, but not assign sequences to them). Because the actions themselves can be our technical internal implementation, not public, they can be added, removed, changed ... If we grant access to all actions indiscriminately, then this will restrain us in the future, we will have to invest additional efforts to maintain backward compatibility.

In any case, it is not worth solving this problem by imposing a side effect on a method that is not intended for this, and the fact that it is now called for all actions is just an implementation detail that can be changed, and accordingly it will stop fulfilling the additional responsibility imposed on it ...

### DIFF
--- a/src/framework/shortcuts/internal/shortcutsregister.cpp
+++ b/src/framework/shortcuts/internal/shortcutsregister.cpp
@@ -270,15 +270,9 @@ Notification ShortcutsRegister::shortcutsChanged() const
     return m_shortcutsChanged;
 }
 
-const Shortcut& ShortcutsRegister::shortcut(const std::string& actionCode)
+const Shortcut& ShortcutsRegister::shortcut(const std::string& actionCode) const
 {
-    const auto& shortCut = findShortcut(m_shortcuts, actionCode);
-    if (shortCut.action.empty()) {
-        Shortcut sc = { actionCode, 0 };
-        m_shortcuts.push_back(sc);
-        return findShortcut(m_shortcuts, actionCode);
-    }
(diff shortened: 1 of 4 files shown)


Eism (approved):

wizofaus:
> 

That was the one change in that PR I would have been happy to take out,  even if the workaround was to modify shortcuts.xml to include empty entries for actions with no shortcuts. 
I would say though that having actions as "publicly" known is fine and parallels how many other applications that support scripting, toolbar/menu customization and shorts do it too. Each action may then be implemented by internal functions that are not visible to end users,  so I don't see an issue with not being free to change implementation details. 
But the shortcuts dialog is still quite broken,  I'm happy to work on it if we can get an agreement as to the best way forward (including removing my kludge!)

Issue #9987, opened 2021-12-06
[MU4 Issue] Grace note ledger lines are too long


oktophonie:
The length (and thickness) of grace note ledger lines should scale with the size of the note.

Currently default grace note size is 70% of a normal note. The thickness of the grace note ledger lines _does_ scale properly, but their length does not; they are slightly too long (about 80% that of a normal ledger line):

![image](https://user-images.githubusercontent.com/20604032/144855937-82d1a7c9-50b8-44df-9609-736667b6dc5e.png)

(The same behaviour can be observed in 3.6)

PR #9989, opened 2021-12-06
fix grace note ledger line lengths

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/9987

This PR multiplies a note's ledger line extension by the note's size. This way, grace notes have ledger lines that are appropriately wide.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -855,8 +855,8 @@ void Chord::addLedgerLines()
 
             //ledger lines need the leftmost point of the notehead with a respect of bbox
             x = note->pos().x() + note->bboxXShift();
-            if (x - extraLen < minX) {
-                minX  = x - extraLen;
+            if (x - extraLen * note->mag() < minX) {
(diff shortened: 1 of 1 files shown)


Nick-Mazuk:
The vtests are failing as expected.

Jojo-Schmitz:
Not quite, see https://github.com/musescore/MuseScore/runs/4432982318?check_suite_focus=true#step:14:371, it fails on an assertion, so not all vests do run.

Jojo-Schmitz:
And on the 2 vests that do swho issues, only one os retaled to ledger lines on grace notes (accudental-9), the other (accidental-5) also has (quite small) ledger line differences, but no grace notes.

oktophonie:
The chords in accidental-5 are actually cue-size notes (as they are whole notes it's not extremely obvious at first glance), so the differences here are to be expected.

Jojo-Schmitz:
Ah, I see. Seems a `vtests` label is in order ;-)

oktophonie (approved):

RomanPudashkin (approved):

Issue #9777, opened 2021-11-18
[MU4 Issue] VTests not fails on crash


igorkorsukov:
**Describe the bug**
If a crash occurs during the generation of the current images, then the remaining images will not be generated, only those that have been generated will be compared, and the test will not be marked as failed (if the comparison is successful)

**To Reproduce**
1. Make PR with problem with some vtest score (should be crash with some score)
2. On CI will be run vtest
3. The vtest artifact will show that the images were generated only before the problem file, after the crash, the rest of the images were not generated. And vtest passed successfully.

**Expected behavior**
vtest should be fail with any crash

**Screenshots**
see https://github.com/musescore/MuseScore/pull/9738

Jojo-Schmitz:
Could we also make the vtests artifacts (much) smaller, by only archiving the 'compare' directory('s content)? The other 2 directories ('current' and 'ref') don't really serve any purpose IMHO. The 3.x vtests didn't include them either BTW.
Such an artifact is currently 7.4MB, with almost all of that due to 'current' and 'ref'.
Downloading and checking the vteste results would be much faster if those unneeded files weren't there.

njvdberg:
I see the generation of vtests fails on an assert, is it an option to use a "normal" build for vtests instead of a "debug" build?
Then mscore shouldn't crash on an assert, should it??

Jojo-Schmitz:
I'd rather get the assert fixed though

njvdberg:
I agree it is better to fix the assert but now a `Q_ASSERT` is causing a crash which stops further vtests while reporting an incorrect success. In that case I think leaving the assert for now and run a normal build is best of 2 bad cases.

Jojo-Schmitz:
The lesser evil... OTH the assert is easier to fix, so the low hanging fruit: just comment it out.
Extra benefit: the crash is gone in the FlatPak version of MuseScore too (which for some stupid reason gets build in Debug mode)

PR #9991, opened 2021-12-06
Make VTests fail when mscore crashes during PNG generation

cbjeukendrup:
Resolves: #9777 

I'm also working on (temporary) fixes for the crashes.

### DIFF
--- a/thirdparty/haw_logger/logger/log_base.h
+++ b/thirdparty/haw_logger/logger/log_base.h
@@ -42,7 +42,7 @@
 #define DEPRECATED LOGD() << "This function deprecated!!"
 #define DEPRECATED_USE(use) LOGD() << "This function deprecated!! Use:" << use
 #define NOT_IMPLEMENTED LOGW() << "Not implemented!!"
-#define NOT_IMPL_RETURN NOT_IMPLEMENTED return
+#define NOT_IMPL_RETURN NOT_IMPLEMENTED; return
 #define NOT_SUPPORTED LOGW() << "Not supported!!"
 #define NOT_SUPPORTED_USE(use) LOGW() << "Not supported!! Use:" << use
 
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -1000,8 +1000,14 @@ void Beam::layout2(std::vector<ChordRest*> chordRests, SpannerSegmentType, int f
         // fix horizontal spacing of stems
         LayoutBeams::respace(&chordRests);
     }
-    Chord* startChord = toChord(chordRests[0]);
-    Chord* endChord = toChord(chordRests[chordRests.size() - 1]);
+
+    // FIXME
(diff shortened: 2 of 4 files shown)


cbjeukendrup:
Yay, the test is faling as expected. So now I'll push the commits that "fix" the crashes. Note those fixes will only be effective when this PR is merged, because the VTests check also builds the master branch, which doesn't have the fixes yet.

@Nick-Mazuk I think the second of the two fixes is something for you to look at.

Nick-Mazuk (commented):
So the end goal of this code is to get the chords that the beam attaches to.

Does MuseScore support beams that hang over rests at the endpoints?

thirdparty/haw_logger/logger/log_base.h
```diff
@@ -42,7 +42,7 @@
 #define DEPRECATED LOGD() << "This function deprecated!!"
 #define DEPRECATED_USE(use) LOGD() << "This function deprecated!! Use:" << use
 #define NOT_IMPLEMENTED LOGW() << "Not implemented!!"
-#define NOT_IMPL_RETURN NOT_IMPLEMENTED return
+#define NOT_IMPL_RETURN NOT_IMPLEMENTED; return
```
Jojo-Schmitz, line 45: Why?

Jojo-Schmitz:
> Does MuseScore support beams that hang over rests at the endpoints?

Yes

thirdparty/haw_logger/logger/log_base.h
cbjeukendrup: Otherwise it would expand to `LOGW() << "Not implemented!!" return`, wouldn't it? That is not valid.

Jojo-Schmitz: Oops, of course. I misses that it is 3 'words'

thirdparty/haw_logger/logger/log_base.h
```diff
@@ -42,7 +42,7 @@
 #define DEPRECATED LOGD() << "This function deprecated!!"
```
Jojo-Schmitz, line 42: Here an "is" is missing. And "function" probably should be "method"?

thirdparty/haw_logger/logger/log_base.h
```diff
@@ -42,7 +42,7 @@
...
 #define DEPRECATED_USE(use) LOGD() << "This function deprecated!! Use:" << use
```
Jojo-Schmitz, line 43: same

cbjeukendrup: I think I'll rather remove them altogether. They are not used, as far as I can find.

Jojo-Schmitz: They are taken from the original source. Maybe getting them fixed there is in order, @igorkorsukov ?

igorkorsukov: Thanks.
I will fix them later

PR #9992, opened 2021-12-06
Refactor VTests to build reference and current in parallel

cbjeukendrup:
Oops, I planned to test this on my own fork first...


Jojo-Schmitz:
Maybe, while at it, also only upload the "compare" part of all the images? That's what the vtests in 3.x do and IMHO is sufficient. Less space needed, less time to download and check. None ideally (if no diffs)

igorkorsukov:
@cbjeukendrup Would you please also rename the artifact, like in executable builds

Jojo-Schmitz:
"Artifact name: , is incorrectly provided" ?

cbjeukendrup:
Still not there yet... 😞

cbjeukendrup:
Finally! 
@igorkorsukov If you confirm that it works as it should when the VTests are failing, I will remove the topmost commit and then this should be ready for merge!

igorkorsukov:
> Finally! @igorkorsukov you can see here how it works if the VTests are failing: https://github.com/musescore/MuseScore/actions/runs/1550456992
> 
> And here how it works if they are succeeding: https://github.com/musescore/MuseScore/actions/runs/1550651602
> 
> I think this is ready for merge now!

Excellent!!

cbjeukendrup:
So now, I'm also a CI Engineer! 🥳

PR #10000, opened 2021-12-07
[MU4] Fix #306189: Extend Tremolo palette

Jojo-Schmitz:
to allow for "Divide measured tremolo by [2-4.6]", `tremoloDivisiDots[2-4,6]`, implemented as articulations, actually even as staccatos, but (currently) without any playback effect on tremolos (as 'normal' staccatos don't work on those either).

Resolves: *https://musescore.org/en/node/106081* and *https://musescore.org/en/node/306189* as well as a recent discussion in the OpenScore Contributors' Telegram chat and the notation channel on the MuseScore Discord server.

Not sure yet how to deal with MusicXML ex- and import, or whether at all, @lvinken ? For now just the staccato itself is exported in case of those added symbols


tools/fonttools/smufl2sym-in-trans.json
```diff
@@ -260,6 +260,10 @@
   "chantCaesura": null,
   "coda": null,
   "codaSquare": null,
+  "tremoloDivisiDots2": null,
+  "tremoloDivisiDots3": null,
+  "tremoloDivisiDots4": null,
+  "tremoloDivisiDots6": null,
```
Jojo-Schmitz, line 266: oops, wrong (alphabethical) order

src/palette/internal/palettecreator.cpp
```diff
@@ -515,6 +515,19 @@ PalettePtr PaletteCreator::newTremoloPalette()
...
         sp->appendElement(tremolo, tremolo->subtypeName());
     }
+
+    static SymIdList dots {
+        SymId::tremoloDivisiDots2,
+        SymId::tremoloDivisiDots3,
+        SymId::tremoloDivisiDots4,
+        SymId::tremoloDivisiDots6
+    };
+    // include additional symbol-based fingerings (temporarily?) implemented as articulations
```
Jojo-Schmitz, line 525: comment needs to get amended

Jojo-Schmitz:
Why do the vtests fail?

Jojo-Schmitz:
There's an issue with this: if these dots are below the notehead, they are too close to it (or too close to the bottom staff line).

Above they do look OK. 'Normal' staccatos don't have this issue.

Jojo-Schmitz:
What's wrong with the Linux AppImage build?

cbjeukendrup:
> There's an issue with this: if these dots are below the notehead, they are too close to it (or too close to the bottom staff line).

That is because these SMuFL symbols are apparently only meant to be placed above notes. Compare their placement relative to the baseline with the staccato articulation:

| articStaccatoAbove | articStaccatoBelow | tremoloDivisiDots2 |
|-|-|-|
|Schermafbeelding 2021-12-08 om 20 59 29|Schermafbeelding 2021-12-08 om 21 01 26|Schermafbeelding 2021-12-08 om 21 01 45|

cbjeukendrup:
Of course, if it is really desired to have a "below the staff" version, you can offset the symbol by `bbox().height()`, but I would expect that SMuFL has a reason to include only the "above" version. @oktophonie What's your opinion?

oktophonie:
If the SMuFL spec implies they should only go above notes, we should probably respect that.

That said, even symbols which should always go above sometimes need to go below (when applying to a second voice with stems down), though generally these are then mirrored vertically (which would work fine here).

If the aim is to use these symbols as stand-ins for repeated staccatos, I'd prefer to avoid a hack just to make that work - because if we wish to do that, we should instead implement a better solution that a) uses the primary staccato symbol and b) can work with other articulation marks as well.

Jojo-Schmitz:
Well, the original request for these shows them under the notes.

lvinken:
With respect to MusicXML, I can't find anything in the current spec related to tremolo division dots. This may not be supported yet. Also, the SMuFL description "Divide measured tremolo by [2-4.6]" suggest the semantics differ from staccato. Exporting as staccato could be incorrect (even though it would be visually similar).

Jojo-Schmitz:
OK, thanke, I've removed them from MusicXMK export.

ecstrema:
Good Job on getting all the x000 to 10000 @Jojo-Schmitz

src/engraving/types/symnames.cpp
```diff
@@ -5879,10 +5879,10 @@ constexpr const std::array<const char*, size_t(SymId::lastSym) + 1> SymNames::s_
...
     "Combining tremolo 4",
     "Combining tremolo 5",
-    "Divide measured tremolo by 2",
-    "Divide measured tremolo by 3",
-    "Divide measured tremolo by 4",
-    "Divide measured tremolo by 6",
+    QT_TRANSLATE_NOOP("symUserNames", "Divide measured tremolo by 2"),
+    QT_TRANSLATE_NOOP("symUserNames", "Divide measured tremolo by 3"),
+    QT_TRANSLATE_NOOP("symUserNames", "Divide measured tremolo by 4"),
+    QT_TRANSLATE_NOOP("symUserNames", "Divide measured tremolo by 6"),
```
cbjeukendrup, line 5885: The context should be `engraving/sym` nowadays

cbjeukendrup:
What is the status of this: https://github.com/musescore/MuseScore/pull/10000#issuecomment-988819153 ?

Jojo-Schmitz:
No change as far as I can tell, but let's rebase first

src/engraving/types/symnames.cpp
Jojo-Schmitz: Yup, done

src/engraving/types/symnames.cpp
```diff
@@ -5879,10 +5879,10 @@ constexpr const std::array<const char*, size_t(SymId::lastSym) + 1> SymNames::s_
     "Combining tremolo 3",
     "Combining tremolo 4",
     "Combining tremolo 5",
-    "Divide measured tremolo by 2",
-    "Divide measured tremolo by 3",
-    "Divide measured tremolo by 4",
-    "Divide measured tremolo by 6",
+    QT_TRANSLATE_NOOP("engraving/sym", "Divide measured tremolo by 2"),
+    QT_TRANSLATE_NOOP("sengraving/sym", "Divide measured tremolo by 3"),
```
cbjeukendrup, line 5883: Typo here...

Jojo-Schmitz: %&$§
Fixed

RomanPudashkin (approved):

oktophonie (approved):

oktophonie (approved):

PR #10005, opened 2021-12-07
Try fix vtests

igorkorsukov:


cbjeukendrup:
Wouldn't it be worth to wait until I finish PR #9992, to see if that helps? (I'll do that very soon!)

igorkorsukov:
> Wouldn't it be worth to wait until I finish PR #9992, to see if that helps? (I'll do that very soon!)

Ok, it's just that now all the vtests are failing, as far as I can see, the problem with the comparison of ready PNG

igorkorsukov:
I check locally, for the same code generated different images :)

image

Moreover, two pictures can be generated, instead of one (which means two pages instead of one)
Something is calculated with a random result, probably some variable is not initialized and therefore a different result

igorkorsukov:
@cbjeukendrup 

At the moment, I propose to remove the problematic score (tremolo-1.mscx) from the vtests and deal with it separately.
We can merge this PR and the vtests will pass successfully
Or we can wait for the end of your work
What do you think will be better?

cbjeukendrup:
Indeed, that one test is not working well. Let's disable it (I've also done that in my PR too).
PR #9992 is now ready (final test round), so I think it's better to merge that one directly, to prevent a conflict.

HemantAntony:
I didn't know there was a problem with the vtests. I was a bit confused when it failed because I didn't do anything related to Scores

cbjeukendrup:
> I didn't know there was a problem with the vtests. I was a bit confused when it failed because I didn't do anything related to scores

I understand your confusion! But is really wasn't your fault :)

HemantAntony:
That's good to know :)

PR #10006, opened 2021-12-07
[MU4] Removed some shortcuts/actions

RomanPudashkin:
We do not plan to support these shortcuts/actions in MU4


Jojo-Schmitz:
Many of these are very usefull shortcuts/actions. "toggle-visible" for example is pretty important IMHO, also "full-measure-rest", "toggle-mmrest" and all the puct up/down ones

vpereverzev (approved):

RobFog:
Am I right to assume that the shortcut for toggling multi-measure rests was one of the shortcuts removed?

Jojo-Schmitz:
Yes, that was one of them, and not juct the shortcut, the entire action, so no shortcut possible anymore.
Unless I'm missing something, that is...

RobFog:
That'd be odd.

bkunda:
I've just looked through the list of actions and can confirm we **will** now be porting over:

clef-bass
pitch-up-diatonic-alterations
pitch-down-diatonic-alterations
toggle-autoplace
autoplac-enabled

Some of these initially belonged to a lower priority list, but now that the most important commands from #8908 have been ported we can turn our attention to these.

There are a few that I thought might be questionable, but perhaps someone can enlighten me as to their usefulness:

mirror-note: seems to produce an undesirable engraving result

These ones all seem to duplicate the same function as their "toggle" equivalent:
sharp2-post
sharp-post
nat-post
flat-post
flat2-post
set-visible
unset-visible

Do let me know if there is a clear need for these, though, and we can restore them as well 🙂

Jojo-Schmitz:
Sometimes you need a note of the same pitch twice on the same stem, like for example when notating 2 voices, like Soprano and Alto, in one staff and by using just one voice, to denote that both 'voices' sing that same pitch rather than just one of them. For that you'd need to flip one head onto the other side if the stem

bkunda:
You make a good point. 
@RomanPudashkin Can we please also reinstate mirror-note? 
Thanks @Jojo-Schmitz

Jojo-Schmitz:
Those non-toggle shortcuts/commands can be usefull for a list- or range selection with mixed content, like vilsible and invisble elementes, where toggle would invert that, but a 'visble' switch would make all visible, (i.e. not change those those visible already)

RobFog:
@bkunda What about toggle-mmrest?

Jojo-Schmitz:
Needed esp. in (linked) parts (which have them on by default) in case additional notes need to get entered there. Finding the same spot in the main score to add them there might be cumbersome

bkunda:
This all makes sense. I will chat with @RomanPudashkin about reinstating these 👍🏻
I think toggle-mmrest is actually working?
toggle-mmrest

RobFog:
If toggle-mmrest is still there, that's great. :-)

Jojo-Schmitz:
Good. Wonder then what about this in this PR got removed?

RobFog:
Perhaps the option to assign other shortcuts?

PR #10007, opened 2021-12-07
[MU4] Fix "dotted eighth = quarter" metric modulation

Jojo-Schmitz:
Follow up to #5901, where I screwed up the math.

Same issue and fix in and for 3.x

PR #10010, opened 2021-12-08
Beam custom direction

Nick-Mazuk:
Previously, if a beam group was flipped (i.e., the direction is not the default), the slant for the beam was not calculated correctly. More specifically, beams would often be flat if it were supposed to be slanted, or vice versa. This would occur either when there were multiple voices on a single staff, or if the beam group was flipped manually.

This should now be fixed.


Nick-Mazuk:
With the exception of this vtest, the vtests are failing as expected. I need to think about this case in the morning.

Screen Shot 2021-12-07 at 10 30 59 PM

oktophonie:
Is this (the flipped B) not just down to the middle-line-contextual-stem-direction-flipping (now undone with #9969 )?

Jojo-Schmitz:
Ah, I'm having the same issue in #10007

Jojo-Schmitz:
And indeed it dues. So just rebase and it'll be fine

oktophonie (approved):

RomanPudashkin (approved):

PR #10018, opened 2021-12-08
Clearing property value types [Step 13]

igorkorsukov:


src/engraving/types/typesconv.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include "log.h"
+
+using namespace mu::engraving;
+
+template<typename T>
+struct Item
+{
+    T type;
+    QString xml;
+    QString userName;
```
cbjeukendrup, line 33: Currently, the username _is_ translatable (because of `Q_TRANSLATE_NOOP`), but it is never really translated, as far as I can see. I could imagine that it is sometimes desired to have the untranslated username, and sometimes the translated version.

So you could change the userName member to `const char*`, and then add an additional function below, like `TConv::toTranslatedUserName`, which translates the username using `mu::qtrc` and returns a QString.

Alternatively, if you want to keep the translating outside of this module, the `TConv::toUserName` function should return `const char*`, since `mu::qtrc` needs that, and converting from QString to `const char*` every time is not so nice.

igorkorsukov: You are right, I didn’t think about it.
I will change `userName` to `const char*`, and add `qtrc` to `toUserName` method.

igorkorsukov: added qtrc

vpereverzev (approved):

Issue #10023, opened 2021-12-08
[MU4 Issue] Tablature staves have weird beam/stem placement


DmitryArefiev:
**Describe the bug**
Tablature staves (TABS) have wierd beam/stem placement

![beams on tab](https://user-images.githubusercontent.com/90187801/145294302-dafa31d3-08e2-4d10-9fe8-645489bff1cf.PNG)

**Desktop:**
Win/Mac/Lin

oktophonie:
Another consequence of #9792 - fix coming!

RobFog:
"wierd" → weird

PR #10039, opened 2021-12-10
add support for tab staff beams

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/10023

After the beam refactoring in #9792, beams on tablature staves got really messed up. This PR fixes those issues for both "common" and "full" staves.


Nick-Mazuk:
Codestyle fails due to a missing space in beams.h. However, I didn't touch this line of code, so this error existed before this PR. I'm not going to fix it because:

1. This is probably a false positive. The needed extra space goes against our normal code style
2. The line of code in question is irrelevant to this PR. If it needs to be fixed, it should be fixed in a new PR.

Nick-Mazuk:
The vtests are also failing with the desired behavior. The vtests label is appropriate here.

Jojo-Schmitz:
The culprit line for the condestyle check failure came in via 6ea0b9e

oktophonie (approved):

Issue #10043, opened 2021-12-10
[Installer] MU4 installer should not contain "3"  and License Agreement should be updated (Windows)


DmitryArefiev:
1. MU4 installer should not contain 3 (from previous version)

![01](https://user-images.githubusercontent.com/90187801/145563747-7eb9cd2e-16de-41b0-8f79-1ab1980d3e64.PNG)

2. End-User License Agreement should be updated 

![LA](https://user-images.githubusercontent.com/90187801/145565213-30ab76f0-fc5f-440b-b39b-ef5f0ee38467.PNG)

Jojo-Schmitz:
The bogus 3 is in [.../build/packaging/installer_background_wix.png](https://github.com/musescore/MuseScore/blob/master/build/packaging/installer_background_wix.png)
The outdated copyright year in [.../LICENSE.rtf](https://github.com/musescore/MuseScore/blob/master/LICENSE.rtf)

cbjeukendrup:
For the 1st part: see also https://github.com/musescore/MuseScore/pull/7219, where I tried to implement the new images a long time ago, but we did not include them in the 3.6 release (as we had planned earlier) because we weren't sure whether it would work correctly and there were some problems that prevented us from testing it.

The new top banner would become this:
![](https://github.com/cbjeukendrup/MuseScore/raw/fb7244a585f0ffa3acfc4a266c5fae722634ee49/build/packaging/installer_banner_wix.png)

And the new background would become this:
![](https://github.com/cbjeukendrup/MuseScore/raw/fb7244a585f0ffa3acfc4a266c5fae722634ee49/build/packaging/installer_background_wix.png)

@jessjwilliamson Are you still happy with these versions, or would you like to create updated ones by now?

Jojo-Schmitz:
They are is twice as big as the ones they replace, does that matter?

cbjeukendrup:
That was one of the things we were not sure about. I'm afraid it will; for macOS it did (but for macOS it works completely different). I think we just need to change the DPI from 72 to 144, and then it should work (for macOS that was the solution).

jessjwilliamson:
> For the 1st part: see also #7219, where I tried to implement the new images a long time ago, but we did not include them in the 3.6 release (as we had planned earlier) because we weren't sure whether it would work correctly and there were some problems that prevented us from testing it.
> 
[...]

PR #10044, opened 2021-12-10
[MU4]  Update copyright year and GPL version in LICENSE.rtf and Windows Installer images

Jojo-Schmitz:
Resolves: #10043 (the 2nd part)
Edit: Fix for the 1st part added too now


LICENSE.rtf
```diff
@@ -1,8 +1,8 @@
 {\rtf1\ansi\ansicpg1252\deff0\nouicompat{\fonttbl{\f0\fnil\fcharset0 Courier New;}{\f1\fnil\fcharset77 Courier New;}}
 {\*\generator Riched20 6.3.9600}\viewkind4\uc1 
 \pard\sa200\sl276\slmult1\f0\fs18\lang12 MuseScore, free and open source music notation software.\par
-Copyright (C) 1999-2020 MuseScore BVBA and others.\par
-This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation and reproduce below with the following exception:\par
+Copyright (C) 1999-2021 MuseScore BVBA and others.\par
```
Jojo-Schmitz, line 4: Or should we make it 2022 already?

Jojo-Schmitz: Is that a 'yes'? Or wait until we're really in 2022 (there are more places to make that change, I believe)

Jojo-Schmitz: Guess it really depends on when an Alpha gets released, for Beta, RC and Final Release we'd certainly need to change it anyhow

Jojo-Schmitz: @igorkorsukov cherry-pick as you see fit...

cbjeukendrup: >there are more places to make that change

Anyway, one of those places would be src/main/CMakeLists.txt, line 78, for the MACOSX_BUNDLE_COPYRIGHT.

For the rest, I can't really think of other places where it would be exposed to the user. However, it is of course at the top of each source file, but that's probably better to solve in a separate "happy new (copyright) year" pull request.

Jojo-Schmitz: Added

src/main/CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ elseif(OS_IS_MAC)
     set(MACOSX_BUNDLE_LONG_VERSION_STRING ${MUSESCORE_VERSION_FULL})
     set(MACOSX_BUNDLE_SHORT_VERSION_STRING ${MUSESCORE_VERSION})
     set(MACOSX_BUNDLE_BUNDLE_VERSION ${MUSESCORE_VERSION_FULL})
-    set(MACOSX_BUNDLE_COPYRIGHT "Copyright © 1999-2021 MuseScore BVBA and others. Published under the GNU General Public License.")
+    set(MACOSX_BUNDLE_COPYRIGHT "Copyright © 1999-2022 MuseScore BVBA and others. Published under the GNU General Public License.")
```
Jojo-Schmitz, line 78: Shouldn't a " version 3" get added here?

cbjeukendrup: Yes, good idea (I see you added it already), so that it's at least consistent with the other places.

vpereverzev (approved):

PR #10056, opened 2021-12-10
fix GH#10055: setting key signature mode in the inspector crashes

### LINKED ISSUE #10055: [MU4 Issue] Setting key signature mode in the inspector crashes
**Describe the bug**
Setting key signature mode in the inspector crashes.

 - OS: Windows 11

Howard-C:
Fixes #10055.

### DIFF
--- a/src/engraving/libmscore/key.h
+++ b/src/engraving/libmscore/key.h
@@ -61,24 +61,6 @@ enum class Key {
     DELTA_ENHARMONIC = 12
 };
 
-//---------------------------------------------------------
-//   KeyMode
-//---------------------------------------------------------
-
-enum class KeyMode {
-    UNKNOWN = -1,
-    NONE,
-    MAJOR,
-    MINOR,
-    DORIAN,
-    PHRYGIAN,
-    LYDIAN,
-    MIXOLYDIAN,
-    AEOLIAN,
-    IONIAN,
-    LOCRIAN
-};
-
 static inline bool operator<(Key a, Key b) { return static_cast<int>(a) < static_cast<int>(b); }
 static inline bool operator>(Key a, Key b) { return static_cast<int>(a) > static_cast<int>(b); }
 static inline bool operator>(Key a, int b) { return static_cast<int>(a) > b; }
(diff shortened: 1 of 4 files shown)


src/engraving/libmscore/key.h
```diff
@@ -61,24 +61,6 @@ enum class Key {
...
-    DORIAN,
-    PHRYGIAN,
-    LYDIAN,
-    MIXOLYDIAN,
-    AEOLIAN,
-    IONIAN,
-    LOCRIAN
-};
-
 static inline bool operator<(Key a, Key b) { return static_cast<int>(a) < static_cast<int>(b); }
```
igorkorsukov, line 64: I think these operators need to be moved too

src/engraving/property/propertyvalue.cpp
```diff
@@ -237,7 +237,8 @@ PropertyValue PropertyValue::fromQVariant(const QVariant& v, P_TYPE type)
     case P_TYPE::DYNAMIC_TYPE:     return PropertyValue(DynamicType(v.toInt()));
     case P_TYPE::DYNAMIC_RANGE:    return PropertyValue(DynamicRange(v.toInt()));
     case P_TYPE::DYNAMIC_SPEED:    return PropertyValue(DynamicSpeed(v.toInt()));
-    case P_TYPE::HOOK_TYPE:    return PropertyValue(HookType(v.toInt()));
+    case P_TYPE::HOOK_TYPE:        return PropertyValue(HookType(v.toInt()));
+    case P_TYPE::KEYMODE:          return PropertyValue(KeyMode(v.toInt()));
```
igorkorsukov, line 241: You also need to add to the `toQVariant` method

Jojo-Schmitz: Like up after line 172 add `case P_TYPE::KEYMODE:        return static_cast<int>(value<KeyMode>());`, right?

igorkorsukov: right!

src/engraving/libmscore/key.h
Howard-C: But these operators are for `Key`, not `KeyMode`? The definition of `Key` itself has not been moved, should it be moved as well?

src/engraving/property/propertyvalue.cpp
Howard-C: Done

src/engraving/libmscore/key.h
igorkorsukov: Oh, you're right, I was wrong, no need to move them

igorkorsukov (approved):

DmitryArefiev:
Tested #10055 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Thanks!

PR #10058, opened 2021-12-11
Fix floating VST windows on Windows

cbjeukendrup:
Resolves: https://github.com/musescore/MuseScore/pull/9830#issuecomment-979959093

The "hack" did not work well on Windows, but there appears to be a much more graceful solution. Unfortunately, that solution does not work on macOS, so on macOS we still use the old solution.

I must say that I have no idea about the situation on Linux. Maybe it works on Linux the same as on Windows, maybe the same as on macOS, or even different. Currently, I let Linux use the Windows solution.

### DIFF
--- a/src/framework/vst/view/abstractvsteditorview.cpp
+++ b/src/framework/vst/view/abstractvsteditorview.cpp
@@ -38,8 +38,34 @@ AbstractVstEditorView::AbstractVstEditorView(QWidget* parent)
     : QDialog(parent)
 {
     setAttribute(Qt::WA_NativeWindow, true);
-    updateStayOnTopness();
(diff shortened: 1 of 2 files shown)


DmitryArefiev:
Tested on **Win10** - this looks fine  with my VSTs

@jessjwilliamson  Can you check it on your Windows as well please?

Thanks!

jessjwilliamson:
> Tested on **Win10** - this looks fine with my VSTs
> 
> @jessjwilliamson Can you check it on your Windows as well please?
> 
> Thanks!

Yep! looks good, able to access inside VSTs now :D and window still remains in front! nice

RomanPudashkin (approved):

Issue #9262, opened 2021-09-24
[MU4 Issue] The panel tabs go beyond the panel if more than 3 tabs 


mattblo:
**Describe the bug**
If, for example, the Selection filter is displayed in addition to "Properties (Inspector)", "Paletts" and "Instruments", this strange look occurs.

**To Reproduce**
1. Open the Selection Filter

**Expected behavior**
@Tantacrul Show an arrow to navigate through the tabs?

**Screenshots**
![Strange Design with more than 3 Tabs in left Panel](https://user-images.githubusercontent.com/63284090/134650850-46ff70de-cad8-45cf-babb-8d6d537fd4c0.png)

**Desktop (please complete the following information):**
 - OS: Windows 11

cbjeukendrup:
This is what it is supposed to be like:
<img width="407" alt="Schermafbeelding 2021-09-24 om 11 25 15" src="https://user-images.githubusercontent.com/48658420/134651786-488ba2d2-9a2d-4a33-802b-c22e1f140876.png">
So no-active tabs shrink their width proportionally.

bkunda:
Well spotted @mattblo and @cbjeukendrup! 
In looking at this I've also noticed that the panel doesn't resize in Windows and MacOS (despite there being a resize handle).  
@Tantacrul is this panel supposed to resize horizontally? If not, then we may need to remove that resize handle.

Tantacrul:
These panels are meant to be resizable (although there is a limit to how much they can be resized). Resizing horizontally was working before. 

Thnks @mattblo : I forgot to log this.

Tantacrul:
@RomanPudashkin, @Eism, @vpereverzev - to get this right may take a bit of collaboration with me (to determine the easiest solution). @jessjwilliamson made the initial designs, so she's a good person to ask too.

@jessjwilliamson: in a nutshell, we have a few options with how tabs can work when they resize. Let's get some feedback and potentially make the design simpler if we need to. I feel we have other potential options... but what you've designed may very well be fine. Regarding horizontal resizing of panel, we just need to make sure users can't narrow them to the point where they look really weird.

jessjwilliamson:
> @RomanPudashkin, @Eism, @vpereverzev - to get this right may take a bit of collaboration with me (to determine the easiest solution). @jessjwilliamson made the initial designs, so she's a good person to ask too.
> 
[...]

PR #10059, opened 2021-12-11
Make panel tabs scrollable in case of overflow

cbjeukendrup:
Resolves: #9262

Note: this is different from the originally planned design. I looked into ways to implement that design, and got the logic for it correctly, but then I got stuck with binding loops, not-working bindings, polish loops, non-informative warnings, and hangs and crashes... 

So I decided to go for a slightly simpler solution, which appears to work well. (I also made sure it works well with keyboard navigation.)

I had to make changes to the KDDockWidgets library to make clicking at a tab panel work when the tabs are scrolled. Maybe we should offer this change to KDAB, since I think it may make sense for other projects too.

In case anyone is interested in seeing what I tried for the original design, it's here: https://github.com/cbjeukendrup/MuseScore/commit/e54032190e1d32ca02796c396568b86cae3c269e
You'll find that it kind of works, but especially when you just launch the app, it sometimes behaves very strangely.

### DIFF
--- a/src/appshell/qml/dockwindow/DockFrame.qml
+++ b/src/appshell/qml/dockwindow/DockFrame.qml
@@ -142,10 +142,11 @@ Rectangle {
             anchors.bottom: parent.bottom
             anchors.left: parent.left
 
-            width: contentWidth
+            // NOTE: + 1, because we don't need to see the separator of the rightmost tab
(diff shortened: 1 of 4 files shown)


cbjeukendrup:
Rebased to resolve conflicts.

Issue #9385, opened 2021-10-06
[MU4 Task] New design spec for inserting bars


bkunda:
This is a new design that slightly tweaks how users insert bars (measures) in MuseScore:

https://www.dropbox.com/s/s4a5yf9dzypoyf3/UX_Design_Spec_Adding-bars_V2.pdf?dl=0

In a nutshell, we would like to:

- Add a "bars" dropdown section to the properties panel
- Add a popup to this new section, which users can use to quickly specify how many bars they want to enter into a score and where
- Rename current labels for adding bars in the menu bar and contextual menus

We are aiming to have this improved functionality ready for the release of MuseScore 4. 

Please let me know if you have any questions about this.
Thanks!

bkunda:
> "The button to finally add the bars simply has "+". In many pieces of software, this design pattern with an input and a "+" button below it mean "add another input element, please!" as if it were a list. So perhaps adding some verbiage like "Insert" could help make the meaning more clear."

Yes this is an interesting one! The problem with adding some verbiage, such as "Insert bar", is that we also use this to trigger different popups and dialogs. I'm thinking that, on this one, it might be a matter of the user working out how to use this affordance by trying it out. 

> "Likewise, I wasn't entirely sure what the trash can icon meant until it was explained in slide 10. To me it looked like a "Delete the thing right next to it" pattern, which of course doesn't make sense… you can't delete the "insert bars" button! I'm not sure if there's a reason to change it as it's quite elegant once you know what it does. It was just not obvious when I first saw it."

Similarly with this; we also use this pattern in other places, such as the New Score dialog. As you say, it's quite elegant...once you work out what it does! 

<img width="1022" alt="Screenshot 2021-10-07 at 5 24 44 pm" src="https://user-images.githubusercontent.com/86290556/136331071-6b06fccc-0642-441b-96af-e911814af44b.png">

Our routes here might be to either reassess the add/remove button pattern in general, and/or to go ahead and implement this as designed for now and see whether users can easily discover how these buttons work during our early testing phases. We might be surprised!

DmitryArefiev:
Hi @bkunda, can you check this on your side as well please?

I've tested briefly on Win/Linux, and it looks fine. I have only one suggestion: maybe not to close pop-up dialog after each clicking +


https://user-images.githubusercontent.com/90187801/151558373-b37da86b-120e-4245-be90-00ada44f11bf.mp4

Thanks!

bkunda:
[...]

PR #10062, opened 2021-12-12
Implemented new design spec for inserting measures

cbjeukendrup:
Resolves: #9385

Notes: 
- As usually, I used the US spelling, i.e. "measures" instead of "bars". The British translators will "translate" it into proper British English.
- I made the "Delete measures" button with the trash icon less wide than in the design, to keep it consistent with the rest of the app.
- The "Delete measures" button is one pixel taller than the "Insert measures" button. This is a general problem (see #8098) and should be fixed separately.
- I added some tooltips where I thought it would make sense.


cbjeukendrup:
Rebased to resolve conflicts.

cbjeukendrup:
Rebased again to resolve conflicts.

RomanPudashkin (approved):

DmitryArefiev:
Hi @cbjeukendrup thank you for implementing this! All looks good!

About one improvement mentioned in  https://github.com/musescore/MuseScore/issues/9385#issuecomment-1024281077
(leave the popup opened until the user closes it) 
Should it be a separate task?

Thanks!

cbjeukendrup:
@DmitryArefiev Yes, we could see it as a separate task, which I will solve in a next PR. It turns out that the Measure Settings Model is destroyed and recreated every time the user selects another measure, which causes the popup to be closed even if I remove the code that closed it explicitly. So, it seems a more general issue, which is good for a separate task.

DmitryArefiev:
@cbjeukendrup Good! Created #10411

Thank you!

src/appshell/view/appmenumodel.cpp
```diff
@@ -470,11 +470,11 @@ MenuItemList AppMenuModel::makeTupletsItems()
 MenuItemList AppMenuModel::makeMeasuresItems()
 {
     MenuItemList items {
-        makeMenuItem("insert-measure"),
-        makeMenuItem("insert-measures"),
+        makeMenuItem("insert-measures-after-selection", qtrc("notation", "Insert after selection…")),
```
Jojo-Schmitz, line 473: Maybe "Append after selection"?

src/appshell/view/appmenumodel.cpp
```diff
@@ -470,11 +470,11 @@ MenuItemList AppMenuModel::makeTupletsItems()
...
+        makeMenuItem("insert-measures", qtrc("notation", "Insert before selection…")),
         makeSeparator(),
-        makeMenuItem("append-measure"),
-        makeMenuItem("append-measures")
+        makeMenuItem("insert-measures-at-start-of-score", qtrc("notation", "Insert at start of score…")),
+        makeMenuItem("append-measures", qtrc("notation", "Insert at end of score…"))
```
Jojo-Schmitz, line 477: and "Append to score" or "Append at end of score")?

cbjeukendrup: I'm passing these questions to @bkunda, who made the design. 🙂

Jojo-Schmitz: Or " Add measures ...„ for all 4 strings

bkunda: Just to clarify: is the question here what the menu item labels should be in the "Add" > "Measures" menu?

Jojo-Schmitz: Yes, and only that. The functionality is not questioned and very welcome

Maybe we could add "insert to beginning is scetion" and "append to end of section", but that is not as important as the other additions at all.

Jojo-Schmitz: In 3.x we had "Append measures" (to mean Append measures at end of score) and "Insert measures" (to mean insert before selection), I'd rather keep the new texts as close as possible to those.
So "Insert measures at start of score", Insert measures before selection", "Append/Insert measures to/after selection" (not really sure about the best wording here), "Apend measures to end of score"

bkunda: This is what I put in my design spec. The idea was to align the labels in this menu with the new dropdown feature: 

Screenshot 2022-02-01 at 6 15 23 pm

I tried to avoid the word "Append". It of course makes sense, but is generally not a super-frequently used word in English, and I reasoned that "Insert+location" communicated essentially the same message. 

In saying this, I'm wondering in my own design whether the word "Insert" is even necessary. The action sequence is, after all, `Add`>`Measures`>`After selection` (or whatever other location designation).

Jojo-Schmitz: I'd swap "Insert after selection" and "Insert before selection", seems more logical (and maybe rename the former to "Append to selection"). Bit in the contect of the "Add > Measure" indeed Insert and Append are not needed, In the shortcuts menu though this is different.
So "Add" > "Measure" > "Before selection" / "After Selection" / "At start of score" / "At end of score" seems good to me

Jojo-Schmitz: Also the menus needs to show the shortcuts

Jojo-Schmitz: What about single measures?

bkunda: Ok I've just checked the implementation in master again and I think can better see the scope of the issue now (my apologies, I don't think I was completely understanding what the issue was before).

The "Add" menu is as I've designed it, and I think it works well: 
Screenshot 2022-02-01 at 8 46 23 pm

The `+` menu, however, has inconsistent terminology and functionality:

Screenshot 2022-02-01 at 8 45 42 pm

My recommendation would be that we align the options in the `+` menu with what we have in the `Add` menu. It would therefore make sense to retain the use of the word "insert" as this provides better context for the options in the `+` menu. 

All menu items trigger the "Number of measures to insert" dialog, which comes pre-filled with a single measure (this is how single measures can be entered). 

Shortcuts do need tweaking as well now, because the dialog that gets triggered doesn't delineate between measures inserted at the selection point, and measures inserted at the end of the score. 

Since the present PR has already been merged, I'll raise these tweaks as a separate issue.

Jojo-Schmitz: That, for single measures, would be more complicated then before, a simple 'Ins' inserted a single measure before the selection and a Ctrl+Ins appended one measure to the end of the score, very simple and efficient, now is a much more involved

bkunda: Sorry, I didn't mean to suggest we would change the way the shortcuts work (I also quite like how easy they make inserting a single measure). I only meant to suggest that the dialog no longer gives adequate information for the user to know where they are entering bars (on `⌥⇧B` command).

Jojo-Schmitz: I see. Still the Insert/Append one measure seem missing from the menu?

bkunda: I'm afraid this was a by-product of our efforts to both clarify the intended result of the menu actions while also simplifying the language.

The problems in MS3 are that:
- "Append" can be an impenetrable term for a lot of people
- The effect of "Insert measures" and "Insert one measure" never made clear _where_ the inserted measure would be applied (I.e. _before_ or _after_ the selection point)? The user had to actually click on the label to work out where the new measure was inserted (and because the new bar(s) never got highlighted, it was doubly confusing trying to work out the whereabouts of the new measure(s)).
 
In trying to move away from the word "Append", while also making it very clear in the label _where_ the inserted measure would be applied, some simplification in the logic was required, lest we end up with a whole string of wordy options ("insert one measure before selection point", "insert one measure after selection point", "insert measures before selection point", "insert measures after selection point", "insert one measure at the end of the score"...etc, etc.). This included moving away from the distinction between inserting just one measure, and inserting multiple measures. We rationalised that this wouldn't be such a bad trade-off, however, as inserting measures is typically an action that doesn't usually take place with any tremendous frequency in the course of creating a score (It's usually a much less frequent function than, say, inserting notes and rests, or adding dynamics or articulation marks to a score).

A (hopefully) nice result of this is that the user now has additional options for specifying where they want the inserted measures to appear ("at start of score" and "after selection point" are new). On top of this, we've doubled the number of access points for inserting measures, exposing this functionality in what we hope will be a much more visible way. Obviously I'll be testing all this in user testing in the near future, but we are hoping all these revisions make for a less, rather than more, complicated user experience.

RobFog: I agree that "append" is better avoided.

PR #10066, opened 2021-12-13
add new rule for determining flat beams

Nick-Mazuk:
Adds new rule for determining flat beams. The dictator is the note on the higher end of the beam group if the direction is up, or the note on the lower end of the beam group if the direction is down. So if the note right next to the dictator is at the same level of the dictator, and if the chord would otherwise be slanted if the dictator didn't exist, the the chord is now slanted.

image


Jojo-Schmitz:
Seems all the vtest failures are expected, so a 'vtests' lable seem in order?

oktophonie (approved):

Jojo-Schmitz:
That merge commit really doesn't belong into a PR

PR #10071, opened 2021-12-13
Fix #9902 - keyboard editing for slurs/lines not working properly

### LINKED ISSUE #9902: [MU4 Issue] Can't adjust line or slur length via keyboard
It is not possible to make lines longer or shorter via the keyboard. Affects "spanner" elements with draggable handles:

- hairpins _(also unable to adjust height of "jaws")_
- slurs & ties _(also unable to adjust curvature)_
- cresc./dim.
- ottavas (8va/8vb)
- pedal markings
- trill lines
- etc.

__Expected behaviour__

1. On **entering** (i.e. adding/inserting) one of the above elements, it should become selected and the **rightmost** handle should gain keyboard focus.

image

2. On **navigating** to one of the above elements with Alt+Left/Alt+Right, the **leftmost** handle should gain keyboard focus.

image

3. Pressing the Tab key should move focus clockwise to one of the other handles (Shift+Tab to go anticlockwise). But see suggested change below.

**Actual behaviour**

1. Line elements are not selected when they are entered in the score.
2. The drag handles do not gain keyboard focus during navigation with Alt+Left and Alt+Right (you must press Alt+Shift+E to edit the line).
3. The Tab shortcut moves focus away from the score instead of moving to a different handle. This is sometimes desirable, sometimes not. See suggestion below.

**Suggested change to Tab navigation**

MuseScore 3 used the Tab key to navigate between drag handles in a closed loop, which prevented Tab being used to navigate to other UI elements like the Inspector (see bug [MU#299941](https://musescore.org/en/node/299941)). In MuseScore 4, perhaps we should "break the loop"? This means:

- Pressing **Tab** would still move you to the **next handle (right)**, but when you reach the **rightmost handle**, pressing Tab again would navigate to a UI panel on the **right** of the score (or below it, like the status bar).
- Pressing **Shift+Tab** would still move you to the **previous handle (left)**, but when you reach the **leftmost handle**, pressing Shift+Tab again would navigate to a UI panel on the **left** of the score (e.g. the Palettes).

Basically, each handle would be treated as a separate UI element for Tab and Shift+Tab, and when you run out of handles you then you start to navigate the rest of the UI.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/9902)

Keyboard editing not working when slurs first selected (but not in editing mode), and tab not working reliably to cycle between active grip handle

### DIFF
--- a/src/engraving/playback/playbackeventsrenderer.cpp
+++ b/src/engraving/playback/playbackeventsrenderer.cpp
@@ -96,16 +96,6 @@ void PlaybackEventsRenderer::renderRestEvents(const Ms::Rest* rest, mpe::Playbac
 
 void PlaybackEventsRenderer::renderArticulations(const Ms::Chord* chord, PlaybackContext&& ctx, mpe::PlaybackEventList& result) const
 {
-    if (renderChordArticulations(chord, std::forward<PlaybackContext>(ctx), result)) {
-        return;
-    }
-
-    renderNoteArticulations(chord, std::forward<PlaybackContext>(ctx), result);
-}
-
-bool PlaybackEventsRenderer::renderChordArticulations(const Ms::Chord* chord, PlaybackContext&& ctx,
-                                                      mpe::PlaybackEventList& result) const
-{
     TRACEFUNC;
 
     for (const auto& pair : ctx.commonArticulations.data()) {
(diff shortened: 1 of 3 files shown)


wizofaus:
BTW I also discovered a crash if one of the first things you try to do is edit a slur using the keyboard, as "notation->score()" is returning a null pointer when the ScoreCallbacks object is first initialised.

Issue #9210, opened 2021-09-21
[MU4 Issue] Rearranging the instruments, leads to incorrect brackets


mattblo:
**Describe the bug**
Rearranging the instruments, leads to incorrect brackets

**To Reproduce**
1. Create a new Score "SATB Closed Score + Piano"
2. Drag the "women" under the "men" voice

**Expected behavior**
Set the brackets correct or remove them if they are not needed.

**Screenshots**
Bevor rearranging
![Vorher](https://user-images.githubusercontent.com/63284090/134126761-f36888f6-f3d1-468d-b5c5-0b70fa392a14.png)

After rearranging
![nachher](https://user-images.githubusercontent.com/63284090/134126789-a6abe4b6-0f5a-4490-82d6-b86da816cc9f.png)

**Desktop (please complete the following information):**
 - OS: Windows 10

njvdberg:
Worked fine in MU3, I'll have a look.

njvdberg:
When creating a new score  "SATB Closed Score + Piano", the score order is `Choir` which is correct.
However, when changing the order of the staves, the score order becomes `Custom` instead of `Choir (customized)`. Since the `Custom` score order does nothing with bracketing and barlines the bracketing stays where it is (starting at "women" and spanning 2 staves) which is wrong.

mattblo:
Does this mean that this behavior is intentional and nothing needs to be changed? I would at least expect that the parenthesis is removed or added again when I restore the order as at the beginning.

mattblo:
If you set the order to Orchestral, the brackets for women and men are removed. If you set it back to Choir, they will not be added again...

https://user-images.githubusercontent.com/63284090/144417339-9aaae686-4098-4361-9191-c19420ff98c0.mp4

njvdberg:
>Does this mean that this behavior is intentional and nothing needs to be changed? I would at least expect that the parenthesis is  removed or added again when I restore the order as at the beginning.

No, this is not intentional and will be solved (I'm working on it already).

> If you set the order to Orchestral, the brackets for women and men are removed. If you set it back to Choir, they will not be added again...

Indeed, when you delete the brackets, selecting the same score order again doesn't recreate the brackets again. Switching it to another score order, e.g. `Orchestral`, and than back to `Choir` **will** recreate the brackets. Nevertheless something I have to look at.

mattblo:
You may also want to take a look at #10040, #10034, #10038

njvdberg:
[...]

PR #10084, opened 2021-12-14
System objects now appear on template-specified tracks

asattely:
Templates include information about which staves have system objects above them (i.e. tempo text, voltas, jump markers, system text, etc.), and this PR enforces those objects on the appropriate staves.
image

There is also pre-existing functionality where you can hold ctrl when dragging a volta to a staff and it shows on that staff instead of the preset system staves:
image

This PR also deals with saving and loading of all system objects, regardless of staff.

### DIFF
--- a/share/instruments/orders.xml
+++ b/share/instruments/orders.xml
@@ -2,7 +2,7 @@
 <museScore>
     <Order id="orchestral">
         <name>Orchestral</name>
-        <section id="woodwind" showSystemMarkings="true">
+        <section id="woodwind">
             <family>flutes</family>
             <family>oboes</family>
             <family>clarinets</family>
--- a/share/templates/02-Choral/02-SATB_+_Organ/02-SATB_+_Organ.mscx
+++ b/share/templates/02-Choral/02-SATB_+_Organ/02-SATB_+_Organ.mscx
@@ -44,6 +44,9 @@
       <family>organs</family>
       <unsorted/>
       </Order>
+    <SystemObjects>
+      <Instance staffId="5" barNumbers="false"/>
+      </SystemObjects>
     <Part>
       <Staff id="1">
         <StaffType group="pitched">
--- a/share/templates/02-Choral/03-SATB_+_Piano/03-SATB_+_Piano.mscx
+++ b/share/templates/02-Choral/03-SATB_+_Piano/03-SATB_+_Piano.mscx
@@ -44,6 +44,9 @@
       <family>organs</family>
       <unsorted/>
       </Order>
+    <SystemObjects>
+      <Instance staffId="5" barNumbers="false"/>
+      </SystemObjects>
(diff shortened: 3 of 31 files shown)


asattely:
Bah. Killed the utests. I'll give it a look tomorrow

bkunda:
This is looking super nice so far! Thanks so much @asattely 😀

Not sure if you're already onto this (and so apologies if I'm pointing out something you already know!): System objects don't seem to be respecting changes made to top-staff instruments from the Instruments dialog (after the template has been opened). 

In this example:

- Adding a piccolo above the flute: system objects stay with the flute and don't move up a staff
- Removing the violin 1, then re-adding it: system objects move down to violin 2 in first instance (expected), but don't jump back up again to the newly-added violin 1
- Lastly, my intuition told me that I might be able to adjust the bracket around the string staves, in the hopes I could "trick" the system objects to jumping back up. I was being a tad optimistic, I think 😅

system-objects-1

wizofaus:
Ah, glad to see someone was brave enough to take this on, I'd looked into it a bit but got bogged down with other stuff - good work though!

njvdberg:
@bkunda Since this feature makes use of the score ordering the points you mention are expected. If you add a piccolo the score order is changed to `Custom`. My guess is every is fine after you change the score order back to `Orchestral`. This is related to #9210 and other related issues I'm looking at right now.

@asattely I was working on this a year ago but got distracted. Good to see you picked this up :+1:.

asattely:
So, the unit test is failing on split measures, and apparently the tempo isn't being written into the files being diff'd with the reference score. I am having a lot of trouble reproducing this in any scores I create myself. I'm going to keep at this tomorrow to see if there's anything weird I'm missing. The rest of the utests have been fixed, and the vtests magically started being okay again.

asattely:
I've added some xml to the save file so that scores can save their list of system object staves. This will be useful in the short-term if we want to make more fine adjustments to templates, and later on when we implement configuration of this stuff.

Unfortunately since it modifies the mscx file slightly, I expect every utest to fail now, so there's going to be a lot of fixing of reference files in my future.

asattely:
The utest I was worried about before seems to have fixed itself with a rebase, though I seem to have introduced a new problem with the MM rests which I will investigate now.

asattely:
That'll probably do it! vtests are passing, the utests passed last time so I am confident in those, and we should be good for some good old fashioned testin'

Jojo-Schmitz:
The vtests fail at multimeasure rests, interrupted by a system element (here: rehearsal mark or time signature), by loosing those.

oktophonie (approved):

asattely:
Fixed one more bug with jump markers appearing twice in an excerpt, and rebased. As long as all the checks pass, which they should, this one will be ready to gooooo

PR #10086, opened 2021-12-14
Fixed crash on edit text style

igorkorsukov:
An attempt was made to compare a value of type string ("20") with a value of type int (20).
Previously, QVariant converted values implicitly if they are of different types and then compared them.
I do not want to add another conversion hack ... I hope we will identify all such situations of a data type mismatch.

### DIFF
--- a/src/engraving/property/propertyvalue.cpp
+++ b/src/engraving/property/propertyvalue.cpp
@@ -213,7 +213,11 @@ PropertyValue PropertyValue::fromQVariant(const QVariant& v, P_TYPE type)
     case P_TYPE::KEY_MODE:         return PropertyValue(KeyMode(v.toInt()));
     case P_TYPE::TEXT_STYLE:       return PropertyValue(TextStyleType(v.toInt()));
 
+    // other
+    case P_TYPE::ACCIDENTAL_ROLE: return PropertyValue(Ms::AccidentalRole(v.toInt()));
+    case P_TYPE::SUB_STYLE:       return PropertyValue(v.toInt());
     default:
+        UNREACHABLE;
         break;
     }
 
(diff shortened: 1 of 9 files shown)


Eism (approved):

src/engraving/property/propertyvalue.cpp
```diff
@@ -241,7 +241,9 @@ PropertyValue PropertyValue::fromQVariant(const QVariant& v, P_TYPE type)
 
     // other
     case P_TYPE::ACCIDENTAL_ROLE: return PropertyValue(Ms::AccidentalRole(v.toInt()));
+    case P_TYPE::SUB_STYLE:       return PropertyValue(v.toInt());
     default:
+        UNREACHABLE;
```
Jojo-Schmitz, line 246: If this is truely unreachable, because all possible values are dealt with, I'd rather remove the default branch and get a compiler warnign about an unhandled case as soon as another `P_TYPE` got added, but got forgotten here.

Jojo-Schmitz: But a log more have been added meanwhile it seems, so here's a merge conflict

src/engraving/style/styledef.h
```diff
@@ -27,11 +27,10 @@
 #include <vector>
 
 #include "property/propertyvalue.h"
+#include "libmscore/property.h"
 #include "config.h"
 
 namespace Ms {
-enum class Pid : int;
-
```
Jojo-Schmitz, line 34: These 2 changes are already in

src/engraving/property/propertyvalue.cpp
igorkorsukov: > If this is truely unreachable, because all possible values are dealt with, I'd rather remove the default branch and get a compiler warnign about an unhandled case as soon as another `P_TYPE` got added, but got forgotten here.

Fixed

PR #10104, opened 2021-12-15
Clearing TextStyleType (ex Tid) type [Step 16] 

igorkorsukov:


src/engraving/rw/compat/read206.cpp
```diff
@@ -1151,14 +1153,18 @@ bool Read206::readNoteProperties206(Note* note, XmlReader& e, ReadContext& ctx)
...
+        return QString();
     }
+    beginIdx += QString("<style>").size();
+
+    int endIdx = xmlTag.indexOf("</style>");
+    IF_ASSERT_FAILED(endIdx > 0) {
+        return QString();
+    }
+
+    QString s = xmlTag.mid(beginIdx, (endIdx - beginIdx));
```
igorkorsukov, line 1167: @cbjeukendrup FYI the return of the previous implementation was like "<style>System</style>", so the style enum value was never found, `DEFAULT` was always used

igorkorsukov: vtest failure expected, because fixed read of text style for 2.06 format

cbjeukendrup: Ah! Very nice that you found this! I remember seeing things like "Unknown text style: <style>System</style>" in the output, and never guessed it would have to do with this!

The problem is that `match.captured()` is actually `match.captured(0)`, which gives the substring that matched the _whole_ regex. Instead, it should have been `match.captured(1)`, to get the substring that matched the first capture group. ([Docs](https://doc.qt.io/qt-5/qregularexpression.html#extracting-captured-substrings))

So you could simplify the fix, just by changing `match.captured()` with `match.captured(1)`. But of course, it is also possible to go without QRegularExpression, like your solution. I guess that might give slightly better performance?

Jojo-Schmitz: Which is what it die in 3.x (using `QRegExpr` and it's `cap(1)`)

igorkorsukov: >  I guess that might give slightly better performance?

I think so, and I think that there is no need to use a regex if we just want to parse the **structured** text.

vpereverzev (approved):

PR #10093, opened 2021-12-14
[MU4] Implemented the chord symbols navigation

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/9708

PR #10116, opened 2021-12-16
[MU4] Fixed compilation errors

RomanPudashkin:
See https://github.com/musescore/MuseScore/pull/10093


Jojo-Schmitz:
I gues the vtests fail is expected, as that previous commit just doesn't build at all, which is what this PR is supposed to fix. So a `vtests` label seems in order

Eism (approved):

PR #10126, opened 2021-12-17
[MU4] Fix #10050: Wrong category name in New Score Wizard>Choose from template

### LINKED ISSUE #10050: [MU4 Issue] Wrong category name in New Score Wizard>Choose from template
**Additional context**

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10050

### DIFF
--- a/share/templates/categories.json
+++ b/share/templates/categories.json
@@ -1,0 +1,91 @@
+[
+    {
+      "title": "General",
+      "files":
+      [
+          "01-General/01-Treble_Clef/01-Treble_Clef.mscx",
+          "01-General/02-Bass_Clef/02-Bass_Clef.mscx",
+          "01-General/03-Grand_Staff/03-Grand_Staff.mscx"
+      ]
+    },
+    {
+      "title": "Choral",
+      "files":
+      [
+          "02-Choral/01-SATB/01-SATB.mscx",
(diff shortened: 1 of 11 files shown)


vpereverzev (approved):

share/templates/categories.json
```diff
@@ -0,0 +1,93 @@
+{
+  "categories": [
```
igorkorsukov, line 2: I believe either does not need this tag (just `[]` as the root object), or the file should be called `templates.json`

share/templates/categories.json
```diff
@@ -0,0 +1,93 @@
...
+    {
+      "title": "General",
```
igorkorsukov, line 4: Do we have a table for translating these names?

DmitryArefiev:
Tested #10050 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

share/templates/categories.json
RomanPudashkin: removed this tag

share/templates/categories.json
RomanPudashkin: We should have such a table. Translations will work in the same way as they do in MU3 but instead of passing folder names to qApp->translate, we will pass strings from this JSON

PR #10129, opened 2021-12-19
Fix #9726: Update mixer on replace instrument

### LINKED ISSUE #9726: [MU4 Issue] Part name on Mixer doesn't change
**Describe the bug**
When editing the part name from the Instruments panel, it doesn't change the name on the mixer either. 

https://user-images.githubusercontent.com/71656189/141502466-3a4bcbf7-7693-4274-9229-9cf013434bf8.mp4

 - OS: Windows and MacOS

HemantAntony:
Resolves: #9726 

This commit updates the mixer when an instrument is replaced.

### DIFF
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -600,6 +600,11 @@ void PlaybackController::setupSequenceTracks()
     partList.onItemRemoved(this, [this](const Part* part) {
         removeTrack(part->id());
     });
+
+    partList.onItemChanged(this, [this](const Part* part) {
+        addTrack(part->id(), part->partName().toStdString());
+        removeTrack(part->id());
+    });
 }
 
 void PlaybackController::setupSequencePlayer()


HemantAntony:
In my opinion, this is a temporary fix. A better fix should be made. Though this appears to work, the order does change. So the new instrument is last

src/playback/internal/playbackcontroller.cpp
```diff
@@ -600,6 +600,11 @@ void PlaybackController::setupSequenceTracks()
     partList.onItemRemoved(this, [this](const Part* part) {
         removeTrack(part->id());
     });
+
+    partList.onItemChanged(this, [this](const Part* part) {
+        addTrack(part->id(), part->partName().toStdString());
+        removeTrack(part->id());
```
Jojo-Schmitz, line 606: Isn't that the wrong way round and should remove first and add last?

HemantAntony: Yes, that's what doesn't make sense. The other way doesn't work. When you do it the other way, the instrument gets removed, but the last instrument in the mixer gets duplicated

RomanPudashkin:
I don't think this fix is reliable enough. Let's close this PR to give other developers a chance to work on this bug

PR #10131, opened 2021-12-20
fix whole note tremolo positions

Nick-Mazuk:
Positions one-note tremolos on whole notes identically to those on stemmed notes. I also removed the magic numbers by creating two new style parameters. That way there's a single source of truth for the position for the tremolos.


Jojo-Schmitz:
Vtests crash at generating the current PNGs, utests creash with a failed assertion.

Jojo-Schmitz:
Still the same assertion failute: `ASSERT: "MStyle::valueType(idx) == mu::engraving::P_TYPE::REAL" in file ../src/engraving/style/style.h, line 64`

oktophonie (approved):

PR #10158, opened 2021-12-22
[Mu3] fix #327681 illegal repeatsegment crash

jeetee:
illegal repeatsegment inserted due to a certain open volta, jump and end repeat combination

Resolves: https://musescore.org/en/node/327681

Fixes interpretation order so the end repeat is covered by the open volta if a jump also exists in the same measure.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows the *3.x* coding rules
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #10159, opened 2021-12-22
[Mu4] fix #327681 illegal repeatsegment crash

jeetee:
illegal repeatsegment inserted due to a certain open volta, jump and end repeat combination

Resolves: https://musescore.org/en/node/327681

MU4 version of PR #10158
Fixes interpretation order so the end repeat is covered by the open volta if a jump also exists in the same measure.

- [N/A - patch applied according to what builds in Mu3] I made sure the code compiles on my machine

### DIFF
--- a/src/engraving/libmscore/repeatlist.cpp
+++ b/src/engraving/libmscore/repeatlist.cpp
@@ -866,7 +866,6 @@ void RepeatList::unwind()
                     && ((*repeatListElementIt)->getRepeatCount() < (*repeatListElementIt)->measure->repeatCount())
                     ) {
                     // Honor the repeat
-                    Q_ASSERT(rs != nullptr);
                     if ((rs != nullptr) && (!rs->isEmpty())) {
                         push_back(rs);
                     }
--- a/src/engraving/libmscore/repeatlist.cpp
+++ b/src/engraving/libmscore/repeatlist.cpp
@@ -500,8 +500,15 @@ void RepeatList::collectRepeatListElements()
                             ) {
                             // The previous volta was supposed to end before us
                             // or open volta ends together with us -> insert the end
-                            sectionRLElements.push_back(new RepeatListElement(RepeatListElementType::VOLTA_END, volta, toMeasure(mb)));
-                            volta = nullptr;
+                            if (!mb->repeatEnd()) {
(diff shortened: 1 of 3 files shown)


src/engraving/libmscore/repeatlist.cpp
```diff
@@ -848,7 +855,10 @@ void RepeatList::unwind()
                     && ((*repeatListElementIt)->getRepeatCount() < (*repeatListElementIt)->measure->repeatCount())
                     ) {
                     // Honor the repeat
-                    push_back(rs);
+                    Q_ASSERT(rs != nullptr);
```
Jojo-Schmitz, line 858: Not sure we're still (supposed to be) using `Q_ASSERT()` in master?

Jojo-Schmitz (approved):

jeetee:
We might want to update Q_ASSERTS to IF_ASSERT_FAILED as per https://github.com/musescore/MuseScore/wiki/Error-handling; but I've not included it here yet, as it would touch more in this file than relevant to the issue.
Perhaps that replacement could be done in a code sweep commit?

Jojo-Schmitz:
Rebase needed. Issue came up again in https://musescore.org/en/node/335346

Jojo-Schmitz (approved):
While I don't claim to fully understand it all, the proposed change does make sense to me ;-)

jeetee:
@cbjeukendrup would be nice to have this merged in before beta; one less crash to worry about then ;-)

cbjeukendrup:
I've asked @vpereverzev to review this, since we're already entering extra careful mode :)

When it comes to asserts, we have two flavours:
- simple (C++ built-in?) `assert(…)`
- the `IF_ASSERT_FAILED` macro:
    ```c++
    IF_ASSERT_FAILED(condition) {
        // code that will be run when condition is unexpectedly false
        // (if asserts are enabled (debug build), then the assert would trigger an abort and this point won't be reached)
    }
    ```

jeetee:
The Q_ASSERT was there more as a glorified debug breakpoint; the further handling of the scenario is the same whether the assert would've been hit or not.

vpereverzev (approved):

jeetee:
ty!

Issue #10085, opened 2021-12-14
[MU4 Issue] Scrolling is too sensitive in various menus


bkunda:
**Describe the bug**
Scrolling in a number of menus remains very sensitive, and in particular doesn't always stop when the mouse wheel is stopped (in other words, the scroll animation continues after the user has stopped the mouse wheel from moving). 

This is particularly noticeable in the following long menus:

- Preferences > Keyboard shortcuts
- Preferences > Appearance > Font face
- Preferences > Import > Character set used when importing binary files (both dropdowns)
- Add-ons > Languages
- Score > Customise toolbar

Aside from the above, I suspect this is a global issue throughout the program.

**Demo 1** Showing customise toolbar, and comparison expected scroll behaviour with OS finder dialog, where the scroll behaviour stops as soon as the mouse wheel stops moving:

https://user-images.githubusercontent.com/86290556/145936148-edfdc2e8-0de9-4427-a205-c405d34b236b.mov

**Demo 2** Languages menu

https://user-images.githubusercontent.com/86290556/145936186-65b4cc16-6da4-4d4d-a4c8-54b48b54482f.mov

**Demo 3** Keyboard shortcuts menu

https://user-images.githubusercontent.com/86290556/145936214-96a54091-4e81-4b7e-b601-d021e878ad26.mov

**Expected behavior**
Scrolling should stop as soon as mouse wheel has stopped moving. It would probably be sufficient that scroll speed and general behaviour match MS3.6.

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1)
 - Windows 11

**Additional context**
See #7805, not yet merged and potentially not a global fix

cbjeukendrup:
My experience:
- On macOS, scrolling feels perfectly natural with the built-in trackpad and with Apple's Magic Mouse 2 (it starts as soon as you start scrolling, and stops as soon as you stop; the scrolling speed is comparable with native macOS apps).
- But with a simple external Logitech mouse, it is indeed a bit problematic. The ease-in-out animations are too slow, so there is always some delay, and if you turn the wheel a tiny bit, it will continue scrolling quite far.
[...]

PR #10162, opened 2021-12-23
[MU4] Set maximum flick velocity to 1500 for all listviews in the project

Eism:
Resolves: #10085

### DIFF
--- a/src/appshell/qml/Preferences/PreferencesPage.qml
+++ b/src/appshell/qml/Preferences/PreferencesPage.qml
@@ -57,7 +57,7 @@ Rectangle {
         }
     }
 
-    Flickable {
+    StyledFlickable {
         id: flickable
         anchors.fill: parent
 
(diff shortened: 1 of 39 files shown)


vpereverzev (approved):

cbjeukendrup:
How about creating a reusable StyledFlickable component that has the correct value for all properties that we use every time? Like `maximumFlickVelocity`, `boundsBehaviour`, `Scrollbar.vertical`, `clip`, etc. 
I think that would make it easier to add new Flickables without forgetting properties, and ensure that all Flickables stay consistent.

Eism:
@cbjeukendrup 
We can't just create a new Flickable component and use it for all views in the project because views are inherited from Flickable.

Ok, we can create a styled component for each view (StyledListView, StyledGridView, StyledFlickable, etc).
And it is better than it looks now

cbjeukendrup:
Ah, of course, that was the catch... Thanks!

Eism:
@cbjeukendrup fixed

llui85:
@Eism Would it be possible to add an setting option to disable the "flick"/smooth scrolling behaviour entirely as well?

PR #10167, opened 2021-12-24
Changed Align type implementation

igorkorsukov:


src/engraving/types/types.h
```diff
@@ -62,46 +62,45 @@ enum class GlissandoStyle {
...
 };
 
-constexpr Align operator|(Align a1, Align a2)
-{
-    return static_cast<Align>(static_cast<char>(a1) | static_cast<char>(a2));
-}
+enum class AlignH {
+    LEFT,
+    RIGHT,
+    HCENTER
```
vpereverzev, line 76: it might make sense to just call it AlighH::Center

vpereverzev (approved):

Issue #10076, opened 2021-12-13
[MU4 Issue] New Score> Additional Score Info: Title, Subtitle and Composer fields should select after one mouse click (Improvement)


DmitryArefiev:
**Describe the bug**
After #9186, we have auto filled Title, Subtitle and Composer with default text. 

![01](https://user-images.githubusercontent.com/90187801/145817721-2034b843-82fd-43c8-a575-a1c55ae0fb2e.PNG)

But when user wants to change them with a different text he needs to erase them at first. 
On **Win, Linux**, after you click with mouse the value selects and deselects for a moment which is strange and not convenient.
We should give to user a **quick way** to fill out Title, Subtitle and Composer with his own text

https://user-images.githubusercontent.com/90187801/145818305-b6e2ad76-0be7-46de-8990-b0a6ba6ec22e.mp4

HemantAntony:
Hey, when you edit Title, Composer, Subtitle, there is already text there. But, I was thinking if the "text" in it, could be like Lyricist and Copyright. So something like this:

https://user-images.githubusercontent.com/88136054/147342500-88eaa774-c491-42a3-9518-e583e925f4a1.mp4

Howard-C:
@HemantAntony I agree with this solution. Does this mean you already implemented the fix?

HemantAntony:
I can have the fix ready with a couple of hours. So, the text would be like this:
1. Title: Name of the Score
2. Composer: Name of the Composer
3. Subtitle: Subtitle
4. Lyricist: Name of the Lyricist
5. Copyright: Copyright
Got any suggestions for the text?

Howard-C:
Now that I think about it, I find these texts to be completely redundant… Are they really necessary? Aren't "Title", "Composer", "Subtitle", "Lyricist", "Copyright" already self-explanatory?

HemantAntony:
Yeah I was thinking about that too. We already know what it is. So just remove the text and keep it blank?

Howard-C:
The only good thing about the original implementation, I think, is the "Untitled Score" default value for the "Title" field. Maybe we can keep that and put it the same way as the "Optional" texts, but I really don't find the rest of the texts useful.

HemantAntony:
Okay, so I will grey out "Untitled Score" and remove all the other text

Howard-C:
This is only my opinion, others might disagree

DmitryArefiev:
[...]

PR #10173, opened 2021-12-27
Changed Additional Score Info default values

HemantAntony:
Resolves: #10076 

This commit changes default values of Additonal Score Info. With this
commit, the default values have been "hinted", that is greyed out.
Also, the default value of all five fields except Title has been
removed since they didn't seem to be adding any meaningful information.

### DIFF
--- a/src/project/qml/MuseScore/Project/internal/GeneralInfoItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/GeneralInfoItem.qml
@@ -28,7 +28,7 @@ Column {
     id: root
 
     property string title: ""
-    property alias info: textField.currentText
+    property alias info: textField.hint
 
     property alias navigation: textField.navigation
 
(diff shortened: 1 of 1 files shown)


src/project/qml/MuseScore/Project/internal/GeneralInfoView.qml
```diff
@@ -81,8 +81,6 @@ Column {
 
             title: qsTrc("project", "Composer")
 
-            info: qsTrc("project", "Composer / arranger")
```
RomanPudashkin, line 84: You shouldn't remove this value (the same for the subtitle field), it should be used for all scores by default

src/project/qml/MuseScore/Project/internal/GeneralInfoItem.qml
```diff
@@ -45,7 +45,7 @@ Column {
 
     TextInputField {
         id: textField
-        hint: qsTrc("project", "Optional")
+        hint: ""
```
RomanPudashkin, line 48: The hint property should be empty by default, so you can remove this line

DmitryArefiev:
Tested #10076 on Win10, Mac12, LinuxMint 20.2 - PASS

Please merge into master

Issue #10172, opened 2021-12-27
[MU4 Issue] Closing project should change the window from Score to Home


HemantAntony:
**Describe the bug**
In my opinion, when you close a project, the current window should change from Score to Home.

**To Reproduce**
1. Create a Score
2. Close it now
3. The current window is still Score

**Expected behavior**
The window should change from Score to Home

**Screenshots**

https://user-images.githubusercontent.com/88136054/147434364-ec3ee7e8-8020-4b28-96a3-897eaf6dee10.mp4

**Desktop (please complete the following information):**
 - OS: Windows 11

RomanPudashkin:
@bkunda what do you think about this? :)

bkunda:
@RomanPudashkin I agree this is a sensible suggestion 😀

PR #10174, opened 2021-12-27
Fixed Notation Page not updating to Home Page when closing project and first launch setup

HemantAntony:
Resolves: #10172 

Closing the project and first launch set up now changes the current page from notation page to home page. 

### DIFF
--- a/src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
+++ b/src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
@@ -27,8 +27,8 @@ FirstLaunchSetupModel::FirstLaunchSetupModel(QObject* parent)
     : QObject(parent)
 {
     m_pages = {
-        Page { "ThemesPage.qml", "musescore://notation" },
-        Page { "PlaybackPage.qml", "musescore://notation", /*canSkip*/ true },
+        Page { "ThemesPage.qml", "musescore://home" },
+        Page { "PlaybackPage.qml", "musescore://home", /*canSkip*/ true },
         Page { "TutorialsPage.qml", "musescore://home?section=learn" }
     };
 }
--- a/src/project/internal/projectactionscontroller.cpp
+++ b/src/project/internal/projectactionscontroller.cpp
@@ -39,6 +39,7 @@ using namespace mu::framework;
 using namespace mu::actions;
 
 static mu::Uri NOTATION_PAGE_URI("musescore://notation");
+static mu::Uri HOME_PAGE_URI("musescore://home");
 static mu::Uri NEW_SCORE_URI("musescore://project/newscore");
 
 void ProjectActionsController::init()
@@ -153,7 +154,7 @@ Ret ProjectActionsController::openProject(const io::path& projectPath_)
 
     //! Step 2. If the project is already open in the current window, then just switch to showing the notation
     if (isProjectOpened(projectPath)) {
-        return openNotationPageIfNeed();
+        return openPageIfNeed(NOTATION_PAGE_URI);
     }
 
     //! Step 3. Check, if the project already opened in another window, then activate the window with the project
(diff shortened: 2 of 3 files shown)


HemantAntony:
Also before this PR is merged, shouldn't the corresponding page for Themes Page be the Home Page:
image
instead of:
image

src/appshell/view/firstlaunchsetup/firstlaunchsetupmodel.cpp
```diff
@@ -123,4 +125,8 @@ bool FirstLaunchSetupModel::askAboutClosingEarly()
 void FirstLaunchSetupModel::finish()
 {
     configuration()->setHasCompletedFirstLaunchSetup(true);
+
+    if (!interactive()->isOpened(HOME_PAGE_URI).val) {
+        interactive()->open(HOME_PAGE_URI);
```
cbjeukendrup, line 130: Not completely sure about this one. I don't think we will want to open the home page unconditionally. For example, the user might have launched MuseScore for the first time by opening a MSCZ file. In that case, the user wants to see the Score tab after finishing the first launch setup.

HemantAntony: Oh ok. Well, I did it because whenever I build and run MuseScore in Qt Creator and I close the window, it takes me to the score page when home page would have been better for me

cbjeukendrup:
> Also before this PR is merged, shouldn't the corresponding page for Themes Page be the Home Page:
> image
> instead of:
> image
> @cbjeukendrup

I think this is a question for @bkunda!
(FWIW, I tend to agree.)

HemantAntony:
> I think this is a question for @bkunda! (FWIW, I tend to agree.)
This would actually fix the bug I was going after in the first place

bkunda:
@HemantAntony @cbjeukendrup Yes I agree, it should be the Home>Scores screen on initial startup (as it is also for every subsequent startup). 
Part of the design rationale was to give users a feel for how the interface will change when they set their appearance settings in the "getting started" dialog for the first time. The Home tab is actually better than the Score tab for this because the accent colours are easily seen in two locations (as opposed to only one in the Score tab). 
So yes, please go ahead with this!

src/project/internal/projectactionscontroller.cpp
```diff
@@ -212,6 +213,15 @@ mu::Ret ProjectActionsController::openNotationPageIfNeed()
     return interactive()->open(NOTATION_PAGE_URI).ret;
 }
 
+Ret ProjectActionsController::closeNotationPageIfNeed()
```
RomanPudashkin, line 216: Better to rename this method to openHomePageIfNeed. Also, its implementation duplicates openNotationPageIfNeed, so you can create a separate method for the implementation and reuse it twice in openNotationPageIfNeed and openHomePageIfNeed

src/project/internal/projectactionscontroller.cpp
```diff
@@ -307,6 +317,11 @@ bool ProjectActionsController::closeOpenedProject(bool quitApp)
     if (result) {
         globalContext()->setCurrentProject(nullptr);
 
+        Ret ret = closeNotationPageIfNeed();
```
RomanPudashkin, line 320: We should open the home page only if quitApp == false. So the implementation will be like this:

if (quitApp) {
dispatcher()->dispatch("quit", actions::ActionData::make_arg1<bool>(false));
} else {
Ret ret = closeNotationPageIfNeed();
...
}

DmitryArefiev:
Hi @HemantAntony @bkunda, When Getting Started dialog is working it shouldn't switch to Score page when going it's going Playback page, right? 

https://user-images.githubusercontent.com/90187801/150980397-94a0feaa-5310-4f3c-aaf1-1fd0fbb7b3c2.mp4

HemantAntony:
> Hi @HemantAntony @bkunda, When Getting Started dialog is working it shouldn't switch to Score page when going it's going Playback page, right?

I'm pretty sure it was like that before. But, I agree with you

bkunda:
Yes once it starts on home page it should stay there until it moves to the "Learn" section on the final screen.

HemantAntony:
Okay then, I'll add that too

src/project/internal/projectactionscontroller.cpp
HemantAntony: I created a new method for opening pages and deleted openHomePageIfNeed and openNotationPageIfNeed. This method accepts a uri as its argument. I hope this is okay

RomanPudashkin (approved):

DmitryArefiev:
> Okay then, I'll add that too

Thanks! All works now

DmitryArefiev:
Tested #10172 on Win10, Mac12 - FIXED      

Please merge into master

PR #10176, opened 2021-12-27
[MU4] playback_events_rendering_p1

vpereverzev:
This PR introduces the core functionality of the new playback events rendering system. The main differences from the previous system are:
- The system consists of small autonomous classes. Thus, each particular articulation subtype is processed in its own separate handler (renderer)
- The new system aimed to work using our own events format (MPE), which gives us a lot more playback details than both versions of MIDI

In general, each small handler falls into two categories:
- Meta parsers - collecting the articulations data from EngravingItems
- Renderers - building the playback events using the articulations meta data being collected 

Interesting notes:
- In order to reduce a run-time overhead (like vtables function calls) the implementation uses rare [CRTP idiom](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) which is based on the static polymorphism.
- I'm planning to optimize the performance in the next PRs. However, judging by the current measurements, it already runs 24% faster than the previous system, which is pretty funny considering we get a lot more detailed playback events here


src/engraving/playback/metaparsers/chordarticulationsparser.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+        int spannerDurationTicks = spannerTo - spannerFrom;
+
+        PlaybackContext spannerContext = ctx;
+        spannerContext.nominalTimestamp = timestampFromTicks(chord->score(), spannerFrom);
+        spannerContext.nominalDuration = durationFromTicks(ctx.beatsPerSecond, spannerDurationTicks);
+        spannerContext.nominalPositionTick = spannerFrom;
+        spannerContext.nominalDurationTicks = spannerDurationTicks;
+
+        SpannersMetaParser::instance()->parse(spanner, std::move(spannerContext), result);
```
igorkorsukov, line 108: Solution with `::instance` is temporary? I think it better to use IoC

src/engraving/playback/playbackcontext.h
```diff
@@ -0,0 +1,104 @@
...
+#include "playback/utils/pitchutils.h"
+
+namespace mu::engraving {
+struct PlaybackContext {
+    mpe::timestamp_t nominalTimestamp = 0;
+    mpe::duration_t nominalDuration = 0;
+    mpe::dynamic_level_t nominalDynamicLevel = 0;
+    int nominalPositionTick = 0;
+    int nominalDurationTicks = 0;
+    qreal beatsPerSecond = 0;
```
igorkorsukov, line 40: I recently added a separate type for this :) https://github.com/musescore/MuseScore/blob/master/src/engraving/types/bps.h

vpereverzev: Heh, I actually noticed this type when resolved conflicts, but forgot to change this particular one

I'll refine it in the next PR

igorkorsukov (approved):

Issue #10166, opened 2021-12-24
[MU4 Issue] Score jumps when entering note with mouse because dragging is turned on in Note Input mode


DmitryArefiev:
Score jumps when entering note with mouse because dragging is turned on in Note Input mode

https://user-images.githubusercontent.com/90187801/147351906-e4f1a582-5db6-4257-a96c-a07c171dfd33.mp4

**Desktop:**
 - OS: all

**Additional context**
Recent builds

Tantacrul:
Possibly related: https://github.com/musescore/MuseScore/issues/10168

igorkorsukov:
@DmitryArefiev Please check, PR https://github.com/musescore/MuseScore/pull/10177

DmitryArefiev:
Tested #10177 on Win10, LinuxMint 20.2 - FIXED

DmitryArefiev:
> Possibly related: #10168

Looks like it's different because not fixed with the current PR

PR #10177, opened 2021-12-27
Fixed notation drag

igorkorsukov:


src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -581,8 +568,10 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
     }
 
     // move canvas
-    m_view->moveCanvas(dragDelta.x(), dragDelta.y());
-    m_isCanvasDragged = true;
+    if (!m_view->isNoteEnterMode()) {
```
Eism, line 571: for optimization: I think we can calculate the value of the note enter mode(see 534 line) and reuse it here

igorkorsukov: fixed

Eism (approved):

DmitryArefiev:
Tested #10166 on Win10, LinuxMint 20.2 (Mac will check in master) - FIXED      

Please merge into master

PR #10182, opened 2021-12-27
Fix #9538: Fixed no language selected by default

### LINKED ISSUE #9538: [MU4 Issue] Preferences: no language selected by default
See https://youtu.be/0Y0X10yiOYg?t=61

HemantAntony:
Resolves: #9538 

This commit fix no language selected by default. The problem was that if
your system language was not part of the existing list of languages,
there would be no default language. Therefore, I have introduced
a fallback language which is English (US) which would be used as the
default language in the specified case

### DIFF
--- a/src/appshell/qml/Preferences/GeneralPreferencesPage.qml
+++ b/src/appshell/qml/Preferences/GeneralPreferencesPage.qml
@@ -57,6 +57,10 @@ PreferencesPage {
                 root.hideRequested()
                 preferencesModel.openUpdateTranslationsPage()
             }
+
+            onChangeCurrentLanguage: function(newLanguage) {
+                preferencesModel.currentLanguageCode(newLanguage);
+            }
         }
 
         SeparatorLine { }
--- a/src/appshell/qml/Preferences/internal/LanguagesSection.qml
+++ b/src/appshell/qml/Preferences/internal/LanguagesSection.qml
@@ -37,6 +37,9 @@ BaseSection {
 
     signal languageSelected(string languageCode)
     signal updateTranslationsRequested()
+    signal changeCurrentLanguage(string newLanguage)
+
+    property string fallbackLanguage: "en_US";
 
     Row {
         spacing: 12
@@ -50,7 +53,7 @@ BaseSection {
             textRole: "name"
             valueRole: "code"
 
-            currentIndex: dropdown.indexOfValue(root.currentLanguageCode)
+            currentIndex: findIndexOfValue(root.currentLanguageCode)
 
             navigation.name: "LanguagesBox"
             navigation.panel: root.navigation
@@ -58,6 +61,15 @@ BaseSection {
 
             onCurrentValueChanged: {
                 root.languageSelected(dropdown.currentValue)
+            }
+
+            function findIndexOfValue(languageCode) {
+                var index = dropdown.indexOfValue(languageCode)
+                if (index < 0) {
+                    changeCurrentLanguage(root.fallbackLanguage)
+                    index = dropdown.indexOfValue(root.fallbackLanguage)
+                }
+                return index
             }
         }
 


cbjeukendrup:
This PR does not really tackle the root cause of the problem. The following things are causing problems:
- The default language is "system". For me, this expands to `nl_NL`, the code for Dutch. However, the code that MuseScore uses for Dutch is `nl`. So, `nl_NL` will not be found in the list with language codes, so the dropdown will be empty.
- By default, only a selection of languages seems to be "installed". For example, Dutch is _not_ installed by default. So, even if all language codes were correct, the dropdown will still be empty if your system language is not "installed". 
  
The selection of pre-installed languages is a bit strange: for example, English is not included. I'm not sure at all whether it is a good idea to have not all languages installed by default (@Tantacrul what's your opinion on this?). At least we would have to overcome the problems mentioned above.

With the risk to digress too much, I think there's more to say about languages. The interaction between the Preferences dialog and Add-ons > Languages is not ideal in my opinion, mainly regarding UX. For example, the user has to _download_ a language using the Add-ons page, but then has to open the Preferences dialog in order to _use_ it. 
Or, if the user opens the Preferences dialog, hoping to select their language, they may not find it in the list, even though MuseScore _does_ support that language. They somehow need to figure out that they have to go to the Languages page because maybe they can still install the language from there. (And then, to use it, they need to go back to the Preferences dialog.) 
@Tantacrul maybe you or one of the other designers might want to refine this a bit.

Also, the language is set to "system" only at the first launch. After changing the language, there is no way to change it back to "system".

Tantacrul:
I do think we should have a few languages installed by default and I'm open to suggestion about how many / which.

You're also right that we're going to have to slightly clean up the UX a tiny bit here too. Doubt it would need anything major but there is a slight weirdness.

HemantAntony:
Okay I will close this PR

Jojo-Schmitz:
> I do think we should have a few languages installed by default and I'm open to suggestion about how many / which.
> 
> You're also right that we're going to have to slightly clean up the UX a tiny bit here too. Doubt it would need anything major but there is a slight weirdness.

MuseScore 1, 2 and 3 always had all supported languages installed (and the option to update those to the latest translations), and fell back to the internal texts (which are US English, the en_US 'translation' we have is for just some after-release typo fixes and the plural forms).
I don't see any reason to change any of this.

Only thing that would be nice is the ability to add more translations on the fly, rather than having to wait for the next MuseScore release (the list of available languages is built into MuseScore currently). Not sure this is possible, maybe via the Extension mechanism?

Qt's locale feature automagically uses e.g. "nl" if the system is set to "nl_NL", if an "nl_NL" translationd doesn't exists but an "nl" translation does exist.

Jojo-Schmitz:
Another nice addition would be a multi-level fallback, e.g. take untranslated "pr_BR" string from "pt", and only if not translated there from "en". Or maybe any of the languages using Cyrillic to fall back to Russian first, then English. Or the 3 "ar_XX" falling back to plain "ar" first, "ca@catalan" to "ca", "cn_XX" to one another, etc. Again, not sure this is possible at all.

Tantacrul:
> MuseScore 1, 2 and 3 always had all supported languages installed (and the option to update those to the latest translations)

I think this may have just been a communication mistake when putting together MS4. If we had a version of the languages installed in MS1-3, then we should do it for 4 too.

Jojo-Schmitz:
As far as I can tell MuseScore 4 (builds from master) still come with all the languages. Of course the translations are (very) outdated and incomplete meanwhile. But I guess that'll change around Alpha release.

Tantacrul:
OK, I'm confused about the precise issue then. We have them installed but are just not displaying them properly, I suppose. Anyway, I've marked this as a task to get tackled in the new year.

Jojo-Schmitz:
I have no idea what thos PR was supposed to fix, to be honest

HemantAntony:
> I have no idea what this PR was supposed to fix, to be honest

Well the issue is marked. You may not be having this issue. I did because my system language is Indian English which is not a language MuseScore provides (I can understand why). The function which evaluates the default language returned -1 for the index as Indian English was not in the list of languages. Its a bit complicated to explain. But the main problem is that there was no fallback language in case of something like this. So the default language was empty

Jojo-Schmitz:
But certainly all texts in English, so not an issue as far as I can tell

PR #10184, opened 2021-12-28
fix build issues on main branch

Nick-Mazuk:
The main branch currently has many build errors. This PR fixes them by adding the required static casts:

```
/src/engraving/utests/playbackeventsrendering_tests.cpp:610:9: error: non-constant-expression cannot be narrowed from type 'int' to 'short' in initializer list [-Wc++11-narrowing]
        nominalPitchLevel - PITCH_LEVEL_STEP,
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/src/engraving/utests/playbackeventsrendering_tests.cpp:610:9: note: insert an explicit cast to silence this issue
        nominalPitchLevel - PITCH_LEVEL_STEP,
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        static_cast<short>(                 )
```

### DIFF
--- a/src/engraving/utests/playbackeventsrendering_tests.cpp
+++ b/src/engraving/utests/playbackeventsrendering_tests.cpp
@@ -360,9 +360,9 @@ TEST_F(PlaybackEventsRendererTests, SingleNote_Turn_Regular)
     int expectedSubNotesCount = 4;
     duration_t expectedDuration = QUARTER_NOTE_DURATION / expectedSubNotesCount;
     pitch_level_t nominalPitchLevel = pitchLevel(PitchClass::F, 4);
-    std::vector<pitch_level_t> expectedPitches = { nominalPitchLevel + PITCH_LEVEL_STEP,
+    std::vector<pitch_level_t> expectedPitches = { static_cast<short>(nominalPitchLevel + PITCH_LEVEL_STEP),
                                                    nominalPitchLevel,
-                                                   nominalPitchLevel - PITCH_LEVEL_STEP,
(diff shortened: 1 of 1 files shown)


src/engraving/utests/playbackeventsrendering_tests.cpp
```diff
@@ -360,9 +360,9 @@ TEST_F(PlaybackEventsRendererTests, SingleNote_Turn_Regular)
     int expectedSubNotesCount = 4;
     duration_t expectedDuration = QUARTER_NOTE_DURATION / expectedSubNotesCount;
     pitch_level_t nominalPitchLevel = pitchLevel(PitchClass::F, 4);
-    std::vector<pitch_level_t> expectedPitches = { nominalPitchLevel + PITCH_LEVEL_STEP,
+    std::vector<pitch_level_t> expectedPitches = { static_cast<short>(nominalPitchLevel + PITCH_LEVEL_STEP),
```
vpereverzev, line 363: 'short' is a local implementation of int_fast_16 on your machine. You can't just cast it here, otherwise it would work only for you. I suggest to get a local variable with pitch_level_type. Like nominalPitchLevel one

vpereverzev (changes requested):

src/engraving/utests/playbackeventsrendering_tests.cpp
Jojo-Schmitz: See #10188

Issue #9096, opened 2021-09-09
[MU4 Issue] Cannot use keyboard shortcuts to zoom in


Nick-Mazuk:
**Describe the bug**

Cannot use keyboard shortcuts (⌘+ or ⌘-) to zoom in/out until you manually zoom with the menu items.

**To Reproduce**
1. Build and run MuseScore with `./build.cmake && ./build.cmake run`
2. Open up the app
3. Create a new score or open up an existing one
4. Press ⌘+ or ⌘- to zoom in/out — it won't work
5. Select to View > Zoom in, the score zooms in
6. Press ⌘+ or ⌘- again to zoom in/out, it works

**Expected behavior**

Using the keyboard shortcuts should work first try.

**Desktop (please complete the following information):**
 - OS: macOS 11.5.2 x86
 - MuseScore: master branch version

**Additional context**

I've been having a similar (potentially unrelated) issue where I cannot zoom in/out by pinching on the trackpad.

wizofaus:
That sounds like a general problem with shortcut keys not being able to activate commands unless the score window has the "keyboard focus", but I'm not sure.  The zoomin/zoomout actions are implemented and should work as is though (and on Windows work using CTRL+*+* and CTRL+*-*).

Nick-Mazuk:
That would explain why several keyboard shortcuts don't seem to be working. I just assumed many of them weren't programmed in yet.

wizofaus:
There are a lot that have only been implemented in a PR I've had up for a while now, but those two should be fine, do they work if you click on the score window first?

Nick-Mazuk:
They don't work when clicking on the score first.

wizofaus:
Very odd, can't think of any reason that activating the commands via the menu would start allowing the shortcuts to work.
Certainly works fine for me as is.  Though Ctrl+0 to zoom 100% doesn't work on the nightly build (only in my branch, which I don't entirely understand the reason for as I didn't change anything related to that command!)

Nick-Mazuk:
Is there anything I could do to help you replicate it on your end? Perhaps this is a macOS thing?

In the meantime, I've verified the same issue compiling it locally with factory settings ``, as well as using the latest nightly build ([MuseScoreNightly-202109110431-master-92d372f.dmg](https://ftp.osuosl.org/pub/musescore-nightlies/macos/4x/nightly/?C=M;O=D)).

wizofaus:
Are you running it from source - can you see all the log output?

[...]

PR #10193, opened 2021-12-29
Fix #10192, #9096: Fixed zoom in not working with Ctrl++

### LINKED ISSUE #10192: [MU4 Issue] Zoom in doesn't work with Ctrl + + but does with Ctrl + Shift + +
**Describe the bug**
If you were to use the shortcut Ctrl + +, it wouldn't zoom in. However, if you used Ctrl + Shift + +, it works.

**To Reproduce**
1. Create a Score
2. Use the shortcut Ctrl + + (as in Hold Ctrl and the Equal Buttons)
3. It does not work
4. Now do the same thing with Ctrl + Shift + +, it works

**Expected behavior**
Ctrl + + should zoom in similar to Chrome

**Screenshots**

 - OS: Windows 11
 
**Additional context**
Using Ctrl + - zooms out

HemantAntony:
Resolves: #9096
Resolves: #10192

The problem was the shortcut was just Ctrl + +. This meant I had to use
shift + = to get the + similar to how you type in + (By holding shift +
=). So, I changed the shortcut sequence to Ctrl+=

### DIFF
--- a/src/framework/shortcuts/data/shortcuts.xml
+++ b/src/framework/shortcuts/data/shortcuts.xml
@@ -834,7 +834,7 @@
     </SC>
   <SC>
     <key>zoomin</key>
-    <seq>Ctrl++</seq>
+    <seq>Ctrl+=</seq>
     </SC>
   <SC>
     <key>zoomout</key>


src/framework/shortcuts/data/shortcuts.xml
```diff
@@ -794,7 +794,7 @@
     </SC>
   <SC>
     <key>zoomin</key>
-    <seq>Ctrl++</seq>
+    <seq>Ctrl+=</seq>
```
RomanPudashkin, line 797: I don't think that it is a good idea to change the shortcut in this way. We should find out why Ctrl++ doesn't work properly

Jojo-Schmitz: To get to = I'd have to press Shift, on a German QWERTZ keyboard... whereas the + is accessible without Shift on this keyboard. So replacing one nuisance with another...

HemantAntony: Well, I think when the Shortcut was written as `Ctrl++`, it implies that plus also has to be pressed. But there is no direct plus key. You have to press `Shift + =` to get a plus. Hence, when `Ctrl + +` implies `Ctrl + Shift + =`. Now, when you click `Ctrl + Shift + =` , it does imply Zoom in, hence it work while `Ctrl + =` does not

Jojo-Schmitz: If you need Shift to get to =, so be it. Doesn't have any bearing to the shoutcut itself. Or at least shouldn't (and didn't in 3.x)
Same is true for many other shortcuts where the keys differ depending on zhe keyboard usedd, like {[]}-_.:,;+* and many more

HemantAntony: I'm also curious, do you have to press `Ctrl + Shift + =` to zoom in your browser because of your keyboard?

Jojo-Schmitz: No, as + is accessible without Shift on my keyboard, it is Ctrl++ here.

HemantAntony: Oh it seems like there are two shortcuts, I just checked I can zoom in with `Ctrl + Shift + =` (`Ctrl + +`) as well

Jojo-Schmitz: And s `Ctrl`+`+` just works for me in the current builds too.

HemantAntony: Yeah, but the "universal" zoom in button, in my experience, is `Ctrl + =` (So no need to press shift too). That's why I tried fixing this

Jojo-Schmitz: Well, no, it is not. The universal key is `Ctrl`+`+` for zooming in and  `Ctrl`+`-` for zooming out. As said, whether  you have to press `Shift` to get to either is irrelevant and depending on keyboard layout

Tantacrul:
Can confirm that the shortcut ⌘+ and ⌘- are working, although I had to reset to factory for it to work.

Now that I've done that, it seems to work for new and old scores on Mac.

HemantAntony:
@RomanPudashkin please don't merge yet, I think I'll just add the new shortcut and also keep the old one too, so two shortcuts.

Zooming in chrome works through both shortcuts

Tantacrul:
@HemantAntony - what are both shortcuts? I want to remove this idea of having multiple shortcuts for one action. It massively complicates the experience and serves basically no purpose.

HemantAntony:
> @HemantAntony - what are both shortcuts? I want to remove this idea of having multiple shortcuts for one action. It massively complicates the experience and serves basically no purpose, other than remote, niche (and largely theoretical) ones.

Okay, I agree. Then this PR is ready for review/merge @RomanPudashkin

HemantAntony:
This PR should also fix #9096 along with #10192

Tantacrul:
I've tested it and it works

Tantacrul:
@RomanPudashkin - are you happy with this PR? If so, can you merge?

RomanPudashkin:
@Tantacrul 

I am afraid that this fix will not work reliably enough. After this fix, the "zoom-in" shortcut will be highly dependent on the keyboard layout of the user. Ideally, we need to figure out how it worked before (in MU3). I can merge this PR, but I'm afraid this fix will hide the real problem from us

Tantacrul:
Then let's not merge

MarcSabatella:
> Ideally, we need to figure out how it worked before (in MU3)

It didn't :-).  Not reliably, anyhow.  I'd say it worked on about 1/3 of the systems I ever ran MuseScore on (all of which place "+" above "=").  I typically had to customize the zoom command to use Ctrl+= in addition.

RomanPudashkin:
> > Ideally, we need to figure out how it worked before (in MU3)
> 
> It didn't :-). Not reliably, anyhow. I'd say it worked on about 1/3 of the systems I ever ran MuseScore on (all of which place "+" above "="). I typically had to customize the zoom command to use Ctrl+= in addition.

Hmm, very interesting :) Thank you for the information!

@Tantacrul I will merge it in that case :) It doesn't seem to cause more problems than we have now

RomanPudashkin:
@HemantAntony thank you very much!

PR #10197, opened 2021-12-29
Fix #10072: Fixed cursor not on last inputted note

### LINKED ISSUE #10072: [Note Input] When triggering Note Input mode, the cursor (incorrectly) always defaults to the start of the score
**Describe the bug**
See video

https://user-images.githubusercontent.com/21022311/145804480-0fba750b-d3b4-4bb4-b50e-4c63f60b6a1f.mp4

**Expected behaviour**
The note input selection cursor should behave in the same way as MS3: appearing on the last selection beat on the score.

HemantAntony:
Resolves: #10072 

- [x]I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)

### DIFF
--- a/src/notation/internal/notationnoteinput.h
+++ b/src/notation/internal/notationnoteinput.h
@@ -92,6 +92,8 @@ private:
     async::Notification m_stateChanged;
     async::Notification m_noteAdded;
 
+    EngravingItem* m_lastInputedItem = nullptr;
+
     ScoreCallbacks* m_scoreCallbacks = nullptr;
 };
 }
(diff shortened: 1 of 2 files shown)


src/notation/internal/notationnoteinput.h
```diff
@@ -92,6 +92,8 @@ class NotationNoteInput : public INotationNoteInput, public async::Asyncable
     async::Notification m_stateChanged;
     async::Notification m_noteAdded;
 
+    EngravingItem* m_lastInputedItem = nullptr;
```
RomanPudashkin, line 95: Ms::InputState already stores the position of the last entered note, so this pointer looks redundant. Maybe we don't use it correctly?

HemantAntony: Oh okay. I didn't know about that

PR #9593, opened 2021-10-27
Fix realtime noteentry 8th crash (fixes #321716)

fxthomas:
Resolves: https://musescore.org/en/node/321716

To reproduce (on MuseScore 3.6.2):

* Connect MIDI keyboard (via JACK/Pipewire in my case)
* Enter Realtime (Manual) note input mode
* Select "8th notes"
* Enter 1 note and hit the manual input button 4 times on the MIDI keyboard, the 4th makes MuseScore crash

----

- [?] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [?] I created the test (mtest, vtest, script test) to verify the changes I made

PR #10199, opened 2021-12-29
[MU4] Fix #321716: Realtime note entry 8th crash

Jojo-Schmitz:
Port of #9593 (and is included in #9000 too)

Resolves: https://musescore.org/en/node/321716

### DIFF
--- a/src/engraving/libmscore/noteentry.cpp
+++ b/src/engraving/libmscore/noteentry.cpp
@@ -203,7 +203,7 @@ Note* Score::addPitch(NoteVal& nval, bool addFlag, InputState* externalInputStat
         //   We could split the duration at the barline and continue into the next bar, but this would create extra
         //   notes, extra ties, and extra pain. Instead, we simply truncate the duration at the barline.
         Fraction ticks2measureEnd = is.segment()->measure()->ticks() - is.segment()->rtick();
-        duration = is.duration() > ticks2measureEnd ? ticks2measureEnd : is.duration().fraction();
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

src/engraving/libmscore/noteentry.cpp
```diff
@@ -203,7 +203,7 @@ Note* Score::addPitch(NoteVal& nval, bool addFlag, InputState* externalInputStat
         //   We could split the duration at the barline and continue into the next bar, but this would create extra
         //   notes, extra ties, and extra pain. Instead, we simply truncate the duration at the barline.
         Fraction ticks2measureEnd = is.segment()->measure()->ticks() - is.segment()->rtick();
-        duration = is.duration() > ticks2measureEnd ? ticks2measureEnd : is.duration().fraction();
+        duration = is.duration().fraction() > ticks2measureEnd ? ticks2measureEnd : is.duration().fraction();
```
worldwideweary, line 206: This is where I was saying I would imagine that `.ticks()` instead of `.fraction() `  would be more appropriate here, but I've not found a problem with Note Entry through a MIDI device in master branch as with #9000...

Jojo-Schmitz: As per https://github.com/Jojo-Schmitz/MuseScore/commit/3681b72032f35e731ff3e2bc933528a9ae594777#commitcomment-75143487 this has been a false alert 😉

PR #10200, opened 2021-12-29
add support for beam breaks

Nick-Mazuk:
Adds back in support for beam breaks. Behavior should be identical to v3.6.

Supersedes https://github.com/musescore/MuseScore/pull/9979

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -672,7 +672,7 @@ void Beam::createBeamSegment(Chord* startChord, Chord* endChord, int level)
         );
 }
 
-bool Beam::calcIsBeamletBefore(Chord* chord, int i, int level) const
+bool Beam::calcIsBeamletBefore(Chord* chord, int i, int level, bool isAfter32Break, bool isAfter64Break) const
 {
     // if first or last chord in beam group
     if (i == 0) {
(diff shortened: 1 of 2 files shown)


Nick-Mazuk:
vtests: looking at the results, the vtests should fail. However, not all failed as expected and I'm working on a fix for those.

Nick-Mazuk:
More debugging for the vtests needed

Nick-Mazuk:
Vtests look good for this PR.

There are still a few things that should be fixed, but those are for separate PRs and not caused by this one.

Nick-Mazuk:
Ready to merge unless anyone has comments.

oktophonie (approved):

oktophonie:
Just to belatedly explain what this PR is for (!):

In MU3.6 (and before) inner beams are broken according to the note groupings defined for the time signature. For instance, in a _x_/4 bar, this was the result:
image
where the break of the inner beams between the 4th and 5th notes is automatic, not manually specified via the corresponding Beam Properties palette.

With the beam code rewrite in MU4 this functionality was lost, and as such the above example looks like this:
image


This PR is meant to reinstate the behaviour from MU3.6. As such, the changes in vtest are to be expected.

DmitryArefiev:
Tested on Win10, Mac12 - Looks good      

Please merge into master

DmitryArefiev (approved):

PR #10203, opened 2021-12-31
Fix #7654: Added ability to hide multiple instruments at once

### LINKED ISSUE #7654: [MU4 Issue] If multiple instruments selected and click to hide them, only one item hidden

HemantAntony:
Resolves: #7654
Closes: #7940

This PR solved #7654 and also adds the ability to select multiple instruments using Ctrl

### DIFF
--- a/src/instrumentsscene/view/instrumentspaneltreemodel.cpp
+++ b/src/instrumentsscene/view/instrumentspaneltreemodel.cpp
@@ -414,6 +414,21 @@ bool InstrumentsPanelTreeModel::moveRows(const QModelIndex& sourceParent, int so
     return true;
 }
 
+void InstrumentsPanelTreeModel::toggleVisibilityOfSelectedRows(bool visible)
+{
+    if (!m_selectionModel || !m_selectionModel->hasSelection()) {
+        return;
+    }
+
+    QModelIndexList selectedIndexes = m_selectionModel->selectedIndexes();
+
+    for (QModelIndex index : selectedIndexes) {
+        AbstractInstrumentsPanelTreeItem* item = modelIndexToItem(index);
+
+        item->setIsVisible(visible);
+    }
+}
+
 QItemSelectionModel* InstrumentsPanelTreeModel::selectionModel() const
 {
     return m_selectionModel;
--- a/src/instrumentsscene/qml/MuseScore/InstrumentsScene/InstrumentsPanel.qml
+++ b/src/instrumentsscene/qml/MuseScore/InstrumentsScene/InstrumentsPanel.qml
@@ -276,6 +276,10 @@ Item {
                                 onPopupClosed: {
                                     flickable.contentY = contentYBackup
                                 }
+
+                                onVisibilityChanged: function(visible) {
+                                    instrumentsTreeModel.toggleVisibilityOfSelectedRows(visible);
+                                }
                             }
                         }
 
--- a/src/instrumentsscene/qml/MuseScore/InstrumentsScene/internal/InstrumentsTreeItemDelegate.qml
+++ b/src/instrumentsscene/qml/MuseScore/InstrumentsScene/internal/InstrumentsTreeItemDelegate.qml
@@ -49,6 +49,8 @@ FocusableControl {
     signal popupOpened(var popupX, var popupY, var popupHeight)
     signal popupClosed()
 
+    signal visibilityChanged(bool visible)
+
     QtObject {
         id: prv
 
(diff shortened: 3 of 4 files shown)


src/instrumentsscene/view/instrumentspaneltreemodel.cpp
```diff
@@ -48,13 +48,10 @@ static QString notationToKey(const INotationPtr notation)
 }
 
 InstrumentsPanelTreeModel::InstrumentsPanelTreeModel(QObject* parent)
-    : QAbstractItemModel(parent)
+    : QAbstractItemModel(parent), m_selectionModel(new ItemMultiSelectionModel(this))
 {
     m_partsNotifyReceiver = std::make_shared<async::Asyncable>();
 
-    m_selectionModel = new ItemMultiSelectionModel(this);
-    m_selectionModel->setAllowedModifiers(Qt::ShiftModifier);
```
cbjeukendrup, line 56: Does moving instruments up and down still work correctly if there is a non-contiguous selection?

HemantAntony: It didn't at that time. It got all messed up.

I don't know about the present condition. But I think I will remove that commit. I don't think the problem of not being able to select instrument is too big (@Tantacrul hope you agree)

src/instrumentsscene/view/instrumentspaneltreemodel.cpp
```diff
@@ -414,6 +411,31 @@ bool InstrumentsPanelTreeModel::moveRows(const QModelIndex& sourceParent, int so
...
+    bool toVisible = true;
+
+    for (QModelIndex index : selectedIndexes) {
+        AbstractInstrumentsPanelTreeItem* item = modelIndexToItem(index);
+        item->isVisible() ? visibleItems++ : hiddenItems++;
+    }
+
+    if (visibleItems > hiddenItems) {
+        toVisible = false;
+    }
```
cbjeukendrup, line 431: Would it be an idea to base the new visibility on the old visibility of the instrument that was clicked, instead of on the number of visible selected instruments?

i.e. If there are two instruments selected, and the first is visible, but the second not, and you click on the eye icon for the first instrument, then both instruments become invisible, because the first instrument was visible.
And if you'd click the eye icon for the second instrument, both instruments become visible, because the second instrument was not visible. 

I think that would feel more predictable / easier for the user to figure out what the pattern is.

src/instrumentsscene/view/instrumentspaneltreemodel.cpp
```diff
@@ -414,6 +411,31 @@ bool InstrumentsPanelTreeModel::moveRows(const QModelIndex& sourceParent, int so
     return true;
 }
 
+void InstrumentsPanelTreeModel::changeVisibilityOfSelectedRows()
```
cbjeukendrup, line 414: Tiny thing: I would call it <code><b>toggle</b>VisibilityOfSelectedRows</code>

cbjeukendrup:
@HemantAntony Tiny remark: the commit message of the first commit mentions the wrong issue number. So if you are going to force-push this once more, it'd be great if you could fix that too.

src/instrumentsscene/view/instrumentspaneltreemodel.cpp
HemantAntony: Okay that seems like a better idea

src/instrumentsscene/view/instrumentspaneltreemodel.cpp
cbjeukendrup: I believe I had long ago created an algorithm for moving non-contiguous selections up and down. It was aimed at the parts dialog (never got merged into there though), but could probably be adapted for the instruments panel. I'll try to reuse it after this PR gets merged. But for now, let's indeed drop the commit from this PR.

HemantAntony:
Also linked #7940 to be closed if this PR gets accepted

Tantacrul:
I've tested it. Working great.

Tantacrul:
@RomanPudashkin - same question :)

Are you happy with this solution? If so, it is working well from a functional point of view

RomanPudashkin:
> @RomanPudashkin - same question :)
> 
> Are you happy with this solution? If so, it is working well from a functional point of view

Yes, I'm happy! Let's merge it :)

PR #10206, opened 2022-01-04
[MU4] Turn `enum : char` into `enum : signed char` if it contains negative values

Jojo-Schmitz:
Needed for all platforms/compilers/implementations where `char` is `unsigned char` by default, like e.g. ARM

### DIFF
--- a/src/engraving/libmscore/textbase.h
+++ b/src/engraving/libmscore/textbase.h
@@ -46,7 +46,7 @@ enum class FrameType : char {
 //   VerticalAlignment
 //---------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


igorkorsukov:
@Jojo-Schmitz  What do you think about omitting the type? (to be int by default)

jeetee:
I vote for omitting a storage type if we don't care about it (which we usually shouldn't).

Jojo-Schmitz:
It seemed to have been a good idea back when I did all this, leading to a significantly smaller memory footprint.

jeetee:
Memory footprint could be a reason to care about it..

PR #10207, opened 2022-01-04
Fix #10019: set min height/width for dock panels

### LINKED ISSUE #10019: [MU4 Issue] Dock panels have no minimum height (especially problematic when they are undocked) [Good first issue]
**Describe the bug**
"Vertical" dockable panels (Palettes, Instruments, etc.) have no minimum height. This is not very noticeable when they are docked, but becomes a real problem when they are floating. 

Similarly, horizontal panels (Mixer, etc.) have no minimum width. 

https://user-images.githubusercontent.com/48658420/145260710-68430c31-b250-4d67-8840-c0f7336f3ed7.mov

**Proposed solution**
I would suggest setting a minimum height of 200 pixels, for all four vertical panels. All panels become scrollable if their content is taller than themselves, so it really doesn't need to be much. With a height of 200, they are still usable (not very comfortable, but users with a small screen may have no choice).
For the horizontal panels, I would suggest a minimum width of 300 px.

Hint: See NotationPage.qml.

patrickmartinasin:
Resolves: #10019

Added minimum height of 200px to vertical dock panels and minimum width of 300px to horizontal dock panels, as suggested in the issue.

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -91,8 +91,10 @@ DockPage {
     }
 
     readonly property int verticalPanelDefaultWidth: 300
-    readonly property int verticalPanelMinimumHeight: 200
-    readonly property int horizontalPanelMinimumWidth: 300
+    readonly property int verticalPanelMinimumHeightFloating: 200
+    readonly property int verticalPanelMinimumHeightDocked: 50
+    readonly property int horizontalPanelMinimumWidthFloating: 300
+    readonly property int horizontalPanelMinimumWidthDocked: 100
 
     readonly property int horizontalPanelMinHeight: 100
     readonly property int horizontalPanelMaxHeight: 520
@@ -223,7 +225,7 @@ DockPage {
             width: root.verticalPanelDefaultWidth
             minimumWidth: root.verticalPanelDefaultWidth
             maximumWidth: root.verticalPanelDefaultWidth
-            minimumHeight: root.verticalPanelMinimumHeight
+            minimumHeight: floating ? root.verticalPanelMinimumHeightFloating : root.verticalPanelMinimumHeightDocked
 
             groupName: root.verticalPanelsGroup
 
@@ -249,7 +251,7 @@ DockPage {
             width: root.verticalPanelDefaultWidth
             minimumWidth: root.verticalPanelDefaultWidth
             maximumWidth: root.verticalPanelDefaultWidth
-            minimumHeight: root.verticalPanelMinimumHeight
+            minimumHeight: floating ? root.verticalPanelMinimumHeightFloating : root.verticalPanelMinimumHeightDocked
 
             groupName: root.verticalPanelsGroup
 
@@ -277,7 +279,7 @@ DockPage {
             maximumWidth: root.verticalPanelDefaultWidth
 
             groupName: root.verticalPanelsGroup
-            minimumHeight: root.verticalPanelMinimumHeight
+            minimumHeight: floating ? root.verticalPanelMinimumHeightFloating : root.verticalPanelMinimumHeightDocked
             
             dropDestinations: root.verticalPanelDropDestinations
 
@@ -298,7 +300,7 @@ DockPage {
             width: root.verticalPanelDefaultWidth
             minimumWidth: root.verticalPanelDefaultWidth
             maximumWidth: root.verticalPanelDefaultWidth
-            minimumHeight: root.verticalPanelMinimumHeight
+            minimumHeight: floating ? root.verticalPanelMinimumHeightFloating : root.verticalPanelMinimumHeightDocked
 
             groupName: root.verticalPanelsGroup
 
@@ -325,7 +327,8 @@ DockPage {
             height: 368
             minimumHeight: root.horizontalPanelMinHeight
             maximumHeight: root.horizontalPanelMaxHeight
-            minimumWidth: root.horizontalPanelMinimumWidth
+
+            minimumWidth: floating ? root.horizontalPanelMinimumWidthFloating : root.horizontalPanelMinimumWidthDocked
 
             groupName: root.horizontalPanelsGroup
 
@@ -360,7 +363,7 @@ DockPage {
             height: 200
             minimumHeight: root.horizontalPanelMinHeight
             maximumHeight: root.horizontalPanelMaxHeight
-            minimumWidth: root.horizontalPanelMinimumWidth
+            minimumWidth: floating ? root.horizontalPanelMinimumWidthFloating : root.horizontalPanelMinimumWidthDocked
 
             groupName: root.horizontalPanelsGroup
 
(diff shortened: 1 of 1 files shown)


patrickmartinasin:
My first ever pull request on GitHub, so I hope I've done it correctly!

src/appshell/qml/NotationPage/NotationPage.qml
```diff
@@ -91,6 +91,8 @@ DockPage {
     }
 
     readonly property int verticalPanelDefaultWidth: 300
+    readonly property int verticalPanelMinimumHeight: 200
```
RomanPudashkin, line 94: I think we should have different minimum values for floating and docked panels. This value may be too big for vertically stacked panels. For example, the status bar disappears if I place the panels vertically on my screen (because there is not enough free vertical space for it)

Screenshot 2022-01-10 at 11 06 34

RomanPudashkin: @bkunda could you advise us on the most appropriate min/max sizes for vertical/horizontal panels, depending on whether they are docked or not?

patrickmartinasin: Good point. I've added another commit to allow this. I've gone for min height docked of 50px and min width docked of 100px, but can easily change if advised.

duanjr:
I did the same work . But when the panels are floating, they still have no minimum height. I wonder if I did something wrong or just these changes cannot set minimum height when they are floating. Did your local version perform correctly?

bkunda:
I actually don't see any issue with @cbjeukendrup's suggestion of minimum height 200px. 

I've tested it on the smallest monitor I have (13") at a resolution of 1440x900, and 200px conveniently shows just the right amount of the properties panel (when stacked) to partially reveal the presence of important functionalities (to which the user must scroll for access, but at least they can still see they are there):

Screenshot 2022-01-11 at 5 14 13 pm

I'm also unable to reproduce any issues with the status bar; on both Mac and Windows the panels appear to be drawn behind the status bar, so it is always visible (unless I missed something here, @RomanPudashkin?)

Moreover, 200px looks quite nice when panels are undocked. The user can see 2 complete rows of palette items which, especially if they've set up a custom palette, might be all they need for working on a project with a small screen:

Screenshot 2022-01-11 at 5 16 36 pm

The subject of this issue mentions "width", however I'm not sure we can even adjust the width of these panels in MU4 (I certainly can't, despite the presence of resize handles). I also can't see any responsive design work in Figma concerning width adjustment, so perhaps this is not even yet a feature (perhaps @jessjwilliamson can confirm this). 

Do let me know if I've missed something here (quite likely 😅). Otherwise, I'd go with 200px for minimum height for all panels. When vertically stacked, the maximum any given panel could go would therefore be whatever space is left minus the 200px of each stacked panel (if that makes sense...).

bkunda:
Now as for those horizontal panels 😀:

**Note input toolbar**
On MacOS, I can't even resize this panel at the moment (same as for playback controls). This is perhaps a bug.
I can resize in Windows – to the point of making the entire panel disappear! But I do note that there is actually some responsive design work for this in Figma that doesn't seem to have been implemented yet: https://www.figma.com/file/l0TQUrNFZmEfrZnfHhFDfb/?node-id=2871%3A0

Not sure how old or valid these designs are, so perhaps @jessjwilliamson could confirm. But they do seem to make sense to me. It would mean that the minimum width would be something in the order of 400px (wide enough to take in the `augmentation dot` button.

**Playback panel**
I can't see any design work for this one, but the panel is at any rate not particularly wide, so perhaps it doesn't need to be resizable at all.

bkunda:
> > 
> 
> @bkunda Those different width versions of the Note input bars are supposed to be demonstrated if they were docked into Musescore when in a narrow app window - for example, how will it respond if in 13" screen. It wasn't intentional for the bar to be resized actually (especially vertically!! which I know is a bug in Windows). Play panel doesn't resize either.

That's good to know @jessjwilliamson. 
@RomanPudashkin, I didn't actually find the width of the note input toolbar to be a problem on my small screen, so if it was never meant to be resized, then I'd say we don't need to specify a min-width (but we should disable resizing on Windows – and Linux too, if it is an issue there).

bkunda:
> > > 
> > 
> > 
> > @bkunda Those different width versions of the Note input bars are supposed to be demonstrated if they were docked into Musescore when in a narrow app window - for example, how will it respond if in 13" screen. It wasn't intentional for the bar to be resized actually (especially vertically!! which I know is a bug in Windows). Play panel doesn't resize either.
> 
> That's good to know @jessjwilliamson. @RomanPudashkin, I didn't actually find the width of the note input toolbar to be a problem on my small screen, so if it was never meant to be resized, then I'd say we don't need to specify a min-width (but we should disable resizing on Windows – and Linux too, if it is an issue there).

...but maybe check to see if you can put your hands on a more recent PDF spec of the note input toolbar (too see if it has anything in it about responsiveness). This was before my time at MS so all I have to go by are the Figma files :-)

Tantacrul:
@bkunda - If I'm reading you right, this PR looks good to go? Do you approve it?

If so, then let's resolve the conflicts and get it merged.

bkunda:
> @bkunda - If I'm reading you right, this PR looks good to go? Do you approve it?
> 
> If so, then let's resolve the conflicts and get it merged.

I can't actually test this because there is nothing in the "Checks" tab (and no artifacts I can find either). 

Am I missing something obvious here? 😕

cbjeukendrup:
The problem is that the builds have expired. 
@patrickmartinasin Could you please rebase and force-push, to resolve the conflict and to get new builds?

patrickmartinasin:
> The problem is that the builds have expired. @patrickmartinasin Could you please rebase and force-push, to resolve the conflict and to get new builds?

I'm sorry, but how do I do this, please?

HemantAntony:
Correct me if I'm wrong, but you should open terminal and then type `git pull --rebase upstream master`. You will have conflicts that should be resolved. Then you should do `git push -f origin branch_name`

patrickmartinasin:
> The problem is that the builds have expired. @patrickmartinasin Could you please rebase and force-push, to resolve the conflict and to get new builds?

OK, I think I've done that now.

bkunda:
I don't think this is ready to be merged yet, sorry.
A few issues I've spotted: 

1. Min. panel size doesn't seem to apply when panels are docked, so there is still an undesirable overlap of elements:

https://user-images.githubusercontent.com/86290556/171385137-74c69b1a-fd4b-46de-9ae4-f855deafef3e.mov

2. Mixer resize now seems broken: it is no longer possible to expand the width of the mixer

https://user-images.githubusercontent.com/86290556/171385213-72e66993-0c29-4272-8319-c9d9569ce6bb.mov

_(Compare with latest nightly):_

https://user-images.githubusercontent.com/86290556/171386219-58372fee-6d1c-4ede-91c3-197b2b7da517.mov

RomanPudashkin:
Closing as there is a [new attempt](https://github.com/musescore/MuseScore/pull/22136) to fix this bug...

PR #10211, opened 2022-01-04
Fix #328003 score property crash

jonnymoo:
Resolves: https://musescore.org/en/node/328003

Following accessibility changes saving score properties causes MuseScore to crash

- [ x] I signed [CLA](https://musescore.org/en/cla)
- [ x] I made sure the code compiles on my machine
- [ x] I made sure there are no unnecessary changes in the code
- [ x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [ x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [ x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ x] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/src/notation/view/widgets/scoreproperties.cpp
+++ b/src/notation/view/widgets/scoreproperties.cpp
@@ -28,6 +28,7 @@
 #include "ui/view/widgetutils.h"
 
 #include "translation.h"
+#include "notation/internal/scorepropertiesitem.h"
 
 using namespace mu::notation;
 using namespace mu::project;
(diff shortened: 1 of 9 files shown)


src/notation/view/widgets/scoreproperties.cpp
```diff
@@ -248,31 +249,31 @@ bool ScorePropertiesDialog::save()
         QLayoutItem* tagItem   = scrollAreaLayout->itemAtPosition(i, 0);
         QLayoutItem* valueItem = scrollAreaLayout->itemAtPosition(i, 1);
         if (tagItem && valueItem) {
-            QLineEdit* tag   = static_cast<QLineEdit*>(tagItem->widget());
-            QLineEdit* value = static_cast<QLineEdit*>(valueItem->widget());
 
-            QString tagText = tag->text();
+            ScorePropertiesItem item(tagItem->widget(), valueItem->widget());
```
RomanPudashkin, line 253: This class is a bit redundant. You can just add missing nullptr checks where the crash occurs

jonnymoo: Is that why the test runs have all failed except on windows? Have I added the file incorrectly?

I split it out into a new class to a) fullfill the single responsibility principle of SOLID and b) it made it much easier to unit test, but if it creates other issues then yes of course just doing the cast check in the score properties page will work. 

I’ll have a think how I can TDD that…

jonnymoo: Ps I think you are correct btw - I’ll go and simplify.

RomanPudashkin: These principles are good but in this case, your solution overcomplicates the dialog implementation. In addition, the dialog will be rewritten using the qml / c++ pair in the future (or completely replaced with something else). Anyway, thanks for your work!

jonnymoo: Simple fix as per above on way....

PR #10212, opened 2022-01-05
fix #10202: Added a check on whether zoom has already been initialized

### LINKED ISSUE #10202: [MU4 Issue] Saving causes zoom to go to 100%
**Describe the bug**

When saving with the keyboard shortcut `⌘S`, the screen zooms to 100%, regardless of the current zoom level.

**To Reproduce**
1. Zoom on the score to anything aside from 100% (doesn't matter how, either using the icons, the menu bar, pinch to zoom, or `⌘+` and `⌘-`)
2. Use the keyboard shortcut `⌘S` to save the file
3. The zoom level changes

**Expected behavior**

Saving with the keyboard shortcut should not affect the zoom level.

**Screenshots**

https://user-images.githubusercontent.com/5659171/147709762-52ab22b2-7462-47ca-8f08-70911bb352b3.mov

 - OS: macOS 12

.1, Apple Silicon

**Additional context**

I'm unsure if this is a macOS specific bug. It has been present for at least a week or two now and I've been forgetting to add a bug report for this.

VanSHOE:
Resolves: #10202 

I added a check on whether zoom has already been initiated instead of setting it to false everytime, as this is what is called upon save. Another solution is to just remove this call on the save function, please let me know if that alternative seems better.

### DIFF
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -76,6 +76,11 @@ void NotationViewInputController::init()
         if (!m_isZoomInited) {
             initZoom();
         }
+
+        auto masterNotation = globalContext()->currentMasterNotation();
+        if (!masterNotation) {
+            m_isZoomInited = false;
+        }
     });
 }
 
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -73,7 +73,7 @@ void NotationViewInputController::init()
     }
 
     globalContext()->currentMasterNotationChanged().onNotify(this, [this]() {
-        if (m_isZoomInited == false) {
+        if (!m_isZoomInited) {
             initZoom();
         }
     });
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -73,8 +73,9 @@ void NotationViewInputController::init()
     }
 
     globalContext()->currentMasterNotationChanged().onNotify(this, [this]() {
-        m_isZoomInited = false;
-        initZoom();
+        if (m_isZoomInited == false) {
+            initZoom();
+        }
     });
 }
 


cbjeukendrup:
Good that you found out that this issue was caused by the `currentMasterNotationChanged` notification!

I'm not sure whether it is desired that the zoom is re-initialized when the master notation is changed. There is more work to do regarding zoom anyway, so I think your changes are fine for now. 

I'm also not sure whether it is desired that the `currentMasterNotationChanged` is fired whenever the notation is saved (see `src/project/internal/projectactionscontroller.cpp:562`). In fact, it stays the same MasterNotation (and the same NotationProject, etc.), but only the path changes. Maybe we should instead create a dedicated notification for this on the MasterNotation class, or better the NotationProject class. Maybe @Eism can comment on this; see b311ffeac49019e8a86412a33e28da351b158908 where this was implemented, and 534b5c3b978bfc409a86a29e1f31ae2860881321 where it was changed by @HemantAntony.

VanSHOE:
I see
So should I let this be?

I noticed another thing which was then when I save after the file already exists, the filePath is still just ""
image
Which causes this test to ALWAYS fail which I feel was not intentional?

cbjeukendrup:
Hm, yeah, there you've found another problem. I was already wondering about this. It seems that `currentNotationProject()->metaInfo().filePath` gets the path from `Ms::MasterScore::fileInfo()`, and I wonder whether that path is ever updated. So therefore, it will be `""`. 

I think some refactoring needs to be done; a good first step would be to move the QFileInfo from `Ms::MasterScore` to `mu::engraving::EngravingProject`. If you feel comfortable to dive into this, feel free to do so; otherwise, For one of the team members could do it.

But back to this pull request: I'd propose to leave it as it is for now, and wait for feedback from the team. If you like, you can also try implementing the dedicated "path changed" notification I mentioned in my previous comment.

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -73,8 +73,9 @@ void NotationViewInputController::init()
     }
 
     globalContext()->currentMasterNotationChanged().onNotify(this, [this]() {
-        m_isZoomInited = false;
-        initZoom();
+        if (m_isZoomInited == false) {
```
Jojo-Schmitz, line 76: Maybe better `if (!m_isZoomInited) {`?

VanSHOE: On it

Jojo-Schmitz: Seems better in sync with
```
    if (viewport().isValid() && !m_inputController->isZoomInited()) {
        m_inputController->initZoom();
```
in notationpaintview.cpp, which is the only (other) place where this is checked (and only then initialized)

Eism:
I think it should be

```
    globalContext()->currentMasterNotationChanged().onNotify(this, [this]() {
        if (!m_isZoomInited) {
            initZoom();
        }

        auto masterNotation = globalContext()->currentMasterNotation();
        if (!masterNotation) {
            m_isZoomInited = false;
        }
    });
```
Previously, the zoom was initialized when opening the score page(NotationPaintView::onViewSizeChanged) and changing the master notation.

Your solution will not work in the case:
1. Open project(the score page will open)
2. Close project
3. Open a new project from File->Open recent->project

The zoom should reset to default

So, let's reset the zoom on close project

VanSHOE:
Do note that after this commit, a crash has started happening as mentioned here https://github.com/musescore/MuseScore/issues/10233

Eism:
@VanSHOE I can't reproduce the crash on Windows(after applying changes from this PR).

Please give me more information(detailed steps, stacktrace, video) so that I can help you.

VanSHOE:
@Eism I apologize, a lot of work has me occupied for about 2 weeks, I will get on this again ASAP

VanSHOE:
@Eism Here, I have made a video on how to do the crash
The version of the program is the same that I made my branch on 2 weeks ago

https://user-images.githubusercontent.com/75690289/151364681-49ed1f6c-7920-4b95-9351-071645f2d45d.mp4

PR #10216, opened 2022-01-06
fix #10189: Autosave dedicated file

### LINKED ISSUE #10189: [MU4 Issue] Auto save overwrites existing score
**Describe the bug**
Auto save saves **into** the existing score instead a dedicated backup score

**To Reproduce**
1. Open any existing score.
2. Make any change (adding/changing note, whatever). Now the score is marked with an star.
3. Wait until auto-save kicks in.
4. Now the star is gone, indicating the score is saved. This shouldn't happen!
5. Close the score (File -> Close), the scores closes without the "Do you want to save changes tot the score..." dialog.
6. Open the score again, the previous made changes are stored. However, the original score is expected (as load in step 1).

**Expected behavior**
Auto save should save the score into a dedicated score file. Only after an explicit save (using the "Save" command) the score should be saved in the original file.
The original file should **never** be overwritten by the auto-save.

**Screenshots**
n.a.

 - Linux

**Additional context**
Now the behavior of auto-save is changed with respected to MU3 (and (almost?) other program).

VanSHOE:
Resolves: #10189 

Instead of overwriting the same file, the program now makes a new file and does not modify the original file on its own
I added another save type called autosave, which I felt would help distinguish this save later on if someone wants to work on this feature later on (I tried using Savecopy before, but it felt stranger hence made another type)

Note this requires that the user should NEVER have two files of the name \<name\>.mscz and autosave_\<name\>.mscz as the program would OVERWRITE the latter treating it as an autosave file without warnings

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -373,12 +373,16 @@ mu::Ret NotationProject::saveScore(const io::path& path, SaveMode saveMode)
         m_engravingProject->setPath(path.toQString());
     }
 
+    std::string auto_path = m_engravingProject->path().toStdString();
+    std::string name = m_engravingProject->title();
+    auto_path.erase(auto_path.end() - name.length() - 5, auto_path.end());
+    auto_path += ("autosave_" + name + ".mscz");
+
     if (saveMode == SaveMode::Autosave) {
-        std::string original_path = m_engravingProject->path().toStdString();
-        std::string name = m_engravingProject->title();
-        original_path.erase(original_path.end() - name.length() - 5, original_path.end());
-        original_path += ("autosave_" + name + ".mscz");
-        m_engravingProject->setPath(QString::fromStdString(original_path));
+        m_engravingProject->setPath(QString::fromStdString(auto_path));
+    }
+    else {
+        remove(auto_path.c_str());
     }
     Ret ret = doSave(true);
     if (!ret) {
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -336,7 +336,8 @@ mu::Ret NotationProject::save(const io::path& path, SaveMode saveMode)
     case SaveMode::Save:
     case SaveMode::SaveAs:
     case SaveMode::SaveCopy:
-        return saveScore(path);
+    case SaveMode::Autosave:
+        return saveScore(path, saveMode);
     }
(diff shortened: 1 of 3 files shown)


VanSHOE:
Please note that I have tested this only on Windows till now

njvdberg:
Thanks :+1: 
Two questions (at the moment I'm not able to check myself)

1. Is the auto save file removed after a regular save of the score file?
2. If a score file is opened **and** an auto save file exists, will there be a question whether the auto save file should be opened? This is the behavior in MU3 (and many other applications I know).

VanSHOE:
The first part I have done now, please do verify if it works (only tested on windows right now)

The second part is not there. I feel it is best to have an autosave that does not bother the user much. If the user thinks they have made a mistake they can simply rename the autosave to their original name (kind of like how blender handles its .blend1 etc. files)

bkunda:
Hey @VanSHOE, I've had a look at this on both Mac and Win and cannot seem to restore the autosaved file. I expect this might have something to do with the fact that we haven't yet ported over a dialog that prompts the user to actually load the autosaved file – this is something I'll log in a separate issue.
@vpereverzev it might be worth you taking a look at this as it might relate to our discussion about auto save from yesterday :-)

VanSHOE:
Hey, by restore do you mean after you do a normal save? Because then the autosave file is deleted.
Secondly, the autosave file is separate to the actual project, and you have to open it up separately. Or does the file not exist at all?

bkunda:
> Hey, by restore do you mean after you do a normal save? Because then the autosave file is deleted. Secondly, the autosave file is separate to the actual project, and you have to open it up separately. Or does the file not exist at all?

No, the way I tested it was to do a normal save, then make a bunch of changes and wait for the auto save to kick in, then restart my computer and reboot MuseScore. In the current nightlies, reopening the file from Home>Score preserves the autosaved changes, but this isn't the case in this build. In 3.6, the user has to "OK" a "Do you want to restore session" dialog, which will then open the autosaved copy.

njvdberg:
> In 3.6, the user has to "OK" a "Do you want to restore session" dialog, which will then open the autosaved copy.

For me, this would be the preferred behavior.

Eism:
@VanSHOE 
There are some problems with your approach here. So it has been resolved in a separate PR #10273.

Anyway, thanks for your work!

PR #10014, opened 2021-12-08
Fix #8172: Replaced instruments updated

HemantAntony:
Resolves: [#8172](https://github.com/musescore/MuseScore/issues/8172)

This commit now updates the instruments panel and parts dialog to
display the part name of new instrument.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [ ] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

Issue #10128, opened 2021-12-18
[MU4 Task] Clean up the work with `Instruments`, `Parts` and `Excerpts`, their names/titles, and Instrument changes


cbjeukendrup:
**Terms used in the code**
To prevent misunderstandings, let's recall some definitions from earlier discussions. The monospace font indicates that I refer to the term as used in the code, while the normal font is used to refer to daily-life terms that are (or should be) also used in the UI.
- An `Instrument` is a representation of an instance of a physical instrument, like Flute, Piccolo, Violin, etc.. Each `Instrument` has a name.
- A `Part` has the same role as a Player in Dorico; i.e. it represents a person (or a group of persons playing the same voice) that plays one `Instrument` at a time, but may switch to another `Instrument` during the piece. Examples of `Parts` are a violin player, a violins section, and a flutist who switches to the piccolo for some bars. In daily life, one may say that a certain score has three instruments, but in the code, we would say that that `Score` has three `Parts`. Each `Part` has a name.
- An `Excerpt` may be regarded the same way as a Layout in Dorico. It represents a score that contains only a selection of the `Parts` from the full score (`MasterScore`). In daily life, we tend to refer to `Excerpts` as "parts" (which we also do in the UI). Each `Excerpt` has a name.
  The selection of `Parts` may technically be empty, but in practice, this serves little purpose.
  In by far the most cases, an `Excerpt` contains only one `Part`, and the title of the `Excerpt` equals the name of the `Part`. 
  However, `Excerpts` containing more than one `Part` are not uncommon. For instance, for a piece for orchestra and choir, one may want to create an `Excerpt` containing only the four choir `Parts`. Or, for a violin sonata, the pianist will probably like to see the violin `Part` printed in his `Excerpt` too (but in this case, the full score could also have been used).

**No one-to-one relations**
It is important to realize that there is _not_ necessarily a one-to-one relation between `Instruments` and `Parts`, nor between `Parts` and `Excerpts`. 
- An `Instrument` is played in only one `Part`;
- A `Part` may be performed on one or more `Instruments` alternately;
- A `Part` is printed in one or more `Excerpts`;
- An `Excerpt` displays zero or more `Parts`.

In fact, there is nothing new or difficult about this, but we should make sure to keep this in mind always. Currently, there are quite some places in MU4 where we assume one-to-one relations, and many of them actually stem from MU3.

**Problems**
[...]

Issue #10122, opened 2021-12-17
[MU4 Issue] Instrument name does not update in parts dialog and its tab on replace


HemantAntony:
(Updated by @Tantacrul, @DmitryArefiev )
**Describe the bug**
When opening some older scores, the 'Replace Instrument' feature does not work. However, it **does** work for new scores.

BUG№1 - **FIXED** in #13097
To Reproduce:
1. Open  [string quartet 2.zip](https://github.com/musescore/MuseScore/files/9594429/string.quartet.2.zip)
2. Open Violin I part
3. Go to score and replace the Violin I with another instrument, like a Flute, using the Instruments panel
4. Observe that the **instrument label in score/mixer** and **sounds do not update**.

https://user-images.githubusercontent.com/90187801/190918230-036284c9-7ec6-4971-9e1f-ade697bb5a60.mp4

BUG№2 (**STILL OCCURS**)
To Reproduce:
1. Open  [string quartet 2.zip](https://github.com/musescore/MuseScore/files/9594429/string.quartet.2.zip)
2. Open Violin I part
3. Go to score and replace the Violin I with another instrument, like a Flute, using the Instruments panel
4. Observe that the **instrument name in part, in its tab and in parts dialog do not update**.

https://user-images.githubusercontent.com/90187801/190918628-ed607a95-dde3-4545-a3eb-344fb7ff3031.mp4

**Expected behavior**
The instrument's name should be updated in the parts dialog and its tab

Initial video from bug report

https://user-images.githubusercontent.com/88136054/146490932-a6ced8f6-6d20-4c0c-9d38-c2142dcf0a55.mp4

**Desktop (please complete the following information):**
Win, Mac, Linux

HemantAntony:
I am working on this

Tantacrul:
I don't know how widespread this problem is anymore. It could just be an issue with my file, which was originally a MusicXML import into 3.6. Other comments welcome!

HemantAntony:
I can still reproduce this

Also if unclear, I'm no longer working on this. 5 months is too long :)

Tantacrul:
OK, I've moved it to alpha then if you are still experiencing it.

DmitryArefiev:
@RomanPudashkin Says there are two different issues here: 
* **one** about not updating sounds/name in Mixer 
* and **another** about not updating instrument name in part, in parts dialog and in part's tab

The **issue №1** is fixed in #13097 
The **issue №2** still happens 

[...]

PR #10221, opened 2022-01-07
Fix #9745: Changed instrument connection to instrument title and instrument name

### LINKED ISSUE #9745: [MU4 Issue] Instrument name doesn't change in instruments panel when replacing instrument
**Describe the bug**
When replacing an instrument, the instrument's name in the instrument panel doesn't update correctly.

https://user-images.githubusercontent.com/86290556/141954519-9d35c968-fae7-4a0a-87a6-cd41bcffa899.mp4

**Expected behavior**
1. Both the instrument panel label AND the part name should update with the name of the replacement instrument 
2. The instrument panel label should follow the name of the instrument in the main score, not the part name
3. The ability to create separate instrument names between the main score and the part should be retained

 - MacOS 11.6
 - Windows 11

HemantAntony:
Resolves: #9745

Prior to this commit, if you were to change the part name in the
instruments dialog, it would automatically get updated in the instrument
title (the text for the corresponding instrument button). However, with
this commit, this "connection" is changed to instrument name and the
instrument title, that is, if you now change the instrument name, the
same would be the instrument title. Also, this commit does not fix the
entire issue. Another problem mentioned in this issue had been fixed
earlier in a separate pull request #10014

### DIFF
--- a/src/instrumentsscene/view/parttreeitem.cpp
+++ b/src/instrumentsscene/view/parttreeitem.cpp
@@ -48,7 +48,7 @@ void PartTreeItem::init(const notation::Part* masterPart)
     }
 
     setId(part->id());
-    setTitle(part->partName().isEmpty() ? part->instrument()->name() : part->partName());
+    setTitle(part->instrument()->name());
     setIsVisible(visible);
     setIsEditable(partExists);
     setIsExpandable(partExists);
--- a/src/instrumentsscene/qml/MuseScore/InstrumentsScene/internal/InstrumentsTreeItemDelegate.qml
+++ b/src/instrumentsscene/qml/MuseScore/InstrumentsScene/internal/InstrumentsTreeItemDelegate.qml
@@ -389,7 +389,6 @@ Item {
                     popup = popupLoader.createPopup(instrumentSettingsComp, this)
 
                     item["partId"] = model.itemRole.id
-                    item["partName"] = model.itemRole.title
                     item["instrumentId"] = model.itemRole.instrumentId()
 
                 } else if (root.type === InstrumentsTreeItemType.STAFF) {
--- a/src/instrumentsscene/view/instrumentsettingsmodel.cpp
+++ b/src/instrumentsscene/view/instrumentsettingsmodel.cpp
@@ -42,13 +42,13 @@ void InstrumentSettingsModel::load(const QVariant& instrument)
     QVariantMap map = instrument.toMap();
     m_instrumentKey.partId = ID(map["partId"]);
     m_instrumentKey.instrumentId = map["instrumentId"].toString();
-    m_partName = map["partName"].toString();
 
     const Part* part = notationParts()->part(m_instrumentKey.partId);
     if (!part) {
         return;
     }
 
+    m_partName = part->partName();
     m_instrumentName = part->instrument()->name();
     m_instrumentAbbreviature = part->instrument()->abbreviature();
 


HemantAntony:
If anyone is wondering, the fix was only the change in parttreeitem.cpp. However, this would cause the part name to be the instrument title (and indirectly the instrument name). As a result, the other fixes were necessary

bkunda:
Tested in MacOS and Windows. FIXED.

src/instrumentsscene/view/parttreeitem.cpp
```diff
@@ -48,7 +48,7 @@ void PartTreeItem::init(const notation::Part* masterPart)
     }
 
     setId(part->id());
-    setTitle(part->partName().isEmpty() ? part->instrument()->name() : part->partName());
+    setTitle(part->instrument()->name().isEmpty() ? part->partName() : part->instrument()->name());
```
Jojo-Schmitz, line 51: I believe the intrument name can never be empty, only the part name can, at least that's how it was in 3.x IIRC

src/instrumentsscene/view/parttreeitem.cpp
```diff
@@ -48,7 +48,7 @@ void PartTreeItem::init(const notation::Part* masterPart)
     }
 
     setId(part->id());
-    setTitle(part->partName().isEmpty() ? part->instrument()->name() : part->partName());
+    setTitle(part->instrument()->name());
```
Jojo-Schmitz, line 51: Still the part name can be different from the instrument name, like when a part contains more than one instrument, can't it? Like I usually ha a "Choir" part, containing the "Women" and Men" 'instruments'.

HemantAntony: I didn't know you can have multiple instruments in a single part. When you say part, do you mean like this with the separate score:
image

Jojo-Schmitz: Yes, exactly.

bkunda:
I've had a look at this build and from my end, it achieves what it aims to do, which is make the instrument panel label match the "Name on main score" 👍🏻

instrument-name-fix

There still remain other issues, but I believe these are being discussed elsewhere.

To be clear, yes the part name should be able to be set as something different from the instrument name.
@HemantAntony, it might be worth you reading through #10128 (if you haven't already), which outlines some of the complications still to resolve re: instruments and parts 🙂

src/instrumentsscene/view/parttreeitem.cpp
HemantAntony: But, this works as intended. I think there might be a confusion regarding the naming conventions. Though we say part to refer to the scores, its apparently different in the source code

HemantAntony:
> I've had a look at this build and from my end, it achieves what it aims to do, which is make the instrument panel label match the "Name on main score" 👍🏻
> 
> instrument-name-fix
> 
> There still remain other issues, but I believe these are being discussed elsewhere.
> 
> To be clear, yes the part name should be able to be set as something different from the instrument name. @HemantAntony, it might be worth you reading through #10128 (if you haven't already), which outlines some of the complications still to resolve re: instruments and parts 🙂

I agree that the naming conventions have to finalized. Its really confusing 😕.

HemantAntony:
I have found another possible issue. When I change the instrument name in one part (part as in score), it does not get updated across all parts. This is probably an issue unless its intended

bkunda:
> I have found another possible issue. When I change the instrument name in one part (part as in score), it does not get updated across all parts. This is probably an issue unless its intended

@HemantAntony is this perhaps related to your other issue #10122?

HemantAntony:
> > I have found another possible issue. When I change the instrument name in one part (part as in score), it does not get updated across all parts. This is probably an issue unless its intended
> 
> @HemantAntony is this perhaps related to your other issue #10122?

I think it probably is. But I'm not exactly sure. So, I will add "possibly related to #10122" as part of additional context

HemantAntony:
> > I have found another possible issue. When I change the instrument name in one part (part as in score), it does not get updated across all parts. This is probably an issue unless its intended
> 
> @HemantAntony is this perhaps related to your other issue #10122?

I don't think so. It seems like a separate issue

DmitryArefiev:
Tested #9745 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

PR #10224, opened 2022-01-07
fix #10223: Save selection now works

### LINKED ISSUE #10223: [MU4 Issue] `Save selection...` does nothing
**Describe the bug**
`Save selection...` doesn't save a file. Seems to perform no function. 
This is currently working in 3.6, so it seems it might be a regression in MS4.

**To Reproduce**
1. Open or create a new score containing some engraved elements
2. Select a range of music (e.g. a few bars)
3. Select `File`>`Save selection...`
4. Choose your destination file location and click `Save`
No file will be created

**Expected behavior**
A new file containing only the selected measures should be created at the chosen file destination (Per MS3.6).

 - MacOS 11.6 (M1)
 - Windows 11

VanSHOE:
Resolves: #10223 
The save selection feature now works
There was a crash when nothing was selected, and I have also implemented a check for that in `src/engraving/libmscore/scorefile.cpp`
```cpp
// outdated, this check has been moved outside now
if (selectionOnly && _selection.staffStart() == _selection.staffEnd()) {
        return;
}
```
Please do verify that the changes made are correct and do not break anything, though I have tested the changes (on Windows) and will continue to test further and report back (if any issues) after creating the PR

Edit: It is quite late here right now but I feel like just returning might not be a very good solution, will have to look more into that

Edit_2: The check has been moved outside before the file is created. A new error code has been created for the same and a new function has been created which checks if the selection is empty.

### DIFF
--- a/src/engraving/libmscore/scorefile.cpp
+++ b/src/engraving/libmscore/scorefile.cpp
@@ -82,10 +82,6 @@ void Score::write(XmlWriter& xml, bool selectionOnly, compat::WriteScoreHook& ho
     // then some layout information is missing:
     // relayout with all parts set visible
 
-    if (selectionOnly && _selection.staffStart() == _selection.staffEnd()) {
-        return;
-    }
-
     QList<Part*> hiddenParts;
     bool unhide = false;
     if (styleB(Sid::createMultiMeasureRests)) {
--- a/src/engraving/libmscore/scorefile.cpp
+++ b/src/engraving/libmscore/scorefile.cpp
@@ -82,6 +82,10 @@ void Score::write(XmlWriter& xml, bool selectionOnly, compat::WriteScoreHook& ho
     // then some layout information is missing:
     // relayout with all parts set visible
 
+    if (selectionOnly && _selection.staffStart() == _selection.staffEnd()) {
+        return;
+    }
+
     QList<Part*> hiddenParts;
     bool unhide = false;
     if (styleB(Sid::createMultiMeasureRests)) {
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -522,12 +522,9 @@ mu::Ret NotationProject::saveSelectionOnScore(const mu::io::path& path)
     // Write project
     std::string suffix = io::suffix(info.fileName());
     MscWriter::Params params;
-    QString org_path = m_engravingProject->path();
-    m_engravingProject->setPath(path.toQString());
-    params.filePath = m_engravingProject->path();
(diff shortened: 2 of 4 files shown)


src/project/internal/notationproject.cpp
```diff
@@ -522,20 +522,24 @@ mu::Ret NotationProject::saveSelectionOnScore(const mu::io::path& path)
     // Write project
     std::string suffix = io::suffix(info.fileName());
     MscWriter::Params params;
+    QString org_path = m_engravingProject->path();
```
wizofaus, line 525: I don't think we're using snake_case variable names, try to stick to camelCase.  And personally I'd call it originalPath.  But it's not clear that you *should* have to change the path property for the project just to be able to save to a different filename - can't you just change params.filePath?

src/project/internal/notationproject.cpp
```diff
@@ -522,20 +522,24 @@ mu::Ret NotationProject::saveSelectionOnScore(const mu::io::path& path)
...
 
     MscWriter msczWriter(params);
-    Ret ret = writeProject(msczWriter, false);
+    Ret ret = writeProject(msczWriter, true);
 
     if (ret) {
         QFile::setPermissions(info.filePath(),
                               QFile::ReadOwner | QFile::WriteOwner | QFile::ReadUser | QFile::ReadGroup | QFile::ReadOther);
     }
     LOGI() << "success save file: " << info.filePath();
```
wizofaus, line 541: doesn't info.filePath( ) still refer to the original filename?

src/engraving/libmscore/scorefile.cpp
```diff
@@ -82,6 +82,10 @@ void Score::write(XmlWriter& xml, bool selectionOnly, compat::WriteScoreHook& ho
     // then some layout information is missing:
     // relayout with all parts set visible
 
+    if (selectionOnly && _selection.staffStart() == _selection.staffEnd()) {
+        return;
```
igorkorsukov, line 86: Maybe better check this condition before opening the file (NotationProject::saveSelectionOnScore or NotationProject::writeProject), so that an empty file is not created.
We can also return a "Nothing selected" error so that we can show the result to the user.

src/project/internal/notationproject.cpp
igorkorsukov: here's our code style info https://github.com/musescore/MuseScore/wiki/CodeGuidelines

igorkorsukov (approved):

DmitryArefiev:
Tested #10223 on Win10, Mac12, LinuxMint 20.2 - FIXED

PR #10231, opened 2022-01-10
[MU4] mpe_rendering_optimizations

vpereverzev:
- Optimized a performance with shallow-copy containers
- Replace redundant move operators with constant references
- Used BeatsPerSecond type instead of qreal


src/engraving/playback/metaparsers/internal/arpeggiometaparser.cpp
```diff
@@ -62,9 +62,9 @@ void ArpeggioMetaParser::doParse(const Ms::EngravingItem* item, const PlaybackCo
 
     mpe::ArticulationMeta articulationMeta;
     articulationMeta.type = type;
-    articulationMeta.pattern = ctx.profile->pattern(type);
+    articulationMeta.pattern = ctx.profile->at(type);
```
igorkorsukov, line 65: I believe the previous name was better

src/engraving/playback/playbackeventsrenderer.cpp
```diff
@@ -64,18 +64,24 @@ void PlaybackEventsRenderer::renderNoteEvents(const Ms::Chord* chord, const mpe:
...
-    ctx.nominalPositionTick = chord->tick().ticks();
-    ctx.nominalDurationTicks = chord->durationTypeTicks().ticks();
-    ctx.beatsPerSecond = chord->score()->tempomap()->tempo(ctx.nominalPositionTick).val;
-    ctx.nominalTimestamp = timestampFromTicks(chord->score(), ctx.nominalPositionTick);
-    ctx.nominalDuration = durationFromTicks(ctx.beatsPerSecond, ctx.nominalDurationTicks);
-    ctx.nominalDynamicLevel = nominalDynamicLevel;
-    ctx.profile = profile;
+    int chordPosTick = chord->tick().ticks();
+    int chordDurationTicks = chord->durationTypeTicks().ticks();
+    BeatsPerSecond bps = chord->score()->tempomap()->tempo(chordPosTick).val;
```
igorkorsukov, line 69: There is no need to get a val

src/framework/mpe/mpetypes.h
```diff
@@ -385,53 +395,16 @@ struct ArticulationPatternSegment
     ExpressionPattern expressionPattern;
 };
 
-using ArticulationPattern = std::map<duration_percentage_t, ArticulationPatternSegment>;
+using ArticulationPattern = SharedMap<duration_percentage_t, ArticulationPatternSegment>;
 
-struct ArticulationsProfile
+struct ArticulationsProfile : public SharedMap<ArticulationType, ArticulationPattern>
```
igorkorsukov, line 400: Not a fan of this solution, I think the previous one was better.
There are several disadvantages when inheriting from standard components:
* The interface is too large, it may be that some methods are unnecessary or harmful.
* Too abstract interface (for example, `T pattern (...)` method is better than `T at (...)`)
* Difficult to debug - e.g. set breakpoint or log
* Difficult to change implementation

vpereverzev: Me either. I've changed it by analogy with ArticulationMap, in which I need almost the entire SharedHashMap interface and it is more convenient to use inheritance there.

However, now I see that it's unnecessary to use inheritance here. Returned the previous one. Thanks

src/engraving/playback/playbackeventsrenderer.cpp
vpereverzev: thanks, fixed.

Forgot to remove it when changed type

src/engraving/playback/metaparsers/internal/arpeggiometaparser.cpp
vpereverzev: reverted to the previous option

src/framework/mpe/mpetypes.h
igorkorsukov: Yes, inheritance is fine for `ArticulationMap`, because it's a `map` :)

igorkorsukov (approved):

PR #10250, opened 2022-01-12
[Qt 6 preparation] Explicitly capture QML signal handler arguments

cbjeukendrup:
Cherry-picked from #10108 

@vpereverzev, @igorkorsukov, @Eism, @RomanPudashkin, it would be great if we could also do this consequently in new code from now (and also ask this from other contributors), so that it will work directly with Qt 6.


Eism:
I can't find any information about the difference between:
```
function onDataChanged(newData) {
}

onDataChanged: (newData) {
}

onDataChanged: function (newData) => {
}
```
@cbjeukendrup do you know the answer?

cbjeukendrup:
I'm not sure whether the first option will work. When using the `Connections` object, you _do_ need to use that syntax (that rule got introduced in Qt 5.15, or a bit earlier). But I've never seen this being used when handling signals directly in the "declaration" of the object itself; I believe you can only use the property-like syntax (`onDataChanged: ...`) in that case.

The second and third look unfamiliar to me. Probably you meant: 
```qml
onDataChanged: (newData) => { // This one with arrow
  doSomething()
}

onDataChanged: function(newData) { // This one without arrow
  doSomething()
}
```
These are just the JavaScript syntaxes for anonymous functions, and QML supports them both. 
(See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions.)

I _guess_ even this would be valid: 
```qml
function doSomething(newData) {
  ...
}

onDataChanged: doSomething
```
But I haven't tried it and don't think it makes more sense.

I made the choice to use the `onDataChanged: function(newData) { }` syntax consistently, because I slightly prefer how it looks, and I think it's the most readable (even though it's a tiny bit more verbose). What would be your preference?

Eism:
I agree with you, `onDataChanged: function(newData) { }` is the most readable

Eism (approved):

PR #10235, opened 2022-01-11
[MU4] Menu item refactoring

Eism:
Resolves:  #10125 #8697

PR #10251, opened 2022-01-12
[Qt 6 preparation] Simplify propagating model data into delegate in GridViewSectional

cbjeukendrup:
Cherry-picked from #10108

I found out that one of the preferred ways to propagate properties to a Loader's sourceComponent is to set those properties on the Loader itself. The Loader will set itself as the _context_ for the sourceComponent, so that the sourceComponent can use those properties. So, it seems that we won't need `onLoaded: { … }` in some cases. When writing new code or refactoring old code, we can keep this in mind. 

In Qt 6, the old way did not work with the icons in the Note Input Bar. (_Why_ that is the case, is still a big riddle to me...) Luckily, the new way does work.

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/GridViewSectional.qml
@@ -19,7 +19,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-import QtQuick 2.12
+import QtQuick 2.15
 
 import "internal"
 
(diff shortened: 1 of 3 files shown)


Eism:
@cbjeukendrup I tested your solution on 5.15.2
There are a lot of warnings in qml if you try to change the order of the elements using the note input customise popup

For example

```
09:38:30.335 | WARN  | main_thread | Qt         | qrc:/qml/MuseScore/NotationScene/NoteInputBar.qml:130: TypeError: Cannot read property 'description' of null
09:38:30.335 | ERROR | main_thread | AppShell   | run: error: qrc:/qml/MuseScore/NotationScene/NoteInputBar.qml:131: TypeError: Cannot read property 'shortcut' of null
```

cbjeukendrup:
Thanks for checking, I'll have another look at it!

cbjeukendrup:
I updated this after #10235. The part about the NoteInputBar was not really actual anymore, but the simplification in GridViewSectional still makes sense.

Eism (approved):

PR #10261, opened 2022-01-13
Fix shadow note on small staves

cbjeukendrup:
It looked like this:
Schermafbeelding 2022-01-09 om 00 20 23


DmitryArefiev:
@cbjeukendrup Tested on Win10, Mac12, LinuxMint 20.2 - NOW IT **LOOKS GOOD**    

Thanks!

small

PR #10266, opened 2022-01-13
Fix #328003 score property crash

jonnymoo:
Resolves: https://musescore.org/en/node/328003

Following accessibility changes saving score properties causes MuseScore to crash

[ x] I signed CLA
[ x] I made sure the code in the PR follows the coding rules
[ x] I made sure the code compiles on my machine
[ x] I made sure there are no unnecessary changes in the code
[ x] I made sure the title of the PR reflects the core meaning of the issue you are solving
[ x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
[ x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
[ ] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/src/notation/view/widgets/scoreproperties.cpp
+++ b/src/notation/view/widgets/scoreproperties.cpp
@@ -249,7 +249,7 @@ bool ScorePropertiesDialog::save()
         QLayoutItem* valueItem = scrollAreaLayout->itemAtPosition(i, 1);
         if (tagItem && valueItem) {
             QString tagText = "";
-            
+
             // Is the property a standard one?
             QLabel* labelItem = dynamic_cast<QLabel*>(tagItem->widget());
             if (labelItem) {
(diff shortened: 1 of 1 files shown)


src/notation/view/widgets/scoreproperties.cpp
```diff
@@ -245,31 +245,43 @@ bool ScorePropertiesDialog::save()
...
+            }
+
             QLineEdit* value = static_cast<QLineEdit*>(valueItem->widget());
 
-            QString tagText = tag->text();
             if (tagText.isEmpty()) {
                 interactive()->warning(trc("notation", "MuseScore"),
                                        trc("notation", "Tags can't have empty names."));
-                tag->setFocus();
+                lineEditItem->setFocus();
```
RomanPudashkin, line 270: The crash can still occur here because of the null pointer. Also, I can't save new values of tags for some reason (always get this warning about empty names)...

jonnymoo: Rearranged the logic to make it more clearer. The validations would only ever fail on none standard properties - so this makes it clearer and the null checks now make more sense.

DmitryArefiev:
Tested #10289 on Win10, Mac12, LinuxMint 20.2 - Crash no longer occurs      

Please merge into master

Issue #10254, opened 2022-01-12
[MU4 Issue] "Restore previous session" dialog not yet implemented


bkunda:
**Describe the bug**
In MS3.6, this dialog appears when MuseScore is reopened after quitting unexpectedly: 

<img width="713" alt="Screenshot 2022-01-12 at 7 33 03 pm" src="https://user-images.githubusercontent.com/86290556/149091938-2caa92e4-e785-47dd-8013-27520f6542e3.png">

This dialog does not yet appear when restarting MuseScore 4 (nightlies).

**To Reproduce**
1. Create a new score and save it
2. Make some changes to the score but don't save them
3. Wait for the autosave time to lapse (set in `Preferences`), then force a reboot of your machine 
4. When computer restarts, open MuseScore. No dialog appears.

The implication of this is that the user is currently unable to recover a file that has been autosaved.

**Expected behavior**
Given we don't yet have a loading screen, I wonder if it is possible for this dialog to appear over the `Home`>`Scores` screen on startup instead?

<img width="1403" alt="Screenshot 2022-01-12 at 7 36 00 pm" src="https://user-images.githubusercontent.com/86290556/149092533-0ff24b2e-4bd2-4034-a2f3-d854cb6a801e.png">

Behaviour should be the same as in MS3.6: I.e. it should restore the last autosaved copy of the last open project file. 

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1)
 - Windows 11

**Additional context**
Relates to PR #10216

cbjeukendrup:
> force a reboot of your machine

Force-quitting MuseScore should be enough 😉 (macOS: easiest way: Right-click MuseScore icon in the Dock, press the "Option/alt" key and choose "Force Quit"; Windows: Task Manager)

Another option is DevTools > Crash Handler > Crash.

All these options _should_ produce a "Restore session?" dialog at next launch.

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED in #10273

PR #10273, opened 2022-01-14
[MU4] Sessions manager

Eism:
Resolves: #10254


src/appshell/internal/appshellconfiguration.cpp
```diff
@@ -238,3 +270,43 @@ std::string AppShellConfiguration::currentLanguageCode() const
 
     return locale.bcp47Name().toStdString();
 }
+
+mu::io::path AppShellConfiguration::sessionDataPath() const
+{
+    return globalConfiguration()->userAppDataPath() + "/session";
```
igorkorsukov, line 276: I think better to add extension (.json ?)

igorkorsukov (approved):

src/appshell/internal/appshellconfiguration.cpp
Eism: It is a directory for session.
The extension for session is .json. See sessionFilePath and 
static const mu::io::path SESSION_FILE("/session.json");

Jojo-Schmitz:
Why does MuseScore crash on generating the PNGs for the vtests?

Eism:
> Why does MuseScore crash on generating the PNGs for the vtests?

I am working on it now

DmitryArefiev:
Tested #10254 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

PR #10277, opened 2022-01-16
Fix #9220: [MU4 Issue] Enabling high-contrast switches to dark theme and swaps theme icons

### LINKED ISSUE #9220: [MU4 Issue] Enabling high-contrast switches to dark theme and swaps theme icons
**Describe the bug**
From Jessica Williamson : Just realized something about the theme placements between normal mode and high contrast mode. On normal we have light mode on left, dark on right. When high contrast mode is on, dark version is now on left, white on right. Wouldn't it make sense if the high contrast ones were swapped to match the normal ones? so both light versions on left and darks on right?

**To Reproduce**
1. Go to Edit>Preferences>Appearance
2. Check 'Enable high-contrast' (see video below)

**Expected behavior**
It should not switch from Light to Dark theme and should not swap the icons

**Screenshots**

Enable high-contrast

**Desktop:**
 OS: - Win10/Mac10.15.7/Linux Mint 20.2

jithinjohn2k2:
Resolves: #9220 

Swapped the positions of the order of White and Black Theme codes returned by allStandardThemeCodes(), changed the default value of High Contrast Theme and reordered the instances of Black and White Theme code usages.

### DIFF
--- a/src/appshell/view/firstlaunchsetup/themespagemodel.cpp
+++ b/src/appshell/view/firstlaunchsetup/themespagemodel.cpp
@@ -61,7 +61,7 @@ QVariantList ThemesPageModel::highContrastThemes() const
     QVariantList result;
 
     for (const ThemeInfo& theme : allThemes()) {
-        if (theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE || theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE) {
+        if (theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE || theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE) {
             result << ThemeConverter::toMap(theme);
         }
     }
--- a/src/appshell/view/preferences/appearancepreferencesmodel.cpp
+++ b/src/appshell/view/preferences/appearancepreferencesmodel.cpp
@@ -91,7 +91,7 @@ QVariantList AppearancePreferencesModel::highContrastThemes() const
     QVariantList result;
 
     for (const ThemeInfo& theme : allThemes()) {
-        if (theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE || theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE) {
+        if (theme.codeKey == HIGH_CONTRAST_WHITE_THEME_CODE || theme.codeKey == HIGH_CONTRAST_BLACK_THEME_CODE) {
             result << ThemeConverter::toMap(theme);
         }
     }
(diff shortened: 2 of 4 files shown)


cbjeukendrup (approved):
Looks good to me!

Thanks for updating the order in all those `if` statements too. Of course, it does not make any functional difference, but is absolutely nice for readability and consistency.

jithinjohn2k2:
> Thanks for updating the order in all those `if` statements too. Of course, it does not make any functional difference, but is absolutely nice for readability and consistency.

Yeah, only setDefaultValue() in uiconfiguration.cpp and the return value of allStandardThemeCodes() are logically necessary, others, as you said are for readability and consistency, Sorry, I forgot to mention them in the PR :sweat_smile: 

Thanks for approval @cbjeukendrup :slightly_smiling_face:

RomanPudashkin (approved):

DmitryArefiev:
Hi @jithinjohn2k2, it's fixed when enabling high-contrast in Light theme but **still persists in Dark theme**

https://user-images.githubusercontent.com/90187801/149756427-0360e01a-016a-4d45-bf0d-bd496111b61a.mp4

jithinjohn2k2:
I hope the behavior shown in the video is the expected one as the application is actually remembering the Previous General theme and the Previous High Contrast Theme separately.

image

So it is remembering which theme in each category(High Contrast mode / Non-High Contrast Mode) you have selected previously and just retains it when you switch back. 

For example, if you just open this window where everything is the default, and check the High Contrast box, then select the black mode and if you uncheck the box(Disable High Contrast), it should end up in Light mode(because we did not change) which is the default for the General theme and if you again Enable the High contrast Mode(check the box), it should be showing you the black mode because you left the setting for High Contrast mode in black.

The Light mode in the General Setting and the White mode in the High Contrast Setting are just the initial settings that lasts only until we click on any of the other themes.

I am not sure if I misunderstood something about the way it works after going through the code. I am a beginner here :sweat_smile:

cbjeukendrup:
@jithinjohn2k2 No, the video describes the current behaviour, which should change. It is namely _not_ desired that the last used "version" (light/dark) is remembered per category (HC/non-HC).

The desired behaviour is that non-HC light will switch to HC light, and non-HC dark to HC dark. So, the current "version" (light/dark) will be preserved when switching to another category.

By the way, this PR is perfectly fine and does exactly what it promises. The thing is that issue #9220 contained _two_ (more or less separate) tasks.

jithinjohn2k2:
>The thing is that issue #9220 contained _two_ (more or less separate) tasks.


Ohh, my bad, I didn't notice them. Got the clarity on how it should work.
I will try to make another PR if I am able to solve them completely.

Issue #9929, opened 2021-11-30
[MU4 Issue] Scaling in Windows is not incremental


bkunda:
**Describe the bug**
Changing the display scaling in Windows results in only two levels of scaling in MuseScore (effectively, a small and a large scaling level).

This is especially pronounced when using MS on a large monitor with an extended display configuration (although the more I look into it, the less I'm convinced that connecting an external monitor might be the real cause of this problem). 

Basically, scaling options on a large external display result in MuseScore being displayed either somewhat large, or very small. 

This video attempts to show what happens when the scaling is adjusted: 

https://user-images.githubusercontent.com/86290556/144052213-950c3a2b-f407-4eee-8334-31a543acc347.mp4

**Expected behavior**
Scaling should be incremental in 25% intervals, as per other apps (see MS Word example in above video).

**Desktop (please complete the following information):**
 - Windows 11

**Additional context**
Possibly related to #9263 
Also a problem in MS3.6 (see above video)

DmitryArefiev:
Tested #9956 on Win10, laptop with **FullHD** display  - now MU also changes scaling when switching to **125%**

DmitryArefiev:
@bkunda Would be good to check this on a large display and Win11 as well. Thanks!

bkunda:
@DmitryArefiev I've checked this on my system and scaling now appears wonderfully incremental. Fixed from my end 🙂

DmitryArefiev:
Hi @bkunda, Can you check it on your Windows again please? 
On my **Win10** and latest **master build** I don't see a **different between 100% and 125%** in app size. Maybe something was broken or just something wrong in my environment..

Thanks!

bkunda:
Yeah I'm also having a problem here now. 
There seems to be a slight resizing of the app window between 100% and 125%, but no change in the topbar, toolbar or sidebar sizes. Then there is a huge jump in size from 125% to 150%. 
This is the case for the both monitors (laptop and external monitor).

DmitryArefiev:
@bkunda Thanks for checking! Let's reopen this.

DmitryArefiev:
The same issue #10267 was opened recently.. Let's close the initial bug #9929 :)

bkunda:
Issue now raised in #10267

DmitryArefiev:
Since  it's hard to reproduce #10267 (the issue about File menu placement on 125%) let's reopen the current issue #9929 to fix incremental bug on 125%

[...]

PR #10285, opened 2022-01-18
[MU4] UI fractional scaling

Eism:
Resolves: #9929

### DIFF
--- a/src/appshell/appshell.cpp
+++ b/src/appshell/appshell.cpp
@@ -83,6 +83,15 @@ int AppShell::run(int argc, char** argv)
 
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 
+    //! NOTE: For unknown reasons, Linux scaling for 1 is defined as 1.003 in fractional scaling.
+    //!       Because of this, some elements are drawn with a shift on the score.
+    //!       Let's make a Linux hack and round values above 0.75(see RoundPreferFloor)
+#ifdef Q_OS_LINUX
+    QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
+#elif defined(Q_OS_WIN)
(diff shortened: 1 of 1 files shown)


MarcSabatella:
I don't know what "fractional scaling" is, but does this by chance address the fact that the icons in MU4 and the score itself are way too small on high DPI displays, and that unlike previous releases, the -D command line option can't be used to override this?  This to me is a showstopper issue, as it makes MU4 all but unusable for me right now.  And I'd really like to be able to start moving forward soon on documentation.  See https://github.com/musescore/MuseScore/issues/10354

Eism:
Hey @MarcSabatella!
No, this PR is not fixed #10354. 
I saw your issue and wanted to fix it, but I just forgot about it :( Sorry

So, I tried fixing the custom DPI setting, but that caused a few problems that take time to investigate. Therefore, I will create a separate PR with a fix.

RomanPudashkin (approved):

DmitryArefiev:
Tested #9929 on Win10 - **Now incremental scaling in Windows is FIXED**  
But also a bug with score blurring appears now (I will log it a separate issue)

src/appshell/appshell.cpp
```diff
@@ -83,6 +83,15 @@ int AppShell::run(int argc, char** argv)
 
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 
+    //! NOTE: For unknown reasons, Linux scaling for 1 is defined as 1.003 in fractional scaling.
+    //!       Because of this, some elements are drawn with a shift on the score.
+    //!       Let's make a Linux hack and round values above 0.75(see RoundPreferFloor)
+#ifdef Q_OS_LINUX
+    QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
```
theofficialgman, line 90: @Eism please provide more information. this change breaks fractional scale values. if you have a particular issue with one DE and display server combination, it is best to have it corrected upstream and not break this for everyone else.

with the current code we get:
1X scaling for 0.75-1.749999
2X scaling for 1.75-2.749999
3X scaling for 2.75-3.749999
etc

this is not a good system. using passthrough would be much better.

Issue #9815, opened 2021-11-22
[MU4 Issue] Right-click in master palette triggers redundant context menu


bkunda:
**Describe the bug**
Right clicking any item in the master palette triggers a redundant, "floating" context menu that seems disconnected from the palette itself. Options in the context menu are all disabled.
This context menu remains on screen for some time, even after the master palette dialog has been closed. 

![master-pallete-strange-menu](https://user-images.githubusercontent.com/86290556/142858220-793494a9-8270-4d9c-8a3d-8f25a34bbf8c.gif)

**Expected behavior**
Nothing should be triggered when right-clicking on a master palette item (per MS3.6)

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1)
 - Windows 11

DmitryArefiev:
Tested #10297 on Win10, Mac12, LinuxMint 20.2 - FIXED

PR #10297, opened 2022-01-18
[MU4] Fixed right-click in master palette

Eism:
Resolves: #9815

### DIFF
--- a/src/notation/view/notationpaintview.cpp
+++ b/src/notation/view/notationpaintview.cpp
@@ -68,8 +68,6 @@ NotationPaintView::NotationPaintView(QQuickItem* parent)
     dispatcher()->reg(this, "diagnostic-notationview-redraw", [this]() {
         update();
     });
-
-    qApp->installEventFilter(this);
 }
 
 void NotationPaintView::load()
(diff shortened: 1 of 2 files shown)


src/notation/view/notationpaintview.cpp
```diff
@@ -876,16 +876,20 @@ bool NotationPaintView::event(QEvent* ev)
     return QQuickPaintedItem::event(ev);
 }
 
-bool NotationPaintView::eventFilter(QObject* obj, QEvent* ev)
+bool NotationPaintView::eventFilter(QObject* object, QEvent* event)
```
cbjeukendrup, line 879: Since this event filter only filters events for the NotationPaintView itself, wouldn't it make more sense to convert it to [`bool NotationPaintView::event(QEvent* event)`](https://doc.qt.io/qt-5/qobject.html#event) and remove `installEventFilter(this);` in the constructor?
It seems that the body of the function can stay unchanged, except that you need `return QQuickPaintedItem::event(event);` at the end.

Eism: thanks, fixed

DmitryArefiev:
Tested #9815 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Issue #10337, opened 2022-01-20
[MU4 Issue] Note name displayed in the bottom screen bar is not refreshing


izarys:
**Describe the bug**
Moving the note by up/down arrows does not change the note name displayed in the bottom screen bar.

**To Reproduce**
1. Create an empty score
2. Add a note
3. Click up/down to change the added note
4. Note name displayed in bottom screen bar is not refreshed

**Expected behavior**
The note name is changed accordingly.

**Screenshots**

![musescore_a4](https://user-images.githubusercontent.com/69964767/150300329-e5d7b92b-5692-4498-84cf-da49584fa0b5.PNG)
![musescore_d5_incorrect](https://user-images.githubusercontent.com/69964767/150300355-c4b9eb1e-8eee-42c0-a965-68e422dd06ad.PNG)
You can see that on the second screenshot it says A4, but it should be D5.

**Desktop (please complete the following information):**
 - OS: Windows 10

**Note**
It's working correctly in MuseScore 3.

cbjeukendrup:
Working on it!

PR #10340, opened 2022-01-20
Update notation accessibility information when notation is edited

cbjeukendrup:
Resolves: #10337

### DIFF
--- a/src/notation/internal/notationaccessibility.h
+++ b/src/notation/internal/notationaccessibility.h
@@ -40,7 +40,7 @@ class IGetScore;
 class NotationAccessibility : public INotationAccessibility, public async::Asyncable
 {
 public:
-    NotationAccessibility(const IGetScore* getScore, async::Notification selectionChangedNotification);
+    NotationAccessibility(const Notation* notation);
 
     ValCh<std::string> accessibilityInfo() const override;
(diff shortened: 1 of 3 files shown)


src/notation/internal/notationaccessibility.h
```diff
@@ -40,9 +40,9 @@ class IGetScore;
 class NotationAccessibility : public INotationAccessibility, public async::Asyncable
 {
 public:
-    NotationAccessibility(const IGetScore* getScore, async::Notification selectionChangedNotification);
+    NotationAccessibility(const Notation* notation);
 
-    ValCh<std::string> accessibilityInfo() const override;
```
Eism, line 45: It's better to keep std::string here

Eism: > in the future we will rewrite libmscore to use std::string too?

yes

Eism (approved):

DmitryArefiev:
Tested #10337 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

DmitryArefiev:
@Eism But screenreader doesn't announce the changes when moving notes up/down. Will log a separate bug for that

PR #10346, opened 2022-01-20
Fix #324840: Ctrl+Scrollwheel zoom on Linux X11

Gobbel2000:
Resolves: https://musescore.org/en/node/324840

Qt provides two methods for getting information out of a QWheelEvent: pixelDelta() and angleDelta(), which could both be used in musescore. I assume the recent changes to libinput, that created this problem, either made pixelDelta() available or changed its behavior. Either way, for me that method always returned 120 (or -120) in the y component which then gets interpreted as 24  zoom steps at once. 120 is a good value when it comes from angleDelta(), and in that case it also gets interpreted correctly.

I'm not sure where the value `PIXELSSTEPSFACTOR = 5` comes from, but I can only assume that it works for other systems. The [Qt documentation](https://doc.qt.io/qt-5/qwheelevent.html#pixelDelta) notably says:

> **Note:** On X11 this value is driver specific and unreliable, use angleDelta() instead

which is pretty much what this change does. 

As mentioned in the issue, this problem only arises when using `xf86-input-libinput` version 1.2.0 or higher. So far that mostly affects rolling release distros, but it seems like the new Ubuntu releasing April will have the [new version](https://packages.ubuntu.com/jammy/x11/xserver-xorg-input-libinput) as well.

---

### DIFF
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -398,6 +398,15 @@ void NotationViewInputController::wheelEvent(QWheelEvent* event)
     qreal stepsX = 0.0;
     qreal stepsY = 0.0;
 
+// pixelDelta is unreliable on X11
+#ifdef Q_OS_LINUX
+    if (std::getenv("WAYLAND_DISPLAY") == NULL) {
+        // Ignore pixelsScrolled unless Wayland is used
+        pixelsScrolled.setX(0);
+        pixelsScrolled.setY(0);
+    }
+#endif
+
     if (!pixelsScrolled.isNull()) {
         dx = pixelsScrolled.x();
         dy = pixelsScrolled.y();


src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -363,20 +363,25 @@ void NotationViewInputController::endOfScore()
...
-    QPoint stepsScrolled = event->angleDelta();
-
     int dx = 0;
     int dy = 0;
     qreal stepsX = 0.0;
     qreal stepsY = 0.0;
 
+    QPoint stepsScrolled = event->angleDelta();
+
+// pixelDelta is unreliable on X11
```
RomanPudashkin, line 373: Linux is not only about X11. Will this fix work fine on Wayland?

Jojo-Schmitz: Are we building MuseScore for supporting Wayland? I don't think we do.

Gobbel2000: I tested this on wayland with both KDE and sway and the pixelDelta value isn't provided there, meaning the problem doesn't arise in the first place _for now_. It could be that in the future wayland will provide that in a compliant way, in which case the exception should be limited to X11 only. It could also end up broken as with X11 right now. 

For now the change doesn't do anything on Wayland, but maybe this would need to be reevaluated in the future.

Jojo-Schmitz: But whether it is X11 or Wayland is a runtime thing rather than a compile time one, isn't it?

Gobbel2000: Yeah, the same binary runs on X11 and Wayland, so that would have to be done at runtime. I believe the best way to do that is by looking at certain environment variables.

Jojo-Schmitz: I guess that'd be `$XDG_SESSION_TYPE`, which should be `x11`or `wayland` Or not used at all on non-Linux/UNIX platforms?

Jojo-Schmitz: Or `$DISPLAY` vs. `$WAYLAND_DISPLAY` vs. neither

Gobbel2000: I now put up a version that does not affect Wayland by checking `$WAYLAND_DISPLAY` that should be safe for most situations. I still have it dependent on `Q_OS_LINUX` so as to not make any assumptions about the environment variables on other platforms.

bmarwell:
Can I somehow propose this inclusion to someone?

If not, I'm going to mail the maintainers of the Linux distro packages to include this. I get my MuseScore from the AUR, that should be doable.

MarcSabatella:
See https://github.com/musescore/MuseScore/issues/11198.  It seems the issue is known on MU4, and apparently perhaps the existence of this PR, but I don't know the status.  Wouldn't hurt to add a link to this if you can verify this solution actually works on MU4 - so much about the UI code has changed, it might no longer apply.

RomanPudashkin:
@Gobbel2000 could you please rebase it?

Gobbel2000:
I have now squashed and rebased on the current tip of master, the patch still works as before. This patch is for MU4, but the same applies to version 3 as well, mostly just in a different file. The patch for that is already part of #9000. 

In the meantime there was also an update to Wayland (1.21) recently that added smooth scrolling support. I have retested everything there as well and that implementation still doesn't create any problems from what I could see. This further justifies the exception to not use this fix on Wayland.

It would be great if this gets merged for 4.0.

RomanPudashkin (approved):

bmarwell:
Thank you so much, @RomanPudashkin!!!

PR #10352, opened 2022-01-21
[MU4] Fix #10169: Key shortcut for resizing hairpin not working as MS3

### LINKED ISSUE #10169: Key shortcut for resizing hairpin not working as MS3. Hairpin visualisation does not update when resized via shorcut.
**Describe the bug**
There is a really important shortcut capability for lines (especially Hairpins) that isn't quite working right for MS4. There is also a visualisation problem where the hairpin connection lines do not update (see video below).

**To Reproduce**
(See video)
1. Apply a hairpin to some notation.
2. Select the hairpin with you mouse
3. Try to use the shortcut (**shift + alt + right arrow / left arrow**) to resize the hairpin. Note that this doesn't work
4. If you click on the central handle for any line, this shortcut _**does**_ work

**Expected behavior**
Match exact functionality of MS3. All lines should be able to be resized by selecting them and applying the above shortcut. Since lines are core to how MuseScore works, a large number of users will notice if it isn't working identically. Interestingly, slurs **do** actually work as expected with (**shift + alt + right arrow / left arrow**).

----

In addition, when you select the central handle to move a line and use the shortcut to resize it, the pink connection lines do not update correctly. Observe my attached video where I compare MS4 to MS3.

https://user-images.githubusercontent.com/21022311/147373382-3623e924-62df-4c27-8162-b002cda747f4.mp4

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10169

### DIFF
--- a/src/engraving/libmscore/engravingitem.cpp
+++ b/src/engraving/libmscore/engravingitem.cpp
@@ -397,6 +397,11 @@ bool EngravingItem::onTabStaff() const
     return stt ? stt->isTabStaff() : false;
 }
 
+bool EngravingItem::hasGrips() const
+{
+    return gripsCount() > 0;
+}
+
 int EngravingItem::track() const
 {
     return _track;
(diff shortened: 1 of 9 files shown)


Eism (approved):

DmitryArefiev:
Tested #10169 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

PR #10357, opened 2022-01-22
Fix unpredictable jumping when zooming while there is a selection

cbjeukendrup:
Resolves: not reported yet

The idea is that the pressing Ctrl++ or Ctrl+- would zoom to the center of the selection if there is a selection, otherwise to the center of the view. The former was not implemented correctly (`selection->canvasBoundingRect()` is in logical coordinates and needs to be converted to "physical" coordinates, but this was done incorrectly).

### DIFF
--- a/src/notation/view/notationpaintview.cpp
+++ b/src/notation/view/notationpaintview.cpp
@@ -22,7 +22,6 @@
 #include "notationpaintview.h"
 
 #include <QPainter>
-#include <QWindow>
 
 #include "actions/actiontypes.h"
 #include "stringutils.h"
@@ -503,6 +502,11 @@ std::pair<qreal, qreal> NotationPaintView::constraintCanvas(qreal dx, qreal dy) 
     return { dx, dy };
 }
 
+PointF NotationPaintView::viewportTopLeft() const
+{
+    return toLogical(PointF(0.0, 0.0));
+}
+
 RectF NotationPaintView::viewport() const
 {
     return toLogical(RectF(0.0, 0.0, width(), height()));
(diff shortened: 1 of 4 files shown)


RomanPudashkin (approved):

DmitryArefiev:
@cbjeukendrup Thank for implementing this! Comparing with previous build it's much better.

I just noticed the selection is not properly placed directly to the center when zooming in if that selection is somewhere on the edge.

https://user-images.githubusercontent.com/90187801/151719148-6a8b2773-0545-4230-981e-e63e5641ae57.mp4


Probably requires a separate issue?

cbjeukendrup:
@DmitryArefiev Yes, I think so. This PR is just to fix that the "focus point" for zooming is the center of the selection. So if the selection is on the edge, it will zoom in to that edge, but no effort is made move the selection into the view or closer to the center of the view. I wonder a bit how this should work exactly; wouldn't it become quite jumpy when we would move the view so that the selection becomes the center? 

I see a couple of options:
- Always placing the selection in the center of the view when zooming (may be too jumpy)
- Placing the selection in the center of the view only when it would have become out of view after zooming (still jumpy)
- Only jump the smallest possible distance to keep move the selection just into the view, so that it is at the edge closest to where it was;
- Only jump such distance so that not a smaller part of the selection becomes visible (i.e. if the selection _was_ already out of view, it will stay out of view, but not become _more_ out of view than it was)
- Jumping by some fraction of distance between the center of the view and the center of the selection, so that the center of the selection gradually converges to the center of the view after zooming in/out a lot of times.

(Yes, we can make something simple as scrolling and zooming as complicated as we want!) 😄

PR #10359, opened 2022-01-23
Fix #8241: [MU4 Issue] No file path visible in recent score list

### LINKED ISSUE #8241: [MU4 Issue] No file path visible in recent score list
**Describe the bug**
When trying to open a score from `File->Open recent` (or form the `Score` page) the name of the **score** is visible but there is no way to know which file it is.

**To Reproduce**
1. Open a score
2. Save it using `Save as...`
3. Close the score
4. Go the the `Score` page (or use `File -> Open recent`). Now two identical entries are shown although it are two different files (and most likely different scores).

**Expected behavior**
At least a hint which file is behind an entry in the recent files list.

**Screenshots**

 - OS: Linux (but I guess this is the same for other platforms).

**Additional context**
It is common practice to have to make a copy of a document to try something and ensure the the original document is not destroyed. In that case `File -> Save as` is used to make that copy. But in the recent files list only the `workTitle` of the score properties is shown which hasn't changed. When loading a score with a certain name, there is no way to tell which **file** is actually loaded. The only way to find out is load the file and check the score properties. If it is the wrong file, try another.

iamjithinjohn:
Resolves: #8241 

Swapped the positions of the order of White and Black Theme codes returned by allStandardThemeCodes(), changed the default value of High Contrast Theme and reordered the instances of Black and White Theme code usages.

### DIFF
--- a/src/appshell/view/appmenumodel.cpp
+++ b/src/appshell/view/appmenumodel.cpp
@@ -120,7 +120,7 @@ void AppMenuModel::setupConnections()
         workspacesItem.setSubitems(makeWorkspacesItems());
     });
 
-    connect(qApp, &QApplication::applicationStateChanged, this, [this](Qt::ApplicationState state){
+    connect(qApp, &QApplication::applicationStateChanged, this, [this](Qt::ApplicationState state) {
         if (state != Qt::ApplicationActive) {
             resetNavigation();
         }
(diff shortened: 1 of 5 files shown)


Jojo-Schmitz:
Check and fix the code style mistakes, please.

iamjithinjohn:
> Check and fix the code style mistakes, please.

Sorry for that, I didn't configure the formatter properly, fixed now :slightly_smiling_face:

RomanPudashkin:
Fixed in https://github.com/musescore/MuseScore/pull/10473

Issue #10234, opened 2022-01-10
[MU4 Issue] [Accessibility] Text editing events are not announced by screen readers


shoogle:
**Describe the bug**
When entering lyrics or stave text, or when typing in a text frame, screen readers should announce text characters as they are typed, selected, deselected and navigated with the cursor keys. At the moment, the screen reader is silent during this process.

![image](https://user-images.githubusercontent.com/11011881/148830897-debe0bea-b75e-4eda-af83-651c06aef91a.png)

**To Reproduce**
1. Start the screen reader (Narrator, NVDA, VoiceOver, Orca, etc.).
2. Click on a note or rest in any score.
3. Press Ctrl+T (macOS: Cmd+T) to enter stave text.
4. Start typing text characters and listen to the screen reader.
5. Use the arrow keys to navigate through the typed character. Keep listening to the screen reader.
6. Hold Shift and continue pressing arrow keys to select and deselect typed characters. Keep listening to the screen reader.
7. Repeat steps 4 to 6 in your platform's native text editor (e.g. Notepad, TextEdit) and compare the screen reader experience.

**Expected behavior**
The screen reader should behave the same in MuseScore during text editing as it does in the platform's native text editor. It should announce text characters as they are typed, navigated, selected and deselected.

**Desktop (please complete the following information):**
 - OS: Windows, macOS, Linux

**Additional context**
In MuseScore 3, this was achieved by [sending a message to the screen reader](https://github.com/musescore/MuseScore/blob/3.x/mscore/scoreaccessibility.cpp#L93-L103) that contained the [exact text characters that should be spoken](https://github.com/musescore/MuseScore/blob/2513676e512d29d554cb6c4d37d3efaf53ea2c5b/libmscore/textbase.cpp#L555-L670), but this was more of a hack than an actual solution.

In MuseScore 4 it should be achieved by subclassing [QAccessibleTextInterface](https://doc.qt.io/qt-5/qaccessibletextinterface.html) similarly to how [we already subclass QAccessibleInterface and QAccessibleValueInterface](https://github.com/musescore/MuseScore/blob/f30d808e5dd84fc732291ab3b222dfa2c55095ae/src/framework/accessibility/internal/accessibleiteminterface.h#L33). However, I'm not sure how we would integrate this QAccessibleTextInterface subclass into the accessibility tree given that it would be used by some elements but not others.

Issue #10329, opened 2022-01-19
[MU4 Issue] Some shortcuts with Alt (Alt+Shift+E to Edit Text) stopped working on Windows/Linux


DmitryArefiev:
**To Reproduce**
1. Select text 
2. Hit **Alt+Shift+E** to edit - **it triggers Menu Bar instead of editing**

https://user-images.githubusercontent.com/90187801/150197617-9543d773-75ed-4ab8-9fc1-cdea4f560256.mp4

**Desktop:**
 - OS: Win/Lin

**Additional context**
**Alt+V** also doesn't work (e.g. if to assign it to create accent

PR #10371, opened 2022-01-24
[MU4] Accessibility. Score text editing

Eism:
Resolves: #10234 #10329

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -32,8 +32,8 @@ using namespace Ms;
 
 bool AccessibleItem::enabled = true;
 
-AccessibleItem::AccessibleItem(Ms::EngravingItem* e)
-    : m_element(e)
+AccessibleItem::AccessibleItem(Ms::EngravingItem* e, Role role)
+    : m_element(e), m_role(role)
 {
 }
 
@@ -54,7 +54,7 @@ AccessibleItem::~AccessibleItem()
 
 AccessibleItem* AccessibleItem::clone(Ms::EngravingItem* e) const
 {
-    return new AccessibleItem(e);
+    return new AccessibleItem(e, m_role);
 }
 
 void AccessibleItem::setup()
@@ -146,7 +146,7 @@ const IAccessible* AccessibleItem::accessibleChild(size_t i) const
 
 IAccessible::Role AccessibleItem::accessibleRole() const
 {
-    return IAccessible::Role::ElementOnScore;
+    return m_role;
 }
 
 QString AccessibleItem::accessibleName() const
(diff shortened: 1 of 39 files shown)


src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -165,26 +165,139 @@ QString AccessibleItem::accessibleDescription() const
...
+    textCursor->setSelectLine(startCoord.first);
+    textCursor->setSelectColumn(startCoord.second);
+
+    result = textCursor->selectedText();
+
+    delete textCursor;
+    return result;
+}
+
+int AccessibleItem::accesibleCharacterCount() const
```
igorkorsukov, line 291: missing letter `s` ;)

igorkorsukov (approved):

shoogle:
Awesome! Unfortunately I get a crash when I try to edit frame text with VoiceOver enabled on macOS.

shoogle:
Crash on macOS is fixed, thanks! Speech is working perfectly when editing frame text (title, subtitle & composer, etc.) and lyrics, but it's not working for stave text or tempo markings.

Your implementation is a bit different to what I imagined. I thought text elements would need a normal AccessibleItem to represent the element in its normal state, plus a child AccessibleTextItem to represent the item in its editing state. However, your method of adding text capabilities to normal AccessibleItems seems to work just fine!

DmitryArefiev:
On my side on Mac (Monterey, M1), speech is not working when **entering text** (Title, Staff text, Tempo, Lyrics). 
But selection/deselection works.

shoogle:
After the latest commits I get the opposite problem to @DmitryArefiev. Now speech works on macOS when entering text but not when selecting it or navigating through it with the arrow keys. Strangely, if I hold the Alt key while navigating then speech works better for some reason.

Eism:
@shoogle thanks for testing!
On which text elements does the selection not work for you?

igorkorsukov (approved):

DmitryArefiev:
On my side, **Mac** looks fine now - speech is working when entering text (Title, Staff text, Tempo, Lyrics) as well as selection and deselection.

Eism:
Ok, let's merge it. 
@shoogle If this still doesn't work for you, please open a separate issue.

shoogle:
@Eism, it turns out I accidentally enabled a VoiceOver setting called "Quick Nav" which makes the arrow keys behave very differently to how they do normally. This affected ordinary keyboard navigation within the score as well as text editing.

image

"Quick Nav" mode is toggled on and off by pressing the Right and Left arrow keys at the same time while VoiceOver is running. I must have done this accidentally at some point while testing.

Speech during text editing works perfectly when Quick Nav is disabled. Sorry for the noise!

PR #10398, opened 2022-01-28
[MU4] gradual_tempo_changes_processing

vpereverzev:
Introduced processing of gradual tempo changes and covered the basic cases with unit-tests

Here is a demo (slowed down for some reason known only to my screen capturer):
Peek 2022-01-27 20-33

Important note:
Despite the fact that gradual tempo changes are already taken into account when calculating the tempomap, the existing playback system does not work properly with tempo map at all. Soon it will be replaced by a new playback system, which will provide a better playback experience and solve this problem as well


src/engraving/libmscore/score.cpp
```diff
@@ -535,6 +536,29 @@ void Score::fixTicks()
...
+        const TempoChangeRanged* tempoChange = toTempoChangeRanged(spannerItem);
+        if (!tempoChange) {
+            continue;
+        }
+
+        int tickPositionFrom = tempoChange->tick().ticks();
+        BeatsPerSecond currentBps = tempomap()->tempo(tickPositionFrom);
+        BeatsPerSecond newBps = currentBps * tempoChange->tempoChangeFactor();
+
+        std::map<int, double> tempoCurve = TConv::easingValueCurve(tempoChange->ticks().ticks(), 4, std::abs(currentBps.val - newBps.val), tempoChange->easingMethod());
```
jeetee: Change the first param to `tickPositionFrom` for clarity (and no double fetching).
Wouldn't it be better to make the hardcoded 4 here depend on the spanned duration instead? Looking at the uses for the current TempoChanges plugin there are often more markings inserted (each segment there) especially in curved ways.
So I'd make this depend on the amount of change and rate of change instead; striving for a calculation every half beat approximately.

src/engraving/libmscore/score.cpp
```diff
@@ -535,6 +536,29 @@ void Score::fixTicks()
...
+
+        for (const auto& pair : tempoCurve) {
+            tempomap()->setTempo(tickPositionFrom + pair.first, BeatsPerSecond(currentBps.val + pair.second));
```
jeetee, line 561: This sets the tempo markers as type FIX into the map; not sure if we want to use the type::RAMP here instead (which seems to be reserved for supporting this)

jeetee: Although I'm fine with discrete markings here as well; it kind of depends on whether we need to be able to differ between actual discrete fixed markings and such a RAMP marking for export for example

src/engraving/libmscore/timesig.cpp
```diff
@@ -601,6 +602,24 @@ QString TimeSig::accessibleInfo() const
     return QString("%1: %2").arg(EngravingItem::accessibleInfo(), timeSigString);
 }
 
+void TimeSig::added()
+{
+    IF_ASSERT_FAILED(score()) {
+        return;
+    }
+
+    score()->fixTicks();
```
jeetee: Does time signature affect this? I thought not.

src/engraving/libmscore/score.cpp
```diff
@@ -486,7 +486,7 @@ void Score::addMeasure(MeasureBase* m, MeasureBase* pos)
 }
 
 //---------------------------------------------------------
-//    fixTicks
+//    setUpTempoMap
```
jeetee, line 489: might I suggest `generateTempoMap` or `refreshTempoMap` instead of setUp (which feels weird English)

src/engraving/utests/tempomap_data/absolute_tempo_80_to_120_bpm/absolute_tempo_80_to_120_bpm.mscx
```diff
@@ -0,0 +1,396 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<museScore version="4.00">
+  <programVersion>4.0.0</programVersion>
```
jeetee, line 3: test files used to not include programVersion information if you generated them when running musescore with -t, not sure if including this information in these files is now intentional?

src/engraving/libmscore/breath.h
```diff
@@ -79,6 +79,8 @@ class Breath final : public EngravingItem
     EngravingItem* nextSegmentElement() override;
     EngravingItem* prevSegmentElement() override;
     QString accessibleInfo() const override;
+    void added() override;
+    void removed() override;
```
RomanPudashkin, line 83: Better to move these methods to the private part of the class

src/engraving/libmscore/engravingobject.h
```diff
@@ -274,6 +274,9 @@ class EngravingObject
     void writeProperty(XmlWriter& xml, Pid id) const;
     void writeStyledProperties(XmlWriter&) const;
 
+    virtual void added() {}
+    virtual void removed() {}
```
RomanPudashkin, line 278: What do you think about renaming these methods to onAdded and onRemoved? Also, I would suggest moving these methods under the protected section

src/engraving/libmscore/score.cpp
vpereverzev: I know it's pretty confusing (I mean tick and ticks), but these two values are not similar. (tempoChange->ticks() vs tempoChange->tick())

So there is no double fetching and it uses a proper one

jeetee: True, read straight over it.
One of my gripes with TempoMap as well and proposal to rename ticks to len in more places

src/engraving/libmscore/score.cpp
vpereverzev: I think FIX type is much easier to understand and maintain, so I'd stick with this approach.

However, if we'll want to achieve a lower sampling rate, it would make sense to make the parameter stepCount configurable by user

src/engraving/libmscore/timesig.cpp
vpereverzev: Unfortunately, the current implementation of former Score::fixTicks includes lots of time signature calculations and building the timesigMap. I think we'll be able to extract this method and use it separately soon

src/engraving/utests/tempomap_data/absolute_tempo_80_to_120_bpm/absolute_tempo_80_to_120_bpm.mscx
vpereverzev: yep, it's absolutely intentional thing nowadays

src/engraving/libmscore/breath.h
vpereverzev: done

src/engraving/libmscore/engravingobject.h
vpereverzev: well, I prefer to not use such names anywhere except of qml to be honest

RomanPudashkin (approved):

src/engraving/libmscore/engravingobject.h
jeetee: onAdded and onRemoved indicate the thing being the event handler itself, not the notifier (cfr javascript/qml)

Issue #10120, opened 2021-12-16
Score Migration


Tantacrul:
**Task description**
The following documents specify updates to our score migration process given the engraving changes made for MS4.

Simon has specified the alterations here: (Apologies about this being a restricted document. We'll post the details publicly shortly)
https://alice.mu.se/display/UK/Score+migration+for+MuseScore+4.0

Updated Figma assets here:
https://www.figma.com/file/l0TQUrNFZmEfrZnfHhFDfb/?node-id=2584%3A0

![image](https://user-images.githubusercontent.com/21022311/146411587-e86034da-b08c-4662-a56d-72a870e60d89.png)

![image](https://user-images.githubusercontent.com/21022311/146412611-d54074f9-fed1-4819-9c20-897df6663bdd.png)

Jojo-Schmitz:
The patch level is not mentioned? Like 3.6.2, 2.2.1, 3.2.3, etc? That would give a bit of extra info without having to dive into Score Properties. And is what 3.x did.

Tantacrul:
An oversight. Just fixed it.

![image](https://user-images.githubusercontent.com/21022311/146438858-48ff2a51-d39d-429a-89e3-ebcc1997a8e3.png)

igorkorsukov:
@DmitryArefiev Please check, PR https://github.com/musescore/MuseScore/pull/10133

DmitryArefiev:
@igorkorsukov Tested #10133 on Win10, Mac12, LinuxMint 20.2 - found some things:

1. The version at the top doesn't correspond with the MU version were the score was created. 
E.g. **This file was created in 2.2.5** this message appears for score created in **MU 2.0.1**.  
In fact the dialogs just show 2.2.5 or 3.3.2

2. The wrong dialog applied for score created in **MU3.6.2**

This: 

![Capture](https://user-images.githubusercontent.com/90187801/150406536-4290c1cb-1544-44d7-aec6-f73087e42d8e.png)

Instead of this:

![Capture2](https://user-images.githubusercontent.com/90187801/150406703-2c54bcfa-8ea8-4082-a827-8664b9bf3064.PNG)

3. Since the two dialogs (for 2.2.5 and 3.3.2) are for two kinds of scores, **'Don't ask again'** checkbox should work independently for each of them (minor issue)

DmitryArefiev:
[...]

Issue #9070, opened 2021-09-08
[MU4 Issue] mscz score file saved in v3 is loaded using "Emmentaler" font 


wizofaus:
**Describe the bug**
Scores created in v3.6 are loaded in v4.0 with the wrong musical symbols/text font ("Emmentaler").

**To Reproduce**
1. Create a score in v3.6 (can be empty) and save it
2. Open it in v4
3. Go to Format|Style|Score and observe that Musical Symbols font is Emmentaler (even though it was Leland in 3.6)

**Expected behavior**
Font should remain as Leland.

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**

oktophonie:
Just to add that this problem is concealed if the 'use Leland' box is ticked, but that shouldn't be necessary in this case (similarly for Emmentaler). To be addressed as part of #10120

oktophonie:
In fact, this problem appears to be more general: it looks like the style defaults (for a 3.6 file) are being overwritten by the style sheet from a previous version (3.5, perhaps?), some obvious indicators of which are the page numbers appearing at the bottom of the page rather than at the top, etc. Ticking the 'use Leland'/'use Edwin' buttons muddy the water as they will reapply the music and text font settings, but other unrelated settings will still change unexpectedly.

PR #10400, opened 2022-01-28
[MU4] Migration dialog. Stabilization

RomanPudashkin:
Resolves: 
https://github.com/musescore/MuseScore/issues/10120#issuecomment-1017846773

### DIFF
--- a/src/appshell/commandlinecontroller.cpp
+++ b/src/appshell/commandlinecontroller.cpp
@@ -275,7 +275,9 @@ void CommandLineController::apply()
         }
 
         //! NOTE Don't write to settings, just on current session
-        projectConfiguration()->setMigrationOptions(migration, false);
+        for (project::MigrationType type : project::allMigrationTypes()) {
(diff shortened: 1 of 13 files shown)


oktophonie:
Can confirm this now shows the correct dialog according to the version of the file being opened!

The 'Automatic spacing' checkbox option seems to have no effect, which is a regression from 3.6. (The 'Leland' option seems to work fine; the 'Edwin' option works selectively, but that is the same as in 3.6 and should be handled as an entirely separate issue.)

RobFog:
Is #9070 related?

oktophonie:
Certainly not impossible; it looks like the right 'basic' style sheet isn't always being used for the version of the file, which could cause unpredictable things to happen when subsequent changes are applied.

vpereverzev (approved):

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - PASS

Issue #10209, opened 2022-01-04
When adding slurs in MS4, you cannot use Keyboard shortcuts to extend / contract them as with MS3


Tantacrul:
**Describe the bug**
The desired functionality was actually working in MS4 until a few days ago. I was away on holiday so didn't notice exactly when this happened. Basically, if you create a slur (I'm doing this by pressing 'S' in the attached video), you should be able to extend / contract it immediately using the shortcuts: **Shift+Command+Left/Right arrow** or **Shift+Alt+Left/Right arrow**.

This is also true for hairpins, pedal marks and various other lines. It is one of the most important shortcuts in the app and must function the same way as MS3. This is related to the bug: https://github.com/musescore/MuseScore/issues/10169

**To Reproduce**
See video

https://user-images.githubusercontent.com/21022311/148116986-dbc6f1c1-c53c-4320-8f1b-8b888eda2d28.mov

Mac: Catalina.

wizofaus:
I believe https://github.com/musescore/MuseScore/pull/10185 fixes this (was reported as https://github.com/musescore/MuseScore/issues/9902)

DmitryArefiev:
Tested  on Win10, Mac12, LinuxMint 20.2 - FIXED in #10401

Tantacrul:
Awesome :)

PR #10401, opened 2022-01-28
[MU4] Fix #9902: Can't adjust line or slur length via keyboard

### LINKED ISSUE #9902: [MU4 Issue] Can't adjust line or slur length via keyboard

RomanPudashkin:
Resolves: 
https://github.com/musescore/MuseScore/issues/10209
https://github.com/musescore/MuseScore/issues/9902

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -1938,10 +1938,10 @@ void Score::cmdAddOttava(OttavaType type)
 }
 
 //---------------------------------------------------------
-//   addHairpin
+//   addHairpins
 //---------------------------------------------------------
 
-void Score::addHairpin(HairpinType type)
+std::vector<Hairpin*> Score::addHairpins(HairpinType type)
 {
     // special case for two selected chordrests on same staff
     bool twoNotesSameStaff = false;
@@ -1954,12 +1954,14 @@ void Score::addHairpin(HairpinType type)
         }
     }
 
+    std::vector<Hairpin*> hairpins;
+
     // add hairpin on each staff if possible
     if (selection().isRange() && selection().staffStart() != selection().staffEnd() - 1) {
         for (int staffIdx = selection().staffStart(); staffIdx < selection().staffEnd(); ++staffIdx) {
             ChordRest* cr1 = selection().firstChordRest(staffIdx * VOICES);
             ChordRest* cr2 = selection().lastChordRest(staffIdx * VOICES);
-            addHairpin(type, cr1, cr2, /* toCr2End */ true);
+            hairpins.push_back(addHairpin(type, cr1, cr2, /* toCr2End */ true));
         }
     } else if (selection().isRange() || selection().isSingle() || twoNotesSameStaff) {
         // for single staff range selection, or single selection,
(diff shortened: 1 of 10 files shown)


src/engraving/libmscore/edit.cpp
```diff
@@ -1959,7 +1959,7 @@ void Score::addHairpin(HairpinType type)
         for (int staffIdx = selection().staffStart(); staffIdx < selection().staffEnd(); ++staffIdx) {
             ChordRest* cr1 = selection().firstChordRest(staffIdx * VOICES);
             ChordRest* cr2 = selection().lastChordRest(staffIdx * VOICES);
-            addHairpin(type, cr1, cr2, /* toCr2End */ true);
+            return addHairpin(type, cr1, cr2, /* toCr2End */ true);
```
Jojo-Schmitz, line 1962: Wouldn't that leave the function in the first round of that `for` loop? The comment says `// add hairpin on each staff if possible`

RomanPudashkin: Thanks, I didn't notice this for loop :(

RomanPudashkin: fixed

DmitryArefiev:
Tested #10209 #9902 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

vpereverzev (approved):

PR #10408, opened 2022-01-30
Preserve the Theme setting while enabling/disabling High contrast

iamjithinjohn:
Resolves: #9220 

Add some cases to switch back to the same mode of Theme instead of switching back
to the previous mode of enabled-high-contrast or disabled-high-contrast modes.

### DIFF
--- a/src/framework/ui/internal/uiconfiguration.cpp
+++ b/src/framework/ui/internal/uiconfiguration.cpp
@@ -37,8 +37,6 @@ using namespace mu::framework;
 using namespace mu::async;
 
 static const Settings::Key UI_THEMES_KEY("ui", "ui/application/themes");
-static const Settings::Key UI_PREVIOUS_GENERAL_THEME("ui", "ui/application/previousGeneralTheme");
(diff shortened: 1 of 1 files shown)


cbjeukendrup (commented):
Although this PR in practice solves the issue, I would go for a solution that is a bit more rigorous. You can in fact eliminate both `UI_PREVIOUS_HIGH_CONTRAST_THEME` and `UI_PREVIOUS_GENERAL_THEME`, since those values can now be calculated rather than that they need to be stored and retrieved. 

So, the body of `UiConfiguration::setCurrentTheme` can be cut down to just `settings()->setSharedValue(UI_CURRENT_THEME_CODE_KEY, Val(codeKey));`. Instead, some work needs to be done in `UiConfiguration::setIsHighContrast`.

iamjithinjohn:
Oh yes, I got that now. I didn't notice that those values would be of no use when we are just switching between the two **based** on the previous choice :neutral_face: 

Thanks for pointing that out. I'll update the PR with those changes.

cbjeukendrup (approved):

DmitryArefiev:
Tested #9220 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Thanks!

PR #10439, opened 2022-02-03
[MU4] notation_playback_layer_p1

vpereverzev:


src/framework/mpe/mpetypes.h
```diff
@@ -476,6 +504,15 @@ struct ArticulationMeta
     duration_t overallDuration = 0;
     pitch_level_t overallPitchChangesRange = 0;
     dynamic_level_t overallDynamicChangesRange = 0;
+
+    bool operator==(const ArticulationMeta& other) const
+    {
+        return type == other.type
```
RomanPudashkin, line 510: What about timestamp? Should we add it to this comparison?

vpereverzev: sure, thanks

src/engraving/playback/playbackmodel.cpp
```diff
@@ -87,31 +94,84 @@ void PlaybackModel::load(Ms::Score* score, async::Channel<int, int, int, int> no
...
+    const PlaybackEventList& eventsFromTick = trackPlaybackData->second.originData.at(timestamp);
+
+    PlaybackEventsMap result;
+
+    if (item->isChord()) {
+        for (const Ms::Note* note : Ms::toChord(item)->notes()) {
+            findEventsForNote(note, eventsFromTick, result[timestamp]);
+        }
+    } else {
+        findEventsForNote(Ms::toNote(item), eventsFromTick, result[timestamp]);
```
RomanPudashkin, line 144: I would suggest adding a check here to see if the item is really a note

RomanPudashkin (approved):

src/engraving/playback/playbackmodel.cpp
vpereverzev: item->isPlayable() from above will ensure this

PR #10447, opened 2022-02-04
fix #10288: Save Copy works as intended

### LINKED ISSUE #10288: [MU4 Issue] `Save a copy...` replaces active score with copy
**Describe the bug**
`File` > `Save a copy...` currently replaces the original working score with the newly saved (copy) score.

**To Reproduce**
1. Save a new score
2. Click `File` > `Save a copy...`, and give your score a different filename
3. Click `Save`
4. The score with the new filename is now active in the project window

**Expected behavior**
Per MS3.6, the new file created by `Save a copy...` should actually only be saved in the user's designated file directory, leaving the original file (i.e. not the new file) active in the project window.
In other words, `Save a copy...` should work as per MS3.6.

**Demo video**
https://www.dropbox.com/s/tjiuhsnod2jcm64/Save%20a%20copy%20demo.mov?dl=0

 - MacOS 11.6 (M1)
 - Windows 11

VanSHOE:
Resolves: #10288 

The save a copy now makes another file and does not change the current project

I have moved the current code which was common for Save, SaveAs and SaveCopy without modifying it
Please let me know if that requires any modification (one of the conditions at line 378 (after the PR changes) in it is null and void, so I am not sure whether to remove it or if there is some better condition that should be put in its place)

The above mentioned condition is this (line 378)
`if (saveMode != SaveMode::SaveCopy || oldFilePath == savePath) `

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -360,14 +360,14 @@ mu::Ret NotationProject::save(const io::path& path, SaveMode saveMode)
     case SaveMode::SaveSelection:
         return saveSelectionOnScore(path);
     case SaveMode::Save:
-    case SaveMode::SaveAs:
-    case SaveMode::SaveCopy: {
+    case SaveMode::SaveAs: {
         io::path oldFilePath = m_engravingProject->path();
 
         io::path savePath = path;
         if (!savePath.empty()) {
             m_engravingProject->setPath(savePath.toQString());
-        } else {
+        }
+        else {
             savePath = m_engravingProject->path();
         }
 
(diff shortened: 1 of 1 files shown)


src/project/internal/notationproject.cpp
```diff
@@ -360,14 +360,14 @@ mu::Ret NotationProject::save(const io::path& path, SaveMode saveMode)
...
-    case SaveMode::SaveCopy: {
+    case SaveMode::SaveAs: {
         io::path oldFilePath = m_engravingProject->path();
 
         io::path savePath = path;
         if (!savePath.empty()) {
             m_engravingProject->setPath(savePath.toQString());
-        } else {
+        }
+        else {
```
VanSHOE, line 370: I acknowledge that I moved this else to the next line unintentionally, I will fix this alongside any other suggested fix

PR #10466, opened 2022-02-06
Prepared the possibility to have non-local scores

cbjeukendrup:
**SaveLocation**
The ultimate goal of this PR is to introduce the SaveLocation type. It describes the location of a project: it can be a local project, identified by a path on the disk, or a cloud project, identified by a URL or ID (but that part is not implemented yet).

**Paths**
It turned out we depend quite a lot on having a path. We even have a feature that can display the path in a score (by writing `$f` or `$F` in the header/footer fields). 
Also Images do something with the path of the score, trying to look up an image by a relative path, but I think we should remove that feature and always store images in the project file.
For the rest, many places don't use the EngravingProject yet, so the logic for paths is spread over MasterScore, EngravingProject and NotationProject. This is not a great situation.

**IFileInfoProvider**
So I created the `IFileInfoProvider` interface, that will provide the MasterScore and EngravingProject with information about the file. For places where we need to work with just a MasterScore or EngravingProject (mainly in unit tests), this will be a QFileInfoProvider, which works based on QFileInfo and just gives information about a file on disk.
Where we work with NotationProject, the NotationProject itself will be the IFileInfoProvider; for projects on disk, it will give information about a file on disk (quite similar to how QFileInfoProvider works), and for cloud projects, it will give information about the cloud project (but that part is not implemented yet).
(I still consider the possibility to create a separate ProjectFileInfoProvider class, to which the NotationProject will delegate the file info provision.)

**Behaviour changes**
So far, the behavior after this PR should be exactly the same as before. The only change so far, is that the window title for newly created scores will be the title of the new score instead of just "Untitled"; this replicates MU3 behaviour.

**What's next**
I think the next step should be to clean up the work with titles, excerpt names and filenames, because there are at least some confusing method names in the code, and again there are many "sources of truth": the filename, the thing returned by MasterScore::title(), the title from the "metatags". To clean this up, most changes are just internal and will not result in changes in behaviour; the only change I'll make in this area is that the list of Recent Scores will show a filename instead of title.


src/project/internal/notationproject.cpp
```diff
@@ -168,7 +166,7 @@ mu::Ret NotationProject::doLoad(engraving::MscReader& reader, const io::path& st
 
     // Create new engraving project
     EngravingProjectPtr project = EngravingProject::create();
-    project->setPath(reader.params().filePath);
+    project->setFileInfoProvider(shared_from_this());
```
Eism, line 169: There is a problem that I have fixed in this PR #10414
Passing a pointer to a child blocks deletion of the parent, because there are owners(child) of the parent

You should replace shared_from_this with weak_from_this

src/project/internal/notationproject.h
```diff
@@ -55,11 +56,21 @@ class NotationProject : public INotationProject
     NotationProject();
     ~NotationProject() override;
 
-    io::path path() const override;
-
     Ret load(const io::path& path, const io::path& stylePath = io::path(), bool forceMode = false) override;
     Ret createNew(const ProjectCreateOptions& projectInfo) override;
 
+    io::path path() const override;
```
RomanPudashkin, line 62: It is a bit confusing that this method is not part of IFileInfoProvider

cbjeukendrup: I did not put it in IFileInfoProvider because that would require to use io::path in the engraving module, and I thought we have avoided that so far (and I think IFileInfoProvider should be in the engraving module because its purpose is to provide the MasterScore and EngravingProject with all necessary info). 
But if we think it's okay to use io::path in engraving, then I could add `path()` to the IFileInfoProvider interface, although I wonder whether it will ever be used from there.

src/engraving/infrastructure/io/compat/qfileinfoprovider.h
```diff
@@ -0,0 +1,49 @@
...
+#include <QFileInfo>
+
+namespace mu::engraving {
+class QFileInfoProvider : public IFileInfoProvider
+{
+public:
+    explicit QFileInfoProvider(const QFileInfo& fileInfo);
+    explicit QFileInfoProvider(const QString& filePath);
+
+    QString absoluteDirPath() const override;
```
RomanPudashkin, line 36: A lot of these methods are already implemented in path.h. So we should use path.h instead of reimplementing these methods here

src/engraving/libmscore/masterscore.h
```diff
@@ -219,11 +219,8 @@ class MasterScore : public Score
 
     MasterScore* unrollRepeats();
 
-    QFileInfo* fileInfo() { return &info; }
-    const QFileInfo* fileInfo() const { return &info; }
-    void setName(const QString&);
-
-    const QFileInfo& sessionStartBackupInfo() const { return _sessionStartBackupInfo; }
+    IFileInfoProviderPtr fileInfo() const { return m_fileInfoProvider; }
```
RomanPudashkin, line 222: Better to move the implementation to .cpp

src/engraving/engravingproject.h
```diff
@@ -89,7 +83,7 @@ class EngravingProject : public std::enable_shared_from_this<EngravingProject>
 
     Err doSetupMasterScore(Ms::MasterScore* score);
 
-    QString m_path;
+    IFileInfoProviderPtr m_fileInfoProvider;
```
RomanPudashkin, line 86: I can't find where this member is used

src/project/internal/notationproject.h
RomanPudashkin: You can freely use the framework in the engraving module :) The main restriction is not to use Qt in the engraving module

src/project/internal/notationproject.h
```diff
@@ -90,6 +103,9 @@ class NotationProject : public INotationProject
     notation::MasterNotationPtr m_masterNotation = nullptr;
     ProjectAudioSettingsPtr m_projectAudioSettings = nullptr;
     ProjectViewSettingsPtr m_viewSettings = nullptr;
+
+    SaveLocation m_saveLocation = SaveLocation::makeInvalid();
+    QFileInfo m_cachedFileInfo;
```
RomanPudashkin, line 108: I think, EngravingProject can provide this information directly without any caches here because it already stores a pointer to IFileInfoProvider

src/project/internal/notationproject.cpp
cbjeukendrup: In some situations, the file info provider should be "owned" by the engraving project, and in some situations not. 

For example, when creating a standalone QFileInfoProvider, it should be deleted when the EngravingProject is destroyed. 
But when we set the NotationProject as the info provider for the engraving project, the engraving project should obviously not delete it. 

I think a solution would be to introduce a new `class ProjectFileInfoProvider : public IFileInfoProvider` that will be passed to the engraving project, instead of passing the NotationProject itself. This way, the ProjectFileInfoProvider could be owned by the EngravingProject without problems; and the ProjectFileInfoProvider can hold a weak reference to the NotationProject. Does that sound like a good idea?

RomanPudashkin (approved):

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - PASS

Please merge into master

PR #10470, opened 2022-02-06
Bug fix: incorrect layout of multimeasure rests

mike-spa:
Bug description:
Multimeasure rests are initially layed out correctly (hence why it went unnoticed), but during any editing operation their width can only increase, and never decrease, so they quickly get messed up (see screeshot). 
Screenshot from 2022-02-06 16-58-54

How to replicate:
Create multimeasure rests and play around with them using stretch commands or line breaks.

Solution:
The error started from this [commit](https://github.com/musescore/MuseScore/commit/427f573f34a39cc9f8c8244a022dd14d662d9670) of my horizontal spacing PR, specifically from eliminating this [block of code](https://github.com/musescore/MuseScore/commit/427f573f34a39cc9f8c8244a022dd14d662d9670#diff-52ba1d84eecac29ce33da0360e3bc07d8a7ae6f1b7230763caa1dcd31af4a9a9L4219-L4232). However, this code was simply covering a deeper error, namely the fact that the minimum horizontal distance of a segment was calculated via the width of the corresponding shape, but this doesn't make sense for MM rests because their width (unlike all other shapes) changes with context. I think this should fix it.

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -4212,7 +4212,7 @@ void Measure::computeWidth(Fraction minTicks, qreal stretchCoeff)
     //
     Shape ls(first ? RectF(0.0, -1000000.0, 0.0, 2000000.0) : RectF(0.0, 0.0, 0.0, spatium() * 4));
 
-    x = s->minLeft(ls);
+    x = s->isMMRestSegment() ? 0 : s->minLeft(ls);
 
     if (s->isStartRepeatBarLineType()) {
         System* sys = system();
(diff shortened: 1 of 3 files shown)


mike-spa:
(v test failure may be due to the width of MM rests now being slightly different, though the difference should be very small)

Marr11317:
some vtests look odd:
image

you can download the tests [here](https://github.com/musescore/MuseScore/suites/5196113531/artifacts/158452128)] and open the html file in the folder (once uncompressed) to view the diff.

Jojo-Schmitz:
Seems the vtest failure shows the improved/corrected layout and so a `vtests` label seems due

oktophonie (approved):

RomanPudashkin (approved):

PR #10471, opened 2022-02-06
Fix #10441 - Omit lyrics extender line if too short

### LINKED ISSUE #10441: Lyrics: Extender lines can be too short.
**Describe the bug**
I've noticed that extender lines for lyrics (created using the shortcut **Shift + -**) don't always range far enough:

Here's an example from MS3 and 4:
image

Here's an older example from MS2... although this only looks clearer partially because the font is narrower
image

**See video**
https://user-images.githubusercontent.com/21022311/152386503-db3beba9-e54d-49e1-b3fd-117b5b8c2032.mov

mike-spa:
Comment

Resolves: [10441](https://github.com/musescore/MuseScore/issues/10441)

As discussed in the related issue, the extender line for lyrics should be omitted when too short, as it serves no purpose. See the screen capture attached for a quick look. The threshold value is now hardcoded to 1 sp as suggested by @oktophonie. In future, it probably deserves a dedicated style setting.

https://user-images.githubusercontent.com/93707756/152693861-5bb7c9bc-e270-46bd-bb68-6c26afd77ca7.mp4

### DIFF
--- a/src/engraving/libmscore/lyricsline.cpp
+++ b/src/engraving/libmscore/lyricsline.cpp
@@ -417,8 +417,10 @@ void LyricsLineSegment::layout()
 
     // MELISMA vs. DASHES
     const double minMelismaLen = 1 * sp; // TODO: style setting
+    const double minDashLen  = score()->styleS(Sid::lyricsDashMinLength).val() * sp;
+    const double maxDashDist = score()->styleS(Sid::lyricsDashMaxDistance).val() * sp;
+    double len = pos2().rx();
     if (isEndMelisma) {                   // melisma
-        double len = pos2().rx();
         if (len < minMelismaLen) { // Omit the extender line if too short
             _numOfDashes = 0;
         } else {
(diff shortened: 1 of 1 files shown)


src/engraving/libmscore/lyricsline.cpp
```diff
@@ -416,8 +416,14 @@ void LyricsLineSegment::layout()
     }
 
     // MELISMA vs. DASHES
+    static double minMelismaLen = 1 * sp; // TODO: style setting
```
cbjeukendrup, line 419: I think it shouldn't be static since `sp` isn't static either. But it could be const.

mike-spa: Yeah, good point, it should be const. Though I think it can still be static: in this context it means that it is a property of the class, so it doesn't need to be reinstantiated by every lyricsLine object. Even if sp isn't static

cbjeukendrup: I still think it can't be static. `sp` may be different every time this method is called, so minMelismaLen must also be initialized/updated every time. Now, it will be initialized the first time this method is called, but after that, the value will never change anymore.

mike-spa: Mmh, you're right. I was assuming that sp is always the same within a given score, but indeed it can change. Then the variable should just be a const double

Jojo-Schmitz:
I wonder why no vtest fails, seems to indicate no such too short melisma is part of the tests, but maybe it better should?

mike-spa:
> I wonder why no vtest fails, seems to indicate no such too short melisma is part of the tests, but maybe it better should?

Yeah, I guess it's a relatively rare occurrence, but it may be worth having it in the tests. In general, is there a handbook or something on how to create new tests? That's something I'd like to learn.

Jojo-Schmitz:
Check vtest/scores/lyrics-[1-9].mscx

oktophonie (approved):

RomanPudashkin (approved):

Issue #9189, opened 2021-09-19
[MU4 Issue] Recent scores list shows score title (instead of filename) with trailing \ and too much vertical space


wizofaus:
**Describe the bug**
Recent scores list shows the score title with a trailing \ and too much vertical space

**To Reproduce**
1. Start musescore, access File|Open
2. Type "<git workspace root>\src\engraving\tests\unrollrepeats_data\clef-key-ts-test.mscx" into the File name field and hit Enter
3. Browser to File | Open Recent and observe first entry in recent scores list - shows as `Clef-Key-TS test case\`

**Expected behavior**
Should just show "clef-key-ts-test.mscx" with normal vertical spacing.
(While I do understand the appeal of showing the Title rather than the filename, it's not adequate if you have various copies of the same score saved with different filenames for various reasons, and it looks odd if some of your scores don't have titles - in fact my Recent files list has just the word "Title" occurring about 7 times and I have no idea which is which.  And there's very few other applications that do this, even v3 shows the file name)

**Screenshots**
![image](https://user-images.githubusercontent.com/5364225/133915108-a3ed3e95-64a7-4cde-a1ed-c244ac8bcfca.png)

**Desktop (please complete the following information):**
 - Windows 10

**Additional context**

riaanvn:
Seems to be a duplicate of https://github.com/musescore/MuseScore/issues/8241

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED in #10473

PR #10473, opened 2022-02-06
Clean up the work with "(work) titles" and "names"

cbjeukendrup:
Resolves: #8241
Resolves: #9189 

Builds on: #10466

Now, we have the following things and use them consequently
- Work title: the title of the score, stored in metatags
- Name:
  - For `Excerpt` scores ("parts"): the name of the `Excerpt`, for example, "Flute", or "Violin II"
  - For MasterScores: the filename without extension
  used in the notation tab bar, and as the suggested filename when exporting a score;
- Filename (with extension); now used in the following places:
  - Recent Scores list
  - Home page
  - Window title


src/engraving/libmscore/excerpt.h
```diff
@@ -20,84 +20,79 @@
...
     int nstaves() const;
     bool isEmpty() const;
 
-    QMultiMap<int, int>& tracks() { return _tracks; }
-    void setTracks(const QMultiMap<int, int>& tracks);
-
-    MasterScore* oscore() const { return _oscore; }
-    Score* partScore() const { return _partScore; }
-    void setPartScore(Score* s);
+    QMultiMap<int, int>& tracksMapping() { return m_tracksMapping; }
```
njvdberg, line 65: Renaming `tracks()` into `tracksMapping()` seems to be an unrelated change and should be a separate change?

njvdberg: Same for changing other methods in `excerps.h`/`excerpts.cpp`, like `oscore()` and `partScore()`.

cbjeukendrup: Yes, you're right, I should better split that commit into a separate PR.

src/notation/iexcerptnotation.h
```diff
@@ -36,8 +36,8 @@ class IExcerptNotation
 
     virtual bool isCreated() const = 0;
 
-    virtual QString title() const = 0;
-    virtual void setTitle(const QString& title) = 0;
+    virtual QString name() const = 0;
+    virtual void setName(const QString& title) = 0;
```
RomanPudashkin, line 40: You forgot to rename the argument

RomanPudashkin (approved):

src/notation/iexcerptnotation.h
cbjeukendrup: Thanks, fixed!

PR #10483, opened 2022-02-07
[MU4] Slur Revamp

asattely:
This is a first step on the journey to having better-handled slurs. There are still some strange behaviors that need to be sussed out and fixed, but we don't have a lot of time to spend making it perfect, so for the time being, here are the changes.
- "Floating" endpoint vertical positioning fixed
- Collision avoidance between stem-side slurs and hooks
- Collisions between slurs and ties no longer happen
- Collisions between slurs and other slurs (both nested and similar start/end points)
- Slur endpoints now anchor to the middle of tenuto marks if necessary
- Endpoint _and_ height adjustment for slurs interacting with objects on the staff (this is one that still needs some work)
- Endpoint staff line avoidance

I expect the vtests to explode in a fiery ball of failure, by the way, so that's next

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -39,6 +39,7 @@
 #include "libmscore/tuplet.h"
 #include "libmscore/volta.h"
 #include "libmscore/system.h"
+#include "libmscore/slur.h"
 
 #include "layoutbeams.h"
 #include "layoutharmonies.h"
@@ -885,6 +886,10 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
     Fraction etick = useRange ? lc.endTick : system->measures().back()->endTick();
     auto spanners = score->spannerMap().findOverlapping(stick.ticks(), etick.ticks());
 
+    // ties
+    doLayoutTies(system, sl, stick, etick);
+
+    // slurs
     std::vector<Spanner*> spanner;
     for (auto interval : spanners) {
         Spanner* sp = interval.value;
@@ -919,18 +924,6 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
 
     std::vector<Dynamic*> dynamics;
     for (Segment* s : sl) {
-        for (EngravingItem* e : s->elist()) {
-            if (!e) {
-                continue;
-            }
-            if (e->isChord()) {
-                Chord* c = toChord(e);
-                for (Chord* ch : c->graceNotes()) {
-                    layoutTies(ch, system, stick);
-                }
-                layoutTies(c, system, stick);
-            }
-        }
         for (EngravingItem* e : s->annotations()) {
             if (e->isDynamic()) {
                 Dynamic* d = toDynamic(e);
@@ -1187,6 +1180,22 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
     }
 }
 
+void LayoutSystem::doLayoutTies(System* system, std::vector<Ms::Segment*> sl, const Fraction& stick, const Fraction& etick)
+{
+    for (Segment* s : sl) {
+        for (EngravingItem* e : s->elist()) {
+            if (!e || !e->isChord()) {
+                continue;
+            }
+            Chord* c = toChord(e);
+            for (Chord* ch : c->graceNotes()) {
+                layoutTies(ch, system, stick);
+            }
+            layoutTies(c, system, stick);
+        }
+    }
+}
+
 void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, bool align)
 {
     std::vector<SpannerSegment*> segments;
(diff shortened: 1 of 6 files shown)


src/engraving/layout/layoutsystem.cpp
```diff
@@ -920,7 +938,7 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
                 }
                 layoutTies(c, system, stick);
             }
-        }
+        }*/
```
Jojo-Schmitz, line 941: Why not just deleting these lines and let git keep the history?

src/engraving/libmscore/slur.cpp
```diff
@@ -460,37 +527,96 @@ void SlurSegment::layoutSegment(const PointF& p1, const PointF& p2)
...
             computeBezier();
-            intersection = false;
-            gdist = 0.0;
         }
-        if (intersection && gdist > 0.0) {
+        /*if (intersection && gdist > 0.0) {
             qreal min = minDistance + gdist;
             rypos() += up ? -min : min;
-        }
+        }*/
```
Jojo-Schmitz, line 619: same here

src/engraving/libmscore/slur.cpp
```diff
@@ -696,14 +838,25 @@ void Slur::slurPos(SlurPos* sp)
             sa2 = SlurAnchor::STEM;
         }
     }
+#else
```
Jojo-Schmitz, line 841: and here

src/engraving/libmscore/slur.cpp
```diff
@@ -696,14 +838,25 @@ void Slur::slurPos(SlurPos* sp)
...
     // also link start of slur to stem if start chord & slur are in same direction and there is a hook
+    /*
     if (scr->up() == _up && stem1 && sc->hook()) {
         sa1 = SlurAnchor::STEM;
         // if end chord is in same direction, link end of slur to stem too
         if (ecr->up() == scr->up() && stem2 && (!ecr->beam() || !ecr->beam()->cross())) {
             sa2 = SlurAnchor::STEM;
         }
-    }
+    }*/
```
Jojo-Schmitz, line 859: and here

src/engraving/layout/layoutsystem.cpp
asattely: good catch, missed that one

oktophonie (approved):

src/engraving/libmscore/slur.cpp
Jojo-Schmitz: Still this commented out code?

src/engraving/layout/layoutsystem.cpp
```diff
@@ -882,6 +883,23 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
     Fraction etick = useRange ? lc.endTick : system->measures().back()->endTick();
     auto spanners = score->spannerMap().findOverlapping(stick.ticks(), etick.ticks());
 
+    // ties
+    for (Segment* s : sl) {
+        for (EngravingItem* e : s->elist()) {
+            if (!e) {
```
vpereverzev, line 889: if (!e || !e->isChord())

src/engraving/libmscore/system.cpp
```diff
@@ -1804,6 +1804,52 @@ qreal System::lastNoteRestSegmentX(bool trailing)
...
+//---------------------------------------------------------
+//   firstChordRest
+//    returns the last chordrest of a system for a particular track
+//---------------------------------------------------------
+
+ChordRest* System::firstChordRest(int track)
+{
+    qreal margin = score()->spatium();
+    for (const MeasureBase* mb : measures()) {
+        if (mb->isMeasure()) {
```
vpereverzev, line 1839: if (!mb->isMeasure())

src/engraving/layout/layoutsystem.cpp
```diff
@@ -882,6 +883,23 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
...
+                continue;
+            }
+            if (e->isChord()) {
```
Jojo-Schmitz, line 892: or `if (e && e->isChord())` here, I guess.

src/engraving/libmscore/system.cpp
Jojo-Schmitz: Huh? In that case we'd be castring seomthing that isn't a measure into a measure??

src/engraving/libmscore/slur.h
```diff
@@ -52,7 +52,7 @@ class SlurSegment final : public SlurTieSegment
     bool edit(EditData&) override;
 
     Slur* slur() const { return toSlur(spanner()); }
-
+    void adjustEndpoints();
```
vpereverzev, line 55: rename to adjustEnd**P**oints pls

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1219,6 +1227,43 @@ void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, boo
         }
     }
 
+    if (slurSegments.size() > 1) {
+        const qreal slurCollisionVertOffset = 0.65 * system->spatium();
+        const qreal fuzzyHorizCompare = 0.1 * system->spatium();
+        auto compare = [fuzzyHorizCompare](qreal x1, qreal x2) { return qAbs(x1 - x2) < fuzzyHorizCompare; };
```
vpereverzev, line 1233: always prefer STL algorithms

qAbs -> std::abs

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1219,6 +1227,43 @@ void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, boo
         }
     }
 
+    if (slurSegments.size() > 1) {
+        const qreal slurCollisionVertOffset = 0.65 * system->spatium();
```
vpereverzev, line 1231: it's not quite clear, why you have chosen these specific number. Makes sense to have named constexpr variables

vpereverzev: or a comment, at least

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1187,10 +1191,14 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
 void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, bool align)
 {
     std::vector<SpannerSegment*> segments;
+    std::vector<SlurSegment*> slurSegments;
```
vpereverzev, line 1194: instead of filling and allocating a new vector of these pointers, you could just put a check to the body of the 'for loop" at 1234

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1219,6 +1227,43 @@ void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, boo
         }
     }
 
+    if (slurSegments.size() > 1) {
+        const qreal slurCollisionVertOffset = 0.65 * system->spatium();
+        const qreal fuzzyHorizCompare = 0.1 * system->spatium();
+        auto compare = [fuzzyHorizCompare](qreal x1, qreal x2) { return qAbs(x1 - x2) < fuzzyHorizCompare; };
+        for (SlurSegment* slur1 : slurSegments) {
+            for (SlurSegment* slur2 : slurSegments) {
+                // slurs don't collide with themselves
```
vpereverzev, line 1236: this one looks over complicated and not performant

in general, I think you need to compare only adjacent elements by x point. So that it makes sense to prepare a std::map<int /*x*/, SlurSegment*> (which is sorted by default) and then compare adjacent elements

asattely: I agree that this is cumbersome, but in this loop I test both endpoints, so there could be slurs which are non-adjacent in terms of start point, but which end on the same chordrest:
image
So to test the endpoint collisions every slur segment needs to be compared to every other slur segment as far as I can tell. If there is a better way to do that, please let me know!

src/engraving/libmscore/system.cpp
```diff
@@ -1808,6 +1808,53 @@ qreal System::lastNoteRestSegmentX(bool trailing)
...
+
+ChordRest* System::firstChordRest(int track)
+{
+    qreal margin = score()->spatium();
+    for (const MeasureBase* mb : measures()) {
+        if (!mb->isMeasure()) {
+            continue;
+        }
+        const Measure* measure = static_cast<const Measure*>(mb);
+        for (const Segment* seg = measure->first(); seg; seg = seg->next()) {
```
Jojo-Schmitz, line 1847: Doesn't build, at least not with MSVC: 
"first()" is not a member of 'stdvector<Ms::ChordRest *, std::allocator<Ms::ChordRest *>>'

src/engraving/libmscore/slur.cpp
```diff
@@ -765,11 +912,16 @@ void Slur::slurPos(SlurPos* sp)
             pt.rx() = hw2 * 0.5;
         }
         // don't allow overlap with beam
-        qreal yadj = ec->beam() ? 0.75 : -0.25;
+        qreal yadj;
+        if (ec->beam() && ec->beam()->elements().first() != ec) {
```
Jojo-Schmitz, line 916: Doesn't build, at least not with MSVC:
"first()" is not a member of 'stdvector<Ms::ChordRest *, std::allocator<Ms::ChordRest *>>'

Jojo-Schmitz: Should it be `        if (ec->beam() && ec->beam()->elements().front() != ec) {` instead?

Jojo-Schmitz: That actually dows work, see #10851

asattely: After a bit of investigation, commit https://github.com/musescore/MuseScore/commit/94d7ee71f85e7fddc342bfe4208a499af5476133 seems to have changed QVector (which does have a `first` method) with std::vector (which does not). Your suggestion is correct!

PR #10488, opened 2022-02-08
Fix MS #314800 Add Ledger Lines to Key Signatures

sammik:
Resolves: *([MS #314800](https://musescore.org/en/node/314800))*

Added ledger lines to key signatures. If accidental was above, or belove staff, ledger lines were missing and it was difficualt to recognize its position. Now it is fixed.

~~Possible conflicts with #10609, will need to rebase on top of it.~~
Rebased on top of current master.

### DIFF
--- a/src/engraving/libmscore/keysig.cpp
+++ b/src/engraving/libmscore/keysig.cpp
@@ -26,6 +26,7 @@
 #include "rw/xml.h"
 #include "types/symnames.h"
 #include "types/typesconv.h"
+#include "draw/pen.h"
 
 #include "staff.h"
 #include "clef.h"
@@ -379,13 +380,30 @@ void KeySig::layout()
 void KeySig::draw(mu::draw::Painter* painter) const
 {
     TRACE_OBJ_DRAW;
+    using namespace mu::draw;
     painter->setPen(curColor());
     double _spatium = spatium();
     double step = _spatium * (staff() ? staff()->staffTypeForElement(this)->lineDistance().val() * 0.5 : 0.5);
+    int lines = staff() ? staff()->staffTypeForElement(this)->lines() : 5;
+    double ledgerLineWidth = score()->styleMM(Sid::ledgerLineWidth) * mag();
+    double ledgerExtraLen = score()->styleS(Sid::ledgerLineLength).val() * _spatium;
     for (const KeySym& ks: _sig.keySymbols()) {
         double x = ks.xPos * _spatium;
         double y = ks.line * step;
         drawSymbol(ks.sym, painter, PointF(x, y));
+        // ledger lines
+        double _symWidth = symWidth(ks.sym);
+        double x1 = x - ledgerExtraLen;
+        double x2 = x + _symWidth + ledgerExtraLen;
+        painter->setPen(Pen(curColor(), ledgerLineWidth, PenStyle::SolidLine, PenCapStyle::FlatCap));
+        for (int i = -2; i >= ks.line; i -= 2) { // above
+            y = i * step;
+            painter->drawLine(LineF(x1, y, x2, y));
+        }
+        for (int i = lines * 2; i <= ks.line; i += 2) { // below
+            y = i * step;
+            painter->drawLine(LineF(x1, y, x2, y));
+        }
     }
     if (!explicitParent() && (isAtonal() || isCustom()) && _sig.keySymbols().empty()) {
         // empty custom or atonal key signature - draw something for palette


sammik:
KeySig-LedgerLines

src/engraving/libmscore/keysig.cpp
```diff
@@ -315,6 +315,18 @@ void KeySig::draw(mu::draw::Painter* painter) const
...
         drawSymbol(ks.sym, painter, PointF(ks.pos.x(), ks.pos.y()));
+        // draw ledgerLines
+        double _spatium = spatium();
+        double xshift = (ks.sym == SymId::accidentalSharp) ? (_spatium * .15) : (_spatium * .25);
+        int _line = int(ks.pos.y() / _spatium);
+        while (_line < 0) {
+            drawSymbol(SymId::legerLine, painter, PointF((ks.pos.x() - xshift), (_line * _spatium)));
+            ++_line;
+        }
+        while (_line >= 5) {
```
cbjeukendrup, line 326: Here you assume that a staff has 5 lines, but that might not always be the case. Better use `staff() ? staff()->staffTypeForElement(this)->lines() : 5` instead of `5`. As an optimization, you could take that out of the for-loop and store it in a variable.

cbjeukendrup: Seems like you'll also need to account for staff line distance. Look at this failing VTest for example:
Schermafbeelding 2022-02-08 om 11 06 49
It's just a 5-line staff, but with a staff line spacing of 2.50 sp.

Jojo-Schmitz: Wouldn't `staff() ? staff()->staffTypeForElement(this)->lineDistance().val() : 1.0;` help here?

cbjeukendrup: Personally I would expect the distance between the ledger lines to be the same as the distance between the staff lines. That of course relates to the other bug you just discovered. If you would like to solve that bug yourself (while on it), I could assign it to you.

cbjeukendrup: Also, that VTest still seems to fail the same way as before... you will indeed need to use `staff() ? staff()->staffTypeForElement(this)->lineDistance().val() : 1.0;`, as Jojo suggests.

sammik: Yes, distance should be same, as distance between lines, but as you can see, ledger lines corresponds with accidentals. I will chek something. Thank You.
One question `staff()->staffTypeForElement(this)->lineDistance().val()` or `staff()->lineDistance(tick())` would be OK?

Jojo-Schmitz: The latter looks simpler, but for symmetry alone I'd rather use the former?

cbjeukendrup: The former contains a small optimization compared to the latter, because it avoids a call to `tick()` in case there is only one staff type on that staff (which is the case in most scores, I'd think). For some elements, `tick()` is a very trivial call, but for others, it involves a bit more.

src/engraving/libmscore/keysig.cpp
```diff
@@ -315,6 +315,19 @@ void KeySig::draw(mu::draw::Painter* painter) const
     painter->setPen(curColor());
     for (const KeySym& ks: _sig.keySymbols()) {
         drawSymbol(ks.sym, painter, PointF(ks.pos.x(), ks.pos.y()));
+        // draw ledgerLines
+        qreal _spatium = spatium();
+        qreal xshift = (ks.sym == SymId::accidentalSharp) ? (_spatium * .15) : (_spatium * .25);
```
Jojo-Schmitz, line 320: Maybe make this qreal x = ks.pos.x() - (ks.sym == SymId::accidentalSharp) ? (_spatium * .15) : (_spatium * .25); and use that in the 2 drawSymbol() calls?

sammik: In fact, they are never called both, every time just one of them, so what is the reason for this change? Pretyfng code?

Jojo-Schmitz: Ah, I missed that

Jojo-Schmitz: Won't harm anyway

sammik:
I hope, finaly its fixed.
Thank You @Jojo-Schmitz and @cbjeukendrup for Your patience and help!
Ledger_spacing
(It looks may be weird a bit, because of that other bug.)

Jojo-Schmitz:
Hmm, my backport to 3.x doesn't work well:
image

Jojo-Schmitz:
Have a sample score to test this? Preferably for MuseScore 3?

sammik:
> Hmm, my backport to 3.x doesn't work well: image using the sample score from #10497, maybe that's why.
> 
> Have a sample score to test this? Preferably for MuseScore 3?
Somehow one line became wrong in Your backport. 
https://github.com/Jojo-Schmitz/MuseScore/commit/4ba24e6e45ce966e168bc55d5bdf2cea303190a9#r66315029

src/engraving/libmscore/keysig.cpp
```diff
@@ -313,8 +313,22 @@ void KeySig::draw(mu::draw::Painter* painter) const
...
     TRACE_OBJ_DRAW;
     painter->setPen(curColor());
+    qreal _spatium = spatium();
+    int lines = staff() ? staff()->staffTypeForElement(this)->lines() : 5;
+    qreal lineDist = staff() ? staff()->staffTypeForElement(this)->lineDistance().val() : 1;
     for (const KeySym& ks: _sig.keySymbols()) {
         drawSymbol(ks.sym, painter, PointF(ks.pos.x(), ks.pos.y()));
+        // draw ledgerLines
+        qreal x = ks.pos.x() - ((ks.sym == SymId::accidentalSharp) ? (_spatium * .15) : (_spatium * .25));
+        int i = int(ks.spos.y() / lineDist);
```
Jojo-Schmitz, line 323: This kind of casts are frowned upon, in master at least, better use the more verbose `static_cast<int>(...)`

cbjeukendrup:
@sammik I think it would be good if we would merge #10609 first, and then rebase this PR on top of it. So could you please fix the review comments on that PR?

sammik:
> @sammik I think it would be good if we would merge #10609 first, and then you rebase this PR on top of it. So could you please fix the review comments on that PR?

Sorry, I was playing. 
It should be fixed, so once it will be merged, Ill rebase this one.

sammik:
vtests failure is good

before
custom-keysig-2-1 ref

after
custom-keysig-2-1

cbjeukendrup:
I had another look at this today, and I thought, maybe we should not use the `SymId::legerLine` etc. symbols, but instead, just draw lines, using the `Painter` object. This enables us to make the line length dynamic, so that it will be good for every accidental regardless of how wide it is. It also enables us to follow the ledger line thickness specified in the style settings. This is how it is done for chords too. Probably a good example of this is in `ShadowNote::draw`. The relevant `Sid`s are `Sid::ledgerLineLength` (the amount to be added at each side of the accidental) and `Sid::ledgerLineWidth` (the line thickness).

sammik:
> I had another look at this today, and I thought, maybe we should not use the `SymId::legerLine` etc. symbols, but instead, just draw lines, using the `Painter` object. This enables us to make the line length dynamic, so that it will be good for every accidental regardless of how wide it is. It also enables us to follow the ledger line thickness specified in the style settings. This is how it is done for chords too. Probably a good example of this is in `ShadowNote::draw`. The relevant `Sid`s are `Sid::ledgerLineLength` (the amount to be added at each side of the accidental) and `Sid::ledgerLineWidth` (the line thickness).

Done!

PR #10511, opened 2022-02-09
Fix #10490: Spacing breaks by changing minimum measure width

### LINKED ISSUE #10490: [MU4 Issue] Spacing within measure is braking after changing minimum measure width value in Style settings
**To Reproduce**
1. Open a score
2. Go to **Style**>**Measure** and change **Minimum measure width**

Capture

https://user-images.githubusercontent.com/90187801/152981868-2948fbd2-3d71-4e36-aa3d-58bc5471176d.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Broken recently

mike-spa:
Resolves [10490](https://github.com/musescore/MuseScore/issues/10490)

The changes to horizontal spacing had broken this functionality. It is now restored. Also the minimum width of multimeasure rests should now work as expected.

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -391,10 +391,14 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     qreal stretchCoeff = 1;
     qreal prevWidth = 0;
     int iter = 0;
-    static double epsilon = score->spatium() * 0.05; // For reference: this is smaller than the width of a note stem
-    static constexpr int maxIter = 400; // Limits the number of iterations, just for safety. In reality, less than 3 iterations are required on average.
+    double epsilon = score->spatium() * 0.05; // For reference: this is smaller than the width of a note stem
     static constexpr float multiplier = 1.4; // Empirically optimized value which allows the fastest convergence of the following algorithm.
-
+    static constexpr int maxIter = 400;
+    // Different systems need different numbers of iterations of the following loop to reach the target width.
+    // The average is less than 3 iterations, and the maximum I've ever seen (very rare) is 30-40 iterations.
+    // maxIter just serves as a safety exit to not get stuck in the loop in case a system can't be justified
+    // (which can only happen if errors are made before getting here). It's set to a very high value to make
+    // that the system really can't be justified, and it isn't just a "tricky" one needing more iterations.
     while (abs(newRest) > epsilon && iter < maxIter) {
         stretchCoeff *= (1 + multiplier * newRest / curSysWidth);
         for (MeasureBase* mb : system->measures()) {
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -393,12 +393,12 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     int iter = 0;
     double epsilon = score->spatium() * 0.05; // For reference: this is smaller than the width of a note stem
     static constexpr float multiplier = 1.4; // Empirically optimized value which allows the fastest convergence of the following algorithm.
-    static constexpr int maxIter = 400;
+    static constexpr int maxIter = 100;
     // Different systems need different numbers of iterations of the following loop to reach the target width.
     // The average is less than 3 iterations, and the maximum I've ever seen (very rare) is 30-40 iterations.
     // maxIter just serves as a safety exit to not get stuck in the loop in case a system can't be justified
     // (which can only happen if errors are made before getting here). It's set to a very high value to make
-    // that the system really can't be justified, and it isn't just a "tricky" one needing more iterations.
+    // sure that the system really can't be justified, and it isn't just a "tricky" one needing more iterations.
     while (abs(newRest) > epsilon && iter < maxIter) {
         stretchCoeff *= (1 + multiplier * newRest / curSysWidth);
         for (MeasureBase* mb : system->measures()) {
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -169,9 +169,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // check if lc.curMeasure fits, remove if not
         // collect at least one measure and the break
 
-        static constexpr double acceptanceRange = 1.025; // We allow the initial width of the system to be slightly
-        // larger than the target system width. This avoids having to put a measure in the next system just because
-        // it overshoots the width by a tiny amount.
+        double acceptanceRange = system->isSqueezable() ? 1.025 : 1; // A value slightly larger than 1 allows systems
+        // to be initially slightly larger than the target width and be justified by tightening rather than stretching.
+        // However, we must first make sure that the system *can* be tightened. isSqueezable() checks for that.
         bool doBreak = (system->measures().size() > 1) && ((curSysWidth + ww) > systemWidth * acceptanceRange);
         if (doBreak) {
             breakMeasure = ctx.curMeasure;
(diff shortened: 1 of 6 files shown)


Jojo-Schmitz:
"breaks", not "brakes" I think

Jojo-Schmitz:
Seems a `vtest` label is in order, the failed vtests are actually good

oktophonie:
Still a bit of oddness here; for example, I increased the min measure width to 13sp:
image
and the systems (except first and last) extend into the right margin. This also happens with various higher values.

mike-spa:
Oh, bummer. That is a consequence of the "acceptance range" I've implemented, which lets systems be initially *slightly* larger (3-4 mm, in this case) than the target width and be justified by "tightening" rather than stretching, which is good for making smarter choices on how many measures fit on a system. This, unfortuntely, is the single edge case which breaks it: every single measure being locked by the minimum measure width, so the system can't be justified by tightening, because none of the measures can be narrower. Accounting for this case would require lots of work, which would probably need a separate PR and separate testing. 
For now, I'm going to roll back on that funcionality, so measures will be justified only by stretching and this case won't be a problem. I've also improved the efficiency of the justification algorithm overall. I expect these changes to cause further v-test failures (commit coming soon).

mike-spa:
(utests are failing on beams, like my first PR. Fix coming soon.)

oktophonie (approved):

src/engraving/libmscore/measure.cpp
```diff
@@ -3968,16 +3968,40 @@ void Measure::checkTrailer()
...
+    const double minWidth = isMMRest() ? score()->styleMM(Sid::minMMRestWidth) : score()->styleMM(Sid::minMeasureWidth);
+
+    Segment* s = nullptr;
+    double newSegWidth = 0.0;
+    double diff = 0.0;
+    int iter = 0;
+    double x = 0.0;
+    double multiplier = 1.1; // Empirical factor for fast convergence of the following loop
+
+    while (w < minWidth && iter < 200) { // Just for safety. In practice, it rarely takes more than 2 iter
```
RomanPudashkin, line 3987: This algorithm looks too overcomplicated. It does a lot of things at the same time. Ideally, we should have something like this:

```
void EngravingItem::setWidth(w) {
m_width = std::max(w, minWidth())
}
```
And this is roughly what the previous implementation looked like

mike-spa: Sorry for my late reply. Yes, that method was messy, sorry. I think I've now come up with a different, more elegant solution. One that also allows to maintain the "acceptance range" in system justification while accounting for the edge case (code coming soon).

On a more general note though: the previuos implementation wasn't actually simpler, it was hiding the complexity elsewhere. Now, because of many reasons connected to my spacing improvements (I'm happy to go into all details if needed), that specific "elsewhere" is not available anymore, so I need to perform more operations here, making the code a bit more cumbersome. I believe it's a worthy compromise though.

mike-spa:
Working on this issue has now allowed me to improve on other aspects of the code connected to it. The justification of the measures to `minMeasureWidth` or `minMMRestWidth` is now performed differently, in a way that better integrates with the previous spacing improvements. 
Most notably, I've introduced the `bool` member `Measure::isWidthLocked()` which is true for measures whose width is "locked" by the `minMeasureWidth` parameter. Having this information greatlly helps optimizing layout operations, particularly system justification.

oktophonie (approved):

MarcSabatella:
FWIW, isWidthLocked() sounds like a first step to implementing a fixed measure width facility - whether that's specificying an absolute width in measure properties, dragging a barline, or a style setting to say you want equal measure widths on each system.  The latter in particular is a common request, since it's found often in handwritten jazz charts and in music for children, and people expect software to be able to emulate that look as well.

Jojo-Schmitz:
That "fixed measure width facility" got lost in MuseScore 2 (along with a fixed number of measures per system, IMHO on the latter part was probelmatic), was available in MuseScore 1.

mike-spa:
> FWIW, isWidthLocked() sounds like a first step to implementing a fixed measure width facility - whether that's specificying an absolute width in measure properties, dragging a barline, or a style setting to say you want equal measure widths on each system. The latter in particular is a common request, since it's found often in handwritten jazz charts and in music for children, and people expect software to be able to emulate that look as well.

You just made me realise that part of the code I've now written has potentially much wider applications than I thought, so it definitely deserves to be on a dedicated method. Thanks!

oktophonie:
100% agree with Marc about the need for some sort of 'system config' object (which could be the same thing as a system break or connected to it) for exactly the reasons he describes - plus it can include details about bracketing or labelling, vertical justification, all sorts of fun stuff like that.

src/engraving/layout/layoutsystem.cpp
```diff
@@ -391,18 +391,20 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     qreal stretchCoeff = 1;
     qreal prevWidth = 0;
     int iter = 0;
-    static double epsilon = score->spatium() * 0.08; // For reference: this is approximately as small as the width of a note stem
-    static constexpr int maxIter = 200; // Limits the number of iterations, just for safety. In reality, most systems require less then 10 iterations.
-    static constexpr float multiplier = 1.5; // Empirically optimized value which allows the fastest convergence of the following algorithm.
+    static double epsilon = score->spatium() * 0.05; // For reference: this is smaller than the width of a note stem
```
cbjeukendrup, line 394: This shouldn't be static because `score` is not static either. (otherwise, the spatium of the very first score in the program will be used for the rest of all time)

src/engraving/libmscore/measure.cpp
```diff
@@ -4258,6 +4239,31 @@ void Measure::computeWidth(Fraction minTicks, qreal stretchCoeff)
...
+        setWidthLocked(false);
+    }
+}
+
+void Measure::setWidthToTargetValue(Segment* s, qreal x, bool isSystemHeader, Fraction minTicks, qreal stretchCoeff, qreal targetWidth)
+{
+    // The input parameters of this method rely on Measure::computeWidth(), so always call this method from there
+    const double epsilon = 0.1 * spatium();
+    static constexpr double multiplier = 1.4; // Empirical value for fastest convergence of the following loop
+    static constexpr int maxIter = 200; // Just for safety, in reality just 2 iterations are typically needed.
```
RomanPudashkin, line 4260: Could you please explain why you use this particular number here? It is much greater than the number mentioned in the comment. It gives the feeling that this algorithm will work unpredictably

RomanPudashkin: The same question for 

static constexpr int maxIter = 400;

in LayoutSystem::collectSystem

src/engraving/layout/layoutsystem.cpp
mike-spa: Thanks, that dates back to my first PR. I'll fix that

src/engraving/libmscore/measure.cpp
```diff
@@ -4257,7 +4257,12 @@ void Measure::setWidthToTargetValue(Segment* s, qreal x, bool isSystemHeader, Fr
     // The input parameters of this method rely on Measure::computeWidth(), so always call this method from there
     const double epsilon = 0.1 * spatium();
     static constexpr double multiplier = 1.4; // Empirical value for fastest convergence of the following loop
-    static constexpr int maxIter = 200; // Just for safety, in reality just 2 iterations are typically needed.
+    static constexpr int maxIter = 200;
+    // Different measures need different numbers of iterations of the following loop to reach the target width.
+    // The average is about 2 iterations, and the maximum I've ever seen (very rare) is around 10-15 iterations.
+    // maxIter just serves as a safety exit to not get stuck in the loop in case a measure can't be justified
+    // (which can only happen if errors are made before getting here). It's set to a very high value to make
+    // sure that the measure really can't be justified, and it isn't just a "tricky" one needing more iterations.
```
Jojo-Schmitz, line 4265: just like here ;-)

src/engraving/libmscore/measure.cpp
RomanPudashkin: In any case, it is unclear why you used two different values for maxIter and why they are so big. Especially, considering this observation:

>I haven't seen so far any system taking more than 30-40 iterations (and only in very rare cases)
 
So, they should both be reduced to 40 (or even less, if this case is unrealistic). Ideally, we should use the same constant in both places

RomanPudashkin: 400 and 200 are too big numbers. We can have two different constants, but their values should definitely be reduced to something more realistic

Jojo-Schmitz: I agree. Maybe 64. Would cover all known cases and still have enough headroom.

Jojo-Schmitz: Or 42, the answer to all questions ;-)

mike-spa: Ok, I've gone back and tried to test as many edge cases as I could think of. For system justification I've seen a maximum of 50 iterations needed, for measure justification I've seen a maximum of 20. So `maxIter` is now set to 100 for systems, and 50 for measures. It really shouldn't be less than that: you _never_ design a system to fail on cases that may be just slightly harder than the hardest ones you tested. Also because `maxIter` is almost irrelevant in practice: if a system converges to the target after just 3 iterations, it will exit the loop at 3 iterations, regardless of `maxIter` being 100 or 1000.

RomanPudashkin (approved):

PR #10518, opened 2022-02-10
Fix #10440: Re-add Remove measures to Context menu

### LINKED ISSUE #10440: Context menu no longer has 'Remove measures' option. We should re-add it with some UX cleanup.
**Describe the bug**
Removing measures is still quite hard in MS4. One method from MS3 was to right click on selected measures and opt to 'remove' them. This was always a little tough to find because there is another option called **'Delete'** which seems much more likely to fulfil that function.

I'm guessing someone assumed these two functions were duplicating functionality and removed one of them.

image

**Expected behavior**
We need to do two things:

1. We need to add a 'Delete measures' option (and still use the delete icon) - this should be tied to the existing shortcut: Ctrl (command) +backspace
2. Under that we put the existing 'Delete' feature but rename it to 'Clear measures' (and still use the delete symbol)

**Additional context**

duanjr:
Resolves: [#10440](https://github.com/musescore/MuseScore/issues/10440)

In order not to interfere with other menus，I only rewrite context menu when select measure(s).
I also add Icon to "time-delete", for it's only used here(except shortcut), I think it's ok to add that.

https://user-images.githubusercontent.com/71218187/153349899-d57d84c1-9a00-486c-8e81-2ccbe79d2f46.mp4

### DIFF
--- a/src/notation/view/notationcontextmenumodel.cpp
+++ b/src/notation/view/notationcontextmenumodel.cpp
@@ -21,6 +21,8 @@
  */
 #include "notationcontextmenumodel.h"
 
+#include "translation.h"
+
 #include "ui/view/iconcodes.h"
 
 using namespace mu::notation;
@@ -90,9 +92,9 @@ MenuItemList NotationContextMenuModel::makeMeasureItems()
     items << makeSeparator();
 
     MenuItem* clearItem = makeMenuItem("notation-delete");
-    clearItem->setTitle("Clear measures");
+    clearItem->setTitle(qtrc("notation", "Clear measures"));
     MenuItem* deleteItem = makeMenuItem("time-delete");
-    deleteItem->setTitle("Delete measures");
+    deleteItem->setTitle(qtrc("notation", "Delete measures"));
     items << clearItem;
     items << deleteItem;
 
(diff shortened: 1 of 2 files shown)


src/notation/view/notationcontextmenumodel.cpp
```diff
@@ -80,7 +80,22 @@ MenuItemList NotationContextMenuModel::makeDefaultCopyPasteItems()
...
+        makeMenuItem("notation-cut"),
+        makeMenuItem("notation-copy"),
+        makeMenuItem("notation-paste"),
+        makeMenuItem("notation-swap"),
+    };
+
+    items << makeSeparator();
+
+    MenuItem* clearItem = makeMenuItem("notation-delete");
+    clearItem->setTitle("Clear measures");
```
RomanPudashkin, line 93: You missed translations here. Should be:

clearItem->setTitle(qtrc("notation", "Clear measures"))

The same for deleteItem

duanjr: Thanks for pointing it out. I already corrected it.

RomanPudashkin (approved):

PR #10534, opened 2022-02-11
[MU4] Fix #321074: Pitch bend sounds incorrectly when export to MIDI

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/321074

(affects MuseScore up to 3.6.2 and the 3.6.2_backend on musescore.com too)

### DIFF
--- a/src/importexport/midi/internal/midiconfiguration.cpp
+++ b/src/importexport/midi/internal/midiconfiguration.cpp
@@ -36,7 +36,7 @@ static const Settings::Key EXPORTRPNS_KEY("iex_midi", "io/midi/exportRPNs");
 void MidiConfiguration::init()
 {
     settings()->setDefaultValue(SHORTEST_NOTE_KEY, Val(mu::engraving::Constants::division / 4));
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Has this been obsoleted by #16062?

cbjeukendrup:
@Jojo-Schmitz I don't think so. https://github.com/musescore/MuseScore/pull/16062 seems to fix something about multiple pitch bends writing to the same channel (?), while this PR is still necessary to specify the correct range for pitch bend events.

cbjeukendrup (approved):

cbjeukendrup:
Exporting these RPNs by default seems to me indeed the right thing to do, since they are required to make it sound correctly. The only reason that we didn't do it is apparently https://musescore.org/en/node/37431, namely that some software years ago couldn't handle it. Still I think it's fine to enable it by default, as the user can always disable it if necessary.

cbjeukendrup:
@Jojo-Schmitz Could you please rebase this to make sure that it doesn't conflict with the current state of the unit tests? (I guess it doesn't, because those unit tests haven't really been touched, but you never know)

PR #10536, opened 2022-02-11
[MU4] Optimized drawing of a large score

Eism:
Generally, it is reverting a part of PR #6274 because it caused lags when, for example, a note was shifted.
For now, it's easier to revert these changes than to try to optimize the current solution.

Score on which the bug can be reproduced
[smell.zip](https://github.com/musescore/MuseScore/files/8047960/smell.zip)


vpereverzev (approved):

Jojo-Schmitz:
What went wrong with the vtetsts? The diff is **huge**, but doesn't show any **real** diff.

PR #10549, opened 2022-02-11
Fix MS#318983 Custom Keysig - different accidentals per octaves

sammik:
Resolves: *([Musescore #318983](https://musescore.org/en/node/318983))*

- if there are more accidentals for same note, every affects only single octave, other octaves remain in unchanged (natural) state

https://github.com/musescore/MuseScore/assets/1646034/ecf7d110-7098-4a9e-b63e-ce983478e943

Wery useful for scordatura scores. Quite often used there (f.ex. Biber: Mysteries)


sammik:
@oktophonie what do You think?

sammik:
Vtest failure asexpected.

before (first accidental is ignored)
custom-keysig-5-1 ref

after (added ability to have different accidentals per octave, as described above)
custom-keysig-5-1

mike-spa:
Sorry to be the killjoy this time, but we're not going to merge this. 
- The feature is too unclear as it stands: does each accidental from the key signature apply only to its own specific octave? Does one of them apply to all the other octaves? If so, which of the two? I appreciate the use case, but it's quite clear to me that we have to design a proper feature for this, we can't just make a "fix".
- Before implementing this kind of hyper specialized application we probably need to take a fresh look at custom key signatures in general.
- As far as I can tell, not having this feature isn't a blocker, it just requires (annoying, I'm sure) manual adjustments. 
- I find the code very obscure, as I struggle to understand what it's doing even though I understand the intended outcome.

@oktophonie

sammik:
@mike-spa No problem :)

1. feature
a) Does each accidental apply only to its own specific octave? - YES
b) Does one of them apply to all the other octaves? - NO If so, which of the two? - The one, user added first. But as it is difficualt (or impossible) to distinguish that, I choosed "NO" here, so "a)"
2. we need to take a fresh look at custom key signatures in general. Yes, I agree. But when You will doing that, please, try to preserve as much flexibility, as possible. Now, Custom key signatures in MuseScore are really customisable. None of other notation apps I know have this level of flexibility.
3. this feature isn't a blocker - Well, yes, but... Broken Key signature cannot be blocker by definition, as KS just makes some accidentals to disappear in measure by default, and user can always hide them by hand. In this point of view also non-working convential KS is not blocker, but I am not sure, if it would be marked so. But yes, there are not many people on earth, who want this concrete special feature.
4. I find the code very obscure - Yes, I agree, it may seem so :) I should have commented some places. I will make some "review" comments no, just in case, You will work on Custom KS later, so may be, You can find it useful.

Issue #10460, opened 2022-02-06
[MU4 Issue] Many UI elements don't respond to keyboard input


MarcSabatella:
I am finding that very few UI elements that traditionally respond to keyboard input work as expected in MU4 - spinboxes, dropdown lists, defaults button in dialogs, etc.  So for example

Spinbox issue
1) go to Edit / Preferences
2) click within the "Auto save every" spinbox
3) press Up arrow to increase
Result: spinbox loses focus, no change is applied

Dropdown issue
1) select any text element in a score
2) click in the Font dropdown
3) type either Z or similar to move to the end of the list of fonts
Result so far: dropdown repositions to show fonts starting with that letter, as expect
4) use cursor to select one of the fonts shown
Result: dropdown closes, no change is applied

Note that if you type a letter A-G, instead of repositioning the dropdown, a note is entered into the score.

Default button issue:
1) File / Open
2) click any score in the list
3) press Enter to open the select score
Result: nothing happens

Checkbox issue:
1) Edit / Preferences
2) click the checkbox for "Auto save every" to disable it
3) press Space to toggle it back
Result: nothing happens

and so on.

I find I can operate at least some controls by keyboard *if* I don't touch the mouse at all during the process.  Basically, if the element receives focus via the keyboard, then the keyboard often works to activate it.  So the good news is, blind users who literally use the keyboard only may not see many problems here (although I can't get spinboxes to work no matter what).  But for most of us who use some combination of clicking and keyboard, it's not good at all.

MarcSabatella:
Is there any way this could be bumped in priority to be dealt with before alpha?  In more regular use of MU4 over the past few weeks, this is by far the single biggest usability issue I encounter, and a much bigger regression over MU3 than any single missing feature.

Eism:
@MarcSabatella 
I think it will be fixed in #9839
And I'll start fixing it soon

MarcSabatella:
Fantastic, thanks!

Eism:
@MarcSabatella 

>  Default button issue:
> 
> File / Open
> click any score in the list
> press Enter to open the select score
> Result: nothing happens

Where is "File / Open"?

MarcSabatella:
I mean the File menu, then Open.  It’s really the same for most controls throughout the UI - keyboard just doesn’t work.

Eism:
But File->Open launches the system's standard file open dialog, we can't change the behaviour of this dialog.
What platform do you have this behaviour on?

[...]

PR #10566, opened 2022-02-14
Fix #9769: rewrite-text-size-menu

### LINKED ISSUE #9769: [MU4 Issue] Text resize menu should be a text box + context menu
**Describe the bug**
The current design pattern for resizing text is a dropdown menu, which limits users to a finite set of pre-determined text size values.

Screenshot 2021-11-18 at 12 59 46 am

**Expected behavior**
Instead, this should be an editable text box + context menu containing preset sizes.
This context menu should **replace** the current dropdown control, as it will function better for this kind of setting (i.e. we don't want to use the dropdown anymore). 

Screenshot 2021-11-18 at 12 57 10 am

Thanks!

duanjr:
Resolves: [#9769](https://github.com/musescore/MuseScore/issues/9769)

 make size menu be an editable text box + context menu containing preset sizes.

https://user-images.githubusercontent.com/71218187/153905265-ddba03fb-7049-4266-be0e-6d52e9f5d11a.mp4

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/Dropdown.qml
@@ -44,6 +44,8 @@ Item {
 
     property int popupWidth: root.width
     property int popupItemsCount: 18
+    property alias popupX: popup.x
+    property alias popupY: popup.y
 
     property alias dropIcon: dropIconItem
     property alias label: mainItem.label
--- a/src/inspector/view/qml/MuseScore/Inspector/text/TextInspectorView.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/text/TextInspectorView.qml
@@ -116,7 +116,7 @@ InspectorSectionView {
                 }
             }
 
-            DropdownPropertyView {
+            InspectorPropertyView {
                 id: sizeSection
                 anchors.left: parent.horizontalCenter
                 anchors.leftMargin: 2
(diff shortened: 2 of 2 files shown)


bkunda:
Fantastic that this has been done. Thanks @duanjr!

One issue I have found after testing this build on MacOS is that pressing `Return` after changing the font size value seems to have an undesirable effect on horizontal score spacing:

https://user-images.githubusercontent.com/86290556/165516592-8c9dd092-1d62-4f40-9fae-a9865bfa4e31.mov

Using `Return` to "confirm" the change of value seems to me to be a pretty predictable user behaviour, so this horizontal spacing issue would need to be addressed before this could be merged.

Also I would request that there be no rounded corners between the text field and the dropdown menu button (circled):

Screenshot 2022-04-27 at 10 12 05 pm

There was some discussion in our team about potential issues with introducing new components, although I'm not sure this fix actually does that. @Eism could you please check this and comment? Obviously if it's going to create some potential headaches for us then I'd happily come up with an ever simpler solution 🙂

bkunda:
> I think what you're seeing here is not specifically related to the change. The problem was already present in the alpha builds but has been mostly fixed since: the score retains keyboard focus even after clicking controls in other windows. So hitting Enter or using the Up/Down within the Properties window indeed edits the score if you have something selected. In this case, it added a line break. I'm guessing this branch is just missing whatever code was added to fix this during the past couple of weeks. I've been using a build that incorporates a partial fix but I don't know if all the relevant code has been merged yet.
> 
> See for example #10460 and #11193.

Ok it looks like work on this issue is already underway then, which is great. 
@Eism my remaining issue with this PR is the simple UI fix (noted above).
Other than that, it's over to you to assess whether this component is all good to introduce at this stage 🙂

bkunda:
> > Ok it looks like work on this issue is already underway then, which is great. @Eism my remaining issue with this PR is the simple UI fix (noted above). Other than that, it's over to you to assess whether this component is all good to introduce at this stage 🙂
> 
> I have one more question about the UI fix: we have a border for the inputfield in this branch.The border will be highlight when input. If we remove the rounded corners then what is the border's behavior?

I would think that the border would still be highlighted on hover, like our other inputfield elements. 
To be clear, it's really just the top and bottom right corners of the inputfield border that I don't think should be rounded. 
We are now getting into fine UI territory here, so am quickly tagging in @jessjwilliamson for her input too.

shoogle:
Hi @duanjr. Please sign the CLA over at https://musescore.org/cla. Don't just tick the box here on GitHub. Thanks!

Tantacrul:
What's the status of this?

@bkunda - you happy?
@duanjr - have you signed the CLA as per @shoogle's request? Also, can you resolve the conflicts? We're dying to merge this badboy!

duanjr:
> What's the status of this?
> 
> @bkunda - you happy? @duanjr - have you signed the CLA as per @shoogle's request? Also, can you resolve the conflicts? We're dying to merge this badboy!

Yes I signed it. 
As about the issue itself, maybe solve the ui issue before trying to merge it?
I successfully remove the radius in two corners of the inputfield and adjust the border's shape, but I was stuck at removing those of the dropdon Icon. The problem is like this:
Qml actually does not allow set different radius for each corner, either different border for each side. So we need to take some alternative measures. I decided to draw a rectangle to hide the corner and draw another to hide the left boder of the first one:
```
Rectangle {
      id: hideRadiusRectangle
      width: textInputField.background.radius
      height: textInputField.background.height
      anchors {
          bottom: textInputField.background.bottom
          right: textInputField.background.right
      }
      color: textInputField.background.color
      border.color: textInputField.background.border.color
      border.width: hideRadiusRectangle.background.border.width
      Rectangle {
          anchors.right: hideRadiusRectangle.left
          anchors.top: hideRadiusRectangle.top
          color: hideRadiusRectangle.color
          height: textInputField.background.height - hideRadiusRectangle.border.width * 2
          width: hideRadiusRectangle.border.width
          anchors.topMargin: hideRadiusRectangle.border.width
          anchors.rightMargin: -hideRadiusRectangle.border.width
      }
  }
```
But we cannot do similar work about the dropdown icon, since the `opacity` of its background is not 1, so when we draw another rectangle with the same color. The overlapping part's color will be darker.

image

I made some other attempts under the above ideas but all failed. I think one possible solution is to reduce the dropdown icon part's width , set its radius to 0, and draw a column (of 1 quarter circle, 1 ranctangle , 1 quarter circle, we can use QtQuick.Shapes to draw a quarter circle) to the right side to achieve the expected apperance . In that case, those semi-opaque color blocks will not overlap. But this solution may be too complex for such a small ui. I wonder if we could have a better solution.

Tantacrul:
I don't really care about the visuals right now. I think those are 'good enough' and someone will tweak it later.

I do still see a few significant functional issues though:

1. When entering a number, the text updates immediately. This goes against the general convention that the user has to either click away (deselect) the text box in order to confirm or just press 'Return'
2. When I press the 'Return' while the text box is in focus, it creates a system break on my score, which is super janky
3. The text box often fails to reflect the size of the text and regularly goes blank or gives the wrong measurement

https://user-images.githubusercontent.com/21022311/167290822-d2c81d24-15f2-4585-a2b6-f395f56ab85c.mov

@Eism or @RomanPudashkin - would it be better that we do this ourselves?

PR #10569, opened 2022-02-15
Fix instrument IDs for harmonicas and winds

shoogle:
Fixes historic problems with instrument IDs in instruments.xml:

- The IDs for the 10 hole harmonicas mistakenly included the number 12 instead of the number 10.

- The ID for the "Winds" instrument on grand staff was capitalised when all other IDs are lowercase.

Also fixes the ID and description of the "10 Hole G Diatonic Harmonica". The description said "high G" and the ID included an unnecessary dash.

Changes to instrument IDs affects compatibility with older scores. This has been accounted for in PR #10570, which replaces old IDs with new ones when a score is loaded. See commit 53b7b0685d and the changes it introduced to file [src/project/internal/projectmigrator.cpp](https://github.com/musescore/MuseScore/pull/10570/files#diff-300df7f140304e50db06105e080af6e9f60bd9999fccad7ef7834aa141774b70R134-R149).


share/instruments/instruments.xml
```diff
@@ -3600,13 +3600,13 @@
                         <program value="22"/> <!--Harmonica-->
                   </Channel>
             </Instrument>
-            <Instrument id="harmonica-d12-g">
+            <Instrument id="harmonica-d10g">
```
shoogle, line 3603: @oktophonie, I took the opportunity to remove the dash before the "g" in this ID to make it consistent with the others.

share/instruments/instruments.xml
```diff
@@ -3600,13 +3600,13 @@
...
                   <family>harmonicas</family>
                   <trackName>10 Hole Diatonic Harmonica</trackName>
                   <longName>10 Hole G Diatonic Harmonica</longName>
                   <shortName>Harm.</shortName>
                   <dropdownName meaning="tuning">G</dropdownName>
-                  <description>10-hole diatonic harmonica pitched in high G.</description>
+                  <description>10-hole diatonic harmonica pitched in G.</description>
```
shoogle, line 3609: @oktophonie, I think this description was wrong so I changed it. There is another 10 hole diatonic harmonica that really is the "high G" variant.

shoogle:
Merging unblocked. The necessary compatibility code to handle opening of older scores has now been added by PR #10570 in [src/project/internal/projectmigrator.cpp](https://github.com/musescore/MuseScore/pull/10570/files#diff-300df7f140304e50db06105e080af6e9f60bd9999fccad7ef7834aa141774b70R134-R149).

PR #10570, opened 2022-02-15
Fix #10499

### LINKED ISSUE #10499: [MU4 Task] Implement score migration routine for scores from previous versions being opened in MU4.0
Implement score migration routine for **all** scores from previous versions being opened in MU4.0.

This list of changes is non-exhaustive but is quite enough to start with :)

- [ ] Ensure `beamDistance` is mapped to `useWideBeams` (`true` if `beamDistance` is nearer 1.00 than 0.75, otherwise `false`, which is the default; see [this commit](https://github.com/musescore/MuseScore/commit/f7a2a5a97d027be4d4ecb244672225398d481f52))
- [ ] Ensure that custom beam positions are retained (this is possible as they are expressed as absolute values, not as offsets)
- [ ] Ensure that explicitly-set beam direction is retained
- [ ] Reset all manual adjustments to slurs and ties, except for explicitly-set direction
- [ ] Reset layout stretch throughout the score
- [ ] Ensure that any and all `showSystemMarkings` attributes inherited from orders.xml are excised (they were never used, and their continued presence in files will only cause confusion)
- [ ] Ensure that the `<SystemObjects>` element is always empty
- [ ] Reinterpret the style settings for `endBarDistance` and `doubleBarDistance` so that they will give the same visual result as before, according to the new calculation (see [#9138](https://github.com/musescore/MuseScore/pull/9138))
- [ ] Set `dynamicsFontSize` to 10pt in every case (normally we honour settings in the file/previous defaults, but this is a rare exception)

## TBC / for discussion
- Do/can we retain custom stem lengths?

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/10499

A variety of fixes to score migration! The checklist in the linked issue has been fully addressed, except for the one about maintaining bar adjustments, as that will be fixed by [#10130 ](https://github.com/musescore/MuseScore/pull/10130).


oktophonie (changes requested):

oktophonie (approved):

Jojo-Schmitz:
Actually it dies of a failed assertion:
ASSERT: "MStyle::valueType(idx) == mu::engraving::P_TYPE::REAL" in file ../src/engraving/style/style.h, line 64

src/project/internal/projectmigrator.cpp
```diff
@@ -66,12 +67,29 @@ Ret ProjectMigrator::migrateEngravingProjectIfNeed(engraving::EngravingProjectPt
...
+    doubleBarDistance -= style->styleMM(Ms::Sid::doubleBarWidth);
+    style->set(Ms::Sid::doubleBarDistance, doubleBarDistance / sp);
+    qreal endBarDistance = style->styleMM(Ms::Sid::endBarDistance);
+    endBarDistance -= (style->styleMM(Ms::Sid::barWidth) + style->styleMM(Ms::Sid::endBarWidth)) / 2;
+    style->set(Ms::Sid::endBarDistance, endBarDistance / sp);
+    qreal repeatBarlineDotSeparation = style->styleMM(Ms::Sid::repeatBarlineDotSeparation);
+    qreal dotWidth = project->masterScore()->scoreFont()->width(Ms::SymId::repeatDot, 1.0);
+    repeatBarlineDotSeparation -= (style->styleMM(Ms::Sid::endBarWidth) + dotWidth) / 2;
+    style->set(Ms::Sid::repeatBarlineDotSeparation, repeatBarlineDotSeparation / sp);
+    project->masterScore()->setResetDefaults();
```
RomanPudashkin, line 91: This whole new block of code doesn't look like it belongs to ProjectMigrator. Perhaps it should be moved to some public method of MasterScore

asattely: I agree, though my main fear about that is that this only should happen once, on first open of an old score. I still have some work to do on this one so I'll keep searching for a better location for it

asattely:
These vtest issues are weird because they are not happening on my machine, so I suspect it's something about build configurations. I'm going to continue to investigate this today

src/project/internal/projectmigrator.cpp
```diff
@@ -146,11 +147,54 @@ Ret ProjectMigrator::migrateProject(engraving::EngravingProjectPtr project, cons
     if (ok && opt.isApplyAutoSpacing) {
         ok = resetAllElementsPositions(score);
     }
-
+    if (score->mscVersion() <= 302) {
+        for (Part* part : score->parts()) {
```
RomanPudashkin, line 151: Please move this code to a separate private method

src/project/internal/projectmigrator.cpp
```diff
@@ -146,11 +147,54 @@ Ret ProjectMigrator::migrateProject(engraving::EngravingProjectPtr project, cons
...
+    }
     if (ok && score->mscVersion() != Ms::MSCVERSION) {
         score->undo(new Ms::ChangeMetaText(score, "mscVersion", MSC_VERSION));
     }
 
+    if (_resetStyleSettings) {
+        // there are a few things that need to be updated no matter which version the score is from (#10499)
+        // primarily, the differences made concerning barline thickness and distance
+        // these updates take place no matter whether or not the other migration options are checked
+        qreal sp = score->spatium();
```
RomanPudashkin, line 182: Please move this code to a separate private method

oktophonie (approved):

PR #10573, opened 2022-02-15
[MU4] Various fixes for the workspaces dialog

RomanPudashkin:


src/workspace/view/newworkspacemodel.cpp
```diff
@@ -88,8 +88,24 @@ void NewWorkspaceModel::setWorkspaceName(const QString& name)
...
+    validateWorkspaceName();
+    emit workspaceNameChanged();
+}
+
+void NewWorkspaceModel::validateWorkspaceName()
+{
+    m_errorMessage.clear();
+
+    if (m_allWorkspaceNames.contains(m_workspaceName)) {
+        m_errorMessage = QString("'%1' does already exist, please, choose a different name").arg(m_workspaceName);
```
cbjeukendrup, line 100: Shouldn't this be translatable? And how about a wording like this:
> A workspace with the name '%1' already exists. Please choose a different name.

src/workspace/qml/MuseScore/Workspace/NewWorkspaceDialog.qml
```diff
@@ -93,6 +93,14 @@ StyledDialogView {
             }
         }
 
+        StyledTextLabel {
+            Layout.topMargin: 12
+            Layout.fillWidth: true
+
+            horizontalAlignment: Qt.AlignLeft
+            text: workspaceModel.errorMessage
```
cbjeukendrup, line 101: Shouldn't we say explicitly `visible: !isEmpty`? Or did you omit that on purpose?

RomanPudashkin: Usually, this is not a good idea, because the content of the dialog will bounce when this label appears

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

PR #10577, opened 2022-02-15
"Last system fill threshold" should also apply to section breaks

mike-spa:
This is a tiny fix.
The setting "Last system fill threshold" (from Style -> Page), by which we don't justify the last system if its width is smaller than a given amount, should also apply to section breaks.
For reference, this is what the program does now (and also MU3.6)
Screenshot from 2022-02-07 10-52-09
but this would be the correct layout
Screenshot from 2022-02-07 10-53-49

I don't see any obvious downsides to making this the default behaviour. @oktophonie

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -381,8 +381,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     // proportional to the difference between the current length and the target length.
     // After few iterations, the length will converge to the target length.
     qreal newRest = systemWidth - curSysWidth;
-    if (ctx.curMeasure == 0 && ((curSysWidth / systemWidth) <= score->styleD(Sid::lastSystemFillLimit))) {
-        // We do not stretch last system if curSysWidth is <= lastSystemFillLimit
+    if ((ctx.curMeasure == 0 || (lm && lm->sectionBreak()))
+        && ((curSysWidth / systemWidth) <= score->styleD(Sid::lastSystemFillLimit))) {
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
That request goes a step further though, asking for a section break property whether or not to follow that setting.
And for the last system to use it regardless of trailing vertical- or text frames

Jojo-Schmitz:
I wonder what that crash in the utests is about

Jojo-Schmitz:
(my backport to 3.x crashes too, not sure whether that's good or bad news)

src/engraving/layout/layoutsystem.cpp
```diff
@@ -381,8 +381,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     // proportional to the difference between the current length and the target length.
     // After few iterations, the length will converge to the target length.
     qreal newRest = systemWidth - curSysWidth;
-    if (ctx.curMeasure == 0 && ((curSysWidth / systemWidth) <= score->styleD(Sid::lastSystemFillLimit))) {
-        // We do not stretch last system if curSysWidth is <= lastSystemFillLimit
+    if ((ctx.curMeasure == 0 || system->lastMeasure()->sectionBreak())
```
Jojo-Schmitz, line 384: Could use `lm->sectionBreak()` here. And check whether lm exists first...

Jojo-Schmitz: Actually that check does fix the crash: `if ((ctx.curMeasure == 0 || (lm && lm->lastMeasure()->sectionBreak()))`

mike-spa: yup, good catch! thanks

oktophonie (approved):

PR #10545, opened 2022-02-11
[MU4] Fix #10542: Crash after adding Figured bass

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10542

PR #10584, opened 2022-02-16
[MU4] Fix #329469: Editing Measure numbers causes crash

Jojo-Schmitz:
While at it fix other potential crashes in the same area

Resolves: https://musescore.org/en/node/329469

I'm not sure at all whether this is the right way to fix this. Probably measure numbers should not pretend to be editable in the first place, or need to get made editable.

### DIFF
--- a/src/engraving/libmscore/textbase.cpp
+++ b/src/engraving/libmscore/textbase.cpp
@@ -3066,21 +3066,28 @@ AccessibleItem* TextBase::createAccessible()
 
 void TextBase::notifyAboutTextCursorChanged()
 {
-    accessible()->accessiblePropertyChanged().send(accessibility::IAccessible::Property::TextCursor, Val());
+    if (accessible()) {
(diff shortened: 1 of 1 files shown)


RomanPudashkin:
I guess we need a similar fix:

https://github.com/musescore/MuseScore/pull/10545

If an element does not have accessible(), it means it was not created via the factory

Jojo-Schmitz:
Only here these are not supposed to be editable, in contrast to figured bass?

RomanPudashkin:
It is better to create all elements via the factory :) But I agree that we should have some kind of protection against null pointers here

RomanPudashkin:
Could you please add a null check to notifyAboutTextCursorChanged() ?

RomanPudashkin (approved):

Jojo-Schmitz:
Done

Jojo-Schmitz:
No idea where to plug in the factory though

PR #10586, opened 2022-02-16
[MU4] Fix #329470: Wrong/missing transpositioning on MusicXML import

Jojo-Schmitz:
If the optional element "diatonic" is missing, calculate it from "chromatic".

Resolves: https://musescore.org/en/node/329470
Same issue in 3.x.

Also adds/enables all (most?) available MusicXML unit tests (commenting out those that fail, working on those is out of score of this PR).

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass1.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass1.cpp
@@ -36,6 +36,7 @@
 #include "engraving/libmscore/timesig.h"
 #include "engraving/libmscore/spanner.h"
 #include "engraving/libmscore/bracketItem.h"
+#include "engraving/libmscore/utils.h"
 
 #include "engraving/types/symnames.h"
 #include "engraving/style/style.h"
@@ -2496,6 +2497,9 @@ void MusicXMLParserPass1::transpose(const QString& partId, const Fraction& tick)
     }
 
     if (_parts[partId]._intervals.count(tick) == 0) {
+        if (!interval.diatonic && interval.chromatic) {
+            interval.diatonic = chromatic2diatonic(interval.chromatic);
+        }
         _parts[partId]._intervals[tick] = interval;
     } else {
         qDebug("duplicate transpose at tick %s", qPrintable(tick.toString()));
--- a/src/importexport/musicxml/tests/tst_mxml_io.cpp
+++ b/src/importexport/musicxml/tests/tst_mxml_io.cpp
@@ -93,14 +93,21 @@ private slots:
     void barStyles() { mxmlIoTest("testBarStyles"); }
     void barStyles2() { mxmlIoTest("testBarStyles2"); }
     void barStyles3() { mxmlIoTest("testBarStyles3"); }
+    // void beamEnd() { mxmlIoTest("testBeamEnd"); } fails
+    void beams1() { mxmlIoTest("testBeams1"); }
+    void beams2() { mxmlIoTest("testBeams2"); }
+    void beams3() { mxmlIoTestRef("testBeams3"); }
     void breaksImplExpl() { mxmlMscxExportTestRefBreaks("testBreaksImplExpl"); }
     void breaksMMRest() { mxmlMscxExportTestRefBreaks("testBreaksMMRest"); }
     // void breaksManual() { mxmlIoTestRefBreaks("testBreaksManual"); } fail after sync with 3.x
     // void breaksPage() { mxmlMscxExportTestRefBreaks("testBreaksPage"); } fail after sync with 3.x
-    // void breaksSystem() { mxmlMscxExportTestRefBreaks("testBreaksSystem"); } fail after sync with 3.x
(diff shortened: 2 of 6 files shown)


Jojo-Schmitz:
@lvinken: please review

src/importexport/musicxml/internal/musicxml/importmxmlpass1.cpp
```diff
@@ -2498,6 +2498,12 @@ void MusicXMLParserPass1::transpose(const QString& partId, const Fraction& tick)
     }
 
     if (_parts[partId]._intervals.count(tick) == 0) {
+        if (!interval.chromatic) {
+            qFatal("Mandatory element \"chromatic\" missing");
```
Jojo-Schmitz, line 2502: Better just a `qDebug()`, as 0 might be a legal value.

src/importexport/musicxml/internal/musicxml/importmxmlpass1.cpp
```diff
@@ -2498,6 +2498,12 @@ void MusicXMLParserPass1::transpose(const QString& partId, const Fraction& tick)
...
+            return;
```
Jojo-Schmitz, line 2503: And then of course no exit from this method.

src/importexport/musicxml/internal/musicxml/importmxmlpass1.cpp
```diff
@@ -2498,6 +2498,12 @@ void MusicXMLParserPass1::transpose(const QString& partId, const Fraction& tick)
...
+        } else if (!interval.diatonic) {
+            interval.diatonic = round(interval.chromatic * 7.0 / 12.0);
```
Jojo-Schmitz, line 2505: Better use `cromatic2diatonic(int semitones)` [from utils.cpp, line 596](https://github.com/musescore/MuseScore/blob/28bf1776dd5737d8edc2f55422d97886a0115a1b/src/engraving/libmscore/utils.cpp#L596)

lvinken:
As indeed diatonic is optional but chromatic is required, the code change is correct. I apparently overlooked the possibility of missing diatonic, which led to an invalid interval in my implementation. Thanks for fixing this.

Note that a missing chromatic would be caught by the schema validation.

A testfile would be a nice addition, to prevent this requirement from being missed in future.

Jojo-Schmitz:
Should I drop the check for the mandatory "chromatic", as it is caught elsewhere already?

Jojo-Schmitz:
I see a ``testChangeTranspose.xml`.xml`, which we might use for this, by making a copy (e.g. `testChangeTranspose-no-diatonic.xml`) and removing the optional `diatonic`, but I don't see where that test is actually run, so don't know where and how to add it (or them both)

src/importexport/musicxml/tests/tst_mxml_io.cpp
```diff
@@ -98,6 +98,8 @@ private slots:
     // void breaksManual() { mxmlIoTestRefBreaks("testBreaksManual"); } fail after sync with 3.x
     // void breaksPage() { mxmlMscxExportTestRefBreaks("testBreaksPage"); } fail after sync with 3.x
     // void breaksSystem() { mxmlMscxExportTestRefBreaks("testBreaksSystem"); } fail after sync with 3.x
+    void changeTranspose() { mxmlIoTest("testChangeTranspose"); }
```
Jojo-Schmitz, line 101: This test is in 3.x, but not in master, not sure why?

src/importexport/musicxml/tests/tst_mxml_io.cpp
```diff
@@ -98,6 +98,8 @@ private slots:
...
+    void changeTransposeNoDiatonic() { mxmlIoTestRef("testChangeTranspose-no-diatonic"); }
```
Jojo-Schmitz, line 102: This test however fails. If I import `testChangeTranspose-no-diatonic.xml` and export it again (using the -d option and no formatting export in preferences, as apparently these tests do too) I don't see and differences between that export and `testChangeTranspose-no-diatonic-ref.xml`.

src/importexport/musicxml/tests/tst_mxml_io.cpp
Jojo-Schmitz: I wonder how many more such tests are dormant, @lvinken?

src/importexport/musicxml/tests/tst_mxml_io.cpp
Jojo-Schmitz: Acrually I checked it with `vim`'s diff. The 'real' diff shows the problem: DOS line endings in the ref. file, UNIX line endings in the exported one. Is that maybe the problem?

lvinken: White-space characters have bitten me before when mixing Linux and Windows diff ...

Jojo-Schmitz: AHRGL!!! I'm just too stupid, the file name of the ref file was wrong, `...-ref.xml` rather than `..._ref.xml`!

lvinken:
Looks good to me, thanks.

Issue #10541, opened 2022-02-11
[MU4 Issue] Other than MSCX/MSCZ file formats don't show in Recent scores list (MIDI, MXL, GP etc.)


DmitryArefiev:
**To Reproduce**
1. Open **MIDI** or **GP** or **MXL** or other supported my MU files
2. After closing, they should appear in **Home>Scores** and **File>Open recent**

https://user-images.githubusercontent.com/90187801/153596847-1a103eb3-69b2-432e-bef0-71b550bf8bcc.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

Issue #10343, opened 2022-01-20
[MU4 Issue] MSCX scores are displayed without any logo in Recent scores


DmitryArefiev:
**To Reproduce**
1. Open [01.zip](https://github.com/musescore/MuseScore/files/7907209/01.zip) 
2. Go to Home>Scores

https://user-images.githubusercontent.com/90187801/150388559-42c8652c-b709-417e-b4c0-3f9c3f4666a6.mp4

**Expected behavior**
It should not be empty

**Screenshots**

**Desktop (please complete the following information):**
 - OS:

PR #10600, opened 2022-02-17
Show imported files in Recent Files lists

cbjeukendrup:
Resolves: #10343
Resolves: #10541

We will need a placeholder image to show as the thumbnail for those imported files on the Home page.


jessjwilliamson:
Hey @cbjeukendrup!
Me and Martin discussed to have instead of 1 generic thumbnail, we want to have 3:

1. MusicXML files (.mxl, .musicxml, .xml)
2. MIDI files (.mid, .midi, .kar)
3. Other files (MU logo) 

image
image

I exported PNGs which should be sized to 80 x 110 pixels on top of the white sheet template (similar to how the new score template is)
These apply to all themes.
Figma mockups are in 'IMPORT FILE THUMBNAILS' for references

PNGS:
https://www.dropbox.com/sh/wvu1ta1u72yo29s/AACcH9dTuYOcRVZDPpcRO443a?dl=0

Thank you!

cbjeukendrup:
Done!
Schermafbeelding 2022-05-10 om 19 04 43

jessjwilliamson:
Awesome!!

src/project/view/recentprojectsmodel.cpp
```diff
@@ -34,9 +36,31 @@ using namespace mu::notation;
...
+const QString SCORE_THUMBNAIL_PIXMAP_KEY("thumbnailPixmap");
+const QString SCORE_THUMBNAIL_URL_KEY("thumbnailUrl");
 const QString SCORE_TIME_SINCE_MODIFIED_KEY("timeSinceModified");
 const QString SCORE_ADD_NEW_KEY("isAddNew");
+
+const QString THUMBNAIL_PLACEHOLDER_MIDI_URL("qrc:/resources/Placeholder_MIDI.png");
+const QString THUMBNAIL_PLACEHOLDER_MXML_URL("qrc:/resources/Placeholder_MXML.png");
+const QString THUMBNAIL_PLACEHOLDER_OTHER_URL("qrc:/resources/Placeholder_Other.png");
+
+QString thumbnailPlaceholderUrlBySuffix(const std::string suffix)
```
RomanPudashkin, line 48: This code doesn't look like it belongs to this file. Ideally, RecentProjectsProvider should provide the meta info object from which we can directly get the thumbnail without any additional stuff in this model

RomanPudashkin: Also, this function should be static. And you missed &

src/project/view/scorethumbnail.cpp
```diff
@@ -21,22 +21,22 @@
...
-    if (pixmap.isNull()) {
-        return;
-    }
+    return m_thumbnail;
+}
 
-    m_thumbnail = pixmap.value<QPixmap>();
+void ScoreThumbnail::setThumbnail(QPixmap pixmap)
+{
+    m_thumbnail = pixmap;
```
RomanPudashkin, line 38: I would suggest using std::move here

src/project/view/recentprojectsmodel.cpp
cbjeukendrup: Moved it to QML. (The only extra thing that the model does now, is providing the suffix, so that QML does not need to figure out the suffix from the path.)

RomanPudashkin (approved):

Issue #9106, opened 2021-09-12
[MU4 Issue] Staff/part properties | Advanced style properties | Preview doesn't work


wizofaus:
**Describe the bug**
The preview control in the "Edit staff type" dialog doesn't work (no preview is shown at all).

**To Reproduce**

1. Access "staff/part properties" from the context menu for a staff
2. Click on 'Advanced style properties' to open the 'Edit staff' dialog
3. Note that the preview control is empty

**Expected behavior**
As per v3, preview should show a preview of what the staff will look like

**Screenshots**

![preview](https://user-images.githubusercontent.com/90187801/153830020-fd88a095-46f7-4c2b-880e-86a78d937f51.PNG)

**Desktop (please complete the following information):**
Windows 10

**Additional context**
Doesn't occur in **MU3**

PR #10605, opened 2022-02-17
Fix preview in Staff/Part Properties > Advanced

cbjeukendrup:
Resolves: #9106 

Now, the scaling of the previews in Staff/Part Properties > Advanced and in the Time Sig Properties dialog should be similar to MuseScore 3. This would need to be tested with various kinds of platforms and monitors.

### DIFF
--- a/src/commonscene/exampleview.cpp
+++ b/src/commonscene/exampleview.cpp
@@ -145,6 +145,7 @@ void ExampleView::setScore(Score* s)
 
     ScoreLoad sl;
     _score->doLayout();
+    resetMatrix();
     update();
 }
 
--- a/src/commonscene/exampleview.cpp
+++ b/src/commonscene/exampleview.cpp
@@ -41,32 +41,29 @@ using namespace mu;
 using namespace mu::engraving;
 
 namespace Ms {
-//---------------------------------------------------------
-//   ExampleView
-//---------------------------------------------------------
-
 ExampleView::ExampleView(QWidget* parent)
     : QFrame(parent)
 {
-    _score = 0;
+    m_score = nullptr;
     setAcceptDrops(true);
     setFocusPolicy(Qt::StrongFocus);
     resetMatrix();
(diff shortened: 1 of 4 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #9106 on Win10 (FullHD), Mac12 (Retina), LinuxMint 20.2 (FullHD)

It looks small on Win and Linux with FullHD displays:
Capture

@bkunda Can you check this on your 4k display please?

cbjeukendrup:
It seems there is a broader issue with scaling/zoom at the moment in MU4. For example, when zooming to 100%, the size of the notation does not correspond with the size of a physical sheet of paper. This was the case in MU3, and I also fixed this once for MU4, but apparently it is broken again and I haven't figured out why.

bkunda:
> @bkunda Can you check this on your 4k display please?

@DmitryArefiev It's mostly working on my 4k display. Preview staff renders as expected at my screen's recommended monitor scaling (150%). 

The noteheads do some quirky things when you change the scaling. For instance, setting it down to 100% then returning back to 150% seems to leave the noteheads looking very small and disconnected from their stems. But this is rectified by closing and reopening the dialog.

DmitryArefiev:
> It seems there is a broader issue with scaling/zoom at the moment in MU4. For example, when zooming to 100%, the size of the notation does not correspond with the size of a physical sheet of paper. This was the case in MU3, and I also fixed this once for MU4, but apparently it is broken again and I haven't figured out why.

@cbjeukendrup Maybe do you want a separate issue and merge the current PR?

cbjeukendrup:
@DmitryArefiev That seems a good idea. Let's log a general issue about scaling in notation views (because, as I said, in the main notation view, the scaling corresponding to 100% zoom is also not correct).

DmitryArefiev:
@cbjeukendrup Yes, logged #10686 

Thanks

Issue #10606, opened 2022-02-17
[MU4 Issue] Zooming in one window affects the zoom level reported in the status bar of other windows


cbjeukendrup:
**Describe the bug**
If you zoom using Ctrl+Shift++ or Ctrl+-, the zoom percentage in the status bar is updated, as expected. However, it is updated not only in the current window, but also in other windows, while the actual zoom level is _not_ updated in those other windows.

In this video, check out the status bars of the two windows:

https://user-images.githubusercontent.com/48658420/154559950-81df9760-3a0b-4d99-bad0-2864e12c1546.mov

**Expected behavior**
The zoom level in one window should not interfere with the zoom level in other windows in any way.

**Platform information**
 - OS: reported on macOS, should apply to all

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED in #10607

Issue #9143, opened 2021-09-15
[MU4 Issue] Pinch to zoom does not update zoom level


Nick-Mazuk:
**Describe the bug**

When using pinch to zoom (implemented in https://github.com/musescore/MuseScore/pull/9108), the zoom percentage in the lower right hand corner does not update.

This also means that when you subsequently zoom by any means aside from pinching, the zoom percentage updates to the incorrect value. The only way to reset to the correct zoom value is by choosing a zoom level.

**To Reproduce**
1. Open up any document
2. Pinch to zoom in our out
3. Look at the number in the bottom right

**Expected behavior**

The number should update according to the actual zoom level.

**Screenshots**

https://user-images.githubusercontent.com/5659171/133516834-d544f974-744c-49be-87d3-8b8905708ec3.mov

**Desktop (please complete the following information):**
 - OS: macOS 11.5.2

cbjeukendrup:
I mentioned this in the description of #9108, but thank you for reporting it as an issue, this way it won't get forgotten.
I'm planning to fix this when tackling #7879, so it would be great if somebody could assign me to this issue.

Nick-Mazuk:
> I mentioned this in the description of #9108, but thank you for reporting it as an issue, this way it won't get forgotten.

Sorry, I totally missed that you mentioned this issue in the description! In general, this feature is working _really_ well. I've already used it quite a bit today!

cbjeukendrup:
No problem :)
And yes, I'm also using this feature all the time, now that it is implemented! Actually, in MU3 it worked too, so it's extra important that it works in MU4 too. MU4 seems to have slightly less lag than MU3, but the way.

DmitryArefiev:
The same with Ctr/Cmd+mouse scroll

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED in #10607

PR #10607, opened 2022-02-17
Fix updating zoom level in status bar on pinch-to-zoom or Ctrl + scroll

cbjeukendrup:
Resolves: #9143 
Resolves: #10606

### DIFF
--- a/src/notation/view/notationpaintview.h
+++ b/src/notation/view/notationpaintview.h
@@ -100,7 +100,8 @@ public:
     qreal currentScaling() const override;
     void setScaling(qreal scaling, const PointF& pos) override;
     void scale(qreal factor, const PointF& pos);
-    Q_INVOKABLE void scale(qreal factor, const QPointF& pos);
+
+    Q_INVOKABLE void pinchToZoom(qreal scaleFactor, const QPointF& pos);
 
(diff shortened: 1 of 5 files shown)


src/notation/view/notationpaintview.h
```diff
@@ -100,7 +100,8 @@ class NotationPaintView : public QQuickPaintedItem, public IControlledView, publ
     qreal currentScaling() const override;
     void setScaling(qreal scaling, const PointF& pos) override;
     void scale(qreal factor, const PointF& pos);
-    Q_INVOKABLE void scale(qreal factor, const QPointF& pos);
+
+    Q_INVOKABLE void pinchToZoom(qreal scaleFactor, const QPointF& pos);
```
RomanPudashkin, line 104: The name of the method is not perfect. At first glance, it seems like it converts something into something else. Also, the name of the argument should be the same as for NotationPaintView::scale (you can use scaleFactor in both places)

RomanPudashkin (approved):

src/notation/view/notationpaintview.h
RomanPudashkin: >Would "onPinchToZoomGesture" be better

No, the current one is better :)

PR #10516, opened 2022-02-09
fix #10497 Custom Key Sig, accidentals placement in scaled line distance

sammik:
Resolves: *(#10497)*

*(If line distance was different than 1, accidentals in custom key signatures were placed in wrong vertical position. Fixed.)*

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

fix_scaling_bug

Issue #10497, opened 2022-02-08
[MU4 Issue] Custom Key Signatures scaling bug


sammik:
**Describe the bug**
When Line DIstance is changed, Accidentals in custom key signature douesnt adapt it, are misplaced.

**To Reproduce**
1. Create Custom Key Signature
2. Add it to score
3. Change Staff Properties - Line Distance
4. Accidentala are misplaced

**Expected behavior**
Accudentals should adapt their positions.

**Screenshots**
![Key_Signature_Accidentals_Scaling_Bug](https://user-images.githubusercontent.com/1646034/153034173-95dd443d-aaa9-4270-afc3-f54b7d122f08.png)

**Platform information**
 - OS: Linux (but probably exist on all platforms)

**Additional context**
If Key signature is added before, or after scaling doesnt play role.

zipped mscz file attached
[Key Signature Accidentals Scaling Bug.zip](https://github.com/musescore/MuseScore/files/8025777/Key.Signature.Accidentals.Scaling.Bug.zip)

Jojo-Schmitz:
Probably affects MuseScore 3 too, right?

DmitryArefiev:
Tested in master build (Win10, Mac12, LinuxMint 20.2) - FIXED in #11280

PR #10609, opened 2022-02-18
[MU4] Fix #10517 KeySig refactor

### LINKED ISSUE #10517: [MU4 Task] KeySym refactor
**Task description**
Remove `KeySym` members  `pos` and `spos`, and instead of them, add
```
int line;
double xPos; // in staff spatium units
```
Why?
`pos` and `spos` are "visual" members (spos.y depends on line distance), so same Key Signature had different Key Symbols spos values in different staff settings. 
In custom key signatures, spos was (incorrectly) used as "logical" member instead, which caused visual bug #10497 in staffs with altered line distance.

`line` and `xPos` are "logical" members, which clean up code logic and brings more flexibility to Key Signatures.
(For example it opens up opportunities to adapt custom key signatures per clef, or to transpose them in future)

sammik:
Resolves: #10517 
Also Resolves: #10497, Closes: #10516, Closes: #10543 

Remove `KeySym` members `pos` and `spos`, and instead of them, add
```
int line;
double xPos; // in staff spatium units
```
### Why?
`pos` and `spos` are "visual" members (spos.y depends on line distance), so same Key Signature had different Key Symbols spos values in different staff settings.
In custom key signatures, spos was (incorrectly) used as "logical" member instead, which caused visual bug https://github.com/musescore/MuseScore/issues/10497 in staffs with altered line distance.

`line` and `xPos` are "logical" members, which clean up code logic and brings more flexibility to Key Signatures.
(For example it opens up opportunities to adapt custom key signatures per clef, or to transpose them in future)

### Save / Import Export
KeySym is saved into file like this:
```
<KeySym>
  <sym>accidentalSharp</sym>
  <pos xPos="0" line="0"/>
  </KeySym>

```
Before this change, it was `<pos x="0" y="0"/>`. Code is compatible with old files, they are opened correctly.

Import and export to musicxml works too.

--
I used `qreal stepDistance = staff() ? staff()->staffTypeForElement(this)->lineDistance().val() * 0.5 : 0.5;` in few places, so I also changed `qreal stepDistance = staff() ? staff()->lineDistance(tick()) * 0.5 : 0.5;` in one place to it, in case of consistency.

--
tested on Linux (Ubuntu Studio 21.10)

### DIFF
--- a/src/engraving/libmscore/key.h
+++ b/src/engraving/libmscore/key.h
@@ -77,8 +77,8 @@ static inline Key operator-=(Key& a, const Key& b) { return a = Key(static_cast<
 
 struct KeySym {
     SymId sym;
-    mu::PointF spos;       // position in spatium units
-    mu::PointF pos;        // actual pixel position on screen (set by layout)
+    int line;       // relative line position (first staffline: line == 0, first gap: line == 1, ...)
+    double xPos;    // x position in staff spatium units
 };
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/keysig.cpp
+++ b/src/engraving/libmscore/keysig.cpp
@@ -93,11 +93,11 @@ qreal KeySig::mag() const
 
 void KeySig::addLayout(SymId sym, int line)
 {
-    qreal stepDistance = staff() ? staff()->lineDistance(tick()) * 0.5 : 0.5;
+    qreal _spatium = spatium();
+    qreal step = _spatium * (staff() ? staff()->staffTypeForElement(this)->lineDistance().val() * 0.5 : 0.5);
     KeySym ks;
     ks.sym = sym;
     qreal x = 0.0;
-    qreal y = qreal(line) * stepDistance;
     if (_sig.keySymbols().size() > 0) {
         KeySym& previous = _sig.keySymbols().back();
         qreal accidentalGap = score()->styleS(Sid::keysigAccidentalDistance).val();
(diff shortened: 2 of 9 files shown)


sammik:
Failing visual test is actually good, as this also fixes custom key sig visual bug.

Jojo-Schmitz:
so a `vtests` label is in order I guess

sammik:
> so a `vtests` label is in order I guess

You mean, I shoul tick "I created the test" option too?

sammik:
> No, someone with sufficient priviledge should add the `vtests` label

Ah, OK, thanks for explanation.

cbjeukendrup (commented):
Looks like you've done a good job! Just a few minor comments.

src/engraving/libmscore/key.h
```diff
@@ -77,8 +77,8 @@ static inline Key operator-=(Key& a, const Key& b) { return a = Key(static_cast<
 
 struct KeySym {
     SymId sym;
-    mu::PointF spos;       // position in spatium units
-    mu::PointF pos;        // actual pixel position on screen (set by layout)
+    int line;       //relative line position (first staffline: line == 0, first gap: line == 1, ...)
```
Jojo-Schmitz, line 80: I'd rather see a space between the `//` and the comment's text

src/engraving/libmscore/keysig.cpp
```diff
@@ -437,7 +445,10 @@ void KeySig::read(XmlReader& e)
                     }
                     ks.sym = id;
                 } else if (t == "pos") {
-                    ks.spos = e.readPoint();
+                    int idx = _sig.keySymbols().size();
+                    ks.xPos = e.doubleAttribute("xPos", e.doubleAttribute("x", idx));  //if xPos not there, use x, or accidetal index
+                    ks.line = e.intAttribute("line", static_cast<int>(e.doubleAttribute("y", 2) * 2)); //if line not there, use y, or middle line
```
Jojo-Schmitz, line 450: here too

cbjeukendrup: And maybe put these comments above the line they belong to, because this way the lines are too long.

sammik:
Should be done. Thank You both.

cbjeukendrup (approved):
Looks good to me, thanks!

cbjeukendrup:
Looks like a good optimization :)
I still think `_spatium = spatium()` makes sense, because the `_spatium` variable is used more than once. By creating a variable, the spatium only needs to be calculated/retrieved once.

sammik:
Replaced by #11280

PR #10613, opened 2022-02-18
Update style defaults

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/10395
Some style defaults changes as well as updated some .mss files.

PR #10616, opened 2022-02-19
Cleaned up implementation of MMRest and MeasureRepeat

cbjeukendrup:
Follow-up for #9886 

Appearance on various kinds of staves:
Schermafbeelding 2022-02-19 om 00 35 20

Inspector:
Schermafbeelding 2022-02-19 om 01 28 32 Schermafbeelding 2022-02-19 om 01 31 41


oktophonie (approved):

oktophonie:
Actually now I look at the tests: since the default `mmRestNumberPos` has now been changed in `styledef.cpp` all the tests/templates will need to be updated too. (@asattely can perhaps help make this much less painful as he's just had to do a similar thing in #10613 - maybe the easiest thing is to add that change to that PR).

Jojo-Schmitz:
I wonder whether the numbers atop these mmrests should be optional? They do seem redundant to me

Jojo-Schmitz:
Also see https://github.com/musescore/MuseScore/pull/10613#issuecomment-1045033944, so @asattely is on the templates

Jojo-Schmitz:
I wonder whether the numbers atop these (Multi)MeasureRepeats (!) should be optional (like they are for mmrests, and non-existent for the single MeasureRepeat)? They do seem redundant to me. Also have the potential to collide with barlines.

Also see https://github.com/musescore/MuseScore/pull/10613#issuecomment-1045033944, so @asattely is on the templates.

Jojo-Schmitz:
Also the "reset to default" and "set to style" buttons seems to be missing? Or is that something MU4 doesn't provide anymore in general?

oktophonie:
Indeed, the numbers are somewhat redundant on multi-measure repeats, since all the bars actually shown.

Jojo-Schmitz:
Not quite sure why they had been added in the first place, @IsaacWeiss ?

oktophonie:
I suppose if the multi-measure repeats were condensed to a single 'bar' (like mm rests) then having the number would make sense, i.e. something like
image
but whether that's ever been done in real life I must admit I don't know :)

Jojo-Schmitz:
In that case they a) would make sence and b) won't collide with barlines.

cbjeukendrup:
I updated the templates and everything else, just to make the tests pass on this PR (but maybe that change should still ultimately be part of #10613). I'm not totally convinced yet that it is a good idea to store _all_ style defaults in _all_ score files though, but that is a separate discussion.

I agree it seems sensible to make the number for measure repeats optional. I should probably better do that in a separate PR though, and maybe it should not have much priority right now.

Jojo-Schmitz:
Optional and even off by default? But yes, in a separate PR

cbjeukendrup:
In the VTests, it looked like the mmrest numbers got shifted to the right a tiny bit. After some puzzling, I found a solution. Now, all changes seem expected.

src/engraving/libmscore/mmrest.cpp
```diff
@@ -81,21 +81,21 @@ void MMRest::draw(mu::draw::Painter* painter) const
...
+
     if (score()->styleB(Sid::oldStyleMultiMeasureRests)
         && m_number <= score()->styleI(Sid::mmRestOldStyleMaxMeasures)) {
         // draw rest symbols
-        x = (m_width - m_symsWidth) * 0.5;
+        qreal x = (m_width - m_symsWidth) * 0.5;
         qreal spacing = score()->styleMM(Sid::mmRestOldStyleSpacing);
         for (SymId sym : m_restSyms) {
-            y = (sym == SymId::restWhole ? -spatium() : 0);
+            qreal&& y = (sym == SymId::restWhole ? -spatium() : 0);
```
RomanPudashkin, line 98: It is better not to use && here

RomanPudashkin (approved):

src/engraving/libmscore/mmrest.cpp
cbjeukendrup: Fixed!

PR #10617, opened 2022-02-19
[MU4] Top-staff system objects maintain position

asattely:
Top-staff system objects would be moved around with the staff they're attached to. This is the behavior of non-top staff system objects, but I added an exception for ones on track 0.

### DIFF
--- a/src/engraving/libmscore/segment.cpp
+++ b/src/engraving/libmscore/segment.cpp
@@ -859,15 +859,15 @@ void Segment::sortStaves(QList<int>& dst)
     }
     for (EngravingItem* e : _annotations) {
         ElementType et = e->type();
-        if (!e->systemFlag()
-            || (et == ElementType::REHEARSAL_MARK)
-            || (et == ElementType::SYSTEM_TEXT)
-            || (et == ElementType::PLAYTECH_ANNOTATION)
-            || (et == ElementType::JUMP)
-            || (et == ElementType::MARKER)
-            || (et == ElementType::TEMPO_TEXT)
-            || (et == ElementType::VOLTA)
-            || (et == ElementType::TEXTLINE && e->systemFlag())) {
+        if (!e->systemFlag() || (e->isLinked()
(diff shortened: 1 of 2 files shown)


oktophonie (approved):

oktophonie (approved):

src/engraving/libmscore/segment.cpp
```diff
@@ -859,15 +859,15 @@ void Segment::sortStaves(QList<int>& dst)
...
-        if (!e->systemFlag()
-            || (et == ElementType::REHEARSAL_MARK)
-            || (et == ElementType::SYSTEM_TEXT)
-            || (et == ElementType::PLAYTECH_ANNOTATION)
-            || (et == ElementType::JUMP)
-            || (et == ElementType::MARKER)
-            || (et == ElementType::TEMPO_TEXT)
-            || (et == ElementType::VOLTA)
-            || (et == ElementType::TEXTLINE && e->systemFlag())) {
+        if (!e->systemFlag() || (e->isLinked()
```
RomanPudashkin, line 862: This code looks very complicated to read/debug. I would suggest simplifying it. The most complicated part of it can be simplified as follows:

static const std::set<ElementType> allowedTypes {
ElementType::REHEARSAL_MARK,
ElementType::PLAYTECH_ANNOTATION
...
}

if (!e->systemFlag() || (e->isLinked() && allowedTypes.contains(et)) {
...
}

This condition can probably be simplified even further

oktophonie:
[Prevent duplicate system objects on one stave when moving instruments](https://trello.com/c/0iwJVLJn/37-prevent-duplicate-system-objects-on-one-stave-when-moving-instruments)

src/engraving/libmscore/segment.cpp
asattely: Good idea. I've updated it with that suggestion. `set::contains()` is  c++20 it seems, but `::find()` works fine for that.

RomanPudashkin (approved):

Issue #10618, opened 2022-02-19
Some of the Palette Options in MS3 have not been ported to MS4 (in the '...' menu)


Tantacrul:
**Describe the bug**
In MS3.6, there are a few useful options in the '...' menu, next to the search bar:

<img width="545" alt="image" src="https://user-images.githubusercontent.com/21022311/154809457-f0586ddb-07a7-43ce-add1-b2823f1a2512.png">

However, there is one option missing which used to be available in previous versions: the ability to open a palette by single-clicking on it. This is a common complaint from users of older versions of MuseScore. So, in addition to the options in the image above, we should also reintroduce the checkbox for the option '**Single-click to open a palette**'

They should be added to the new '...' menu here:
<img width="300" alt="image" src="https://user-images.githubusercontent.com/21022311/154809587-c6fc52cb-8853-43cc-84e2-b1bc1d1fa116.png">

I seem to recall that a member of the community did some work on this last year. Maybe @Howard-C or @cbjeukendrup? Perhaps I'm wrong.

cbjeukendrup:
Yes, it was me who worked on that for MU3.6 :)
I'd be happy to reimplement it for MU4.

Tantacrul:
Wow, that long ago?

Tantacrul:
@cbjeukendrup, I've removed everyone else and left the task to you. Cheers!

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED in #10619

PR #10619, opened 2022-02-20
Restore Palettes panel context menu options

cbjeukendrup:
Resolves: #10618
Resolves: https://musescore.org/en/node/305131
Schermafbeelding 2022-02-20 om 00 59 16

Until very recently I did not know about the 'Single-click to open a palette' option; that must have been before my time. I hope that it works as desired. I wonder why that has been absent since MU3.4, since it was not _that_ difficult to implement, and things like dragging still work. Hopefully I'm not missing something.

I also made a refinement for dragging palettes: now, the dragged palette will be positioned correctly relative to the mouse (instead of just with the top left corner at the position of the mouse).

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -214,6 +214,10 @@ DockPage {
 
             PalettesPanel {
                 navigationSection: palettesPanel.navigationSection
+
+                Component.onCompleted: {
+                    palettesPanel.contextMenuModel = contextMenuModel
+                }
             }
         },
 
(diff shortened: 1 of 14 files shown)


RomanPudashkin (approved):

Jojo-Schmitz:
That single-click got lost when the palettes got chanmged to be QML based.
Thanks for brining it back!

Jojo-Schmitz:
I don't recall though this having to be an option, IIRC it was the standard behavoir and no way to configure it?

cbjeukendrup:
I think having an option for it is just nice, but we could enable it by default. It really works quite nicely, so I'd say, why not?

Jojo-Schmitz:
Yes, option __and__ default, please ;-)

DmitryArefiev:
Tested #10618 and https://musescore.org/en/node/305131 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

Issue #7440, opened 2021-02-06
my interface is VERY BIG in 150% screen scale, how to solve it?


zarhani:
Hi I am using 2K screen with 150% screen scaling, and when I launch MuseScore, the interface is very big that buttons nearly take all the space of my screen...How to solve this problem?

MarcSabatella:
Hello!  You will probably need to either set, or unset QT_AUTO_SCREEN_SCALE_FACTOR, or else use the -D command line option to specify the correct resolution in DPI.  If you need further assistance, please ask for help in the Support forum on musescore.org, we do not use GitHub for this.

zarhani:
oops，I am sorry to discuss this in Github :-(   
I added -D command in the shortcut and It solved, thanks

Howard-C:
You can just close this issue.

PR #10621, opened 2022-02-20
Fix MS #7440 Displaying custom key signatures in multiple clefs

sammik:
Resolves: [*(MS #7440)*](https://musescore.org/en/node/7440)

Custom key signatures are adapted to different clefs.

Changes are made on top of #10609 commit 52ee1dd753744da5659735b29df6c2c14df21c71 which assumes, it is merged first.
Actual change ("keysig per clef") for review is second commit 0bb0d5ccf8070b92cc28db7ca9291a85da523a22

Tested on Linux (Ubuntu Studio 21.10)

vtest failure expected and welcome

KeySig_Clefs

This "just works", it is not ideal, but at the moment, I dont see better solution. (May be, I miss something). It is "solution a" as described bellow.

I think, for the future, custom key signatures needs deeper rethink / refactor. See "solution b" below

### Relation  between `key` and `keysignature` 
Convential
`key` is definition, `key symbols` in `keysignature` represent it, related on `clef`
`keySymbol.line` is representation of `key symbols` in concrete `clef`

Custom
definition is missing, `key symbols` are used to "fake" definition, but `clef` relation is missing inside `keysignature`.

### Possible solution
a) use `keySymbol.line` in custom key signatures as definition, which means, it is always stored as in treble clef.
- pros: easy adaptation at the moment; doesn't reuquire refactor
- cons: limited options for the future; different meaning of `keySymbol.line` in custom and convential sygnatures; 

b) add new "definition" of customkey in (or next to [^1]) `key` containing list of symbols, each defined as (my proposal):
 1) note (one of c, d, e, f, g, a, b)
 2) symbol (sharp, flat, natural, quatertone up, ...)
 3) xPos alteration (default xPos depends of symbol's index, 1.st symbol 1sp, 2.nd 2sp), so on second symbol xPos=-1 means it stays at 1sp.
 4) octave alteration (default line position depends on conventional position as defined in clefTable in clef.cpp)

- pros: clear logic; consistency of `keySymbol.line` meaning in both, conventional and custom signatures; new options and easier enhancements for future (transposing custom key signatures, microtonal accidentals, ...)
- cons: needs refactor

[^1]: Probably better to have them both, side by side, in case of possible visual modifications in editor as described in https://musescore.org/en/node/329177

### DIFF
--- a/src/engraving/layout/layoutmeasure.cpp
+++ b/src/engraving/layout/layoutmeasure.cpp
@@ -662,7 +662,7 @@ void LayoutMeasure::getNextMeasure(const LayoutOptions& options, LayoutContext& 
         const Drumset* drumset
             = staff->part()->instrument(measure->tick())->useDrumset() ? staff->part()->instrument(measure->tick())->drumset() : 0;
         AccidentalState as;          // list of already set accidentals for this measure
-        as.init(staff->keySigEvent(measure->tick()), staff->clef(measure->tick()));
+        as.init(staff->keySigEvent(measure->tick()));
 
         for (Segment& segment : measure->segments()) {
             // TODO? maybe we do need to process it here to make it possible to enable later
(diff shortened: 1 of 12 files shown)


Jojo-Schmitz:
Looks like (pun inteded) the vtests failures are expected so a `vtests` lable seems due

sammik:
Replaced by more complex #11280

Issue #10635, opened 2022-02-21
[MU4 Issue] Multiple pop-ups are being opened at the same time


jessjwilliamson:
**Describe the bug**
Pop-ups (commonly found in Properties) are being opened/activated at the same time when they should be opened one at a time.
This causes the pop-ups to overlap each other

https://user-images.githubusercontent.com/71656189/155008730-c0f0da4c-5c22-4667-9811-f7f8e19bbd5a.mp4

DmitryArefiev:
Tested on Win10, Mac12 - FIXED in #10640

PR #10640, opened 2022-02-22
[MU4] Inspector popups multi opening

Eism:
Resolves: #10635


src/inspector/view/qml/MuseScore/Inspector/InspectorForm.qml
```diff
@@ -44,6 +44,20 @@ Rectangle {
...
+        property var openedPopup: null
+
+        function closePreviousOpenedPopup(newOpenedPopup) {
+            if (Boolean(openedPopup) && openedPopup !== newOpenedPopup) {
+                openedPopup.close()
+            }
+
+            openedPopup = newOpenedPopup
+        }
+    }
```
Howard-C, line 59: Why is a `QtObject` needed here? Can't the variable and the function be declared directly?

Eism: The property declaration approach in QtObject is an imitation of private properties in qml(the engine does not provide such a possibility).

RomanPudashkin (approved):

DmitryArefiev:
Tested #10635 on Win10, Mac12 - FIXED      

Please merge into master

PR #10646, opened 2022-02-22
[MU4] Fix #329800: Crash on importing Capella .capx file

Jojo-Schmitz:
apparently due to a grace note attached to a rest.

Resolves: *https://musescore.org/en/node/329800*
Resolves:  #18446

Can't test this fix in MU4, apparently Capella import doesn't work at all currently (!?!), but in MU3, which has the same issue, this very same change (see e821bd2) does prevent the crash.

Edit meanwhile Capella import works in Mu4

### DIFF
--- a/src/importexport/capella/internal/capella.cpp
+++ b/src/importexport/capella/internal/capella.cpp
@@ -419,7 +419,7 @@ static bool findChordRests(BasicDrawObj const* const o, Score* score, const int 
     int graceNumber1 = 0;
     bool foundcr1 = false;
     Fraction tick2 = tick;
-    foreach (NoteObj* nobj, objects) {
+    for (NoteObj* nobj : objects) {
         BasicDurationalObj* d = 0;
         if (nobj->type() == CapellaNoteObjectType::REST) {
             d = static_cast<BasicDurationalObj*>(static_cast<RestObj*>(nobj));
@@ -468,14 +468,14 @@ static bool findChordRests(BasicDrawObj const* const o, Score* score, const int 
         if (seg->segmentType() != SegmentType::ChordRest) {
             continue;
         }
-        ChordRest* cr = static_cast<ChordRest*>(seg->element(track));
+        ChordRest* cr = toChordRest(seg->element(track));
         if (cr) {
-            if (graceNumber1 > 0) {       // the spanner is starting from a grace note
-                Chord* chord = static_cast<Chord*>(cr);
-                foreach (Chord* cc, chord->graceNotes()) {
+            if ((graceNumber1 > 0) && cr->isChord()) {       // the spanner is starting from a grace note
+                Chord* chord = toChord(cr);
+                for (Chord* cc : chord->graceNotes()) {
                     --graceNumber1;
                     if ((graceNumber1 == 0) && (!cr1)) {
-                        cr1 = static_cast<ChordRest*>(cc);             // found first ChordRest
+                        cr1 = toChordRest(cc);             // found first ChordRest
                     }
                 }
             }
@@ -489,14 +489,14 @@ static bool findChordRests(BasicDrawObj const* const o, Score* score, const int 
         if (seg->segmentType() != SegmentType::ChordRest) {
             continue;
         }
-        ChordRest* cr = static_cast<ChordRest*>(seg->element(track));
+        ChordRest* cr = toChordRest(seg->element(track));
         if (cr) {
-            if ((graceNumber > 0) && (cr->type() == ElementType::CHORD)) {       // the spanner is ending on a grace note
-                Chord* chord = static_cast<Chord*>(cr);
-                foreach (Chord* cc, chord->graceNotes()) {
+            if ((graceNumber > 0) && cr->isChord()) {       // the spanner is ending on a grace note
+                Chord* chord = toChord(cr);
+                for (Chord* cc : chord->graceNotes()) {
                     --graceNumber;
                     if ((graceNumber == 0) && (!cr2)) {
-                        cr2 = static_cast<ChordRest*>(cc);             // found 2nd ChordRest
+                        cr2 = toChordRest(cc);             // found 2nd ChordRest
                     }
                 }
             }
--- a/src/engraving/dom/measure.cpp
+++ b/src/engraving/dom/measure.cpp
@@ -1265,6 +1265,9 @@ void Measure::insertStaff(Staff* staff, staff_idx_t staffIdx)
 RectF Measure::staffabbox(staff_idx_t staffIdx) const
 {
     System* s = system();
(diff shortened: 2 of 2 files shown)


src/importexport/capella/internal/capella.cpp
```diff
@@ -463,14 +463,14 @@ static bool findChordRests(BasicDrawObj const* const o, Score* score, const int
         if (seg->segmentType() != SegmentType::ChordRest) {
             continue;
         }
-        ChordRest* cr = static_cast<ChordRest*>(seg->element(track));
+        ChordRest* cr = toChordRest(seg->element(track));
         if (cr) {
-            if (graceNumber1 > 0) {       // the spanner is starting from a grace note
-                Chord* chord = static_cast<Chord*>(cr);
-                foreach (Chord* cc, chord->graceNotes()) {
+            if (graceNumber1 > 0 && cr->isChord()) {       // the spanner is starting from a grace note
```
Jojo-Schmitz, line 468: This additional check is the core of this fix, the rest is more or less cosmetic.

src/importexport/capella/internal/capella.cpp
```diff
@@ -463,14 +463,14 @@ static bool findChordRests(BasicDrawObj const* const o, Score* score, const int
         if (seg->segmentType() != SegmentType::ChordRest) {
             continue;
         }
-        ChordRest* cr = static_cast<ChordRest*>(seg->element(track));
+        ChordRest* cr = toChordRest(seg->element(track));
         if (cr) {
-            if (graceNumber1 > 0) {       // the spanner is starting from a grace note
-                Chord* chord = static_cast<Chord*>(cr);
-                foreach (Chord* cc, chord->graceNotes()) {
+            if ((graceNumber1 > 0) && cr->isChord()) {       // the spanner is starting from a grace note
```
Jojo-Schmitz, line 468: This additional check is the core of this fix, the rest is more or less cosmetic.
(And that change is basically 'stolen' from further down)

src/importexport/capella/internal/capella.cpp
```diff
@@ -484,14 +484,14 @@ static bool findChordRests(BasicDrawObj const* const o, Score* score, const int
         if (seg->segmentType() != SegmentType::ChordRest) {
             continue;
         }
-        ChordRest* cr = static_cast<ChordRest*>(seg->element(track));
+        ChordRest* cr = toChordRest(seg->element(track));
         if (cr) {
-            if ((graceNumber > 0) && (cr->type() == ElementType::CHORD)) {       // the spanner is ending on a grace note
-                Chord* chord = static_cast<Chord*>(cr);
-                foreach (Chord* cc, chord->graceNotes()) {
+            if ((graceNumber > 0) && cr->isChord()) {       // the spanner is ending on a grace note
```
Jojo-Schmitz, line 489: (namely from here)

Jojo-Schmitz:
In a build from master? I wasn't able to open any in my local (Windows, MSVC) build

Jojo-Schmitz:
the test passed here too (and does run the Capella tests, so my problem got to be a local build issue)

Jojo-Schmitz:
Does crash when loading the score from https://musescore.org/en/node/329800, so back to the drawing board :-(

Jojo-Schmitz:
No need (nor way) to care for the hang in MinGW, as master doesn't build with MinGW anymore.
But now (after a rebase) the MCVS build hangs rather than crashes...

cbjeukendrup:
It looks like this would also resolve #18446.

Jojo-Schmitz:
I'll rebase it, after vacation (in about 2 weeks)

cbjeukendrup:
The changes in capella.cpp look good to me, but I've some doubts about those in measure.cpp. At least, I'd change it to:
```c++
IF_ASSERT_FAILED(s) {
    return RectF();
}
// else: do the same as before
```
because this method should just not be called when the measure has no system yet. 

It would be good to find out _why_ there is no system yet. So far, it looks to me like it's a problem with this score. In SystemLayout::collectSystem, there is a call to `layoutSystemElements`, and at that point all measures in that system should have their `system()` determined. But inside `layoutSystemElements`, some tuplet appears that is attached to a measure that does _not_ occur in the system; so perhaps the next measure, which has no system yet at that point. And when the score is loaded (after making the changes in this PR), it is reported as corrupted. 

Conclusion: maybe this fix is okay as a safety net to prevent crashes for corrupted scores, but then we should make clear that it's only meant as a safety net, using `IF_ASSERT_FAILED`. 

Another possibility is that the file is fine but the capella import code. That would be more difficult to find, of course...

Jojo-Schmitz:
Done, along with a fresh rebase

RomanPudashkin (approved):

zacjansheski:
The file from https://musescore.org/en/node/329800 doesn't crash on these builds, but it doesn't open either.  Score appears to be loading indefinitely.  Tried on MacOS and Windows 11

Jojo-Schmitz:
Well, that's what I stated on May 10...

Jojo-Schmitz:
It seems to spin on tzhis:
```cpp
            while (true) {
                double desiredY = proportionAlongX * (endY - startY) + startY;
                bool beamClearsAnchor = (m_up && RealIsEqualOrLess(desiredY, anchor.y() + reduction))
                                        || (!m_up && RealIsEqualOrMore(desiredY, anchor.y() - reduction));
                if (beamClearsAnchor) {
                    break;
                }

                if (isFlat || (isSmall && dictator == pointer)) {
                    dictator += m_up ? -1 : 1;
                    pointer += m_up ? -1 : 1;
                } else if (std::abs(dictator - pointer) == 1) {
                    dictator += m_up ? -1 : 1;
                } else {
                    pointer += m_up ? -1 : 1;
                }

                startY = (isStartDictator ? dictator : pointer) * m_spatium / 4 + tolerance;
                endY = (isStartDictator ? pointer : dictator) * m_spatium / 4 + tolerance;
            }
 ```
Apparently `beamClearsAnchor``never becomes true`

Jojo-Schmitz:
Looks like an entirely different issue, not related to the orginal one at all

Jojo-Schmitz:
But it seems that capella score has more than one problem, even in 3my 3.7, which has the same fix, only the crash is gone (and no hang), but the score gets reported as being corrupt. 
Problem is in pinpointing it its shere size, 960 measures...

Jojo-Schmitz:
NB: I don't see any tremolos in the score, so it got to be a beam causing that loop. There are a lot of them...

cbjeukendrup:
We'll merge this now because it's at least an improvement of the situation, and the problem Zac found earlier is not really related to the subject of this PR and not quite trivial to solve.

PR #10095, opened 2021-12-14
update max slopes

Nick-Mazuk:
Now that most of the refactored beams is sorted out, this PR adds the optimal max slopes.

**Note:** Unit tests still need to be fixed.

PR #10649, opened 2022-02-22
Fix beam unit tests

asattely:
These are the unit test adjustments for PR https://github.com/musescore/MuseScore/pull/10095

Beam slope has been changed, so the unit tests needed to be changed as well.

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -617,28 +617,30 @@ int Beam::computeDesiredSlant(int startNote, int endNote, int middleLine, int di
     }
     // for 2-indexed interval i (seconds, thirds, etc.)
     // maxSlopes[i] = max slope of beam for notes with interval i
-    static const int maxSlopes[] = { 1, 2, 3, 4, 5, 6 };
+    static constexpr std::array maxSlopes = { 1, 2, 3, 4, 5, 6, 7 };
 
     // calculate max slope based on distance between first and last chords
     qreal beamWidth = _elements[_elements.size() - 1]->stemPos().x() - _elements[0]->stemPos().x();
     beamWidth /= spatium();
-    int maxSlope = 6;
+    int maxSlope = maxSlopes.back();
     if (beamWidth < 3.0) {
         maxSlope = maxSlopes[0];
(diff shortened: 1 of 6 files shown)


asattely:
ignore this until it builds, i'm going to have to incorporate some changes from the draft PR before this will pass checks

oktophonie (approved):

RomanPudashkin:
You should fix the commit messages. For example, the last commit should describe what kind of bug it fixes. "extra tweaks" should probably be combined with the first commit. The PR description (and its title) should be corrected as well

Screenshot 2022-02-23 at 11 33 05

src/engraving/libmscore/beam.cpp
```diff
@@ -617,28 +617,30 @@ int Beam::computeDesiredSlant(int startNote, int endNote, int middleLine, int di
     }
     // for 2-indexed interval i (seconds, thirds, etc.)
     // maxSlopes[i] = max slope of beam for notes with interval i
-    static const int maxSlopes[] = { 1, 2, 3, 4, 5, 6 };
-
+    static const int maxSlopes[] = { 1, 2, 3, 4, 5, 6, 7 };
```
RomanPudashkin, line 620: Better to use constexptr std::array here

src/engraving/libmscore/beam.cpp
```diff
@@ -617,28 +617,30 @@ int Beam::computeDesiredSlant(int startNote, int endNote, int middleLine, int di
...
+    static const int numSlopes = sizeof(maxSlopes) / sizeof(maxSlopes[0]);
     // calculate max slope based on distance between first and last chords
     qreal beamWidth = _elements[_elements.size() - 1]->stemPos().x() - _elements[0]->stemPos().x();
     beamWidth /= spatium();
-    int maxSlope = 6;
+    int maxSlope;
```
RomanPudashkin, line 625: This should be initialized with some value by default

asattely:
Squashed, made improvements, as long as all the tests pass this one should be good to go

oktophonie (approved):

RomanPudashkin (approved):

PR #10650, opened 2022-02-22
[MU4] Remove all showSystemObjects functionality from ScoreGroup

asattely:
system object information does not rely on orders.xml anymore, so it shouldn't ever be in the <Section> xml element. ignore from previous versions.

This may cause the utests to fail because the xml is going to be slightly different, so let's see how that goes

### DIFF
--- a/src/engraving/utests/beam_data/beamNoSlope.mscx
+++ b/src/engraving/utests/beam_data/beamNoSlope.mscx
@@ -24,53 +24,6 @@
     <metaTag name="translator"></metaTag>
     <metaTag name="workNumber"></metaTag>
     <metaTag name="workTitle"></metaTag>
-    <Order id="orchestral">
-      <name>Orchestral</name>
-      <instrument id="voice">
-        <family id="voices">Voices</family>
-        </instrument>
-      <section id="woodwind" brackets="true" showSystemMarkings="true" barLineSpan="true" thinBrackets="true">
-        <family>flutes</family>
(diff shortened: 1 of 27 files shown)


oktophonie (approved):

asattely:
Taking this opportunity to remove <Order> tags from reference utest files that don't specifically test it because it's just unnecessary

oktophonie (approved):

src/engraving/utests/beam_data/beamNoSlope.mscx
```diff
@@ -24,53 +24,6 @@
     <metaTag name="translator"></metaTag>
     <metaTag name="workNumber"></metaTag>
     <metaTag name="workTitle"></metaTag>
-    <Order id="orchestral">
```
RomanPudashkin, line 27: Could you please explain how this will work now? I'm curious where the instrument order information will now be stored

asattely: There were only a few utest files that had this `<Order>` element in them; the vast majority don't. Since these tests aren't meant to test that object, there's no reason to have it in there. `beamNoSlope` for example shouldn't be concerned with the `<Order>` element at all. This doesn't reflect how normal scores are saved, this is just a change to make the utests more consistent.

RobFog:
Rebase needed.

RomanPudashkin (approved):

Issue #10631, opened 2022-02-21
[MU4 Issue] File extension is duplicated in "Save as..." dialog


oktophonie:
**Describe the bug**
When saving via the "Save as..." dialog, the file extension is duplicated in the file name field. If the file is saved, this is then reflected in the filename. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Open an existing file, or create a new one and save it once.
2. Go to "File -> Save as..."
3. Behold the "File name" field.

**Expected behaviour**
No duplicate of the file extension should be added to the end of the filename.

**Screenshots**
<img width="313" alt="image" src="https://user-images.githubusercontent.com/20604032/154981263-d493bff6-eb21-441c-a6e4-8c2da7fcbd68.png">

**Platform information**
 - Windows 11
 - macOS

Issue #10589, opened 2022-02-16
[MU4 Issue] "Do you want to save..." dialog should not show file extension


DmitryArefiev:
Doesn't occur with new scores.

1. Open recent score
2. Make some changes
3. Close score

![Capture](https://user-images.githubusercontent.com/90187801/154309293-b7f2e7eb-8d5e-4114-a4f6-87b9419033d8.PNG)

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

cbjeukendrup:
This happens to be fixed by PR #10552. (It will show the extension only when it is "interesting", i.e. when it is something different than mscz.)

PR #10652, opened 2022-02-22
Clean up some code relating to paths and saving status, preparing for Save to Cloud features

cbjeukendrup:
Closes: #10552
Resolves: #10288
Resolves: #10631

The solution with "SaveLocation" was unnecessarily complicated, so we keep it simple and identify scores by an `io::path`. There will be a hidden directory where cloud projects are stored/cached. We distinguish cloud scores by checking if their path is inside that hidden directory.

### DIFF
--- a/src/engraving/libmscore/masterscore.h
+++ b/src/engraving/libmscore/masterscore.h
@@ -112,6 +112,10 @@ class MasterScore : public Score
     // We can't yet, because m_project is not set on every MasterScore
     IFileInfoProviderPtr m_fileInfoProvider;
 
+    bool m_isNewlyCreated { false }; /// true if the file has never been saved yet
+    bool m_saved { false };
+    bool m_autosaveDirty { true };
+
     void reorderMidiMapping();
     void rebuildExcerptsMidiMapping();
     void removeDeletedMidiMapping();
@@ -221,6 +225,15 @@ public:
 
     IFileInfoProviderPtr fileInfo() const;
     void setFileInfoProvider(IFileInfoProviderPtr fileInfoProvider);
+
+    bool isNewlyCreated() const;
+    void setNewlyCreated(bool val);
+
+    bool saved() const;
+    void setSaved(bool v);
+
+    bool autosaveDirty() const;
+    void setAutosaveDirty(bool v);
 
     QString name() const override;
 
--- a/src/framework/global/io/path.cpp
+++ b/src/framework/global/io/path.cpp
@@ -29,6 +29,11 @@
 
 using namespace mu::io;
 
+path::path(const QByteArray& s)
+    : m_path(s)
+{
+}
+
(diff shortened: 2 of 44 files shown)


src/engraving/libmscore/masterscore.h
```diff
@@ -222,6 +226,15 @@ class MasterScore : public Score
     IFileInfoProviderPtr fileInfo() const;
     void setFileInfoProvider(IFileInfoProviderPtr fileInfoProvider);
 
+    bool isNewlyCreated() const { return m_isNewlyCreated; }
```
RomanPudashkin, line 229: Please, move the implementation to cpp

src/project/internal/projectautosaver.cpp
```diff
@@ -84,19 +82,40 @@ void ProjectAutoSaver::removeProjectUnsavedChanges(const io::path& projectPath)
...
+        return projectAutoSavePath;
+    }
+
     return io::filename(projectAutoSavePath, false);
 }
 
 mu::io::path ProjectAutoSaver::projectAutoSavePath(const io::path& projectPath) const
 {
-    return projectPath + AUTOSAVE_SUFFIX;
+    return projectPath + "." + AUTOSAVE_SUFFIX;
```
RomanPudashkin, line 94: You can reuse appendingSuffix here

src/framework/global/io/path.cpp
```diff
@@ -49,6 +54,24 @@ bool path::empty() const
     return m_path.isEmpty();
 }
 
+path path::appendingComponentsOf(const path& other) const
```
RomanPudashkin, line 57: Maybe just appendingComponent? Its name should be similar with appendingSuffix

src/framework/global/io/path.cpp
```diff
@@ -49,6 +54,24 @@ bool path::empty() const
...
+{
+    if (m_path.endsWith('/') || other.m_path.startsWith('/')) {
+        return m_path + other.m_path;
+    }
+
+    return m_path + '/' + other.m_path;
+}
+
+path path::appendingSuffix(const QString& suffix) const
```
RomanPudashkin, line 66: You can use io::path for suffixes

RomanPudashkin (approved):

src/project/internal/projectautosaver.cpp
cbjeukendrup: Nice one, thanks! Fixed (other comments too).

DmitryArefiev:
Tested #10288 #10631 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

@cbjeukendrup #10589 still occurs..

cbjeukendrup:
> #10589 still occurs..

Ah, yes, that's right. It will be fixed in the next PR in the series: #10662.

Issue #10653, opened 2022-02-23
[MU3 and MU4 Issue] Crash while converting to whole note


jason-cannon:
**Describe the bug**
A Crash to desktop occurs when selecting notes and converting them to whole notes twice. My guitar teacher was experiencing this issue and I am unsure if these are the steps he took exactly, but I reproduce the same CTD. I believe the issue has something to do with the mixture between notes and rhythm marks

**To Reproduce**
1. Open the attached muse score
2. Select the last measure with both notes and rythm marks
3. press 7 (convert to whole note) twice
4. CTD occurs

**Expected behavior**
Note should be converted to whole note and not crash

**Platform information**
 - OS: Windows

[Hero_of_War.zip](https://github.com/musescore/MuseScore/files/8121997/Hero_of_War.zip)

Jojo-Schmitz:
I can't reproduce.

Jojo-Schmitz:
Edit: I can (reading helps, press 7 twice)

Jojo-Schmitz:
Possible fix:
```
diff --git a/libmscore/edit.cpp b/libmscore/edit.cpp
index 3461bb9503..b455f30137 100644
--- a/libmscore/edit.cpp
+++ b/libmscore/edit.cpp
@@ -4130,7 +4130,7 @@ static Element* findLinkedVoiceElement(Element* e, Staff* nstaff)
       Measure* measure = segment->measure();
       Measure* m       = score->tick2measure(measure->tick());
       Segment* s       = m->findSegment(segment->segmentType(), segment->tick());
-      return s->element(dtrack);
+      return s ? s->element(dtrack) : nullptr;
       }

 //---------------------------------------------------------
@@ -4169,6 +4169,8 @@ static Chord* findLinkedChord(Chord* c, Staff* nstaff)
       Segment* s  = c->segment();
       Measure* nm = nstaff->score()->tick2measure(s->tick());
       Segment* ns = nm->findSegment(s->segmentType(), s->tick());
+      if (!ns)
+            return 0;
       Element* ne = ns->element(dtrack);
       if (!ne->isChord())
             return 0;
@@ -4922,6 +4924,8 @@ void Score::undoAddElement(Element* element)
                         if (cr1->staffIdx() != cr2->staffIdx())
                               sm = cr2->staffIdx() - cr1->staffIdx();
                         Chord* c1 = findLinkedChord(cr1, score->staff(staffIdx));
+                        if (!c1)
+                              return;
[...]

PR #10654, opened 2022-02-23
[MU4] Fix GH#10653: Crash converting to whole note

Jojo-Schmitz:
Resolves: #10653

Same issue with (and originally reported for) MU3

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -4830,6 +4830,9 @@ static EngravingItem* findLinkedVoiceElement(EngravingItem* e, Staff* nstaff)
     Measure* measure = segment->measure();
     Measure* m       = score->tick2measure(measure->tick());
     Segment* s       = m->findSegment(segment->segmentType(), segment->tick());
+    if (!s) {
+        return 0;
+    }
     return s->element(dtrack);
 }
 
@@ -4874,6 +4877,9 @@ static Chord* findLinkedChord(Chord* c, Staff* nstaff)
     }
     Measure* nm = nstaff->score()->tick2measure(s->tick());
     Segment* ns = nm->findSegment(s->segmentType(), s->tick());
+    if (!ns) {
+        return 0;
+    }
     EngravingItem* ne = ns->element(dtrack);
     if (!ne->isChord()) {
         return 0;
@@ -5668,6 +5674,9 @@ void Score::undoAddElement(EngravingItem* element, bool ctrlModifier)
                     sm = static_cast<int>(cr2->staffIdx() - cr1->staffIdx());
                 }
                 Chord* c1 = findLinkedChord(cr1, score->staff(staffIdx));
+                if (!c1) {
+                    return;
+                }
                 Chord* c2 = findLinkedChord(cr2, score->staff(staffIdx + sm));
                 Note* nn1 = c1->findNote(n1->pitch(), n1->unisonIndex());
                 Note* nn2 = c2 ? c2->findNote(n2->pitch(), n2->unisonIndex()) : 0;


Tantacrul:
Cool! Just tested it and it seems to be solving the crash. @RomanPudashkin / @Eism - if you are happy with this, it would be nice to merge (removes a P0).

Thanks @Jojo-Schmitz !

Jojo-Schmitz:
What about merging this?

Jojo-Schmitz:
Any particular reason __not__ to merge this?

Tantacrul:
@bkunda - when you get a moment, can you test this and assuming all's good, recommend to merge?

I can't quite manage it at the mo.

bkunda:
> @bkunda - when you get a moment, can you test this and assuming all's good, recommend to merge?
> 
> I can't quite manage it at the mo.

The logs for this run have apparently expired so I can no longer download any artefacts.

Jojo-Schmitz:
Downside of letting PRs sit for such a long time...
Rebased and force-pushed now.

src/engraving/libmscore/edit.cpp
```diff
@@ -4830,6 +4830,9 @@ static EngravingItem* findLinkedVoiceElement(EngravingItem* e, Staff* nstaff)
     Measure* measure = segment->measure();
     Measure* m       = score->tick2measure(measure->tick());
     Segment* s       = m->findSegment(segment->segmentType(), segment->tick());
+    if (!s) {
```
RomanPudashkin, line 4833: This looks very suspicious. Where do these null pointers come from? It looks like this PR fixes the consequence of the problem, not the cause

Jojo-Schmitz: Well, I don't quite understand the question, obviously this one comes from `findSegment()`.

Jojo-Schmitz: That in turn used `findSegmentR()`
```
//   findSegmentR
//    Search for a segment of type st at measure relative
//    position t.

Segment* Measure::findSegmentR(SegmentType st, const Fraction& t) const
{
    Segment* s;
    if (t > (ticks() * Fraction(1, 2))) {
        // search backwards
        for (s = last(); s && s->rtick() > t; s = s->prev()) {
        }
        while (s && s->prev() && s->prev()->rtick() == t) {
            s = s->prev();
        }
    } else {
        // search forwards
        for (s = first(); s && s->rtick() < t; s = s->next()) {
        }
    }
    for (; s && s->rtick() == t; s = s->next()) {
        if (s->segmentType() & st) {
            return s;
        }
    }
    return 0;
}
```

RomanPudashkin: Why does it return null in this case? It usually indicates some other, more complicated problem

Jojo-Schmitz: Dunno. The method does on occasions and on purpose return a `nullptr`. So callers should take care of those and not relying on it to never do that.

Jojo-Schmitz: If it were illegal, an assert would be due in that `findSegmentR()`

RomanPudashkin:
This PR fixes one symptom of the problem (crash in one scenario), but it doesn't fix the problem completely (removed element in the selection). For example, if I cherry-pick the changes from the PR, I can't select anything in the note input bar after the steps from the issue (due to the corrupted selection list). I guess we will be able to reproduce this crash in other scenarios as well (even with these changes)

Jojo-Schmitz:
Ah, I see.
Don't see the fix for that though :-(
Still won't invalidate my pointer checking in this PR, would it?

RomanPudashkin:
Let's try this fix https://github.com/musescore/MuseScore/pull/13183

It prevents removed elements from being added to the selected element list

Jojo-Schmitz:
That's fine, and of course better than my attempt here, but #13183 still allows Measure::findSegmentR(`)` to return 0, which still may cause a crash whenever the return falue of this is dereferenced. Should be prevented by an `assert()` at least.

Issue #10658, opened 2022-02-23
[MU4 Issue] Can't save to MSCX uncompressed format on macOS


shoogle:
**Describe the bug**
In the __Save As__ dialog, users can manually rename scores so the name ends `.mscx` to allow saving in the uncompressed format. This works on Windows and Linux, but on macOS a system dialog appears warning that this is not an allowed extension. This may be related to the fact that we do not give users an explicit option to save in the uncompressed format (see #9132).

**To Reproduce**
1. Create or open a score
2. Go to __File > Save As__
3. Chang name to `anything.mscx`
4. Click __Save__ and notice the warning dialog appear with no option to use `.mscx`

**Expected behavior**
The score should save in the uncompressed format, either as a single MSCX file or as a project folder containing an MSCX and other files usually found inside MSCZ archives. No warning dialog should appear.

**Screenshots**
<img width="906" alt="image" src="https://user-images.githubusercontent.com/11011881/155337946-10663bc2-ec27-4cdf-a112-437850d19283.png">

**Platform information**
 - OS: macOS

**Additional context**
It is important that users are able to save scores in an uncompressed format. This is useful for advanced editing, fixing corruptions, and storage in a Git repository. Ideally, the option should be visible in the Save As dialog (i.e. #9132 should also be fixed). It may even be necessary to fix #9132 to prevent the warning dialog on macOS from appearing.

Issue #9132, opened 2021-09-15
[MU4 Issue] "Save As..." dialog missing file type selector


bkunda:
In MuseScore 3.6, the "Save As..." dialog contains this dropdown, allowing users to select their file save format from between uncompressed .mscx  or the regular .mscz.
<img width="799" alt="Screenshot 2021-09-15 at 20 50 42" src="https://user-images.githubusercontent.com/86290556/133420655-0bf714cb-c126-490d-a496-942a5354ceae.png">

In the MacOS nightly, this dropdown selector is not visible.
In the Windows nightly, this dropdown only shows the option for .mscz

**Expected behavior**
.mscx should be available as an option in the dropdown selectors

Macbook Air M1, MacOS 11.5.2
Windows 10

wizofaus:
My understanding is that this is by design - I believe the intention to support saving as uncompressed files into a directory. .mscx files aren't guaranteed to hold all necessary information on their own, though they are very useful for troubleshooting and development purposes (e.g. for creating unit test reference files).  It's arguably not a huge loss of functionality as you can always unzip .mscz files yourself after saving, but that can be a bit annoying when you're used to just being able to save them directly in v3.

riaanvn:
Checking with a 2022-01-10 nightly on macOS 12.1 show be both the mscz and mscz file options (e.g. identical to the v3 options: 
<img alt="v4 save as" src="https://user-images.githubusercontent.com/4170969/149002345-8d3c182c-4fc3-4c6e-8b0f-1ad0ba9e8ad1.png">

Side note: this string still needs to be updated to reflect the new version; still referencing version 4.

Issue #10556, opened 2022-02-13
[MU4 Task] Opening files from previous major versions should be handled as import


cbjeukendrup:
**Task description**
When opening a file from MuseScore 1, 2, or 3, it should be handled as if the file is imported from other file formats. This way, the user will either have to confirm that they want to override the old file, or select a new location. Together with the "Migration" dialog when opening a score, this will help making the user aware that they are potentially overriding the old file that was compatible with the older version of MuseScore.

Tantacrul:
100%. Targeting at alpha.

DmitryArefiev:
Fixed in #11782

PR #10660, opened 2022-02-23
Restored the ability to save Uncompressed MuseScore files

cbjeukendrup:
Resolves: #9132
Resolves: #10658 (as far as possible)

Now, you can select a file type in the "Save score" dialog:
Schermafbeelding 2022-02-23 om 17 15 22

When you change the file type using this dropdown, the extension in the filename field is automatically updated. However, it does not work the other way round: if you type "mscx" as the extension while "Compressed MuseScore file" is selected in the dropdown, macOS will show the dialog as in #10658. We can't change this, since it is entirely controlled by Qt and macOS.

Also interesting: if you type an extension that is not registered with any app (for example, if you end the filename with ".hello"), macOS will add ".mscz" without letting you know (or mscx or mscs, based on the currently selected option in the dropdown). So "Score.hello" would be changed to "Score.hello.mscz".

I noticed that saving in mscx and mscs format reveals at least three bugs:
- Mscs files are not registered with the operating system, so macOS has no idea what it should do with them and opens them desperately in Text Editor

- When saving as mscx, a folder with the given filename is created, and inside that folder, a mscx file is created. But when opening that mscx file the next time and then saving it, another folder is created at the same level as the mscx file, and another mscx file is created inside that new folder. This gives a structure like this:
    ```
    Scores
     |-- MyScore
         |--MyScore.mscx
         |--style.mss
         |--…
         |--MyScore
             |--MyScore.mscx
             |--style.mss
             |--…
    ```
    This can continue endlessly...

- When selecting the save location for the new score, macOS thinks that a mscx file will be created in the selected folder. So, if a mscx file with the same name already exists in that directory, it will warn the user about replacing the file. However, in practice, a folder is created (and the mscx file will be inside), and macOS does not warn about replacing any existing folder with the same name. 

So, considering all these bugs, I'm not sure how good of an idea it is to expose these options to regular users...

### DIFF
--- a/src/project/internal/saveprojectscenario.cpp
+++ b/src/project/internal/saveprojectscenario.cpp
@@ -104,13 +104,21 @@ RetVal<io::path> SaveProjectScenario::askLocalPath(INotationProjectPtr project, 
 
     QStringList filter;
     filter << qtrc("project", "MuseScore file") + " (*.mscz)"
-           << qtrc("project", "Uncompressed MuseScore folder (experimental)") + " (*.mscx)"
+           << qtrc("project", "Uncompressed MuseScore folder (experimental)")
            << qtrc("project", "Uncompressed MuseScore file") + " (*.mscs)";
 
     io::path selectedPath = interactive()->selectSavingFile(dialogTitle, defaultPath, filter.join(";;"));
 
     if (selectedPath.empty()) {
         return make_ret(Ret::Code::Cancel);
+    }
+
+    if (!engraving::isMuseScoreFile(io::suffix(selectedPath))) {
+        // Then it must be that the user is trying to save a mscx file.
+        // At the selected path, a folder will be created,
+        // and inside the folder, a mscx file will be created.
+        // We should return the path to the mscx file.
+        selectedPath = selectedPath.appendingComponent(io::filename(selectedPath)).appendingSuffix(engraving::MSCX);
     }
 
     configuration()->setLastSavedProjectsPath(io::dirpath(selectedPath));
--- a/src/project/internal/saveprojectscenario.cpp
+++ b/src/project/internal/saveprojectscenario.cpp
@@ -102,9 +102,12 @@ RetVal<io::path> SaveProjectScenario::askLocalPath(INotationProjectPtr project, 
 
     io::path defaultPath = configuration()->defaultSavingFilePath(project, filenameAddition);
 
-    QString filter = qtrc("project", "MuseScore File") + " (*.mscz)";
+    QStringList filter;
+    filter << qtrc("project", "MuseScore file") + " (*.mscz)"
+           << qtrc("project", "Uncompressed MuseScore folder (experimental)") + " (*.mscx)"
+           << qtrc("project", "Uncompressed MuseScore file") + " (*.mscs)";
 
-    io::path selectedPath = interactive()->selectSavingFile(dialogTitle, defaultPath, filter);
+    io::path selectedPath = interactive()->selectSavingFile(dialogTitle, defaultPath, filter.join(";;"));
 
     if (selectedPath.empty()) {
         return make_ret(Ret::Code::Cancel);
--- a/src/project/internal/saveprojectscenario.cpp
+++ b/src/project/internal/saveprojectscenario.cpp
@@ -102,10 +102,15 @@ RetVal<io::path> SaveProjectScenario::askLocalPath(INotationProjectPtr project, 
 
     io::path defaultPath = configuration()->defaultSavingFilePath(project, filenameAddition);
 
-    QStringList filter;
-    filter << qtrc("project", "MuseScore file") + " (*.mscz)"
-           << qtrc("project", "Uncompressed MuseScore folder (experimental)")
-           << qtrc("project", "Uncompressed MuseScore file") + " (*.mscs)";
+    QStringList filter {
+        qtrc("project", "MuseScore file") + " (*.mscz)",
+        qtrc("project", "Uncompressed MuseScore folder (experimental)")
+#ifdef Q_OS_WIN
+        + " (*.)"
+#else
+        + " (*)"
+#endif
+    };
 
     io::path selectedPath = interactive()->selectSavingFile(dialogTitle, defaultPath, filter.join(";;"));
 
--- a/build/Linux+BSD/mscore.1.in
+++ b/build/Linux+BSD/mscore.1.in
@@ -100,7 +100,7 @@ GNU General Public Licence (GPLv2).
 @MAN_PORTABLE@Create symbolic links to make it easier to launch
 @MAN_PORTABLE@MuseScore from the command line when needed.
 @MAN_PORTABLE@.It
-@MAN_PORTABLE@Make sure MSCZ, MSCX and MusicXML files are recognised
+@MAN_PORTABLE@Make sure .MSCZ, .MSCX, .MSCS and MusicXML files are recognised
 @MAN_PORTABLE@by the system and the correct icons are displayed.
 @MAN_PORTABLE@.El
 @MAN_PORTABLE@.Pp
@@ -350,10 +350,12 @@ internal file sanity check log (JSON)
 MPEG Layer III (lossy compressed audio)
 .It Li mpos
 measure positions (XML)
-.It Li mscx
-uncompressed MuseScore file
 .It Li mscz
-compressed MuseScore file
+MuseScore File
+.It Li mscx
+MuseScore Uncompressed File
+.It Li mscs
+MuseScore Uncompressed File
 .It Li musicxml
 uncompressed MusicXML file
 .It Li mxl
--- a/build/Linux+BSD/musescore.xml.in
+++ b/build/Linux+BSD/musescore.xml.in
@@ -4,15 +4,30 @@
     <!-- MuseScore documents -->
 
     <mime-type type="application/x-musescore">
-        <comment>MuseScore compressed score</comment>
+        <comment>MuseScore File</comment>
         <glob pattern="*.mscz"/>
         <sub-class-of type="application/zip"/>
         <icon name="application-x-musescore@MSCORE_INSTALL_SUFFIX@"/>
     </mime-type>
 
     <mime-type type="application/x-musescore+xml">
-        <comment>MuseScore uncompressed score</comment>
+        <comment>MuseScore Uncompressed File</comment>
         <glob pattern="*.mscx"/>
+        <sub-class-of type="application/xml"/>
+        <root-XML namespaceURI="" localName="museScore"/>
+        <icon name="application-x-musescore@MSCORE_INSTALL_SUFFIX@+xml"/>
+        <magic>
+            <match type="string" value="&lt;?xml" offset="0">
+                <match type="string" value="&lt;museScore" offset="0:128">
+                    <match type="string" value="&lt;Score" offset="0:512"/>
+                </match>
+            </match>
+        </magic>
+    </mime-type>
+
+    <mime-type type="application/x-musescore+xml">
+        <comment>MuseScore Uncompressed File</comment>
+        <glob pattern="*.mscsx"/>
         <sub-class-of type="application/xml"/>
         <root-XML namespaceURI="" localName="museScore"/>
         <icon name="application-x-musescore@MSCORE_INSTALL_SUFFIX@+xml"/>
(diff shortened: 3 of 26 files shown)


shoogle:
> When selecting the save location for the new score, macOS thinks that a mscx file will be created in the selected folder.

What if the `.mscx` option was renamed "MuseScore Project Folder" with no file extension? Does this solve the issues with apparent conflicts, etc?

igorkorsukov (approved):

cbjeukendrup:
I fixed the bugs mentioned above; it seems to work ok now.

src/project/internal/saveprojectscenario.cpp
```diff
@@ -102,14 +102,25 @@ RetVal<io::path> SaveProjectScenario::askLocalPath(INotationProjectPtr project,
 
     io::path defaultPath = configuration()->defaultSavingFilePath(project, filenameAddition);
 
-    QString filter = qtrc("project", "MuseScore File") + " (*.mscz)";
+    QStringList filter;
+    filter << qtrc("project", "Compressed MuseScore file") + " (*.mscz)"
+           << qtrc("project", "Uncompressed MuseScore folder")
+           << qtrc("project", "Uncompressed MuseScore file") + " (*.mscs)";
```
RomanPudashkin, line 108: Do we really need to expose this mscs format for all users? What is the purpose of this?

DmitryArefiev:
@cbjeukendrup It doesn't specify (show) the extension for mscx files

https://user-images.githubusercontent.com/90187801/161520115-0ef1fc5c-2575-4deb-974b-ca5955003efd.mp4

src/project/internal/saveprojectscenario.cpp
shoogle: Now that MSCX is made up of multiple files, I think @igorkorsukov created the MSCS format as a way to have an uncompressed format that includes all data in a single file. This could be useful for storing scores in a Git repository, for example, though it's not an absolute necessity of course given that MSCX can be used for that purpose.

cbjeukendrup:
> It doesn't specify (show) the extension for mscx files when saving

@DmitryArefiev That is unfortunately necessary, because when saving a mscx file, you are actually choosing where the folder for that mscx file will be created, rather than the mscx file itself. 
The fact that Windows says "a file with this name already exists" might be a problem of Windows. It might be the case that it doesn't allow you to have a file and a folder where the filename without extension has the same name as the folder. But I'm not sure about that.

src/project/internal/saveprojectscenario.cpp
cbjeukendrup: Yes, that's what MSCS does. I could imagine that it is better to remove rather MSCX and keep MSCS. Because the MSCX format consists of multiple files in a folder, it is likely to cause users trouble (for example, any extra files that are put in the folder of a MSCX file will be lost when the user re-saves the MSCX file). The MSCS format is just a single file, so that's just safe.

DmitryArefiev:
@cbjeukendrup I have some doubts about putting back mscx format.. since it has a new form (several files in one folder comparing with a single file in MU3). In case when the user opens an mscz file he will get the same single file, but with mscx it will be a multiple files.. 

Also, I found a bug using that scenario: if to open an mscx created in MU3, then re-save it as mscx in MU4: 
1. file can't be closed after multiple saving 
3. MU4 deletes all contain of that folder with mscx file from MU3.


https://user-images.githubusercontent.com/90187801/161821796-e931f536-647e-45af-86ce-05ddd3ea6b55.mp4

And another bug with mscs format: I can reopen mscs file





https://user-images.githubusercontent.com/90187801/161823139-c3d92bf6-c3fc-4388-84cd-1d91a014bcaa.mp4

cbjeukendrup:
@DmitryArefiev I absolutely share your doubts. Handling MSCX files from MU3 properly will be tough, and will make the code not necessarily nicer... that would be another reason to not expose the MSCX format.

I fixed the problem with reading MSCS files.

DmitryArefiev:
> @DmitryArefiev I absolutely share your doubts. Handling MSCX files from MU3 properly will be tough, and will make the code not necessarily nicer... that would be another reason to not expose the MSCX format.
> 
> I fixed the problem with reading MSCS files.

@cbjeukendrup Yeah, I checked the issue about reopening MSCS and it seems to be fine now. But I guess we still need to think about putting back MSCX and MSCS formats..

@Tantacrul @bkunda @jessjwilliamson  Please review. Do we need to add the ability for common users to save scores in MSCX (new version) and MSCS formats?

jeetee:
I know I'm not one of the mentionees (that should be a word) but will advocate for the (small) userbase that uses text-based versioning systems. Those do need a way to save and work with uncompressed files.

cbjeukendrup:
I'm still inclined to think (the idea of) the old MSCX format was not that bad... but that may be naive.

MarcSabatella:
> I know I'm not one of the mentionees (that should be a word) but will advocate for the (small) userbase that uses text-based versioning systems. Those do need a way to save and work with uncompressed files.

I concur.  There are other use cases people have for this, like being able to do a quick global search & replace on text elements or font styles and other hacks.  These cases are not a huge deal if the MSCC can still be extracted from the MSCZ, but versioning does pretty much require text to be useful.  I have no idea what the complications are or the cost of implementing it, but the benefit is non-zero.

src/project/internal/saveprojectscenario.cpp
shoogle: I think both formats are imperfect but in different ways. I don't think it is possible to keep everyone satisfied with just one of them.

shoogle:
I store [1.5k scores](https://github.com/OpenScore/Lieder) in a Git repository so I certainly hope we keep at least one uncompressed format :)

> any extra files that are put in the folder of a MSCX file will be lost when the user re-saves the MSCX file

That seems like an implementation detail rather than inherent to the folder idea, though fixing it may not be entirely straightforward. If it's not easily fixable, I would support hiding the MSCX folder option behind an advanced preference. I wouldn't want to remove MSCX entirely because it is easier to parse than MSCS, and because the use of CDATA sections in MSCS breaks syntax highlighting in the XML and JSON trees.

> The MSCS format is just a single file, so that's just safe.

It's a single file until you [insert an image](https://musescore.org/en/handbook/3/images) in the score. Images are still carried around as separate files. They could be embedded as CDATA but it could be corrupted when editing in a text editor, and the sheer size of the image would cause many text editors to become unresponsive or crash.

DmitryArefiev:
Besides these bugs 
> 1. file can't be closed after multiple saving
> 2. MU4 deletes all contain of that folder with mscx file from MU3.

also found 


3. on **Windows**, still save can't **Uncompressed MuseScore folder** format - MU just saves a regular MSCZ file with mscz extension (without additional folders)
4. In file browser window when saving a file, I think regular format shouldn't shown as **Compressed MuseScore file**, it should be just **MuseScore file** (like in MU3 and in previous builds)

DmitryArefiev:
@Tantacrul Please review. Do we want to put MSCX (new version) and MSCS formats to regular saving mechanism or maybe to put somewhere to Export window or using plugins etc.?

As I see, most of issues are with MSCX (new version) file format (but I didn't finish all testing)

shoogle:
I just spoke about this with @Tantacrul. He said the following:

1. The MSCZ compressed option should just be called "MuseScore file".
2. Let's keep the MSCX "Uncompressed MuseScore folder" option.
    - This is a useful option for some users. It won't be the default in Save As so minor bugs are ok.
        - (Suggestion of my own: we could call it "Uncompressed MuseScore folder (experimental)" while it has bugs.)
    - For now, let's not try to fix the deleted files problem. Instead we should just show a warning when you save (including with Ctrl+S) that any extra files will be deleted. Ideally we would only show the warning if extra files are detected.
3. Let's remove the MSCS "Uncompressed MuseScore file" option for ordinary users.
    - @vpereverzev says the limitations of this format mean it is only useful to a tiny minority: probably just developers.
    - Therefore, @Tantacrul says we should keep the ability to save in this format but it should be hidden behind a "Developer Tools" option or similar.
        - It should be a general option to enable all developer tools. We don't want an advanced preference like "Enable saving in the MSCS format" because that will pollute preferences with options that are only useful to a tiny, tiny minority.

Tantacrul:
Thanks @shoogle!

That's actually a good general point: if a setting is something only contributors (or coders who are able to modify MuseScore themselves) will make use of, it should go in 'DevTools' and not in 'Advanced' in Preferences.

cbjeukendrup:
@DmitryArefiev I'm planning to continue working on this PR this week! I'll use the information from the comments here, so there's no need for a separate issue for now :)

cbjeukendrup:
@DmitryArefiev I finally fixed the issues mentioned above. Now, any previously existing files in a mscx folder should be preserved when saving a mscx. It turned out that there was a not-too-awkward way to do this, so I chose that solution, instead of creating a warning message.

Tantacrul:
Any reason why this hasn't been merged?

cbjeukendrup:
I believe the latest version is still being tested by @DmitryArefiev.

RobFog:
It would seem a rebase is necessary?

cbjeukendrup:
@RobFog Oh, yeah, that too. Thanks for the notice!

shoogle:
On Windows, I named my file "Untitled" with no file extension, and chose "Uncompressed folder" as the format:

image

Result: file "Untitled.mscz" was created (i.e. an MSCZ file rather than MSCX).

Then I tried saving with the name "Untitled.mscx". This succeeded in saving an MSCX file but it was in the current directory rather than in a new subdirectory with the chosen name.

cbjeukendrup:
So that problem still occurs. That's very unfortunate, since it seems to be the Windows file dialog that is causing this. Apparently it does not allow the situation that the user can choose between a file type with an extension and a file type without one. (The fact that the latter will be a folder in our case, does not influence the file dialog in any way.)

We could decide to take this bug for granted for now, and log and fix it separately. Maybe someone who has much experience with Windows file dialogs could advice us.

cbjeukendrup:
@shoogle Thanks a lot! The `(*.)` option seems to work!

shoogle:
@cbjeukendrup, awesome! Tested again on Windows:

1. Extra files added to the the Pictures, Thumbnails or META-INF subfolders are deleted on Save. (To create the Pictures folder, drag-and-drop an image file onto a note or frame in your score, then save.)

2. When saving MSCZ, the MSCX file inside is given the wrong file extension (it uses ".mscz" instead of ".mscx"), which results in an error about a corrupt file when attempting to reopen it.

DmitryArefiev:
@cbjeukendrup Tested on Win10, Mac12, LinuxMint 20.2 - Found some issues related to workflow with MU3 mscx scores

1. Video for the issue №2 from Peter:

> When saving MSCZ, the MSCX file inside is given the wrong file extension (it uses ".mscz" instead of ".mscx"), which results in an error about a corrupt file when attempting to reopen it.


https://user-images.githubusercontent.com/90187801/166694233-01e1bc5c-752f-4ee8-bb2c-14419274b9a6.mp4

2. When opening an old mscx MU3 scores from one folder and re-saving in a new version of mscx format, 
**MU4 creates folders and setting files inside that common folder but for all scores together and not separately for each score**. The issue can be solved if we trigger **Save As** dialog in that case and user can choose a regular mscz format or experimental uncompressed mscx format (with creating a new folder for each MU3 file)


https://user-images.githubusercontent.com/90187801/166698579-6f26674b-f7f5-49e8-99fa-33099681cadc.mp4

3. When opening an old mscx MU3 score and triggering Save As dialog, **MU4 takes the name of the folder and not the initial file name**

https://user-images.githubusercontent.com/90187801/166699049-ab416cc9-cdb9-465c-9ad4-ab3b126588c8.mp4

cbjeukendrup:
@shoogle 
> Extra files added to the the Pictures, Thumbnails or META-INF subfolders are deleted on Save

I did this on purpose for now; only direct child files of the folder containing the mscx will be preserved.

@DmitryArefiev I would like to solve the issues about opening MU3 scores separately, together with #10556.

DmitryArefiev:
@cbjeukendrup Tested all mentioned issues in the PR on Win10, Mac12, LinuxMint 20.2 - **FIXED** 
Can be merged to master (if code review is not required) 

Found one new issue on Linux, logged that separately #11615 
Other issues about opening MU3 scores will log later and link with #10556

igorkorsukov (approved):

PR #10661, opened 2022-02-23
[MU4] Fix #10593 Fix drag-and-drop of slurs

### LINKED ISSUE #10593: [MU4 Issue] Slur is positioned in a wrong place when added from the palettes by drag 'n' drop
https://user-images.githubusercontent.com/90187801/154445339-abf86105-9933-4ae7-95bc-d6006e4b0dd3.mp4

**Platform information**
 macOS, Windows, Linux

**Additional context**
Doesn't occur when adding slur with shortcut or adding slur from the palette by NOT drag 'n' drop
Doesn't occur in **MU3**

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/10593

Slurs can be added by clicking on the palette or pressing the hotkey, but dragging and dropping slurs wasn't working. This PR fixes that, but I'm not 100% confident about my fix being 'correct'. I would love some input about where this fix should have gone, if I put it in the wrong place.

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -67,6 +67,7 @@
 #include "libmscore/mscore.h"
 #include "libmscore/linkedobjects.h"
 #include "libmscore/tuplet.h"
+#include "libmscore/slur.h"
 
 #include "masternotation.h"
 #include "scorecallbacks.h"
@@ -1200,7 +1201,6 @@ bool NotationInteraction::drop(const PointF& pos, Qt::KeyboardModifiers modifier
     case ElementType::NOTE:
     case ElementType::CHORD:
     case ElementType::SPACER:
-    case ElementType::SLUR:
     case ElementType::BAGPIPE_EMBELLISHMENT:
     case ElementType::AMBITUS:
     case ElementType::TREMOLOBAR:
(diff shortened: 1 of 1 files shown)


DmitryArefiev:
Tested #10593 on Win10, Mac12, LinuxMint 20.2 - FIXED

oktophonie (approved):

src/notation/internal/notationinteraction.cpp
```diff
@@ -1111,6 +1112,7 @@ bool NotationInteraction::drop(const PointF& pos, Qt::KeyboardModifiers modifier
     case ElementType::VIBRATO:
     case ElementType::PALM_MUTE:
     case ElementType::HAIRPIN:
+        //case ElementType::SLUR:
```
RomanPudashkin, line 1115: It is better to remove all commented code. There is no need for it

src/notation/internal/notationinteraction.cpp
```diff
@@ -1240,6 +1242,17 @@ bool NotationInteraction::drop(const PointF& pos, Qt::KeyboardModifiers modifier
         accepted = true;
     }
     break;
+    case ElementType::SLUR:
+    {
+        EngravingItem* el = dropTarget(m_dropData.ed);
```
RomanPudashkin, line 1247: I'm curious how this worked before (for example, in MU3)... Because there may be a more general problem (for a lot of elements, not just slurs)

asattely: I looked into it and the MU3 code is completely different as far as I can tell. It looks like this code was written for MU4, and I wonder if dragging and dropping slurs worked at all since the move.

This problem could definitely affect other things but I made sure to test every spanner in the palette and they all seemed to work fine! If other problems are found we know where to look for the solution at least.

RomanPudashkin (approved):

PR #10663, opened 2022-02-24
instruments.xml: Fix saxophone pitch range and hand bell transposition

shoogle:
Run share/instruments/update_instruments_xml.py to fetch the latest online spreadsheet. Changes introduced here:

- Resolve https://musescore.org/en/node/329569 - Saxophone pitch range.

- Fix https://musescore.org/en/node/328336 - Hand Bells should sound an octave higher than written.

The hand bells have been given an 8va octave clef in concert pitch mode but not in transposed pitch mode (normal mode). This ensures notes appear on the same staff lines in both modes.

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -2963,8 +2963,8 @@
                   <musicXMLid>wind.reed.saxophone.soprano</musicXMLid>
                   <clef>G</clef>
                   <barlineSpan>1</barlineSpan>
-                  <aPitchRange>56-89</aPitchRange>
-                  <pPitchRange>56-94</pPitchRange>
+                  <aPitchRange>56-87</aPitchRange>
+                  <pPitchRange>56-91</pPitchRange>
                   <transposeDiatonic>-1</transposeDiatonic>
                   <transposeChromatic>-2</transposeChromatic>
                   <Channel>
@@ -3020,8 +3020,8 @@
                   <musicXMLid>wind.reed.saxophone.alto</musicXMLid>
                   <clef>G</clef>
                   <barlineSpan>1</barlineSpan>
-                  <aPitchRange>49-82</aPitchRange>
-                  <pPitchRange>49-87</pPitchRange>
+                  <aPitchRange>49-80</aPitchRange>
+                  <pPitchRange>49-92</pPitchRange>
                   <transposeDiatonic>-5</transposeDiatonic>
                   <transposeChromatic>-9</transposeChromatic>
                   <Channel>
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
I believe the templates using those instruments need to get adjusted too

Jojo-Schmitz:
Also you seem to have forgotten the 'plain' Saxophone (which is a Tenor Saxophone as far as I know)

Jojo-Schmitz:
Edit: no you didn't, sorry for the noise

share/instruments/instruments.xml
```diff
@@ -6639,14 +6639,18 @@
...
+                  <transposingClef staff="2">F</transposingClef>
+                  <concertClef staff="2">F8va</concertClef>
                   <aPitchRange>36-97</aPitchRange>
                   <pPitchRange>36-97</pPitchRange>
                   <singleNoteDynamics>0</singleNoteDynamics>
+                  <transposeDiatonic>7</transposeDiatonic>
+                  <transposeChromatic>12</transposeChromatic>
                   <Channel>
                         <!--MIDI: Bank 0, Prog 112; MS General: Tinker Bell-->
                         <program value="112"/> <!--Tinkle Bell-->
```
Jojo-Schmitz, line 6656: `Tinker Bell` vs. `Tinkle Bell`?

shoogle: "Tinker Bell" is the name of a [fictional character](https://en.wikipedia.org/wiki/Tinker_Bell) so it's probably a mistake here, although not necessarily since I could imagine this sound being used to represent that character (a fairy).

Anyway, if it's a mistake then it is a mistake in the MS General SoundFont rather than the spreadsheet or instruments.xml, which is correctly reporting the name used in the SoundFont. "Tinkle Bell" on the `<program>` line refers to the name in General MIDI / Roland GS which is [often different](https://docs.google.com/spreadsheets/d/1SwqZb8lq5rfv5regPSA10drWjUAoi65EuMoYtG-4k5s/edit#gid=133112496) to the name used in the SoundFont.

Jojo-Schmitz: Try to google "Tinkle Bell" and be surprised...

oktophonie (approved):

shoogle:
> I believe the templates using those instruments need to get adjusted too.

The plan is to remove all instrument data from template scores so that only the instrument IDs remain: the instrument data would be repopulated when the template is opened based on current data in instruments.xml. But if we don't get around to doing that then we will indeed need to update the existing templates with the new data manually.

Jojo-Schmitz:
That'd indeed be pretty good. Although It may not work for all templates, ISTR that some deliberatly deviate from instruments.xml
That for sure was the case for the SATB closes score templates (before the Women and Men 'instruments made it into instruments.xml), it is not needed there any longer, but there may be more/others.

PR #10665, opened 2022-02-24
[MU4] Fix staccato and tenuto appearing below notes when they shouldn't

asattely:
Before: 
image

After:
image

There was just some non-direction-agnostic code in the laying out of certain articulations, so that's been fixed.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -3746,18 +3746,19 @@ void Chord::layoutArticulations()
         bool headSide = bottom == up();
         qreal x = centerX();
         qreal y = 0.0;
-
+        const qreal halfBeamSp = score()->styleS(Sid::beamWidth).val() * score()->spatium() * 0.5;
         if (bottom) {
             if (!headSide && stem()) {
+                auto userLen = stem()->userLength();
                 if (_up) {
-                    y = downPos() - stem()->length();
+                    y = downPos() - stem()->length() - userLen;
                     if (beam()) {
-                        y -= score()->styleS(Sid::beamWidth).val() * _spatium * .5;
+                        y -= halfBeamSp * beam()->mag();
                     }
                 } else {
-                    y = upPos() + stem()->length();
+                    y = upPos() + stem()->length() - userLen;
                     if (beam()) {
-                        y += score()->styleS(Sid::beamWidth).val() * _spatium * .5;
+                        y += halfBeamSp * beam()->mag();
                     }
                 }
                 int line   = lrint((y + 0.5 * _spStaff) / _spStaff);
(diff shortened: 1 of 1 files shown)


asattely:
Added a bit of logic for ignoring manual adjustments to stem length, so things like this are possible:
image

Jojo-Schmitz:
Are these vtests differences wanted? Before:
layout-sequence-2-1 ref
After:
layout-sequence-2-1

Jojo-Schmitz:
But what about this, Before:
slurs-3-1 ref
After:
slurs-3-1
Esp. in measure 1 in bass clef that doesn't look good, and different from measure 2 in bass clef, without good reason?
Measurre 3 in basclef looks bad in theitthe case

oktophonie:
Yes, these are all exactly as expected. Here's this test in 3.6:

image

The articulations have their anchor set differently from bar to bar, which is the reason for the differences.

asattely:
Yeah, the differences are due to the anchor point not being respected, which this PR fixes

Jojo-Schmitz:
OK, so it seems 3.x is not affected by th at all is seems, Even if I backport this PR to 3.x I can't spot any difference, not in those 3 vtests at least.

oktophonie:
Yes, this was a regression introduced in 4.0 development (likely connected with the new beam code somewhere).

Jojo-Schmitz:
I see, thanks

oktophonie (approved):

src/engraving/libmscore/chord.cpp
```diff
@@ -3749,9 +3749,17 @@ void Chord::layoutArticulations()
...
         if (bottom) {
             if (!headSide && stem()) {
-                y = upPos() + stem()->length();
-                if (beam()) {
-                    y += score()->styleS(Sid::beamWidth).val() * _spatium * .5;
+                auto userLen = stem()->userLength();
+                if (_up) {
+                    y = downPos() - stem()->length() - userLen;
+                    if (beam()) {
+                        y -= score()->styleS(Sid::beamWidth).val() * _spatium * .5;
```
RomanPudashkin, line 3756: This line

score()->styleS(Sid::beamWidth).val() * _spatium * .5;

is present in 4 places. It is better to add a temporary variable for it and reuse the variable 4 times

asattely: Okay!

oktophonie:
[Fix vertically displaced articulation marks](https://trello.com/c/cgSmEOWX/7-fix-vertically-displaced-articulation-marks)

src/engraving/libmscore/chord.cpp
```diff
@@ -3746,18 +3746,19 @@ void Chord::layoutArticulations()
         bool headSide = bottom == up();
         qreal x = centerX();
         qreal y = 0.0;
-
+        const qreal halfBeamSp = score()->styleS(Sid::beamWidth).val() * _spatium() * 0.5;
```
Jojo-Schmitz, line 3749: `_spatium()` should rather be `_spatium`

src/engraving/libmscore/chord.cpp
```diff
@@ -3746,18 +3746,19 @@ void Chord::layoutArticulations()
...
         if (bottom) {
             if (!headSide && stem()) {
+                auto userLen = stem()->userLength();
                 if (_up) {
-                    y = downPos() - stem()->length();
+                    y = downPos() - stem()->length() - userLen;
                     if (beam()) {
-                        y -= score()->styleS(Sid::beamWidth).val() * _spatium * .5;
+                        y -= halfBeamSp * beam()->mag();
```
Jojo-Schmitz, line 3756: Doesn't `_spatium` (as part ot `halfBeamSp`) already contain `mag`?

asattely: `beam->mag()` can be different from `mag` (which is based on the magnitude of the chord) in situations like this
image
... as far as I can tell. So instead of using `_spatium` for that one I'll use `score()->spatium()` so that the chord magnitude isn't included in the calculation

Jojo-Schmitz: OK, still my last question: why not doing this only once?

asattely: i guess we could, but `beam()` is not guaranteed to be non-null until the check, so we'd need to add another check where the definition is. I don't see it as too big of a deal though

RomanPudashkin (approved):

PR #10668, opened 2022-02-25
[MU4] Fix #9995: Double/halve note duration skips values in note entry mode

### LINKED ISSUE #9995: [MU4 Issue] Double/halve note duration skips values in note entry mode
**Describe the bug**
Using key commands **Q** and **W** quarters/quadruples the note value in input mode (in both standard and tablature staves).

**Expected behaviour**
**Q** and **W** should (respectively) double/halve note duration consistently across standard and tablature staves. 
When in **note input** mode, these commands should effect the to-be-entered ChordRest.
When in **normal** mode, these commands should act on the currently selected ChordRest(s).

**To Reproduce**
1. Create a score
2. Select bar
3. Hit **c**
4. Hit **q** several times, hit **w** several times

**Side-by-side comparison**
double-half-note-duration

 - MacOS 11.6 (M1)
 - Windows 11

**Additional context**
See also #9914 and #8908

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/9995

### DIFF
--- a/src/framework/shortcuts/data/shortcuts-Mac.xml
+++ b/src/framework/shortcuts/data/shortcuts-Mac.xml
@@ -998,14 +998,6 @@
     <seq>NumPad+0</seq>
     </SC>
   <SC>
-    <key>pad-note-increase-TAB</key>
-    <seq>W</seq>
-    </SC>
-  <SC>
-    <key>pad-note-decrease-TAB</key>
-    <seq>Q</seq>
-    </SC>
-  <SC>
     <key>rest-TAB</key>
     <seq>;</seq>
     </SC>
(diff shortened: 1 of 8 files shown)


bkunda:
Tested in MacOS and Windows. Shortcuts work well across standard and TAB staves. 
One related change I did notice (from MS3 behaviour) is that the `0` key no longer enters an open string in tablature staves (instead, it enters a rest). Not sure if this was changed somewhere along the line, but I think it should continue to enter open string "0" -- particularly as this key is also necessary to enter the "10" fret. 
@RomanPudashkin let me know if this can be addressed in this PR, or if you'd like me to raise this separately :-)

RomanPudashkin:
>One related change I did notice (from MS3 behaviour) is that the 0 key no longer enters an open string in tablature staves

@bkunda I can reproduce this bug in the master, so let's rise it as a separate issue

Eism (approved):

bkunda:
> 

Sure! I'll do this now

PR #10673, opened 2022-02-25
[MU4] Fix #10655: Correct footer space in vertical justified layout calculations

### LINKED ISSUE #10655: [MU4 Issue] Interaction between vertical stave justification and footer collision avoidance is faulty
In MU4 the 'music top/bottom margin' works differently from before - it measures from the margin to the nearest stave line, not to the highest or lowest notation item - and logic has been added to also prevent music colliding with headers and footers where present. This was implemented in #9193. This very lovely illustration explains the difference (thanks @bkunda!):

footer-collision-v2

Currently this works exactly as intended around headers (that is: moving the note up does not cause the stave to move down until the note comes within 1sp of the header):

https://user-images.githubusercontent.com/20604032/155312584-bb293556-e332-4ce9-9a6b-70298b1f3352.mp4

And it works as intended around footers, when vertical justification of staves is turned **off**:

https://user-images.githubusercontent.com/20604032/155312699-aeeca074-5325-4c09-88c4-af491cbd627d.mp4

But when vertical justification of staves is turned **on**, it appears that the 'music bottom margin' is being interpreted the _old_ way, and measuring to the lowest item of notation on the bottom stave, and so the music is being pushed up even when the notation is nowhere near the footer:

https://user-images.githubusercontent.com/20604032/155312794-e3fa12bb-fc88-42f5-9560-412d39cf26b9.mp4

And it doesn't even do that entirely consistently, but I haven't yet figured out what the exact behaviour is. In any case, the behaviour should be identical to the previous video.

(For this example I increased the 'max. page fill distance' to be sure that it was nothing to do with that.)

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/10655

The original issue describes the issue very well with lots of useful visuals. This ended up being an error in the calculation of the 'remaining' space after the first layout pass. `footerMusicGap` is a const that represents the gap between the top of the footer and the bottom of the music in the bottom staff, labeled in the image in the linked issue as '1sp'.

### DIFF
--- a/src/engraving/layout/layoutpage.cpp
+++ b/src/engraving/layout/layoutpage.cpp
@@ -562,10 +562,14 @@ void LayoutPage::distributeStaves(const LayoutContext& ctx, Page* page, qreal fo
         }
     }
     --ngaps;
+    const qreal staffLowerBorder = score->styleMM(Sid::staffLowerBorder);
+    const qreal extraHeight = yBottom - prevYBottom;
+    const qreal combinedBottomMargin = page->bm() + footerPadding;
+    const qreal marginToStaff = page->bm() + staffLowerBorder;
+    qreal spaceRemaining{ qMin(page->height() - combinedBottomMargin - yBottom, page->height() - marginToStaff - prevYBottom) };
 
-    qreal spaceRemaining { page->height() - page->bm() - footerPadding - score->styleMM(Sid::staffLowerBorder) - yBottom };
     if (nextSpacer) {
-        spaceRemaining -= qMax(0.0, nextSpacer->gap() - spacerOffset - score->styleMM(Sid::staffLowerBorder));
+        spaceRemaining -= qMax(0.0, nextSpacer->gap() - spacerOffset - staffLowerBorder);
     }
     if (spaceRemaining <= 0.0) {
         return;


oktophonie:
Getting there!
The measurement, when vertical justification is _on_, is not yet correct. If I set Music Bottom Margin to 0sp we end up with this:

image

or with the footer removed:

image

In any case three distances are to be enforced:

1. 0sp between the bottom margin and the very lowest item on the bottom stave
2. 1sp between the footer, if present, and the very lowest item on the bottom stave
3. the distance specified by musicBottomMargin between the bottom margin and the bottom stave line of the bottom stave

with whichever of these that would put the bottom stave at the highest position prevailing. And notation (notes, etc) should be able to move within the musicBottomMargin space without causing the staves to move up/down until they get too close to either the margin or the footer (just as it works when vertical justification is turned off).

oktophonie:
[Interaction between vertical stave justification and footer collision avoidance is faulty](https://trello.com/c/C9vsUfEM/40-interaction-between-vertical-stave-justification-and-footer-collision-avoidance-is-faulty)

Jojo-Schmitz:
I guess the (rather small) vtests differences are expected and therefote a `vtests` label is in order?

oktophonie (approved):

RomanPudashkin (approved):

PR #10675, opened 2022-02-26
Fix memory management issue with StaffTypeChange

cbjeukendrup:
Resolves: [MS#329892](https://musescore.org/en/node/329892)

A crash would occur when enabling multi-measure rests where a staff type change existed at the last measure in a series of full-measure rests. In that situation, the StaffTypeChange of that last measure was `added` to the newly created MMRest measure; while doing so, there was an attempt do `delete` the StaffTypeChange's staff type, which pointed to a StaffTypeChange in the staff's StaffTypeList, and thus was a normal member, and not allocated via `new`. A crash occurs when trying to `delete` that kind of memory.

Note: this PR just solves the crash, which reveals that the interaction between mmrests and staff type changes is sometimes a bit strange...

### DIFF
--- a/src/engraving/libmscore/stafftypechange.h
+++ b/src/engraving/libmscore/stafftypechange.h
@@ -34,7 +34,8 @@ class StaffType;
 
 class StaffTypeChange final : public EngravingItem
 {
-    StaffType* _staffType { 0 };
+    StaffType* m_staffType { nullptr };
+    bool m_ownsStaffType = false;
     qreal lw;
 
     friend class mu::engraving::Factory;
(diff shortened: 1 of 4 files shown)


Jojo-Schmitz:
It is not only crashing _where a staff type change existed at the last measure in a series of full-measure rests._ but even if amidst a multimeasure rests (or amidst a bunch of empty measuresa wne ehen then enabling mm rests). But your change fixes that too.

src/engraving/libmscore/stafftypechange.h
```diff
@@ -46,14 +47,15 @@ class StaffTypeChange final : public EngravingItem
     void draw(mu::draw::Painter*) const override;
 
 public:
+    ~StaffTypeChange();
```
RomanPudashkin, line 50: You missed override here

RomanPudashkin (approved):

src/engraving/libmscore/stafftypechange.h
cbjeukendrup: Fixed, thanks!

PR #10687, opened 2022-02-28
[MU4] Fix #10644: Percussion Input Tab is unusable in dark mode. Positioning broken. Shortcut visualisations missing

### LINKED ISSUE #10644: Percussion Input Tab is unusable in dark mode. Positioning broken. Shortcut visualisations missing.
**Describe the bug**
See video below. The Note Input mode has the following problems:

- Illegible in dark mode: the BG should be the same as the score (the same as MS3 in dark mode)
- Shortcuts are missing
- Attempting to reposition the panel breaks it
- Expanding or contracting the app's size breaks the positioning of the panel too

https://user-images.githubusercontent.com/21022311/155123669-89aec483-5946-488b-9fc2-cfd21de7f881.mov

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10644

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -372,9 +372,15 @@ DockPage {
             objectName: pageModel.drumsetPanelName()
             title: qsTrc("appshell", "Drumset Tools")
 
-            height: 100
-            minimumHeight: 100
-            maximumHeight: 200
+            height: 64
+            minimumHeight: 64
+            maximumHeight: 64
+
+            //! NOTE: hidden by default
+            visible: false
+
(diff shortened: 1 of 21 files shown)


vpereverzev (approved):

bkunda:
@RomanPudashkin 

All mostly fixed here 🙂

The shortcut visualisations in Windows are a bit too big (they actually collide with the stave). 



Also I noticed the controls to undock the panel are gone. This is consistent with MS3, however, so I don't see it as being a problem at the moment. What does remain misleading is that there is still a resize handle that appears when you hover over the top of the panel. This probably shouldn't appear if the panel can't be resized (not a big issue though). 

I'll raise a new issue about some remaining interactions that still require implementing. 

Many thanks!

PR #10708, opened 2022-03-02
[MU4] instrument_id_recovering

vpereverzev:


RomanPudashkin (approved):

src/engraving/playback/metaparsers/notearticulationsparser.cpp
```diff
@@ -143,9 +143,13 @@ void NoteArticulationsParser::parseSpanners(const Ms::Note* note, const Renderin
 {
     for (const Ms::Spanner* spanner : note->spannerFor()) {
         int spannerFrom = spanner->tick().ticks();
-        int spannerTo = spanner->tick().ticks() + spanner->ticks().ticks();
+        int spannerTo = spanner->tick().ticks() + std::abs(spanner->ticks().ticks());
```
Jojo-Schmitz, line 146: IMHO `int spannerTo = spannerFrom + std::abs(spanner->ticks().ticks());` wouldl be clearer hwere (and in sync with how is is done above in chordarticulationsparser.cpp)

PR #10713, opened 2022-03-02
Improvements to horizontal spacing

mike-spa:
This PR contains several improvements that @oktophonie and I have been working on. Apart from posssible bug fixes or small corrections, this may be somewhat the "definitive" horizontal spacing for MU4.

The PR is split into 3 commits which are logically separate and may be also reviewed and merged separately, if needed.

1. I've rolled back the `acceptanceRange` concept. This used to let systems be initially slightly wider than the margins, and justify them by squeezing rather than stretching. The idea is good, but it is still too risky and it can break under some edge cases. Hopefully I can bring it back in future.

2. I've developed better logic to prevent the spacing of long notes to explode in the presence of very short ones. The previous logic was hacky and only worked with one specific spacing formula. The new one is mathematically solid, and works in general regardless of which spacing formula one may choose.

3. An overall improvement of the spacing algorithm.
- The spacing formula has been changed in favor of a better one.
- Some parameters, namely `minNoteDistance`, `stretch`, and `measureSpacing` interact differently with the spacing algorithm. The most significant change is `measureSpacing`. This parameter now does not act as an overall multiplier, but affects the *ratio* between the space assigned to longer vs shorter notes. This also required a minor UI change. The "Measure -> Spacing" style setting has been renamed to "Measure -> Spacing curve". Now this parameter can only take values between 1 and 2, where 1 corresponds to a flat spacing curve (meaning all notes receive equal space regardless of duration), and 2 corresponds to strict time-space layout. The default value of "Measure -> Spacing curve" is now 1.5. This required to change all the template files so that they load with the correct value. Additionally, I'm making sure that the value is reset to default for any score migrating from an older version of Musescore.

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -104,9 +104,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // we need to recompute the layout of the previous measures. When updating the width of these
         // measures, curSysWidth must be updated accordingly.
         if (ctx.curMeasure->isMeasure()) {
-            if (toMeasure(ctx.curMeasure)->computeTicks() < minTicks) {
+            if (toMeasure(ctx.curMeasure)->shortestChordRest() < minTicks) {
                 prevMinTicks = minTicks; // We save the previous value in case we need to restore it (see later)
-                minTicks = toMeasure(ctx.curMeasure)->computeTicks();
+                minTicks = toMeasure(ctx.curMeasure)->shortestChordRest();
                 changeMinSysTicks = true;
                 for (MeasureBase* mb : system->measures()) {
                     if (mb == ctx.curMeasure) {
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -171,9 +171,10 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // check if lc.curMeasure fits, remove if not
         // collect at least one measure and the break
 
-        double acceptanceRange = system->isSqueezable() ? 1.025 : 1; // A value slightly larger than 1 allows systems
-        // to be initially slightly larger than the target width and be justified by tightening rather than stretching.
-        // However, we must first make sure that the system *can* be tightened. isSqueezable() checks for that.
+        // acceptanceRange slightly larger than 1 allows systems to be initially slightly larger than the target width
+        // and be justified by squeezing rather than stretching. However, we must first make sure that the system *can*
+        // be squeezed. I'm temporarily rolling back the idea (still a bit too risky in some edge cases) [M.S.]
+        double acceptanceRange = 1;
         bool doBreak = (system->measures().size() > 1) && ((curSysWidth + ww) > systemWidth * acceptanceRange);
         if (doBreak) {
             breakMeasure = ctx.curMeasure;
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -4101,6 +4101,7 @@ void Measure::computeWidth(Segment* s, qreal x, bool isSystemHeader, Fraction mi
     qreal minNoteSpace = score()->noteHeadWidth() + score()->styleMM(Sid::minNoteDistance);
     qreal usrStretch = std::max(userStretch(), qreal(0.1)); // Avoids stretch going to zero
     usrStretch = std::min(usrStretch, qreal(10)); // Higher values may cause the spacing to break (10 is already ridiculously high and no user should even use that)
+    qreal durStretch = 1;
 
     while (s) {
         s->rxpos() = x;
@@ -4114,6 +4115,7 @@ void Measure::computeWidth(Segment* s, qreal x, bool isSystemHeader, Fraction mi
             s = s->next();
             continue;
         }
+
         Segment* ns = s->nextActive();
         while (ns && ns->allElementsInvisible()) {
             ns = ns->nextActive();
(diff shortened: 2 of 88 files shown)


MarcSabatella:
The "spacing curve" change is exciting, strict proportional spacing is a semi-common request!

src/engraving/libmscore/measure.cpp
```diff
@@ -4043,12 +4037,22 @@ float Measure::durationStretch(Fraction curTicks, const Fraction minTicks) const
...
+    // and all other ratios are scaled accordingly.
+    //------------------------------------------------------------------------
+    static constexpr double maxRatio = 32.0;
+    double minSysTicks = double(minTicks.ticks());
+    double maxSysTicks = double(system()->maxSysTicks().ticks());
+    double maxSysRatio = maxSysTicks / minSysTicks;
+    double ratio = double(curTicks.ticks()) / minSysTicks;
+    if (maxSysRatio > maxRatio) {
+        double A = minSysTicks * (maxRatio - 1) / (maxSysTicks - minSysTicks);
+        double B = (maxSysTicks - maxRatio * minSysTicks) / (maxSysTicks - minSysTicks);
```
asattely, line 4054: for these three lines, it's better to use parentheses to make sure the order of operations is what you want it to be. There have been a number of bugs I've been fixing recently that stem from people being unclear about the order of operations, so it's better to write
`((maxSysTicks - maxRatio) * minSysTicks) / (maxSysTicks - minSysTicks);`, for example

src/engraving/libmscore/measure.cpp
```diff
@@ -4057,11 +4061,13 @@ float Measure::durationStretch(Fraction curTicks, const Fraction minTicks) const
...
-    qreal str = 1 - slope + slope * sqrt(qreal(curTicks.ticks()) / qreal(minTicks.ticks()));
+    // qreal str = 1 - slope + slope * sqrt(ratio);
+    // Custom spacing
+    double str = pow(slope, log2(ratio));
 
-    if (minTicks > longNoteThreshold) {
+    if (minSysTicks > longNoteThreshold) {
         // Avoids long notes being too narrow in the absense of shorter notes.
-        str = str * (1 - 0.5 + 0.5 * sqrt(qreal(minTicks.ticks()) / qreal(longNoteThreshold.ticks())));
+        str = str * (1 - 0.6 + 0.6 * sqrt(minSysTicks / longNoteThreshold));
```
asattely, line 4070: a few magic numbers in here, it might help to make named constants to help people who want to work on this code in the future! Or at least a comment explaining what these `0.6` values are.

Jojo-Schmitz:
The mtests failures need to get fixed, most probably by adjusting the ref files

asattely:
> I guess I should probably just update all the vtest files to have the parameter already set to 1.5?
As far as I know, you don't need to update vtests manually! Someone with the ability will add the 'vtests' label to it as long as all the vtests show expected results

Jojo-Schmitz:
@shoogle ?

mike-spa:
> edit: Oh, I see what you mean. Yeah potentially you will have to fix the files the vtests use. Definitely ask someone like Peter about that

Yeah sorry I expressed myself quite poorly. What's confusing me is that if I manually open any (old) vtest file with MU4, the migration routine will update the "Measure -> Spacing" style parameter to its new default value, as it should. But this doesn't seem to happen during the actual running of the vtests. They seem to run without updating the parameter. So if we now put the vtest label, the new reference images would be generated with the wrong style value.

mike-spa:
I've added a further commit (again, it can be reviewed and merged separately if needed). Besides the 3 previous points, we now have:

4. Improvements of the spacing calculations when rythms overlap non-trivially. The best example is polyrythms. We go from this:
Screenshot from 2022-03-03 18-44-01
to this:
Screenshot from 2022-03-03 18-37-52
This ensures correct spacing as long as there is enough horizontal space. The proportions will still be distorted when notes start coming close to each other. This, however, will be solved when we let notes from different voices overlap vertically. This will require a complete overhaul of the collision checking algorithm.

Besides that, I've also implemented @asattely 's suggestions. Next, I'll look into those vtests and also fix the utests failures.

mike-spa:
The vtests now are all loading the correct default values of `measureSpacing = 1.5` (thanks @cbjeukendrup and @asattely for the great help). After some minor bug fixes, they now correctly show the updated horizontal spacing in all cases so, pending @oktophonie 's review, I think the vtest label is in order. utests (and there were *a lot* of them) have been also fixed. 
I'm now not planning to make any additional change, so it is in principle ready for review.

src/engraving/libmscore/segment.cpp
```diff
@@ -2691,4 +2691,28 @@ qreal Segment::minHorizontalDistance(Segment* ns, bool systemHeaderGap) const
...
+// shortest chordRest.
+//------------------------------------------------------
+Fraction Segment::shortestChordRest() const
+{
+    Fraction shortest = Fraction::max(); // Initializing at arbitrary high value
+    Fraction cur = Fraction::max();
+    for (auto elem : elist()) {
+        if (!elem || !elem->staff()->show() || !elem->isChordRest()) {
+            continue;
+        } else {
```
vpereverzev, line 2709: no need for "else" if you have "continue" above

src/engraving/layout/layoutsystem.cpp
```diff
@@ -171,9 +171,10 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
         // check if lc.curMeasure fits, remove if not
         // collect at least one measure and the break
 
-        double acceptanceRange = system->isSqueezable() ? 1.025 : 1; // A value slightly larger than 1 allows systems
+        //double acceptanceRange = system->isSqueezable() ? 1.025 : 1; // A value slightly larger than 1 allows systems
```
vpereverzev, line 174: don't leave it like this, better to remove this line if you don't need it

mike-spa:
> Has this been replaced by #10934?

Correct. I'm gonna close this one.

mike-spa:
Replaced by [#10934](https://github.com/musescore/MuseScore/pull/10934)

Issue #8969, opened 2021-08-30
[MU4 Issue] Refine 'auto' tuplet bracket mode behaviour


oktophonie:
1. Where a tuplet's 'Bracket type' is 'Auto', if the tuplet applies to only a part of a beamed group, but there is no subdivision of the internal beams matching the extent of the tuplet, the bracket _should_ be shown.

![tuplet-brackets](https://user-images.githubusercontent.com/20604032/131334774-47cd7ffa-8c6b-4853-b3ed-c81fb46aaf6e.png)

The first bar here shows the default beaming that MuseScore provides for this rhythm, in which case the absence of the bracket is fine. If the inner beams are explicitly joined, however, the bracket should be shown.

2. Conversely, if a beam is joined over a rest, and that beam now spans an entire tuplet group, the bracket should disappear:
3. 
![image](https://user-images.githubusercontent.com/20604032/146919413-c31c674e-b797-48b7-a7d1-d656b5caf4df.png)

Nick-Mazuk:
I'll take this one

Nick-Mazuk:
This may be worth waiting until the beaming project is finish. Determining whether or not the beam is broken requires code that will almost certainly change after rewriting the beaming code.

PR #10718, opened 2022-03-03
refine 'auto' tuplet bracket mode behaviour

Nick-Mazuk:
Resolves: https://github.com/musescore/MuseScore/issues/8969

Improves the 'auto' behavior for tuplet brackets. In short, if a tuplet's start/end notes are ambiguous, a bracket is automatically added.

Notes:

- This PR follows the guidelines laid out in #8969 with one addition. For a tuplet to not have brackets, it must be defined by beam breaks _and_ all notes inside the tuplet must have the same number of beams. For instance, the following example doesn't have a bracket:Screen Shot 2022-03-02 at 8 22 30 PM
Yet the following example does need a bracket even though it's defined by a beam break:Screen Shot 2022-03-02 at 8 23 32 PM
While this example may seem clear, I was noticing that more complex examples were not. I can be persuaded to change this behavior.
- The issue also includes the following example in which the second tuplet _should not_ have a bracket:
image
In theory, this PR fixes this issue. However, I could not figure out how to beam over the rest so I could not verify it.

@oktophonie ready for your review

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -812,7 +812,7 @@ void Beam::createBeamletSegment(Chord* chord, bool isBefore, int level)
         );
 }
 
-void Beam::calcBeamBreaks(Chord* chord, int level, bool& isBroken32, bool& isBroken64) const
+void Beam::calcBeamBreaks(const Chord* chord, int level, bool& isBroken32, bool& isBroken64) const
 {
     BeamMode beamMode = chord->beamMode();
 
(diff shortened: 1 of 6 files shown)


Nick-Mazuk:
vtests fail as desired

Jojo-Schmitz:
As to your question "how to beam over the rest" see https://musescore.org/en/node/11370

oktophonie (approved):

RomanPudashkin (approved):

Issue #9860, opened 2021-11-24
[MU4 Issue] Note Input with USB MIDI Keyboard doesn't work on Mac (macOS Monterey)


DmitryArefiev:
**Describe the bug**
I have Arturia KeyLab Essential 61 (connected via USB) to my M1 laptop with Monterey and it doesn't work in MU4 
(but works fine in MU3 on the same laptop)

All settings are set up  right (see screenshots)

**To Reproduce**
1. Connect USB MIDI Keyaboard
2. Click on bar
3. Hit N 
4. Play something on MIDI Keyaboard

**Screenshots**
MU3

![MU3](https://user-images.githubusercontent.com/90187801/143258073-f6b1ed3c-2fb6-41c3-a235-4be8e8d92596.png)

MU4

![MU4](https://user-images.githubusercontent.com/90187801/143258128-0cbcae3d-156e-47e1-915b-5e6043192910.png)

![MU4 play panel](https://user-images.githubusercontent.com/90187801/143258140-108ddf47-8459-48cd-b57f-a6896083e034.png)

**Desktop:**
 - OS: Mac 12.0.1

**Additional context**
Works fine on Windows10

DmitryArefiev:
Tested on Mac12 - FIXED in #10726

PR #10726, opened 2022-03-03
Fix MIDI input on macOS 11.0 and higher

cbjeukendrup:
Resolves: #9860

The culprit was bad use of memcpy in Event::fromRawData: I forgot to multiply the number of "words" (a.k.a. `uint32_t`) with the number of bytes per word (i.e. `sizeof(uint32_t)`).

### DIFF
--- a/src/framework/midi/midievent.h
+++ b/src/framework/midi/midievent.h
@@ -136,7 +136,8 @@ struct Event {
     static Event fromRawData(const uint32_t* data, size_t count)
     {
         Event e;
-        memcpy(e.m_data.data(), data, std::min(count, e.m_data.size()));
+        size_t numBytes = std::min(count, e.m_data.size()) * sizeof(uint32_t);
+        memcpy(e.m_data.data(), data, numBytes);
         return e;
     }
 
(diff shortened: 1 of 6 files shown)


src/framework/midi/midievent.h
```diff
@@ -136,7 +136,8 @@ struct Event {
     static Event fromRawData(const uint32_t* data, size_t count)
     {
         Event e;
-        memcpy(e.m_data.data(), data, std::min(count, e.m_data.size()));
+        size_t numBytes = std::min(count, e.m_data.size()) * sizeof(uint32_t);
```
Jojo-Schmitz, line 139: Better `sizeof(data)`?

cbjeukendrup: No, `sizeof(data)` would be the size of a pointer (to uint32_t), which equals 8, on 64bit systems. 
`sizeof(*data)` or `sizeof(data[0])`  would have been an option, but I don't think it would be _much_ better.

Jojo-Schmitz: That's what I meant, it it would be better, as then the data type could change without the code breaking

PR #10730, opened 2022-03-03
[MU4] Manual Beam Adjustments

asattely:
A merge / fix of https://github.com/musescore/MuseScore/pull/10130

Manual adjustments to beams are now possible. However, there were a few bugs to fix:
- Stem direction would sometimes be incorrect
- Thumbnail generation would launch layout() with different values that would move everything around
- Changing direction of stems within a beam would cause graphical problems
- A small 8th space slant, which is necessary for unmodified 32nd note beams, was still present when the beams were adjusted manually

This PR fixes all of those issues!


oktophonie:
To be superseded by a new version shortly.

PR #10731, opened 2022-03-03
[MU4] playback_tuning

vpereverzev:


RomanPudashkin (approved):

.gitignore
```diff
@@ -52,10 +52,6 @@ vtest/Pictures
 dependencies
 
 # Downloaded files during build process
-MuseScore_General.sf3
-MuseScore_General_License.md
-MuseScore_General_Changelog.txt
-MuseScore_General_Readme.md
```
cbjeukendrup, line 58: Now that we are adding the sound font to the repo, could we also remove the downloading from the build process? (i.e. DownloadSoundFont.cmake etc.)

Jojo-Schmitz: I'm not sure whether adding a ~80MB binary file to the repo is a good idea. A couple years back we got rid of that and for IMHO good reason

vpereverzev: > Now that we are adding the sound font to the repo, could we also remove the downloading from the build process? (i.e. DownloadSoundFont.cmake etc.)

let's keep them, they shouldn't run if there is already a sound-font in the "share" folder

cbjeukendrup: DownloadSoundFont.cmake rather compares the contents of the share/sound/SF_VERSION file with https://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General/VERSION.

Jojo-Schmitz:
Is it intentional that this really is the former HQ soundfont?

cbjeukendrup:
@Jojo-Schmitz Yes it is. The size of the HQ sf is still acceptable, and more importantly: the soundfont is now loaded in a different way, so that only the sounds that are actually used will be loaded, not the whole soundfont. So it won't cause much more memory usage, only better and more sounds.

thirdparty/fluidsynth/fluidsynth-2.1.4/src/synth/fluid_voice.c
```diff
@@ -196,7 +196,11 @@ static void fluid_voice_initialize_rvoice(fluid_voice_t *voice, fluid_real_t out
     fluid_voice_update_modenv(voice, FALSE, FLUID_VOICE_ENVFINISHED,
                               0xffffffff, 0.0f, 0.0f, -1.0f, 1.0f);
 
+#ifdef ENABLE_DEFAULT_COMPRESSION
```
Jojo-Schmitz, line 199: That's not really `Disable built-in Fluid IIR filter via environmental variable`, but rather via pre-processor...
Can't we have an API (or really an environment variable) for this?

Why has it been disabled in the first place? I see no reason being mentioned anywhere.

Please check https://github.com/Jojo-Schmitz/MuseScore/issues/438#issuecomment-2617832287

PR #10744, opened 2022-03-04
Added keyboard navigation for the Save to cloud dialogs

cbjeukendrup:
Added keyboard navigation for two dialogs:
- "How would you like to save?"
- "Save to cloud"

And fixed some problems with requesting focus when the "Save to cloud" dialog is opened from the "How would you like to save?" dialog. 

Also added a nice macro for the "Defer" struct, so that we can use it like this:
```c++
DEFER {
   doSomethingOnScopeExit();
   // It almost looks like a built-in language feature
}; // Except for that semicolon there
```

Inspired by what I've seen at the Swift project, but slightly simpler under the hood.

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledDialogView.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/StyledDialogView.qml
@@ -63,7 +63,7 @@ DialogView {
     }
 
     onOpened: {
-        navSec.requestActive()
+        Qt.callLater(navSec.requestActive)
     }
 
     onClosed: {
--- a/src/project/qml/MuseScore/Project/AskSaveLocationTypeDialog.qml
+++ b/src/project/qml/MuseScore/Project/AskSaveLocationTypeDialog.qml
@@ -64,6 +64,15 @@ StyledDialogView {
         RowLayout {
             spacing: 24
 
+            NavigationPanel {
+                id: optionsNavPanel
+                name: "SaveLocationOptionsButtons"
+                enabled: parent.enabled && parent.visible
+                direction: NavigationPanel.Horizontal
+                section: root.navigationSection
+                order: 1
+            }
+
             SaveLocationOption {
                 title: qsTrc("project", "To the Cloud (free)")
                 description: qsTrc("project", "Files are saved privately on your own personal account. \
@@ -71,6 +80,11 @@ You can share drafts with others and publish your finished scores publicly too."
                 buttonText: qsTrc("project", "Save to the cloud")
 
                 imageSource: "internal/SaveToCloud/images/Cloud.png"
+
+                navigation.panel: optionsNavPanel
+                navigation.column: 1
+                navigation.accessible.name: qsTrc("project", "Save to the cloud (free)")
+                navigation.accessible.description: description
 
                 onButtonClicked: {
                     root.done(SaveLocationType.Cloud)
@@ -84,6 +98,11 @@ You can share drafts with others and publish your finished scores publicly too."
 
                 imageSource: "internal/SaveToCloud/images/Laptop.png"
 
+                navigation.panel: optionsNavPanel
+                navigation.column: 2
+                navigation.accessible.name: qsTrc("project", "Save on your computer")
+                navigation.accessible.description: description
+
                 onButtonClicked: {
                     root.done(SaveLocationType.Local)
                 }
@@ -92,11 +111,24 @@ You can share drafts with others and publish your finished scores publicly too."
 
         SeparatorLine { Layout.margins: -24 }
 
+        NavigationPanel {
+            id: dontAskAgainPanel
+            name: "DontAskAgain"
+            enabled: dontAskAgainCheckbox.enabled && dontAskAgainCheckbox.visible
+            section: root.navigationSection
+            order: 2
+            accessible.name: dontAskAgainCheckbox.text
+        }
+
         CheckBox {
             id: dontAskAgainCheckbox
+
             width: parent.width
             text: qsTrc("project", "Don’t show again")
             checked: !root.askAgain
+
+            navigation.panel: dontAskAgainPanel
+            navigation.order: 1
 
             onClicked: {
                 root.askAgain = !root.askAgain
--- a/src/project/qml/MuseScore/Project/SaveToCloudDialog.qml
+++ b/src/project/qml/MuseScore/Project/SaveToCloudDialog.qml
@@ -78,15 +78,27 @@ StyledDialogView {
             spacing: 20
 
             StyledTextLabel {
+                id: titleLabel
                 text: root.visibility === CloudVisibility.Public
-                ? qsTrc("project", "Publish to MuseScore.com")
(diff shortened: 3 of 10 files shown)


Eism:
@cbjeukendrup I found a crash on your branch on macos(not reproducible on master and on Linux on your branch). But I can’t debug (for some reason, breake points are triggered in random places), because of it I can’t send the stack trace. Maybe you can debug

https://user-images.githubusercontent.com/10116828/156846127-7f53634b-0809-4a2a-8bed-82e06e4c42af.mov

cbjeukendrup:
@Eism thanks! I thought I had also seen this crash on the master branch, but that must have been something different; now I also can reproduce it only on my branch. 
I'm debugging it now, and it seems that the crash happens in StyledDialogView.qml, in the `onClosed` signal handler (that's why the stack trace is so strange: it happens in QML code). I think the problem is that `root.navigationParentControl` points to a control inside the "How would you like to save" dialog, which has already been deleted because it is closed. One of the reasons that I'm thinking this, is that the crash does not occur if you open the Save to cloud dialog directly via File > Save to cloud.
Would you have a suggestion on how to fix this? Maybe we should open the Save to Cloud dialog through something like Qt.callLater, or QMetaObject::invokeMethod in C++. But I would not be sure how to do that in `SaveProjectScenario::askSaveLocation`.

Eism:
@cbjeukendrup quick and probably right fix it is reseting navigationParentControl when it is destroyed

```
    if (!m_navigationParentControl) {
        ui::INavigationControl* ctrl = navigationController()->activeControl();
        //! NOTE At the moment we have only qml navigation controls
        QObject* qmlCtrl = dynamic_cast<QObject*>(ctrl);
        setNavigationParentControl(qmlCtrl);

        connect(qmlCtrl, &QObject::destroyed, this, [this]() {
            setNavigationParentControl(nullptr);
        });
    }
```

cbjeukendrup:
@Eism thanks, I did that, and it works!

src/project/qml/MuseScore/Project/AskSaveLocationTypeDialog.qml
```diff
@@ -72,6 +72,9 @@ You can share drafts with others and publish your finished scores publicly too."
 
                 imageSource: "internal/SaveToCloud/images/Cloud.png"
 
+                navigationPanel.section: root.navigationSection
+                navigationPanel.order: 1
```
Eism, line 76: I think, 'Save to the cloud' and 'Save to computer' should be inside the same panel for navigating with arrow keys instead of Tab

src/project/qml/MuseScore/Project/SaveToCloudDialog.qml
```diff
@@ -132,13 +151,32 @@ StyledDialogView {
             }
 
             RowLayout {
+                id: buttonsRow
                 Layout.alignment: Qt.AlignRight
                 spacing: 12
 
+                NavigationPanel {
+                    id: buttonsNavPanel
+                    name: "SaveToCloudOptions"
```
Eism, line 160: SaveToCloudButtons

src/project/qml/MuseScore/Project/SaveToCloudDialog.qml
```diff
@@ -132,13 +151,32 @@ StyledDialogView {
...
+                    enabled: buttonsRow.enabled && buttonsRow.visible
+                    direction: NavigationPanel.Horizontal
+                    section: root.navigationSection
+                    order: 2
+
+                    onNavigationEvent: function(event) {
+                        if (event.type === NavigationEvent.AboutActive) {
+                            event.setData("controlIndex", [saveButton.navigation.row, saveButton.navigation.column])
```
Eism, line 168: We usually solve this problem (button activation priority when panel is activated) by setting correct buttons order
Save button must be first (order = 1)

src/project/qml/MuseScore/Project/internal/SaveToCloud/SaveLocationOption.qml
```diff
@@ -92,6 +94,7 @@ ColumnLayout {
             FlatButton {
                 id: button
                 accentButton: true
+
```
Eism, line 97: When navigating through the buttons, the user should hear full information about button's section.
button.accessible.name = root.title + " " + root.description + " " button.text

src/project/qml/MuseScore/Project/SaveToCloudDialog.qml
```diff
@@ -101,6 +112,9 @@ StyledDialogView {
                         Layout.fillWidth: true
                         currentText: root.name
 
+                        navigation.panel: optionsNavPanel
+                        navigation.row: 1
```
Eism, line 116: need to add accessibility
accessible.name = <Save to cloud/musescore.com text> + " " + <Name label text> + " " currentText

src/project/qml/MuseScore/Project/SaveToCloudDialog.qml
```diff
@@ -118,12 +132,17 @@ StyledDialogView {
...
                         model: [
                             { value: CloudVisibility.Private, text: qsTrc("project", "Private") },
                             { value: CloudVisibility.Public, text: qsTrc("project", "Public") }
                         ]
 
                         currentIndex: indexOfValue(root.visibility)
+
+                        navigation.panel: optionsNavPanel
+                        navigation.row: 2
+
```
Eism, line 145: accessible.name: <Visibility text> + " " + currentText

src/project/qml/MuseScore/Project/internal/SaveToCloud/SaveLocationOption.qml
cbjeukendrup: I put the description in `accessible.description`. Wouldn't that be appropriate?

RomanPudashkin (approved):

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledDialogView.qml
```diff
@@ -63,7 +63,7 @@ DialogView {
     }
 
     onOpened: {
-        navSec.requestActive()
+        Qt.callLater(navSec.requestActive)
```
Eism, line 66: @cbjeukendrup 
Do we really need to call later the section activation?
Could you please explain what problem did you solve?

Because of this decision, it is now necessary to delay the activation of the control in all dialogs.
see https://github.com/musescore/MuseScore/pull/10906/commits/ade54f2e86b979e0a537b495e7aefc4d00c14668

cbjeukendrup: It was necessary in order to fix the activation when opening the "Save to cloud" dialog from the "How would you like to save?" dialog. I tried it now and the fix is still necessary. 

Now that I look at it again, I see that the reason that it wasn't working, was that there was a `callLater` in `onClosed` (a few lines below), to activate the parent control. If we do that without `callLater`, then we also don't need `callLater` in `onOpened`.

Issue #10742, opened 2022-03-04
[MU4 Issue] Paste text when editing doesn't work on Linux


DmitryArefiev:
**To Reproduce**
1. Double click on Title
2. Ctr+A to select all
3. Ctr+X to cut (text disappears)
4. Ctr+V to paste - **nothing happens**

Also doesn't work if to use commands from File menu>Edit>Paste

**Platform information**
Linux only

**Additional context**
Doesn't occur in **MU3**

Tantacrul:
Bumping to beta since it is an existing issue

DmitryArefiev:
> Bumping to beta since it is an existing issue

@Tantacrul I just realized that made a typo in the Description : Wanted to write "Doesn't **occur** in MU3 instead of "Doesn't **work** in MU3".. Let's put this again to Alpha2 (but with P2) since it's critical when working with text in score.

If disagree just move to Alpah2 again. 

Sorry for confusion!

DmitryArefiev:
FYI @Eism

DmitryArefiev:
Fixed in #12157

PR #10754, opened 2022-03-06
Check the mode of clipboard usage while copying depending on the platform

iamjithinjohn:
Resolves: https://github.com/musescore/MuseScore/issues/10742

Add platform dependant check while copying the selection to determine whether to use the global clipboard or global mouse selection and use the enum type corresponding to the part of the system clipboard used by setText().

https://user-images.githubusercontent.com/73242397/156912069-8ab1b6c6-32dc-40d5-8f44-c9783032cb5c.mp4

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -3089,7 +3089,12 @@ void NotationInteraction::copySelection()
         m_editData.element->editCopy(m_editData);
         Ms::TextEditData* ted = static_cast<Ms::TextEditData*>(m_editData.getData(m_editData.element).get());
         if (!ted->selectedText.isEmpty()) {
-            QGuiApplication::clipboard()->setText(ted->selectedText, QClipboard::Clipboard);
+            #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
+            QClipboard::Mode mode = QClipboard::Clipboard;
+            #else
+            QClipboard::Mode mode = QClipboard::Selection;
+            #endif
+            QGuiApplication::clipboard()->setText(ted->selectedText, mode);
         }
     } else {
         QMimeData* mimeData = selection()->mimeData();


src/notation/internal/notationinteraction.cpp
```diff
@@ -3089,7 +3089,12 @@ void NotationInteraction::copySelection()
         m_editData.element->editCopy(m_editData);
         Ms::TextEditData* ted = static_cast<Ms::TextEditData*>(m_editData.getData(m_editData.element).get());
         if (!ted->selectedText.isEmpty()) {
-            QGuiApplication::clipboard()->setText(ted->selectedText, QClipboard::Clipboard);
+            #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
```
RomanPudashkin, line 3092: I've seen this piece of code in three places already... I think we should add our wrapper (to the ui module) for QClipboard and use it instead of QApplication::clipboard() throughout the application. The interface for it will look like this:

```
namespace mu::ui {
class IClipboard {
public:
    virtual void setText(const QString& text) = 0;
    virtual QString text() const = 0;
    
    const QMimeData* mimeData() = 0;
    virtual void setMimeData(QMimeData* data) = 0;
}
}
```

And the implementation for the  setText method:

```
void Clipboard::setText(const QString& text) 
{
#if defined(Q_OS_MAC) || defined(Q_OS_WIN)
    QClipboard::Mode mode = QClipboard::Clipboard;
#else
    QClipboard::Mode mode = QClipboard::Selection;
#endif
    QGuiApplication::clipboard()->setText(ted->selectedText, mode);
}
```

DmitryArefiev:
Hi @iamjithinjohn, can you rebase on last master please?

RomanPudashkin:
Thanks for your help! I cherry-picked your commit:
https://github.com/musescore/MuseScore/pull/12157

Issue #10751, opened 2022-03-05
Return default UI font to OS default. Goodbye Fira Sans! 


Tantacrul:
**Describe the bug**
As it says on the tin: the OS defaults are simply more legible than Fira Sans.

Also, pay attention to #10736 to make sure that the strange 'no space after a comma' problem occurs.

cbjeukendrup:
I suddenly had a concern about this. So apparently, the system font used by Qt on macOS is a kind of crippled version where the comma has no width and there are probably other problems too. 
When I install SF Pro manually via Apple's developer website, and use that instead of the system font, it looks okay. But we are not allowed to package this font with MuseScore, as stated by its license. So... what do we do? It looks like our only options are using the crippled system font or using a non-default font on macOS.

cbjeukendrup:
There is also some good news: the problem seems fixed in Qt 6.

vpereverzev:
let's keep Fira Sans until 4.1 then. After MU4.0 we'll migrate on Qt6, it does seem to be quite stable already

vpereverzev:
> let's keep Fira Sans until 4.1 then. After MU4.0 we'll migrate on Qt6, it does seem to be quite stable already

@Tantacrul Are you OK with that?

Tantacrul:
> @Tantacrul Are you OK with that?

Annoyingly, I don't think we can stick with Fira. It look really, really bad on Windows and not great on Mac either. On any machine I've used, reverting to the OS default instantly makes the app look a lot better and more legible.

Releasing with a clearly inferior font is just too big a deal. I've heard a lot of complaints about it already and nothing positive.

cbjeukendrup:
@Tantacrul @vpereverzev I created this solution: https://github.com/musescore/MuseScore/pull/10758
Do you agree about it?

cbjeukendrup:
I tried calling the native macOS APIs, and they also return `.AppleSystemUIFont` as the default font. So the problem is really in Qt and how Qt _uses_ that font. 

The comma issue also occurs in the very simplest possible Qt app, so it's not something that we are doing wrong. 

I found this report (among others about the same issue) in the Qt Bug Tracker: https://bugreports.qt.io/browse/QTBUG-88495 
This suggests that it was fixed in Qt 5.15.3, which we can't use because it is commercial only 😭.

cbjeukendrup:
This is the patch they used to fix it: https://codereview.qt-project.org/c/qt/qtbase/+/322102
[...]

Issue #10736, opened 2022-03-04
Spaces after commas not working. Typo correction & copy change for 'Learn' section


Tantacrul:
**Describe the bug**
The text when videos can't load in the 'Learn' section needs a tiny bit of revision. There are two ',' symbols with no spaces after them and (being pedantic) you shouldn't have the word 'but' after a comma.

<img width="1453" alt="image" src="https://user-images.githubusercontent.com/21022311/156675766-eddbead1-4ba6-470f-b040-49fa7a400a03.png">

**Desired**
**"Sorry, we are unable to load these videos right now"**
"Please check your internet connection or try again later"

cbjeukendrup:
Actually, there _is_ a space, but apparently in that font something strange happens with commas. Here's how it looks in Fira Sans:
<img width="644" alt="Schermafbeelding 2022-03-04 om 01 36 11" src="https://user-images.githubusercontent.com/48658420/156676713-1feb0407-f102-477c-98ff-b740ba303747.png">
I remember that I've seen that problem with commas eating spaces earlier... I think it was in the score migration dialog in MU3.6.

Tantacrul:
Interesting! Although the copy should change anyway... we should look into this.

We're definitely planning to revert back to the OS default UI fonts anyway. This example is SF Pro.

Tantacrul:
I've changed it to a P2 since it is shaping up to be a more significant problem.

cbjeukendrup:
@Tantacrul curious observation:
- If you choose `.AppleSystemUIFont` in the dropdown in Preferences, this issue occurs;
- If you choose `SF Pro`, the issue does not occur. 

So apparently, `.AppleSystemUIFont` points to a wrong version of the font. 

Also, there is SF Pro, SF Pro Display, SF Pro Rounded and SF Pro text. Do you know which is the correct one?

Somehow, in MuseScore, these system fonts don't look like in the rest of macOS to me. But it must be an optical illusion, because if I take a screenshot of MuseScore and overlay it on a screenshot of another app, it turns out to be perfectly the same.

PR #10758, opened 2022-03-06
Switch back to the system font as the default UI font

cbjeukendrup:
Resolves: #10751
Temporary partial solution for: #10736 (the part about spaces after commas)

So, the problem is that Qt makes a mistake when rendering text in the default macOS font, causing spaces after commas and periods to disappear. This will be solved when we switch to Qt 5.15.3 or later or Qt 6. 
For the meantime, this PR includes a workaround for the (very) few users who have installed the SF Pro font manually, because in the manually installed version the problem does not occur.


cbjeukendrup:
@Tantacrul @vpereverzev would it be an idea to merge this PR as it is now, before the next alpha release, and later revisit the idea of switching to Qt 5.15.3 to fix the comma rendering issue?

Tantacrul:
I have no objection

vpereverzev (approved):

PR #10761, opened 2022-03-08
Fix #321095: Support MusicXML 4.0 (export)

lvinken:
Resolves: https://musescore.org/en/node/321095

Update the MusicXML version to 4.0 on export (second and final part of #321095). This prepares for future use of MusicXML 4.0 features.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -7265,9 +7265,9 @@ void ExportMusicXml::write(QIODevice* dev)
     _xml << "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n";
     _xml
         <<
-        "<!DOCTYPE score-partwise PUBLIC \"-//Recordare//DTD MusicXML 3.1 Partwise//EN\" \"http://www.musicxml.org/dtds/partwise.dtd\">\n";
+        "<!DOCTYPE score-partwise PUBLIC \"-//Recordare//DTD MusicXML 4.0 Partwise//EN\" \"http://www.musicxml.org/dtds/partwise.dtd\">\n";
 
-    _xml.startObject("score-partwise version=\"3.1\"");
+    _xml.startObject("score-partwise version=\"4.0\"");
 
     work(_score->measures()->first());
     identification(_xml, _score);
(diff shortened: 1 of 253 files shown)


Jojo-Schmitz:
First part (Import MusicXML 4.0, resp. update the validation schema files) was in #10060

Jojo-Schmitz:
Any known (and good) reason why the vtests fail?

src/importexport/musicxml/tests/data/testMaxNumberLevel.mscx
```diff
@@ -0,0 +1,3198 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Jojo-Schmitz, line 1: This test is not yet used, or is it?

Jojo-Schmitz: Adding `void maxNumberLevel() { mxmlImportTestRef("testMaxNumberLevel"); }` to https://github.com/musescore/MuseScore/blob/master/src/importexport/musicxml/tests/tst_mxml_io.cpp seems to work and pass

src/importexport/musicxml/tests/data/testMaxNumberLevel_ref.xml
```diff
@@ -0,0 +1,2315 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
+<score-partwise version="3.1">
```
Jojo-Schmitz, line 3: Shouldn't this be 4.0?

Jojo-Schmitz: That is what's keeping the unit test from passing

lvinken:
Fixed. Should have mentioned it yesterday: I wasn't done yet.

Jojo-Schmitz:
Thanks.

vpereverzev:
@lvinken looks like some of your recent changes has broken a couple of MXML unit tests.

a876cef65fa1053e4fcc508c34377f47027090f8

Could you please take a look?

Jojo-Schmitz:
And it most likely isn't @lvinken's failt but mine, from #10586's 2nd commit, which adds/enables a bunch of those tests, which in turn were not prepared for MusicXML 4.0

PR #10763, opened 2022-03-08
Revive the "Follow system theme" feature

cbjeukendrup:
There is now an opt-in feature that lets MuseScore follow the OS theme regarding light vs dark. This also works if high contrast is enabled. 
On some OSs, it would also be possible to let MuseScore follow the OS regarding high contrast. That seems less useful to me though, so I've removed the code for it for now.

### DIFF
--- a/src/appshell/view/firstlaunchsetup/themespagemodel.cpp
+++ b/src/appshell/view/firstlaunchsetup/themespagemodel.cpp
@@ -33,9 +33,32 @@ ThemesPageModel::ThemesPageModel(QObject* parent)
 
 void ThemesPageModel::load()
 {
+    uiConfiguration()->isFollowSystemTheme().notification.onNotify(this, [this]() {
+        emit isFollowSystemThemeChanged();
+    });
+
     uiConfiguration()->currentThemeChanged().onNotify(this, [this]() {
         emit themesChanged();
     });
+}
+
+bool ThemesPageModel::isFollowSystemThemeAvailable() const
+{
+    return uiConfiguration()->isFollowSystemThemeAvailable();
+}
+
+bool ThemesPageModel::isFollowSystemTheme() const
+{
+    return uiConfiguration()->isFollowSystemTheme().val;
+}
+
+void ThemesPageModel::setFollowSystemTheme(bool enabled)
+{
+    if (enabled == isFollowSystemTheme()) {
+        return;
+    }
+
+    uiConfiguration()->setFollowSystemTheme(enabled);
 }
 
 ThemeList ThemesPageModel::allThemes() const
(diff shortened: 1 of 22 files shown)


Jojo-Schmitz:
rebase needed

Jojo-Schmitz:
Wouldn't a CI function that reports such needed rebases automagically really nice?
Would it be possible?

cbjeukendrup:
Months ago I proposed to Github to make this a built-in feature: https://github.com/github/feedback/discussions/4335 (didn't get any response though...)
Using CI could be possible, but may be a lot of work. I'd say it's not worth it to invest a lot of time into it at this moment.

src/appshell/view/firstlaunchsetup/themespagemodel.cpp
```diff
@@ -96,17 +119,18 @@ QString ThemesPageModel::currentThemeCode() const
 
 void ThemesPageModel::setCurrentThemeCode(const QString& themeCode)
 {
-    if (themeCode == currentThemeCode()) {
+    if (themeCode == currentThemeCode() && !isFollowSystemTheme()) {
         return;
     }
 
+    setFollowSystemTheme(false);
```
RomanPudashkin, line 126: This logic is duplicated in two places. It should be rather at the configuration level (so that you can call
`uiConfiguration()->setCurrentTheme(...)` from any part of the program without additionally calling `setFollowSystemTheme(false)`)

cbjeukendrup: Thanks, done!

src/framework/ui/iuiconfiguration.h
```diff
@@ -49,15 +47,21 @@ class IUiConfiguration : MODULE_EXPORT_INTERFACE
...
     virtual bool isHighContrast() const = 0;
     virtual void setIsHighContrast(bool highContrast) = 0;
 
-    virtual void resetCurrentThemeToDefault(const ThemeCode& codeKey) = 0;
-
     virtual const ThemeInfo& currentTheme() const = 0;
-    virtual void setCurrentTheme(const ThemeCode& codeKey) = 0;
-    virtual void setCurrentThemeStyleValue(ThemeStyleKey key, const Val& val) = 0;
     virtual async::Notification currentThemeChanged() const = 0;
+    virtual void setCurrentTheme(const ThemeCode& codeKey, bool overrideFollowSystemTheme) = 0;
```
RomanPudashkin, line 58: I'm not sure it is a good idea to expose "overrideFollowSystemTheme" in this method. Changing the current theme and enabling the "follow system theme" option are conflicting operations

cbjeukendrup: I went for a slightly different solution (namely splitting it into "setCurrentTheme" (public) and "doSetCurrentTheme" (private).

PR #10579, opened 2022-02-15
Fix 9696: place text and icon properly

duanjr:
Resolves: [#9696](https://github.com/musescore/MuseScore/issues/9696)

Though the detail data may be different, it does look as expected.
Besides, compare to code just several lines above the change I made, it's reasonable to believe that the original code is written by carelessness. (write 'top' as 'left')

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

before change:
D6HDFR38$XA37Q%`RX02T@X
after change:
5B X{QQTU$@27@D9W~2MFKX

Issue #9696, opened 2021-11-10
[MU4 Issue] Text and icon buttons are not placed properly


jessjwilliamson:
**Describe the bug**
The text and icons have a strange placement on button and would like to point out to clean this up and match it as the Figma version.
Buttons like these are often found in the Properties pallete.

Screenshot below shows green labels - Figma how it should be. Red labels - current build.
![image](https://user-images.githubusercontent.com/71656189/141133102-4bf95639-bf3c-4d80-9192-f7381b4f9c7d.png)

**Desktop (please complete the following information):**
 - OS: Windows and Mac

Issue #9681, opened 2021-11-09
[MU4 Issue] The search button in Palettes is still 32 x 30


jessjwilliamson:
**Describe the bug**
This bug is still not being fixed, it needs to be 30x30!

**Screenshots**
![image](https://user-images.githubusercontent.com/71656189/140962152-9461ff90-0259-4173-9442-a1365e2fddc3.png)

**Desktop (please complete the following information):**
 - OS: Windows, Mac

cbjeukendrup:
Ha, a third report of this problem :) (see also #9278 and #8098)

The problem is that the size of the buttons is calculated dynamically based on the content size. This has the disadvantage that icon buttons are not square, and that text-only buttons are 29px instead of 30px high. 

So we should probably give icon-only buttons a fixed size of 30x30px. But we should also do something for text buttons.

We could
- specify a _minimum_ height of 30px for text buttons (but then it starts looking ugly for font sizes of 14+ pixels, because text buttons are then 31px)
- specify a _fixed_ height of 30px for text buttons (but then, text could get clipped at larger font sizes)
- anything else?

What would be the best solution?

RomanPudashkin:
See https://github.com/musescore/MuseScore/issues/8098

PR #10789, opened 2022-03-11
Fix the size of FlatButton

cbjeukendrup:
Resolves: #8098 (and its three duplicates but those have already been closed)
Resolves: #9681
Resolves: #9696

Closes: #8114
Closes: #10579
(sorry, but a more thorough fix for both problems was necessary)

Text-only and icon-only buttons are both 30px tall - always. And icon-only buttons will be perfectly square. For custom buttons, the size will by default be determined by the content item.

This solution also appears to work well with Qt 6. With the old solution, there were some errors printed to the console.


src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
```diff
@@ -148,76 +177,101 @@ FocusScope {
...
 
     Component {
-        id: defaultContentComponent
+        id: verticalContentComponent
+
+        ColumnLayout {
+            spacing: 4
+
+            StyledIconLabel {
+                Layout.alignment: Qt.AlignCenter
```
Eism, line 191: should be Qt.AlignHCenter

Eism (approved):

jessjwilliamson:
@DmitryArefiev @Eism I confirm this is good can this be merged into master please? :)

Issue #10756, opened 2022-03-06
[MU4 Issue] An octave higher when playing the note C-flat


Fince10086:
**Describe the bug**
MU4 has an octave higher when playing the note C-flat.

**Screenshots**
![image](https://user-images.githubusercontent.com/87692117/156922561-81c86244-d379-42fd-a48a-935d68127ce5.png)

**Platform information**
Windows and Linux (not tested on macOS)

cbjeukendrup:
There are more problems with notes with accidentals... 
- F## sounds like Eb
- Cb, Cbb, A## sound an octave to high
- B#, B##, Db, Dbb sound an octave to low

vimpostor:
I have fixed this in #10798 :)

PR #10798, opened 2022-03-13
[MU4] Fix broken pitch classes

vimpostor:
Resolves: #10756

- The first commit fixes the actual issue. The `actualOctave()` function was shifting in the wrong direction, but actually we had to reverse apply the accidental (not apply again like before this patch) to get back to the original note.
- The second commit fixes a wrong pitch class mapping

- [ ] I created the test (mtest, vtest, script test) to verify the changes I made (Does this apply here?)


vimpostor:
Attached is a score to test the behaviour. With my patch every onbeat note has the same pitch as the following offbeat note (as should be the case). Before my patch, there were different kind of problems, e.g. wrong octave or flat out wrong notes.
[10756-pitchclasses.zip](https://github.com/musescore/MuseScore/files/8239551/10756-pitchclasses.zip)

vpereverzev (approved):

vpereverzev:
nice catch. Thanks

PR #10808, opened 2022-03-14
Fix #10794 - ensure string sections are in "common" group

### LINKED ISSUE #10794: Add the 'section' strings to the 'common' category
**Describe the bug**
Currently, in the '**common**' category of the add instruments / new score dialog, we only show solo viol instruments. We shoudl also add the sections here too and rename them slightly.

To avoid confusing english speakers, who might not easily notice the difference between singular and plural, I think we should use the following naming convention:

- Violin
- Violins (section)
- Viola
- Violas (section)
- Violoncello
- Violoncellos (section)
- Contrabass
- Contrabasses (section) 

image

I considered adding 'solo' to the end of the solo instruments but felt it may get confused with the 'make soloist' button.

@shoogle - any thoughts?
@wizofaus - another simple enough one, if you have the time :)

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/10794)

String sections in common group

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -13210,7 +13210,7 @@
             </Instrument>
             <Instrument id="violins">
                   <family>orchestral-strings</family>
-                  <trackName>Violins</trackName>
+                  <trackName>Violins (section)</trackName>
                   <longName>Violins</longName>
                   <shortName>Vlns.</shortName>
                   <description>Violin section.</description>
@@ -13234,6 +13234,7 @@
                         <controller ctrl="32" value="21"/> <!--Bank LSB-->
                         <program value="44"/> <!--Tremolo Strings-->
                   </Channel>
+                  <genre>common</genre>
                   <genre>orchestra</genre>
             </Instrument>
             <Instrument id="viola">
@@ -13265,7 +13266,7 @@
             </Instrument>
             <Instrument id="violas">
                   <family>orchestral-strings</family>
-                  <trackName>Violas</trackName>
+                  <trackName>Violas (section)</trackName>
                   <longName>Violas</longName>
                   <shortName>Vlas.</shortName>
                   <description>Viola section.</description>
@@ -13289,6 +13290,7 @@
                         <controller ctrl="32" value="31"/> <!--Bank LSB-->
                         <program value="44"/> <!--Tremolo Strings-->
                   </Channel>
+                  <genre>common</genre>
                   <genre>orchestra</genre>
             </Instrument>
             <Instrument id="violoncello">
@@ -13321,7 +13323,7 @@
             </Instrument>
             <Instrument id="violoncellos">
                   <family>orchestral-strings</family>
-                  <trackName>Violoncellos</trackName>
+                  <trackName>Violoncellos (section)</trackName>
                   <longName>Violoncellos</longName>
                   <shortName>Vcs.</shortName>
                   <description>Violoncello section.</description>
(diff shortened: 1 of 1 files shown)


shoogle (changes requested):
Sorry I didn't see this earlier, but this change has to be made in the [online spreadsheet](https://docs.google.com/spreadsheets/d/1SwqZb8lq5rfv5regPSA10drWjUAoi65EuMoYtG-4k5s/edit#gid=516529997) first rather than directly in instruments.xml. This [Python script](https://github.com/musescore/MuseScore/blob/master/share/instruments/update_instruments_xml.py) must be run to download the spreadsheet and generate a new instruments.xml.

I have now updated the spreadsheet, so if you run the Python script `share/instruments/update_instruments_xml.py` then it will update instruments.xml for you. It will look very similar to the one you created manually, but `common` will go before `orchestral` in the list of genres for consistency with the rest of the file.

If you have trouble running the Python script then you could move the genres manually on this occasion and let me run the Python script in future.

wizofaus:
Ah, yes, I hadn't noticed it said it was generated at the top. No luck running the py script though:

UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 5162: character maps to <undefined>

Not sure what that's about.
Will just adjust manually for now.  Shouldn't the spreadsheet be in git if it's the source of truth?

wizofaus:
`python -X utf8 share\instruments\instruments.xml` did the trick

Jojo-Schmitz:
At least for the other 2 .py files  and on Windows, a `set PYTHONIOENCODING=utf-8`  is sufficient (and that is documented on those files)

shoogle (approved):

shoogle:
Thanks for this!

> set PYTHONIOENCODING=utf-8 is sufficient... Apparently doesn't work in this case though...

Interesting. [PEP 540](https://peps.python.org/pep-0540/) has more suggestions (e.g. `set PYTHONUTF8=1`, `unset PYTHONLEGACYWINDOWSFSENCODING`).

The scripts work fine for me on macOS but I ran into problems running them on Linux because the version of Python available in Ubuntu 20.04's default repositories is outdated. macOS has Python 3.9 via Homebrew but Ubuntu 20.04 only has Python 3.8.

> Shouldn't the spreadsheet be in git if it's the source of truth?

It depends whether you see the instrument data as part of MuseScore's source code, or if you see it as an asset/resource that MuseScore interacts with at runtime via an API. I'd argue it's the latter, so it's ok to store it in a separate "repository" and update it here occasionally like we do with the other assets and resources (e.g. images, fonts, SoundFonts).

Git isn't really designed for dealing with diffs of tabular data. It works, but it will show an entire row as changed rather than just an individual cell, and it won't show the column heading. Google Sheets has a version history that shows exactly what changed and who changed it, and it has the ability to restore previous versions. If somebody needs to make changes then I can give that person permission to edit only the columns that they actually need to edit rather than the entire sheet. Plus it's more convenient than an offline spreadsheet because you can use it in the browser (no need to install other software) and changes are synced instantly so there is no possibility of merge conflicts.

By the way, the W3C Music Notation Community Group (i.e. the people who create MusicXML, SMuFL, and MNX) are starting an effort to create a shared resource of instrument data that can be used by all notation software. This will probably take the form of an SQL database on a website somewhere that anybody can ask to contribute to, which is basically a glorified online spreadsheet.

shoogle:
If you want to see the history of changes in the spreadsheet then you can just look at the [GitHub history for instruments.xml](https://github.com/musescore/MuseScore/commits/master/share/instruments/instruments.xml). I've been careful to explain the changes in commit messages each time, though I don't think it's really necessary to say more than "Updated instrument data" like we do when we update fonts and other resources.

Issue #10420, opened 2022-02-01
Choose Instruments Panel: Search categories are often incorrect.


Tantacrul:
**Describe the bug**
When searching for an instrument, the categories shown on the left can often be wrong. For example, when searching for 'Snare', I became confused because I wanted the 'Marching Band' type. When I checked the categories on the left, they did not highlight the 'Percussion - Marching' category.

![image](https://user-images.githubusercontent.com/21022311/151946753-2fc850b5-3edc-4628-b4f8-846781219007.png)

The next snare down, also has this problem:
![image](https://user-images.githubusercontent.com/21022311/151946841-3074fa1b-e32b-4e9f-aace-7a82b90f8122.png)

I also notice that the category dropdown still say 'Common', where it should say (while the user is searching) 'All instruments'. 

Here's another example:
![image](https://user-images.githubusercontent.com/21022311/151947189-146422af-95be-46f3-be7e-9863160423fe.png)

And one more:
![image](https://user-images.githubusercontent.com/21022311/151947292-eb15d0c0-50dc-4c67-818f-4462032941ce.png)

Tantacrul:
The reason I'm marking this as a P2 (and not a P3) is because it causes confusion and reduces communication, which makes it harder to select the correct instrument.

In MS3, this handy thing existed, which made your instrument choice very simple to understand. Without this bug being fixed, MS4 is a bit more obscure.
![image](https://user-images.githubusercontent.com/21022311/151949013-cae3675d-1597-4df4-b043-e02d437fb046.png)

oktophonie:
The marching instruments are awkward as they have similar names to other instruments but different instrument definitions (different labels, numbers of lines, etc.)

Fortunately there are only four of them. How about we change the _score labels_ to:
- Snares / S.
- Tenors / T.
- Toms / Tom
- Basses / B.

And their names in the instrument panel to the same, perhaps with "(marching)" suffixed? (Since "Tenors" and "Basses" remain ambiguous).

We should remove "Concert" from the instrument labels for the orchestral Snare Drum, Bass Drum, Tom-toms etc. too (but they could remain in the name used in the instrument panel, for extra clarity).

shoogle:
To be clear, as you type characters in the search box, no category is highlighted on the left. This is the **correct** behaviour (because the search results can include instruments from multiple categories).

![image](https://user-images.githubusercontent.com/11011881/152075038-f5808e72-6df4-4a7b-9c39-cdd1b17c843c.png)

[...]

Issue #9131, opened 2021-09-15
[MU4 Issue] Order of templates categories is incorrect in MacOS


bkunda:
Three issues here:
1. The list of categories in the "Choose from template" dialog is out of order in MacOS.
2. Please rename the category "templates" to "My Templates" (note capitalisation) and place at the bottom of the list.
3. Please rename the tab item "Choose from template" to "Create from template"

Current list:
<img width="1273" alt="Screenshot 2021-09-15 at 20 42 15" src="https://user-images.githubusercontent.com/86290556/133419417-b8ce9725-569f-402a-b31c-ad8c2ad4639f.png">

**Expected behavior**

1. Order of categories should be as per numbered list here: https://github.com/musescore/MuseScore/tree/master/share/templates
2. "templates" should be renamed to "My Templates" and appear at the bottom of the list
3. "Choose from template" tab label should read "Create from template"
MacBook Air M1, MacOS 11.5.2

Jojo-Schmitz:
and "template" should get renamed to "Templates", "Choose from template" to "Choose from templates"
"My templates might better shown at the top, or just after General, like in MuseScore 3

Jojo-Schmitz:
Not sure about  "My Templates", the capitalization, didn't we just recently decided to no longer do that "Title Case" thing but use "Sentence case" all the way?

bkunda:
I suggest "My Templates" because it apparently stores user-created templates. In this case, "templates" on its own does not communicate this. 

You are right about sentence case, and I'm usually a stickler for consistency in this kind of thing. In this dialog, however, I don't actually mind the use of title case (E.g. "Voice and piano" looks somewhat odd by comparison). But I'm happy to be convinced otherwise on this one. 

"Choose from templates" (in the plural) is technically correct, but I don't think "choose from template" (in the singular) is necessarily _incorrect_. In this case, the user is choosing to create a new score from an existing template (just one, not many). They are of course selecting to do this from a list of many templates. So both options are probably acceptable.

Jojo-Schmitz:
"Choose template", not "Choose **from** template"

bkunda:
I take your point, but actually I think the UI copy for these tab items actually works well the way it is. Having three words instead of two makes for a nice visual distinction between the two tabs. My preference would be to leave it and to focus on addressing the category ordering.

Jojo-Schmitz:
Then it is plural...
Native speakers to the front?

wizofaus:
[...]

PR #10828, opened 2022-03-16
[MU4] Choose Instruments Panel: Search categories are often incorrect.

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10420
Resolves: https://github.com/musescore/MuseScore/issues/9131


DmitryArefiev:
Tested #10420 #9131 on Win10, Mac12, LinuxMint 20.2 - FIXED

bkunda:
This is looking good! A couple things I've just noticed, which might be the subject of separate issues. @oktophonie you're opinions on the below would be helpful :-) 

- Voice + Piano is in "Choral". I would expect to see this in "Chamber", or even "Solo" (depending on the musical genre). But definitely not choral. 

- Regarding the user-defined templates category: Will we only be able to recognise templates created in MS4? I had one already stored in my designated "templates" folder, which I believe was created in MS3, but it on its own did not trigger the "My templates" category in the "New score" dialog (I.e. when I _only_ had .mscz files from MS3 in my templates folder, the "My templates" category failed to show in MS4). 

On the last point, we might just need to add some documentation around this in the handbook if it is the case that only template files created in MS4 will be visible in the New Score dialog (and I could understand why this might have to be the case from an engraving POV).

Eism (approved):

PR #10833, opened 2022-03-17
Fix #10780 - irrelevant staff types shown for all instruments

### LINKED ISSUE #10780: Too many 'Stave Types' displayed in the Instruments Panel compared to MS 3.6
**Describe the bug**
See video for the full explanation. 

https://user-images.githubusercontent.com/21022311/157869299-c7763456-7895-4059-a422-fdfc497caacd.mov

1. Step one would be to hide any options not displayed in MuseScore 3.6
2 Step two is to also remove any other redundant options (which I'm not 100% sure about yet. There may be reasons there is a 'Standard' and a 'Custom Standard' staff definition for each instrument but I'm not sure what it is.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/10780)

Full list of all staff types was being shown for all instruments, rather than those matching the instrument's staff type group.

### DIFF
--- a/src/instrumentsscene/view/staffsettingsmodel.cpp
+++ b/src/instrumentsscene/view/staffsettingsmodel.cpp
@@ -44,6 +44,13 @@ void StaffSettingsModel::load(const QString& staffId)
     m_config = notationParts()->staffConfig(m_staffId);
 
     m_type = staff->staffType()->type();
+    auto presets = Ms::StaffType::presets();
+    auto match = std::find_if(presets.begin(), presets.end(), [staff](const Ms::StaffType& t) {
+        return t.isSameStructure(*staff->staffType());
+    });
+    if (match != presets.end()) {
+        m_type = match->type();
+    }
 
     m_voicesVisibility.clear();
     for (const QVariant& voice: staff->visibilityVoices()) {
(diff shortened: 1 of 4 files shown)


src/instrumentsscene/view/staffsettingsmodel.cpp
```diff
@@ -53,27 +53,35 @@ void StaffSettingsModel::load(const QString& staffId)
...
-    for (notation::StaffType type: notation::allStaffTypes()) {
-        QVariantMap obj;
+    for (const Ms::StaffType& type : Ms::StaffType::presets()) {
+        if (type.group() == group || type.type() == m_type) {
+            QVariantMap obj;
 
-        obj["text"] = staffTypeToString(type);
-        obj["value"] = static_cast<int>(type);
+            obj["text"] = staffTypeToString(type.type());
+            obj["value"] = static_cast<int>(type.type());
```
Jojo-Schmitz, line 73: whitespace damage ;-)

src/instrumentsscene/view/staffsettingsmodel.h
```diff
@@ -35,24 +35,25 @@ class StaffSettingsModel : public QObject
 
     INJECT(instruments, context::IGlobalContext, context)
 
-    Q_PROPERTY(QString staffType READ staffType NOTIFY staffTypeChanged)
+    Q_PROPERTY(int staffType READ staffType NOTIFY staffTypeChanged)
     Q_PROPERTY(QVariantList voices READ voices NOTIFY voicesChanged)
     Q_PROPERTY(bool isSmallStaff READ isSmallStaff NOTIFY isSmallStaffChanged)
     Q_PROPERTY(bool cutawayEnabled READ cutawayEnabled NOTIFY cutawayEnabledChanged)
-
+
```
Jojo-Schmitz, line 42: whitespace damage

src/instrumentsscene/view/staffsettingsmodel.cpp
```diff
@@ -53,27 +60,44 @@ void StaffSettingsModel::load(const QString& staffId)
...
 QVariantList StaffSettingsModel::allStaffTypes() const
 {
     QVariantList result;
+    auto staff = notationParts()->staff(m_staffId);
+    if (staff == nullptr) {
+        return result;
+    }
+    int maxLines = 0;
+    bool isPercussion = false;
+    if (staff->part() && staff->part()->instrument()) {
```
RomanPudashkin, line 76: This code can be greatly simplified by adding temporary variables and reusing them instead of staff->part()->instrument()->...

src/instrumentsscene/view/staffsettingsmodel.cpp
```diff
@@ -53,27 +60,44 @@ void StaffSettingsModel::load(const QString& staffId)
...
+            maxLines = staff->part()->instrument()->stringData()->frettedStrings();
+        }
+        isPercussion = staff->part()->instrument()->useDrumset();
+    }
 
-    for (notation::StaffType type: notation::allStaffTypes()) {
-        QVariantMap obj;
+    for (const Ms::StaffType& type : Ms::StaffType::presets()) {
+        if (isPercussion ? type.group() == Ms::StaffGroup::PERCUSSION
+            : type.group() == Ms::StaffGroup::STANDARD || type.group() == Ms::StaffGroup::TAB && type.lines() <= maxLines) {
```
RomanPudashkin, line 85: This condition looks too complicated to read/debug. You should at least move it to temporary variables

wizofaus:
I don't seem to be able to reply to the comments made above. I disagree that `if (staff->part() && staff->part()->instrument())` can be greatly simplified, the addition of a temporary variable is not going to significantly improve performance or readable here.  The second example I agree is a little complex at first glance, but temporary variables isn't really going to help, the nature of the logic basically:

if this is a percussion instrument, then the preset type group must be percussion, else the preset type group can either be standard, or it can be TAB provided that the number of tab lines is less than that of the instrument.
I suppose it could be rewritten as 
```
auto include = false;
if ( type.group() == Ms::StaffGroup::PERCUSSION)
  include = isPercussion;
else if (type.group() == Ms::StaffGroup::STANDARD)
  include = true;
else if (type.group() == Ms::StaffGroup::TAB)
  include = type.lines() <= maxLines;
if (include) {
```

but I vastly prefer the functional simplicity of the original expression
.  If you want to rewrite it in a way that makes more sense to you, I don't mind.

RomanPudashkin:
Thank you so much for the fix! I created a new PR in which I fixed all the code-review issues

wizofaus:
Except it didn't from what I just tested, it appears to still allow selecting all possible staff types rather than filtering the list to what makes sense for the given instrument. @RomanPudashkin

Issue #10259, opened 2022-01-13
[MU4 Issue] Properties in drop-downs are not updated when changing object selection 


bkunda:
**Describe the bug**
When changing an object's property in a dropdown menu in **Properties**, the value in said dropdown menu does not change when a different object is selected – unless the selection is cleared first.   

This was first noticed when changing the font face and/or size of a text object: 

![GIF 1-13-22 at 21 03 50](https://user-images.githubusercontent.com/86290556/149309851-26e0c040-b3ce-4136-98cb-e812b8ba6dcd.gif)

@boryskuzmenko confirms this problem is general across all possible score elements (including articulations, ornaments, dynamics, etc.). 

**Expected behavior**
Selecting any score object should cause all relevant dropdown menus in Properties to update with the correct values for that selection.

**Desktop (please complete the following information):**
 - MacOS 11.6 (M1)
 - Windows 11

**Additional context**
Related to https://github.com/musescore/MuseScore/issues/8813
Note also: new design for text size dropdown in https://github.com/musescore/MuseScore/issues/9769

maryszabo:
Hi, is this still available? If so, I am new to this (and first time working on open source code). I would really appreciate any tips or advice on how to approach solving this problem.

RobFog:
> Hi, is this still available? If so, I am new to this (and first time working on open source code). I would really appreciate any tips or advice on how to approach solving this problem.

Looks like it is. Maybe @cbjeukendrup can help? :)

cbjeukendrup:
Sure!

@maryszabo Hi! I investigated a bit into this issue myself, and after some testing I found that it seems to be specific to dropdown items, rather than having to do anything with text specifically. For example, the inspector panel for key signatures also contains a dropdown, which shows the same problem.

I also discovered that it only happens when you have just changed the value in the dropdown. So, if you select one item, don't change any settings, and then directly select another item, it updates fine; but if you select one item, then change the value of a dropdown, and then select another item, the value of the dropdown will not update anymore, until you clear the selection (because then the dropdown is destroyed altogether and will be recreated when you make a new selection).

So, to fix this issue, I would suggest to look the file DropdownPropertyView.qml. If you add some logging statements like this, you can see in the console what happens. What I see there when doing so, is not completely expected.

[...]

PR #10846, opened 2022-03-18
Update Font Text And Size when changing the the focus element after selecting an option

iamjithinjohn:
Resolves: #10259

Modify activated() signal to trigger whenever currentIndex or when the user selects an option from the dropdown. Pass the index values as per the context to update the model and the dropdown in order. Remove OnCurrentValueChanged function.


RomanPudashkin:
Thank you for your help! I have prepared a more global fix for this problem:
https://github.com/musescore/MuseScore/pull/10862

It is based on the idea of @cbjeukendrup:
https://github.com/musescore/MuseScore/issues/10259#issuecomment-1022674028

PR #8892, opened 2021-08-20
[MU4] Removed old TODO

RomanPudashkin:

PR #10849, opened 2022-03-20
Fix AppleClang 13.1 compiler warnings

cbjeukendrup:
The update to Xcode 13.3 includes AppleClang 13.1, which gives some new compiler warnings (actually, a lot of them, almost 600, but only a few _unique_ ones).

- It got stricter about the "rule of three": even if you declare a copy constructor or copy assignment operator with `= default` or `= delete`, it treats it as "user declared", and wants you to explicitly declare the other too. In some cases, I just removed the `= default` declaration, because it's redundant, and sometimes, I added a copy constructor with `= default`.

- It got smarter about "unused variables": it now warns about variables that are assigned to but never read (before, these warnings only appeared in "deep analysis" mode). This reveals some strange situations, and I was not always sure how to handle them. You'd say that someone must have added those variables with a reason, but it's not easy to track down which reason (Git blame often shows the famous "Changed code style" commit, or even worse, "Initial commit").

### DIFF
--- a/src/importexport/musedata/internal/musedata.cpp
+++ b/src/importexport/musedata/internal/musedata.cpp
@@ -44,6 +44,8 @@
 #include "libmscore/timesig.h"
 #include "libmscore/segment.h"
 
+#include "log.h"
+
 using namespace mu::engraving;
 
 namespace Ms {
@@ -53,7 +55,7 @@ namespace Ms {
 
 void MuseData::musicalAttribute(QString s, Part* part)
 {
-    QStringList al = s.mid(3).split(" ", Qt::SkipEmptyParts);
+    QStringList al = s.mid(2).split(" ", Qt::SkipEmptyParts);
     foreach (QString item, al) {
         if (item.startsWith("K:")) {
             int key = item.midRef(2).toInt();
@@ -67,23 +69,22 @@ void MuseData::musicalAttribute(QString s, Part* part)
         } else if (item.startsWith("T:")) {
             QStringList tl = item.mid(2).split("/");
             if (tl.size() != 2) {
-                qDebug("bad time sig <%s>", qPrintable(item));
+                LOGD() << "bad time sig: " << item;
                 continue;
             }
             int z = tl[0].toInt();
             int n = tl[1].toInt();
             if ((z > 0) && (n > 0)) {
-//TODO                        score->sigmap()->add(curTick, Fraction(z, n));
                 Measure* mes = score->tick2measure(curTick);
                 Segment* seg = mes->getSegment(SegmentType::TimeSig, curTick);
                 TimeSig* ts = Factory::createTimeSig(seg);
                 Staff* staff = part->staff(0);
                 ts->setTrack(staff->idx() * VOICES);
+                ts->setSig(Fraction(z, n));
                 seg->add(ts);
             }
         } else if (item.startsWith("X:")) {
         } else if (item[0] == 'C') {
-#if 0 // TODO
             int staffIdx = 1;
             int col = 2;
             if (item[1].isDigit()) {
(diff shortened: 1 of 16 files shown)


src/importexport/musedata/internal/musedata.cpp
```diff
@@ -83,28 +83,30 @@ void MuseData::musicalAttribute(QString s, Part* part)
             }
         } else if (item.startsWith("X:")) {
         } else if (item[0] == 'C') {
+            /* TODO
```
Jojo-Schmitz, line 86: I'd prefer an `#if 0 ... #endif` for disabling a longer code section

src/importexport/musicxml/internal/musicxml/exportxml.cpp
```diff
@@ -1207,8 +1207,10 @@ void ExportMusicXml::calcDivisions()
                 // special number 0 -> don’t show staff number in
                 // xml output (because there is only one staff)
 
+                /*
                 int sstaff = (staves > 1) ? st - strack + VOICES : 0;
                 sstaff /= VOICES;
+                */
```
Jojo-Schmitz, line 1213: Here we could as well just remove those lines?
@lvinken ?

src/notation/internal/notationparts.cpp
```diff
@@ -1044,7 +1044,7 @@ void NotationParts::sortParts(const PartInstrumentList& parts, const QList<Ms::S
     QList<int> staffMapping;
     QList<int> trackMapping;
     int runningStaffIndex = 0;
-    bool sortingNeeded = false;
+    //bool sortingNeeded = false;
```
Jojo-Schmitz, line 1047: here too

src/notation/internal/notationparts.cpp
```diff
@@ -1055,7 +1055,7 @@ void NotationParts::sortParts(const PartInstrumentList& parts, const QList<Ms::S
 
             trackMapping.append(originalStaves.indexOf(staff));
             staffMapping.append(actualStaffIndex);
-            sortingNeeded |= actualStaffIndex != runningStaffIndex;
+            //sortingNeeded |= actualStaffIndex != runningStaffIndex;
```
Jojo-Schmitz, line 1058: here too

thirdparty/beatroot/Induction.cpp
```diff
@@ -160,7 +160,7 @@ AgentList Induction::beatInduction(const AgentParameters &params,
             b = bestn[index];
                               // Adjust it, using the size of super- and sub-intervals
             double newSum = clusterMean[b] * clusterScore[b];
-            int newCount = clusterSize[b];
+            //int newCount = clusterSize[b];
```
Jojo-Schmitz, line 163: Maybe here too, but I'm not sure how to deal with third party code

src/importexport/musicxml/internal/musicxml/exportxml.cpp
lvinken: Yes, of course. Text search shows variable sstaff is not used in that function. Probably left in place by accident after copying some code.

cbjeukendrup:
I decided to remove those unused variables entirely, which makes things a lot cleaner. On the other hand, it hides traces of potential mistakes.

src/importexport/musedata/internal/musedata.cpp
```diff
@@ -83,28 +83,35 @@ void MuseData::musicalAttribute(QString s, Part* part)
             }
         } else if (item.startsWith("X:")) {
         } else if (item[0] == 'C') {
+#if 0 // TODO
```
RomanPudashkin, line 86: I don't like this way of fixing warnings. We should either remove this code completely or fix it. My experience tells me that such TODOs can exist for years without any attempt to fix them

https://github.com/musescore/MuseScore/pull/8892

src/notation/internal/notationparts.cpp
```diff
@@ -1055,15 +1054,12 @@ void NotationParts::sortParts(const PartInstrumentList& parts, const QList<Ms::S
 
             trackMapping.append(originalStaves.indexOf(staff));
             staffMapping.append(actualStaffIndex);
-            sortingNeeded |= actualStaffIndex != runningStaffIndex;
             ++runningStaffIndex;
         }
         ++partIndex;
     }
 
-    //if (sortingNeeded) {
```
RomanPudashkin, line 1064: I think this is a wonderful opportunity to ask @asattely why he disabled this code, and @njvdberg why this code is needed :)

src/importexport/musedata/internal/musedata.cpp
Jojo-Schmitz: Removing is fine by me. That code had not been touched in 10 years, seems a left over tehm the 1.x code moved to Github, to become MuseScore 2, the initial commit on Github. So really old and stale...

cbjeukendrup: I decided to choose not the easiest solution, but hopefully the most correct one, namely fixing the TODO right now. I've no idea what MuseData even is, but found the spec on the internet and tried to understand the code, and corrected it as far as I could. However, the MuseData module is probably the most broken of all; it does not have any unit tests and when I tried to import a MuseData file the result was corrupted in every measure and couldn't even be opened.

Jojo-Schmitz: Do you have any MuseData sample files?

RomanPudashkin (approved):

src/importexport/musedata/internal/musedata.cpp
cbjeukendrup: I grabbed this one: https://bitbucket.org/musedata/bach/src/master/bg/cant/0001/stage2/01/01 
That looks like a MuseData file.

Jojo-Schmitz: Thanks. It does open fine in MuseScore 1.3, saving as mscz and opening that in 3.6.2 reports only one single corruption, but opening the .md in 2.3.2 or 3.6.2 reports tons of corruptions (as does my backport of this chenges here to 3.x, that does though detect the correct timesig, 1.3 does too). It opens them though when ignoring these,

src/notation/internal/notationparts.cpp
njvdberg: @RomanPudashkin Needed is maybe too much but this was just an optimization to prevent an unnecessary instance on the undo stack. Apparently the implementation of system objects might require always sorting but that's what @asattely can comment on.

Issue #10823, opened 2022-03-16
[MU4 Issue] Crash when deleting instruments in some scores


DmitryArefiev:
**To Reproduce**
1. Open  [7.zip](https://github.com/musescore/MuseScore/files/8261694/7.zip) or [8.zip](https://github.com/musescore/MuseScore/files/8261697/8.zip)
2. Open Instruments window and delete instrument

https://user-images.githubusercontent.com/90187801/158581229-4f4b3264-5975-4c79-8e5d-9d7a00e36a97.mp4

**Platform information**
macOS, Windows, Linux

DmitryArefiev:
@asattely Hi! I've created them in latest MU4. You need to add a tempo mark at first and then add a new staff on top of the score

https://user-images.githubusercontent.com/90187801/159022988-a947dab0-53f6-4d96-8969-a6ab845a2be4.mp4

DmitryArefiev:
Fixed in #10855

PR #10855, opened 2022-03-21
[MU4] Fix incorrect placement of system objects and add checking on load

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/10823

The issue was that there was a bug in the system object code that didn't produce any graphical errors (so the vtests still passed) but saved the system object to the incorrect staff in XML. That's been fixed now, and also some code has been added to the loading process to make sure that if we have invalid system objects in the xml the program won't crash anymore.


DmitryArefiev:
Tested #10823 on Win10, Mac12, LinuxMint 20.2 - FIXED      

Please merge into master

PR #8199, opened 2021-05-29
[MU3] Fix #321730: Crash on corrupt MXL import

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/321730

PR for master in #8200

PR #10860, opened 2022-03-22
[MU4] Fix #330595: Crash on selecting entire score after XML import

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/330595 (the 2nd part, the first part is not an issue with master, looking at its fix for 3.x seems to indicate that this is because #8530 hasn't been ported to master yet? For 3.x and master that crash had been fixed in #8199 resp. #8200, we'd need to make sure it doesn't break again when porting those backend fixes to master)
Applies to 3.x too.
Strange enough: There is no crash when saving the imported XML as an MSCZ, closing and reopening it and doing that "select all" then.

### DIFF
--- a/src/engraving/libmscore/select.cpp
+++ b/src/engraving/libmscore/select.cpp
@@ -508,7 +508,7 @@ void Selection::appendChord(Chord* chord)
             if (note->tieFor()->endElement()->isNote()) {
                 Note* endNote = toNote(note->tieFor()->endElement());
                 Segment* s = endNote->chord()->segment();
-                if (s->tick() < tickEnd()) {
+                if (!s || s->tick() < tickEnd()) {
                     _el.push_back(note->tieFor());
                 }
             }
@@ -517,7 +517,7 @@ void Selection::appendChord(Chord* chord)
             if (sp->endElement()->isNote()) {
                 Note* endNote = toNote(sp->endElement());
                 Segment* s = endNote->chord()->segment();
-                if (s->tick() < tickEnd()) {
+                if (!s || s->tick() < tickEnd()) {
                     _el.push_back(sp);
                 }
             }


src/engraving/libmscore/select.cpp
```diff
@@ -527,7 +527,7 @@ void Selection::appendChord(Chord* chord)
             if (sp->endElement()->isNote()) {
                 Note* endNote = toNote(sp->endElement());
                 Segment* s = endNote->chord()->segment();
-                if (s->tick() < tickEnd()) {
+                if (s && s->tick() < tickEnd()) {
```
Jojo-Schmitz, line 530: This particular crash happens here

src/engraving/libmscore/select.cpp
```diff
@@ -518,7 +518,7 @@ void Selection::appendChord(Chord* chord)
             if (note->tieFor()->endElement()->isNote()) {
                 Note* endNote = toNote(note->tieFor()->endElement());
                 Segment* s = endNote->chord()->segment();
-                if (s->tick() < tickEnd()) {
+                if (s && s->tick() < tickEnd()) {
```
Jojo-Schmitz, line 521: This change is just for extra safety.

src/engraving/libmscore/select.cpp
Jojo-Schmitz: Alternative fix: `if (!s || s->tick() < tickEnd()) {`, avopids the crash too. Which would be more correct?

src/engraving/libmscore/select.cpp
```diff
@@ -507,7 +507,7 @@ void Selection::appendChord(Chord* chord)
         if (note->accidental()) {
             _el.append(note->accidental());
         }
-        foreach (EngravingItem* el, note->el()) {
+        for (EngravingItem* el: note->el()) {
```
Jojo-Schmitz, line 510: (while being in the vincinity)

src/engraving/libmscore/select.cpp
wizofaus: I'd be more worried why a chord would ever not have a segment?

Jojo-Schmitz: Well, then check the sample score from that forum post.

Jojo-Schmitz: Yes, this is curing the symptom rather than the cause. But that isn't a bad thing, or is it?

Jojo-Schmitz: It is only happening directly after the XML import, not anymore once the score has been saved as mscz (and a close/reopen).

wizofaus: Well the issue is as you noted it's not clear what the right behaviour is in this case if a chord doesn't have a segment, and it seems like the sort of thing that might have consequences elsewhere.  I'd be trying to fix it in the XML importer first, that's all.
I'd also note that MS codebase, and even the short sample above, would probably be twice as big if it had to check every pointer first, due to the lack of a null propagation operator in C++.

Jojo-Schmitz: Don't check for error conditions you can't handle... but do check those you can.

I've now tried both methods on the xml from the forum post: the resulting scores are 100% identical.

Jojo-Schmitz: But yes, of course it'd be better to fix the XML import, I just don't know where to look. @lvinken ?

lvinken: Indeed, MusicXML import should not result in invalid data. Will try to find the root cause.

lvinken: See comment added to https://musescore.org/en/node/330595 for possible root cause, shouldn't be too hard to track down and fix in a structural way.

Jojo-Schmitz:
This one too is waiting to get reviewed merged since almost a year now.

PR #10873, opened 2022-03-23
Bring back some engraving debug painting tools from MU3

cbjeukendrup:
Can be enabled by setting the `ENGRAVING_PAINT_DEBUGGER_ENABLED` CMake option to `ON` _and_ setting `MScore::showSkylines` to `true`; there's no UI for it.

@mike-spa Please try if this is helpful for you, or whether I can improve the visual representation.

Schermafbeelding 2022-03-23 om 15 55 53


src/engraving/paint/debugpaint.cpp
```diff
@@ -80,3 +83,95 @@ void DebugPaint::paintElementsDebug(mu::draw::Painter& painter, const QList<Ms::
...
+#ifndef NDEBUG
+    if (MScore::showCorruptedMeasures) {
+        painter.setPen(Pen(Color::redColor, 4.0));
+        painter.setBrush(BrushStyle::NoBrush);
+
+        double _spatium = score->spatium();
+
+        for (const System* system : page->systems()) {
+            for (const MeasureBase* mb : system->measures()) {
+                if (mb->type() == ElementType::MEASURE) {
```
RomanPudashkin, line 163: Please use inverted conditions to avoid multi-level nested code:

```
if (mb->type() != ElementType::MEASURE) {
    continue;
}
```
...

RomanPudashkin, line 163: Please use inverted conditions to avoid multi-level nested code:

```
if (mb->type() != ElementType::MEASURE) {
    continue;
}
```
...

src/engraving/libmscore/skyline.h
```diff
@@ -26,6 +26,7 @@
 #include <vector>
 
 #include "infrastructure/draw/geometry.h"
+#include "infrastructure/draw/painter.h"
```
RomanPudashkin, line 29: No need to include this header. You should use forward declaration instead

src/engraving/paint/debugpaint.h
```diff
@@ -37,10 +38,13 @@ class PaintDebugger;
 class DebugPaint
 {
     INJECT_STATIC(engraving, diagnostics::IEngravingElementsProvider, elementsProvider)
-public:
+    INJECT_STATIC(engraving, ui::IUiConfiguration, uiConfiguration)
```
RomanPudashkin, line 41: Please, use EngravingConfiguration instead

src/engraving/paint/debugpaint.cpp
```diff
@@ -80,3 +83,95 @@ void DebugPaint::paintElementsDebug(mu::draw::Painter& painter, const QList<Ms::
...
+                            s->shapes().at(i).paint(painter);
+                            painter.translate(-pt);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+#ifndef NDEBUG
```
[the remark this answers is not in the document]
cbjeukendrup: The problem is that `Measure::corrupted()` and everything supporting it is hidden behind those `#ifndef NDEBUG` things. Should I remove them, so that those methods are also available in release builds?

src/engraving/paint/paint.h
```diff
@@ -26,17 +26,21 @@
...
+
 namespace Ms {
 class EngravingItem;
 }
 
 namespace mu::engraving {
 class Paint
 {
-public:
+    INJECT_STATIC(engraving, ui::IUiConfiguration, uiConfiguration)
```
RomanPudashkin, line 39: It seems that uiConfiguration is not used in this class

RomanPudashkin (approved):

PR #10874, opened 2022-03-23
[MU4] Fix #10477: Mixer and timeline should be tabbed by default, and height adjustment interaction needs fixing

### LINKED ISSUE #10477: [MU4 Issue] Mixer and timeline should be tabbed by default, and height adjustment interaction needs fixing
**Describe the bug**
Opening the timeline before the mixer stacks the two panels rather than tabbing them together. 
There is also some glitchy interaction when resizing these panels vertically. 

See this video:
(N.B: In the video, when I say "docked" by default, I actually mean "tabbed". Docking is already working as default behaviour). 
https://www.dropbox.com/s/fx3jlv5ny2ewfwz/mixer-timeline-docking-issue.mov?dl=0

**Expected behavior**
1. Mixer and timeline should be tabbed by default when both are opened (regardless of the order in which they are opened)
2. Vertical height of these panels when tabbed should be the same as that of the mixer by default (when opened by itself)
3. The user should be able to change the height of the tabbed panels up to the maximum value currently specified for the mixer, and without any glitches.

**Platform information**
 - MacOS 12.1 (M1)
 - Windows 11

**Additional context**
Relates to #9241

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10477

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -95,6 +95,9 @@ DockPage {
     readonly property int horizontalPanelMinHeight: 100
     readonly property int horizontalPanelMaxHeight: 520
 
+    readonly property string verticalPanelsGroup: "VERTICAL_PANELS"
+    readonly property string horizontalPanelsGroup: "HORIZONTAL_PANELS"
+
     readonly property var verticalPanelDropDestinations: [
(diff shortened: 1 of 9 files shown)


DmitryArefiev:
Tested #10477 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Eism (approved):

PR #10898, opened 2022-03-24
Fix #10848: Onboarding process no longer starts in the main 'Score' tab. Rewording of Tutorial description needed. App should start maximised.

### LINKED ISSUE #10848: Onboarding process no longer starts in the main 'Score' tab. Rewording of Tutorial description needed. App should start maximised.
**Describe the bug**
When opening the app for the first time, the opening screen overlays the 'Home' tab. The 'Score' tab should be selected at this point, so the user has a better idea of how the theme colour will affect the main UI. This will also make the transition to the 'Learn' section more obvious.

image

Secondly, when the popup describes the tutorial section, the wording should be revised. Rather than saying 'this is the 'Learn section', it should be clearer, saying something like: 'Behind me, is the 'Learn' section...', or 'When you close this popup, you will see the 'Learn' section', etc. I don't have an exact wording in mind. The current wording just seems a little strange, since it seems to be referring to itself. Almost as if it is saying 'This is the 'Learn' section of this popup'.

image

Lastly, when the app is first launched, it should open maximised. Currently, it open in a little window, which is a bit annoying. Each time, my first action is to maximise it. This will avoid the problem of the popup overlapping most of the UI too.

cbjeukendrup:
Resolves: #10848

### DIFF
--- a/src/appshell/qml/FirstLaunchSetup/TutorialsPage.qml
+++ b/src/appshell/qml/FirstLaunchSetup/TutorialsPage.qml
@@ -29,7 +29,7 @@ import MuseScore.AppShell 1.0
 
 Page {
     title: qsTrc("appshell", "Video tutorials")
-    explanation: qsTrc("appshell", "This is the ‘Learn’ section, where you’ll find tutorials to get you started\n(Video tutorials require an internet connection)")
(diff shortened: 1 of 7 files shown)


boryskuzmenko:
Bug https://github.com/musescore/MuseScore/issues/10848 was tested.
FIXED.

Please, merge to master

RomanPudashkin (approved):

Issue #10686, opened 2022-02-28
[MU4 Issue] General issue with scaling/zoom in MU4


DmitryArefiev:
**Describe the bug**
See https://github.com/musescore/MuseScore/pull/10605#issuecomment-1045200331

> It seems there is a broader issue with scaling/zoom at the moment in MU4. For example, when zooming to 100%, the size of the notation does not correspond with the size of a physical sheet of paper. This was the case in MU3, and I also fixed this once for MU4, but apparently it is broken again and I haven't figured out why.

==================================================================
**Bug 1**
Staff/part properties>Advanced style properties: Staff preview looks small on Win/Linux, FullHD displays 

![01](https://user-images.githubusercontent.com/90187801/155975833-b3db5e86-064c-484a-92df-d9d6536d22a4.PNG)

**To Reproduce**
1. Select a staff
2. Right click menu>Staff/part properties>Advanced style properties>Staff preview - compare with MU3

**Platform information**
Win10/Linux on FullHD displays (100%)

===================================================================

PR #10903, opened 2022-03-25
Fix default zoom, and scaling in ExampleView

cbjeukendrup:
(the latter is used in the TimeSigProperties dialog and Staff/Part Properties > Advanced)

Inspired by MuseScore 3.

Tested on multiple resolutions on M1 MacBook Pro built-in display with macOS and Windows via Parallels. 

Resolves: #10686

### DIFF
--- a/src/commonscene/exampleview.cpp
+++ b/src/commonscene/exampleview.cpp
@@ -84,7 +84,7 @@ ExampleView::ExampleView(QWidget* parent)
 
     m_stateMachine->start();
 
-    m_defaultScaling = 0.8 * notationConfiguration()->guiScaling() * notationConfiguration()->notationScaling();
+    m_defaultScaling = 0.9 * notationConfiguration()->notationScaling();
 }
 
 ExampleView::~ExampleView()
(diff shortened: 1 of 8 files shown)


igorkorsukov (approved):

DmitryArefiev:
@cbjeukendrup It's good on Win and Mac. But on Linux it's not centered properly tab staves are cutted:

linux

Should I log that separately?

cbjeukendrup:
@DmitryArefiev Yes, I think it would be good to log that separately. The difficulty is that I don't have Linux, so it is difficult for me to try out solutions...

DmitryArefiev:
..oh, I also found on Windows on 150% scaling the preview is too small

Untitled

Also a separate issue?

cbjeukendrup:
Hm, that's probably related. Maybe I should still have another look at this PR. So no need for a separate issue yet!

DmitryArefiev:
Ok, thanks! I just created an issue for centered case on Linux - #11289

DmitryArefiev:
> Yes, I'll rebase it once more to make sure no problems will occur, and then let's merge it. I have tried several solutions to be able to work on the problem, but with everything I tried, Windows had something to prevent me from editing the scaling settings and testing this issue.

Yes, let's just log it for the future. I'll do it

PR #10907, opened 2022-03-25
Replaced QMap to std::map for engraving

igorkorsukov:


handrok (approved):

cbjeukendrup:
Due to GitHub's recent "degraded performance", no checks have been run for this PR. It appears that the checks would otherwise have failed; at least that is what happens in PR #10924.

igorkorsukov:
> Due to GitHub's recent "degraded performance", no checks have been run for this PR. It appears that the checks would otherwise have failed; at least that is what happens in PR #10924.

Oh, I didn't pay attention to it, there was no red cross and I decided that everything was fine (

cbjeukendrup:
One of the test failures seems to be due to the fact that std::map.insert does something different than QMap.insert. Seems like std::map.insert_or_assign is more appropriate. Particularly in `Score::setMetaTag`.

Jojo-Schmitz:
That though probably won't fix the crashes in utests and vtests, or would it?

Jojo-Schmitz:
And indeed it does not.
```
std::[map](https://www.cplusplus.com/reference/map/map/)::at
      mapped_type& at (const key_type& k);
const mapped_type& at (const key_type& k) const;
Access element
Returns a reference to the mapped value of the element identified with key k.

If k does not match the key of any element in the container, the function throws an [out_of_range](https://www.cplusplus.com/out_of_range) exception.
```
vs.
```
const T QMap::value(const Key &key, const T &defaultValue = T()) const
Returns the value associated with the key key.

If the map contains no item with key key, the function returns defaultValue. If no defaultValue is specified, the function returns a [default-constructed value](https://doc.qt.io/qt-5/containers.html#default-constructed-value). If there are multiple items for key in the map, the value of the most recently inserted one is returned.
```

Jojo-Schmitz:
There are 5 places in the code where value() has been replaced by at(), using try and catch(...) there seems to fix most of the crashes. Not sure which of those are really needed and what's the remaining crash though

src/engraving/libmscore/pitch.cpp
```diff
@@ -33,11 +36,13 @@ int PitchList::pitchOffset(int tick) const
...
+
+    std::vector<int> ticks = mu::keys(*this);
+    auto it = std::upper_bound(ticks.cbegin(), ticks.cend(), tick);
+    if (it == ticks.cend()) {
         return 0;
     }
-    --i;
-    return i.value();
+    --it;
+    return at(*it);
```
Jojo-Schmitz, line 46: Replacing this with 
'''
    try {
        return at(*it);
    }
    catch (...) {
        return 0;
    }
'''
fixes the vtests crashes and all but one utest crash

Jojo-Schmitz:
See https://github.com/musescore/MuseScore/pull/10907/commits/fef003d7052bd575da022de530a8be446949acf3#r836384740

Jojo-Schmitz:
Also don't quite understand why now a MusicXML test fails, `testExcludeInvisibleElements.xml`

src/engraving/libmscore/pitch.cpp
Jojo-Schmitz: Easier/smaller possible fix:
```
    return mu::value(*this, (*it));
```

igorkorsukov:
fixed all the problems, fixes are merged into the master
https://github.com/musescore/MuseScore/pull/10931

PR #10920, opened 2022-03-26
Fix #10791: focus is setted when switch from instruments and template…

### LINKED ISSUE #10791: When the New Score dialog is triggered, the search bar should automatically be in focus
**Describe the bug**
When the **New Score** dialog is triggered, the search bar should automatically be in focus so users can immediately begin typing.

image

tzarjakob:
… options but not when the wizard is opened directly

### DIFF
--- a/src/engraving/layout/layoutharmonies.cpp
+++ b/src/engraving/layout/layoutharmonies.cpp
@@ -21,7 +21,10 @@
  */
 #include "layoutharmonies.h"
 
+#include <map>
+
 #include "realfn.h"
+#include "containers.h"
 
 #include "libmscore/fret.h"
 #include "libmscore/harmony.h"
@@ -57,7 +60,7 @@ void LayoutHarmonies::alignHarmonies(const System* system, const std::vector<Seg
     // Contains harmonies/fretboard per segment.
     class HarmonyList : public QList<EngravingItem*>
     {
-        QMap<const Segment*, QList<EngravingItem*> > elements;
+        std::map<const Segment*, QList<EngravingItem*> > elements;
         QList<EngravingItem*> modified;
 
         EngravingItem* getReferenceElement(const Segment* s, bool above, bool visible) const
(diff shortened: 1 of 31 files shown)


cbjeukendrup:
The reason that it does not work when opening the dialog, is probably here:
https://github.com/musescore/MuseScore/blob/57f790d6ed2fee969547daababf30bb21bb7c70c/src/project/qml/MuseScore/Project/NewScoreDialog.qml#L62

First, focus is set to the search field, but then, by that line, focus is set back to something else. So, you need to remove or change that line.

If that does not fix the problem, do the following:
1. in StyledDialogView.qml, add a property like this:
    ```qml
    property bool isDoActiveFirstControlOnOpen: true
    ```
2. Also in StyledDialogView.qml, modify the `onOpened` handler:
    ```qml
    onOpened: {
        if (root.isDoActiveFirstControlOnOpen) {
            Qt.callLater(navSec.requestActive)
        }
    }
    ```
3. In NewScoreDialog.qml and in InstrumentsDialog.qml, add `isDoActiveFirstControlOnOpen: false`.

---

Also, I think you should not modify TextInputField for this. Specifically, I don't really feel good about putting `forceActiveFocus()` inside the property getter of the `focus` property. Instead, do the following:
1. In InstrumentsView.qml and TitleListView.qml, add a function like this:
   ```qml
    function focusSearchField() {
        searchField.navigation.requestActive()
    }
    ```
4. In ChooseInstrumentsPage.qml and CreateFromTemplatePage.qml, add a function like this:
   ```qml
    function focusOnFirst() {
        instrumentsView.focusSearchField() // templatesView in CreateFromTemplatePage.qml
    }
    ```
5. In ChooseInstrumentsAndTemplatesPage.qml, find the Loader and add the following:
   ```qml
        onLoaded: {
            if (item && item.focusOnFirst) {
                Qt.callLater(item.focusOnFirst)
            }
        }
    ```
That way, whenever a different page is loaded, the `focusOnFirst` function of that new page will be called.
It's slightly more work, but a bit more correct.

tzarjakob:
Thank you, very useful. I've cleaned up and I'll fix the problem according to your advice. Soon I'll share the update.

RobFog:
Instead of closing the pull request, you can mark it as a draft.

tzarjakob:
Yes, I wanted to clean up (don't want to keep changes), but you are right, could have been marked as draft

Jojo-Schmitz:
Better not use your master branch, but rather some topic branch

Eism:
Hey @tzarjakob 
I agree with @cbjeukendrup and I prefer the second version of the fixes he suggested
If it's not difficult for you, please implement it in #10791

tzarjakob:
I have already implemented following his advice in #10920 but I don't see a second version of his suggestions.

Eism:
@tzarjakob 
Ah, I got it

Issue #9232, opened 2021-09-22
[MU4 Task] On-screen Piano Keyboard


jessjwilliamson:
Document and designs for on-screen piano keyboard now ready to be implemented :)

I've sent you the link to the pdf spec on Discord.
The design assets for it in PIANO KEYBOARD in 'MS4 Main Interface Panels page' on Figma - not master!

![image](https://user-images.githubusercontent.com/71656189/159942439-a588223e-6398-44d5-8d08-da4ca4c5bdb3.png)

Issue #9168, opened 2021-09-17
[MU4 Issue] View menu commands - Synthesizer, Piano, Score Comparison don't seem to work


wizofaus:
**Describe the bug**
The View menu commands Synthesizer, Piano, Score Comparison don't seem to work

**To Reproduce**

1. Open or create a new score
2. Select any of the commands from the view menu - Synthesizer, Piano, Score Comparison
3. Nothing happens...

**Expected behavior**
Behaviour as per v3

**Screenshots**

**Desktop (please complete the following information):**
Windows 10

**Additional context**
Nightly build from https://github.com/musescore/MuseScore/actions/runs/1244135645

Presumably this should be done as 3 separate issues, but not sure what the intention is with supporting these.

PR #10926, opened 2022-03-28
Implemented On-screen Piano Keyboard

cbjeukendrup:
Resolves: #9168 (in that issue, only the Score comparison tool remains, but it should be a separate task)
Resolves: #9232

https://user-images.githubusercontent.com/48658420/160305581-458b66c2-6975-4224-9803-3d69e0cdb540.mov

In the first part of the video, you see the zooming and scrolling options of the new keyboard. You can zoom using a pinch gesture on your trackpad, or Ctrl+Scroll, or using the options in the context menu.

In the second half, you see the keyboard being used to input notes. The selected notes on the score are also reflected on the keyboard, just like in MU3. When you select notes in a chord, the non-selected notes of that chord are also highlighted, in a lighter color. The highlight color follows the accent color of the UI. 

In the video, the middle C is C3. After making the video, I decided to change it to C4, which means that the lowest MIDI note is C-1, not C-2 as specified in the design. I thought C4 is more common.

The thing that this keyboard can _not_ yet do, compared to MU3, is displaying the pitches that are currently played. In MU3, the logic to update the piano keyboard was inside the sequencer. Let's think of a better solution for MU4 :)


mattblo:
The build from these artefacts from this PR crashes when, for example, the Mahler Symphony No.9 1st Movement is opened (https://musescore.com/santinocorleone/mahler-symphony-no-9-1st-movement).

This is not the case in the last NightlyBuilds...
At least in my case I was able to reproduce this a few times. NightlyBuilds open (albeit after some waiting time), the builds from this PR crash.

src/notation/view/pianokeyboard/pianokeyboardpanelcontextmenumodel.cpp
```diff
@@ -0,0 +1,183 @@
...
+    });
+
+    item->setArgs(ActionData::make_arg1<int>(numberOfKeys));
+
+    return item;
+}
+
+void PianoKeyboardPanelContextMenuModel::updateKeyWidthScalingItems()
+{
+    qreal roundedCurrentScaling = 1.0;
```
RomanPudashkin, line 168: You should add static constants for theses numbers and use them instead

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+#include <QPainter>
+
+#include "pianokeyboardcontroller.h"
+
+#include "log.h"
+
+using namespace mu::notation;
+
+static const QColor backgroundColor(36, 36, 39);
```
RomanPudashkin, line 32: I would suggest moving all such colors to the configuration level

src/notation/view/pianokeyboard/pianokeyboardview.h
```diff
@@ -0,0 +1,120 @@
...
+
+    std::optional<piano_key_t> keyAt(QPointF position) const;
+
+    static constexpr piano_key_t MIN_KEY = 0;
+    static constexpr piano_key_t MAX_NUM_KEYS = 128;
+
+    piano_key_t m_lowestKey = MIN_KEY;
+    piano_key_t m_numberOfKeys = MAX_NUM_KEYS;
+
+    PianoKeyboardController* m_controller;
```
RomanPudashkin, line 100: You missed nullptr here

src/notation/view/pianokeyboard/pianokeyboardview.h
```diff
@@ -0,0 +1,120 @@
...
+
+    void moveCanvas(qreal dx);
+    void updateScrollBar();
+
+    void wheelEvent(QWheelEvent* event) override;
+    void mousePressEvent(QMouseEvent* event) override;
+    void mouseMoveEvent(QMouseEvent* event) override;
+    void mouseReleaseEvent(QMouseEvent* event) override;
+
+    std::optional<piano_key_t> keyAt(QPointF position) const;
```
RomanPudashkin, line 92: Should be const QPointF&

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    m_scrollBarPosition = newPosition;
+    m_scrollBarSize = newSize;
+    emit scrollBarChanged();
+}
+
+void PianoKeyboardView::setScrollBarPosition(qreal position)
+{
+    m_scrollOffset = -position* m_keysAreaRect.width();
```
RomanPudashkin, line 409: `-position* m_keysAreaRect.width();`

Wrong indentation :)

Also, it is better to check if scollOffset was actually changed before calling update()

```
qreal newScrollOffset = -position * m_keysAreaRect.width();
if (m_scrollOffset == newScrollOffset) {
    return;
}

adjustKeysAreaPosition();
update();
```

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+    m_keyWidthScaling = newScaling;
+    m_scrollOffset *= correctedFactor;
+    m_scrollOffset += x * (1 - correctedFactor);
+
+    emit keyWidthScalingChanged();
+    calculateKeyRects();
+    update();
+}
+
+qreal abs2d(qreal x, qreal y)
```
RomanPudashkin, line 377: This function should be static. In addition, I would suggest moving it to the beginning of the file

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+}
+
+void PianoKeyboardView::scale(qreal factor, qreal x)
+{
+    setScaling(m_keyWidthScaling * factor, x);
+}
+
+void PianoKeyboardView::setScaling(qreal scaling, qreal x)
+{
+    qreal newScaling = std::clamp(scaling, 0.5, 2.0);
```
RomanPudashkin, line 365: Perhaps these numbers are from the PianoKeyboardPanelContextMenuModel? You should use static constants here instead of raw numbers

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+    qreal correctedFactor = newScaling / m_keyWidthScaling;
+
+    m_keyWidthScaling = newScaling;
```
RomanPudashkin, line 368: The same as 
https://github.com/musescore/MuseScore/pull/10926/files#r839346249

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+    m_blackKeyRects.clear();
+    m_whiteKeyRects.clear();
+
+    m_spacing = std::min(2.0 * m_keyWidthScaling, 2.0);
+
+    qreal whiteKeyWidth = 32.0 * m_keyWidthScaling;
+    qreal blackKeyWidth = 20.0 * m_keyWidthScaling;
+
+    qreal whiteKeyHeight = std::min(height() - m_spacing, 4.27 * whiteKeyWidth);
+    qreal blackKeyHeight = whiteKeyHeight * 0.625;
```
RomanPudashkin, line 98: What do these 4.27, 0.625 mean?

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+        painter->translate(rect.topLeft());
+
+        QColor fillColor = m_whiteKeyStateColors[m_controller->keyState(key)];
+
+        painter->fillPath(path, fillColor);
+
+        if (key % 12 == 0) {
+            // Draw octave label
+            qreal octaveLabelBottomOffset = std::min(8.0 * m_keyWidthScaling, 8.0);
```
RomanPudashkin, line 216: Magic numbers

src/notation/view/pianokeyboard/pianokeyboardview.h
```diff
@@ -0,0 +1,120 @@
...
+
+    Q_PROPERTY(int numberOfKeys READ numberOfKeys WRITE setNumberOfKeys NOTIFY numberOfKeysChanged)
+    Q_PROPERTY(qreal keyWidthScaling READ keyWidthScaling WRITE setScaling NOTIFY keyWidthScalingChanged)
+
+    Q_PROPERTY(qreal scrollBarPosition READ scrollBarPosition WRITE setScrollBarPosition NOTIFY scrollBarChanged)
+    Q_PROPERTY(qreal scrollBarSize READ scrollBarSize NOTIFY scrollBarChanged)
+
+public:
+    explicit PianoKeyboardView(QQuickItem* parent = nullptr);
+    ~PianoKeyboardView();
```
RomanPudashkin, line 52: You missed override here

src/notation/view/pianokeyboard/pianokeyboardview.cpp
```diff
@@ -0,0 +1,476 @@
...
+    updateScrollBar();
+}
+
+void PianoKeyboardView::determineOctaveLabelsFont()
+{
+    m_octaveLabelsFont.setFamily(QString::fromStdString(uiConfiguration()->fontFamily()));
+    m_octaveLabelsFont.setPixelSize(uiConfiguration()->fontSize() * std::min(m_keyWidthScaling, 1.5));
+}
+
+void PianoKeyboardView::updateKeyStateColors()
```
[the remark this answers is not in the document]
cbjeukendrup: I would not prefer that. NotationConfiguration is already _huge_, and these methods and colors are very specific to the piano keyboard. I would prefer to keep these things close to the rest of the code for the keyboard.

RomanPudashkin (approved):

PR #10928, opened 2022-03-28
fixed #9658: Resolve current screen for popups by parent item

### LINKED ISSUE #9658: [MU4 Issue] menus wrongly drawn on second monitor
**Describe the bug**
If the MU4 window is maximized, the main dropdown menus will all be drawn disconnected, on a second monitor.
 
**To Reproduce**
1. Install a second monitor.  Place it above laptop.
2. Launch MU4.  Maximize the window on the laptop (i.e. lower monitor).
3. Click the File menu (or Edit, View, Add, etc.)
4. The menu "drops down" from the top of the upper monitor.  It is inaccessible because it is disconnected and the mouse can't reach it (before it disappears, that is).
5.  If MU4 is maximized on the upper monitor, the menus are drawn *behind* the window.  Again, inaccessible.

**Expected behavior**
A dropdown menu should be drawn within and above the main app window always.

**Screenshots**
will do if people can't replicate this

**Desktop**
 - OS: Windows 10.

**Additional context**
none

[EDIT - now that I think about it, here's some additional context.  Hi, I'm David, I'm new here.  I have a lot to learn, but I intend to stay onboard as far as I can.

Eism:
Resolves: #9658

### DIFF
--- a/src/framework/uicomponents/view/popupview.cpp
+++ b/src/framework/uicomponents/view/popupview.cpp
@@ -174,7 +174,8 @@ void PopupView::open()
         m_window->setResizable(m_resizable);
     }
 
-    m_window->show(m_globalPos.toPoint(), m_openPolicy != OpenPolicy::NoActivateFocus);
+    QScreen* screen = resolveScreen();
+    m_window->show(screen, m_globalPos.toPoint(), m_openPolicy != OpenPolicy::NoActivateFocus);
 
     m_globalPos = QPointF(); // invalidate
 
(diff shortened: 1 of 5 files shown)


RomanPudashkin (approved):

bkunda:
FIXED on MacOS.

Unfortunately a similar issue persists in Windows. The first problem is that the side panels appear wrongly scaled (so the three-dots icon is not visible), but after re-sizing the panel the context menu still appears in a different place to where it should be:

https://user-images.githubusercontent.com/86290556/160404572-e035f64d-2b7c-41ee-a85d-e51102bd1947.mp4

Eism:
fixed popups for Windows

> The first problem is that the side panels appear wrongly scaled 

will fix it in a separate PR

bkunda:
FIXED in Windows

PR #10944, opened 2022-03-29
[MU4] Fix user dir default to "MuseScore4" or "MuseScore4Development"

Jojo-Schmitz:
or whatever the `applicationName()` is going to be, for it to not collide with the default user dir of MuseScore 1.x

### DIFF
--- a/src/framework/global/internal/globalconfiguration.cpp
+++ b/src/framework/global/internal/globalconfiguration.cpp
@@ -102,7 +102,7 @@ io::path GlobalConfiguration::userBackupPath() const
 
 io::path GlobalConfiguration::userDataPath() const
 {
-    static io::path p = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation) + "/MuseScore";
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
I firmly believe this needs to be in the Alpha...

src/framework/global/internal/globalconfiguration.cpp
```diff
@@ -102,7 +102,7 @@ io::path GlobalConfiguration::userBackupPath() const
 
 io::path GlobalConfiguration::userDataPath() const
 {
-    static io::path p = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation) + "/MuseScore";
+    static io::path p = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation) + "/MuseScore4";
```
cbjeukendrup, line 105: How about making it `MuseScore4Development` for development builds? That was MuseScore 3's behaviour at least.

So that would probably be `QCoreApplication::applicationName()`, like in MU3.

Jojo-Schmitz: Great minds think alike, I was working on this already and have just pushed it

RomanPudashkin (approved):

PR #10959, opened 2022-03-29
fix #10674 - ensure "select similar in range" and "select dialog" menu options work as expected

### LINKED ISSUE #10674: [MU4 Issue] Select submenu missing from context menu if range selection present
The context menu that appears on right-click an element is missing the "Select" submenu if a range selection is present.  This is an key regression, since losing that menu negates the ability to ever use the "Select: similar in the range" item that is normally present when there is no range, or to use the dialog.  BTW, do we really need the word "the" in "similar in the range"?  And if so, maybe we should add it "similar in same staff"?

To reproduce:

1. Make a range selection
2. Right-click a note

Result: contextual menu is missing the "Select" submenu

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/10674)

A number of behaviours around "select similar in range" and the "select dialog" weren't correct.

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -3359,7 +3359,7 @@ void Score::selectAdd(EngravingItem* e)
     SelState selState = _selection.state();
 
     if (_selection.isRange()) {
-        select(0, SelectType::SINGLE, 0);
+        select(e, SelectType::SINGLE, 0);
         return;
     }
 
--- a/src/notation/inotationinteraction.h
+++ b/src/notation/inotationinteraction.h
@@ -269,6 +269,8 @@ public:
 };
 
 using INotationInteractionPtr = std::shared_ptr<INotationInteraction>;
+
+EngravingItem* contextItem(INotationInteractionPtr);
 }
 
 #endif // MU_NOTATION_INOTATIONINTERACTION_H
--- a/src/notation/internal/notationuiactions.cpp
+++ b/src/notation/internal/notationuiactions.cpp
@@ -256,22 +256,22 @@ const UiActionList NotationUiActions::m_actions = {
              ),
     UiAction("select-similar",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Select: similar"),
+             QT_TRANSLATE_NOOP("action", "Similar"),
              QT_TRANSLATE_NOOP("action", "Select all similar elements")
              ),
     UiAction("select-similar-staff",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Select: similar in same staff"),
-             QT_TRANSLATE_NOOP("action", "Select all similar elements in same staff")
+             QT_TRANSLATE_NOOP("action", "Similar on this staff"),
+             QT_TRANSLATE_NOOP("action", "Select all similar elements on the same staff")
              ),
     UiAction("select-similar-range",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Select: similar in the range"),
+             QT_TRANSLATE_NOOP("action", "Similar in this range"),
              QT_TRANSLATE_NOOP("action", "Select all similar elements in the range selection")
              ),
     UiAction("select-dialog",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Select dialog"),
+             QT_TRANSLATE_NOOP("action", "More..."),
              QT_TRANSLATE_NOOP("action", "Select all similar elements with more options")
              ),
     UiAction("notation-delete",
--- a/src/notation/view/notationcontextmenumodel.cpp
+++ b/src/notation/view/notationcontextmenumodel.cpp
@@ -161,24 +161,23 @@ MenuItemList NotationContextMenuModel::makeHarmonyItems()
 
 MenuItemList NotationContextMenuModel::makeSelectItems()
 {
-    MenuItemList items {
-        makeMenuItem("select-similar"),
-        makeMenuItem("select-similar-staff"),
-        makeMenuItem("select-similar-range"),
-        makeMenuItem("select-dialog"),
-    };
-
-    return items;
+    if (isSingleSelection()) {
+        return MenuItemList { makeMenuItem("select-similar"), makeMenuItem("select-similar-staff"), makeMenuItem("select-dialog") };
+    } else if (canSelectSimilarInRange()) {
+        return MenuItemList { makeMenuItem("select-similar-range"), makeMenuItem("select-dialog") };
+    } else if (canSelectSimilar()) {
+        return MenuItemList{ makeMenuItem("select-dialog") };
+    }
+    return MenuItemList();
 }
 
 MenuItemList NotationContextMenuModel::makeElementItems()
 {
     MenuItemList items = makeDefaultCopyPasteItems();
-
-    if (isSingleSelection()) {
-        items << makeMenu(qtrc("notation", "Select"), makeSelectItems());
+    MenuItemList selectItems = makeSelectItems();
+    if (!selectItems.isEmpty()) {
+        items << makeMenu(qtrc("notation", "Select"), selectItems);
     }
-
     return items;
 }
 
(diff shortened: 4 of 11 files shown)


src/engraving/libmscore/score.cpp
```diff
@@ -3345,7 +3345,7 @@ void Score::selectAdd(EngravingItem* e)
     SelState selState = _selection.state();
 
     if (_selection.isRange()) {
-        select(0, SelectType::SINGLE, 0);
```
wizofaus, line 3348: this could have never done anything useful...

src/notation/inotationinteraction.h
```diff
@@ -268,6 +268,8 @@ class INotationInteraction
 };
 
 using INotationInteractionPtr = std::shared_ptr<INotationInteraction>;
+
+EngravingItem* contextItem(INotationInteractionPtr);
```
wizofaus, line 272: free function because C++ doesn't suppose extension methods...

src/notation/view/notationcontextmenumodel.cpp
```diff
@@ -166,6 +165,8 @@ MenuItemList NotationContextMenuModel::makeElementItems()
 
     if (isSingleSelection()) {
         items << makeMenu(qtrc("notation", "Select"), makeSelectItems());
+    } else if (canSelectSimilarInRange()) {
+        items << makeMenu(qtrc("notation", "Select"), MenuItemList { makeMenuItem("select-similar-range"), makeMenuItem("select-dialog") });
     }
```
wizofaus, line 170: this might not be quite right actually, the select-dialog should probably be available even if there's a list selection (multiple individual elements selected) and you right-click on an element?

src/notation/view/widgets/selectnotedialog.cpp
```diff
@@ -234,39 +234,38 @@ void SelectNoteDialog::apply() const
         return;
     }
 
-    EngravingItem* selectedElement = interaction->selection()->element();
+    EngravingItem* selectedElement = contextItem(interaction);
     if (!selectedElement) {
         return;
     }
 
-    FilterElementsOptions options = noteOptions();
```
wizofaus, line 242: in reference-based languages (C#/Java etc.) this would have worked fine, but in C++ it's a slice!

src/notation/view/widgets/selectnotedialog.cpp
```diff
@@ -234,39 +234,38 @@ void SelectNoteDialog::apply() const
...
+    FilterNotesOptions options = noteOptions();
 
     std::vector<EngravingItem*> elements = notationElements->elements(options);
     if (elements.empty()) {
         return;
     }
+    if (isInSelection()) {
```
wizofaus, line 248: OK so all this code is the same as that in selectdialog.cpp - but a) it was already like that, I just had to add more duplicate code b) I assume this will be rewritten to use QtQuick/QML at some point.

bkunda:
I am testing this PR now and will provide comments here

bkunda:
Have tested this on Mac.
I like how the context menu options are now...well, contextual! (I.e. only the options relevant to the selection are shown). My only suggestion/request would be a slight tweak to the microcopy:

- Select > Similar
- Select > Similar on this staff
- Select > Similar in this range
- Select > More...

I don't think we need the repetition of the word "Select: xxx", and the label "Select dialog" is quite confusing (are we selecting a dialog?)

As far as the dialog interactions themselves go, it all seems to work as expected, although I can't figure out the "Action" section, in particular the "Add to selection", which seems to suggest that whatever the selection will be as a result of this dialog will be added to an existing selection, which I can't seem to get to work. Anyway, this might be more an interaction design issue with the existing dialog itself, and I'm not going to worry too much about it in the grand scheme of things at this stage.

One question @wizofaus: can these dialogs be accessed in any way other than via the context menu? 

Thanks so much for putting this together. @wizofaus if you can please fix up the menu labels I'll then ping one of our devs for a review soon :-)

jeetee:
"Add to selection" is usually used when starting from a list selection rather than a range selection (if not, the end result will become a list selection of all that was in the range + the additional selected elements).

These different select options in MU3 can be triggered either via context menu or with a (non-default set) shortcut.

wizofaus:
> Have tested this on Mac. I like how the context menu options are now...well, contextual! (I.e. only the options relevant to the selection are shown). My only suggestion/request would be a slight tweak to the microcopy:
> 
> * Select > Similar
> * Select > Similar on this staff
> * Select > Similar in this range
> * Select > More...
> 
> I don't think we need the repetition of the word "Select: xxx", and the label "Select dialog" is quite confusing (are we selecting a dialog?)
> 

I agree the menu labels seem odd, but I assume that was following some initial design (as it's different to v3).  If @Tantacrul is happy with your suggestion (it's essentially going back to what v3 had), I'm all for it.

In principle all these commands can be mapped to shortcut keys but for whatever reason they're not available currently (at least, "select similar in range" wasn't), and as I explained elsewhere there's a problem because at least that command and "select dialog" (select more...) depend on the last clicked-on element if there's more than 1 element selected.
But from memory they didn't work via shortcut keys in v3 either.

wizofaus:
So I can easily change the sub-menu items to just have the text "Similar", "Similar on staff" etc. as requested - there's just a slight risk that in future if these items get moved elsewhere or if we allow users to customize where they're shown, they might be not make so much sense.
image


The next problem is that now that I've added support for accessing "Select > More..." when a *list* of individual elements are selected, the problem is you can't right-click on an individual element that's part of the list without deselecting it.  So if you've carefully selected 5 or 6 notes holding down CTRL, then want to use one of them as a reference point for adding/subtracting to the selection, right-clicking on it even while still holding down CTRL will deselect it.  I would say this behaviour is unexpected and doesn't match what either v3 or most other windows apps do.  I'd say that's out of scope for this issue, but there is another similar one I'll put a comment on.

jeetee:
I agree that the list selection being dropped can be considered a separate issue; filed as #11008

wizofaus:
Ok have pushed with menu labels changed, "select > more" now works when there's a list selection, and have hidden the unused "search in selection" radio button.

bkunda:
> So I can easily change the sub-menu items to just have the text "Similar", "Similar on staff" etc. as requested - there's just a slight risk that in future if these items get moved elsewhere or if we allow users to customize where they're shown, they might be not make so much sense. image

Thanks for this. I'm not too worried about this because it's something we'll have to deal with if we re-design this feature.

bkunda:
> The next problem is that now that I've added support for accessing "Select > More..." when a _list_ of individual elements are selected, the problem is you can't right-click on an individual element that's part of the list without deselecting it.

Yes this is a problem, and I now see you've already created a new PR to address it, so thank you!

bkunda:
> Ok have pushed with menu labels changed, "select > more" now works when there's a list selection, and have hidden the unused "search in selection" radio button.

I've just checked this latest build on Mac and the labels still need correcting please:

"Similar in the same staff" should read "Similar on this staff"
"Similar in the range" should read "Similar in this range"

 Great that the "More" option now works on list selections! 🙏🏻

Jojo-Schmitz:
Some simple codestyle fixes needed

wizofaus:
BTW it turns out "Select > similar" doesn't work for slurs (and other similar items) when the grip handles are showing.  Looking into it though it's probably not strictly related to anything here.

bkunda:
> BTW it turns out "Select > similar" doesn't work for slurs (and other similar items) when the grip handles are showing. Looking into it though it's probably not strictly related to anything here. However it is a trivial fix, might as well just add to this one seeing as it's related.

Good spot. I was going to suggest a separate issue, but seeing as how you're fixing it here, that saves work for everyone!
Many thanks 🙂

RobFog:
Rebase needed.

wizofaus:
I don't know what the hold up with getting this merged was. I'd rather hand it over to the core team TBH.

wizofaus:
I did just re-test this after the rebase and it appears to be working fine...

Tantacrul:
@bkunda - perhaps one last test + confirmation before recommending merge?

bkunda:
Tested on MacOS. FIXED. 👍🏻

src/notation/internal/notationuiactions.cpp
```diff
@@ -256,22 +256,22 @@ const UiActionList NotationUiActions::m_actions = {
...
     UiAction("select-similar-range",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Select: similar in the range"),
+             QT_TRANSLATE_NOOP("action", "Similar in this range"),
              QT_TRANSLATE_NOOP("action", "Select all similar elements in the range selection")
              ),
     UiAction("select-dialog",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Select dialog"),
+             QT_TRANSLATE_NOOP("action", "More..."),
```
Jojo-Schmitz, line 274: Can we please have a real ellipsis here, a single 3-dot glyph rather than 3 single e dots?

Jojo-Schmitz: See #11506

Issue #10700, opened 2022-03-01
[MU4 Task] Identify non-default styles in saved files


oktophonie:
**Task description**

Currently in MU4, _every_ style setting is saved as part of the file package (in `score_style.mss`), and inside the .mscx file, the complete instrument definition data (from `instruments.xml`) for each instrument is copied to an `<Instrument>` element inside each `<Part>`.

It is proposed that instead, we include only style settings that _differ_ from the default settings (as given in `styledef.cpp`) - which was the behaviour pre-MU4 (and in fact continues to be the behaviour in that the non-default settings are still saved into the `.mscx` currently, as well as the 'complete' style information going into the `.mss` file) - and also only the instrument data that differs from that of `instruments.xml` (by default, an unaltered instrument could be represented simply by e.g.  `<Instrument id="piano" />`).

As well as a modest reduction of file size, this also helps in development since, whenever a style has its default changed, or an instrument definition is tweaked, we do not have to go through all the templates and tests to update them to match. If these files require custom style or instrument data it can still be included where necessary.(Special care will be required for tests which explicitly address the application of styles!)

For score migration, we already include style sheets with the legacy defaults for previous file versions (`legacy-style-defaults-v1|2|3|302.mss`) which are used to provide the 'base' styles for any scores created in previous versions, so much of this functionality already exists. We could do something similar for instrument definitions if necessary.

When loading a file, assuming the `.mscx` or `.mss` file contains a reference to the file format version (which they currently do), all 'missing', i.e. default, information can be repopulated. When saving, all settings and instrument information will be compared against the defaults, and included only where it differs.

Explicitly saving an .mss file from within MuseScore would, of course, still save the complete set of styles.

One potential disadvantage is that this renders styles generally implicit rather than explicit, i.e. if the underlying defaults change then the settings of the score will also change. However, these changes should only happen as part of a new software version (and/or file format version), and as long as we supply complete style sheets for each version, we can continue to honour the settings of the version in which the file was previously saved, with only those exceptions as required by the specific migration path being taken (i.e. for styles which are now deprecated, have a different interpretation so require a recalculation, etc.)

Comments/outrage welcome!

shoogle:
It turns out that we _do_ need to save all the styles after all, including the default styles. However, we _also_ need to save a tag or other indicator to tell us whether the saved style was default or not. See https://github.com/musescore/MuseScore/pull/10963#issuecomment-1095529655.

shoogle:
[...]

PR #10963, opened 2022-03-30
[MU4] Remove excess style information from save file

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/10700

In 4.0 the style setting data saving and loading changed in order to include every style setting in every save file. This PR reverts back to previous functionality. See the linked issue for details.


asattely:
The vtest failures seem to be due to the horizontal spacing adjustments that are going on potentially. Style information being in the save file XML means that there is work to be done vis a vis adding it back into the reference files for the utests. I'll be finished doing that tomorrow!

oktophonie (approved):

shoogle:
@vpereverzev was against merging this for compatibility reasons. We thought we had that covered by loading the [legacy default styles](https://github.com/musescore/MuseScore/tree/master/src/engraving/data/styles), which would give us backwards compatibility with older MuseScore versions. However, it wouldn't give us *forwards compatibility* with future MuseScore versions.

Let's imagine we want to change the default styles in MuseScore 4.1. If a MuseScore 4.1 user sends a score to a MuseScore 4.0 user, MuseScore 4.0 is not going to know about any changes to default styles in MuseScore 4.1, hence all styles need to be saved in the file. We could ignore the problem and tell users to upgrade, but students and teachers might not be able to install updates on their school/university computers, and they need to be able to send files to each other.

So @vpereverzev was right: we need to keep all styles in the file, including the default styles. However, as @cbjeukendrup mentioned, it would be useful to know which styles are the default ones and which ones were explicitly chosen by the user. This could be done with a `source="user"` attribute in score_styles.mss:

```xml
<museScore version="4.00">
  <Style>
    ...
    <beginRepeatLeftMargin>1</beginRepeatLeftMargin> <!-- default style -->
    <minNoteDistance source="user">0.4</minNoteDistance> <!-- custom style chosen by user -->
    <barNoteDistance source="user">1.3</barNoteDistance>
    <barAccidentalDistance>0.65</barAccidentalDistance>
    <noteBarDistance source="user">1.5</noteBarDistance>
    <measureSpacing>1.5</measureSpacing>
    ...
  </Style>
</museScore>
```

If we do this now for files saved in MuseScore 4.0, it would enable us to offer a feature in MuseScore 4.1 where users can choose to upgrade the default styles but keep all the custom ones as they were, so the time users spent in customising their styles is not lost. Without this feature, users would be forced to choose between keeping their custom styles with the old defaults, or resetting all styles to the new defaults in MuseScore 4.1.

Issue #10967, opened 2022-03-30
[MU4 Issue] MU4 crashes when opening MU3 score


henkdegroot:
When I open the scores provided in the zip file, MU4 crashes.

Using the private alpha release on Windows 10.

[MU3 scores.zip](https://github.com/musescore/MuseScore/files/8377377/MU3.scores.zip)

Jojo-Schmitz:
One of therm stems from 3.5.2, the other from 3.6.2

Jojo-Schmitz:
Stack trace:
```
                                                                                                                                                                
                                                                                                                                                                
1   Ms::Slur::layoutSystem(Ms::System *)                                                                                                               0x7447ea 
2   mu::engraving::LayoutSystem::processLines(Ms::System *, std::vector<Ms::Spanner *>, bool)                                                          0x666e8d 
3   mu::engraving::LayoutSystem::layoutSystemElements(mu::engraving::LayoutOptions const&, mu::engraving::LayoutContext&, Ms::Score *, Ms::System *)   0x6697ac 
4   mu::engraving::LayoutSystem::collectSystem(mu::engraving::LayoutOptions const&, mu::engraving::LayoutContext&, Ms::Score *)                        0x66edf6 
5   mu::engraving::LayoutPage::collectPage(mu::engraving::LayoutOptions const&, mu::engraving::LayoutContext&)                                         0x757943 
6   mu::engraving::Layout::doLayout(mu::engraving::LayoutOptions const&, mu::engraving::LayoutContext&)                                                0x65cfa6 
7   mu::engraving::Layout::doLayoutRange(mu::engraving::LayoutOptions const&, mu::engraving::Fraction const&, mu::engraving::Fraction const&)          0x66fccf 
8   Ms::Score::doLayoutRange(mu::engraving::Fraction const&, mu::engraving::Fraction const&)                                                           0x4be349 
9   Ms::Score::update(bool)                                                                                                                            0x5b1982 
10  Ms::Score::endCmd(bool)                                                                                                                            0x5b20f6 
11  mu::project::ProjectMigrator::migrateProject(std::shared_ptr<mu::engraving::EngravingProject>, mu::project::MigrationOptions const&)               0x99e73e 
12  mu::project::ProjectMigrator::migrateEngravingProjectIfNeed(std::shared_ptr<mu::engraving::EngravingProject>)                                      0x99ecdd 
[...]

PR #10978, opened 2022-03-30
Fix more tie crashes

cbjeukendrup:
Resolves: #10967

### DIFF
--- a/src/engraving/libmscore/slur.cpp
+++ b/src/engraving/libmscore/slur.cpp
@@ -1348,14 +1348,14 @@ SpannerSegment* Slur::layoutSystem(System* system)
     // adjust for ties
     p1 = sPos.p1;
     p2 = sPos.p2;
-    Chord* sc = startChord();
-    Chord* ec = endChord();
     bool constrainLeftAnchor = false;
 
     // start anchor, either on the start chordrest or at the beginning of the system
     if (sst == SpannerSegmentType::SINGLE || sst == SpannerSegmentType::BEGIN) {
-        // on chordrest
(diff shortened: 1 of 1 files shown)


DmitryArefiev:
Tested #10967 on Win10, Mac12, LinuxMint 20.2 - **FIXED**      

Please merge into master

src/engraving/libmscore/slur.cpp
```diff
@@ -1522,9 +1524,9 @@ SpannerSegment* Slur::layoutSystem(System* system)
             Chord* c = toChord(cr);
             Tie* tie = nullptr;
             PointF endPoint;
-            if (c->notes()[0]->tieBack() && !c->notes()[0]->tieBack()->isInside() && c->notes()[0]->tieFor()->up() == up()) {
+            if (c->notes()[0]->tieBack() && !c->notes()[0]->tieBack()->isInside() && c->notes()[0]->tieBack()->up() == up()) {
```
Jojo-Schmitz, line 1527: I tried this, didn't help...

src/engraving/libmscore/slur.cpp
```diff
@@ -1522,9 +1524,9 @@ SpannerSegment* Slur::layoutSystem(System* system)
...
                 // there is a tie that ends on this chordrest
-                tie = sc->notes()[0]->tieBack();
+                tie = c->notes()[0]->tieBack();
```
Jojo-Schmitz, line 1529: ...but didn''t spot this, so the combination of both should indeed do the trick ;-)

PR #11021, opened 2022-03-31
fix #11008 - various mouse selection issues

### LINKED ISSUE #11008: [MU4 Issue] Secondary (right) click loses list selection
**Describe the bug**
Not being able to secondary click an element while retaining a list selection loses the ability of MU3 to "Add to/Remove from" the selection using the select dialog.

**To Reproduce**
1. Select more than one score element using Ctrl/Cmd+click
2. Right click an element
Notice the list selection is lost

**Expected behavior**
List selection is maintained

**Additional context**
I kind of understand dropping the list selection *if* only one other item was selected; but would advocate for not dropping it to allow more advanced filtering of selected items when using the selection dialog as was possible in MU3.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/11008)

Fixes not being able to secondary click an element while retaining a list selection 

### DIFF
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -536,14 +536,9 @@ bool NotationViewInputController::needSelect(const ClickContext& ctx) const
     if (!ctx.hitElement) {
         return false;
     }
-
-    bool result = true;
-
-    if (ctx.event->button() == Qt::MouseButton::RightButton) {
-        result = !viewInteraction()->selection()->range()->containsPoint(ctx.logicClickPos);
-    }
-
-    return result;
+    return !(ctx.event->modifiers() & Qt::ControlModifier)
+           && !viewInteraction()->selection()->range()->containsPoint(ctx.logicClickPos)
+           && !ctx.hitElement->selected();
 }
 
 void NotationViewInputController::handleLeftClick(const ClickContext& ctx)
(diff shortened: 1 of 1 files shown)


src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -536,14 +536,9 @@ bool NotationViewInputController::needSelect(const ClickContext& ctx) const
     if (!ctx.hitElement) {
         return false;
     }
-
-    bool result = true;
-
-    if (ctx.event->button() == Qt::MouseButton::RightButton) {
-        result = !viewInteraction()->selection()->range()->containsPoint(ctx.logicClickPos);
-    }
```
wizofaus, line 544: This function determines what happens during mouse *press*/down - and in that case, which button you press is irrelevant (that only becomes relevant on mouse release).  But whether CTRL is held down, or the element is already selected definitely is relevant.

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -691,6 +686,10 @@ void NotationViewInputController::mouseReleaseEvent(QMouseEvent*)
         && !interaction->isDragStarted() && !noteInput->isNoteInputMode()) {
         interaction->clearSelection();
     }
+    if (event->button() == Qt::MouseButton::LeftButton && m_view->toLogical(event->pos()) == m_beginPoint
+        && !(event->modifiers() & Qt::KeyboardModifier::ControlModifier) && hitElement()) {
+        interaction->select({ hitElement() });
```
wizofaus, line 691: On releasing the left mouse button, if you clicked on an element, and haven't moved the mouse, and aren't holding down CTRL, then the existing selection should be cleared and replaced with the clicked-on element.

RomanPudashkin:
Thanks for your help! I used your solution as a reference and simplified/improved it a bit
https://github.com/musescore/MuseScore/pull/11114

Issue #11024, opened 2022-03-31
[MU4 Issue] When entering notes in continuous view, measure width doesn't calculate properly 


DmitryArefiev:
**Describe the bug**
Mentioned in #10980:

@mike-spa FYI. This fix doesn't work perfectly. It looks like we need to do some refactoring, because Measure::computeWidth tries to use the system width, which will be available later, so we can't calculate the width correctly

**To Reproduce**
1. Switch to continuous view
2. Start to enter notes
3. See measure width - it's calculated as in page view
4. Switch to continuous view
5. Switch back to page view - now it's calculated (updated) right

https://user-images.githubusercontent.com/90187801/161048684-cb6c8b07-fac6-42c7-ad40-13f8cc344b7c.mp4

**Expected behavior**
Compare with MU3

**Platform information**
 macOS, Windows, Linux

DmitryArefiev:
Fixed in #11025

PR #10980, opened 2022-03-30
[MU4] Fix #10973: Score becomes total empty in continuous view

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10973
Resolves: https://github.com/musescore/MuseScore/issues/10955

@mike-spa FYI. This fix doesn't work perfectly. It looks like we need to do some refactoring, because Measure::computeWidth tries to use the system width, which will be available later, so we can't calculate the width correctly

Issue #10973, opened 2022-03-30
[MU4 Issue] Score becomes total empty in continuous view


DmitryArefiev:
**Describe the bug**
Score becomes total empty in continuous view after score left edge goes out of screen.
Doesn't occur with one staff score

https://user-images.githubusercontent.com/90187801/160805866-8bb7035d-fee6-4e92-878d-7c2cad5dfd6d.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur with vertical continuous view

DmitryArefiev:
Fixed in #10980

Issue #10955, opened 2022-03-29
[MU4 Issue] Improper scaling in continuous view (measure width becomes too large)


Tegze:
**Describe the bug**
Horizontal scaling is too large in continuous view

**To Reproduce**
1. Create a new score with two or more instruments
2. Switch to continuous view (bottom right)

**Expected behavior**
Scaling should be based on the current notes

**Screenshots**
![image](https://user-images.githubusercontent.com/38992967/160693671-6acb57e8-867a-4e9f-8599-71bf70e6741c.png)
![image](https://user-images.githubusercontent.com/38992967/160693702-d4e52b1f-a373-401c-b19b-c0c489462189.png)

**Platform information**
Windows 10, macOS, Linux

**Additional context**
Continuous view works properly with one instrument

DmitryArefiev:
https://user-images.githubusercontent.com/90187801/160809629-e93ff0bf-e362-4f05-9225-e484239645de.mp4

Can reproduce on my side as well. Measure width becomes too large in continuous view.

Thanks for catching this!

DmitryArefiev:
Can be related to #10973

DmitryArefiev:
Fixed in #10980

PR #11025, opened 2022-03-31
Fix spacing in continuous view

mike-spa:
Resolves: [10955](https://github.com/musescore/MuseScore/issues/10955)
[10973](https://github.com/musescore/MuseScore/issues/10973)
[11024](https://github.com/musescore/MuseScore/issues/11024)

Completes the initial solution from PR [10980](https://github.com/musescore/MuseScore/pull/10980) by @RomanPudashkin.

The deleted lines were originally done to help with an edge case in page view but that isn't necessary anymore. 
Additionally, MM rests need to be disregarded when computing the longest note of a system, otherwise they introduce "fake" long notes and cause additional wrong layouts.

The two things together should fix the issue @DmitryArefiev

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -4291,9 +4291,6 @@ void Measure::computeWidth(Fraction minTicks, qreal stretchCoeff)
     } else {
         setWidthLocked(false);
     }
-    if (width() > maxWidth) {
-        setWidthToTargetValue(s, x, isSystemHeader, minTicks, stretchCoeff, maxWidth);
-    }
 }
 
 void Measure::setWidthToTargetValue(Segment* s, qreal x, bool isSystemHeader, Fraction minTicks, qreal stretchCoeff, qreal targetWidth)
@@ -4555,7 +4552,7 @@ Fraction Measure::maxTicks() const
     Segment* s = first();
     Fraction maxticks = Fraction(0, 1);
     while (s) {
-        if (s->enabled()) {
+        if (s->enabled() && s->isChordRestType() && !s->isMMRestSegment()) {
             maxticks = std::max(maxticks, s->ticks());
         }
         s = s->next();


mike-spa:
vtests look ok to me. The only meaningful difference is on MM rests and is as expected.

RomanPudashkin (approved):

mike-spa:
@RomanPudashkin if you're waiting for the green flag from the vtest before merging, that isn't going to happen, probably. Some weird issue is going on with Github and, even though Simon already put the vtest label, it still notifies the failure. Before, the vtest label used to trigger new vtests and they would go green even if differences were found. Not sure what's going on now.

DmitryArefiev:
@mike-spa I see now the measure width in continuous view the same as in page view. This is intentional, right?
Previously, I remember it was smaller then in page view 

Now:

https://user-images.githubusercontent.com/90187801/161115842-623156e7-a0dd-4252-96e6-e1f06f1f6358.mp4

Previously:

https://user-images.githubusercontent.com/90187801/161115897-31126dfb-d4c0-4960-b3c9-35ae8e291ba1.mp4

mike-spa:
@DmitryArefiev Thank you for your feedback! Yes, that is the intended behavior. Basically, in MU3.6 (and also in MU4, but that was an an error which started [here](https://github.com/musescore/MuseScore/pull/10511)) no duration-spacing was applied in continuous view, hence why it was so narrow. Now, duration-spacing is done in `Measure::computeWidth()`, which is used both in the page view and the continuous view, therefore we get equal width (or similar, actually, cause in page view there is the additional step of justifying the systems). 
In future, it may be worth having a dedicated version of `computeWidth()` for continuous view, because indeed continuous view may benefit from slighly narrower appearance. But that's for a later stage :)

mike-spa:
Correction: in MU3.6 some duration-spacing was also applied. Not properly, but it was applied ;)

DmitryArefiev:
@mike-spa Yeah, now measure spacing more convenient in continuous view (not so tight now). 
Thanks for detailed explanation!

Issue #10917, opened 2022-03-26
[MU4 Issue] render bug when select same notes in different voices and scaling


Fince10086:
**Describe the bug**
When selecting the same notes in different voices and scaling, the order of the voices covered will be random.

https://user-images.githubusercontent.com/87692117/160241021-b521681a-a99d-4bc1-8833-d6acb9418b53.mp4

**Expected behavior**
the voice 4 should always cover voice 3, and the 3 cover 2, 2 cover 1. Just likes in MuseScore 3:
![image](https://user-images.githubusercontent.com/87692117/160241165-badc9d26-0d6d-470e-897e-336a57dd0ea5.png)

**Additional context**

KaiVinter:
It doesn't seem to be random, though. It follows the 1-3-2-4 Pattern (1 & 3 ↑ and 2 & 4 ↓)

boryskuzmenko:
**FIXED**
Tested on macOS, Win11, Linux

PR #11026, opened 2022-03-31
Fix sorting elements for drawing and selecting

cbjeukendrup:
Resolves: #10917 

- if you select overlapping elements in different voices, the element with the greatest voice will be drawn on top;
- if you click at a place with multiple elements, generally, the same element would be selected that would be drawn on top if you would select all those elements. 

This is slightly different from MU3, but I think it feels more consistent.


cbjeukendrup:
The vtest differences are not really differences...

Jojo-Schmitz:
Why do the vtests claim to be different? Just curious

cbjeukendrup:
I wouldn't know a logical explanation, but maybe it has to do with antialiasing, that some pixels become by chance a tiny bit lighter?

boryskuzmenko:
https://github.com/musescore/MuseScore/issues/10917 bug is **FIXED**
Tested on macOS, Win11, Linux

Please, merge to Master

Issue #10859, opened 2022-03-21
[MU4 Issue] [Improvement] Score tab should indicate if it has an opened score when switching to other tabs (Home, Publish, DevTools)


DmitryArefiev:
**Describe the bug**
Score tab should indicate if it has an opened score when switching to other tabs (Home, Publish, DevTools)
Score tab ("Score" text) can be bold and maybe highlighted but **NOT underlined** showing that there is an opened score already, when user doing something on Home, Publish or DevTool tab.

https://user-images.githubusercontent.com/90187801/159350217-87a621ff-4468-4f89-be64-9762053457af.mp4

**Platform information**
macOS, Windows, Linux

**Expected result**
If there is an opened score in MU, **Score tab should be bold but not underlined** when clicking/navigating Home or Publish or DevTools tab.

DmitryArefiev:
@Tantacrul @bkunda @jessjwilliamson What do you think about this improvement (and the expected result)?

Tantacrul:
I'm not sure I understand exactly. So, if the Score tab has no opened score, if I click on it, it should just highlight and the text should be bold but there should be no underline?

I do like the idea of trying to trying to help indicate to the user when a score is open.

DmitryArefiev:
> I'm not sure I understand exactly. So, if the Score tab has no opened score, if I click on it, it should just highlight and the text should be bold but there should be no underline?

I think there should not be changes in the behavior (it should be bold, highlighted and underlined when click on it even if it has no opened score (empty))

The changes here are only when user has an opened score and wants to click on Home/Publish/DevTools tabs. In that case Score tab should be bold and maybe highlighted but **NOT underlined**

Sorry for my English 🙂

Tantacrul:
OK, I get it now :)

I don't really know, to be honest. Jess? Could you try this idea out as a mockup when you've time?

DmitryArefiev:
https://user-images.githubusercontent.com/90187801/160824223-d069fbb9-55d3-4eff-a763-15d5541d6416.mp4

@Tantacrul @jessjwilliamson Please see the video. 

1. Roman has made the initial solution for now: **"Score" text is bold** when score is opened and user navigates other tabs. 

2. Also, Roman suggested to mark score preview pictures with some mark (check mark for example) on Home>Score page, to mark score (scores) that are opened.  Something like we want to do with a cloud scores as I understand. 
This item requires a design decision. 


What do you think about №1 and №2?  (as you see, №1 can be implemented already)

[...]

PR #11032, opened 2022-03-31
[MU4] Fix #10859: Score tab should indicate if it has an opened score when switching to other tabs 

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10859

PR #11033, opened 2022-03-31
Fix 10859: Score tab is always bold if it has an opened score

duanjr:
Resolves: [#10859](https://github.com/musescore/MuseScore/issues/10859)

Score tab is always bold if it has an opened score.

### DIFF
--- a/src/appshell/qml/MainToolBar.qml
+++ b/src/appshell/qml/MainToolBar.qml
@@ -32,6 +32,8 @@ Item {
     height: radioButtonList.height
 
     property alias navigation: navPanel
+    property alias navigation: navCtrl
+    property bool isNotationOpened: false
 
     property string currentUri: "musescore://home"
     property var items: [
@@ -96,6 +98,7 @@ Item {
 
             checked: modelData["uri"] === root.currentUri
             title: modelData["title"]
+            isBoldOnly: (modelData["title"] === qsTrc("appshell", "Score")) && !checked && isNotationOpened
 
             onToggled: {
                 root.selected(modelData["uri"])
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -90,6 +90,9 @@ DockPage {
         Qt.callLater(pageModel.init)
     }
 
+
+    property bool isNotationOpened:pageModel.isNotationOpened
+
     readonly property int verticalPanelDefaultWidth: 300
 
     readonly property int horizontalPanelMinHeight: 100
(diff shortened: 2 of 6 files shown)


src/appshell/qml/MainToolBar.qml
```diff
@@ -32,6 +32,8 @@ Item {
     height: radioButtonList.height
 
     property alias navigation: navPanel
+    property alias navigation: navCtrl
```
duanjr, line 35: this line is added to avoid merge conflict, sorry for forgetting pull before my work

RomanPudashkin:
LOL, I was first :)
https://github.com/musescore/MuseScore/pull/11032

src/appshell/qml/MainToolBar.qml
```diff
@@ -96,6 +98,7 @@ Item {
 
             checked: modelData["uri"] === root.currentUri
             title: modelData["title"]
+            isBoldOnly: (modelData["title"] === qsTrc("appshell", "Score")) && !checked && isNotationOpened
```
RomanPudashkin, line 101: I believe this is not the right way of solving the problem. Qml is not the best place for such logic

RomanPudashkin:
Next time, make sure that the issue you're planning to fix isn't being solved by someone else

src/appshell/qml/MainToolBar.qml
duanjr: You are right and I have seen how you dealt with it.

duanjr:
> Next time, make sure that the issue you're planning to fix isn't being solved by someone else

Ah yes, I noticed it was assigned to you but there're so many issues assigned to you.
So I made some try without saying first and it went unexpectedly smoothly. 
I will pay attention next time and really thanks for your guiding.

RomanPudashkin:
> Ah yes, I noticed it was assigned to you but there're so many issues assigned to you.

Yes, that's why I'm against this "let's assign all new issues to the in-house developers" tactic :( As a result, this can lead to some really confusing situations

HemantAntony:
> > Ah yes, I noticed it was assigned to you but there're so many issues assigned to you.
> 
> Yes, that's why I'm against this "let's assign all new issues to in-house developers" tactic :( As a result, this can lead to some really confusing situations

I also face this confusion. Usually, I don't say that I am going to solve an issue because I might not be able to. If I can't and I said I would do it, it leads to an awkward situation

RobFog:
> I also face this confusion. Usually, I don't say that I am going to solve an issue because I might not be able to. If I can't and I said I would do it, it leads to an awkward situation

Don't worry. If you take on an issue and then find out it is too difficult, that's fine, nobody will bite your head off. 😊

Issue #9823, opened 2021-11-22
[MU4 Issue] Plugins list should be an alphabetical order


jessjwilliamson:
**Describe the bug**
I have recently installed a plugin pack that has a long list of plugins inside and it made me realise how the names of the plugins do not list alphabetically in MS4. This applies to all plugins, not just this particular one showing here.

![image](https://user-images.githubusercontent.com/71656189/142884957-ba3be257-2c17-4210-91a1-6b153fe4489c.png)

Compared to Reaper, Logic and Ableton, they do list this as such and it would be much easier for the user to locate a name of a plugin in this way when there is a big list.

![image](https://user-images.githubusercontent.com/71656189/142884991-646f8b85-1faa-45c4-907f-def9437ec0d5.png)
![image](https://user-images.githubusercontent.com/71656189/142885045-6b9a5f3a-2e1c-4eca-a614-5c1a32e4bc43.png)
<img width="1280" alt="Screenshot 2021-11-22 at 17 46 25" src="https://user-images.githubusercontent.com/71656189/142910084-379f2bce-61da-41ae-878c-fd50da4379b9.png">

PR #11049, opened 2022-04-01
Sort the list of available (VST) plugins

cbjeukendrup:
Resolves: #9823


src/playback/view/internal/inputresourceitem.cpp
```diff
@@ -174,4 +175,16 @@ void InputResourceItem::updateAvailableResources(const AudioResourceMetaList& av
         AudioResourceMetaList& resourcesMetaList = resourcesByVendor[meta.vendor];
         resourcesMetaList.push_back(meta);
     }
+
+    for (auto& [type, resourcesByVendor] : m_availableResourceMap) {
```
RomanPudashkin, line 179: The exact same code is presented in two places. Please don't duplicate it

RomanPudashkin, line 179: The exact same code is presented in two places. Please don't duplicate it

cbjeukendrup: Fixed!

RomanPudashkin (approved):

jessjwilliamson:
Hell yeah! Thank you so much!

Issue #9381, opened 2021-10-05
[MU4 Issue] Brackets palette does switch to display dark mode


Nick-Mazuk:
**Describe the bug**

When using dark mode, all the brackets in the brackets palette are still black (and not white like all other palettes are).

<img width="301" alt="Screen Shot 2021-10-05 at 10 21 13 AM" src="https://user-images.githubusercontent.com/5659171/136072082-315d78f8-c4cc-4fea-8b26-ae93bdbcce34.png">

**Desktop (please complete the following information):**
 - OS: macOS 11.6

nanu1605:
@Nick-Mazuk @RomanPudashkin  Hi, I am pretty new to open source but would love to work on this issue. Could you please help me with the issue more.

duanjr:
I have made a pr to fix this issue.
Though it performs correctly in all four themes, I don't think it's good enough to be merged.
I noticed that we once fixed a similar issue [#8714](https://github.com/musescore/MuseScore/issues/8714) in this [commit](https://github.com/cbjeukendrup/MuseScore/commit/88e5daa537eaf23e9dfc76dfca119c1d7cdaf143). But when I reset to that commit, the brackets still have the wrong color.
Could someone give me some guidance?

boryskuzmenko:
Bugfix is verified
OS: macOS, Windows
**FIXED**

PR #11072, opened 2022-04-03
Fix 9381: set correct color for brackets in dark mode

duanjr:
Resolves: #9381 

This issue occurs because class `Bracket` uses `BrackItem` to store some properties, which is different from many other `EngravingItem`.
When we set the color of the bracket in `PaletteCellIconEngine::paintPaletteElement` through `element->setProperty`, we actually set it for `BrackItem` instead of `Bracket` itself. So when the `draw` function ask for color it only gets the default color for light mode.

### DIFF
--- a/src/engraving/libmscore/bracket.h
+++ b/src/engraving/libmscore/bracket.h
@@ -109,6 +109,8 @@ public:
     void editDrag(EditData&) override;
     void endEditDrag(EditData&) override;
 
+    mu::draw::Color color() const override { return _bi->color(); }
+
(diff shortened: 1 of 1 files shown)


duanjr:
image

RomanPudashkin (approved):

boryskuzmenko:
https://github.com/musescore/MuseScore/issues/9381 bugfix is nverified
OS: macOS, Windows
**FIXED**

Please, merge to the Master.

PR #11074, opened 2022-04-03
Simplify finding SndFile package using PkgConfig

cbjeukendrup:
Also seems to fix finding the correct one if you need the Intel version on an ARM Mac when you have both installed using HomeBrew.


cbjeukendrup:
Ah, that's a pity, this does not work on Linux, for a reason that is unknown to me. Unless someone could pick this up for me (I have no Linux at the moment), I'm afraid we'll have to close this.

AntonioBL:
Ciao.
Under Linux, with cmake 3.16.3, `pkg_check_modules` does not seem to fill in the variables `LIBSNDFILE_PKGCONF_INCLUDE_DIRS` and `LIBSNDFILE_PKGCONF_LIBRARY_DIRS` but those paths can be found instead in the variables `LIBSNDFILE_PKGCONF_INCLUDEDIR` and `LIBSNDFILE_PKGCONF_LIBDIR`
The help page for `pkg_check_modules` lists them as well, but I do not understand why the other variables (<>_LIBRARY_DIRS and <>_INCLUDE_DIRS) are left empty
https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules

PR #11090, opened 2022-04-04
[MU4] Slurs only adjust for slurs on their staff, and endpoint adjustment fix

asattely:
There were a few bugs in the slur adjustment code that caused slurs to adjust for slurs not on their system. This has been fixed. There was also a bit of a bug in the slur endpoint avoidance code which has been fixed now as well.

There still seems to be a strange issue with accidentals not being taken into account in the collision detection code (which this PR or the previous slur PR never touched), and I can't help but think that this issue didn't exist when the last slur revamp was written. So the shapewise collision stuff for accidentals still needs to be looked at a bit closelier in the future. This PR won't deal with that however!

### DIFF
--- a/src/engraving/libmscore/slur.cpp
+++ b/src/engraving/libmscore/slur.cpp
@@ -313,7 +313,7 @@ void SlurSegment::adjustEndpoints()
     // point 1
     int lines = staff()->lines(tick());
     auto adjustPoint = [staffLineMargin](bool up, qreal ysp) {
-        qreal y1offset = ysp - round(ysp);
+        qreal y1offset = ysp - floor(ysp);
         qreal adjust = 0;
         if (up) {
             if (y1offset < staffLineMargin) {
@@ -353,6 +353,9 @@ void SlurSegment::computeBezier(mu::PointF p6o)
     qreal _spatium  = spatium();
     qreal shoulderW;                // height as fraction of slur-length
     qreal shoulderH;
+    if (autoplace()) {
+        adjustEndpoints();
+    }
     //
     // pp1 and pp2 are the end points of the slur
     //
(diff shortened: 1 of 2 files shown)


asattely:
vtest failures are expected, I believe (@oktophonie can confirm or deny)

oktophonie:
The endpoints of slurs are not avoiding ties any more, when crossing a system:

slur-tie-collision

oktophonie (approved):

src/engraving/libmscore/slur.cpp
```diff
@@ -1426,15 +1426,11 @@ SpannerSegment* Slur::layoutSystem(System* system)
             Chord* c = toChord(cr);
             Tie* tie = nullptr;
             PointF endPoint;
-            if (c->notes()[0]->tieFor() && !c->notes()[0]->tieFor()->isInside() && c->notes()[0]->tieFor()->up() == _up) {
-                // there is a tie that starts on this chordrest
-                tie = c->notes()[0]->tieFor();
-                endPoint = tie->segmentAt(0)->ups(Grip::START).pos();
-            }
-            /*if (c->notes()[0]->tieBack() && !c->notes()[0]->tieBack()->isInside() && c->notes()[0]->tieBack()->up() == _up) {
+            if (c->notes()[0]->tieBack() && !c->notes()[0]->tieBack()->isInside() && c->notes()[0]->tieBack()->up() == _up) {
```
RomanPudashkin, line 1429: Please, add a temp variable for c->notes()[0]->tieBack() instead of constantly copy-pasting it

asattely:
Good call! The changes are pushed.

RomanPudashkin (approved):

Issue #11023, opened 2022-03-31
[MU4 Issue] "Disable Vertical justification of staves" setting is being ignored in imported MS3 files


rgreen5:
**Describe the bug**

It seems that MS4 is ignoring the "Disable vertical justification of staves" setting which was previously applied to MS3 files and applying automatic vertical justification instead.

**To Reproduce**

1. Open the attached score. 
[staves_justification_issue.mscz.zip](https://github.com/musescore/MuseScore/files/8388993/staves_justification_issue.mscz.zip)

2. Go to Format > Style > Page. Notice that the "Enable vertical justification of staves" option is ticked. However, this was not the option specified option in MS3 which was "Disable …". This has clearly spoilt the layout of the score.
3. Open the same score in MS3 and notice the correct layout. Note also the "Disable vertical justification of staves" option is selected in Format > Style > Page.

https://user-images.githubusercontent.com/90187801/161567314-827a464c-396c-4531-aff1-77c03ed91e99.mp4

**Expected behavior**
For imported MS3 files the correct staff justification option should be carried over into MS4.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2059995262, revision: github-musescore-musescore-cafefd7
- occurs on Mac and Win as well

DmitryArefiev:
Attached a video demo to the description

Jojo-Schmitz:
In 3.6 "Enable vertical spread" was the default, hard coded in `MuseScore::init()`.

PR #11091, opened 2022-04-04
[MU4] Change default for enableVerticalSpread to be in line with 3.6

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/11023

The 302 default for `enableVerticalSpread` was false, so the 302 defaults file had to be changed to be in line with it.

### DIFF
--- a/src/engraving/data/styles/legacy-style-defaults-v302.mss
+++ b/src/engraving/data/styles/legacy-style-defaults-v302.mss
@@ -17,7 +17,7 @@
     <akkoladeDistance>6.5</akkoladeDistance>
     <minSystemDistance>8.5</minSystemDistance>
     <maxSystemDistance>15</maxSystemDistance>
-    <enableVerticalSpread>1</enableVerticalSpread>
+    <enableVerticalSpread>0</enableVerticalSpread>
     <spreadSystem>2.5</spreadSystem>
     <spreadSquareBracket>1.2</spreadSquareBracket>
     <spreadCurlyBracket>1.1</spreadCurlyBracket>


oktophonie (approved):

oktophonie:
The default (as defined in `style.cpp`) in MuseScore 3.6 (= file format version 302) is indeed `false`, so this fix is correct. However, it becomes set to `true` at some point for all new scores created in 3.6, which is a bit confusing. I'm not sure now whether this was a deliberate decision or is just an oddity.

Jojo-Schmitz:
It is explicitly set to `true` on start of MuseScore (in 3.x), if not in test or template mode:
```
      // initialize current page size from default printer
#ifndef QT_NO_PRINTER
      if (!MScore::testMode && !MScore::saveTemplateMode) {
            QPrinter p;
            if (p.isValid()) {
//                  qDebug("set paper size from default printer");
                  QRectF psf = p.paperRect(QPrinter::Inch);
                  MScore::defaultStyle().set(Sid::pageWidth,  psf.width());
                  MScore::defaultStyle().set(Sid::pageHeight, psf.height());
                  MScore::defaultStyle().set(Sid::enableVerticalSpread, true);
                  }
            }
#endif
```
Since 4916486, MuseScore 3.6 Alpha

oktophonie:
You mean moot, I guess ;)

The point is that only the settings that differ from the defaults given in `style.cpp` (now `styledef.cpp`) are saved in the file. So, if the user disables vertical spread in the 3.6 file, that property will _not_ be explicitly included, but if they leave it set to enabled (which is what they will generally be presented with for a new file, for the reasons you mention), it _will_ be included, since that's technically not the default.

When it comes to migration, we need to reference a style sheet that contains all the defaults for the previous MU versions (as taken from `style.cpp`), and the default in 3.6/302 is indeed `false`, so this PR does make sense, though I can't deny it's all a bit confusing.

Jojo-Schmitz:
Yes, I did. I noticed that it looked wrong but couldn't spot the error ;-)

And yes, I see now. That option not being set in 302 means disabled. So the PR seems correct.

RomanPudashkin (approved):

Issue #9960, opened 2021-12-02
[MU4 Issue] Cross staff beaming is screwed up


worldwideweary:
**Describe the bug**
Cross-staff beams need some work.
Not sure if this is just partially implemented and left hangin' or was screwed up in a previous attempt to be thorough

**To Reproduce**
Use cross-staff shortcut upon 8th notes from bass to treble clef:

![MS-crossstaffbeam](https://user-images.githubusercontent.com/7139517/144519056-c13b8c9e-67e9-498e-898b-f1df0bd2d6cf.png)

Obviously should be like MS3
![MS3-crossstafbeam](https://user-images.githubusercontent.com/7139517/144519321-9052bafb-3918-44f7-ada5-bdf5ff736e97.png)
:

oktophonie:
This is a consequence of #9792 (as the description there mentions, cross-staff beaming remains to be implemented). This will be fixed shortly as part of this ongoing overhaul.

worldwideweary:
Ok cool. Unless suggested otherwise, I'll keep this open as a separate issue for now instead of closing it so that after #9792 is finalized, the cross-staff beaming can be tested in and of itself (or if there's anything else funny going on with it afterwards).

PR #11092, opened 2022-04-04
[MU4] Cross staff beams

asattely:
(partially) Resolves: https://github.com/musescore/MuseScore/issues/9960

Cross-staff beams now function almost perfectly, but there is one problem with manually-adjusted cross staff beams. That is a pretty niche issue so it can be fixed in a future PR; the rest of the stuff in this PR is pretty high priority so it needs to go in as soon as possible.

There is also an issue with cross-staff slurs, which will be addressed soon!

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -436,6 +436,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 firstMeasure = false;
             }
             mb->setPos(pos);
+            mb->setParent(system);
             Measure* m = toMeasure(mb);
             m->layoutMeasureElements();
             m->layoutStaffLines();
@@ -456,6 +457,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
 
     layoutSystemElements(options, ctx, score, system);
     system->layout2(ctx);     // compute staff distances
+    for (MeasureBase* mb : system->measures()) {
+        mb->layoutCrossStaff();
+    }
     // TODO: now that the code at the top of this function does this same backwards search,
     // we might be able to eliminate this block
     // but, lc might be used elsewhere so we need to be careful
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -374,9 +374,9 @@ void Beam::layout1()
     if (_direction != DirectionV::AUTO) {
         _up = _direction == DirectionV::UP;
     } else if (_maxMove > 0) {
-        _up = true;
-    } else if (_minMove < 0) {
         _up = false;
+    } else if (_minMove < 0) {
+        _up = true;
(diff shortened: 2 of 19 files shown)


src/engraving/libmscore/beam.cpp
```diff
@@ -917,8 +921,9 @@ void Beam::offsetBeamToRemoveCollisions(std::vector<ChordRest*> chordRests, int&
         }
         Chord* chord = toChord(chordRest);
         PointF anchor = chordBeamAnchor(chord);
+
         qreal proportionAlongX = (anchor.x() - startX) / (endX - startX);
-        while (true) {
+        while (true && proportionAlongX) {
```
cbjeukendrup, line 926: I have some remarks about this line:
- `true && ` could be removed
- the qreal `proportionAlongX` might not become exactly 0, but maybe something like 0.00000001. We often use `RealIsNull` from realutils.h for this. 

Also, it seems that proportionAlongX is not modified inside the loop. Therefore, it might be confusing to use it as a loop condition. Maybe something like this is more suited:
```c++
if (RealIsNull(proportionAlongX)) {
    continue;
}

while (true) {
    ...
```

src/engraving/layout/layoutsystem.cpp
```diff
@@ -431,6 +431,7 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
                 firstMeasure = false;
             }
             mb->setPos(pos);
+            mb->setParent(system);
```
vpereverzev, line 434: why would we need this? Does it mean that we're getting measures from system->measures() without any parent?

asattely: This PR is not ready to merge yet and I've been spending the past few days trying to clean it up a little bit. That having been said, yes, measures don't have an explicitly set parent as they're being laid out (because we don't know what system the measure will be in until layout happens) but we need to set the parent as soon as possible once we know, because cross-staff stuff needs to know things like distance between staves in a system etc.

That's my understanding, at least, after talking to some people who did work on this before. Again this one is still a work in progress and should have something at least functional by the end of the week

src/engraving/libmscore/beam.cpp
asattely: a very good point! I don't know how that slipped in there but that's very bush league of me. It'll be fixed in the next push

asattely:
Most of the vtest differences seem to be improvements, though a lot of them appear to be due to the rebase I did rather than anything that this PR changes. There are still a few little graphical glitches here and there, which I'm looking into now

asattely:
fixed some vtests, squashed, as long as nothing went pearshaped this is ready for review, testing, and merge!

oktophonie:
In the interests of ever getting this increasingly complex PR merged, the issues with the couple of problematic vtests (beams-12-1 and cross-2-1, which pertain to manual adjustments) are to be fixed separately.

oktophonie (approved):

Issue #8143, opened 2021-05-23
[MU4 Issue] Cannot load MuseScore3 score


njvdberg:
**Describe the bug**
When loading [Test..._Beethoven__Symphony_No._9__Op._125.mscz](https://musescore.com/openscore/scores/5733014) I had to abort after 2 hours running with 100% CPU usage (just one single core). In there 2 hours there was only some logging for the first 2 minutes:
```
19:10:09.103 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: file-open
19:10:17.270 | WARN  | main_thread | Qt         | libpng warning: iCCP: CRC error
19:10:23.259 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.289 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.290 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.291 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.291 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.292 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.292 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.293 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.293 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:23.294 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.447 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.447 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.447 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.447 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.448 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.448 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.448 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.448 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.448 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.448 | WARN  | main_thread | Qt         | unmapped drum note 46
19:10:40.628 | WARN  | main_thread | Qt         | unmapped drum note 46
[...]

PR #4784, opened 2019-03-08
fix #285591: accidentals should tuck under notes

MarcSabatella:
The intent of this is to fix https://musescore.org/en/node/285591, and I'm coming close but running into issues.  So I'm submitting this PR as a work in progress, hoping maybe someone else can see a way out.

The good news is that my initial analysis was correct, simply leaving accidentals and other elements out of _spaceLw and _spaceRw *does* solve the problem.  It turns out we can't just remove the code calculating those variables because this is also the code used to *position* the symbols attached to chords.  But I was able to separate out the positioning from the allocation of space easily enough.  There is a slight complication with grace notes (which use the same code recursively), but nothing we shouldn't be able to solve.  I was able to get the following with little effort:

image

This is much better than 3.0.4 (which is similar to both 2.3.2 and current master):

image

Unfortunately, there is a problem with certain beamed note groups where the final stem direction and length is not known until long after segment spacing has been finalized.  In these cases, it is quite possible that at the point we are calculating the minimum distance between segment shapes, we might decide it is safe to tuck an accidental above a note, but it turns out to create overlap with the previous stem once its final direction and length are determined.  Thus we get the following:

image

This was the case in earlier 3.0 builds as well, we just didn't notice it (or maybe someone did, but it never got addressed).

The only solution I see is to figure out the beaming earlier in the layout process, but then there are also reasons why it needs to be as late as it does, so I don't see a way out.

Maybe someone else wants to take it from here?

PR #11099, opened 2022-04-05
Collisions overhaul (and other improvements to horizontal spacing)

mike-spa:
This PR rewrites a substantial part of horizontal spacing, particularly the way we perform collision checks. Most notably (and _finally_) now items can tuck (or kern) above/below each other.

In more detail, what I did is:
1. Removed the logic within the code that was preventing elements from kerning. 
2. Implemented a system which calculates the spacing at "high resolution", meaning it considers every single graphical piece of an item (notehead, stem, hook, ledger lines, etc...).
3. Managed several exceptions and special cases related to kerning (basically, cases when items should _not_ kern).
4. Refactored the interaction of all the user-defined spacing settings (specifically those from Style -> Measure).
5. Corrected a small error related to the spacing of polyrythms.
6. Corrected mid-system clefs positioning.
7. Implemented a smarter way of deciding how many measures fit in one system. (I had tried this already in previous PRs but it was breaking on edge cases. Now it is finally stable.)

A side benefit of allowing items to tuck is that we save a huge amount of horizontal space that we were previously wasting, which lets us:
8. Increase the default value of minNoteDistance to a much more reasonable 0.6 spaces. 
This is done in a separate commit, as it requires to also change all template files.

Basically, every single visual test is expected to fail, and I've usually had also utests failing on beams, so that's also to be expected.

@MarcSabatella if I may bother you, it'd be great if you could link some of the old issues that will be fixed by this (one of which you mentioned already). Thanks for the amazing support!

Before:
Screenshot from 2022-04-01 12-35-00

After:
Screenshot from 2022-04-01 12-36-27

Before:
Screenshot from 2022-04-05 17-13-49
Screenshot from 2022-04-05 17-12-48

After:
Screenshot from 2022-04-05 17-05-42


MarcSabatella:
This is fantastic!  Happy to update the issues once this is merged.  But, I did want to verify something.  When I took a crck at fixing this a couple of years ago, the issue I ran into was that at the point where we were calculating the spacing requirements for a given note, we didn't yet know for sure if the previous note was upstem or downstem if it was beamed.  That's because beam directions weren't computer until later.  So in any case where the initially guessed stem direction was up, but the actual stem direction turned out to be down, we would make the wrong call regarding the tucking.  See for instance my examples in https://github.com/musescore/MuseScore/pull/4784

I can totally believe this might not be an issue any \more because the beam code has been rewritten.  But I wouldn't mind hearing confirmation.

src/engraving/layout/layoutsystem.cpp
```diff
@@ -168,14 +168,14 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
...
+
         // check if lc.curMeasure fits, remove if not
         // collect at least one measure and the break
-
-        // acceptanceRange slightly larger than 1 allows systems to be initially slightly larger than the target width
-        // and be justified by squeezing rather than stretching. However, we must first make sure that the system *can*
-        // be squeezed. I'm temporarily rolling back the idea (still a bit too risky in some edge cases) [M.S.]
-        double acceptanceRange = 1;
-        bool doBreak = (system->measures().size() > 1) && ((curSysWidth + ww) > systemWidth * acceptanceRange);
+        double acceptanceRange = 0.3 * system->squeezableSpace();
```
Marr11317, line 174: Where does the 0.3 come from?

src/engraving/layout/layoutmeasure.cpp
```diff
@@ -745,6 +745,26 @@ void LayoutMeasure::getNextMeasure(const LayoutOptions& options, LayoutContext&
     }
 
     LayoutBeams::createBeams(score, ctx, measure);
+    /* HACK: The real beam layout is computed at much later stage (you can't do the beams until you know
```
Marr11317, line 748: @MarcSabatella This may be the answer to your question

Marr11317:
Just to be sure: are these vtests correct?

Lyrics 6-1
image

Here the note order changed. Is it expected?
image

changed file?
image

No lyrics under timesig: 
image

(multiple of these)
image

weird
image

and 
image

I have probably missed some more.

Marr11317:
One tiny question: Is there any performance overhead for this more complex layout algorithm?

mike-spa:
> This is fantastic! Happy to update the issues once this is merged. But, I did want to verify something. When I took a crck at fixing this a couple of years ago, the issue I ran into was that at the point where we were calculating the spacing requirements for a given note, we didn't yet know for sure if the previous note was upstem or downstem if it was beamed. That's because beam directions weren't computer until later. So in any case where the initially guessed stem direction was up, but the actual stem direction turned out to be down, we would make the wrong call regarding the tucking. See for instance my examples in #4784
> 
> I can totally believe this might not be an issue any \more because the beam code has been rewritten. But I wouldn't mind hearing confirmation.

@MarcSabatella stems were a real head scratcher. It turned out that now direction _was_ known at the point of horizontal spacing, but _extension_ was not. And yes, @Marr11317 pointed out the answer. Not an elegant solution at all (and I may look for a better one in future) but one that seems to do the job.

src/engraving/layout/layoutsystem.cpp
mike-spa: Nothing, it just looks good. May be worth a comment though, thanks.

mike-spa:
@Marr11317 Clefs/timesigs/keysigs are an exception to kerning, meaning they shouldn't kern above/below other items. Lyrics are the exception to the exception to kerning. Gonna look into that.

src/engraving/libmscore/score.cpp
```diff
@@ -5401,6 +5405,160 @@ void Score::doLayoutRange(const Fraction& st, const Fraction& et)
...
+void Score::createPaddingTable()
+{
+    double minimumPaddingUnit = 0.1 * spatium();
+    for (int i=0; i < int(ElementType::MAXTYPE); ++i) {
+        for (int j=0; j < int(ElementType::MAXTYPE); ++j) {
+            _paddingTable[ElementType(i)][ElementType(j)] = minimumPaddingUnit;
+        }
+    }
+
+    const double ledgerPad = 0.25 * spatium();
```
Marr11317, line 5417: should'nt that ledger padding come from the score's configured staff weight?

src/engraving/libmscore/score.cpp
```diff
@@ -5401,6 +5405,160 @@ void Score::doLayoutRange(const Fraction& st, const Fraction& et)
...
+    const double ledgerLength = styleMM(Sid::ledgerLineLength);
+
+    /* NOTE: the padding value for note->note is NOT minNoteDistance, because minNoteDistance
+     * should only apply to notes of the same voice. Notes from different voices should be
+     * allowed to get much closer. So we set the general padding at minimumPaddingUnit,
+     * but we introduce an appropriate exception for same-voice cases in Shape::minHorizontalDistance().
+     */
+    _paddingTable[ElementType::NOTE][ElementType::NOTE] = minimumPaddingUnit;
+    _paddingTable[ElementType::NOTE][ElementType::LEDGER_LINE] = 0.35 * spatium();
```
Marr11317, line 5426: dito

src/engraving/libmscore/score.h
```diff
@@ -513,6 +513,9 @@ class Score : public QObject, public EngravingObject
     void assignIdIfNeed(Staff& staff) const;
     void assignIdIfNeed(Part& part) const;
 
+    PaddingTable _paddingTable;
+    void createPaddingTable();
```
Marr11317, line 517: is the padding table updated when the user changes score properties?

src/engraving/libmscore/score.cpp
```diff
@@ -5401,6 +5405,160 @@ void Score::doLayoutRange(const Fraction& st, const Fraction& et)
     }
 }
 
+void Score::createPaddingTable()
+{
+    double minimumPaddingUnit = 0.1 * spatium();
```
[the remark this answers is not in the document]
oktophonie: This is a hard limit on the distance that two items can come within; it only applies in certain special situations, e.g. polyrhythms between notes in two voices where the normal `minNoteDistance` does not apply. 0.01 would be too small, but like all the other settings this is to become user-configurable eventually.

src/engraving/libmscore/score.cpp
```diff
@@ -5401,6 +5405,160 @@ void Score::doLayoutRange(const Fraction& st, const Fraction& et)
...
+        elem.second = _paddingTable[ElementType::NOTE][elem.first];
+    }
+    for (auto& elem: _paddingTable) {
+        elem.second[ElementType::STEM] = _paddingTable[elem.first][ElementType::NOTE];
+    }
+    _paddingTable[ElementType::STEM][ElementType::NOTE] = styleMM(Sid::minNoteDistance);
+    _paddingTable[ElementType::STEM][ElementType::STEM] = 0.85 * spatium();
+    _paddingTable[ElementType::STEM][ElementType::ACCIDENTAL] = 0.35 * spatium();
+    _paddingTable[ElementType::STEM][ElementType::LEDGER_LINE] = 0.35 * spatium();
+    _paddingTable[ElementType::LEDGER_LINE][ElementType::STEM] = 0.35 * spatium();
```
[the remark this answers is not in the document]
oktophonie: The magic numbers emanate from me (@mike-spa working here under my brutal and unforgiving instruction).

In the end we want all the values of the padding table to be user-configurable. We can't do that right now as there's a high design cost and it will need a lot of thought to work out how to expose them all, since there are so many values and potential interactions between them.

For now we're taking the settings that do already exist (in the Style -> Measure/Bar dialog) and using them in the relevant slots in the table, and then deriving other values from them where possible. For the remainder I have supplied defaults for the time being, which aren't arbitrary.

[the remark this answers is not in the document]
Marr11317: Sounds good then. It can always be made configurable when users ask for it I guess.

oktophonie (approved):

Jojo-Schmitz:
"8 - engraving_utests (SEGFAULT)"

mike-spa:
> "8 - engraving_utests (SEGFAULT)"

Yeah I saw that. I never had that kind failure: how can I replicate that specific test on my machine to debug it?

Jojo-Schmitz:
Not sure at all. There are other failures too, diffs. Maybe those in the end result in a crash?

mike-spa:
Yeah I'm now going through the diffs and fix those, then I'll force push again and let's see what happens

Jojo-Schmitz:
utests still fail, but no longer crash

mike-spa:
@igorkorsukov I got a compiler error after the last rebase related to a QList vs std conflict. I hope I've fixed it correctly

src/engraving/rw/scorereader.cpp
```diff
@@ -140,6 +140,10 @@ Err ScoreReader::loadMscz(Ms::MasterScore* masterScore, const mu::engraving::Msc
         }
     }
 
+    // These are necessary to be sure that ALL the layout is correctly updated when loading any file
+    masterScore->styleChanged();
+    masterScore->update();
+
```
Eism, line 146: @mike-spa can we remove this block of code?
This code takes a lot of time when opening a big score.
You can test on the score from this PR https://github.com/musescore/MuseScore/issues/8143

This code does not solve all the problems with the long opening, but it significantly reduces the time.
And the comment says that this code may not be executed, we just make sure.

mike-spa: @Eism I had to do that because many scores were not loading the new spacing settings correctly, so for now we can't just remove the code, unfortunately. But there is probably a smarter way to do it, I can try to look into that. Just for my understanding: which one of them is more time consuming? `styleChanged()`or `update()`?

Side note: #8143 was opened before this PR was merged, so there must be also something else going on, I guess.

Eism: Score::update

image

mike-spa: Excellent, thank you. I'll try to look into it as soon as I can!

PR #11100, opened 2022-04-05
Fix #10964: Editing Tab Staff types is reflected in score

### LINKED ISSUE #10964: [MU4 Issue] Tablature Guitar "Edit Staff Type" menu not saving
**Describe the bug**
The "Fret Marks" and "Note Values" sections of "Edit Staff Type" for guitars doesn't save settings. 

**To Reproduce**
1. Create a new score
2. Add a guitar with tablature (e.g., "Electric Guitar (tablature)")
3. Select a measure and right-click
4. Click "Staff/part properties..."
5. Under "Staff properties" click "Advanced style properties..."
6. Make a change under "Fret Marks" or "Note Values"
7. Click "OK"
8. Click "Apply"
9. Click "OK"
10. Repeat 3-5
11. Observe no settings under "Fret Marks" or "Note Values" was saved

**Expected behavior**
Settings inside of the "Fret Marks" and "Note Values" should be saved.

**Screenshots**
image

**Platform information**
Windows 10

**Additional context**
All settings inside of the "Edit Staff Type" window will save except items under "Fret Marks" and "Note Values". Setting a template does not keep the set template inside the dropdown after clicking "OK", unsure if this is the intended behavior though. 

Only tested on Electric and Acoustic tablature guitars. Issue doesn't occur for instruments without this menu (e.g., Piano).

miiizen:
Resolves: #10964

### DIFF
--- a/src/notation/view/widgets/editstaff.cpp
+++ b/src/notation/view/widgets/editstaff.cpp
@@ -470,26 +470,15 @@ void EditStaff::applyStaffProperties()
 {
     StaffConfig config;
     config.visible = m_orgStaff->visible();
-    config.linesColor = color->color();
-    config.visibleLines = invisible->isChecked();
+
     config.userDistance = spinExtraDistance->value() * m_orgStaff->score()->spatium();
-    config.scale = mag->value() / 100.0;
-    config.isSmall = isSmallCheckbox->isChecked();
     config.cutaway = cutaway->isChecked();
     config.showIfEmpty = showIfEmpty->isChecked();
-    config.linesCount = lines->value();
-    config.lineDistance = lineDistance->value();
-    config.showClef = showClef->isChecked();
-    config.showTimeSignature = showTimesig->isChecked();
-    config.showKeySignature = editStaffTypeDialog->getStaffType().genKeysig();
-    config.showBarlines = showBarlines->isChecked();
-    config.showStemless = editStaffTypeDialog->getStaffType().stemless();
-    config.showLedgerLinesPitched = editStaffTypeDialog->getStaffType().showLedgerLines();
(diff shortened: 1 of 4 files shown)


src/notation/view/widgets/editstaff.cpp
```diff
@@ -492,6 +492,7 @@ void EditStaff::applyStaffProperties()
     config.clefTypeList = m_instrument.clefType(m_orgStaff->rstaff());
 
     notationParts()->setStaffConfig(m_orgStaff->id(), config);
+    m_orgStaff->setStaffType(Fraction(0, 1), editStaffTypeDialog->getStaffType());
```
RomanPudashkin, line 495: Unfortunately, here we need a fix of a different kind

1. we must change the staff type only by score()->undo(new Ms::ChangeStaffType(staff, newStaffType)) (i.e. in NotationParts::doSetStaffConfig). Otherwise it will be impossible to undo these changes (ctrl + z)

2. it seems that StaffConfig partially duplicates Ms::StaffType. I would suggest removing these duplicated fields from StaffConfig and adding Ms::StaffType as its member

Thus, we can pass the changed staff type directly to NotationParts and apply it using score()->undo(new Ms::ChangeStaffType(...))

miiizen: My apologies, thought the solution looked a bit naive!  Thanks for the guidance, I've implemented that.

RomanPudashkin: Thank you! Much better now! But you should rebase your PR: there should be only one commit

RomanPudashkin (approved):

DmitryArefiev:
Tested #10964 on Win10, Mac12, LinuxMint 20.2 - **FIXED**      

Please merge into master

PR #11102, opened 2022-04-06
[MU4] fix #9356: Set the correct magnification on icons in custom palettes

### LINKED ISSUE #9356: [MU4 Issue] Custom palette icons zoomed in
**Describe the bug**
Custom palette icons zoomed in

**To Reproduce**
1. Create a score
2. Click Left side "Palettes" panel
3. Click Add Palettes
4. Click Create Custom palette
5. type name
6.  Click "Create"
7. Expand one of the existing Palettes list
8. Drag & drop palette from expanded list to the Custom palette list

**Expected behavior**
Palettes icons should be the same size in Custom list as in other lists

**Screenshots**

https://user-images.githubusercontent.com/71372310/135734460-2d914784-d50a-49e5-a476-440d9406bbc2.mp4

Windows 
Linux
MacOS

helloworld12321:
Resolves: <https://github.com/musescore/MuseScore/issues/9356>

In the "Palettes" sidebar, some icons are zoomed in or out to make them more legible. (For example, clefs are shown at 0.8x magnification; fingering indicators are shown at 1.5x magnification.)

Previously, items in custom palettes were _not_ being scaled. As a result, clefs in custom palettes looked very large, and fingering indicators in custom palettes looked very small.

This PR fixes that behavior; now, when an item is added to any palette, it is given a magnification value according to what type of element it is.

**Caveats:**

- You may need to click the "reset palette" button in order for these changes to take effect.

- ~~Some elements—in particular, action icons—still appear slightly too large in custom palettes, for reasons I don't entirely understand.~~ [Fixed]

- Some palettes have cells with varying magnifications. Previously, the logic to accomplish this was spread between `palettecreator.cpp` and `palette.cpp`. I've consolidated this logic in `palette.cpp` in the `scaleCell()` method.

- Previously, some tempo text icons were displayed at 0.975x scale, and others were displayed at 0.845x scale. To keep the logic in `scaleCell()` simple, I split the difference and set them both to 0.9x scale.

### DIFF
--- a/src/palette/internal/palette.cpp
+++ b/src/palette/internal/palette.cpp
@@ -45,6 +45,10 @@ using namespace mu;
 using namespace mu::palette;
 using namespace Ms;
 
+static Palette::Type guessTypeFrom(const Ms::EngravingItem* e);
+static void scaleCell(PaletteCellPtr cellPtr);
+static qreal appropriateMagnificationFor(const Palette::Type& type);
+
 Palette::Palette(Type t, QObject* parent)
     : QObject(parent), m_type(t)
 {
@@ -104,11 +108,7 @@ PaletteCellPtr Palette::insertElement(size_t idx, ElementPtr element, const QStr
     }
 
     PaletteCellPtr cell = std::make_shared<PaletteCell>(element, name, mag, tag, this);
-
-    auto cellHandler = cellHandlerByPaletteType(m_type);
-    if (cellHandler) {
-        cellHandler(cell);
-    }
+    scaleCell(cell);
 
     m_cells.emplace(m_cells.begin() + idx, cell);
     return cell;
@@ -122,11 +122,7 @@ PaletteCellPtr Palette::appendElement(ElementPtr element, const QString& name, q
     }
 
     PaletteCellPtr cell = std::make_shared<PaletteCell>(element, name, mag, tag, this);
-
-    auto cellHandler = cellHandlerByPaletteType(m_type);
-    if (cellHandler) {
-        cellHandler(cell);
-    }
+    scaleCell(cell);
 
     m_cells.emplace_back(cell);
     return cell;
@@ -154,6 +150,7 @@ bool Palette::insertCells(size_t idx, std::vector<PaletteCellPtr> cells)
 
     for (PaletteCellPtr& c : cells) {
         c->setParent(this);
+        scaleCell(c);
     }
 
     m_cells.insert(m_cells.begin() + idx, std::make_move_iterator(cells.begin()),
@@ -274,11 +271,7 @@ bool Palette::read(XmlReader& e)
                 continue;
             }
 
-            auto cellHandler = cellHandlerByPaletteType(m_type);
-            if (cellHandler) {
-                cellHandler(cell);
-            }
-
+            scaleCell(cell);
             m_cells.push_back(cell);
         } else {
             e.unknown();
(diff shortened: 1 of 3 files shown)


helloworld12321:
Before:

https://user-images.githubusercontent.com/56209343/161885136-2015336f-47f6-4733-b46f-67f4408475b4.mp4

After:

https://user-images.githubusercontent.com/56209343/161885159-5a11d252-3d4e-4d5f-ae44-895b73dd2adf.mp4

helloworld12321:
Action icons are still a bit too large: 🙁

Screenshot from 2022-04-05 21-09-02

src/palette/internal/palette.h
```diff
@@ -161,8 +161,10 @@ class Palette : public QObject
     void showWritingPaletteError(const QString& path) const;
 
     Type guessType() const;
+    static Type guessTypeFrom(const Ms::EngravingItem* e);
```
RomanPudashkin, line 164: I would suggest removing these definitions from this header. You can declare all the static functions at the top of the palette.cpp file

RomanPudashkin (approved):

src/palette/internal/palette.h
helloworld12321: Done, thank you! :slightly_smiling_face:

boryskuzmenko:
Hi, @helloworld12321 
Thank you for your fix! But some of elements still changes their scale when adding to a custom palette. Due to different element cell dimensions. Could you please fix them too.

Capture 2022-04-08_01-07-28_PM

src/palette/internal/palette.cpp
```diff
@@ -515,102 +508,143 @@ Palette::Type Palette::guessType() const
...
-        return Type::Bracket;
+        return Palette::Type::Bracket;
     case ElementType::ACCIDENTAL:
-        return Type::Accidental;
+        return Palette::Type::Accidental;
     case ElementType::ARTICULATION:
+        if (toArticulation(e)->isOrnament()) {
+            return Palette::Type::Ornament;
+        } else if (toArticulation(e)->isLuteFingering()) {
+            return Palette::Type::Fingering;
```
helloworld12321, line 536: Lute fingerings are implemented as articulations, but belong to the "fingerings" palette.

src/palette/internal/palettecreator.cpp
```diff
@@ -231,7 +231,6 @@ PalettePtr PaletteCreator::newKeySigPalette()
 {
     PalettePtr sp = std::make_shared<Palette>(Palette::Type::KeySig);
     sp->setName(QT_TRANSLATE_NOOP("palette", "Key signatures"));
-    sp->setMag(0.8);
```
helloworld12321, line 234: We're handling element-scaling on a per-cell basis now, rather than a per-palette basis.

The user can still set a custom magnification on a palette, if, say, they want the whole "Articulations" palette to be scaled up. This composes nicely with the per-cell element scaling.

src/palette/internal/palette.cpp
```diff
@@ -515,102 +508,143 @@ Palette::Type Palette::guessType() const
...
+    case Palette::Type::Clef:
+    case Palette::Type::BagpipeEmbellishment:
+    case Palette::Type::Line:
+    case Palette::Type::TimeSig:
+        return 0.8;
+    case Palette::Type::Repeat:
+    case Palette::Type::Text:
+        return 0.85;
+    case Palette::Type::Tempo:
+        return 0.9;
```
helloworld12321, line 640: Previously, different cells in the "tempo" palette had different magnifications (some were ≈0.98, some were ≈0.85). A magnification of 0.9 is a compromise between the previous values.

src/palette/internal/palette.cpp
```diff
@@ -515,102 +508,143 @@ Palette::Type Palette::guessType() const
...
-            if (bracket->bracketType() == BracketType::BRACE) {
-                bracket->setStaffSpan(0, 1);
-                cellPtr->mag = 1.2;
-            }
-        };
+qreal appropriateMagnificationFor(const Palette::Type& type)
+{
+    switch (type) {
+    case Palette::Type::Bracket:
+        return 0.6;
```
helloworld12321, line 628: Previously, brackets had a magnification of 0.7. I made them slightly smaller by default so that they don't get cropped when added to custom palettes.

src/palette/internal/palette.cpp
```diff
@@ -515,102 +508,143 @@ Palette::Type Palette::guessType() const
...
         }
     }
     }
 
-    return Type::Custom;
+    return Palette::Type::Custom;
 }
 
-std::function<void(PaletteCellPtr)> Palette::cellHandlerByPaletteType(const Palette::Type& type) const
+void scaleCell(PaletteCellPtr cellPtr)
```
helloworld12321, line 601: No need to return a lambda; we can just operate on a `PaletteCellPtr` directly! :relaxed:

helloworld12321:
@boryskuzmenko Ready for re-review—thanks for your patience! :smiley:

boryskuzmenko:
Hi, @helloworld12321 
The fix can't be tested properly due to regression. Some of elements can't be added to custom palette.
Please, investigate and fix

https://user-images.githubusercontent.com/101250347/164703599-6fc47dd6-bd1a-4f77-a128-7731f42665da.mov

helloworld12321:
@boryskuzmenko Oh dear, my apologies. I'll look into the regression.

Thanks for your review!

helloworld12321:
Closing, since

1. I wasn't able to track down @boryskuzmenko's regression—I couldn't reproduce it locally.
2. Unfortunately, I just moved to a new city and no longer have the box I was doing development on.

Very sorry for not following through with this! 🙁

PR #6248, opened 2020-06-22
Refactor score unwinding

jeetee:
Resolves: issues with Jump/repeat combinations: [#287447](https://musescore.org/en/node/287447), [#299320](https://musescore.org/en/node/299320), [#300362](https://musescore.org/en/node/300362), [#304795](https://musescore.org/en/node/304795)
[#284887: crash on playback end in Debug build](https://musescore.org/en/node/284887)

Refactored unwinding to mimic "real" interpretation chain by a human over the previous copy-and-paste-rewind logic for jumps.

Also tracks section break pauses on RepeatSegments (but not used to fix related issue [#32696](https://musescore.org/en/node/32696) yet)

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/Documentation/blob/master/WorkflowAndGuidelines/CodeGuidelines.md)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

PR #11125, opened 2022-04-09
[MU3 backend] Fix #325825, fix #331098: Crash on MP3 export and playback with an open volta before a section break on a frame

Jojo-Schmitz:
which affects "Save online" too.

Resolves: https://musescore.org/en/node/325825 and https://musescore.org/en/node/331098, backport of #9574

### DIFF
--- a/libmscore/repeatlist.cpp
+++ b/libmscore/repeatlist.cpp
@@ -545,7 +545,7 @@ void RepeatList::collectRepeatListElements()
                         if (volta != nullptr) {
                               //if (volta->endMeasure()->tick() < mb->tick()) {
                                     // The previous volta was supposed to end before us (open volta case) -> insert the end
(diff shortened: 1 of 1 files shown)


jeetee:
Extremely low-risk fix which breaks no existing mtests for repeat playback.

Jojo-Schmitz:
BTW: It is a 3.5.1 regression, no crash with 3.5.0.

jeetee:
Indeed, the bug was introduced by #6248

RomanPudashkin:
3.6.2_backend is closed for any changes

jeetee:
Too bad, as this means the 3.6.2_backend is not on par with the actual 3.6.2

Jojo-Schmitz:
Indeed. IHMO a bad decision

PR #11140, opened 2022-04-11
[MU4] Fix #10406: Cannot drag elements from 'Master Palette' to 'Palettes'

### LINKED ISSUE #10406: Cannot drag elements from 'Master Palette' to 'Palettes'
**Describe the bug**

1. Cannot drag items from the **Master Palette** into **Palettes**
2. **Master Palette** window can get hidden behind main window (existing MS3 bug)

(See video)

https://user-images.githubusercontent.com/21022311/151660315-ee3ee2bc-3dc1-43cc-b95c-ba8151665f6e.mov

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10406

### DIFF
--- a/src/framework/uicomponents/view/topleveldialog.cpp
+++ b/src/framework/uicomponents/view/topleveldialog.cpp
@@ -1,0 +1,75 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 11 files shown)


src/framework/uicomponents/view/topleveldialog.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "topleveldialog.h"
+
+#include <QApplication>
+
+using namespace mu::uicomponents;
+
+TopLevelDialog::TopLevelDialog(QWidget* parent)
+    : QDialog(parent)
+{
+    setAttribute(Qt::WA_NativeWindow, true);
```
cbjeukendrup, line 32: I recall this window attribute was specific to the VST editors, because their content needs to be attached to a native window. I don't expect it will be necessary for all top level dialogs.

RomanPudashkin: Thank you! Fixed. But unfortunately, I had to remove the setTransientParent call from the constructor (now it will be called from TopLevelDialog::event). In the constructor, this call causes the crash of the application if the WA_NativeWindow attribute is not set.

PR #6489, opened 2020-08-24
[MU3] Prepare for updated libsndfile

Jojo-Schmitz:
This enables us to update libsndfile from the currently used 1.0.25 to 1.0.29 or later (1.0.30 is available meanwhile too, see https://github.com/erikd/libsndfile), which apparently changed the name of the lib from `libsndfile-1`to just `sndfile`.
In a separate step the dependencies file for Windows may get updated. If the newer one is found on build, that is taken, else the older one.
This in turn might enable us to add opus support, see https://musescore.org/en/node/288994

PR #11143, opened 2022-04-11
[MU4] Allow for newer libsndfile

Jojo-Schmitz:
Was part of #6483 (as a port of https://github.com/musescore/MuseScore/pull/6489), see c615ba3, but apparently broke again later.

For a newer libsndfile version to get used on GitHub CI too, the dependencies.7z needs to get updated, see https://github.com/musescore/MuseScore/pull/6489#issuecomment-679888471, which pointe at libsndfile1.0.30, but actually meanwhile we should be taking it from https://github.com/libsndfile/libsndfile/releases/tag/1.1.0, the currently latest.

### DIFF
--- a/src/main/CMakeLists.txt
+++ b/src/main/CMakeLists.txt
@@ -350,7 +350,12 @@ if (OS_IS_WIN)
             )
 
     if (SNDFILE_LIB)
-        install(FILES ${DEPENDENCIES_DIR}/libsndfile-1.dll DESTINATION bin)
+        if (CC_IS_MINGW)
+            install(FILES ${SNDFILE_LIB} DESTINATION bin)
+        else()
+            find_library(dll_sndfile NAMES "sndfile" "libsndfile-1" PATHS ${DEPENDENCIES_DIR} NO_DEFAULT_PATH)
+            install(FILES ${dll_sndfile} DESTINATION bin)
+        endif()
     endif()
 
     # Install ssl


Jojo-Schmitz:
Nope, that's not good, works only of the old libsndfile-1.dll is also available and then copdoes both.

src/main/CMakeLists.txt
```diff
@@ -350,7 +350,12 @@ if (OS_IS_WIN)
             )
 
     if (SNDFILE_LIB)
-        install(FILES ${DEPENDENCIES_DIR}/libsndfile-1.dll DESTINATION bin)
+        if (CC_IS_MINGW)
+            install(FILES ${SNDFILE_LIB} DESTINATION bin)
+        else()
+            find_library(dll_sndfile NAMES "sndfile" "libsndfile-1" PATHS ${DEPENDENCIES_DIR} NO_DEFAULT_PATH)
```
igorkorsukov, line 356: We have /build/cmake/FindSndFile.cmake, I believe, find should be there

Jojo-Schmitz: Even before my change we apparenerly don't (fully) rely on FindSndFile.cmake (which, as mentioned, for MSVC finds the .lib file, not the .dll), but rather hardcode it.

igorkorsukov: I got it, thanks

DmitryArefiev:
Hi @Jojo-Schmitz I have a problem with libsndfile file on Windows, app doesn't start and shows an error message. 
Previous master build doesn't have that file and runs fine.

https://user-images.githubusercontent.com/90187801/165716509-06aeb51f-ebc8-48c0-8091-d542ad65906d.mp4

Jojo-Schmitz:
Pretty strange as that dll is seen in the bin dir (to the roght)?

Jojo-Schmitz:
the lib vs dll issue is what my PR was trying to fix.

cbjeukendrup:
On the right is a different build.

Jojo-Schmitz:
Ah, I see.
Still puzzled as to why it doesn't work

Jojo-Schmitz:
It doesn't locally either. I'm 100% sure it did before though

Jojo-Schmitz:
I'm currently trying whether adding `set(CMAKE_FIND_LIBRARY_SUFFIXES ".dll")` fixes it

Jojo-Schmitz:
Seems it does. Added right before that `find_library(dll_sndfile NAMES "sndfile" "libsndfile-1" PATHS ${DEPENDENCIES_DIR} NO_DEFAULT_PATH)`

Taking this? Any better idea?

cbjeukendrup:
Yeah... or what if you try to put that in FindSndFile.cmake? I'd actually still prefer to have it all in one place.

Jojo-Schmitz:
OK, lemmy try something...

Jojo-Schmitz:
We'd need separate detections for LIB and DLL for this, esp. for MSVC

Jojo-Schmitz:
Check #11444. Seems the cleanest solution to me. I should have done it that way from the start

Issue #11060, opened 2022-04-02
[MU4 Issue] Actions without shortcuts are missing in Preferences


rgreen5:
**Describe the bug**

In MU4 Preferences > Shortcuts, the actions list is missing all the actions that did not have defined shortcuts in MU3.

For example, in MU3 there are actions for adding intervals above or below the current note. Only the "above" actions had shortcuts assigned.

![image](https://user-images.githubusercontent.com/11011881/172459052-802665c9-40e1-44f3-91e5-ec2c55122bb9.png)

The actions list in MU4 only has the "above" intervals, so it is not possible to assign shortcuts to "below" intervals at all.

![image](https://user-images.githubusercontent.com/11011881/172482802-2af3fc97-3720-44c8-932e-b87854b80d1b.png)

**Expected behavior**

All actions in MS3 should be available in MS4 as well.

https://user-images.githubusercontent.com/101250347/161807657-966ad3d1-64c6-4458-9d58-a0ea49dda132.mp4

**Platform information**
OS: Linux Mint 20.1 (Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2059995262, revision: github-musescore-musescore-cafefd7), MacOS, WIN 11

boryskuzmenko:
Thank you for logging the issue. 
Screen record of the issue added

MarcSabatella:
Right now from what I can see, the most glaringly obvious missing shortcuts are the navigation commands that are absolutely essential for accessibility as well as for keyboard use in general. In particular, the shortcuts for next/previous staff or voice. These commands seem to just be gone, unless they've been renamed to something I can't find, which is certainly possible.

Without these commands, it's impossible to navigate or edit anything but the top staff of a score without the mouse. So if they are truly missing, adding them back should be P0, as it's a complete showstopper for accessibility, and also quite bad for anyone trying to enter notes via keyboard.

Next most important missing shortcuts to me are F2 (special characters dialog) and Z (symbols palette). I would also put in votes for Shift+K/T for key/time signatures palettes. These are ones I find myself most often being disappointed by.

cbjeukendrup:
@igorkorsukov In PR #11144, you commented that the solution to this issue would be to put all actions that _can_ have shortcuts in shortcuts.xml, even if they have no shortcut by default. 

Keeping task #11804 in mind (it's closed but still needs to be done), I had another idea. 

I propose that we give each UiAction in cpp a new field called `category`. We could have action categories like this:
```cpp
[...]

PR #11144, opened 2022-04-11
fix: #11060; Extract leftover action codes into the shortcuts list

VanSHOE:
Resolves: #11060 

It reads from a map object m_actions of the `uiactionsregister`, whose keys contain all the action codes as far as I can understand. Do let me know in case this is not the case.  

Secondly, there are a few null shortcut names

image

Please let me know how to proceed with the same

### DIFF
--- a/src/framework/ui/iuiactionsregister.h
+++ b/src/framework/ui/iuiactionsregister.h
@@ -41,6 +41,7 @@ public:
     virtual const UiAction& action(const actions::ActionCode& code) const = 0;
     virtual UiActionState actionState(const actions::ActionCode& code) const = 0;
     virtual async::Channel<actions::ActionCodeList> actionStateChanged() const = 0;
+    virtual std::vector<std::string> getKeys() const = 0;
 };
 }
 
--- a/src/framework/shortcuts/view/shortcutsmodel.cpp
+++ b/src/framework/shortcuts/view/shortcutsmodel.cpp
@@ -25,6 +25,7 @@
 #include "ui/view/iconcodes.h"
 #include "translation.h"
 #include "log.h"
+#include <unordered_set>
 
 using namespace mu::shortcuts;
 using namespace mu::ui;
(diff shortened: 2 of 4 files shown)


src/framework/ui/iuiactionsregister.h
```diff
@@ -27,7 +27,7 @@
 #include "iuiactionsmodule.h"
 #include "uitypes.h"
 #include "async/channel.h"
-
+#include <unordered_map>
```
VanSHOE, line 30: Ah sorry... I left this here by mistake, will remove it next time. Anyways have to squash commits later on

igorkorsukov:
This solution opens up all the actions for shortcuts, but there are a few problems: 
1. Not all actions are "public", there are some internal ones that can be added and removed at any time. 
2. There are too many shortcuts, it is not worth increasing their number even more 
3. The names of the actions may change (for example, just in case of an incorrect name, a new understanding, etc.), but it is not necessary that the shortcuts break.

Therefore, adding shortcuts is arranged differently: 
There is a file `shortcuts.xml` is a list of actions and shortcuts that we want to provide to the user, it's like a public API (unlike uiactionsregister, which is a private service). 
If some kind of shortcut is missing, then it needs to be added to `shortcuts.xml` . Exactly what is written in `shortcuts.xml`  we can keep it stable, everything else can change and break for the user.

VanSHOE:
Oh sure, I'll revert this change and add those shortcuts to the XML
Which ones should I add?

igorkorsukov:
> Which ones should I add?

This is the main difficulty - we need to find out which ones are missing

VanSHOE:
> > Which ones should I add?
> 
> This is the main difficulty - we need to find out which ones are missing

I apologize, I have back to back examinations these 10 days. If not done till first week of May, I will definitely go through the actions and see which ones are required and add them as needed

PR #11156, opened 2022-04-11
[MU4] Fix #11018: Number inputs can't be scrolled

### LINKED ISSUE #11018: [MU4 Issue] [UX] Number inputs can't be scrolled
**Describe the bug**
In MU3, when a numerical input is focused and the pointer hovering over it, scrolling changes the value.
In MU4 this does not occur.

**To Reproduce**
1. Open a score
2. Select a note
3. Open the "Playback" section of the Properties
4. Focus the "Velocity" input
5. Try and scroll, and observe that the value is not changed.

**Expected behavior**
Scrolling over the input while focused will change the value in the direction scrolled.

**Platform information**
 - OS: Linux

poypoyan:
Resolves: #11018

Docs:
https://doc.qt.io/qt-5/qml-qtquick-mousearea.html
https://doc.qt.io/qt-6/qml-qtquick-mousearea.html

- [x] I signed [CLA](https://musescore.org/en/cla) ([Ameer Lubang](https://musescore.org/en/user/3060033))

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/IncrementalPropertyControl.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/IncrementalPropertyControl.qml
@@ -112,6 +112,8 @@ Item {
     TextInputField {
         id: textInputField
 
+        property int scrolled: 0
+
         anchors.top: parent.top
         anchors.bottom: parent.bottom
 
@@ -147,6 +149,39 @@ Item {
 
             onIncreaseButtonClicked: { root.increment() }
             onDecreaseButtonClicked: { root.decrement() }
+        }
+
+        mouseArea.onWheel: function(wheel) {
+            if (textInputField.activeFocus == true) {
+                let pixelY = wheel.pixelDelta.y
+                let angleY = wheel.angleDelta.y
+
+                // This is set below. For angleY, make sure it is <= 120,
+                // because in many actual mouse wheels, one scroll sets
+                // angleY to +/- 120.
+                let oneScroll = 0
+
+                if (pixelY != 0) {
+                    scrolled += pixelY
+                    oneScroll = 60
+                } else if (angleY != 0) {
+                    scrolled += angleY
+                    oneScroll = 120
+                }
+                if (scrolled >= oneScroll) {
+                    root.increment()
+                    scrolled = 0
+                } else if (scrolled <= -oneScroll) {
+                    root.decrement()
+                    scrolled = 0
+                }
+            } else {
+                wheel.accepted = false
+            }
+        }
+
+        mouseArea.onExited: {
+            scrolled = 0
         }
 
         onCurrentTextEdited: function(newTextValue) {


MarcSabatella:
As long as you have to click or otherwise give focus to the field, I'm not so sure it matters if the mouse remains over it or not when initiating scrolling.  Although I recognize it isn't common.  But, if the scroll wheel starts affecting fields that don't even have focus, that would definitely be a problem.

More importantly to me, see also https://github.com/musescore/MuseScore/issues/10460, which is about keyboard control.  After clicking, it is very important to be able to work the spinbox via keyboard.   Same for other controls, but it's especially crucial for spinboxes.  That has been identified as a P1 issue and I definitely agree, it's actually the most awkward thing in the entire UI to me.

Anyhow, it will be important to ensure that fixes for these issues are compatible with each other.

poypoyan:
@MarcSabatella Oh I notice this too.
@Marr11317 Yes, number field scrolls just when mouse hovered.

I will now work on both.

poypoyan:
Update: number field scrolls only when in activeFocus (text input).

There is still one difference from this and MU3: here, when number field is in activeFocus, when you scroll the inspector and the number field go through, the number field scrolls, **but the cursor/inspector scroll doesn't stop**. In MU3, the cursor/inspector scroll stops.

llui85:
Thank you so much for working on this! I **really** appreciate it.

Just tested your latest push (on Linux), the scrolling is far too sensitive. Taking a look at your code, what's happening is that MU3 bases the change in value of the distance scrolled, whereas yours assumes a fixed velocity, which isn't accurate with a touchpad .

Here's a screen recording of me doing a slowish two finger scroll up and down the height of my touchpad in MU3. Note how the change in value is pretty much the same either direction, because the physical movement was the same:

https://user-images.githubusercontent.com/47158068/163360187-eda5cbd0-671a-4ecd-a2c3-ed9b31db6e76.mov

___

It looks like you can use the `pixelDelta` property of the event, then falling back to `angleDelta` if it's missing. See the docs and code sample here: https://doc.qt.io/qt-5/qwheelevent.html#angleDelta

boryskuzmenko:
> Update: number field scrolls only when in activeFocus (text input).
> 
> There is still one difference from this and MU3: here, when number field is in activeFocus, when you scroll the inspector and the number field go through, the number field scrolls, **but the cursor/inspector scroll doesn't stop**. In MU3, the cursor/inspector scroll stops.

It can be scrolled on macOS as well. Thanks. But value does not correspond scroll direction on touchpad. When moving fingers up - value goes to negative values.
Also, speed of value changes is too high (e.g. impossible to raise on one point)

poypoyan:
Hopefully this is less sensitive. The "When moving fingers up - value goes to negative values" in trackpad is expected behavior afaik.

Marr11317:
Why use a new mousearea when the `TextInputFied` Already has one? Couln't it be reused?

poypoyan:
Just learned to use TextInputField's MouseArea instead. I'll just fix the code style.

boryskuzmenko:
@poypoyan almost done) should be a bit more sensitive (closer to your first variant). The best way to use MU3 as a reference.
Also, could you please change the direction to opposite. It is another bug but it will be great if you do it in scope of this one.

llui85:
Perfect! On Linux, direction and sensitivity are exactly right.

It might be best to shelve the direction issue for later, until we can confirm the expected direction on all platforms. Does the `inverted` property of the event help in any way?

poypoyan:
Hopefully, `inverted` fixes the direction. In the documentation, it is said that there are many platforms that always return `false` regardless of scroll direction. That would require more complicated fixes.

For the macOS sensitivity, let me know. I just guesstimated `oneScroll = 60`. Setting lower value makes it more sensitive.

boryskuzmenko:
@poypoyan Everything ok from your side. But I was wrong about direction on macOS. Please, return it back, my bad.. The same situation on all another platforms when using touchpad instead of mouse. 
Thank you

boryskuzmenko:
**https://github.com/musescore/MuseScore/pull/11156#:~:text=Resolves%3A-,%2311018,-Docs%3A%0Ahttps%3A//doc is FIXED**
Tested on macOS12, Win11, Ubuntu20

Please, merge to master

Eism (approved):

PR #11158, opened 2022-04-11
fix #11157: Checking that string is not empty before erasing

### LINKED ISSUE #11157: [MU4 Issue] Crash on opening preferences (master branch)
**Describe the bug**
As soon as I open preferences on the latest master branch, the program crashes

**To Reproduce**
1. First update to the latest master branch
2. Open MuseScore
3. Click on Edit > Preferences
4. MuseScore crashes

https://user-images.githubusercontent.com/75690289/162833197-c6d08f82-2d42-40df-b3a3-28f53fca0c72.mp4

**Expected behavior**
It should safely open the preferences dialogue box

**Screenshots**
Video already attached

**Platform information**
 - OS: Windows 10

**Additional context**
Running the application in debug mode in QT creator does hit a breakpoint but continuing the application allows it to open the preferences option.

https://user-images.githubusercontent.com/75690289/162833793-60bd059b-3cd0-48b5-b4ca-b977b84b3935.mp4

VanSHOE:
Resolves: #11157 

A simple check to ensure that `pop_back` is not called on an empty string.

### DIFF
--- a/src/framework/global/io/path.cpp
+++ b/src/framework/global/io/path.cpp
@@ -246,6 +246,9 @@ std::string mu::io::pathsToString(const paths& ps, const std::string& delim)
     }
 
     for (size_t i = 0; i < delim.length(); ++i) {
+        if (result.empty()) {
+            break;
+        }
         result.pop_back();
     }
 


src/framework/global/io/path.cpp
```diff
@@ -246,7 +246,9 @@ std::string mu::io::pathsToString(const paths& ps, const std::string& delim)
     }
 
     for (size_t i = 0; i < delim.length(); ++i) {
-        result.pop_back();
+        if (result.size()) {
```
RomanPudashkin, line 249: It is better to rewrite this code as follows:

```
for (size_t i = 0; i < delim.length(); ++i) {
    if (result.empty()) {
        break;
    }

    result.pop_back();
}
```

VanSHOE: Sure, I had considered the same but realized that after the commit. 
I'll push it right away!

RomanPudashkin: Or even like this:

result.erase(result.length() - delim.length())

RomanPudashkin (approved):

PR #11163, opened 2022-04-12
[MU4] Fix #10461: Elements often cannot be dragged (if they were previously selected)

### LINKED ISSUE #10461: [MU4 Issue] Elements often cannot be dragged (if they were previously selected)
As far as I can tell, you cannot change the position of elements by dragging, except sometimes when you can/  Haven't found the pattern yet

1) new score
2) Ctrl+T to add staff text, anything at all
3) now try to drag the text elsewhere
Result: either it drags as expect, or else you are placed into edit mode and dragging isn't possible

I'm sure there is some rule being followed here as to when the click to initiate drag activates edit mode versus actually allowing the drag - maybe it has something to do with how long you wait before moving the mouse, or else it's about where in the text you click, not sure.  But for quite a while I was convinced dragging just plain didn't work at all.  I'm only successful maybe 10-20% of the time.

https://user-images.githubusercontent.com/90187801/162727978-4a84c28c-30e4-4a58-8453-9e276b28a859.mp4

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10461

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -1182,7 +1182,7 @@ void NotationActionController::openSelectionMoreOptions()
     }
 }
 
-void NotationActionController::startEditSelectedElement()
+void NotationActionController::startEditSelectedElement(const ActionData& args)
 {
     auto interaction = currentNotationInteraction();
     if (!interaction) {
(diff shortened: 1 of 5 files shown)


DmitryArefiev:
Tested #10461 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #11169, opened 2022-04-12
[engraving] replaced QList to std. Part 1

igorkorsukov:


handrok (approved):

Jojo-Schmitz:
Why would this PR change vtests?

igorkorsukov:
> Why would this PR change vtests?

I am investigating this problem, different by 4-16 pixels... (

igorkorsukov:
> Why would this PR change vtests?

there was a different sorting, now the same

Jojo-Schmitz:
Good to have those tests ;-)

PR #11175, opened 2022-04-13
[MU3] Backend log fix

Eism:
Disabled logging for backend

### DIFF
--- a/mscore/musescore.cpp
+++ b/mscore/musescore.cpp
@@ -4097,34 +4097,10 @@ static bool processNonGui(const QStringList& argv)
 //   Message handler
 //---------------------------------------------------------
 
-#if defined(QT_DEBUG) && defined(Q_OS_WIN)
-static void mscoreMessageHandler(QtMsgType type, const QMessageLogContext& context, const QString& msg)
+static void mscoreMessageHandler(QtMsgType, const QMessageLogContext&, const QString&)
 {
-    QTextStream err(stderr);
-    QByteArray localMsg = msg.toLocal8Bit();
-
-    switch (type) {
-    case QtInfoMsg:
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Wouldn't using a non-Debgg build have disabled that logging automgically?

Jojo-Schmitz:
Also how is disabling some code that is for Windows only going to help on the backend, which is Linux? I'm confused...

Eism:
@Jojo-Schmitz 
> Wouldn't using a non-Debug build have disabled that logging automagically?

qWarning in qt library are printed to stdout. Extra logs interfere with the parsing of command results. We can disable logging only in this way.

> Also how is disabling some code that is for Windows only going to help on the backend, which is Linux? I'm confused...

That is the problem.
Previously, the handler worked only for Windows, for Linux (backend) the logs were printed as they are to the console. Log output is now disabled for all platforms on the backend

Jojo-Schmitz:
I see, thanks for explaining. 
So we might have done this instead:
```
static void mscoreMessageHandler(QtMsgType type, const QMessageLogContext& context, const QString& msg)
{
#if ! defined(QT_DEBUG)
      Q_UNUSED(type);
      Q UNUSED(context);
      Q_UNUSED(msg)
#else
...
#endif
}
```

Jojo-Schmitz:
Of even
```
static void mscoreMessageHandler(QtMsgType type, const QMessageLogContext& context, const QString& msg)
{
      if (debugMode) {
...
      }
}
```
to have them available when using the -d option

Jojo-Schmitz:
BTW: `qWarning()` does not send to `stdout` but to `stderr`, so a ` 2>/dev/null` should get rid of it too

PR #11184, opened 2022-04-13
[MU4] lame_thirdparties

vpereverzev:


RomanPudashkin (approved):

Jojo-Schmitz:
This does add a whole bunch of compiler warnings in at least MinGW, about Shifting a negative signed value being undefined, fallthough in switch/case, missing braces in empty else statement, 'packed' attribube being ognored and some others

PR #11187, opened 2022-04-13
[MU4] Fix #10795: Instrument playback should be dynamic, based on the users selection on the score

### LINKED ISSUE #10795: Instrument playback should be dynamic, based on the users selection on the score.
**Task description**
In most other notation apps (excluding Finale), users can quickly solo an instrument or hear a range of specific instruments by making a selection on the score. This is an incredibly useful feature that removes the need to fiddle with Mute/Solo buttons in the mixer and is long overdue in MuseScore.

It should work like this:

If a user selects an **element** on the score and then presses 'play', playback will behave as normal, including all instruments.

image

If the user makes the following type of selection (using 'Shift' or by selecting the measure), then playback will only be heard for the selected stave. In the example below, the user has selected the viola and cello, so only those instruments should be audible when the user presses 'play'.

image

Bear in mind that multiple selections like this would NOT affect playback (see below). Only the 'blue box' selection should do that.
image

This behaviour should not have any effect on the mixer UI or settings. In other words, the users selection should not alter the states of the **solo/mute** buttons. Also, if a user mutes an instrument in the mixer, it won't be heard regardless of the user's selection on the score.

Undoubtedly, this behaviour will be new for some users. However, when they become accustomed to it, the benefit will very evident.

RomanPudashkin:
Resolves: #10795

### DIFF
--- a/src/framework/audio/internal/worker/clock.cpp
+++ b/src/framework/audio/internal/worker/clock.cpp
@@ -49,12 +49,22 @@ void Clock::forward(const msecs_t nextMsecs)
         return;
     }
 
-    if (newTime > m_timeDuration) {
+    if (newTime >= m_timeDuration) {
+        setCurrentTime(m_timeDuration);
         pause();
         return;
     }
 
-    m_currentTime = newTime;
+    setCurrentTime(newTime);
+}
+
+void Clock::setCurrentTime(msecs_t time)
+{
+    if (m_currentTime == time) {
+        return;
+    }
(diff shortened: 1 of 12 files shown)


vpereverzev (approved):

DmitryArefiev:
Tested #10795 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

vpereverzev (approved):

PR #11197, opened 2022-04-14
[MU4] Fix (MSVC) compiler warnings

Jojo-Schmitz:
reg. size_t vs. int, after #11194

### DIFF
--- a/src/engraving/libmscore/harmony.cpp
+++ b/src/engraving/libmscore/harmony.cpp
@@ -2008,7 +2008,7 @@ void Harmony::addDegree(const HDegree& d)
 //   numberOfDegrees
 //---------------------------------------------------------
 
-int Harmony::numberOfDegrees() const
+size_t Harmony::numberOfDegrees() const
 {
     return _degreeList.size();
 }
--- a/src/engraving/libmscore/part.cpp
+++ b/src/engraving/libmscore/part.cpp
@@ -229,7 +229,7 @@ void Part::write(XmlWriter& xml) const
 
 int Part::nstaves() const
 {
-    return _staves.size();
+    return static_cast<int>(_staves.size());
 }
 
 const std::vector<Staff*>* Part::staves() const
@@ -267,7 +267,7 @@ void Part::setShortNames(std::list<StaffName>& name, const Fraction& tick)
 
 void Part::setStaves(int n)
 {
-    int ns = _staves.size();
+    int ns = static_cast<int>(_staves.size());
     if (n < ns) {
         qDebug("Part::setStaves(): remove staves not implemented!");
         return;
--- a/src/engraving/libmscore/scoretree.cpp
+++ b/src/engraving/libmscore/scoretree.cpp
@@ -874,7 +874,7 @@ EngravingObject* BSymbol::scanChild(int idx) const
 
 int BSymbol::scanChildCount() const
 {
-    return _leafs.size();
+    return static_cast<int>(_leafs.size());
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/scoretree.cpp
+++ b/src/engraving/libmscore/scoretree.cpp
@@ -78,7 +78,7 @@ EngravingObject* Score::scanChild(int idx) const
     return pages()[idx];
 }
 
-int Score::scanChildCount() const
+size_t Score::scanChildCount() const
 {
     return pages().size();
 }
@@ -98,9 +98,9 @@ EngravingObject* Page::scanChild(int idx) const
     return systems()[idx];
 }
 
-int Page::scanChildCount() const
+size_t Page::scanChildCount() const
 {
-    return static_cast<int>(systems().size());
+    return systems().size();
 }
 
 //---------------------------------------------------------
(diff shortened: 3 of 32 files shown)


src/framework/global/containers.h
```diff
@@ -61,6 +63,7 @@ inline T value(const std::vector<T>& vec, size_t idx)
         return nullptr;
     }
 
+    UNREACHABLE;
     return T();
```
igorkorsukov, line 67: it can be reachable, I think better add 'else`

src/engraving/libmscore/scoretree.cpp
```diff
@@ -874,7 +874,7 @@ EngravingObject* BSymbol::scanChild(int idx) const
 
 int BSymbol::scanChildCount() const
 {
-    return _leafs.size();
+    return static_cast<int>(_leafs.size());
```
igorkorsukov, line 877: I believe, better change method signature to `size_t BSymbol::scan ChildCount() const`

src/engraving/libmscore/part.cpp
```diff
@@ -229,7 +229,7 @@ void Part::write(XmlWriter& xml) const
 
 int Part::nstaves() const
 {
-    return _staves.size();
+    return static_cast<int>(_staves.size());
```
igorkorsukov, line 232: I believe, better change method signature to `size_t Part::nstaves() const`

src/engraving/libmscore/harmony.cpp
```diff
@@ -2010,7 +2010,7 @@ void Harmony::addDegree(const HDegree& d)
 
 int Harmony::numberOfDegrees() const
 {
-    return _degreeList.size();
+    return static_cast<int>(_degreeList.size());
```
igorkorsukov, line 2013: I believe, better change method signature to `size_t Harmony::numberOfDegrees() const`

src/framework/global/containers.h
Jojo-Schmitz: Hmm, MSVC apparently thinks it is not? And an else branch would equally (be reported as) unreachable?

No, apparently not, so I'll go for the `else`

src/engraving/libmscore/scoretree.cpp
Jojo-Schmitz: I general I tend to agree, but that requires more changes in lots other places too

Jojo-Schmitz: See b52db31

src/engraving/libmscore/part.cpp
Jojo-Schmitz: Same comment as above, many more changes needed in lots other places

src/engraving/libmscore/harmony.cpp
Jojo-Schmitz: This one is an easy change, will do.

src/engraving/libmscore/scoretree.cpp
Jojo-Schmitz: You wanna go that way?

src/engraving/libmscore/part.cpp
Jojo-Schmitz: This one seems even worse, requiring more casting elsewhere (or even much bigger changes)

Jojo-Schmitz: Not worth the effoert IMHO

Jojo-Schmitz: not leading to cleaner code at a reasonable price

src/engraving/libmscore/scoretree.cpp
igorkorsukov: I believe it's worth it

Jojo-Schmitz: This one indeed might. Also it is done ;-)
Not so the other one, below.

src/engraving/libmscore/part.cpp
igorkorsukov: I think it's better to do this later when most of the code will be changed to std

Jojo-Schmitz:
Hmm, that merge was a bit early, I was just about to check whether any of the changes cause issues/warnings with MinGW, and sure enough this is the case..
Oh well, next PR coming...

Jojo-Schmitz:
Better revert that last commit, it just causes tons of warnings with MinGW (and presumable other gcc based compilers).
Fixing those would be another huge change, esp. when trying to avoid introducing other MSVC warnings (and on fixing those causeing other MinGW warnings)

igorkorsukov:
> Hmm, that merge was a bit early, I was just about to check whether any of the changes cause issues/warnings with MinGW, and sure enough this is the case..
> Oh well, next PR coming...

Oh, sorry, I merged quickly to reduce conflicts (

Jojo-Schmitz:
My fault (too) I guess, I should have marked it DRAFT...
I also would have squashed the 3rd commit into the 1st. But now I'm glad I did not, makes reverting that 3rd commit far easier.
I believe we're better of witout it, for now, and possibly revisit all these size_t issues later,  _when most of the code will be changed to std_

igorkorsukov:
I looked, commit revert not need, I try fix most warning in a next PR

Jojo-Schmitz:
There are also still tons of warnings with MSVC, not sure why (or rather why I hadn't seen them earlier).
And even if I do revert that commit! I'm severely confused now.

PR #11213, opened 2022-04-15
[engraving] replaced QList to std. Part 4

igorkorsukov:
I tried to reduce the number of warnings, but there are still a lot of them, the  work continues...

### DIFF
--- a/src/engraving/libmscore/barline.cpp
+++ b/src/engraving/libmscore/barline.cpp
@@ -178,7 +178,7 @@ static void undoChangeBarLineType(BarLine* bl, BarLineType barType, bool allStav
     break;
     case BarLineType::START_REPEAT: {
         Measure* m2 = m->isMMRest() ? m->mmRestFirst() : m;
-        for (int staffIdx = 0; staffIdx < m2->score()->nstaves(); ++staffIdx) {
+        for (size_t staffIdx = 0; staffIdx < m2->score()->nstaves(); ++staffIdx) {
             if (m2->isMeasureRepeatGroupWithPrevM(staffIdx)) {
                 MScore::setError(MsError::CANNOT_SPLIT_MEASURE_REPEAT);
                 return;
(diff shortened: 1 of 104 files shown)


handrok (approved):

src/engraving/libmscore/barline.cpp
```diff
@@ -570,7 +570,7 @@ bool BarLine::isBottom() const
     if (!_spanStaff) {
         return true;
     }
-    int idx = staffIdx();
+    size_t idx = staffIdx();
     if (idx == score()->nstaves() - 1) {
```
Jojo-Schmitz, line 574: Here `idx` could potentially go negative, if `nstaves()` is 0, or is that something not to consider?

src/engraving/libmscore/bracketItem.cpp
```diff
@@ -51,7 +51,7 @@ mu::engraving::PropertyValue BracketItem::getProperty(Pid id) const
     case Pid::BRACKET_COLUMN:
         return _column;
     case Pid::BRACKET_SPAN:
-        return _bracketSpan;
+        return int(_bracketSpan);
```
Jojo-Schmitz, line 54: Hmm, not `static_cast<int>(_bracketSpan)`?

src/engraving/libmscore/bracketItem.cpp
```diff
@@ -67,7 +67,7 @@ bool BracketItem::setProperty(Pid id, const engraving::PropertyValue& v)
         staff()->changeBracketColumn(column(), v.toInt());
         break;
     case Pid::BRACKET_SPAN:
-        _bracketSpan = v.toInt();
+        _bracketSpan = size_t(v.toInt());
```
Jojo-Schmitz, line 70: similar

src/engraving/libmscore/barline.cpp
igorkorsukov: I looked that there is an equality check, so no problem if there was -1

src/engraving/libmscore/bracketItem.cpp
igorkorsukov: fixed

src/engraving/libmscore/bracketItem.cpp
igorkorsukov: fixed

PR #11215, opened 2022-04-15
remote binaries from repo

lyrra:
- [v ] I signed [CLA](https://musescore.org/en/cla)


src/diagnostics/CMakeLists.txt
```diff
@@ -102,12 +102,15 @@ endif(BUILD_CRASHPAD_CLIENT)
...
+#    set(CPAD_ROOT_PATH ${CMAKE_CURRENT_LIST_DIR}/crashpad_handler)
+#    if (OS_IS_LIN)
+#        install(PROGRAMS ${CPAD_ROOT_PATH}/linux/x86-64/crashpad_handler DESTINATION bin)
+#    elseif(OS_IS_WIN)
+#        install(PROGRAMS ${CPAD_ROOT_PATH}/windows/x86-64/crashpad_handler.exe DESTINATION bin)
+#    elseif(OS_IS_MAC)
+#        install(PROGRAMS ${CPAD_ROOT_PATH}/macos/x86-64/crashpad_handler DESTINATION ${Mscore_SHARE_NAME}Contents/MacOS)
+#    endif()
+#endif(BUILD_CRASHPAD_CLIENT)
+
```
Jojo-Schmitz, line 116: Hmm. I guess they should get installed, but build first?

lyrra: Yes, not sure if it builds but install doesn't find the built binaries. Or if they aren't built at all.

Jojo-Schmitz: But that's a problem than needs to get sorted first

lyrra: Yes, meanwhile I'll convert this PR to a draft.

igorkorsukov: crashpad_handler prebuild binaries, without which crash dumps will not work

lyrra: Ok, so you can't build the crashpad-handler from sources (easily), and the crashpad handler will do network activity (https://chromium.googlesource.com/crashpad/crashpad/+/HEAD/handler/crashpad_handler.md).

I'm updating the PR to remove the prebuilt binaries and turn the options off by default.

src/diagnostics/CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ set(MODULE_SRC
 # --- Crashpad ---
 # Not building with MinGW, so turned off for MinGW
 include(GetCompilerInfo)
-set(BUILD_CRASHPAD_CLIENT ON)
+set(BUILD_CRASHPAD_CLIENT OFF)
```
igorkorsukov, line 74: Why did you decide to turn off crash dumps?

lyrra: It because I can't build the crashpad-handler from source. It looks like the source is there, but is it missing some CMakeList file or what is missing?

igorkorsukov: In this case, you should disable the use of crashpad only for yourself, and not as a whole in the project. (you should probably add the cmake  option)

lyrra: yes I'll rework it to have an global option to disable crashpad.

tibequadorian:
Currently building MuseScore 4.0 for Void Linux and this is an issue. The binaries should not be installed when built with `BUILD_CRASHPAD_CLIENT=OFF`.

Issue #11195, opened 2022-04-14
Accessibility - Export dialog does not take the keyboard focus.


Tantacrul:
**Describe the bug**
See image: when going to file > export, the keyboard focus remains on the main application. This is particularly bad because it is the only place where a user can export Braille (apart from any score output). Hence, it's a P0.

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/21022311/163377728-744a7471-37af-47e3-9aca-4376c9e7e94c.png">

Tested on Mac OS only.

VanSHOE:
If this helps anyone:
https://user-images.githubusercontent.com/75690289/163384198-4f19d201-fe92-457a-9372-4637ae889824.mp4
Issue not there on Windows as far as I can understand

Tantacrul:
Ah, I've discovered some nuance. Updated here with explanation:

https://user-images.githubusercontent.com/21022311/163386360-b43dba9c-89cd-4e12-a878-3eb0855f7942.mov

shoogle:
I've noticed this occurs occasionally with the New Score Wizard too. It usually happens after switching focus to another app (or changing virtual desktop in fullscreen mode) and then switching focus back to MuseScore again: focus will go to MuseScore's main window rather than the dialog.

Pressing the Tab key a few times is usually enough to get focus back to the dialog, but this shouldn't be necessary. If a modal dialog like Export or the NSW is visible then it shouldn't be possible to focus the main window at all.

DmitryArefiev:
Fixed in #11233

Issue #10919, opened 2022-03-26
[MU4 Issue] Tabbing around while inputting additional score information is out of order.


MajesticString:
**Describe the bug**
Upon creating a new score, when inputting "Additional Score Information", tabbing around doesn't select the right inputs

**To Reproduce**
1. Create a new score (ctrl + N or other)
2. Get to the Additional Score Information
3. Click the title input
4. Press tab

**Expected behavior**
The selected input would be Composer, Instead, it selects the "key signature"

**Screenshots**
![2022-03-26_12-08-51_AdobeCreativeCloudExpress (2)](https://user-images.githubusercontent.com/66224939/160253801-3fbe7a22-e58c-47a6-96ba-b69287165f9e.gif)
(sorry its so low quality)

**Platform information**
 - OS: Windows 11

RobFog:
@vpereverzev This should probably be targeted for the public alpha. There have been plenty of reports.

DmitryArefiev:
Fixed in #11233

Issue #9839, opened 2021-11-23
[MU4 Issue] Navigation with arrows doesn't work together with mouse clicks (in new style windows)


DmitryArefiev:
**To Reproduce**
**Case 1**
1. Open **Instruments** window (or **Preferences**)
2. Click on any instrument
3. Start to navigate with ↑↓ arrows - it **doesn't work** (you need to start navigation with tab and then arrows start to work)

https://user-images.githubusercontent.com/90187801/143014038-832871b6-54b1-488a-92b9-ad9a0cee3788.mp4

**Case 2**
1. Open **Instruments** window
2. Click on search field
3. Type something (e.g. flute)
4. Start to navigate through the list with ↑↓ arrows - it **doesn't work** (you need to start navigation with tab and then arrows start to work)

https://user-images.githubusercontent.com/90187801/158849269-4ad8b9d3-a6b9-4513-bcc3-d78d08c3fd50.mp4

**Desktop:**
Win10/Mac12/LinuxMint20

**Additional context**
Doesn't occur in **MU3**
In old style windows (Master Palette, Style) it works fine

Eism:
also see #10919

DmitryArefiev:
Fixed in #11233

DmitryArefiev:
> Expected behavior
> The selected input would be Composer, Instead, it selects the "key signature"

@MajesticString Regarding the expected result, MU4 uses **arrows** to navigate within a group and **Tab** to switch between item groups

PR #11233, opened 2022-04-18
[MU4] Navigation improvements

Eism:
Resolves: #9839
Resolves: #10919
Resolves partially: # 10460
Resolves: #11195

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -419,6 +419,12 @@ DockPage {
 
         Component.onCompleted: {
             root.notationView = notationView
+
+            root.setDefaultNavigationControl(defaultNavigationControl)
+        }
+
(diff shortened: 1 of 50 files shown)


MarcSabatella:
Thanks, it works, albeit with a bit of an odd gltich in that the *first* cursor action does not produce any real result.  If I select a note then click in a field like "Offset" under "Appearance", the first press of cursor up or down just changes the displayed "0" to "0.00".  Then after that it starts incrementing/decrementing normally.  With integer values like "Body text size" in the Preferences dialog, it's a little more apparent that what is actually happening is that the first press changes focus from the *contents* of the field to the field itself.  Would be nice to clean that up.  But it's definitely an improvement.

FWIW, it seems to even for dropdowns, at least the font selection in the Properties dialog.  But not most others in Preferences.

igorkorsukov (approved):

PR #11245, opened 2022-04-19
Fix (MSVC) compiler warning

Jojo-Schmitz:
reg. "discarding return value of function with 'nodiscard' attribute"

Fixes just one odd warning, still leaving some 1005 warnings C4267 "'return': conversion from 'size_t' to 'int', possible loss of data" (along with 4 long term warnings C4505 "'checkNotifySignalValidity_KDDockWidgets__DockWidgetQuick': unreferenced function with internal linkage has been removed" all of which are due to a 'moc' limitation, so can't get avoided as far as I can tell)

### DIFF
--- a/src/engraving/libmscore/repeatlist.cpp
+++ b/src/engraving/libmscore/repeatlist.cpp
@@ -428,7 +428,7 @@ void RepeatList::collectRepeatListElements()
                 }
                 // Cross-section of the repeatList
                 std::vector<int> endings = remainder->endings();
-                std::remove_if(endings.begin(), endings.end(), [&volta](const int& ending) {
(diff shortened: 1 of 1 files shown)


igorkorsukov:
I introduced a bug here :(, I will fix it in a next PR

Jojo-Schmitz:
So for once a compiler warning revealed a bug ;-)

igorkorsukov:
fixed here https://github.com/musescore/MuseScore/pull/11246/files#diff-1e5ed86898f898777f75da53b53122137f41791d0de716469e459473077a88c8R431

Jojo-Schmitz:
So close this PR?

igorkorsukov:
> So close this PR?

I think, yes

PR #11247, opened 2022-04-19
[MU4] Fix some (MinGW) compiler warnings

Jojo-Schmitz:
* reg. "comparison of unsigned expression < 0 is always false" _(all, actaully the one and only)_
* reg. "'%lu' expects argument of type 'long unsigned int', but argument has type 'std::vector<Ms::Staff*>::size_type' {aka 'long long unsigned int'}" _(or similar, all)_
* reg. "comparison of integer expressions of different signedness: 'int' and 'size_t' {aka 'long long unsigned int'}" _(just a handfull of some 100, but better than none ;-))_


igorkorsukov (approved):

Jojo-Schmitz:
Still 52 warnings in MinWG and 1005 (!) in MSVC, all about `int` vs. `size_t`. Very easy to miss the tree in that forrest!

PR #11250, opened 2022-04-19
[MU4] Fix some more compiler warnings

Jojo-Schmitz:
* MinGW: reg. "comparison of integer expressions of different signedness: 'int' and 'size_t' {aka 'long long unsigned int'}", this one-line change fixes 51 of 94.
As `curIdx` being `-1` is used elsewhere to detect whether an operation is (not) undoable, that cast is unavoidable as far as I can tell
* MSVC: `size_t` vs. `int`, fixes 271 warnings of 1005

### DIFF
--- a/src/engraving/libmscore/bend.cpp
+++ b/src/engraving/libmscore/bend.cpp
@@ -184,7 +184,7 @@ void Bend::layout()
 
     mu::draw::FontMetrics fm(font(_spatium));
 
-    int n   = m_points.size();
+    size_t n   = m_points.size();
     qreal x = m_noteWidth;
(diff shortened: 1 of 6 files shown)


src/engraving/property/propertyvalue.h
```diff
@@ -45,6 +45,7 @@ enum class P_TYPE {
     // Base
     BOOL,
     INT,
+    LLU, // unsigned long long, size_t
```
Jojo-Schmitz, line 48: Better name needed?

igorkorsukov (approved):

PR #11259, opened 2022-04-20
[MU4] Fix more (MSVC) compiler warnings

Jojo-Schmitz:
reg. `size_t` vs. `int`, still not all (and some of these changes result in new warnings elsewhere), but 496 less than before (still 452 left).

### DIFF
--- a/src/plugins/api/score.h
+++ b/src/plugins/api/score.h
@@ -203,10 +203,10 @@ public:
 //      Q_INVOKABLE void updateRepeatList(bool expandRepeats) { score()->updateRepeatList(); } // TODO: needed?
 
     /// \cond MS_INTERNAL
-    int nmeasures() const { return score()->nmeasures(); }
-    int npages() const { return score()->npages(); }
-    int nstaves() const { return score()->nstaves(); }
-    int ntracks() const { return score()->ntracks(); }
+    int nmeasures() const { return static_cast<int>(score()->nmeasures()); }
+    int npages() const { return static_cast<int>(score()->npages()); }
+    int nstaves() const { return static_cast<int>(score()->nstaves()); }
+    int  ntracks() const { return static_cast<int>(score()->ntracks()); }
     /// \endcond
 
     /**
--- a/src/converter/internal/compat/notationmeta.cpp
+++ b/src/converter/internal/compat/notationmeta.cpp
@@ -60,8 +60,8 @@ mu::RetVal<std::string> NotationMeta::metaJson(notation::INotationPtr notation)
     json["poet"] =  poet(score);
     json["mscoreVersion"] =  score->mscoreVersion();
     json["fileVersion"] =  score->mscVersion();
-    json["pages"] =  score->npages();
-    json["measures"] =  score->nmeasures();
+    json["pages"] =  static_cast<int>(score->npages()); // no = operator for size_t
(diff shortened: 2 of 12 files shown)


src/plugins/api/score.h
```diff
@@ -89,13 +89,13 @@ class Score : public Ms::PluginAPI::ScoreElement
...
+    Q_PROPERTY(size_t nmeasures READ nmeasures)
     /** Number of pages (read only) */
-    Q_PROPERTY(int npages READ npages)
+    Q_PROPERTY(size_t npages READ npages)
     /** Number of staves (read only) */
-    Q_PROPERTY(int nstaves READ nstaves)
+    Q_PROPERTY(size_t nstaves READ nstaves)
     /** Number of tracks (#nstaves * 4) (read only) */
-    Q_PROPERTY(int ntracks READ ntracks)
+    Q_PROPERTY(size_t ntracks READ ntracks)
```
igorkorsukov, line 98: I'm not sure if this will work, Qt may not support size_t when put to Qml.

Jojo-Schmitz: Leaving them `int` here...

src/plugins/api/score.h
```diff
@@ -203,10 +203,10 @@ class Score : public Ms::PluginAPI::ScoreElement
...
 
     /// \cond MS_INTERNAL
-    int nmeasures() const { return score()->nmeasures(); }
-    int npages() const { return score()->npages(); }
-    int nstaves() const { return score()->nstaves(); }
-    int ntracks() const { return score()->ntracks(); }
+    size_t nmeasures() const { return score()->nmeasures(); }
+    size_t npages() const { return score()->npages(); }
+    size_t nstaves() const { return score()->nstaves(); }
+    size_t ntracks() const { return score()->ntracks(); }
```
Jojo-Schmitz, line 209: ... but changing to `size_t` here, would that work?

Jojo-Schmitz: It does build at least

Jojo-Schmitz: and without new warnings, as far as I can tell

src/plugins/api/score.h
Jojo-Schmitz: Hmm, there's an `ulonglong` available as a type on `Q_PROPERTY()`. Question is whether this is "platform-proof"

Jojo-Schmitz: Nope, that doesn't even compile

Jojo-Schmitz: with both, `size_t` and `int` it does

src/plugins/api/score.h
Jojo-Schmitz: Actually with `int` futher up, it does generate 4 new warnings here

igorkorsukov: I think, better keep `int`, and do like `int ntracks() const { return static_cast<int>(score()->ntracks()); }`

Jojo-Schmitz: But only here in this file, right?

Jojo-Schmitz: Done

igorkorsukov: > But only here in this file, right?

right, only for Qml Api

igorkorsukov (approved):

PR #11263, opened 2022-04-20
[MU4] Fix more compiler warnings

Jojo-Schmitz:
Down to 359 in MSVC, 49 in MinGW


src/notation/view/widgets/editstaff.cpp
```diff
@@ -278,7 +278,7 @@ void EditStaff::updateNextPreviousButtons()
 {
     int staffIdx = m_orgStaff->idx();
 
-    nextButton->setEnabled(staffIdx < (m_orgStaff->score()->nstaves() - 1));
+    nextButton->setEnabled(staffIdx < (static_cast < int > (m_orgStaff->score()->nstaves()) - 1));
```
Jojo-Schmitz, line 281: No idea why uncrustify wants it this way...

PR #11272, opened 2022-04-21
Slurs default to up when contained chords have mixed stem direction

asattely:
Slurs should default to upwards direction when they contain a mix of down- and up-stemmed notes. This fixes that.


asattely:
This PR has a solution that doesn't work, please don't merge it yet!

asattely:
There is a single vtest failing at the moment, which may require fixing (but possibly not). I'll know for sure tomorrow.

oktophonie (approved):

RomanPudashkin (approved):

PR #11280, opened 2022-04-21
[MU4] Custom Keysig transposable

sammik:
Resolves: #10517
Also Resolves: #10497 and [MS7440](https://musescore.org/en/node/7440)

This is more complex PR, which covers #10609 #10543

In_Action

It contains 3 commits.

### 1. KeySym refactor
Replace KeySym members pos and spos, by 
```
int line; 
double xPos; // in staff spatium units
```
**Why?**
`pos` and `spos` are "visual" members (spos.y depends on line distance), so same Key Signature had different Key Symbols spos values in different staff settings.
In custom key signatures, spos was (incorrectly) used as "logical" member instead, which caused visual bug #10497 in staffs with altered line distance.

`line` and `xPos` are "logical" members, which clean up code logic and brings more flexibility to Key Signatures.
This commit replaces PR #10609 which is adapted here to use std:vector instead of QList

### 2. Adaptation for clef - added new CustDef class

**Why?**
Relation between key and keysignature

Convential keys
`key` is definition, `key symbols` in `keysignature` represent it, related on `clef`
`keySymbol.line` is representation of key `symbols` in concrete `clef`

Custom keys
definition is missing, `key symbols` are used to "fake" definition, but `clef` relation is missing inside `keysignature`.

I added a new "definition" of customkey contains list of symbols, each defined as:
1.  note / degree (one of c, d, e, f, g, a, b, or 0, 1, 2, …) /used note at the begining, but changet it do degree, as more flexibile
2.  symbol (sharp, flat, natural, quatertone up, ...)
3.  xPos Alteration (default xPos depends of symbol's index, 1.st symbol 1sp, 2.nd 2sp), so on second symbol xAlt=-1.4 means it stays at 1. symbols position.
4.  octave alteration (default line position depends on conventional position as defined in clefTable in clef.cpp)
schematic idea is 
`Custom_Definitions * Clef -> Key_Symbols`
(This replaces #10621 in better way.)

### 3. Transpositions
a) NOT USED 
It is possible to transpose custom defiitions themself, everytime as relative step, but problem could be „open space“ – infinite increasing / decreasing, something like ##### could happen, but we have limited amount of symbols. Also we would need to take a care, to don’t chage initial key signature (symbols order, …) after transpositions there and back.
Schematic idea would be:
`KeySig_Definition * Clef -> Key_Symbols`
`KeySig_Definition * Transposition -> new KeySig_Definition`

**b) USED** 
We can understand user definition as root „`scale definition`“ (like in standard keys is major scale) and `key` as transpositions information.
Schematic idea is:
`Scale_Definition * Key -> KeySig_Definition (temporary) * Clef -> Key_Symbols`
or we can avoid middle step and get shroter 
`Scale_Definition * Key * Clef -> Key_Symbols`

I choosed b). I am not 100% sure, if it was right choice, but hope so.

- Most of accidentals are transposed like this:
 transposing-accidentals

- If symbol of transposed accidental is unknown, sharp, or flat is put before it.
 transposing-unknown

- Convential accidentals takes place first. 
convential-first

- If same note has custom accidental, convential accidental is dropped out. Adapt-accidental-C-G

- Naturals are used only when user explicitely select them (not by transpositions).
 drop-natural
(If I understood @oktophonie [Simons suggestion](https://discord.com/channels/818804595450445834/818855201959051275/963698591908659201) correctly.)

Key(signature) is stored:
```
<KeySig>
    <accidental>-1</accidental> // key
    <custom>1</custom> // is custom
    <CustDef> // list of accidentals
      <sym>accidentalDoubleSharp</sym>
      <def degree="3" xAlt="0" octAlt="0"/>
      </CustDef>
    <CustDef>
      <sym>accidentalThreeQuarterTonesFlatArrowDown</sym>
      <def degree="6" xAlt="0.699868" octAlt="0"/>
      </CustDef>
    </KeySig>

```

Vtest failure expected


Jojo-Schmitz:
MusicXML tests failing for reasons unrelated to this PR

sammik:
> MusicXML tests failing for reasons unrelated to this PR

Oh, thanks! 
I was just investigating, whats going on.

Jojo-Schmitz:
Fixed in #10442's 2nd commit

Jojo-Schmitz:
a Rebase should fix it

sammik:
vtest fail is good

before:
custom-keysig-3-1 ref
custom-keysig-4-1 ref

after:
custom-keysig-3-1
custom-keysig-4-1

PR #11281, opened 2022-04-21
[engraving] replaced QList to std. Part 11

igorkorsukov:


src/notation/view/widgets/timeline.cpp
```diff
@@ -1785,11 +1785,11 @@ int Timeline::getHeight() const
 //   Timeline::correctStave
 //---------------------------------------------------------
 
-int Timeline::correctStave(int stave)
+StaffIdx Timeline::correctStave(StaffIdx stave)
```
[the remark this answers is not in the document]
Jojo-Schmitz: timeline.h.line 223 actually

src/engraving/libmscore/stafftype.h
```diff
@@ -422,7 +422,7 @@ class StaffType
     bool isDrumStaff() const { return _group == StaffGroup::PERCUSSION; }
     // static functions for font config files
     static std::vector<QString> fontNames(bool bDuration);
-    static bool fontData(bool bDuration, int nIdx, QString* pFamily, QString* pDisplayName, qreal* pSize, qreal* pYOff);
+    static bool fontData(bool bDuration, size_t nIdx, QString* pFamily, QString* pDisplayName, qreal* pSize, qreal* pYOff);
```
handrok, line 425: Is nIdx fontIdx? Should be renamed to an explicit name

Jojo-Schmitz: Those variable names all use a prefix, b for bool, n for numeric, p for pointer, etc.
So changing just nIdx would certainly be wrong

src/engraving/layout/layoutpage.cpp
```diff
@@ -68,8 +68,8 @@ void LayoutPage::getNextPage(const LayoutOptions& options, LayoutContext& lc)
         lc.page = lc.score()->pages()[lc.curPage];
         std::vector<System*>& systems = lc.page->systems();
         lc.pageOldMeasure = systems.empty() ? nullptr : systems.back()->measures().back();
-        const int i = mu::indexOf(systems, lc.curSystem);
-        if (i > 0 && systems[i - 1]->page() == lc.page) {
+        const SystemIdx i = mu::indexOf(systems, lc.curSystem);
+        if (!mu::is_invalid_idx(i) && i > 0 && systems[i - 1]->page() == lc.page) {
```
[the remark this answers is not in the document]
Jojo-Schmitz: the `!mu::is_invalid_idx(i) && ` is not needed, as in the next step `i` is tested for being larger than 0, so it can't be -1.

src/engraving/libmscore/mscore.h
```diff
@@ -82,28 +84,26 @@ static constexpr int MSCVERSION = 400;
...
-static constexpr int VOICES = 4;
+static constexpr size_t VOICES = 4;
 
-static constexpr int INVALID_INDEX = -1;
-
-inline constexpr int staff2track(int staffIdx, int voiceIdx = 0)
+inline constexpr TrackIdx staff2track(StaffIdx staffIdx, VoiceIdx voiceIdx = 0)
 {
-    return staffIdx >= 0 ? staffIdx * VOICES + voiceIdx : INVALID_INDEX;
+    return !mu::is_invalid_idx(staffIdx) ? staffIdx * VOICES + voiceIdx : mu::invalid_idx;
```
RomanPudashkin, line 91: I would suggest adding mu::is_valid_idx to avoid such inverted conditions

src/engraving/layout/layoutpage.cpp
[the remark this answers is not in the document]
Jojo-Schmitz: Not sure what size you want checked?

src/engraving/libmscore/mscore.h
igorkorsukov: This will be incorrect. I can check that the index is definitely not valid, but this does not mean that it is valid, the validity of the index depends on the context of its use.

Example 1
```
size_t idx = mu::invalid_id;
```
This index is definitely not valid

Example 2
```
size_t idx = 9;
```
Is this a valid index? - it is impossible to say definitely

Example 3
```
std::vector<int> vec3 = {1, 2, 3};
size_t idx 3 = 9;

int v = vec.at(idx);
```
In this context, the index is definitely not valid

Jojo-Schmitz: Of course right. `mu::might_be_valid_idx()`  :-)

src/engraving/infrastructure/internal/engravingconfiguration.cpp
```diff
@@ -57,7 +57,7 @@ void EngravingConfiguration::init()
         m_scoreInversionChanged.notify();
     });
 
-    for (int voice = 0; voice < Ms::VOICES; ++voice) {
+    for (size_t voice = 0; voice < Ms::VOICES; ++voice) {
```
RomanPudashkin, line 60: VoiceIdx

src/engraving/libmscore/figuredbass.cpp
```diff
@@ -1230,7 +1230,8 @@ void FiguredBass::layoutLines()
         // probably be implemented.
     }
 
-    int i, len;
+    SystemIdx i;
```
RomanPudashkin, line 1233: Please add default values to these variables

src/engraving/libmscore/measure.cpp
```diff
@@ -1476,7 +1476,7 @@ bool Measure::acceptDrop(EditData& data) const
     PointF pos = data.pos;
     EngravingItem* e = data.dropElement;
 
-    int staffIdx;
+    StaffIdx staffIdx;
```
RomanPudashkin, line 1479: Please add default values to these variables

src/engraving/libmscore/measure.cpp
```diff
@@ -2472,9 +2472,9 @@ bool Measure::isCutawayClef(int staffIdx) const
     if (!empty) {
         return false;
     }
-    int strack;
-    int etrack;
-    if (staffIdx < 0) {
+    TrackIdx strack;
```
RomanPudashkin, line 2475: Please add default values to these variables

src/engraving/libmscore/figuredbass.cpp
igorkorsukov: There are a lot of places where there is no default value. I fixed it somewhere. But this is beyond the scope of the current task, so I did not edit everywhere.

src/engraving/libmscore/navigate.cpp
```diff
@@ -372,13 +372,13 @@ ChordRest* Score::upStaff(ChordRest* cr)
 ChordRest* Score::downStaff(ChordRest* cr)
 {
     Segment* segment = cr->segment();
-    int tracks = nstaves() * VOICES;
+    TrackIdx tracks = nstaves() * VOICES;
 
-    if (cr->staffIdx() == static_cast<int>(nstaves()) - 1) {
+    if (cr->staffIdx() == nstaves() - 1) {
```
RomanPudashkin, line 377: Will this code work correctly if the staff list is empty?

igorkorsukov: it will be the same as it was before

handrok (approved):

Jojo-Schmitz:
A couple code style fixes needed...

RomanPudashkin (approved):

Issue #10922, opened 2022-03-27
[MU4 Task] (Re-)Implement playback for multi-measure repeats


cbjeukendrup:
**Task description**
The new MPE playback rendering system does not yet support playback for _multi_-measure repeats (single-measure repeats do work). The old MIDI rendering system did (before it was replaced for playback; it is still used for MIDI export nowadays).
<img width="841" alt="Schermafbeelding 2022-03-27 om 15 25 00" src="https://user-images.githubusercontent.com/48658420/160283660-9631d895-d5df-444d-a16f-71bc05bb20c9.png">

cbjeukendrup:
@vpereverzev Actually, I think there are some significant differences between "measure repeats" and other kinds of repeats:
- measure repeats apply only to the staves on which they appear, while other repeats apply to all staves
- other repeats are about jumping through the score (as in, the playback cursor should jump), while in the case of measure repeats, the normal playback flow just goes on, and you only hear the music from previous measures being "echoed". 

Do you really think that RepeatSegment/RepeatList should be extended for this?

jeetee:
RepeatSegments are entirely **not** foreseen for this type of logic and indeed act as a playback roadmap for entire measures.
I'd rather expect the equivalent of the old renderPlayEvents(element) to do the substitution here. (just my 2cents)

RobFog:
There was support for playback of multi-measure before the playback architecture was replaced. Maybe the previous implementation can inform the new one.

Tantacrul:
I should point out that in the latest nightly, regular repeats aren't working. Don't know if this has also been intentionally switched off for now, or not.

<img width="518" alt="image" src="https://user-images.githubusercontent.com/21022311/177040168-ef8786fc-c768-40cd-b8c2-e785eb8535f8.png">

Tantacrul:
Moving this to Post Beta. For this release we only strictly need to get ordinary repeats working.

RobFog:
Now that the beta is out I hope this pull request gets reviewed soon. 🙃

cbjeukendrup:
@RobFog Unfortunately, there are performance issues with this PR. It turns out to be tough to find a good solution that is not a completely awful hack and doesn't affect performance badly. Other team members have looked at it too, but haven't found the perfect solution either. So I highly doubt that this will be merged soon. However, you're right that this maybe deserves a friendly bump :)

@vpereverzev FYI :)

Tantacrul:
[...]

PR #11285, opened 2022-04-21
(Re-)Implement playback for (Multi-)Measure Repeats

cbjeukendrup:
Resolves: #10922

Actually only the last commit contains the fix; the rest is about some small cleanups in Libmscore.

### DIFF
--- a/src/engraving/libmscore/repeatlist.cpp
+++ b/src/engraving/libmscore/repeatlist.cpp
@@ -26,7 +26,6 @@
 #include "marker.h"
 #include "measure.h"
 #include "score.h"
-#include "segment.h"
 #include "tempo.h"
 #include "types.h"
 #include "volta.h"
@@ -119,9 +118,9 @@ const std::vector<const Measure*>& RepeatSegment::measureList() const
 
 RepeatList::RepeatList(Score* s)
 {
-    _score = s;
-    idx1  = 0;
-    idx2  = 0;
+    m_score = s;
+    m_idx1  = 0;
+    m_idx2  = 0;
 }
 
 //---------------------------------------------------------
@@ -152,7 +151,7 @@ int RepeatList::ticks() const
 
 void RepeatList::update(bool expand)
 {
-    if (!_scoreChanged && expand == _expanded) {
+    if (!m_scoreChanged && expand == m_expanded) {
         return;
     }
 
@@ -162,7 +161,7 @@ void RepeatList::update(bool expand)
         flatten();
     }
 
-    _scoreChanged = false;
+    m_scoreChanged = false;
 }
 
 //---------------------------------------------------------
@@ -171,7 +170,7 @@ void RepeatList::update(bool expand)
 
 void RepeatList::updateTempo()
 {
-    const TempoMap* tl = _score->tempomap();
+    const TempoMap* tl = m_score->tempomap();
 
     int utick = 0;
     double t  = 0;
@@ -199,10 +198,10 @@ int RepeatList::utick2tick(int tick) const
     if (tick < 0) {
         return 0;
     }
-    unsigned ii = (idx1 < n) && (tick >= at(idx1)->utick) ? idx1 : 0;
+    unsigned ii = (m_idx1 < n) && (tick >= at(m_idx1)->utick) ? m_idx1 : 0;
     for (unsigned i = ii; i < n; ++i) {
         if ((tick >= at(i)->utick) && ((i + 1 == n) || (tick < at(i + 1)->utick))) {
-            idx1 = i;
+            m_idx1 = i;
             return tick - (at(i)->utick - at(i)->tick);
         }
     }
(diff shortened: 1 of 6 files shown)


src/engraving/libmscore/repeatlist.cpp
```diff
@@ -1027,6 +1031,49 @@ void RepeatList::unwind()
...
+    m_expanded = true;
+}
+
+const Measure* RepeatList::playbackMeasure(size_t staffIdx, const Measure* measure) const
+{
+    if (m_measureRepeatsMapping.find(staffIdx) == m_measureRepeatsMapping.cend()) {
+        return measure;
+    }
+
+    const MeasureRepeatsStaffMapping& staffMapping = m_measureRepeatsMapping.at(staffIdx);
```
vpereverzev, line 1043: better to use the result of m_measureRepeatsMapping.find(staffIdx) here

src/engraving/playback/playbackmodel.cpp
```diff
@@ -258,52 +258,74 @@ void PlaybackModel::updateEvents(const int tickFrom, const int tickTo, const tra
...
+
+                        collectChangesTracks(trackId, trackChanges);
                     }
+                }
+            }
 
-                    m_renderer.render(item, tickPositionOffset, ctx.appliableDynamicLevel(segmentStartTick + tickPositionOffset),
-                                      ctx.persistentArticulationType(segmentStartTick + tickPositionOffset), std::move(profile),
-                                      m_playbackDataMap[trackId].originEvents);
+            for (Ms::Segment* segment = measure->first(); segment; segment = segment->next()) {
```
vpereverzev, line 315: Don't quite understand the motivation. Now we have 2 loops for segments for each repeatSegment.
Performance is highly critical here, so I'd propose you to return previous implementation.

src/engraving/libmscore/repeatlist.cpp
```diff
@@ -1027,6 +1031,49 @@ void RepeatList::unwind()
...
+                const Measure* playbackMeasure = measure;
+                for (int i = 0; playbackMeasure && i < measure->measureRepeatNumMeasures(staffIdx); i++) {
+                    playbackMeasure = playbackMeasure->prevMeasure();
+                }
+
+                if (playbackMeasure) {
+                    if (staffMapping.find(playbackMeasure) != staffMapping.cend()) {
+                        playbackMeasure = staffMapping.at(playbackMeasure);
+
+                        Q_ASSERT(staffMapping.find(playbackMeasure) == staffMapping.cend());
```
vpereverzev, line 1070: always prefer to use the one from log.h

src/engraving/libmscore/repeatlist.cpp
```diff
@@ -1027,6 +1031,49 @@ void RepeatList::unwind()
     }
 
     updateTempo();
-    _expanded = true;
+    m_expanded = true;
+}
+
+const Measure* RepeatList::playbackMeasure(size_t staffIdx, const Measure* measure) const
```
vpereverzev, line 1037: Not quite undesrtand what's going on here. I mean, the method requires a measure and provides another measure which is considered as "playback" measure.

Could you explain your motivation? :)

cbjeukendrup: Suppose you have a score like this: 
Schermafbeelding 2022-03-27 om 15 25 00
For the first two measures, you will just hear those measures themselves. But for the third measure, you should hear the first measure. So, the third measure will appear as a key in the map, with the first measure as its value. 
The same goes for the fourth and second measure: for the fourth measure, you should hear the second measure, so the fourth measure will appear as a key with the second measure as its value. 

And since measure repeats are a per-staff thing, there will be a separate map for each staff.

RomanPudashkin:
I would suggest transferring all commits with refactoring to a separate PR

cbjeukendrup:
> I would suggest transferring all commits with refactoring to a separate PR

Done! https://github.com/musescore/MuseScore/pull/11423

src/engraving/libmscore/repeatlist.cpp
cbjeukendrup: There is not just an assert function in log.h; only IF_ASSERT_FAILED etc., but that is pointless in this case.
Also, Q_ASSERT is used in numerous places, also in new code, and also as the 'backend' of the things in log.h. 
Of course, we can replace all Q_ASSERT with something like MU_ASSERT which we #define to Q_ASSERT, but that's a separate job. Shall I do that?

src/engraving/playback/playbackmodel.cpp
cbjeukendrup: Let's give ourselves an overview:

Old situation:
```
for each measure:
    for each segment:
        for each track (segment->elist())
            render the element in that segment on that track
        render the metronome for that segment
```

New situation: 
```
for each measure: 
    for each staff:
        determine the "playback measure" for the combination of the measure and the staff
        for each segment of the "playback measure":
            for each track of that staff:
                render the element in that segment on that track
    
    (not specific to any staff)
    for each segment of the measure:
        render the metronome for that segment
```
The motivation is as follows: for each measure, we need to handle each staff separately, because we need to determine the playback measure based on the combination of staff and measure. And we can't do anything with segments before we have determined the playback measure. So the segments loop really needs to be nested inside the staves loop. And then we need one more segments loop to render the metronome, which is not specific to any staff.

src/engraving/playback/playbackmodel.cpp
```diff
@@ -241,69 +241,102 @@ void PlaybackModel::updateEvents(const int tickFrom, const int tickTo, const tra
...
+            Q_ASSERT(staffMeasureList.size() == measureList.size());
 
-                if (segmentStartTick > tickTo || segmentEndTick <= tickFrom) {
-                    continue;
-                }
+            for (size_t i = 0; i < measureList.size(); ++i) {
+                const Measure* measure = measureList.at(i);
+                const Measure* staffMeasure = staffMeasureList.at(i);
+
+                int measureTicksOffset = repeatTickOffset + (measure->tick().ticks() - staffMeasure->tick().ticks());
```
cbjeukendrup, line 271: This is still quite ugly, but it is necessary to get `measureTicksOffset`. Has anyone a suggestion how I could solve this in a different way?

cbjeukendrup:
Rebased to fix conflicts.

Jojo-Schmitz:
rebase needed

PR #11290, opened 2022-04-21
[MU4] flac_thirdparties

vpereverzev:


RomanPudashkin (approved):

thirdparty/opus/CMakeLists.txt
```diff
@@ -269,8 +269,6 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "(armv7-a)")
   add_sources_group(opus celt ${celt_sources_arm})
 endif()
 
-target_compile_definitions(opus PRIVATE -D__OPTIMIZE__)
-
```
Jojo-Schmitz: Iis that removal intentional? You added those like just one commit before.

Jojo-Schmitz: It seems causing one of the (surprisingly few) warning MinGW gives.
Adding it back though results in compiler __errrors__ in MinGW (but not in MSVC)

PR #11301, opened 2022-04-22
[MU4] Fix Compiler warnings

Jojo-Schmitz:
In thirdparty/opus and with MSVC there are 106 warnings C4244 (conversion from 'xxx' to 'yyy', possible loss of data), 69 warnings C4310 (cast truncates contant value) and 9 warnings C4701 (potentially unititialized local variables), I think we should not deal with those, but disable them?
Edit: looks like #11346 is going to fix that.

Then there are 5 MinGW warnings about invalid PCH, which I believe to stem from thirdparty/flac, so also something we should probably not deal with other than to disable them.

Leaves us with (currently, moving target) 773 warnings C4267 (conversion from 'size_t' to 'int', possible loss of data)...


Jojo-Schmitz:
More to follow (there's enough of them left :-()

Jojo-Schmitz:
In thirdparty/opus and with MSVC there are 106 warnings C4244 (conversion from int to xxx, possible loss of data), 69 warnings C4310 (cast truncates contant value) and 9 warnings C4701 (potentially unititialized local variables), I think we should not deal with those, but disable them?

Jojo-Schmitz:
Leaves us with 948 warnings C4267 (conversion from 'xxx'' to''yyy', possible loss of data

Jojo-Schmitz:
Then there are 5 MinGW warnings about invalid PCH, which I believe to stem from thirdparty/flac, so also something we should prpbably not deal with other than to disaböe them.

src/engraving/libmscore/barline.cpp
```diff
@@ -124,7 +124,7 @@ static void undoChangeBarLineType(BarLine* bl, BarLineType barType, bool allStav
                 BarLine* sbl = toBarLine(e);
                 for (Staff* lstaff : sbl->staff()->staffList()) {
                     Score* lscore = lstaff->score();
-                    int ltrack = lstaff->idx() * VOICES;
+                    staff_idx_t ltrack = lstaff->idx() * VOICES;
```
cbjeukendrup, line 127: `track_idx_t`?

Jojo-Schmitz: Well, `lstaff->idx()` gives a `staff_idx_t`, it is chat the variable __name__ `ltrack` that suggests it ti be a track (and it is later used as such too. Under the cober both types are `size_t` anyway, so ti doesn't really matter, or does it?

Jojo-Schmitz: Well, `lstaff->idx()` gives a `staff_idx_t`, it is just the variable __name__ `ltrack` that suggests it to be a `track_idx_t` (and it is later used as such too). Under the cover both types are `size_t` anyway, so it doesn't really matter, or does it?

Jojo-Schmitz: So it seems a matter of coding style, use the type suitable for initialization, or the type for the later use (and cast if need be on the other occasions, not needed here though)?

Jojo-Schmitz:
Darn, no idea what I did to break that unit test "copypasteTuplet01"

igorkorsukov:
@Jojo-Schmitz Oh, I'm doing the same thing to finish my task. Now we will have a lot of conflicts and some of the work will be lost, how do we solve this problem now? It is necessary that one person completes the correction.

Jojo-Schmitz:
Just go ahead with your's. I'll check the conflicts after.
Would be great though is I'd get an explanatiion as to why the unit test fails

Jojo-Schmitz:
Maybe you could cherry-pick the first 4 commits?

PR #11308, opened 2022-04-22
[MU4] crashes_in_new_parts

RomanPudashkin:


DmitryArefiev:
Tested the crash with hiding all instruments in part with added system text on Win10, Mac12, LinuxMint 20.2 - **FIXED**

igorkorsukov (approved):

Issue #10522, opened 2022-02-10
[MU4 Issue] Changing Score properties settings doesn't make score dirty and some fields don't save


DmitryArefiev:
**Describe the bug**
1. Changing Score properties settings doesn't make score dirty (see video to compare with MU3)

2.  Some fields don't save - **workTitle** and **creationDate**. If they shouldn't be editable then should be greyed out

https://user-images.githubusercontent.com/90187801/153402027-eb069153-7491-41bb-97d6-d6cd9aaed78a.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

sammik:
Possibly related info at forum.
https://musescore.org/en/node/327924#comment-1110792

Jojo-Schmitz:
I disagree on `creationDate` being readonly. Needed when importing a file, which sets the creatinDate to the import date (as thatr is whxn this score gets created) rather than the creation date of the imported file, so may need to get corrected

cbjeukendrup:
Then we either need to come up with a decent date picker instead of just a textfield, or we should always treat it as plain text rather than trying to parse it as an ISO date (see [src/project/internal/notationproject.cpp:651](https://github.com/musescore/MuseScore/blob/37f1382e992cd23f395c5a766e580a3e76e7265c/src/project/internal/notationproject.cpp#L651) and [src/engraving/libmscore/page.cpp:463](https://github.com/musescore/MuseScore/blob/37f1382e992cd23f395c5a766e580a3e76e7265c/src/engraving/libmscore/page.cpp#L463)).

Jojo-Schmitz:
Either way is fine by me.

HemantAntony:
This is likely solved by the new Project Properties Dialog (#10630)

PR #11312, opened 2022-04-23
Fixes and simplifications for the Score Properties dialog

cbjeukendrup:
- Resolves: #10522

- Simplifies the dialog, by eliminating the concept of the dialog being "dirty" and having a "save" button. It made the dialog unnecessarily complicated for the user, and could cause confusion about closing the dialog vs closing the score. Removing it brings this dialog in line with other dialogs throughout the app. 

- Replaces the buttons at the bottom of the dialog with a QDialogButtonBox, fixing a problem with the order of the buttons on some platforms. 
(I had to fight a bit to get the Tab order right; finally, I decided to go the hard way of setting the tab order for every widget, because anything more elegant didn't work, probably due to bugs in Qt.)

- (Also includes a tiny color fix proposed by Jessica.)

### DIFF
--- a/src/converter/internal/compat/backendapi.cpp
+++ b/src/converter/internal/compat/backendapi.cpp
@@ -721,7 +721,7 @@ Ret BackendApi::updateSource(const io::path& in, const std::string& newSource, b
     ProjectMeta meta = project.val->metaInfo();
     meta.source = QString::fromStdString(newSource);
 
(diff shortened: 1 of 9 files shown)


cbjeukendrup:
After @HemantAntony's great work on rewriting this dialog in QML, this PR is not necessary anymore, so I'll close it!

Issue #10732, opened 2022-03-03
[MU4 Issue] Rewind and Play buttons don't behave as in MU3 when barline, line, slur etc. are selected 


DmitryArefiev:
( **Bug 1** )
When barline (or repeat line)  is selected,  Rewind button doesn't move selection to the first note (rest) of score - **UPDATE**: **NOT A BUG**, expected in **MU4**

**Bug 2** (most important)
When barline (or repeat line), lines (hairpin, ped), slur etc. are selected, Play button doesn't start playback from that location

https://user-images.githubusercontent.com/90187801/156643760-a723c477-d067-4a7f-b91e-d6120673e275.mp4

https://user-images.githubusercontent.com/90187801/157649478-30e47944-d6ea-42e7-bcfe-876f6b1348e6.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
I'm not sure about Bug 1, is this intentional or not for MU4.  But it doesn't correspond to MU3 behavior.

cbjeukendrup:
@DmitryArefiev I have indeed experienced cases where the space bar does not start playback at all (nothing seems to happen). Specifically, that seems to happen when selecting an element that will show grips when it is selected, such as barlines, slurs, etc.. Is that what you mean? That is indeed a separate bug. 

About bug 1, I think that there is indeed an intentional change in behavior. In MU3, the rewind button seems to work indirectly, by selecting the first element and thus triggering the playback cursor to jump to that element. But in MU4, the rewind button directly talks to the playback system. 

However, I discovered that when there is a range selection, the rewind button will rewind to the beginning of the range, rather than to the beginning of the score. In the code, it looks like this was added deliberately by Roman in 965710a7b4697a700d50e1683ffe6c8e53744eb1, but I'm not sure if that is how @Tantacrul has it in mind.

Tantacrul:
I didn't specifically state that this is how it should work but I may have miscommunicated.

My feeling is that the rewind button should be 'dumb' (not contextual). However, I would like second opinions. What do you think @cbjeukendrup ?

cbjeukendrup:
I also think that the rewind button should always rewind to the beginning of the piece, or if looping is enabled, to the beginning of the loop. 
At the moment that you _make_ a range selection, the playback position is already rewinded to the start of that range selection, so it would be pointless or even confusing if the rewind button would also rewind to that selection (it would seem like the rewind button is broken, because nothing happens, because the playback position is already at the place you're "rewinding" to).

[...]

PR #11322, opened 2022-04-24
When clicking any element, seek its playback tick

cbjeukendrup:
Not just "playable" elements, i.e. notes, chords...

Resolves: #10732
Resolves: #11493

### DIFF
--- a/src/notation/inotationplayback.h
+++ b/src/notation/inotationplayback.h
@@ -54,6 +54,7 @@ public:
     virtual midi::tick_t secToPlayedTick(float sec) const = 0;
     virtual midi::tick_t secToTick(float sec) const = 0;
 
+    virtual RetVal<midi::tick_t> playPositionTickByRawTick(midi::tick_t tick) const = 0;
     virtual RetVal<midi::tick_t> playPositionTickByElement(const EngravingItem* element) const = 0;
 
     enum BoundaryTick : midi :: tick_t {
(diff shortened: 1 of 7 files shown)


cbjeukendrup:
@DmitryArefiev Oops! Should be better in the next build :)

DmitryArefiev:
> @DmitryArefiev Oops! Should be better in the next build :)

Сan you rebase please? The fix looks good on Mac, but some strange things on Linux.. And the build Win doesn't start because of libsndfile issue.

Thanks!

cbjeukendrup:
@DmitryArefiev Done!

DmitryArefiev:
Tested #10732 on Win10, Mac12, LinuxMint 20.2 - ** Looks good now. FIXED**      

Other related issue I will log separately

src/notation/inotationplayback.h
```diff
@@ -56,6 +56,7 @@ class INotationPlayback
 
     virtual RectF playbackCursorRectByTick(midi::tick_t tick) const = 0;
 
+    virtual RetVal<midi::tick_t> playPositionTickByRawTick(midi::tick_t tick) const = 0;
```
RomanPudashkin, line 59: The purpose of this method is quite unclear. What is the difference between "regular" and "raw" ticks?

cbjeukendrup: It's basically a wrapper around `score()->repeatList().tick2utick(tick)`, which converts from "ticks without considering repeats" (i.e. just incrementing linearly from start to end of score) to "ticks with considering repeats" (i.e. unrolling all repeat sections before `tick`). Do you have a better idea for a name?

DmitryArefiev:
> Other playback related issues I will log separately

Found related #11674

DmitryArefiev:
@RomanPudashkin Please review :)

PR #11346, opened 2022-04-25
Reduced warning show for MSVC

igorkorsukov:
I tried to eliminate all warnings with one big PR, but the practice has shown that this is very dangerous. So I decided to temporarily turn off warnings for now (about int vs size_t) and fix them with small PRs


src/engraving/CMakeLists.txt
```diff
@@ -221,3 +221,11 @@ set(MODULE_LINK ${ENGRAVING_INFRASTRUCTURE_LINK} qzip)
 include(SetupModule)
 
 set_property(TARGET ${MODULE} APPEND PROPERTY AUTOMOC_MACRO_NAMES "BEGIN_QT_REGISTERED_ENUM")
+
+if (MSVC)
+    # temporarily disabled some warnings (int vs size_t)
+    target_compile_options(${MODULE} PUBLIC /wd4244)
+    target_compile_options(${MODULE} PUBLIC /wd4267)
```
Jojo-Schmitz, line 228: This is the big one. IMHO the other 3 warnings are easily fixed, see the 2nd to 4th commit of #11301

RomanPudashkin (approved):

src/engraving/CMakeLists.txt
Jojo-Schmitz: Actually only that one line is needed, the other warnings don't need to get disabled, as they don't show in the first place (that is even if not ignored)

Issue #9964, opened 2021-12-03
Expand click+drag interaction to all palette elements 


bkunda:
**Task description**
@vpereverzev, @RomanPudashkin, @Eism, @igorkorsukov we would be interested to know how feasible it is to implement this task for MS4 🙂.

The basic idea is that when the user clicks and drags a palette object onto their score, the pink connector line should immediately find the nearest applicable score object and attach itself to that. When the connector line is showing, the user can release their mouse click (even if they are a long way away from the score) and the object will be applied in the correct position. 
(Note this behaviour is almost identical to how hairpins and other line objects currently work. We would just like to expand it out to _all_ palette objects). 

Using the connector line to show the target score object also means we can remove some of the current cursor iconography, including the red "unavailable" icon in Windows (see pp. 16-17 in the spec for details).

Here is the spec: 
[UX-Spec_Palette-drag-interaction_v.1.pdf](https://github.com/musescore/MuseScore/files/7649385/UX-Spec_Palette-drag-interaction_v.1.pdf)

Let me know what you think! 
Thanks

Tantacrul:
I am bumping this to 4.x.

Annoying, since this is a terrible, terrible problem! However, it isn't a regressions, since the problem exists in MS3.6

PR #11364, opened 2022-04-25
Added connector lines for slurs

HemantAntony:

This commit creates connector lines to the nearest object when you drag a slur from Palettes

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -1099,6 +1099,7 @@ bool NotationInteraction::isDropAccepted(const PointF& pos, Qt::KeyboardModifier
     case ElementType::TRILL:
     case ElementType::HAIRPIN:
     case ElementType::TEXTLINE:
+    case ElementType::SLUR:
         return dragTimeAnchorElement(pos);
     case ElementType::IMAGE:
     case ElementType::SYMBOL:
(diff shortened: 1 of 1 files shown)


HemantAntony:
@bkunda This implements a small part of #9964. This my first time doing things in engraving which is why I created this PR to know if there is anything I can improve on

Jojo-Schmitz:
A rebase should fix the failing MusicXML tests

Tantacrul:
This is very promising!

One thing that would massively elevate it and significantly reduce confusion is if objects only try to snap to things that are relevant.

For example, in this video, the slurs should avoid trying to attach to blank measures. I understand they are using the same logic as hairpins... but I think there will be a few items that will need tweaked logic to make sure they don't try to snap to too many things that will produce a null result.

https://user-images.githubusercontent.com/21022311/165522741-c4b7b3a7-4b27-40ee-a4da-c7e1c1f660e0.mov

HemantAntony:
I did think of that. But I though hairpins also had wrong logic. It didn't make sense to me when I added a hairpin to a rest

Jojo-Schmitz:
What about "open ended slurs"? Not every slur needs to start or end on a note/chord.

HemantAntony:
> @HemantAntony - yes. perhaps we could remove the auto-snapping to empty measures across the board and see what that's like?

Okay, I'm working on a solution now. It should find the nearest element that can accept the required element (in this case, a slur)

HemantAntony:
> @HemantAntony - yes. perhaps we could remove the auto-snapping to empty measures across the board and see what that's like?

Okay, I'm working on a solution now. It should find the nearest element that can accept the required element (in this case, a slur)

HemantAntony:
I'm closing the PR as this issue is a bit difficult for me

Issue #11350, opened 2022-04-25
[MU4 Task] New colour definition - Hyperlink


jessjwilliamson:
Tagging @Tantacrul into this too as it has brought to our attention that the hyperlinks in some dialogs and other places contain hyperlinks with no colour being defined. 

Just added the 'HYPERLINK_COLOUR' in the Figma Master colour palette in all 4 themes.

Here are some examples with hyperlinks with updated colours.
**Another to include is the 'About Qt...' dialogue**

![image](https://user-images.githubusercontent.com/71656189/165105294-3ce6ed7e-6c2c-418c-a031-7c92a2bbda5b.png)
![image](https://user-images.githubusercontent.com/71656189/165105316-7e6b7b48-22d3-443f-b7c4-d193be8e4148.png)
![image](https://user-images.githubusercontent.com/71656189/165105532-90081771-9594-40cf-9c4b-07510b7c6f7c.png)

boryskuzmenko:
**FIXED**

Issue #11338, opened 2022-04-25
[MU4 Issue] New red and yellow accent colour in light mode


jessjwilliamson:
Red and yellow accent colours in **light mode** have been updated, also approved by @Tantacrul 

Red: #F28585
Yellow: #E0CC87

boryskuzmenko:
**FIXED**

Issue #11337, opened 2022-04-25
[MU4 Issue] Drop shadow less visible vertically/underneath pop-up


jessjwilliamson:
**Describe the bug**
Me and @Tantacrul noticed recently in most pop-ups that there is a lot of drop shadow going on on both horizontal sides but much less on the bottom of the pop-up/vertically. 
The issue is affecting the dark mode the most - it does not 'pop' very well.
Light mode is much more visible to see the issue. 

The bottom needs some more shadow intensity, around same value as the sides

![image](https://user-images.githubusercontent.com/71656189/165076854-81fce5ef-016a-49e3-900f-e3767d2605f2.png)

This is how the drop shadow should look like here:
![image](https://user-images.githubusercontent.com/71656189/165079922-8b7eaa27-4788-48d0-a5be-033df2c5a9df.png)

Tantacrul:
Making a P2 because it helps popups look more like popups in dark mode (they are hard to distinguish from the BG)

boryskuzmenko:
**FIXED**

PR #11369, opened 2022-04-25
More color tweaks

cbjeukendrup:
- Update selected state color of NotationSwitchButton
- Resolves: #11337
- Resolves: #11338
- Resolves: #11350
- Resolves: #11400
- Resolves: #11402 
- Some other minor tweaks to drop shadows to make them match Figma more closely


RomanPudashkin (approved):

cbjeukendrup:
@jessjwilliamson It turns out that you will need to reset the theme to see the effect, using the "Reset to default" button at the bottom of the Appearance page in Preferences. You'll need to do that for each theme that you have modified. After that, you will see the new link color.

jessjwilliamson:
> @jessjwilliamson It turns out that you will need to reset the theme to see the effect, using the "Reset to default" button at the bottom of the Appearance page in Preferences. You'll need to do that for each theme that you have modified. After that, you will see the new link color.

Ahh thank you! Fixed :)

PR #11394, opened 2022-04-26
Improvements to grace notes spacing

mike-spa:
Grace notes are currently handled very primitively in Musescore. My previous horizontal spacing PR fixed some of the most outstanding issues, namely the fact that grace notes were tucking above time- and key signatures (and they shouldn't) but were not tucking above notes from a different voice (and they should).

This PR fixes most of the remaining problems with grace notes spacing:
- Grace notes are too close to each other
- Grace notes are _way_ too close to the main note
- Accidentals cause unnecessary wide gaps
- Grace-notes-after cause a totally pointless space gap in the second voice which is quite funny.
- Grace-notes-after have no collision checks with the other voices.

I order to achieve this, the code for grace notes layout needed to be substantially scrapped and rewritten. Most notably: there aren't anymore separate methods for the layout of grace-notes-before and grace-notes-after. Grace-notes-after are "attached" to the following segment and spaced as if they were grace-notes-before of that segment. This is quite elegant, as now all grace notes share one, very simple, layout function. The only delicate part is making sure that this operation of "attaching" grace notes to the appropriate segment is performed at the right moment in the code and updated correctly upon every user interaction.

Needs review and testing by @oktophonie.

Before:
MU3GraceNotes
After:
Screenshot from 2022-04-26 14-17-05

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -3220,7 +3220,7 @@ void Measure::layoutMeasureElements()
         }
 
         // After the rest of the spacing is calculated we position grace-notes-after.
-        s.positionGraceNotesAfter();
+        LayoutChords::repositionGraceNotesAfter(&s);
 
         for (EngravingItem* e : s.elist()) {
             if (!e) {
@@ -3640,11 +3640,6 @@ qreal Measure::createEndBarLines(bool isLastMeasureInSystem)
                 m_segments.insert(s1, s2);
             }
         }
-    }
-
-    // May have grace notes attached to it so we need to lay them out
-    for (unsigned stfIdx = 0; stfIdx < score()->staves().size(); ++stfIdx) {
-        LayoutChords::layoutGraceNotes(seg, stfIdx);
     }
 
     // fix segment layout
@@ -4572,7 +4567,7 @@ void Measure::stretchMeasureInPracticeMode(qreal targetWidth)
             continue;
         }
         // After the rest of the spacing is calculated we position grace-notes-after.
-        s.positionGraceNotesAfter();
+        LayoutChords::repositionGraceNotesAfter(&s);
         for (EngravingItem* e : s.elist()) {
             if (!e) {
                 continue;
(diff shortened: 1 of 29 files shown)


oktophonie (approved):

src/engraving/libmscore/shape.cpp
```diff
@@ -102,6 +102,24 @@ bool Shape::sameVoiceExceptions(const EngravingItem* item1, const EngravingItem*
     return false;
 }
 
+bool Shape::graceToMainExceptions(const EngravingItem* item1, const EngravingItem* item2) const
+{
+    // Handles exceptions in the cases of main-to-grace and grace-to-main note spacing
```
vpereverzev, line 107: I'd suggest you to rewrite it on the following manner:

if (!item1 || !item1->isNote()) {
    return false;
}

....

filter out the very common cases first and then go down to details

vpereverzev: you're free to change it in the next PRs, tho

src/engraving/libmscore/measure.cpp
```diff
@@ -4331,6 +4336,33 @@ void Measure::computeWidth(Fraction minTicks, qreal stretchCoeff)
     }
 }
 
+void Measure::updateGraceNotes()
```
vpereverzev, line 4339: I believe this method should be somewhere at layoutchord.cpp

src/engraving/libmscore/segment.cpp
```diff
@@ -2620,4 +2626,49 @@ Fraction Segment::shortestChordRest() const
     }
     return shortest;
 }
+
+/* layoutGraceNotes()
+ * Needs to be called when creating the segment shape. Make sure it is called *last*,
+ * so that grace notes can be correctly padded against the segment shape.*/
+void Segment::layoutGraceNotes(int staffIdx)
```
vpereverzev, line 2633: https://github.com/musescore/MuseScore/pull/11394/files#r866118008

the same story about this one

mike-spa:
Code has been updated following @vpereverzev comments and rebased on latest master. Ready to be merged if all is ok

vpereverzev (approved):

PR #11409, opened 2022-04-26
[MU4] Fix #11310: Selection filter should not be enabled by default when users first launch MuseScore 4

### LINKED ISSUE #11310: Selection filter should not be enabled by default when users first launch MuseScore 4.
**Describe the bug**
Recently, the selection filter became enabled by default and tabbed with the other panels. 

We shouldn't do this. It's a niche panel with limited usefulness and should be kept off unless the user really needs it. It is also less crucial now that our properties panel allows you to edit specific types of object without having to deselect every other type of object.

In addition, the interface looks cluttered with the selection filter enabled and will introduce another layer of complexity for new users. Let's put it back where it belongs!

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11310


Eism (approved):

DmitryArefiev:
Tested #11310 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #11438, opened 2022-04-27
[MU4] Fix Compiler warnings

Jojo-Schmitz:
* Fix MSVC compiler warnings reg. conversion from `size_t` to `int` (C4245)
* Fix MinGW compiler warning reg. wrong format specifier
* Fix MinGW compiler warnings reg. comparison of integer expressions of different signdedness (-Wsign-compare)
* Fix MSVC compiler warnings reg. variable declarations hiding global declarations (C4459) resp.  class members (C4458)
 
Yes, @igorkorsukov, I know patience and all that ;-)
But these show in MSVC even with C4267 being disabled and also in MinGW

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -2452,7 +2452,7 @@ EngravingItem* Score::move(const QString& cmd)
                 }
             }
             // segment for sure contains chords/rests,
-            int size = int(seg->elist().size());
+            size_t size = seg->elist().size();
             // if segment has a chord/rest in original element track, use it
             if (track < size && seg->element(track)) {
                 trg  = seg->element(track);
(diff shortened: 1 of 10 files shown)


src/engraving/libmscore/cmd.cpp
```diff
@@ -2452,15 +2452,15 @@ EngravingItem* Score::move(const QString& cmd)
                 }
             }
             // segment for sure contains chords/rests,
-            int size = int(seg->elist().size());
+            size_t size = int(seg->elist().size());
```
igorkorsukov, line 2455: now we can remove the cast to  `int`

Jojo-Schmitz: oops, sure, gimme a couple minutes...

PR #11444, opened 2022-04-28
{MU4] Fix finding libsndfile for MSVC

Jojo-Schmitz:
As discussed in https://github.com/musescore/MuseScore/pull/11143#issuecomment-1111934130 ff.


igorkorsukov (approved):

Jojo-Schmitz:
Mind to update dependencies.z7 with the latest libsndfile stuff now?
So we're really all benefitting from this, and not just those who build locally and know how to get that stuff done?

cbjeukendrup:
Personally, I'm not against. I've been using libsndfile 1.1.0 (the latest version) for a while now on macOS, without any problems. So it might work on Windows too.

Jojo-Schmitz:
It does, as far as I can tell, using libsndfile 1.0.31, (which I believe was the latest mid-March?)
Updating to 1.1.0 as we speak ;-)

igorkorsukov:
I would prefer to just get rid of libsndfile :), for this we need to rewrite the soundfont loader for fluidsynth 
In order not to use different libraries for the same thing in different places of the project, and to reduce dependencies

PR #11453, opened 2022-04-28
Slurs adjust horizontally for other slurs' endpoints

asattely:
image

Slurs now adjust slightly to avoid the above pictured issue. The distance they move may need to be changed after @oktophonie reviews it


oktophonie (approved):

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1246,6 +1246,7 @@ void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, boo
     if (segments.size() > 1) {
         //how far vertically an endpoint should adjust to avoid other slur endpoints:
         const qreal slurCollisionVertOffset = 0.65 * system->spatium();
+        const qreal slurCollisionHorizOffset = 0.15 * system->spatium();
```
RomanPudashkin, line 1249: What does this number mean?

asattely: That's the amount that the slurs move in opposite directions horizontally so as not to be running into each other

Issue #11387, opened 2022-04-26
[MU4 Issue] Instrument - Replace Instrument - Clef changes do not occur


lengthwave:
**Describe the bug**
When selecting Replace Instrument from the instrument menu and replacing with an instrument that uses a different default clef, no clef change occurs.

**To Reproduce**
1. Go to Instruments
2. Click on Instrument settings
3. Click on Replace Instrument
4. Select instrument with a default clef change
5. Instrument name and playback changes
6. Clef does not change

**Expected behavior**
When an instrument is replaced with another instrument that has a different default clef, instrument clef should change.

**Screenshots**

https://user-images.githubusercontent.com/30370609/165275450-76bacba9-d9b3-4018-8c45-a68601899370.mov

**Platform information**
OS: macOS 10.16
Arch.: x86_64
MuseScore version (64-bit): 3.6.2.548020600
revision: github-musescore-musescore-3224f34

lengthwave:
@Tantacrul - FYI

Tantacrul:
Confirmed. When replacing an instrument, using this panel, it should include information that gets added when you create an instrument change.

Example: 
- Create a new score and add a Horn
- Click on the second measure and drag an 'Instrument change' element from palettes, then choose to change to a cello
- Notice how a bass clef now gets added
<img width="529" alt="image" src="https://user-images.githubusercontent.com/21022311/165286262-96e6d7b6-a5b0-4945-970b-4f009e4d60b1.png">

However, when you use the instruments panel to 'Replace' an instrument (which is a master replace action, not an instrument change), the clef does not get added.

Example: 
- Create a new score and add a Horn
- Go to the instruments panel, click on the settings cog, choose 'Replace instrument' and pick a cello
- Notice that the cello is now being used but the bass clef hasn't been added

Tantacrul:
By the way, there is another 'Replace Instrument' option in the Staff/Part properties dialog that **actually does it right**

I think we should use this logic instead.

<img width="756" alt="image" src="https://user-images.githubusercontent.com/21022311/165289309-d7320ac4-e82c-42fc-875f-024337365bd3.png">

Issue #11386, opened 2022-04-26
[MU4 Issue] Instrument - Replace Instrument - Transposing to non-transposing instrument


lengthwave:
**Describe the bug**
When using the Replace Instrument function to change a transposing instrument to non-transposing the instrument is changed, but not the transposition.

**To Reproduce**
1. Create a score with a transposing instrument
2. Select the corresponding instrument from the Instruments tab
3. Click on Settings icon
4. Click on Replace Instrument
5. Select non-transposing instrument
6. See instrument change in score (but not key signature and transposition)

**Expected behavior**
Instrument change from transposing to non-transposing instrument updates the key signature and transposition

**Screenshots**

https://user-images.githubusercontent.com/30370609/165271988-2eeadcf7-e83b-49e0-9d25-7b7e59745e80.mov

**Platform information**
OS: macOS 12.2
Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2224366007
revision: github-musescore-musescore-9a7fabf

lengthwave:
@Tantacrul - FYI

Tantacrul:
It's worth trying this in MuseScore 3 to see how it behaves. In this case, the key signature for the oboe (C major/A minor) should be used.

DmitryArefiev:
Fixed in #11454

PR #11454, opened 2022-04-29
Fix changing transposition and clef when replacing instrument using Instruments Panel

cbjeukendrup:
Resolves: #11386
Resolves: #11387


RomanPudashkin (approved):

DmitryArefiev:
Tested #11386 #11387 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #11458, opened 2022-04-29
[MU4] Fix more compiler warnings, as seen in MinGW and MSVC

Jojo-Schmitz:
* Fix MSVC compiler warning reg. signed/unsigned mismatch, C4245
* Fix MSVC compiler warnings reg. conversion from `size` to `int`, possible loss of data, C4267
* Fix MinGW compiler warnings reg. -Wtype-limits, -Wsign-compare and -Wformat= caused by the above change
* Fix MSVC compiler warnings reg. truncation from `double` to `float`, C4305 and reg. unreachable code, C4702

### DIFF
--- a/src/engraving/libmscore/instrtemplate.cpp
+++ b/src/engraving/libmscore/instrtemplate.cpp
@@ -163,7 +163,7 @@ void InstrumentGroup::read(XmlReader& e)
             if (t == 0) {
                 t = new InstrumentTemplate;
                 t->articulation.insert(t->articulation.end(), articulation.begin(), articulation.end());             // init with global articulation
-                t->sequenceOrder = instrumentTemplates.size();
+                t->sequenceOrder = static_cast<int>(instrumentTemplates.size());
                 instrumentTemplates.push_back(t);
             }
             t->read(e);
@@ -475,7 +475,7 @@ void InstrumentTemplate::read(XmlReader& e)
             extended = e.readInt();
         } else if (tag == "staves") {
             staffCount = e.readInt();
-            bracketSpan[0] = staffCount;
+            bracketSpan[0] = static_cast<int>(staffCount);
 //                  for (int i = 0; i < staves-1; ++i)
 //                        barlineSpan[i] = true;
         } else if (tag == "clef") {             // sets both transposing and concert clef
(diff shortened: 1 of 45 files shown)


src/engraving/libmscore/instrtemplate.cpp
```diff
@@ -306,7 +306,7 @@ void InstrumentTemplate::write(XmlWriter& xml) const
     }
     stringData.write(xml);
     if (staffCount > 1) {
-        xml.tag("staves", staffCount);
+        xml.tag("staves", static_cast<int>(staffCount));
```
Jojo-Schmitz, line 309: Or a new `QVariant`? But how exactly?

igorkorsukov: It is better to use `int`, because there will be a value type check inside...

Very soon we will get rid of QVariant completely, this code will be changed

Jojo-Schmitz: Via your last PR #11510 that change is now in the repository anyway, as far as I can tell.

Jojo-Schmitz:
Finally! All compiler warnings related to `size_t` vs. `int` that were shown in MSVC and/or MinGW are gone.

src/engraving/libmscore/beam.cpp
```diff
@@ -1345,7 +1345,7 @@ void Beam::layout2(const std::vector<ChordRest*>& chordRests, SpannerSegmentType
             int bottomLastLine = bottomLast ? bottomLast->upNote()->line() : 0;
             if (chordRests.size() == 2 && chordRests[0]->staffMove() != chordRests[1]->staffMove()) {
                 // if there are only two notes, one on each staff, special case
-                qreal horizDistance = chordRests[1]->stemPosX() - chordRests[0]->stemPosX();
+                //qreal horizDistance = chordRests[1]->stemPosX() - chordRests[0]->stemPosX();
```
Jojo-Schmitz, line 1348: Commenting it out only, as I'm unsure how/whether it relates to the comment above

Jojo-Schmitz: @asattely ?

Jojo-Schmitz: Or does @Nick-Mazuk know something about this, as this change is a contination of #10130?

Jojo-Schmitz:
@igorkorsukov I really wouldn't mind getting this PR merged ;-)
Just had to rebase and adjust it once again...

src/engraving/libmscore/beam.cpp
Nick-Mazuk: I don't see that code anywhere in that file anymore, so I can't be of much help.

@Jojo-Schmitz for future reference, you can use `git blame` to see who added the code.

igorkorsukov (approved):

PR #11465, opened 2022-04-30
Fix #11382: Inconsistent text sizes

### LINKED ISSUE #11382: [MU4 Issue] Palettes - Tempo - Inconsistent text sizes
**Describe the bug**
In the Palettes - Tempo the text sizes for different items are inconsistent. Some giant, some tiny.

**To Reproduce**
1. Go to Palettes
2. Click on Tempo
3. Scroll through Palette items
4. See inconsistency of text size

**Expected behavior**
1. The below palette objects could take the following properties specifications by default, which would align their optical size with that of existing text objects (E.g. "Grave", "Largo"):
X: 0sp
Y: 0.4sp
Content scale: 1.3

- accel.
- allarg.
- rall....
- rit...
- Swing
- Straight
- calando
- lentando
- morendo
- precipitando
- smorz.
- sost...
- string.

2. Musical notes, which are currently taken from Leland Text, should be scaled at a ratio of 5:3 relative to their adjacent numerical tempo specification (See example below, where I have scaled the half-note in the first cell (marked "correct") to 26.5pt alongside the "= 80", which is set at 16pt).

**Screenshots**
CURRENT
Screen Shot 2022-04-26 at 12 03 57 PM

EXPECTED
tempo-palette-revised

(N.B. For demonstration purposes, only the first cell in the above example has the correct font size ratio of musical note to numerical tempo indicator.)  

**Platform information**
OS: macOS 12.2
Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2224366007
revision: github-musescore-musescore-9a7fabf

@HemantAntony you may want to have a look at this and #11383, as you did the most recent work on tidying-up palettes 🙂

HemantAntony:
Resolves: #11382 

This commit fixes the differences in the size of certain palette cells of the Tempo Palette by increasing magnification. This also fixes the differences in size of the note symbol and the text in Tempo Palette

### DIFF
--- a/src/engraving/libmscore/textbase.cpp
+++ b/src/engraving/libmscore/textbase.cpp
@@ -787,11 +787,11 @@ mu::draw::Font TextFragment::font(const TextBase* t) const
 
     QString family;
     if (format.fontFamily() == "ScoreText") {
-        if (t->explicitParent() && t->isDynamic()) {
+        if (t->isDynamic()) {
             family = t->score()->scoreFont()->fontByName(t->score()->styleSt(Sid::MusicalSymbolFont))->family();
             // to keep desired size ratio (based on 20pt symbol size to 10pt text size)
             m *= 2;
-        } else if (t->explicitParent() && t->isTempoText()) {
+        } else if (t->isTempoText()) {
             family = t->score()->styleSt(Sid::MusicalTextFont);
             // to keep desired size ratio (based on 20pt symbol size to 12pt text size)
             m *= 5.0 / 3.0;
--- a/src/palette/internal/palettecreator.cpp
+++ b/src/palette/internal/palettecreator.cpp
@@ -1345,14 +1345,16 @@ PalettePtr PaletteCreator::newTempoPalette(bool defaultPalette)
         auto item = makeElement<TempoChangeRanged>(gpaletteScore);
         item->setTempoChangeType(pair.first);
         item->setBeginText(pair.second);
-        sp->appendElement(item, pair.second);
+        sp->appendElement(item, pair.second, 1.3)->yoffset = 0.4;
     }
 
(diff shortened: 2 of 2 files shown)


HemantAntony:
@bkunda Please check this PR out. You may have to delete your workspaces file to apply changes

cbjeukendrup:
> You may have to delete your workspaces file to apply changes

Or just resetting the tempo palette, using the option in its '...' menu, could be enough.

HemantAntony:
> Or just resetting the tempo palette, using the option in its '...' menu, could be enough.
I didn't know that. That would have saved me some time :)

cbjeukendrup (approved):

bkunda:
Tested on MacOS. Looks good!

cbjeukendrup:
Yes... once you have made any changes to any of your palettes, you will need to reset them in order to see any changes that _we_ make. With Palettes, it seems even more complicated to merge the changes than with toolbars, so I'm afraid we'll have to live with it, in this case.

bkunda:
> Yes... once you have made any changes to any of your palettes, you will need to reset them in order to see any changes that _we_ make. With Palettes, it seems even more complicated to merge the changes than with toolbars, so I'm afraid we'll have to live with it, in this case.

Thanks for this clarification. Again, will this _only_ affect users who have previously downloaded a copy of MS4 (In any build version)? (I.e. users who have _never_ downloaded a copy of MS4 _won't_ have to reset their palettes, correct?)

cbjeukendrup:
> users who have never downloaded a copy of MS4 won't have to reset their palettes, correct?

Correct!

(and I believe, users who _have_ downloaded MS4 but never made any changes to their "workspace" also won't have to reset their palettes, but I'm not 100% sure about that.)

Issue #11264, opened 2022-04-20
[MU4 Issue] Text-lines: Property to allow user to adjust horizontal gap between text and line


rgreen5:
AFAICS, there is no way for the user to adjust the horizontal gap between text and its line in a text-line. All text-lines have the same default space. 

**Expected behavior**
A user-adjustable gap between text and line in text-lines.

**Screenshots**

Some published works favour the guitar barre line line flush to the text: 

![barre_line_(noad)](https://user-images.githubusercontent.com/87246583/164247712-1b11e5c2-85d0-4297-8aaa-3bf8182fe110.png)

![barre_line_(harris)](https://user-images.githubusercontent.com/87246583/164248383-4a9f413f-5ee7-4a85-a47c-6fdc49528f86.png)

Others have a space:

![barre_line_(noad1)](https://user-images.githubusercontent.com/87246583/164248854-011ab9d8-361d-4ef2-be70-010937b499c3.png)

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2059995262, revision: github-musescore-musescore-cafefd7

Tantacrul:
If I understand this bug correctly, then you can edit the vertical position of text on a line by selecting the line and going to the 'Properties' panel and selecting the 'Text' tab. Vertical offsets can be found there for most (if not all) text lines.

<img width="728" alt="image" src="https://user-images.githubusercontent.com/21022311/166006858-f4ffa50b-c983-4ae3-af48-55bf6b68a566.png">

If this doesn't answer your request then I'll reopen this bug. 

Thanks!
Martin

rgreen5:
Clicking on the "Text" tab _does_ give you access to a property called "Offset"  but, this should only affect vertical and _not_ horizontal alignment (that it affects both, currently, is a bug; see #11266).

So, AFAICS, there needs to be both a "Vertical offset" button _and_ a "Gap (horizontal)" button in the Text tab.

So, could this be reopened?

Tantacrul:
I don't think we need a horizontal offset option. It was overkill in MS3 and certainly the kind of thing that should live in an engraving styles section and not in the inspector. Also, there is already another bug about the diagonal positioning problem - I've recommended renaming the vertical position option there.

quinnouyang:
[...]

PR #11484, opened 2022-05-01
Fix and clarify vertical text offset for lines

### LINKED ISSUE #11266: [MU4 Issue] Text-lines: "Offset" doesn't work as it should
**Describe the bug**
If you create a text-line the "Offset" property causes the text to move at a diagonal, rather than vertically.

**To Reproduce**
1. Add a Staff text line from the line palette to the score.
2. With the text-line selected, in the Properties side bar, chose Text-line > Text.
3. Adjust the "Offset property"

RESULT: The text moves at a diagonal in relation to the line.

**Expected behavior**

The text should move vertically. In addition, IMV, the labelling should be better: "Vertical offset" for example.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2059995262, revision: github-musescore-musescore-cafefd7

quinnouyang:
Resolves  #11266, tangential to #11264.

Vertical text offset would incorrectly update both the X- and Y-positions, so I fixed the X-positions to `0` when updating. Also clarified the title from "Offset" to "Vertical offset."

Re. #11264: We were considering adding a horizontal offset option that purely moved the text (and largely ignored the line, e.g. allowing the text to overlap the line), but that feels pointless. Should further decide on what options to (not) include for line text.

Final result with example:
image

### DIFF
--- a/src/inspector/models/notation/lines/textlinesettingsmodel.cpp
+++ b/src/inspector/models/notation/lines/textlinesettingsmodel.cpp
@@ -93,7 +93,8 @@ void TextLineSettingsModel::createProperties()
         m_continiousText = buildPropertyItem(Ms::Pid::CONTINUE_TEXT);
 
         m_continiousTextVerticalOffset = buildPropertyItem(Ms::Pid::CONTINUE_TEXT_OFFSET, [this](const Ms::Pid pid, const QVariant& newValue) {
-            onPropertyValueChanged(pid, QPointF(0, newValue.toDouble()));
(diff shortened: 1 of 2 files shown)


Tantacrul:
Tested. Working great. Thanks!

RomanPudashkin:
Please fix the problems with the code style

quinnouyang:
Fixed the codestyling! (Sorry, fumbled around earlier from attempting to squash commits...)

PR #11485, opened 2022-05-01
fix #331007: [MusicXML import] support tie element

lvinken:
Resolves: https://musescore.org/en/node/331007

Create a tie on MusicXML import when either a tie or a tied element is present.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -105,6 +105,13 @@ static std::shared_ptr<mu::iex::musicxml::IMusicXmlConfiguration> configuration(
 //#define DEBUG_VOICE_MAPPER true
 
 //---------------------------------------------------------
+//   function declarations
+//---------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


Jojo-Schmitz:
@lvinken:  My backport to 3.x fails an mtest, "testUnterminatedTies" (see [here](https://github.com/musescore/MuseScore/runs/6245663871?check_suite_focus=true#step:5:808)), seems related to this change. any idea whow to deal with it?

Jojo-Schmitz:
That test doesn't seem to exist in master (see https://github.com/musescore/MuseScore/blob/a623af3beb5264655ab572a8ad9a812150dd5e38/src/importexport/musicxml/tests/tst_mxml_io.cpp#L255-L256)

RomanPudashkin (approved):

Jojo-Schmitz:
I see, that test isn't in 3.x either, but had been last added with #8556 into the 3.6.2_backend_musicxml branch

lvinken:
It is quite unfortunate that MusicXML import in the backend and the application itself are not in sync ...

Jojo-Schmitz:
Yes, it is. And unfortunate that the author of those is no longer around.

PR #11491, opened 2022-05-01
Remove SND options from Staff/Part Properties and hairpin playback popup

### LINKED ISSUE #11417: Remove option for 'Single Note Dynamics' in the Staff/Part Properties dialog
**Describe the bug**
This option (at the bottom left corner of the dialog) no longer functions, since we have replaced the entire engine and how it works. It was always a pointless option to expose to the user anyway. Good riddance!

image

quinnouyang:
Resolves #11417 

Title. Removed relevant XML from `editstaff.ui` and legacy code from `editstaff.cpp`. Also removed from `hairpinplaybackmodel`.

Noticed quite a bit of non-UI/functionality code surrounding SND, but I don't know if those are still necessary so I let them be.

## Staff/Part Properties
| Before | After |
| ------------- | ------------- |
| image  | image  |

## Hairpin Playback Popup
| Before | After |
| ------------- | ------------- |
| image  | image  |

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/general/playback/internal/HairpinsExpandableBlank.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/general/playback/internal/HairpinsExpandableBlank.qml
@@ -79,23 +79,12 @@ ExpandableBlank {
             minValue: 0
         }
 
-        CheckBoxPropertyView {
-            id: singleNoteCheckBox
-
-            navigation.name: "Use single note dynamics"
-            navigation.panel: root.navigation.panel
-            navigation.row: velocityChangeSection.navigationRowEnd + 1
-
-            text: qsTrc("inspector", "Use single note dynamics")
-            propertyItem: root.model ? root.model.useSingleNoteDynamics : null
-        }
-
         DropdownPropertyView {
             id: velocityChangeTypeSection
 
             navigationName: "Changes in dynamics range"
             navigationPanel: root.navigation.panel
-            navigationRowStart: singleNoteCheckBox.navigation.row + 1
+            navigationRowStart: velocityChangeSection.navigationRowEnd + 1
 
             titleText: qsTrc("inspector", "Changes in dynamics range")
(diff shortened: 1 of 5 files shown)


src/inspector/view/qml/MuseScore/Inspector/general/playback/internal/HairpinsExpandableBlank.qml
```diff
@@ -79,17 +79,6 @@ ExpandableBlank {
             minValue: 0
         }
 
-        CheckBoxPropertyView {
-            id: singleNoteCheckBox
```
RomanPudashkin, line 83: This id is still used in velocityChangeTypeSection

`navigationRowStart: singleNoteCheckBox.navigation.row + 1`

quinnouyang: Got it: replaced the id with the one directly above it. I assume this fixes it, but is there a difference between using `.navigation.row` and `navigationRowEnd`?

RomanPudashkin: You should use navigationRowEnd here:

navigationRowStart:  velocityChangeSection.navigationRowEnd + 1

RomanPudashkin: >is there a difference between using .navigation.row and navigationRowEnd?

Some controls in Inspector may occupy a range of navigation indexes. navigationRowEnd is needed to indicate the end of these occupied indexes

quinnouyang: Thanks, makes sense! Just ammended the commit with `navigationRowEnd`.

RomanPudashkin (approved):

Issue #11456, opened 2022-04-29
[MU4 Issue] Preferences > General > Languages - Check for Update button


lengthwave:
**Describe the bug**
The check for update button in Preferences > Language is a bit confusing.

The placement of the button suggests that the action is connected to submitting the language selection.

**To Reproduce**
1. Go to Preferences > General > Languages
2. Select language from dropdown
3.Click 'Check for updates'
4. See issue

**Expected behavior**
As the check for updates function is logically separate from the language selection action, it could be clearer if there is more visual distinction.

Maybe this can be moved to the Updates section of Preferences?

**Screenshots**
<img width="1212" alt="Screen Shot 2022-04-29 at 12 12 05 PM" src="https://user-images.githubusercontent.com/30370609/165916867-12540a61-2445-49a9-aec3-edb7a27811e2.png">

**Platform information**
OS: macOS 12.2
Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2243301402
revision: github-musescore-musescore-a6ab311

Tantacrul:
This update button is only for languages - hence why it is next to the languages dropdown. I suggest a simple rewording to 

**Check for language updates**

See the 'Updates' section lower down for the **app update**.

Tantacrul:
Marked as P2 because of potential (quite bad) misinterpretation on the behalf of the user.

RobFog:
Would it perhaps makes sense to just move the update checking button below the dropdown?

Tantacrul:
We can move it below and rename it too. Should be enough.

igorkorsukov:
renamed to "Check for language updates" 
https://github.com/musescore/MuseScore/pull/12455

PR #11507, opened 2022-05-03
Fix #11488 Add condition for endNote returning null.

### LINKED ISSUE #11488: Crash when undoing a split measure.
**To Reproduce**
1. Open this score: https://www.dropbox.com/s/dzwvbhobmxo9yzl/MS4_Advert_01.mscz?dl=0

2. Go to measure no.9 and select the following note in the Oboe
image

3. Select **Tools > Measures > Split measure before selected note/rest**

4. Observe that the measure is now split:
image

5. Now press **Ctrl+Z** undo the action (this regularly causes a crash)

6. If it doesn't crash then redo the split measure action again (this should produce the crash)

This crash is not 100% reproducible on new scores. However, once you perform the split action, it doesn't take long for the app to crash when trying to do something else. The score I provided above is one where the problem was most easily reproduced.

donarturo11:
Resolves: *([#11488 ](https://github.com/musescore/MuseScore/issues/11488))*

In some conditions, for example undoing split measure, endNote in tie class returns nullptr and tie condition is false positive. After adding the condition to if (tie) block this operation runs with no crash.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -2109,7 +2109,7 @@ void Chord::layoutPitched()
         // but the allocation of space needs to be performed here
         Tie* tie;
         tie = note->tieBack();
-        if (tie) {
+        if (tie && tie->endNote()) {
             tie->calculateDirection();
             qreal overlap = 0.0;
             bool shortStart = false;


Jojo-Schmitz:
Only the last commit belongs into this PR, the first is entirely unrelated (leongs into resp. stems from #11456) and the middle two are merges that don't belong here either.

donarturo11:
Please check if now is well.

sammik:
> Please check if now is well.

No, still 4 commits there.

If You rebased and Your local branch is OK, You need to force push it (`git push -f`)

donarturo11:
@sammik Thank you very much.

Tantacrul:
Tested. The example I gave in my issue is fixed. @Eism - if you are happy with this implementation, can you merge?

donarturo11:
Has anybody an idea, how to test, if startNote would be a null pointer? Thanks to @Tantacrul's score I found a possibility to being endNote as null pointer. Now I'm thinking if such condition is possible.

Issue #11516, opened 2022-05-04
[MU4 Task] Assign default shortcuts to selected note input commands for adding accidentals 


bkunda:
**Task description**
Would resolve #11500 

The suggestion here is to assign default shortcuts for the below actions as follows: 

- `+` for **Note input: sharp** (and rename action to **Toggle accidental: sharp**)
- `-` for **Note input: flat** (and rename action to **Toggle accidental: flat**)
- `=` for **Note input: natural** (and rename action to **Toggle accidental: natural**)

And while we're at it:

- Change the shortcut for ties to `T`. 

**Note also**: 
- Shortcut `-` should retain its existing functionality for text entry, while also being available for note input mode as described above.
- Existing shortcut `=` should be removed from "Toggle 'automatic placement' for selected elements". A replacement shortcut does not need to be designated at this stage.  

Let me know if there are any questions about this
Thanks!

quinnouyang:
There's two available shortcuts for each in `notationuiactions.cpp`, e.g. for sharps:
- `sharp`, "Toggle accidental: sharp"
- `sharp-post`, "Double ♯ (non-toggle)"
Ignoring the toggle option, I can set the default for `sharp-post`. To double-check, should I re-write these as `add-sharp` and `Add sharp` (I can try to sneak in the ♯ symbol like we do for note duraitons).

bkunda:
I can't actually advise on whether/how this should be re-labelled in the code. Perhaps @MarcSabatella or @cbjeukendrup might have an opinion on this.

But you have made me realise that the word "toggle" should probably be included in the UI label, because this is effectively what the shortcut does. I'm going to correct the issue description to reflect this, but yes the labels in the "Action" list in "Preferences > Shortcuts", should read in the format "Toggle accidental: sharp".

(Don't bother about sneaking in those symbols - they already appear in the left column, and they are problematic substitutes for the actual words from an accessibility and translation POV) 🙂

quinnouyang:
To clarify, I added those symbols in the left column.

Are we toggling or just adding? The differences between `sharp` and `sharp-post`, etc. is whether they toggle or just add.

On a UX POV, I kind of found the `+` shortcut a little jarring because it's the only one out of the three that require the `Shift` key. (And_ we're still figiuring out how shortcuts look and work through Preferences...)

bkunda:
Oh good. Those symbols are helpful in the left column 👍🏻.

[...]

PR #11536, opened 2022-05-05
Set default shortcuts for toggling accidentals and ties

quinnouyang:
Resolves: #11516 

Summary of changes:
- Set default toggles for `sharp` to `+`, `flat` to `-`, `natural` to `=`, and `tie` to `t` (used to be `+`). Removed default shortcut for toggling autoplace.
- Clarified descriptions and translations by replacing unicode symbols with text and specifying if the shortcut is a toggle

Please check if the translations/descriptions are good! Functionality seems to work on my end, **although there is no audio playback from toggling accidentals.** (New issue?)

image


Eism (approved):

Eism:
@bkunda please check the build from this PR

bkunda:
@quinnouyang this is coming along well!
Only one issue I found:
- On Mac, when using a QWERTZ keyboard (German), I am unable to trigger the "natural" command. In this particular case, I need to use `Shift` to access the `=` key, which doesn't seem to work. I also tried deliberately re-assigning the shortcut to `Shift`+`=` but this didn't work either. FWIW, I don't think it is a `Shift` key issue because I also need to use `Shift` on my QWERTY keyboard (because `=` and `+` occupy the same key), and this works fine.

quinnouyang:
@bkunda Thanks! For reference, I can only test with a Mac QWERTY keyboard.

I think the issue is on how we encode "secondary" characters on certain keyboards. From my experimentation, I cannot trigger shortcuts with `!`, `@`, `#`, etc. depiste `+` working, all of which physically require `Shift`+`[number/"primary" character]` **for me.**

I notice that we encoded some characters using their [references](https://dev.w3.org/html5/html-author/charref), e.g. `<` is `&lt;`. Uses of references vary by the shortcut file (keyboard layout). I tried experimenting by getting `!`, `@`, `#` to work by using their references only with no success.

On a tangent, my zoom shortcut doesn't work (`Ctrl`+`+`) on the latest master build. Probably for similar reasons...

bkunda:
> I think the issue is on how we encode "secondary" characters on certain keyboards

Sounds like this might need some further investigation then.
@cbjeukendrup @RomanPudashkin @Eism if you have any pointers here? Is this something we need to tackle internally?

bkunda:
@Eism I'm going to suggest we implement this and then raise the issue described above (about the Shift key not working on some – i.e. QWERTY – keyboards) as a separate issue. 

...unless of course the discussion about "secondary keys" above rings any alarm bells for you.

MarcSabatella:
@bkunda @Eism @Tantacrul Is there a reason not to merge this?  We'd really like to be moving faster on documentation and translation, but still having uncertainty about some of these basic shortcuts is getting in the way.  It would be good to settle on what the default shortcuts for 4.0 will be and make that happen before beta.

Also see https://github.com/musescore/MuseScore/pull/11854

To my knowledge, these two PR's represent the only pending changes to the default shortcuts being considered for 4.0, but if there are others, we'd really like to know before we get too much deeper into documentation and translation.

From my read of the code and understanding of the discussions, the following are the changes we are expecting:

Shortcuts being added
---
* Sharp: +
* Flat: -
* Natural: =

Shortcuts being changed
---
Tie: changing from + to T
Toggle insert mode: Ctrl+I remains the shortcut, but it's now the going to be *real* insert mode, not the weird undocumented buggy one it was in MU3

Shortcuts being removed
---
Toggle automatic placement (was =)

Tantacrul:
I'll leave this with @bkunda

Eism:
> I notice that we encoded some characters using their [references](https://dev.w3.org/html5/html-author/charref), e.g. < is &lt;. Uses of references vary by the shortcut file (keyboard layout). I tried experimenting by getting !, @, # to work by using their references only with no success.

@quinnouyang 
Yes, we should not use such characters in the shortcut file. And I have several PRs in which I fixed such shortcuts. 
So you need to replace `+` with `Shift+=`

bkunda:
> > I notice that we encoded some characters using their [references](https://dev.w3.org/html5/html-author/charref), e.g. < is <. Uses of references vary by the shortcut file (keyboard layout). I tried experimenting by getting !, @, # to work by using their references only with no success.
> 
> @quinnouyang Yes, we should not use such characters in the shortcut file. And I have several PRs in which I fixed such shortcuts. So you need to replace `+` with `Shift+=`

@quinnouyang please let me know when you've made this correction and I'll then re-test.

Eism (approved):

Eism:
@quinnouyang 
Rebase is needed

bkunda:
@Eism I'm afraid the change from using `+` to `Shift`+`=` has worsened the user experience of this fix:
- The regular `+` key on a numpad now no longer toggles the sharp accidental. This affects all users with a full-size keyboard
- QWERTZ users have an even worse experience, because on these keyboards the `+` is a dedicated key. This means that these users won't be able to access toggles for neither sharps nor naturals (naturals didn't work in this fix anyway because of the requirement to use the `Shift` key, which I understood was a separate issue).

Is it absolutely essential we use `Shift`+`=` instead of `+`? I'm afraid I don't understand the technical reasons for this, but if we have to adopt this then I think we risk reducing the discoverability of this new shortcut action, especially for users with numpads.

Eism:
@bkunda 
Does '+' work for you on macOS?

> The regular + key on a numpad now no longer toggles the sharp accidental. This affects all users with a full-size keyboard

We have problems with NumPad and should be solved in a separate PR
For example, https://github.com/musescore/MuseScore/issues/12791

> QWERTZ users have an even worse experience

Do we support QWERTY layout in MuseScore 4? Is there something I don't know? :)

HemantAntony:
fwiw, You might be using the wrong names? Isn't `Shift + =` the same as `+`. When you say `+`, you might be meaning just `=` :)

bkunda:
> @bkunda Does '+' work for you on macOS?
> 
> > The regular + key on a numpad now no longer toggles the sharp accidental. This affects all users with a full-size keyboard
> 
> We have problems with NumPad and should be solved in a separate PR For example, #12791
> 
> > QWERTZ users have an even worse experience
> 
> Do we support QWERTY layout in MuseScore 4? Is there something I don't know? :)

@Eism `+` previously did work for me on macOS _and_ on the numpad, before it was changed to `Shift`+`=`.

As for the QWERTZ layout, this was all working fine as well, with the exception of the new shortcut for `toggle accidental: natural`, which required the use of the `Shift` key. Apparently there is an issue with "secondary keys" ([described here](https://github.com/musescore/MuseScore/pull/11536#issuecomment-1128989958))

bkunda:
> fwiw, You might be using the wrong names? Isn't `Shift + =` the same as `+`. When you say `+`, you might be meaning just `=` :)

No when I refer to the `+` key I am literally referring to keys that only bear the symbol "+" (such as on a numpad), or on a QWERTZ keyboard the symbols "+" and "*" are on the same key. 

(FWIW On a QWERTZ keyboard, `=` and `0` share the same key).

Eism:
@bkunda @quinnouyang 
Sorry, guys, my mistake 
It is should be `+` instead of `Shift+=`

bkunda:
> @bkunda @quinnouyang
> Sorry, guys, my mistake
> It is should be `+` instead of `Shift+=`

ok all good then.
@quinnouyang if you can please revert to your previous solution
I'll then quickly test
And we can merge.

quinnouyang:
Made a new commit because there was a rebase after the old one:)

PR #11545, opened 2022-05-06
[engraving] Removed Q_OBJECT and Q_GADGET

igorkorsukov:


handrok (approved):

cbjeukendrup:
@Jojo-Schmitz Maybe it is because of the `#include "translation.h"` in scorefile.cpp. That makes the translation unit bigger, and apparently too big for MinGW.

igorkorsukov:
Now we can remove `#include "translation.h"` from `scorefile.cpp`, but in the future, we will need it.
I think it's better to disable the unit build for engraving because there are still very large `cpp` files

cbjeukendrup:
Or just reduce the number of files per unit, can't we do that?

igorkorsukov:
> Or just reduce the number of files per unit, can't we do that?

Now the number of files in the unit is a global setting, I don't want to add another setting to setup the module, and there are so many settings :)

I don't think it is necessary to globally reduce the number of files in the unit, in other modules, there are a few lines in each `cpp` file. 
There are a lot of lines of code in the engraving modules in most `cpp` files, they themselves are like units :), so I think disabling the unit build will have little effect for engraving

Jojo-Schmitz:
See https://github.com/musescore/MuseScore/pull/11547#issuecomment-1119503836

PR #11547, opened 2022-05-06
Disabled unity build for engraving module

igorkorsukov:
To prevent problem https://github.com/musescore/MuseScore/pull/11545#issuecomment-1119458761


Jojo-Schmitz:
Instead using `set(CMAKE_UNITY_BUILD_BATCH_SIZE 6)` in src/engraving/CMakeLists.txt seems to fix this too (as far as I can tell 8 is the default there).
And maybe we need/want this for MinGW only, as other build environments don't seem to have that issue?

```
$ git diff
diff --git a/src/engraving/CMakeLists.txt b/src/engraving/CMakeLists.txt
index db5ef3cc02..4da3354ce7 100644
--- a/src/engraving/CMakeLists.txt
+++ b/src/engraving/CMakeLists.txt
@@ -218,6 +218,9 @@ set(MODULE_INCLUDE
 set(MODULE_DEF ${ENGRAVING_INFRASTRUCTURE_DEF})
 set(MODULE_LINK ${ENGRAVING_INFRASTRUCTURE_LINK} qzip)

+if(MINGW)
+    set(CMAKE_UNITY_BUILD_BATCH_SIZE 6)
+endif()
 include(SetupModule)

 set_property(TARGET ${MODULE} APPEND PROPERTY AUTOMOC_MACRO_NAMES "BEGIN_QT_REGISTERED_ENUM")
```

PR #11551, opened 2022-05-06
[MU4] Backend logging fix

Eism:


src/framework/global/log.h
```diff
@@ -30,6 +30,10 @@
 #endif
 #include "thirdparty/haw_logger/logger/log_base.h"
 
+#define ASSERT_FAILED_X(msg) \
```
igorkorsukov, line 33: `FAILED` not need here

igorkorsukov (approved):

PR #11556, opened 2022-05-06
Update and clean up the app icons and file icons

cbjeukendrup:
The new icons are created by @Tantacrul.

I also did some cleanup: renamed "share" to "resources", and moved all icons to resources/icons. Now, we have them all in one place. I removed the MusicXML icons, because they were unofficial and not super beautiful. There are currently no SVG versions available of the new icon.

Note: it will be difficult to see the results of this PR. 
On macOS, it is advisable to move all other versions of the app (even MU3) to the trash, or put them on external storage and then disconnect that storage; anyway, a location where macOS can't find them, so that the build from this PR is the only one.
Then, do the following:
- open Finder
- press Cmd + Shift + G
- type `/Library/Caches` and press Return;
- delete the folder called com.apple.iconservices.store (enter your TouchId or password if required)
- press Cmd + Shift + G
- type `/private/var/folders` and press Return;
- press Cmd + F to search, and search for `com.apple.iconservices`; make sure that you have selected `folders` as the search scope, not "This Mac"
- delete the folder called "com.apple.iconservices"
- while pressing the Option (alt) key, right-click the Finder icon and click "Restart" (alternatively, restart your Mac)
- Now you should finally see the new icons in Finder. 

On Windows, you will need to re-install MuseScore using a MSI installer. When you just use the .7z file from this PR, the new icons will not be associated with MuseScore files. I will create such a MSI installer soon.

On Linux, I have not been able to test this; I'm not sure what is necessary on various versions of Linux to associate the new icons with MuseScore files. Basically, if the old version worked, the new version _should_ work too.


Marr11317:
image

What about "Uncompressed" or "raw" instead of "MSCX"? mscx is rather technical.

 @Tantacrul

cbjeukendrup:
@Marr11317 At least on macOS, it is a common pattern to show the extension in the file icon. A problem with adding some real-language word to the icon, is that it can't be localised, so everyone will see it in English.

Tantacrul:
For the main filetype, I didn't use the extension. In this case, I think it makes sense because it's not for general users.

shoogle:
> On Linux, I have not been able to test this; I'm not sure what is necessary on various versions of Linux to associate the new icons with MuseScore files. Basically, if the old version worked, the new version should work too

For MuseScore's AppImage, simply run:

```Bash
./MuseScore*.AppImage install
```

Note: AppImages in general don't have an `install` option. We implemented the behaviour [here](https://github.com/musescore/MuseScore/blob/aae8a90f065b606e347bc931863209de72d63aa4/build/Linux%2BBSD/portable/portable-utils.in#L11-L12).

---

On Linux the icon files get installed to a shared icon folder. Usually:

```Bash
PREFIX/share/icons/THEME/SIZE/apps       # app launcher icon (e.g. MuseScore, Audacity)
PREFIX/share/icons/THEME/SIZE/mimetypes  # file icons (e.g. MSCZ, AUP)
```

The default THEME is `hicolor`. SIZE can be `16x16`,  `24x24`, ...,   `512x512` for PNGs, or `scalable` for SVGs.

PREFIX is usually one of:

```Bash
~/.local    # per-user install (installed just for current user).
/usr/local  # system-wide install (installed for all users on this machine).
/usr        # system-wide install by the package manager (apt, dnf, yum, etc.).
```

The shared icon folders contain all icons for all applications. Each application specifies the icon it wants to use from this folder:

- The app launcher icon is specified in the application's [.desktop file](https://github.com/musescore/MuseScore/blob/aae8a90f065b606e347bc931863209de72d63aa4/build/Linux%2BBSD/mscore.desktop.in#L11) (see [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys)).
- The file type icons are specified in the application's [.xml mime-info package file](https://github.com/musescore/MuseScore/blob/aae8a90f065b606e347bc931863209de72d63aa4/build/Linux%2BBSD/musescore.xml.in#L10) (see [Shared MIME-info Database Specification](https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html#idm46292897757504)).

With the .desktop, mime-info, and icon files in place, the package manager (or in our case the AppImage install scripts) call some terminal commands to [update the system caches](https://github.com/musescore/MuseScore/blob/aae8a90f065b606e347bc931863209de72d63aa4/build/Linux%2BBSD/portable/portable-utils.in#L376-L378).

---

Note: only the icon name is specified, not the directory or file extension. So if the desktop file specifies `mscore` as the icon, it might get resolved to the full path:

```
~/.local/share/icons/hicolor/64x64/apps/mscore.png
```

The exact file used will depend on the user's system theme and screen resolution. The system chooses the best one that's available.

cbjeukendrup:
@boryskuzmenko When testing this, you don't need to test on macOS, since I already confirmed that it works there, and I can imagine it will not be convenient if you have to remove all other builds from your computer. 

@igorkorsukov In order to test this on Windows, we will need a MSI installer. But when I try to create one from my branch, it fails, because in my repo the necessary secrets can't be accessed. I see two things we can do:
- merge this PR to master, and test afterwards with an installer from the master branch
- create a new branch in the musescore repo with the changes from this PR, and test with an installer from that branch

Which one has your preference? Or are there other options?

Tantacrul:
@cbjeukendrup - I can grab someone who doesn't have MuseScore installed from Twitter who is on Windows / Linux and get them to install this and tell us?

cbjeukendrup:
@Tantacrul I believe that on Windows and Linux, it is not strictly necessary to have no other versions installed; the installers will overwrite the old installation. (On macOS, there is no installer, so there it fully depends on the OS picking up the correct version.)
But to try it out on Windows, we will need such an installer, and I've not yet succeeded in creating one.

Tantacrul:
Just posted this on Twitter now:
https://twitter.com/Tantacrul/status/1525538795432693760

cbjeukendrup:
@Tantacrul It won't work with that Windows version, because it is not a MSI installer. This is just a zip file, and there is nothing that registers the file icons with the system: that is one of the main tasks of the MSI installer. Unfortunately, I don't seem to be able to create such an installer myself, and therefore I asked @igorkorsukov for advice.

Tantacrul:
Ah, I didn't quite realise when you mentioned this before. OK, I've removed the request.

shoogle:
I tested this on Linux and it's working. There are a few minor issues if the user has multiple versions of MuseScore installed but I think it is easiest if I fix those in a separate PR after this one is merged.

@cbjeukendrup, I'm not sure why you renamed the "share" folder. I guess "resources" is more descriptive but changing the folder name has created a diff for all the files inside. Perhaps next time you rename a top-level folder you could do so in a separate commit (ideally a separate PR) to make review easier? Best not do it too often though because it interferes with the history and `git blame`.

cbjeukendrup:
@shoogle Indeed, I renamed it because "resources" made more sense to me. But yes, it's quite an invasive change... I could still move it to a separate commit. Alternatively, we can just stick with the name "share". Let me know what you prefer.

shoogle:
@cbjeukendrup, I think stick with the name "share" for now just because it will make this PR easier to review now and easier to understand in the future. Feel free to keep the changes relating to the rename on a separate branch and submit a separate PR for it after this one is merged so you don't have to keep rebasing all the time.

cbjeukendrup:
@shoogle Done!

I also confirmed again that it works well on macOS (_if_ macOS chooses the correct version of MuseScore, which we can't control, but which will be okay in the released version).

So since it works on macOS and (basically) Linux too, would this PR be ready for merge? Then we can check on Windows later (it should be much easier to create a test installer from the master branch).

shoogle (approved):
@cbjeukendrup, thanks! I'm happy with the code. We've tested it works on Linux and macOS.

I think it should be safe to merge now and test on Windows later, but I'll give @igorkorsukov an opportunity to have a say first.

igorkorsukov:
> Which one has your preference? Or are there other options?

I think, better create a new branch in the musescore repo, do testing, and delete it after task finished

igorkorsukov (approved):

cbjeukendrup:
@igorkorsukov I created a branch in the MuseScore repo and am trying to create an installer from it, but I discovered a very stupid problem.

By now, "build numbers" have reached values larger than 2147483647, the maximum for an Int32. But one of the tools behind WiX tries to parse version numbers using https://docs.microsoft.com/en-us/dotnet/api/system.version.tryparse, which tries to parse each component of the version number into an Int32. 
See https://github.com/musescore/MuseScore/runs/6473426348?check_suite_focus=true#step:9:8079 and https://github.com/musescore/MuseScore/blob/55346de77e0a6cde007d30e851646c591629cf6d/build/packaging/WIX.template.in#L20

That obviously fails and makes that WiX tool crash, which makes the whole packaging fail. So we can't create _any_ installers for Windows until this gets resolved. 

So what can we do? Remove the build number from the version number? Subtract a random number from the build number, and increase that number when build numbers grow _even_ bigger? Use the build number modulo 2147483648?

shoogle:
> Use the build number modulo 2147    483648

That would mean the final digits would bear no resemblance to the original build number, but we *could* use the number modulo 10^9 as that would keep the final 9 digits the same. Or even modulo 10^6 or smaller if we don't want that many digits. Fewer digits would lead to it wrapping more often though, which could cause problems identifying the latest version between releases.

Marr11317:
> By now, "build numbers" have reached values larger than 2147483647, the maximum for an Int32

That's actually funny. It can't mean that Musescore has been built that many times though... Or has it? When is the build number incremented?

cbjeukendrup:
The build numbers are actually the GitHub run ID. I believe that's incremented for each action run for the whole repo. So that increments quite quickly! But it is useful because it's an incremental number that is generated for us.

cbjeukendrup:
After some struggles with creating an installer, I was finally able to check that it works on Windows! You simply run the installer, and then make sure that the installed version is selected as the default app for mscz and mscx files, and then it instantly works. 
Schermafbeelding 2022-05-23 om 00 11 01
If anyone else wants to try it, use the installer from https://github.com/musescore/MuseScore/actions/runs/2367810813. Do note that this is just some development build, but it will be installed as a release version (!), in order to make the registration of the icons happen. So, once you're done with testing, you may want to uninstall this version again (by re-running the installer and selecting the "Remove" option on the second page), in order to avoid annoyances and conflicts later on.

Tantacrul:
Hurrah!

cbjeukendrup:
@igorkorsukov Could you please check if you're happy with the changes that I had to make for the Windows installer? If you are happy, then I think this PR is ready to be merged.

igorkorsukov:
> 

eh, I missed it, and now I'm faced with the same thing when building 3.6.2, I think you made the right decision - to remove the build number

PR #11558, opened 2022-05-07
Fix #11502: Change instrument always applied

### LINKED ISSUE #11502: [MU4 Issue] Cancelling "change instrument" dialog still effects instrument change
**Describe the bug**
Entering a "change instrument" element from the **Text** palette triggers a dialog that allows the user to select which instrument/sound they want for their mid-score instrument change. 

The issue is that, if the user clicks **Cancel** without making any instrument selection, an instrument change is still added to the score, and a (rather random) instrument is assigned for playback. 

Video to demonstrate: 

https://user-images.githubusercontent.com/86290556/166232893-0a51304d-1098-49a7-8864-37410738c2f6.mov

**Expected behavior**
I would expect that pressing **Cancel** in this dialog would cancel the process of adding a "change instrument". The label in the score would therefore be removed and there would be no change to the playback sound of the stave in question.  

**Platform information**
 - MacOS 12.3.1 (M1)

HemantAntony:
Resolves: #11502

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -1316,7 +1316,11 @@ bool NotationInteraction::drop(const PointF& pos, Qt::KeyboardModifiers modifier
 
         EngravingItem* dropElement = el->drop(m_dropData.ed);
         if (dropElement && dropElement->isInstrumentChange()) {
-            selectInstrument(toInstrumentChange(dropElement));
+            if (!selectInstrument(toInstrumentChange(dropElement))) {
+                rollback();
+                accepted = true;
+                break;
+            }
         }
         score()->addRefresh(el->canvasBoundingRect());
         if (dropElement) {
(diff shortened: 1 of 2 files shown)


RomanPudashkin (approved):

bkunda:
Tested on Mac and Windows.

Only minor issue is that there is an unnecessary animation of the bounding box surrounding the "Change instr." element when the "Cancel" button is clicked (thereby closing the dialog): 

https://user-images.githubusercontent.com/86290556/167444439-58197cf3-c8a4-425c-96a9-3a6ef788a683.mov

This only seems to affect MacOS (couldn't reproduce on Windows). 

Also I was unable to verify that the process of actually changing instruments still works because there were no instruments in the instruments dialogs (although I believe this is a bigger issue concerning the latest master build). 

So if this animation could please be removed, then it should be ready to go. 

Thanks!

HemantAntony:
I sadly don't even have that animation for any palette cells in the first place. I think its due to me being on Windows (I wonder if that should be an issue by itself)

So I don't think I can solve this bug

bkunda:
> I sadly don't even have that animation for any palette cells in the first place. I think its due to me being on Windows (I wonder if that should be an issue by itself)
> 
> So I don't think I can solve this bug. Any ideas what to do, Devs?

If it can be fixed, then great. Otherwise I don't think it should stand in the way of this being merged if it can't. It's more important that the interaction with the dialog is corrected (which you've done). Minor UI tweaks can always be tweaked later.

cbjeukendrup:
This is system behaviour on macOS, and indicates that the drop event was rejected by the app.
However, in this specific case I agree the animation does not make much sense. I would suggest to "accept" the drop event in this case.

src/notation/internal/notationinteraction.cpp
```diff
@@ -1316,7 +1316,10 @@ bool NotationInteraction::drop(const PointF& pos, Qt::KeyboardModifiers modifier
 
         EngravingItem* dropElement = el->drop(m_dropData.ed);
         if (dropElement && dropElement->isInstrumentChange()) {
-            selectInstrument(toInstrumentChange(dropElement));
+            if (!selectInstrument(toInstrumentChange(dropElement))) {
+                rollback();
```
cbjeukendrup, line 1320: > "accept" the drop event

... for example, by adding `accepted = true;` here

HemantAntony:
@bkunda It should work now

PR #11560, opened 2022-05-07
[MU4] Use `std::gcd()`

Jojo-Schmitz:
Use C++17's `std::gcd()` rather than our own implementation.


src/engraving/libmscore/edit.cpp
```diff
@@ -3375,6 +3375,10 @@ void Score::cmdCreateTuplet(ChordRest* ocr, Tuplet* tuplet)
     track_idx_t track = ocr->track();
     Measure* measure = ocr->measure();
     Fraction tick = ocr->tick();
+    Fraction an = (tuplet->ticks() * tuplet->ratio()) / tuplet->baseLen().fraction();
+    if (!an.denominator()) {
+        return;
```
Jojo-Schmitz, line 3380: Needs to be before the `undoRemoveElement(ocr);` a few lines down.

igorkorsukov (approved):

Issue #11145, opened 2022-04-11
[MU4 Issue] Unable to close Master Palette window with shortcut Shift+F9 (when it has focus)


DmitryArefiev:
**Describe the bug**
Unable to close Master Palette window with shortcut Shift+F9 (when it has focus)

**To Reproduce**
1. Run MU
2. Open a score
3. Open Master Palette window with shortcut Shift+F9
4. Press Shift+F9 again to close the window - **it doesn't work**

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**
Escape shortcut works fine to close the window
Shift+F9 shortcut works for closing Master Palette window when it doesn't have focus

micpap25:
Is this issue available? It seems like a relatively straightforward fix.

DmitryArefiev:
Yeah, you can take it. Thanks!

Tantacrul:
I tested this on MacOS just now and the shortcut does close the master palette again. Can you check @DmitryArefiev and reopen if you find it's not working?

DmitryArefiev:
> I tested this on MacOS just now and the shortcut does close the master palette again. Can you check @DmitryArefiev and reopen if you still see an issue?

@Tantacrul Rechecked in master. In some cases it's better on Mac. But **still occurs**. I've updated the steps to reproduce on all platforms.

shoogle:
On macOS Shift+F9 works to close the Master Palette regardless of focus. On Linux it only works if MuseScore's main window has focus (i.e. it doesn't work if the Master Palette has focus). I've not tested on Windows yet.

PR #11576, opened 2022-05-09
Accessibility for Special Characters dialog

shoogle:
Fixes the F2 shortcut for the Insert Special Characters dialog and ensures the dialog is accessible to screen readers.


shoogle:
Tested on Windows and macOS. The F2 shortcut works on both platforms. Speech works with Narrator and NVDA. Speech for VoiceOver will be added when we port the dialog to QML: there's no easy way to do it before then unfortunately.

Eism:
@shoogle 
Crash occurs when pressing F2 without text editing
(you should add check in PaletteActionsController)

shoogle:
@Eism, thanks for the hint! I fixed that and added some more improvements. @Tantacrul has approved use of F2 as an alternative to Alt+Shift+E to begin text editing, and Shift+F2 to open the Special Characters dialog. F2 is the standard editing shortcut used in spreadsheet applications (to edit cell data) and File Explorer (to rename a file).

RobFog:
Rebase needed

shoogle:
@RobFog, I rebased the other day to account for the QString -> String change. It merges cleanly now.

Eism (approved):

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 

@shoogle 
1. F2 doesn't work for text editing for me (only Alt+Shift+E works)
2. As I see in the commits, Shift+F2 also should close the dialog but that doesn't work for me (added to #11145 for now)

shoogle:
@DmitryArefiev
> F2 doesn't work for text editing for me (only Alt+Shift+E works)

Strange! F2 and Alt+Shift+E are both working for me on all platforms. If you're using a laptop I guess you already tried pressing Fn with F2? Maybe Help > Revert to Factory Settings and relaunching MuseScore will help? Or maybe there is another program stealing the F2 keypress event (e.g. a screen reader or virtual machine app)?

> Shift+F2 also should close the dialog but that doesn't work for me

It works when the main window has focus (i.e. when the Insert Special Characters dialog doesn't have focus). When the dialog has focus you are able to close it with Esc. I can try to enable Shift+F2 to close the dialog even when it has focus, but it wouldn't work if you manually change the Shift+F2 shortcut to a different key combination in the Preferences (at least not without an overhaul of the shortcuts system).

Eism:
> When the dialog has focus you are able to close it with Esc. I could try to enable Shift+F2 to close the dialog even when it has focus, but I think that's a separate issue really.

I think Esc is enough

src/engraving/libmscore/textedit.cpp
```diff
@@ -249,9 +249,14 @@ bool TextBase::isEditAllowed(EditData& ed) const
         return false;
     }
 
-    bool ctrlPressed  = ed.modifiers & Qt::ControlModifier;
     bool shiftPressed = ed.modifiers & Qt::ShiftModifier;
 
+    if (shiftPressed && ed.key == Qt::Key_F2) {
+        return false; // Toggle Special Characters dialog
```
Eism, line 255: I think we can find a better solution
see https://github.com/musescore/MuseScore/pull/11923#issuecomment-1166914602

Eism: Okey, let's leave it as it is for now. Many changes will be needed

DmitryArefiev:
> Strange! F2 and Alt+Shift+E are both working for me on all platforms. If you're using a laptop I guess you already tried pressing Fn with F2? Maybe Help > Revert to Factory Settings and relaunching MuseScore will help? Or maybe there is another program stealing the F2 keypress event (e.g. a screen reader or virtual machine app)?

@shoogle Yes, after Revert to Factory Settings applying F2 works fine. Thanks!

> It works when the main window has focus (i.e. when the Insert Special Characters dialog doesn't have focus). When the dialog has focus you are able to close it with Esc. I could try to enable Shift+F2 to close the dialog even when it has focus, but I think that's a separate issue really. It should be easy enough to handle the Shift+F2 keypress when the dialog has focus but it would require more effort to get it working if the user has reassigned the shortcut to a different key combination.

Yes, let's track it in #11145 and it's not major because Esc works fine

shoogle:
Ok cool! I guess this is ready to merge then?

PR #11581, opened 2022-05-09
Custom keysig - count symbols width

sammik:
Custom Key Signature - default accidental positions by symbol width, not fixed xstep

Accidentals have various widths (natural is thin, triple flat is fat). If default xstep were fixed, unwanted gaps, or overlays are produced.
Test file - accidentals on default positions (`xAlt = 0`)
before (xStep = 1.4 sp):
Width-before

after (xstep depends od symbol width):
Width-after

other minor tweaks:
- add more transposable symbols (`Sharp Sharp`, `Natural Sharp`, `Natural Flat`)
- replaced `QString` by `std::string` on some places
- replaced `score()->spatium() / mag()` by `spatium()` (which already has mag calculation included)

### DIFF
--- a/src/engraving/libmscore/key.cpp
+++ b/src/engraving/libmscore/key.cpp
@@ -46,6 +46,21 @@ const SymId accTable[] = {
     SymId::accidentalDoubleSharp,
     SymId::accidentalTripleSharp,
     SymId::noSym,
+    //  natural sharp
+    SymId::accidentalNatural,
+    SymId::accidentalNaturalSharp,
+    SymId::accidentalSharp,
+    SymId::noSym,
+    //  natural flat
+    SymId::accidentalDoubleFlat,
+    SymId::accidentalNaturalFlat,
+    SymId::accidentalNatural,
+    SymId::noSym,
+    //  natural sharp sharp
+    SymId::accidentalSharp,
+    SymId::accidentalSharpSharp,
+    SymId::accidentalTripleSharp,
+    SymId::noSym,
     // Gould quarter tone
     //  arrow down
     SymId::accidentalFiveQuarterTonesFlatArrowDown,
--- a/src/engraving/libmscore/key.h
+++ b/src/engraving/libmscore/key.h
@@ -105,7 +105,6 @@ class KeySigEvent
     bool _forInstrumentChange{ false };
     std::vector<CustDef> _customKeyDefs;
     std::vector<KeySym> _keySymbols;
-    double _xstep       { 1.4 };
 
     void enforceLimits();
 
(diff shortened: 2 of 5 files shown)


RomanPudashkin:
Could you please add some unit-tests (only for changes in the engraving module) that will test your fix?

sammik:
> Could you please add some unit-tests (only for changes in the engraving module) that will test your fix?

@RomanPudashkin sorry, I need some guidance here. Or @cbjeukendrup ?
Any hint, or doc, or wiki please?
Thank You!

RomanPudashkin:
Sorry, I meant vtests. You already have an experience of adding new vtests :)
https://github.com/musescore/MuseScore/pull/10519

Or you can modify some existing vtests and add your case to them

sammik:
> Sorry, I meant vtests. You already have an experience of adding new vtests :) #10519
> 
> Or you can modify some existing vtests and add your case to them

@RomanPudashkin One question / trable with this new test.
I need to use new file format, as introduced in #11280 
but tests are written in MS3 (or older) format and new MS4 mscx ignores page layout (this info is saved in separate file)

So I can:
1) ignore page format setting, which means test png will be unnecessarily big.
or
2) save this .mss file in vtests folder too (I am not sure, if it would be included by test script)
or
3) I write header as if it were MS3 `<programVersion>3.6.2</programVersion>` but in fact will use new format `def` instead of `pos` (it is not part of migration scripts, so it shouldnt be a problem in master)

sammik:
Updated test file to show accidentals in default xposition (`xAlt=0`).

vtest fail is IMHO good, see
before
custom-keysig-1-1 ref

after
custom-keysig-1-1

@oktophonie I think, now You can look at it.

Thanks

oktophonie:
How does one even create a custom key signature? I can do one via the palettes but then it has the spacing that I specify (according to where I drag the symbols to), not a 'default' spacing.

sammik:
> How does one even create a custom key signature? I can do one via the palettes but then it has the spacing that I specify (according to where I drag the symbols to), not a 'default' spacing.

@oktophonie Yes, that is, how custom key editor (creator) works right now. But if I understood, it needs to be completly rewritten anyway (now it is widget), so I hope, new editor will respect this (and hopefully much more).

You can set default values (or create custom keysig) by altering mscx file at the moment.

I attached test file with accidentals in default positions.
[CusKeySig.zip](https://github.com/musescore/MuseScore/files/8780267/CusKeySig.zip)

sammik:
Eventualy I can adapt #10454 to shift xposition by multiples of default position.

sammik:
I updated editor, so now accidentals takes default positions (unless Control ispressed).
@oktophonie are You happy with it now?
Thank You!

https://user-images.githubusercontent.com/1646034/170921432-126fc566-28b4-4fe6-8488-b9107c6fe658.mp4

RomanPudashkin (approved):

Issue #11358, opened 2022-04-25
[MU4 Issue] Parts - Scrollbar Prevents Closing Tabs


lengthwave:
**Describe the bug**
When more parts are open than the width of the window, a floating scrollbar appears on top of tabs. This scrollbar prevents ability to close tabs.

**To Reproduce**
1. Open or create score with many parts
2. Open all parts
3. Try to close part tabs

**Expected behavior**
Ability to delete tabs.

**Screenshots**
<img width="1188" alt="Screen Shot 2022-04-25 at 5 57 14 PM" src="https://user-images.githubusercontent.com/30370609/165118202-f4e127c3-efa5-41b9-bb73-cfc9936d9e12.png">

**Platform information**
OS: macOS 12.2, Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2214419458
revision: github-musescore-musescore-df96a36

lengthwave:
@Tantacrul - FYI

Tantacrul:
Here is @jessjwilliamson design to fix this issue
https://www.dropbox.com/s/bjs56hs8nx3v8cn/Scrolling%20Tabs.pdf?dl=0

Issue #11243, opened 2022-04-19
[MU4 Issue] 2 scrollbars in font face menu, Preferences


jessjwilliamson:
One scroll bar is working but there is another one stuck on the top of the menu, also in a different size!

https://user-images.githubusercontent.com/71656189/163978494-f42b51ba-67cb-4329-9254-4b304c338dc2.mp4

PR #11582, opened 2022-05-09
Replace the scrollbar in the notation tabs by arrow buttons, and give the tabs a context menu

cbjeukendrup:
Resolves: #11243 (it was already fixed in dropdowns, but this PR fixes the underlying problem so that it will be fixed everywhere)
Resolves: #11358


jessjwilliamson:
Excellent work! :D 
In the pdf spec, I also mentioned about access to scroll through the tabs with a mouse wheel or trackpad. Can this be enabled? It doesn't seem to work right now!

cbjeukendrup:
Scrolling horizontally with mouse or trackpad seems to work as before. Or did you mean that scrolling vertically should cause the tabs to move horizontally? (like what happens currently when pressing the shift key)

jessjwilliamson:
> Scrolling horizontally with mouse or trackpad seems to work as before. Or did you mean that scrolling vertically should cause the tabs to move horizontally? (like what happens currently when pressing the shift key)

So shift and mouse wheel don't work anyway to scroll horizontally.
I meant that just normal mouse wheel (vertically) can still scroll horizontally through the tabs.

src/notation/qml/MuseScore/NotationScene/internal/NotationSwitchScrollArrowButton.qml
```diff
@@ -0,0 +1,67 @@
...
+
+FlatButton {
+    id: root
+
+    width: 22
+    height: parent.height
+
+    opacity: 1.0
+
+    backgroundItem: Rectangle {
```
Eism, line 34: Please explain why you redefine the background item?

src/notation/qml/MuseScore/NotationScene/internal/NotationSwitchButton.qml
```diff
@@ -96,6 +100,52 @@ FlatRadioButton {
...
+            acceptedButtons: Qt.RightButton
+
+            onClicked: function(mouse) {
+                contextMenuLoader.show(Qt.point(mouse.x, mouse.y))
+            }
+
+            ContextMenuLoader {
+                id: contextMenuLoader
+
+                items: {
```
Eism, line 117: Please move this data to NotationSwitchListModel

src/notation/qml/MuseScore/NotationScene/internal/NotationSwitchScrollArrowButton.qml
cbjeukendrup: I did it because we don't want the rounded corners in this case, _and_ we want the background a different color, _and_ we want the normal state opacity to be 1.0 instead of ui.theme.buttonOpacityNormal. 
Of course, we can add more properties to FlatButton to control the corner radius and the base state opacity, but I decided not to do that because FlatButton already has so many properties and I don't wanted to complicate it even further.

src/notation/qml/MuseScore/NotationScene/internal/NotationSwitchPanel.qml
```diff
@@ -61,57 +62,113 @@ Rectangle {
...
 
-            text: model.title
-            needSave: model.needSave
+                readonly property bool canCloseOtherTabs: notationsView.count > 2
+                                                          || (notationsView.count === 2 && index === 0)
+                readonly property bool canCloseAllTabs: notationsView.count > 1
 
-            ButtonGroup.group: notationsView.radioButtonGroup
-            checked: index === notationsView.currentIndex
+                contextMenuItems: notationSwitchModel.contextMenuItems(canCloseOtherTabs, canCloseAllTabs)
```
cbjeukendrup, line 114: I'm keeping the calculation of `canCloseOtherTabs` and `canCloseAllTabs` in QML, so that we can make use of QML's property binding, and `notationSwitchModel.contextMenuItems` will be called again automatically when those properties change.

cbjeukendrup:
@jessjwilliamson About the mouse/trackpad scrolling issue, I tried on both macOS and Windows, but I don't see a difference between the behaviour from this PR and from the master branch. I can just scroll left and right with my trackpad and with my mouse, and the tabs will move. What am I missing?

jessjwilliamson:
> @jessjwilliamson About the mouse/trackpad scrolling issue, I tried on both macOS and Windows, but I don't see a difference between the behaviour from this PR and from the master branch. I can just scroll left and right with my trackpad and with my mouse, and the tabs will move. What am I missing?

That's interesting. I've just both OS each with their own mice. Mac it works with shift and scroll with mouse. On my Windows this doesn't work. I'm not sure if that's my Windows PC or my mouse issue.. weird

cbjeukendrup:
@jessjwilliamson I found out what's going on, see https://github.com/musescore/MuseScore/issues/11625#issuecomment-1126050174

src/notation/qml/MuseScore/NotationScene/internal/NotationSwitchPanel.qml
Eism: It is bad practice to rely on such dependence.
Let's move the logic to the model

For example 
add contextMenuRequested signal for NotationSwitchButton
on handler get the context menu items from model like notationSwitchModel.contextMenuItems(index)
(canCloseOtherTabs and canCloseAllTabs should be in the model)

OR (and maybe better than previous version)
add contextMenuItems as a role for item in the model

Eism (approved):

src/notation/view/notationswitchlistmodel.h
```diff
@@ -49,6 +49,11 @@ class NotationSwitchListModel : public QAbstractListModel, public async::Asyncab
     Q_INVOKABLE void load();
     Q_INVOKABLE void setCurrentNotation(int index);
     Q_INVOKABLE void closeNotation(int index);
+    Q_INVOKABLE void closeOtherNotations(int index);
```
RomanPudashkin, line 52: What does this function do?

cbjeukendrup: It closes all notations except the specified one and the master notation. (The name of this method is based of the name of this functionality in the UI.)

DmitryArefiev:
@cbjeukendrup I think it would be nice to make arrow buttons work in auto repeat mode when holding mouse button and not to click each time to scroll the parts bar

https://user-images.githubusercontent.com/90187801/171414934-79e48cd0-23d8-41db-9cb9-789e6947dd54.mp4

If it's not a hard I can create a issue/task for that

cbjeukendrup:
@DmitryArefiev Nice suggestion! That should be possible. Let's create a task for it :)

DmitryArefiev:
> @DmitryArefiev Nice suggestion! That should be possible. Let's create a task for it :)

@cbjeukendrup Cool! Created #11864 and assigned to you

PR #11584, opened 2022-05-09
Update `dr_libs` dependency

cbjeukendrup:
dr_flac.h v0.12.20 -> v0.12.38
dr_mp3.h v0.6.17 -> v0.6.17
dr_wav.h v0.12.12 -> v0.13.6

This should be a step towards compiling in ARM64 on Windows with Qt 6.3.


Jojo-Schmitz:
Along with #10206

vpereverzev:
there is no need in this update, sooner or later we'll get rid of this dependency, since we're working with lame/flac/opus libraries directly

cbjeukendrup:
Okay! Then I'll close this.

Issue #11577, opened 2022-05-09
[MU4 Issue] Fix scrollbar issue on Mixer


jessjwilliamson:
**Describe the bug**
An issue mentioned by @Tantacrul that a solution needs to be solved with the Mixer's scrollbar which is currently causing some visual issues on the instrument names.

![image](https://user-images.githubusercontent.com/71656189/167431705-c7eccd83-d9e9-41e8-a5de-33421e4ecf62.png)

**Screenshots**
Changing the scrollbar in Mixer - which will now have its own padding, rather than overlapping the instrument names.

Only when the instruments do exceed the width in the Mixer, the scrollbar (with its padding) will appear permanently. 
If instruments in mixer does not exceed width, no scrollbar will appear (including the padding - this does mean the instrument channels and labels will be adjusted vertically to accommodate the whole scrollbar space)

![image](https://user-images.githubusercontent.com/71656189/167428444-17713734-e210-483a-b9a8-1f60a7821e8d.png)

**Additional context**
Assets for this can be found in the 'MS4 Main Interface Panels > MIXER' in the Scrolling horizontally section.

Issue #11367, opened 2022-04-25
[MU4 Issue] Mixer - Scrollbar masks instrument label


lengthwave:
**Describe the bug**
When there are a large number of instruments, the scrollbar on the bottom of the mixer masks/obscures instrument names.

**To Reproduce**
1. Open/create score with more instruments than width of mixer
2. Go to Mixer
3. Scroll left/right
4. See issue

**Expected behavior**
Scrollbar should be below mixer, not floating on top of labels. 

**Screenshots**

<img width="1509" alt="Screen Shot 2022-04-26 at 12 06 29 AM" src="https://user-images.githubusercontent.com/30370609/165175550-fa2bf7f2-ea53-40cf-b309-5b1538bbfa2a.png">

OS: macOS 12.2
Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2214419458
revision: github-musescore-musescore-df96a36

lengthwave:
@Tantacrul - FYI

Tantacrul:
@jessjwilliamson - I guess we are probably going to have to sort that out. :)

PR #11585, opened 2022-05-10
Fix the horizontal scrollbar in the Mixer

cbjeukendrup:
It was overlapping with the content.

Resolves: #11367
Resolves: #11577


jessjwilliamson:
Absolutely amazing!! Thank you!
There is one bug though. It does not work when it's undocked :) 

https://user-images.githubusercontent.com/71656189/167605791-8543e88d-f4db-4302-82a3-d4577810eda9.mp4

cbjeukendrup:
That seems not reproducible on macOS. I think it is related to the dock widgets system on Windows, and not to this PR, since it occurs in other dock panels too (for example, the piano keyboard).

jessjwilliamson:
> That seems not reproducible on macOS. I think it is related to the dock widgets system on Windows, and not to this PR, since it occurs in other dock panels too (for example, the piano keyboard).

Okay, this is just a Windows problem then

RomanPudashkin (approved):

Issue #11522, opened 2022-05-04
[MU4 Issue] Assigning a shortcut to "go to top staff" doesn't appear to work


wizofaus:
**Describe the bug** (Edited by Martin)

1. When the shortcut is triggered it correctly selects the topmost staff but the view does not scroll up so we can see that selection (it does this in MS3). Because the scrolling does not work, the user cannot tell whether the shortcut has worked correctly.
2. It should be renamed to 'Select the top staff'

Tantacrul:
Can you possibly convert this to a set of expected behaviour. It's hard to understand.

Title: It's not working
Description: Well, it _is_ working, etc.

quinnouyang:
@Tantacrul I think they're specifying three things:
- It correctly* selects the left-most element of the selected measure (if none selected, then of the entire score in the topmost staff). E.g. Select a note/rest on beat 3 of a 4/4 measure; the shortcut selects the element on beat 1. Or select nothing; the shortcut selects the element on beat 1 on the first staff.
- *Expected behavior is not clear from the title, "Go to top staff." Is "go to the first beat in selected measure" correct?
- When the shortcut selects an element is out of view, the view does not scroll to it. This is desired for "Page: top of first", which brings you to the top of the first page _without selecting the top + left-most element_. I presume we'd want it here too.

MarcSabatella:
No, I don’t thinks that it at all.  “Go to top staff” is a command that literally moves the cursor/selection to the top staff of your score.

The main bug here is that if you try to assign a shortcut to this command, it doesn’t work.  The dialog allows you to add the shortcut and hit OK.  But if you try to use the shortcut, nothing happens.

The strange part of the the bug is that the command shows up in the shortcut list at all.  That’s strange because it’s the only command with no default shortcut to appear in the list.  All other commands with no default shortcut are missing from the list - a separate bug submitted a while back and referenced quite a few times recently.

For some reason, the “go to top staff” command seems immune to the bug that prevent shortcutless-commands from appearing in the list, but the command also seems to not work if you do assign it a shortcut.

That’s what I read in the original report, and it’s easily reproducible as described.

wizofaus:
My description is as stands, but at least after restarting MuseScore, the shortcut key *does* work (I have observed occasions where you need to restart MuseScore for newly-assigned shortcut keys to start working), it's that the behaviour isn't really what I'd expect (particularly wrt not scrolling the selected note into view).

[...]

PR #11586, opened 2022-05-10
fix: 11060; Added quite many shortcuts from MS3

VanSHOE:
Resolves: #11060 

Follow up of https://github.com/musescore/MuseScore/pull/11144.
This simply adds quite many shortcuts from MS3.
I went through the list of MS3 shorcuts and tried to find as many actions I could find there that were also there in the code of MS4 but missing from the shortcuts list and added them in XML

Several parts could require changes, especially the context
All the actions had notation-opened context, so I added the same in XML as well, but please do verify if that is correct
Do note that not all MS3 shortcuts have been added since action codes for most could not be found by me. but a significant chunk definitely has been added

While verifying the changes, lines may look quite a lot, but the three files essentially have the same changes. (so the changes are essentially tripled)


src/framework/shortcuts/data/shortcuts.xml
```diff
@@ -1226,4 +1231,545 @@
...
+    </SC>
+  <SC>
+    <key>edit-style</key>
+    <seq></seq>
+    <ctx>notation-opened</ctx>
+    </SC>
+  <SC>
+    <key>interval-2</key>
+    <seq></seq>
+    </SC>
```
VanSHOE, line 1408: Somehow I have missed a couple contexts here
I have noted this and will fix this alongside any further requested changes

MarcSabatella:
Of these, Special Characters is the most important from an accessibility and general usability standpoint, especially for people writing in languages other than English where you almost constantly need to access these.  The dialog exists already and can be opened from the Properties panel, but that's definitely far from ideal.  I'm assuming that re-adding a command to simply open the dialog would be not be difficult.

The command to open the symbols palette, like the also-missing commands to open the key signatures and time signatures palettes, is a bit different.  What it really does is open the Master Palette but do so already focused on the correct section.  So it's kind of a special version of the command to open the master palette, and it was probably a bit of a hack.  These individual commands are quite useful, but realistically, if it's hard to implement them in the current framework, a passable substitute would be if the master palette remembered which section you were last viewing.  So if you are dealing in symbols a lot, you open the master palette, change to Symbols (possible in a single keystroke - press "End"), do what you want, then the next time you open the master palette, it's already focused on the symbols section.

VanSHOE:
Update: I believe the action itself exists which is called `show-keys`. However the shortcuts themselves are not picked up while typing. This particular issue might need a rework.

MarcSabatella:
At least in MU3, if a shortcut claimed to be active in text edit mode, then it really was active in text edit mode. People manage to shoot ourselves in the foot with this reasonably often, so I'm perfectly happy if this is no longer possible.

I don't have much of an opinion on whether or not a code change to allow function keys while in edit mode should be a separate PR.  I guess I'd lean towards saying do it separately to maximize the chances of this one being approved sooner, in case there are issues with the code for the other.

VanSHOE:
Now assuming the above PR gets approved, I'll push to set visibility of shortcut to true here

VanSHOE:
Now, assuming the above PR is approved, I'll push (I won't force push as a precaution for easy rollback) to change the visibility of this shortcut to true.

VanSHOE:
I have pushed this, and after seeing the xml, I notice some actions are repeated so I would need to do some extra cleanup as well

VanSHOE:
Please have a look @bkunda @MarcSabatella. 
The shortcut should be working now. PR has also been rebased to latest master

MarcSabatella:
It looks good to me.  I will say, it's surprising to me that we should need to add all these empty shortcuts explicitly just to get things to appear in the list; I'd have assumed a simple code change somewhere would be easier.  In 3.x, the shortcuts.xml file contained only the commands with default shortcuts, and it worked fine.

But I know from the "top staff" command, which had an empty shortcut all along, that it does basically work.  See https://github.com/musescore/MuseScore/issues/11522, though, to be sure it *really* works, as that command has some quirks that I don't know if anyone completely figured out.

MarcSabatella:
I see.  This public/private actions thing is new; it used to be that all actions (aka commands) were the same.  I see now you were specifically advised to use this approach, so, all is well.

igorkorsukov:
@VanSHOE 
There is some kind of problem there, an assert is triggered at the start of the application, it looks like the same shortcut is written twice in the file with different contexts

image

VanSHOE:
@igorkorsukov Also a second doubt I had was that in my project I was also working on auto addition of shortcuts, so would this PR be necessary after that?

igorkorsukov:
@VanSHOE at the moment I have revert back this PR.

>I added here did in fact need to have notation-opened context? (Apart from the ones already present)
We need to investigate each shortcut. For example, for `pitch-up`, the context should be `notation-staff-not-tab`, because another action should be performed for the tab staff

igorkorsukov:
> @igorkorsukov Also a second doubt I had was that in my project I was also working on auto addition and categorisation of shortcuts, so would this PR be necessary after that?

In fact, probably after your work this PR will not be needed

PR #11604, opened 2022-05-11
Fix professional pitch range and musicXMLid of Tarogato instrument

cbjeukendrup:
I updated the spreadsheet and ran the `update_instruments_xml.py` script to update `instruments.xml`.


shoogle:
Thanks! Do you have a source for the pitch range change? Obviously its wrong for the max professional pitch to be lower than the max amateur one, but it might be the amateur one that's too low rather than the professional one being too high (cc. @oktophonie).

PR #11608, opened 2022-05-11
[MU4] Fix incorrect beaming for tuplets

asattely:
Tuplet sub-beams were not behaving like in 3.6, so this PR fixes them to work like the old ones. If we need to change them based on new design specs for 4.0 we can do that, but this one at least restores the original behavior.

There is also something to be gained from investigating how beamed-together tuplets determine whether brackets are necessary.

Before (example):
image

After (different example):
image

edit: also, tuplet beams were weirdly askew, which I also fixed.

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -96,8 +96,8 @@ Beam::Beam(const Beam& b)
 {
     _elements     = b._elements;
     _id           = b._id;
-    for (const LineF* bs : b._beamSegments) {
-        _beamSegments.push_back(new LineF(*bs));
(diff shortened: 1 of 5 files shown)


asattely:
vtests are split between better and worse, i will jump on that presently

asattely:
vtests lookin' beautious, I will recommend this get the vtest tag.

Note that some of the vtests still look a little haggard--particularly ones where there are split beams within a single staff (that is, one where the height of the beam is manually adjusted). Manual adjustments are still something that need to be tackled, and are outside of the scope of this PR.

oktophonie (approved):

RomanPudashkin (approved):

PR #11609, opened 2022-05-11
MusicXML: Export invisible noteheads as "none"

shoogle:
Avoid setting `<note print-object="none">` because that hides other parts of the note in addition to the head, which causes problems for some MusicXML importers and Braille conversion tools. Use `<notehead>none</notehead>` instead.

Resolves: https://musescore.org/en/node/327118

MusicXML documentation for:
- [`<note print-object="no">`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/note/)
- [`<notehead>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/notehead/) and its ["none" value](
https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/notehead-value/).


Eism (approved):

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Tested on  [headless.zip](https://github.com/musescore/MuseScore/files/8709640/headless.zip) from https://musescore.org/en/node/327118

PR #11624, opened 2022-05-13
Fix #9345: KeySig in hidden staves cause additional space

### LINKED ISSUE #9345: [MU4 Issue] Space is allotted unnecessarily for key signatures on hidden staves
unnecessary-time-signature-space

In the second system, the alto saxophone stave is hidden but empty space is allotted for the key signature that would appear on it. Space should only be made for key signatures that are actually visible on the system.

The same problem affects key changes (see the second bar on the second system).

See discussion at: https://musescore.org/en/node/50316

mike-spa:
Fix #9345. A simple but effective solution. As a side benefit, we now avoid creating the shapes for hidden staves and thus save unnecessary operations.

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -605,6 +605,14 @@ void LayoutSystem::hideEmptyStaves(Score* score, System* system, bool isFirstSys
         SysStaff* ss = system->staff(staff->idx());
         ss->setShow(true);
     }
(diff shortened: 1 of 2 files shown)


mike-spa:
Vtests show the correct result: the differences are due to the bass clefs in hidden staves, which used to cause added space (bass clef is slightly wider than the treble clef). Now that is also corrected.

Eism (approved):

oktophonie (approved):

Issue #11455, opened 2022-04-29
[MU4 Task] "Replace Instrument" dialog: allow double-clicking instrument to choose it


cbjeukendrup:
**Task description**
In the "Replace Instrument" dialog, it would be a nice improvement if we could double-click an instrument name in the list of available instruments to choose that instrument, without pressing the OK button. 

**Hints for contributors**
The code for the "Replace Instrument" dialog lives in `InstrumentsDialog.qml`. This file is shared between the "Add instruments" dialog (which can be opened from the Instruments panel) and the "Replace Instrument" dialog. The difference between these dialogs is made by the `canSelectMultipleInstruments` property. 
When adding double-clicking for the "Replace Instrument" dialog, make sure not to break it in the "Add Instruments" dialog.

PR #11630, opened 2022-05-13
fix:#11455: Double clicking instrument automatically submits

VanSHOE:
Resolves: #11455 

A simple feature that allows double-clicking on an instrument in the "Replace Instrument" dialogue to select it rather than having to click the OK button.

### DIFF
--- a/src/instrumentsscene/qml/MuseScore/InstrumentsScene/ChooseInstrumentsPage.qml
+++ b/src/instrumentsscene/qml/MuseScore/InstrumentsScene/ChooseInstrumentsPage.qml
@@ -39,6 +39,8 @@ Rectangle {
 
     property NavigationSection navigationSection: null
 
+    signal submitRequested()
+
     function instruments() {
         if (root.canSelectMultipleInstruments) {
             return instrumentsOnScoreView.instruments()
@@ -142,6 +144,10 @@ Rectangle {
 
             onAddSelectedInstrumentsToScoreRequested: {
                 prv.addSelectedInstrumentsToScore()
+
+                if (!root.canSelectMultipleInstruments) {
+                    root.submitRequested()
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/instrumentsscene/qml/MuseScore/InstrumentsScene/ChooseInstrumentsPage.qml
```diff
@@ -142,6 +142,10 @@ Rectangle {
 
             onAddSelectedInstrumentsToScoreRequested: {
                 prv.addSelectedInstrumentsToScore()
+                if(!root.canSelectMultipleInstruments && invoker)
+                {
+                    invoker.submit()
+                }
```
cbjeukendrup, line 148: And here, you would "emit" that signal:
```qml
    root.submitRequested()
```
(and don't forget to remove the check for `invoker` too)

src/instrumentsscene/qml/MuseScore/InstrumentsScene/InstrumentsDialog.qml
```diff
@@ -55,7 +62,7 @@ StyledDialogView {
 
             canSelectMultipleInstruments: root.canSelectMultipleInstruments
             currentInstrumentId: root.currentInstrumentId
-
+            invoker: root
```
cbjeukendrup, line 65: And here, you would listen for the signal, and handle it:
```qml
    onSubmitRequested: {
        root.submit()
    }
```

This way, the ChooseInstrumentsPage does not depend on specific characteristics of any of its parent items ("characteristics" as in having a `submit()` function). But it just tells the parent that something happened, and the parent can act as it sees fit.

You will notice this pattern of having `...Requested()` signals very often in our QML code. It is good to use similar patterns whenever you need to communicate something upwards in the view hierarchy (from child component to parent component).

cbjeukendrup (commented):
Thanks for your contribution! I left three comments that form a story together :)

Beside those comments, I would like to ask you to pay a little bit more attention to the code style. Please add some newlines before and after signals and functions that you declare, add a space between `if` and `(`, and put the opening `{` at the same line as the `if`.

cbjeukendrup (approved):
Thanks!

Issue #11533, opened 2022-05-05
[MU4 Issue] Project Properties - Worktitle and Creation date fields don't save data properly


DmitryArefiev:
**To Reproduce**
1. Add some text to **Worktitle** - click OK
2. Reopen Project Properties  - **text doesn't save** (BUG1)
3. Add some text to **Creation date** - click OK
4. Reopen Project Properties  - **field is empty** (BUG2)

https://user-images.githubusercontent.com/90187801/166938708-50e005d9-3faf-4196-a6f0-f23fbc1f0a93.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

VanSHOE:
@DmitryArefiev Regarding the date issue: it must be in the format: `ddd MMM d yyyy` as Qt::TextDate only accepts that format as seen [here in docs](https://doc.qt.io/qt-5/qt.html#:~:text=Description-,Qt%3A%3ATextDate,-0) 
I'll be making a PR related to the title issue in a couple of minutes though

Please let me know if the issue regarding the date still needs fixing though, I'll be happy to contribute!

cbjeukendrup:
About the date, I believe @HemantAntony is still planning to build a decent date picker component in the future, so that the user doesn't need to type the date in a certain format. We're waiting with that until we switch to Qt 6, because the QML component for making such a date picker will be replaced by another component in Qt 6, and it's bad if the work has to be done twice.

VanSHOE:
> About the date, I believe @HemantAntony is still planning to build a decent date picker component in the future, so that the user doesn't need to type the date in a certain format. We're waiting with that until we switch to Qt 6, because the QML component for making such a date picker will be replaced by another component in Qt 6, and it's bad if the work has to be done twice.

Should I mark in the PR that is partially resolves this issue then?

VanSHOE:
> > About the date, I believe @HemantAntony is still planning to build a decent date picker component in the future, so that the user doesn't need to type the date in a certain format. We're waiting with that until we switch to Qt 6, because the QML component for making such a date picker will be replaced by another component in Qt 6, and it's bad if the work has to be done twice.
> 
> Should I mark in the PR that is partially resolves this issue then?

cbjeukendrup:
Yes, that's a good idea.

GrahamDickins:
Dear all,

[...]

PR #11634, opened 2022-05-14
fix: 11533; Title is also checked when checking equality of meta struct

VanSHOE:
Part of #11533 

When checking the equality of meta info struct, the title attribute was not considered. I have added that now.

The date issue mentioned was because of the wrong format of date inputted. To make it clear, should I add the format in brackets after "Creation date"?

### DIFF
--- a/src/project/projecttypes.h
+++ b/src/project/projecttypes.h
@@ -168,6 +168,7 @@ struct ProjectMeta
     bool operator==(const ProjectMeta& other) const
     {
         bool equal = filePath == other.filePath;
+        equal &= title == other.title;
         equal &= subtitle == other.subtitle;
         equal &= composer == other.composer;
         equal &= lyricist == other.lyricist;


src/project/projecttypes.h
```diff
@@ -169,6 +169,7 @@ struct ProjectMeta
     {
         bool equal = filePath == other.filePath;
         equal &= subtitle == other.subtitle;
+        equal &= title == other.title;
```
cbjeukendrup, line 172: I think it would be great to keep the struct members in the same order as they are declared (just to avoid confusion in the future). So could you please put the line about `title` above `subtitle`?

VanSHOE: Should I also move the thumbnail before the creation date?

cbjeukendrup: Nah, let's leave that one at the bottom, since it's a special case.

cbjeukendrup (approved):

cbjeukendrup:
@DmitryArefiev this PR only intends to solve BUG1, so you only need to check that one :)

src/project/projecttypes.h
Jojo-Schmitz: It is unusual, but possible, for a **score** to have a _partName_ too, shouldn't that get checked too?

cbjeukendrup: If there is a partName, that will be part of `additionalTags`, so it is indeed checked too.

Jojo-Schmitz: Ah, I see

PR #11646, opened 2022-05-16
Fix #10505: Added button order

### LINKED ISSUE #10505: [MU4 Issue] Wrong OK/Save/Yes buttons order in dialogs on Windows (uses Mac style)
**Describe the bug**
MU4 windows builds now using **Mac style** for OK/Cancel buttons order in dialogs. 
It should be changed to Windows style - **OK/Save/Yes buttons should be from the left side** 

**MU4:**

MU4
MU4 2
MU4 3

**MU3:**
MU3
MU3 2
MU3 3

**Expected behavior**
OK/Save buttons should be from the left side (See MU3 (or other Win app) to compare)

**Platform information**
Windows 10

**Additional context**
Doesn't occur in **MU3**

Some dialogs have the **right** order:  
Style, Page settings, Add/Remove system breaks, Transpose

Another dialogs with the **wrong** order:
New Score wizard, Export, Score properties, Parts, Reset palette

HemantAntony:
Resolves: #10505 

This PR adds a ButtonBox (along with ButtonBoxItem) component for creating and arrange commonly used button in the bottom panels of dialogs


HemantAntony:
@RomanPudashkin This PR is ready for review

HemantAntony:
@DmitryArefiev One other change this PR implements is that there is at least one button in a dialog being accented

Buttons of following dialogs have been rearranged:
About MusicXML Dialog
Preferences Dialog
Export Dialog
New Score Dialog
Save to Cloud Dialog
Parts Dialog
Workspaces Dialog
New Workspace Dialog
Palette Properties Dialog
Project Properties Dialog
Edit Grid Size Dialog (From appearance settings)
Palette Cell Properties Dialog
Edit Shortcut Dialog
Select Measures Count Dialog
Edit Midi Mapping Dialog
Instruments Dialog

DmitryArefiev:
> > @HemantAntony I've tested on Windows and found 3 dialogs still have wrong order:
> 
> There's something missing

@Jojo-Schmitz Yeah, thanks! Pressed Ctr+Return to early :)
Edited now.

DmitryArefiev:
> > @DmitryArefiev One other change this PR implements is that there is at least one button in a dialog being accented
> > Buttons of following dialogs have been rearranged: About MusicXML Dialog Preferences Dialog Export Dialog New Score Dialog Save to Cloud Dialog Parts Dialog Workspaces Dialog New Workspace Dialog Palette Properties Dialog Project Properties Dialog Edit Grid Size Dialog (From appearance settings) Palette Cell Properties Dialog Edit Shortcut Dialog Select Measures Count Dialog Edit Midi Mapping Dialog Instruments Dialog
> 
> @HemantAntony I've tested on Windows and found 3 dialogs still have wrong order:
> 
>     1. Do you want to save changes..
> 
>     2. Restore Palette
> 
>     3. Save to cloud
> 
> 
> buttons1 buttons2 buttons3

@HemantAntony If you want we can merge what we have now and fix those three dialogs later

HemantAntony:
@DmitryArefiev I have fixed issue 3. Issues 1 and 2 are almost fixed. They are currently causing accessibility problems. I'm trying to work on it, but it does seem a bit difficult for me. I'll tell you tomorrow if I will be able to fix those dialogs

HemantAntony:
@DmitryArefiev 5 days went by quickly :). I have updated the PR

DmitryArefiev:
> @DmitryArefiev 5 days went by quickly :). I have updated the PR

@HemantAntony Cool! Now those dialogs have right order. But I found a regression (after latest changes):
some dialogs have wrong centering to the left edge instead of right edge (on Mac build as well)


Capture
Capture2
Capture3

HemantAntony:
@DmitryArefiev I have updated the PR. The artifacts are ready

DmitryArefiev:
> @DmitryArefiev I have updated the PR. The artifacts are ready

Yes, now it looks good. Thanks!

DmitryArefiev:
@Eism Please review :)

Tantacrul:
Would be good to merge this bad boy when @Eism has time. (Assuming you are happy with it)

HemantAntony:
I have resolved conflicts :)

src/framework/uicomponents/view/buttonboxmodel.h
```diff
@@ -0,0 +1,188 @@
...
+
+        LayoutButton(QString _text, ButtonType _buttonType,  ButtonRole _buttonRole, bool _isAccent, bool _isLeftSide = false,
+                     QString _navigationName = "")
+            : text(_text), buttonType(_buttonType),  buttonRole(_buttonRole), isAccent(_isAccent), isLeftSide(_isLeftSide),
+            navigationName(_navigationName)
+        {
+        }
+    };
+
+    const char* c = "uicomponents"; //To make Following Hash look neat
```
cbjeukendrup, line 136: This doesn't work: the context must be written inline, otherwise `lupdate` won't recognise the translatable string. Unfortunately.

(I'm currently collecting all PRs that affects translatable strings, so that's why I'm randomly posting a random comment 😁)

cbjeukendrup:
Rebase needed, by the way (again)

Eism:
@HemantAntony 
Sorry for the late review. 

I would suggest thinking about improving these points:
1. What to do if there are unique buttons?  [first](https://github.com/musescore/MuseScore/pull/11646/files#diff-f1984e943deb351c54822999df5bfaf568bf9755319f8f7136073af9aa0ef768R116), [second](https://github.com/musescore/MuseScore/pull/11646/files#diff-0e31828af21286452a3ce366bf2515ac0973fa5529bf0cebb5dcb0c899a43c3fR164)
    The component must be able to get a list of buttons and sort them
    
2. What to do if some buttons have unique properties? see [delegateProperties](https://github.com/musescore/MuseScore/pull/11646/files#diff-29ba6aa2e2c635c2d3368ae42f60a95f34fd7382c9b7938f131296a1b06cc913R214)
   The component must be able to get a list of custom buttons with unique properties


I would suggest designing the API for the ButtonBox component as for Interactive
```
    virtual Result question(const std::string& title, const std::string& text, **const Buttons& buttons**, const Button& def = Button::NoButton,
                            const Options& options = {}) const = 0;

    virtual Result question(const std::string& title, const Text& text, **const ButtonDatas& buttons**, int defBtn = int(Button::NoButton),
                            const Options& options = {}) const = 0;
```

The component receives two lists as input - a list of button enumerators and a list of custom buttons.

```
property list<int> buttons // buttons enums, we should construct ButtonBoxItems in ButtonBox and ignore customButtons
property list<ButtonBoxItem> customButtons // we should ignore buttons and use only customButtons
```

Only after the list of buttons is ready to sort them

HemantAntony:
Hey, so I've started working on your suggestions. Just to clarify your first suggestion, you are thinking of allowing Dialogs to explicitly order buttons? `ButtonBoxItem` already has a `buttonRole` property which you can set. These buttons are then ordered by roles

Eism:
> you are thinking of allowing Dialogs to explicitly order buttons?

No
I like your structure - button ordering in a separate ButtonBox component.

In the first suggestion, I propose to do 
```
        ButtonBox {
            customButtons: [
                ButtonBoxItem { text: "Cancel", role: ButtonBoxModel.Cancel } // need to think about how best to create global buttons
                ButtonBoxItem { text: "Add", role: ButtonBoxModel.Custom + 1 }
            ]
        }
```
Instead of
```
        ButtonBox {
            buttons: ButtonBoxModel.Cancel

            ButtonBoxItem { text: "Add" }
        }
```

HemantAntony:
Okay thanks! I understood. Could you also clarify your second suggestion too please :)

Eism:
Yes, sure
 
 If we need a button with unique behavior or properties, then using the previous example we can solve the problem like this
 
```
     ButtonBox {
        customButtons: [
            ButtonBoxItem { text: "Cancel", role: ButtonBoxModel.Cancel } // need to think about how best to create global buttons
            ButtonBoxItem { 
                text: "Add" 
                role: ButtonBoxModel.Custom + 1 
            
                accessible.ignored: true
                navigation.onActiveChanged: {
                    // do something
                }
            }
        ]
    }
```

Jojo-Schmitz:
Rebase needed

Any other reason not to merge it?

bkunda:
Hi @HemantAntony. Any updates on this one? 
@Tantacrul curious to hear your thoughts about whether this is still something we ought to pursue (the current dialog button orders have been in place for a little while now).

Tantacrul:
Is this the Mac VS Windows thing?

I'll give my unvalidated opinion on this. The ordering of cancel / save, etc. is a topic that gets a lot of developers really riled up but which I do not believe makes the slightest bit of difference in reality.

We all have to deal with modals across tablet apps, mobile apps, websites, OS's and have just learned to choose the button that has the right text inside it.

So my opinion is: it doesn't matter either way, therefore inertia is the best option.

Jojo-Schmitz:
Yes it is Mac vs. Windows. Qt knows about the differences, it and when. QButtonBox is used, which MuseScore uses in some parts, but not in others, so the button order is inconsistent.

MarcSabatella:
I have never particularly noticed that there was any sort of specific standard for the placement of buttons as I'm constantly moving back and forth between Linux, Windows, ChromeOS, iOS, and Android - plus web apps - and hence everything is all over the place anyhow.  I would personally have a slight preference for MuseScore behaving the same on all platforms, but to be honest I don't care.

What I *do* wonder about, though, is what makes the most sense accessibility-wise.  Tab takes you to the first button in the button box.  To me, it feels natural that this should be OK and not Cancel when those are the options, or Save vs Don't Save unless it's a situation where we really want the default action to be not to save.  @shoogle - this is something James Bowden from RNIB touched on in that document I forwarded earlier, although he's specifically talking about the new score wizard which is problematic in other ways as well.  If Tab can reach both OK and Cancel, it matters less which comes first.  But if Tab only reaches one of the buttons and you need to switch to cursor for the others, that could make it more important to choose wisely.

shoogle:
@MarcSabatella, we should fix the accessibility problem separately. Tab should visit all buttons in dialogs.

Back to the issue at hand, personally I'm not convinced that button order matters enough to justify the extra time and effort needed to design, develop, test and maintain a different button order for each platform. Also, I quite like the fact that MuseScore uses the same button order on all platforms, though I appreciate us multi-platform users will be in the minority.

Attempting to follow the OS button order can lead to strange UI bugs/inconsistencies. Consider the strange positioning of the **'Apply to all parts'** button in the _Format > Style_ dialog on Windows. This placement might be correct for a standard 'Apply' button, but I would argue that 'Apply to all parts' is not a standard 'Apply', so it seems wrong to put it next to 'OK' and 'Cancel'.

macOS | Windows
:---:|:---:
MuseScore_Styles_macOS | image

These screenshots are from MuseScore 3, but I just checked and it's still like that in MuseScore 4.

Tantacrul:
> so it seems wrong to put it next to 'OK' and 'Cancel'.

@shoogle it's an action button, so should go next to the other action buttons. I think the current placement of 'Apply to all parts is wrong', although I won't comment on the ordering.

HemantAntony:
@Tantacrul @bkunda Hello, I don't think I will be able to complete this PR soon. I had to implement a custom "Container" in QML to sort it out which is a lot of work. So, it would take way longer to fix any bugs that come up (not to mention code reviews).

But, right now, apparently there is a QML component made for this ['DialogButtonBox`](https://doc.qt.io/qt-6/qml-qtquick-controls-dialogbuttonbox.html#details). This is implemented in Qt 6.5 LTS. So, I know there is a PR for updating to Qt 6 by @cbjeukendrup. I'm not sure if its specifically Qt 6.5, though. I think we should wait until there is an update to Qt 6.5 and then implement this. Not to mention, as you said, this really isn't an important bug.

PR #11650, opened 2022-05-16
Fix #10888: Enable Tuplet Button Always

### LINKED ISSUE #10888: [MU4 Issue] Hiding and showing the tuplet button causes it to appear disabled
**Describe the bug**
When the tuplet button in the note input toolbar is hidden then shown again, it returns in an apparently disabled (greyed-out) state. 

disabled-tuplet-icon

This doesn't affect its functionality, but it does present as a bit strange because the behaviour is not replicated across other buttons in the toolbar. 

**Expected behavior**
Triplet icon shouldn't appear greyed out when changing visibility states.
When nothing is selected in the score, clicking on the tuplet button should perform no action (as per current behaviour of "Tie", "Slur" and "Flip direction" buttons).

**Platform information**
 - MacOS 12.1 (M1)
- Windows 11 

**Note**
The weird scrolling behaviour you see will be the subject of another issue (@Tantacrul FYI)

HemantAntony:
Resolves: #10888

This PR always enables the Tuplet Button regardless of the selected element

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -154,16 +154,16 @@ void NotationActionController::init()
     registerAction("add-tenuto", [this]() { addArticulation(SymbolId::articTenutoAbove); });
     registerAction("add-staccato", [this]() { addArticulation(SymbolId::articStaccatoAbove); });
 
-    registerAction("duplet", [this]() { putTuplet(2); });
-    registerAction("triplet", [this]() { putTuplet(3); });
-    registerAction("quadruplet", [this]() { putTuplet(4); });
-    registerAction("quintuplet", [this]() { putTuplet(5); });
-    registerAction("sextuplet", [this]() { putTuplet(6); });
-    registerAction("septuplet", [this]() { putTuplet(7); });
-    registerAction("octuplet", [this]() { putTuplet(8); });
-    registerAction("nonuplet", [this]() { putTuplet(9); });
+    registerAction("duplet", [this]() { putTuplet(2); }, &Controller::noteOrRestSelected);
+    registerAction("triplet", [this]() { putTuplet(3); }, &Controller::noteOrRestSelected);
+    registerAction("quadruplet", [this]() { putTuplet(4); }, &Controller::noteOrRestSelected);
+    registerAction("quintuplet", [this]() { putTuplet(5); }, &Controller::noteOrRestSelected);
+    registerAction("sextuplet", [this]() { putTuplet(6); }, &Controller::noteOrRestSelected);
+    registerAction("septuplet", [this]() { putTuplet(7); }, &Controller::noteOrRestSelected);
+    registerAction("octuplet", [this]() { putTuplet(8); }, &Controller::noteOrRestSelected);
+    registerAction("nonuplet", [this]() { putTuplet(9); }, &Controller::noteOrRestSelected);
     registerAction("custom-tuplet", &Controller::putTuplet);
-    registerAction("tuplet-dialog", &Controller::openTupletOtherDialog);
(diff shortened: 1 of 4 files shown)


HemantAntony:
@bkunda This always enabled the Tuplet button even if something non applicable is selected

bkunda:
Hi @HemantAntony,
Thanks for working on this. It is _almost_ perfect 🙂

If no applicable element is selected when the tuplet button is clicked, it should just perform **no** action, rather than triggering the menu. 

The behaviour should be the same as for the tie and slur buttons:

https://user-images.githubusercontent.com/86290556/168599246-29b53d84-8518-43d4-a7e3-3f4763c053ba.mov

Thanks!

Eism:
@HemantAntony I don't see any updates on this PR
Do you plan to finish fixing the issue? Or you need help?

HemantAntony:
@Eism Sorry! I forgot about this. It seems like an easy fix. I'm going to work on this

HemantAntony:
@bkunda I know you said the behaviour should be similar to that of Slurs and Ties

But I think the Cross-staff beaming button is similar to how the tuplet buttons should behave. The cross-staff beaming button just disables the menu items:

https://user-images.githubusercontent.com/88136054/170654830-ac6f0ab8-523d-443f-adc9-1cff5cbf5071.mp4


But, if you do want the behaviour to be that of Slurs and ties, then shouldn't the cross-staff beaming button also behave like that?

bkunda:
Actually I think you're right here, @HemantAntony. Because the tuplet button triggers a menu, rather than working as a "switch", it would make sense if its behaviour was the same as other menu-triggering buttons, like the cross-staff beams button. 

Thanks for this suggestion :-)

HemantAntony:
I have updated the PR, but something to note is that the tuplet buttons will be disabled is the Add menu too

https://user-images.githubusercontent.com/88136054/170811177-dbb2e1db-c7b5-4448-a475-708950e986c4.mp4

Eism:
@bkunda 
Also, they will also be enabled if there is a selection on the score, no matter which one
Previously it depended on a note or a rest was selected
Do you agree?

bkunda:
> @bkunda
> Also, they will also be enabled if there is a selection on the score, no matter which one
> Previously it depended on a note or a rest was selected
> Do you agree?

Can it still be conditional on whether a note or rest is selected? 
If I understand you right, the user will see these options enabled if they select _anything_ (e.g. text objects). This doesn't make much sense.

HemantAntony:
> Can it still be conditional on whether a note or rest is selected? If I understand you right, the user will see these options enabled if they select _anything_ (e.g. text objects). This doesn't make much sense.

Yeah I accidently added that. I'll add that condition for note / rest selection too

bkunda:
> > Can it still be conditional on whether a note or rest is selected? If I understand you right, the user will see these options enabled if they select _anything_ (e.g. text objects). This doesn't make much sense.
> 
> Yeah I accidently added that. I'll add that condition for note / rest selection too

Great. Thanks @HemantAntony. Probably worth me giving it one quick last test before we merge. Could you let me know when it's ready to go please?

src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1729,6 +1729,17 @@ Ms::EngravingItem* NotationActionController::selectedElement() const
     return selection ? selection->element() : nullptr;
 }
 
+bool NotationActionController::noteOrRestSelected() const
+{
+    EngravingItem* element = selectedElement();
```
Eism, line 1734: Just copy the logic from resolveTupletEnabled
Your code will not work for range selection

bkunda: Good spot, @Eism!

HemantAntony:
@Eism I have updated the PR

Issue #11181, opened 2022-04-13
[MU4 Issue] Improvement. Score view is not move back to the current playback cursor during playback


boryskuzmenko:
**Describe the bug**
We have playback option 'Pan score automatically'. When this option is disabled I can move/scroll a score sheet without any limits, when this option is enabled (default) I can't really move around the score (the same is in Sibelius). My point is that it looks weird and is not really good from UX side. So, I propose to look to approach which used in UG playback: during playback you can scroll around the tab until you do scrolling, but after around 1 sec after no action with the tab user is moved back to the current playback position. 
Pros from this improvement:
- user can move around a score and select any note/measure 
- in any case (I propose after 2 sec) cursor will returned back to the current playback position
- no visual artifacts like this (when 'Pan score automatically' is enabled)
- consistency between Muse products

**Screenshots**
**MU4 ('Pan score automatically' in both states)**

https://user-images.githubusercontent.com/101250347/163205894-5618393d-888e-426c-af0d-33fa99a6370e.mov

**UG**

https://user-images.githubusercontent.com/101250347/163206688-4a8b1111-708b-4ce8-b463-ace6e3344463.mov

**Platform information**
 - OS: macOS, Windows, Linux

Tantacrul:
Nice suggestion!

boryskuzmenko:
🟢 FIXED
Tested by @Tantacrul

PR #11655, opened 2022-05-16
[MU4] Fix #10779: Initiating 'Play' automatically scrolls the score to the top of the page. Vertical scrolling should be disabled

### LINKED ISSUE #10779: Initiating 'Play' automatically scrolls the score to the top of the page. Vertical scrolling should be disabled.
**Describe the bug**
In this example (see video below), I have a large score with multiple instruments. Pressing play seems to vertically scroll the score back to the top of the page.

Obviously, this makes sense if the playback is occurring on a different system. For example, a solo piano score will require vertical scrolling if the user is zoomed in. However, when the system being played back is in view, there should be no vertical scrolling.

https://user-images.githubusercontent.com/21022311/157868158-f6ecf667-9cde-4d67-b709-500be858c814.mov

**Expected behaviour**
Where there is only one system on a page, the vertical position of the score should be retained during playback. This vertical position should be held (I.e. **not** changed) when playback moves to other pages containing only one system. 

Where playback moves to a page containing more than one system, the score should be repositioned according to present behaviour in MS4. 

See this document for further illustration of this: 
[playback-panning-v1.pdf](https://github.com/musescore/MuseScore/files/8581890/playback-panning-v1.pdf)

(N.B. I have tried to accurately paraphrase @MarcSabatella's descriptions below. @MarcSabatella please advise in the comments if any of this needs correcting. Also note that we are going for a lean fix here that addresses the present issue in situations with one system on a page only. Pages containing multiple systems should not be affected by this fix).

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10779
Resolves: https://github.com/musescore/MuseScore/issues/11181
Closes: https://github.com/musescore/MuseScore/pull/10799

### DIFF
--- a/src/notation/notationtypes.h
+++ b/src/notation/notationtypes.h
@@ -33,6 +33,7 @@
 
 #include "libmscore/engravingitem.h"
 #include "libmscore/page.h"
+#include "libmscore/system.h"
 #include "libmscore/durationtype.h"
 #include "libmscore/mscore.h"
 #include "libmscore/masterscore.h"
@@ -58,6 +59,7 @@
 
 namespace mu::notation {
 using Page = Ms::Page;
+using System = Ms::System;
 using EngravingItem = Ms::EngravingItem;
 using ElementType = Ms::ElementType;
 using PropertyValue = engraving::PropertyValue;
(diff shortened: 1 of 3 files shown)


Eism (approved):

boryskuzmenko:
🟢 **FIXED**
Tested by @Tantacrul

Issue #11410, opened 2022-04-26
[MU4 Issue] Too easy to select measures or click wrong element


shoogle:
## Description

When trying to click on a note in the score, it is too easy to miss the note and click on the measure behind it instead.

## Recommended solution / Expected behaviour

Set the default "selection proximity" value (found in Preferences > Canvas) to 5px.

### Report from an accessibility tester:

> It is taking a lot of getting used to, selecting a single note vs the entire measure. I probably select the measure 3 or 4 times for every time I try to select a note. My mouse cursor must be touching just outside of a note I want to select. This feature, selecting the entire measure might be better off with a mouse modifier like Alt/Option
>
> MacOS Monterey 12.3.1 16GM RAM Apple M1
> Using Zoom (system magnification tool, not video conferencing software)

I think it is mainly a problem on macOS, though other platforms may be affected to a lesser extent.

### macOS cursor problem

On macOS the cursor/pointer has border and the cursor's "hotspot" is on the inside of this border rather than the outside, which I find to be rather unintuitive. This means you are not quite clicking where you think you are.

<img width="300" alt="macOS cursor hotspot placement" src="https://user-images.githubusercontent.com/11011881/165360668-3f72715f-ba58-4ba0-b402-81a97a4e8928.svg">

This video shows how it can lead to problems selecting objects in the score. **(Video has sound!)**

https://user-images.githubusercontent.com/11011881/165380364-c2598e02-ef76-4ee1-83f7-8e83162f1803.mp4

RobFog:
If the cursor behaves the same in Calculator (and likely elsewhere), I'm not convinced we should deviate from that.

shoogle:
@RobFog, I agree to an extent. When there are two similarly sized objects next to each other then I agree that we should follow the macOS convention. So we should consider this a click on the 5 button and not on the 8:

![image](https://user-images.githubusercontent.com/11011881/165390148-b8c756e0-a7dc-4808-900b-e0f56dbc0494.png)

But when there is a very small object in front of a very large object (or in front of an empty background), then I think it is safe to assume that this is a click on the object and not on the background:

![image](https://user-images.githubusercontent.com/11011881/165389692-23ae0460-fe2f-406e-ba9f-bdbead8eb008.png)

I'm just suggesting that we should make the click region a bit bigger for small objects. I certainly don't think we should add an artificial offset to the cursor position in MuseScore just to make up for my gripe with cursors on macOS!

[...]

PR #11661, opened 2022-05-17
Set default selection proximity to 5px

cbjeukendrup:
Resolves: #11410

This is the simplest suggested solution. If there are other concrete suggestions for making the "hit element" behaviour smarter, I could implement those too.


Eism:
@cbjeukendrup can we close this PR?

cbjeukendrup:
Yes, that's probably the outcome for now.

PR #11667, opened 2022-05-18
[MU4] Fix MSCV/MinGW compiler warnings

Jojo-Schmitz:
reg. unused variables, C4198 resp. -Wunused-variable and -Wunused-but-set-variable, introduced by #11608


src/engraving/libmscore/beam.cpp
```diff
@@ -939,9 +939,6 @@ void Beam::calcBeamBreaks(const ChordRest* chord, const ChordRest* prevChord, in
 
     // get default beam mode -- based on time signature preferences
     const Groups& group = chord->staff()->group(chord->measure()->tick());
-    Fraction stretch = chord->staff()->timeStretch(chord->measure()->tick());
```
RomanPudashkin, line 942: @asattely FYI

RomanPudashkin (approved):

Issue #11665, opened 2022-05-17
[MU4 Issue] Preferences>General>Folders: VST3 and SoundFonts pathes are cleaned up after closing dialogs with X or Escape


DmitryArefiev:
**To Reproduce**
1. Go Preferences>General>Folders
2. Make sure you have added directories to VST3 and SoundFonts
3. Click on folder icon to open Add directory dialog
4. Close it with X or Esc   

https://user-images.githubusercontent.com/90187801/168864352-10a1de52-c629-41bc-a05f-bf08e23b198e.mp4

**Platform information**
macOS, Windows, Linux

DmitryArefiev:
Fixed in #11680

PR #11680, opened 2022-05-18
fix: #11665; Multi-Selection directory box doesn't return empty path on being crossed out anymore

VanSHOE:
Resolves: #11665 

Instead of returning an empty list of paths whenever the dialogue box returned an error, we now return the previous list, which prevents the deletion of directories as described in the issue.

This does not prevent you from emptying the list manually. You may still clear the list as usual but remember to use the OK button to confirm that.

### DIFF
--- a/src/framework/global/internal/interactive.cpp
+++ b/src/framework/global/internal/interactive.cpp
@@ -170,7 +170,7 @@ io::paths Interactive::selectMultipleDirectories(const QString& title, const io:
 
     RetVal<Val> paths = open("musescore://interactive/selectMultipleDirectories?" + params.join("&").toStdString());
     if (!paths.ret) {
-        return io::paths();
+        return selectedDirectories;
     }
 
     io::paths result;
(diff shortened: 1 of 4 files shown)


src/framework/global/internal/interactive.cpp
```diff
@@ -170,7 +170,7 @@ io::paths Interactive::selectMultipleDirectories(const QString& title, const io:
 
     RetVal<Val> paths = open("musescore://interactive/selectMultipleDirectories?" + params.join("&").toStdString());
     if (!paths.ret) {
-        return io::paths();
+        return selectedDirectories;
```
Eism, line 173: Thanks!
I would also ask to get rid of this logic in the dialog.
And remove m_originDirectories from the model.
image

VanSHOE: @Eism Done! Please have a look

Eism (approved):

src/framework/global/internal/interactive.cpp
Eism: thanks!

DmitryArefiev:
Tested #11665 on Win10, Mac12, LinuxMint 20.2 - **FIXED**      

Thanks!

Issue #11294, opened 2022-04-21
[MU4 Issue] Crash when pressing arrow key when Bracket selected in Edit mode


cbjeukendrup:
**Describe the bug**
A crash occurs when you press an arrow key while a Bracket is selected in Edit mode. 

**To Reproduce**
1. Create a score with multiple instruments, and add a Bracket if not yet present;
2. Double-click the bracket to enter "Edit mode"
3. Press one of the arrow keys
A crash occurs. 

**Expected behavior**
No crash; maybe, the bracket should respond to the key press.

**Platform information**
 - OS: reported on macOS

DmitryArefiev:
Fixed in #11681

PR #11681, opened 2022-05-18
Fix crash when pressing arrow key with bracket in Edit Mode

cbjeukendrup:
Resolves: #11294

Again the same type of problem: during the edit, the brackets get deleted and replaced during re-layout, but somewhere we still try to use the old bracket which causes the crash. 

I had to do some refactoring in order to get the ScoreCallbacks object consistently registered as a "viewer" of the score, so that we can use the `onElementDestruction` callback in order to be able to remove references to the deleted object. After that, we attempt to select the newly created bracket.

### DIFF
--- a/src/notation/internal/igetscore.h
+++ b/src/notation/internal/igetscore.h
@@ -22,6 +22,8 @@
 #ifndef MU_NOTATION_IGETSCORE_H
 #define MU_NOTATION_IGETSCORE_H
 
+#include "async/notification.h"
+
 namespace Ms {
 class Score;
 }
@@ -33,6 +35,7 @@ public:
     virtual ~IGetScore() = default;
 
     virtual Ms::Score* score() const = 0;
+    virtual async::Notification scoreInited() const = 0;
 };
 }
 
--- a/src/engraving/libmscore/mscoreview.h
+++ b/src/engraving/libmscore/mscoreview.h
@@ -55,7 +55,7 @@ public:
 
     virtual void moveCursor() {}
 
-    virtual void adjustCanvasPosition(const EngravingItem*, int /*staffIdx*/ = -1) {}
+    virtual Score* score() const { return m_score; }
     virtual void setScore(Score* s) { m_score = s; }
     virtual void removeScore() {}
 
@@ -70,13 +70,11 @@ public:
     virtual void lyricsMinus() {}
     virtual void lyricsUnderscore() {}
 
-    virtual void onElementDestruction(EngravingItem*) {}
-
     virtual const mu::Rect geometry() const = 0;
 
     const std::vector<EngravingItem*> elementsAt(const mu::PointF&) const;
     EngravingItem* elementNear(const mu::PointF& pos) const;
-    Score* score() const { return m_score; }
+    virtual void adjustCanvasPosition(const EngravingItem*, int /*staffIdx*/ = -1) {}
 
 protected:
     Score* m_score = nullptr;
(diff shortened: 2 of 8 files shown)


Eism:
@cbjeukendrup 
Does this PR resolve #11294?

cbjeukendrup:
@Eism Yes, it does! Added it to the description.

src/notation/inotationinteraction.h
```diff
@@ -266,6 +268,8 @@ class INotationInteraction
 
     virtual void showItem(const Ms::EngravingItem* item, int staffIndex = -1) = 0;
     virtual async::Channel<ShowItemRequest> showItemRequested() const = 0;
+
+    virtual ScoreCallbacks* scoreCallbacks() = 0;
```
RomanPudashkin, line 272: We should not expose ScoreCallbacks (and onElementDestruction) in this interface

RomanPudashkin: I would suggest replacing MuseScoreView::onElementDestruction with async::Channel<EngravingItem*> MuseScoreView::elementDestroyed(). After this change, we will be able to connect to the channel where needed and update all local pointers to avoid crashes

src/notation/internal/igetscore.h
```diff
@@ -33,6 +35,7 @@ class IGetScore
     virtual ~IGetScore() = default;
 
     virtual Ms::Score* score() const = 0;
+    virtual async::Notification scoreChanged() const = 0;
```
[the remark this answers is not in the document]
cbjeukendrup: The problem is that `setScore` can be called at any time, outside the constructor, which happens for example in MasterNotation and ExcerptNotation. This means that we need to inform the ScoreCallbacks that the score has been replaced.

Anyway, my ultimate goal was to make sure that the `ScoreCallbacks` object is registered and unregistered as a "viewer" for the score at the correct times. So any better solution to achieve that would be preferable.

src/notation/internal/scorecallbacks.cpp
```diff
@@ -22,16 +22,44 @@
...
+}
+
+ScoreCallbacks::~ScoreCallbacks()
+{
+    if (m_score) {
+        m_score->removeViewer(this);
+    }
+}
+
+void ScoreCallbacks::setScore(Ms::Score* score)
```
RomanPudashkin, line 50: With this function we will have two independent ways to get a score (via m_score or via m_getScore), which is pretty confusing

src/notation/internal/igetscore.h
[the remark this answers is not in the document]
RomanPudashkin: It should be renamed to scoreInited(), at least

src/notation/inotationinteraction.h
cbjeukendrup: > We should not expose ScoreCallbacks (and onElementDestruction) in this interface

We need to use it in NotationNoteInput. It looks like the MuseScoreView interface was designed to have only one instance of it for each notation, so I would prefer if NotationNoteInput reuses the ScoreCallbacks from NotationInteraction. 

Ultimately I think the whole MuseScoreView interface needs some rethinking. We could split it into smaller groups of related callbacks, and maybe move some functionality to the score itself (for example, the `elementDestroyed` notification could be part of the score class).

RomanPudashkin (approved):

DmitryArefiev:
Tested #11294 on Win10, Mac12, LinuxMint 20.2 - **FIXED. No crash**

PR #11705, opened 2022-05-20
[MU4] Fixing a few extra rules for slants in cross-staff beams

asattely:
There were a few more rules for 1/4-space slants in cross-staff beams that had to be added in, since the regular way that is determined is mired in assumptions that all notes are all on the same staff. I've also cut down on the size of the cross-staff code in layout2().


asattely:
Not to be merged yet, by the way, I'm pushing it to see what the tests look like

oktophonie (approved):

PR #11717, opened 2022-05-22
Fix #11378. Add palettes not found message

### LINKED ISSUE #11378: [MU4 Issue] Palettes - Search - No context indication when results not found
**Describe the bug**
A short and clear description of what the bug is.

**To Reproduce**
1. Go to Palettes
2. Type something into search box that will not be found
3. Panel is blank without context

**Expected behavior**
If no results are found, should be some context provided (ex: No results found)

**Screenshots**

https://user-images.githubusercontent.com/30370609/165256662-2bad8eb9-2ed0-4b54-88a0-cdf974883d89.mov

**Platform information**
OS: macOS 12.2
Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2224366007
revision: github-musescore-musescore-9a7fabf

softcat477:
Resolves: #11378

Add a _No results found_ message when there are no results.


src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml
```diff
@@ -47,6 +47,8 @@ StyledListView {
     property string filter: ""
     property bool searchOpened: false
 
+    property bool isResultFound: true
```
RomanPudashkin, line 50: I would recommend declaring this property as follows:
isResultFound: count === 0

You can remove onCountChanged after this change

softcat477: Thanks for the feedback! I changed to `isResultFound: count !== 0` and removed `onCountChanged` to keep `PalettesPanel.qml` untouched.

RomanPudashkin (approved):

PR #11719, opened 2022-05-23
Replaced QFile to mu::io::File for gp import

igorkorsukov:


src/importexport/guitarpro/internal/importgtp.cpp
```diff
@@ -2940,9 +2941,9 @@ Score::FileError importGTP(MasterScore* score, const QString& name)
     // otherwise it's an older version - check the header
     else if (strcmp(&header[1], "FIC") == 0) {
         uchar l;
-        fp.read((char*)&l, 1);
+        fp.read((uint8_t*)&l, 1);
         char ss[30];
-        fp.read(ss, 30);
+        fp.read(reinterpret_cast<uint8_t*>(ss), 30);
```
Jojo-Schmitz, line 2946: Mind to explay why using different ways of casting here?

igorkorsukov: I did same )

alexpavlov96 (approved):

Issue #11699, opened 2022-05-19
[MU4 Issue] "Restore session?" dialog position is not center of screen


schoettl:
**Describe the bug**
The "restore session?" dialog after a MuseScore crash seems to be centered around the screen position 0,0. That is, only part of the right "Yes" button is visible. Other dialogs like Help/About and Open/Save dialogs are centered correctly.

**To Reproduce**
1. Open MuseScore and open a score file
2. Let MuseScore crash somehow (or kill the process with `killall -9 musescore` or so)
3. Check the position of the restore dialog.

**Expected behavior**
Should be centered on screen.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/1301843/169375269-968abd8c-5a95-4a0b-841d-fee135f93484.png)
This shows the top-left corner of my left-most primary X screen.

**Platform information**
- Arch Linux with Xmonad window manager. All dialogs are floating correctly (not tiled).
- I have multiple screens connected! Cannot test currently how it works with only one
- Latest dev version from today.

Tantacrul:
I'm marking this as a P3 for now because with multiple windows, it's a solvable problem (but definitely not desirable). However, if anyone can reproduce this for other cases, I'll raise the P level.

Thanks! Nice catch :)

cbjeukendrup:
@schoettl I'm unable to reproduce this issue myself, but it should probably be fixed in #11720. Could you please try the test builds from that PR? (They should be ready in 20 minutes from now.)

schoettl:
Thanks for your efforts!

PS: I checked out our PR and built myself. See PR.

cbjeukendrup:
@schoettl This issue was actually not really fixed, was it?

Anyway, I've a second attempt here: #13542
Would be nice if you could check it out!

PR #11720, opened 2022-05-23
Fix positioning dialogs when the main window isn't loaded yet

cbjeukendrup:
Resolves: #11699

### DIFF
--- a/src/framework/uicomponents/view/dialogview.cpp
+++ b/src/framework/uicomponents/view/dialogview.cpp
@@ -22,6 +22,7 @@
 
 #include "dialogview.h"
 
+#include <QScreen>
 #include <QWindow>
 
 #include "log.h"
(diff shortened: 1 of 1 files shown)


schoettl:
Thanks for this patch. The position has changed but now it's reproducibly here, when MuseScore is started on my middle screen:
grafik

Oh, and it's on the exact same *absolute* (Xinerama?) position, when MuseScore is started on the left or right screen. The dialog always appears on the middle screen and not at the center of the screen.

My xrandr output, if it helps is:
```
Screen 0: minimum 8 x 8, current 5520 x 1080, maximum 32767 x 32767
eDP1 connected 1920x1080+3600+0 (normal left inverted right x axis y axis) 310mm x 170mm
  …
DP1 disconnected (normal left inverted right x axis y axis)
DP2 disconnected (normal left inverted right x axis y axis)
DP2-1 disconnected (normal left inverted right x axis y axis)
DP2-2 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 530mm x 300mm
   …
DP2-3 connected 1680x1050+1920+0 (normal left inverted right x axis y axis) 490mm x 320mm
   …
HDMI1 disconnected (normal left inverted right x axis y axis)
HDMI2 disconnected (normal left inverted right x axis y axis)
VIRTUAL1 disconnected (normal left inverted right x axis y axis)
```

Eism:
The main window must be loaded before opening the restore session dialog
See [startupscenario](https://github.com/musescore/MuseScore/blob/master/src/appshell/internal/startupscenario.cpp#L84)

There may be a problem that we send the opening signal early in PopupView. 
I recommend to do the same with the open as with the close - send an open signal only when qt sends a signal.
See [PopupView::onHiden](https://github.com/musescore/MuseScore/blob/master/src/framework/uicomponents/view/popupview.cpp#L202)

But the solution needs to be thoroughly tested because it's a global change.

Issue #10681, opened 2022-02-27
[MU4 Issue] Application seems to lose/gain focus (highlight) while traversing the top text menu items


worldwideweary:
**Describe the bug**
At least on my Linux distro (Mint - Release: Una), the application title and the "V" icon (for max/minimize etc)  flashes while mousing through the file-type menu, (and the application icon in taskbar selects/deselects concurrently):

https://user-images.githubusercontent.com/7139517/155895463-12c22a07-0473-4836-b082-bb44a012a62b.mp4

It would seem that MuseScore is losing and regaining focus in the process.

cbjeukendrup:
Indeed, whenever a menu, popup or tooltip is opened, the main window looses focus. This should be solvable since for example Qt Creator's menus (which are also non-native) don't show this behaviour. (of course, 'under the hood', the popup window _should_ get focus, but it should not have a visual effect.)

llui85:
Possibly related to #10218?

tibequadorian:
I noticed that in a ~15 sec session where I heavily mouse through the panel, some functions in the debug output are more often called and have significantly more sum time:
```
Function                                                    Call time         Call count        Sum time
NavigationController::onActiveRequested                     0.169 ms          110               18.547 ms
UiActionsRegister::updateEnabledAll                         0.121 ms          112               13.575 ms
NavigationControl::setPanel                                 0.002 ms          1223              2.601 ms
NavigationController::doDeactivateSection                   0.020 ms          75                1.482 ms
NavigationController::doDeactivatePanel                     0.015 ms          80                1.175 ms
NavigationController::doDeactivateControl                   0.001 ms          747               0.873 ms
UiContextResolver::currentUiContext                         0.006 ms          113               0.709 ms
```
compared to an idle session of the same length:
```
Function                                                    Call time         Call count        Sum time
NavigationController::onActiveRequested                     0.143 ms          1                 0.143 ms
UiActionsRegister::updateEnabledAll                         0.086 ms          3                 0.259 ms
UiContextResolver::currentUiContext                         0.003 ms          4                 0.012 ms
```
I hope this helps.

Tantacrul:
> Indeed, whenever a menu, popup or tooltip is opened, the main window looses focus. This should be solvable since for example Qt Creator's menus (which are also non-native) don't show this behaviour. (of course, 'under the hood', the popup window _should_ get focus, but it should not have a visual effect.)

Is this is still an issue?

[...]

Issue #9056, opened 2021-09-07
[MU4 Issue] Tooltips appearing with undesirable animation


bkunda:
Tooltips are currently appearing with an undesirable animation, which appears as if scooting/swooping in from the right.

See demo video: https://www.dropbox.com/s/koft4o07wu472kc/tooltip-animation.mov?dl=0

Expected behaviour: tool tips should simply appear (without animated transitions), as per MU3.6.

MacBook Air M1, MacOS 11.5.2

HemantAntony:
This appears to be fixed in the latest master

boryskuzmenko:
Hi, @HemantAntony. Thanks for your fix!
The animation is removed but there's still some delay before a tip appearing in MU4. About 1 sec. If you see MU3 behavior there's no delay. Could you please fix this as well?

https://user-images.githubusercontent.com/101250347/166307067-4ac036fb-1dd6-43bd-bd9e-594a83ac3df0.mov

HemantAntony:
@boryskuzmenko I never made a fix, I just commented on whether this issue occurs or not. But I'll take the thanks anyway :)

#11504 removes the tooltip intervals for all buttons in MU4 (So not just the toolbar). Tell me if this is not wanted

boryskuzmenko:
@HemantAntony Works great now. Thanks!
**FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

boryskuzmenko:
🟡 **Partly FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

See the issue https://github.com/musescore/MuseScore/issues/11876 for more details

Issue #9055, opened 2021-09-07
[MU4 Issue] Dropdown animation transition is too slow


bkunda:
When moving from one dropdown menu to another, the slow transition time means that the first menu is still partially visible while the second menu appears above it. This produces a weird optical result. 

My preference would be for a cleaner transition, where one menu disappears (completely and quickly) before the next one appears. 

Probably easier to see in this demo video: https://www.dropbox.com/s/0zkv568558l2a8i/dropdown-menu-transition-slow.mov?dl=0

MacBook Air M1, MacOS 11.5.2

Tantacrul:
I have reprioritised this bug because it has bad interaction consequences.

In particular, if you are moving quickly, the context menu often doesn't keep up, so you have to go back and roll over options more slowly.

I'm not certain whether this is due to an animation or not. If it is, we should completely remove the animation. Options should pop out instantaneously.

boryskuzmenko:
🟡 **Partly FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

See the issue https://github.com/musescore/MuseScore/issues/11876 for more details

PR #11750, opened 2022-05-25
[MU4] Sub menu close animation fix

Eism:
Resolves: #9055
Resolves: #9056
Resolves: #10681
Resolves: #11860


bkunda:
@Eism this works like a dream on Mac ☺️. Fixed!

But on Windows, I still see an apparent delay when moving from one sub menu to another (particularly noticeable on longer menus):

https://www.dropbox.com/s/vvg5l1pel76gmrr/windows-menu-lag.mp4?dl=0

Tooltips in Windows experience the same problem:

https://www.dropbox.com/s/ei9y4go21dwqkeb/windows-tooltip-lag.mp4?dl=0

Eism:
@bkunda please recheck latest changes

bkunda:
@Eism this issue still occurs in Windows for me

Eism:
@bkunda 
I added animation for tooltips and menus. It should have solved your problem
Please recheck

bkunda:
> @bkunda
> I added animation for tooltips and menus. It should have solved your problem
> Please recheck

@Eism I'm afraid this makes it worse. There should be no need for any additional animation. When the user leaves one menu and moves to another one, the first sub-menu should immediately be replaced by the next one.

You had it perfect in a previous build for MacOS (see below):

https://user-images.githubusercontent.com/86290556/171169365-84fcb553-f34e-41f1-bd90-987c365d024e.mov

If this behaviour can just be replicated for all operating systems, that would be great.

Eism:
Okay, the new solution is ready

@bkunda please recheck 
The problem is that I can't reproduce the bug. My solutions are my guesses. I hope this will be the last one

bkunda:
Ok this is a lot better. 
I still get a tiny lag in Windows, but I am starting to wonder whether this is due to the very slow processing power of my machine.
At any rate, it feels much better in Windows now and it is perfect on my Mac. 
Thanks @Eism! This is actually a really important fix.

Tantacrul:
Wow. This is so much better (tested on MacOS 12)

boryskuzmenko:
@bkunda the issue is reproducible on Win11 and Linux. Are you sure it's OK and we can close this issue?

bkunda:
> 

@boryskuzmenko sorry can you please clarify: are you also still seeing a lag when switching between sub-menus in Windows and Linux? 

If so (and it's not just my slow machine), then we really do need to fix that. 

But maybe I've misunderstood your comment 🙂

bkunda:
> How serious is this lag?
> Can we fix this later? For example, it almost does not reproduce for me.
> This PR fixes a lot of problems with tooltips and menus, and I also optimized the solution from the technical side.

It's pretty minor for me on my Windows machine, although still noticeable. 

Borys's comment made me think he could still reproduce the lag in Windows and Linux, so I'm just hoping he can confirm if this is still the case. 

If other testers  can _almost_ not reproduce it (I.e. it's still there but not a real problem) then perhaps we can merge the present solution because it _is significantly better_, and then raise a new (lower-priority) issue for a future fix to perfect the sub-menus only in those operating systems that remain affected.

boryskuzmenko:
Ok, lets do that. I'll create another issue for that

RomanPudashkin (approved):

boryskuzmenko:
🟢 **Partly FIXED** 
See the issue https://github.com/musescore/MuseScore/issues/11876 created after closing this PR

Issue #10877, opened 2022-03-23
[MU4 Issue] Backwards/Forwards buttons (In 'General' / Properties panel) have no effect


MarcSabatella:
Edited by Martin:

What we need is the same options you find in most creation software. Basically four buttons that do the following:

**Forwards** (looks at the items in front of it and moves in front of the nearest item)
**To front** (looks at the items in front of it and moves in front of the item furthest to the front)
**Backwards** (looks at the items behind it and moves behind the nearest item)
**To back** (looks at the items behind it and moves behind the one furthest back)

**Design mockup**
<img width="292" alt="APPEARANCE_POPUP_SNAP_OFF_V2c" src="https://user-images.githubusercontent.com/86290556/171202958-e853820b-2cc8-4c76-b6b2-eacf85c4c180.png">

Link to Figma file (internal team only): https://www.figma.com/file/l0TQUrNFZmEfrZnfHhFDfb/Musescore-Notation-App-MASTER?node-id=1621%3A74746

DmitryArefiev:
@MarcSabatella @Tantacrul As I see, Backward/Forwards buttons are equivalent of Stacking order (Z) feature in MU3 (up/down controls). 
But in MU3 the user seeing the changes of actions in the numeric field. 
In MU4 we don't have that numeric field and user doesn't see the changes in Properties (only in score).

https://user-images.githubusercontent.com/90187801/170740314-2caf4171-87db-43ba-9a8c-89f8cdf93a69.mp4

So one of the fix can be putting a numeric value to Arrange section showing the object's layer in score.
The same as MU3 does in the stacking order field:


https://user-images.githubusercontent.com/90187801/170740905-bf8c3b0f-e0f9-44e3-846f-269052bd267f.mp4



(I guess we can delete extra zeros and use only xx or x format)

Tantacrul:
@DmitryArefiev: I think we need a more sensible interface for this. A textbox (like in MS3) is just ridiculous.

What we need is the same options you find in most creation software:

**Forwards** (looks at the items in front of it and moves above the closest one)
**To front**  (looks at the items in front of it and moves above the top one)

MarcSabatella:
[...]

PR #11760, opened 2022-05-25
Resolve #11253 - Add 'To Back' and 'To Front' buttons to 'Arrange'

### LINKED ISSUE #11253: 'Arrange' back/forward feature missing two options
**Describe the bug**
At the moment, the 'Arrange' feature allows users to move an item behind or in front of another item. However, standard practice is to also have options to arrange 'To front' and 'To back'.

To front: this would move the selected item in front of anything else in the score.
To back: this would move the selected item to the back of the score

image

This feature was never fully completed (by design) due to the transition from MS3 to MS4 - as this new inspector was originally created to live in V3.

OfekZeevi:
Resolves: [#11253](https://github.com/musescore/MuseScore/issues/11253)

*I've added a "To Back" and "To Front" buttons to the "Arrange" section of an item's "Appearance" settings. These would immediately push the selected items to the front or back of the score (in the z-axis).*

### DIFF
--- a/src/inspector/models/general/appearance/appearancesettingsmodel.cpp
+++ b/src/inspector/models/general/appearance/appearancesettingsmodel.cpp
@@ -111,12 +111,12 @@ std::vector<Ms::EngravingItem*> AppearanceSettingsModel::getAllElementsInPage()
     return page->elements();
 }
 
-void AppearanceSettingsModel::pushBackInOrder()
+void AppearanceSettingsModel::pushBackwardsInOrder()
 {
     m_arrangeOrder->setValue(m_arrangeOrder->value().toInt() - REARRANGE_ORDER_STEP);
 }
 
-void AppearanceSettingsModel::pushFrontInOrder()
+void AppearanceSettingsModel::pushForwardsInOrder()
 {
     m_arrangeOrder->setValue(m_arrangeOrder->value().toInt() + REARRANGE_ORDER_STEP);
 }
--- a/src/inspector/models/general/appearance/appearancesettingsmodel.cpp
+++ b/src/inspector/models/general/appearance/appearancesettingsmodel.cpp
@@ -103,6 +103,14 @@ void AppearanceSettingsModel::loadOffsets()
     });
 }
 
+std::vector<Ms::EngravingItem*> AppearanceSettingsModel::getAllElementsInPage()
+{
+    requestElements();
+    Ms::EngravingItem* element = m_elementList.first();
+    Ms::Page* page = toPage(element->findAncestor(Ms::ElementType::PAGE));
+    return page->elements();
+}
+
 void AppearanceSettingsModel::pushBackInOrder()
 {
     m_arrangeOrder->setValue(m_arrangeOrder->value().toInt() - REARRANGE_ORDER_STEP);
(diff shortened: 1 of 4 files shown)


Eism:
@OfekZeevi please change the destination branch from 4.0_alpha to master

DmitryArefiev:
> These would immediately push the selected items to the front or back of the score (in the z-axis).

@OfekZeevi Hi, found that with some elements **when clicking To front  you can't override value 'to be in front' of the second element.** 
When making text in front of notes by clicking To front it makes text in front. Now select the notes and click To front, it doesn't make notes in front of text. You need to select text and click To back.

https://user-images.githubusercontent.com/90187801/170698719-fe134378-16d4-4b00-8e4a-b2f8e68cd845.mp4

src/inspector/models/general/appearance/appearancesettingsmodel.cpp
```diff
@@ -103,16 +103,48 @@ void AppearanceSettingsModel::loadOffsets()
...
     m_arrangeOrder->setValue(m_arrangeOrder->value().toInt() - REARRANGE_ORDER_STEP);
 }
 
-void AppearanceSettingsModel::pushFrontInOrder()
+void AppearanceSettingsModel::pushForwardsInOrder()
 {
     m_arrangeOrder->setValue(m_arrangeOrder->value().toInt() + REARRANGE_ORDER_STEP);
 }
 
+void AppearanceSettingsModel::pushToBackInOrder()
```
MarcSabatella, line 124: FWIW, this seems to be the exact same approach that would also be needed to fix the existing Forwards/Backwards buttons, which basically do almost nothing now because they only move by REARRANGE_ORDER_STEP rather than finding the previous/next element and pushing past that.  So currently they require you to click Forwards literally dozens of times before you see anything happen.

I mention this here because ideally I think both of these issues should be addressed to together so the code is designed well to handle both commands with a minimum of duplication.

OfekZeevi:
> > These would immediately push the selected items to the front or back of the score (in the z-axis).
> 
> @OfekZeevi Hi, found that with some elements **when clicking To front you can't override value 'to be in front' of the second element.** When making text in front of notes by clicking To front it makes text in front. Now select the notes and click To front, it doesn't make notes in front of text. You need to select text and click To back.
> bandicam.2022-05-27.15-11-12-809.mp4

Thanks! I'm looking into that. It seems like maybe there's a bigger issue here with the z-value of text elements, but I'm not yet sure.

src/inspector/models/general/appearance/appearancesettingsmodel.cpp
OfekZeevi: Thanks, that's an interesting point. I'll look into that to see if there's an elegant solution for both problems.

OfekZeevi: I've updated the pull request to support that feature as well :)

OfekZeevi:
> > These would immediately push the selected items to the front or back of the score (in the z-axis).
> 
> @OfekZeevi Hi, found that with some elements **when clicking To front you can't override value 'to be in front' of the second element.** When making text in front of notes by clicking To front it makes text in front. Now select the notes and click To front, it doesn't make notes in front of text. You need to select text and click To back.
> bandicam.2022-05-27.15-11-12-809.mp4

I've figured it out! When you select a "note" element, you're not actually selecting the note but rather an element of type "Chord" that contains it. The chord isn't actually printed on the page, it's just a logical container for the notes within. For this reason, when playing with the front/back buttons, the z-value of the chord changes but that doesn't matter because the notes themselves remain at their original z-value. This explains why you can't move the notes by pressing those buttons (like you showed in the video). It's important to mention that this means the forwards/backwards buttons never affected notes, even before the changes I've made here.
I would like to point out that maybe this is intentional - from digging through the code, it seems like the order in which notes are printed matters, e.g. so that note-heads will be drawn after their note-flags and such. This makes me think that perhaps the fact that you're not able to change the z-value of notes is a feature and not a bug.
Either way, I think this issue requires a separate ticket so we can discuss it properly, and again - it isn't actually affected by my changes. So I decided to leave this issue unresolved for now.
Please let me know if you disagree :)

DmitryArefiev:
> > > These would immediately push the selected items to the front or back of the score (in the z-axis).
> > 
> > 
> > @OfekZeevi Hi, found that with some elements **when clicking To front you can't override value 'to be in front' of the second element.** When making text in front of notes by clicking To front it makes text in front. Now select the notes and click To front, it doesn't make notes in front of text. You need to select text and click To back.
> > bandicam.2022-05-27.15-11-12-809.mp4
> 
> I've figured it out! When you select a "note" element, you're not actually selecting the note but rather an element of type "Chord" that contains it. The chord isn't actually printed on the page, it's just a logical container for the notes within. For this reason, when playing with the front/back buttons, the z-value of the chord changes but that doesn't matter because the notes themselves remain at their original z-value. This explains why you can't move the notes by pressing those buttons (like you showed in the video). It's important to mention that this means the forwards/backwards buttons never affected notes, even before the changes I've made here. I would like to point out that maybe this is intentional - from digging through the code, it seems like the order in which notes are printed matters, e.g. so that note-heads will be drawn after their note-flags and such. This makes me think that perhaps the fact that you're not able to change the z-value of notes is a feature and not a bug. Either way, I think this issue requires a separate ticket so we can discuss it properly, and again - it isn't actually affected by my changes. So I decided to leave this issue unresolved for now. Please let me know if you disagree :)

@OfekZeevi  Thanks for explanation! Agree not to fix that in this PR.

BTW #10877 also looks much better now in this PR. But the position of buttons should be re-arranged now (to the new design). It's described in the description in #10877. Maybe you want to implement this as well in the current PR? :)

DmitryArefiev:
@Eism Can you review please :)

Eism (approved):

Issue #11759, opened 2022-05-25
[MU4 Task] Thumbnail icon to be created for GP files


boryskuzmenko:
**Task description**
I believe we should have to separate GP files from MU files on the Recent board. To do that specific icon should be created and implemented.
Also will be good to use this icon as a file icon for GP files in case when there's no Guitar Pro installed (just thoughts)
MU4 file icon as an example:

![Capture 2022-05-25_08-48-05_PM](https://user-images.githubusercontent.com/101250347/170330665-a8d9ea5a-565f-468a-b3b6-eed309a14dd5.png)

Tantacrul:
@jessjwilliamson - I know I said not to... but can we make a GP image>? :)

I'd suggest a guitar silhouette or plectrum symbol or something like that.

jessjwilliamson:
> 

hehe.

Of course :)

DmitryArefiev:
Fixed in #11773

Issue #11757, opened 2022-05-25
[MU4 Issue] File icons look pixelated on Windows/Linux (FullHD, 100% scaling)


DmitryArefiev:
**Windows10, FullHD, 100%**
![Capture](https://user-images.githubusercontent.com/90187801/170513152-e7db5d36-4fb9-40a2-84fa-f34817f955da.PNG)

Windows10, FullHD, 150%
![Capture2](https://user-images.githubusercontent.com/90187801/170511763-eb5009aa-cee5-4c04-a47c-b4103c939338.PNG)

Mac12, Retina display
![Screenshot 2022-05-26 at 5 44 17 PM](https://user-images.githubusercontent.com/90187801/170512883-24c3118a-a961-4648-bf71-d24e0dde2acc.png)

**Platform information**
 Windows, Linux - FullHD, 100

HemantAntony:
I can reproduce on my laptop. The icon is too sharp

DmitryArefiev:
Fixed in #11773

PR #11773, opened 2022-05-26
Recent Scores: Add Guitar Pro file icon and fix pixelated icons on low-res screens

cbjeukendrup:
Resolves: #11757 
(I hope that this fix works, because the screen of my non-retina laptop is so poor that I can barely see the difference! But it seems slightly better.)

Resolves: #11759
Schermafbeelding 2022-05-26 om 19 06 32


RomanPudashkin (approved):

DmitryArefiev:
Tested #11757 #11759 on Win10, Mac12, LinuxMint 20.2 - **FIXED**     

> Resolves: https://github.com/musescore/MuseScore/issues/11757
> (I hope that this fix works, because the screen of my non-retina laptop is so poor that I can barely see the difference! But it seems slightly better.)

@cbjeukendrup Yeah! Now it's better. Thanks!

Issue #11616, opened 2022-05-12
[MU4 Issue] Some issues when openning a MU3 mscx scores 


DmitryArefiev:
**Describe the bug**

From https://github.com/musescore/MuseScore/pull/10660#issuecomment-1117362634

1. When opening an old mscx MU3 scores from one folder and re-saving in a new version of mscx format, 
**MU4 creates folders and setting files inside that common folder but for all scores together and not separately for each score**. The issue can be solved if we trigger **Save As** dialog in that case (after clicking Save) and user can choose a regular mscz format or experimental uncompressed mscx format (with creating a new folder for each MU3 file)

https://user-images.githubusercontent.com/90187801/166698579-6f26674b-f7f5-49e8-99fa-33099681cadc.mp4

2. When opening an old mscx MU3 score and triggering Save As dialog, **MU4 takes the name of the folder and not the initial file name**

https://user-images.githubusercontent.com/90187801/166699049-ab416cc9-cdb9-465c-9ad4-ab3b126588c8.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
#10556
#11615

DmitryArefiev:
Fixed in #11782

Issue #11615, opened 2022-05-12
When saving uncompressed MuseScore folder format, MU doesn't remove an extension and saves the score as mscz file on Linux (occurs on Windows/Mac if file extensions are not hidden)


DmitryArefiev:
**To Reproduce**
1. Create a score 
2. Add some notes
3. Save
4. In Save score dialog, select Uncompressed MuseScore folder format (notice MU shows mscz extension)
5. Click OK - **MU saves the score in mscz format**

![uncompr](https://user-images.githubusercontent.com/90187801/168086064-c91df167-eb86-4293-a928-09a76e580b51.png)

**Workaround**:
**Remove the extension manually** when saving Uncompressed MuseScore folder format 

**Platform information**
Linux Mint 20.2

**Additional context**
See #10660

Also, the bug can be related to #10836

DmitryArefiev:
> Should be rechecked after #10836 fix

Nope.. still occurs after #10836

DmitryArefiev:
I think can be decreased to P2

bkunda:
@DmitryArefiev pls test for reproducibility, and whether either PR above fixes it

nukeop:
I'm still seeing this on Mac.

Issue #10836, opened 2022-03-18
[MU4 Issue] Wrong explorer in latest build on Linux (Qt's dialog instead of Native dialog)


DmitryArefiev:
Wrong explorer in latest build on Linux when using Save as or Open dialogs

**Actual:**
![Снимок экрана от 2022-03-17 18-27-50](https://user-images.githubusercontent.com/90187801/158977876-f13cfc1b-0281-406b-bf57-88698f071c24.png)

**Expected:**
![Снимок экрана от 2022-03-17 18-26-11](https://user-images.githubusercontent.com/90187801/158977856-da886826-9042-46f6-986f-2282f4ba43b7.png)

**Platform information**
 Linux Mint 20.2

**Additional dialog**
- Qt's file dialog is **buggy** (e.g. it doesn't save score if it was entered without extension and I can't open a score from a flash drive)

- In the "wrong", unexpected save/open dialog, all key presses are consumed by the main window. E.g. Return inserts a line break in the score instead of opening a file (see comment below)

Jojo-Schmitz:
Is that Native file dialog vs. Qt's file dialogs?

DmitryArefiev:
Yes, looks like it is. 

> MuseScore 3 has an (advanced) option to change between them

Thanks! I've founded it in MU3 (and it's turned on by default) but it's missing in MU4
![Capture](https://user-images.githubusercontent.com/90187801/158992604-a2430412-3563-4082-a528-c6873e506c22.PNG)

Eism:
I tested a little and noticed that the dialog is different when building on my laptop(Ubuntu) and on CI.

So, first we need to understand the reason for this behavior.

Laptop
![image](https://user-images.githubusercontent.com/10116828/158993261-3de56e24-49ac-4b33-8ff6-f0c6c5ea3c36.png)

CI
![image](https://user-images.githubusercontent.com/10116828/158993355-9b4c1310-560f-4229-991a-b39bdd30952a.png)

schoettl:
Not sure if I should create a new issue for that:
In the "wrong", unexpected save/open dialog, all key presses are consumed by the main window. E.g. Return inserts a line break in the score instead of opening a file.

DmitryArefiev:
> Not sure if I should create a new issue for that:
> In the "wrong", unexpected save/open dialog, all key presses are consumed by the main window. E.g. Return inserts a line break in the score instead of opening a file. (Today's master branch on Linux.)

[...]

PR #11782, opened 2022-05-27
Fixes for opening/saving MSCX scores and scores from older versions

cbjeukendrup:
Resolves: #10556
~ResoIves: #11615 (Couldn't verify, but I did the same for Linux as I did to fix this problem on Windows, so it might work)~
Resolves: #11616

### DIFF
--- a/src/plugins/api/qmlpluginapi.cpp
+++ b/src/plugins/api/qmlpluginapi.cpp
@@ -134,7 +134,8 @@ Score* PluginAPI::readScore(const QString& name, bool noninteractive)
     mu::engraving::MasterScore* score = msc()->openScore(name, !noninteractive);
     if (score) {
         if (noninteractive) {
-            score->setNewlyCreated(false);
+            // TODO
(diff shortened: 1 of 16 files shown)


src/plugins/api/qmlpluginapi.cpp
```diff
@@ -134,7 +134,8 @@ Score* PluginAPI::readScore(const QString& name, bool noninteractive)
     Ms::MasterScore* score = msc()->openScore(name, !noninteractive);
     if (score) {
         if (noninteractive) {
-            score->setNewlyCreated(false);
+            // TODO
```
Eism, line 137: what are the plans for this todo?

Eism (approved):

DmitryArefiev:
Tested #10556 #11616 on Win10, Mac12, LinuxMint 20.2 - **FIXED**      

#11615 **still occurs**. Let's recheck it after #10836 is fixed

PR #11784, opened 2022-05-27
Fix some clang compiler warnings

cbjeukendrup:
About missing override and about copying of loop variable


igorkorsukov (approved):

cbjeukendrup:
One more fix coming...

PR #11786, opened 2022-05-27
[MU4] Fix compiler warnings

Jojo-Schmitz:
* Fix QtCreator/MinGW compiler (?) warnings reg. unrecognized escape sequence (since #11556)
* Fix MSVC compiler warnings reg. conversion from 'size_t' to 'int', possible loss of data (C4267)
 (since #11822, #11825, #11849)

### DIFF
--- a/share/icons/windows_icons.rc
+++ b/share/icons/windows_icons.rc
@@ -1,3 +1,3 @@
-IDI_ICON1   ICON  DISCARDABLE "AppIcon\MS4_AppIcon.ico"
-IDI_ICON2   ICON  DISCARDABLE "MsczIcon\MS4_MsczIcon.ico"
-IDI_ICON3   ICON  DISCARDABLE "MscxIcon\MS4_MscxIcon.ico"
+IDI_ICON1   ICON  DISCARDABLE "AppIcon\\MS4_AppIcon.ico"
+IDI_ICON2   ICON  DISCARDABLE "MsczIcon\\MS4_MsczIcon.ico"
+IDI_ICON3   ICON  DISCARDABLE "MscxIcon\\MS4_MscxIcon.ico"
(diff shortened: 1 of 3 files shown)


share/icons/windows_icons.rc
```diff
@@ -1,3 +1,3 @@
-IDI_ICON1   ICON  DISCARDABLE "AppIcon\MS4_AppIcon.ico"
-IDI_ICON2   ICON  DISCARDABLE "MsczIcon\MS4_MsczIcon.ico"
-IDI_ICON3   ICON  DISCARDABLE "MscxIcon\MS4_MscxIcon.ico"
+IDI_ICON1   ICON  DISCARDABLE "AppIcon\\MS4_AppIcon.ico"
+IDI_ICON2   ICON  DISCARDABLE "MsczIcon\\MS4_MsczIcon.ico"
+IDI_ICON3   ICON  DISCARDABLE "MscxIcon\\MS4_MscxIcon.ico"
```
cbjeukendrup, line 3: I'm quite unsure about this one, because on MSVC it works correctly with a single backslash (I tested the builds from CI). We would need to re-test it if we change this. It surprises me that MinGW says anything about this file at all. It should be handled by the RC compiler, which compiles it to a .o file, and _that_ should be sent to MinGW for linking with the other sources. (See src/main/CMakeLists.txt.)

Jojo-Schmitz: Try the artifact?

Jojo-Schmitz: If it is a path, maybe a `/` instead would work too?

cbjeukendrup: I tried some things and searched a bit on the internet, and probably your change is correct. I tried editing the rc file using Visual Studio, which has a kind of graphical editor for these files, and its output uses `\\`. So that must be right.

PR #11787, opened 2022-05-27
Add new instrument: Bell Tree

shoogle:
Add the Bell Tree instrument to MuseScore and assign it a playback sound. ~I picked Glockenspiel as the closest sound in General MIDI but~ I created a new MPE sound ID for the Bell Tree specifically.

~https://en.wikipedia.org/wiki/Bell_tree
https://www.youtube.com/watch?v=rhStPnEcXrk (I used C4 to C8 as the pitch range based on this video, but I'm not sure how accurate that is.)~

The changes to instruments.xml need to be reviewed by @oktophonie, @Tantacrul and @lengthwave. Once those are approved, the playback changes will need to be reviewed by the dev team before this is merged. If necessary, I could drop the playback commit and leave it to someone more familiar with the new sound system.

__Edit:__ The consensus was that the Bell Tree and Mark Tree should be added as _unpitched_ percussion instruments, so I updated the PR to do this. See discussion below.


Jojo-Schmitz:
Isn't this related to https://musescore.org/en/node/264854 ?

shoogle:
@Jojo-Schmitz, yet it is. It seems that there is a MIDI sound for the bell tree at drum note 84, but if we use that sound then you would always get a fixed-rate glissando (like [this](https://youtu.be/rhStPnEcXrk?t=29)). It wouldn't be possible to play notes individually (like [this](https://youtu.be/rhStPnEcXrk?t=19)), or a partial glissando (like [this](https://youtu.be/rhStPnEcXrk?t=33)), upwards glissando, or glissando at different speeds.

Perhaps I misunderstood the requirement though and the bell tree should be treated as an unpitched percussion instrument rather than as a pitched one? (I was told it uses a 5 line staff.)

Tantacrul:
@shoogle. My mistake. I noticed that we have recordings of single hits for the Bell Tree, so I thought it might make sense to make it a 5 line instrument. However, that is quite non-standard. I'm just being a silly billy.

It should be single line. And the initial hit will produce a glissando (up). You can specify gliss up and gliss down and that's about it. Same with the Mark Tree, incidentally. And we actually need that instrument definition too!

@oktophonie and @lengthwave - let me know if you have thoughts.

Jojo-Schmitz:
As per the links above bell tree and mark tree are the same thing, aren't they?

_Bell tree is the more generic name for the specific instrument known as the "Mark tree". See https://en.wikipedia.org/wiki/Mark_tree._

lengthwave:
Bell tree and mark tree are different instruments, but commonly confused.

https://en.wikipedia.org/wiki/Bell_tree

Jojo-Schmitz:
BTW, "Bell Tree" would (or might) be translated to "Schellenbaum" in German, but that's an entirely different thing, see https://de.wikipedia.org/wiki/Schellenbaum

shoogle:
I've fixed the Bell Tree and added the Mark Tree as well. @oktophonie, @lengthwave, please download the CI build and let me know if the instruments have been set up as you expect.

Tantacrul:
Hey @oktophonie - can you let me know if these look OK to you?

Tantacrul:
Forget that. I just tested it. Seems to be working as expected. Thanks @shoogle!

@RomanPudashkin / @Eism / @cbjeukendrup - can you merge this if you are happy with it?

Also, @shoogle - these two instruments don't have definitions in our XML. Can you add those? I'll ask Vasily to create MPE IDs.

cbjeukendrup:
This looks fine to me, but I can't say much about the correctness of the sound IDs etc.

Also, a rebase is needed.

oktophonie:
I've seen the Bell Tree, at least, on a 5-line stave (with perc clef) more often than not, e.g.

image
(MacMillan)

image
(Adams)

image
(Penderecki)

But see no issue in defining it as a one-line stave (which is certainly encountered in the wild too) by default, which also makes playback mapping more straightforward. Users can reconfigure it as they need.

oktophonie (approved):

shoogle:
> I've seen the Bell Tree, at least, on a 5-line stave [...] But see no issue in defining it as a one-line stave [...] which also makes playback mapping more straightforward

Indeed. Our SoundFont (and the Roland GS MIDI standard) only has a playback sound on one note for the Bell Tree, so even if we used a 5 line staff for it, users would still only be able to add one "pitch" to that staff with the right playback sound. Playback would be wrong for all the other "pitches", so it's much better to use a single line staff that only allows the correct note to be added.

src/framework/mpe/soundid.h
```diff
@@ -193,11 +193,13 @@ enum class SoundId
     Tablas,
     Timbales,
     Anvil,
+    BellTree,
```
cbjeukendrup, line 196: You will also need to add these below in `static const std::unordered_map<SoundId, String> ID_STRINGS`.

vpereverzev:
@shoogle rebase this branch one more time pls. Going to merge that one

PR #11788, opened 2022-05-27
Grace note beams commit by @Nick-Mazuk

asattely:
Do not commit yet, this is a manual rebase of the work that Nick had done for grace note beams.


Nick-Mazuk:
Hey @asattely, FYI if you don't want someone to merge a PR, you can set it as Draft. Then if someone tries to merge it, GitHub will make it very clear that it's not ready to be merged yet. All done for you!

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests

oktophonie (approved):

oktophonie (approved):

asattely:
It won't let me update the main comment of the PR, so:

Beams, and specifically, stem shortening within beams, needed to be looked at much more closely. Most of the work for this PR was done by Nick, but needed to be completed and the scope creeped up bit by bit.

- Grace note stems are now properly shortened. They are computed as full-length stems and then carefully scaled down by magnitude and look much better now.
- Various fixing of the way that stems are shortened when they are outside the staff, which has some differences between grace notes and regular ones
- Stems not on the end of beams (i.e. in 3+ note beams) now are able to be shortened down to a minimum stem length rather than the full default length for the stem pushing the beam away.
- Slash noteheads (and rhythmic slashes) above or below the staff with stems pointing away from the staff now have proper stem shortening applied to them. This was mostly a problem with slashed grace notes or rhythmic slashes.

cbjeukendrup:
Rebase needed (probably since #12163)

asattely:
> Rebase needed (probably since #12163)

Yup, all the changes of qreal to double required me to rebase all of my changes that included qreal in them!

src/engraving/libmscore/chord.cpp
```diff
@@ -1425,34 +1425,37 @@ int Chord::calcMinStemLength()
...
-        if (minStemLength < cutout) {
-            minStemLength = hookOffset;
-        } else {
-            minStemLength += hookOffset - cutout;
+        if (_hook) {
+            double smuflAnchor = _hook->smuflAnchor().y() * (_up ? 1 : -1);
+            int hookOffset = (_hook->height() + smuflAnchor) / _spatium * 4 - 2;
+            // TODO: when the SMuFL metadata includes a cutout for flags, replace this with that metadata
+            // https://github.com/w3c/smufl/issues/203
+            int cutout = up() ? 4 * _relativeMag : 6 * _relativeMag;
```
cbjeukendrup, line 1434: Are you sure that `cutout` must be an int and not a double? Or should it maybe be rounded? (since `_relativeMag` is a double)

src/engraving/libmscore/chord.cpp
```diff
@@ -1425,34 +1425,37 @@ int Chord::calcMinStemLength()
...
+            if (beams() >= 2) {
+                cutout -= 2;
```
cbjeukendrup, line 1436: Is it correct that this is _not_ multiplied by `_relativeMag`?

src/engraving/libmscore/beam.cpp
```diff
@@ -1031,23 +979,44 @@ void Beam::offsetBeamToRemoveCollisions(const std::vector<ChordRest*> chordRests
...
     double endY = (isStartDictator ? pointer : dictator) * spatium() / 4 + tolerance;
+    // using <= or >= for doubles is not a very good plan because 0 != 0.000000000001
+    auto fuzzyCompare = [](double a, double b) {
+        double diff = a - b;
+        if (abs(diff) < 0.001) {
+            return 0;
+        } else {
+            return diff < 0 ? -1 : 1;
+        }
+    };
```
cbjeukendrup, line 993: Maybe you can replace this with `RealIsEqualOrLess` and `RealIsEqualOrMore` from `realfn.h`?

src/engraving/libmscore/beam.cpp
```diff
@@ -262,7 +262,7 @@ void Beam::draw(mu::draw::Painter* painter) const
     if (_beamSegments.size() > 1 && d > M_PI / 6.0) {
         d = M_PI / 6.0;
     }
-    double ww = lw2 / sin(M_PI_2 - atan(d));
+    double ww = (_beamWidth / 2.0) / sin(M_PI_2 - atan(d));
```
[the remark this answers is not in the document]
asattely: yes it certainly can! Missed that one, thanks.

src/engraving/libmscore/chord.h
```diff
@@ -166,6 +167,8 @@ class Chord final : public ChordRest
     void setScore(Score* s) override;
     double chordMag() const;
     double mag() const override;
+    void calcRelativeMag();
```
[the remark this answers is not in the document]
asattely: It definitely could be!

src/engraving/libmscore/chord.cpp
asattely: Yeah. The calculations are all integers because they represent quantities of 0.25sp, so 1 = 0.25sp, 2 = 0.5, etc.

The fact that all of the calculations need integers and `_relativeMag` was a double was one of the biggest challenges with this one. This line can be changed to `floor(...)` instead of just casting to integer (though they're positive anyway). I'll change it to be explicitly floored but cutout not being a double would just push the issue further down the line when it is subtracted from `hookOffset` or added to `minStemLength` or whatever.

src/engraving/libmscore/chord.cpp
asattely: Ah, same explanation as above on this one

src/engraving/libmscore/beam.cpp
asattely: Oh, cool, didn't know those were in there. TIL

PR #11795, opened 2022-05-28
Fix #11776: Playback panel shows wrong tempo

### LINKED ISSUE #11776: [MU4 Issue] Play panel shows incorrect tempo
**Describe the bug**
When adding "Quarter note = 80 BPM" element from the palette, the play panel shows "♩= 79" in nightly build

**To Reproduce**
1. Create new score
2. Add "Quarter note = 80 BPM" element from tempo palette to the score
3. See play panel

**Expected behavior**
Play panel should show "♩= 80"

**Screenshots**
スクリーンショット (19)

**Platform information**
 - OS: Windows 11

**Additional context**
Doesn't occur in Private Alpha

HemantAntony:
Resolves: #11776

### DIFF
--- a/src/notation/internal/notationplayback.cpp
+++ b/src/notation/internal/notationplayback.cpp
@@ -299,7 +299,7 @@ const Tempo& NotationPlayback::tempo(tick_t tick) const
         return empty;
     }
 
-    m_currentTempo.valueBpm = static_cast<int>(score()->tempomap()->tempo(tick).toBPM().val);
+    m_currentTempo.valueBpm = static_cast<int>(std::round(score()->tempomap()->tempo(tick).toBPM().val));
 
     return m_currentTempo;
 }


src/notation/internal/notationplayback.cpp
```diff
@@ -299,7 +299,7 @@ const Tempo& NotationPlayback::tempo(tick_t tick) const
         return empty;
     }
 
-    m_currentTempo.valueBpm = static_cast<int>(score()->tempomap()->tempo(tick).toBPM().val);
+    m_currentTempo.valueBpm = std::round(score()->tempomap()->tempo(tick).toBPM().val);
```
Jojo-Schmitz, line 302: but `std::round()` does return a `double`, so that cast to `int` would still be needed (as `valueBpm` is an `int`), for the compiler not to warn, as far as I can tell?

HemantAntony: Yeah that's correct. I realized that after creating the PR and forgot to push the changes. Thanks! :)

Jojo-Schmitz: Strange enough neither MSVC nor MinGW do warn about it?!

HemantAntony: Yeah, that's probably correct. I figured it out because of a hunch and not because of a warning

Tantacrul:
Tested. The tempo display problem is fixed. Thanks! @HemantAntony

RomanPudashkin (approved):

PR #11799, opened 2022-05-29
Fixes for non-unity build

madebr:
Resolves: 

These patches fixes compile error when building MuseScore with `-DBUILD_UNITY=OFF`

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -34,6 +34,7 @@
 #include "engraving/engravingerrors.h"
 #include "engraving/style/defaultstyle.h"
 
+#include "iprojectautosaver.h"
 #include "notation/notationerrors.h"
 #include "projectaudiosettings.h"
 #include "projectfileinfoprovider.h"
--- a/src/framework/uicomponents/view/internal/popupviewclosecontroller.cpp
+++ b/src/framework/uicomponents/view/internal/popupviewclosecontroller.cpp
@@ -23,6 +23,7 @@
 #include "popupviewclosecontroller.h"
 
 #include <QApplication>
+#include <QWindow>
 
 using namespace mu::uicomponents;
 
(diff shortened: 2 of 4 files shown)


jeetee:
It looks like you've also (accidentally) changed the line endings on [popupviewclosecontroller.cpp](https://github.com/musescore/MuseScore/pull/11799/files#diff-787088af1942da1c9b4ec025e76a1a5b5bb2209b0f2cbf6486f97023f902a687) which results in a full file changed diff

madebr:
> It looks like you've also (accidentally) changed the line endings on [popupviewclosecontroller.cpp](https://github.com/musescore/MuseScore/pull/11799/files#diff-787088af1942da1c9b4ec025e76a1a5b5bb2209b0f2cbf6486f97023f902a687) which results in a full file changed diff

Oops, fixed this.
My editor was confused with the mixing of line-endings.
The cpp sources are CRLF, and the cmake scripts are LF.

May I suggest to add an [editorconfig config file](https://editorconfig.org/) to the root of the folder?

src/project/internal/notationproject.cpp
```diff
@@ -37,6 +37,7 @@
 #include "notation/notationerrors.h"
 #include "projectaudiosettings.h"
 #include "projectfileinfoprovider.h"
+#include "project/iprojectautosaver.h"
```
cbjeukendrup, line 40: You should not need `project/` here, because `project` is an include directory for the target of which this file is part.

madebr: Removed the prefix.

Jojo-Schmitz:
Shpuldn't just one commit be enough for this?

cbjeukendrup:
@madebr I believe (almost) everything in the repo is/should be LF. You could experiment with the `core.autocrlf` git setting, so that files will always be converted to LF when committing.

madebr:
> Shouldn't just one commit be enough for this?

This repo does not squash pr's?

cbjeukendrup:
> This repo does not squash pr's?

Occasionally, but generally we create merge commits.

Eism (approved):

Issue #11834, opened 2022-05-31
[MU4 Issue] unable to unset subscript / superscript from text (regression)


sammik:
**Describe the bug**
Once subscript or superscript is set to text, it is impossible to unset it

**To Reproduce**
1. Select text
2. Click Properties -> More -> Subscript (or Superscript)
3. Unable to unset it

**Expected behavior**
Clicking button second time unset property 

**Screenshots**

https://user-images.githubusercontent.com/1646034/171291085-9f8c67c6-a256-4cf6-ba0d-7c5fedc2ef2d.mp4

**Platform information**
 - OS: [tested on Linux]

**Additional context**
Only MU4 bug
In MS 3.6.2 it works as expected

PR #11835, opened 2022-05-31
Text inspector: fix turning off subscript/superscript

cbjeukendrup:
Resolves: #11834

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/text/TextSettings.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/text/TextSettings.qml
@@ -88,7 +88,7 @@ Column {
                 iconCode: modelData["iconRole"]
                 checked: root.model && !root.model.textScriptAlignment.isUndefined ? root.model.textScriptAlignment.value === modelData["typeRole"]
                                                                                    : false
-                onToggled: {
+                onClicked: {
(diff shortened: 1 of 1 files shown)


src/inspector/view/qml/MuseScore/Inspector/text/TextSettings.qml
```diff
@@ -88,7 +88,7 @@ Column {
                 iconCode: modelData["iconRole"]
                 checked: root.model && !root.model.textScriptAlignment.isUndefined ? root.model.textScriptAlignment.value === modelData["typeRole"]
                                                                                    : false
-                onToggled: {
+                onClicked: {
```
RomanPudashkin, line 91: Could you please explain why this helps?

cbjeukendrup: I found that the `onToggled` signal is not emitted at all if the user clicks the radio button while it is already checked (because normally, the radio button would not be toggled in that case but just stay checked, so in general no need for a `toggled` signal).
But the `clicked` signal _is_ emitted, every time that the user clicks the button, also when it is already checked. That's exactly what we need in this case.

RomanPudashkin (approved):

Issue #11606, opened 2022-05-11
[MU4 Issue] TAB. Duration stems are rendered in wrong way after Standard staff type changing to 6str-tab Common


boryskuzmenko:
**To Reproduce**
1. Open any score with standard notation staff
2. Change the staff type to 6str.-tab Common
3. Pay attention to the beams on the staff

**Screenshots**
![Capture 2022-05-11_06-51-17_PM](https://user-images.githubusercontent.com/101250347/167905063-eca71143-b7b3-4e72-9c6c-9f9491dba843.png)

https://user-images.githubusercontent.com/101250347/167905101-f05881ed-3b7e-452b-9a4e-cd12eb8e07cf.mov

**Expected behavior**
Duration stems are rendered according to the 6str-tab Common staff type design

![Capture 2022-05-11_07-55-36_PM](https://user-images.githubusercontent.com/101250347/167905400-3ab5492d-ee78-4ee3-98db-dab848930690.png)

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

**Additional context**
Related to #11092 changes

alexpavlov96:
Started to reproduce after this PR
https://github.com/musescore/MuseScore/pull/11092

asattely:
@boryskuzmenko Could you include the score file you are using? I'm having difficulty reproducing this one with the https://github.com/musescore/MuseScore/pull/11857 branch. It may be something in the score itself that could be making the layout system go crazy.

boryskuzmenko:
@asattely it was reproducible on any score before, but I can't reproduce it now. Seems like it's fixed 
Could you, please, also take a look at this issue if it has similar bug with beam? https://github.com/musescore/MuseScore/issues/12144

asattely:
> @asattely I don't know if it's the same bug but looks like:
> 
>  Capture.2022-07-11_06-21-05_PM.mov 
> reproduces on any group of notes with beam
It's not the same bug, and I tried to reproduce it but couldn't, there is probably some other stuff going on in the score file you are using. If it's still a problem then open a new issue for it, and if you would, please include your test file in the issue.

boryskuzmenko:
> It's not the same bug, and I tried to reproduce it but couldn't, there is probably some other stuff going on in the score file you are using. If it's still a problem then open a new issue for it, and if you would, please include your test file in the issue.

Separate issue created https://github.com/musescore/MuseScore/issues/12471

boryskuzmenko:
[...]

PR #11857, opened 2022-06-01
[MU4] Fix incorrect placement of beams when displayed beside tab staff

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/11606

Beams on 'common' tab staves have different offsets than ones from pitched staves. The offset needed for pitched staves are not necessary for tab staves.


boryskuzmenko (commented):
@asattely Looks like this PR doesn't fix the issue. Please, take a look

PR #11862, opened 2022-06-02
Added use AsciiString for write xml

igorkorsukov:


src/engraving/utests/utils/scorerw.cpp
```diff
@@ -102,6 +102,9 @@ EngravingItem* ScoreRW::writeReadElement(EngravingItem* element)
     // read element
     //
 
+    std::cout << buffer.data().constData() << "\n";
```
RomanPudashkin, line 105: Do we need this?

RomanPudashkin (approved):

src/engraving/utests/utils/scorerw.cpp
igorkorsukov: removed, thanks!

Issue #10812, opened 2022-03-15
[MU4 Task] Implement selected plugins to be pre-installed with MS4


bkunda:
**Task description**
Since plugins will need to be either modified or rebuilt for compatibility with MS4, we have selected a small handful from the existing set available on [musescore.org](https://musescore.org/en/plugins) that can be packaged together with the download of MS4. 

These plugins are considered to be already reasonably well designed and potentially offer real benefit to a considerable number of users. This list can and will of course grow, but bearing in mind available technical resources and the project timeline, we've narrowed these plugins down to the following list:

(Listed in order of priority for implementation):

**File management**
[ABC_ImpEx](https://musescore.org/en/project/abcimpex)
[Batch Convert](https://musescore.org/project/batch-convert)

**Composing/arranging tools**
[Add and remove courtesy accidentals](https://musescore.org/project/add-and-remove-courtesy-accidentals)
[Mirror Intervals](https://musescore.org/en/project/mirror-intervals)
[New Retrograde](https://musescore.org/en/project/new-retrograde)
[Note Names](https://musescore.org/project/note-names)
~~Harp Pedal Diagram~~

**Colour notes**
Colornotes (already packaged with MS3.6)

**Playback**
[Tunings and Temperaments](https://musescore.org/en/project/tunings-and-temperaments) (two separate plugins)
[Modal Tuning](https://musescore.org/en/project/modal-tuning) (a fork of the above)

Image assets for these plugins have already been created by @jessjwilliamson and sent to the developer team.

@Eism let me know if you have any questions!
Thanks

bkunda:
I've crossed-out "harp pedal diagram" because this is now a GSoC project, which will imbed comparable functionality within MS itself.

PR #11866, opened 2022-06-02
[MU4] Plugins actualization

Eism:
Resolves: #10812


jeetee:
Is the ABC Import(Export) plugin coming in a different PR?
If there won't be a musescore hosted webservice like ABC_import.qml used to use and there is an expectancy to use local tools instead; will MuseScore bundle those tools?

Eism:
@jeetee 
The best solution would be to make requests to the server. Do you know why the server is not available?

Jojo-Schmitz:
Or integrate https://musescore.org/en/project/abcimpex, as #10812 suggests

jeetee:
An advantage to the offline approach is that it works when offline (hello, I'm captain obvious!).
The CON to me is having to provide/support the local version of the tool then (including the eventual python environment) which seems like a fairly big bundling on our side.

The advantage of the server approach is ease of plugin installation for the user. If we go that route I feel hosting the service at .org or .com makes a lot of sense, to be able to fix server issues like the current plugin is having.
Self-hosting might also open up the ability for export as well?

Another option (the most clean and preferred imho, but almost certainly not for 4.0) is to translate that tool into C++ and make ABC a built-in import/export format.

Eism:
@jeetee I agree with you
@Tantacrul Do we need this plugin in 4.0 or can we delay the implementation?

Tantacrul:
> @jeetee I agree with you
> 
> @Tantacrul Do we need this plugin in 4.0 or can we delay the implementation?

I think it is fine to delay this one until you have an ideal technical solution. Can you create a separate task for it and we'll bump to 4.x?

MarcSabatella:
I have only minimal experience with AppSpot, but I suspect it’s something very simple to recreate.  I know something similar can be done through GitHub - GitHub.io, I think?  I can try to look into this.

src/plugins/api/qmlpluginapi.h
```diff
@@ -81,19 +81,21 @@ class PluginAPI : public mu::engraving::QmlPlugin
     INJECT(plugins, mu::context::IGlobalContext, context)
 
     /** Path where the plugin is placed in menu */
-    Q_PROPERTY(QString menuPath READ menuPath WRITE setMenuPath)
```
RomanPudashkin, line 84: I would recommend not changing the old API. It definitely requires some redesing, but that's not for this PR

src/plugins/view/pluginview.cpp
```diff
@@ -103,6 +136,9 @@ void PluginView::run()
     //connect(m_view, &QQuickView::closing, this, &PluginView::finished);
     connect(m_view, SIGNAL(closing(QQuickCloseEvent*)), this, SIGNAL(finished()));
 
+    if (pluginHasUi()) {
```
RomanPudashkin, line 139: I would recommend rewriting this code as follows:

```
void PluginView::run()
{
    if (!pluginHasUi()) {
        m_qmlPlugin->runPlugin();
        return;
    }
    
    destroyView();
    m_view = new QQuickView(engine(), nullptr);
    ...
}
```

src/plugins/view/pluginview.cpp
```diff
@@ -35,6 +35,12 @@ PluginView::PluginView(const QUrl& url, QObject* parent)
 {
     m_component = new QQmlComponent(engine(), url);
     m_qmlPlugin = qobject_cast<mu::engraving::QmlPlugin*>(m_component->create());
+
+    connect(m_qmlPlugin, &mu::engraving::QmlPlugin::close, [this]() {
```
RomanPudashkin, line 39: I would suggest renaming this signal to closeRequested

RomanPudashkin (approved):

jeetee:
FYI: I just got a basic python "hello world" enabled on my server, so I might be able to run the convertor (perhaps even both directions) in the coming week(s) via https://musescore.jeetee.net/

I'd still prefer a .org hosted version, but due to the amount of users complaining about the current plugin failures it can act as a duct tape stop-gap intermediate solution

Tantacrul:
@jeetee - I think we can sort that out!

DM me and we'll figure out the details

Issue #11747, opened 2022-05-25
[MU4 Issue] [GP] Rendering issue with drum part after importing


boryskuzmenko:
**Describe the bug**
The problem could be that some of beams belong to one voice but pointed to different directions.
The issue also reproducible after saving imported file

**To Reproduce**
1. Open test file
[drums issue.gp.zip](https://github.com/musescore/MuseScore/files/8770107/drums.issue.gp.zip)
2. Pay attention to the mess in written drum fills

**Expected behavior**
Drum part is rendered according to the default notation rules

**Screenshots**
![Capture 2022-05-25_12-45-35_PM](https://user-images.githubusercontent.com/101250347/170239434-a225669d-c99c-43e4-af01-ce97f665a16a.png)

boryskuzmenko:
🟢 **FIXED** in https://github.com/musescore/MuseScore/pull/11873
Tested on macOS 12, Win 11, Ubuntu 20

PR #11873, opened 2022-06-02
Fix stem direction in drum staff beams

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/11747

The stem direction needs to be set properly in `Chord::computeUp()` whether or not an explicit parent is set.


asattely:
The vtest failures actually demonstrate correct behavior--these vtests have custom stem directions for the grace note beams which are now respected

boryskuzmenko (approved):

alexpavlov96 (approved):

boryskuzmenko:
🟢 **FIXED**

PR #11882, opened 2022-06-03
[MU4] Fix #11856: Crash when moving multiple staves 

### LINKED ISSUE #11856: [MU4 Issue] Crash when moving multiple staves
**Describe the bug**
When in the instrument panel staves of two or more instruments are selected, a move or remove of these staves will crash MU4
**To Reproduce**
1. Create a score with e.g. a Piano and Organ.
2. Select the first staff of the Piano.
3. Select a staff of the Organ (Ctrl--Shift-Click on e.g. first staff of the Organ).
4. Click on the "Down" button.
5. MU4 crashes.

**Expected behavior**
One of
1. Select staves are moved or
2. Selection of staves of multiple instruments is blocked or
3. If staves of multiple instruments are selected, moving is disabled.

**Screenshots**
N.a.

**Platform information**
 - Linux

**Additional context**
This issue was found when looking into #10034. The idea is, `Add staff` will create a new staff **after** the selected staff. However, when staves of multiple instruments are selected it is not obvious what is supposed to happen. Is the staff created for the instrument for which `Add staff` is clicked only or will a staff be created after **all** selected staves. 
To solve issue #10034 solution 2 is preferred. With option 3 `Add staff` should be blocked also. Option 1 might require another `Add staff` option (e.g. via the popup menu of a staff).

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11856

### DIFF
--- a/src/framework/uicomponents/view/itemmultiselectionmodel.cpp
+++ b/src/framework/uicomponents/view/itemmultiselectionmodel.cpp
@@ -60,14 +60,22 @@ QList<int> ItemMultiSelectionModel::selectedRows() const
 void ItemMultiSelectionModel::select(const QModelIndex& index)
 {
     Qt::KeyboardModifiers modifiers = QApplication::keyboardModifiers();
-    bool modifiersAllowed = m_allowedModifiers & modifiers;
 
+    //! NOTE: always treat simultaneously pressed Ctrl and Shift as Ctrl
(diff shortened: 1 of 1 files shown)


njvdberg:
Checked PR and is indeed solving the reported issue by making it impossible selecting staves of multiple instruments.

njvdberg (approved):

Eism (approved):

boryskuzmenko:
🟢 **FIXED**
Tested on macOS 12, Win 11

Issue #11697, opened 2022-05-19
[MU4 Issue] Fix dialog messaging for assigning shortcuts


bkunda:
**Describe the bug**
UI microcopy in the "Enter Shortcut Sequence" dialog would make more sense to adopt the pattern established in the "Assign MIDI mapping" dialog. 

**Current "Enter Shortcut Sequence** dialog
<img width="535" alt="Screenshot 2022-05-19 at 11 39 33 pm" src="https://user-images.githubusercontent.com/86290556/169307045-ece8aa9f-e86a-4518-88eb-4ade72128b47.png">
_Note: problems with buttons extending beyond the dialog frame (See #11474)_

**Current "Assign MIDI mapping" dialog** 
<img width="537" alt="Screenshot 2022-05-19 at 11 40 41 pm" src="https://user-images.githubusercontent.com/86290556/169307295-b2b4ec38-aa69-4969-861f-3bf043780f51.png">

**Desired "Enter Shortcut Sequence dialog (Expected behaviour)**
<img width="538" alt="new-shortcut-1" src="https://user-images.githubusercontent.com/86290556/169308182-176caa60-ff2c-4892-9619-2e94edb9d9ad.png">

In this dialog:
- Icons should already be created in `musescoreicon` and would appear if they appear already in **Preferences** > **Shortcuts**
- If an icon is not available, only the text label will be shown (centred). E.g.
<img width="538" alt="new-shortcut-2" src="https://user-images.githubusercontent.com/86290556/169308226-198d0042-01a5-4fc1-8f27-24e07d68b371.png">

Thanks!

bkunda:
Closing this for now as we have already made alternative changes to the assign shortcut dialog

PR #11887, opened 2022-06-05
[MU4] Fix musescore#11697: Fix dialog messaging for assigning shortcuts

iwoithe:
Resolves: #11697

**Screenshots**

image

image


src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -150,6 +150,7 @@ QVariant ShortcutsModel::currentShortcut() const
 
     QVariantMap obj;
     obj["title"] = actionTitle(sc.action);
+    obj["icon"] = static_cast<int>(action(sc.action).iconCode);
```
cbjeukendrup, line 153: I think it would be good to rewrite this as:
```suggestion
    const UiAction& action = this->action(sc.action);

    QVariantMap obj;
    obj["title"] = action.title;
    obj["icon"] = static_cast<int>(action.iconCode);
```
That way, you look up the `UiAction` only once, instead of twice (first through `actionTitle()`, then directly).

src/framework/shortcuts/qml/MuseScore/Shortcuts/ShortcutsPage.qml
```diff
@@ -74,6 +74,7 @@ Item {
 
         function startEditCurrentShortcut() {
             editShortcutDialog.startEdit(shortcutsModel.currentShortcut, shortcutsModel.shortcuts())
+            editShortcutDialog.currentShortcut = shortcutsModel.currentShortcut
```
cbjeukendrup, line 77: I think it is preferable to move this line to EditShortcutDialog.startEdit.

Or, maybe even better, get rid of the `currentShortcut` property and instead add a Q_PROPERTY to `EditShortcutModel`, and use that in QML.

src/framework/shortcuts/qml/MuseScore/Shortcuts/EditShortcutDialog.qml
```diff
@@ -69,13 +71,36 @@ Dialog {
...
             StyledTextLabel {
                 anchors.horizontalCenter: parent.horizontalCenter
 
                 width: parent.width
-                text: qsTrc("shortcuts", "Define keyboard shortcut")
-                horizontalAlignment: Text.AlignLeft
-                font:ui.theme.headerBoldFont
+                text: qsTrc("shortcuts", "Press a key or sequence of keys to assign a shortcut to this function.")
+                horizontalAlignment: Text.AlignHCenter
+                font: ui.theme.bodyFont
```
cbjeukendrup, line 103: Since these two properties have their default value, you may as well remove them.

src/framework/shortcuts/qml/MuseScore/Shortcuts/EditShortcutDialog.qml
```diff
@@ -35,6 +35,7 @@ Dialog {
     function startEdit(shortcut, allShortcuts) {
         open()
         model.load(shortcut, allShortcuts)
+        model.currentShortcut = shortcut
```
cbjeukendrup, line 38: You could remove this line and the `setCurrentShortcut` method in C++ (i.e. make `currentShortcut` a readonly property), and instead put the contents of `setCurrentShortcut` in the `load` method.

Jojo-Schmitz:
seems a `String` label is in order, as this PR ;-)

src/framework/shortcuts/qml/MuseScore/Shortcuts/EditShortcutDialog.qml
```diff
@@ -68,13 +68,30 @@ Dialog {
...
+            }
+
             StyledTextLabel {
                 anchors.horizontalCenter: parent.horizontalCenter
 
                 width: parent.width
-                text: qsTrc("shortcuts", "Define keyboard shortcut")
-                horizontalAlignment: Text.AlignLeft
-                font:ui.theme.headerBoldFont
+                text: qsTrc("shortcuts", "Press a key or sequence of keys to assign a shortcut to this function.")
```
Jojo-Schmitz, line 94: Sure about the trailing fullstop?

bkunda:
@iwoithe @cbjeukendrup We have since made alternative changes to the assign shortcut dialog, and I don't think we should attempt any further changes at this stage. 
Thank you for tackling this though @iwoithe. I'm sorry that we won't be merging this at this stage. 
I will close #11887 for now.

src/framework/shortcuts/qml/MuseScore/Shortcuts/EditShortcutDialog.qml
iwoithe: I was following the designs in the issue description. Perhaps it should have been removed.

> BTW: that merge commit is not supposed to be part of this PR, better rebase using something like git pull --rebase upstream master; git push -f

Yeah, sorry about that. I was kind of pressed for time earlier today when I went through the merge conflict.

iwoithe:
> Thank you for tackling this though @iwoithe.

No problem.

> I'm sorry that we won't be merging this at this stage.

That's alright. I learnt more on MS's source code from solving this issue. I am looking forward to tackling more issues in the future. :slightly_smiling_face:

Issue #11809, opened 2022-05-30
[MU4 Task] Adjust alignment and drawing of text lines


oktophonie:
## Alignment of text relative to line

<img width="1241" alt="image" src="https://user-images.githubusercontent.com/20604032/171012116-eec72dcb-7926-4191-be29-100c47d20125.png">

This example shows a default 'cresc.' line, with the line thickness increased to 0.5sp to exaggerate the issue.
The origin point of the item is the vertical centre of the line. The text can be aligned relative to this point. (These vertical alignment options, 'top', 'centre', 'bottom', 'baseline', are currently removed from the Properties panel in MU4, but may be restored later.)

Currently, 'bottom' or 'baseline' alignment means that the bottom of the bounding box of the text, or the baseline of the text (respectively) will vertically align with the vertical _centre_ of the line. I propose that it should instead align with the _bottom_ of the line, i.e. the text will be lower by half of the line width, resulting in this:

<img width="788" alt="image" src="https://user-images.githubusercontent.com/20604032/171014244-c98277f5-bd58-424e-98c8-a1af1232e18c.png">

The same thing should happen for the 'baseline' setting; for 'top', the top of the text bounding box should align with the _top_ of the line; for 'centre', the behaviour should stay as it currently is (centre of text aligns with centre of line).

I believe that lyrics extension lines are already aligned in this 'correct' way. Compare here the _cresc._ line (above) to the lyrics (below):

<img width="951" alt="image" src="https://user-images.githubusercontent.com/20604032/171016013-c20cf090-968c-4b56-b749-ffe5105e4f49.png">

The alignment should be identical.

## Width of dashes and gaps

Even if the width of the dash and the gap are set the same (and 3 and 3, as shown here, appears to approximate to the 'default' dashed line), the dash is clearly wider than the gap:

<img width="772" alt="image" src="https://user-images.githubusercontent.com/20604032/171016699-1836bb7e-6649-4f0a-a1f7-b5179e013251.png">
(Screenshot from MU3.6, but the behaviour is identical in MU4.)

It seems reasonable for a user to expect that, if these two values are the same, then the widths should be the same. This should be corrected. (Is a cap being drawn on each dash, perhaps?)

Furthermore, the thickness of the line seems to significantly affect both widths. Here the top line has a thickness of 0.1sp and the bottom line has a thickness of 0.2sp:

[...]

PR #11904, opened 2022-06-06
Improvements for Text Lines

cbjeukendrup:
Resolves: #11809


cbjeukendrup:
The current VTest failure looks more or less expected to me. The line width for pedal markings is 0.15sp, and the "Ped." mark was shifted down (using the "offset" property) by 0.15sp, so its baseline ended up 0.075sp below the bottom of the line. Now, the "Ped." marking will be shifted down to align its baseline with the bottom of the line, so an additional offset is not necessary anymore.

oktophonie (approved):

RomanPudashkin (approved):

Issue #11894, opened 2022-06-06
[MU4 Task] Add large double flat & double sharp to 'common symbols'


Tantacrul:
**Task description**
There is a small oversight here, where the regular sized double sharp/flat symbols are missing, which means that they cannot easily be added to text descriptions. This would make Harp pedal changes really awkward to create.

<img width="694" alt="image" src="https://user-images.githubusercontent.com/21022311/172146987-1e73adc3-ad92-4b17-8cbf-c368858d26a9.png">

PR #11905, opened 2022-06-06
Special Characters dialog: add double flat and sharp

cbjeukendrup:
It looks like they were removed at some point because according to a comment in the code they are "not supported in most fonts". But since they do work in Edwin, the default font, let's bring them back.

Resolves: #11894


Jojo-Schmitz:
I believe there were never in, other than in that commented out form.

RomanPudashkin (approved):

PR #11907, opened 2022-06-06
[MU4] Fix MSVC compiler warning

Jojo-Schmitz:
reg.	declaration of 'articulation' hides global declaration (C4459), since #11865

Follow-up to #11868, apparently I missed to include it there, but now it reverts that PR and renames the conflicting global declaration instead.

### DIFF
--- a/src/engraving/playback/metaparsers/chordarticulationsparser.cpp
+++ b/src/engraving/playback/metaparsers/chordarticulationsparser.cpp
@@ -117,8 +117,8 @@ void ChordArticulationsParser::parseSpanners(const Chord* chord, const Rendering
 
 void ChordArticulationsParser::parseArticulationSymbols(const Chord* chord, const RenderingContext& ctx, mpe::ArticulationMap& result)
 {
-    for (const Articulation* artic : chord->articulations()) {
-        SymbolsMetaParser::parse(artic, ctx, result);
+    for (const Articulation* articulation : chord->articulations()) {
+        SymbolsMetaParser::parse(articulation, ctx, result);
     }
 }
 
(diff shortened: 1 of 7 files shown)


cbjeukendrup:
Actually I would be interested to find that notorious global `articulation` declaration. I can't imagine that `articulation` is the best name for that global declaration.

Jojo-Schmitz:
We could maybe rename it to `midiArticulation`?

src/engraving/playback/metaparsers/chordarticulationsparser.cpp
```diff
@@ -117,8 +117,8 @@ void ChordArticulationsParser::parseSpanners(const Chord* chord, const Rendering
 
 void ChordArticulationsParser::parseArticulationSymbols(const Chord* chord, const RenderingContext& ctx, mpe::ArticulationMap& result)
 {
-    for (const Articulation* artic : chord->articulations()) {
-        SymbolsMetaParser::parse(artic, ctx, result);
+    for (const mu::engraving::Articulation* articulation : chord->articulations()) {
```
Jojo-Schmitz, line 120: Not sure we need/want that `mu::engraving::` here?

src/notation/view/noteinputbarmodel.cpp
```diff
@@ -436,8 +436,8 @@ std::set<SymbolId> NoteInputBarModel::resolveCurrentArticulations() const
 
     auto chordArticulations = [](const Chord* chord) {
         std::set<SymbolId> result;
-        for (Articulation* artic: chord->articulations()) {
-            result.insert(artic->symId());
+        for (Articulation* articulation: chord->articulations()) {
```
Jojo-Schmitz, line 439: or whether we'd need that `mu::engraving::` here?

cbjeukendrup (approved):

Issue #11864, opened 2022-06-02
[MU4 Task] Add ability to arrow buttons work in auto repeat mode when holding mouse button (to scroll part tabs)


DmitryArefiev:
**Task description**
https://github.com/musescore/MuseScore/pull/11582#issuecomment-1143609319

**To Reproduce**
1. Open a big score
2. Open all part (you see arrow buttons at the top right)
3. Click one of them and hold the key - **it should trigger auto repeat mode and start scroll process**

![Capture](https://user-images.githubusercontent.com/90187801/171617203-91fa782e-11fa-4213-8504-a5bf505b202d.PNG)

**Additional context**
#11582

cbjeukendrup:
@DmitryArefiev No problem, I added it to the PR! I was initially afraid that it would be too fast if it was the same as the inspector. But now I realize that _if_ the user clicks and holds, they probably _want_ to scroll fast, so it's probably a good change.

PR #11314, opened 2022-04-23
Dynamic behaviour of clear text button when text is present in the Search Field

the-next-ceo:
Resolves: *(direct link to the issue)*

*(short description of the changes and the motivation to make the changes)*

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [x] I created the test (mtest, vtest, script test) to verify the changes I made

https://user-images.githubusercontent.com/66167731/164948789-9e0205a6-dfe8-4cc6-9e5e-f5e67518d212.mov

PR #11910, opened 2022-06-06
Enable press and hold for notation tabs scroll buttons

cbjeukendrup:
Resolves: #11864

(While looking at IncrementalPropertyControl.qml for a quick reference, I found that there were some QML warnings in that file so I quickly fixed those too.)

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/IncrementalPropertyControl.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/IncrementalPropertyControl.qml
@@ -49,7 +49,7 @@ Item {
     implicitHeight: 30
     implicitWidth: parent.width
 
-    navigation.name: root.objectName != "" ? root.objectName : "IncrementalControl"
+    navigation.name: Boolean(root.objectName) ? root.objectName : "IncrementalControl"
 
     function increment() {
         var value = root.isIndeterminate ? 0.0 : currentValue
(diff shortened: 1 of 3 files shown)


src/notation/qml/MuseScore/NotationScene/internal/NotationSwitchScrollArrowButton.qml
```diff
@@ -26,6 +26,8 @@ import MuseScore.UiComponents 1.0
 FlatButton {
     id: root
 
+    signal scrollRequested
```
Eism, line 29: Please add round brackets

src/framework/uicomponents/qml/MuseScore/UiComponents/IncrementalPropertyControl.qml
```diff
@@ -49,7 +49,7 @@ Item {
     implicitHeight: 30
     implicitWidth: parent.width
 
-    navigation.name: root.objectName != "" ? root.objectName : "IncrementalControl"
+    navigation.name: root.objectName ?? "IncrementalControl"
```
Eism, line 52: My opinion about '??'
https://github.com/musescore/MuseScore/pull/11314#pullrequestreview-957450120

I suggest not to use '??'

cbjeukendrup: Okay, I reverted it to the `Boolean(...) ? ... : ...` solution. I think we shouldn't let a buggy IDE limit which language features we use though...

DmitryArefiev:
Tested #11864 on Win10, Mac12, LinuxMint 20.2 - **DONE**

RomanPudashkin (approved):

Issue #11897, opened 2022-06-06
[MU4 Issue] Unable to run latest builds on Linux


DmitryArefiev:
**Describe the bug**
Unable to run latest build on Linux

try these
https://github.com/musescore/MuseScore/actions/runs/2445567865
https://github.com/musescore/MuseScore/actions/runs/2446767239

see [terminal output.txt](https://github.com/musescore/MuseScore/files/8844994/terminal.output.txt)

**Platform information**
Linux Mint 20.2

Jojo-Schmitz:
See also https://musescore.org/en/node/332616

PR #11911, opened 2022-06-07
[MU4] fixed #11478: keyboard navigation breaks at key and time signatures if >1 staff

### LINKED ISSUE #11478: [MU4 Issue] Accessibility: keyboard navigation breaks at key and time signatures if >1 staff
Navigation of the score one element at a time via Alt+Left/Right it how blind users use MuseScore.  Currently in MU4 (as of the build I'm using from a few days ago, anyhow), navigation breaks when you reach a key or time signature, if there is more than one staff.

To reproduce:

1) new score, grand staff
2) start pressing Alt+Right to navigate

Result: when you reach the key or time signature, you will note they are selected across both staves.  This much is kind of normal when *clicking*, in that we do "often" want selecting a time signature to select them across all staves.  But navigation by keyboard needs to select just the one on the current staff.  Otherwise the screen reader doesn't know what to read.  And more importantly, right now pressing Alt+Right again does nothing - navigation gets stuck at the list selection.

Eism:
Resolves: #11478

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -2468,7 +2468,7 @@ void NotationInteraction::moveElementSelection(MoveDirection d)
         endEditElement();
     }
 
-    select({ toEl }, SelectType::SINGLE);
+    select({ toEl }, SelectType::REPLACE);
     resetHitElementContext();
     showItem(toEl);
 


RomanPudashkin (approved):

DmitryArefiev:
Tested #11478 on Win10, Mac12 - **FIXED**    

Testing on Linux is blocked by #11897

PR #11921, opened 2022-06-08
External wav

jsm222:
Resolves: #11801 

Add an external load  media button to add a wave traack 48000 32bit. Cancel filedialog to remove the track.

### DIFF
--- a/src/framework/audio/internal/worker/externalaudiosource.cpp
+++ b/src/framework/audio/internal/worker/externalaudiosource.cpp
@@ -23,12 +23,12 @@
 using namespace mu;
 using namespace mu::audio;
 
-ExternalAudioSource::ExternalAudioSource(const TrackId trackId, const io::Device* playbackData)
+ExternalAudioSource::ExternalAudioSource(const TrackId trackId, io::Device* playbackData)
     : m_trackId(trackId), m_playbackData(playbackData)
 {
     ONLY_AUDIO_WORKER_THREAD;
-    (const_cast<io::Device*>(playbackData))->open(mu::io::Device::ReadOnly);
-    b = (const_cast<io::Device*>(playbackData))->readAll();
+    playbackData->open(mu::io::Device::ReadOnly);
+    b = playbackData->readAll();
 
     drwav_init_memory(&m_wav, b.constData(), b.size(), nullptr);
     if (m_wav.fmt.bitsPerSample != 32) {
--- a/src/framework/audio/internal/worker/externalaudiosource.h
+++ b/src/framework/audio/internal/worker/externalaudiosource.h
@@ -29,7 +29,7 @@ class ExternalAudioSource : public ITrackAudioInput, public async::Asyncable
 {
 public:
     ExternalAudioSource();
-    explicit ExternalAudioSource(const TrackId trackId, const io::Device*);
+    explicit ExternalAudioSource(const TrackId trackId, io::Device*);
 
     bool isActive() const override;
     void setIsActive(const bool active) override;


jsm222:
> You don't need to cast to QIODevice, see src/framework/global/io/iodevice.h.
The cast is there to loose the const..

cbjeukendrup:
> The cast is there to loose the const..

In that case, you should const_cast to `io::Device`, and not use any Qt stuff. It is just coincidence that the current solution doesn't crash, because `QIODevice` and `io::Device` are entirely unrelated classes.

src/framework/audio/internal/worker/externalaudiosource.cpp
```diff
@@ -0,0 +1,77 @@
+#include "externalaudiosource.h"
```
Jojo-Schmitz, line 1: copyright header missing

src/framework/audio/internal/worker/externalaudiosource.h
```diff
@@ -0,0 +1,39 @@
+#ifndef EXTERNALAUDIOSOURCE_H
```
Jojo-Schmitz, line 1: here too

src/framework/audio/internal/worker/tracksequence.cpp
```diff
@@ -28,11 +28,11 @@
 #include "internal/audiothread.h"
 #include "clock.h"
 #include "eventaudiosource.h"
+#include "externalaudiosource.h"
 #include "sequenceplayer.h"
 #include "sequenceio.h"
 #include "audioengine.h"
 #include "audioerrors.h"
-
```
Jojo-Schmitz, line 35: better keep that empty line

src/framework/audio/internal/worker/tracksequence.cpp
```diff
@@ -94,7 +94,6 @@ RetVal2<TrackId, AudioParams> TrackSequence::addTrack(const std::string& trackNa
     trackPtr->outputHandler = mixer()->addChannel(newId, trackPtr->inputHandler).val;
     trackPtr->setInputParams(requiredParams.in);
     trackPtr->setOutputParams(requiredParams.out);
-
```
Jojo-Schmitz, line 97: here too

src/playback/internal/playbackcontroller.cpp
```diff
@@ -809,6 +846,7 @@ void PlaybackController::setupSequenceTracks()
 {
     m_trackIdMap.clear();
 
+
```
Jojo-Schmitz, line 849: this here is too much though

src/playback/internal/playbackcontroller.cpp
```diff
@@ -453,7 +485,6 @@ void PlaybackController::setCurrentPlaybackStatus(PlaybackStatus status)
     if (m_currentPlaybackStatus == status) {
         return;
     }
-
```
Jojo-Schmitz, line 456: change not needed, may just cause merge conflicts later

src/playback/internal/playbackcontroller.cpp
```diff
@@ -642,7 +673,6 @@ void PlaybackController::addTrack(const InstrumentTrackId& instrumentTrackId, co
     }
 
     uint64_t notationPlaybackKey = reinterpret_cast<uint64_t>(notationPlayback().get());
-
```
Jojo-Schmitz, line 645: same here

src/framework/audio/internal/worker/externalaudiosource.cpp
```diff
@@ -0,0 +1,106 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include "externalaudiosource.h"
+using namespace mu;
+using namespace mu::audio;
+
+ExternalAudioSource::ExternalAudioSource(const TrackId trackId, const io::Device* playbackData)
```
cbjeukendrup, line 26: Why is the `const` necessary in `const io::Device*`? It looks like it isn't, and removing it makes the code simpler and better.

jsm222: Agree, I just tested it
```diff
diff --git a/src/framework/audio/internal/worker/externalaudiosource.cpp b/src/framework/audio/internal/worker/externalaudiosource.cpp
index d0d4067360..3bf057d2f4 100644
--- a/src/framework/audio/internal/worker/externalaudiosource.cpp
+++ b/src/framework/audio/internal/worker/externalaudiosource.cpp
@@ -23,12 +23,12 @@
 using namespace mu;
 using namespace mu::audio;
 
-ExternalAudioSource::ExternalAudioSource(const TrackId trackId, const io::Device* playbackData)
+ExternalAudioSource::ExternalAudioSource(const TrackId trackId, io::Device* playbackData)
     : m_trackId(trackId), m_playbackData(playbackData)
 {
     ONLY_AUDIO_WORKER_THREAD;
-    (const_cast<io::Device*>(playbackData))->open(mu::io::Device::ReadOnly);
-    b = (const_cast<io::Device*>(playbackData))->readAll();
+    playbackData->open(mu::io::Device::ReadOnly);
+    b = playbackData->readAll();
 
     drwav_init_memory(&m_wav, b.constData(), b.size(), nullptr);
     if (m_wav.fmt.bitsPerSample != 32) {
diff --git a/src/framework/audio/internal/worker/externalaudiosource.h b/src/framework/audio/internal/worker/externalaudiosource.h
index a2682909a3..4a6e2be373 100644
--- a/src/framework/audio/internal/worker/externalaudiosource.h
+++ b/src/framework/audio/internal/worker/externalaudiosource.h
@@ -29,7 +29,7 @@ class ExternalAudioSource : public ITrackAudioInput, public async::Asyncable
 {
 public:
     ExternalAudioSource();
-    explicit ExternalAudioSource(const TrackId trackId, const io::Device*);
+    explicit ExternalAudioSource(const TrackId trackId, io::Device *);
 
     bool isActive() const override;
     void setIsActive(const bool active) override;
``

bkunda:
@jsm222 we appreciate this as a proof of concept, but have come to the conclusion that it's unfortunately not ready to be included in a release. This is partly because of the limitation in supported file formats, and also given the age of the PR and the resulting conflicts, it might be more convenient if we implement this feature ourselves when we have the resources available to do so in a more comprehensive way.

PR #11923, opened 2022-06-08
Allowing function keys while typing for calling shortcuts

VanSHOE:
By-Product of: https://github.com/musescore/MuseScore/pull/11586
This PR intends to filter out function keys while text-editing as no shortcuts are picked up while typing. But they are required for shortcuts like `show-keys`. 

Thank you @cbjeukendrup for guidance!


Eism:
@VanSHOE 
If we decide to change the "show keys" keyboard shortcut to non-functional keys, your solution will break
I think we should fix this with shortcut context

Eism:
will be fixed in #11576

VanSHOE:
Understood! I too think, like your comment in #11576, that this would require quite a bigger change.

Issue #11044, opened 2022-03-31
[MU4] Staff instrument alignment not correct when using two lines


henkdegroot:
Instruments names using two lines are not all "right aligned". The style is set to right align, however the first line is left aligned.
All text should be right aligned.

Staff instrument names as shown in MU3:
![image](https://user-images.githubusercontent.com/13550012/161150825-85e00d1b-933a-44a1-b427-8d1f6a8cf8bf.png)

Staff instrument names as shown in MU4:
![image](https://user-images.githubusercontent.com/13550012/161150994-36101064-8ec5-4b7c-b793-24aee8828213.png)

Tantacrul:
Oh, wait. I misread the issue. Reopening. @oktophonie - thoughts?

Tantacrul:
This feels a bit more serious than a P3 then.

PR #11945, opened 2022-06-08
Fix alignment of instrument names with multiple lines

cbjeukendrup:
- The multiple lines are now correctly aligned relative to each other again
- Instead of counting `Sid::instrumentNameOffset` twice, correctly take the brackets width into account for determining the system left margin if instrument names are present
- Fix a possible bug where bracket width was counted using integers instead of qreals
- Added a comment that the `System::totalBracketOffset` method does not work (fixing it is a separate job...)
- And a tiny optimization / code style improvement in TextBlock::layout().

Resolves: #11044

### DIFF
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -254,27 +254,22 @@ void System::adjustStavesNumber(size_t nstaves)
 }
 
 //---------------------------------------------------------
-//   systemNamesWidth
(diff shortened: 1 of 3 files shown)


oktophonie (approved):

RomanPudashkin (approved):

henkdegroot:
@cbjeukendrup Thank you so much! Looking good now.

Issue #11214, opened 2022-04-15
[MU4 Issue] [Accessibility] Unable to continue navigation with arrows after Esc from text editing


DmitryArefiev:
**To Reproduce**
1. Navigate with Alt+arrow and select some text
2. Enter editing mode (Alt+Shift+E or Return)
3. Esc to finish editing
4. Try to continue navigation with Alt+arrow - **it doesn't work**

https://user-images.githubusercontent.com/90187801/163594055-3f72cb3c-ddf0-4c0c-86e6-cd8fb94719a8.mp4

**Platform information**
macOS, Windows, Linux

DmitryArefiev:
Fixed in #11214

PR #11951, opened 2022-06-09
Fix #312589: Frames with text are not accessible

HemantAntony:
Resolves: [#312589](https://musescore.org/en/node/312589)

### DIFF
--- a/src/engraving/libmscore/navigate.cpp
+++ b/src/engraving/libmscore/navigate.cpp
@@ -43,6 +43,77 @@
 using namespace mu;
 
 namespace mu::engraving {
+using ElementPair = std::pair<EngravingItem*, int>; // element and its index as a child
+
+//---------------------------------------------------------
+//   isChild1beforeChild2
+//    compares if two children are order correctly
+//    from top-left to bottom-right
+//---------------------------------------------------------
+
+static bool isChild1beforeChild2(const ElementPair& child1, const ElementPair& child2)
+{
+    Q_ASSERT(child1.first->parent() == child2.first->parent());
+
+    PointF p1 = child1.first->pos() + child1.first->bbox().center();
+    PointF p2 = child2.first->pos() + child2.first->bbox().center();
+
+    // If one child is *clearly* above the other then visit the higher one first
+    qreal verticalSeparation = p2.y() - p1.y();
+    if (verticalSeparation > 10) {
+        return true;
+    } else if (verticalSeparation < -10) {
+        return false;
+    }
+
+    // Children are roughly aligned vertically. Visit the one on the left first
+    qreal horizontalSeparation = p2.x() - p1.x();
+    if (horizontalSeparation > 0.0) {
+        return true;
+    } else if (horizontalSeparation < 0.0) {
+        return false;
+    }
+
+    // Children are aligned horizontally so check exact vertical position.
+    if (verticalSeparation > 0.0) {
+        return true;
+    } else if (verticalSeparation < 0.0) {
+        return false;
+    }
+
+    // Elements are at exact same position so fall back to using child index to determine order.
+    if (child1.second < child2.second) {
+        return true;
+    }
+
+    return false;
+}
+
+//---------------------------------------------------------
+//   toChildPairsSet
+//    return a set of children ordered
+//    on the basis of position
+//---------------------------------------------------------
+
+static std::set<ElementPair, decltype(& isChild1beforeChild2)> toChildPairsSet(const EngravingItem* element)
+{
+    std::set<ElementPair, decltype(& isChild1beforeChild2)> children(&isChild1beforeChild2);
+    int index = 0;
+
+    for (EngravingObject* object : element->children()) {
+        EngravingItem* engravingItem = toEngravingItem(object);
+
+        if (!engravingItem) {
+            continue;
+        }
+
+        children.insert(std::pair<EngravingItem*, int>(engravingItem, index));
+        index++;
+    }
+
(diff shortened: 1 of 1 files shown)


HemantAntony:
@shoogle This PR solves [#312589](https://musescore.org/en/node/312589)

shoogle:
@HemantAntony, thanks for the fix! This works well for Vertical Frames that contain Title, Subtitle, Composer and Lyricist text, but it is possible for these frames to contains other text styles as well (or it would be except the feature is broken in MU4; I'll create an issue about that separately).

## Example Score created in MU3
- [Frame Navigation.zip](https://github.com/musescore/MuseScore/files/8878493/Frame.Navigation.zip) (you need to change the extension to `.mscz` because of GitHub)

The file demonstrates more complicated use of frames with text, including:

### 1. Text Frames with different text styles
image

### 2. Horizonal Frame with multiple instances of the Frame style
image

### 3. Vertical Frame with custom styles and multiple instances of each style
image

## Desired behaviour

Your code already handles (1) correctly: press Alt+Right to select the Text Frame, then press Alt+Right again to select the text element inside.

For (2) and (3), you need to make sure that Alt+Right is able to get to every text element in the Horizonal / Vertical Frame. At the moment it skips elements that don't use the Title, Subtitle, Composer or Lyricist style, and it only accesses the last instance of that style (e.g. "Title 2" but not "Title 1").

The order in which text elements are accessed is not important at this stage, so I suggest you just access them in whatever order the elements appear in the frame's list of children. The screen reader will say "Title" or "Composer" before it announces the actual title or composer name, so it doesn't matter if the order is different sometimes because the user added the composer before they added the title, etc.

Whatever the order is for Alt+Right, make sure that Alt+Left accesses all the elements in the reverse of that order.

shoogle:
As discussed on Discord, @HemantAntony is going to attempt to traverse text elements in order based on their position on the page, starting in the top left and working towards the bottom right. I've updated the example file in the [comment above](https://github.com/musescore/MuseScore/pull/11951#issuecomment-1152295522) to add "Composer 3" and "Lyricist 3" at the exact same positions as "Composer 2" and "Lyricist 2" respectively. This is just to make sure that they are all traversed by Alt+Right. The exact order (2 then 3, or 3 then 2) is not important.
image

src/engraving/libmscore/navigate.cpp
```diff
@@ -760,8 +818,61 @@ EngravingItem* Score::prevElement()
...
+            for (EngravingObject* object : list) {
+                EngravingItem* engravingItem = toEngravingItem(object);
+
+                if (!engravingItem) {
+                    continue;
+                }
+
+                boxChildren.insert(std::pair<EngravingItem*, int>(engravingItem, index));
+                index++;
+            }
```
shoogle, line 867: Most of this code is identical to `nextElement()` so it should be factored out into a separate `static` function that takes the box as an argument and returns the set of child+index pairs. You can use the following declarations to make the code neater:

```c++
using ElementPair = std::pair<EngravingItem*, int>; // element and its index in parent

static bool compareElementPairs(const ElementPair& child1, const ElementPair& child2)
{
    PointF p1 = ...;
    PointF p2 = ...;
    // etc.
}

static std::set<ElementPair, decltype(&compareFunction)> childPairsSet(const EngravingItem* element)
{
    std::set<ElementPair, decltype(&compareFunction)> children(&compareElementPairs);

   // add elements to set

   return children;
}
```

Put that somewhere near the top of `navigate.cpp` outside of any functions, then inside `nextElement()` and `previousElement()` you just have to do:

```c++
auto boxChildren = childPairsSet(e);
// now use the set
```

shoogle (changes requested):
I've suggested some code improvements but the implementation works very well indeed!

src/engraving/libmscore/navigate.cpp
```diff
@@ -695,6 +766,26 @@ EngravingItem* Score::nextElement()
         case ElementType::VBOX:
         case ElementType::HBOX:
         case ElementType::TBOX: {
+            auto boxChildren = toChildPairsSet(e);
+
+            if ((getSelectedElement()->type() == ElementType::VBOX
```
Eism, line 771: Please move getSelectedElement() to a variable and use it

HemantAntony:
@Eism I updated the PR

Eism (approved):

src/engraving/libmscore/navigate.cpp
```diff
@@ -824,6 +930,20 @@ EngravingItem* Score::prevElement()
         case ElementType::VBOX:
         case ElementType::HBOX:
         case ElementType::TBOX: {
+            auto boxChildren = toChildPairsSet(e);
+
+            for (auto child = boxChildren.rbegin(); child != boxChildren.rend(); child++) {
+                if (getSelectedElement() != child->first) {
```
shoogle, line 936: @HemantAntony, you should also store `getSelectedElement()` in a variable outside the loop here.

HemantAntony: Oh right, I saw it in only one place and forgot that it was in a loop

shoogle (approved):

DmitryArefiev:
Tested [#312589](https://musescore.org/en/node/312589) on Win10, Mac12, LinuxMint 20.2 - **FIXED** (now is able to navigate with Alt+arrows between different text inside one frame, e.g. Compose - Subtitle - Title)

Also there is annoyance bug when switching between text elements after editing #11214 which also effects editing process inside text frame

PR #11952, opened 2022-06-09
fixed #8957: Update style when changing local properties

### LINKED ISSUE #8957: [MU4 Issue] Custom tremolo style set as default doesn't affect new tremolos, nor does it survive upon reload
**Describe the bug**
Custom tremolo style set as default doesn't affect new tremolos, nor does it survive upon reload.

**To Reproduce**
1. Need to have #8955's fix, otherwise the inspector won't work;
2. Create a minim-based two-note tremolo;
3. Set the tremolo style to a custom one (the second or the third one), then set the custom style as default in the inspector;
4. Add a similar tremolo, notice it doesn't automatically have the default custom style applied (bug 1);
5. Close the file and open again, notice the custom style is gone (bug 2).

**Expected behavior**
4. The new tremolo should automatically have the default custom style applied;
5. The custom style, which has been set as default, should survive a reload.

Eism:
Resolves: #8957

### DIFF
--- a/src/engraving/libmscore/tremolo.cpp
+++ b/src/engraving/libmscore/tremolo.cpp
@@ -171,7 +171,8 @@ void Tremolo::setTremoloType(TremoloType t)
         _lines = 1;
         break;
     }
-    computeShape();
+
+    styleChanged();
 }
 
 //---------------------------------------------------------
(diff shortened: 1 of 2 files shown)


DmitryArefiev:
Tested #8957 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #11953, opened 2022-06-09
Fix #307530: MusicXML export for hidden staves

HemantAntony:
Resolves: [#307530](https://musescore.org/en/node/307530)

@shoogle FYI

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -1692,21 +1692,22 @@ void MusicXMLParserPass2::part()
     _hasDrumset = hasDrumset(instruments);
 
     // set the parts first instrument
-    setPartInstruments(_logger, &_e, _pass1.getPart(id), id, _score, _pass1.getInstrList(id), _pass1.getIntervals(id), instruments);
+    Part* part = _pass1.getPart(id);
+    setPartInstruments(_logger, &_e, part, id, _score, _pass1.getInstrList(id), _pass1.getIntervals(id), instruments);
 
     // set the part name
     auto mxmlPart = _pass1.getMusicXmlPart(id);
-    _pass1.getPart(id)->setPartName(mxmlPart.getName());
+    part->setPartName(mxmlPart.getName());
     if (mxmlPart.getPrintName()) {
-        _pass1.getPart(id)->setLongName(mxmlPart.getName());
+        part->setLongName(mxmlPart.getName());
     }
     if (mxmlPart.getPrintAbbr()) {
-        _pass1.getPart(id)->setPlainShortName(mxmlPart.getAbbr());
+        part->setPlainShortName(mxmlPart.getAbbr());
     }
     // try to prevent an empty track name
-    if (_pass1.getPart(id)->partName() == "") {
+    if (part->partName() == "") {
         QString instrId = _pass1.getInstrList(id).instrument(Fraction(0, 1));
-        _pass1.getPart(id)->setPartName(instruments[instrId].name);
+        part->setPartName(instruments[instrId].name);
     }
 
 #ifdef DEBUG_VOICE_MAPPER
@@ -1738,10 +1739,10 @@ void MusicXMLParserPass2::part()
     }
 
     // stop all remaining extends for this part
-    Measure* lm = _pass1.getPart(id)->score()->lastMeasure();
+    Measure* lm = part->score()->lastMeasure();
     if (lm) {
         track_idx_t strack = _pass1.trackForPart(id);
-        track_idx_t etrack = strack + _pass1.getPart(id)->nstaves() * VOICES;
+        track_idx_t etrack = strack + part->nstaves() * VOICES;
         Fraction lastTick = lm->endTick();
         for (track_idx_t trk = strack; trk < etrack; trk++) {
             _extendedLyrics.setExtend(-1, trk, lastTick);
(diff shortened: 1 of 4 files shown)


RomanPudashkin:
Thanks for the fix! Could you add a unit test to check this fix? You can use this PR as an example:
https://github.com/musescore/MuseScore/pull/11413

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -2328,6 +2328,13 @@ void MusicXMLParserPass2::staffDetails(const QString& partId)
         t->setFrets(25);      // sensible default
     }
 
+    QString visible = _e.attributes().value("print-object").toString();
+    if (visible == "no") {
+        _score->staff(staffIdx)->setVisible(false);
+    } else if (visible != "yes") {
+        _logger->logError(QString("print-object should be \"yes\" or \"no\""));
+    }
```
shoogle, line 2336: If possible without major restructuring, once the import is nearly complete and all staves have been made visible or invisible in MuseScore's internal structures, you should check whether all staves in each `Part` are invisible and if so make the `Part` itself invisible, and make all the staves inside it visible again (they won't be shown if the `Part` is invisible. This will help to ensure that the import matches the export.

shoogle (changes requested):

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
HemantAntony: I think a better fix would be to make the `Part` invisible when all its staves are invisible. This is the current behaviour:
image

Do you think there would be edge cases where this would not be desired? (@Tantacrul too)

Tantacrul: I think you're right. We'd just need to make sure we avoid janky situations.
 

- If all your staves are empty and you add a new stave, it should be **visible** by default, thereby turning back on the visibility of your part
- (visa-versa) If you remove that visible stave, which means there are no more visible staves, then the part turns back off
- Turning on/off the part visibility affects the visibility of **all** its staves in all circumstances (to avoid loads of additional unnecessary logic)

Tantacrul: But it isn't necessary to do any of the above unless there is a good technical reason for it. From a UX point of view, it is fine as it is. However, changing it t be the way @HemantAntony suggest's isn't a problem either.

shoogle: It's probably best to do as @HemantAntony suggests and make the `Part` invisible if all of its staves are invisible, and the staves invisible if the Part is made invisible. There are other ways it could work but that way is probably the least complicated option.

src/importexport/musicxml/tests/tst_mxml_io.cpp
```diff
@@ -155,6 +155,7 @@ private slots:
     void hello() { mxmlIoTest("testHello"); }
     void helloReadCompr() { mxmlReadTestCompr("testHello"); }
     void helloReadWriteCompr() { mxmlReadWriteTestCompr("testHello"); }
+    void hiddenStaves() { mxmlIoTest("testHiddenStaves"); }
```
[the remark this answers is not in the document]
HemantAntony: @shoogle I don't need to do this for non-show `Parts` case because when you export to MusicXML, it writes the `Part` as hidden if all its staves are also hidden (unlike mscz exports)

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
HemantAntony: I sadly had to abandon this way of doing it though this is the ideal way of doing it. It turns out doing this would require some larger changes to the existing code. So I'm doing it the way @shoogle proposed. So the Part won't be hidden if all its staves are

HemantAntony:
I'll fix the unit tests tomorrow

Jojo-Schmitz:
It'll be an easy fix

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -1777,9 +1778,18 @@ void MusicXMLParserPass2::part()
...
         // Note: part has been read, staff type already set based on clef type and staff-details
         // but may be incorrect for a percussion staff that does not use a percussion clef
-        setStaffTypePercussion(_pass1.getPart(id), drumset);
+        setStaffTypePercussion(part, drumset);
     }
 
+    bool showPart = false;
+    for (Staff* staff : part->staves()) {
+        if (staff->visible()) {
+            showPart = true;
```
shoogle, line 1787: You should add a `break;` here.

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
```diff
@@ -2328,6 +2338,13 @@ void MusicXMLParserPass2::staffDetails(const QString& partId)
         t->setFrets(25);      // sensible default
     }
 
+    QString visible = _e.attributes().value("print-object").toString();
+    if (visible == "no") {
+        _score->staff(staffIdx)->setVisible(false);
+    } else if (visible != "yes") {
+        _logger->logError(QString("print-object should be \"yes\" or \"no\""));
+    }
```
shoogle, line 2346: What happens if the `print-object` attribute was not specified? Will that also print an error message in the log? I don't think we want it to do that.

shoogle (changes requested):

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
HemantAntony: I can't believe I forgot that

src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
HemantAntony: Yeah, I didn't notice that. I added a check to see if its empty

shoogle (approved):
Looks good! While testing I noticed a very strange problem with hiding staves _after_ a MusicXML file has been imported into MuseScore, but the problem already exists in the nightly builds so I created a separate issue for it (see #12186). I think this one is ready to merge as-is.

RomanPudashkin (approved):

PR #11957, opened 2022-06-09
[MU4] Fix #11874: New SoundFont notification dialog cannot accommodate entire text string

### LINKED ISSUE #11874: [MU4 Issue] New SoundFont notification dialog cannot accommodate entire text string
**Describe the bug**
In this dialog, which is triggered upon attempting to install a previously-installed SoundFont, there doesn't seem to be enough room in the header text block to accommodate the entire string (should end with the words "already exists.", regardless of how long the filename is).

**To Reproduce**
1. Drag-and-drop a previously-installed .sf2 file into the MuseScore 4 score window 
2. This dialog is triggered

**Expected behavior**
The final words in the text string ("...already exists.") should always be visible, regardless of how long the filename is. 

**Platform information**
 - MacOS 12.3.1 (M1)
 - Windows 11

**Additional context**
Discovered when testing PR #11828

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11874

### DIFF
--- a/src/framework/audio/internal/soundfontrepository.cpp
+++ b/src/framework/audio/internal/soundfontrepository.cpp
@@ -86,10 +86,12 @@ mu::Ret SoundFontRepository::addSoundFont(const SoundFontPath& path)
     }
 
     if (fileSystem()->exists(newPath.val)) {
-        title = qtrc("audio", "File \"%1\" already exists.")
-                .arg(newPath.val.toQString()).toStdString();
+        title = trc("audio", "File already exists. Do you want to overwrite it?");
 
-        btn = interactive()->question(title, trc("audio", "Do you want to overwrite it?"), {
+        std::string body = qtrc("audio", "File path: %1")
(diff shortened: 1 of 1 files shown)


Eism (approved):

cbjeukendrup:
Why don't we allow the text to take more than two lines? (See StandardDialogPanel.qml, line 153)
We will probably need that anyway in some places to fit translated versions of strings. 

The same goes for the body text label. It is limited to three lines, which is for example not enough for the "revert to factory settings" warning dialog:
Schermafbeelding 2022-06-09 om 14 16 21

RomanPudashkin:
> Why don't we allow the text to take more than two lines? (See StandardDialogPanel.qml, line 153)

This won't help if the file path length is long enough

RomanPudashkin:
Here is the example. The user will not be able to see "already exists", no matter how big the number in maximumLineCount is
Screenshot 2022-06-09 at 14 22 16

PR #11965, opened 2022-06-09
[MU4] Fix #11451: Chords are not played during Left/Right navigation

### LINKED ISSUE #11451: [MU4 Issue] Chords are not played during Left/Right navigation
**Describe the bug**
In MuseScore 4, only the top note of chords is played when you navigate through the score with the Left and Right arrow keys. In MuseScore 3 all notes were played. This is important for blind users because otherwise they will not know that the selected note belongs to a chord.

**Additional context**
When navigating with Alt+Right and Alt+Left, it is correct for only one note at a time to be played. This is because those shortcuts give access to all notes in the chord whereas Left and Right alone only give access to the top note.

### Example 1: Navigation with Right/Left arrow key _alone_
**UiActions: `play-next-chord`/`play-prev-chord`**

1. The note before a chord is selected.
  image

2. Press the Right arrow key. The top note of the chord is selected and **you should hear a sound**.
  image
**Expected behavior:** You should hear all three notes in the chord, even though only the top note is actually selected.
**Actual behavior:** You only hear the top note. **WRONG!**

3. Press the Right arrow key again. The rest after the chord is selected (and you should hear nothing).
  image

### Example 2: Navigation with Left/Right arrow keys _with Alt modifier_
**UiActions: `next-element`/`prev-element`**

1. The note before a chord is selected.
  image

2. Press Alt+Right arrow. The top note of the chord is selected and **you should hear a sound**.
  image
**Expected behavior:** This time you should only hear the top note.
**Actual behavior:** You only hear the top note. **CORRECT!**

3. Press Alt+Right arrow again. This time the middle note of the chord is selected (and you only hear the middle note).
  image

**Please fix Example 1 without breaking Example 2. Thanks!**

---

**Related edit (Martin)**
When in Note Input mode on a **Tablature stave**, pressing **up** or **down** to jump between notes in a chord - you will hear one note in the chord being played each time. The result should behave in the same way as MuseScore 3: no sound should be heard in this specific instance.

image

**Platform information**
 - OS: Windows

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11451

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -851,7 +851,8 @@ void NotationActionController::move(MoveDirection direction, bool quickly)
         return;
     }
 
-    EngravingItem* selectedElement = interaction->selection()->element();
+    const EngravingItem* selectedElement = interaction->selection()->element();
+    bool playChord = false;
 
     switch (direction) {
     case MoveDirection::Up:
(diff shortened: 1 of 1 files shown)


DmitryArefiev:
Tested #11451 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Eism (approved):

PR #11966, opened 2022-06-09
[MU4] Fix MinGW/MSVC compiler warnings

Jojo-Schmitz:


src/engraving/rw/xmlreader.cpp
```diff
@@ -220,9 +220,6 @@ String XmlReader::readXml()
 {
     static int count = 0;
     ++count;
-    if (count == 6) {
-        int k = -1;
-    }
```
igorkorsukov, line 225: fixed here https://github.com/musescore/MuseScore/pull/11967
(`static int count = 0;` need to remove also)

src/framework/global/types/string.cpp
```diff
@@ -110,6 +110,7 @@ char16_t Char::fromAscii(char c)
 
 char Char::toAscii(bool* ok) const
 {
+    UNUSED(ok);
```
igorkorsukov, line 113: fixed here https://github.com/musescore/MuseScore/pull/11967
(`ok` needs to be pass to `toAscii(m_ch)`)

PR #11970, opened 2022-06-10
fixed #8588: Correct staff types in staves after inserting/removing measures

### LINKED ISSUE #8588: [MU4 Issue] Staff size rendered incorrectly after inserting measures
**Describe the bug**
If a staff contains staff size changes, inserting an extra measure before it moves the staff change element but renders the extra measures at the wrong size. On saving/closing/re-opening it renders correctly.
 
**To Reproduce**
1. Create a new score
2. From the Layout palette add a staff change element at measure 2 and specify "cue size" using the inspector
3. Select the first measure and choose Insert One Measure from the Add menu
4. Note that newly added measure is shown at cue size even though the staff change element moved along to the following measure

**Expected behavior**
Newly added measure should render at normal size.

**Screenshots**

 - Windows 10

**Additional context**

Eism:
Resolves: #8588

### DIFF
--- a/src/engraving/libmscore/staff.cpp
+++ b/src/engraving/libmscore/staff.cpp
@@ -1196,6 +1196,17 @@ const StaffType* Staff::staffTypeForElement(const EngravingItem* e) const
     return &_staffTypeList.staffType(e->tick());
 }
 
+bool Staff::isStaffTypeStartFrom(const Fraction& tick) const
+{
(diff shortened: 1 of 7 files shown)


DmitryArefiev:
Tested #8588 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #11971, opened 2022-06-10
[engraving] Replacement QString to String. Step 9

igorkorsukov:


RomanPudashkin (approved):

src/notation/view/widgets/selectdialog.cpp
```diff
@@ -47,18 +48,18 @@ SelectDialog::SelectDialog(QWidget* parent)
...
+        subtype->setText(qtrc("notation", m_element->subtypeName()));
         break;
     case ElementType::SLUR_SEGMENT:
-        subtype->setText(qApp->translate("elementName", m_element->subtypeName().toUtf8()));
+        subtype->setText(qtrc("notation", m_element->subtypeName()));
         break;
     case ElementType::FINGERING:
     case ElementType::STAFF_TEXT:
-        subtype->setText(qApp->translate("TextStyle", m_element->subtypeName().toUtf8()));
+        subtype->setText(qtrc("notation", m_element->subtypeName()));
```
cbjeukendrup, line 62: I'm afraid these new translation contexts are wrong. They need to be the same as where the string is defined. For example, the case of `m_element->typeUserName()`, it must be `"elementName"`, because the string is defined in that context, see EngravingObject::typeUserName().

Jojo-Schmitz: I agree.

PR #11979, opened 2022-06-10
Improvements to articulation placement

mike-spa:
The main observation which prompted this work is that staccato and tenuto marks are laid out way too far when they are on the stem side, which in turn makes slurs look wrong as they are pushed too high:

Screenshot from 2022-06-10 17-52-28

As we looked into it, we found that also ornaments (which share the same code as articulations) are laid out unreasonably far from the staff, which often causes problems with vertical spacing. 

Additionally, the settings in Style->Articulation are obscure and do different things than what they say. For instance, "Stem distance" affects the position of ornaments above the staff even if there is no stem. But it doesn't affect the distance of articulations... from stems.

https://user-images.githubusercontent.com/93707756/173109884-4ee073e9-c89a-410a-a327-3ba26679d900.mp4

Additionally, there are other small bugs such as the ones showed here:

https://user-images.githubusercontent.com/93707756/173107768-3bc8bdf9-1897-4a01-9d98-64b7ffe2d17b.mp4

This PR is an effort to tidy up these aspects and make them workable without doing any major code redesign. We consider this an interim solution, which will do for now.
- A couple of UI options have been renamed and made sure they do what they say.
- A couple of default values have been changed.
- The calculations for articulation positioning have been tweaked to obtain more natural results.
- The bugs have been fixed.
Example after:
Screenshot from 2022-06-10 17-57-16

@oktophonie


oktophonie:
Most vtests looking delicious; a couple need some consideration:

- gonville-4-1
- layout-sequence-2-1

PR #11985, opened 2022-06-11
Fix #8512 support ALT+drag to make copy

### LINKED ISSUE #8512: [MU4 Issue] CTRL+SHIFT+drag to duplicate elements not working
**Describe the bug**
Holding down CTRL & SHIFT and dragging any element doesn't behave as per MU3 (though MU3 seems broken for many element types too)

**To Reproduce**
1. Create a score with pedal marks, dynamics, trill lines etc.
2. Hold down CTRL & SHIFT and try to drag them to different locations, including onto the palette to create a custom palette entry
3. Element doesn't move (specifically, the grips don't move horizontally) and no copy is created when releasing mouse, and no palette entries can be created
4. Create a slur, and try the same thing - element can be moved vertically but nothing else

**Expected behavior**
For me, the expected behaviour is that CTRL+SHIFT+drag should create as close a possible of a copy of whatever element is being copied (or a new palette entry if dragged there).
However MU3's behaviour seems to be that
a) for most lines (hairpin, pedal marks, trills, etc.), CTRL+SHIFT+drag easily allows creating a copy of the element, but the length is not remembered, instead it seems to always be created such that it ends at the end of the current measure.
b) for clefs, time/key signatures, it attempts to duplicate it at the *start* of the dropped measure, regardless of where you drop it
c) for other elements (including accidentals & dynamics, text, slurs etc), CTRL+SHIFT+drag works mostly, but you need to be very very careful what you select as the target (which is tricky because of the 🚫 cursor when you're not quite in the right location, making it hard to determine where the tip of the mouse cursor would be), and for slurs the length and control points are all lost, it just creates a new default-shape slur from the target note to the following note.

However, for v4.0, the decision at this point is that ALT+drag to duplicate elements more or less the way it works in v3 (except that ALT+drag also turns off auto-place, to be discussed...)

**Screenshots**

 - Windows 10, built with MSVC

**Additional context**
My preference would be to fix CTRL+SHIFT+drag to always create the best possible copy for all elements at the exact location dropped, and make it easier to find possible targets, but at a minimum we should restore MU3 behaviour.
Also, the behaviour seems a little odd in MU3 even for the elements that DO work - it only cares about whether CTRL+SHIFT are held down at the point the mouse-down event occurs - you can then release them and the drag/copy behaviour is still active. OTOH, if you first press & hold the left mouse button (but don't move the mouse), then press CTRL+SHIFT, then move the mouse, the behaviour seem to be the same as CTRL+drag, which seems to allow just adjusting horizontal position (without re-anchoring).
The code for MU3 is ScoreView::mousePressEventNormal with the call to cloneElement() - there doesn't seem to be any equivalent code in MU4 yet at all.

NB palette customisation is covered by https://github.com/musescore/MuseScore/issues/11006

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/8512

WIP - just for demonstration, need to confirm ALT+drag behaviour causing both copy and turning off auto-place

### DIFF
--- a/src/appshell/view/navigableappmenumodel.cpp
+++ b/src/appshell/view/navigableappmenumodel.cpp
@@ -221,6 +221,9 @@ bool NavigableAppMenuModel::processEventForOpenedMenu(QEvent* event)
 
 bool NavigableAppMenuModel::processEventForAppMenu(QEvent* event)
 {
+    if (event->type() == QEvent::Type::MouseButtonPress) {
+        m_needActivateHighlight = false;
+    }
     QKeyEvent* keyEvent = dynamic_cast<QKeyEvent*>(event);
     if (!keyEvent) {
         return false;
--- a/src/engraving/libmscore/line.cpp
+++ b/src/engraving/libmscore/line.cpp
@@ -41,7 +41,6 @@
 #include "system.h"
 #include "textline.h"
 #include "utils.h"
-
 #include "log.h"
 
 using namespace mu;
@@ -587,6 +586,24 @@ LineSegment* LineSegment::rebaseAnchor(Grip grip, Segment* newSeg)
     return nullptr;
 }
 
+void LineSegment::handleStaffChange()
+{
+    staff_idx_t oldIdx = spanner()->staffIdx();
+    auto seg = spanner()->startSegment();
+    qreal y = canvasPos().y() - system()->canvasPos().y();
+    auto newIdx = system()->searchStaff(y);
+    if (newIdx != oldIdx) {
+        auto oldY = pagePos().y();
+        spanner()->undoChangeProperty(Pid::TRACK, newIdx * VOICES);
+        for (auto seg : spanner()->spannerSegments()) {
+            seg->setTrack(newIdx * VOICES);
+        }
+        auto newY = pagePos().y();
+        setOffset(PointF(offset().x(), offset().y() + oldY - newY));
+        setOffsetChanged(true);
+    }
+}
+
 //---------------------------------------------------------
 //   rebaseAnchors
 //---------------------------------------------------------
@@ -676,6 +693,10 @@ void LineSegment::rebaseAnchors(EditData& ed, Grip grip)
 
         rebaseAnchor(Grip::START, seg1);
         rebaseAnchor(Grip::END, seg2);
+
+        if (!(ed.modifiers & Qt::KeyboardModifier::ControlModifier) || !!(ed.modifiers & Qt::KeyboardModifier::ShiftModifier)) {
+            handleStaffChange();
+        }
     }
     default:
         break;
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -888,6 +888,16 @@ void NotationInteraction::startDrag(const std::vector<EngravingItem*>& elems,
     qreal proximity = configuration()->selectionProximity() * 0.5f / scaling;
     m_scoreCallbacks.setSelectionProximity(proximity);
 
+    if (elems.size() == 1 && (m_editData.modifiers & Qt::KeyboardModifier::AltModifier)) {
+        if (elems[0]->isSpannerSegment()) {
+            auto copy = toSpannerSegment(elems[0])->spanner()->clone();
+            score()->undoAddElement(copy);
+        } else {
+            auto copy = elems[0]->clone();
+            score()->undoAddElement(copy);
+        }
+    }
+
     if (isGripEditStarted()) {
         m_editData.element->startEditDrag(m_editData);
         return;
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -616,7 +616,9 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
 
             DragMode mode = DragMode::BothXY;
             if (keyState & Qt::ShiftModifier) {
-                mode = DragMode::OnlyY;
+                if (!(keyState & Qt::ControlModifier)) {
+                    mode = DragMode::OnlyY;
+                }
             } else if (keyState & Qt::ControlModifier) {
                 mode = DragMode::OnlyX;
             }
--- a/src/engraving/libmscore/dynamic.cpp
+++ b/src/engraving/libmscore/dynamic.cpp
@@ -444,30 +444,7 @@ std::unique_ptr<ElementGroup> Dynamic::getDragGroup(std::function<bool(const Eng
 
 mu::RectF Dynamic::drag(EditData& ed)
 {
-    RectF f = EngravingItem::drag(ed);
-
-    //
-    // move anchor
-    //
-    KeyboardModifiers km = ed.modifiers;
-    if (km != (ShiftModifier | ControlModifier)) {
-        staff_idx_t si = staffIdx();
-        Segment* seg = segment();
-        score()->dragPosition(canvasPos(), &si, &seg);
-        if (seg != segment() || staffIdx() != si) {
-            const PointF oldOffset = offset();
-            PointF pos1(canvasPos());
-            score()->undo(new ChangeParent(this, seg, si));
-            setOffset(PointF());
-            layout();
-            PointF pos2(canvasPos());
-            const PointF newOffset = pos1 - pos2;
-            setOffset(newOffset);
-            ElementEditDataPtr eed = ed.getData(this);
-            eed->initOffset += newOffset - oldOffset;
-        }
-    }
-    return f;
+    return EngravingItem::drag(ed);
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/engravingitem.cpp
+++ b/src/engraving/libmscore/engravingitem.cpp
@@ -2128,7 +2128,7 @@ void EngravingItem::startDrag(EditData& ed)
     eed->pushProperty(Pid::AUTOPLACE);
     eed->initOffset = offset();
     ed.addData(eed);
-    if (ed.modifiers & AltModifier) {
+    if (ed.modifiers & AltModifier && ed.modifiers & ShiftModifier) {
         setAutoplace(false);
     }
 }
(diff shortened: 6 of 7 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -876,6 +876,16 @@ void NotationInteraction::startDrag(const std::vector<EngravingItem*>& elems,
     qreal proximity = configuration()->selectionProximity() * 0.5f / zoom;
     m_scoreCallbacks.setSelectionProximity(proximity);
 
+    if (elems.size() == 1 && (m_editData.modifiers & Qt::KeyboardModifier::AltModifier)) {
+        if (elems[0]->isSpannerSegment()) {
+            auto copy = toSpannerSegment(elems[0])->spanner()->clone();
+            score()->undoAddElement(copy);
+        } else {
+            auto copy = elems[0]->clone();
+            score()->undoAddElement(copy);
```
wizofaus, line 885: there are some element types (e.g. accidentals, rests) where this doesn't really work properly and might cause a memory-leak or potentially odd behaviour, though from testing I didn't notice anything except that you simply don't see the copied element.

Tantacrul:
It works as expected when alt+dragging one element. However, I ran into some issues when I alt+dragged elements more than once.

https://user-images.githubusercontent.com/21022311/173189310-88fefcff-524d-4951-99d8-4f12cbbe43aa.mov

wizofaus:
So that's because you're not actually dragging the articulation to re-attach to another note, and now you have multiple articulations all belonging to the same note, which it doesn't really know how to render sensibly. As discussed elsewhere, this is the default (and seemingly only) behaviour v4 supports for dragging articulations, and for v3 the normal behaviour unless using ctrl+shift+drag to create a copy.  In fact ALL types of draggable element (except lines and dynamics) have this problem - when you drag them they don't re-attach themselves to the nearest note/rest/measure/etc. - they're still anchored to their original position, and if you delete/clear that measure the element disappears etc.  This is true in v3 too.
I can probably kludge it so that ALT+drag does cause this re-attachment for dragging articulations, but I don't really see why ALT+drag should cause that but not regular drag.

Tantacrul:
Well, it's hairpins in particular that are the real problem (and they'd be the things you'd be more likely to want to copy).

I don't quite know what to suggest to improve them.

I think there's no point having alt+drag to copy articulations if they stay attached to a previous notehead. It serves a different purpose to normal dragging, so it makes sense that it would behave differently.

wizofaus:
But if I'm going to change the behaviour for "copy" drag then I'd really need to change the behaviour for regular drag too (as in, that would keep the code the simplest and the least likely to result in unexpected behaviour).
The hairpin/line dragging is basically the same thing, in that dragging doesn't re-attach to the staff you're dragging to, which makes lines even weirder than I thought - they do auto-reattach to other *measures*, but not to other staves!

Why *wouldn't* we have it that when dragging *anything* by default it logically re-attaches itself to wherever you drag it to?
Providing of course there's a key combo (e.g. CTRL+SHIFT) that can be used to keep the non-reattaching behaviour where that's needed.

wizofaus:
Also, can we move this discussion elsewhere, putting it on the PR seems entirely the wrong place...happy for you to suggest where.  At any rate I think I'll leave off doing more on it until we have some sort of decision as to what behaviour we want.  But personally I'd very much like to see:
a) all draggable elements should behave the same - with no modifier keys held down, or only ALT, while dragging they auto-attach themselves to the nearest sensible anchor (either a ChordRest or a segment), whether that be on the current staff or another one.  But using some modifier key combination (e.g. SHIFT+CTRL), no auto-reattach occurs.
b) the only difference holding down ALT makes is that a copy of the original element is also left in place.
c) to turn off auto-place after dragging both ALT+SHIFT must be held down (or just get rid of this entirely, not clear what advantage if gives)
d) the code to move the keyboard focus to the menu on pressing ALT needs to detect that a mouse button hasn't been pressed in the meantime

Tantacrul:
I don't disagree with your point about dragging. However, I do note that the MS3 implementation is a lot less problematic with copy+dragging Hairpins.

I don't see why we shouldn't just try your suggestion and see if we like it.

wizofaus:
My main concern is that it's going to require a lot of low-level code changes (including in the engraving library), and doing this so close to a beta probably isn't ideal.
A compromise might be to disable copy-dragging of elements like articulations that don't behave sensibly now, and accept the unexpected behaviour for dragging lines (incl. hairpins) to other staves. As it is this doesn't behave nicely in v3 either - it doesn't maintain the correct shape/extent of the line when copy-dragging to other staves.

I'll see how hard it is to fix the staff-reattachment, but I suspect it won't be trivial.

Tantacrul:
OK. I didn't realise it would be so fundamental. In that case, definitely let's not fiddle with dragging behaviour then.

wizofaus:
Well ultimately that's your call - I'm happy to work on it because I think it will be a significant improvement, but it may be best shunted to 4.1.

wizofaus:
Had some success with this - it seems relatively straightforward to have lines (including hairpins) automatically change tracks/staves as you drag them up and down, however there's some recalculations that need to happen to ensure they stay in the right position that I haven't quite figured out yet.  But it's definitely not trivial.

wizofaus:
Ok, I've pushed a change that allows re-anchoring of lines (including hairpins etc.) when dragging vertically over other staves.  To *prevent* this and preserve existing behaviour you need to hold down CTRL+SHIFT while dragging.  I actually think it works quite nicely and can't see at all why it's not the default behaviour.
So if you want to create a copy of an existing hairpin, say, then drag it somewhere so that it stays attached to the current staff, you need to have ALT held down when you *begin* dragging, then you can release ALT (but keep left mouse button down), and then hold down CTRL+SHIFT to continue dragging (or you could just use ALT+CTRL+SHIFT+drag for the whole operation).
Still working on articulations etc.

src/engraving/libmscore/line.cpp
```diff
@@ -587,6 +586,25 @@ LineSegment* LineSegment::rebaseAnchor(Grip grip, Segment* newSeg)
...
+    qreal y = canvasPos().y() - system()->canvasPos().y();
+    auto newIdx = system()->searchStaff(y);
+    if (newIdx != oldIdx) {
+        auto oldY = pagePos().y();
+        spanner()->setTrack(newIdx * VOICES);
+        for (auto seg : spanner()->spannerSegments()) {
+            seg->setTrack(newIdx * VOICES);
+        }
+        auto newY = pagePos().y();
+        _changedPos = PointF(_changedPos.x(), _changedPos.y() + oldY - newY);
```
wizofaus, line 602: maybe could just use setOffsetChanged( )

wizofaus:
I also now have a change I could push that will allow the same for other element types - basically any element that attaches to either a note/rest (e.g. articulations) or that attaches to a segment (all text elements, including dynamics) can now be dragged so that it auto-reparents/re-anchors onto whatever you drag it on to.
It does look a bit odd when you first start dragging because it's trying to show two such element types in the same position, but once you drag it away the copy returns to the expected position.  But needs a bit more testing, as there's likely to be cases that don't work as expected (e.g. it lets you try to attach articulations to rests etc.)

Tantacrul:
Just tested it again. Made one or two points. Let me know what you think.

https://user-images.githubusercontent.com/21022311/173236833-4e138649-2641-4072-b5f2-fa2ec4e38b69.mov

wizofaus:
As for your video - the pending change I have does allow dragging text to re-anchor to other notes/rests, but as I said, it's a bit too sensitive and in particular when dragging left tends to anchor to the last note in the previous measure even when you don't want it to (it's even worse for "Segno" and other repeat marks, which anchor to the start of the measure).
I definitely had this discussion on a musescore.org forum where various existing long-time users argued against it, precisely because of this concern (there seemed to be a belief that in most cases that you'd be dragging elements around, you don't want them to re-anchor). But I still strongly believe it *can* be made to work, it's just about finding the right balance, and accepting that for outlying situations you'll need to hold down additional keys (e.g. CTRL+SHIFT) in order to drag without re-anchoring.

wizofaus:
Ok have pushed another change that allows you to play around with dragging other element types - the ones that definitely should work now are articulations, all text elements and repeat marks (segno etc.).
There's a heuristic in place to prevent "over-eager" re-anchoring that basically means when dragging *left*, it doesn't re-anchor until the right-side of the element moves into the bounding box of the new anchor, and likewise when dragging *right* it doesn't re-anchor until the left-side moves into the bounding box of the new anchor.  This allows dragging text elements etc. around near a barline without it annoyingly shifting between measures, though note for e.g. a Segno symbol on a barline, it can still tend to jump to the previous measure a bit eagerly.  But you can always hold down CTRL+SHIFT while dragging to prevent this, or if you really do want an element to stay anchored where it is regardless of where you drag it to.
Accidentals still don't work as expected.
The one remaining issue I need to fix is that releasing the ALT key after using the mouse to drag should *not* switch keyboard focus to the menu.

src/appshell/view/navigableappmenumodel.cpp
```diff
@@ -221,6 +221,9 @@ bool NavigableAppMenuModel::processEventForOpenedMenu(QEvent* event)
 
 bool NavigableAppMenuModel::processEventForAppMenu(QEvent* event)
 {
+    if (event->type() == QEvent::Type::MouseButtonPress) {
+        m_needActivateHighlight = false;
```
wizofaus, line 225: this is necessary as otherwise after using ALT+drag the keyboard focus switches to the menu

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -640,7 +640,9 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
 
             DragMode mode = DragMode::BothXY;
             if (keyState & Qt::ShiftModifier) {
-                mode = DragMode::OnlyY;
+                if (!(keyState & Qt::ControlModifier)) {
+                    mode = DragMode::OnlyY;
```
wizofaus, line 644: CTRL+SHIFT means "don't re-anchor", don't want to also restrict X movement

Tantacrul:
Yeah, I like this. It works fairly well. I notice a few issues.

- Hairpins on grand staves, like piano are hard to move without constant snapping above and below
- Undo tends to not place items where they originally were (possibly related to a different issue?)

But as a proof of concept, I can't see why we wouldn't arrive at something close to this in 4.X

wizofaus:
FIXED: "Hairpins on grand staves, like piano are hard to move without constant snapping above and below" (it was not specific to grand staves, and I accept that it might still snap to the next or previous staff unexpectedly in some cases, but again, CTRL+SHIFT will avoid this.  The actual bug was that it was jumping around unexpectedly while re-anchoring between staves).

Working on Undo/Redo.

I also found an *existing* weird bug with dragging lines (like hairpins) horizontally, but it's a bit hard to describe. Will play around a bit more.

wizofaus:
Turns out the left/right moving heuristic wasn't really working well, esp. for long text elements that span multiple measures etc.  Fine-tuned it so that it's just moving to the left that's a special case - basically it won't re-anchor when dragging left until the right-side of the element you're moving is to the left of its current parent.  For long text that can mean having to drag left then right to get it to anchor to the measure you want, but at least it will allow you to drag it a long way left without auto-reanchoring, e.g:

image

It will only re-anchor once you move the whole element to the left of its current parent, but you may then need to drag right again to anchor it as expected.

cbjeukendrup:
@Tantacrul @bkunda Since the Ctrl+Shift+Drag functionality has been implemented in https://github.com/musescore/MuseScore/pull/13018 in a similar way to how it worked in MS3, I wonder to what extent this is still actual. From what I read in the commits here, it does seem a nice improvement over the existing functionality and a potentially useful reimplementation.

Tantacrul:
@cbjeukendrup - that looks to have been merged a while back. The PR's I've looked at are expired and I'm also not 100% certain what to test at this point 😄 

What do you recommend?

Tantacrul:
To be more helpful, I'm happy to test a new implementation and offer thoughts. I can see there was some good stuff in this... and I can do some of it in a current version of MS4, although I'm not massively fond of the choice of key combinations, which are non-standard.

The Alt key is generally how one does this in most creation apps, not Command+Shift (on MacOS)

wizofaus:
I'm happy to do the work to resolve the conflicts here, though I haven't tried to build the software for a while!

wizofaus:
Is this something worth trying to clean up and get mergeable @cbjeukendrup ? Otherwise happy to just close it.

cbjeukendrup:
@wizofaus I'm sorry but maybe this one not. It looks like this doesn't add much anymore on top of the current Ctrl+Shift+drag interaction, and the existing implementation of that (which was created based on MS3, after this PR had been opened) seems a bit less risky than this PRs solution. So if we ever want to change the implementation, there should be a good reason to put effort into that, and right now there isn't, afaics.

PR #11988, opened 2022-06-11

Jojo-Schmitz:
reg. 
* unreachable code (C4702)
* 'initializing': conversion from 'size_t' to 'int', possible loss of data (C4267)
* declaration hides class member (C4458)

### DIFF
--- a/src/engraving/libmscore/stafftype.cpp
+++ b/src/engraving/libmscore/stafftype.cpp
@@ -278,7 +278,7 @@ bool StaffType::isSimpleTabStaff() const
         return true;
 
     default:
-        return false;
+        break;
     }
 
     return false;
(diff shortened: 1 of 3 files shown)


src/engraving/libmscore/stafftype.cpp
```diff
@@ -280,8 +280,6 @@ bool StaffType::isSimpleTabStaff() const
     default:
         return false;
     }
-
-    return false;
 }
```
cbjeukendrup, line 283: Removing this makes other compilers emit warnings. In my experience the universal fix is:
```c++
    default: 
        break;
    }

    return false;
}
```

Jojo-Schmitz: Done

Jojo-Schmitz: Doesn't even compile though, needs a `break;` in the `default:` case

cbjeukendrup: I did write `break;` 😉

Jojo-Schmitz: Outch, reading helps...

cbjeukendrup (approved):

Issue #9948, opened 2021-12-01
[MU4 Issue] Timeline still requires horizontal scroll bar (with proper functionality)


DmitryArefiev:
Horizontal scroll bar still needs to be implemented in Timeline and it should has a proper functionality (like in MS3.6)

Demo video:
https://www.dropbox.com/s/7e50gxc8irms83j/timeline-crashes.mov?dl=0 

The UPDATED state of scroll bar in MS4:

https://user-images.githubusercontent.com/90187801/173817182-cc8a9de8-d9ad-4fc5-8d85-9363ba7ff7c3.mp4

**Expected behavior**
- Horizontal scroll bar should be operational at the bottom of the timeline (per MS3.6)
- Scroll bar should work as a scroll bar (see the second video)

PR #11995, opened 2022-06-12
WidgetView: fix updating the size of the contained QWidget

cbjeukendrup:
Resolves: #9948

(The scrollbars were just there, but the size of the Timeline widget was not updated properly, so the scrollbars ended up being out of view.)

### DIFF
--- a/src/framework/ui/view/widgetview.cpp
+++ b/src/framework/ui/view/widgetview.cpp
@@ -125,7 +125,7 @@ QWidget* WidgetView::qWidget() const
 void WidgetView::updateSizeConstraints()
 {
     if (qWidget()) {
-        qWidget()->setMinimumSize(width(), height());
+        qWidget()->setFixedSize(width(), height());
     }
 }
 


RomanPudashkin (approved):

DmitryArefiev:
Tested #9948 on Win10, Mac12, LinuxMint 20.2

@cbjeukendrup I see your fix, now the scroll bar behaves better when resizing timeline vertically . But the **scroll bar still doesn't work as it should** when you drag it. I propose don't close ##9948 (I've edited the bug and added the video with scroll bar's actual state). 

https://user-images.githubusercontent.com/90187801/173818386-d627a8df-62c5-4255-bcad-a8b8f456dba2.mp4

Let's promote your changes to master since it's better now.

Thanks!

cbjeukendrup:
Ah, that's a strange issue. Seems like either QML or QWidgets is thinking that it is a touch based interface instead of mouse based... I'll look at it separately.

Issue #11199, opened 2022-04-14
[MU4 Issue] Mid-score instrument changes are reflected confusingly in mixer


bkunda:
**Describe the bug**
Adding a mid-score instrument change ("To `Instrument`") to a stave results in a new channel strip being created between the metronome and master faders in the mixer. This is very confusing from a UX POV (please see the below video for demo/explanation):

https://www.dropbox.com/s/wrn5wb3oe4129cs/mid-score-instrument-changes-in-mixer.mov?dl=0

**Expected behavior**

1. Channel strips created as a result of a "change instrument" score direction should appear immediately adjacent to their corresponding "main instrument". In the example mockup below, the Violin I stave has had a "To Flute" direction applied. 
2. The channel strip label should reflect the newly-assigned (I.e. "changed") instrument, and this should be placed in parentheses. 
3. The opacity of the channel strip label container should be set to 25%, or half the current opacity of the label background as specified in the current designs (which are currently set at 50%; @jessjwilliamson please confirm if I've misread the design files).

<img width="827" alt="Screenshot 2022-04-14 at 11 27 59 pm" src="https://user-images.githubusercontent.com/86290556/163400831-4f59ad89-aae3-451a-b49f-c75d0cf1c582.png">

**Platform information**
 - All

Issue #10815, opened 2022-03-15
[MU4 Issue] Mixer issues: metronome track position and failed refresh when instruments are deleted


bkunda:
**Describe the bug**
Two (or three) issues here:

1. The metronome track should remain aligned to the immediate left of the master track when new instruments are added. This currently does not take place (i.e. new instruments in the mixer are stacked to the right of the metronome track). 
2. Removing instruments by the `undo` command fails to remove them from the mixer.
3. Removing instruments from the **Instruments** dialog fails to remove them from the mixer. 

Demo video: https://www.dropbox.com/s/44giq806fskwihb/three-mixer-issues.mov?dl=0

**Expected behavior**
As above: 
1. Mixer should remain aligned to immediate left of Master track when new instruments are added. 
2. Tracks should be removed when instruments are removed. 3. 

**Platform information**
 - MacOS 12.1 (M1)

jessjwilliamson:
Absolutely agree!

Tantacrul:
Seems like a P1 candidate to me.

cbjeukendrup:
Working on it as we speak! :)

PR #12009, opened 2022-06-14
Mixer: fix channel ordering issues and fix display of channels from instrument changes

cbjeukendrup:
Resolves: #11199
Resolves: #10815

To do:
- a mysterious crash occurs when adding an instrument change

### DIFF
--- a/src/playback/qml/MuseScore/Playback/internal/MixerSoundSection.qml
+++ b/src/playback/qml/MuseScore/Playback/internal/MixerSoundSection.qml
@@ -47,7 +47,7 @@ MixerPanelSection {
             height: 26
 
             supportsByPassing: false
-            resourceItemModel: channelItem.inputResourceItem
+            resourceItemModel: channelItem.inputResourceItem ?? null
 
             navigationPanel: channelItem.panel
             navigationRowStart: root.navigationRowStart
--- a/src/engraving/libmscore/part.cpp
+++ b/src/engraving/libmscore/part.cpp
@@ -475,6 +475,17 @@ const Instrument* Part::instrument(Fraction tick) const
     return _instruments.instrument(tick.ticks());
 }
 
+const Instrument* Part::instrumentById(const std::string& id) const
+{
+    for (const auto& pair: _instruments) {
+        if (pair.second->id().toStdString() == id) {
+            return pair.second;
+        }
+    }
+
+    return nullptr;
+}
+
 //---------------------------------------------------------
 //   instruments
 //---------------------------------------------------------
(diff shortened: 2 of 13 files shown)


Eism:
@bkunda 
Please check this PR

bkunda:
Tested on MacOS 🟢
(Was unable to install on Windows...kept getting a message "The application was unable to start correctly (0xc0000106)")

On MacOS, everything seems to be working as expected. All issues in #10815 were resolved, and I had success inserting instrument changes too, so that resolves #11199 too.

Eism:
@cbjeukendrup, can you help Bradley with Windows?

cbjeukendrup:
I don't experience any problems if I try this build on Windows... maybe re-downloading helps?

DmitryArefiev:
@cbjeukendrup I tested this on **Windows** and **Linux** - #10815 looks fine, but **with #11199 I had a crashes** in some cases:



https://user-images.githubusercontent.com/90187801/182902889-20ecc3ad-6dd0-4463-abd4-17aff609596a.mp4

Can you try this on your side?

1. Open [audio_Basic.zip](https://github.com/musescore/MuseScore/files/9261559/audio_Basic.zip)
2. Select the 2nd measure in Violin I
3. Change instrument to Clarinet

cbjeukendrup:
@DmitryArefiev Thanks for testing! Yes, I can now also reproduce the crash again. I'm trying to find out what's going on exactly, but it's not obvious so far!

cbjeukendrup:
@DmitryArefiev I changed something that fixes the crash for me and seems to have no side effects (all parameters in the Mixer still update properly if you change them). 

@vpereverzev That change is in the last commit. I'm not very confident that it is correct, but it seems to work. Also, Asan reported:
```
heap-use-after-free channel.h:126 in deto::async::Channel<QAccessibleEvent*>::ChannelInvoker::doInvoke(int, void*, deto::async::NotifyData const&)
```
and the source of this use-after-free seemed to be in the async library when `AsyncMode::AsyncSetRepeat` is used.

src/playback/qml/MuseScore/Playback/internal/MixerSoundSection.qml
```diff
@@ -47,7 +47,7 @@ MixerPanelSection {
             height: 26
 
             supportsByPassing: false
-            resourceItemModel: channelItem.inputResourceItem
+            resourceItemModel: channelItem.inputResourceItem ?? null
```
Eism, line 50: ohh

Eism (approved):

src/playback/qml/MuseScore/Playback/internal/MixerSoundSection.qml
cbjeukendrup: @Eism The bug is fixed in Qt Creator 8 :)

Issue #10316, opened 2022-01-19
[MU4 Task] [GP] Display slides correctly


alexpavlov96:
**Task description**
Slides to/from note are shown a bit big and far from note

**Current state:**
![Screenshot 2022-01-19 at 16 26 15](https://user-images.githubusercontent.com/24373905/150139744-79a1072a-547d-4ca7-97c7-0a035f94e70e.png)

**Expected:**
![Screenshot 2022-01-19 at 16 25 55](https://user-images.githubusercontent.com/24373905/150139710-efbd2372-c465-45dd-8427-a26753bb45df.png)

oktophonie:
I feel like these slides ought to match the positioning of a glissando line between notes a third apart:
![image](https://user-images.githubusercontent.com/20604032/172385268-0d9c9f37-4347-477c-ade8-3502a7370985.png)

The side of the slide that is attached to the note has a fixed position (and moves if the note is moved); this should match that of the glissando. The 'free' end can be manually adjusted, but by default it should match the equivalent vertical position of a glissando line starting a third away from the destination note. The horizontal distance between the two endpoints should also be shorter (I'd suggest 1.5sp).

boryskuzmenko:
🟢 **FIXED** in https://github.com/musescore/MuseScore/pull/12011
Tested on macOS 12, Win 11, Ubuntu 20

alexpavlov96:
Still needs fix for import from guitar pro (remove especially created class "Slide" and use the "ChordLine" for import, but take care of slide for each note in the chord)

mike-spa:
> Still needs fix for import from guitar pro (remove especially created class "Slide" and use the "ChordLine" for import, but take care of slide for each note in the chord) - I'm doing it

**Yes**, I knew this would come up at some point, so I'm actually glad you brought it up. I agree that the `Slide` class should be removed and we should just use the `ChordLine`. The main problem with ChordLine is that, as the name suggests, it can only be attached to chords. Why _on earth_ it was done this way, I honestly have no idea, to me it is completely obvious that these kind of lines should be attached to the individual notes. Now the problem is that in our code base probably the most consequential property of each object is its `parent`. Changing the parent of ChordLine from `Chord` to `Note` requires to change almost every single instance that the ChordLine is used. It is a lot of work, but I already have a working branch where I have done some of that, before giving up. So I can take care of this, if that's ok for you

mike-spa:
[...]

PR #12011, opened 2022-06-14
Improvements to guitar slides and bends

mike-spa:
Resolves: #10316 

- Made all the marks smaller, as they were previously ridiculously large.
- Improved the positioning, which now also clears accidentals and augmentation dots.
- Added appropriate padding for horizontal spacing
- Prevent kerning over notes of the same voice.

Before:
Screenshot from 2022-06-14 13-23-45
After:
Screenshot from 2022-06-14 13-25-39

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -2193,6 +2193,10 @@ void Chord::layoutPitched()
     _spaceLw = lll;
     _spaceRw = rrr;
 
+    for (Note* note : _notes) {
+        note->layout2();
+    }
+
     for (EngravingItem* e : el()) {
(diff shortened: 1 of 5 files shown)


alexpavlov96 (approved):

MarcSabatella:
FWIW, I liked the size of the bends (as opposed to slides) better before.  I suppose there might be different conventions for guitar versus horns, but I'm much more accustomed to falls in particular being at least as large as in MU3 - often larger.  Would be good to compare across some published music in different genres.

MarcSabatella:
You'll also note that Sammy's copyist was fine with having scoops overlap the preceding note, although I could see not wanting that sometimes.  Ideally to me, it would default to not allowing overlapping, but disabling autoplace for the bend would allow it.

Issue #12005, opened 2022-06-14
[MU4 Issue] Crash on clicking on any button in part name's context menu


HemantAntony:
**Describe the bug**
After you right click on a part name, click on any option, you get a crash. Aside from this way of reproducing you don't even need to click on a button. You can just click outside the popup.

**To Reproduce**
1. Right click on a part name
2. Click on any option
3. MuseScore crashes

**Expected behavior**
MuseScore should not crash on clicking button in the context menu

**Screenshots**
This following shows both ways of reproducing the crash

https://user-images.githubusercontent.com/88136054/173555720-78ef0528-2f24-4ab3-b40c-57d6eab6fa97.mp4

**Platform information**
 - OS: Windows 11

**Additional context**
@Tantacrul imho, this is a critical crash

Tantacrul:
Added a P0. Thanks @HemantAntony !

wizofaus:
NB this is only recent, wasn't happening in a build from a few days ago (but reproduced with last night's build).

DmitryArefiev:
> NB this is only recent, wasn't happening in a build from a few days ago (but reproduced with last night's build). Very easy to reproduce in debugger and see cause, related to accessibility changes (selectedElement->accessible() is returning nullptr)
> 
> Most likely fixed by #12013 ?

Yes, confirm the crash in previous master and was not able to repro after merging #12013 to master

PR #12013, opened 2022-06-14
fixed #11984: Note input bar crash fix

### LINKED ISSUE #11984: [MU4 Issue] Toolbar buttons disable for lines and layouts only after hover
**(edited by @Tantacrul)**

**Describe the bug**
When the user selects a line or frame, it appears that the toolbar is supposed to disallow the user from doing certain actions (such as adding a sharp or marcato marking). However, those buttons are still enabled until the user hovers over a button long enough for the tooltip to appear.

Seems to affect at least the following:
* All Lines (Slurs, Crescendos, Trills, etc.)
* Barlines (Repeats, Normal Barlines, etc.)
* Vertical / Horizonal Frames

**To Reproduce**
1. Create a new score with a piano (should work for any instrument).
2. Add a crescendo hairpin to the rest in measure 1.
3. Select the crescendo hairpin
4. Hover over the voice 2 button in the toolbar (or any button that's not tuplet or flip direction).
5. Wait until the tooltip appears.
6. Note that most of the toolbar buttons are now disabled (can't be clicked).

**Crash details**
On MacOS, select a crescendo line and hover over Voice2 to produce a 100% reproducible crash.

https://user-images.githubusercontent.com/21022311/173198369-cc203a75-53a4-4ed5-b926-cd5cfbfb5b3c.mov

**Expected behavior**
- If the intention is for the buttons to be disabled, then it should happen immediately
- Otherwise, the buttons should not become disabled at all (should just do nothing)

**Screencast**

https://user-images.githubusercontent.com/2625852/173157563-24a2ab1c-33c5-4ae0-a014-5ec8f5cc4121.mp4

**Platform information**
 - OS: Windows
 - Build: Nightly MuseScoreNightly-202206090431-master-8a3bf7c-x86_64

Eism:
Resolves: #11984

### DIFF
--- a/src/notation/internal/notationaccessibility.cpp
+++ b/src/notation/internal/notationaccessibility.cpp
@@ -83,6 +83,7 @@ void NotationAccessibility::setEnabled(bool enabled)
     };
 
     EngravingItem* selectedElement = selection()->element();
+    AccessibleItem* selectedElementAccItem = selectedElement ? selectedElement->accessible() : nullptr;
 
     for (AccessibleRoot* root : roots) {
         root->setEnabled(enabled);
@@ -92,8 +93,12 @@ void NotationAccessibility::setEnabled(bool enabled)
             continue;
         }
 
-        if (selectedElement && selectedElement->accessible()->accessibleRoot() == root) {
-            root->setFocusedElement(selectedElement->accessible());
+        if (!selectedElementAccItem) {
+            continue;
+        }
+
+        if (selectedElementAccItem->accessibleRoot() == root) {
+            root->setFocusedElement(selectedElementAccItem);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/notation/internal/notationaccessibility.cpp
```diff
@@ -92,8 +93,12 @@ void NotationAccessibility::setEnabled(bool enabled)
             continue;
         }
 
-        if (selectedElement && selectedElement->accessible()->accessibleRoot() == root) {
-            root->setFocusedElement(selectedElement->accessible());
+        if (!selectedElementAccItem) {
```
wizofaus, line 96: confirmed that this fixes quite a large number of crashes!

Tantacrul:
Just tested on MacOS. This fixes my issue logged here: https://github.com/musescore/MuseScore/issues/12014

HemantAntony:
> Just tested on MacOS. This fixes my issue logged here: #12014

Hey, can you test #12005 too? There is a chance this might solve that too (@wizofaus)

RomanPudashkin (approved):

DmitryArefiev:
Tested #12005 #12014 #11984  in master build (after merging #11984) on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Issue #10965, opened 2022-03-30
[MU4 Issue] Windows 11 Snap layouts not shown in MU4


GabeS573:
**Describe the bug**
When hovering over the Maximize button in the toolbar of MuseScore, the snap layouts are not displayed.

Also, if MuseScore is snapped next to another app, it often is too large to fit the available space, probably due to the minimum size of the app window. See screen recording below.

**To Reproduce**
1. Open MuseScore
2. Hover over the Maximize button (do not click on it)
3. The snap layouts do not appear.
4. Open another app (i.e. notepad) and snap it to the left of the screen using the snap layouts.
5. When shown which other apps you can snap to the right side of screen, click on MuseScore
6. Depending on your screen size, MuseScore may be too large to fit the space.

**Expected behavior**
Snap layouts should be available when hovering over the maximize button, just like they were in MuseScore 3. Also, when snapped, MuseScore should fit the available space like it did in MuseScore 3 (screenshot below).

**Screenshots**

Video:

https://user-images.githubusercontent.com/90187801/160885524-1e9b2bc5-fe99-4f6c-bbe7-26a80e090f46.mp4

MuseScore 3:

![Screenshot 2022-03-29 203455](https://user-images.githubusercontent.com/91428485/160746324-9d090e2e-0fae-419a-88d7-7c1a9ef86bd9.jpg)

MuseScore 4:

![Screenshot 2022-03-29 203430](https://user-images.githubusercontent.com/91428485/160746285-cfb4264c-34be-496d-85b5-1dca8f0dabf6.jpg)

**Platform information**
 - OS: Windows 11

**Additional context**
As noted above, snap layouts worked perfectly with MuseScore 3. For the issue of the window not being small enough to snap correctly next to other apps, this may be because MuseScore 3 had a very small minimum window size.

GabeS573:
For reference, here are the minimum sizes of MuseScore 3 and MuseScore 4:

![Screenshot 2022-03-29 203911](https://user-images.githubusercontent.com/91428485/160746724-b263ac83-053e-4b4a-962d-4e92f676cb82.jpg)

DmitryArefiev:
Thanks for logging!

(I converted the attached video and re-added to the description to be able to watch it inside web browser )

Tantacrul:
@cbjeukendrup - can't this simply be fixed by lowering the minimum size of MS4? I seem to recall we discussed this on another issue.

cbjeukendrup:
[...]

PR #12015, opened 2022-06-14
Dynamically calculate the minimum size of the dock window

cbjeukendrup:
Resolves: #10965 (kind of)

This makes sure that the minimum size of the window is always as low as it can be according to the calculations made by the docking system, based on the current minimum sizes of all docked dock widgets. For example, if you undock the note input toolbar, which has a large minimum width, then the minimum width of the window becomes lower. If you re-dock it, then the window will adjust its size. By undocking both the note input bar and the playback toolbar, you can get a pretty low minimum width. I'd say that counts as a solution for #10965, at least for the time being.

The advantage is that this eliminates some problems that arise when users use different text sizes or when translated strings take more space. For example, this causes the "Home/Score/Publish" toolbar to become wider, which might make the required minimum width larger than a hard-coded minimum width, so that the docking system gets crazy. We avoid this situation by letting the dock system determine the minimum size.

This PR also solves the problems that arose in PR #10207, so after merging this, we can just set the minimum height of the dock panels to 200px and it will be fine.

### DIFF
--- a/src/appshell/qml/AppWindow.qml
+++ b/src/appshell/qml/AppWindow.qml
@@ -35,11 +35,14 @@ ApplicationWindow {
 
     title: titleProvider.title
 
-    width: 1150
-    height: 800
+    // Minimum width and height are set in Main.qml files
+    onMinimumWidthChanged: {
+        width = Math.max(width, minimumWidth)
+    }
 
-    minimumWidth: 1150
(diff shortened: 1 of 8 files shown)


cbjeukendrup:
Closing this for now, since it was decided that this is probably too risky for 4.0. I'll save it somewhere for later.

PR #12029, opened 2022-06-15
fixed #11998: Fixed undefined behavior when calling a virtual method from a destructor

### LINKED ISSUE #11998: [MU4 Issue] Score::sigmap() can enter infinite loop and/or eventually stack overflow
**Describe the bug**
Has happened various times now while working on MuseScore that after certain operations (esp. save then close), Score::sigmap() enters infinite recursion because _masterScore is a partly destructed MasterScore, which should never happen.
This causes the app to freeze up and usually crash eventually.

**To Reproduce**

Unknown at this point, will update when I have something more concrete, but just now it happened when saving then closing the score.

**Expected behavior**
App should not freeze-up or crash

**Screenshots**

**Platform information**
Windows 10, running on Visual Studio, RelWithDebugInfo build from latest code

**Additional context**
Example stack-trace of this being hit on closing the score

       etc.
 	MuseScore4.exe!mu::engraving::Score::sigmap() Line 5624	C++
 	MuseScore4.exe!mu::engraving::Score::sigmap() Line 5624	C++
 	MuseScore4.exe!mu::engraving::Score::sigmap() Line 5624	C++
 	MuseScore4.exe!mu::engraving::Score::sigmap() Line 5624	C++
 	MuseScore4.exe!mu::engraving::EngravingItem::barbeat() Line 2504	C++
 	MuseScore4.exe!mu::engraving::EngravingItem::formatBarsAndBeats() Line 2849	C++
 	MuseScore4.exe!mu::engraving::AccessibleItem::accessibleName() Line 157	C++
 	MuseScore4.exe!mu::accessibility::AccessibilityController::stateChanged(mu::accessibility::IAccessible * aitem, mu::accessibility::IAccessible::State state, bool arg) Line 229	C++
 	MuseScore4.exe!deto::async::AbstractInvoker::invoke(int type, const deto::async::NotifyData & data) Line 46	C++
 	MuseScore4.exe!deto::async::Channel<enum mu::accessibility::IAccessible::State,bool>::send(const mu::accessibility::IAccessible::State & <d_0>, const bool & <d_1>) Line 32	C++
 	MuseScore4.exe!mu::engraving::AccessibleItem::notifyAboutFocus(bool focused) Line 103	C++
 	MuseScore4.exe!mu::engraving::AccessibleRoot::setFocusedElement(mu::engraving::AccessibleItem * e) Line 61	C++
 	MuseScore4.exe!mu::engraving::AccessibleItem::~AccessibleItem() Line 48	C++
...
>	MuseScore4.exe!mu::engraving::EngravingProject::~EngravingProject() Line 51	C++

```
EngravingProject::~EngravingProject()
{
    delete m_masterScore;
}
```
Which I'd say explains it - the MasterScore is partly deleted, so the virtual table is pointing to Score::sigmap.

Not sure what the safest fix is - but it seems to me having AccessibleItem::~AccessibleItem() call `root->setFocusedElement( )` is rather dangerous, or at least should pass in a flag indicating it's because the element is being destructed and hence notifyAboutFocus( ) shouldn't be called.
Note however a logging statement is part of the stack trace:

`LOGE() << "for item: " << aitem->accessibleName() << " parent is null";`

Eism:
Reproduced if add the define ACCESSIBILITY_LOGGING_ENABLED to accessibilitycontroller.cpp

AccessibleItem destructor -> AccessibilityController::unreg -> AccessibleItem::name -> EngravingItem::formatBarsAndBeats -> EngravingItem::barbeat -> Score::sigmap (virtual)

Resolves: #11998

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -40,6 +40,8 @@ AccessibleItem::AccessibleItem(EngravingItem* e, Role role)
 
 AccessibleItem::~AccessibleItem()
 {
+    m_element = nullptr;
+
     AccessibleRoot* root = accessibleRoot();
     if (root && root->focusedElement() == this) {
         root->setFocusedElement(nullptr);
@@ -49,8 +51,6 @@ AccessibleItem::~AccessibleItem()
         accessibilityController()->unreg(this);
         m_registred = false;
     }
-
-    m_element = nullptr;
 }
 
 AccessibleItem* AccessibleItem::clone(EngravingItem* e) const
@@ -104,6 +104,10 @@ void AccessibleItem::notifyAboutFocus(bool focused)
 
 const IAccessible* AccessibleItem::accessibleParent() const
 {
+    if (!m_element) {
+        return nullptr;
+    }
+
     EngravingObject* p = m_element->parent();
     if (!p || !p->isEngravingItem()) {
         return nullptr;
@@ -115,6 +119,11 @@ const IAccessible* AccessibleItem::accessibleParent() const
 size_t AccessibleItem::accessibleChildCount() const
 {
     TRACEFUNC;
+
+    if (!m_element) {
+        return 0;
+    }
+
     size_t count = 0;
     for (const EngravingObject* obj : m_element->children()) {
         if (obj->isEngravingItem()) {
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -49,8 +51,6 @@ AccessibleItem::~AccessibleItem()
         accessibilityController()->unreg(this);
```
wizofaus, line 51: there is *still* undefined behaviour here, because unreg calls a virtual function on the passed-in argument
@Eism 
But...it never gets executed because accessibleRoot( ) will always return nullptr after setting m_element to nullptr, hence this code is pointless...

PR #12056, opened 2022-06-16
[MU4] vst_scanning

vpereverzev:


src/framework/vst/internal/vstmodulesmetaregister.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+bool VstModulesMetaRegister::isEmpty() const
+{
+    return m_metaMap.empty() && m_paths.empty();
+}
+
+void VstModulesMetaRegister::readMetaList()
+{
+    m_file.open(io::IODevice::OpenMode::ReadWrite);
```
RomanPudashkin, line 138: IODevice::OpenMode::ReadOnly would be better here

src/framework/vst/internal/vstconfiguration.h
```diff
@@ -23,17 +23,22 @@
...
 class VstConfiguration : public IVstConfiguration
 {
+    INJECT(vst, framework::IGlobalConfiguration, globalConfig)
 public:
     void init();
 
     io::paths_t userVstDirectories() const override;
     void setUserVstDirectories(const io::paths_t& paths) override;
     async::Channel<io::paths_t> userVstDirectoriesChanged() const override;
+    io::path_t knownPluginsList() const override;
```
RomanPudashkin, line 41: What do you think about knownPluginsFile?

src/framework/vst/internal/vstmodulesmetaregister.cpp
RomanPudashkin: Hmm, now I understand why it's necessary

RomanPudashkin (approved):

src/framework/vst/internal/vstconfiguration.h
vpereverzev: actually, forgot to rename it

Issue #12047, opened 2022-06-16
[MU4 Issue] MuseScore_amd64 failed to run "global_test" on release and debug configuration with MSVC on windows. 


rachelhe123:
**Describe the bug**
MuseScore_amd64 failed to run "global_test" on release and debug configuration with MSVC on windows. Could you please help look at this issue or provide some workarounds? Thanks in advance.

**To Reproduce**
1. git clone https://github.com/musescore/MuseScore F:\gitP\musescore\MuseScore
2. mkdir F:\gitP\musescore\MuseScore\build_amd64
3. cd F:\gitP\musescore\MuseScore\build_amd64
4. cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 -DBUILD_UNIT_TESTS=ON -DCMAKE_BUILD_TYPE=Release .. 
5. msbuild /m /p:Platform=x64 /p:Configuration=Release mscore.sln /t:Rebuild 
6. ctest --build-config Release --output-on-failure -E "engraving_utests|iex_braille_tests|iex_capella_tests|iex_musicxml_tests"

**Expected behavior**
For MuseScore_amd64, "global_test" test should be run pass on the windows.

**Error Message**
[----------] Global test environment tear-down
[==========] 55 tests from 9 test suites ran. (791 ms total)
[  PASSED  ] 51 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] Global_IO_IODeviceTests.Open_ReadOnly
[  FAILED  ] Global_IO_IODeviceTests.Open_WriteOnly
[  FAILED  ] Global_IO_FileInfoTests.IsRelative
[  FAILED  ] Global_IO_FileInfoTests.IsAbsolut

**Platform information**
 OS: Windows

**Setup**
VS 2019 + Windows Server 2019

**Additional context**
test log
[test.log](https://github.com/musescore/MuseScore/files/8916366/test.log)
[test.log.3.out.txt](https://github.com/musescore/MuseScore/files/8916368/test.log.3.out.txt)

Tantacrul:
@cbjeukendrup - since I can see a PR from you, I've assigned you. Please close when finished.

cbjeukendrup:
@igorkorsukov Do you agree with my previous comment? If so, we can close this.

PR #12061, opened 2022-06-16
Fix FileInfo tests on Windows

cbjeukendrup:
- fix unsupported unescaped () in regex
- fix FileInfo::relative and FileInfo::absolute

Note: these changes fix the tests, but I'm not sure if they are 'universally correct' for Windows.

Resolves: #12047

### DIFF
--- a/src/framework/global/io/fileinfo.cpp
+++ b/src/framework/global/io/fileinfo.cpp
@@ -145,26 +145,6 @@ QDateTime FileInfo::lastModified() const
     return fileSystem()->lastModified(m_filePath);
 }
 
-#if defined(Q_OS_WIN)
-bool FileInfo::isRelative() const
-{
-    return m_filePath.isEmpty()
-           || (m_filePath.at(0) != '/'
-               && !(m_filePath.length() >= 2 && m_filePath.at(1) == ':'));
-}
-
-bool FileInfo::isAbsolute() const
-{
-    return (m_filePath.length() >= 3
-            && m_filePath.at(0).isLetter()
-            && m_filePath.at(1) == ':'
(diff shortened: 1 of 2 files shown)


igorkorsukov (approved):

src/framework/global/io/fileinfo.cpp
```diff
@@ -145,37 +145,25 @@ QDateTime FileInfo::lastModified() const
...
 }
 
-#if defined(Q_OS_WIN)
 bool FileInfo::isRelative() const
 {
-    return m_filePath.isEmpty()
-           || (m_filePath.at(0) != '/'
-               && !(m_filePath.length() >= 2 && m_filePath.at(1) == ':'));
+    return !isAbsolute();
 }
```
wizofaus, line 151: not sure if we need to support unc paths etc., then there's the "\\?\..." convention...are all pathnames in MuseScore passed around using forward slash?

src/framework/global/io/fileinfo.cpp
```diff
@@ -145,37 +145,25 @@ QDateTime FileInfo::lastModified() const
...
             && m_filePath.at(0).isLetter()
             && m_filePath.at(1) == ':'
             && m_filePath.at(2) == '/')
-           || (m_filePath.length() >= 2
-               && m_filePath.at(0) == '/'
-               && m_filePath.at(1) == '/');
-}
-
+           || (!m_filePath.isEmpty()
+               && (m_filePath.at(0) == '/'
```
wizofaus, line 161: A path starting with \ is relative in windows - but relative to the current *drive*, not the current directory.
Whereas \\?\c:\xyz is absolute, as is \\hostname\share\

PR #12062, opened 2022-06-16
[MU4] Fix #11919 - Beam options changing number of beams

### LINKED ISSUE #11919: [MU4 Issue] Beaming options in Properties can make 32nd beam look like 16th
**Describe the bug**
Using the beaming options in the Properties panel, you can make beamed 32nd notes look like 16th notes.

**To Reproduce**
1. Create a score with 32nd notes
2. Select their beam
3. In the properties panel, click the "begin 32" beaming option (it has the wrong icon/tooltip by the way, see #9977)

Result: the notes look like 16th notes

**Screenshots**

https://user-images.githubusercontent.com/48658420/172494542-9bc35fa6-673b-49c2-8583-4775974394e4.mov

**Expected behavior**
This situation should not be possible. 

I think we should reimagine the beaming options to make something that's both comprehensive and comprehensible and always produces valid results. 
The current situation feels like we are providing some random options, which together after some fiddling enable you to create the situation that you want if you're working with 8th or 16th notes. 
Instead, I think we should consider a WYSIWYG control that lets you choose for each beam level whether the note should be connected to the notes at the left and right sides. If you disconnect both sides, you will get a "fractional beam"/beamlet or flag, otherwise a connecting beam will be created as you specify. 
For now, if you disconnect the primary beam, the secondary etc. should be disconnected too, but in the future we can drop this restriction, making it possible to create unconventional beamings.
(These are just musings though, probably we'll have to stick with what we have now until after 4.0.)

**Platform information**
 - OS: reported on macOS

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/11919

Originally, there was a bit of a bug when all notes are Start32 (which I guess is really Start16) that made the loop terminate early if all the notes required beamlets instead of beams. Fixed:
image
image


asattely:
Note that the first two notes seem to have a solid beam between them, but those are just 'before' and 'after' beamlets that are colliding into each other:
image

cbjeukendrup:
Looks like something went wrong here with rebasing...

asattely:
> Looks like something went wrong here with rebasing...

Whoa it sure did. Fixed! I wonder how that happened.

Issue #12028, opened 2022-06-15
[MU4 Issue] When opening a non-existing (deleted) file, you get the corruption dialog instead of missing file dialog


DmitryArefiev:
**To Reproduce**
1. Open score A from a Flash drive
2. Close score A
3. Eject Flash drive
4. Try to open score A from Recent scores - **you get the corruption dialog instead of missing file dialog**

### Actual result:

![Capture](https://user-images.githubusercontent.com/90187801/224660204-0f31815b-376a-4a18-8507-b872bf6321df.PNG)

### Previous actual result:

**MU shows "File open error" instead of "File not found" message** 

![Capture2](https://user-images.githubusercontent.com/90187801/224660268-ea311f24-a913-47e9-91ca-0abdbdfbacea.png)

**Expected behavior**
MU should show message about the file is missing

**Platform information**
macOS, Windows, Linux

**Additional context**
#8884

DmitryArefiev:
@Tantacrul Putting to Beta2 with P3 but I think can be moved to 4.x.x as well

cbjeukendrup:
It looks like this issue has become a bit more strange. Now, when opening a non-existing file, you get the corruption dialog! When you click "Show details", you see "Error opening the file".

DmitryArefiev:
Yes, the same on my side:

![Capture](https://user-images.githubusercontent.com/90187801/220156894-380feab9-484c-4c1d-b967-c017786134e4.PNG)

FYI @Eism

PR #12067, opened 2022-06-17
[MU4] Fix musescore#12028: Fix messaging when opening non-existenting scores

iwoithe:
Resolves: #12028


src/engraving/infrastructure/io/mscreader.cpp
```diff
@@ -28,6 +28,8 @@
 #include "io/fileinfo.h"
 #include "serialization/zipreader.h"
 #include "serialization/xmlstreamreader.h"
+#include "engraving/engravingerrors.h"
+#include "libmscore/masterscore.h"
```
wizofaus, line 32: you'll need something like this

```
namespace mu::engraving {
    static inline bool operator!=(const char* a, const AsciiStringView b)  { 
        return ::operator!=(a, b); 
    }
}
```
to avoid the name hiding problem with operator!= here.

Jojo-Schmitz:
The unit tests fail with an assert(): https://github.com/musescore/MuseScore/runs/6932119546?check_suite_focus=true#step:8:752

src/engraving/infrastructure/io/mscreader.cpp
```diff
@@ -426,13 +438,13 @@ StringList MscReader::XmlFileReader::fileList() const
     m_device->seek(0);
     XmlStreamReader xml(m_device);
     while (xml.readNextStartElement()) {
-        if ("files" != xml.name()) {
+        if (xml.name() != "files") {
```
Jojo-Schmitz, line 441: why this (and the following)?

src/engraving/infrastructure/io/mscreader.cpp
```diff
@@ -244,24 +246,29 @@ MscReader::ZipFileReader::~ZipFileReader()
...
 {
     m_device = device;
     if (!m_device) {
         m_device = new File(filePath);
         m_selfDeviceOwner = true;
     }
 
+    if (!FileInfo::exists(filePath)) {
+        LOGD() << "not exists path: " << filePath;
+        return scoreFileErrorToRet(Score::FileError::FILE_NOT_FOUND, filePath);
```
wizofaus, line 259: this is wrong in the case that the filePath is just a dummy name (for logging and whatever other purposes), but the actual data passed in is via the "device" argument, which is why one of the unit tests is failing

src/engraving/infrastructure/io/mscreader.cpp
wizofaus: well a) there's no good reason to put it in that order in the first place, as it reads very unnaturally, and prevents nothing, and b) it wasn't compiling due to a name hiding problem with operator != (though I'm not 100% certain why swapping the args resolves this)

Jojo-Schmitz: It did work before the PR though, didn't it?

wizofaus: yes but a header file was introduced that introduced an operator!= in the mu::engraving namespace, hence the name hiding.

src/engraving/infrastructure/io/mscreader.cpp
iwoithe: What is the preferred solution to this?

wizofaus: If you move this inside the !m_device check it might work.  Otherwise to the calling function.

iwoithe:
Closing due to a mistake I made when pushing some changes. Will open a new PR.

PR #12077, opened 2022-06-17
[MU4] Fix compiler warnings

Jojo-Schmitz:


src/engraving/libmscore/lyrics.cpp
```diff
@@ -319,8 +319,8 @@ void Lyrics::layout()
 //                   LOGD("create leading, trailing <%s> -- <%s><%s>", qPrintable(text), qPrintable(leading), qPrintable(trailing));
             const TextBlock& tb = textBlock(0);
 
-            const qreal leadingWidth = tb.xpos(leading.size(), this) - tb.boundingRect().x();
-            const int trailingPos = text.size() - trailing.size();
+            const qreal leadingWidth = tb.xpos(static_cast<int>(leading.size()), this) - tb.boundingRect().x();
+            const int trailingPos = static_cast<int>(text.size() - trailing.size());
```
cbjeukendrup, line 323: Looks like it's better to update TextBlock::xpos to use size_t, doesn't it?

Jojo-Schmitz: Probably, but that'd be for another PR?

Jojo-Schmitz: Or another commit ;-)

src/engraving/libmscore/lyrics.cpp
```diff
@@ -320,7 +320,7 @@ void Lyrics::layout()
             const TextBlock& tb = textBlock(0);
 
             const qreal leadingWidth = tb.xpos(leading.size(), this) - tb.boundingRect().x();
-            const int trailingPos = text.size() - trailing.size();
+            const int trailingPos = static_cast<int>(text.size() - trailing.size());
```
cbjeukendrup, line 323: That makes this cast unnecessary (change the variable type to size_t)

src/engraving/libmscore/textedit.cpp
```diff
@@ -589,7 +589,7 @@ bool TextBase::edit(EditData& ed)
         score()->undo(new InsertText(_cursor, s), &ed);
 
         int startPosition = cursor->currentPosition();
-        notifyAboutTextInserted(startPosition, startPosition + s.size(), s);
+        notifyAboutTextInserted(startPosition, startPosition + static_cast<int>(s.size()), s);
```
cbjeukendrup, line 592: Maybe the same can be done here with `notifyAboutTextInserted`.

Jojo-Schmitz: Not today anymore

cbjeukendrup: Fair enough, seems like that would involve larger changes...

Jojo-Schmitz: That's just one reason, the other is that I have a life too 😉

PR #12080, opened 2022-06-17
fixed #12074: Added check for crash prevention

### LINKED ISSUE #12074: [MU4 Issue] Сrash when moving bracket
**To Reproduce**
1. Open a score

**Platform information**
macOS, Windows, Linux

**Additional context**
Recent builds

Eism:
Resolves: #12074

### DIFF
--- a/src/engraving/accessibility/accessibleroot.cpp
+++ b/src/engraving/accessibility/accessibleroot.cpp
@@ -119,7 +119,7 @@ void AccessibleRoot::updateStaffInfo(const AccessibleItem* newAccessibleItem, co
 
     if (newAccessibleItem && newAccessibleItem->element()->hasStaff()) {
         staff_idx_t newStaffIdx = newAccessibleItem->element()->staffIdx();
-        staff_idx_t oldStaffIdx = oldAccessibleItem ? oldAccessibleItem->element()->staffIdx() : nidx;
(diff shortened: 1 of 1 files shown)


src/engraving/accessibility/accessibleroot.cpp
```diff
@@ -119,7 +119,7 @@ void AccessibleRoot::updateStaffInfo(const AccessibleItem* newAccessibleItem, co
 
     if (newAccessibleItem && newAccessibleItem->element()->hasStaff()) {
         staff_idx_t newStaffIdx = newAccessibleItem->element()->staffIdx();
-        staff_idx_t oldStaffIdx = oldAccessibleItem ? oldAccessibleItem->element()->staffIdx() : nidx;
+        staff_idx_t oldStaffIdx = oldAccessibleItem && oldAccessibleItem->registered() ? oldAccessibleItem->element()->staffIdx() : nidx;
```
wizofaus, line 122: ok I put up a fix that's almost the same. But I think it's wrong - the problem is that AccessibilityRoot::m_focusedElement is often referring to an already deleted object, which we should ensure doesn't happen using some sort of smart pointer.

RomanPudashkin:
See https://github.com/musescore/MuseScore/pull/12101

PR #12081, opened 2022-06-17
Fix #12074 prevent crash when dragging certain elements (e.g. brackets)

### LINKED ISSUE #12074: [MU4 Issue] Сrash when moving bracket

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/12074

### DIFF
--- a/src/engraving/accessibility/accessibleroot.cpp
+++ b/src/engraving/accessibility/accessibleroot.cpp
@@ -119,7 +119,7 @@ void AccessibleRoot::updateStaffInfo(const AccessibleItem* newAccessibleItem, co
 
     if (newAccessibleItem && newAccessibleItem->element()->hasStaff()) {
         staff_idx_t newStaffIdx = newAccessibleItem->element()->staffIdx();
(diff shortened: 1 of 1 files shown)


src/engraving/accessibility/accessibleroot.cpp
```diff
@@ -119,7 +119,7 @@ void AccessibleRoot::updateStaffInfo(const AccessibleItem* newAccessibleItem, co
 
     if (newAccessibleItem && newAccessibleItem->element()->hasStaff()) {
```
wizofaus, line 120: It does concern me that `newAccessibleItem->element()` is never nullptr (from what I observed) but  `oldAccessibleItem->element()` can be - nor sure in what cases AssessibleItem::element() should be returning null...

cbjeukendrup:
Hm... this is _almost_ an exact copy of #12080!

RomanPudashkin:
Here is a more reliable fix for this crash (and all similar crashes in the future):
https://github.com/musescore/MuseScore/pull/12101

Issue #11792, opened 2022-05-27
'Feathered Beams' option has disappeared from the Inspector


Tantacrul:
**Describe the bug**
The feathered beams option that we designed for MS4 (and which was working) has gone missing. It's purpose was to accomplish these effects:
<img width="822" alt="image" src="https://user-images.githubusercontent.com/21022311/170777767-f90c751e-7a3c-4b4b-8fbf-0a447f680fc5.png">

Here is an example of the design in Figma (in the 'Inspector' page)
<img width="738" alt="image" src="https://user-images.githubusercontent.com/21022311/170776592-23ffad7c-d1ea-47c6-8705-33e009d1a7f7.png">

The idea is this:

This button sets: **Grow left = 1' / Grow right = 0** on a selected beam (which can be accomplished in MS3)
<img width="183" alt="image" src="https://user-images.githubusercontent.com/21022311/170777131-e661527e-bb58-479d-ba1c-056970ea31a4.png">

This button sets: **Grow left = 0' / Grow right = 1** on a selected beam (which can be accomplished in MS3)
<img width="203" alt="image" src="https://user-images.githubusercontent.com/21022311/170777248-71c08d60-4f66-48e4-b11d-1923439b1532.png">

This is not new functionality, so should be very simple to implement (replacing the text boxes in MS3). If there is any confusion, just contact me and I'll take you through it.

RomanPudashkin:
I'm working on it

Tantacrul:
@RomanPudashkin - I suspect that this might be blocked behind @asattely's work on the beaming code. Perhaps he may want to have a say in this too. I'm unsure.

cbjeukendrup:
@Tantacrul I see that currently the wrong icons are used:
<img width="299" alt="Schermafbeelding 2022-06-08 om 22 56 57" src="https://user-images.githubusercontent.com/48658420/172715851-31d4b1a9-843e-49af-8641-e1008e161d1b.png">
I will fix that. 

I was also thinking about the screen reader text and tooltip (in palettes). Currently the inspector uses "Left" and "Right", but that might be confusing and not for everyone intuitive. The palette uses "slower" and "faster". But I think even better would be "Decelerate" and "Accelerate". Do you agree?

cbjeukendrup:
Pushed a fix to #11934!

Now the next problem becomes visible: the correct icons are not correctly centered in their bounding box:
[...]

PR #12082, opened 2022-06-17
Fix Feathered Beams

cbjeukendrup:
Resolves: #11792 

This fixes the engraving of Feathered Beams.
Schermafbeelding 2022-06-18 om 00 00 53

It also takes slightly less normal cases into account:
Schermafbeelding 2022-06-18 om 00 01 56
I don't think this makes much sense musically, but at least nothing weirder than this happens. 

In the case that feathering is applied to a beam where subbeams are broken, or where there are subbeams that do not span the whole beam, the slope of those subbeams will be 'interpolated' as if they do span the whole beam. We could make a different choice here, but that does not seem to give much nicer results:
<details><summary>Screenshot of alternative considered:</summary>
Schermafbeelding 2022-06-18 om 00 18 52

</details>


Jojo-Schmitz:
There's one unit test failing (on top of the most probably expected vtests failures), https://github.com/musescore/MuseScore/runs/6975545469?check_suite_focus=true#step:8:790

oktophonie (approved):

RomanPudashkin (approved):

asattely:
> There's one unit test failing (on top of the most probably expected vtests failures), https://github.com/musescore/MuseScore/runs/6975545469?check_suite_focus=true#step:8:790

That one is failing due to an issue that is fixed in https://github.com/musescore/MuseScore/pull/11788

As soon as it's merged, a rebase will fix the test.

Issue #9037, opened 2021-09-06
[Inspector] "Force Horizontal" and "Beam height" not working / reflecting beam height


wizofaus:
**Describe the bug**
Force horizontal and beam height text fields not working in the inspector.

https://user-images.githubusercontent.com/21022311/167021190-d23e51fc-02c6-442c-b039-d74f76ee07dc.mov

wizofaus:
Pretty sure this is actually at least partly my fault, I'd changed the internal representation of 'beamPos' to `QPair<float,float>` but missed a couple of spots where it's still using PointF.

Tantacrul:
@oktophonie - I've assigned this to you for now and given it a provisional P2. Feel free to reassign, reprioritise or close as needed.

Tantacrul:
Uploading 01_Beam_Height_Settings_Not_Working_in_Inspector.mov…

Also adding this here, since I'd imagine beam height parameters and 'force horizontal' functionality in the inspector should be fixed together.

For example, if changing force horizontal, it will affect the values in the text fields here. Elevating to P0 because this is non-functional.

wizofaus:
My comment explained what I'd found while trying to investigate why this wasn't working, which is likely to be useful to whatever developer takes this on, so I'm not sure why it got deleted?

PR #12085, opened 2022-06-18
Fix the "Force horizontal" beam property, and the interaction with other properties

cbjeukendrup:
Resolves: #9037

Note: this is largely the work of @mike-spa, and I built some more UI fixes on top of it.


mike-spa (approved):

RomanPudashkin (approved):

src/inspector/view/qml/MuseScore/Inspector/notation/notes/BeamSettings.qml
```diff
@@ -278,22 +271,24 @@ FocusableItem {
...
+                                navigation.name: "BeamHeightRightControl"
                                 navigation.panel: root.navigationPanel
                                 navigation.row: lockButton.navigation.row + 1
-                                navigation.accessible.name: beamHeight.titleText + " " + qsTrc("inspector", "Left") + " " + currentValue
+                                navigation.accessible.name: beamHeight.titleText + " " + qsTrc("inspector", "Right") + " " + currentValue
 
-                                onValueEdited: function(newValue) { root.model.beamVectorY.value = newValue }
+                                onValueEdited: function(newValue) {
+                                    root.model.beamHeightRight.value = newValue
+                                }
```
wizofaus, line 291: This appears to leave the "Force Horizontal" toggle as a large "flat button", which seems an odd control to use for a toggle.
I actually thought there was some logic in moving away from "Force horizontal" being a toggle, and rather simply adjusting the current stem heights so the beam became horizontal - the problem with the toggle in v3 is that you can turn it on, but then adjust the beam angle so it's no longer horizontal, but the "force horizontal" checkbox remains checked!  Is this addressed here?

Jojo-Schmitz: In MU3 that "force horizontal" remains check, but gets disabled is you modify the beam. That's OK IMHO. More probelmatic is that then "Userdefined position" also is disabled (and checked), yet the Position X/Y settings are used (and active)

cbjeukendrup: I agree that "Force horizontal" could as well be a checkbox, which will make its meaning clearer. We're still thinking about that a bit, but probably we'll change it. 

With this PR, using "Force horizontal" and "Userdefined position" are not mutually exclusive anymore, so both controls will be enabled all the time. They (should) work together nicely now, so that after choosing that you want the beam to be flat, you can still manually adjust its height.

oktophonie: The new behaviour seems much better now to me, than in MU3.6.

The only remaining question for me is: if a beam has been manually moved, then 'Force horizontal' is clicked, the beam resets to its default position, and is then made horizontal. Perhaps the beam should, in this case, stay where it was moved, but be made horizontal? (I'm not really sure)

cbjeukendrup: I agree that would be better. Just resetting it was the easy option, but probably we can apply some smartness here. Will try it out.

cbjeukendrup: I replaced the Force horizontal button with a checkbox. 

About saving the modified beam height when toggling 'Force horizontal':
When you _enable_ it, then it is not difficult to keep the height and just make the beam flat. 
But when you _disable_ it, it seems difficult to go back to the old shape without fully resetting, because that is not stored anywhere. So the beam would stay flat, which may give the user the feeling that it didn't work. A solution could be to reset the position when the user _disables_ 'force horizontal', and keep it when the user _enables_ it. Is that okay?

oktophonie: My first thought is that it's fine to leave the beam in its horizontal position when the user unchecks the box. The user is then a position where they can edit the endpoints independently, or they can easily manually reset the beam with Ctrl+R. 

I agree that it's a bit odd that seeing no change in the score when the option is turned off is a bit confusing, but having the beam also (potentially) drastically change position might also be confusing behaviour.

I can't say this is a particularly strong preference though. Maybe @Tantacrul or others have an opinion.

cbjeukendrup: Okay, so now I implemented that: when disabling Force horizontal on a custom-positioned beam, it will stay flat but allow customizations again.

oktophonie (approved):

PR #12088, opened 2022-06-18
Support slides/bends on multiple notes of a chord

mike-spa:
Resolves: #10316 

Currently, Musescore does not support inserting slides and bends on multiple notes of a chord. This is particularly important in guitar notation, but there may be plenty of other applications, so we should support it, regardless of guitar.

The origin of the problem is that these type of marks are handled by the `ChordLine` class, whose `parent` item is `Chord`, so all the marks are attached to a Chord, not to an individual note. The "proper" solution would be to change the parent of ChordLine from `Chord` to `Note`, but this would require to modify almost every bit of code where this class is used. 

Instead, this is a simple temporary solution which allows to support the feature. A small modification to the read and write functions was also necessary to ensure that they are reloaded correctly upon reopening a file. Nothing changes when opening older files.

BEFORE:
Screenshot from 2022-06-18 16-45-18
Screenshot from 2022-06-18 16-43-27

AFTER:
Screenshot from 2022-06-18 16-45-03
Screenshot from 2022-06-18 16-43-08

### DIFF
--- a/src/engraving/libmscore/chordline.h
+++ b/src/engraving/libmscore/chordline.h
@@ -49,6 +49,7 @@ protected:
     bool modified;
     qreal _lengthX;
     qreal _lengthY;
+    Note* _note;
     static constexpr double _baseLength = 1.0;
 
     friend class Factory;
@@ -92,6 +93,9 @@ public:
 
     bool isToTheLeft() const { return _chordLineType == ChordLineType::PLOP || _chordLineType == ChordLineType::SCOOP; }
     bool isBelow() const { return _chordLineType == ChordLineType::SCOOP || _chordLineType == ChordLineType::FALL; }
+
+    void setNote(Note* note) { _note = note; }
+    Note* note() const { return _note; }
 };
 
 extern const char* scorelineNames[];
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1200,6 +1200,9 @@ void Chord::write(XmlWriter& xml) const
         _tremolo->write(xml);
     }
     for (EngravingItem* e : el()) {
+        if (e->isChordLine() && toChordLine(e)->note()) { // this is now written by Note
+            continue;
+        }
         e->write(xml);
     }
     xml.endElement();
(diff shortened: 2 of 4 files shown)


alexpavlov96:
@boryskuzmenko this should fix the problem that we cannot put the chordlines on different notes on a chord (from the palette). So this bug should be created and attached here. It still doesn't fix the *https://github.com/musescore/MuseScore/issues/10316*

alexpavlov96 (approved):

mike-spa:
Added new commit that fixes the layout also in the case of linked staff+TAB. Seems to work nicely!

Screenshot from 2022-06-19 23-08-10

alexpavlov96:
@mike-spa magician!

src/engraving/libmscore/chordline.h
```diff
@@ -92,6 +93,9 @@ class ChordLine final : public EngravingItem
 
     bool isToTheLeft() const { return _chordLineType == ChordLineType::PLOP || _chordLineType == ChordLineType::SCOOP; }
     bool isBelow() const { return _chordLineType == ChordLineType::SCOOP || _chordLineType == ChordLineType::FALL; }
+
+    void setNote(Note* note) { _note = note; }
```
vpereverzev, line 97: that's a bit strange if chordline is attached to one of the notes within a chord. I'd rather expect a chord as a base point for a chord-line

vpereverzev: well, we'll fix this up later on

mike-spa: Yes that is the problem. IMO, this class should have been called `NoteLine` and made it with a `Note` parent, not `Chord` parent, because these type of lines are conceptually similar to glissandos, so they should be attached to _notes_. But changing the whole class implementation is not an option right now, so I did this temporary fix, until we do it properly.

src/engraving/libmscore/chordline.h
```diff
@@ -49,6 +49,7 @@ class ChordLine final : public EngravingItem
     bool modified;
     qreal _lengthX;
     qreal _lengthY;
+    Note* _note;
```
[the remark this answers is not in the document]
mike-spa: yup, I'll do that right away

Issue #11827, opened 2022-05-31
[MU4 Task] Implement `F7` shortcut to toggle "Instruments" panel visibility


bkunda:
**Task description**
To resolve #10484 

The "instruments" panel currently has no shortcut to toggle its visibility, unlike its associated "properties" and "palettes" panels. 

The proposal here is to allocate the default shortcut `F7` to toggle the visibility of the "instruments" panel. 

This behaviour would be identical to that of the `F8` (toggle "properties") and `F9` (toggle "palettes") keys.

The name in the Preferences > Shortcuts should be "Instruments" (to establish consistency with other panel elements).

Lastly, the shortcut command in Preferences > Shortcuts currently designated "Instruments..." should be renamed to "Add/change instruments..." 

Thanks!

HemantAntony:
@bkunda I have a fix: #11830

@cbjeukendrup I hope you don't mind. It was an easy issue :)

cbjeukendrup:
@HemantAntony No problem! :)

PR #12089, opened 2022-06-18
fix #11896 and #12022: There are two dialogs called "Instruments"... and "Change time signature" button...

### LINKED ISSUE #11896: There are two dialogs called "Instruments"; cannot pre-specify staff type to be added
**Describe the issue**

1. There are now two dialogs with the name **Instruments**. We should rename the 'Instruments' popup to be called **Add or remove instruments**

### LINKED ISSUE #12022: "Change time signature" button (Properties sidebar) should be "Time signature properties"
**Describe the bug**

if you want to change an existing time signature, the method of choice should be to select it and click on a new time signature from the palettes.

In MS4, if you select an existing time signature, there is a button in the Properties sidebar called "Change time signature". The wording is wrong since this is _not_ the function of the dialog – which is actually to change the text _appearance_ of the time sig. or the beaming of notes under its "jurisdiction".

IMV, this button should revert to "Time signature properties" (as the option was in MS3 and before).

**To Reproduce**

1. Select a time signature.
2. Inspect the properties sidebar for the "Change time signature" button.

**Expected behavior**

The button should be called "Time signature properties" (or suchlike).

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2499731453, revision: github-musescore-musescore-d8b6263

poypoyan:
Resolves #11896, resolves #12022

These are just label changes.

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/notation/timesignatures/TimeSignatureSettings.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/notation/timesignatures/TimeSignatureSettings.qml
@@ -143,7 +143,7 @@ Column {
     FlatButton {
         width: parent.width
 
-        text: qsTrc("inspector", "Change time signature")
+        text: qsTrc("inspector", "Time signature properties")
 
         navigation.name: "ChangeButton"
         navigation.panel: root.navigationPanel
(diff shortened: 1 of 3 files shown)


cbjeukendrup:
Looks good! 

I thought, for an extra improvement, you could also update the action name in instrumentsuiactions.cpp, line 34. Currently it says `Add/change instruments…`, but I think it should be updated to `Add or remove instruments…`. This string is currently not displayed in the UI, but it may still be good to update this so that we won't forget.

cbjeukendrup (approved):

HemantAntony:
@cbjeukendrup, @bkunda had suggested the `Add/change instruments…` in #11827

PR #12091, opened 2022-06-19
Fix #11522 - ensure element is scrolled into view when using top-staff command

### LINKED ISSUE #11522: [MU4 Issue] Assigning a shortcut to "go to top staff" doesn't appear to work

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/11522) and fixes random crash

### DIFF
--- a/src/notation/internal/notationselectionrange.cpp
+++ b/src/notation/internal/notationselectionrange.cpp
@@ -164,8 +164,6 @@ mu::engraving::Segment* NotationSelectionRange::rangeStartSegment() const
 {
     mu::engraving::Segment* startSegment = score()->selection().startSegment();
 
-    startSegment->measure()->firstEnabled();
-
     if (!startSegment) {
         return nullptr;
     }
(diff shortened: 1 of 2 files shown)


src/notation/internal/notationselectionrange.cpp
```diff
@@ -164,8 +164,6 @@ mu::engraving::Segment* NotationSelectionRange::rangeStartSegment() const
 {
     mu::engraving::Segment* startSegment = score()->selection().startSegment();
 
-    startSegment->measure()->firstEnabled();
```
wizofaus, line 167: doesn't do anything useful and crashes if startSegment is nullptr which is definitely can be in some cases

RomanPudashkin (approved):

PR #12096, opened 2022-06-19
Remove "Images" field from Preferences > Folders

cbjeukendrup:
In MU3, as far as I can see, this field was only used by the Image Capture feature, to choose where captured images would be saved. This feature is delayed for MuseScore 4.0, and when we reimplement it, I'm pretty sure that the UX will be totally different, and most probably won't involve a field in Preferences.

### DIFF
--- a/src/appshell/view/preferences/folderspreferencesmodel.h
+++ b/src/appshell/view/preferences/folderspreferencesmodel.h
@@ -72,8 +72,7 @@ private:
         Templates,
         Plugins,
         SoundFonts,
-        VST3,
-        Images
+        VST3
     };
 
     enum class FolderValueType {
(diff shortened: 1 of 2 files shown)


src/appshell/view/preferences/folderspreferencesmodel.h
```diff
@@ -73,7 +73,6 @@ class FoldersPreferencesModel : public QAbstractListModel, public async::Asyncab
         Plugins,
         SoundFonts,
         VST3,
```
Eism, line 75: Please remove the comma

Eism (approved):

src/appshell/view/preferences/folderspreferencesmodel.h
cbjeukendrup: Done!

PR #12098, opened 2022-06-19
fix #12097 - ensure new/changed notes/rests are always scrolled into view

### LINKED ISSUE #12097: [MU4 Issue] If current measure moves to new system after changing durations it's not scrolled into view
**Describe the bug**
Quite often changing the duration of existing notes/rests in a measure or entering in new notes can cause it to become too wide and move to the next system. This is fine, but it should at least be automatically scrolled into view if so.
This is especially a problem with note entry.

**To Reproduce**
1. Create a new score using an ensemble template like Rock Band (only one system fitting on screen by default)
2. Click the "quaver" button (or hit the 4 key)
3. Select the last measure of the first system in, say, the Drum part
4. Enter a note (e.g. hit A)
5. Entered note (and containing measure) is no longer visible (it has moved to next system)

**Expected behavior**
Selected note/rest should remain visible after moving automatically to new system, as per v3.
NB the same occurs when "padding" a rest by changing its duration.

**Screenshots**

**Platform information**
 - Windows 10

**Additional context**

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/12097

Ensure newly added or changed notes/rests are always kept in view

### DIFF
--- a/src/notation/internal/notationnoteinput.cpp
+++ b/src/notation/internal/notationnoteinput.cpp
@@ -254,10 +254,6 @@ void NotationNoteInput::putNote(const PointF& pos, bool replace, bool insert)
 
     notifyNoteAddedChanged();
     notifyAboutStateChanged();
-
-    if (mu::engraving::ChordRest* chordRest = score()->inputState().cr()) {
-        m_interaction->showItem(chordRest);
-    }
 }
 
 void NotationNoteInput::removeNote(const PointF& pos)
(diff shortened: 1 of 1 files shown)


src/notation/internal/notationnoteinput.cpp
```diff
@@ -254,10 +254,6 @@ void NotationNoteInput::putNote(const PointF& pos, bool replace, bool insert)
 
     notifyNoteAddedChanged();
     notifyAboutStateChanged();
-
-    if (mu::engraving::ChordRest* chordRest = score()->inputState().cr()) {
-        m_interaction->showItem(chordRest);
-    }
```
wizofaus, line 260: logic should apply to all actions (putNote/padNote/addNote)

RomanPudashkin (approved):

PR #12099, opened 2022-06-20
fix #9429 - various bugs editing barline properties

### LINKED ISSUE #9429: [Inspector] In the 'Barlines' section, the 'Left (start) repeat' option behaves incorrectly
**Describe the bug**
This is a problem with MS3 that we have inherited for MS4

**Video**

https://user-images.githubusercontent.com/21022311/136942895-765e2ea7-a0cc-4197-b8c1-4602306d2d56.mov

wizofaus:
Resolves: #9429

Fixes a number of bugs with barline properties:

a) when changing a barline to "start repeat", it was changing the barline *after* the one selected
b) for a "start repeat" barline added at the beginning of the system, it was allowing changing the type, despite this not being meaningful. The controls are now hidden for such barlines
c) for a final measure barline (end of section) it was allowing changing the type to "start repeat" or "start & end repeat", neither of which worked. Those options are now hidden for final measure barlines.
d) when changing from "start repeat" or "start & end repeat" to any other sort of barline, the start repeat barline was not getting removed as expected, even if the new barline type was correctly drawn in the same location, and the barline was not selected after doing so (under the hood it's actually now a different barline...)

### DIFF
--- a/src/engraving/libmscore/barline.cpp
+++ b/src/engraving/libmscore/barline.cpp
@@ -67,22 +67,37 @@ static void undoChangeBarLineType(BarLine* bl, BarLineType barType, bool allStav
     if (!m) {
         return;
     }
-
-    if (barType == BarLineType::START_REPEAT) {
-        m = m->nextMeasure();
-        if (!m) {
-            return;
-        }
-    } else if (bl->barLineType() == BarLineType::START_REPEAT) {
-        if (barType != BarLineType::END_REPEAT) {
-            m->undoChangeProperty(Pid::REPEAT_START, false);
-        }
+    bool isStartRepeat = barType == BarLineType::END_START_REPEAT || barType == BarLineType::START_REPEAT;
+    if (bl->barLineType() == BarLineType::START_REPEAT) {
         m = m->prevMeasure();
         if (!m) {
+            if (!isStartRepeat) {
+                bl->measure()->undoChangeProperty(Pid::REPEAT_START, false);
+            }
             return;
         }
     }
-
+    if (m->nextMeasure()) {
+        if (isStartRepeat || bl->barLineType() == BarLineType::START_REPEAT || m->system() == m->nextMeasure()->system()) {
+            m->nextMeasure()->undoChangeProperty(Pid::REPEAT_START, isStartRepeat);
+        }
+        if (!isStartRepeat) {
+            if (bl->score()->selection().element() == bl) {
+                bl->score()->select(const_cast<BarLine*>(m->endBarLine()));
+            }
+            if (bl->barLineType() == BarLineType::START_REPEAT) {
+                bl->score()->onElementDestruction(bl);
+                if (m->system() != m->nextMeasure()->system()) {
+                    return; // don't touch the previous barline if it's on a different system
+                }
+                bl = const_cast<BarLine*>(m->endBarLine());
+            }
+        }
+    }
+    bool isEndRepeat = barType == BarLineType::END_START_REPEAT || barType == BarLineType::END_REPEAT;
+    if (isEndRepeat || (m->nextMeasure() && m->system() == m->nextMeasure()->system())) {
+        m->undoChangeProperty(Pid::REPEAT_END, isEndRepeat);
+    }
     switch (barType) {
     case BarLineType::END:
     case BarLineType::NORMAL:
@@ -189,8 +204,8 @@ static void undoChangeBarLineType(BarLine* bl, BarLineType barType, bool allStav
         }
         for (Score* lscore : m2->score()->scoreList()) {
             Measure* lmeasure = lscore->tick2measure(m2->tick());
-            if (lmeasure) {
-                lmeasure->undoChangeProperty(Pid::REPEAT_START, true);
+            if (lmeasure && lmeasure->nextMeasure()) {
+                lmeasure->nextMeasure()->undoChangeProperty(Pid::REPEAT_START, true);
             }
         }
     }
@@ -1503,18 +1518,8 @@ bool BarLine::setProperty(Pid id, const PropertyValue& v)
 void BarLine::undoChangeProperty(Pid id, const PropertyValue& v, PropertyFlags ps)
 {
     if (id == Pid::BARLINE_TYPE && segment()) {
-        const BarLine* bl = this;
         BarLineType blType = v.value<BarLineType>();
-        if (blType == BarLineType::START_REPEAT) {     // change next measures endBarLine
-            if (bl->measure()->nextMeasure()) {
-                bl = bl->measure()->nextMeasure()->endBarLine();
-            } else {
-                bl = 0;
-            }
-        }
-        if (bl) {
-            undoChangeBarLineType(const_cast<BarLine*>(bl), v.value<BarLineType>(), true);
-        }
+        undoChangeBarLineType(this, v.value<BarLineType>(), true);
     } else {
         EngravingObject::undoChangeProperty(id, v, ps);
     }
--- a/src/engraving/utests/barline_tests.cpp
+++ b/src/engraving/utests/barline_tests.cpp
@@ -344,33 +344,43 @@ TEST_F(BarlineTests, barline179726)
     Measure* m = score->firstMeasure();
 
     // drop NORMAL onto initial START_REPEAT barline will remove that START_REPEAT
-    dropNormalBarline(m->findSegment(SegmentType::StartRepeatBarLine, m->tick())->elementAt(0));
+    Segment* seg = m->findSegment(SegmentType::StartRepeatBarLine, m->tick());
+    EXPECT_TRUE(seg);
+    dropNormalBarline(seg->elementAt(0));
     EXPECT_EQ(m->findSegment(SegmentType::StartRepeatBarLine, Fraction(0, 1)), nullptr);
 
     // drop NORMAL onto END_START_REPEAT will turn into NORMAL
-    dropNormalBarline(m->findSegment(SegmentType::EndBarLine, m->endTick())->elementAt(0));
-    BarLine* bar = static_cast<BarLine*>(m->findSegment(SegmentType::EndBarLine, m->endTick())->elementAt(0));
+    seg = m->findSegment(SegmentType::EndBarLine, m->endTick());
+    EXPECT_TRUE(seg);
+    dropNormalBarline(seg->elementAt(0));
+    seg = m->findSegment(SegmentType::EndBarLine, m->endTick());
+    EXPECT_TRUE(seg);
+    BarLine* bar = static_cast<BarLine*>(seg->elementAt(0));
     EXPECT_TRUE(bar);
     EXPECT_EQ(bar->barLineType(), BarLineType::NORMAL);
 
     m = m->nextMeasure();
 
     // drop NORMAL onto the END_REPEAT part of an END_START_REPEAT straddling a newline will turn into NORMAL at the end of this meas
-    dropNormalBarline(m->findSegment(SegmentType::EndBarLine, m->endTick())->elementAt(0));
-    bar = static_cast<BarLine*>(m->findSegment(SegmentType::EndBarLine, m->endTick())->elementAt(0));
+    seg = m->findSegment(SegmentType::EndBarLine, m->endTick());
+    EXPECT_TRUE(seg);
+    dropNormalBarline(seg->elementAt(0));
+    bar = static_cast<BarLine*>(seg->elementAt(0));
     EXPECT_TRUE(bar);
     EXPECT_EQ(bar->barLineType(), BarLineType::NORMAL);
 
     m = m->nextMeasure();
 
     // but leave START_REPEAT at the beginning of the newline
-    bar = static_cast<BarLine*>(m->findSegment(SegmentType::StartRepeatBarLine, m->tick())->elementAt(0));
+    seg = m->findSegment(SegmentType::StartRepeatBarLine, m->tick());
+    bar = static_cast<BarLine*>(seg->elementAt(0));
     EXPECT_TRUE(bar);
 
     // drop NORMAL onto the meas ending with an END_START_REPEAT straddling a newline will turn into NORMAL at the end of this meas
     // but note I'm not verifying what happens to the START_REPEAT at the beginning of the newline...I'm not sure that behavior is well-defined yet
     dropNormalBarline(m);
-    bar = static_cast<BarLine*>(m->findSegment(SegmentType::EndBarLine, m->endTick())->elementAt(0));
+    seg = m->findSegment(SegmentType::EndBarLine, m->endTick());
+    bar = static_cast<BarLine*>(seg->elementAt(0));
     EXPECT_TRUE(bar);
     EXPECT_EQ(bar->barLineType(), BarLineType::NORMAL);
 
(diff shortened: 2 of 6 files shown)


src/engraving/libmscore/barline.cpp
```diff
@@ -67,21 +67,23 @@ static void undoChangeBarLineType(BarLine* bl, BarLineType barType, bool allStav
...
+    }
+    if (m->nextMeasure()) {
+        bool isStartRepeat = barType == BarLineType::END_START_REPEAT || barType == BarLineType::START_REPEAT;
+        m->nextMeasure()->undoChangeProperty(Pid::REPEAT_START, isStartRepeat);
+        if (!isStartRepeat) {
+            if (bl->score()->selection().element() == bl) {
+                bl->score()->select(const_cast<BarLine*>(m->endBarLine()));
+            }
+            if (bl->barLineType() == BarLineType::START_REPEAT) {
+                bl->score()->onElementDestruction(bl);
```
wizofaus, line 81: NotationInteraction still maintains a reference to this otherwise and attempts to draw it until the selection is changed

src/engraving/libmscore/barline.cpp
```diff
@@ -1503,18 +1505,8 @@ bool BarLine::setProperty(Pid id, const PropertyValue& v)
...
 {
     if (id == Pid::BARLINE_TYPE && segment()) {
-        const BarLine* bl = this;
         BarLineType blType = v.value<BarLineType>();
-        if (blType == BarLineType::START_REPEAT) {     // change next measures endBarLine
-            if (bl->measure()->nextMeasure()) {
-                bl = bl->measure()->nextMeasure()->endBarLine();
-            } else {
-                bl = 0;
-            }
```
wizofaus, line 1513: I literally have no idea why this code was here...

wizofaus:
TEST_F(BarlineTests, barline179726) is failing with a SEGFAULT, not too surprising, will likely have to change that test anyway.

src/engraving/utests/barline_tests.cpp
```diff
@@ -344,65 +344,83 @@ TEST_F(BarlineTests, barline179726)
     Measure* m = score->firstMeasure();
 
     // drop NORMAL onto initial START_REPEAT barline will remove that START_REPEAT
-    dropNormalBarline(m->findSegment(SegmentType::StartRepeatBarLine, m->tick())->elementAt(0));
```
wizofaus, line 347: no changes to what's actually being tested here, just ensuring we get clear/helpful error messages if `findSegment` returns `nullptr`

Tantacrul:
I've checked this out and it seems to work as expected!

I also tried out your case for the 'end of system' behaviour, which also seems to work as expected. Never thought to check something like that.

Also, the repeats seem to be playing back correctly too.

Tantacrul:
If you're ready, I can make a request to review and merge?

src/engraving/libmscore/barline.cpp
RomanPudashkin: As far as I can tell, this element is not destroyed here, so using onElementDestruction here is incorrect

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
```diff
@@ -56,19 +56,28 @@ Column {
...
+                        { text: qsTrc("symUserNames", "Right and left repeat sign"), value: BarlineTypes.TYPE_END_START_REPEAT },
+                        { text: qsTrc("symUserNames", "Dashed barline"), value: BarlineTypes.TYPE_DASHED },
+                        { text: qsTrc("symUserNames", "Final barline"), value: BarlineTypes.TYPE_FINAL },
+                        { text: qsTrc("symUserNames", "Dotted barline"), value: BarlineTypes.TYPE_DOTTED },
+                        { text: qsTrc("symUserNames", "Reverse final barline"), value: BarlineTypes.TYPE_REVERSE_END },
+                        { text: qsTrc("symUserNames", "Heavy barline"), value: BarlineTypes.TYPE_HEAVY },
+                        { text: qsTrc("symUserNames", "Heavy double barline"), value: BarlineTypes.TYPE_DOUBLE_HEAVY },
+                    ]
+            return model.filter(function(item) { 
+                return (item.value != BarlineTypes.TYPE_START_REPEAT && item.value != BarlineTypes.TYPE_END_START_REPEAT) ||
```
RomanPudashkin, line 76: This code should be in the model. qml isn't the right place for such logic

src/engraving/libmscore/barline.cpp
wizofaus: Not deleted as an in-memory object no, but the measure layout code destroys it by removing the owning segment (actually I suspect there's a memory leak, the destructor never seemed to get called, but it no longer belongs to the score).
But without this, the element appears to remain on the screen on the next paint because it's referred to by the NotationInteraction as the "edited object".

src/engraving/libmscore/barline.cpp
wizofaus: Hard to see how that issue could have anything to do with barlines at all!

MarcSabatella: Ooops, sorry, I had the wrong link - fixed now.

wizofaus: Ah ok.  But it is rather surprising 3.6 is still so broken in this regard - there's literally no way I can see of making the first barline in a score a "start repeat" using just the inspector, unless you insert a dummy measure at the beginning first.

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
wizofaus: Basically I'm suggesting to leave the list of barline types defined in the qml file where it is now, but call a filter function that's exposed by the C++ model to determine which ones should be shown.  Does that sound reasonable?

wizofaus: Or should I use SortFilterProxyModel? There's quite a few examples of that, though I'm not sure about one using a Q_INVOKABLE filter.

wizofaus: SortFilterProxyModel seemed excessively complex for what I needed, so I hope my "compromise" solution is ok...even though I admit it's the only use of the javascript filter function in any of the qml models.

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
```diff
@@ -56,19 +56,27 @@ Column {
...
+                        { text: qsTrc("symUserNames", "Dashed barline"), value: BarlineTypes.TYPE_DASHED },
+                        { text: qsTrc("symUserNames", "Final barline"), value: BarlineTypes.TYPE_FINAL },
+                        { text: qsTrc("symUserNames", "Dotted barline"), value: BarlineTypes.TYPE_DOTTED },
+                        { text: qsTrc("symUserNames", "Reverse final barline"), value: BarlineTypes.TYPE_REVERSE_END },
+                        { text: qsTrc("symUserNames", "Heavy barline"), value: BarlineTypes.TYPE_HEAVY },
+                        { text: qsTrc("symUserNames", "Heavy double barline"), value: BarlineTypes.TYPE_DOUBLE_HEAVY },
+                    ]
+            return model.filter(function(item) { return root.barlineSettingsModel && root.barlineSettingsModel.isStyleChangingAllowed 
+                && root.barlineSettingsModel.shouldShowType(item.value) })
+        }
```
wizofaus, line 77: I had hoped this could be simpler, but in order to ensure it does get called every time as needed and worked in all cases, this was the simplest I could make it...if I understood better how Qt Quick translated between javascript and C++ I'm sure it could be improved.

src/engraving/libmscore/barline.cpp
MarcSabatella: The Inspector was never intended to be a supported way of creating repeat barlines in the first place.  And in fact, it's only possible for single-staff scores - scores with multiple staves gray this option out, because the Inspector changes only the current staff (in MU3 anyhow) and it would be inconsistent to have a repeat on one staff but not another.

So, between the fact that it's not actually the officially supported way of creating repeats (the palette is), and it isn't possible to do via the Inspector at all except for single-staff scores, and of course it has had this obvious bug for the past couple of years but only two or three reports, it seems pretty safe to say, most people don't create repeats this way to begin with.

wizofaus: Actually I've performed off scores that had repeat marks only in two staves,  while the rest of the ensemble played through composed music. I wouldn't necessarily recommend it, but it can work. I assume there'd be some way to kludge it...

MarcSabatella: Right, it definitely can happen that you have scores with mismatched repeats like this, but we didn't want it to be too easy to create that accidentally, which is why it's disabled in the Inspector.  The current workaround would be to add the repeat barline from the Symbols palette and position it manually.  Definitely room for improvement.

Offhand I can't think of a good reason to keep the repeat options in the Inspector given that they are only active for single-staff scores and in that case they don't do anything differently from the palette, aside from appearing in the wrong measure.

Also, in case it isn't obvious: normally, the Inspector in MU3 applies a barline change to just that staff, similar to using Ctrl+click from the palette.  Ctrl is our standard shortcut to mean "apply this to a single staff only even though it would normally affect all".  But Ctrl+click on the repeat barlines still adds to all staves.

Tantacrul: Well, they work in MS4 across staves. If it could just be made consistent and not cause crashes, then that's an improvement over 3. However, if it is rife with issues, then I'd potentially remove it from the Inspector.

All in all - if it works, keep it. Don't limit it to one instrument. That doesn't make sense to the user.

Agreed about manual placement of symbols to achieve single instrument repeats in a score with multiple instruments.

MarcSabatella: I'm mostly in agreement.  But I would just observe, it *is* important to allow the user to set barline properties for just a single staff.  It's the key to how you can override barline spans, or hide barlines for just one staff, or create "mensurstrich" etc.  Quite a few real-world effects require modifying barlines properties independently per staff.

Luckily most of the settings in the MU4 properties panel do continue to work that way - it seems only the "Style" has been modified to affect all staves.  To me that's a mistake - only one staff's barline is selected, so only that should be affected.  But it would be fine if clicking a barline actually selected across all staves like we do for key and time signatures.  And then Ctrl+click could still be used to select just a single staff's barline and thus affect it's properties.  That to me would be more consistent and logical without giving up functionality.  Only issue would be the discoverability of Ctrl+click as a method of limiting selection to a single staff, which is definitely a consideration and a reason why things are as they are in MU3.

MarcSabatella: So bottom line for me - right now Style works differently from the other barline properties.  It does nothing different from the palette except impose limitations and bugs.  I think the best course of action is to remove this.  Leaving the other properties as they are - affecting only the current staff.

wizofaus: Removing the repeat options from the properties panel would still require code changes to support what happens when you do select a repeat barline - in that case you wouldn't want to show any option to charge the barline type (only height/ span etc.). But honestly I think what I've implemented here works fine,  unless QA find some specific issues,  there are definitely some cases that mightn't work as expected.

wizofaus: I just tried with v4 *and* v3, using CTRL+click to select a barline for just a single staff doesn't work - in fact, it looks the same as regular click (only the barline on the staff selected is highlighted), but any property changes apply to all staves. So how am I supposed to change the barline style (e.g. just to 'dotted') for a single staff using properties panel/inspector?

MarcSabatella: Not sure which properties you tried, but in both MU3 and MU4, all properties I tried affect just the selected staff *except* for Style, which affects them all.  Eg, try toggling visibility or any of the span options or setting an X offset.  So right now it's inconsistent.  Most properties affect only what's selected, except Style, which affects that barline across all staves.

And yes, Ctrl+click *currently* is the same as regular click - both select just the current staff.  I was suggesting we *could* change regular click to select across all staves similarly to key and time signatures, *if* we wanted to fix the consistency problem by having *all* properties affect all staves.

wizofaus: Yes, I'm only testing the "BarlineType" property (which is shown as "Style" in the inspector/properties panel).  How do I change it for a single staff (other than dragging from the palette)?

MarcSabatella: As I have been saying, that's the one property that behaves inconsistently - it's the only property that currently applies to all staves.  The only current way to apply the "Style" property to only selected staves is via the palette using Ctrl.

This is why I am pointing out that this setting serves no uniquely useful purpose currently, and that since it's *also* inconsistent with the other properties, I'm totally fine with seeing it removed.  At least right now I don't see a downside.

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
```diff
@@ -56,19 +56,27 @@ Column {
...
+                        { text: qsTrc("symUserNames", "Right and left repeat sign"), value: BarlineTypes.TYPE_END_START_REPEAT },
+                        { text: qsTrc("symUserNames", "Dashed barline"), value: BarlineTypes.TYPE_DASHED },
+                        { text: qsTrc("symUserNames", "Final barline"), value: BarlineTypes.TYPE_FINAL },
+                        { text: qsTrc("symUserNames", "Dotted barline"), value: BarlineTypes.TYPE_DOTTED },
+                        { text: qsTrc("symUserNames", "Reverse final barline"), value: BarlineTypes.TYPE_REVERSE_END },
+                        { text: qsTrc("symUserNames", "Heavy barline"), value: BarlineTypes.TYPE_HEAVY },
+                        { text: qsTrc("symUserNames", "Heavy double barline"), value: BarlineTypes.TYPE_DOUBLE_HEAVY },
+                    ]
+            return model.filter(function(item) { return root.barlineSettingsModel && root.barlineSettingsModel.isStyleChangingAllowed 
+                && root.barlineSettingsModel.shouldShowType(item.value) })
```
wizofaus, line 76: Are you OK with this solution @RomanPudashkin? I haven't really managed to find anything else that doesn't involve a much riskier/bigger change than I would think makes sense this close to Beta.

RobFog:
Rebase needed.

wizofaus:
Perhaps but it seems there's no plan to incorporate these fixes.

src/inspector/models/notation/barlines/barlinesettingsmodel.h
```diff
@@ -54,9 +56,12 @@ class BarlineSettingsModel : public AbstractInspectorModel
     PropertyItem* hasToShowTips() const;
 
     bool isRepeatStyleChangingAllowed() const;
+    bool isStyleChangingAllowed() const;
 
 signals:
     void isRepeatStyleChangingAllowedChanged();
+    void isStyleChangingAllowedChanged();
+    void isFinalBarlineChanged();
```
cbjeukendrup, line 64: It looks like nobody is listening to this signal :)

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
```diff
@@ -56,19 +56,27 @@ Column {
...
+                        { text: qsTrc("symUserNames", "Double barline"), value: BarlineTypes.TYPE_DOUBLE },
+                        { text: qsTrc("symUserNames", "Left (start) repeat sign"), value: BarlineTypes.TYPE_START_REPEAT },
+                        { text: qsTrc("symUserNames", "Right (end) repeat sign"), value: BarlineTypes.TYPE_END_REPEAT },
+                        { text: qsTrc("symUserNames", "Right and left repeat sign"), value: BarlineTypes.TYPE_END_START_REPEAT },
+                        { text: qsTrc("symUserNames", "Dashed barline"), value: BarlineTypes.TYPE_DASHED },
+                        { text: qsTrc("symUserNames", "Final barline"), value: BarlineTypes.TYPE_FINAL },
+                        { text: qsTrc("symUserNames", "Dotted barline"), value: BarlineTypes.TYPE_DOTTED },
+                        { text: qsTrc("symUserNames", "Reverse final barline"), value: BarlineTypes.TYPE_REVERSE_END },
+                        { text: qsTrc("symUserNames", "Heavy barline"), value: BarlineTypes.TYPE_HEAVY },
+                        { text: qsTrc("symUserNames", "Heavy double barline"), value: BarlineTypes.TYPE_DOUBLE_HEAVY },
```
cbjeukendrup, line 73: These translation context have been changed to `engraving/sym`.

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
```diff
@@ -56,19 +56,27 @@ Column {
...
+                    ]
+            return model.filter(function(item) { return root.barlineSettingsModel && root.barlineSettingsModel.isStyleChangingAllowed 
+                && root.barlineSettingsModel.shouldShowType(item.value) })
+        }
+
+        model: getModel()
```
cbjeukendrup, line 79: It is possible to do: 
```qml
        model: {
            let barlineTypes = [
                        { text: qsTrc("symUserNames", "Single barline"), value: BarlineTypes.TYPE_NORMAL },
                        ...
                    ]
            return barlineTypes.filter(function(item) { return root.barlineSettingsModel && root.barlineSettingsModel.isStyleChangingAllowed 
                && root.barlineSettingsModel.shouldShowType(item.value) })
        }
```

Or a micro-optimized version:
```qml
        model: {
            if (!root.barlineSettingsModel || !root.barlineSettingsModel.isStyleChangingAllowed) {
                return []
            }

            let barlineTypes = [
                        { text: qsTrc("symUserNames", "Single barline"), value: BarlineTypes.TYPE_NORMAL },
                        ...
                    ]
            return barlineTypes.filter(function(item) { root.barlineSettingsModel.shouldShowType(item.value) })
        }
```

src/inspector/view/qml/MuseScore/Inspector/notation/barlines/BarlineSettings.qml
cbjeukendrup: You could create a Q_PROPERTY in C++ with type QVariantList, and do the filtering in C++. The getter might look somewhat like this:
```c++
QVariantList result;

for (BarLineType type: theBarlineTypes) {
    if (shouldInclude(type)) {
        QVariantMap map;
        map["value"] = static_cast<int>(type);
        map["text"] = getTheUserNameAsQStringForType(type);
        result << map;
    }
}

return result;
```
This will be translated in QML to a JS array containing JS objects just like when you define it in QML.

wizofaus:
This appears to be no longer relevant as it seems you can't change a barline between a start/end repeat and a regular barline in the properties panel any more?  If that's true this can be closed @cbjeukendrup

cbjeukendrup:
If I recall correctly, we "fixed" the problem by removing that functionality, because we were in a rush before a release and this PR wasn't considered ready yet. We might want to bring the functionality back, and fix the problem by merging this PR. @bkunda what do you think?

Issue #11883, opened 2022-06-03
Bracket / Brace can't be removed with [Delete]; no UNDO possible; leads to crash (Linux only)


rgreen5:
**Describe the bug**

Clicking on a bracket and pressing [Del] should delete. But this doesn't work and leads to a crash.

**To Reproduce**

1. Create a guitar staff/tab score.
2. Click on measure one and click on a bracket in the Brackets palette.
3. Click once on the bracket/brace and press [Del]
RESULT: Nothing happens (can crash at this time)
4. Click on the brace/bracket one or more times
RESULT: The program crashes.

**Expected behavior**

Pressing Delete should immediately remove the bracket/brace.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2432394363, revision: github-musescore-musescore-1ff2049
Doesn't occur on **Win** and **Mac**

Tantacrul:
Won't reproduce on Mac. (I'm using backspace to delete the bracket and brace, if that's useful to know).

@DmitryArefiev - can you reproduce?

RomanPudashkin:
@Tantacrul I can reproduce it

DmitryArefiev:
Oh.. @RomanPudashkin told me the secret and now I can reproduce it on all platforms. You need to use trackpad (not a mouse) and delete bracket from context menu

DmitryArefiev:
#11917 fixed related bug on Win and Mac with a little bit different steps.

The current issue is still occurs. But it's Linux only issue. I'll modify summary and description a little bit

wizofaus:
Was able to crash it running under Visual Studio on Windows. Again related to AccessibleRoot::m_focusedElement pointing to a deleted object, which leads to undefined behaviour.  While there are two PRs already that may prevent this from crashing, they won't prevent undefined behaviour, as neither prevent dereferencing m_focusedElement when it's pointing at a deleted object.

>	MuseScore4.exe!deto::async::AbstractInvoker::invoke(int type, const deto::async::NotifyData & data) Line 30	C++
 	MuseScore4.exe!deto::async::Channel<enum mu::accessibility::IAccessible::State,bool>::send(const mu::accessibility::IAccessible::State & <d_0>, const bool & <d_1>) Line 32	C++
 	MuseScore4.exe!mu::engraving::AccessibleItem::notifyAboutFocus() Line 106	C++
[...]

PR #12101, opened 2022-06-20
[MU4] Fix #12074: Сrash when moving bracket

### LINKED ISSUE #12074: [MU4 Issue] Сrash when moving bracket

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11883

### DIFF
--- a/src/engraving/layout/layout.cpp
+++ b/src/engraving/layout/layout.cpp
@@ -190,15 +190,6 @@ void Layout::doLayoutRange(const LayoutOptions& options, const Fraction& st, con
         }
     } else {
         for (System* s : m_score->_systems) {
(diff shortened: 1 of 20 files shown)


wizofaus:
I tried out this branch and now I can't use the mouse to resize brackets at all?  And if I use the keyboard (up/down arrows) it doesn't properly reselect after each keypress, so it only works once.  Likewise if I adjust the "Span" control in the Bracket properties inspector.
(these were the first things I tried because I noticed a lot of code had been modified that I remember putting there to support these operations).

PR #12102, opened 2022-06-20
[engraving] Replacement QString to String. Step 15

igorkorsukov:


RomanPudashkin (approved):

igorkorsukov:
Yes, I tried to use `std::basic_regex<char16_t>;`, but in the constructor it looks for the locale of `char16_t`, does not find it and sends an exception, the problem is in the lack of the locale of `char16_t`

Alternatively, we can make our own wrapper class to hide the complexity... this idea came to me late, I did not do it... In general, I don't like regular expressions, so I don't make it easier to use them, I hope we will use the less. But if we still need them often, then it's better to make a wrapper

PR #12110, opened 2022-06-20
fix #12074 - prevent crash after certain operations that delete elements

### LINKED ISSUE #12074: [MU4 Issue] Сrash when moving bracket

wizofaus:
Resolves:  https://github.com/musescore/MuseScore/issues/12074

Logic issue with destruction of AccessibleItem .

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -40,12 +40,11 @@ AccessibleItem::AccessibleItem(EngravingItem* e, Role role)
 
 AccessibleItem::~AccessibleItem()
 {
-    m_element = nullptr;
-
     AccessibleRoot* root = accessibleRoot();
     if (root && root->focusedElement() == this) {
         root->setFocusedElement(nullptr);
     }
+    m_element = nullptr;
 
     if (m_registred && accessibilityController()) {
         accessibilityController()->unreg(this);


src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -40,12 +40,11 @@ AccessibleItem::AccessibleItem(EngravingItem* e, Role role)
 
 AccessibleItem::~AccessibleItem()
 {
-    m_element = nullptr;
-
     AccessibleRoot* root = accessibleRoot();
     if (root && root->focusedElement() == this) {
         root->setFocusedElement(nullptr);
     }
+    m_element = nullptr;
```
wizofaus, line 47: accessibleRoot() returns nullptr if m_element is null, so root was never getting informed of our deletion

RomanPudashkin: 1. this change will cause regression of this crash:
https://github.com/musescore/MuseScore/pull/12029

2. I strongly recommend not fixing this crash now. The problem is that we very often recreate brackets. And we store pointers to these temporal objects in a lot of places. Such fixes only temporarily hide the global problem. My PR will improve the situation with brackets:
https://github.com/musescore/MuseScore/pull/12101

Yes, I know, it is a bit unstable right now, but I'm working on it

wizofaus: I don't actually think that's true, the undefined behaviour was happening via the call to `accessibilityController()->unreg(this);`.
And this code can crash as is without brackets being involved at all (deleting staff labels to the left of the staff etc.).
Further as it is currently ```if (root && root->focusedElement() == this) {
        root->setFocusedElement(nullptr);
    }``` is logically pointless code, as `root` will always be `nullptr`.

wizofaus:
But it's not just brackets, it can crash deleting other items too. And the current AccessibleItem destructor has logically inaccessible code...

PR #12121, opened 2022-06-21
fixed #12054: Learn page playlists fix

### LINKED ISSUE #12054: [MU4 Issue] Video tutorials (in 'Learn') are very often 'unable to load'
**Describe the bug**
As you can see below, this is the most common message I see when I click on the learn section, despite being connected to the internet. (MacOS). I'm wondering whether there is a refresh setting that could check from time to time to see if the user is connected, so the videos can display? It only sometimes works.

image

Eism:
Resolves: #12054

### DIFF
--- a/src/learn/ilearnservice.h
+++ b/src/learn/ilearnservice.h
@@ -34,6 +34,8 @@ class ILearnService : MODULE_EXPORT_INTERFACE
 public:
     virtual ~ILearnService() = default;
 
+    virtual void refreshPlaylists() = 0;
+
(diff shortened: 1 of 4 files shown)


cbjeukendrup:
Oops, now that I see the fix, I realize that I must have broken this while fixing memory leaks in #10702! Anyway, those memory leaks were also not good...

DmitryArefiev:
Now works fine in latest master builds (Win/Mac) on my side

Tantacrul:
Same!

Issue #12117, opened 2022-06-21
[MU4 Issue] [Assert crash] Crash on exporting to music xml


HemantAntony:
**Describe the bug**
Exporting to MusicXML crashes MuseScore

**To Reproduce**
1. Create a score with a piano
2. Go to export dialog
3. Switch to MusicXML export
4. Export, MuseScore crashes

**Expected behavior**
MuseScore is not supposed to crash

**Screenshots**
**I will add screenshots after a bit of time**

**Platform information**
 - OS: Windows 11

HemantAntony:
You should not be able to reproduce without being in Debug mode. So I don't think this will affect regular users. But still worth testing just in case

wizofaus:
Can't reproduce, running MSVC on Win 10 with latest source.

cbjeukendrup:
You can only reproduce it if the "All layout" option is checked in the export dialog.

wizofaus:
Nope, tried that...

PR #10456, opened 2022-02-05
[MU4] Fix #34451: Add rest position in musicxml export

Jojo-Schmitz:
Resolves: https://musescore.org/en/node/34451 (and replaces PR #9955)

Export rest position attributes (default-x, default-y) in the same way it has already been done for notes, by making the code for `notePosition()` generic, so it can handle both.

PR #12122, opened 2022-06-21
Fix assert in MusicXML Export

cbjeukendrup:
Resolves: #12117

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -3617,7 +3617,7 @@ QString ExportMusicXml::elementPosition(const ExportMusicXml* const expMxml, con
         const double pageHeight  = expMxml->getTenthsFromInches(expMxml->score()->styleD(Sid::pageHeight));
 
         const auto meas = elm->findMeasure();
-        IF_ASSERT_FAILED(!meas) {
+        IF_ASSERT_FAILED(meas) {
             return res;
         }
 


src/importexport/musicxml/internal/musicxml/exportxml.cpp
```diff
@@ -3617,7 +3617,7 @@ QString ExportMusicXml::elementPosition(const ExportMusicXml* const expMxml, con
         const double pageHeight  = expMxml->getTenthsFromInches(expMxml->score()->styleD(Sid::pageHeight));
 
         const auto meas = elm->findMeasure();
-        IF_ASSERT_FAILED(!meas) {
+        IF_ASSERT_FAILED(meas) {
             return res;
```
wizofaus, line 3621: Not a fan of his IF_ASSERT_FAILED macro, it's not obvious what it might from reading it.
Maybe
I'd actually prefer just 
```
  if (ASSERT_TRUE(!meas)) {
}
```
Such that the macro didn't invert the condition.

Jojo-Schmitz: DEBUG, Not RelWithDebInfo?

Jojo-Schmitz: OK, that might be the problem then

Anyway, seems that bug was a screwup of mine when porting @julienripoche`s #9955 in my #10456
And yes, that strange inverted lugic is what threw me off.

Jojo-Schmitz: Mind to check ef1a2cf whether that needs to get inverted too? @cbjeukendrup ?

PR #12123, opened 2022-06-22

Jojo-Schmitz:
reg. 'argument': conversion from 'size_t' to 'rapidjson::SizeType', possible loss of data (C4267)


Jojo-Schmitz:
in MinGW there are some warnings reg. "unknown option after '#pragma GCC diagnostic' kind [-Wpragmas]", all with the rapidjason stuff. At one place the code is hinting at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59124

Jojo-Schmitz:
Those pragma warnings in rapidjson should go away with #12142. Not sure about those this PR fixes, but I'd guess so, so let's get #12142 merged first

PR #12133, opened 2022-06-22
[MU4] Fix typo, including compalibility check

Jojo-Schmitz:
Same issue with MU3


Jojo-Schmitz:
No idea why, after a needed rebase to resolve a merge conflict, it now fails the vtests?

src/engraving/style/style.cpp
```diff
@@ -328,6 +328,8 @@ void MStyle::read(XmlReader& e, compat::ReadChordListHook* readChordListHook)
             set(Sid::dontHideStavesInFirstSystem, e.readBool());
         } else if (tag == "beamDistance") { // beamDistance maps to useWideBeams in 4.0
             set(Sid::useWideBeams, e.readDouble() > 0.75);
+        } else if (tag == "pedalListStyle") { // pre-3.6.3/4.0 typo
+            set(Sid::pedalLineStyle, e.readInt());
```
cbjeukendrup, line 332: I think it should nowadays be:
```suggestion
            set(Sid::pedalLineStyle, TConv::fromXml(e.readAsciiText(), LineType::SOLID));
```
That will probably make the VTests happy.

Jojo-Schmitz: Hmm, that is very different from the other special cases above this one?!?

Jojo-Schmitz:
Maybe to late to merge this into master now, as it'd cause 4.0.0 and 4.0.1 to not read that setting properly.

cbjeukendrup:
Yeah, that's a good point... I've added it to the list of things to consider when changing the file format, so that we won't forget about it again.

Jojo-Schmitz:
> Yeah, that's a good point... I've added it to the list of things to consider when changing the file format, so that we won't forget about it again.

Where's that list?

igorkorsukov (approved):

PR #12142, opened 2022-06-23
[engraving] Replaced QJsonXXX to JsonXXX

igorkorsukov:


vpereverzev (approved):

src/framework/global/types/bytearray.h
```diff
@@ -37,7 +37,7 @@ class ByteArray
 
     ByteArray();
     ByteArray(const uint8_t* data, size_t size);
-    ByteArray(const char* str);
+    ByteArray(const char* str, size_t size = static_cast<size_t>(-1));
```
Jojo-Schmitz, line 40: not `mu::nidx`?

igorkorsukov: I'm not sure that `mu::nidx`  is semantically suitable here, I didn't want to make unnecessary dependencies

PR #12183, opened 2022-06-27
fixed #12179: Fixed invalid pointers

### LINKED ISSUE #12179: [MU4 Issue] Crash after press Space two times when entering lyrics
**To Reproduce**
1. Enter lyrics mode (Ctr/Cmd+L)
2. Hit Space two times

https://user-images.githubusercontent.com/90187801/175990975-2788de70-a483-446f-9a6e-7042b58eb255.mp4

**Platform information**
macOS, Windows, Linux

Eism:
Resolves: #12179
Resolves: #12074

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -42,11 +42,6 @@ AccessibleItem::~AccessibleItem()
 {
     m_element = nullptr;
 
-    AccessibleRoot* root = accessibleRoot();
-    if (root && root->focusedElement() == this) {
-        root->setFocusedElement(nullptr);
-    }
-
     if (m_registred && accessibilityController()) {
         accessibilityController()->unreg(this);
         m_registred = false;
@@ -84,7 +79,7 @@ AccessibleRoot* AccessibleItem::accessibleRoot() const
     }
 
     RootItem* rootItem = m_element->explicitParent() ? score->rootItem() : score->dummy()->rootItem();
-    return dynamic_cast<AccessibleRoot*>(rootItem->accessible());
+    return dynamic_cast<AccessibleRoot*>(rootItem->accessible().get());
 }
 
 const EngravingItem* AccessibleItem::element() const
@@ -113,7 +108,7 @@ const IAccessible* AccessibleItem::accessibleParent() const
         return nullptr;
     }
 
-    return static_cast<EngravingItem*>(p)->accessible();
+    return static_cast<EngravingItem*>(p)->accessible().get();
 }
 
 size_t AccessibleItem::accessibleChildCount() const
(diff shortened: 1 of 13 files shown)


DmitryArefiev:
Tested #12179 #12074 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -42,11 +42,6 @@ AccessibleItem::~AccessibleItem()
 {
     m_element = nullptr;
 
-    AccessibleRoot* root = accessibleRoot();
-    if (root && root->focusedElement() == this) {
-        root->setFocusedElement(nullptr);
-    }
-
     if (m_registred && accessibilityController()) {
         accessibilityController()->unreg(this);
```
wizofaus, line 46: undefined behaviour - accessibilityController->unreg( ) calls a virtual function on `this`, which is partially destructed here

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -84,7 +79,7 @@ AccessibleRoot* AccessibleItem::accessibleRoot() const
     }
 
     RootItem* rootItem = m_element->explicitParent() ? score->rootItem() : score->dummy()->rootItem();
-    return dynamic_cast<AccessibleRoot*>(rootItem->accessible());
+    return dynamic_cast<AccessibleRoot*>(rootItem->accessible().get());
```
wizofaus, line 82: why do we need dynamic_cast? (yes I know it was already there...)

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -113,7 +108,7 @@ const IAccessible* AccessibleItem::accessibleParent() const
         return nullptr;
     }
 
-    return static_cast<EngravingItem*>(p)->accessible();
+    return static_cast<EngravingItem*>(p)->accessible().get();
```
wizofaus, line 111: any reason not change signature to return smart pointer?

src/engraving/accessibility/accessibleitem.cpp
Eism: It is not about all virtual functions, see the case with Score::sigmap()
When we try to call the child method in EngravingItem::barbeat, the parent method is called

src/engraving/accessibility/accessibleitem.cpp
Eism: since accessible() is the parent, we need to return the child

src/engraving/accessibility/accessibleitem.cpp
Eism: because we don't store this object in this class, we get it from element and just return

src/engraving/accessibility/accessibleitem.cpp
wizofaus: Ah ok, I misread - AccessibleRoot derives from AccesibleItem, which is a bit counter-intuitive.  Still, I probably would have defined RootItem to return an AccessibleRoot rather than do the case here.

src/engraving/accessibility/accessibleitem.cpp
wizofaus: Sure, but now you've lost track of the fact that there's a copy of the pointer somewhere...

Eism: Thanks, I created tech debt. We can't implement it now because it will require a lot of changes
https://github.com/musescore/MuseScore/issues/12208

PR #12194, opened 2022-06-28
[engraving] Removed Qt dependencies. Step 3

igorkorsukov:


handrok (approved):

src/framework/global/tests/string_tests.cpp
```diff
@@ -53,6 +53,9 @@ TEST_F(Global_Types_StringTests, Char_Digit)
         EXPECT_FALSE(ch.isDigit());
         EXPECT_EQ(ch.digitValue(), -1);
     }
+
+    QString str = "ggg";
+    bool ret = str >= 7;
```
cbjeukendrup, line 58: Looks like this is either incomplete or a left-over :)

igorkorsukov: oops )

handrok (approved):

Issue #10096, opened 2021-12-15
[MU4 Issue] Enter key doesn't work as expected when using Ctrl+F / Find


wizofaus:
**Describe the bug**
Ctrl+F brings up the "Find / Go To" search field, but after typing in the desired location (e.g. 2 for measure 2), hitting enter to close the search field and return focus to the score doesn't work as per MuseScore 3, instead seemingly inserting a line break if the notation window did have the keyboard focus, or nothing at all otherwise, and in neither case dismissing the search field.

**To Reproduce**
1. Open/create any score
2. Use the keyboard shortcut Ctrl+F
3. Type 2 (measure 2 should get selected) then enter
4. Nothing happens (except that the search field no longer has the focus so keystrokes activate shortcuts etc.)
5. Now click somewhere on the score and use Ctrl+F again
6. Type 2 (measure 2 should get selected) then enter
7. Line break is inserted but search field is not dismissed 

**Expected behavior**
As per MuseScore 3, enter key should dismiss the search field and return focus to the score

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

Kissh12:
can I work on this issue?

Tantacrul:
@Kissh12 - it doesn't seem like there has been much movement? Are you intending on working on this or shall I reassing?

SoAsEr:
What's the intended visual behavior? Do we want to keep the live preview? What is the advantage of the live preview if it doesn't adjust the scroll?

wizofaus:
What do you mean by live preview?  I described the expected behaviour pretty clearly.
It's still behaving the same way - esp. problematic with enter very unexpectedly adding a line break.
I guess I can have a look at this but the way Qt handles keyboard events is painful...

SoAsEr:
What I did was eliminate the as you type it highlights the correct measure and instead only made it highlight and jump (the jump did not happen previously) when the user pressed enter/return.  Is that desired or should the editor never jump? Or should it jump at every key press?

wizofaus:
Personally I wouldn't have an issue with that, but not my call... @Tantacrul @bkunda ?
Would it be particularly hard to leave in the instant-jump-on-keypress?  I can see it being inefficient for super large scores where you want to type, e.g. "296" which causes it to jump to measure 2, then 29, then 296...

SoAsEr:
[...]

PR #12198, opened 2022-06-28
Fix find

SoAsEr:
Resolves https://github.com/musescore/MuseScore/issues/10096

The find dialog was totally broken. This makes it usable with enter and return, and makes it reachable with keyboard navigation.

I changed the dialog to only activate when "enter" was pressed to avoid jumping around unnecessarily. I made it highlight the first measure of a page if we are searching for a page.

I think there should be some explanation of the format required to search for rehearsal marks or pages, but that was outside of the scope of the issue.

I hope this pull request is okay!

- [] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [] I created the test (mtest, vtest, script test) to verify the changes I made

### DIFF
--- a/src/engraving/layout/layoutmeasure.cpp
+++ b/src/engraving/layout/layoutmeasure.cpp
@@ -33,6 +33,7 @@
 #include "libmscore/lyrics.h"
 #include "libmscore/marker.h"
 #include "libmscore/part.h"
+#include "libmscore/textlinebase.h"
 
 #include "layout.h"
 #include "layoutcontext.h"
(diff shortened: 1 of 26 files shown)


src/out/build/x64-Debug/.cmake/api/v1/query/client-MicrosoftVS/query.json
```diff
@@ -0,0 +1 @@
+{"requests":[{"kind":"cache","version":2},{"kind":"cmakeFiles","version":1},{"kind":"codemodel","version":2}]}
```
cbjeukendrup, line 1: You accidentally added the `src/out/` folder to the PR. Please remove it :)

cbjeukendrup (commented):
A couple of code-style related comments. Also take a look at the automated code style check for the C++ code, which is currently failing: https://github.com/musescore/MuseScore/runs/7101353896?check_suite_focus=true

PR #12199, opened 2022-06-28
Fix raising the main window on macOS

cbjeukendrup:
There is an issue when raising the main window on macOS (when another instance needs to be activated): instead of being raised and activated, the main window stays at the z-index where it was, and only _resizes_.

That's fixed with this PR. The solution works flawlessly on macOS, even when the window is minimized, and it also works well regarding keyboard focus. 

However, considering how complicated this code was before this PR, there is probably a reason that it was the way it was. If this solution indeed breaks it on other platforms, I'll just put it in an `#ifdef Q_OS_MAC`.


cbjeukendrup:
I updated this PR to make the change macOS-only, since indeed it causes problems on Windows.

For testing: 
1. Open a score
2. Open another score, which opens in a different instance
3. In that new instance, go to "File" > "Open recent" and try to re-open that same file as you originally opened in the first instance. The window containing this file should appear correctly.

igorkorsukov:
@cbjeukendrup could you please add comment near each #ifdef Q_OS_MAC about this

cbjeukendrup:
@igorkorsukov Done!

igorkorsukov (approved):

DmitryArefiev:
@cbjeukendrup Tested on Mac13.1 - **looks fine**. Should I test Win/Linux?

cbjeukendrup:
@DmitryArefiev Thanks! macOS is enough; for Windows and Linux the code is unchanged.

Issue #12188, opened 2022-06-28
[MU4 Task] Publish to MuseScore.com should be hidden until it's properly implemented on musescore.com


DmitryArefiev:
**Task description**
Just hide **Publish to MuseScore.com** button grey out on Publish tab (and in File menu)

![Capture](https://user-images.githubusercontent.com/90187801/176138384-5af05e2e-31cd-45fd-9920-874a9688a581.PNG)

DmitryArefiev:
Hi @cbjeukendrup , let's just **hide the button** (from Publish tab and File menu). I'll rewrite the summary.
Thanks!

cbjeukendrup:
@DmitryArefiev Thanks, I updated the PR!

DmitryArefiev:
Fixed in #12200

PR #12200, opened 2022-06-28
Disable saving/publishing to cloud until it is fully implemented

cbjeukendrup:
Resolves: #12188

### DIFF
--- a/src/appshell/view/appmenumodel.cpp
+++ b/src/appshell/view/appmenumodel.cpp
@@ -139,8 +139,9 @@ MenuItem* AppMenuModel::makeFileMenu()
         makeMenuItem("file-save-as"),
         makeMenuItem("file-save-a-copy"),
         makeMenuItem("file-save-selection"),
-        makeMenuItem("file-save-to-cloud"),
-        makeMenuItem("file-publish"),
+        // Todo(save-to-cloud): hidden until fully implemented
+        //makeMenuItem("file-save-to-cloud"),
+        //makeMenuItem("file-publish"),
         makeSeparator(),
         makeMenuItem("file-import-pdf"),
         makeMenuItem("file-export"),
--- a/src/appshell/view/publish/publishtoolbarmodel.cpp
+++ b/src/appshell/view/publish/publishtoolbarmodel.cpp
@@ -22,8 +22,6 @@
 
 #include "publishtoolbarmodel.h"
 
-#include "log.h"
-
 using namespace mu::appshell;
 using namespace mu::uicomponents;
 
@@ -38,7 +36,8 @@ void PublishToolBarModel::load()
 
     MenuItemList items {
         makeMenuItem("print"),
-        makeMenuItem("file-publish"),
+        // Todo(save-to-cloud): hidden until fully implemented
+        //makeMenuItem("file-publish"),
         makeMenuItem("file-export")
     };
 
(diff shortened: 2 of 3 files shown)


DmitryArefiev:
Tested on Win/Mac

@cbjeukendrup I've noticed two additional changes:
1. Save to cloud is also removed from File menu
2. When triggering Export window on Publish page, it  greys out Print and Export buttons

https://user-images.githubusercontent.com/90187801/177131204-7212b481-de35-44f0-b7de-6f385721bf49.mp4

The first point is make sense but I have a doubts about №2

cbjeukendrup:
@DmitryArefiev Thanks, that should be better now! And yes, it was on purpose that I also hid the actions in the File menu. 

The PR also fixes another issue: before, the File > Save, File > Export and File > Print actions were always disabled when you are in Publish mode. Now, those actions will also be enabled in Publish mode, because I thought that would make sense (since "print" and "export" are enabled in the toolbar too).

DmitryArefiev:
> @DmitryArefiev Thanks, that should be better now! And yes, it was on purpose that I also hid the actions in the File menu.
> 
> The PR also fixes another issue: before, the File > Save, File > Export and File > Print actions were always disabled when you are in Publish mode. Now, those actions will also be enabled in Publish mode, because I thought that would make sense (since "print" and "export" are enabled in the toolbar too).

@cbjeukendrup Yeah, now it's OK. Thanks!

DmitryArefiev:
Tested #12188 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

src/project/internal/projectuiactions.cpp
```diff
@@ -38,23 +38,23 @@ const UiActionList ProjectUiActions::m_actions = {
...
              ),
     UiAction("file-close",
-             mu::context::UiCtxNotationOpened,
+             mu::context::UiCtxNotationOrPublishOpened,
              QT_TRANSLATE_NOOP("action", "Close"),
              QT_TRANSLATE_NOOP("action", "Close current score")
              ),
     UiAction("file-save",
-             mu::context::UiCtxNotationOpened,
+             mu::context::UiCtxNotationOrPublishOpened,
```
RomanPudashkin, line 46: Looks like now the user won't be able to save their score from the home page

cbjeukendrup: That was already the case, and will indeed also be the case after this PR. Personally, I could see a way that this makes sense from a UX point of view, because on the home page, you don't _see_ the score, and on the other pages you do. Not a very strong point though, so if you think these actions should also be available on the homepage too, then I'll change them to `UiCtxAny`.

RomanPudashkin: >Not a very strong point though, so if you think these actions should also be available on the homepage too, then I'll change them to UiCtxAny.

@Tantacrul what do you think?

RomanPudashkin: UiCtxAny is not suitable in this case, since these actions should be blocked if the user does not have any score open

RomanPudashkin: I would recommend removing all these changes from your PR and keeping only what is relevant to the original issue

cbjeukendrup: Yeah, that's the best for now. Done!

RomanPudashkin (approved):

Issue #12001, opened 2022-06-14
[Optional Task] Add Finale's Maestro and Broadway fonts


Tantacrul:
**Task description**
This is a task that does not need to be implemented for 4.0. However, if a community member is interested, these are the fonts that it would make the most sense for us to include:

- **Finale Maestro (& Finale Maestro Text)**
- **Finale Broadway (& Finale Broadway Text**

You can get them here: https://makemusic.zendesk.com/hc/en-us/articles/1500013053461-MakeMusic-Fonts-and-Licensing-Information

Note: we would only merge this if the fonts have no issues, which should be the case if they are now indeed proper SMuFL fonts, especially after the cleanup we did for SMuFL support last year.

Jojo-Schmitz:
Unfortunaltely they are not just plain downloadable, you'd have it download an msi (Windows) or pkg (Mac) file and install them, to then extract, add to MuseScore and uninstall them again.

cbjeukendrup:
I'v just extracted them from the pkg installer and will put the extracted versions in MuseScore 😎

Jojo-Schmitz:
Yeah, it seems possible to open the msi file with 7Zip, as I just found out

PR #12203, opened 2022-06-28
Add Finale Maestro and Finale Broadway fonts

cbjeukendrup:
Resolves: #12001

At a first glance, I see no problems with these fonts.


RomanPudashkin (approved):

Tantacrul:
@oktophonie - we have a little task for you! :)

Tantacrul:
I've noticed that the ottava marks are fine for Broadway but not for Maestro (at least, I doubt Maestro is meant to look like this...)

 **Broadway** 
image

**Maestro**
image

cbjeukendrup:
That '8' seems to be a Maestro Text problem. I believe we are using the "Text" variant of the font in that place, but Maestro Text does not contain the ottava symbols, also not that '8' which should be at U+E510. However, at U+E888 I seem to recognize that small 8 that I see in your example. And I suspect that the '15' is from either Leland or Bravura (whichever is used as a fallback).

Tantacrul:
Thought so. Perhaps I can modify Maestro text for now (until they fix it themselves)?

cbjeukendrup:
> 2\. I don't think we use the text fonts for 8ve markings - mainly because I never added them to LelandText.

We really are. That would mean that ottava markings fall back to Bravura, and that seems to be the case...

Tantacrul:
Got it. Thanks!

PR #12204, opened 2022-06-29
fixes #10096

### LINKED ISSUE #10096: [MU4 Issue] Enter key doesn't work as expected when using Ctrl+F / Find

SoAsEr:
Resolves https://github.com/musescore/MuseScore/issues/10096

The find dialog was totally broken. This makes it usable with enter and return, and makes it reachable with keyboard navigation.

I changed the dialog to only activate when "enter" was pressed to avoid jumping around unnecessarily. I made it highlight the first measure of a page if we are searching for a page.

I think there should be some explanation of the format required to search for rehearsal marks or pages, but that was outside of the scope of the issue.

Third attempt. I just completely deleted the branch and manually copied in my changes and then ran it through uncrustify.

[x] I signed [CLA](https://musescore.org/en/cla)
[x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
[x] I made sure the code compiles on my machine
[x] I made sure there are no unnecessary changes in the code
[x] I made sure the title of the PR reflects the core meaning of the issue you are solving
[x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
[ ] I made sure the commit message title starts with "fix #424242:" if there is a related issue
[ ] I created the test (mtest, vtest, script test) to verify the changes I made


src/notation/view/searchpopupmodel.cpp
```diff
@@ -32,14 +32,22 @@ void SearchPopupModel::load()
     });
 }
 
-void SearchPopupModel::search(const QString& text)
+bool SearchPopupModel::search(const QString& text)
 {
     mu::engraving::EngravingItem* element = notation()->elements()->search(text.toStdString());
     if (element) {
+        if (element->isPage()) {
```
SoAsEr, line 39: Should rehearsal marks also have the same behavior of selecting the rehearsal marks they're attached to?

src/notation/view/searchpopupmodel.cpp
```diff
@@ -32,14 +32,22 @@ void SearchPopupModel::load()
...
+            const mu::engraving::Page* p = static_cast<const mu::engraving::Page*>(element);
+            element = p->firstMeasure();
+            if (!element && !p->systems().empty() && !p->systems().front()->measures().empty()) {
+                element = p->systems().front()->measures().front();
+            }
+        }
         notation()->interaction()->select({ element }, SelectType::SINGLE);
```
SoAsEr, line 46: Here we select the element but do not end note input mode, meaning that we jump to a measure and select it, but if we continue to write music, we will jump back to where we were. I don't know whether or not this is desirable and would like some feedback from a design person

HemantAntony:
Can you give this PR a better name that is related to this issue. You can see other already existing PRs when you click on 'pull requests' in the top panel.

src/engraving/libmscore/page.cpp
```diff
@@ -180,16 +194,16 @@ Text* Page::layoutHeaderFooter(int area, const String& ss) const
         if (!text) {
             text = Factory::createText((Page*)this, TextStyleType::HEADER);
             text->setFlag(ElementFlag::MOVABLE, false);
-            text->setFlag(ElementFlag::GENERATED, true);       // set to disable editing
+            text->setFlag(ElementFlag::GENERATED, true);           // set to disable editing
```
wizofaus, line 197: what's actually changed here?

SoAsEr: I touched nothing this was uncrustify

src/framework/uicomponents/qml/MuseScore/UiComponents/TextInputField.qml
```diff
@@ -173,18 +174,25 @@ FocusScope {
...
                     event.accepted = false
-
                     root.focus = false
                     root.textEditingFinished()
                 }
             }
 
             Keys.onPressed: function(event) {
+                //enter and return don't go through this event
+                //https://stackoverflow.com/questions/9677371/qml-keys-onenterpressed-issue
```
wizofaus, line 184: don't think adding this is all that helpful but ok

src/engraving/libmscore/page.cpp
```diff
@@ -309,7 +323,7 @@ double Page::footerExtension() const
 
 void Page::scanElements(void* data, void (* func)(void*, EngravingItem*), bool all)
 {
-    for (System* s :_systems) {
+    for (System* s : _systems) {
```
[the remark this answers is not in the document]
SoAsEr: All of this changed when I ran uncrustify. I did nothing here

src/notation/view/searchpopupmodel.cpp
wizofaus: Any reason why they shouldn't?

wizofaus:
I've tested this @Tantacrul and it appears fine to me, definitely a big improvement over the current behaviour.  No reason not to merge that I can see...

Tantacrul:
Yeah. Seems fine to me too. @RomanPudashkin / @Eism - same request, if you are happy.

src/engraving/libmscore/page.cpp
cbjeukendrup, line 1: Please undo the unrelated changes to the whitespace. If these were really made by Uncrustify, make sure that you are using version 0.73.0 and that it reads from the correct config file. (Or just don't run Uncrustify, and format the code correctly yourself, and rely on the GitHub Actions code style check.)

src/engraving/libmscore/page.h
cbjeukendrup, line 1: Same about the whitespace changes in this file

cbjeukendrup (commented):
A rebase is needed. Also, here are some review comments; it turns out that some of them are only cosmetic, but that's important too :)

Eism:
@SoAsEr @cbjeukendrup 
It seems to me that this PR is no longer relevant since the navigation was fixed in #14485 and #14515, and we closed the issue
Or is there something I'm missing?

cbjeukendrup:
Looks like you're right, @Eism!

Sorry @SoAsEr! I think we'll have to close this. If you think that this PR still contains something that's missing in the current master branch, please create a fresh new PR for that.

PR #12213, opened 2022-06-29
[engraving] Removed QRegularExpression from Lyrics::layout

igorkorsukov:


igorkorsukov:
> std::regex is famously terribly implemented in every standard library and can never be changed due to ABI breaks. I don't know what QT's implementation is, but hopefully its closer to boost's than the standard libraries. Is this actually beneficial?

We have a strategic goal to remove the dependence of engraving on Qt and anything, only std.

I changed the implementation without using regular expressions, I believe it's easier and more efficient

cbjeukendrup:
I guess the text in this sample code on cppreference must be interpreted as a hint 😉
Schermafbeelding 2022-06-29 om 23 21 03

SoAsEr:
There are good implementations out there (Boost and Re2 come to mind). With c++20 and the new contexpr possibilities we could get a really good implementation.

igorkorsukov:
> I guess the text in this sample code on cppreference must be interpreted as a hint wink Schermafbeelding 2022-06-29 om 23 21 03



> I guess the text in this sample code on cppreference must be interpreted as a hint wink Schermafbeelding 2022-06-29 om 23 21 03

Problems with Unicode (

PR #12214, opened 2022-06-29
Fix #10205: Beam groups calculation fix

### LINKED ISSUE #10205: [MU4 Issue] Beamlets (fractional beams) go in the wrong direction
**Describe the bug**

The vtests for beams-11.musx are currently wrong. Specifically, the first beamlet in each of the bottom two measures should face the other direction.

Screen Shot 2022-01-03 at 5 16 28 PM

**Expected behavior**

MuseScore 3 currently renders the beamlets like this. This is desired.

Screen Shot 2022-01-03 at 5 17 48 PM

 - OS: macOS 12.1, Apple Silicon

**Additional context**

This was caused by beams rewrite #9792.

The line of code that fixed this in MS3 is 

https://github.com/musescore/MuseScore/blob/4291c9ef4a3eb63f5f0f04cd0b69575aa70db409/src/engraving/libmscore/beam.cpp#L1970-L1972

Related: https://musescore.org/en/node/42856

asattely:
There are some considerations that went into this one. Generally, the idea is that if notes are in the middle of a beam group, their beamlets go right, unless they are the last note of a group, in which case they go left.

image
In this case, the beamlet finds itself at the end of a beam group, so the beamlet goes left.

image
In this case, the beamlet finds itself in the middle of a beam group, so the beamlet goes right.

image
This is the interesting one. The beamlet is at the end of a beam group, so it goes left. But also, according to the 1/16 beam groups, the primary beam is also broken. According to the 1/8 beam group, the primary beam seems to be unbroken the whole way through the bar.

The way this PR addresses the disagreement is to determine where the third note falls. It doesn't happen to fall on any concrete 8th beat, so the next smallest note length is checked. It _does_ happen to fall on a 16th beat, so the beam groups for that one are used, and the beam group dictates that there is a clean break in the primary beam.

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -978,11 +978,13 @@ void Beam::createBeamSegments(const std::vector<ChordRest*>& chordRests)
         bool breakBeam = false;
         bool previousBreak32 = false;
         bool previousBreak64 = false;
+        int prevRests = 0;
 
         for (uint i = 0; i < chordRests.size(); i++) {
             ChordRest* chordRest = chordRests[i];
             ChordRest* prevChordRest = i < 1 ? nullptr : chordRests[i - 1];
             if (!chordRest->isChord()) {
+                prevRests++;
                 continue;
             }
             Chord* chord = toChord(chordRest);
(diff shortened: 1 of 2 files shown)


asattely:
Resolves https://github.com/musescore/MuseScore/issues/10205
This PR also addresses the issue @oktophonie linked in the original issue:
image

asattely:
Fixed a couple of issues that was causing some very weird beams in the utests to go crazy.

oktophonie (approved):

RomanPudashkin (approved):

PR #12227, opened 2022-06-30
[MU4] Fix MSVC/MinGW compiler warnings, fix MinGW compiler error

Jojo-Schmitz:
* Fix MSVC compiler warnings:
    * reg. declaration hides class member (C4458)
    * reg. macro redefinition (C4005)
    * reg. conversion from 'size_t' to 'int', possible loss of data (C4267)
    * reg. unreachable code (C4702)

* Fix MinGW compiler error:
    * reg. a header file not found
    Still doesn't link though, due to "undefined reference to `PKEY_Device_FriendlyName'"

* Fix MinGW compiler warning:
    * reg. comparison of integer expressions of different signedness (-Wsign-compare)

* Fix MOC compiler warning:
    * reg. name 'label' is already in use


vpereverzev:
I'd encourage to not touch 3rd-party sources here

src/appshell/view/devtools/settingslistmodel.cpp
```diff
@@ -103,6 +103,7 @@ QString SettingListModel::typeToString(Val::Type t) const
     case Val::Type::Color:     return "Color";
     case Val::Type::List:      return "List";
     case Val::Type::Map:       return "Map";
+    default: break;
```
Jojo-Schmitz, line 106: Will conflict with #12239

thirdparty/google_crashpad_client/third_party/mini_chromium/mini_chromium/base/compiler_specific.h
```diff
@@ -79,7 +79,9 @@
 #elif defined(__clang__)
 #define FALLTHROUGH [[clang::fallthrough]]
 #else
+#ifndef FALLTHROUGH
 #define FALLTHROUGH
 #endif
```
cbjeukendrup, line 84: I think we should rather rename our own macro to MU_FALLTHROUGH (in log.h)

cbjeukendrup:
Probably we'll have to disable the warning in flac by adding another `target_no_warning` in thirdparty/flac/CMakeLists.txt.

thirdparty/google_crashpad_client/third_party/mini_chromium/mini_chromium/base/compiler_specific.h
Jojo-Schmitz: Ah, yes, i didn't even check where the conflict is

Jojo-Schmitz:
I'll go that a try

thirdparty/google_crashpad_client/third_party/mini_chromium/mini_chromium/base/compiler_specific.h
Jojo-Schmitz: Done. Actually that marco isn't really needed as `// FALLTHROUGH` comment works just fine cross-platform and is used in several places already, whereas that `FALLTHROUGH()` marco is use in only 2 places (and one file)

PR #12241, opened 2022-07-02
fix #12225, support style file import path

### LINKED ISSUE #12225: [MU4 Task] Preferences > Import > Style used for import not yet connected
In Preferences > Import, the "Style used for import" setting is not yet connected. Its value is stored in the MusicXML* configuration, but never read from there. 

\* If this is a MusicXML specific setting, it should be in the MusicXML section on the page; if it is a general import setting, it should not be stored in the MusicXML configuration but in a more general place.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/12225)

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -150,7 +150,7 @@ mu::Ret NotationProject::load(const io::path_t& path, const io::path_t& stylePat
 
     std::string suffix = !format.empty() ? format : io::suffix(path);
     if (!isMuseScoreFile(suffix)) {
-        return doImport(path, stylePath, forceMode);
+        return doImport(path, stylePath.empty() ? notationConfiguration()->styleFileImportPath() : stylePath, forceMode);
     }
 
     MscReader::Params params;
--- a/src/appshell/view/preferences/importpreferencesmodel.cpp
+++ b/src/appshell/view/preferences/importpreferencesmodel.cpp
@@ -87,7 +87,7 @@ QString ImportPreferencesModel::fileDirectory(const QString& filePath) const
 
 QString ImportPreferencesModel::styleFileImportPath() const
 {
-    return musicXmlConfiguration()->styleFileImportPath().toQString();
+    return notationConfiguration()->styleFileImportPath().toQString();
 }
 
 QString ImportPreferencesModel::currentOvertureCharset() const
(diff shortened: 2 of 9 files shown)


src/project/internal/notationproject.cpp
```diff
@@ -150,7 +150,7 @@ mu::Ret NotationProject::load(const io::path_t& path, const io::path_t& stylePat
 
     std::string suffix = !format.empty() ? format : io::suffix(path);
     if (!isMuseScoreFile(suffix)) {
-        return doImport(path, stylePath, forceMode);
+        return doImport(path, stylePath.empty() ? notationConfiguration()->styleFileImportPath() : stylePath, forceMode);
```
cbjeukendrup, line 153: Hm... now that I see that `styleFileImportPath()` is only used inside the project module, I think it should probably be in ProjectConfiguration... sorry.

wizofaus: I did think that...

wizofaus:
What's happening with this? Do I need to request a review from the core team?

Tantacrul:
@wizofaus - I've just installed this PR. Can you let me know how to test this?  It's one of those features I've never used. I'll then advocate for merge city the moment I've confirmed it works! Sorry for delay.

wizofaus:
The linked issue explains what the problem is...

cbjeukendrup:
@Tantacrul In Preferences, go to the Import page, choose "use style file:" and select a .mss style file. This style file will be used when importing for example a MusicXML file (or any imported non-MuseScore file type afaik). 

To get a .mss style file, just create a score, change some style settings in a noticeable way, use Format > Save Style… to save it.

Tantacrul:
> The linked issue should explain what the problem is...but I agree it needs repro steps etc. As far as I understand, it's the default set of styles used for scores created by importing other formats (musicxml etc.)

I read it but couldn't figure it out. Otherwise I wouldn't have asked. Also don't have the time to spend with trial and error figuring out what it is.

Tantacrul:
Well, based on Casper's explanation, I can confirm it is working (despite there being quite a few styles that are not saving / loading properly - but that's a separate issue)

Of course, it is an almost completely incomprehensible setting. Glad I know what it does so we can eventually replace it with something that makes some kind of intuitive sense.

@RomanPudashkin / @Eism - I recommend merging if you are happy with the PR.

RomanPudashkin:
@wizofaus please rebase it

wizofaus:
Not clear why it needs it but OK, done

Issue #12237, opened 2022-07-01
[MU4 Issue] Wrong pitch for MuseScore General playback device in exported audio (Windows)


DmitryArefiev:
https://user-images.githubusercontent.com/90187801/176912302-6eaadfcd-ba5e-4a61-8984-c5f9024515d5.mp4

**Platform information**
Windows

**Additional context**
Changing sample rate export setting to 32kHz or 48kHz doesn't fix the issue

Tantacrul:
Also happening on MacOS. I changed this to P0.

vpereverzev:
Fixed in #12566

Issue #11267, opened 2022-04-20
[MU4 Issue] MuseScore General Piano sample pitch is lower for some cents


DmitryArefiev:
**Describe the bug**
MuseScore General Piano sample pitch is lower for some cents comparing with VST piano

![Capture](https://user-images.githubusercontent.com/90187801/164258537-e1990679-c799-40bb-9ec1-4da124f2845c.PNG)

**Platform information**

**Additional context**

cbjeukendrup:
I have the following theory about this. A while ago, we had this issue: #9225. The playback from the soundfont was approximately three quarter tones (150 cent) too high. We suspected that it had to do with a mismatch between the sample rate used by Fluid (44100 Hz) and the sample rate used by the rest (48000 Hz). As a temporary solution, Fluid was tuned down by exactly 150 cents.

If you take A = 440 Hz, multiply that by 48000/44100, convert it to cents, subtract 150 cents, and convert back to Hz, you get approximately 439 Hz, which corresponds with the screenshot in the issue description. Coincidence?

I also think I found out _why_ the sample rate of Fluid does not match. The order of things is currently the following:
1. a FluidSynth object is created, and it calls `init()` in the constructor
2. in `init()`: if sampleRate is already initialized (`> 0`), then set the sampleRate to the `fluid_settings_t` object (never happens, because sampleRate is _not_ already initialized, so the value in the `fluid_settings_t` object will stay at the default of 44100 Hz)
3. still in `init()`: a `fluid_synth_t` object is created from the settings object
4. Later: `setSampleRate` gets called, to initialize the sample rate. 
5. in `setSampleRate`: the new sampleRate is written to the `fluid_settings_t` object, but the `fluid_synth_t` object, that had already been created by the time that `setSampleRate` is called, is _not_ updated! (Because `"synth.sample-rate"` is not a "real-time" setting.)

@vpereverzev I'm not sure about how to fix this myself, but maybe it is helpful information for you. Or if you're already aware of all of this, sorry for the noise then :)

DmitryArefiev:
Tested in latest master on Win10, Mac12, LinuxMint 20.2 - **Now A sounds like in MU3. FIXED**

Closing

Issue #10804, opened 2022-03-14
[MU4 Issue] Main pitch should not change after changing default sample rate in audio device settings (Windows10, Realtek Audio Device) 


DmitryArefiev:
**Describe the bug**
On my ASUS laptop I found weird behavior when MU changes its pitch after changing sample rate in Realtek Audio Default settings (Internal Audio Device). 
Pitch should not change and be correct (like it behaves in **MU3**)

Pitch is **not correct** when sample rate is **44kHz**:

https://user-images.githubusercontent.com/90187801/158224147-9144ab40-0513-4698-8777-c46a6080e7e5.mp4

Pitch is **correct** when sample rate is **48kHz**:

https://user-images.githubusercontent.com/90187801/158224292-edc5d672-4e0f-4c79-bd87-30945562050b.mp4

**Expected behavior**
Pitch should be correct in 44kHz and in 48kHz

**Platform information**
Windows10, Realtek Audio Device, ASUS Laptop VivoBook

**Additional context**
Doesn't occur in **MU3**
Should be rechecked after fixing #11748

DmitryArefiev:
related to #9225

Tantacrul:
Two questions:

1. Is this still a problem?
2. Shouldn't this be targeted at **Beta 1**, at least? Also, doesn't it seem more like a P1 than a P2? Especially since I imagine users would never figure out how to fix it.

DmitryArefiev:
@Tantacrul Yes, still occurs in latest master and agree with P2.  Also putting back audio device settings to Preferences might fix the issue #11748

Tantacrul:
> @Tantacrul Yes, still occurs in latest master.  Also putting back audio device settings to Preferences might fix the issue #11748 
> 
> 
> 
> (regarding priority Vasilij was unable to reproduce this issue on his machine with a different version of Realtek Audio Manager)

When - why did audio device get taken out? Seems very unusual not to provide that option. Is there a separate issue for that?

vpereverzev:
@DmitryArefiev should be fine after my fixes to alpha-2. Please, check one more time on the latest master

DmitryArefiev:
> @DmitryArefiev should be fine after my fixes to alpha-2. Please, check one more time on the latest master

@vpereverzev Yes, it looks now! Checked on the **latest master**, the same ASUS laptop, Windows 10

PR #12242, opened 2022-07-02
Trying to fix Fluid synth tuning problems

cbjeukendrup:
In https://github.com/musescore/MuseScore/issues/11267#issuecomment-1127055159, I wrote that I had discovered that the slight detuned-ness of FluidSynth had to do with a sample rate mismatch, and also how that mismatch was caused. I tried to fix it, but that turns out to be complicated: 
- It is possible but strongly discouraged to change the sample rate of an existing FluidSynth (see `fluid_synth_set_sample_rate` in fluid_synth.c)
- The recommended solution is to delete the fluid synth and create a new one with the desired sample rate. But then we would need to backup and restore the whole state of the synth (loaded sounds, midi controllers, ...) which seems not realistic to me.

So then this hack came to my mind: we just keep doing what we are doing, namely compensating by changing the tuning of the synth, but just do it slightly better, instead of hard-coding a number that is only correct if the app sample rate is 48000Hz (#10804) and actually even not exactly correct but a few cents off (#11267). 

So, this hopefully:
Resolves: #11267 
Resolves: #10804 
Resolves: #12237


cbjeukendrup:
@DmitryArefiev It would be really great if you could try this out and let me know if it solves anything :)

DmitryArefiev:
> @DmitryArefiev It would be really great if you could try this out and let me know if it solves anything :)


@cbjeukendrup  Sure. I've checked briefly #12237 #11267 #10804 on Windows and it seems to be OK now, but @vpereverzev says the current PR should be code reviewed at first..

cbjeukendrup:
@DmitryArefiev Thanks for checking! Yes, it still needs review, but I thought in this case it would be good to have a brief test first, because if it doesn't solve anything then a code review is useless :)

cbjeukendrup:
Replaced by #12566 + #12574

PR #12244, opened 2022-07-02
Implemented new default Dash/Gap lengths for Lines

cbjeukendrup:
- For the line style property, we will now use a dedicated type instead of directly using (Q)PenStyle values.
- The "Dash" and "Custom dash" line types have been merged into a single "Dash" type, so now you can edit the Dash/Gap length properties whenever the Dash type is selected.
- The Dash/Gap length properties have been hooked up to the Styles system, so that we can set default values for them, which are also user-editable in the EditStyle dialog.

Reading old scores will work perfectly fine after this PR, but scores saved in builds from this PR might not be opened properly in older builds (the line style properties may not be read correctly).

Note: in this PR, the lines are still drawn using "square caps", meaning that the dashes are slightly longer than specified (just like before). This will be fixed in #11904. 

I would prefer this PR to be merged first, and then I will make #11904 even better.


cbjeukendrup:
Small change of plans: I removed the commits about which I was not yet super happy from #11904. I will redo those after this has been merged. So the preferred merge order is now:
1. #11904
2. #12244 (this PR)
3. Those commits containing the fix for the Square caps that I'll redo after these PRs have been merged

cbjeukendrup:
The VTest failures look good to me, even the change in `line-dashed-1`, in the second measure in the second staff, is expected: for that line (which is actually a "Let ring"), the gap size is customized but the dash size not, so the dash size will now follow the updated default, which is 4.0 for "Let rings" instead of the old generic default of 5.0.

oktophonie (approved):

RomanPudashkin (approved):

Issue #11313, opened 2022-04-23
[MU4 Task] Highlight voice colours during playback


shoogle:
During playback in MuseScore 3, noteheads under the playback cursor were coloured based on the voice they belong to: blue for voice 1, green for voice 2, etc. We should do this in MuseScore 4 as well.

### Playback in MuseScore 3

![](https://user-images.githubusercontent.com/11011881/163734607-75f37a59-0497-4184-9d2f-6fa9cfd5d740.png)

In MuseScore 4, noteheads are not currently coloured like this during playback, but [we would like them to be](https://github.com/musescore/MuseScore/issues/11221#issuecomment-1101185137).

DmitryArefiev:
Other playback cursor issue #11221

PR #12258, opened 2022-07-04
[MU4] Highlight voice colours during playback

Eism:
Resolves: #11313


Eism:
I close the PR because we decided to implement it to 4.x

Issue #12263, opened 2022-07-04
[MU4 Issue] All elements in Palettes > More section are missing


DmitryArefiev:
![Capture](https://user-images.githubusercontent.com/90187801/177172297-113192d8-ad60-415d-8bd7-440b3511288d.PNG)

https://user-images.githubusercontent.com/90187801/177172315-88ae152f-c160-4d1e-af70-7105e6c43006.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Was broken recently

DmitryArefiev:
@Tantacrul  Adding to Alpha2 since it was broken recently (please change if needed)

bkunda:
Yes I can verify that the PR I just re-tested from @HemantAntony (https://github.com/musescore/MuseScore/pull/12103) didn't have this problem.

cbjeukendrup:
I may have found the cause, and it's pretty stupid and my fault...

DmitryArefiev:
Fixed in #12268

Issue #12254, opened 2022-07-03
[MU4 Issue] unable to create custom key signature


sammik:
**Describe the bug**
Unable to create new custom key signature in Palette

**To Reproduce**
1. Palettes - Key Signatures
2. Click on 'More'
3. No option to create new Key Signature

**Expected behavior**
User should be abletocreate custom Key Signature

**Screenshots**
![No Custom KeySig](https://user-images.githubusercontent.com/1646034/177059871-a9a81453-2af9-4698-8466-4a0839f56694.png)

**Platform information**
 - OS: [checked on Linux]

**Additional context**
Master Pallette works.
Regression. It worked in MU4 some time ago.

DmitryArefiev:
Fixed in #12268

PR #12268, opened 2022-07-04
Fix "More" popups in Palettes

cbjeukendrup:
Resolves: #12254
Resolves: #12263

I already feared that this was the problem, and the final hint was this printed error:
```
19:33:28.115 | ERROR | main_thread | AppShell   | run: error: qrc:/qml/MuseScore/Palette/internal/PaletteTree.qml:633: Error: Unknown method return type: FilterPaletteTreeModel*
19:33:28.115 | WARN  | main_thread | Qt         | qrc:/qml/MuseScore/Palette/internal/PaletteTree.qml:633: Error: Unknown method return type: FilterPaletteTreeModel*
```

So anything relating to a Q_PROPERTY or Q_INVOKABLE method must be fully qualified (still nowadays). 

The good news is that the elements in the "More" popup will just be back after this PR (they have not been removed, just didn't show up).

### DIFF
--- a/src/palette/internal/paletteprovider.h
+++ b/src/palette/internal/paletteprovider.h
@@ -49,15 +49,8 @@ class PaletteElementEditor : public QObject, public async::Asyncable
     INJECT(palette, framework::IInteractive, interactive)
     INJECT(palette, IPaletteProvider, paletteProvider)
 
-    AbstractPaletteController* _controller = nullptr;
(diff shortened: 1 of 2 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #12254 #12263 in master (after merging) on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #12269, opened 2022-07-04
Harp pedalling diagrams and harp notation palette

miiizen:
This pull request adds harp pedalling diagrams to MuseScore for [GSoC 2022](https://musescore.org/en/user/3773138/blog/2022/06/01/gsoc-2022-harp-pedalling-diagrams). 
These diagrams will be editable through a UI which will be built in the coming weeks.  

Debussy-MS-1

**Known issues**
- [x] Save in score
- [x] Inspector
- [ ] Style properties
- [x] Style (diagram size, placement) is not final and will be amended.  
- [x] Text diagrams are not finished.  A start has been made.
- [x] Screenreader not yet working
- [x] Some issues with internal xml representation causing loss of data (eg. between palette and score) - input on this very welcome!
- [x] Harp diagram not using the Leland font when it is set in styledef.cpp and the style menu in the program
---


src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,250 @@
...
+                    break;
+                case PedalPosition::SHARP:
+                    diagram.append(strName + String("<sym>accidentalSharp</sym>, "));
+                    break;
+                }
+            }
+        }
+        // trailing comma
+        if (diagram.size() != 0) {
+            diagram.truncate(diagram.size() - 2);
```
RomanPudashkin, line 129: Magic number. Please explain what it means

src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    setXmlText(diagram);
+}
+
+const String HarpPedalDiagram::getStringName(HarpString str)
```
RomanPudashkin, line 136: This method should be static and should be declared directly in this file. Also, HarpString should be renamed to HarpStringType (and the argument to "type" accordingly). After these manipulations the method itself should be renamed to harpStringTypeToString

src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,250 @@
...
+    if (!xml.context()->canWrite(this)) {
+        return;
+    }
+    xml.startElement(this);
+    writeProperty(xml, Pid::HARP_IS_DIAGRAM);
+
+    // Write vector of harp strings.  Order is always D, C, B, E, F, G, A
+    xml.startElement("pedalState");
+    for (int idx = 0; idx < _pedalState.size(); idx++) {
+        LOGD() << "String " << idx << " : " << " Pedal state " << static_cast<int>(_pedalState[idx]);
```
RomanPudashkin, line 206: Please remove this log

src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+#include "harppedaldiagram.h"
+
+#include "segment.h"
+
+#include "rw/xml.h"
+
+#include "log.h"
+
+namespace mu::engraving {
```
RomanPudashkin, line 31: just use `using namespace mu::engraving;` instead of this

src/engraving/libmscore/harppedaldiagram.h
```diff
@@ -0,0 +1,91 @@
...
+};
+
+class HarpPedalDiagram final : public TextBase
+{
+    std::vector<PedalPosition> _pedalState;
+
+    bool _isDiagram = true;
+
+public:
+    // Constructors - too many?
```
RomanPudashkin, line 47: Please remove all TODOs, commented code blocks, and such comments

src/palette/internal/palettecreator.cpp
```diff
@@ -1671,3 +1674,25 @@ PalettePtr PaletteCreator::newFretboardDiagramPalette()
...
     return sp;
 }
+
+PalettePtr PaletteCreator::newHarpPalette()
+{
+    PalettePtr sp = std::make_shared<Palette>(Palette::Type::Harp);
+    sp->setName(QT_TRANSLATE_NOOP("palette", "Harp notation"));
+    sp->setGridSize(100, 28);   // TODO: Size of cell - double check this
+    sp->setDrawGrid(true);
+    sp->setVisible(true); //TODO: maybe?
```
RomanPudashkin, line 1684: Please solve all these TODOs or just remove them

src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,250 @@
...
+                String strName = getStringName(static_cast<HarpString>(idx));
+                switch (_pedalState[idx]) {
+                case PedalPosition::FLAT:
+                    diagram.append(strName + String("<sym>accidentalFlat</sym>, "));
+                    break;
+                case PedalPosition::NATURAL:
+                    diagram.append(strName + String("<sym>accidentalNatural</sym>, "));
+                    break;
+                case PedalPosition::SHARP:
+                    diagram.append(strName + String("<sym>accidentalSharp</sym>, "));
```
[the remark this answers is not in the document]
miiizen: The text diagrams in the design I'm following from @Tantacrul are only supposed to show the changes over previous settings, but yes the above is what I'd expect as a sort of initial state.
I'm still working on the text diagrams so can change this if need be!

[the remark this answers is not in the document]
cbjeukendrup: Yes, showing only changes is good, but even then Dorico puts D,C,B on a separate line. I'm wondering if we should do that too.
Schermafbeelding 2022-07-07 om 16 10 41

src/engraving/style/styledef.cpp
```diff
@@ -788,6 +788,24 @@ const std::array<StyleDef::StyleValue, size_t(Sid::STYLES)> StyleDef::styleValue
     { Sid::stringNumberFrameBgColor,      "stringNumberFrameBgColor",      PropertyValue::fromValue(draw::Color::transparent) },
     { Sid::stringNumberOffset,            "stringNumberOffset",            PointF(0.0, 0.0) },
 
+    { Sid::harpPedalDiagramFontFace,          "harpPedalDiagramFontFace",          "Leland" },
```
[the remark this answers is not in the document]
oktophonie: Also consider using these symbols for the accidentals (they will be a better size): https://www.w3.org/2019/03/smufl13/tables/standard-accidentals-for-chord-symbols.html

miiizen:
Both types of pedal diagrams are now finished.  The UI to edit them can be submitted as a separate PR or added to this one.  
I'll squash and rebase later in case there's much to change.  
Thanks!

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1209,6 +1209,18 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
             }
         }
     }
+
+    //-------------------------------------------------------------
+    // Harp pedal diagrams
```
cbjeukendrup, line 1214: By putting this at the bottom of the function, harp pedal diagrams will be stacked 'outside' of everything else. For example, if they are above the staff, they will appear even above tempo markings and rehearsal marks. @oktophonie What would be the correct stacking order?

oktophonie:
Like other indications of playing technique I'd expect them to stay close to the stave without too much getting in between. This is pretty easy if they are between the staves or below the bottom one. For placing them above, someone with access to a lot of solo harp material might better advise. I'd expect them to go outside of 8va markings but inside anything else(?)

src/palette/internal/palettecell.cpp
```diff
@@ -99,6 +99,7 @@ const char* PaletteCell::translationContext() const
     case ElementType::BREATH:
     case ElementType::FERMATA:
     case ElementType::SYMBOL:
+    case ElementType::HARP_DIAGRAM:
```
cbjeukendrup, line 102: For `ElementType::HARP_DIAGRAM`, this method should return "palette" (it should correspond with:
```c++
sp->appendElement(pedalDiagram, QT_TRANSLATE_NOOP("palette", "Harp pedal diagram"));
                                                   ^~~~~~~
```
_and_
```c++
sp->appendElement(pedalTextDiagram, QT_TRANSLATE_NOOP("palette", "Harp pedal text"));
                                                       ^~~~~~~
```

src/engraving/libmscore/factory.cpp
```diff
@@ -163,6 +164,7 @@ static const ElementName elementNames[] = {
     { ElementType::STAFFTYPE_CHANGE,     "StaffTypeChange",      QT_TRANSLATE_NOOP("elementName", "Staff type change") },
     { ElementType::HARMONY,              "Harmony",              QT_TRANSLATE_NOOP("elementName", "Chord symbol") },
     { ElementType::FRET_DIAGRAM,         "FretDiagram",          QT_TRANSLATE_NOOP("elementName", "Fretboard diagram") },
+    { ElementType::HARP_DIAGRAM,         "HarpPedalDiagram",          QT_TRANSLATE_NOOP("elementName", "Harp Pedal Diagram") },
```
cbjeukendrup, line 167: Minor thing: for ui texts, we use "Sentence case" instead of "Title Case". So this one would be "Harp pedal diagram".

src/palette/internal/palettecreator.cpp
```diff
@@ -1671,3 +1674,23 @@ PalettePtr PaletteCreator::newFretboardDiagramPalette()
...
+    sp->setDrawGrid(true);
+    sp->setVisible(true);
+
+    auto pedalDiagram = Factory::makeHarpPedalDiagram(gpaletteScore->dummy()->segment());
+    pedalDiagram->updateDiagramText();
+    sp->appendElement(pedalDiagram, QT_TRANSLATE_NOOP("palette", "Harp pedal diagram"));
+
+    auto pedalTextDiagram = Factory::makeHarpPedalDiagram(gpaletteScore->dummy()->segment());
+    pedalTextDiagram->setIsDiagram(false);
+    pedalTextDiagram->updateDiagramText();
```
cbjeukendrup, line 1692: You could also consider to call `updateDiagramText()` in `layout()`, so that you don't need to remember to call it and so that it will always be up-to-date when other diagrams earlier in the score are modified.

(If you decide to do so, then my comment in src/engraving/libmscore/chordrest.cpp doesn't apply anymore)

cbjeukendrup:
A rebase is needed. It would have my preference to merge this PR first, so that we can review the new changes in #12597 in isolation.

src/engraving/layout/layoutsystem.cpp
cbjeukendrup: @miiizen, as suggested by @oktophonie, let's place them for now "outside" ottavas but "inside" anything else. That would mean moving this block of code to just above https://github.com/musescore/MuseScore/blob/3ee3c7eaa517cfa0aca563fae476739f4ca2b894/src/engraving/layout/layoutsystem.cpp#L1090-L1094

src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,342 @@
...
+    case HarpStringType::D:
+        return String("D");
+    case HarpStringType::E:
+        return String("E");
+    case HarpStringType::F:
+        return String("F");
+    case HarpStringType::G:
+        return String("G");
+    }
+    return String("");
```
cbjeukendrup, line 54: Better just `String()` instead of `String("")`.

Also, those `String` literals should be prefixed with `u`. So: `String(u"A")`. 

Otherwise, you are converting from `const char*` to QString and then from QString to String. That's undesired because we want that the engraving module doesn't depend on Qt.

cbjeukendrup:
Some quick tips for rebasing, in case it's useful:
- About the #includes and forward declarations, I have recently done a PR to sort those alphabetically. This gives somewhat strange conflicts; to solve them, just choose the change from the master branch (or maybe Git will call it "Updated upstream") and then re-insert the #include that you inserted, please in the correct alphabetical order.
- `static const ElementName elementNames[]` has been moved to `typesconv.cpp`
- `ElementType` has been moved to `types/types.h` as opposed to `libmscore/types.h`
- `QT_TRANSLATE_NOOP` has been replaced with `TranslatableString` (in many cases);
- In PaletteCreator.cpp, some other new palettes have been added; you can simply preserve both changes

So in this case, the best strategy is often to choose the changes from the master branch and then reapply your additions. Of course, during rebasing, you can still see your changes (as they were before the rebase) here on GitHub, which you can use as a reference.

Finally: you could save the SHA of your last commit, 340ad21e138dc486a9294bd6e3e1fd838decdbd7, so that if it goes completely wrong you can do `git checkout 340ad21e138dc486a9294bd6e3e1fd838decdbd7` to go back to the state before rebase.

Jojo-Schmitz:
Huh?

miiizen:
Sorry for the confusion, please see #12597!

Jojo-Schmitz:
But how does that match with https://github.com/musescore/MuseScore/pull/12269#issuecomment-1225058301 ?

cbjeukendrup:
@Jojo-Schmitz We changed the plans a bit; it is not very convenient for @miiizen to continuously maintain two PRs. We decided to move on with #12597.

PR #12274, opened 2022-07-05
More improvements for Text Lines

cbjeukendrup:
This builds on #11904, so that PR should be reviewed and merged first. (However, this PR just contains the other one, so it is possible to merge just this one, but it may be more difficult to review both of them at once.)

This makes the drawing of text lines with different line styles more advanced:

- For solid lines:
    - we can just draw the line as a polyline, so that the corners will always look perfect (except for T-hooks, those need to be drawn separately)

- For dashed lines:
    - for each line piece individually, we want the dashes to be distributed over the length of that line piece, so we draw each piece separately.
    - for the dashes themselves, we want to use FlatCaps (otherwise, the dashes take more space than specified relative to the gaps), but for the 90° and 45° hooks, the line needs to be extended (as happens with SquareCaps) for the corners to look neat. For non-acute angles, we can make it look perfect by extending the lines slightly further based on the angle between them. For acute angles, such trick does not exist, so those will look as bad as in MU3. Fortunately, those are much less common (only occur when using crazy steep diagonal textlines).

- For dotted lines:
    - Again, we want to distribute the dots for each linepiece separately.
    - The dots are drawn by drawing very short dashes with round caps. Thus, since we distribute the dots, the outer dots will be circles centered around the start and end points of the line.
    - Because the end point of one linepiece will coincide with the start point of the next, we don't need to extend the lines.

So it should look good now it almost all cases. I also tried to minimize the amount of calculations (only perform them if the result is really necessary), to keep it efficient.


oktophonie (approved):

cbjeukendrup:
I added something new to the VTests: in addition to the "diff" image where the differences are highlighted in red, I added a GIF image that alternately shows the old and new image. In my opinion, this makes it easy to get an idea of how big the differences are. I also slightly increased the resolution, because it seems that at small resolutions, the paint engine sometimes starts behaving weirdly, making the tests quite useless.

RomanPudashkin (approved):

PR #12280, opened 2022-07-05
General improvements to slur shape and collision avoidance

mike-spa:
Adapting the shape and position of slurs to avoid collisions with the underlying items, and at the same time obtain a good looking arc, is quite a tricky matter. A vast improvement over version 3.6 was already done by @asattely. 

This PR extends the idea and includes a more general algorithm for the optimization of the slur shape and end points positions. Specifically:
- The Bezier pointes are adjusted at the same time as the end points and have a lot more flexibility in doing so.
- Also the width (and not only the height) of the arc is optimized.
- The arc is allowed to become slightly asymmetrical if the contour of the music is asymmetrical.
- Unnecessary asymmetries in end point height (making the slurs look tilted) are avoided.
- "Non-ugliness" rules are introduced to avoid bad looking arcs.
- Position of cross-staff slurs is corrected (though a lot more work will be needed there in future).

Here is a very artificial example constructed to show the difference in the absolute worst cases.
Before:
Screenshot from 2022-07-05 19-06-36
After:
Screenshot from 2022-07-05 19-15-08

@oktophonie @Tantacrul

### DIFF
--- a/src/engraving/layout/layoutpage.cpp
+++ b/src/engraving/layout/layoutpage.cpp
@@ -336,17 +336,15 @@ void LayoutPage::collectPage(const LayoutOptions& options, LayoutContext& ctx)
             continue;
         }
         auto spanners = ctx.score()->spannerMap().findOverlapping(stick, etick);
-        std::vector<Spanner*> spanner;
         for (auto interval : spanners) {
             Spanner* sp = interval.value;
-            if (sp->tick().ticks() < etick && sp->tick2().ticks() >= stick) {
-                if (sp->isSlur()) {
-                    ChordRest* scr = sp->startCR();
-                    ChordRest* ecr = sp->endCR();
-                    if (scr && ecr && scr->vStaffIdx() != ecr->vStaffIdx()) {
-                        toSlur(sp)->layout();
-                    }
-                }
+            if (!sp->isSlur()) {
+                continue;
+            }
+            ChordRest* scr = sp->startCR();
+            ChordRest* ecr = sp->endCR();
+            if (scr && ecr && scr->vStaffIdx() != ecr->vStaffIdx()) {
+                toSlur(sp)->layout();
             }
         }
     }
--- a/src/engraving/layout/layoutpage.cpp
+++ b/src/engraving/layout/layoutpage.cpp
@@ -42,6 +42,7 @@
 #include "libmscore/chord.h"
 #include "libmscore/tremolo.h"
 #include "libmscore/barline.h"
+#include "libmscore/slur.h"
 
 #include "layoutsystem.h"
 #include "layoutbeams.h"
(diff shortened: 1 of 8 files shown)


oktophonie (approved):

src/engraving/libmscore/chord.cpp
```diff
@@ -3973,30 +3973,56 @@ void Chord::layoutArticulations2()
...
     }
 }
 
+bool Chord::isSlurStartEnd() const
+{
+    auto spanners = score()->spannerMap().findOverlapping(tick().ticks(), tick().ticks());
+    if (spanners.empty()) {
+        return false;
+    }
+    for (auto sp : spanners) {
```
vpereverzev, line 4011: you might have a lot of spanners here, even from other staffs. Makes sense to compare staff indexes of a spanner and a chord here to avoid unnecessary calculations

src/engraving/libmscore/chord.cpp
```diff
@@ -3973,30 +3973,56 @@ void Chord::layoutArticulations2()
...
             }
-            r.translate(s->pos() + m->pos());
-            m->system()->staff(vStaffIdx())->skyline().add(r);
         }
     }
 }
 
+bool Chord::isSlurStartEnd() const
+{
+    auto spanners = score()->spannerMap().findOverlapping(tick().ticks(), tick().ticks());
```
vpereverzev, line 4007: avoid double calculations of tick().ticks() (add some variable above)

src/engraving/libmscore/slur.cpp
```diff
@@ -349,139 +366,327 @@ void SlurSegment::adjustEndpoints()
...
+        const unsigned npoints = 20;
+        std::vector<RectF> slurRects;
+        slurRects.reserve(npoints);
+        // Define separate collision areas (left-mid-center)
+        SlurCollision collision;
+
+        // CHECK FOR COLLISIONS
+        unsigned iter = 0;
+        do {
+            collision.left = collision.mid = collision.right = false;
```
vpereverzev, line 397: do you really need to set these values on every iteration?

src/engraving/libmscore/slur.cpp
```diff
@@ -349,139 +366,327 @@ void SlurSegment::adjustEndpoints()
...
+                        if (rightSection) {
+                            collision.right = true;
+                        }
+                    }
+                }
+                if (seg == endSeg) {
+                    break;
+                }
+            }
+            // In the even iterations, adjust the shape
```
vpereverzev, line 459: not quite sure about this one. What was the motivation?

src/engraving/libmscore/slur.cpp
mike-spa: Yes, because at every iteration I modify the shape of the slur, so I must check for collisions again. Also, if I don't reset them I never exit the loop.

src/engraving/libmscore/slur.cpp
mike-spa: Basically, if I see that the slur collides with something, I adjust it in small steps: one step, I adjust the shape; next step, I adjust the end points; next I adjust the shape again; next the end points again; until there are no collisions anymore. The simplest way I could think for doing this is to do the shape in the even iterations, and the end points in the odd iterations.

src/engraving/layout/layoutpage.cpp
```diff
@@ -322,6 +323,34 @@ void LayoutPage::collectPage(const LayoutOptions& options, LayoutContext& ctx)
...
+        }
+        etick = system->endTick().ticks();
+        if (stick == 0 && etick == 0) {
+            continue;
+        }
+        auto spanners = ctx.score()->spannerMap().findOverlapping(stick, etick);
+        std::vector<Spanner*> spanner;
+        for (auto interval : spanners) {
+            Spanner* sp = interval.value;
+            if (sp->tick().ticks() < etick && sp->tick2().ticks() >= stick) {
```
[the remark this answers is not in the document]
mike-spa: Done! 👍

src/engraving/layout/layoutpage.cpp
```diff
@@ -322,6 +323,34 @@ void LayoutPage::collectPage(const LayoutOptions& options, LayoutContext& ctx)
...
+        long int etick = 0;
+        if (system->firstMeasure()) {
+            stick = system->firstMeasure()->tick().ticks();
+        }
+        etick = system->endTick().ticks();
+        if (stick == 0 && etick == 0) {
+            continue;
+        }
+        auto spanners = ctx.score()->spannerMap().findOverlapping(stick, etick);
+        std::vector<Spanner*> spanner;
```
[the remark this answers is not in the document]
mike-spa: Done! 👍

src/engraving/layout/layoutpage.cpp
```diff
@@ -322,6 +323,34 @@ void LayoutPage::collectPage(const LayoutOptions& options, LayoutContext& ctx)
...
+        for (auto interval : spanners) {
+            Spanner* sp = interval.value;
```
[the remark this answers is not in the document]
mike-spa: Done! 👍

src/engraving/libmscore/chord.cpp
mike-spa: Done! 👍

src/engraving/libmscore/chord.cpp
mike-spa: Done! 👍

src/engraving/libmscore/slur.cpp
vpereverzev: then add a "reset" method into the structure

mike-spa: ha, good point! done 👍

PR #12297, opened 2022-07-07
Fixed no output when debugging

HemantAntony:

There was not output when you debug. This commit fixes this

### DIFF
--- a/src/framework/global/thirdparty/haw_logger/logger/logdefdest.cpp
+++ b/src/framework/global/thirdparty/haw_logger/logger/logdefdest.cpp
@@ -3,6 +3,10 @@
 #include <iostream>
 #include <cassert>
 
+#ifdef _WIN32
+#include <Windows.h>
+#endif
+
 using namespace haw::logger;
 
 MemLogDest::MemLogDest(const LogLayout& l)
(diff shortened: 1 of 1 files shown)


HemantAntony:
I'm not really sure if this is best solution or if I need to creating this PR upstream. I also don't know if this will affect logging in Mac and Linux

src/framework/global/thirdparty/haw_logger/logger/logdefdest.cpp
```diff
@@ -71,5 +73,8 @@ std::string ConsoleLogDest::name() const
 
 void ConsoleLogDest::write(const LogMsg& logMsg)
 {
-    std::clog << m_layout.output(logMsg) << std::endl;
+    std::string stemp = m_layout.output(logMsg);
+    std::wstring temp = std::wstring(stemp.begin(), stemp.end());
+    OutputDebugString(temp.c_str());
+    OutputDebugString(L"\n");
```
igorkorsukov, line 79: This solution can only work for Windows, it won't work for Linux and Macos, it won't even compile

cbjeukendrup: @igorkorsukov Is there a specific reason that we use std::clog? It seems to send all output to stderr instead of stdout, and I'm not sure if that's desirable.

igorkorsukov: There is no special reason, in fact it may be better to replace it with `std::cout`

HemantAntony: Okay, I will do that too

HemantAntony: @igorkorsukov No, it doesn't work. I had to use the `OutputDebugString()` function. Also, this also  gets displayed along with the regular logs when you are debugging. These were getting printed before this commit:
image

Issue #12190, opened 2022-06-28
[MU4 Improvement] Ability to add glissando to all chord notes when whole chord is selected


boryskuzmenko:
**Describe the bug**
Ability to add glissando to all chord notes when whole chord is selected

**Screenshots**

https://user-images.githubusercontent.com/101250347/176152297-3a28bc51-6f06-45e9-ae3a-78f6593b727b.mov

Tantacrul:
👍

boryskuzmenko:
🟢 **FIXED** in https://github.com/musescore/MuseScore/pull/12306
Tested on macOS 12, Win 11, Ubuntu 20

PR #12306, opened 2022-07-07
Allow adding glissando to all notes when range selected

mike-spa:
Resolves part of #12190 

This just seems to require one line of code, so I thought it was worth doing it. The second point of the issue, i.e. inserting a glissando between two specifically selected notes, I think requires more work (because there isn't an interaction model for it as far as I can understand) so it should be postponed. 

Note: the choice of end notes for each glissando line is fixed by my other PR #12303 which should be merged together (or before) this.

Before:

https://user-images.githubusercontent.com/93707756/177803061-97fb5a17-2dfe-44c9-ab4f-5cddcdad443a.mp4

After:

https://user-images.githubusercontent.com/93707756/177803109-11751eaa-1f69-41f2-936c-769be2dbafbb.mp4


boryskuzmenko:
🟢 **FIXED**

Please, merge to master

The issue about adding a glissando between two selected notes is reported separately https://github.com/musescore/MuseScore/issues/12468

cbjeukendrup (approved):

Issue #12287, opened 2022-07-06
[MU4 Issue] Two-note tremolos between whole notes are misplaced


oktophonie:
![image](https://user-images.githubusercontent.com/20604032/177529228-aef0b824-fad1-4653-b067-b8f1c9148cea.png)

**Describe the bug**
When the tremolo is upstemmed, the vertical position of the tremolos is miscalculated (see the first two bars above).
The vertical position of the tremolo should match that of the equivalent situation where stems are present (i.e. as between quarter/half notes).

Horizontally offset the ends of the tremolos inward from the inside of each notehead by whatever the standard stem:tremolo gap is.

PR #12307, opened 2022-07-07
Fix 12287: Fixing tremolo positions for stemless notes

### LINKED ISSUE #11899: Crashes on score with stemless notation
Hi, not sure this is the correct place to report this.

 I am trying to open a file created in MS3.6 with MS4 but it crashes before opening. The file contains stemless notation, empty text frames (a box surrounding the notes created through staff text with empty content and a rectangular frame) and lines stretching across multiple lines. I also have a lot of hidden content (mostly rests from empty bars).
After some testing (I deleted elements one by one to see what stops loading the file) the crash seems to be caused by stemless notation. However, if I set my my score to display stems and add some other stemless notation to a different staff in MS3 the file opens correctly

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12287
(and also incidentally resolves the stemless notation crash mentioned in https://github.com/musescore/MuseScore/issues/11899)

Tremolo whole notes are now calculated as if the whole notes had stems, with some adjustment so that the tremolo will never be over the top of the notehead.
image

This also fixes #11899, which contained a regression due to the beams refactor #11788.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1585,9 +1585,7 @@ int Chord::calc4BeamsException(int stemLength) const
 
 double Chord::calcDefaultStemLength()
 {
-    if (!_stem) {
-        return 0.0;
-    }
+    // returns default length even if the chord doesn't have a stem
 
     double _spatium = spatium();
 
@@ -1768,6 +1766,10 @@ void Chord::calcRelativeMag()
 //! May be called again when the chord is added to or removed from a beam.
 void Chord::layoutStem()
 {
+    // we should calculate default stem length for this chord even if it doesn't have a stem
+    // because this length is used for tremolos or other things that attach to where the stem WOULD be
+    _defaultStemLength = calcDefaultStemLength();
+
     if (!shouldHaveStem()) {
         removeStem();
         return;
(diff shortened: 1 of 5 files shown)


src/engraving/libmscore/chord.cpp
```diff
@@ -1585,9 +1585,9 @@ int Chord::calc4BeamsException(int stemLength) const
 
 double Chord::calcDefaultStemLength()
 {
-    if (!_stem) {
+    /*if (!_stem) {
```
RomanPudashkin, line 1588: You should either remove this code or uncomment it

src/engraving/libmscore/tremolo.h
```diff
@@ -42,6 +42,9 @@ enum class TremoloStyle : signed char {
 
 class Tremolo final : public EngravingItem
 {
+    // TODO: This should be a style setting, to replace tremoloStrokeLengthMultiplier
+    const double stemGapSp = 0.65;
```
RomanPudashkin, line 46: Please declare this in cpp. Also, it should be `static constexpr`

asattely:
Done and done!

RomanPudashkin (approved):

src/engraving/libmscore/tremolo.cpp
```diff
@@ -360,9 +372,6 @@ void Tremolo::layoutTwoNotesTremolo(double x, double y, double h, double spatium
     //---------------------------------------------------
     //   Step 1: Calculate the position of the tremolo (x, y)
     //---------------------------------------------------
-
-    y += (h - bbox().height()) * .5;
```
Jojo-Schmitz, line 364: This removal causes a MSVC compiler warning C4100 reg. "unreferenced formal parameter"  `double h`.

PR #12318, opened 2022-07-08
Fix #12316: Limit Gap values to 2 decimals in Inspector

### LINKED ISSUE #12316: Number values for frames allow too many digits after the decimal point
**Describe the bug**
Text frames can sometimes have values that are so long, they exceed the width of the text box. We need to limit these to two points after the decimal.

image

Here's another example of the default for a Symphony Orchestra template
image

**To Reproduce**
1. Open the Symphony Orchestra template
2. Click on the title frame
3. Observe the 'Gap above' and 'Gap below' value

HemantAntony:
Resolves: #12316 

### DIFF
--- a/src/framework/global/val.cpp
+++ b/src/framework/global/val.cpp
@@ -101,9 +101,9 @@ bool Val::toBool() const
         return str == VAL_TRUE;
     }
     default:
-        return isNull() ? false : true;
+        break;
     }
-    return false;
+    return isNull() ? false : true;
 }
 
 int Val::toInt() const
--- a/src/appshell/qml/Main.wasm.qml
+++ b/src/appshell/qml/Main.wasm.qml
@@ -33,7 +33,7 @@ ApplicationWindow {
     height: 480
 
     visible: true
-    title: qsTr("Muse Score")
+    title: qsTr("MuseScore")
 
     header: ToolBar {
         contentHeight: 40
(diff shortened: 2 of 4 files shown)


Tantacrul:
Just tested. Seems to be working as expected! @RomanPudashkin / @Eism, if you are happy, can you merge?

Thanks @HemantAntony !

src/framework/global/val.cpp
```diff
@@ -100,10 +100,8 @@ bool Val::toBool() const
         std::string str = std::get<std::string>(m_val);
         return str == VAL_TRUE;
     }
-    default:
-        return isNull() ? false : true;
     }
-    return false;
+    return isNull() ? false : true;
```
RomanPudashkin, line 104: This will cause a warning because after this change not all possible types will be covered in the switch statement

HemantAntony: I actually didn't get that warning. I was trying to fix an "unreachable" warning. I'll just delete the `return false;`

Tantacrul:
Still seems to be working. @RomanPudashkin - if you are happy?

src/framework/global/val.cpp
cbjeukendrup: It differs per compiler. MSVC tends to complain about unreachable code if you put the return statement in the `default` case and also put something at the bottom of the function, while Clang is afraid that the function doesn't always return a value if you don't. The general solution would be like this:
```c++
    default:
        break;
    }
    return isNull() ? false : true;
```

RomanPudashkin:
> Still seems to be working. @RomanPudashkin - if you are happy?

Waiting for my [comment](https://github.com/musescore/MuseScore/pull/12318#discussion_r916938854) to be resolved :)

HemantAntony:
Sorry! Switching between branches is a bit of a pain. Hence I was putting it off :)

RomanPudashkin (approved):

PR #12329, opened 2022-07-09
Improvements/fixes for MIDI output on macOS

cbjeukendrup:
- fix some integer type mismatch warnings
- calculate the proper size for MIDIPacketLists on macOS 10.15 and earlier
- listen for device name changes (if the name of a device changes, update the UI in Preferences)
- on macOS 11.0 and later:
    - correct the given size for messages (instead of byte count, e.g. `4 * sizeof(UInt32)`, we should pass word count, e.g. `4`)
    - if the connected receiver does not support the MIDI 2.0 protocol, use the MIDI 1.0 protocol
    - add an option to disable the use of MIDI 2.0 protocol even when the receiver reports that it supports it (see commit message for why this is useful)
        Schermafbeelding 2022-07-09 om 18 41 16

### DIFF
--- a/src/framework/midi/internal/platform/win/winmidioutport.cpp
+++ b/src/framework/midi/internal/platform/win/winmidioutport.cpp
@@ -149,6 +149,11 @@ MidiDeviceID WinMidiOutPort::deviceID() const
     return m_deviceID;
 }
 
+bool WinMidiOutPort::supportsMIDI20Output() const
+{
+    return false;
+}
+
 mu::Ret WinMidiOutPort::sendEvent(const Event& e)
 {
     if (!isConnected()) {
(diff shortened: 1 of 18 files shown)


src/framework/midi/internal/platform/win/winmidioutport.cpp
```diff
@@ -149,6 +149,11 @@ MidiDeviceID WinMidiOutPort::deviceID() const
     return m_deviceID;
 }
 
+bool WinMidiOutPort::supportsMIDI20Output() const
+{
+    return false;
```
Eism, line 154: Doesn't Windows support midi 2.0?

cbjeukendrup: At least our implementation of the Windows MIDI port converts everything to MIDI1.0 before sending it. So we don't really support sending MIDI2.0 events. If Windows itself supports MIDI2.0 (I don't know this), we can update our implementation to make use of it, and then we can change this to `true`.

Eism (approved):

PR #12308, opened 2022-07-07
fixed #10453: Voice pitch changes when moving note up and down

Eism:
Resolves: #10453

PR #12340, opened 2022-07-11
Accessibility: Improve speech output for notes and common elements

shoogle:
Screen readers will now announce:

- Accidentals in note names (e.g. "A♯", "B♭", etc.).

- [Fix [MU#312625](https://musescore.org/en/node/312625)] Notehead group type (e.g. "cross", "triangle", etc.).

- [Fix [MU#312628](https://musescore.org/en/node/312628)] When notes or rests use cross-staff notation.

- [Fix [MU#312627](https://musescore.org/en/node/312627)] When elements are invisible.

- [Fix [MU#307442](https://musescore.org/en/node/307442)] Tuplet length for custom tuplets (more than 9 notes).

Replaces PR #8877.

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -25,13 +25,26 @@
 #include "../libmscore/score.h"
 #include "../libmscore/measure.h"
 
+#include "translation.h"
 #include "log.h"
 
+using namespace mu;
 using namespace mu::engraving;
 using namespace mu::accessibility;
 using namespace mu::engraving;
 
 bool AccessibleItem::enabled = true;
+
+static QString readable(QString s)
+{
+    // Remove undesired pauses in screen reader speech output
+    s.remove(u':');
+
+    // Handle desired pauses
+    s.replace(u';', u','); // NVDA doesn't pause for semicolon
+
+    return s;
+}
 
 AccessibleItem::AccessibleItem(EngravingItem* e, Role role)
     : m_element(e), m_role(role)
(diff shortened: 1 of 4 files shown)


shoogle:
Tested working with NVDA, Narrator, VoiceOver, Orca.

For testing purposes, use Ctrl+Shift+Up or Ctrl+Shift+Down to move some notes across staves, then make the note invisible or change the notehead shape via the Properties panel.

image


Create a custom tuplet with 10 or more notes via:
image image 
image

And check speech output for the above.

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -25,14 +25,34 @@
...
 
+static QString readable(QString s)
+{
+    // Remove undesired pauses in screen reader speech output
+    s.remove(u':');
+
+    // Handle desired pauses
+    s.replace(u';', u','); // NVDA doesn't pause for semicolon
+
+    // Replace common Unicode characters that screen readers can't pronounce
```
Eism, line 46: Hm, there is my fix of this bug https://github.com/musescore/MuseScore/pull/12308/commits/5e0a8460564e6ce2ecc730faa5e5f93d39453cc4

shoogle: Ok, it seems that this really does just occur with Narrator so we can probably consider the non-spoken accidentals Microsoft's bug rather than our own. Your version of the fix will still be useful though. Firstly for Narrator, but also to VoiceOver because it says "music flat sign" instead of just "flat", etc, which gets annoying quite quickly.

Eism (approved):

PR #12347, opened 2022-07-11
[MU4] Fix #11474: Shortcuts section in Preferences is still very broken

### LINKED ISSUE #11474: Shortcuts section in Preferences is still very broken
**Describe the bug**
The shortcuts panel has the following issues.

1. When holding the shift modifier and pressing a number (in this example '0'), the keys that appear are **⇧ )** rather than **⇧ 0**
As a result, the newly defined shortcut doesn't work

image

2. If you search for 'Triplet' and then clear the shortcut for it ( which is **⌘ 3** or **Ctrl 3**). Now try to define the same shortcut again and an error will tell you it is already in use. (See image below). This means a user cannot clear a shortcut from one action in order to assign it to another.
image

3. In one case (possibly more, I've not checked) the same shortcut is defined by default to multiple actions:
image
image

We need to remove all default duplicates 

4. We no longer want to reveal shortcut sequences. They cause incredible confusion to users who just want to specify a simple shortcut and cause constant crashes in MS3. We **may** decide to support them again in later versions but only if there is enough demand for them.

5. We should remove the ability to have to 'Add' another shortcut definition. There should only be one shortcut (or shortcut sequence) per action. For this reason, we need to remove the 'Add' button and only have the **Replace** button, which should be renamed to **Apply**. This will also massively help the UX.

6. By removing the button (and functionality) for the 'add' button, this will help with this last visual bug, which is that the cancel button is falling out of frame 
image

Please send to @bkunda or myself for testing. If anything is unclear in this description, let me know and I'll happily jump on a call or chat to clarify

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11474

### DIFF
--- a/src/framework/shortcuts/view/editshortcutmodel.cpp
+++ b/src/framework/shortcuts/view/editshortcutmodel.cpp
@@ -160,16 +160,6 @@ void EditShortcutModel::replaceOriginSequence()
     emit applyNewSequenceRequested(m_originSequence);
 }
 
-void EditShortcutModel::addToOriginSequence()
-{
-    if (!m_originSequence.isEmpty()) {
-        m_originSequence += "; ";
-    }
-    m_originSequence += inputtedSequence();
-
-    emit applyNewSequenceRequested(m_originSequence);
-}
-
 QString EditShortcutModel::inputtedSequence() const
 {
     return m_inputtedSequence.toString();
--- a/src/framework/shortcuts/view/editshortcutmodel.cpp
+++ b/src/framework/shortcuts/view/editshortcutmodel.cpp
@@ -29,6 +29,7 @@
 #include "log.h"
 
 using namespace mu::shortcuts;
+using namespace mu::framework;
 
 static const QString ANY_CONTEXT("any");
 
@@ -39,13 +40,22 @@ EditShortcutModel::EditShortcutModel(QObject* parent)
 
 void EditShortcutModel::load(const QVariant& originShortcut, const QVariantList& allShortcuts)
 {
-    clear();
+    TRACEFUNC;
+
+    clearNewSequence();
+
+    m_allShortcuts = allShortcuts;
+    m_potentialConflictShortcuts.clear();
 
     QVariantMap originShortcutMap = originShortcut.toMap();
     QString originCtx = originShortcutMap.value("context", ANY_CONTEXT).toString();
     bool isOriginCtxAny = originCtx == ANY_CONTEXT;
 
     for (const QVariant& shortcut : allShortcuts) {
+        if (shortcut == originShortcut) {
+            continue;
+        }
+
         if (isOriginCtxAny) {
             m_potentialConflictShortcuts << shortcut;
             continue;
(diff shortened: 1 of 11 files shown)


src/framework/shortcuts/view/editshortcutmodel.cpp
```diff
@@ -39,42 +38,45 @@ EditShortcutModel::EditShortcutModel(QObject* parent)
...
     for (const QVariant& shortcut : allShortcuts) {
-        if (isOriginCtxAny) {
-            m_potentialConflictShortcuts << shortcut;
+        if (shortcut == originShortcut) {
             continue;
         }
 
         QVariantMap map = shortcut.toMap();
-        QString ctx = map.value("context", ANY_CONTEXT).toString();
+        std::string ctx = map["context"].toString().toStdString();
```
igorkorsukov, line 57: Avoid using the `[]` operator for reading, because in this case, the container detach occurs

src/framework/shortcuts/view/editshortcutmodel.cpp
```diff
@@ -139,38 +135,59 @@ QString EditShortcutModel::originSequenceInNativeFormat() const
...
-    return m_inputtedSequence.toString(QKeySequence::NativeText);
+    return m_newSequence.toString(QKeySequence::NativeText);
 }
 
-QString EditShortcutModel::errorMessage() const
+QString EditShortcutModel::conflictWarning() const
 {
-    return m_errorMessage;
-}
+    QString title = m_conflictShortcut["title"].toString();
```
igorkorsukov, line 145: avoid using the [] operator to read, there may  detach

Eism (approved):

Tantacrul:
OK, I've tested this in detail and am now happy with it. For anyone interested, we have made the following changes in this PR, so that Shortcuts are much more usable and easy to assign / reassign.

**Three main points:**
- We have removed the ability to set shortcut sequences. These are nice in concept but cause endless UX problems and are understood by almost no one. In addition, they often crash MS3 or just don't work. This is not a legacy we want to bring into MS4. In future, if there is demand, we will implement a separate and more stable implementation of shortcut sequences.
- The same shortcuts can still be used for multiple different actions but the ability to do this should only be available to developers, who can consult the context priority (a more robust system than existed in MS3). For users, it is enormously confusing and illogical to allow the same shortcut to be applied to different actions, so we disallow it.
- Two different shortcuts can no longer be applied to the same action. This improves the comprehensibility of the Shortcuts panel by quite a lot.
- We have made improvements to the UX for reassigning shortcuts. For example, if a user assigns the shortcut 'M' to the mixer, they will be told that this same shortcut is being used by **multi-measure rests** and will be given the option to re-assign it immediately. They no longer have to seek out the original action and 'clear' it first.

With all the above achieved, shortcuts finally feel usable and comprehensible. We have also applied a much clearer and logical naming convention to shortcuts (handled in a separate PR). This will do the following:

1. Combine similar actions together in the list, which massively reduces the scattered nature of the existing list
2. Makes each action easier to understand

Once 4.0 is released, we will also create categories for shortcuts, so they are easier to find.

PR #12362, opened 2022-07-12
Avoid translation duplication (note)

igorkorsukov:
Removed 30 entries

this `.ts` file is uploaded to Transifex, we can see how this looks there

### DIFF
--- a/share/locale/musescore_en.ts
+++ b/share/locale/musescore_en.ts
@@ -147,1638 +147,1495 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../../src/notation/internal/notationuiactions.cpp" line="1225"/>
+        <location filename="../../src/notation/internal/notationuiactions.cpp" line="1250"/>
         <source>Print</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../../src/notation/internal/notationuiactions.cpp" line="413"/>
+        <location filename="../../src/notation/internal/notationuiactions.cpp" line="438"/>
         <source>Undo</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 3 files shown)


cbjeukendrup:
@igorkorsukov I just realized that there is a flaw with this approach. By writing 
```c++
static const QString Enter_note_X(QT_TRANSLATE_NOOP("action", "Enter note %1"));
```
the translatable string becomes `"Enter note %1"`. 

But in the action, we write:
```
Enter_note_X.arg(noteC)
```
So the action will store the string `"Enter note C"`.

When we later try to get the translated action name, we try to search the translation for `"Enter note C"`. But it doesn't exist. The translation system only has a translation for `"Enter note %1"`, not for `"Enter note C"`.

So, in summary, a general rule: *we should never use .arg in places where we store the untranslated version of a translatable string*. In other words: *only apply .arg on strings that are _already_ translated*.

That means that we can't avoid these duplicates at all, when it comes to the list of actions.

Another place where a lot of duplicates occur, is in the engraving module, in typesconv.cpp. With the current implementation of this file, we have the same problem: we are _storing_ the untranslated version of a translatable string, so we can't use .arg. However, in this file, we could relatively easily change the implementation to something like this:
```c++
switch (...) {
case ...:
    return mtrc("engraving", "... %1").arg(mtrc("engraving", "..."))
...
}
```
(because with that kind of implementation, the string is translated before .arg is used). However, this will require a lot of extra lines of code, which is quite tedious... I'd think the performance shouldn't be a difference in performance though.

Also, I wonder how bad those duplicates are. Translators can also copy+paste :)
And it's not worse than in MuseScore 3.

So, considering everything, what do you think, should I change the implementation of the relevant methods in typesconv.cpp, or is it not worth the effort and the extra code?

PR #12363, opened 2022-07-12
[MU4] [Tab] Force correct stem direction on non-full tab staves

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/11606

When beams and stems had custom directions and then were switched to tab-common, those stem directions were not being overridden by the staff type. Now they are.

Note: these custom stem directions are still preserved, in case you change back to tab-full or regular notation.


boryskuzmenko:
🟢 **FIXED**

Please, merge to master

RomanPudashkin:
Please add some tests to prevent this fix from regressions

asattely:
> Please add some tests to prevent this fix from regressions

Done, and rebased!

RomanPudashkin (approved):

asattely:
Rebased again!

PR #12364, opened 2022-07-12
[MU4] Fix some translatable strings

Jojo-Schmitz:
mostly to fix strings to use to US English, 6 to use singular/plural forms and 2 others to avoid leading/trailing newlines.

### DIFF
--- a/src/engraving/libmscore/measurerepeat.cpp
+++ b/src/engraving/libmscore/measurerepeat.cpp
@@ -297,7 +297,7 @@ Fraction MeasureRepeat::ticks() const
 
 String MeasureRepeat::accessibleInfo() const
 {
-    return mtrc("engraving", "%1; Duration: %2 measure(s)").arg(EngravingItem::accessibleInfo()).arg(numMeasures());
+    return mtrc("engraving", "%1; Duration: %n measure(s)", nullptr, numMeasures()).arg(EngravingItem::accessibleInfo());
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/noteentry.cpp
+++ b/src/engraving/libmscore/noteentry.cpp
@@ -387,8 +387,9 @@ void Score::putNote(const Position& p, bool replace)
     staff_idx_t staffIdx = track2staff(_is.track());
     if (m->isMeasureRepeatGroup(staffIdx)) {
         auto b = MessageBox::warning(trc("engraving", "Note input will remove measure repeat"),
-                                     trc("engraving", "There is a measure repeat here.")
-                                     + trc("engraving", "\nContinue with adding note and delete measure repeat?"));
+                                     trc("engraving", "This measure contains a measure repeat."
+                                                      " If you enter notes here, it will be deleted."
+                                                      " Do you want to continue?"));
         if (b == MessageBox::Cancel) {
             return;
         }
--- a/src/engraving/types/typesconv.cpp
+++ b/src/engraving/types/typesconv.cpp
@@ -242,8 +242,8 @@ static const std::vector<Item<ElementType> > ELEMENT_TYPES = {
     { ElementType::COMPOUND,             "Compound",             QT_TRANSLATE_NOOP("engraving", "Compound") },
     { ElementType::CHORD,                "Chord",                QT_TRANSLATE_NOOP("engraving", "Chord") },
     { ElementType::SLUR,                 "Slur",                 QT_TRANSLATE_NOOP("engraving", "Slur") },
-    { ElementType::ELEMENT,              "EngravingItem",        QT_TRANSLATE_NOOP("engraving", "EngravingItem") },
-    { ElementType::ELEMENT_LIST,         "ElementList",          QT_TRANSLATE_NOOP("engraving", "EngravingItem list") },
+    { ElementType::ELEMENT,              "EngravingItem",        QT_TRANSLATE_NOOP("engraving", "Element") },
+    { ElementType::ELEMENT_LIST,         "ElementList",          QT_TRANSLATE_NOOP("engraving", "Element list") },
     { ElementType::STAFF_LIST,           "StaffList",            QT_TRANSLATE_NOOP("engraving", "Staff list") },
     { ElementType::MEASURE_LIST,         "MeasureList",          QT_TRANSLATE_NOOP("engraving", "Measure list") },
     { ElementType::HBOX,                 "HBox",                 QT_TRANSLATE_NOOP("engraving", "Horizontal frame") },
@@ -381,7 +381,7 @@ static const std::vector<Item<NoteHeadScheme> > NOTEHEAD_SCHEMES = {
     { NoteHeadScheme::HEAD_AUTO,                "auto",                QT_TRANSLATE_NOOP("engraving", "Auto") },
     { NoteHeadScheme::HEAD_NORMAL,              "normal",              QT_TRANSLATE_NOOP("engraving", "Normal") },
     { NoteHeadScheme::HEAD_PITCHNAME,           "name-pitch",          QT_TRANSLATE_NOOP("engraving", "Pitch names") },
-    { NoteHeadScheme::HEAD_PITCHNAME_GERMAN,    "name-pitch-german",   QT_TRANSLATE_NOOP("engraving", "German Pitch names") },
+    { NoteHeadScheme::HEAD_PITCHNAME_GERMAN,    "name-pitch-german",   QT_TRANSLATE_NOOP("engraving", "German pitch names") },
     { NoteHeadScheme::HEAD_SOLFEGE,             "solfege-movable",     QT_TRANSLATE_NOOP("engraving", "Solf\u00e8ge movable Do") },  // &egrave;
     { NoteHeadScheme::HEAD_SOLFEGE_FIXED,       "solfege-fixed",       QT_TRANSLATE_NOOP("engraving", "Solf\u00e8ge fixed Do") },    // &egrave;
     { NoteHeadScheme::HEAD_SHAPE_NOTE_4,        "shape-4",             QT_TRANSLATE_NOOP("engraving", "4-shape (Walker)") },
--- a/src/framework/global/dataformatter.cpp
+++ b/src/framework/global/dataformatter.cpp
@@ -44,48 +44,29 @@ String DataFormatter::formatTimeSince(const Date& dateTime)
         return mtrc("global", "Today");
     }
 
-    if (days == 1) {
-        return mtrc("global", "Yesterday");
-    }
-
-    if (days < 7) {
-        return mtrc("global", "%1 days ago").arg(days);
+    if (days >= 1 && days < 7) {
+        //: Omit the "%n" for the singular translation and translate to (the equivalent of) "Yesterday"
+        return mtrc("global", "%n day(s) ago", nullptr, days);
     }
 
     int weeks = days / 7;
 
-    if (weeks == 1) {
-        return mtrc("global", "Last week");
-    }
-
-    if (weeks == 2) {
-        return mtrc("global", "Two weeks ago");
-    }
-
-    if (weeks == 3) {
-        return mtrc("global", "Three weeks ago");
-    }
-
-    if (weeks == 4) {
-        return mtrc("global", "Four weeks ago");
+    if (weeks >= 1 && weeks <= 4) {
+        //: Omit the "%n" for the singular translation and translate to (the equivalent of) "Last week"
+        return mtrc("global", "%n week(s) ago", nullptr, weeks);
     }
 
     Date currentDate = currentDateTime.date();
     constexpr int monthInYear = 12;
     int months = (currentDate.year() - dateTime.year()) * monthInYear + (currentDate.month() - dateTime.month());
 
-    if (months == 1) {
-        return mtrc("global", "Last month");
-    }
-
-    if (months < monthInYear) {
-        return mtrc("global", "%1 months ago").arg(months);
+    if (months >= 1 && months < monthInYear) {
+        //: Omit the "%n" for the singular translation and translate (the equivalent of) to "Last month"
+        return mtrc("global", "%n month(s) ago", nullptr, months);
     }
 
     int years = currentDate.year() - dateTime.year();
-    if (years == 1) {
-        return mtrc("global", "1 year ago");
-    }
 
-    return mtrc("global", "%1 years ago").arg(years);
+    //: Omit the "%n" for the singular translation and translate to (the equivalent of) "Last year"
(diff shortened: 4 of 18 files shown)


src/engraving/libmscore/measurerepeat.cpp
```diff
@@ -297,7 +297,7 @@ Fraction MeasureRepeat::ticks() const
 
 String MeasureRepeat::accessibleInfo() const
 {
-    return mtrc("engraving", "%1; Duration: %2 measure(s)").arg(EngravingItem::accessibleInfo()).arg(numMeasures());
+    return mtrc("engraving", "%1; Duration: %n measure(s)").arg(EngravingItem::accessibleInfo()).arg(numMeasures());
```
RomanPudashkin, line 300: Looks like this change is incorrect

src/engraving/libmscore/noteentry.cpp
```diff
@@ -387,8 +387,8 @@ void Score::putNote(const Position& p, bool replace)
     staff_idx_t staffIdx = track2staff(_is.track());
     if (m->isMeasureRepeatGroup(staffIdx)) {
         auto b = MessageBox::warning(trc("engraving", "Note input will remove measure repeat"),
-                                     trc("engraving", "There is a measure repeat here.")
-                                     + trc("engraving", "\nContinue with adding note and delete measure repeat?"));
+                                     trc("engraving", "There is a measure repeat here.") + "/n"
```
cbjeukendrup, line 390: Wrong slash? (`/n` instead of `\n`)

src/engraving/libmscore/noteentry.cpp
```diff
@@ -387,8 +387,8 @@ void Score::putNote(const Position& p, bool replace)
...
+                                     + trc("engraving", "Continue with adding note and delete measure repeat?"));
```
cbjeukendrup, line 391: Actually, do we want a line break here at all? It looks like one continuous text. Personally I would change it like this:
```suggestion
                                     trc("engraving", "There is a measure repeat here. If you enter notes here, "
                                                      "it will be deleted. Do you want to continue?"));
```
(The string itself shouldn't contain a line break, I'm just spreading the literal over two lines.)

src/engraving/libmscore/noteentry.cpp
Jojo-Schmitz: Duh!

src/engraving/libmscore/noteentry.cpp
Jojo-Schmitz: Yes, I'd agree here

src/engraving/libmscore/noteentry.cpp
```diff
@@ -387,8 +387,8 @@ void Score::putNote(const Position& p, bool replace)
     staff_idx_t staffIdx = track2staff(_is.track());
     if (m->isMeasureRepeatGroup(staffIdx)) {
         auto b = MessageBox::warning(trc("engraving", "Note input will remove measure repeat"),
-                                     trc("engraving", "There is a measure repeat here.")
-                                     + trc("engraving", "\nContinue with adding note and delete measure repeat?"));
+                                     trc("engraving", "There is a measure repeat here."
+                                                      "Continue with adding note and delete measure repeat?"));
```
cbjeukendrup, line 391: A space is missing at the end of the first line. Now the string wil be
```
"There is a measure repeat here.Continue with adding note and delete measure repeat?"
```
Also, how about changing the string to something like this:
```
"This measure contains a measure repeat. If you enter notes here, it will be deleted. Do you want to continue?"
```
Isn't that better/nicer English?

src/engraving/types/typesconv.cpp
```diff
@@ -242,8 +242,8 @@ static const std::vector<Item<ElementType> > ELEMENT_TYPES = {
     { ElementType::COMPOUND,             "Compound",             QT_TRANSLATE_NOOP("engraving", "Compound") },
     { ElementType::CHORD,                "Chord",                QT_TRANSLATE_NOOP("engraving", "Chord") },
     { ElementType::SLUR,                 "Slur",                 QT_TRANSLATE_NOOP("engraving", "Slur") },
-    { ElementType::ELEMENT,              "EngravingItem",        QT_TRANSLATE_NOOP("engraving", "EngravingItem") },
-    { ElementType::ELEMENT_LIST,         "ElementList",          QT_TRANSLATE_NOOP("engraving", "EngravingItem list") },
+    { ElementType::ELEMENT,              "EngravingItem",        QT_TRANSLATE_NOOP("engraving", "Engraving item") },
+    { ElementType::ELEMENT_LIST,         "ElementList",          QT_TRANSLATE_NOOP("engraving", "Engraving item list") },
```
cbjeukendrup, line 246: I doubt about these ones; in other places in the UI, we still call it "Elements", for example "select similar elements", and I think "Element" is better as a user-exposed string than "Engraving item". I'd vote for reverting the usernames to "Element".

src/framework/global/dataformatter.cpp
```diff
@@ -44,48 +44,25 @@ String DataFormatter::formatTimeSince(const Date& dateTime)
...
     }
 
-    if (days == 1) {
-        return mtrc("global", "Yesterday");
-    }
-
-    if (days < 7) {
-        return mtrc("global", "%1 days ago").arg(days);
+    if (days >= 1 && days < 7) {
+        return mtrc("global", "%n day(s) ago", nullptr, days);
```
cbjeukendrup, line 48: I'd prefer to keep "Yesterday", "Last week" and "Last month". Just because "Yesterday" sounds a lot better to me than "1 day ago".
(Apart from that, I think it's good to use %n day(s) etc. here, to account for languages that have different forms for "dual" and plural.)

src/engraving/types/typesconv.cpp
Jojo-Schmitz: ok, done

src/engraving/libmscore/noteentry.cpp
Jojo-Schmitz: yep, done

src/framework/global/dataformatter.cpp
Jojo-Schmitz: This is still possible, via the translations for the singular form, I've now added a comment to that extent, which should shown on Transifex and so gie the translators a good hint.
IMHO the easier code wins here, and avoids the confusion that as per your idea the singlular would translated but never used.

Jojo-Schmitz: BTW: These singlular/plural things are one reason why we want an en_US 'translation'. Another is some strings from SMuFL which are really British English, but out of our hands to fix.

src/inspector/view/qml/MuseScore/Inspector/general/appearance/internal/HorizontalSpacingSection.qml
```diff
@@ -62,17 +62,17 @@ Item {
...
         anchors.right: parent.right
 
-        navigationName: "Bar width"
+        navigationName: "Measure width"
         navigationPanel: root.navigationPanel
         navigationRowStart: leadingValue.navigationRowEnd + 1
 
-        titleText: qsTrc("inspector", "Bar width")
-        propertyItem: root.barWidth
+        titleText: qsTrc("inspector", "Measure width")
```
Jojo-Schmitz, line 74: Not sure whether this is really meant?

cbjeukendrup: I think this change is okay. This affects the USER_STRETCH property; I'm not sure that this property is meant to be exposed in the Inspector (in MU3 it was not AFAIK), since the reset button behaves a bit weirdly. Anyway, let's reports those bugs later...

src/framework/global/dataformatter.cpp
```diff
@@ -44,48 +44,29 @@ String DataFormatter::formatTimeSince(const Date& dateTime)
         return mtrc("global", "Today");
     }
 
-    if (days == 1) {
-        return mtrc("global", "Yesterday");
```
RomanPudashkin, line 48: I strongly recommend not changing the implementation of this function

src/project/qml/MuseScore/Project/internal/MeasuresSettings.qml
```diff
@@ -46,15 +46,15 @@ FlatButton {
...
+                return qsTrc("project", "Pickup:") + " " +
                         model.pickupTimeSignature.numerator + "/" + model.pickupTimeSignature.denominator
             }
 
             return qsTrc("project", "No pickup")
         }
 
         font: ui.theme.largeBodyFont
-        text: model.measureCount + qsTrc("project", " measures, \n") + pickupMessage
+        text: model.measureCount + qsTrc("project", " measures.") + "\n" + pickupMessage
```
RomanPudashkin, line 57: Please explain the purpose of this change (I mean replacing ',' with '.')

cbjeukendrup, line 57: This should be
```suggestion
        text: qsTrc("project", "%1 measures").arg(model.measureCount) + ", \n" + pickupMessage
```

src/project/view/additionalinfomodel.cpp
```diff
@@ -318,19 +318,19 @@ QString AdditionalInfoModel::tempoAccessibleName(int noteIcon, bool withDot) con
...
-        name += qtrc("project", "semiquaver");
+        name += qtrc("project", "16th note");
         break;
     case MusicalSymbolCode::QUAVER:
-        name += qtrc("project", "quaver");
+        name += qtrc("project", "8th note");
         break;
     case MusicalSymbolCode::CROTCHET:
-        name += qtrc("project", "crotchet");
+        name += withDot ? qtrc("project", "quarter note") : qtrc("project", "Quarter note");
```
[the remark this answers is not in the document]
cbjeukendrup: We need to change this. We are assuming that all languages put the adjective "Dotted" at the start of the string, which is a very invalid assumption. This whole method should be:

```c++
QString AdditionalInfoModel::tempoAccessibleName(int noteIcon, bool withDot) const
{
    switch (static_cast<MusicalSymbolCode>(noteIcon)) {
    case MusicalSymbolCode::SEMIQUAVER:
        return withDot ? qtrc("project", "Dotted 16th note") : qtrc("project", "16th note");
    case MusicalSymbolCode::QUAVER:
        return withDot ? qtrc("project", "Dotted eighth note") : qtrc("project", "Eight note");
    case MusicalSymbolCode::CROTCHET:
        return withDot ? qtrc("project", "Dotted quarter note") : qtrc("project", "Quarter note");
    case MusicalSymbolCode::MINIM:
        return withDot ? qtrc("project", "Dotted half note") : qtrc("project", "Half note");
    case MusicalSymbolCode::SEMIBREVE:
        return withDot ? qtrc("project", "Dotted whole note") : qtrc("project", "Whole note");
    default:
        break;
    }

    return QString();
}
```

src/project/qml/MuseScore/Project/internal/MeasuresSettings.qml
Jojo-Schmitz: Potential next text starts with an uppercase letter.

src/project/view/additionalinfomodel.cpp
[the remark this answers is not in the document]
Jojo-Schmitz: Yes, that's better

src/project/qml/MuseScore/Project/MigrationDialog.qml
```diff
@@ -133,7 +133,7 @@ StyledDialogView {
                     horizontalAlignment: Qt.AlignLeft
                     verticalAlignment: Qt.AlignVCenter
 
-                    text: qsTrc("project", "This file was created in MuseScore %1").arg(dialog.appVersion)
+                    text: qsTrc("project", "This file was last saved with MuseScore %1").arg(dialog.appVersion)
```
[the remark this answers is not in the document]
cbjeukendrup: Let's change it to <code>"This file was <b><u>last saved</u></b> <b><u>in</u></b> MuseScore %1"</code> and everyone is happy.

src/framework/global/dataformatter.cpp
cbjeukendrup: Even if we decide to keep "Yesterday" etc. as a separate string, we will still need to use the `%n` way for the number of days. This article has a very good explanation why it's necessary: https://doc.qt.io/archives/qq/qq19-plurals.html

By the way, looking at how complex the rules for some languages are, I think we _do_ need to keep "Yesterday" as a separate string. Some languages use the "singular" form not only for `n == 1`.

src/project/qml/MuseScore/Project/internal/MeasuresSettings.qml
Jojo-Schmitz: OK. I'll leave the comma in the translatable string, for LTR languages

src/project/qml/MuseScore/Project/MigrationDialog.qml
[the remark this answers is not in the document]
Jojo-Schmitz: Well, "in" sounds wrong to me, but I'm not a native English speaker, so changed it.

igorkorsukov:
@Jojo-Schmitz @cbjeukendrup 

I suggest to merge this PR so that there are fewer conflicts. i.e. to discuss and edit smalls PRs. 
Do you have any objections?

cbjeukendrup:
Except for a few remaining comments, I'm very happy with this. So once those get fixed, I'll merge it!

src/project/qml/MuseScore/Project/MigrationDialog.qml
```diff
@@ -212,7 +212,7 @@ StyledDialogView {
...
```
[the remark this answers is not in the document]
Jojo-Schmitz: OK, but there are also still a few other plces that use "MuseScore version the score was last saved with"

src/project/qml/MuseScore/Project/internal/MeasuresSettings.qml
Jojo-Schmitz: done

src/framework/global/dataformatter.cpp
Jojo-Schmitz: running out of time, holidays starting and 2 weeks AFK...

Jojo-Schmitz:
> Except for a few remaining comments, I'm very happy with this. So once those get fixed, I'll merge it!

Please merge and amend as needed, I won't find the time to work on this any further until August

cbjeukendrup:
I'll do that! Thanks a lot, Jojo!

Issue #11173, opened 2022-04-12
[MU4 Issue] Huge cursor is rendered when editing dynamics on a score


boryskuzmenko:
**To Reproduce**
1. Create a score
2. Add fff dynamic mark to the score
3. Double-click the mark to edit
4. Pay attention to the cursor size

**Expected behavior**
Cursor size corresponds to the dynamics font size when editing dynamics on a score

**Screenshots**

https://user-images.githubusercontent.com/101250347/163053457-f135e9b6-d7eb-4be7-99db-dcb3cd12a3e6.mov

**Platform information**
 - OS: macOS, Windows, Linux

tzarjakob:
Can I work on this one?

Tantacrul:
@tzarjakob - have you made any progress?

tzarjakob:
No, you can unassign me

PR #12380, opened 2022-07-13
Avoid huge cursor when editing dynamics text

mike-spa:
Resolves: #11173 

In some cases (such as dynamics), symbols from the music font are used within blocks of text. The cursor then wants to adapt itself to the maximum height of the font in use, but in the case of the music font this means it becomes huge (see issue).
This PR ensures that the size of the cursor is always related to the text font.

@oktophonie @Tantacrul


Tantacrul:
I'm going to hand this review over to @bkunda.

I'm a little snowed under at the moment.

bkunda:
Looking at this now

bkunda:
Tested in Windows and Mac. Works as expected 🟢. FIXED

RomanPudashkin (approved):

PR #12392, opened 2022-07-14
fixed #11365: Implemented playback loading progress

### LINKED ISSUE #11365: Preloader for Audio Samples in Bottom Bar
**Task description**
When opening / creating a new score, we should have a preloader in the status bar that informs users that MS4 is still loading and preparing audio samples. This will be particularly useful for larger scores that use the new orchestral library.

image

For the developer who works on this, the dimensions can be found in Figma under **BOTTOM INFO BAR**

While this is displayed, it should replace the status bar info. When it is no longer needed, the status bar info can be displayed again (as shown in the image above).

Eism:
Resolves: #11365


RomanPudashkin (approved):

Tantacrul:
This works better than I thought it would! Can't find an issue with it.

Issue #11330, opened 2022-04-25
New MS4 Loading Screen


Tantacrul:
**Task description**
Essentially, this should work exactly like the MS3 loading screen, including the dynamic loading text.

I have provided it as a Figma file (so, all vector) but can provide a PNG version if needed.

Here's what it looks like
<img width="1157" alt="image" src="https://user-images.githubusercontent.com/21022311/165067152-de61ba3b-111d-49eb-984d-5bbf0d90a9e7.png">

We should use the text field here to describe important processes:

- **Scanning for VST plugins: name of plugin** (this is important, since there may be a lot of scanning involved the first time a user opens MS4)
- **Loading (process, file, etc.)** (I don't have suggestions here - anything the development team thinks is relevant / anything else that impacts opening time

(Ignore the 'restoring session' text. This is just a placeholder example)

RobFog:
I'd prefer "Restoring session" to "Resoring session". :-P

Jojo-Schmitz:
You're mentioning a sore point there...

;-)

HemantAntony:
@Tantacrul This looks really cool!

Tantacrul:
Resoring is a new technical process... that...

Tantacrul:
> You're mentioning a sore point there...
> 
> ;-)

Nah, it's cool to correct my typos :)

HemantAntony:
I wonder if there will be holiday variations of the splash screen like in MU3

PR #12396, opened 2022-07-14
Implement Loading Screen

cbjeukendrup:
Resolves: #11330

Schermafbeelding 2022-07-14 om 12 44 34

- If the user has set a custom font, this won't be used because the configuration is not initialized yet at the time that the loading screen is created
- Texts on the loading screen will not be translatable, since the translation system is not yet initialized.

If desired, these problems can be worked around by adding another method to `modularity::IModuleSetup` which will be called before the loading screen is created. 

This implementation uses QSplashScreen from QWidgets, because our custom IInteractive system is not yet initialized. However, this barely has any disadvantages as far as I can see, and it's an easy, compact implementation.


src/appshell/view/internal/splashscreen.h
```diff
@@ -0,0 +1,52 @@
...
+    //! We only use it for querying the default system font
+    INJECT(appshell, ui::IUiConfiguration, uiConfiguration)
+
+public:
+    SplashScreen();
+
+private:
+    void drawContents(QPainter* painter) override;
+
+    QSvgRenderer* m_backgroundRenderer;
```
RomanPudashkin, line 48: = nullptr

RomanPudashkin (approved):

PR #12398, opened 2022-07-14
fixed #11492: Renamed MuseScore_General to MS Basic

### LINKED ISSUE #11492: Rename "MuseScore_General_HQ" sound font to "MS Basic"
**Task description**
Hear me out...

One of the awkward things about our new mixer is that we've necessarily needed to keep each track quite narrow. As a result, some VST and sound font titles get truncated (a very common issue across plenty of DAWs and other notation apps).

In the case of **MuseScore_General_HQ** (which is now our default), I feel we can't have a situation where this is what people are greeted with when they first open the app:
image

Or possibly:
image

Eism:
Resolves: #11492


RomanPudashkin (approved):

cbjeukendrup:
Wouldn't we also need to rename it on the server where it is downloaded from? https://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General/
(see also build/cmake/DownloadSoundFont.cmake)

build/cmake/DownloadSoundFont.cmake
```diff
@@ -2,7 +2,7 @@
 
 set (SF_VERSION_REMOTE_FILE ${PROJECT_BINARY_DIR}/SF_VERSION)
 set (SF_VERSION_LOCAL_FILE ${PROJECT_SOURCE_DIR}/share/sound/SF_VERSION)
-set (DOWNLOAD_URL_ROOT "https://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General")
+set (DOWNLOAD_URL_ROOT "https://ftp.osuosl.org/pub/musescore/soundfont/MS Basic")
```
igorkorsukov, line 5: I recommend moving to S3, this ftp was used for historical reasons when there was no S3

Eism: moved

DmitryArefiev:
Tested #11492 on Win10, Mac12, LinuxMint 20.2 - **DONE**

PR #12404, opened 2022-07-14
[MU4] Fix bug re: slur direction for voices

asattely:
The issue:
image

Slurs were checking their whole length for voices, but they weren't checking the final tick. Now, the check happens not only for the duration of the slur, but also for the duration of both CR's:
image

### DIFF
--- a/src/engraving/libmscore/slur.cpp
+++ b/src/engraving/libmscore/slur.cpp
@@ -1321,15 +1321,21 @@ SpannerSegment* Slur::layoutSystem(System* system)
 
             _up = !(startCR()->up());
 
-            Measure* m1 = startCR()->measure();
-
+            Measure* m = startCR()->measure();
             if (c1 && c2 && !c1->isGrace() && isDirectionMixture(c1, c2)) {
                 // slurs go above if there are mixed direction stems between c1 and c2
                 // but grace notes are exceptions
                 _up = true;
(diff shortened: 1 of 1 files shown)


src/engraving/libmscore/slur.cpp
```diff
@@ -1327,7 +1327,7 @@ SpannerSegment* Slur::layoutSystem(System* system)
                 // slurs go above if there are mixed direction stems between c1 and c2
                 // but grace notes are exceptions
                 _up = true;
-            } else if (m1->hasVoices(startCR()->staffIdx(), tick(), ticks()) && c1 && !c1->isGrace()) {
+            } else if (m1->hasVoices(startCR()->staffIdx(), tick(), ticks() + c2->ticks()) && c1 && !c1->isGrace()) {
```
cbjeukendrup, line 1330: Are you sure that we don't need a nullptr check for `c2` here? Maybe wrap the whole if/elseif block in an `if (c1 && c2)` block?

asattely: oh, good call!

asattely:
There was also an issue when the different voices were in a different measure:
image

Which is also fixed:
image

src/engraving/libmscore/slur.cpp
asattely: We still need to get the tick / ticks for that CR, so I'll change it to check the ChordRest instead of the Chord

oktophonie (approved):

RomanPudashkin (approved):

PR #12414, opened 2022-07-14
Fix: #11060: Add missing shortcuts to Preferences

VanSHOE:
Resolves: #11060 
This PR intends to add the missing shortcuts in shortcuts list.
It also adds the shortcut contexts to the cpp files alongside newly added descriptions (which are now displayed in the Preferences page instead of the titles)

Note: There is an existing issue that you can't clear shortcuts that already have default shortcuts. I have not tried to resolve that issue in this PR because it has consequences to adding new default shortcuts. 

### DIFF
--- a/src/framework/shortcuts/view/shortcutsmodel.cpp
+++ b/src/framework/shortcuts/view/shortcutsmodel.cpp
@@ -22,6 +22,7 @@
 
 #include "shortcutsmodel.h"
 
+#include "ui/view/iconcodes.h"
 #include "translation.h"
 #include "log.h"
 
@@ -47,12 +48,12 @@ QVariant ShortcutsModel::data(const QModelIndex& index, int role) const
     const Shortcut& shortcut = m_shortcuts.at(index.row());
 
     switch (role) {
-    case RoleTitle: return this->action(shortcut.action).title;
+    case RoleTitle: return actionText(shortcut.action);
     case RoleIcon: return static_cast<int>(this->action(shortcut.action).iconCode);
     case RoleSequence: return sequencesToNativeText(shortcut.sequences);
     case RoleSearchKey: {
         const UiAction& action = this->action(shortcut.action);
-        return QString::fromStdString(action.code) + action.title + sequencesToNativeText(shortcut.sequences);
+        return QString::fromStdString(action.code) + action.title + action.description + sequencesToNativeText(shortcut.sequences);
     }
     }
 
@@ -68,6 +69,17 @@ const QString& ShortcutsModel::actionTitle(const std::string& actionCode) const
 {
     const UiAction& action = this->action(actionCode);
     return action.title;
+}
+
+const QString& ShortcutsModel::actionDescription(const std::string& actionCode) const
+{
+    const UiAction& action = this->action(actionCode);
+    return action.description;
+}
+
+const QString& ShortcutsModel::actionText(const std::string& actionCode) const
+{
+    return actionDescription(actionCode) == "" ? actionTitle(actionCode) : actionDescription(actionCode);
 }
 
 int ShortcutsModel::rowCount(const QModelIndex&) const
@@ -92,9 +104,15 @@ void ShortcutsModel::load()
     beginResetModel();
     m_shortcuts.clear();
 
-    for (const Shortcut& shortcut : shortcutsRegister()->shortcuts()) {
-        if (actionTitle(shortcut.action).isEmpty()) {
+    for (const UiAction& action : uiactionsRegister()->getActions()) {
+        if (action.title.isEmpty() || action.description.isEmpty()) {
             continue;
+        }
+
+        Shortcut shortcut = shortcutsRegister()->shortcut(action.code);
+        if (shortcut.action != action.code) {
+            shortcut.action = action.code;
+            shortcut.context = action.scCtx;
         }
 
         m_shortcuts << shortcut;
@@ -105,7 +123,7 @@ void ShortcutsModel::load()
     });
 
     std::sort(m_shortcuts.begin(), m_shortcuts.end(), [this](const Shortcut& s1, const Shortcut& s2) {
-        return actionTitle(s1.action) < actionTitle(s2.action);
+        return actionText(s1.action) < actionText(s2.action);
     });
 
     endResetModel();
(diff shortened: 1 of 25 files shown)


shoogle:
**Important:** This should be merged **after** #12426 has been merged. Once #12426 is merged @VanSHOE can rebase this one easily by dropping the final commit and re-running his Python script to insert the new action descriptions.

src/framework/ui/uitypes.h
```diff
@@ -188,25 +189,33 @@ enum class Checkable {
...
-    UiAction(const actions::ActionCode& code, UiContext ctx, const QString& title, Checkable ch = Checkable::No)
-        : code(code), context(ctx), title(title), description(title), checkable(ch) {}
-    UiAction(const actions::ActionCode& code, UiContext ctx, const QString& title, const QString& desc, Checkable ch = Checkable::No)
-        : code(code), context(ctx), title(title), description(desc), checkable(ch) {}
-    UiAction(const actions::ActionCode& code, UiContext ctx, const QString& title, const QString& desc, IconCode::Code icon,
+    UiAction(const actions::ActionCode& code, UiContext ctx, std::string scCtx, Checkable ch = Checkable::No)
+        : code(code), uiCtx(ctx), scCtx(scCtx), checkable(ch) {}
+
+    UiAction(const actions::ActionCode& code, UiContext ctx, std::string scCtx, const QString& title,  Checkable ch = Checkable::No)
+        : code(code), uiCtx(ctx), title(title), scCtx(scCtx), checkable(ch) {}
```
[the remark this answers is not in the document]
VanSHOE: Just pushed it @cbjeukendrup, please have a look

VanSHOE:
Rebased due to merge-conflict

igorkorsukov (approved):

shoogle:
@VanSHOE, rebase needed.

shoogle (approved):

bkunda:
Hurrah! 🙏🏻🎉

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -70,6 +71,17 @@ const QString& ShortcutsModel::actionTitle(const std::string& actionCode) const
...
 
+const QString& ShortcutsModel::actionDescription(const std::string& actionCode) const
+{
+    const UiAction& action = this->action(actionCode);
+    return action.description;
+}
+
+const QString& ShortcutsModel::actionText(const std::string& actionCode) const
+{
+    return actionDescription(actionCode) == "" ? actionTitle(actionCode) : actionDescription(actionCode);
```
cbjeukendrup, line 82: It looks like `actionDescription(actionCode) == ""` will never be true, because in `load()`, you do:
```c++
         if (action.title.isEmpty() || action.description.isEmpty()) {
             continue;
         }
```
Maybe you meant `&&` instead of `||`? Or is the emptiness check just redundant?

VanSHOE: I agree that this should never really be required. But just as a safety measure, if somehow a description was empty and it was allowed in the list(maybe inserted via some other measure and modified later on), it should then default to the title.

Eism:
@VanSHOE 
Crash: Select any element on the score and press Backspace
Please fix
image

src/framework/shortcuts/view/shortcutsmodel.cpp
Eism: @VanSHOE 
Please optimize the code, you are calculating the description twice.

VanSHOE:
Sure, I'm on it. Will create a separate PR with the crash fix, and also the action description optimization and tag it here

shoogle:
@VanSHOE, that might be appropriate as a temporary fix. You could submit that as a PR while you go looking for the real problem.

The real problem is that an action is created somewhere without being assigned an explicit context. The assert is there as a safeguard to flag when this occurs, so ideally you should find the contextless action and give it an explicit context. Also, it might help to have this assert at the point where actions are registered (i.e. during startup) rather than just having it at the point where actions are used. If it happens during startup then it can't be missed.

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -116,7 +134,9 @@ bool ShortcutsModel::apply()
     ShortcutList shortcuts;
 
     for (const Shortcut& shortcut : qAsConst(m_shortcuts)) {
-        shortcuts.push_back(shortcut);
+        if (!shortcut.sequences.empty()) {
```
Eism, line 137: @VanSHOE 
Can you explain to me why you filter empty shortcuts here?

cbjeukendrup: Otherwise we fill the shortcuts register (and thus the user shortcuts xml file) with _all_ actions that exist, while only a few have a shortcut.

However, @VanSHOE and me are aware that this solution is not perfect, because if the user deletes a default shortcut, this deletion is not saved.

So maybe we should accept that we store a big amount of unnecessary information in the shortcuts register. 

Or we should make this filter clever, by only saving user shortcuts that differ from the default ones instead of only saving non-empty user shortcuts. Maybe that's the best solution.

VanSHOE: There is another issue which occurs even if we do store the ones that are currently different from the default ones. 
If a default shortcut is added in the development process to the xml which was previously not assigned anything, it would not get assigned to any one if we choose store empty values in the register.

I was thinking that we have another flag with each shortcut where we can mark if it was user cleared. In that case, we do write it to the shortcuts register.

Other shortcuts that are just not assigned any shortcut can continue to not exist in the register.

Eism: There are more problems than benefits. I will revert this changes in https://github.com/musescore/MuseScore/pull/12890

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -92,11 +104,17 @@ void ShortcutsModel::load()
     beginResetModel();
     m_shortcuts.clear();
 
-    for (const Shortcut& shortcut : shortcutsRegister()->shortcuts()) {
-        if (actionTitle(shortcut.action).isEmpty()) {
+    for (const UiAction& action : uiactionsRegister()->getActions()) {
+        if (action.title.isEmpty() || action.description.isEmpty()) {
```
Eism, line 108: @VanSHOE
Could you explain why here OR?
Should be AND?

PR #12416, opened 2022-07-15
Fix #12402: Close popups when MuseScore looses focus

### LINKED ISSUE #12402: [MU4 Issue] Menu won't close when you click on the desktop panel/taskbar
**Describe the bug*
In MS3, if you had a menu open it would close if you click on the panel (linux) which lies outside the MS program window (Taskbar in Windows). However, the menu stays open if you do the same thing in MS4.

A similar issue was fixed in #10993.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2667920094, revision: github-musescore-musescore-0f1e7fb

iwoithe:
Resolves: #12402

**Additional Notes**
- Is `doUpdateEventFiletrs()` meant to be `doUpdateEventFilters()`? If so I would be happy to change this in either this PR or a new PR.
- I am aware of the method `PopupViewCloseController::onApplicationStateChanged()` but wasn't sure if I should use it. If the preferred solution is to use this method then I'll update the PR.

### DIFF
--- a/src/framework/uicomponents/view/internal/popupviewclosecontroller.cpp
+++ b/src/framework/uicomponents/view/internal/popupviewclosecontroller.cpp
@@ -53,6 +53,22 @@ void PopupViewCloseController::setActive(bool active)
     doUpdateEventFiletrs();
 }
 
+bool PopupViewCloseController::isDialog() const
+{
+    return m_isDialog;
+}
+
+void PopupViewCloseController::setIsDialog(bool isDialog)
+{
+    if (m_isDialog == isDialog) {
+        return;
+    }
+
+    m_isDialog = isDialog;
+
+    doUpdateEventFiletrs();
+}
+
 QQuickItem* PopupViewCloseController::parentItem() const
 {
     return m_parentItem;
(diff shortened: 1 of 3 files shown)


src/framework/uicomponents/view/internal/popupviewclosecontroller.cpp
```diff
@@ -115,6 +131,12 @@ bool PopupViewCloseController::eventFilter(QObject* watched, QEvent* event)
         }
     }
 
+    if (QEvent::ApplicationStateChange == event->type() && !m_isDialog) {
```
Eism, line 134: There is a processing of the application state change signal in the init method. see [line 37](https://github.com/musescore/MuseScore/blob/master/src/framework/uicomponents/view/internal/popupviewclosecontroller.cpp#L37)
Is this connect not working?

iwoithe: > There is a processing of the application state change signal in the init method

It did not work at one point when I was trying to reproduce the issue, however I have just checked and the issue does not exist on `master`. Not sure what I managed to do to reproduce the issue.

Apologies for this unneeded PR.

What is wished to be done regarding the following (from **Additional Notes**)?

> Is `doUpdateEventFiletrs()` meant to be `doUpdateEventFilters()`? If so I would be happy to change this in ~either this PR or~ a new PR.

Eism: > What is wished to be done regarding the following (from Additional Notes)?

@iwoithe yes, should be doUpdateEventFilters

> Apologies for this unneeded PR.

Don't worry. Thanks for your help

PR #12426, opened 2022-07-15
Add ampersand (&) to action titles to enable mnemonic access keys in menus

shoogle:
Add ampersand to action titles that appear in menus, excluding any titles that are also used outside of the menu as a tooltip or button text elsewhere in the UI (e.g. "Parts", "Mixer", "Undo", "Redo" and "Slur"). Adding it to those titles would have made the '&' visible in the UI.

I followed these guidelines in choosing mnemonics: https://docs.microsoft.com/en-gb/windows/win32/uxguide/cmd-menus#access-keys

**Important:** This should be merged before #12414 because that one can be easily rebased whereas this one cannot.

### DIFF
--- a/src/appshell/internal/applicationuiactions.cpp
+++ b/src/appshell/internal/applicationuiactions.cpp
@@ -46,7 +46,7 @@ const UiActionList ApplicationUiActions::m_actions = {
              ),
     UiAction("fullscreen",
              mu::context::UiCtxAny,
-             QT_TRANSLATE_NOOP("action", "Enter full screen"),
+             QT_TRANSLATE_NOOP("action", "&Full screen"),
              QT_TRANSLATE_NOOP("action", "Enter full screen"),
              Checkable::Yes
              ),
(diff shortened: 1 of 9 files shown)


src/appshell/internal/applicationuiactions.cpp
```diff
@@ -46,7 +46,7 @@ const UiActionList ApplicationUiActions::m_actions = {
              ),
     UiAction("fullscreen",
              mu::context::UiCtxAny,
-             QT_TRANSLATE_NOOP("action", "Enter full screen"),
+             QT_TRANSLATE_NOOP("action", "&Full screen"),
```
shoogle, line 49: Title change approved by @Tantacrul

src/notation/internal/notationuiactions.cpp
```diff
@@ -356,32 +356,32 @@ const UiActionList NotationUiActions::m_actions = {
              ),
     UiAction("slash-rhythm",
              mu::context::UiCtxNotationOpened,
-             QT_TRANSLATE_NOOP("action", "Toggle 'rhythmic slash notation'")
+             QT_TRANSLATE_NOOP("action", "Toggle rhythmic sl&ash notation")
```
shoogle, line 359: Title change approved by @bkunda.

igorkorsukov:
I believe that it is not necessary to insert ampersands into the translatable strings. As far as I know, the position of the ampersand depends on the shortcut, if the shortcut has a different value, then the ampersand should be in a different place. Or I have erroneous information?

cbjeukendrup:
@igorkorsukov These ampersands are separate from shortcuts. They define so-called mnemonics: a menu item can be triggered by pressing Alt + the letter after the ampersand. Those can be seen as very local shortcuts, that are only active if the menu containing them is currently opened. 

Because the text is different in every language, the mnemonic also needs to be different. That is the reason that the ampersand needs to be in the translatable string, so that the translator can put it where they want. The only restriction is that mnemonics must be unique per menu; it is the translator's responsibility to ensure that. (That's not great... but it's always been like that.)

igorkorsukov:
> @igorkorsukov These ampersands are separate from shortcuts. They define so-called mnemonics: a menu item can be triggered by pressing Alt + the letter after the ampersand. Those can be seen as very local shortcuts, that are only active if the menu containing them is currently opened.
> 
> Because the text is different in every language, the mnemonic also needs to be different. That is the reason that the ampersand needs to be in the translatable string, so that the translator can put it where they want. The only restriction is that mnemonics must be unique per menu; it is the translator's responsibility to ensure that. (That's not great... but it's always been like that.)

thank you for your explanation!

Eism (approved):

Issue #12387, opened 2022-07-13
[MU4 Issue] Key signature changes don't show for transposing instruments when not in concert pitch


DmitryArefiev:
**Describe the bug**
See compare with previous build and latest master

https://user-images.githubusercontent.com/90187801/178800709-3ed459ac-6c61-492d-8560-682a2d608415.mp4

**Steps to reproduce**
1. Open Wind Quartet template
2. Turn on/off concert pitch mode - you should see key signature change in Clarinet instrument

This occurs only if you start from a scratch without adding a new key signatures to the score

**Platform information**
macOS, Windows, Linux

**Additional context**

DmitryArefiev:
Another related weird thing

https://user-images.githubusercontent.com/90187801/178801873-280e509f-f1f6-453e-8e1e-f0dccfabe224.mp4

Eism:
@cbjeukendrup 
I assigned you because #12240 is the cause of this problem

DmitryArefiev:
Fixed in #12427

PR #12427, opened 2022-07-15
Get rid of KeyMode when creating a new score

cbjeukendrup:
Since a KeyMode is never communicated to QML, the score creation options passed from QML won't contain a KeyMode either. Still, we try to get the KeyMode, to assign it to `scoreOptions.keyMode`, in `NewScoreModel::parseOptions`. Thus, we get the default int value namely 0, and convert that to a KeyMode, which gives KeyMode::NONE, which is the "dangerous" mode for custom/atonal key signatures. Key signatures with this mode behave very differently when it comes to transposing instruments, which is undesired for normal key signatures. 

Thus, we get rid of the ScoreCreationOptions.keyMode entirely for now, since it will never contain a correct value. (And in practice, the keyMode doesn't matter anything at all, the only place where it is used is in MusicXML export.) This is a bit of a quick fix for now, and we might consider bringing it back later.

Resolves: #12387

### DIFF
--- a/src/notation/internal/masternotation.cpp
+++ b/src/notation/internal/masternotation.cpp
@@ -138,7 +138,6 @@ static void createMeasures(mu::engraving::Score* score, const ScoreCreateOptions
         ks.setMode(KeyMode::NONE);
     } else {
         ks.setKey(scoreOptions.key);
-        ks.setMode(scoreOptions.keyMode);
     }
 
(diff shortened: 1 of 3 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #12387 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Issue #12389, opened 2022-07-13
[MU4 Task] Audit translatable strings


cbjeukendrup:
**Progress:**
- [x] InstrumentsXML
- [x] accessibility
- [x] action
- [x] appshell
- [x] appshell/about
- [x] appshell/dock
- [x] appshell/gettingstarted
- [x] appshell/preferences
- [x] audio
- [x] cloud
- [x] engraving
- [x] engraving/sym
- [x] global
- [x] iex_capella
- [x] iex_guitarpro
- [x] iex_midi
- [x] iex_musicxml
- [x] inspector
- [x] instruments
- [x] languages
- [x] learn
- [x] mpe
- [x] notation
- [x] notation/editstyle
- [x] notation/timeline
- [x] palette
- [x] palette/unicodeRanges
- [x] playback
- [x] plugins
- [x] project
- [x] project/export
- [x] project/migration
- [x] project/newscore
- [x] project/properties
- [x] project/save
- [x] shortcuts
- [x] ui
- [x] uicomponents
- [x] workspace

PR #12428, opened 2022-07-15
Audit translatable strings. Step 5

cbjeukendrup:
See #12389


src/palette/qml/MuseScore/Palette/internal/Palette.qml
```diff
@@ -190,6 +190,7 @@ StyledGridView {
                 }
             }
 
+            //: Caption of a buttton to reveal more elements
```
igorkorsukov, line 193: `buttton` - typo

cbjeukendrup: Thanks, fixed!

PR #12435, opened 2022-07-15
run_lupdate.sh: don't forget .h, .mm and, most notably, .ui files

cbjeukendrup:
Resolves: https://discord.com/channels/818804595450445834/818852942882275379/997602246252564491

That brings the total number of strings from 3660 to 4819...

### DIFF
--- a/share/locale/musescore_en.ts
+++ b/share/locale/musescore_en.ts
@@ -2,6 +2,4694 @@
 <!DOCTYPE TS>
 <TS version="2.1" language="en">
 <context>
+    <name>AlignSelect</name>
+    <message>
+        <location filename="../../src/notation/view/widgets/align_select.ui" line="51"/>
+        <source>Align left edge of text to reference point</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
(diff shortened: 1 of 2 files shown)


cbjeukendrup:
Unfortunately, there is no way to set the context explicitly for .ui files. I'm not planning to spend much time going through these strings from .ui files, since they are unchanged since MU3, and the .ui files will be removed in the future anyway.

igorkorsukov (approved):

igorkorsukov:
> By the way, @igorkorsukov, since these strings and their contexts should largely be exactly the same as in MU3, is there any chance that we can do some Transifex magic to restore these translations from MU3? That would save the translators tons of work... or will Transifex still somehow remember the old translations, once we put back these strings?

I think the translation will be keeped if the context doesn't change.

cbjeukendrup:
That sounds hopeful! I'll merge this PR then. Is it okay if I run `CI_Run_lupdate_to_tx` after that?

igorkorsukov:
> That sounds hopeful! I'll merge this PR then. Is it okay if I run `CI_Run_lupdate_to_tx` after that?

Yes, you can run lupdate to tx

PR #12444, opened 2022-07-17
Fix #12372: remember maximized/windowed state when restoring minimized window

### LINKED ISSUE #12372: [MU4 Issue] App dialog position not restored correct after minimize on Windows
**Describe the bug**
When the app dialog is minimized, I expected it will restore the state it was before I minized the app. This works fine in MU3 but does not work in MU4. When the app is restored, it returns to the initial dialog size (not maximized).

**To Reproduce**
1. Start MU4
2. Click the Maximize option
3. Click the Minimize option
4. Click the app icon in the taskbar to restore the window

**Expected behavior**
The app should be restored to the previous state.

**Screenshots**
Showing the issue in MU4:
MU4-Maximize-minimize

Showing expected working using MU3:
MU3-Maximize-minimize

**Platform information**
 - OS: Windows 11

martin-danhier:
Resolves: https://github.com/musescore/MuseScore/issues/12372

Hello,

I found a Qt bug report about this issue: https://bugreports.qt.io/browse/QTBUG-69711. However, I'm not sure if it is a bug or a breaking change, because in the version where this bug was introduced, a new [`QWindow::setWindowStates`](https://doc.qt.io/qt-6/qwindow.html#setWindowStates) method was added, which allows to set multiple window states at the same time. This can be used to minimize without losing the previous state.

Using this function, it works as expected. However, I didn't find a way to use it directly in QML with the `ApplicationWindow`, so I called it via the `MainWindowManager`.

Since the bug report only mentioned Windows, I only changed the Windows-related code, but it might be worth it to try it on other OS's as well.


Eism:
@martin-danhier thanks for fixing!
I have suggestions: please rename the showMinimizedWin methods to showMinimizedWithSavePreviousState and leave a comment(from AppWindow.qml) only in the model.

Also please squash the two commits into one(drop fix codestyle)

martin-danhier:
@Eism Thanks for your feedback! I applied the changes you suggested.

DmitryArefiev:
Tested #12372 on **Win10**, Mac12, LinuxMint 20.2 - **FIXED**

Eism (approved):

PR #12446, opened 2022-07-18
[gp] Updated utest refs

igorkorsukov:


handrok (approved):

src/importexport/guitarpro/tests/environment.cpp
```diff
@@ -50,6 +50,7 @@ static mu::testing::SuiteEnvironment importexport_se(
     mu::engraving::ScoreRW::setRootPath(mu::String::fromUtf8(iex_guitarpro_tests_DATA_ROOT));
 
     mu::engraving::MScore::testMode = true;
+    mu::engraving::MScore::testWriteStyleToScore = false;
```
cbjeukendrup, line 53: @igorkorsukov Are you sure that the style shouldn't be tested? I think it should, since the style can influence which EngravingItems are visible on tab staves and which not. (@alexpavlov96 you might also want to comment)

igorkorsukov: > @igorkorsukov Are you sure that the style shouldn't be tested? I think it should, since the style can influence which EngravingItems are visible on tab staves and which not. (@alexpavlov96 you might also want to comment)

This is a writing of the style in the score file, it has nothing to do with the import of the gp (perhaps the style that is read during import is important). 
We refused to store the style in the score file. 
At the moment style is written to a score file in test mode only because of compatibility with the reference files of the engraving tests. We need to update them.

PR #12449, opened 2022-07-18
Fix #12278: bound line2pitch if line below minimum

### LINKED ISSUE #12278: [MU4 Issue] Odd behavior when dragging note downwards
**Describe the bug**
Dragging a note downwards results in odd behavior, causing note to eventually jump upwards.

**To Reproduce**
1. Enter a note
2. Keep dragging the note down.
3. See error

**Expected behavior**
It would be ideal if the note 'locked' in place when it reaches the lowest pitch. This behavior isn't exhibited when dragging the note upwards.

**Screenshots**

https://user-images.githubusercontent.com/61588096/177323164-0df9da71-808d-47ed-a3df-b499c18ce7b6.mov

**Platform information**
 - OS: macOS
 - Version: 4.0.0.2607659352

martin-danhier:
Resolves: https://github.com/musescore/MuseScore/issues/12278

Hello,

This issue appears to be caused by the `line2pitch` function. If the line was below the minimum, the function would add octaves until the note is valid. Replacing that behavior by a simply setting notes that are too low to the minimum fixes the issue.

However, I suppose this behavior had a purpose in one of the usage of the `line2pitch` function. But in this case, why wouldn't there be a similar behavior for notes that are too high?

### DIFF
--- a/src/engraving/libmscore/utils.cpp
+++ b/src/engraving/libmscore/utils.cpp
@@ -455,10 +455,10 @@ int line2pitch(int line, ClefType clef, Key key)
 {
     int l      = ClefInfo::pitchOffset(clef) - line;
     int octave = 0;
-    while (l < 0) {
-        l += 7;
-        octave++;
+    if (l < 0) {
+        l = 0;
     }
+
     octave += l / 7;
     l       = l % 7;
 


Tantacrul:
@bkunda - if you get time, can you test this one and say whether you're happy that it fixes the problem? If so, then we can push to merge.

Thanks @martin-danhier !

bkunda:
🟢 Tested on MacOS and Windows. This does seem to fix the problem! 
Thank you @martin-danhier!

Eism (approved):

sammik (commented):
Apparently late to the party, but I think, in this case `octave += l / 7` doesn't make sense (semantically, result is the same, of course).

I think this is more clear (sorry, I dont know, how to review directly into code):
```
    int l = ClefInfo::pitchOffset(clef) - line;
    if (l < 0) {
        l = 0;
    }

    int octave = l / 7;
    l = l % 7;
```

PR #12451, opened 2022-07-18
Extending minimum tie length implementation to apply at start of system

mike-spa:
At the moment, minimum tie length is only enforced between adjacent chords. That was by design in my first implementation. @oktophonie pointed out that it can be necessary to enforce it also at the start of a system. That case requires special handling, which is now implemented.
Additionally, the space between the end of the header and the start of the tie (which used to be determined by an anonimous variable set to 1 sp) has been moved into a dedicated style setting (though not exposed in UI).

Before:
Immagine 2022-07-15 181819

After
Immagine 2022-07-15 182043

### DIFF
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -844,10 +844,8 @@ void TieSegment::addLineAttachPoints()
     // Add tie attach point to start and end note
     Note* startNote = tie()->startNote();
     Note* endNote = tie()->endNote();
-    if (endNote && endNote->findMeasure() == startNote->findMeasure()) {
-        startNote->addLineAttachPoint(ups(Grip::START).pos(), tie());
-        endNote->addLineAttachPoint(ups(Grip::END).pos(), tie());
-    }
+    startNote->addLineAttachPoint(ups(Grip::START).pos(), tie());
+    endNote->addLineAttachPoint(ups(Grip::END).pos(), tie());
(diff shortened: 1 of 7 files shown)


oktophonie (approved):

src/engraving/libmscore/tie.cpp
```diff
@@ -839,10 +839,8 @@ void TieSegment::addLineAttachPoints()
     // Add tie attach point to start and end note
     Note* startNote = tie()->startNote();
     Note* endNote = tie()->endNote();
-    if (endNote && endNote->findMeasure() == startNote->findMeasure()) {
-        startNote->addLineAttachPoint(ups(Grip::START).pos(), tie());
-        endNote->addLineAttachPoint(ups(Grip::END).pos(), tie());
-    }
+    startNote->addLineAttachPoint(ups(Grip::START).pos(), tie());
+    endNote->addLineAttachPoint(ups(Grip::END).pos(), tie());
```
RomanPudashkin, line 843: Is the null-check no longer needed here?

mike-spa: Good point. In principle, the null-check isn't strictly necessary, because startNote and endNote are never null. But I think it's safer if I null-check anyway, just in case we encounter a broken tie or something. I'll quickly fix this

RomanPudashkin (approved):

PR #12455, opened 2022-07-18
[lang] added request of restart for change lang

igorkorsukov:
image

### DIFF
--- a/src/appshell/view/preferences/generalpreferencesmodel.cpp
+++ b/src/appshell/view/preferences/generalpreferencesmodel.cpp
@@ -35,10 +35,12 @@ void GeneralPreferencesModel::load()
 {
     languagesConfiguration()->currentLanguageCode().ch.onReceive(this, [this](const QString& languageCode) {
         emit currentLanguageCodeChanged(languageCode);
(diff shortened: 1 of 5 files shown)


src/appshell/view/preferences/generalpreferencesmodel.cpp
```diff
@@ -35,10 +35,12 @@ void GeneralPreferencesModel::load()
 {
     languagesConfiguration()->currentLanguageCode().ch.onReceive(this, [this](const QString& languageCode) {
         emit currentLanguageCodeChanged(languageCode);
+        sesIsNeedRestart(true);
     });
 
     languagesService()->languages().ch.onReceive(this, [this](const LanguagesHash&) {
         emit languagesChanged(languages());
+        sesIsNeedRestart(true);
```
igorkorsukov, line 43: I expected that the notification would be sent only if the language changed, but it seems to be always sent. I hope it doesn't matter that now any language update check will require us to restart the application.

Eism (approved):

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - **PASS**

PR #12460, opened 2022-07-18
Small fixes in the EditPitch dialog

cbjeukendrup:
- fix memory leaks
- only allow single selection
- remove non-working clef icons

### DIFF
--- a/src/framework/ui/view/widgetnavigationfix.cpp
+++ b/src/framework/ui/view/widgetnavigationfix.cpp
@@ -27,9 +27,9 @@
 
 using namespace mu::ui;
 
-bool WidgetNavigationFix::fixNavigationForTableWidget(const WidgetNavigationFix::NavigationChain* chain, int key)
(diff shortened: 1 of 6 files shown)


igorkorsukov (approved):

src/framework/ui/view/widgetnavigationfix.cpp
```diff
@@ -27,29 +27,29 @@
 
 using namespace mu::ui;
 
-bool WidgetNavigationFix::fixNavigationForTableWidget(const WidgetNavigationFix::NavigationChain* chain, int key)
+bool WidgetNavigationFix::fixNavigationForTableWidget(const WidgetNavigationFix::NavigationChain& chain, int key)
```
cbjeukendrup, line 30: @Eism, just to be sure: was there any reason to use pointers here?

cbjeukendrup: Okay, thanks. I'm confident that the change is safe, so I'll merge it.

Issue #12432, opened 2022-07-15
[MU4 Issue] Crash after adding slur to linked staff (in some cases)


DmitryArefiev:
See video for steps

https://user-images.githubusercontent.com/90187801/179276338-3f7e6da6-0399-4607-8dbd-756c295664a2.mp4

**Steps to reproduce**
1. Open String Quartet template
2. Open all parts
3. In Score, add staff to Violin I (Instruments panel>Add staff)
4. In Violin I part, add linked staff to the bottom staff (Instruments panel>Bass clef>Settings icon>Create a linked staff)
5. In Score>Violin I section>Bottom staff (3rd staff), add C note and hit S (add slur)

**Platform information**
 Windows10, Mac12

wizofaus:
Happy to look at this if there's a chance we could get the repro steps in text?

wizofaus:
However I was able to get it crash based on the description of the bug:


read access violation
this was nullptr
 	MuseScore4.exe!mu::engraving::EngravingObject::type() Line 222	C++
 	MuseScore4.exe!mu::engraving::EngravingObject::isChord() Line 309	C++
>	MuseScore4.exe!mu::engraving::Spanner::findStartChord() Line 948	C++
 	MuseScore4.exe!mu::engraving::Spanner::startChord() Line 886	C++
 	MuseScore4.exe!mu::engraving::LayoutSystem::processLines(mu::engraving::System * system, std::vector<mu::engraving::Spanner *,std::allocator<mu::engraving::Spanner *>> lines, bool align) Line 1283	C++
 
Steps were similar to video, though I haven't worked out exactly what sequence is necessary (create from string quartet template, open parts, in one of the parts, then in a part, add a staff then add a linked staff, then try to add notes and a slur to the linked staff). Even if it doesn't crash initially it often does on undo or closing score etc.
Also had it crash here:

>	MuseScore4.exe!mu::engraving::Staff::part() Line 144	C++
 	MuseScore4.exe!mu::notation::NotationAccessibility::singleElementAccessibilityInfo() Line 181	C++
 	MuseScore4.exe!mu::notation::NotationAccessibility::updateAccessibilityInfo() Line 115	C++
 	MuseScore4.exe!mu::notation::NotationAccessibility::{ctor}::__l2::<lambda>() Line 54	C++
 	[External Code]	
[...]

PR #12475, opened 2022-07-19
[MU4] Fix 12432: Ensure staves that are not added to an excerpt are unlinked (and deleted)

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12432

There was a bug in the interaction between `MasterNotationParts::appendStaff()` and `NotationParts::appendStaff()`. When creating linked clones of staves to add to excerpts, the linked clone is created before it's even determined whether we need to add that clone to the excerpt. If not, the process of creating a linked clone saves itself in all of its would-be siblings lists of siblings. The fix for that was to make sure that we unlink any staff that is not added to the excerpt in question.

I also noticed that there was a bit of a memory leak, where new Staves were being allocated and not being deleted in the situation described above, so I added some `delete`s in there. I know MuseScore is a huge codebase with tons of macros and whatever magic Qt does, so if there's some behind-the-scenes reference counting going on that makes those `delete`s unnecessary, I can remove them. But I didn't find anything to suggest there was!

### DIFF
--- a/src/notation/internal/notationparts.cpp
+++ b/src/notation/internal/notationparts.cpp
@@ -475,13 +475,17 @@ void NotationParts::setStaffConfig(const ID& staffId, const StaffConfig& config)
     notifyAboutStaffChanged(staff);
 }
 
-void NotationParts::appendStaff(Staff* staff, const ID& destinationPartId)
+bool NotationParts::appendStaff(Staff* staff, const ID& destinationPartId)
 {
     TRACEFUNC;
 
+    IF_ASSERT_FAILED(staff) {
+        return false;
+    }
+
     Part* destinationPart = partModifiable(destinationPartId);
-    if (!staff || !destinationPart) {
-        return;
+    if (!destinationPart) {
+        return false;
     }
 
     startEdit();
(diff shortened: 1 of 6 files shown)


src/notation/internal/notationparts.cpp
```diff
@@ -481,6 +481,8 @@ void NotationParts::appendStaff(Staff* staff, const ID& destinationPartId)
 
     Part* destinationPart = partModifiable(destinationPartId);
     if (!staff || !destinationPart) {
+        staff->unlink();
```
cbjeukendrup, line 484: `staff` may be null at this point, in which case the added lines will cause a crash. Better create a separate assert for `staff`:
```
    TRACEFUNC;

    IF_ASSERT_FAILED(staff) {
        return;
    }

    Part* destinationPart = partModifiable(destinationPartId);
    if (!destinationPart) {
        staff->unlink();
        ...
```
(I think it's fair to use an assert for `staff`, isn't it?)

(same in `appendLinkedStaff`)

asattely: Oh, really good catch, I dont know where my brain is today

asattely: Is there a reason that's a better choice than 

`if (!staff || !destinationPart) {
        if (staff) {
            staff->unlink();
            delete staff;
        }
        return;
    }`

Just a regular null check?

cbjeukendrup: Mainly for semantical reasons, I would use an assert for `staff`, to indicate that we consider it a mistake if `staff` is null. But functionally it won't really make a difference. 

(In any case I personally think it would look clearer if we'd do the check for `staff` separately, at the top of the function, instead of inside the `if (!staff || !destinationPart)` block.)

src/notation/internal/notationparts.cpp
```diff
@@ -495,9 +501,15 @@ void NotationParts::appendLinkedStaff(Staff* staff, const ID& sourceStaffId, con
...
+    IF_ASSERT_FAILED(staff) {
+        return;
+    }
+
     Staff* sourceStaff = staffModifiable(sourceStaffId);
     Part* destinationPart = partModifiable(destinationPartId);
-    if (!staff || !sourceStaff || !destinationPart) {
+    if (!sourceStaff || !destinationPart) {
+        staff->unlink();
+        delete staff;
```
RomanPudashkin, line 512: I'm not a big fan of when a function deletes a passed argument (and it's not a data cleaning function like qDeleteAll). So a better solution would be to return bool from this function and delete the object at the level above

asattely: No problemo, done and done. I wasn't sure if it was a wise idea to mess with the interface itself, but I guess the only two classes that implement it are the ones I'm working on, so I guess it's not a big deal

RomanPudashkin (approved):

PR #12479, opened 2022-07-19
Fix conflict checking with multiple shortcuts

VanSHOE:
When comparing two shortcuts, before it did a single string comparison between the conjoined string form of the sequences. This caused an issue because if there are two shortcuts, and one has, for example, `Ctrl+Z` as the shortcut and some other shortcut, say as its for UNDO `Ctrl+Z; Alt + Backspace`, then even if I assign `Ctrl+Z` to some other shortcut it would not conflict because the `Alt+Backspace` would cause the string to not match and hence not cause the conflict.

image
This allowed me to set the same shortcut to say Redo
image

Another such example can been seen here (the shortcuts of marcato were set by me in shortcuts.xml)
image

This PR just compares each shortcut and flags the conflict on first sequence match

### DIFF
--- a/src/framework/shortcuts/view/editshortcutmodel.cpp
+++ b/src/framework/shortcuts/view/editshortcutmodel.cpp
@@ -116,14 +116,18 @@ void EditShortcutModel::inputKey(int key, Qt::KeyboardModifiers modifiers)
 void EditShortcutModel::checkNewSequenceForConflicts()
 {
     m_conflictShortcut.clear();
-    QString input = newSequence();
+    const std::string input = newSequence().toStdString();
 
     for (const QVariant& shortcut : m_potentialConflictShortcuts) {
         QVariantMap map = shortcut.toMap();
 
-        if (map.value("sequence").toString() == input) {
-            m_conflictShortcut = map;
-            return;
+        std::vector<std::string> toCheckSequences = Shortcut::sequencesFromString(map.value("sequence").toString().toStdString());
+
+        for (const std::string& toCheckSequence : toCheckSequences) {
+            if (input == toCheckSequence) {
+                m_conflictShortcut = map;
+                return;
+            }
         }
     }
 }


src/framework/shortcuts/view/editshortcutmodel.cpp
```diff
@@ -117,13 +117,20 @@ void EditShortcutModel::checkNewSequenceForConflicts()
 {
     m_conflictShortcut.clear();
     QString input = newSequence();
+    std::vector<std::string> inputSequences = Shortcut::sequencesFromString(input.toStdString());
```
RomanPudashkin, line 120: This conversion is not needed because we no longer allow the user to assign multiple shortcuts to one action

src/framework/shortcuts/view/editshortcutmodel.cpp
```diff
@@ -117,13 +117,20 @@ void EditShortcutModel::checkNewSequenceForConflicts()
...
     for (const QVariant& shortcut : m_potentialConflictShortcuts) {
         QVariantMap map = shortcut.toMap();
 
-        if (map.value("sequence").toString() == input) {
-            m_conflictShortcut = map;
-            return;
+        std::vector<std::string> toCheckSequences = Shortcut::sequencesFromString(map.value("sequence").toString().toStdString());
+
+        for (std::string inputSequence : inputSequences) {
+            for (std::string toCheckSequence : toCheckSequences) {
```
RomanPudashkin, line 128: Please, use const std::string&

VanSHOE:
@RomanPudashkin I have incorporated the suggestions. Please have a look

src/framework/shortcuts/view/editshortcutmodel.cpp
```diff
@@ -121,9 +121,13 @@ void EditShortcutModel::checkNewSequenceForConflicts()
     for (const QVariant& shortcut : m_potentialConflictShortcuts) {
         QVariantMap map = shortcut.toMap();
 
-        if (map.value("sequence").toString() == input) {
-            m_conflictShortcut = map;
-            return;
+        std::vector<std::string> toCheckSequences = Shortcut::sequencesFromString(map.value("sequence").toString().toStdString());
+
+        for (const std::string& toCheckSequence : toCheckSequences) {
+            if (input.toStdString() == toCheckSequence) {
```
RomanPudashkin, line 127: Please convert the "input" string outside the loops

VanSHOE: Done, also made it const

Issue #12440, opened 2022-07-16
[MU4 Issue] Custom grace note positions in tablature are lost when MS3 file is opened


rgreen5:
**Describe the bug**

When you open an MS3 file in MS4, the _custom_ positions of _tablature_ grace notes are lost making the tablature very messy. (Tablature grace notes at default positions are not affected.)

**To reproduce**

1. Open the attached MS3 speciemen score. 
[specimen_score.zip](https://github.com/musescore/MuseScore/files/9125608/specimen_score.zip)

2. Inspect the grace note positions.
3. Compare it with MS3.

RESULT: In the tablature, measure 2, the grace note position, which was at default, is preserved. In measure 4, the custom-positioned grace notes (and the slur around them) are way out of position. They have to be reset by the user and repositioned.
![grace_notes](https://user-images.githubusercontent.com/87246583/179352090-fd0885fb-275f-40d0-8a4a-f045ba764588.png)

**Expected behavior**

Custom grace note positions in tablature should be preserved.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2680600421, revision: github-musescore-musescore-315cfb3

mike-spa:
Two separate problems are causing this issue. 
1) A small error in grace notes layout which causes them to apply a larger offset than requested. This is now fixed in #12469.
2) People used to apply manual offsets because default positioning was often bad in older versions, particularly with grace notes. In 4.0, however, the default position is usually correct, and applying the old offsets will most likely make it wrong. We will takle this separately, probably by resetting grace notes position when migrating older scores.

PR #12485, opened 2022-07-20
Reset grace notes offset when migrating from pre-4.0

mike-spa:
Resolves: #12440 

The default grace notes spacing used to be quite bad in older versions: way too narrow in normal staves, way too wide in TAB staves. Users would often correct by applying manual offset to individual notes. However, the default spacing in 4.0 is much improved, which means that applying the old offsets would often result in a completely wrong layout. This PR resets manual offset of grace notes when importing from 4.0. @oktophonie 

Musescore 3.6 default:
MU3Default
Musescore 3.6 with manual corrections:
MU3Offset
Musescore 4 with the same offsets retained:
MU4Offset
Musescore 4 default:
MU4Default

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1219,6 +1219,10 @@ void Chord::read(XmlReader& e)
             e.unknown();
         }
     }
+    // Reset horizontal offset of grace notes when migrating from before 4.0
+    if (isGrace() && score()->mscVersion() < 400) {
+        rxoffset() = 0;
+    }
 }
 
 //---------------------------------------------------------


oktophonie (approved):

src/engraving/libmscore/chord.cpp
```diff
@@ -1295,6 +1295,10 @@ bool Chord::readProperties(XmlReader& e)
     } else {
         return false;
     }
+    // Reset horizontal offset of grace notes when migrating from before 4.0
+    if (isGrace() && score()->mscVersion() < 400) {
+        rxoffset() = 0;
+    }
```
cbjeukendrup, line 1301: I think this should be in the `read` method, because this `readProperties` method is called in a loop, basically once for every xml tag inside the <Chord> tag, and thus not all properties are initialized yet at the bottom of the `readPropeties` method (only one property is initialized per call to this method).

mike-spa: Ha! You're totally right. Gonna do that now.

cbjeukendrup (approved):

PR #12490, opened 2022-07-20
[MU4] Fix #12465: Force instrument selection upon removing last instrument

### LINKED ISSUE #12465: [MU4 Issue] Crash when using playback toolbar in a score with no instruments
**Describe the bug**

**To Reproduce**
1. Create a score
3. Save and close the score
4. Reopen the score
A crash occurs.

**Expected behavior**

**Platform information**
 - OS: reported on macOS

hodori0719:
Resolves: https://github.com/musescore/MuseScore/issues/12465

Though the issue specifically describes crashes with the metronome feature on instrument-less scores, such scores are prone to an unspecified number of other bugs, including not being able to add back any instruments after saving & crashing on adding or selecting any measures. Rather than accounting for the other edge cases I believe it will be easier and more sensible to take the approach consistent with both the Instrument Dialog & MU3 of disallowing the user to interact with empty scores in general. The fix I suggest is to immediately force the user to add at least one instrument upon removing the last instrument through the side panel.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1219,6 +1219,10 @@ void Chord::read(XmlReader& e)
             e.unknown();
         }
     }
+    // Reset horizontal offset of grace notes when migrating from before 4.0
+    if (isGrace() && score()->mscVersion() < 400) {
+        rxoffset() = 0;
+    }
 }
 
 //---------------------------------------------------------


HemantAntony:
Also you should create a specific branch next time. Its better than using master

hodori0719:
What a mess! First time using Github desktop. :) Trying again now.

HemantAntony:
Cool! Next time you might not have to create a new PR. If you need help just say so :)

PR #12492, opened 2022-07-20
[MU4] Fix #12465: Force instrument selection upon removing last instrument

### LINKED ISSUE #12465: [MU4 Issue] Crash when using playback toolbar in a score with no instruments

hodori0719:
Resolves: https://github.com/musescore/MuseScore/issues/12465

Though the issue specifically describes crashes with the metronome feature on instrument-less scores, such scores are prone to an unspecified number of other bugs, including not being able to add back any instruments after saving & crashing on adding or selecting any measures. Rather than accounting for the other edge cases I believe it will be easier and more sensible to take the approach consistent with both the Instrument Dialog & MU3 of disallowing the user to interact with empty scores in general. The fix I suggest is to immediately force the user to add at least one instrument upon removing the last instrument through the side panel.

### DIFF
--- a/src/engraving/utests/barline_tests.cpp
+++ b/src/engraving/utests/barline_tests.cpp
@@ -42,7 +42,7 @@ static const String BARLINE_DATA_DIR(u"barline_data/");
 //   BarlineTests
 //---------------------------------------------------------
 
-class BarlineTests : public ::testing::Test
+class Engraving_BarlineTests : public ::testing::Test
 {
 public:
 };
@@ -68,7 +68,7 @@ static const double BRACKET0_HEIGHT_MAX     = 30;
 static const double BRACKET_HEIGHT_MIN      = 17;
 static const double BRACKET_HEIGHT_MAX      = 20;
 
-TEST_F(BarlineTests, barline01)
+TEST_F(Engraving_BarlineTests, barline01)
 {
     Score* score = ScoreRW::readScore(BARLINE_DATA_DIR + "barline01.mscx");
     EXPECT_TRUE(score);
@@ -112,7 +112,7 @@ TEST_F(BarlineTests, barline01)
 //
 //    NO REFERENCE SCORE IS USED.
 //---------------------------------------------------------
-TEST_F(BarlineTests, barline02)
+TEST_F(Engraving_BarlineTests, barline02)
 {
     Score* score = ScoreRW::readScore(BARLINE_DATA_DIR + "barline02.mscx");
     EXPECT_TRUE(score);
@@ -149,7 +149,7 @@ TEST_F(BarlineTests, barline02)
 //
 //   NO REFERENCE SCORE IS USED.
 //---------------------------------------------------------
-TEST_F(BarlineTests, barline03)
+TEST_F(Engraving_BarlineTests, barline03)
 {
     Score* score = ScoreRW::readScore(BARLINE_DATA_DIR + "barline03.mscx");
     EXPECT_TRUE(score);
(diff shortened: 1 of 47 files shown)


src/instrumentsscene/qml/MuseScore/InstrumentsScene/InstrumentsPanel.qml
```diff
@@ -117,7 +120,7 @@ Item {
             Layout.rightMargin: 20
 
             //: Text between %1 and %2 will be printed in a bold font.
-            text: qsTrc("instruments", "There are no instruments in your score. To choose some, press %1Add%2, or use the shortcut %1‘i’%2")
+            text: qsTrc("instruments", "There are no instruments in your score. Add at least one instrument.")
                   .arg("<b>").arg("</b>")
```
HemantAntony, line 124: You can delete this line. The `.arg()` replaces the `%1` with string passed as its parameter (so `<b>`)

hodori0719: Thanks, spotted

src/instrumentsscene/qml/MuseScore/InstrumentsScene/InstrumentsPanel.qml
```diff
@@ -106,6 +106,9 @@ Item {
 
             onRemovingRequested: {
                 instrumentsTreeModel.removeSelectedRows()
+                while (instrumentsTreeModel.isEmpty) {
```
RomanPudashkin, line 109: I don't think this is the right way to solve the problem. At the very least, this solution should be coordinated with the  design team (@Tantacrul or @bkunda)

bkunda: In principle, I think it's okay to immediately prompt the user to add instruments if they, for whatever reason, decide to delete every instrument in their score (via the Instruments panel). What I don't like is the persistently re-appearing "Add instruments" dialog whenever the "Cancel" button is clicked. Perhaps better would be to trigger a modal dialog whenever "Cancel" is clicked _in this specific context_, which reads something along the lines of "Please add at least one instrument to continue working in this score" (And a nice-to-have might then be a "Restore deleted instruments" button so the user can return to the stage they were at before this dialog was automatically re-triggered).  

In saying this, the solution presented here does leave open a loophole, which is where the user has more than one project open at once. Quitting MuseScore from another opened project allows you to save the "empty" one without having to add any new instruments. It is then possible to open the "empty" score, in which any interaction with the playback panel continues to trigger a crash. Fixing _this_ issue would then give rise to some further UX complications, like forcing the user to add some instruments before being able to save and close their session (which is becoming too messy from my POV). 

In light of this, if there is another technical solution that doesn't change any of the present UX, I'd say let's explore that first.

Tantacrul:
Just want to make sure of one thing. It is fine for a score to have no instruments. We specifically wanted to allow the ability for users to remove all scores if they want to. @bkunda

hodori0719:
Ah, I see. In that case, here is at least the list of issues I found: empty scores crash on interaction with playback after saving. Empty scores crash upon using "transpose," "note input"/N, or "select all"/ctrl-A. Empty scores arbitrarily crash while using "find"/ctrl-F. Frames persist and work after removing last instrument but disappear and are unaddable upon saving and reloading. Inserting measures at any point on an instrument-less score causes a crash. Inserting measures at start of score after adding an instrument also causes crash. Attempting to create parts causes crash.

There may very well be more, I only tested the most obvious ones. Most of these functionalities are impractical and could simply be disabled in an empty score, but the problem with frames should at least be fixed, since that seems to defeat the only purpose of scores like this. Also, the only way of "recovering" an empty score is by first adding an instrument and then adding measures to the end of the score (rather unintuitive), so this should also be fixed.

PR #12510, opened 2022-07-21
Ignore beam custom offset from before 3.0

mike-spa:
(because they used to be expressed in different coordinates and will mess up the score).
@oktophonie

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -1779,28 +1779,11 @@ void Beam::read(XmlReader& e)
             setGrowLeft(e.readDouble());
         } else if (tag == "growRight") {
             setGrowRight(e.readDouble());
-        } else if (tag == "y1") {
-            if (fragments.empty()) {
-                fragments.push_back(new BeamFragment);
-            }
-            BeamFragment* f = fragments.back();
-            int idx = (_direction == DirectionV::AUTO || _direction == DirectionV::DOWN) ? 0 : 1;
-            _userModified[idx] = true;
-            f->py1[idx] = e.readDouble() * _spatium;
-        } else if (tag == "y2") {
-            if (fragments.empty()) {
-                fragments.push_back(new BeamFragment);
-            }
-            BeamFragment* f = fragments.back();
-            int idx = (_direction == DirectionV::AUTO || _direction == DirectionV::DOWN) ? 0 : 1;
-            _userModified[idx] = true;
-            f->py2[idx] = e.readDouble() * _spatium;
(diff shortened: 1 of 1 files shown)


src/engraving/libmscore/beam.cpp
```diff
@@ -1779,15 +1779,15 @@ void Beam::read(XmlReader& e)
             setGrowLeft(e.readDouble());
         } else if (tag == "growRight") {
             setGrowRight(e.readDouble());
-        } else if (tag == "y1") {
+        } else if (tag == "y1" && score()->mscVersion() >= 300) {
```
cbjeukendrup, line 1782: The following would be slightly more efficient (and maybe more beautiful code):
```c++
} else if (tag == "y1") {
    if (score()->mscVersion() < 300) {
        // before 3.0, custom offset used to be expressed in different coordinate systems, 
        // and thus would mess up the score
        e.skipCurrentElement();
    } else {
       if (fragments.empty()) {
            fragments.push_back(new BeamFragment);
       }
       BeamFragment* f = fragments.back();
       ...
```
Because that way you always stop comparing the tag if the it matches "y1". Otherwise, you do a lot of unnecessary comparisons in the else-ifs after "y1".

src/engraving/libmscore/beam.cpp
```diff
@@ -1801,7 +1801,7 @@ void Beam::read(XmlReader& e)
             _userModified[idx] = true;
             double _spatium1 = spatium();
 
-            while (e.readNextStartElement()) {
+            while (e.readNextStartElement() && score()->mscVersion() >= 300) {
```
cbjeukendrup, line 1804: Here, for < 3.0 scores, you do make the beam fragment userModified, but you don't set the py1/py2 values, so those will be zero. Doesn't that mean that the beam will always be flat, straddling the topmost staff line? Maybe better to not make it userModified in this case:
```c++
        } else if (tag == "Fragment") {
            if (score()->mscVersion() < 300) {
                e.skipCurrentElement();
            } else {
                BeamFragment* f = new BeamFragment;
                int idx = (_direction == DirectionV::AUTO || _direction == DirectionV::DOWN) ? 0 : 1;
                _userModified[idx] = true;
                double _spatium1 = spatium();

                while (e.readNextStartElement()) { 
                    ...

cbjeukendrup: That might also fix the vtests...

mike-spa:
@cbjeukendrup always spot on! Thank you!

cbjeukendrup (approved):

PR #12513, opened 2022-07-22
Fix #9697: align opened scores in the middle instead of the top left

### LINKED ISSUE #9697: Score page should be aligned to the middle when opened
**Describe the bug**
I'm not sure if this issue has been reported yet.
Currently, the score is being aligned top left corner when opened/created new score. It should be aligned in the middle as MS3 does.

**Expected behavior**

image

**Screenshots**

image

 - OS: Windows, MacOS

martin-danhier:
Resolves: #9697

Hello,

I first tried to fix this issue by simply centering the view horizontally with a small margin from the top vertically, but I didn't find the result very helpful for long scores with more than one page or continuous views. I thus thought it made sense to make the choice of the default position depend on the view mode and the number of pages.

Here is what it looks like so far:

## Page mode, 1 page

Center horizontally, and align top with a small margin.

## Page mode, 2 pages or more

It aligns the score on the top left with a small margin.

I initially thought to center the first page, but I think the space is better used by showing more of the score.

## Horizontal continuous mode

Center vertically, and align left with a small margin.

## Vertical continuous mode

Same as 1 page. This is the fallback alignment.

## Switching tabs

I also noticed that the score was being moved to the top left each time one of the "Score"/"Publish"/... tabs was changed.

With a simple fix, it can stay as it was in the previous tab (in the code, it feeled like an unwanted side effect). Though, we might want to reset the default position instead of keeping the previous one?

I included the fix, but idk if I should create a separate issue?

Let me know what you think :-)

### DIFF
--- a/src/notation/view/abstractnotationpaintview.cpp
+++ b/src/notation/view/abstractnotationpaintview.cpp
@@ -119,6 +119,35 @@ void AbstractNotationPaintView::initNavigatorOrientation()
     });
 }
 
+void AbstractNotationPaintView::initPosition()
+{
+    const PointF center        = this->canvasCenter();
+    const ViewMode currentMode = notation()->viewMode();
+    constexpr float margin     = 150.0;
+
+    // Multiple pages (default mode)
+    if (currentMode == ViewMode::PAGE) {
+        PageList pages = notation()->elements()->pages();
+
+        // More than 1 page: align it in the top-left with margin
+        if (pages.size() > 1) {
+            doMoveCanvas(margin, margin);
+        }
+        // 1 page: center it
+        else {
+            doMoveCanvas(center.x(), margin);
+        }
+    }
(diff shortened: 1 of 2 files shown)


martin-danhier:
Hello, I rebased the changes to the latest version of master, and it still works.

There is just a slight problem when:
1. the score is loaded in page mode with several parts opened
2. the main score is changed to continuous mode
3. a part is opened: it will center horizontally as if it was in page mode, even though it is in continuous mode.

However, this is fixed with the changes from my other PR for independent view modes #12618. Together, the two PR work correctly for independent and centered views.

cbjeukendrup:
Hello! I know it's "a bit" late, but thanks for this contribution! Apologies that we didn't comment on it earlier. We could not merge it directly, but a variation on your implementation has been merged via https://github.com/musescore/MuseScore/pull/24483 and will be part of the upcoming release. So thanks again!

Issue #12504, opened 2022-07-21
[MU4 Issue] View/Zoom drop-down in status bar doesn't show


DmitryArefiev:
![Untitled](https://user-images.githubusercontent.com/90187801/180221260-e97da556-1ac8-4186-a9c9-9c56eff842f0.png)

https://user-images.githubusercontent.com/90187801/180221728-9de0c3c1-6a6a-4f02-bab8-9e1e7efc6bf9.mp4

**Screenshots**

**Platform information**
 macOS, Windows, Linux

DmitryArefiev:
Could be related to splash screen implementation

PR #11939, opened 2022-06-08
fixed #11935: Disabled controls after closing score

Eism:
Resolves: #11935

PR #10757, opened 2022-03-06
Implemented independent zooming and scrolling per notation

cbjeukendrup:
Resolves: #7879

PR #12515, opened 2022-07-22
Fix the zoom dropdown in Status Bar

cbjeukendrup:
Resolves: #12504 

Looks like it was caused by #11939.


Eism:
@cbjeukendrup Nope!
It's not caused by my PR, it's caused by your PR #10757

You can send me a bottle of wine by way of an apology

cbjeukendrup:
Oops! Of course... 
That PR has been open for so long, and then it was suddenly merged. I should probably do some investigation whether it hasn't caused more unexpected bugs.

Eism (approved):

PR #12521, opened 2022-07-22
Fixing stem direction glitch for custom-positioned cross-staff beams

mike-spa:
Fixing a glitch related to stems and custom position of cross staff beams.
@oktophonie @asattely 
Before:
image
After:
image


cbjeukendrup:
@mike-spa I found this case where _non_-customized beams also produce this glitch:
Schermafbeelding 2022-07-22 om 18 39 05
(the notes belong to the center staff, one is moved down, one is moved up)

cbjeukendrup:
Hm... it seems that we're not yet entirely ready to handle these strange "both-directions-cross-staff" beams...
Schermafbeelding 2022-07-22 om 18 57 39
Schermafbeelding 2022-07-22 om 18 50 54

mike-spa:
@cbjeukendrup good catch! Last commit should have fixed it, I believe.

First bar: custom position. Second bar: default.
image

oktophonie (approved):

cbjeukendrup:
Found another one. Enter notes like this:
Schermafbeelding 2022-07-25 om 20 30 58
and then set the stem direction of one of the notes explicitly to "up" or "down". The result is something similar to this:
Schermafbeelding 2022-07-25 om 20 32 53

oktophonie:
MU3.6 just ignores modified stem directions in cross-stave beams, so we should just continue to do the same - no doubt there's a cleverer solution that could be considered for the future, though that might be a bit of a logic minefield.

mike-spa:
There you go, the last commit simply disallows setting a custom stem/beam direction with cross staff beams, which is what happens in 3.6

src/engraving/libmscore/chord.cpp
```diff
@@ -1007,19 +1007,21 @@ void Chord::computeUp()
...
+                    if (move > 0) {
+                        _up = staffMove() > 0;
+                        firstCr->setUp(firstCr->staffMove() > 0);
+                        lastCr->setUp(lastCr->staffMove() > 0);
+                    } else {
+                        _up = staffMove() >= 0;
+                        firstCr->setUp(firstCr->staffMove() >= 0);
+                        lastCr->setUp(lastCr->staffMove() >= 0);
+                    }
+                    break;
```
cbjeukendrup, line 1023: Couldn't the `break` be moved out of the `if (!_beam->userModified()) {` block?

mike-spa: I did, at some point, then I somehow lost it. Done now!

cbjeukendrup (approved):

RomanPudashkin:
It would be great to add some tests to protect this fix from regressions

mike-spa:
> It would be great to add some tests to protect this fix from regressions

@RomanPudashkin done! Again, no idea what is going on with the mac build (my other PR seems to be failing on windows). Maybe something is going on in the master

PR #12524, opened 2022-07-23
fix #12481/#12480 - fixes for styled properties in inspector

### LINKED ISSUE #12481: [MU4 Issue] Inspector and other things don't update when a style value is changed by undo/redo
**Describe the bug**
If a style value is changed not directly, but via an undo or redo action, the `styleChanged` notification is not emitted. Thus, certain objects in the UI that rely on this notification won't update properly in this situation. For example, the flag type selector in the "Stem" Inspector.

**To Reproduce**
1. select a note, and go to the "stem" inspector
2. open the Style dialog, go to the "Notes" and change the flag type
3. notice that the Inspector correctly updates now, because a style setting was changed directly
4. Press OK
5. Now, use the "undo" button or press Ctrl/Cmd+Z. The score updates properly (because that happens inside the engraving module), but the Inspector not (because that should have happened via the `styleChanged` notification from the notation module, but that notification is not emitted this time)

Or another way: **THIS IS LOGGED SEPARATELY** #13437
1. have a score with transposing instruments (not necessary, but useful for checking behaviour)
2. enable concert pitch using the checkbox in the Status bar
3. press undo
Expected: concert pitch is disabled again, and the concert pitch checkbox reflects this
Actual: although the score is updated correctly (because that happens inside the engraving module) but the checkbox not (because that should have happened via the `styleChanged` notification from the notation module, but that notification is not emitted this time)
If you click the checkbox again after this, unexpected things may happen.

In both cases:
Afterwards, if you do anything else that causes the `styleChanged` notification, the state of the UI is (of course) updated/fixed again.

**Expected behavior**

**Screenshots**

https://user-images.githubusercontent.com/48658420/179816948-1df456df-d039-48b4-9b09-354e691bce3c.mov

**Platform information**
 - OS: reported on macOS

**Additional context**
We can use a stopgap solution by listening for `undoStack()->stackChanged()` or `notationChanged()` notifications instead (or additionally). Or by emitting the `styleChanged` notification whenever one of these notifications is emitted.

But that will cause _a lot_ of unnecessary notifications and updates. A real fix might be not so easy; it would require a smart solution for communication between the engraving module and the notation module. 

It is not certain that this kind of problem is limited to the `styleChanged` notification. Other things may have similar problems in combination with "undo", although I can't think of any right now.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/12481)

### DIFF
--- a/src/inspector/models/abstractinspectormodel.cpp
+++ b/src/inspector/models/abstractinspectormodel.cpp
@@ -278,11 +278,12 @@ void AbstractInspectorModel::onPropertyValueChanged(const mu::engraving::Pid pid
 
         mu::engraving::PropertyFlags ps = element->propertyFlags(pid);
 
-        if (ps == mu::engraving::PropertyFlags::STYLED) {
-            ps = mu::engraving::PropertyFlags::UNSTYLED;
+        PropertyValue propValue = valueToElementUnits(pid, newValue, element);
+        if (ps != mu::engraving::PropertyFlags::NOSTYLE) {
+            ps = (propValue == element->propertyDefault(pid)) ? mu::engraving::PropertyFlags::STYLED
+                 : mu::engraving::PropertyFlags::UNSTYLED;
         }
 
-        PropertyValue propValue = valueToElementUnits(pid, newValue, element);
         element->undoChangeProperty(pid, propValue, ps);
     }
 
--- a/src/inspector/models/notation/notes/stems/stemsettingsmodel.cpp
+++ b/src/inspector/models/notation/notes/stems/stemsettingsmodel.cpp
@@ -62,6 +62,10 @@ void StemSettingsModel::createProperties()
     context()->currentNotation()->style()->styleChanged().onNotify(this, [this]() {
         emit useStraightNoteFlagsChanged();
     });
+
+    context()->currentNotation()->undoStack()->stackChanged().onNotify(this, [this]() {
+        emit useStraightNoteFlagsChanged();
+    });
 }
 
 void StemSettingsModel::requestElements()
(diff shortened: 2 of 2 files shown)


src/inspector/models/abstractinspectormodel.cpp
```diff
@@ -278,11 +278,12 @@ void AbstractInspectorModel::onPropertyValueChanged(const mu::engraving::Pid pid
 
         mu::engraving::PropertyFlags ps = element->propertyFlags(pid);
 
-        if (ps == mu::engraving::PropertyFlags::STYLED) {
-            ps = mu::engraving::PropertyFlags::UNSTYLED;
+        PropertyValue propValue = valueToElementUnits(pid, newValue, element);
+        if (ps != mu::engraving::PropertyFlags::NOSTYLE) {
+            ps = (propValue == element->propertyDefault(pid)) ? mu::engraving::PropertyFlags::STYLED
+                 : mu::engraving::PropertyFlags::UNSTYLED;
```
wizofaus, line 284: Ensure that when resetting a property value back to the default it gets the "styled" flag, so it continues to track that default.

src/inspector/models/notation/notes/stems/stemsettingsmodel.cpp
```diff
@@ -60,7 +60,11 @@ void StemSettingsModel::createProperties()
     });
 
     context()->currentNotation()->style()->styleChanged().onNotify(this, [this]() {
-        emit useStraightNoteFlagsChanged();
+        loadProperties();
```
wizofaus, line 63: this might be needed in other inspector models too

wizofaus:
Closing this as the main issue has already been fixed, though there's still an issue with "reset to default".

PR #12533, opened 2022-07-23
Audit translatable strings. Step 8

cbjeukendrup:
See #12389 

(After this one, there will be one more step, but that one will be so exciting that it deserves a separate PR.)

### DIFF
--- a/src/appshell/qml/DevTools/MPE/ArticulationsProfileEditorView.qml
+++ b/src/appshell/qml/DevTools/MPE/ArticulationsProfileEditorView.qml
@@ -63,7 +63,8 @@ Rectangle {
 
                     width: parent.width - openButton.width - saveButton.width - createButton.width
 
-                    hint: qsTrc("mpe", "Profile path")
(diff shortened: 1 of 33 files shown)


src/appshell/qml/DevTools/MPE/ArticulationsProfileEditorView.qml
```diff
@@ -63,7 +63,7 @@ Rectangle {
 
                     width: parent.width - openButton.width - saveButton.width - createButton.width
 
-                    hint: qsTrc("mpe", "Profile path")
+                    hint: "Profile path" //qsTrc("mpe", "Profile path")
```
igorkorsukov, line 66: I think, we can remove comments

cbjeukendrup: I decided to keep them, but only in a simpler form, to make them less obstructive. When we are going to expose this tool to users, we can restore translatability with a simple RegEx search & replace, using these comments.

PR #12540, opened 2022-07-24
Flexible translation of instrument names

shoogle:
Give translators more control over how instrument names will appear in other languages. Add translation comments to make their job easier.

### DIFF
--- a/src/notation/internal/notationparts.cpp
+++ b/src/notation/internal/notationparts.cpp
@@ -37,15 +37,39 @@ using namespace mu::engraving;
 
 static const mu::engraving::Fraction DEFAULT_TICK = mu::engraving::Fraction(0, 1);
 
-static QString formatInstrumentTitleOnScore(const QString& instrumentName, const Trait& trait, int instrumentNumber)
+static QString formatInstrumentTitleOnScore(const QString& instrumentName, const Trait& trait)
 {
-    QString numberPart = instrumentNumber > 0 ? " " + QString::number(instrumentNumber) : QString();
+    // Comments for translators start with //:
 
-    if (trait.type != TraitType::Transposition || trait.isHiddenOnScore) {
-        return instrumentName + numberPart;
+    if (trait.type == TraitType::Transposition && !trait.isHiddenOnScore) {
+        //: %1=name ("Horn"), %2=transposition ("C alto"). Example: "Horn in C alto"
+        return mu::qtrc("notation", "%1 in %2", "Transposing instrument displayed in the score")
+               .arg(instrumentName, trait.name);
     }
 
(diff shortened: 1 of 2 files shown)


src/notation/internal/notationparts.cpp
```diff
@@ -37,15 +37,39 @@ using namespace mu::engraving;
 
 static const mu::engraving::Fraction DEFAULT_TICK = mu::engraving::Fraction(0, 1);
 
+static QString formatInstrumentTitleOnScore(const QString& instrumentName, const Trait& trait)
+{
+    // Comments for translators start with //:
+
+    if (trait.type == TraitType::Transposition && !trait.isHiddenOnScore) {
+        //: %1=name ("Horn"), %2=transposition ("C alto"). Example: "Horn in C alto"
+        return mu::qtrc("notation", "%1 in %2", "Transposing instrument displayed in the score")
```
cbjeukendrup, line 46: Keep in mind that the "disambiguation" needs to be repeated exactly in all places where we want to use the same translatable string. Therefore, in general, I don't consider it a good place to add long comments, rather for a few simple words that are not prone to typos. 
However, since it is very unlikely that we will ever use these specific strings in other places, I'd say this is okay for now.

cbjeukendrup (approved):
Looks good to me! Two comments that don't really require action in my opinion but might be good to discuss.

src/notation/internal/notationparts.cpp
```diff
@@ -37,15 +37,39 @@ using namespace mu::engraving;
 
 static const mu::engraving::Fraction DEFAULT_TICK = mu::engraving::Fraction(0, 1);
 
+static QString formatInstrumentTitleOnScore(const QString& instrumentName, const Trait& trait)
+{
+    // Comments for translators start with //:
```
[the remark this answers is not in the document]
shoogle: I think it is sensible to _not_ say what `//:` does except in a few key places where it is useful and doesn't lead to unnecessary bloat. I think this is one of those places where it is worth saying because these format strings are vitally important and the comments here really are essential.

src/notation/internal/notationparts.cpp
shoogle: I think a shorter disambiguation string would require a longer translation comment to explain what it means. I'll experiment with `lupdate` to see whether/how it handles multi-line translation comments.

cbjeukendrup: My experience is that `lupdate` strips newlines from `//:` comments. 

Anyway, since each of these strings is used in only one place for now, I'm happy with it as it is, but it's just something to keep in mind for other cases.

PR #12541, opened 2022-07-24
Fix #333774: MusicXML import of slashed grace notes

lvinken:
Resolves: https://musescore.org/en/node/333774

MuseScore's current (3.6.2 and 4.0) MusicXML importer emulates MuseScore's GUI (see the grace note palette) and supports only 1/8 grace notes with slash. Grace notes with slash and durations 1/4, 1/16 or 1/32 are imported as non-slashed. Given that the GUI also allows changing an acciaccatura's duration (e.g. from 1/8 to 1/16th) while keeping the slash, the importer should also support this.


RomanPudashkin (approved):

shoogle (approved):

shoogle:
Works great, thanks!

Issue #11976, opened 2022-06-10
[MU4 Issue] "Echo"-like effect on certain instruments 


bkunda:
**Describe the bug**
This is an issue I've so far only encountered with some of the "auxiliary percussion" instruments in templates like "European concert band" and "Small pit percussion".

@Tantacrul has apparently heard this in more places, so we may expand this issue.

The following instruments within "auxiliary percussion" sound with a quasi echo/delay effect.

- Closed hi-hat
- Pedal hi-hat
- Snare side-stick
- Concert snare drum

See this video to compare MS3 with MS4 behaviour: 

https://user-images.githubusercontent.com/86290556/173083015-0377dc4a-b541-4aad-b802-67510330fd9e.mov

My guess is this is an artefact of the reverb.

This is a problem because it can aurally convey a rhythmic idea that differs from what the user notated.

**Expected behavior**
Playback sound should be clean with no aural impression of echo/delay

**Screenshots**

**Platform information**
 - Tested on MacOS 12.3.1 (M1)

rgreen5:
This issue would be much easier to test if  #11234 could be made a priority fix; at the moment you can't play any instruments, even from the virtual piano keyboard, without entering notation as well. The reverb unit also needs to be accessible so testers can see and change the reverb properties (predelay, amount, etc).

Issue #11974, opened 2022-06-10
[MU4 Issue] Remaining mapping for percussion sounds


bkunda:
**Describe the bug**
The following unpitched percussion instruments still have incorrect sound mapping: 

- Chains
- Temple blocks
- Flexatone

And this tuned percussion instrument has no sound:

- Tuned klaxon horns

Instructions for correct mapping are included in [this spreadsheet](https://docs.google.com/spreadsheets/d/1p544iRhT_kGtkQd0u2qdgEoKb0ZHggvUg8FARhhZNMY/edit?usp=sharing).

**Platform information**
 - Tested on MacOS 12.3.1 (M1)

Issue #11011, opened 2022-03-31
[MU4 Issue] Reverb effect is not applied to the piano and some percussion instruments (tambourine)


boryskuzmenko:
**To Reproduce**
1. Create a score with woodwind, strings, brass, piano and tambourine instruments
2. Create some notes for each instrument
3. Click the notes
4. Pay attention to the sound of piano and tambourine 

**Expected behavior**
Reverb effect is applied to all instruments in score

**Screenshots**

https://user-images.githubusercontent.com/101250347/161028419-dd51c330-20ac-4767-89a2-e62014a2d888.mp4

**Platform information**
 - OS: [macOS, Windows]

Tantacrul:
Also, some have far too much reverb, as noted here (https://github.com/musescore/MuseScore/issues/11976)
 @vpereverzev - when we get around to this task, I can quickly test with you to get to a reasonable result.

Tantacrul:
https://github.com/musescore/MuseScore/issues/11976

boryskuzmenko:
🟢 **FIXED** in https://github.com/musescore/MuseScore/pull/12545

PR #12545, opened 2022-07-25
reverb_problems

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11011
Resolves: https://github.com/musescore/MuseScore/issues/11976
Resolves: https://github.com/musescore/MuseScore/issues/11974


bkunda:
@RomanPudashkin Tested this draft PR on MacOS and Windows.

✅ Reverb settings appear correct for the instruments mentioned in #11011 and #11976

🟠 We should still address some missing reverb on other instruments in the **drumset** instrument group (bass drums and toms), as it's quite clear that they don't have any reverb compared to other sounds in the group. 

Also instruments mentioned in #11974 appear to work now as per MS3.6, so any other changes we want to make to (E.g. assigning the **flexatone** a single-line staff) could perhaps be addressed in a separate issue as they are improvements on MS3.6 and not fixes.

vpereverzev (approved):

PR #12551, opened 2022-07-25
Few fixes and optimizations for PR #12414

VanSHOE:
This involves any further fixes required for the PR #12414 

Progress:
- [x] Optimize ActionText
- [x] Default Context is now `any` (Hotfix. This might be changed later in a separate PR)

I will continue to add resolutions to any more issues found to this PR

### DIFF
--- a/src/framework/shortcuts/view/shortcutsmodel.cpp
+++ b/src/framework/shortcuts/view/shortcutsmodel.cpp
@@ -79,7 +79,8 @@ const QString& ShortcutsModel::actionDescription(const std::string& actionCode) 
 
 const QString& ShortcutsModel::actionText(const std::string& actionCode) const
 {
-    return actionDescription(actionCode) == "" ? actionTitle(actionCode) : actionDescription(actionCode);
+    const QString description = actionDescription(actionCode);
+    return description.isEmpty() ? actionTitle(actionCode) : description;
 }
 
 int ShortcutsModel::rowCount(const QModelIndex&) const
--- a/src/framework/ui/uitypes.h
+++ b/src/framework/ui/uitypes.h
@@ -190,7 +190,7 @@ struct UiAction
 {
     actions::ActionCode code;
     UiContext uiCtx = UiCtxAny;
-    std::string scCtx = "NULL";
+    std::string scCtx = "any";
     QString title;
     QString description;
     IconCode::Code iconCode = IconCode::Code::NONE;


src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -253,7 +253,14 @@ bool ShortcutsRegister::readFromFile(ShortcutList& shortcuts, const io::path_t&
 
         Shortcut shortcut = readShortcut(reader);
         if (shortcut.isValid()) {
-            shortcuts.push_back(shortcut);
+            if (uiactionsRegister()->action(shortcut.action).isValid()) {
```
Eism, line 256: This code should not be here.
The ShortcutsRegister::readFromFile method should only read data from the file, it should not filter.

Let's just fix the crash. Let's make a hot fix - the default context is any. I hope this doesn't create other critical bugs. At the moment it is a blocking critical bug. Needs to be fixed quickly

And then you'll figure out the problem in another PR

VanSHOE: @Eism Please have a look. Just pushed it

Eism (approved):

src/framework/shortcuts/internal/shortcutsregister.cpp
shoogle: @Eism, the cause of the crash is that these shortcuts in shortcuts.xml don't have a corresponding UiAction in the code:
```cpp
play-next-chord
play-prev-measure
note-breve-TAB
symbols
backspace
viewmode
key-signatures
play-prev-chord
zoom-in-horiz-pre
realtime-advance
next-score
revision
accessibility-dev-show-tree
seek-begin
time-signatures
seek-end
rest-TAB
zoom-in-vert-pre
previous-score
note-longa-TAB
text-word-left
diagnostics-show
play-next-measure
text-word-right
zoom-out-horiz-pre
zoom-out-vert-pre
plugin-creator
startcenter
```

The filtering was dropping any shortcuts that don't have an action, such as these ones above, or others if the user makes a typo in one of the `<key>` tags in shortcuts.xml.

@bkunda is going through this list to see if there are any that we can simply remove from shortcuts.xml because we don't need them anymore, though it still seems like it might be good to filter shortcuts without an action when the file is read in case there are typos. Filtering them will free up the shortcut key so it can be assigned to other things when the user tries to define a new shortcut key in Preferences.

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -79,7 +79,8 @@ const QString& ShortcutsModel::actionDescription(const std::string& actionCode)
 
 const QString& ShortcutsModel::actionText(const std::string& actionCode) const
 {
-    return actionDescription(actionCode) == "" ? actionTitle(actionCode) : actionDescription(actionCode);
+    const QString description = actionDescription(actionCode);
+    return description.isEmpty() ? actionTitle(actionCode) : description;
```
cbjeukendrup, line 83: This produces a compiler warning: you are returning a reference to `description`, which is a temporary object and will be deleted at the bottom of this method. So the reference ends up pointing to a deleted object. 

Correct would be:
```suggestion
    const QString& description = actionDescription(actionCode);
    return description.isEmpty() ? actionTitle(actionCode) : description;
```
This does work, because `actionDescription` also returns a reference, so you basically forward that reference this way. 

Also, this method is still not quite efficient: both `actionTitle` and `actionDescription` need to look up the action in the actions register. It would be better to look up the action once, then look at its title/description members directly.

Anyway, this method will be overhauled in my next PR, so this is just for information.

VanSHOE: I see... Thank you for letting me know of this issue :)

PR #12553, opened 2022-07-26
Generate placeholder translations

shoogle:
The use of generated translations enables developers to see which strings have been correctly marked for translation **without having to wait for a proper translation to be made available**.

Fake translations are displayed in the UI as source text in double angle quotes `« »` preceded by a superscript `tr`:

image

A lack of a fake translation indicates a place where translations are not being used correctly in the code.

image

This shows that the "Parts" and "Mixer" button text is not being translated. Nor are the tooltips for toolbar buttons like "Double sharp".

---

Special markup is used to identify plural translations:

    Source text:                       %n measure(s)
    Translation (1st plural form): ᵗʳ¹«%n measure(s)»
    Translation (2nd plural form): ᵗʳ²«%n measure(s)»
    Etc.

See the second page of the New Score dialog for an example of plurals.

image&nbsp;image

---

Replacement markers %1, %2, etc. for QString arg() are also identified:

    Source text:     Add %1 to chord
    Translation: ᵗʳ«Add ⌜%1⌝ to chord»

Arguments must be translated separately to the main string for translation to work properly:

    Non-translated argument:    ᵗʳ«Add ⌜D⌝ to chord»
    Translated argument:     ᵗʳ«Add ⌜ᵗʳ«D»⌝ to chord»

image

Here the format strings are translated but the instrument data (name and traitName) is not being translated.

---
The Python script could be modified to display more information in the translated strings such as context, disambiguation, comments or file name and line number of the string in the code or in the .ts file.

### DIFF
--- a/.github/workflows/ci_linux_mu4.yml
+++ b/.github/workflows/ci_linux_mu4.yml
@@ -73,6 +73,13 @@ jobs:
             fi 
         fi
 
+        DO_PLACEHOLDER_TRANSLATIONS='false'
+        if [[ "$DO_BUILD" == "true" ]]; then
+          if [[ "$BUILD_MODE" == "nightly_build" || "$BUILD_MODE" == "devel_build" ]]; then
+            DO_PLACEHOLDER_TRANSLATIONS='true'
+          fi
+        fi
+
         DO_UPLOAD_SYMBOLS='false'
         SENTRY_PROJECT=${{ github.event.inputs.sentry_project }}
         SENTRY_URL=""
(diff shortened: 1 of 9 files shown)


tools/translations/placeholder_en_GB.py
```diff
@@ -0,0 +1,80 @@
...
+# to check that translations are being used correctly. This is useful when
+# proper translations are not yet available for certain strings.
+
+# Steps:
+#   1. Add Qt's bin folder to $PATH
+#   2. Call run_lupdate.sh
+#   3. Run this script
+#   4. Call run_lrelease.sh
+#   5. Compile & run MuseScore
+#   6. In Preferences > General, set language to en_GB
```
shoogle, line 14: Instructions for how to use the Python script

igorkorsukov:
Cool idea!!
Maybe add a separate language? 
Select it automatically for devel builds
And run this script when building on CI

cbjeukendrup:
(FWIW: Many of those cases where we forget to translate strings, will already be fixed in my next PR)

igorkorsukov (approved):

PR #12560, opened 2022-07-26
Fix #12506: check other staves for keysigs when deleting measures

### LINKED ISSUE #12506: [MU4 Issue] Deleting first measure in score removes Key Signatures in transposing instruments when score is not in concert pitch
**To Reproduce**
1. Open Wind quintet template
2. Make sure the concert pitch is unchecked
3. Delete the first measure (Ctr/Cmd+Delete) - **key signatures in Clarinet and Horn should not be removed**

https://user-images.githubusercontent.com/90187801/180238020-223f5ac4-6a10-4369-83d9-8c1de21662eb.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Also occurs in **MU3**

from @sammik :
1) occurs only in C major
2) and only if Key signature in first measure wasnt changed before removing first measure
(if you change initial ks to G major and than to C major, bug not there)

martin-danhier:
Resolves: #12506

Hello,

This issue was caused by this chain of events:
- When creating a score from a template, no key signature is added for staves in C major because it is empty
- Then, when disabling concert pitch, only transposing instruments are updated, so only two key signatures are added in a wind quintet, in the third and fourth staves
- When deleting the first measures, the function looks at the first staff to see if a key signature is deleted, in order to get the key and add it if necessary in the first measure after the deleted ones. Since there is no key signature in the first staff in C major, it assumes that there is no key change and doesn't add it afterwards.

I managed to fix it with two fixes:
- I updated the code that creates new scores from templates to add key signatures even if the key signature is "empty"
- I updated the code that checks for key signatures in the deleted measures to look in other staves if nothing was found in the first one (for backward compatibility with scores that were created before the other fix)

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -2767,17 +2767,26 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
             }
             sts = m->findSegment(SegmentType::KeySig, m->tick());
             if (sts && !lastDeletedKeySig) {
-                lastDeletedKeySig = toKeySig(sts->element(0));
-                if (lastDeletedKeySig) {
-                    lastDeletedKeySigEvent = lastDeletedKeySig->keySigEvent();
-                    if (!styleB(Sid::concertPitch) && !lastDeletedKeySigEvent.isAtonal()) {
-                        // convert to concert pitch
-                        transposeKeySigEvent = true;
-                        Interval v = staff(0)->part()->instrument(m->tick())->transpose();
-                        if (!v.isZero()) {
-                            lastDeletedKeySigEvent.setKey(transposeKey(lastDeletedKeySigEvent.key(), v,
-                                                                       lastDeletedKeySig->part()->preferSharpFlat()));
+                // Search all staves for a key sig until one is found
+                // Sometimes, there is no key sig in a staff, but there is a transposing instrument
+                // that has one in another one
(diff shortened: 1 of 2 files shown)


src/engraving/libmscore/edit.cpp
```diff
@@ -2767,17 +2767,26 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
...
-                if (lastDeletedKeySig) {
-                    lastDeletedKeySigEvent = lastDeletedKeySig->keySigEvent();
-                    if (!styleB(Sid::concertPitch) && !lastDeletedKeySigEvent.isAtonal()) {
-                        // convert to concert pitch
-                        transposeKeySigEvent = true;
-                        Interval v = staff(0)->part()->instrument(m->tick())->transpose();
-                        if (!v.isZero()) {
-                            lastDeletedKeySigEvent.setKey(transposeKey(lastDeletedKeySigEvent.key(), v,
-                                                                       lastDeletedKeySig->part()->preferSharpFlat()));
+                // Search all staves for a key sig until one is found
```
mike-spa, line 2770: Just for my understanding: now that you've modified the other piece of code to always create the keySig, why is this search still necessary?

martin-danhier: Hello, if I remember correctly, it was for backwards compatibility with scores created before the fix.

RomanPudashkin:
Closing since the original issue has been fixed

PR #12561, opened 2022-07-26
Reorganizing stacking order of some score elements

mike-spa:
Disclaimer: in some of these cases there may be more than one "correct" solution, depending on context. However, part of the current order can be definitely rationalized. Codewise, it is just about changing the order in which the layouts are called, I haven't actually added anything.

In detail:
- Fermatas must be inside of dynamics, hairpins, and text lines.
- Tempo should be outside of everything, including guitar diagrams and voltas.
- Tempo change lines should be at the same stacking position as tempo.
- Rehearsal marks are the only items that should go outside of tempo, and the "segno" mark should be at the same stacking position as rehearsal marks.

@oktophonie 

Before:
before
After:
after


mike-spa:
vtests look good to me!

oktophonie (approved):

RomanPudashkin (approved):

PR #12562, opened 2022-07-26
Slur: correction when cross-system

mike-spa:
A correction to the slur collision-avoidance logic. Otherwise a piece of the slur may try to avoid collisions also with items that are not in the same system. @oktophonie 
slurIssue

### DIFF
--- a/src/engraving/libmscore/slur.cpp
+++ b/src/engraving/libmscore/slur.cpp
@@ -1444,6 +1444,11 @@ SpannerSegment* Slur::layoutSystem(System* system)
             }
             Chord* c1 = startCR()->isChord() ? toChord(startCR()) : 0;
             Chord* c2 = endCR()->isChord() ? toChord(endCR()) : 0;
+            if (startCR()->measure()->system() != endCR()->measure()->system()) {
+                // If the end chord is in a different system its direction may
+                // have never been computed, so we need to compute it here.
+                c2->computeUp();
+            }
 
             if (_sourceStemArrangement != -1) {
                 if (_sourceStemArrangement != calcStemArrangement(c1, c2)) {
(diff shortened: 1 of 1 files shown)


cbjeukendrup:
Found an issue:
Schermafbeelding 2022-07-26 om 19 56 57
(Note the collision in bar 20).

In the second system, the first segment that contains a ChordRest in voice 2 is in the second measure, so the first measure won't be looked at.
Probably, you need to replace `system()->firstChordRest(startCR->track())` with something else that just gets the first ChordRest segment of the system, not taking into account the voice of `startCR`. It looks like you'll need to create a new method in the `System` class for this. (I believe you can't just use the very first segment, because I believe that will include the system header.)

mike-spa:
> In the second system, the first segment that contains a ChordRest in voice 2 is in the second measure, so the first measure won't be looked at. Probably, you need to replace `system()->firstChordRest(startCR->track())` with something else that just gets the first ChordRest segment of the system, not taking into account the voice of `startCR`. It looks like you'll need to create a new method in the `System` class for this. (I believe you can't just use the very first segment, because I believe that will include the system header.)

Incredibly, I had thought about this one 😂 Indeed, it would be nice to have some System::firstChordRestSegment() and System::lastChordRestSegment() methods, but I wasn't sure it would be worth adding two new methods just for this. I mean, as you showed, the only way that this becomes a problem is if you slur over a full bar rest in one of the voices, which is something one shouldn't do. But regardless, those methods may be useful elsewhere, so I may just do them.

On a separate note: testing the kind situation you posted there reveald a completely separate glitch. If you close and open that file, you'll probably see the slur picking the wrong direction. I am about to push a new commit to fix that.

mike-spa:
@cbjeukendrup turns out it wasn't necessary to implement those two methods. I think the current solution I've pushed fixed both the anti-collidion glitch and the wrong direction glitch.

oktophonie (approved):

oktophonie (approved):

src/engraving/libmscore/slur.cpp
```diff
@@ -377,10 +377,37 @@ void SlurSegment::avoidCollisions(PointF& pp1, PointF& p2, PointF& p3, PointF& p
...
+    if (isSingleEndType()) {
+        endSeg = endCR ? endCR->segment() : nullptr; // last of the slur
+    } else {
+        endSeg = system()->lastMeasure()->last(); // last of the system
+    }
+
+    // Collect all the segments spanned by this slur segment in a single vector
+    std::vector<Segment*> segList;
+    Segment* seg = startSeg;
+    while (seg && seg->tick() <= endSeg->tick()) {
```
cbjeukendrup, line 398: Looks like this will go wrong if `endSeg` is `nullptr`. It looks like if `startSeg` or `endSeg` is `nullptr`, the code below is not executed anyway. Maybe it makes sense to add something like the following here above?
```c++
if (!startSeg || !endSeg) { // maybe also `|| startSeg == endSeg`?
    return;
}
```

Also, maybe in the `if` statement in line 417, you could replace `startSeg && endSeg && startSeg != endSeg` with `segList.empty()`.

src/engraving/libmscore/slur.cpp
```diff
@@ -1462,23 +1467,31 @@ SpannerSegment* Slur::layoutSystem(System* system)
 
             _up = !(startCR()->up());
 
+            // Check if multiple voices
+            bool multipleVoices = false;
             Measure* m1 = startCR()->measure();
-            if (c1 && c2 && !c1->isGrace() && isDirectionMixture(c1, c2)) {
+            ChordRest* cr2 = endCR() ? endCR() : startCR();
+            while (m1 && m1->tick() <= cr2->tick()) {
+                if ((m1->hasVoices(startCR()->staffIdx(), tick(), ticks() + (cr2 ? cr2->ticks() : startCR()->ticks())))
```
cbjeukendrup, line 1475: Sorry, another tiny thing. Since `endCR()` is never nullptr here (see line 1441), the line `ChordRest* cr2 = endCR() ? endCR() : startCR();` is not necessary. Just use `endCR()` instead of `cr2`. 

For the same reason, `(cr2 ? cr2->ticks() : startCR()->ticks())` can be replaced with simply `endCR()->ticks()`.

src/engraving/libmscore/slur.cpp
```diff
@@ -377,17 +377,50 @@ void SlurSegment::avoidCollisions(PointF& pp1, PointF& p2, PointF& p3, PointF& p
...
+    }
+
     // Collision clearance at the center of the slur
     double clearance = 0.75 * spatium(); // TODO: style
     // balance: determines how much endpoint adjustment VS shape adjustment
     // we will do (0 = only shape, 1 = only end point)
     double balance = 0.4; // TODO: style (maybe?)
 
-    if (startSeg && endSeg && startSeg != endSeg && autoplace()) {
+    if (autoplace()) {
```
cbjeukendrup, line 423: I just realize that this `if` statement can be replaced with 
```c++
if (!autoplace()) {
    return;
}
```
at the very top of the function. That further reduces the number of nested { } blocks. 

Better to do this in a separate commit though, since it creates a lot of changed lines where only the indent level has really changed.

mike-spa: Actually, if autoplace is off it doesn't even make sense to call this method, so I can move this check when the method is called.

cbjeukendrup (approved):

PR #12563, opened 2022-07-26
Audit translatable strings. Finale

cbjeukendrup:
Qt's translation system is not perfect:
- Since we're always working with just strings, it is not clear where we are using not-yet-translated strings or already-translated strings. So sometimes we are attempting to translate a string twice (i.e. pass an already-translated string to the translation functions), or we completely forget to translate a string, as currently happening in anything with AbstractMenuModel (app menu bar, note input bar...)

- When we store the not-yet-translated version of a string and want to pass it to the translation functions later, we must make sure that we don't modify the stored string. For example, the following won't work:
    ```c++
    // Declaration of translatable string
    int thingNo = 42;
    String myFailingTranslatableString = String(QT_TRANSLATE_NOOP("context", "Thing no. %1")).arg(thingNo);

    // Later:
    String myFailedTranslation = mtrc("context", myFailingTranslatableString);
    ```
    In this case, the translation system knows the translation for `"Thing no. %1"`, but we are trying to retrieve the translation of `"Thing no. 42"`, which the translation system doesn't know. This has gone wrong in PR #12362.

- Every time that we want to reuse a string, we need to type exactly the same context, key, and disambiguation. This also applies when a string is "declared" using `QT_TRANSLATE_NOOP`, and translated later, in a different part of the code: When calling the translation function, we need to type the same context as where we declared the string, even though this may be in a totally different module. So if we change the context of a translatable string, we have to manually find all places where this string might be used, and update the context there too. That's very difficult and error-prone, since you need to track down all places where the string is stored and retrieved from there...

    Even worse: it may be that we are receiving strings from different contexts, and need to translate them in one place. Something like this is happening in the Palettes. How do we know the context for a string, if we only know the string itself? We can't. We can only guess, like the heuristic in `PaletteCell::translationContext()` is doing. 

So, enough ways to break things!

But that's not what we want to do. Therefore, I'm introducing a new type: `TranslatableString`. It solves al three problems mentioned above.

It stores the combination of context/string/disambiguation. So you create a TranslatableString in one place, and then just call its `.translated()` method anywhere, and you will always get the correct translation. 

Since it is a distinct type, you can't accidentally use the string without translating it. You either have to pass it through as a TranslatableString, so that it stays translatable, or you pass the translated version of the string (uhm oh, yes, you can of course still pass that already-translated string back to the translation functions, which stay untouched by this PR. But you just won't do that, okay?)

But the most advanced thing is that you can also call `.arg` on TranslatableString. Using a far-too-complicated mechanism, the arguments are stored in the TranslatableString, but the string itself stays unmodified, so that the translation system can still find the translation. And when you call `.translated()`, the arguments are applied to the translated string before it is returned. Due to the far-too-complicated way this is implemented, it is ensured at compile time that you don't pass arguments to `TranslatableString`'s `.arg` method that can't be passed to `String`'s `.arg` method.

There are places where you sometimes need to use a translatable string, and sometimes just a plain string. For example, for UiActions, you usually want the title to be translatable, but sometimes it should just be a file path or the name of a plugin. This use case is also catered for by TranslatableString: using `TranslatableString::untranslatable(...)`, you can create a TranslatableString that doesn't have a context, which gets special treatment when calling its `.translated()` method (but `.arg` still works). 

(If we ever need it, we can even do this:)
```c++
TranslatableString operator+(const TranslatableString& lhs, const TranslatableString& rhs)
{ 
     return TranslatableString::untranslatable("%1%2").arg(lhs, rhs);
}
```

And what about finding TranslatableString with `lupdate`? We simply add `TranslatableString` as an alias for the `translate` function in `run_lupdate.sh`. `lupdate` doesn't care that this is the constructor of a type rather than a function.
**NOTE:** There is one small catch though. The following string won't be found by `lupdate`:
```c++
TranslatableString doesntWork("context", "string");
```

This `TranslatableString` type allows to remove some duplications from the list of translatable strings: for example, instead of "Fret 1", "Fret 2", "Fret 3", we now have "Fret %1" and use `.arg`.

However, we are not yet using TranslatableString to its full potential. In Palettes, the cell names also need to be read from and written to a file, which is something we can't do super easily with TranslatableString, especially if we want to take into account the `.arg` arguments. Thus, we are not using it yet for palette cell names, and instead still relying on the `PaletteCell::translationContext()` heuristic. For this heuristic to work, all cell names related to a certain ElementType need to be in the same translation context, for example forcing us to put unrelated strings in the "engraving/sym" context. Very unfortunate that those palettes spoil the fun for big parts of the engraving module. Anyway, let's tackle this later, when we overhaul Palettes...


cbjeukendrup:
My worst fear has become truth: GCC is unable to compile it (even though it is valid C++17, isn't it?). It fails to deduce the template parameter pack inside the TranslatableString.arg methods. That's quite a bummer. 

So I'll have to specify those parameters explicitly. That is a tiny bit less fun, because it breaks the future possibility to use C++20 concepts to add deduction guides to the `TranslatableString::Arg` template, so that we can also do `.arg(u"string literal")`. This won't work currently, since the type is inferred to `const char16_t[15]` instead of `String`. As a workaround, we can use `.arg<String>(u"string literal")` or `.arg(String(u"string literal"))`. So, since it wouldn't make much of a difference as long as we use C++17, it is only a tiny bit less fun, but still, less fun.

jeetee:
u"string" is expected to be `const char16_t*` encoded as UTF-16 AFAIK (at least according to MSVC docs)

Doesn't appending 's' help: `u"string"s would then either evaluate to std::string or to std::u16string as I understand it.

Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#slstr12-use-the-s-suffix-for-string-literals-meant-to-be-standard-library-strings and https://docs.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp?view=msvc-170

igorkorsukov:
Great idea!

cbjeukendrup:
@jeetee At least on Clang, the type of `"..."` is `const char[]`. If it is an rvalue, it is "decayed" to `const char*` when passing it to a function or constructor, so therefore you can construct a `String` from it. But in this case, the `const char[]` is _stored_ somewhere, which turns it into a lvalue, so that the implicit conversion to `const char*` doesn't work anymore (according to Clang), and thus you can't construct a `String` or anything else from it, thus you can't pass it to `.arg`.
And using things like std::decay_t or rvalue reference overloads doesn't really help, it causes more trouble than it solves...

Anyway, it seems like I have now found another solution!

igorkorsukov (approved):

src/framework/global/types/translatablestring.h
```diff
@@ -0,0 +1,173 @@
...
+#include "types/string.h"
+
+#ifndef NO_QT_SUPPORT
+#include <QString>
+#endif
+
+namespace mu {
+//! Note: in order to make the string visible for `lupdate`,
+//! you must write TranslatableString(...) explicitly.
+struct TranslatableString {
```
shoogle: It's quite a long name for something that will appear in lots of places. How about calling the struct `Translatable` since it doesn't inherit from `String`?

src/framework/global/types/translatablestring.h
```diff
@@ -0,0 +1,173 @@
...
+    const char* context = nullptr;
+    const char* str = nullptr;
+    const char* disambiguation = nullptr;
+
+    TranslatableString() = default;
+    TranslatableString(const char* context, const char* str, const char* disambiguation = nullptr)
+        : context(context), str(str), disambiguation(disambiguation) {}
```
shoogle, line 47: I think the constructor will need an `int n` parameter so that `lrelease` knows to request translations with plurals (look for `<numerusform>` in the .ts files). You could store `n` as a member variable but still allow requesting translations with a different number later on:

```c++
    inline String translated(int num = -1) const
    {
        // ...

        assert(num >= -1);
        assert(num == -1 || n != -1);
        String res = mtrc(context, str, disambiguation, num == -1 ? n : num);

        // ...

        return res;
    }
```
With `assert(n >= -1)` in the constructor.

src/framework/global/types/translatablestring.h
```diff
@@ -0,0 +1,173 @@
...
+
+    inline bool isValid() const
+    {
+        // Not null or empty
+        return context && context[0] && str && str[0];
```
shoogle: Do we ever want an empty translatable string? Perhaps this line should just be an assertion in the constructor?

shoogle (commented):
This is a very clever system! I thought we might need delayed access to translated and non-translated versions of strings at some point but I never figured out how to make delayed `arg` replacements work. Mind you, I'm not sure I fully understand your solution to that particular problem, but if it works it works!

shoogle:
Looks like I got in slightly late with the review 🙂

src/framework/global/types/translatablestring.h
cbjeukendrup: We certainly want the ability to default-construct a TranslatableString. For example, to use it as the default return value for a method, or for "optional" arguments for methods (i.e. if you pass nothing, an empty string will be passed).

src/framework/global/types/translatablestring.h
cbjeukendrup: Good point. I believe `lupdate` won't look at the signature of this constructor though, only at the callers. But indeed, if you pass a string containing `%n`, but don't specify a fourth parameter, `lupdate` is apparently confused and gives a warning. 
Currently, we actually never need to use TranslatableString with %n strings, but when we start doing that, we should keep this in mind.

PR #12575, opened 2022-07-27
Fixing a crash with stemless notes with continued ties

asattely:
There is a crash caused by a small bug in https://github.com/musescore/MuseScore/pull/12569. This fixes it.


oktophonie:
Blame me for not testing better.

oktophonie (approved):

PR #12576, opened 2022-07-27
Translatable strings: Instruments

cbjeukendrup:
Last but not least: instrument names, score order names and template names were not being translated yet. I added them back to the lupdate/lrelease workflow.
The translations are not yet being used in the UI. This can be solved separately and is less urgent, as it doesn't affect the work of the translators. 

Contains / builds on #12563, which should be merged first.


share/instruments/instruments.xml
```diff
@@ -6551,6 +6551,24 @@
                   <genre>concertband</genre>
                   <genre>marching</genre>
             </Instrument>
+            <Instrument id="bell-tree">
```
shoogle, line 6554: Sorry, it's my fault this got in. I've removed this instrument from the spreadsheet for now. Could you drop this commit and then run `update_instruments_xml.py` and `lupdate` again in each subsequent commit where you ran them before? You can ignore any merge conflicts as these will disappear after you run the script to regenerate the files.

shoogle (commented):
Looks good!

igorkorsukov (approved):

share/instruments/instruments.xml
cbjeukendrup: Done!

shoogle (approved):

share/instruments/update_instruments_xml.py
```diff
@@ -360,83 +360,87 @@ def disambiguation(instrumentId: str, nameType: str):
...
             for nameType in ['trackName', 'longName', 'shortName']:
-                add_translatable_string_if_not_null(f, 'InstrumentsXML', instrument[nameType],
+                add_translatable_string_if_not_null(f, 'engraving/instruments', instrument[nameType],
                                                     disambiguation(instrumentId, nameType))
                 
             if instrument['traitName'] != '[hide]':
-                add_translatable_string_if_not_null(f, 'InstrumentsXML', instrument['traitName'],
+                add_translatable_string_if_not_null(f, 'engraving/instruments', instrument['traitName'],
                                                     disambiguation(instrumentId, 'traitName'), 
                                                     'traitName codes: * most common for instrument; () hide in score')
```
cbjeukendrup, line 434: @shoogle Considering your explanation about the meaning of `*`, shouldn't we update/extend this comment? Maybe we should put a link to some external documentation here, since a decent explanation of the possibilities translators have would maybe not fit in a comment.

shoogle: You're right. I'll put something together quickly so you have a URL to include.

The comment could also mention the traitType for the instrument (transposition, tuning or course).

PR #12578, opened 2022-07-28
Added disable implicit shared option

igorkorsukov:
On MacOS with clang there are problems with debugging - the value of the string is not visible. It's probably because use the shared pointer. So, I added an option that disables implicit shared (using shared pointer)


igorkorsukov:
it didn't help

PR #12149, opened 2022-06-23
[MU4] MinGW compile fix

Eism:
Resolves: https://github.com/musescore/MuseScore/pull/12115#issuecomment-1164581520

PR #12580, opened 2022-07-28
[MU4] Fixed device name convert

Eism:
It is partial revert of #12149
Later we will think about the best solution without warning

### DIFF
--- a/src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
+++ b/src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
@@ -49,10 +49,9 @@ struct WinCoreData {
     HANDLE hEvent;
 };
 
-static LPWSTR stringToLPWSTR(const std::string& str)
+static std::wstring stringToWString(const std::string& str)
 {
-    static std::wstring wstr = QString::fromStdString(str).toStdWString();
-    return LPWSTR(wstr.c_str());
+    return QString::fromStdString(str).toStdWString();
 }
 
 static std::string lpwstrToString(const LPWSTR& lpwstr)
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
```diff
@@ -51,8 +51,7 @@ struct WinCoreData {
 
 static LPWSTR stringToLPWSTR(const std::string& str)
 {
-    static std::wstring wstr = QString::fromStdString(str).toStdWString();
-    return LPWSTR(wstr.c_str());
+    return LPWSTR(QString::fromStdString(str).toStdWString().c_str());
 }
```
cbjeukendrup, line 55: I think the correct way is:
```c++
// Here
static std::wstring stringToWstring(const std::string& str)
{
    return QString::fromStdString(str).toStdWString();
}

// In CoreAudioDriver::open
...
    } else {
        std::wstring wstr = stringToWstring(outputDeviceId);
        hr = pEnumerator->GetDevice(wstr.c_str(), &pDdevice);
    }
```
The idea is that the `std::wstring` needs to be alive always while we use its `c_str()` pointer. Basically, we should never call `c_str()` on a temporary `std::string` or `std::wstring`.

boryskuzmenko:
**Tested**
Please, merge to master

src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
Eism: thanks, fixed

cbjeukendrup (approved):

PR #12583, opened 2022-07-28
[MU4] Implemented updating playback range when interacting with articulations and accidentals

Eism:


bkunda:
Sorry to say this crash still persists in this build on MacOS:

Also the hyphens themselves don't enter as expected. 

https://user-images.githubusercontent.com/86290556/182624642-f64bdcd5-d6e3-46e7-be42-4ea4f62dff9b.mov

PR #12584, opened 2022-07-28
Incremental improvements to horizontal spacing

mike-spa:
UPDATED DESCRIPTION.
This PR contains three separate improvements (made in the three separate commits), all related to horizontal spacing. Some of them seem like a lot of code, but 90% of it is actually just refactoring.
1) When enforcing minimum measure width we should not take into account the space occupied by the header, otherwise the result is irregular. I've added the check to do this, and moved all the minMeasureWidth logic to a dedicated method.
Before:
minMeasWidthAFTER
After:
minMeasWidthBEFORE

2) The longest note of a system (which is needed for the spacing calculation) is calculated at the system level and passed down until Measure::durationStretch(), instead of computing it inside durationStretch(). This potentially saves thousands of pointless loops through the system segments.

3) System justification has been updated to use a bisection algorithm rather than a gradient descent, which makes it stronger against edge cases. Side benefit: I could reduce the safety exit from 100 down to 50 iteration. I also moved the logic to a dedicated method, which I probably should have done earlier.

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -4386,23 +4386,43 @@ void Measure::computeWidth(Fraction minTicks, double stretchCoeff)
     computeWidth(s, x, isSystemHeader, minTicks, stretchCoeff);
 
     // Check against minimum width and increase if needed
-    double minWidth = isMMRest() ? score()->styleMM(Sid::minMMRestWidth) : score()->styleMM(Sid::minMeasureWidth);
-    double maxWidth = system()->width() - system()->leftMargin(); // maximum available system width (left margin accounts for possible indentation)
-
-    // System width may not yet be available for the linear mode (e.g. continuous view)
-    // Will use the minimum width from the style in this case
-    if (maxWidth <= 0) {
-        maxWidth = minWidth;
-    }
-
-    minWidth = std::min(minWidth, maxWidth); // Accounts for a case where the user may set the minMeasureWidth to a value larger than the available system width
-
+    double minWidth = computeMinMeasureWidth();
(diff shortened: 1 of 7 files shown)


mike-spa:
vtests looking good now

oktophonie (approved):

src/engraving/libmscore/measure.cpp
```diff
@@ -4386,23 +4386,41 @@ void Measure::computeWidth(Fraction minTicks, double stretchCoeff)
...
-
-    // System width may not yet be available for the linear mode (e.g. continuous view)
-    // Will use the minimum width from the style in this case
     if (maxWidth <= 0) {
+        // System width may not yet be available for the linear mode (e.g. continuous view)
+        // Will use the minimum width from the style in this case
         maxWidth = minWidth;
     }
-
     minWidth = std::min(minWidth, maxWidth); // Accounts for a case where the user may set the minMeasureWidth to a value larger than the available system width
```
cbjeukendrup, line 4407: I tested it, and it seems this line is not entirely doing its job. With a high enough minimum measure width, measures can now become wider than the system, but only to a certain extent, not infinitely. 
Schermafbeelding 2022-07-30 om 04 08 34

Probably, the cause is that you return `minWidth + startPosition` at the bottom of the function, which is equal to `maxWidth + startPosition` in the case that `maxWidth` initially was smaller than `minWidth`.

I think the solution is to return `std::min(minWidth + startPosition, maxWidth)` instead, and drop the line `minWidth = std::min(minWidth, maxWidth);`.

mike-spa: Good catch! Should be fixed now

PR #12593, opened 2022-07-29
[MU4] Notes can share the same accidentals

asattely:
The issue:
image

This is a longstanding issue, and one that predominantly affects scores migrated from MU1, which treated unison accidentals extremely weirdly. Now, two unison notes will only have one copy of the same accidental.


oktophonie (approved):

asattely:
I will add a vtest for this one, please hold

RomanPudashkin (approved):

Issue #12550, opened 2022-07-25
[MU4 Issue] Splash-screen is not displayed when running MU4 on Linux


boryskuzmenko:
**To Reproduce**
1. Open the last master build https://github.com/musescore/MuseScore/actions/runs/2729870681
2. Pay attention to the frame on the center of screen

**Expected behavior**
Splash-scree is displayed when running MU4

**Screenshots**
![Screenshot from 2022-07-25 20-08-36](https://user-images.githubusercontent.com/101250347/180836443-d810c73d-d8b1-4f56-b3eb-b5808efb7457.png)

**Platform information**
 - OS: Linux

Eism:
On Windows, the background should be transparent

![image](https://user-images.githubusercontent.com/10116828/180945226-dce302a8-bf03-4355-a0af-4543e8edb1fa.png)

sammik:
In my desktop (Ubuntu studio, KDE, Plasma), splash screen is there (latest nightly - 28.7.).
Corners are not transparent.

https://user-images.githubusercontent.com/1646034/181507022-499b0d72-f51b-479c-83b3-60f0f251cda8.mp4

MarcSabatella:
For me on Debian (running under ChromeOS, so no Linux-side window manager), the splash screen is there, and annoyingly, it's also there again every time I create a new score.  I guess because a new instance of MuseScore is created for each new score, but only the first should create the splash screen.

HemantAntony:
But what if the score is quite large and it takes time to load. Won't it be annoying if there is no indication that the score is being opened?

cbjeukendrup:
The loading of the score happens after the app has already launched, so when the splash screen is already gone. 

If we want, we can make the splash screen appear only for the first instance that is being launched, but I wouldn't know why we'd do that. For secondary instances, it takes exactly as long to load as for the first instance, and there is no indication that a new instance is being loaded, so the user has to wait multiple seconds while nothing seems to happen.

MarcSabatella:
I would just say, I can't recall ever seeing a splash screen used for that - they are normally startup only.  Instead, a separate loading progress dialog would normally be used if needed.  As it is, seeing the splash screen again look like a bug, since it is unexpected based on how other programs work.

boryskuzmenko:
🟢 **FIXED**
Can't be reproduce anymore

PR #12595, opened 2022-07-29
Partly reimplement SplashScreen

cbjeukendrup:
Resolves: #12943

Fixes a macOS bug: it was not possible anymore to open a file from Finder with MuseScore unless MuseScore was already running. (MuseScore would launch correctly, but forget to open the file, since the QOpenEvent would already be consumed by hidden calls to `QApplication::processEvents` in the QSplashScreen implementation. ~This is avoided by specifying `QEventLoop::ExcludeUserInputEvents`~ EDIT: that apparently doesn't work reliably either (see below); implemented a different solution.)

_Might_ fix some problems on Linux too (from #12550), but this is really a blind guess, since I still don't have a Linux installation to test things myself.

### DIFF
--- a/src/appshell/view/internal/splashscreen.cpp
+++ b/src/appshell/view/internal/splashscreen.cpp
@@ -22,7 +22,9 @@
 
 #include "splashscreen.h"
 
+#include <QApplication>
 #include <QPainter>
+#include <QScreen>
 #include <QSvgRenderer>
 
 #include "translation.h"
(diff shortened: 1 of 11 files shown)


RomanPudashkin (approved):

src/appshell/view/internal/splashscreen.cpp
```diff
@@ -43,16 +45,34 @@ static const QColor versionNumberColor("#22A0F4");
...
-    : QSplashScreen(QPixmap(splashScreenSize)),
+    : QWidget(nullptr, Qt::SplashScreen | Qt::FramelessWindowHint),
     m_backgroundRenderer(new QSvgRenderer(imagePath, this))
 {
     setAttribute(Qt::WA_TranslucentBackground);
+    setSize(splashScreenSize);
 
     // Can't make translatable, because translation system not yet initialized
-    showMessage("Loading…");
+    m_message = "Loading…";
```
Jojo-Schmitz, line 55: Still true that translations are not yet intialized? ISTR that this is now done quite a bit earlier?

cbjeukendrup: #12712 has not been merged yet...

Jojo-Schmitz: Would it make a difference? Is so I'd make the text translatable here already.

boryskuzmenko:
@cbjeukendrup Unfortunately, I can still reproduce the issue.
I could only open a few scores from finder but most of my scores can't be opened from finder (but can be from MU)
Here're three scores to check the issue
[Archive.zip](https://github.com/musescore/MuseScore/files/9425284/Archive.zip)

cbjeukendrup:
@boryskuzmenko That's unfortunate, I'll have another look. 
Could you tell me, does this PR improve anything regarding #12550?

cbjeukendrup:
Rebased to fix conflicts.

@RomanPudashkin Please have a look at the latest version of this :)

cbjeukendrup:
Rebased again.

It would be nice if this could be merged before beta, because it solves #12943 which might be quite an annoying issue for macOS users.

RomanPudashkin (approved):

DmitryArefiev:
Tested #12943 on Win10, **Mac12.5**, LinuxMint 20.2 - **FIXED**

PR #12597, opened 2022-07-29
Harp pedalling diagrams and UI

miiizen:
This pull request adds harp pedalling diagrams to MuseScore for [GSoC 2022](https://musescore.org/en/user/3773138/blog/2022/06/01/gsoc-2022-harp-pedalling-diagrams).

Pedal diagram demo

*Known issues*
 - [x] Opening via shortcut
 - [x] Stacking of diagrams around other elements
 - [x] Placement of the popup on the score
 - [x] The arrow at the top of the UI
 - [x] Probably some spacing within the UI

---


src/notation/qml/MuseScore/NotationScene/internal/ElementPopupLoader.qml
```diff
@@ -0,0 +1,115 @@
...
+    QtObject {
+        id: prv
+
+        function setPopupType(elementType) {
+            type = elementType
+        }
+
+        function componentByType(type) {
+            switch (type) {
+            case 52: return harpPedalComp
```
miiizen, line 52: this is unclear, need to expose the ElementType enum to QML

src/engraving/libmscore/note.cpp
```diff
@@ -1436,6 +1436,13 @@ void Note::draw(mu::draw::Painter* painter) const
                 painter->setPen(selected() ? engravingConfiguration()->warningSelectedColor() : engravingConfiguration()->warningColor());
             }
         }
+        // Warn if notes are unplayable based on previous harp diagram setting
```
miiizen, line 1439: Every notehead causes a search for a previous diagram currently - I'm thinking this isn't sustainable and a change of approach might be necessary

cbjeukendrup: Many similar things (clef, key sig, time sig) are stored in the Staff class, indeed as a `tick`-to-`…` map. Maybe that'd work for harp pedal state too?

(Disadvantage is of course that it would require extra effort to apply it to both staves instead of only the staff to which the diagram belongs technically. However, I believe this is a more general problem, that applies to dynamics too.)

src/palette/internal/palettecreator.cpp
```diff
@@ -2011,6 +2014,23 @@ PalettePtr PaletteCreator::newPitchPalette(bool defaultPalette)
...
+    sp->setGridSize(90, 30);
+    sp->setDrawGrid(true);
+    sp->setVisible(true);
+
+    auto pedalDiagram = Factory::makeHarpPedalDiagram(gpaletteScore->dummy()->segment());
+    sp->appendElement(pedalDiagram, QT_TRANSLATE_NOOP("palette", "Harp pedal diagram"));
+
+    auto pedalTextDiagram = Factory::makeHarpPedalDiagram(gpaletteScore->dummy()->segment());
+    pedalTextDiagram->setIsDiagram(false);
+    sp->appendElement(pedalTextDiagram, QT_TRANSLATE_NOOP("palette", "Harp pedal text"));
```
cbjeukendrup, line 2033: Tiny one: is it intended that this one says "Harp pedal text", while other places use "Harp pedal text _diagram_"?

cbjeukendrup (commented):
Some (probably last) minor comments about everything except the changes in the notation module! (those I will review later)

cbjeukendrup:
This PR needs a rebase too; this time all conflicts are just about the #includes.

src/engraving/libmscore/note.cpp
miiizen: Yes, I think it'd sort a few inefficiencies.  Would storing it in the `Part` class be an acceptable solution to make it appear on both staves?

cbjeukendrup: Interesting idea. It looks like the Part class has not really been used yet for storing data like this, so I'm not completely sure what the implications might be, but we could definitely try it. Maybe it turns out to be a very good idea.

miiizen: It definitely does what I want but I'm not sure what the wider implications would be!  There don't seem to be any issues so far.

src/engraving/libmscore/select.cpp
```diff
@@ -1026,6 +1027,8 @@ ByteArray Selection::symbolListMimeData() const
         case ElementType::HAIRPIN:
             seg = toHairpin(e)->startSegment();
             break;
+        case ElementType::HARP_DIAGRAM:
+            seg = toHarpPedalDiagram(e)->segment();
```
cbjeukendrup, line 1031: Missing a `break;` here?

src/notation/qml/MuseScore/NotationScene/internal/HarpPedalPopup.qml
```diff
@@ -0,0 +1,242 @@
...
+        // not enough room below stave to open so open above
+        if (root.y > notationView.height) {
+            root.y = pos.y - size.y
+        }
+    }
+
+    ColumnLayout {
+        id: menuItems
+
+        RowLayout {
```
cbjeukendrup, line 82: Would it be possible to refactor this into a GridLayout? I think that could result in cleaner code. Also, for the dummy items: it would be preferable to solve this using Layout.leftMargin and Layout.rightMargin, but if you really need dummy items, then I think you can just use an `Item` rather than a dummy ColumnLayout.

src/notation/qml/MuseScore/NotationScene/internal/HarpPedalGridColumn.qml
```diff
@@ -0,0 +1,105 @@
...
+
+    Layout.preferredWidth: 20
+    Layout.preferredHeight: 120
+
+    function checkPedalState(string, state) {
+        var pedalState = harpModel.pedalState
+        if (pedalState[string] == state) {
+            return true
+        }
+        return false
```
cbjeukendrup, line 46:
```suggestion
        return harpModel.pedalState[string] === state
```

src/notation/qml/MuseScore/NotationScene/internal/HarpPedalPopup.qml
```diff
@@ -0,0 +1,242 @@
...
+                }
+            }
+
+            Repeater {
+                width: parent.width
+
+                model: [
+                    { buttonsId: "DStringButtons", stringId: 0, stringName: "D" },
+                    { buttonsId: "CStringButtons", stringId: 1, stringName: "C" },
+                    { buttonsId: "BStringButtons", stringId: 2, stringName: "B" },
```
cbjeukendrup, line 140: I think the string names will need to be translatable.

src/notation/qml/MuseScore/NotationScene/internal/HarpPedalGridColumn.qml
```diff
@@ -0,0 +1,105 @@
...
+    RoundedRadioButton {
+        Layout.fillWidth: true
+        Layout.fillHeight: true
+        spacing: 5
+        checked: checkPedalState(repeaterModel.stringId, 2)
+
+        navigation.name: repeaterModel.stringName + " Sharp"
+        navigation.panel: navPanel
+        navigation.order: repeaterModel.stringId * 3 + 3
+        navigation.accessible.name: repeaterModel.stringName + qsTrc("notation", " sharp")
```
cbjeukendrup, line 101: We got to be careful here regarding translations. IIRC, in German there is no word for "sharp" or "flat" but they include the accidental in the note name (e.g. F Sharp is Fis). (@Jojo-Schmitz Please correct me if I'm wrong 😉)

So I think we'll need a matrix of 3x7 translatable strings, namely the cartesian product of [flat, natural, sharp] and the 7 notes.

Jojo-Schmitz: Well, we'd have "Auflösungszeichen", "Kreuz" and "Be", but not as a postfix to a notename or whatever `stringName` may contain? Also a leading space in a translatable string is a bad idea.

src/notation/qml/MuseScore/NotationScene/internal/HarpPedalPopup.qml
miiizen: Including the diagram/text buttons in the grid?  I remember running into trouble with using the separator line in the grid too.

src/notation/view/internal/harppedalpopupmodel.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+bool HarpPedalPopupModel::belowStave() const
+{
+    if (m_diagram != nullptr && (m_diagram->getProperty(mu::engraving::Pid::PLACEMENT) == mu::engraving::PlacementV::BELOW)) {
+        return true;
+    }
+    return false;
+}
+
+QPointF HarpPedalPopupModel::staffPos() const
```
miiizen, line 119: I've noticed one remaining issue - the popup tries to avoid all staves instead of just the stave the diagram is attached to as measure contains the whole system.

src/engraving/libmscore/harppedaldiagram.cpp
```diff
@@ -0,0 +1,370 @@
...
+    // Write vector of harp strings.  Order is always D, C, B, E, F, G, A
+    xml.startElement("pedalState");
+    for (int idx = 0; idx < _pedalState.size(); idx++) {
+        xml.tag("string", { { "name", idx } }, static_cast<int>(_pedalState[idx]));
+    }
+    xml.endElement();
+
+    TextBase::writeProperties(xml);
+    xml.endElement();
+}
```
cbjeukendrup, line 279: You'll need to rebase once more; the changes are the following:
- The `read` method has been removed definitively from all `EngravingItem` classes;
- The `write` method too; its functionality has been moved to the `TWrite` class;
- The headers `"rw/xml.h"` and `"rw/400/tread.h"` are no longer being included in files for `EngravingItem` classes.

cbjeukendrup (commented):
Another round of review comments! It's really very good now, so this is probably the last review round from me and then I'll ask a colleague to check it too.

Tantacrul:
Had a poke at the diagram to see how things are going with it. Found a few things and thought I'd share. Sorry if this is a bit premature and you were planning us to test later.

https://user-images.githubusercontent.com/21022311/235263589-30a26e9d-470c-4c01-a67d-56f4bd9343e5.mp4

cbjeukendrup:
@miiizen A rebase is needed again...

src/engraving/libmscore/harppedaldiagram.h
```diff
@@ -0,0 +1,94 @@
...
+
+#ifndef __HARPPEDALDIAGRAM_H__
+#define __HARPPEDALDIAGRAM_H__
+
+#include "pitchspelling.h"
+#include "textbase.h"
+#include "pitchspelling.h"
+
+using namespace mu;
+//using namespace mu::engraving;
```
RomanPudashkin, line 31: Please remove this line

RomanPudashkin:
This PR looks great (and ready to be merged), thank you so much! The only issue that it has is the broken unit tests. If you can't fix them yourself (or don't have enough time), I suggest disabling them (I'll fix them later). Or you can try rebasing this PR one more time, maybe this will help

RomanPudashkin (approved):

miiizen:
Thanks for the feedback!
Rebasing hasn't fixed them - afraid I won't have time to try to fix the tests til after Tuesday, feel free to have a look yourself.

cbjeukendrup:
Fortunately, I think the test failures are no mystery at all. The default z index of an element is defined as its ElementType casted to `int` multiplied by 100. Because the HARP_DIAGRAM type got inserted, the default z-index for images was incremented (because the `IMAGE` element type comes after `HARP_DIAGRAM`).

The reason that the z index is written to the file in those specific files, is that it is "customised" in those places. And the reason that it is customised _and_ gets changed while reading/writing the score, is `BSymbol::add` (and `Image` is derived from `BSymbol`). What happens is that when reading the score, the z index of the inner image gets set to the z index of the outer image minus one. The z index of the outer image was not customised, so will user the new default value. Therefore the z index of the inner image will be changed too.

So the fix would be to just update the contents of `src/engraving/tests/parts_data/part-image.mscx` and `src/engraving/tests/parts_data/part-image-parts.mscx` according to  https://github.com/musescore/MuseScore/actions/runs/4952108865/jobs/8857989773?pr=12597#step:9:3323

PR #12605, opened 2022-07-29
Small improvements to cross-staff slurs

mike-spa:
I've removed some checks that were no longer relevant, and improved the logic that determines whether the slur is cross-staff (see following example: the slur should be considered cross-staff even though start and end chord have the same `vStaffIdx()`). I've also moved this to a dedicated method to avoid repeated code. @oktophonie 
Before:
slurCrossStaffBEFORE

After:
slurCrossStaffAFTER


oktophonie (approved):

RomanPudashkin:
Could you please add a vtest for this case?

mike-spa:
> Could you please add a vtest for this case?

@RomanPudashkin done! The vtest is obviously failling as expected. No idea what's going on with the Windows build (seems to be failing on something connected to translations? any idea @cbjeukendrup?).

RomanPudashkin (approved):

cbjeukendrup:
Something in the "Setup dependencies" step seems to have failed, which is often a temporary glitch caused by something external. Not very good that this failure doesn't cause an early exit... 

Anyway, I'll restart it, and then it _should_ be fine.

Jojo-Schmitz:
Also a a `vtests` label?

Issue #12603, opened 2022-07-29
[MU4 Issue] Wrong symbols are rendered instead of values in Zoom drop-down


boryskuzmenko:
**Expected behavior**
Real zoom values are rendered

**Screenshots**
<img width="439" alt="image" src="https://user-images.githubusercontent.com/101250347/181800758-6fc4650d-fe92-4d73-9ba6-cb1b26e91943.png">

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

PR #12606, opened 2022-07-29
Fix zoom percentages dropdown

cbjeukendrup:
Resolves: #12603 

Turns out my `TranslatableString::fake_deduction_guides_t` was working a bit too aggressively, interpreting even plain numbers as convertible to String. That had somewhat strange effects. Get rid of it completely for now, since we are not really using it anyway.

Furthermore, actually completely separate from that, fixes String.arg for strings that contain a percent sign not followed by 1-9.

### DIFF
--- a/src/appshell/view/notationstatusbarmodel.cpp
+++ b/src/appshell/view/notationstatusbarmodel.cpp
@@ -198,23 +198,19 @@ void NotationStatusBarModel::load()
             }
         }
     });
-
-    emit availableViewModeListChanged();
-    emit availableZoomListChanged();
 }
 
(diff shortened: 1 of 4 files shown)


DmitryArefiev:
Hi @cbjeukendrup, I found another issue: **zero appears and checked**

Screenshot 2022-08-01 124354

Do you want to fix it in the current PR?

DmitryArefiev:
Tested #12603 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

igorkorsukov (approved):

PR #12609, opened 2022-07-30
Add `instruments` to `.tx/config`

cbjeukendrup:
so that it will be pushed/pulled by tx


cbjeukendrup:
Sorry, I'm going to merge this now without review. It's better to do this sooner rather than later, and I'm confident enough that this change is correct. And nothing can go wrong, as I have created a local backup of all translations on Transifex right now.

Issue #12608, opened 2022-07-30
[MU4 Issue] Some text strings in toolbar are using incorrect data


bkunda:
**Describe the bug**
I suspect that since #12414 was merged, some of the text strings in the UI are now using data from `proposed new description (used in preferences)` instead of `Description.` (Referring here to the MScore Actions spreadsheet @VanSHOE @shoogle).  

The idea behind #12414 was to change only the action titles in the preferences dialog, not elsewhere in the UI. 

This problem can be seen, for example, in the playback settings menu (current nightly) (E.g. "toggle enable MIDI input"):

<img width="267" alt="Screenshot 2022-07-30 at 9 58 04 am" src="https://user-images.githubusercontent.com/86290556/181861726-5c021bc5-35b2-4021-b40d-fe9ad160bb93.png">

This menu should look like this:

<img width="286" alt="Screenshot 2022-07-30 at 9 58 51 am" src="https://user-images.githubusercontent.com/86290556/181861760-4d26f5d5-32a4-40e0-b9d8-d54b941420e6.png">

Tooltips have also been affected. For example, "Manage parts" now reads as "Parts...", which is less useful as a tooltip. 

**Expected behavior**
Tooltips and text labels in the UI should retain their data from `Description` instead of `Proposed new description (used in preferences`)

**Platform information**
 - MacOS 12.3.1

**Additional context**
Flagging as priority because this should be addressed before string freeze

cbjeukendrup:
This is why I had doubts about using the description in #12414. It means that all tooltips in the playback toolbar, the note input bar and the parts/mixer toolbar now use the same text as in Preferences, which is often not desirable. 

So I think each action needs up to three things:
1. Every action that can have a shortcut must have a _name_, that is used in Preferences. (So that would be what's currently called `description` since #12414.)
2. Every action that is exposed in a menu or toolbar must additionally have a _title_, that is displayed in the menu or used by screen readers and for the tooltip title. (So that's what's currently already called title.)
(Actions that are never exposed in a menu or toolbar should preferably _not_ have such a title, to reduce the amount of unnecessary translatable strings. That means probably that we should remove quite a few existing titles.)
3. Some of the UI-exposed actions may additionally have a _description_. This may be used as extra screen reader info and as the second line in a tooltip, to give the user some extra hint about potentially non-obvious actions. (So this should generally be the `description` as it was before #12414, but in most cases just empty.)

[...]

PR #12611, opened 2022-07-30
fix: #12608 Menu items in the playback settings now read titles

VanSHOE:
Partial(only playback settings) Fix for [[MU4 Issue] Some text strings in toolbar are using incorrect data](https://github.com/musescore/MuseScore/issues/12608)

Hotfix for above-mentioned issue. This simply reads the menu-items text from title of the actions instead of the descriptions.
It also changes the titles to match the titles desired in the issue

### DIFF
--- a/src/playback/internal/playbackuiactions.cpp
+++ b/src/playback/internal/playbackuiactions.cpp
@@ -72,7 +72,7 @@ const UiActionList PlaybackUiActions::m_settingsActions = {
     UiAction("midi-on",
              mu::context::UiCtxAny,
              mu::context::CTX_ANY,
-             TranslatableString("action", "MIDI input"),
+             TranslatableString("action", "Enable ‘MIDI input’"),
              TranslatableString("action", "Toggle MIDI input"),
              IconCode::Code::MIDI_INPUT,
              Checkable::Yes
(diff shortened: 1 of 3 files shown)


src/playback/view/playbacktoolbarmodel.cpp
```diff
@@ -93,7 +93,7 @@ void PlaybackToolBarModel::updateActions()
     MenuItemList settingsItems;
 
     for (const UiAction& action : PlaybackUiActions::settingsActions()) {
-        settingsItems << makeActionWithDescriptionAsTitle(action.code);
```
cbjeukendrup, line 96: You can get rid of this whole method, since it was not more than an ugly hack just for these two cases...

cbjeukendrup (approved):
Happy from a technical point of view!

src/playback/internal/playbackuiactions.cpp
```diff
@@ -72,7 +72,7 @@ const UiActionList PlaybackUiActions::m_settingsActions = {
     UiAction("midi-on",
              mu::context::UiCtxAny,
              mu::context::CTX_ANY,
-             TranslatableString("action", "MIDI input"),
+             TranslatableString("action", "Enable MIDI input"),
              TranslatableString("action", "Toggle ‘Enable MIDI input’"),
```
cbjeukendrup, line 76: @bkunda How about changing `Toggle ‘Enable MIDI input’` to just `Toggle MIDI input`?

bkunda: I think just go with `Enable MIDI input` (the check mark alongside it is sufficient to indicate whether this is enabled or not).

cbjeukendrup: @bkunda Yes, for in the menu `Enable MIDI input` is good, together with the checkmark. 

But I was wondering about Preferences, because there is no checkmark of course. So I could imagine using "Toggle" there, to emphasize that it is an on/off thing. On the other hand, it might be confusing to use a different name in Preferences than in the menu, maybe in this particular case even slightly more than in other cases. So what do you think?

Issue #12610, opened 2022-07-30
[MU4 Issue] Empty state message in Instruments panel contains hard-coded keyboard shortcut


cbjeukendrup:
**Describe the bug**
If a score contains no instruments, the Instruments Panel displays a message:

> There are no instruments in your score. To choose some, press **Add**, or use the shortcut **'i'**

In this message, the shortcut 'i' is hard-coded, so if you have changed this keyboard shortcut in Preferences, the message will be incorrect. 

**To Reproduce**
1. Change the keyboard shortcut for "Add/remove instruments"
2. Open a score and delete all instruments
3. Note the incorrect message in the instruments panel

**Expected behavior**
The shortcut should be dynamic, or we shouldn't show it at all in this message. 

Additionally, I have the following requests:
- The shortcut should be displayed in uppercase (because we always use uppercase for shortcuts, and on a physical keyboard the letters are also uppercase)
- We should make it clearer by replacing "shortcut" with "keyboard shortcut"
- I'd think that the message should end with a '.' (full stop), especially since it consists of multiple sentences. In many places in the designs, a full stop is omitted. This feels incomplete and unnatural to me, but maybe that's because I'm not a native English speaker?

@Tantacrul Please let me know to what extent you agree with these requests :)

**Platform information**
 - OS: reported on macOS

PR #12615, opened 2022-07-31
Some last corrections to translatable strings

cbjeukendrup:
Resolves: #12610

### DIFF
--- a/src/instrumentsscene/view/instrumentspaneltreemodel.cpp
+++ b/src/instrumentsscene/view/instrumentspaneltreemodel.cpp
@@ -37,6 +37,8 @@ using namespace mu::notation;
 using namespace mu::uicomponents;
 using ItemType = InstrumentsTreeItemType::ItemType;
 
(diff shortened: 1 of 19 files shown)


src/instrumentsscene/view/instrumentspaneltreemodel.cpp
```diff
@@ -558,6 +564,11 @@ bool InstrumentsPanelTreeModel::isEmpty() const
     return m_rootItem ? m_rootItem->isEmpty() : true;
 }
 
+QString InstrumentsPanelTreeModel::addInstrumentsKeyboardShortcut() const
+{
+    return shortcuts::sequencesToNativeText({ shortcutsRegister()->shortcut(ADD_INSTRUMENTS_ACTIONCODE).sequences[0] });
```
RomanPudashkin, line 569: this code will cause a crash if the sequence list is empty

cbjeukendrup: Thanks, fixed!

RomanPudashkin (approved):

PR #12618, opened 2022-07-31
Fix #11621: implemented independent view modes (page, continuous, ...) per notation

### LINKED ISSUE #11621: [MU4 Issue] Switching view (Page view -> Continuous view etc.) should work independently for score and parts
https://user-images.githubusercontent.com/90187801/168129478-f4d1f2c3-f3e8-4692-b722-a7bc29a4b009.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
The same was fixed for zooming/scrolling in #7879
Doesn't occur in **MU3**

martin-danhier:
Resolves: #11621

Half of the work was already done since notations already have independent view modes, but each time the mode was modified, all notations were updated.

I modified the view settings in order to store separate view mode for each excerpt, that will be stored in the save file alongside the global one.

New save files should still work with previous versions (as it is only an addition), and old save files should be usable in the new version as well.

**Video:**

https://user-images.githubusercontent.com/36641081/182047468-dc60e11c-4629-4022-832f-1c062a2d4621.mp4

### DIFF
--- a/src/notation/imasternotation.h
+++ b/src/notation/imasternotation.h
@@ -49,6 +49,8 @@ public:
     virtual void addExcerpts(const ExcerptNotationList& excerpts) = 0;
     virtual void removeExcerpts(const ExcerptNotationList& excerpts) = 0;
 
+    virtual std::optional<size_t> findExcerptIndex(INotationPtr excerpt) const = 0;
+
     virtual void setExcerptIsOpen(const INotationPtr excerptNotation, bool opened) = 0;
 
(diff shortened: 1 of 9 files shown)


RomanPudashkin:
We will implement a different solution a bit later. It will be a pretty global change in the file structure, so we will do it before POST-BETA. The new solution will be more flexible than the current one in the master branch. Thanks for your help!

See https://github.com/musescore/MuseScore/issues/11621#issuecomment-1271377164

PR #12635, opened 2022-08-01
[MU4] Fix some bogus translatable strings

Jojo-Schmitz:
* a frame around text is a "rectangle" rather than a "square", most of the time (and a square is a rectangle, but not always vice versa)
* strings (of plucked instruments) are "Always open" rather than just "Open", when not over the fretboard, that term is at least clearer
* "arpeggio" should be plural here
* "Midi" to "MIDI", being an acronym
* "Spacing curve" to "Spacing ratio"
* "tool bar" to "toolbar"
* Remove leading space
* "MIDI device mapping" to "MIDI mappings"
* "bar" to "measure"
* Clarify some translation needs, for sharp, flat, natural, reg. accessibility
* "Mask H-bar around number" to "Break H-bar if it collides with number"

See also #12664, for changes to intruments.xml / instrumentsxml.h

### DIFF
--- a/src/appshell/qml/MainToolBar.qml
+++ b/src/appshell/qml/MainToolBar.qml
@@ -61,7 +61,7 @@ Item {
         id: navPanel
         name: "MainToolBar"
         enabled: root.enabled && root.visible
-        accessible.name: qsTrc("appshell", "Main tool bar") + " " + navPanel.directionInfo
+        accessible.name: qsTrc("appshell", "Main toolbar") + " " + navPanel.directionInfo
     }
 
     RadioButtonGroup {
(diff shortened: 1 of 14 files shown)


Wim4rk (approved):

cbjeukendrup:
About those Qt translation files, I would prefer to not put them in the repo. Instead, we should just install the qm files directly from the Qt installation (on Windows, `C:\Qt\5.15.2\msvc_64\translations`, or something like that). Would it be okay to remove this change from this PR for now, so that we can just merge it, and think about this later?

Jojo-Schmitz:
Well, the point of adding those to the repo is that they are __not__ available (or in a broken form, like 2 of them) from the Qt installation and are on top of those.

Jojo-Schmitz:
But yes, I can put that into a separate PR if need be.

Jojo-Schmitz:
Done

shoogle (approved):

cbjeukendrup (approved):

PR #12650, opened 2022-08-02
Grace notes improvements and bug-fixes 

mike-spa:
Grace note spacing was not taking into account staff moves, resulting in a quite dramatic bug.
Before:
graceNoteBugBEFORE
After:
graceNoteBugAFTER


cbjeukendrup:
This also seems like a good candidate for a VTest. Let's make a habit of adding a VTest for every PR where it might be applicable. This way, we build a strong test suite over time, that will protect us against regressions. 

I found another issue with cross staff grace notes. In this example, the first four eighth notes are moved to the staff below, and then, for each of them, I added a grace note before and a grace note after. The grace notes appear on the upper staff, but I think it would be more ideal if they inherit the `staffMove` from the "main" note by default. Also, remarkably, the grace note after appears _before_ the grace note before, except for the fourth one, which appears at the very end of the measure. (To check which one is which, I select them and check the status bar.)
Schermafbeelding 2022-08-02 om 19 11 09

cbjeukendrup:
... And if I finally select one of the main notes and move them back to the staff above, the same bug as before the PR returns...
Schermafbeelding 2022-08-02 om 19 20 42

mike-spa:
Wtf. Grace notes after are laid out before the main note, also on current master, and not only in moved staves. That's a big regression, no idea how it started, or how it went unnoticed.

mike-spa:
@oktophonie @cbjeukendrup this PR has become significantly bigger than when it started. Now grace notes should be much more solid though. From my side, I think it is ready for review/testing. From a first look, the vtests show exactly the correction they are supposed to. Tomorrow I'll make new ones specific for the bug I described at the beginning, and also for the cross-staff spacing.

Jojo-Schmitz:
I guess a `vtests` label is in order for this PR?

oktophonie (approved):

RomanPudashkin (approved):

PR #12651, opened 2022-08-02
[MU4] Fix MinGW build

Jojo-Schmitz:


src/framework/audio/internal/platform/win/wincoreaudiodriver.cpp
```diff
@@ -27,6 +27,9 @@
 #include "windows.h"
 #include "audioclient.h"
 
+#if defined __MINGW32__ // hack against "undefined reference" when linking in MinGW
+#define INITGUID
+#endif
```
igorkorsukov, line 32: If this solution works, then I think it's better

Jojo-Schmitz: It does. The other is commented out, but would work too

igorkorsukov: I suggest keep this, and remove the second one (remove comments)

Jojo-Schmitz: OK, gimmi some secs...

Jojo-Schmitz: done

PR #12652, opened 2022-08-02
[MU4] Fix #333979: Crash when changing time signature on a 'hanging' slur

Jojo-Schmitz:
Resolves: *https://musescore.org/en/node/333979*

Same issue (and fix) in MU3. Problem happens when there is a time sig change getting applied and later there is another time sig change already with a slur crossing from before into after that, further than to just the 1st beat after the change.
This change fixes the crash, said slur however is getting lost in the process.


Jojo-Schmitz:
Any reason not to merge this?

Jojo-Schmitz:
@vpereverzev mind to review and merge?

Jojo-Schmitz:
Hmm, strange, I didn't close it, but it seems not needed anymore, after a rebase?

cbjeukendrup:
Apparently @asattely made the same change in b444d36e7f533295eb2fe5f62f2242c2fadedfb7

Jojo-Schmitz:
Yep, meanwhile I found that out too ;-)

PR #12660, opened 2022-08-03
Add refresh_token to logout request to musescore api

bentl:
Include refresh token as query param to logout request to musescore api

### DIFF
--- a/src/cloud/internal/cloudservice.cpp
+++ b/src/cloud/internal/cloudservice.cpp
@@ -271,6 +271,10 @@ void CloudService::signOut()
 
     QUrl signOutUrl = prepareUrlForRequest(configuration()->loginApiUrl());
     if (!signOutUrl.isEmpty()) {
+        QUrlQuery query(signOutUrl.query());
+        query.addQueryItem(REFRESH_TOKEN_KEY, m_refreshToken);
+        signOutUrl.setQuery(query);
+
         QBuffer receivedData;
         Ret ret = m_networkManager->get(signOutUrl, &receivedData, headers());
 


src/cloud/internal/cloudservice.cpp
```diff
@@ -270,6 +270,9 @@ void CloudService::signOut()
     TRACEFUNC;
 
     QUrl signOutUrl = prepareUrlForRequest(configuration()->loginApiUrl());
+    QUrlQuery query(signOutUrl.query());
+    query.addQueryItem(REFRESH_TOKEN_KEY, m_refreshToken);
+    signOutUrl.setQuery(query);
```
cbjeukendrup, line 275: Wouldn't this conflict with what `prepareUrlForRequest` does? At least it overrides the query that is set by `prepareUrlForRequest`, because as far as I know, `setQuery` overrides any existing query.

src/cloud/internal/cloudservice.cpp
```diff
@@ -270,6 +270,9 @@ void CloudService::signOut()
     TRACEFUNC;
 
     QUrl signOutUrl = prepareUrlForRequest(configuration()->loginApiUrl());
```
cbjeukendrup, line 272: Actually, a bit confusing that the method is called `loginApiUrl` while it returns the logout URL...

src/cloud/internal/cloudservice.cpp
bentl: Haven't found other option to add additional query parameter to the url. I decided to get all query parameters from the url to QUrlQuery and add additional parameter. Request should not lose any data.

src/cloud/internal/cloudservice.cpp
bentl: Yeap, it needs to be fixed as well. I decided to not include refactoring to this PR.

src/cloud/internal/cloudservice.cpp
cbjeukendrup: Ah, sorry, I missed that you copy the existing query from the URL: <code>QUrlQuery query(<b>signOutUrl.query()</b>);</code>
So that's fine :)

The remaining thing I wonder about, is whether these three added lines should be inside the `if (!signOutUrl.isEmpty()) {` statement. The URL created by `prepareUrlForRequest` will be empty if the `m_accessToken` is empty, because in that case it does not make sense to proceed with the request. But if we add the refresh token to the query of the URL, it will not be empty anymore, so we try to do a request with a URL that only contains a query but for the rest nothing. That's not good. So I think we should only add the refresh token if the URL returned by `prepareUrlForRequest` was not empty.

bentl: Understood. I've moved it inside the `if`

cbjeukendrup (approved):

PR #12675, opened 2022-08-03
Fix uploading translations from Transifex to S3

cbjeukendrup:
After some reports on Discord, I took a look and it appears that the tx2s3 workflow has been silently failing for quite a while, so language updates were not working. This PR should fix it.

It looks like this has to do with https://github.com/musescore/MuseScore/pull/12447, although I don't understand how we missed the failure if it's true that it started happening with that PR.


igorkorsukov (approved):

igorkorsukov:
oops

PR #12712, opened 2022-08-07
Init translation system early, before modules that use it

cbjeukendrup:
Some modules perform tasks during their initialization that rely on the translation system. For example, the loading of instrument names, and the creation of default ui themes. Thus, the translation system needs to be initialized before everything else. To facilitate this, we give modules a new "onPreInit" method, which is meant to do initialization steps that don't depend on anything else and need to be done very early.
We take the opportunity to make the Splash Screen translatable, and also to init UI module settings early, so that we can use user-chosen fonts in the Splash Screen. 

(Although this solution is not totally unstructured, it feels a bit ad-hoc; by design the modules should not depend on each other, but in practice they just do, and we work around that by adding yet another initialization step. If the graph of dependencies between initialization steps becomes more complicated in the future, we might consider a proper system with dependency management, so that initialization steps are always done in the correct order. However, such a system is bound to be complicated and introduce overhead, so I think that's not yet worth the trouble.) 

Edit: the part about `onPreInit` has been merged through another PR, so now we're simply utilising that.

I have doubts though whether it is good at all that instrument names are translated during loading and then stored in translated form. For now, (after this PR), that is totally fine, but if we ever want to support changing language without restarting the app, it's bad, and then we'll need to switch to TranslatableString. 

Similar concern about UI theme names, but even worse: there, the translated names are even stored in the configuration. This means that even if the user does restart the app after changing the language, the names of the themes will still appear as in the old language, until a factory reset. The idea behind this situation is that we want to keep open the technical possibility to support totally custom themes with custom names. However, names of built-in themes should always be properly translated, so I'll change that in the future. (Edit: this is being fixed by #13743)

### DIFF
--- a/src/appshell/view/internal/splashscreen.cpp
+++ b/src/appshell/view/internal/splashscreen.cpp
@@ -52,8 +52,6 @@ SplashScreen::SplashScreen()
     setAttribute(Qt::WA_TranslucentBackground);
     setSize(splashScreenSize);
 
-    // TODO: this is just to make it translatable, but translation won't
-    // actually work here because translation system not yet initialized
     m_message = qtrc("appshell", "Loading…");
 
     repaint();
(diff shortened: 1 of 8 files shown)


Jojo-Schmitz:
Is this now initialized early enough for the Usage message, so we chould/should make that translatable too?

cbjeukendrup:
Rebased to fix conflicts.

cbjeukendrup:
@igorkorsukov It would be nice to merge this, since it fixes the problems mentioned in https://github.com/musescore/MuseScore/issues/13742#issuecomment-1277935227.

PR #12713, opened 2022-08-07
Apply fixing instrument IDs when reading legacy scores, instead of during migration

cbjeukendrup:
This is functionally crucial, so it needs to happen always, not only when "migrating" a score through the UI but always when reading a score, and not only for the MasterScore, but also for parts.

### DIFF
--- a/src/engraving/rw/compat/read302.h
+++ b/src/engraving/rw/compat/read302.h
@@ -25,6 +25,7 @@
 #include "engravingerrors.h"
 
 namespace mu::engraving {
+class Instrument;
 class MasterScore;
 class Score;
 
@@ -41,6 +42,8 @@ public:
 
 private:
     static bool readScore302(Score* score, XmlReader& e, ReadContext& ctx);
+
+    static void fixInstrumentId(Instrument* instrument);
 };
 }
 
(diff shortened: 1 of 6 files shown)


src/engraving/rw/compat/read302.h
```diff
@@ -39,6 +40,8 @@ class Read302
 
     static Err read302(MasterScore* masterScore, XmlReader& e, ReadContext& ctx);
 
+    static void fixInstrumentId(Instrument* instrument);
```
RomanPudashkin, line 43: This is a very strange place for this function. I would suggest making it private for Instrument and calling it from Instrument::updateInstrumentId()

cbjeukendrup: I think it's not _so_ strange, since this function is purely related to reading of 3.6 scores. I would not like to "clutter" the Instrument class with auxiliary single-use methods like this.
However, while thinking about your comment, I realized that this applies to 3.6.x scores _only_, and not to older scores (because for older scores, there were no instrument ids at all). So I'll change that. That hopefully improves the situation :)

RomanPudashkin (approved):

PR #12715, opened 2022-08-08
fix #12433: Text in Style Setting Dialog cut-off

### LINKED ISSUE #12433: [MU4 Issue] Text in Style Setting Dialog cut-off
**Describe the bug**
A short and clear description of what the bug is.

**To Reproduce**
1. Open Style Settings
2. Select Beams option

**Expected behavior**
Text and checkbox should be visible.
At the moment this is the only dialog I found having this issue. Other dialogs in style settings seem to be okay.

**Screenshots**
image

**Platform information**
 - OS: Windows 10

**Additional context**
Test on Nightly generated 15-Jul 04:53.

poypoyan:
Resolves: #12433

This is the easiest fix. ~May be there's a more "systematic" fix, something like depending on screen resolution, I don't know.~


cbjeukendrup (approved):
For now, this is okay. In the future, we could look for a more complicated 'automatic' solution; on the other hand, since the use of QQuickWidget is a temporary solution anyway (eventually, we'll rewrite the whole dialog in QML), that's not really worth it. So, thanks for the fix!

PR #12717, opened 2022-08-08
[MU4] Add (back) some missing qt_*.ts files

### LINKED ISSUE #12700: [MU4 Issue] Missing Qt translation
**Describe the bug**
Not all Qt button (and perhaps some dialog options) are translated to Dutch.
This might apply to other languages as well. This depends on the Qt version being used to compile the code.
Qt 5.x does not include the Dutch text file library.
For MS3 this was resolved by including the "qt_nl.ts" file in the locale folder (being distributed as .qm with the install).
I believe this file is even customized for some texts which are missing in some language files.

**To Reproduce**
1. Open MS4
2. Select Dutch (Nederlands) as the language
3. Load a Score or create a new one
4. Open a dialog which contain standard Qt buttons (i.e. Page Settings - Pagina Instellingen)
5. Check the buttons. Cancel should be Annuleer in Dutch and OK is normally written as Ok.

**Expected behavior**
Dialog buttons should showing in selected language.
NOTE: I believe to recall it was not only buttons in MS 3 (or might have been MS 2 even) which should English while using Dutch language. Some common dialog text messages would appear "untranslated" as well.

**Screenshots**
image

**Platform information**
 - OS: Windows 10 using MS4 Nightly dated: 5th Aug 2022

**Additional context**
As mentioned there are more languages in the MS3 share/locale folder.

Jojo-Schmitz:
that are not provided by Qt (or in a bogus form).
See e.g. fe849f3, 7067c5a, 8fb3961, 83b2689, 504b954 for some history when and why they got added/changed.

If Qt provides them, those from Qt take precedence, as they are getting copied later.

Fixes #12700


cbjeukendrup:
It looks like this PR will need to wait until #12917 is merged. After that, only the first commit of this PR will be necessary.

Jojo-Schmitz:
Yes, that 2nd commit is just a hack, I probably didn't even intend to push it.

Jojo-Schmitz:
Rebase and removed that 2nd commit

Jojo-Schmitz:
I don't see though why it would need to wait on #12917?

henkdegroot:
@cbjeukendrup will this be merged before the beta release?

RobFog:
A rebase is necessary.

henkdegroot:
@Jojo-Schmitz can you rebase this please?

Jojo-Schmitz:
Done, but I'm not clear how, when and where the `lrelease` is done nowadays, there these qt_*.ts files would need to get added, I think.

cbjeukendrup:
@Jojo-Schmitz The `gen-qt-projectfile` scripts are not in use anymore (you could delete them, and `gen-instruments-projectfile` too; otherwise I'll delete them in an upcoming cleanup PR).

Instead, see `tools/translations/run_lrelease.sh`.

Jojo-Schmitz:
Done

cbjeukendrup (commented):
This looks good to me now, but I'll ask @igorkorsukov to have a look too.

However, for Dutch, I don't see any difference (probably because Qt's files take priority over the manually added one). Also, I took a look at the added ts file for Dutch and I'm really shocked by the quality. This can't be right. Indeed, I can't find the mistakes in the translation files that currently come with Qt, so I think we'd better get rid of the manually added Dutch file. Probably the same goes for nl_BE.

henkdegroot:
@cbjeukendrup, if you check the Opmaak > Stijl dialog you should notice that this PR fixes the OK and Cancel button (as shown in the currently nightly) vs Ok and Annuleren when using the artifact generate by this PR.

image

cbjeukendrup:
@henkdegroot That's strange... for me on macOS it is "OK" and "Annuleren" anyway, even without this PR... maybe that suggests something is wrong on Windows specifically?

henkdegroot:
@cbjeukendrup maybe localization on macOS is handled differently for these default buttons?

igorkorsukov (approved):

igorkorsukov:
can we merge this pr? @cbjeukendrup do you have any objections?

cbjeukendrup:
I'm happy with it, let's merge it!

PR #12727, opened 2022-08-08
[MU4] Small adjustments to beam placement / slope

asattely:
image
This beam's start position is in a 'floater' position, but was still considered outside the staff, so it was fine for pointers. Now, that is an invalid beam position for both pointers and dictators:
image

Additionally, the horizontal length of the beam required for certain slope constraints have been changed to be more forgiving and slightly slantier.

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -1138,12 +1138,12 @@ void Beam::offsetBeamWithAnchorShortening(std::vector<ChordRest*> chordRests, in
     int newDictator = dictator;
     int newPointer = pointer;
     int reduce = 0;
-    while (!isValidBeamPosition(newDictator, isStartDictator, isAscending, isFlat, staffLines, beamCount)) {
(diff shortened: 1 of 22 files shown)


src/engraving/libmscore/beam.cpp
```diff
@@ -1138,12 +1138,12 @@ void Beam::offsetBeamWithAnchorShortening(std::vector<ChordRest*> chordRests, in
     int newDictator = dictator;
     int newPointer = pointer;
     int reduce = 0;
-    while (!isValidBeamPosition(newDictator, isStartDictator, isAscending, isFlat, staffLines, beamCount)) {
+    while (!isValidBeamPosition(newDictator, isStartDictator, isAscending, isFlat, staffLines, false)) {
```
RomanPudashkin, line 1141: I would suggest using a global static constant (you can declare it at the top of the file, and make sure that it will be declared as constexpr) instead of just "false" here in other places

asattely: Did you one better and just removed it from that function entirely, since the main issue I was trying to solve only needed that value for `isBeamInsideStaff`

oktophonie (approved):

PR #12733, opened 2022-08-09
[MU4] Fix assert fail caused by unreliable tick counts

asattely:
The fix for https://github.com/musescore/MuseScore/pull/12214 relied on integer ticks for its calculations, which are unreliable when dealing with extremely small fractions (due to rounding errors), so now it uses Fractions, which are more reliable, the whole way through.


Jojo-Schmitz (approved):
I believe to have stumbled across that assert myself quite a few times without ever really fully understandig why (and even less how to fix), but your code change certainly makes sense.

vpereverzev (approved):

PR #12737, opened 2022-08-09
fixed #12128: Scrolling through long list of plugins causes the context menu to close

### LINKED ISSUE #12128: [MU4 Issue] Scrolling through long list of plugins causes the context menu to close
**Describe the bug**
When there are lot of options, attempting to scroll using the context menu causes the context menu to close, so you cannot access many of the plugins.

This is a bug found using Marek's PC.. who has 700+ plugins installed!
For a user that has lots of plugins, being able to navigate and try to scroll down becomes difficult to do as demonstrated here:
This issue is happening not just in master but any Audio FX slot too. 

https://user-images.githubusercontent.com/71656189/175028584-065241f3-ac76-447e-af6a-6c05fee8671d.mp4

Eism:
Resolves: #12128


jessjwilliamson:
This has been fixed :) Thank you so much!

PR #12738, opened 2022-08-09
Fix translatability of line styles in EditStyle

cbjeukendrup:
Resolves: https://discord.com/channels/818804595450445834/818852942882275379/1006444044479631400

### DIFF
--- a/src/notation/view/widgets/editstyle.cpp
+++ b/src/notation/view/widgets/editstyle.cpp
@@ -28,6 +28,7 @@
 #include <QSignalMapper>
 
 #include "translation.h"
+#include "types/translatablestring.h"
 
 #include "alignSelect.h"
 #include "colorlabel.h"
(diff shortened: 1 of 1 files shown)


igorkorsukov (approved):

src/notation/view/widgets/editstyle.cpp
```diff
@@ -172,13 +173,13 @@ class EditStyle::LineStyleSelect : public QObject
...
+static const TranslatableString lineStyles[] = {
     //: line style
-    QT_TRANSLATE_NOOP("notation/editstyle", "Continuous"),
+    TranslatableString("notation/editstyle", "Continuous"),
     //: line style
-    QT_TRANSLATE_NOOP("notation/editstyle", "Dashed"),
+    TranslatableString("notation/editstyle", "Dashed"),
     //: line style
-    QT_TRANSLATE_NOOP("notation/editstyle", "Dotted"),
+    TranslatableString("notation/editstyle", "Dotted"),
```
Jojo-Schmitz, line 182: There are still a few `QT_TRANSLATE_NOOP` elsewhere in the code, is that a problem?

cbjeukendrup: Not per se. It becomes a problem at the moment that we forget to sync the context passed to `qtrc` with the context specified in `QT_TRANSLATE_NOOP`.

Jojo-Schmitz: How to tell one from the other?

cbjeukendrup: The contexts must simply match. But it may be difficult to track down where a string ends up being used, so if you update the context in the `QT_TRANSLATE_NOOP`, you will have to find out manually where that string can go and where it is translated using `qtr`, and update the context there too.
(Is that what you mean?)

Jojo-Schmitz: Yeah, I guess that's what I meant

PR #12752, opened 2022-08-10
Shortcut Categories

VanSHOE:
This adds shortcut category as defined [here](https://musescore.org/en/user/3638616/blog/2022/06/01/gsoc22-global-shortcuts#:~:text=Categorizing%20shortcuts%20in%20shortcut%20preferences)
image

The shortcut categories are currently draft categories and may change later on.
A known issue is that the opening and closing of categories can be a bit stuttery when opening/ closing categories


src/framework/ui/uitypes.h
```diff
@@ -187,47 +189,89 @@ enum class Checkable {
...
+    Dialogspanels,
+    Noteinput,
+    Workspace,
+    Plugins
+};
+
 struct UiAction
 {
+    inline static QStringList categories
+        = { "Undefined", "Internal", "Tablature", "Viewing & Navigation", "Playback", "Layout & Formatting", "Selecting & Editing",
```
shoogle, line 217: Lists values should be written out on separate lines like the enum values above to make for easier reading, scrolling and commenting. They also need to be translated:
```c++
{
    TranslatableString("framework/actioncategory", "Undefined"),
    TranslatableString("framework/actioncategory", "Internal"), // Now I can comment on this value
    TranslatableString("framework/actioncategory", "Tablature"),
    // TranslatableString("framework/actioncategory", "Viewing & Navigation"), // And comment out this value altogether
    ...
}
```
But I'm not sure if these strings should really be in a static list or just return values of a switch statement? (Maybe another dev can comment on this?) Also, they should probably be defined in a `.cpp` file along with the body of the `getCategory()` function.

src/framework/ui/uitypes.h
```diff
@@ -187,47 +189,89 @@ enum class Checkable {
...
              Checkable ch = Checkable::No)
-        : code(code), uiCtx(ctx), scCtx(scCtx), title(title), iconCode(icon), checkable(ch) {}
+        : code(code), category(cat), uiCtx(ctx), scCtx(scCtx), title(title), iconCode(icon), checkable(ch) {}
 
     bool isValid() const
     {
         return !code.empty();
     }
 
+    QString getCategory() const
```
shoogle, line 258: I would call this function `categoryName()`.  But I'm not sure it really belongs here. It should probably be defined more like this:
```c++
QString name(ActionCategory cat) {
    // code
}
```
And live outside the UiAction struct. It should be declared in a `.h` file and defined in a `.cpp` file.

shoogle (commented):
Works pretty well with the mouse, although it breaks accessibility with the keyboard. Try navigating the Preferences > Shortcuts page with the Tab key and arrow keys on your keyboard and you'll see what I mean (compare the behaviour of your PR to the latest nightly builds on master). Ideally the Spacebar and Return keys would be used to expand categories like in the Palettes.

There's no stuttering for me when expanding/collapsing the categories, though the text seems to remain visible for a few milliseconds longer than I would expect. The text should needs to disappear before the rectangles are fully collapsed. Perhaps the should lose opacity gradually rather than instantly disappearing? Or maybe the text could be gradually covered by the rectangle below?

Category names should be in **bold** according to the design. I would also suggest using separate columns for icons and action names to avoid the left edge of action names being staggered like this...

image

Icons should be centred in the first column and action names left-aligned in the next column.

bkunda:
This is coming along really nicely @VanSHOE!
I would suggest removing the accordion animation altogether – then there should be no stuttering/lag at all. We don't have animations like this elsewhere in the app, so it would work just fine I think to just have the category contents appear/disappear when their menu header is opened/closed. 
I do need to finesse and proof some of these category definitions. I already spotted one error (certainly my fault), and I expect there will be others. But actually having them "working" now and appearing as they should in the UI is going to help tremendously with this. Thank you so much!

src/framework/ui/uitypes.h
```diff
@@ -187,9 +189,53 @@ enum class Checkable {
...
+    MUSICALSYMBOLS,
+    DIALOGSPANELS,
+    NOTEINPUT,
+    WORKSPACE,
+    PLUGINS
+};
+
 struct UiAction
 {
+    inline static std::vector<TranslatableString> categories
```
cbjeukendrup, line 216: I'm inclined to suggest moving this vector to ShortcutsModel, since that is the only place where it is used and I think this is rather UI-related than logic-related so might fit better into a UI-related class.

VanSHOE: Sure I'll move it there!

VanSHOE:
@bkunda I see, I will remove the animation
However, the opening/ closing still will not be instant though it is better than an animation that stutters

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -163,6 +188,15 @@ QVariant ShortcutsModel::currentShortcut() const
...
 }
 
+QString ShortcutsModel::getCategory(ui::ActionCategory category) const
+{
+    if ((int)category + 1 >= 0 && (int)category + 1 < categories.size()) {
+        return categories[(int)category + 1].qTranslated();
+    }
+//    LOGE() << "Categories: " << categories.size() << "| Requested Category: " << (int)category << " by: " << code;
+    return "NULL";
+}
```
shoogle, line 198: Thanks for moving the function, but why did you go back to using C-style casts? It should be:
```C++
    QString getCategoryName(ui::ActionCategory category) const
    {
        size_t c = static_cast<int>(category) + 1;
        IF_ASSERT_FAILED(c < categories.size()) {
            return QString();
        }
        return categories[c].qTranslated();
    }
```

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -188,6 +188,15 @@ QVariant ShortcutsModel::currentShortcut() const
     return shortcutToObject(sc);
 }
 
+QString ShortcutsModel::getCategory(ui::ActionCategory category) const
+{
+    if ((int)category + 1 >= 0 && (int)category + 1 < categories.size()) {
+        return categories[(int)category + 1].qTranslated();
+    }
```
cbjeukendrup, line 195: It might be better to do this method as a switch statement: 
```c++
switch (category) {
case ...: return ...;
...
}

// Unreachable
return {};
```
This way, you don't need to worry about syncing the order in the enum and in this method, and you'll get a compiler warning here if you forget to handle new cases here.

Also, maybe better rename the method to `categoryName(...)` (we tend not to use `get` prefixes).

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -57,11 +58,32 @@ QVariant ShortcutsModel::data(const QModelIndex& index, int role) const
         return QString::fromStdString(action.code) + action.title.qTranslated() + action.description.qTranslated()
                + sequencesToNativeText(shortcut.sequences);
     }
+    case RoleSection:
+        return SectionName(shortcut);
+    case RoleSectionValue:
+        return SectionName(shortcut).toString() + sequencesToNativeText(shortcut.sequences);
```
cbjeukendrup, line 64: In SectionName, you convert from QString to QVariant and here you're converting back. That could be more efficient. Just return QString in SectionName.

src/framework/shortcuts/view/shortcutsmodel.cpp
```diff
@@ -57,11 +58,32 @@ QVariant ShortcutsModel::data(const QModelIndex& index, int role) const
...
 
     return QVariant();
 }
 
+const QVariant ShortcutsModel::SectionName(const Shortcut& shortcut) const
+{
+    return this->getCategory(this->action(shortcut.action).category);
+}
+
+QStringList ShortcutsModel::getSections() const
```
cbjeukendrup, line 77: A few code style comments in this area:
- `SectionName` should start with a lowercase `s`
- I don't like unnecessary uses of `this->` because it decreases readability (makes a simple thing look complicated)
- `getSections` preferably without `get` prefix (and then also with lowercase `s`)

cbjeukendrup (commented):
I think you can actually get rid of `getCategory` and just do the stuff in `ShortcutsModel::sectionName(...)`.

src/framework/shortcuts/view/shortcutsmodel.cpp
VanSHOE: Ah, so sorry, must have missed it during handling a merge conflict

src/framework/shortcuts/view/shortcutsmodel.cpp
VanSHOE: Sure Ill incorporate this and above suggestions

src/framework/shortcuts/internal/shortcutsregister.cpp
```diff
@@ -32,6 +34,7 @@
 using namespace mu::shortcuts;
 using namespace mu::framework;
 using namespace mu::async;
+using namespace mu::ui;
```
cbjeukendrup, line 37: Are the changes in this file fix-ups for previous changes, or are they left-overs of uncommitted experiments?

src/framework/shortcuts/qml/MuseScore/Shortcuts/ShortcutsPage.qml
```diff
@@ -54,6 +54,8 @@ Item {
     Component.onCompleted: {
         shortcutsModel.load()
 
+        shortcutsView.allSections = shortcutsModel.getSections()
+        console.log("All sections are: " + shortcutsView.allSections)
```
cbjeukendrup, line 58: (please remove before merge)

src/framework/shortcuts/view/shortcutsmodel.h
```diff
@@ -51,15 +51,38 @@ class ShortcutsModel : public QAbstractListModel, public async::Asyncable
...
+            TranslatableString("framework/actioncategory", "File"),
+            TranslatableString("framework/actioncategory", "Selection & Navigation"),
+            TranslatableString("framework/actioncategory", "Text & Lyrics"),
+            TranslatableString("framework/actioncategory", "Chord symbols & figured bass"),
+            TranslatableString("framework/actioncategory", "Measures"),
+            TranslatableString("framework/actioncategory", "Musical Symbols"),
+            TranslatableString("framework/actioncategory", "Dialogs & Panels"),
+            TranslatableString("framework/actioncategory", "Note Input"),
+            TranslatableString("framework/actioncategory", "Workspace"),
+            TranslatableString("framework/actioncategory", "Plugins") };
```
cbjeukendrup, line 73: Not sure if this is the best translation context. @shoogle What do you think? Maybe `"action/category"`, to keep it consistent with `"action"` which is the context of UIActions?
(probably "framework/" is not right anyway because it's not part of the framework module anymore)

src/framework/ui/uitypes.h
```diff
@@ -27,6 +27,7 @@
 #include <cstring>
 #include <QString>
 #include <QMetaType>
+#include <set>
```
cbjeukendrup, line 30: This one seems unnecessary

src/framework/ui/uitypes.h
```diff
@@ -187,9 +189,32 @@ enum class Checkable {
...
+    FILE,
+    SELECTIONNAVIGATION,
+    TEXTLYRICS,
+    CHORDSYMBOLSFIGUREDBASS,
+    MEASURES,
+    MUSICALSYMBOLS,
+    DIALOGSPANELS,
+    NOTEINPUT,
+    WORKSPACE,
+    PLUGINS
```
cbjeukendrup, line 211: Although there's some inconsistency within the project, I think in new code it is more common to use CamelCase. It has may preference, because it looks more friendly and I think it's better readable.

Also, in any case, the words should be separated, e.g. `VIEW_NAVIGATION` or `ViewNavigation`, `LAYOUT_FORMATTING` or `LayoutFormatting`, etc.

src/framework/uicomponents/qml/MuseScore/UiComponents/internal/ValueListSection.qml
```diff
@@ -0,0 +1,78 @@
...
+                font: ui.theme.bodyBoldFont
+                text: "section"
+                horizontalAlignment: Text.AlignLeft
+            }
+        }
+
+
+    }
+
+
```
cbjeukendrup, line 77: (please fix code style before merge)

src/framework/ui/uitypes.h
```diff
@@ -35,6 +36,7 @@
 #include "view/iconcodes.h"
 #include "shortcuts/shortcutstypes.h"
 #include "context/shortcutcontext.h"
+#include "log.h"
```
cbjeukendrup, line 39: This one also seems unnecessary now that `getCategory` has been moved

src/framework/uicomponents/qml/MuseScore/UiComponents/ValueList.qml
```diff
@@ -175,6 +209,50 @@ Item {
...
+                console.log("Hiding: " + section)
+                hideSection(section)
+            } else {
+                console.log("Opening: " + section)
+                showSection(section)
+            }
+        }
+
+        function isSectionExpanded(section) {
+            console.log("Checking " + section +  ": " + (section in collapsed))
```
cbjeukendrup, line 224: (please remove the logging statements before merge)

src/framework/shortcuts/qml/MuseScore/Shortcuts/internal/ShortcutsBottomPanel.qml
```diff
@@ -81,6 +83,22 @@ RowLayout {
 
     Item { Layout.fillWidth: true }
 
+    FlatButton {
+        id: toggleVisible
+
+        Layout.preferredWidth: root.buttonWidth
+
+        text: qsTrc("shortcuts", "Toggle Visibility")
```
cbjeukendrup, line 91: Should be sentence case (`"Toggle visibility"`)

cbjeukendrup:
I have a feeling that we should think a little bit about how the tree view is implemented in QML. Currently, we're basically abusing a ListView and setting the height of invisible items to zero. This has bad implications for keyboard navigation (it will treat the "hidden" rows as visible, because technically they are).

You could try to use QML's TreeView component (rather than ListView). TreeView is from `QtQuick.Controls 1`, and thus has a totally different idiom than ListView (it's a bit more complicated). 

It would be quite a big operation, because it would mean that the ShortcutsModel will also need to become a QAbstractItemModel rather than a QAbstractListModel. 

I'm not sure whether this is worthwhile. I can't predict whether this will give better performance or not at all. On the other hand, one could argue that it is a lot cleaner. Maybe @shoogle and other developers could advise about this?

(One other note about this: Qt 6.0-6.2 did not contain a TreeView component at all, but since Qt 6.3 there's a new one, which is more in ListView style, and works completely different from the `QtQuick.Controls 1` one at the QML side. So, using TreeView now will mean having to implement it when we switch to Qt 6.)

VanSHOE:
@cbjeukendrup Keyboard navigation with respect to the items was fixed, i.e. if you collapse them they wont be selectable with keyboard anymore.

The issue however is that sections themselves can't be selected with a keyboard

cbjeukendrup:
Ah, missed that, my fault. So that's one less problem to worry about...

shoogle:
> using TreeView now will mean having to implement it when we switch to Qt 6

If the Qt 5 TreeView is implemented very differently to the Qt 6 TreeView then I think it would be better to stick with the current approach (i.e. the hacked ListView) for as long as we use Qt 5, and only think about using a proper TreeView after we have switched to Qt 6.

> sections themselves can't be selected with a keyboard

Of course if this is not solvable then that might be an argument for using a proper TreeView now, but hopefully it will be solvable one way or another.

src/framework/shortcuts/view/shortcutsmodel.h
VanSHOE: Please let me know if this is supposed to be done. Till then I'll push some of the other changes

src/framework/shortcuts/view/shortcutsmodel.cpp
VanSHOE: @cbjeukendrup I can do this however, the categories vector is required to send all sections to QML as well. It is in `ShortcutsPage.qml`
```qml
  Component.onCompleted: {
      shortcutsModel.load()

      shortcutsView.allSections = shortcutsModel.sections()
      topPanel.setSearchText(root.shortcutCodeKey)
  }
```

So I thought maybe the get category could just use this list itself to preserve Lines of Code. Please let me know if I should still go with the switch case approach in section name

cbjeukendrup:
Hi @VanSHOE! We would like to revisit this PR, and we are planning to include it in MuseScore 4.1. The first thing that's necessary for that is a rebase, both to fix the conflicts and to get new artifacts so that we can test it again (the old builds have expired). It would be great if you could do that!

src/framework/ui/uitypes.h
```diff
@@ -187,9 +187,32 @@ enum class Checkable {
     Yes
 };
 
+enum class ActionCategory {
+    Undefined = -1,
+    Internal,
```
cbjeukendrup, line 192: It looks like the `Internal` category is not really used; its role already seems to be fulfilled by `Undefined`. Do we need it?

VanSHOE:
@cbjeukendrup 

Sure Im on it!

shoogle:
I'm archiving this for now, but we should take another look if we decide to split Preferences > Shortcuts into categories.

I note that Qt 6 now has a [QML TreeView](https://doc.qt.io/qt-6/qml-qtquick-treeview.html) component that we could possibly use for this purpose.

However, we still have our own [LegacyTreeView](https://github.com/musescore/MuseScore/tree/a31436c662afff058b0322657a7a4f4d35ad3df7/src/framework/uicomponents/qml/Muse/UiComponents/LegacyTreeView) component. I'm not sure why. Perhaps the Qt6 TreeView isn't suitable, or maybe we just haven't got around to porting our existing UI components to use it rather than our LegacyTreeView?

cbjeukendrup:
New code should definitely use the new QML TreeView. Only in the Instruments panel, the replacement is blocked by the issue mentioned in https://github.com/musescore/MuseScore/pull/29443. In other places, one could go ahead and replace the legacy tree view.

Issue #12774, opened 2022-08-10
[MU4 Issue] Long tooltips in palettes don’t break


akulisch:
**Describe the bug**
Longer tooltips of symbols in the palettes don’t have enough space to be fully within their box. 

**To Reproduce**
1. Switch to German (for the examples below; same probably also happens in other languages; maybe even English)
2. Hover mouse over (e.g.) “25 small diesis up, (25S, ~5:13S, ~37S, 5C plus 5C), 2° up [53 EDO]” or add the Accordion palette and hover over (e.g.) Right hand, 3 ranks, 4' stop + lower tremolo 8' stop + 8' stop + upper tremolo 8' stop + 16' stop

**Expected behavior**
Either the tooltip box needs its maximum width removed, or, preferably, line breaks have to be inserted.

**Screenshots**
![image](https://user-images.githubusercontent.com/15906252/184038832-1707e6de-0116-40d3-aa8c-dcd967b767dd.png)
![image](https://user-images.githubusercontent.com/15906252/184038981-bbc4b7e7-056a-4be9-8a63-d53dd047c4fc.png)
![image](https://user-images.githubusercontent.com/15906252/184039063-87382362-f0ef-42cb-9eed-ca98bd5359f1.png)

**Platform information**
- OS: Windows 10 Pro 10.0.19044 Build 19044
- MuseScore Version: v4.0_alpha2

Tantacrul:
Yes, this is super annoying. It would be great to simply have a character limit per line and then bump it down to a second line when needed.

abariska:
FYI @Tantacrul and all. 
Comment from @cbjeukendrup:

> I fixed it for QML tooltips in general, so it should work for all of them, in Palettes, Inspector, Note Input Toolbar…
> However, the master palette is implemented in Qt Widgets. It uses Qt’s tooltip implementation, so it might be different from other tooltips, but we can’t influence that unfortunately.

So, there's no word cutting for now in Master palette but long words still don't wrap. Could be fixed in future

<img width="1025" alt="CleanShot 2022-09-21 at 16 41 56" src="https://user-images.githubusercontent.com/101250347/191521834-7841f364-2fc8-4ee7-a88e-b1ef40e97ea6.png">

abariska:
🟢 **FIXED** in https://github.com/musescore/MuseScore/pull/12801

PR #12801, opened 2022-08-11
Some text wrap fixes

cbjeukendrup:
Resolves: #12774
Resolves: https://discord.com/channels/818804595450445834/818852942882275379/1007159359941394523

Enabling word wrap is nice, but when the width of the text field is not properly limited, it doesn't help much...

Also: we could consider replacing all uses of `Text.WordWrap` with `Text.Wrap`, because the latter allows wrapping in the middle of a word in "emergency" cases (word longer than available line length). The former would put words that are too long still on one line, and either clip them or let them just go outside the boundaries of the `StyledTextLabel`. I did not make this change here, but would be happy to do so if we decide it is desired.

### DIFF
--- a/src/appshell/qml/FirstLaunchSetup/Page.qml
+++ b/src/appshell/qml/FirstLaunchSetup/Page.qml
@@ -20,7 +20,6 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 import QtQuick 2.15
-import QtQuick.Layouts 1.15
 
 import MuseScore.Ui 1.0
 import MuseScore.UiComponents 1.0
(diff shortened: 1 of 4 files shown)


RomanPudashkin (approved):

abariska:
@cbjeukendrup I'm not sure it's a real case when word longer than current line length (after fix). So, I'd leave it as was fixed for now.

abariska:
🟢 **TESTED**
Please, merge to master

cbjeukendrup:
> @cbjeukendrup I'm not sure it's a real case when word longer than current line length (after fix). So, I'd leave it as was fixed for now.
> Please, add a note about what is the symbol limit for a line?

The maximum number of characters per line is not specified; instead, it is limited by the width. For tooltips, we specify a maximum width of 300px, for info/warning/error dialogs it is 420px.

cbjeukendrup:
Rebased. @RomanPudashkin, I think this PR is already ready to go (reviewed & tested)

Issue #12785, opened 2022-08-11
[MU4 Issue] Beams beginning with a rest are not displayed


bakajikara:
**Describe the bug**
Beams beginning with a rest are not displayed

**To Reproduce**
1. Enter notes
2. Change beam property of rest
3. See error
4. Save and reopen the score
5. See another error

See video:

https://user-images.githubusercontent.com/102702868/184100926-727c6a81-60e3-4a98-acef-83b49a320107.mp4

**Expected behavior**
Beams are displayed as in MU3: 
![スクリーンショット (38)](https://user-images.githubusercontent.com/102702868/184100375-9c5f9c5f-5a51-4aae-ab9d-e97ffc3beeb7.png)

**Platform information**
 - OS: Windows 11

**Additional context**
Checked with Public Alpha

oktophonie:
Thanks for the report; you'll be pleased to know we're actually already working on a fix for this at the moment (keep an eye out for a PR this week)!

PR #12802, opened 2022-08-11
Fix 12785: Beams starting or ending with rests

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12785

Beams starting or ending on rests are now rendered properly. This extends to cross-staff beams as well!


oktophonie (approved):

src/engraving/libmscore/beam.cpp
```diff
@@ -604,7 +604,9 @@ int Beam::getMaxSlope() const
     // maxSlopes[i] = max slope of beam for notes with interval i
 
     // calculate max slope based on distance between first and last chords
-    double beamWidth = _elements[_elements.size() - 1]->stemPos().x() - _elements[0]->stemPos().x();
+    double endX = chordBeamAnchorX(_elements[_elements.size() - 1], ChordBeamAnchorType::Start);
+    double startX = chordBeamAnchorX(_elements[0], ChordBeamAnchorType::End);
```
cbjeukendrup, line 608: Why are we using `ChordBeamAnchorType::Start` for `endX`, and `ChordBeamAnchorType::End` for `startX`?

PR #12803, opened 2022-08-11
[MU4] Fix #11617: Regression: tempo override control missing (formerly in Play Panel)

### LINKED ISSUE #11617: [MU4 Issue] Regression: tempo override control missing (formerly in Play Panel)
Previous versions of MuseScore had a slider to override the tempo of a piece, allowing you to play it back faster or slow than the marked tempo. This was used very extensively by people wanting to practice along with the playback, or by people wanting to listen more closely to each chord of the playback to be sure there are not wrong notes etc.  You could even turn it down to a ridiculous low and then use the arrow keys during playback to achieve a "scrubbing" effect.

Merely changing the initial tempo marking is not the same thing at all, since that won't affect any subsequent tempo changes.  The tempo slider worked across the entire piece.

image

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11617

### DIFF
--- a/src/engraving/libmscore/tempo.cpp
+++ b/src/engraving/libmscore/tempo.cpp
@@ -72,7 +72,7 @@ TempoMap::TempoMap()
 {
     _tempo    = 2.0;          // default fixed tempo in beat per second
     _tempoSN  = 1;
-    _relTempo = 1.0;
+    _tempoMultiplier = 1.0;
 }
 
 //---------------------------------------------------------
@@ -124,7 +124,7 @@ void TempoMap::normalize()
             e->second.tempo = tempo;
         }
         int delta = e->first - tick;
-        time += double(delta) / (Constants::division * tempo.val * _relTempo.val);
+        time += double(delta) / (Constants::division * tempo.val * _tempoMultiplier.val);
         time += e->second.pause;
         e->second.time = time;
         tick  = e->first;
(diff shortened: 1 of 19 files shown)


MarcSabatella:
❤️

DmitryArefiev:
Tested #11617 on Win10, Mac12, LinuxMint 20.2 - **FIXED**      

(please notice the tempo slider doesn't work on Windows - #12817)

Obliquely:
### Great to see this - and issue with loop playback and tempo% change
Great to see this... **and** there seems to be an issue with loop playback. If you adjust the Tempo% slider when you have a loop set, then, it will play notes beyond the end of the loop (if you increase Tempo over the starting %) or it will fail to play all the notes in the loop (if you increase Tempo % below starting %). (So, totally, there is a workaround - just unset and reset the loop after tempo change -  *and* setting a loop and adjusting the tempo is, I think, a common operation for people who might be using the Tempo slider at all.)

I'm guessing there needs to be some updating of the loop playback state when the Tempo% slider is moved.

Obliquely:
In MS 3.6.2 double clicking the equivalent of the tempo% slider (in the Play Panel) would restore the tempo% to 100%. I'm not sure if that double-click slider to reset to default is part of the design language of MS4. I think it's valuable and certainly regularly make use of it in MS 3.6.2 (though I think I implemented or fixed implementation in the Play Panel for 3.6.2 so I would say that 😀).

vpereverzev (approved):

PR #12824, opened 2022-08-12
[MU4] Fix #11494: Note input can start at location not on screen

### LINKED ISSUE #11494: [MU4 Issue] Note input can start at location not on screen
EDITED BY @Tantacrul 

We want to retain the 'no jumping' rule from MuseScore 3 when a user resumes Note Input mode. There are two points (which I've demonstrated in the video below):

1. When a user triggers note input mode, if the previous position where the user last inputted notes is out of view, **do not jump back to to that view** (This is by far the more important point. **_When this is achieved, we may close the issue_**)

2. If the previous position is not in view, then move the note input cursor to the **top-left visible stave**. If this is too hard / not achievable, we should discuss whether a delay until 4.x is the right move.

https://user-images.githubusercontent.com/21022311/174475265-41bd0179-6bb1-42da-8aa5-4e082ce015c0.mov

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11494

### DIFF
--- a/src/appshell/qml/NotationPage/NotationPage.qml
+++ b/src/appshell/qml/NotationPage/NotationPage.qml
@@ -419,7 +419,7 @@ DockPage {
         name: "MainNotationView"
 
         isNavigatorVisible: pageModel.isNavigatorVisible
-        isAccessibilityEnabled: true
+        isMainView: true
 
         Component.onCompleted: {
             root.notationView = notationView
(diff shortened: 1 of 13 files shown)


DmitryArefiev:
Tested #11494 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Issue #12769, opened 2022-08-10
[MU4 Issue] Text palet contains user selected language and english elements


henkdegroot:
**Describe the bug**
When using a different language, in my case Dutch, the Text palet is showing translated items. A very good thing!
However, when I click the More button the English versions of the same element is available.

**To Reproduce**
1. Select another language then English
2. Restart MS4
3. Open an existing or start a New Score
4. Check the Text palet

**Expected behavior**
The english equivalent of the element should not be shown in the more section.

**Screenshots**
Highlighted a few in the image, more exist.
![image](https://user-images.githubusercontent.com/13550012/184022874-4be188f5-042a-4378-ab52-caf575c94927.png)

**Platform information**
 - OS: Windows 10 Dutch

**Additional context**
Using the Alpha2 release

cbjeukendrup:
Does resetting the Text palette help in this case?

henkdegroot:
Tried that (twice) and did not make a difference

henkdegroot:
And also tested with Polski and got the same result.

cbjeukendrup:
@henkdegroot Can you still reproduce this? As @DmitryArefiev found while testing #12853, we can't reproduce this anymore in the latest master.

henkdegroot:
@cbjeukendrup, issue still exists in Beta...however I just downloaded the Nightly from this morning and cannot reproduce this anymore. So does seem something has fixed. Also happy to see Dutch in the splash screen (instead of Loading...) :)

DmitryArefiev:
Tested in latest master on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Closing

PR #12853, opened 2022-08-14
Take translatability into account when comparing palette cells

cbjeukendrup:
Resolves: #12769

Note: some elements still appear both in the palette itself and in the 'More' popup. This is because although they look the same, those elements don't produce the same MIME data, see `isEquivalent` in palette.cpp. So this is a separate issue, and it looks like it also occurred in MU3.

### DIFF
--- a/src/palette/internal/palette.cpp
+++ b/src/palette/internal/palette.cpp
@@ -220,8 +220,7 @@ static bool isEquivalent(const EngravingItem& e1, const EngravingItem& e2)
 
 int Palette::indexOfCell(const PaletteCell& cell, bool matchName) const
 {
-    const EngravingItem* el = cell.element.get();
-    if (!el) {
+    if (!cell.element) {
         return -1;
     }
 
@@ -231,8 +230,11 @@ int Palette::indexOfCell(const PaletteCell& cell, bool matchName) const
             continue;
         }
 
-        const EngravingItem* exElement = localCell.element.get();
-        if (exElement && !isEquivalent(*exElement, *el)) {
+        if (cell.untranslatedElement) {
+            if (!localCell.untranslatedElement || !isEquivalent(*localCell.untranslatedElement, *cell.untranslatedElement)) {
+                continue;
+            }
+        } else if (!localCell.element || !isEquivalent(*localCell.element, *cell.element)) {
             continue;
         }
 


src/palette/internal/palette.cpp
```diff
@@ -231,8 +230,11 @@ int Palette::indexOfCell(const PaletteCell& cell, bool matchName) const
             continue;
         }
 
-        const EngravingItem* exElement = localCell.element.get();
-        if (exElement && !isEquivalent(*exElement, *el)) {
+        if (cell.untranslatedElement) {
```
RomanPudashkin, line 233: Why do we need two instances of the same element? Or what is the difference between cell.untranslatedElement and cell.element?

cbjeukendrup: This is something specific for `TextBase` derived elements. For these elements, `element` is (as always) the element that is drawn in the palette and that gets added to the score, and `untranslatedElement` is just what the name might suggest: a copy of `element`, but with the untranslated text. 

If the cell is made translatable (i.e. `untranslatedElement` is not null), both `element` and `untranslatedElement` are written to disk when the PaletteCell is written to disk. When the user changes the language and restarts MuseScore, when reading the PaletteCell, we look at the text of `untranslatedElement`, and guess the translation context using the `translationContext` method. That way, we can reconstruct the translated `element`, so that `element` will always use the language that the user has currently selected. 

(Note: this is historical code. Or course, instead of storing a whole copy of the element, we could also just store the untranslated text as a string (or preferably even `TranslatableString`, once we have a way to write `TranslatableStrings` to XML). However, that solution would cause difficulties if we ever want to store an element in Palettes that contains _two_ translatable strings (for example, a text line with begin text and end text). So I propose to leave it as it is for now.)

RomanPudashkin (approved):

RomanPudashkin:

cbjeukendrup:
Done!

DmitryArefiev:
@cbjeukendrup I **can't reproduce** #12769 in master build (Windows10)

**Dutch** (Nederlands)

https://user-images.githubusercontent.com/90187801/197571680-96ee2575-c41f-4168-a39c-135d8009edf1.mp4

**Polski**

https://user-images.githubusercontent.com/90187801/197571963-2feeed20-9b11-4526-ae0c-d68b5b2ebe4d.mp4

cbjeukendrup:
@DmitryArefiev Hm, I can't reproduce it anymore either... maybe it got fixed as a side effect of something.

cbjeukendrup:
Closing, since the corresponding issue was magically fixed in master branch, maybe as a side effect of #12712.

PR #12861, opened 2022-08-15
[MU4] Fix some menus/shortcut strings

### LINKED ISSUE #12921: [MU4 Issue] Ampersand is present in Export and Save shortcuts name
**Expected behavior**
No ampersands shortcuts names

**Screenshots**
Capture 2022-08-18_02-54-51_PM copy

**Additional context**
Could be fixed along with https://github.com/musescore/MuseScore/issues/12922

### LINKED ISSUE #13164: [MU4 Issue] In Properties: Section break, "Measure" (US) has been replaced by "Bar" (UK)
**Describe the bug**

Even if US spelling is selected in Preferences, the word "Bar" is still used in Properties: Section break. i.e. "Reset bar numbers for new section".

Jojo-Schmitz:
i.e. differentiate them between the menu and the shortcuts. Esp. have the accelerator key only in the menu, not in the shortcut strings and also, for shortcuts have a better description (which in the Menu comes from their context, here: the File menu)

Changes/adds a few string from/to the translations

Fixes #12921
Fixes #13164, see also #11916
Related to #13295 too

### DIFF
--- a/src/appshell/view/publish/publishtoolbarmodel.cpp
+++ b/src/appshell/view/publish/publishtoolbarmodel.cpp
@@ -35,9 +35,9 @@ void PublishToolBarModel::load()
     AbstractMenuModel::load();
 
     MenuItemList items {
-        makeMenuItem("print"),
-        makeMenuItem("file-publish"),
-        makeMenuItem("file-export")
+        makeMenuItem("print", TranslatableString("action", "Print")),
+        makeMenuItem("file-publish", TranslatableString("action", "Publish to MuseScore.com")),
+        makeMenuItem("file-export", TranslatableString("action", "Export"))
     };
 
     setItems(items);
--- a/src/engraving/libmscore/fret.cpp
+++ b/src/engraving/libmscore/fret.cpp
@@ -1522,7 +1522,10 @@ String FretDiagram::screenReaderInfo() const
             }
         }
 
-        //: Omit the "%n " for the singular translation (and the "(s)" too)
+        //: Please provide a translation for the singular and all plural forms available in your language.
+        //: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
+        //: You may translate singular case to the equivalent of "one dot" or even just "dot" instead of the "1 dot" form, but only do this
+        //: if you are sure that the translation is only used for a single value of "%n".
         String dotsInfo = mtrc("engraving", "%n dot(s) on fret(s) %1", "", dotsCount).arg(fretInfo);
 
         detailedInfo = String(u"%1 %2 %3 %4").arg(detailedInfo, stringIdent, markerName, dotsInfo);
@@ -1565,6 +1568,10 @@ String FretDiagram::screenReaderInfo() const
 
     String basicInfo = String(u"%1 %2").arg(translatedTypeUserName(), chordName);
 
+    //: Please provide a translation for the singular and all plural forms available in your language.
+    //: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
+    //: You may translate singular case to the equivalent of "one string" instead of the "1 string" form, but only do this
+    //: if you are sure that the translation is only used for a single value of "%n".
     String generalInfo = mtrc("engraving", "%n string(s) total", "", _strings);
 
     String res = String(u"%1 %2 %3").arg(basicInfo, generalInfo, detailedInfo);
--- a/src/framework/global/dataformatter.cpp
+++ b/src/framework/global/dataformatter.cpp
@@ -44,21 +44,23 @@ String DataFormatter::formatTimeSince(const Date& date)
         return mtrc("global", "Today");
     }
 
-    if (days == 1) {
-        return mtrc("global", "Yesterday");
-    }
-
     if (days < 7) {
+        //: Please provide a translation for the singular and all plural forms available in your language.
+        //: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
+        //: Possible translations for the 0 case and any value > 6 are not used, yet mandatory.
+        //: You may translate singular case to the equivalent of "Yesterday" instead of the "1 day ago" form, but only do this
+        //: if you are sure that the translation is only used for a single value of "%n".
         return mtrc("global", "%n day(s) ago", nullptr, days);
     }
 
     int weeks = days / 7;
 
-    if (weeks == 1) {
-        return mtrc("global", "Last week");
-    }
-
     if (weeks <= 4) {
+        //: Please provide a translation for the singular and all plural forms available in your language.
+        //: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
+        //: Possible translations for the 0 case and any value > 6 are not used, yet mandatory.
+        //: You may translate singular case to the equivalent of "Last week" instead of the "1 week ago" form, but only do this
+        //: if you are sure that the translation is only used for a single value of "%n".
         return mtrc("global", "%n week(s) ago", nullptr, weeks);
     }
 
@@ -66,15 +68,21 @@ String DataFormatter::formatTimeSince(const Date& date)
 
     int months = (currentDate.year() - date.year()) * monthsInYear + (currentDate.month() - date.month());
 
-    if (months == 1) {
-        return mtrc("global", "Last month");
-    }
-
     if (months < monthsInYear) {
+        //: Please provide a translation for the singular and all plural forms available in your language.
+        //: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
+        //: Possible translations for the 0 case and any value > 6 are not used, yet mandatory.
+        //: You may translate singular case to the equivalent of "Last month" instead of the "1 month ago" form, but only do this
+        //: if you are sure that the translation is only used for a single value of "%n".
         return mtrc("global", "%n month(s) ago", nullptr, months);
     }
 
     int years = currentDate.year() - date.year();
 
+    //: Please provide a translation for the singular and all plural forms available in your language.
+    //: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
+    //: Possible translations for the 0 case and any value > 6 are not used, yet mandatory.
+    //: You may translate singular case to the equivalent of "Last year" instead of the "1 year ago" form, but only do this
+    //: if you are sure that the translation is only used for a single value of "%n".
     return mtrc("global", "%n year(s) ago", nullptr, years);
 }
--- a/src/project/internal/projectuiactions.cpp
+++ b/src/project/internal/projectuiactions.cpp
@@ -31,14 +31,14 @@ const UiActionList ProjectUiActions::m_actions = {
              mu::context::UiCtxAny,
              mu::context::CTX_ANY,
              TranslatableString("action", "&Open…"),
-             TranslatableString("action", "Open…"),
+             TranslatableString("action", "Open"),
              IconCode::Code::OPEN_FILE
              ),
     UiAction("file-new",
              mu::context::UiCtxAny,
              mu::context::CTX_ANY,
              TranslatableString("action", "&New…"),
-             TranslatableString("action", "New…"),
+             TranslatableString("action", "New"),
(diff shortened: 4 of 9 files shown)


cbjeukendrup:
@bkunda Could you please review these changes? (by checking the code changes; for each action, the first `TranslatableString` currently represents the title in the menus, and the second one represents the name in Preferences; if there's only one, that will be used in both places.)

(There are a couple of things that I feel unsure about:
- Do we want to use the word "Project"? If so, shouldn't we do it consistently? In most places, we currently use "New _score_" etc. instead.
- Mentioning "Project" in all these strings adds some verbosity, and I'm not sure whether it adds much clarity
- Especially about `"Save selection of project"` I'm not sure. Also (minor detail): `"Save a copy of project"` sounds suboptimal to me; I'd say either `"Save copy of project"` or `"Save a copy of the project"`.)

Jojo-Schmitz:
We do already hare "Project properties", so I thought to add that to the save, open, save as too, but I'm not too wild on this. But then is should get removed from the properties too, and called "Score properties", like in MU3

Jojo-Schmitz:
Main point of the PR is that neither the ellipsis not the & should show in the shortcuts list

henkdegroot:
@Jojo-Schmitz, when I was checking the Dutch language, I tried adding the & sign for the Export option in the menu. This then also added the & to the Export buttons in the GUI. I noticed the PR includes the & sign now for the Export "menu" option, so I believe the result will be that the button will show the & sign now.

I mean this is where I noticed the & appearing (not shown in this dialog as this is the Alpha2 release in English):
image

Will the code be updated to resolve this (i.e. remove the & from the string or use another string)?

Jojo-Schmitz:
Good question, I believe it won't, as there the shortcurt string (including the icon) seems to be getting used, but let me check

Jojo-Schmitz:
Mhh, the &s do show there, for Print and Export, not good

src/framework/global/dataformatter.cpp
```diff
@@ -44,37 +44,33 @@ String DataFormatter::formatTimeSince(const Date& date)
         return mtrc("global", "Today");
     }
 
-    if (days == 1) {
-        return mtrc("global", "Yesterday");
-    }
-
     if (days < 7) {
+        //: Make sure to translate singular and plural form and whatever else your language has here.
+        //: No need to handle the 0 case here, for singular best translate to the equivalent of "Yesterday")
```
Jojo-Schmitz, line 49: See [also https://github.com/musescore/MuseScore/pull/12996#discussion_r953625676 ff.

src/appshell/view/publish/publishtoolbarmodel.cpp
```diff
@@ -35,10 +35,10 @@ void PublishToolBarModel::load()
     AbstractMenuModel::load();
 
     MenuItemList items {
-        makeMenuItem("print"),
+        makeMenuItem("print", TranslatableString("action", "Print project")),
         // Todo(save-to-cloud): hidden until fully implemented
-        //makeMenuItem("file-publish"),
-        makeMenuItem("file-export")
+        //makeMenuItem("file-publish", TranslatableString("action", "Save project to cloud")),
```
Jojo-Schmitz, line 40: Note to self: conflicts with #13024, but would be easy to resolve

Jojo-Schmitz: Rebased

src/engraving/types/typesconv.cpp
```diff
@@ -906,7 +906,7 @@ static const std::vector<Item<TextStyleType> > TEXTSTYLE_TYPES = {
     { TextStyleType::TITLE,             "title",                TranslatableString("engraving", "Title") },
     { TextStyleType::SUBTITLE,          "subtitle",             TranslatableString("engraving", "Subtitle") },
     { TextStyleType::COMPOSER,          "composer",             TranslatableString("engraving", "Composer") },
-    { TextStyleType::POET,              "poet",                 TranslatableString("engraving", "Poet") },
+    { TextStyleType::POET,              "poet",                 TranslatableString("engraving", "Lyricist") },
```
Jojo-Schmitz, line 909: See also #13111, which has the same change

src/engraving/libmscore/fret.cpp
```diff
@@ -1522,7 +1522,8 @@ String FretDiagram::screenReaderInfo() const
             }
         }
 
-        //: Omit the "%n " for the singular translation (and the "(s)" too)
+        //: Make sure to translate singular and plural form and whatever else your language has here.
+        //: Best omit the "%n " for the singular translation (and the "(s)" too).
```
shoogle, line 1526: French uses the singular form for the `n = 0` case (they say the equivalent of "zero dot" rather than "zero dots") so in general it is not safe to tell people to omit the `%n` for any cases, though I appreciate we don't use the `n = 0` case here.

You could say this instead:
> Please provide a translation for the singular and all plural forms available in your language.
> See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms

And add an explanation on that page about when it is and isn't safe to use a "wordy" number.

src/framework/global/dataformatter.cpp
```diff
@@ -44,37 +44,36 @@ String DataFormatter::formatTimeSince(const Date& date)
...
 
-    if (days == 1) {
-        return mtrc("global", "Yesterday");
-    }
-
     if (days < 7) {
+        //: Make sure to translate singular and plural form and whatever else your language has here.
+        //: No need to handle the 0 case here, nor any value > 6,
+        //: for singular best translate to the equivalent of "Yesterday")
         return mtrc("global", "%n day(s) ago", nullptr, days);
```
shoogle, line 51: I think it is best to get a translation for all plural forms. Transifex won't let you save without it anyway.

It may be safer to state the singular case as:
> You may translate singular cases to the equivalent of "Yesterday" instead of the "1 day ago" form, but only do this if you are sure that the translation is only used for a single value of "%n".

And also give the link to that Wiki page.

---

FWIW, if we wanted to do this properly we would probably code it like this:
```C++
//: Translate as "[no-equivalent]" if your language doesn't have a single word / phrase
//: that means "yesterday" other than the literal equivalent of "1 day ago".
String yesterday = mtrc("global", "Yesterday");

//: Translate as "[no-equivalent]" if your language doesn't have a single word / phrase
//: that means "2 days ago" in your language other than the literal equivalent of "2 days ago".
String yesteryesterday = mtrc("global", "Yesteryesterday");

//: Translate as "[no-equivalent]" if your language doesn't have a single word / phrase
//: that means "3 days ago" in your language other than the literal equivalent of "3 days ago".
String yesteryesteryesterday = mtrc("global", "Yesteryesteryesterday");

if (days == 1 && yesterday != "[no-equivalent]") {
        return yesterday;
    }

if (days == 2 && yesteryesterday != "[no-equivalent]") {
        return yesteryesterday;
    }

if (days == 3 && yesteryesteryesterday != "[no-equivalent]") {
        return yesteryesteryesterday;
    }

//: Please provide a translation for the singular and all plural forms available in your language.
//: See https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms
return mtrc("global", "%n day(s) ago", nullptr, days);
```
But that's clearly more trouble than it's worth for something like this.

src/project/internal/projectuiactions.cpp
```diff
@@ -31,91 +31,93 @@ const UiActionList ProjectUiActions::m_actions = {
              mu::context::UiCtxAny,
              mu::context::CTX_ANY,
              TranslatableString("action", "&Open…"),
-             TranslatableString("action", "Open…"),
+             TranslatableString("action", "Open project"),
```
shoogle, line 34: @bkunda, what do you think about this change to the string displayed in Preferences?

src/engraving/libmscore/fret.cpp
Jojo-Schmitz: 0 is not the singular case then, or is it? Whatever: there's no "0" translation in French on Transifex, just 1, many and other. I guess orther does cover the 0 case, if so its  current translation "%n point**s** sur la frette %1" needs to get adjusted.

Still that sentence seems correct, as far as I can tell. Your proposal is too though.

src/framework/global/dataformatter.cpp
Jojo-Schmitz: Check again, please

src/engraving/libmscore/fret.cpp
shoogle: We'll have to experiment to see whether MuseScore uses the `1` translation for the `n = 0` case in French.

There may be languages where the `1` case is used for all numbers ending in 1 (`n = 1, 11, 21, 31...`, etc.) or something like that, so I would be wary of encouraging special cases except by experienced translators who know exactly what they are doing.

Jojo-Schmitz: Fine by me, hence I'm using your wording now.

I'd rather leave it to you to adjust https://github.com/musescore/MuseScore/wiki/Help-translate-MuseScore#plural-forms

src/appshell/view/appmenumodel.cpp
```diff
@@ -132,7 +132,7 @@ MenuItem* AppMenuModel::makeFileMenu()
     MenuItemList fileItems {
         makeMenuItem("file-new"),
         makeMenuItem("file-open"),
-        makeMenu(TranslatableString("appshell", "Open &recent"), recentScoresList, "menu-file-open", openRecentEnabled),
+        makeMenu(TranslatableString("appshell/file", "Open &recent"), recentScoresList, "menu-file-open", openRecentEnabled),
```
cbjeukendrup, line 135: I quite like this, since it is also helpful for translators to find the string. I thought, maybe we can make it even clearer by making the context `"appshell/menubar/file"` etc. 
(I think "menubar" is the clearest, better than something like "appmenu" or "mainmenu", what do you think?)

Jojo-Schmitz: I'm not sure how helpfil the `appshell` part is. But just `menu` should be enough, this string here is not in the menubar itself too, its 'parent', "&File" is, this is ie below that.

cbjeukendrup:
@Jojo-Schmitz Sorry, I should maybe have told you earlier, but as you've probably already noticed, I cherry-picked the "non-controversial" part of your changes into #13523 🙂

Jojo-Schmitz:
@cbjeukendrup I just noticed that some of the commits got lost on rebase, so thanks for telling me why ;-)

bkunda:
I see this fix has successfully removed the "&" from the "publish" tab buttons, which is great. 
However I'm not in favour of the proposed new button labels. I think the word "project" is unnecessary and potentially confusing, especially when actions like "print" and "export" are already so common and uncontentious.
Also "Save project to cloud" should actually be "Publish to MuseScore.com", as it currently is in the latest nightlies. This button triggers a flow that only uploads your score as a publicly visible document, so we don't want the label to suggest the possibility of private cloud storage.

Jojo-Schmitz:
OK, let's drop the term 'Project' then

Jojo-Schmitz:
Done, please recheck

cbjeukendrup:
@Jojo-Schmitz I'm sorry, but I think we'll need to put this PR on hold for a few reasons. I'm still not convinced that the changes in `DataFormatter` are a good idea. They make the situation worse for languages where the "singular" case does _not_ only correspond with %n = 1.
Also, the changes in PublishToolBarModel will become unnecessary anymore when #13639 gets merged, which provides a more general solution to the problem of mnemonics. 
And @bkunda is still not really happy with the changes in ProjectUiActions.

The added translators comments are fine, but I see you already added them on Transifex too, so I'm not sure how useful it is to put them in the code too; it is not quite elegant to have all those big comments in there.

Jojo-Schmitz:
Anything I could or should do here?

Jojo-Schmitz:
OK, what parts of this PR are agreeable and which are not and/or need more discussions?

bkunda:
> OK, what parts of this PR are agreeable and which are not and/or need more discussions?

Is this PR still even needed? It seems to me that the issues it resolves (https://github.com/musescore/MuseScore/issues/12921 and https://github.com/musescore/MuseScore/issues/13164) have already been fixed.
Am I perhaps missing something?

Jojo-Schmitz:
Hmm, indeed seems all resolved (all but the dataformatter simplifications, which had been challenged anyway)
So let's close it

Issue #12466, opened 2022-07-19
[MU4 Issue] When opening score that contains no instruments, "Loading audio samples" stays forever


cbjeukendrup:
**Describe the bug**
If you open a score, the "Loading audio samples" message appears in the Status Bar. However, if you open a score that contains no instruments, the loading message stays forever.

**To Reproduce**
1. Create a score
2. Remove all instruments using the instruments panel
3. save and close the score
4. reopen the score
The "Loading audio samples" message appears and stays forever.

**Expected behavior**
If there are no instruments, nothing needs to be loaded, so the loader shouldn't appear at all. However, in practice, the metronome track would still need to be loaded, so the loader should appear only for that metronome track and disappear after that. 

**Platform information**
 - OS: reported on macOS

nicolaiskogheim:
In the current version of MuseScore [1], it is not possible to save a score without instruments, so this should not be possible to reproduce.

[1] MuseScore version (64-bit): 4.2.1-240230937, revision: github-musescore-musescore-d757433

rgreen5:
I wonder if this is perhaps related to #18928.

Issue #11909, opened 2022-06-06
[MU4 Issue] All instruments can be deleted via Inspector Instruments 


boryskuzmenko:
**Describe the bug**
At least one instrument should be present on score and you shouldn't be able to delete the last one. This rule works for creating score procedure and Instruments dialog, but doesn't for Instruments tab in Inspector. You can delete the last instrument by clicking Backspace.
As result we have another weird behavior. After all instruments deleted playback still can be run.

**To Reproduce**
1. Create a score
2. In Inspector delete all instruments by clicking Backspace
3. Press Play

**Expected behavior**
At least one instrument should be present on score and you shouldn't be able to delete the last one

**Screenshots**
![Capture 2022-06-07_12-52-06_AM](https://user-images.githubusercontent.com/101250347/172256349-1f9b755b-0084-4f22-b036-07cbffef8a84.png)

**Platform information**
 - OS: macOS, Windows, Linux

shoogle:
You're right: there should always be at least one instrument in the score, and possibly in each part (`Excerpt`) too.

It's not because we want to stop users creating empty scores. It's because there will be other bugs if we allow empty scores. For example, text frames are stored in the top staff in the MSCX, so those frames will be lost when the file is saved with no staves and reopened later on. The safest option is to disallow empty scores.

cbjeukendrup:
I'm not sure about disallowing empty `Excerpts` though, because when you create a new one, it is empty at first too.

shoogle:
@cbjeukendrup, that's a good point. I tried it now and noticed more bugs with deleting instruments in `Excerpts` so I think we should concentrate on the main score here and deal with `Excepts` in a separate issue.

Tantacrul:
So this is actually my doing! I intentionally asked to get rid of the old MS3 behaviour of needing to have one instrument present on the score at all times because I keep finding that the MS3 rule results in frustrating experience under certain circumstances.

Mainly, when users want to change a score with one instrument to have something else instead (like they do with those 1 staff piano scores), they have to add the instrument they want first and then delete the one they don't want, which is really weird.

Also, there's no good UX reason to force the behaviour of always needing one instrument.

If there are associated issues, we should solve them individually. For example:
- If the play feature still works when the score is empty, I think it should be disabled.
- If there is UI that can be interacted with while the score is empty it should (potentially) also be disabled

Tantacrul:
I should also mention that we'll need the concept of the score being empty in future anyway.

[...]

PR #12878, opened 2022-08-16
Fix #12864: fix crashes when there is no staff

### LINKED ISSUE #12864: [MU4 Issue] Crash when deleting staff from single-staff instrument/template
**Describe the bug**
When using Simple template score crashes after deleting its single staff (video 1)
When using any single-staff instrument score crashes after deleting it's single staff and clicking Add Staff button (video 2)

**Expected behavior**
No crash during any action with instruments

**Screenshots**
Video 1

https://user-images.githubusercontent.com/101250347/184632336-678a8799-85c8-4bd8-9546-bff337dcb289.mov

Video 2

https://user-images.githubusercontent.com/101250347/184632356-0abed7e3-ffcd-4eb4-a159-f3aa9bd6d9c9.mov

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

martin-danhier:
Resolves: #12864 

Hello,

There were various places where the code tried to access a staff even if there were none. I added additional checks to prevent this.
Now it works as intended:

https://user-images.githubusercontent.com/36641081/184849252-f43563b0-7f50-48d6-8e82-2e1b2a9dc342.mp4

Have a good day,
Martin

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -3833,6 +3833,10 @@ void Score::setScoreOrder(ScoreOrder order)
 
 void Score::updateBracesAndBarlines(Part* part, size_t newIndex)
 {
+    if (part->nstaves() == 0) {
+        return;
+    }
+
     bool noBracesFound = true;
     for (size_t indexInPart = 0; indexInPart < part->nstaves(); ++indexInPart) {
         size_t indexInScore = part->staves()[indexInPart]->idx();
@@ -3868,8 +3872,8 @@ void Score::updateBracesAndBarlines(Part* part, size_t newIndex)
                 undoAddBracket(part->staves()[0], 0, tp->bracket[0], part->nstaves());
             }
         }
-    } else {
-        updateBracketSpan(newIndex, newIndex - 1);
+    } else if (newIndex == 1) {
+        updateBracketSpan(1, 0);
     }
 }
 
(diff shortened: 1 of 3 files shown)


RomanPudashkin:
@Tantacrul what do you think about disabling the ability to remove ALL staves from a selected instrument (to avoid having instruments without staves)? The user will be able to remove all instruments from the score though. This would greatly reduce the number of such bugs and simplify the implementation of the instrument panel

Tantacrul:
> @Tantacrul what do you think about disabling the ability to remove ALL staves from a selected instrument (to avoid having instruments without staves)? The user will be able to remove all instruments from the score though. This would greatly reduce the number of such bugs and simplify the implementation of the instrument panel

Completely agree. I did not realise users were able to do this actually!

martin-danhier:
Yeah, it would make more sense. I can look into it if you want!

martin-danhier:
@RomanPudashkin @Tantacrul 

I modified the instrument panel to disallow deleting the last staff of an instrument.
I didn't remove the initial fix, because without it there could still be crashes when staves are reordered then deleted.

https://user-images.githubusercontent.com/36641081/184870338-f3bc9fe6-21b3-4f69-92e0-28e191e9b698.mp4

---

However, there still are many problems with scores without any instruments, as mentioned in #11909 #12465 #12466.
Also, when loading an empty score, the score will be completely blank and nothing will work at all (I will report it later if I don't find an existing report for it):

https://user-images.githubusercontent.com/36641081/184871347-1dec4477-82af-4ca7-ab70-088dcd1ca903.mp4

Should I extend the modification and disallow removing the last instrument as well ?

RomanPudashkin:
Thank you for your help! I have prepared a better solution in terms of reliability and ease of maintenance
https://github.com/musescore/MuseScore/pull/12891

PR #12892, opened 2022-08-16
[cleanup] Fix #includes in Engraving module

cbjeukendrup:
What this does:
- Remove unnecessary (obsolete) #includes, and add missing ones, increasing correctness 
- Reduce the number of included header files (directly or indirectly) per cpp file. This should slightly reduce the compile time, because we give the compiler as few information as possible. Probably not huge though.
- For certain header files, reduces the number of cpp files they are included in (directly or indirectly). This means that fewer cpp files need to be recompiled when editing such a header file. This is in some cases a big improvement for whoever needs to edit those header files frequently.
- Sort #includes alphabetically where relevant (=almost everywhere). This makes it just tidier and more glanceable and prevents double #includes. Please, let's stick with alphabetical order in the future, it's no effort.
- Rename a few files to something more sensible
- Fix non-unity build

Powered by Qt Creator's `clangd` diagnostics :)


Jojo-Schmitz:
.../src/engraving/libmscore/edit.cpp (now) apparently needs `#include <climits>`.

PR #12895, opened 2022-08-16
config.h is generated by autoconf

lyrra:


cbjeukendrup:
It'd also need to be add it to `.gitignore`, shouldn't it?

Jojo-Schmitz:
Nect to https://github.com/musescore/MuseScore/blob/aeea6b1e27e465b9f777eb5519fac8b601749176/.gitignore#L62 I guess

lyrra:
Does this still compiles for msvc? I'm guessing the CI uses mingw which has access to autoconf and can rebuild config.h (or opusenc is disabled in the CI build).

Jojo-Schmitz:
No, CI uses MSVC

lyrra:
Ok, so it might be generated by cmake, I'll do some tests..

lyrra:
> Next to
> 
> https://github.com/musescore/MuseScore/blob/aeea6b1e27e465b9f777eb5519fac8b601749176/.gitignore#L62

That file should go as well. It seems cmake flac generates its own config.h.
I'm pushing a similar fix to opusenc that makes cmake also generate into builds.

Though, it seems if you remove those config.h files, it still can build, and I'm guessing it is picking up some other incorrect config.h

Issue #12886, opened 2022-08-16
[MU4 Issue] Navigation breaks down after enabling plugins


DmitryArefiev:
**To Reproduce**
1. Make sure all plugins are disabled
2. Navigate with keyboard
3. Select one and enable it
4. Try to continue navigation with Tab - **it doesn't work**

https://user-images.githubusercontent.com/90187801/184869148-4d882412-77e5-49ff-820b-b73154372a81.mp4

**Platform information**
 macOS, Windows, Linux

DmitryArefiev:
Hi @cbjeukendrup , @Eism found that the issue was caused by #12854

DmitryArefiev:
Fixed in #12897

PR #12854, opened 2022-08-14
Select only one plugin at a time in 'Enabled' and 'Disabled' lists together

cbjeukendrup:
Resolves: #12818
Resolves: #12826

PR #12897, opened 2022-08-16
Fix navigation after enabling plugin

cbjeukendrup:
Resolves: #12886 

Seems like there's a Qt bug / strange behaviour, which was just revealed by #12854. Problem is that the list view items have not yet been loaded at the time that we call `requestActive`, even though the items _should_ exist. Apparently, assigning something to the `currentIndex` property of the grid view somehow forces the items to be loaded. 

Also tweaks the navigation/selection borders of Plugins and Scores thumbnails yet a bit further.

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
@@ -29,8 +29,14 @@ Rectangle {
     property NavigationControl navigationCtrl: null
     property bool drawOutsideParent: true
 
+    //! NOTE: sometimes, the contrast between the item and the navigation focus
+    //! border might be very low, for example when the item is an image.
+    //! Therefore, we sometimes add some padding between the item and the border,
+    //! to keep the border easily distinguishable.
+    property real padding: 0
+
     anchors.fill: parent
-    anchors.margins: drawOutsideParent ? -border.width : 0
+    anchors.margins: drawOutsideParent ? -border.width - padding : 0
 
     visible: navigationCtrl ? navigationCtrl.highlight : false
 
@@ -39,6 +45,6 @@ Rectangle {
     border.color: ui.theme.fontPrimaryColor
     border.width: ui.theme.navCtrlBorderWidth
     radius: Number(parent.radius) > 0
-            ? (drawOutsideParent ? parent.radius + root.border.width : parent.radius)
+            ? parent.radius - anchors.margins
             : 0
 }
--- a/src/plugins/qml/MuseScore/Plugins/PluginsPage.qml
+++ b/src/plugins/qml/MuseScore/Plugins/PluginsPage.qml
@@ -222,7 +222,7 @@ Item {
 
         onClosed: {
             prv.resetSelectedPlugin()
-            Qt.callLater(resetNavigationFocus)
+            resetNavigationFocus()
         }
 
         function resetNavigationFocus() {
--- a/src/plugins/qml/MuseScore/Plugins/PluginsPage.qml
+++ b/src/plugins/qml/MuseScore/Plugins/PluginsPage.qml
@@ -102,9 +102,7 @@ Item {
 
         anchors.top: parent.top
         anchors.left: parent.left
-        anchors.leftMargin: root.sideMargin
         anchors.right: parent.right
-        anchors.rightMargin: root.sideMargin
         anchors.bottom: panel.visible ? panel.top : parent.bottom
 
         contentWidth: width
@@ -113,20 +111,12 @@ Item {
         topMargin: topGradient.height
         bottomMargin: 24
 
-        ScrollBar.vertical: StyledScrollBar {
-            parent: flickable.parent
-
-            anchors.top: parent.top
-            anchors.bottom: panel.visible ? panel.top : parent.bottom
-            anchors.right: parent.right
-
-            visible: flickable.contentHeight > flickable.height
-            z: 1
-        }
-
         Column {
             id: column
+
             anchors.fill: parent
+            anchors.leftMargin: root.sideMargin
+            anchors.rightMargin: root.sideMargin
 
             spacing: 24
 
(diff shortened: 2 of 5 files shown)


DmitryArefiev:
Hi @cbjeukendrup, it this ready for testing?

DmitryArefiev:
OK, got it! I'll test after review from @Eism

src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
```diff
@@ -39,6 +39,6 @@ Rectangle {
     border.color: ui.theme.fontPrimaryColor
     border.width: ui.theme.navCtrlBorderWidth
     radius: Number(parent.radius) > 0
-            ? (drawOutsideParent ? parent.radius + root.border.width : parent.radius)
```
Eism, line 42: Please explain why this change is needed?

src/plugins/qml/MuseScore/Plugins/PluginsPage.qml
```diff
@@ -102,9 +102,7 @@ Item {
 
         anchors.top: parent.top
         anchors.left: parent.left
-        anchors.leftMargin: root.sideMargin
```
Eism, line 105: It is better to return as it was. Like in scores and learn pages

Eism (approved):

Eism (changes requested):
Sorry, approved accidentally

src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
cbjeukendrup: It fixes the corner radius of the navigation border for the thumbnails on the Scores page. In ScoreItem, we set the margin for the NavigationBorder explicitly, to create some padding between the border and the thumbnail, and with this change, the corner radius of the border is adjusted accordingly.

src/plugins/qml/MuseScore/Plugins/PluginsPage.qml
cbjeukendrup: The change is necessary because otherwise the navigation/selection border would be cut off. We can't set `clip` to `false`, because that will result in the items being visible above and below the scrollable area.

(In the current master branch, the clipping problem does not occur, because the items have an inset of 4px. But that way, they are not aligned with the "Enabled"/"Disabled" title text, which looks a bit ugly. )

src/plugins/qml/MuseScore/Plugins/internal/PluginsListView.qml
```diff
@@ -51,6 +51,10 @@ Column {
     signal navigationActivated(var itemRect)
 
     function focusOnFirst() {
+        // Strange hack to force the view to load the items. Without this, `view.itemAtIndex(0)` might be null
```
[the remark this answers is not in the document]
cbjeukendrup: I'm afraid no, but I'll take a look at the Qt source code, maybe that gives some insight why this hack works and if there's a better way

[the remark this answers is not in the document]
cbjeukendrup: Seems I found something that works

src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
Eism: What's wrong with the borders for the Score item in master? Looks nice

image

src/plugins/qml/MuseScore/Plugins/internal/PluginItem.qml
```diff
@@ -50,29 +50,17 @@ Item {
...
         width: parent.width
         height: 144
 
-        color: "transparent"
-        radius: root.radius + border.width
-
-        border.color: ui.theme.fontPrimaryColor
-        border.width: root.selected ? 2 : 0
-
-        NavigationFocusBorder {
```
Eism, line 66: Why you decided to use a custom navigation border instead of this component?

cbjeukendrup: The border for navigation and selection looks exactly the same, so I thought it is redundant to use two separate rectangles for these purposes. If you prefer, I could reuse `NavigationFocusBorder` and just override its `visible` property to make it also visible when the item is selected.

src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
cbjeukendrup: It's subtle, _very_ subtle, but it is not entirely correct. I've tried to indicate the difference in this image:
Scherm­afbeelding 2022-08-29 om 10 57 24
With the old version (left), it will look like the padding between the border and the thumbnail is a bit bigger in the corners. With the new version, it will look and feel like 2px all around.

src/plugins/qml/MuseScore/Plugins/internal/PluginItem.qml
Eism: They perform different functions and it is better not to mix them. 
It should be clear from the code where the selection rectangle is and where the navigation is.

src/framework/uicomponents/qml/MuseScore/UiComponents/NavigationFocusBorder.qml
Eism: ok, thanks for clarification

cbjeukendrup: I decided to add a dedicated "padding" property to `NavigationFocusBorder`, to make it more clear what we're doing.

src/plugins/qml/MuseScore/Plugins/internal/PluginItem.qml
cbjeukendrup: Okay, I separated them!

Eism (approved):

DmitryArefiev:
Tested #12886 on Win10, Mac12 - **FIXED**  
(can't check linux build, something wrong with recent CI builds for linux, will check on master)

cbjeukendrup:
> something wrong with recent CI builds for linux

Oh no... I'm afraid this is related to #12963...

Issue #11900, opened 2022-06-06
[MU4 Issue] Unable to drag multiple selected lines (hairpins, slurs, dim. etc) with mouse


DmitryArefiev:
**Describe the bug**
Unable to drag multiple selected lines (hairpins, slurs, dim. etc) with mouse.
(For multiple selection use Cmd/Ctr)

https://user-images.githubusercontent.com/90187801/172189635-fac20bbb-ff69-465a-93bf-58d623e14823.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

DmitryArefiev:
Can be related to #8512

DmitryArefiev:
Fixed in #12910

PR #12910, opened 2022-08-17
[MU4] Fix #12587: Images cannot be moved via drag or keyboard

### LINKED ISSUE #12587: [MU4 Issue] Images cannot be moved via drag or keyboard
**Describe the bug**
Cursor keys should nudge elements, and do for most (sometimes requiring edit mode), but for images, no nudging seems possibility.  This is a regression from MU3.

**To Reproduce**
1. In an empty, paste an image (that you capture to your clipboard elsewhere) onto the first rest
2. Try to move it via keyboard or mouse
3. See error - the image resizes rather than moves

**Expected behavior**
Image should resize if you drag it explicitly, or select a handle to adjust by keybaord, but default operation of both drag and cursor keys should be nudge.  Or, if a conscious choice is being made that resizing should be the default, then there should be an obvious way to move instead.  But edit mode doesn't work, nor does deselecting then dragging.

**Platform information**
- Linux

**Additional context**
I was able to drag by right-clicking - actually, to be more accurate, by dragging with two fingers on my touchpad.  This is entirely non-obvious and probably a bug in itself.

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12587
Resolves: https://github.com/musescore/MuseScore/issues/11900

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -1244,6 +1244,8 @@ void NotationActionController::startEditSelectedElement(const ActionData& args)
     if (interaction->textEditingAllowed(element)) {
         PointF cursorPos = !args.empty() ? args.arg<PointF>(0) : PointF();
         interaction->startEditText(element, cursorPos);
+    } else if (element->hasGrips()) {
+        interaction->startEditGrip(element, element->defaultGrip());
     } else {
         interaction->startEditElement(element);
     }
--- a/src/notation/inotationinteraction.h
+++ b/src/notation/inotationinteraction.h
@@ -141,6 +141,7 @@ public:
     virtual bool isHitGrip(const PointF& pos) const = 0;
     virtual void startEditGrip(const PointF& pos) = 0;
     virtual void startEditGrip(EngravingItem* element, mu::engraving::Grip grip) = 0;
+    virtual void endEditGrip() = 0;
 
     virtual bool isElementEditStarted() const = 0;
     virtual void startEditElement(EngravingItem* element) = 0;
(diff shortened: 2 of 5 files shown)


src/notation/internal/notationactioncontroller.cpp
```diff
@@ -1244,6 +1244,8 @@ void NotationActionController::startEditSelectedElement(const ActionData& args)
     if (interaction->textEditingAllowed(element)) {
         PointF cursorPos = !args.empty() ? args.arg<PointF>(0) : PointF();
         interaction->startEditText(element, cursorPos);
+    } else if (element->hasGrips()) {
```
wizofaus, line 1247: but an Image *does* have grips, surprised that this makes a difference...

wizofaus:
It's certainly likely that this is better fix than the one I came up with, but does it still allow resizing via the bottom-right grip handle?

src/notation/internal/notationactioncontroller.cpp
RomanPudashkin: This code is needed for the "Edit element" item in the context menu. It automatically selects the default grip point of an element and starts "edit mode"

wizofaus: Ok but I believe if you use that mode then you *can't* use the mouse to move an image.  But that's the same as v3.

RomanPudashkin:
> does it still allow resizing via the bottom-right grip handle?

Yes, sure :)

 >BTW: very clear commit messages, great to see! I hope this means we're no longer expected to always squash commits?
 
We need to squash only those commits that don't make much sense. For example, commits with code-style fixes, commits with some minor code-review fixes, etc. Small commits with detailed descriptions are still welcome

Eism (approved):

DmitryArefiev:
Tested #12587 #11900 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

Issue #12823, opened 2022-08-12
[MU4 Issue] Inclomplete message when checking for language update


Jojo-Schmitz:
**Describe the bug**
Inclomplete message when checking for language update

**To Reproduce**
1. Edit > Preferences > Languages
2. Make sure language is set to "System default"
3. Click on 'Check for language updates'
4. See error:
![image](https://user-images.githubusercontent.com/1786669/184360503-e46d1804-bbe2-4a80-aaf3-efc90f0665b4.png)

**Expected behavior**
The language should be named, either "Deutsch" (in my case) or "System Language" (or similar):
![image](https://user-images.githubusercontent.com/1786669/184361843-981a67fe-987e-4bee-8342-b104c56f827f.png)
like it is when having selected "Deutsch" as the language
(and yes, that string hadn't been translated until a minute ago)

**Platform information**
 - OS: Windows, MuseScore self-built from latest master branch

Issue #12702, opened 2022-08-06
[MU4 Issue] Language update check not working when set to System Default


henkdegroot:
**Describe the bug**
At the start of MS4, it detects my system language is Dutch and shows the GUI with Dutch strings.
When I check in preferences, I noticed the language is set to "System Default".
Click the "Check for update" button now and the system tells me it is up-to-date.
I know this is not correct, as not all text is shown in Dutch.
When I change the "System Default" to "Nederlands" (Dutch) and click the "Check for update" button, it will download and update the  latest language and update.

**To Reproduce**
See "describe the bug"

**Expected behavior**
Check for update should work while the "System Default" is selected.

**Platform information**
 - OS: Windows 10 with Dutch language

henkdegroot:
@cbjeukendrup, this is the issue I referenced on Discord.

Tantacrul:
@cbjeukendrup - let me know if you're happy for me to assign this to you.

cbjeukendrup:
Yes, in fact I'm already working on it :)

Issue #12535, opened 2022-07-23
[MU4 Issue] "System default" language is incorrect


mxmou:
**Describe the bug**

At least on Windows, the "system default" language depends on the regional format setting instead of the preferred display language.

**To Reproduce**

1. Change the date format in the Windows settings to something different from the display language.
2. Set the language to "system default" in MuseScore preferences.

**Expected behavior**

The correct setting would be used.

**Platform information**

 - OS: Windows

**Additional context**

The problem is probably caused by [this function](https://github.com/musescore/MuseScore/blob/master/src/languages/internal/languagesservice.cpp#L406-L413) which should call [`QLocale::uiLanguages()`](https://doc.qt.io/qt-6/qlocale.html#uiLanguages) instead of `QLocale::name()`.

The same issue exists in MuseScore 3.

PR #12917, opened 2022-08-17
Rework languages module and fix updating language

cbjeukendrup:
Resolves: #12535
Resolves: #12702
Resolves: #12823
Resolves: reports on Discord that updating language doesn't work reliably

The languages module was still based on the old design where languages were 'installable' like add-ons. This PR reworks the module, so that it contains only code for the new design (languages are always packed with MuseScore, but you can download updated versions). This simplifies the code quite a lot.


src/appshell/view/preferences/generalpreferencesmodel.cpp
```diff
@@ -54,30 +54,19 @@ void GeneralPreferencesModel::load()
...
 {
-    QString currentLanguageCode = this->currentLanguageCode();
-    LanguageStatus::Status languageStatus = languagesService()->languageStatus(currentLanguageCode);
-    if (languageStatus != LanguageStatus::Status::NeedUpdate) {
-        Language currentLanguage = languagesService()->languages().val.value(currentLanguageCode);
-        QString msg = mu::qtrc("appshell/preferences", "Your version of %1 is up to date").arg(currentLanguage.name);
-        interactive()->info(msg.toStdString(), "");
-        return;
-    }
+    QString currentLanguageCode = effectiveCurrentLanguageCode();
```
RomanPudashkin, line 57: effective? What does it mean? :)

src/framework/global/types/retval.h
```diff
@@ -31,9 +31,12 @@ template<typename T>
 struct RetVal {
     Ret ret;
     T val;
+
     RetVal() = default;
     RetVal(const Ret& r)
         : ret(r) {}
+    RetVal(const T& v)
```
RomanPudashkin, line 38: Not sure about this change. Such constructors very often cause various kinds of problems because they do things implicitly

src/languages/ilanguagesservice.h
```diff
@@ -35,14 +35,15 @@ class ILanguagesService : MODULE_EXPORT_INTERFACE
...
-    virtual LanguageStatus::Status languageStatus(const QString& languageCode) const = 0;
+    virtual bool hasPlaceholderLanguage() const = 0;
+    virtual const Language& placeholderLanguage() const = 0;
 
-    virtual RetCh<LanguageProgress> install(const QString& languageCode) = 0;
-    virtual RetCh<LanguageProgress> update(const QString& languageCode) = 0;
-    virtual Ret uninstall(const QString& languageCode) = 0;
+    virtual LanguageProgressChannel update(const QString& languageCode) = 0;
+
+    virtual ValCh<bool> needRestartToApplyLanguageChange() const = 0;
```
RomanPudashkin, line 46: I would suggest not using ValCh. Use two separate methods here (one for notification and one for getting the actual value). We decided to drop this class in the future. For now, we should at least not use it in new parts of the code

src/languages/ilanguagesservice.h
```diff
@@ -35,14 +35,15 @@ class ILanguagesService : MODULE_EXPORT_INTERFACE
...
     virtual ValCh<Language> currentLanguage() const = 0;
 
-    virtual LanguageStatus::Status languageStatus(const QString& languageCode) const = 0;
+    virtual bool hasPlaceholderLanguage() const = 0;
+    virtual const Language& placeholderLanguage() const = 0;
 
-    virtual RetCh<LanguageProgress> install(const QString& languageCode) = 0;
-    virtual RetCh<LanguageProgress> update(const QString& languageCode) = 0;
-    virtual Ret uninstall(const QString& languageCode) = 0;
+    virtual LanguageProgressChannel update(const QString& languageCode) = 0;
```
RomanPudashkin, line 44: Can we use framework::Progress instead?

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
 
 void LanguagesService::init()
```
RomanPudashkin, line 60: Please add TRACEFUNC here

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
-        closeOperation(languageCode, languageProgressStatus);
-    }, Asyncable::AsyncMode::AsyncSetRepeat);
-    QtConcurrent::run(this, &LanguagesService::th_install, languageCode, languageProgressStatus, languageFinishChannel);
-
-    return result;
+    return true;
 }
 
-RetCh<LanguageProgress> LanguagesService::update(const QString& languageCode)
+void LanguagesService::loadLanguages()
```
RomanPudashkin, line 104: Please add TRACEFUNC

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
+        m_languagesHash.insert(it.key(), lang);
     }
 
-    return make_ret(Err::NoError);
+    m_placeholderLanguage.code = PLACEHOLDER_LANGUAGE_CODE;
+    m_placeholderLanguage.name = "«Placeholder translations»";
 }
 
 void LanguagesService::setCurrentLanguage(const QString& languageCode)
 {
```
RomanPudashkin, line 133: There should be a check to make sure that the current language actually changed:

```
if (languageCode == currentLanguageCode()) {
    return;
}

...
```

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
+    };
 
-bool LanguagesService::checkLanguageFilesHash(const QString& languageCode, const LanguageFiles& languageFiles) const
-{
-    io::paths_t filePaths = configuration()->languageFilePaths(languageCode);
-    int filesSize = static_cast<int>(filePaths.size());
-    if (filesSize != languageFiles.size()) {
-        return false;
-    }
+    auto tryCode = [hasCode](QString code) -> QString {
```
RomanPudashkin, line 185: What does this lambda do?

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
-            m_translatorList.append(translator);
-        } else {
-            LOGE() << "Error load translate " << filePath;
-            delete translator;
+    m_updateOperationsHash.insert(languageCode, progressChannel);
+
+    async::Channel<Ret>* languageFinishChannel = new async::Channel<Ret>();
+    languageFinishChannel->onReceive(this, [this, languageCode, languageFinishChannel](const Ret& ret) {
+        if (ret.code() == static_cast<int>(Err::AlreadyUpToDate)) {
+            QString msg = mu::qtrc("appshell/preferences", "Your version of %1 is up to date").arg(m_languagesHash[languageCode].name);
```
RomanPudashkin, line 282: appshell/preferences?

RomanPudashkin: Looks like this code should be at a level above

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
+    io::Buffer buff(&ba);
+    ZipReader zipReader(&buff);
 
-    Ret unpack = languageUnpacker()->unpack(languageCode, languageArchivePath, configuration()->languagesUserAppDataPath());
-    if (!unpack) {
-        LOGE() << "Error unpack " << unpack.toString();
-        finishChannel->send(unpack);
+    for (const auto& info : zipReader.fileInfoList()) {
+        io::path_t userFilePath = configuration()->languagesUserAppDataPath().appendingComponent(info.filePath);
+        ret = fileSystem()->writeFile(userFilePath, zipReader.fileData(info.filePath.toStdString()));
```
RomanPudashkin, line 377: As far as I understand, you need to use mi::WriteResourceLockGuard here

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
+    QLocale locale(lang.code);
+    QLocale::setDefault(locale);
+    qApp->setLayoutDirection(locale.textDirection());
 
-    for (const QString& fileType: languageFileTypes) {
-        QJsonObject instrumentsFile = languageObject.value(fileType).toObject();
-        if (!instrumentsFile.isEmpty()) {
-            files << LanguageFile(instrumentsFile.value("file_name").toString(),
-                                  instrumentsFile.value("hash").toString());
+    if (m_inited) {
```
RomanPudashkin, line 170: As far as I understand, there is no longer any need for this code, since we ask the user to restart the app anyway

src/languages/internal/languagesservice.cpp
```diff
@@ -53,470 +59,340 @@ static QString downloadingStatusTitle()
...
+    m_currentLanguage.set(lang);
 }
 
-bool LanguagesService::isLanguageExists(const QString& languageCode) const
+QString LanguagesService::effectiveLanguageCode(const QString& languageCode)
 {
-    io::paths_t files = configuration()->languageFilePaths(languageCode);
-    return !files.empty();
-}
+    auto hasCode = [this](const QString& code) -> bool {
```
RomanPudashkin, line 181: This lambda looks a bit redundant...

src/languages/internal/languagesservice.cpp
cbjeukendrup: It tries decreasingly specific versions of the specified language code. For example, the system ui language might be `nl_NL`. This lambda tries if we have that. That's not the case, so we try a simpler version, just `nl`. We do have that one, so it will be returned by this lambda.

src/languages/ilanguagesservice.h
cbjeukendrup: I will remove it, but actually I have started to like it, since it reduces the amount of members required to do a relatively simple thing. So, just for my understanding, what is wrong with it?

src/framework/global/types/retval.h
cbjeukendrup: Yeah, I doubted about it too. Probably I should just be not too lazy to write `RetVal<QString>::make_ok()` :)

src/appshell/view/preferences/generalpreferencesmodel.cpp
cbjeukendrup: The specified language code (for example, "system") needs to be transformed into a code for which we provide a translation (for example, "en_US" or "nl"). The latter is what I call the "effective" language code.

src/appshell/view/preferences/generalpreferencesmodel.cpp
```diff
@@ -54,30 +55,27 @@ void GeneralPreferencesModel::load()
...
+            interactive()->info(msg.toStdString(), "");
+        }
+    });
 }
 
 QVariantList GeneralPreferencesModel::languages() const
 {
-    ValCh<LanguagesHash> languages = languagesService()->languages();
-    QList<Language> languageList = languages.val.values();
+    LanguagesHash languages = languagesService()->languages();
```
RomanPudashkin, line 77: `const LanguagesHash&. 
`

Also, ideally, this line should be removed. We can iterate over the container using iterators
`QList<Language> languageList = languages.values(); `

src/appshell/view/preferences/generalpreferencesmodel.cpp
RomanPudashkin: This "efficient" term is very confusing. Why can't LanguagesService::currentLanguageCode() return the code in the proper format?

src/appshell/view/preferences/generalpreferencesmodel.cpp
cbjeukendrup: There are two possibilities:
```c++
    QList<Language> languages = languagesService()->languages().values();

    std::sort(languages.begin(), languages.end(), [](const Language& l, const Language& r) {
        return l.code < r.code;
    });

    QVariantList result;

    for (const Language& language : languages) {
        QVariantMap languageObj;
        languageObj["code"] = language.code;
        languageObj["name"] = language.name;
        result << languageObj;
    }
```

```c++
    QVariantList result;

    for (const Language& language : languagesService()->languages()) {
        QVariantMap languageObj;
        languageObj["code"] = language.code;
        languageObj["name"] = language.name;
        result << languageObj;
    }

    std::sort(result.begin(), result.end(), [](const QVariant& l, const QVariant& r) {
        return l.toMap().value("code").toString() < r.toMap().value("code").toString();
    });
```

In the first version, the sorting is probably more efficient, but the `.values()` is less efficient. 
In the second version, the iteration is faster but the sorting is more cumbersome. (Apparently, if you iterate over a QHash in `for (… : …)` style, you get the values only, not a key+value pair like with std::map.)

I've a light preference for the first version, what's your opinion?

src/appshell/view/preferences/generalpreferencesmodel.cpp
cbjeukendrup: `effective`, not `efficient`. I believe the word `effective` has a secondary meaning like "actual", "being used", "in fact"; that is the meaning I'm referring to in this case. 

We need both the specified code and the "actual" code: when determining which item should be selected in the list in Preferences, we need the specified code (otherwise, we could not show the user that the `"system"` language is selected), and when doing any operation with LanguagesService, we need the "actual" code, because `"system"` is not a language itself.

Would it be clearer to call the method `actualLanguageCode`, or `realLanguageCode`, or `usableLanguageCode`? Personally I think none of those is really better...

Eism: > We need both the specified code and the "actual" code: when determining which item should be selected in the list in Preferences, we need the specified code (otherwise, we could not show the user that the "system" language is selected), and when doing any operation with LanguagesService, we need the "actual" code, because "system" is not a language itself.

I think, only service should know about this logic. 

The logic is identical to audio and midi device selection. There, the model is very simple - get from the service a list of all devices(contains a system default) and the current device(might be a system default) and displays the current state.
The selected device is transferred to the configurator and written to the settings. The service subscribes to changing the current device from the configurator and connects to the selected device.

That is, model and configuration does not know about the logic with the system default. All logic is placed in the service.

cbjeukendrup:
Rebased!

src/languages/ilanguagesservice.h
```diff
@@ -35,14 +36,18 @@ class ILanguagesService : MODULE_EXPORT_INTERFACE
...
-    virtual ValCh<LanguagesHash> languages() const = 0;
-    virtual ValCh<Language> currentLanguage() const = 0;
+    virtual const LanguagesHash& languages() const = 0;
+    virtual Language language(const QString& languageCode) const = 0;
+    virtual const Language& currentLanguage() const = 0;
+    virtual async::Notification currentLanguageChanged() const = 0;
 
-    virtual LanguageStatus::Status languageStatus(const QString& languageCode) const = 0;
+    virtual bool hasPlaceholderLanguage() const = 0;
+    virtual const Language& placeholderLanguage() const = 0;
```
vpereverzev, line 45: What is the purpose of "placeholder" language?

RomanPudashkin (approved):

src/languages/ilanguagesservice.h
cbjeukendrup: It is the thing that @shoogle added here: #12553

Eism:
@cbjeukendrup 
There is one case of incorrect selection of the system language.

Tested on macOS.
The language of the system is English (en), the second language is Russian. The selected region is any country other than the UK or the USA. Then the algorithm incorrectly selects the language of the system.
It tries to compare en_CY and en with en_UK and en_GB (just en is missing in languages hash) and doesn't find it.

The system language is selected as ru, since languages hash contains just ru.

cbjeukendrup:
@Eism Ah! That must be the solution to #13926! I'll jump on this asap!

PR #12932, opened 2022-08-19
Fix #12674: Update MuseScore icons

### LINKED ISSUE #12674: [MU4 Task] Update the app icon used inside MuseScore itself
**Task description**
The new purple app icon is not yet used in all places inside the app itself:
- [ ] MuseScore > About MuseScore
- [ ] Home > My account (when logged out)
- [ ] (Installer imagery - that's probably a separate topic)
- [ ] (Anywhere else?)

HemantAntony:
Resolves: #12674 


cbjeukendrup:
@jessjwilliamson I notice that the new macOS dmg background image is a lot bigger than the old one (4.1 MB instead of 23 KB). Maybe the image needs to be optimized?

(Also, a nitpick: I see the extension has been changed from tiff to tif. I believe tiff with double f is the correct one, and tif is a historical thing from the time that Windows's file system could only work with 3-character filename extensions. Anyway, there is no functional difference.)

jessjwilliamson:
> @jessjwilliamson I notice that the new macOS dmg background image is a lot bigger than the old one (4.1 MB instead of 23 KB). Maybe the image needs to be optimized?
> 
> (Also, a nitpick: I see the extension has been changed from tiff to tif. I believe tiff with double f is the correct one, and tif is a historical thing from the time that Windows's file system could only work with 3-character filename extensions. Anyway, there is no functional difference.)

Thank you, no don't worry that's absolutely okay to mention the nitpicking :)

Eism:
Hm, what is the reason for changing svg to png?

cbjeukendrup:
> Hm, what is the reason for changing svg to png?

I believe I've heard Martin saying that the new logo was initially created as a vector graphic, but some color effects had to be applied (using Photoshop or something like that) which can only be done in PNG. It would be difficult to recreate the effects using pure SVG, so if we can live with PNG that would be preferable.

HemantAntony:
But it actually shouldn't be difficult to recreate the logo in svg. You can just take the current MU3 svg file and then just add a gradient to the background. You can use the colours from the PNG logos. If its okay with you, I can take a look @Tantacrul

cbjeukendrup:
Well, if I've understood it correctly, the problem is that it _is not_ a simple gradient from color 1 to color 2 anymore, because of those effects. (I don't know if people would be able to see the difference, since the logo is usually displayed at very small sizes, but... yeah.)

Tantacrul:
Yeah, it isn't a graphic made in SVG. It has additional effects that SVG does not support.

The new MacOS background should not be 4.1mb

@jessjwilliamson - did you run it through ImageOptim? (All PNGs should be run through that)

jessjwilliamson:
> Yeah, it isn't a graphic made in SVG. It has additional effects that SVG does not support.
> 
> The new MacOS background really shouldn't be **4.1mb** :)
> 
> @jessjwilliamson - did you run it through ImageOptim? (All PNGs should be run through that)

I've already updated the image and sent to Hemant :)

HemantAntony:
Ready for review / merge

Tantacrul:
The logo here looks weird in some way. Like it has been scaled up or really compressed or something.

image

Tantacrul:
And again here too. It's just jagged and low res looking? Any ideas why?

image

Tantacrul:
I also think we should use this version instead (scaled down accordingly. It's quite massive)
https://www.dropbox.com/s/cuqqo22g8jltz4u/MS4_MacOS_1024_1024.png?dl=0

It just looks a little nicer.

Tantacrul:
I do like the MacOS installation screen though.

HemantAntony:
I think its because I used a png instead of svg. svg files scale nicely since its vector.

This didn't affect me probably due to our different display properties

I will add the one you linked now and update the PR

cbjeukendrup:
@HemantAntony I think it is because you set `sourceSize` (as opposed to `width`/`height`). Either you should not set it (so that the image will be loaded at full resolution) or you should multiply it with screen resolution (`sourceSize: Qt.size(width * Screen.devicePixelRatio, height * Screen.devicePixelRatio)`).

HemantAntony:
Okay thanks! I'll update the pr again

HemantAntony:
@cbjeukendrup There is actually a difference between the two. Using `devicePixelRatio` leads to a granier image:

image


Before:
image

cbjeukendrup:
https://doc.qt.io/qt-5/qml-qtquick-image.html#mipmap-prop might help

Tantacrul:
Hmm. Don't know why but it's still blurry looking.

image

Shall we just pick a fixed size and I'll give you the PNG at those exact dimensions? 162x162px?

Tantacrul:
Sorry @HemantAntony - I bet there's some silly reason for this. It's just quite obviously blurry looking on high res screens.

HemantAntony:
@cbjeukendrup I tried mipmap. It sort of worked, but now its too blurry:
image

The left one is the one using `mipmap` while the right one is using `sourceSize` like before
This is my code:
image

cbjeukendrup:
@Tantacrul Maybe you can provide a version of the image that is something like 2x the desired size. That's what we do for most images and that seems to work OK. That way, we don't need to fiddle with mipmaps and sourceSize etc.

(@HemantAntony `Layout.preferredWidth: 100 * Screen.devicePixelRatio` is incorrect; it will make the image bigger on high-res screens. Multiplication with `Screen.devicePixelRatio` is only necessary for the `sourceSize` property, because that determines at which size the image will be loaded in memory. But for other sizes, high-res scaling will be done automatically.)

Tantacrul:
@cbjeukendrup @HemantAntony - sure. I can whip that up in a few secs. Just on question:

What is the pixel size the image is actually displaying at in code? It's sometimes tricky to figure out because I'm unsure what 100% scale is (is it for low resolution screens or high def screens etc.)

cbjeukendrup:
Generally, the numbers in the code represent "logical" pixels (as in, on a standard screen, 1 logical pixel = 1 physical pixel, and on a high-res screen 1 logical pixel might cover multiple physical pixels.) So, the values in the code always match the values in Figma.

But for images, `sourceSize` is an exception; it does two things:
- influence the "implicit" size of the image -> in this regard, it acts the same as other size properties
- influence the amount of pixels loaded in memory -> in this regards, it is scaling-unaware, so if you set sourceSize to 100x100, you get no more than 100x100 pixels, which means that on high-res displays the image will look low-res (because on a high-res screen, you want 200x200 "physical" pixels of image, to spread over 100x100 software pixels.)

(Hopefully this is a somewhat understandable explanation)

HemantAntony:
@Tantacrul can you try the artifact from this PR one more time?

Tantacrul:
It's now looking like this on Mac.  
image

And it still looks a little blurry.

I think Casper might be right: that I should just send you an updated image at x2 the required size?

HemantAntony:
Okay let's try that. Hopefully it works🤞🏾

Tantacrul:
If this is causing a lot of problems, we can just make a special SVG and try to match the image as closely as possible, although we can obviously just use an image (which we've done elsewhere). 

If .svg, we may need to wait until @jessjwilliamson is back from vacation next week.

Sorry @HemantAntony. I don't really know what to suggest!

HemantAntony:
The size of the image is 100 x 100

Tantacrul:
Apologies for the delay. Here's the image at 200x200.

https://www.dropbox.com/s/vex1br1557fmvzz/MS4_LOGO_200x200.png?dl=0

Tantacrul:
OK, that's looking fine to me. @RomanPudashkin / @Eism - if you are happy, can this one be merged?

Thanks @HemantAntony !

jessjwilliamson:
@Tantacrul Blurry on my 4k screen. I'm more than happy to create and provide SVG version if that will help? 
image

Tantacrul:
Yeah, perhaps best @jessjwilliamson

jessjwilliamson:
@HemantAntony Got svg version for you! :D https://www.dropbox.com/s/me1gfiy238hyzag/MS4_LOGO_SVG.svg?dl=0

HemantAntony:
@jessjwilliamson @Tantacrul I made a PR: #13112 :)

jessjwilliamson:
DELICIOUS! Perfect! Thank you Hemant!

Tantacrul:
If @jessjwilliamson is happy, I'm happy. @Eism / @RomanPudashkin - this can be merged, if you are happy. We just replaced the image with an SVG.

Sorry for the round trip @HemantAntony and thanks!

Issue #12806, opened 2022-08-11
[MU4 Issue] Side Panels Width gets extremely wide when closing them, opening them again and then opening the Mixer


KaiVinter:
**Describe the bug**
Pretty self explanatory.

**Requisites**
You need to have the Mixer closed and the side panels open by default when opening the score, otherwise this doesn't seem to happen.

**To Reproduce**
1. Prepare MS4 with the requisites above
2. Open a score
3. Open Mixer (nothing happens, it's just so you can see when it doesn't happen)
4. Close Mixer
5. Close all the Side Panels (Palettes, Properties and Instruments)
6. Reopen the Side Panels
7. Open Mixer again.
8. See error

**Expected behavior**
Side panels should not get any wider when opening the Mixer.

**Videos**

With the requisites:

https://user-images.githubusercontent.com/88064903/184242886-095a80e2-e0a6-4d8f-822a-56badefdf567.mp4

Without the requisites (no bug):

https://user-images.githubusercontent.com/88064903/184242997-56203693-5df8-4fb9-ba79-9166f0229bc6.mp4

**Platform information**
Tested on MacOS Mojave 10.14.6
Tested on MS4 Alpha 2

**Additional context**
There's usually a crash after closing MS4 with the score open, I'm not sure if it's always or only after reproducing this bug.

rgreen5:
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2844416344, revision: github-musescore-musescore-ae252cd

Same issue in Linux.

DmitryArefiev:
Fixed in #12937

Issue #12789, opened 2022-08-11
[MU4 Issue] Drumset palette layout is broken after opening/closing Mixer 


boryskuzmenko:
**To Reproduce**
1. Create a score from Battery Percussion template
2. Activate note input ('N')
3. Open mixer
4. Close mixer

**Expected behavior**
Drumset palette layout is not changed after opening/closing Mixer 

**Screenshots**

https://user-images.githubusercontent.com/101250347/184123008-5355e9e2-68a3-43ba-b134-16dae98e412f.mov

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

**Additional context**
The issue more noticeable on Windows

DmitryArefiev:
Fixed in #12937

PR #12431, opened 2022-07-15
[MU4] Fix #12421: Part, Mixer buttons and Playback panel are not centered after opening/creating a score

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12421

Issue #12421, opened 2022-07-15
[MU4 Issue] Part, Mixer buttons and Playback panel are not centered right after opening/creating a score


DmitryArefiev:
![Capture](https://user-images.githubusercontent.com/90187801/179209140-30c25165-1eca-48cc-8ed2-1c4648fbfb60.PNG)

see video 

https://user-images.githubusercontent.com/90187801/179209190-bae5cf00-7f5b-44b4-98d8-ebce0653fdec.mp4

**To Reproduce**
1. Open or create a score
2. Look at Part, Mixer buttons - **they are not centered properly** 

**Platform information**
 macOS, Windows, Linux

PR #12937, opened 2022-08-19
[MU4] Fix #12925: Dock panel does not respect maximum size anymore after moving it from leftmost column to other location

### LINKED ISSUE #12925: [MU4 Issue] Dock panel does not respect maximum size anymore after moving it from leftmost column to other location
**To Reproduce**
1. Open a score on the "Score" page.
2. Have some panels docked in the leftmost column.
3. Move one or more panels to a different dock location in a different column.
See error: the moved panel does not respect its maximum size anymore; you can resize it very wide.
4. Open the Mixer 
See error: the moved panel becomes automatically very wide (it seems that the space is distributed equally between the notation view and the moved panel)
5. Switch to another page and switch back to the "Score" page.
Note that everything is fine again, until you move a panel again to the left column and then again move it to another location.

**Expected behavior**
Maximum size should be respected.

**Screenshots**

https://user-images.githubusercontent.com/48658420/185448445-9f8ca1b4-38b9-45e7-9d55-6eac4e429464.mov

**Platform information**
 - OS: reported on macOS

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12925
Resolves: https://github.com/musescore/MuseScore/issues/12806
Resolves: https://github.com/musescore/MuseScore/issues/12789
Resolves: https://github.com/musescore/MuseScore/issues/12976

Reworked [my solution](https://github.com/musescore/MuseScore/pull/12431/commits/d89f2cb6146b0d51605aecca7515c7a097d3b825) for the bug https://github.com/musescore/MuseScore/issues/12421, because it doesn't cover some cases

### DIFF
--- a/src/appshell/view/dockwindow/internal/dockbase.cpp
+++ b/src/appshell/view/dockwindow/internal/dockbase.cpp
@@ -620,10 +620,14 @@ void DockBase::listenFloatingChanges()
         //! and emit floatingChanged() after that
         QTimer::singleShot(0, this, [this]() {
             updateFloatingStatus();
+
+            if (m_floating) {
+                applySizeConstraints();
+            }
         });
 
-        *frameConn
-            = connect(frame, &KDDockWidgets::Frame::isInMainWindowChanged, this, &DockBase::updateFloatingStatus, Qt::UniqueConnection);
+        *frameConn = connect(frame, &KDDockWidgets::Frame::isInMainWindowChanged,
+                             this, &DockBase::onIsInMainWindowChanged, Qt::UniqueConnection);
     });
 
     connect(m_dockWidget->toggleAction(), &QAction::toggled, this, [this]() {
@@ -638,6 +642,12 @@ void DockBase::updateFloatingStatus()
     bool floating = m_dockWidget && m_dockWidget->floatingWindow();
 
     doSetFloating(floating);
+}
+
+void DockBase::onIsInMainWindowChanged()
+{
+    applySizeConstraints();
+    updateFloatingStatus();
 }
 
 void DockBase::doSetFloating(bool floating)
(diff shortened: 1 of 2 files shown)


src/appshell/view/dockwindow/internal/dockbase.cpp
```diff
@@ -640,6 +642,12 @@ void DockBase::updateFloatingStatus()
     doSetFloating(floating);
 }
 
+void DockBase::onIsInMainWindowChanged()
+{
+    applySizeConstraints();
+    updateFloatingStatus();
```
cbjeukendrup, line 648: Is this method executed only when `isInMainWindow` has changed to `true`, or also when it has changed to `false`? If the latter, isn't this duplicating the code in line +621 to +624, or is that code invoked in different situations?

Or, to rephrase my question: when is `KDDockWidgets::Frame::isInMainWindowChanged` emitted and when `KDDockWidgets::DockWidgetQuick::parentChanged`?

cbjeukendrup (approved):
Looks good! I have one question though.

src/appshell/view/dockwindow/internal/dockbase.cpp
RomanPudashkin: >Is this method executed only when isInMainWindow has changed to true, or also when it has changed to false? 

It is called in both cases

>If the latter, isn't this duplicating the code in line +621 to +624, or is that code invoked in different situations?

The code in lines 621-624 is needed for one particular case: when the dock becomes undocked and loses its origin frame. KDDockWidgets::Frame::isInMainWindowChanged won't be called in this case (we just haven't yet connected to this signal from a new frame). This code ensures that the size constraints will be applied to the floating dock

RomanPudashkin: PS: docks do not always change their parent when they are undocked. Several docks can use the same frame at the same time. And that frame really belongs to only one of them (the first one who created it)

cbjeukendrup: Ah, thanks for the clarification!

DmitryArefiev:
Tested #12925 #12806 #12806 #12976 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #12940, opened 2022-08-20
fix #11159 allow changing stem colour

### LINKED ISSUE #11159: [MU4 Issue] Stems cannot be colored
**Describe the bug**
Discovered while testing for #11096 and somewhat related.
If a stem is selected and then a color is chosen, the stem's color does not change and the notehead's color changes.
I believe this is because, as described in the linked issue, the entire chord is selected by the "General" tool and this is the way that MuseScore renders the note's color value.

**To Reproduce**
1. Create a new note.
2. Click the note's stem.
3. Go to **General > Appearance > Color**
4. Change the note's color in the Color Picker.
5. Click "OK"

**Expected behavior**
The color of the stem should change, and the color of the notehead should not change.

**Platform information**
 - OS: Windows 11

**Additional context**
This is not a problem on MU3

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/11159)

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1391,6 +1391,9 @@ void Chord::processSiblings(std::function<void(EngravingItem*)> func) const
     }
     for (Note* note : _notes) {
         func(note);
+        for (NoteDot* dot : note->dots()) {
+            func(dot);
+        }
     }
     for (EngravingItem* e : el()) {
         func(e);
@@ -3754,7 +3757,7 @@ void Chord::undoChangeProperty(Pid id, const PropertyValue& newValue)
 
 void Chord::undoChangeProperty(Pid id, const PropertyValue& newValue, PropertyFlags ps)
 {
-    if (id == Pid::VISIBLE) {
+    if (id == Pid::VISIBLE || id == Pid::COLOR) {
         processSiblings([=](EngravingItem* element) {
             element->undoChangeProperty(id, newValue, ps);
         });


src/engraving/libmscore/chord.cpp
```diff
@@ -1391,6 +1391,9 @@ void Chord::processSiblings(std::function<void(EngravingItem*)> func) const
     }
     for (Note* note : _notes) {
         func(note);
+        for (NoteDot* dot : note->dots()) {
+            func(dot);
```
wizofaus, line 1395: It might seem Note itself should do this, but in principle a Note can be selected independently of its augmentation dots, whereas selecting a *chord* always means selecting all the sub-components.

Tantacrul:
Here you can see that when I make a range selection on a particular bar, I get a persistent crash. Other than that it seems to be working as I'd expect.

https://user-images.githubusercontent.com/21022311/185738781-bd56f161-1271-4387-87d8-5e2acc0387ef.mov

Tantacrul:
Here's a link to the file I tested on
https://www.dropbox.com/s/inkh6elaj54o5ou/New_All_Forever_01.mscz.zip?dl=0

wizofaus:
The changes I made only affect what happens when applying property changes to chords, not when selecting. But I haven't rebased with latest master so maybe I should do that.  It crashes a lot for me in debug mode too, and was before my changes were in place.

RomanPudashkin:
fixed in https://github.com/musescore/MuseScore/pull/15970

PR #12949, opened 2022-08-21
fixed #11323: Fixed navigation on the New Score Wizard

### LINKED ISSUE #11323: [MU4 Issue] Keyboard navigation in New Score Wizard popups
**Describe the bug**
Keyboard navigation is broken or unintuitive in the Time signature, Tempo and Measures popups within the New Score dialog.

image

The arrow keys are currently used to navigate the popups, but this results in focus getting stuck in number boxes, such as the beats per minute (BPM) field in the **Tempo popup** (default value = 120).

image

Within the number box, the Left and Right arrow keys are used to move the text caret, and the Up and Down arrow keys are used to operate the stepper buttons. This means the arrow keys are not available to navigate away from the number box to focus other controls in the popup.

This could be solved by using the Tab key for navigation within the popups. However, the focus order currently implemented for the Tab key is incomplete, leading to navigation that is either broken or unintuitive depending on which popup you are in.

---

## Time signature popup

image

Here it is difficult to focus the denominator (8) dropdown using the keyboard. Tabbing is not fully implemented.

### Desired focus order

KeyNav for NSW Time Signature Popup

- The radio buttons are grouped together as a single Tab object.
    - The arrow keys are used to select an individual radio button from within the group.
- Shift+Tab navigates in the opposite direction to that shown for the Tab key.
- The Up arrow key navigates in the opposite direction to that shown for the Down arrow key.

## Tempo popup

image

Here, all controls are reachable via the keyboard, but if you press Shift+Tab from the BPM (12) number box then it unexpectedly moves focus all the way to the top of the popup, so tabbing is not fully implemented.

### Desired focus order

KeyNav for NSW Tempo Popup

- The note duration buttons are grouped together as a single Tab object.
    - The arrow keys are used to select an individual note duration from within the group.
- Shift+Tab navigates in the opposite direction to that shown for the Tab key.
- The Left arrow key navigates in the opposite direction to that shown for the Right arrow key.

## Measures popup

image

Here it is impossible to focus the denominator (8) dropdown using the keyboard. Tabbing is not fully implemented.

### Desired focus order

KeyNav for NSW Measures Popup

- Shift+Tab navigates in the opposite direction to that shown for the Tab key.

Eism:
Resolves: #11323

### DIFF
--- a/src/project/qml/MuseScore/Project/internal/NewScore/MeasuresSettings.qml
+++ b/src/project/qml/MuseScore/Project/internal/NewScore/MeasuresSettings.qml
@@ -110,6 +110,8 @@ FlatButton {
             }
 
             TimeSignatureFraction {
+                id: pickupMeasure
+
                 Layout.topMargin: 12
                 Layout.leftMargin: 32
 
@@ -118,9 +120,8 @@ FlatButton {
                 availableDenominators: root.model.timeSignatureDenominators()
                 enabled: withPickupMeasure.checked
 
-                navigationPanel: content.navigationPanel
-                navigationRowStart: 1
-                navigationColumnStart: 0
+                navigationSection: popup.navigationSection
+                navigationPanelOrderStart: 2
 
                 onNumeratorSelected: function(value) {
                     root.model.setPickupTimeSignatureNumerator(value)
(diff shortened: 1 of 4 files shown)


shoogle (approved):
Works great, thanks!

Tested on macOS

RomanPudashkin (approved):

PR #12961, opened 2022-08-22
[MU4] Fix #12134: Changing pickup measure breaks playback 

### LINKED ISSUE #12134: [MU4 Issue] Changing pickup measure breaks playback
**Describe the bug**
When you change the pickup measure's length, the playback sticks to both the old length and the new length.

**To Reproduce**

1. Make a new MuseScore file
2. Enable a pickup measure of any length
3. Write some music including that pickup measure and the next one
4. Change the length of the pickup measure

After this, the playback of the next measures will trigger when it would for the original pickup measure length. This breaks the playback in different ways depending on whether you made the pickup measure length shorter or longer.
- If you made it longer, it will start the next measure before it should and again when it should
- If you made it shorter, it will start the next measure when it should and again after it should
Both of these versions interfere with the playback in awkward ways I can't really describe in text. You can try it out to see what I mean.

**Expected behavior**
I'd expect the playback to adapt to the new pickup measure length instead of also keeping the old length and playing both at the same time.

**Platform information**
OS: Windows

**Additional context**

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12134

### DIFF
--- a/src/engraving/playback/playbackmodel.cpp
+++ b/src/engraving/playback/playbackmodel.cpp
@@ -392,8 +392,12 @@ void PlaybackModel::updateEvents(const int tickFrom, const int tickTo, const tra
 bool PlaybackModel::hasToReloadTracks(const std::unordered_set<ElementType>& changedTypes) const
 {
     static const std::unordered_set<ElementType> REQUIRED_TYPES = {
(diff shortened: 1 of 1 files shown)


boryskuzmenko:

Please, merge to master

vpereverzev (approved):

PR #12964, opened 2022-08-22
Fix #8731: Implement clone instrument button

### LINKED ISSUE #8731: [MU4 Task] Instruments cleanup. Clone instrument and reset all formatting
1. implement clone instrument
2. implement the reset all formatting(need to discuss)
3. show a tooltip when hovering over the clone instrument button

image

image

HemantAntony:
Resolves: #8731

### DIFF
--- a/src/instrumentsscene/view/instrumentsettingsmodel.cpp
+++ b/src/instrumentsscene/view/instrumentsettingsmodel.cpp
@@ -23,6 +23,7 @@
 
 #include "log.h"
 #include "translation.h"
+#include "libmscore/excerpt.h"
 
 using namespace mu::instrumentsscene;
 using namespace mu::notation;
@@ -101,6 +102,30 @@ void InstrumentSettingsModel::resetAllFormatting()
 
     const Part* masterPart = masterNotationParts()->part(m_instrumentKey.partId);
     notationParts()->replacePart(m_instrumentKey.partId, masterPart->clone());
+}
+
+void InstrumentSettingsModel::cloneInstrument()
+{
+    if (!notationParts()) {
+        return;
+    }
+
+    Part* part = notationParts()->part(m_instrumentKey.partId)->clone();
+
+    if (!part) {
+        return;
+    }
+
+    part->setPartName(part->partName() + String(" ") + mtrc("instruments", "(copy)"));
+
+    engraving::Excerpt* excerpt = engraving::Excerpt::createExcerptFromPart(part);
+    notation::IMasterNotationPtr masterNotation = context()->currentMasterNotation();
+
+    if (!excerpt || !masterNotation) {
+        return;
+    }
+
+    masterNotation->addExcerpt(excerpt);
 }
 
 QString InstrumentSettingsModel::instrumentName() const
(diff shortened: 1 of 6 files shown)


HemantAntony:
Not sure who to tag, @Tantacrul can you test this? Should I add a tooltip or open the cloned part?

src/instrumentsscene/view/instrumentsettingsmodel.cpp
```diff
@@ -103,6 +104,30 @@ void InstrumentSettingsModel::resetAllFormatting()
...
+        return;
+    }
+
+    Part* part = notationParts()->part(m_instrumentKey.partId)->clone();
+
+    if (!part) {
+        return;
+    }
+
+    part->setPartName(part->partName() + String(" (copy)"));
```
cbjeukendrup, line 119: You need to take translatability into account

HemantAntony: You won't be able to do that, right? You can make " (copy)" translatable. But won't different languages have different sentence structures.

But anyway, I will make it translatable since its already being done for when you duplicate a part in the Parts dialog

Tantacrul:
What is this feature?

> Not sure who to tag, @Tantacrul can you test this? Should I add a tooltip or open the cloned part?

I checked out the build but can't figure out what the proposed functionality is!

HemantAntony:
Its basically to duplicate a part (https://github.com/musescore/MuseScore/issues/8731)

https://user-images.githubusercontent.com/88136054/187025365-a2df6b28-a681-46e0-b518-262e89257c36.mp4

Tantacrul:
OK, I see.

However, I do think the proposed UX is not right. Since it is called 'Duplicate in part', that really does indicate that the cloning should happen in the Parts dialog, not the instruments dialog. This would save two extra steps.

Is there a difference between this and the existing 'Duplicate' function?
image

Tantacrul:
If the functions are the same, I would say that I definitely do not think it should be accessible from the instruments panel simply due to the reason that it **doesn't appear to do anything**. Moreover, we'll be making a lot of changes to the Instruments panel over the next few releases to make it much more useful and I think this button is simply going to get in the way.

cbjeukendrup:
@HemantAntony It looks like this is a severe instance of the usual confusion about instruments, parts and excerpts. The idea of #8731 is that you can have multiple instances of the same "instrument" (the thing that's called `Part` in the code) in one "part" (the thing that's called `Excerpt` in the code); not that you can duplicate `Excerpts` (because that can already be done in the "Parts" dialog.)

HemantAntony:
Oh okay, I see. Thanks!

@cbjeukendrup I am really starting to hate the different terminologies in code :)

cbjeukendrup:
Oh yes, it's terrible. On the other hand, you just need to find it out once, and once you know it, it isn't really problematic. 

I don't think we should change the terminology in the code. I'm happy enough that we have something that makes some sort of sense. It's just unfortunate that the terminology used in real life is so ambiguous.
(Not to mention the extra confusion that will occur when renaming things, and potential compatibility issues...)

PR #12977, opened 2022-08-23
[MU4] Fix #12831: Accessibility navigation commands (Alt+Left/Right) do not set playback position

### LINKED ISSUE #12831: [MU4 Issue] Accessibility navigation commands (Alt+Left/Right) do not set playback position
**Describe the bug**
If you navigate through the score with Alt+Left an Alt+Right, then start playback, playback continues where it left off last time rather than starting at the location you selected.

**To Reproduce**
1. Load a score
2. Start playback
3. Pause it
4. Use Alt+Left/Right to navigate to a specific note in the score (other than where you paused playback)
5. Press Play

**Result**
Playback resume where it left off

**Expected behavior**
Playback should start at the selected note, same as if you had either clicked it or navigated there using Left/Right without Alt

**Screenshots**
https://youtu.be/q1uDP7wmY2o

**Platform information**
 - OS: Windows

**Additional context**
This is pretty critical, as Alt+Left/Right are a blind user's normal way of navigating a score, more so than Left/Right on their own.  That's because Alt+Left/Right navigate through all symbols, not just notes.  And while we sighted users can see based on the playback cursor that plaback is starting in the wrong place, this won't be apparent at all to blind users!

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12831

### DIFF
--- a/src/engraving/libmscore/select.cpp
+++ b/src/engraving/libmscore/select.cpp
@@ -761,6 +761,10 @@ void Selection::updateState()
 
 void Selection::setState(SelState s)
 {
+    if (_state == s) {
+        return;
+    }
+
     _state = s;
     _score->setSelectionChanged(true);
 }
@@ -776,6 +780,8 @@ String Selection::mimeType() const
     case SelState::RANGE:
         return String::fromAscii(mimeStaffListFormat);
     }
+
+    return String();
 }
 
 ByteArray Selection::mimeData() const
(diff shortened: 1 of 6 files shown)


DmitryArefiev:
Tested #12831 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

vpereverzev (approved):

PR #12979, opened 2022-08-23
Grace notes beam direction fix when multiple voices

mike-spa:
Grace note beams on the second voice should be downwards.
image
image


oktophonie (approved):

src/engraving/libmscore/beam.cpp
```diff
@@ -372,18 +372,22 @@ void Beam::layout1()
     //
     // determine beam stem direction
     //
+    ChordRest* firstNote = _elements.front();
```
RomanPudashkin, line 375: You should check whether the list is empty or not

asattely (approved):
Other than Roman's null check, this all looks cromulent to me!

src/engraving/libmscore/beam.cpp
mike-spa: done!

Issue #12950, opened 2022-08-21
[MU4 Issue] Opening MU2 file leads to crash


sammik:
**Describe the bug**
When try to open MU2 file, program crashes.
[Love_of_My_Life_QUARTET.zip](https://github.com/musescore/MuseScore/files/9389992/Love_of_My_Life_QUARTET.zip)

**To Reproduce**
1. Unzip attached score
2. Try to open it
3. Program crashes

<details><summary>Shell log</summary>
/lib/x86_64-linux-gnu/libjack.so.0
/lib/x86_64-linux-gnu/libnss3.so
QEventLoop: Cannot be used without QApplication
QEventLoop: Cannot be used without QApplication
23:45:01.557 | INFO  | main_thread | GlobalModule | onInit: log path: /home/samko/.local/share/MuseScore/MuseScore4Development/logs/MuseScore_220821_234501.log
23:45:01.557 | INFO  | main_thread | GlobalModule | onInit: === Started MuseScore 4.0.0-Development, build number 2897153606 ===
23:45:01.611 | INFO  | main_thread | DiagnosticsModule | onInit: success start crash handler
23:45:02.487 | WARN  | main_thread | IpcSocket  | connect: failed connect to server
23:45:02.487 | WARN  | main_thread | IpcSocket  | connect: failed connect to server
23:45:02.488 | INFO  | main_thread | IpcSocket  | connect: success connected to ipc server
23:45:02.488 | INFO  | 140646845249088 | IpcServer  | listen: id: "2457cab9741d414389e7022b0ee95414" 
23:45:04.801 | WARN  | main_thread | AbstractMenuModel | makeMenuItem: not found action: Modal_Tuning
23:45:04.802 | WARN  | main_thread | AbstractMenuModel | makeMenuItem: not found action: Temperaments
23:45:04.802 | WARN  | main_thread | AbstractMenuModel | makeMenuItem: not found action: notenames
23:45:04.802 | WARN  | main_thread | AbstractMenuModel | makeMenuItem: not found action: tuning
23:45:05.271 | WARN  | 140646086080064 | Qt         | QSocketNotifier: Socket notifiers cannot be enabled or disabled from another thread
23:45:05.271 | WARN  | 140646086080064 | Qt         | QSocketNotifier: Socket notifiers cannot be enabled or disabled from another thread
23:45:19.296 | WARN  | main_thread | Qt         | QQmlComponent: Created graphical object was not placed in the graphics scene.
[...]

PR #12982, opened 2022-08-23
Fix crash that could occur when opening 206 or 114 scores

cbjeukendrup:
Resolves: #12950 

When reading an element in a pre-301 score, it creates a `LinkedObjects` instance and adds that instance to `e.context()->linkIds()`. See https://github.com/musescore/MuseScore/blob/bf553783109f6b6b2d0759b35087fd819aab50eb/src/engraving/libmscore/engravingitem.cpp#L1031-L1059

But sometimes, the created element needs to be deleted, because it turns out we don't need it (e.g. when older scores needed an explicit element but in newer scores we don't need that anymore, or when it turns out that the element is empty.

Since the `LinkedObjects` instance is deleted when the last `EngravingItem` that was linked to it is deleted, we need to remove that `LinkedObjects` instance from the `linkIds()` of the reader context, otherwise it will contain a deleted object which we try to read later on.

At a first glance, it looks like we're already doing that, for example here: https://github.com/musescore/MuseScore/blob/bf553783109f6b6b2d0759b35087fd819aab50eb/src/engraving/rw/compat/read206.cpp#L2649-L2654

Well, the reality is that _this_ is where it goes wrong. We try to remove the `LinkedObjects` from `ctx.linkIds()`, where `ctx` is the `ReadContext` for the _excerpt score_ that is currently being read. But it was added to `e.context()->linkIds()`, where `e.context()` is the context belonging to the reader and thus corresponds with the _master score_, or rather the whole reading process. So we add it to one place, and then try to remove it from a totally different place.

The result is that the deleted `LinkedObjects` instance remains in the `e.context()->linkIds()`, and (occasionally) causes a crash (at least a use-after-free) when we try to use it in https://github.com/musescore/MuseScore/blob/bf553783109f6b6b2d0759b35087fd819aab50eb/src/engraving/rw/compat/read206.cpp#L3381-L3385

Once all of this is discovered, the solution is simple: just use `e.context()->linkIds()` instead of `ctx.linkIds()` when working with `LinkedObjects` and `linkIds()`.

### DIFF
--- a/src/engraving/rw/compat/read114.cpp
+++ b/src/engraving/rw/compat/read114.cpp
@@ -1878,7 +1878,7 @@ static void readMeasure(Measure* m, int staffIdx, XmlReader& e, ReadContext& ctx
                 LOGD("remove keysig c at tick 0");
                 if (ks->links()) {
                     if (ks->links()->size() == 1) {
-                        mu::remove(ctx.linkIds(), ks->links()->lid());
+                        mu::remove(e.context()->linkIds(), ks->links()->lid());
                     }
                 }
             } else {
--- a/src/engraving/rw/compat/read206.cpp
+++ b/src/engraving/rw/compat/read206.cpp
@@ -1820,7 +1820,7 @@ bool Read206::readChordProperties206(XmlReader& e, ReadContext& ctx, Chord* ch)
 //    symbols which were not available for use prior to 3.0
 //---------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


igorkorsukov (approved):

DmitryArefiev:
Tested #12950 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #12990, opened 2022-08-23
Fix #12954: Add child elements of spanners to scans

### LINKED ISSUE #12954: [MU4 Issue] Can't add accidentals to trills
**Describe the bug**
When attempting to add an accidental to a trill line, nothing happens. 

**To Reproduce**
1. Click on a note
2. Add a trill line from the "Lines" palette
3. Click on the trill line
4. Click on an accidental from the "Accidentals" palette
5. Nothing happens

**Expected behavior**
An accidental appears somewhere next to/above the trill line.

**Screenshots**

**Platform information**
OS: macOS 12.5, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2897146315, revision: github-musescore-musescore-89ef1de

**Additional context**
Works in 3.6.

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12954

This issue was due to the fact that spanner child elements (such as accidentals) were not being scanned (thus were also not being added to the BSP tree). I believe this is the safest way to do it, but if someone has a more idiomatic version of this one, let me know and I'll change it.


asattely:
I've implemented a different solution, where I just re-add the override for scanChildren that used to be in 3.6. I have no idea why it was removed, but this seems like the much better solution. cf https://github.com/musescore/MuseScore/blob/4566605d92467b0f5a36b3731b64150500e48583/libmscore/trill.cpp#L233

oktophonie (approved):

RomanPudashkin (approved):

PR #12992, opened 2022-08-23
re: #12691 Avoid applying mag twice for grace notes

asattely:
Continuation from https://github.com/musescore/MuseScore/pull/12691

From what I can tell, mag() used to be 1.0 for grace notes, so the style grace magnitude had to be applied separately. Now, mag() accounts for that so the exception is no longer necessary. There is still some spacing issues in cross-staff grace beams, so I'm pingponging this one back to @mike-spa for all the spacing goodness! For the time being, check the `grace-7` vtest, which illustrates the issue being solved in this PR.

### DIFF
--- a/src/engraving/libmscore/tremolo.cpp
+++ b/src/engraving/libmscore/tremolo.cpp
@@ -419,12 +419,7 @@ void Tremolo::layoutTwoNotesTremolo(double x, double y, double h, double spatium
     // the outer edge of the stems (non-default beam style)
     double x2 = _chord2->stemPosBeam().x();
     if (!stem2 && _chord2->up()) {
-        double nhw = score()->noteHeadWidth();
-        if (_chord2->noteType() != NoteType::NORMAL) {
-            nhw *= score()->styleD(Sid::graceNoteMag);
-        }
-        nhw *= _chord2->mag();
-        x2 -= nhw;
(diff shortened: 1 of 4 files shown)


asattely:
In the slur code, there was a little bit of hackery to account for this weird magnitude problem, so that's been fixed now as well!

src/engraving/libmscore/tremolo.cpp
```diff
@@ -420,9 +420,6 @@ void Tremolo::layoutTwoNotesTremolo(double x, double y, double h, double spatium
     double x2 = _chord2->stemPosBeam().x();
     if (!stem2 && _chord2->up()) {
         double nhw = score()->noteHeadWidth();
-        if (_chord2->noteType() != NoteType::NORMAL) {
-            nhw *= score()->styleD(Sid::graceNoteMag);
-        }
         nhw *= _chord2->mag();
         x2 -= nhw;
```
cbjeukendrup, line 424: I'd maybe simplify this to either
```c++
x2 -= score()->noteHeadWidth() * _chord2->mag();
```
or
```c++
x2 -= _chord2->noteHeadWidth();
```

asattely: Good idea! I made `Chord::noteHeadWidth` public and just used that.

oktophonie (approved):

PR #12996, opened 2022-08-24
Fix #12995: Fixed score date not updating

### LINKED ISSUE #12995: [MU4 Issue] Score date on MuseScore 4 always say 'TODAY'
**Describe the bug**
Dates for recent scores are no longer working as before and always say TODAY

image

**Platform information**
 - MacOS
 - (not checked Windows / Linux)

HemantAntony:
Resolves: #12995

### DIFF
--- a/src/framework/global/types/datetime.cpp
+++ b/src/framework/global/types/datetime.cpp
@@ -135,12 +135,12 @@ int64_t Date::daysTo(const Date& d) const
     toTM(meTM, *this);
 
     std::tm otherTM = {};
-    toTM(otherTM, *this);
+    toTM(otherTM, d);
 
     std::time_t meTime = std::mktime(&meTM);
     std::time_t otherTime = std::mktime(&otherTM);
     if (meTime != (std::time_t)(-1) && otherTime != (std::time_t)(-1)) {
-        double difference = std::difftime(meTime, otherTime) / (60 * 60 * 24);
+        double difference = std::difftime(otherTime, meTime) / (60 * 60 * 24);
         return difference;
     }
     return 0;


cbjeukendrup:
Uhm oh, I was just working on it too. Anyway, we have found the same solution :)

HemantAntony:
Ah I wasn't sure about my solution :)

src/framework/global/dataformatter.cpp
```diff
@@ -59,7 +59,7 @@ String DataFormatter::formatTimeSince(const Date& date)
     }
 
     if (weeks <= 4) {
-        return mtrc("global", "%n week(s) ago", nullptr, weeks);
+        return mtrc("global", "%n weeks ago", nullptr, weeks);
```
cbjeukendrup, line 62: Why these changes? It is intentional that these strings contain these parentheses, which will be translated away on Transifex (for singular/plural/... separately).

HemantAntony: There won't be a case for singular. Its already covered by the if statement above. It will be called 'Last week' instead of '1 week ago'. Similarly for days. That is, unless there are edge cases

Jojo-Schmitz: True, but might do for other forems that some languages have

Jojo-Schmitz: The consense was, when I proposed and added those, to have the strings as they are and translated them accordingly.

Jojo-Schmitz: Although I'd rather delete the singular cases from the code and add an instruction to translate those as Today, Yesterday, Last week, Last month etc.
BTW: why isn't there a "Last year"? And why didn't you treat the months case the same?

Jojo-Schmitz: Check 35b94f5 for a possible (and IMHO better) solution

cbjeukendrup: @Jojo-Schmitz The problem with that solution was that for some languages, the "singular" value is not used only for "1", but also for "11". So with that solution, in some languages "11 months ago" would also be translated with "Last month".

Jojo-Schmitz: Sure about that? Isn't then (with 11) a 3rd form used?

shoogle: @HemantAntony, I think it's best to leave the parentheses in. It's [normal to write plurals this way in Qt-based source code](https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals) so translators will be used to seeing it with parentheses. It's purely a coincidence that English only has 2 forms of plural (`n = 1` and `n != 1`) and that one of those forms corresponds exactly to the "yesterday" case (`n = 1`). Other languages have [different forms of plural](https://doc.qt.io/archives/qq/qq19-plurals.html#whatstheproblemwithpluralforms) for [different values of `n`](https://doc.qt.io/qt-5/i18n-plural-rules.html).

Jojo-Schmitz: On Transifex I see "1", "other" and "plural", for some languages, and "1" to me clearly means "just one item", regardless whether this is called "singular" in that language or not.

HemantAntony: I have re-added the brackets. I thought it was an unintentional decision :)

cbjeukendrup: See the table at the bottom of https://doc.qt.io/archives/qq/qq19-plurals.html

cbjeukendrup (approved):

src/framework/global/dataformatter.cpp
Jojo-Schmitz: Well, at least for days and weeks the 11 case just won't ever happen. The years case is not covered by that extra "Last year", so not really affected here, leaves the case of "11 months". And a certian uncertainty how Transifex handles that

Jojo-Schmitz: So we could keep my propose change at least for days and weeks, and the comment for years probably too

Jojo-Schmitz: Languages affected by that 11 (21, 31. etc) problem: Latvian (Transifex: 0, 1, other), Lithuanian (Transifex: 1, few, many, other), Macedonian (no such translation for MuseScore) and Russian  (Transifex: 1, few, many, other), additionaly for 101, 201, etc: Czech (Transifex: 1, few, many, other)

cbjeukendrup: Ah... so Transifex isn't fully compatible with Qt in this regard. So _that_ must be the reason that `lrelease` says things like this for some languages:
```
Removed plural forms as the target language has less forms.
If this sounds wrong, possibly the target language is not set or recognized.
```
No idea whether/how we can fix it. I'm afraid this will have to be reported to Transifex.

Jojo-Schmitz: What languages does it report this on?

cbjeukendrup: See for example https://github.com/musescore/MuseScore/runs/7995324566?check_suite_focus=true#step:11:1 for the full lrelease output (that's just from a random GitHub Actions run)

Jojo-Schmitz: Picking on one, fr (French): it has 1, many and other on Transifex, but only 2 variants in that Qt table

Jojo-Schmitz: It has only 2 in Qt Linguist too... pl (Polish ) has 4 variants on Transifex, 3 in Qt Linguist (and in that Qt table too)

Issue #13000, opened 2022-08-24
[MU4 Issue] UI: Timeline panel's scrollbar is click through


fp22june:
(Info added by @cbjeukendrup)

**Steps to reproduce**
1. Open a score with multiple staves (4+ recommended)
2. Open the Timeline, and if necessary, undock it and resize it horizontally, so that the scrollbars become visible
3. Try to move the scrollbars

**Expected:** scrollbar moves, follows mouse, and view scrolls accordingly
**Actual:** not the scrollbar reacts, but the view reacts to mouse press as if the scrollbar doesn't exist

**Additional Context**
Happens in Alpha2

cbjeukendrup:
It looks like the first issue can be solved with a reasonably small amount of effort. About the second issue, I'm not sure, I don't understand immediately what's causing that. Maybe I'll find it.

cbjeukendrup:
@boryskuzmenko I've pushed a fix for the first issue to the PR. I'm doing a quick attempt to find the cause of the second issue, but it looks like it would take more time than it's worth.

boryskuzmenko:
🟢 **FIXED** in https://github.com/musescore/MuseScore/pull/13003
Except one small bug which was reported separately https://github.com/musescore/MuseScore/issues/13086

PR #13003, opened 2022-08-25
Fix scrollbars in Timeline

cbjeukendrup:
Now, all child widgets, including scrollbars, receive events when they should.

(This is in fact still not a super thorough advanced solution; that would be recreating almost all event handling of QWidgetWindow (a private Qt class) in WidgetView etc.. But that would be such a large undertaking with so little added value, that it's not worth the effort.)

Resolves: #13000


RomanPudashkin (approved):

boryskuzmenko:

Please, merge to master

PR #13004, opened 2022-08-25
Fix #13002: Vertical text alignment buttons are swapped

### LINKED ISSUE #13002: [MU4 Issue] Vertical text alignment buttons are swapped
**Describe the bug**

Icons and functionality for some of the vertical text alignment buttons have been switched compared to MuseScore 3.

**To find the vertical alignment buttons**

1. Select a text object (e.g. the title).
2. Open the Inspector / Properties panel.
3. Look for Text > Alignment > Vertical alignment options.

MU3 Inspector panel | MU4 Properties panel
:---:|:---:
image | image

### Wrong icons

Look closely at the icons for the final two buttons and you'll see that they have swapped places.

MuseScore 3 | MuseScore 4
:---:|:---:
image | image
Dotted line is lower on the left | Dotted line is lower on the right

### Wrong functionality

Three of the buttons have been "wired" incorrectly. Only valign-center (image) is correct. The others have been mixed up.

To see this:

1. Select the "Title" or "Composer" text in the title frame.

image

2. Check which vertical alignment option is highlighted in the UI.

Selected text | MuseScore 3 | MuseScore 4
---|---|---
Title | image | image
Composer | image | image

### Expected behavior

For text alignment buttons:

- Icons should be in the same order as they were in MuseScore 3.
- Buttons should have the same functionality as they did in MuseScore 3.

Once these are both fixed, here is where frame text should appear for each alignment option in turn:

image image image image

You can check they are correct by saving the score in MSCX format and examining it in a text editor. Look for the `<align>` tags.

HemantAntony:
Resolves: #13002 


HemantAntony:
Am I the only one that thinks there should be tooltips for those buttons in Properties?

src/framework/ui/view/iconcodes.h
```diff
@@ -99,8 +99,8 @@ class IconCode
         ARROW_UP = 0xEF37,
         POSITION_ARROWS = 0xEF38,
         TEXT_ALIGN_BASELINE = 0xEF39,
-        TEXT_ALIGN_ABOVE = 0xEF3A,
-        TEXT_ALIGN_UNDER = 0xEF3B,
+        TEXT_ALIGN_UNDER = 0xEF3A,
+        TEXT_ALIGN_ABOVE = 0xEF3B,
```
Jojo-Schmitz, line 103: Strange, I always believed "UNDER" being the opposite/counterpart to "OVER", and "ABOVE" the opposite/counterpart of "BELOW"...

HemantAntony: I agree, Above is the opposite to Below. I'm not sure of the Under and Over though

Jojo-Schmitz: So renaming TEXT_ALIGN_UNDER to TEXT_ALIGN_BELOW seems a good option. Ot even keeps the legnth, so doesn't srew up score formatting

cbjeukendrup: Why not just `TEXT_ALIGN_TOP` and `TEXT_ALIGN_BOTTOM`?

HemantAntony: @cbjeukendrup Yeah I agree with you. Excel and Google Sheets uses them same terms

Jojo-Schmitz: Works too, and probably even better as it matches other similar defines

shoogle: Thanks, I was about to make the same suggestion (i.e use "Top" and "Bottom" instead of "Above" and "Below"). When the reference point is the same, text that uses the **top** alignment option will actually appear _below_ text that uses the **bottom** alignment option.

image

Similarly, **align left** will put text on the _right_ of the reference point, so we should be consistent with terminology.

HemantAntony: @shoogle Nice! I was also thinking of adding tooltips to those buttons. There aren't any now. What do you think?

shoogle: @HemantAntony, sure! I was thinking the same and was about to create an issue. But let's merge this and put the tooltips in a separate PR.

shoogle (approved):

src/framework/ui/view/iconcodes.h
HemantAntony: Okay cool!

cbjeukendrup (approved):

PR #13011, opened 2022-08-25
fixed #12832: Concert pitch button does not read it state

### LINKED ISSUE #12832: [MU4 Issue] Accessibility: concert pitch button does not read it state
**Describe the bug**
When navigating via keybaord to the concert pitch button, the screen reader does not read its state (on or off)

**To Reproduce**
1. Start screen reader (tested with NVDA and Narrator)
2. From main score, press Tab to reach the bottom toolbar
3. Press Right to reach the Concert Pitch button

**Result**
Screen reader does read the name of the button, but not its state, so you have no idea if it is enabled or not.  Pressing Space toggles the button, but again, there is no feedback

**Expected behavior**
Button should report that it is a toggle, and read its state

**Platform information**
 - OS: Windows

**Additional context**
In general, it seems all toggle buttons report their initial state, so simply changing the type of the control to toggle should fix this.  But, I also note that even for proper toggle buttons in MuseScore, when you press space, it doesn't generally report the result of the action.  Not sure if that's something we have much control over, but it feels off to me.

Eism:
Resolves: #12832

### DIFF
--- a/src/notation/qml/MuseScore/NotationScene/ConcertPitchControl.qml
+++ b/src/notation/qml/MuseScore/NotationScene/ConcertPitchControl.qml
@@ -37,6 +37,9 @@ FlatButton {
     isNarrow: true
     margins: 4
 
+    navigation.accessible.role: MUAccessible.CheckBox
+    navigation.accessible.checked: checked
+
     contentItem: Row {
         spacing: 6
 


RomanPudashkin (approved):

DmitryArefiev:
Tested #12832 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #13013, opened 2022-08-25
fixed #12807: No NVDA feedback within score options panel

### LINKED ISSUE #12807: [MU4 Issue] Accessibility: no NVDA feedback within score options panel
**Describe the bug**
NVDA users cannot select the key or time signature for their scores, or the tempo or length, using the wizard, because screen reader does not reader the contents of these options.  Narrator works, however.

**To Reproduce**
1. Start NVDA
2. Create new score for any instruments
3. In the "Additional score information" panel, select the key signature control
4. Use keyboard to navigate through the keys

**Result**
Nothing is read aloud

**Expected behavior**
Each key should be read as you navigate

**Platform information**
 - OS: Windows / NVDA

**Additional context**
Same for the time signature, tempo, and measures control

Eism:
Resolves: #12807

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -167,6 +167,11 @@ const IAccessible* AccessibleItem::accessibleChild(size_t i) const
     return nullptr;
 }
 
+QWindow* AccessibleItem::accessibleWindow() const
+{
+    return nullptr;
+}
+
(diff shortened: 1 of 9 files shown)


igorkorsukov (approved):

DmitryArefiev:
Tested # on Win10, NVDA - **FIXED**

But found the same occurs on Linux.  Logged separately #13311

PR #13024, opened 2022-08-26
upload_score_fix

RomanPudashkin:


Eism (approved):

src/appshell/view/publish/publishtoolbarmodel.cpp
```diff
@@ -36,8 +36,7 @@ void PublishToolBarModel::load()
 
     MenuItemList items {
         makeMenuItem("print"),
-        // Todo(save-to-cloud): hidden until fully implemented
-        //makeMenuItem("file-publish"),
+        makeMenuItem("file-publish"),
```
cbjeukendrup, line 39: It would also need to be re-exposed in AppMenuModel and AskSaveLocationTypeDialog.qml

RomanPudashkin: fixed

PR #13029, opened 2022-08-26
Stem slash redesign

mike-spa:
Fixing our stem slashes, which were all over the place (partially also due to all the work on beams). The new vtests (and plenty of old ones) should clearly show the problems and the improvement.


oktophonie (approved):

src/engraving/libmscore/stemslash.cpp
```diff
@@ -67,22 +68,62 @@ void StemSlash::setLine(const LineF& l)
...
-    double w  = chord()->upNote()->bboxRightPos() * .7;
-    setLine(LineF(PointF(x + w, y - h2), PointF(x - w, y + h2)));
+
+    double endX = 0;
+    double endY = 0;
+
+    if (hook) {
+        auto musicFont = score()->styleSt(Sid::MusicalSymbolFont);
+        // HACK: adjust slash angle for fonts with "fat" hooks. In future, we must use smufl cutOut
+        if (c->ticks() < Fraction(1, 8) && !straight
```
cbjeukendrup, line 105: Could we also use `c->beams() >= 2` or something in that direction here?

src/engraving/libmscore/beam.cpp
```diff
@@ -47,6 +47,7 @@
 #include "system.h"
 #include "tremolo.h"
 #include "tuplet.h"
+#include "stemslash.h"
```
cbjeukendrup, line 50: I'd prefer to keep the #includes sorted alphabetically, just to keep it a bit tidy and avoid chaos. (Recently I did a PR to fix the sorting in basically all libmscore files, so it would be lovely to try to keep it clean from now on 🙂)

(Please forgive me for complaining about such a trivial useless thing)

src/engraving/libmscore/stemslash.cpp
```diff
@@ -57,7 +58,7 @@ void StemSlash::draw(mu::draw::Painter* painter) const
 void StemSlash::setLine(const LineF& l)
 {
     line = l;
-    double w = score()->styleMM(Sid::stemWidth) * .5;
+    double w = _width * .5;
     setbbox(RectF(line.p1(), line.p2()).normalized().adjusted(-w, -w, 2.0 * w, 2.0 * w));
 }
```
cbjeukendrup, line 63: Since this method is called in exactly one place, what do you think about just inlining this code in the layout() method, and getting rid of the _width member? (Or do you foresee the method might be useful in other places in the future?)

src/engraving/libmscore/stemslash.cpp
mike-spa: Ha! Thanks, didn't know we had that

src/engraving/libmscore/beam.cpp
mike-spa: No worries, we all like tidiness ;)

src/engraving/libmscore/stemslash.cpp
mike-spa: Yeah I actually thought the same, gonna do that

src/engraving/style/styledef.h
```diff
@@ -215,6 +215,9 @@ enum class Sid {
     staffLineWidth,
     ledgerLineWidth,
     ledgerLineLength,
+    stemSlashHeight,
+    stemSlashAngle,
+    stemSlashWidth,
```
cbjeukendrup, line 220: When seeing `stemSlashHeight` and `stemSlashWidth` so close to each other, one might be confused about what these settings do. Have I understood correctly that `stemSlashHeight` determines the position, and `stemSlashWidth` the line thickness? Perhaps we could rename `stemSlashWidth` to `stemSlashThickness`, and `stemSlashHeight` to `stemSlashPosition`? (To me the word "height" seems to suggest that it has to do with the size of the slash itself, which is not the case.)
Let me know what you think.

cbjeukendrup (approved):
One question about naming things (the most difficult problem in computer science), but for the rest I'm happy!

src/engraving/style/styledef.h
oktophonie: MuseScore historically has used 'width' rather than 'thickness' (mostly) in its settings/variables. But I agree that using 'thickness' would be much less confusing in anything new (and ideally, one day, when overhauling the old ones). SMuFL also uses 'thickness' in its `engravingDefaults`.

mike-spa: Yup, I definitely prefer position and thickness, as you suggest, I just kept those name cause many other things are named that way, as Simon said. Gonna change it now.

Issue #13032, opened 2022-08-26
[MU4 Issue] Crash after actions with clef on staff


boryskuzmenko:
**To Reproduce**
1. Open any score (e,g, 
[Hairpins.mscz.zip](https://github.com/musescore/MuseScore/files/9434445/Hairpins.mscz.zip)
2. Select clef and in its Properties > Appearance change the Leading space parameter
3. Via Staff/part properties make clef hidden (Show Clef to False) and Apply
4. Apply Undo

**Expected behavior**
Last action is reverted in case of using Undo

**Screenshots**

https://user-images.githubusercontent.com/101250347/186940114-cdfb9059-c337-4739-8919-69516ba3cb0b.mov

DmitryArefiev:
Fixed in #13036

PR #13036, opened 2022-08-26
Prevent crash after performing specific actions with clef

cbjeukendrup:
Resolves: #13032

Problem: we can't delete the clef, because it is still referenced (and necessary) in the undo stack.

I'm not 100% sure whether this creates a memory leak, but I fear it does. Anyway, a memory leak is less bad than a crash, but at some point we'll really have to properly sort out memory management regarding the combination of the layout process with the undo stack.

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -3758,7 +3758,9 @@ void Measure::addSystemHeader(bool isFirstSystem)
                 clef->layout();
             } else if (clef) {
                 clef->parentItem()->remove(clef);
-                delete clef;
+                if (clef->generated()) {
+                    delete clef;
+                }
             }
             //cSegment->createShape(staffIdx);
             cSegment->setEnabled(true);


src/engraving/libmscore/measure.cpp
```diff
@@ -3746,7 +3746,6 @@ void Measure::addSystemHeader(bool isFirstSystem)
                 clef->layout();
             } else if (clef) {
                 clef->parentItem()->remove(clef);
-                delete clef;
```
RomanPudashkin, line 3749: As far as I remember, we cannot remove generated elements, so they don't go to UndoStack. Can we rewrite this code as follows:

```
if (clef->generated()) {
    delete clef;
}
```

Will this fix the crash?

cbjeukendrup: Thanks, that solution works too!

RomanPudashkin (approved):

cbjeukendrup:
There's a different issue that's revealed by this fix: 
1. Instead of modifying "Leading space", modify "Offset"
2. Staff Part Properties > disable "Show clef"
3. undo

expected: clef comes back and it's properties are the same as before (modified offset is kept)
actual: clef comes back but offset is lost (and technically, it's a different clef, but that doesn't matter from user POV)

This also occurs in MU3.

In MU3, if you press undo multiple times instead of once, a crash will occur, but in MU4 this crash does not occur; instead, nothing happens.

DmitryArefiev:
> There's a different issue that's revealed by this fix:
> 
>     1. Instead of modifying "Leading space", modify "Offset"
> 
>     2. Staff Part Properties > disable "Show clef"
> 
>     3. undo
> 
> 
> expected: clef comes back and it's properties are the same as before (modified offset is kept) actual: clef comes back but offset is lost (and technically, it's a different clef, but that doesn't matter from user POV)
> 
> This also occurs in MU3.
> 
> In MU3, if you press undo multiple times instead of once, a crash will occur, but in MU4 this crash does not occur; instead, nothing happens.

@cbjeukendrup But it's not fixed in your current PR, right?

cbjeukendrup:
@DmitryArefiev Indeed, not in this PR. Requires a bit more research :)

DmitryArefiev:
Cool. Thanks!

DmitryArefiev:
Tested # on Win10, Mac12, LinuxMint 20.2 - **FIXED**

DmitryArefiev:
> There's a different issue that's revealed by this fix:
> 
>     1. Instead of modifying "Leading space", modify "Offset"
> 
>     2. Staff Part Properties > disable "Show clef"
> 
>     3. undo
> 
> 
> expected: clef comes back and it's properties are the same as before (modified offset is kept) actual: clef comes back but offset is lost (and technically, it's a different clef, but that doesn't matter from user POV)
> 
> This also occurs in MU3.
> 
> In MU3, if you press undo multiple times instead of once, a crash will occur, but in MU4 this crash does not occur; instead, nothing happens.

Created #13266 for this

Issue #12947, opened 2022-08-21
[MU4 Issue] Spanning element vertical flip (x key) doesn't work at linebreaks


schoettl:
**Describe the bug**
When I press `x` to vertically flip the position of *multiple* spanning elements, this does not work for those elements that go over two staff lines.

- Flipping the position for a single spanning element that goes over two staff lines does work.
- Flipping elements also work when multiple elements are manually selected (by holding the Ctrl key)!

**To Reproduce**
1. Select first line, right click, select all similar elements
2. Press `x`
3. Check if all selected elements have been flipped

**Expected behavior**
The `x` command should change all selected elements.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/1301843/185792830-340cb2f0-9a18-45c7-9a98-9e52fc4cf7b9.png)
![grafik](https://user-images.githubusercontent.com/1301843/185792915-e247c488-ae0d-4e15-92fb-7ec55093216a.png)

**Platform information**
 - OS: Linux

**Additional context**
Same problem on latest MuseScore 4 and MuseScore 3.6.2.

Sample score to reproduce:
[test-score.zip](https://github.com/musescore/MuseScore/files/9389167/test-score.zip)

MarcSabatella:
Most likely, it’s getting flipped twice, since effectively it is selected twice - the two segments of the line are separate elements.  We’ve had to special-case a number of similar situations before, probably should here too.

schoettl:
Ah, that makes sense. Same thing applies to other spanning elements, e.g. the keyboard bracket. But not for slurs.

Should I try to find the problem / special case handling in the code?

MarcSabatella:
If you can that would be fantastic!  It would be easier for someone who already know their way around, of course, but if you have any interest in getting involved in development, this is exactly the way we often recommend getting started - "scratching your own itch"!

schoettl:
Looks like the special cases are here in edit.cpp `cmdFlip`

```c++
        …
        } else if (e->isTremolo()) {
            auto tremolo = toTremolo(e);
            flipOnce(tremolo, [tremolo]() {
                DirectionV dir = tremolo->up() ? DirectionV::DOWN : DirectionV::UP;
                tremolo->undoChangeProperty(Pid::STEM_DIRECTION, dir);
            });
        } else if (…) {
        …
```

But I have some problems in understanding:

0. Is this the right place and `flipOnce` the function which fixes this?
[...]

PR #13043, opened 2022-08-27
Fix flipping some elements

cbjeukendrup:
- some elements were just forgotten
- spanners spanning an even number of systems were flipped an even number of times, so in the end they are not flipped

Resolves: #12947

Of course, such a long list of `e->is…()` checks is ugly and not very maintainable (easy to forget to add new elements). But adding all required functions as members to the EngravingItem class isn't great either; I don't like to clutter this huge class even further with single-use methods (which are still visible in every cpp file so increase compilation time...).


RomanPudashkin (approved):

DmitryArefiev:
Tested #12947 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #13066, opened 2022-08-29
Slur over not-enabled segments bug fix

mike-spa:
Fixes this bug:

https://user-images.githubusercontent.com/93707756/187212878-000af9cd-d623-469d-b83c-171fca5851dd.mp4


src/engraving/libmscore/slur.cpp
```diff
@@ -473,6 +473,9 @@ void SlurSegment::avoidCollisions(PointF& pp1, PointF& p2, PointF& p3, PointF& p
         }
         // Check collisions
         for (Segment* seg : segList) {
+            if (!seg->enabled()) {
+                continue;
+            }
```
cbjeukendrup, line 478: It might be a little bit more efficient to filter out disabled segments directly when computing `segList`. How about that?

mike-spa: right, good point!

cbjeukendrup (approved):

PR #13083, opened 2022-08-30
[MU4] Fix #12480: Inspector doesn't update when changing property default in Style dialog

### LINKED ISSUE #12480: [MU4 Issue] Inspector doesn't update when changing property default in Style dialog
**Describe the bug**
If you edit the default value for a property using the Style dialog, the new value is not reflected in the Inspector, even though the property of the item itself does change.

**To Reproduce**
1. On a score, select an element in the Inspector (for example a note) and find a "styled" property in the Inspector.
2. Open the Style dialog, and find the corresponding style setting.
3. Change this setting. Notice that the item on the score is updated (unless the property of that item was modified of course), but that the Inspector keeps showing the old value. 

**Expected behavior**
If the value of a property of an item is updated, the Inspector should reflect this, even if the update happens because the property is following its corresponding style setting that is updated.

**Screenshots**

https://user-images.githubusercontent.com/48658420/179811844-1fdec3df-310c-419b-9d35-53b60ca23cbc.mov

**Platform information**
 - OS: reported on macOS

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12480
Resolves: https://github.com/musescore/MuseScore/issues/12481


DmitryArefiev:
Tested #12481 #12480 on Win10, LinuxMint 20.2 - **FIXED**

PR #13084, opened 2022-08-30
Add information about MS version to sign-up and authorize links on musescore

bentl:
Add information about MS version to sign-up and authorize links on musescore.

Sign-up link is opened by clicking on 'Create new account' button. Oauth flow is not started at this point. I'm going to check possibility later and add as separate PR.

For authorization I set ModifyParametersFunction to QOAuth2AuthorizationCodeFlow to add one more query parameter to the URL. 

### DIFF
--- a/src/cloud/iauthorizationservice.h
+++ b/src/cloud/iauthorizationservice.h
@@ -35,6 +35,7 @@ class IAuthorizationService : MODULE_EXPORT_INTERFACE
 public:
     virtual ~IAuthorizationService() = default;
 
+    virtual void signUp() = 0;
     virtual void signIn() = 0;
     virtual void signOut() = 0;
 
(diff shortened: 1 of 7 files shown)


RomanPudashkin (approved):

cbjeukendrup:
About the signUp URL, this just opens the sign up page in the browser, but doesn't start the OAuth flow, right? So the user would first click "Create account" in the desktop app, then create the account online, then go back to the desktop app and click "Sign in" and then in the browser log in with the newly created account. 
From UX point of view, I think it would be nice if the user could just start the OAuth flow by clicking "Create account", so that when the account is created, the user is immediately redirected to the app and logged in in the app. So actually, the "Sign in" and "Create account" buttons would do almost exactly the same, but the only difference is that "Sign in" redirects you to https://musescore.com/user/login and "Create account" to https://musescore.com/user/register. Would that be possible?

bentl:
@cbjeukendrup Thanks for the comment. I have a task in my todo list to research possibility to implement sign-up with oauth in the way as you described. If it would be possible I'll add as a new PR.
It requires some more effort so firstly I added sign-up flow just to have MS version on register flow.

cbjeukendrup:
Okay, cool!

PR #13101, opened 2022-08-31
A definitive solution to cross-beam spacing (also Fix #13120)

### LINKED ISSUE #13120: [MU4 Issue] Note positions are not recalculated when chord is flipped due to beam drag
**Describe the bug**
Note positions in a chord should be recalculated if the chord is flipped following a beam drag (see video).

https://user-images.githubusercontent.com/93707756/187906569-2633b54b-f15b-499f-acf9-3d545436ad67.mp4

The solution in itself seems ovious, but it has implications for horizontal spacing which need to be accounted for (see here the notes are positioned correcty, but horizontal spacing now is wrong). Currently working on it.
image

mike-spa:
(UPDATE: last commit also fixes #13120. The system implemented in this PR made the solution much simpler).

Horizontal spacing needs special consideration when chords cross up/down a beam. My previous solution was very limited, as it was only considering staff-moves (i.e., decide the spacing based on which staff the chord is), and that way doesn't take into account that
- Cross-staff beams default in-between staves, but they can be dragged above/below the staves, in which case the spacing becomes wrong.
- Beams that don't have any cross-staff chords may still have cross-beam chords and therefore need spacing adjustment.
(see videos for quick understanding).
My new solution looks at the actual position of the beam with respect to the chord. This way, the spacing is correct in all situations. The difference should be clear from the videos and from the vtests.

Before:

https://user-images.githubusercontent.com/93707756/187701699-c3dba4b6-813a-4951-9fe8-8a645667a2fb.mp4

After:

https://user-images.githubusercontent.com/93707756/187701721-dc5d98dc-3b5d-4afa-80c7-252f4938edfc.mp4


oktophonie (approved):

src/engraving/layout/layoutsystem.cpp
```diff
@@ -107,7 +107,9 @@ System* LayoutSystem::collectSystem(const LayoutOptions& options, LayoutContext&
     while (ctx.curMeasure) {      // collect measure for system
         oldSystem = ctx.curMeasure->system();
         system->appendMeasure(ctx.curMeasure);
-
+        if (system->hasCrossStaffOrModifiedBeams()) {
```
cbjeukendrup, line 110: It looks like system->hasCrossStaffOrModifiedBeams() is a somewhat expensive operation. Would it be possible to call it only once, outside the loop?

src/engraving/libmscore/segment.cpp
```diff
@@ -2706,28 +2706,63 @@ bool Segment::hasAccidentals() const
...
+    for (EngravingItem* e : elist()) {
+        if (!e || !e->isChordRest() || !e->staff()->visible()) {
+            continue;
+        }
+        ChordRest* thisCR = toChordRest(e);
+        if (!thisCR->visible() || thisCR->isFullMeasureRest()) {
+            continue;
+        }
+        if (!thisCR->beam()) {
+            return;
```
cbjeukendrup, line 2734: Just to be sure, is it intended that in this case, _crossBeamType stays `{false, false}` even though `upDown` or `downUp` might be true? (I guess this will mean that if on at least _one_ staff there is no "cross beam", no adjustment will be made, is that right?)

Also, it looks like in the following case, the `_crossBeamType` of the first segment is `{false, false}`, so the spacing between the first two eight notes won't be adjusted. I suppose that this is intentional for now, and we might refine the algorithm even further in the future?
Scherm­afbeelding 2022-09-02 om 17 40 21

src/engraving/libmscore/segment.cpp
```diff
@@ -2706,28 +2706,63 @@ bool Segment::hasAccidentals() const
...
+            if (thisCR->up() && !nextCR->up()) {
+                upDown = true;
             }
-            if (toChordRest(el)->staffMove() < 0 && toChordRest(el)->beam()) {
-                crossStaffContent.movedUp = true;
+            if (!thisCR->up() && nextCR->up()) {
+                downUp = true;
+            }
+            if (upDown && downUp) {
+                return;
```
cbjeukendrup, line 2761: This basically means that for any CrossBeamType, it will never happen that both upDown and downUp are true at the same time, is that right? (I was thinking, in that case, you could make it an enum with 3 cases as well. However, I'm not sure whether that has any benefits; it looks like it's quite convenient like it is now.)

src/engraving/layout/layoutsystem.cpp
mike-spa: Quick answer is no, because every loop adds one more measure, which may add cross-staff or modified beams to the system, so I need to re-check. I may try to make it more efficient by only looking at the last measure when it's added, but that may cause issues if beams are connected across measures. I'll look into it.

src/engraving/libmscore/segment.cpp
mike-spa: Quick answer: yes. Adjusting for cross-beams should never interfere with spacing of "normal" notes in any other voice or staff, so I'm being very strict and I'm not applying any correction if there is even a single non-cross-beam note anywhere in the same segment or the next. Second measure is as expected here, first measure could be improved. I think I'll refine it at a later stage. For now, I'd say it can go as it is.

src/engraving/libmscore/segment.cpp
mike-spa: Indeed. If the same segment contains both a down/up and an up/down situation, we shouldn't correct for any of them.

RomanPudashkin (approved):

RomanPudashkin:
Please rebase your PR, there are a lot of redundant commits

mike-spa:
> Please rebase your PR, there are a lot of redundant commits

Done, thanks! 👍

Issue #12794, opened 2022-08-11
[MU4 Issue] User Text Styles missing in Properties tab


sammik:
**Describe the bug**
In Properties tab, User Text Styles are missing, it is not possible, to set new custom formattings.

**To Reproduce**
1. Create Staff Text (select note, press Ctrl+t)
2. Click on Properties - More
3. Scroll down to Text Style
4. no User Style there

**Expected behavior**

**Platform information**
 - OS: [reported on Linux]

MarcSabatella:
Ouch, pretty critical omission, surprised it wasn't noticed yet!  The user styles can still be customized via Format / Style / Text Styles, but no way I can see to actually assign one.  I can verify they do work when loading scores from MuseScore 3 that are using user styles.  The style settings are honored, but the name of the style shows as empty in the Properties window.  Hopefully a simple fix!

Tantacrul:
Unfortunately, this is no longer in scope for MS4.0

Unless @oktophonie - you think engraving can handle it? For now, I'm marking as a P0 and assigning to 4.x.

When we reopen this issue (after MS4.0 is out), this will sit at the top of the priority list.

cbjeukendrup:
It looks like this is as simple as adding some values to the list in the Inspector. The engraving side should just work, unchanged since MS3.

Tantacrul:
OK, I'll pop it in post beta as something we may get time to fix... but might not too.

MarcSabatella:
FWIW, since MU3 files display user styes correctly, definitely not an engraving issue.  And since Format / Style / Text Styles displays the user styles correctly, probably there is code there that could just be copied over for the Properties panel.

PR #13111, opened 2022-08-31
Expose all text styles, including user styles, in the Inspector

cbjeukendrup:
Resolves: #12794

Include user text styles, taking into account their customized name, if any.

@Tantacrul This will make it slightly different from MS3, where only a subset of text styles was displayed. For fingering text elements, only 4 styles were shown, for other text elements 40 styles or so; now we will show all 56. All these styles are exposed in Format > Style > Text Styles anyway, in MS3 too. I think they could be useful, for example, when you use a text element to create a very special tuplet marking that is not supported using a 'real' tuplet, you will want to use the Tuplet text style, so that it will look the same as other tuplets. However, if there are styles that you think should be excluded, we can easily do so.

### DIFF
--- a/src/inspector/models/text/textsettingsmodel.cpp
+++ b/src/inspector/models/text/textsettingsmodel.cpp
@@ -26,7 +26,8 @@
 #include "types/commontypes.h"
 #include "types/texttypes.h"
 
-#include "libmscore/textbase.h"
+#include "engraving/libmscore/textbase.h"
+#include "engraving/types/typesconv.h"
 
 #include "translation.h"
 #include "log.h"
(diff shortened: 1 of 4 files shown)


MarcSabatella:
FWIW, I always had mixed feelings about the restriction we did in MU3 to only show some styles in the Inspector dropdown.  The idea was to make it harder for people to shoot themselves in the foot by picking a style that really wouldn't make sense for the selected element type, but, meh.

Tantacrul:
Perhaps @bkunda you could just sanity check this?

Sorry but I'm definitely going to be up the walls tomorrow,Mon/Tue

bkunda:
@cbjeukendrup I don't see any problem with exposing all these text styles in this dropdown menu. It seems useful to have them there, the list isn't offensively long, and it is all sufficiently buried under the "More..." button such that, should one be inclined to look for something there in the first place, it would probably be expected that _more_ options (not fewer) would be available. So this gets the 👍🏻👍🏻 from me.

RomanPudashkin (approved):

RomanPudashkin:

cbjeukendrup:
@RomanPudashkin Done!

PR #13121, opened 2022-09-01
(test)

mike-spa:
(just collecting the vtest, I need to see if this block of code can be removed, as I think)


HemantAntony:
Can't you run the vtests locally? like the utests?

mike-spa:
> Can't you run the vtests locally? like the utests?

If it is possible, I surely don't know how 😂 (But I don't think it is possible on windows)

PR #13128, opened 2022-09-02
Force screen readers to say accessible description on all platforms

shoogle:
Previously VoiceOver (macOS) and Narrator (Windows) did not say accessible descriptions, which meant certain information was not available to users of those screen readers.

### DIFF
--- a/src/engraving/accessibility/accessibleitem.cpp
+++ b/src/engraving/accessibility/accessibleitem.cpp
@@ -199,13 +199,7 @@ QString AccessibleItem::accessibleDescription() const
         return QString();
     }
 
-    QString result;
-#ifdef Q_OS_MACOS
-    result = accessibleName() + " ";
-#endif
-
-    result += readable(m_element->accessibleExtraInfo());
-    return result;
+    return readable(m_element->accessibleExtraInfo());
 }
 
 QVariant AccessibleItem::accessibleValue() const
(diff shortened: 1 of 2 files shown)


Eism (approved):

shoogle:
@DmitryArefiev, you can check screen reader descriptions are working by navigating to toolbar buttons. The name and description of those is the same as the tooltip.

image

**Accessible name:** _"Tie"_
**Accessible description:** _"Add tied note"_

Within the score, information about the selected element is given as the name while contextual information about nearby elements is given as the description.

image

**Accessible name:** _"Note C5 quarter, bar 2 beat 1."_
**Accessible description:** _"Tempo: Andante; dynamic: p; start of hairpin crescendo."_

DmitryArefiev:
@shoogle The first case is working in master as well.


The second case working fine in your PR.  But it still doesn't work on **Linux**
This case I mean:


> image
> 
> **Accessible name:** _"Note C5 quarter, bar 2 beat 1."_ **Accessible description:** _"Tempo: Andante; dynamic: p; start of hairpin crescendo."_

DmitryArefiev:
> @DmitryArefiev, both cases are working for me on Linux (Ubuntu 20.04). Here is the second one: **(VIDEO HAS SOUND 🔈)**
> 
>  simplescreenrecorder-2022-09-08_01.40.49.mp4 
> The description of the hairpin is read when the starting note is selected.
> 
> Interestingly, there is no speech for the hairpin when the hairpin itself is selected, but that's also true in master and it appears to be a separate bug to what I was aiming to fix in this PR.

@shoogle I rechecked again and now it works properly.. Sorry! Probably tested a wrong build.

PR #13134, opened 2022-09-02
Fix #11645: Draw bar lines and clefs in cutaway situations

### LINKED ISSUE #11645: [MU4 Issue] Regression: Missing starting barline in measure following invisible measure.
Making a measure invisible (rightclick > Measure Properties > untick Visible) causes the next measure's starting barline to disappear, which is not the case in MuseScore 3. I'm unable to re-add the missing barline.

image

mike-spa:
Resolves: #11645 

As far as I could tell, everything was still there, we were just not drawing it, so this should fix it. Meanwhile, I've taken the chance to set the length of the partial staff lines for cutaway clefs to a more appropriate value.
Before:
image
Bug fix:
image
Improved staff lines length on cutaway clef:
image

Also added a vtest to protect from regressing again on this.


src/engraving/libmscore/segment.cpp
```diff
@@ -1294,9 +1294,18 @@ void Segment::scanElements(void* data, void (* func)(void*, EngravingItem*), boo
...
-            continue;
+        bool thisMeasureVisible = measure()->visible(staffIdx) && score()->staff(staffIdx)->show();
+        if (!all && !thisMeasureVisible) {
+            Measure* nextMeasure = measure()->nextMeasure();
+            bool nextMeasureVisible = nextMeasure
+                                      && nextMeasure->system() == measure()->system()
+                                      && nextMeasure->visible(staffIdx)
+                                      && score()->staff(staffIdx)->show();
+            if (!(isEndBarLineType() && (nextMeasureVisible || measure()->isCutawayClef(staffIdx))
+                  || isClefType() && measure()->isCutawayClef(staffIdx))) {
```
cbjeukendrup, line 1305: Always a bit dangerous to use `||` and `&&` "at the same level". Many compilers even warn about this. Better add some parentheses, to remove any ambiguity.

mike-spa: 👍

src/engraving/libmscore/measure.cpp
```diff
@@ -234,7 +234,10 @@ void Measure::layoutStaffLines()
     for (MStaff* ms : m_mstaves) {
         if (isCutawayClef(staffIdx) && (score()->staff(staffIdx)->cutaway() || !visible(staffIdx))) {
             // draw short staff lines for a courtesy clef on a hidden measure
-            ms->lines()->layoutPartialWidth(width(), 4.0, true);
+            Segment* clefSeg = findSegmentR(SegmentType::Clef, ticks());
+            double partialWidth = clefSeg ? clefSeg->width() + clefSeg->minLeft() + score()->styleMM(Sid::clefLeftMargin)
+                                  + score()->styleMM(Sid::barWidth) : 0.0;
```
cbjeukendrup, line 239: This seems to assume that the barline is always a single barline. If that's not the case, the lines become too short:
Scherm­afbeelding 2022-09-02 om 16 31 15
It would be ideal if we could use the actual width of the actual barline. Has that already been calculated at this point?

mike-spa: Nah, that was just a stupid way of calculating it, I've now pushed a new one that should be correct independently of the barline. Good catch

oktophonie (approved):

cbjeukendrup (approved):

mike-spa:
@cbjeukendrup I've rebased it now, you can go ahead and merge it if you're ok with it

PR #13138, opened 2022-09-02
Fix #9429: Removed repeat types in barline properties

### LINKED ISSUE #9429: [Inspector] In the 'Barlines' section, the 'Left (start) repeat' option behaves incorrectly

HemantAntony:
Resolves: #9429

### DIFF
--- a/src/inspector/models/notation/barlines/barlinesettingsmodel.cpp
+++ b/src/inspector/models/notation/barlines/barlinesettingsmodel.cpp
@@ -83,6 +83,13 @@ void BarlineSettingsModel::resetProperties()
     m_hasToShowTips->resetToDefault();
 }
 
+void BarlineSettingsModel::updatePropertiesOnNotationChanged()
+{
+    loadPropertyItem(m_type, [](const QVariant& elementPropertyValue) -> QVariant {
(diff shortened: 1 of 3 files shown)


cbjeukendrup:
If you now select an existing repeat barline, you get this:
Scherm­afbeelding 2022-09-02 om 20 11 33
which is basically unavoidable if we remove the repeat barline options from the dropdown.

Personally I'm not convinced that this is a good idea, especially since a PR with a real solution already exists.

MarcSabatella:
If we're confident about the real solution, great - I only proposed removing the setting when it was suggested that the solution might be incomplete or risky.

Tantacrul:
We need to go with this solution for 4.0 (on recommendation from @RomanPudashkin) simply because the alternative solution will take more time to fully work through and merge.

Right now, I'm OK with the slightly awkward problem of the blank dropdown because it is better than the existing situation.

RomanPudashkin (approved):

PR #13146, opened 2022-09-03
[MU4] Disable writing .mxml from the GUI

Jojo-Schmitz:
still possible via command line

Fixes an unintended side effect of #13122


Jojo-Schmitz:
As far as I can tell mxml writing came in via 025baf6, not sure whether that was intentional though, might well have been a typo (@igorkorsukov?) , corrected later by adding back mxl in bf328ec (@Eism?)

Jojo-Schmitz:
May interfere with 76997dc

shoogle (changes requested):
MuseScore 3 couldn't export `.mxml` via the command line so I'm going to prevent this being merged until we find out why it was added in MuseScore 4. If it was an accident then most of https://github.com/musescore/MuseScore/pull/13122 needs to be reverted.

Jojo-Schmitz:
I believe my 'forensics' above explain when and where this happend. As for the why only @Eism can tell, I guess.

I'd have a PR for reverting this all ready within minutes (have it ready locally, just need to push it)

Jojo-Schmitz:
Actually see #13151

Jojo-Schmitz:
See also #13175

shoogle:
Closing in favour of #13175 and #13151

PR #13149, opened 2022-09-03
Require authorization before saving to cloud

cbjeukendrup:
I created a RequireAuthorizationDialog, which can be reused in other placed where the user needs to be logged in to perform a certain action.


src/project/internal/saveprojectscenario.cpp
```diff
@@ -171,6 +171,14 @@ RetVal<SaveLocation::CloudInfo> SaveProjectScenario::askCloudLocation(INotationP
 RetVal<SaveLocation::CloudInfo> SaveProjectScenario::doAskCloudLocation(INotationProjectPtr project, bool canSaveLocallyInstead,
                                                                         CloudProjectVisibility defaultVisibility) const
 {
+    if (!authorizationService()->userAuthorized().val) {
+        Ret ret = authorizationService()->requireAuthorization(
+            trc("project/save", "Login or create a free account on musescore.com to save this score to the cloud."));
```
RomanPudashkin, line 176: Looks like this text can be used directly in  RequireAuthorizationDialog.qml without having to pass it as an argument

RomanPudashkin (approved):

src/project/internal/saveprojectscenario.cpp
cbjeukendrup: The reason to pass it from here, is that this text explicitly mentions saving a score to the cloud, but in the future, we may want to reuse this dialog but with a different text.

RomanPudashkin: Yes, makes sense, I didn't read the whole string

PR #13151, opened 2022-09-03
[MU4] Remove read/write .mxml

Jojo-Schmitz:
This does revert #13122 and replaces (Read: is an alternative to) #13146

It additionally eliminates `mxml` format writing entirely and also condenses MacOSXBundleInfo.plist.in (as fas as MusicXML is concerned)

### DIFF
--- a/build/Linux+BSD/musescore.xml.in
+++ b/build/Linux+BSD/musescore.xml.in
@@ -103,8 +103,7 @@
     <mime-type type="application/vnd.recordare.musicxml">
         <!-- See "Container file" at http://www.musicxml.com/for-developers/musicxml-dtd/ -->
         <comment>MusicXML compressed score</comment>
-        <glob pattern="*.mxl"/><!-- preferred extension must go first -->
-        <glob pattern="*.mxml" weight="40"/><!-- reduced weight (default 50) to avoid matching Macromedia XML files -->
+        <glob pattern="*.mxl"/>
         <sub-class-of type="application/zip"/>
         <icon name="application-vnd.recordare.musicxml@MSCORE_INSTALL_SUFFIX@"/>
     </mime-type>
--- a/build/MacOSXBundleInfo.plist.in
+++ b/build/MacOSXBundleInfo.plist.in
@@ -236,7 +236,6 @@
 			<key>CFBundleTypeExtensions</key>
 			<array>
 				<string>mxl</string>
-				<string>mxml</string>
 			</array>
 			<key>CFBundleTypeMIMETypes</key>
 			<array>
--- a/build/packaging/WIX.template.in
+++ b/build/packaging/WIX.template.in
@@ -100,7 +100,6 @@
             <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".xml" Value="" Type="string" />
             <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".musicxml" Value="" Type="string" />
             <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".mxl" Value="" Type="string" />
-            <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".mxml" Value="" Type="string" />
             <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".cap" Value="" Type="string" />
             <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".capx" Value="" Type="string" />
             <RegistryValue Root="$(var.RegistryRoot)" Key="SOFTWARE\Classes\Applications\$(var.ExeKey)\SupportedTypes" Name=".scw" Value="" Type="string" />
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -3080,7 +3080,7 @@ void MusicXMLParserDirection::dashes(const QString& type, const int number,
         b->setLineStyle(LineType::DASHED);
         // TODO brackets and dashes now share the same storage
         // because they both use ElementType::TEXTLINE
-        // use mxml specific type instead
+        // use MusicXML specific type instead
         starts.append(MusicXmlSpannerDesc(b, ElementType::TEXTLINE, number));
     } else if (type == "stop") {
         auto b = spdesc._isStarted ? toTextLine(spdesc._sp) : Factory::createTextLine(_score->dummy());
--- a/src/importexport/musicxml/internal/musicxmlreader.cpp
+++ b/src/importexport/musicxml/internal/musicxmlreader.cpp
@@ -38,7 +38,7 @@ mu::Ret MusicXmlReader::read(MasterScore* score, const io::path_t& path, const O
     std::string suffix = mu::io::suffix(path);
     if (suffix == "xml" || suffix == "musicxml") {
         err = importMusicXml(score, path.toQString());
-    } else if (suffix == "mxl" || suffix == "mxml") {
+    } else if (suffix == "mxl") {
         err = importCompressedMusicXml(score, path.toQString());
     }
     return make_ret(err, path);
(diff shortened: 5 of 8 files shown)


src/converter/internal/compat/backendapi.cpp
```diff
@@ -50,11 +50,11 @@ using namespace mu::io;
 
 static const std::string PNG_WRITER_NAME = "png";
 static const std::string SVG_WRITER_NAME = "svg";
-static const std::string SEGMENTS_POSITIONS_WRITER_NAME = "sposXML";
-static const std::string MEASURES_POSITIONS_WRITER_NAME = "mposXML";
+static const std::string SEGMENTS_POSITIONS_WRITER_NAME = "spos";
+static const std::string MEASURES_POSITIONS_WRITER_NAME = "mpos";
```
shoogle, line 54: This change was not in #13122. Please remove it from this PR.

src/notation/notationmodule.cpp
```diff
@@ -113,8 +113,8 @@ void NotationModule::resolveImports()
 
     auto writers = modularity::ioc()->resolve<project::INotationWritersRegister>(moduleName());
     if (writers) {
-        writers->reg({ "sposXML" }, std::make_shared<PositionsWriter>(PositionsWriter::ElementType::SEGMENT));
-        writers->reg({ "mposXML" }, std::make_shared<PositionsWriter>(PositionsWriter::ElementType::MEASURE));
+        writers->reg({ "spos" }, std::make_shared<PositionsWriter>(PositionsWriter::ElementType::SEGMENT));
+        writers->reg({ "mpos" }, std::make_shared<PositionsWriter>(PositionsWriter::ElementType::MEASURE));
```
shoogle, line 117: Same.

shoogle (changes requested):
This PR should do nothing except:

1. Revert all of #13122 (except some condensing in MacOSXBundleInfo.plist.in).
2. Remove `.mxml` as an export option on the command line.

Adding any other changes to this PR is just confusing.

src/converter/internal/compat/backendapi.cpp
Jojo-Schmitz: done, will PR separately

src/notation/notationmodule.cpp
Jojo-Schmitz: done, will PR separately

Jojo-Schmitz:
Stiull @Eism would need to review this, esp. the chages that may affect the backend

src/converter/internal/compat/backendapi.cpp
```diff
@@ -54,7 +54,7 @@ static const std::string SEGMENTS_POSITIONS_WRITER_NAME = "sposXML";
 static const std::string MEASURES_POSITIONS_WRITER_NAME = "mposXML";
 static const std::string PDF_WRITER_NAME = "pdf";
 static const std::string MIDI_WRITER_NAME = "midi";
-static const std::string MUSICXML_WRITER_NAME = "mxml";
+static const std::string MUSICXML_WRITER_NAME = "mxl";
```
shoogle, line 57: I suspect that this is supposed to be `mxml`. But keep it in the PR for now.

It looks like the string `mxml` used in the website API (it appears in the 3.x code [here](https://github.com/musescore/MuseScore/blob/2513676e512d29d554cb6c4d37d3efaf53ea2c5b/mscore/file.cpp#L3569)), and this probably caused the original confusion that saw [`.mxml` added as an export format](https://github.com/musescore/MuseScore/pull/8344/files#diff-b3721830027f45451d2b81cff70b29584ca58ec71e95b07cd2effe26d8f33f5bR74). I could be wrong though so let's wait for confirmation.

Jojo-Schmitz: It also is uesd as the extension apparently, that at least is my understanding.

Jojo-Schmitz: Actually delving deeper into the code I'm now pretty sure that this `mxml` is **also** used as the extension, not **just** as a key in the metadata.json

Jojo-Schmitz: So we may also want to change `midi` to `mid`, to keep the downloadable formats the same as with MU3 (and a 3 char extension maximises compatibility)

Jojo-Schmitz:
See also #13175 (which probably will cause merge conflicts)

src/converter/internal/compat/backendapi.cpp
shoogle: `mxml` is fixed in #13175.

I kept `midi` as it is. It seems that `.midi` is widely recognised as a MIDI file even if `.mid` is the normal extension. Both MU3 and MU4 allow the creation of MIDI files with `.mid`, `.midi` and even `.kar` as the extension. This is achieved in the GUI by manually changing the file extension in the operating system's Export dialog that appears after our Export dialog.

image

We don't want dedicated options for `.midi` and `.kar` because:

1. `.mid` is preferred over `.midi`.
2. `.kar` is not quite the same as a MIDI file. As I understand, the Karaoke format is a subset of MIDI files where only the top staff has lyrics. To properly support this we would need to:
    1. Actually export lyrics to MIDI files (currently we don't export any lyrics to MIDI file).
    2. Prevent lyrics being exported to all except the top staff when `.kar` is selected.

shoogle:
#13175 is merged now so writing `.mxml` is no longer possible. Please rebase this so that it just prevents reading `.mxml`.

src/converter/internal/compat/backendapi.cpp
Jojo-Schmitz: That's not the point: Up till now MuseScore.com offeres a MIDI download as ".mid". and also the MuseScore program exports as ".mid" only.
Now, at least on musescore.com, that apparently changes to ".midi", and that does seem inconsistent.

Jojo-Schmitz:
Rebased, and included another commit dealing with .mid vs .midi

src/converter/internal/compat/backendapi.cpp
shoogle: I think you might have misunderstood what the `--score-media` command does. It doesn't write any files to disk. It simply sends this JSON data to `stdout`:
```JSON
{
"pngs":  ["BASE64DATA"],
"svgs":  ["BASE64DATA"],
"sposXML": "BASE64DATA",
"mposXML": "BASE64DATA",
"pdf": "BASE64DATA",
"midi": "BASE64DATA",
"mxml": "BASE64DATA",
"metadata": {"composer":"Composer","duration":4,"fileVersion":302,"hasHarmonies":"false","hasLyrics":"false","keysig":0,"lyrics":"","measures":2,"mscoreVersion":"3.6.2","pageFormat":{"height":297,"twosided":"true","width":210},"pages":1,"parts":[{"harmonyCount":0,"hasDrumStaff":"false","hasPitchedStaff":"true","hasTabStaff":"false","instrumentId":"keyboard.piano","isVisible":"true","lyricCount":0,"name":"Piano","program":0}],"poet":"","previousSource":"","subtitle":"","tempo":0,"tempoText":"","textFramesData":{"composers":["Composer"],"poets":[],"subtitles":[],"titles":["Title"]},"timesig":"4/4","title":"Title"}
}
```
So the output contains the *contents* of a MIDI file (as [Base64](https://en.wikipedia.org/wiki/Base64#Base64_table_from_RFC_4648) data); it doesn't specify a *name* for the MIDI file, therefore the website is free to call the file whatever it likes.

The output (including the names of the `"midi"` and `"mxml"` object keys) has not changed since MuseScore 3, yet the website is already offering those files with the correct `.mid` and `.mxl` extensions. PNG and SVG images are being used correctly on the website despite those key names in this file ending in the letter "s".

Jojo-Schmitz: As far as I can tell it does not only that, but **also** does the corresponding export

Jojo-Schmitz: Check `PNG_WRITER_NAME` and `SVG_WRITER_NAME? :  it is "png" resp. "svg", singular, but in your `score-media` output it is plural. Check `SEGMENTS_POSITIONS_WRITER_NAME` and `MEASURES_POSITIONS_WRITER_NAME` and compare with 
```
        writers->reg({ "sposXML" }, std::make_shared<PositionsWriter>(PositionsWriter::ElementType::SEGMENT));
        writers->reg({ "mposXML" }, std::make_shared<PositionsWriter>(PositionsWriter::ElementType::MEASURE));
```
and how that differes from MuseScore 3

shoogle: It has to *generate* a MIDI file in order to embed it within the JSON as Base64 data. This involves running the MIDI export code, but the file is "exported" into memory, not saved to the hard drive.

Jojo-Schmitz: Ah, I see! So the backend still produces ".mid" files for download?

Jojo-Schmitz: Still my change won't harm here, or would it?

shoogle: > So the backend still produces ".mid" files for download?

Yes.

> Still my change won't harm here, or would it?

It does no harm because you kept the JSON key as `"midi"`, but it's not really necessary or related to this PR.

Jojo-Schmitz: OK, moved to another PR

shoogle (approved):

PR #13175, opened 2022-09-05
Prevent creation of .mxml files

shoogle:
The standard file extension for compressed MusicXML is `.mxl` not `.mxml`.

Prior to this commit we allowed compressed MusicXML files to be created on the command line with either of these commands:

    mscore file.mscz -o file.mxl    # should succeed
    mscore file.mscz -o file.mxml   # should fail but was successful

This caused confusion in PR #13122 which saw MXML added as an export option in the GUI.

image

Now the GUI option is removed and the `-o file.mxml` command will correctly fail with non-zero exit status, thereby preventing the creation of non-standard files.

---

The string "mxml" is also used as a key in the JSON output of this command:

    mscore file.mscz --score-media

This commit avoids changing that behaviour.


Jojo-Schmitz:
See also #13146 and #13151

Eism (approved):

PR #13178, opened 2022-09-05
First system indentation correction (Fix #13166)

### LINKED ISSUE #13166: [MU4 Issue] First system indent is 1 sp too large
**Describe the bug**
The default indent of 5 sp results in a larger indent than it should

**To Reproduce**
1. New score, treble clef template
2. Reduce first system indent to 0.5 sp
3. Postion the score against the edge of the window to more easily gauge the size of the indent
4. See error: it's clearly more than a staff space

**Expected behavior**
Indent should be exactly 0.5 sp

**Screenshots**
This GIF shows me changing the indent between 0.0 sp and 0.5 sp; you can see it's clearly way more than that

indent

**Platform information**
 - OS: Linux

**Additional context**
It's the same for any non-zero value: always about 1 sp too big.  Another way to check is to add a horizontal frame on another system of the same width, and compare.  The default 5 sp intent is clearly about 1 sp wider than a 5 sp horizontal frame.

mike-spa:
Resolves: #13166

### DIFF
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -376,22 +376,24 @@ void System::layoutSystem(const LayoutContext& ctx, double xo1, const bool isFir
     //---------------------------------------------------
     double maxNamesWidth = instrumentNamesWidth();
 
+    double indent = maxNamesWidth > 0 ? maxNamesWidth + instrumentNameOffset : 0.0;
     if (isFirstSystem && firstSystemIndent) {
-        maxNamesWidth = std::max(maxNamesWidth, styleP(Sid::firstSystemIndentationValue) * mag());
+        indent = std::max(indent, styleP(Sid::firstSystemIndentationValue) * mag());
+        maxNamesWidth = indent - instrumentNameOffset;
     }
 
     double maxBracketsWidth = totalBracketOffset(ctx);
     double bracketsWidth = layoutBrackets(ctx);
     double bracketWidthDifference = maxBracketsWidth - bracketsWidth;
 
-    if (RealIsNull(maxNamesWidth)) {
+    if (RealIsNull(indent)) {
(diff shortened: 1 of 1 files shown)


cbjeukendrup:
I found another edge case: if there _are_ instrument names, but they are very short, the first system indent ends up having the value of `Sid::firstSystemIndentationValue` + `Sid::instrumentNameOffset`. So basically, I think we should take `std::max` over `namesWidth + instrumentNameOffset` and `firstSystemIndentationValue`. However, for aligning the instrument names at the bottom of the method, we need the total indent (excluding brackets) minus `instrumentNameOffset`. If `firstSystemIndentationValue > namesWidth + instrumentNameOffset`, then that value will not be equal to `namesWidth`. I think the algorithm could look like this:

```c++
double namesWidth = instrumentNamesWidth();

double indent = namesWidth;
if (!RealIsNull(namesWidth)) { // If there are instrument names
    indent += instrumentNameOffset // take offset into account
} 

if (isFirstSystem && firstSystemIndent) { // Apply minimum indent for first system
    indent = std::max(indent, styleP(Sid::firstSystemIndentationValue) * mag());

    // Adjust namesWidth to take first system indentation into account
    // because it will be used for aligning the instrument names later on
    namesWidth = indent - instrumentNameOffset;
}

double bracketsWidth = layoutBrackets(ctx);

// And then use indent like this:
if (RealIsNull(indent)) {
    double maxBracketsWidth = totalBracketOffset(ctx);
    
    if (score()->styleB(Sid::alignSystemToMargin)) {
        _leftMargin = maxBracketsWidth - bracketsWidth;
    } else {
        _leftMargin = maxBracketsWidth;
    }
} else {
    _leftMargin = indent + bracketsWidth;
}

mike-spa:
Now should be right

oktophonie (approved):

RomanPudashkin (approved):

Issue #13159, opened 2022-09-04
[MU4 Issue] Crash after adding a note, saving, then changing its pitch (with Screen Reader enabled)


MarcSabatella:
**Describe the bug**
Current MuseScore 4 builds crash for me reliably if I add a note, save, then try to change the pitch of the note just entered

**To Reproduce**
1. Open any score - or create a new one with treble clef template
2. Select first measure
3. Add a single note quarter note C5 ("N 5 C")
4. Press Esc to leave note input mode
5. Save
6. Press Down to move the pitch of the note down

** Result **
Crash

**Expected behavior**
Note should simply change pitch, no crash

**Screenshots**
![save crash](https://user-images.githubusercontent.com/1799936/188321674-02b7d22f-2585-4515-81d9-c10efe263574.gif)

**Platform information**
 - OS: Linux, Windows

**Additional context**
It was quite the process to track this down, but in the end, it's exactly this simple - on both my Linux and Windows systems.

Note, actually entering (or at least changing pitch of) a note before the save is important.  Merely editing text or adding other markings and then saving won't trigger a crash on the next pitch change.  Has to be an operation involving pitch - either adding one or changing pitch of an existing one.

Also, it's not just up/down that triggers the crash - anything that changes pitch, like adding an accidental, or dragging the note up or down.  But other operations work fine after saving - you can add notes, change duration of existing notes or move them between voices, add articulations, etc.  And these operations often temporarily fix the problem - changing pitch after doing one of these other "safe" operations causes the pitch change to work.  Until the next save, then a pitch change will again crash.

MarcSabatella:
I am now completely convinced autosave is involved.  If I start MuseScore from the command line (AppImage), the console log shows me that the autosave *does* occur after only a minute.  At least most of the time - sometimes it takes longer.  make it's based on the last autosave operation, not the last change?

Anyhow, for 100% reproducible results, then, wait for the autosave message to appear on the console.

wizofaus:
I'm not convinced at all, I sat there adjust a note pitch with the arrow key continually right through several auto-saves (during which it noticeably pauses) and not a single crash.  But I've had quite a number of crashes doing things unrelated to changing note pitch, including just selecting "edit element" from the right-click content menu.

MarcSabatella:
[...]

Issue #12821, opened 2022-08-12
[MU4 Issue] Crash on moving notes up/down octave using keyboard (sometimes)


emielbeinema:
**Describe the bug**
In note input mode, when using the keyboard to enter notes, Musescore often crashes when moving notes up/down an octave (using `Ctrl + Arrow up/down`). This does not happen on all occasions, but in bars where it does happen, it appears to happen consistently.

Musescore also crashes the same note is moved up an octave outside note input mode.

<details><summary>Console output</summary>

`13:10:24.314 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: note-f`
`13:10:24.937 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: note-f`
`13:10:25.250 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: pad-note-4`
`13:10:25.551 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: note-f`
`13:10:25.809 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: pad-note-2`
`13:10:26.135 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: note-f`
`13:10:27.433 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: note-f`
`13:10:28.431 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: pad-note-8`
`13:10:28.969 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: note-e`
```13:10:29.412 | INFO  | main_thread | ActionsDispatcher | dispatch: try call action: pitch-up-octave`
[843539:843539:20220812,131029.446273:WARNING thread_snapshot_linux.cc:111] Unknown scheduling policy 1073741826
[843539:843539:20220812,131029.457730:ERROR elf_dynamic_array_reader.h:61] tag not found
[843539:843539:20220812,131029.461069:ERROR elf_dynamic_array_reader.h:61] tag not found
[843539:843539:20220812,131029.465471:ERROR elf_dynamic_array_reader.h:61] tag not found
[843539:843539:20220812,131029.465678:ERROR elf_dynamic_array_reader.h:61] tag not found
[843539:843539:20220812,131029.465844:ERROR elf_dynamic_array_reader.h:61] tag not found
[843539:843539:20220812,131029.468893:ERROR elf_dynamic_array_reader.h:61] tag not found
[...]

Issue #12835, opened 2022-08-12
[MU4 Issue] - Arrow keys cause crash


drbirkenstocks:
**Describe the bug**
When attempting to adjust text element within a frame using arrow keys, MU4 crashes. Not sure if this happens with other text elements.

**To Reproduce**
1. Select Lyricist text
2. Use keyboard arrows to move incrementally
3. Program crashes

**Expected behavior**
Text will move incrementally

**Platform information**
 - Windows 11

Grillzombie:
No crash on 
OS: Linux Mint 20.3, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2844416344, revision: github-musescore-musescore-ae252cd

MarcSabatella:
Crashes for me on Debian, attempting to move any text element within a frame by clicking then press a cursor key.

charlesneimog:
Crash on Windows 11 too (not all times)

MarcSabatella:
This crash is no longer 100% reproducible for me.  But, a possibly related crash that is also not totally reproducible: some percentage of the time, up or down arrows will also crash with a *note* selected rather than a *frame text*.  I haven't found specific steps to reproduce, so I'm not opening a new issue yet.  But it happens often enough to be a concern, and I guess it's *possible* the crashes will turn out to be related, as the command to handle the Up and Down keys is - or at least was in MU3 - the same.  So maybe the crash happens in the part of the code responsible for figuring out if we're dealing with text or a note?

Tantacrul:
Again, I'm wondering if this is affecting linux worse than mac / pc because I've not managed to reproduce it (yet)

MarcSabatella:
I see the crash probably 2-3 times a day when using the up or down keys to transpose a note on Linux, but unfortunately, I can't say what's different about the cases where it happens and the cases where it doesn't.  My sense is that it seems to happens most if it's the very first thing I do upon returning to MuseScore after switching to work on something else for a while.  So, I'll be using MuseScore with no problem, go to type up an email or whatever, then flip back MuseScore, decide to change a note, and it crashes.  That could be a complete red herring, or it *could* be, something to do with loading of soundfonts, and the issue might actually be happening just on clicking the note but I don't see the crash until hitting the cursor key.

But, it just happened on Linux while typing this message, in pretty much exactly that scenario - very thing thing I did upon flipping back to MU4 half way through the above paragraph.  Using last night's build.

[...]

PR #13186, opened 2022-09-05
[MU4] Crash after adding a note, saving, then changing its pitch

Eism:
Resolves: #13159

### DIFF
--- a/src/engraving/libmscore/engravingitem.cpp
+++ b/src/engraving/libmscore/engravingitem.cpp
@@ -111,6 +111,7 @@ EngravingItem::EngravingItem(const EngravingItem& e)
 
     //! TODO Please don't remove (igor.korsukov@gmail.com)
     //m_accessible = e.m_accessible->clone(this);
+    m_accessibleEnabled = e.m_accessibleEnabled;
 }
 
 EngravingItem::~EngravingItem()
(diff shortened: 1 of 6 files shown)


DmitryArefiev:
Tested #13159 on Win10, LinuxMint 20.2 - **FIXED**

src/engraving/libmscore/engravingitem.cpp
```diff
@@ -176,6 +177,18 @@ AccessibleItemPtr EngravingItem::createAccessible()
...
 
+void EngravingItem::notifyAboutNameChanged()
+{
+    if (!selected()) {
+        return;
+    }
+
+    if (m_accessible) {
+        doInitAccessible();
+        m_accessible->accessibleRoot()->notifyAboutFocuedElemntNameChanged();
```
MarcSabatella, line 188: "notifyAboutFocuedElemntNameChanged"?  Is that the name of one of our functions?  Two typos for the price of one :-)

MarcSabatella:
Justed tested the current artifact (the one that passed the tests!) and am happy to see that it too seems to work well.

src/engraving/libmscore/engravingitem.cpp
Eism: Ah, I was rushing

cbjeukendrup:
Does this maybe also resolve #12821? It looks like that could be the same issue as #13159.

MarcSabatella:
These are almost certainly all the same issue, except the text portion of #12835 which isn't reproducible anymore anyhow.

MatheusWittmann:
I've just downloaded the last version available, and I can reproduce exactly this error :(

MarcSabatella:
Right, this PR isn’t merged yet.

RomanPudashkin (approved):

Issue #13172, opened 2022-09-05
[MU4 Issue] Add - Intervals/Notes - Below doesn't work (Win/Linux)


sammik:
**Describe the bug**
Add - Intervals - Below doesn't work.
Add - Notes - Below doesn't work.

![Capture2](https://user-images.githubusercontent.com/90187801/188477583-cd3417ac-294e-4ab9-a3ab-ae7fa957ba50.PNG)

![Capture](https://user-images.githubusercontent.com/90187801/188477611-7441eaf2-b514-4436-8080-91a24b49a4a1.PNG)

**To Reproduce**
1. Select note(s)
2. Click menu Add - Intervals/Notes - Something below
3. Nothing happens

**Expected behavior**
Interval(s) below note(s) are added.

**Platform information**
 - OS: [reported on Linux]

DmitryArefiev:
The same with Add - Notes. 
Added additional info to the description.

Thanks for logging!

DmitryArefiev:
Related to #11947

wizofaus:
Should be fixed by https://github.com/musescore/MuseScore/pull/13189

DmitryArefiev:
Fixed in #13189

PR #13189, opened 2022-09-06
fix #11947 - submenu commands don't always activate using mouse

### LINKED ISSUE #11947: [MU4 Issue] Many commands in Add / Text do nothing (Win/Linux)
**Describe the bug**
If you try to add elements from the Add / Text menu, nothing happens for several of the commands.

**To Reproduce**
1. select a note in a score
2. Add / Text / Sticking
3. See error - nothing happens

**Expected behavior**
Normally, you should get an empty element in edit mode, with a visible cursor, ready for you to type.  This works as expected for Title, Staff Text, and a few other element types, but not for others - for others, nothing at all happens.

Among the commands that don't work are Sticking, Roman numeral analysis, and Nashville number, which is especially unfortunate since there is no *other* way to add these right now.  Most others have shortcuts that do work, or are also present in palettes.  Also, for Instrument change, you *can* add it from the palette, but there is a bug (present in MU3 as well) where the key signature may be added incorrectly.  In MU3, adding from Add / Text has been the recommended workaround.

**Platform information**
 - OS: Linux

**Additional context**

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/11947)

Proper handling of opened area for cascading sub-menus etc.

### DIFF
--- a/src/appshell/qml/platform/AppMenuBar.qml
+++ b/src/appshell/qml/platform/AppMenuBar.qml
@@ -39,12 +39,20 @@ ListView {
 
     model: appMenuModel
 
+    function openedArea(menuLoader) {
+        if (menuLoader.isMenuOpened) {
+            if (menuLoader.menu.subMenuLoader && menuLoader.menu.subMenuLoader.isMenuOpened)
+                return openedArea(menuLoader.menu.subMenuLoader)
+            return Qt.rect(menuLoader.menu.x, menuLoader.menu.y, menuLoader.menu.width, menuLoader.menu.height)
+        }
+        return Qt.rect(0, 0, 0, 0)
+    } 
+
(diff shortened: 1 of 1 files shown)


HemantAntony:
I wonder if this fix would also solve #13172

wizofaus:
Looks like it does, yep.

HemantAntony:
I think you can link it by adding `Resolves: #13172` after the first `Resolves` in the PR description

RomanPudashkin (approved):

DmitryArefiev:
Tested #11947 #13172 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

@wizofaus Thanks for fixing!

PR #13192, opened 2022-09-06
[MU4] Fix MSVC compiler warning

Jojo-Schmitz:
reg.  conversion from 'size_t' to 'int', possible loss of data  (C4267)

### DIFF
--- a/src/engraving/libmscore/stafftype.cpp
+++ b/src/engraving/libmscore/stafftype.cpp
@@ -843,7 +843,7 @@ String StaffType::tabBassStringPrefix(int strg, bool* hasFret) const
 //    rect    the rect of the 'blue rectangle' showing the input position
 //---------------------------------------------------------
 
-void StaffType::drawInputStringMarks(mu::draw::Painter* p, int string, int voice, const RectF& rect) const
+void StaffType::drawInputStringMarks(mu::draw::Painter* p, int string, voice_idx_t voice, const RectF& rect) const
 {
     if (_group != StaffGroup::TAB) {
         return;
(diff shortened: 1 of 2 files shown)


src/notation/view/noteinputcursor.cpp
```diff
@@ -48,7 +48,7 @@ void NoteInputCursor::paint(mu::draw::Painter* painter)
         const StaffType* staffType = state.staff ? state.staff->staffType() : nullptr;
 
         if (staffType) {
-            staffType->drawInputStringMarks(painter, state.currentString, state.currentVoiceIndex, cursorRect);
+            staffType->drawInputStringMarks(painter, state.currentString, static_cast<int>(state.currentVoiceIndex), cursorRect);
```
cbjeukendrup, line 51: I'd rather update the signature of `StaffType::drawInputStringMarks` to take `voice_idx_t` instead of `int`.

Jojo-Schmitz: True, and in this case esp. easy, as it is used in only one place

Jojo-Schmitz: Done

src/engraving/libmscore/stafftype.cpp
```diff
@@ -843,7 +843,7 @@ String StaffType::tabBassStringPrefix(int strg, bool* hasFret) const
 //    rect    the rect of the 'blue rectangle' showing the input position
 //---------------------------------------------------------
 
-void StaffType::drawInputStringMarks(mu::draw::Painter* p, int string, int voice, const RectF& rect) const
+void StaffType::drawInputStringMarks(mu::draw::Painter* p, int string, voice_idx_t voice, const RectF& rect) const
```
Jojo-Schmitz, line 846: Do we want/need `mu::engraving::voice_idx_t` here?

RomanPudashkin (approved):

PR #13204, opened 2022-09-06
Spacing should not kern over chords with articulations or chords which start/end slurs

mike-spa:
This PR is based on #13218, so that should be merged first. I've split the two things to collect vtests separately.

Before:

kerningBefore

After:

kerningAfter


cbjeukendrup:
Another rebase is needed...

mike-spa:
@cbjeukendrup done, thank you! 👍

cbjeukendrup:
Maybe I've found a bug:
1. Create a situation with chords and accidentals where kerning would occur
2. Add a slur -> kerning disabled
3. Flip slur -> kerning still disabled, but I would expect it to be enabled again.
Scherm­afbeelding 2022-09-19 om 11 30 02

mike-spa:
Ha! Great catch. That's because in #13218  I hadn't extracted the full direction logic, but only the "auto" portion of it. I've corrected that now

cbjeukendrup:
I found it's still not 100% "waterproof" when multiple slurs are used. After some fiddling with adding/removing/flipping slurs, I managed to create this situation (it's somewhat artificial, but I think it could happen with non-artificial multi-slur situations too):
Scherm­afbeelding 2022-09-19 om 12 56 43
It must be that the `_startSlur` of the first chord becomes the downwards slur, and the big upwards slur gets overlooked.

mike-spa:
> It must be that the `_startSlur` of the first chord becomes the downwards slur, and the big upwards slur gets overlooked.

Yeah, the chord is storing only one `_startSlur` and one `_endSlur`. To make it general, I would have to make a vector of startSlurs and a vector of endSlurs and I wasn't sure it would be worth it, that's why I didn't (and I actually expected this to possibly happen). I could try though, maybe the performance hit isn't significant

src/engraving/libmscore/slur.cpp
```diff
@@ -1741,56 +1723,75 @@ SpannerSegment* Slur::layoutSystem(System* system)
...
+            c2->computeUp();
+        }
 
-    _up = !(startCR()->up());
+        if (slurSegment && _sourceStemArrangement != -1 && _sourceStemArrangement != calcStemArrangement(c1, c2)) {
+            // copy & paste from incompatible stem arrangement, so reset bezier points
+            for (int g = 0; g < (int)Grip::GRIPS; ++g) {
+                slurSegment->ups((Grip)g) = UP();
+            }
+        }
```
cbjeukendrup, line 1758: I'm doubtful about having this in `computeUp()`. It looks like we should be able to do `computeUp()` once for the whole slur ("once" as opposed to "for every segment"), and move this segment-dependent logic back to `layoutSystem`. Or am I missing something?

vpereverzev: `I think it's possible to move that segment-related part out of this method though`

do it

cbjeukendrup:
> > It must be that the `_startSlur` of the first chord becomes the downwards slur, and the big upwards slur gets overlooked.
> 
> Yeah, the chord is storing only one `_startSlur` and one `_endSlur`. To make it general, I would have to make a vector of startSlurs and a vector of endSlurs and I wasn't sure it would be worth it, that's why I didn't (and I actually expected this to possibly happen). I could try though, maybe the performance hit isn't significant

I was thinking about another alternative solution: instead of vectors of slurs, we could store: `bool _startsOrEndsUpSlurs` and `bool _startsOrEndsDownSlurs`. That probably requires that we call `chord->computeIsSlurStartEnd` after `LayoutMeasure::preComputeSlurUp` has been done. That's a bit more dangerous because we need to ensure that `preComputeSlurUp` and `computeIsSlurStartEnd` are done in the correct order and both are done before computing kerning, but I can imagine the performance will be better. Would that be possible? 
(Also something I'd like to hear @vpereverzev's feedback on first 🙂)

For the rest, it seems unnecessary to me to have different vectors for start and end slurs; we can just throw them all together in one vector.

src/engraving/libmscore/slur.cpp
mike-spa: done!

cbjeukendrup:
Yeah, the cost of _constructing/filling_ those vectors may well be low, but then you need to _iterate_ over them in `Note::doComputeKerningType` for every note of the chord. By computing the boolean values once for the chord, you can trivially retrieve them in `Note::doComputeKerningType`.

mike-spa:
I've optimized it further and made sure that the computation is performed only once per chord. The measured performance cost (i.e. the time taken to perform `Score::doLayoutRange()`) is now +5.5% on a huge piano concerto, +5.7% on the Start Wars, +3.6% on a huge file but with not so many articulation and slurs.

mike-spa:
By the way, when #13466 is merged, the performance cost of this PR will become much smaller, maybe even negligible. I'll wait for the other to be merged and the rebase and see.

RomanPudashkin:
@mike-spa please rebase it

mike-spa:
> @mike-spa please rebase it

@RomanPudashkin sorry, still working on it to reduce performance cost. Will be probably ready tomorrow.

mike-spa:
@vpereverzev @RomanPudashkin I've eliminated the performance cost of this PR. I've also optimized some small things around it, and now it even seems to be some 1-2% faster than master.
Master:
master
PR:
optimizedPR

If the vtest don't show any problem, this one is good to go for me

mike-spa:
vtests are good!

PR #13216, opened 2022-09-07
fix #13200 - allow using space to remove hyphen

### LINKED ISSUE #13200: [MU4 Issue] Lyrics - unable to delete hyphen; regression from MU3
**Describe the bug**
It is not possible to delete hyphen in Lyrics

**To Reproduce**
1. Enter "lyrics mode" (Ctrl + L)
2. Write some text
3. Create hyphen (-)
4. Try to remove it (Shift+space; space) or some other way

**Expected behavior**
When going to previous syllabe with hyphen, pressing "space" should replace "hyphen" (-) with "next sylabe" ( )

**Screenshots**
MU4 - bug:

https://user-images.githubusercontent.com/1646034/188654897-f8a0c39a-19ff-4601-90f6-2f3cf32300f6.mp4

MU3 - Works:

https://user-images.githubusercontent.com/1646034/188655017-f2b9850c-2aa8-408d-9762-791d77ca6509.mp4

**Platform information**
 - OS: [reported on Linux]

**Additional context**
Regression from MU3

wizofaus:
Resolves: #13200

### DIFF
--- a/src/notation/inotationinteraction.h
+++ b/src/notation/inotationinteraction.h
@@ -240,7 +240,7 @@ public:
     virtual void addLyricsVerse() = 0;
 
     // Text navigation
-    virtual void navigateToLyrics(MoveDirection direction) = 0;
+    virtual void navigateToLyrics(MoveDirection direction, bool moveOnly = false) = 0;
     virtual void navigateToLyricsVerse(MoveDirection direction) = 0;
 
     virtual void navigateToNextSyllable() = 0;
--- a/src/notation/internal/notationuiactions.cpp
+++ b/src/notation/internal/notationuiactions.cpp
@@ -1589,17 +1589,23 @@ const UiActionList NotationUiActions::m_actions = {
              IconCode::Code::TUNING_FORK,
              Checkable::Yes
              ),
+    UiAction("next-word",
+             mu::context::UiCtxNotationFocused,
+             mu::context::CTX_NOTATION_TEXT_EDITING,
+             TranslatableString("action", "Next word"),
+             TranslatableString("action", "Go to next word")
+             ),
     UiAction("next-text-element",
              mu::context::UiCtxNotationFocused,
              mu::context::CTX_NOTATION_TEXT_EDITING,
              TranslatableString("action", "Next text element"),
-             TranslatableString("action", "Go to next syllable")
+             TranslatableString("action", "Go to next text element")
              ),
     UiAction("prev-text-element",
              mu::context::UiCtxNotationFocused,
              mu::context::CTX_NOTATION_TEXT_EDITING,
              TranslatableString("action", "Previous text element"),
-             TranslatableString("action", "Go to previous syllable")
+             TranslatableString("action", "Go to previous text element")
              ),
     UiAction("next-beat-TEXT",
              mu::context::UiCtxNotationFocused,
--- a/src/framework/shortcuts/data/shortcuts-Mac.xml
+++ b/src/framework/shortcuts/data/shortcuts-Mac.xml
@@ -880,8 +880,12 @@
     <seq>Z</seq>
     </SC>
   <SC>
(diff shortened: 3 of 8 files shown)


src/notation/internal/notationuiactions.cpp
```diff
@@ -1589,17 +1589,23 @@ const UiActionList NotationUiActions::m_actions = {
              IconCode::Code::TUNING_FORK,
              Checkable::Yes
              ),
+    UiAction("next-word",
+             mu::context::UiCtxNotationFocused,
+             mu::context::CTX_NOTATION_TEXT_EDITING,
+             TranslatableString("action", "Next word"),
+             TranslatableString("action", "Go to next word")
```
wizofaus, line 1596: new strings here, does the translation team need to know?

src/notation/internal/notationuiactions.cpp
```diff
@@ -1589,17 +1589,23 @@ const UiActionList NotationUiActions::m_actions = {
...
+             ),
     UiAction("next-text-element",
              mu::context::UiCtxNotationFocused,
              mu::context::CTX_NOTATION_TEXT_EDITING,
              TranslatableString("action", "Next text element"),
-             TranslatableString("action", "Go to next syllable")
+             TranslatableString("action", "Go to next text element")
```
wizofaus, line 1602: these were just wrong

cbjeukendrup:
@wizofaus It looks like we have the same problem with Shift+Space / `prev-text-element`: 
```xml
  <SC>
    <key>prev-text-element</key>
    <seq>Shift+Space</seq>
    </SC>
  <SC>
    <key>prev-text-element</key>
    <seq>Left</seq>
  </SC>
```
Looks like we'll also need a `prev-word` action.

Also, I think you'll need to update `shortcuts_AZERTY.xml` too.

src/notation/inotationinteraction.h
```diff
@@ -240,7 +240,7 @@ class INotationInteraction
     virtual void addLyricsVerse() = 0;
 
     // Text navigation
-    virtual void navigateToLyrics(MoveDirection direction) = 0;
+    virtual void navigateToLyrics(MoveDirection direction, bool moveOnly = false) = 0;
```
cbjeukendrup, line 243: Here you have `bool moveOnly = false`, but in `notationinteraction.h` you have `bool moveOnly = true`. That's scary...

There are people who totally ban the use of default argument values in virtual methods for this reason... I don't think we should go that far, but we must be careful.

sammik:
> @wizofaus It looks like we have the same problem with Shift+Space / `prev-text-element`:
> Looks like we'll also need a `prev-word` action.

@cbjeukendrup I don't think so. This (`prev-text-element`) is how it worked in MU3, isn't it? 
(Space should removes hyphen, but Shift + Space shouldn't)

cbjeukendrup:
@sammik Ah, you're right. I had misunderstood the description in the issue: I read "Shift+Space _or_ Space", but you meant "Shift+Space _and then_ Space". Of course, it should be the latter.

@wizofaus, in that case, I think we should maybe replace 
```xml
  <SC>
    <key>prev-text-element</key>
    <seq>Shift+Space</seq>
    </SC>
  <SC>
    <key>prev-text-element</key>
    <seq>Left</seq>
  </SC>
```
with 
```xml
  <SC>
    <key>prev-text-element</key>
    <seq>Shift+Space</seq>
    <seq>Left</seq>
    </SC>
```
Or we remove one of the two instances of `<SC> <key>prev-text-element</key> ...`. 

I also think we should make it consistent in all shortcuts xml files, so shortcuts.xml, shortcuts_Mac.xml and shortcuts_AZERTY.xml should all be the same for these actions.

RomanPudashkin:
Thank you! I cherry-picked your commit:
https://github.com/musescore/MuseScore/pull/14892

PR #13217, opened 2022-09-07
Fix #13213: Save as command deletes initial file

### LINKED ISSUE #13213: [MU4 Issue] Save As command deletes the initial score from disk
**To Reproduce**
1. Create score
2. Save it as '1'
3. Reopen 
4. Using **Save as**, re-save it as '1a' - **score '1' is deleted from disk and both scores are empty in Recent scores** 
 

https://user-images.githubusercontent.com/90187801/188708418-71fb4ff9-92a9-4960-928d-57c50e38c494.mp4

**Platform information**
Windows, Mac, Linux

HemantAntony:
Resolves: #13213

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -593,9 +593,9 @@ mu::Ret NotationProject::makeCurrentFileAsBackup()
     }
 
     io::path_t backupFilePath = filePath + "~";
-    ret = fileSystem()->move(filePath, backupFilePath, true);
+    ret = fileSystem()->copy(filePath, backupFilePath, true);
     if (!ret) {
-        LOGE() << "failed to move from: " << filePath << ", to: " << backupFilePath;
+        LOGE() << "failed to copy: " << filePath << " to: " << backupFilePath;
         return ret;
     }
 


src/project/internal/notationproject.cpp
```diff
@@ -593,9 +593,9 @@ mu::Ret NotationProject::makeCurrentFileAsBackup()
     }
 
     io::path_t backupFilePath = filePath + "~";
-    ret = fileSystem()->move(filePath, backupFilePath, true);
+    ret = fileSystem()->copy(filePath, backupFilePath, true);
     if (!ret) {
-        LOGE() << "failed to move from: " << filePath << ", to: " << backupFilePath;
+        LOGE() << "failed to copy: " << filePath << " to: " << backupFilePath;
         return ret;
```
wizofaus, line 599: while I agree this is better, are you sure it doesn't break regular save, which appears to try to move from "xxxx_saving" to "xxxx", which is likely to fail if xxxx stil exists?

HemantAntony: Oh I didn't think about that. Let me check

RomanPudashkin: >are you sure it doesn't break regular save, which appears to try to move from "xxxx_saving" to "xxxx", which is likely to fail if xxxx still exists?

It shouldn't because we call copy with replace = true. Anyway, I asked @DmitryArefiev  to check both Save and Save as

RomanPudashkin (approved):

src/project/internal/notationproject.cpp
HemantAntony: When I checked it nothing broke. But I thought my solution might not be the correct fix :)

DmitryArefiev:
Tested #13213 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #13230, opened 2022-09-07
Fix grace notes stem shape glitch

mike-spa:
Resolves a glitch with grace notes which can cause slurs to change shape when a relayout is triggered.

https://user-images.githubusercontent.com/93707756/188924009-c424792c-18e9-4d1b-9e14-6b31af16e08e.mp4

### DIFF
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -711,32 +711,10 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
     //-------------------------------------------------------------
 
     for (Segment* s : sl) {
-        for (EngravingItem* e : s->elist()) {
-            if (!e || !e->isChordRest() || !score->staff(e->staffIdx())->show()) {
-                // the beam and its system may still be referenced when selecting all,
-                // even if the staff is invisible. The old system is invalid and does cause problems in #284012
-                if (e && e->isChordRest() && !score->staff(e->staffIdx())->show() && toChordRest(e)->beam()) {
-                    toChordRest(e)->beam()->resetExplicitParent();
-                }
-                continue;
-            }
-            ChordRest* cr = toChordRest(e);
-
-            // layout beam
-            if (LayoutBeams::isTopBeam(cr)) {
-                Beam* b = cr->beam();
-                b->layout();
-            }
-            if (!cr->isChord()) {
-                continue;
-            }
-            for (Chord* grace : toChord(cr)->graceNotes()) {
-                Beam* beam = grace->beam();
-                if (beam && !beam->elements().empty() && beam->elements().front() == grace) {
-                    beam->layout();
-                }
-            }
+        if (!s->isChordRestType()) {
+            continue;
         }
+        LayoutBeams::layoutNonCrossBeams(s);
         // Must recreate the shapes because stem lengths may have been changed!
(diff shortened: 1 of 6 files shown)


oktophonie (approved):

src/engraving/layout/layoutsystem.cpp
```diff
@@ -727,6 +727,14 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
                 Beam* b = cr->beam();
                 b->layout();
             }
+            if (!cr->isChord()) {
+                continue;
+            }
+            for (Chord* grace : toChord(cr)->graceNotes()) {
+                if (grace->beam() && grace->beam()->elements().front() == grace) {
```
RomanPudashkin, line 734: Please check if the list is empty before calling front()

mike-spa: 👍

RomanPudashkin: Please don't use these endless grace->beam() calls. Add a temp variable and reuse it. This will not only simplify the code but also help to avoid possible performance problems

mike-spa: 👍👍

src/engraving/libmscore/chord.cpp
```diff
@@ -4202,4 +4202,13 @@ void GraceNotesGroup::setPos(double x, double y)
         chord->movePos(PointF(x, y));
     }
 }
+
+Shape GraceNotesGroup::shape() const
+{
+    Shape shape;
+    for (Chord* grace : (std::vector<Chord*>) * this) {
```
cbjeukendrup, line 4209: Is the cast to `(std::vector<Chord*>)` really necessary? For me, it also seems to compile fine without.

mike-spa: Huh, first time I tried it didn't work without the cast, I must have done something wrong then. Indeed works fine without

src/engraving/layout/layoutsystem.cpp
```diff
@@ -727,6 +727,15 @@ void LayoutSystem::layoutSystemElements(const LayoutOptions& options, LayoutCont
...
             }
+            if (!cr->isChord()) {
+                continue;
+            }
+            for (Chord* grace : toChord(cr)->graceNotes()) {
+                Beam* beam = grace->beam();
+                if (beam && !beam->elements().empty() && beam->elements().front() == grace) {
+                    beam->layout();
+                }
+            }
```
cbjeukendrup, line 738: I suddenly remember that I've seen very similar code in `LayoutMeasure::getNextMeasure`. Would it make sense to make this one exactly the same as that? For example, shouldn't we use `LayoutBeams::isTopBeam` here?

Or maybe even better: add a static method in LayoutBeams that can be called both here and in `LayoutMeasure::getNextMeasure`.

cbjeukendrup (approved):

src/engraving/layout/layoutsystem.cpp
cbjeukendrup: Oops, didn't entirely mean to approve already... anyway, I think this will be the last comment :)

mike-spa: Done, I've put that into a dedicated method 👍

cbjeukendrup (approved):

PR #13237, opened 2022-09-08
fix #13215: support syllable-paste lyrics paste

### LINKED ISSUE #13215: [MU4 Issue] Regression - Lyrics don't paste word by word anymore
In MS3 I was able to use Ctrl-v repeatedly to paste lyrics one word at a time from the clipboard. This does not work in MS4 anymore and only the first word is pasted... repeatedly.
I tested this in the latest Alpha release

1. Create a score
2. Copy a paragraph of lyrics from a text editor to the clipboard
3. Use Ctrl+L to go into lyric entry mode
4. Use Ctrl+V repeatedly to paste each word associated with subsequent notes

**Expected behavior**
I expect each word in the copied text to paste one by one under subsequent notes in the score like they did in MS3

**Platform information**
 - OS: ArchLinux

**Additional context**
This is something I need for my workflow. I hope it gets added again in the final version.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/13215)

Actually I don't like this solution...but nobody's come up with anything better


src/notation/internal/notationinteraction.cpp
```diff
@@ -5044,8 +5080,9 @@ void NotationInteraction::addMelisma()
     mu::engraving::PlacementV placement = lyrics->placement();
     mu::engraving::PropertyFlags pFlags = lyrics->propertyFlags(mu::engraving::Pid::PLACEMENT);
     Fraction endTick = segment->tick(); // a previous melisma cannot extend beyond this point
-
+    int clipTextPos = m_clipboardTextPos;
     endEditText();
+    m_clipboardTextPos = clipTextPos;
```
wizofaus, line 5085: this ensures explicitly pressing _ while pasting words doesn't lose what position in the clipboard text you're up to.

macxcool: You need to do the same with - for multi-syllabic words

Jojo-Schmitz:
Rebase needed

Eism:
fixed in #17524

PR #13245, opened 2022-09-08
Fix stem slash glitch

mike-spa:
Resolves:

https://user-images.githubusercontent.com/93707756/189122278-7ea80ea7-a436-4a84-a27e-4308ba3e2bf5.mp4

### DIFF
--- a/src/engraving/libmscore/stemslash.cpp
+++ b/src/engraving/libmscore/stemslash.cpp
@@ -116,4 +116,26 @@ void StemSlash::layout()
     _width = score()->styleMM(Sid::stemSlashThickness) * graceNoteMag;
     setbbox(RectF(line.p1(), line.p2()).normalized().adjusted(-_width / 2, -_width / 2, _width, _width));
 }
+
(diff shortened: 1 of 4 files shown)


oktophonie (approved):

src/engraving/libmscore/stemslash.cpp
```diff
@@ -116,4 +116,26 @@ void StemSlash::layout()
...
+    Chord* nextChord = nullptr;
+    if (nextParent->isChord()) {
+        nextChord = toChord(nextParent);
+    } else if (nextParent->isNote()) {
+        nextChord = toChord(nextParent->parentItem());
+    }
+    if (!nextChord) {
+        return KerningType::KERNING;
+    }
+    if (nextChord->beam() && nextChord->beam() == this->chord()->beam()) {
```
cbjeukendrup, line 135: (For the record: it isn't really necessary to check do a null check for `nextChord->beam()` before comparing it with another pointer; you're just comparing the pointers' addresses, not dereferencing their pointees.)

cbjeukendrup (approved):

PR #13253, opened 2022-09-08
[MU4] Fix compiler warnings

Jojo-Schmitz:
* MSVC: reg. conversion from 'size_t' to 'int', possible loss of data (C4267).
* MinGW: reg. comparison of integer expressions of different signedness [-Wsign-compare].
* MSVC: reg. cast truncates constant value (C4310).
* MSVC: reg. "the usage of 'method' requires the compiler to capture 'this' but the current default capture mode does not allow it (C4573)", by disabling it.
* MSVC: reg. "unreferenced function with internal linkage has been removed", by disabling them resp. removing the code.
* MSCV/MinGW: reg. unreferenced formal parameter (C4100) resp. unused parameter [-Wunused-parameter].
* MinGW: reg. this statement may fall through [-Wimplicit-fallthrough=] (at least one of which is a bug!).
* MSVC: reg. unreachable code (C4702).
* MinGW: reg. suggest parentheses around '&&' within '||' [-Wparentheses].
* MinGW: reg. variable set but not used [-Wunused-but-set-variable].
* MinGW: reg. suggest parentheses around assignment used as truth value [-Wparentheses].
* MSVC: reg. declaration hides global declaration (C4459).
* MSVC: reg. declaration hides function parameter (C4457).
* MSVC: reg. conversion requires a narrowing conversion (C4838)
* MSVC: reg. "structure was padded due to alignment specifier (C4324)", by disabling it
* GitHub CI: reg. Node.js 12 actions being deprecated vs. 16
* GitHub CI: reg. `set-output` command being deprecated
* XCode: reg. loop variable is always a copy because the range of type 'QJsonArray' does not return a reference [-Wrange-loop-analysis]
* MSVC: reg.  local variable is initialized but not referenced (C4189)

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -1376,7 +1376,8 @@ void Beam::setValidBeamPositions(int& dictator, int& pointer, int beamCountD, in
                     continue;
                 }
                 // we can use dictator beam position because all of the notes have the same beam position
-                if (currOffset = findValidBeamOffset(dictator, cr->beams(), staffLines, isStartDictator, isAscending, isFlat)) {
+                currOffset = findValidBeamOffset(dictator, cr->beams(), staffLines, isStartDictator, isAscending, isFlat);
+                if (currOffset) {
                     break;
                 }
             }
--- a/src/engraving/libmscore/chordline.cpp
+++ b/src/engraving/libmscore/chordline.cpp
@@ -472,6 +472,7 @@ bool ChordLine::setProperty(Pid propertyId, const PropertyValue& val)
         break;
     case Pid::CHORD_LINE_STRAIGHT:
         setStraight(val.toBool());
+        break;
     case Pid::CHORD_LINE_WAVY:
         setWavy(val.toBool());
         break;
--- a/src/engraving/libmscore/stretchedbend.cpp
+++ b/src/engraving/libmscore/stretchedbend.cpp
@@ -89,11 +89,11 @@ void StretchedBend::fillDrawPoints()
     }
 
     m_drawPoints.clear();
-    auto isExtraPointHelper = [this](int i, std::function<bool(int, int)> compare) {
+    auto isExtraPointHelper = [this](size_t i, std::function<bool(int, int)> compare) {
         return compare(m_points[i - 1].pitch, m_points[i].pitch) && compare(m_points[i].pitch, m_points[i + 1].pitch);
     };
 
-    auto isExtraPoint = [isExtraPointHelper](int i) {
+    auto isExtraPoint = [isExtraPointHelper](size_t i) {
         return isExtraPointHelper(i, std::less<int>())
                || isExtraPointHelper(i, std::greater<int>())
                || isExtraPointHelper(i, std::equal_to<int>());
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -755,7 +755,7 @@ Bracket* System::createBracket(const LayoutContext& ctx, BracketItem* bi, size_t
     //
     if (span > 1
         || (bi->bracketSpan() == span)
-        || (span == 1 && score()->styleB(Sid::alwaysShowBracketsWhenEmptyStavesAreHidden)) && bi->bracketType() != BracketType::SQUARE
+        || (span == 1 && score()->styleB(Sid::alwaysShowBracketsWhenEmptyStavesAreHidden) && bi->bracketType() != BracketType::SQUARE)
         || (span == 1 && score()->styleB(Sid::alwaysShowSquareBracketsWhenEmptyStavesAreHidden)
             && bi->bracketType() == BracketType::SQUARE)) {
         //
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -817,14 +817,7 @@ size_t System::getBracketsColumnsCount()
 void System::setBracketsXPosition(const double xPosition)
 {
     for (Bracket* b1 : _brackets) {
-        // Obtain correct distance
-        double bracketDistance = 0.0;
         BracketType bracketType = b1->bracketType();
-        if (bracketType == BracketType::NORMAL || bracketType == BracketType::LINE) {
-            bracketDistance = score()->styleMM(Sid::bracketDistance);
-        } else if (bracketType == BracketType::BRACE) {
-            bracketDistance = score()->styleMM(Sid::akkoladeBarDistance);
-        }
         // For brackets that are drawn, we must correct for half line width
         double lineWidthCorrection = 0.0;
         if (bracketType == BracketType::NORMAL || bracketType == BracketType::LINE) {
--- a/src/framework/audio/internal/synthesizers/fluidsynth/fluidsequencer.cpp
+++ b/src/framework/audio/internal/synthesizers/fluidsynth/fluidsequencer.cpp
@@ -27,9 +27,9 @@ using namespace mu::audio;
 using namespace mu::midi;
 using namespace mu::mpe;
 
-static constexpr mpe::pitch_level_t MIN_SUPPORTED_LEVEL = mpe::pitchLevel(PitchClass::C, 0);
+static constexpr mpe::pitch_level_t MIN_SUPPORTED_PITCH_LEVEL = mpe::pitchLevel(PitchClass::C, 0);
 static constexpr note_idx_t MIN_SUPPORTED_NOTE = 12; // MIDI equivalent for C0
-static constexpr mpe::pitch_level_t MAX_SUPPORTED_LEVEL = mpe::pitchLevel(PitchClass::C, 8);
+static constexpr mpe::pitch_level_t MAX_SUPPORTED_PITCH_LEVEL = mpe::pitchLevel(PitchClass::C, 8);
 static constexpr note_idx_t MAX_SUPPORTED_NOTE = 108; // MIDI equivalent for C8
 
 void FluidSequencer::init(const PlaybackSetupData& setupData)
@@ -228,15 +228,15 @@ channel_t FluidSequencer::channel(const mpe::NoteEvent& noteEvent) const
 
 note_idx_t FluidSequencer::noteIndex(const mpe::pitch_level_t pitchLevel) const
 {
-    if (pitchLevel <= MIN_SUPPORTED_LEVEL) {
+    if (pitchLevel <= MIN_SUPPORTED_PITCH_LEVEL) {
         return MIN_SUPPORTED_NOTE;
     }
 
-    if (pitchLevel >= MAX_SUPPORTED_LEVEL) {
+    if (pitchLevel >= MAX_SUPPORTED_PITCH_LEVEL) {
         return MAX_SUPPORTED_NOTE;
     }
 
-    float stepCount = MIN_SUPPORTED_NOTE + ((pitchLevel - MIN_SUPPORTED_LEVEL) / static_cast<float>(mpe::PITCH_LEVEL_STEP));
+    float stepCount = MIN_SUPPORTED_NOTE + ((pitchLevel - MIN_SUPPORTED_PITCH_LEVEL) / static_cast<float>(mpe::PITCH_LEVEL_STEP));
 
     return stepCount;
 }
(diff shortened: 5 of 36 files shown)


src/plugins/api/apitypes.h
```diff
@@ -481,18 +481,18 @@ enum class NoteHeadGroup {
...
-    INVALID       = char(mu::engraving::NoteType::INVALID)
+    NORMAL        = unsigned char(mu::engraving::NoteType::NORMAL),
+    ACCIACCATURA  = unsigned char(mu::engraving::NoteType::ACCIACCATURA),
+    APPOGGIATURA  = unsigned char(mu::engraving::NoteType::APPOGGIATURA),
+    GRACE4        = unsigned char(mu::engraving::NoteType::GRACE4),
+    GRACE16       = unsigned char(mu::engraving::NoteType::GRACE16),
+    GRACE32       = unsigned char(mu::engraving::NoteType::GRACE32),
+    GRACE8_AFTER  = unsigned char(mu::engraving::NoteType::GRACE8_AFTER),
+    GRACE16_AFTER = unsigned char(mu::engraving::NoteType::GRACE16_AFTER),
+    GRACE32_AFTER = unsigned char(mu::engraving::NoteType::GRACE32_AFTER),
```
Jojo-Schmitz, line 494: This one is 128 and as such too large for signed char

src/plugins/api/apitypes.h
```diff
@@ -481,18 +481,18 @@ enum class NoteHeadGroup {
...
+    INVALID       = unsigned char(mu::engraving::NoteType::INVALID)
```
Jojo-Schmitz, line 495: This one is 255, as as such too large for signed char

cbjeukendrup:
Wow. That looks like a MSVC quirk again! `this` is not even meaningful in that context.

Anyway, maybe this works?
```diff
-    QObject::connect(qmlEngine, &QQmlEngine::warnings, [](const QList<QQmlError>& warnings) {
+    QObject::connect(qmlEngine, &QQmlEngine::warnings, [=](const QList<QQmlError>& warnings) {
```

jeetee:
I'd rather expect you'd need the [=] on line 38 instead, as QObject::connect is not a static method?

cbjeukendrup:
Probably you're right. But QObject::connect is surely static here, but there is one overload that is non-static. That probably confuses MSVC. We have seen that before, that time with `disconnect` instead of `connect`.

Jojo-Schmitz:
> Wow. That looks like a MSVC quirk again! `this` is not even meaningful in that context.
> 
> Anyway, maybe this works?
> 
> ```diff
> -    QObject::connect(qmlEngine, &QQmlEngine::warnings, [](const QList<QQmlError>& warnings) {
> +    QObject::connect(qmlEngine, &QQmlEngine::warnings, [=](const QList<QQmlError>& warnings) {
> ```

No, doesn't help, still the same warning



> I'd rather expect you'd need the [=] on line 38 instead, as QObject::connect is not a static method?

No, doesn't even compile!

jeetee:
> Probably you're right. But QObject::connect is surely static here,

The static call should perhaps have the `receiver` parameter for MSVC to not confuse it with the method signature? Perhaps try `QObject::connect(qmlEngine, &QQmlEngine::warnings, this, [=](const QList<QQmlError>& warnings) {` given that is what the member-connect is supposed to expand to? (https://doc.qt.io/qt-5/qobject.html#connect-2)

Jojo-Schmitz:
> > Probably you're right. But QObject::connect is surely static here,
> 
> The static call should perhaps have the `receiver` parameter for MSVC to not confuse it with the method signature? Perhaps try `QObject::connect(qmlEngine, &QQmlEngine::warnings, this, [=](const QList<QQmlError>& warnings) {` given that is what the member-connect is supposed to expand to? (https://doc.qt.io/qt-5/qobject.html#connect-2)
> 
> EDIT: and drop the capture indication as the body doesn't need it: `QObject::connect(qmlEngine, &QQmlEngine::warnings, this, [](const QList<QQmlError>& warnings) {`

Neither compiles

Jojo-Schmitz:
OK, let's disable it

src/engraving/libmscore/chordline.cpp
```diff
@@ -469,6 +469,7 @@ bool ChordLine::setProperty(Pid propertyId, const PropertyValue& val)
         break;
     case Pid::CHORD_LINE_STRAIGHT:
         setStraight(val.toBool());
+        break;
```
Jojo-Schmitz, line 472: This fixes a bug!

Jojo-Schmitz:
More and more warnings are ramping up, I'd appreciate a review and merge...

src/plugins/api/apitypes.h
```diff
@@ -481,18 +481,18 @@ enum class NoteHeadGroup {
 };
 Q_ENUM_NS(NoteHeadGroup);
 
-enum class NoteType : char {
+enum class NoteType : unsigned char {
```
cbjeukendrup, line 484: Maybe we should change it to `uint8_t`. It's shorter 😉
We can also simply remove it (so that we get `int`). In fact, `mu::engraving::NoteType` is `int` too, so I doubt that we gain much by using a different type for the api version.

src/framework/audio/internal/synthesizers/fluidsynth/fluidsequencer.cpp
```diff
@@ -152,6 +152,7 @@ void FluidSequencer::appendControlSwitch(EventSequenceMap& destination, const mp
 void FluidSequencer::appendPitchBend(EventSequenceMap& destination, const mpe::NoteEvent& noteEvent,
                                      const mpe::ArticulationTypeSet& appliableTypes, const channel_t channelIdx, const int noteIdx)
 {
+    UNUSED(noteIdx);
```
cbjeukendrup, line 155: I think we can just remove this parameter completely

src/engraving/libmscore/stretchedbend.cpp
```diff
@@ -89,14 +89,14 @@ void StretchedBend::fillDrawPoints()
     }
 
     m_drawPoints.clear();
-    auto isExtraPointHelper = [this](int i, std::function<bool(int, int)> compare) {
+    auto isExtraPointHelper = [this](size_t i, std::function<bool(size_t, size_t)> compare) {
```
cbjeukendrup, line 92: I think the compare function should be unmodified, because `m_points[i - 1].pitch` etc. are `int`
```suggestion
    auto isExtraPointHelper = [this](size_t i, std::function<bool(int, int)> compare) {
```

src/engraving/libmscore/stretchedbend.cpp
```diff
@@ -89,14 +89,14 @@ void StretchedBend::fillDrawPoints()
...
     };
 
-    auto isExtraPoint = [isExtraPointHelper](int i) {
-        return isExtraPointHelper(i, std::less<int>())
-               || isExtraPointHelper(i, std::greater<int>())
-               || isExtraPointHelper(i, std::equal_to<int>());
+    auto isExtraPoint = [isExtraPointHelper](size_t i) {
+        return isExtraPointHelper(i, std::less<size_t>())
+               || isExtraPointHelper(i, std::greater<size_t>())
+               || isExtraPointHelper(i, std::equal_to<size_t>());
```
cbjeukendrup, line 99: (these too)
```suggestion
    auto isExtraPoint = [isExtraPointHelper](size_t i) {
        return isExtraPointHelper(i, std::less<int>())
               || isExtraPointHelper(i, std::greater<int>())
               || isExtraPointHelper(i, std::equal_to<int>());
```

src/engraving/libmscore/beam.cpp
```diff
@@ -1377,7 +1377,7 @@ void Beam::setValidBeamPositions(int& dictator, int& pointer, int beamCountD, in
                     continue;
                 }
                 // we can use dictator beam position because all of the notes have the same beam position
-                if (currOffset = findValidBeamOffset(dictator, cr->beams(), staffLines, isStartDictator, isAscending, isFlat)) {
+                if ((currOffset = findValidBeamOffset(dictator, cr->beams(), staffLines, isStartDictator, isAscending, isFlat))) {
```
cbjeukendrup, line 1380: I would prefer a more unambiguous notation:
```suggestion
                currOffset = findValidBeamOffset(dictator, cr->beams(), staffLines, isStartDictator, isAscending, isFlat);
                if (currOffset) {
```

src/engraving/libmscore/system.cpp
```diff
@@ -753,7 +753,7 @@ Bracket* System::createBracket(const LayoutContext& ctx, BracketItem* bi, size_t
     //
     if (span > 1
         || (bi->bracketSpan() == span)
-        || (span == 1 && score()->styleB(Sid::alwaysShowBracketsWhenEmptyStavesAreHidden)) && bi->bracketType() != BracketType::SQUARE
+        || ((span == 1 && score()->styleB(Sid::alwaysShowBracketsWhenEmptyStavesAreHidden)) && bi->bracketType() != BracketType::SQUARE)
```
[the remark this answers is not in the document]
Jojo-Schmitz: Fine by me

src/plugins/api/apitypes.h
Jojo-Schmitz: OK, let's remove it then

src/framework/musesampler/internal/apitypes.h
```diff
@@ -66,7 +66,7 @@ typedef struct ms_DynamicsEvent
 
 typedef struct ms_PedalEvent
 {
-    long _location_us;
+    long long _location_us;
```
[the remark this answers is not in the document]
Jojo-Schmitz: I see.

src/framework/audio/internal/synthesizers/fluidsynth/fluidsequencer.cpp
Jojo-Schmitz: OK

src/engraving/libmscore/stretchedbend.cpp
Jojo-Schmitz: OK

src/engraving/libmscore/stretchedbend.cpp
Jojo-Schmitz: OK

src/engraving/libmscore/beam.cpp
Jojo-Schmitz: OK

src/engraving/libmscore/system.cpp
```diff
@@ -815,14 +815,7 @@ size_t System::getBracketsColumnsCount()
...
 {
     for (Bracket* b1 : _brackets) {
-        // Obtain correct distance
-        double bracketDistance = 0.0;
         BracketType bracketType = b1->bracketType();
-        if (bracketType == BracketType::NORMAL || bracketType == BracketType::LINE) {
-            bracketDistance = score()->styleMM(Sid::bracketDistance);
-        } else if (bracketType == BracketType::BRACE) {
-            bracketDistance = score()->styleMM(Sid::akkoladeBarDistance);
-        }
```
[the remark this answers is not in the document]
mike-spa: Good catch @Jojo-Schmitz, this was un unintentional leftover from a previous solution I'd tried.

mike-spa (approved):

Jojo-Schmitz:
There are 3 new warning:
`src\framework\audio\internal/audiobuffer.h(62,63,64): warning C4324: 'mu::audio::AudioBuffer': structure was padded due to alignment specifier``

cbjeukendrup:
MSVC complains that we're making those structs/members bigger than their intrinsic size by specifying alignment. But in this case, that is exactly what we want to do; we want them to be in different memory cache lines, and therefore they must be padded. So we should just disable these warnings.

Jojo-Schmitz:
OK, it is disabled now ;-)

src/framework/vst/internal/vstplugin.cpp
```diff
@@ -173,7 +173,7 @@ void VstPlugin::stateBufferFromString(VstMemoryStream& buffer, char* strData, co
     static Steinberg::int32 numBytesRead = 0;
 
     buffer.write(strData, strSize, &numBytesRead);
```
Jojo-Schmitz, line 175: This seems to need some casting too, but I can't get it working.

Jojo-Schmitz:
Hmm, what on earth is now causing that unit test failure at https://github.com/musescore/MuseScore/actions/runs/3872958358/jobs/6602418643#step:9:1537 ?
As far as I can tell I did  othing but a rebase

cbjeukendrup:
Seems related to my work on the tests; that test seems to fail sometimes, and sometimes not...

Jojo-Schmitz:
OK, tests worked this time

igorkorsukov (approved):

PR #13260, opened 2022-09-09
Added option to run autobot tests on startup

HemantAntony:

When you add the environment variable "MU_RUN_AUTOBOT_TESTS_ON_START" and set it to 1, MuseScore on startup will run all autobot tests

### DIFF
--- a/src/appshell/internal/startupscenario.cpp
+++ b/src/appshell/internal/startupscenario.cpp
@@ -99,6 +99,12 @@ void StartupScenario::run()
     });
 
     interactive()->open(startupUri);
+
+    bool result = false;
+    std::istringstream(std::getenv("MU_RUN_AUTOBOT_TESTS_ON_START")) >> result;
+    if (result) {
+        dispatcher()->dispatch("autobot-show-scripts", actions::ActionData::make_arg1<bool>(result));
+    }
 }
 
 bool StartupScenario::startupCompleted() const
--- a/src/autobot/internal/autobotactionscontroller.cpp
+++ b/src/autobot/internal/autobotactionscontroller.cpp
@@ -31,7 +31,13 @@ static const mu::UriQuery SHOW_SCRIPTS_URI("musescore://autobot/scripts?sync=fal
 void AutobotActionsController::init()
 {
     dispatcher()->reg(this, "autobot-show-batchtests", [this]() { openUri(SHOW_BATCHTESTS_URI); });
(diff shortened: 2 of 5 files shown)


HemantAntony:
I'll also add instructions to wiki on how to do this when this gets merged :)

cbjeukendrup:
Interesting! Is this also a step towards running these tests on CI (for example, as a separate step in the `CI_utests` workflow)? That would be pretty cool. At least it would allow to catch some crashes, but we can probably add some automatic sanity checks to the test cases themselves to make them super useful on CI.

One suggestion, what do you think about using a command line option instead of an environment var? For example
```sh
./musescore --run-autobot-tests
```

src/appshell/internal/startupscenario.cpp
```diff
@@ -99,6 +99,12 @@ void StartupScenario::run()
     });
 
     interactive()->open(startupUri);
+
+    bool result = false;
+    std::istringstream(std::getenv("MU_RUN_AUTOBOT_TESTS_ON_START")) >> result;
+    if (result) {
+        dispatcher()->dispatch("autobot-show-scripts", actions::ActionData::make_arg1<bool>(result));
+    }
```
cbjeukendrup, line 107: Shouldn't this be done inside the async::Async::call block, i.e. after `m_startupCompleted = true;`? Or maybe in `onStartupPageOpened`?

HemantAntony:
I actually didn't know about that. I just did it because I thought it was annoying having to click on Diagnostic->Autobot->Show scripts. But I can definitely add something for that to this PR

I just used Environment variables because I saw there were a couple of those from the [wiki](https://github.com/musescore/MuseScore/wiki/Autobot-overview#configuration). I'll work on the command line options :)

src/appshell/internal/startupscenario.cpp
HemantAntony: It probably should. I didn't see that 😅

igorkorsukov:
I think running tests using the command line option is better. Environment variables are convenient for setting environment variables :) - those settings that do not change from run to run.

cbjeukendrup:
@HemantAntony Would you still like to revisit this PR?

HemantAntony:
Yeah I'm actually working on this PR. I implemented everything and it works great. But the problem is that the steps in the example tests are not working properly. It seems like they are outdated. I'm currently working on that part @cbjeukendrup

cbjeukendrup:
Closing due to inactivity

PR #13265, opened 2022-09-09
Invisible time sig fix

mike-spa:
Corrects the spacing in case of a mixture of visible/invisible items and visible/hidden staves. This substitutes the solution I did [here](https://github.com/musescore/MuseScore/pull/11624) with what should be a more general one. New vtest should explain the difference.


oktophonie (approved):

src/engraving/libmscore/segment.cpp
```diff
@@ -1194,9 +1194,19 @@ bool Segment::allElementsInvisible() const
...
-            return false;
+    for (Staff* staff : score()->staves()) {
+        if (!staff->visible()) {
+            continue;
+        }
+        int staffIdx = staff->idx();
+        if (system() && staffIdx < system()->staves().size() && !system()->staves().at(staffIdx)->show()) {
+            continue;
+        }
+        for (int track = staffIdx * VOICES; track < staffIdx * VOICES + VOICES; ++track) {
```
cbjeukendrup, line 1205: We should work with `staff_idx_t` and `track_idx_t` respectively instead of `int`. 

Also, maybe it's better to calculate `staffIdx * VOICES + VOICES` only once (instead of for every iteration of the inner loop) and store it into a variable. 

And I estimate it would also be beneficial to call `system()` only once, outside the outer loop (`for (Staff* staff : score()->staves()) {`).

cbjeukendrup: Hm, another small optimization: we loop over `score()->staves()`, but then for every Staff, we look up it's index using `staff->idx()`. I think would be more efficient to loop by using a counter:
```c++
for (staff_idx_t staffIdx = 0; staffIdx < score()->nstaves(); ++staffIdx) {
    const Staff* staff = score()->staves().at(staffIdx); // taking the nth element of a vector is a very efficient operation

    if (!staff->visible()) {
        …
```
(Not much more beautiful code, but probably quite a bit more efficient. We're probably still talking about microseconds though, but those also count.)

mike-spa: Done, thanks! 👍

cbjeukendrup (approved):

Issue #12658, opened 2022-08-03
[MU4 Issue] Default workspace name not translatable, because logic relies on it


cbjeukendrup:
**Describe the bug**
The name of the default workspace (`"Default"`) is not translatable, and we can't easily make it translatable because there is some logic relying on it (logic to determine which one is the built-in default workspace).

**To Reproduce**
1. See status bar, and see the source code, `workspacetypes.h`

We should make a better way to determine which workspace is the default one, instead of relying on the name. (I don't think we should still have logic relying on the name after we make it translatable. Relying on translatable names is asking for trouble, if you ask me.)

cbjeukendrup:
In fact, the only reason that we need to check whether a workspace is the default one, is that we don't allow deleting the default workspace. Do we really want to disallow this? Or do we rather want the the user can simply not delete the last remaining workspace (i.e. there always has to exist at least one workspace)?

Tantacrul:
@cbjeukendrup - we specified that the default workspace can't be deleted in case users really need it to revert settings they messed up. I don't have opinions on how this should be achieved.

Tantacrul:
yeah, you're right. Oh well then. The option to just enforce '1 workspace at all times' is good.

We'll need to spruce up this system in other releases.

PR #13279, opened 2022-09-10
Fix #10601: Added Edit and Reset workspace button

### LINKED ISSUE #10601: [MU4 Issue] Edit, Reset workspace options are missing in Workspace menu
**Describe the bug**
Edit, Reset workspace options are missing in Workspace menu (comparing with MU3)

https://user-images.githubusercontent.com/90187801/154520772-71b46019-be33-416b-940d-c8a92414eba3.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

HemantAntony:
Resolves: #10601 

### DIFF
--- a/src/appshell/view/appmenumodel.cpp
+++ b/src/appshell/view/appmenumodel.cpp
@@ -612,7 +612,9 @@ MenuItemList AppMenuModel::makeWorkspacesItems()
     }
 
     items << makeSeparator()
-          << makeMenuItem("configure-workspaces");
+          << makeMenuItem("edit-workspace")
+          << makeMenuItem("configure-workspaces")
+          << makeMenuItem("reset-workspace");
 
     return items;
 }
--- a/src/workspace/internal/workspaceactioncontroller.cpp
+++ b/src/workspace/internal/workspaceactioncontroller.cpp
@@ -29,13 +29,82 @@ using namespace mu::actions;
 void WorkspaceActionController::init()
 {
     dispatcher()->reg(this, "select-workspace", this, &WorkspaceActionController::selectWorkspace);
+    dispatcher()->reg(this, "edit-workspace", this, &WorkspaceActionController::editCurrentWorkspace);
     dispatcher()->reg(this, "configure-workspaces", this, &WorkspaceActionController::openConfigureWorkspacesDialog);
+    dispatcher()->reg(this, "reset-workspace", [this]() { setCurrentWorkspaceName(DEFAULT_WORKSPACE_NAME); });
 }
 
 void WorkspaceActionController::selectWorkspace(const ActionData& args)
 {
     std::string selectedWorkspace = !args.empty() ? args.arg<std::string>(0) : "";
     setCurrentWorkspaceName(selectedWorkspace);
+}
+
+void WorkspaceActionController::editCurrentWorkspace()
+{
+    IWorkspacePtrList workspaces = workspacesManager()->workspaces();
+    std::string workspaceNames;
+
+    for (const IWorkspacePtr& workspace: workspaces) {
+        workspaceNames = workspaceNames + workspace->name() + ",";
+    }
+    workspaceNames.pop_back();
+
+    UriQuery uri("musescore://workspace/create");
+    uri.addParam("sync", Val(true));
+    uri.addParam("workspaceNames", Val(workspaceNames));
+    uri.addParam("editWorkspace", Val(true));
+
+    RetVal<Val> obj = interactive()->open(uri);
+    if (!obj.ret) {
+        return;
+    }
+
+    IWorkspacePtr currentWorkspace = workspacesManager()->currentWorkspace();
+    if (!currentWorkspace) {
+        return;
+    }
+
+    QVariantMap meta = obj.val.toQVariant().toMap();
+    std::string name = meta.value("name").toString().toStdString();
+    if (name != currentWorkspace->name()) {
+        IWorkspacePtrList updatedWorkspaces;
+        for (IWorkspacePtr workspace : workspaces) {
+            if (workspace != currentWorkspace) {
+                updatedWorkspaces.emplace_back(workspace);
+            }
+        }
+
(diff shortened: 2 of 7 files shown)


src/workspace/internal/workspaceactioncontroller.cpp
```diff
@@ -29,7 +29,9 @@ using namespace mu::actions;
 void WorkspaceActionController::init()
 {
     dispatcher()->reg(this, "select-workspace", this, &WorkspaceActionController::selectWorkspace);
+    dispatcher()->reg(this, "edit-workspace", this, &WorkspaceActionController::editCurrentWorkspace);
     dispatcher()->reg(this, "configure-workspaces", this, &WorkspaceActionController::openConfigureWorkspacesDialog);
+    dispatcher()->reg(this, "reset-workspace", [this]() { setCurrentWorkspaceName(DEFAULT_WORKSPACE_NAME); });
```
cbjeukendrup, line 34: As I explained in https://github.com/musescore/MuseScore/issues/12658#issuecomment-1215150531, this will not really reset the workspace... not sure what to do about that.

HemantAntony: How about basing the default workspace of a "Default workspace template" not exposed to users. But we can call it the default workspace since for the users its the workspace that they use (by default).

So, whenever we want to reset it to default workspace, we just delete the default workspace and create a new one based off of the template.

src/appshell/view/appmenumodel.cpp
```diff
@@ -599,7 +599,9 @@ MenuItemList AppMenuModel::makeWorkspacesItems()
     }
 
     items << makeSeparator()
-          << makeMenuItem("configure-workspaces");
+          << makeMenuItem("edit-workspace")
+          << makeMenuItem("configure-workspaces")
+          << makeMenuItem("reset-workspace");
```
Eism, line 604: Maybe it is not a good place for 'edit-workspace' and 'reset-workspace'. 
Because these two options affect to current workspace, 'configure-workspaces' affects all workspaces. 

@Tantacrul 
What do you think about adding these two operations to the Workspaces dialog for each workspace? Like in Parts dialog

src/workspace/internal/workspaceactioncontroller.cpp
Eism: Actually 'reset workspace' doesn't work for me in MU3 at all. 
So I don't know how it is should work. 

In the beginning, I thought that this should reset the current workspace to default state.
But here I see that we are resetting the selection to the default workspace, not the state

Tantacrul:
I'm unable to test this because the builds have expired. @HemantAntony - can you fix this up and I'll take a look? Thanks!

HemantAntony:
@Tantacrul It should be okay now

Tantacrul:
@Eism 

> What do you think about adding these two operations to the Workspaces dialog for each workspace? Like in Parts dialog

I agree. @HemantAntony - I don't think this is the right place to put those options because they are hard to find and are much less clear than they would be if they were located in the 'Configure Workspaces' popup. I'd suggest leaving the options in the top menu but also creating some UI options in the Workspace popup itself.

Perhaps @bkunda can help with this?

Tantacrul:
Actually, I'm confused about what 'Reset Workspace' is supposed to do. In this build, it seems to simply choose the default workspace and that's it. Even if I change the default workspace, pressing 'Reset' doesn't do anything.

If this is all it does, then I don't think we should have this option at all.

MarcSabatella:
I can tell you what I would *want* it to, which is not what it has done in the past: I'd want it to reset any customizations I had made during that particular editing session.  So if I had a nice carefully customized workspace I had been using for weeks, then one day I accidentally messed it up, I could easily go back to the way it was when I started that session.  This would be easy to implement I think since as far as I know we don't actually write the customizations out until your either close the window or change workspaces.  So basically it would just amount to reloading the current workspace.

What it actually did in MU3 was something rather different, and sort of useful but not really.  When you create a workspace, it's cloned from whatever workspace you are in.  So you can start form either Basic and add to it, or from Advanced and subtracted from it.  Reset takes you back to either Basic or Advanced - whichever your workspace was based on.  You can also base one custom workspace off of another, which is nice, but reset still takes you all the way back to Basic or Advanced rather than back to the workspace you actually based yours on.  Seems like there was some technical reason for that but I don't remember.

Tantacrul:
So, I guess you'd need two options then (which is fine).

**Reset workspace** does what it says - resets the selected workspace back to the initial default the app is loaded with. Anything else is just confusing.

**Undo changes to workspace in this session** - again, does what it says, although I'd like to think of a snappier name :)

Tantacrul:
@HemantAntony - if you think you'd be comfortable implementing a design like we've just discussed, I can put that together for you?

HemantAntony:
@Tantacrul That would be great! I'll work on it :)

HemantAntony:
@Tantacrul Just pinging you in case you forgot

Tantacrul:
Hey @HemantAntony, I wasn't aware I needed to do anything. I did just try to grab a build but it had exprired.

Have you implemented changes and want them tested?

HemantAntony:
Ah no 😁. You had said that you would implement a design for 'Edit workspace' and 'Undo recent changes' in the above comments

Tantacrul:
Sure. I think you're right that it needs a visualisation (and possibly a little double-checking on the UX). I'll bump this over to @bkunda and @jessjwilliamson. They'll likely be able to bang it out quite quickly and give it to you.

Right now, I'm a bit too consumed with other things to do the design myself. Apologies!

RobFog:
With  #10601 supposedly implemented through #26384, can this pull request be closed, @Eism, @cbjeukendrup?

Eism:
Yes, we should close this

@HemantAntony hey! Thanks for your work! 
We have slightly reworked the workspaces, you can see how it looks in the master :)

HemantAntony:
@Eism yeah no problem :)

Issue #10951, opened 2022-03-29
[MU4 Issue] The user should be able to unmute hidden instruments


DmitryArefiev:
**Describe the bug**
See #9357

If you hide an instrument using the 'Instruments' panel, the instrument should be muted in the mixer. 
The user **should be able to unmute** it if they want

https://user-images.githubusercontent.com/90187801/160684542-0789944b-9c0a-4f97-a989-697c02ee5476.mp4

**Expected result**
You should give user ability to unmute hidden instruments in Mixer 
(or implement something in Instruments panel - see https://github.com/musescore/MuseScore/issues/9357#issuecomment-950046752 )

**Platform information**
macOS, Windows, Linux

MarcSabatella:
This isn't a feature request, it's a regression - it has always worked in the past and lots of people rely on it as a way to get all sorts playback effects you can't otherwise (rhythm section parts notated with just slashes, various ornaments and other notations not supported for playback directly.  I have verified this with one of my big band scores; no drum playback.

The good news is, it seems you *can* still access the mute controls, if instead of hiding the *instrument*, you hide its *staff* (or staves).  So it seems the functionality is still there (thankfully!) - it just for some reason isn't working as it should when hiding the full instrument instead of its staff/staves.

It's still a regression in that it means any imported 3.x score using invisible playback staves won't play correctly without employing a hard-to-discover workaround.  And even for new scores, it isn't necessarily obvious that you even can hide the staves individually or that this would work when hiding the instrument doesn't.

Tantacrul:
Never mind. I found it.

Tantacrul:
@bkunda - would be good to perhaps consider a design here. Perhaps just the ability to toggle on/off the sound in the mixer?

bkunda:
> @bkunda - would be good to perhaps consider a design here. Perhaps just the ability to toggle on/off the sound in the mixer? Perhaps something in the Instruments panel itself?

Do we really need a design for this? (Maybe I'm missing something).

I think at this stage our goal should simply be to restore the functionality as it was in MS3.6, which would mean: 

1. Not automatically muting an instrument that has its stave(s) hidden in the instruments panel
2. Allowing the user the ability to mute and unmute any hidden instrument in the mixer

[...]

PR #13282, opened 2022-09-10
Don't force-mute hidden staves

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/10951

Scores that use invisible playback staves were playing incorrectly - the invisible staves were automatically, with no ability to unmute.  This applies to existing score, but it also means new scores could not create invisible playback staves at all.  This PR restores the MU3 behavior, where hiding an instrument does not automatically mute it.  The user can still mute and unmuted the instrument normally.

(I did not include a test for this, as I'm not sure how these work anymore)

### DIFF
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -1043,13 +1043,14 @@ void PlaybackController::updateMuteStates()
         }
 
         const Part* part = notationParts->part(instrumentTrackId.partId);
-        bool isPart = part;
+        bool isMasterNotation = m_notation == m_masterNotation->notation();
+        bool isPartActive = part && (isMasterNotation || part->show());
 
         auto soloMuteState = audioSettings()->soloMuteState(instrumentTrackId);
 
         bool shouldBeMuted = soloMuteState.mute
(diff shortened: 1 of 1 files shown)


cbjeukendrup:
I'm not sure whether this has a desired effect on excerpts. If you show and then hide again a staff in an excerpt, that staff would unexpectedly still playback.

MarcSabatella:
Good point!  It was a little too good to be true to hope that simply change would do it :-).  But still, it probably needs just a slight modification where we still check for part->show() but only in excerpts.  Any recommendation on the best way to check if this is an excerpt in this context?

cbjeukendrup:
I believe the best way currently is the following:
```c++
bool isMasterNotation = m_notation == m_masterNotation->notation();
```
(The INotation interface lacks a method to check if it's a master notation, so this hack is used in numerous places...)

MarcSabatella:
Thanks!  I had noticed constructs like that but it seemed hacky :-)

RomanPudashkin:
I am going to archive this PR as we are going to implement a more general solution for future releases. Thank you!

@bkunda FYI

Issue #13258, opened 2022-09-09
[MU4 Issue] Address Sanitizer on GitHub CI is not working, and unit tests contain multiple undetected memory errors


cbjeukendrup:
We are thinking that we are using Address Sanitizer (asan) when running unit tests on GitHub CI:
https://github.com/musescore/MuseScore/blob/e30f0723617e69dfaa8e59b87a588280f306a29a/.github/workflows/ci_utests.yml#L11-L14

But apparently that doesn't work. I tried out what would happen if we enable asan using the dedicated CMake option. See my changes here: https://github.com/cbjeukendrup/MuseScore/commit/aa7e2cbc48f694ab563ad7b591caacebd4d3f22e

With these changes, asan starts detecting multiple errors: https://github.com/cbjeukendrup/MuseScore/runs/8259012256?check_suite_focus=true#step:8:1090

That must mean that the old way of enabling asan had no effect.

We should fix the errors reported by asan, and then switch to the working way of enabling asan, to catch some regressions before they get merged.

This is not super alarming, since many errors reported by asan hardly ever produce a crash in normal builds, but still, sometimes they _do_ cause a (hardly reproducible/understandable) crash.

PR #13284, opened 2022-09-10
Fix enabling Address Sanitizer on CI

cbjeukendrup:
Resolves: #13258

Contains real fixes for some of the reported memory errors, and temporary dirty fixes for the other ones.


cbjeukendrup (commented):
The remaining errors that still need a proper solution can be categorized as follows:
- Often, child elements are already deleted in the destructor of EngravingObject when their parent is deleted. We should research whether existing "ad hoc" logic to delete child elements should be adjusted or removed.
- The relation between parent tuplets and child DurationElements is not clear. Research is needed to determine which one should delete the other.
- Sometimes both a parent element and its child are in separate `RemoveElement` actions in the UndoStack. When the UndoStack is deleted, the elements stored in the commands will be deleted; in that case, the child element is sometimes deleted by the parent, and sometimes by the UndoStack. It's not clear which is correct.

cbjeukendrup:
Every input on this would be highly appreciated!

cbjeukendrup:
I turned on Address Sanitizer for the vtests too. Apparently on Linux, Address Sanitizer also includes Leak Sanitizer, and woah... that's a huge amount of leaks. You can see them here: https://github.com/musescore/MuseScore/actions/runs/3177233016/jobs/5177448025#step:7:1031 (it takes a while for that the huge log is loaded).

I'm going to turn that off for now, because it's really too much to handle right now. So we'll only have Address Sanitizer.

Hopefully we manage to fix the remaining existing errors, so that we can merge this, since it would be quite a nice tool to catch regressions. For example it would have prevented the crash that's fixed by #13594.

igorkorsukov (approved):

igorkorsukov (approved):

cbjeukendrup:
This is finally ready! I've found a proper solution for all problems, and also added some compiler warning fixes.

src/engraving/libmscore/tuplet.cpp
```diff
@@ -893,6 +897,21 @@ void Tuplet::remove(EngravingItem* e)
...
+{
+    assert(!_beingDestructed);
+    _allElements.insert(de);
+}
+
+void Tuplet::removeDurationElement(DurationElement* de)
+{
+    if (_beingDestructed) {
+        return;
+    }
```
cbjeukendrup, line 910: This is only an optimisation; it has no functional value

.github/workflows/ci_utests.yml
```diff
@@ -49,4 +46,6 @@ jobs:
         # So, we remove unnecessary tools, for more free space
         sudo docker system prune -a -f
         sudo rm -rf /usr/local/lib/android
-        sudo bash ./build/ci/linux/runutests.sh
\ No newline at end of file
+        sudo bash ./build/ci/linux/runutests.sh
+      env: 
+        ASAN_OPTIONS: "detect_leaks=0"
```
Jojo-Schmitz, line 51: Never check for error conditions you can't handle ;-)

igorkorsukov (approved):

PR #13280, opened 2022-09-10
Fix #13248: Added dialog message when adding text

HemantAntony:
Resolves: #13248

*(short description of the changes and the motivation to make the changes)*

Added dialog message when adding text and nothing is selected

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

Issue #13248, opened 2022-09-08
[MU4 Issue] Missing dialog message content for when adding certain text elements without selection


bkunda:
**Describe the bug**
This dialog is triggered when attempting to add a text element without a note/rest selection:

<img width="310" alt="Screenshot 2022-09-08 at 3 46 04 pm" src="https://user-images.githubusercontent.com/86290556/189140291-8462d638-4f97-44f0-9806-cf68c98383f6.png">

The dialog is missing critical message content.

This relates to all of these text elements (except "Figured bass", which for some reason is still returning the correct dialog messaging):

<img width="217" alt="Screenshot 2022-09-08 at 3 54 28 pm" src="https://user-images.githubusercontent.com/86290556/189140719-a4a7339a-a261-4318-a050-ca0776aa4aaf.png">

**To Reproduce**
1. Create a new score
2. Select Add > Text > System text 
3. See above faulty dialog
4. Repeat for other text elements in the same menu

**Expected behavior**
This dialog should be triggered when attempting to add one of the above text elements and when no note/rest is selected:

<img width="474" alt="Screenshot 2022-09-08 at 3 57 47 pm" src="https://user-images.githubusercontent.com/86290556/189141336-87b2ff20-3e9b-4d7c-ba23-30ec5694daee.png">

**Platform information**
 - MacOS 12.3.1 (M1)
 - Windows 11

HemantAntony:
@bkunda are you sure a dialog like this is needed in the first place. I remember talking about this in separate issue. It was sort of agreed on that a dialog is not really needed. See https://github.com/musescore/MuseScore/issues/10888#issuecomment-1079021187

Tantacrul:
I think in this instance the dialog is needed because the rules behind when you can / can't add text are quite unintuitive.

I get caught by this problem a lot. So... we should do exactly what is suggested in @bkunda's description.

In future, we'll just change the behaviour to avoid this problem. However, we don't do it until 4.x

Nice catch @bkunda !

Tantacrul:
@RomanPudashkin - back to you

HemantAntony:
@bkunda Please check #13280

PR #13287, opened 2022-09-11
fix #13286 - ensure no empty error message shown when adding text and nothing selected

### LINKED ISSUE #13286: [MU4 Issue] Empty dialog when adding text (chord symbols, lyrics etc.) and nothing selected
**Describe the bug**
If you press Ctrl+K to insert chord symbols before selecting anything, you get a strange empty dialog (with an option not to show again).  This applies to various types of text input.

**To Reproduce**
1. Open or create any score, ensure nothing selected
2. Press Ctrl+K (Ctrl+L and other similar commands too)
3. Observe dialog with no text

**Expected behavior**
Should explain something needs to be selected.
Note that if you select a measure/range first you DO get a dialog saying a note/rest needs to be selected, which is also unexpected - there's no reason it can't just start adding the chord symbol/harmony element at the first note/rest in the range. However this latter behaviour is the same in v3.6.

**Screenshots**

**Platform information**
 - Windows 10

**Additional context**
Built from latest source, 11-Sep-22.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/13286)

### DIFF
--- a/src/framework/uicomponents/view/popupview.cpp
+++ b/src/framework/uicomponents/view/popupview.cpp
@@ -71,6 +71,9 @@ PopupView::PopupView(QQuickItem* parent)
 
 PopupView::~PopupView()
 {
+    if (m_window) {
+        m_window->setOnHidden(std::function<void()>());
+    }
     m_contentItem->deleteLater();
     delete m_closeController;
 }
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -3880,16 +3880,12 @@ void NotationInteraction::addTextToTopFrame(TextStyleType type)
 
 mu::Ret NotationInteraction::canAddTextToItem(TextStyleType type, const EngravingItem* item) const
 {
-    if (!item) {
-        return false;
-    }
-
     if (isVerticalBoxTextStyle(type)) {
-        return item->isVBox();
+        return item && item->isVBox();
     }
 
     if (type == TextStyleType::FRAME) {
-        return item->isBox();
+        return item && item->isBox() ? make_ok() : make_ret(Err::EmptySelection);
     }
 
     static const std::set<TextStyleType> needSelectNoteOrRestTypes {
@@ -3918,7 +3914,7 @@ mu::Ret NotationInteraction::canAddTextToItem(TextStyleType type, const Engravin
             ElementType::CHORD,
         };
 
-        bool isNoteOrRestSelected = mu::contains(requiredElementTypes, item->type());
+        bool isNoteOrRestSelected = item && mu::contains(requiredElementTypes, item->type());
         return isNoteOrRestSelected ? make_ok() : make_ret(Err::NoteOrRestIsNotSelected);
     }
 


cbjeukendrup:
@wizofaus Your issue is a duplicate of #13248, and this PR is a duplicate of #13280 😅

HemantAntony:
I actually did think so. But his solution was different from mine.

wizofaus:
I did search but couldn't find anything. Hemant's solution will show the same "select a note or rest" message in every case you add any type of text element and nothing is selected, which wasn't an assumption I thought it was safe to make.

Tantacrul:
@HemantAntony / @wizofaus - are you in agreement about which of the issues we should choose to test?

wizofaus:
Actually I'm not convinced that either solution covers all cases properly - really that function to determine if you can add a text element based on the current selection should never return just "false", or you'll get the same empty message. I'd prefer to change this one to ensure that's never possible.

wizofaus:
In particular it's still showing an empty message if you use the "add text frame" command which doesn't appear to be in any menu by default, but you can define a shortcut for and use.

src/framework/uicomponents/view/popupview.cpp
```diff
@@ -71,6 +71,9 @@ PopupView::PopupView(QQuickItem* parent)
 
 PopupView::~PopupView()
 {
+    if (m_window) {
+        m_window->setOnHidden(std::function<void()>());
+    }
```
wizofaus, line 76: Occasionally the "onHidden" handler would crash if it activated after this object had been deleted

src/notation/internal/notationinteraction.cpp
```diff
@@ -3880,16 +3880,12 @@ void NotationInteraction::addTextToTopFrame(TextStyleType type)
...
-    }
-
     if (isVerticalBoxTextStyle(type)) {
-        return item->isVBox();
+        return item && item->isVBox();
     }
 
     if (type == TextStyleType::FRAME) {
-        return item->isBox();
+        return item && item->isBox() ? make_ok() : make_ret(Err::EmptySelection);
```
wizofaus, line 3888: technically the message is wrong, the selection might not be empty, just not valid for this command - but the only way this can happen is by assigning a shortcut key to the "add frame text" command.

src/notation/internal/notationinteraction.cpp
```diff
@@ -3880,16 +3880,12 @@ void NotationInteraction::addTextToTopFrame(TextStyleType type)
 
 mu::Ret NotationInteraction::canAddTextToItem(TextStyleType type, const EngravingItem* item) const
 {
-    if (!item) {
-        return false;
-    }
-
     if (isVerticalBoxTextStyle(type)) {
-        return item->isVBox();
+        return item && item->isVBox();
```
wizofaus, line 3884: NB currently this can never happen because the only calling function already handles this case

Tantacrul:
@bkunda - when you have time, can you test this and let me know if you're happy that it solves the 'don't show an empty error message' issue?

bkunda:
> @bkunda - when you have time, can you test this and let me know if you're happy that it solves your '**don't show an empty error message**' issue?

Yes this fix does the job 👍🏻

I'd really prefer it if the text formatting could be corrected for both this and also the dialog with the "figured bass" alternate messaging (if this is possible):

_Current_
Screenshot 2022-09-19 at 12 50 11 pm

_Preferred_
Screenshot 2022-09-19 at 12 51 20 pm

But I seem to recall an earlier discussion where this was deemed not possible, in which case this PR would have to be accepted as-is.

bkunda:
> In particular it's still showing an empty message if you use the "add text frame" command and no horizontal or vertical frame is selected. 

I did notice that no messaging dialog is triggered for frames. The corresponding dialog in MS3 bore this message: 

Screenshot 2022-09-19 at 12 58 10 pm

For now I'd be happy if this messaging could be copied-over to the new dialog style.

bkunda:
The instruction to "please select a bar" is not incorrect. If the user follows this instruction, they will be able to achieve their goal of inserting a frame. Certainly, it isn't the _only_ element they could select to insert a frame, but listing out every element they could potentially select would be cumbersome and unhelpful, and trying to generalise the message could potentially lead to unintended results (E.g. "please select something...", which could see the user selecting an element that _doesn't_ permit a frame). 

It's not brilliant messaging, but it's a reasonable solution and I don't think there's any great danger in keeping it for now. 

As for the figured bass, I've had another look at how this works and another solution here – perhaps a simpler one – would be just to have the one dialog for all these text elements (I.e. "please select a note or rest..."). You can select a figured bass number and then Add > Figured bass, but this doesn't add a figured bass so much as enter you into edit mode for the selected figured bass. So having one consistent dialog for all text elements here (Or rather, not having an exceptional dialog for figured bass) would actually make more sense. 

And yes, my comments above concerned the UI. Incidentally, the "don't show again" was considered a courtesy for users who found this kind of prompt frustrating. And we weren't too worried about the UX here because the user would be aware that they've elected to not show this type of message. However if this generates a lot of confusion then of course we are always open to tweaking the UX once we get more feedback from users.

MarcSabatella:
I would also mention, in the past it always *was* necessary to select a full bar when inserting a frame.  Which makes some logical sense, because even if you select some random element *within* the bar, the frame isn't going to literally be inserted right there - it's still going to be inserted in front of the bar as a whole.  And the palette at least does continue to require that, even if the Add menu does not.  So I'd think of the occasional ability to select a random element within a bar as more of a bone to people with sloppy fingers than as something we want to encourage.

However, there is one element type that it *is* perhaps important to know you can select - another frame.  So ideally, the message would say "select a bar or frame".  And of course, it should actually say "measure" in the code, then the UK translation would change it to "bar".

bkunda:
That makes sense to me @MarcSabatella.

So @wizofaus if I were to refine the specification of this fix, based on our discussion above, I would say: 

1. Just this one dialog for all text elements that currently trigger it (incl. figured bass):
Screenshot 2022-09-19 at 5 55 56 pm

2. And this dialog for frames: 
Screenshot 2022-09-19 at 5 56 28 pm

I would also add that the checkbox in both cases should be **unchecked** by default.

Does this all make sense?

wizofaus:
I believe that will need to wait for 4.1 given the "string freeze".
I think we can agree that this PR at least improves the current situation in 4.0.

bkunda:
> I believe that will need to wait for 4.1 given the "string freeze".
> I think we can agree that this PR at least improves the current situation in 4.0.

Yes I believe you're right here. 

Is the UI tweak something you were able to do @wizofaus?

cbjeukendrup:
I'm inclined to say we shouldn't worry _too_ much about string freeze, since we already have a bunch of changes/additions that are waiting to be pushed anyway. Since there's still enough time until the release, I think a few strings less or more doesn't make a huge difference.

Tantacrul:
> I'm inclined to say we shouldn't worry too much about string freeze,

Exactly. We did mention that there would be a few changes - but we'd keep them to a minimum. We generally agreed it was best to initiate translations rather than wait for the last few stragglers to make their way in.

bkunda:
> > I'm inclined to say we shouldn't worry too much about string freeze,
> 
> Exactly. We did mention that there would be a few changes - but we'd keep them to a minimum. We generally agreed it was best to initiate translations rather than wait for the last few stragglers to make their way in.

Ok well let's push this forward then :-) 

@wizofaus please let me know if/when you can implement the tweaks we've [specified here](https://github.com/musescore/MuseScore/pull/13287#issuecomment-1251218002). I'll give it another test before recommending it for merging.

RomanPudashkin (approved):

Issue #12336, opened 2022-07-11
[MU4 Issue] Values do not always update when moving between text styles in style dialog


oktophonie:
**To Reproduce**
Open the Style dialog and choose 'Text Styles' on the left:
![image](https://user-images.githubusercontent.com/20604032/178233631-cd423bde-6259-4e5b-98a8-109e5532fcac.png)
'Title' is selected in the list by default; note that 'Size' is 22pt.

Now choose another style, e.g. 'Hairpin':
![image](https://user-images.githubusercontent.com/20604032/178233871-051a9ba7-253e-4a87-b9c5-8b89ca65f696.png)
The actual size of this style is 10pt, but not that  'Size' does not update (it remains at 22pt).

_Sometimes_ the size changes as you move around this list, but not consistently. I'm not sure if this affects only the 'Size' field, or others too.

It also seems to affect only certain styles for some reasons - Hairpin and Volta are two. But this does make it effectively impossible to change the values in these styles.

**Expected behavior**
All values should update correctly.

**Screenshots**

**Platform information**
 - OS: Arch Linux

rgreen5:
Looks similar to #11262?

oktophonie:
Note that this particularly seems to afflict the Hairpins style (possibly among others), to the extent that changing styles here has no effect. This means there's no way to change font for hairpin text (which also includes cresc./dim. text lines), short of creating a style sheet with them explicitly in and applying it.

PR #13292, opened 2022-09-11
Fix editing Hairpin text style and others in EditStyle

cbjeukendrup:
Resolves: #12336
Resolves: #13420

The problem (also existed in MU3, by the way): 
A "TextStyle" was defined as a set of 14 (*) pairs of a style ID and a property ID. In EditStyle, we assumed that the property IDs are always the same. This is not true: for example, some text styles work with `Pid::FONT_FACE`, others with `Pid::BEGIN_FONT_FACE`. So for styles with the latter, loading or editing them doesn't work.

The solution is to introduce a dedicated TextStylePropertyType enum, that can always be used to identify properties of a text style. That fixes the problem for now, with not too big refactoring.

However, there's a lot more to be done with text styles. It is bad that the Pids are stored as part of the TextStyle; they should depend on the EngravingItem in question, rather than the TextStyle, since a text style is not necessarily bound to a certain kind of EngravingItem. The result of the current situation is that if you apply the "hairpin" style to a regular text element, we try to modify `Pid::BEGIN_FONT_FACE`, but that has no effect on the text element, because the text element listens to `Pid::FONT_FACE`. This is not fixed in this PR, and would be slightly less urgent and more risky. If someone from the community would like to volunteer, you're welcome!

I'm considering the idea to do a huge refactor of TextStyles in 4.x. They could be much more powerful; for example, it would be very useful if styles could be based on others, so that if you want to chance the font family for the whole score, you only need to change it for the base style, instead of for every individual style. Anyway, that's for the future... 

(\*) For one or more TextStyles, it turns out that some values are missing, because some style IDs do not exist! That does not cause serious problems though (but this is the reason that we need TextStylePropertyType::Undefined = 0, because if the std::array constructor contains not enough elements, the remaining ones will be zero-initialized.)


RomanPudashkin (approved):

RomanPudashkin:

cbjeukendrup:
Done!

PR #13308, opened 2022-09-12
Slur over grace notes refinement

mike-spa:
Refined slur collision checks to include grace notes of the first chord (which were excluded until now). Incidentally, I've also optimized the code, so the whole collision avoidance now runs approximately 10% faster despite having added more logic to it. The difference is expected to show in the vtests.


oktophonie (approved):

src/engraving/libmscore/slur.cpp
```diff
@@ -366,8 +366,44 @@ void SlurSegment::adjustEndpoints()
...
+    Shape segShape = seg->staffShape(startStaffIdx).translated(seg->pos() + seg->measure()->pos());
+    // If cross-staff, also add the shape of second staff
+    if (slur()->isCrossStaff() && seg != startCR->segment()) {
+        staff_idx_t endStaffIdx = (endCR->staffIdx() != startStaffIdx) ? endCR->staffIdx() : endCR->vStaffIdx();
+        SysStaff* startStaff = system()->staves().at(startStaffIdx);
+        SysStaff* endStaff = system()->staves().at(endStaffIdx);
+        double dist = endStaff->y() - startStaff->y();
+        Shape secondStaffShape;
+        secondStaffShape.add(seg->staffShape(endStaffIdx).translated(seg->pos() + seg->measure()->pos()));
+        secondStaffShape.translate(PointF(0.0, dist));
```
cbjeukendrup, line 381: Why not 
```c++
Shape secondStaffShape = seg->staffShape(endStaffIdx).translated(seg->pos() + seg->measure()->pos()));
secondStaffShape.translate(PointF(0.0, dist));
```
or even 
```c++
Shape secondStaffShape = seg->staffShape(endStaffIdx).translated(seg->pos() + seg->measure()->pos() + PointF(0.0, dist)));
```
?

src/engraving/libmscore/slur.cpp
```diff
@@ -366,8 +366,44 @@ void SlurSegment::adjustEndpoints()
...
+            return false;
+        }
+        // Lyrics and fingering
+        const EngravingItem* item = shapeEl.toItem;
+        if (item->isLyrics() || item->isFingering()) {
+            return true;
+        }
+        // Items belonging to its own start chord
+        EngravingItem* parentItem = item->parentItem();
+        if (parentItem && parentItem->isChord() && toChord(parentItem) == startCR) {
```
cbjeukendrup, line 396: Do you specifically want to check for Chords rather than ChordRests here? 

If no, then this could be just `if (item->parentItem() == startCR)`. Non-nullness of `item->parentItem()` is then guaranteed because `startCR` is not null; and `toChord` returns exactly the same pointer as you feed it, only C++ interprets it as a pointer to a different type, but it will be the same pointer with the same address. And if the address is the same, you are already sure that it is _exactly_ the same object in memory and that it _must_ be a ChordRest because `startCR` is a ChordRest.

If yes, then `toChord` is still not necessary, but won't do much harm either, since it does (almost) nothing at runtime, but rather at compile time.

src/engraving/libmscore/slur.cpp
mike-spa: 👍 Though I prefer the first option, as I'd still rather keep the vertical translation (which is the tricky one cause it depends on staff distance) visually separate from the horizontal.

src/engraving/libmscore/slur.cpp
mike-spa: In the rare occurrence that a slur starts on a rest, the rest normally doesn't carry any item (flags are especially annoying) that may interfere with the slur position, hence why I was excluding it. Though I think it is indeed easier (and more general) if I just compare the ChordRest pointer

cbjeukendrup (approved):

PR #13309, opened 2022-09-12
Separate flows for "Save to cloud" and "Publish"

cbjeukendrup:
There are two flows:
- "Save to cloud": turns the project into a cloud score. This is very similar to the classic "Save as" command: the only difference is that the score will be saved on MuseScore.com rather than locally. Any subsequent File > Save or Cmd+S will also save the score to the cloud, just like with classic "Save as".

    (in fact, my current idea is that a local copy of the score will be stored in a cache folder on your computer, to facilitate some things, but that's an implementation detail.)

- "Publish": MS3 style uploading. Uploads a copy of the score to MuseScore.com once, but leaves the score itself where it is (which may be either locally or in the cloud). The ability to re-upload to an existing score is planned to be retained, just like it worked in MS3.

Although some dialogs and some logic is shared between the two flows, we should not confuse them. This PR (almost fully) implements proper separation between the flows, while still keeping the code DRY ("don't repeat yourself").
This PR also documents some of the remaining tasks as TODOs.

We should be careful with the concept of "source URLs". Each of the "flows" has its own version of it: the "Save to cloud" flow _needs_ a URL because that determines where the score is saved, and the "Publish" flow _can optionally use_ a URL that tells where this score was published last time. Since it is also possible to publish a project that is saved in the cloud, we need to store both URLs separately. In this regard, this PR doesn't change anything yet.

### DIFF
--- a/src/project/internal/projectactionscontroller.cpp
+++ b/src/project/internal/projectactionscontroller.cpp
@@ -99,7 +99,7 @@ INotationSelectionPtr ProjectActionsController::currentNotationSelection() const
 
 bool ProjectActionsController::canReceiveAction(const actions::ActionCode& code) const
 {
-    if (!m_saveToCloudAllowed) {
+    if (m_isUploadingProject) {
         if (code == "file-save-to-cloud" || code == "file-publish") {
             return false;
         }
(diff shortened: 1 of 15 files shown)


src/project/internal/projectactionscontroller.cpp
```diff
@@ -508,7 +535,7 @@ void ProjectActionsController::saveProjectToCloud(const SaveLocation::CloudInfo&
     m_uploadingProgress = uploadingService()->uploadScore(*projectData, meta.title, meta.source);
 
     m_uploadingProgress->started.onNotify(this, [this]() {
-        m_saveToCloudAllowed = false;
+        m_isUploadingProject = false;
```
RomanPudashkin, line 538: You need to change the behavior of this flag after renaming it

cbjeukendrup: 🤦 
Fixed!

RomanPudashkin (approved):

PR #13312, opened 2022-09-12
[MU4] Straight flags now spaced better

asattely:
Traditional flags on notes are reduced outside the staff such as to make sure the hook ends up at a certain vertical point, but this is meaningless for straight flags. The shortening that straight-flagged notes do now follows the same rules as beams:
image

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1555,7 +1555,11 @@ int Chord::maxReduction(int extensionOutsideStaff) const
         { 0, 1, 1, 1, 1 }, // 2 beams
         { 0, 0, 0, 1, 1 }, // 3 beams
     };
-    int beamCount = _hook ? 0 : beams();
+    bool hasTradHook = _hook && !score()->styleB(Sid::useStraightNoteFlags);
+    int beamCount = hasTradHook ? 0 : beams();
+    if (_hook && !hasTradHook) {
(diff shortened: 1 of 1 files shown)


oktophonie (approved):

src/engraving/libmscore/chord.cpp
```diff
@@ -1830,6 +1836,15 @@ void Chord::layoutStem()
 {
     // we should calculate default stem length for this chord even if it doesn't have a stem
     // because this length is used for tremolos or other things that attach to where the stem WOULD be
```
cbjeukendrup, line 1838: Shouldn't this comment stay together with the line `_defaultStemLength = calcDefaultStemLength();`?

cbjeukendrup (approved):
It feels a little bit scary to layout the hook potentially before the stem has been created. This means we can't use the stem in Hook::layout. However, as long as we are aware of that, it's not per se a problem, and I currently don't see any reason why we should depend on the stem in Hook::layout. To be fair, it's the other way around at the moment. 
For the rest a cosmetic comment only!

RomanPudashkin:
@asattely please rebase it

RomanPudashkin (approved):

RomanPudashkin:
@asattely sorry, but please rebase it again...

src/engraving/libmscore/chord.cpp
asattely: yep, good catch, fixed

asattely:
Rebased

PR #13314, opened 2022-09-12
fix #13303 - ensure tie flip via x shortcut works properly

### LINKED ISSUE #13303: [MU4 Issue] Flipping tie with 'x' does not trigger redraw
**Describe the bug**
When flipping ties via the keyboard shortcut 'x', a redraw is not triggered (i.e. the tie _is_ flipped, but this is not visibly evident until something else triggers a redraw).

Also if this 'flip' is undone with Ctrl+Z,  the tie is removed - that is, it looks like the flip action isn't put on the undo stack either.

Flipping either via the toolbar button or in the Properties panel works as expected. (Slurs work as expected in all cases.)

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a tie to two notes of the same pitch
2. Select the tie
3. Press 'x'; observe nothing happens
4. Trigger a relayout (by adding a system break, or whatever) - note the tie will now flip

**Expected behaviour**
Ties should visibly flip immediately, and undoing them should revert the flip.

wizofaus:
Resolves: [*(direct link to the issue)*](https://github.com/musescore/MuseScore/issues/13303)

### DIFF
--- a/src/engraving/libmscore/slur.cpp
+++ b/src/engraving/libmscore/slur.cpp
@@ -149,10 +149,6 @@ static ChordRest* searchCR(Segment* segment, track_idx_t startTrack, track_idx_t
 
 bool SlurSegment::isEditAllowed(EditData& ed) const
 {
-    if (ed.key == Key_X && !ed.modifiers) {
-        return true;
-    }
-
     if (ed.key == Key_Home && !ed.modifiers) {
         return true;
     }
@@ -188,14 +184,10 @@ bool SlurSegment::edit(EditData& ed)
 
     Slur* sl = slur();
 
-    if (ed.key == Key_X && !ed.modifiers) {
-        sl->undoChangeProperty(Pid::SLUR_DIRECTION, PropertyValue::fromValue<DirectionV>(sl->up() ? DirectionV::DOWN : DirectionV::UP));
-        sl->layout();
-        return true;
-    }
     if (ed.key == Key_Home && !ed.modifiers) {
         ups(ed.curGrip).off = PointF();
         sl->layout();
+        triggerLayout();
         return true;
     }
 
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -116,10 +116,6 @@ void TieSegment::draw(mu::draw::Painter* painter) const
 
 bool TieSegment::isEditAllowed(EditData& ed) const
 {
-    if (ed.key == Key_X && !ed.modifiers) {
-        return true;
-    }
-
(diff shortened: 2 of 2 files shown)


src/engraving/libmscore/tie.cpp
```diff
@@ -117,7 +117,7 @@ void TieSegment::draw(mu::draw::Painter* painter) const
 bool TieSegment::isEditAllowed(EditData& ed) const
 {
     if (ed.key == Key_X && !ed.modifiers) {
-        return true;
+        return false;
```
wizofaus, line 120: actually I have no idea why this code is here at all, 'x' already maps to the 'flip' action which works as expected when a tie is selected.

cbjeukendrup: If it works when returning false, yes, then let's remove the whole special case. In that case, please also remove the corresponding code in `TieSegment::edit`, `SlurSegment::isEditAllowed` and `SlurSegment::edit`. 

So now I finally understand exactly what's happening: both "Slur" and "Tie" have this strange special case for flipping using the X key, probably it was necessary at some point in history, but "Slur" correctly uses `undoChangeProperty` but "Tie" uses `setSlurDirection` which skips the undo stack and also fails to update the layout. So that's why it does work for Slur but not for Tie!

src/engraving/libmscore/slur.cpp
```diff
@@ -149,10 +149,6 @@ static ChordRest* searchCR(Segment* segment, track_idx_t startTrack, track_idx_t
 
 bool SlurSegment::isEditAllowed(EditData& ed) const
 {
-    if (ed.key == Key_X && !ed.modifiers) {
-        return true;
-    }
-
```
wizofaus, line 155: none of the special handling for Key_X was necessary

src/engraving/libmscore/slur.cpp
```diff
@@ -188,14 +184,10 @@ bool SlurSegment::edit(EditData& ed)
...
 
-    if (ed.key == Key_X && !ed.modifiers) {
-        sl->undoChangeProperty(Pid::SLUR_DIRECTION, PropertyValue::fromValue<DirectionV>(sl->up() ? DirectionV::DOWN : DirectionV::UP));
-        sl->layout();
-        return true;
-    }
     if (ed.key == Key_Home && !ed.modifiers) {
         ups(ed.curGrip).off = PointF();
         sl->layout();
+        triggerLayout();
```
wizofaus, line 190: existing problem in v3.6 that using "home" to reset grip position didn't always trigger a redraw

oktophonie (approved):

RomanPudashkin (approved):

PR #13320, opened 2022-09-13
fix #13318 - ensure default dash/gap length for hairpins is sensible

### LINKED ISSUE #13318: Hairpin dash settings broken - causing invisibility
**Describe the bug**
When applying the middle 'dash' setting for hairpins, the default 'dash' and 'gap' settings are '0', so the hairpin becomes invisible. In addition, when changing the settings to something visible and saving it as a style, the style does not apply to other hairpins, which also default to 0, 0.

**See video**

https://user-images.githubusercontent.com/21022311/189772620-49ab4662-300a-44e4-80fe-f581b76f1b8e.mov

**Platform information**
 - MacOS (not tried other platforms yet)

wizofaus:
Resolves: https://github.com/musescore/MuseScore/issues/13318

### DIFF
--- a/src/engraving/libmscore/hairpin.cpp
+++ b/src/engraving/libmscore/hairpin.cpp
@@ -625,9 +625,9 @@ Sid Hairpin::getPropertyStyle(Pid pid) const
     case Pid::LINE_STYLE:
         return isLineType() ? Sid::hairpinLineLineStyle : Sid::hairpinLineStyle;
     case Pid::DASH_LINE_LEN:
-        return isLineType() ? Sid::hairpinLineDashLineLen : Sid::NOSTYLE;
+        return Sid::hairpinLineDashLineLen;
     case Pid::DASH_GAP_LEN:
-        return isLineType() ? Sid::hairpinLineDashGapLen : Sid::NOSTYLE;
+        return Sid::hairpinLineDashGapLen;
     default:
         break;
     }
@@ -643,9 +643,9 @@ Hairpin::Hairpin(Segment* parent)
 {
     initElementStyle(&hairpinStyle);
 
+    resetProperty(Pid::HAIRPIN_TYPE);
     resetProperty(Pid::BEGIN_TEXT_PLACE);
     resetProperty(Pid::CONTINUE_TEXT_PLACE);
-    resetProperty(Pid::HAIRPIN_TYPE);
     resetProperty(Pid::LINE_VISIBLE);
 
     _hairpinCircledTip     = false;


src/engraving/libmscore/hairpin.cpp
```diff
@@ -643,9 +643,9 @@ Hairpin::Hairpin(Segment* parent)
 {
     initElementStyle(&hairpinStyle);
 
+    resetProperty(Pid::HAIRPIN_TYPE);
```
wizofaus, line 646: no actual reason to move this, but originally I had tried putting it before initElementStyle( ) to see if it helped (it didn't).

src/engraving/libmscore/hairpin.cpp
```diff
@@ -625,9 +625,9 @@ Sid Hairpin::getPropertyStyle(Pid pid) const
     case Pid::LINE_STYLE:
         return isLineType() ? Sid::hairpinLineLineStyle : Sid::hairpinLineStyle;
     case Pid::DASH_LINE_LEN:
-        return isLineType() ? Sid::hairpinLineDashLineLen : Sid::NOSTYLE;
+        return Sid::hairpinLineDashLineLen;
```
wizofaus, line 628: can't say I understand why these shouldn't be used for any sort of hairpin - as it is, there doesn't appear to be away of changing these style values for the whole score anyway

Tantacrul:
This seems to do the trick. @cbjeukendrup, @RomanPudashkin, or @Eism,

If you are happy with this, can it be merged?

cbjeukendrup:
There's actually a lot of ways to solve this:
1. (what this PR does) use the same style setting for dash/gap length of hairpins and crescendo lines (disadvantage: to users, it's totally unclear that hairpins and crescendos will be affected by the same setting, even though they are visually totally different elements)
2. add a new (unexposed?) style setting for dash/gap length of hairpins
3. use no style setting but a hard-coded default for dash/gap length of hairpins
4. make sure that `EngravingObject::propertyFlags` returns `NOSTYLE` when `getPropertyStyle` returns `Sid::NOSTYLE`, so that the existing fallback hardcoded default in `SLine::propertyDefault` is used (disadvantage: `EngravingObject::propertyFlags` might be performance critical so we may need to keep it as simple as possible)
5. improve the NOSTYLE check in `SLine::propertyDefault`

I don't have a strong preference for which one is the most correct solution. @oktophonie Maybe you can advise?

Tantacrul:
> (what this PR does) use the same style setting for dash/gap length of hairpins and crescendo lines (disadvantage: to users, it's totally unclear that hairpins and crescendos will be affected by the same setting, even though they are visually totally different elements)

@cbjeukendrup: I don't understand this point. I can't figure out how hairpins and slurs are linked from a user's POV. A change to one does not appear to affect the other. Saving one as a default style doesn't affect the other, etc.

I'll now defer to mr. @oktophonie

wizofaus:
The change just means that if you use the "save as default style for this score" option for dash/gap length for a "cresc. . . " or "dim. . ." line, it will also be used as the default style for the dash/gap length for regular (< >) hairpins.

cbjeukendrup:
@Tantacrul FWIW, the issues has nothing to do with slurs, it is about _hairpins_ and _cresc..._ / _decresc..._ text lines.

For some reason, these two things are implemented in the code as one type of element. Still, some aspects of their appearances can be customized separately for hairpins and for cresc lines, using distinct style settings. 

(These settings are not exposed in the Format > Style dialog, but the way to interact with them is in the inspector, by using the "Save as default style for this score" option.)

Now, the problem is that not all required style settings exist:
|Appearance aspect|How hairpins handle it|How cresc lines handle it|
|-|-|-|
|Line style (i.e. continuous, dashed, dotted)|using dedicated style setting for hairpins|using dedicated style setting for cresc lines|
|Dash length (i.e. in case of a dashed line, the length of the dashes)|**??**|using dedicated style setting for cresc lines|
|Gap length (i.e. in case of a dashed line, the space between the dashes)|**??**|using dedicated style setting for cresc lines|

The question is simply what to do with the **??**. Options:
1. (what this PR does) reuse the style setting for cresc lines, so that changing the default for cresc lines will affect hairpins too and vice versa (I don't think that's desirable, because users who really care about the fine details may absolutely want different defaults for cresc lines and hairpins because they are simply totally different elements)
2. Create the missing style settings for dash/gap length of hairpins
3. For dash/gap length of hairpins, don't use customizable style settings but use hardcoded defaults so that users cannot change the default (not ideal for users who care about it, since they'll have to customize each dashed hairpin manually)

4 and 5 from my previous comment are a bit more technical but they ultimately mean something like "making sure that the hardcoded fallback defaults for all text lines gets used instead of invalid values" (but that has the same disadvantage for users as option 3).

wizofaus:
I only mentioned slurs originally because the screen capture included one demonstrating that dashed line functionality was working fine in that case (and as a user it seems odd I can configure gap/dash length for hairpins and other lines but not slurs). Ultimately as long as hairpins support the dashed style without becoming invisible I'd think 99% of users will be fine with it.

Tantacrul:
If they're already implemented in code as one element (meaning @wizofaus didn't introduce this), then it's something we should solve later. In that case, I think we should merge this issue.

As @wizofaus says - it'll solve the fundamental issue.

PR #13345, opened 2022-09-14
[MU4] a few necessary beam adjustments

asattely:
Beams were being unnecessarily shortened, when the outside beam that slants away from the staff is in floater position
image
That's been fixed, and now the primary beam can be 0.5sp above the staff as it slants away.

Also, max-lengthened stems in beams weren't extending past the middle line properly in the case of 3+ beams
image
They now can extend past the middle line:
image


oktophonie (dismissed):

cbjeukendrup (approved):

cbjeukendrup:
Some unit tests need updating...

PR #13347, opened 2022-09-14
fixed #12846: Menus do not read their names (Linux)

### LINKED ISSUE #12846: [MU4 Issue] Accessibility: menus do not read their names (Windows/Linux)
**Describe the bug**
Upon pressing Alt and navigating the menu bar, the names of the menus (File, Edit, etc) are not read by screen reader

**To Reproduce**
1. Start screen reader (tested with NVDA and Orca)
2. Press Alt to focus on the menu bar
3. Use left and right cursor keys to navigate the menus

**Result**
Nothing is read at all

**Expected behavior**
The name of each menu should be read (File, Edit, etc)

**Platform information**
 - OS: Windows, Linux

**Additional context**
In MuseScore 3, Alt alone did not focus on the menu bar; you need Alt+F or other direct mnemonic to open a specific menu.  I think it's an improvement that Alt focuses on the menu bar without needing know any specific menmonic.  But, it's not very useful if the names of the menus don't read!

On Windows, it appears that Tab may also move focus to the menu (after the bottom toolbar), but still, nothing is read aloud.  On Linux, Tab goes directly from the bottom toolbar to the Home/Score/Publish toolbar.  That's actually more what I expect, I think.

Eism:
Resolves: #12846

### DIFF
--- a/src/appshell/qml/platform/AppMenuBar.qml
+++ b/src/appshell/qml/platform/AppMenuBar.qml
@@ -63,6 +63,14 @@ ListView {
         }
     }
 
+    AccessibleItem {
+        id: panelAccessibleInfo
+
+        visualItem: root
+        role: MUAccessible.Panel
+        name: qsTrc("appshell", "Application menu")
+    }
+
     Component.onCompleted: {
         appMenuModel.load()
     }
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #12846 on LinuxMint 20.2 - **FIXED**

PR #13354, opened 2022-09-14
Note dot alignment correction

mike-spa:
Chords in different voices should align their augmentation dots individually when they can. This means that `dotPosX()` is now a member of the `Chord` class and not the `Segment` class.


oktophonie (approved):

src/engraving/libmscore/segment.h
```diff
@@ -216,9 +216,6 @@ class Segment final : public EngravingItem
     bool hasElements(track_idx_t minTrack, track_idx_t maxTrack) const;
     bool allElementsInvisible() const;
 
-    double dotPosX(staff_idx_t staffIdx) const { return _dotPosX[staffIdx]; }
-    void setDotPosX(staff_idx_t staffIdx, double val) { _dotPosX[staffIdx] = val; }
```
cbjeukendrup, line 220: It looks like the `_dotPosX` member of `Segment` can be removed now. It seems that its values are never read anymore.

src/engraving/layout/layoutchords.cpp
```diff
@@ -876,6 +878,7 @@ void LayoutChords::layoutChords3(const MStyle& style, std::vector<Note*>& notes,
         }
 
         Chord* chord = note->chord();
+        chords.insert(chord);
```
cbjeukendrup, line 881: Although this implementation is pretty elegant, I wonder whether it would be more efficient to collect chords at the side of the caller (where we must be iterating over chords anyway to collect the notes) and pass them here. I'm not sure whether that's really better and I guess you've already considered that option, so I'm just curious why you chose this solution.

mike-spa: Nothing very deep, I just thought it would be a bit clumsy to be passing to this function both the vector of notes _and_ the vector of the chords containing them, cause anyway one always implies the other. Though you're almost certainly right that it would be more efficient, I'm gonna test it

src/engraving/libmscore/segment.h
mike-spa: Yup, forgot that. Done now!

src/engraving/layout/layoutchords.cpp
```diff
@@ -113,7 +115,7 @@ void LayoutChords::layoutChords1(Score* score, Segment* segment, staff_idx_t sta
                     hasGraceBefore = true;
                 }
                 layoutChords2(c->notes(), c->up());               // layout grace note noteheads
-                layoutChords3(score->style(), c->notes(), staff, 0);              // layout grace note chords
+                layoutChords3(score->style(), chords, c->notes(), staff);              // layout grace note chords
```
cbjeukendrup, line 118: Shouldn't this particular one be `{ c }` instead of `chords`? At least that would be necessary if you want to match the behaviour that you get when collecting chords inside `layoutChords3`.

mike-spa: Right, though I think it can't be list-initialized like that when passing by reference. The even better thing in that case would be to pass nothing. Or an empty vector

mike-spa: No you're right, cause grace notes may be dottes as well (though I've never seen one). I'll push the fix

cbjeukendrup: For const references, you _can_ pass a "temporary value", like a list initializer, only for non-const references or pointers you'll need a variable.

Since you don't need to modify the vector inside `layoutChords3` I recommend changing the signature of `layoutChords3` to take a const reference. Only use non-const references when you _really_ need to.

cbjeukendrup (approved):

PR #13360, opened 2022-09-15
Fix tuplet brackets end points

mike-spa:
The layout is currently a bit messy, the endpoints of the brackets terminate asymmetrically, and the related style settings don't make much sense. This PR just makes sure that the endpoints are computed properly and that the styles do what they actually say.
Before:
image

After:
image


src/engraving/libmscore/tuplet.cpp
```diff
@@ -315,25 +315,31 @@ void Tuplet::layout()
 
     double xx1 = p1.x();   // use to center the number on the beam
 
+    double leftNoteEdge = 0.0; // system coordinates
```
cbjeukendrup, line 318: > `// system coordinates`

Doesn't `abbox` return page coordinates?

cbjeukendrup (approved):
Looks good to me! (apart from a nitpick about a comment)

src/engraving/libmscore/tuplet.cpp
mike-spa: 👍

oktophonie (approved):

PR #13368, opened 2022-09-15
save_to_cloud_flow

RomanPudashkin:


src/project/internal/projectactionscontroller.cpp
```diff
@@ -375,10 +375,14 @@ bool ProjectActionsController::saveProject(const io::path_t& path)
         return false;
     }
 
-    // TODO(save-to-cloud):
-    // - check whether the score is *currently* public (might have been changed on .com)
-    //   (if yes, show warning)
-    // - save indeed to the cloud, generating MP3 as necessary
+    if (project->isCloudProject()) {
+        RetVal<CloudProjectInfo> info = saveProjectScenario()->askCloudLocation(project);
```
cbjeukendrup, line 379: I've become very reluctant to comment because apparently my understanding isn't 100% correct either, but I doubt about this: I thought that when the user just presses Ctrl+S in a score that has already been saved to the cloud, the score should just be saved without asking a new location (that should only happen when using "Save as...", or when saving a score for the first time). @bkunda @Tantacrul please check if you agree about this :)

src/project/internal/projectconfiguration.cpp
```diff
@@ -263,14 +265,19 @@ async::Channel<io::path_t> ProjectConfiguration::userProjectsPathChanged() const
     return m_userScoresPathChanged;
 }
 
-io::path_t ProjectConfiguration::cloudProjectsPath() const
+io::path_t ProjectConfiguration::cloudProjectPath(const io::path_t& projectName) const
 {
-    return globalConfiguration()->userAppDataPath() + "/cloud_projects";
+    return cloudProjectsPath() + '/' + projectName + '.' + engraving::MSCZ;
```
cbjeukendrup, line 270: Are we sure that `projectName` contains no scary characters like `'/'`? I was personally considering the option to use the score id that's also used on MuseScore.com.

src/project/projecttypes.h
```diff
@@ -83,9 +83,7 @@ enum class CloudProjectVisibility {
 };
 
 struct CloudProjectInfo {
-    QUrl sourceUrl;
```
cbjeukendrup, line 86: I'll quickly tell what was my idea with this:
I thought we would have to save this information inside the mscz file (for example, in a json file like we do for audio settings etc), because we will need the source URL when opening the cloud project (to download the latest version of the project before opening, since it might have been modified and uploaded from another computer). 
Also, if we would use my idea of using the score id as the filename in the `cloud_projects` folder, we will need to read the display name from the project file, to show it correctly in "Recent Scores".
For these reasons, I thought it would be good to store those things in a separate json file in the mscz so that we can quickly retrieve them without having to read the whole score.

src/project/internal/projectactionscontroller.cpp
Tantacrul: Yes, @cbjeukendrup is right. If it is **already** saved **Privately** to the cloud, then CTRL+S should just save it with no dialogs. (Same as pressing CTRL+S on a local file).

However, if the score has visibility set to **Public**, and presses CTRL+S, it's different because it's live and requires new audio to be uploaded. In that case, they need to re-upload new audio and there is a specific dialog informing of this first. (You can see it in the design). 

Just in case anyone is interested, we're putting in this friction for saving **Public** scores to encourage people not to save unfinished work privately.

cbjeukendrup: Okay, that's clear. I'd like to add two things (although they are a bit obvious):
- between saves, the user might have changed the visibility of the score using the score manager on MuseScore.com. So, when the user presses Ctrl+S, we should first download the latest score metadata to see if the score is _actually_ public, and act according to that, rather than to whatever information we have cached.
- To save time and energy, we don't need to do anything when the user presses Ctrl+S in a score where all changes have already been saved/uploaded successfully (i.e. score not "dirty"). In other words, do nothing when nothing needs to be done.

RomanPudashkin: I misunderstood this part of the design. Thank you for correcting me

src/project/projecttypes.h
RomanPudashkin: I have reverted this change for now, but I doubt that storing this information in a separate file would be useful. Especially considering the fact that we still have to download this information from musescore.com...

src/project/internal/projectconfiguration.cpp
RomanPudashkin: > I was personally considering the option to use the score id

I don't like the idea to use scoreId because it becomes available at a pretty late point in time. A ton of things can go wrong while uploading the score to musescore.com. It is much safer to save the project before sending it so that if anything happens the user does not lose their data. But you are absolutely right that using a filename also has its drawbacks. Let's keep things as they are for now, I'll think about this more closely a bit later

cbjeukendrup: Yes, the fact that we don't know the id yet for new cloud scores is one of the problems for which I couldn't think of a solution yet. 
Of course, we can simply encode the filename by replacing special characters. We have `io::escapeFileName`, but I'm not sure how "waterproof" this method is. Another issue with this method is that if we don't watch out, the user will see the escaped version of the name in "Recent Scores", which is not ideal.
Maybe this is a comical idea, but alternatively we can base64 encode the name and use the result as the filename, and decode it before showing it in Recent Scores. QByteArray has builtin methods for this. 

A disadvantage of basing the filename on the score name is that it might case issues when the user renames the score on MuseScore.com. Of course, we need to store the source URL / score id at least _somewhere_, so we can still find out that the score has been renamed and that we'll need to rename it in our `cloud_projects` folder too, but it might be slightly more complicated. 

Anyway, I agree, let's keep it as it is for now, and give it some thought.

vpereverzev (approved):

PR #13376, opened 2022-09-15
[MU4] Allow tremolo stem gap to reduce when trem is very small

asattely:
The issue:
image
Traditional tremolos had no mitigation for becoming extremely small (and in some cases negatively long). This PR caps it at 0.6sp until the gaps become less than 0.4sp:
image


oktophonie (approved):

cbjeukendrup (approved):

cbjeukendrup:
Actually it would be nice to have a VTest that covers this situation, for example what you show on the image in the PR description. Could you please add that?

cbjeukendrup:
@asattely @oktophonie I noticed something about the slope of tremolos. If you draw a line along the tremolo, that line will intersect the stem, instead of touching its endpoint:
Scherm­afbeelding 2022-09-19 om 17 44 57
(left is before this PR, right is after)

It seems not to be caused by this PR, but only becomes more apparent. Is this correct?

oktophonie (approved):

PR #13384, opened 2022-09-16
first_uploaded_project

RomanPudashkin:


src/project/internal/projectactionscontroller.cpp
```diff
@@ -657,6 +657,15 @@ void ProjectActionsController::onProjectSuccessfullyUploaded(const QUrl& urlToOp
         return;
     }
 
+    QUrl scoreMangerUrl = configuration()->scoreManagerUrl();
+
+    if (configuration()->showDetailedProjectUploadedDialog()) {
+        interactive()->open(QString("musescore://project/uploaded?scoresManagerUrl=%1")
+                            .arg(scoreMangerUrl.toString()).toStdString());
```
cbjeukendrup, line 664: Isn't it safer / more performant if we do simply
```c++
UriQuery query("musescore://project/uploaded");
query.addParam("scoresManagerUrl", Val(scoreMangerUrl.toString());
interactive()->open(query);
```
to avoid parsing the URI (which might even go wrong if scoreMangerUrl ever contains a '?' character)?

(also, there's a typo in `scoreMangerUrl`)

RomanPudashkin: Yes, it looks much better, thanks!

Eism (approved):

Issue #12847, opened 2022-08-13
[MU4 Issue] - printing is truncated on page 1


drbirkenstocks:
**Describe the bug**
When printing directly from MU4, page 1 is truncated; subsequent pages do not appear affected.

**To Reproduce**
1. From MU4, Ctrl(Cmd)+P to print
2. Print document
3. Page 1 is truncated

**Expected behavior**
Page 1 will print in its entirety as expected.

**Screenshots**
See below.

**Platform information**
 - Windows 11
![MU4 print](https://user-images.githubusercontent.com/111152765/184507296-3062d5ac-feb2-4eff-a192-817a0474eb76.jpeg)

sammik:
May be related with #12773 (as output looks similar)

DmitryArefiev:
**Can reproduce** on my side with **physical printer** and **Microsoft Print To PDF** (Windows10)

oktophonie:
I have also encountered this (Win 11 + Artix Linux)

DmitryArefiev:
Tested in **latest master** on Win10, Mac12 - **FIXED**

Issue #12773, opened 2022-08-10
[MU4 Issue] PNG Export - images with higher resolution than 360DPI cropps score


sammik:
**Describe the bug**
Scores exported to PNG images with higher resolution, than "default" 360 dpi are cropped (canvas size is OK, but score is cropped)

**To Reproduce**
1. Open score
2. Export - PNG
3. Set high resolution (for example 600 dpi,oranything higher than 360)
4. See resulting image - score is cropped

**Expected behavior**
Full score should be there

**Screenshots**
Exported score:
![Untitled score-1](https://user-images.githubusercontent.com/1646034/184037533-af7a0609-ad64-4e87-8bdc-206377d6c527.png)

Should be:
![Untitled score-1-screenshot](https://user-images.githubusercontent.com/1646034/184037563-aa1cef55-5e18-4858-b922-27bf5df6d32f.png)

**Platform information**
 - OS: [reported on Linux]

**Additional context**
In MU3, it worked well.

PR #13392, opened 2022-09-16
Printing: Fix clipping on first page / Painter: update matrix when setting window or viewport

cbjeukendrup:
Resolves: #12773
Resolves: #12847

### DIFF
--- a/src/framework/draw/painter.cpp
+++ b/src/framework/draw/painter.cpp
@@ -49,10 +49,10 @@ Painter::Painter(QPaintDevice* dp, const std::string& name)
     init();
 }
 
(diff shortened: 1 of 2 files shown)


src/framework/draw/painter.cpp
```diff
@@ -68,6 +68,7 @@ void Painter::init()
     State st;
     st.worldTransform = m_provider->transform();
     st.isWxF = true;
+    st.transform = st.worldTransform;
```
cbjeukendrup, line 71: I'm not 200% convinced about this, but I think it's necessary to keep everything in a consistent state. Not based on practice though, only on theory..

aswadshiraz (approved):

igorkorsukov (approved):

PR #13394, opened 2022-09-16
Add (and fix) tests for draw::Painter

cbjeukendrup:
Includes: #13392


RomanPudashkin (approved):

DmitryArefiev:
Hi @cbjeukendrup , what should be tested here? 
Thanks!

DmitryArefiev:
> @DmitryArefiev A few things you could verify:
> 
>     * the main notation view is still drawn at the correct scale (100% should correspond with physical size of sheet of paper, or at least it should not be worse than before this PR)
> 
>     * the example in Staff/Part Properties > Advanced style properties should still be rendered at the correct size
> 
>     * the palette cell elements are still drawn at correct size.
> 
> 
> Thanks!

Tested those things on Win10, Mac12, LinuxMint 20.2 - **PASS**

PR #13406, opened 2022-09-17
Fix #13342: Multiple tempo markings get added

### LINKED ISSUE #13342: [MU4 Issue] Multiple stacked tempo marking added to top staff
**Describe the bug**
If you select a range across multiple staves and add a tempo marking, it adds multiple stacked tempo markings to the top staff.

**To Reproduce**
1. Create or open a multi-staff score such as String Quartet
2. Use Ctrl+A to select the whole score (or click on first measure of top staff then shift+click on a measure in a lower staff)
3. Add any tempo marking from the palette
4. The tempo marking is shown repeated, stacked vertically, one for each staff selected, but all above the top staff

**Expected behavior**
Just a single tempo marking should be added, the same as if only a single staff had been selected.

**Screenshots**
image

**Platform information**
Windows 10

**Additional context**
Debug build from master branch as of 14-Sep-2022.

HemantAntony:
Resolves: #13342 

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -1711,8 +1711,12 @@ bool NotationInteraction::applyPaletteElement(mu::engraving::EngravingItem* elem
 
             // A text should only be added at the start of the selection
             // There shouldn't be a text at each element
-            for (staff_idx_t staff = firstStaffIndex; staff < lastStaffIndex; staff++) {
-                applyDropPaletteElement(score, firstSegment->firstElement(staff), element, modifiers);
+            if (element->systemFlag()) {
+                applyDropPaletteElement(score, firstSegment->firstElement(0), element, modifiers);
+            } else {
+                for (staff_idx_t staff = firstStaffIndex; staff < lastStaffIndex; staff++) {
+                    applyDropPaletteElement(score, firstSegment->firstElement(staff), element, modifiers);
+                }
             }
         } else {
             track_idx_t track1 = sel.staffStart() * mu::engraving::VOICES;


src/notation/internal/notationinteraction.cpp
```diff
@@ -1711,8 +1711,12 @@ bool NotationInteraction::applyPaletteElement(mu::engraving::EngravingItem* elem
 
             // A text should only be added at the start of the selection
             // There shouldn't be a text at each element
-            for (staff_idx_t staff = firstStaffIndex; staff < lastStaffIndex; staff++) {
-                applyDropPaletteElement(score, firstSegment->firstElement(staff), element, modifiers);
+            if (element->isTempoText()) {
```
cbjeukendrup, line 1714: This fixes the problem for tempo text only, but it's problematic for any system-related item, for example System Text or Rehearsal Marks. Better to use `element->systemFlag()` instead.

src/notation/internal/notationinteraction.cpp
```diff
@@ -1711,8 +1711,12 @@ bool NotationInteraction::applyPaletteElement(mu::engraving::EngravingItem* elem
...
+                applyDropPaletteElement(score, firstSegment->firstElement(firstStaffIndex), element, modifiers);
```
cbjeukendrup, line 1715: I think it should be applied to staff idx 0, at least that's what's done for system text _lines_, so I guess it applies to just system text too. (Maybe it has no functional implications, but at least it's tidier.)

cbjeukendrup (approved):

oktophonie (approved):

DmitryArefiev:
Tested #13342 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #13410, opened 2022-09-17
Fix crash when pasting or Ctrl+Shift+dragging certain elements (beam, note flag a.k.a. `Hook`)

cbjeukendrup:
These elements are not "creatable" by name, so also not "pasteable". But the code to handle such uncreatable elements had some small mistakes which causes a crash.

Resolves: #14159


cbjeukendrup:
@RomanPudashkin It would be great to merge this, since there is a recent report of a similar crash that is also fixed by this PR. 

(I already updated the PR description to mention that newly reported crash too.)

RomanPudashkin (approved):

PR #13414, opened 2022-09-19
Fix #13402: Slur collision correction

### LINKED ISSUE #13402: [MU4 Issue] Poor rendition of flipped slurs
**Describe the bug**

Flipped ties in MS4 are being poorly rendered in MS4 compared to MS3. e.g.

Fipped slur in MS3
slur_ms3

Same flipped slur in MS4
slur_ms4

**To Reproduce**

Open the attached score in both MS3 and MS4. The problem slurs are in measures 1 and 3.
[slur_rendition_issue.zip](https://github.com/musescore/MuseScore/files/9591624/slur_rendition_issue.zip)

**Expected behavior**

Slur rendition should be as good as in MS3.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-3072092433, revision: github-musescore-musescore-007bc62

mike-spa:
Resolves: #13402 
Better specification of which items to ignore when computing collisions. Also added vtest.


src/engraving/libmscore/slur.cpp
```diff
@@ -36,6 +36,7 @@
 #include "mscoreview.h"
 #include "navigate.h"
 #include "part.h"
+#include "rest.h"
```
cbjeukendrup, line 39: It looks like this #include is not used, is it?

mike-spa: yup, forgot to remove that

cbjeukendrup (approved):

PR #13423, opened 2022-09-19
[MU4] Compute snapped beam adjustments irrespective of page position

asattely:
The issue:
image

With snapping turned on, beams would snap to weird different places. This was due to the fact that it was applying snapping to the anchor's page position, which can change if there are margins or titles (or other staves) that push it down the page. Snapping is now done first and then page position is added.

There's a vtest which I hope will show the difference


oktophonie:
Looks good!
Small/resized staves seem to cause a problem however. If I have this (normal size stave, first beam is default position but second has its height set to 2 left and 4 right):

image

If I make this a small stave:

image

The default position is good, the customised one is not.

oktophonie (approved):

cbjeukendrup:
This will mess up scores where custom beam position snapping is disabled (which isn't likely to happen since there's no UI to do that, but still). We should add `pagePos().y()` in all cases.

src/engraving/libmscore/beam.cpp
```diff
@@ -1439,13 +1439,15 @@ void Beam::layout2(const std::vector<ChordRest*>& chordRests, SpannerSegmentType
...
-        double startY = fragments[frag]->py1[fragmentIndex] + pagePos().y();
-        double endY = fragments[frag]->py2[fragmentIndex] + pagePos().y();
+        double startY = fragments[frag]->py1[fragmentIndex];
+        double endY = fragments[frag]->py2[fragmentIndex];
         if (score()->styleB(Sid::snapCustomBeamsToGrid)) {
             const double quarterSpace = spatium() / 4;
-            startY = round(startY / quarterSpace) * quarterSpace;
-            endY = round(endY / quarterSpace) * quarterSpace;
+            startY = (round(startY / quarterSpace) * quarterSpace);
+            endY = (round(endY / quarterSpace) * quarterSpace);
```
cbjeukendrup, line 1447: Now these parentheses here aren't necessary anymore... anyway, looks good to me now!

PR #13427, opened 2022-09-20
[MU4] Missed case on slur-tie adjustment

asattely:
The issue:
https://user-images.githubusercontent.com/89263931/191334637-d7be94a8-1886-4186-891e-0b98dcf3264b.mp4

Notice how the slur adjusts for a tie that starts on the same note, but is not at the vertical position to be adjusted for. This has now been fixed.


oktophonie (approved):

cbjeukendrup (commented):
Looks good to me, over to @vpereverzev!

RomanPudashkin (approved):

PR #13433, opened 2022-09-21
[MU4] Make trems act like beams re: slur attach points

asattely:
The issue:
image

I've taken this opportunity to factor out some of the beam-specific code into more generic beam/trem methods. Eventually the plan is to refactor the whole thing to apply to tremolos. But for now, slur attach points for trems works as beams now.

I reckon this will probably need a vtest, so I'll do that tomorrow morning if necessary.


asattely:
Also I am aware that I am having uncrustify issues, i am looking at that now

asattely:
Welp. My git game is utterly out of control right now. Fixing this soon.

oktophonie (approved):

PR #13447, opened 2022-09-22
[MU4] Implemented detailed progress status for audio export

Eism:


jessjwilliamson:
Works great :) 
The UI needs improvement to make it consistent with the app. This needs to be the same as any other dialogs we have:
- Rounded corners
- Stroke
- Drop shadow

Tested this on light mode and the dialog was blending onto the score paper behind.

RomanPudashkin (approved):

jessjwilliamson:
Here's design based on what I said above. ALSO there is no cancel button, which is a must - if user accidentally exports then regrets, there's no way to cancel.

image

Tantacrul:
@jessjwilliamson - we decided not to implement a cancel button, so the header and button need to go.

However, the round corners and drop shadow (especially dropshadow) should stay

cbjeukendrup:
> the round corners and drop shadow (especially dropshadow) should stay

This would be achieved by not making the window "frameless" but letting it have its platform-native title bar / header (like what's shown in the design).
There must be better ways to prevent closing the window than hiding its frame. For example, `Qt::WindowCloseButtonHint`, and/or not accepting close events might work well.

Tantacrul:
If it's not a completely trivial fix, I suggest we do it after the MS4 launch. Makes much more sense to focus on the highest priority issues instead.

PR #13450, opened 2022-09-22
Solidify the 1/4sp constraint for certain beams

asattely:
Due to a few missed cases, the 1/4sp slant has not been treated as a constraint, in certain situations involving a lot of other beam adjustment to avoid wedges and slivers etc.
image

This has been fixed up:
image


asattely:
The vtests revealed that now that the special 1sp slants are working, there was actually a bit of a bug regarding entirely cross-staff beams being set to up or down based on their cross staff direction, overriding manual direction. That's been fixed as well:
image

oktophonie (approved):

cbjeukendrup (commented):
Looks good to me, over to @RomanPudashkin / @vpereverzev!

PR #13097, opened 2022-08-31
[MU4] Fix #10122: Instrument name does not update in parts dialog and its tab on replace

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10122

PR #13459, opened 2022-09-23
assert_fix

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/pull/13097#discussion_r978542335


cbjeukendrup (approved):
Looks good, the only possible disadvantage I can see is that it is now different in `MasterNotation::setupNewScore` and `NotationProject::loadTemplate` (in the first case, the master notation takes care of locking, in the other case it's done in the project). Not sure if that matters.

PR #13468, opened 2022-09-23
Fix #13461 - treble instruments have only treble clefs

### LINKED ISSUE #13461: [MU4 Issue] Add staff > Bass clef added to Treble clef instruments in scores with already added staves
**Describe the bug**
The issue was mostly fixed in #12353, but there is still a one case when Bass clef added to Treble clef instruments when using Add staff command

**To Reproduce**
1. Create score with Violin 
2. Add staff to Violin

3. Save and reopen (or open [17.zip](https://github.com/musescore/MuseScore/files/9633781/17.zip) )
4. Add staff to Violin again - **added staff has a Bass clef instead of Treble clef**

https://user-images.githubusercontent.com/90187801/191965718-efe4c907-7cbc-4c58-b3e8-310c271b021c.mp4

**Expected behavior**
Treble clef should be added to Treble clef instruments
Bass clef should be added to Bass clef instruments

**Platform information**
 macOS, Windows, Linux

**Additional context**
#12353

sammik:
Resolves: #13461 
Resolves: #13610

There is no reason for checking staff index (as it is not relevant information for detecting instrument)

In second commit, I changed Treble Clef template, to have "second staf bass clef" information, as it is, in fact piano instrument.

### DIFF
--- a/share/templates/01-General/01-Treble_Clef/01-Treble_Clef.mscx
+++ b/share/templates/01-General/01-Treble_Clef/01-Treble_Clef.mscx
@@ -82,6 +82,7 @@
         <minPitchA>21</minPitchA>
         <maxPitchA>108</maxPitchA>
         <instrumentId>keyboard.piano</instrumentId>
+        <clef staff="2">F</clef>
         <Articulation>
           <velocity>100</velocity>
           <gateTime>95</gateTime>
--- a/src/engraving/libmscore/instrument.cpp
+++ b/src/engraving/libmscore/instrument.cpp
@@ -1533,7 +1533,7 @@ ClefTypeList Instrument::clefType(size_t staffIdx) const
 {
     if (staffIdx >= _clefType.size()) {
         if (_clefType.empty()) {
-            return ClefTypeList(staffIdx == 1 ? ClefType::G : ClefType::F);
+            return ClefTypeList();
         }
         return _clefType[0];
     }


cbjeukendrup:
> // TODO check for instrument - second staff of Piano (only) should have Bas Clef

@sammik Since a new staff is always added at the bottom currently, I think we can apply a fairly simple heuristic here: just use the clef of the lowest existing staff. As far as I can tell, you would need to replace `return _clefType[0];` with `return _clefType.back()` to achieve that. What do you think about this idea?

sammik:
> > // TODO check for instrument - second staff of Piano (only) should have Bas Clef
> 
> @sammik Since a new staff is always added at the bottom currently, I think we can apply a fairly simple heuristic here: just use the clef of the lowest existing staff. As far as I can tell, you would need to replace `return _clefType[0];` with `return _clefType.back()` to achieve that. What do you think about this idea?

@cbjeukendrup I dont think, it helps. In fact, if there is single staff treble clef piano score (treble clef template, ...) and user adds staff, he expect bass clef for second staff (that was reason /IMHO/ why staffidx check was there)

So I rewert it and, add instrument family check, and now, it should work as expected.

Anyway, I have no idea, why also piano single staff bass clef template has `_clefType.empty()`.

cbjeukendrup:
I think that checking for the family isn't a great solution... at least, keyboards are not the only instruments with treble and bass clef, harp has also both...

Maybe it would be nice to try to retrieve the `InstrumentTemplate` from the `Instrument` (maybe by instrument id?), and if we can indeed find the template, look at the clefs for the template, and use those to extrapolate if the instrument's clef list is empty or the requested index is too high.

Also, I'm not sure if all this logic should happen in `Instrument::clefType(…)`. It looks like this method was designed to be a simple getter, and UI-related logic like guessing the desired clef for a newly added staff should rather happen in a higher level.

sammik:
> Also, I'm not sure if all this logic should happen in `Instrument::clefType(…)`. It looks like this method was designed to be a simple getter, and UI-related logic like guessing the desired clef for a newly added staff should rather happen in a higher level.

Agree with simple getter. I changed PR to be: 
- If Cleftype is empty, add default (which is treble). (Now it is also consistent with logic in `setClefType`)
- If staffIdx is higher then Cleftype size, add first. (This one I didnt change)


Apart of it, I added missing bass clef to Bass clef template.

And if we want (do we?) to Treble clef temlate have next staff bass clef, it can be done inside template by `<clef staff="2">F</clef>`, that is, what Piano instrument have by default.

Btw, looking into templates, I found out, template title "Bass Clef", "Treble Clef", ... is ignored and replaced by "Untitled Score". Is this intended?

cbjeukendrup:
@sammik Just to be sure we're on the same page, I was not talking about project templates, but about the `InstrumentTemplate` class. It basically contains all info about how an instrument will be created when adding it to the score. When adding an instrument, an `Instrument` is created from an `InstrumentTemplate`. The `Instrument` in the score may then be modified (for example, staves are added or removed). 

I meant that maybe we can somehow figure out from which `InstrumentTemplate` the `Instrument` was created, and then use that `InstrumentTemplate` to make a guess about the clef for the added staff. This would be useful for piano and harp: if the score somehow contains a piano or harp with only one staff (like what happens in the "Treble clef" score template), we can use the `InstrumentTemplate` to find out that the instrument should have had multiple staves and that the clef for the second staff should be a bass clef.

sammik:
> @sammik Just to be sure we're on the same page, I was not talking about project templates, but about the `InstrumentTemplate` class. 

Yes, I understood You.

But normally, it will never happen, that harp, or piano with only one staff has no clef info. 

If you create harp score and remove one staff, clef info remains there. If you create single staff score, violin for exapmple, and change instrument to harp, clef info is automatically added and saved as `<clef staff="2">F</clef>`

So it seems, we dont need this `InstrumentTemplate` check at all, do we? 

`_clefType.empty()` normaly happens only when opening score containing instrument with treble clef. 
Or somehow corrupted file (like Bas Clef template was).

cbjeukendrup (approved):
This looks good to me, at least for now, and it fixes the issue. 

In the future we might want to come up with some heuristic for guessing the desired clef, and we might also want to restructure the code around adding staves a little bit; currently there is some logic in `StaffControlTreeItem::appendNewItem()` which I think should be in (Master)NotationParts. I'll log this as a task for 4.x

Issue #13460, opened 2022-09-23
[MU4 Task] Replace "callback received" page in save-to-cloud flow


bkunda:
**Task description**
Currently, when the user attempts to save a score to the cloud, if they are not already logged into their MuseScore.com account within the desktop app, their browser is triggered and this message is displayed: 

<img width="358" alt="Screenshot 2022-09-23 at 2 51 50 pm" src="https://user-images.githubusercontent.com/86290556/191964476-aba77e8d-06dc-4768-9c28-587ced90e164.png">

At this stage in the user flow, it would be preferable if a page on musescore.com with this design were triggered instead: 

<img width="1348" alt="Screenshot 2022-09-23 at 2 56 23 pm" src="https://user-images.githubusercontent.com/86290556/191965067-d6f51fe3-b0d4-4e0c-8460-ef6fff4cd1f4.png">

Note: The "Take me there" button is not crucial if there are technical limitations that prevent it being implemented. 

Let me know if you have any questions.
Thanks!

shoogle:
The "callback received" page is generated by Qt. @cbjeukendrup, do you know if it's possible for us to redirect to a real webpage instead?

PR #13472, opened 2022-09-23
Implement custom `QAbstractOAuthReplyHandler` subclass to redirect to fancy MuseScore.com "Success!" page

cbjeukendrup:
Inspired by [`QOAuthHttpServerReplyHandler`](https://codebrowser.dev/qt5/qtnetworkauth/src/oauth/qoauthhttpserverreplyhandler.cpp.html). (We could decide to make the implementation even simpler since we don't use everything of it.)

Resolves: #13460

TODO: currently, this will redirect to the MuseScore.com homepage, for demo purposes. The MuseScore.com team needs to implement the real page, and give us the URL, so that we can redirect to that page instead.

### DIFF
--- a/src/cloud/internal/cloudconfiguration.cpp
+++ b/src/cloud/internal/cloudconfiguration.cpp
@@ -106,6 +106,12 @@ QUrl CloudConfiguration::signUpUrl() const
     return QUrl("https://musescore.com/oauth/authorize-new");
 }
 
+QUrl CloudConfiguration::signInSuccessUrl() const
+{
+    // TODO
+    return QUrl("https://musescore.com");
+}
+
 QUrl CloudConfiguration::scoreManagerUrl() const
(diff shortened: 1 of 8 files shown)


src/cloud/internal/cloudconfiguration.cpp
```diff
@@ -106,6 +106,12 @@ QUrl CloudConfiguration::signUpUrl() const
     return QUrl("https://musescore.com/oauth/authorize-new");
 }
 
+QUrl CloudConfiguration::signInSuccessUrl() const
+{
+    // TODO
+    return QUrl("https://musescore.com");
```
shoogle, line 112: FYI, the homepage URL redirects to https://musescore.com/dashboard for logged-in users.

shoogle (approved):

Issue #11960, opened 2022-06-09
[MU4 Issue] Bracket width is calculated incorrectly when determining left margins


oktophonie:
**Describe the bug**
Width of brackets is not calculated correctly when determining left margins, which should be consistent through the score (otherwise implemented in #9277).

![image](https://user-images.githubusercontent.com/20604032/172862330-d4c95123-4a9e-41c9-8778-5af0ab2ede36.png)

Saith @cbjeukendrup: "in this case the brackets start pushing the instrument names off the page, instead of pushing the systems to the right [...] it becomes apparent that the code for calculating the "total" brackets width (i.e. if all staves were visible) does not work at all: it just returns the brackets width for the current system, so the result is that each system has its own left margin so the systems end up not being aligned consistently. "

PR #13474, opened 2022-09-24
Fix left side alignment of systems with brackets

cbjeukendrup:
Resolves: #11960


oktophonie (approved):

src/engraving/libmscore/system.cpp
```diff
@@ -344,26 +343,80 @@ double System::layoutBrackets(const LayoutContext& ctx)
...
+    size_t columns = 0;
+    for (const Staff* staff : ctx.score()->staves()) {
+        for (auto bi : staff->brackets()) {
+            columns = std::max(columns, bi->column() + 1);
+        }
+    }
 
-    score()->setStyleValue(Sid::hideEmptyStaves, hideEmptyStaves);
-    return offset;
+#if (!defined (_MSCVER) && !defined (_MSC_VER))
```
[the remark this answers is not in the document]
cbjeukendrup: Ok, replaced with just std::vector.

It's all over the place though in libmscore. How does VLA actually work and does it have any potential advantages over std::vector? The accompanying comment seemed to suggest that VLAs are not on the heap but on the stack, but is that really the case even though the size of the VLA is unknown in advance? And does stack vs heap still matter much on modern computers?

cbjeukendrup:
I made the `totalBracketOffset` method another tiny bit smarter, because now it will exclude brackets that would be invisible for every system because one or more of their staves are hidden for the whole score. To put it simpler (but less correct), it now excludes brackets for always-invisible staves.

src/engraving/layout/layoutcontext.h
```diff
@@ -69,6 +69,8 @@ class LayoutContext
     Fraction startTick;
     Fraction endTick;
 
+    double totalBracketsWidth = -1.0;
```
RomanPudashkin, line 72: Why isn't it 0 by default?

RomanPudashkin (approved):

src/engraving/layout/layoutcontext.h
cbjeukendrup: We need to detect whether it had already been calculated or not. Even after it has been calculated, it may be 0, so we need to choose another distinguishable initial value.

PR #13478, opened 2022-09-24
Added workaround to issue template

HemantAntony:
This PR added a workaround section to issue templates


HemantAntony:
Should I be adding the `If applicable, add a workaround` as a comment to prevent the extra work of deleting it

RomanPudashkin:
I don't think that changing the template is a good idea. People will almost never fill this new field, so we will see this useless "If applicable, add a workaround" in every new issue. It is better to use the comments section for adding a workaround (or add it to one of the existing fields)

PR #13480, opened 2022-09-24
fix #13239 Add removing excerpts inside MasterScore

### LINKED ISSUE #13239: [MU4 Issue] Parts are not deleted after deleting those instruments from score
**To Reproduce**
1. Create score with some instruments 
2. Open all parts
3. Close all parts
4. Delete some instruments from score
5. Add new instruments to score
6. Open parts - **deleted instruments in step 4  are still there**

https://user-images.githubusercontent.com/90187801/189067301-3d4d9a99-b785-4175-b7ac-53b1d3255578.mp4

**Expected result**
Conditions: Wind Quartet score (Flute, Oboe, Clarinet, Bassoon)

If **Flute part contains only Flute** instrument inside, removing Flute from score closes and **deletes Flute part**
If **Flute part contains Flute+Oboe+Calrinet** instruments inside, removing Flute from score **deletes ONLY Flute instrument from Flute part**

**Platform information**
macOS, Windows, Linux

donarturo11:
Resolves: #13239 

Add remove excerpts inside MasterScore

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -2559,6 +2559,15 @@ void Score::removePart(Part* part)
             break;
         }
     }
+    for (auto excerpt : masterScore()->excerpts()) {
+        if (!excerpt) {
+            continue;
+        }
+        if (excerpt->containsPart(part)) {
+            masterScore()->removeExcerpt(excerpt);
+            break;
+        }
+    }
 
     masterScore()->rebuildMidiMapping();
     setInstrumentsChanged(true);


src/engraving/libmscore/score.cpp
```diff
@@ -2559,6 +2559,11 @@ void Score::removePart(Part* part)
             break;
         }
     }
+    for (auto excerpt : masterScore()->excerpts()) {
+        if (excerpt) {
+            masterScore()->removeExcerpt(excerpt);
```
RomanPudashkin, line 2564: This code just removes all excerpts when removing any part, it doesn't fix the problem

donarturo11: I tested with debugger, that ```_excerpt``` after create parts is too empty, so ```if(_excerpt)``` condition is always false and never is executed. I checked in the same place, that adding parts adds axcerpts to MasterScore(). I can search another solution.

donarturo11: @RomanPudashkin I put new commit. Is it better? I would ask if searching in this place is a good idea or better search in another place?

donarturo11: @RomanPudashkin Yesterday I improved this code. Now removes only excerpts, which contain part, not all. I think, that now is ready to review.

RomanPudashkin: No, this solution also won't work. Now, this code will remove all excerpts containing this part, which is not the desired behavior. I don't recommend fixing this bug right now because it is quite complex and requires a deep understanding of how the system works

PR #13483, opened 2022-09-25
Fix #13454: Fingering & Sticking should be added to all notes

### LINKED ISSUE #13454: [MU4 Issue] Regression: fingering and sticking not added to all selected notes
**Describe the bug**
If you select a range of notes and apply a fingering element (including any custom text you define yourself from the palette), it is not applied to all notes, but instead only to the first note

**To Reproduce**
1. Select a range of notes
2. Click on a fingering element from the palette
3. See error: the fingering is only added to the first note

**Expected behavior**
Fingering should be added to every element in the selection

**Screenshots**

**Platform information**
 - OS: Linux

**Additional context**
This worked correctly in MU3, but broke with a change to affect the behavior of other text such as dynamic or tempo marking.  For dynamics and tempo, it definitely makes sense to only apply them to just the first element, since by their nature, they *affect* the subsequent notes too.  Arguably true for staff text as well.  But fingering is more like articulation in that it *should* be applied to every note in the selection.

Same for sticking.  There is no predefined sticking palette, but if a user creates one, they would definitely expect it to be applied to each note in a selection just as articulations are.  Similarly if one defines a custom fingering or sticking text using a particular special character or otherwise implementing a special annotation that you might reasonably want to apply to all notes of a selection.

Fix is simple - where this special handling for text was introduced a few months ago, it should exclude fingering and sticking:

https://github.com/musescore/MuseScore/blob/master/src/notation/internal/notationinteraction.cpp#L1707

Workaround is to make a list selection instead of a range selection.

HemantAntony:
Resolves: #13454

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -1826,7 +1826,7 @@ bool NotationInteraction::applyPaletteElement(mu::engraving::EngravingItem* elem
                 spanner->styleChanged();
                 score->cmdAddSpanner(spanner, i, startSegment, endSegment);
             }
-        } else if (element->isTextBase()) {
+        } else if (element->isTextBase() && !element->isFingering() && !element->isSticking()) {
             mu::engraving::Segment* firstSegment = sel.startSegment();
             staff_idx_t firstStaffIndex = sel.staffStart();
             staff_idx_t lastStaffIndex = sel.staffEnd();


cbjeukendrup (approved):

DmitryArefiev:
Tested #13454 on Win10, Mac12 - **FIXED**    

Thanks for fixing!

PR #13488, opened 2022-09-26
Fix #269926 MusicXML export: don't duplicate fermatas near grace notes

shoogle:
Previously, if a note had both a fermata and a grace note then the grace note would gain a fermata during export.

Resolves: https://musescore.org/en/node/269926

Replaces #7621


RomanPudashkin (approved):

DmitryArefiev:
Tested https://musescore.org/en/node/269926 on Win10, Mac12 - **FIXED**

PR #13492, opened 2022-09-26
[MU4] Fix #12140: Tempo marking ignored when they follow a rit, accel or other gradual tempo change

### LINKED ISSUE #12140: [MU4 Issue] Tempo marking ignored when they follow a rit, accel or other gradual tempo change
A tempo marking placed just after a gradual tempo change is ignored in playback, but moving the marking any further allows it to be read.

image

1. Add a rit, accel, allargando,  or rall (for a whole bar or multiple full bars)
2. Add a tempo marking on the next downbeat.
3. Play back
4. Tempo stays what it was set to by the rit, etc.
5. Move the tempo marking even half a beat later, and it is not ignored.

**Expected behavior**
The tempo should jump to the latest marking. I would also expect to be able to add "a tempo" to achieve a similar functionality in my use case, which is fairly common.

**Platform information**
 - OS: Windows 10

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/12140

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -575,7 +575,7 @@ void Score::setUpTempoMap()
         BeatsPerSecond newBps = currentBps * tempoChange->tempoChangeFactor();
 
         std::map<int, double> tempoCurve = TConv::easingValueCurve(tempoChange->ticks().ticks(),
-                                                                   4,
+                                                                   4 /*stepsCount*/,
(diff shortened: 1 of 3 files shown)


vpereverzev (approved):

DmitryArefiev:
Tested #12140 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

PR #13496, opened 2022-09-26
Style default tweaks

mike-spa:
No functional changes, just tweaked a bunch of style defaults following @oktophonie 's specifications.


oktophonie (approved):

alexpavlov96 (approved):

cbjeukendrup (approved):

cbjeukendrup:
There are more .mss files across the repo that might need updating though. For example, if I search for `<lyricsMinBottomDistance>2</lyricsMinBottomDistance>` in the whole repo, I get 119 results, not only those 41 templates. Shouldn't the others be updated too?

mike-spa:
@cbjeukendrup If I search for the style key in .mss files in the repo I only find the legacy default specifications, which shouldn't be changed, and the utests, which I also don't think should be changed (and they're mostly playback tests anyway).

image

PR #13502, opened 2022-09-27
Fix #10228, #303351: Range selections not spoken by screen readers

### LINKED ISSUE #10228: [MU4 Issue] [Accessibility] Range Selections are not spoken by screen readers
**Describe the bug**
When a range of notes is selected, the screen reader should say "Range Selection" followed by the start and end location in terms of bars and beats similar to that displayed in the Status Bar at the bottom of the screen.

image

Here the screen reader should say _"Range selection: starts staff 1, measure 3, beat 3; ends staff 3, measure 5, beat 1.75."_

**To Reproduce**
1. Open an existing score or enter notes in a new score.
2. Start the screen reader (Narrator, NVDA, VoiceOver, Orca, etc.)
3. Select a range of notes (e.g. hold Shift and press Up, Down, Left or Right arrow keys) and listen to the screen reader.
4. Continue extending the range selection and listening to the screen reader.

**Expected behavior**
The screen reader should say the contents of the status bar: _"Range selection: starts... ends..."_. Instead it just says the name of the last note in the range selection.

 - OS: All (tested on Windows and macOS)

**Additional context**
There is code to do this in [src/notation/internal/notationaccessibility.cpp](https://github.com/musescore/MuseScore/blob/c816fb01719e81575579e132b150237f0f8685db/src/notation/internal/notationaccessibility.cpp#L102-L115) but it appears to be unused at the moment, and I'm not sure how it integrates with the element accessibility code in [src/engraving/accessibility](https://github.com/musescore/MuseScore/tree/master/src/engraving/accessibility).

HemantAntony:
Resolves: #10228 


src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -187,11 +189,25 @@ QString AccessibleItem::accessibleName() const
...
+    Selection selection = m_element->score()->selection();
+    QString selectionText = "";
+
+    if (selection.isRange() && selection.endSegment()) {
+        Segment* prevSeg = selection.endSegment()->prev1(SegmentType::ChordRest);
+        if (m_element->tick() != selection.tickStart() || m_element->tick() != prevSeg->tick()) {
+            if (m_element->tick() == selection.tickStart()) {
+                selectionText = "Selection begins";
+            } else if (prevSeg && m_element->tick() == prevSeg->tick()) {
+                selectionText = "Selection ends";
```
cbjeukendrup, line 201: Not translatable?

HemantAntony: Its just spoken out loud. So I assumed there was no need for it

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -187,11 +189,25 @@ QString AccessibleItem::accessibleName() const
...
+            }
+        }
+    }
 
     barsAndBeats.remove(u';'); // Too many pauses in speech
 
-    QString name = QString("%1%2%3%4%5")
+    QString name = QString("%1%2%3%4%5%6")
                    .arg(!commandInfo.isEmpty() ? (commandInfo + "; ") : "")
+                   .arg(selectionText)
```
cbjeukendrup, line 210: I think you need to do this:
```suggestion
                   .arg(!selectionText.isEmpty() ? (selectionText + "; ") : "")
```
otherwise it will be glued together with the next word

src/engraving/accessibility/accessibleitem.cpp
shoogle: It should be translated like the rest of the UI, otherwise the button names would be in one language and the selection messages in a different one. The screen reader will just read whatever it is told to read, so it's up to us to translate it.

HemantAntony: Ah okay. I didn't know that

src/engraving/libmscore/select.cpp
```diff
@@ -698,8 +698,44 @@ void Selection::setRangeTicks(const Fraction& tick1, const Fraction& tick2, staf
...
+                i++;
+            }
+            el.emplace_back(elements);
+        }
+
+        for (auto it = el.rbegin(); it != el.rend(); it++) {
+            for (EngravingItem* e : *it) {
+                e->setSelected(true);
+            }
+        }
```
shoogle, line 734: We really need some comments in this function to explain what you are doing and why it is necessary.

1. It's really hard to see what you are doing differently to before. If it wasn't for the `r` in `rbegin()` and `rend()` then your 38 lines of new code would be exactly equivalent to the 2 lines that were already there.

2. Even once I spotted the `r` and realised that you were calling `e->setSelected(true)` on elements in a slightly different order, it's not obvious why the order matters.

You're also relying on the `_el` list being in a certain initial ordering. It works most of the time, but it breaks when there's an articulation on the final note in the range selection.

I think there is a much simpler way to solve it that doesn't rely on `_el` being ordered in a certain way:
```C++
void Selection::update()
{
    for (EngravingItem* e : _el) {
        e->setSelected(true); // also tells accessibility that e has focus
    }
    // Only one element can have focus at a time, so currently the final
    // element in _el has focus. That's ok for a LIST selection because it
    // corresponds to the last element the user clicked on.
    if (ChordRest* cr = activeCR()) {
        // User is performing a RANGE selection. Let's focus a note/rest in the activeCR.
        EngravingItem* e = cr->isRest() ? toEngravingItem(cr) : toChord(cr)->upNote();
        assert(e->selected()); // was selected in loop above (e is somewhere in _el)
        e->setSelected(true); // HACK: select it again so accessibility thinks it has focus
    }
    updateState();
}
```

shoogle (changes requested):
@HemantAntony, thanks for this! You implemented the specification exactly as I requested, but now having tried it out I think that it is a bit too strange to hear "selection ends" before "selection starts". I'd like to switch to something a bit closer to what we had in MuseScore 3, but still here the name of the element that changes as you extend the selection.

I tweaked your code to experiment with a few different forms of output. I think this is best:

**Accessible name:** _"Note C3 half, measure 5 beat 1, selected"_

**Accessible description:** _"Range selection starts measure 3 beat 3 staff 3 (piano); ends measure 5 beat 1 staff 1 (flute)"_

Since I've already written the code I might as well share it. You can find it [here](https://github.com/musescore/MuseScore/compare/master...shoogle:MuseScore:303351-range_selection_note_not_read) and add it to the PR (just drop/amend your current commit).

The only thing I didn't do was add the "staff X (instrument Y)" bit to the output so feel free to add this yourself. It should be easy enough. See [AccessibleRoot::updateStaffInfo()](https://github.com/musescore/MuseScore/blob/a78c97285cc127fb151a6966b0dec65fcbcce29c/src/engraving/accessibility/accessibleroot.cpp#L121) for a guide. When the start and end staff are the same neither should be included in the output (so in that case it would just be measure and beat numbers like it is at the moment).

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -206,7 +208,41 @@ QString AccessibleItem::accessibleDescription() const
...
     }
 
-    return readable(m_element->accessibleExtraInfo());
+    Selection selection = m_element->score()->selection();
+
+    if (!selection.isRange()) {
+        return readable(m_element->accessibleExtraInfo());
+    }
+
+    String start = selection.startSegment()->formatBarsAndBeats();
```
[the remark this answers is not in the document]
HemantAntony: I also added checks for `endSegment`, `staff1` and `staff2`

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -206,7 +208,43 @@ QString AccessibleItem::accessibleDescription() const
         return QString();
     }
 
-    return readable(m_element->accessibleExtraInfo());
+    Selection selection = m_element->score()->selection();
```
Eism, line 211: I would recommend moving the code below this into AccessibleRoot. In accessibleName() use accessibleRoot too

```
class AccessibleRoot
{
    bool isRangeSelection() const;
    QString rangeSelectionInfo();
};

AccessibleRoot* root = accessibleRoot();
if (root.isRangeSelection()) {
    return readable(root.rangeSelectionInfo());
}
    
return readable(m_element->accessibleExtraInfo());
```

shoogle: Looks good to me! @HemantAntony, please move the range selection code to `AccessibleRoot`.

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -206,7 +208,43 @@ QString AccessibleItem::accessibleDescription() const
...
+    }
+
+    String start = selection.startSegment()->formatBarsAndBeats();
+    String end;
+    if (Segment* endSeg = selection.endSegment()->prev1(SegmentType::ChordRest)) {
+        end = endSeg->formatBarsAndBeats();
+    }
+
+    start.remove(u';'); // Too many pauses in speech
+    end.remove(u';');
```
shoogle, line 224: This `remove()` can go inside the if statement above so it only happens when there is some end text.

src/engraving/accessibility/accessibleitem.cpp
```diff
@@ -206,7 +208,43 @@ QString AccessibleItem::accessibleDescription() const
         return QString();
     }
 
-    return readable(m_element->accessibleExtraInfo());
+    Selection selection = m_element->score()->selection();
+
+    if (!selection.isRange() || !selection.startSegment() || !selection.endSegment()) {
```
shoogle, line 213: Since you use `startSegment()` and `endSegment()` later on, it's best to store them in a variable rather than calling them repeatedly.

src/engraving/accessibility/accessibleroot.cpp
```diff
@@ -170,3 +171,64 @@ void AccessibleRoot::setMapToScreenFunc(const AccessibleMapToScreenFunc& func)
...
+bool AccessibleRoot::isRangeSelection() const
+{
+    if (auto focusedItem = m_focusedElement.lock()) {
+        const EngravingItem* item = focusedItem->element();
+        if (item) {
+            return item->score()->selection().isRange();
+        }
+    }
+    return false;
+}
```
shoogle, line 184: I think you can just do:
```C++
bool AccessibleRoot::isRangeSelection() const
{
    return element()->score()->selection().isRange();
}
```

src/engraving/accessibility/accessibleroot.cpp
```diff
@@ -170,3 +171,64 @@ void AccessibleRoot::setMapToScreenFunc(const AccessibleMapToScreenFunc& func)
...
+QString AccessibleRoot::rangeSelectionInfo()
+{
+    auto focusedItem = m_focusedElement.lock();
+    if (!focusedItem) {
+        return QString();
+    }
+    const EngravingItem* item = focusedItem->element();
+    if (!item) {
+        return QString();
+    }
```
shoogle, line 195: Instead of using `m_focusedElement` I think this is sufficient:
```
Score* score = element()->score();

// use later for score->selection() and score->staff()
```

shoogle (changes requested):

src/engraving/accessibility/accessibleroot.cpp
```diff
@@ -170,3 +171,64 @@ void AccessibleRoot::setMapToScreenFunc(const AccessibleMapToScreenFunc& func)
...
+    String start = startSegment->formatBarsAndBeats();
+    String end;
+    if (Segment* endSeg = endSegment->prev1(SegmentType::ChordRest)) {
+        end = endSeg->formatBarsAndBeats();
+        end.remove(u';'); // Too many pauses in speech
+    }
+
+    start.remove(u';');
```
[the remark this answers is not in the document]
shoogle: Actually, you won't need the `if (Segment* endSeg...)` statement now if you make the change I suggest [here](https://github.com/musescore/MuseScore/pull/13502/files#r983334314), so you can put the two `removes` back together again with just one comment.

src/engraving/accessibility/accessibleroot.cpp
```diff
@@ -170,3 +171,64 @@ void AccessibleRoot::setMapToScreenFunc(const AccessibleMapToScreenFunc& func)
...
+
+    QString staffInstrument1, staffInstrument2;
+    staff_idx_t startStaff = selection.staffStart();
+    staff_idx_t endStaff = selection.staffEnd() - 1;
+
+    if (startStaff != endStaff) {
+        Staff* staff1 = m_element->score()->staff(startStaff);
+        Staff* staff2 = m_element->score()->staff(endStaff);
```
shoogle, line 218: Remember to use the new `score` variable here instead of calling `m_element->score()`

shoogle (approved):
Works great, thanks! Tested on macOS.

Eism:
@HemantAntony 
Please fix the cause of the unit tests crash

6: engraving_utests: ../src/engraving/libmscore/engravingobject.h:638: mu::engraving::Chord* mu::engraving::toChord(mu::engraving::EngravingObject*): Assertion `e == 0 || e->isChord()' failed.
 6/13 Test  #6: engraving_utests .................Child aborted***Exception:  30.93 sec
test 7
      Start  7: iex_bb_tests

cbjeukendrup:
I think the following fix would be better:
```c++
EngravingItem* e = cr->isChord() ? toChord(cr)->upNote() : cr;
```
(Only convert to chord if we're _really_ sure that it's a chord)

shoogle (approved):

Issue #13077, opened 2022-08-30
[MU4 Issue] "save as default" for offset fails; always changes default to 0


MarcSabatella:
**Describe the bug**
When using the "save as default style for this score" command in Properties / Appearance / Offset, instead of making the current values the default for the score, the default is instead always changed to 0 horizontal, 0 vertical

**To Reproduce**
1. Empty score
2. Enter a staff text in both of the first two measures
3. In Properties / Appearance, change the X & Y offset to 1 and -3 respectively
4. Click "..." / Save as default style for this score
5. See error: the second staff text actually changes offset to 0, 0

**Expected behavior**
Second staff text should have changed to 1, -3 like the first.

**Screenshots**
![image](https://user-images.githubusercontent.com/1799936/187349505-d175b55d-7050-4564-9a17-f5f50b4b6a01.png)

**Platform information**
 - OS: Linux

**Additional context**
Other "save as default style for this score" commands work as expected, somehow only the offset is messed up.  But, the bug with offset applies not just to staff text but to all element types I tried.

Not sure if it's related, but after this, it also becomes hard to edit the offset via keyboard - typing new values into the spinbox doesn't seem to work as it usually does.  At least temporarily; it does fix itself.  So very likely not related, but I mention it just in case.

MarcSabatella:
@Tantacrul could this get prioritized?

Tantacrul:
Done. Thanks

Tantacrul:
@oktophonie - is this one you're guys can do?

oktophonie:
@Tantacrul We'd prefer not, as it's basically a UI matter. Strikes me as a @cbjeukendrup special, if he has the capacity!

DmitryArefiev:
Fixed in #13521

PR #13521, opened 2022-09-28
Inspector: fix applying offsets to style

cbjeukendrup:
Splitting properties over multiple PropertyItems turns out to cause problems.

Resolves: #13077

### DIFF
--- a/src/inspector/CMakeLists.txt
+++ b/src/inspector/CMakeLists.txt
@@ -58,6 +58,8 @@ set(MODULE_SRC
     ${CMAKE_CURRENT_LIST_DIR}/models/inspectorlistmodel.h
     ${CMAKE_CURRENT_LIST_DIR}/models/propertyitem.cpp
     ${CMAKE_CURRENT_LIST_DIR}/models/propertyitem.h
+    ${CMAKE_CURRENT_LIST_DIR}/models/pointfpropertyitem.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/models/pointfpropertyitem.h
     ${CMAKE_CURRENT_LIST_DIR}/models/general/generalsettingsmodel.cpp
     ${CMAKE_CURRENT_LIST_DIR}/models/general/generalsettingsmodel.h
     ${CMAKE_CURRENT_LIST_DIR}/models/general/appearance/appearancesettingsmodel.cpp
--- a/src/inspector/models/abstractinspectormodel.cpp
+++ b/src/inspector/models/abstractinspectormodel.cpp
@@ -475,6 +475,26 @@ PropertyItem* AbstractInspectorModel::buildPropertyItem(const mu::engraving::Pid
 {
     PropertyItem* newPropertyItem = new PropertyItem(propertyId, this);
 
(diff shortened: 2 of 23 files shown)


RomanPudashkin:
> @RomanPudashkin I see that in commit [cdd36db](https://github.com/musescore/MuseScore/commit/cdd36db14e3faddfd3c5c321b130d3d3f9998c50) you removed the ability so set the horizontal offset because that was "undesired". What's undesired about it?

This question should be addressed to @Tantacrul or @bkunda

RomanPudashkin (approved):

Tantacrul:
@cbjeukendrup - I don't quite recall making any decision about horizontal offset one way or another. Could you possibly describe the problem from the user's POV? What objects are we talking about?

It may be the case that we hid horizontal offset for certain items because (in specific cases), it was a useless option.

cbjeukendrup:
It concerns almost all elements from the Lines palette; specifically it affects what you see in the "Text" tab in Properties.
Scherm­afbeelding 2022-09-29 om 11 32 06
With this PR, you will see a full offset control, before, it was only a vertical offset field. 

It might be that horizontal offset is not applicable in all cases, but in some cases it is absolutely useful. So I propose to expose the setting anyway for now, and later we can build complicated logic to hide it in certain cases where it is less useful.

Tantacrul:
> It might be that horizontal offset is not applicable in all cases, but in some cases it is absolutely useful. So I propose to expose the setting anyway for now, and later we can build complicated logic to hide it in certain cases where it is less useful.

What are the cases where it is useful?

I would be more inclined to not show it unless there is a really strong case.

MarcSabatella:
In MU3, changing the horizontal offset for text on lines was, as far as I can tell, the same functionally as adjusting the left endpoint.  So it was in that sense useless, and I might guess that is why it was removed.

But, in my opinion it *should* have done something better all along - it should have moved the text relative to the line.  So you can have text closer or further from the actual start of the line than the default.  That would be useful for any number of custom line types, including things like using an arrow or some other Unicode character as the text.

DmitryArefiev:
Tested #13077 on Win10, Mac12, LinuxMint 20.2 - **FIXED**

DmitryArefiev:
@cbjeukendrup Should I check other things except described in #13077 ?

cbjeukendrup:
@DmitryArefiev Except for the offset control in Appearance > Offset, the other offset controls are also fixed now:
- Notehead offset in Notehead > Show more
- Stem offset and Flag offset in Stem > Show more
- For text offset, the horizontal offset setting is back, because that setting is necessary to create certain kinds of custom text lines. In the future we might want to tweak the behaviour of this setting, or we might want to hide it under certain circumstances (although I don't see any advantage in that). But for now it is just visible and working as in MU3.

Tantacrul:
I've tested all of these and they seems good to go.

I am a little unhappy with adding back in the horizontal offset option (because it is broadly valueless) but as @MarcSabatella says, we can always make it do something valuable in future.

Let's leave for now and keep on truckin' through the rest of the issues!

PR #13522, opened 2022-09-28
Articulation tweaks

mike-spa:
Three separate jobs:
- A small modification of default styles.
- Correcting the alignment of composite staccato/tenuto marks.
- Improving the efficiency when fetching the category of a given articulation mark (it doesn't makes sense to perform all those comparisons every time we need to query it, makes more sense to compute them once and store them).

Before:
image17055
After:
image18575


src/engraving/libmscore/articulation.cpp
```diff
@@ -676,54 +677,35 @@ double Articulation::mag() const
...
+
+    _isMarcato = _symId == SymId::articMarcatoAbove || _symId == SymId::articMarcatoBelow
+                 || _symId == SymId::articMarcatoStaccatoAbove || _symId == SymId::articMarcatoStaccatoBelow
+                 || _symId == SymId::articMarcatoTenutoAbove || _symId == SymId::articMarcatoTenutoBelow;
+
+    _isLuteFingering = _symId == SymId::articMarcatoStaccatoAbove || _symId == SymId::articMarcatoStaccatoBelow
+                       || _symId == SymId::articTenutoStaccatoAbove || _symId == SymId::articTenutoStaccatoBelow
+                       || _symId == SymId::articAccentStaccatoAbove || _symId == SymId::articAccentStaccatoBelow
+                       || _symId == SymId::articMarcatoTenutoAbove || _symId == SymId::articMarcatoTenutoBelow
+                       || _symId == SymId::articTenutoAccentAbove || _symId == SymId::articTenutoAccentBelow;
```
cbjeukendrup, line 708: Looking at the old definition of `isLuteFingering()`, this doesn't seem right to me...

src/engraving/libmscore/articulation.h
```diff
@@ -122,6 +122,14 @@ class Articulation final : public EngravingItem
     };
     static AnchorGroup anchorGroup(SymId);
 
+    bool _isDouble = false;
+    bool _isTenuto = false;
+    bool _isStaccato = false;
+    bool _isAccent = false;
+    bool _isMarcato = false;
+    bool _isLuteFingering = false;
```
cbjeukendrup, line 130: To reduce memory usage, you _could_ store these using a single field of flags. But I'd first ask @vpereverzev if he thinks that's better.

Anyway, to do that, you would create an enum like this:
```c++
enum class ArticulationCategory: char { // char makes sure that this will use only 1 byte, 
                                        // which consists of 8 bits, 
                                        // which is enough to store these 6 flags
    None = 0x00
    Double = 0x01
    Tenuto = 0x02
    Staccato = 0x04
    Accent = 0x08
    Marcato = 0x10
    LuteFingering = 0x20
}
```
Then create a Flags from that enum: (you need to `#include "flags.h"` for this)
```c++
DECLARE_FLAGS(ArticulationCategories, ArticulationCategory)
DECLARE_OPERATORS_FOR_FLAGS(ArticulationCategories)
```

Then replace all these members with:
```suggestion
    ArticulationCategories m_categories = ArticulationCategory::None;
```

Then replace the `is...` methods with:
```c++
    bool isDouble() const { return m_categories & ArticulationCategory::Double; }
```
etc.

Then modify `computeCategories` like this:
```c++
m_categories.setFlag(ArticulationCategory::Double, 
                     _symId == SymId::articMarcatoStaccatoAbove || _symId == SymId::articMarcatoStaccatoBelow
                     || _symId == SymId::articTenutoStaccatoAbove || …)

File: src/engraving/libmscore/articulation.cpp (hunk 1 above)
Comment (mike-spa) [reply src/engraving/libmscore/articulation.cpp:708]: Shit, wrong copy-paste. Glad you noticed

File: src/engraving/libmscore/articulation.h (hunk 2 above)
Comment (mike-spa) [reply src/engraving/libmscore/articulation.h:130]: Thanks! I'll try!

**Review by oktophonie — APPROVED** [review]

File: src/engraving/libmscore/articulation.h [hunk 3]
```diff
@@ -180,12 +195,12 @@ class Articulation final : public EngravingItem
...
-    bool isStaccato() const;
-    bool isAccent() const;
-    bool isMarcato() const;
-    bool isLuteFingering() const;
+    bool isDouble() const { return m_categories & ArticulationCategory::DOUBLE; }
+    bool isTenuto() const { return m_categories & ArticulationCategory::TENUTO; }
+    bool isStaccato() const { return m_categories & ArticulationCategory::STACCATO; }
+    bool isAccent() const { return m_categories & ArticulationCategory::ACCENT; }
+    bool isMarcato() const { return m_categories & ArticulationCategory::MARCATO; }
+    bool isLuteFingering() { return m_categories & ArticulationCategory::LUTE_FINGERING; }
```
cbjeukendrup, line 203: `const` got lost here :)

PR #13541, opened 2022-09-29
More fixes for translatable strings

cbjeukendrup:
Resolves: https://discord.com/channels/818804595450445834/818852942882275379/1024521115558744115
Resolves: https://discord.com/channels/818804595450445834/818852942882275379/1024918469382254612


src/notation/view/widgets/breaksdialog.cpp
```diff
@@ -37,6 +37,16 @@ BreaksDialog::BreaksDialog(QWidget* parent)
     setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
     intervalButton->setChecked(true);
     intervalBox->setValue(DEFAULT_INTERVAL);
+
+    //: `%1` will be replaced with a number input field.
+    //: Text before it will appear before that number field, text after will appear after the field.
+    QString text = qtrc("notation/add-remove-system-breaks", "Break systems every %1 measures");
+    QRegularExpression regex("(.*)%1(.*)");
```
RomanPudashkin, line 44: Why do we use regexp here?

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
```diff
@@ -113,12 +121,25 @@ RadioButtonGroup {
...
+        RowLayout {
             id: numberOfSavesItem
 
-            spacing: 0
+            width: parent.width
+            spacing: 12
+
+            //: `<span>%n</span>` will be replaced with a number input field.
+            //: Text before it will appear before that number field, text after will appear after the field.
+            readonly property string text: qsTrc("project/save", "Every: <span>%n</span> saves", "", settingsModel.numberOfSaves)
```
RomanPudashkin, line 132: Very confusing piece of code. Why do we need it?

Eism:
@cbjeukendrup 
I would recommend not using discord links. Please describe the problem in the body of the PR.

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
cbjeukendrup: It's maybe indeed a bit complicated, I'll explain.

For the translators, it is very important that the text "Every: %n saves" is one big string, instead of split over multiple strings. This has to do with different word order in different languages, of course.

Thus, what we do, is making it one string, and replacing the "%n" with a number input field. Concretely, this means we need to split the string and put the part before %n before the spinbox, and the part after "%n" after the spinbox.

Recall that a translatable string containing "%n" allows translators to give multiple translations for singular / plural / whatever forms their language has. 
For example:
1 Save
2 Saves
(English is a bad example in this case because it has only two forms and in this particular case the singular form will never occur, but you get the idea. There are languages with way more complex rules, so we really need to support that.)

However, this comes with a small implication: to make it work correctly, we _must_ specify the value of "n" as the fourth argument to `qsTrc`. This is necessary to:
- make sure that `lupdate` correctly finds the string and identifies it as a "%n" string
- make sure that we retrieve the correct form of the translation (singular/plural/...)

But it implies that the "n" is already filled in in the string, so we get "Every: 2 saves" instead of "Every: %n saves". Normally that's great, but in this case it's a tiny bit inconvenient, since it makes it harder to determine where to split the string.

So we do a trick, and surround the number with something we can recognize but won't harm either, namely `<span>` and `</span>`.

Now, we can easily use a regular expression to split the string.

src/notation/view/widgets/breaksdialog.cpp
cbjeukendrup: Actually this one is simple enough to use a simple `.split("%1")`. Will do!

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
cbjeukendrup: (I added some documentation in the code too)

Eism: No. 
This is a difficult task and it is better not to waste your time on it.
Not only your time, but also the time of translators, the time of developers who will read this code and a lot of documentation for it. The code should be simple.

It is better to let the user see 1 saves, I think he will not be offended much.

cbjeukendrup: I honestly don't see what's so complicated about it. It took me literally a few seconds to come up with this idea. And correct grammar _is_ important. In my opinion, you can put as many effort in an app as you want, but if it contains language mistakes, it just looks unprofessional. 

Anyway, opinions aside - if we choose a less perfect solution, with %1 instead of %n, it won't be much less complicated. The only simplification is that we can replace the regular expression with `.split("%1")`. Is that worthwhile?

shoogle: I'm also not totally convinced the plural is worth worrying about in this case, but if we do want to do this then I think the following would be sufficient to explain what is going on to developers and translators:

```QML
//: `<span>%n</span>` will be replaced with a number input field for the user to pick a number.
//: Text before `<span>%n</span>` will appear before the number field, text after will appear after the field.
readonly property string text: qsTrc("project/save", "Every: <span>%n</span> save(s)", "", settingsModel.numberOfSaves)
// We need `<span>` tags to be able to split the string after it has been translated and `%n` replaced with a number.
```

Notice I replaced "saves" with "save(s)" to give a hint about the plural.

Presumably this should also be done with "every %1 measures" above.

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
```diff
@@ -146,23 +177,31 @@ RadioButtonGroup {
...
-                measureUnitsSymbol: qsTrc("project/save", "Saves")
+                measureUnitsSymbol: numberOfSavesItem.textPart2InSpinbox ? numberOfSavesItem.textPart2 : ""
                 step: 1
                 decimals: 0
 
                 navigation.panel: root.navigationPanel
-                navigation.row: index
+                navigation.row: numberOfSavesItem.index
                 navigation.column: 1
+                navigation.accessible.name: qsTrc("project/save", "Every n saves")
```
shoogle, line 192: Maybe write this as `"Every N saves"` in case capitalisation affects pronunciation by screen readers? Also, I was initially confused and thought you meant `"Every %n save(s)"`, so a capital N would help avoid that confusion.

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
RomanPudashkin: >I honestly don't see what's so complicated about it. It took me literally a few seconds to come up with this idea
 
 This doesn't mean that your solution is easy to understand for other developers. It looks incredibly overcomplicated
 
 I like Peter's suggestion. Just change the unit text to "save(s)". But tbh, I doubt very much that we need to do anything with this text at all

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
```diff
@@ -146,23 +167,31 @@ RadioButtonGroup {
...
-                measureUnitsSymbol: qsTrc("project/save", "Saves")
+                measureUnitsSymbol: numberOfSavesItem.textPart2InSpinbox ? numberOfSavesItem.textPart2 : ""
                 step: 1
                 decimals: 0
 
                 navigation.panel: root.navigationPanel
-                navigation.row: index
+                navigation.row: numberOfSavesItem.index
                 navigation.column: 1
+                navigation.accessible.name: qsTrc("project/save", "Every N saves") + ": " + currentValue
```
cbjeukendrup, line 182: @shoogle / @Eism What do you think about the accessibility descriptions? I realized that I should maybe do it slightly different than what I'm doing now:
- the radio button gets "Every N saves" as its accessibility name
- the number input field gets just its value as its name. 

Is that a good idea? Or do you prefer another option?

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
cbjeukendrup: Okay, I resorted to using simply %1, to avoid the scary regular expression. Let's hope nobody will notice the imperfection :)

But FWIW, we _must_ make it one string and splitting it. I'm not doing this for academical reasons, but because translators already reported that they can't work with these separate strings. 

Also, if the two words are in two different strings, translators have no clue that they form one phrase together, or what they actually mean (unless we start adding enormous comments and hints etc.). If you see the whole phrase, it's totally clear and unambiguously translatable. If I were a translator and saw just the word "Saves" for example, I could come up with many possible Dutch translations with totally different meanings and I would have no idea which one is correct.

shoogle: I agree it should be one string. The `<span>` idea was actually quite an elegant solution. I think it was just the large comment above made it look more complicated than it really was.

src/project/qml/MuseScore/Project/AudioGenerationSettings.qml
shoogle: You wouldn't normally make the value part of the name unless you're doing some kind of hack. The screen reader should report the value separately to the name.

For example, you might hear the screen reader say _"Visible checkbox not checked"_. The name of the checkbox is just "Visible". The screen reader knows that it has to tell you the role ("checkbox") and the value ("not checked") so you don't need to include that information in the name.

 You'd probably need to use "Every N saves" as the name for both the radio button and the input field (or the input field could be called something like "Saves before regenerating audio"). The screen reader should announce the value separately to this, but you'd need to test to confirm. There are instructions for how to enable your build-in screen reader [here](https://github.com/musescore/MuseScore/wiki/Accessibility-Testing#popular-screen-readers).

cbjeukendrup: I now made it "Every N saves" for both the radio button and the spin box.

We might want to refine the way screen readers treat our spinboxes; currently, they treat them as plain text fields. This was better in MS3.

cbjeukendrup:
@RomanPudashkin Could we please merge this? It would be beneficial for translation quality.

RomanPudashkin (approved):

PR #13545, opened 2022-09-30
fixed applying style for rehearsal marks

alexpavlov96:
fixed applying style for rehearsal marks


handrok (approved):

abariska:
🔴 **To be closed** 
The fix is reverted due to the unexpected result

PR #13558, opened 2022-09-30
Save to cloud: tweak behaviour of asking about audio upload settings

cbjeukendrup:
- Remove the "don't ask again" checkbox
- Ask the question just once (to change decision, user can go to Preferences > Cloud)
- Ask the question just whenever the user is saving privately to the cloud for the first time, no matter whether that score was newly created or previously stored on disk


RomanPudashkin (approved):

bkunda:
🟢 Tested on MacOS. All looks good! 
Thanks @cbjeukendrup

PR #13563, opened 2022-10-01
Fix #9476: Fix accidental note input bar and palette interaction

### LINKED ISSUE #9476: [MU4 Issue] When an accidental is selected on the score, clicking an accidental in the Note Input Bar does nothing
**Describe the bug**
When an accidental is selected on the score, I would expect intuitively that I can change that accidental by clicking a different accidental in the Note Input Bar, or in the Accidentals Palette. This is not the case; instead, nothing happens. I can only change the accidental by first selecting the note.

iwoithe:
Resolves: #9476

### DIFF
--- a/src/engraving/libmscore/accidental.cpp
+++ b/src/engraving/libmscore/accidental.cpp
@@ -383,6 +383,10 @@ bool Accidental::acceptDrop(EditData& data) const
 {
     EngravingItem* e = data.dropElement;
 
+    if (e->type() == ElementType::ACCIDENTAL) {
+        return true;
+    }
+
     if (e->isActionIcon()) {
         ActionIconType type = toActionIcon(e)->actionType();
         return type == ActionIconType::PARENTHESES
@@ -400,6 +404,9 @@ EngravingItem* Accidental::drop(EditData& data)
 {
     EngravingItem* e = data.dropElement;
     switch (e->type()) {
+    case ElementType::ACCIDENTAL:
+        score()->changeAccidental(note(), toAccidental(e)->accidentalType());
+        break;
     case ElementType::ACTION_ICON:
         switch (toActionIcon(e)->actionType()) {
         case ActionIconType::PARENTHESES:
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -1960,8 +1960,27 @@ void Score::toggleAccidental(AccidentalType at, const EditData& ed)
 
 void Score::changeAccidental(AccidentalType idx)
 {
-    for (Note* note : selection().noteList()) {
-        changeAccidental(note, idx);
+    for (EngravingItem* item : selection().elements()) {
+        Accidental* accidental = 0;
+        Note* note = 0;
+        switch (item->type()) {
+        case ElementType::ACCIDENTAL:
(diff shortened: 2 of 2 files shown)


src/engraving/libmscore/accidental.cpp
```diff
@@ -22,6 +22,10 @@
 
 #include "accidental.h"
 
+#include "editdata.h"
+#include "engravingitem.h"
+#include "engravingobject.h"
+#include "infrastructure/symbolfont.h"
```
cbjeukendrup, line 28: These four #includes are unnecessary/undesired :)

src/engraving/libmscore/cmd.cpp
```diff
@@ -1909,6 +1911,21 @@ void Score::toggleAccidental(AccidentalType at, const EditData& ed)
 
 void Score::changeAccidental(AccidentalType idx)
 {
+    if (selection().isAccentidentalsOnly()) {
```
cbjeukendrup, line 1914: I think we can remove this "if" condition and return statement; we should allow a mixed selection. For example, if my selection is something like in the first measure of the following example, I would like to end up with something like the second measure:
Scherm­afbeelding 2022-12-02 om 23 26 46

src/engraving/libmscore/cmd.cpp
```diff
@@ -27,6 +27,8 @@
 
 #include <assert.h>
 
+#include "engravingitem.h"
+#include "engravingobject.h"
```
cbjeukendrup, line 31: These includes are also not necessary. They happen automatically by the other includes. 

(You're right that in general these "transitive includes" are not always great, but in the case of these particular headers, I think we should allow it.)

cbjeukendrup (commented):
Thanks for this! 

One general point: we prefer if you _rebase_ the PR branch on top of master, instead of _merging_ master into your PR branch. Those merge commits inside PRs make the history difficult to understand...

src/engraving/libmscore/cmd.cpp
iwoithe: Do I then remove the `isAccentidentalsOnly()` method (as it's not used anymore)?

cbjeukendrup: Yes, indeed.

src/engraving/libmscore/accidental.cpp
```diff
@@ -559,6 +563,9 @@ EngravingItem* Accidental::drop(EditData& data)
 {
     EngravingItem* e = data.dropElement;
     switch (e->type()) {
+    case ElementType::ACCIDENTAL:
+        score()->changeAccidental(toAccidental(e)->accidentalType());
```
cbjeukendrup, line 567: I think you need to change this to <code>score()->changeAccidental(<b>note()</b>, toAccidental(e)->accidentalType());</code> to fix the issue you asked about on Discord 🙂

RomanPudashkin (approved):

iwoithe:
@RomanPudashkin Rebased and squashed. :slightly_smiling_face:

zacjansheski:
Tested on Mac OS 11.6.5, Ubuntu 20.04.6, Windows 11
https://github.com/musescore/MuseScore/issues/9476 FIXED

Issue #10894, opened 2022-03-24
[MU4 Issue] Preferences>Appearance>Reset to default option doesn't work


DmitryArefiev:
**To Reproduce**
1. Go to Preferences>Appearance and change some settings
2. Trigger **Reset to default** - **nothing happens**

https://user-images.githubusercontent.com/90187801/159985657-d5705a81-6d9a-41cd-bd10-cce40f3e35f3.mp4

**Expected behavior**
Reset to default should reset all changes to default on Appearance page

**Platform information**
 macOS, Windows, Linux

cbjeukendrup:
The button does reset the theme colors (accent color, etc.) and the "Background" color. But indeed, because it's placed at the bottom of the page, you'd expect that it resets all options on the page.

HemantAntony:
So, the fix would be to change the position of the Reset Button?

cbjeukendrup:
@Tantacrul The button does something, namely it resets any changes made to the current theme (e.g. customized accent color etc.). I think this is especially useful for high contrast themes, because they have many customizable options. I believe that this button was originally designed to do just that. 

But since the button is placed at the bottom of the page, a user may expect that it will reset everything on the whole page. Should we alter the behaviour of the button to do this? Or should we move the button closer to the settings that it affects?

Tantacrul:
Ah, OK. That's really strange. The intention was for it to reset all options in 'Appearance'.

Tantacrul:
To be clear: this button should just reset all settings in 'Appearance'.

If this causes any significant problems, we can just remove the button for now :)

cbjeukendrup:
@Tantacrul I have a proposed fix in #13568, would be great if you could try if that's the desired behaviour!

PR #13568, opened 2022-10-01
Preferences > Appearance: reset button should reset *all* elements on that page

cbjeukendrup:
Resolves: #10894

The button is now a lot more dangerous :) 
(but the "Cancel" button still undoes the appearance reset, so no worries)

(Also contains a tiny fix for the location for wallpapers; see the first commit and its message)


RomanPudashkin (approved):

Tantacrul:
Yes. I think this is a clever implementation. Perhaps a confirmation dialog might make it a little clearer.

For example '**Are you sure you'd like to reset **light, high contrast** to its default settings?**'

But given string freeze - and the fact that I worked it out pretty easily, this probably isn't necessary. Good to merge.

Issue #13573, opened 2022-10-01
[MU4 Issue] [macOS] Uncaught exception when closing MuseScore


cbjeukendrup:
**Describe the bug**
When closing MuseScore, the last line that I see in the Qt Creator output window is:
```
libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```

macOS doesn't show a "MuseScore quit unexpectedly" dialog though, and the crash happens when pretty much everything is already de-initialised, so the severity is probably low.

When you add a breakpoint for "exception thrown", you can see the stack trace. The problematic `std::lock_guard` is the one in `Logger::write`. At the start of the stack trace, the destructor of CoreMidiInPort is called, which calls its `disconnect` method, which contains a LOGD statement.

PR #13574, opened 2022-10-01
Disconnect MIDI ports in deinit, not when global objects are destructed

cbjeukendrup:
Resolves: #13573

### DIFF
--- a/src/framework/midi/internal/platform/lin/alsamidiinport.h
+++ b/src/framework/midi/internal/platform/lin/alsamidiinport.h
@@ -35,9 +35,10 @@ class AlsaMidiInPort : public IMidiInPort, public async::Asyncable
 {
 public:
     AlsaMidiInPort() = default;
-    ~AlsaMidiInPort() override;
+    ~AlsaMidiInPort() = default;
 
     void init();
+    void deinit();
 
     MidiDeviceList availableDevices() const override;
     async::Notification availableDevicesChanged() const override;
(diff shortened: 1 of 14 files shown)


src/framework/midi/internal/platform/lin/alsamidiinport.h
```diff
@@ -35,9 +35,10 @@ class AlsaMidiInPort : public IMidiInPort, public async::Asyncable
 {
 public:
     AlsaMidiInPort() = default;
-    ~AlsaMidiInPort() override;
+    ~AlsaMidiInPort() = default;
```
cbjeukendrup, line 38: I'm not quite sure whether this one is correct. The reason I'm doubting is that we have a member `std::shared_ptr<Alsa> m_alsa;`, and the Alsa struct is defined only in the cpp file. I would expect that `shared_ptr` needs to know the definition of Alsa in its destructor (because it might need to delete the underlying pointer, and therefore it needs to know the size of the Alsa struct). Therefore I'm thinking that the destructor of AlsaMidiInPort etc. needs to be defined in the cpp file, so that the destructor of m_alsa is (implicitly) done in cpp where the side of Alsa is known. But for some reason it still compiles like this, so maybe the destructor of shared_ptr does something else than I'm thinking?

Edit: seems that's the case... apparently it does something with a "deleter", that is created when the shared_ptr is constructed. That explains a lot.

Eism (approved):

Issue #13369, opened 2022-09-15
[MU4 Issue] Tremolo playback issues


bkunda:
**Describe the bug**
Two parts to this issue:
1. "Eighth through stem" tremolo not playing back (except for crotchet notes). "16th through stem" tremolo plays back as "Eighth through stem" should. "32nd through stem" plays back as "16th through stem" should, and so on.2. 
2. Tremolo on triplets plays back with dotted rhythm3. 

See demo video: https://www.dropbox.com/s/ldqrw138v3y07s2/tremolo-issues.mov?dl=0

Test files:
[tremolo-test-files.zip](https://github.com/musescore/MuseScore/files/9576505/tremolo-test-files.zip)

**Expected behavior**
1. Practically speaking, a tremolo stroke attached to a note stem should double the number of notes played at halve the value of the note to which it is attached. 
2. There should be no dotting of the playback rhythm 

**Platform information**
 - MacOS 12.3.1 (M1)

cbjeukendrup:
@bkunda I have a potential fix (still needs to be code-reviewed) for the first issue: #13575. 

The triplet issue turns out to be a more general problem, so I logged a separate issue for it in #13576.
(What's actually going on there, is that the durations of the tremolo notes are not taking into account the fact that they are in a tuplet; so they behave like they stem from a normal eighth note. That's why it sounds like the different notes are overlapping a bit.)

PR #13575, opened 2022-10-02
Fix playback of tremolos on 8th notes or shorter

cbjeukendrup:
Resolves: #13369 (the part about tremolos on eighth notes and shorter; the part about triplets not yet but that turns out to be a more general problem; will log a separate issue later)

To do: add a test case
I have a test case in mind, namely a score with 2 staves, where the upper staff contains notes with tremolos, and the lower staff contains the same information but written with short notes instead of tremolos. Both staves should generate equal playback data. I think the best way to render and compare the two staves would be using PlaybackModel (so the test case would be added in playbackmodel_tests.cpp), but I was not sure.


bkunda:
🟢 Tested on MacOS. This all works as expected.
Thanks @cbjeukendrup!

Issue #13162, opened 2022-09-04
[MU4 Issue]  A plugin's  `requiresScore: false` seems to get ignored


Jojo-Schmitz:
**Describe the bug**
A plugin's  `requiresScore: false` seems to get ignored, in order for such a plugin to become active in the Plugins menu a score needs to be loaded

**To Reproduce**
1. Load and and activate the [Batch Convert](https://musescore.org/en/project/batch-convert) or the [ABC Import](https://musescore.org/en/project/abc-import) plugin.
2. (Re)Start MuseScore, don't open a score

**Expected behavior**
In MuseScore 2 and 3 that setting, `requiresScore: false`, resulted in the plugin menu enty to become active even without a score being loaded

**Platform information**
 - OS: Windows (but no reason tu assume this to be different on macOS or Linux

Jojo-Schmitz:
Example: https://musescore.org/en/project/batch-convert

Eism:
Related to #8859, which we moved to 4.x

Jojo-Schmitz:
I see, thanks. Makes us falling back to pre 2.x times in this respect though...

cbjeukendrup:
@Jojo-Schmitz Could you please try #13582?

Jojo-Schmitz:
Seems to work!

PR #13582, opened 2022-10-02
Respect plugins' `requiresScore: false`

cbjeukendrup:
Resolves: #13162

### DIFF
--- a/src/plugins/internal/pluginsservice.cpp
+++ b/src/plugins/internal/pluginsservice.cpp
@@ -122,6 +122,7 @@ PluginInfoMap PluginsService::readPlugins() const
         info.version = view.version();
         info.enabled = pluginsConfigurationHash.value(info.codeKey).enabled;
         info.categoryCode = view.categoryCode();
+        info.requiresScore = view.requiresScore();
 
         QString thumbnailName = view.thumbnailName();
         if (!thumbnailName.isEmpty()) {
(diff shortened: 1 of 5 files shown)


Jojo-Schmitz (approved):
Resolves the issue

RomanPudashkin:
As far as I can tell, none of our standard plugins use this API. So why do we need to fix it?

>To Reproduce
Load and and activate the batch-convert plugin.

This plugin won't work in 4.0 anyway

Jojo-Schmitz:
Why won't it work?

Jojo-Schmitz:
Hmm, well, it in does, reports "ERROR on reading [file]"

Jojo-Schmitz:
Which means that `readScore(fileFullPath, true)` fails, for siome reason. So it seems there's another bug in the Plugin framework

cbjeukendrup:
That method is indeed not implemented at the moment.

Jojo-Schmitz:
Reported in #13593

jeetee:
> none of our standard plugins use this API. So why do we need to fix it?

Because of existing MS3 plugins using it?

cbjeukendrup:
> So why do we need to fix it?

Also, why would we _not_ fix it? It might make some people a bit happier, with minimal effort 🙂

Jojo-Schmitz:
What's holding this PR from getting merged?

Jojo-Schmitz:
It is not just Batch Convert that would benefit from this, but also ABC Import. The workaround of having to open another unrelated score is just nasty and annoying

cbjeukendrup:
Rebased. @RomanPudashkin Could you please review and merge this? I don't really see any disadvantages, and it would make some people a tiny bit happier, so why not :)

Jojo-Schmitz:
It is the lesser of the 2 issues the ABC plugin has (the other being the missing file I/O API). Still worth having it fixed though

RomanPudashkin (approved):

Issue #11218, opened 2022-04-15
[MU4 Issue] Improving workflow of Panels (Palettes, Properties, Instruments and Selection Filter)


KaiVinter:
**Description:**
On MS3 you could view all the panels at once, even when stacked, so everything was pretty easily accessible. However in MS4, since the 4 main panels are stacked 'on tabs' on the same place, I think it would be a major improvement for the workflow to make them appear in the front whenever they are triggered, specially for people who are more used to use keyboard shortcuts.

I mean:
If a panel is open but not visible (only tabbed) it should become visible whenever it's triggered, instead of having to trigger it twice for closing and then reopening.
If a panel is visible, it should be closed when triggered.

Tantacrul:
Can you clarify what you mean by 'triggered'?

Incidentally, the panel layout in MS4 is more flexible than three. We just started with a different default layout. MS3 also had tabs but not by default.

If you drag one of the tabs somewhere else, you'll notice you can stack them vertically, horizontally or move them to the other side of the screen.

Incidentally, we'll be releasing with extensive documentation, so users won't need to figure this out without a helping hand.

KaiVinter:
With 'triggered' I mean called, [Ex. If I press F9 I 'trigger'/toggle the Palettes panel], probably a translation error from my side.

Sure you can stack them as in MS3, but I want to use tabs. And whenever I toggle/call a panel that is not shown in the front with a shortcut, it closes. And I think it makes sense that if I press a keyboard shortcut for a panel, it opens/shows in the front instead of closing.

Tantacrul:
Oh, I see. That makes sense.

Tantacrul:
OK, so I agree we should match the MS3 functionality:

For the following panels: **Selection filter, Properties, Instruments & Palettes** :

- If the panel is tabbed but hidden, then the shortcuts should display it (make it the selected tab) and not close it
- If the panel is already selected and the user triggers the shortcut, then the panel should be closed

KaiVinter:
Exactly.

In sketch code it would be something like:

'''
when > toggle panel {
     if > closed > then > open & show in the front.
     if > shown in the front > then > close.
     if > open in another tab > then > show in the front.
}
'''

MarcSabatella:
See also https://github.com/musescore/MuseScore/issues/12873 - the palette search command should also either open or "top" the palettes panel.  Probably doesn't make sense for this command to ever close the panel though.

[...]

PR #13583, opened 2022-10-02
Address #11218: Change toggle behavior of vertical panels

zmelson:
Resolves: [Issue #11218](https://github.com/musescore/MuseScore/issues/11218)

Changed DockPageView::setDockOpen() to select vertical panels rather than close them when open but not selected. Added function toggleTabSelection() to DockPanelView class to accomplish this. Changes behavior for both shortcuts and drop-down menus as they make the same function calls.

### DIFF
--- a/src/appshell/view/dockwindow/dockpageview.cpp
+++ b/src/appshell/view/dockwindow/dockpageview.cpp
@@ -27,6 +27,7 @@
 #include "dockcentralview.h"
 #include "dockpanelview.h"
 #include "dockstatusbarview.h"
+#include "appshelltypes.h"
 
 #include "ui/view/navigationcontrol.h"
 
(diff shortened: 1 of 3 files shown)


src/appshell/view/dockwindow/dockpageview.cpp
```diff
@@ -188,7 +189,13 @@ void DockPageView::setDockOpen(const QString& dockName, bool open)
         return;
     }
 
-    if (!open) {
+    bool relevantPanel =
```
RomanPudashkin, line 192: You should not add local solutions to base classes/components. This class knows nothing about specific panels, like Palettes or Properties, and we shouldn't give it any such knowledge

RomanPudashkin:
closing due to a long period of inactivity

src/appshell/view/dockwindow/dockpageview.cpp
hanoixan: I'm new to the project and looking at picking up this issue. 

Instead of local name-based solutions like this, I imagine there should be some per-tab configuration property along the lines of "open makes hidden tab current". Does this pattern exist in MuseScore already? Or, is there a better existing way to data drive this?

PR #13584, opened 2022-10-02
fix #8964: Added checks for when no score is open

### LINKED ISSUE #8964: [MU4 Issue] Palettes/Properties panel fully enabled even when no score open
**Describe the bug**
The Palettes and the Properties panel do nothing if no score is open yet appear to be fully enabled, even though clicking on various controls/elements does nothing.

**To Reproduce**
1. Start application, don't open any score
2. Click on "Score" tab (next to "Home")
3. Observe that in Palettes tab all items/controls etc. are fully enabled.  This does allow you customize the palette, but clicking on any of the elements does nothing.
4. Likewise for the Properties panel - the various toggles for "Invisible"/"Frames" etc. are all clickable and appear to toggle but do nothing. The 'Page settings' and 'Style settings' button do nothing either.

**Expected behavior**
As per MU3, if no score is opened, the inspector should show nothing (or be disabled) and likewise for palettes, or at least, if we are going to enable customizing without a score being open, it should be obvious that the various elements you can normally add to a score are disabled (NB MU3 doesn't allow customizing if no score is open).
 
**Screenshots**

Windows 10

**Additional context**

5yM01:
Resolves: #8964 

Added checks for when no score is open so that the properties cannot be changed or used by the user if the case that no score is open is true. Checks it in scoredisplaysettingsmodel.cpp so that the variables that dictate whether the properties can be used aren't always true even when no score is open.

### DIFF
--- a/src/inspector/models/score/scoredisplaysettingsmodel.cpp
+++ b/src/inspector/models/score/scoredisplaysettingsmodel.cpp
@@ -123,6 +123,9 @@ void ScoreSettingsModel::setShouldShowInvisible(bool shouldShowInvisible)
     if (m_shouldShowInvisible == shouldShowInvisible) {
         return;
     }
+    else if (context()->currentNotation() == nullptr) {
+        return;
+    }
 
     dispatcher()->dispatch("show-invisible");
     updateShouldShowInvisible(shouldShowInvisible);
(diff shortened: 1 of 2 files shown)


RomanPudashkin:
This fix is incorrect. We should disable the panel, not add checks for null
https://github.com/musescore/MuseScore/issues/8964#issuecomment-1079781550

Or even just ignore this issue, because the current behavior does not cause any problems (e.g. crashes)

PR #13587, opened 2022-10-02
RecipeDebian musescore does not use qtwebkit

nmariusp:


shoogle:
Thanks for you contribution! However, I don't think we use RecipeDebian anymore. I think it was only ever used for ARM builds, which we stopped supporting due to the maintenance burden and because Qt doesn't provide ARM binaries for Linux.

Issue #13263, opened 2022-09-09
Importing MS2 files: 'View' options in Inspector can show the opposite of what they are supposed to show


Tantacrul:
**Describe the bug**

This seems to affect files coming from MS2.

Opening [this score](https://www.dropbox.com/s/6xi9qvk6qu3a2rq/Greensleeves_Guitar_Easy.mscz?dl=0) from V2.02, I've noticed that the inspector view options are inverted, meaning when something is hidden in the UI, it becomes visible on the score.

<img width="294" alt="image" src="https://user-images.githubusercontent.com/21022311/189335217-a1da9429-fe23-4531-a32d-4f63116aa829.png">

**To Reproduce**
1. On score and accept new defaults
2. Click to show the option for **formatting** in the inspector
3. Observe this actually hides formatting
4. The same is true for frames

It is possible to get the inspector to 'wake up' and display things properly. I'm not sure how I did it. One thing I did was make an object invisible and then turn on/off the show **invisible** option.

HemantAntony:
@Tantacrul I can't seem to reproduce this in the latest master

sberla365:
I have noticed the same issue with files entirely made in MU4, as the attached. Sometimes when opening the file, if no interaction with the file is made (no clicks on the score or elements of it, the show/hide buttons are inverted (eye closed shows, eye open hides). It does not happen every time though. A simple click on the score (everywhere, whether an element or the page itself) resets to the correct behaviour of the show elements

[Test 2.zip](https://github.com/musescore/MuseScore/files/9640813/Test.2.zip)

sberla365:
I took a short screencast of the behaviour 
[Screencast from 25-09-22 14:12:48.webm](https://user-images.githubusercontent.com/55401477/192145502-f7fe0904-a640-446e-b818-4cea9e5c55da.webm)

abariska:
🟢 **FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

PR #13588, opened 2022-10-02
Fix that view options in Properties panel sometimes work inverted

cbjeukendrup:
Resolves: #13263

Fix updating the data stored in `ScoreSettingsModel`.

Alternative considered: not storing any data in `ScoreSettingsModel`, but retrieving the actual values from the current notation (or default value if no notation is open) every time the getter is called. But that might give worse performance.


RomanPudashkin (approved):

abariska:
Please, merge to master

PR #13594, opened 2022-10-03
Fix a (potential) crash with dereferencing `end()` iterator

cbjeukendrup:
Resolves: a crash reported on Discord with a certain score


src/engraving/libmscore/segment.cpp
```diff
@@ -2792,15 +2792,15 @@ void Segment::computeCrossBeamType(Segment* nextSeg)
...
         return;
     }
 
-    std::sort(springs.begin(), springs.end(), [](Spring& a, Spring& b){ return a.preTension < b.preTension; });
+    std::sort(springs.begin(), springs.end(), [](const Spring& a, const Spring& b) { return a.preTension < b.preTension; });
     double inverseSpringConst = 0.0;
     double force = 0.0;
     double curForce = 0.0;
-    for (auto spring = springs.begin(); spring != springs.end(); ++spring) {
+    for (auto spring = springs.cbegin(); (spring + 1) != springs.cend(); ++spring) {
```
cbjeukendrup, line 2803: @mike-spa A few lines here below, you are reading from `(spring + 1)`, which would be equal to `springs.end()` before this change. That causes a crash (sometimes), because reading from `springs.end()` is invalid.

So I changed it so that the last spring will not be considered, so that there is always a "next" spring. Is that desired behaviour? Or should we consider all springs, and do something special for the last one?

cbjeukendrup: Okay, judging from the VTest failures, this is _not_ the correct way to fix it!

So I really need your help, @mike-spa :)

mike-spa: Urgh, what a silly oversight, sorry 🤦‍♂️ Yeah that doesn't work because the loop must be still allowed to enter the final element of the vector. The solution, if I'm not mistaken, is to check `spring + 1` just before dereferencing it. So the `if` statement a few lines below should become:
```
if (spring + 1 == springs.end() || curForce < (spring + 1)->preTension) {
      break;
}
```
This, in turn, should also allow to remove the exit condition from the `for` statement itself, since we're gonna exit anyway.

mike-spa: Or better yet, you could increase the iterator just before the if, so that can also be removed from the `for` statement.
It could look like this:
```
for (auto spring = springs.begin();;) {
        inverseSpringConst += 1 / spring->springConst;
        width += spring->width;
        curForce = width / inverseSpringConst;
        ++spring;
        if (spring == springs.end() || curForce < spring->preTension) {
            break;
        }
    }
```

mike-spa (approved):

src/engraving/libmscore/segment.cpp
mike-spa: Excellent!

PR #13616, opened 2022-10-04
Full bar rest bug fix

mike-spa:
Full bar rests in multiple voices situations cause extra space at the beginning of the measure and alignment errors for dynamics. See for example: 
image

This fixes it.


oktophonie (approved):

cbjeukendrup (commented):
Looks fine to me! Over to @RomanPudashkin

PR #13630, opened 2022-10-05
gp-import: importing chord names instread of fret diagrams

alexpavlov96:
*importing chord names instread of fret diagrams*


handrok (approved):

abariska:
🟢 **TESTED**

PR #13634, opened 2022-10-05
A few bug fixes and improvements to system header and brackets

mike-spa:
- The distance of staff names to the system should scale not only with the staff size (as it does now) but also with text size (taking into account the possibility that the text size may not "follow staff size").
- Indentation of the first system should not include the brackets width.
- Square brackets should never appear on a single staff, even if "show bracket if spanning a single staff" is enabled. So I've created a dedicated style setting for square brackets (not exposed).
- The begin bar line of a system should not be drawn if the system has only one visible staff. At the moment, it was working correctly only if the score itself has just the one staff.


mike-spa:
@cbjeukendrup I get this weird u-test failur where it seems to crash or something, but I don't understand why it's happening and I can't reproduce it locally. Any idea what's going on?

cbjeukendrup:
Yes, should be fine after a rebase now!

Issue #13295, opened 2022-09-11
[MU4 Issue] apersands in items under Publish tab, non accesible by keyboard


sammik:
**Describe the bug**
In Publish tab, thereare ampersands in items - Pu&blish to ..., &Export...

**To Reproduce**
1. Click "Publish" tab
2. See items
3. try to navigate with keyboard

**Expected behavior**
Underlined letter, accesible by "Alt" (or letter)

**Screenshots**
![apersands](https://user-images.githubusercontent.com/1646034/189547484-b2002362-24ae-4618-919d-43f052034151.png)

**Platform information**
 - OS: [reported on Linux]

wizofaus:
Odd that I do see the same for "Publish" but not for "Export" (in the code there's no & before Export).

bkunda:
The dreaded ampersand has also managed to work it's way into the "Print" button as well...

<img width="432" alt="Screenshot 2022-10-05 at 5 11 45 pm" src="https://user-images.githubusercontent.com/86290556/194096436-edac63ed-c84f-4ce1-b248-d99c6d4d960e.png">

@Tantacrul FYI
@RomanPudashkin if it's at all possible, we'd really like to get this fixed before Beta goes out. Let us know what you think!

Tantacrul:
Yes. It's particularly problematic that the '**P&ublish to MuseScore.com**' button has this problem. We'll get a handful of grief for that :)

cbjeukendrup:
Can't we just get rid of those ampersands and mnemonics? They keep being problematic forever. Anyway, we probably can't, otherwise people will start complaining...

The reason that they popup everywhere, is that we define the "title" of an action globally. So the same string will be used for the menu bar, for the toolbars and any other place where we use that action. For the menu bar, the string needs to contain the ampersand, so that will appear in all other places too. 

So what is the desired solution? Of course, we can process each string to remove the ampersand before displaying it, but is that not bad for performance?

Another issue is that there is not a very clear distinction between strings where translators can and can't add ampersands. (A translation may namely have ampersands where English can't have it, because in English all letters might be used within a menu but in other languages there might be a free letter left.) 
Ideally we'd put strings that can have ampersands in a separate context, so that we can easily label them on Transifex to make them distinguishable, and also so that they won't collide with strings that should not contain ampersands. 
[...]

Issue #12922, opened 2022-08-18
[MU4 Issue] Shortcuts which don't have ampersand present in search results when using ampersand as a search key


abariska:
**Screenshots**
![Capture 2022-08-18_02-54-51_PM](https://user-images.githubusercontent.com/101250347/185395279-e886075a-9602-45c4-a987-baf28078c19d.png)

**Additional context**
Could be fix along with https://github.com/musescore/MuseScore/issues/12921

PR #13639, opened 2022-10-05
Introduce MnemonicString to guarantee handling of mnemonics ampersands

cbjeukendrup:
Resolves: #13295 
Resolves: #12922

This idea was initially proposed by @shoogle, based on my TranslatableString idea, to force at compile time that we don't forget to handle those mnemonics. At first I was a little bit worried about the performance, but we don't have much choice; we have to process every string anyway. Another thing was that some of the work actually needed to be done at the QML side. But I moved all hard work to C++, which is said to be faster than doing it in QML. 
In the end, the performance cost is not _very_ bad:
```
Main thread. Top 150 by sum time (total count: 158)
Function                                                    Call time         Call count        Sum time          
MnemonicString::qTranslatedWithoutMnemonics                 0.018 ms          13998             258.135 ms        
MnemonicString::qTranslatedWithMnemonicUnderlines           0.017 ms          5782              98.640 ms         
```
(This is after a not-very-short stress test where I moved my mouse through various menus to quickly open them all, opened the list of shortcuts in Preferences, etc.. It was on an M1 Mac though.)

To avoid huge diffs (and a _lot_ of work that is not easily automatable), I made the constructor of MnemonicString implicit, so that we don't need to replace TranslatableString(...) with MnemonicString(...) everywhere.

### DIFF
--- a/src/framework/ui/types/mnemonicstring.cpp
+++ b/src/framework/ui/types/mnemonicstring.cpp
@@ -1,0 +1,69 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2022 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include "mnemonicstring.h"
+
+using namespace mu::ui;
+
+static QString processMnemonic(const QString& str, bool showUnderlines)
+{
+    if (str.size() <= 1) {
+        return str;
+    }
+
+    QString result;
+    result.reserve(str.size());
+
+    const QChar* c = str.data();
+    int l = str.length();
+    while (l) {
+        if (*c == '&') {
+            ++c;
+            --l;
+            if (!l) {
+                break;
+            }
+
+            if (showUnderlines && *c != '&') {
+                result.append(QStringLiteral("<u>")).append(*c).append(QStringLiteral("</u>"));
+                ++c;
+                --l;
+                continue;
+            }
+        }
+
+        result.append(*c);
+        ++c;
+        --l;
+    }
+
+    return result;
+}
+
+QString MnemonicString::qTranslatedWithMnemonicUnderline() const
+{
+    return processMnemonic(m_raw.qTranslated(), true);
+}
+
+QString MnemonicString::qTranslatedWithoutMnemonic() const
+{
+    return processMnemonic(m_raw.qTranslated(), false);
+}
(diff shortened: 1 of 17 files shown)


src/framework/ui/types/mnemonicstring.cpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+#include "mnemonicstring.h"
+
+#ifndef NDEBUG
+#include "log.h"
+#endif
+
+using namespace mu::ui;
+
+static QString processMnemonics(const QString& str, bool show)
```
RomanPudashkin, line 30: showUnderline?

RomanPudashkin: It would be great to have some utests for MnemonicString since logic here isn't that trivial

src/framework/uicomponents/view/menuitem.cpp
```diff
@@ -46,14 +46,19 @@ QString MenuItem::id() const
     return m_id;
 }
 
-mu::TranslatableString MenuItem::title() const
+MnemonicString MenuItem::title() const
```
RomanPudashkin, line 49: looks like this function is unused

src/framework/ui/types/mnemonicstring.cpp
Eism: There can be only one mnemonic in the text. Should be processMnemonic

Also, you can see how qt guys solves a similar issue with mnemonic - QKeySequence::mnemonic

src/framework/uicomponents/qml/MuseScore/UiComponents/internal/StyledMenuItem.qml
```diff
@@ -92,7 +92,7 @@ ListItemBlank {
             text += " " + qsTrc("ui", "Menu")
         }
 
-        return Utils.removeAmpersands(text)
```
Eism, line 95: I'm not sure that we get text without mnemonics here.

cbjeukendrup: `text` is based on `itemPrv.title`, which contains certainly no mnemonic ampersands. The other strings that are appended to it also certainly don't contain ampersands (and for the translators, I think it is clear enough that they should not add ampersands here).

src/framework/ui/types/mnemonicstring.cpp
cbjeukendrup: Yes, I based my implementation on Qt's. However, in Qt this routine is implemented in at least 3 different places and in each place in a slightly different way, so I chose the "greatest common divisor" of all their implementations.

RomanPudashkin (approved):

DmitryArefiev:
@cbjeukendrup Tested #13295 #12922 #12921 on Win10, Mac12, LinuxMint 20.2 

#13295 #12922 are **FIXED**  

But #12921 **still occurs** (Lin/Win)


Capture

cbjeukendrup:
@DmitryArefiev These ones are issues with the translations, and must be fixed on the Transifex side (that's the only way).

In fact, they _are_ already fixed on Transifex, so after updating the translations (Preferences > General), the problem shouldn't occur anymore.

DmitryArefiev:
> after updating the translations (Preferences > General), the problem shouldn't occur anymore

@cbjeukendrup Yes, that works! But after **Factory reset** the problem returns

cbjeukendrup:
@DmitryArefiev That's right; when doing a factory reset, the whole folder with user app data is deleted, and updated translations are also stored in that folder. So after a factory reset, the translations that were originally packaged with the build are used again.

We will at some point update the translation files that are stored in the repository, so that the translations packaged with PR builds will be updated. For the released version, we don't need to worry; during the release build process, the very latest translations are pulled from Transifex anyway.

DmitryArefiev:
> @DmitryArefiev That's right; when doing a factory reset, the whole folder with user app data is deleted, and updated translations are also stored in that folder. So after a factory reset, the translations that were originally packaged with the build are used again.
> 
> We will at some point update the translation files that are stored in the repository, so that the translations packaged with PR builds will be updated. For the released version, we don't need to worry; during the release build process, the very latest translations are pulled from Transifex anyway.

@cbjeukendrup Ah, got it! Thanks!

PR #13642, opened 2022-10-06
Fix #13524

### LINKED ISSUE #13524: [MU4 Issue] Inspector > Note > Stem: enabledness of "Flag offset" doesn't update always correctly
**Describe the bug**
The idea is: when the selected note has such a duration that it has a flag, the flag-related settings in Inspector > Note > Stem become enabled, otherwise they are disabled. 

But when you change the duration of a note in such way that the note will gain or loose its flag while the inspector is open, the enabled-ness is not updated.

Also, when you first select an eighth note, so that "Flag offset" is enabled, and then a quarter note, the enabled-ness is not updated either. In this case, the other way around it works correctly though.

**To Reproduce**
1. Select quarter note
2. In Properties > Stem > Show More, notice that "Flag offset" is disabled (correct)
3. While keeping the inspector open, change the duration of the note to eighth
4. Notice that "Flag offset" is still disabled (should become enabled)

Other way:
1. Select eighth note
2. In Properties > Stem > Show More, notice that "Flag offset" is enabled (correct)
3. Select quarter note
4. Notice that "Flag offset" is still enabled (should become disabled)

The other way around (first selecting quarter note, then eighth note) does work correctly.

donarturo11:
Resolves: #13524 

Add enable mode as false when model flag or stem is empty.
In some cases stem or flag is not used so I added disabling stem offset too.


cbjeukendrup:
@donarturo11 Please rebase and retest; the following lines should be removed (for hookModel too):
```qml
                    horizontalOffset: root.stemModel ? root.stemModel.horizontalOffset : null
                    verticalOffset: root.stemModel ? root.stemModel.verticalOffset : null
```
(`horizontalOffset` and `verticalOffset` together have been replaced by `propertyItem`.)

Also, please use rebasing instead of merging; i.e. 
```sh
git pull --rebase upstream master
```
if upstream points to https://github.com/musescore/MuseScore.

donarturo11:
Ok. Thank you. I'll make it yet this evening.

donarturo11:
@cbjeukendrup Already made. Could you look, please?

cbjeukendrup:
@donarturo11 Thanks! I just tested the PR, and I found that the "enabledness" is now correctly updated when in all cases when a different note is selected.

However, the other issue still occurs for me: if you change the duration of a note so that a flag is added or removed, while keeping the note selected, the enabledness is not yet updated correctly: 

https://user-images.githubusercontent.com/48658420/194955090-8bd57a7b-ac75-41a0-a0bd-a51a95f0cefe.mov

The underlying problem must be a bit deeper. The problem is possibly that the addition or removal of the flag/stem is not noticed at all by `ElementRepositoryService`, which is responsible for exposing selected (and related) elements to the Inspector Models. That would be my bet. So you'd need to somehow make sure that `ElementRepositoryService::updateElementList` is _also_ called when the notation is changed, but preferably only when the selected chords (if any) are modified, because only in that case a flag or stem might be added or removed. (Otherwise, we'd be updating the model for every change which is an unnecessary waste of performance.)

Let me know if you get the chance to dive into this!

donarturo11:
@cbjeukendrup I'm not sure if I'm near resolve this problem. I noticed, that when clicking another note, models are refreshed.
I debugged [elementrepositoryservice.cpp:60](https://github.com/musescore/MuseScore/blob/58a64c72b14a7f4ba04969250408b39bb314d34b/src/inspector/internal/services/elementrepositoryservice.cpp#L60). I noticed that if I select other note, like in first case, ```m_rawElementList != newRawElementList = true``` and entire function returns true. If I deduced well, if I select another but single note, new selection state should be identical with state saved in class, but If I change note duration, where is less elements, new list shouldn't be identical but should be different, so mentioned element list comparison should be truth and function should return true. 
Tomorrow probably I won't have a lot of time to sit to computer to find solution (entire day I'll working in my garden), but I'll try to do it in Thursday, or earlier if I find some time.

cbjeukendrup:
@donarturo11 That sounds right to me!

So I think that `ElementRepositoryService::updateElementList` is called correctly when the selection is changed, but it seems like it is not called when you change the duration of the selected note (without changing the selection). 

In `InspectorListModel::listenSelectionChanged()`, a callback is registered to update the element list every time that the selection is changed. Initially, I thought we'd need to register another callback to update the element list also when the notation is edited. So, together with the surrounding code, somewhat like this:
```c++
    updateElementList();

    INotationPtr notation = context()->currentNotation();
    if (notation) {
        notation->interaction()->selectionChanged().onNotify(this, updateElementList);

        notation->notationChanged().onNotify(this, updateElementList); // this line is added
    }
```

However, this does not work: because the selection (as returned by `notation->interaction()->selection()`) is still the same, `ElementRepositoryService::needUpdateElementList` (which is eventually called) will return false (because the lists are equal), so the Inspector is not updated.

But even though the "raw" selection has not changed, the result of `ElementRepositoryService::findStems` _will_ be different. So the Inspector would still need to be updated. 

A possible solution would be that when the notation is changed, we update the Inspector regardless of whether the "raw" selection was changed or not. That could be done by adding an extra parameter to `updateElementList` and `needUpdateElementList` like `bool wasNotationEdited` which will be true if the reason that we're trying to update is that the notation was changed.

Final note: if this "sketch" solution turns out to work, then it would need to be refined a bit regarding performance: when the notation was edited, we only need to update the inspector if the selection contains "chord-related" elements (notes, flags, stems, chords, dots), since the problem only occurs with those elements. To improve performance even further, you could replace `notation->notationChanged()` with `notation->undoStack()->changesChannel()`, and then look at the `changedTypes` of the changes range; only if it contains types of chord-related elements, then the inspector needs to be updated.

That's a long story, but I hope it's helpful!

Issue #12600, opened 2022-07-29
[MU4 Issue] AppImage desktop file has wrong Name, leading to missing icon


MarcSabatella:
**Describe the bug**
If you install the AppImage (via the install option), then run from the command line, the window manager is unable to match the process to the desktop file, leading to a missing icon

**To Reproduce**
1. run the AppImage from the command line with the "install" option
2. run the AppImage from the command line again, without the "install" option
3. look at your taskbar, or whatever your window manager provides to show icons for your running programs

**Result**
Unless your window manager is doing some special magic to try to find the right desktop file, you'll see only a generic icon for MuseScore 4, not the actual MuseScore icon

**Expected behavior**
MuseScore icon should appear on the taskbar

**Screenshots**
Here's what I see - the icon to right is what represents MuseScore 4 for me:
![image](https://user-images.githubusercontent.com/1799936/181789981-3bcb0ab4-6d02-4e42-9a3b-5c7e157fb6d2.png)

**Platform information**
 - OS: Linux - debian, running within ChromeOS (but presumably similar issues exist for other window managers)

**Additional context**
This issue is fixed simply by changing the Name field in the desktop file to be what the application actually registers itself as on startup: MuseScore4Development at the moment, presumably it will be other things for milestone releases.

HemantAntony:
Something related, on my Ubuntu, when I build MuseScore, it does not show up with an icon. However, since then I have installed MuseScore 2.0 (`sudo apt install musescore`). Now the icon for MU4 becomes MU2's icon. I have only tested this with compiled builds

MarcSabatella:
As per https://musescore.org/en/node/316255, this is definitely a more widespread issue.  An important but trivially simple fix for anyone who knows how the AppImage packaging process works - just get the "Name" field right.

shoogle:
I'm afraid I can't reproduce this as my desktop environment (GNOME) on Ubuntu 20.04 seems to be clever enough to pair the process with the .desktop file, even when run from the command line (perhaps it looks at the Exec line of the .desktop file to get the binary location?).

Anyway, I would rather change the process name to match the `Name` field in the .desktop file rather than the other way around. If the Name field in the .desktop file was changed to a fixed string like "MuseScore" then you could only have one version of the app installed at a time because only one version would show up in your OS's application launcher. Version numbers and a "-portable" or "-portable-nightly" suffix were appended to the Name in MuseScore 3; so that different versions and build types would show up separately in the app launcher.

[...]

PR #13651, opened 2022-10-06
AppImage: Add '4' suffix to install paths to prevent conflicts with MU3 on Linux

shoogle:
This results in the AppImage's internal binary being called `mscore4-portable` instead of `mscore-portable`. AppImage resources are installed to paths like:

    bin/mscore4-portable
    share/applications/mscore4-portable.desktop
    share/icons/hicolor/SIZE/apps/mscore4-portable.png
    share/icons/hicolor/SIZE/mimetypes/application-x-musescore4-portable.png
    share/mime/packages/musescore4-portable.xml
    share/man/man1/mscore4-portable.1.gz

This enables MuseScore 4 to be installed alongside MuseScore 3 on Linux.

It was necessary to change the app name in the .desktop file to avoid MuseScore 4.0 being called "MuseScore 4.04-portable". The new name is "MuseScore 4.0 Portable", with "Portable" there to distinguish our AppImage from distribution packages in the OS's application launcher.

---

FYI, our AppImage can be installed using the `install` option on the command line:
```
chmod +x MuseScore*.AppImage
./MuseScore*.AppImage install
```

### DIFF
--- a/build/Linux+BSD/mscore.desktop.in
+++ b/build/Linux+BSD/mscore.desktop.in
@@ -1,6 +1,6 @@
 [Desktop Entry]@Variables_substituted_by_CMAKE_on_installation@
 Version=1.0@This_is_version_of_Desktop_Entry_Specification@@It_is_NOT_MuseScore_version@
-Name=@MUSESCORE_NAME@ @MUSESCORE_VERSION@@MSCORE_INSTALL_SUFFIX@
+Name=@DESKTOP_LAUNCHER_NAME@
 GenericName=Music notation
 GenericName[de]=Notensatz
 GenericName[fr]=Notation musicale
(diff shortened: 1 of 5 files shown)


MarcSabatella:
Is there any way we could also fix the desktop file?  See https://github.com/musescore/MuseScore/issues/12600.  The "Name" field in the desktop file needs to match whatever the process registers for itself at run time, otherwise many window managers cannot assign the proper icon.  This was broken in MU3; it would be great to get this right in MU4.

shoogle:
@MarcSabatella, I would prefer to solve #12600 in PR #13654 so as not to delay this one.

igorkorsukov (approved):

shoogle:
I've moved this code into PR #13654

PR #13654, opened 2022-10-07
Update AppImage installation files

### LINKED ISSUE #12600: [MU4 Issue] AppImage desktop file has wrong Name, leading to missing icon

shoogle:
~~Attempt to fix #12600. I can't reproduce the problem so I'll need others to download and install the CI build to test.~~

- Update and reformat the .desktop and .appdate.xml file.
- Fix #12600.
- Prevent conflicts when MU3 and MU4 are installed on the same machine.

See individual commit messages for details.


MarcSabatella:
For me this does not work.  The application is still registered as MuseScore4Development as far as the desktop is concerned, so that is what the name needs to be for it to work.  I have no idea where that name comes from, but it's the displayed by, among other things, accerciser.

shoogle:
@MarcSabatella, please try again. I think the secret is adding `StartupWMClass=MuseScore4Development` in the `.desktop` file, but I've also changed the name of the `.desktop` file to match the recommended value `org.musescore.MuseScore.desktop` and taken the opportunity to update the AppStream `.appdata.xml` file.

MarcSabatella:
Bingo!  I think you're right about StartupWMClass being the correct answer.  Obviously it will need to be plain "MuseScore 4" or whatever for release builds, but it *looks* to me like your change will do the right thing here (assuming the variables I see being referenced have the values it looks like they should).

I have been using Name only because I discovered through trial and error that it worked :-).  One user-visible difference is that with the Name change, the tooltip on the icon actually reads MuseScore4Development, but with this version, it reads "MuseScoreDev 4.0 Portable".  So, that's nice - it means we have independent control over the *name* of the application versus how the desktop association is made.

I wasn't aware of recommended naming of the desktop file, but I see it now in the freekdesktop spec, so good call there as well.

Anyhow, thanks for looking at this!  It's always a challenge to fix issues you can't reproduce, but I also appreciate your going the extra mile to find the *right* fix.

igorkorsukov (approved):

Issue #12790, opened 2022-08-11
[MU4 Issue] Staccato notes under slurs play loud


la-bel:
**Describe the bug**
When you make notes staccato under a phrase mark, they play much louder than surrounding notes.

**To Reproduce**
1. Create a score
2. Enter two or more notes
3. Make one of the notes staccato
4. Add a slur over the notes
5. Play back the notes: the staccato note is much louder than the others. This appears to be relative to the general dynamic.

**Expected behavior**
The staccato notes to be at or close to the general dynamic.

**Screenshots**
<img width="904" alt="Screenshot 2022-08-11 at 12 23 00" src="https://user-images.githubusercontent.com/63920211/184122762-e1cdf93d-4d92-49e0-b905-ec5e3ec58d91.png">

**Platform information**
 - OS: macOS

la-bel:
This also happens with other articulations including (at least) staccatissimo, tenuto and staccatissimo stroke.

DmitryArefiev:
Also can reproduce it on my side.

https://user-images.githubusercontent.com/90187801/184181989-bb8ec591-f477-410c-84f5-2eaf4df68fc1.mp4

Thanks for logging!

(can be related to #12678)

abariska:
I'd say more. Notes with staccato plays louder in a few more cases:
- when staccato and slur are applied at the same time but on the different staves 
- when slur is applied to the note before or after the note with staccato
- when applied combined articulation which contain staccato (staccato+accent or staccato+tenuto)
Good to use this score for reproducing 
[Danse_a_deux.mscz_FILES.zip](https://github.com/musescore/MuseScore/files/9551298/Danse_a_deux.mscz_FILES.zip)

MarcSabatella:
Same for arpeggio - see https://github.com/musescore/MuseScore/issues/13340

DmitryArefiev:
Fixed in #13656

Issue #11933, opened 2022-06-08
Accents not producing an audible change in dynamic. Marcato needs more audible change.


Tantacrul:
**Describe the bug**
In the video below, you'll see that **accents** do not appear to increase the dynamics and **marcato** only a _little_ bit. (It is possible accents are working but the % increase in volume is not easy to hear)

**Video**

https://user-images.githubusercontent.com/21022311/172640024-878b0163-2d18-4573-9006-e55dfcde0638.mov

**Expected behavior**

1. Accents should increase by an audible amount (like how Marcato currently sounds, but slightly more)
2. Marcato should be slightly louder than accents

PR #13656, opened 2022-10-07
[MU4] staccato_loudness

vpereverzev:
Resolves: #12790

### DIFF
--- a/src/framework/mpe/mpetypes.h
+++ b/src/framework/mpe/mpetypes.h
@@ -806,6 +806,10 @@ private:
             for (auto& pair : m_averageDynamicOffsetMap) {
                 pair.second /= dynamicChangesCount;
             }
+        } else if (dynamicChangesCount == 0) {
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #12790 on Win10, Mac12 - **FIXED**

MarcSabatella:
This is *much* better for sure - Reunion almost sounds like music now! :-) - but regular accent marks still don't play at all that I can tell.  Marcato does.  not sure if that was meant to have been fixed this or not.

DmitryArefiev:
> This is _much_ better for sure - Reunion almost sounds like music now! :-) - but regular accent marks still don't play at all that I can tell. Marcato does. Not sure if that was meant to have been fixed this or not.

@MarcSabatella  Oh I see that now (comparing with MS3). Please log a separate issue when you have time. The current PR fixes loud issues when slurs involved.

Thanks!

MarcSabatella:
There is such an issue already - see https://github.com/musescore/MuseScore/issues/11933

PR #13657, opened 2022-10-07
Improved score loading time (and some general cleanup regarding multi-measure rests)

mike-spa:
When opening a score, we currently have a `Score::relayoutForStyles()` method which forces a complete relayout if the score contains mmRests or a specific type of measure repeats. That is obviously inefficient: it can end up literally doubling the score loading time for no good reason. The reason this method was created is (probably) to fix a couple of our trademark "first layout is wrong but corrects itself after relayout" problems, by brute-forcing a complete score relayout. I have now removed this relayout and (hopefully) fixed the problems at their origin.
- In the case of MMrests, the issue is caused by the fact that their shape is accounted for when computing horizontal spacing, but the shape itself also changes its width to fill the measure, so in the end we were effectively spacing the MMrest against itself. This is something I had already stuggled with when working on horizontal spacing and had to bodge some kind of solution. The more correct solution, which I've implemented now, is to simply _ignore_ the shape of mmRests. Multi-bar rests are by definition empty measures, so there is never going to be any need of spacing other items against an MMrest. This way, I could also remove my previous bodge solution and simplify the whole thing.
- In the case of these measure repeats, it appears to be just a missing layout call.

I've also taken the change to make a small tweak to the width calculations for MMrests, which at the moment can sometimes end up being disproportionately wide.

### DIFF
--- a/src/engraving/libmscore/measure.cpp
+++ b/src/engraving/libmscore/measure.cpp
@@ -3273,6 +3273,7 @@ void Measure::layoutMeasureElements()
                     e->setPosX(measureWidth - .5 * e->width());
                 } else {
                     // full measure rest or one-measure repeat, center within this measure
+                    e->layout();
                     e->setPosX((x2 - x1 - e->width()) * .5 + x1 - s.x() - e->bbox().x());
                 }
                 s.createShape(staffIdx);            // DEBUG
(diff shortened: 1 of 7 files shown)


mike-spa:

PR #13659, opened 2022-10-07
Only layout a score if the score is open!

mike-spa:
Well, what the title says. At the moment, when loading a score, we are also computing the layout of every part, even if all parts are closed. On a big piece with many parts, this dramatically increases the loading time of the project. 
Now, unless there is something deeper that I am missing here (which is very likely), I think it makes more sense that we only compute the layout of things that are going to appear on the screen. So, at first, we don't layout closed scores, and we lay them out when they are opened. This, of course, could mean that there is a small lag when the user opens one or more parts, but I think it is much preferable than bunching all of that waiting time at the beginning (and using it to lay out a huge amount of stuff that the user may likely not even need).

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -419,6 +419,9 @@ void Score::update(bool resetCmdState)
         ms->deletePostponed();
         if (cs.layoutRange()) {
             for (Score* s : ms->scoreList()) {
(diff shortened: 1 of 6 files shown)


RomanPudashkin:
Cool optimization! But please make sure that this change won't break the export: the user can export parts that haven't been opened yet

mike-spa:

PR #13667, opened 2022-10-08
[MU4] accent_adjustments

vpereverzev:
Resolves: #11933

### DIFF
--- a/src/framework/mpe/resources/general_keyboard_articulations_profile.json
+++ b/src/framework/mpe/resources/general_keyboard_articulations_profile.json
@@ -14,35 +14,35 @@
                         },
                         {
                             "offsetPosition": 1000,
-                            "offsetValue": 1500
+                            "offsetValue": 2000
                         },
                         {
                             "offsetPosition": 2000,
-                            "offsetValue": 3000
+                            "offsetValue": 4000
                         },
                         {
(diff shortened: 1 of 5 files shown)


Tantacrul:
I can see from an exported waveform that marcato is a tiny bit louder than accent and I think that's fine.
image

The accents are audible - especially for piano. Part of me thinks we should push it a **tiny** bit more but for now, it's probably best to merge this PR and then see how it all hangs together once we start using key switches in VST.

RomanPudashkin (approved):

MarcSabatella:
FWIW, in MU3, accents were *crazy* loud for piano, a constant source of complaints.  So more subtle will be a welcome change.

For marcato, it's worth noting that in the jazz world (and also funk, salsa, etc), marcato normally implies *shorter* as well as louder.  That's what MU3 did, and jazzers were happy with that, but classical folks not so much.  Now, I guess it will be the other way around.  I gather that eventually it will be possible for users to customize this, but meanwhile, if it's possible to customize the definitions in the jazz templates, that would be nice.  I see there are still "gateTime" tags in the mscx file, but they don't seem to be doing anything.  I tried changing to durationFactor but no luck there either.  Is there something else I can do to manually customize the jazz template mscx file to give shorter marcato?

Issue #13585, opened 2022-10-02
[MU4 Task] Fix enums in the plugin API (but how?)


cbjeukendrup:
**Task description**
Currently, enums are broken in the plugin API. This affects many plugins, also those few that we want to pack with MuseScore, so it really needs to be solved.

**How are enums designed to work in the plugin API?**
An enum is exposed as an object of type `Enum`, which is not much more than a map from the name of an enum case to the value of that enum case. For example, the ElementType enum gets represented somewhat like this:
```js
{ 
    "CHORD" : int(ElementType::CHORD),
    ...
}
```
So it can be used in QML as `Element.CHORD`, which has type int, where `Element` is the name under which the enum is exposed.

**How is such an enum object constructed?**
Such an object is populated with the help of QMetaEnum. We create a QMetaEnum using `QMetaEnum::fromType<ElementType>()`, and then loop over the key/value pairs that it offers.
QMetaEnum offers these data based on the data that is collected by Qt's meta object compiler (moc). 

**And that is where it goes wrong...**
There are basically two limitations that make our life very difficult:
- The moc can not look in included files
- When using Q_NAMESPACE to expose enums that are in a namespace to the moc, all those enums must be in the same file, because if you put Q_NAMESPACE in the same namespace in multiple files, the moc will generate the definition of the static meta object in all the moc files for all those files, leading to conflicting redefinitions. 

**Solutions**
1. We could try to organise everything in such a way that we work around it. That means, we put all enums that we want to expose from the engraving module in one file, put Q_NAMESPACE macro in that file, and make all enums Q_ENUM_NS. It would probably work, but it's not nice, since we need to put a lot of stuff in one file, most notably the SymId enum too. 
(Note that the trick of putting the actual SymId enum in symid_p.h and then #including that, _does not work_ either, because moc can't look in included files, so it can't see the enum and ignores the Q_ENUM_NS macro.)

2. We could _copy_ all enums to the plugin api module, so that the engraving module stays clean as it is. The obvious big disadvantage is that we need to keep all that stuff in sync, which is guaranteed to go wrong at some point. And... copying SymId is not a particularly fun thing to do...

3. We could put all enums in the Engraving module in Q_GADGET wrapper classes (or Q_NAMESPACE wrapper namespaces), and use a typedef to be able to use the enum normally, i.e. 
    ```c++
    class SymId_wrapper {
        Q_GADGET
     public:
        enum E {
            .......
        };
        Q_ENUM(E)
    };
[...]

Issue #12867, opened 2022-08-15
[MU4 Issue] Courtesy accidental plugins do not work


MarcSabatella:
**Describe the bug**
I can enable and run the courtesy accidental plugins, but none of them do anything

**To Reproduce**
1. Enable Add courtesy accidentals plugin
2. Create score in C major, 4/4
3. Add quarter notes C C C Bb | B (so the B is in the second measure
4. Select all
5. Run Add courtesy accidentals

**Result**
Nothing happens

**Expected behavior**
The B should receive a courtesy natural

**Screenshots**
![image](https://user-images.githubusercontent.com/1799936/184644640-950fd8f1-6cae-4e0a-9362-25a74aa18754.png)

**Platform information**
 - OS: Linux

**Additional context**
The other courtesy accidentals plugins (remove, configure) do not seem to work either.  But other plugins do.

MarcSabatella:
Actually, "Color notes" works, but neither "new retrograde" nor "mirror intervals" do.

Jojo-Schmitz:
@heuchi is the original author (of these Courtesy accidental plugins) and might be interested in this

heuchi:
It's new-major-version time again ;-)
I hope the plugin frame work hasn't changed too much.
I don't really have time to look into this now and I haven't even looked at version 4, either. So it might take a while...
But anyway, thanks for the notification.

heuchi:
I tried to look at this, but I ran into some difficulties. I don't know if this is the best place to share it, but since they seem to be of different origins but all affect plugin development, I'd list them here. Feel free to suggest a different place. :-)

I'm using the latest nightly build AppImage for Linux ("MuseScoreNightly-latest-x86_64.AppImage" from Aug 28)

1. Custom qml files (in private "Plugins" folder) are only found if they are not symbolic links. (minor but impractical)
2. There seems to be no easy way to reload a plugin (other than restarting MuseScore). The "Reload" button in the plugin manager doesn't seem to do anything. (again minor, but impractical)
3. console.log("some debug message") calls don't appear on the terminal window where I started the AppImage, and I wasn't able to find them anywhere else either. Where do they go? (major problem for debugging)

On a side note: "Color notes" works but cannot be undone.

heuchi:
I managed to compile MuseScore myself again, which made the console.log() problem go away. The nightly builds don't seem to be debug builds, so they don't show the messages created by plugins.

[...]

PR #13669, opened 2022-10-08
Fix enums in the Plugin API

cbjeukendrup:
Resolves: #13585 
Resolves: #12867

Uses "solution 2" from #13585 

Added a test case to verify the fix


igorkorsukov (approved):

DmitryArefiev:
Tested #12867 on Win10, Mac12 - **FIXED**

Issue #13429, opened 2022-09-20
[MU4 Issue] Articulation is deleted after changing note duration


abariska:
**Describe the bug**
Two cases are present:
- One articulation is added to a note (case 1)
- Two articulations are added to a note (case 2)

**To Reproduce**
1. Create a simple score
2. Add a quarter note
3. Add staccato to the note
4. Change duration of the note

To get case 2 on the step '3' add staccato and accent

**Expected behavior**
Articulation s are not affected by duration change

**Screenshots**
Case 1

https://user-images.githubusercontent.com/101250347/191368104-3aa5c7fb-9285-40f4-a492-bb693fa32054.mp4

Case 2

https://user-images.githubusercontent.com/101250347/191368139-c6fa3fac-fa9b-450b-9f06-7569bd416618.mp4

oktophonie:
I notice that articulations are also deleted (Case 1) when moving notes from one voice to another:

https://user-images.githubusercontent.com/20604032/192024778-d673dcd0-8bd5-4651-ace0-5e98f7d72981.mp4

Tantacrul:
This is actually a really bad regression from V3, so I've bumped it to Beta 1. @oktophonie - is it something for your team do you think?

cbjeukendrup:
PR created: #13675

DmitryArefiev:
Fixed in #13675

DmitryArefiev:
> I notice that articulations are also deleted when moving notes from one voice to another:
> box-220923-1836-39.mp4

The same issue in MuseScore3. Should be logged separately and solved in 4.x or later

PR #13675, opened 2022-10-09
Preserve articulations when changing chord duration

cbjeukendrup:
Resolves: #13429

@Eism I see that you had added this code in 68e609bab08561db95fc5fbf0b439fa1dc01ce07. What was it intended to do?

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -3305,12 +3305,6 @@ void Score::padToggle(Pad p, const EditData& ed)
             undoChangeChordRestLen(cr, _is.duration());
         } else {
             changeCRlen(cr, _is.duration());
-
-            for (const SymId& articulationId: _is.articulationIds()) {
-                Articulation* na = Factory::createArticulation(cr);
-                na->setSymId(articulationId);
-                toggleArticulation(cr, na);
-            }
         }
     }
 }


Tantacrul:
Just tested it. Seems to work exactly as expected. Hopefully, this is an easy one to merge because it was quite a big usability regression! :D

Thanks @cbjeukendrup!

Eism:
@cbjeukendrup 
It's hard for me to remember exactly what happened(two years have passed), but everything seems to work as expected after your changes.

We had a design document for this feature. We just need to test the user cases from this document.
The problem is that I can't find this document(@Tantacrul)

Tantacrul:
This is the document where I put together the design for the articulations section of the note input bar. You'll need to skip the first few pages.
https://www.dropbox.com/s/ms5554bsej5xnxu/Note_Input_Bar_v03.pdf?dl=0

DmitryArefiev:
@Tantacrul I found some changes in behavior (comparing with MS3) when triggering Input mode after selecting note with articulation:

 

https://user-images.githubusercontent.com/90187801/195113364-f7584deb-263f-4402-95f9-7f9c22f7ab24.mp4

Tantacrul:
@DmitryArefiev - I'm not too worried about that. This difference has existed for months and no one has complained. I'm not even sure it needs to be changed.

My sense is that we **have** to fix the problem logged in my issue - and we can potentially add additional tweaks either in post beta or in 4.x. What do you think?

DmitryArefiev:
> @DmitryArefiev - I'm not too worried about that. This difference has existed for months and no one has complained. I'm not even sure it needs to be changed.
> 
> My sense is that we **have** to fix the problem logged in my issue - and we can potentially add additional tweaks either in post beta or in 4.x. What do you think?

@Tantacrul You are right! I though it was introduced by the current PT but the same result in master build. 

I totally agree to think/discuss that difference in behavior later.

DmitryArefiev:
Tested #13429 on Win10, Mac12, LinuxMint 20.2 - **FIXED**    

@Eism Please review

Eism (approved):

Issue #13660, opened 2022-10-07
[MU4 Issue] Arpeggio re-renders slur shape after reopening score


DmitryArefiev:
https://user-images.githubusercontent.com/90187801/194597351-d82488b7-4272-4852-96d9-38dd9eb728b4.mp4

**Platform information**
macOS, Windows, Linux]

DmitryArefiev:
@oktophonie @Tantacrul FYI

(Not sure if it's logged already)

mike-spa:
This is caused by the bounding box of the arpeggio (here in yellow) being wrongly computed at first:
<img width="1027" alt="image" src="https://user-images.githubusercontent.com/93707756/194601445-582c0535-2633-4f50-8615-84256a5fd5e0.png">
and then fixing itself after a relayout:
<img width="1066" alt="image" src="https://user-images.githubusercontent.com/93707756/194601531-8e6cf276-b96c-4e1c-9ae2-011539e5445f.png">
Will get to it first thing on Monday!

DmitryArefiev:
Fixed in #13724

PR #13686, opened 2022-10-10
Fix arpeggio bug

mike-spa:
Resolves: #13660 

The bounding box of arpeggios was wrongly computed at the first layout, which was causing the slurs to be misplaced.
arpeggios


oktophonie (approved):

mike-spa:

PR #13708, opened 2022-10-12
[MU4] Avoid reading multiple instance of the same system object

asattely:
The issue:
image
System markings were being duplicated when ranges of measures were being read and rewritten, like for example when a time signature is added. This was because each linked clone of the system object was also being read, and in turn added, which automatically adds the system marking on all applicable system staves.

Now, range only reads the "main" system object (which is track 0, always), and when writing it will be added to all system object staves as necessary.


asattely:
Note that system text lines, such as _rit._ or _accel._ currently create two copies of themselves on the top staff. This is fixed in https://github.com/musescore/MuseScore/pull/13422, which is at the time of writing still open.

oktophonie (approved):

PR #13710, opened 2022-10-12
Revert "Introduce MnemonicString to guarantee handling of mnemonics ampersands"

vpereverzev:
Reverts musescore/MuseScore#13639


vpereverzev:
@alexpavlov96 see the utests output. Needs to be fixed in a separate PR

PR #13716, opened 2022-10-12
Mnemonics attempt 2

cbjeukendrup:
Still figuring out why linking fails on CI, since locally it builds perfectly fine for me

### DIFF
--- a/src/appshell/qml/platform/AppMenuBar.qml
+++ b/src/appshell/qml/platform/AppMenuBar.qml
@@ -116,6 +116,7 @@ ListView {
         property var item: model ? model.itemRole : null
         property string menuId: Boolean(item) ? item.id : ""
         property string title: Boolean(item) ? item.title : ""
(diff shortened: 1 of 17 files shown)


src/framework/ui/types/mnemonicstring.h
```diff
@@ -0,0 +1,71 @@
...
+ */
+#ifndef MU_UI_MNEMONICSTRING_H
+#define MU_UI_MNEMONICSTRING_H
+
+#include <QString>
+
+#include "types/translatablestring.h"
+
+namespace mu::ui {
+class MnemonicString
```
igorkorsukov, line 30: I think better move this type to global module, near TranslatableString

cbjeukendrup: Yeah, that's probably the best way to handle the linking problems. I'll do that.

RomanPudashkin (approved):

DmitryArefiev:
Re-tested #12922 #13295 and Mixer UI on Win10, Mac12, LinuxMint 20.2 - **PASS**

PR #11126, opened 2022-04-09
Set search filter model as master palette model instead of user palette model (#11113)

mccoydtromb:
Set search filter model as master palette model instead of user palette model

Resolves: #11113 

changed the search filter model to the master palette model. It was previously the user palette model. Now all palette elements are available to the search.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [ ] I created the test (mtest, vtest, script test) to verify the changes I made

PR #13720, opened 2022-10-12
[MU4] Fix #11113: Palette search excludes elements behind "More" button

### LINKED ISSUE #11113: [MU4 Issue] Palette search excludes elements behind "More" button
**Describe the bug**

To avoid cluttering the palettes, some elements are hidden behind a "More" button. Currently these elements do not appear in search results, yet elements from hidden palettes (the ones behind the "Add palettes" button) do appear in search results.

For the sake of discoverability, searches in the palettes should return all matching elements even if those elements are not usually visible in the palettes.

**Example**

The "More" list of the Articulations palette contains an element called "Snap pizzicato above", but this item does not appear in search results. We want it to appear.

"Snap pizzicato above" exists in Articulations > More | Searching for "snap" does not reveal it
---|---
image | image

### Other considerations

When fixing the above, we want to **preserve the following behaviour** which is correct as it is now.

**Add palettes**

Searching for "grace" returns elements from the Bagpipe embellishments palette even though this palette is hidden behind the "Add palettes" button. We want this behaviour to continue.

"Bagpipe embellishments" exists in the 'Add palettes' popup | Searching for "grace" includes Bagpipe embellishments
---|---
image | image

**Custom palettes**

Users can create or load custom palettes and drag additional elements into those palettes. These elements are included in search results even though they may also appear in other palettes. We want this to continue.

Here "Trill" has been added to a custom palette | Searching for "trill" includes the custom palette
---|---
image | image

Basically, we want palette searches to include everything!

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11113

See: https://github.com/musescore/MuseScore/pull/11126

### DIFF
--- a/src/palette/internal/paletteprovider.cpp
+++ b/src/palette/internal/paletteprovider.cpp
@@ -601,7 +601,7 @@ void PaletteProvider::init()
 
     m_searchFilterModel = new PaletteCellFilterProxyModel(this);
     m_searchFilterModel->setFilterCaseSensitivity(Qt::CaseInsensitive);
-    m_searchFilterModel->setSourceModel(m_userPaletteModel);
+    m_searchFilterModel->setSourceModel(m_masterPaletteModel);
 
     m_visibilityFilterModel = new QSortFilterProxyModel(this);
     m_visibilityFilterModel->setFilterRole(PaletteTreeModel::VisibleRole);


Eism (approved):

Tantacrul:
Working for me now. :)

cbjeukendrup:
What we ultimately want to do, is indeed to make both default and custom elements show up. But this is technically complicated, so therefore we decided to go with this solution, at least temporarily, even though it has issues. The reason to prefer the master palettes model over the user palettes model, is that the master palette model will also allow searching for elements that are hidden or deleted in the user's palettes, which make the search feature a lot more useful (users probably already know where they put their custom items, but might not know where to find items that they have never made visible).

worldwideweary:
It's understandable that search has been made to serve the beginners. In the mean time, it might be a good idea to update the handbook to describe the current behavior and how it deviates from previous functionality.

I'd also venture to say that it isn't *a lot* more useful due to the fact that once a user knows where a palette item is (which "category" of palette), they *know* it, and if it isn't in their current palettes, they can add it to their workspace and no longer require searching, whereas the functionality to quickly bring a particular palette to the forefront for applying to score has a recurring use, let alone the ability to re-arrange vertically/horizontally/edit properties/find custom palettes added in by the user, but again I'm biased because I've been using the program for years rather than coming in not knowing what I'm doing. Hope it works out!

PR #13723, opened 2022-10-12
Fix a bunch of GitHub Actions deprecations

cbjeukendrup:
See:
- https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
    - https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/ci_backend.yml
+++ b/.github/workflows/ci_backend.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@0.9.1
+      uses: styfle/cancel-workflow-action@0.11.0
       with:
         access_token: ${{ github.token }}
     - name: Clone repository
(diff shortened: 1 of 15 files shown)


cbjeukendrup:
This fixes all things that can be fixed on our side; for the rest, the "sub-actions" that we use will have to be updated. They will probably fix it in 0.0.1 versions, which will get used automatically because we only specify the major version `v3` for the relevant actions.

But anyway: to be on the safe side, **let's not merge this PR until after 4.0** (even though the risk is low because it only involves changes to very basic CI actions that hardly interfere with the build process, and thus have much lower risk than updating a compiler for example). I'm just putting it here so that we have it ready when they suddenly "hard-deprecate" the things.

igorkorsukov (approved):

cbjeukendrup:
Rebased!

igorkorsukov (approved):

PR #13724, opened 2022-10-12
Optimization: avoiding unnecessary layout calls

mike-spa:
Resolves: https://github.com/musescore/MuseScore/issues/13660

This PR collects together three of my (not yet merged) PRs, namely #13686, #13657 and #13659. Those will be now closed.

The point of this PR is to avoid unnecessary calls to `Score::doLayoutRange()`, which is the main engraving layout function and is extremely expensive. Basically:
- I've eliminated unnecessary layout of closed scores (those are only laid out when they are opened or when they are exported).
- I've eliminated unnecessary relayout for mmRests and measure repeats (this was just a brute-force workaround for a bug, but now I've fixed the bug at the origin so I can eliminate these calls).

The other bug fixes (separated by commit) are necessary because these are our typical "relayout bugs", meaning that the layout is initially wrong and correct itself after a relayout. Now that I have eliminated the relayouts, these bugs have surfaced and have been fixed (which is a good thing). _The vtest differences are due exclusively to the bug fixes_. I've run the tests locally and made sure that the optimization commits leave the vtests absolutely unchanged.

How much of an advantage this PR can give depends on how many parts exist and how many are opened. On an orchestral score, it can make a pretty big difference:

Before (28 layout calls):
before

After (1 single layout call):
after

This needs careful reviewing, especially about the non-layout of closed scores, and making sure that everything still works, especially on export. Everything I've tried seems to work as it should.

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -337,7 +337,7 @@ void Score::undoRedo(bool undo, EditData* ed)
 ///   and (always) updating the redraw area.
 //---------------------------------------------------------
 
-void Score::endCmd(bool rollback)
+void Score::endCmd(bool rollback, bool layoutAllParts)
 {
     if (undoStack()->locked()) {
         return;
@@ -357,7 +357,7 @@ void Score::endCmd(bool rollback)
         undoStack()->current()->unwind();
     }
 
-    update(false);
+    update(false, layoutAllParts);
 
     ScoreChangesRange range = changesRange();
 
@@ -408,7 +408,7 @@ void CmdState::dump()
 //    layout & update
 //---------------------------------------------------------
 
-void Score::update(bool resetCmdState)
+void Score::update(bool resetCmdState, bool layoutAllParts)
 {
     TRACEFUNC;
 
@@ -419,6 +419,9 @@ void Score::update(bool resetCmdState)
         ms->deletePostponed();
         if (cs.layoutRange()) {
             for (Score* s : ms->scoreList()) {
+                if (!s->isOpen() && ms->scoreList().size() > 1 && !layoutAllParts) {
+                    continue;
+                }
                 s->doLayoutRange(cs.startTick(), cs.endTick());
             }
             updateAll = true;
--- a/src/engraving/utests/exchangevoices_tests.cpp
+++ b/src/engraving/utests/exchangevoices_tests.cpp
@@ -104,7 +104,7 @@ TEST_F(Engraving_ExchangevoicesTests, undoChangeVoice)
     // change voice
     score->startCmd();
     score->changeSelectedNotesVoice(1);
-    score->endCmd();
+    score->endCmd(false, /*layoutAllParts = */ true);
     EXPECT_TRUE(ScoreComp::saveCompareScore(score, writeFile1, reference1));
 
     // undo
--- a/src/notation/internal/excerptnotation.cpp
+++ b/src/notation/internal/excerptnotation.cpp
@@ -133,7 +133,6 @@ void ExcerptNotation::setName(const QString& name)
 
     excerptTitle->setPlainText(name);
     score()->setMetaTag(u"partName", name);
-    score()->doLayout();
 
     notifyAboutNotationChanged();
 }
--- a/src/notation/internal/masternotation.cpp
+++ b/src/notation/internal/masternotation.cpp
@@ -431,6 +431,12 @@ mu::ValNt<bool> MasterNotation::needSave() const
 
 void MasterNotation::initExcerpts(const ExcerptNotationList& excerpts)
 {
+    // Scores that are closed may have never been laid out, so we lay them out now
+    for (auto score : masterScore()->scoreList()) {
+        if (!score->isOpen()) {
+            score->doLayout();
+        }
+    }
     for (IExcerptNotationPtr excerptNotation : excerpts) {
         ExcerptNotation* impl = get_impl(excerptNotation);
 
--- a/src/notation/internal/masternotation.cpp
+++ b/src/notation/internal/masternotation.cpp
@@ -431,12 +431,6 @@ mu::ValNt<bool> MasterNotation::needSave() const
 
 void MasterNotation::initExcerpts(const ExcerptNotationList& excerpts)
 {
-    // Scores that are closed may have never been laid out, so we lay them out now
-    for (auto score : masterScore()->scoreList()) {
-        if (!score->isOpen()) {
-            score->doLayout();
-        }
-    }
     for (IExcerptNotationPtr excerptNotation : excerpts) {
         ExcerptNotation* impl = get_impl(excerptNotation);
 
--- a/src/project/internal/exportprojectscenario.cpp
+++ b/src/project/internal/exportprojectscenario.cpp
@@ -56,6 +56,14 @@ bool ExportProjectScenario::exportScores(const INotationPtrList& notations, cons
         return false;
     }
 
+    // Scores that are closed may have never been laid out, so we lay them out now
+    for (INotationPtr notation : notations) {
+        mu::engraving::Score* score = notation->elements()->msScore();
+        if (!score->isOpen()) {
+            score->doLayout();
+        }
+    }
+
     return exportScores(notations, chosenPath, unitType, openDestinationFolderOnExport);
 }
 
--- a/src/engraving/layout/layoutchords.cpp
+++ b/src/engraving/layout/layoutchords.cpp
@@ -602,6 +602,9 @@ double LayoutChords::layoutChords2(std::vector<Note*>& notes, bool up)
             }
         }
         note->setMirror(mirror);
+        if (chord->stem()) {
+            chord->stem()->layout(); // needed because mirroring can cause stem position to change
+        }
 
         // accumulate return value
         if (!mirror) {
--- a/src/engraving/libmscore/arpeggio.cpp
+++ b/src/engraving/libmscore/arpeggio.cpp
@@ -188,6 +188,30 @@ double Arpeggio::calcTop() const
 }
 
 //---------------------------------------------------------
+//   computeHeight
+//---------------------------------------------------------
+
+void Arpeggio::computeHeight(bool includeCrossStaffHeight)
+{
+    Chord* topChord = chord();
+    if (!topChord) {
+        return;
+    }
+    double y = topChord->upNote()->pagePos().y() - topChord->upNote()->headHeight() * .5;
+
+    Note* bottomNote = topChord->downNote();
+    if (includeCrossStaffHeight) {
+        track_idx_t bottomTrack = track() + (_span - 1) * VOICES;
+        EngravingItem* element = topChord->segment()->element(bottomTrack);
+        Chord* bottomChord = (element && element->isChord()) ? toChord(element) : topChord;
+        bottomNote = bottomChord->downNote();
+    }
+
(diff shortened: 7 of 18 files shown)


Tantacrul:
Yeah, this is quite noticeably faster when loading my 'go to' set of orchestral scores.

src/notation/inotation.h
```diff
@@ -96,6 +96,9 @@ class INotation
 
     // notify
     virtual async::Notification notationChanged() const = 0;
+
+    // score
+    virtual mu::engraving::Score* score() = 0;
```
cbjeukendrup, line 101: FWIW, it could already be accessed using `notation->elements()->msScore()`.

src/notation/internal/excerptnotation.cpp
```diff
@@ -133,7 +133,9 @@ void ExcerptNotation::setName(const QString& name)
 
     excerptTitle->setPlainText(name);
     score()->setMetaTag(u"partName", name);
-    score()->doLayout();
+    if (score()->isOpen()) {
+        score()->doLayout();
```
RomanPudashkin, line 137: I think this call can be completely removed from here

src/engraving/utests/utils/scorerw.cpp
```diff
@@ -73,6 +73,7 @@ MasterScore* ScoreRW::readScore(const String& name, bool isAbsolutePath, ImportF
         score = nullptr;
     } else {
         for (Score* s : score->scoreList()) {
+            s->setIsOpen(true);
```
RomanPudashkin, line 76: We should not open scores implicitly here

src/engraving/libmscore/cmd.cpp
```diff
@@ -419,6 +419,9 @@ void Score::update(bool resetCmdState)
         ms->deletePostponed();
         if (cs.layoutRange()) {
             for (Score* s : ms->scoreList()) {
+                if (!s->isOpen() && ms->scoreList().size() > 1) {
```
RomanPudashkin, line 422: I'm wondering if it is possible to call layout only for the current score (which is visible on the screen at the moment)

src/project/internal/exportprojectscenario.cpp
```diff
@@ -56,6 +56,13 @@ bool ExportProjectScenario::exportScores(const INotationPtrList& notations, cons
         return false;
     }
 
+    // Parts that are closed may have never been laid out, so we lay them out now
+    for (INotationPtr notation : notations) {
```
RomanPudashkin, line 60: I would recommend moving this code to MasterNotation::initExcerpts

src/notation/internal/excerptnotation.cpp
mike-spa: done, thanks!

src/engraving/utests/utils/scorerw.cpp
mike-spa: Yeah, I was trying to get the unit tests to succeed. I need to find another solution (and for some reason I can't run them on my machine, so I need to push the commit to the PR and wait for them)

src/project/internal/exportprojectscenario.cpp
mike-spa: done, thanks!

src/engraving/utests/utils/scorerw.cpp
shoogle: Which OS are you on? The unit tests only really work on Linux right now. If you're on Windows you could try building the tests within WSL2, though I wouldn't recommend trying to build/run the GUI that way.

mike-spa: Ah, good to know that, thanks! I'm on windows now, but I do have the linux environment set up on my old laptop, I could try to run the tests there

mike-spa:
u-tests are now all succeeding and the v-tests show the expected result!

oktophonie (approved):

vpereverzev (approved):

vpereverzev:
@mike-spa looks juicy, nice job. Rebase your branch and then I'll merge this

src/notation/internal/masternotation.cpp
```diff
@@ -417,6 +417,12 @@ mu::ValNt<bool> MasterNotation::needSave() const
 
 void MasterNotation::initExcerpts(const ExcerptNotationList& excerpts)
 {
+    // Scores that are closed may have never been laid out, so we lay them out now
+    for (auto score : masterScore()->scoreList()) {
```
RomanPudashkin, line 421: Unfortunately, this change is incorrect. Now we won't layout scores that haven't been opened yet. Let's revert this particular change to your previous version. I apologize for misleading you :(

mike-spa: Actually, for me it seems to work correctly with your first suggestion, as it is now (see video for example, I export the Soprano part without opening it, and it is laid out correctly). In which case exactly do think this will not work?

https://user-images.githubusercontent.com/93707756/196369716-36c9f732-b2f8-4411-8362-eb6ef96914d0.mp4

src/engraving/utests/exchangevoices_tests.cpp
```diff
@@ -104,7 +104,7 @@ TEST_F(Engraving_ExchangevoicesTests, undoChangeVoice)
     // change voice
     score->startCmd();
     score->changeSelectedNotesVoice(1);
-    score->endCmd();
+    score->endCmd(false, /*layoutAllParts = */ true);
```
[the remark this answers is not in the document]
mike-spa: The problem with all the unit tests is that they modify the score, but they never explicitely _open_ the score (meaning, they don't call `Score::setIsOpen()`). But this is not a realistic test case: in reality, as a user, you _have_ to open a score if you want to modify it. Closed scores cannot be modified, so it is not necessary to lay them out when saving. 
So, there are two options to fix the unit tests: a) Explicitely opening the scores before modifying them, or b) Using this flag, which forces the layout of all the parts. I decided to do b) because I thought having the possibility to force a layout of all the parts might be useful for something else in future. However, I can also do a), it will take few minutes. Let me know what you think it's best to do :)

vpereverzev:
> @mike-spa looks juicy, nice job. Rebase your branch and then I'll merge this

actually, @RomanPudashkin has found a couple of blocking issues here, will be looking on the progress of your changes

mike-spa:
@RomanPudashkin @vpereverzev I need some more indiciation on how to proceed.

> Let's revert this particular change to your previous version.

1) Ok on this one ☝️. I have reverted it to the previous version, though I couldn't reproduce the issue you mentioned. For me, your first suggestion seemed to work fine. Anyway, it's a one minute job to do either version, so let me know  :)

> Could you please explain why we need this flag? It looks like this test found a bug: we don't call layout for closed scores before saving them

2) About this one ☝️. The problem with the utests is that they modify the scores without opening them (meaning, the unit tests never do `Score::setIsOpen(true)`). But this in not a realistic test case: in reality you _have_ to open a score to modify it. Closed score can't be modified, so we don't need to lay them out when saving. There are two ways to fix the unit test: a) Open the scores before modifying them, or b) Force the layout of all closed scores using a flag. At the moment I have done b) cause I thought the flag may also be useful in future, but maybe the more correct thing is to do a). I honestly don't know which one is better, so let me know what you think.

src/notation/internal/masternotation.cpp
RomanPudashkin: >In which case exactly do think this will not work?

Just create a score containing several instruments (don't open any of them) and then try to export some of them. None of these instruments will be in masterScore()->scoreList(), so we won't call doLayout() for them

src/engraving/utests/exchangevoices_tests.cpp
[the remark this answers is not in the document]
RomanPudashkin: >But this is not a realistic test case: in reality, as a user, you have to open a score if you want to modify it.

OK, how will it work in this case:  the user creates a score, opens a part, modifies it, and then closes it. After that, the user continues modifying the master score and then saves the whole project. I feel that we won't call doLayout() for the previously opened part

src/notation/internal/masternotation.cpp
mike-spa: Ah! Ok, got it. Now I've tried testing (using the latest version of this PR) a worst-case scenario: score is initially empty, instruments are added from the instruments panel, file has never been saved, and parts have never been opened. When exporting all parts to pdf, they are all correctly laid out (which means doLayout() was called). So I think we're good now.

https://user-images.githubusercontent.com/93707756/196901546-302abd8d-91a5-4582-bfbc-3fcef235f032.mp4

src/engraving/utests/exchangevoices_tests.cpp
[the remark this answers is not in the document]
mike-spa: So, here's what I found after lots of testing. You are totally right, we do not call doLayout() for the previously opened part before saving. As a result, yes, it _is_ possible that the saved xml will contain small mismatches between score and parts. See for example this file [part_OutOfSyncExample.mscx.zip](https://github.com/musescore/MuseScore/files/9831216/part_OutOfSyncExample.mscx.zip): the xml contains an `<accidental>` tag in the score, but not in the part.

However, I think that this is really _not_ a problem. For two reasons.
1) The mismatch can only happen with layout-dependent items, i.e. items that can be different between score and parts (for example: accidentals, beams, custom offsets etc...). The fundamental information of the score (measures, chords, notes, durations, etc) is not layout-dependent, so that will never go out of sync, so we're safe there.
2) Yes, the layout of the part is out of sync, but what can the user do with this part? a) Open it, in which case we call doLayout() and it goes back in sync with the score. b) Export it, in which case we also call doLayout(), and it goes back in sync with the score. (This can be tested on that same file: opening or exporting the Flute part, it will display a b _flat_, correctly). So, practically, the xml captures a specific moment in time where the part was out of sync, but the user will never see that, and that's what matters. In other words: we don't need the part to be in sync _in the xml_, it has to be in sync _when the user needs it_.

So, in the end, I _think_ this is all working as it should. Hope this makes sense!

RomanPudashkin (approved):

src/engraving/utests/exchangevoices_tests.cpp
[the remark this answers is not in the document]
shoogle: I would still consider this to be a mini bug. The current version of MuseScore might be able to deal with this slight inconsistency between the score and the parts, but I don't think there's any guarantee that future versions will deal with this appropriately. It could lead to a "corrupt score" warning dialog in the future, or incorrect behaviour if future versions allow for more differences between the score and the parts than the current version does.

Also, it's a bit of an edge case, but MSCX is an open format so MuseScore might not be the only program reading these files. We shouldn't rely on a quirk of how MuseScore reads scores to allow us to be a bit sloppy about how we write them.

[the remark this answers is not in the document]
shoogle: Anyway, I just thought I'd point out some potential problems here beyond what was already discussed. They appear to be edge cases though so they might be work risking if the performance gain is significant.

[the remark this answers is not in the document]
mike-spa: That's a very valid point, which I also thought about. The way I see it is: whatever change we may do in the future, there is one immutable principle: the master score is always up to date (because it has to be open when you work on a project) and always commands the layout of a part. So I think we'll be safe. In any case, should we realize that this is a problem, reverting this change will also be painless. Let's see how it will go :)

Issue #12900, opened 2022-08-17
[MU4 Issue] cant start play or move cursor to beginning after export to ogg


lyrra:
commit aeea6b1

export to wav or mp3 works.

Tantacrul:
Can't replicate bug because I don't have enough information. Closing for now.

lyrra:
Please reopen: reproducible in mingw. Not reproducible in linux.

Tantacrul:
Can you flesh out the information. Add an image or two. Explain the problem in more detail and add reproduction steps.

We can't solve an issue we only vaguely understand.

lyrra:
The menu panes isn't shown due to it being rendered in its own window, but around the middle of the video I'm doing an export to ogg.





https://user-images.githubusercontent.com/243748/195172134-950f514b-3199-45d2-b339-9d67665ac142.mp4

lyrra:
Fixed in commit c1a6462 , closing.

PR #13726, opened 2022-10-12
bugfix ogg encode

lyrra:
Resolves: #12900 

fixed incorrect usage of opus API


cbjeukendrup:
Rebase needed. Is this (and the issue it's solving) still actual?

lyrra:
Fixed in commit c1a6462 , closing.

PR #13733, opened 2022-10-13
fixed #12465: Crash when using playback toolbar in a score with no instruments

### LINKED ISSUE #12465: [MU4 Issue] Crash when using playback toolbar in a score with no instruments

Eism:
Resolves: #12465

### DIFF
--- a/src/notation/imasternotation.h
+++ b/src/notation/imasternotation.h
@@ -52,6 +52,9 @@ public:
     virtual void setExcerptIsOpen(const INotationPtr excerptNotation, bool opened) = 0;
 
     virtual INotationPartsPtr parts() const = 0;
+    virtual bool hasParts() const = 0;
+    virtual async::Notification hasPartsChanged() const = 0;
+
     virtual INotationPlaybackPtr playback() const = 0;
 };
 
--- a/src/notation/view/noteinputbarmodel.cpp
+++ b/src/notation/view/noteinputbarmodel.cpp
@@ -21,10 +21,12 @@
  */
 #include "noteinputbarmodel.h"
 
-#include "log.h"
+#include "async/notifylist.h"
 #include "types/translatablestring.h"
 
 #include "internal/notationuiactions.h"
+
+#include "log.h"
 
 using namespace mu;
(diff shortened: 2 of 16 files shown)


src/notation/imasternotation.h
```diff
@@ -52,6 +52,8 @@ class IMasterNotation
     virtual void setExcerptIsOpen(const INotationPtr excerptNotation, bool opened) = 0;
 
     virtual INotationPartsPtr parts() const = 0;
+    virtual bool isPartsEmpty() const = 0;
```
RomanPudashkin, line 55: I would suggest renaming it to hasParts

src/notation/view/noteinputbarmodel.cpp
```diff
@@ -186,6 +189,20 @@ void NoteInputBarModel::onNotationChanged()
         undoStack()->stackChanged().onNotify(this, [this]() {
             updateState();
         });
+
+        async::NotifyList<const Part*> partList = masterNotationParts()->partList();
```
RomanPudashkin, line 193: It is better to add somewhere (for example, to NotationParts) one additional notification. It will notify us when the list becomes empty or not. So this code (and other similar blocks) can be rewritten as follows:

```
masterNotationParts()->hasPartsChanged(this, [this]() {
    emit isInputAllowedChanged();
});
```

Tantacrul:
Tested on MacOS. Play bar is appropriately disabled when there are no instruments and I am unable to use a shortcut to make it play. Seems good to go!

DmitryArefiev:
@Tantacrul It's fine about missing full stops? (discussion in #12465)

Tantacrul:
I would just add the full stop. We've done it elsewhere and it's not 'wrong' to do so. Let's also agree to codify it in our style guides.

RomanPudashkin (approved):

Issue #13740, opened 2022-10-13
[MU4 Issue] Crash when trying to notate a rest with articulation


violine1101:
**Describe the bug**
When I try to notate a rest while an articulation toggle is activated (e.g. staccato), MuseScore crashes.

The crash does not happen in all situations. For example, it does not consistently happen if the rest replaces an existing note. However, the reproduction steps below appear to be reliable.

**To Reproduce**
1. Create an empty score
2. Enable note input mode (`N`)
3. Enable staccato (`Shift`+`S`)
4. Place a (quarter note) rest (`0`)
5. MuseScore crashes

**Expected behavior**
MuseScore should not crash.

**Video**
https://user-images.githubusercontent.com/12451842/195648366-b81945fe-8730-4fbd-989b-f155a8deaf19.mp4

**Platform information**
 - OS: Windows 10
 - MuseScore version: Tested on both the public alpha, and nightly version `2022-10-10 04:55`

lyrra:
confirm crash on linux
<details><summary>backtrace</summary>

```
#3  0x00007ffff5684662 in __GI___assert_fail (assertion=0x555557d8fd1e "e == 0 || e->isChord()", file=0x555557d8f908 "src/engraving/libmscore/engravingobject.h", line=640, 
    function=0x555557d8fcd0 "mu::engraving::Chord* mu::engraving::toChord(mu::engraving::EngravingObject*)") at assert.c:101
#4  0x0000555555c7c519 in mu::engraving::toChord (e=0x555561312d70) at src/engraving/libmscore/engravingobject.h:640
#5  0x0000555555c81846 in mu::engraving::Score::setNoteRest (this=0x55555cdba3d0, segment=0x55556131fd40, track=0, nval=..., sd=..., stemDirection=mu::engraving::DirectionV::AUTO,
    forceAccidental=false, articulationIds=std::set with 1 element = {...}, rhythmic=false, externalInputState=0x0) at src/engraving/libmscore/cmd.cpp:915
#6  0x0000555555cc2660 in mu::engraving::Score::enterRest (this=0x55555cdba3d0, d=..., externalInputState=0x0) at src/engraving/libmscore/edit.cpp:3559
#7  0x0000555555cc22f1 in mu::engraving::Score::cmdEnterRest (this=0x55555cdba3d0, d=...) at src/engraving/libmscore/edit.cpp:3538
```

</details>

PR #13746, opened 2022-10-13
bugfix: check if chord if updateArticulations

lyrra:
Resolves: #13740

This is a naive fix, as it will only avoid a crash.

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -912,7 +912,9 @@ Segment* Score::setNoteRest(Segment* segment, track_idx_t track, NoteVal nval, F
             tick += ncr->actualTicks();
 
             if (!articulationIds.empty()) {
-                toChord(ncr)->updateArticulations(articulationIds);
+                if (ncr->isChord()) {
+                    toChord(ncr)->updateArticulations(articulationIds);
+                }
             }
         }
 


cbjeukendrup:
> This is a naive fix, as it will only avoid a crash.

I think it is actually okay; rests with articulations are simply not supported, and so it is okay to simply ignore articulations when entering a rest.

src/engraving/libmscore/cmd.cpp
```diff
@@ -912,7 +912,9 @@ Segment* Score::setNoteRest(Segment* segment, track_idx_t track, NoteVal nval, F
             tick += ncr->actualTicks();
 
             if (!articulationIds.empty()) {
-                toChord(ncr)->updateArticulations(articulationIds);
+                if (ncr->isChord()) {
+                    toChord(ncr)->updateArticulations(articulationIds);
```
cbjeukendrup, line 916: Alternatively: do this a few lines higher directly when the chord is created, so that you can avoid the `ncr->isChord()` check and use the `chord` variable instead.

For example, around line 892:
```c++
if (!articulationIds.empty()) {
    chord->updateArticulations(articulationIds);
}
```

Maybe that solution looks a little bit cleaner.

RomanPudashkin (approved):

PR #13749, opened 2022-10-13
Fix drawing barlines in Palettes

cbjeukendrup:
This was broken relatively recently.

|Before|After|
|-|-|
|Scherm­afbeelding 2022-10-13 om 22 38 52|Scherm­afbeelding 2022-10-13 om 22 36 57|

### DIFF
--- a/src/engraving/libmscore/barline.h
+++ b/src/engraving/libmscore/barline.h
@@ -69,8 +69,8 @@ class BarLine final : public EngravingItem
     int _spanFrom           { 0 };         // line number on start and end staves
     int _spanTo             { 0 };
     BarLineType _barLineType { BarLineType::NORMAL };
-    mutable double y1;
-    mutable double y2;
+    mutable double y1 = 0.0;
+    mutable double y2 = 0.0;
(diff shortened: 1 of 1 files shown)


mike-spa (approved):

src/engraving/libmscore/barline.cpp
```diff
@@ -1276,7 +1276,7 @@ void BarLine::layout()
     // after staff distances are known). This is a temporary layout.
     double _spatium = spatium();
     y1 = _spatium * .5 * _spanFrom;
-    if (RealIsEqual(y2, 0.0)) {
+    if (RealIsEqual(y2, 0.0) || score()->isPaletteScore()) {
```
RomanPudashkin, line 1279: Could you please explain why we need this `score()->isPaletteScore()` here? I'm not a big fan of them. Too much knowledge of palettes in libmscore

cbjeukendrup: @RomanPudashkin I found an alternative solution that doesn't involve anything palette-specific. Probably this is a better solution anyway.

RomanPudashkin (approved):

Issue #12869, opened 2022-08-15
[MU4 Issue] Pedal playback not working for "pedal change"


MarcSabatella:
**Describe the bug**
The standard "pedal change" marking \_\_\_/\\\_\_\_ does not work; it lifts the pedal off but does not return it to the on position

**To Reproduce**
1. Piano score
2. Add four quarter notes C D E D to first measure and again to the second
3. Add a |___/ pedal marking from the first C to the second
4. Add a \\___| pedal marking from the second C to the end of the measure
5. Play

*Result**
Pedal lifts in bar 2 but does not re-engage

**Expected behavior**
Second measure should have pedal engaged just like the first

**Screenshots**
![image](https://user-images.githubusercontent.com/1799936/184670289-e552b221-9a5b-4f63-9f0c-1c33fdb3d5fc.png)

**Platform information**
 - OS: Linux

**Additional context**
The old-style Ped___* markings work fine because they don't overlap.  The issue seems to only occur on the pedal changes - the places where the end of one pedal coincides the the start of the next to create the characteristic pedal change \_\_\_/\\\_\_\_.  If I were to guess, I'd say the pedal off and on messages are simply being sent in the wrong order.

MarcSabatella:
Bump - could this get prioritized?

Tantacrul:
Yeah, I've been noticing this myself too.

MarcSabatella:
FWIW, this is now the only significant issue remaining with "Reunion" - https://musescore.com/user/2975/scores/39069

Tantacrul:
Just had a call with Vasily an hour ago, where we placed this issue in our top 3. Won't be much longer

FireController1847:
I'm still having issues with pedal changes on this front, particularly when migrating from old versions of MuseScore. Notes are cut off early, general playback with MS Sounds is unreliable and unclean. Every once in a while a note goes max velocity as well. I can send example files but the songs are copyrighted so I'll have to email them or send them through other means.

Ander3232:
Any progress with this? It seems like a pretty basic thing. And how do you include the "Ped." text symbol at the beginning of the extended line?

MarcSabatella:
The issue reported here was fixed a year ago, which is why this issue is marked “closed”.  If you are having issues using a current version of MuseScore, that is probably something different.  Please open a new thread and ZIP and attach the score you are having trouble with and give precise steps to reproduce the problem.

[...]

PR #13759, opened 2022-10-14
[MU4] pedal_change

vpereverzev:
Resolves: https://github.com/musescore/MuseScore/issues/12869

### DIFF
--- a/src/engraving/libmscore/spannermap.cpp
+++ b/src/engraving/libmscore/spannermap.cpp
@@ -22,6 +22,7 @@
 
 #include "spannermap.h"
 #include "spanner.h"
+#include "part.h"
 
 #include "log.h"
 
@@ -45,11 +46,13 @@ SpannerMap::SpannerMap()
 
 void SpannerMap::update() const
 {
-    std::vector<interval_tree::Interval<Spanner*> > intervals;
-    for (auto i : *this) {
-        intervals.push_back(interval_tree::Interval<Spanner*>(i.second->tick().ticks(), i.second->tick2().ticks(), i.second));
-    }
-    tree = interval_tree::IntervalTree<Spanner*>(intervals);
+    IntervalList regularIntervals;
+    IntervalList collisionFreeIntervals;
+
+    collectIntervals(regularIntervals, collisionFreeIntervals);
+
+    tree = interval_tree::IntervalTree<Spanner*>(regularIntervals);
+    collisionFreeTree = interval_tree::IntervalTree<Spanner*>(collisionFreeIntervals);
     dirty = false;
 }
 
@@ -57,13 +60,20 @@ void SpannerMap::update() const
 //   findContained
 //---------------------------------------------------------
 
-const std::vector<interval_tree::Interval<Spanner*> >& SpannerMap::findContained(int start, int stop) const
+const SpannerMap::IntervalList& SpannerMap::findContained(int start, int stop, bool excludeCollisions) const
 {
     if (dirty) {
         update();
     }
+
     results.clear();
-    tree.findContained(start, stop, results);
+
+    if (excludeCollisions) {
+        collisionFreeTree.findContained(start, stop, results);
+    } else {
+        tree.findContained(start, stop, results);
+    }
+
     return results;
 }
 
@@ -71,14 +81,65 @@ const std::vector<interval_tree::Interval<Spanner*> >& SpannerMap::findContained
 //   findOverlapping
 //---------------------------------------------------------
 
-const std::vector<interval_tree::Interval<Spanner*> >& SpannerMap::findOverlapping(int start, int stop) const
+const SpannerMap::IntervalList& SpannerMap::findOverlapping(int start, int stop, bool excludeCollisions) const
 {
     if (dirty) {
         update();
     }
+
     results.clear();
-    tree.findOverlapping(start, stop, results);
+
+    if (excludeCollisions) {
+        collisionFreeTree.findOverlapping(start, stop, results);
+    } else {
+        tree.findOverlapping(start, stop, results);
+    }
+
     return results;
+}
+
+void SpannerMap::collectIntervals(IntervalList& regularIntervals, IntervalList& collisionFreeIntervals) const
+{
+    using IntervalsByType = std::map<ElementType, IntervalList>;
+    using IntervalsByPart = std::map<ID, IntervalsByType>;
+
+    IntervalsByPart intervalsByPart;
+
+    //!Note Because of the current UX of spanners adjustments spanners collision is a regular thing,
+    //!     so we have to manage those cases when two similar spanners (e.g. Pedal line) are overlapping
+    //!     with each other.
+    constexpr int collidingSpannersPadding = 1;
+
+    for (const auto& pair : *this) {
+        int newSpannerStartTick = pair.second->tick().ticks();
+        int newSpannerEndTick = pair.second->tick2().ticks();
+
+        IntervalsByType& intervalsByType = intervalsByPart[pair.second->part()->id()];
+        IntervalList& intervalList = intervalsByType[pair.second->type()];
+
+        if (!intervalList.empty()) {
+            auto lastIntervalIt = intervalList.rbegin();
+            if (lastIntervalIt->stop >= newSpannerStartTick) {
+                lastIntervalIt->stop = newSpannerStartTick - collidingSpannersPadding;
+            }
+        }
+
+        intervalList.emplace_back(interval_tree::Interval<Spanner*>(newSpannerStartTick,
+                                                                    newSpannerEndTick,
+                                                                    pair.second));
+
+        regularIntervals.push_back(interval_tree::Interval(newSpannerStartTick,
+                                                           newSpannerEndTick,
+                                                           pair.second));
+    }
+
+    for (const auto& pair : intervalsByPart) {
+        for (const auto& intervals : pair.second) {
+            collisionFreeIntervals.insert(collisionFreeIntervals.end(),
+                                          intervals.second.begin(),
+                                          intervals.second.end());
+        }
+    }
 }
 
 //---------------------------------------------------------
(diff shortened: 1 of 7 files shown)


RomanPudashkin (approved):

Tantacrul:
Just tested this for Fluid only. Working well on the scores I tested! Great stuff.

Tantacrul:
@MarcSabatella - it seems that for both you and I, this is the final fix that makes a complex test score of mine now play back perfectly.

MarcSabatella:
Is that horn change really relevant?  Anyhow, I tested, and unfortunately it isn't working for me on Reunion - I think because of the dynamics, because when I remove them, it works.  Also, the pedal markings now disappear on save/reload.  And the end hooks is missing if it crosses systems.

MarcSabatella:
Actually, not just the end hook - the pedal line doesn't continue across syste,s at all, if there is another starting at the same place:

image

(note the missing continuation of the pedal line going into the last system)

MarcSabatella:
More relevant info - if I add the pedal lines back using last night's build then load into the artifact from this one, it does mostly work even with the dynamics, but, the last note of the first measure of the pedal passage getds chopped off.  That happens in the nightly too, if I try shortening the line to not overlap.

MarcSabatella:
Here's the actual score I am testing (just rename to ".mscz").  Unlike the "official" version on musescore.com, this one was created from scratch in a recent MU4 build.  You'll need to add the pedal lines back in because they were lost on the save again.
[Reunion-40-scratch.zip](https://github.com/musescore/MuseScore/files/9787850/Reunion-40-scratch.zip)

Tantacrul:
I'm getting the strangest error @MarcSabatella 
image

MarcSabatella:
This *is* the file, I just renamed it from msc to zip so GitHub would accept it.  Just rename it back - no need to use zip at all.

Tantacrul:
> This _is_ the file, I just renamed it from msc to zip so GitHub would accept it. Just rename it back - no need to use zip at all.

I did rename it and I'm getting this bizarre error.

MarcSabatella:
As I continue to experiment, I get all sorts of kind of randomy different results, but none of them amount ot it actually working correctly.  Mostly, I always see the non-continuation, I always hear the last two measure not pedal, and I usually hear the first measure cut off too soon.  Maybe the tempo changes also somehow get in the way?  Or the arpeggios, or the haripins, or the slurs.  Anyhow, if you're not hearing exactly what I described, no big surprise.

Tantacrul:
I think I got the problem. It's the rallentando in Reunion. I've replicated it here:
https://www.dropbox.com/s/3pzst6nncxdhgly/TEMPO_LINE_VS_PEDAL_LINE.mscz?dl=0

MarcSabatella:
Yep, I think you're right.  Since the pedal lines were lost on the save, I guessed on adding them back in (every half note?).

BTW, it's all but the last pedal line that gets lost on save.

vpereverzev:
@MarcSabatella @Tantacrul  there is a new fix coming on. Please check it again

MarcSabatella:
The more recent build fixes the engraving issue and the saving issue, but it still fails on playback using the score I attached earlier.

vpereverzev:
> The more recent build fixes the engraving issue and the saving issue, but it still fails on playback using the score I attached earlier.

Yea, @Tantacrul was right. There was a separate issue whith gradual tempo change (which is a spanner) clashing with other spanners, like pedal line. Just pushed some changes here

cbjeukendrup:
@vpereverzev I think all those other horns need to be adjusted as well, from `"a-horn"` up to `"c-horn"`. (Maybe others too? but those are already distinguished by SoundSubCategory::Alto or Bass.)

Another thought (this one is rather about semantics): since English Horn is a completely different instrument, and has `SoundId::Oboe` (which seems good to me), I suggest that simply `{ SoundId::Horn, SoundCategory::Winds, {}, {} }` should point to the (French) horn sound anyway. (Just like in real life: if you say simply "horn", you're much more likely to mean French horn than English horn.) But I don't know whether we can decide about this, or that it is part of the MPE standard.

vpereverzev:
> @vpereverzev I think all those other horns need to be adjusted as well, from `"a-horn"` up to `"c-horn"`. (Maybe others too? but those are already distinguished by SoundSubCategory::Alto or Bass.)
> 
> Another thought (this one is rather about semantics): since English Horn is a completely different instrument, and has `SoundId::Oboe` (which seems good to me), I suggest that simply `{ SoundId::Horn, SoundCategory::Winds, {}, {} }` should point to the (French) horn sound anyway. (Just like in real life: if you say simply "horn", you're much more likely to mean French horn than English horn.) But I don't know whether we can decide about this, or that it is part of the MPE standard.

we'll be updating MuseSounds mappings separately, don't involve them in this discussion. That's entire different and bloody exhausting story :D

MarcSabatella:
Getting closer for sure!

Still, though, on that same score, the last three notes of bar 17 (the first measure with the pedal), as well as the entirety of the last two bars, don't sustain.  Try setting the "meno mosso" marking in bar 17 to 60 BPM or so, and listen with headphones.  You can clearly hear the sustain stop aroiund when the fourth note "B" plays in the LH in bar 17 (and then resume in bar 18).  And again, there is no sustain whatsoever on the big cross-staff arpeggio.  That one works if I remove the pedal line in the previous bar, or if I remove the rallentando.

It's also still the case that if I delete the pedal lines and add them back, they don't play at all after the first three notes of bar 17 - none of the remaining bars sustain at all.  Not unless I then do something else to trigger the playback info to be recalculated, like select the passage then press Ctrl+Down followed by Ctrl+Up.  So it seems it (partially) works on score load, but not immediately on adding the pedal, unless you do something else to force it to happen.

Not sure if it's coincidence or not, but there is also an odd effect in bar 19 - again, best experienced at a slower tempo, and through headphones.  Although both bar 17 and 19 have diminuendos, there is a very noticeable *increase* in volume half way through bar 19.  I'm guessing that's about the samples themselves.  But somehow it seems to me they also increase the volume of the previously sustain notes, if the pedal is on.  Maybe somehow that relates to the dropping of the sustain in bar 17 too?

And - do the horn changes really have anything to do with this?  Just wondering why they are part of this PR.

Tantacrul:
I've tested again using reunion and re-writing the passage myself to isolate the remaining issue. It is clear to me now that this happens when **the end of a tempo line meets the beginning of a pedal line**.

[Link to new score demonstrating the issue
](https://www.dropbox.com/s/pw7u5n9z5dhvhoh/TEMPO_LINE_VS_PEDAL_LINE_V2.mscz?dl=0)

It's a bit tricky to understand visually because the snapping works differently. Here the **rall** ends at the end of the measure and the pedal line initiates on the first beat.

So I'm not sure if they are on the same beat or not... or whether they're just really close together.
image

Regardless, we need to replace this entire system in future releases!

vpereverzev:
> I've tested again using reunion and re-writing the passage myself to isolate the remaining issue. It is clear to me now that this happens when **the end of a tempo line meets the beginning of a pedal line**.
> 
> [Link to new score demonstrating the issue ](https://www.dropbox.com/s/pw7u5n9z5dhvhoh/TEMPO_LINE_VS_PEDAL_LINE_V2.mscz?dl=0)
> 
> It's a bit tricky to understand visually because the snapping works differently. Here the **rall** ends at the end of the first measure and the pedal line initiates on the first beat of the second measure.
> 
> So I'm not sure if they are on the same beat or not... or whether they're just really close together. image
> 
> Regardless, we need to replace this entire system in future releases!

Did you tested this on the latest build? There shouldn't be clash with tempo lines anymore. Although I've found an oppiste issue where lines can clash with other lines from **previous measures** because of the "anchors"

MarcSabatella:
The most recent build I see is [MU4_3253434985_Lin_13759_[MU4]](https://github.com/musescore/MuseScore/suites/8786017110/artifacts/398963764).  Thus one definitely fails on Reunion.  Be sure to use the version I attach to the this thread - that's the one created within MuseScore 4 using gradual tempo changes.  The version on musecore.com was created in an older version and thus fakes the gradual tempo changes using normal tempo text.

For me it has exactly the problems I stated above:

1) the last three notes of bar 17 do not sustain
2) the last two bars do not sustain
3) if I delete the pedal lines and re-add them, they do not play at all after the first three notes of bar 17, until I do something to force playback info to be updated

I re-downloaded the score I attached earlier to be sure I'm testing the same score.  Here is the direct link:

https://github.com/musescore/MuseScore/files/9787850/Reunion-40-scratch.zip

Tantacrul:
@vpereverzev I tested it on this PR this morning. I actually downloaded it twice to be sure. Unless there's something I'm missing?

vpereverzev:
> @vpereverzev I tested it on this PR this morning. I actually downloaded it twice to be sure. Unless there's something I'm missing?

Just pushed the latest changes. The case with tempo lines should be fixed

cbjeukendrup:
@vpereverzev Look at this annoying example... [Breaking_SpannersMetaParser_spannerDuration.mscz.zip](https://github.com/musescore/MuseScore/files/9802893/Breaking.SpannersMetaParser.spannerDuration.mscz.zip)
Scherm­afbeelding 2022-10-17 om 18 49 45

Of course, this is just an artificial example constructed to break the `SpannersMetaParser::spannerDuration` method, but maybe more natural examples would be broken too.

vpereverzev:
> @vpereverzev Look at this annoying example... [Breaking_SpannersMetaParser_spannerDuration.mscz.zip](https://github.com/musescore/MuseScore/files/9802893/Breaking.SpannersMetaParser.spannerDuration.mscz.zip) Scherm­afbeelding 2022-10-17 om 18 49 45
> 
> Of course, this is just an artificial example constructed to break the `SpannersMetaParser::spannerDuration` method, but maybe more natural examples would be broken too.

There is still a number of issues which we can't really properly fix so far. As we discussed with @Tantacrul we'd make a fix for the originally mentioned problems in the issue attached. As for the rest of them.. there is a couple of dirty workarounds and ideas I'd try to implement later on, after all the high-priority issues

PR #13763, opened 2022-10-14
Fix #13725: Note input on percussion staff with only one selectable element in the drumset palette should pre-select this element

### LINKED ISSUE #13725: Note input on percussion staff with only one selectable note should pre-select this note
For example, I'm adding a "finger snap" instrument staff to my score.

Selecting a bar on the finger snap staff and entering input mode (key "n") I get a note hint when I move the mouse cursor over the staff's rests, and when I click I even hear the finger snap sound, but no note appears.

It is not intuitively clear that I must first click the only available note in the drumset notation section at the lower left to be able to add notes to this staff. As it is the only available note in this case it could be selected by default when entering input mode. This saves the experienced user a click and the inexperienced user some head-scratching and a few google searches to find the solution (an old post from 2010: https://musescore.org/en/node/4805 redirecting to the Drumset notation help page).

The second best solution would be to somehow indicate that the user must choose a note from the drumset notation section first.

It may be a minor issue but oftentimes the little things matter as far as usability is concerned. A related issue is that when input mode is exited MuseScore forgets the previously selected percussion note on re-entry. From a usability standpoint it might be better to keep the previous selection (on a per-staff basis).

These issues affect the latest version 3 release as well as the current state on the master branch (OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-, revision: github-musescore-musescore-)

leomeyer:
Resolves: [issue 13725](https://github.com/musescore/MuseScore/issues/13725):
Note input on percussion staff with only one selectable note should pre-select this note

Can be tested by adding a "Slap" instrument staff, selecting the first rest and pressing "n".
The only available drumset palette element is automatically selected (no sound is played).
On a percussion staff with more than one possible notes (e. g. Drumset) there is no automatic pre-selection.
I have added a reference to the issue as a code comment. Please remove it as desired.

As well, the following usability improvements have been made:
Remember last entered drum note for each percussion staff in the score (also, save it to the .mscx file). If a note is selected when entering note input mode this note becomes the currently preferred note and is remembered. If there is only one drum note it is always selected. Otherwise, when entering note input for the first time on a percussion staff no note is selected.


leomeyer:
Please note that there are additional commits that are an attempt at pre-selecting the drum note depending on the last input and/or selection (sorry, I forgot to squash them...)

leomeyer:
I feel that I have to think about the drum input a little more.

Issue #12815, opened 2022-08-12
[MU4 Issue] Crash when closing a score after disabling/enabling mixer several times


mattblo:
**Describe the bug**
Musescore 4 crashes when the score is closed, both with and without saving. Before that, I think the mixer has to be enabled and disabled several times. At least that's how I was able to reproduce it.

**To Reproduce**
1. Open one of those score for example
[test scores.zip](https://github.com/musescore/MuseScore/files/9314987/test.scores.zip)
2. See the video

**Expected behavior**
No crash!

**Screenshots**

https://user-images.githubusercontent.com/63284090/184318881-bbfdec45-8496-4e40-88aa-eb81d3ad0228.mp4

**Platform information**
 - OS: Windows 11
 
 **Additional information**
 - Musescore 4 has already been reset to factory settings before the video

Tantacrul:
Making a P1 because I'm worried it may lead to occasional crashes when someone opens / closes the mixer only once.

Issue #13764, opened 2022-10-14
[MU4 Issue] MuseScore quits instead of just closing score


HemantAntony:
**Describe the bug**
When you press `Ctrl + W` to close the open score, it 

**To Reproduce**
1. Open any score
2. Press `Ctrl + W`
3. MuseScore quits

**Expected behavior**
Only the score should close instead of the whole app

**Screenshots**

https://user-images.githubusercontent.com/88136054/195892448-f3472933-134a-4cc2-ad52-782e6d1391bc.mp4

**Platform information**
 - OS: Ubuntu 22.04 LTS

**Additional context**
For the Devs, this problem occurs because there is more than one instance of the app open. Hence [`quitApp`](https://github.com/musescore/MuseScore/blob/7ff90ee56c584c9916291ad84d0a70f9baa2c903/src/project/internal/projectactionscontroller.cpp#L54) becomes true, hencing forcing MU4 to quit. I would've attempted to fix this. But this lead to bunch of things about IpcSocket (It sounds too complicated :) )

Also, if no one can reproduce this, ask for any other info you might need

HemantAntony:
This might be related to #13698. However, in that issue it states that the issue is a crash. Hence I thought it might be a different issue. But they both do look similar

Eism:
@HemantAntony 
Yes, it's the same bug(#13698)
Please check the build from #13774

Eism:
Ah, I missed additional information.

> For the Devs, this problem occurs because there is more than one instance of the app open.

This is expected behaviour. See https://github.com/musescore/MuseScore/issues/9966

HemantAntony:
@Eism but that behaviour is only when there are more than 1 projects open. This issue occurs when only one project is open. So MuseScore should not close

The bug is that there are two app instances in the code when in reality there is only one

Eism:
@HemantAntony make sure you have only one instance open.
I can't reproduce

HemantAntony:
I only do have one instance open

If no one can reproduce, I guess I will try to figure out what the problem (emphasis on try :) )

AntonioBL:
Ciao.
I managed to reproduce it most of times with these steps:
1- Open MuseScore
2- Open a recent score ("score1") from the dashboard
3- Open another recent score ("score2") from the "recent files" menu
4- close score2 with `Ctrl + w`
5- close score1 with `Ctrl + w` -> MuseScore quits
[...]

PR #13774, opened 2022-10-15
fixed #13698: MuseScore crashes after closing score

### LINKED ISSUE #13698: MU4 crash in Linux
**Description:**
Today I experienced a crash running the Nightly build from Oct, 10th.
Upon closing the open file it crashed (after refusing to save changes to the open file) with the following message:

/tmp/.mount_MuseSc7ltvD2/AppRun: Zeile 26: 44217 Speicherzugriffsfehler "${APPDIR}/bin/mscore-portable-nightly" "$@"

OS: Ubuntu 20.04
Appimage
=== Started MuseScore 4.0.0-Development, build number 3216906122 ===

Logfile from that session attached.

**To Reproduce**
1. Open first score (Pachelbel)
2. Play it by clicking the play button
3. While playing open the mixer (F10)
4. Press CTRL-S to save the score
5. Close the score (click the X)
6. Open second score (Schumann from 3.6.0)
7. Play it by clicking the play button
8. Close the score by clicking the X and click "Don't save"
9. Crash (app disappeared completely)

I could reproduce the crash once but no luck afterwards so I am not sure whether the steps above lead to the crash. The only sure thing is step 9 (closing the changed score from Schumann).

[MS4.zip](https://github.com/musescore/MuseScore/files/9756126/MS4.zip)

Klaus

Eism:
Resolves: #13698
Resolves: #12815

### DIFF
--- a/src/framework/global/thirdparty/deto_async/async/internal/abstractinvoker.cpp
+++ b/src/framework/global/thirdparty/deto_async/async/internal/abstractinvoker.cpp
@@ -55,9 +55,15 @@ void AbstractInvoker::invoke(int type, const NotifyData& data)
 void AbstractInvoker::invokeCallback(int type, const CallBack& c, const NotifyData& data)
 {
     assert(c.threadID == std::this_thread::get_id());
+
+    if (!containsReceiver(c.receiver)) {
+        return;
+    }
+
     if (c.receiver && !c.receiver->isConnectedAsync()) {
         return;
     }
+
     doInvoke(type, c.call, data);
 }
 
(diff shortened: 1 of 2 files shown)


HemantAntony:
@Eism My issue (#13764) still occurs

DmitryArefiev:
Tested #13698 on Win10, Mac12, **LinuxMint 20.2** - **FIXED**

src/framework/global/thirdparty/deto_async/async/internal/abstractinvoker.cpp
```diff
@@ -192,3 +198,16 @@ void AbstractInvoker::removeQInvoker(QInvoker* qi)
     std::lock_guard<std::mutex> lock(m_qInvokersMutex);
     m_qInvokers.remove(qi);
 }
+
+bool AbstractInvoker::containsReceiver(Asyncable* receiver) const
```
vpereverzev, line 202: Feels like this can hit the performance?

One other thing, I thought we're removing the necessary callbacks at destruction point of a receiver? @igorkorsukov FYI

Eism: I implemented this solution after discussing with Igor, we came to the conclusion that this is the best solution at the moment. Usually, an Invoker doesn't have many receivers.

> One other thing, I thought we're removing the necessary callbacks at destruction point of a receiver?

Yes, we remove all callbacks in the destructor of the receiver. The problem is that we delete the receiver after passing this receiver to the Invoker. At the time of the creation of the Invoker, the receiver was valid; when the callback was called, the receiver is destroyed.

RomanPudashkin (approved):

Issue #13281, opened 2022-09-10
[MU4 Issue] Apply current palette element command not implemented


MarcSabatella:
**Describe the bug**
MuseScore 3 had a command "apply current palette element" to easily reapply the most recently-used palette element.  This was a great usability feature and especially valuable for blind users even though it had no default shortcut.  This command is missing in MU4.

**To Reproduce**
1. Go to Edit / Preferences / Shortcuts
2. Search for "palette"
3. See error: no "Apply current palette element" command available

**Expected behavior**
The "apply current palette element" should be available, and ideally, should have a default shortcut assigned even though there was none in MU3.  FWIW, I normally used "\\".

**Platform information**
 - OS: Linux

**Additional context**
I know there is some new functionality coming for shortcuts via the GSoC project, but on the assumption that none of the pending work for that is being targeted for 4.0, re-implementing this one command would be a big help for now.

There are a handful of other commands that were dropped from MU3, some consciously, some perhaps less so.  This to me is the one most worth adding back for 4.0.

wizofaus:
You normally used ""?
There is still "applyCurrentPaletteElement" in PalettesPanel.qml but nothing calling it...hard to judge whether it will actually work.

MarcSabatella:
Sorry, I typed a backslash but GitHub ate it.  Fixed now.

MarcSabatella:
I can make an attempt, if I can get a little help.  I knew how things worked in MU3 but it still is a lot of trial and error for me in MU4.

@wizofaus - I assume you're talking about the code here:

https://github.com/musescore/MuseScore/blob/master/src/palette/qml/MuseScore/Palette/PalettesPanel.qml#L46-L47

It looks like the actual work then is done here:

https://github.com/musescore/MuseScore/blob/master/src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml#L111-L120

which looks about right to me (not that I know my way QML at all).  So, it seems promising.

In theory, then, all we'd need to do is make sure the command exists and calls the applyCurrentPaletteElement function.  The place where the command would be to be defined, I assume, is here:

https://github.com/musescore/MuseScore/blob/master/src/palette/qml/MuseScore/Palette/PalettesPanel.qml#L46-L47

And then after defining tyhe command itself there, I'd need to register it here:

https://github.com/musescore/MuseScore/blob/master/src/notation/internal/notationactioncontroller.cpp

[...]

PR #13779, opened 2022-10-15
Apply current palette element (DRAFT)

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/13281

Add handler for apply-current-palette-element

### DIFF
--- a/src/palette/internal/paletteuiactions.cpp
+++ b/src/palette/internal/paletteuiactions.cpp
@@ -44,6 +44,12 @@ const UiActionList PaletteUiActions::m_actions = {
              TranslatableString("action", "Palette search"),
              TranslatableString("action", "Search palettes")
              ),
+    UiAction("apply-current-palette-element",
+             mu::context::UiCtxNotationOpened,
+             mu::context::CTX_ANY,
+             TranslatableString("action", "Apply current palette element"),
+             TranslatableString("action", "Apply current palette element")
+             ),
     UiAction("time-signature-properties",
              mu::context::UiCtxNotationOpened,
              mu::context::CTX_NOTATION_OPENED,
--- a/src/palette/internal/paletteuiactions.cpp
+++ b/src/palette/internal/paletteuiactions.cpp
@@ -46,7 +46,7 @@ const UiActionList PaletteUiActions::m_actions = {
              ),
     UiAction("apply-current-palette-element",
              mu::context::UiCtxNotationOpened,
-             mu::context::CTX_ANY,
+             mu::context::CTX_NOTATION_OPENED,
              TranslatableString("action", "Apply current palette element"),
              TranslatableString("action", "Apply current palette element")
              ),
--- a/src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml
+++ b/src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml
@@ -116,6 +116,9 @@ StyledListView {
             // in search and no cell selected: apply the first found element
             const parentIndex = paletteTreeDelegateModel.modelIndex(0);
             index = paletteModel.index(0, 0, parentIndex);
+        } else if (!index.valid) {
+            // no cell selected and not in search: use current index from previous click
+            index = currentIndex;
         }
 
         paletteController.applyPaletteElement(index, Qt.NoModifier);
--- a/src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml
+++ b/src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml
@@ -117,8 +117,8 @@ StyledListView {
             const parentIndex = paletteTreeDelegateModel.modelIndex(0);
             index = paletteModel.index(0, 0, parentIndex);
         } else if (!index.valid) {
-            // no cell selected and not in search: use current index from previous click
-            index = currentIndex;
+            // no cell selected: use current item index
+            index = currentItem.modelIndex;
         }
 
         paletteController.applyPaletteElement(index, Qt.NoModifier);
--- a/src/palette/qml/MuseScore/Palette/PalettesPanel.qml
+++ b/src/palette/qml/MuseScore/Palette/PalettesPanel.qml
@@ -65,6 +65,10 @@ Item {
         onPaletteSearchRequested: {
             palettesPanelHeader.startSearch()
         }
+        
+        onApplyCurrentPaletteElementRequested: {
+            root.applyCurrentPaletteElement()
+        }
     }
 
     ColumnLayout {
--- a/src/palette/qml/MuseScore/Palette/internal/Palette.qml
+++ b/src/palette/qml/MuseScore/Palette/internal/Palette.qml
@@ -526,7 +526,7 @@ StyledGridView {
             }
 
             onClicked: {
-                if (!paletteView.paletteController.applyPaletteElement(paletteCell.modelIndex, ui.keyboardModifiers())) {
+                if (paletteView.paletteController.applyPaletteElement(paletteCell.modelIndex, ui.keyboardModifiers())) {
                     updateSelection()
                 }
             }
--- a/src/palette/view/paletterootmodel.cpp
+++ b/src/palette/view/paletterootmodel.cpp
@@ -29,6 +29,9 @@ PaletteRootModel::PaletteRootModel(QObject* parent)
     dispatcher()->reg(this, "palette-search", [this]() {
         emit paletteSearchRequested();
     });
+    dispatcher()->reg(this, "apply-current-palette-element", [this]() {
+        emit applyCurrentPaletteElementRequested();
+    });
 }
 
 PaletteRootModel::~PaletteRootModel()
--- a/src/palette/view/paletterootmodel.h
+++ b/src/palette/view/paletterootmodel.h
@@ -49,6 +49,7 @@ public:
 
 signals:
     void paletteSearchRequested();
+    void applyCurrentPaletteElementRequested();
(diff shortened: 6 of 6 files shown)


cbjeukendrup:
Forgot to commit paletterootmodel.h? The signal `applyCurrentPaletteElementRequested` needs to be declared there.

MarcSabatella:
Thanks for the pointer on the signal declaration; I found there to do that just as you were posting.  I had meanwhile been wondering if somehow QML did that automatically, like how it seems to automatically associate the "on" handler with the signal itself.

Anyhow, I don't have a build system right now, so I'm just creating this using GitHub online, which seems to limit me to one file per commit.  Hopefully when I'm done I'll figure out how to squash - assuming this all actually works.  If the actual handling code in PaletteTree.qml doesn't do what it needs to here, I'm probably out of luck in solving this myself.

Thanks again.  FWIW, to my knowledge there is no MU4-specific code of mine anywhere in the application, so that's a dual incentive for me wanting to get this in.  But mostly, as I said in the issue, it feels like the biggest single missing feature to me.  I really missed this command trying to add pedal markings over and over while testing the pedal playback PR!

MarcSabatella:
... and, unfortunately, it doesn't work.  Console shows me the command *is* being called, but beyond that, I can't say what isn't working.  I'll poke around a little more to see if I can figure something out.

MarcSabatella:
Actually, no mystery at all. It works *if* the "current palette element" is the result of keyboard navigation of the palette.  Not if it was the result of simply clicking the palette.  So I'd need to either *set* the current index when clicking the palette, or else, fall back in applyCurrentElement to whoever the most-recently-click element is being stored.

src/palette/internal/paletteuiactions.cpp
```diff
@@ -44,6 +44,12 @@ const UiActionList PaletteUiActions::m_actions = {
              TranslatableString("action", "Palette search"),
              TranslatableString("action", "Search palettes")
              ),
+    UiAction("apply-current-palette-element",
+             mu::context::UiCtxNotationOpened,
+             mu::context::CTX_ANY,
```
MarcSabatella, line 49: Perhaps this should be CTX_NOTATION_OPENED?  Doesn't make sense to run this command without a score, but I don't understand the difference between the two contexts here

cbjeukendrup:
> So I'd need to either _set_ the current index when clicking the palette,

That sounds sensible to me, isn't it?

src/palette/internal/paletteuiactions.cpp
cbjeukendrup: Indeed, CTX_NOTATION_OPENED makes sense here. The first context determines when an action is enabled (although there may be additional logic for enabled-ness), the second one is the shortcut context, which determines when the keyboard shortcut can be activated. This is useful when the same shortcut is used for different actions in different contexts.

MarcSabatella:
> > So I'd need to either _set_ the current index when clicking the palette,
> 
> That sounds sensible to me, isn't it?

It does, but also a little dangerous.  If clicking sets paletteSelectionModel.currentIndex, then it's going to be set until it's not, and then the next block of code won't have a chance to do what it's supposed to do (substitute the first matching element after a search).  And who knows what else might depend on clicking not setting any more permanent state.  So, I'm trying to key off something else I think might be set to a useful vaue, based on how it is used elsewhere.

Realistically, I'm still pretty clueless about QML and am mostly just pushing buttons and hoping for the best.  But, I'll keep trying a while longer...

src/palette/qml/MuseScore/Palette/internal/PaletteTree.qml
```diff
@@ -114,6 +114,9 @@ StyledListView {
             // in search and no cell selected: apply the first found element
             const parentIndex = paletteTreeDelegateModel.modelIndex(0);
             index = paletteModel.index(0, 0, parentIndex);
+        } else if (!index.valid) {
+            // no cell selected and not in search: use current index from previous click
+            index = currentIndex;
```
cbjeukendrup, line 119: This seems incorrect to me. `paletteController.applyPaletteElement` expects a QModelIndex, not an int, like `currentIndex` is. 

Also, when I run it locally, this code is never called. For me, it works fine without these three added lines, regardless of whether the current palette element was selected using keyboard navigation or by clicking on it.

MarcSabatella: Yes, I figured the type mismatch when I tried running it from the build artifact.  Like I said, I'm really just guessing here!  For me, it definitely doesn't work, though.  The code above *is* called - I got error messages to the console about the mismatch (which is how I discovered it).  Strange that it's otherwise for you, though.  Anyhow, trying something else now...

MarcSabatella:
My latest version feels like it should work, but doesn't.  I'm going to have to put this aside - there's just too much I don't know to have any confidence I'd be able to fix this without inadvertently breaking something else.

I am still wondering why it works for @cbjeukendrup but not for me in the case of clicking the palette.  I wonder if it is connected in any way to the fact the palette search crashes for me (in master) almost all the time.  I take it that's not something others see.  Between that crash and the discrepancy in the behavior here for this PR, it makes me wonder if there isn't some sort of failure to initialize something properly in the in the palette selection model code.

I'll try to find reliable steps to reproduce the crash and submit a separate report.

worldwideweary:
Some observations:

In MS3.6 I liked to have "Enter" be my [apply current element] so I could [type in search bar] a specific item that I'd named, and press [Enter] and be done and back on the score. With these changes so far, even when "Enter" is set in the shortcut, it triggers rather a [nav-trigger-control]

However, when within the score rather than typing in the palette search, "Enter" will work and register properly. 
The index is always the first on the list though. If I click on another palette item besides the first one and then try using the command again later, it's the first item on the list rather than most recent item.... as if to say the currentItem doesn't get updated when pressing the mouse on a cell. 

It *does* update when using keyboard within the palettes to highlight a cell. That is, if the highlighter remains, this command will apply the element once again.  MS3 didn't show a highlight after applying a palette item, but that element would still be applied later by using the command.

worldwideweary:
Went ahead and looked at the code. It *seems* this should be the case:

`internal/Palette.qml`
around line 550:

```cpp
     onClicked: {
         if (paletteView.paletteController.applyPaletteElement(paletteCell.modelIndex, ui.keyboardModifiers())) {
```
That is, the code currently has a *not* operator which  should be removed. By doing so, the highlighter remains on a selected item that was placed from the palette, and in turn, after doing so, the `applyCurrentPaletteElement` will place that item 

Still have the damned problem of not being able to place the first item showing during search though, as again, either you have to press the command twice, or nothing happens at all if [Enter] is assigned as the command. At least that's what I've found so far. Hope this helps.

-526,9 +550,11 @@ StyledGridView {
             }
 
             onClicked: {
-                if (!paletteView.paletteController.applyPaletteElement(paletteCell.modelIndex, ui.keyboardModifiers())) {
+                console.log("Delegate Model onClicked");
+                if (paletteView.paletteController.applyPaletteElement(paletteCell.modelIndex, ui.keyboardModifiers())) {

MarcSabatella:
I still have no actual build environment set up, so I am limited to simple tweaks where I can use the PR artifact for testing.  That is obviously not ideal - not for me or for the build system.  I did try a quick update a few weeks ago to see if I could get this working for click as opposed to keyboard, but still no dice.

Someday I will probably take the time to get a full development workflow again.  But meanwhile, I'd love it if you @worldwideweary or someone else could over the task of straightening this out.  As of 4.1 we have now added back almost all of the other missing commands from MU3.  This remains the glaring exception.  While not a well-known command for the general public, it was actually very significant for accessibility.

worldwideweary:
I have a build, but haven't set up so as to perform a PR. So far though everything seems decent except for the search text not seeming to have any response to commands, which is essential in my opinion. Hope this works out soon. I really want to see MS 4.2/4.3 void of all these little problems!

worldwideweary:
Good news! I've got something working here using [Enter] as the shortcut definition (which was extra work besides other input like an F5 or something). At least I'm on to something. The code approach may be wrong, but I'll try to do a PR this week sometime to see what happens. Also demonstrating that after clicking something else besides the first thing also works. Nevermind the musical notation:

https://github.com/musescore/MuseScore/assets/7139517/b6ccbb86-906d-4be6-a9ff-bc5bf91a7678

MarcSabatella:
Great!  Make sure, though, that it isn't hard-coded to rely on Enter, since that is not available as a shortcut: it's the default for adding system breaks.  While the focus is in the palette, of course, it should apply the selected element, but the apply current palette element command needs to work with focus in the score of course.

worldwideweary:
> Great! Make sure, though, that it isn't hard-coded to rely on Enter,

Oh, of course. It was just extra work to allow for it to *also* use enter.  And as the demonstration shows above by selecting the notes moving forward via keyboard, it also works while score is focused

On the side, one big bummer about this though is I just realized custom palette item names don't show up on the search results. That's integral to my workflow and one of the reasons why I figured this out. So I'm more reluctant to keep going (like getting the palette to show when search is triggered without its already showing).Maybe I can fix that too, but it can be time-consuming sometimes figuring out what to do and where with this stuff coming in from ground zero with MS4's different codebase

worldwideweary:
#13281 It would be cool if, so long as artifacts are built successfully, you could give this a test sometime! And anyone else for that matter.  I left some comments in the pull-request also if you're interested

MarcSabatella:
I'll leave this open for now just in case we want separate PR's, but from my perspective this is now superceded by #18652

Issue #13799, opened 2022-10-17
[MU4 Issue] List of recent files grows infinitely


cbjeukendrup:
**Describe the bug**
The list of recent files grows infinitely; it can become unlimitedly long.

**Expected behavior**
@Tantacrul / @bkunda, what shall we do?

Possible solutions:
- limit by number, for example max 100 recent files
- limit by modified date, for example max 1 year since modified. But actually, this is not an option, because it would imply that a file that is opened without modifying and last modified more than 1 year ago will not appear in the list.
- limit by opened date, but that would require storing the opening dates in the list of recent files, because not all operating systems support "opened" date.

Of course, we don't want to set the limit too low, because a filled home page looks 100x nicer than with just a few scores.

Tantacrul:
100 makes sense. I know we spoke about this before and either it got undone or it fell through the cracks.

ecstrema:
I mean, in any real-world case, can it cause a performance bottleneck?

I mean, if 5 *different* files are opened **every** day for 10 years, and the average path length is 40 chars, then it'll take 0.73 MB of memory, which is less than 0.2% of the memory used by MuseScore with a small score opened and no VST loaded.

I'd rather have a convenient way of finding all my files, not just the last 100. In VSCode, I regularly look for "recent" projects that I haven't opened in a year, so definitely further away than 100.

Solutions: 
 - Maybe use 1 000? Unless you're a bot, few people will ever open a thousand scores on the same computer, but lots will open 100.
 - Don't put a limit

What do you think? I know it's a very small detail.

Tantacrul:
I think it becomes unwieldy to think of it as something that should contain that many projects. Practically speaking, it'd be better if one just saved their work in a way that made it easier to find.

One point: we do want to eventually tie 'your scores' to a registered account if you are working online. This isn't the same thing exactly but we may very well add a filter that lets you see all your online scores represented on the desktop app.

cbjeukendrup:
Lowering the priority, because since https://github.com/musescore/MuseScore/issues/16918 it does not cause big performance problems anymore. I still don't really like the fact that there is an infinitely growing list somewhere (especially since recent files are absolutely unimportant data and most people would only ever care about the 100 most recent files at most). OTOH, as @ecstrema says, the list won't grow _that_ quickly, and of course each entry is only a small amount of data.

Tantacrul:
[...]

PR #13804, opened 2022-10-17
Limit the number of recent files to 100

cbjeukendrup:
Resolves: #13799 

After this PR, starting at the first moment you open another file, the list of recent files will be trimmed to 100 entries.

### DIFF
--- a/src/project/internal/recentprojectsprovider.cpp
+++ b/src/project/internal/recentprojectsprovider.cpp
@@ -40,8 +40,17 @@ void RecentProjectsProvider::init()
 ProjectMetaList RecentProjectsProvider::recentProjectList() const
 {
     if (m_dirty) {
+        TRACEFUNC;
+
         io::paths_t paths = configuration()->recentProjectPaths();
         m_recentList.clear();
+
+        static constexpr size_t MAX_RECENT_FILES = 100;
+
+        if (paths.size() > MAX_RECENT_FILES) {
+            paths.resize(MAX_RECENT_FILES);
+        }
+
         for (const io::path_t& path : paths) {
             ProjectMeta meta;
             if (engraving::isMuseScoreFile(io::suffix(path))) {


src/project/internal/projectactionscontroller.cpp
```diff
@@ -972,7 +972,15 @@ void ProjectActionsController::prependToRecentScoreList(const io::path_t& filePa
...
 
     recentScorePaths.insert(recentScorePaths.begin(), filePath);
+
+    static constexpr size_t MAX_RECENT_FILES = 100;
+
+    if (recentScorePaths.size() > MAX_RECENT_FILES) {
+        recentScorePaths.resize(MAX_RECENT_FILES);
+    }
+
     configuration()->setRecentProjectPaths(recentScorePaths);
```
vpereverzev, line 982: that's OK for configuration to know about all the possible recent files. The real problem is in model/view. Move this limitation into "RecentProjectsModel"

cbjeukendrup: > that's OK for configuration to know about all the possible recent files

Really? Every time that we want to add (or "bump") a recent file, we need to search through the whole list, to find if the file is already in the list, and if yes, erase it from there. If the list grows longer and longer, this will take more and more time. Of course I understand that this amount of time will likely still be negligible for most users, but it just seems wasteful. Especially if we never actually show all recent files to the user, so it has completely zero benefit to store them.

If you don't like to have this logic here, we can use an alternative solution; in `ProjectConfiguration::setRecentProjectPaths`, where the list is actually saved, we iterate over the list of paths to convert them to JSON. Instead of iterating over all of them, we could iterate over the first 100 only. That seems not much better to me, but it's just an alternative.

vpereverzev: Again, that's a responsibility of model/view. The original data should remain consistent.

Usually, you'd expect to have "technical" limitation which user would never reach, but you still want to have "some" upper bound (let's say 1000) and "visual" limitation of model/view which saves some performance for you.

The reason why I'm asking you to differentiate this is pretty simple - let's say, tomorrow we'll decide to show 150 items instead of 100, since the "original" data would remain consistent you'll not miss anything

> Every time that we want to add (or "bump") a recent file, we need to search through the whole list, to find if the file is already in the list

BTW If that's a problem you were going to resolve by that limit, I'd suggest to change a container on unordered_set/map and do the necessary sorting on the model's side

cbjeukendrup:
I found that the long list of recent files causes a huge performance bottleneck when launching MuseScore and when opening a score; this happens in `RecentProjectsProvider::recentProjectList()`. So let's do the limitation in that method, rather than in any higher level like `RecentProjectsModel`. That doesn't solve the performance problem completely, but might help a little bit.

(If you open the frontmost file, so that the list of recent files stays the same, the score opens almost instantly; if you open the exact same file but it's somewhere else in the list, so that the whole list of recent files is reloaded and for every recent file the metadata needs to be re-read, it takes like 15 seconds.)

RomanPudashkin:
@cbjeukendrup I think, we can close this (since now we have your async-loading optimization)

cbjeukendrup:
Yes, let's close it. I'm still not a fan of having an infinitely growing list somewhere, especially if that list contains such unimportant information as recent files while nobody is interested in more than, say, 100 recent files. But I guess we can live with it.

Issue #13806, opened 2022-10-17
[MU4 Issue] Zoom to page width command missing


MarcSabatella:
**Describe the bug**
The "zoom to page width" comamdn from MU3 is missing

**To Reproduce**
1. Go to Edit / Preferences / Shortcuts
2. Search for "Zoom"
3. See error: no zoom to page width command

**Expected behavior**
Zoom to page width should be an available command

**Platform information**
 - OS: Linux

**Additional context**
The underlying code is present, but the command it not hooked up.

In MU3, this actually worked as a toggle, switching between page width and whatever the previous setting was.  That's somewhat useful but the current code does not appear to support it.  Not a big deal to me; the command is worth having anyhow.   I will try a PR for this...

DmitryArefiev:
Implemented in #13807

Before:

![Capture](https://user-images.githubusercontent.com/90187801/235750855-30ee4129-9d6d-4ed0-8c86-7c47e85b263f.PNG)

After:

![Capture2](https://user-images.githubusercontent.com/90187801/235750889-aeeaf046-01ab-4ba2-91c6-df8ba70bc8ed.png)

PR #13807, opened 2022-10-17
Reimplement "zoom to page width" command

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/13806

The needed code was present; this simply adds the UiAction to enable it.

(no test applicable)

### DIFF
--- a/src/notation/internal/notationuiactions.cpp
+++ b/src/notation/internal/notationuiactions.cpp
@@ -1451,6 +1451,18 @@ const UiActionList NotationUiActions::m_actions = {
              TranslatableString("action", "Zoom to page width"),
              TranslatableString("action", "Zoom to page width")
              ),
+    UiAction("zoom-whole-page",
+             mu::context::UiCtxNotationOpened,
+             mu::context::CTX_ANY,
+             TranslatableString("action", "Zoom to whole page"),
+             TranslatableString("action", "Zoom to whole page")
+             ),
+    UiAction("zoom-two-pages",
+             mu::context::UiCtxNotationOpened,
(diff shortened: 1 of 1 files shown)


MarcSabatella:
This works, but note, it does introduce a new string.

Also, the page width command (executed from the status bar) is a little awkward even without this PR - it doesn't also re-center the score.  That's probably best looked at in conjunction with other zoom behaviors.  I'll let others worry about that.

cbjeukendrup:
@MarcSabatella I found that the there is a similar problem for the actions `zoom-whole-page` and `zoom-whole-page`. Would you mind fixing those too?

MarcSabatella:
Yes, I can do that, assuming it's as simple as the page width (the code is still there and just needed to be connected).  I assume you mean zoom-whole-page and zoom-two-pages?

MarcSabatella:
As far as I can tell, these never existed before as separate commands (not in MU3 anyhow), but there does seem to be code to handle them, we'll see if it actually works.  I just went with a generic text description for now ("Zoom to whole page" and "Zoom to two pages") which is lame but maybe someone wants to suggest some a little better like "Zoom to fit whole page"?  For now, I want to be sure it actually works :-)

MarcSabatella:
For the record, yes, the commands work, as far as I'm concerned the PR is ready for review.

cbjeukendrup (approved):

DmitryArefiev:
Tested #13806 on Win10, Mac13, LinuxUbuntu 22.04 - **PASS**

Thanks!

Issue #13783, opened 2022-10-16
[MU4 Issue] Drum notation: MIDI keyboard can only enter the selected drum hit


ShibuyaCyana:
To reproduce:
1. add a drumset
2. press N
3. for now MIDI keyboard can enter drum notes depending on which key I press on it, eg. pressing C results in a kick drum and D results in a snare
4. select a drum hit from the drum hit menu below, say hi-hat
5. now whichever key I press on my MIDI keyboard, it results in a hi-hat.
6. now I can't quite this mode and revert to the status in step 3, without restarting MuseScore.

This problem also occurs in MS3.

PR #13821, opened 2022-10-18
Drum midi entry fix

### LINKED ISSUE #12936: [MU4 Issue] Percussion beams on wrong side
**Describe the bug**
Percussion staff stems and beams can be on the wrong side when entered from the on-screen Piano. I haven't tested a MIDI keyboard, but it is possible that the problem exists there as well.
 
**To Reproduce**

1. Create a 5-line percussion staff score.
2. In note-input mode, ensure voice 1 is selected and enter some eighth-note closed hihats (F#2) in measure 1  using the on-screen piano.
3. Select voice 2 and enter some 1/4 note bass drum beats (C2) in the same measure using the on-screen piano.

RESULT: Voice 1 beams are on the wrong side of the notehead.

percussion_beams

 Curiously, if you break all the beams using the "no beam" icon then UNDO, the beams appear on the correct side.

**Expected behavior**

Voice 1 Stems should point upwards in this case.

**Platform information**
 OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2887048614, revision: github-musescore-musescore-61484bc

mike-spa:
Resolves #12936 and #13783 

I'm not an expert here, but I think when entering from midi it isn't necessary to reassign `nval.pitch`, as it is already defined by the midi event (this causes #13783). Also, I think when no note type is selected from the drum input menu, `InputState::drumNote` defaults to -1, which causes to skip the following if body and therefore we don't set the correct track and stem direction for the input note (which causes #12936).


MarcSabatella:
I'm not an expert on this code either, but this *does* look like exactly the right fix based on the reported symptoms.

oktophonie (approved):

RomanPudashkin (approved):

Issue #13831, opened 2022-10-19
[MU4 Issue] Crash when doing ctrl+A (select all)


mike-spa:
Crash when doing crtl+A (select all) on this score.

[MAHLER Symphony No 9 - I (4).zip](https://github.com/musescore/MuseScore/files/9820603/MAHLER.Symphony.No.9.-.I.4.zip)

PR #13832, opened 2022-10-19
Fix crash on select all

mike-spa:
Resolves: #13831 

Was caused by calling `toChord(elem)` on an element that isn't a Chord.


src/inspector/internal/services/elementrepositoryservice.cpp
```diff
@@ -207,6 +207,9 @@ QList<mu::engraving::EngravingItem*> ElementRepositoryService::findNotes() const
             const mu::engraving::Beam* beam = mu::engraving::toBeam(elementBase);
 
             for (mu::engraving::ChordRest* chordRest : beam->elements()) {
+                if (!chordRest->isChord()) {
+                    continue;
+                }
                 for (mu::engraving::Note* note : engraving::toChord(chordRest)->notes()) {
```
Eism, line 213: Please add an empty line before this

Eism (approved):

oktophonie (approved):

Issue #11392, opened 2022-04-26
[MU4 Issue] Large MuseScore3 score opens 4x longer comparing with MU3


DmitryArefiev:
**Describe the bug**

**To Reproduce**
Compare [Test..._Beethoven__Symphony_No._9__Op._125.mscz](https://musescore.com/openscore/scores/5733014) loading in MU4 and MU3 - **now MU4 opens that score 3-4x time longer**

**Expected behavior**

**Platform information**
macOS, Windows, Linux

**Additional context**

DmitryArefiev:
Tested in **latest master** on Win10, Mac12.5 - **STILL OCCURS**

Tantacrul:
Moving to beta 2, since we have made a decent amount of performance gains. With luck @mike-spa's PR about part generation doesn't break anything and we get even more gains.

However, for Beta 1, we're satisfied with current performance and want to focus on broken audio / crash issues

mike-spa:
Another thing I noticed with one of my files: it takes much longer to open the MU3 file than it takes to open the same identical file once it's been saves in MU4 for the first time. This means that something very inefficient is likely going on during the migration process. I'm going to take a look at that as well and see if I can improve it.

DmitryArefiev:
Results on Windows10 after #13835 optimization 

MuseScore**3**:  14sec+6sec=**20sec**
MuseScore**4 after #13835** optimization:  35sec+1m15sec= **110sec** 
MuseScore**4 master**:  53sec+1m38sec=**151sec**

DmitryArefiev:
> Results on **Windows10** after #13835 optimization
> 
> MuseScore**3**: 14sec+6sec=**20sec** MuseScore**4 after #13835** optimization: 35sec+1m15sec= **110sec** MuseScore**4 master**: 53sec+1m38sec=**151sec**

Results on **Windows10** after #14240 

MuseScore**4 after** #14240 optimization: 33+59=**92sec**
MuseScore**4 master**: 34+1m35sec=**129sec**

DmitryArefiev:
Results on **Windows10** after #14271
 
MuseScore**4 after #14271** optimization 15+30=**45sec**

DmitryArefiev:
Comparing On **Windows10**:

MuseScore**4 OLD build**: 59sec+1m34sec=**153sec**
MuseScore**4 after all optimizations**: 15sec+30sec=**45sec**
MuseScore**3.6.2**: 15sec+6sec=**21sec**

[...]

PR #13835, opened 2022-10-19
Score opening optimization

mike-spa:
Does not fully resolve, but definitely helps with: #11392 

The modification leaves the method functionally identical (becase the part following the first return does nothing if diff == 0). Tested on the Beethoven 9 score, the loading time is reduced by nearly 30%.


RomanPudashkin (approved):

DmitryArefiev:
Tested #11392 on Win10, Mac12, LinuxMint 20.2 - **yes, #11392 better now**

Issue #13840, opened 2022-10-20
[MU4 Issue] options in migration dialog are locked (regression)


sammik:
**Describe the bug**
It is not able to change options in migration dialog

**To Reproduce**
1. Open old score
2. Untick "New font Leland"
3. It is not possible

**Expected behavior**
Possible to tick / untick options

**Screenshots**

https://user-images.githubusercontent.com/1646034/196875061-a3b4a0b8-b68d-4f56-887c-0ad729944754.mp4

**Platform information**
 - OS: [reported on Linux]

**Additional context**
[Discussed](https://github.com/musescore/MuseScore/issues/13824#issuecomment-1284381412) in #13824

Issue #13824, opened 2022-10-18
[MU4 Issue] unable to tick "Auto spacing" in migration dialog for 2.0.2 file. Option should be removed.


sammik:
**Describe the bug**
Opening 2.0.2 file, migration dialog appears, but it is not possible to tick "Auto spacing"

**To Reproduce**
Steps to reproduce the behaviour:
1. Open 2.0.2 file
2. Try to tick "Auto spacing"

**Expected behaviour**
The auto spacing option should no longer be available. Let's just remove it (it defaults to being on, like all other scores)

**Screenshots**

https://user-images.githubusercontent.com/1646034/196492935-d9f042e0-eb60-47a8-bc16-fab46e0bd7c0.mp4

**Platform information**
 - OS: [reported on Linux and MacOS]

MarcSabatella:
I am not sure what that box is even supposed to mean.  Does it refer to automatic placement (aka autoplace)?

I'll put it out there that there are almost no cases where disabling autoplace would produce reasonable results.  There's just no way MuseScore 4 (or 3) can completely reproduce all the quirks of the MuseScore 2 layout algorithm, so you will likely end up with a worse mess than leaving it on.

MuseScore 3 offers the option of not checking that box, but as far as I can tell, it has no effect whatsoever.  Automatic placement is always enabled.  I don't see any point in having MuseScore 4 provide a way to disable it on import.

Tantacrul:
Agreed (with consultation from @oktophonie).

Can we just remove the option? I'll update the issue to reflect this.

MarcSabatella:
Unless it does something useful that I’m not aware of I’dd be very happy to see it go away.  I suppose maybe it might relate to vertical justification rather than autoplace?  Even if so, I see no reason to expose that in the dialog.

sammik:
"Auto spacing" (aka autoplace) should be on by default, but it is not always. Have no idea,what it depends on.

But I found, **other options are also blocked from ticking / unticking**.

Also 3.2.3 files.

https://user-images.githubusercontent.com/1646034/196693818-881f0a80-bb57-48ed-9f55-e11a9156a1d5.mp4

It is regression. In Alpha it worked.

So I would recommend to update bug description to cover **actual problem - non accesible tickers**.

MarcSabatella:
[...]

PR #13850, opened 2022-10-20
Fix checkboxes in MigrationDialog

cbjeukendrup:
Resolves: #13824 ~(Note: does not remove the checkbox; simply fixes it. That seems currently the lowest-risk solution.)~
Resolves: #13840

### DIFF
--- a/src/project/internal/projectmigrator.cpp
+++ b/src/project/internal/projectmigrator.cpp
@@ -107,7 +107,7 @@ Ret ProjectMigrator::askAboutMigration(MigrationOptions& out, const QString& app
     query.addParam("migrationType", Val(migrationType));
     query.addParam("isApplyLeland", Val(out.isApplyLeland));
     query.addParam("isApplyEdwin", Val(out.isApplyEdwin));
-    query.addParam("isApplyAutoSpacing", Val(out.isApplyAutoSpacing));
+    query.addParam("isApplyAutomaticPlacement", Val(out.isApplyAutomaticPlacement));
     RetVal<Val> rv = interactive()->open(query);
     if (!rv.ret) {
         return rv.ret;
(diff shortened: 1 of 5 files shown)


cbjeukendrup:
Removed the "Autoplace" checkbox now; automatic placement will be applied always. Will need some testing with MS2 scores.

src/project/internal/projectmigrator.cpp
```diff
@@ -167,7 +164,7 @@ Ret ProjectMigrator::migrateProject(engraving::EngravingProjectPtr project, cons
         m_resetStyleSettings = false;
     }
 
-    if (ok && opt.isApplyAutomaticPlacement) {
+    if (ok && score->mscVersion() < 300) {
```
MarcSabatella, line 163: FWIW, I'm not so sure this is really what I'd expect either.  But, if that's actually all the checkbox was doing before - controlling whether manual adjustments were reset or not - it's harmless to keep the box.  It's just terribly named, then - if the point of the box is the reset positions to the (new/better) defaults, that's what it should say.

Still, since it's trivial to do Ctrl+A / Ctrl+R after loading the score, I'd just as soon see the checkbox go away, *and* see this block of code away go away with it, so older scores simply come through as is, and the user can then decide if they want to reset positions.  After all, if we do reset them here, it will be impossible to get them back (unless "undo" immediately on import accomplishes that, but still, that's not obvious).

oktophonie: Is the point not that this resets manual adjustments made in files from pre-MU3, i.e. before the autoplace era, whicih are now likely to be meaningless? Seems better to me to present the user with something (hopefully) sensible rather than something messed-up which they need to take an explicit step (even if "trivial") to fix. In any case the checkbox should go.

MarcSabatella: I assume that is the point, yes.  But, we have several years of experience here already to tell us, it's really a toss-up as to whether resetting all positions when importing MU2 scores into MU3 produces better results than not.  Some of those manual adjustments might be counterproductive, others might be unnecessary but harmless, others still actually important for the layout effect desired.

But, it's a bigger jump from MU2 to MU4, obviously; so that much more likely those old manual adjustments are not good.  Therefore, maybe not a *terrible* idea to just toss them.  I was just surprised to realize that this is what this checkbox does, and suddenly not so sure getting rid of it is warranted.

Eism (approved):

oktophonie (approved):

PR #13851, opened 2022-10-20
Fix #12544: Fill empty part with masterscore's information

### LINKED ISSUE #12544: [MU4 Issue] User-created parts show generic title/subtitle/composer etc.
**Describe the bug**
If the user creates a custom part in the Parts dialog, this part contains text fields that only show generic title/subtitle/composer info, rather than the same info as the full score. 

**To Reproduce**
1. Create a score, and on the "additional info" page, enter a custom title, subtitle, and/or composer. 
1. Open the Parts dialog
2. Click "Create new part"
3. Open the newly created part

**Expected behavior**
Newly created parts should show the same info as the full score and the "default" parts.

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12544

New empty parts were being created with hardcoded default values. Now those values match the master score's.

### DIFF
--- a/src/notation/internal/excerptnotation.cpp
+++ b/src/notation/internal/excerptnotation.cpp
@@ -92,10 +92,18 @@ void ExcerptNotation::fillWithDefaultInfo()
         }
     };
 
-    setText(TextStyleType::TITLE, qtrc("notation", "Title"));
-    setText(TextStyleType::COMPOSER, qtrc("notation", "Composer / arranger"));
-    setText(TextStyleType::SUBTITLE, "");
-    setText(TextStyleType::POET, "");
+    auto getText = [&score](TextStyleType textType, const QString& defaultText) {
+        if (Text* t = score->getText(textType)) {
+            return t->plainText().toQString();
+        } else {
+            return defaultText;
+        }
+    };
+
+    setText(TextStyleType::TITLE, getText(TextStyleType::TITLE, ""));
+    setText(TextStyleType::COMPOSER, getText(TextStyleType::COMPOSER, ""));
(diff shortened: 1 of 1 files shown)


src/notation/internal/excerptnotation.cpp
```diff
@@ -94,10 +94,18 @@ void ExcerptNotation::fillWithDefaultInfo()
...
+            return t->plainText().toQString();
+        } else {
+            return defaultText;
+        }
+    };
+
+    setText(TextStyleType::TITLE, qtrc("notation", getText(TextStyleType::TITLE, "")));
+    setText(TextStyleType::COMPOSER, qtrc("notation", getText(TextStyleType::COMPOSER, "")));
+    setText(TextStyleType::SUBTITLE, qtrc("notation", getText(TextStyleType::SUBTITLE, "")));
+    setText(TextStyleType::POET, qtrc("notation", getText(TextStyleType::POET, "")));
```
cbjeukendrup, line 108: I don't think `qtrc` should be here. The values of `getText(…)` are potentially not translatable strings; some of them might be even entered by the user. So we should not try to translate them, otherwise we loose performance and might get strange results (for example, if the user enters something that _does_ happen to be recognised by the translation system).

asattely: Ah, good point, will fix that.

oktophonie (approved):

RomanPudashkin:
@asattely please rebase it

RomanPudashkin (approved):

asattely:
> @asattely please rebase it
Rebased!

Issue #13649, opened 2022-10-06
[MU4 Issue] Crash when changing time signature in score with parts


dragonwithafez:
**Describe the bug**
When trying to add a time signature change in a score where part scores exist, the software crashes. Occurs whether trying to add the time signature to a part score or the main score.

**To Reproduce**
1. Create a new score
2. Create a part score for one of the instruments
3. Try to add a time signature change anywhere in the score, i.e. from 4/4 to 2/4
4. Observe crash

**Expected behavior**
Software should not crash.

**Platform information**
 - OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-, revision: github-musescore-musescore-

**Additional context**
Application output from before crash:

> 15:20:53.470 | DEBUG | main_thread | UiActionsRegister | updateEnabledAll: currentCtx: UiCtxNotationOpened
> 15:20:53.736 | DEBUG | main_thread | UiActionsRegister | updateEnabledAll: currentCtx: UiCtxNotationFocused
> 15:20:53.803 | DEBUG | main_thread | Score      | tick2measure: tick2measure 57600 (max 54720) not found

lyrra:
confirmed on linux.
<details><summary> stacktrace:</summary>

```
(gdb) ba
#0  0x0000555555dc0650 in std::vector<mu::engraving::MStaff*, std::allocator<mu::engraving::MStaff*> >::size (this=0x160)
    at /usr/include/c++/10/bits/stl_vector.h:919
#1  0x0000555555db82e1 in mu::engraving::Measure::setMeasureRepeatCount (this=0x0, n=0, staffIdx=0)
    at src/engraving/libmscore/measure.cpp:2986
#2  0x0000555555ce43b4 in mu::engraving::Excerpt::cloneStaff2 (srcStaff=0x55555aee4d90, dstStaff=0x55555c0740e0, startTick=..., endTick=...)
    at src/engraving/libmscore/excerpt.cpp:1395
#3  0x0000555555e1c749 in mu::engraving::ScoreRange::write (this=0x7fffffff8180, score=0x55555c1914a0, tick=...)
    at src/engraving/libmscore/range.cpp:727
#4  0x0000555555c94a25 in mu::engraving::Score::rewriteMeasures (this=0x55555c1914a0, fm=0x55555b1eb470, lm=0x55555adfe660, ns=..., 
    staffIdx=18446744073709551615) at src/engraving/libmscore/edit.cpp:1011
[...]

PR #13856, opened 2022-10-21
Fix crash when changing TimeSig with parts

mike-spa:
Resolves: #13649


oktophonie (approved):

src/engraving/libmscore/excerpt.cpp
```diff
@@ -1391,6 +1391,9 @@ void Excerpt::cloneStaff2(Staff* srcStaff, Staff* dstStaff, const Fraction& star
 
     for (Measure* m = m1; m && (m != m2); m = m->nextMeasure()) {
         Measure* nm = score->tick2measure(m->tick());
+        if (!nm) {
```
RomanPudashkin, line 1394: Could you please explain why it is null? I don't really get it from the commit description

RomanPudashkin (approved):

PR #13859, opened 2022-10-21
Trying to fix authorization problems on Windows

cbjeukendrup:
Added some logging, hoping it uncovers problems

### DIFF
--- a/src/cloud/internal/cloudservice.cpp
+++ b/src/cloud/internal/cloudservice.cpp
@@ -122,6 +122,30 @@ void CloudService::init()
     connect(m_oauth2, &QOAuth2AuthorizationCodeFlow::authorizeWithBrowser, this, &CloudService::openUrl);
     connect(m_oauth2, &QOAuth2AuthorizationCodeFlow::granted, this, &CloudService::onUserAuthorized);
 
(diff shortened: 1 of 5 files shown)


cbjeukendrup:
I'm going to close this now, since Bradley and I _finally_ found what was wrong. 

TL;DR: `C:\Windows\System32\msvcr100.dll` was missing on Bradleys (virtual) Windows machine, which caused the libssl and libcrypto libraries to fail loading, which caused any https connection in MuseScore to fail, which caused downloading the access token to fail.

@DmitryArefiev So if we ever get a report of a user who is unable to log in in the app, or if the videos on the Learn page will never load, or if they always get a "unable to connect to MuseScore.com" popup when trying to publish, then their `C:\Windows\System32\msvcr100.dll` might be missing. Do you have by any chance a place where you document such "possible solutions for problems that are out of our control"?

DmitryArefiev:
> @DmitryArefiev So if we ever get a report of a user who is unable to log in in the app, or if the videos on the Learn page will never load, or if they always get a "unable to connect to MuseScore.com" popup when trying to publish, then their `C:\Windows\System32\msvcr100.dll` might be missing. Do you have by any chance a place where you document such "possible solutions for problems that are out of our control"?

I will put this info into my internal document "User cases". Thanks!

But the best place for potential cases like that (+solution) is https://musescore.org/en/project/issues/musescore

Issue #13569, opened 2022-10-01
[MU4 Issue] Tuplets in a staff negate tuplet signs in lower staves


ModernMozart:
**Describe the bug**

A tuplet in one staff cancels out tuplet markings in lower staves that occur at the same time, but only when one note fills the entire tuplet in the lower staff.

**To Reproduce**

1. create a tuplet in one staff 
2. In a lower staff, create a tuplet in the same beat where one note fills the entire tuplet, e.g. a dotted quarter in a triplet of 3 eighths.
3. you now get a bugged tuplet that doesn't respond to options in the inspector that disappears upon reloading the score.

**Expected behavior**
A tuplet marking in both staves with tuplets, that both respond independently to options in the inspector
**Screenshots**

<img width="698" alt="image" src="https://user-images.githubusercontent.com/73030322/193421902-35f17c43-094c-4912-b5f1-79001357fa0c.png">

**Platform information**
macOS 11.6, M1

[Tuplet bug.mscz.zip](https://github.com/musescore/MuseScore/files/9690908/Tuplet.bug.mscz.zip)

**Additional context**

MarcSabatella:
I feel this could be related to https://musescore.org/en/node/334771, which was missing steps to reproduce, but I can confirm I've seen similar.

erinic04:
![image](https://user-images.githubusercontent.com/21023353/195947504-61e36638-9ba1-4f20-8543-bbf72e612091.png)
Could be related

erinic04:
Here's more about the issue https://discord.com/channels/818804595450445834/818804809784229898/1030589827999072286

PR #13860, opened 2022-10-21
Fix tuplet layout bug

mike-spa:
Resolves: #13569 

I don't think we should break out of that loop, given that all tracks may contain (possibly multiple) tuplets. Added a vtest to show the issue and relative solution.


oktophonie (approved):

cbjeukendrup (approved):
Looks good to me! @RomanPudashkin over to you!

PR #13972, opened 2022-10-24
Fix #13886: Improved cross-staff validity checks

### LINKED ISSUE #13886: [MU4 Issue] Crash when moving staves in case of cross staff notation
**Describe the bug**
In scores with cross-staff notation, moving the staves of an instrument can cause a crash.

**To Reproduce**
1. Create score with grand staff
2. add note on top staff
3. Cmd+Shift+Down to move the note to the bottom staff
4. using the instruments panel, swap the staves OR delete the bottom staff
-> crash 

Also interesting: if you _hide_ the bottom staff instead of deleting it, the note moves back to the top staff but is still on the line as if it is on the bottom staff.

https://user-images.githubusercontent.com/48658420/197366860-1867fb36-48e6-4098-97bf-725dd15c2142.mov

**Expected behavior**
No crash, but... maybe simply move the note back to the top staff when the top staff becomes the bottom staff?

**Additional context**
Also occurs in MS 3.6.2

mike-spa:
Resolves: #13886

The main problem is that `Note::updateRelLine()` is the wrong place to be checking the validy of cross-staff moves (rests can also be cross-staff). The validity of `_staffMove` should be checked by ChordRest because it is a ChordRest property, and it should be checked much earlier. 

The first commit fixes that, and does the only correct thing to do: if the destination staff is invalid, reset the _staffMove to zero (i.e., bring it back to the origin staff).

The second commit is more of a "nice to have": if the destination staff is invalid, we store the _staffMove before resetting it, and re-apply it in case the destination staff become valid again (for example, it is unhidden). @oktophonie @cbjeukendrup let me know what you think. The PR can be also merged without the second commit, in which case when a _staffMove becomes invalid it will simply be reset and forgotten.

https://user-images.githubusercontent.com/93707756/197523266-d37a0c83-a404-4061-9a7f-96164315d871.mp4

### DIFF
--- a/src/engraving/layout/layoutmeasure.cpp
+++ b/src/engraving/layout/layoutmeasure.cpp
@@ -690,6 +690,11 @@ void LayoutMeasure::getNextMeasure(const LayoutOptions& options, LayoutContext& 
                     if (!cr) {
                         continue;
                     }
+                    // Check if requested cross-staff is possible
+                    if (cr->staffMove() || cr->storedStaffMove()) {
+                        cr->checkStaffMoveValidity();
+                    }
+
                     double m = staff->staffMag(&segment);
                     if (cr->isSmall()) {
                         m *= score->styleD(Sid::smallNoteMag);
--- a/src/engraving/layout/layoutsystem.cpp
+++ b/src/engraving/layout/layoutsystem.cpp
@@ -1479,11 +1479,6 @@ void LayoutSystem::updateCrossBeams(System* system, const LayoutContext& ctx)
                 }
             }
         }
-        for (Segment& seg : toMeasure(mb)->segments()) {
-            if (seg.next()) {
-                seg.computeCrossBeamType(seg.next());
-            }
-        }
     }
 }
 
--- a/src/engraving/libmscore/chordrest.cpp
+++ b/src/engraving/libmscore/chordrest.cpp
@@ -1426,4 +1426,34 @@ void ChordRest::undoAddAnnotation(EngravingItem* a)
     a->setParent(seg);
     score()->undoAddElement(a);
 }
+
+void ChordRest::checkStaffMoveValidity()
+{
+    if (!staff()) {
+        return;
+    }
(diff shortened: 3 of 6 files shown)


cbjeukendrup (commented):
This generally looks good to me, especially the first commit. 

Because `_storedStaffMove` is just a runtime member field, it will obviously be forgotten when the file is saved and closed, but that's probably fine.

Also, I'm not sure how `_storedStaffMove` will interact with Undo/Redo. But since it is read and written only during the layout process, and the logic in this method looks fine, I don't think it will cause issues. (And if it does, that's really a tiny problem, since this is very niche anyway.)

mike-spa:
> it will obviously be forgotten when the file is saved and closed

Yes, that is intentional, I don't think it would be worth writing to file something like this. Besides, if you open a file you closed a while ago, you've likely forgotten that you used to have cross-staff notes, so you'd be probably surprised by seeing them jump when you (for instance) add a staff. I think it's better this way.

> Also, I'm not sure how `_storedStaffMove` will interact with Undo/Redo.

As far as I could test, it still works as expected :)

oktophonie (approved):

src/engraving/layout/layoutmeasure.cpp
```diff
@@ -690,6 +690,9 @@ void LayoutMeasure::getNextMeasure(const LayoutOptions& options, LayoutContext&
                     if (!cr) {
                         continue;
                     }
+                    // Check if requested cross-staff is possible
+                    cr->checkStaffMoveValidity();
```
RomanPudashkin, line 694: It seems that from now on it will be called much more often than before. Won't it hurt performance?

mike-spa: That can change from score to score, but in general we are actually calling this logic _less_ often, because now we call it once per chord:
image
but before we used to call it once per note:
image
Now I've also moved that if condition on the caller side, so we avoid entering the funciton at all if we don't need it. So performance-wise we should be equal or better than before

RomanPudashkin (approved):

PR #13990, opened 2022-10-24
Fix translations for local debug builds

jeetee:
Partial revert of ca8b96f23c892757d87061af37a22d8c04a3e39b to allow local debug builds to generate translation files. If not, those builds will assert out upon launch due to missing translations..

Resolves: Local debug builds crash on [this assert](https://github.com/musescore/MuseScore/blob/master/src/languages/internal/languagesservice.cpp#L276) during launch due to the missing qm-files.

Note: This is a stop-gap to allow contributors to build, run and test. Likely a better (configurable) option is required to allow CI to have the up to date languages while still allowing local builds as well.


cbjeukendrup:
Igor fixed it just today... see #13964

jeetee:
Ah.. missed that; closing this duct tape in favor of his real solution.

PR #12960, opened 2022-08-22
Fix #12788: Slur with lyrics

mike-spa:
Resolves #12788 

The origin of the problem is in the way that lyrics shapes are handled, but that is something that will be addressed later on, as we're planning to overhaul lyrics at some point in 4.x versions. For now, slur should simply ignore lyrics when looking for collisions.

Issue #12690, opened 2022-08-05
[MU4 Issue] Key signatures from Instrument change from top staff are duplicated to new staves 


DmitryArefiev:
**Describe the bug**
Key signatures from Instrument change (transposing instrument) from top staff are duplicated to new staves 

**To Reproduce**
1. Open [KS.zip](https://github.com/musescore/MuseScore/files/9268239/KS.zip)
2. Add Flute - **Key signatures should not be copied from Oboe 1 to Flute**

https://user-images.githubusercontent.com/90187801/183071253-ccc3c59d-a5ba-4a48-92b8-c7dbdef8ab13.mp4

And also unable to delete key signatures independently (you need to create Instrument change again to have proper KS)

https://user-images.githubusercontent.com/90187801/183073472-d3afc306-4c07-4959-bcdd-8ba52f1088e2.mp4

**Platform information**
 macOS, Windows, Linux

**Additional context**
* This bug also happens with user-created local key signatures; not only with local key signatures that are created as a result of adding an instrument change.
 (To create a local key signature, hold Ctrl/Cmd while adding it from the Palette; to select a key signature on one staff only, hold Ctrl/Cmd while clicking it, which allows you to delete it for that staff only.)

* Also occurs in **MU3**

DmitryArefiev:
Could be removed to Beta2 since the bug occurs in MU3

cbjeukendrup:
This bug also happens with user-created local key signatures; not only with local key signatures that are created as a result of adding an instrument change.

(To create a local key signature, hold Ctrl/Cmd while adding it from the Palette; to select a key signature on one staff only, hold Ctrl/Cmd while clicking it, which allows you to delete it for that staff only.)

DmitryArefiev:
Thanks! Added to the description

PR #13999, opened 2022-10-24
Fix #12960: filter key signature changes to only include scorewide ones

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12690

Previously, all of the time signatures were copied from the top staff when creating a new staff. Now, they are only added if there is a key signature on every staff on that tick.

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -2793,12 +2793,43 @@ KeyList Score::keyList() const
     // find the keylist of the first pitched staff
     KeyList tmpKeymap;
     Staff* firstStaff = nullptr;
+    bool isFirstStaff = true;
     for (Staff* s : masterScore()->staves()) {
         if (!s->isDrumStaff(Fraction(0, 1))) {
             KeyList* km = s->keyList();
-            tmpKeymap.insert(km->begin(), km->end());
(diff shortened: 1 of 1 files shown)


asattely:
Wow xcode really doesn't like that one. Gimme a bit to write it in a more OS-inclusive way!

src/engraving/libmscore/score.cpp
```diff
@@ -2795,10 +2795,40 @@ KeyList Score::keyList() const
...
         if (!s->isDrumStaff(Fraction(0, 1))) {
+            static bool firstStaff = true;
             KeyList* km = s->keyList();
-            tmpKeymap.insert(km->begin(), km->end());
-            firstStaff = s;
-            break;
+            if (firstStaff) {
+                firstStaff = false;
+                tmpKeymap = *km;
+                firstStaff = s;
```
cbjeukendrup, line 2803: Personally I'm confused by this section... two variables with the same name? A static variable?

asattely: Sorry, pushed a bad version, i have pushed the fix for that and also fixed what i think xcode was angry about

oktophonie (approved):

RomanPudashkin (approved):

PR #14007, opened 2022-10-25
fixed #10399: Closing the last score shouldn't close MU

### LINKED ISSUE #10399: [MU4 Issue] Closing the last score shouldn't close MU (if many scores were opened)
https://user-images.githubusercontent.com/90187801/151548181-4dbeabb6-08ea-4e66-a593-c32135628c21.mp4

**To Reproduce**
1. Open score A
2. Open score B 
3. Close score B (with X button)
4. Close score A (with X button) - it should not close MU 

**Platform information**
 - OS: Win10 (didn't try Mac, Lin)

Eism:
Resolves: #10399

When the program instance is closed -> the sockets are destroyed -> the server updates the list of open instances.

What happened:
The user closes the score and the window closes. The program starts cleaning up the created objects, one of which is MidiDevicesListener. The listener contains a thread in which the list of available devices is updated, to optimize resource consumption, the thread goes to sleep for 500ms after each check. In the listener's destructor, the thread is terminated, but in order to terminate the thread, the program needs to wait until it wakes up. There are two such listeners: midiIn + midiOut = 1s. The user can close the main instance during this time.

The same with AudioThread

Notify another instances about quitting before quitting application

### DIFF
--- a/src/appshell/internal/applicationactioncontroller.cpp
+++ b/src/appshell/internal/applicationactioncontroller.cpp
@@ -170,6 +170,10 @@ void ApplicationActionController::quit(bool isAllInstances, const io::path_t& in
             interactive()->openUrl(QUrl::fromLocalFile(installerPath.toQString()));
         }
 
+        if (multiInstancesProvider()->instances().size() > 1) {
+            multiInstancesProvider()->notifyAboutInstanceWasQuited();
+        }
+
         QCoreApplication::quit();
     }
 }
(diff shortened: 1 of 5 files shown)


igorkorsukov (approved):

DmitryArefiev:
Tested #10399 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #14015, opened 2022-10-25
changed layout ordering of staff text, system text, instrumentchanged

handrok:
Stave text is closer to stave than system text because of their names. Stave is a part of System so it looks like stave text shoud be closer to stave than system text.


alexpavlov96 (approved):

abariska:
🟢 **TESTED**
Please, merge to master

Issue #14058, opened 2022-10-26
[MU4 Issue] Crash when extract all parts from some old big scores 


DmitryArefiev:
**To Reproduce**
1. Open [Test..._Beethoven__Symphony_No._9__Op._125.zip](https://github.com/musescore/MuseScore/files/9869040/Test._Beethoven__Symphony_No._9__Op._125.zip) or [Gluck_-_Iphignie_en_Aulide__Vocal_Score.zip](https://github.com/musescore/MuseScore/files/9869044/Gluck_-_Iphignie_en_Aulide__Vocal_Score.zip)
2. Open Parts
3. Open all

https://user-images.githubusercontent.com/90187801/198017105-941fe50b-7db0-476c-b212-65754f13a79a.mp4

**Platform information**
macOS, Windows

**Additional context**

lyrra:
Shrinked that gigantic mscz file into this mscx, crashes with the same backtrace which suggest same bug is triggered

```
<?xml version="1.0" encoding="UTF-8"?>
<museScore version="4.00">
  <Score>
    <Part id="1">
      <Staff id="1"> </Staff>
      </Part>
    <Staff id="1">
      <TBox> </TBox>
      <Measure>
        </Measure>
      </Staff>
    </Score>
  </museScore>
```

<details><summary>backtrace</summary>

```
Backtrace from original mscx file:
#0  0x00005555559b67d0 in mu::engraving::MeasureBase::prev (this=0x0) at src/engraving/libmscore/measurebase.h:112
#1  0x00005555559960e4 in mu::engraving::MeasureBaseList::fixupSystems (this=0x5555ba02c9f0) at src/engraving/libmscore/score.cpp:296
#2  0x0000555555995de8 in mu::engraving::MeasureBaseList::insert (this=0x5555ba02c9f0, fm=0x5555b5bd3a60, lm=0x5555b5bd3a60) at src/engraving/libmscore/score.cpp:227
#3  0x0000555555af7f68 in mu::engraving::InsertRemoveMeasures::insertMeasures ( this=0x5555b5bd3c10) at src/engraving/libmscore/undo.cpp:2041
#4  0x0000555555cdb1a2 in mu::engraving::InsertMeasures::redo (this=0x5555b5bd3c10) at src/engraving/libmscore/undo.h:913
[...]

PR #14068, opened 2022-10-26
fix 14058

lyrra:
Resolves: #14058 

- [ x] I signed [CLA](https://musescore.org/en/cla)


RomanPudashkin:
Unfortunately, these changes do not fix the problem

PR #14079, opened 2022-10-26
Fix #13986: Adjust spanners when measures are deleted

### LINKED ISSUE #13986: [MU4 Issue] Slurs are not deleted when measures was deleted
**Describe the bug**
The issue is reproducible for deleting measures from score or from part

**To Reproduce**
1. Open the score 
[Mozart_Symphony_No._40_in_G_Minor_K._550_I._Molto_Allegro 3.mscz.zip](https://github.com/musescore/MuseScore/files/9853258/Mozart_Symphony_No._40_in_G_Minor_K._550_I._Molto_Allegro.3.mscz.zip)
2. Delete few measures with slurs (e.g. 19-21)
3. Pay attention to extra slurs on the score
4. Go Undo
5. Open any part 
6. Delete few measures 
7. Go to the full score and check extra slurs

**Expected behavior**
Related slurs are deleted along with measures

**Screenshots**

https://user-images.githubusercontent.com/101250347/197576094-4bd7b1b9-0ead-4aa2-b6e8-0462f4dddd4c.mp4

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/13986

First, if the highlighted measure here is deleted, the endpoints of the slur do not change:
image

If these highlighted measures are deleted, the slur's start point will be moved up to the first beat of the third measure:
image

(The same will happen when the second two measures are deleted--its endpoint will adjust back to the 4th beat of the first measure)

Finally, when this measure is deleted, all of these slurs will be removed as well:
image

The middle slur will be deleted because both of its endpoints are inside of the removed range. The outer slurs could potentially be adjusted backwards like the ones above, except that would put the endpoint of the slur on the same note as the startpoint. So instead of dealing with zero-length slurs, they are removed.

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -2809,6 +2809,12 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
         }
     }
 
+    if (!preserveTies) {
+        // we're not preserving slurs and ties where they are, so the endpoints need to be either
+        // moved or the slur has to be deleted
+        reconnectSlurs(mbStart, mbEnd);
+    }
+
     // get the last deleted timesig & keysig in order to restore after deletion
     KeySigEvent lastDeletedKeySigEvent;
     TimeSig* lastDeletedSig   = 0;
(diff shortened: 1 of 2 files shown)


asattely:
Grace notes are also taken into consideration when considering what the last (or first) cr of a measure is:
image

When the highlighted measures are deleted, the slurs adjust to the appropriate grace note rather than deleting because they're on the same tick.

oktophonie (approved):

src/engraving/libmscore/edit.cpp
```diff
@@ -2801,6 +2801,116 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
...
+        // Reconnect or remove slurs that intersect with these deleted measures
+        Fraction sTick = mbStart->tick();
+        Fraction eTick = mbEnd->tick();
+        for (auto overlappingSlur : spannerMap().findOverlapping(sTick.ticks(), eTick.ticks())) {
+            Spanner* sp = overlappingSlur.value;
+            if (!sp->isSlur()) {
+                continue;
+            }
+            ChordRest* cr1 = sp->startCR();
+            ChordRest* cr2 = sp->endCR();
```
cbjeukendrup, line 2814: Don't we need a sanity check before using cr1 and cr2? The slur layout code contains many of those.

cbjeukendrup (commented):
Looks good to me for the rest, over to @RomanPudashkin!

src/engraving/libmscore/edit.cpp
asattely: Certainly couldn't hurt! Added.

src/engraving/libmscore/edit.cpp
```diff
@@ -2808,6 +2808,120 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
...
+            if (!cr1 || !cr2) {
+                // this is an invalid slur
+                continue;
+            }
+            MeasureBase* m = mbStart;
+            bool adjust1 = false;
+            bool adjust2 = false;
+            do {
+                adjust1 = adjust1 ? true : cr1->measure() == m;
```
RomanPudashkin, line 2830: You should call these cr1->measure()/cr2->measure() outside of the while loop

src/engraving/libmscore/edit.cpp
```diff
@@ -2808,6 +2808,120 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
...
+                // search for last cr in this voice
+                Segment* seg = mNext->first(SegmentType::ChordRest);
+                while (seg) {
+                    firstAvailableCr = seg->cr(sp->track());
+                    if (!firstAvailableCr || !firstAvailableCr->isChord()) {
+                        seg = seg->next(SegmentType::ChordRest);
+                        continue;
+                    }
+                    auto gracesBefore = toChord(firstAvailableCr)->graceNotesBefore();
+                    if (firstAvailableCr->tick() > sp->endCR()->tick()) {
```
RomanPudashkin, line 2859: Please call these sp->endCR() (and the other things that are not changes between iterations) outside the loop

src/engraving/libmscore/edit.cpp
```diff
@@ -2808,6 +2808,120 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
         }
     }
 
+    if (!preserveTies) {
```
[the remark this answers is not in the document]
asattely: Done!

PR #14085, opened 2022-10-27
fixed hiding of articulations

alexpavlov96:


handrok (approved):

cbjeukendrup:
@alexpavlov96 Could you please explain which problem this is solving?
Currently you're only checking for the "above" variants of articulations, e.g. `articStaccatoAbove`. How can you be sure that the "below" variants will never occur, e.g. `articStaccatoBelow`? 
Also, there's a bug with this: if you add the "accent staccato" articulation (i.e. both an accent and a staccato marking on the same chord), then both will be shown, regardless of any style setting.

Also tagging @mike-spa, who has worked on articulations recently.

alexpavlov96:
@cbjeukendrup thanks, you're right! Please review: https://github.com/musescore/MuseScore/pull/14099

mike-spa:
Hey, I know it takes time and it is often annoying, but it is a good thing to provide a short description of what the PR is trying to solve (and maybe a screenshot, when it involves engraving stuff). That way we are also aware of what is going on and, like in this case, we could have helped sooner :) Anyway, I've approved the new PR, it is indeed just a matter of calling `computeCategories()` before they are needed :)

Tantacrul:
From now on all PRs that change behaviour in MuseScore's engraving **must** include an explanation and **must** put @oktophonie or @mike-spa as a reviewer.

@vpereverzev - just so you can see.

Issue #14037, opened 2022-10-25
[MU4 Issue] Chord symbols placed on the first beat of a measure and attached to an invisible rest get placed too close to the barline


Riisus:
**Describe the bug**
As the title says, if you attach a chord symbol on an invisible rest on the first beat of a measure, it gets placed too close to the preceding barline.

**To Reproduce**
1. Go to any score
2. Place a chord on a rest in the first beat of a measure that has a barline before it
3. Make the rest attached to the chord symbol invisible (If there are other instruments, they must have an invisible rest there as well).

**Expected behavior**
Chord symbols placed on invisible rests remain correctly offset from a preceding barline.

**Screenshots**
<img width="760" alt="chord symbol offset" src="https://user-images.githubusercontent.com/116495080/197857090-dff9a483-c002-4176-ae0a-0e2e92079da2.png">
<img width="678" alt="chord symbol offset 2" src="https://user-images.githubusercontent.com/116495080/197858497-7c909f8b-c1e0-4bf4-8349-86f781c2f8d3.png">

**Platform information**
 - MacOS 13.0

oktophonie:
Further, I'd say the invisible rest is too close to the barline.

Riisus:
Yes they are, but only ones with a chord symbol attached to them. Invisible rests without a chord symbol are fine.

oktophonie:
Even worse!

PR #14086, opened 2022-10-27
Invisible chord spacing fix

mike-spa:
Resolves: #14037 

While working on this issue, I've found that some general code clean up was _really_ necessary.

- Commit 1: eliminated a duplicate method (`hasAccidental(Segment* )` in measure.cpp was duplicating `Segment::hasAccidental()`. The method should only exist in the Segment class.)
- Commit 2: cleaned up the shape/bounding box logic for Harmony items. Specifically, horizontal padding used to be implemented by artificially increasing the size of the bounding boxes. That is a _bad_ idea. The bounding box should always represent the minimum area occupied by the item and the padding should be added _between_ the bounding boxes. There are a few more places where this old idea is still used (most notably, lyrics). Those will be gradually also fixed in future. Also, for some reason, the bounding box of Harmony items was added twice (?).
- Commit 3: actual bug fix. Correcting some logic to cope better with the spacing of invisible items.

All the vtest differences are due to the correction in harmony bounding box calculations. And I've added one more vtest to protect the bug fix.


oktophonie (approved):

src/engraving/libmscore/measure.cpp
```diff
@@ -4408,6 +4411,16 @@ double Measure::computeFirstSegmentXPosition(Segment* segment)
         // The start-repeat should overlap the end-repeat of the previous measure
         x -= score()->styleMM(Sid::endBarWidth);
     }
+    // Do a final check of chord distances (invisible items may in some cases elude the 2 previous steps)
+    if (segment->isChordRestType()) {
+        double barNoteDist = static_cast<double>(score()->styleMM(Sid::barNoteDistance));
```
cbjeukendrup, line 4416: I think the preferred way is calling `.val()`: 
`double barNoteDist = score()->styleMM(Sid::barNoteDistance).val()`.

cbjeukendrup (commented):
For the rest looks good to me, over to @RomanPudashkin!

cbjeukendrup:
One of the commits here turned out to be useful for #14312. So, @RomanPudashkin, it would be great to merge this relatively soon :)

cbjeukendrup:
@mike-spa A rebase is needed :)

src/engraving/libmscore/measure.cpp
cbjeukendrup: @mike-spa It seems like this comment got overlooked :)

PR #14092, opened 2022-10-27
Fix #13858: Removed velocity settings for dynamics and hairpins

### LINKED ISSUE #13858: Remove velocity settings for dynamics and hairpins
**Task description**
This is a step we need to take to have a simpler, universal system for velocity change in MuseScore 4. In MuseScore 3, there are competing velocity parameters which the user then needs to be able to learn to either combine or prioritise. It is too complex and unworkable for the future.

So for MS4, velocity can only be placed on notes themselves.

Hence, these options need to be removed of hairpins and dynamics:

image

image

---

Moreover, we now should also remove the tab in 'playback' that distinguishes between 'General' and 'Dynamics'
image

HemantAntony:
Resolves: #13858


Tantacrul:
This MacOS check is taking forever :)

Eism:
restarted

Tantacrul:
OK, just tested. Looking good!

@Eism or @RomanPudashkin - if you're happy, can this baby be floated down the river merge?

src/inspector/view/qml/MuseScore/Inspector/general/playback/PlaybackDynamicsSettings.qml
```diff
@@ -55,13 +55,6 @@ Item {
         }
 
         SeparatorLine { anchors.margins: -12 }
```
Eism, line 57: line should also be removed

HemantAntony: I have updated the PR again :)

PR #14096, opened 2022-10-27
fixed #11846: Content is shaking when scrolling on trackpad

### LINKED ISSUE #11846: [MU4 Issue] [macOS] Content is shaking when scrolling on trackpad
**Describe the bug**
The issue is reproducible in cases when scrolling with inertia 

**To Reproduce**
1. Open any score
2. Open lots of pallets in inspector (scroll bar should appear )
3. Scroll a bit and leave trackpad to get inertia scrolling 
4. Try to do it with different intensity 

**Expected behavior**
Content is moving smoothly when scrolling on trackpad 

**Screenshots**

https://user-images.githubusercontent.com/101250347/171407344-355675d4-b22c-49ef-a6db-7515a01037b3.mov

**Platform information**
 - OS: macOS

Eism:
Resolves: #11846

### DIFF
--- a/src/appshell/CMakeLists.txt
+++ b/src/appshell/CMakeLists.txt
@@ -118,6 +118,8 @@ if (OS_IS_MAC)
     set(MODULE_SRC ${MODULE_SRC}
         ${CMAKE_CURRENT_LIST_DIR}/view/internal/platform/macos/macosappmenumodelhook.mm
         ${CMAKE_CURRENT_LIST_DIR}/view/internal/platform/macos/macosappmenumodelhook.h
+        ${CMAKE_CURRENT_LIST_DIR}/view/internal/platform/macos/macosscrollinghook.cpp
+        ${CMAKE_CURRENT_LIST_DIR}/view/internal/platform/macos/macosscrollinghook.h
     )
     set_source_files_properties(
         ${CMAKE_CURRENT_LIST_DIR}/view/internal/platform/macos/macosappmenumodelhook.mm
--- a/src/appshell/appshellmodule.cpp
+++ b/src/appshell/appshellmodule.cpp
@@ -66,6 +66,7 @@
 #ifdef Q_OS_MAC
 #include "view/appmenumodel.h"
 #include "view/internal/platform/macos/macosappmenumodelhook.h"
+#include "view/internal/platform/macos/macosscrollinghook.h"
 #else
 #include "view/navigableappmenumodel.h"
 #endif
@@ -80,6 +81,10 @@ static std::shared_ptr<ApplicationActionController> s_applicationActionControlle
 static std::shared_ptr<ApplicationUiActions> s_applicationUiActions = std::make_shared<ApplicationUiActions>(s_applicationActionController);
 static std::shared_ptr<AppShellConfiguration> s_appShellConfiguration = std::make_shared<AppShellConfiguration>();
 static std::shared_ptr<SessionsManager> s_sessionsManager = std::make_shared<SessionsManager>();
+
+#ifdef Q_OS_MAC
+static std::shared_ptr<MacOSScrollingHook> s_scrollingHook = std::make_shared<MacOSScrollingHook>();
+#endif
 
 static void appshell_init_qrc()
 {
@@ -189,6 +194,10 @@ void AppShellModule::onInit(const IApplication::RunMode&)
     s_applicationActionController->init();
     s_applicationUiActions->init();
     s_sessionsManager->init();
+
+#ifdef Q_OS_MAC
+    s_scrollingHook->init();
+#endif
 }
 
 void AppShellModule::onDeinit()
--- a/src/appshell/view/internal/platform/macos/macosscrollinghook.cpp
+++ b/src/appshell/view/internal/platform/macos/macosscrollinghook.cpp
@@ -1,0 +1,46 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "macosscrollinghook.h"
+
+#include <QWheelEvent>
+
+using namespace mu::appshell;
+
+void MacOSScrollingHook::init()
+{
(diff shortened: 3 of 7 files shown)


bkunda:
@Eism This is mostly fine – I can't see any shaking, and it even solves my problem with the Logitech MX Master 3, so I'm delighted! 

The inertia time does seem to have increased though. I'm wondering if it's at all possible to still reduce this?
Hope this video helps! 

https://user-images.githubusercontent.com/86290556/198546165-085f210c-a9f3-451f-87b5-3a144bada946.mp4

cbjeukendrup:
@Eism For me, this seems to make scrolling with the trackpad almost unusable. If I move my fingers just a millimetre, MuseScore already scrolls way to far.

However, I did find a variation on your solution that seems to work well for me:
```c++
bool MacOSScrollingHook::eventFilter(QObject* watched, QEvent* event)
{
    if (event->type() != QEvent::Wheel) {
        return false;
    }

    QWheelEvent* wheelEvent = dynamic_cast<QWheelEvent*>(event);
    if (wheelEvent->phase() == Qt::ScrollEnd) {
        return true;
    }

    return false;
}
```

(Also, since the scrolling hook only uses Qt stuff, I see no reason to make it an ObjC++ file. I'd simply use `.cpp`.)

bkunda:
This is definitely an issue worth testing on multiple machines I think. 
@Eism when you're ready for more testers, I'd suggest also asking @shoogle and @Tantacrul who both use Macs.

Tantacrul:
Gimme the nod when it's ready for testing and I'll try it.

Tantacrul:
Quick question though: why are we focusing attention on a P3? The jittering trackpad problem is annoying but not a dealbreaker. I wouldn't spend too much time on this.

Eism:
@cbjeukendrup Yes, it looks better, thanks.

@Tantacrul I think this is a critical bug and we need to fix it before release. This bug annoyed me every single day on mac.

Tantacrul:
> @cbjeukendrup Yes, it looks better, thanks.
> 
> @Tantacrul I think this is a critical bug and we need to fix it before release. This bug annoyed me every single day on mac.

@Eism - that's the best possible reason then. Thanks!

Eism:
@bkunda @Tantacrul 
ready for testing

bkunda:
I'm not getting any shakiness when I use the trackpad, and for me the inertia level feels just about right, so this is all good.

Unfortunately, my old problem with the MX Master 3 is back (when I scroll in some menus, nothing happens unless I manually click and drag the scrollbar first), but fixing that was never the point of the issue in question – although it is interesting that the problem was fixed for a moment there 🤔 I suspect this is something that I'll have to raise separately?

If @cbjeukendrup and @Tantacrul find no issues with this, then it's probably looking good.

cbjeukendrup:
@bkunda I'm afraid that the fix for the MX Master 3 problem is the same thing that created the hyper-scroll-speed / inertia problem on my trackpad.

Basically, scroll data is reported in two ways: pixel delta and angle delta. Pixel delta is the amount of pixels to be scrolled, calculated by the OS (not all mouse+OS combinations support this). Angle delta is the "raw" amount of steps scrolled with the mouse. 
Apple's trackpads and Magic Mouses report pixel delta, which works well. However, the angle data of these mouses is not reliable, apparently (it creates the hyper-scroll-speed / inertia problem).
For the MX Master 3, it is apparently the other way around: it does report pixel delta, but that is unreliable; instead, the angle delta does work.

Elnurs initial fix was to kill the pixel delta completely, so that always the angle delta was used as a fallback. So, that fixes the MX Master 3, but makes life much worse for Apple's own mouses and trackpads.

Personally I'm happy with the solution as it is now!

shoogle:
Perhaps this situation merits a new option in Preferences > Advanced?

    application/filterScrollingEvents (default: checked)

Then we can tell people with the MX Master 3 mouse to uncheck this option and everyone is happy.

Personally I haven't experienced the issue with scrolling on macOS Big Sur and I don't notice any difference in scrolling with this PR compared to the latest nightly build. (Lucky me! 😉)

Tantacrul:
I have discovered a reproducible crash with this PR. You can see it happen at the start of the video and also the end too. I do like this PR because it gets rid of the really annoying trackpad jitter (which I experience constantly on MacOS).

However, the easily replicated crashes and strange Palettes behaviour needs a look.
https://www.dropbox.com/s/e9kwu6zo75o1rr6/Scrolling-Crash.mov?dl=0

cbjeukendrup:
The reason that Palettes are different from other things is probably the following: in palettes, each "row" might have a different height. The way QML's list views work is that they only render the currently visible rows, and extrapolate the total height based on the visible ones. For palettes that doesn't work because every row is different, so that's why you get the weird over scroll problem.

However, _apparently_, QML does depend on the "scrolling has ended" event, because in reaction on that event it "fixes" the situation. As an alternative solution, we can try what happens if we set the ScrollPhase of _all_ scroll events to NoPhase.

The crash seems to be related to our tooltips.

bkunda:
@Eism I think this is a "back over to you" situation.
Let us know when you're ready for more testing :-) 

Also I don't mind @shoogle's suggestion of exposing a setting in Preferences > Advanced for MX Master 3 mouse users if it's the _only_ way to solve the problem at the moment (Obviously configuring the scrolling behaviour to work universally well across all devices is preferable though). I'd take consolation in the fact that people who encounter this issue will only have to research the solution and change the setting _once_ to fix their problem.

And @Tantacrul's video showed some really annoying issues that I've also noticed: large empty space at the end of the scroll.

Eism:
@Tantacrul @bkunda @cbjeukendrup 
A couple more changes to fix the issues that Martin mentioned
Please check on your machines

Tantacrul:
Wow. This works really nicely on my Mac! I can't find any problems with it! Nice work :)

cbjeukendrup:
I can still reproduce the problem with palettes... if I add all palettes, and then expand all of them, and then scroll really fast, I still get this:
Scherm­afbeelding 2022-11-05 om 18 51 25

I can't reproduce the crash anymore though.

However, the side effect of the fix for crash is that when you hover a button in a non-scrollable area (for example, note input bar) so that a tooltip appears, and you touch your scroll wheel or make a scroll gesture on your trackpad, the tooltip will disappear. Anyway, that's absolutely no big deal to me; I assume we can live with that.

Tantacrul:
Interestingly, I'm unable to replicate the palettes scrolling issue. 

1. Regarding the observation that the scrollwheel / trackpad dismisses tooltips - I don't think we should spend any time on this. It doesn't feel like undesirable behaviour and I doubt users would suffer from any problems here.

I suggest (assuming a clear dev review and no more bugs with this implementation) that we merge this and move on to more significant issues.

Thanks @Eism! Great work :)

cbjeukendrup:
When I add at least the "Bagpipe embellishments" palette (that super tall one), I can really easily replicate the scrolling issue. But this is probably the best situation we can get right now; I agree that we should maybe not spend much more time on this issue. And when we switch to Qt 6, all trouble is over anyway... (well, that's a massive overstatement, but it looks like they did fix some scrolling issues.)

Tantacrul:
Ah, you're right. It is reproducible with the bagpipe embellishments palette.

Yeah, I don't think we should spend any more time on this - especially since the Bagpipe Embellishments palette is complete nonsense anyway. We should just replace it with a sane replacement in future (I know a person who has already suggested a way better solution).

@Eism - I think we're good to go!

Eism:
I agree

cbjeukendrup (approved):

PR #14106, opened 2022-10-27
making fret circle not selectable and fixed bounding rect

alexpavlov96:


cbjeukendrup (approved):

src/engraving/libmscore/fretcircle.cpp
```diff
@@ -68,10 +67,13 @@ void FretCircle::layout()
...
     if (!tabEllipseEnabled()) {
         _skipDraw = true;
+        setbbox(RectF());
         return;
     }
 
-    setbbox(ellipseRect());
+    double lw = spatium() * circleWidth;
+    m_rect = ellipseRect();
+    setbbox(m_rect.adjusted(-lw, -lw, lw, lw));
```
cbjeukendrup, line 76: Actually, you need to adjust only by the _half_ line width. The ellipse is drawn in such way that one half of the line is inside the rect, and the other half outside.

cbjeukendrup (approved):

Issue #14123, opened 2022-10-28
[MU4 Issue] Exported audio should ignore selection and always export full score audio


MarcSabatella:
**Describe the bug**
If you happen to have a range selected when you publish to musescore.com, only the select staves are exported

**To Reproduce**
1. Open a score for multiple instruments
2. Select one measure in one instrument
3. File / Export or File / Publish to musescore.com
4. Listen to exported/published score
6. Hear error: only the select staff plays

**Expected behavior**
Full score should play

**Additional context**
The feature of soloing selected staves is nice although still a bit divisive in that many of us continue to accidentally invoke it because it's so much easier to select a measure rather than a note to start playback.  But, at least we realize our mistake instantly when we don't hear what we expect.  Whereas, when exporting audio or publishing to musescore.com, it might literally take an hour to export the audio for a large score with Muse Sounds, and it's *incredibly* frustrating to then realize we have to redo it.  I would claim that as useful as soloing staves by selection might be, it should be ignored for export / publish.

Zanctarian:
It'd be nice for expandability upon that feature as well. I get so frustrated when I accidentally click the measure instead of the note! It'd be nice to be able to click, lets say, the instrument name on the left side instead.. or at the very least a toggle to enable or disable the feature!

DaddyLudwig:
I tend to click on the measure too as force of habit from earlier MuseScore versions.

cbjeukendrup:
I'll work on this, this is a pretty bad bug and also affects score uploading to .com.

@Tantacrul Extra question: should solo/mute state in the Mixer be honoured by audio export?

teetow:
> Extra question: should solo/mute state in the Mixer be honoured by audio export?
It's a reasonable expectation, especially if you're looking to export a "no soloist" version, for example. This would be a lot of extra work otherwise -- exporting separated tracks and re-assembling them in a DAW.

Tantacrul:
Solo / Mute should absolutely be respected and I agree 100% that selection should have no bearing on what is exported.

cbjeukendrup:
Okay. That's what #14129 does :)

MarcSabatella:
[...]

PR #14129, opened 2022-10-28
Don't solo selected staves during audio export

cbjeukendrup:
Resolves: #14123
Resolves: #14265
It's a somewhat dirty fix, but I'm not sure if we can do much better right now.

### DIFF
--- a/src/playback/internal/playbackcontroller.h
+++ b/src/playback/internal/playbackcontroller.h
@@ -105,6 +105,8 @@ public:
 
     void applyProfile(const SoundProfileName& profileName) override;
 
+    void setExportingAudio(bool isExporting) override;
+
     bool canReceiveAction(const actions::ActionCode& code) const override;
 
 private:
@@ -211,6 +213,8 @@ private:
 
     framework::Progress m_loadingProgress;
     std::list<engraving::InstrumentTrackId> m_loadingTracks;
+
+    bool m_isExportingAudio = false;
 };
 }
 
--- a/src/playback/internal/playbackcontroller.h
+++ b/src/playback/internal/playbackcontroller.h
@@ -105,7 +105,8 @@ public:
 
     void applyProfile(const SoundProfileName& profileName) override;
 
-    void setExportedNotation(notation::INotationPtr notation) override;
+    void setNotation(notation::INotationPtr notation) override;
+    void setIsExportingAudio(bool exporting) override;
 
     bool canReceiveAction(const actions::ActionCode& code) const override;
 
@@ -214,8 +215,7 @@ private:
     framework::Progress m_loadingProgress;
     std::list<engraving::InstrumentTrackId> m_loadingTracks;
 
-    /// The notation for which we're currently exporting audio
-    notation::INotationWeakPtr m_exportedNotation;
+    bool m_isExportingAudio = false;
 };
 }
 
--- a/src/playback/internal/playbackcontroller.h
+++ b/src/playback/internal/playbackcontroller.h
@@ -105,7 +105,7 @@ public:
 
     void applyProfile(const SoundProfileName& profileName) override;
 
-    void setExportingAudio(bool isExporting) override;
+    void setExportedNotation(notation::INotationPtr notation) override;
 
     bool canReceiveAction(const actions::ActionCode& code) const override;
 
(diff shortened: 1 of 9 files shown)


vpereverzev:
1. Remove dependency from IPlaybackController as well as setExportingAudio method
2. Add notification to IPlayback facade `virtual async::Channel<RenderMode> modeChanged() const = 0;`
3. Add implementation of this method to playback.h
4. Subscribe Playback facade 
```
AudioEngine::instance()->modeChanged().onNotify([this]() {
    m_modeChangedChannel.send(AudioEngine::instance()->mode());
}
```
5. Subscribe PlaybackController on IPlayback::modeChanged notifications and do your things.

AudioEngine is the only one who knows what is the current mode and why, also it's not desirable to add dependencies between modules on IPlaybackController unless we don't have any other options

cbjeukendrup:
@vpereverzev I'm afraid of one problem: will it not be too late if PlaybackController listens for notifications from the audio engine?
Wouldn't we get something like this?

1. [audio worker thread] set render mode to offline
2. [audio worker thread] immediately start rendering audio file
3. [main thread] oh, uhm, hey, the render mode has been changed. Let's update the mute states of the tracks. 
Oh, oops, audio export was already started, so now it's wrong for the first few milliseconds of the audio file.

Or is that nonsense?

MarcSabatella:
FWIW, at least in the past, a number of other commands that had issues with active selection would simply clear the selection before executing, then restore it afterwards.  So, that's another approach to try if this one seems not right.

cbjeukendrup:
@vpereverzev I just made this a little bit worse so that it now also resolves #14265 😄

src/playback/internal/playbackcontroller.h
```diff
@@ -211,6 +213,9 @@ class PlaybackController : public IPlaybackController, public actions::Actionabl
 
     framework::Progress m_loadingProgress;
     std::list<engraving::InstrumentTrackId> m_loadingTracks;
+
+    /// The notation for which we're currently exporting audio
+    notation::INotationPtr m_exportedNotation = nullptr;
```
RomanPudashkin, line 218: You should use INotationWeakPtr here

vpereverzev:
@cbjeukendrup 
I guess it's fine to work this out via playback controller in the context of 2 attached issues. The only problem I can see here - instead of `setExportedNotation ` method it'd be better to switch "current notation" from global context onto desired one. PlaybackController would react on thins notification at "PlaybackController::onNotationChanged()". Then when exporting has been finished - switch current notation back

vpereverzev:
> @cbjeukendrup I guess it's fine to work this out via playback controller in the context of 2 attached issues. The only problem I can see here - instead of `setExportedNotation ` method it'd be better to switch "current notation" from global context onto desired one. PlaybackController would react on thins notification at "PlaybackController::onNotationChanged()". Then when exporting has been finished - switch current notation back

have you tried that already?

vpereverzev:
> > @cbjeukendrup I guess it's fine to work this out via playback controller in the context of 2 attached issues. The only problem I can see here - instead of `setExportedNotation ` method it'd be better to switch "current notation" from global context onto desired one. PlaybackController would react on thins notification at "PlaybackController::onNotationChanged()". Then when exporting has been finished - switch current notation back
> 
> have you tried that already?

if that causes some undesirable problems - add method "PlaybackController::setNotation" and call it from "PlaybackController::onNotationChanged" and from "AbstractAudioWriter::doWriteAndWait"

cbjeukendrup:
@vpereverzev I think something like your second suggestion would be better indeed, because otherwise the whole UI would switch to the different notation, which is maybe not desirable and of course comes with a performance cost.

However, I still see some difficulties. In PlaybackController::onNotationChanged, the controller subscribes itself to a bunch of notifications, and I don't know if I should move this to the new `setNotation` method or not. 
- if I do move this code, then the controller will be subscribed to a lot of notifications for the exported notation, while that is not necessary at all.
- if I don't move this code, then the controller will still be subscribed to the notifications of the old notation, but `m_notation` is the exported notation. So actually the controller is in an inconsistent state in this case.

Since the UI is currently blocked during audio export, these notifications are probably not received anyway, but still it feels not like a very clean solution to me. At least it will not be more elegant than with a simple `setExportedNotation` method, I'm afraid. I'll give it another try though.

cbjeukendrup:
@vpereverzev I gave it a try; please have a look :)

DmitryArefiev:
Tested #14123 #14265 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

vpereverzev:
I'd not cherry pick that one. @cbjeukendrup please make a separate PR for RC

Issue #11211, opened 2022-04-15
[MU4 Issue] "Hide instrument name if there is only 1 instrument" not always applied


RootCubed:
**Describe the bug**
The style setting "Hide instrument name if there is only 1 instrument" (Style -> Score) is not honoured after adding and removing instruments to/from a part score. This happens in the master score too, but there it might be intended behaviour (?).

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a score with at least two instruments.
2. Open the part score of an instrument.
3. Observe that the instrument name is hidden, since the option is enabled by default.
4. Make another instrument visible, then invisible.
5. Observe that the instrument name is not hidden anymore, even though the option is enabled.

Video demonstration attached.

https://user-images.githubusercontent.com/41953062/163548692-9eb25634-0778-4135-914e-de0c385134fd.mp4

**Expected behavior**
The style setting "Hide instrument name if there is only 1 instrument" should always be applied.

**Platform information**
 - OS: Windows (but most likely all)

Tantacrul:
@oktophonie - not sure if this is an issue you guys should take on (I'm guessing not) but thought I'd alert you to it just in case.

oktophonie:
Score and parts should behave the same, I think. And we should interpret "when there is 1 instrument" to mean "when only 1 instrument is visible" (rather than "when the score contains only 1 instrument" - an academic distinction now, really, since it's so easy to show/hide instruments in parts.) Will sort.

PR #14133, opened 2022-10-28
Fix visibility of instrument name when single part

mike-spa:
Resolves: #11211

Given the way that parts interact with the instrument panel, the decision of showing the name shouldn't be taken based on how many parts are in the score, but rather how many of them are visible.

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -5375,6 +5375,17 @@ Part* Score::partById(const ID& partId) const
     return nullptr;
 }
 
+int Score::visiblePartCount() const
+{
+    int count = 0;
+    for (const Part* part : _parts) {
+        if (part->show()) {
+            ++count;
+        }
+    }
+    return count;
+}
+
 ShadowNote& Score::shadowNote() const
 {
     return *m_shadowNote;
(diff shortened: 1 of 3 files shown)


src/engraving/libmscore/score.cpp
```diff
@@ -5375,6 +5375,17 @@ Part* Score::partById(const ID& partId) const
     return nullptr;
 }
 
+int Score::visibleParts() const
```
RomanPudashkin, line 5378: It should be renamed to `visiblePartCount`

src/engraving/libmscore/score.cpp
```diff
@@ -5375,6 +5375,17 @@ Part* Score::partById(const ID& partId) const
...
+{
+    int count = 0;
+    for (Part* part : _parts) {
```
RomanPudashkin, line 5381: `const Part* part`

RomanPudashkin (approved):

src/engraving/libmscore/score.cpp
mike-spa: 👍

src/engraving/libmscore/score.cpp
mike-spa: 👍

oktophonie (approved):

Issue #13914, opened 2022-10-23
[MU4 Issue] Cut off endings of text labels in the Preferences dialog


prokoudine:
**Describe the bug**

When running MuseScore in Russian, some text labels in the Preferences dialog seem to be ellipsized for no good reason. There's plenty of free space to display the entire text. I checked the translation, it's somewhat longer than the original text in English, but it isn't in the "OH MY GOD, IT'S HUGE!!!11!" department :)

I haven't seen this anywhere else yet (but the two messages in the screenshot below), but then again not all of the Preferences dialog has been translated yet. 

**Screenshot**

![image](https://user-images.githubusercontent.com/57467/197391351-0ed7981e-3935-4660-944a-b97bc4361afb.png)

**Platform information**

OS: Ubuntu 22.04.1 LTS, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-3289917644, revision: github-musescore-musescore-1ac395a

Issue #13773, opened 2022-10-15
[MU4 Issue] Various text clipping issues


cbjeukendrup:
**Describe the bug**
Due to localisation and other reasons, various texts require more space than anticipated, causing clipping and other problems.
Let's collect these problems in this issue.

- [ ] MuseScore 3.6 migration dialog with Dutch (Nederlands) localisation ("MuseScore 3.…")
    <img width="400" alt="Scherm­afbeelding 2022-10-15 om 14 53 01" src="https://user-images.githubusercontent.com/48658420/195987797-26c20f89-b4aa-457f-bbd1-efd86563b97d.png">
- [ ] #15537
- [ ] #15411
- [ ] #15395
- [x] Cloud upload success dialog in German: https://musescore.org/en/node/342293
    <img width="400" src="https://musescore.org/sites/musescore.org/files/2023-01/Bildschirm%C2%ADfoto%202023-01-14%20um%2018.06.49.png">

- [x] Workspace dialog in Czech
    <img width="400" src="https://media.discordapp.net/attachments/818804809784229898/1055891439499886672/image.png">
- [ ] #15572

- [x] Barline inspector in Dutch
    <img width="300" alt="Scherm­afbeelding 2022-12-24 om 20 44 11" src="https://user-images.githubusercontent.com/48658420/209449269-eb6c89a3-5be0-4c8c-8686-61f3e9fa2c63.png">

- [x] The playback toolbar when the piece has more than 100 bars
    
    https://user-images.githubusercontent.com/48658420/195987842-f52c7500-ad77-449d-a95f-375f28f16240.mov

- [x] Preferences > "Reset Preferences" with German localisation
    <img width="463" alt="Scherm­afbeelding 2022-10-15 om 15 07 07" src="https://user-images.githubusercontent.com/48658420/195988022-9eee60cb-79fc-4dbf-9b32-ba7aa4164d23.png">
    (General problem: we must _never_ give a text button a fixed width; if we want the default width of the button to be bigger, we can use the `margins` or `minWidth` properties. But never set `width` directly.)

**Note:**
Please add any other text clipping problems that you find as comments to this issue!

DmitryArefiev:
@cbjeukendrup Dutch it's Nederlands in Preferences>Languages, right?

If yes, please also use 'Nederlands' in bug description

cbjeukendrup:
(But that one is not yet fixed by the PR)

DmitryArefiev:
[...]

PR #14139, opened 2022-10-28
Fix various text clipping issues

cbjeukendrup:
Resolves: #13914 
and some problems from #13773

### DIFF
--- a/src/appshell/qml/Preferences/internal/ProgrammeStartSection.qml
+++ b/src/appshell/qml/Preferences/internal/ProgrammeStartSection.qml
@@ -51,12 +51,14 @@ BaseSection {
         width: parent.width
 
         delegate: Row {
+            width: parent.width
             spacing: root.columnSpacing
 
             RoundedRadioButton {
                 anchors.verticalCenter: parent.verticalCenter
 
-                width: root.columnWidth
+                width: filePicker.visible ? Math.max(implicitWidth, root.columnWidth)
+                                          : parent.width
 
                 checked: modelData.checked
                 text: modelData.title
@@ -72,6 +74,8 @@ BaseSection {
             }
 
             FilePicker {
+                id: filePicker
+
                 pathFieldWidth: root.columnWidth
                 spacing: root.columnSpacing
 
@@ -102,6 +106,8 @@ BaseSection {
         height: contentHeight
 
         delegate: CheckBox {
+            width: parent.width
+
             text: modelData.title
             checked: modelData.visible
 
--- a/src/appshell/qml/Preferences/internal/PreferencesButtonsPanel.qml
+++ b/src/appshell/qml/Preferences/internal/PreferencesButtonsPanel.qml
@@ -57,7 +57,7 @@ Rectangle {
         FlatButton {
             anchors.left: parent.left
 
-            width: 160
+            minWidth: 160
 
             navigation.panel: root.navigation
             navigation.order: 2
@@ -76,7 +76,6 @@ Rectangle {
             navigation.panel: root.navigation
             navigation.order: 3
 
-            width: 132
             text: qsTrc("global", "Cancel")
 
             onClicked: {
@@ -92,7 +91,6 @@ Rectangle {
             navigation.panel: root.navigation
             navigation.order: 1
 
-            width: 132
             accentButton: true
 
             text: qsTrc("global", "OK")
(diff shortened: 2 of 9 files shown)


src/appshell/qml/Preferences/internal/ProgrammeStartSection.qml
```diff
@@ -51,12 +51,14 @@ BaseSection {
...
 
         delegate: Row {
+            width: parent.width
             spacing: root.columnSpacing
 
             RoundedRadioButton {
                 anchors.verticalCenter: parent.verticalCenter
 
-                width: root.columnWidth
+                width: filePicker.visible ? Math.max(implicitWidth, root.columnWidth)
```
Eism, line 60: I would like to hear the opinion of the designers on this issue.

In your solution, the problem is that the first column on this page may have a different width than first columns from other pages. May look ugly.

Eism:
@cbjeukendrup 

Also not sure about the dynamic calculation of button widths... and again I would like the designers to investigate 'Various text clipping issues' and only after that we would change the behaviour.

Summary:
I think we can merge a fix for https://github.com/musescore/MuseScore/issues/13914 (change is not so dangerous), https://github.com/musescore/MuseScore/issues/13773 should be fixed in a separate PR.

vpereverzev:
> @cbjeukendrup
> 
> Also not sure about the dynamic calculation of button widths... and again I would like the designers to investigate 'Various text clipping issues' and only after that we would change the behaviour.
> 
> Summary: I think we can merge a fix for #13914 (change is not so dangerous), #13773 should be fixed in a separate PR.

@cbjeukendrup can you isolate the changes in proposed way?

cbjeukendrup:
Yes, on it right now!

src/appshell/qml/Preferences/internal/ProgrammeStartSection.qml
cbjeukendrup: In principle, the first column _does_ have the same width as on other pages; only in the emergency case that the text in front of the filePicker box needs more space, we give it that space. It is not necessary in English, but we must be prepared for languages where it _is_ necessary. Also, since this page has rather different controls than other pages, I think it won't be a massive problem if this particular page doesn't exactly follow the same column layout in a few languages.

cbjeukendrup:
@Eism I now included a more conservative fix for https://github.com/musescore/MuseScore/issues/13773. We should simply not set a fixed width for a button, otherwise it's guaranteed that you get clipping in some languages. The solution is that we use the `minWidth` property instead, so that we follow the design in principle, but in the emergency case that a language requires more space, we allow the button to grow (instead of letting the text grow beyond the border of the button). @Tantacrul are you happy with that approach?

Eism:
@cbjeukendrup 
Not in the proposed way, but okay

Tantacrul:
> @Eism I now included a more conservative fix for #13773. We should simply not set a fixed width for a button, otherwise it's guaranteed that you get clipping in some languages. The solution is that we use the `minWidth` property instead, so that we follow the design in principle, but in the emergency case that a language requires more space, we allow the button to grow (instead of letting the text grow beyond the border of the button). @Tantacrul are you happy with that approach?
> 
> (This change should be safe since we are already using the `minWidth` property without problems. However, if you really want it excluded from this PR, I'll move it to a next PR.)

It's not easy for me to imagine the repercussions of the approach if it is applied universally. Will some buttons and other items get smaller than they are now? Will some buttons end up growing so much they extend into weird places, etc?

Tantacrul:
Is it possible to simply provide local solutions to each of the main offending areas. I see this being mentioned above, so I'm unsure if this is what you are doing or not.

cbjeukendrup:
Yes, this only applies to specific buttons, it's not a general change that affects everything.

Tantacrul:
OK, so it's tricky to test because I'm unsure what the problems are.

For example, I'm still seeing this (should obviously go to two lines):
image

This is fine though:
image


So, given that this is a partial fix... I think it's very low risk to just merge it.

Eism (approved):

Eism:
@Tantacrul, agree. I will merge it

cbjeukendrup:
> For example, I'm still seeing this (should obviously go to two lines):
> image

Yeah, that one is not yet solved by this PR. If we want it to go to two lines, then the dialog must be taller, which gives strange empty space in languages where it doesn't need two lines. 

Ideally, the height of the dialog would be calculated automatically, so that it always fits perfectly, but that doesn't work well in QML: the size is calculated _after_ the dialog has been opened and positioned, in other words, the dialog is centered on the screen before the correct height is known. The dialog ends up being not correctly centered. (That's why the standard warning/error/info dialogs are not centered correctly: for those, we do calculate the size dynamically. But those are not so tall, so the offset will be small enough to be not disruptive.)

Issue #13670, opened 2022-10-08
[MU4 Issue] Incorrect behaviour when "resetting shapes and positions" for various "lines"


cbjeukendrup:
**Describe the bug**
If you use the "Reset shapes and positions" action on various lines, unexpected behaviour occurs. It affects at least gradual tempo changes and "let ring" lines. The behaviour is the following:
- in debug builds, an assertion failure causes a crash
- in release builds (i.e. what you get from GitHub), the line width is set to zero.

The cause is some internal confusion about whether the line width is specified in `Spatium` or in `Millimetre` (that's a class in the code but I don't believe it really represents millimetres, but rather something like points or pixels).

**To Reproduce**
1. Select one of the mentioned lines
2. Format > Reset shapes and positions

**Expected behavior**
Should reset (among other things) the line width, in the same way as when you reset line width using the Inspector.

oktophonie:
(Sort-of-duplicate of #13500, and #13318 is related)

Issue #13500, opened 2022-09-26
[MU4 Issue] Resetting some types of text lines sets line thickness to 0


oktophonie:
**Describe the bug**
Some text lines, when reset (Ctrl+R, or Format > Reset shapes and positions), have their thickness set to 0. As well as making them invisibly thin, this also messes up the dash/gap size (which relates to the line thickness).

**To Reproduce**
Steps to reproduce the behaviour:
1. Create some text lines (palm mute, let ring, gradual tempo change all seem affected; ottavas seem immune. I have not checked every possible type).
2. Select them (individually, or together, with Ctrl+A)
3. Reset their shapes and positions (Ctrl+R)
4. Note that the thickness of some is now set to 0

**Expected behaviour**
Thickness should be reset to the default thickness for a line of that type.
If the default thickness is actually 0, it should rather match the standard thickness (that the palette items of each type of line have).

This video shows me resetting different types one by one; as you can see, the 8va is not affected:

https://user-images.githubusercontent.com/20604032/192330650-86bee3ef-4b82-4b0c-b274-c4e8f448f9fe.mp4

Note: #13318 is a not unrelated problem.

PR #14149, opened 2022-10-28
Fix line thickness bug on reset

mike-spa:
Resolves: #13500 
Resolves: #13670 

As far as I can understand, the problem is that `_lineWidth` is encoded in `Millimetre`, but the style setting from which it takes its value is expressed in `Spatium`. The issue (either an assert failure in debug mode or zero-width lines in release mode) arises because propertyvalue.h doesn't have a Spatium to Millimeter conversion. And that conversion can't be done by `PropertyValue` because, obviously, you can only convert Spatium to Millimeter if you know the global `spatium()` of the score. Hope this makes sense.

### DIFF
--- a/src/engraving/libmscore/line.cpp
+++ b/src/engraving/libmscore/line.cpp
@@ -1498,7 +1498,11 @@ bool SLine::setProperty(Pid id, const PropertyValue& v)
         _lineColor = v.value<mu::draw::Color>();
         break;
     case Pid::LINE_WIDTH:
-        _lineWidth = v.value<Millimetre>();
+        if (v.type() == P_TYPE::MILLIMETRE) {
+            _lineWidth = v.value<Millimetre>();
+        } else if (v.type() == P_TYPE::SPATIUM) {
+            _lineWidth = v.value<Spatium>().toMM(spatium());
+        }
         break;
     case Pid::LINE_STYLE:
         _lineStyle = v.value<LineType>();


oktophonie (approved):

src/engraving/libmscore/line.cpp
```diff
@@ -1498,7 +1498,11 @@ bool SLine::setProperty(Pid id, const PropertyValue& v)
         _lineColor = v.value<mu::draw::Color>();
         break;
     case Pid::LINE_WIDTH:
-        _lineWidth = v.value<Millimetre>();
```
RomanPudashkin, line 1501: There are similar conversions already in AbstractInspectorModel (valueToElementUnits and valueFromElementUnits). Maybe the bug is in one of them?

src/engraving/libmscore/line.cpp
```diff
@@ -1498,7 +1498,11 @@ bool SLine::setProperty(Pid id, const PropertyValue& v)
...
+        if (v.type() != P_TYPE::SPATIUM) {
```
RomanPudashkin, line 1501: It should be v.type() == P_TYPE::MILLIMETRE

src/engraving/libmscore/line.cpp
cbjeukendrup: No, this doesn't involve the Inspector. In fact, the Inspector works correctly; what's wrong is when you press Cmd+R to "reset" the element. That doesn't go through the Inspector.

src/engraving/libmscore/line.cpp
mike-spa: done, thanks 👍

RomanPudashkin (approved):

PR #14154, opened 2022-10-28
Inspector: Avoid using "notehead type" to mean different things

shoogle:
We already use "notehead type" to mean the shape of the notehead so we shouldn't use it to mean the duration as well. This affects the Note Inspector and the Ambitus Inspector.

## Ambitus

image

In the Properties panel, **"Notehead type"** becomes **"Notehead duration"** where appropriate.

Before | After
---|---
image | image

## Notes

image

In the Properties panel, **"Notehead type (visual only)"** becomes **"Override visual duration"** to indicate that this option changes the visual appearance of the notehead without affecting playback or horizontal spacing.

Before | After
---|---
 |

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/notation/notes/HeadSettings.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/notation/notes/HeadSettings.qml
@@ -141,7 +141,7 @@ FocusableItem {
 
                 NoteheadTypeSelector {
                     id: noteHeadTypeSection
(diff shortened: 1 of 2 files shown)


shoogle:
@Tantacrul, this is what was proposed on Discord. I found it affects the Ambitus Inspector as well as the Note Inspector.

Tantacrul:
Looks good

RomanPudashkin (approved):

PR #14183, opened 2022-10-30
Fix formatting links in Interactive standard dialogs

cbjeukendrup:
Before: 
Scherm­afbeelding 2022-10-30 om 12 14 59

After:
Scherm­afbeelding 2022-10-30 om 12 13 18

By default, text will be formatted automatically (interpreted as HTML or Markdown etc. as appropriate). This is the default for StyledTextLabel too, so for consistency we do this in Interactive too.

### DIFF
--- a/src/engraving/engravingerrors.h
+++ b/src/engraving/engravingerrors.h
@@ -83,7 +83,7 @@ inline Ret make_ret(Err err, const io::path_t& filePath = "")
                .arg(u"https://musescore.org");
         break;
     case Err::FileOld300Format:
-        text = mtrc("engraving", "This file was last saved with a development version of 3.0.");
+        text = mtrc("engraving", "This file was last saved in a development version of 3.0.");
         break;
     case Err::FileCorrupted:
         text = mtrc("engraving", "File \"%1\" is corrupted.").arg(filePath.toString());
--- a/src/framework/global/iinteractive.h
+++ b/src/framework/global/iinteractive.h
@@ -81,17 +81,19 @@ public:
     using ButtonDatas = std::vector<ButtonData>;
 
     enum class TextFormat {
-        PlainText = 0,
+        Auto = 0,
+        PlainText,
         RichText
     };
 
     struct Text {
         std::string text;
-        TextFormat format = TextFormat::PlainText;
+        TextFormat format = TextFormat::Auto;
+
         Text() = default;
         Text(const char* t)
-            : text(t), format(TextFormat::PlainText) {}
-        Text(const std::string& t, const TextFormat& f = TextFormat::PlainText)
+            : text(t), format(TextFormat::Auto) {}
+        Text(const std::string& t, const TextFormat& f = TextFormat::Auto)
             : text(t), format(f) {}
     };
 
--- a/src/framework/ui/view/interactiveprovider.cpp
+++ b/src/framework/ui/view/interactiveprovider.cpp
@@ -291,6 +291,7 @@ void InteractiveProvider::fillStandardDialogData(QmlLaunchData* data, const QStr
 {
     auto format = [](IInteractive::TextFormat f) {
         switch (f) {
+        case IInteractive::TextFormat::Auto:      return Qt::AutoText;
         case IInteractive::TextFormat::PlainText: return Qt::PlainText;
         case IInteractive::TextFormat::RichText:  return Qt::RichText;
         }


src/engraving/engravingerrors.h
```diff
@@ -83,7 +83,7 @@ inline Ret make_ret(Err err, const io::path_t& filePath = "")
                .arg(u"https://musescore.org");
         break;
     case Err::FileOld300Format:
-        text = mtrc("engraving", "This file was last saved with a development version of 3.0.");
+        text = mtrc("engraving", "This file was last saved in a development version of 3.0.");
```
Jojo-Schmitz, line 86: `using` might be better?

cbjeukendrup: I believe this was discussed before, and the decision was to use "in". I'm just making this one consistent with the rest.

Jojo-Schmitz:
I'd rather keep html stuff away from translatable texts
```
    case Err::FileTooOld:
        text = mtrc("engraving", "This file was last saved in a version older than 2.0.0. "
                                 "You can convert this score by opening and then "
                                 "saving in MuseScore version 2.x. "
                                 "Visit the "%1MuseScore download page%2 to obtain such a 2.x version.")
               .arg(u"<a href=\"https://musescore.org/download#older-versions"\">).arg(u"</a>");
        break;
    case Err::FileTooNew:
        text = mtrc("engraving", "This file was saved using a newer version of MuseScore. "
                                 "Visit the %1MuseScore website%2 to obtain the latest version.")
               .arg(u"<a href=\"https://musescore.org"\">).arg(u"</a>");
        break;
```

cbjeukendrup:
@Jojo-Schmitz That was my first iteration too, but it turned out that it is better to keep the html in the string. That gives these nice things on Transifex:
Scherm­afbeelding 2022-10-30 om 12 43 10
Which is better than strange %1 and %2 and then having to add explanations about what those mean.

Jojo-Schmitz:
It may though open the door for malicous miss-translations

cbjeukendrup:
@Jojo-Schmitz But how is this more vulnerable than with %1...%2? The user can't see nor modify the contents of the html tags; only sees those shapes created by Transifex. 
The only thing a user could do is _adding_ malicious html, but that can happen equally with %1...%2. 
Or am I missing something?

Jojo-Schmitz:
True, with html code it is just more obvious.

RomanPudashkin (approved):

bkunda:
@cbjeukendrup can you point me towards other dialogs that contain interactive links, so I can test this?
(I admit I wasn't aware we had any like this; we usually seem to prefer buttons to html links).

cbjeukendrup:
The quickest on I can think of it the dialog that appears when opening a score that has a too old file format. That is also the dialog in the picture in the PR description. Here is one: [bach-bc2.mscx.zip](https://github.com/musescore/MuseScore/files/9901802/bach-bc2.mscx.zip)

bkunda:
> The quickest one I can think of is the dialog that appears when opening a score that has a too old file format. That is also the dialog in the picture in the PR description. Here is one: [bach-bc2.mscx.zip](https://github.com/musescore/MuseScore/files/9901802/bach-bc2.mscx.zip)

Thanks, I'll check this. 
Is there a way of finding out if there are any others?

cbjeukendrup:
@bkunda Actually the dialog shown in the PR description is the only one that can be tested. The only other dialog that contains a link is the "file too new" dialog, but we currently don't have any files that are newer than MuseScore 4 of course. 

The "You're up-to-date" dialog does contain a simple link to musescore.org, but that one was already working correctly.

For the rest, the About dialogs contain some links, but those go through different logic than these simple error message dialogs.

bkunda:
So this PR really is to fix the URL formatting of this one dialog then?

At the risk of opening a can of worms – and distracting from the issue at hand – I have to say that this dialog has rather confused me and I'd like some clarity (perhaps from @Tantacrul and/or @oktophonie) about whether we even need it at all.

English version of the dialog above:
 
Screenshot 2022-10-31 at 3 17 11 pm

Firstly, I was able to successfully open the very old MuseScore file (from v. 0.9.6) by pressing the **Ignore** button, which seems strange enough (Using an "ignore" action to open something). It also means that there wasn't ever really an **Open error** in the first place. 

Also, given that the **Ignore** button then triggers this dialog, I'm not sure why we need to worry the user with the interim "Open error" step:

Screenshot 2022-10-31 at 3 18 13 pm

Unless there's something very important I'm missing – some particular old versions perhaps that just won't ever be openable unless the file is opened in the correct (old) version?

cbjeukendrup:
> So this PR really is to fix the URL formatting of this one dialog then?

In practice, yes, although this will also fix URL formatting in potential future dialogs of this kind. 

About whether the "File too old" error is really necessary, I can't say anything sensible about this. It also existed in MS3, and I assume someone created it for a reason, but I've no idea which reason. Maybe it was laziness because nobody wanted to fix the compatibility score reading code properly? 
@MarcSabatella @jeetee @Jojo-Schmitz Do you by chance know the history of this?

Tantacrul:
I have never seen this dialog before, lol!

Can't imagine it's worth touching now, since so few scores will actually use it.

bkunda:
Ok well in light of this discussion, I'm happy to say that this PR at least does what it sets out to do (I.e. the URL _is_ formatted correctly and works as a clickable link after having tested on macOS).

🟢 So, FIXED @Eism @RomanPudashkin.

And we can deal with this dialog later.

jeetee:
For the history part, the 1.x score import code for 3.x wasn't really well tested and a blunt port of 2.x. As a result of it, the result of directly importing a 1.x score lost more markup if you opened it directly in 3.0 rather than going into 2.x first and then importing that (through the better read20x.. logic).

I don't recall which things were lost in the read1.. logic, but it was enough of a test report issue to introduce that warning.

PR #14186, opened 2022-10-30
Fix rounded corners for SplashScreen on macOS

cbjeukendrup:
Before: if you look closely, you see that there is a sort of background behind the rounded corners. 
Scherm­afbeelding 2022-10-30 om 14 25 44

After: 
Scherm­afbeelding 2022-10-30 om 14 22 01

### DIFF
--- a/src/appshell/view/internal/splashscreen.cpp
+++ b/src/appshell/view/internal/splashscreen.cpp
@@ -34,19 +34,26 @@ using namespace mu::appshell;
 
 static const QString imagePath(":/resources/LoadingScreen.svg");
 
-static const QSize splashScreenSize(810, 405);
+static constexpr QSize splashScreenSize(810, 405);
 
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

bkunda:
Tested, seems to be fine on MacOS. 
FWIW though, I couldn't actually reproduce the original problem in the latest nightly, so I haven't been able to do a "real" test where I can report a change/fix.

DmitryArefiev:
@cbjeukendrup I'm also don't see a difference between master and PR's build on Mac (don't see the bug in both cases)

But let's merge if it's fixed on your side

DmitryArefiev:
@cbjeukendrup Oh.. Nope, I tried on macOS12.6.  I'll install OS13 later to see how MU looks. Thanks!

PR #14204, opened 2022-10-31
implemented dead slapped element

alexpavlov96:
Implemented dead slapped element for importing from guitar pro:
Screenshot 2022-10-31 at 10 36 17

[dead-slapped.gp.zip](https://github.com/musescore/MuseScore/files/9899416/dead-slapped.gp.zip)

### DIFF
--- a/src/engraving/libmscore/deadslapped.cpp
+++ b/src/engraving/libmscore/deadslapped.cpp
@@ -1,0 +1,96 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "deadslapped.h"
+
+#include "rest.h"
+#include "staff.h"
+#include "log.h"
+
+using namespace mu;
+using namespace mu::engraving;
+
+namespace mu::engraving {
+//---------------------------------------------------------
+//    DeadSlapped
+//--------------------------------------------------------
+
+DeadSlapped::DeadSlapped(Rest* rest)
+    : EngravingItem(ElementType::DEAD_SLAPPED, rest)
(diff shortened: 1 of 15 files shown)


handrok (approved):

src/engraving/libmscore/deadslapped.cpp
```diff
@@ -0,0 +1,96 @@
...
+//---------------------------------------------------------
+//   fillPath
+//---------------------------------------------------------
+
+void DeadSlapped::fillPath()
+{
+    constexpr double crossThinknessPercentage = 0.1;
+    double height = m_rect.height();
+    double width = m_rect.width();
+    double crossThickness = width * crossThinknessPercentage;
```
mike-spa, line 72: Tiny detail: I would not express the thickness in terms of the width, cause one may want to adjust the two values separately. I think it would be better to express the thickness with its own value, written in terms of spatium()

src/engraving/libmscore/deadslapped.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include "durationelement.h"
+
+namespace mu::engraving {
+//---------------------------------------------------------
+//    @@ DeadSlapped
+///     This class implements a dead slapped element.
+//---------------------------------------------------------
+
+class DeadSlapped : public EngravingItem
```
mike-spa, line 34: General comment, @cbjeukendrup please correct me if I'm wrong here. Since DeadSlapped is, essentially, a rest, we could consider deriving this class from Rest or ChordRest. At the moment, if I understand correctly, DeadSlapped is defined as just a graphical element, and you are drawing it in place of the rest when you need it. That seems a bit hacky to me. Deriving it from Rest would let this class inherit lots of stuff that may turn out useful when we build proper support for this element in Musescore. Doesn't need to be done now, anyway, we can also do it later.

src/engraving/libmscore/rest.cpp
```diff
@@ -334,6 +343,13 @@ void Rest::layout()
     for (EngravingItem* e : el()) {
         e->layout();
     }
+
+    _skipDraw = false;
+    if (_deadSlapped) {
+        _skipDraw = true;
+        return;
+    }
```
mike-spa, line 351: Yeah, this seems a bit hacky to me (see also my previous comment). If I understand correctly, we are just drawing the dead slapped instead of the rest. I think it can do for now, but I'd consider switching to a more solid solution in future

mike-spa (approved):

src/engraving/libmscore/deadslapped.cpp
```diff
@@ -0,0 +1,96 @@
...
+DeadSlapped::DeadSlapped(Rest* rest)
+    : EngravingItem(ElementType::DEAD_SLAPPED, rest)
+{
+}
+
+void DeadSlapped::draw(mu::draw::Painter* painter) const
+{
+    TRACE_OBJ_DRAW;
+    painter->setPen(draw::Pen(curColor()));
+    painter->setBrush(draw::BrushStyle::SolidPattern);
```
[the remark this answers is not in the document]
alexpavlov96: thanks, i'll change in next PR

abariska:
🟢 **Tested**

PR #14211, opened 2022-10-31
fixed track calculation for chord rest linked elements

alexpavlov96:
before linked elements of chord/rest type had a track of main staff, not linked

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1393,9 +1393,6 @@ void Chord::processSiblings(std::function<void(EngravingItem*)> func) const
     for (Note* note : _notes) {
         func(note);
     }
-    for (EngravingItem* e : el()) {
-        func(e);
-    }
     for (Chord* chord : _graceNotes) {    // process grace notes last, needed for correct shape calculation
         func(chord);
     }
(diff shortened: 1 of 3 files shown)


handrok (approved):

abariska:
🟢 **Tested**

src/engraving/libmscore/excerpt.cpp
```diff
@@ -1229,6 +1229,12 @@ void Excerpt::cloneStaff(Staff* srcStaff, Staff* dstStaff, bool cloneSpanners)
                             score->undoAddElement(ne1);
                         }
                     }
+
+                    ChordRest* neCr = toChordRest(ne);
+                    for (EngravingItem* crItem : neCr->el()) {
+                        crItem->setTrack(dstTrack);
```
cbjeukendrup, line 1235: I think a different fix is preferable: for ChordRest and Chord, this happens inside their `setTrack` method. See `ChordRest::setTrack` and `Chord::setTrack`. So it is probably better to update those methods (or maybe the `processSiblings` method of those classes instead)

alexpavlov96: @cbjeukendrup I added proccessing of _el in ChordRest, not sure if that's expected, but it works for the task as well

mike-spa (approved):

cbjeukendrup (approved):

Issue #13426, opened 2022-09-20
[MU4 Issue] Lyrics: Underlines are not beeing considered by the anti-collision automation


oMrSmith:
![image](https://user-images.githubusercontent.com/78870770/191314326-1d923096-d70e-4451-8826-b2580f4e4cd5.png)

MarcSabatella:
FWIW, this is something we'd have loved to get working in MuseScore 3, but we just couldn't figure out a good way, since things were being done in the wrong order to make it feasible.  See https://musescore.org/en/node/280139 for some discussion.  No doubt it's possible, but it's not just a simple oversight that I would expect to be easily fixed, unless more has changed about layout than I think is actually the case.  But if you can pull it off @asattely , that would be pretty awesome!

oktophonie:
Best leave it until 4.X then.

PR #14214, opened 2022-10-31
Fixing (part of) lyrics lines collision bugs

mike-spa:
Resolves: #13426 

Does anybody know any specific reason why `LyricsLine` was _not_ added to the skyline? I'm a bit confused as to why it wasn't done before, but honestly I can't find any reason why we shouldn't do it.

Let me also clarify one thing. Avoiding lyrics-line collisions with items _of the same staff_ is one problem. In thas case, it is the lyrics that should move down, and it is currently complicated because of alignment issues. That is _not_ addressed in this PR (but I am going to fix it in future). Avoiding lyrics-line collisions with items _of the staff below_ is a **_completely separate_** problem, because in that case it is the staff below that should move down, lyrics don't need to do anything, except add the line to the skyline (at least as far as I understand). 
So here's before:
lyricsLineBefore
and after:
lyricsLineAfter

NOTE: I've also made a change to the triggerLayout() method because the lyrics extender may go into the next system(s) so we need to also trigger the layout for those (which is done by `Spanner::triggerLayout()`).


MarcSabatella:
I know I tried working on lyrics line autoplace issues this at one point and gave up because I couldn't find a good solution.  But to be honest, I don't think it occurred to me to just solve that part of the problem - the collisions between staves / systems - by adding the lyrics line to the skyline, without addressing what I perceived to be the more important problem (getting things working within a single staff).  But, no doubt, what you are doing here does seem helpful and quite reasonable.

See https://musescore.org/en/node/280139 if you haven't already for some of the history here.  Note in particular the discussion brought up by @shoogle (https://musescore.org/en/node/280139#comment-935536) regarding "exclusion zones" that might reasonably be expected in some cases even when no lyrics lines are present - but also to the observations about why this isn't always desired.

MarcSabatella:
As for why the lyrics lines weren't added to the skyline in the first place, I don't know.  But, I can speculate - these facilities were developed independently at around the same time.  Lyrics lines existed long before autoplace, but, they were implemented a couple of times along the way, and in a pretty awkward way, and they probably got missed when the skyline stuff was added.

oMrSmith:
<3

PR #14260, opened 2022-11-02
fix #11362: Workspace Name Character Limit

### LINKED ISSUE #11362: [MU4 Issue] [Workspaces] Workspace Name Character Limit
**Describe the bug**
There is an ambiguous error when typing reeeeeeeeaaaaaallly long Workspace names. 
It could be better to give a fixed character limit prompt to the user instead of "cannot be used".

**To Reproduce**
1. Go to 'Workspaces'
2. Click on 'Create new workspace'
3. Create a really, really long name
4. See issue

Note: trying to test the limits of this, i did ignore the warning and was able to get it to accept it. Surprisingly, nothing broke with such a long name. 

**Expected behavior**
A warning that tells that workspace name should be less than X characters.

**Screenshots**
Screen Shot 2022-04-25 at 6 54 19 PM
Screen Shot 2022-04-25 at 6 55 18 PM
Screen Shot 2022-04-25 at 6 55 34 PM

**Platform information**
OS: macOS 12.2, Arch.: x86_64
MuseScore version (64-bit): 4.0.0-2214419458
revision: github-musescore-musescore-df96a36

poypoyan:
Resolves: #11362

Just added an extra condition in newworkspacemodel.cpp. Need review for the error message.

### DIFF
--- a/src/workspace/view/newworkspacemodel.cpp
+++ b/src/workspace/view/newworkspacemodel.cpp
@@ -106,6 +106,11 @@ void NewWorkspaceModel::validateWorkspaceName()
                          .arg(m_workspaceName);
         return;
     }
+
+    if (m_workspaceName.size() > 24) {
+        m_errorMessage = qtrc("workspace", "Workspace name should not have more than 24 characters. Please choose a different name.");
+        return;
+    }
 }
 
 void NewWorkspaceModel::setUseUiPreferences(bool needUse)


RomanPudashkin (approved):

src/workspace/view/newworkspacemodel.cpp
```diff
@@ -106,6 +106,11 @@ void NewWorkspaceModel::validateWorkspaceName()
                          .arg(m_workspaceName);
         return;
     }
+
+    if (m_workspaceName.size() > 24) {
```
Eism, line 110: As @Tantacrul said https://github.com/musescore/MuseScore/issues/11362#issuecomment-1108761914
> I suggest a cap of 24 characters. A user should not be able to enter more than that in the text field.

We should just ignore character input without showing an error message.

bkunda:
Thanks @poypoyan for working on this!

My comments on this are:

- We shouldn't show an error message when the text input field is empty. Just keep the "Select" button disabled.
- We should allow an empty space after a character (and again, not show the error message "xxx cannot be used as a workspace name"). It's entirely plausible that a user might accidentally add a space after their workspace name, and this error message doesn't let them know _why_ their name can't be used. If it's _absolutely impossible_ for technical reasons to allow a space at the end of a workspace name, then we at least need a more helpful error message (E.g. "Please delete the space at the end of your workspace name"). But preferably, the space should be allowed and there should be no need for this error message.
- We should allow more than 24 characters. I would suggest perhaps 48 instead. 

This demo video is rather long (about 4:00), but you can see how confused I got 😅
https://www.dropbox.com/s/b71cpm6dy9wh4qq/workspace-error-message.mov?dl=0

Thank you so much again!

Tantacrul:
One comment @bkunda - I was suggesting a lower character limit so that the UI doesn't end up really wide:
image

bkunda:
> One comment @bkunda - I was suggesting a lower character limit so that the UI doesn't end up really wide: image

I don't disagree, but I think that 24 characters is a little _too_ constrictive.

Even 40 characters feels a little more relaxed: 

Screenshot 2022-11-03 at 10 40 46 am

Screenshot 2022-11-03 at 10 43 47 am

Tantacrul:
We haven't met in the middle here so I now look weak in the eyes of the community if I give way to you. The only way forward is to double down, so I'm insisting on a character limit of 10.

(OK, 40's fine)

RomanPudashkin:
fixed in https://github.com/musescore/MuseScore/pull/14288

PR #14264, opened 2022-11-02
gp-import: not importing grace note without main chord

alexpavlov96:
fixed problem with importing grace note in the end of measure, 
not adding it cause chord cannot have only grace notes in MU
Screenshot 2022-11-02 at 09 44 30


handrok (approved):

abariska:
🟢 **TESTED**
Please, merge

PR #14272, opened 2022-11-02
replaced unnecessary styles with guitar pro configuration

alexpavlov96:
removed styles used for import guitar pro and replaced them with experimental configuration

### DIFF
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -558,7 +558,6 @@ Note::Note(Chord* ch)
     _playEvents.push_back(NoteEvent());      // add default play event
     _cachedNoteheadSym = SymId::noSym;
     _cachedSymNull = SymId::noSym;
-    _fretConflictResolveSupported = score()->styleB(Sid::fretConflictResolveSupported);
 }
 
 Note::~Note()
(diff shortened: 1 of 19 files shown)


handrok (approved):

src/engraving/libmscore/note.cpp
```diff
@@ -558,7 +558,7 @@ Note::Note(Chord* ch)
     _playEvents.push_back(NoteEvent());      // add default play event
     _cachedNoteheadSym = SymId::noSym;
     _cachedSymNull = SymId::noSym;
-    _fretConflictResolveSupported = score()->styleB(Sid::fretConflictResolveSupported);
+    _fretConflictResolveSupported = !engravingConfiguration()->guitarProImportExperimental();
```
cbjeukendrup, line 561: I'm thinking that `_fretConflictResolveSupported` should maybe not be a stored member. I would prefer if `fretConflictResolveSupported` was defined like this:
```c++
bool Note::fretConflictResolveSupported() const
{
    return !engravingConfiguration()->guitarProImportExperimental();
}
```

abariska:
🟢 **TESTED**
Please, merge to master

cbjeukendrup (approved):
I'm happy with this now, but _only_ if @RomanPudashkin is too :)

Personally, I think this is the best solution for now. It is clear that we should not be abusing Sid and StyleDef for this. It seems unrealistic to me that all of the experimental features will be implemented fully right now before the release, so we need an interim solution, namely using the `I…Configuration` classes. Having one "blanket" method in EngravingConfiguration makes the code in the EngravingModule hard to understand, so therefore I had suggested to Alex to introduce multiple methods with descriptive names. This makes it easy to finish the experimental features / workarounds one by one, because it is clear which code belongs to which feature/workaround.

RomanPudashkin (approved):

PR #14297, opened 2022-11-03
Fix #14160 Tie direction fix for <showBackTied>-enabled tablature staves

### LINKED ISSUE #14160: [MU4 Issue] Ties in tablature are positioned the wrong way round
**Describe the bug**

It appears that MS4 is setting the position (above/below) of ties in tablature wrongly – at least in imported MS3 files. e.g.

MS3
ms3_slur

MS4
ms4_slur

**To Reproduce**

1. Open the attached MS3 file in both MS3 and MS4. 
[ties_wrong_way.mscz.zip](https://github.com/musescore/MuseScore/files/9893561/ties_wrong_way.mscz.zip)

2. Compare and contrast the positions of ties in mm 1,2,3, and 4.

RESULT: The ties which were correctly oriented upwards in MS3, now point downwards.

**Expected behavior**

Ties should be correctly oriented by default, as they were in MS3 (and should respect positions set in MS3).

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223020418, revision: github-musescore-musescore-fa12b58

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/14160

There is a property that save files can have called `<showBackTied>`, which makes ties visible even on simple (i.e. stemless) tab staves. The tie direction calculation did not take these staves into account (that is, it tries to put the tie on the "stem side" but there is no stem). Now it does!

### DIFF
--- a/src/engraving/libmscore/tie.cpp
+++ b/src/engraving/libmscore/tie.cpp
@@ -560,9 +560,9 @@ void TieSegment::adjustY(const PointF& p1, const PointF& p2)
         // Each line position in the staff has a set endpoint Y location
         double newAnchor;
         if (isUp) {
-            newAnchor = floor(line / 2.0) + ((line & 1) ? staffLineOffset : -staffLineOffset);
+            newAnchor = floor(line / 2.0) + (line & 1 ? staffLineOffset : -staffLineOffset);
         } else {
-            newAnchor = floor((line + 1) / 2.0) + ((line & 1) ? -staffLineOffset : staffLineOffset);
+            newAnchor = floor((line + 1) / 2.0) + (line & 1 ? -staffLineOffset : staffLineOffset);
         }
 
         // are the endpoints within the staff?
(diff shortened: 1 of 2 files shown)


oktophonie (approved):

src/engraving/libmscore/tie.cpp
```diff
@@ -1141,11 +1143,16 @@ TieSegment* Tie::layoutFor(System* system)
...
-                // in polyphonic passage, ties go on the stem side
-                _up = c1->up();
+            bool simpleException = st && st->isSimpleTabStaff();
+            if (st && st->isSimpleTabStaff()) {
+                _up = !(c1->voice() & 1);
             } else {
-                _up = !c1->up();
+                if (c1->measure()->hasVoices(c1->staffIdx(), c1->tick(), c1->actualTicks())) {
+                    // in polyphonic passage, ties go on the stem side
+                    _up = simpleException ? !(c1->voice() & 1) : c1->up();
```
RomanPudashkin, line 1152: I have seen this code in many places, it is better to move it to some function with a clear name

`voice() & 1`

For example, you can add such functions to mscore.h:

```
inline constexpr bool isOddVoice(voice_idx_t voiceIdx) 
{
    return voiceIdx & 1;
}

inline constexpr bool isEvenVoice(voice_idx_t voiceIdx) 
{
    return !(voiceIdx & 1);
}
```

And use them here

asattely: Done! I've only replaced the ones in tie.cpp. I've used the names `isUpVoice` and `isDownVoice` instead, because of the the code is 0-indexed but the display in-app is 1-indexed, which could create some unnecessary problems.

RomanPudashkin (approved):

Issue #10435, opened 2022-02-03
[MU4 Issue] Fretboard diagram chord symbol is doubled in edit mode


DmitryArefiev:
https://user-images.githubusercontent.com/90187801/152333863-add047ae-7675-4f48-bffb-0122f95578f5.mp4

**To Reproduce**
1. Add a fretboard diagram to the note from Palettes
2. Edit the chord symbol

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MU3**

Tantacrul:
Confirmed on todays' build.
<img width="193" alt="image" src="https://user-images.githubusercontent.com/21022311/171036956-e580ee88-766c-4f81-a009-3c1487e63598.png">

DmitryArefiev:
Fixed in #14312

PR #14312, opened 2022-11-04
Fix double drawing of chord symbol in Edit Mode, and related issues

cbjeukendrup:
Problem with the old situation: TextBase aligns text relative to its pos(), while Harmony laid out alignment-agnostic and compensated by setting pos(). But in Edit Mode, Harmony uses TextBase's methods, and otherwise it uses its own method, so both methods still need to draw in the same place. In MS3, this seems to be accomplished by changing pos() when Edit Mode is entered or left. But nowadays, the full layout method is not called anymore when entering EditMode, so pos() will not be set correctly.

We fix this by making the alignment behaviour of Harmony similar to TextBase's. Both now align text relative to their pos(), so that pos() can now be the same in Edit Mode and otherwise. This approach allows for some small de-jankifications.

Finally, this tweaks the alignment relative to FretDiagrams; the visual center of the diagram was calculated incorrectly, and right alignment was too far to the right.

Resolves: #10435


cbjeukendrup:
I've taken one commit from #14086, that helps fixing the VTest issues that were caused by this. So it would be great if that PR could be merged soon, to unblock this one.

RomanPudashkin (approved):

DmitryArefiev:
Tested #10435 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

MarcSabatella:
Is the merging of this PR why my own PR builds are showing vtest failures relating to chord symbols, even though my code has nothing to do with them?

Issue #13687, opened 2022-10-10
[MU4 Issue] Tied note does not sound when entered


rgreen5:
**Describe the bug**

In MS3, entering a tied note in note input mode caused the note to sound. In MS4 tied notes are entered without any sound.

**Platform information**

 OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-3216906122, revision: github-musescore-musescore-eb51740

MarcSabatella:
@Tantacrul - this is an important accessibility issue; without this feedback it's hard for a blind user to know if they entered a note.  Made especially important since "T" is a new shortcut they won't be trusting yet (and realistically, if they previously were testing the alpha, it may or may not work).

Tantacrul:
Thanks @MarcSabatella

MarcSabatella:
See https://github.com/musescore/MuseScore/pull/14347, which fixes this

Issue #12935, opened 2022-08-19
[MU4 Issue] "R" command to add a repeated note does not also play the note


MarcSabatella:
**Describe the bug**
When in Note Input mode, pressing 'R' repeats the last entered note. However, you do not hear the new note (we did in MS3). This is a problem for blind users.

**To Reproduce**
1. New score
2. First measure, note input
3. Type C (enters a C and you hear it)
4. Press R (adds a new 'C' - but you don't hear it)

Tantacrul:
I've altered the description and explanation to just make it a little clearer.

MarcSabatella:
@Tantacrul - I meant to list this one last week as one of the accessibility issues worth addressing.

I *think* it's a one-line fix, the same as https://github.com/musescore/MuseScore/issues/13687 - just a matter of adding a call to setPlayNote() to the right place.  I can probably find the right spots with some trial and error (or if someone wants to advise me).  I will see if I can pull that off; would be nice to contribute something concrete to the code for 4.0

MarcSabatella:
See https://github.com/musescore/MuseScore/pull/14347, which fixes this

PR #14347, opened 2022-11-05
Play note

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/13687
Resolves: https://github.com/musescore/MuseScore/issues/12935

Added calls to playSelectedElement when executing the tie or repeat selection commands in note input mode.

### DIFF
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -1090,6 +1090,7 @@ void NotationActionController::addTie()
 
     if (noteInput->isNoteInputMode()) {
         noteInput->addTie();
+        playSelectedElement(true);
     } else {
         interaction->addTieToSelection();
     }
@@ -1110,6 +1111,7 @@ void NotationActionController::chordTie()
 
     if (noteInput->isNoteInputMode()) {
         noteInput->addTie();
+        playSelectedElement(true);
     } else {
         interaction->addTiedNoteToChord();
     }
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -1036,6 +1036,15 @@ void NotationActionController::repeatSelection()
     if (!ret && !ret.text().empty()) {
         interactive()->error("", ret.text());
     }
+
+    auto noteInput = interaction->noteInput();
+    if (!noteInput) {
+        return;
+    }
+
+    if (noteInput->isNoteInputMode()) {
+        playSelectedElement(true);
+    }
 }
 
 void NotationActionController::pasteSelection(PastingType type)


Tantacrul:
On Mac, neither the 'T' or 'R' shortcut result in sound for MS Basic, VSTi or Muse Sounds.

Does it work on Win/Lin Marc?

MarcSabatella:
Works on Linux.  You are in note mode?  For me, typing C T or C R with this PR works.  Will test Windows later.

Tantacrul:
Ah, yes, it works in Note Input mode.

However, the same functions should really trigger a sound when you are in select mode too. Give it a try and tell me if you agree.

MarcSabatella:
To me that's debatable, but if it were to be implemented, it would need to be rather more selective.

For ties, the reason to need this in note input mode is that a new note *is* being added.  In normal mode, it's not a new note being added - you're just connecting two already-existing notes.  Adding a marking to already existing note generally doesn't play the note - doesn't work that way for articulations, dynamics, etc.  So it would seem inconsistent to do this only for ties.  But not difficult to implement if we are looking for new behavior as opposed to just fixing the regression.

For repeat selection, new notes are indeed added, but note this command works on selections of any size, not just single notes.  It clearly shouldn't play anything if you're repeating an entire 8-measure passage, any more than copy/paste does.  But for a single note, it certainly could.  And then the question would be, should copy/paste of a single note *also* play?

Anyhow, MU3 didn't do any of that, it just did the playback in note input mode.  And I'm pretty sure that's all blind users expect as well, and all I'd consider essential to be able to produce reasonable tutorials that don't seem broken.  Anything more complex would probably require someone else to work on a design and then to do the implementation, as it would go beyond the one-line fixes I am doing here.

Tantacrul:
I think we could safely say that R should produce a new sound, regardless of the number of instruments. If that isn't possible, then for just one instrument (since that's likely to cover the vast majority of use cases). The rest - like copy/paste are all debatable and certainly not issues that need block the 'R' example.

T is different, as you say, so I think you're right in that instance.

MarcSabatella:
Hmm, I wasn't talking about multiple *instruments* at all - I was talking about multiple *measures*, since that's actually the most common use case for the "R" command outside of note input mode (eg, to make a second copy of a whole 8-bar phrase).  And here we clearly shouldn't try to initial playback of an entire phrase.

Having "R" play the whole vertical slice if you've selected a single chord but across multiple instruments - not sure that's possible given the current code.  It doesn't seem set up to play a single note across multiple staves at all - only a single note or a single chord on a single staff.  Notice you also don't get audible feedback if you select a single note across multiple staves and hit Up or Down to transpose, etc.  And trying to make that happen would open another can of worms too, because if you select a single quarter note on one staff then extend the selection downward, it could well encompass two eighth notes on the staff below, and then what should be played?

So, I don't think it's really feasible to try to play a whole range selection, whether it's multiple measures or a single note across multiple staves.  But on the other hand, with a single *note* is selected, getting "R" to play it makes perfect sense and should be simple enough, so I'll make that change.

MarcSabatella:
The change I made "should" work to have it play in normal mode, but it doens't, and I think i know why.  It's because upon pressing "R", you're back to having a range selection instead of a single note, and the playSelectedElement command doens't handle ranges, as noted.

So no doubt possible to address this, but it's a bigger job.  The good news, the code should still be good as is to work in note input mdoe only, and if someone else wants to address making range selections in general play somehow, the "R" command will start working.

Tantacrul:
Ah, for muliple measures, I think we don't have a choice. If I select more than one note (horizontally) and press 'R', there's nothing we can do then. No point playing anything back.

Tantacrul:
So, if that's all @MarcSabatella, then I can confirm that:

- R shortcut is now producing a sound when in Note Input mode
- T shortcut is also now producing a sound when in Note Input mode

If that's all, then @Eism @RomanPudashkin - if you're happy with this, can it be merged?

MarcSabatella:
This is all from my perspective.  I do agree it would have been nice to get "R" to produce a sound with a single note selected in normal mode, but it turns out that's a bigger job with bigger ramifications.  Definitely worth tackling someday, but no reason this change here need wait for that.

RomanPudashkin (approved):

PR #14365, opened 2022-11-06
Fix asking save path for mscx folders on macOS and Linux

cbjeukendrup:
Resolves: #11615

@DmitryArefiev I couldn't test on Linux, but something similar appeared to happen on macOS, depending on one of the Preferences in Finder. So now it works well on macOS, and hopefully on Linux too. 

As a solution/workaround, on macOS and Linux, we ask for a .mscx path instead. Then, we remove the mscx extension and create a folder at that path, and put the mscx file inside that folder. 

However, this means that we will write to a different place than the file dialog thinks. As a result, the file dialog might fail to display the "do you want to replace" question correctly:
- if a mscx file with the same name exists in the same parent folder as where we will be creating our folder, the file dialog will incorrectly display a question, because it thinks we will override the mscx file (but we aren't).
- if a folder with the same exists in the same parent folder as where we will be creating our folder, the file dialog will _not_ display a question, because it doesn't know we will override that folder.

We cannot solve the first point (apart from by using forbidden private macOS APIs), but to address the second point, we display a question dialog ourselves. 

So, in summary, for testing, the following things are important to verify:
- when saving as mscx, a folder will be created at the specified path, and a mscx file (plus other files) will be inside that folder
- our custom "do you want to replace..." dialog appears in cases where the file dialog fails to display it. And sometimes, the file dialog does display such a question while it shouldn't; we will have to live with this. But I think this is okay, since "mscx folders" are (permanently) marked as an experimental feature anyway.


Eism:
@cbjeukendrup 
I think the problem is how we manage the QFileDialog/qml FileDialog.

If I understand the problem correctly (when changing the filter, the extension is not removed and the user always saves mscz), then we can do the following:
QFileDialog has a filterSelected signal. Depending on the selected filter, we should change the extension. It may also be necessary to change the fileMode - if user selected mscx, then it should be QFileDialog::Directory(but maybe removing the extension is enough).

DmitryArefiev:
@cbjeukendrup Should I test this now or do you need to make changes?

cbjeukendrup:
@DmitryArefiev First I'll need to discuss it with Elnur...

@Eism indeed, when the user selects a filter option, normally, the file dialog updates the extension accordingly. But in the case of mscx folders, there is no extension; the file pattern would be `*`. But beside "no extension", this pattern also matches "any extension", including .mscz which would be currently selected. So the file dialog is happy and changes nothing, leaving mscz as the extension. 

(And on macOS, something extra mysterious is going on; by default, the extension is not visible in the filename field in the dialog, and the behaviour when we or the user would add an extension manually is a bit unpredictable. There is a Finder preference that influences this too: Finder > Preferences > Advanced > Show all file name extensions. So this is yet another complication.)

I tried your suggestion of removing the extension manually when the filterSelected signal is emitted, but is doesn't seem to work, at least not on macOS. Setting file mode to Directory doesn't help much either: in that case, the "do you want to replace" dialog will not appear at all. (Apparently, the combination of "saving" and "directory" doesn't really work; probably only "open file" / "save file" / "open directory" are working).

Also, I see some architectural challenges: this strategy would require very close interaction with QFileDialog, but that is buried deep inside `Interactive(Provider)`. I'm not sure how we should interact with it in SaveProjectScenario. 

So I see two remaining options:
1. This PR as it is right now
2. A variation on this PR, namely instead of using a fake mscx extension, checking `QFileDialog::selectedNameFilter` to see whether the user was saving as mscz or mscx. This would also require changes to the `Interactive(Provider)` infrastructures, but that does seem doable.

DmitryArefiev:
Tried on LinuxMint - #11615 **still occurs**

cbjeukendrup:
@DmitryArefiev That's a pity... and quite surprising to me. Could you maybe send me a screen recording, so that it might give me a clue of where it goes wrong? 

The reason that it surprises me so much, is that this PR creates quite a "normal" situation for the file dialog: just two filter options, one with `.mscz` and one with `.mscx`. All our strange logic about creating folders instead of files is basically kept away from the file dialog. So if even this doesn't work, I'm starting to think that the problem might be caused by the Mint file dialog itself, rather than by us... Anyway, a recording might hopefully be helpful :)

Eism:
@cbjeukendrup 
The problem is still that when the user switch the filter, the extension doesn't change.

> A variation on this PR, namely instead of using a fake mscx extension, checking QFileDialog::selectedNameFilter to see whether the user was saving as mscz or mscx. This would also require changes to the Interactive(Provider) infrastructures, but that does seem doable.

Yes, it can help, but we need to think again whether this problem is so important to us.

https://user-images.githubusercontent.com/10116828/204510770-89537333-3429-44b0-a88b-cd42d95cb570.mp4

cbjeukendrup:
> The problem is still that when the user switch the filter, the extension doesn't change.

If this doesn't even work in a standard situation like this, then it must be a problem with that file dialog. We can try to use the `QFileDialog::selectedNameFilter` option, but we can't do much more than that.

> but we need to think again whether this problem is so important to us.

I agree; maybe we should ask Martin to reprioritise it?

Tantacrul:
I think this feature isn't important at all and I'm unclear why it is even being provided.

DmitryArefiev:
> I think this feature isn't important at all and I'm unclear why it is even being provided.

@Tantacrul As I remember, @shoogle knows the benefits of using mscx in MU4

DmitryArefiev:
@Tantacrul Concerning #11615, I retested and found this bug occurs on **Windows** and **Mac** as well.. But it's less critical because **file extensions are not shown by default** on Windows and Mac. 
If to turn on 'Show file extensions', the bug happens there too.

In the current PR #14365 the issue is fixed for **Mac**. At least we can merge those changes now and fix the part for Windows and Linux in Post-Release.

The easy solution is just to hide ability to save MSCX format in Release and renew it in Post-Release

DmitryArefiev:
How it works on **Mac** (if extensions are not hidden in OS) after #14365 

https://user-images.githubusercontent.com/90187801/205281953-89eb7577-6d9f-49d2-ab71-3145aaf443d4.mov

On **Windows** (with not hidden extensions) and **Linux** it looks like this

https://user-images.githubusercontent.com/90187801/205282434-2c79f575-d919-452e-bc50-3c5438986374.mp4

cbjeukendrup:
@DmitryArefiev I think it is even _more_ problematic when extensions are hidden (on macOS): you can't _see_ the extension, but it is still the wrong one, but you won't even know. So if you just switch to the mscx file type, then the extension is still mscz, and a mscz file will be created. (_before_ this PR)

cbjeukendrup:
@DmitryArefiev I made another tweak so that the fix should now also work on Windows

DmitryArefiev:
> @DmitryArefiev I made another tweak so that the fix should now also work on Windows

It still doesn't work on **Linux**.

On **Windows** and **Mac** it's good except one regression comparing with master: 
now it's doing old mscx file visible and shows Replace dialog but if to click Yes it won't replace old MSCX file and creates new MSCX folder

https://user-images.githubusercontent.com/90187801/205317748-2e259ab6-e1ca-4a3e-8dc8-2f166fabadb1.mp4

DmitryArefiev:
@shoogle How important to implement new MSCX format in 4.0? Can we postpone it to Post-Release?

shoogle:
@DmitryArefiev, for the release, I think we should keep the current behaviour as it is right now in master:

- MSCX is available as an option in the Save As dialog, marked as "experimental".
- You need to manually remove ".mscz" from the file name to get it to save as MSCX.

cbjeukendrup:
I'm closing this PR, because 
* it has complex conflicts
* judging from the comments above, it did not fully solve the problems
* it's so long ago that I have forgotten most about it by now

(See also https://github.com/musescore/MuseScore/pull/18645 and https://github.com/musescore/MuseScore/pull/27701)

PR #14386, opened 2022-11-07
fixed #11600: Vague MIDI assigning info in MIDI Device Mappings

### LINKED ISSUE #11600: [MU4 Issue] Vague MIDI assigning info in MIDI Device Mappings
So far, when assigning a note with a MIDI controller/keyboard, it is displayed as 'Note XX'. For example, Note C4 on the keyboard. It is able to recognise any note in any octave.

image

However, with other buttons on a MIDI keyboard such as a play or stop button, they are all labelled as 'Controller 0'. If they have a MIDI CC attached to it, it would be good to show the MIDI CC number value instead of 'Controller 0'.

Example, I have a Novation Impulse keyboard, I want to assign the play score button on MS4 to the play button on the Novation keyboard. Doing, this, MIDI Device Mappings recognises this button as 'Controller 0'. This label also shows up in other different buttons as well.

image

Eism:
Resolves: #11600

### DIFF
--- a/src/framework/shortcuts/internal/midiremote.cpp
+++ b/src/framework/shortcuts/internal/midiremote.cpp
@@ -209,7 +209,11 @@ bool MidiRemote::needIgnoreEvent(const Event& event) const
         return true;
     }
 
-    bool release = event.opcode() == Event::Opcode::NoteOff;
+    static const QList<Event::Opcode> releaseOps {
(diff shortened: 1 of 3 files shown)


jessjwilliamson:
That looks great! @abariska can you also test this looks good to you too?

abariska:
Already tested. Looks good to me

PR #14410, opened 2022-11-08
fixed tie position when note fret has a circle

alexpavlov96:


src/engraving/libmscore/tie.cpp
```diff
@@ -966,6 +967,24 @@ void Tie::slurPos(SlurPos* sp)
...
     }
+
+    /// adding extra space above ties for notes in circles
+    if (engravingConfiguration()->enableExperimentalFretCircle() && staff()->staffType()->isCommonTabStaff()) {
+        auto adjustTie = [](Chord* ch, PointF& coord) {
+            const Fraction halfFraction = Fraction(1, 2);
+            if (ch && ch->ticks() >= halfFraction) {
+                for (EngravingItem* item : ch->el()) {
+                    if (item && item->isFretCircle()) {
+                        coord -= PointF(0, toFretCircle(item)->offsetFromUpNote());
```
cbjeukendrup, line 978: It looks like this line will only work correctly for ties that are above the note, but they can also be below notes.

abariska:
🟢 **TESTED**
Please, merge to master

handrok (approved):

PR #14412, opened 2022-11-08
fixed #13028: More than one midi parameter can be mapped to one MU parameter

### LINKED ISSUE #13028: [MU4 Issue] Mapping. More than one midi parameter can be mapped to one MU parameter
**Describe the bug**
In this case I use NI Maschine MK3 Mikro (pad controller)
When I map a pad to Play action I just press the pad. In the settings I see corresponding midi input and note. But at the same time aftertouch parameter was catch when I pressed the pad and also mapped. I'm not sure it was mapped to Play also or to rewind but when I press the pad and unpress for a while playback is stuck on the first note. When aftertouch parameter is changed it plays from the start again and again. 
Second part of story.
When I map another pad to Rewind action aftertouch is also mapping to the Loop action, so when I press the pad it move playback to start but at the same time enable/disable Loop.

Also, I believe that only 0 and 127 values should be recieved for midi controllers which send values 0-127 during mapping. So they should be used as on/off controller till we have only simple toggle elements for mapping.

**Expected behavior**
Only Note On midi event is used for mapping when any note pressed

**Screenshots**

https://user-images.githubusercontent.com/101250347/188182484-b6958ffd-8eb4-4649-83cc-c12e790f3dcf.mov

**Platform information**
 - OS: macOS, Windows

**Additional context**
Could be linked with https://github.com/musescore/MuseScore/issues/13030 and https://github.com/musescore/MuseScore/issues/13031

Eism:
Resolves: #13028

### DIFF
--- a/src/framework/shortcuts/internal/midiremote.cpp
+++ b/src/framework/shortcuts/internal/midiremote.cpp
@@ -209,6 +209,11 @@ bool MidiRemote::needIgnoreEvent(const Event& event) const
         return true;
     }
 
+    if (event.opcode() != Event::Opcode::NoteOn && event.opcode() != Event::Opcode::NoteOn
+        && event.opcode() != Event::Opcode::ControlChange) {
+        return true;
+    }
+
     static const QList<Event::Opcode> releaseOps {
         Event::Opcode::NoteOff
     };


abariska:
🟢 **Tested** on macOS 12

RomanPudashkin (approved):

Issue #14426, opened 2022-11-08
[MU4 Issue] Crash when dragging glissando line


DmitryArefiev:
**To Reproduce**
1. Click and drag gliss grip
2. Click and drag gliss text

https://user-images.githubusercontent.com/90187801/200642774-ed01b96b-bbbd-42a9-8aac-ac126c46cec5.mp4

**Platform information**
macOS, Windows, Linux

lyrra:
Reproduced on linux.
<details><summary> backtrace + patch </summary>

```
#0  0x0000555555a8f6d4 in mu::engraving::ElementEditData::pushProperty (this=0x0, pid=mu::engraving::Pid::OFFSET2) at src/engraving/libmscore/engravingitem.h:606
#1  0x0000555555e766a2 in mu::engraving::LineSegment::startDrag (this=0x55556066e960, ed=...) at src/engraving/libmscore/line.cpp:221
#2  0x0000555555db7f4e in mu::engraving::SingleElementGroup::startDrag (this=0x55555e545b30, ed=...) at src/engraving/libmscore/elementgroup.cpp:29
#3  0x00005555564fdf1b in mu::notation::NotationInteraction::startDrag(std::vector<mu::engraving::EngravingItem*, std::allocator<mu::engraving::EngravingItem*> > const&, mu::PointX<double> const&, std::function<bool (mu::engraving::EngravingItem const*)> const&) (this=0x55555c0bd8a0, elems=std::vector of length 1, capacity 1 = {...}, eoffset=..., isDraggable=...)
    at src/notation/internal/notationinteraction.cpp:904
#4  0x00005555564a6cba in mu::notation::NotationViewInputController::startDragElements (this=0x55555cba97c0, elementsType=mu::engraving::ElementType::GLISSANDO_SEGMENT, elementsOffset=...)
    at src/notation/view/notationviewinputcontroller.cpp:742


class ElementEditData
{
...
    void pushProperty(Pid pid)
    {
        propertyData.push_back(PropertyData({ pid, e->getProperty(pid), e->propertyFlags(pid) })); <--- LINE 606
    }
...
}

----------------------------------------------

Possible fix (doesn't crash, can still drag by text (not possible in MU3)):

diff --git a/src/engraving/libmscore/line.cpp b/src/engraving/libmscore/line.cpp
index 2aa8631d1e..245be20c1b 100644
--- a/src/engraving/libmscore/line.cpp
+++ b/src/engraving/libmscore/line.cpp
[...]

PR #14433, opened 2022-11-09
Fix crash when dragging glissando

cbjeukendrup:
Resolves: #14426 

We assume that element data has just been created, so we use it without thinking about it. But that is not always the case; for non-movable LineSegment subclasses, no edit data will be created (see EngravingItem::startDrag, called from LineSegment::startDrag).

Adding a proper check reveals the next problem: Undoing dragging a glissando (segment) doesn't work well. Glissando segments are not marked movable, but in practice they are (is that maybe the bigger problem? Is the `MOVABLE` flag not honoured?? Anyway...). 
This causes problems: undoing a drag operation doesn't work well. To fix this, mark glissando segments as officially movable, so that the necessary data is recorded to make dragging glissando segments properly undoable.

For reference: in MS3, glissando segments were indeed not really movable. You could move them by dragging their middle grip, but if you try to move the glissando segment by dragging it itself, you end up panning the score view instead.

### DIFF
--- a/src/engraving/libmscore/glissando.cpp
+++ b/src/engraving/libmscore/glissando.cpp
@@ -72,7 +72,7 @@ static constexpr double GLISS_PALETTE_HEIGHT = 4.0;
 //=========================================================
 
 GlissandoSegment::GlissandoSegment(Glissando* sp, System* parent)
-    : LineSegment(ElementType::GLISSANDO_SEGMENT, sp, parent)
+    : LineSegment(ElementType::GLISSANDO_SEGMENT, sp, parent, ElementFlag::MOVABLE)
 {
 }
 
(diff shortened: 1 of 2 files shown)


RomanPudashkin (approved):

DmitryArefiev:
@cbjeukendrup Found two issues

Score 1: the gliss lines are shifted after opening the score in PR's build
Score 2: can't drag some gliss lines (but the same in master)

 

https://user-images.githubusercontent.com/90187801/200809621-f6282cbb-c059-4b4f-97e8-9d7b6bac47c9.mp4


 
[1 and 2.ZIP](https://github.com/musescore/MuseScore/files/9970053/1.and.2.ZIP)

cbjeukendrup:
@DmitryArefiev I fixed point 1, which also fixes the VTests. Fortunately it was a very minor fix :)

About point 2, maybe we should address that later, since it also matches the behaviour of the master branch and MS3.

DmitryArefiev:
> @DmitryArefiev I fixed point 1, which also fixes the VTests. Fortunately it was a very minor fix :)

Yes, looks good

> About point 2, maybe we should address that later, since it also matches the behaviour of the master branch and MS3.

Totally agree

DmitryArefiev:
Tested #14426 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #14287, opened 2022-11-03
[MU4 Issue] Layout at starts of bars is faulty, intermittently, in parts with multiple instruments


oktophonie:
**Describe the bug**
In parts, sometimes the padding between the barline and the first item after the barline is lost.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open a score with multiple instruments
2. Generate the parts (or at least one part)
3. Make _another_ instrument visible in that part
4. Note that at some places there is now no padding after the barline:
![image](https://user-images.githubusercontent.com/20604032/199686523-33d6e7bd-f848-4515-8777-0982019c0656.png)
![image](https://user-images.githubusercontent.com/20604032/199686568-fd5feb7f-3bf5-4f26-a6ef-1313195fd442.png)
![image](https://user-images.githubusercontent.com/20604032/199686619-1651a210-7e9b-4ed3-aa52-cf9d7131a8e6.png)

As far as I can tell, the places that this happens correspond to bars which are the first bar of a system _in the score_.

Note: if you then re-hide the second instrument, the problem remains.

Related to this (because it happens in bars which are the first bar of a system in the score):
1. Generate parts, go to one part, and show a second instrument in the part
2. Hide that instrument
3. Show the instrument _again_
4. Note that in some bars the clef and key signature become restated at that point(!):
![image](https://user-images.githubusercontent.com/20604032/199688112-bb816724-bc3f-4d8b-a10b-376fe3262c48.png)

These extra items will disappear if you hide the second instrument, will also reappear if you ever unhide it.
If you make a _different_ instrument visible, they will not be shown on that stave the _first_ time you do so, but will then appear any subsequent times.

**Platform information**
OS: Artix Linux, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223070418, revision: github-musescore-musescore-36a8ffa (Nightly 3 Nov)

Issue #11186, opened 2022-04-13
[MU4 Issue] Some items get duplicated or don't show between parts and score (when creating a new part/showing instruments)


mike-spa:
**Describe the bug**
Some items become duplicated between parts and score when creating a new part and adding instruments (see video attached). Especially clefs, time signatures and key signatures.

**Additional context**
Current master branch. Linux platform.
(also occurs on Mac/Win)

(The screen recording doesn't show the pop up menus, for unknown reasons)

https://user-images.githubusercontent.com/93707756/163221072-365f3935-daa3-49c5-80b5-9918f226fc5a.mp4

[video.zip](https://github.com/musescore/MuseScore/files/8487953/video.zip)
-----------------------------------------
**Part 1**:
- When creating a new part, **time signatures and key signatures are duplicated in the new part** (after showing other instruments). But all looks OK after reopening the score

https://user-images.githubusercontent.com/90187801/163380031-dcf70deb-f292-438f-b2e5-dd0d55e86a75.mp4

**Part 2**:
- When showing another instruments in the part, **fermatas are duplicated in the score** and **don't show in the part** (in unhidden instruments)

https://user-images.githubusercontent.com/90187801/163379012-4018fd19-f7c8-4ce2-9daa-2a989ee8cb97.mp4

**Part 3**:
- When creating a new part, **Tempo (and system text) don't show in the new part** (see video from Part 1)

DmitryArefiev:
@mike-spa I've reattached your video in zip package for easy download.

DmitryArefiev:
Found the same things with fermatas and Tempo (system) text. Divided into 3 parts and added to the description.

DmitryArefiev:
can be related to #11154

DmitryArefiev:
Fixed in #11260

PR #14446, opened 2022-11-09
Don't clone header segments from score to parts

mike-spa:
Resolves: #14287 

Header segments (beginBarLine, headerClef, timeSig, keySig) should not be cloned from score to parts because they are layout-dependent (and very likely will end up in different places in the part than they were in the score). More specifically, when cloning these segments we were loosing track of the `header()` flag, which was messing up with `Measure::removeSystemHeader()` and was causing these faulty spacing and mid-system headers problems. These segments are always generated at layout-time anyway, so it is unnecessary to clone them (which also saves a bit of time and reduces the number of total segments going around in parts).

image

image

Additional commit: also don't duplicate system objects when adding a new instruments to the part.

image

### DIFF
--- a/src/engraving/libmscore/excerpt.cpp
+++ b/src/engraving/libmscore/excerpt.cpp
@@ -1429,6 +1429,10 @@ void Excerpt::cloneStaff2(Staff* srcStaff, Staff* dstStaff, const Fraction& star
             TupletMap tupletMap;          // tuplets cannot cross measure boundaries
             track_idx_t dstTrack = map.at(srcTrack);
             for (Segment* oseg = m->first(); oseg; oseg = oseg->next()) {
+                if (oseg->header()) {
+                    // Generated at layout time, should not be cloned
+                    continue;
+                }
                 Segment* ns = nm->getSegment(oseg->segmentType(), oseg->tick());
                 EngravingItem* oef = oseg->element(trackZeroVoice(srcTrack));
                 if (oef && !oef->generated() && (oef->isTimeSig() || oef->isKeySig())) {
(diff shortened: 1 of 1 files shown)


oktophonie (approved):

src/engraving/libmscore/excerpt.cpp
```diff
@@ -1470,10 +1474,7 @@ void Excerpt::cloneStaff2(Staff* srcStaff, Staff* dstStaff, const Fraction& star
                     }
 
                     for (EngravingItem* e : oseg->annotations()) {
-                        if (e->generated()) {
-                            continue;
-                        }
-                        if (e->track() != srcTrack) {
+                        if (e->generated() || e->track() != srcTrack || e->systemFlag()) {
```
RomanPudashkin, line 1477: This change causes a regression of this issue: https://github.com/musescore/MuseScore/issues/11186

See the video:

https://user-images.githubusercontent.com/67867444/201324070-7585c16b-cf28-4b72-9bdd-2500c47c7e21.mp4

mike-spa: Ha! Thanks, I'll see if I can find a different fix

src/engraving/libmscore/excerpt.cpp
```diff
@@ -1470,10 +1474,9 @@ void Excerpt::cloneStaff2(Staff* srcStaff, Staff* dstStaff, const Fraction& star
                     }
 
                     for (EngravingItem* e : oseg->annotations()) {
-                        if (e->generated()) {
-                            continue;
-                        }
-                        if (e->track() != srcTrack) {
+                        if (e->generated()
+                            || (e->track() != srcTrack && !e->systemFlag()) // system items must be cloned even if they are on different tracks
+                            || (e->systemFlag() && score->nstaves() > 1)) { // ...but only once!
```
RomanPudashkin, line 1479: Ideally, you should assign `score->nstaves() > 1` to a temp variable outside the loop and use it here (to avoid redundant calculations). You can fix this in your next PR

RomanPudashkin (approved):

Issue #12829, opened 2022-08-12
[MU4 Issue] Default page size is A4 instead of Letter in North America


MarcSabatella:
**Describe the bug**
In North America, the standard paper size is Letter, and this should be the default page size in MuseScore.  But in MuseScore 4, it is A4, even in North America.

**To Reproduce**
1. Live in North America :-)
2. Create a new score, using either Choose instruments or Create from template
3. Go to Format / Page Settings

**Result**
Score is A4

**Expected behavior**
Score should be Letter

**Platform information**
 - OS: Windows

**Additional context**
In MU3, we used a bad hack to determine the default page size: we queried the default printer to find its default paper size.  This worked most of the time, but failed always on Linux, and occasionally on macOS and on Windows, particularly if your default printer was not accessible (e.g., turned off) or if the default printer was a virtual printer like a "print to PDF" utility.  With MU4, I'm not sure if the same hack is being attempted, but bottom line, it isn't working for me in North America even though my default (virtual) printer is set up with Letter as the default papr size.

Using the default printer to determine default page size is, as I suggested, a bad hack.  But for various reasons, no one could seem to agree on the better way that would work reliably across platforms.  Hopefully for MuseScore 4 we can do better!

MarcSabatella:
@Tantacrul - I forgot about this one, but this will be an enormous headache for American users.

Tantacrul:
Marc. Any chance you can submit a PR for this - or cajole a community member? It would help us focus on the trickier issues.

MarcSabatella:
Unfortunately, I don't know what the best solution would be - no one could agree on it back in the day.

But I'll try to drum up a consensus and see what I can do.  No doubt, it's a very simple fix no matter what strategy is chosen.

MarcSabatella:
My current PR https://github.com/musescore/MuseScore/pull/14457 is, I think, the best I personally can do for now.  It doesn't solve the problem for the user automaticall, but it at least makes sure the workarounds that have always existed (since this didn't always work reliably in MU3 either) continue to work in MU4.  So it's way better than nothing.

[...]

Issue #12828, opened 2022-08-12
[MU4 Issue] Custom default page size not honored when creating score from template


MarcSabatella:
**Describe the bug**
If you've specific a custom style file in your preferences in order to force MuseScore to use Letter instead of A4 (for example), this is ignored when creating a score from template

**To Reproduce**
1. Create a style (MSS) file to force page size to Letter (eg, set up a score that way, then Format / Save Style)
2. Set it as the default in Edit / Preferences / Score
3. Restart MuseScore to make sure the change takes effect
4. Create a new score from the treble template
5. Go to Format / Page Settings

**Result**
Score is A4

**Expected behavior**
Score should be Letter

**Platform information**
 - OS: Windows

**Additional context**
This is a very common issue in North America, where the default page size should normally be Letter instead of A4, but for any number of reasons, MuseScore has traditionally often failed to figure this out.  So the workaround has always been, use a custom style file to force the page size to letter.  In MuseScore 3, this works even when creating scores from templates, because we took special precautions to prevent templates from forcing A4 on everyone.  In MuseScore 4 this seems to not work any more.  Creating a score by choosing instruments picks up the page size from the style file, but creating from template, the A4 that is presumable hard-coded into it "wins".

MarcSabatella:
See also https://github.com/musescore/MuseScore/issues/12829, which is the reason I need to use the custom style in the first place.  To some extent, this is probably always going to be needed on some systems.  If that were not an issue on my own system - if my default page size was already Letter - I guess it's still an open question as to whether this would work for templates or only when choosing instruments.

PR #14457, opened 2022-11-10
Allow user to override default page size

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/12828
Resolves: https://github.com/musescore/MuseScore/issues/12829 (partially)

MuseScore currently sets all new scores to A4 page size.  This is inappropriate in North America; the default should be Letter.  There is no great way to make this determination automatically, but MuseScore does provide a preference setting to specify a custom style file for new scores and for parts.  Unfortunately, this setting was being completely ignored for parts, and also for scores created from templates as opposed to "from scratch".

This PR fixes the bug where the custom style for parts was being completely ignored, and also allows the page size setting from the user's custom style to take precedence over that of the template, if the template was using the default page size (the template still wins if it is explicitly specifying a larger or smaller page size).  It also allows the environment variable PAPERSIZE (used on some Linux systems) to provide an additional override, and it establishes a framework to allow the default page size to be made a user-visible preference in the future, which could perhaps be set to a more reasonable locale-dependent default automatically.

See commit messages for further details.


MarcSabatella:
The way I am approaching this is not the only way these problems could be addressed, and I'm happy to discuss further.  If someone wants to implement this using another approach instead, feel free to borrow my code.  One way or another, though, it's pretty important that those of us in North America have a way of makiing sure our scores default to Letter instead of A4.

Tantacrul:
@RomanPudashkin @Eism - could you merge this if you are happy with the implementation. I'm unable to test it myself because I'm in the UK. I've asked a few community members to have a look and confirm it's working.

MarcSabatella:
I've updated my PR and made my best attempt at setting the page size automatically, by checking the system locale and setting the default page size to letter if the country is any of the ones listed in the Wikipedia article as using that size.  There is no guarantee this will yield the right answer always, but at least this PR also fixes the workaround.

Here is the PR to test: https://github.com/musescore/MuseScore/issues/14498

Jojo-Schmitz:
Seems a rebase is needed?
You may get the "without Qt" check fixed with some `#ifdef`?

MarcSabatella:
> Seems a rebase is needed? You may get the "without Qt" check fixed with some `#ifdef`?

Rebased.  I'm happy to add ifdef around the Qt-specific code, but I'd need guidance in how to do this.  Is it just "#ifndef NO_QT_SUPPORT" as I see used a few other places?  I'll give that a shot.  The locale autodetect won't work, of course, but it will compile and still be better than the current situation, since at least the page size specified in preferences will be honored.

src/engraving/style/defaultstyle.cpp
```diff
@@ -47,15 +48,37 @@ DefaultStyle* DefaultStyle::instance()
...
+    case QLocale::UnitedStates:
+    case QLocale::Canada:
+    case QLocale::Mexico:
+    case QLocale::Chile:
+    case QLocale::Colombia:
+    case QLocale::CostaRica:
+    case QLocale::Panama:
+    case QLocale::Guatemala:
+    case QLocale::DominicanRepublic:
+    case QLocale::Philippines:
```
Jojo-Schmitz, line 74: Isn't it Nicaragua, El Salvador, Belize, Puerto Rico and Venzuela too?
As far as I can tell the Dominical Republic ises A4

image

MarcSabatella: My list is based on the Wikipedia article on letter size - https://en.m.wikipedia.org/wiki/Letter_(paper_size). But I’m happy to update based on more authoritative documentation.

cbjeukendrup:
@MarcSabatella I think that preferably the Qt stuff should be hidden at least in EngravingConfiguration (maybe even further away, but let's use EngravingConfiguration as a start). 

Perhaps you could create a `Size defaultPageSize()` method in (I)EngravingConfiguration, and hide the Qt Locale logic inside that method, and simply call the method in the DefaultStyle class. (You'll need to INJECT_STATIC `IEngravingConfiguration` in the DefaultStyle class (just like we do in the EngravingItem class.)

MarcSabatella:
Hmm, that's getting a little beyond what I'm comfortable with - this architecture of the code still is very new to me, and I'm operating without a build system, just using the online GitHub interface to make simple changes where I can, depending on CI to build for me, etc.  But I'm happy to donate my code here to someone who wants to then make it fit the architecture better.

src/engraving/style/defaultstyle.cpp
Jojo-Schmitz: Mine on the Image from https://en.wikipedia.org/wiki/Paper_size

PR #14470, opened 2022-11-11
Choose instruments: show trait name when there is only one option

### LINKED ISSUE #14150: [MU4 Improvement] Instrument selection not clear for new/first time use
**Describe the bug**
Some instruments have the option to select a different key while others don't. Excellent!
However when these are combined in the same dialog it is not clear which key is used for the instruments that do not have this selection option.
In my example I search for tuba and get several results. From the first step, I can't tell which key is used for the Bass Tuba (treble clef).
I need to select the instrument first, then look at the bottom of the dialog so see the extra information or add the instrument to the score and see it automatically add the "in Eb" text.
With this design it also feels as if there are options missing. I.e. why the different keys for the Bass Tuba but not for the Bass Tuba (treble clef)? 
I am sure there are more instruments which have this.

**To Reproduce**
1. Create a new score
2. In "choose instruments" search for tuba
3. Check the screen

**Expected behavior**
Make the key used for the instruments more obvious in the dialog. Perhaps just always show the key (even if there is only 1 option)?

**Screenshots**
image

**Platform information**
 - OS: Windows 10

**Additional context**

shoogle:
Fix #14150

This solves an ambiguity for instruments like the clarinet that have muliple traits (i.e. multiple transpositions, tunings or course options) in the New Score / Instruments dialog.

image

When the genres filter is set to 'Common' rather than 'All instruments', only one clarinet transposition is available and the trait dropdown is hidden. However, this meant that the user was unable to tell which trait is being used (i.e. whether it is the clarinet in A, Bb or C, etc.). I solved this by adding the trait name to the instrument name when there is only one option.

### Before
image

### After
image

 An alternative solution would be to make the trait dropdown always visible, but it would be frustrating for the user to expand a dropdown just to find that it only has one option.

### DIFF
--- a/src/instrumentsscene/view/instrumentlistmodel.cpp
+++ b/src/instrumentsscene/view/instrumentlistmodel.cpp
@@ -284,9 +284,19 @@ void InstrumentListModel::loadInstruments()
     QList<CombinedInstrument> instruments;
 
     for (const InstrumentName& instrumentName : templatesByInstrumentName.keys()) {
+        const InstrumentTemplateList& templates = templatesByInstrumentName[instrumentName];
         CombinedInstrument instrument;
-        instrument.name = instrumentName;
-        instrument.templates = templatesByInstrumentName[instrumentName];
+
+        if (templates.length() == 1) {
+            // Only one trait option so let's display it in the instrument name.
+            const InstrumentTemplate* templ = templates.at(0);
+            instrument.name = formatInstrumentTitle(instrumentName, templ->trait);
+        } else {
+            // Multiple traits to choose from so don't add any to instrument name yet.
+            instrument.name = instrumentName;
+        }
+
+        instrument.templates = templates;
         instrument.currentTemplateIndex = 0;
 
         instruments << instrument;


shoogle:
@oktophonie, notice in the screenshots above I added the transposition to the piccolo even though "in C" wouldn't be displayed in the actual score. I did this for these reasons:

1. I think it's best to know for certain which instrument you are getting before you try to add it. This will help to minimize the risk of mistakes.
    - For example, if an inexperienced user is transcribing a score that contains a "Piccolo in Db", they might grab the instrument called "Piccolo" and assume it's the right one when it's not.
2. Once added, the "in C" transposition is displayed in the **Your score** column of the dialog, so I think it is best to show it in both columns for consistency.

If you still want to hide the transpositions that would be hidden in the score then I can do so. The dialog would look like this:

image

oktophonie (approved):

src/instrumentsscene/view/instrumentlistmodel.cpp
```diff
@@ -284,9 +284,19 @@ void InstrumentListModel::loadInstruments()
     QList<CombinedInstrument> instruments;
 
     for (const InstrumentName& instrumentName : templatesByInstrumentName.keys()) {
+        const InstrumentTemplateList templates = templatesByInstrumentName[instrumentName];
```
RomanPudashkin, line 287: Minor optimization:
const InstrumentTemplateList **&** templates = templatesByInstrumentName[instrumentName];

RomanPudashkin (approved):

Issue #14184, opened 2022-10-30
[MU4 Issue] macOS: shortcut Cmd+F for "Find" doesn't work


cbjeukendrup:
**To Reproduce**
1. Press Cmd+F

Expected: "Find" panel opens

Actual:
- if there is a selection on the score, then the "Insert measures" popup appears, and the log output says that MuseScore thinks you pressed Ctrl+Ins
- if there is no selection, nothing happens. 

**Platform information**
 - OS: happens on macOS; on Linux it works fine (as reported by Marc); probably macOS specific

**Additional context**
Activating the "Find" panel using the "Edit" menu in the menu bar works fine. So it's purely a shortcuts mapping issue.

MarcSabatella:
In Linux at least, the search box appears.  It just doesn’t actually work - entering a measure number doesn’t take you there.  It select the measure but the view doesn’t change.

cbjeukendrup:
@MarcSabatella Yes, when I open the search box via the Edit menu, I get the same behaviour as you describe. But I think that is covered by #10096.

Tantacrul:
Since this was working not too long ago, it's a bad regression but I'm guessing easy to fix too.

cbjeukendrup:
It looks like this is not as trivial as it seems. It turns out that `MacOSShortcutsInstanceModel::doActivate ` is mapping Ctrl+F to Ctrl+Ins. So I think this is rather something for @Eism, who authored `MacOSShortcutsInstanceModel`.

Eism:
@cbjeukendrup 
I can't reproduce on the nighty build

cbjeukendrup:
@Eism That's strange... I can certainly reproduce still, both on nightlies and my own builds. Deleting my shortcuts.xml doesn't help. Can it be a macOS 13 issue?

cbjeukendrup:
Yes, really strange... I can still reproduce. I deleted my shortcuts.xml once again and I'm 100% sure I'm on the latest master, and it still happens for me.

Tantacrul:
I'm moving this to a P3 since it seems hard to reproduce. This should be the last thing we work on in this release.

Tantacrul:
It happened to me today. (MacOS 12.1)

I then cleared the **command+F** shortcut and re-added it again and then it worked. _Really_ odd.

Tantacrul:
So this is conclusively fixed right? I think we just forgot to close this issue.

cbjeukendrup:
[...]

PR #14485, opened 2022-11-12
Fix "Find" command

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/10096

Fixes the "Find" and "Get Location" commands to actually go to the selected location, and to close the search box on Esc.

The main issue preventing the command from working at all was just a missing "showItem()" call after selecting the appropriate element in the "Find" and "Get location" commands.  The failure to close the search panel on Esc was just a missing "onTextEditingFinished()" handler.  There was also missing code to handle showing a full page, so I added that to the implementation of "showItem()".

One thing this PR does not do is fix the behavior where keystrokes that represent valid editing shortcuts get passed to the score.  So hitting Enter when done with a search remains unwise - it closes the search panel, but still adds a system break to the score.  I am not sure how to fix that without changing the behavior of TextInputField, which seems dangerous, so I left that alone.  Maybe someone else can find a better fix for that.

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -5343,6 +5343,7 @@ void NotationInteraction::getLocation()
             score()->setPlayNote(true);
         }
         select({ e }, SelectType::SINGLE);
+        showItem(e);
     }
 }
 
@@ -5390,6 +5391,10 @@ void NotationInteraction::showItem(const mu::engraving::EngravingItem* el, int s
     } else if (el->isSpanner()) {
         EngravingItem* se = static_cast<const mu::engraving::Spanner*>(el)->startElement();
         m = static_cast<Measure*>(se->findMeasure());
+    } else if (el->isPage()) {
+        const mu::engraving::Page* p = static_cast<const mu::engraving::Page*>(el);
+        mu::engraving::System* s = !p->systems().empty() ? p->systems().front() : nullptr;
+        m = s && !s->measures().empty() ? s->measures().front() : nullptr;
     } else {
         // attempt to find measure
         mu::engraving::EngravingObject* e = el->parent();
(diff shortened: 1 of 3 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -5390,6 +5391,10 @@ void NotationInteraction::showItem(const mu::engraving::EngravingItem* el, int s
     } else if (el->isSpanner()) {
         EngravingItem* se = static_cast<const mu::engraving::Spanner*>(el)->startElement();
         m = static_cast<Measure*>(se->findMeasure());
+    } else if (el->isPage()) {
+        const mu::engraving::Page* p = static_cast<const mu::engraving::Page*>(el);
```
wizofaus, line 5395: isn't there a toPage( ) function?

src/notation/internal/notationinteraction.cpp
```diff
@@ -5343,6 +5343,7 @@ void NotationInteraction::getLocation()
             score()->setPlayNote(true);
         }
         select({ e }, SelectType::SINGLE);
+        showItem(e);
```
wizofaus, line 5346: I actually would've thought the select function would have a flag to indicate you want it scrolled into view - that'd probably be the default expectation...

Tantacrul:
Testing this, I can't get Command+F to work on my Mac.

And I can also confirm that I can't get it to work on any recent build. It seems to be working for the updated Beta release though, which makes me wonder whether there might be an issue with nightlies or PRs?

MarcSabatella:
The issue with Cmd+F not always working on macOS as the default shortcut is separate - see https://github.com/musescore/MuseScore/issues/14184.  Meanwhile, you can still test this by defining your own shortcut for Find, or activating it from the Edit menu.

Tantacrul:
OK, that's weird. I actually just cleared the shortcut and then re-added it (Command + F) and then it works.

RomanPudashkin (approved):

PR #14512, opened 2022-11-14
Fix #14459: Don't adjust staves for instrument name if the staff is hidden

### LINKED ISSUE #14459: [MU4 Issue] Cutaway staff causes unwanted horizontal gap at start of score
**Describe the bug**

A cutaway in an imported MS3 staff causes an unexplained gap at the start of the score in MS4.

**To Reproduce**

1. Open the attached score in both MS3 and MS4. 
[cutaway_issue.zip](https://github.com/musescore/MuseScore/files/9980226/cutaway_issue.zip)
2. Compare the layout of both scores. Note that a cutaway staff occurs in mm 38-39 of the original score.

RESULT: There is an unexplained gap at the start of the score in MS4. Apart from that the score seems to be OK.
image

**Expected behavior**

Score should look the same as it did in MS3.
image

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223140418, revision: github-musescore-musescore-e2a6c15

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/14459

The problem described in the issue linked above ended up being the staves adjusting for long instrument names at the start of the score even when the staff was hidden. This has been fixed so that the instrument names of hidden staves no longer factor in to the margin.

### DIFF
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -274,7 +274,7 @@ double System::instrumentNamesWidth()
 
     for (staff_idx_t staffIdx = 0; staffIdx < score()->nstaves(); ++staffIdx) {
         const SysStaff* staff = this->staff(staffIdx);
-        if (!staff) {
+        if (!staff || !staff->show()) {
             continue;
         }
 


asattely:
This creates more problems than it's worth at the moment and the workaround is fairly simple: uncheck hide empty staves and remove the instrument names from the cutaway staves, then rehide. The first system will no longer adjust for unseen long instrument names.

The reason why we want to keep adjusting for invisible names is because that way the systems will still line up and we won't run into situations like this:
image

So for the time being I think the workaround is what we want to use in the linked issue.

PR #14515, opened 2022-11-14
[MU4] Score 'Find/Go to' panel fix

Eism:
Resolves: #10096

Added navigation for panel and fixed the problem with pressing Enter and Esc


RomanPudashkin (approved):

Tantacrul:
@Eism - yes, this is working perfectly for me!

MarcSabatella:
Brilliant, thanks for finishing the job!  I had figured out that accepting the event was the key to getting the right behavior on Enter, but my attempts still had other shortcuts being passed to the score, so things like Up/Down were transposing.  Looks like this is fixed too now with the navigation changes.

PR #14528, opened 2022-11-14
gp-import: imported string number

alexpavlov96:


handrok (approved):

abariska (commented):
String numbers which already exist in palettes should be used.
To be fixed after adding string numbers till 10 by engraving team

handrok (approved):

abariska (approved):

PR #14532, opened 2022-11-15
Fix build with BUILD_UNITY being off

Jojo-Schmitz:


src/update/internal/updateactioncontroller.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include "translation.h"
 #include "log.h"
+#include "globalmodule.h"
```
Jojo-Schmitz, line 26: there might be a better solution here?

cbjeukendrup: You can just remove `using namespace mu::framework;` instead, because nothing from `framework` is used.

cbjeukendrup: And it looks like `#include "translation.h"` and `#include "log.h"` can go too...

Jojo-Schmitz: That's indeed much better ;-)

src/notation/inotation.h
```diff
@@ -23,6 +23,7 @@
 #define MU_NOTATION_INOTATION_H
 
 #include <QString>
+#include <QQuickItem>
```
cbjeukendrup, line 26: I don't think this should be in `inotation.h`...

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -23,6 +23,8 @@
 
 #include <QMimeData>
 #include <QtMath>
+#include <QWheelEvent>
+#include <QTimer>
```
cbjeukendrup, line 27: ...but rather here.

Also, I think it would be better to drop `<QWheelEvent>` from here, and instead `#include <QtEvents>` in `notationviewinputcontroller.h` (and then remove the `class QMouseEvent;` forward declaration from that file). This is because many kinds of Q…Events are used in the declarations and definitions of members of NotationViewInputController. 

Final request: I prefer to keep the includes alphabetically sorted within the "groups".

Jojo-Schmitz: Same for QQuickItem?

src/notation/inotation.h
Jojo-Schmitz: OK

src/notation/view/notationviewinputcontroller.cpp
cbjeukendrup: Nah, this looks fine to me now. 

Regarding the sorting, I meant sorting it like this:
```c++
#include <QMimeData>
#include <QQuickItem>
#include <QTimer>
#include <QtMath>
```
(By the way, Qt Creator has an option to sort lines in Edit > Advanced, which I find very convenient)

Jojo-Schmitz: Of course, didn't get my ABC right...

cbjeukendrup (approved):

src/notation/view/notationviewinputcontroller.h
```diff
@@ -22,6 +22,8 @@
 #ifndef MU_NOTATION_NOTATIONVIEWINPUTCONTROLLER_H
 #define MU_NOTATION_NOTATIONVIEWINPUTCONTROLLER_H
 
+#include <QEvent>
```
cbjeukendrup, line 25: Oh, sorry, I missed something here: this should be `<QtEvents>` (≠ `<QEvent>`). `<QEvent>` is an alias for `"qcoreevent.h"`, which won't help here. We need `<QtEvents>` instead, which is an alias for `"QtGui/events.h". That contains the definitions of the events we need.

Jojo-Schmitz: done ;-)

PR #14545, opened 2022-11-15
Fix #12530 (candidate) Apply !showbacktied behavior per-system rather than per-measure

### LINKED ISSUE #12530: [MU4 Issue] Tablature: "Show back-tied notes" option and "continuation parentheses" are not working properly
**Describe the bug**

I suspect these issues are linked. You can easily check the correct behaviour in MS3.

(1) When the "Show back-tied notes" option is OFF, MS4 is continuing to show the tied notes with spurious parentheses (system-spanning ties are OK).

ms4_off

The _correct_ behaviour is that subsequent tied notes and ties should be invisible; and parentheses, and their enclosed fretmarks, should only _automatically_ appear if the tie would have extended from one system to another – as in MS3.

(2) When "Show back-tied notes" option is ON, spurious parentheses are again appearing in error – even on system-spanning ties.

ms4_on

The _correct_ behaviour is that, in this case, parentheses should _never_ appear – as in MS3.

**To Reproduce**

1. Open the MS4 file, 
2. Observe the appearance of tablature ties when  "Show back-tied notes" is OFF and ON (Staff / Part properties > Advanced style properties)

**Expected behavior**

Tied notes and ties in tablature should always behave as in MS3.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2722421418, revision: github-musescore-musescore-5e7884d

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/12530

See discussion in the linked issue for details.
image


oktophonie (approved):

vpereverzev:
@Tantacrul quite a risky one, bumping to post-release, since it's a P3 anyway

src/engraving/libmscore/note.cpp
```diff
@@ -2360,9 +2356,18 @@ void Note::layout()
 
 void Note::layout2()
 {
+    const StaffType* tab = staff()->staffTypeForElement(this);
```
RomanPudashkin, line 2359: it should be "staffType", not "tab"

RomanPudashkin:
any new tests?

src/engraving/libmscore/note.cpp
```diff
@@ -2379,19 +2384,15 @@ void Note::layout2()
...
-                // with TAB's, dot Y is not calculated during layoutChords3(),
-                // as layoutChords3() is not even called for TAB's;
-                // setDotY() actually also manages creation/deletion of NoteDot's
-                setDotY(DirectionV::AUTO);
-
-                // use TAB default note-to-dot spacing
-                dd = STAFFTYPE_TAB_DEFAULTDOTDIST_X * spatium();
-                d = dd * 0.5;
-            }
+        if (staff()->isTabStaff(chord()->tick()) && tab->stemThrough()) {
```
RomanPudashkin, line 2387: This staff()->isTabStaff(chord()->tick()) is already called at least twice in one function

asattely: no longer!

RomanPudashkin (approved):

Issue #14210, opened 2022-10-31
[MU4 Issue] Brackets behave inconsistently in parts as instruments are shown or hidden


oktophonie:
**Describe the bug**
A short and clear description of what the bug is.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create, for example, a string quartet setup (or use the String Quartet template)
2. File > Parts > Open all, then go to the Violin I part
3. Switch to the Instruments panel
4. Show or hide instruments in any combination. When more than one instrument is shown, they will be joined by a thick bracket as expected
5. Now go to any other part
6. As far as I can tell, if you show the Violin I part _before any other_, the various instruments will be bracketed together as expected (even if the Violin I is subsequently hidden). If you open any of the other parts, they will not. I assume this is because the bracket originates from the Vl I stave.

**Expected behaviour**
Brackets should be applied consistently, according to the score layout.,

**Screenshots**
Vl II part where Vla was shown:
![image](https://user-images.githubusercontent.com/20604032/199030805-a9d5cf96-8a28-453d-bf01-6ccb3ec42edc.png)

Vl II part where Vl I was shown, _then_ Vla was shown, then Vl I hidden again:
![image](https://user-images.githubusercontent.com/20604032/199030920-3278f738-5da6-4b25-878c-75dfac710a49.png)

**Platform information**
OS: Artix Linux, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223040427, revision: github-musescore-musescore-a5ef471

PR #14551, opened 2022-11-16
Fix brackets and barline in parts

### LINKED ISSUE #15006: [MU4 Issue] Grand staff barlines not connected when inserted in score w/ multiple instruments
**Describe the bug**
Instruments that use grand staves (e.g. piano, harp) have their barlines not connected to each other when score has multiple instruments involved.

**To Reproduce**
1. Go to _New Score_.
2. Insert multiple instruments w/ at least one instrument that uses a grand staff
3. Complete score creation as of liking

**Expected behavior**
Grand staff barlines should be conjoined in this case.

**Screenshots**
image

**Platform information**
 - OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223410309, revision: github-musescore-musescore-c26899b

**Additional context**
- Otherwise, this doesn't occur when the instruments in question are inserted _after_ the score is created.
- I'm surprised this is not logged yet (or at least, I can't find any logged issues related to this) since I've observed this since the alpha release, but I assume this must be already known. I'll anyhow log this in just in case. Close if duplicate or unnecessary.

mike-spa:
Resolves: #14210

Currently, system brackets and barlines extensions are mapped on a staff-by-staff basis when cloning staves from score to parts. It is impossible for this to work correctly, because many of the staves that are spanned by a given bracket or barline in the score may literally not exist in the parts, and indeed we run into all kinds of trouble:
If the first staff of the bracket isn't in the part, the bracket will not be cloned:
image image

If the first staff of the bracket _is_ in the part, the bracket will be cloned but will span as many staves as it does in the score, leading to incorrect layout in the part:
image

Barlines can become wrongly connected:
image

The only reliable way to correctly map brackets and barlines from score to parts is to map them on a score-by-score basis (basically, look at brackets and barlines in the score, see how many of the spanned staves are also present in the part, and adapt accordingly). This has to be done when the part is generate or when a new instrument is added to it.

Edit:
The last commit also resolves #15006 , which had wrong barline extensions on piano at score creation.

Edit-edit:
Last commit has been moved to separate PR.

### DIFF
--- a/src/engraving/libmscore/engravingobject.cpp
+++ b/src/engraving/libmscore/engravingobject.cpp
@@ -677,6 +677,22 @@ bool EngravingObject::isLinked(EngravingObject* se) const
 }
 
 //---------------------------------------------------------
+//   findLinkedInScore
+///  if exists, returns the linked object in the required
+///  score, else returns null
+//---------------------------------------------------------
+
+EngravingObject* EngravingObject::findLinkedInScore(Score* score) const
+{
+    if (score == this || !_links || _links->empty()) {
+        return nullptr;
+    }
+    auto findElem = std::find_if(_links->begin(), _links->end(),
+                                 [score](EngravingObject* engObj) { return engObj && engObj->score() == score; });
+    return findElem != _links->end() ? *findElem : nullptr;
+}
+
+//---------------------------------------------------------
 //   undoUnlink
 //---------------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


oktophonie (approved):

RomanPudashkin:
@mike-spa

>Edit:
The last commit also resolves https://github.com/musescore/MuseScore/issues/15006 , which had wrong barline extensions on piano at score creation.

Is it possible to move it to a separate PR? For RC we should make as few changes as possible

Issue #13631, opened 2022-10-05
[MU4 Issue] crash in homegrown string library


lyrra:
Another bug, time to switch to libunistring?
commit height: bac0270ed
(gdb) ba
#0  0x00007ff617ad8bac in std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::size (this=0x0) at mingw64/include/c++/12.2.0/bits/basic_string.h:1064
#1  0x00007ff617ad8f5f in std::__cxx11::basic_string<char16_t, std::char_traits<char16_t>, std::allocator<char16_t> >::operator std::basic_string_view<char16_t, std::char_traits<char16_t> > (this=0x0) at mingw64/include/c++/12.2.0/bits/basic_string.h:937
#2  0x00007ff616612200 in mu::String::arg (this=0x8c4a1fb790, val=...) at src/framework/global/types/string.cpp:870
#3  0x00007ff6179dce89 in mu::mpe::PlaybackSetupData::toString (this=0x7ff618079f40 <mu::mpe::GENERIC_SETUP_DATA>) at src/framework/mpe/events.h:313
#4  0x00007ff617620ac5 in __static_initialization_and_destruction_0 (__initialize_p=0x1, __priority=0xffff) at src/framework/mpe/events.h:346
#5  0x00007ff617622694 in _GLOBAL__sub_I__ZN2mu9engraving19ChordLineMetaParser7doParseEPKNS0_13EngravingItemERKNS0_16RenderingContextERNS_3mpe15ArticulationMapE () at src/engraving/playback/metaparsers/internal/chordlinemetaparser.cpp:67

cbjeukendrup:
In which situation does this crash occur? What are the steps to reproduce it?

lyrra:
Mingw build, crashes at startup without any arguments.
I'll try to dig some more if nothing obvious pops out from the stacktrace above.

lyrra:
Thinking if subCategorySet is empty and string-join cant handle it (if so, it wasn't empty some commits ago).
Trying out below patch to see whats going on:
```
diff --git a/src/framework/mpe/events.h b/src/framework/mpe/events.h
index 4a9e465904..673d062885 100644
--- a/src/framework/mpe/events.h
+++ b/src/framework/mpe/events.h
@@ -33,6 +33,8 @@
 #include "mpetypes.h"
 #include "soundid.h"
 
+#include "log.h"
+
 namespace mu::mpe {
[...]

PR #14555, opened 2022-11-16
Fix GH#13631: Crash in homegrown string library

Jojo-Schmitz:
Resolves: #13631

### DIFF
--- a/src/framework/mpe/soundid.h
+++ b/src/framework/mpe/soundid.h
@@ -398,7 +398,7 @@ enum class SoundSubCategory
     Last
 };
 
-static const std::unordered_map<SoundId, String> ID_STRINGS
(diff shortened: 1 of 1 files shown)


lyrra:
Confirming this PR resolves the crash during startup (on mingw).

Jojo-Schmitz:
@igorkorsukov mind to review and merge?

igorkorsukov (approved):

igorkorsukov:
> @igorkorsukov mind to review and merge?

I believe that we need to move these lists to a cpp. The only difficulty is that we need to decide whether to hide behind the interface or make a module with a hard dependency, require it when linking. But it is certainly better to make it a separate task.

PR #14561, opened 2022-11-16
Set default page size based on locale

cbjeukendrup:
Resolves: https://github.com/musescore/MuseScore/issues/12828
Resolves: https://github.com/musescore/MuseScore/issues/12829

Building on the work that @MarcSabatella did in #14457, thus:
Closes: #14457

### DIFF
--- a/src/engraving/iengravingconfiguration.h
+++ b/src/engraving/iengravingconfiguration.h
@@ -48,6 +48,8 @@ public:
     virtual io::path_t partStyleFilePath() const = 0;
     virtual void setPartStyleFilePath(const io::path_t& path) = 0;
 
+    virtual SizeF defaultPageSize() const = 0;
+
     virtual String iconsFontFamily() const = 0;
 
     virtual draw::Color defaultColor() const = 0;
--- a/src/engraving/internal/engravingconfiguration.cpp
+++ b/src/engraving/internal/engravingconfiguration.cpp
@@ -21,6 +21,13 @@
  */
 #include "engravingconfiguration.h"
 
+#include <cstdlib>
+
+#ifndef NO_QT_SUPPORT
+#include <QLocale>
+#include <QPageSize>
+#endif
+
 #include "global/settings.h"
 #include "draw/types/color.h"
 #include "libmscore/mscore.h"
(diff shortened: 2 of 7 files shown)


MarcSabatella:
Thanks for updating this!  Definitely looks cleaner, and losing the pointers for default style makes a lot of sense.

One other thing worth mentioning:

I was deliberately letting the default style overriding the template *only* for page size.  For other style settings, the template wins, even when it's the same as base style.  So, a user who wants their default music font to be Bravura, or staff size to always be extra large, or text to always use FreeSerif, will still find the templates force the usual MU4 defaults.  In MU3, we actually let the user defaults override *all* template style settings.  You could create really whatever collection of default style settings you wanted and they would all be honored when creating scores from templates.

But, that's not because we do anything special upon creating the score from the template.  It's because in MU3, templates (like all scores) don't contain explicit style tags for any settings that are unchanged from the base style.  So the default style that is initially assigned to the newly created score never gets replaced by style settings from the template, except where the template explicitly deviates from base style.  

In MU4, however, the MSS file for each score contains *all* style settings.  I understand this was a conscious change, and I can see reasons for it - so that if we subsequently change a default style setting, existing scores are not adversely affected.  We *could* consider not doing this for templates, however - only storing *modified* style settings in templates.  That could mean having a special "save as template" command - not a bad idea in my opinion, as it would expose this functionality to the user.

But, if we wanted to mimic the MU3 behavior when creating scores from templates, *without* modifying the templates or adding a new user-visible command, we could modify the code in MasterNotation::applyOptions() to run through all style settings and for any that are currently same as base style, replace them with the corresponding default style setting (as opposed to special casing the page settings as we are here in the PR).

Whether or not we *want* user style defaults to continue "win" over templates in this way as they did in MU3 is debatable, but probably some users expect it.

RomanPudashkin (approved):

cbjeukendrup:
About the VTest difference, my understanding is the following:
Now, the default page size is determined by `QPageSize::size(defaultPageSizeIsLetter() ? QPageSize::Letter : QPageSize::A4, QPageSize::Inch)`. Before, it was determined by `210.0 / INCH` and `297.0 / INCH`. That gives apparently a slightly different value, which causes rounding differences during layout, which causes VTest differences. So I think it's nothing major. 

But to be on the safe side: @oktophonie, if you have a moment, it would be great if you could take a quick look on this too :)

oktophonie (approved):

RomanPudashkin:

cbjeukendrup:
Done!

Issue #11740, opened 2022-05-25
For notes, the "v" keyboard shortcut is not connected to "visible" toggle in property panel


mike-spa:
**Describe the bug**
The "v" keyboard shortcut for "visible" _does_ cause the selected item to become invisible, but that is not connected to the "visible" toggle in the property panel for that item. It looks as if the shortcut makes it invisible on screen but does not change the underlying `visible()` property. Indeed, if I make notes invisible via the property toggle, they will be spaced as invisible, but if I use the shortcut their positioning is not affected.

**@Tantacrul edit:** Worth mentioning that this only seems to happen for notes. It doesn't happen for most other elements.

So if I click on a notehead or stem and press 'V', the whole note (notehead, stem, flag) become invisible but the visible checkbox is not marked.

**To Reproduce**
Play around with the "v" shortcut and the "visible" toggle.

**Expected behavior**
Unless I'm missing something, I would expect the shortcut to act as if I'm pressing the toggle, and the two to be connected.

**Screenshots**

https://user-images.githubusercontent.com/93707756/170217604-21f202dd-f7d2-4e44-a997-fc792159ca87.mp4

**Platform information**
 - OS: Linux

**Additional context**
Current master

bkunda:
@VanSHOE this might be a good one for you to kick-start your work on shortcuts :-)

VanSHOE:
The main issue is that the inspector does not refresh upon pressing v

Regarding this, I can cause a refresh of the inspector page by sending a notification on `selectionChanged()` and catching it at the inspectorModel whenever the hotkey is pressed
But with that some other issues are coming up (essentially, the inspector is updated well for the whole measure selections but not for individual selections)
I'll continue to look into the same

cbjeukendrup:
@VanSHOE Have a look at `AbstractInspectorModel::updatePropertiesOnNotationChanged()` and how that's used.

VanSHOE:
> @VanSHOE Have a look at `AbstractInspectorModel::updatePropertiesOnNotationChanged()` and how that's used.

Thank you! Ill have a look into that!

cbjeukendrup:
[...]

PR #14580, opened 2022-11-17
Fix #13487: Retain MU3 behavior by hiding dots when notehead (or rest) is hidden

### LINKED ISSUE #13487: [MU4 Issue] Making dotted note, or rest invisible doesnt make dot invisible - regression
**Describe the bug**
When user makes dotted note, or rest invisible, dot reamins vible.

**To Reproduce**
1. Select doted note (or rest)
2. Press "v", or untick "Visible"
3. Dot remains visible.

**Expected behavior**
Dot is marked invisible,or visible together with note (or rest)

**Screenshots**

https://user-images.githubusercontent.com/1646034/192146645-cc06869b-acdd-4e7a-9d12-8acc9ecaf222.mp4

**Platform information**
 - OS: [reported on Linux]

**Additional context**
May be (or may be not) related to #13114

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/13487

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -46,6 +46,7 @@
 #include "mscore.h"
 #include "navigate.h"
 #include "note.h"
+#include "notedot.h"
 #include "noteevent.h"
 #include "part.h"
 #include "score.h"
@@ -1392,6 +1393,9 @@ void Chord::processSiblings(std::function<void(EngravingItem*)> func) const
     }
     for (Note* note : _notes) {
         func(note);
+        for (NoteDot* noteDot : note->dots()) {
+            func(noteDot);
+        }
     }
     for (Chord* chord : _graceNotes) {    // process grace notes last, needed for correct shape calculation
         func(chord);
--- a/src/engraving/libmscore/rest.cpp
+++ b/src/engraving/libmscore/rest.cpp
@@ -1091,6 +1091,21 @@ bool Rest::setProperty(Pid propertyId, const PropertyValue& v)
     return true;
 }
 
+void Rest::undoChangeProperty(Pid id, const PropertyValue& newValue)
+{
+    undoChangeProperty(id, newValue, propertyFlags(id));
+}
+
+void Rest::undoChangeProperty(Pid id, const PropertyValue& newValue, PropertyFlags ps)
+{
+    if (id == Pid::VISIBLE) {
+        for (NoteDot* nd : m_dots) {
+            nd->undoChangeProperty(id, newValue, ps);
+        }
+    }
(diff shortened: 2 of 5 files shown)


asattely:
This PR deals with visibility when the user selects a part of the note and presses the `v` shortcut. However, due to https://github.com/musescore/MuseScore/issues/11740 (I believe), the properties panel currently does not work when you attempt to hide only the dot--instead, it hides or shows the entire chord when any part of that chord is selected. I am now looking into that issue and may (but not necessarily will) fold that fix into this PR if it's simple. If it's not simple then I'll submit a PR for that first and we can merge this afterwards.

asattely:
After a buncha back-and-forth with @oktophonie and @Tantacrul, this PR has been amended with the ability for note dots (as a whole, though if we want it to work on an individual basis that's an easy fix) to be affected by the properties panel independently from the rest of the chord. If people want individual dots hidden, they can still use the `v` shortcut.

Tantacrul:
OK, this is working fine for me. Thanks very much!

Tantacrul:
Actually. Hold on... I just got a crash when pressing V on a note within a chord. Very easy to reproduce.

https://user-images.githubusercontent.com/21022311/204672954-b5e1dac1-62a6-4f12-b033-21ecbaa4fb87.mov

oktophonie:
I can replicate this crash but only very unpredictably - i.e. if I make a few chords, and toggle the invisibility on and off over and over again (perhaps mixing in toggling the visibility of a whole bar) it will crash __eventually__, but it takes some time!

Tantacrul:
When I test in Windows it is a little harder but I've now done it about a dozen times, so there's definitely a problem. Mac may be easier to test on

Tantacrul:
Tested again on Windows. Unfortunately can still get a crash (after a few attempts) where I make a note (that sits within a chord) invisible and then undo it.

I think we're going to have to bump this to 4.x since the instability issue makes it dangerous to merge now.

asattely:
I'm unable to reproduce any crash on windows, using the Debug or RelWithDebInfo builds, or even with the build artifact from this PR. @Tantacrul if you are still able to reproduce this on windows, could you take a video of it so I know the exact steps that are happening?

Tantacrul:
@oktophonie @asattely @RomanPudashkin 

I think we need to park this for a moment. @oktophonie, @bkunda and I have just agreed on a proper design improvement, which seeks to address this problem in a more robust way. In the process I've learned a lot about how this whole system works. Once we've agreed on the design I'll share it and we can then plan the best way forward.

One current confusion is that @RomanPudashkin and @asattely have similar PRs which do slightly different things.

Issue #12887, opened 2022-08-16
[MU4 Issue] "Cannot split tuplet" error is not given when changing time signature


oktophonie:
**Describe the bug**
When changing a time signature in existing music, which causes the following music to be rebarred, if it were impossible for the music to be rebarred due to a tuplet that would now have to cross a barline, MU3 would alert the user with an error window:

![image](https://user-images.githubusercontent.com/20604032/184880270-53cf4127-b409-4a1b-93d1-b1058e8eee25.png)

MU4 currently just fails silently. (As though the "Show this message again" checkbox had been unticked, though it hasn't.)

UPDATE: related is this:
![image](https://user-images.githubusercontent.com/20604032/186210443-b38d279a-8ad7-4a19-aa83-5986fb8815f2.png)
which 3.6 would give if you Add -> Tuplet -> Other and put in a ratio that makes sense (i.e. 4:3 on an undotted note). Quoth @asattely: "_the message box is libmscore calling a qt warning box unilaterally, and there was a big push to divorce libmscore from qt, so that's not an option anymore_".

**To Reproduce**
Steps to reproduce the behaviour:
In a score such as this:
![image](https://user-images.githubusercontent.com/20604032/184886574-369134a0-8315-4720-bb17-fb7083f74f06.png)
change the initial time signature to 3/4, for example.

**Expected behavior**
The user should get some feedback about the rebarring not happening.

cbjeukendrup:
It looks like this applies to many more error dialogs. `MScore::setError` is used in a lot of places, but the errors reported through that function are never handled. 
It must be said though that this is not _always_ a problem because in some cases there might be some error prevention on higher levels like NotationInteraction and NotationActionController.
But to name just one example, the "cannot split measure repeat" dialog never appears (it should be shown when trying to split or join measures where one of them contains a measure repeat). There are many more, but all seem somewhat niche, most of them relating to tuplets, local time signatures or measure repeats.
It must also be said that this is only a user information problem: simply nothing happens. It's not that MuseScore or the score can get into an inconsistent state.

PR #14601, opened 2022-11-18
Add missing error dialogs that should be the result of MScore errors

cbjeukendrup:
Resolves: #12887

The error handling in MS3 was relatively straightforward: when an error occurred, this was stored in MScore and the at the end of the current command, the command would be rolled back in Score::endCmd. Then, there's some other method that is _also_ called after every command, which checks the error and displays them. It was perhaps not the most sophisticated system, but at least somewhat consistent. 

In MS4, it has become quite a mess. We added some ad-hoc error handling methods, and some "error prevention", sometimes with or without appropriate UI message. The MScore error field was not checked, resulting in a lot of missing UI messages. 

To some extent, I'm bringing back the MScore error checking, and removing some of the ad-hoc logic. It is certainly not beautiful, and in the future we must come up with something more clever, but at least it brings back the missing error dialogs for 4.0.

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -23,6 +23,8 @@
 #include <map>
 #include <set>
 
+#include "infrastructure/messagebox.h"
+
 #include "accidental.h"
 #include "articulation.h"
 #include "barline.h"
@@ -243,7 +245,9 @@ Tuplet* Score::addTuplet(ChordRest* destinationChordRest, Fraction ratio, Tuplet
 
     Fraction fr = f * Fraction(1, _ratio.denominator());
     if (!TDuration::isValid(fr)) {
-        // todo: there needs to be some kind of user feedback for failure to add tuplet
+        MessageBox::warning(mtrc("engraving", "Cannot create tuplet with ratio %1 for duration %2")
+                            .arg(_ratio.toString(), f.toString()).toStdString(),
+                            std::string(), { MessageBox::Ok });
         return nullptr;
     }
 
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -218,6 +218,38 @@ void NotationInteraction::rollback()
     m_undoStack->rollbackChanges();
 }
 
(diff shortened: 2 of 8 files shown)


src/engraving/libmscore/edit.cpp
```diff
@@ -99,20 +101,6 @@ static ChordRest* chordOrRest(EngravingItem* el)
...
-//   getSelectedNote
-//---------------------------------------------------------
-
-Note* Score::getSelectedNote()
-{
-    EngravingItem* el = selection().element();
-    if (el && el->isNote()) {
-        return toNote(el);
-    }
-    MScore::setError(MsError::NO_NOTE_SELECTED);
```
wizofaus, line 112: surely we can do better than using a static/global variable though...why not make it a proper member?

src/notation/internal/notationinteraction.cpp
```diff
@@ -216,6 +218,149 @@ void NotationInteraction::rollback()
...
+    }
+
+    return {};
+}
+
+void NotationInteraction::checkAndShowMScoreError() const
+{
+    TRACEFUNC;
+
+    MsError err = MScore::_error;
```
wizofaus, line 257: again, why not use an actual instance variable/field?  This error might have been set in another part etc...not relevant to this notation window.

RomanPudashkin (approved):

src/notation/internal/notationinteraction.cpp
```diff
@@ -216,6 +216,159 @@ void NotationInteraction::rollback()
     m_undoStack->rollbackChanges();
 }
 
+static std::string MScoreErrorToString(MsError err)
```
Eism, line 219: It would be cool to put this methods in a separate file.
I'm sure it will annoy me. So long to scroll.

cbjeukendrup: Done!

Eism: I meant also the content of checkAndShowMScScoreError

Tantacrul:
OK, I can trigger the two dialogs I'm after. I think we're good.

Issue #13760, opened 2022-10-14
Chord Symbols should 'play' once when they have been created.


Tantacrul:
**Describe the bug**
In MS3, after a chord has been created, we hear the sound of the chord. We do not hear this in MS4.

**To Reproduce**
See video:

https://user-images.githubusercontent.com/21022311/195846946-a48ce379-6cbe-4b23-b0d0-a053db6d2d34.mov

**Additional context**
Please ignore my comments on volume / velocity in the video. These are covered in a different issue already.

Also, I'm aware it's an F **minor** 7, so don't mention that you pedants!

MarcSabatella:
FWIW, the chord *does* play during input if you hit Space to move to the next beat/note (or use any of the navigation shortcuts for chord symbol entry) - it only fails is when terminating chord symbol input.  So obviously the code to play during input is there already, it's just not being triggered in that case.

Issue #13602, opened 2022-10-03
[MU4 Issue] Assertion failure when entering a chord symbol in a just created/opened score (crash in debug builds)


cbjeukendrup:
**Describe the bug**
A short and clear description of what the bug is.

**To Reproduce**
NOTE: The crash will only occur in Debug builds, so not in any build created on GitHub Actions.
1. Create a score
2. Select the first measure rest
3. Press Cmd/Ctrl+K
4. Type "C"
5. **Press Enter** <- this specifically causes the assertion failure (if you just click on empty space to finish editing the chord symbol, it works fine)
-> Assertion failure: `(curCmd == 0), function reopen, file undo.cpp, line 475.`.

**Platform information**
 - OS: reported on macOS

**Additional context**
Might not be limited to chord symbols and newly created scores

lyrra:
duplicate of? https://github.com/musescore/MuseScore/issues/13105

Issue #13105, opened 2022-08-31
[MU4 Issue] enter chordsymbol+return assert


lyrra:
Enter a note,
press ctrl-k
Type in some chord symbol and press enter.
Assert.

MarcSabatella:
But see also https://github.com/musescore/MuseScore/issues/10119 - Return is not actually *supposed* to do anything in particular here, although there are proposals for possible future behaviors.  Definitely crashes aren't good, but assert failures won't crash in release builds.

PR #14603, opened 2022-11-18
Play chord symbol after creating/editing and fix assertion failure

cbjeukendrup:
Resolves: #13105
Resolves: #13602
Resolves: #13760

### DIFF
--- a/src/notation/internal/notationinteraction.h
+++ b/src/notation/internal/notationinteraction.h
@@ -142,7 +142,7 @@ public:
     const TextBase* editedText() const override;
     async::Notification textEditingStarted() const override;
     async::Notification textEditingChanged() const override;
-    async::Notification textEditingEnded() const override;
+    async::Channel<TextBase*> textEditingEnded() const override;
 
     // Grip edit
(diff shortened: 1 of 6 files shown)


src/notation/internal/notationinteraction.h
```diff
@@ -50,6 +52,7 @@ class NotationInteraction : public INotationInteraction, public async::Asyncable
 {
     INJECT(notation, INotationConfiguration, configuration)
     INJECT(notation, ISelectInstrumentsScenario, selectInstrumentScenario)
+    INJECT(notation, playback::IPlaybackController, playbackController)
```
RomanPudashkin, line 55: You should not inject this interface here. It would be better if INotationInteraction::textEditingEnded sent the edited element. We can use this element to play it

cbjeukendrup: Thanks, done!

Tantacrul:
Tested this. It's looking good! Thanks @cbjeukendrup!

@RomanPudashkin @vpereverzev - good to merge!

RomanPudashkin (approved):

Issue #10672, opened 2022-02-25
[Inspector] 'Hide stem' option should also hide note flags


Tantacrul:
**Describe the bug**
Here, the 'hide stem' option is ticked but the flag is still displayed.

<img width="666" alt="image" src="https://user-images.githubusercontent.com/21022311/155810753-31d11793-a385-44c6-a533-2e3589b6c44f.png">

**Expected behavior**
Hiding the stem should also hide flags

Tantacrul:
Note that we do this in MuseScore 3:
<img width="399" alt="image" src="https://user-images.githubusercontent.com/21022311/155810965-a97a8282-3af8-438c-977b-2348ca6ce62b.png">

nogramix:
@Tantacrul can we expect that the "Stemless"-option will also come for MS4? It was very useful in MS3.

Tantacrul:
It's added to 'release', which means it needs to be in for 4.0

cbjeukendrup:
@Tantacrul In MS3, we _did_ hide the flag, but _didn't_ hide the beam. Is it intended that this will be different in MS4? 

If yes, _how_ should we hide the beam? Should we hide the entire beam, or simply exclude the stemless chord from the beam so that the surrounding chords will still be beamed as appropriate? 

So, which of these options is preferable?
<img width="517" alt="Scherm­afbeelding 2022-11-19 om 15 28 20" src="https://user-images.githubusercontent.com/48658420/202855749-a20a1b7b-d6c5-4959-9dee-9ebfbe3ab3d8.png">

The latter makes the most sense to me, but I'm not quite familiar with the purpose of stemless notation, so I thought I'd just quickly ask.

Tantacrul:
I would have thought the MS3 way is the only way to go. @oktophonie?

cbjeukendrup:
Okay, that makes me quite happy 🙂. Hiding the whole beam is the current MS4 behaviour, but it is implemented in a somewhat hacky way at Inspector level (as opposed to Engraving level).

If we decide to replicate the MS3 behaviour, we're done after #14611 (please check it out!).

If we want to automatically exclude stemless chords from beams, we should implement this at Engraving level, as an addition to #14611. My proposed solution is to set the default beam mode to "no beam" for stemless chords; that's in fact how I created that example image. Perhaps this is something for the engraving team to tackle, ___if___ we're going to do it.

(I created that example image in MS3, which demonstrates that with MS3's behaviour every combination is possible.)

PR #14611, opened 2022-11-19
Inspector: Fix the "Stemless" property

cbjeukendrup:
It is a property of the chord rather than the stem, so we need a ChordSettingsModel.

(By the way, I think we'll need to move the (stem) direction property to this new model too, since direction is also relevant for chords without stem, and in fact it is also a chord property rather than a stem property. But that's a separate issue.)

Resolves: #10672


Tantacrul:
Just tested. This looks right to me! Also like the string change to stemless!

cbjeukendrup:
> Also like the string change to stemless!

The credits for this go to MuseScore 3 :)

oktophonie (approved):

RomanPudashkin (approved):

Issue #14620, opened 2022-11-20
[MU4 Issue] Crash when trying to add Composer text to 3.6.2 score via right click menu (one score)


DmitryArefiev:
[4791818.zip](https://github.com/musescore/MuseScore/files/10050472/4791818.zip)

https://user-images.githubusercontent.com/90187801/202920881-ad78c4bb-9216-47ee-bf2a-f01743f1f54a.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Doesn't occur in **MS3**

PR #14625, opened 2022-11-20
Fix crash when adding text where default font is musical font

cbjeukendrup:
If you had selected Leland (or another musical symbol font) as the default font for some text type in Format > Style > Text Styles, then a crash would occur upon adding a text element of that type.

Resolves: #14620

Alternative way to reproduce the crash:
1. Create a new score with Leland as the musical symbols font
2. In Format > Style > Text Styles, go to "Staff"
3. Change the default font family to "Leland" (not "Leland Text")
4. Press OK
5. Create a Staff Text by selecting a rest and pressing Cmd/Ctrl+T

### DIFF
--- a/src/engraving/libmscore/textbase.cpp
+++ b/src/engraving/libmscore/textbase.cpp
@@ -252,7 +252,9 @@ RectF TextCursor::cursorRect() const
     mu::draw::Font _font  = fragment ? fragment->font(_text) : _text->font();
     if (_font.family() == _text->score()->styleSt(Sid::MusicalSymbolFont)) {
         _font.setFamily(_text->score()->styleSt(Sid::MusicalTextFont), draw::Font::Type::MusicSymbolText);
-        _font.setPointSizeF(fragment->format.fontSize());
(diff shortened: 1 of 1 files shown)


src/engraving/libmscore/textbase.cpp
```diff
@@ -252,7 +252,9 @@ RectF TextCursor::cursorRect() const
     mu::draw::Font _font  = fragment ? fragment->font(_text) : _text->font();
     if (_font.family() == _text->score()->styleSt(Sid::MusicalSymbolFont)) {
         _font.setFamily(_text->score()->styleSt(Sid::MusicalTextFont), draw::Font::Type::MusicSymbolText);
-        _font.setPointSizeF(fragment->format.fontSize());
+        if (fragment) {
```
cbjeukendrup, line 255: As far as I know, it's a legitimate situation that `fragment` is null, namely when the text is empty because it's just created. So this fix is not simply hiding a deeper problem :)

RomanPudashkin (approved):

PR #14626, opened 2022-11-21
Fix #14404: size_t underflow error in cmdExplode

### LINKED ISSUE #14404: Musescore crash if you attempt to explode meaures containing chords with different numbers of notes.
**Describe the bug**
Tools/Explode crashes if you try to use it on measures that have chords with mismatching numbers of notes.

**To Reproduce**
Create a new score.
Add three staves to the score.
In measure 1 of the top staff, enter one chord with 3 notes and a second chord with 2 notes.
select the measure and extend your selection to include the two staves below it.
Choose Tools/Explode from the menu.
Musescore will crash.

**Expected behavior**
Musescore should be able to handle this situation gracefully and not crash.  It would be helpful if it could do its best to explode this measure, even though it has invalid data.  Eg, it could just enter a rest where the missing note is.

**Platform information**
Wundows 10

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/14404

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -3210,7 +3210,7 @@ void Score::cmdExplode()
                     // keep note "i" from top, which is backwards from nnotes - 1
                     // reuse notes if there are more instruments than notes
                     size_t stavesPerNote = std::max((lastStaff - srcStaff) / nnotes, static_cast<size_t>(1));
-                    size_t keepIndex = std::max(nnotes - 1 - (i / stavesPerNote), static_cast<size_t>(0));
+                    size_t keepIndex = static_cast<size_t>(std::max(static_cast<int>(nnotes) - 1 - static_cast<int>(i / stavesPerNote), 0));
                     Note* keepNote = c->notes()[keepIndex];
                     for (Note* n : notes) {
                         if (n != keepNote) {


src/engraving/libmscore/cmd.cpp
```diff
@@ -3210,7 +3210,7 @@ void Score::cmdExplode()
                     // keep note "i" from top, which is backwards from nnotes - 1
                     // reuse notes if there are more instruments than notes
                     size_t stavesPerNote = std::max((lastStaff - srcStaff) / nnotes, static_cast<size_t>(1));
-                    size_t keepIndex = std::max(nnotes - 1 - (i / stavesPerNote), static_cast<size_t>(0));
+                    size_t keepIndex = static_cast<size_t>(std::max(static_cast<int>(nnotes) - 1 - static_cast<int>(i / stavesPerNote), 0));
```
RomanPudashkin, line 3213: Too many static_cast... What exactly went wrong with this expression?

asattely: `nnotes - 1 - (i / stavesPerNote)` is `size_t`, which means that doing a subtraction and calling `std::max(x, 0)` doesn't help when x underflows to `size_t::MAX`. This was the cause of the crash. So I cast that as a signed integer so we can get negative numbers, and then cast the thing back to `size_t` when we are guaranteed a minimum value of zero.

RomanPudashkin: I see, now I understand the problem :) But shouldn't a similar change be applied to the line above?

asattely: `lastStaff` is never less than `srcStaff` (because of the `for` loop restrictions) and `nnotes` is unsigned, so underflow isn't going to happen on that line I think

RomanPudashkin (approved):

RomanPudashkin:
@asattely please open a PR with these changes for the 4.0_rc branch

Jojo-Schmitz:
Why not just use `git cherry-pick`?

Issue #14606, opened 2022-11-19
[MU4 Issue] Crashes when splitting a rest into shorter ones


SamYeh:
**Describe the bug**
MuseScore crashes when splitting a half rest into two quarter rests.

**To Reproduce**
1. Download [this score](https://musescore.com/user/55169571/scores/9017174). Its contents are show below.

![image](https://user-images.githubusercontent.com/10248807/202823040-178db7ff-a69a-4738-8d74-5619603cee53.png)

2. Click on the half rest in bar 2.
4. Press 5 on computer keyboard to split it into two quarter rests.
5. Observe that MuseScore crashes, terminates itself immediately.

These steps reproduce the bug 100% of the time on my computer.

**Expected behavior**
MuseScore should split the rest without crashing.

**Platform information**
 - OS: Windows 11
 - Build: MuseScoreNightly-223220403-master-fd02c8d-x86_64

oktophonie:
Duplicate of https://github.com/musescore/MuseScore/issues/14200 (but we'll use this example file as a test to confirm any fix)

PR #14635, opened 2022-11-21
Fix #14200 and #14606: nullcheck in findLinkedVoiceElement

### LINKED ISSUE #14200: [MU4 Issue] Score becomes unusable after a certain point, causing consistent crashes
**Describe the bug**
I was simply working on a score and then I find that whenever I press a specific set of numbers past a specific measure, the program crashes. I don't exactly know what is causing it to crash but when I press 5, 6, 7, 8, or 9 on a measure past the written measure 131, the crash happens.

**To Reproduce**
1. Open the score [Stone_and_Mortar ms4.zip](https://github.com/musescore/MuseScore/files/9897388/Stone_and_Mortar.ms4.zip)
2. Navigate to measure 131, the first measure on the 11th page
3. Click on a rest on any staff
4. Press the shortcuts 5, 6, 7, 8, or 9 (1, 2, 3, 4, and 0 are safe)
5. Program will crash

**Expected behavior**
The rest should've adjusted to the duration pressed, but instead the program crashes

**Platform information**
 - OS: Windows 11 64 bit, x64-based processor
 - Tested on both the beta in MuseHub and the recent nightly build

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/14200
Resolves: https://github.com/musescore/MuseScore/issues/14606

can't call `->element()` on nullptr!

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -4856,7 +4856,7 @@ static EngravingItem* findLinkedVoiceElement(EngravingItem* e, Staff* nstaff)
     Measure* measure = segment->measure();
     Measure* m       = score->tick2measure(measure->tick());
     Segment* s       = m->findSegment(segment->segmentType(), segment->tick());
-    return s->element(dtrack);
+    return s ? s->element(dtrack) : nullptr;
 }
 
 //---------------------------------------------------------


src/engraving/libmscore/edit.cpp
```diff
@@ -4856,7 +4856,7 @@ static EngravingItem* findLinkedVoiceElement(EngravingItem* e, Staff* nstaff)
     Measure* measure = segment->measure();
     Measure* m       = score->tick2measure(measure->tick());
     Segment* s       = m->findSegment(segment->segmentType(), segment->tick());
-    return s->element(dtrack);
```
RomanPudashkin, line 4859: Why is it null?

asattely: In the score in issue https://github.com/musescore/MuseScore/issues/14606, it's due to a different number of measures in the part vs. the masterscore, which I think probably shouldn't be possible? I've tried to recreate that score and I couldn't.

That makes me think the problem in https://github.com/musescore/MuseScore/issues/14200 is similar, though there are a lot more parts. I've asked for clarification from the submitter of that issue as to what steps were taken to see if I can narrow down a reason for that issue. I am running a debug build and #14200 is not crashing for me on master, but if a nullcheck here fixes it I can only assume there is a similar situation going on.

asattely: The level above has a nullcheck though: https://github.com/musescore/MuseScore/blob/13f3d886c65017e591a9e69f0a10e4f43b4b6dd7/src/engraving/libmscore/edit.cpp#L4938

Which means that we already expect that `findLinkedVoiceElement` can return `nullptr`.

PR #14638, opened 2022-11-21
[MU4] Fix #14597: Crash when appending measures to new score #14636

### LINKED ISSUE #14597: [MU4 Issue] Crash when appending measures to new score
**Describe the bug**
Appending measures with Ctrl+B immediately upon creating a score crashes

**To Reproduce**
1. Create a new score with treble clef template, all default settings
2. When score appears, press Ctrl+B
3. See error: crash

**Expected behavior**
Measure should be appended.

**Platform information**
 - OS: Linux

**Additional context**
For me, this is reproducible 90% of the time if I literally do the Ctrl+B first thing after creating the score.  I did wait for "loading audio samples" to complete (I have MS Basic as my default sounds).  But if I do anything else first, then Ctrl+B works.  This includes, simply scrolling.  Or just waiting an additional 5 seconds.  So I assume there is something still going on in the background during those first few seconds.

RomanPudashkin:
Resolves: #14597

### DIFF
--- a/src/engraving/libmscore/masterscore.cpp
+++ b/src/engraving/libmscore/masterscore.cpp
@@ -193,6 +193,12 @@ void MasterScore::updateRepeatListTempo()
     _repeatList2->updateTempo();
 }
 
+void MasterScore::updateRepeatList()
+{
+    _repeatList->update(MScore::playRepeats);
+    _repeatList2->update(false);
+}
+
 //---------------------------------------------------------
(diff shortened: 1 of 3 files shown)


DmitryArefiev:
Tested #14597 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Eism (approved):

Issue #14350, opened 2022-11-05
[MU4 Issue] Change Actual Measure Duration in a mmRest result in crash when revert it Ctrl+Z


Gai-Luron:
**Describe the bug**
Change Actual Measure Duration in a mmRest result in crash when revert it Alt-z

**To Reproduce**
1 - Create a Score as this with mmRest
![image](https://user-images.githubusercontent.com/1835871/200137020-2fa19987-56b4-4a0d-abef-dd530dda23e9.png)
2 - Open measure Properties of the mmRest and change actual measure duration
![image](https://user-images.githubusercontent.com/1835871/200137036-32dc52e3-12c5-4dae-a039-585a732b7c7d.png) 
Change to 
![image](https://user-images.githubusercontent.com/1835871/200137051-cbc59273-26fd-4532-8607-062e90f41826.png)
3 - Apply
![image](https://user-images.githubusercontent.com/1835871/200137062-01f12dca-3b82-4da1-958b-612dc87471f3.png)
4 - Ctrl-Z result in crash

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

cbjeukendrup:
Yep, reproduced. I think the easiest solution would be disallowing the Measure Properties dialog for mmrests, like we do apparently in MS3.

Gai-Luron:
You can't disable Measure Properties because we need it to access strech measure even in mmRest.

cbjeukendrup:
But it was disabled in MS3 too, right? Or was there another way of accessing it than via the context menu?

But of course, if someone manages to fix this and possible other bugs with Measure Properties and MMRests, I have no problem with allowing it.

In MS4, you can also access the stretch property in Properties > Appearance > Measure width. And you can use the keyboard shortcuts `{` and `}` of course.

Gai-Luron:
Why not disable only the Measure actual field? Anyway, use of arrows button to go next or previous measure skip MMRest and I think it's not the good behaviour. MMRest it's a measure like other. I have take a look at code of MMRest in the past and I know that's no easy to manage that.
Do that you think it's better. No problem for me. ;-)

MarcSabatella:
[...]

PR #14642, opened 2022-11-21
Fix MMRest crash when doing ctrl+Z

mike-spa:
Resolves: #14350

Just a missing null-check, it seems.

### DIFF
--- a/src/notation/internal/notationselectionrange.cpp
+++ b/src/notation/internal/notationselectionrange.cpp
@@ -217,7 +217,9 @@ const
     std::vector<RangeSection> sections;
 
     const mu::engraving::Segment* startSegment = rangeStartSegment;
-    for (const mu::engraving::Segment* segment = rangeStartSegment; segment && (segment != rangeEndSegment);) {
+    Fraction rangeEndTick = rangeEndSegment->tick();
+    for (const mu::engraving::Segment* segment = rangeStartSegment;
+         segment && segment != rangeEndSegment && segment->tick() <= rangeEndTick;) {
         mu::engraving::System* currentSegmentSystem = segment->measure()->system();
 
         mu::engraving::Segment* nextSegment = segment->next1MMenabled();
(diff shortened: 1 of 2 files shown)


oktophonie (approved):

src/notation/internal/notationselectionrange.cpp
```diff
@@ -221,7 +221,7 @@ const
         mu::engraving::System* currentSegmentSystem = segment->measure()->system();
 
         mu::engraving::Segment* nextSegment = segment->next1MMenabled();
-        mu::engraving::System* nextSegmentSystem = nextSegment->measure()->system();
+        mu::engraving::System* nextSegmentSystem = nextSegment ? nextSegment->measure()->system() : segment->measure()->system();
```
RomanPudashkin, line 224: Why is it null?

mike-spa: I just noticed that Segment::next1MM() (which is called by Segment::next1MMenabled()) can explicitely return a null, so its output should be null-checked. Why exactly it returns null in this case, I'm not sure, I can look into it

RomanPudashkin: >Why exactly it returns null in this case, I'm not sure

We should find out why it returns null

mike-spa: It seems that we were skipping the exit condition of the loop, so it was running until the end of the score (where a null return is normal). Now it should be good

RomanPudashkin: 1. You should create a temp variable for this expression outside the loop
`rangeEndSegment->tick()`

2. How is this change related to Ctrl+Z? I don't get it from the commit message

Also, please notice that there is some mess in the commit message

Screen Shot 2022-11-22 at 11 35 46

mike-spa: I think I was finally able to find the root problem: when doing ctrl+Z we were tying to reset the selection, but we were computing the wrong end-segment in `Selection::updateSelectedElements()`, which was messing up the rest. Now it's fixed (but I still changed the exit condition of the loop to make it safer).

RomanPudashkin (approved):

oktophonie (approved):

RomanPudashkin:
@mike-spa please open the same PR, but for the 4.0_rc branch

PR #14653, opened 2022-11-22
fixed #14489: Master Palette crashes MS4 due to application/paletteScale stuck on 0

### LINKED ISSUE #14489: [MU4 Issue] Master Palette crashes MS4 due to application/paletteScale stuck on 0
For working on a HiDPI device, I mistakenly set the `Preferences > Advanced > application/paletteScale` to "5", realised my mistake and then set it to "2". It worked OK and I continued my session. The next time I opened MS4, some of the Palettes' symbols were missing, showing only a horizontal line, and others garbled as it rendered much to large.

Examples:

Palette symbols missing Crazy large palette symbols

I tried to open the Master palette, and MS4 immediately crashed (closed/disappeared without warning message). After some troubleshooting, I realised that the issue is with the application/paletteScale setting. It is now stuck reverting back to "0" after each MS4 restart, and I have to change it back to 1 each time for the Master palette to function without crashing MS4. Even changing this back to "1" does not solve the standard palette issues, and the grid in the Master Palette is also somewhat screwed up since then...

Master Palette woes

When using the `Reset to default` button at the top of the Advanced page, it does the same as manually changing it to "1", so the outcome is the same as above. However, when clicking on the `Reset preferences` (for all) at the bottom, MS4 hangs indefinitely. 

Reset Preferences 

When closing the Preferences dialog (immediately or even after hours), Windows spits out the standard "not responding" window.

Not responding

**Steps to reproduce the behaviour:**
1. Go to `Preferences > Advanced` and change `application/paletteScale` value to "0". 
2. Click OK.
3. Close MS4, then reopen.
4. Open the Master palette.
5. MS4 crashes. If not, repeat steps 1 - 4, but change the value to "5" when at step 1.
6. When this crashes MS4, open it again, and view `application/paletteScale` to confirm that it has changed to "0" after MS4 restart.

**Expected behaviour**
The `application/paletteScale` should remain as the user have set it, not revert to "0". All palettes should have the correct size symbols and grids, not mismatched as in the screenshots above.

**Steps tried to fix the issue:**
1. I uninstalled MS4
2. Deleted the Program Files folder as well as the AppData folder
3. Ran CCleaner (Custom clean plus Registry clean)
4. Restarted Windows
5. Reran CCleaner as above
6. Reinstalled MS4

When opening MS4 after all this, it remembered my previous Scores, still showing them on the Home screen, and the issue described still remains. Thus there are still one or more locations where orphan settings are stored. As a temporary workaround for this issue I would like to completely purge MS4 from my system, removing all traces, then do a completely clean install. How would I go about doing this? 

**Platform information**
OS: Windows 10 (21H2 version19044.2251)

### **EDIT: WORKAROUND**
1. I opened the `Diagnostic paths` window from the menu `Diagnostic > System > Show paths...`
2. Took a screenshot
3. Repeated "Steps tried..." above, but made sure to purge all the folders in the Diagnositc paths window when at step 2.
4. Did not need to restart Windows... (it was overkill but I was deperate)
5. Reinstalled MS4 and got greeted by the first time setup, with all setting back to a fresh default.

This bug is still out there, but for now this workaround did the trick for me.

Eism:
Resolves: #14489

### DIFF
--- a/src/appshell/qml/DevTools/Preferences/SettingsPage.qml
+++ b/src/appshell/qml/DevTools/Preferences/SettingsPage.qml
@@ -109,10 +109,23 @@ ColumnLayout {
 
                 Loader {
                     id: loader
-                    property var val: valRole
+                    property var val: valueRole
+                    property var minValue: minValueRole
+                    property var maxValue: maxValueRole
                     anchors.fill: parent
                     sourceComponent: root.componentByType(typeRole)
-                    onLoaded: loader.item.val = loader.val
+                    onLoaded: {
+                        loader.item.val = loader.val
+
+                        if (loader.item.minValue !== undefined) {
+                            loader.item.minValue = loader.minValue
+                        }
+
+                        if (loader.item.maxValue !== undefined) {
+                            loader.item.maxValue = loader.maxValue
+                        }
+                    }
+
                     onValChanged: {
                         if (loader.item) {
                             loader.item.val = loader.val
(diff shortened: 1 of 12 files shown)


DmitryArefiev:
Tested #14489 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #14557, opened 2022-11-16
[MU4 Issue] Size adjustments to chord symbols lost on save / reload


MarcSabatella:
**Describe the bug**
If you change the size of a single chord symbol using Properties, that change is lost when you save the file

**To Reproduce**
1. New score
2. Add chord symbols to the first two measures
3. Change the second to 18 pt font using the Properties panel
4. Save
5. Close
6. Reload
7. See error: chord symbol has reverted to the default size

**Expected behavior**
Size change should persist

**Screenshots**
Expected:
![image](https://user-images.githubusercontent.com/1799936/202282375-07836148-99e4-4d1a-8fa5-6ee400f8f2bc.png)

Actual:
![image](https://user-images.githubusercontent.com/1799936/202282297-37fd0d96-f2bb-4a67-9e08-bf0048e95a61.png)

**Platform information**
 - OS: Linux

**Additional context**
From the mscx file, I can see the size info was never written to the file (whereas it is for staff text).  The same issue occurs whether you change the overall size by single-clicking the chord symbol, or the size of all the characters within by double-clicking.

Issue #12756, opened 2022-08-10
[MU4 Issue] Custom formatting of text objects lost in copy and paste operations


rgreen5:
**Describe the bug**

When a text object is copied and pasted, and it contains a SPACE with a custom font-size, the space disappears! Same thing happens if you save and reopen the file without copying the object

**To Reproduce**

1. Open a score
2. Click on a note say, then press Ctrl + T to add staff text.
3. Enter the characters "Greater than", SPACE, "Less than" (i.e. "< >")
4. Copy the text object to another fretmark, say.
RESULT: All font properties are preserved as expected.
5. Select the space in the text object and increase its font-size to 30, say. Then copy the object again.
RESULT: The space has disappeared.

**Expected behavior**

All font-properties, whether object-level or character-level, should be preserved during cut/copy/paste operations (and any other operations such as saving to a palette, or saving the file).

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2829968415, revision: github-musescore-musescore-2d21253

rgreen5:
**Additional issue**
There is also an allied problem with lost formatting when copying text objects in _imported MS3 files_.

1. Open the attached MS3 file in MS4. [harmonics.zip](https://github.com/musescore/MuseScore/files/9315524/harmonics.zip)
2. Copy and paste a harmonic _symbol_ (not the fretmark) from one of the existing fretmarks in measure 1 onto a _corresponding_ fretmark in measure 2.
RESULT: Custom formatting is lost during the operation and the harmonic symbol no longer fits the fretmark.

**Expected result**

All style and character formatting should be preserved during operations.

oktophonie:
Something else I have found triggers a reset of custom formatting: resetting (unrelated) styles to default. See video. I have, however, not been able to reproduce this consistently.

https://user-images.githubusercontent.com/20604032/190986882-ef84a406-ccc7-41f6-8143-d4c8c1482fbf.mp4

bkunda:
I can add to this issue that custom offsets for page numbers are reset to 0 when a file is closed and re-opened.

MarcSabatella:
[...]

PR #14663, opened 2022-11-22
Text formatting bug fix

mike-spa:
Resolves: #12756 (not the original space-character problem, which is secondary, but the copy-paste problem, and especially the unintentional style reset problem that @oktophonie reported there, which is extremely dangerous as it can literally cancel hours of work).
Also resolves: #14557 

The origin of the problem is in `TextBase::writeProperties()`. At the moment, we are deliberately _not_ writing some of the font properties to the xml. The change was made in [#8815](https://github.com/musescore/MuseScore/pull/8815) with the purpose that now font properties are not written to the xml at an object level but at a character level. I don't know if someone can comment on that particular choice @RomanPudashkin @cbjeukendrup, but I'm afraid it has to be reverted, at least for now, because it causes a lot of problems:

- By not writing the object property, we mess up the `PropertyFlags` logic. Specifically, it will not be set to UNSTYLED, which means it will not be written (also during copy-paste) and it will be reset by `styleChanged()` even when it shouldn't (this was the cause of #12756).
- Some text-based items, such as Harmony, do not write any actual string to the xml, so they can't write the property at the character-level. If we don't write the property at an object level, it will not be written at all (this was the cause of #14557).

This PR reverts that change and adds some checks to make sure we don't write to the xml invalid tags, which causes other layout problems on save-reload or copy-paste.

Still, we are left with the problem that character-based formatting is possible, and in that case the xml tag is only present at a character level. Again, we need to make sure that in such cases the relevant propetyFlags are also marked as UNSTYLED, otherwise the formatting could be unintentionally lost upon copy-pasting or style resets. For this purpose I've introduce TextBase::checkCustomFormatting(), which simply looks into the string to see if formatting tags are present. I suspect there are smarter ways to do that, but that's the solution I've found atm. Let me know.

This one will need careful testing.


oktophonie (approved):

RomanPudashkin (approved):

RomanPudashkin:
@mike-spa please rebase it

mike-spa:
@vpereverzev done!

PR #14684, opened 2022-11-23
[MU4] parallel_audio_rendering

vpereverzev:


vpereverzev:
@Tantacrul @DmitryArefiev 
please, check these builds out

- WIN - https://github.com/musescore/MuseScore/suites/9466105241/artifacts/448274202
- MAC - https://github.com/musescore/MuseScore/suites/9466101733/artifacts/448288963

RomanPudashkin (approved):

src/framework/global/concurrency/taskscheduler.h
```diff
@@ -0,0 +1,211 @@
...
+            return 1;
+        }
+
+        thread_pool_size_t optimalCapacity = maxCapacity / 2;
+
+        if (desiredThreadCount <= 0) {
+            return optimalCapacity;
+        }
+
+        return desiredThreadCount;
```
shoogle, line 166: This is *way* outside my area of expertise, but should there be a check like this here?
```c++
if (desiredThreadCount > optimalCapacity) {
    return optimalCapacity;
}
```
Or perhaps this?
```c++
if (desiredThreadCount > maxCapacity) {
    return maxCapacity;
}
```

vpereverzev: nope, number of allocated threads not necessarily should be <= of logical/physical cores. It might be deliberate choice of a consumer

src/framework/audio/internal/worker/mixer.cpp
```diff
@@ -128,12 +130,30 @@ samples_t Mixer::process(float* outBuffer, samples_t samplesPerChannel)
...
-        std::fill(m_writeCacheBuff.begin(), m_writeCacheBuff.end(), 0.f);
+    std::vector<std::future<std::vector<float> > > futureList;
+
+    for (size_t i = 0; i < m_mixerChannels.size(); ++i) {
+        MixerChannelPtr channel = m_mixerChannels.at(i);
+
+        std::future<std::vector<float> > future = TaskScheduler::instance()->submit([this, samplesPerChannel,
+                                                                                     channel]() -> std::vector<float> {
+            thread_local std::vector<float> buffer(samplesPerChannel * audioChannelsCount(), 0.f);
+            thread_local std::vector<float> silent_buffer(samplesPerChannel * audioChannelsCount(), 0.f);
```
cbjeukendrup, line 141: This might be a dumb question but: as far as I understand, `thread_local` is similar to `static`, but then for the lifetime of a thread instead of the whole program. Doesn't that mean that it can cause problems to initialise it based on `samplesPerChannel`, which is variable instead of static? 

For example, it may happen that this code runs a second time in the same thread (due to TaskScheduler), is that correct? 

If that happens, and `samplesPerChannel` is different during the second call, then the size of these buffers is _not_ changed, right? (because they are already initialised and never updated again).

Isn't that a problem? `channel->process(…)` will be thinking that it is writing to a buffer of the size based on _new_ `samplesPerChannel`, but in reality it is writing to a buffer of the size based on _old_ `samplesPerChannel`. If the expected/new size is bigger than the real/old size, I'd think there will be a buffer overflow.

vpereverzev: Driver may request variable number of samples, but when it comes to rendering we should keep samplesPerChannel consistent. See audioconfiguration::renderingStep

cbjeukendrup: Ah, I see. So in practice, `samplesPerChannel` and `audioChannelsCount()` are constants. And now I also understand that we _write_ to the buffer in steps of a fixed size, and the driver can _read_ as many as it wants each time. Nice to know!

Tantacrul:
We have been testing this build for 2 days now across MacOS and PC. Apart from one export crash (which was then fixed), we noticed:

- A very large increase in exporting speed for larger scores
- A much better playback experience that removed all stuttering, distortion & clicks (on the machines we tested)
- In all cases, the kept the buffer size at 1024 only

We tested each of the audio export formats at the most important export settings:
- .Wav - 32K, 44.1K, 48K
- .MP3 - 44.1K, 48K (@192 & 128 bitrate)
- .OGG - 44.1K, 48K
- .Flac - 44.1K, 48K

We also tested uploads to MuseScore.com that included MS Basic, Muse Sounds and VSTi.

I uploaded a full orchestral score to MuseScore.com that was using Ozone 9 in the Master Channel: https://musescore.com/user/31604498/scores/8726328

Considering the success of all the above, we think now is the time to merge.

@abariska @bkunda @RomanPudashkin @DmitryArefiev @shoogle @jessjwilliamson @oktophonie

Issue #14570, opened 2022-11-17
[MU4 Issue] A4 scores created in MU3 get imported as Letter


MarcSabatella:
**Describe the bug**
Scores created in MU3 that use the default page size of A4 get converted to Letter on import into MU4

**To Reproduce**
1. In MU3, create a score that uses the default page size of A4
2. Load that score into MU4
3. Go to Format / Page Settings
4. See error: page size is now Letter

**Expected behavior**
Page size should remain A4

**Platform information**
 - OS: Linux

**Additional context**
The problem is that the [legacy 302 style defaults file](https://github.com/musescore/MuseScore/blob/master/src/engraving/data/styles/legacy-style-defaults-v302.mss) is wrong - it has letter as the page size.  The height/width/printablewidth settings need to be fixed.

Note the problem is only visible for scores that truly use the built-in default A4 settings.  If your system currently defaults to Letter in MU3 (because of the printer check, or because you have a custom default style set in preferences) and manually set the page size to A4 - or if you change the page size to something else then back to A4 - then the problem might not occur.  That's because MU3 might see the page size as not being exactly equal to A4 due to round-off, and thus, the modified page size is written to the file.  Whereas on systems where A4 is the true default and you leave the page size alone, the page size does not normally get written.

Tantacrul:
@oktophonie

Tantacrul:
OK, good to know!

cbjeukendrup:
@MarcSabatella If you have a potential solution, feel free to make a PR for it. I don't have much time for coding the next few days. Also, as a small separate PR, it may be merged sooner than as part of a large-ish PR.

MarcSabatella:
PR submitted, see note in PR comment about how to test.

MarcSabatella:
I had forgotten about this after https://github.com/musescore/MuseScore/pull/14561 was merged, but this bug still exists and was just reported by a user today.  I believe my PR is still valid.

Jojo-Schmitz:
Yet another showstopper (for me), fix should really make it into 4.0.2

PR #14686, opened 2022-11-23
Fix page size on 302 import

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/14570

The legacy 302 style file has the wrong default page size (Letter instead of A4), causing scores using the A4 default to change to Letter on import.

### DIFF
--- a/src/engraving/data/styles/legacy-style-defaults-v302.mss
+++ b/src/engraving/data/styles/legacy-style-defaults-v302.mss
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <museScore version="4.00">
   <Style>
-    <pageWidth>8.5</pageWidth>
-    <pageHeight>11</pageHeight>
+    <pageWidth>8.26772</pageWidth>
+    <pageHeight>11.6929</pageHeight>
     <pagePrintableWidth>7.08661</pagePrintableWidth>
     <pageEvenLeftMargin>0.590551</pageEvenLeftMargin>
     <pageOddLeftMargin>0.590551</pageOddLeftMargin>


MarcSabatella:
As far as I know the pagePrintableWidth is correct as is.  This would need to be tested on a system for which A4 actually is the default.  In MU3, create a score, check the page size and margins (should be A4 and 15 mm), save it, then open in MU4 and verify the page and margins are correct.  There have often been small discrepancies in margin size due to round-off error (eg, 15 mm changing to 15.06 mm), but I *think* this will avoid that.

Tantacrul:
@oktophonie - can you do the honors?

oktophonie:
Having tried this, a migrated score has the correct page size and margins, but I also tried it on a recent nightly build and it worked fine there too. So doesn't this really need to be tested on a system where Letter is the default?

Jojo-Schmitz:
Well, testing on a system with A4 being the default is a sanity test, whether that fix broke anything there.

MarcSabatella:
Hmm, it *shouldn't* work correctly in a nightly on A4-default systems.  But, that could be the rounding error I mentioned actually works around this problem on some systems, by forcing MuseScore to write the page size to the file even though it's ostensibly A4 (just not *exactly* equal to what the default A4 size is).

But, if you have a 3.6.2 file that truly uses the default - where the page size was *not* written to the file - it will import into current 4.0 builds as Letter.  That's the bug I'm fixing here.

So even if, due to the rounding error, scores you create from scratch on your system have the page size written to the file, plenty of *other* A4 scores don't have their page sizes written to the file.  Here's a random example:

https://musescore.com/user/26911315/scores/9086213

Load this into 3.6.2 and you will see it's A4.  Load it into a current 4.0 build and you will see it's Letter (and with badly messed up margins).  This PR fixes that and this should be testable on any system.

cbjeukendrup (approved):
I tested this with a MS3 file where the page size was not written to the file, and indeed this file uses the incorrect size on the latest master branch, and the correct size with this PR. And since the default page format was A4 in MS3, it would make sense if it's also A4 in our "MS3 defaults" file. So I think this is ready for merging!

(waiting until tomorrow in case anyone disagrees)

bkunda:
🟢 Re-tested in 4.0.2. Remains FIXED.

Issue #13926, opened 2022-10-23
[MU4 Issue] Musescore UI not in System Default language


teetow:
**Describe the bug**
Musescore shows the incorrect language by default, not respecting its own setting "System Default."

**To Reproduce**
1. Go to Edit > Preferences
2. Verify "Language" set to "System Default" (English in my case)
3. Note that the Musecore UI is in Swedish, which is not my selected Display Language (see screenshots)
4. Silently wonder why every single multiplatform app gets this wrong

**Expected behavior**
Musescore UI being in English, as per its Language setting

**Screenshots**

Windows Display Language setting

![image](https://user-images.githubusercontent.com/183582/197399876-fa149eff-80b8-4f96-b31a-0fafb6bb22b3.png)

**Platform information**
 - OS: Windows 10

**Additional context**
As mentioned, this is a common screwup in applications where devs mistakenly interpret User Regional formatting settings (stuff like date and time formatting) to mean System Language. Please be one of the good guys here.

cbjeukendrup:
I'm afraid this might be a Qt problem. We get the system language from Qt, using [`QLocale::system().uiLanguages()`](https://doc.qt.io/qt-5/qlocale.html#uiLanguages). 
Although Qt intends that this is the correct way to retrieve the display language, we might try another way.

teetow:
Hmm. I whipped up a quick test app (in PyQt5, granted) and it gave me the following values for the settings I could scrounge up from various SO answers:

![image](https://user-images.githubusercontent.com/183582/197405163-b8c85897-ebf7-46b0-9425-586d3344c717.png)

So, the only place where Swedish shows up is indeed in `QLocale::system()::uiLanguages()`. Now, I can't really make heads or tails of all the stuff in `languageService.cpp` but I'm seeing references to `languageCode` which makes me believe that (at least on my machine) the _last_ code in that list gets picked as the assumed system language.

Good news is, you have two other methods that (again, on my machine) yields English -- `QLocale.system().language()` and `QLocale.system().name()`. So maybe there's some hope.

Tantacrul:
@cbjeukendrup - can you prioritise (4.x, post-beta, release etc.)? If we think it's an easy, quick fix, then release would be appropriate.

cbjeukendrup:
@teetow There's a small chance that this is fixed by #14701. Could you please try the builds from that PR?

teetow:
[...]

PR #14701, opened 2022-11-23
LanguagesService: redirect "en" language code to "en_US"

cbjeukendrup:
If the system language code happens to be something like "en_CY", that will decay to "en". But we don't have that, only "en_US". That causes some problems. To fix that, we map "en" to "en_US".

Might resolve: #13926 
May resolve: https://github.com/musescore/MuseScore/pull/12917#issuecomment-1323621684

### DIFF
--- a/src/languages/internal/languagesservice.cpp
+++ b/src/languages/internal/languagesservice.cpp
@@ -215,6 +215,18 @@ QString LanguagesService::effectiveLanguageCode(const QString& languageCode) con
                 return code;
             }
 
+            static const std::map<QString, QString> SHORT_TO_LONG_FALLBACK_CODES {
+                { "en", "en_US" },
+                { "hi", "hi_IN" },
+                { "mn", "mn_MN" },
+                { "zh", "zh_CN" }
+            };
+
+            auto it = SHORT_TO_LONG_FALLBACK_CODES.find(code);
+            if (it != SHORT_TO_LONG_FALLBACK_CODES.cend()) {
+                return it->second;
+            }
+
             int rightmost = code.lastIndexOf('_');
             if (rightmost <= 0) {
                 break;


cbjeukendrup:
@Eism Could you please try if this solves the issue for you? 

There are multiple ways to solve this; I'm now doing it without touching the languages hash, but alternatively we could make sure that we put "en" in the hash instead of "en_US", by making a special case in the `languages.json` reading code. We can't edit `languages.json` directly though, because the language codes need to match the codes from Transifex, and there we use "en" as the "source language" and "en_US" as one of the "target languages", so we can't make any changes there.

Eism:
@cbjeukendrup 
There are other problematic languages in the list of available languages, for example zh (there are zh_CN, zh_HK, zh_TW, but no zh).

Maybe instead of 
```
if (m_languagesHash.contains(code)) {
    return code;
}
```
should be
```
std::sort(languages.begin(), languages.end(), [](const Language& l, const Language& r) {
        return l.code < r.code;
});

for (const Language& language : languages) {
    if (language.code.startsWith(code)) {
        return language.code;
    }
}
```

The first found in the list will be selected, for now it is en_GB. (Or does the language have to be US?)

cbjeukendrup:
Yes, the language should preferably be en_US if no other variant is specified explicitly. 

Anyway, I will try to come up with a more general mechanism to work around these problems.

src/languages/internal/languagesservice.cpp
```diff
@@ -215,6 +215,22 @@ QString LanguagesService::effectiveLanguageCode(const QString& languageCode) con
                 return code;
             }
 
+            // For some languages, we only have long, specific codes in LanguagesHash / languages.json.
+            // If the user's system language is just "en", or "en_CY" for example, this will not be found.
+            // So, for short codes that are at risk, we map them to long codes that we do have.
+            // Should be kept in sync with languages.json.
```
Eism, line 221: I think this comment is redundant. A lot of terms from the code that can change

src/languages/internal/languagesservice.cpp
```diff
@@ -215,6 +215,22 @@ QString LanguagesService::effectiveLanguageCode(const QString& languageCode) con
...
+            static const std::map<QString, QString> SHORT_TO_LONG_CODES {
```
Eism, line 222: I think we can move it to configuration
Like fallbacksForLanguages

cbjeukendrup: I prefer not moving it to configuration, because this relates quite closely to this `tryCode` algorithm, and actually all logic about which languages we support happens in LanguagesService. I hope this is okay :)

Eism (approved):

Eism:
should be merged to RC

PR #14723, opened 2022-11-24
primary_secondary_instrument

RomanPudashkin:


vpereverzev (approved):

src/engraving/playback/mapping/stringssetupdataresolver.cpp
```diff
@@ -270,5 +270,17 @@ const PlaybackSetupData& StringsSetupDataResolver::doResolve(const Instrument* i
         return empty;
     }
 
+    static const std::unordered_set<SoundId> supportPrimaryAndSecondaryCategories {
+        SoundId::Violoncello,
+        SoundId::VioloncelloSection,
```
cbjeukendrup, line 275: Shouldn't this be `Violin` instead of `Violoncello`? (at least, in Muse Sounds, Violin does support 1 & 2, Violoncello doesn't.)

PR #14734, opened 2022-11-25
Fix: Beam types -> Beam type

Gootector:
The first argument: should be "Beam type" because we select beam type, not types. The second argument: "types" occurs only once in strings to translate - only in this case.

Greetings,
Gootector


shoogle (approved):
Looks good! Here's where the change occurs:
image

bkunda:
This makes perfect sense to me. Well argued!

cbjeukendrup (approved):

PR #14747, opened 2022-11-25
[MU4 RC] Add new instrument: Bell Tree

shoogle:
Port PR #11787 for the MU4 Release Candidate (`4.0_rc` branch).

PR #14670, opened 2022-11-22
Fix chord symbol style file filter

MarcSabatella:
Resolves: https://github.com/musescore/MuseScore/issues/14002

The button to load chord symbol style files uses the wrong filter (*.mss instead of *.xml).  This means that the dialog shows the wrong files - it shows "regular" style files rather than chord symbol description files.

- [x] I signed [CLA](https://musescore.org/en/cla)
- [x] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] I made sure the code compiles on my machine
- [x] I made sure there are no unnecessary changes in the code
- [x] I made sure the title of the PR reflects the core meaning of the issue you are solving
- [x] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
- [x] I made sure the commit message title starts with "fix #424242:" if there is a related issue
- [] I created the test (mtest, vtest, script test) to verify the changes I made

PR #14761, opened 2022-11-25
[MU4 RC] Cherry-picking two trivial string change PRs

cbjeukendrup:
Includes: 
#14750
#14734

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -1153,7 +1153,7 @@
             </Instrument>
             <Instrument id="g-soprano-ocarina">
                   <family>ocarinas</family>
-                  <trackName>G Soprano Ocarina</trackName>
+                  <trackName>Soprano Ocarina</trackName>
                   <longName>G Soprano Ocarina</longName>
                   <shortName>G S. Oc.</shortName>
                   <traitName type="tuning">G</traitName>
@@ -1171,7 +1171,7 @@
             </Instrument>
             <Instrument id="f-soprano-ocarina">
                   <family>ocarinas</family>
-                  <trackName>F Soprano Ocarina</trackName>
+                  <trackName>Soprano Ocarina</trackName>
                   <longName>F Soprano Ocarina</longName>
                   <shortName>F S. Oc.</shortName>
                   <traitName type="tuning">F</traitName>
--- a/share/instruments/instrumentsxml.h
+++ b/share/instruments/instrumentsxml.h
@@ -596,26 +596,26 @@ QT_TRANSLATE_NOOP3("engraving/instruments", "Contrabass Recorder", "contrabass-r
 //: shortName for Contrabass Recorder; Please see https://github.com/musescore/MuseScore/wiki/Translating-instrument-names
 QT_TRANSLATE_NOOP3("engraving/instruments", "Cb. Rec.", "contrabass-recorder shortName"),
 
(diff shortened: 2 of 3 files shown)


shoogle:
I think it would be better to cherry-pick the actual code commits (as was done in PR #14749 and PR #14747) rather than the merge commits. If you cherry-pick a merge commit then we will probably end up with yet another merge commit.

Gootector:
Let's see - for example: the branch 4.0_alpha is 21 commits ahead and 3660 commits master. So, 21 commits from alpha aren't in the master branch.
I think that MuseScore repo should take example from Audacity repo:
The master is always switched to the next big release. After the release of big release, PRs for patch release are sending into the new branch. With the release of patch release, patch branch is merged into the master. The master contains all commits from the other branches.
I hope I helped.

cbjeukendrup:
@shoogle Done! Indeed a better idea. Also, it seems that cherry-picking a merge commit turns it into a "normal" commit, containing all the changes of the merged commits together.

@Gootector Yes, that's another possible strategy. The big advantage is that the "… commits ahead" message will say something useful: currently, we "cherry-pick" all commits to the RC branch, which means that it becomes technically a different commit, so after cherry-picking one commit, GitHub will say: "1 commit ahead and 1 commit behind". I think that's what happened with the alpha branch: those "ahead" commits are commits that were cherry-picked from the master branch. On the other hand, if we would create the commits on the RC branch, and then merge the RC branch into master, the commits will stay the same, but "glued" to the master branch by a merge commit. That also brings us to the disadvantage: those merge commits create a very non-linear commit history, making it slightly more difficult to track down where a certain bug was introduced. But maybe that's only a small problem.
Anyway, maybe something to try out next time :)

shoogle:
@Gootector
> 21 commits from alpha aren't in the master branch.

When commits are cherry-picked between branches, the SHA changes, so Git and Github no longer identify them as being the same even though the code itself has not changed. In reality, all code on the `4.0alpha` branch also made it to `master`.

Audacity's method is worth investigating, but it has some downsides of its own. Firstly the non-linear history that @cbjeukendrup mentioned, but also...

> With the release of patch release, patch branch is merged into the master.

This means that useful fixes on the patch branch are not available on the master branch until after the patch release has been made.

I'd say the big advantage of Audacity's method is not having to cherry-pick each and every commit individually. However, it puts the burden of rebasing everything all on one person, and rebasing gets harder the longer it is delayed. Eventually some of those SHAs probably wouldn't match even with Audacity's method.

cbjeukendrup:
(Of course, it's also possible to merge the patch branch every few days, or even every few hours. That solves the problems of having to merge everything at once, and of not having important fixes in master for a while.)

shoogle:
True. Regular merges would probably result in fewer changed SHAs as well (once one SHA changes all subsequent SHAs under the same merge commit must change too, but SHAs in future merges can still match). However, the more often you do it the less work you save compared to cherry-picking.

It would also make the history even more non-linear (lots of small merges vs. one big one). I'm not sure whether that would make it easier or harder to follow.

MarcSabatella:
FWIW, I'm in favor of anything that makes it easy for folks like me to be able to submit occasional small PR's and not have to do them each twice.  I still can't tell what if anything I would need to do to make sure https://github.com/musescore/MuseScore/pull/14669 or https://github.com/musescore/MuseScore/pull/14670 make it into the release.  They were both merged into master but perhaps not rc.

Issue #13968, opened 2022-10-24
[MU4 Issue] Work title gets overridden by "Untitled Score" on musescore.com in Save/Publish flows


bkunda:
**Describe the bug**
The work title for both **privately saved** _and_ **published** scores on musescore.com changes to "Untitled Score" when a file is closed and reopened in the desktop app. 

This can be rectified by manually entering a title in Project Properties > Work title, which leads me to suspect that an aspect of this bug could be that entering a title in the "Save to cloud" dialog doesn't populate this field (as it perhaps should 🤷‍♂️). 

**To Reproduce**
See demo video: https://www.dropbox.com/s/swivp5oy0h0ezod/untitled-score-issue-privatecloud.mov?dl=0

1. Create a new score
2. Save the score to the cloud and give it a unique filename
3. Open the score online: **work title is correct**
4. Quit and restart MuseScore desktop app
5. Save the score and open it online: **work title shows as "Untitled Score"**
6. In the desktop app, open Project Properties > Work title, and give your score a unique name in "Work title"
7. Save the score and open it online: work title shows as "Untitled Score"
8. Quit and restart MuseScore desktop app
9. Save the score and open it online: **work title is correct**

**Expected behavior**
The main point is that the work title online should never revert back to "Untitled Score" when restarting the desktop app. It should reflect what the user entered when they saved/published their score to the cloud.

Purely by way of suggestion: it would make sense to me if the "Work title" field in Project Properties would show the title entered by the user in the "Save to cloud" dialog when the score is saved for the first time. This would then give the user a mechanism to change the work title if they want at a later date.
(As a corollary to this: if the user were to change the work title online, it should also update in this field in the desktop app). 

**Platform information**
 - MacOS 12.6 (M1)

FYI @Tantacrul

PR #14784, opened 2022-11-27
Fix name when saving project to cloud after re-opening

cbjeukendrup:
Resolves: #13968

When reopening a project, the cloud name for that project is unknown. When re-saving the project, we used the work title as a fallback, but this should be the filename.

Also, when the score is renamed on MuseScore.com, we respect this during saving, so that we don't override it back to the old name.

### DIFF
--- a/src/project/internal/notationproject.cpp
+++ b/src/project/internal/notationproject.cpp
@@ -408,9 +408,8 @@ bool NotationProject::isCloudProject() const
 const CloudProjectInfo& NotationProject::cloudInfo() const
 {
     if (!m_cloudInfo.isValid()) {
-        auto tags = m_masterNotation->masterScore()->metaTags();
-        m_cloudInfo.name = tags[WORK_TITLE_TAG].toQString();
-        m_cloudInfo.sourceUrl = tags[SOURCE_TAG].toQString();
+        m_cloudInfo.name = io::filename(m_path, false).toQString();
+        m_cloudInfo.sourceUrl = m_masterNotation->masterScore()->metaTags()[SOURCE_TAG].toQString();
     }
 
     return m_cloudInfo;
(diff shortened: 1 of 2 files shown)


MarcSabatella:
For some of my scores, even though I have set a reasonable workTitle, every time I re-publish to musescore.com, it keeps reverting to the filename.  I had never figured out why this only happened for some scores and not others, and thus I never opened an issue for it as I have no idea how to reproduce the problem.  But based on your second commit here, I am hopeful that it will turn out to be this.  Should I expect a build using your PR to fix this for existing scores?

FWIW, one such score is https://musescore.com/user/2975/scores/8880078, which has a valid workTitle but keeps reverting to the filename as a title every time I publish.

RomanPudashkin (approved):

cbjeukendrup:
@MarcSabatella This PR is about re-saving rather than re-publishing. With this PR, when simply pressing Ctrl+S on an existing cloud score, the name on MuseScore.com will be respected, instead of reverting to the workTitle or filename. 

Maybe it's worth trying the builds, to see if this improves the situation for you.

MarcSabatella:
Unfortunately, no change.

I'm still not clear is on this: is it by design that every time I try to re-publish a score already published, it defaults to the filename instead of using the workTitle or the name previously set on msuescore.com?  And that the rights field always reverts to "None" even if I've previously set it to "Attribution"?  I gather it *is* by design that unlisted scores turn into private every time you re-publish.  These are all significant regressions IMHO.

cbjeukendrup:
@MarcSabatella This seems to be a separate problem then; I think it would be better to create a new issue about it. I assume that with "it defaults to the filename" you mean the defaults in the "Publish" dialog?

MarcSabatella:
Correct.  Will open the new issue.

Tantacrul:
I've tested this and it's working.

Issue #14495, opened 2022-11-12
[MU4 Issue] "sfp" doesn't play p after repetition.


LuisLiszt:
**Describe the bug**
A sfp dynamic plays normally but after a repetition it will play f instead of p after the dynamic. See video...

**To Reproduce**
1. Recreate score in the video.

**Expected behavior**
Sfp should play p after the dynamic.

**Video**

https://user-images.githubusercontent.com/55204261/201479048-b1ef1fdd-a27d-41d5-a72c-bc0d5e85e5f9.mp4

**Platform information**
 - OS: Windows

**Additional context**
Probably related to #13543

DmitryArefiev:
Fixed in #14788

Issue #13543, opened 2022-09-30
[MU4 Issue] fp dynamic plays as f (no drop to p)


MarcSabatella:
**Describe the bug**
The "fp" dynamic (and similar) does not play correctly - you get the "f" but not the change to "p" after.

**To Reproduce**
1. Score for trumpet
2. Add a whole note
3. Add an "fp" dynamic to it
4. Play
5. Hear error: volume is "f" throughout

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/1799936/193162583-50f5d71a-f5b6-45e2-8870-e063b2c174ae.png)

**Platform information**
 - OS: Linux

**Additional context**
This is one issue I hadn't caught yet using my MuseScore Cafe theme score - https://musescore.com/user/2975/scores/6656770.  That's the MuseScore 3 version; I'm working on a MuseScore 4 update to take advantage of things like the fall playback.

MarcSabatella:
I don't know for sure if it was ever working in MU4, but I tend to doubt it.  I check nightly builds on this score on a pretty regular basis, but until recently there were enough other problems that this one could easily have escaped notice.  Oldest build I have handy to test is the alpha 2 from early August, and I can confirm it was a problem then.  Sorry for not catching it sooner, but dynamics were a complete mess (mostly because of all the articulation weirdnesses), so it was very easy to miss.

FWIW, the three scores I'm always checking are this one ("MuseScore Café" theme), Reunion, and the one I sent you to check with Muse Sounds - the "Music Master Class" theme.  All three sounded pretty bad just a couple of months ago, but gradually almost everything has been fixed.   At this point, Music Master Class is fine (at least with MS Basic and VST instruments - not sure if Muse Sounds will still show the odd portamento effect we heard a few weeks ago).  Reunion is good except for pedal playback.  And the MuseScore Café is good except for this "fp" issue plus falls/glissandi on tied notes - https://github.com/musescore/MuseScore/issues/12971.  Well, also https://github.com/musescore/MuseScore/issues/13544 but that is understandably being put off, and I'm now working around that using "doits" which are close enough.

cbjeukendrup:
@RomanPudashkin Are you working on this currently? I have an idea for a solution, based on 3.x, which I could have ready today.

DmitryArefiev:
Fixed in #14788

PR #14788, opened 2022-11-27
Fixes for "fp"-like and "sf"-like dynamics

cbjeukendrup:
Resolves: #13543
Resolves: #14495


MarcSabatella:
Unfortunately this doens't seem to work if there is a crescendo starting on that note (very common in conjunction with "fp"):

https://user-images.githubusercontent.com/1799936/204159817-5127a0ef-a937-45f9-b8d0-fc3fd9da0d38.mp4

I'd also say, it sounds pretty artifical even without the fp, at least in this example (MS Basic, trumpet).

cbjeukendrup:
@MarcSabatella I made two improvements:
- the transition from forte to piano is now smooth (instead of first forte, then piano, with nothing in-between), which should make it more natural
- the basic case of fp + hairpin is fixed, albeit in a somewhat hacky way. I'm not sure if this will be approved. At the moment, working with the combination of hairpins and dynamic markings is a bit cumbersome. Hopefully this will be improved in the future.

MarcSabatella:
Your hack looks straightforward enough to me, hopefully it's considered OK!

DmitryArefiev:
Tested #13543 #14495 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #13903, opened 2022-10-23
[MU4 Issue] Holding down the mixer button causes Musescore to crash


erinic04:
**Describe the bug**
Holding the mixer button for about a second on larger scores causes Musescore to crash

**To Reproduce**
1. Open a larger score
2. Hold the mixer button (F10 by default)
3. Observe the crash

**Expected behavior**
Musescore shouldn't crash

**Screenshots**

**Platform information**
 - OS: [e.g. macOS, Windows, Linux]

**Additional context**

Tantacrul:
What OS are you using? I can't replicate it with the shortcut or the UI.

erinic04:
I'm using Windows. Also, it needs to be a larger score for it to work. I did it reproducibly on a 7 minute marching band full score (with Muse Sounds if that makes a difference)

cbjeukendrup:
@erinic04 It seems that I can only reproduce this issue during playback. Is that the case for you too?

erinic04:
It wasn't the case before, not sure if it is now

cbjeukendrup:
@erinic04 Could you please try if this problem is fixed for you in the builds from #14790? (see https://github.com/musescore/MuseScore/wiki/Downloading-and-running-test-builds#downloading-builds-from-pull-requests for how to download it)

erinic04:
Yeah it's fixed for me (whether holding the button while the score is playing or not)

erinic04:
I tested it on a full orchestra score that's 600 measures long so we good

zacjansheski:
@cbjeukendrup should this still have the label "needs info"?

I have reproduced in the nightly, but it takes quite a bit and I found it it easier to reproduce when the mixer is docked.

https://github.com/musescore/MuseScore/assets/69917893/fccf8d26-468b-4aa0-95e3-be591f60a136
[13903_ZJ.zip](https://github.com/musescore/MuseScore/files/11637064/13903_ZJ.zip)

[Mvt 4_Beethoven__Symphony_No._9__Op._125.zip](https://github.com/musescore/MuseScore/files/11637093/Mvt.4_Beethoven__Symphony_No._9__Op._125.zip)

cbjeukendrup:
I think we have enough info indeed :)

I checked it again, and found indeed that the same crash still happens and the same fix still works.

Issue #9467, opened 2021-10-14
[MU4 Issue] Crash after openning-closing Mixer multiple times during playback (on big score)


DmitryArefiev:
**To Reproduce**
**Case 1**:
1. Open [Big_Score.zip](https://github.com/musescore/MuseScore/files/7346223/Big_Score.zip)
2. Open Mixer
3. Start Playback
4. Close-Open Mixer multiple times (see video below)

UPDATE (23 Mar 2022): Now it is harder to reproduce the crash: you need to close-open Mixer ~20 times instead of ~5.
Reducing priority.

**Case 2** (fixed in #14790, can be tested on Mac for now):
1. Open [Big_Score.zip](https://github.com/musescore/MuseScore/files/7346223/Big_Score.zip)
2. Open Mixer
3. Start Playback
4. Close-Open Mixer multiple times 
5. Stop Playback
6. Close score (not MS app)

![crash playback+mixer](https://user-images.githubusercontent.com/90187801/137324206-67f6b511-49c8-4462-bc30-753d209f5820.gif)

**Desktop:**
 - OS: Win10/Mac11
 (didn't try on Linux)

**Additional context**
Sometimes can happen after first closing of Mixer (on big scores)

Tantacrul:
I have tested big score (which is a masterpiece, by the way) and rapidly opened and closed the mixer about 100 times and not experienced one crash.

Perhaps time to close the issue?

DmitryArefiev:
I've tried on Win and Mac and **still have the crash on Mac** (M1 laptop)

https://user-images.githubusercontent.com/90187801/166913203-dda55222-dddd-4b7b-99a7-1d4a56b1b792.mov

Maybe let's just low the priority to P3 and move to Beta 2? (since it's not regular workflow to open Mixer several times so quickly)

AntonioBL:
Ciao.
When running under AddressSanitizer, just after closing the Mixer there is a heap-use-after-free error: something (the playback?) is trying to access a `MixerChannelItem` which was deleted when closing the Mixer.
Here is the backtrace of the log (note the `freed by thread T0` and `previously allocated by thread T0` sections):
```
=================================================================
==13688==ERROR: AddressSanitizer: heap-use-after-free on address 0x6130008c4490 at pc 0x55556629ded5 bp 0x7fffffffc760 sp 0x7fffffffc750
READ of size 8 at 0x6130008c4490 thread T0
[...]

PR #14790, opened 2022-11-27
Fix crash when opening/closing mixer very quickly during playback

cbjeukendrup:
Resolves: #13903
Resolves: #9467

What happened: 

0. The Mixer is subscribed to the audio signal changes, and does this from the main thread.
1. In the audio thread, in `AudioSignalsNotifier::updateSignalValues(…)`, we call `audioSignalChanges.send(audioChNumber, signalVal);`.
2. Because the mixer is subscribed from the main thread but "we" are on the audio thread, this results in the allocation of a `QInvoker` in `AbstractInvoker::invoke(int type, const NotifyData& data)`. Also, a call to this `QInvoker`'s `invoke()` method, followed by the deletion of this `QInvoker` is queued to `QueuedInvoker`.
3. In the constructor of this `QInvoker`, it adds itself to the `AbstractInvoker`'s list of `m_qInvokers` by calling `AbstractInvoker::addQInvoker(…)`.
---
4. Then, for some reason, `AbstractInvoker::removeCallBack(…)` is called, which "invalidates" the `QInvoker`; that means: the `QInvoker`'s pointer to the `AbstractInvoker` is set to `nullptr`. 
---
5. Then, the `QInvoker`'s `invoke()` method is still called (because that had been queued already, see point 2). This doesn't do much :)
6. And, the `QInvoker` is deleted, because that was also queued. 
7. Normally, the destructor would remove the `QInvoker` from the `AbstractInvoker`'s list of `m_qInvokers` by calling `AbstractInvoker::removeQInvoker(…)`. But in this case, the `QInvoker` was already invalidated, so it has no reference to the `AbstractInvoker` anymore, so it _doesn't_ remove itself from the list of `m_qInvokers`. So that list now contains a pointer to deleted memory! ⚠️⚠️⚠️
---
8. Later, `AbstractInvoker::removeCallBack(…)` is called again, for some callback (which one exactly isn't relevant).
9. This method loops through the list of `m_qInvokers`, and reads a bit of all of them, to look whether they belong to the callback that is being removed (and thus need to be invalidated). 
10. And thus it also reads from the deleted `QInvoker` that was still in the list! 💥

### DIFF
--- a/src/framework/global/thirdparty/kors_async/async/internal/abstractinvoker.cpp
+++ b/src/framework/global/thirdparty/kors_async/async/internal/abstractinvoker.cpp
@@ -39,6 +39,8 @@ AbstractInvoker::~AbstractInvoker()
     for (QInvoker* qi : m_qInvokers) {
         qi->invalidate();
     }
+
+    m_qInvokers.clear();
 }
 
 void AbstractInvoker::invoke(int type)
@@ -147,9 +149,11 @@ void AbstractInvoker::removeCallBack(int type, Asyncable* receiver)
 
     {
         std::lock_guard<std::mutex> lock(m_qInvokersMutex);
-        for (QInvoker* qi : m_qInvokers) {
+        for (auto it = m_qInvokers.begin(); it != m_qInvokers.end(); ++it) {
+            QInvoker* qi = *it;
             if (qi->call.call == c.call) {
                 qi->invalidate();
+                m_qInvokers.erase(it);
                 break;
             }
         }


igorkorsukov (approved):

src/framework/global/thirdparty/deto_async/async/internal/abstractinvoker.cpp
```diff
@@ -12,7 +12,7 @@ AbstractInvoker::AbstractInvoker()
 
 AbstractInvoker::~AbstractInvoker()
 {
-    std::lock_guard<std::mutex> lock(m_qInvokersMutex);
+    std::lock_guard lock(m_qInvokersMutex);
     for (QInvoker* qi : m_qInvokers) {
         qi->invalidate();
```
cbjeukendrup, line 17: Wait, it's still not good: now I'm removing things from `m_qInvokers` while iterating over it! Will fix.

cbjeukendrup:
Rebased again :)

DmitryArefiev:
@cbjeukendrup Tested on Win/Mac with [Mvt.4_Beethoven__Symphony_No._9__Op._125.zip](https://github.com/musescore/MuseScore/files/14210783/Mvt.4_Beethoven__Symphony_No._9__Op._125.zip)


on **Win11**, after some iterations, **Mixer stops to open**:


https://github.com/musescore/MuseScore/assets/90187801/2eba8bad-103f-4f8e-adb6-2afd075c4696



on **Mac13.6**, it **still crashes**:

https://github.com/musescore/MuseScore/assets/90187801/2dae88fe-db75-4d6a-9556-0cc94de242af

cbjeukendrup:
I've looked into this PR again. My conclusion is that this PR does solve the first problem that causes a crash (and potentially some crashes-on-quit), so it would be good to merge it.
But I also observe that there are at least two other problems, which is why there may still be crashes after this PR. One of them is a very obscure crash in QtQuick and/or KDDockWidgets (probably, that is the one seen by @DmitryArefiev). The other can probably only be reproduced with Address Sanitizer enabled and is a data race in AbstractInvoker. 
About the issue seen on Windows, mentioned above: that looks like a QtQuick and/or KDDockWidgets bug, maybe related to the crash seen on Mac. It is unlikely that it is a deadlock, and even more unlikely that it is caused by this PR. 

Note that the issue is slightly more difficult to reproduce now that long-pressing F10 does no longer repeatedly fire the Toggle Mixer keyboard shortcut, but it can still be reproduced by just pressing the F10 key very fast manually.

I've looked into fixing the data race in AbstractInvoker, but that turns out to be complicated, and also dangerous because it is very easy to introduce deadlocks or performance issues. I will discuss this with @igorkorsukov and solve it in a different PR, if we decide that it's worth the effort and risk. 

@RomanPudashkin @DmitryArefiev Do you think it's okay to merge this PR in its current state?

RomanPudashkin (approved):

igorkorsukov (approved):

DmitryArefiev:
Tested #9467 on Win10, Mac13.6. There is another crash on Windows - #23505. But on **Mac** it looks fine, especially **Case 2** from  #9467 (comparing with master)

Issue #14787, opened 2022-11-27
[MU4 Issue] Finale Maestro / Broadway: repeat dots appear too low


cbjeukendrup:
Expected behaviour (Leland and all other fonts):
<img width="108" alt="Scherm­afbeelding 2022-11-27 om 15 22 52" src="https://user-images.githubusercontent.com/48658420/204140259-99ddd6dc-fb6f-4949-bd4b-d02cb7d43686.png">

Actual behaviour with Finale Maestro and Finale Broadway:
<img width="108" alt="Scherm­afbeelding 2022-11-27 om 15 28 51" src="https://user-images.githubusercontent.com/48658420/204140548-03d05d4e-83af-4f6c-b413-66271fa850c2.png">

This is obviously an issue with these fonts rather than with MuseScore, but we might want to add a workaround for the time being. 

(That brings us to another question: how do we know when there are updates available for the Finale fonts? Do we know where to download updated versions of these fonts?)

Jojo-Schmitz:
Somhwo sounds familiar.
See https://github.com/musescore/MuseScore/blob/fab9db01fa9e52f39ae74600dd5a412f1b950860/src/engraving/libmscore/barline.cpp#L505-L512

Jojo-Schmitz:
IHMO the contrary is the case: "Emmentaler, Gonville, MuseJazz: repeat dots appear too high" with a workaround for that being in place, but by accident (ommission) also affecting Finale Maestro / Broadway

Tantacrul:
@oktophonie - which set of fonts is at fault here? 😄

I'm going to make an uninformed observation that if you have two camps:

1. Gonville, MuseJazz and Emmentaler 
2. Bravura, Leland and Petaluma

Group 1. is more likely to be wrong! I say this having had to make a ton of changes to Emmentaler last year so it corresponded with the SMuFL spec properly. It'll be just my luck now to discover I'm wrong.

cbjeukendrup:
@Tantacrul See https://github.com/musescore/MuseScore/pull/14797; Jojo has created a fix already

Jojo-Schmitz:
Add Finale Maestro and Finale Broadway to the 2nd camp, makes it a 3 : 5 majority vote ;-)
In the beginning (MuseScore 2!) It was 2 : 1 (Emmentaler, Gonville vs. Bravura).

oktophonie:
Bravura is, as always, the model. 
[...]

PR #14797, opened 2022-11-28
[MU4] Fix GH#14787: Finale Maestro / Broadway: repeat dots appear too low

Jojo-Schmitz:
Resolves: #14787

### DIFF
--- a/src/engraving/libmscore/barline.cpp
+++ b/src/engraving/libmscore/barline.cpp
@@ -502,10 +502,10 @@ void BarLine::drawDots(Painter* painter, double x) const
         y1l = st->doty1() * _spatium;
         y2l = st->doty2() * _spatium;
 
-        //workaround to make Bravura, Petaluma and Leland font work correctly with repeatDots
-        if (!(score()->engravingFont()->name() == "Leland"
-              || score()->engravingFont()->name() == "Bravura"
-              || score()->engravingFont()->name() == "Petaluma")) {
+        //workaround to make Emmentaler, Gonville and MuseJazz font work correctly with repeatDots
+        if (score()->engravingFont()->name() == "Emmentaler"
+            || score()->engravingFont()->name() == "Gonville"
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Might be easier and shorter to check for Emmentaler, Gonville and MuseJazz?
Or, as these are 'our' fonts', to fix those?

Jojo-Schmitz:
That alternative fix would turn the issue into (a workaround for) "Emmentaler, Gonville, MuseJazz: repeat dots appear too high"

cbjeukendrup:
> Or, as these are 'our' fonts', to fix those?

@oktophonie What do you think about this? Can we reasonably easily fix those fonts themselves?

cbjeukendrup:
I like the latter!

Tantacrul:
I can fix Emmentaler, Gonville and MuseJazz. I may not be able to do it for a few days so perhaps we may need this hack instead.

Jojo-Schmitz:
That hack is in already, I just corrected it to also work for Finale Maestro and Finale Broadway, by simply inverting it.

cbjeukendrup:
@Tantacrul @oktophonie What do you think about merging this for now, and then fixing "our" three fonts later when we have more time, so that we can remove the workaround by that time?

Jojo-Schmitz:
That's indeed the whole and only purpose of this PR ;-)

Jojo-Schmitz:
Came up again in https://musescore.org/en/node/343759
Can we have this for 4.0.2 too, please?

cbjeukendrup:
I've nominated the linked issue for 4.0.2.

Jojo-Schmitz:
Ah, the issue, nit the PR, I see now, thanks and sorry for the noise

Jojo-Schmitz:
Any update on fixing the `repeatDot` position in Emmentaler, Gonville, MuseJazz, so we can get rid of the hacky workaround from this PR here?
See also #27960...

PR #14798, opened 2022-11-28
fixed signed integer overflow detected by sanitizer

handrok:


src/framework/global/thirdparty/haw_logger/logger/logger.cpp
```diff
@@ -61,7 +61,7 @@ DateTime DateTime::now()
     dt.time.hour = tm->tm_hour;
     dt.time.min = tm->tm_min;
     dt.time.sec = tm->tm_sec;
-    dt.time.msec = static_cast<int>(ms_d.count() - (sec * 1000));
+    dt.time.msec = static_cast<int>(ms_d.count() - (static_cast<long long>(sec) * 1000));
```
Jojo-Schmitz, line 64: Whoudn't `dt.time.msec = static_cast<int>(ms_d.count() - (sec * 1000LL));` be better? As the latter is the culprit for the potential overflow.

handrok: done

igorkorsukov (approved):

Issue #13344, opened 2022-09-14
[MU4 Issue] Editing dynamics is broken


wizofaus:
**Describe the bug**
Attempting to edit the text inside a dynamic element often returns in incorrect fonts being used or text disappearing entirely

**To Reproduce**
1. Create a new score
2. Add an ***f*** dynamic
3. Triple click to select the element and enter edit mode (or just enter edit mode and select the "f" - you can also use various other methods, as long as the text is selected while in edit mode)
4. Type p
5. A p in a strange font shows
6. Click away or press ESC to end editing mode
7. The dynamic disappears completely

**Expected behavior**
Text for dynamics should be editable as expected.  Note that pressing 'p' *isn't* actually expected to show a ***p*** dynamic marking - that requires using ctrl+shift+p. But it should show a regular p in the default font.

**Screenshots**
N/A

**Platform information**
Windows 10

**Additional context**
Latest nightly build, Sep-14-2022.
NB similar issues happen for other text elements after inserting musical symbols, though don't seem quite as bad.
I'd put a heap of work into this about a year ago and it was all working :(

MarcSabatella:
I see a similar effect - I'm guessing the same thing - editing tempo markings, and that's a more critical problem I think, because editing tempo is incredibly common.  In particular:

1) Alt+Shift+T to add a tempo marking
2) Ctrl+A to select the default "(quarter) = 80"
3) type something like "tempo primo"

Same issue: the font seems random (presumably as a result of the special character at the beginning of the text), and it's likely to disappear upon completion of editing - when you press Esc.  But worse, the tempo marking is there, it's not visible.  You can find it only via navigation by Alt+Left/Right.  So you're going to have a lot of people with random invisible 80 BPM tempo markings all over their scores until they figure the workaround.  Which is to not do the Ctrl+A to select all, but to instead first type the new text, then delete the old.

rgreen5:
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-3224155784, revision: github-musescore-musescore-c6f3e75

Same issue with tempo text.

wizofaus:
Confirmed it's still happening with latest build, and as Marc notes, it's pretty serious as it actually applies to *any* text element (that happens to contain musical symbols) and the invisible elements left over can wreak havoc.

Issue #12978, opened 2022-08-23
[MU4 Issue] Tempo text acting weird after modification


cgpqci:
**Describe the bug**
Tempo text, specifically the one with note+metronome value (e.g. "♩ = 80"), changes font and disappears after modifying in a certain way.

**To Reproduce**
1. Insert tempo text (note+metronome value)

![image](https://user-images.githubusercontent.com/90231404/186110933-bd37c13b-0b0d-4c5c-98d3-59877eb2a247.png)

2. Select whole text through shortcut <kbd>Ctrl</kbd>+<kbd>A</kbd>.

![image](https://user-images.githubusercontent.com/90231404/186111310-3aaf10d1-eced-4130-a4cf-9ce7c279cf2b.png)

3. Clear with <kbd>Backspace</kbd> key. (_optional_)
4. Enter new text.

![image](https://user-images.githubusercontent.com/90231404/186111567-23e7a43d-ef23-452f-b3e2-81f6dbd2c7f8.png)

5. Exit Edit mode (_click outside the text box_).

![image](https://user-images.githubusercontent.com/90231404/186111949-45603f06-2826-41f2-a220-6586943f5ea2.png)
(_The text disappears._)

**Expected behavior**
The text should remain consistent; no such bug should occur.

**Screenshots**
See above

**Platform information**
 - OS: Windows 10
 - Version: MuseScoreNightly-202208230452-master-64960b1-x86_64

**Additional context**
This also happens on the previous nightly builds, including the released Alpha builds.

PR #14808, opened 2022-11-28
Fix broken editing dynamics when selection starts on music symbol

mike-spa:
Resolves: #13344
Resolves: #12978

The problem arises because, when doing that particular editing operation, we set the cursor to "ScoreText" font, which is used in `TextFragment::font()` to return either the music-font (say, Leland) or the music-text-font (say, Leland Text) family. The problem is that neither of those contain alphabet characters, so the UI will show a random fallback font and the charactes typed won't be registered. 
I _think_ that the system is supposed to work with the cursor always set to the text font. If you then enter special symbols, they will be registered as separate TextFragments and will be set to the music-font or music-text-font as needed. In principle, at least as far as I understand, it shouldn't be necessary to set the cursor to the music fonts. @MarcSabatella @wizofaus this seems to solve the issue, but do let me know if you find it causes problems elsewhere that I'm not seeing.

### DIFF
--- a/src/engraving/libmscore/textedit.cpp
+++ b/src/engraving/libmscore/textedit.cpp
@@ -981,8 +981,6 @@ bool TextBase::deleteSelectedText(EditData& ed)
                 break;
             }
             TextCursor undoCursor(*_cursor);
-            // can't rely on the cursor's current format as it doesn't preserve the special font "ScoreText"
-            undoCursor.setFormat(*_layout[_cursor->row()].formatAt(static_cast<int>(_cursor->column())));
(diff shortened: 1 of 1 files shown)


oktophonie (approved):

wizofaus:
The bug didn't seem to be related to select-all though? What happens now if you undo deleting a range of text?

mike-spa:
Undo works fine

https://user-images.githubusercontent.com/93707756/204480717-8c8972ac-2793-434d-aba2-474262d3e010.mp4

RomanPudashkin (approved):

mike-spa:
Done! 👍

Issue #14789, opened 2022-11-27
[MU4 Issue] Added sharp to the chord signature ?!


wyldcrd:
**Describe the bug**
Cannot use the shortcut **CTRL + #** (Option + #) to place a # symbol in a chord signature. This is a regression on older versions of MS4.
(https://musescore.org/en/node/337234)

**To Reproduce**
1. Option + # when adding an accidental to a chord 

**Expected behavior**
To be able to add these to chords:
<img width="304" alt="image" src="https://user-images.githubusercontent.com/21022311/204341132-e684a332-2d22-456b-9d4d-a87811e0713f.png">

**Platform information**
 - OS: macOS Monterey 

**Additional context**
https://musescore.org/en/node/337234
(Reposted for Claudio Riffero)

Tantacrul:
Urg. I noticed this over the weekend. @cbjeukendrup - if you are looking for something urgent in Release, this is a really bad one.

cbjeukendrup:
@Tantacrul I was a bit confused at first, because on a Dutch keyboard the only key that requires the Option key is the `€` character. But apparently on other keyboards even the `#` needs this. 

Anyway, PR created: https://github.com/musescore/MuseScore/pull/14812

I could obviously only use the `€` character to test, but this must also fix it for the `#` character on British keyboards.

PR #14812, opened 2022-11-28
macOS: fix typing Option-key characters in Chord Symbols, Figured Bass, Fingering, Lyrics and Sticking

cbjeukendrup:
Resolves: #14789

For certain keyboard layouts on macOS, some characters can only be typed by pressing the Option (Alt) key. For example, On British keyboards, you have this:
Scherm­afbeelding 2022-11-28 om 19 21 35 (image source: https://support.apple.com/nl-nl/HT201794)
To type the `#` character, you need to press <kbd>Option</kbd> + <kbd>3</kbd>.

This was broken in MuseScore, because we block <kbd>Option</kbd> + <kbd>[almost any key]</kbd> because that would be a "navigation key" (I'm not completely sure what that means). But I think it's pretty obvious that <kbd>Option</kbd> + <kbd>3</kbd> (and similar) are not navigation keys, so we should not block those. Now we only block specific keys as being "navigation keys", whatever that means.

Because I'm so unsure what a "navigation key" is supposed to mean, this will require a thorough review and testing :)

### DIFF
--- a/src/engraving/libmscore/textbase.h
+++ b/src/engraving/libmscore/textbase.h
@@ -505,12 +505,16 @@ public:
 inline bool isTextNavigationKey(int key, KeyboardModifiers modifiers)
 {
     if (modifiers & TextEditingControlModifier) {
-        static const std::set<int> standardTextOperationsKeys {
-            Key_Space, // Ctrl + Space inserts the space symbol
-            Key_A // select all
+        static const std::set<int> controlNavigationKeys {
+            Key_Left,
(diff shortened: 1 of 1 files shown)


Tantacrul:
Works great @cbjeukendrup ! Thanks a lot :)

Tantacrul:
@RomanPudashkin @Eism - this one looks good to me.

RomanPudashkin (approved):

Issue #14716, opened 2022-11-24
[MU4 Issue] Sound lags behind cursor position during playback


rgreen5:
**Describe the bug**

Just noticed on this particular nightly (though it may have been present but unnoticed before).

Play back the attached (short) score, with MS Basic, and listen out for the lag between the cursor position and sound.
[time_lag.zip](https://github.com/musescore/MuseScore/files/10082925/time_lag.zip)

RESULT: Sound seems to be lagging perhaps an eighth to a quarter note behind the cursor position

**Expected behavior**

During playback, the cursor position should coincide with the sound heard.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223280405, revision: github-musescore-musescore-2756533

Grillzombie:
All ok with Linux Mint 21
OS: Linux Mint 21, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223240408, revision: github-musescore-musescore-56b4855

erinic04:
Is your buffer size larger? That's what that does lol

DmitryArefiev:
Yes, I've noticed that too. Some recent changes


https://user-images.githubusercontent.com/90187801/203858295-cb601dcd-b0d5-427c-9af1-d4ac7fafe5c4.mp4
[piano.zip](https://github.com/musescore/MuseScore/files/10087085/piano.zip)

PR #14829, opened 2022-11-29
[MU4] buffer_capacity

vpereverzev:
Resolves: #14716 

Forgot to bring back original value after some experiments


vpereverzev:
tested by @Tantacrul

PR #14834, opened 2022-11-29
Fix #14807: Hide Count-in option

### LINKED ISSUE #14807: [TASK] Since we are not implementing 'Count In' in MS4.0, we need to hide the UI for it
**Describe the bug**
This option needs to be hidden until we have the 'Count-in' feature re-implemented in 4.X

image

**Expected behavior**
Option is not visible

HemantAntony:
Resolves: #14807

### DIFF
--- a/src/playback/internal/playbackuiactions.cpp
+++ b/src/playback/internal/playbackuiactions.cpp
@@ -108,14 +108,14 @@ const UiActionList PlaybackUiActions::m_settingsActions = {
              IconCode::Code::PAN_SCORE,
              Checkable::Yes
              ),
-    UiAction("countin",
-             mu::context::UiCtxAny,
-             mu::context::CTX_ANY,
-             TranslatableString("action", "Enable count-in when playing"),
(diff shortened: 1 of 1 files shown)


src/playback/internal/playbackuiactions.cpp
```diff
@@ -108,14 +108,14 @@ const UiActionList PlaybackUiActions::m_settingsActions = {
...
              ),
-    UiAction("countin",
-             mu::context::UiCtxAny,
-             mu::context::CTX_ANY,
-             TranslatableString("action", "Enable count-in when playing"),
-             TranslatableString("action", "Enable count-in when playing"),
-             IconCode::Code::COUNT_IN,
-             Checkable::Yes
-             ),
+//    UiAction("countin",                                      // See #14807
```
RomanPudashkin, line 111: You should keep only this change. All changes in other files should be excluded from this PR

Tantacrul:
@RomanPudashkin - as long as you're happy, this one does what it says on the tin 😄 

image

@HemantAntony thanks so much!

Issue #13027, opened 2022-08-26
[MU4 Issue] Slide play property causes timing errors in the score


rgreen5:
**Describe the bug**

In the attached score, there is a playback timing problem in all the measures with a slide in them.

**To Reproduce**

* Open the attached file and play it back. 
[slide_playback_bug.zip](https://github.com/musescore/MuseScore/files/9432912/slide_playback_bug.zip)

RESULT: In meaure 1, for example, the quarter note (A#) in voice 1, beat 3, does not start on the 3rd beat but an 1/8th note late. Removing the slide (on the preceding G) restores correct playback timing.

![measure_screenshot](https://user-images.githubusercontent.com/87246583/186900402-9991a664-fd98-4ee8-b7d6-14bdfcd4a762.png)

**Expected behavior**

Playback should accurately mirror what's shown in the score.

**Platform information**
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2931312884, revision: github-musescore-musescore-c4515b8

Tantacrul:
I'm not actually experiencing this on my build. I wonder whether it has been fixed? @rgreen5 - can you test again?

Tantacrul:
Oh wait, I hear it now.

i slowed it down to **half note = 30** and switched on the metronome. It sounded fine.
I then sped it up to **half note = 50* and can confirm the A# now plays late.

<img width="686" alt="image" src="https://user-images.githubusercontent.com/21022311/188951360-39268124-d719-49d7-a670-89e484e0bf75.png">

rgreen5:
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-3289917644, revision: github-musescore-musescore-1ac395a

Another test case here, but this time the slide is causing the notes in mm 1 and 3 to sound too early:

![slide_playback_issue](https://user-images.githubusercontent.com/87246583/197384737-14239691-4e0e-4e01-9c1a-001f255b9631.png)

[slide_playback_issue.zip](https://github.com/musescore/MuseScore/files/9846130/slide_playback_issue.zip)

PR #14843, opened 2022-11-30
[MU4] fluid_synth_pitch_bend_intervals

vpereverzev:
Resolves: #13027

Turned out that FluidSynth can't carry this amount of messages on lower sampler rates, which produces huge delays after pitch bend messages being sent


RomanPudashkin (approved):

Tantacrul:
Happy with this one. Rhythm problem is fixed.

Issue #13877, opened 2022-10-22
Single Note Dynamics should be disabled for Piano


ShawTakeuchi:
**Describe the bug**

Single Note Dynamic (SND) appear to be used for piano.  This is true for MS Basic, VST but not Muse Sounds.

Single note dynamic (SND) should be "off" for percussive instruments such as Piano or Drums.  The "hairpin" should *not* affect the sound volume on the sustained notes.

**To Reproduce**
See image:
<img width="529" alt="image" src="https://user-images.githubusercontent.com/21022311/197384788-80162012-349e-44fe-ade6-c0910d0699e2.png">

For the release, single note dynamics should be switched off for:

- Piano
- Celesta
- Vibraphone
- Harpsichord
- Marimba
- Harp
- Guitars
- All other instruments in the 'plucked strings' category
- All percussion instruments

Tantacrul:
Thanks for pointing this out. I have updated the issue to target a fix for this release. I do not think SND should be on for any instrument that can't support it naturally. Perhaps later, we might provide a way for users activate them but I see that as an ultra-niche use case.

ShawTakeuchi:
As of MS 4.0.2 / MuseHub refreshed on 2023.03.15.

*  This issue seems to be fixed for MuseHub Piano!

*  It is not yet fixed for 

1)  MS Basic 
2)  3rd party Soundfont (tested with MS3 default soundfont port by Mark Sabatella)
3)  3rd party VSTs (tested with EastWest Opus Steinway)

I created a mscv test file with various use-cases if helpful - See attached.

cc: @MarcSabatella 

[Tester-20230315d.zip](https://github.com/musescore/MuseScore/files/10983598/Tester-20230315d.zip)

rgreen5:
OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230651545, revision: github-musescore-musescore-dbe7c6d / MS Basic

I can confirm that _**pf**_ _**fp**_ dynamics are still switched on for classical guitar;, though not working correctly as the last dynamic of the pair persists beyond the next note.

DaddyLudwig:
[...]

PR #14848, opened 2022-11-30
[MU4] filtered_gradual_volume_change

vpereverzev:
Resolves: #13877


vpereverzev:
Closing that one. Unfortunately, the proper fix would be quite risky

Issue #13435, opened 2022-09-21
[MU4 Task] Checking update on Linux


DmitryArefiev:
**Task description**
It is already implemented on Mac and Windows but still required additional work on **Linux**

**Linux**

![Screenshot from 2022-09-21 16-06-23](https://user-images.githubusercontent.com/90187801/191513113-f428a243-c060-49cc-8ae4-9c90a7086fe4.png)

**Windows**

![Capture](https://user-images.githubusercontent.com/90187801/191513269-3f451ded-6036-4c14-bc8f-e86f89b9b4f7.PNG)

DmitryArefiev:
@Tantacrul On Linux it just shows progress window and downloads the app file to Downloads folder. And there is no need to close the current because it doesn't replace initial file. 

How it looks in Audacity:

![01](https://user-images.githubusercontent.com/90187801/191788663-987f36f4-42de-482d-b66f-6526ac8e0e11.png)
![02](https://user-images.githubusercontent.com/90187801/191788671-511bd123-7dd4-4c2a-8bc2-d736759eec0e.png)
![03](https://user-images.githubusercontent.com/90187801/191788675-bc7369d1-a5ce-4a27-980b-9f60dd907a50.png)

But there is no any message after download progress window (Picture 2) in what location file was downloaded.

What about to add a message something like: "The newest version of MuseScore has been successfully downloaded to your Downloads folder" ?

Tantacrul:
Moving to post beta. However, yes, I agree.

We should revamp the update checking / new update implementation once MS4.0 is out. I would like it to behave like MuseScore 3 in future.

DmitryArefiev:
Done in #14850

PR #14850, opened 2022-11-30
[MU4] Checking update on Linux

Eism:
Resolves: #13435


DmitryArefiev:
Tested #13435 on LinuxUbuntu 22.04, LinuxMint 22.01 - **FIXED**

Issue #13853, opened 2022-10-20
[MU4 Issue] Playback curtailed on unison between voices


la-bel:
**Describe the bug**
When two voices are playing the same note for different durations, the longer playback is cut short to the length of the shorter playback.

**To Reproduce**
1. Create a new score with a sustaining instrument (like a saxophone)
2. Add a whole note in voice 1
3. Add a much shorter note in voice 2 starting on the same beat and playing the same note
4. Play the score - the longer note does not sustain. 

**Expected behavior**

The longer note should play for its written length.

**Screenshots**
<img width="964" alt="Screenshot 2022-10-20 at 21 43 07" src="https://user-images.githubusercontent.com/63920211/197054011-f2159fad-0ce6-44a7-abd9-8fcbca58bb4d.png">
Example

**Platform information**
 - OS: macOS

MarcSabatella:
Not good indeed!  This was an ongoing issue with lots of debate over the years in previous versions of MuseScore as to what the desired behavior actually is with respect to the MIDI playback, and what the best way to achieve that behavior is.  The current implementation in MuseScore 3 is the best compromise people were able to come up with, and probably makes a good model.  See https://musescore.org/en/node/12971

la-bel:
Thanks for the context! I couldn't think of ways to phrase it to see if there had been previous discussion.

MarcSabatella:
@Tantacrul this is very serious

Tantacrul:
Seems like something that can easily be solved now that we're using a new system.

Desired behaviour: the long and short duration in both voices should both be heard to their full extent.

maestroharmony:
@Tantacrul sorry the ping but wouldn't this trigger Phasing issue like the same sample? I always thought that the reason this couldn't be achievable is because It's the same sample being triggered. I would suggest WHEN this happens, there would be a system that could trigger samples from a half/whole step down to use a different sample, otherwise it will become out-of-phase.

Tantacrul:
> @Tantacrul sorry the ping but wouldn't this trigger a phasing issue like the same sample? I always thought that the reason this couldn't be achievable is because It's the same sample being triggered. I would suggest WHEN this happens, there would be a system that could trigger samples from a half/whole step down/up to use a different sample, otherwise it will become out-of-phase.

Let us worry about that.

MarcSabatella:
[...]

PR #14858, opened 2022-12-01
[MU4] multivoice_playback_fluid

vpereverzev:
Resolves: #13853


MarcSabatella:
Wow, Vasily!  I tried a whole bunch of cases and it's working as expected for me.

PR #14890, opened 2022-12-02
Fix cross system glissando bug

mike-spa:
Resolves: #15295
A nasty bug with glissandos when start-end notes cross systems (see vtest). Basically, the computation wasn't taking into account that, within each system, the y-coordinate of the staff may be different.


oktophonie (approved):

Tantacrul:
@RomanPudashkin @vpereverzev - can you let us know if this is safe to merge for 4.0.2

src/engraving/libmscore/system.h
```diff
@@ -235,6 +235,8 @@ class System final : public EngravingItem
     double squeezableSpace() const;
     bool hasCrossStaffOrModifiedBeams();
 
+    static double yStaffDifference(System* system1, staff_idx_t staffIdx1, System* system2, staff_idx_t staffIdx2);
```
RomanPudashkin, line 238: Do we really need this function here? I think a better place for it would be libmscore/utils.h (if it will be needed outside of glissando.cpp)

src/engraving/libmscore/system.cpp
```diff
@@ -2144,4 +2144,19 @@ bool System::hasCrossStaffOrModifiedBeams()
...
+
+//----------------------------------------------------------------
+// yStaffDifference
+// Static utility function. Returns the difference of y-coordinate
+// between any two given staves in any two given systems.
+//----------------------------------------------------------------
+
+double System::yStaffDifference(System* system1, staff_idx_t staffIdx1, System* system2, staff_idx_t staffIdx2)
+{
+    if (!system1 || !system2 || staffIdx1 > system1->staves().size() || staffIdx2 > system2->staves().size()) {
```
RomanPudashkin, line 2156: staffIdx1 **==** system1->staves().size() is also an invalid input

src/engraving/libmscore/system.cpp
```diff
@@ -2144,4 +2144,19 @@ bool System::hasCrossStaffOrModifiedBeams()
...
     return false;
 }
+
+//----------------------------------------------------------------
+// yStaffDifference
+// Static utility function. Returns the difference of y-coordinate
+// between any two given staves in any two given systems.
+//----------------------------------------------------------------
+
+double System::yStaffDifference(System* system1, staff_idx_t staffIdx1, System* system2, staff_idx_t staffIdx2)
```
RomanPudashkin, line 2154: Both system1 and system2 should be const

src/engraving/libmscore/system.cpp
```diff
@@ -2144,4 +2144,19 @@ bool System::hasCrossStaffOrModifiedBeams()
...
+{
+    if (!system1 || !system2 || staffIdx1 > system1->staves().size() || staffIdx2 > system2->staves().size()) {
+        // Safety checks
+        return 0.0;
+    }
+    return system1->staff(staffIdx1)->y() - system2->staff(staffIdx2)->y();
```
RomanPudashkin, line 2160: It would be better to get this staff from system1->staves(). So your algorithm would look like this:

```
const std::vector<SysStaff*>& staves1 = system1->staves();
const std::vector<SysStaff*>& staves2 = system2->staves();
if (staffIdx1 >= staves1.size() || staffIdx2 >= staves2.size()) {
    return 0.0;
}

return staves1[staffIdx1]->y() - staves1[staffIdx2]->y();
```

Or you can rewrite it as follows:

```
const Staff* staff1 = system1->staff1(staffIdx1);
const Staff* staff2 = system2->staff2(staffIdx2);

if (!staff1 || !staff2) {
    return 0.0;
}

return staff1->y() - staff2->y()
```

mike-spa:
@RomanPudashkin thank you! I've implemented all your suggestions, let me know if it's good to go 👍

RomanPudashkin (approved):

Issue #14915, opened 2022-12-03
[MU4 Issue] Grace note slurs are broken under certain conditions.


bakajikara:
**Describe the bug**s.
Grace note slurs are broken when you delete measure under certain conditions.

**To Reproduce**
See video

https://user-images.githubusercontent.com/102702868/205448966-64eff729-be17-4a98-8ecb-118063ef5fb9.mp4

**Platform information**
 - Windows 11, RC1

**Additional context**

PR #14953, opened 2022-12-05
Fix bug for grace note slur upon deleting measure

mike-spa:
Resolves: #14915 

Basically, the slur endpoint was being updated before the end-chord was, so it was still using the "old" (i.e. before-removing-measure) position of such chord, which caused it to wrongly end up in the next measure (and therefore in the next system). As far as I can tell, this happens inside `Score::deleteMeasures()` because we are calling `undoInsertTime()` (which updates the spanners) before `undoRemoveMeasure()` (which updates all the ticks). Moving the call of undoInsertTime() after the for loop seems to solve the problem.

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -2846,7 +2846,6 @@ void Score::deleteMeasures(MeasureBase* mbStart, MeasureBase* mbEnd, bool preser
     Fraction startTick = mbStart->tick();
     Fraction endTick   = mbEnd->tick();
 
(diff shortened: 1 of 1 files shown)


oktophonie (approved):

RomanPudashkin (approved):

Vykoiazon:
There is still be a problem with grace not slurs after deleting bars, I can post the full report on another issue but it looks like this:

https://user-images.githubusercontent.com/120861729/208555413-03292ade-5933-48e2-9de9-26ea57bc2a5b.mp4

Issue #14952, opened 2022-12-05
[MU4 Issue] Issues using shortcut **Shift** + **right arrow** to extend slurs and system text lines


bkunda:
**Describe the bug**
Two parts to this bug, both of which concern the use of the shortcut **Shift** + **&rarr;**:
1. (This now moved to #14992): Slurs can no longer be extended between systems (**regression**: this was apparently fixed by [#10515](https://github.com/musescore/MuseScore/pull/10515))
2. System text lines cannot be extended past the bar in which they are inserted

**To Reproduce**
See demo video:

https://user-images.githubusercontent.com/86290556/205672644-9953a997-788d-41e7-8ddf-a276a7ba1353.mp4

**Expected behavior**
The shortcut **Shift** + **&rarr;** should allow both slurs and system text lines to be extended between systems. The behaviour should basically be the same as for staff text lines and hairpins (which currently work as expected). 

**Platform information**
 - macOS 12.6 (M1)
 - RC

**Additional context**
@Tantacrul Even though the slurs issue is technically a regression, I'm wondering whether it's benign enough to leave it in post-release.

Tantacrul:
If it's a regression on a previously fixed issue, we should fix it here. Also, it's a really bad bug anyway.

mike-spa:
👀

DmitryArefiev:
**Issue №1**. Slurs can no longer be extended between systems (**if attached to the last note in system**) - is not a regression. As I see now, it also happens in MS3 and was not fixed in #10515

But **Issue №2** System text lines cannot be extended past the bar in which they are inserted - is really **regression**. Should be fixed in 4.0

mike-spa:
Issue n.2 started with #14397, which fixed other serious issues. I'm looking into it now

Tantacrul:
Perfect. Thanks both!

DmitryArefiev:
Fixed in #14982

DmitryArefiev:
Issue 2 is fixed. For Issue 1 it's better to create a separate bug.

bkunda:
> Issue 2 is fixed. For Issue 1 it's better to create a separate bug.

@DmitryArefiev done: https://github.com/musescore/MuseScore/issues/14992

PR #14982, opened 2022-12-06
Fix shift + rightArrow system line edit

mike-spa:
Resolves: #14952 

The regression started in #14397. 
The correct layout of these lines requires that they end at the left-most segment at that tick (which is often the endBarLine of the previous measure). I originally did it by modifying the endSegment() of the line, but this was preventing it to jump to the next measure. Now I do it at layout stage without modifying the endSegment().

### DIFF
--- a/src/engraving/libmscore/line.cpp
+++ b/src/engraving/libmscore/line.cpp
@@ -981,6 +981,15 @@ PointF SLine::linePos(Grip grip, System** sys) const
             x += cr->segment()->pos().x() + m->pos().x();
         } else {
             Segment* segment = (grip == Grip::START) ? startSegment() : endSegment();
+            if (grip == Grip::END && segment && segment->rtick().ticks() == 0) {
+                // The line should end on the left-most segment at this tick. If endSegment() is the first of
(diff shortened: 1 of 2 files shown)


src/engraving/libmscore/line.cpp
```diff
@@ -981,6 +981,15 @@ PointF SLine::linePos(Grip grip, System** sys) const
             x += cr->segment()->pos().x() + m->pos().x();
         } else {
             Segment* segment = (grip == Grip::START) ? startSegment() : endSegment();
+            if (grip == Grip::END && segment->rtick().ticks() == 0) {
```
RomanPudashkin, line 984: make sure that `segment` is not null

RomanPudashkin (approved):

DmitryArefiev:
Tested #14952 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

bkunda:
I've also just tested this in macOS. Can confirm it works as expected now.

oktophonie (approved):

PR #14990, opened 2022-12-06
[MU4] Fix #14752: Changing volume faders causes VST effects to be removed

### LINKED ISSUE #14752: [MU4 Issue] Changing volume faders causes VST effects to be removed
Double clicking on the master fader knob, which my intention is to reset the value back to 0, it is also ending up deleting the master effects which should not happen at all! FXs should still stay in place no matter what.

The only way to get those effects back is to close without saving then re-open project again.

MacOS Ventura

https://user-images.githubusercontent.com/71656189/204025982-a20874f1-37a2-435c-bb0d-f05aeb262574.mov

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/14752

### DIFF
--- a/src/playback/view/internal/mixerchannelitem.cpp
+++ b/src/playback/view/internal/mixerchannelitem.cpp
@@ -155,6 +155,8 @@ void MixerChannelItem::loadOutputParams(AudioOutputParams&& newParams)
         emit mutedChanged();
     }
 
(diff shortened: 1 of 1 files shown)


vpereverzev (approved):

DmitryArefiev:
Tested #14752 on Win10, Mac12 - **FIXED**

Issue #14811, opened 2022-11-28
[MU4 Issue] Trill line trills all staffs of the instrument


LuisLiszt:
**Describe the bug**
Trill line trills all notes of one instrument, in this example the bass staff of the piano.

**To Reproduce**
1. Create a score 
2. Place notes in two staffs
3. Place trill line

**Expected behavior**
Trill should only trill the notes of it's assigned staff like the normal trill ornament.

**Screenshots**

https://user-images.githubusercontent.com/55204261/204349198-31b07c7e-fe18-4823-9a4e-b9d420f499a9.mp4

**Platform information**
 - OS: Windows

**Additional context**
Probably relates to the problems that existed previously with the trill line.

LuisLiszt:
Yes, same with MSBasic!

PR #14991, opened 2022-12-06
[MU4] spanners_rendering

vpereverzev:
Resolves: #14811


Tantacrul:
Tested on MacOS and Win 10

The reported issue is no longer happening.

PR #14993, opened 2022-12-06
Fix #14944: Prevent adding multiple beam fragments

### LINKED ISSUE #14944: [MU4 Issue] Cannot manually move beam that crosses barline
**Describe the bug**
Beams that cross barline cannot (seemingly) be manually moved.

**To Reproduce**

image

Steps to reproduce the behaviour:
1. Create a beam across a barline
2. Try to drag any of the handles
3. The beam will not move; the _handles_ will (sometimes); in any case, the act of attempting this will set the beam as having a 'custom position', and if the pitches then change, the beam will stay in the same place. This is the result of transposing the initial image up an octave, after moving the beam was tried:
image

**Expected behaviour**
Beams the cross barlines should move like any others. (Note that beams are currently not drawn when they cross system breaks, though.)

**Platform information**
OS: Artix Linux, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223362132, revision: github-musescore-musescore-d050fb0 (RC1)

**Additional context**
For another beam-over-barline quirk that may be tangentially related, see https://github.com/musescore/MuseScore/issues/13624

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/14944

The problem was that during layout, the second measure of a cross-measure beam didn't have its system explicitly set, which made the beam think that the second measure was on a different system. Now, that codepath only happens when chordrests are _explicitly_ on different systems.

### DIFF
--- a/src/engraving/layout/layoutbeams.cpp
+++ b/src/engraving/layout/layoutbeams.cpp
@@ -332,15 +332,20 @@ void LayoutBeams::createBeams(Score* score, LayoutContext& lc, Measure* measure)
                 BeamMode mode = cr->beamMode();
                 if (mode == BeamMode::MID || mode == BeamMode::END || mode == BeamMode::BEGIN32 || mode == BeamMode::BEGIN64) {
                     ChordRest* prevCR = score->findCR(measure->tick() - Fraction::fromTicks(1), track);
+                    Beam* prevBeam = prevCR->beam();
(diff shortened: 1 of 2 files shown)


oktophonie (approved):

RomanPudashkin:
@asattely please rebase it

RomanPudashkin (approved):

Issue #14958, opened 2022-12-05
[MU4 Issue] Falls play as chromatic instead of portamento (MS Basic)


MarcSabatella:
**Describe the bug**
In the RC, falls play as chromatic instead of portamento when using MS Basic

**To Reproduce**
1. Create a score for trumpet
2. Add a whole note
3. Add a fall (from arpeggios & glissandi palette)
4. Play
5. Hear error: fall is chromatic

**Expected behavior**
Fall should use portamento.  Similarly for the other similar elements (doit, etc)

**Screenshots**

https://user-images.githubusercontent.com/1799936/205691346-b59e0877-afc4-4968-89f3-3a3d9f59dc93.mp4

**Platform information**
 - OS: Linux

**Additional context**
This was working well up until shortly before the RC.  But I can't say for sure when it broke.

Tantacrul:
Putting in Release, since this was working correctly a few weeks ago.

PR #14997, opened 2022-12-06
[MU4] pitch_bend_intervals

vpereverzev:
Resolves: #14958


Tantacrul:
Tested on MacOS and Win 10

- The pitch bend sounds much better at different tempos. I tested falls, plops, etc.
- I also tested the original issue with slides, where they caused proceeding notes to be delayed. There is no regression with this fix either.

Issue #12775, opened 2022-08-10
[MU4 Issue] Build fail with `-DBUILD_JACK=ON`


hfiguiere:
**Describe the bug**
Build fail with -DBUILD_JACK=ON because the macro PKGCONFIG1 is unknown

**To Reproduce**
1. configre the build on Linux wtih `-DBUILD_JACK=ON`
2. build
3. Build fail because PKGCONFIG1 isn't known.

**Expected behavior**
Either it find jack or it doesn't.

**Screenshots**
n/a

**Platform information**
 - OS: Linux

**Additional context**

You need this patch:
```diff
diff --git a/build/cmake/FindJack.cmake b/build/cmake/FindJack.cmake
index 6713e3aeef..58cc8cee14 100644
--- a/build/cmake/FindJack.cmake
+++ b/build/cmake/FindJack.cmake
@@ -3,6 +3,7 @@
 ## Find JACK >= JACK_MIN_VERSION
 ##
 
+include(UsePkgConfig1)
 
 IF(BUILD_JACK)
      IF(MINGW OR MSVC)
```

Tantacrul:
I'm moving Jack issues to 4.x, which is when we will begin to address them.

Closing until 4.0 is released, after which, I'll reopen.

tibequadorian:
This is not a Jack issue and @hfiguiere already provides a patch. This should be fixed before the 4.x release.

cbjeukendrup:
`-DBUILD_JACK=ON` does nothing, so you can just drop it.

PR #14999, opened 2022-12-06
Fix cmake error when building with jack on Linux

jamesjer:
Resolves: #12775

Attempting to build with JACK support on Linux fails:
```
CMake Error at build/cmake/FindJack.cmake:31 (PKGCONFIG1):
  Unknown CMake command "PKGCONFIG1".
Call Stack (most recent call first):
  CMakeLists.txt:184 (include)
```
This PR adds the necessary include to ensure the PKGCONFIG1 macro is defined.

- [x] I created the test (mtest, vtest, script test) to verify the changes I made (actually N/A)


cbjeukendrup (approved):
This is correct, but be aware that `-DBUILD_JACK=ON` does nothing, apart from trying to find Jack by calling this CMake file (there is simply no Jack support in MS4.0, unfortunately).

jamesjer:
Ah, good to know, thank you.  I've been the MuseScore maintainer for the Fedora Linux distribution for awhile now.  I'm in the process of figuring out how to build 4.0 RC1, and stumbled across this.  In that case, I will simply stop attempting to build in Jack support for now.

Issue #15012, opened 2022-12-07
[MU4 Issue] Crash when pasting note that has a fall, doit, plop (etc.) attached to it.


Tantacrul:
**Describe the bug**
See video. Seems to affect all the fall, doit, plop, etc. markings.

https://user-images.githubusercontent.com/21022311/206156072-6c1b1f88-8dbf-43da-b30c-cffc0e59a0eb.mov

**Platform information**
 - OS: MacOS

**Additional context**
Crashes on RC

PR #15019, opened 2022-12-07
[MU4] note_paste_crash

vpereverzev:
Resolves: #15012 

The parent of the pasted note is being set later on at ChordRest::drop(..)


RomanPudashkin (approved):

Tantacrul:
Seems to be working great.

image

PR #15079, opened 2022-12-09
[MU4] Lyrics copy paste fix

Eism:
Fixed this bugs:
- If the user selected text in the lyrics and inserted new text, then insertion at the end occurs, not replacement
- When copying a text from another text box and pasting it to lyrics, formatting text is inserted into the lyric.

### DIFF
--- a/src/engraving/libmscore/lyrics.cpp
+++ b/src/engraving/libmscore/lyrics.cpp
@@ -431,9 +431,12 @@ void Lyrics::layout2(int nAbove)
 
 void Lyrics::paste(EditData& ed, const String& txt)
 {
-    MuseScoreView* scoreview = ed.view();
+    String correctedText = txt;
+    //! NOTE: Remove formating info. For example, for "<info><info>text" will be "text"
+    correctedText = correctedText.remove(std::regex("\\<(.*?)\\>"));
+
     String regex = String(u"[^\\S") + Char(0xa0) + Char(0x202F) + u"]+";
-    StringList sl = txt.split(std::regex(regex.toStdString()), mu::SkipEmptyParts);
+    StringList sl = correctedText.split(std::regex(regex.toStdString()), mu::SkipEmptyParts);
     if (sl.empty()) {
         return;
     }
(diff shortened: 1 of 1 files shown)


Eism:
@oktophonie FYI
Pasting formatting for lyrics does not work at the moment (release, master), you can test it by copying any text from the score (for example, from the title) and pasting it into the lyrics.

In this PR, I have explicitly disabled pasting formatting for lyrics.

Inserting formatting for lyrics should be implemented as part of a separate task. Of course, if we need this feature for the lyrics.

DmitryArefiev:
> If the user tried to insert text with symbols, then only the first part before the symbol is inserted. For example, for “1. Bla Bla” the result will be “1.“

This should be like in **MS3**: when pasting a text with spaces it should interpret each space as a move to next syllable. 
Actually, it doesn't work as it should in **master** as well - #13215

oktophonie (approved):

DmitryArefiev:
> Fixed this bugs:
> 
>     * If the user selected text in the lyrics and inserted new text, then insertion at the end occurs, not replacement
> 
>     * When copying a text from another text box and pasting it to lyrics, formatting text is inserted into the lyric.

Tested these on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

RomanPudashkin (approved):

Tantacrul:
Excellent. Thanks guys!

sammik:
I thought "Bug 1" was not bug, but feature.

It was used quite often, when pasting lyrics sylabe-by-sylabe.

User prepared "sy-la-be". But sometimes, it needed to be "syla-be". So after "sy-" user went back and continues pasting and "sy-" became "syla-", which was great and fast.

Of course, sometimes user wanted replace sylabe, instead of adding it, but in my experience, it was way less often, than adding sylabe to previous one.

oktophonie:
When you press the right arrow to go back to a previous syllable, the whole string of that syllable is selected:
image
and in this situation it seems like completely predictable and expected behaviour that pasting, or typing anything, would replace the syllable.

To append to the end instead, all one needs to do is press the right arrow once first.

Issue #14916, opened 2022-12-03
[MU4 Issue] "Advance cursor" (figured bass/chord symbols) command not working


Der-Leiermann:
**Describe the bug**
Edit by @bkunda: 

The shortcuts `Cmd`+`1`...`9` no longer work to advance the cursor by durational value when entering chord symbols or figured bass. This prevents the user from, for example, indicating multiple chords over a sustained note. It also makes navigation more tedious, especially when entering figured bass.

We are fairly certain the current effect on chord symbols in particular is a recent regression ([see discussion below](https://github.com/musescore/MuseScore/issues/14916#issuecomment-1339549369)).

**To Reproduce**
Steps to reproduce the behavior (example here is for figured bass):
1. Enter half note.
2. Select the note and press `Command` + `G`.
3. Enter figure.
4. Press `Command` + `5`.
5. See that the cursor has not advanced

**Expected behavior**
The cursor should have advanced by the value of a quarter note

**Platform information**
 - OS: macOS Ventura 13.0.1 (also tested on macOS 12.6)
 - MuseScore version: Release Candidate 1

MarcSabatella:
I can confirm this, but I would add, you really should enter an actual figure between pressing Cmd+G and then changing the duration.  Not sure that would be defined.  But certainly, after entering the figure, Cmd+number should work, and currently does not (well,m Ctrl+number on keyboards other than Mac).

sammik:
I can confirm on Linux.

And it is regression from (from 3.6.2, where everything worked) and also from RC0 (where duration worked, but "space" didnt /it started playback/). Here, in RC1 it is oposite ("space" works as "next note", but duration doesnt).

Der-Leiermann:
@MarcSabatella Thanks, I updated the steps. Also, apologies for accidentally closing. :)

bkunda:
I too can confirm this, but unfortunately I can't check the regression status against an earlier build, because every old build I try to run crashes on my machine when opening! 🤷‍♂️
Anyway, it is a sufficiently bad one in that, without the navigation commands, the user has to rely on the mouse to click on every note for which they wish to add figured bass.

MarcSabatella:
No, I was unaware there was a problem until I decided to try to see if I could find a workaround for the figured bass problem, then it dawned on me to try it for chord symbols.

It's the same bug since it's the same command, and probably the same fix will address both.

[...]

Issue #13148, opened 2022-09-03
[MU4 Issue] Alt+Left/Right no longer navigate while entering text (or navigate between figured bass, fingering)


MarcSabatella:
**Describe the bug**
Alt+Left/Right are supposed to be available to navigate from note to note while entering staff text or other elements that don't support navigation via Space/Shift+space

**To Reproduce**
1. add several notes to score
2. select first
3. Ctrl+T to enter staff text
4. type "one", then Alt+Right
5. Result: nothing happens

**Expected behavior**
What should happen is as in MuseScore 3 - entry of the current text is completed, cursor moves on to next note to allow you to enter another staff text.  Similar to how Space works when entering lyrics.

**Platform information**
 - OS: Linux

**Additional context**
This was a very handy feature for educators adding annotations to scores

FrancRos31:
Same issue with dinamics: alt + right/left doesn't move the selection but change the dinamic offset.

PR #4521, opened 2018-12-30
fix #278175: add fingering mode (also works for staff text)

MarcSabatella:
See https://musescore.org/en/node/278175 for some discussion.  Adding an easy way to add quickly move through the score adding fingerings - like we have for lyrics - is a common request.  It always seemed like it would be a lot of work to introduce a whole new mode and the necessary semantics for it, but recently it occurred to me there is a much simpler approach that leverages the existing next/previous element commands (alt+left/right).

The basic gist of my change is to tweak the behavior of alt+left/right so that if you are in text edit mode, we not only move to the next note, but also terminate edit mode on the original text and then invoke it on a text attached to the next note.

Bottom line is that entering fingerings now can work almost exactly like lyrics.  Staff text too.  Selection moves to next/previous note; if there is a text of the same type there, you can edit it, if not, one is added.

One might wish it were Space/Shift+Space that invokes this, but we can't do that without introducing a whole new state and all the state transitions and handling of cases that means throughout the code.  After all, we can hardly reserve Space for a command during regular text edit.

If someone wants to do the work of adding a whole new mode just for fingering edit, we could still use the same textTab() function I introduced as the handler, and still keep the alt+left/right behavior in addition.

PR #15081, opened 2022-12-09
Fix "Advance cursor" shortcuts for figured bass/chord symbols

cbjeukendrup:
Resolves: #14916

Problem was that TextBase::isEditAllowed returned "true" way too often; even for shortcuts that it doesn't handle at all. Therefore, these shortcuts are blocked from going through the normal shortcuts/actions system, which prevents the corresponding actions from working. 

For testing:
We basically need to make sure that every possible shortcut you might want to use in text edit mode works. And that needs to be done for _every_ text type: lyrics, chord symbols, figured bass, fingering, sticking, but also normal staff texts and text boxes.

I discussed this with @vpereverzev today; probably, it won't be suitable for the rc branch, as this is quite a risky change since it feels like this whole TextBase / "editing" stuff is a bit of a mess.


sammik:
I checked this PR and doesnt seem to work for Figured bass, nor Chords.

It brings also new issue - ESC doesnt work in Lyrics, in Chords, in Text, in Figured Bass.


https://user-images.githubusercontent.com/1646034/206876194-87284553-4db8-4d64-b2c0-ede5eda4d40c.mp4

Linux Ubuntu Studio 22.04

cbjeukendrup:
@sammik Strange... on Mac, it works well now. I'll try on Windows, to see if it's indeed still broken on non-Mac.

I could reproduce the Esc problem though; that should be fixed now.

sammik:
> @sammik Strange... on Mac, it works well now. I'll try on Windows, to see if it's indeed still broken on non-Mac.
> 
> I could reproduce the Esc problem though; that should be fixed now.

I checked win.
It seems, that numpad doesnt work now also for duration in note entry.
Also duration in Figured bass doesnt work for me (I checked numpad, and standard numbers).

MarcSabatella:
As far as I know from dozens of reports, numpad doesn't work period in MU4, on any system, for anything.  I don't have a numpad to test myself.

sammik:
> As far as I know from dozens of reports, numpad doesn't work period in MU4, on any system, for anything. I don't have a numpad to test myself.

Numpad normally works for me. Also in latest nigthly (in note input mode).

src/engraving/libmscore/textedit.cpp
```diff
@@ -243,33 +243,115 @@ void TextBase::insertText(EditData& ed, const String& s)
...
+    case Key_Underscore:
+        return true;
+    // Select all is handled by us
+    case Key_A:
+        if (ctrlPressed && !shiftPressed) {
+            return true;
+        }
+        break;
+    // Several commands
+    // TODO: looks like we're hard-coding keyboard shortcuts here, but what if user has edited those?
```
RomanPudashkin, line 309: I guess we should disable the ability to edit them in the preferences. There is no real way to change these shortcuts because they are hardcoded in TextBase and its derived classes. It would take a lot of refactoring to add a real ability to change these shortcuts, which would lead to a lot of regressions. So I don't think it's worth our effort. So instead of adding this comment, I would suggest opening an issue about removing these shortcuts from the preferences

RomanPudashkin (approved):

DmitryArefiev:
@cbjeukendrup Tested on Win10/Mac13 - it works on Mac but **DOESN'T WORK** on **Windows**

cbjeukendrup:
Yeah, this wasn't quite ready for testing, as found in the comments above... Sorry, I didn't know you were already assigned, otherwise I would have unassigned you :)

DmitryArefiev:
> Yeah, this wasn't quite ready for testing, as found in the comments above... Sorry, I didn't know you were already assigned, otherwise I would have unassigned you :)

OK. No problem :)

cbjeukendrup:
Now it should finally be fixed on Windows too!

DmitryArefiev:
> Now it should finally be fixed on Windows too!

Yes, now it looks fine on Windows and Linux too!

DmitryArefiev:
Tested #14916 on Win10, Mac12, LinuxMint 22.04 - **FIXED**

igorkorsukov (approved):

MarcSabatella:
FWIW, also see https://github.com/musescore/MuseScore/issues/13148.  Alt+Left/Right is meant to provide note-to-note navigation between staff texts.  I assume that code is still present somewhere but isn't being hit.  Also see https://github.com/musescore/MuseScore/issues/15156.  Ctrl+B and Ctrl+I are broken too.  Neither of these are new regressions since this PR, but it seems the code in this PR is probably relevant to any fix.

MarcSabatella:
@cbjeukendrup Looking over the code changes here and thinking through what is going on, I definitely agree a major overhaul someday would be nice.  But for now, I'd be happy with a "bandaid" fix for the specific issues I see:

1) Ctrl+Left/Right not working in chord symbols or figured bass - that's a regression from 4.0.2
2) Ctrl+Underscore doesn't work in lyrics (should insert underscore) - regression from MU3 but not working in 4.0.2 either
3) Alt+Left/Right not working for staff text (should work like Space in fingering) - regression from MU3 but not working in 4.0.2 either

For 1), I *think* the best places to fix this would be in Harmony::isEditAllowed() and FiguredBass::isEditAllowed().  In 4.0.2 I guess it must have worked because Ctrl+Left/Right was being detected by isTextNavigationKey() but isn't anymore, and adding that back feels like a step backwards, so special casing it in Harmony & FiguredBass is my vote.  I'm happy to try a PR for this, unless you'd rather.

For 2), I'm really at a loss; I have no idea where that is supposed to be handled.  Presumably the same place Ctrl+Minus gets handled to insert a hyphen, but I can't even find that.

For 3), I wrote the original code for this in MU3 but too much has changed for me to say how to best address it.  I was hooking into the next-element and prev-element handlers but that's more complicated now.  I think *maybe* the most logical place to reproduce the functionality is https://github.com/musescore/MuseScore/blob/5cda56428e423623e7330de4ef08f686c286c853/src/notation/internal/notationinteraction.cpp#L2661.  The necessary support code seems intact, it's textTab() and is being used successfully by fingering.  Here's how it was done in MU3 - https://github.com/musescore/MuseScore/pull/4521/files#diff-3e7cf0f7e03151067c94f814562f7c5abcca42c7cf729ec8dd10991e0935ce84

PR #15088, opened 2022-12-10
Instruments: Make Bb trumpet the default instead of C trumpet

### LINKED ISSUE #14957: [MU4 Issue] Instruments dialog - search for "trumpet" defaults to C
**Describe the bug**
If you search for "trumpet" in the instruments dialog, the default choice offered is C, but it should be Bb.

**To Reproduce**
1. Create a new score
2. Under "Choose instruments", type "trumpet" into the search box
3. See error: the default choice offered is C, and you need to change to Bb manually via the dropdown

**Expected behavior**
The default should be Bb, with C available via the dropdown

**Screenshots**
Screenshot 2022-12-05 9 16 08 AM

**Platform information**
 - OS: Linux

**Additional context**
It's not that way for cornet, horn, clarinet, or other instruments offered in a variety of transpositions - they all default to the most common variant.  I do recognize that in orchestral music, C trumpet might be about as common as Bb, but it's not common at all outside that world.  And Bb trumpet is the one listed if, instead of searching, you select the Brass family in the Common genre.  So it makes no sense that C would be offered as the default choice in the search.

This is not a recent regression; it's been this way for as long as that drop-down has been a thing, I believe.  MU3 didn't have this; it just listed all matching instruments in one long list.

shoogle:
Now Bb will be the default option in the Trumpet dropdown in the NSW and Instruments dialog.

Fix #14957

Before | After
---|---
image | image

I also updated the description of the Bb trumpet:

Before | After
---|---
image | image

Ignore the missing space after the full stop. This is a known bug in how Qt strings are displayed on old versions of macOS. In reality there is a space character there that is displayed on other platforms.

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -4693,7 +4693,7 @@
                   <trackName>Trumpet</trackName>
                   <longName>Trumpet</longName>
                   <shortName>Tpt.</shortName>
-                  <traitName type="transposition">*C</traitName>
+                  <traitName type="transposition">C</traitName>
                   <description>Trumpet in C. Nowadays the most common orchestral trumpet.</description>
                   <musicXMLid>brass.trumpet.c</musicXMLid>
                   <clef>G</clef>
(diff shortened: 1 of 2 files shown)


oktophonie (approved):

RomanPudashkin (approved):

Tantacrul:
Tested this again. Looking good!

PR #15097, opened 2022-12-10
Add ability to skip certain instruments in update_instruments_xml.py

shoogle:
If an instrument's traitName is set to "[skip]" in the online spreadsheet, when update_instruments_xml.py is run it will ignore that instrument, causing it to be omitted from instruments.xml.

This provides a way to add new instruments to the spreadsheet during development without having them appear in the program until we are ready for them. It also makes it easier to make changes to the spreadsheet across multiple PRs in parallel. The [skip] tag should be used for new instruments up until the PR adding them is merged, at which point the [skip] tag will be removed to allow other PRs to be rebased simply by running the script again.

For example, the [skip] tag is currently being used to prevent the instrument changes in PR #14703 from appearing in instruments.xml when update_instruments_xml.py is run on the master branch. After that PR is merged, the [skip] tag will be removed so that those instruments do appear on master.

### DIFF
--- a/share/instruments/update_instruments_xml.py
+++ b/share/instruments/update_instruments_xml.py
@@ -153,10 +153,12 @@ for family in families.values():
 
 # Add articulations to XML tree
 for articulation in articulation_defaults.values():
-    el = ET.SubElement(root, 'Articulation')
+    el = ET.Element('Articulation')
     to_attribute(el, articulation, 'name')
     to_subelement(el, articulation, 'velocity')
     to_subelement(el, articulation, 'gateTime')
+    if el.find('*') is not None:
+        root.append(el)
 
 # Add groups and instruments to XML tree
 for group in groups.values():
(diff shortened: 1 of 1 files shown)


cbjeukendrup (approved):
Looks good to me. Regarding how this is supposed to be used: is the idea that when working for example on https://github.com/musescore/MuseScore/pull/14703 itself, you temporarily remove the `[hide]` tags from the spreadsheet before running this script, and then afterwards put them back in the spreadsheet so that other PRs are not affected?

shoogle:
@cbjeukendrup, yes exactly. If you want to submit a PR or edit an existing one, temporarily remove the `[skip]` tag from the relevant instruments in the spreadsheet, run the script to generate instruments.xml, then put the `[skip]` tag back again so other PRs are not affected.

BTW, the `[skip]` tag is different to the `[hide]` tag:

- `[hide]` - Include the instrument in instruments.xml but don't display it in the "Choose instruments" dialog. MuseScore will recognise this instrument when opening older scores from MU3 but it won't allow users to create new scores with this instrument. Used for backwards compatibility purposes.
- `[skip]` - Exclude the instrument from instruments.xml entirely. Used to allow parallel contributions to the spreadsheet.

We'll probably get rid of `[hide]` soon and replace it with a different way of identifying deprecated instruments.

PR #15103, opened 2022-12-11
Fix: retry the command -> retry

Gootector:
The "...and retry the command" occurs only once in strings to translate - only in this case. These cases above are with "or" instead of "/"; and this case name is: Err::SelectCompleteTupletOrTremolo.

Greetings,
Gootector

### DIFF
--- a/src/notation/notationerrors.h
+++ b/src/notation/notationerrors.h
@@ -60,7 +60,7 @@ inline Ret make_ret(Err err)
                + "\n" + trc("notation", "Please select a measure and retry");
         break;
     case Err::SelectCompleteTupletOrTremolo:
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

DmitryArefiev:
No way to re-test it from QA side. And actually, it's a really minor change

bkunda:
> No way to re-test it from QA side. And actually, it's a really minor change

I was just about to write the same thing 😅. @DmitryArefiev are you okay moving this to "Done – Tested Again" in this case?

DmitryArefiev:
> > No way to re-test it from QA side. And actually, it's a really minor change
> 
> I was just about to write the same thing 😅. @DmitryArefiev are you okay moving this to "Done – Tested Again" in this case?

Absolutely!

PR #15107, opened 2022-12-11
Fix #13276: Removed updates on editing style textfields

### LINKED ISSUE #13276: [MU4 Issue] Typing values into style fields triggers updates on each keystroke
**Describe the bug**
Perhaps more of a request or suggestion than a 'bug', but I'll explain why it is a problem.

For example: let's say I want to change my page width to 184mm. I open format > Page settings and select the value currently in the page width field (210mm) for me.

I can get the width I want by typing "184". But the score updates on every keystroke, and other settings can be affected too. In this case, as soon as I type "1", the score becomes 1mm wide, and also the _margins_ become 1mm (constrained as such by the width):

image

and when I type the "8" and "4" the margins obviously do not increase again, so while I have successfully set my desired page width, my margins are now messed up.

This applies to all fields in the Page settings dialog and also the main Style dialog.

As well as being annoying, a full recomputation of the score layout happens with each keystroke too. In the case of a large, complex score this can be very slow, and if you are typing multiple digits into a style field, you are likely to see nonsense until they are all typed (plus it runs the risk of messing the layout up in unpredictable ways).

Would it be possible for the 'value updated' message to be sent not on every keystroke, but only by some appropriate trigger, for instance tabbing or clicking away from the field?

HemantAntony:
Resolves: #13276


oktophonie:
This looks great, though I'd [agree with Marc](https://github.com/musescore/MuseScore/issues/13276#issuecomment-1345564168) that it would be even better if it worked for instrument labels too; is that possible?

RomanPudashkin (approved):

oktophonie (approved):

Issue #14349, opened 2022-11-05
Musescore 4 MIDI Import rounds tempo changes to integers


vsthem:
MuseScoreNightly-223090408-master-2c34155-x86_64

Import the MIDI File with Tempo changes that include decimal places, ex: 90.55BPM

Current behavior: Takes tempo changes and rounds them to integers

Expected behavior: Import Tempo Changes with decimal places without rounding them to the nearest integer

Attempting to use JACK and Ardour to sync Musescore to audio that was not recorded to a click, and later tempo mapped. For the sync to work, Musescore must be able to import more exact tempo changes.

cbjeukendrup:
I think we should make this configurable. Not all users will like to see a "quarter note = 90.55" marking on their score... 
A "round tempo" checkbox in Preferences > Import > MIDI (on by default) would be nice.

This seems to be a good first issue for new contributors (or maybe a good second issue). As a starting point, see the following things:
- `importmidi_tempo.cpp`
- the `IMidiImportExportConfiguration` class (and where it is used)
- `MidiSection.qml` for the Preferences GUI
- the `ImportPreferencesModel` class for the import preferences model

shachiang428:
Can I be assigned to this issue? This would be my first contribution.

cbjeukendrup:
@shachiang428 Yes, you're welcome to work on this!

shachiang428:
@vsthem do you perhaps have the MIDI file in question and can you upload it to this issue?

vsthem:
> 
[midi_export_0.zip](https://github.com/musescore/MuseScore/files/10203680/midi_export_0.zip)
wouldn't let me upload a midi so I zipped it. Should have a tempo of 66.889 at Measure 33

PR #15115, opened 2022-12-12
Fix issue #14349: added togglable round tempo MIDI

shachiang428:
Resolves: (https://github.com/musescore/MuseScore/issues/14349)

Some users may find it frustrating that when importing MIDI files automatically rounds tempos to integers. This feature addition allows users to toggle the round tempo option as a solution to this issue.


src/appshell/qml/Preferences/internal/MidiSection.qml
```diff
@@ -29,9 +29,12 @@ BaseSection {
     title: qsTrc("appshell/preferences", "MIDI")
 
     property alias shortestNotes: shortestNotesBox.model
+    property alias roundTempo: roundTempoBox.checked
+
     property int currentShortestNote: 0
```
cbjeukendrup, line 34: Minor code style thing: the `shortestNotes` and `currentShortestNote` properties are closely related, so it would be better to keep them together. I propose to change the order like this:
```suggestion
    property alias shortestNotes: shortestNotesBox.model
    property int currentShortestNote: 0

    property alias roundTempo: roundTempoBox.checked
```

src/appshell/view/preferences/importpreferencesmodel.h
```diff
@@ -54,7 +54,7 @@ class ImportPreferencesModel : public QObject, public async::Asyncable
     Q_PROPERTY(bool needUseDefaultFont READ needUseDefaultFont WRITE setNeedUseDefaultFont NOTIFY needUseDefaultFontChanged)
 
     Q_PROPERTY(int currentShortestNote READ currentShortestNote WRITE setCurrentShortestNote NOTIFY currentShortestNoteChanged)
-
+    Q_PROPERTY(bool roundTempo READ roundTempo WRITE setRoundTempo NOTIFY roundTempoChanged)
```
cbjeukendrup, line 57: Please keep the empty line after this one, for better readability (same for the other places in this file)

src/importexport/midi/internal/midiimport/importmidi_tempo.cpp
```diff
@@ -102,7 +107,9 @@ void setTempoToScore(Score* score, int tick, double beatsPerSecond)
 
 double roundToBpm(double beatsPerSecond)
 {
-    return qRound(beatsPerSecond * 60.0) / 60.0;
+    auto conf = mu::modularity::ioc()->resolve<mu::iex::midi::IMidiImportExportConfiguration>("iex_midi");
+    bool round = conf ? conf->roundTempo() : true;
+    return round ? qRound(beatsPerSecond * 60.0) / 60.0 : (beatsPerSecond * 60.0) / 60.0;
```
cbjeukendrup, line 112: You can simplify this calculation a little bit:
```suggestion
    return round ? qRound(beatsPerSecond * 60.0) / 60.0 : beatsPerSecond;
```

The `* 60.0 / 60.0` is only relevant when rounding: we need to convert from beats per second to beats per minute, because we want to round to whole beats per minute, not necessarily whole beats per second. That's why we do that. But when not rounding, it's meaningless extra work.

cbjeukendrup (commented):
Looks very good already; only a few very small remarks!

src/appshell/view/preferences/importpreferencesmodel.h
shachiang428: Think I pushed all the new suggestions, let me know if the readability looks good for this!

fidwell:
Is there any chance of this being picked back up? I would love to have this feature!

cbjeukendrup:
Rebased (16103 new commits... 🙂)

PR #15117, opened 2022-12-12
Fix #15038: Recompute default stem length when hook is added

### LINKED ISSUE #15038: [MU4 Issue] When beaming changes, correct stem lengths and beam positions are not shown until re-render
**Describe the bug**
When changes are made to beams, the correct stem lengths and beam positions are not always shown immediately, but only when a re-render is triggered. This happens in all sorts of situations; here is a very simple one to illustrate the issue.

If we take this:
image
and split the beam, we get this:
image
The beam has split, and in this case stem _directions_ have been adjusted accordingly; but the stem _lengths_ are wrong (here both are 1/4sp too short). If you make a change that causes the layout to be re-rendered then they correct themselves:
image

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/15038

The stems were poorly sized because the computation for stem length came before any beams were removed / hooks were added. Now there is a recalculation of stem length.


oktophonie:
Looks good; it is also necessary to re-render when stem directions change. When the first note of a) is split from the group, for example, we get b). (The stem direction should flip, which does happen when the layout is recomputed.)
image

oktophonie (approved):

RomanPudashkin:
@asattely please pay attention to the failed utests

PR #15130, opened 2022-12-12
Update pull request template and fix some ambiguities

cbjeukendrup:
Based on commonly seen ""mistakes"" in PR descriptions, but mainly because now GitHub is mainly used rather than musescore.org's issue tracker, and to prevent confusion because GitHub issue numbers are starting to overlap with musescore.org issue numbers. From now on, we simply ask to link to the GitHub issue so that it will be closed automatically, and don't mention musescore.org anymore. 

(My future vision is that users can continue to create issues on musescore.org, and that we implement an automated way to transfer well-written non-duplicate issues to GitHub, so that the musescore.org issue ends up tracking the status of the GitHub issue, being closed automatically when the GitHub issue is closed etc..)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,14 +1,14 @@
-Resolves: *(direct link to the issue)*
+Resolves: #NNNNN <!-- Replace `NNNNN` with a GitHub issue number, or a direct link if the issue is not on GitHub -->
 
-*(short description of the changes and the motivation to make the changes)*
+<!-- Add a short description of and motivation for the changes here -->
 
-<!-- Use "x" to fill the checkboxes below like [x] -->
+<!-- Replace `[ ]` with `[x]` to fill the checkboxes below -->
 
-- [ ] I signed [CLA](https://musescore.org/en/cla)
-- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
-- [ ] I made sure the code compiles on my machine
-- [ ] I made sure there are no unnecessary changes in the code
-- [ ] I made sure the title of the PR reflects the core meaning of the issue you are solving
-- [ ] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
-- [ ] I made sure the commit message title starts with "fix #424242:" if there is a related issue
-- [ ] I created the test (mtest, vtest, script test) to verify the changes I made
+- [ ] I signed the [CLA](https://musescore.org/en/cla)
+- [ ] The title of the PR describes the problem it addresses
+- [ ] Each commit's message describes its purpose and effects, and references the issue it resolves
+- [ ] If changes are extensive, there is a sequence of easily reviewable commits
+- [ ] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
+- [ ] There are no unnecessary changes
+- [ ] The code compiles and runs on my machine, preferably after each commit individually
+- [ ] I created a unit test or vtest to verify the changes I made (if applicable)


shoogle (commented):
Good idea to update the PR template and drop `Fix #xxxx` from commit titles! There are only 72 characters available in commit titles (and ideally less than 50 characters) so I've always hated wasting those on a numeric tag that can just as easily go in the body of the message. Much better to use the title for a useful description IMO.

You could remove "I made sure" from the start of each bullet point as it's not really necessary. Also, take a look at [Audacity's PR template](https://github.com/audacity/audacity/blob/master/.github/pull_request_template.md) as I think they have said basically the same as us but in fewer words. I think coding rules is the only one we have that they don't have at all.

I noticed some projects have a GitHub Action that check to see whether the contributor has signed their CLA. It would be great if we could set something like that up at some point, though it would require custom coding so it's probably not worth the effort right now.

iwoithe:
> I noticed some projects have a GitHub Action that check to see whether the contributor has signed their CLA

Would something like [CLA Assistant](https://cla-assistant.io/) work for MuseScore?

shoogle:
> Would something like [CLA Assistant](https://cla-assistant.io/) work for MuseScore?

Yes, at first glance it looks like it would! 🙂

> Having commits link to the issue they fix enables:

(1), (2) and (4): Commits should still link to issues, just not in the title. The issue number should go in the body of the message. This will still close the issue automatically.

(3): IMO it's far more useful to see what the commit actually does rather than an issue number, which is pretty meaningless on its own. Usually there's room for an issue number and a message, but not always, especially if the commit fixes multiple issues. Anyway, if you are viewing the blame on GitHub then you can click on the commit title to see the full message, including links to any PRs and issues that were affected. If it's offline Git then you just need to run `git show 3e6da6e` (or whatever the short commit SHA is as displayed in the blame) and Git will show the full message for that commit, including any issue numbers. You can do this in a second Terminal tab so that you can keep `git blame` open in the first tab.

jeetee:
And just like that I've discovered of using `-m` twice with git commit -_-
Consider my objection removed!

shoogle:
Personally I use `git commit` without any `-m`. This opens in an editor allowing you to type a multi-line message. You can configure which editor is used by setting the `EDITOR` environment variable (e.g. `export EDITOR=nano`).

cbjeukendrup:
@shoogle I just pushed a second iteration, indeed inspired by Audacity!

.github/pull_request_template.md
```diff
@@ -1,14 +1,14 @@
...
-- [ ] I made sure the code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
-- [ ] I made sure the code compiles on my machine
-- [ ] I made sure there are no unnecessary changes in the code
-- [ ] I made sure the title of the PR reflects the core meaning of the issue you are solving
-- [ ] I made sure the commit message(s) contain a description and answer the question "Why do those changes fix that particular issue?" or "Why are those changes really necessary as improvements?"
-- [ ] I made sure the commit message title starts with "fix #424242:" if there is a related issue
-- [ ] I created the test (mtest, vtest, script test) to verify the changes I made
+- [ ] I signed the [CLA](https://musescore.org/en/cla)
+- [ ] The title of the PR describes the issue it addresses
+- [ ] Each commit's message describes its purpose and effects and references the issue it resolves
```
jeetee, line 9: I'd personally drop the 's from commit here; but I'm non-native, so what do I know

jeetee (approved):

.github/pull_request_template.md
shoogle: Either would work, though keeping `'s` is slightly more correct because it makes it clear we are talking about the purpose of the commit, not the purpose of the commit *message*.

However, we need a comma after "effects". It should be:

> Each commit's message describes its purpose and effects, and references any issues it resolves.

And I suggest rewording the one above to be:

> The title of the PR describes the ~~issue~~ __problem__ it addresses.

This avoids using the word "issue" to mean a generic problem as well as a specific numbered issue from our tracker.

shoogle (commented):
Looks good! I like the new wording. I just have small suggestions for two of the points.

shoogle (approved):

PR #15155, opened 2022-12-13
Fix #14460: Only put beamlets before if the beam level above is also before

### LINKED ISSUE #14460: [MU4 Issue] Faulty beamlet direction
**Describe the bug**
Applying "break inner beams (semiquaver/16th)" to the third note here:
image
creates this:
image
Where the beamlet on the second note is in the wrong direction.

This is probably a specific example of a slightly more general problem in the logic.

Note that this is _not_ a regression caused by #14429 - that change simply made it a bit easier to encounter it, specifically in this example (which is now beamed this way automatically, but one could have created it manually before, with the same result):
image

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/14460

### DIFF
--- a/src/engraving/libmscore/beam.cpp
+++ b/src/engraving/libmscore/beam.cpp
@@ -894,11 +894,15 @@ bool Beam::calcIsBeamletBefore(Chord* chord, int i, int level, bool isAfter32Bre
     // next note has a beam break
     ChordRest* nextChordRest = _elements[i + 1];
     ChordRest* currChordRest = _elements[i];
+    ChordRest* prevChordRest = _elements[i - 1];
     if (nextChordRest->isChord()) {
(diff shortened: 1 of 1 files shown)


oktophonie (dismissed):

oktophonie:
Setting 'join beams' on the second rest here
image
results in
image
where the beamlet should, instead, point in the same direction as the main beam.

asattely:
> Setting 'join beams' on the second rest here image results in image where the beamlet should, instead, point in the same direction as the main beam.

Sorry this slipped under my radar for so long, this is fixed and also the PR is rebased!

oktophonie (approved):

RomanPudashkin (approved):

PR #15174, opened 2022-12-14
Narrow spacing improvements

mike-spa:
Draft


src/engraving/layout/layoutsystem.cpp
```diff
@@ -1505,3 +1510,102 @@ void LayoutSystem::restoreTies(System* system)
     Fraction etick = system->measures().back()->endTick();
     doLayoutTies(system, segList, stick, etick);
 }
+
+void LayoutSystem::manageNarrowSpacing(System* system, double& curSysWidth, double targetSysWidth)
```
RomanPudashkin, line 1514: This algorithm looks very suboptimal. We iterate over this system->measures() list many many times (including some implicit iterations, such as inside system->minSysTicks/maxSysTicks). Considering how slow the layout process is, we should at least do some performance tests before merging this

src/engraving/layout/layoutsystem.cpp
```diff
@@ -1505,3 +1510,102 @@ void LayoutSystem::restoreTies(System* system)
...
+{
+    const Fraction minTicks = system->minSysTicks();
+    const Fraction maxTicks = system->maxSysTicks();
+    static constexpr double step = 0.2;
+    static constexpr double squeezeLimit = 0.3;
```
RomanPudashkin, line 1519: Please add a comment explaining why these constants have these values

mike-spa:
@RomanPudashkin yeah, I know it's extremely inefficient, but it's the only way to squeeze everything while still producing an acceptable engraving result. The good news is that `LayoutSystem::manageNarrowSpacing()` will _never_ be called by the default layout. The conditions for calling this function only happen if the user specifically requires to fit in the system more measures than can normally fit (again, also quite rare). In fact, I've tried the usual Beethoven 9 score and this function never gets called, so the final layout time is even slightly better than current master:

Master:
master

PR:
thisPR

Besides that, I've removed the calls to minSysTicks and maxSysTicks cause I've realized that the values could be passed down by the caller. And I've added comments on the constants.

RomanPudashkin (approved):

abariska (approved):

mike-spa:
Ok, I don't understand why the vtests seem to get stuck: I can run them locally on my machine (Windows) and they run fine. I'll try on my Linux machine and see if something changes.

mike-spa:
Update on vtests: they run perfectly fine both on my Windows machine and on my Linux machine. So I really don't know what to do with them here

igorkorsukov:
@mike-spa Please, disable (comment out) the drawdata steps so they don't block you (they are in alpha mode now), I'll see what's wrong later https://github.com/musescore/MuseScore/blob/master/.github/workflows/ci_vtests.yml#L153

mike-spa:
Ok, I've removed the problematic bit from the vtests and rebased it, let's hope it works 👍 (btw the current master crashes on startup)

RomanPudashkin (approved):

PR #15205, opened 2022-12-15
Fix #15166: Ignore StemDirection element for drum staves on pre-4.0 import

### LINKED ISSUE #15166: [MU4 Issue] Broken stems in drum score
**Describe the bug**
In a drum score, trying to write a) bass drum, pedal hi-hat or high floor tom and b) other parts of the drum set in one voice while otherwise using MuseScore defaults (stem directions downward for bass drum, pedal hi-hat note, high floor tom) results in broken stems. This happens both when entering notes in a newly created drum score in MuseScore 4 and when opening a drum score created in MuseScore 3 in MuseScore 4.

By default, MuseScore's drum input places bass drum, pedal hi-hat and high floor tom in the second voice but in drum notation it is also common to place all parts of the drum set in one voice.

**To Reproduce**
1. Create a drum score in MuseScore 4.
2. Write consecutive beamed notes (16ths, quavers, …) of bass drum and snare drum hits with both placed in the same voice.

or
1. Open the attached score in MuseScore 4: [CultOfPersonality-IntroFill.zip](https://github.com/musescore/MuseScore/files/10228439/CultOfPersonality-IntroFill.zip)
2. See the broken stems in bars 1, 3 and 5.

**Expected behavior**
No broken stems

**Screenshots**
MuseScore 3: MuseScore3
MuseScore 4: MuseScore4

**Platform information**
 - OS: macOS 11.7.2, final release of MuseScore 4

asattely:
Resolves: #15166 

If you look in the 3.6 save file provided in the issue, the stem directions are wrong. And indeed, if you open the score in 3.6 and look at the stem directions for each note, you can see that they are inaccurate. The reason it looks good in 3.6 is that 3.6 drum staves were incapable of rendering custom stem directions on drum staves.

So, the solution as I see it is to ignore `StemDirection` for drum staves when importing pre-4.0 scores, as it is essentially junk information.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -1256,7 +1256,9 @@ void Chord::read(XmlReader& e)
 bool Chord::readProperties(XmlReader& e)
 {
     const AsciiStringView tag(e.name());
-
+    // issue #15166: MU3 saves wildly incorrect stem directions for drum staves, so we need to ignore them
+    // on import
+    bool readStemDirection = !staff()->staffType()->isDrumStaff() || score()->mscVersion() >= 400;
     if (tag == "Note") {
         Note* note = Factory::createNote(this);
         // the note needs to know the properties of the track it belongs to
@@ -1301,7 +1303,7 @@ bool Chord::readProperties(XmlReader& e)
         StemSlash* ss = Factory::createStemSlash(this);
         ss->read(e);
         add(ss);
-    } else if (readProperty(tag, e, Pid::STEM_DIRECTION)) {
+    } else if (readStemDirection && readProperty(tag, e, Pid::STEM_DIRECTION)) {
     } else if (tag == "noStem") {
         _noStem = e.readInt();
     } else if (tag == "Arpeggio") {
@@ -4169,7 +4171,7 @@ void GraceNotesGroup::layout()
         Chord* grace = this->at(i);
         Shape graceShape = grace->shape();
         Shape groupShape = _shape;
-        mu::remove_if(groupShape, [grace](ShapeElement& s){
+        mu::remove_if(groupShape, [grace](ShapeElement& s) {
             if (!s.toItem || (s.toItem->isStem() && s.toItem->vStaffIdx() != grace->vStaffIdx())) {
                 return true;
             }


MarcSabatella:
I think there is more to it than that though.  The issue report mentions you can always create this same error condition from scratch in MuseScore 4.  It doesn't provide the steps to do so, but I managed to trigger the issue as follows:

1) add drumset instrument
2) note input mode
3) select eighth note duration
4) click bass drum icon in drum palette
5) click voice 1 button on toolbar
6) click in score to enter bass drum into voice 1
7) select snare drum icon on toolbar
8) click in score to enter snare drum

Result:

image

The problem happens because the bass drum has a hard-coded stem direction (sensible because it's normally in voice 2) and this is retained even when forcing the note to appear in voice 1.  And as mentioned in the issue, there are people who notate drum parts that way at times.

So, somehow, this should resolved so that it isn't a problem in new scores either.

RobFog:
Thanks, Marc.

Adding to that: A solution that might come to mind is to change the default stem direction for bass drum to upwards when we place it in voice 1. But I'm not sure that's the right thing to do because we probably don't want the stem to go up when we write a non-beamed bass drum note, for example a quarter note?

MarcSabatella:
That depends on your own personal style of notation, of course.  I wouldn't want my bass drum stems up ever.

Also, we need to do the right thing (whatever that means) if the bass drum is first *added* in voice 2 and then *changed* to voice 1 later.  Right now that seems to work OK in the case I just tried, but perhaps there is some particular sequence of operations that triggers the same bug.

RobFog:
You'll find that most drum transcriptions you find online from reputable sources today write bass drum stems up when transcribing grooves. See for example:
* Drumeo ([website](https://www.drumeo.com)), probably the world's largest online drum teaching website: https://drumeoblog.s3.amazonaws.com/beat/wp-content/uploads/2022/06/10120437/Foo-Fighters-Everlong-1.pdf
* Tim Buell ([website](https://www.timbuellmusic.com)), who transcribes for Zildjian among others: https://timbuellmusicpublicdownloads.s3.us-east-2.amazonaws.com/2019/VFjamsLIVE-MasterFile.pdf

This way of writing drum scores is also mentioned as a legitimate variant in Norman Weinberg's _Guide to Standardized Drumset Notation_.

PR #15207, opened 2022-12-15
Issue templates: add link to Muse Hub issue tracker

cbjeukendrup:
Muse Hub issues should now be directed to the dedicated issue tracker. This PR mentions this on the "New issue" page. Once this is merged, the page will look like this: (tested on my own fork)
Scherm­afbeelding 2022-12-15 om 21 18 42

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,4 @@
+contact_links:
+  - name: Muse Hub issue
+    url: https://musehub.zendesk.com/
+    about: All issues to do with installing apps and sounds via Muse Hub.


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,4 @@
+contact_links:
+  - name: Muse Hub issue
+    url: https://musehub.zendesk.com/
+    about: Issues with the Muse Hub application, like installation issues
```
cbjeukendrup, line 4: Not sure what would be the best here; maybe "Issues with the Muse Hub application, rather than MuseScore itself" is better

shoogle: Perhaps this would be clearer:

> **Muse Hub issues**
> All issues to do with installing applications and sounds via Muse Hub.

I think it should be clear that includes other Muse Hub problems, like usability issues, even if they are are not directly related to installing apps and sounds. I'm more worried about non-Hub problems being reported on the Muse Hub tracker (e.g. installation issues not relating to Muse Hub, or issues using Muse Sounds in MuseScore).

cbjeukendrup: Yeah, changed it to your suggestion!

shoogle (approved):

Issue #15237, opened 2022-12-16
[MU4 Issue] Crash when opening (generating) parts in score from "Treble clef" or "Bass clef" template


cbjeukendrup:
**To Reproduce**
1. Create score from Treble Clef or Bass Clef template
2. Open Parts dialog
3. Select the "Piano" part (the only one)
4. Open it
-> CRASH

**Additional context**
Fix coming in a few minutes

abariska:
🟢 **FIXED**
Tested on macOS 12

PR #15238, opened 2022-12-16
Fix crash when extracting parts when master score contains a bracket/barline that has a bigger span than the number of available staves

cbjeukendrup:
Resolves: #15237


mike-spa (approved):
Good catch, thank you!

abariska:
🟢 **TESTED on macOS 12**
Please, merge to master

PR #15270, opened 2022-12-17
Fix #14903 - Remove hidden "S/A", "T/B", "T/L", "B/B" texts from Pale…

### LINKED ISSUE #14903: [MU4 Task] Remove hidden "S/A", "T/B", "T/L", "B/B" texts from Palette > Text > More, now that they don't make sense anymore
**Task description**
In Palettes > Text > More, there are four elements that look somewhat strange when you see them for the first time: "S/A", "T/B", "T/L", "B/B". What makes them even more mysterious, is that they are made invisible by default.

What these elements do: they used the MS3 options for audio routing / channels; these options could namely be activated using staff texts. These elements route different "notational"/graphical voices to different audio channels. For example, "S/A" sends notes from voice 1 and 3 to a Soprano channel, and 2 and 4 to an Alto channel. So these elements were massively useful for condensed scores where the S/A voices share a staff.

However, in MS4, we don't support this kind of channel routing yet, and when we will, it will probably work differently than what we had in MS3. In any case, at this moment, these elements in the Palette are completely useless, and only a possible source of confusion. Therefore I propose to remove them.

rzvncj:
…tte > Text > More

Resolves: #14903


cbjeukendrup (approved):
Looks perfect!

PR #15293, opened 2022-12-18
Init OAuth lazily

cbjeukendrup:
- so that we don't use the resources of having a TCP server running all the time
- so that firewall warnings will only appear at the moment that the user tries to log in, not on every launch. This makes it easier for users to understand that these warnings are related to the login process.

    (This is the dialog I mean: 
    Scherm­afbeelding 2022-12-18 om 02 30 25
"Do you want to allow mscore.app to accept incoming connections?"

    This dialog appears every time that MuseScore starts its TCP server, if macOS's default firewall is enabled)

### DIFF
--- a/src/cloud/internal/cloudservice.cpp
+++ b/src/cloud/internal/cloudservice.cpp
@@ -104,6 +104,26 @@ void CloudService::init()
 {
     TRACEFUNC;
 
+    multiInstancesProvider()->resourceChanged().onReceive(this, [this](const std::string& resourceName) {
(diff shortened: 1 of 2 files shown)


RomanPudashkin (approved):

abariska:
🟢 **TESTED on macOS 12**
Please, merge to master

cbjeukendrup:
See https://musescore.org/en/node/343568: someone reported that their internet breaks when they open MuseScore. It is likely that this is caused by the OAuth "server", so this PR should improve the situation for that.

Issue #15297, opened 2022-12-18
[MU4 Issue] Some of the tempi have learned to fly...


quinnouyang:
https://user-images.githubusercontent.com/90884224/208274968-a47c5225-fc48-4a37-b6e1-87cba80719d9.mov

**To Reproduce**
1. Open the Master Palette
2. Click to Tempo
3. Wave your cursor over the grid. Magic! (Direction does not matter–any movement moves the same group of tempi vertically.)
(This doesn't seem to occur for any other tabs in the Master Palette.)

**Expected behavior**
This should not happen.

**Platform information**
 - OS: macOS Monterey 12.6, Macbook Air M1 with built-in trackpad (if that matters)

cbjeukendrup:
Another way to reproduce the same issue in the regular Palettes:

https://user-images.githubusercontent.com/48658420/208277483-87b270ae-9845-46b0-8583-c3ecb8409f26.mov

On every redraw of the palette cell, the text is moved upwards a little bit.

quinnouyang:
~~Who's going to displace those specific palette options enough times to notice?~~

PR #15300, opened 2022-12-18
Fix flying gradual tempo changes in Palettes / Master Palette

cbjeukendrup:
On every redraw of the palette cell (which included a call to `layout()`) they would be moved upwards a little bit.

Resolves: #15297

@mike-spa Do you think that the positioning of gradual tempo changes needs to be re-tested thoroughly after this change?

### DIFF
--- a/src/engraving/libmscore/gradualtempochange.cpp
+++ b/src/engraving/libmscore/gradualtempochange.cpp
@@ -342,7 +342,7 @@ Sid GradualTempoChangeSegment::getPropertyStyle(Pid id) const
 void GradualTempoChangeSegment::layout()
 {
     TextLineBaseSegment::layout();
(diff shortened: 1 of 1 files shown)


cbjeukendrup:
Hm, yeah, it's indeed problematic for the placement below the staff...

mike-spa:
Yeah, this is gonna need a deeper look. I remember why I did that change, but now that I look at it I'm thinking I should have done it differently. I'll see if I can find a better solution

MarcSabatella:
FWIW, in the past there were a number of layout() functions that had hacks like checking parent to see if this was being laid in a score or on a palette, so things could be special-cased as needed.  Not pretty and I'm not sure if that's still something used elsewhere or not, but this is exactly the sort of case where that could make sense.

cbjeukendrup:
I think I'll close this, as Michele is already looking for a better solution.

Issue #15152, opened 2022-12-13
[MU4 Issue] Crash when adding notehead parentheses second time after Undo


abariska:
The issue could be linked with https://github.com/musescore/MuseScore/issues/15150

**To Reproduce**
1. Create a simple score
2. Select a notehead 
3. Via Properties add parentheses to notehead
4. Apply Undo and deselect the note
5. Select the note again and again add parentheses to notehead via Propeerties

**Expected behavior**
No crash. Parentheses are added again

**Screenshots**

https://user-images.githubusercontent.com/101250347/207383452-e9ca4cae-fd32-4666-bcf1-b63accc358bd.mp4

**Platform information**
 - OS: macOS

abariska:
🟢 **FIXED**
Tested on macOS 12

PR #15303, opened 2022-12-18
Make code for adding/removing parentheses from notes more solid

cbjeukendrup:
Fix crash when doing add parentheses -> undo -> re-add parentheses

Resolves: #15152 

What happened:
- add parentheses -> undo stack creates new "AddElement" action pointing to the parentheses (to which the note is also pointing)
- undo -> the "AddElement" frame is popped from the undo stack, and pushed on the redo stack
- add parentheses again -> since we're doing a new action, the redo stack is cleared, so the actions on the redo stack are deleted, so they perform some cleanup; the "AddElement" action deletes the parentheses it points to. But the note thinks that they still exist, and tries to re-add them by calling `score()->undoAddElement(_leftParenthesis);`. That goes wrong, because they were deleted.

Solution:
Add more robust tracking to the note whether its parentheses are still alive or not. When they are removed (either deleted manually, or by undoing an "AddElement" action, or by changing the "hasHeadParentheses" property) the note sets them to nullptr. So now _leftParenthesis != nullptr if and only if there is _actually_ a left parenthesis.

### DIFF
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -1252,12 +1252,23 @@ void Note::add(EngravingItem* e)
         m_bend = toStretchedBend(e);
     // fallthrough
     case ElementType::FINGERING:
-    case ElementType::SYMBOL:
     case ElementType::IMAGE:
     case ElementType::TEXT:
     case ElementType::BEND:
         _el.push_back(e);
         break;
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

abariska:
🟢 **TESTED on macOS 12**
Please, merge to master

Issue #11570, opened 2022-05-08
Guitar Tab fret numbers have a visible white box behind them


Tantacrul:
**Describe the bug**

<img width="666" alt="image" src="https://user-images.githubusercontent.com/21022311/167304436-782c3863-7005-43cc-bfbc-7e697b15d945.png">
These white boxes need to be set to match the 'Paper colour', which is set in preferences.
<img width="538" alt="image" src="https://user-images.githubusercontent.com/21022311/167304492-3004b0f6-1c0f-4357-9af9-964ed61703e6.png">

Looks really bad if you change the score background colour to anything other than white (by default we don't use white).

Here's how they look when the score is inverted:
<img width="1338" alt="image" src="https://user-images.githubusercontent.com/21022311/167304466-9436345a-e7fb-4e98-9ae8-7f6d3e7973b2.png">

Tantacrul:
🥇 for me!

I guess we'll also need to be careful that by fixing the colour problem, they still appear white when exported as PDF :)

cbjeukendrup:
The real solution would be not to draw a background at all, but make a gap in the staff lines. That way, it would also work correctly when exporting transparent PNGs or SVGs. I believe the reason that it is not done yet that way, is that it's difficult :)

Tantacrul:
@cbjeukendrup - it does seem to have been done for MuseScore 3 though?

<img width="635" alt="image" src="https://user-images.githubusercontent.com/21022311/167712286-5258f564-e99a-4562-b0d3-00a9d819f444.png">

cbjeukendrup:
MU3 uses the workaround of using the background color of the page. But when exporting to transparent images, indeed it does not work:
<img width="511" alt="Schermafbeelding 2022-05-10 om 22 09 56" src="https://user-images.githubusercontent.com/48658420/167713434-3bcd0ba7-6c97-4a0a-8af4-f016ddc72892.png">

Tantacrul:
Better than what MS4 does, at least :)

I do agree with your suggestion for what we should actually do, of course :)

But at the least, we should match MS3.

Tantacrul:
@bkunda / @oktophonie / @mike-spa - it'd be lovely to fix this. It's such a dumb issue that surely has the simplest fix known to man.

- Hack solution: make the squares the same colour as the BG.
- Proper solution: implement a way to 'cut out' or mask out stuff around the numbers. It's a thing we need elsewhere anyway (dynamics on barlines for example)

[...]

PR #15312, opened 2022-12-18
Use page color for fret number background

FigmentBoy:
Resolves: #11570 

Renders the page background color behind tablature numbers instead of just rendering a white rectangle

### DIFF
--- a/src/engraving/internal/engravingconfiguration.h
+++ b/src/engraving/internal/engravingconfiguration.h
@@ -28,6 +28,7 @@
 #include "global/iglobalconfiguration.h"
 #include "ui/iuiconfiguration.h"
 #include "accessibility/iaccessibilityconfiguration.h"
+#include "notation/inotationconfiguration.h"
 #include "importexport/guitarpro/iguitarproconfiguration.h"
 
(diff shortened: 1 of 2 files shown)


FigmentBoy:
Will fix vtest later this evening once I arrive home

src/engraving/internal/engravingconfiguration.h
```diff
@@ -38,6 +39,7 @@ class EngravingConfiguration : public IEngravingConfiguration, public async::Asy
     INJECT(engraving, mu::framework::IGlobalConfiguration, globalConfiguration)
     INJECT(engraving, mu::ui::IUiConfiguration, uiConfiguration)
     INJECT(engraving, mu::accessibility::IAccessibilityConfiguration, accessibilityConfiguration)
+    INJECT(engraving, notation::INotationConfiguration, notationConfiguration)
```
RomanPudashkin, line 42: This is not a good solution. Now we have a circular dependency between the engraving and notation modules

FigmentBoy: I guess I hadn't fully considered that. I'll close this PR for now

Issue #15165, opened 2022-12-14
[MU4 Issue] Crash after specific actions with Instruments


abariska:
**Describe the bug**
Probably related to https://github.com/musescore/MuseScore/issues/15127
Test score [smoke 2.mscz.zip](https://github.com/musescore/MuseScore/files/10258986/smoke.2.mscz.zip)

**To Reproduce**
1. Open the score
2. Choose one instrument and add a stave for it
3. Ctrl+Z
4. Add a linked staff for the same instrument
5. Ctrl+Z
6. Delete the instrument
7. Ctrl+Z

**Expected behavior**
No crash

**Screenshots**

https://user-images.githubusercontent.com/101250347/207611950-34a54269-275d-4d66-bf51-ef5bb4dcff44.mp4

**Platform information**
 - OS: macOS

Harmoniker1:
Can't seem to reproduce this on Windows 11…

cbjeukendrup:
@abariska I think I have a fix; it seems to be caused by the articulations. But to be sure, could you please upload the score so that I can check?

abariska:
> could you please upload the score so that I can check?
@cbjeukendrup added to description

cbjeukendrup:
@abariska Thanks! Indeed my PR turns out to fix it for this score too!

abariska:
🟢 **FIXED**
Tested on macOS 12

PR #15351, opened 2022-12-18
Fix crash when doing Add Linked Staff / Undo / DeleteInstrument / Undo on staff with articulation

cbjeukendrup:
Fix crash when performing the following actions on a staff containing a chord with an articulation:

1. Create a linked staff
2. Undo
3. Delete the instrument
4. Undo

Resolves: #15165 

What happened, as far as I understand:
- Create linked staff → a new articulation is created for the linked staff, and it is linked to the original articulation
- Undo → the linked articulation is removed from the score and continues to live in the undo stack, but it is not properly unlinked from the original articulation 
- Delete the instrument → the because of doing a new action, the redo stack is cleared and the actions that were in it are cleaned up; the linked articulation is deleted, but the original one still contains a link pointing to it
- Undo → while restoring the instrument, the articulation is restored too, and somehow this causes us to do something with the linked articulation that was deleted, so we're working with freed memory. This caused the crash. 

Solution: when creating a linked staff, link the linked articulation in an undoable way (like we already do for the chord), so that the link is properly removed when undoing, and the original articulation has no pointers to freed memory.

### DIFF
--- a/src/engraving/libmscore/chord.cpp
+++ b/src/engraving/libmscore/chord.cpp
@@ -295,7 +295,7 @@ Chord::Chord(const Chord& c, bool link)
     for (Articulation* a : c._articulations) {      // make deep copy
         Articulation* na = new Articulation(*a);
         if (link) {
-            na->linkTo(a);
+            score()->undo(new Link(na, a));
         }
         na->setParent(this);
         na->setTrack(track());


RomanPudashkin (approved):

abariska:
🟢 **TESTED on macOS 12**
Please, merge to master

Issue #15323, opened 2022-12-18
[MU4 Issue] "Save as default style" not working for certain text properties


rgreen5:
**Describe the bug**

When you alter a property value in the Properties panel and press "Save as default style for this score" this should update the Text Styles menu (Format > Styles > Text Styles) and should also update all the text instances in the score.

However this does not seem to be happening for all text properties.

Related to #12480?

**To Reproduce**

Open the attached file: 
[text_style_update_issue.zip](https://github.com/musescore/MuseScore/files/10253679/text_style_update_issue.zip)

1. Select the text in the first _text_ frame ("This is the first movement"). It is styled as Frame text.
2. Change the font-size and press "Save as default style for this score" to update the styles menu.
3. Check the Text styles menu and all instances of the same type of text.

RESULT: All OK. Everything has been updated as expected.

4. Now change the horizontal alignment setting to "Center" and "Save as default style for this score".

RESULT: Neither the Text Styles menu nor other instances of frame text have been updated to the new alignmnet.

**Expected behavior**

Updating the style default should work for ALL text properties.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223520506, revision: github-musescore-musescore-5485621

PR #15376, opened 2022-12-19
fix GH#15323: Align properties can't save to style

Harmoniker1:
Resolves: #15323 

To update the Align style value properly, it should be set as a `QVariantList` of 2 variants, but the default call to `updateStyleValue()` only gives it a single `QVariant`, which means an assertion failure is triggered when attempting to treat this `QVariant` as a `QVariantList` in `PropertyValue::fromQVariant()`, returning a `PropertyValue` of undefined type.

In this commit I added an optional parameter to `buildPropertyItem()` and `initPropertyItem()` that can be used to specify the style update callback, just like the existing property update callback parameter, so that `updateStyleValue()` can be given a customised `newValue` when needed.

### DIFF
--- a/src/inspector/models/abstractinspectormodel.h
+++ b/src/inspector/models/abstractinspectormodel.h
@@ -165,7 +165,8 @@ protected:
     void setElementType(mu::engraving::ElementType type);
 
     PropertyItem* buildPropertyItem(const mu::engraving::Pid& pid, std::function<void(const mu::engraving::Pid propertyId,
-                                                                                      const QVariant& newValue)> onPropertyChangedCallBack = nullptr);
+                                                                                      const QVariant& newValue)> onPropertyChangedCallBack = nullptr, std::function<void(const mu::engraving::Sid styleId,
+                                                                                                                                                                         const QVariant& newValue)> onStyleChangedCallBack = nullptr);
     PointFPropertyItem* buildPointFPropertyItem(const mu::engraving::Pid& pid, std::function<void(const mu::engraving::Pid propertyId,
                                                                                                   const QVariant& newValue)> onPropertyChangedCallBack = nullptr);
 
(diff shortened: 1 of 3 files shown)


src/inspector/models/abstractinspectormodel.h
```diff
@@ -165,7 +165,8 @@ public slots:
     void setElementType(mu::engraving::ElementType type);
 
     PropertyItem* buildPropertyItem(const mu::engraving::Pid& pid, std::function<void(const mu::engraving::Pid propertyId,
-                                                                                      const QVariant& newValue)> onPropertyChangedCallBack = nullptr);
+                                                                                      const QVariant& newValue)> onPropertyChangedCallBack = nullptr, std::function<void(const mu::engraving::Sid styleId,
+                                                                                                                                                                         const QVariant& newValue)> onStyleChangedCallBack = nullptr);
```
cbjeukendrup, line 169: Poor Uncrustify... 😅

cbjeukendrup (approved):

src/inspector/models/abstractinspectormodel.h
Harmoniker1: Ikr :p

DmitryArefiev:
Tested #15323 on Win10, Mac12, LinuxMint 22.04 - **FIXED**

DmitryArefiev:
@cbjeukendrup Is this reviewed and ready to be merged? (just don't know what the comments about the code above mean :)

cbjeukendrup:
Yes, it's ready! 

(Uncrustify is the tool that we use to check and fix the code style. But this tool is... not always perfect.)

DmitryArefiev:
Oh cool!

MarcSabatella:
I see that a number of other Properties panel bugs are fixed for 4.0.2 via https://github.com/musescore/MuseScore/pull/16038 - would it be possible for this one to be included as well?

Issue #15358, opened 2022-12-19
[MU4 Issue] Increasing duration of a measure creates corruption


MarcSabatella:
**Describe the bug**
Increasing the duration of a measure creates a corrupt measure - the extra beats are not filled with rests, but a rest is placed in the *next* measure

**To Reproduce**
1. Create new score, 4/4
2. Fill a measure with quarter notes
3. Measure Properties
4. Change Actual Duration to 5/4
5. OK
6. See error: still only four beats showing in the measure, but an errant quarter rest appears in the next, replacing the measure rest

**Expected behavior**
The measure selected should not have five beats, and the next measure should be unchanged

**Screenshots**
![image](https://user-images.githubusercontent.com/1799936/208349414-99b7f335-e92b-45d1-ae67-03c90650519d.png)

**Platform information**
 - OS: Linux

**Additional context**
Not sure how this could have slipped by, but it's hugely critical.

eitanmuir:
Yes! A number of issues with time signature changes. I have a project where this keeps corrupting my files - see the 'ghost' two bars at the end of all parts (but not in the full score)

Windows 10 Pro 22H2 build 19045.2364

[Eitan Muir - Gamelan Piece 3 [2022-12-19].zip](https://github.com/musescore/MuseScore/files/10257086/Eitan.Muir.-.Gamelan.Piece.3.2022-12-19.zip)

PR #15380, opened 2022-12-19
Fix corruption when changing measure length

cbjeukendrup:
Resolves: #15358

Was introduced in f25a94103a81607c24af5df270c96d18fcc81623


cbjeukendrup:
A slightly more ideal solution would be to decouple fixing the ticks from setting up the tempo map. Fixing the ticks must often be done immediately, and must happen in all parts, while setting up the tempo map can absolutely be delayed until the end of the layout process and only needs to happen in the master score. Also, when simply changing the tempo, fixing the ticks is not necessary.

Issue #15298, opened 2022-12-18
[notation issue] beam direction ignoring default, disappearing stems


Vykoiazon:
**Describe the bug**
On a percussion staff, sometimes beam direction got reversed for unknown reasons and also made the stems disappear.

**To Reproduce**
1. on a single line percussion staff (like bass drum), add a note
3. press ctrl + 3/4/5 to make it into a tuplet
4. see the beam direction flip and if more notes are added to the rest of the tuplet, the stems disappear
5. this can also happen for non tuplets but I do not know how to reproduce that (example in screenshot)

**Expected behavior**
On a single line percussion staff (like bass drum), beam direction is defaulted to down so all notes including tuplets should also have beams facing downwards.

**Screenshots**
![image](https://user-images.githubusercontent.com/120861729/208274728-8930188b-3730-488d-b1d3-4d0f701ff67e.png)

**Platform information**
windows

**Additional context**
It also happened in musescore 3 but the stems did not disappear before.

oktophonie:
This is very similar to https://github.com/musescore/MuseScore/issues/15166 and we shuld therefore ensure it is also fixed by https://github.com/musescore/MuseScore/pull/15205

eitanmuir:
I have also noticed this on Percussion staves and staves with less than the standard 5 lines

PR #15383, opened 2022-12-19
Fix #15166: Only use custom stem directions when stem is unbeamed

### LINKED ISSUE #15166: [MU4 Issue] Broken stems in drum score

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/15166
A better fix for this one (than PR https://github.com/musescore/MuseScore/pull/15205, which treats it as an import issue) is to just ignore explicit stem direction if the chord is beamed. This will allow stem directions to flip when the beam is manually adjusted, as well as weird drum situations where stem directions are hardcoded even when beamed in a certain direction.

(partially) Resolves: https://github.com/musescore/MuseScore/issues/15298
There is an unrelated issue where converting a chord from a normal chord to a tuplet resets its stem direction to `AUTO`, which is not solved by this PR. However, when other chords are added to the tuplet in a percussion staff, the added chords have explicit stem direction so the beam will use that. Basically, when determining the direction of the beam, it looks through all the chords to see if there are any explicit stem directions, and if so, uses that.

Chords not maintaining their explicit stem direction when tupletified is fixed by https://github.com/musescore/MuseScore/pull/15394.

Resolves: https://github.com/musescore/MuseScore/issues/15401
Determining stem direction for notes within beams was thrown off when the beam both started and ended with a rest. Now the direction calculation for the notes within a beam match the calculation for the beam itself--the first and last chords of a beam are determined by ignoring the rests and using the first and last actual chords of a beam


oktophonie (approved):

asattely:
> Does this relate to https://musescore.org/en/node/264750 ?
It does indeed. As long as the chords have explicit stem direction, the beam will abide by that.

oktophonie (approved):

Eism:
@asattely 
Please fix utests

asattely:
Rebased, utests good!

src/engraving/libmscore/beam.cpp
```diff
@@ -318,11 +318,22 @@ void Beam::layout1()
...
+            // search through the beam for the first chord with explicit stem direction and use that.
+            // if there is no explicit stem direction, default to the direction of the first stem.
+            bool firstUp = false;
+            bool firstChord = true;
+            DirectionV explicitDirection = DirectionV::AUTO;
             for (ChordRest* cr :_elements) {
                 if (cr->isChord()) {
-                    _up = cr->up();
-                    break;
+                    if (toChord(cr)->stemDirection() != DirectionV::AUTO) {
```
RomanPudashkin, line 328: Please, never call the same operations multiple times even if they are relatively fast to perform. Call this stemDirection() once, write its result to a temp variable, and reuse it

RomanPudashkin:
> Could you please add some utests for Chord::computeUp()?

@asattely Any progress with this?

RomanPudashkin:
> I feel that the functionality is better tested via a vtest, since it just loads the file and makes sure the stem directions laid out correctly based on the new computeUp functionality.

Could you please rebase your PR? We will see if the vtests show the difference. If they are all "green" after rebasing, then we should add more tests to cover all possible cases

oktophonie (approved):

RomanPudashkin (approved):

Issue #15368, opened 2022-12-19
[MU4 Issue] Regression: Multi-measure rests break "hide when empty"


jeetee:
**Describe the bug**
A multi-measure rest is not considered "empty" for the "Hide when empty" logic

**To Reproduce**
1. Open https://musescore.org/sites/musescore.org/files/2022-12/Nur%20ein%20Wort.mscz
2. Notice how on the first and last system, all staves are shown even though "hide when empty" style setting (and first system subsetting too) are active.
3. Press "M" to toggle multimeasure rests off
4. Empty staves are now hidden (as expected)
5. Press "M" to toggle multimeasure rests back on
6. Error: Empty staves are made visible again

**Expected behavior**
A multimeasure rest condensed measure should be regarded as equally empty as when expanded.

**Platform information**
 - OS: Verified on Windows 10, but platform agnostic

**Additional context**
This is a regression compared to MS3.6.2

oMrSmith:
(I confirm.)

PR #15418, opened 2022-12-20
Correction to staff hiding logic

mike-spa:
Resolves: #15368


RomanPudashkin (approved):

mike-spa:
@Tantacrul @vpereverzev Would it be ok to include this in 4.0.2? I think the fix is quite low risk, and the bug can cause a pretty serious problem with staff hiding that has no workaround (see video), which is getting quite a few complaints.

https://user-images.githubusercontent.com/93707756/214044260-9192fb6b-d5ae-44cb-a495-2427923fd6b2.mp4

PR #15428, opened 2022-12-20
In textLineSegment layout, align line to text, not text to line

mike-spa:
Resolves: #15297 

The issue was caused by [this line](https://github.com/musescore/MuseScore/blob/master/src/engraving/libmscore/gradualtempochange.cpp). That was a quick solution I made to fix the undesirable baseline difference shown here:
image

However, that baseline difference is caused by the fact that we are moving the text down by half-line width to match the baseline to the following line. The better solution (which I admit may sound a bit academic) is that we should align the line to the text, not the text to the line. This way the text isn't offset and that original correction I did isn't needed anymore. 

The vtests will show small differences (all due to a half-line-width offset).

### DIFF
--- a/src/engraving/libmscore/textlinebase.cpp
+++ b/src/engraving/libmscore/textlinebase.cpp
@@ -306,18 +306,22 @@ void TextLineBaseSegment::layout()
         _offset2.setY(0);
     }
 
-    auto alignText = [tl](Text* text) {
+    auto alignBaseLine = [tl](Text* text, PointF& pp1, PointF& pp2) {
+        PointF widthCorrection(0.0, tl->lineWidth() / 2);
         switch (text->align().vertical) {
         case AlignV::TOP:
-            text->movePosY(-tl->lineWidth() / 2);
+            pp1 += widthCorrection;
+            pp2 += widthCorrection;
             break;
         case AlignV::VCENTER:
             break;
         case AlignV::BOTTOM:
-            text->movePosY(tl->lineWidth() / 2);
+            pp1 -= widthCorrection;
+            pp2 -= widthCorrection;
             break;
         case AlignV::BASELINE:
-            text->movePosY(tl->lineWidth() / 2);
+            pp1 -= widthCorrection;
+            pp2 -= widthCorrection;
             break;
         }
     };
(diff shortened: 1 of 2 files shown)


src/engraving/libmscore/textlinebase.cpp
```diff
@@ -306,18 +306,22 @@ void TextLineBaseSegment::layout()
         _offset2.setY(0);
     }
 
-    auto alignText = [tl](Text* text) {
+    auto alignBaseLine = [tl](Text* text, PointF& pp1, PointF& pp2) {
+        PointF widthCorrection(0.0, tl->lineWidth() / 2);
```
cbjeukendrup, line 310: I'm not sure whether this takes into account whether the line is visible or not (`textLineBase()->lineVisible()`)

cbjeukendrup:
@mike-spa This seems to cause a regression for the alignment of the end text if that is different from the alignment of the start text, albeit that this is a somewhat hypothetical situation.

Before:
Scherm­afbeelding 2023-01-30 om 18 41 26

After: 
Scherm­afbeelding 2023-01-30 om 18 43 08

Since it is so hypothetical, is that a problem?

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -451,6 +447,15 @@ void TextLineBaseSegment::layout()
     if (tl->lineVisible() || !score()->printing()) {
         pp1 = PointF(l, 0.0);
 
+        // Make sure baseline of text and line are propertly align (accounting fot line thickness)
+        if (isSingleBeginType() && (tl->beginTextPlace() == TextPlace::LEFT || tl->beginTextPlace() == TextPlace::AUTO)
```
RomanPudashkin, line 451: This code is impossible to understand/debug. It also has a lot of duplicated calls (e.g., isSingleBeginType() and tl->beginTextPlace())

RomanPudashkin:
@mike-spa any updates on Casper's comments?

src/engraving/libmscore/textlinebase.cpp
mike-spa: Ok I'll try to improve it, thanks

mike-spa:
> @mike-spa any updates on Casper's comments?

For the first comment, I'll look into it. For the second one, I can't think of any practical case where that could be an issue, though I'll leave it to @oktophonie to say the final word

src/engraving/libmscore/textlinebase.cpp
mike-spa: If the line is not visible the function will not be called, as expected, because of [this line](https://github.com/musescore/MuseScore/pull/15428/files#diff-6db3e5aff93698409228e959022f9efb004707d57d265492950b4a7ff5af9da6R447)

src/engraving/libmscore/textlinebase.cpp
mike-spa: Done, should be much more readable now 👍

oktophonie:
I can't say the hypothetical issue concerns me much.

RomanPudashkin (approved):

PR #15448, opened 2022-12-20
Fix #15422: Treat grace note beams as small

### LINKED ISSUE #15422: [MU4 Issue] Grace note beam direction is sometimes off
**Describe the bug**
The beam should be heading downwards rather than upwards. In this case it only happens when the grace notes are 32nd or shorter
image

**Platform information**
 - OS: Windows 11

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/15422

For some reason, grace note beams have a `mag()` of 1.0, which is wild considering the beam actually is not full-size. To fix this, I just treat grace note beams the same way we'd treat any other beam with a mag of less than 1.

Note that this fix only checks the first chord of the beam, since grace notes and regular notes can't be beamed together, so if the first chord is grace then the entire beam is.


ecstrema:
I believe that would be worth a vtest

oktophonie (approved):

RomanPudashkin:
@asattely please pay attention to the failed utests

RomanPudashkin (approved):

RomanPudashkin (approved):

Issue #15093, opened 2022-12-10
Melisma remains after changing note value


Aurivox:
If you have lyrics with melisma and have changed the values, the melisma remains. In MS 3 you were able to remove them by editing the lyrics and type the space bar. In MS 4 you have to delete the lyric

(Current MuseHub OS: macOS 13.0, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223362133, revision: github-musescore-musescore-d050fb0)
![Bildschirm­foto 2022-12-10 um 09 25 50](https://user-images.githubusercontent.com/103463374/206841328-dfd71371-ee69-4346-93a5-d1f2065264a1.jpg)

oktophonie:
Seems logical to me that if a melisma is reduced to one note then the line should just disappear:
![image](https://user-images.githubusercontent.com/20604032/206851838-11098d59-1554-4f8f-b991-713139b68d3b.png)
_b)_ and _c)_ already work as described here so only _a)_ is the problem.
(This is not a regression from 3.6, but the spacebar behaviour is different, as noted.)

Whether the syllable should automatically become centered when the melisma is removed is a question.

Aurivox:
In MS 3 it's the same behavior with the underline
![Bildschirm­foto 2022-12-10 um 16 51 35](https://user-images.githubusercontent.com/103463374/206863761-b39fc3a7-c1e5-44ba-b23b-26ec3db2172a.jpg)

oktophonie:
An issue that may be considered related (or not).

If a syllable after a dash is deleted, the dash remains (it should go), and also the __previous__ syllable, if it was centred, becomes left-aligned (it shouldn't):
![image](https://user-images.githubusercontent.com/20604032/209385895-ff4f3603-c715-432a-ba6b-cb69c084e1f8.png)
![image](https://user-images.githubusercontent.com/20604032/209385957-f1131b29-6c41-4db2-9b28-cb942c927452.png)

Jojo-Schmitz:
Came up again in the German forum, see https://musescore.org/de/node/343537
Very anoying to have to remove the syllable in order to be able to dget rid of the melisma

PR #15459, opened 2022-12-20
Fix #15407: Hide zero-length lyrics dashes

### LINKED ISSUE #15407: [MU4 Issue] Lyrics layout breaks on incomplete input
**Describe the bug**
Lyrics self-intersect when ending with a minus (`-`).

**To Reproduce**
1. Enter more than one note.
2. Click on the first note
3. Type Ctrl-L to enter lyrics mode
4. Type "Froh-" without quotes
5. Press the `Esc` key, or click some empty part of the score, to abort entering lyrics

**Expected behavior**
I wouldn't be surprised by any of the following behaviors:
1. Because the minus key is supposed to indicate a single connected word and the user refused to write the rest of the word, the minus is removed, and only "Froh" remains.
2. Because the user refused to enter anything further, the lyrics "Froh - " appear.
3. A dialog appears asking the user what they want.
4. Because the input doesn't make sense, MuseScore reinterprets the input as "Froh_".

**Actual behavior**
The layout algorithm seems to think that "Froh" is the only remaining text node, and does the layout accordingly. However, the "-" (minus symbol) also appears, in a place that would be reasonable if "Froh" hadn't moved in the meantime. The result is that "Froh" and "-" overlap awkwardly.

**Screenshots**
I wrote "Fröh-", but that doesn't really make a difference:
Bildschirmfoto_2022-12-20_01-02-35
I didn't edit that screenshot, the weird line in the text is the minus sign.

**Platform information**
 - OS: Linux, German locale, but that probably doesn't matter.

**Additional context**
Am I spamming too much or should I keep going? Writing 8 bug reports in a row starts feeling like I should stop doing that.

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/15407 and https://github.com/musescore/MuseScore/issues/15093

See comment, the solution for this one has changed.

### DIFF
--- a/src/engraving/libmscore/lyrics.cpp
+++ b/src/engraving/libmscore/lyrics.cpp
@@ -159,11 +159,11 @@ bool Lyrics::isMelisma() const
 
     // hyphenated?
     // if so, it is a melisma only if there is no lyric in same verse on next CR
-    if (_syllabic == LyricsSyllabic::BEGIN || _syllabic == LyricsSyllabic::MIDDLE) {
+    if (_separator && (_syllabic == LyricsSyllabic::BEGIN || _syllabic == LyricsSyllabic::MIDDLE)) {
         // find next CR on same track and check for existence of lyric in same verse
-        ChordRest* cr  = chordRest();
+        ChordRest* cr = chordRest();
         if (cr) {
-            Segment* s     = cr->segment()->next1();
+            Segment* s = cr->segment()->next1();
             ChordRest* ncr = s ? s->nextChordRest(cr->track()) : 0;
             if (ncr && !ncr->lyrics(_no, placement())) {
                 return true;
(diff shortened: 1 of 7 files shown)


asattely:
This one ended up being a little more in-depth than I expected it to be.

- If the user places exactly one underscore (or dash) and then ends editing, the melisma is deleted and the alignment of the lyric is reset.
- If the user places multiple dashes, the lyric will be left-justified (as it spans multiple notes), but if the editing ends with that string of dashes, the original lyric will again have its alignment reset and the "melisma" will be deleted.

oktophonie (approved):

src/engraving/libmscore/lyrics.cpp
```diff
@@ -766,4 +792,16 @@ KerningType Lyrics::doComputeKerningType(const EngravingItem* nextItem) const
     }
     return KerningType::KERNING;
 }
+
+void Lyrics::removeInvalidSegments()
```
RomanPudashkin, line 796: What is considered an invalid segment? I don't really understand the purpose of this method

src/engraving/libmscore/lyrics.cpp
```diff
@@ -230,6 +230,31 @@ bool Lyrics::isMelisma() const
     return false;
 }
 
+//---------------------------------------------------------
+//   prevLyrics
+//    - find the lyric (if any) before this one (not including lines)
+//    - currently used to determine the first lyric of a melisma
+//---------------------------------------------------------
+
+Lyrics* Lyrics::prevLyrics() const
```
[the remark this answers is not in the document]
asattely: I'm afraid I don't understand this one. You want the lyrics class to be in navigate.h?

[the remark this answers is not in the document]
RomanPudashkin: No, only this function :)

src/engraving/libmscore/lyrics.cpp
asattely: That said, I have added comments to the method!

oktophonie (approved):

src/engraving/libmscore/navigate.cpp
```diff
@@ -965,4 +966,29 @@ EngravingItem* Score::prevElement()
...
     return score()->firstElement();
 }
+
+//---------------------------------------------------------
+//   prevLyrics
+//    - find the lyric (if any) before this one (not including lines)
+//    - currently used to determine the first lyric of a melisma
+//---------------------------------------------------------
+
+Lyrics* Lyrics::prevLyrics() const
```
RomanPudashkin, line 976: This is not exactly what I meant :) You should declare a function in navigate.h (similar to prevChordRest):

`extern Lyrics* prevLyrics(const Lyrics* lyrics);
`

Then use this new function in TextBase::endEdit(EditData& ed):

```
 if (isLyrics()) {
     Lyrics* prevLyrics = prevLyrics(toLyrics(this));
     if (prevLyrics) {
         ...
     }
 }


```

asattely: For the record, why do we want to do this? There are no other Lyrics methods in navigate.cpp, so I am wondering what the point of all of this is?

RomanPudashkin: Because it can iterate implicitly over the whole score. And in general, has low cohesion in this class (it does not depend on anything within this class, and can easily be moved out of it and take it as an argument). We want engraving elements to be as simple as possible

Issue #15314, opened 2022-12-18
[MU4 Issue] Wrong key signature when i add a transposition instrument in this score 


Gai-Luron:
1 - Open this file : [Test Score.zip](https://github.com/musescore/MuseScore/files/10253555/Test.Score.zip)
![image](https://user-images.githubusercontent.com/1835871/208291853-a647eec8-9fd0-4bbf-95c9-eddc2cd662e8.png)

2 - Add new Alto Saxophone 2
![image](https://user-images.githubusercontent.com/1835871/208291864-2c87d558-f6c8-4577-956b-74d14a88080a.png)

3 - Result in 3 Sharps key signature  Instead of 1
![image](https://user-images.githubusercontent.com/1835871/208291873-a3b5d422-f476-4685-85b7-ded2705c91df.png)

OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223510501, revision: github-musescore-musescore-5485621

mike-spa:
We can reproduce the issue only on the file you attached, but not on newly created scores. Now, because of the nature of the problem, if the file was created with any build preceding the fix (which happened in #14849), the issue will be still there because it was saved into the file. It would be great if you can try to grab the latest nightly build from the top of the list [here](https://ftp.osuosl.org/pub/musescore-nightlies/windows/4x/nightly/?C=M;O=D) and let us know if you can reproduce it on a newly created file

oktophonie:
Thanks for this! (I tried it myself now and can reproduce the problem)

Regarding deleting a key signature (step 4) - I think that is intentional, and it's also what 3.6 does - it basically leaves you with an 'open/atonal' key signature, rather than 'C major/A minor'. If you want a different key then you apply it from the palette in the usual way.

Gai-Luron:
Thank's for the answer. Ok for the Atonal mode, but if I put on a key signature again, this atonal mode must disapear, isn't it?. 
The first time i remove the key signature is not intentional.

mike-spa:
@Gai-Luron thanks for your help in figuring this out! The linked PR should fix it

PR #15479, opened 2022-12-21
Fix key sig bug

mike-spa:
Resolves: #15314

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -4657,18 +4657,15 @@ void Score::undoChangeKeySig(Staff* ostaff, const Fraction& tick, KeySigEvent ke
             ks->undoChangeProperty(Pid::GENERATED, false);
             undo(new ChangeKeySig(ks, nkey, ks->showCourtesy()));
         } else {
-            // do not create empty keysig unless custom or atonal
-            if (tick.isNotZero() || nkey.key() != Key::C || nkey.custom() || nkey.isAtonal()) {
-                KeySig* nks = Factory::createKeySig(s);
-                nks->setParent(s);
-                nks->setTrack(track);
-                nks->setKeySigEvent(nkey);
-                undo(new AddElement(nks));
-                if (lks) {
-                    undo(new Link(lks, nks));
-                } else {
-                    lks = nks;
-                }
+            KeySig* nks = Factory::createKeySig(s);
+            nks->setParent(s);
+            nks->setTrack(track);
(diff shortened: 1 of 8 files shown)


sammik:
@mike-spa I dont know, why it was imortant, not to create C (empty) key signature, but I guess, there was some reason.
Are we sure, it is not relevant enymore?
If so, I saw many places in code, where similar code ("not create emptykey sig") is present. Isnt is related with this PR?

https://github.com/musescore/MuseScore/blob/3f9b94d3e54c554b9e046cfb1249fdc0601c2711/src/engraving/libmscore/measure.cpp#L3868-L3873

https://github.com/musescore/MuseScore/blob/3f9b94d3e54c554b9e046cfb1249fdc0601c2711/src/engraving/libmscore/edit.cpp#L4652-L4653

https://github.com/musescore/MuseScore/blob/3f9b94d3e54c554b9e046cfb1249fdc0601c2711/src/engraving/rw/measurerw.cpp#L408-L413

And possibly many others.

MarcSabatella:
I can provide a little historical background that might be relevant.  There was a conscious design decision many years ago to represent C major by the absence of a key signature based on a general feeling that it was bad to have elements in the score that showed no visible trace of their existence.  I personally never shared that feeling, but did what I could to make things work despite any number of complications that design caused.  So some of those comments are likely mine, as is some of the code, but I'd be very happy to see it simplified.

There is nothing wrong in my book with C major being represented by an explicit element.  I don't even think the code actively *requires* C major to be handle with no key signature, as there are still ways you can end up with a C major key signatures element in the score.  But they do tend to get cleaned up on save/reload, so you'd want to make sure that isn't still happening.

sammik:
Same fix is part of PR #12560 which fixes issue #12506

mike-spa:
Thanks both for providing context! To me, eliminating the keySig event from a staff just because it isn't graphically shown is a profound mistake. The fact that the key isn't shown is purely incidental, and managing the showing/not showing just requires some very simple logic. Eliminating it altogether, on the contrary, effectively deletes a fundamental piece of information from our score representation that can't be reconstructed, and it keeps causing bugs and problems everywhere. 
The clearest example is the problems I fixed in #14849. There is a fundamental difference between global key signatures (i.e. the ones that apply to all staves) and local key signatures (i.e. the ones that apply only to individual staves). The first need to be repeated in all extracted parts, regardless of transposition, the latter doesn't. Problem is: if your score has instruments with different transposition and some of them happen to be in C,

oktophonie (approved):

Issue #15226, opened 2022-12-16
[MU4 Task] New dialogs for corruption warnings


bkunda:
**Task description**
This task would resolve the issues in #11807 

There are two variations of the same dialog to be implemented.

1. When the user attempts to open a corrupted file, and the file _can still be opened_

<img width="527" alt="Group 198" src="https://user-images.githubusercontent.com/86290556/211331237-a8ff4bb5-d4c7-4df6-8b77-da7133dbfb0d.png">

<img width="527" alt="Expanded" src="https://user-images.githubusercontent.com/86290556/211331268-2707d398-ad4f-4903-a8fa-a6ca1e6d4c31.png">

<img width="527" alt="Expanded1" src="https://user-images.githubusercontent.com/86290556/211331289-fff5b732-e85b-4713-ab57-c882121c56bd.png">

2. When the user attempts to open a corrupted file, and the file _cannot be opened_

<img width="527" alt="Group 1598" src="https://user-images.githubusercontent.com/86290556/208117493-c9fa2928-b688-4d97-b8a2-89843e40c8b5.png">

<img width="527" alt="Expanded" src="https://user-images.githubusercontent.com/86290556/208117510-7aec00fc-be8c-4515-8a33-d8e5ccf5e14a.png">

(In this case, **Get help** redirect to [the musescore.org support forum](https://musescore.org/en/forum/6)).

Full UX specification has been sent to @RomanPudashkin for implementation

@Tantacrul @oktophonie @abariska FYI

eitanmuir:
Noting that in Windows 10 there is no option to Show Details - please add this feature too as restarting a whole score is a bit of a headache!

cbjeukendrup:
@bkunda One thought about this design: it feels a little bit random to me that this dialog mentions musescore.com. The fact that uploading might not work properly is just one small aspect. Something like the following would be more complete and informative:

> **File "The_Avengers_Theme_Song_full_orchestra.mscz" is
> corrupted and may contain errors**
> 
> This can cause MuseScore to malfunction and might lead to crashes. 

(You can probably find a nicer wording for this.)

If we really want, we could add "In particular, saving to musescore.com may fail unless these errors are resolved."

sam-the-programmer:
We need a show details feature on Musescore 4 when a file is corrupted.

cbjeukendrup:
[...]

Issue #14560, opened 2022-11-16
[MU4 Issue] No space after dot or comma in UI on macOS


shoogle:
Whenever a UI string contains a dot (period/full stop) followed by a space, or a comma followed by a space, the space is missing in the UI.

**Screenshots**

<img width="700px" src="https://user-images.githubusercontent.com/11011881/202306710-54eb723c-3497-40ae-b43c-7cbc4bc96699.png"/>

<img width="700px" src="https://user-images.githubusercontent.com/11011881/202307366-d9ed710c-d388-47e3-96ab-8e87abd140f8.png"/>

<img width="700px" src="https://user-images.githubusercontent.com/11011881/202306379-8952d10a-5e1d-4280-af3e-62686867c4ae.png"/>

**Platform information**
 - OS: macOS Big Sur

shoogle:
Closing as this is a [Qt bug](https://bugreports.qt.io/browse/QTBUG-88495) that is fixed in Qt 6. It's also fixed in Qt 5.15.3 but sadly that's [not available for open source](https://www.phoronix.com/news/Qt-5.15.3-LTS-Released), at least [not via the official Qt Online Installer](https://raymii.org/s/blog/Qt_5.15.3_OpenSource_released.html).

cbjeukendrup:
Indeed, this is fixed in 5.15.3 and higher. The latest available open source version is 5.15.7 at the moment. It is provided by HomeBrew, and I use it on a daily basis. If we want, we can easily put that on our S3 server, from where it is downloaded during CI builds.

(HomeBrew even seems to provide an ARM64 version, but I can't get that to work with MuseScore.)

shoogle:
That's something to check with the other developers. It should be safe to upgrade to a patch release, but it's not guaranteed of course. We might prefer to stick with what we are currently using for the time being.

PR #15488, opened 2022-12-21
error_details

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/15226

### DIFF
--- a/src/framework/ui/view/internal/errordetailsmodel.cpp
+++ b/src/framework/ui/view/internal/errordetailsmodel.cpp
@@ -1,0 +1,77 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0-only
+ * MuseScore-CLA-applies
+ *
+ * MuseScore
+ * Music Composition & Notation
+ *
+ * Copyright (C) 2021 MuseScore BVBA and others
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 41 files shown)


bkunda:
@RomanPudashkin I've tested this on macOS using all your supplied scores, as well as the "Avengers" score we were previously using. 
✅ Both "types" of error dialog are being triggered correctly (I.e. the "open anyway" version, and the "get help" version)
✅ "Get help" link points to correct URL
✅ "Show details" text strings are correctly laid-out
✅ "Copy" functionality works
Let us know when the accessibility navigation is implemented and I'll give that one quick look too 🙂

src/framework/ui/view/internal/errordetailsmodel.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "errordetailsmodel.h"
+
+#include <QApplication>
+#include <QClipboard>
+#include <QTextDocument>
+
+#include "translation.h"
```
Eism, line 29: don't need

Eism (approved):

Jojo-Schmitz:
Seems this PR needs thee "Strings" label, as it adds translatable strings?
And wouldn't that make it not fitting 4.0.1?

bkunda:
@RomanPudashkin I've given this another test on macOS – all looks good (and keyboard navigation works well too).
Just one quibbly point: could we please have a space after the punctuation marks at these points? 

Screenshot 2023-01-18 at 11 42 09 am

cbjeukendrup:
@bkunda That's the same problem again as #14560...

bkunda:
> @bkunda That's the same problem again as #14560... (where the only solution is to update to Qt 5.15.3+)

oh god, how droll. 
Ok then @RomanPudashkin this is good to go from my POV!

Issue #15475, opened 2022-12-21
[MU4 Issue] Placing or removing dynamics affects accent placement on stave below it


meesfrensel:
**Describe the bug**
When adding or  removing a dynamic (mf, f, ff etc), accents above the bar below the dynamics marking shift up by about 2cm. They should remain in the (correct) place (where they were).

**Update:** this happens in a few more situations but I haven't been able to put my finger on what exactly the trigger is. For example, when I changed the author text, a few accents jumped around as well.

Let me know if you need any more information!

**To Reproduce**
1. Create a score and add notes to span multiple lines
2. Add an accent
3. Place a dynamics mark on the stave right above
4. Notice that the accents shifted up by about 2cm
5. Issue can be fixed for the whole bar by replacing accent mark

**Expected behavior**
The dynamics marking appeared the way I expected, but I also expected the accents on the bar below to remain in the place they were before

**Screenshots**
Before: 
![image](https://user-images.githubusercontent.com/33722705/208922955-4af25ae7-5660-4d59-809a-fed0f3ab9fd7.png)

After placing forte marking:
![image](https://user-images.githubusercontent.com/33722705/208923019-dfaa457e-d086-4623-9fbe-1ebd0861006a.png)

After removing one of the accents:
![image](https://user-images.githubusercontent.com/33722705/208923190-a452b59c-84a2-4a61-bec9-950d1d5e585b.png)

**Platform information**
 - OS: Windows 10 Version 2009
 - Arch.: x86_64
 - MuseScore version (64-bit): 4.0.0-223472200
 - revision: github-musescore-musescore-5485621

MarcSabatella:
I can reproduce in a simple test score:

1) new score, treble clef templates
2) fill first two systems with accented half notes
3) add an "mp" dynamic to first note of score

Result: the accent on the first half notes of the second system jumps up

![image](https://user-images.githubusercontent.com/1799936/208932626-684592d0-0f20-40a4-a894-5cf42bb1c01b.png)

YuanCQA9826:
In fact, If notes are facing down, the accent will go to move down

PR #15516, opened 2022-12-22
Fix articulation glitch

mike-spa:
Resolves: #15475 

`layoutArticulation2()` wasn't actually needed there

### DIFF
--- a/src/engraving/layout/layoutmeasure.cpp
+++ b/src/engraving/layout/layoutmeasure.cpp
@@ -880,7 +880,6 @@ void LayoutMeasure::computePreSpacingItems(Measure* m)
             isFirstChordInMeasure = false;
 
             chord->layoutArticulations();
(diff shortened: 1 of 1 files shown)


oktophonie (approved):

src/engraving/layout/layoutmeasure.cpp
```diff
@@ -880,7 +880,6 @@ void LayoutMeasure::computePreSpacingItems(Measure* m)
             isFirstChordInMeasure = false;
 
             chord->layoutArticulations();
-            chord->layoutArticulations2();
```
jnhagelberg, line 883: Ideally this should be refactored so you do not need to call two different methods to lay out the articulations.  It seems odd to need to lay them out twice.

RomanPudashkin (approved):

Issue #14546, opened 2022-11-16
[MU4 Task] Optimize LayoutSystem::collectSystem


RomanPudashkin:
**Task description**

Resolves: https://github.com/musescore/MuseScore/issues/11392

This function is much slower in MU4 than in the previous versions of the application. The execution time of it (on my computer) when opening the large score from issue #11392:
<img width="564" alt="Screen Shot 2022-11-16 at 10 36 42" src="https://user-images.githubusercontent.com/67867444/202135484-46620e98-118f-49a4-acc1-a83f8fe56053.png">

For comparison, the MU3 version:
```
Main thread. Top 150 by sum time (total count: 13)
Function                                                    Call time         Call count        Sum time
Score::collectSystem                                        20.179 ms         648               13076.309 ms
Score::layoutSystemElements                                 13.687 ms         556               7609.958 ms
Score::update                                               2397.282 ms       3                 7191.846 ms
Score::getNextMeasure                                       0.491 ms          5326              2615.643 ms
Ms::processLines                                            0.743 ms          3336              2479.808 ms
System::layoutSystem                                        0.531 ms          1112              590.605 ms
```

Thus, in MU3 it is called **2 times more often** when opening the large score, but it wastes much **less time**.  The problem is not only in LayoutSystem::collectSystem, but also in some subfunctions it calls: LayoutSystem::layoutSystemElements, LayoutMeasure::getNextMeasure, LayoutSystem::processLines, etc. They all perform much worse in MU4.

Our task is to find the slowest parts of the code and try to optimize them

RomanPudashkin:
@asattely @mike-spa I ported our profiler to MU3, so you can use my branch to measure performance in MU3. See:
https://github.com/RomanPudashkin/MuseScore/commit/e62c0c21c8d5c6b1180345efafdecc03f42e44a7

mike-spa:
@RomanPudashkin that's awesome, thanks! 
Just for additional context: `LayoutSystem::collectSystem()` is where 99% of all the engraving functions are called from. Anything we do in almost any part of the engraving engine will reflect on this method, so the task could also be renamed to "Optimize engraving engine". 
Now, the engraving engine of MU4 is, in every possible way, more sophisticated than it was in MU3 so, to a certain extent, MU3 wasn't "faster", it was just "simpler". Still, MU4 probably shouldn't be _twice_ as slow, so I think there is still good room for optimization (and it's great that we can use the profiler in MU3 now). We'll work on it!

[...]

PR #15531, opened 2022-12-23
Optimizations and code improvement

mike-spa:
Resolves: #14546 
(More of an incremental improvement, really)

I am collecting in this PR a few optimizations and code improvements I've worked on in the past weeks. Depending on which score this is tested on, overall I see a performance gain of 5-10% in total layout time of the score. Not huge, but not too bad either. Each commit deals with one specific optimization so it probably makes sense to review them separately (@vpereverzev for the code, @oktophonie for the engraving, though it's really just some tiny slur differences).


oktophonie (approved):

Chris-Clawson:
Was this ever merged with trunk? I am an end user and see no improvement in MS 4.01

Jojo-Schmitz:
No, that's why it is still open ;-)

RomanPudashkin (approved):

Issue #15510, opened 2022-12-22
[MU4 Issue] System-wide elements (incl Tempo, rehearsal markings) are gona until restart if a part is made without the first staff of an instrument


randoguyname:
**Describe the bug**
System-wide elements such as Tempo changes and rehearsal marks disappear after making a new part with only the 2nd or 3rd (or both) etc. staff of a multi-staff instrument.
**To Reproduce**
1. Create a new file with the template Great Stave (works with other multi-staff instruments as well)
2. Add a tempo or rehearsal mark element to the score
3. Go to parts
4. Create a new part
5. Show only the bass staff of the piano (this can be done in the instruments panel by showing the piano itself then hiding the treble staff)
6. See the tempo or rehearsal mark diasappear

Then:
Save and reopen the file and see that it now exists on the part
OR
Add a new rehearsal mark or tempo element on the main score and see that it appears in the part.

**Expected behavior**
Such elements are not lost when creating new parts

**Platform information**
 - macOS, Monterey

oktophonie:
Note (for emphasis): only happens when creating a _new_ part, not if you open the 'existing' Piano part from the score.

PR #11260, opened 2022-04-20
[MU4] Fix #11186: Some items get duplicated or don't show between parts and score (when creating a new part/showing instruments)

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/11186

PR #15536, opened 2022-12-23
Fix system flag on parts creation

mike-spa:
Resolves: #15510 

So the bug happens because [here ](https://github.com/musescore/MuseScore/blob/master/src/engraving/libmscore/excerpt.cpp#L1462) we call `Score::undoAddElement()`, inside of which we are switching off the system flag for these objects (on the line I've now deleted). But these _are_ system objects, so they _should_ have the system flag. I worry that this LOC was needed for something else though. @asattely this one probably needs your help.

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -5229,8 +5229,10 @@ void Score::undoAddElement(EngravingItem* element, bool addToLinkedStaves, bool 
         std::vector<Staff* > staffList;
 
         if (!addToLinkedStaves || (ctrlModifier && isSystemLine)) {
-            element->setSystemFlag(false);
             staffList.push_back(element->staff());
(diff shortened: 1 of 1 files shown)


asattely:
That part is there for system objects that have been dropped with the ctrl modifier (voltas, mostly) whose system object flag defaults to `true` but need to be explicitly non-systemobjectified in cases where the ctrl modifier is used.

The `addToLinkedStaves` parameter was added by https://github.com/musescore/MuseScore/pull/11260, which I haven't investigated fully but it is interesting to note that in the calling method you linked it is always calling `undoAddElement` with `addToLinkedStaves` set to `false`. If you have more specific questions I'd be happy to help further!

oktophonie (approved):

RomanPudashkin (approved):

PR #15547, opened 2022-12-24
Fix more possible crashes due to modifying segment annotations during loop

cbjeukendrup:
In #15391, we saw a crash that was caused by modifying the list of annotations of a segment while looping over that list. It turned out that there are a few more cases like this; let's fix them preventively instead waiting until these crashes will indeed occur.

### DIFF
--- a/src/engraving/layout/layoutmeasure.cpp
+++ b/src/engraving/layout/layoutmeasure.cpp
@@ -361,8 +361,8 @@ void LayoutMeasure::createMMRest(const LayoutOptions& options, Score* score, Mea
 
         // remove stray elements (possibly leftover from a previous layout of this mmr)
         // this should not happen since the elements are linked?
-        for (EngravingItem* e : s->annotations()) {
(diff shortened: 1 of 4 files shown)


src/engraving/layout/layoutmeasure.cpp
```diff
@@ -361,8 +361,8 @@ void LayoutMeasure::createMMRest(const LayoutOptions& options, Score* score, Mea
 
         // remove stray elements (possibly leftover from a previous layout of this mmr)
         // this should not happen since the elements are linked?
-        for (EngravingItem* e : s->annotations()) {
-            // look at elements in mmr
+        const auto annotations = s->annotations(); // make a copy since we alter the list
+        for (EngravingItem* e : annotations) { // look at elements in mmr
```
cbjeukendrup, line 365: This one just happened to me, so it's certainly not theory-only :)

RomanPudashkin (approved):

PR #15563, opened 2022-12-24
add "a Tempo" in the Tempo palette

rtbo:
Aims to provide "a Tempo" text mark to reset the tempo after a gradual change. At the moment it is a simple tempo text that defaults to 120 BPM and can be edited to whatever the user wants.

It is already better than present situation, but I'd prefer the item to automatically pick-up the tempo prior to previous gradual change. ~~Don't really know how to achieve this though.~~


randoguyname:
In regards to the comment of automatically picking up the tempo; to the developer who works on this: keep in mind that users will expect that to update automatically by default as well, but also to be able to manually select a tempo. Perhaps by default a checkbox titled "Detect tempo automatically" should be ticked by default, which then updates with changes to the score, and users can uncheck that and do whatever they want.

rtbo:
I added `TEMPO_RESET_PREVIOUS` property on `TempoText` and added corresponding check box in the inspector panel.
Whenever it is checked, the tempo spinbox is disabled and previous explicit (or default) tempo is restored.
When unchecked, the item behaves just like any other `TempoText` item.

image

The checkbox will be there for every `TempoText` so I chose generic text description ("reset previous tempo"). I'm open to suggestion for better name/description.

TODO:
 - add a tooltip on the checkbox
 - add a test
 - french translation

rtbo:
I changed terminology to "restore previous tempo".
Here is with tooltip

image

rtbo:
I've run tests manually of all functions, but it's not obvious what I could do as automated test.
Is it possible to script playback start on a test score and may be check the tempomap?
Or is it OK to not include automated test for this?

rtbo:
Regarding UX, maybe the additional `Restore previous tempo` check-box isn't very valuable.
Isn't it better if the `Follow written tempo` do what's right when the text is `a tempo` (case insensitive).
The fact that `a tempo` is proposed in the tempo palette self-documents the feature in a much better way IMO.

@CombinedAtom @randoguyname @Tantacrul  opinions?

Jojo-Schmitz:
The Frensh translation doesn't belong into this PR

Tantacrul:
> Regarding UX, maybe the additional `Restore previous tempo` check-box isn't very valuable. Isn't it better if the `Follow written tempo` do what's right when the text is `a tempo` (case insensitive)? The fact that `a tempo` is proposed in the tempo palette self-documents the feature in a much better way IMO.
> 
> @CombinedAtom @randoguyname @Tantacrul opinions?

Just tested this. Nice work!

If you mean that the functionality should read the characters 'a tempo' and only then apply the correct logic, I would advise against that. We have generally agreed to avoid trying to deduce intent from written text due to how unreliable it is - and how it needs to localise if a different language is chosen. I think having a checkbox (enabled by default) that you can disable makes more sense. Sorry if I've misunderstood your question.

I actually like 'Restore previous tempo' setting and I think we should get rid of 'follow written tempo'.
- 'Restore previous tempo' already does what we need
- Turning it off allows the user to specify an exact tempo anyway, so all needs are covered

A few points which I think will help the whole thing feel a little easier to understand:

1. I'd place the 'a tempo' palette in the main section, at the bottom of the list. Don't keep it in 'More'. It's a primary option, I feel.
2. Change the text from 'override written tempo' to simply 'set tempo'.

Then I think we're good! Thanks very much.

I'll review again when you're done.

cbjeukendrup:
@Tantacrul If we remove the "follow written tempo" functionality, that also has implications for normal tempo markings (because the new "a tempo" item is in fact just a tempo marking, and therefore it shares its properties widget with normal tempo markings). 

So we need to design a widget that would work well with both "normal" tempo markings and "a tempo" markings. 

How about something like this?
Scherm­afbeelding 2023-03-18 om 14 06 13

(OBVIOUSLY, the "metric modulation" option should not be there yet, but I've added it to convey the point of this design)

Tantacrul:
Can we not just have different options for both things, since the user won't care about what goes on under the hood? I'd rather leave any discussion of tempo markings out of this.

cbjeukendrup:
> I'd rather leave any discussion of tempo markings out of this.

Of course I understand that, but unfortunately it isn't that easy. 

The problem is that MuseScore has no way of distinguishing an "a tempo" marking from any other tempo marking (with this implementation). 

Comparing the text is no option, because who says that the user will call this item "a tempo"? They might also write "Tempo I". So anything about comparing text is not more than a fragile heuristic.
The rigorous solution would be to implement the "a tempo" item as a separate class with a separate item type, but I'm not sure whether it is a good idea to create yet another new class in Libmscore.
Another option would be to add just a `bool isATempo` to the Tempo class, but that's not great either.
(Conclusion: from a technical POV it's not a very nice thing to implement.)

Beside that, from a user point of view, I wonder whether "a tempo" should get a different treatment at all. If it did, then a strange situation would occur when the user changes their mind and changes the text from "a tempo" to an explicit metronome marking. MuseScore would still think that it is an "a tempo" item, because it was initially created as such, and thus it displays the "a tempo" options, while the user thinks that it is just a tempo item, and wonders why the Inspector displays a different set of options than for any other tempo item. Instead of just changing the text, the user would need to remove the "a tempo" item and add a new normal tempo marking. 

So I guess the main question is: would a user intuitively expect that "a tempo" is a different kind of item than any other tempo marking and therefore you can't change one into the other without deleting and re-adding? 
My personal answer is no, I would expect them to be the same kind of thing, only the effect on the playback tempo is slightly different. But that's just my answer.

Tantacrul:
I think the case of someone changing 'A tempo' to a metric marking is an unlikely thing to happen - and there are ways we could make it less likely to happen too.

So, let me narrow my issue with UX in this implementation: we basically have two checkboxes with options that are hard to distinguish from each other.

Worse, it puts an illogical option in all other tempo markings, which I think is a non-starter. This is a classical problem from MS3. 'Restore previous tempo' makes exactly zero sense here:

image

The illogical extra option is not worth the benefit of an 'a tempo' marking.

I think they have to be separate objects.

Tantacrul:
Actually, to be a little clearer: I don't necessarily think they need to be different objects. Just that they need different options to appear in the Properties panel because the existing solution lumps both items with extra options they don't need and the overall comprehensibility of the app takes a hit. 

I think it's a bridge too far, as is unfortunately.

I do appreciate the explanation though @cbjeukendrup.

cbjeukendrup:
Okay. @rtbo I suggest to implement this in the following way:
- Change the new `TEMPO_RESET_PREVIOUS` property to determine whether the tempo marking is an "a tempo" marking (rather than whether the marking does indeed use the previous tempo or that the user has customised the playback tempo)
- Instead, use the `TEMPO_FOLLOW_TEXT` property to determine whether the tempo marking does indeed use the previous tempo or that the user has customised the playback tempo

So:
||`TEMPO_RESET_PREVIOUS` == false|`TEMPO_RESET_PREVIOUS` == true|
|-|-|-|
|`TEMPO_FOLLOW_TEXT` == false | This is a normal tempo marking where the user has overridden the playback tempo | This is an "a tempo" marking where the user has overridden the playback tempo |
|`TEMPO_FOLLOW_TEXT` == true | This is a normal tempo marking where the playback tempo is determined automatically based on what's written in the text of the marking | This is an "a tempo" marking where the playback tempo is determined automatically by looking at the previous tempo marking |

Then, you'll need to create separate Inspector views for "a tempo" and normal markings. To see an example of how you can create different views/models based on the properties of an item, you can start by looking at `HairpinLineSettingsModel` (and the places where it's used), which differentiates based on whether the selected hairpin is a crescendo or decrescendo.

rtbo:
I'm not sure to get the UX right.

What options are to be proposed for the "a tempo" mark?

Either the user can override the tempo to a specific one, in such case the same options as regular tempo apply:
 - "follow text" check box
 - "override tempo" spin button (enabled if "follow text" is unchecked)

Or the user don't have this choice, in which case there is no option to display at all. (`FOLLOW_TEXT` and `RESTORE_PREVIOUS` properties are true and can't be changed on such object)

In either case, I would change the "Tempo" section name by "Restore previous tempo (a tempo)" for clarity for the user.

Tantacrul:
I would have one option

- A checkbox that says 'Set tempo' (rather than 'override tempo', which is less clear).
- If you check the box, then it enables a text field where the specific tempo can be set.

rtbo:
Here is what I have now

image

Ideally I'd like to have the checkbox where the greyed out title is.

src/inspector/view/qml/MuseScore/Inspector/notation/tempos/TempoSettings.qml
```diff
@@ -58,7 +58,7 @@ Column {
         propertyItem: root.model ? root.model.tempo : null
         enabled: root.model ? !root.model.isEmpty && !followWrittenTempoCheckbox.checked : false
 
-        measureUnitsSymbol: qsTrc("inspector", "BPM")
+        measureUnitsSymbol: qsTrc("inspector", " BPM")
```
cbjeukendrup, line 61: Translatable strings should not contain leading or trailing whitespace, because translators have no reasonable way to notice that. Generally this is solved by putting the extra space outside the translatable string:
```suggestion
        measureUnitsSymbol: " " + qsTrc("inspector", "BPM")
```

Jojo-Schmitz: Other than that we've never separated the numbers from the units by a space, but there had been pretty lengthy discussion about this.

rtbo:
I've rebased the PR on master.
Is there anything left preventing the merge?

src/inspector/models/notation/tempos/temposettingsmodel.cpp
```diff
@@ -27,18 +27,23 @@
...
-    setModelType(InspectorModelType::TYPE_TEMPO);
-    setTitle(qtrc("inspector", "Tempo"));
+    Q_ASSERT(modelType == InspectorModelType::TYPE_TEMPO || modelType == InspectorModelType::TYPE_TEMPO_RESTORE_PREVIOUS);
+    setModelType(modelType);
+    if (modelType == InspectorModelType::TYPE_TEMPO) {
+        setTitle(qtrc("inspector", "Tempo"));
+    } else {
+        setTitle(qtrc("inspector", "Restore previous tempo (a tempo)"));
+    }
     setIcon(ui::IconCode::Code::METRONOME);
```
cbjeukendrup, line 40: For readability, it would be preferable to add a few newlines:
```suggestion
    Q_ASSERT(modelType == InspectorModelType::TYPE_TEMPO || modelType == InspectorModelType::TYPE_TEMPO_RESTORE_PREVIOUS);
    
    setModelType(modelType);
    
    if (modelType == InspectorModelType::TYPE_TEMPO) {
        setTitle(qtrc("inspector", "Tempo"));
    } else {
        setTitle(qtrc("inspector", "Restore previous tempo (a tempo)"));
    }
    
    setIcon(ui::IconCode::Code::METRONOME);
```

cbjeukendrup (commented):
This looks very good! A few minor comments.

cbjeukendrup:
A rebase is needed again...

Jojo-Schmitz:
It is a completely painless and conflict-free rebase, I guess you could rebase and merge in one step.
It does build cleanly on Windows/MSVC too.

I'll submit a new PR (even if only for the tests to run), see #18017

Jojo-Schmitz:
All tests in #18017 passed, so I guess this here is ready for (rebase and) merge.

rtbo:
Great!
Do I have to squash commits as well?

bkunda:
Are we supposed to see a different option in Properties for the new "a tempo" object (per [this comment above](https://github.com/musescore/MuseScore/pull/15563#issuecomment-1497331816))?

At the moment (on macOS) I am seeing the same options for both regular tempo marks and "a tempo"

https://github.com/musescore/MuseScore/assets/86290556/ad1b4719-c9fe-4beb-8d56-e3ac1be0f034

FWIW I did (as I always do) perform a factory reset, and reset the "Tempo" palette before re-testing.
Version number: OS: macOS 13.1, Arch.: x86_64, MuseScore version (64-bit): 4.1.0-231680012, revision: github-musescore-musescore-78be98f

Given that we don't have an "a tempo" object in master though, I assume I haven't made a silly mistake and that I am in fact testing on the correct build!

Aside from this, there is certainly some playback functionality tied to the "a tempo" object I tested. Some issues with it though were that:

- "Follow written tempo" does not recall the first tempo marking assigned to the start of the piece
- Overriding the tempo, then reverting back to "follow written tempo", sees the overridden tempo persist

I made a (apologies) rather lengthy 6min demo video of my test: https://www.dropbox.com/s/16brsjwaky4jphd/Screen%20Recording%202023-06-19%20at%2010.03.27%20pm.mov?dl=0

Unless I've made some mistake in how I've either understood the comments above (especially [this most recent one containing a design mockup](https://github.com/musescore/MuseScore/pull/15563#issuecomment-1497331816)), or in the version I'm testing, I have to admit that I'm quite reticent about merging this now for 4.1 – particularly since we are now entering a "feature freeze" state and are really trying to focus on stabilisation and regression testing. Anything connected to playback is also especially risky, so we want to make sure the feature is as bulletproof as possible before merging it. 

But I am extremely glad this has been worked on – it is a much, much needed feature! I will be happy to ensure it gets prioritised for the next release (4.2) if these issues can be ironed out.

lgvr123:
As mentioned on the [forum](https://musescore.org/en/node/339398#comment-1195939), there are various "a tempo" forms.

The traditional "a tempo" (_back to the previous tempo_), the "Tempo primo" (_resume the original (first) tempo_), ...

What about labelling the tempo marks (such as in the jumps) and referring to those labels in the "a tempo" to indicate which tempo to reuse ?

a tempo2

rtbo:
Hi @bkunda , Thanks for the test and feedback.
What you observe is obviously not what is intended, but I also reproduce it on my side.

Weirdly, the test file I built before I rebased on master 2 weeks ago still works fine (both properties UI and playback).
I have to dig why it is now broken.

Regarding playback, the feature is designed to reset to the last set tempo in the score. So in your test, "a tempo" on bar 16 would reset to 220 BPM, and therefore have no effect. In fact it is effective only after a progressive tempo change, unless "Set specific tempo" option is used.

rtbo:
It should work now.
@bkunda can you repeat your test?

rtbo:
@lgvr123 I added "tempo primo" in the palette. It is similar to "a tempo" but will restore the initial score tempo.

Image collée

rtbo:
Rebased on master and solved conflict (again :smile:).
I can remove "tempo primo" if you want this PR more focused.

AceAttorneyMaster111:
Any idea when this PR will be merged?

Jojo-Schmitz:
First another rebase is needed

src/engraving/dom/masterscore.h
```diff
@@ -141,6 +143,11 @@ class MasterScore : public Score
     TempoMap* tempomap() const override { return _tempomap; }
     async::Channel<ScoreChangesRange> changesChannel() const override { return m_changesRangeChannel; }
 
+    BeatsPerSecond tempoPrimo() const override { return _tempoPrimo; }
+    void resetTempoPrimo() { _tempoPrimo = Constants::DEFAULT_TEMPO; _tempoPrimoSet = false; }
+    void setTempoPrimo(BeatsPerSecond tempoPrimo) { _tempoPrimo = tempoPrimo; _tempoPrimoSet = true; }
+    bool tempoPrimoSet() const { return _tempoPrimoSet; }
```
cbjeukendrup, line 149: It looks like these methods are only used inside `Score::setUpTempoMap()`. Therefore, I think it would be preferable to create a local variable inside `Score::setUpTempoMap()` instead of adding this as a member of the `MasterScore` class.

src/engraving/dom/masterscore.h
```diff
@@ -77,6 +77,8 @@ class MasterScore : public Score
     UndoStack* _undoStack = nullptr;
     TimeSigMap* _sigmap;
     TempoMap* _tempomap;
+    BeatsPerSecond _tempoPrimo;
+    bool _tempoPrimoSet = false;
```
cbjeukendrup, line 81: For this combination of things, it is often nice to use `std::optional`.

src/engraving/dom/masterscore.h
rtbo: It is used in `Score::setUpTempoMap` and `Score::rebuildTempoAndTimeSigMaps`.
As the latter is also private and only called once from `setUpTempoMap`, I localized the variable in `setUpTempoMap` and pass it as mutable reference paremeter to `rebuildTempoAndTimeSigMaps`.

bkunda:
@cbjeukendrup @Tantacrul I've just quickly tested this (macOS only) and it seems to be working well. 
Really liking the "Tempo primo" option too. 
UX seems to align with what was discussed above. 
Happy with this one! Thanks so much @rtbo :-)

bkunda:
Ah sorry! One more thing we might need to reconsider: 

Screenshot 2023-11-08 at 12 21 10 pm

(This always seems to be afterthought!)

"Restore previous tempo" is a very long title for these buttons, and is likely to not localize well. 

We could also think of a slightly modified icon (this can be done separately – no need to block the present PR).

But the label itself reflects the section header, and probably should be decided on here.

Jojo-Schmitz:
Just drop the "Restore"? "Previous tempo" or even "Prev. tempo" might be enough

bkunda:
> Just drop the "Restore" (both actually)? "Previous tempo" or even "Prev. tempo" might be enough, as should be "Initial tempo"

My concern with this is that the use of the verb in the Properties panel is currently clarifying the UX by making it clear what the action does.

Alternatively, how about we simply using the actual object name – the function of which should be clear enough to the user who wishes to use it? 

Screenshot 2023-11-08 at 12 50 23 pm

Jojo-Schmitz:
Ah, so "A tempo" and "Tempo primo". Better indeed.

bkunda:
@rtbo if you're able to make those brief section heading changes I suggested [above](https://github.com/musescore/MuseScore/pull/15563#issuecomment-1801734414), I think we'll be able to get this one past the post for the next release. 

@cbjeukendrup @Tantacrul FYI.

src/engraving/dom/engravingitem.h
```diff
@@ -259,6 +259,10 @@ class EngravingItem : public EngravingObject
 
     virtual int subtype() const { return -1; }                    // for select gui
 
+    // Index to a set of properties if the same element type can represent
+    // different set of properties
+    virtual int propset() const { return subtype(); }
```
RomanPudashkin, line 264: This is not a good solution. Since this is only relevant for Inspector, you need to solve this problem at the AbstractInspectorModel level:

1. declare a static function in AbstractInspectorModel:
`ElementKey AbstractInspectorModel::makeKey(const EngravingItem*)` -  it will create the correct key for TempoText (and other engraving items)

2. call this function from InspectorListModel::setElementList

rtbo:
I should be able to work on the new requests over the weekend.

For my information, where is this "notation" panel in the UI ? Never seen that before.

> Screenshot 2023-11-08 at 12 28 09 pm

cbjeukendrup:
> For my information, where is this "notation" panel in the UI ? Never seen that before.

It's what you see in the Inspector when multiple elements of different types are selected in the score.

bkunda:
I've given this a final test on macOS. All good to go from my end. 
Thank you so much @rtbo! People are going to love this!

Issue #15499, opened 2022-12-22
[MU4 Issue] Remove default Qt button icons 


Ciubix8513:
## Default Qt button icons appear in some dialogues

Found in:
- Page settings
- About Qt 
- Transpose
- Style
- Create tuplet
- Staff text properties
- Staff/Part properties

There are probably more places where this issue can be found 

**Screenshots**
![image](https://user-images.githubusercontent.com/57328501/209102086-e71f8c09-3332-4928-97a1-3a27f688ac75.png)

**Platform information**
 - OS: Linux

Jojo-Schmitz:
Why remove them?
They don't show in Windows BTW, so what OS are you seeing them in?

Ciubix8513:
Consistency, other windows don't have them.

For example here are the buttons of Prefrences window:
![image](https://user-images.githubusercontent.com/57328501/209110684-b4a5d261-ebd7-4df8-a8e2-46e823915b66.png)

They also clash with the clean UI of MU4

cbjeukendrup:
I think the best way to solve this is using QStyle, by overriding the `standardPixmap` and/or `standardIcon` methods. We already have a QStyle subclass, see uitheme.cpp, so adding this method override should be doable enough to call this a good first issue for anyone with a little bit of experience with Qt (or not even that).

PR #15583, opened 2022-12-25
FIX 15499: remove default Qt icons

Da-Viper:
Resolves: #15499 

- [X] I created a unit test or vtest to verify the changes I made (if applicable)


Da-Viper:
I am not sure if this should be the intended behaviour for all icon types so I only disabled them for DialogButtons.

Issue #15500, opened 2022-12-22
Musescore 4 missing font rendering on startup? MacOS Mojave


ianchia:
I just started up Musescore 4 for the first time (MacOS Mojave 10.14.6, MBP 2012 i7) and it is missing almost all, but not quite all of the font rendering onscreen? I see "Musescore 4" at the top, "Getting Started" but all the buttons and dialogs are missing text labels.
![Musescore 4 missing font display](https://user-images.githubusercontent.com/76875/209109409-fcc7af6d-fd6e-4be3-826a-2f35ec4a3b4e.jpg)

cbjeukendrup:
Strange... all text that's missing should use the system font. It's apparently not a general font rendering issue, because icons in the toolbars appear correctly (and those are also characters of a font). The few other texts that do appear are rendered by the system rather than by MuseScore.

Do you have the "SF Pro" font installed on your Mac? Did you ever have it installed but uninstalled it? Is there anything special about fonts on your system that you can think of?

ianchia:
I have SF Pro installed, but to test further, I downloaded fresh copies of all SF Pro and SF variants from apple.com and reinstalled. Launched MuseScore 4 with the same result. Rebooted, and relaunched again - still with same rendering issues on MuseScore 4.

See attachment with Font Book showing SF Pro installed and buggy MuseScore 4 rendering.

There's nothing unusual about fonts on my system. I use a range of apps in Adobe Creative Cloud as well as the Affinity Designer and Photo apps and they work fine with my fonts.

Any other ideas I can try to debug this? Thanks!

![Missing Muse Score font rendering - SF Pro installed](https://user-images.githubusercontent.com/76875/209419120-ca6b6116-1c29-48c0-b614-8de1592025d1.jpg)

cbjeukendrup:
@ianchia I have created a test version, which might resolve the issue. You can download it in ca. 40 minutes from now from https://github.com/cbjeukendrup/MuseScore/actions/runs/3775802804 (scroll down to "Artifacts").

ianchia:
@cbjeukendrup Thank you Casper. This one renders fine! However, under the Learn tab under Get Started, it says "Sorry we are unable to load these videos right now. Please check your internet connection or try again later." Not sure if this is because this is a a development build? Musescore does correctly pull my profile image down after launch so it is connected to the internet.

cbjeukendrup:
[...]

PR #15593, opened 2022-12-26
Remove SF Pro workaround on macOS

cbjeukendrup:
Resolves: #15500

See the comment accompanying the workaround to read what it did. This appeared to cause quite a big problem for someone, while it tried to _solve_ only a small problem. So, getting rid of it is an obvious choice, I think.

Also, I _still_ hope we will switch to Qt 5.15.3 (or even better 5.15.7) soon, which will really fix the problem that the workaround was trying to solve.

### DIFF
--- a/src/framework/ui/internal/uiconfiguration.cpp
+++ b/src/framework/ui/internal/uiconfiguration.cpp
@@ -604,20 +604,6 @@ std::string UiConfiguration::defaultFontFamily() const
 {
     std::string family = QFontDatabase::systemFont(QFontDatabase::GeneralFont).family().toStdString();
 
-#ifdef Q_OS_MACOS
-    // The macOS default font *is* SF Pro, but under the name ".AppleSystemUIFont".
-    // In Qt, the version under this name has an issue with the width of the comma
-    // character (see https://github.com/musescore/MuseScore/issues/10736). This does
-    // not occur in user-installed versions of the font. So if the user has one, we
-    // use it, otherwise, it will be just the system version.
-    // TODO(qt-6): This can be removed when switching to Qt 6.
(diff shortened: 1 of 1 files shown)


igorkorsukov (approved):

DmitryArefiev:
@cbjeukendrup Should I test it on my side? I don't have Mojave installed and doesn't reproduce after installing SF Pro.

cbjeukendrup:
@DmitryArefiev No, this doesn't need to be tested further. The reason I had waited with merging this, is that I had asked on the forums and on Discord if more people were getting this problem, and which OS version they used, hoping to find out if this can happen on any version, or that we only need to disable the workaround for macOS 10.14. I've got not a single reply, so still have no idea. I guess it's fine to disable the workaround on all versions then, since it only solves a pretty small problem (and I still hope we will soon switch to Qt 5.15.8 to fix the problem completely).

Jojo-Schmitz:
Is Qt 5.15.8 publicly available? Or anything newer than Qt 5.15.2? Where?

DmitryArefiev:
@cbjeukendrup Got it. Thanks!

cbjeukendrup:
> Is Qt 5.15.8 publicly available? Or anything newer than Qt 5.15.2? Where?

Yep. It's all here: https://download.qt.io/official_releases/qt/5.15/

I believe the story is that they have an agreement with KDE (or was it someone else?) that obligates them to make every version of Qt open source after at most 1 year after the commercial release. So that's what they do. But they only give you the sources, in this hidden place, and I bet they prefer that as few people as possible know about it.

Jojo-Schmitz:
OK, that's source code only, so not easy to use (in additon to not easy to find)

PR #15596, opened 2022-12-26
Fix #15409: Update score after resetting page styles

### LINKED ISSUE #15409: [MU4 Issue] Reset All Page Settings to Default doesn't update score
**Describe the bug**
See video

https://user-images.githubusercontent.com/102702868/208554342-9891d0d3-c78a-4d5a-9acf-a45caafd32a7.mp4

**Platform information**
 - OS: Windows 11

iwoithe:
Resolves: #15409

https://user-images.githubusercontent.com/73422868/209488626-d811bb8f-80e2-45f9-a018-34b557894003.mp4

### DIFF
--- a/src/notation/internal/notationstyle.cpp
+++ b/src/notation/internal/notationstyle.cpp
@@ -67,6 +67,7 @@ void NotationStyle::setStyleValue(const StyleId& styleId, const PropertyValue& n
 void NotationStyle::resetStyleValue(const StyleId& styleId)
 {
     score()->resetStyleValue(styleId);
+    score()->update();
     m_styleChanged.notify();
 }
 


src/notation/view/widgets/pagesettings.cpp
```diff
@@ -268,6 +268,8 @@ void PageSettings::on_resetPageStyleButton_clicked()
     }
 
     updateValues();
+    globalContext()->currentNotation()->undoStack()->commitChanges();
```
iwoithe, line 271: I must admit that I am not entirely sure why this is required, but it does not fully work without it.

cbjeukendrup: A bit of context about what this line does:

Each thing that the user does needs to be undoable. So before a change, we call `undoStack()->prepareChanges()`, to start recording changes. After the changes, we call either `rollbackChanges()` or `commitChanges()`, to revert the changes, or to "commit" them. A block of committed changes can be undone/redone.

In the case of the PageSettings dialog, we call `prepareChanges` when the user opens the dialog, `rollbackChanges` when the user clicks Cancel and `commitChanges` when the user clicks OK. So it is not desired that we commit the changes earlier than that, so we need another solution.

I looked why the score does update correctly after _changing_ a style value and not when _resetting_. The difference seems to be in the NotationStyle class:
Scherm­afbeelding 2022-12-26 om 15 53 52
The `resetStyleValue` method seems to be missing a call to `score()->update()`.

cbjeukendrup (approved):

cbjeukendrup (approved):
Thanks! But before we merge it, could you please rebase it? so that new test builds are created so that the testers can give it a try.

zacjansheski:
Tested on Mac OS 11.6.5, Ubuntu 20.04.6, Windows 11

https://github.com/musescore/MuseScore/issues/15409 FIXED

PR #15604, opened 2022-12-26
Align the rows in "Create from template" to the rows in "Choose instruments" in the New Score wizard

ericssonpaul:
Previously, when creating a new score the rows in these two tabs weren't aligned. Having them being the same is a bit more aesthetically pleasing.

Before:
Choose_instruments_before
Choose_from_template_before

After:
Choose_instruments_after
Choose_from_template_after


cbjeukendrup:
@bkunda Please check if you agree with these new column widths on the "Create from template" page :)

Issue #15552, opened 2022-12-24
[playback issue] sustained notes stop sounding at end of a bar


Vykoiazon:
**Describe the bug**
The sustained notes stop playing when they should still be sounding at the end of a bar,
maybe caused by the tempo slowing?

**To Reproduce**
1. add rhythms like this where somebody has sustained notes but another part has moving notes before a barline
![image](https://user-images.githubusercontent.com/120861729/209421131-4fabeb0f-f3d3-46e2-8931-ec40904cec51.png)

2. make the second bar a slower tempo and add a rit on the first bar toward the second bar (does not need two different tempo numbers like the example)
4. press play and hear the sustained notes stop around half a beat early
5. if it does not work then maybe test what happens if accents or grace notes are added

**Screenshots**

https://user-images.githubusercontent.com/120861729/209421170-7a670530-fb75-4d4f-ba36-d755fcc20025.mp4

cbjeukendrup:
Yes,  I think it is indeed related to the ritenuto, or more generally, that the tempo at the start of the note differs from the tempo at the end of the note. For example, in this extreme example, the dotted whole note ends _way_ too early:
<img width="344" alt="Scherm­afbeelding 2022-12-24 om 18 26 14" src="https://user-images.githubusercontent.com/48658420/209446002-8f820502-ee19-485a-85d9-3f9c813c633e.png">

zacjansheski:
Combining this with https://github.com/musescore/MuseScore/issues/14369 for reference

Issue #14369, opened 2022-11-06
[MU4 Issue] Tempo changes in a measure cause notes to be cut off or sustain too much (metronome mark or accel./rit.)


LRenn0:
**Describe the bug**
When the tempo changes within a measure, notes such as whole notes will not hold for their complete value.

**To Reproduce**
1. Make a score with multiple instruments
2. Put a whole note on one instrument's part
3. Put something other than a whole note on the other instrument's part in the same measure and change the tempo within that measure (not on beat 1)
4. On playback, the whole note will not be held to the end of the measure

**Expected behavior**
On playback, the note that was played before the tempo change will hold through the tempo change and to the end of the measure.

**Screenshots**
![image](https://user-images.githubusercontent.com/117554969/200189800-7b8c1d4b-8949-432d-a2a1-5f4906a96311.png)
In a setup like this, the whole note from violin 1 will hold for 4 counts at 120 bpm. Violin 2 will play the first half note for 2 counts at 120 bpm and the next half note for 2 counts at 60 bpm. This means that the whole note being played by violin 1 will be cut off before the end of the measure and violin 2 will continue holding the half note until the end of the measure.

--

Added by @DmitryArefiev 

Good example from #17909

https://github.com/musescore/MuseScore/assets/90187801/65e884ca-eecf-4186-b8fa-63d19e156888

**Platform information**
 - OS: Windows

Tantacrul:
Confirmed. Created an example score to demonstrate the issue.
https://www.dropbox.com/s/wmnouy3ha1j6ygu/Tempo%20change%20mid%20bar.mscz?dl=0

Happens with MS Basic and Muse Sounds, so must be an MPE thing.

erinic04:
Also happens with soundfonts, seems to just be a global thing

zacjansheski:
Additional issue reportsd
https://github.com/musescore/MuseScore/issues/17863
https://github.com/musescore/MuseScore/issues/15552

DmitryArefiev:
Another similar issue 
#17909

DmitryArefiev:
Fixed in #15609

PR #15609, opened 2022-12-26
Fix issues with tempo changes during notes

cbjeukendrup:
Resolves: #15552
Resolves: #14369

There are several problems:
1. When converting the duration of an element from ticks to microseconds, we do that based on the tempo at the start of the element, which gives a wrong value if there are tempo changes during the element. Instead, we should look at the end and start timestamps, and use their difference as the duration.
2. We convert to microseconds too early. This means that calculations for grace notes and tremolos for example need to be done in microseconds, without any information about the tempo during the element. For example, if there is a ritenuto during a tremolo, the tremolo would need to slow down, but we can't know if we use microseconds. Instead, we should calculate the position of every tremolo note in ticks, and then convert those ticks to microseconds, so that the tremolo will slow down correctly.
3. (Not yet solved) To MuseSampler, we can only communicate the tempo at the beginning of the note; we have no way to tell MuseSampler about tempo changes during the note. This could potentially cause problems with MuseSampler (and maybe even confuse it).


BoBogus:
In my own submission of this issue (#17863), I realized a potential (if perhaps a bit inelegant) solution that can at least improve the issue with long notes.  If you watch the tempo indicator on the upper left corner, you'll notice it doesn't gradually change, but rather changes in leaps every quarter note. I just realized that I could use that to my advantage by breaking a problematic half note into two tied quarter notes. The interpretation is musically the same, but the software ends up calculating it properly so the note stops in time.

So a potential solution (at least, until something more elegant can be discovered) could be for the software to automatically treat long notes as a chain of tied notes no longer than a quarter, either everywhere or just whenever they occur within a tempo change (unless the necessary calculations are memory inefficient).

cbjeukendrup:
@BoBogus The solution used by this PR is more general and more straight-forward: instead of calculating the length based on the tempo at the start timestamp, we should calculate the end timestamp and use the difference between the start and end timestamp as the length. So, the idea is very simple, but the difficulty is to properly apply it in _all_ places.

cbjeukendrup:
I've now fully implemented this solution in all places. It seems to work well, but I'm afraid that it is prone to rounding errors, since the resolution of "ticks" is only 480 ticks per quarter note. Maybe we should simply change that fact, because it is problematic in more situations: #16253. Ideas are welcome.

cbjeukendrup:
Actually the PR might indeed improve the situation regarding fermatas somewhat, but that's not its main goal; fully solving the "fermata consolidation" problem would be a separate task.

DmitryArefiev:
Tested #14369 on Win10 - **FIXED**

PR #15613, opened 2022-12-27
Antialiase plugin thumbnails

iwoithe:
Resolves: N/A

This PR fixes a subtle, but visible, sharpness to the plugin thumbnails.

**Before**

Screenshot of before

**After**

Screenshot of after

### DIFF
--- a/src/plugins/qml/MuseScore/Plugins/internal/PluginItem.qml
+++ b/src/plugins/qml/MuseScore/Plugins/internal/PluginItem.qml
@@ -20,6 +20,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 import QtQuick 2.15
+import QtQuick.Window 2.15
 import QtGraphicalEffects 1.0
 
 import MuseScore.Ui 1.0
@@ -63,6 +64,8 @@ Item {
             anchors.fill: parent
 
             fillMode: Image.PreserveAspectCrop
+
+            sourceSize: Qt.size(width * Screen.devicePixelRatio, height * Screen.devicePixelRatio)
 
             layer.enabled: true
             layer.effect: OpacityMask {


DmitryArefiev:
Tested on Win10 (QHD and FullHD), Mac13 (Retina), LinuxUbuntu (FullHD) 

On Windows it's fine on FullHD display but I see a little blurriness on QHD display. 

@iwoithe As I understand, we don't have another icons with better resolution for high DPI displays, right?

DmitryArefiev:
@jessjwilliamson Can you check this briefly on Windows please? (but if you have QHD or 4K display)

iwoithe:
> @iwoithe As I understand, we don't have another icons with better resolution for high DPI displays, right?

I don't have a high DPI display.

The exact problem (should have mentioned this earlier, sorry) is this:
When scaled up, images are smoothed by default in QML (see [here](https://doc.qt.io/qt-6/qml-qtquick-image.html#smooth-prop)). However, when an image is scaled down, nothing is done, hence the subtle sharpness. This PR adds the [mipmap](https://doc.qt.io/qt-6/qml-qtquick-image.html#mipmap-prop) property which smooths out images when they are downsized.

So I don't believe better resolutions will fix this issue. In fact, higher resolutions *may* make it worse.

jeetee:
#random thought: Is it an option to switch to SVGs for these images instead?

Eism:
@iwoithe 
Actually, I have never used mipmap in my work, and as we can see it does not work in all cases.

This problem can be solved with `sourceSize: Qt.size(width, height)` instead of `mipmap`. 
Please check on your machine.

cbjeukendrup:
`sourceSize: Qt.size(width, height)` does not work for high resolution displays (the resolution of the image becomes too low). It seems like `Qt.size(width * Screen.devicePixelRatio, height * Screen.devicePixelRatio)` does work correctly.

Eism:
please rebase it

DmitryArefiev:
Tested on Win10, Mac12, LinuxUbuntu 22.04 - **Seems to be better now on Windows high-res displays**    

(FYI @bkunda or @jessjwilliamson You check it briefly on Windows 2K or 4K display. Just to make sure it did not become too blurry after antialiasing)

Eism (approved):

PR #15633, opened 2022-12-28
WASAPI Driver: don't crash (or rather hang) when starting client fails

cbjeukendrup:
We were waiting infinitely for the WASAPI_Client_Started event, but when there was an error while starting the client, that event would never be signaled. So MuseScore would be stuck infinitely on launch.

So, in the cases where MS would hang before this PR, playback won't work, but at least MuseScore will launch. 

This is maybe helpful in some of the cases of "crash on launch".

### DIFF
--- a/src/framework/audio/internal/platform/win/wasapiaudioclient.cpp
+++ b/src/framework/audio/internal/platform/win/wasapiaudioclient.cpp
@@ -29,8 +29,9 @@ using namespace winrt::Windows::Foundation;
 using namespace winrt::Windows::Media::Devices;
 using namespace winrt::Windows::Devices::Enumeration;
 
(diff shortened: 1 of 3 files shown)


cbjeukendrup:
I asked quite a lot of people on the MuseScore.org forums to test this, and so far the results seem exactly what I expected: MuseScore does at least launch now, only playback doesn't work. 

In fact, I sent them a test build with some extra logging and it turns out that (not very surprisingly) the problem comes from `m_audioClient->InitializeSharedAudioStream`, which throws the `AUDCLNT_E_UNSUPPORTED_FORMAT` error.

vpereverzev (approved):

vpereverzev:
looks fine to me

vpereverzev:
@DmitryArefiev test that one pls

Issue #15638, opened 2022-12-29
[MU4 Issue] The instrument names of hidden staffs affect indentation on the first system


bakajikara:
**Describe the bug**
On the first system, the name of the hidden staff affects the indentation.

**To Reproduce**
See video

https://user-images.githubusercontent.com/102702868/209924040-310f43bb-b9ee-4300-8d0a-31d3270299c9.mp4

**Expected behavior**
On the first system, the instrument names of the hidden staffs should not affect the indentation.

**Platform information**
 - OS: Windows 11

L0uisc:
@oktophonie Have you started working on this yet? I'd like to take it on to help me learn the code base.

oktophonie:
I'm not working on it, I'm merely the Engraving Overseer to have assigned myself so that I can keep track of the issue. If you'd like to take a look at it, please feel free!

Jojo-Schmitz:
It does also happen in Mu3, just slighly different: the space is not added at the beginning, but the measures of the 1st system seem to get increased and may cause an automatic system break at an earlier measure as it'd otherwise happen

Issue #8968, opened 2021-08-30
[MU4 Issue] Left origin of staves should be consistent throughout score


oktophonie:
The left edge of the staves should be a consistent position through the entire score, excepting: a) the first system (which may have full-length, as opposed to abbreviated, instrument names, or which may have an indent specified in Style -> Score), and b) any system where this is specifically overridden, e.g. by adding a frame to increase the indent.

Currently, MuseScore calculates the beginning of the staves _for each system independently_, according to the longest instrument label in use, or the bracket that extends furthest to the left, resulting in a ragged left:

![margins-before](https://user-images.githubusercontent.com/20604032/131327873-a7385d55-aed4-4c78-b71d-948f23e84899.png)

The beginnings should all be aligned, according to whichever needs the largest space:

![margins-after](https://user-images.githubusercontent.com/20604032/131327953-3d20449a-510a-4a6f-aaf5-87e9fbfc1bc2.png)

Further, it is conventional for brackets and braces, where there are no instrument labels, to extend into the left margin. So rather than:

![margins2-before](https://user-images.githubusercontent.com/20604032/131328450-3f484d95-1c7d-4152-a4c1-2dfcb8041c31.png)

we should see:

![margins2-after](https://user-images.githubusercontent.com/20604032/131328461-1d42c45a-52ec-4501-aded-b601f49bd4b2.png)

Though very common, this is not universal. The important part is the left origin of the staves remaining consistent. Keeping brackets/braces out of the left margin could be achieved in Style -> Score with an option, either:

- to toggle this behaviour explicitly, or
- to apply an indent to all staves _after_ the first, of a modifiable value in sp. Like the first-system indent, this would effectively be a 'minimum' and would almost certainly be trumped by the length of any instrument labels when present.

**Should fix:**
https://musescore.org/en/node/99136
https://musescore.org/en/node/317246
(somewhat indirectly)

PR #15645, opened 2022-12-29
15638 hidden instr names takes space

L0uisc:
Resolves: #15638 

Fixes instruments names from hidden staves influences measure start position.

### DIFF
--- a/src/engraving/libmscore/system.cpp
+++ b/src/engraving/libmscore/system.cpp
@@ -274,7 +274,7 @@ double System::instrumentNamesWidth()
 
     for (staff_idx_t staffIdx = 0; staffIdx < score()->nstaves(); ++staffIdx) {
         const SysStaff* staff = this->staff(staffIdx);
-        if (!staff) {
+        if (!staff || !staff->show()) { // Fix for #15638: ignore hidden staves when calculating instrument names width.
             continue;
         }
 


L0uisc:
I'm not familiar enough with the testing framework yet to write a test for this change. Do you think it's necessary for this change?

L0uisc:
vtests are failing, but I think they should be updated. The failures are where a staff with a longer instrument name is hidden. Issue is, I have no idea how these tests work. I'll check the wiki and then come back here if I need help.

cbjeukendrup:
Note for manual testing: we need to make sure this doesn't break https://github.com/musescore/MuseScore/issues/8968.

Regarding the automatic tests: the existing VTests are probably enough in this case, so no need to add another test. You don't need to update the VTests manually; they are automatically compared to the latest master branch. So if we merge your PR, then future PRs will compare with the master branch including your PR.

L0uisc:
Hmm... I'll maybe need to figure out a way to take all systems on the page into account here. Well, that complicates things a bit. I should've known it's too good to be true ;-)

L0uisc:
image

It does regress that. I guess I'd need a two pass approach to page layout then. That complicates matters considerably.

L0uisc:
So thinking further: do we even want to fix this? Are we going to assume that instrument names won't be so dissimilar in length to cause wasted space? Because implementing a system that can remove unnecessary space from hidden instrument names *and* keep the systems aligned on a page will be a non-trivial monstrosity of code.

I think I should close the PR, since it's really trivial. I'll rather start fresh from a current master branch if I decide to try implementing an algorithm that can handle this issue and keep the systems aligned. For now I think the fix I provided does more harm than the issue it's trying to fix.

This exercise did serve the purpose to familiarize me with the layout system, though, so not all is lost.

cbjeukendrup:
One more thought: one reason _not_ to fix this issue, is that it causes different system left margins on each page, and I'm not sure if _that_ is desirable. Maybe that is also one of the reasons that @oktophonie preferred that I implemented the left-alignment in a "score-wide" way rather than per page.

L0uisc:
Yes, I think the reported issue is enough of an edge case that the fix is "don't do it". Best case I can see some heuristic type algorithm that can try to improve things for a finite number of tries and pick the best option, but first MU4 needs to become stable and optimized as is. We already have users struggling with performance on less powerful machines.

Issue #15546, opened 2022-12-23
[MU4 Issue] Cannot delete or reset "auto-generated" parts


cbjeukendrup:
**Describe the bug**
In the Parts dialog, it is not possible to delete "auto-generated" parts, even after they have really been created by opening them. This is currently disabled on purpose, because it is not implemented; see https://github.com/musescore/MuseScore/issues/11349#issuecomment-1172462901. 

The possibility to delete an auto-generated part is necessary in the following cases:
- You messed up the part badly and want to reset it to start over again
- MuseScore messed up the part by corrupting it; that happens sometimes currently. For scores where this has happened, the best way to recover them is deleting the part(s) containing the corruption; currently that is not possible.

sammik:
I think this needs design first.

I see several problems with #15665 behaviour.

1) In User POV, it is strange, if I delete part, and next time opening Parts dialog, "deleted" part is there again.
So I think, if deleting parts is allowed, than parts should remain deleted. May be some new button to "Recreate generic parts".
Or dont allow "deleting" generic parts and allow "recreate" or "restore default" or something like that instead.

2) If parts were not already open, "delete" is missing in generic parts. But is present in custom parts. (To reproduce, create new String quartet score, open parts dialog and try to delete part.)

cbjeukendrup:
Yes, I agree this will need some thoughts again. @Tantacrul's idea was to hide the concept of "creating parts" from the user, because some users might be unable to understand it. That seems a good to me, as long as it doesn't have any negative consequences for users who do understand it, which it hasn't indeed. 

But when it comes to deleting parts, this design, or rather the current implementation of it, is actually only _more_ confusing: deleted parts (that were auto-created initially) suddenly reappear when you reopen the dialog, and then you can't re-delete them anymore until you open them. I'm afraid that users who don't know the underlying thoughts may experience this as janky.

So I propose that we slightly tweak the design and implementation:

1. When the user creates a new score, or when the user adds new instruments, we generate parts for the missing instruments. (Additionally, we might also do that for MS 3.x/2.x/1.x scores, which might not have all parts yet.)
2. But after that, we forget which parts were generated by MuseScore and which were user-created; all of them get equal treatment. In particular, all of them can be deleted.
3. We provide a way to easily re-create deleted parts, namely a popup inside the parts dialog with a "Create empty part" button, and a list of instruments that don't have a part so that the user can click those instruments to create a new part for them. 

[...]

PR #15665, opened 2022-12-31
Allow deleting auto-created parts

cbjeukendrup:
Resolves: #15546

When you delete an auto-created part, it will re-appear when you re-open the parts dialog, so that you can open it again to re-create it.

### DIFF
--- a/src/notation/imasternotation.h
+++ b/src/notation/imasternotation.h
@@ -52,9 +52,8 @@ public:
     virtual void initExcerpts(const ExcerptNotationList& excerpts) = 0;
     virtual void addExcerpts(const ExcerptNotationList& excerpts) = 0;
     virtual void removeExcerpts(const ExcerptNotationList& excerpts) = 0;
+    virtual void resetExcerpt(IExcerptNotationPtr excerpt) = 0;
     virtual void sortExcerpts(ExcerptNotationList& excerpts) = 0;
-
-    virtual IExcerptNotationPtr resetReplaceExcerpt(IExcerptNotationPtr excerpt) = 0;
 
     virtual void setExcerptIsOpen(const INotationPtr excerptNotation, bool opened) = 0;
 
--- a/src/notation/imasternotation.h
+++ b/src/notation/imasternotation.h
@@ -54,6 +54,8 @@ public:
     virtual void removeExcerpts(const ExcerptNotationList& excerpts) = 0;
     virtual void sortExcerpts(ExcerptNotationList& excerpts) = 0;
 
+    virtual IExcerptNotationPtr resetReplaceExcerpt(IExcerptNotationPtr excerpt) = 0;
+
     virtual void setExcerptIsOpen(const INotationPtr excerptNotation, bool opened) = 0;
 
     virtual INotationPartsPtr parts() const = 0;
--- a/src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
+++ b/src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
@@ -34,7 +34,8 @@ ListItemBlank {
 
     property int currentPartIndex: -1
 
-    property bool isInited: false
+    property bool canReset: false
+    property bool canDelete: false
 
     property int sideMargin: 0
 
@@ -42,6 +43,7 @@ ListItemBlank {
     navigation.accessible.name: title
 
     signal copyPartRequested()
+    signal resetPartRequested()
     signal removePartRequested()
     signal partClicked()
 
@@ -160,10 +162,13 @@ ListItemBlank {
 
         MenuButton {
             Component.onCompleted: {
-                var operations = [ { "id": "duplicate", "title": qsTrc("notation", "Duplicate") },
-                                   { "id": "rename", "title": qsTrc("notation", "Rename") }]
+                var operations = [
+                            { "id": "duplicate", "title": qsTrc("notation", "Duplicate") },
+                            { "id": "rename", "title": qsTrc("notation", "Rename") },
+                            { "id": "reset", "title": qsTrc("notation", "Reset"), "enabled": root.canReset }
+                        ]
 
-                if (root.isInited) {
+                if (root.canDelete) {
                     operations.push({ "id": "delete", "title": qsTrc("notation", "Delete") })
                 }
 
@@ -182,6 +187,9 @@ ListItemBlank {
                     break
                 case "rename":
                     root.startEditTitle()
+                    break
+                case "reset":
+                    root.resetPartRequested()
                     break
                 case "delete":
                     root.removePartRequested()
--- a/src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
+++ b/src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
@@ -34,7 +34,7 @@ ListItemBlank {
 
     property int currentPartIndex: -1
 
-    property bool isCustom: false
+    property bool isInited: false
 
     property int sideMargin: 0
 
@@ -163,7 +163,7 @@ ListItemBlank {
                 var operations = [ { "id": "duplicate", "title": qsTrc("notation", "Duplicate") },
                                    { "id": "rename", "title": qsTrc("notation", "Rename") }]
 
-                if (root.isCustom) {
+                if (root.isInited) {
                     operations.push({ "id": "delete", "title": qsTrc("notation", "Delete") })
                 }
 
--- a/src/notation/view/partlistmodel.cpp
+++ b/src/notation/view/partlistmodel.cpp
@@ -179,15 +179,9 @@ void PartListModel::doResetPart(int partIndex)
         return;
     }
 
-    bool isCurrentNotation = context()->currentNotation() == m_excerpts[partIndex]->notation();
-
-    m_excerpts[partIndex] = masterNotation()->resetReplaceExcerpt(m_excerpts[partIndex]);
+    masterNotation()->resetExcerpt(m_excerpts[partIndex]);
 
     emit dataChanged(index(partIndex), index(partIndex));
-
-    if (isCurrentNotation) {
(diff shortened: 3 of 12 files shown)


Tantacrul:
So, I just tried this - and it's still quite weird. I had to re-read the issue thread to understand what was going on. Basically, to me (and to just about all users), this has two points of confusion:

1. Some parts can be deleted but not others. It's unclear why. (although I do _**know**_ why now!)
2. When you delete a part, it's suddenly back there when you reopen the parts dialog, which is even weirder

As a simple patch: I'd suggest the following to help make things clearer to the user:

1. Rename the delete function to 'Reset'. In that way, the user might be able to grasp that it only works on scores that have been edited and will not be confounded when the option appears back in the dialog
2. Add a 'reset' option to all parts but disable them when they can't be triggered
3. When the user does reset a part, make sure it reappears in the Parts dialog. The user will not figure out that they need to close it and reopen it.
4. Add an '**are you sure you want to reset this part. This action cannot be undone**' dialog when the user chooses to reset a part.

Tantacrul:
Seems to be an issue with the MacOS build, unfortunately...

cbjeukendrup:
Should be better this time!

Tantacrul:
This certainly feels a lot easier to understand for sure.

Only one thing I noticed: when I reset a part, and reselect the part (without opening it) the reset option doesn't become disabled:

image


It's not a big deal but if possible to fix, it'd be nice.

cbjeukendrup:
@Tantacrul Unfortunately that seems difficult to solve, at least for custom parts. For auto-created parts, it would be possible, but for custom parts, we have no way of storing a non-initialised custom part, so we must initialise them immediately. That means that their "reset" button will be enabled, because we have no way to know whether the user has already ever touched that part. We can add all those things, but it would be a bit of work and for compatibility reasons I'm not convinced that we should do this in a patch release. 

Do you think it would be good to fix it for auto-created parts only, or shall we keep it consistent for auto-created and custom parts?

Tantacrul:
I think we should just leave it for now. It's not confusing. It's only a tiny bit eye-raising and won't cause anyone any trouble.

Thanks!

Eism (approved):

src/notation/view/partlistmodel.cpp
```diff
@@ -149,6 +152,44 @@ void PartListModel::selectPart(int partIndex)
...
+    }
+
+    bool isCurrentNotation = context()->currentNotation() == m_excerpts[partIndex]->notation();
+
+    m_excerpts[partIndex] = masterNotation()->resetReplaceExcerpt(m_excerpts[partIndex]);
+
+    emit dataChanged(index(partIndex), index(partIndex));
+
+    if (isCurrentNotation) {
+        context()->setCurrentNotation(context()->currentMasterNotation()->notation());
```
Eism, line 189: If the part is opened, then you reset to the main score
Is this the expected behavior?

cbjeukendrup: I changed it; now, the resetted part will just stay open.

src/notation/view/partlistmodel.cpp
```diff
@@ -149,6 +152,44 @@ void PartListModel::selectPart(int partIndex)
...
+    }
+
+    bool isCurrentNotation = context()->currentNotation() == m_excerpts[partIndex]->notation();
+
+    m_excerpts[partIndex] = masterNotation()->resetReplaceExcerpt(m_excerpts[partIndex]);
+
+    emit dataChanged(index(partIndex), index(partIndex));
+
+    if (isCurrentNotation) {
+        context()->setCurrentNotation(m_excerpts[partIndex]->notation());
```
RomanPudashkin, line 189: This code looks quite strange. Why do we need to change the current notation?

src/notation/imasternotation.h
```diff
@@ -54,6 +54,8 @@ class IMasterNotation
     virtual void removeExcerpts(const ExcerptNotationList& excerpts) = 0;
     virtual void sortExcerpts(ExcerptNotationList& excerpts) = 0;
 
+    virtual IExcerptNotationPtr resetReplaceExcerpt(IExcerptNotationPtr excerpt) = 0;
```
RomanPudashkin, line 57: A clearer interface would look like this:

`virtual void resetExcerpt(IExcerptNotationPtr excerpt) = 0:`

Or

`virtual void resetExcerptToDefault(IExcerptNotationPtr excerpt) = 0:`

src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
```diff
@@ -160,8 +161,11 @@ ListItemBlank {
...
             Component.onCompleted: {
-                var operations = [ { "id": "duplicate", "title": qsTrc("notation", "Duplicate") },
-                                   { "id": "rename", "title": qsTrc("notation", "Rename") }]
+                var operations = [
+                            { "id": "duplicate", "title": qsTrc("notation", "Duplicate") },
+                            { "id": "rename", "title": qsTrc("notation", "Rename") },
+                            { "id": "reset", "title": qsTrc("notation", "Reset"), "enabled": root.isInited }
+                        ]
 
                 if (root.isCustom) {
```
RomanPudashkin, line 170: As far as I can tell, this code will no longer work because you renamed the property

src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
```diff
@@ -34,14 +34,15 @@ ListItemBlank {
 
     property int currentPartIndex: -1
 
-    property bool isCustom: false
+    property bool isInited: false
```
RomanPudashkin, line 37: I would suggest something like "resetEnabled"

src/notation/view/partlistmodel.cpp
cbjeukendrup: Because the current excerpt has been deleted and replaced with a new one by `masterNotation()->resetReplaceExcerpt(m_excerpts[partIndex]);`

src/notation/qml/MuseScore/NotationScene/internal/PartDelegate.qml
cbjeukendrup: Thanks, I forgot to add a new `isCustom` property. Now I named it `canDelete`.

src/notation/imasternotation.h
cbjeukendrup: But that would mean we have to reset the `IExcerptNotationPtr` while leaving it in place. That is potentially a lot more work and prone to make things go out of sync. Just replacing it with a fresh one was a lot more straight forward. But if you think it's better I'll go for it.

RomanPudashkin: > That is potentially a lot more work

This is not a very good argument. Looking at this interface, I can't tell what it does. I have to open the .cpp file to understand that.

> Just replacing it with a fresh one was a lot more straight forward.

You can still destroy the old engraving::Excerpt*  and replace it with a fresh one inside the implementation of the function. What kind of difficulties does this cause?

cbjeukendrup: The big problem of that approach is that it requires us to change the `Score*` of an existing `Notation`. At least I have never actively kept in mind that we would ever have to do that, so who knows where we accidentally store references to the old `Score*`.

At a first glance, it seems to work fine though, so we can try. But to me it still feels like a risky thing to do.

By the way, my implementation was somewhat inspired by `MasterNotation::updateExcerpts()`.

src/notation/view/partlistmodel.cpp
```diff
@@ -149,6 +152,40 @@ void PartListModel::selectPart(int partIndex)
...
+    doResetPart(partIndex);
+}
+
+void PartListModel::doResetPart(int partIndex)
+{
+    if (!isExcerptIndexValid(partIndex)) {
+        return;
+    }
+
+    bool isCurrentNotation = context()->currentNotation() == m_excerpts[partIndex]->notation();
```
RomanPudashkin, line 182: This variable is no longer used

RomanPudashkin (approved):

DmitryArefiev:
@cbjeukendrup I found two minor issues:

1. When Reset part and Undo/Redo, it switches focus to the score

 

https://user-images.githubusercontent.com/90187801/220932419-915dadcf-b0ba-4ac0-b512-fbd20495671b.mp4

2.  After resetting parts, they missing as tabs after reopening the score


https://user-images.githubusercontent.com/90187801/220932819-4dc15f9d-6caa-48ec-b99b-34a39670fef1.mp4

DmitryArefiev:
@Tantacrul @bkunda Quick design question: In auto-created parts, should it delete other instruments added to that part (by eye option)?


https://user-images.githubusercontent.com/90187801/220933793-57064a74-901c-43c0-8ffd-622b7440a426.mp4

Tantacrul:
> @Tantacrul @bkunda Quick design question: In auto-created parts, should it delete other instruments added to that part (by eye option)?

I'm not _too_ bothered either way. However, if it doesn't cost lots of time, I would argue that the score should return to its original state (in this example, only Violin 1 should be visible)

cbjeukendrup:
> @cbjeukendrup I found two minor issues:
> 
> 1. When Reset part and Undo/Redo, it switches focus to the score
> 
> 
> 2. After resetting parts, they missing as tabs after reopening the score

@DmitryArefiev Issue 2 turned out to be easily solvable, so I created https://github.com/musescore/MuseScore/pull/16511. (@RomanPudashkin you might want to include that one in 4.0.2 too).

Issue 1 seems more difficult, since it is actually related to how undo works with regard to the wrappers around libmscore. Those wrapper classes are namely re-created when "undo" is called, in a way that we can't detect that a part has been replaced with another one, rather than deleted. So, when undoing, the part that had been reset disappears (because it's being replaced with the old version of that part) but MuseScore thinks it is being closed, so therefore it switches to the master score, and it does not notice that a few microseconds later a new part is created to replace the disappeared part.

Issue #14209, opened 2022-10-31
[MU4 Task] Implement fix for ability to set custom hooks on line (resolves #13653)


bkunda:
**Task description**
This design attempts to resolve the two problems identified in #13653, namely:
- Expose options to set a start hook on a line
- Allow negative values for hook height

1. The **Properties** panel will show the following settings when a **Line** is selected (Note: This applies specifically and _only_ to the palette element "Line", not to other kinds of lines such as "System text line", "Pedal (straight hooks)", etc.). 
    
<img width="752" alt="Frame 327" src="https://user-images.githubusercontent.com/86290556/199017625-9cf5b1e2-e2ee-4d8e-9e0b-76c32fc83d5e.png">

Note that, by default, controls for "Start hook height" and "End hook height" will be disabled. They become enabled when the user selects a start/end hook: 

<img width="665" alt="Frame 326" src="https://user-images.githubusercontent.com/86290556/199017126-da6d2b51-ca0f-4ae4-8c41-a62bdd01bf7c.png">

2. Please remove the minimum value on the spin boxes so as to allow negative values.

Let me know if there are any questions 🙂
Thanks!

MarcSabatella:
Also missing is the option to allow diagonal.  Although FWIW I'd be just as happy to not require a checkbox for that, but most lines really should be constrained horizontal, so there would still need to be some way to specify this.

I came up with a nice little workaround for my own usage today.  I set up a line with start & end hooks that allows diagonals in MU3, saved the score, loaded it into MU4, then added that line to my palette.

bkunda:
Do you mean the ability to make a diagonal line? (because we already have two ways of creating this in MS4)

MarcSabatella:
Now that is weird, I would swear it wasn't working a couple of hours ago.  I guess it works for ordinary lines, but not pedal?

bkunda:
> Now that is weird, I would swear it wasn't working a couple of hours ago. I guess it works for ordinary lines, but not pedal?

It appears you can't make a pedal line diagonal. Is there a case where this would be desired/useful?

oktophonie:
Very rarely seen in real life but possible. Excuse my gorgeous photography:
![image](https://user-images.githubusercontent.com/20604032/212716003-da39b186-b7f9-4456-845b-b02bb542be37.png)
Not something I'd be in a rush to enable personally.

But diagonal lines are useful for indicating a gradual release though:
[...]

PR #15668, opened 2022-12-31
Implement #14209-2: Allow setting line hook heights to negative values

Gravifer:
Resolves: #14209-2 #15871 

Text lines under staves may be expected to have hooks pointing up, 
as the behavior of built-in lines, e.g., the pedal line. 
User should be allowed to set hook height to negative values in the Properties tab.


Gravifer:
An itsy bitsy change, I don't think this would break anything. I hope this PR of minimal effort does not seem distasteful.

cbjeukendrup (approved):

oktophonie (approved):

Issue #15685, opened 2023-01-02
[MU4 Issue] Musescore will crash when the audio output device is not set to 44100 Hz 


99hcl:
**Describe the bug**
Musescore will not open when an audio device is set to anything above 44100 Hz, simply showing a splash screen and closing without warning. It will crash mid-session if the audio device is set to a sampling frequency above 44100 Hz mid-session.

**To Reproduce**
1. Open Windows "Sound" setting, set primary audio device to any combination of channels and sampling frequency above 44100 Hz.
2. Attempt to open Musescore. Splash screen will be shown, but program will not launch.

Alternatively:
1. Set audio device to 44100 Hz and open Musescore normally.
2. Change audio device to 48000 Hz or above via the Windows Sound setting. A warning that programs using the device may stop working will appear. Click OK.
3. Musescore crashes.

Note that in both scenarios the application is prohibited from taking exclusive control of the audio device via Windows Sound setting.

**Expected behavior**
Needless to say, the expected behaviour would be for MU to work out of the box with whatever combination of channels and sampling frequencies I set my audio device to (or at the very least just 48k), or, **even better**: allow me to request a sampling frequency and preferred API along with buffer size from inside the I/O settings panel itself (just like REAPER, for example). Of course, for this to happen the program needs to open first. 

**Platform information**
 - OS: Win10, 64-bit

**Additional context**
Musescore has had problems with higher sampling frequencies since forever, and it seems this problem carried over to MU4. In MU3 it would stop all other applications from playing sounds, and hijack the device to 44100 Hz, even when it was blocked from taking exclusive control of the audio device via Windows settings. It would also not playback any sounds when started in a higher frequency. Multiple "Restart Audio and MIDI Devices" later, and maybe changing the API to something other than ASIO could make it work on a good day. In MU4 there is seemingly no way to set the API manually or reset the devices, so no workaround is possible. It's worth noting that after attempting to make MU4 work with my Behringer UMC404HD by setting it to a lower frequency, the option to make it 4 channel, 24 bit, 48000 Hz has disappeared entirely out of my Windows settings. I'll probably have to reinstall the driver to get it back.

I've also seen other issues being reported where users had to unplug their headphones in order to launch MU4 without crashing and I suspect this is what's at play.

cbjeukendrup:
Hi! Could you please try if the build from https://github.com/musescore/MuseScore/suites/11044663711/artifacts/561149134 works correctly?

99hcl:
[...]

Issue #15637, opened 2022-12-29
[MU4 Issue] It doesn't work with MuseScore4 installer from MuseHub in Windows 10.


anjwl:
**Describe the bug**
MuseScore4 closes or crashes at start.

**To Reproduce**
1. Download MuseScore4 from MuseHub
2. Open MuseScore4
3. Wait until Loading window appears

**Expected behavior**
It should work with MuseScore4 installer from MuseHub.

**Screenshots**

**Platform information**
 - OS: 64-bit Window 10

**Additional context**
It's same as https://github.com/musescore/MuseScore/issues/13899.
But it doesn't work with MuseScore4 installer from MuseHub in Windows 10.

L0uisc:
Can you try to attach the logs as you did with #13899? I'm also on Windows 10 64 bit and I have no issues with MuseHub, so there must be something unique to your system that triggers the crash.

cbjeukendrup:
Could you do the following: 

1. Go to https://www.dropbox.com/s/cv4vvn3lutdezna/MuseScore-4.0.0.223622217-x86_64.zip?dl=0 and click "Download" in the top left corner (you don't need to log in).
3. The downloaded file is a zip file; unpack it.
4. Inside the resulting unpacked folder, you should find a subfolder called `bin`, and inside that folder you should find `MuseScore4.exe`. This is the file you need; double-click it to launch MuseScore.
5. It should crash again, just like the official version 4.0.0; or maybe it does load but playback doesn't work.
6. Now, in Windows Explorer, please go to `%localappdata%\MuseScore\MuseScore4Development\logs` to find the logs. (Note that this is a different location than for the logs of the official version: this test version puts the logs in `MuseScore4Development`, while the official version puts them in `MuseScore4`.)
7. Here you should find the logs from the last attempt to run the test version. Please upload them here; you may need to rename them to have .txt at the end, so that the forum lets you upload them. 
8. When you are ready, you can simply remove the downloaded folder; with this procedure, nothing is installed on your computer, so you don't need to uninstall anything either.

For many people, this crash appeared to be related to audio devices. As a workaround, you could try changing some settings or switching to a different device for the time being. It would be great if you could give some details about the audio device that causes the crash: which sample rate and bit depth does it use? And has it more than two audio channels or some kind of "spatial audio" mode enabled?

anjwl:
@cbjeukendrup It does load but playback doesn't work. This is my logs folder. (I can't find the logs.)
[...]

Issue #15235, opened 2022-12-16
[MU4 Issue] MU4 won't open, crash on splashscreen - Windows 11 (WASAPI related)


jaymzgordon:
**Describe the bug**
I've installed (and uninstalled/re-installed, both from the MuseHub and standalone) the latest version and it crashes on the splash screen every time. I've tried admin mode and various compatibility modes but no luck.

I'm on Windows 11 Pro, latest update. My audio device is a PreSonus Studio 1810c.
CPU: AMD Ryzen 7 3700X
RAM: 64GB
GPU: NVIDIA GeForce GTX 1050 Ti

I'm also running Sonarworks/Sound ID Reference but I've tried booting MuseScore without that running and still hits the same issue.

**To Reproduce**
1. Click on app icon
2. Splashscreen shows 'loading...'
3. App doesn't open

**Expected behavior**
App opens

**Screenshots**
Most recent log file
<img width="848" alt="Screenshot 2022-12-16 185715" src="https://user-images.githubusercontent.com/120748200/208169391-a243e9cb-c73e-43a7-b1ee-ed5674de4724.png">

**Platform information**
 - OS: Windows 11 Pro

**Additional context**

webfischi:
For me it started after I updated Windows, GPU and the driver of my Sound BlasterX AE-5 Plus today, all settings stayed the same as far as I could see, it was always set to 5.1 Surround 24 Bit 48kHz, it starts with every other output except the output I actually use, which is Speaker (Sound BlasterX AE-5 Plus). SPIDIF from the sound card: no crash, headphone output from my AT2020USB+: no crash, Nvidia HDMI Audio: no crash, Voice Meter: no crash. Hopefully the next patch is around the corner.

cbjeukendrup:
Hello! Could you please try if the build from https://github.com/musescore/MuseScore/suites/11044663711/artifacts/561149134 works correctly?

jaymzgordon:
That version managed to boot up and playback audio just fine (no install, just ran the .exe from the bin folder). Haven't had much time to test beyond playing back an old score, but it looks like you've cracked it. :)

heisluft:
I also had this issue. While Musescore does startup as expected with the build linked by @cbjeukendrup, the playback is still high-pitched and faster as usual as described in #15698 that was supposed to hack around this exact problem. Does the linked build include all commits by the pr and if it does, could somebody either re-open this issue or should I open a new one?

DmitryArefiev:
[...]

Issue #15029, opened 2022-12-07
[MU4 Issue] Crash when disconnecting headphones on Windows


krischan941:
**Describe the bug**
Musescore crashes when I disconnect my bluetooth In-Ear-Headphones by putting them back in the in-ears case. It crashes every time doing so.

__Edit by @shoogle:__ it affects other kinds of headphones too, including USB and 3.5mm audio jack.

**To Reproduce**
1. Connect headphones to your PC via Bluetooth, USB, or 3.5mm audio jack
2. start Musescore (and play back some content)
3. pause playback
4. disconnect headphones --> Crash

**Expected behavior**
It doesn't crash and instead maps the output to the PC speakers as my other applications do.

**Platform information**
 - OS: Windows 11 on a Huawei Matebook 2020 AMD

**Additional context**
I installed Musescore via Musehub today.

tjfsteele:
Musescore also won't load if my default audio device (laptop speakers) is selected, but it will if a Bluetooth audio device is selected.

Valdessir:
I experienced a crash as well in the release version with wired earphones. I didn't experience a crash in the beta, nightlies, nor the RC, but an automatic mapping to speakers.

shoogle:
@Valdessir, were your wired earphones USB or standard 3.5mm audio jack?

Valdessir:
@shoogle 

Audio jack.

shoogle:
Ok thanks. It was mentioned in #15308 that it affects USB headphones too, so it appears that the crash occurs whenever the current audio device gets disconnected, regardless of connection type.

tjfsteele:
It also prevents musescore from launching in some cases

brondeau01:
I have the same issue with Musescore 4 crashing on startup.  I was able to get it to work by:
1.  Setting the audio to 16bit 44kHz
2.  Setting my Sound Blaster Z SE to Direct Mode. Normal mode will not work.

I think they may a bit of work to do with the audio engine.  From what I can tell, it won't work set to 24bit 96 or 48k.  Well at least on my system.  (Core I9, 64GB RAM, Windows 11)

MrSimmmons:
I use a pair of Logitech pro X wireless headphones (USB wireless dongle) and have had no issues. Even when swapping the output device to my speakers.
[...]

PR #15698, opened 2023-01-03
WASAPI: use closest supported mix format if error occurs

cbjeukendrup:
Builds on: #15633

and thus resolves: one of the crashes on launch
and additionally resolves: playback problems

Should resolve: https://github.com/musescore/MuseScore/issues/15637
Should resolve: https://github.com/musescore/MuseScore/issues/15685
Should resolve: https://github.com/musescore/MuseScore/issues/15235
Should resolve: https://github.com/musescore/MuseScore/issues/15029
(not 100% sure because it's difficult to tell from those issue descriptions whether it is indeed the same crash that this PR solves or a different one, for example caused by a VST)

And these musescore.org issues:
https://musescore.org/en/node/338043
https://musescore.org/en/node/338149
https://musescore.org/en/node/338271
https://musescore.org/en/node/338378
https://musescore.org/nl/node/339691
https://musescore.org/en/node/340458
https://musescore.org/en/node/340562
https://musescore.org/en/node/341307
https://musescore.org/en/node/341487
and probably there are more reports of the same issue

---

So, after the crash was solved by #15633 by adding proper error handling, it was time to fix the error itself. Turns out that we specify a wave format to the WASAPI device without checking if that format is actually supported by that device. In particular, WASAPI seems not very happy if we specify that we want two channels while the device has more. 

<details>
<summary>EDIT: The following has been postponed:</summary>
What I try to do in this PR is that we check if the format is supported, and if not, use the closest supported format. However, that gives one more problem: because the device needs samples for >2 channels and we only provide samples for 2 channels, the playback is much too fast and therefore also high-pitched; similar to https://github.com/musescore/MuseScore/issues/14921, which seems like it wasn't really fixed yet, so maybe that will be fixed by this PR too.

My solution to that problem is a bit of a "proof of concept" right now; indeed it fixes the problem, but it is a bit ugly. This problem will need some more thought. 
Deeper in the audio module, we seem to rely quite a lot on the assumption that everything has only two channels. For example: `mu::audio::dsp::balanceGain`, the hardcoded 2 in `AudioModule::setupAudioDriver`, and my comment at https://github.com/musescore/MuseScore/pull/14684#discussion_r1030872639. But because everything has a method called `audioChannelsCount()`, this assumption is not very explicit. Not sure whether this is ideal.
And apparently there is also no guarantee that the driver always supports exactly two channels, so we need to do a conversion somewhere. Right now, I have added this in the Wasapi driver itself, but it feels not great. Probably we should better do that on a different level.
</details>

---

But now we do only the following: when an error occurs, we try to use the "closest supported" format. The `IsFormatSupported` method is claimed to be untrustworthy, so we apply trial and error. This may mean that the playback will be high-pitched if the WASAPI device only supports more than two channels (while MuseScore only supports just two). This will need to be fixed later.

### DIFF
--- a/src/framework/audio/internal/platform/win/wasapiaudioclient.cpp
+++ b/src/framework/audio/internal/platform/win/wasapiaudioclient.cpp
@@ -122,13 +122,16 @@ hstring WasapiAudioClient::defaultDeviceId() const
     return MediaDevice::GetDefaultAudioRenderId(AudioDeviceRole::Default);
 }
 
-void WasapiAudioClient::asyncInitializeAudioDevice(const hstring& deviceId) noexcept
+void WasapiAudioClient::asyncInitializeAudioDevice(const hstring& deviceId, bool useClosestSupportedFormat) noexcept
 {
     try {
         // This call can be made safely from a background thread because we are asking for the IAudioClient3
         // interface of an audio device. Async operation will call back to
         // IActivateAudioInterfaceCompletionHandler::ActivateCompleted, which must be an agile interface implementation
         m_deviceIdString = deviceId;
+
+        m_useClosestSupportedFormat = useClosestSupportedFormat;
+        m_deviceState = DeviceState::Uninitialized;
 
         com_ptr<IActivateAudioInterfaceAsyncOperation> asyncOp;
         check_hresult(ActivateAudioInterfaceAsync(m_deviceIdString.c_str(), __uuidof(IAudioClient3), nullptr, this, asyncOp.put()));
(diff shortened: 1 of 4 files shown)


src/framework/audio/internal/platform/win/wasapiaudioclient.cpp
```diff
@@ -536,11 +571,24 @@ void WasapiAudioClient::getSamples(uint32_t framesAvailable)
...
-        m_sampleRequestCallback(nullptr, data, actualBytesToRead);
+    if (actualFramesToRead > 0) {
+        if (clientFrameSize > muFrameSize) {
+            std::vector<uint8_t> tempBuffer(actualFramesToRead* muFrameSize);
+            m_sampleRequestCallback(nullptr, tempBuffer.data(), tempBuffer.size());
+
+            for (uint32_t i = 0; i < actualFramesToRead; ++i) {
+                std::memcpy(data + i * clientFrameSize, tempBuffer.data() + i * muFrameSize, muFrameSize);
+                std::memset(data + i * clientFrameSize + muFrameSize, 0, clientFrameSize - muFrameSize);
+            }
```
cbjeukendrup, line 588: This is the ugly part...

L0uisc: Not that ugly if there's a comment in the code referencing this PR. It makes sense for now for a quick fix for the crashes to just create a bigger sample buffer with blank data for all the extra channels not created by MuseScore.

cbjeukendrup:
@vpereverzev It would be great if we could include a version of this fix in the next patch release :)

Jojo-Schmitz:
Wouldn't it need to get out its Draft state?

Eism (approved):

bkunda:
@DmitryArefiev it seems that #15029 and #15235 have been re-tested and marked as "Done - Tested Again". I can also verify that #15685 has been fixed. I think this only leaves #15637 to re-test (which I can't do because I'm not running Windows 10). Is this all that's now blocking us from being able to mark the present issue (#15698) as "Done - Tested Again"?

L0uisc:
I can test on Win10. I can't promise it will be today, but it will be before Sunday.

DmitryArefiev:
> I can test on Win10. I can't promise it will be today, but it will be before Sunday.
@L0uisc Oh, thanks!

PR #15715, opened 2023-01-05
Fix #15472: Deal with deleted system object staves

### LINKED ISSUE #15472: [MU4 Issue] Crash when reopening score after deleting staff where system objects would appear
**To Reproduce**
1. Create a score **from the Symphony orchestra template**
2. Open the "Add/remove instruments" dialog
3. Remove the strings instruments (violins and everything below)
→ The violins staff was namely one of the staves where system objects would be shown
4. Press OK
5. Save the score
→ The mscx now contains:
    ```xml
        <SystemObjects>
          <Instance staffId="0" barNumbers="false"/>
          </SystemObjects>
    ```
6. Reopen the score
→ `systemObjectStaves` now contains a nullptr; see `Read400::readScore400`, especially the lines:
    - `int staffIdx = e.attribute("staffId").toInt() - 1;` and `sysStaves.push_back(staffIdx);`, which puts a `-1` in `sysStaves`
    - `for (int idx : sysStaves) { score->addSystemObjectStaff(score->staff(idx)); }`, which tries to get the staff at index -1, which returns nullptr
8. Save the score again
→ Crash; see `Score::write`, from the line `if (!systemObjectStaves.empty()) {`

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/15472

This PR does three things:

1. When a system object staff is deleted, that staff is not saved to the file (and if there are no extra system object staves, the `<SystemObjects>` tag is not included in the file at all.
2. When a file is loaded, invalid system object staves (i.e. idx < 2) are ignored
3. When a system object is added to a score, it is not added to deleted system object staves.

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -5340,7 +5340,9 @@ void Score::undoAddElement(EngravingItem* element, bool addToLinkedStaves, bool 
             for (Score* s : scoreList()) {
                 staffList.push_back(s->staff(0)); // system objects always appear on the top staff
                 for (Staff* staff : s->systemObjectStaves()) {
-                    staffList.push_back(staff);
+                    if (staff->idx() != mu::nidx) {
+                        staffList.push_back(staff);
+                    }
                 }
             }
         }
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -5340,9 +5340,11 @@ void Score::undoAddElement(EngravingItem* element, bool addToLinkedStaves, bool 
             for (Score* s : scoreList()) {
                 staffList.push_back(s->staff(0)); // system objects always appear on the top staff
                 for (Staff* staff : s->systemObjectStaves()) {
-                    if (staff->idx() != mu::nidx) {
-                        staffList.push_back(staff);
+                    IF_ASSERT_FAILED(staff->idx() != mu::nidx) {
+                        continue;
                     }
+
+                    staffList.push_back(staff);
                 }
             }
         }
--- a/src/engraving/libmscore/scorefile.cpp
+++ b/src/engraving/libmscore/scorefile.cpp
@@ -168,16 +168,27 @@ void Score::write(XmlWriter& xml, bool selectionOnly, compat::WriteScoreHook& ho
     }
 
     if (!m_systemObjectStaves.empty()) {
-        // write which staves currently have system objects above them
-        xml.startElement("SystemObjects");
+        bool saveSysObjStaves = false;
         for (Staff* s : m_systemObjectStaves) {
-            // TODO: when we add more granularity to system object display, construct this string per staff
-            String sysObjForStaff = u"barNumbers=\"false\"";
-            // for now, everything except bar numbers is shown on system object staves
-            // (also, the code to display bar numbers on system staves other than the first currently does not exist!)
-            xml.tag("Instance", { { "staffId", s->idx() + 1 }, { "barNumbers", "false" } });
+            if (s->idx() != mu::nidx) {
+                saveSysObjStaves = true;
+                break;
+            }
         }
-        xml.endElement();
+        if (saveSysObjStaves) {
+            // write which staves currently have system objects above them
+            xml.startElement("SystemObjects");
+            for (Staff* s : m_systemObjectStaves) {
+                if (s->idx() != mu::nidx) {
+                    // TODO: when we add more granularity to system object display, construct this string per staff
+                    String sysObjForStaff = u"barNumbers=\"false\"";
+                    // for now, everything except bar numbers is shown on system object staves
+                    // (also, the code to display bar numbers on system staves other than the first currently does not exist!)
+                    xml.tag("Instance", { { "staffId", s->idx() + 1 }, { "barNumbers", "false" } });
+                }
+            }
+            xml.endElement();
+        }
     }
 
     xml.context()->setCurTrack(0);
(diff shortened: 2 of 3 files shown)


asattely:
The code style failure is dead wrong, unless I'm doing something wildly incorrect here

cbjeukendrup:
Well, you are correct, but Uncrustify is (pretty wildly) wrong! See https://github.com/musescore/MuseScore/actions/runs/3849233551/jobs/6557988913#step:3:2779 for what it wants you to do. Unfortunately, Uncrustify is the boss...

src/engraving/libmscore/edit.cpp
```diff
@@ -5224,7 +5224,9 @@ void Score::undoAddElement(EngravingItem* element, bool addToLinkedStaves, bool
             for (Score* s : scoreList()) {
                 staffList.push_back(s->staff(0)); // system objects always appear on the top staff
                 for (Staff* staff : s->getSystemObjectStaves()) {
-                    staffList.push_back(staff);
+                    if (staff->idx() != mu::nidx) {
+                        staffList.push_back(staff);
+                    }
```
cbjeukendrup, line 5229: So this one is to ensure that if _still_ some non-existing staff makes it into `systemObjectStaves` (for example because it was left in there after removing a staff?)...

src/engraving/libmscore/scorefile.cpp
```diff
@@ -169,16 +169,27 @@ void Score::write(XmlWriter& xml, bool selectionOnly, compat::WriteScoreHook& ho
...
+            if (s->idx() != mu::nidx) {
+                saveSysObjStaves = true;
+                break;
+            }
+        }
+        if (saveSysObjStaves) {
+            // write which staves currently have system objects above them
+            xml.startElement("SystemObjects");
+            for (Staff* s : systemObjectStaves) {
+                if (s->idx() != mu::nidx) {
```
cbjeukendrup, line 183: ...and these ones are to make sure that non-existing staves don't get written to the file...

src/engraving/rw/read400.cpp
```diff
@@ -156,7 +156,9 @@ bool Read400::readScore400(Score* score, XmlReader& e, ReadContext& ctx)
                     // TODO: read the other attributes from this element when we begin treating different classes
                     // of system objects differently. ex:
                     // bool showBarNumbers = !(e.hasAttribute("barNumbers") && e.attribute("barNumbers") == "false");
-                    sysStaves.push_back(staffIdx);
+                    if (staffIdx > 0) {
+                        sysStaves.push_back(staffIdx);
+                    }
```
cbjeukendrup, line 161: ...and this one is to ensure that we don't insert nullptrs into `systemObjectStaves` when reading a file that was saved before this PR, right?

cbjeukendrup (commented):
To quickly check if I correctly understand the purpose of each change:

src/engraving/libmscore/edit.cpp
asattely: That's correct--even when the staff is removed it remains in memory as part of the undo stack, I believe.

src/engraving/libmscore/scorefile.cpp
asattely: Yep! We don't save zombie staves to the file because the file doesn't save the undo stack.

src/engraving/rw/read400.cpp
asattely: Yep!

asattely:
Re: uncrustify:
Okay, I'll see if I can revert that line of code. I didn't even change that line, it's just that my uncrustify seems to do it correctly for some reason so there are a few landmine code files that if I change anything in them uncrustify goes berserk. I'll just have to disable the hook for that particular commit.

cbjeukendrup:
The code style check uses version 0.73 of Uncrustify, so you might want to make sure that your commit hook uses that too. 

(We decided that it's the most convenient to use a fixed version, because each version behaves slightly differently, and on GitHub there's some delay before it uses updated versions, and it's inconvenient if every few weeks your local hook already has the update but the code style check not yet.)

oktophonie (approved):

RomanPudashkin:
@asattely please rebase it

src/engraving/libmscore/edit.cpp
cbjeukendrup: Since @mike-spa's work about system objects, is the situation that `systemObjectStaves` contains an invalid staff still to be expected? Or should we turn this into an assertion? For example, 
```suggestion
                    IF_ASSERT_FAILED(staff->idx() != mu::nidx) {
                        continue;
                    }
                    
                    staffList.push_back(staff);
```

src/engraving/libmscore/scorefile.cpp
cbjeukendrup: Same question here actually

cbjeukendrup:
I think part of this PR is still needed to deal with existing scores from 4.0.0 where it is wrong (i.e. scores that already contain 
```xml
    <SystemObjects>
      <Instance staffId="0" barNumbers="false"/>
      </SystemObjects>
```

src/engraving/libmscore/edit.cpp
asattely: Real good point! I tested that change with a rebase and it does appear that system object staves are being deleted from the list when they're removed. I'll make this change here and also the other place you indicated below.

src/engraving/libmscore/scorefile.cpp
cbjeukendrup: It looks like an empty `SystemObjects` tag won't do any harm, and it is very unlikely to happen anyway (indeed it would happen only when something goes wrong). So I don't think it's worth the extra loop. But I'm happy with any decision :)

In any case, I think it would be nice to replace each `if (s->idx() != mu::nidx) {` with `IF_ASSERT_FAILED(s->idx() != mu::nidx) { continue; }`, to make clear that this is an unexpected scenario.

asattely: fair enough, changed!

cbjeukendrup (approved):

Issue #15752, opened 2023-01-08
Properties > Fret diagrams: when applying scale value to style, in fact 100x the scale value is applied to style


cbjeukendrup:
**Describe the bug**
In the Inspector, in some cases, there is a conversion between the value displayed in the UI and the internal value. For example, a scale value is internally stored as a real number, but in the UI presented as a percentage. But today, while digging in the code, I found something strange: this conversion is not applied when applying the current value to style. This results for example in the following bug:

**To Reproduce**
1. Create score
2. Add fret diagram
3. In Properties > Settings, change scale
4. Apply to style
-> Boom

**Screenshots**

https://user-images.githubusercontent.com/48658420/211175424-c4adb403-8687-431d-af52-f66194438e06.mov

**Additional context**
This was simply the first example I found, but this may very well apply to other elements and properties too; maybe with even more janky results.

cbjeukendrup:
It seems this has been fixed (by e6457fb41ce568ef882444b5aa42293f751eec97). @DmitryArefiev or @zacjansheski could one of you confirm that? (retesting in 4.2 build should be enough)

zacjansheski:
Fixed in 4.2

PR #15772, opened 2023-01-08
Inspector offset property: fix editing x and y value individually when multiple elements are selected with potentially different values for x and/or y

### LINKED ISSUE #15452: [MU4 Issue] X and Y offset values cannot be managed separately when multiple items are selected
**Describe the bug**
Bear with me on this one...

If I select two items and they have the same X _and_ Y offset, then those values will be shown in the Appearance popup, as one might expect:
issue-1

Let's say I change the X offset of one of these items:
issue-2

Now, if I select the two items, the X and Y boxes both show "--", which normally indicates 'there are multiple values':
issue-3
This is technically only true for the X offset, because the Y offset is the same for both. It would, therefore, be nice that if only one of the values is consistent for all the selected items, that this value be displayed.

The real problem is that it is impossible in this situation to apply the same X or Y offset to all of the selected items at once (which is not a particularly unusual thing to want to do; and to have to do it for a whole load of items one by one is very tedious).

As an example, let's say I wanted to set the Y offset of both the selected items to 3, but preserve the different X offsets that each one has. The first oddness is that when I select the value in the box it changes from '--' to '0':
issue-4

And if I then change this value to 3, it will apply a Y offset of 3 to both items as expected, but also will reset the X offset of all the items to 0!
issue-5

**Expected behaviour**
Ideally,

- when multiple items are selected, if all items have the same X offset, or the same Y offset, then this value should be shown in the box
- editing such a value should affect only that offset (X or Y), and have no influence on the other one, for any of the selected items.

**Platform information**
 - 4.0 release, tested on Linux (Arch) and Win 11

cbjeukendrup:
Resolves: #15452 
~Res0lves: #15752~ (already fixed meanwhile)

To fix issue #15452, we need to do two things:
1. Create separate property items for x and y (and do something ugly to make resetting and applying to style work correctly)
2. Create a special setter function that work per element, so that if you set the x value, for each element individually the y value will be preserved.

So we need to add another customisation point to `buildPropertyItem`.

We can also see that different properties need different kinds of customisation possibilities:
- some need nothing
- some only need to specify a simple element-agnostic function to convert from UI value to internal PropertyValue
- some need to specify an element-specific function to convert from UI value to internal PropertyValue; in those cases we need to force the caller to specify a way to convert to a style value too, which solves [#15752](https://github.com/musescore/MuseScore/issues/15752).
- some need to specify a fully custom setter function; often they only need to call some extra method after calling the default behaviour, so we add a convenient way to retrieve the default function. 

In this PR, I added all these ways. I also eliminated an unnecessary conversion from QVariant to QVariant, by allowing to specify a direct conversion between PropertyValue and QVariant.


ghost (approved):

cbjeukendrup:
Finally rebased this dreadful PR. After looking at it again, I think it's not that dreadful at all; it may look a bit complicated, but it isn't really. Also, the complicatedness is mostly confined to AbstractInspectorModel; any attempt to simplify that, would result in some awkwardness elsewhere, so it's a bit of a trade-off.
However, there are still some refinements to be made, so I'm marking it as draft.

PR #15775, opened 2023-01-09
Add missing includes (mostly Qt)

thesamesam:
Fix compile issues because of missing includes.

  Add missing includes (mostly Qt)

  When compiling MuseScore, I hit various compile failures because
  of missing includes. I suspect it built before because of either
  unity builds or transitive includes which change between versions
  of Qt.

  e.g.:
  ```
  src/workspace/workspace_autogen/OTUN7GXZT4/moc_newworkspacemodel.cpp:163:29: error: variable ‘QVariant _r’ has initializer but incomplete type
  [...]
  /var/tmp/portage/media-sound/musescore-9999/work/musescore-9999/thirdparty/beatroot/AgentList.h:39:7: error: ‘size_t’ does not name a type
     39 |       size_t size() { return list.size(); }
        |       ^~~~~~
  ```

- [ ] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)

### DIFF
--- a/src/appshell/iapplicationactioncontroller.h
+++ b/src/appshell/iapplicationactioncontroller.h
@@ -22,6 +22,9 @@
 #ifndef MU_APPSHELL_IAPPLICATIONACTIONCONTROLLER_H
 #define MU_APPSHELL_IAPPLICATIONACTIONCONTROLLER_H
 
+#include <QDragEnterEvent>
+#include <QDropEvent>
+#include <QDragMoveEvent>
 #include <QEvent>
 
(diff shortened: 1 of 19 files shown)


igorkorsukov (approved):

Jojo-Schmitz:
Oh boy, this introduced a whole bunch of new compiler warnings

Jojo-Schmitz:
Forget about it, after a 'make clean' these are gone, sorry for the noise

thesamesam:
No problem.

Issue #15561, opened 2022-12-24
[MU4 Issue] Problems with measure rest and multibar rest width


oktophonie:
In this score (3x 3/4 bars followed by a 4/4 bar, repeated)
![image](https://user-images.githubusercontent.com/20604032/209440255-fddac7b6-dd5f-4b74-92fe-03755aca3460.png)
turning multirests on (press 'M') demonstrates a few issues:
![image](https://user-images.githubusercontent.com/20604032/209440275-cab77c1e-224b-4028-b34f-89d60bbbf587.png)

Most obviously, the first system is escaping its bounds.

Secondly, the multirests are narrower than the single bars, which is very confusing. A single bar may be _as_ narrow, perhaps, but shouldn't be narrower.

Lastly, the first 3-bar rest in each system is wider than the others, presumably because of how the system header is/is not accounted for in the calculations.

PR #15778, opened 2023-01-09
Fix MMRest bug

mike-spa:
Resolves: #15561 

The issue was, in part, due to a couple of errors in the handling of the MMRest themselves (specifically, not correctly taking into account items _before_ the MMRest but in the same bar) . Those are now fixed. The vtests differences will show the corrected results.

The other part is more trivial, i.e. that the minimum width for MMRests is set by default to 4 sp (Style -> Rests). This makes no sense, given that the minimum width for normal measures is 8sp and MMRests should never be narrower than normal measures. I propose raising the default value to 6sp. Considering that MMRests also have two enforced margins that are set by default at 1.2sp, the resulting minimum possible width of an MMRests measure is 8.4sp, which ensures they can never be narrower than a single empty measure.

### DIFF
--- a/share/templates/01-General/01-Treble_Clef/score_style.mss
+++ b/share/templates/01-General/01-Treble_Clef/score_style.mss
@@ -345,7 +345,7 @@
     <concertPitch>0</concertPitch>
     <createMultiMeasureRests>0</createMultiMeasureRests>
     <minEmptyMeasures>2</minEmptyMeasures>
-    <minMMRestWidth>4</minMMRestWidth>
+    <minMMRestWidth>6</minMMRestWidth>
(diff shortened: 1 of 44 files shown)


oktophonie (approved):

RomanPudashkin:
@mike-spa any new vtests/utests?

mike-spa:
@RomanPudashkin the changes show up in many of the existing vtests already (basically, all that contain MMRests), so there's no need to add more. They are going all green just cause @oktophonie put the vtest label on this PR :)

oktophonie:
(I even checked the vtests again to be sure, the changes still all look beautiful to me.)

RomanPudashkin (approved):

PR #15781, opened 2023-01-09
Fix #12970: Limit time signature numerator to maximum 63

### LINKED ISSUE #12970: [MU4 Issue] Too big custom time signature causes the program to stop responding
**The bug**
Program stops responding when entering too big numbers (e.g. 1000) in a custom time signature. Same issue in v3.6

**To Reproduce**
1. Open the window for creating a custom time signature from the palette
2. Enter a big number (e.g. 1000) 

create_time_signature_window

**Platform information**
 - Windows 11

**Suggestion for fix**
Give the Value field a range like in the New score wizard (1 - 63)

rzvncj:
Resolves: #12970

This change sets an upper limit of 63 to the value of the numerator of new custom time signatures. If this limitation is not enforced, the application will hang when large values  (e.g. 1000) are typed in.

### DIFF
--- a/src/palette/view/widgets/timedialog.ui
+++ b/src/palette/view/widgets/timedialog.ui
@@ -79,7 +79,7 @@
               <number>1</number>
              </property>
              <property name="maximum">
-              <number>999999</number>
+              <number>63</number>
              </property>
              <property name="value">
               <number>4</number>


bkunda:
Is this ready for testing?

Currently I see this in Checks:

Screenshot 2023-01-11 at 12 00 54 pm

Jojo-Schmitz:
I believe you can (re)start the CI actions?

rzvncj:
> Is this ready for testing?
> 
> Currently I see this in Checks:
> 
> Screenshot 2023-01-11 at 12 00 54 pm

This is probably not a question for me, but I did test it manually before pushing and opening the pull request up. I don't know if it's supposed to trigger automated testing or not.

oktophonie:
The checks should run automatically, but sometimes GitHub just doesn't bother for some reason.
We can only manually re-run them from the site if they've already run once - if they haven't run at all I don't think we can do anything.

Adding another commit (even empty) should trigger them though, with any luck.

Jojo-Schmitz:
`git commit --amend; git push -f`

rzvncj:
Here you go.

bkunda:
Superb. Thanks all. Will test when everything's ready.

Jojo-Schmitz:
Hmm, such large numerator`s are possible in Mu3 and do work there, no hang, so this PR here seems to cure the symptoms rather than the real cause?

rzvncj:
> Hmm, such large numerators are possible in Mu3 and do work there, no hang, so this PR here seems to cure the symptoms rather than the real cause?

That was the requested fix in the description of https://github.com/musescore/MuseScore/issues/12970 (see "Suggestion for fix"). Of course, it's up to you if this pull request is merged or not, I understand if you want a different way to go (in which case it'd probably be good to update the description of the issue).

Either way, I don't really see a large numerator actually being useful in the real world, but perhaps they are for somebody.

rzvncj:
> Hmm, such large numerators are possible in Mu3 and do work there, no hang, so this PR here seems to cure the symptoms rather than the real cause?

#12970 also says _"Same issue in v3.6"_, so that's a bit confusing. I don't have a v3 MuseScore handy so I can't claim it's one way or the other.

Jojo-Schmitz:
I just check in Mu3 (3.6.2, to be exact) and 9999 as a numerator of **a measure property** a) is possible and b) just works

Jojo-Schmitz:
As a timesig it indeed hangs, so I rest my case

bkunda:
🟢 Tested on macOS. Fixed. 
The fact that the maximum value is now consistent with the equivalent field in the New Score dialog is I think satisfactory. We can discuss accommodating higher values if there are reasonable use cases that require them.
Thank you @rzvncj!

@RomanPudashkin FYI

rzvncj:
> green_circle Tested on macOS. Fixed. The fact that the maximum value is now consistent with the equivalent field in the New Score dialog is I think satisfactory. We can discuss accommodating higher values if there are reasonable use cases that require them. Thank you @rzvncj!

Happy to help! Thanks for the quick review / test!

oktophonie:
https://en.wikipedia.org/wiki/List_of_musical_works_in_unusual_time_signatures#Upper_number_more_than_59
It would be better not to limit this just to avoid encountering a bug, if we can fix the underlying problem itself.

bkunda:
> https://en.wikipedia.org/wiki/List_of_musical_works_in_unusual_time_signatures#Upper_number_more_than_59 https://en.wikipedia.org/wiki/List_of_musical_works_in_unusual_time_signatures#Combined_unusual_signatures
> 
> OK, it's not common, but it would be better not to limit this just to avoid encountering a bug, if we can fix the underlying problem itself.

@RomanPudashkin when you have some time, can you please look further into this?

rzvncj:
Please feel free to discard and close this pull request if you decide to go with another way of fixing the issue.

Issue #15601, opened 2022-12-26
[MU4 Issue] Compile for ARM64


sclsj:
**Describe the bug**
Following the default commands for compiling results in failed compilation because the code is in x86_64 while external libraries are arm64. I cannot find an option to compile for arm64.

**To Reproduce**
1. Apple Silicon mac, arm64 homebrew, install deps
2. Clone git
3. `cmake -P build.cmake -DCMAKE_BUILD_TYPE=Release`
4. Build failed.

**Expected behavior**
Default target is arm64 on arm64 systems.

**Screenshots**
N/A

**Platform information**
 - OS: macOS Monterey

**Additional context**
Error output:
```
[ 49%] Linking CXX executable global_tests
ld: warning: ignoring file /opt/homebrew/opt/qt@5/lib/QtNetworkAuth.framework/QtNetworkAuth, building for macOS-x86_64 but attempting to link with file built for macOS-arm64
<lots of content removed because error log exceeds maximum issue length>
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [src/framework/mpe/tests/CMakeFiles/mpe_test.dir/build.make:183: src/framework/mpe/tests/mpe_test] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:4046: src/framework/mpe/tests/CMakeFiles/mpe_test.dir/all] Error 2
[ 49%] Linking CXX static library libui.a
[ 49%] Built target ui
gmake: *** [Makefile:146: all] Error 2
CMake Error at build.cmake:249 (message):
  Build step failed with status 2.  See output above for details.
```

cbjeukendrup:
Building for ARM architecture with Qt 5 is not yet possible. There is a very experimental Qt 6 branch where it _is_ supported, see #10108. But the problems with Qt 6 are still so big, that I don't expect we will switch to it until MuseScore 5.

When I have time, I will investigate it once more though, since HomeBrew now indeed conveniently provides an ARM version of Qt 5. But last time I tried with this version, I got a lot of errors that weren't immediately clear.

In your case, the problem is that you have installed the ARM version of Qt but you are building MuseScore in x86_64 mode (because that's hardcoded in the CMake files).

[...]

PR #15785, opened 2023-01-09
Support building for arm64 on Apple Silicon Macs with arm64 Qt version

cbjeukendrup:
Resolves: #15601

On Macs with Apple Silicon, this PR makes it as simple as this:
- Want to build in arm64 mode? Install dependencies in arm64 mode (for example, just via HomeBrew, which conveniently provides Qt 5.15.8) and enable the CMake option `BUILD_MACOS_APPLE_SILICON`.
- Want to build in x86_64 mode? Install dependencies in x86_64 mode (for example, using the x86_64 version of HomeBrew, or via the official Qt installer) and make sure that CMake looks for the x86_64 versions, for example by setting `CMAKE_PREFIX_PATH`.

(This won't enable us to create arm64 releases, because GitHub actions does not yet support Apple Silicon. To make arm64 releases, we would need to create a universal binary of Qt, for example by merging the arm64 and x86_64 versions from HomeBrew using the `lipo` tool, and the same for the other dependencies, and then tell the compiler to cross-compile to create universal binaries. This is far from trivial though. However, the good news is that it seems eventually possible to create universal builds with Qt 5, which is way more realistic than switching to Qt 6 at the moment.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,7 @@ set(MUSESCORE_BUILD_CONFIG "dev" CACHE STRING "Build config")
 # - testing - for testing versions (alpha, beta, RC)
 # - release - for stable release builds
 
+option(BUILD_MACOS_APPLE_SILICON "Build for Apple Silicon architecture. Only applicable on Macs with Apple Silicon, and requires suitable Qt version." OFF)
 option(BUILD_64 "Build 64 bit version of editor" ON)
 option(BUILD_PORTABLEAPPS "Windows build for PortableApps.com" OFF)
 option(BUILD_FOR_WINSTORE "Build for the Windows Store." OFF)
--- a/build/cmake/SetupBuildEnvironment.cmake
+++ b/build/cmake/SetupBuildEnvironment.cmake
@@ -132,8 +132,19 @@ endif()
 
 # APPLE specific
 if (OS_IS_MAC)
-      set(CMAKE_OSX_ARCHITECTURES x86_64)
-      set(MACOSX_DEPLOYMENT_TARGET 10.14)
-      set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14)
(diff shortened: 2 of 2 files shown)


igorkorsukov (approved):

igorkorsukov (approved):

Arnie97:
> GitHub actions does not yet support Apple Silicon

[New M1 runners have just been introduced in October.](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/) Unfortunately, they're only available for GitHub Team or GitHub Enterprise Cloud plans. Anyway it still helps: the self-hosted runner image now becomes available for Apple Silicon.  I'll try it on my MacBook Pro to see if I can get a easily reproducible env for compiling MuseScore.

Arnie97:
> GitHub actions does not yet support Apple Silicon

[New M1 runners have just been introduced in October.](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/) Unfortunately, they're only available for GitHub Team or GitHub Enterprise Cloud plans. Anyway it still helps: the self-hosted runner image now becomes available for Apple Silicon.  I'll try it on my MacBook Pro to see if I can get an easily reproducible env for compiling MuseScore.

cbjeukendrup:
It should be possible to create a _Universal_ (Intel and Apple Silicon in one) build using GitHub Actions. We have switched to Qt 5.15.9, which supports Universal builds. The only thing that's missing is Universal binaries for some other dependencies. I'm working on that; there is a chance that this is going to happen for MS4.2!
If you have a M1 Mac yourself, and you install all dependencies (i.e. Qt and libsndfile) via arm64 HomeBrew to get the arm64 versions, it should be trivial to compile MuseScore for Apple Silicon, you just need to enable the `MUE_COMPILE_BUILD_MACOS_APPLE_SILICON` CMake option.

Issue #15796, opened 2023-01-10
[MU4 Issue] Severe crash and file corruption on templates with duplicate system marks


mike-spa:
**To Reproduce**
1. Create a new file from one of the templates which have duplicate system marks (for example: SATB + Piano)
2. Save the file
3. Open the instrument panel, and delete the instrument which is supposed to hold the duplicate system mark (in this case, the Piano)
4. Try to insert from palette any system text or object (tempo mark, or rehearsal mark, or whatever) : **program will crash**
5. Reopen the program after the crash, and reopen the the file
6. **The program will now crash any time you try to save the file**, which means the entire file is now unusable.

Edit: the crash can actually be reproduced also without steps 3 & 4.

cbjeukendrup:
Seems related to https://github.com/musescore/MuseScore/issues/15735

DmitryArefiev:
Fixed in #15806

DmitryArefiev:
> Seems related to #15735

@cbjeukendrup Unfortunately crash with segno still occurs in #15806

PR #15806, opened 2023-01-10
Fix crash on deleting "system object" staff

mike-spa:
Resolves: #15796 

When deleting a staff, we weren't deleting it from the list of "system object staves", so further down the line we end up trying to reference a staff that doesn't exist anymore.

@vpereverzev @Tantacrul this one definitely needs to go into the patch release

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -5153,6 +5153,8 @@ void Score::undoRemoveStaff(Staff* staff)
 
     staff->undoUnlink();
 
+    mu::remove_if(systemObjectStaves, [staff](Staff* s){ return s == staff; });
+
     undo(new RemoveStaff(staff));
 }
 


oktophonie (approved):

src/engraving/libmscore/edit.cpp
```diff
@@ -5153,6 +5153,9 @@ void Score::undoRemoveStaff(Staff* staff)
 
     staff->undoUnlink();
 
+    // If the staff we're removing is a "system object staff", it must be removed from the list
```
vpereverzev, line 5156: the comment is, essentially, a duplication of what I can see from the code itself. Instead, I'd prefer to see explanation here of "why do we need to remove it here"

mike-spa: Uhm, but I guess "why remove it here" is also self-explaining, given that the function is called "removeStaff"? Or did I misunderstand? Maybe I'd better just remove the comment

vpereverzev: yeah, that was just leading to unnecessary confusion

vpereverzev:
What's the deal with unit-tests? @mike-spa

mike-spa:
> What's the deal with unit-tests? @mike-spa

As far as I can tell, that test isn't even using this function, so I'm quite confused. Let me try to re-run it

vpereverzev:
@DmitryArefiev and this one too, should be quite easy to test

vpereverzev (approved):

vpereverzev:
> > What's the deal with unit-tests? @mike-spa
> 
> As far as I can tell, that test isn't even using this function, so I'm quite confused. Let me try to re-run it

which one wasn't working? Looks like we have some weird test with random results in master.. (judging by other opened PRs)

mike-spa:
> which one wasn't working?

Engraving_PartsTest::doRemoveMeasureRepeat()

vpereverzev:
> > which one wasn't working?
> 
> Engraving_PartsTest::doRemoveMeasureRepeat()

Try to launch it locally with and without your changes to make sure

vpereverzev:
@mike-spa alright, it's not your fault at all. So we can merge this once @DmitryArefiev will have this tested

@cbjeukendrup looks like you have turned on these tests recently, so that now they randomly fail in all the fresh PRs. See https://github.com/musescore/MuseScore/pull/15821 or https://github.com/musescore/MuseScore/actions/runs/3882835455 for example

I'm suggesting to disable them for now and figure out why they're so unstable

DmitryArefiev:
Tested #15796 on Win10, Mac12, LinuxMint 22.04 - **FIXED**

vpereverzev:
@mike-spa make another one for 4.0.1

PR #15815, opened 2023-01-11
fix #15292: Opening a score (clicking its file) launches the homepage instead of opening the score view (macOS only)

### LINKED ISSUE #15292: [MU4 Issue] Opening a score (clicking its file) launches the homepage instead of opening the score view (macOS only)
Haven't seen a report of this which I'd find odd, considering opening a file is pretty common to do.

**To Reproduce**
1. Ensure MS4 is not running. Scores open properly if there is an existing instance running
1. Open any MS4 score (older score versions prompt the update notice and open the score properly).
2. Note that the program launches to the homepge. It does not navigate to the Score tab nor does it show it there either.

**Expected Behavior**
Opening a score should open an instance that presents the Score view.

**Platform information**
 - OS: macOS Monterey 12.6

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/15292

### DIFF
--- a/src/appshell/appshellmodule.cpp
+++ b/src/appshell/appshellmodule.cpp
@@ -200,6 +200,14 @@ void AppShellModule::onInit(const IApplication::RunMode&)
 #endif
 }
 
+void AppShellModule::onAllInited(const framework::IApplication::RunMode&)
+{
+    //! NOTE: process QEvent::FileOpen as early as possible if it was postponed
+#ifdef Q_OS_MACOS
+    qApp->processEvents();
+#endif
+}
+
 void AppShellModule::onDeinit()
 {
     s_sessionsManager->deinit();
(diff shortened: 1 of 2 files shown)


vpereverzev (approved):

src/appshell/internal/applicationactioncontroller.cpp
```diff
@@ -41,6 +41,11 @@ using namespace mu::actions;
 void ApplicationActionController::preInit()
 {
     qApp->installEventFilter(this);
+
+    //! NOTE: process QEvent::FileOpen as early as possible if it was postponed
+#ifdef Q_OS_MACOS
+    qApp->processEvents();
+#endif
```
igorkorsukov, line 48: I think it's better to call `processEvents` in `onAllInited` because in `preInit` modules are not initialized

RomanPudashkin: Yes, let's try

DmitryArefiev:
Tested #15292 on Mac13.1 - **FIXED**

bkunda:
🟢 Tested on macOS. Fixed.

Issue #15204, opened 2022-12-15
[MU4 Issue] Numeric keyboard and standard numeric keys are broken (specific AZERTY?)


cadiz2:
**Describe the bug**
As written in the title, numeric keyboard and standard numeric keys don't respond at all, or under certain circumstances 

**To Reproduce**
1. Create a score for piano (Treble clef) and Guitar Tablature
2. Input notes in the piano staff and try to change note values or/and input rests ("0")  with the numeric keyboard. Result: failure. 
4. Now input notes in the TAB staff
5. Surprise: the numeric keyboard works, I can choose different notes values
6. BUT: try to enter fret numbers with standard numeric keys. New failure. To get a result, I am forced to use letters A for 0, B for 1, C for 2 etc. 

**Expected behavior**
Numeric keyboard and numeric keys work as expected (ie as version 3.6.2)

**Platform information**
 - Windows 10

**Additional context**
I don't know if it's a specific AZERTY keyboard issue. I would say probably because this problem had already appeared when version 3 was released, and a bug report on the French forum confirmed this malfunction. 
Also: https://musescore.org/en/4.0#comment-1153340

opjazz:
The numeric keyboard on Mac OS doesn't work yet.

deadwanderer:
Just downloaded 4.0 and noticed this same thing. Setting note duration using numpad keys no longer works on 4.0. Can still set them using the number keys above the regular keyboard, but this is way less usable than numpad for duration and keyboard for note values.

Flying-Roger:
I noticed the same problem, really blocking for me as I constantly use the keypad when encoding or modifying scores.
After some investigation, I found out that this can simply be solved by exporting the key shortcuts definitions into an xml file and editing it to add add a few definitions, like this one:
    <SC>
        <key>pad-note-2</key>
        <seq>6</seq>
        <seq>Num+6</seq>
    </SC>
Actually, these definitions, like "Num+6", are missing for the pad-note-x definitions, while they are already present for the tab-pad-note-x definitions, for example:
    <SC>
        <key>pad-note-2-TAB</key>
        <seq>Shift+6</seq>
        <seq>Num+6</seq>
    </SC>
I recommend that all these definitions are included in the default shortcuts of the released version of Musescore 4.

liferesearcher57:
Yes, I'm having the same issue. But I was hoping I wouldn't have to go through all of that export xml file etc. Can't they just fix the problem?

[...]

Issue #14505, opened 2022-11-13
[MU4 Issue] Shortcut file Mismatch beetween Num+ and NumPad+


Gai-Luron:
**Describe the bug**
Shortcut file Mismatch beetween Num+ and NumPad+

**To Reproduce**
1. Réinit MuseScore préférences
2. Numpad 1,2 3 dosn't work in french keyboard
3. Export Shortcuts to xml file
4. Load in xml editor shortcuts file
5. All numpad shortcuts have been generated with "Num+".
6. Replace All "Num+" with "NumPad+"
7. Import modified shortcuts file
8. Numpad 1,2 3 dosn't work anyway
9. exit from Musescore 4
10. Run MuseScore 4
11. Numpad 1,2,3 Work Now

Not that NumPad shortcut dosn't  appear in shorcut préférences but work.
![image](https://user-images.githubusercontent.com/1835871/201520101-77ef64d9-aa09-4d86-ac07-5f54cd5676f0.png)

OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223170411, revision: github-musescore-musescore-7b85777

Eism:
@Gai-Luron 
Please check the build from https://github.com/musescore/MuseScore/pull/15819

Jojo-Schmitz:
Tested: durations (1-9) work, /, +, - and Enter works, but . (resp. , in my case) does not,
Windows 11, German QWERTZ keyboard, German language settings

Changing to
```
    <SC>
        <key>pad-dot</key>
        <seq>.</seq>
        <seq>Num+.</seq>
        <seq>Num+,</seq>
    </SC>
 ```
makes 'dot' work for me

NumPad+0 doesn't work for Rest 
```
    <SC>
        <key>rest</key>
        <seq>0</seq>
        <seq>Num+0</seq>
    </SC>
 ```
fixes that. Both are vital shortcuts that did work in MU3 out of the box

Issue #12956, opened 2022-08-22
[MU4 Issue] Numpad enter doesn't add line/system break


HarryAllen1:
**Describe the bug**
When a line-break needs to be added, the enter-key can be used to add one. However, the numpad enter doesn't add a line break (both with and without num-lock) 

**To Reproduce**
1. Create a new score
2. Select a measure
3. Press numpad enter

**Expected behavior**
A line-break is added

**Platform information**
 - OS: Windows 11, nightly build

Jojo-Schmitz:
None of the NumPad keys seems to work, e.g. note entry and entering durations. 
 "+" for tie works though, as does "/" for adding grace notes, Acciaccaturas

sammik:
It is the same, as in MU3 (NumPad Enter didnt work in 3.6.2, Linux)

Eism:
Enter and Return are different keys. And system break should not be called on Enter if Return is assigned to it.

I can assign Enter for the system break in MU4 and it works.

@DmitryArefiev please test on your machine
If the reassignment of Return to Enter works for you, then let's close this issue.

Jojo-Schmitz:
IIRC in Mu3 both, Return and Enter, were assigned to system break

Eism:
> IIRC in Mu3 both, Return and Enter, were assigned to system break.
No. Only Return

Jojo-Schmitz:
OK, I just double checked and stand corrected ;-)

Still it should though

Eism:
So, Is it not a bug? Because the behavior is the same as in MU3

Jojo-Schmitz:
The Bug is that you can't make Mu4 using NumPad Enter, trying to do so just exits the dialog.
2nd bug is that it apparently is not possible to __add__ another shortcut, just replaing one for another seems possible

Jojo-Schmitz:
All mentioned in #issuecomment-1239684980, 4 months ago...

Eism:
The first bug I fixed - see https://github.com/musescore/MuseScore/pull/14715
2nd bug - https://github.com/musescore/MuseScore/issues/13324

Jojo-Schmitz:
No, that 2nd bug is not about shortcut sequences (e.g. pressing a, then b, then c), separatend by commas, but about alternative shortcuts, separated be semicolons

Eism:
@Jojo-Schmitz see https://github.com/musescore/MuseScore/pull/12347#issuecomment-1184183216

[...]

PR #15819, opened 2023-01-11
fixed #12791: Num pad not responding for duration selection

### LINKED ISSUE #12791: [MU4 Issue] Num pad not responding for duration selection
**Describe the bug**
Using num pad no longer works for selecting durations. I have checked Num Lock (makes numpad into arrows which navigate around buttons or notes). The problem started after I had been experimenting with using MuseScore with a MIDI keyboard. To remove the MIDI keyboard from the input slot I also clicked "Reset Preferences", so it may be due to that.
I tried re-adding the numbers as shortcuts, but that didn't do it.
Interesting point is that the numpad dot still selects dotted note, and + for ties and / for acciaccaturas still work.

I can only guess that Reset Preferences does this, but haven't been able to test from a starting point again.

**Expected behavior**
Use num pad numbers same as number row to select durations

**Platform information**
 - OS: Windows 10
 - MuseScore 4 Alpha

Eism:
Resolves: #12791
Resolves: #12956
Resolves: #15204
Resolves: #14505

### DIFF
--- a/src/framework/shortcuts/data/shortcuts-Mac.xml
+++ b/src/framework/shortcuts/data/shortcuts-Mac.xml
@@ -291,6 +291,7 @@
   <SC>
     <key>rest</key>
     <seq>0</seq>
+    <seq>Num+0</seq>
     </SC>
   <SC>
     <key>realtime-advance</key>
@@ -315,6 +316,7 @@
   <SC>
     <key>sharp</key>
     <seq>+</seq>
+    <seq>Num++</seq>
     </SC>
   <SC>
     <key>nat</key>
@@ -323,6 +325,7 @@
   <SC>
     <key>flat</key>
     <seq>-</seq>
+    <seq>Num+-</seq>
     </SC>
   <SC>
     <key>stretch+</key>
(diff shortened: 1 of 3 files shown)


RomanPudashkin (approved):

Jojo-Schmitz:
Could you add Numpad+Enter as an additional shortcut for System Break?
Would fix #12956

Eism:
@Jojo-Schmitz thanks, makes sense for 
And NumPad + dot for dotted duration?
And maybe NumPad ++/- for sharp/flat?
Maybe NumPad + / 

Did not find a sequence *

Jojo-Schmitz:
Double/half duration, W and Q. Also with Shift

Jojo-Schmitz:
OTH instead of flat/sharp I'd rather have NumPad + for Tie, maybe Numpad + - for Slur? Would allow easy note entry with one hand on Numpad and the other on mouse, excatly my favourite workflow

Jojo-Schmitz:
> Why we should use NumPad+/ and NumPad+* for half and double duration?

Because those are otherwise unused? I'm open for better suggestions though. Not using them at all seems a waste.

Jojo-Schmitz:
Maybe acciaccatura/appoggiatura? Might even be better.

Eism:
For other changes please create a separate issue

Jojo-Schmitz:
I'd need to check, but maybe add NumPad + , for those locales/keyboads that use , as the decimal separator

Jojo-Schmitz:
Tested: durations (1-9) work, /, +, - and Enter works, but . (resp. , in my case) does not,
Windows 11, German QWERTZ keyboard, German language settings

Changing to
```
    <SC>
        <key>pad-dot</key>
        <seq>.</seq>
        <seq>Num+.</seq>
        <seq>Num+,</seq>
    </SC>
 ```
makes 'dot' work for me. Needed for all of continental Europe as far as I can tell

NumPad+0 doesn't work for Rest 
```
    <SC>
        <key>rest</key>
        <seq>0</seq>
        <seq>Num+0</seq>
    </SC>
 ```
fixes that. Both are vital shortcuts that did work in MU3 out of the box

So we're almost there ;-)

Jojo-Schmitz:
Great! Now Mac and AZERTY users would need to test I guess

Jojo-Schmitz (approved):

Jojo-Schmitz:
Darn, that dreaded spurious unit test failure again...

Flying-Roger:
See my recent post on https://github.com/musescore/MuseScore/issues/15204:

I noticed that there is a discrepancy between the way detects the numeric keypad while editing a score with respect to what it does while defining shortcuts:

while defining a shortcut, pressing 0 on the alphanumerical keyboard or pressing 0 on the numeric keypad has exactly the same effect: "0" is recorded for the shortcut.
while editing a score, pressing 0 on the alphanumerical keyboard does what is defined as "0" in the shortcut xml file, while pressing 0 on the numeric keypath does is not detected as "0", but as "Num+0", thereby requiring a specific definition in the shortcut xml file.
This is a bug : the consequence is that it is impossible to define a shortcut on the numerical keypad via the shortcut definition menu. The workaround is to export the shortcuts in xml, edit the xml to add the Num+x definitions and reimport the xml file.
I see that this topic is noted as "fixed", but I recommend to make sure that the root cause reported here above is also fixed.

PR #15821, opened 2023-01-11
Fix #15798: Use a collapsible panel for additional text settings

### LINKED ISSUE #15798: [MU4 Task] To make 'More' drop-downs in Properties consistent
**Task description**
There're different implementation for 'More' drop-downs in Properties. E.g. in Note section it's implemented as a cllapsable section called "Show more" and for Text section as a "More.." button with dialog style drop-down. Will be great to make these cases consistent and choose only one variant of implementation.

@Tantacrul @bkunda please, confirm

abariska 2023-01-10 at 12 38 35

iwoithe:
Resolves: #15798 - specifically the following

> the only real change is to replace the 'More' button in the ‘Text‘ properties panel to a ‘Show more‘ (collapsible) panel

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/text/TextInspectorView.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/text/TextInspectorView.qml
@@ -322,31 +322,23 @@ InspectorSectionView {
             }
         }
 
-        PopupViewButton {
-            id: textAdvancedSettingsButton
+        ExpandableBlank {
+            id: showItem
+            isExpanded: false
 
-            width: contentColumn.width
-            anchorItem: root.anchorItem
+            title: isExpanded ? qsTrc("inpsector", "Show less") : qsTrc("inspector", "Show more")
+
+            visible: root.model ? !root.model.isEmpty : false
+            width: parent.width
 
             navigation.panel: root.navigationPanel
             navigation.name: "TextAdvancedSettings"
             navigation.row: insertSpecCharactersButton.navigation.row + 1
 
-            text: qsTrc("inspector", "More…")
-            visible: root.model ? !root.model.isEmpty : false
-
-            popupContent: TextSettings {
+            contentItemComponent: TextSettings {
                 model: root.model
-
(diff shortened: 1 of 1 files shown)


Tantacrul:
Need to sign the CLA before we'll consider merging anything.

src/inspector/view/qml/MuseScore/Inspector/text/TextInspectorView.qml
```diff
@@ -322,31 +322,23 @@ InspectorSectionView {
...
-            }
-
-            onEnsureContentVisibleRequested: function(invisibleContentHeight) {
-                root.ensureContentVisibleRequested(invisibleContentHeight)
-            }
-
-            onPopupOpened: {
-                root.popupOpened(popup, control)
+                navigationPanel: root.navigationPanel
+                navigationRowStart: showItem.navigation.row
```
cbjeukendrup, line 341: To be on the safe side, this should be <code>showItem.navigation.row <b><i>+ 1</i></b></code>. I know that `TextSettings` has its first control at `navigationRowStart + 1`, so there is no real problem yet, but in general, `navigationRowStart` is supposed to be the first index that is safe to be used inside the section, so that should be `showItem.navigation.row + 1`.

cbjeukendrup (commented):
Except for one minor comment, this looks good to me, and it seems to work well too! @bkunda Could you please also test it and check if you're happy from a design POV?

Tantacrul:
I definitely think this is an improvement. 

One thing that would be great to do (although probably in a separate PR) would be to remember whether any given panel is open or collapsed.

For example, if I collapse 'Text', it remains collapsed the next time I select an element. If I open 'Show more', it should also stay open too.

Might be another worthy Community project?

Tantacrul:
Thanks @iwoithe

iwoithe:
@cbjeukendrup Done. :slightly_smiling_face:

cbjeukendrup (approved):
Reviewed and tested once more. Looks good, thanks!

PR #15822, opened 2023-01-11
Introducing text settings for tempo change

mike-spa:
At the moment, "Gradual tempo change" items use the style settings of the "Tempo" text, but they really should have their own, like all other items. Otherwise, it's impossible to customize the appearance of any "Gradual tempo change" item without modifying the global style settings of the "Tempo". Of course, the default settings are still the same as the "Tempo" settings. This PR just creates the dedicates settings and adds the entry to the list.

image


oktophonie (approved):

XiaoMigros:
Wouldn't a name like "Tempo expression" better describe this setting? I feel like "Gradual tempo change" is not encompassing enough..

oktophonie:
But a 'gradual tempo change' is what they are.
Also the name has not arisen from this PR - that's what they are called already in the UI, as you can see if you select one:
image

XiaoMigros:
I see. But my point still stands; "a tempo" for example should, in my opinion, be stylised the same as "rit." and similar, and it is not usually a gradual change.

DaddyLudwig:
I've seen a tempo marked both ways, in the style of a tempo marking like Allegro or Presto and in the style of a gradual change like rit. or accel. so it doesn't really matter to me.

XiaoMigros:
The only times I've seen 'a tempo' marked like 'Allegro', 'rit.' and similar were also marked like 'Allegro'.
It's a tiny detail that doesn't really matter, but if it was up to me I would opt for a more inclusive/comprehensive category name. Do with that what you will!

RomanPudashkin (approved):

PR #15825, opened 2023-01-11
gp-import: fixed octava import when gp file is bugged

alexpavlov96:
fixed behaviour when guitar pro file contains random octava signs on rests (bug of guitar pro)


alexpavlov96:
@abariska it wasn't regression, but good catch! Will fix in the same PR

handrok (approved):

abariska (approved):

PR #15826, opened 2023-01-11
Fix #15473: No platform menu bar support on Linux

### LINKED ISSUE #15473: [MU4 Issue] Menu bar Doesn't behave like a normal menu bar
**Describe the bug**
A short and clear description of what the bug is:
The Menubar is always visible in the Application menu. Even if Global menu is enabled in a Panel
It looks like this:
image

Expected behavior (another application):
image

**To Reproduce**
- i use KDE Plasma 5.26.4 on KDE Neon.
- installed the application via the AppImmage provided on the Musescore website
- added global menu to the plasma-panel
- every app has their menu now in the global menu and not in the application window, exept musescore

adazem009:
Resolves: #15473 

Native menu bar support is only implemented for macOS. This pull request contains changes that will enable it on Linux-based systems as well (for example the global menu bar in KDE Plasma).
Uses D-Bus (com.canonical.AppMenu.Registrar) to check whether global menu is available.

I've tested this in KDE Plasma and it properly detects whether global menu is enabled.

Global menu bar

### DIFF
--- a/src/appshell/qml/platform/linux/Main.qml
+++ b/src/appshell/qml/platform/linux/Main.qml
@@ -30,17 +30,28 @@ import "../../"
 AppWindow {
     id: root
 
-    AppMenuBar {
+    Loader {
         id: appMenuBar
 
         anchors.top: parent.top
         anchors.left: parent.left
         anchors.right: parent.right
+    }
 
-        appWindow: root
+    Loader {
+        id: platformMenuBar
     }
 
     Component.onCompleted: {
+        platformMenuBar.setSource("../PlatformMenuBar.qml", { "loadSubitems": false });
+        if (platformMenuBar.item.available) {
+            platformMenuBar.item.load();
+            appMenuBar.active = 0;
+        } else {
+            appMenuBar.setSource("../AppMenuBar.qml", { "appWindow": root });
+            platformMenuBar.active = 0;
+        }
+
         window.init()
     }
 
--- a/src/appshell/view/appmenumodel.cpp
+++ b/src/appshell/view/appmenumodel.cpp
@@ -72,6 +72,11 @@ void AppMenuModel::load()
     appMenuModelHook()->onAppMenuInited();
 }
 
+bool AppMenuModel::isGlobalMenuAvailable()
+{
+    return uiConfiguration()->isGlobalMenuAvailable();
+}
+
 void AppMenuModel::setupConnections()
 {
     recentProjectsProvider()->recentProjectListChanged().onNotify(this, [this]() {
--- a/build/cmake/FindQt5.cmake
+++ b/build/cmake/FindQt5.cmake
@@ -41,6 +41,13 @@ if (WIN32)
       )
 endif(WIN32)
 
+if (OS_IS_LIN)
+    set(_components
+      ${_components}
+      DBus
+      )
+endif()
+
 find_package(Qt5Core ${QT_MIN_VERSION} REQUIRED)
 
 foreach(_component ${_components})
--- a/src/appshell/appshell.qrc
+++ b/src/appshell/appshell.qrc
@@ -105,6 +105,7 @@
         <file>qml/platform/win/AppSystemButtons.qml</file>
         <file>qml/platform/AppMenuBar.qml</file>
         <file>qml/platform/AppButtonBackground.qml</file>
+        <file>qml/platform/PlatformMenuBar.qml</file>
         <file>resources/LoadingScreen.svg</file>
     </qresource>
 </RCC>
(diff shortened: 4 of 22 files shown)


src/appshell/qml/platform/linux/Main.qml
```diff
@@ -40,14 +43,112 @@ AppWindow {
         appWindow: root
     }
 
+     PLATFORM.MenuBar {
```
cbjeukendrup, line 46: I think it would be preferable to extract this into a separate component, and simply instantiate this component on both macOS and Linux. 

Also, on Linux, you could wrap it in a `Loader`, and only make this `Loader` active when the native menu bar is indeed necessary. This prevents that unnecessary components are instantiated and unnecessary property bindings and signal handlers are evaluated.

src/appshell/view/appmenumodel.cpp
```diff
@@ -72,6 +74,15 @@ void AppMenuModel::load()
     appMenuModelHook()->onAppMenuInited();
 }
 
+bool AppMenuModel::nativeMenuBarAvailable()
+{
+    // When any QMenuBar is created, it sets the nativeMenuBar property to true
+    // if it can be used as the native menu bar.
+    // This dummy menu bar is created just to check native menu bar availability.
+    QMenuBar dummyMenuBar;
+    return dummyMenuBar.isNativeMenuBar();
```
cbjeukendrup, line 83: I'm not a big fan of creating such a dummy object; in the case of QMenuBar that doesn't look like a particularly trivial operation, since the underlying menu bar is immediately created, only to be destroyed a few nanoseconds later, when the method returns. 

It would be preferable to use just the code that detects whether the native menu bar will be created. I took a quick look at the Qt source code, and it seems to boil down to this function: https://codebrowser.dev/qt5/qtbase/src/platformsupport/themes/genericunix/qgenericunixthemes.cpp.html#118

cbjeukendrup (commented):
Thanks for you contribution! A few comments.

src/appshell/view/appmenumodel.cpp
adazem009: Ok, thanks, I should have looked into the source code...
By the way, is it okay to add this function to AppMenuModel? I think it's the most appropriate class for it.

cbjeukendrup: Yes, that seems fine to me!

adazem009: To use D-Bus, MuseScore has to link with the Qt D-Bus module. I'm not sure if it's enough to just add it to `build/cmake/FindQt5.cmake`, though.

src/appshell/view/appmenumodel.cpp
```diff
@@ -21,6 +21,9 @@
  */
 #include "appmenumodel.h"
 
+#include <QtDBus/QDBusConnection>
+#include <QtDBus/QDBusConnectionInterface>
```
cbjeukendrup, line 25: I think these would also need to be wrapped in `#ifdef Q_OS_LINUX`

adazem009: Yeah, I've forgotten about those, I'll fix that...

src/appshell/view/appmenumodel.cpp
```diff
@@ -72,6 +75,30 @@ void AppMenuModel::load()
...
 
+bool AppMenuModel::isGlobalMenuAvailable()
+{
+#if defined(Q_OS_MACOS)
+    return true;
+#elif defined(Q_OS_LINUX)
+    return isDBusGlobalMenuAvailable();
+#else
+    return false;
+#endif
```
igorkorsukov, line 86: I think better move this method to `IPlatformTheme` interface and use it here

adazem009: It would be better to keep the `isGlobalMenuAvailable()` method in `AppMenuModel` (because it's already exposed to QML) and add platform-specific methods to subclasses of `IPlatformTheme`

igorkorsukov: Yes, exactly

igorkorsukov (approved):

igorkorsukov:
@adazem009  Great! 

In order for the unit tests to be successful, could you please update with the master

cbjeukendrup:
The unit test failure is also on the latest master branch, but only sometimes. I still need to clean that up...

adazem009:
> @adazem009 Great!
> 
> In order for the unit tests to be successful, could you please update with the master

Unit tests are successful now.

igorkorsukov:
@cbjeukendrup do you have any other suggestions? or is it ready for merge? :)

adazem009:
> @cbjeukendrup do you have any other suggestions? or is it ready for merge? :)

It's ready for merge. I've already tested it on Windows and macOS just to be sure that it doesn't break anything.

cbjeukendrup:
It's probably okay, but I'll quickly check it later today!

cbjeukendrup (approved):

DmitryArefiev:
@cbjeukendrup I don't have **KDE Plasma**. I have only **Ubuntu** and **Mint** and as I know it's not easy to enable Global menu there..

I've tested this PR on Win10,Mac13 and Linux Mint and didn't found any differences with master. So, if it's safe we can merge it to master.

cbjeukendrup:
Thanks for checking! In that case I'll merge it.

DmitryArefiev:
Great. Thanks!

PR #15831, opened 2023-01-12
Correct cross-beam spacing in narrow situations

mike-spa:
Bit of an edge case, admittedly, but it needs this solution. Can happen when spacing is particularly narrow (to reproduce it quickly, just reduce the stretch of the measure with ctrl+{ )
Before:
image
After:
image


oktophonie (approved):

RomanPudashkin:
@mike-spa please rebase it. I also expect to see some new tests

mike-spa:
Rebased! About new vtests: this PR is part of the preparatory work for the larger #15174. I have added several new vtests to that, which will cover all of these cases and many others

RomanPudashkin (approved):

PR #15849, opened 2023-01-13
Fix #15779: Add rest section to properties panel

### LINKED ISSUE #15779: [MU4 Task] Add ‘Rest‘ section in Properties panel
**Task description**
This task would resolve #15446.

A new ‘Rest‘ section should appear in the Properties panel whenever a rest is selected in the score. 

For now, it will only contain the **Beam type** settings, as currently found in the ‘Note‘/‘Beam‘ section of the Properties panel.

Screenshot 2023-01-09 at 3 19 49 pm

_N.B. The mockup above shows the variant of the **Beam type** settings as described in #9892. If practicable, it could be good to tackle both of these issues together._

When multiple elements – including a rest – are selected in the score, a new ‘Rest‘ button will be shown in the Properties panel. Note that the icon for this button is now available in the latest version of musescoreicon. 

Screenshot 2023-01-09 at 3 12 46 pm

Thanks! 

FYI @oktophonie @Tantacrul @jessjwilliamson

iwoithe:
Resolves: #15779

https://user-images.githubusercontent.com/73422868/212297755-5290db19-3e5c-4bdc-8151-a8b43c759b2d.mp4

---


iwoithe:
@bkunda Ready for testing :slightly_smiling_face:

cbjeukendrup (commented):
Since Chords and Rests have a common base class `ChordRest`, and the beam mode property is in fact a property of this class, it would be great if we had only one beam mode model that can be used for both of them. 

Depending on Bradley's opinion about my comment at https://github.com/musescore/MuseScore/issues/15779#issuecomment-1382840938, this would mean:
- either specify the element type (CHORD or REST) as a parameter of the constructor of the beam settings model
- or create a way to retrieve both Chords and Rests from the ElementRepositoryService.

bkunda:
> Depending on Bradley's opinion about my comment at [#15779 (comment)](https://github.com/musescore/MuseScore/issues/15779#issuecomment-1382840938), this would mean:

@cbjeukendrup let's go with your option 2 at this stage (to which @Tantacrul has already given his assent).

iwoithe:
> Pls let me know when it's ready for testing and I'll give it a look.

@bkunda PR is now ready for testing

bkunda:
I have finally had a chance to give this a test, and I think this PR does achieve what it sets out to: namely, incorporating the ‘rest‘ section in **Properties** with relevant beaming options. 

I think there are other issues that should perhaps be solved separately, and agree with @Tantacrul that a better solution would probably be to have a self-contained ‘Beam‘ category that affects both notes and rests (whatever is selected), and perhaps also contains other useful options to, for example, adjust the length of beams that extend over rests (attempting to do this currently results in very strange behaviour), as well an ‘end beam‘ (or ‘break beam right‘) option. As it stands, the behaviour of beam options doesn't always feel easy to intuit, so there can be a lot of moving back and forwards between the beam options panels for rests and notes in an attempt to achieve the desired result. It would ultimately be better to have these options in one place.

I think @oktophonie's observation about not being able to apply beaming options to rests of a quarter note or longer ([see comment](https://github.com/musescore/MuseScore/issues/15779#issuecomment-1404678024)) reveals a problematic limitation. @iwoithe do let us know if it can be resolved as part of your PR. But I suspect it is perhaps better to raise this as a separate issue. 

So with all this in mind, I'd be happy to pass this PR (pending of course a code review @cbjeukendrup) as it does offer some necessary functionality, but I'd then create both an issue to fix the quarter note rest limitation, and a mini-spec to establish a new ‘Beam‘ category that could eventually replace this in a later release.

iwoithe:
> I'd then create both an issue to fix the quarter note rest limitation, and a mini-spec to establish a new ‘Beam‘ category that could eventually replace this in a later release.

Yeah, raise as separate issues.

Not entirely sure what's causing the quarter note rest limitations. Happy to work on the `Beam` category when further details are given. :slightly_smiling_face:

oktophonie:
I raised the beam-over-quarter issue separately: #16798
@iwoithe - if you'd like to work on it, just leave a comment there

RomanPudashkin (approved):

Issue #12222, opened 2022-06-30
[MU4 Issue] Insert measures dialog doesn't support Enter, buttons in wrong order for Windows


wizofaus:
**Describe the bug**
Insert measures dialog has "cancel" and "ok" buttons in the wrong order, requires pressing ESC twice to cancel, and enter doesn't work to activate "OK".

**To Reproduce**
1. Create/open any score
2. Use context menu on any measure and choose "insert measures"
3. Observe order of buttons - OK is on the right.
4. Press ESC. Nothing happens.
5. Press Enter. Nothing happens.

**Expected behavior**
Dialog should work as per standard windows dialogs where OK is on left, Cancel is on right, Enter activates the former and Esc activates the latter

**Screenshots**

**Platform information**
Windows 10

**Additional context**
Debug build run from another user's PR branch recently branched off master, confirmed on reasonably recent nightly build.
The Esc key does work if you press it twice strangely.

HemantAntony:
Hopefully, #11646 should solve the order problem

cbjeukendrup:
Some extra information from #14155 (duplicate):

> I need to push "Tab" then push "Enter" (Other Bug, The "Enter" of numpad dont work)

wizofaus:
BTW the OK/Cancel being switched constantly throws me in MuseScore, it's in at least a couple of other places too.

Jojo-Schmitz:
Just came up again in https://musescore.org/en/node/343772

fenugrec:
Just confirming this (Enter not closing the dialog) also happens on Linux.

Jojo-Schmitz:
Just came up again in https://musescore.org/en/node/366310

wizofaus:
Esc works OK now, but not Enter.

sammik:
> Esc works OK now, but not Enter.

I still need to press Esc twice. (Linux)

shoogle:
Esc is the shortcut to stop interacting with the text field, then you must press it a second time to close the dialog. Right now this is by design, but if we switch back to using Tab for navigation instead of arrow keys (#20056) then we can remove this shortcut, enabling Esc to close the dialog directly.

The Enter/Return key should behave as described in https://github.com/musescore/MuseScore/issues/19739#issuecomment-2255680284. It should push whichever button is highlighed in blue. That should be "Cancel" if "Cancel" has focus. If anything else has focus it be "OK".

PR #15880, opened 2023-01-14
Fix 12222: Enter and Esc keys now work with insert measures dialog

Salavin:
Resolves: #12222

- Added keypress listener for when enter/return/esc are pressed while input is focused and called the respective functions for each key.
- Added `okClick()` function to reduce code copying, and added `cancelClick()` to pair with it.

### DIFF
--- a/src/notation/qml/MuseScore/NotationScene/SelectMeasuresCountDialog.qml
+++ b/src/notation/qml/MuseScore/NotationScene/SelectMeasuresCountDialog.qml
@@ -36,6 +36,15 @@ StyledDialogView {
 
     property int measuresCount: 1
 
+    function okClick() {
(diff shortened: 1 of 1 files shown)


Salavin:
@cbjeukendrup could I possibly get a review of this?

cbjeukendrup (commented):
Although this is a nice solution for this particular case, I'm wondering whether we should look for something more general that works for other dialogs too. @Eism what do you think?

Eism:
> I'm wondering whether we should look for something more general that works for other dialogs too. @Eism what do you think?

Agree with you

RomanPudashkin:
Closing due to a long period of inactivity. Please reopen when you implement a more general solution. Thanks!

Issue #15769, opened 2023-01-08
[MU4 Issue] Crash when dropping staff type change on multi-measure rest measure


cbjeukendrup:
**To Reproduce**
1. Create score
2. Enable multi-measure rests by pressing M on the keyboard
3. From the Layout palette, drop the Staff Type Change element on the measure containing the multi-measure rest symbol (not on the symbol itself because that does nothing).
-> Crash

DmitryArefiev:
Fixed in #15891

PR #15891, opened 2023-01-15
Fix crash when applying staff type change to MMRest measure

cbjeukendrup:
Resolves: #15769 

We should not apply the staff type change to the MMRest measure, but rather to one of the underlying measures. Elements that are applied to the MMRest measure but not to the underlying measures are deleted during layout (that's fine I guess), but we still use them when constructing the ChangesInfo (`mu::engraving::UndoMacro::changesInfo()`; `mu::engraving::compoundObjects(mu::engraving::EngravingObject const*)`).

This fix does introduce a bit of jank: when you add a staff type change to a MMRest measure, you won't see it while MMRests are enabled (only when they are disabled). This can be fixed using this patch: https://github.com/cbjeukendrup/MuseScore/commit/0b5a160c521794edd2b01f21f05198b930cfab30

The reason that I'm not doing that, is that that introduces a crash when you delete the staff type change while MMRests are enabled. This crash is complicated to fix, because of how StaffTypeChanges work.

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -67,6 +67,7 @@
 #include "libmscore/staff.h"
 #include "libmscore/stafflines.h"
 #include "libmscore/stafftype.h"
+#include "libmscore/stafftypechange.h"
 #include "libmscore/system.h"
 #include "libmscore/textedit.h"
 #include "libmscore/textframe.h"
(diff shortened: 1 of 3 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -1731,6 +1742,21 @@ bool NotationInteraction::applyPaletteElement(mu::engraving::EngravingItem* elem
                     break;
                 }
             }
+        } else if (element->isStaffTypeChange()) {
+            Measure* measure = sel.startSegment() ? sel.startSegment()->measure() : nullptr;
+            if (measure) {
+                // Partially copied from NotationInteraction::applyDropPaletteElement
```
RomanPudashkin, line 1748: Ideally, we should not copy this code

RomanPudashkin (approved):

DmitryArefiev:
Tested #15769 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #14296, opened 2022-11-03
[MU4 Issue] Several problems with system-line objects


mike-spa:
**Describe the bug(s)**
We have _serious_ problems handling system-line objects when they are created from a staff that isn't the first of the score. Specifically, when the system line is created on a note, but the first staff doesn't contain any chordRest in the same segment, the result is basically random. All of these issue affect System-text-line items but also, crucially, tempo change lines. All of these problems have a common cause in the well known, general problem that lines can't be attached to rythmic positions but only to chord/rests. However, they are also separate problems with specific separate causes.

1. This is the most serious one: if the start point of the system line is mid-way through a measure, the line will not be saved, and thus goes lost upon closing-reopening the score.

https://user-images.githubusercontent.com/93707756/199779860-d2af3413-46ff-4d98-ad12-a835b61e3e60.mp4

2. If the start point is mid way of a measure, layout goes completely out of place. But only after a certain measure. 

https://user-images.githubusercontent.com/93707756/199776596-93ee4430-9c85-4d05-908d-62d80e0ced74.mp4

3. Layout goes completely out of place, but oppositely, if mid-way of the first measure.

https://user-images.githubusercontent.com/93707756/199777010-0d7fa739-9d58-4447-9f99-11eb0b230f9a.mp4

4. If entered with range selection (shift+click) the line will go above the top staff as expected. But if entered with list selection (ctrl + click), the line goes above the selected staff. 

https://user-images.githubusercontent.com/93707756/199778513-31f6a59e-19aa-47e2-8f4f-013f494cb5d5.mp4

5. If the end point of the selection is mid-way through a measure, it will extend until the end of the measure, and not until the selected point.

https://user-images.githubusercontent.com/93707756/199780463-b26ae766-a5cf-45df-b2a7-12878e7ccba0.mp4

6. The issue repoerted in #14212, namely that the line disappear if it has been put on a rest and then a note takes its place. This is the less serious of all the issue IMO, but still annoying.

https://user-images.githubusercontent.com/93707756/199783494-6413b00a-fa50-4572-bf08-f35ebe938588.mp4

mike-spa:
Oh, and I forgot another very obvious (and very bad) one.

7. Tempo lines (unlike tempo text) will disappear if the top staff is hidden

https://user-images.githubusercontent.com/93707756/199930041-1ce6ddd3-14c0-439c-8f6d-7baa6f780b91.mp4

mike-spa:
[...]

PR #15931, opened 2023-01-17
More tempo line fixes

mike-spa:
Resolves: #14296

Fixing the next bunch of issues with tempo lines.
- Fixed wrong staff on copy-pasting
- Fixed mid-measure tempo indication disappearing in parts
- ~Fixed wrong staff of tempo lines upon reordering instruments~
- ~Fixed lost tempo lines upon save-reopen if they start or end mid-measure~

The layout of tempo lines (actually, all system-level text lines) is still imperfect because (let's all repeat together):

_We cAN't AtTAcH THinGs To RhyTHmIC PoSiTiONS_

Most commonly: things will just snap to the start/end of a measure if there is nothing to attach to within that measure. Can't really do anything about that at the moment.

### DIFF
--- a/src/engraving/layout/layoutmeasure.cpp
+++ b/src/engraving/layout/layoutmeasure.cpp
@@ -459,12 +459,26 @@ static bool breakMultiMeasureRest(const LayoutContext& ctx, Measure* m)
         return true;
     }
 
+    // Break for spanners/textLines in this measure
     auto sl = ctx.score()->spannerMap().findOverlapping(m->tick().ticks(), m->endTick().ticks());
     for (auto i : sl) {
         Spanner* s = i.value;
-        // break for first measure of volta or textline and first measure *after* volta
-        if ((s->isVolta() || s->isGradualTempoChange() || s->isTextLine()) && (s->tick() == m->tick() || s->tick2() == m->tick())) {
+        if ((s->isVolta() || s->isGradualTempoChange() || s->isTextLine())
+            && ((s->tick() >= m->tick() && s->tick() < m->endTick()) || (s->tick2() >= m->tick() && s->tick2() < m->endTick()))) {
             return true;
+        }
+    }
+    // Break for spanners/textLines starting or ending mid-way inside the previous measure
+    Measure* prevMeas = m->prevMeasure();
+    if (prevMeas) {
+        auto prevMeasSpanners = ctx.score()->spannerMap().findOverlapping(prevMeas->tick().ticks(), prevMeas->endTick().ticks());
+        for (auto i : prevMeasSpanners) {
+            Spanner* s = i.value;
+            if ((s->isVolta() || s->isGradualTempoChange() || s->isTextLine())
+                && ((s->tick2() > prevMeas->tick() && s->tick2() < prevMeas->endTick())
+                    || (s->tick() > prevMeas->tick() && s->tick() < prevMeas->endTick()))) {
+                return true;
+            }
         }
     }
 
@@ -502,7 +516,6 @@ static bool breakMultiMeasureRest(const LayoutContext& ctx, Measure* m)
     }
 
     // Break for annotations found mid-way into the previous measure
-    Measure* prevMeas = m->prevMeasure();
     if (prevMeas) {
         for (Segment* s = prevMeas->first(); s; s = s->next()) {
             for (EngravingItem* e : s->annotations()) {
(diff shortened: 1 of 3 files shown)


oktophonie (approved):

cbjeukendrup (commented):
The code looks good to me, although I must admit that I don't know this part of the code well enough to understand all the details behind every change.

mike-spa:
UPDATE
I have removed a couple of commits from this PR.
1. The PR does _not_ try anymore to solve the last issue mentioned [here](https://github.com/musescore/MuseScore/pull/15931#issue-1536188563) (see Casper's comment [here](https://github.com/musescore/MuseScore/issues/14296#issuecomment-1381839350) for a more detailed description of the issue itself). The root cause of the issue is that, in this scenario, the line ends up with no valid start- or end-segment, so it will also not be written in the file. My previous solution tried to fix the problem by changing how the writing of spanners works, but a) It's too dangerous, b) Because the writing of spanners changed, it was causing dozens and dozens of utest failures, c) It is really not the solution we ultimately want. What we want is proper attachment to rythmic position, so that the line does _not_ end up with invalid start- or end anchors. And, as we know, this is part of a larger job. So we'll just have to live with that issue until we can fix it properly.
2. The second-to-last issue mentioned [here](https://github.com/musescore/MuseScore/pull/15931#issue-1536188563) was meanwhile solved by [another more complete PR](https://github.com/musescore/MuseScore/pull/15938) so it wasn't necessary anymore. 

@oktophonie please do take a look if you can, and check that everything else works as intended. @RomanPudashkin the tests are now fixed so I think it's ready for review

oktophonie (approved):

src/engraving/layout/layoutmeasure.cpp
```diff
@@ -501,7 +515,28 @@ static bool breakMultiMeasureRest(const LayoutContext& ctx, Measure* m)
...
+    // Break for annotations found mid-way into the previous measure
+    if (prevMeas) {
+        for (Segment* s = prevMeas->first(); s; s = s->next()) {
+            for (EngravingItem* e : s->annotations()) {
+                if (!e->visible()) {
+                    continue;
+                }
+                if ((e->isRehearsalMark()
+                     || e->isTempoText()
+                     || ((e->isHarmony() || e->isStaffText() || e->isSystemText() || e->isTripletFeel() || e->isPlayTechAnnotation()
```
RomanPudashkin, line 527: This block of code is impossible to read/debug. I think you could use static std::set/unordered_set here instead of adding a lot of checks to this condition

src/engraving/layout/layoutmeasure.cpp
```diff
@@ -459,14 +459,28 @@ static bool breakMultiMeasureRest(const LayoutContext& ctx, Measure* m)
...
         }
     }
+    // Break for spanners/textLines starting or ending mid-way inside the previous measure
+    Measure* prevMeas = m->prevMeasure();
+    if (prevMeas) {
+        auto prevMeasSpanners = ctx.score()->spannerMap().findOverlapping(prevMeas->tick().ticks(), prevMeas->endTick().ticks());
+        for (auto i : prevMeasSpanners) {
+            Spanner* s = i.value;
+            if ((s->isVolta() || s->isGradualTempoChange() || s->isTextLine())
+                && ((s->tick2() > prevMeas->tick() && s->tick2() < prevMeas->endTick())
```
RomanPudashkin, line 478: Same as https://github.com/musescore/MuseScore/pull/15931/files#r1115733878

+ a lot of redundant calls (for example, prevMeas->tick() is called twice here). This also creates performance problems

src/engraving/libmscore/excerpt.cpp
```diff
@@ -1420,6 +1420,21 @@ void Excerpt::cloneStaff2(Staff* srcStaff, Staff* dstStaff, const Fraction& star
                     }
                 }
 
+                for (EngravingItem* e : oseg->annotations()) {
+                    if (e->generated()
+                        || (e->track() != srcTrack && !(e->systemFlag() && e->track() == 0)) // system items must be cloned even if they are on different tracks
+                        || (e->track() != srcTrack && e->systemFlag() && e->findLinkedInScore(score))) { // ...but only once!
```
RomanPudashkin, line 1426: This condition has become even more complicated than it was before

RomanPudashkin (approved):

PR #15935, opened 2023-01-17
Added another case in which the score is considered corrupted

Eism:
There have been a bunch of issues as a result of which we have corrupted scores. Added one more case by which we can calculate corrupted scores and the user can fix them.


vpereverzev (approved):

Jojo-Schmitz:
rebase needed

Issue #15835, opened 2023-01-12
[MU4 Issue] Some duplicated system objects (Tempo, Rehearsal mark) are change position after deleting instrument and don't return after Undo


DmitryArefiev:
https://user-images.githubusercontent.com/90187801/212084590-6202e617-9f43-4c7e-bd62-343f4ae4a363.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
Occurs in **4.0.0** as well

DmitryArefiev:
Fixed in #15938

DmitryArefiev:
Retested on Win10

Issue #15735, opened 2023-01-06
[MU4 Issue] Crash when deleting staff containing system objects and all staves below in existing score with system objects and "Segno" marking


cbjeukendrup:
**To Reproduce**
1. Create a score from a template with system objects, like SATB+Piano or Symphony Orchestra
2. Add a "Segno" marking in some measure 
3. Delete the staff that shows the system objects and all staves below if any (so in the case of SATB+Piano, the Piano, and in the case of Symphony Orchestra, the whole strings section)
-> instant crash

**Additional context**
Originally reported in https://musescore.org/en/node/341484
I can't reproduce it with a Tempo marking rather than a Segno marker; for the Tempo, instead of a crash, the marking would be duplicated on the top staff.
Not yet fixed by https://github.com/musescore/MuseScore/pull/15715

DmitryArefiev:
Retested on Win10

PR #15938, opened 2023-01-17
Fix system objects issues

mike-spa:
Resolves: #15835
Resolves: #15735 
Resolves: #16063 

In my previous PR #15806 I had made a zero-risk immediate fix to solve the crash / file corruption issue. That has gone into the first patch release. Here I'm trying to do a more complete solution, which should also work correctly when doing ctrl+z and deleting staves.

### DIFF
--- a/src/engraving/libmscore/engravingitem.cpp
+++ b/src/engraving/libmscore/engravingitem.cpp
@@ -451,7 +451,7 @@ staff_idx_t EngravingItem::staffIdxOrNextVisible() const
     }
     bool foundStaff = false;
     if (!m->system()->staff(si)->show()) {
-        std::vector<Staff*> soStaves = score()->getSystemObjectStaves();
+        const std::vector<Staff*> soStaves = score()->systemObjectStaves();
         for (staff_idx_t i = 0; i < soStaves.size(); ++i) {
             staff_idx_t idxOrig = soStaves[i]->idx();
             if (idxOrig == si) {
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -2631,7 +2631,7 @@ void Score::insertStaff(Staff* staff, staff_idx_t ridx)
 
     for (auto i = staff->score()->spanner().cbegin(); i != staff->score()->spanner().cend(); ++i) {
         Spanner* s = i->second;
-        if (s->systemFlag()) {
+        if (!moveDownWhenAddingStaves(s, idx)) {
             continue;
         }
         if (s->staffIdx() >= idx) {
@@ -2737,6 +2737,10 @@ void Score::removeStaff(Staff* staff)
 
     mu::remove(_staves, staff);
     staff->part()->removeStaff(staff);
+
+    if (isSystemObjectStaff(staff)) {
+        mu::remove(m_systemObjectStaves, staff);
+    }
 
     updateStavesNumberForSystems();
 }
@@ -2905,7 +2909,7 @@ void Score::cmdRemoveStaff(staff_idx_t staffIdx)
 void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
 {
     std::vector<staff_idx_t> moveTo;
-    for (Staff* staff : systemObjectStaves) {
+    for (Staff* staff : m_systemObjectStaves) {
         moveTo.push_back(staff->idx());
     }
     // rebuild system object staves
@@ -2913,7 +2917,7 @@ void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
         staff_idx_t newLocation = mu::indexOf(dst, i);
         if (newLocation == mu::nidx) { //!dst.contains(_staves[i]->idx())) {
             // this staff was removed
-            for (size_t j = 0; j < systemObjectStaves.size(); j++) {
+            for (size_t j = 0; j < m_systemObjectStaves.size(); j++) {
                 if (_staves[i]->idx() == moveTo[j]) {
                     // the removed staff was a system object staff
                     if (i == _staves.size() - 1 || mu::contains(moveTo, _staves[i + 1]->idx())) {
@@ -2924,7 +2928,7 @@ void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
                     }
                 }
             }
-        } else if (newLocation != _staves[i]->idx() && mu::contains(systemObjectStaves, _staves[i])) {
+        } else if (newLocation != _staves[i]->idx() && mu::contains(m_systemObjectStaves, _staves[i])) {
             // system object staff was moved somewhere, put the system objects at the top of its new group
             staff_idx_t topOfGroup = newLocation;
             String family = _staves[dst[newLocation]]->part()->familyId();
@@ -2936,11 +2940,11 @@ void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
                     topOfGroup--;
                 }
             }
-            moveTo[mu::indexOf(systemObjectStaves, _staves[i])] = dst[topOfGroup];
+            moveTo[mu::indexOf(m_systemObjectStaves, _staves[i])] = dst[topOfGroup];
         }
     }
-    for (staff_idx_t i = 0; i < systemObjectStaves.size(); i++) {
-        if (moveTo[i] == systemObjectStaves[i]->idx()) {
+    for (staff_idx_t i = 0; i < m_systemObjectStaves.size(); i++) {
+        if (moveTo[i] == m_systemObjectStaves[i]->idx()) {
             // this sysobj staff doesn't move
             continue;
         } else {
@@ -2951,7 +2955,7 @@ void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
                 }
                 Measure* m = toMeasure(mb);
                 for (EngravingItem* e : m->el()) {
-                    if ((e->isJump() || e->isMarker()) && e->isLinked() && e->track() == staff2track(systemObjectStaves[i]->idx())) {
+                    if ((e->isJump() || e->isMarker()) && e->isLinked() && e->track() == staff2track(m_systemObjectStaves[i]->idx())) {
                         if (moveTo[i] == mu::nidx) {
                             // delete this clone
                             m->remove(e);
@@ -2971,7 +2975,7 @@ void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
                                 || e->isTripletFeel()
                                 || e->isTempoText()
                                 || isSystemTextLine(e)) {
-                                if (e->track() == staff2track(systemObjectStaves[i]->idx()) && e->isLinked()) {
+                                if (e->track() == staff2track(m_systemObjectStaves[i]->idx()) && e->isLinked()) {
                                     if (moveTo[i] == mu::nidx) {
                                         s->removeAnnotation(e);
                                         e->unlink();
@@ -2987,10 +2991,10 @@ void Score::sortSystemObjects(std::vector<staff_idx_t>& dst)
             }
             // update systemObjectStaves with the correct staff
             if (moveTo[i] == mu::nidx) {
-                systemObjectStaves.erase(systemObjectStaves.begin() + i);
+                m_systemObjectStaves.erase(m_systemObjectStaves.begin() + i);
                 moveTo.erase(moveTo.begin() + i);
             } else {
-                systemObjectStaves[i] = _staves[moveTo[i]];
+                m_systemObjectStaves[i] = _staves[moveTo[i]];
             }
         }
     }
@@ -3036,7 +3040,7 @@ void Score::sortStaves(std::vector<staff_idx_t>& dst)
         voice_idx_t voice    = sp->voice();
         staff_idx_t staffIdx = sp->staffIdx();
         staff_idx_t idx = mu::indexOf(dst, staffIdx);
-        if (idx != mu::nidx) {
+        if (idx != mu::nidx && !sp->isTopSystemObject()) {
             sp->setTrack(idx * VOICES + voice);
             if (sp->track2() != mu::nidx) {
                 sp->setTrack2(idx * VOICES + (sp->track2() % VOICES));        // at least keep the voice...
@@ -5596,6 +5600,21 @@ Part* Score::partById(const ID& partId) const
     }
 
     return nullptr;
+}
+
+void Score::clearSystemObjectStaves()
+{
+    m_systemObjectStaves.clear();
+}
+
+void Score::addSystemObjectStaff(Staff* staff)
+{
+    m_systemObjectStaves.push_back(staff);
+}
+
+bool Score::isSystemObjectStaff(Staff* staff) const
+{
+    return mu::contains(m_systemObjectStaves, staff);
 }
 
 int Score::visiblePartCount() const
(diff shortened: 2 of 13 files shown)


cbjeukendrup:
This seems to resolve #15735 🥳

DmitryArefiev:
Tested on **Win10** - found a **crash** when parts are opened


https://user-images.githubusercontent.com/90187801/213509197-04e91ec1-3625-46bd-9e76-1c04f5dab89f.mp4

mike-spa:
@DmitryArefiev great catch, thanks! Should be fixed now

src/engraving/libmscore/score.cpp
```diff
@@ -2648,6 +2648,15 @@ void Score::insertStaff(Staff* staff, staff_idx_t ridx)
         }
     }
 
+    if (staff->isSystemObjectStaff()) {
+        bool isAlreadyInSystemObjectList = std::find_if(systemObjectStaves.begin(), systemObjectStaves.end(),
+                                                        [staff](Staff* stf){ return stf == staff; })
+                                           != systemObjectStaves.end();
```
cbjeukendrup, line 2654: Couldn't you use `mu::contains(systemObjectStaves, staff)` here?

src/engraving/libmscore/score.cpp
```diff
@@ -2736,6 +2745,10 @@ void Score::removeStaff(Staff* staff)
     mu::remove(_staves, staff);
     staff->part()->removeStaff(staff);
 
+    if (staff->isSystemObjectStaff()) {
+        mu::remove_if(systemObjectStaves, [staff](Staff* stf){ return stf == staff; });
```
cbjeukendrup, line 2749: and `mu::remove(systemObjectStaves, staff)` here?

src/engraving/libmscore/score.cpp
mike-spa: Thanks! Didn't know we had them

DmitryArefiev:
> @DmitryArefiev great catch, thanks! Should be fixed now

@mike-spa Yeah, it's fixed now! But I found some other issues.. :)

DmitryArefiev:
1. **Crash** when deleting all instruments with Segno in score and then Undo (related to #15735)


https://user-images.githubusercontent.com/90187801/214616551-913604ef-9137-49fc-a59d-4810d8699992.mp4

DmitryArefiev:
2. Segno duplicated when deleting top part of instruments. Tempo duplicated when deleting top part of instruments with opened parts (https://github.com/musescore/MuseScore/issues/15835)


https://user-images.githubusercontent.com/90187801/214617271-c4d9a15b-bf32-469c-9d5e-1a4f74e37fce.mp4

DmitryArefiev:
3. Found an issue with doubled jumps in part. Logged separately because that also happens in master - [#16063](https://github.com/musescore/MuseScore/issues/16063)

mike-spa:
@DmitryArefiev thanks so much for your help. The last commit I've pushed should (_I hope to God_) fix the original issues, plus all the subsequent ones you've found. Let me know!

DmitryArefiev:
> @DmitryArefiev thanks so much for your help. The last commit I've pushed should (_I hope to God_) fix the original issues, plus all the subsequent ones you've found. Let me know!

@mike-spa Yes! All fixed now! 

..I only found another crash with score with different system objects but can't reproduce it on a new score, and it also happens in master. Logged separately #16174

DmitryArefiev:
Tested #15835 #15735 #16063 on Win10, Mac12 - **FIXED**

src/engraving/libmscore/staff.h
```diff
@@ -98,6 +98,7 @@ class Staff final : public EngravingItem
     bool _hideSystemBarLine  { false };         // no system barline if not preceded by staff with barline
     bool _mergeMatchingRests { false };         // merge matching rests in multiple voices
     HideMode _hideWhenEmpty  { HideMode::AUTO };      // hide empty staves
+    bool _isSystemObjectStaff { false };    ///< must be kept in sync with Score::systemObjectStaves
```
RomanPudashkin, line 101: Do we really need this flag here? I think we can add a method to Score that will give us the same information:

`bool Score::isSystemObjectStaff(const Staff* staff) const;
`

This way, we avoid the need for synchronization between different objects

src/engraving/libmscore/score.cpp
```diff
@@ -3034,7 +3043,7 @@ void Score::sortStaves(std::vector<staff_idx_t>& dst)
         voice_idx_t voice    = sp->voice();
         staff_idx_t staffIdx = sp->staffIdx();
         staff_idx_t idx = mu::indexOf(dst, staffIdx);
-        if (idx != mu::nidx) {
+        if (idx != mu::nidx && !(sp->systemFlag() && staffIdx == 0)) {
```
RomanPudashkin, line 3046: What problem does this condition solve?

src/engraving/libmscore/segment.cpp
```diff
@@ -508,7 +508,9 @@ void Segment::insertStaff(staff_idx_t staff)
 
     for (EngravingItem* e : _annotations) {
         staff_idx_t staffIdx = e->staffIdx();
-        if (staffIdx >= staff && !e->isTopSystemObject()) {
+        bool keepSystemObjectOnTop = e->systemFlag() && staffIdx == 0
+                                     && !(staff + 1 < score()->nstaves() && score()->staff(staff + 1)->isSystemObjectStaff());
```
RomanPudashkin, line 512: I have already seen similar conditions in this PR several times. All of them should be simplified, comments should be added to them. It is very difficult to read/debug them in their current state. In addition, they can potentially be moved to a function if they are duplicates

src/engraving/libmscore/staff.h
mike-spa: Yeah, the reason I introduced that flag is because I didn't know how to retrieve the property when the staff gets deleted and then re-inserted by ctrl+Z. But that's a very clumsy solution, I agree. I now realize that the right way to do it is probably to just store that info in the undo stack, as @cbjeukendrup suggested

src/engraving/libmscore/segment.cpp
mike-spa: Yeah, all those conditions are totally horrible. Basically, they try to account for all the different combinations that you can delete/reinsert staves and how to deal with the systemObjects in each case (sometime the object should be deleted with the staff, sometimes not, sometimes it must be re-inserted with the staff, sometimes not, etc etc...). And each of those cases is slightly different, so it's hard to extract this logic into a re-usable piece. I'll do my best to tidy it up and make it as clear as possible

mike-spa:
@RomanPudashkin I've done my best to tidy things up into a more sensible form. Let me know 👍

src/engraving/libmscore/score.h
```diff
@@ -638,9 +638,10 @@ class Score : public EngravingObject
     Staff* staffById(const ID& staffId) const;
     Part* partById(const ID& partId) const;
 
-    void clearSystemObjectStaves() { systemObjectStaves.clear(); }
-    void addSystemObjectStaff(Staff* staff) { systemObjectStaves.push_back(staff); }
-    std::vector<Staff*> getSystemObjectStaves() { return systemObjectStaves; }
+    void clearSystemObjectStaves() { m_systemObjectStaves.clear(); }
+    void addSystemObjectStaff(Staff* staff) { m_systemObjectStaves.push_back(staff); }
+    std::vector<Staff*> getSystemObjectStaves() { return m_systemObjectStaves; }
```
RomanPudashkin, line 643: I would suggest removing this method

src/engraving/libmscore/score.cpp
```diff
@@ -2738,6 +2742,10 @@ void Score::removeStaff(Staff* staff)
     mu::remove(_staves, staff);
     staff->part()->removeStaff(staff);
 
+    if (staff->isSystemObjectStaff()) {
```
RomanPudashkin, line 2745: isSystemObjectStaff() should be declared inside Score. This code looks strange now

src/engraving/libmscore/score.h
```diff
@@ -638,9 +638,10 @@ class Score : public EngravingObject
     Staff* staffById(const ID& staffId) const;
     Part* partById(const ID& partId) const;
 
-    void clearSystemObjectStaves() { systemObjectStaves.clear(); }
-    void addSystemObjectStaff(Staff* staff) { systemObjectStaves.push_back(staff); }
-    std::vector<Staff*> getSystemObjectStaves() { return systemObjectStaves; }
+    void clearSystemObjectStaves() { m_systemObjectStaves.clear(); }
```
RomanPudashkin, line 641: These methods should be implemented inside the *.cpp file, not here

src/engraving/libmscore/score.cpp
```diff
@@ -2617,7 +2617,7 @@ void Score::removePart(Part* part)
 //   insertStaff
 //---------------------------------------------------------
 
-void Score::insertStaff(Staff* staff, staff_idx_t ridx)
+void Score::insertStaff(Staff* staff, staff_idx_t ridx, bool systemObjectStaff)
```
RomanPudashkin, line 2620: What about Score::appendStaff(Staff* staff)? Either this flag should not be here (and we will call score->addSystemObjectStaff explicitly), or you should also add it to Score::addStaff

src/engraving/libmscore/engravingitem.h
```diff
@@ -563,6 +563,9 @@ class EngravingItem : public EngravingObject
     std::vector<Spanner*>& startingSpanners() { return _startingSpanners; }
     std::vector<Spanner*>& endingSpanners() { return _endingSpanners; }
 
+    bool allowRemoveWhenRemovingStaves(staff_idx_t startStaff, staff_idx_t endStaff = 0) const;
```
RomanPudashkin, line 566: These functions do not feel as if they belong to this class. They are more like global functions somewhere in utils

src/engraving/libmscore/engravingitem.h
```diff
@@ -563,6 +563,9 @@ class EngravingItem : public EngravingObject
...
+    bool moveDownWhenAddingStaves(staff_idx_t startStaff, staff_idx_t endStaff = 0) const;
```
RomanPudashkin, line 567: It's not quite clear to me what happens if startStaff == 0 and endStaff == 0. I guess the default value for endStaff should be mu::nidx, right?

src/engraving/libmscore/engravingitem.cpp
```diff
@@ -2371,6 +2371,49 @@ std::pair<int, float> EngravingItem::barbeat() const
...
+    if (endStaff == 0) { // Default initialized
+        endStaff = startStaff + 1;
+    }
+    if (score()->nstaves() == 1) {
+        return true;
+    }
+    if (staffIdx() < startStaff || staffIdx() >= endStaff) {
+        return false;
+    }
+    Staff* nextRemaining = score()->staff(endStaff);
```
RomanPudashkin, line 2389: It should be const

src/engraving/libmscore/engravingitem.cpp
```diff
@@ -2371,6 +2371,49 @@ std::pair<int, float> EngravingItem::barbeat() const
...
+    if (staffIdx() == mu::nidx || startStaff == mu::nidx || endStaff == mu::nidx) {
+        return false;
+    }
+    if (endStaff == 0) { // Default initialized
+        endStaff = startStaff + 1;
+    }
+    if (staffIdx() < startStaff) {
+        return false;
+    }
+    Staff* nextAfterInserted = score()->staff(endStaff);
```
RomanPudashkin, line 2409: It should be const

src/engraving/libmscore/engravingitem.cpp
```diff
@@ -2371,6 +2371,49 @@ std::pair<int, float> EngravingItem::barbeat() const
...
+bool EngravingItem::allowRemoveWhenRemovingStaves(staff_idx_t startStaff, staff_idx_t endStaff) const
+{
+    // Sanity checks
+    if (staffIdx() == mu::nidx || startStaff == mu::nidx || endStaff == mu::nidx) {
+        return false;
+    }
+    if (endStaff == 0) { // Default initialized
+        endStaff = startStaff + 1;
+    }
+    if (score()->nstaves() == 1) {
```
RomanPudashkin, line 2383: This check should be before `if (endStaff == 0) { // Default initialized`

src/engraving/libmscore/engravingitem.cpp
```diff
@@ -2371,6 +2371,49 @@ std::pair<int, float> EngravingItem::barbeat() const
     return std::pair<int, float>(bar + 1, beat + 1 + ticks / static_cast<float>(ticksB));
 }
 
+bool EngravingItem::allowRemoveWhenRemovingStaves(staff_idx_t startStaff, staff_idx_t endStaff) const
+{
+    // Sanity checks
+    if (staffIdx() == mu::nidx || startStaff == mu::nidx || endStaff == mu::nidx) {
+        return false;
+    }
```
RomanPudashkin, line 2379: Add an empty line after each if statement (same for moveDownWhenAddingStaves)

src/engraving/libmscore/undo.h
```diff
@@ -306,6 +306,7 @@ class RemoveStaff : public UndoCommand
 
     Staff* staff = nullptr;
     staff_idx_t ridx = mu::nidx;
+    bool wasSystemObjectStaff = false;
```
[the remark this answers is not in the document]
mike-spa: Not really, cause systemObjectStaff is not a property of the staff, so the staff doesn't "have" that property when it's being inserted

src/engraving/libmscore/engravingitem.h
mike-spa: Actually I need that, cause I wanted to make it possible for the user of this function to just provide the startStaff (if they want to delete only one staff). By setting the default to endStaff = 0, I can distinguish by an invalid input (endStaff == mu::nidx, in which case I exit early) and a default initialization (in which case I just set it to one-past the startStaff). Hope this makes sense

mike-spa:
@RomanPudashkin all done as you suggested (except the two things I pointed out) 👍

src/engraving/libmscore/engravingitem.h
RomanPudashkin: >  wanted to make it possible for the user of this function to just provide the startStaff (if they want to delete only one staff)

My concern is what happens if the user wants to delete **only the first staff** in the score (so startStaff == 0 and endStaff == 0). It seems that if we called these functions as follows (we wanted to check **only the first staff**, not the whole score):

`allowRemoveWhenRemovingStaves(0)
`

We would get this result (not what we expected):

`allowRemoveWhenRemovingStaves(0, score()->nstaves() - 1)
`

Am I right?

mike-spa: No in that case the function reaches [this check](https://github.com/musescore/MuseScore/pull/15938/files#diff-571f2ae508434dfd554cfdbcfb0427062e955fd81490dec44d09259917d7f562R1166-R1168), and if endStaff == 0 we set it to startStaff + 1. So, in the end, calling allowRemoveWhenRemovingStaves(0) is equivalent to calling allowRemoveWhenRemovingStaves(0, 1). Which is what we want, cause endStaff is assumed to be the index one-past the staff we want to remove

RomanPudashkin:
@oktophonie @DmitryArefiev this PR needs to be retested

RomanPudashkin (approved):

DmitryArefiev:
> @oktophonie @DmitryArefiev this PR needs to be retested

Looks good on my side

Tested #15835 #15735 #16063 (and other issues mentioned above) on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #15947, opened 2023-01-18
Fix #342825: MusicXML import: G-Clef 15ma bassa imports as G-Clef

Jojo-Schmitz:
Fixes https://musescore.org/en/node/342825

Seems a day one bug (or ommission), even Mu1 gets that wrong


Jojo-Schmitz:
@lvinken's official blessing would be nice ;-)

RomanPudashkin (approved):

PR #15952, opened 2023-01-18
Implement playback for multi-measure repeats

cbjeukendrup:
Resolves: #10922 

Another attempt, hopefully much better than https://github.com/musescore/MuseScore/pull/11285 :)

The profiler says  about PlaybackModel::updateEvents:
before: between 450 and 467 ms
after: between 467 and 482 ms
for the score "Searching Four" from Muse Hub in a debug+asan build


Jojo-Schmitz:
> before: between 450 and 467 ms
> before: between 467 and 482 ms 

??

cbjeukendrup:
> > before: between 450 and 467 ms
> > before: between 467 and 482 ms
> 
> ??

oops, fixed

Jojo-Schmitz:
No movement on this? Can't we have this in 4.0.2?

cbjeukendrup:
I'm pretty sure we won't have it in 4.0.2, since that is already in final testing state. However, 4.1 would be nice indeed 😉

Jojo-Schmitz:
Tested, works for me

bkunda:
Tested on macOS and Windows. Works for me too!

cbjeukendrup:
Rebased to fix the "conflict" (there was no real conflict but GitHub was just scared about a file that had been moved)

Jojo-Schmitz:
It'd be a real pitty if this PR would miss 4.1.0

src/engraving/libmscore/measurerepeat.cpp
```diff
@@ -63,15 +63,17 @@ Measure* MeasureRepeat::firstMeasureOfGroup() const
     return measure()->firstOfMeasureRepeatGroup(staffIdx());
 }
 
-const Measure* MeasureRepeat::referringMeasure() const
+const Measure* MeasureRepeat::referringMeasure(const Measure* measure) const
```
RomanPudashkin, line 66: IF_ASSERT_FAILED(measure) {
    return nullptr;
}

src/engraving/libmscore/measurerepeat.cpp
```diff
@@ -63,15 +63,17 @@ Measure* MeasureRepeat::firstMeasureOfGroup() const
...
 {
-    Measure* firstMeasureRepeat = firstMeasureOfGroup();
+    Measure* referringMeasure = firstMeasureOfGroup()->prevMeasure();
 
-    if (!firstMeasureRepeat) {
```
RomanPudashkin, line 70: You should check if the result of firstMeasureOfGroup() is null (as it was before)

src/engraving/libmscore/measurerepeat.h
```diff
@@ -43,6 +43,8 @@ class MeasureRepeat final : public Rest
     MeasureRepeat* clone() const override { return new MeasureRepeat(*this); }
     EngravingItem* linkedClone() override { return EngravingItem::linkedClone(); }
 
+    static constexpr int MAX_NUM_MEASURES = 4;
+
     void setNumMeasures(int n) { m_numMeasures = n; }
```
RomanPudashkin, line 48: You should add IF_ASSERT_FAILED to this method to check if n <= MAX_NUM_MEASURES

src/engraving/playback/playbackmodel.cpp
```diff
@@ -369,21 +369,51 @@ void PlaybackModel::processSegment(const int tickPositionOffset, const Segment*
             continue;
         }
 
-        if (item->type() == ElementType::MEASURE_REPEAT) {
-            const MeasureRepeat* measureRepeat = toMeasureRepeat(item);
-            const Measure* currentMeasure = measureRepeat->measure();
-            const Measure* referringMeasure = measureRepeat->referringMeasure();
+        if (isFirstSegmentOfMeasure) {
+            auto processMeasureRepeat
```
RomanPudashkin, line 373: You should probably move this code into a separate private method instead of using the lambda

src/engraving/playback/playbackmodel.cpp
```diff
@@ -369,21 +369,51 @@ void PlaybackModel::processSegment(const int tickPositionOffset, const Segment*
...
+                    return;
+                }
+
+                int currentMeasureTick = currentMeasure->tick().ticks();
+                int referringMeasureTick = referringMeasure->tick().ticks();
+                int repeatPositionTickOffset = currentMeasureTick - referringMeasureTick;
+
+                bool isFirstSegmentOfRepeatedMeasure = true;
+
+                for (Segment* seg = referringMeasure->first(); seg; seg = seg->next()) {
```
RomanPudashkin, line 390: **const** Segment* seg

src/engraving/playback/playbackmodel.h
```diff
@@ -109,7 +109,7 @@ class PlaybackModel : public async::Asyncable
                       ChangedTrackIdSet* trackChanges = nullptr);
 
     void processSegment(const int tickPositionOffset, const Segment* segment, const std::set<staff_idx_t>& staffIdxSet,
-                        ChangedTrackIdSet* trackChanges);
+                        ChangedTrackIdSet* trackChanges, bool isFirstSegmentOfMeasure);
```
RomanPudashkin, line 112: Since trackChanges here is an output argument, I would suggest changing the order of the arguments:
processSegment(..., isFirstSegmentOfMeasure, trackChanges)

Issue #15960, opened 2023-01-19
[MU4 Issue] Cannot create tuplet from durations of 5 or 9 units (regression)


oktophonie:
**Describe the bug**
Cannot create tuplets from durations which are 5 or 9 units (e.g. a bar rest in 5/8 or 9/8). 7 seems okay.
This is a regression from 3.6.

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a 5/8 or 9/8 time signature to a bar
2. Select the bar rest
3. Try to create a tuplet with Ctrl+number
4. Behold error;
![image](https://user-images.githubusercontent.com/20604032/213386007-9964694f-8c11-4e01-90b7-9c70fd9a237c.png)

You can still add the tuplet by specifying the ratio explicitly (Add > Tuplets > Other), but this should not be necessary, and wasn't in 3.6.

Likely related to changes in https://github.com/musescore/MuseScore/pull/12931 and/or https://github.com/musescore/MuseScore/pull/13233.

**Expected behaviour**
MuseScore should be able to determine the denominator automatically, as it did in 3.6, and create the expected tuplet (e.g. x:5 or x.9).

See:
https://musescore.org/en/node/339357
https://musescore.org/en/node/342841

PR #15966, opened 2023-01-19
Fix tuplets entry bug

mike-spa:
Resolves: #15960

### DIFF
--- a/src/engraving/libmscore/tuplet.cpp
+++ b/src/engraving/libmscore/tuplet.cpp
@@ -1431,4 +1431,19 @@ void Tuplet::addMissingElements()
              missingElementsDuration.numerator(), missingElementsDuration.denominator());
     }
 }
+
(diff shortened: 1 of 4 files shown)


oktophonie (approved):

src/engraving/libmscore/tuplet.cpp
```diff
@@ -1431,4 +1431,16 @@ void Tuplet::addMissingElements()
              missingElementsDuration.numerator(), missingElementsDuration.denominator());
     }
 }
+
+int Tuplet::computeTupletDenominator(int numerator, Fraction totalDuration)
+{
+    Fraction noteDuration = totalDuration / numerator;
```
RomanPudashkin, line 1437: It would be better to make sure that we don't divide by zero here

mike-spa: right, good point

RomanPudashkin (approved):

Issue #15805, opened 2023-01-10
[MU4 Issue] Slur overlapping with Tie


Gobbel2000:
**Describe the bug**
When entering the following notes (the A's are tied, then all notes are connected by a slur), the slur overlaps with the tie.

![slur_overlap](https://user-images.githubusercontent.com/26202156/211567008-9d4ee60b-1fb3-41c1-a7e1-e98d799c0ad3.png)

**Expected behavior**
There should be a gap between both lines.

**Platform information**
 - OS: Arch Linux
 - Version: Latest master branch (`a6ec87df`)

**Additional context**
Could be similar to #15317 but is not resolved by #15482.

Jojo-Schmitz:
Just for the record: this is a regression vs. 3.x:
![image](https://user-images.githubusercontent.com/1786669/234061605-7466a1da-b183-432f-a498-6bc3302e6a5b.png)

PR #15986, opened 2023-01-20
Implementing collision avoidance for slurs VS ties and tuplets

mike-spa:
Resolves: #15805

This wasn't a bug, but more of a missing feature. Collision avoidance with ties and tuplets was never implemented (and also wasn't in 3.6) because it is a bit tricky. This is a first attempt at it. Seems to work quite nicely, but like most things it'll probably need gradual refinement.


mike-spa:
(contains some good stuff, but we'll come back to this once we have a more complete solution)

PR #16009, opened 2023-01-22
Add setting to choose whether invisible parts should be muted or not

### LINKED ISSUE #13787: [MU4 Issue] Playback setup in Mixer becomes greyed out after next build installation (Windows only)
Capture

**Describe the bug**
Strange issue. It happens after installation of the next build on Windows. The same option in Top menu>View works fine. 

Factory reset fixes the problem

**To Reproduce**

**Platform information**
 Windows 10
Doesn't occur on Mac/Linux

cbjeukendrup:
Resolves: #15945

This is a (possibly temporary) solution for #15945, that allows users to choose whether invisible parts should be muted:
image
The value of this setting is saved with the score.

If you enable this setting (default), then things will behave like they do now.
If you disable it, then MuseScore won't mute any invisible parts at all, not even in part scores, so that users can do whatever they want. 

This is obviously not a full solution, but at least it provides users who need to override MuseScore's behaviour with a way to do so.

The second commit should resolve #13787.

### DIFF
--- a/src/playback/internal/playbackcontroller.cpp
+++ b/src/playback/internal/playbackcontroller.cpp
@@ -47,6 +47,7 @@ static const ActionCode COUNT_IN_CODE("countin");
 static const ActionCode PAN_CODE("pan");
 static const ActionCode REPEAT_CODE("repeat");
 static const ActionCode PLAY_CHORD_SYMBOLS_CODE("play-chord-symbols");
+static const ActionCode TOGGLE_MUTE_INVISIBLE_PARTS_CODE("toggle-mute-invisible-parts");
 static const ActionCode PLAYBACK_SETUP("playback-setup");
 
 void PlaybackController::init()
@@ -63,6 +64,7 @@ void PlaybackController::init()
     dispatcher()->reg(this, METRONOME_CODE, this, &PlaybackController::toggleMetronome);
     dispatcher()->reg(this, MIDI_ON_CODE, this, &PlaybackController::toggleMidiInput);
     dispatcher()->reg(this, COUNT_IN_CODE, this, &PlaybackController::toggleCountIn);
+    dispatcher()->reg(this, TOGGLE_MUTE_INVISIBLE_PARTS_CODE, this, &PlaybackController::toggleMuteInvisibleParts);
     dispatcher()->reg(this, PLAYBACK_SETUP, this, &PlaybackController::openPlaybackSetupDialog);
 
     globalContext()->currentNotationChanged().onNotify(this, [this]() {
@@ -70,6 +72,8 @@ void PlaybackController::init()
     });
 
     globalContext()->currentProjectChanged().onNotify(this, [this]() {
+        notifyActionCheckedChanged(TOGGLE_MUTE_INVISIBLE_PARTS_CODE);
+
         if (m_currentSequenceId != -1) {
             resetCurrentSequence();
             return;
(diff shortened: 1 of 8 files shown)


cbjeukendrup:
@bkunda Could you please try this out to see if this might be a sensible temporary solution for 4.0.2?

Jojo-Schmitz:
Could we maybe make this the default for scores imported from earlier MuseScore versions?

MarcSabatella:
I can't really tell - is this setting global for the score & all parts, or is it per score & part?  It kind of feels like maybe it's the former, but I think the latter would be much better.  In fact, I'd suggest that would actually fix the problem pretty completely.  The score should default to playing invisible staves, the parts should default to muting them.  As far as I'm concerned that would virtually eliminate user surprise and solve pretty much all the use cases.  The defaults will be perfect for most use cases, and turning off the mute for parts as needed will solve the special use case of wanting  to hear the score while looking at a part.

If it's a global control, is far less ideal, because you'll constantly need to be toggling the option on and off.

cbjeukendrup:
It's a global toggle for the whole score, because we have no way yet to store playback info per part score. (That's also what's holding us back in implementing a full solution; it's not necessarily difficult, but we need to be sure that that is indeed the way to go.)

It might be a bit more useful if I made it a tri-state thing:
1. Mute invisible parts always
2. Mute invisible parts in part scores (=MS3 default)
3. Mute invisible parts never (can be achieved in MS3)

How about that?

Jojo-Schmitz:
Sounds (pun intended) OK to me.
And make option 2 the default for pre 4.x scores ?

cbjeukendrup:
Yeah, that's probably doable. Will look at it tomorrow.

MarcSabatella:
I would want option 2 to be the default for new scores as well.  The only ones that should arguably get option 1 by default are existing 4.0 and 4.0.1 scores, and even then to me it's debatable, as I see no evidence that any user actually wants this behavior by default.

So in fact, if it were up to me I'd drop option 1 completely and just offer options 2 & 3, as a single checkbox "Mute invisible instruments in parts" (defaulting to on).  But, a second checkbox "Mute invisible instruments in score" (defaulting to off) would be harmless, and maybe a little less confusing than the tri-state option.  But I'll let others worry about the UI aspect.

In any case, I would also like to see the wording mention invisible *instruments*, not invisible *parts*, the control over invisibility is in the Instruments panel and not the Parts dialog.  Talking about invisible parts is confusing.

Jojo-Schmitz:
I like this better too

Tantacrul:
I'm afraid that I disagree on multiple levels here.

1. The toggle is undiscoverable, so the only way a user will know to find it is to be told by a community member or read it in the manual. This does not benefit the general user because they won't know it exists.
2. We should **never** resort to temporary UX hacks. It makes the app confusing, illogical and messy.

As I've mentioned elsewhere, I'm currently looking at a solution that allows users to mute/unmute at will. Still working on it but it won't be long. Sorry. We spent years trying to clean up the mishmash of unrelated settings strewn all over MS3. We don't want to start reintroducing them again in 4.

RomanPudashkin:
> The second commit should resolve https://github.com/musescore/MuseScore/issues/13787.

@cbjeukendrup please open a separate PR containing this commit. Thank you!

cbjeukendrup:
> @cbjeukendrup please open a separate PR containing this commit. Thank you!

Done, see #16155!

Issue #16012, opened 2023-01-22
UI is hard to use when the user is moving the mouse while clicking


Tantacrul:
**Describe the bug**
See this video uploaded by a user who is unable to select many items on the interface: https://www.youtube.com/watch?v=KiGPBrbKeN0

I'm guessing there is some issue where mouse clicks are being cancelled when the cursor is being moves (even a little bit).

I'm not sure what the solution is:
- Trigger UI events on MouseUp, rather than MouseDown?
- Ignore cursor movement when clicking on UI elements?
- Establish a permitted amount of cursor drift when clicking?

**Expected behavior**
This user should be having none of these problems.

cbjeukendrup:
I found the reason; it's relatively easy to solve, fortunately.

We do most things "on click" (rather than "mouse up" or "mouse down"). From the QML docs:
> A click is defined as a press followed by a release, both inside the [MouseArea](https://doc.qt.io/qt-5/qml-qtquick-mousearea.html) (pressing, moving outside the [MouseArea](https://doc.qt.io/qt-5/qml-qtquick-mousearea.html), and then moving back inside and releasing is also considered a click).

So it should work... but it doesn't.

And that turns out to be because we have attached a signal handler to the `pressAndHold` signal. In this handler, we simply forward the signal by calling another signal. This forwarded signal is then used in a few places throughout the app. But the handler where this happens, is thus connected to the `pressAndHold` signal for every FlatButton.
When QML detects that something is connected to this signal, the button starts apparently behaving differently and the click signal doesn't work anymore as described; it only works if you keep the mouse still. I'm inclined to call this a QML bug, but that doesn't really bring us further...
So to fix it for 99,9% of all buttons*, we should simply make sure that nothing is connected to that `pressAndHold` signal. The easiest way to do it, is by removing that "forwarded" signal, and in places where it was used simply connect to `mouseArea.pressAndHold` directly. A slightly more advanced way would be to put the "forwarder" handler in a `Connections` object that is only `enabled` when something is connected to the forwarded signal. It's currently not possible to check that in QML directly, so we'll have to write a small C++ utility function to achieve that. Not sure whether that's worth the effort.

\* Namely those where we don't have any functionality on "pressAndHold"


Alternatively, we could try if something like TapHandler works better than MouseArea, but I have a gut feeling that it will be only worse.

DmitryArefiev:
Retested on Win10

PR #16013, opened 2023-01-23
Fix clicking FlatButton if you (accidentally) move the mouse between press and release

cbjeukendrup:
See https://github.com/musescore/MuseScore/issues/16012#issuecomment-1399656043

Resolves: #16012

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/FlatButton.qml
@@ -93,7 +93,8 @@ FocusScope {
     }
 
     signal clicked(var mouse)
-    signal pressAndHold(var mouse)
+    // There are intentionally no "forwarded" signals here from the MouseArea, like `pressAndHold`
+    // See https://github.com/musescore/MuseScore/issues/16012#issuecomment-1399656043
 
     objectName: root.text
 
@@ -288,10 +289,6 @@ FocusScope {
 
         onPressed: {
             ui.tooltip.hide(root, true)
-        }
-
-        onPressAndHold: function(mouse) {
-            root.pressAndHold(mouse)
         }
 
         onContainsMouseChanged: {
--- a/src/notation/qml/MuseScore/NotationScene/NoteInputBar.qml
+++ b/src/notation/qml/MuseScore/NotationScene/NoteInputBar.qml
@@ -168,12 +168,20 @@ Item {
                 }
             }
 
-            onPressAndHold: {
-                if (menuLoader.isMenuOpened || !hasMenu) {
-                    return
-                }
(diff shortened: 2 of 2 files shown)


RomanPudashkin (approved):

Tantacrul:
Hey @MarcSabatella & @bkunda 

When you get a chance can you test this and report back. I'm not near a PC right now (trackpad on MacOS) so can't test this.

cbjeukendrup:
I don't think that a physical mouse or not will make much of a difference. Probably QML can't even see which kind of device you're using. This is really all at software level.

Tantacrul:
> I don't think that a physical mouse or not will make much of a difference. Probably QML can't even see which kind of device you're using. This is really all at software level.

I'm just unable to practically notice a difference with my fancy Mac trackpad.

bkunda:
Tested this on Windows and Mac as a side-by-side comparison with 4.0.1. I can confirm there is definitely a noticeable increase in the range of mouse cursor movement accepted by the app while still triggering the required action. So this fix has definitely addressed the problem (Incidentally, I tested it on all the UI elements identified in the YT video linked in #16012, plus a range of other elements too).

The only additional point I might flag - which was not highlighted in the abovementioned video - is that a similar kind of additional mouse cursor movement when clicking on a tabbed panel (palettes, properties, instruments) can very easily cause the panel to become undocked. 

Not sure if this could also be addressed in this same PR, but I imagine this could also cause similar frustration if a person unintentionally moved their mouse slightly when trying to switch to a different tab. 

Otherwise, this PR seems to have addressed the problem.

Tantacrul:
So in other words, we need to increase the threshold for how much a user must move their mouse for these elements i order to undock them?

@RomanPudashkin - that seems like a sensible improvement. Is it easy to do?

RomanPudashkin:
> Is it easy to do?

It seems so, but I'm not 100% sure :)

@cbjeukendrup try to play around with startDragDistance(), which is defined in Utils_p.h. Unfortunately, there is no public API for modifying this setting

bkunda:
I've just tried this out at @Eism's request. If I'm honest, I can't really detect any change in behaviour here (when comparing to the current stable release): I.e. In this PR the **palettes**, **properties**, **instruments** and **filter selection** panels are still just as easy to undock with a very slight movement of the mouse (while mouse click is down) as they are in v.4.0.1.

bkunda:
@cbjeukendrup I've just tested this (with @Eism's help) with a value of 40 (startDragDistance) and this seems to have solved the problem of unintended panel undocking nicely. I believe this PR will be ready to merge once this change has been applied to your PR 🙂.

Thanks!

Eism:
I will merge it and fix the problem in separate PR

Issue #15615, opened 2022-12-27
[MU4 Issue] MS Basic and SoundFonts are playing largely out of tune, while Muse Sounds and most VSTs are fine (using 192kHz sample rate output)


MineCake147E:
**Describe the bug**
The frequency of the A4 should be exactly 440Hz, but the MS Basic produces the same sound at 1915Hz, about 4.35 times higher. 
The pitch differences between SoundFont and Muse Sound are consistent among the different SoundFonts, including MS Basic.  
The number 4.35 is close to 192000/44100, where 192000 is the output sampling frequency.  

**To Reproduce**
1. Get some device that runs MuseScore 4 and has an 192kHz output.
2. Set sampling frequency of output device to 192kHz.
3. Launch MuseScore 4.
4. Create an empty score with piano part.
5. Play the A4 note with Grand Piano from Muse Sound.
6. Play the A4 note with MS Basic.
7. You'll see the pitch difference of these two.

**Expected behavior**
Piano from MS Basic plays A4 at 440Hz.

**Screenshots**
<details><summary>Image of MuseScore 4</summary>

![image](https://user-images.githubusercontent.com/31585494/209610617-935c0f1f-d276-4cf8-9b4a-8dcba551805f.png)

</details>
<details><summary>Spectrum analysis of A4 note played with Piano from MS Basic, in Audacity</summary>

![image](https://user-images.githubusercontent.com/31585494/209610713-78db18df-5a3f-49db-b8e0-17a2b1c27187.png)

</details>

**Platform information**
- PC1: Modified DELL XPS 8700
  - OS: Windows 10 Home 22H2 Build 19045.2364
  - CPU: Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
  - RAM: 32.0 GB DDR3 DIMM
  - Graphics: NVIDIA GeForce GTX 1060 (6 GB) from 玄人志向
  - Audio: VB-Audio VoiceMeeter VAIO @ 24bit 192kHz Stereo
  - Storage: 
    - OS (C:) : Western Digital WDC WDS100T2B0A-00SM50 1000.2 GB SATA SSD
    - WORKSPACE (D:) : Seagate ST1000DM003-1ER162 1000.2 GB 7200RPM SATA HDD
    - DATA (E:) : Western Digital WDC WD10EZEX-22MFCA0 1000.2 GB 7200RPM SATA HDD
[...]

Issue #16014, opened 2023-01-23
[MU4 Issue] Playback is choppy, disoriented, and higher pitched for no reason at all.


ParaliyzedEvo:
**Describe the bug**
As soon as I updated to version 4.01 (I think) the playback started to sound choppy, disoriented, and higher pitched.

**To Reproduce**
1. Idk how to reproduce since all you have to do is press the playback and you get the problem

**Expected behavior**
Playback is supposed to sound perfect and good ig?

**Screenshots**
I got a video instead of a screenshot
https://user-images.githubusercontent.com/107338855/213964061-67a3e2e9-1201-45ca-b6d3-a1f75d80aa52.mp4

**Platform information**
 - OS: Windows 11

**Additional context**
It started when I updated Musescore

MarcSabatella:
It works fine normally, but it could be some sort of incompatibility with your particular audio device or its settings.  Please say what device you are using and what settings you've made in your OS for it (eg, the sample rate).

ParaliyzedEvo:
> It works fine normally, but it could be some sort of incompatibility with your particular audio device or its settings. Please say what device you are using and what settings you've made in your OS for it (eg, the sample rate).
![Screenshot_20230122_104224](https://user-images.githubusercontent.com/107338855/213967173-fd813271-d36f-4353-9dad-0f6d6c0884b6.png)

MarcSabatella:
Try a more reasonable sample rate like 44.1 or 48 kHz.

ParaliyzedEvo:
> Try a more reasonable sample rate like 44.1 or 48 kHz.

Ok, I'll try tomorrow since I'm going to sleep

cbjeukendrup:
@ParaliyzedEvo I've made a fix that may resolve this issue: https://github.com/musescore/MuseScore/pull/16018

Would be great if you could try out the test builds for this version (they're ready in 30 minutes). See https://github.com/musescore/MuseScore/wiki/Downloading-and-running-test-builds#Downloading-Builds-from-Pull-Requests for how to download these builds.

ParaliyzedEvo:
> @ParaliyzedEvo I've made a fix that may resolve this issue: #16018
> 
> Would be great if you could try out the test builds for this version (they're ready in 30 minutes). See https://github.com/musescore/MuseScore/wiki/Downloading-and-running-test-builds#Downloading-Builds-from-Pull-Requests for how to download these builds.

Thanks

[...]

PR #16018, opened 2023-01-23
Fix playback with FluidSynth at higher sample rates

cbjeukendrup:
Resolves: #15615
Resolves the high-pitch part of: #16014

I `git blame`d the FluidSynth source code to see if there is a reason that FluidSynth only supports sample rates up to 96kHz, but the only thing I found was that it has just always been like that, for the past 20 years. So I expect it should be possible to move this limit up a bit. 

Also adds support for exporting audio at higher sample rates. Since we (intend to) support those for live playback, I see no problem in supporting them for export too.

### DIFF
--- a/thirdparty/fluidsynth/fluidsynth-2.3.3/src/synth/fluid_synth.c
+++ b/thirdparty/fluidsynth/fluidsynth-2.3.3/src/synth/fluid_synth.c
@@ -228,7 +228,7 @@ void fluid_synth_settings(fluid_settings_t *settings)
     fluid_settings_register_int(settings, "synth.audio-groups", 1, 1, 128, 0);
     fluid_settings_register_int(settings, "synth.effects-channels", 2, 2, 2, 0);
     fluid_settings_register_int(settings, "synth.effects-groups", 1, 1, 128, 0);
-    fluid_settings_register_num(settings, "synth.sample-rate", 44100.0f, 8000.0f, 96000.0f, 0);
+    fluid_settings_register_num(settings, "synth.sample-rate", 44100.0f, 8000.0f, 384000.0f, 0);
     fluid_settings_register_int(settings, "synth.device-id", 0, 0, 127, 0);
 #ifdef ENABLE_MIXER_THREADS
     fluid_settings_register_int(settings, "synth.cpu-cores", 1, 1, 256, 0);


cbjeukendrup:
Updated to support 384kHz too, after https://github.com/musescore/MuseScore/issues/15615#issuecomment-1401403831

src/importexport/audioexport/internal/audioexportconfiguration.cpp
```diff
@@ -69,6 +69,6 @@ void AudioExportConfiguration::setExportSampleRate(int rate)
 
 const std::vector<int>& AudioExportConfiguration::availableSampleRates() const
 {
-    static const std::vector<int> rates { 32000, 44100, 48000 };
+    static const std::vector<int> rates { 32000, 44100, 48000, 96000, 192000 };
```
Jojo-Schmitz, line 72: Shouldn't this go up to 384000 too now?

cbjeukendrup: Okay, done. I'm not sure if we should tempt people to use such crazily high sample rates, but on the other hand, if we support it for playback, why not for export, for those people that really want it...

DmitryArefiev:
@cbjeukendrup I can't check playback on 192kHz (don't have a audio device supports that). Checked only audio export (Win/Mac).
Seems to be fine. Only in **MP3 export settings**, the sample rates **higher than 48kHz should be removed**

MP3 doesn't support it

Capture3

Capture2

And actually, if you export MP3 in 96-348 it gives MP3 file with 48kHz anyway

cbjeukendrup:
@DmitryArefiev Thanks, I removed them for MP3. Now the settings for MP3 and other audio types are separated, so they won't interfere.

DmitryArefiev:
@cbjeukendrup Now I started to doubt that we need 384000Hz in Export sample rate settings :) Too much values we have now
I think it's very very rare case. 

@bkunda Please approve that we need or not 384000Hz in Export format settings for WAV/FLAC


Screenshot 2023-08-31 112855

(also we can we can get rid of 32000Hz since it's rare case for WAV too)

RomanPudashkin (approved):

cbjeukendrup:
@DmitryArefiev Since the user in https://github.com/musescore/MuseScore/issues/15615 already reported that it's working for them, I will merge it after the checks succeed, is that okay?

Issue #15979, opened 2023-01-19
[MU4 Issue] Accessibility: keyboard focus remains in menu when adding text


MarcSabatella:
**Describe the bug**
If you use the keyboard to access the Add menu to add text to your score, keyboard focus remains in the menu, making it impossible to type letters that correspind to menu mnemonics (eg, "F" for "File")

**To Reproduce**
1. Select any note or rest
2. Press Alt+A to open the Add menu
3. Cursor down to Text
4. Press Right to open the Text submenu
5. Cursor down to Staff text
6. Press Enter to add a staff text
7. Start typing
8. See errror: you can't type the letters F, E, V, A, O, T, P, H, or D, because these keys open the corresponding menu (File, Edit, etc)

**Expected behavior**
Keyboard focus should be on the score so you can type normally

**Platform information**
 - OS: Linux

**Additional context**
Workaround is to press Esc to return focus to the score before you begin typing.  Orca at least does read the menu as it opens, giving you a clue that something has gone wrong.   Not sure if that's true for all screenreaders.

jmwmusic:
@MarcSabatella could you retest this it seems to be working fine for me now with NVDA.

MarcSabatella:
I can confirm it is now working as expected for me on Linux.  I don't see a specific PR that fixed it, but it's also possible it was fixed by a Qt update.

PR #16020, opened 2023-01-23
fixed #15897: Entering Roman numeral analysis crashes when using screen reader

### LINKED ISSUE #15897: [MU4 Issue] Entering Roman numeral analysis crashes when using screen reader
**Describe the bug**
Enter RNA text crashes MuseScore if a sxcreen reader isrunning

**To Reproduce**
1. New score (be sure to have a screen reader running)
2. Select first rest
3. Add / Text / Roman numeral analysis
4. Type I
5. Press Esc
6. See error: crash

**Expected behavior**
The text is entered normally

**Screenshots**

**Platform information**
Confirmed using NVDA, JAWS, and Narrator on Windows, also Orca on Linux.  In some cases apparently the crash might happen sooner, like right after step 3, depending on how quickly you type I guess.

See https://musescore.org/en/node/342415 for more info

Eism:
Resolves: #15897

The C++ standard library's regex implementation does not support lookarounds.

### DIFF
--- a/src/engraving/libmscore/harmony.cpp
+++ b/src/engraving/libmscore/harmony.cpp
@@ -2104,11 +2104,14 @@ String Harmony::generateScreenReaderInfo() const
             { u"i", u"1" },
         };
         static std::vector<std::pair<String, String> > symbolReplacements {
-            { u"b", u"♭" },
-            { u"h", u"♮" },
-            { u"#", u"♯" },
             { u"bb", u"𝄫" },
             { u"##", u"𝄪" },
+            { u"h", u"♮" },
+            { u"\\♮", u"h" }, // \h should be h, so need to correct replacing in the previous step
+            { u"#", u"♯" },
+            { u"\\♯", u"#" }, // \# should be #, so need to correct replacing in the previous step
+            { u"b", u"♭" },
+            { u"\\♭", u"b" }, // \b should be b, so need to correct replacing in the previous step
             // TODO: use SMuFL glyphs and translate
             //{ "o", ""},
             //{ "0", ""},
@@ -2119,10 +2122,7 @@ String Harmony::generateScreenReaderInfo() const
             aux.replace(r.first, r.second);
         }
         for (auto const& r : symbolReplacements) {
-            // only replace when not preceded by backslash
-            String s = u"(?<!\\\\)" + r.first;
-            std::regex re(s.toStdString());
-            aux.replace(re, r.second);
+            aux.replace(r.first, r.second);
         }
         // construct string one  character at a time
         for (size_t i = 0; i < aux.size(); ++i) {


igorkorsukov (approved):

MarcSabatella:
Thanks!  This definitely prevents the crash, and reads correctly for "\b".  But the same change would need to be made for "#" and "h", so "\#" reads "number" rather than "backslash sharp", and "\h" reads "h" instead of "backslash natural".  Actually for me using Orca natural doesn't read at all currently, but I think that's just that Orca doesn't know this Unicode symbol.

The way your code works, I don't think this technique is needed for "bb" or "##", because if you literally want "bb", you need to type \b\b.  The code you have in place already handles that.

Eism:
@MarcSabatella fixed

MarcSabatella:
Looks great, thanks for dealing with this!

NoahCarverMusic:
Excellent. Thank you so much, @Eism and @MarcSabatella, for your work on this. It's greatly appreciated.

NoahCarverMusic:
Hello. Now using MuseScoreNightly-230270506-4.0.2-3e9f131-x86_64 and MuseScoreNightly-230270305-master-a5382f4-x86_64. The Roman numeral analysis crash has been fixed, but text entry seems not to be possible. Pressing enter on the RNA menu item seems to send focus to the text field but entry is not possible. Further, the right and left arrows continue to move focus through the menu bar. Pressing escape seems to return focus to the score as arrow keys move focus through text fields for NRA entry, however again, text entry seems not to work as once I press escape again, no RNA can be found in the score. Tested with latest stable JAWS and NVDA.

MarcSabatella:
Yes, there is still a bug where focus remains in the menu - see https://github.com/musescore/MuseScore/issues/15979.  Pressing Esc does work, but it is possible you may have hit it too many times, or at the wrong point.  Try pressing Esc exactly once, immediately after activating the RNA menu item.  This works for me.  But it will definitely be best to get the focus bug fixed!

MarcSabatella:
Also, it should work now to assign a keyboard shortcut to RNA - except unfortunately the shortcut dialog is currently not accessible, it's impossible to save the shortcut and close the dialog using the keyboard alone.

NoahCarverMusic:
Here's a video demo of what I'm experiencing. I am only pressing escape once as shown in the video, but this bug still occurs. Hope this helps.

https://youtu.be/GDqIj7M1HlA

MarcSabatella:
Thanks for the video!  It definitely seems that you are doing the correct things, and I think the screen reader feedback sounds correct.  However, unfortunately, your application is not large enough to show the score - the mixer is covering it all.  So I am unable to see much more about what is actually happening.  However, one thing I *do* see in the mixer is that when you select the Roman numeral analysis item in the menu, the mixer adds a new channel for the chord symbol (which I guess is a bug since these do not actually play), so it is at least *trying* to add this.  When you first hit Esc, the channel remains, and it does when you press "I" as well, but then when you hit Esc again, the channel is removed.  This is consistent with what I'd expect if the "I" in fact did *not* get sent to the edit field.  In that case, since no text was entered, the new analysis text would simply deleted upon hitting Esc since it is empty.

And yet when I reproduce exactly those same steps, all is well for me.

So can you perhaps do the same thing again but first make sure the mixer is not open?  F10 toggles it, or you can use the View menu.  You can F6 or tab around the interface to be sure it isn't open.  Just that one little section starting from navigating the menu up through when you navigate through the score to show the text isn't there.

NoahCarverMusic:
I tried it again tonight... and it works now both with the mixer turned on or off. There were no changes to screen reader configuration, I'm using the same build, MuseScoreNightly-230270506-4.0.2-3e9f131-x86_64, there were no system changes that I'm aware of, and I followed the same steps as outlined in the video I posted a few days ago. Put simply, it works now, and I have no idea why. Bizarre.

Eism:
@NoahCarverMusic 
I fixed your issue and merged to master
This is a direct link to Windows build https://github.com/musescore/MuseScore/suites/10655099799/artifacts/532549830

NoahCarverMusic:
After more testing, it seems that @Eism's fix works as intended (barring the focus issues mentioned by @MarcSabatella) when MuseScore 4 launches as a maximized window or is manually maximized by the user. However, the Roman numeral input is not recognized by MS4 if the application is launched but its window is not maximized either automatically or manually. I'll try to get this on video. I'm seeing this with both JAWS and NVDA on Windows 11; unfortunately, I can't test with other operating systems at the moment.

Also, @Eism, thanks for the info. I'm aware your issue was merged to master; I really appreciate your work. Both @MarcSabatella and I are now in agreement that RNA input with a screen reader works as intended now barring what I seem to have gradually discovered over the past few days regarding MS4 not recognizing RNA input when its window is not maximized at least on Windows. I've been testing a build that (as far as I know) contains the fix given it no longer crashes when entering RNA, however I can also test using the build you just posted.

Issue #16021, opened 2023-01-23
[MU4 Issue] 'Expand repeats' checkbox can't be unchecked in Midi export dialog


abariska:
**To Reproduce**
1. Create a score
2. Open Export dialog
3. Choose Midi export
4. Try to uncheck Expand repeats

**Expected behavior**
'Expand repeats' checkbox can be unchecked

**Screenshots**

https://user-images.githubusercontent.com/101250347/214040654-977f9e22-e132-41e2-b34d-e5a0cfe8beb1.mp4

**Platform information**
 - OS: macOS 12.6

abariska:
🟢 **FIXED**
Tested on macOS 12

Issue #16029, opened 2023-01-23
[MU4 Issue/Task] Inconsistencies regarding "Expand repeats" setting


cbjeukendrup:
**Describe the bug**
There are various problems with the "expand repeats" setting:
- When exporting audio, we use the value that is used for in-app playback
- That results in #15720 
(MuseScore.com apparently expects that we always expand repeats when uploading audio)
- For MIDI export: #16021
(I already fixed the checkbox locally, but now that the checkbox itself is working, these underlying problems surface)
- The `Score::extractLyrics` method assumes that `masterScore()->setExpandRepeats(…)` works (but it does nothing), possibly resulting in wrong lyrics (this method is used both for copying lyrics and for the MuseScore.com backend)

**The problem**
We have not one, not two, not three, but ***four*** sources of truth:
- `MScore::playRepeats`
- `MasterScore::_expandRepeats`
- `PlaybackModel::m_expandRepeats`
- `NotationConfiguration::isPlayRepeatsEnabled()`

Not very surprising that we have bugs...

abariska:
@cbjeukendrup 
- Regression in the PR according to https://github.com/musescore/MuseScore/issues/16021 

Precondition - Expand repeats' is enabled. 
Notes from further measures are overlayed the repeat section  when played second time

https://user-images.githubusercontent.com/101250347/225894251-356324e8-ede3-4670-9862-253c418ebc6f.mp4


- Audio and cursor are still not synchronised when score exported with 'Play repeats' disabled

https://user-images.githubusercontent.com/101250347/225897293-c89843aa-36ea-47af-99ba-25a1114edd45.mp4


[Test score](https://github.com/musescore/MuseScore/files/11001004/repeat.midi.mscz.zip)

abariska:
🟢 **TESTED on macOS 12

PR #16033, opened 2023-01-23
Cleanup "Expand repeats" option; fix it for MIDI export

cbjeukendrup:
Resolves: #16021 
Improves: #16029 

This eliminates one source of truth, namely MScore::playRepeats, and creates a clear structure for the other three: notation configuration determines playback model, and playback model (and other things) determine master score. 
Also optimise the use of the two RepeatLists in MasterScore; one is for expanding, the other not. Previously, one was for general use and switched between expanded and not expanded every time; the other only for a few specific methods.

### DIFF
--- a/src/engraving/libmscore/masterscore.cpp
+++ b/src/engraving/libmscore/masterscore.cpp
@@ -62,8 +62,8 @@ MasterScore::MasterScore(std::weak_ptr<engraving::EngravingProject> project)
     _undoStack   = new UndoStack();
     _tempomap    = new TempoMap;
     _sigmap      = new TimeSigMap();
-    _repeatList  = new RepeatList(this);
-    _repeatList2 = new RepeatList(this);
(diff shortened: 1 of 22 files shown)


RomanPudashkin (approved):

DmitryArefiev:
@cbjeukendrup Is this ready for testing?

cbjeukendrup:
@DmitryArefiev I have just rebased, and I think it should be ready now. A while ago, Borys had already tested it, but found an issue: https://github.com/musescore/MuseScore/issues/16029#issuecomment-1473721434 but I think this is resolved by the rebase, because I suspect it has been fixed by https://github.com/musescore/MuseScore/pull/17174. I haven't yet got the chance to verify that myself though.

DmitryArefiev:
Cool! Thanks!

abariska:
🟢 **TESTED on macOS 12

PR #16034, opened 2023-01-23
Fix #15861: Printing scores prints square the number of requested copies

### LINKED ISSUE #15861: [MU4 Issue] Printing scores prints square the number of requested copies
**Describe the bug**
Printing scores prints square the number of requested copies

**To Reproduce**
1. File > Print
2. Select number of copies: 2, 3 or 4
3. Print
4. Get 4, 9 or 16 copies printed, respectivly

**Expected behavior**
Well, oviously getting the exact number of requested copies...

**Platform information**
 - OS: Windows

**Additional context**
Came up in https://musescore.org/en/node/341823 and again in https://musescore.org/en/node/342447

cbjeukendrup:
On most systems, the QPrinter already prints $n$ copies of what we feed it with, so we ourselves don't need to feed it with $n$ copies too.

Resolves: #15861

Note: I have not tested this by really printing; only sending it to the printer queue and checking the info there. That's probably sufficient, but if anyone wants to test with real paper I'd recommend to create a score and delete all instruments and text frames, so that the score is only an empty page; that way, you can just reuse the paper after testing.

### DIFF
--- a/src/print/internal/printprovider.cpp
+++ b/src/print/internal/printprovider.cpp
@@ -66,7 +66,8 @@ Ret PrintProvider::printNotation(INotationPtr notation)
     INotationPainting::Options opt;
     opt.fromPage = printerDev.fromPage() - 1;
     opt.toPage = printerDev.toPage() - 1;
-    opt.copyCount = printerDev.copyCount();
+    // See https://doc.qt.io/qt-5/qprinter.html#supportsMultipleCopies
+    opt.copyCount = printerDev.supportsMultipleCopies() ? 1 : printerDev.copyCount();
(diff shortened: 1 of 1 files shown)


igorkorsukov (approved):

Jojo-Schmitz (approved):
Confirmed to fix the issue!
(And thanks for the paper-saving tip)

Jojo-Schmitz:
A simple `git cherry-pick` should be all there is to porting to to 4.0.2

cbjeukendrup:
Indeed; Roman takes care of this periodically, by cherry-picking a batch of PRs every time.

Issue #15987, opened 2023-01-20
[MU4 Task] Set up translations for new strings in 4.0.2


cbjeukendrup:
**Task description**
Version 4.0.2 will contain some new strings. We need to upload these to Transifex, without removing any strings that are currently there, because version 4.0.1 and 4.0.0 still rely on those strings.

**So it is very important that nobody pushes any new strings to Transifex right now.**

Two possible solutions:
- Make an option in the "push to Transifex" GitHub Action, that allows us to choose whether "obsolete" strings should be removed or kept; that means we would push the new strings without removing the old ones. This way, both 4.0.0, 4.0.1 and 4.0.2 can keep using the languages from `http://extensions.musescore.org/4.0/languages/`.
- Create a new endpoint for uploading changes from Transifex, for example `http://extensions.musescore.org/4.0.2/languages/`, so that all new changes will be sent to that place, while the old place stays unchanged. This has the disadvantage that users of 4.0.0 and 4.0.1 won't receive any translation updates while 4.0.2 is in development (not even released yet).

Jojo-Schmitz:
> **Task description** Version 4.0.2 will contain some new strings. We need to upload these to Transifex, with removing any strings that are currently there, because version 4.0.1 and 4.0.0 still rely on those strings.

with**out** removing any strings that are currently there

Responsible for that is the use, or here rather the non-use, of the `-no-obsolete` commandline option of the `lupdate` command

cbjeukendrup:
> with**out**

Phew, silly typo! Corrected.

> Responsible for that is the use, or here rather the non-use, of the `-no-obsolete` commandline option of the `lupdate` command

Indeed, that's what I was thinking of. We just need a way to tell the CI action to use this flag.

Jojo-Schmitz:
Or rather to **not** use it, we'd want to keeo those 'obsolete' strings (as they are not obsolete in 4.0.0 and 4.0.1). Not using that was common practise for all minor 2.x and 3.x releases, using it just once (and possibly manually) when moving to a new minor or major version, but not for patch releases

Jojo-Schmitz:
I believe this is done now

cbjeukendrup:
Last remaining step is to merge the changes to Master too. Apparently PRs to other branches don't automatically close issues.

Jojo-Schmitz:
Ah, I see. But it might get removed from the 4.0.2 project ;-)

PR #16055, opened 2023-01-25
Update translations for 4.0.2

cbjeukendrup:
Resolves: #15987

If we pass the `-no-obsolete` flag to `lupdate`, any obsolete strings are removed. However, since translations for 4.0.2 are pushed to the same S3 URL as for the 4.0.0 and 4.0.1 release, the translations will need to be compatible with all those versions; if we would remove the translations that are obsolete in 4.0.2, then 4.0.0 and 4.0.1 will be broken, because they still depend on these translations.

This is what we have done / should do:

Only now, for 4.0.2:
1. We released 4.0.1 (which contained no string changes, so no need to worry about 4.0.0).
2. We created the 4.0.2 branch, where the next patch release is being developed.
3. Checkout the 4.0.1 branch (yes, .1)
4. Run `lupdate` _with_ `-no-obsolete`, so that we have the "clean" translations for 4.0.0 and 4.0.1 (without any strings that were already obsoleted before the release of 4.0.0).
5. Commit these changes to the 4.0.2 branch (yes, .2); so 4.0.2 contains exactly the strings that were used in 4.0.0 and 4.0.1.
6. Whenever a string changes in 4.0.2, we run `lupdate` _without_ `-no-obsolete`, and push that to Transifex, but _don't_ commit this to the repo
→ Transifex then has everything that was in 4.0.1, plus new and changed strings that are _currently_ in the 4.0.2 branch. 
The reason that we don't commit is: if strings are obsoleted during development, they should be removed. If we _don't_ use `-no-obsolete`, that won't happen; if we _do_ use it, the obsolete strings from 4.0.1 are removed too, so that's no option either. Therefore, every time, we should use the 4.0.1 strings as a base, and then run `lupdate` on 4.0.2 without `-no-obsolete`, to add the _current_ 4.0.2 strings on top of it.

In general, while developing 4.0.(n+1):
1. 4.0.n has just been released, and the 4.0.(n+1) branch has been created
2. On 4.0.n, the translation files contain exactly the strings that were included in the 4.0.(n-1) release. 
3. On 4.0.n, run `lupdate` _without_ `-no-obsolete` to generate the definitive 4.0.n translations without deleting any 4.0.(n-1) translations, so that the resulting translations file will work with 4.0.n and all previous versions
4. Commit these changes to the 4.0.(n+1) branch
5. Whenever a string changes in the 4.0.(n+1) development branch, we run `lupdate` _without_ `-no-obsolete`, and push that to Transifex, but _don't_ commit this to the repo
    - Transifex then has:
        - everything that is new (and not obsoleted during development) of 4.0.(n+1)
        - everything that was used by 4.0.n and previous versions
    - The repo then still has only the 4.0.n strings in the 4.0.(n+1) branch, so that next time we want to push changes to Transifex, there won't be any strings that were obsoleted during the development of 4.0.(n+1) but never included in any release.

Then, when we are ready to stop translations for 4.0.x and when we are ready to start translations for 4.1 (the master branch):
1. Switch to a new URL for translations that will be used for 4.1.x
7. On the master branch, always run `lupdate` _with_ `-no-obsolete` and freely commit this to the repo; we don't need to care about any old strings anymore, and during active development, the use of `-no-obsolete` is necessary, so that strings that were changed and will never make it into any release are removed from the translation files
8. When the official 4.1.0 release has been released, do the process for patch releases as described above for 4.0.x.

I hope this is somewhat understandable; it is really very simple and easy, but the only thing that's difficult about it, is to explain it in few words :)
Anyway, I think I'm going to take care of this myself for now, so don't worry :)

### DIFF
--- a/.github/workflows/ci_lupdate.yml
+++ b/.github/workflows/ci_lupdate.yml
@@ -4,9 +4,13 @@ on:
   workflow_dispatch:
     inputs:
       publish:
-        description: 'Publish to Transifex: on - publish'
+        description: 'Publish to Transifex (on/off)'
         required: false
         default: 'off'
+      cleanup_obsolete:
+        description: 'Clean up obsolete strings (on/off)'
+        required: true
+
 jobs:
   lupdate:
     runs-on: ubuntu-20.04
@@ -27,17 +31,24 @@ jobs:
           fi   
         fi
 
+        LUPDATE_ARGS=''
+        if [ "${{ github.event.inputs.cleanup_obsolete }}" == "on" ]; then
+          LUPDATE_ARGS='-no-obsolete'
+        fi
+
         echo "DO_PUBLISH=$DO_PUBLISH" >> $GITHUB_ENV
         echo "DO_PUBLISH: $DO_PUBLISH"
         echo "BUILD_NUMBER=$BUILD_NUMBER" >> $GITHUB_ENV
         echo "BUILD_NUMBER: $BUILD_NUMBER"
-  
+        echo "LUPDATE_ARGS=$LUPDATE_ARGS" >> $GITHUB_ENV
+        echo "LUPDATE_ARGS: $LUPDATE_ARGS"
+
     - name: Setup environment
       run: |
         sudo bash ./build/ci/translation/qt_install.sh 
     - name: Run lupdate
       run: |
-        sudo bash ./build/ci/translation/run_lupdate.sh
+        sudo bash ./build/ci/translation/run_lupdate.sh ${{ env.LUPDATE_ARGS }}
     - name: Publish to Transifex
       if: env.DO_PUBLISH == 'true'
       run: |
--- a/share/locale/instruments_en.ts
+++ b/share/locale/instruments_en.ts
@@ -193,13 +193,6 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../instruments/instrumentsxml.h" line="265"/>
-        <source>Flute in E♭, sound a minor third above the standard flute.</source>
-        <comment>soprano-flute description</comment>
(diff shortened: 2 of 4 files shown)


Jojo-Schmitz:
Tle lupdate commit seems to have removed strings, which it probably shopuldn't have?

igorkorsukov:
@cbjeukendrup It turns out that on CI, we never need to run lupdate with the -no-obsolete option, maybe then it's better to remove it?  
since, if we need to commit, then we need to run it locally... or we need to make the commit  on CI

cbjeukendrup:
@igorkorsukov When we are in pre-release mode for the next major or minor update, then we do need to run on CI with -no-obsolete. 
(Because then we don't need to preserve strings for older versions, because then we upload the translations to a new URL on S3.)

Jojo-Schmitz:
with**out** `-no-obsolete` should/needs to be the default. Only when changing the download resource, so likely with a minor or major release, running it **once** with `-no-pbsolete` should be done (but doesn't strickly need to, only downside is that no longer used strings are sill out there, eating resources)

cbjeukendrup:
@Jojo-Schmitz That is not unexpected; the files that are currently in the repo were updated a long time ago, and the strings that are removed were probably obsoleted even before 4.0.0.

Jojo-Schmitz:
Ah, OK. Let's hope that to be the case ;-)

Jojo-Schmitz:
So always between any start of a new translation round and the release. Never after release, unless, a new translation round starts.

cbjeukendrup:
Exactly. (And during the development of patch releases, we run not _just_ without -no-obsolete, but specifically everything on top of the definitive strings file of the previous patch release; that is to prevent that strings that got obsoleted during the development of the patch release stick around.)

igorkorsukov (approved):

PR #16056, opened 2023-01-25
fixed before beat midi render of grace notes

alexpavlov96:


jeetee:
Pardon me, but how is this supposed to fix controlling the grace note timing?

You've added an additional local parameter to the NPlayEvent but nobody is reading from it?
In addition you've removed the effect that `graceOffsetOn/Off` have on the actual on and off times (which *are* being used)?

Can you elaborate on your thought process?

alexpavlov96:
> Pardon me, but how is this supposed to fix controlling the grace note timing?
> 
> You've added an additional local parameter to the NPlayEvent but nobody is reading from it? In addition you've removed the effect that `graceOffsetOn/Off` have on the actual on and off times (which _are_ being used)?
> 
> Can you elaborate on your thought process?

@jeetee I it wasn't ready, I ended up just cutting off the negative values

jeetee:
Would this now imply that it becomes impossible to get a grace note to play back before its beat?

jeetee:
Ah yes, had to expand the code view a bit to realise "on" was no longer ontime but ticks.
LGTM then.

handrok (approved):

PR #2669, opened 2016-06-15
fix #115191: wrong translation on English (AU)

Jojo-Schmitz:
map it to English (GB)

PR #16061, opened 2023-01-25
Let the Aussies and Kiwis ;-) fall back to British English

Jojo-Schmitz:
rather than to American English, just like it was in 3.x and 2.x (for the Aussies at least)

See the old https://musescore.org/en/node/115191, fixed in 2.0.4, via #2669


RobFog:
> Not sure about whether to do the same for New Zealand, en_NZ, too.

The same should be done for en_NZ as well, yes. See https://en.wikipedia.org/wiki/New_Zealand_English.

cbjeukendrup (approved):

bkunda:
🟢 Re-tested in OS: macOS 13.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230651546, revision: github-musescore-musescore-dbe7c6d. Remains FIXED.

PR #16062, opened 2023-01-25
export-midi: more careful calculation of pitch bend

handrok:
Problem is different elements change ME_PITCHBEND state of the same channel. They do it independently. Solution: don't render these events all at once, but collect all the changes and apply them once.


cbjeukendrup:
Wouldn't it be a good idea to move the MIDI export code out of the `engraving` module, to the `iex_midi` module at some point?

handrok:
> Wouldn't it be a good idea to move the MIDI export code out of the engraving module, to the iex_midi module at some point?

You are right. Actualy I've already started it. It's not quick because there are a lot of connections between midi and libmscore classes. For example Part has Instrument, Instrument has channels, bank, program. Channels origin from midi world. So to break up a connections takes effort.

alexpavlov96 (approved):

Jojo-Schmitz:
Is this somehow retated to my #10534 ?

Issue #15850, opened 2023-01-13
[MU4 Issue] Articulations too far from notes on single-line stave


oktophonie:
**Describe the bug**
Articulations, except staccato/tenuto, are too far from the notehead on a single-line stave (any other number of lines is fine):
![image](https://user-images.githubusercontent.com/20604032/212306663-6cb26a33-521a-430f-b691-27a4f2bba867.png)

**Platform information**
OS: Artix Linux, Arch.: x86_64, MuseScore version (64-bit): 4.0.1-230121751, revision: github-musescore-musescore-9b70a8c

PR #16078, opened 2023-01-26
The clickable bounding box of a item should not necessarily coincide with the engraving bounding box

mike-spa:
Resolves: #15850 

In our codebase, the clickable area of an item coincides with its bounding box. This is, of course, fine in most cases, but sometimes you may need the clickable area to be different (larger, usually). 
In this example, the bounding box of a staff becomes too small if the staff has only one line (like percussion), so it's very hard to click on it. To solve the issue, the bounding box of the staff was artificially increased, but this in turn was messing up the engraving, which is really silly. So I've introduced a virtual `clickableBoundingBox` field which returns by default the "normal" bounding box but can be overridden by items to return something else. I feel like this should have been done before. Anyway, that solves the engraving issue of #15850 while also providing a better solution than [#4210](https://github.com/musescore/MuseScore/pull/4210) which originally caused it.


src/engraving/libmscore/engravingitem.h
```diff
@@ -320,6 +320,12 @@ class EngravingItem : public EngravingObject
     virtual Shape shape() const { return Shape(bbox(), this); }
     virtual double baseLine() const { return -height(); }
 
+    virtual mu::RectF clickableBBox() const { return _bbox; }
+    mu::RectF clickablePageBoundingRect() const { return clickableBBox().translated(pagePos()); }
+    Shape clickableShape() const { return Shape(clickableBBox(), this); }
```
cbjeukendrup, line 325: I think this should be virtual too and return `shape()` instead by default. Otherwise, elements will be clickable in their whole bounding rectangle rather than only in their `shape()`, like currently. One element type where the difference will be very noticeable is multi-measure rests.

mike-spa: Good one, thanks!

oktophonie (approved):

src/engraving/libmscore/bsp.cpp
```diff
@@ -155,7 +155,7 @@ std::vector<EngravingItem*> BspTree::items(const RectF& rec)
     std::vector<EngravingItem*> l;
     for (EngravingItem* e : findVisitor.foundItems) {
         e->itemDiscovered = false;
-        if (e->pageBoundingRect().intersects(rec)) {
+        if (e->clickablePageBoundingRect().intersects(rec)) {
```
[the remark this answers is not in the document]
mike-spa: Turns out: it's not needed! I think it was the first thing I tried, and then it stayed there. I've removed it now as it seems to work fine without. Thanks!

src/engraving/libmscore/engravingitem.h
```diff
@@ -320,6 +320,12 @@ class EngravingItem : public EngravingObject
     virtual Shape shape() const { return Shape(bbox(), this); }
     virtual double baseLine() const { return -height(); }
 
+    virtual mu::RectF clickableBBox() const { return _bbox; }
+    mu::RectF clickablePageBoundingRect() const { return clickableBBox().translated(pagePos()); }
```
RomanPudashkin, line 324: This method is not used

src/engraving/libmscore/engravingitem.h
```diff
@@ -320,6 +320,12 @@ class EngravingItem : public EngravingObject
     virtual Shape shape() const { return Shape(bbox(), this); }
     virtual double baseLine() const { return -height(); }
 
+    virtual mu::RectF clickableBBox() const { return _bbox; }
```
RomanPudashkin, line 323: Probably we should use the term "hit" instead of "clickable" for consistency

mike-spa:
Removed the unused method and renamed everything to "hit" 👍

RomanPudashkin (approved):

PR #16086, opened 2023-01-26
Fix MSVC compiler warnings

Jojo-Schmitz:
* reg. int vs. size_t (C4267)
* reg. unused function parameters (C4100), introduced by #16062

Todo: a warning reg. conversion of float (C4244), introduced by #16062


igorkorsukov (approved):

Jojo-Schmitz:
I'd still like somenone, probably @handrok, to look after that remaining warning (which seems beyond my pay grade)

PR #16110, opened 2023-01-28
Canadian like British English

Jojo-Schmitz:
Feedback of Canadian needed

### DIFF
--- a/src/languages/internal/languagesservice.cpp
+++ b/src/languages/internal/languagesservice.cpp
@@ -220,6 +220,7 @@ QString LanguagesService::effectiveLanguageCode(const QString& languageCode) con
                 { "ca_ES_valencia", "ca@valencia" },
                 { "en_AU", "en_GB" },
                 { "en_NZ", "en_GB" },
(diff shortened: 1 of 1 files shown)


Jojo-Schmitz:
Some discussion at https://discord.com/channels/818804595450445834/818852942882275379/1068971663909204019 ff.

```
As a Canadian, most of us use more US English with the main exception being using the "our" spelling in words like colour, honour, and harbour.

Do you use "measure" or "bar" more often?

In school we used measure.

I'd personally say that it can be done by having a choice of either British or US versions. That's how computer spell check works here, and people choose which one they prefer so there is no Canadian one. At least not with most operating systems and apps I've looked at.
The only time Canada really has a different language is with the version of French used here, and possibly Gaelic as well, but that's a lot more niche.
```

Eism:
> I don't think we really want this, as per the feedback from native Canadiens

@Jojo-Schmitz Does this mean we should close this PR?

Jojo-Schmitz:
I think so, yes

PR #16133, opened 2023-01-30
[TESTING] stretch system to a page in a case system's width exceeds page's width

handrok:
When page width is not default, it's smaller. Or there are a lot of 32'th notes. System exceeds page:

Screenshot 2023-01-30 at 12 41 46

New:
Screenshot 2023-01-30 at 12 43 55

It's not perfect solution. For example notes climb on top of each other. It looks weird. But in most cases this compromise is fine.


oktophonie:
Closing, as this will be addressed by https://github.com/musescore/MuseScore/pull/15174

Issue #16042, opened 2023-01-24
[MU4 Task] New corrupted file dialogs


bkunda:
**Task description**
Two parts to this task: 

1. This new dialog would be triggered when the user attempts to save/publish a score to the cloud that was already corrupted upon opening: 

<img width="555" alt="Group 14189" src="https://user-images.githubusercontent.com/86290556/214262751-22db8ee7-91bb-4dfa-aff1-5390871a4e05.png">

2. These new dialogs would be triggered if the user tries to save a "clean" (i.e. non-corrupted) score, which has _become_ corrupted since it was opened: 

_Version where the user attempts a save to the cloud_
 <img width="603" alt="Group 14192" src="https://user-images.githubusercontent.com/86290556/214263355-09451b7f-6f2a-4a2b-977c-94c39514c3b4.png">

_Version where the user attempts a local save_
<img width="596" alt="Group 141932" src="https://user-images.githubusercontent.com/86290556/214263409-b69f9f52-b356-4f79-9942-3ef9f83e9fbc.png">

In both cases, "revert to last saved" triggers this confirmation dialog: 
<img width="527" alt="Group 14194" src="https://user-images.githubusercontent.com/86290556/214263490-28bfb61e-c3c0-4fb6-83a0-360f08752d4e.png">

Figma files have been sent directly to @Eism.
@shoogle @cbjeukendrup @Tantacrul FYI

Thanks!

PR #16151, opened 2023-01-31
[MU4] Check for corruption before attempting to save 

Eism:
Resolves: #16042

### DIFF
--- a/src/notation/view/internal/undoredomodel.cpp
+++ b/src/notation/view/internal/undoredomodel.cpp
@@ -67,6 +67,10 @@ void UndoRedoModel::load()
         });
     });
 
+    context()->currentProjectChanged().onNotify(this, [this]() {
+        emit stackChanged();
+    });
+
     emit stackChanged();
 }
 
--- a/src/project/inotationproject.h
+++ b/src/project/inotationproject.h
@@ -59,7 +59,7 @@ public:
     virtual void markAsUnsaved() = 0;
 
     virtual ValNt<bool> needSave() const = 0;
-    virtual bool canSave() const = 0;
+    virtual Ret canSave() const = 0;
 
     virtual Ret save(const io::path_t& path = io::path_t(), SaveMode saveMode = SaveMode::Save) = 0;
     virtual Ret writeToDevice(QIODevice* device) = 0;
--- a/src/project/internal/projectactionscontroller.cpp
+++ b/src/project/internal/projectactionscontroller.cpp
@@ -32,6 +32,7 @@
 #include "engraving/infrastructure/mscio.h"
 #include "engraving/engravingerrors.h"
 #include "cloud/clouderrors.h"
+#include "projecterrors.h"
 
 #include "log.h"
 
@@ -58,7 +59,7 @@ void ProjectActionsController::init()
     });
 
     dispatcher()->reg(this, "file-save", [this]() { saveProject(); });
-    dispatcher()->reg(this, "file-save-as", this, &ProjectActionsController::saveProjectAs);
+    dispatcher()->reg(this, "file-save-as", [this]() { saveProjectAs(); });
     dispatcher()->reg(this, "file-save-a-copy", this, &ProjectActionsController::saveProjectCopy);
     dispatcher()->reg(this, "file-save-selection", this, &ProjectActionsController::saveSelection);
     dispatcher()->reg(this, "file-save-to-cloud", this, &ProjectActionsController::saveToCloud);
@@ -378,41 +379,31 @@ IInteractive::Button ProjectActionsController::askAboutSavingScore(INotationProj
     return result.standardButton();
 }
 
-bool ProjectActionsController::canSaveProject() const
+Ret ProjectActionsController::canSaveProject() const
 {
     auto project = currentNotationProject();
     if (!project) {
         LOGW() << "no current project";
-        return false;
+        return make_ret(Err::NoProjectError);
     }
 
-    if (!project->canSave()) {
-        LOGW() << "project could not be saved";
-        return false;
-    }
-
-    return true;
+    return project->canSave();
 }
 
 bool ProjectActionsController::saveProject(const io::path_t& path)
 {
-    if (!canSaveProject()) {
-        warnSaveIsNotAvailable();
-        return false;
-    }
-
     auto project = currentNotationProject();
 
     if (!project->isNewlyCreated()) {
         if (project->isCloudProject()) {
-            return saveProjectToCloud(project->cloudInfo());
+            return saveProjectAt(SaveLocation(SaveLocationType::Cloud, project->cloudInfo()));
         }
 
-        return saveProjectLocally();
+        return saveProjectAt(SaveLocation(SaveLocationType::Local));
     }
 
     if (!path.empty()) {
-        return saveProjectLocally(path);
+        return saveProjectAt(SaveLocation(SaveLocationType::Local, path));
     }
 
     RetVal<SaveLocation> location = saveProjectScenario()->askSaveLocation(project, SaveMode::Save);
@@ -420,33 +411,23 @@ bool ProjectActionsController::saveProject(const io::path_t& path)
         return false;
     }
 
-    return saveProjectAt(location.val, SaveMode::Save);
+    return saveProjectAt(location.val);
 }
 
-void ProjectActionsController::saveProjectAs()
+void ProjectActionsController::saveProjectAs(bool force, SaveLocationType preselectedType)
 {
-    if (!canSaveProject()) {
-        warnSaveIsNotAvailable();
-        return;
-    }
-
     auto project = currentNotationProject();
 
-    RetVal<SaveLocation> location = saveProjectScenario()->askSaveLocation(project, SaveMode::SaveAs);
+    RetVal<SaveLocation> location = saveProjectScenario()->askSaveLocation(project, SaveMode::SaveAs, preselectedType);
     if (!location.ret) {
         return;
     }
 
-    saveProjectAt(location.val, SaveMode::SaveAs);
+    saveProjectAt(location.val, SaveMode::SaveAs, force);
 }
 
 void ProjectActionsController::saveProjectCopy()
 {
-    if (!canSaveProject()) {
-        warnSaveIsNotAvailable();
-        return;
-    }
-
     auto project = currentNotationProject();
 
     RetVal<SaveLocation> location = saveProjectScenario()->askSaveLocation(project, SaveMode::SaveCopy);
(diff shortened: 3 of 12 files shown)


src/engraving/libmscore/undo.cpp
```diff
@@ -292,11 +292,7 @@ UndoStack::UndoStack()
 
 UndoStack::~UndoStack()
 {
-    size_t idx = 0;
-    for (auto c : list) {
-        c->cleanup(idx++ < curIdx);
-    }
-    DeleteAll(list);
```
cbjeukendrup, line 299: Where has `DeleteAll(list);` gone in `UndoStack::clear()`? It looks like we would still need it to prevent memory leaks, don't we?

src/engraving/libmscore/cmd.cpp
```diff
@@ -331,6 +331,24 @@ void Score::undoRedo(bool undo, EditData* ed)
...
+{
+    if (readOnly()) {
+        return;
+    }
+
+    cmdState().reset();
+
+    auto undoStack = this->undoStack();
+    while (!undoStack->isClean()) {
+        undoStack->undo(ed);
```
cbjeukendrup, line 344: This seems to me a very risky way of reverting to the last saved state. If the score has become corrupted, I would trust the UndoStack even less. In some cases with corruptions or crashes that I've worked on, it looked indeed like undoing only made it worse. 
I thought we would just close and completely re-read the file from disk, to remove any risk.

src/project/internal/projectactionscontroller.cpp
```diff
@@ -378,58 +379,43 @@ IInteractive::Button ProjectActionsController::askAboutSavingScore(INotationProj
...
-        warnSaveIsNotAvailable();
-        return false;
-    }
-
     auto project = currentNotationProject();
 
     if (!project->isNewlyCreated()) {
         if (project->isCloudProject()) {
-            return saveProjectToCloud(project->cloudInfo());
+            return saveProjectAt(SaveLocation(SaveLocationType::Cloud, project->cloudInfo())); // todo force
```
RomanPudashkin, line 399: Is this TODO still needed?

src/project/internal/projectactionscontroller.cpp
```diff
@@ -938,15 +918,130 @@ void ProjectActionsController::warnPublishIsNotAvailable()
...
+    if (btn == saveAnywayBtn.btn) {
+        saveProjectAt(location, SaveMode::Save, true);
+    } else if (btn == revertToLastSavedBtn.btn) {
+        revertCorruptedScoreToLastSaved();
+    }
+}
+
+void ProjectActionsController::warnScoreCorruptionUponOpenning(const std::string& errorText)
+{
+    std::string title = trc("project", "Your score cannot be saved");
```
RomanPudashkin, line 1026: warnScore**CorruptionUponOpenning** but "Your score cannot be **saved**"

Wrong text?

src/engraving/libmscore/cmd.cpp
Eism: yep, fixed

RomanPudashkin (approved):

src/notation/view/internal/undoredomodel.cpp
```diff
@@ -67,6 +67,10 @@ void UndoRedoModel::load()
         });
     });
 
+    context()->currentProjectChanged().onNotify(this, [this]() {
+        load();
```
cbjeukendrup, line 71: I think we can simply call `emit stackChanged();` here. No need to re-subscribe to notifications, if I'm not mistaken.

src/project/internal/projectactionscontroller.cpp
```diff
@@ -938,15 +918,145 @@ void ProjectActionsController::warnPublishIsNotAvailable()
...
+    int btn = interactive()->error(title, body, errorText, buttons, defaultBtn).button();
+
+    if (btn == saveAnywayBtn.btn) {
+        saveProjectAt(location, SaveMode::Save, true);
+    } else if (btn == revertToLastSavedBtn.btn) {
+        revertCorruptedScoreToLastSaved();
+    }
+}
+
+void ProjectActionsController::warnCorruptedScoreUponOpenningCannotBeSaved(const std::string& errorText)
```
cbjeukendrup, line 1026: I'm not sure whether this method is following the design...
When the score _was already corrupted_ at the time that it was opened:
- saving locally should just work without any warning;
- saving/publishing to the cloud should be prohibited.

It looks like this method also prevents saving the score locally; and the title should be `Your score cannot be uploaded to the cloud` (and the "Cancel" button should be an "OK" button).

Excuse me if this is not true or if the design has been changed without that I know.

src/project/internal/projectactionscontroller.cpp
```diff
@@ -938,15 +918,145 @@ void ProjectActionsController::warnPublishIsNotAvailable()
...
+    IInteractive::ButtonDatas buttons;
+    buttons.push_back(interactive()->buttonData(IInteractive::Button::Cancel));
+
+    IInteractive::ButtonData saveCopyBtn(IInteractive::Button::CustomButton, trc("project", "Save a copy"), newlyCreated /*accent*/);
+    buttons.push_back(saveCopyBtn);
+
+    int defaultBtn = saveCopyBtn.btn;
+
+    IInteractive::ButtonData revertToLastSavedBtn(saveCopyBtn.btn + 1, trc("project", "Revert to last saved"),
+                                                  true /*accent*/);
```
cbjeukendrup, line 982: Aren't we forgetting `buttons.push_back(revertToLastSavedBtn);`?

src/project/inotationproject.h
```diff
@@ -58,8 +58,10 @@ class INotationProject
 
     virtual void markAsUnsaved() = 0;
 
+    virtual void revertChanges() = 0;
```
cbjeukendrup, line 61: It looks like we're not using this method because it's fully handled in `ProjectActionsController::revertCorruptedScoreToLastSaved`.

src/project/internal/projectactionscontroller.cpp
```diff
@@ -938,15 +918,145 @@ void ProjectActionsController::warnPublishIsNotAvailable()
...
+        interactive()->buttonData(IInteractive::Button::Cancel),
+        getHelpBtn
+    }, getHelpBtn.btn).button();
+
+    if (btn == getHelpBtn.btn) {
+        interactive()->openUrl(configuration()->supportForumUrl());
+    }
+}
+
+void ProjectActionsController::revertCorruptedScoreToLastSaved()
```
cbjeukendrup, line 1043: I'm missing this dialog from the design: 
Scherm­afbeelding 2023-02-01 om 13 24 58

cbjeukendrup:
For testing: a convenient deterministic way of corrupting a score is #16151. (Since #15935, the score is indeed detected as corrupted in that case.)

(@RomanPudashkin so for that purpose, it would be nice to wait before merging #16165 until this pull request is fully tested 🙂)

Eism:
> (@RomanPudashkin so for that purpose, it would be nice to wait before merging https://github.com/musescore/MuseScore/pull/16165 until this pull request is fully tested 🙂)

Yes, I used this case for testing. 
No need to block the merging of 16165, I will not rebase the master until the end of testing

bkunda:
@Eism tested on Win and macOS. Just a few things: 

1. When saving locally after a save-to-cloud attempt (using the ‘save a copy‘ button): MuseScore doesn't seem to recognise that the score has been saved, even though a file has actually been created on the computer [(See demo video)](https://www.dropbox.com/s/ty9xix9v8i88pf5/save-copy-issue.mov?dl=0).
2. When attempting to save/publish to cloud an _already corrupted score_, this dialog is inconsistent with the design: 

_Current implementation_
Screenshot 2023-02-06 at 12 04 49 pm

_Design_
214262751-22db8ee7-91bb-4dfa-aff1-5390871a4e05

3. A minor typo in this dialog (actually an error in my design):
Screenshot 2023-02-06 at 12 08 23 pm

The last sentence should read **This action cannot be undone** (I.e. "cannot" instead of "can not"). 

Thank you!

Eism:
@bkunda thanks!

1. the problem is also in master. @cbjeukendrup, your [commit](https://github.com/musescore/MuseScore/commit/de3457#diff-ba955a69c6d48fc8e1ad2625cdaf6abeffa16d1fd53774ebe5c57d7aabe069dfL370) is the cause of this bug. I'll leave it for you. @bkunda please create new issue
2. fixed. But the Ok button is an accent
3. fixed

cbjeukendrup:
Also, I think the "Save a copy" (or "Save as") button in the corruption dialog should immediately open the "save locally" dialog, because it does not make sense to suggest saving to the cloud if we already know that the score is corrupted.

bkunda:
> What you're expecting rather sounds like "Save as":

Yes on reflection I think this might actually make more sense. The behaviour of "Save a copy" will probably stump most users (it obviously stumped me!), and their main goal if they go for this option is most likely going to be to send their corrupted file to someone on the forum for assistance, so the behaviour should be as familiar as possible. 

So I guess this should probably be something like "Save to computer". If the file was already saved, it would perform a regular save function. If the file was not already saved, it would trigger the local "Save as" dialog. 

The UI copy would then be updated to something like: "You can also save the score locally and get help to fix the errors on musescore.org". 

> Also, I think the "Save a copy" (or "Save as") button in the corruption dialog should immediately open the "save locally" dialog, because it does not make sense to suggest saving to the cloud if we already know that the score is corrupted.

Completely agree with this 🙂

bkunda:
Hi @Eism,

So one change to the specification please: 

When the user attempts to save a "clean" (i.e. uncorrupted) score to the cloud _which has become corrupted since it was opened_, they see this dialog: 
 
Group 14192

The **Save As...** button would immediately trigger the regular operating system "save" dialog, _not_ MuseScore's "how would you like to save?" dialog.

Hope this is clear 🙂 Let me know when it's ready for further testing.

Eism:
@bkunda 
Fixed. Please test this build https://github.com/Eism/MuseScore/actions/runs/4143485042

(Builds in this PR don't start due to conflicts)

bkunda:
@Eism thanks! Just one more thing I've found (actually another gap in the design – my apologies).

For a new score that has _never been saved before_, and which becomes corrupted while the user is working: when the user attempts to save the score to the cloud for the first time, they currently see this dialog (in your build):

Screenshot 2023-02-10 at 4 24 56 pm

The options are right, but the messaging doesn't make sense (my oversight here). Kindly please change the body copy to: 

"This score has become corrupted and contains errors. You can fix the errors manually, or save the score to your computer and get help for this issue on musescore.org."

Thank you!

Eism:
@bkunda 
Fixed, but also rebased to the current master, so I'm not sure if you can make a score corrupted.

bkunda:
> @bkunda
> Fixed, but also rebased to the current master, so I'm not sure if you can make a score corrupted.

Yeah I just noticed this. No worries, I shall take your word that this last fix has been made 😀. Many thanks!

Issue #16125, opened 2023-01-29
[MU4 Issue] [Workspaces] UI glitch in workspaces window when adding too much workspaces


DmitryArefiev:
![Capture](https://user-images.githubusercontent.com/90187801/215351270-3000b99d-1c96-4b94-b1bc-43ebc63eb9a2.PNG)
![Capture2](https://user-images.githubusercontent.com/90187801/215351278-4b9f14f0-dbc8-47a5-860a-bbb62986fd42.PNG)

https://user-images.githubusercontent.com/90187801/215351289-7bd692f6-09b0-4e15-a153-e6f5e417f696.mp4

**Platform information**

 macOS, Windows, Linux

octopols:
I would like to work on this issue, can someone please assign it to me?

octopols:
I'll work on this as soon as my classes ends

PR #16154, opened 2023-01-31
Fix for issue #16125 | [MU4 Issue] [Workspaces] UI glitch in workspaces window when adding too much workspaces

octopols:
Resolves: #16125 

added `clip: true` property to `src\workspace\qml\MuseScore\Workspace\internal\WorkspacesView.qml` which overrides the `clip: false` property used in `StyledListView` in `src\framework\uicomponents\qml\MuseScore\UiComponents\RadioButtonGroup.qml`

### DIFF
--- a/src/workspace/qml/MuseScore/Workspace/WorkspacesDialog.qml
+++ b/src/workspace/qml/MuseScore/Workspace/WorkspacesDialog.qml
@@ -86,7 +86,7 @@ StyledDialogView {
         WorkspacesView {
             id: view
             Layout.fillWidth: true
-            Layout.fillHeight: true
+            Layout.preferredHeight: root.height - 190
             Layout.leftMargin: -parent.anchors.leftMargin
             Layout.rightMargin: -parent.anchors.rightMargin
             leftPadding: parent.anchors.leftMargin
--- a/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
+++ b/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
@@ -77,7 +77,7 @@ RadioButtonGroup {
 
         function onSelectedWorkspaceChanged(selectedWorkspace) {
             if (Boolean(selectedWorkspace)) {
-                root.positionViewAtIndex(selectedWorkspace.index + 1, ListView.Contain)
+                root.positionViewAtIndex(selectedWorkspace.index, ListView.Contain)
             }
         }
     }
--- a/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
+++ b/src/workspace/qml/MuseScore/Workspace/internal/WorkspacesView.qml
@@ -46,6 +46,7 @@ RadioButtonGroup {
     }
 
     spacing: 0
+    clip: true
     orientation: Qt.Vertical
 
     interactive: height < contentHeight
@@ -76,7 +77,7 @@ RadioButtonGroup {
 
         function onSelectedWorkspaceChanged(selectedWorkspace) {
             if (Boolean(selectedWorkspace)) {
-                root.positionViewAtIndex(selectedWorkspace.index, ListView.Contain)
+                root.positionViewAtIndex(selectedWorkspace.index + 1, ListView.Contain)
             }
         }
     }


octopols:
https://user-images.githubusercontent.com/90182257/215740229-e7d32755-08ee-4ea3-8210-9735cd97cad4.mp4

cbjeukendrup (approved):

Jojo-Schmitz:
Why? You apparently managed to get rid of the commit(s)?!?

cbjeukendrup:
Well... _all_ commits, including the one that should have been preserved ;-)

@octopols Maybe you could benefit from the info at https://musescore.org/en/handbook/developers-handbook/finding-your-way-around/git-workflow?

DmitryArefiev:
@octopols 
1. In Checks tab, there are more then one build for each OS. Can you remove remove unnecessary builds and leave only the last version (which will be tested) ?

2. When there are many workspaces and the active workspace is at the end of the list, the focus is too low. Is it easy to fix within the current PR? (or I can log a separate issue)


https://user-images.githubusercontent.com/90187801/217815417-26a56a9f-2ac8-4c11-a4cd-ceb0c2cade13.mp4

Thanks!

cbjeukendrup:
> In Checks tab, there are more then one build for each OS. Can you remove remove unnecessary builds and leave only the last version (which will be tested) ?

@octopols To do this, it probably helps if you just rebase and force-push. (It's likely a GitHub glitch anyway, not your fault)

octopols:
https://user-images.githubusercontent.com/90182257/218264191-186bf2f3-d2b2-4ea9-907f-fe29928fa03e.mp4

DmitryArefiev:
@octopols Yeah, it's better! Thanks! 

 But I found two cases when auto scrolling still doesn't work: 1. When select 'Default' workspace 2. Right after creating a new workspace

https://user-images.githubusercontent.com/90187801/218494045-9ed8d6c6-c9f2-49bc-a4f0-d27c092c2162.mp4

octopols:
I think I found what the issue is, the list auto scrolls to a position which is somehow hidden so what I did was I auto scrolled to a position bellow the index which was selected so the selected index was in view. This breaks when the last index is selected as the list cannot be scrolled bellow the last index? I'll see what I can do to fix this, will update soon.

octopols:
https://user-images.githubusercontent.com/90182257/218705535-9a3d6e1c-e2cb-431e-b240-3404e08a29d5.mp4

@DmitryArefiev I hope this solution works for all cases.

DmitryArefiev:
@octopols I tried the last build Windows and it **looks fine**. Thanks!

Can continue testing or do you want to fix CI_vtest?

DmitryArefiev:
> > Can continue testing or do you want to fix CI_vtest?
> 
> broken in the vtest in this PR is irrelevant, the problem on CI, so here they can be ignored

@igorkorsukov  Cool. Thanks!

@octopols I continue testing then

igorkorsukov:
@octopols please update with the master to fix the vtests

octopols:
@DmitryArefiev I updated the branch, vtests passes now.

DmitryArefiev:
> @DmitryArefiev I updated the branch, vtests passes now.

@octopols Cool. Thanks!

Tested #16125 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

src/workspace/qml/MuseScore/Workspace/WorkspacesDialog.qml
```diff
@@ -86,7 +86,7 @@ StyledDialogView {
         WorkspacesView {
             id: view
             Layout.fillWidth: true
-            Layout.fillHeight: true
+            Layout.preferredHeight: root.height - 190
```
Eism, line 89: @octopols 
This is not the right solution, it's bad to use magic numbers, it's not clear why this number, why not 200 :)

Positioning on the delegate works through positionViewAtIndex. The problem is that positioning doesn't correctly calculate delegate heights.
This is how qml works, need to give time to render the view and only then position on the selected delegate.

You need to call positioning delayed:
` Qt.callLater(root.positionViewAtIndex, selectedWorkspace.index, ListView.Contain)
`

Please prepare a PR

octopols: qt_callLater

Should I call it like this?

Eism: yes

octopols: I called it like in the image and reverted the changes in WorkspacesDialog.qml but `positionViewAtIndex` still scrolls it behind the WorkspaceBottomPannel somehow.


When I first tried to solve this problem I scrolled to an index bellow the index which was selected using `root.positionViewAtIndex, selectedWorkspace.index + 1, ListView.Contain` 

But this breaks when the last index is selected. 
I tried using `ListView.Center`, `positionViewAtEnd()` still no help, even tried enclosing them inside `onCountChanged` but it still scrolled behind the bottom pannel.

octopols: > qt_callLater
> 
> Should I call it like this?

I had clip enabled when I tested this, nothing else was changed

Issue #15718, opened 2023-01-05
[MU4 Issue] Middle-click drag does not work in note-input mode


tyzoid:
**Describe the bug**
Middle-click drag does not pan the score around while in note input mode, but does when note input is disabled.

**To Reproduce**
1. Select Note Input mode
2. Press and hold the Middle Mouse Button on the score
3. Drag the score around while holding the middle mouse button

**Expected behavior**
The score should drag around while in note input mode

**Screenshots**
[musescore-middle-drag-bug.webm](https://user-images.githubusercontent.com/974758/210705031-4c472a4f-b6bb-43cc-913a-957e69aa7c24.webm)

**Platform information**
 - OS: Arch Linux, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-, revision: github-musescore-musescore-

**Additional context**
N/A

PR #15037, opened 2022-12-08
fixed #12589 and #11656: Don't drag objects with middle button

Eism:
Resolves: #12589
Resolves: #11656

Was broken in #10147

PR #16162, opened 2023-01-31
fix: #15718 :Middle click will drag the score in note-input mode.

Tejkh1:
middle click will drag the score in note-input mode.

Resolves: #15718

I added the feature in which we can drag the score on middle-click in note-input mode.


Tejkh1:
> Is this not already fixed in master by #15037?

Sir, I dont know about that ,actually this is my first pull request.

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -722,6 +722,10 @@ void NotationViewInputController::mouseMoveEvent(QMouseEvent* event)
         m_view->moveCanvas(dragDelta.x(), dragDelta.y());
         m_isCanvasDragged = true;
     }
+    if (m_view->isNoteEnterMode()) {
+        m_view->moveCanvas(dragDelta.x(), dragDelta.y());
+        m_isCanvasDragged = true;
+    }
```
cbjeukendrup, line 728: Note that this piece of code can be simplified as just 
```c++
    m_view->moveCanvas(dragDelta.x(), dragDelta.y());
    m_isCanvasDragged = true;
```
without any if statements (since `isNoteEnterMode == m_view->isNoteEnterMode()`)

Also, in note input mode I think this should only happen for middle mouse button. Wouldn't it affect left mouse button too, with the current code?

PR #16172, opened 2023-02-01
Fix #341149: Note Names plugin not translated

Jojo-Schmitz:
Resolves: #16176 and https://musescore.org/en/node/341149

It also now (and as a prerequisite) makes the notenames from the Ambitus Inspector translatable, so does add a few strings to the translations


cbjeukendrup:
Quite a witty solution! But also a bit hacky...

@igorkorsukov This reminds me of another thing to keep in mind when designing a new plugins framework: we should ideally provide a proper way for plugins to have their own translations, so that they don't need to rely on hacks like this (and can also have strings that don't appear anywhere in MuseScore itself).

Jojo-Schmitz:
Yes, I admit it to be quite hacky...

cbjeukendrup:
But wait! All these note names also appear in AmbitusSettings.qml, but they're not translatable there. That would need to be fixed. If you would like to do that, that would be great; you could put them in the "global" context, and then reuse them in the note names plugin.

Jojo-Schmitz:
Ah!, let me look...

igorkorsukov:
> @igorkorsukov This reminds me of another thing to keep in mind when designing a new plugins framework: we should ideally provide a proper way for plugins to have their own translations, so that they don't need to rely on hacks like this (and can also have strings that don't appear anywhere in MuseScore itself).

Yes, I totally agree with you. Plugins should have their own translations, it's not difficult to do (I've already done this in another project :))

Jojo-Schmitz:
Done, made AmbitusSettings.qml translatable and nor fully use those in notenames.qml

Jojo-Schmitz:
> Yes, I totally agree with you. Plugins should have their own translations, it's not difficult to do (I've already done this in another project :))

@igorkorsukov: I'm perfectly aware of that, the problem though with the builtin plugins is to get their strings translated on Transifex, along with MuseScore's other strings

src/inspector/view/qml/MuseScore/Inspector/notation/ambituses/AmbitusSettings.qml
```diff
@@ -42,41 +42,41 @@ Column {
...
-        { text: "A♭", value: AmbitusTypes.TPC_A_B },
-        { text: "A", value: AmbitusTypes.TPC_A },
-        { text: "A♯", value: AmbitusTypes.TPC_A_S },
-        { text: "A♯♯", value: AmbitusTypes.TPC_A_SS },
-        { text: "B♭♭", value: AmbitusTypes.TPC_B_BB },
-        { text: "B♭", value: AmbitusTypes.TPC_B_B },
-        { text: "B", value: AmbitusTypes.TPC_B },
-        { text: "B♯", value: AmbitusTypes.TPC_B_S },
-        { text: "B♯♯", value: AmbitusTypes.TPC_B_SS }
+        { text: qsTrc("global", "C♭♭"), value: AmbitusTypes.TPC_C_BB },
```
Jojo-Schmitz, line 45: Using 'global' as the context here rather than 'inspector', as a good part of these strigs (C, C♯, D♭, D, D♯, E♭, E, F, F♯, G, G♯, A, A♯, B♭, and B) are available in the global context already.

igorkorsukov:
> @igorkorsukov: I'm perfectly aware of that, the problem though with the builtin plugins is to get their strings translated on Transifex, along with MuseScore's other strings

I think we can at the moment just add a pass  lupdate through the plugins folder with the same tc file
https://github.com/musescore/MuseScore/blob/master/tools/translations/run_lupdate.sh#L48

Jojo-Schmitz:
@cbjeukendrup mind to review and merge?

share/plugins/note_names/notenames.qml
```diff
@@ -97,32 +100,32 @@ MuseScore {
...
-               case 23: text.text = qsTranslate("accidental", "Natural arrow up") + text.text; break;
-               case 24: text.text = qsTranslate("accidental", "Natural arrow both") + text.text; break;
-               case 25: text.text = qsTranslate("accidental", "Sori") + text.text; break;
-               case 26: text.text = qsTranslate("accidental", "Koron") + text.text; break;
+               case  1: text.text = qsTranslate("engraving/sym", "Sharp") + text.text; break;
+               case  2: text.text = qsTranslate("engraving/sym", "Flat") + text.text; break;
+               case  3: text.text = qsTranslate("engraving/sym", "Double sharp") + text.text; break;
+               case  4: text.text = qsTranslate("engraving/sym", "Double flat") + text.text; break;
+               case  5: text.text = qsTranslate("engraving/sym", "Natural") + text.text; break;
+               case  6: text.text = qsTranslate("engraving/sym", "Flat-slash") + text.text; break;
```
cbjeukendrup, line 108: It looks like this list is outdated (when compared to accidental.cpp), and not all these names appear in symnames.cpp. But since it's hidden behind an `if (false)` anyway, it's maybe not worth the tourble to fix that... that do you think?

Jojo-Schmitz: Let's do away with them, seems a remnant from MuseScore 1

Jojo-Schmitz:
Could this please finally get merged?

cbjeukendrup (approved):

cbjeukendrup:
The reason I hadn't merged it yet was probably that I was thinking about making `lupdate` go through plugins too. But actually that doesn't prevent us in any way to merge this PR right now...

Jojo-Schmitz:
Thanks!

Issue #16174, opened 2023-02-01
[MU4 Issue] Crash after select all and delete all measures (one score)


DmitryArefiev:
**To Reproduce**
1. Open [segno  crash.mscz.TXT](https://github.com/musescore/MuseScore/files/10556927/segno.crash.mscz.TXT)
2. Select all
3. Delete measures

https://user-images.githubusercontent.com/90187801/216042162-d9d21beb-578a-425a-8ee4-45a895720deb.mp4

**Platform information**
macOS, Windows, Linux

**Additional context**
I **can't reproduce** it from scratch.. only on the attached score
**MS4.0.0** also crashes with that score

cbjeukendrup:
@DmitryArefiev It turns out that the parts of this score are corrupted. (This is not detected yet, see #15988.)

I'm not sure to what extent we are supposed to fix crashes that are caused by corrupted scores.

cbjeukendrup:
Anyway, I've created this fix: #16175 
It will let the user delete the measures without crashing, which could be a good way to get rid of the corruptions in certain cases.

DmitryArefiev:
Fixed in #16175

DmitryArefiev:
Retested on Win10

PR #16175, opened 2023-02-01
Don't crash when deleting measures in corrupted score

cbjeukendrup:
to help the user to fix corruptions by deleting the affected measures.

Resolves: #16174


src/engraving/libmscore/edit.cpp
```diff
@@ -6327,9 +6327,16 @@ void Score::undoRemoveMeasures(Measure* m1, Measure* m2, bool preserveTies)
     //  handle ties which start before m1 and end in (m1-m2)
     //
     for (Segment* s = m1->first(); s != m2->last(); s = s->next1()) {
+        IF_ASSERT_FAILED(s) {
+            // This score is corrupted; handle it without crashing,
```
RomanPudashkin, line 6331: It would be better to output this message to LOGE

RomanPudashkin (approved):

DmitryArefiev:
Tested #16174 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #16006, opened 2023-01-22
[MU4 Issue] Nonsensical behaviour of offsets when changing stave size


oktophonie:
**Describe the bug**

Here's a simple example from a score (string quartet template). The rests, accent and stave text all have X and Y offsets applied:
![image](https://user-images.githubusercontent.com/20604032/213913127-68d9faee-607b-48c5-8f1a-b971b3a3d24d.png)

Here is the Violin I part:
![image](https://user-images.githubusercontent.com/20604032/213913226-46acf00a-7f48-4095-9cac-cd61898e5a8e.png)

Note that:
- X and Y offsets to text are reset
- X and Y offsets to rests and articulation are SCALED to the new stave size (score size/part size, in this case 1.3/1.75, so scaled by a factor of 0.743)

The former I can accept, at least for now (the position of text may need to be changed anyway, with possibly different text sizes and radically different layout etc) but the latter makes NO sense. The offsets of items directly attached to notes, or to rests (which need to have specific positions on the stave) should not change!

I haven't tested yet _all_ types of item to see how their behaviour may differ.

It turns out this isn't about generating parts per se: it happens whenever the spatium size is changed (which will often happen when parts are generated).

**Platform information**
This is in MU4, but the behaviour is inherited from 3.6 or earlier.

PR #16180, opened 2023-02-01
Avoid scaling offset twice when spatium is changed

mike-spa:
Resolves: #16006

It turns out the that issue happened because Score::spatiumChanged() was being inadvertently called twice when changing the staff size in a piece. This should fix it.

### DIFF
--- a/src/engraving/tests/parts_data/part-all-appendmeasures.mscx
+++ b/src/engraving/tests/parts_data/part-all-appendmeasures.mscx
@@ -1267,7 +1267,7 @@
                 <Fingering>
                   <linked>
                     </linked>
-                  <offset x="1.06662" y="-4.71271"/>
+                  <offset x="1.05809" y="-4.67501"/>
                   <text>2</text>
                   </Fingering>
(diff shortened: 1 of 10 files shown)


oktophonie (approved):

src/engraving/tests/parts_data/part-all-appendmeasures.mscx
```diff
@@ -1267,7 +1267,7 @@
                 <Fingering>
                   <linked>
                     </linked>
-                  <offset x="1.06662" y="-4.71271"/>
+                  <offset x="1.05809" y="-4.67501"/>
```
RomanPudashkin, line 1270: Why did you change the test data?

mike-spa: Ok I've double checked to be sure. In all of these tests, the spatium in the master score is slightly different than the spatium in the part, so all of these tests had slightly wrong offsets because of the wrong scaling calculations. Now all the new offset values should be correct.

Issue #14338, opened 2022-11-05
[MU4 Issue] Missclicking while selecting several things, although still holding control shouldn't remove all the selections.


Grim-O:
When you make a selection (holding the control button) and then try to select something else but accidentally click on empty space, the selection is cleared. 

It would be awesome if the selection is kept because it often happens that I'm selecting things quickly and misclick or select tiny things such as staccato markings and also misclick.

MarcSabatella:
Indeed, and see also https://github.com/musescore/MuseScore/issues/10225.  These are two related features that were implemented together (same code) in MuseScore 3 but seem to have been lost in MuseScore 4.

cbjeukendrup:
Might be a good first issue; hint: `NotationViewInputController`.

MarcSabatella:
I assume you mean not clearing the selection would be easy first issue.  I think reimplementing the select-element-underneath behavior is rather trickier.

octopols:
> Might be a good first issue; hint: `NotationViewInputController`.

I removed the condition which called clearselection in that file which I think fixes the issue but after using it for a while the experience feels quite odd, like whenever I want to get rid of selections I have to click somewhere outside the canvas or use my arrow keys.

Grim-O:
the way I think it should work (and how I've seen it work in other programs) is that when you hold control, the selection isn't cleared regardless of where you click and when you don't hold control the selection is cleared when clicking anywhere. If you click on another element (possibly part of the larger selection) without control, only that one element should be selected.

octopols:
> the way I think it should work (and how I've seen it work in other programs) is that when you hold control, the selection isn't cleared regardless of where you click and when you don't hold control the selection is cleared when clicking anywhere. If you click on another element (possibly part of the larger selection) without control, only that one element should be selected.

That makes sense, let me look into the code to see if I can make that work.

PR #16190, opened 2023-02-02
Possible fix for Issue #14338: [MU4 Issue] Missclicking while selecting several things, although still holding control shouldn't remove all the selections.

octopols:
Resolves: #14338 

Nested the condition which clears selection inside a condition which checks for control modifier.

### DIFF
--- a/src/notation/view/notationviewinputcontroller.cpp
+++ b/src/notation/view/notationviewinputcontroller.cpp
@@ -790,9 +790,11 @@ void NotationViewInputController::mouseReleaseEvent(QMouseEvent* event)
     INotationNoteInputPtr noteInput = interaction->noteInput();
     const EngravingItem* hitElement = hitElementContext().element;
 
-    if (!hitElement && !m_isCanvasDragged && !interaction->isGripEditStarted()
-        && !interaction->isDragStarted() && !noteInput->isNoteInputMode()) {
-        interaction->clearSelection();
+    if (event->modifiers() != Qt::ControlModifier && event->modifiers() != Qt::ShiftModifier) {
+        if (!hitElement && !m_isCanvasDragged && !interaction->isGripEditStarted()
+            && !interaction->isDragStarted() && !noteInput->isNoteInputMode()) {
+            interaction->clearSelection();
+        }
     }
 
     if (event->button() == Qt::LeftButton && event->modifiers() == Qt::NoModifier) {


zacjansheski:
@octopols This is almost a half year old now :)

Is this working and ready for review?

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -749,9 +749,11 @@ void NotationViewInputController::mouseReleaseEvent(QMouseEvent* event)
     INotationNoteInputPtr noteInput = interaction->noteInput();
     const EngravingItem* hitElement = hitElementContext().element;
 
-    if (!hitElement && !m_isCanvasDragged && !interaction->isGripEditStarted()
-        && !interaction->isDragStarted() && !noteInput->isNoteInputMode()) {
-        interaction->clearSelection();
+    if (event->modifiers() != Qt::ControlModifier) {
+        if (!hitElement && !m_isCanvasDragged && !interaction->isGripEditStarted()
+            && !interaction->isDragStarted() && !noteInput->isNoteInputMode()) {
```
cbjeukendrup, line 754: I think we can combine this into one `if` condition using `&&`. Also, maybe we should do the same if the user has pressed Shift instead of Ctrl.

cbjeukendrup:
It's rather our fault that it has become so old... sorry @octopols! 

It can't be tested right now because the test builds have expired. But if you rebase and make the suggested amendments, and then force-push, then new builds will be created.

src/notation/view/notationviewinputcontroller.cpp
```diff
@@ -790,11 +790,10 @@ void NotationViewInputController::mouseReleaseEvent(QMouseEvent* event)
     INotationNoteInputPtr noteInput = interaction->noteInput();
     const EngravingItem* hitElement = hitElementContext().element;
 
-    if (event->modifiers() != Qt::ControlModifier) {
-        if (!hitElement && !m_isCanvasDragged && !interaction->isGripEditStarted()
-            && !interaction->isDragStarted() && !noteInput->isNoteInputMode()) {
-            interaction->clearSelection();
-        }
+    if (!(event->modifiers() & Qt::ControlModifier)
```
cbjeukendrup, line 793: Sorry, my comment was unclear: I meant `if (event->modifiers != Qt::ControlModifier && event->modifiers() != Qt::ShiftModifier && …)`.

Also, on second thought, I think your solution with two (nested) `if` statements was still better.

cbjeukendrup (approved):
Thanks!

zacjansheski:
@cbjeukendrup @octopols 

Looks like this needs to be rebased and force-pushed

cbjeukendrup:
> @cbjeukendrup @octopols 
> 
> 
> 
> Looks like this needs to be rebased and force-pushed

... because the VTests keep failing for an unclear reason.

octopols:
@zacjansheski I have rebased the branch the builds should be ready.

zacjansheski:
Tested Mac OS 11.6.5, Ubuntu 20.04.6, Windows 11
https://github.com/musescore/MuseScore/issues/14338 FIXED

Issue #13652, opened 2022-10-07
[MU4 Issue] Regression: opacity control missing in color picker


MarcSabatella:
**Describe the bug**
MuseScore 4 lacks the opacity control in the color picker - needed for creating highlighter effects

**To Reproduce**
1. Add a staff text
2. Set the Frame to rectangle in Properties
3. Click the color swatch to set the flll color
4. See error: no opacity control

**Expected behavior**
An opacity control is needed to create translucent effects, such as mimicking a yellow highlighter pen

**Screenshots**
This is how it looked in MuseScore 3, and shows how the feature was commonly used by educators:
![image](https://user-images.githubusercontent.com/1799936/194440528-9a4a0ebb-dc2a-4e42-a88e-3b197a0bb46e.png)

**Platform information**
 - OS: Linux

**Additional context**
In MU3, it was very annoyingly necessary to explicitly set this to 255 if you wanted to actually see the background color at all - the default was 0.  Glad to see that opening this dialog now sets the color to opaque.  But, there still needs to be a way to set it to intermediate values like 127.  And, a small complication - that value, once explicitly set, should remain set and not get arbitrarily changed to 255 every time you open the dialog.  So, whatever magic happens to force the opacity to 255 upon setting a color, should be smart enough not to do this if there is an explicit opacity value already set.

rgreen5:
Bump.

octopols:
I want to work on this, can you please assign me this issue?

octopols:
I was able to add the alpha channel in the colour picker but the task also said that the solution should be smart enough to realise when alpha is 0 and set it to max (atleast that's how I interpreted the problem). 
I was able to compare QColors but is there a way by which I can change the alpha value of an already existing QColor?

cbjeukendrup:
How about https://doc.qt.io/qt-5/qcolor.html#setAlpha?

octopols:
https://user-images.githubusercontent.com/90182257/216559541-6f790556-3ae1-46ac-89c7-a281e85d9625.mp4

Jojo-Schmitz:
@octopols what's the status of this?

Jojo-Schmitz:
Came up again in https://musescore.org/en/node/371565

Jojo-Schmitz:
And again in https://musescore.org/en/node/377166

Kartoffelbreili:
Id like to use it to be able to do the highlighting too. Just like how it is described here: https://musescore.org/en/node/220071

PR #16103, opened 2023-01-27
#13652 [MU4 Issue] Regression: opacity control missing in color picker

octopols:
Resolves: #13652

Color picker now have alpha value. If the alpha value is 0 and the color picker is launched the value is changed to 255 automatically.

- [x] I signed the [CLA](https://musescore.org/en/cla)
- [x] The title of the PR describes the problem it addresses
- [x] Each commit's message describes its purpose and effects, and references the issue it resolves
- [x] If changes are extensive, there is a sequence of easily reviewable commits
- [x] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [ ] There are no unnecessary changes
- [x] The code compiles and runs on my machine, preferably after each commit individually
- [ ] I created a unit test or vtest to verify the changes I made (if applicable)

PR #16201, opened 2023-02-03
Fix for Issue #13652 MU4 Issue] Regression: opacity control missing in color picker

octopols:
Resolves: #13652 
Color picker now have alpha value. If the alpha value is 0 and the color picker is launched the value is changed to 255 automatically.

Sorry I had to close the last PR fixing this issue because of me deleting my original fork. 
This PR is a replacement for #16103 .

### DIFF
--- a/src/framework/global/internal/interactive.cpp
+++ b/src/framework/global/internal/interactive.cpp
@@ -225,7 +225,11 @@ io::paths_t Interactive::selectMultipleDirectories(const QString& title, const i
 
 QColor Interactive::selectColor(const QColor& color, const QString& title)
 {
-    QColor selectedColor = QColorDialog::getColor(color, nullptr, title);
+    QColor colorEditable = color;
+    if (colorEditable.alpha() == 0) {
+        colorEditable.setAlpha(255);
+    }
+    QColor selectedColor = QColorDialog::getColor(colorEditable, nullptr, title, QColorDialog::ShowAlphaChannel);
     return selectedColor.isValid() ? selectedColor : color;
 }
 


octopols:
https://user-images.githubusercontent.com/90182257/216559433-f9f2603f-2cfa-4588-8d1a-76b259f2b01c.mp4

octopols:
Is there a way for me to add a checkbox to QColorDialog to enable or disable checkbox? When I first tackled this issue I couldn't find the qml for QColorDialog and soon realized it's an inbuit component and never really looked into editing the dialog's UI.

cbjeukendrup:
@octopols What do you mean? The color dialog is platform-dependent, i.e. Qt just calls the OS's collar dialog. That means that you don't have any possibilities to customise it, beyond the options offered by Qt.

cbjeukendrup (commented):
I see one issue with this PR: probably not _every_ color picker should have an alpha channel. For example, I think the color pickers in Preferences > Appearance should not have an alpha channel, because that may cause strange results. 

So I think you need to add a property to the `ColorPicker` component in QML to enable/disable the alpha channel per instance of `ColorPicker`. Then, you need to propagate the value of that property via ColorPickerModel to `Interactive::selectColor`, which should get a new parameter for that. Then, based on that parameter, you should pass `QColorDialog::ShowAlphaChannel` or `QColorDialog::ColorDialogOptions()` to the QColorDialog.

octopols:
The whole feel

> 

I was discussing this issue with @VanSHOE recently, and he suggested editing it through Qt Quick components. However, I'm not entirely sure if I conveyed my problem correctly to him.
I will implement the fix you mentioned. However, if we do decide to add a checkbox to enable the alpha channel, is our only option to design a completely new component?

cbjeukendrup:
I'm afraid yes...

octopols:
@cbjeukendrup I'm having a really hard time propagating a bool I created `showAlpha` to `Interactive::selectColor`. I think I need to make some changes in `src\framework\uicomponents\view\colorpickermodel.cpp` but I don't even know where to start. Should I commit the changes I made so far?

octopols:
Screenshot 2023-03-05 003235

cbjeukendrup:
That list of files looks good. The "missing link" will be in `colorpickermodel.h/cpp`. I think the easiest way is to add a `showAlpha` bool parameter the method `Q_INVOKABLE QColor selectColor(const QColor& currentColor);`.

octopols:
I made some changes in the `colorpickermodel.h/cpp` files but I'm still getting an error in the `ColorPicker.qml` file.
ColorPicker qml

These are the changes I made in `colorpickermodel.h`
colorpickermodel h

And these are the changes I made in `colorpickermodel.cpp`
colorpickermodel cpp

cbjeukendrup:
Ah, the problem is that `showAlpha` is not a property of the ColorPickerModel class, but a parameter of the selectColor method. 

That method is called here: 
```qml
    QtObject {
        id: prv

        function selectColor() {
            var selectedColor = colorPickerModel.selectColor(root.color) // <-- here
            root.newColorSelected(selectedColor)
        }
    }
```

So in that place, you need to pass the showAlpha parameter.

octopols:
I have this one last problem (hopefully), how do I set the value of `showAlpha` from `TextSettings.qml`.
how_to_set

cbjeukendrup:
There are two solutions:
1. setting the property of the color picker inside ColorSection.qml (but that means that it will be enabled for all color pickers that are created using ColorSection; not sure if that's desired
2. create a `showAlpha` property alias in ColorSection.qml: inside the root item of the file, you would add `property alias showAlpha: colorPicker.showAlpha`.
After giving ColorSection this property, you can go to TextSettings.qml and use it like just `showAlpha: true`.

octopols:
Thank you very much for your help. The issue has been resolved now.

MarcSabatella:
Sorry if this was already obvious to everyone else, but in case it's useful or relevant - I just realized, the alpha channel control already is present *if* you involve the color picker from the Format / Style / Text Styles.  Which is great, because that's where it is actually needed most.  But of course, it would be great to have in Properties as well

Jojo-Schmitz:
Rebase needed

octopols:
> Rebase needed 

I'll rebase it today, sorry for replying late

cbjeukendrup:
@zacjansheski please check that the colour pickers in the properties panel now have a (working) alpha channel, i.e. the one inside the Appearance popup, and the ones for text border / fill.

zacjansheski:
I'm unable to open these builds, tried on both MacOS and Windows.  Splash screen and then MSS crashes.

@cbjeukendrup please advise

cbjeukendrup:
It turns out I had not properly removed a change from an old version of this PR... should be fixed now!

zacjansheski:
Tested on MacOS 15, Windows 11. Approved
https://github.com/musescore/MuseScore/issues/13652 FIXED

Jojo-Schmitz:
Backporting to 4.6.4?

cbjeukendrup:
Hm... yeah, I'm not against.

Jojo-Schmitz:
Done, see #30918

Issue #10664, opened 2022-02-24
Input cursor on tablature (TAB) staves is smaller than in MU3


DmitryArefiev:
**Describe the bug**
Input cursor on tablature (TAB) staves is smaller than in MU3 

![Capture](https://user-images.githubusercontent.com/90187801/155569858-29724920-185f-4927-88db-167f3c798489.PNG)

**Expected behavior**
Cursor should be the same as in MU3

**Platform information**
macOS, Windows, Linux

jatinAgarwala:
Hello! I would like to try working on this issue if it's still open

RobFog:
It’s all yours. 😊

Tantacrul:
Any movement on this?

somethingstarted:
i'm looking for it but I can't find it. been peaking through the code trying to familiarize myself with musescore last few days. 
what is the highlighter called in the program? and what would be a good file that i can start my search in? 

i've been looking in 
\src\engraving\internal\engravingconfiguration.cpp and 
\src\engraving\libmscore\fret.cpp

but i haven't found the blue highlighter yet. i was thinking about making an IF fret number is 1 leave width as is and if it's 2 characters (like in the example) make it x percent larger or whatever but i wont know if that's a good idea or not till i find where it's at and am able to figure out how to preview/compile  programs with cmakes (i've only done VS solutions so far) 

gonna watch some tutorials on previewing/compiling cmakes and i think once i figure that out, it'll make this much easier and faster for me

Tantacrul:
Maybe @cbjeukendrup might be able to help?

cbjeukendrup:
@somethingstarted In MuseScore 4 (master branch), the code for the note input cursor is in `src/notation/view/noteinputcursor.cpp` (and the methods that are called from there). 

You might want to compare it with MuseScore 3 (3.x branch), where the code seems to be in `PositionCursor::paint`.

Tip: I wouldn't recommend to compile the 3.x branch, but just download the ZIP from GitHub and open it in Visual Studio Code, which is enough to browse the code. I'd also keep it in a separate folder, because when you switch between 3.x and master, you need to delete your build folder so that everything needs to be re-compiled, which takes long, so therefore I don't recommend it. 

Compiling in Qt Creator might be a bit easier and is documented here: https://github.com/musescore/MuseScore/wiki/Compile-in-Qt-Creator

[...]

PR #16107, opened 2023-01-28
#10664 fixing  variable names - and making the blue box wider, with a quick way of further editing of it's side in the future. 

somethingstarted:
Resolves: #10664 

-fixing  variable names to fit camel case guidelines - 
-making the blue box wider, with a quick way of further editing of it's side in the future. 

- [ x] I signed the [CLA](https://musescore.org/en/cla)
- [ x] The title of the PR describes the problem it addresses
- [ x] Each commit's message describes its purpose and effects, and references the issue it resolves
- [ x] If changes are extensive, there is a sequence of easily reviewable commits
- [ x] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [ x] There are no unnecessary changes
- [ x] The code compiles and runs on my machine, preferably after each commit individually
- [ x] I created a unit test or vtest to verify the changes I made (if applicable) n/a but it did run when compiled via QT

PR #16209, opened 2023-02-04
#10664 resizing + recentering curorRect

somethingstarted:
Resolves: #10664 

Get current width, get location, multiply current width by mult amount. recenter 

(needs some math sense it's left justified, so it was a little convoluted converting from center coordinate to edge, but otherwise it'd be off center with the new width)

then once it calculates a new left justified location, it puts it there. 

in reference to (now closed) pulls reference  #16107 


src/notation/view/noteinputcursor.cpp
```diff
@@ -34,12 +34,21 @@ void NoteInputCursor::paint(mu::draw::Painter* painter)
     NoteInputState state = noteInput->state();
     RectF cursorRect = noteInput->cursorRect();
 
+    constexpr auto widthMultiplier = 1.3;
+    constexpr int leftLineWidth = 3;
+
+    // calculate where the box is, and how to center it with new width
+    auto oldWidth = cursorRect.width();   // store old width
+    auto newWidth = cursorRect.width() * widthMultiplier; // set new width
+    auto pixToSlideLeft = (((oldWidth - newWidth) / 2) - (leftLineWidth / 2)); // calculate new center pos
```
Eism, line 43: All cursor position calculations must be in the NotationNoteInput::cursorRect method, NoteInputCursor only paint.

As for the problem, it seems to me that it is enough for us to do the following in NotationNoteInput::cursorRect 
```
    const mu::engraving::StaffType* staffType = staff->staffType(inputState.tick());
    bool isTabStaff = staffType->isTabStaff();

    int sideMargin = isTabStaff ? 16 : 4;
```

The cursor may look not so elegant, but I would not like to complicate the code, as it will be difficult to maintain.
(But if the designers or testers do not like the new solution, then we will think about your solution)

Thanks

somethingstarted: not a problem. I'm here to learn. I can see how yours makes more sense. plus yours fixes it at the source and mine was more of a bandaid.

PR #14822, opened 2022-11-29
fixed #14685: Crash when opening one-measure score with 2nd ending assigned

Eism:
Resolves: #14685

PR #16213, opened 2023-02-04
Fix and re-enable the repeats tests (except two)

cbjeukendrup:
The remaining two are 57 and 67; those were broken by #14822 (but at that time the repeats tests were not yet run at all). @jeetee will have a look to fix that. 

(We're working together here at FOSDEM 😄)

### DIFF
--- a/src/engraving/tests/utils/scorerw.cpp
+++ b/src/engraving/tests/utils/scorerw.cpp
@@ -77,6 +77,15 @@ MasterScore* ScoreRW::readScore(const String& name, bool isAbsolutePath, ImportF
         }
     }
 
+    // While reading the score, some elements might use `score->repeatList()` (which is incorrect
+    // anyway, because the repeatList will be incomplete because the score is incomplete, but some
+    // elements still do it).
+    // `score->repeatList()` calls `_repeatList->update()`; the repeat list then thinks that it is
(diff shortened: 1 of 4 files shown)


src/engraving/tests/utils/scorerw.cpp
```diff
@@ -77,6 +77,15 @@ MasterScore* ScoreRW::readScore(const String& name, bool isAbsolutePath, ImportF
...
     }
 
+    // While reading the score, some elements might use `score->repeatList()` (which is incorrect
+    // anyway, because the repeatList will be incomplete because the score is incomplete, but some
+    // elements still do it).
+    // `score->repeatList()` calls `_repeatList->update()`; the repeat list then thinks that it is
+    // up-to-date from that point. But we weren't finished reading the score, so the score will still
+    // change. We need to tell the repeat list about that, so that it will be updated next time
+    // someone uses it.
+    score->setPlaylistDirty();
```
cbjeukendrup, line 87: This change is not strictly necessary anymore since the changes in spanner.cpp/h; I'd still like to have it here, just in case something else happens to use the repeat list while reading the score, and because it is in EngravingPRoject::doSetupMasterScore too.

igorkorsukov (approved):

PR #16222, opened 2023-02-05
applied solo mute mixer settings from Musescore 3 version and below.

handrok:
In Musescore 4 mixer settings store in audiosettings.json. Previous it was part of mscx file. It stored in Instrument section of mscx file. Now this setting (_solo, _mute) is not used anywhere. Added applying compat mixer settings and applyed it to the mixer.


src/engraving/libmscore/instrument.cpp
```diff
@@ -988,6 +940,10 @@ void InstrChannel::read(XmlReader& e, Part* part)
     }
 
     _mustUpdateInit = true;
+    if (e.context()
+        && e.context()->audioSettingCompat().count(partAudioSetting.instrumentId.partId) == 0) {
+        e.context()->addPartAudioSettingCompat(partAudioSetting);
```
igorkorsukov, line 945: I think it's better to hide this logic with a check that the data is already there in the `addPartAudioSettingCompat` method, we can call it `addUniquePartAudioSettingCompat`

src/engraving/rw/scorereader.cpp
```diff
@@ -143,6 +144,8 @@ mu::Ret ScoreReader::loadMscz(MasterScore* masterScore, const MscReader& mscRead
         }
     }
 
+    audioSettings= std::move(masterScoreCtx.audioSettingCompat());
```
igorkorsukov, line 147: It might be better to make `masterScoreCtx`  as a in-out parameter and set it outside. This will allow us to do other things in the same way, and now if we also need to read something else, we will either have to add another meter or redo this solution...

src/engraving/libmscore/instrument.cpp
handrok: fixed

src/engraving/rw/scorereader.cpp
igorkorsukov: This is just the current implementation, we can change it, for example, replace it with the use the shared pointer

handrok: AudioSettingsCompat is an out parameter of loadMscz function. To add other things is simple. Extending AudioSettingsCompat is needed only. For example volume and pan of instrument are easily can be added. I think AudioSettingsCompat is readable parametr of loadMscz function instead of ReadContext. ReadContext looks like a useful staff of XmlReader.  Adding it to  loadMscz function seems make this function less understood I guess.

handrok: fixed

igorkorsukov (approved):

src/engraving/types/types.h
```diff
@@ -908,6 +908,17 @@ static inline bool operator==(const Key a, const Key b) { return int(a) == int(b
 static inline bool operator!=(const Key a, const Key b) { return static_cast<int>(a) != static_cast<int>(b); }
 static inline Key operator+=(Key& a, const Key& b) { return a = Key(static_cast<int>(a) + static_cast<int>(b)); }
 static inline Key operator-=(Key& a, const Key& b) { return a = Key(static_cast<int>(a) - static_cast<int>(b)); }
+
+struct PartAudioSettingsCompat {
+    InstrumentTrackId instrumentId;
+    bool mute{ false };
+    bool solo{ false };
+    int velocity{ 127 };
```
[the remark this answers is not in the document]
handrok: fixed

PR #16223, opened 2023-02-06
Beam / tremolo layout refactor

asattely:
This is a work-in-progress PR.

The idea of this PR is to factor out our fancy new beam positioning algorithm into its own class, which can be used by both beams and two-note tremolos. This doesn't change anything about beams, but rather radically changes two-note trems' behavior to use the beam layout algorithm.


igorkorsukov:
@asattely please update with the master to fix the vtests

src/engraving/libmscore/beam.h
```diff
@@ -33,18 +33,105 @@ class ChordRest;
...
     bool above; // above level 0 or below? (meaningless for level 0)
 };
 
+struct TremAnchor {
+    ChordRest* chord1;
+    double y1;
+    double y2;
+};
+
+class BeamLayout
```
RomanPudashkin, line 64: I would recommend moving this class to a separate file inside the engraving/layout folder

src/engraving/libmscore/beam.h
```diff
@@ -33,18 +33,105 @@ class ChordRest;
...
+{
+private:
+    static constexpr std::array _maxSlopes = { 0, 1, 2, 3, 4, 5, 6, 7 };
+    enum class BeamType {
+        INVALID,
+        BEAM,
+        TREMOLO
+    };
+    BeamType _beamType{ BeamType::INVALID };
+    EngravingItem* _e{ nullptr };
```
RomanPudashkin, line 74: Please use the = operator instead

src/engraving/libmscore/beam.h
```diff
@@ -33,18 +33,105 @@ class ChordRest;
...
+    Beam* _beam{ nullptr };
```
RomanPudashkin, line 75: All the class members should have the m_ prefix

src/engraving/libmscore/beam.cpp
```diff
@@ -1450,8 +1488,474 @@ void Beam::add8thSpaceSlant(PointF& dictatorAnchor, int dictator, int pointer, i
...
+        // or 2) the trem notes are between staves. what a pain.
+        return _trem->up();
+    }
+}
+
+BeamLayout::BeamLayout(EngravingItem* e)
+{
+    bool isGrace = false;
+    if (!e->isBeam() && !e->isTremolo()) {
+        // right now only beams and trems are supported
```
RomanPudashkin, line 1528: Is it possible to split this class into two independent classes (BeamLayout and TremoloLayout)?

src/engraving/libmscore/beam.cpp
```diff
@@ -1222,7 +1254,12 @@ void Beam::offsetBeamWithAnchorShortening(std::vector<ChordRest*> chordRests, in
     int newDictator = dictator;
     int newPointer = pointer;
     int reduce = 0;
-    while (!isValidBeamPosition(newDictator, isStartDictator, isAscending, isFlat, staffLines, true)) {
+    auto fourBeamException = [staffLines](int beams, int yPos) {
+        yPos += 400;
```
RomanPudashkin, line 1258: What does this number mean?

RomanPudashkin: There are a lot of new magic numbers in this PR. You should at least add a comment to each of them

src/engraving/libmscore/beam.h
asattely: All BeamLayout functionality moved out into `beamlayout.cpp`!

src/engraving/libmscore/beam.h
asattely: done!

src/engraving/libmscore/beam.h
asattely: done!

src/engraving/libmscore/beam.h
```diff
@@ -33,18 +33,105 @@ class ChordRest;
...
+};
 
 struct BeamSegment {
     mu::LineF line;
     int level;
     bool above; // above level 0 or below? (meaningless for level 0)
 };
 
+struct TremAnchor {
+    ChordRest* chord1;
```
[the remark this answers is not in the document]
asattely: done!

src/engraving/libmscore/beam.cpp
asattely: I've added a comment to that line, and I've searched for other places where the numbers aren't immediately obvious and couldn't find other ones. The rest make sense with the understanding that all of the beam layout logic uses units of 0.25sp.

src/engraving/libmscore/beam.cpp
asattely: I feel like the whole point of this PR is to unify beam layout behavior between two-note tremolos and beams. That is to say, everything in this class applies equally to both and the behaviors of both are exactly identical. I have intentionally minimized the amount of code that is not beam-trem agnostic, so as not to need multiple classes.

src/engraving/libmscore/beam.h
```diff
@@ -33,20 +33,107 @@ class ChordRest;
...
+struct TremAnchor {
+    ChordRest* chord1 = nullptr;
+    double y1 = 0.;
+    double y2 = 0.;
+};
+
+class BeamLayout
+{
+private:
+    static constexpr std::array _maxSlopes = { 0, 1, 2, 3, 4, 5, 6, 7 };
```
RomanPudashkin, line 69: As far I can tell, this array isn't used in this file, so it's better to declare it directly in cpp

src/engraving/libmscore/beam.cpp
RomanPudashkin: It should be renamed to BeamAndTremoloLayout at least (to show explicitly what engraving elements this class can handle)

src/engraving/libmscore/beam.h
asattely: done! moved both into `beamtremololayout` .cpp/.h

src/engraving/libmscore/beam.cpp
asattely: `BeamTremoloLayout` was my compromise, hope that's okay!

src/engraving/libmscore/beam.h
asattely: done!

src/engraving/libmscore/beam.h
```diff
@@ -33,20 +33,107 @@ class ChordRest;
...
+    enum class BeamType {
+        INVALID,
+        BEAM,
+        TREMOLO
+    };
+    BeamType m_beamType = BeamType::INVALID;
```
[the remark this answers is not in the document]
asattely: done!

src/engraving/libmscore/beamlayout.cpp
```diff
@@ -0,0 +1,1081 @@
...
+#include "tremolo.h"
+
+using namespace mu;
+using namespace mu::engraving;
+
+namespace mu::engraving {
+BeamLayout::BeamLayout(EngravingItem* e)
+{
+    bool isGrace = false;
+    if (!e->isBeam() && !e->isTremolo()) {
```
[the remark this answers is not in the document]
asattely: done!

RomanPudashkin (approved):

Issue #16219, opened 2023-02-05
[MU4 Issue] Customise note input toolbar popup points to empty space


iwoithe:
**Describe the bug**

The customise toolbar popup does not point to the customise button when the note input toolbar is vertical.

![image](https://user-images.githubusercontent.com/73422868/216811902-b09e535f-542d-4faf-8db2-3b1c5bd64bb0.png)

**To Reproduce**
1. Create a new score
2. Move the note input toolbar to a vertical orientation
3. Click on the customise toolbar button
4. See the popup pointing to empty space

**Expected behavior**
Like all other popups in MuseScore, I expect the popup to point to the button which opened it.

**Platform information**
 - Linux (Slackware 15, Xfce 4.16)

octopols:
Can I work on this issue?

octopols:
As I can see, the pop-up window have a fixed height which I don't think causes an issue when the toolbar is horizontal because there is plenty of space available bellow the customize toolbar icon and the bottom of the window but when the toolbar is vertical the height of the toolbar is bigger than the gap between customize toolbar icon and the application's bottom so it gets pushed up to a different location.
What I'm thinking to do is that I'll check if the toolbar is vertical or not and change the height of the toolbar accordingly.

octopols:
> I can't replicate this, but the current behavior in 4.5.1 is also not ideal: the popup overlaps with the side-docked toolbar, so when you hide/show things, you can't see what's happening beneath it.
> 
> The popup should appear beside the toolbar. We could just remove the pointing arrow for simplicity.
> 
> * If toolbar is left-docked, the popup should appear to its right.
> * If toolbar is right-docked, the popup should appear to its left.
> 
>  Screen.Recording.2025-03-25.at.10.06.42.AM.mov

from what I remember there were ways to open a popup directionally and if we mix some other logic to check if the toolbar is horizontal or vertical we might get what we need, I'll try to setup musescore on my windows device again and check

octopols:
@cbjeukendrup did some other PR fix the original issue as @avvvvve is not able to replicate this?

avvvvve:
Just for the record, there is still some weirdness:

1. Side-dock the note input toolbar
2. Open the customize popup
3. Top-dock the toolbar
4. Open it again, see that the arrow now points down

https://github.com/user-attachments/assets/581b68e2-bb50-4f23-a8d0-cc25fe4a7a70

PR #16226, opened 2023-02-06
Fix for Issue #16219: [MU4 Issue] Customise note input toolbar popup points to empty space

octopols:
Resolves: #16219

Customise note input toolbar popup now points to the settings cog when the toolbar is set vertical.

### DIFF
--- a/src/notation/qml/MuseScore/NotationScene/internal/NoteInputBarCustomisePopup.qml
+++ b/src/notation/qml/MuseScore/NotationScene/internal/NoteInputBarCustomisePopup.qml
@@ -31,7 +31,7 @@ StyledPopupView {
     id: root
 
     contentWidth: 280
-    contentHeight: 600
+    contentHeight: (parent.parent.height < 80) ? 600 : 500
 
     NoteInputBarCustomiseModel {
         id: customiseModel


octopols:
https://user-images.githubusercontent.com/90182257/216927062-3ffccd2f-81d8-4602-aaf8-2103d1b59599.mp4

cbjeukendrup:
This fix looks a bit naive to me. If you resize the MuseScore window to a smaller size, then the bug will probably be back. 

I think a real solution would be to move the arrow to the left side of the popup, in this case. That's a bit more complicated though.

octopols:
I was just checking how the popup behaves with different screen sizes, having a fixed height even to 2 numbers like in this fix will cause some issues (look at the screenshot attached). I'll try to write a function which takes the application size as a parameter and changed the height accordingly.
When the toolbar is vertical I think you're right the best solution is to just put the arrow on the side of the popup.

DmitryArefiev:
@octopols @cbjeukendrup Hi! Is this ready for testing?

octopols:
> @octopols @cbjeukendrup Hi! Is this ready for testing?

Sorry for the late reply but this fix does not completely solves the problem. I'm trying to work on the solution Casper suggested and I hope it doesn't take that long to implement.

DmitryArefiev:
Good. Thanks!

cbjeukendrup:
Closing due to inactivity and per https://github.com/musescore/MuseScore/pull/16226#issuecomment-1418998486

octopols:
> I was just checking how the popup behaves with different screen sizes, having a fixed height even to 2 numbers like in this fix will cause some issues (look at the screenshot attached). I'll try to write a function which takes the application size as a parameter and changed the height accordingly. When the toolbar is vertical I think you're right the best solution is to just put the arrow on the side of the popup. Screenshot (42) Screenshot (43)

@avvvvve can you check if this issue still exists?

PR #16232, opened 2023-02-06
export-midi: removed SND render option

handrok:
RenderMidi used to be a part of playback engine. So it has a lot of features and code. Now MPE is using. RenderMidi is using for exporting score to midi file. Therefore a lot of features in RenderMidi is not needed. In this PR single note dynamic has been removed from midi export.

### DIFF
--- a/src/engraving/compat/midi/midirender.h
+++ b/src/engraving/compat/midi/midirender.h
@@ -33,12 +33,6 @@ class MasterScore;
 class Staff;
 class SynthesizerState;
 
-enum class DynamicsRenderMethod : signed char {
-    FIXED_MAX,
-    SEG_START,
-    SIMPLE
-};
-
 //---------------------------------------------------------
 //   RangeMap
 ///   Helper class to keep track of status of status of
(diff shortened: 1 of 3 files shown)


igorkorsukov (approved):

src/engraving/compat/midi/midirender.h
```diff
@@ -123,8 +114,6 @@ class MidiRenderer
 
     void collectMeasureEvents(EventMap* events, Measure const* m, const MidiRenderer::StaffContext& sctx, int tickOffset,
                               PitchWheelRenderer& pitchWheelRenderer);
-    void collectMeasureEventsSimple(EventMap* events, Measure const* m, const StaffContext& sctx, int tickOffset,
-                                    PitchWheelRenderer& pitchWheelRenderer);
     void collectMeasureEventsDefault(EventMap* events, Measure const* m, const StaffContext& sctx, int tickOffset,
```
alexpavlov96, line 117: maybe then we can just name it "collectMeasureEvents"?

src/engraving/compat/midi/midirender.h
```diff
@@ -105,9 +99,6 @@ class MidiRenderer
     struct StaffContext
     {
```
alexpavlov96, line 100: let's simply pass const Staff* to methods

src/engraving/compat/midi/midirender.h
handrok: collectMeasureEvents method already exist. So renamed collectMeasureEventsDefault to doCollectMeasureEvents. And made this method private

alexpavlov96 (approved):

PR #16238, opened 2023-02-06
export-midi: refactored MidiRender class; removed chunk logics

handrok:
RenderMidi used to be a part of playback engine. So it has a lot of features and code. Now MPE is using. RenderMidi is using for exporting score to midi file. Therefore a lot of features in RenderMidi is not needed. On of these features is render according to the chunks. It was necessary because RenderMidi is used as  playback engine. It was overhead to render full score only if one measure was changed. Now MidiRender is used only once. So removing chunk rendering make code simpler and clearer.

MidiRender.h was changed according to the codestyle. Public method above, private below.

### DIFF
--- a/src/engraving/compat/midi/midirender.cpp
+++ b/src/engraving/compat/midi/midirender.cpp
@@ -77,7 +77,6 @@ using namespace mu;
 using namespace mu::engraving;
 
 namespace mu::engraving {
-static constexpr int MIN_CHUNK_SIZE(10); // measure
 static PitchWheelSpecs wheelSpec;
 
 static bool graceNotesMerged(Chord* chord);
@@ -634,7 +633,6 @@ void MidiRenderer::doCollectMeasureEvents(EventMap* events, Measure const* m, co
 MidiRenderer::MidiRenderer(Score* s)
     : score(s)
 {
-    setMinChunkSize(MIN_CHUNK_SIZE);
 }
 
 //---------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/engraving/compat/midi/midirender.h
```diff
@@ -62,76 +62,39 @@ class RangeMap
...
-    void renderStaffChunk(const Chunk&, EventMap* events, const Staff* sctx, PitchWheelRenderer& pitchWheelRenderer);
+    static const int ARTICULATION_CONV_FACTOR { 100000 };
 
-    void renderSpanners(const Chunk&, EventMap* events, PitchWheelRenderer& pitchWheelRenderer);
+private:
 
-    void renderMetronome(const Chunk&, EventMap* events);
-    void renderMetronome(EventMap* events, Measure const* m, const Fraction& tickOffset);
+    Score* score{ nullptr };
+    int minChunkSize = 0;
```
alexpavlov96, line 83: we don't need minChunkSize now?

src/engraving/compat/midi/midirender.h
```diff
@@ -62,76 +62,39 @@ class RangeMap
...
 
-    void renderStaffChunk(const Chunk&, EventMap* events, const Staff* sctx, PitchWheelRenderer& pitchWheelRenderer);
+    static const int ARTICULATION_CONV_FACTOR { 100000 };
 
-    void renderSpanners(const Chunk&, EventMap* events, PitchWheelRenderer& pitchWheelRenderer);
+private:
 
-    void renderMetronome(const Chunk&, EventMap* events);
-    void renderMetronome(EventMap* events, Measure const* m, const Fraction& tickOffset);
+    Score* score{ nullptr };
```
alexpavlov96, line 82: should be "score = nullptr" (and other places)

src/engraving/compat/midi/midirender.h
handrok: you are correct. missed it

src/engraving/compat/midi/midirender.cpp
```diff
@@ -1095,75 +1051,6 @@ bool MidiRenderer::canBreakChunk(const Measure* last)
     return true;
 }
 
-//---------------------------------------------------------
-//   MidiRenderer::updateChunksPartition
-//---------------------------------------------------------
-
-void MidiRenderer::updateChunksPartition()
```
alexpavlov96, line 1102: also "canBreakChunk", probably everything with "Chunk" =)

src/engraving/compat/midi/midirender.h
handrok: fixed

src/engraving/compat/midi/midirender.cpp
handrok: fixed

alexpavlov96 (approved):

PR #16242, opened 2023-02-06
Fix #16146, Copy-paste plain text outside MuseScore

### LINKED ISSUE #16146: [MU4 Issue] When pasted outside MuseScore copied Measure Text carries markup tags
**Text Copy/Paste bug**
It's important that users can copy and paste text in and out of MuseScore Text objects. But on MacOS when I paste text from a MuseScore text field all the formatting tags are included.

I've attached an example of what I see on pasting outside of MuseScore.

**To Reproduce**
1. Add a Measure Text object
2. Add some formatting like bold and italic
3. Copy
4. Paste to another app
5. The issue: The pasted text includes formatting tags

**Expected behavior**
Text copied from MuseScore should paste elsewhere as formatted text (without markup tags) or as plain text.

**Original forum post**
See MuseScore forum: https://musescore.org/en/node/343914

**Platform information**
 MacOS Ventura 13.1

**Screenshot**
The attached screenshot shows the result on paste.

Formatted Measure text as pasted to another app

softcat477:
Resolves: #16146 

Before this PR, when a user copies a text and past it outside of musescore, the text includes formatting tags (e.g. `<font>` `<b>`). This does not happen in `MU3`, where you can copy-paste a text outside of `MU3` and get plain text. However, `MU4` is able to keep the original format when we copy-paste a text to another text object since it also copies the formatting tags. `MU3` failed to do so because it does not copy the formatting tag.

This PR keeps what `MU4` can do (keep the formatting) while letting users paste plain texts outside `MU4`. It is done using `QMimeData` to store both plain and rich text in different MIME types. The plain text is stored with type `text/plain`, and the rich text is stored in a custom MIME type `application/musescore/richtext`. When pasting texts to another text object in MU4, MU4 searches for rich text in custom type `application/musescore/richtext` and paste the rich text to keep the format. When pasting text outside of musescore, the plain text with type `text/plain` will be used so users can get rid of the formatting tags.

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -3463,7 +3463,10 @@ void NotationInteraction::copySelection()
         m_editData.element->editCopy(m_editData);
         mu::engraving::TextEditData* ted = static_cast<mu::engraving::TextEditData*>(m_editData.getData(m_editData.element).get());
         if (!ted->selectedText.isEmpty()) {
-            QGuiApplication::clipboard()->setText(ted->selectedText);
+            QMimeData* mimeData = new QMimeData();
+            mimeData->setData(TextEditData::mimeRichTextFormat, ted->selectedText.toQString().toUtf8());
+            mimeData->setText(ted->selectedPlainText);
+            QGuiApplication::clipboard()->setMimeData(mimeData);
         }
     } else {
         QMimeData* mimeData = selection()->mimeData();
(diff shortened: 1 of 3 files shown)


src/notation/internal/notationinteraction.cpp
```diff
@@ -3597,8 +3600,11 @@ void NotationInteraction::pasteSelection(const Fraction& scale)
         #else
         QClipboard::Mode mode = QClipboard::Selection;
         #endif
-        QString txt = QGuiApplication::clipboard()->text(mode);
-        toTextBase(m_editData.element)->paste(m_editData, txt);
+        const QMimeData* mimeData = QApplication::clipboard()->mimeData();
+        if (mimeData->hasFormat(TextEditData::mimeRichTextFormat)) {
```
cbjeukendrup, line 3604: This seems to break the ability to paste (plain) text that was copied outside MuseScore, so you'd need to add some else-if or so to handle that.

For the rest this looks really good!

softcat477: Oops, thanks for pointing it out. Yes, it did break. I added an else section so that it will paste texts copied outside Musescore. Those texts are stored in `plain/text`, so `QMimeData().text()` should work. 

I also add `mode` back or else Musescore in linux will break (i guess i don't have a linux on my side)

cbjeukendrup (approved):

RomanPudashkin:
@softcat477 please rebase this PR. Thanks!

softcat477:
it's up.

zacjansheski:
Tested in Windows 11, Mac OS 11.6.5 , Ubuntu 20.04.6

https://github.com/musescore/MuseScore/issues/16146 FIXED

PR #16258, opened 2023-02-07
fixed midi rendering for last measure

alexpavlov96:
regression after https://github.com/musescore/MuseScore/pull/16238
(last measure of score was not treated)


handrok (approved):

abariska:
🟢 **Tested**

Issue #16217, opened 2023-02-05
[MU4 Issue] Copy-paste of 2 voices can create incorrect initial rest.


aige02:
**Describe the bug**
In 4/4, copying the last 1.5 beats of two voices into a measure with 2.5 beats of voice 1 creates a 3 beat rest in voice 2.  I found and reproduced this in a drumset staff.

**To Reproduce**
1. Create a measure with two voices each with a quarter note and an eight note.
2. In another measure, start with 2.5 beats of rest in voice 1 (and no voice 2).
3. Copy the quarter and eighth notes (both voices).
4. Paste the notes after the 2.5 beats.  The 2nd voice should (incorrectly) have a dotted half rest giving it 4.5 beats.
 
**Expected behavior**
The second voice should have either no visible rest or 2.5 beats of rest.

**Screenshots**
![image](https://user-images.githubusercontent.com/20604032/216808298-924deff2-0dbc-4a16-852a-18f0324fe576.png)

**Platform information**
 - Windows 11
 - Nightly build 230350507-4.0.2-aabbec1-x86_64

PR #16260, opened 2023-02-07
Fix incorrect rests when filling measure

mike-spa:
Resolves: #16217 

I'm quite inexperienced in this area of our codebase so I hope this is a good enough solution. Basically, the bug happens because we are assuming that the gap before the pasted notes should be covered by a single rest figure (possibly dotted), but obviously that's not always possible (for instance, if you have a total rest of 5/8 like in this example). 
So I'm now using `toRhythmicDurationList` to break the total rest into smaller ones according to the meter of the measure, this way the rests before a pasted chord will obey the exact same logic as those after.

### DIFF
--- a/src/engraving/libmscore/check.cpp
+++ b/src/engraving/libmscore/check.cpp
@@ -21,12 +21,14 @@
  */
 
 #include "chordrest.h"
+#include "durationtype.h"
 #include "factory.h"
 #include "keysig.h"
 #include "masterscore.h"
 #include "measure.h"
 #include "rest.h"
 #include "segment.h"
+#include "sig.h"
 #include "staff.h"
 #include "tuplet.h"
 #include "utils.h"
@@ -268,15 +270,23 @@ void Measure::fillGap(const Fraction& pos, const Fraction& len, track_idx_t trac
          len.numerator(), len.denominator(),
          stretch.numerator(), stretch.denominator(),
          track);
-    TDuration d;
-    d.setVal(len.ticks());
-    if (d.isValid()) {
+    std::vector<TDuration> durationList;
+    if (useGapRests) { // fill this gap with a single rest
+        TDuration d;
+        d.setVal(len.ticks());
+        durationList.push_back(d);
+    } else { // break the gap into shorter durations if necessary
+        durationList = toRhythmicDurationList(len, true, pos, score()->sigmap()->timesig(tick()).nominal(), this, 0);
+    }
+    Fraction curTick = pos;
+    for (TDuration d : durationList) {
         Rest* rest = Factory::createRest(score()->dummy()->segment());
-        rest->setTicks(len);
+        rest->setTicks(d.fraction());
         rest->setDurationType(d);
         rest->setTrack(track);
         rest->setGap(useGapRests);
-        score()->undoAddCR(rest, this, (pos / stretch) + tick());
+        score()->undoAddCR(rest, this, curTick + tick());
+        curTick += d.fraction();
     }
 }
 


MarcSabatella:
I can at least confirm that this was indeed the original purpose of that function, for handling things like this.  So the code looks as I'd expect it to, assuming the desired result is to have the voice 2 rests at all.  MU3 pasted the notes without pre-filling, but I'd probably rather have them and then decide for myself if they can be removed (or replaced with notes - which is quite difficult if the initial paste results in a "hole" like this).

src/engraving/libmscore/check.cpp
```diff
@@ -232,15 +234,16 @@ void Measure::fillGap(const Fraction& pos, const Fraction& len, track_idx_t trac
          len.numerator(), len.denominator(),
          stretch.numerator(), stretch.denominator(),
          track);
-    TDuration d;
-    d.setVal(len.ticks());
-    if (d.isValid()) {
+    std::vector<TDuration> durationList = toRhythmicDurationList(len, true, pos, score()->sigmap()->timesig(tick()).nominal(), this, 0);
```
cbjeukendrup, line 237: Shouldn't `score()->sigmap()->timesig(tick()).nominal()` be replaced with `_len`, to take irregular measures into account?

MarcSabatella: I am thinking the answer is no, at least not so directly.  For example, a measure of 4/4 might be lengthened to 9/8 to accommodate some sort of cadenza, but that doesn't mean we should use the group rules for compound meter.

Or, another case - you might have a measure of length 4/4 that has been split into two measures of 5/8 and 3/8, but that doesn't mean you want leading rests to actually group according to those time signatures.  They should still group as 4/4.  But, the first partial measure should group as if it is the *first* part of a 4/4 measure, and the second partial measure should group as if it is the *second*.  So for example, if there is to be one beat worth of rest, the first 5/8 portion should show it as a quarter since this truly is beat 1, but the second 3/8 portion should show it as two eighths since this is actually starting on the "&" of 3.

Unfortunately, we don't have a way of knowing whether any given irregular measure is meant to be the first portion or second portion of a split measure, or if indeed it represents a split at all or something else entirely.

The swing playback system has to deal with similar issues of trying to figure out if a shorter-than-nominal measure that has odd number of eighths should be seen as starting on the beat or off the beat.  We used an awkward but somewhat effective hack, basically checking to see if this is the first measure of the piece (making it a pickup) and also if there is a system or similar break delineating the partial measures.  Probably we should be checking for repeats too.

Might be nice to have a general purpose function to determine this, and to do it a bit smarter, like looking behind and ahead for a "matching" partial measure.  Such a function would also come in handy when generating rests as well as when determining beaming for pickup measures, which we basically always get wrong since we treat them as "first portion" rather than "second portion".

RomanPudashkin (approved):

RomanPudashkin:
@mike-spa please rebase it

mike-spa:
Done! :+1:

src/engraving/libmscore/check.cpp
aige02: Since the grouping may not be not evident, one solution is for a measure to have a grouping flag to provide, to the method filling the measure, a hint of "left" or "right". This flag might also be useful when generating flags on notes in that measure (e.g. five eighth notes in a 5/8 partial measure).

This problem also applies to a pickup measure before the first full measure.  MU typically groups a pickup measure from the left instead of from the right.  For example, a 3/8 pickup measure, before a 4/4 full measure, should be initialized as an eighth rest followed by a quarter rest, but MU defaults to quarter then eighth.

src/engraving/libmscore/check.cpp
```diff
@@ -268,15 +270,23 @@ void Measure::fillGap(const Fraction& pos, const Fraction& len, track_idx_t trac
          len.numerator(), len.denominator(),
          stretch.numerator(), stretch.denominator(),
          track);
-    TDuration d;
-    d.setVal(len.ticks());
-    if (d.isValid()) {
+    std::vector<TDuration> durationList;
+    if (useGapRests) { // fill this gap with a single rest
+        TDuration d;
+        d.setVal(len.ticks());
```
MarcSabatella, line 276: OK, thinking about https://github.com/musescore/MuseScore/issues/16841 - what does this do if there is no single duration that can represent the length?  The bug in question was a case of a half rest plus an eighth rest being deleted.  That's a gap you can't fill with a single rest.  I'd think you still need to use the duration list thing even for gap rests, but - I know almost nothing about gap rests, maybe they're special in that way?

aige02: I think this code may have been a kludge.

I've always considered using the single rest when necessitated by pasting to be wrong.  The generated rest(s) should always be the same as if a partial measure (such as a pickup measure) were generated.  The code to generate rests for a partial measure should used whenever a partial measure of rests are generated at the beginning of a measure.

Issue #14685, opened 2022-11-23
[MU4 Issue] Crash when opening one-measure score with 2nd ending assigned


abariska:
**Describe the bug**
This is an edge case but should be fixed 

**To Reproduce**
1. Create a simple score with 1 measure
2. Add 2nd ending from palettes
3. Save the score
4. Open the score

**Expected behavior**
Score is opened

**Screenshots**

https://user-images.githubusercontent.com/101250347/203553053-75f3cb2d-3305-4358-902b-bd9ddef72098.mp4

Eism:
@abariska 
The crash occurs in the beta, but doesn't occur in the nightly build https://github.com/musescore/MuseScore/actions/runs/3537604380

Looks like it's been fixed.

Eism:
Can reproduce

abariska:
🟢 **FIXED**

jeetee:
Verified that 3.6.2 did not crash on this scenario (just some sanity testing before diving into alternative fixes for this).

PR #16263, opened 2023-02-07
Repeat test follow-up and skipped scores

jeetee:
Resolves: #14685 (with additional test case 68) and re-enables repeat tests 57 and 67 as a follow-up to https://github.com/musescore/MuseScore/pull/16213

### DIFF
--- a/src/engraving/libmscore/repeatlist.cpp
+++ b/src/engraving/libmscore/repeatlist.cpp
@@ -834,10 +834,10 @@ void RepeatList::unwind()
                     activeVolta = nullptr;
                     // Start next rs on the following measure
                     Measure const* const possibleNextMeasure = (*repeatListElementIt)->measure->nextMeasure();
-                    rs = new RepeatSegment(playbackCount);
                     if (possibleNextMeasure == nullptr) {
-                        // end of score, but will still encounter section break, notify it
+                        rs = nullptr;                   // end of score, but will still encounter section break, notify it
                     } else {
+                        rs = new RepeatSegment(playbackCount);
                         rs->addMeasure(possibleNextMeasure);
                     }
                 }
--- a/src/engraving/libmscore/repeatlist.cpp
+++ b/src/engraving/libmscore/repeatlist.cpp
@@ -263,7 +263,11 @@ int RepeatList::utime2utick(double secs) const
         }
     }
 
-    ASSERT_X(String(u"time %1 not found in RepeatList").arg(secs));
+    if (!empty()) {
+        ASSERT_X(String(u"time %1 not found in RepeatList").arg(secs));
+    }
+    // else: requesting from an empty map can be expected as a valid scenario
+
     return 0;
 }
 
@@ -1028,14 +1032,16 @@ void RepeatList::unwind()
         }
 
         // Reached the end of this section
-        // Inform the last RepeatSegment that the Section Break pause property should be honored now
-        rs = this->back();
-        repeatListElementIt = sectionIt->cend() - 1;
-        assert((*repeatListElementIt)->repeatListElementType == RepeatListElementType::SECTION_BREAK);
+        if (!this->empty()) {
+            // Inform the last RepeatSegment that the Section Break pause property should be honored now
+            rs = this->back();
+            repeatListElementIt = sectionIt->cend() - 1;
+            assert((*repeatListElementIt)->repeatListElementType == RepeatListElementType::SECTION_BREAK);
 
-        LayoutBreak const* const layoutBreak = toMeasureBase((*repeatListElementIt)->element)->sectionBreakElement();
-        if (layoutBreak != nullptr) {
-            rs->pause = layoutBreak->pause();
+            LayoutBreak const* const layoutBreak = toMeasureBase((*repeatListElementIt)->element)->sectionBreakElement();
+            if (layoutBreak != nullptr) {
+                rs->pause = layoutBreak->pause();
+            }
         }
     }
 
--- a/src/engraving/tests/repeat_data/repeat68.mscx
+++ b/src/engraving/tests/repeat_data/repeat68.mscx
@@ -1,0 +1,148 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<museScore version="4.00">
+  <Score>
+    <LayerTag id="0" tag="default"></LayerTag>
+    <currentLayer>0</currentLayer>
+    <Division>480</Division>
+    <Style>
+      <showMeasureNumberOne>1</showMeasureNumberOne>
+      <measureNumberInterval>1</measureNumberInterval>
+      <measureNumberSystem>0</measureNumberSystem>
+      <Spatium>1.76389</Spatium>
+      </Style>
+    <showInvisible>1</showInvisible>
+    <showUnprintable>1</showUnprintable>
+    <showFrames>1</showFrames>
+    <showMargins>0</showMargins>
+    <metaTag name="arranger"></metaTag>
+    <metaTag name="composer"></metaTag>
+    <metaTag name="copyright"></metaTag>
+    <metaTag name="lyricist"></metaTag>
+    <metaTag name="movementNumber"></metaTag>
+    <metaTag name="movementTitle"></metaTag>
+    <metaTag name="mscVersion">4.00</metaTag>
+    <metaTag name="poet"></metaTag>
+    <metaTag name="source"></metaTag>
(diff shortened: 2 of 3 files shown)


jeetee:
[WIP] (have to run now, but will elaborate later today)

cbjeukendrup (approved):

jeetee:
Elaboration:
While this works, in theory, trying to play back the new repeat68 test file will ASSERT_X crash under debug mode (it is fine under release and commenting out the assert). I'll just sanity check what happens in other scenarios for the 3s additional playback time (which here is the cause of the assert trigger) and likely we only need the assert for non-empty scores.

alexpavlov96:
@jeetee would PR fix playback of this file? If no, any ideas?
[Da coda.mscz.zip](https://github.com/musescore/MuseScore/files/10678834/Da.coda.mscz.zip)

jeetee:
> @jeetee would PR fix playback of this file? If no, any ideas? [Da coda.mscz.zip](https://github.com/musescore/MuseScore/files/10678834/Da.coda.mscz.zip)

Your "Da Coda" is not an actual "To Coda" marking, but a broken jump instruction. Remove it and replace it with a normal "To Coda" from the palettes. You may then double click that to change it's displayed text.

jeetee:
Didn't mean to remove @Eism from the review list, but re-requesting Casper's apparently automatically did that..

ASSERT conditionally set to allow for empty repeatlists.

abariska:
> > @jeetee would PR fix playback of this file? If no, any ideas? [Da coda.mscz.zip](https://github.com/musescore/MuseScore/files/10678834/Da.coda.mscz.zip)
> 
> Your "Da Coda" is not an actual "To Coda" marking, but a broken jump instruction. Remove it and replace it with a normal "To Coda" from the palettes. You may then double click that to change it's displayed text.

Actually, it's another common mark which present in Palettes. So, it should work same way as To Coda mark

jeetee:
> > > @jeetee would PR fix playback of this file? If no, any ideas? [Da coda.mscz.zip](https://github.com/musescore/MuseScore/files/10678834/Da.coda.mscz.zip)
> > 
> > 
> > Your "Da Coda" is not an actual "To Coda" marking, but a broken jump instruction. Remove it and replace it with a normal "To Coda" from the palettes. You may then double click that to change it's displayed text.
> 
> Actually, it's another common mark which present in Palettes. So, it should work same way as To Coda mark
> 
Please open up a new issue for "Da Coda" and "Da Double Coda" markings; they are simply the wrong element types in the palettes and should be corrected. Whomever added them modelled them as Jumps instead of Markers.

abariska:
@jeetee Here it is https://github.com/musescore/MuseScore/issues/16294

Eism:
@jeetee 
I think you are solving the problem incorrectly. In your solution, the repeat list is empty(shouldn't it be empty?) after the method is executed. Because of this, the playback does not work at all.

https://user-images.githubusercontent.com/10116828/218399490-f225c7e8-7731-4c67-916c-c5878c25beab.mov

cbjeukendrup:
@Eism The expected behaviour is that there is no playback at all in this case. (The measure should be skipped because it is in Volta 2.) 
But indeed the infinite time counter will need to be fixed somehow...

Eism:
@cbjeukendrup 
The measure is played in Sibelius. 

Why did you decide that the measure should not be played? Where can I read about it?

https://user-images.githubusercontent.com/10116828/218403520-cde2394b-1028-4372-a5fe-16b73fae3ad7.mov

jeetee:
@Eism Then Sibelius is "wrong" according to standard notation ;-)

The request of such a score is "play whatever is under this volta only the 2nd time, without having a 2nd time". My guess is that Sibelius tries to be "helpful" in allowing playback of such measure assuming the user is wrong. I'm curious to how they would handle exotic, yet correct scores such as the following instead:
image
[swapped-volta-logic.zip](https://github.com/musescore/MuseScore/files/10720160/swapped-volta-logic.zip)
The correct playback is measures 3, 4, 1, 2.

The infinity counter is to my knowledge not caused by this PR though. Here's what I see when building this branch:
skip-volta

jeetee:
And of course also refer to the two existing tests re-enabled by this PR, such as 57 showing that a non-matching volta is expected to be skipped. (57 in Sibelius would be interesting as well).

Eism:
I'll leave it for @Tantacrul 

@jeetee 
About playback, the bug is reproducing for me on your branch, before that, I reproduced on master with your changes. Please make sure you do exactly as shown in the video. (the newly created score)

jeetee:
I can confirm infinity counter by following the from scratch steps from @Eism
It is "fixed" by saving and re-opening the score.

So now we're left with wondering what causes it..

abariska:
@jeetee @cbjeukendrup @Eism 
2nd ending in the first measure is a negative case (unnatural) and it doesn't make any sense to take into account the ending and apply its regular rule. The only good solution is to ignore the ending in this case. Also, it will be OK if, e.g. the piece is divided to two parts (files). Imagine that first ending is in the first file and the second one starts from 2nd ending. User shouldn't be blocked to play the part.

jeetee:
@abariska I respectfully disagree

There is no way to differentiate between a skipped volta at start that makes sense and one that doesn't. You can't skip the volta in this scenario without breaking standard notation or breaking the other (uncommon, yet valid) example given [above](https://github.com/musescore/MuseScore/pull/16263#issuecomment-1427540838)

Likewise there are valid (though admittedly slight "hacky") uses for invisible voltas skipping the start of a score, such as old style incipits or handbel summaries at the start of a score.

As mentioned, the thing itself (a volta skipping the start, or by extension the entire score) is an edge case in itself. To me having it does not warrant breaking standard notation.

As for the use case raised by @abariska: placing a 2nd volta part in a separate score. A user can freely either change the text of a first volta or change the repeatlist of the second volta for that 2nd score to get functioning playback without the need for maintaining special coding for such a (imho) contrived scenario.

Tantacrul:
I'm kind of conflicted in wading into this. Placing a 2nd volta on measure one... in a single measure score... I can see either argument.

Argument 1. It's wrong, so there's no reason to provide an exception for it. It would just not play. That's fine.
Argument 2. We should not **allow** things to be weird, so in this specific case, we should ignore the volta

I really don't strongly favour either approach because it's such an edge case. Part of me thinks: **whatever doesn't take up extra dev time.**

I guess I'd slightly favour argument 1 because it (kind of) makes sense from a technical POV.

abariska:
As a result of the discussion I don't see any reason to merge this PR, cause it doesn't cover all possible negative cases with endings and, actually, creates more bugs to fix. There investigation should be done in this area and then some specific rules to be implemented for negative cases. Due to low priority it won't be done in the nearest future, so lets just leave Elnur's crash fix as an intermediate step

cbjeukendrup:
@abariska It does not create more bugs. The bug about the infinite play time counter turned out to exist in Master too. 

Also, the automated unit tests do make clear that this PR fixes bugs that occurred in musically correct cases; those bugs were introduced by the previous fix. (But we didn't spot them back then, because the unit tests were not executed at that moment.)

jeetee:
Admittedly I'm somewhat biased about this PR as it is mine :)

But let's try to summarize the facts addressed by this PR and the discoveries inside it:
**This PR**
1. The original crash Fix by Elnur, fixed that crash, but introduced two new bugs. The new bugs would've been caught by test57 and 67, but those tests were disabled at the time.

2. This PR reverts the original crash fix and introduces a new one, in addition adding repeat68 as a test to prove the crash was fixed

3. This PR activates those remaining tests so we won't be bitten by this again in the future. Those tests passing also validate the fix from (2)

**Additional bugs** discovered during testing this PR, but not caused by it:
1. "Infinity" playback counter (99s) upon score creation: moved into #16343 which has a possible fix pending in #16345

2. "Da (double) Coda" markings were set as Jumps in the palettes instead of Markers: moved into #16294 with a fix pending in #16305 

**Speculation** about whether the contrived test case (repeat68) that used to cause a crash before Elnur's fix should be interpreted in a more "make something from this illogical request" instead of the current "follow the standard default logic" (as per this PR as well) method.
Keeping in mind that if we want to "make something" from it, it'll require some additional thought to not break all the other cases of standard notation.

abariska:
@jeetee OK, eventually, I agree. 
Lets merge this PR as a regression fix and one day we'll come back to negative cases with endings

Eism (approved):

PR #16265, opened 2023-02-07
midi-export: each string has its own channel

handrok:
Musescore org. has mobile and versions of musescore editor. This product is inside a private repository. So I cant link to the issue. But long story short playback engine of mobile version is still based on RenderMidi class. We need to support correct playback of bends. Problem is bend applies for whole chord instead of note. To fix it each string render to separate channel. This behavior is defined by eachStringHasChannel flag.

From editor point of view nothing changed. Because eachStringHasChannel = false

### DIFF
--- a/src/engraving/compat/midi/midirender.cpp
+++ b/src/engraving/compat/midi/midirender.cpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * SPDX-License-Identifier: GPL-3.0-only
  * MuseScore-CLA-applies
  *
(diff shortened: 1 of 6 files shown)


src/engraving/compat/midi/midirender.cpp
```diff
@@ -597,10 +598,8 @@ void MidiRenderer::doCollectMeasureEvents(EventMap* events, Measure const* m, co
             Chord* chord = toChord(cr);
 
             Instrument* instr = st1->part()->instrument(tick);
-            int subchannel = chord->upNote()->subchannel();
-            int channel = instr->channel(subchannel)->channel();
 
-            events->registerChannel(channel);
+//            events->registerChannel(channel);
```
alexpavlov96, line 602: may we need it later?

handrok: fixed

alexpavlov96 (approved):

PR #16268, opened 2023-02-07
Fix MSVC compiler warnings

Jojo-Schmitz:

### DIFF
--- a/src/importexport/midi/internal/midiexport/exportmidi.cpp
+++ b/src/importexport/midi/internal/midiexport/exportmidi.cpp
@@ -220,6 +220,7 @@ void ExportMidi::writeHeader()
 
 bool ExportMidi::write(QIODevice* device, bool midiExpandRepeats, bool exportRPNs, const SynthesizerState& synthState)
 {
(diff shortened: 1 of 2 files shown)


src/importexport/midi/internal/midiexport/exportmidi.cpp
```diff
@@ -220,6 +220,7 @@ void ExportMidi::writeHeader()
 
 bool ExportMidi::write(QIODevice* device, bool midiExpandRepeats, bool exportRPNs, const SynthesizerState& synthState)
 {
+    UNUSED(midiExpandRepeats);
```
alexpavlov96, line 223: will be brought back in next PR's, but for now ok

alexpavlov96 (approved):

src/importexport/midi/internal/midiexport/exportmidi.cpp
Jojo-Schmitz: I hoped so, this warning the me looked like an error 😜

PR #16275, opened 2023-02-08
Add user directory default paths for MuseSampler

iamtesch:
The MuseSampler library should be able to be loaded from per-user, non-admin privilege directories.  (This will be the preferred install location for future versions of Muse Hub, to reduce dependence on administrative privileges).

This PR attempts to load the sampler first from the specified user directory per platform:
macOS: `~/Library/Application Support/MuseSampler/lib`
linux: `~/.local/share/MuseSampler/lib`
windows: `<userhome>\AppData\Local\MuseSampler\lib`

The fallback is to the current system-wide location, per current behavior.  Other questions noted inline.

### DIFF
--- a/src/framework/musesampler/internal/musesamplerconfiguration.cpp
+++ b/src/framework/musesampler/internal/musesamplerconfiguration.cpp
@@ -34,18 +34,21 @@ static io::path_t DEFAULT_PATH()
 {
     return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/MuseSampler/lib/" + DEFAULT_LIB_NAME;
 }
+
 #elif defined(Q_OS_MAC)
 static const io::path_t DEFAULT_LIB_NAME("libMuseSamplerCoreLib.dylib");
 static io::path_t DEFAULT_PATH()
 {
     return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/MuseSampler/lib/" + DEFAULT_LIB_NAME;
 }
+
 #else
 static const io::path_t DEFAULT_LIB_NAME("MuseSamplerCoreLib.dll");
 static io::path_t DEFAULT_PATH()
 {
     return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "\\MuseSampler\\lib\\" + DEFAULT_LIB_NAME;
 }
+
 #endif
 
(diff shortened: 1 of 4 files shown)


src/framework/musesampler/internal/musesamplerconfiguration.cpp
```diff
@@ -23,27 +23,48 @@
...
 #if defined(Q_OS_LINUX)
-static const io::path_t DEFAULT_PATH("libMuseSamplerCoreLib.so");
+static const io::path_t DEFAULT_LIB_NAME("libMuseSamplerCoreLib.so");
+static io::path_t DEFAULT_PATH()
+{
+    return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/MuseSampler/lib/" + DEFAULT_LIB_NAME;
+}
 #elif defined(Q_OS_MAC)
-static const io::path_t DEFAULT_PATH("/usr/local/lib/libMuseSamplerCoreLib.dylib");
+static const io::path_t DEFAULT_LIB_NAME("libMuseSamplerCoreLib.dylib");
```
iamtesch, line 38: This used to be specified as a full path on macOS for the system directory; this worked as a fallback without this for me, though, on Monterey.  Any reason the full absolute path was used here before?

src/framework/musesampler/internal/musesamplerconfiguration.cpp
```diff
@@ -23,27 +23,48 @@
...
+static io::path_t DEFAULT_PATH()
+{
+    return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/MuseSampler/lib/" + DEFAULT_LIB_NAME;
+}
 #else
-static const io::path_t DEFAULT_PATH("MuseSamplerCoreLib.dll");
+static const io::path_t DEFAULT_LIB_NAME("MuseSamplerCoreLib.dll");
+static io::path_t DEFAULT_PATH()
```
iamtesch, line 45: The three `DEFAULT_PATH` functions could probably be combined with `Qt` logic for path handling.  I wasn't familiar with this, but could look into and change if desired.

Also, I was able to build and test mac + linux locally, but not Windows; I don't have a Windows built setup for MuseScore atm.

vpereverzev (approved):

DmitryArefiev:
Tested Muse Sounds visability in 4.0.2 RC builds on Win10/Mac14/Linux Ubuntu - **all fine**

Issue #16248, opened 2023-02-07
[MU4 Issue] Musescore crashing when opening a project


BryanIsntDead:
**Describe the bug**
Upon opening the attatched mscz file, MU 3 & 4 instantly crash with no warning.

**To Reproduce**
I don't know how to recreate the corrupted file, but just opening it in MU 3 & 4 instantly crashes the app.

**Expected behavior**
MU3 would normally say the file is corrupted, which it doesn't. I know MU4 doesn't do that as of now, but I don't think it would normally just close with no warning.

**Platform information**
 - Windows

**Additional context**
I really want to know if there is a way to make the file openable again because I spent a lot of time making this, and theres a lot I want to do with it. Thank you
[Upload.zip](https://github.com/musescore/MuseScore/files/10670951/Upload.zip)

cbjeukendrup:
Hi! I fixed the cause of this crash, and then in the fixed version of MuseScore I re-saved the file. Now it also seems to open again in the current version of MuseScore: 
[Give us this day 1 copy.mscz.zip](https://github.com/musescore/MuseScore/files/10690466/Give.us.this.day.1.copy.mscz.zip)

BryanIsntDead:
> Hi! I fixed the cause of this crash, and then in the fixed version of MuseScore I re-saved the file. Now it also seems to open again in the current version of MuseScore: [Give us this day 1 copy.mscz.zip](https://github.com/musescore/MuseScore/files/10690466/Give.us.this.day.1.copy.mscz.zip)

Thank you so much! I've worked at trying to get this back for weeks now!

DmitryArefiev:
Fixed in #16284

DmitryArefiev:
Retested on Win10

Issue #16235, opened 2023-02-06
[MU4 Issue] Musescore crashes when opening the score


urbanrieger:
**Describe the bug**
When I open the score arrached, MS4 crashes.

**To Reproduce**
1. Open the attached score (zipped) in MS4. The score was written in MS4. 
2. Musescore 4 opens and crashes immediatly

**Expected behavior**
Musescore 4 shoud open the score.

**Screenshots**
-

**Platform information**
- Windows 11, MuseScore 4.0.1

**Additional context**
-
[Freude_schoener_GoetterfunkenMS4.zip](https://github.com/musescore/MuseScore/files/10664473/Freude_schoener_GoetterfunkenMS4.zip)

cbjeukendrup:
Hi! I fixed the cause of this crash, and then in the fixed version of MuseScore I re-saved the file. Now it also seems to open again in the current version of MuseScore: 
[Freude_schoener_GoetterfunkenMS4 copy.mscz.zip](https://github.com/musescore/MuseScore/files/10690471/Freude_schoener_GoetterfunkenMS4.copy.mscz.zip)

urbanrieger:
Great, this community is awesome. Can you somehow explain, what you did to fix ist?

cbjeukendrup:
I fixed the crash in the source code, and recompiled MuseScore. (So this is not something that regular users can do themselves.)

You can download the fixed version of MuseScore from https://github.com/musescore/MuseScore/pull/16284, following the instructions at https://github.com/musescore/MuseScore/wiki/Downloading-and-running-test-builds#Downloading-Builds-from-Pull-Requests.

DmitryArefiev:
Fixed in #16284

DmitryArefiev:
Retested on Win10

Issue #14303, opened 2022-11-04
[MU4 Issue] "Replace instrument" without choosing instrument makes staff unusable


rgreen5:
**Describe the bug**

When you use the "Replace instrument" option in Staff properties" it is possible to leave the dialog without having chosen an instrument. This causes obvious problems when you return to the staff.

**To Reproduce**

1. Open the attached example score. 
[change_instrument_issue.zip](https://github.com/musescore/MuseScore/files/9937708/change_instrument_issue.zip)

2. Right click on the staff and select staff properties.
3. Click on "Strings plucked" then click OK without selecting an instrument.

RESULT: Fretmarks turn red and playback muted. i.e. the score is unusable. AFAICS, it can be fixed by repeating the operation _correctly_.

**Expected behavior**

Unless the operation is fully completed there should be no change to the staff.

**Platform information**

 OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223080420, revision: github-musescore-musescore-2c34155

Tantacrul:
Haha. Nice catch!

Confirmed. It's possible to confirm the instruments popup without selecting an instrument. I'm going to move this one to Post Release.

MarcSabatella:
As reported in the forum, if you save a score that has had its instrument removed in this fashion, the score crashes on open.  It would be nice to fix both the original problem where this can happen in the first place, and to somehow allow the score to open and assign an instrument to the staff (even just arbitrarily making it piano or whatever, if there is no other information to go on).

DmitryArefiev:
Fixed in #16319

PR #16284, opened 2023-02-08
Fix crash when opening score with instruments without instrumentId and without InstrChannels

cbjeukendrup:
Resolves: #16248
Resolves: #16235

Maybe a candidate for the patch release? Two reports already.

### DIFF
--- a/src/engraving/libmscore/instrument.h
+++ b/src/engraving/libmscore/instrument.h
@@ -361,8 +361,8 @@ public:
     void setUseDrumset(bool val);
     void setAmateurPitchRange(int a, int b) { _minPitchA = a; _maxPitchA = b; }
     void setProfessionalPitchRange(int a, int b) { _minPitchP = a; _maxPitchP = b; }
-    InstrChannel* channel(int idx) { return _channel[idx]; }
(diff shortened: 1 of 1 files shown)


cbjeukendrup:
Now that I think a bit more about it, I'm wondering how it is possible that these instruments without an ID got written to a 4.0.0 file. You'd say that the ID should be determined when opening the score in 4.0.0+, and then it should be written to the file.
Anyway, I still think this fix would make sense...

MarcSabatella:
Perhaps the ID was lost as a result of trying to replace the instrument from 4.0, then backing out of the dialog?  See https://github.com/musescore/MuseScore/issues/14303 for a report of this losing the mixer channel immediately, and a reference to another report in the forum of this then leading to a crash on save/reload.

cbjeukendrup:
Thanks, that sounds very likely. I'll take a look at that issue too.

RomanPudashkin (approved):

Tantacrul:
@RomanPudashkin - if you're happy, I'm happy to include in the patch release.

DmitryArefiev:
Tested #16248 #16235 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #14878, opened 2022-12-01
[MU4 Issue] Plops, doits, scoops, and falls are scaled incorrectly


erinic04:
**Describe the bug**
Plops, doits, scoops, and falls are scaled incorrectly from the staff space dialog in page settings.

**To Reproduce**
1. Add any of these symbols to a note
2. Make the scaling smaller
3. See that they are too thin relative to everything around them

**Expected behavior**
They should be scaled the same as everything else

**Screenshots**
Normal scaling:
<img width="715" alt="Screen Shot 2022-12-01 at 11 52 22 AM" src="https://user-images.githubusercontent.com/21023353/205136033-19a8b81b-7747-4b91-a9da-453b00e7e396.png">
0.030 in scaling:
<img width="702" alt="Screen Shot 2022-12-01 at 11 53 17 AM" src="https://user-images.githubusercontent.com/21023353/205136199-d2a99ec4-9569-45ce-a2b6-4c8b43be23fa.png">

**Platform information**
 - OS: Reported on Mac

**Additional context**
I also see that the ledger lines are too small, but I think that's already been reported recently.

moogly81:
I agree. We should be able to customize the fall, bend, plop doit, ...
- thickness 
- shape (straight or wavy )

in the properties panel. Falls should have the same properties as glissandi.

PR #16292, opened 2023-02-09
Fix chord line scaling

mike-spa:
Resolves: #14878 

The scaling calculations for ChordLine objects were an utter mess. And instead of fixing the scaling, we were using a style setting for line thickness (not exposed to UI, but still there) of 6/1000 of a space (?!?!?!?). Hopefully, this should clean it up.


RomanPudashkin (approved):

RomanPudashkin:
@mike-spa please rebase it

PR #16296, opened 2023-02-09
Fix MSVC compiler warnings

Jojo-Schmitz:
* reg. conversion from 'float' to '_Rx', possible loss of data (C4244)


Jojo-Schmitz:
I don't like this fix too much, review definitly needed here, esp. from @handrok (as it is his code)

igorkorsukov (approved):

PR #16302, opened 2023-02-09
Fix some text clipping issues

cbjeukendrup:
Resolves: #13773 (the three problems mentioned there that don't have a separate issue yet)

- Cloud upload success dialog
- New workspace dialog
- Barline inspector (in fact, more generally: elide text on buttons if it's too wide, instead of overflowing beyond the border of the button)

### DIFF
--- a/src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
+++ b/src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
@@ -230,8 +230,9 @@ Item {
                 }
 
                 contentItem: StyledTextLabel {
-                    height: root.height
+                    // To do: this causes a binding loop warning
                     width: titleLoader.width - 8 // 4px padding on each side
+                    height: root.height
 
                     text: root.title
                 }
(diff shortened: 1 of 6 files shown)


src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
```diff
@@ -234,6 +234,10 @@ Item {
                     width: titleLoader.width - 8 // 4px padding on each side
 
                     text: root.title
+
+                    onImplicitWidthChanged: {
+                        let i = implicitWidth // workaround for strange binding loop warning
```
Eism, line 239: How often does this warning appear?

Because the workaround looks strange. I think either you need to deal with the warning or leave it as it is, maybe someone will find a better solution.

Eism (approved):

src/playback/qml/MuseScore/Playback/internal/AudioResourceControl.qml
cbjeukendrup: It appears not crazily annoyingly often; just sometimes when a Mixer channel is loaded in the view. I've removed the workaround now in the hope that the warning inspires someone, and if it becomes too annoying we can add back the workaround later.

abariska:
🟢 **TESTED on macOS 12**
Please, merge to master

PR #16315, opened 2023-02-11
Fix #16215: Export filetype and directory not memorised

### LINKED ISSUE #16215: [MU4 Issue] Export filetype and directory not memorized
**Describe the bug**

When exporting a file several times in a row, the format is always reset to PDF, and the directory (for the filepicker dialog)is always reset to the directory containing the current `.mscz` file. This makes it needlessly difficult to re-export a file over and over again, for example when making slight incremental adjustments and seeing how the output has changed. (In my case: I wanted to tweak the score such that the generated FLAC-file has just the right length for a particular usage.)

**To Reproduce**
1. Open MuseScore
2. Write a few notes (probably optional, but just in case)
3. In the menu, go to File > Export, this should open a dialog. In the dialog, select a different format (e.g. FLAC), click "export" and select a different folder (e.g. `/tmp/foobar.flac`)
4. Change/add/delete a note (probably optional; included only for realism)
5. Repeat Step 3

**Expected behavior**
The second time the export dialog opens (specifically step 5), MuseScore remembers that I just exported the score as FLAC, and the format defaults to FLAC. Likewise, it should remember which directory was just used, and suggest that directory in the filepicker.

**Actual behavior**
The second time the export dialog opens (specifically step 5), MuseScore behaves just like the first time: I have to select the FLAC format (otherwise the filepicker suggests a PDF filename extension), and I have to manually navigate to the same destination folder again.

Curiously, if I select PDF format and a destination file like `/tmp/weird.flac`, then MuseScore generates a proper FLAC, as if I had selected flac in the dialog. Weird.

**Screenshots**
This is the "export dialog" I mean:
Bildschirmfoto_2023-02-04_23-21-57

**Platform information**
 - OS: Linux, x64, German locale (I don't think any of these seem relevant here though)
 - Installation method: AppImage
 - Version: MuseScore-4.0.1.230121751-x86_64.AppImage

HemantAntony:
Resolves: #16215 

### DIFF
--- a/src/importexport/audioexport/iaudioexportconfiguration.h
+++ b/src/importexport/audioexport/iaudioexportconfiguration.h
@@ -35,7 +35,8 @@ public:
     virtual ~IAudioExportConfiguration() = default;
 
     virtual int exportMp3Bitrate() const = 0;
-    virtual void setExportMp3Bitrate(std::optional<int> bitrate) = 0;
+    virtual void setExportMp3Bitrate(int bitrate) = 0;
+    virtual void setExportMp3BitrateOverride(std::optional<int> bitrate) = 0;
     virtual const std::vector<int>& availableMp3BitRates() const = 0;
 
     virtual int exportSampleRate() const = 0;
--- a/src/importexport/audioexport/internal/audioexportconfiguration.cpp
+++ b/src/importexport/audioexport/internal/audioexportconfiguration.cpp
@@ -28,22 +28,27 @@ using namespace mu::iex::audioexport;
 using namespace mu::framework;
 
 static const Settings::Key EXPORT_SAMPLE_RATE_KEY("iex_audioexport", "export/audio/sampleRate");
-
-static constexpr int DEFAULT_BITRATE = 128;
+static const Settings::Key EXPORT_MP3_BITRATE("iex_audioexport", "export/audio/mp3Bitrate");
 
 void AudioExportConfiguration::init()
 {
     settings()->setDefaultValue(EXPORT_SAMPLE_RATE_KEY, Val(44100));
+    settings()->setDefaultValue(EXPORT_MP3_BITRATE, Val(128));
 }
 
 int AudioExportConfiguration::exportMp3Bitrate() const
 {
-    return m_exportMp3Bitrate ? m_exportMp3Bitrate.value() : DEFAULT_BITRATE;
+    return m_exportMp3BitrateOverride ? m_exportMp3BitrateOverride.value() : settings()->value(EXPORT_MP3_BITRATE).toInt();
 }
 
-void AudioExportConfiguration::setExportMp3Bitrate(std::optional<int> bitrate)
+void AudioExportConfiguration::setExportMp3Bitrate(int bitrate)
 {
-    m_exportMp3Bitrate = bitrate;
+    settings()->setSharedValue(EXPORT_MP3_BITRATE, Val(bitrate));
+}
+
+void AudioExportConfiguration::setExportMp3BitrateOverride(std::optional<int> bitrate)
+{
+    m_exportMp3BitrateOverride = bitrate;
 }
 
 const std::vector<int>& AudioExportConfiguration::availableMp3BitRates() const
--- a/src/project/projecttypes.h
+++ b/src/project/projecttypes.h
@@ -31,7 +31,9 @@
 #include "log.h"
 
 #include "cloud/cloudtypes.h"
+#include "notation/inotation.h"
 #include "notation/notationtypes.h"
+#include "inotationwriter.h"
 
 namespace mu::project {
 struct ProjectCreateOptions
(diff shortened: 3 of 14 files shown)


src/project/inotationproject.h
```diff
@@ -51,6 +51,9 @@ class INotationProject
     virtual const CloudProjectInfo& cloudInfo() const = 0;
     virtual void setCloudInfo(const CloudProjectInfo& info) = 0;
 
+    virtual ExportInfo exportInfo() const = 0;
```
Eism, line 54: I think we can store this information in ExportProjectScenario instead of NotationProject
And it might be like 
std::map<path /* projectPath*/, ExportInfo>

src/project/projecttypes.h
```diff
@@ -88,6 +89,30 @@ struct CloudProjectInfo {
     }
 };
 
+struct ExportInfo {
+    ExportType type;
```
cbjeukendrup, line 93: Do we really need to store a copy of the whole ExportType? Or is storing the ID enough? (Note the existence of the method `ExportDialogModel::selectExportTypeById`)

src/project/projecttypes.h
```diff
@@ -88,6 +88,35 @@ struct CloudProjectInfo {
     }
 };
 
+struct ExportInfo {
+    QString id;
+    io::path_t path;
+
+    ExportInfo() {}
```
cbjeukendrup, line 95: Is there any reason to not write `ExportInfo() = default;`? And why are the copy constructor and `operator=(…)` defined explicitly?

src/project/view/exportdialogmodel.cpp
```diff
@@ -234,7 +243,11 @@ void ExportDialogModel::setExportType(const ExportType& type)
     }
 
     m_selectedExportType = type;
-    emit selectedExportTypeChanged(type.toMap());
+    m_exportPath = "";
+
+    QMetaObject::invokeMethod(qApp, [this, type]() {
+        emit selectedExportTypeChanged(type.toMap());
+    }, Qt::QueuedConnection);
```
cbjeukendrup, line 250: I wonder if it's still necessary to use `QMetaObject` etc here, now that https://github.com/musescore/MuseScore/pull/16837 has been merged

src/importexport/audioexport/internal/audioexportconfiguration.cpp
cbjeukendrup, line 1: The point of the use of `std::optional` here, is that when options are specified *via the command line*, then they should apply to the current run only. So what we really need is:
1. A method to set the bit rate in a way that is persisted. This method should take just `int`, and store it to `settings()`. This method will be used from the UI.
2. A method to _override_ the bit rate for the current run. This method should take `std::optional<int>` and store it in a `std::optional<int>` field in the `AudioExportConfiguration` class. This method will be used from the command line (see `App::applyCommandLineOptions`).
3. A method to _retrieve_ the current bit rate. This method should return `int`; if the field that stores the override  value has a value, then it should return the override value; otherwise, it should return the value from `settings()`. 

The same applies to those things in `ImagesExportConfiguration`.

src/project/view/exportdialogmodel.cpp
```diff
@@ -306,7 +319,19 @@ void ExportDialogModel::setUnitType(UnitType unitType)
...
+            found = true;
+        }
+    }
+
+    if (!found) {
+        return;
+    }
+
     m_selectedUnitType = unitType;
+    configuration()->setExportUnitType(unitType);
```
cbjeukendrup, line 334: I'd think the unit type should get the same treatment as the selected export file type, right?

HemantAntony: Yeah that makes sense

src/project/projecttypes.h
HemantAntony: I was getting a bunch of warnings and errors at first. So when I added them, it solved them. But now it works fine without those functions

src/importexport/audioexport/iaudioexportconfiguration.h
```diff
@@ -35,6 +35,7 @@ class IAudioExportConfiguration : MODULE_EXPORT_INTERFACE
     virtual ~IAudioExportConfiguration() = default;
 
     virtual int exportMp3Bitrate() const = 0;
+    virtual void setExportMp3Bitrate(int bitrate) = 0;
     virtual void setExportMp3Bitrate(std::optional<int> bitrate) = 0;
```
cbjeukendrup, line 39: Now I think it would be nice to rename this method, to make the difference clear. For example, something like `setExportMp3BitrateOverride` will do.

(And then I'd rename `m_exportMp3Bitrate` too, to `m_exportMp3BitrateOverride`.)

(And the same for the png resolution.)

DmitryArefiev:
Tested #16215 on Win10, Mac13, LinuxUbuntu 22.04 - **FIXED**

Eism (approved):

PR #15730, opened 2023-01-05
Fixed off by one error which incorrectly prohibits a tab note moving …

pkeir:
…to a string at the highest fret.

Resolves: #15729 which was also discussed [here](https://musescore.org/en/node/341410).

Issue #15729, opened 2023-01-05
 Tablature note string change request has no effect


pkeir:
Steps to reproduce the issue:

1) Create a tablature staff for a stringed instrument with N frets.
2) Add a note at fret N, on a string which is not the highest in pitch.
3) Select the note, and press CTRL+UP. The note moves to the higher string, as expected.
4) Select the note moved in step 3, and press CTRL+DOWN. The note does not move to the lower string. This is unexpected.

For example, consider a note at fret 19 on the high 'E' string of a 24-fret electric guitar. The note should also be playable at fret 24 of the 'B' string, but the note cannot be moved using CTRL+DOWN.

I am using a recent build of 4.x from the github project's master branch. This also affects MuseScore 3.
OS: Debian GNU/Linux 11 (bullseye), Arch.: x86_64, MuseScore version (64-bit): 4.0.0-, revision: github-musescore-musescore-

The issue was first discussed [here](https://musescore.org/en/node/341410)

PR #16325, opened 2023-02-12
Fix #12437: TAB: Cannot place note on highest fret using Ctrl+Up/Down

### LINKED ISSUE #12437: [MU4 Issue] TAB: Cannot place note on last fret using Ctrl-Up/Down
**Describe the bug**
As described in https://musescore.org/en/node/333596, on guitar tablature, notes cannot be placed on the highest fret of a given string without first entering a lower note and using the arrow keys to move it up to the highest fret.

**To Reproduce**
Using the default 24 fret electric guitar as an example,
1. Create a treble 8vb electric guitar staff
2. Add a linked tablature staff
4. Add a note two octaves above the base pitch of a string (e.g. E4) to the treble staff
5. Attempt to move the note downwards using Ctrl+Down so it is on the 24th fret of a string
6. Note that the note will not go lower than the 19th fret of the string above
7. Change the tablature staff to have 25 frets
8. Repeat step 5, noting that the E4 will now move onto the 24th fret of the low E string

**Expected behavior**
The note should continue to be able to be moved onto the highest fret of the string for every string.

**Screenshots**
https://www.youtube.com/watch?v=j0GdnPhJyCw

**Platform information**
OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-2674626696, revision: github-musescore-musescore-abc123456

**Additional context**
Present all the way back to 2.0's initial tablature implementation.

cbjeukendrup:
Resolves: #12437

### DIFF
--- a/src/engraving/libmscore/stringdata.cpp
+++ b/src/engraving/libmscore/stringdata.cpp
@@ -445,7 +445,7 @@ int StringData::fret(int pitch, int string, int pitchOffset) const
     }
 
     // fret number is invalid or string cannot be fretted
-    if (fret < 0 || fret >= _frets || (fret > 0 && strg.open)) {
+    if (fret < 0 || fret > _frets || (fret > 0 && strg.open)) {
         return INVALID_FRET_INDEX;
     }
     return fret;


cbjeukendrup:
It turns out that the issue that this is solving has been duplicated by https://github.com/musescore/MuseScore/issues/15729, which has been fixed by https://github.com/musescore/MuseScore/pull/15730.

Issue #12804, opened 2022-08-11
[MU4 Issue] appstream file out of date


hfiguiere:
**Describe the bug**
The appstream file in `build/Linux+BSD/org.musescore.MuseScore.appdata.xml.in` is out of date. The last release it lists in 3.3.4. Not even 3.6.2

**To Reproduce**

n/a

**Expected behavior**

Expect to have the proper version

**Screenshots**

n/a

**Platform information**

 - OS: Linux. But literally n/a
 
**Additional context**

This mean that the information will be wrong in GNOME Software or whatever equivalent in a Desktop Environment.

This should be updated at release time or something.

PR #16327, opened 2023-02-12
Update AppStream file

cbjeukendrup:
Resolves: #12804


shoogle (approved):

Jojo-Schmitz:
Looks like harmless enough and needed for 4.0.2 too (including the 4.0.2 update), doesn't it?

Issue #13655, opened 2022-10-07
[MU4 Issue] Fretboard diagrams: "Orientation" property missing


rgreen5:
**Describe the bug**

For __fretboard diagrams__ MS3 had a "Orientation" property in the Inspector, which allowed you to set them horizontally.  See https://musescore.org/en/node/299644.

![fretboard_orientation](https://user-images.githubusercontent.com/87246583/194514327-4ce95d93-23bc-4b03-9937-a96507d19338.png)

However this is missing in MS4, and needs to be restored.

rgreen5:
Bump. See https://musescore.org/en/node/338158.

rgreen5:
Bump. Surely a missing property should receive some priority?

cbjeukendrup:
@rgreen5 Please check https://github.com/musescore/MuseScore/pull/16331 :)

PR #16331, opened 2023-02-12
Fret diagram Inspector: bring back missing Orientation property

cbjeukendrup:
Resolves: #13655

Scherm­afbeelding 2023-02-12 om 15 12 03

@bkunda Could you please check if you agree with this design? I also moved some other controls around a little bit:
- moved the "show nut" checkbox closer to the spin boxes, because all these things are about the "contents" of the diagram
- moved the "clear" button below the diagram preview, because it looked strange to have it immediately below the "orientation" property (it almost looked like it belonged to that property).

### DIFF
--- a/src/inspector/models/notation/fretdiagrams/fretdiagramsettingsmodel.cpp
+++ b/src/inspector/models/notation/fretdiagrams/fretdiagramsettingsmodel.cpp
@@ -21,8 +21,10 @@
  */
 #include "fretdiagramsettingsmodel.h"
 
-#include "translation.h"
 #include "dataformatter.h"
+#include "translation.h"
+
+#include "engraving/libmscore/fret.h"
 
 using namespace mu::inspector;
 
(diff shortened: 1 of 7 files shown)


bkunda:
@cbjeukendrup I'm fine with most of your suggestions. I'd just swap the position of ‘Orientation‘ and ‘Placement on stave‘. ‘Orientation‘ has more to do with how the diagram content appears (including the relationship between the finger positions and the nut), whereas ‘Placement on stave‘ is more an engraving setting concerning the overall object's relationship with the stave. 

Actually, it should probably just be called ‘Position‘, to make it consistent with other elements that have the same setting: 

Screenshot 2023-02-15 at 4 40 39 pm

On a related front, looking at this has made me realise that there are some problems with the design of the ‘General‘ section as well. But perhaps it's best for me to raise this as a separate issue.

cbjeukendrup:
@bkunda Thanks! I made these changes. I noticed that "Placement on staff" was also used for fermatas, so I changed it there too so that we have "Position" everywhere.

Eism (approved):

Eism:
@cbjeukendrup 
Please rebase it

PR #16335, opened 2023-02-12
[WIP] Retranslating the UI without restarting

### LINKED ISSUE #12379: [MU4 Issue] Restart required to change language
**Describe the bug**
When you change the language in Preferences, only this dialog's strings are being updated. The rest of the software does not update its strings. However, once you restart MuseScore, all strings get updated

**To Reproduce**
1. Go to Preferences
2. Click on the dropdown
3. Select German (Deutsch)
4. Only the dialog's strings get updated

**Expected behavior**
All strings should be updated

**Screenshots**

https://user-images.githubusercontent.com/88136054/178742109-ad6ff7f4-fa7d-4ec8-a4b1-d9790d0db6d1.mp4

**Platform information**
 - OS: Windows 11

**Additional context**
I'm running MuseScore on Qt Creator

cbjeukendrup:
Resolves: https://github.com/musescore/MuseScore/issues/12379

There were two problems: 
- If we call the QQmlEngine's retranslateUi method, literally every property binding in the whole app is re-evaluated, regardless of whether it has anything to do with strings or not. This is quite inefficient (duh) and causes problems too. So, we can't use this.
- Things in/from C++ also need to be retranslated. 

So, we solve this in two steps:
- [x] Re-evaluate QML bindings that make use of `qsTrc`

    The whole logic now looks like this:
    - We have the `QmlTranslation` object (which subclasses `QObject`) with `translate` as a `Q_INVOKABLE` method 
    - That allows us to get the `"translate"` function as a `QJSValue`, namely by querying it as a property of that `QObject`
    - We use this `QJSValue` to offer the translation function as a `Q_PROPERTY` of the `UiEngine`, with a `NOTIFY` signal. When we want the UI to be retranslated, we simply emit that signal, and all property bindings that use the translation function will be re-evaluated.
    - However, because the translation function is now a property of the `UiEngine` object (which is available in QML as `ui`), we now need to call the translation function as `ui.trc(…)`. This means that we have to replace 1036 calls to `qsTrc`, resulting in a huge diff. However, I can easily be convinced that we should just make this change and don't worry about a big diff.
    - So, I created another `QJSValue` which is a JS function, which is simply a wrapper around `ui.trc`. We make this function available under the name `qsTrc`. Apparently, QML looks inside this JS function, and sees that we use `ui.trc`, so bindings using `qsTrc` are now also re-evaluated when the `trc` property of `ui` is changed.

    See the commit message for _why_ it is now done like this, and not slightly differently.

- [ ] In C++, retranslate things and notify QML about that
    ~I'm planning to create a `Translatable` interface that any class can subclass; this interface will contain a single method `retranslate()`. This method will be called automagically when the UI needs to be retranslated.~

    Methods/classes/modules to do:
    - [ ] AppearancePreferencesModel::wallpaperPathFilter()
    - [ ] GeneralPreferencesModel::languages()
    - [ ] ImportPreferencesModel::shortestNotes()
    - [ ] ProgrammeStartPreferencesModel::scorePathFilter()
    - [ ] instrtemplate.cpp
    - [ ] StaffType names
    - [ ] Available audio/midi devices
    - [ ] MidiDeviceMappingModel::midiMappingToObject
    - [ ] NavigationPanel::directionInfo()
    - [ ] Importmidi module
    - [ ] Inspector model titles
    - [ ] GradualTempoChangePlaybackModel::possibleEasingMethods()
    - [ ] GlissandoSettingsModel::possibleLineTypes()
    - [ ] OttavaSettingsModel hook types
    - [ ] PedalSettingsModel hook types
    - [ ] SlurAndTieSettingsModel::possibleLineStyles()
    - [ ] TextLineSettingsModel hook types
    - [ ] VibratoSettingsModel::possibleLineTypes()
    - [ ] VoltaSettingsModel hook types
    - [ ] NoteheadSettingsModel::possibleHeadSystemTypes()
    - [ ] TupletSettingsModel
    - [ ] InstrumentListModel
    - [ ] StaffControlTreeItem
    - [ ] StaffTreeItem::init
    - [x] LearnPageModel::classesAuthor()
    - [ ] NotationAccessibility (see also NotationActionController::toggleNoteInput())
    - [ ] EditStyle (currently not necessary)
    - [ ] SearchCommandsParser::availableCommands()
    - [ ] formatInstrumentTitle
    - [ ] NotationSwitchListModel::contextMenuItems
    - [ ] Timeline
    - [ ] Palette module
        - [ ] SpecialCharactersDialog
        - [ ] MasterPalette::addPalette
        - [ ] Palette::translatedName()
        - [ ] PaletteCell::translatedName() / retranslate()
        - [ ] PaletteElementEditor::actionName()
    - [ ] PlaybackController::addTrack
    - [ ] PlaybackController::setupSequenceTracks()
    - [ ] inputresourceitem.cpp SOUNDFONTS_MENU_ITEM_ID
    - [ ] MixerChannelItem::MixerChannelItem
    - [ ] NO_FX_MENU_ITEM_ID()
    - [ ] MixerPanelModel::buildMasterChannelItem()
    - [ ] PluginsModel::data
    - [ ] TemplatesRepository::readTemplates
    - [ ] AdditionalInfoModel (currently not necessary)

Note: there _will_ be bugs, inevitably. But I think if we can get 90% there, it's worth it, and the remaining bugs will be corrected one by one.

### DIFF
--- a/src/framework/ui/internal/uiengine.cpp
+++ b/src/framework/ui/internal/uiengine.cpp
@@ -106,12 +106,16 @@ void UiEngine::setup(QQmlEngine* engine)
 
     m_engine = engine;
     m_theme->init();
-    m_engine->rootContext()->setContextProperty("ui", this);
+    QJSValue ui = m_engine->newQObject(this);
+    QQmlEngine::setObjectOwnership(this, QQmlEngine::CppOwnership);
+    m_engine->globalObject().setProperty("ui", ui);
+
     m_engine->rootContext()->setContextProperty("api", m_api);
 
     QJSValue translator = m_engine->newQObject(m_translation);
-    QJSValue translateFn = translator.property("translate");
-    m_engine->globalObject().setProperty("qsTrc", translateFn);
+    QQmlEngine::setObjectOwnership(m_translation, QQmlEngine::CppOwnership);
+    m_translationFunction = translator.property("translate");
+    m_engine->globalObject().setProperty("qsTrc", m_translationFunction);
 
 #ifdef Q_OS_WIN
     QDir dir(QCoreApplication::applicationDirPath() + QString("/../qml"));
@@ -140,13 +144,9 @@ void UiEngine::addSourceImportPath(const QString& path)
 #endif
 }
 
-void UiEngine::updateTheme()
+void UiEngine::retranslateUi()
 {
-    if (!m_engine) {
-        return;
-    }
-
-    theme()->update();
+    emit translationFunctionChanged();
 }
 
 QmlApi* UiEngine::api() const
@@ -162,6 +162,11 @@ UiTheme* UiEngine::theme() const
 QmlToolTip* UiEngine::tooltip() const
 {
     return m_tooltip;
+}
+
+QJSValue UiEngine::translationFunction() const
+{
+    return m_translationFunction;
 }
 
 InteractiveProvider* UiEngine::interactiveProvider_property() const
--- a/src/framework/ui/internal/uiengine.cpp
+++ b/src/framework/ui/internal/uiengine.cpp
@@ -115,12 +115,17 @@ void UiEngine::setup(QQmlEngine* engine)
     QJSValue translator = m_engine->newQObject(m_translation);
     QQmlEngine::setObjectOwnership(m_translation, QQmlEngine::CppOwnership);
     m_translationFunction = translator.property("translate");
-    m_engine->globalObject().setProperty("qsTrc", m_translationFunction);
+
+    QJSValue qsTrc = m_engine->evaluate(
+        "(function(context, text, disambiguation = \"\", n = -1) {"
+        "    return ui.trc(context, text, disambiguation, n);"
+        "})");
+    m_engine->globalObject().setProperty("qsTrc", qsTrc);
 
 #ifdef Q_OS_WIN
     QDir dir(QCoreApplication::applicationDirPath() + QString("/../qml"));
     m_engine->addImportPath(dir.absolutePath());
- #endif
+#endif
 
     m_engine->addImportPath(":/qml");
 
--- a/src/framework/ui/internal/uiengine.cpp
+++ b/src/framework/ui/internal/uiengine.cpp
@@ -122,6 +122,10 @@ void UiEngine::setup(QQmlEngine* engine)
         "})");
     m_engine->globalObject().setProperty("qsTrc", qsTrc);
 
+    languagesService()->currentLanguageChanged().onNotify(this, [this]() {
+        emit translationChanged();
+    });
+
 #ifdef Q_OS_WIN
     QDir dir(QCoreApplication::applicationDirPath() + QString("/../qml"));
     m_engine->addImportPath(dir.absolutePath());
@@ -147,11 +151,6 @@ void UiEngine::addSourceImportPath(const QString& path)
 #else
     Q_UNUSED(path);
 #endif
-}
-
-void UiEngine::retranslateUi()
-{
-    emit translationFunctionChanged();
 }
 
 QmlApi* UiEngine::api() const
--- a/src/framework/ui/internal/uiengine.h
+++ b/src/framework/ui/internal/uiengine.h
@@ -49,6 +49,8 @@ class UiEngine : public QObject, public IUiEngine
 
     Q_PROPERTY(QQuickItem * rootItem READ rootItem WRITE setRootItem NOTIFY rootItemChanged)
 
+    Q_PROPERTY(QJSValue trc READ translationFunction NOTIFY translationFunctionChanged)
+
     // for internal use
     Q_PROPERTY(InteractiveProvider * _interactiveProvider READ interactiveProvider_property CONSTANT)
 
@@ -60,6 +62,7 @@ public:
     QmlApi* api() const;
     UiTheme* theme() const;
     QmlToolTip* tooltip() const;
+    QJSValue translationFunction() const;
     InteractiveProvider* interactiveProvider_property() const;
     std::shared_ptr<InteractiveProvider> interactiveProvider() const;
 
@@ -67,10 +70,11 @@ public:
     Q_INVOKABLE Qt::LayoutDirection currentLanguageLayoutDirection() const;
 
     // IUiEngine
-    void updateTheme() override;
     QQmlEngine* qmlEngine() const override;
     void clearComponentCache() override;
     void addSourceImportPath(const QString& path) override;
+
+    void retranslateUi() override;
     // ---
 
     void moveQQmlEngine(QQmlEngine* e);
@@ -86,6 +90,8 @@ signals:
 
     void rootItemChanged(QQuickItem* rootItem);
 
+    void translationFunctionChanged();
+
 private:
     UiEngine();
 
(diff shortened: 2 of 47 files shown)


Jojo-Schmitz:
Can this get triggered also on/after a translations update?

cbjeukendrup:
Yes, I'll add that too 👍

Jojo-Schmitz:
Excellent, that did annoy me in 3.x quite a bit, esp. during the translation work

Jojo-Schmitz:
It BTW never really 100% worked, there were some corners of the UI that did not update.

src/framework/ui/internal/uiengine.cpp
```diff
@@ -107,17 +107,27 @@ void UiEngine::setup(QQmlEngine* engine)
...
     QJSValue translator = m_engine->newQObject(m_translation);
     QJSValue translateFn = translator.property("translate");
-    m_engine->globalObject().setProperty("qsTrc", translateFn);
+    m_engine->setObjectOwnership(m_translation, QQmlEngine::CppOwnership);
+    m_translationFunction = QVariant::fromValue(translateFn);
+
+    QJSValue qsTrc = m_engine->evaluate(
+        "(function(context, text, disambiguation = \"\", n = -1) {"
+        "    return ui.trc(context, text, disambiguation, n);"
+        "})");
```
igorkorsukov, line 124: I also thought about something similar, but I didn't know that we can call properties with arguments, it's cool!

src/framework/ui/internal/uiengine.cpp
```diff
@@ -107,17 +107,27 @@ void UiEngine::setup(QQmlEngine* engine)
...
+    QQmlEngine::setObjectOwnership(this, QQmlEngine::CppOwnership);
+    m_engine->globalObject().setProperty("ui", ui);
+
     m_engine->rootContext()->setContextProperty("api", m_api);
 
     QJSValue translator = m_engine->newQObject(m_translation);
     QJSValue translateFn = translator.property("translate");
-    m_engine->globalObject().setProperty("qsTrc", translateFn);
+    m_engine->setObjectOwnership(m_translation, QQmlEngine::CppOwnership);
+    m_translationFunction = QVariant::fromValue(translateFn);
```
igorkorsukov, line 119: Is it necessary to use `QVariant` here, can't we use `QJSValue`?

src/framework/ui/internal/uiengine.h
```diff
@@ -45,6 +45,8 @@ class UiEngine : public QObject, public IUiEngine
 
     Q_PROPERTY(QQuickItem * rootItem READ rootItem WRITE setRootItem NOTIFY rootItemChanged)
 
+    Q_PROPERTY(QVariant trc READ translationFunction NOTIFY translationFunctionChanged)
```
igorkorsukov, line 48: I think better rename `translationFunctionChanged` to just `translationChanged`

igorkorsukov (approved):

cbjeukendrup:
I have made some C++ classes retranslatable. My idea with creating a `Translatable` interface/superclass turned out to be impossible for two reasons:
- To make the class able to listen for language changes, it needs to be a subclass of `async::Asyncable`, but many classes that would inherit from `Translatable` are already `async::Asyncable` themselves too. That makes the conversion to `async::Asyncable` ambiguous, somewhat similar to the [diamond problem](https://en.wikipedia.org/wiki/Virtual_inheritance).
   (I think that removing `Asyncable` as a direct superclass (so that it is only inherited via `Translatable`) is not clean. Maybe it is possible to hack around it with `virtual` inheritance, but that doesn't make the code more elegant; I also thought maybe [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) can help but then it is not elegant anymore either, so that defeats the purpose.)
- I would add the code to subscribe to the notification in the constructor of `Translatable`, but of course in some cases the `languagesService()` is not even created yet when the constructor is called, so that doesn't work. So we would still need to subscribe to the notification manually, which also drastically reduces the usefulness of the `Translatable` class.

So, I just connect each class manually to the notification, and that turns out to be not very painful.

I think that this retranslates now ca. 70% of the UI. Since 4.1 is still months away apparently, I think it might be nice to merge this now, and ask community members to help perfecting it over time. I have added the whole to-do list (as far as it is known) to the PR description. The individual tasks are not difficult and might be nice for GSoC students too.

@igorkorsukov @RomanPudashkin What do you think about this approach?

igorkorsukov:
@cbjeukendrup I would return to the question, why are we doing this? What problem are we solving for the user?  
I don't see any problem in restart the application. Maybe add a "Restart" button for convenience.

Jojo-Schmitz:
* It is cumbersome, esp. in giving User support on the forums, where switching to the user's language is quite helpful.
* Also helpfull while translating, after having updated trsnalations.
* It wasn't needed in Mu3.

igorkorsukov:
> * It is cumbersome, esp. in giving User support on the forums, where switching to the user's language is quite helpfull.
> * It wasn't needed in Mu3. So requiring a restart is a regression.
> * Also helpfull while translating, after having updated translations.
> 
> But all those arguments have been given in the issue, #12379, already (that issue lacks a `regression` tag though)

I believe that these reasons are not worth the effort we want to make (

But let's try
> To make the class able to listen for language changes, it needs to be a subclass of async::Asyncable...

It's easy to work around, we can create an intermediary to subscribe, like this:
```
Translatable::Translatable()
{
    languagesService()->currentLanguageChanged().onNotify(m_asyncable, [this]() {
        ...
    });
};
```
When Translatable is destroyed, m_asyncable will destroyed too and the callback will be removed.

> I would add the code to subscribe to the notification in the constructor of Translatable, but of course in some cases the languagesService() is not even created yet when the constructor is called, so that doesn't work. So we would still need to subscribe to the notification manually, which also drastically reduces the usefulness of the Translatable class.

I see two solutions.
1. We can change the IoC, add a service registration notice to it.
2. An even simpler solution, we can add, for example, to the Application class a notification that the application has been initialized, and accordingly subscribe to this notification and subscribe to languagesService in its processing, by this moment it should exist (or the language module is disabled :) )

igorkorsukov:
>  I think that this retranslates now ca. 70% of the UI. Since 4.1 is still months away apparently, I think it might be nice to merge this now, and ask community members to help perfecting it over time. I have added the whole to-do list (as far as it is known) to the PR description. The individual tasks are not difficult and might be nice for GSoC students too.

I think we need to merge this PR, but turn off the retranslate functional. Otherwise, I'm afraid there will be a lot of issues from users (including those who use nightly builds) that not the entire retranslated, will have to be explained to everyone many times why.

Jojo-Schmitz:
But with a disabled retranslate we won't get reports of those sections that don't retranslated and therefore need fixing

Jojo-Schmitz:
Can we have this being looked at again? Having this feature back is not just useful to users supporting ithers, but also quite useful in connection with smart/curly quotes (like on a German setup having English curly quotes for English lyrics). Haveing to restart MuseScore in between, twice, is quite annoying

cbjeukendrup:
When I've time I'll take a look again. I'll most likely have to start from scratch though, and the list of C++ models that need retranslation is of course totally outdated and incomplete now. Also, despite the big diff it would cause, we might want to replace the `qsTrc` context property with a proper QML singleton first, to get rid of all those qmllint warnings.

PR #16345, opened 2023-02-13
fix #16343 playback model outdated upon score creation

### LINKED ISSUE #16343: Playback time incorrect upon new score creation
**Describe the bug**
When creating a score from a template, initially the score duration is governed by the template duration and not the actual score length.
Discovered by @Eism when testing PR#16263 but affects 4.0.1 and master as well

**To Reproduce**
1. Create New Score
2. Choose the Treble Clef Template
3. Set number of measure to be anything but 48 (number of measures in the template). Verified for values 1, 4 and 60
4. Playback the score
5. Notice the playback time going on to 1:39 and then stopping, regardless of the actual score length.

**Expected behavior**
Newly created score is set with correct playback length.

**Platform information**
 - OS: All

**Additional context**
The playback time of 1:39 is derived from the chosen template:
99secs = 96sec + 3sec ring on = 92160 ticks (at 120 tempo) = 192 quarter notes = 48 measures in 4/4
Using a template with a different number of measures confirms this behavior.

jeetee:
Resolves: #16343

Creating a new score means the notation has changed for it. Which also means the playback model for it has changed. (compared to the template the score may have started from)

- [N/A] I created a unit test or vtest to verify the changes I made (if applicable)

### DIFF
--- a/src/notation/internal/notationplayback.cpp
+++ b/src/notation/internal/notationplayback.cpp
@@ -76,9 +76,11 @@ void NotationPlayback::init(INotationUndoStackPtr undoStack)
 
     m_playbackModel.load(score());
 
-    updateTotalPlayTime();
     m_playbackModel.dataChanged().onNotify(this, [this]() {
         updateTotalPlayTime();
+    });
+    undoStack->stackLockedChanged().onNotify(this, [this]() {
+        m_playbackModel.reload();
     });
 
     configuration()->isPlayRepeatsChanged().onNotify(this, [this]() {
--- a/src/notation/internal/inotationundostack.h
+++ b/src/notation/internal/inotationundostack.h
@@ -54,6 +54,7 @@ public:
     virtual void lock() = 0;
     virtual void unlock() = 0;
     virtual bool isLocked() const = 0;
+    virtual async::Notification stackLockedChanged() const = 0;
 
     virtual async::Notification stackChanged() const = 0;
     virtual async::Channel<ChangesRange> changesChannel() const = 0;
--- a/src/notation/internal/notationundostack.cpp
+++ b/src/notation/internal/notationundostack.cpp
@@ -140,7 +140,10 @@ void NotationUndoStack::lock()
         return;
     }
 
-    undoStack()->setLocked(true);
+    if (!isLocked()) {
+        undoStack()->setLocked(true);
+        notifyAboutLockChanged();
+    }
 }
 
 void NotationUndoStack::unlock()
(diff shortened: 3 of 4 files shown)


src/notation/internal/notationplayback.cpp
```diff
@@ -57,6 +57,7 @@ NotationPlayback::NotationPlayback(IGetScore* getScore,
 {
     notationChanged.onNotify(this, [this]() {
         updateLoopBoundaries();
+        m_playbackModel.reload();
```
jeetee, line 60: This works for the intent of fixing the referred bug, but not being familiar with the normal notification flow I wonder if it perhaps isn't overkill to reload the playbackModel on every notationChange?

cbjeukendrup: Normally, the playback model already updates after every command, based on info from the undo stack. The reason that this doesn't happen during score creation, is that the undo stack is locked (see NotationProject::loadTemplate), so the playback model gets no notifications from it.

So, we either need to make it so that the playback model is reloaded when the undo stack is unlocked, or we need to trigger a reload of the playback model manually after the score creation options have been applied.

jeetee:
Went with the undoStack lock changed notification

src/notation/internal/notationplayback.cpp
```diff
@@ -76,10 +76,12 @@ void NotationPlayback::init(INotationUndoStackPtr undoStack)
 
     m_playbackModel.load(score());
 
-    updateTotalPlayTime();
     m_playbackModel.dataChanged().onNotify(this, [this]() {
         updateTotalPlayTime();
     });
+    undoStack->stackLockedChanged().onNotify(this, [this]() {
```
Eism, line 82: I don't like this solution.
We should not add undostack locking and playback dependency. 

We also lock the undostack in other places, does this mean that we will update the playback every time? Strange

I think we should solve the initial problem - the playback is loaded before the initialization of the master score

jeetee: Then I'll likely have to bow out for now, not being familiar with the web of current notification flows (yet)

jeetee: But from what I could tell we only unlocked the undoStack via that interface during score creation (take not is it not the UndoStack itself, but the INotationUndoStack-wrapper around it.

jeetee:
Closing this PR as "bowing out" and giving the correct status impression for the linked issue so someone can come up with a better approach to fix this.

Issue #16368, opened 2023-02-15
[MU4 Issue] Textline regression: end text cannot be edited


bakajikara:
**Describe the bug**
Most of the textlines do not have a field to edit the end text in the properties. This was present in MuseScore 3.

**To Reproduce**
1. Create a textline
2. Select the line and see the properties panel
3. Text tab does not contain end text field

**Expected behavior**
The end text of the text line should be editable as in MuseScore 3.

**Platform information**
 - OS: Windows 11

Issue #14928, opened 2022-12-04
[MU4 Issue] Unable to suppress Continue Text for pedal lines


Phil-TA:
**Describe the bug**
Musescore 4 provides features for adding and removing text to and from lines (both at the beginning of a line, and at the continuation point where a line continues through a system break).  However, these features are not available for pedal lines.

**To Reproduce**
1. Create a new, blank score.
2. Arrange two systems of a few bars each.
3. In the last bar of the first system, add a Pedal line.
4. Extend the pedal line to continue into the first bar of the second system. The system now adds "(Ped)" continuation text to the line.
5. Select the Pedal line and open the Properties panel.

**Expected behavior**
The Properties panel should have a "Text" section providing access to features which would allow the selected pedal line's continuation text to be changed or removed if required.

**Actual behavior**
The Properties panel has no "Text" features available for editing text associated with the Pedal line. The continuation "(Ped)" text cannot therefore be removed.

**Screenshots**
N/A

**Platform information**
OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223330401, revision: github-musescore-musescore-3277328

**Additional context**
Perhaps this feature been intentionally removed from MuseScore 4?

oktophonie:
Pedal lines should have the same text settings as, for example, cresc/dim or gradual tempo changes:
![image](https://user-images.githubusercontent.com/20604032/206124521-de79e48b-59a3-4a05-a14c-e0e568b61262.png)
This enables not only suppressing the continuation but also other styles of pedal indication, such as just the 'P' symbol, or 'Ped' in normal text, etc. (The 'Show pedal symbol' checkbox will likely then become redundant.)

PR #16369, opened 2023-02-15
Textline properties: reintroduce missing text edit field

bakajikara:
Resolves: #14928 
Resolves: #16368 

I'm new to both the MuseScore code and QML, so please let me know if I'm doing something wrong!


bkunda:
I've had a look at this, and the solutions do now make the text settings more consistent with other text line objects, so this gets a tick from me!

Only thing is that, instead of the **reset** button for 'Beginning text', Text when continuing to a new system', and 'End text', we usually adopt the "three dots" menu icon with these options:

Screenshot 2023-02-21 at 3 13 02 pm

I don't see any reason why this wouldn't also apply to pedal lines:

Screenshot 2023-02-21 at 3 10 07 pm

Other than that, I think this is a good fix. Thank you!

bakajikara:
This is because there is no default setting for pedal line text style, which has not existed since before MuseScore 3.
This is probably intentional, since it is common to use pedal lines with different text in a single score.

bkunda:
@bakajikara I've just made a quick video to demonstrate the problems I'm having with this reset button (one purely visual, the other functional):

https://user-images.githubusercontent.com/86290556/225903336-55fea8ea-a5e8-4cab-988a-0921a33bdf55.mov

I should add that, if it's somehow not possible to save a default pedal style (I can't imagine why this wouldn't be possible, but that's what I gleaned from your message above), then we can just disable the context menu option "Save as default style for this score". 

But ideally, the "reset" button for pedal marks should just return the syntax that triggers the correct symbol (in just the same way that resetting any other text object returns the default text). 

Hope that makes sense!

RobFog:
A rebase is also needed.

bakajikara:
In my last commit, I added a pedal text style. This solves the visual and functional problems described above.

However, there is a problem with this:
Previously, MuseScore had no style setting for pedal text, and the default value for pedal text was an empty string.
So when loading a previous score, the empty strings of the pedal lines are replaced by the default pedal symbol string of the current style.
I am working on solving this problem but do not know how to do it. Can someone please help me?

cbjeukendrup:
@bakajikara I think there are two solutions:
1. (easy) Make the default value for the new style setting an empty string. Then we still provide the ability to change that default style as requested, but for the rest nothing changes. 
2. (more complicated) Handle this while reading older files. This requires us to update the file format number (which we need to do anyway for 4.1), so that you can detect when a file is older than the new number, and when that is the case you set text to an empty string unless something else is explicitly specified. However, since this requires the file format to be updated, which we haven't done yet, this is maybe not directly actionable for you.

bakajikara:
Thank you very much. I decided to use solution 1 for now. (But ideally I think we should use solution 2.)

@bkunda
I have solved the visual problem you mentioned. The functional problem will be solved by changing the default styles sometime before version 4.1 is released.

cbjeukendrup:
@bakajikara A rebase is needed.

cbjeukendrup:
@RomanPudashkin Could you please review it?

RomanPudashkin (approved):

RomanPudashkin:
@bakajikara could you please rebase it again? I'll merge it after that. Thanks!

RomanPudashkin:
@bakajikara could you please also open the same PR but for the 4.1 branch? Thanks!

bakajikara:
@RomanPudashkin Done!

RomanPudashkin:
@bakajikara please fix the failed unit test

RomanPudashkin:
@bakajikara the build errors are fixed here:
https://github.com/musescore/MuseScore/pull/18409

PR #16372, opened 2023-02-15
Getting rid of some casts

Jojo-Schmitz:
...that at some point were needed to avoid compiler warnings, but are no longer

### DIFF
--- a/src/engraving/compat/midi/midirender.cpp
+++ b/src/engraving/compat/midi/midirender.cpp
@@ -351,7 +351,7 @@ static void collectNote(EventMap* events, const Note* note, CollectNoteParams no
 
     NoteEventList nel = note->playEvents();
     size_t nels = nel.size();
-    for (int i = 0, pitch = note->ppitch(); i < static_cast<int>(nels); ++i) {
+    for (size_t i = 0; i < nels; ++i) {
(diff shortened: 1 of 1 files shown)


src/engraving/compat/midi/midirender.cpp
```diff
@@ -351,7 +351,7 @@ static void collectNote(EventMap* events, const Note* note, CollectNoteParams no
 
     NoteEventList nel = note->playEvents();
     size_t nels = nel.size();
-    for (int i = 0, pitch = note->ppitch(); i < static_cast<int>(nels); ++i) {
+    for (size_t i = 0; i < nels; ++i) {
```
alexpavlov96, line 354: @Jojo-Schmitz why those are no longer creating warnings?
And does "for (int = 0, pitch = ... )" create them or you changed just because it's being used just once below?

Jojo-Schmitz: Maybe they never did, not  sure, but while `pitch` needs to be an `int`, i` should not

Jojo-Schmitz: And yes, `note->pitch` is only used once, so IHMO no need to create a separate variable

alexpavlov96 (approved):

Issue #16385, opened 2023-02-16
[MU4 Issue] Tuplet properties: Line thickness value is not rounded


bakajikara:
**Describe the bug**
Depending on the combination of staff space value and line thickness value, the tuplet's line thickness value will appear as a weird decimal.

**To Reproduce**
1. Set staff space to 1.250mm
2. Create a tuplet
3. Set line thickness to 0.12
4. Deselect the tuplet
5. Select the tuplet again and see strange decimal

https://user-images.githubusercontent.com/102702868/219267386-0254f9fb-f7f0-4578-94ff-20fa3d26793e.mp4

**Platform information**
 - OS: Windows 11

PR #16386, opened 2023-02-16
Tuplet properties: rounding line thickness value

bakajikara:
Resolves: #16385

(By the way, if I want to fix an unreported problem, should I create an issue first, or should I create a PR without creating an issue?)


oktophonie:
Since you asked; yes, ideally, create an issue as well as the PR, as you have done. It's a bit of extra work, but it's appreciated! It's helpful for us since we generally triage issues rather than PRs, and also means that explanation and discussion of the issue itself can be kept separate from the specific fix put forward in the PR (not so important in a simple case like this, but for more involved issues is very useful).

oktophonie (approved):

cbjeukendrup (approved):

Issue #15695, opened 2023-01-03
Add Tooltip to Home/Scores tab to read full score titles and file path


sberla365:
I think it would be particularly helpful to have a way to display the full length of the title on the recent scores picker at startup. Even better would be the possibility to show also the file path of each score in the picker. I think of some text that appears on hover similarly to what LibreOffice' Start Centre panel does. 
This would be very helpful when working on different versions of a piece whose name could exceed the maximum allowed length in the picker. Having the file-path would also be useful to distinguish files with the same name but in different folders

![image](https://user-images.githubusercontent.com/55401477/210331932-0c6c6761-7804-42b4-ab2e-25da6bc53932.png)

abariska:
It's really annoying thing. It shouldn't be a complicated fix, so I believe we should do file name tips when hovering

PetJer:
Hello @cbjeukendrup 

Can I try to do this for my first issue?

I have somewhat a general idea for what to do (looking throught the code).
I will be using a **tooltip** for showing the path.

cbjeukendrup:
@PetJer Yes, you're welcome! Using a tooltip sounds like a good idea.

jessicahalvorsen:
Hi @oktophonie @cbjeukendrup ! 

Is this still being worked on or could I try it as a first issue?

cbjeukendrup:
Someone already created a pull request for it a while ago: #16402
That PR is pretty much in the right direction, so probably I or someone else of the team should make some last improvements to it and merge it. 
So it would be better to work on a different issue for now :)

PR #16402, opened 2023-02-16
Adds a file tooltip to a score preview

PetJer:
Resolves: #15695

This PR adds a file tooltip to a score preview. It shows the name, path name and the time since modified of the score.

dev-test-score-tooltip

I would like to know @jessjwilliamson and @bkunda opinions about the positioning (or other). Here is a "troublesome" example:

dev-long-score-name-tooltip

### DIFF
--- a/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
@@ -40,6 +40,7 @@ FocusScope {
     property alias navigation: navCtrl
 
     signal clicked()
+    signal hovered(bool mouseInArea)
 
     NavigationControl {
         id: navCtrl
@@ -323,6 +324,10 @@ FocusScope {
 
         onClicked: {
             root.clicked()
+        }
+
+        onHoveredChanged: {
+            root.hovered(containsMouse)
         }
     }
 }
--- a/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
@@ -33,6 +33,7 @@ FocusScope {
     property string suffix: ""
     property alias timeSinceModified: timeSinceModified.text
     property var thumbnail: null
+    property var path: null
     property bool isAdd: false
     property bool isNoResultFound: false
     property bool isCloud: false
@@ -328,6 +329,7 @@ FocusScope {
 
         onHoveredChanged: {
             root.hovered(containsMouse)
+            console.log(path)
         }
     }
 }
--- a/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
@@ -41,7 +41,6 @@ FocusScope {
     property alias navigation: navCtrl
 
     signal clicked()
-    signal hovered(bool mouseInArea)
 
     NavigationControl {
         id: navCtrl
@@ -328,8 +327,15 @@ FocusScope {
         }
 
         onHoveredChanged: {
-            root.hovered(containsMouse)
-            console.log(path)
+            if (root.isAdd) {
+                return
+            }
+
+            if (containsMouse) {
+                ui.tooltip.show(root, root.name, root.path, root.timeSinceModified)
+            } else {
+                ui.tooltip.hide(root)
+            }
         }
     }
 }
--- a/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/ScoreItem.qml
@@ -332,9 +332,9 @@ FocusScope {
             }
 
             if (containsMouse) {
-                ui.tooltip.show(root, root.name, root.path, root.timeSinceModified)
+                ui.tooltip.show(scoreRect, root.name, root.path, root.timeSinceModified)
             } else {
-                ui.tooltip.hide(root)
+                ui.tooltip.hide(scoreRect)
             }
         }
     }
--- a/src/project/qml/MuseScore/Project/internal/RecentScoresView.qml
+++ b/src/project/qml/MuseScore/Project/internal/RecentScoresView.qml
@@ -129,7 +129,7 @@ StyledGridView {
                 }
                 // TODO: Set right tooltip position
                 if (mouseInArea) {
-                    ui.tooltip.show(scoreItem, score.name, score.path)
+                    ui.tooltip.show(scoreItem, score.name, score.path, scoreItem.timeSinceModified)
                 } else {
                     ui.tooltip.hide(scoreItem)
                 }
--- a/src/project/qml/MuseScore/Project/internal/RecentScoresView.qml
+++ b/src/project/qml/MuseScore/Project/internal/RecentScoresView.qml
@@ -123,18 +123,6 @@ StyledGridView {
                     root.openScoreRequested(score.path)
                 }
             }
-
-            onHovered: (mouseInArea) => {
-                if (isAdd) {
-                    return
-                }
-                // TODO: Set right tooltip position
-                if (mouseInArea) {
-                    ui.tooltip.show(scoreItem, score.name, score.path, scoreItem.timeSinceModified)
-                } else {
-                    ui.tooltip.hide(scoreItem)
-                }
-            }
         }
     }
 }
--- a/src/project/qml/MuseScore/Project/internal/RecentScoresView.qml
+++ b/src/project/qml/MuseScore/Project/internal/RecentScoresView.qml
@@ -90,6 +90,7 @@ StyledGridView {
         width: root.cellWidth
 
         ScoreItem {
+            id: scoreItem
 
             anchors.centerIn: parent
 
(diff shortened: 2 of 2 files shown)


src/project/qml/MuseScore/Project/internal/ScoreItem.qml
```diff
@@ -324,5 +325,17 @@ FocusScope {
         onClicked: {
             root.clicked()
         }
+
+        onHoveredChanged: {
+            if (root.isAdd) {
```
cbjeukendrup, line 330: You would also need to check for `root.isNoResultFound` here:
```suggestion
            if (root.isAdd || root.isNoResultFound) {
```

Tantacrul:
@jessjwilliamson - it would be good to review this and suggest any UI niceties you think could be made.

jessjwilliamson:
> @jessjwilliamson - it would be good to review this and suggest any UI niceties you think could be made.

I believe the score name and time modified shouldn't be included in the tooltip, just the location - as it's already below the thumbnail. To compromise long titles for the score names however, I notice it only limits to one line, so maybe let's make it possible to have 2 lines instead? (then '...') Like this :) @Tantacrul 
image

PetJer:
The latest commit does this:

dev-tooltip-title-bold

But because the tooltip title is bold by default it wraps weirdly compared to non-bold one (or at least in this case because of the long folder name):

dev-tooltip-title-non-bold

Should I better make a new function to show **Path tooltip** specifically, so I don't need to pass some weird bool value for a **path tooltip** (if the latest commit shouldn't have bold text)?

PetJer:
This does the latest commit:
dev-tooltip-title-newest

@jessjwilliamson I added a custom interval for the **Filetooltip** and it can be changed now.
@cbjeukendrup I hope I don't do any unnecessary stuff with the enum!

PetJer:
dev-filetooltip

src/framework/uicomponents/qml/MuseScore/UiComponents/StyledToolTip.qml
```diff
@@ -28,17 +28,24 @@ import MuseScore.UiComponents 1.0
 StyledPopupView {
     id: root
 
+    enum ToolTipType {
```
cbjeukendrup, line 31: Now the enum is defined both in QML and in C++. (That's partially my bad, because I didn't know you would also need it in C++ so therefore I suggested creating it in QML).

It would be better to use just the C++ definition, which can be used in QML by writing `QmlToolTip.Default` and `QmlToolTip.FileToolTip`.

(The type in `property int toolTipType:` should indeed still be `int` when doing that.)

PetJer: I see, didn't know that was possible.

src/project/qml/MuseScore/Project/internal/ScoreItem.qml
```diff
@@ -176,7 +177,7 @@ FocusScope {
                 text: root.name
 
                 wrapMode: Text.WrapAnywhere
-                maximumLineCount: 1
+                maximumLineCount: 2
```
cbjeukendrup, line 180: Does this not mean that the ScoreItem will become too tall and will therefore overlap with the item below it when there are so many recent scores that they appear in multiple rows?

jessjwilliamson:
Hi! just tested the build so far, tool tip looks lovely :) I think the hover area and delay can be tweaked a bit more :) I feel the tooltip is quite 'in your face'. 

- Can we make the hover area to be only on the title of the score and last date modified? So that means the hover doesn't get triggered from the thumbnail, but the details underneath instead. (see screenshot)
Screenshot 2023-03-20 at 11 47 55

-  Also, can we make the hover tooltip delay, reset when hovering another score? I feel it gets stuck in tooltip mode when im hovering through my scores

src/project/qml/MuseScore/Project/internal/ScoreItem.qml
PetJer: No?

filetooltip-overlapping

PetJer:
(Sorry for late update)

Jessica you can now tweak the closing and opening values as you feel :)

PetJer:
@cbjeukendrup the filetooltip/tooltip can escape the window (when not fullscreen). Any problems with that?

filetooltip-escaping

tooltip-escaping

PetJer:
[87fbbc7](https://github.com/musescore/MuseScore/pull/16402/commits/87fbbc761c41b739c95399739a0094d9a9dbda51) also changes the behavior for "normal" tooltips, too. I can revert it back just for them if needed.

cbjeukendrup:
@PetJer I think that is fine. Maybe it would be ideal if the tooltip is shown relative to the currently visible rectangle of the item, so like this:
Scherm­afbeelding 2023-04-09 om 18 54 22

But people on the internet seem quite certain that this is impossible to achieve (QML doesn't provide a way to detect which part of an item is currently inside the viewport) so I think you don't need to try.

cbjeukendrup:
> [87fbbc7](https://github.com/musescore/MuseScore/pull/16402/commits/87fbbc761c41b739c95399739a0094d9a9dbda51) also changes the behavior for "normal" tooltips, too. I can revert it back just for them if needed.

For normal tooltips it was desired behaviour that once one of them is open, it will immediately move along when you move to another button, instead of resetting the timer.
@jessjwilliamson, do you really think that this should be different for these filename tooltips? Maybe the feeling of the tooltip getting stuck will already be solved because the hoverable area is limited now.

jessjwilliamson:
> > [87fbbc7](https://github.com/musescore/MuseScore/pull/16402/commits/87fbbc761c41b739c95399739a0094d9a9dbda51) also changes the behavior for "normal" tooltips, too. I can revert it back just for them if needed.
> 
> For normal tooltips it was desired behaviour that once one of them is open, it will immediately move along when you move to another button, instead of resetting the timer. @jessjwilliamson, do you really think that this should be different for these filename tooltips? Maybe the feeling of the tooltip getting stuck will already be solved because the hoverable area is limited now.

Absolutely agree to keep as it is the desired behaviour for normal tooltips, that should still be able to move along. This reset should only apply to the filename tooltips only.
I just tested this build so far and it looks and feels very good! I approve :)

PetJer:
That should be it 👍

Jojo-Schmitz:
The vtests failures should be entirely unrelated to thie PR:

```
Run bash ./build/ci/vtests/build_and_pack.sh
bash: ./build/ci/vtests/build_and_pack.sh: No such file or directory
Error: Process completed with exit code 127.
```

jessjwilliamson:
@PetJer @cbjeukendrup We'd like this to be a small feature for 4.2. There's been an update on 4.1 to view cloud scores tab too. Can we revisit this PR and make it work to online scores tab too? :)

cbjeukendrup:
@jessjwilliamson But what information should the tooltip display for cloud scores? Those don't really have a file path, so should we show just the name instead?

@PetJer A rebase is needed: 
- ScoreItem has been moved to a different place and has been modified significantly; it already has a `path` property nowadays. So you just need to add back the tooltip.
- RecentScoresView has been completely refactored, but since the `path` property of ScoreItem is already being set, you don't need to make any changes to RecentScoresView anymore.

jessjwilliamson:
> @jessjwilliamson But what information should the tooltip display for cloud scores? Those don't really have a file path, so should we show just the name instead?
> 
> @PetJer A rebase is needed:
> 
> * ScoreItem has been moved to a different place and has been modified significantly; it already has a `path` property nowadays. So you just need to add back the tooltip.
> * RecentScoresView has been completely refactored, but since the `path` property of ScoreItem is already being set, you don't need to make any changes to RecentScoresView anymore.

We can just show the name with the cloud icon next to it, or to make it simple, no action at all.

PetJer:
@cbjeukendrup @jessjwilliamson Displaying the name again seems unnecessary. What about a direct link to .com? So the tooltip says something like "View online" and links to e.g. [here](https://musescore.com/user/16006641/scores/4197961)

jessjwilliamson:
> @cbjeukendrup @jessjwilliamson Displaying the name again seems unnecessary. What about a direct link to .com? So the tooltip says something like "View online" and links to e.g. [here](https://musescore.com/user/16006641/scores/4197961)

That would be quite conflicting to have a clickable tooltip while user may just try to open a cloud score in the app. Might be a bit overwhelming. It could provide only URL link, but it wouldn't be clickable. However cloud scores have cloud files in your system. Is that right @cbjeukendrup ? You can clarify what you think would be best

PetJer:
> 2. Is that on the tool tip itself or on the score name?

For now the score name and tooltip are without the icon, but what were you referring here to:
> We can just show the name with the cloud icon next to it (...)

jessjwilliamson:
> > 2. Is that on the tool tip itself or on the score name?
> 
> For now the score name and tooltip are without the icon, but what were you referring here to:
> 
> > We can just show the name with the cloud icon next to it (...)

It was just an idea, referenced from how Photoshop do it. But i don't think it's necessary for now.

RomanPudashkin:
I am going to archive this PR. We will have to look into this issue internally at some point, as it turned out to be quite complicated

PR #16421, opened 2023-02-17
fixed #9825: The plugins list from context menu exceeds height (for multiple monitors)

### LINKED ISSUE #9825: [MU4 Issue] The plugins list from context menu exceeds height
**Describe the bug**
I'm not sure who to assign but can anyone can help out with this problem?

Taking an example from this plugin patch by Melda, they have a lot of plugins (I believe 37 plugins) but it's not going to fit into one long context menu. On my screen, 32 are visible (of course there will be different results depending on your scale and screen resolution). 

How it is on my screen with 3840 x 2160 resolution with 200% scaling (only showing 32, exceeding above):
image

How it really is showing all 37 plugins (same resolution but 100% scaling):
image
 
Would need to find a solution to shorten the list. Possibly something like this:

https://user-images.githubusercontent.com/71656189/142909842-ade6cbd6-44d0-4056-b4f7-4dc8eb998baa.mov

More chances of situations like this are when finding an instrument or FX in the Mixer with lots of options. This can apply as a general context menu behaviour rule throughout the app, it doesn't have to be specifically for the mixer.

If possible or another solution, please reach out so we can fix this :) Thanks!

 - OS: Windows (and maybe Mac)

Eism:
Resolves: #9825

### DIFF
--- a/src/framework/uicomponents/qml/MuseScore/UiComponents/internal/StyledMenu.qml
+++ b/src/framework/uicomponents/qml/MuseScore/UiComponents/internal/StyledMenu.qml
@@ -74,11 +74,11 @@ MenuView {
 
         var itemsCount = model.length - sepCount
 
-        var anchorItemHeight = Boolean(root.anchorItem) ? root.anchorItem.height : (Screen.desktopAvailableHeight - padding * 2)
+        var anchorItemHeight = root.anchorGeometry().height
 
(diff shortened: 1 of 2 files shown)


RomanPudashkin (approved):

jessjwilliamson:
Fixed :) Thank you!

Issue #13234, opened 2022-09-07
[MU4 Issue] Items aren't shown as deselected after using selection filter


wizofaus:
**Describe the bug**
If you select a range of measures containing notes, lyrics etc., then turn off various elements using the selection filter, the colours are not updated on the score to indicate which elements are no longer selected.

**To Reproduce**
1. Create or open a score with notes and lyrics
2. Select all
3. Turn off "lyrics" in the selection filter
4. Observe that the lyrics are still coloured blue
5. Press the 'v' key to hide all selected elements - only the notes are hidden, not the lyrics, indicating the lyrics were in fact deselected

**Expected behavior**
Deselecting via the selection filter should cause unselected items to display in their regular unselected colour.

**Screenshots**

**Platform information**
 - Windows 11

**Additional context**

KaiVinter:
I just want to bring back attention to this issue. I just came across the same issue when trying to only select one voice.

DmitryArefiev:
Fixed in #16431

PR #16431, opened 2023-02-18
Notify about selection changed after editing selection filter

cbjeukendrup:
Resolves: #13234

We mark the selection as changed, so that in `NotationInteraction::setSelectionTypeFiltered`, `notifyAboutSelectionChangedIfNeed()` will indeed emit a notification.


DmitryArefiev:
Tested #13234 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #16435, opened 2023-02-19
Fix #340288: Concert Band template is missing

Jojo-Schmitz:
because it is "corrupt", `Error=XML_ERROR_PARSING_ELEMENT ErrorID=6 (0x6) Line number=93`, simply a `"` too much.

It is not really missing, just not shown in the list of templates to choose from.

Resolves https://musescore.org/en/node/340288#comment-1175489 ff.


cbjeukendrup (approved):

Jojo-Schmitz:
Isn't this simple enough to make it into 4.0.2?

PR #16439, opened 2023-02-20
Fix #15219 Received MIDI controller is always CC0

### LINKED ISSUE #15219: [MU4 Issue] CC signals from MIDI controller are not correctly recognized in MIDI device mapping
**Describe the bug**
When using a MIDI keyboard with buttons that are assigned to CC signals, the MIDI device mapping only recognize the signal as "CC 0", although another (e.g. CC 24) is assigned. This happens for every trigger pad I tested.
The bug was confirmed in the German Musescore forum section.

**Hardware**
- MIDI keyboard: M-Audio Oxygen pro mini

**Expected behavior**
The CC signals should be distinguished properly

**Platform information**
 - OS: W10

lachbaer:
Control index was not converted from Midi 1.0 to Midi 2.0 when it was not a RPN or NPRN controller.

Resolves: #15219

In case the controller number was not one of the (N)RPN controllers the `case` `break`'d prematurely and the controller data was never set in the Event. I guess the removed lines were some residue from the initial development phase.

I currently don't have a full dev environment to run a debug version. So this PR is actually only theoretical. Can anybody please counter check this? Thanks.

### DIFF
--- a/src/framework/midi/midievent.h
+++ b/src/framework/midi/midievent.h
@@ -767,11 +767,13 @@ struct Event {
                 break;
             //D3.3
             case Opcode::ControlChange: {
-                std::set<uint8_t> skip = { 6, 38, 98, 99, 100, 101 };
-                if (skip.find(index()) == skip.end()) {
-                    break;
-                }
                 switch (index()) {
+                default:
+                    event.setIndex(index());
+                    event.setData(scaleUp(data(), 7, 32));
+                    break;
+                // RPN and NPRN controller messages are no ordenary conrollers
+                // and need special handling in MIDI 2.0
                 case 99:
                     event.setOpcode(Opcode::AssignableController);
                     event.setBank(static_cast<uint16_t>(data()));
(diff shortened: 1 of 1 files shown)


src/framework/midi/midievent.h
```diff
@@ -767,11 +767,13 @@ struct Event {
                 break;
             //D3.3
             case Opcode::ControlChange: {
-                std::set<uint8_t> skip = { 6, 38, 98, 99, 100, 101 };
-                if (skip.find(index()) == skip.end()) {
```
lachbaer, line 771: If current controller number (`index()`) is not in in the _skip_ set then break the _Opcode::ControlChange_ case immediately, resulting in keeping _0_ from the `Event` initialization as the controller number. Therefore removed, because other controller numbers _are_ handled by the `default` case.
A new MIDI event always goes through `toMidi20` after reception and being created as a Midi 1.0 event in the first place.

src/framework/midi/midievent.h
```diff
@@ -767,11 +767,13 @@ struct Event {
...
-                    break;
-                }
                 switch (index()) {
+                default:
```
lachbaer, line 771: Put `default` case to the beginning for code readability, as this is the standard case and not a "last resort fallback". The other cases are specializations.

cesarizu:
I rebased this on top of V4.0.2 and it seems to be working just fine, I get now some actual mappings (using an Arturia MiniLab mkII keyboard)

image

lachbaer:
@Eism There seems to be no update to this PR. Can you review, please? Thank you.

igorkorsukov (approved):

PR #16440, opened 2023-02-20
Fix false filter logic against MIDI system messages

lachbaer:
I spottet a potential problem that accepts system messages (0xFz), though they obviously should be filtered out.

The previous logic was already always true when only the sign bit was set, so 0x80 to 0xFF, effectively not filtering at all.

IMO no cryptic bit-logic is needed. Just checking if the status byte is between 0x80 and 0xF0 (excl.) is enough.

A more 'fancy' way would be `(u.byte[0] >> 7) && (u.byte[0] & 0x70 ^ 0x70)` - but that's less understandable.

### DIFF
--- a/src/framework/midi/midievent.h
+++ b/src/framework/midi/midievent.h
@@ -104,14 +104,8 @@ struct Event {
             uint32_t full;
         } u;
         u.full = data;
-        if (
-            (u.byte[0] & 0x80)
-            || (u.byte[0] & 0x90)
-            || (u.byte[0] & 0xA0)
-            || (u.byte[0] & 0xB0)
-            || (u.byte[0] & 0xC0)
-            || (u.byte[0] & 0xD0)
-            || (u.byte[0] & 0xE0)
+        if ((u.byte[0] >= 0x80)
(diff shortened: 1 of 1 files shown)


Eism:
@lachbaer 
It's hard for me to say whether your solution is correct, the more you solve a potential problem here. Are there articles or projects that support your assumption?

src/framework/midi/midievent.h
```diff
@@ -104,14 +104,8 @@ struct Event {
             uint32_t full;
         } u;
         u.full = data;
-        if (
```
lachbaer, line 107: This condition shall _obvioulsy_ check for MIDI messages that are in the range "Note off message" (0x8y) up to "Pitch bend change" (0xEy) and filter all System Exclusive/System Common messages (0xFy) that are proprietary to specific devices.

However, the check for `u.byte[0] & 0x80` is always true when the MSB (sign bit) is set, and so already makes the complete or'ed condition true. Line 109 to 114 are superflous as they are never reached.

igorkorsukov (approved):

PR #16448, opened 2023-02-20
enabled IIR low pass filter in fluid synth

handrok:
There is a low-pass filter in the fluid, it is always there, when this value is read from the soundfont, then it is taken. If not present, the default filter is applied. The default filter is set to pass everything up to 20 kHz. With a sample rate of 44100, it is ok, half of the sample rate is just the maximum frequency that you will hear in the digital. That is, such a default filter will not cut anything and skip everything. However, there was bug in the editor which was fixed in https://github.com/musescore/MuseScore/pull/16445, each frequency decreased by a factor of 6.7, including the default one. That is, the default filter, which did nothing before, began to close everything above 3 kHz. And also the custom filter from the soundfont began to close 6.7 times less than it was set in the soundfont. So in https://github.com/musescore/MuseScore/commit/6155a6323aef28f5cd2c94e120ea3b77f69f5e2e was decided to disable the filter I think this is no longer needed, taking into account the edits in the pull request https://github.com/musescore/MuseScore/pull/16445


vpereverzev (approved):

Jojo-Schmitz:
rebase needed, it is now in thirdparty/fluidsynth/fluidsynth-2.3.3/src/synth/fluid_voice.c

PR #16453, opened 2023-02-20
Changed draw data structure to tree

igorkorsukov:


vtest/vtest_context.json
```diff
@@ -1,6 +1,6 @@
 {
     "mscore_ref_bin": "./build.debug/install/bin/mscore",
-    "score_dir": "./vtest/scores",
+    "score_dir": "/home/igor/Dev/scores",
```
igorkorsukov, line 3: At the moment, the system is still in a state of development, it does not run on CI, so it is not critical, I will fix it in the next PR

Eism (approved):

Issue #16334, opened 2023-02-12
[MU4 Issue] Converting to mscz on the command line results in zero-sized output files.


kbloom:
**Describe the bug**
Converting to mscz on the command line results in zero-sized output files. 

**To Reproduce**

1. Find a score, e.g. https://github.com/kbloom/sheetmusic/blob/master/Agadelcha.mscz
2. mscore Agadelcha.mscz -o Agadelcha.mscz
3. ls -l Agadelcha.mscz

The resulting file has zero size.

**Expected behavior**
A valid mscz file should be output at Agadelcha.mscz

**Platform information**
Manjaro Linux running the Manjaro package of Musescore 4.0.1

**Additional context**
In this example, I attempt to upgrade a file from Musescore 3 format to Musescore 4 format. However, this also happens when the input is a Musescore 4 file. This happens when the output format is mscz. When the output format is pdf instead, a valid pdf is created. This also happens when the output file has a different  name than the input file.

kbloom:
Since I just reread my bug description, and thought it might be confusing, I wanted to clarify.

 I get a zero file size when:

1. Converting Musescore3 file to Musescore 4 file.
2. Converting Musescore4 file to Musescore 4 file (which one might want to do if they were applying a new stylesheet to the score)
3. Doing the above conversions when the output file has a different name from the input file.
4. Doing the above conversions when the output file overwrites the input file.

The conversion is performed correctly when converting to another file format, such as PDF.

helpy:
Confirmed!
The same happens if I try to convert a mxl file to a mscz file with MuseScore4.
The same command line works with MuseScore3.

Eism:
@kbloom @helpy 
Please check the build from this PR https://github.com/musescore/MuseScore/pull/16461

kbloom:
I have tested your Linux build artifact on a bunch of my scores, and it works.

kbloom:
This didn't get pulled into 4.0.2, did it? Could it get pulled into 4.0.3 if there is a 4.0.3?

Issue #16339, opened 2023-02-12
[MU4 Issue] MSCZ files corrupted beyond repair, because of corrupted ZIP file


cbjeukendrup:
**Describe the bug**
A few times per month, someone somewhere in the world reports a file corruption where the ZIP container file has become corrupted, so that nothing of the contained score can be saved.

This has been the case for years now, and it is exactly the same in MS3 and MS4 for years now. 

The problem exists in several variants:
1. the saved MSCZ file is completely empty, i.e. 0KB
2. the saved MSCZ file contains only zeros, no real data
3. the saved MSCZ file contains seemingly random data
4. the saved MSCZ file still looks a bit like a ZIP file but is still corrupted so that one or more files from it can't be read

In this issue, we will collect cases of this. When adding a case, please make sure to add the following information: 
- **OS** (Windows, macOS, Linux) and **OS version**
- **MuseScore version** in which the file was saved the last time
- **Filename** under which the file was saved
- A link where **the file** in question can be downloaded, or at least a clear description which of the variants mentioned above is applicable

Please keep comments clear and concise, and don't use this thread for long back-and-forth conversations, to avoid making it longer than necessary.

cbjeukendrup:
For example: 

**Original report:** https://github.com/musescore/MuseScore/issues/15113
**OS:** Windows
**MuseScore version:** 4.0.0.223362132 / d050fb0
**Filename:** unknown
**The file:** [corrupted_file.zip](https://github.com/musescore/MuseScore/files/10203464/corrupted_file.zip)
**Variant:** contains only zeros

cbjeukendrup:
**Original report:** https://musescore.org/en/node/342465
**OS:** Windows 10
**MuseScore version:** 4.0.1
**Filename:** `C:\Users*********\Documents\Musescore4\Scores\KONZERT_gdur.mscz`
**Variant:** 0KB

cbjeukendrup:
Original report: https://discord.com/channels/818804595450445834/818804809784229898/1057744302765256874
OS: Windows 10
MuseScore version: 4.0.0 / 5485621
Filename: Figaro.mscz
Variant: corrupted zip file; could be (partially?) repaired using `zip -FF`
I'm not allowed to upload the file here, unfortunately, since it was sent to me privately

Jojo-Schmitz:
See also https://musescore.org/en/node/271185 and the many "What links here".
[...]

PR #16461, opened 2023-02-20
Fix saving files on the command line

cbjeukendrup:
We were making mistakes with opening/closing IODevices

Resolves: #16334

Unfortunately, this doesn't give very obvious clues about #16339, because saving files in the UI and on the command line go through different flows in the code. The only thing we can do is going through all open/close logic, to make sure that it is correct everywhere. I've already done a bit of that and haven't found anything interesting yet.


igorkorsukov (approved):

kbloom:
I'm the original reporter of this bug. I have tested your Linux build artifact on a bunch of my scores, and it works.

Issue #15768, opened 2023-01-08
[MU4 Issue] MuseSounds cannot handle large scores


metasekk:
**Describe the bug**
When a score becomes large (high number of pages whatever the content of the score is - *can be empty* - or medium number of pages with a lot of notes), Muse Sounds stop working and become permanently silent from a certain point onwards.

**To Reproduce**
1. Create a new empty score using MuseSounds in at least one staff.
2. Add more than 1000 bars (2*999 for example).
3. Go to the end of the score, and add notes on the staves using MuseSounds.
4. Hear no sound. Besides, there exists a well-defined bar before which MuseSounds works perfectly and after which it is permanently turned off.

**Expected behavior**
During playback, any staff using MuseSounds should play wherever it is located in the score, regardless of the length and content of the score.

**Screenshots**
I provide a short [video](https://drive.google.com/file/d/1aK4jwHAQkP4d1s6IKKsEmYvMDy7sZUcM/view?usp=share_link).

**Platform information**
 - OS: Windows 10 Version 22H2, 64 bits, x64-based processor
 - Tested and verified on: MuseScore Release version (64-bit): 4.0.0-223472200 with MuseSampler v.0.3.2 ; MuseScoreNightly-230080305-master-3602349-x86_64 with MuseSampler v.0.3.2

**Additional context**
I believe it is most likely to be correlated with https://github.com/musescore/MuseScore/issues/15094. If this is established, then the moment when MuseSounds is turned off depends at least on:

- The number of notes of the score;
- The number and values of tempo markings;

Consequently, in empty scores, MuseSounds can work up to more than 1000 bars (1075 in the case of the video I provide) but when the score contains stuff, the bug can happen at much lower bars (bar 367 in the score provided in https://github.com/musescore/MuseScore/issues/15094).
This variability makes this problem highly disabling.

cbjeukendrup:
This came up a while ago on the MuseScore.org forum too. I shared it with the Muse Sounds team, and @iamtesch told me that they basically already have a fix, but this will require a change in MuseScore itself too, so will be released together with the next MuseScore update.

DmitryArefiev:
Fixed in #16464 but still doesn't work because of an old version of MuseSampler. 

Should be rechecked when **v0.4** is released

DmitryArefiev:
Tested on **4.1 RC build**, Win10, MuseSounds, Muse Sampler library is **0.4.0** - **FIXED**

Issue #15078, opened 2022-12-09
[MU4 Issue] Velocity and pitch settings have no effect with Muse Sounds (and should be disabled if they can't be utilised)


bkunda:
**Describe the bug**
I'm sure there is nothing surprising here, but it's worth documenting that velocity and pitch settings in **Properties** currently have no effect with Muse Sounds instruments.

**To Reproduce**
Demo video: 

https://user-images.githubusercontent.com/86290556/206734224-519ea89a-deaa-4e33-bf1c-e2b5ddda2770.mov

**Expected behavior**
Ideally, these settings should be hooked-up for Muse Sounds instruments where possible. 
If this is not possible, then these settings should be disabled when a Muse Sounds instrument is being used. 

**Platform information**
 - macOS 12.6 (M1) 

**Additional context**
Possibly relates to #14768

mountbest:
We NEED to control volume (velocity) of individual notes (voices) for individual instrument. For example, in guitar music there are typically 3, sometimes 4 voices in a single staff. Without controlling voice/note volumes the guitar playback is very unrealistic.

DaddyLudwig:
And also, there are some piano pieces for which different velocities for the 2 hands are necessary, like this C minor polonaise by Chopin. 

https://user-images.githubusercontent.com/6530761/216480501-322fd882-f1fd-49d5-8a78-8fc243f6636d.MP4

PR #16464, opened 2023-02-21
Update MuseSampler API wrapper for v0.4 support

iamtesch:
Resolves: #15768 (once MuseSampler v0.4 is released)
Resolves Partially: #15078 (addresses most of tuning issues; v0.4 of MuseSampler should resolve remainder)

- New functions in v0.4 fix issue for long scores
- Also uses specific events from v0.3 that support "per-note-tuning", and appropriately calls these functions (note -- still a musesampler bug in playback in v0.3; works when dragging a note but not completely in ordinary playback)
- Moves version checking logic into the libhandler, and adds per-version behavior here for new API functions

### DIFF
--- a/src/framework/musesampler/imusesamplerconfiguration.h
+++ b/src/framework/musesampler/imusesamplerconfiguration.h
@@ -37,8 +37,6 @@ public:
 
     virtual mu::io::path_t backupLibraryPath() const = 0;
     virtual mu::io::path_t userLibraryPath() const = 0;
-
-    virtual std::string minimumSupportedVersion() const = 0;
 };
 }
 
--- a/src/framework/musesampler/internal/apitypes.h
+++ b/src/framework/musesampler/internal/apitypes.h
@@ -64,11 +64,25 @@ typedef struct ms_DynamicsEvent
     double _value; // 0.0 - 1.0
 } ms_DynamicsEvent;
 
+// Added in v0.4
+typedef struct ms_DynamicsEvent_2
+{
+    long long _location_us;
+    double _value; // 0.0 - 1.0
+} ms_DynamicsEvent_2;
+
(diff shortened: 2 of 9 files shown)


vpereverzev (approved):

RomanPudashkin (approved):

DmitryArefiev:
@iamtesch What should be tested here? 

As I understand, #15768 still occurs because of an old version of MuseSampler (0.3.2)

iamtesch:
> @iamtesch What should be tested here?
> 
> As I understand, #15768 still occurs because of an old version of MuseSampler (0.3.2)

Unfortunately, the full scope is a little difficult to test without an internal build of the upcoming 0.4 release.  However, the "per note tuning" does work with this change.  If you have an old version of the MuseSampler lying around (pre 0.3), you can also test to ensure the proper/extant API functions are called.

The core contribution here is really having more robust and custom version handling within the sampler wrapper, so that as the API is extended we can have MuseScore adapt easily, instead of being fixed to a subset of functionality.  Because the release cycles of the sampler and MuseScore aren't identical, we have added support for the known upcoming features before the sampler is actually released so that this support can make it into the next MuseScore release.

DmitryArefiev:
Ah, got it! I just tested general playback with MuseSounds on your builds (Win/Mac/Lin) - all fine. 

Pitch settings are also work good (but they also work in master.. interesting)

I reopened #15768. Will check and close when v.0.4 is ready

MarcSabatella:
Based on this I was kind of expecting to find tuning adjustments working in a current nightly build of 4.1 with Muse Sampler 0.4, but it's not the case.  If I click a note with a tuning adjustment, it's played with that adjustment indeed, but it is not honored in ordinary playback.  Should we be expecting this to work?

Issue #16433, opened 2023-02-19
[MU4 Issue] Custom slur positions lost in imported MS3 files


rgreen5:
**Describe the bug**

The attached score shows a case where custom slur positions in MS3 are being lost when imported into MS4.

MS3:
![ms3_slur](https://user-images.githubusercontent.com/87246583/219956248-91c8c82c-7048-426e-a1da-33dd0591d6b4.png)

MS4:
![ms4_slur](https://user-images.githubusercontent.com/87246583/219956311-e211e7a3-ce5c-4e26-8006-c63c04519287.png)

**To Reproduce**

1. Open the attached MS3 score in MS3 and MS4. 
[custom_slurs.zip](https://github.com/musescore/MuseScore/files/10777168/custom_slurs.zip)
2. Look at the slurs in the bass staff, mm 9-11.

RESULT: The slurs are the wrong shape in MS4. See also #13906.

**Expected behavior**

Custom slur positions should be retained from MS3.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230500504, revision: github-musescore-musescore-8c85e6c

oktophonie:
This is by design/necessity: all custom slur postions will be reset when migrating MU3 -> 4.

The new default position is actually this:
![image](https://user-images.githubusercontent.com/20604032/219970185-73840f21-283c-48fd-bc0e-3b6e4588f874.png)
but for some reason this is not shown immediately: instead we see the strange interim state that you illustrate. This corrects itself (to the above) as soon as a redraw happens. _This_ bug, at least, can be investigated here.

I'd recommend disabling autoplace for these slurs when you readjust them, to sidestep the new collision avoidance system for the middle. At least until we can implement a system where you can properly attach slur endpoints to specific notes in a chord.

PR #16467, opened 2023-02-21
When migrating, reset defaults before calling layout

mike-spa:
Resolves: #16433

As far as I can see, the problem is that we are calling the first score layout before doing the resetting operations required by older scores migration. Moving the layout call _after_ the reset operations seems to fix the issue.


mike-spa:
Nope sorry, don't know why it seemed to be working. It doesn't

Issue #16442, opened 2023-02-20
[MU4 Issue] Crashes when deleting bars or change volta properties in certain files


unclewalter:
**Describe the bug**
When deleting more than two bars, it seems to crash with some files. I've attached a file than seems to cause it to happen.

[Musescore file](https://drive.google.com/file/d/1UHjAAxOGDS0Rf2PJ4GE2m7som3OadHoQ/view?usp=share_link)

**To Reproduce**
1. Select more than two bars
2. Cmd+delete
3. Program crashes

_or_

1. Change number or repeats
2. Program crashes

**Platform information**
 - OS: MacOS 13.2.1 (22D68)

cbjeukendrup:
@unclewalter It looks like some special access is needed to download the attached file from Google Drive. You can upload the file directly on GitHub by wrapping it in a ZIP file.

unclewalter:
[New Song 2.mscz.zip](https://github.com/musescore/MuseScore/files/10783879/New.Song.2.mscz.zip)

oktophonie:
I find this file crashes upon opening(!) so perhaps a corruption issue at play?

cbjeukendrup:
It seems to be related to the repeat list. I've created a stopgap solution in #16470, but I wonder why the situation even happens. @jeetee Could you please also take a look?

jeetee:
I'll try to have a look this week.

I'm more interested into _why_ Harmony is requesting a utick that's outside of the score (according to repeatList). Seems like either Harmony or the RepeatList is not up to date when this call happens.

DmitryArefiev:
Fixed in #16514

PR #16470, opened 2023-02-21
Fix crash when rendering chord symbol for utick past end of repeat list

cbjeukendrup:
Resolves: #16442

Not quite convinced yet that this should be the solution though...

### DIFF
--- a/src/engraving/libmscore/harmony.cpp
+++ b/src/engraving/libmscore/harmony.cpp
@@ -1140,6 +1140,10 @@ Fraction Harmony::ticksTillNext(int utick, bool stopAtMeasureEnd) const
     Segment* cur = seg->next();
     auto rsIt = score()->repeatList().findRepeatSegmentFromUTick(utick);
 
+    if (rsIt == score()->repeatList().end()) {
+        stopAtMeasureEnd = true;
+    }
+
     Measure const* currentMeasure = seg->measure();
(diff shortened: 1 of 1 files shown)


jeetee:
Since one shouldn't be asking to render a Chord Symbols that won't be played back (utick not existing in the score), stopping the playback of it at the end of it's measure is "as good as any" other arbitrary decision on where to stop it.
Another alternative that crossed my mind is to use the last measure of the last repeatsegment instead, which would allow (in theory) the duration to be analyzed up to the next symbol or score end in a non-crashing way as well and might be slightly more of a "correct" interpretation of this Chord Symbol even though it isn't actually part of the playback roadmap.

What this does highlight, to me, is a deeper issue with how to handle skipped parts of the score and whether we should somehow allow "explicit" playback of them by hoisting them into some separate repeatsegments. This might also in part then address the expectancy raised by @abariska in #16263 to somehow allow playback of a skipped segment when explicitly requested (for example by a user clicking inside such a region).
It also showcases that we currently don't really handle "gaps" well, as tick2utick will likely conjure up a utick for a tick inside a skipped part, possibly overlapping with an actual utick. To be addressed in the TempoMap and timing information refactor (which might still need a task somewhere?).

So my suggested change (untested!) would be something like this as the current stopgap:
```
    IF_ASSERT_FAILED(rsIt != score()->repeatList().end()) {
        // Requested playback position is outside of current playback roadmap
        rsIt = std::prev(score()->repeatList().end()); // Try and make the best of it, but don't go past score end
    }
```

PR #16472, opened 2023-02-21
Return old vtest system

igorkorsukov:
new vtest system not stable, so I return old vtest system

### DIFF
--- a/.github/workflows/ci_vtests.yml
+++ b/.github/workflows/ci_vtests.yml
@@ -4,47 +4,239 @@ on:
   pull_request:
     branches:
     - master
+    - 4.0.2
 
 jobs:
   setup:
-    name: "VTests workflow"
+    name: "Setup VTests workflow"
     runs-on: ubuntu-20.04
+    outputs:
+      do_run: ${{ steps.output_data.outputs.do_run }}
+      reference_sha: ${{ steps.output_data.outputs.reference_sha }}
+      artifact_name: ${{ steps.output_data.outputs.artifact_name }}
     steps:
     - name: Cancel Previous Runs
       uses: styfle/cancel-workflow-action@0.11.0
       with:
         access_token: ${{ github.token }}
     - name: Clone repository
-      uses: actions/checkout@v3  
+      uses: actions/checkout@v3
     - name: "Configure workflow"
-      env:
-        pull_request_title: ${{ github.event.pull_request.title }}
       run: |
-        bash ./build/ci/tools/make_build_number.sh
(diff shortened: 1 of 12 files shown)


.github/workflows/ci_vtests.yml
```diff
@@ -4,47 +4,239 @@ on:
...
+    name: "Setup VTests workflow"
     runs-on: ubuntu-20.04
+    outputs:
+      do_run: ${{ steps.output_data.outputs.do_run }}
+      reference_sha: ${{ steps.output_data.outputs.reference_sha }}
+      artifact_name: ${{ steps.output_data.outputs.artifact_name }}
     steps:
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@0.11.0
+      uses: styfle/cancel-workflow-action@0.9.1
```
cbjeukendrup, line 19: You're accidentally downgrading these actions. This one should be `0.11.0`, and the others should be `v3` everywhere instead of `v2`. Otherwise, we will get a bunch of warnings and those old ones will stop working at some point.

igorkorsukov: Thanks for the notification, fixed

.github/workflows/ci_vtests.yml
```diff
@@ -4,47 +4,239 @@ on:
...
+        echo "UPLOAD_ARTIFACT_NAME=$UPLOAD_ARTIFACT_NAME"
 
+  generate_current_pngs:
+    name: "Generate current PNGs"
+    runs-on: ubuntu-20.04
+    needs: setup
+    if: needs.setup.outputs.do_run == 'true' # Can't use env: see https://github.com/actions/runner/issues/480
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v2
```
cbjeukendrup, line 69: All these should be v3 too (just everything with `uses: actions/…`

igorkorsukov: fixed

src/framework/audio/internal/worker/playback.cpp
```diff
@@ -63,6 +63,11 @@ void Playback::deinit()
     disconnectAll();
 }
 
+bool Playback::isInited() const
```
Eism, line 66: You aren't using this method

igorkorsukov: I believe it is useful, so I decided to keep it (after changing the solution when I stopped used it)

Eism (approved):

Issue #16457, opened 2023-02-20
[MU4 Issue] Score becomes corrupted after actions with deleting piano instrument, undo and adding notes to bottom staff


DmitryArefiev:
**To Reproduce**
1. Open [Untitled score.mscz.TXT](https://github.com/musescore/MuseScore/files/10787184/Untitled.score.mscz.TXT)
2. Delete Piano from Instruments panel
3. Undo
4. Add note to bass clef staff
5. Save

https://user-images.githubusercontent.com/90187801/220181366-c17f3237-b6a2-440f-8d59-ba3d8cab5353.mp4

**Platform information**
 Windows10

DmitryArefiev:
Fixed in #16474

rich777300:
I downloaded the score, removed the .txt extension, opened it in MS4.0.1, went into the instruments panel, selected piano, and pressed the trash can icon on the top right of the instruments panel. MuseScore crashed on me, twice.

I'm on Windows 10 64-bit

oktophonie:
@rich777300 Yes, this is as expected. Only in a build of the specific PR that fixes this issue (https://github.com/musescore/MuseScore/pull/16474) would you expect to not experience the crash. (I can confirm it works)

DmitryArefiev:
Retested on Win10

PR #16474, opened 2023-02-21
score_corruption_fix_4.0.2

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/16457


cbjeukendrup:
@RomanPudashkin See also https://github.com/musescore/MuseScore/pull/15964, which would also have fixed this issue

RomanPudashkin:
> @RomanPudashkin See also #15964, which would also have fixed this issue

It hasn't fixed this issue, unfortunately. I'll soon open a separate PR for the master branch

RomanPudashkin:
> but we also needed to update when for example the master staves are changed, right?

Yes, in this scenario we also may get a corrupted score (especially if the user reorders staves in the master score, but not in the part score)

Specifically for this issue, we filled the cache once when loading the score, but Staff::_visibilityVoices was not yet ready at that time, so we could not fill m_tracksMapping. The next time updateTracksMapping() is called, m_tracksMapping will also not be updated because of the cache. So we will always have an invalid state for m_tracksMapping

DmitryArefiev:
Tested #16457 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Eism (approved):

DmitryArefiev:
@RomanPudashkin Found another case with corruption


https://user-images.githubusercontent.com/90187801/220890826-256f3f62-f204-4b7e-a533-f3ef69209f11.mp4

[test1 (3).mscz.TXT](https://github.com/musescore/MuseScore/files/10812918/test1.3.mscz.TXT)

PR #16504, opened 2023-02-23
fixed #15521: Screen readers may read "not selected" when navigating instrument list

### LINKED ISSUE #15521: [MU4 Issue] Screen readers may read "not selected" when navigating instrument list (NVDA only)
**Describe the bug**
Blind users navigating the instrument list (available instruments within a family) hear "not selected", which is confusing to them

**To Reproduce**
1. Enable screen reader (from @DmitryArefiev : only NVDA reads "not selected" on my side, other screen readers are fine)
2. Create a new score
3. Select "Choose instruments"
4. Select Woodwinds
5. Press Tab to move to the the Instruments column
6. Press Down repeatedly
7. Hear error: depending on your screen reader, each instrument may report that it is "not selected"

**Expected behavior**
The name of the instrument should be read without that additional text

 - OS: Windows

**Additional context**
I believe this happens with NVDA but not most other screen readers.  It ends up causing a lot of confusion, because hearing it read "not selected" even after the instrument is actually added to the score makes people think it wasn't in fact added even though it was.  I've heard this feedback from three different users.

Eism:


RomanPudashkin (approved):

DmitryArefiev:
Tested on Narrator,NVDA, VoiceOver and Orca - see details in https://github.com/musescore/MuseScore/issues/15521#issuecomment-1761582337

PR #16507, opened 2023-02-23
Fix #11235: Added the ability to delete individual files from recent scores

### LINKED ISSUE #11235: Allow removing individual files from Recent Scores list
**Describe the enhancement**

In LibreOffice, recently opened files are displayed as thumbnails when you launch the program. Each has a little cross on the right-hand top edge which allows the user to remove individual files from display. In MS, by contrast, the display of scores is all or nothing; you either keep every score you have ever opened on display or delete all of them (File > Open recent > Clear recent files).

SUGGESTION: Implement ability to remove recently opened file icons _individually_ from the "Scores" section.

bakajikara:
Resolves: #11235

Added a cross button in the upper right corner of the thumbnail of each score that allows users to remove the score from the recent scores.

https://user-images.githubusercontent.com/102702868/220952882-e1f84e36-4064-4120-b081-6d86b0a58c19.mp4

### DIFF
--- a/src/project/qml/MuseScore/Project/internal/ScoresPage/ScoreItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/ScoresPage/ScoreItem.qml
@@ -39,6 +39,7 @@ FocusScope {
     property bool isNoResultFound: false
     property bool isCloud: false
     property int cloudScoreId: 0
+    property bool isRemovingScoresAllowed: false
 
     property alias navigation: navCtrl
 
@@ -182,7 +183,7 @@ FocusScope {
                     anchors.right: parent.right
 
                     icon: IconCode.CLOSE_X_ROUNDED
-                    visible: root.path != "" && parent.containsMouse
+                    visible: root.path != "" && root.isRemovingScoresAllowed && parent.containsMouse
                     transparent: true
 
                     onClicked: {
--- a/src/project/qml/MuseScore/Project/internal/ScoresPage/ScoreItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/ScoresPage/ScoreItem.qml
@@ -43,6 +43,7 @@ FocusScope {
     property alias navigation: navCtrl
 
     signal clicked()
+    signal removeRequested()
 
     NavigationControl {
         id: navCtrl
@@ -59,18 +60,6 @@ FocusScope {
         }
 
         onTriggered: root.clicked()
-    }
-
-    MouseArea {
-        id: mouseArea
-
-        anchors.fill: parent
-
-        hoverEnabled: true
-
-        onClicked: {
-            root.clicked()
-        }
     }
 
     Column {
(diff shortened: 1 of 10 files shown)


cbjeukendrup (commented):
The code looks very good! One minor comment: I would rename functions/signals with `clearRecentScore` to `removeRecentScore`, to make it clearer that these functions remove only one score from the list, rather than clearing the whole list.

@bkunda Could you please check whether you agree with this design?

bakajikara:
Thanks for the feedback! I've renamed functions and signals as you mentioned :)

bkunda:
I quite like this idea!
Still a few things to address with the implementation:

First, a couple of issues with cloud scores:
1. After removing all cloud scores via the new ‘X‘ button, all cloud scores in the user's ‘Cloud scores‘ directory suddenly re-appear
2. After removing some cloud scores via the new ‘X‘ button and using **File --> Open Recent --> Clear recent files**, all cloud scores in the user's 'Cloud scores' directory re-appear (including the ones just removed)

https://user-images.githubusercontent.com/86290556/221163271-e3e7c966-8d0c-4d98-bc47-e9a7bfa6985f.mov

And an issue with the UI: This 'X' icon is difficult to make out in dark mode, I suspect because if font_primary_colour is being used, it doesn't have enough contrast against the score background. @jessjwilliamson could you please advise on this? Perhaps we need to use the black colour (with maybe some reduced opacity) in both light and dark modes? 

https://user-images.githubusercontent.com/86290556/221163215-83313dda-2fb5-4b33-8e19-1ca9251d4689.mov

bakajikara:
Regarding the cloud scores issue, [these lines](https://github.com/musescore/MuseScore/blob/a5412ad3936e75119695a07eb6c528329be53dd3/src/project/internal/projectconfiguration.cpp#L121-L123) seem to cause the cloud scores to be loaded when the recent scores list is empty.

I've come up with three solutions (there may be others), but I'm not sure which one is desirable.
1. Remove these lines; both issues 1 and 2 will be solved, but cloud scores will not be displayed until users open the score.
2. Do not show the X button on cloud scores; the usability is somewhat limited, but strange behavior will no longer occur.
3. Do not remove cloud scores on Clear recent scores; this only fixes issue 2, but it will be similar to the current behavior while increasing functionality.

src/project/qml/MuseScore/Project/internal/ScoresPage/ScoreItem.qml
```diff
@@ -384,4 +374,29 @@ FocusScope {
             }
         }
     }
+    
+    MouseArea {
```
cbjeukendrup, line 378: The MouseArea needs to be below (i.e. lower z-index) the CloudScoreIndicatorButton. Otherwise, the CloudScoreIndicatorButton is not hoverable anymore. What is the reason that you moved the MouseArea to the bottom of the file and thus increased its z-index?

src/project/qml/MuseScore/Project/internal/ScoresPage/ScoreItem.qml
```diff
@@ -39,10 +39,12 @@ FocusScope {
     property bool isNoResultFound: false
     property bool isCloud: false
     property int cloudScoreId: 0
+    property bool isRecentView: false
```
cbjeukendrup, line 42: I would prefer to give this property a more generic name, like `isRemovingScoresAllowed`. Reason: it is generally not nice to give a base component knowledge about its derived components.

bakajikara: I understand. I will change it.

JohnodonCode:
@bakajikara are you going to come back to this at any point?

PR #16511, opened 2023-02-23
Fix that previously resetted excerpts are not reopened after reopening the file

cbjeukendrup:
Resolves: part 2 from https://github.com/musescore/MuseScore/pull/15665#issuecomment-1441873817

The reason that I'm calling the method of the `Score` directly, is that the new `Score` is the only thing that gets out of sync with the rest. When calling it via `->notation()->setIsOpen` or `setExcerptIsOpen`, some unnecessary notifications would be emitted.


RomanPudashkin (approved):

DmitryArefiev:
Tested https://github.com/musescore/MuseScore/pull/15665#issuecomment-1441873817 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #12772, opened 2022-08-10
[MU4 Issue] PNG Export dialog - Resolution - unable to enter more than 3 digits 


sammik:
**Describe the bug**
It is not possible, to enter more than 3 digits in Export - PNG - Resolution dialog

**To Reproduce**
1. Open Score
2. Export - PNG
3. Try to enter new resolution higher, than "999"

**Expected behavior**
Any "allowed" resolution should be enterable

**Screenshots**

https://user-images.githubusercontent.com/1646034/184036158-b7b21a6a-7727-40ec-b572-b01402399a40.mp4

**Platform information**
 - OS: [reported on Linux]

**Additional context**
- Higher resolution is possible to achieve by hitting UP arrow.
- In MU3, it worked well.

Tantacrul:
@sammik - what is the use case for having the DPI so high? You've mentioned it worked well on MS3. I'm curious? It'll help me prioritise this.

sammik:
> @sammik - what is the use case for having the DPI so high? You've mentioned it worked well on MS3. I'm curious? It'll help me prioritise this.

@Tantacrul In fact nothing. I just found it, when testing different bug (and checked MU3 as standard issue report routine).
I would say, it cannot affect any real world work, so priority is not real question :)

octopols:
I would like to work on this issue, can someone assign it to me?

octopols:
`intinputvalidator.cpp` acts as a input validator for the text field holding the dpi value, the regular expression it accepts have a minimum length of `1` and maximum length of `3`. 
If it does not affect any other text field in a bad way I can change the regular expression such that it accepts values of max length 4.

Jojo-Schmitz:
Same issue in Mu3, but entirely different cause apparently

PR #16525, opened 2023-02-24
Fix for issue #12772: [MU4 Issue] PNG Export dialog - Resolution - unable to enter more than 3 digits

octopols:
Changed the regular expression to validate integers of maximum length 4.

Resolves: #12772 

### DIFF
--- a/src/framework/uicomponents/view/validators/intinputvalidator.cpp
+++ b/src/framework/uicomponents/view/validators/intinputvalidator.cpp
@@ -48,8 +48,10 @@ QValidator::State IntInputValidator::validate(QString& inputStr, int& cursorPos)
 {
     QValidator::State state = Invalid;
 
-    if (inputStr.contains(QRegularExpression("^\\-?\\d{1,3}$"))) {
-        if (inputStr.contains(QRegularExpression("^\\-?0{2,3}"))
+    //maxDigits is the number of digits in the maximum absolute value between m_top and m_bottom.
+    int maxDigits = std::ceil(std::log10(std::max(std::abs(m_top), std::abs(m_bottom))));
+    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(maxDigits)))) {
+        if ((maxDigits >= 2 && inputStr.contains(QRegularExpression(QString("^\\-?0{2,%1}").arg(maxDigits))))
             || (inputStr.startsWith("-") && inputStr.toDouble() == 0.0)) {
             state = Intermediate;
         } else {
--- a/src/framework/uicomponents/view/validators/intinputvalidator.cpp
+++ b/src/framework/uicomponents/view/validators/intinputvalidator.cpp
@@ -48,10 +48,12 @@ QValidator::State IntInputValidator::validate(QString& inputStr, int& cursorPos)
 {
     QValidator::State state = Invalid;
 
-    //maxDigits is the number of digits in the maximum absolute value between m_top and m_bottom.
-    int maxDigits = std::ceil(std::log10(std::max(std::abs(m_top), std::abs(m_bottom))));
-    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(maxDigits)))) {
-        if ((maxDigits >= 2 && inputStr.contains(QRegularExpression(QString("^\\-?0{2,%1}").arg(maxDigits))))
+    const int maxAbsoluteValue = std::max(std::abs(m_top), std::abs(m_bottom));
+    const int maxNumberOfDigits = maxAbsoluteValue > 0
+                                  ? std::floor(std::log10(maxAbsoluteValue)) + 1
+                                  : 1;
+    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(maxNumberOfDigits)))) {
+        if ((maxNumberOfDigits >= 2 && inputStr.contains(QRegularExpression(QString("^\\-?0{2,%1}").arg(maxNumberOfDigits))))
             || (inputStr.startsWith("-") && inputStr.toDouble() == 0.0)) {
             state = Intermediate;
         } else {


octopols:
This should also allow default zoom percentage (preferences > canvas) to enter more than 3 values as well (the maximum allowed zoom percentage is `1600`)

src/framework/uicomponents/view/validators/intinputvalidator.cpp
```diff
@@ -48,7 +48,7 @@ QValidator::State IntInputValidator::validate(QString& inputStr, int& cursorPos)
 {
     QValidator::State state = Invalid;
 
-    if (inputStr.contains(QRegularExpression("^\\-?\\d{1,3}$"))) {
+    if (inputStr.contains(QRegularExpression("^\\-?\\d{1,4}$"))) {
         if (inputStr.contains(QRegularExpression("^\\-?0{2,3}"))
```
Eism, line 52: should be `2,4`

Jojo-Schmitz: Why? as far as I can tell single digit DPI values are alowed?

As, I see, they apparently were in Mu3, but in Mu4 the minimum is 32

octopols: Hey, I apologize for not addressing this comment earlier. 
The validator is specifically linked to `IncrementalPropertyControl.qml`, and there are other properties with minimum values in single digits.
When I set this validator to 1, 4, the `IntInputValidator::fixup` method inside `intinputvalidator.cpp` will automatically adjust it to the minimum value for DPI if the value entered is below 32. 
Using `2, 4` might lead to issues with properties that have minimum values in single digits, for example, the insert measure popup.

octopols: Perhaps, We could try to define these values as variables and pass a property to QML, which can then set these variables if it's possible with QRegularExpression.

cbjeukendrup: There might be a misunderstanding here. As I understand it, the comment is about line 52, not line 51.
```suggestion
        if (inputStr.contains(QRegularExpression("^\\-?0{2,4}"))
```
Is that true @Eism?

cbjeukendrup: Also, instead of '4' we might want to use something in the direction of `std::ceil(std::log(std::max(std::abs(m_top), std::abs(m_bottom))))`, isn't that better?

octopols: > Also, instead of '4' we might want to use something in the direction of `std::ceil(std::log(std::max(std::abs(m_top), std::abs(m_bottom))))`, isn't that better?

Thank you, Casper! That's exactly what I needed. I'll test it out and make a commit

src/framework/uicomponents/view/validators/intinputvalidator.cpp
```diff
@@ -48,8 +48,10 @@ QValidator::State IntInputValidator::validate(QString& inputStr, int& cursorPos)
 {
     QValidator::State state = Invalid;
 
-    if (inputStr.contains(QRegularExpression("^\\-?\\d{1,3}$"))) {
-        if (inputStr.contains(QRegularExpression("^\\-?0{2,3}"))
+    // The expression is based on the number of digits in the maximum absolute value between m_top and m_bottom.
+    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(std::ceil(std::log10(std::max(std::abs(m_top),
+                                                                                                          std::abs(m_bottom)))))))) {
+        if (inputStr.contains(QRegularExpression("^\\-?0{2,4}"))
```
cbjeukendrup, line 54: The '4' should be replaced here too. And it would be better to extract that intimidating std::ceil(………) expression into a named local variable.

octopols: Yes! the if statement looks way too long, I'll make the necessary changes and commit soon.

src/framework/uicomponents/view/validators/intinputvalidator.h
```diff
@@ -48,6 +48,7 @@ public slots:
 private:
     int m_top = 999;
     int m_bottom = -999;
+    mutable int m_maxDigits = 4;
```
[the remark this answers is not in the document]
octopols: I wasn't sure about the appropriate value for initialization I just used the number I'm replacing. Could you please provide more details on why are we using 3?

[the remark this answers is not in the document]
cbjeukendrup: @octopols I meant just a local variable, not a class member. Like this:
```c++
    int maxDigits = std::ceil(std::log10(std::max(std::abs(m_top), std::abs(m_bottom))));
    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(m_maxDigits)))) {
        …
```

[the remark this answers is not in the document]
octopols: I've made the change

cbjeukendrup (approved):

src/framework/uicomponents/view/validators/intinputvalidator.cpp
```diff
@@ -48,8 +48,10 @@ QValidator::State IntInputValidator::validate(QString& inputStr, int& cursorPos)
 {
     QValidator::State state = Invalid;
 
-    if (inputStr.contains(QRegularExpression("^\\-?\\d{1,3}$"))) {
-        if (inputStr.contains(QRegularExpression("^\\-?0{2,3}"))
+    //maxDigits is the number of digits in the maximum absolute value between m_top and m_bottom.
+    int maxDigits = std::ceil(std::log10(std::max(std::abs(m_top), std::abs(m_bottom))));
+    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(maxDigits)))) {
+        if (inputStr.contains(QRegularExpression(QString("^\\-?0{2,%1}").arg(maxDigits)))
```
cbjeukendrup, line 54: Wait, I realised that this might cause problems if maxDigits is 1. I guess that the expression will be invalid in that case. Easiest fix seems:
```suggestion
        if (maxDigits >= 2 && inputStr.contains(QRegularExpression(QString("^\\-?0{2,%1}").arg(maxDigits)))
```

octopols: Thank you for noticing. I'll update it.

octopols: {2,2} seems to be valid according to the information I found online.

src/framework/uicomponents/view/validators/intinputvalidator.cpp
```diff
@@ -48,8 +48,10 @@ QValidator::State IntInputValidator::validate(QString& inputStr, int& cursorPos)
 {
     QValidator::State state = Invalid;
 
-    if (inputStr.contains(QRegularExpression("^\\-?\\d{1,3}$"))) {
-        if (inputStr.contains(QRegularExpression("^\\-?0{2,3}"))
+    //maxDigits is the number of digits in the maximum absolute value between m_top and m_bottom.
+    int maxDigits = std::ceil(std::log10(std::max(std::abs(m_top), std::abs(m_bottom))));
+    if (inputStr.contains(QRegularExpression(QString("^\\-?\\d{1,%1}$").arg(maxDigits)))) {
+        if (maxDigits >= 2 && inputStr.contains(QRegularExpression(QString("^\\-?0{2,%1}").arg(maxDigits)))
             || (inputStr.startsWith("-") && inputStr.toDouble() == 0.0)) {
```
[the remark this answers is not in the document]
octopols: I added the condition through VS Code in my previous commit. Even Qt Creator is warning me about it 😅 I'll make a new commit.

Jojo-Schmitz:
Same issue in Mu3, but entirely different cause apparently

Jojo-Schmitz:
Why the heck are vtests failing? Esp. disturbing:
symbol-2-1 diff

Maybe a rebase would fix it? Also 6 commit for a 2-line change seems a bit much ;-)

cbjeukendrup:
Yes, a rebase would fix it, and in this case we can squash the commits when merging

cbjeukendrup:
@octopols After thinking about it carefully, I realised that the formula was still not entirely correct and safe, but now I fixed it myself in order not to bother you again

zacjansheski:
Tested on Mac OS 14, Ubuntu 20.04.6, Windows 11
https://github.com/musescore/MuseScore/issues/12772 FIXED

PR #16534, opened 2023-02-24
Fix #13126: Applying a barline to a note on the first beat changes the previous measure's end barline

### LINKED ISSUE #13126: [MU4 Issue] Application of barlines is inconsistent
**Describe the bug**

The following applies to barline applied from the palette:

1. If you apply a barline to the _first_ note of a measure, the barline replaces the start barline of the _following_ measure. However if you apply a barline to any other note except the first, the barline always appears _in front_ of the note.

__Suggestion__: If a barline is applied to the first note of a measure, it should replace the previous end barline (and not the start barline of the next measure).

2. If a barline is applied to a measure, the barline appears _after_ the measure _except_ if the applied barline is a start repeat—in which case it appears _before_ the selected measure.

__Suggestion__: If a barline is applied to a measure, it should always replace the _following_ start barline.

See also #9429.

bakajikara:
Resolves: #13126

When a barline is applied to a note on the first beat, it now changes the end barline of the previous measure.

### DIFF
--- a/src/engraving/libmscore/chordrest.cpp
+++ b/src/engraving/libmscore/chordrest.cpp
@@ -400,6 +400,13 @@ EngravingItem* ChordRest::drop(EditData& data)
             bl->setGenerated(false);
 
             if (tick() == m->tick()) {
+                if (bl->barLineType() != BarLineType::START_REPEAT) {
+                    m = m->prevMeasure();
+                    if (!m) {
+                        delete e;
+                        return 0;
+                    }
+                }
                 return m->drop(data);
             }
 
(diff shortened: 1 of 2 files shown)


src/engraving/libmscore/chordrest.cpp
```diff
@@ -400,6 +400,12 @@ EngravingItem* ChordRest::drop(EditData& data)
             bl->setGenerated(false);
 
             if (tick() == m->tick()) {
+                if (bl->barLineType() != BarLineType::START_REPEAT) {
+                    m = m->prevMeasure();
+                    if (!m) {
+                        return 0;
```
cbjeukendrup, line 406: I believe you would need to add `delete e;` above this line, like in other places in this method.
```suggestion
                        delete e;
                        return 0;
```

bakajikara: Thanks, fixed now!

PR #16542, opened 2023-02-25
Add style support to display capo chords only

hoongern:
Feature proposal: Add a new style option to "display capo chords only". When enabled, chord readability and layout improves (on scores with a dense number of chords), for parts/musicians which only care about the capo chords.

E.g. When enabled with capo fret 1, "Am7/C" would be displayed instead of "Bbm7/Db (Am7/C)".

Given the desired flexibility for some parts to display capo chords, other parts to display both original & capo chords, and to be synced, it seemed easier to write this into the core harmony code rather than using a plugin (although I do not have experience with the codebase and am open to other suggestions on how to achieve this effect). This does seem to mean that saving the score in a build missing the style id erases the option.

If this approach seems reasonable, I could look into writing a test (I couldn't find documentation on running unit tests on Mac OS). I've performed a simple manual test with a mix of chord symbols, nashville chords and roman numerals.


miiizen:
Thanks for this - I've rebased and reworked the PR so we can get this merged as soon as possible!

avvvvve:
UI looks good! Will let @its-not-nice give the engraving go-ahead.

https://github.com/user-attachments/assets/7ca20fb8-5365-40ca-9645-22a86702cdfc

cbjeukendrup:
Another rebase needed

its-not-nice (approved):

Issue #16383, opened 2023-02-15
[MU4 Issue] Dragging objects small distances- hitching and catching


five-over-four:
**Describe the bug**
When you try to drag objects very small distances, the objects tend to get 'caught' on the page such that larger motions of the cursor are required to move them.

**To Reproduce**
1. Place a note in a new score in any configuration and add any movable object that is tied to the note, such as staff text or a fingering marking.
2. Drag the object next to the notehead and zoom in the maximum amount.
3. Now move the object very small amounts near the notehead repeatedly.
4. You'll notice the object hitching, as if the program has trouble deciding whether we're just clicking, or clicking and dragging the object.

**Expected behavior**
When the user first clicks on an object, small movements should not drag it, as the intent can be to simply activate it for editing. However,  after the mouse button has been held for a short period of time, this 'movement lock' should not re-enable until the click has been released. Therefore, when dragging objects around, they should be able to be moved freely without this 'catching' effect.

**Screenshots**
I made a short youtube video illustrating the problem, here: https://www.youtube.com/watch?v=JIwyxp56HRA

**Platform information**
 - OS: Windows 10.

**Additional context**
This issue was made apparent to me while placing fingering markings for the guitar, as those often go next to the noteheads, so any adjustments must be very small.

DmitryArefiev:
Fixed in #16545

PR #16545, opened 2023-02-25
Check minimum drag delta only before starting a drag operation

cbjeukendrup:
Once dragging has started, the dragged thing should move smoothly, and not refuse to be dragged when it is close to its original position.

Resolves: #16383


RomanPudashkin (approved):

DmitryArefiev:
Tested #16383 on Win10, Mac12 - **FIXED**    

On Linux, I found an issue with hitch dragging of objects. That occurs in older builds but not in 4.0.2 release build. I'll recheck that after merging in master.

Issue #16549, opened 2023-02-25
[MU4 Issue] When file is "dirty" because of changes in the Mixer, AutoSave marks it "saved"


cbjeukendrup:
**To Reproduce**
Prerequisite: AutoSave is turned on, preferably with a short time interval
1. Open an existing score
2. Make changes in the Mixer
-> the score is correctly marked "dirty", i.e. containing unsaved changes; this can be seen from the `*` that is appended to the title of its tab in the tab bar
4. Wait until AutoSave does its work
-> the score is incorrectly marked "clean".
4. Close the file
-> as a result of marking the score "clean", there is no popup asking "do you want to save the changes". So, MuseScore just closes the score without ever saving the changes to the real file (as opposed to the AutoSave file).
5. Reopen the file, to find out that the changes were indeed not saved to the real file.

**Expected behaviour**
Even though AutoSave "saves" the file to some hidden file, it should not mark the file as "saved", because the changes are not definitively saved to the original file.

**Platform information**
 - OS: macOS (N/A)

DmitryArefiev:
Fixed in #16552

PR #16552, opened 2023-02-25
Don't mark audio settings / view settings as saved after AutoSave

cbjeukendrup:
Only mark them as saved when *everything* has been saved *successfully* to the *real file* of the project.

So, not after AutoSave; not after "Save a copy" (but certainly after "Save as"); not after writing the audio settings to some buffer but then failing to write that buffer to a file; etc.

Resolves: #16549


Eism (approved):

DmitryArefiev:
Tested #16549 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #16342, opened 2023-02-13
[MU4 Issue] "Note-input" shortcut "N" is not documented in the tooltip


LWChris:
**Describe the bug**
All buttons of the note-input toolbar show their keyboard shortcut in their tool tips, except for the "note-input" toggle itself. It should show the "(N)" addition in the first line.

**To Reproduce**
1. Hover Quarter note button - tool tip says "Quarter note (5; Num+5)"
2. Hover Slur - tool tip says "Slur (S)"
3. Hower Note-input - tool tip says "Default (step-time)"

**Expected behavior**
Tooltip should somehow mention that "N" is the shortcut to enter/leave note-input mode.

**Screenshots**
![image](https://user-images.githubusercontent.com/845152/218361815-8e32bac1-4284-4597-ba7a-c799cc4e1b59.png)

**Platform information**
 - OS: Windows 10

**Additional context**
- MuseScore 4.0.1

Tejkh1:
Thank you

pratyaksh1610:
@cbjeukendrup  I think it is inactive since it's been 2 weeks. Can I raise a PR ?  I have solved the issue.

Tejkh1:
@pratyaksh1610  yes please go ahead.

PR #16559, opened 2023-02-26
Added shortcut "N" on "Note-input" tooltip.

pratyaksh1610:
Resolves: #16342 

Added the shortcut "N" on "Note-input" tooltip.

https://user-images.githubusercontent.com/76740999/221401035-b64e9cf9-8ce3-4697-977b-5051b7267621.mp4

<br>

### DIFF
--- a/src/framework/shortcuts/data/shortcuts-Mac.xml
+++ b/src/framework/shortcuts/data/shortcuts-Mac.xml
@@ -609,6 +609,10 @@
     <seq>Num+8</seq>
     </SC>
   <SC>
+      <key>note-input-steptime</key>
+      <seq>N</seq>
+      </SC>
+  <SC>
     <key>pad-note-1</key>
     <seq>7</seq>
     <seq>Num+7</seq>
--- a/src/framework/shortcuts/data/shortcuts.xml
+++ b/src/framework/shortcuts/data/shortcuts.xml
@@ -609,6 +609,10 @@
     <seq>Num+8</seq>
     </SC>
   <SC>
+    <key>note-input-steptime</key>
+    <seq>N</seq>
+    </SC>
+  <SC>
     <key>pad-note-1</key>
     <seq>7</seq>
     <seq>Num+7</seq>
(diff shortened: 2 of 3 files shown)


Jojo-Schmitz:
Hmm, that change sort of makes sense, but N is the shortcut already, even without being listed in shortcuts.xml. I wonder why?

cbjeukendrup:
I'm afraid this is not entirely the correct solution, indeed because `N` is already the shortcut for the `note-input` action, so now we would have two actions with the same shortcut. 

Instead, I think we need to look in NoteInputBarModel.cpp: in `NoteInputBarModel::currentNoteInputModeAction()`, we just return the action for the current note input method. This way, the action has the name, icon and shortcut specific to the current note input method. But before returning the action, I think we should modify it to take the shortcut of the generic "toggle note input" action (with code `note-input`). So:
1. Query the action specific to the current note input method from the `uiActionsRegister()`;
2. Query the generic `note-input` action from `uiActionsRegister()`;
3. Change the shortcut of the method-specific action to the shortcut of the generic action;
4. Return the method-specific action.

(Note that we are not modifying the actions that are in the `uiActionsRegister()`, but just the local copy that is used in the NoteInputBarModel. That's exactly what we want.)

shoogle (changes requested):
Please implement @cbjeukendrup's solution.

To understand the problem:

1. Click and hold on the note input button.
2. Choose one of the other note input methods, like Re-pitch.

image

Notice that the main button's icon changes to that of the chosen input method, and the tooltip's name, description, and the displayed shortcut also change.

image

This is desirable except for the shortcut. We want the shortcut for the main button to always be displayed as "N" regardless of the chosen input method.

Inside the menu, the specific shortcut (e.g. Ctrl+Shift+I) should always be displayed next to each mode. "N" shouldn't be displayed as the shortcut for any modes in the menu.

limse10:
Hi, I believe I have fixed this issue with the method described by @cbjeukendrup

Check out #16619 

Thank you!

cbjeukendrup:
@limse10 I will check it. However, in the future, please don't create PRs for issues on which another contributor is already actively working. This can lead to wasted efforts and is very not nice for the other contributor.

limse10:
Ah, I understand. I will take note of that for the future. Thank you @cbjeukendrup and sorry @pratyaksh1610 for any inconvenience caused.

cbjeukendrup:
Closing as replaced by #16619

Issue #16554, opened 2023-02-25
[MU4 Issue] Parts created while score in concert pitch have incorrect key signatures


xlf1024:
**Describe the bug**
If you create a *new* part while the main score is in concert pitch, the new part will use the key signature appropriate for concert pitch for instrument pitch, and a reverse-transposed key signature for instrument pitch. The individual notes aren't affected, accidentals are automatically placed to compensate. Also, the default parts aren't affected; even if first opened while concert pitch is active.

**To Reproduce**
1. Create a score with a transposing instrument (e.g. Bb clarinet, e.g. in C major)
2. Switch the main score to concert pitch
3. Enter some notes (e.g. a C major scale)
4. Open the parts window, create a **new** part, open that part
5. Using the instruments panel, reveal the instrument
6. *The part is now displayed with the notes for instrument pitch, but the key signature for concert pitch (in the example from above: key signature for C major, notes D, E, F#, G, B, C#, D). The concert pitch toggle is unchecked*
7. Switch back to the main score, it is displayed in (correct) concert pitch, but the concert pitch toggle is still unchecked.
8. Switch back to the part, and enable the concert pitch toggle
9. *The part is now displayed with the notes for concert pitch, but a reverse-transposed key signature (in the example from above: two flats = Bb major as key signature, notes C, D, E natural, F, G, A, B natural, C)*

**Expected behavior**
The part should use the correct key signatures, independent of which state the main score was in when the part was created (in the example: C major in concert pitch, two sharps = D major in instrument pitch)

**Screenshots**
Main score in instrument pitch
![grafik](https://user-images.githubusercontent.com/33392024/221377803-4ed91e53-8963-4503-b34a-03afedd06bf6.png)
Main score in concert pitch
![grafik](https://user-images.githubusercontent.com/33392024/221377831-fd4d2ac8-7a5c-4cb4-ba45-b0beb18f1b8b.png)
Instrument score in instrument pitch
![grafik](https://user-images.githubusercontent.com/33392024/221377844-96cdb436-0c31-402a-8210-e30b51f51c35.png)
Instrument score in concert pitch
![grafik](https://user-images.githubusercontent.com/33392024/221377862-2646095f-5b4a-4038-8374-01ad7bebad36.png)

**Platform information**
- OS: Windows 10 x64
- MuseScore version (64-bit): 4.0.1-230121751
- revision: github-musescore-musescore-9b70a8c

**Additional context**
- Autogenerated parts aren't affected
[...]

PR #16563, opened 2023-02-26
Fix key signatures regarding concert pitch when cloning staff

cbjeukendrup:
when either the full score is in concert pitch and the part not, or the part is in concert pitch and the score not.

Resolves: #16554

### DIFF
--- a/src/engraving/libmscore/excerpt.cpp
+++ b/src/engraving/libmscore/excerpt.cpp
@@ -1549,6 +1549,23 @@ void Excerpt::cloneStaff2(Staff* srcStaff, Staff* dstStaff, const Fraction& star
 
         cloneSpanner(s, score, dstTrack, dstTrack2);
     }
+
+    bool oscoreConcertPitch = oscore->styleB(Sid::concertPitch);
+    bool scoreConcertPitch = score->styleB(Sid::concertPitch);
+
+    if ((oscoreConcertPitch && !scoreConcertPitch)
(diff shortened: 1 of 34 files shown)


cbjeukendrup:
The third commit is an attempt to fix an assertion failure that happened sometimes but I still haven't found a way to reproduce it reliably. It must be some permutation of toggling concert pitch in the full score and in the part score, adding instruments and saving/closing/reopening, and adding key signatures. 

The fourth commit fixes an additional bug that still occurred after the second commit: 
1. Create score with transposing instrument and concert C major key sig
8. Turn on concert pitch
9. Save close reopen
10. Turn off concert pitch
11. Create new empty part
12. Make the instrument visible in that part
-> the instrument has the wrong key sig in the part

@mike-spa Since you have worked before on issues that turned out to be related to deleting C major key signatures at the beginning of a score, could you please check if this PR looks good to you?

mike-spa (approved):
@cbjeukendrup as far as I can see, this looks perfectly coherent with all the other similar fixes I've done (removing a keySig from the score just because it happens to be C is a _very_ wrong idea IMO, so I'm very glad to see it gone).

oktophonie (approved):

RomanPudashkin:
@cbjeukendrup please pay attention to the failed tests

cbjeukendrup:
@RomanPudashkin Done!

RomanPudashkin (approved):

Issue #16594, opened 2023-02-28
Adding Fermatas during playback with metronome != 100% breaks TempoMap


cplrossi:
### Issue type

General playback bug

### Bug description

If you set the metronome speed to a value other than 100 per cent then start playback, entering Fermatas during playback creates entries in the TempoMap with incorrect values.

For example, if the BPM of the score is 120, suppose you set 50% (60 BPM) then start playback. If you add a normal fermata (multiplier = 2.0) during playback, the fermata will be played at 15 BPM, instead of 30 BPM, and playback will continue at 30 BPM, instead of 60, and this persists with each playback.

### Steps to reproduce

1. Use a sample score, e.g. 3 bars filled with 3 whole notes.
2. Set Metronome to something != 100%
3. Start playback
4. Put a Fermata on second note.
6. Continue playback and/or rewind and start playback

### Screenshots/Screen recordings

https://user-images.githubusercontent.com/57630292/221905411-11495837-d431-4118-a689-e482beb8c811.mp4

### MuseScore Version

non-stable prerelease 4.1.0

### Regression

No.

### Operating system

Arch Linux

### Additional context

If this is considered a bug, and if it's confirmed, I wrote a one line fix and can make a PR.

jeetee:
I'm just wondering whether we *want* to allow this type of editing during score playback (it never crossed my mind to even attempt this)

ruza-net:
Please note that this bug occurs even if the fermata was not added during playback - you only have to increase the tempo with the slider and then add the fermata (see #17401 )

PR #16596, opened 2023-02-28
Fix TempoMap changes during playback

borsattoz:
Resolves: #16594 

This change fixes the incorrect values of TempoMap when putting fermatas on notes during playback with metronome != 100%

### DIFF
--- a/src/engraving/dom/score.cpp
+++ b/src/engraving/dom/score.cpp
@@ -590,7 +590,8 @@ void Score::rebuildTempoAndTimeSigMaps(Measure* measure, std::optional<BeatsPerS
             }
 
             if (!RealIsNull(stretch) && !RealIsEqual(stretch, 1.0)) {
-                BeatsPerSecond otempo = tempomap()->tempo(segment.tick().ticks());
+                BeatsPerSecond multiplier = tempomap()->tempoMultiplier();
+                BeatsPerSecond otempo = tempomap()->tempo(segment.tick().ticks()) / multiplier.val;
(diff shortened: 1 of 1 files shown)


src/engraving/libmscore/score.cpp
```diff
@@ -708,7 +708,8 @@ void Score::rebuildTempoAndTimeSigMaps(Measure* measure)
                 }
             }
             if (stretch != 0.0 && stretch != 1.0) {
-                BeatsPerSecond otempo = tempomap()->tempo(segment.tick().ticks());
+                BeatsPerSecond multiplier = tempomap()->tempoMultiplier();
```
Jojo-Schmitz, line 711: Why an extra variable it if is only used once?

cplrossi: I chose this way because of line length and the fact that e.g. `ntempo` is used once also. But I understand that the name `ntempo` conveys some additional semantics, mine none, and line length would be not so excessive. Should I commit a single line assignment?

Jojo-Schmitz: What does the staff team think?

cbjeukendrup:
This has now been closed, because in the end we chose to fix this by doing a slightly bigger overhaul, see https://github.com/musescore/MuseScore/pull/27549. But the research done for this PR definitely helped us in finding a solution, so credits for that!

borsattoz:
@cbjeukendrup Thank you for your work!

Issue #15268, opened 2022-12-17
[MU4 Issue] Portamento glissandos break playback for random assortment of instruments (synths, keyboards, and voices)


erinic04:
**Describe the bug**
Portamento glissandos cause the note they are attached to to ring out until the note is played again for a seemingly random assortment of instruments. Saxophone, cornet, and accordion all work fine, but synths, keyboards, and voices don't

**To Reproduce**
1. Create a score with portamento glissandos for one of the affected instruments
2. Hear glissando breaking everything

**Expected behavior**
The glissando should not cause the notes to ring out indefinitely

**Screenshots**

https://user-images.githubusercontent.com/21023353/208237548-f7ce42fd-ca2e-4460-9e4a-0c17755fb92a.mp4

https://user-images.githubusercontent.com/21023353/208237551-8571ef68-0242-4846-bc84-e77646004555.mp4

https://user-images.githubusercontent.com/21023353/208237552-753825d8-2338-4185-8284-fe82df55ecd8.mp4

https://user-images.githubusercontent.com/21023353/208237556-8acae5bb-983d-4e10-abd8-851a1ef1115f.mp4

https://user-images.githubusercontent.com/21023353/208237565-bca350bf-398b-462e-9e15-ab69fda2c4a3.mp4

**Platform information**
 - OS: Reported on Windows

**Additional context**

djfun:
see #12838

djfun:
`src/framework/mpe/resources/general_voice_articulations_profile.json` is missing a definition for `ContinuousGlissando`

Jojo-Schmitz:
Any progress on this? Just came up again in https://musescore.org/en/node/345691, and  the above comment seems to indicate where exactly the problem lies?

DmitryArefiev:
Fixed in #16597

Issue #12838, opened 2022-08-12
[MU4 Issue] Portamento playback does not work for voice


MaximilianoGG:
**Describe the bug**
The first note keeps playing (unless the same pitch was previously sounding)
No portamento playback.

**To Reproduce**
1. Insert two different and consecutive notes and add a glissando between them.
2. Click the glissando and go to Properties tab, then click Playback and then select Portamento as Style.
3. Play.
 
**Platform information**
 - OS: Xubuntu Linux 18.04.6

**Additional context**
MuseScore-4.0.0.2826032396-x86_64 version.

https://user-images.githubusercontent.com/7459884/184451438-076b4eaa-a35e-4041-89b0-a55eff29c748.mp4

MarcSabatella:
I can confirm for "voice" instruments specifically.  Seems to work more as expected for winds and strings.

MaximilianoGG:
Yes. That's right. At least for me too.

MaximilianoGG:
(Sorry, I unintentionally closed it.) Yes, the same for me, only voices.

MaximilianoGG:
I'm also experiencing no glissando playback at all (none of the styles) if the first note is tied, and not only for voices
![gliss](https://user-images.githubusercontent.com/7459884/184499788-6f0219ec-5176-4ddc-ae24-5a653bc3ebe0.png)

DmitryArefiev:
Fixed in #16597 (together with #15268)

PR #16597, opened 2023-02-28
Add definition for continuous glissando for voice

djfun:
Resolves: #15268 (and duplicates)

The definition for "ContinuousGlissando" was copied from the winds articulation profile. That seems to resolve the issue.

### DIFF
--- a/src/framework/mpe/resources/general_keyboard_articulations_profile.json
+++ b/src/framework/mpe/resources/general_keyboard_articulations_profile.json
@@ -943,6 +943,111 @@
                 }
             }
         ],
+        "ContinuousGlissando": [
+            {
+                "arrangementPattern": {
+                    "durationFactor": 9900,
+                    "timestampOffset": 0
+                },
+                "expressionPattern": {
+                    "dynamicOffsets": [
+                        {
+                            "offsetPosition": 0,
+                            "offsetValue": 0
+                        },
+                        {
+                            "offsetPosition": 1000,
+                            "offsetValue": 1500
+                        },
+                        {
+                            "offsetPosition": 2000,
+                            "offsetValue": 2500
+                        },
+                        {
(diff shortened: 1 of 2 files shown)


djfun:
@cbjeukendrup @RomanPudashkin is this going to get merged any time soon?

DmitryArefiev:
@djfun I've tried your build on Win10. **Voices** look fine now, but **Organ** and **Synths still have the problem**
Can you fix please it if it's the same?

https://github.com/musescore/MuseScore/assets/90187801/cfb64866-9e84-406f-a33e-5df38ce2e91d

Jojo-Schmitz:
Can (non-electronic) organs to portamentos?

DmitryArefiev:
In real physical life of course not (the same as piano). But for now, it should just work like in MS3.6.2

Jojo-Schmitz:
Why emulate a 3.6.2 bug?

djfun:
imho: if the option for continuous glissando is available to the user for a certain instrument, the playback should also work (or at least not do strange things like it's doing right now)

Maybe the correct choice for keyboard instruments would be to disable the option for the user, but as long as that's not the case, the better alternative is not having buggy playback.

Jojo-Schmitz:
I think this is the same as with single note dynamics, some instruments can do them, others not. So it should be part of the instrument definition

Jojo-Schmitz:
Until then: Garbage in, garbage out, of you do apply portamatone to an instrument incapable of doing it, wierd things are expected to happen

Jojo-Schmitz:
But here it seems 'All Keyboards Are Created Equal" ;-)

DmitryArefiev:
Tested #15268 #12838 on Win10, Mac13, LinuxUbuntu 22.04 - **FIXED**

PR #16629, opened 2023-03-02
Fix #15939: Add special handling for stemless slash heads

### LINKED ISSUE #15939: [MU4 Issue] Tie is not aligned correctly when using rythmic slash notation whole notes
**Describe the bug**
Tie is not aligned correctly when using rythmic slash notation whole notes

**To Reproduce**
Create 2 4/4 bars with a whole note in each. Add a tie on the first one. Then select these 2 bars and toggle rythmic slash notation.

The tie is not align correctly.

It seems to work fine with other note value.

**Expected behavior**
Tie should be centered over the rythmic notes (whole). Others values seens to be fine.

**Screenshots**
Tie 1
Tie 2
Tie 3

**Platform information**
 - OS: macOS 12.6 Macbook Pro 13 m1 2020

**Additional context**
I was using the template "Jazz Lead Sheet"

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/15939

SMuFL anchors are rough at the moment, so there are a number of special considerations that need to be dealt with regarding notehead types that may or may not have special anchors. Slashed noteheads use the stem anchors to estimate good attachment points, but of course whole notes don't have stem attach points. I've estimated 25%/75% of headwidth for these notes, which seems to be a reasonable compromise between most fonts.


oktophonie:
While we're on it would be good to sort out some of the other tie positions in a similar way, if possible.
Gonville:
image
Emmentaler:
image
Perhaps these two are missing the required anchor points; we can add them, but a fallback solution should be implemented in case we encounter other fonts with a similar problem.

Even in the case of Bravura, it would be nicer to connect the middle of the top (or bottom) of the shape rather than its outer edges (and even for the quarters, not just the halves where it's more obvious):
image

Finale Broadway illustrates this problem better:
image

oktophonie (approved):

RomanPudashkin (approved):

PR #16636, opened 2023-03-03
Fix #16317: Added tied, cross bar chords handling.

### LINKED ISSUE #16317: [MU4 Issue] Only one note in chord is retained when chord entry crosses a barline
**Describe the bug**
Only one note in chord is retained when chord entry crosses a barline.

**To Reproduce**
Steps to reproduce the behavior - adding notes with mouse:
1. Open new score in 4/4 time
2. Turn note input on, select half note duration
3. Add C starting on fourth beat of first measure - note appears as quarter note on 4th beat tied to quarter note on first beat of second measure (as expected)
4. Add A on fourth beat of first measure (to start building a chord)
5. Previous C note is deleted, and only A note appears in score
6. Add F note - this deletes the A and leaves only the F
7. Repeat process starting on first beat of measure - note that all three notes remain in chord
8. See attached score for example

Steps to reproduce the behavior - adding notes with MIDI keyboard:
1. Set up score as above
2. Add F A C chord on fourth beat of measure
3. The C note carries over to the next measure, but the A and C notes are missing on the fourth beat - they only appear on the downbeat of the second measure.
4. Repeat process starting on first beat of measure - note that chord appears properly.
5. See attached score for example.

**Expected behavior**
Full chord should carry over the barline.

**Screenshots**

**Platform information**
 - OS: Win 11
 - Bug appears in both MS 4.0.1.230121751 (revision 9b70a8c - the public MS4.0.1 release) and MS 4.0.2.230380506 (revision aabbec1)

**Additional context**
[Chords over barlines.zip](https://github.com/musescore/MuseScore/files/10713758/Chords.over.barlines.zip)

mikekirin:
Resolves: #16317 

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -657,8 +657,31 @@ void Score::expandVoice()
 
 void Score::addInterval(int val, const std::vector<Note*>& nl)
 {
-    for (Note* on : nl) {
+    // Prepare note selection in case there are not selected tied notes and sort them
+    std::vector<Note*> tmpnl;
+    bool shouldSelectFirstNote = nl.size() == 1 && nl[0]->tieFor();
+    for (auto n : nl) {
+        if (std::find(tmpnl.begin(), tmpnl.end(), n) != tmpnl.end()) {
+            continue;
+        }
+        tmpnl.push_back(n);
+        if (n->tieFor()
+            && (std::find(tmpnl.begin(), tmpnl.end(), n->tieFor()->endNote()) == tmpnl.end())) {
+            Note* currNote = n->tieFor()->endNote();
+            do {
+                tmpnl.push_back(currNote);
+                currNote = currNote->tieFor() ? currNote->tieFor()->endNote() : nullptr;
+            }while (currNote);
+        }
+    }
+
+    Note* prevTied = nullptr;
+    Chord* firstChord = nullptr;
+    for (Note* on : tmpnl) {
         Chord* chord = on->chord();
+        if (!firstChord) {
+            firstChord = chord;
+        }
         Note* note = Factory::createNote(chord);
         note->setParent(chord);
         note->setTrack(chord->track());
@@ -726,14 +749,36 @@ void Score::addInterval(int val, const std::vector<Note*>& nl)
             a->setParent(note);
             undoAddElement(a);
         }
-        setPlayNote(true);
+        if (on->tieBack() && prevTied) {
+            Tie* tie = prevTied->tieFor();
+            tie->setEndNote(note);
+            tie->setTick2(note->tick());
+            note->setTieBack(tie);
+            undoAddElement(tie);
+        }
+        if (on->tieFor()) {
+            Tie* tie = Factory::createTie(this->dummy());
+            tie->setStartNote(note);
+            tie->setTick(note->tick());
+            tie->setTrack(note->track());
+            note->setTieFor(tie);
+            prevTied = note;
+        }
 
-        select(note, SelectType::SINGLE, 0);
+        setPlayNote(true);
+        if (shouldSelectFirstNote && firstChord && !firstChord->notes().empty()) {
+            Note* noteToSelect = firstChord->notes()[firstChord->notes().size() - 1];
+            select(noteToSelect, SelectType::SINGLE, 0);
+        } else {
+            select(note, SelectType::SINGLE, 0);
+        }
     }
     if (_is.noteEntryMode()) {
         _is.setAccidentalType(AccidentalType::NONE);
     }
-    _is.moveToNextInputPos();
+    if (!shouldSelectFirstNote) {
+        _is.moveToNextInputPos();
+    }
 }
 
 //---------------------------------------------------------
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -568,6 +568,60 @@ Note* Score::addNote(Chord* chord, const NoteVal& noteVal, bool forceAccidental,
     return note;
 }
 
+Note* Score::addNoteToTiedChord(Chord* chord, const NoteVal& noteVal, bool forceAccidental, const std::set<SymId>& articulationIds,
+                                InputState* externalInputState)
+{
+    assert(!chord->notes().empty());
+    Note* referenceNote = chord->notes()[0];
+
+    while (referenceNote->tieBack()) {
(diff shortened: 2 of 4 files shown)


oktophonie:
Can confirm this fixes the problem for mouse entry! (Cannot test with a MIDI keyboard atm)

Tantacrul:
Confirmed it works for note entry. Perhaps @bkunda or @DmitryArefiev - you could quickly test on a MIDI keyboard? I'm away from my house until Monday.

cbjeukendrup:
I tried this, and now it seems impossible to input something like this:
Scherm­afbeelding 2023-03-06 om 09 45 32
which is very valid, isn't it? 

Well, it's not impossible, but when you try to enter it by adding a whole note in the second measure, you end up with:
Scherm­afbeelding 2023-03-06 om 09 47 00
And then you need to remove the upper note in the first measure to get the desired result.

So maybe this needs a little bit more thought. Personally, I find it a bit unintuitive that if I click in the second measure, a note is added in the first measure; and also that when there is already a note at the position where I'm clicking, the duration that is selected in the note input toolbar is ignored. This feels more like a kind of re-pitch mode.

mikekirin:
@cbjeukendrup It's an easy fix for that. So if I understand it rigth, by clicking to the first note of a tied chrod, there is should be an another tied note in a chord, if you cilck a second(third) note in a tied chord, there should be just that note. Am I right?

cbjeukendrup:
@mikekirin I think that would be great indeed. 

Or we might want to take another step back: whenever you click, even if there is already a tied chord in that location, add a note with the duration that is currently selected in the note input bar. So if you have a _double whole note_ selected, it would turn this:
Scherm­afbeelding 2023-03-06 om 10 12 42
into this:
Scherm­afbeelding 2023-03-06 om 10 13 02

and this:
Scherm­afbeelding 2023-03-06 om 10 14 05
into this: (when clicking in the second measure)
Scherm­afbeelding 2023-03-06 om 10 14 51

and when you have a _whole note_ selected, it would turn this:
Scherm­afbeelding 2023-03-06 om 10 16 56
into this (depending on where you click):
Scherm­afbeelding 2023-03-06 om 10 21 27
Scherm­afbeelding 2023-03-06 om 10 21 57
(not 100% sure about this last one?)

and when you have a _half note_ selected, it would turn this:
Scherm­afbeelding 2023-03-06 om 10 16 56
into (depending on where you click):
Scherm­afbeelding 2023-03-06 om 10 19 28
Scherm­afbeelding 2023-03-06 om 10 17 18




I _guess_ this would feel natural to me. But it's up to @bkunda and @Tantacrul to decide.

cbjeukendrup:
One more thing about this one:
Scherm­afbeelding 2023-03-06 om 10 14 05
when you try to add a _short note_ in the middle of it, I think it should be turned into:
Scherm­afbeelding 2023-03-06 om 10 26 51
Scherm­afbeelding 2023-03-06 om 10 27 43

mikekirin:
@Tantacrul A few more questions.
image

Adding whole + dot note to the first bar, give us tided chord. That's clear.
1. How it shold behave if I put other than whole to the first bar? What replases what?
2. How it shold behave if I put other than half note to the second bar? Do I need to remove the original note and remove tie?

Tantacrul:
> * How it shold behave if I put other than whole to the first bar? What replases what?
> * How it shold behave if I put other than half note to the second bar? Do I need to remove the original note and remove the tie?

Just so I don't misunderstand your questions: could you accompany each with a visual? I definitely don't understand question 2 I'm afraid. Sorry!

mikekirin:
Looks like now everything works as expected.

oktophonie (approved):

Tantacrul:
OK, I think there's been a bit of confusion here.

Can I ask that people don't add any extra capability requests on top of this issue from this point on just so we can get this finished and approved?

---

OK, first: the good stuff. Here I've selected a whole note duration:
image


And when entering it, I get the right result. Great! That one is in the bag
image

---

Now, in this case - this **was working great** but has been removed - probably due to confusion. I am about to place this note over the **beginning** of a tied duration.

image

... and now I get this again, which is the old behaviour.

image

I think we want this instead:

image

---

In addition, if I hold ALT+5 on this note:
image


I should get this:
image


---

Here's where I think the confusion happened.

@cbjeukendrup's point was that we shouldn't go apply ties 'backwards'. For example:

image

We shouldn't apply the tie 'backwards'. It **should** do this:

image

---

One last thing: when whole notes are tied across bars, and you try to add another note to make it a chord:
image


You get this, which is just plain bad:
image

(that is existing behaviour in MS4 by the way). I don't know why this behaves differently but it shouldn't.

mikekirin:
@Tantacrul Thank you for explanation!
The only case which is still confuses me is the last one.
There are two whole tied notes. If I add to the first one, I should get two tied whole notes as a tied chord. Ignore a duration of a note I put. Am I right? If so, I'll finish it today. There is an easy fix.
All other cases are clear.

Tantacrul:
Thanks!

I've modified the last point so it hopefully makes more sense now.

mikekirin:
Mouse input seems to be working well. Need a bit more time to investigate and fix ### **ALT +** bug.

Tantacrul:
OK, getting very close! Thanks so much for your dedication to this!

I think I just found one thing that is weird:

---

In this instance, I want to add a few intervals (let's say a few 3rds). So I press Alt+3...
image

I get the desired result but the selection moves to the end of the tied note:
image

So if I press alt+3 again... it adds the new note on the second chord.
image

I think the selection should **stay on the first chord**. This is consistent with what happens when you use Alt+num on an untied chord.

---

By the way, when I click on the second chord and press alt+3, the result is correct. Just to clarify. This is working great!

image

image

---

Thanks again!

I'm loving your attention to these kinds of problems. There is still loads of work to do across this app on refining the note input model. I could send a loooooooong laundry list!

mikekirin:
https://user-images.githubusercontent.com/16940972/224137838-22873e4f-8a7d-429b-a4c9-b1202a2bcd07.mp4

Tantacrul:
Nailed it! Thanks very much :)

Tantacrul:
If you are done @mikekirin - then @RomanPudashkin / @Eism / @cbjeukendrup - can you review this and if you are happy, merge it?

Tantacrul:
@mikekirin - I got an email from you I'm unable to respond to because there's no mechanism to do so (I don't even know how you sent it!). Could you email me at info.tantacrul@gmail.com? Then we can discuss GSoC. Thanks!

Issue #16618, opened 2023-03-02
Information copied from Help > About contains incorrect OS information


GabeS573:
### Issue type

UI bug

### Bug description

The OS information copied from Help > About contains the wrong OS name and version. It says "OS: Windows 10 Version 2009" when my OS is Windows 11, version 22H2.

### Steps to reproduce

1. Go to Help > About
2. Click the Copy button next to "Revision"
3. Paste the contents into a text editor

### Screenshots/Screen recordings

![image](https://user-images.githubusercontent.com/91428485/222302006-4a33b59f-d710-4671-8152-49a4c68591e6.png)

![image](https://user-images.githubusercontent.com/91428485/222302040-8c45678e-ba38-4217-9697-eae4e56a4b00.png)

### MuseScore Version

MuseScore version (64-bit): 4.0.2-230551516, revision: github-musescore-musescore-d23f054

### Regression

I don't know

### Operating system

Windows 11 Home, version 22H2, OS build 22621.1265

octopols:
This is what I found related to the issue [QT Bug](https://bugreports.qt.io/browse/QTBUG-99225)

Jojo-Schmitz:
It indeed is Qt's fault, we're using their OS detection.
So this will get fixed onece we move to Qt 6.x, see https://github.com/musescore/MuseScore/pull/10108

Jojo-Schmitz:
See #16649

PR #16649, opened 2023-03-03
Fix GH#16618: Information copied from Help > About contains incorrect OS information

Jojo-Schmitz:
on Windows 10 and later

Resolves: #16618

We will get "Windows 11" printed once we switched to Qt 6.2, but even then the appended " or later" is still correct (or at the very least not wrong).
Current output (with Qt 5.15.2 and on Windows 11 22H2):

OS: Windows 10 Version 2009 or later, Arch.: x86_64, MuseScore version (64-bit): 4.1.0-\<buildnumber\>, revision: github-musescore-musescore-\<commitSHA\>

### DIFF
--- a/src/appshell/view/aboutmodel.cpp
+++ b/src/appshell/view/aboutmodel.cpp
@@ -88,7 +88,9 @@ void AboutModel::copyRevisionToClipboard() const
 {
     QApplication::clipboard()->setText(
         QString("OS: %1, Arch.: %2, MuseScore version (%3-bit): %4-%5, revision: github-musescore-musescore-%6")
-        .arg(QSysInfo::prettyProductName())
+        .arg(QSysInfo::prettyProductName()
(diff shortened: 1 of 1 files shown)


src/appshell/view/aboutmodel.cpp
```diff
@@ -88,7 +88,9 @@ void AboutModel::copyRevisionToClipboard() const
 {
     QApplication::clipboard()->setText(
         QString("OS: %1, Arch.: %2, MuseScore version (%3-bit): %4-%5, revision: github-musescore-musescore-%6")
-        .arg(QSysInfo::prettyProductName())
+        .arg(QSysInfo::prettyProductName()
+             + ((QSysInfo::productType() == "windows" && (QSysInfo::productVersion() == "10" || QSysInfo::productVersion() == "11"))
+                ? " " + qtrc("appshell/about", "or later") : ""))
```
Jojo-Schmitz, line 93: No need to translate, version and revision above are not translated either

Jojo-Schmitz:
Any reason not to merge this?

igorkorsukov (approved):

Issue #16646, opened 2023-03-03
[Accessibility issue] Text not visible in DevTools under the KeyNav section


pratyaksh1610:
### Issue type

UI bug

### Bug description

`Last clicked..` text not visible in DevTools under the KeyNav section in dark mode. It is an accessibility issue which shall be taken into account.

### Steps to reproduce

Go to DevTools -> KeyNav

### Screenshots/Screen recordings

![last](https://user-images.githubusercontent.com/76740999/222714562-bc05ff66-2c43-428e-a6ab-e5ed18468477.png)

### MuseScore Version

4.1.0

### Regression

I don't know

### Operating system

Windows 10

PR #16661, opened 2023-03-04
[UI] Fix visibility issue of text in KeyNav section

pratyaksh1610:
Resolves: #16646 

Fix visibility issue of text in KeyNav section.

### DIFF
--- a/src/appshell/qml/DevTools/KeyNav/KeyNavExample.qml
+++ b/src/appshell/qml/DevTools/KeyNav/KeyNavExample.qml
@@ -45,6 +45,7 @@ Rectangle {
             anchors.margins: 8
             verticalAlignment: Text.AlignVCenter
             text: "Last clicked: " + root.lastClickedInfo
+            color: "#000000"
         }
     }
 


src/appshell/qml/DevTools/KeyNav/KeyNavExample.qml
```diff
@@ -45,6 +45,7 @@ Rectangle {
             anchors.margins: 8
             verticalAlignment: Text.AlignVCenter
             text: "Last clicked: " + root.lastClickedInfo
+            isKeyNav: true
```
cbjeukendrup, line 48: In this case, it would be better to write just `color: #000000` in this file. 

Since `color` is a property of `StyledTextLabel` (because that is a subclass of `Text`), you can set it directly in places where StyledTextLabel is instantiated. 
Because of this, and because of the fact that you need to change this property in only one specific case, it is not worth a dedicated `isKeyNav` property. So you don't need to make any changes to StyledTextLabel.qml.

pratyaksh1610: Done

cbjeukendrup (approved):

PR #16671, opened 2023-03-06
[TEST] Fluid: added linear logarifmic function for attenuation value

handrok:


cbjeukendrup:
@handrok Is this still actual? If so, it will need a rebase. 

(Also, the correct spelling is <code>logari<I>th</I>mic</code> 😉)

abariska:
@cbjeukendrup this PR is still needed for investigating audio compression in MU. It's enabled in MU3 but disabled in MU4. It's not a high priority issue, so still collecting the dust in backlog

PR #16673, opened 2023-03-06
Fix a translatable string for consitsency

Jojo-Schmitz:
Resolves: https://discord.com/channels/818804595450445834/818852942882275379/1082200046432432159


cbjeukendrup (approved):

cbjeukendrup:
I'll merge it to master, but I think it's too late for 4.0.2, because we really want to create the final build soon, probably sooner than we can reasonably expect all translators to translate the string, and if we merge it but the translators don't get the chance to translate it, then there is no translation at all. 

Instead, we can fix it using the English "translations", and add a note for translators of other languages (both like you have already done). I hope that is okay.

PR #16699, opened 2023-03-08
Live Braille module, viewer panel & liblouis 

TuanLa1972:
Adds a new panel to MuseScore's main window that displays the selected measure as music braille. Support for text and lyrics in braille is provided by liblouis, a new third-party library. The live braille panel is hidden by default and must be enabled in Preferences > Live Braille.

---

### Note for code reviewers:

Only ~40 files outside the thirdparty/liblouis directory have been changed. You can use this JS snippet to mark all files as 'viewed' and then manually 'unview' the ones you are interested in reviewing.

```JavaScript
// Mark all files as 'viewed' on a GitHub pull request.
// Paste in browser DevTools console on the PR's "Files changed" page.
document.querySelectorAll('.js-reviewed-checkbox').forEach((elem, index) => {
    if (!elem.checked) {
        var clickEvent = new MouseEvent('click');
        elem.dispatchEvent(clickEvent);
    }
    console.log(`${index} files hidden`);
});
console.log('Finished! Run the snippet again if you were expecting more files.');
```
</details>

If you scroll to the bottom of the Files page first then you'll only need to run the snippet once, but it takes longer that way.

---

image

image

### DIFF
--- a/src/braille/braillemodule.cpp
+++ b/src/braille/braillemodule.cpp
@@ -21,12 +21,23 @@
  */
 #include "braillemodule.h"
 
+#include <QQmlEngine>
+
 #include "modularity/ioc.h"
+#include "ui/iuiengine.h"
 #include "project/inotationwritersregister.h"
 
+#include "internal/brailleconfiguration.h"
+#include "internal/brailleconverter.h"
+#include "internal/notationbraille.h"
 #include "internal/braillewriter.h"
 
+#include "inotationbraille.h"
+
+#include "view/braillemodel.h"
+
 using namespace mu::engraving;
+using namespace mu::notation;
 using namespace mu::project;
 
 namespace mu::braille {
@@ -41,5 +52,35 @@ void BrailleModule::resolveImports()
     if (writers) {
         writers->reg({ "brf" }, std::make_shared<BrailleWriter>());
     }
+}
+
+void BrailleModule::registerExports()
+{
+    m_brailleConfiguration = std::make_shared<BrailleConfiguration>();
+    m_brailleConverter = std::make_shared<BrailleConverter>();
+    m_notationBraille = std::make_shared<NotationBraille>();
+
+    modularity::ioc()->registerExport<IBrailleConfiguration>(moduleName(), m_brailleConfiguration);
+    modularity::ioc()->registerExport<IBrailleConverter>(moduleName(), m_brailleConverter);
+    modularity::ioc()->registerExport<INotationBraille>(moduleName(), m_notationBraille);
+}
+
+void BrailleModule::registerUiTypes()
+{
+    using namespace notation;
+
+    qmlRegisterType<BrailleModel>("MuseScore.Braille", 1, 0, "BrailleModel");
+
+    modularity::ioc()->resolve<ui::IUiEngine>(moduleName())->addSourceImportPath(braille_QML_IMPORT);
+}
+
+void BrailleModule::onInit(const framework::IApplication::RunMode& mode)
+{
+    if (framework::IApplication::RunMode::GuiApp != mode) {
+        return;
+    }
+
+    m_brailleConfiguration->init();
+    m_notationBraille->init();
 }
 }
--- a/src/braille/internal/braille.cpp
+++ b/src/braille/internal/braille.cpp
@@ -39,6 +39,7 @@
 #include "libmscore/hairpin.h"
 #include "libmscore/jump.h"
 #include "libmscore/keysig.h"
+#include "libmscore/lyrics.h"
 #include "libmscore/marker.h"
 #include "libmscore/measure.h"
 #include "libmscore/mmrest.h"
@@ -66,7 +67,12 @@
 
 #include "containers.h"
 
+#include "louis.h"
+#include "braille.h"
+
 namespace mu::engraving {
+// Max lyrics num
+#define MAX_LYRICS_NUM                  16
 // Table 1. Page 2. Music Braille Code 2015.
 #define BRAILLE_EQUALS_METRONOME        QString("7")
 #define BRAILLE_MUSICAL_HYPEN           QString("\"")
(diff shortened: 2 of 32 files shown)


igorkorsukov:
@shoogle @TuanLa1972 We should sync this PR with master, after merge https://github.com/musescore/MuseScore/pull/17138, is right?

shoogle:
I had some trouble moving files into `src/braille` so for now I just rebased to remove liblouis.

src/appshell/view/preferences/livebraillepreferencesmodel.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "livebraillepreferencesmodel.h"
+
+#include "log.h"
+
+using namespace mu::appshell;
+
+LiveBraillePreferencesModel::LiveBraillePreferencesModel(QObject* parent)
+    : QObject(parent)
+{
+    std::sort(m_table_list.begin(), m_table_list.end());
```
igorkorsukov, line 32: Better put this code to load method 
(It is better that the constructors do nothing, so that we can create an object without cost, and perform any resource operations at the right time, when we expect that something resource-intensive can be performed. General rule)

src/appshell/view/preferences/livebraillepreferencesmodel.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+void LiveBraillePreferencesModel::load()
+{
+}
+
+bool LiveBraillePreferencesModel::liveBrailleStatus() const
+{
+    return notationConfiguration()->liveBrailleStatus();
```
igorkorsukov, line 41: Better add LiveBrailleConfiguration and use it

src/appshell/view/preferences/livebraillepreferencesmodel.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+QVariantList LiveBraillePreferencesModel::tables() const
+{
+    QVariantList result;
+
+    for (auto val : m_table_list) {
+        result << val;
+    }
+
+    return result;
```
igorkorsukov, line 77: You can use QStringList for m_table_list and return it as is

src/appshell/view/preferences/livebraillepreferencesmodel.h
```diff
@@ -0,0 +1,243 @@
...
+        "isiXhosa contracted braille [xh-za-g2.ctb]",
+        "Chinese current braille without tones, for simplified Chinese characters [zh_CHN.tbl]",
+        "Cantonese braille [zh_HK.tbl]",
+        "Taiwanese bopomofo braille [zh-tw.ctb]",
+        "Chinese common braille, for simplified Chinese characters [zhcn-cbs.ctb]",
+        "Chinese current braille with tones [zhcn-g1.ctb]",
+        "Chinese double-phonic braille [zhcn-g2.ctb]",
+        "isiZulu uncontracted braille [zu-za-g1.utb]",
+        "isiZulu contracted braille [zu-za-g2.ctb]",
+    };
```
igorkorsukov, line 239: Better move this to like LiveBrailleConfiguration and get this table in load method

src/braille/braillemodule.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "braillemodule.h"
+
+#include "modularity/ioc.h"
+
+#include "internal/brailleconverter.h"
+
+using namespace mu::engraving;
+
+namespace mu::braille {
+static std::shared_ptr<BrailleConverter> s_brailleConverter = {};
```
igorkorsukov, line 31: For a number of reasons, we are now using class members rather than static variables (see other modules for examples)

src/braille/internal/braille.cpp
```diff
@@ -20,71 +20,59 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
```
igorkorsukov, line 22: If we moved the export to a new module, then the old export module is not needed and should it be deleted?

igorkorsukov: I would recommend adding a new module and moving (deleting) the old module into two separate tasks, two separate PRs

src/braille/internal/braille.h
```diff
@@ -0,0 +1,198 @@
...
+    void addLyricsItem(Lyrics*);
+
+    bool isEmpty() { return braille_str.isEmpty(); }
+    EngravingItem* getEngravingItem(int pos);
+    std::pair<int, int> getBraillePos(EngravingItem* e);
+
+    void log();
+private:
+    QString braille_str;
+    std::vector<std::pair<EngravingItem*, std::pair<int, int> > > _items;
```
igorkorsukov, line 87: please use `m_camelCase` for members https://github.com/musescore/MuseScore/wiki/CodeGuidelines

src/braille/internal/braille.h
```diff
@@ -0,0 +1,198 @@
...
+public:
+    Braille(Score* s);
+    bool write(QIODevice& device);
+    bool convertMeasure(Measure* m, BrailleEngravingItems* beis);
+    bool convertItem(EngravingItem* el, BrailleEngravingItems* beis);
+
+private:
+    static constexpr int MAX_CHARS_PER_LINE = 40;
+
+    Score* score;
```
igorkorsukov, line 125: please use m_camelCase for members https://github.com/musescore/MuseScore/wiki/CodeGuidelines

igorkorsukov: It is better to immediately initialize to "zero" (nullptr)

src/braille/internal/braille.h
```diff
@@ -0,0 +1,198 @@
...
+
+    /* ------------ Context ------------ */
+    std::vector<Note*> previousNote;
+    std::vector<ClefType> currentCleffType;
+    std::vector<Key> currentKey;
```
igorkorsukov, line 130: please use m_camelCase for members https://github.com/musescore/MuseScore/wiki/CodeGuidelines

And maybe better add struct like Context and use it for semantic clearacy

src/notation/internal/notationconfiguration.cpp
```diff
@@ -202,6 +205,15 @@ void NotationConfiguration::init()
         m_pianoKeyboardNumberOfKeys.set(val.toInt());
     });
 
+    settings()->setDefaultValue(LIVE_BRAILLE_STATUS, Val(false));
```
igorkorsukov, line 208: Better add separate LiveBrailleConfiguration

src/notation/internal/notationlivebraille.cpp
```diff
@@ -0,0 +1,292 @@
...
+#include "braille/internal/louis.h"
+
+#include "log.h"
+
+using namespace mu::notation;
+using namespace mu::async;
+using namespace mu::engraving;
+using namespace mu::io;
+
+NotationLiveBraille::NotationLiveBraille(const Notation* notation)
```
igorkorsukov, line 53: I think this class better move to module braille/internal and interact with current notation via global context

src/braille/internal/braille.cpp
shoogle: @igorkorsukov, do you mean like PR https://github.com/musescore/MuseScore/pull/17604? The intention is for that one to get merged first, then I will drop that commit from this PR.

igorkorsukov: It is not very clear to me why to keep the old export module, in fact it has become just a wrapper.
I think it would be better to finish with the new module, merge this PR, and then move (delete) the old one. But I do not insist.

shoogle: I see. I wanted to move the export code first so that I could reduce the size of the diff when adding the new code, and so that I could keep the export tests passing throughout to make sure I didn't break anything.

What I will do is move all the export code in PR #17604, including the tests, so that the entire export module exists in `src/braille`. Then in this PR I will add the new code for the live braille.

src/braille/qml/MuseScore/Braille/BrailleView.qml
```diff
@@ -0,0 +1,150 @@
...
+                if(shortcut !== "Left" && shortcut !== "Right" &&
+                        shortcut !== "Up" && shortcut !== "Down") {
+                    brailleModel.shorcut = shortcut;
+                    event.accepted = true;
+                }
+            }
+        }
+    }
+
+    ScrollBar.vertical: ScrollBar {}
```
cbjeukendrup, line 149: Not `StyledScrollBar`?

src/braille/internal/braille.cpp
```diff
@@ -310,6 +316,292 @@ namespace mu::engraving {
...
+        item.second.first += len;
+        item.second.second += len;
+        m_items.push_back(item);
+    }
+    if (del) {
+        delete another;
+    }
+}
+
+void BrailleEngravingItems::join(std::vector<BrailleEngravingItems*> lst, bool newline, bool del)
```
cbjeukendrup, line 487: Shouldn't that be `const std::vector<BrailleEngravingItems*>&`?

src/braille/internal/braille.cpp
```diff
@@ -310,6 +316,292 @@ namespace mu::engraving {
...
+                                       "15", "124", "1245", "125", "24" };
+
+std::string Braille_LowerNumbers[] = { "356", "2", "23", "25", "256",
+                                       "26", "235", "2356", "236", "35" };
+
+std::string Braille_LyricLineIndicator = "56-23";
+
+std::string Braille_NumIndicator = "3456";
+
+std::string getBraillePattern(std::string dots)
```
cbjeukendrup, line 329: For all these functions with `std::string` or `QString` as a parameter: shouldn't it be `const std::string&` or `const QString&`?

src/braille/internal/braille.cpp
```diff
@@ -1823,14 +2407,15 @@ QString Braille::brailleMeasure(Measure* measure, int staffCount)
             resetOctave(staffCount);
 
             // Undo filling the missing beats with rests, so we don't have an altered score.
-            score->undoRedo(true, nullptr);
-            score->undoRedo(true, nullptr);
-            score->deselectAll();
+            m_score->undoRedo(true, nullptr);
+            m_score->undoRedo(true, nullptr);
+            m_score->deselectAll();
```
cbjeukendrup, line 2412: The corresponding code in line +2377 to +2387 is commented out. Shouldn't this be commented out too?

Also, there seems to be some duplication of this code in `Braille::brailleMeasure` and `Braille::brailleMeasureItems`. Can/should that be prevented?

src/braille/internal/braille.cpp
```diff
@@ -310,6 +316,292 @@ namespace mu::engraving {
...
+    for (size_t i=0; i < m_items.size(); i++) {
+        m_items[i].second.first += len;
+        m_items[i].second.second += len;
+    }
+}
+
+void BrailleEngravingItems::addEngravingItem(EngravingItem* el, QString braille)
+{
+    //braille = braille.replace(QRegularExpression ("/\\/"), "\\\\");
+    //Manual doubling slashes '\' because Regex doesn't work. Don't know why.
```
cbjeukendrup, line 524: Wouldn't you need to write `\\\\` to match a single backslash? Namely one  `\\` to escape it for C++ (which puts a single backslash in your regex), and then another `\\` to make an escaped backslash in your regex. So, QRegularExpression("\\\\") matches a single backslash. (See also the examples in https://doc.qt.io/qt-5/qregularexpression.html#introduction.)

Also, are the forward slashes part of the pattern? (I'm wondering because in JS for example, forward slashes are used as delimiters of the pattern, but in the case of `QRegularExpression`, such delimiters shouldn't be included.)

And why do you need to use regex? Doesn't the overload that takes simply two QStrings do the job?

src/braille/internal/braille.h
```diff
@@ -55,6 +57,36 @@ class TimeSig;
...
+    BrailleEngravingItems();
+    ~BrailleEngravingItems();
+
+    void clear();
+
+    void join(BrailleEngravingItems*, bool newline = true, bool del = true);
+    void join(std::vector<BrailleEngravingItems*>, bool newline = true, bool del = true);
+
+    QString brailleStr();
+    std::vector<std::pair<EngravingItem*, std::pair<int, int> > >* items();
```
cbjeukendrup, line 72: It might be helpful to add a small comment somewhere about what the `std::pair<int, int>` means.

cbjeukendrup (commented):
Here's a first round of review comments, albeit somewhat superficial.

shoogle:
@cbjeukendrup, thanks for the feedback. I've implemented most of your suggestions.

I'd prefer not to touch the internal code anymore because right now the module only does braille viewing, not braille input, but the braille input code has already been written on [another branch](https://github.com/SMCenter/MuseScore/tree/braille-input). The more I edit this PR now the harder it will be to apply the changes from the other branch later on for MuseScore 4.2. The important thing is that the code works as it is.

shoogle:
@igorkorsukov, what do you think? Can we merge this so people can start testing?

igorkorsukov (approved):

igorkorsukov:
> @igorkorsukov, what do you think? Can we merge this so people can start testing?

I think yes :)

PR #16700, opened 2023-03-08
scan_audio_plugin_option

RomanPudashkin:


src/app/commandlinecontroller.cpp
```diff
@@ -110,6 +110,9 @@ void CommandLineController::parse(const QStringList& args)
     m_parser.addOption(QCommandLineOption("test-case-func", "Call test case function", "name"));
     m_parser.addOption(QCommandLineOption("test-case-func-args", "Call test case function args", "args"));
 
+    // Audio plugin scanner
+    m_parser.addOption(QCommandLineOption("scan-audio-plugin", "Scan an audio plugin", "path"));
```
igorkorsukov, line 114: I believe a name needs to be thought of. If scan - then usually it means talking about a set. Here, as I understand it, we are talking about one plugin, which means that it is not a scan that is being done, but something else.

igorkorsukov (approved):

PR #16713, opened 2023-03-08
Fix #16701: set MIN_DISPLAYED_ZOOM_PERCENTAGE to 5

### LINKED ISSUE #16701: Status Bar - Zoom - Entering values is limited to 25%
### Steps to reproduce

1. Go to Status Bar
2. Double click on zoom % in bottom right
3. Enter 10 or 5 and hit Return - **it changes to 25%**

Expected result: It should be limited to 5% (the same limit value as Zoom minus control has)

https://user-images.githubusercontent.com/90187801/223699437-59be48f2-7465-4217-8f22-3661948deec1.mp4

### MuseScore Version

4.0.2

### Regression

### Operating system

Win/Mac/Lin

limse10:
Resolves: #16701 

Entering values in zoom controls clipped to 25 while zoom out button clipped to 5. This change sets the min zoom clamp to 5 instead of 25.

### DIFF
--- a/src/appshell/view/notationstatusbarmodel.cpp
+++ b/src/appshell/view/notationstatusbarmodel.cpp
@@ -355,7 +355,7 @@ void NotationStatusBarModel::setCurrentZoom(const QString& zoomId)
 
 int NotationStatusBarModel::minZoomPercentage() const
 {
-    return possibleZoomPercentageList().first();
+    return 5;
 }
 
 int NotationStatusBarModel::maxZoomPercentage() const


Eism:
@DmitryArefiev 
After these changes, a few items(5, 10, 15) were also added to the zoom drop-down list. If it's ok, then I will merge it

DmitryArefiev:
> @DmitryArefiev After these changes, a few items(5, 10, 15) were also added to the zoom drop-down list. If it's ok, then I will merge it

I think the drop-down list will be too long in that case. Let's leave it as it is (the smallest value is 25%). The same as in MS3

limse10:
@DmitryArefiev I have pushed updated changes. This means we are decoupling the button min zoom and the dropdown min zoom. Let me know if that is okay. Thank you!

Eism (approved):

DmitryArefiev:
@limse10 Cool! Thanks for fixing!

PR #16714, opened 2023-03-08
Ignore redundant voltas and hide voltas with print-object="no" on mxml import

asattely:
This implements the change made in https://github.com/musescore/MuseScore/pull/8622, dealing with `print-object` attributes for voltas.


shoogle:
I was a bit confused by the PR title, but from the code I see that you only ignore voltas when they are redundant. Voltas that just have the `print-object="no"` flag are hidden rather than ignored (which is good).

asattely:
I dont know what I was thinking with that original title. Fixed!

RomanPudashkin (approved):

Issue #14650, opened 2022-11-22
[MU4 Issue] Cross-staff notation w/ multiple voices causes rests to be displaced abnormally


cgpqci:
**Describe the bug**
In entering music with multiple voices, doing cross-staff notation on one of the voices cause corresponding rests at another voice to be displaced abnormally.

**To Reproduce**
1. Create score as of liking (preferably a grand staff).
2. Split measure on one of the staves into at least two voices.
3. Input a line of music on the lower voice
4. Perform cross-staff notation on lower voice
5. See error

The displacement seems to correspond to where the notes transferred to the lower staff are located. See:

Higher position

![image](https://user-images.githubusercontent.com/90231404/203209639-d184c1d2-6c67-4bbb-a4a9-0fe74cbe0b5a.png)

Lower position

![image](https://user-images.githubusercontent.com/90231404/203209791-e23575b1-6394-4ede-819a-bc1402ac2492.png)

Both staves same clef

![image](https://user-images.githubusercontent.com/90231404/203209892-7c0f26e4-b5c5-49e6-8c8b-56e6832179c8.png)

At lower staff

![image](https://user-images.githubusercontent.com/90231404/203211445-6285b36c-cf4d-4a71-86a5-9b8b46e05b33.png)
![image](https://user-images.githubusercontent.com/90231404/203211519-3a6065d7-0bbc-411e-927a-be6fbab21cd8.png)
![image](https://user-images.githubusercontent.com/90231404/203211637-683fe655-1df9-4801-adc2-ca48d8dc17d4.png)

It does kind of happen at inverse, though it isn't as apparent, and you have to set the notes lower to be so.

![image](https://user-images.githubusercontent.com/90231404/203211976-5e2b5b92-87c4-4d03-b45c-5afb53799177.png)
![image](https://user-images.githubusercontent.com/90231404/203212054-ff0f7ec7-21a8-4c40-9aaf-898a4b3c670c.png)

**Expected behavior**
Rests position at empty voices should only be respecting elements within its local staff.

**Platform information**
OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223072007, revision: github-musescore-musescore-2c34155

oktophonie:
Not a regression, and there's a more comprehensive overhaul of vertical rest displacement planned, but it would be worth looking into whether there's a quick/easy fix for this (as you say, rests just shouldn't adjust for items that are displaced to another stave).

[...]

PR #16719, opened 2023-03-09
A complete overhaul of rest positioning logic

mike-spa:
Resolves: #15301 
Resolves: #14650

This PR contains a complete overhaul of the rest positioning logic, so it probably doesn't make much sense to explain in detail everething that's being done. But just as an overview:

- The default vertical position of rest has been changed to be more consistent in several cases, especially with numbers of staff lines different than 5.
- The code managing the default position has been scrapped and rewritten because it was inadequate, and several functions were performing competing (and conflicting) logic. Now a single function determines the default position, a single function determines (if needed) the voice offset, and a single function determines the whole-rest offset (if needed).
- In multi-voice cases, rests used to be offset by 2sp up or down. Now they are offset by 1sp as default. This is configurable via a tiny new style option added to the Style -> Rest menu.
- A separate function is dedicated to resolving rest VS note vertical collisions.
- A separate function is dedicated to resolving rest VS rest vertical collisions.
- When rests are under beams, beams used to be moved away (in MuseScore 3). That is wrong, so it was temporarily removed from MuseScore 4.0. Now, _rests_ are moved away to avoid colliding with beams.
- Only if rests _can't_ be moved, beams are moved away.
- A dedicated piece of logic controls the case of two rests in two voices both beamed, as it requires a whole procedure of its own.
- Now situations with beamlets are much better managed both in terms of vertical and horizontal spacing.

A whole bunch of test failures is expected. I'm also going to add some more visual tests to cover all the new things I've done.

This PR includes all the work done in #16410 so I'm going to close that one.

(Still to come: logic for alignment of adjacent rests)

UPDATE: the last commit features an additional setting exposed in the Style -> Rest menu (as by @bkunda 's spec), made from a combination of QML and Widgets. It is literally copy-pasted from the implementation of the straight VS traditional flags in Style -> Note.

https://user-images.githubusercontent.com/93707756/225862836-ac357e73-4141-4e75-9b75-4a85bf40c7d2.mp4


oktophonie (approved):

mike-spa:
Profiler logs on Beethoven 9 score:

MASTER:
master

THIS PR:
image

abariska:
@mike-spa Seems like performance is affected. 
I also tested on Bethoven 9 symphonie. So, if talk about macOS , opening time is almost the same with master (36.8 agains 35.8), but layout recalculation time is noticeably bigger (e.g. 61sec against 50sec). I changed leading space for entire selected score from 0 to 5. Similar situation with Windows and Linux. 
Could you try to reduce layout recalculation time? 
abariska 2023-03-20 at 18 38 29

RomanPudashkin (approved):

abariska (approved):
Performance feels good now. Difference between master and the PR is no more than +-1 sec (when talking about a processes which take ~50sec)

Issue #16717, opened 2023-03-09
[Meta] Assignees get unassigned when adding a label manually


cbjeukendrup:
This keeps happening:
<img width="873" alt="Scherm­afbeelding 2023-03-09 om 11 09 28" src="https://user-images.githubusercontent.com/48658420/223990182-6aad7509-fe2e-4c58-903f-8d0a85edf0eb.png">
(That is _not_ the real Tantacrul, but GitHub Actions acting on behalf of him.)

The problem seems that the assignees4labels action does not entirely what we want. I'm currently building a replacement myself.

PR #16721, opened 2023-03-09
Issue triage: switch to our own `assign-by-label` action

cbjeukendrup:
Resolves: #16717 

I wrote my own "assign by label" action, that works slightly smarter than the one we were using previously: whenever a label was added, that one removed all assignees, and then tried to re-add them as appropriate for the current labels but the re-adding did not work well when multiple labels are specified. 
My action looks _which_ label was added or removed, and only assigns or unassigns the people who are specified for that label. It does not unassign other people that were assigned manually for example.

You can see it at https://github.com/cbjeukendrup/assign-by-label, and I have tested it at https://github.com/cbjeukendrup/MuseScore/issues/4. (edit: moved to https://github.com/musescore/github_actions/tree/main/assign-by-label)

### DIFF
--- a/.github/workflows/ci_issues_triage.yml
+++ b/.github/workflows/ci_issues_triage.yml
@@ -1,7 +1,7 @@
 name: "Issue triage"
 on:
   issues:
-    types: [opened, labeled]
+    types: [opened, labeled, unlabeled]
 
 permissions:
   issues: write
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_issues_triage.yml
```diff
@@ -19,41 +19,43 @@ jobs:
...
-    - uses: actions/checkout@v3
+  assign:
+    if: github.event.action == 'labeled' || github.event.action == 'unlabeled'
+    runs-on: ubuntu-latest
+    steps:
     - name: 'Set assignees on issue'
-      uses: kyanagimoto/assignee4label@main
+      uses: ./.github/actions/assign-by-label
       with:
-        github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
```
cbjeukendrup, line 58: Note to self: I should add this back

igorkorsukov:
I would prefer this to be added to a separate repository in the MuseScore organization. What do you think of it? Create a repository? :)

cbjeukendrup:
@igorkorsukov It was @Eism's idea to put it inside the main repo. However, I would be willing to transfer my repo from https://github.com/cbjeukendrup/assign-by-label to the `musescore` organisation.

igorkorsukov:
> @igorkorsukov It was @Eism's idea to put it inside the main repo. However, I would be willing to transfer my repo from https://github.com/cbjeukendrup/assign-by-label to the `musescore` organisation.

I think the point here is that we need an action that does our needs. At the same time, needs can change, and people can change ..., so we need to make sure that we can change this action regardless of the person who created it if necessary (for example, if you go on a long trip :) ).

So, I will now add a new repository

cbjeukendrup:
No, don't create a new repo, I'll transfer mine, I think that's easier

cbjeukendrup:
Okay, I'll try to do it this way.

cbjeukendrup:
Updated this PR; now https://github.com/musescore/github_actions/pull/1 needs to be merged first.

igorkorsukov (approved):

Eism (approved):

PR #16732, opened 2023-03-10
Fix #341456: Increase max time signature scale from 300% to 1000%

faskr:
Resolves: #341456 on musescore.org issue tracker https://musescore.org/en/node/341456

MuseScore 3 allowed for time signatures to be any size up to 1000%, and it has been requested to grant the same ability in MuseScore 4.


cbjeukendrup (approved):

octopols:
Increasing the size of the `maxValue` beyond three digits could potentially result in an issue similar to #12772

faskr:
> Increasing the size of the `maxValue` beyond three digits could potentially result in an issue similar to #12772

It seems that your solution solves the problem. Maybe that pull request should be approved as well.

PR #16743, opened 2023-03-11
fixed #16728: Corruption dialog becomes too tall when language is set to German

### LINKED ISSUE #16728: Corruption dialog becomes too tall when language is set to German
### Issue type

UI bug

### Bug description

The following corruption messaging dialog is badly resized (height is too tall) when the app language is set to German:

Screenshot 2023-03-07 at 1 32 23 pm

### Steps to reproduce

1. Set app language to German
2. Create a new score
3. Save the score locally
4. Enter some notes
5. Corrupt the score
6. Save the score

See dialog (shown in above image)

https://user-images.githubusercontent.com/86290556/224290209-26ccab05-567f-4bae-a7a9-335f22529a98.mov

### MuseScore Version

MuseScore version (64-bit): 4.0.2-230651546, revision: github-musescore-musescore-dbe7c6d

### Regression

Yes, this used to work in a previous version of Musescore 4.x

### Operating system

macOS 13.1

### Additional context

_No response_

Eism:
Resolves: #16728

This PR fixes the dialog size, but there are many warnings in qml. 

Also, there is one problem in my solution - focusing on the accent button does not occur. If the user clicks on space after opening the dialog, details info will be displayed.

### DIFF
--- a/src/framework/ui/qml/MuseScore/Ui/internal/StandardDialog.qml
+++ b/src/framework/ui/qml/MuseScore/Ui/internal/StandardDialog.qml
@@ -48,6 +48,7 @@ StyledDialogView {
 
         property int buttonId: 999
         property string title: detailsLoader.active ? qsTrc("global", "Hide details") : qsTrc("global", "Show details")
+        property bool accent: false
     }
 
     contentWidth: content.implicitWidth
@@ -78,7 +79,7 @@ StyledDialogView {
         mainPanel.readInfo()
     }
 
-    ColumnLayout {
+    Column {
         id: content
 
         width: mainPanel.width
@@ -107,8 +108,8 @@ StyledDialogView {
         Loader {
             id: detailsLoader
 
-            Layout.fillWidth: true
-            Layout.preferredHeight: visible ? implicitHeight : 0
+            width: parent.width
+            height: visible ? implicitHeight : 0
 
             active: false
             visible: active


RomanPudashkin (approved):

bkunda:
Thanks @Eism, please advise when this is ready for testing.
I've just taken a quick look (perhaps prematurely), and also noticed that not all strings are being translated:
 
Screenshot 2023-03-16 at 12 31 09 pm

Eism:
@bkunda 
Did you restart the program after changing the language?

Jojo-Schmitz:
image
image

cbjeukendrup:
@bkunda This is most likely because the translation files that are stored in the repository have not been updated for a year or so, and those files get used in nightly builds and pull request builds. 

Probably "check for language updates" works.

bkunda:
OK that did the job @cbjeukendrup.
Tested on macOS. This dialog now works correctly (insofar as its size is concerned). Thank you @Eism.

bkunda:
> image image
> 
> Not sure how to get to see the 4th button though, is is that master only?

@Jojo-Schmitz this version of the dialog is shown where there is no previously saved version to revert to (it doesn't make sense to show "revert to last saved" if the file hasn't been saved before).

Jojo-Schmitz:
That did happen on existing files

Jojo-Schmitz:
Or is this showing when a non-corrupt files gets opend, but on save (after some changes) gets detected as corrupt?

bkunda:
> That did happen on existing files
> 
> Or is this showing when a non-corrupt files gets opend, but on save (after some changes) gets detected as corrupt?

Are you using the most recent patch release?

bkunda:
> That was 4.0.2

If you could send me these files please, I will try to replicate and see what is going on.

Jojo-Schmitz:
It were scores I grabbed from the forum, as being corrupt.

One from here: https://musescore.org/en/node/346691, the 2nd file, fixed by me via manual editing. Still doesn't open in 4.0.2, but does in 3.6.2 (with corruptions being reported). So an entirely different issue, not yet reporeted, as far as I can tell.

The other is this here (don't recall anymore which thread I got it from):
[For the Beauty of the Earth.zip](https://github.com/musescore/MuseScore/files/10992888/For.the.Beauty.of.the.Earth.zip)

bkunda:
@Jojo-Schmitz ok I get it now.
These scores are already corrupted, so MuseScore won't show the "revert to last saved" option.
The "revert to last saved" feature was implemented to help prevent the user from saving a corruption into an otherwise un-corrupted score. 
But if the score is already corrupted, then all we do is remind you of that fact, and give you the control to "save anyway". 
Hope that makes sense!

Jojo-Schmitz:
Yes, it does, thanks.
Any comment about that score from the forum, why 3.6.w loads it bit 4.0.2 does not?

bkunda:
> Any comment about that score from the forum, why 3.6.w loads it bit 4.0.2 does not?

Yes, this one actually cannot be opened in MuseScore 4 (such is the nature of its specific corruption). Technical support is required to help fix a score like this (if indeed it is fixable). Hence the "get help" button 🙂

Jojo-Schmitz:
I meant: is there an issue for this on GitHub, or should one be created?
Just came accross the same issue again, see https://musescore.org/en/node/338227#comment-1181111 ff.

bkunda:
> I meant: is there an issue for this on GitHub, or should one be created? Just came accross the same issue again, see https://musescore.org/en/node/338227#comment-1181111 ff. Actually only a similar issue, here Mu4 just crashes on opening them

It seems that these are individual issues that concern specific scores, in which case the only place they will get the attention they need is either in the forum or in the #support channel on discord. Github really should be reserved for reproducible issues that are not tied to specific scores.

Issue #16371, opened 2023-02-15
[MU4 Issue] Changing to "Continuous view (vertical)" after "Page view" often results in empty window


rgreen5:
**Describe the bug**

1. Load a multipage score into MS4 in Page view mode.
2. Go to any page except page 1
3. Switch to Continuous view (vertical)

RESULT: The window is now blank. The viewer needs to adjust the viewport to get the score back into view. This happens if the original page in view is any but page 1.

**Expected result**

The viewport should not be empty.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230460504, revision: github-musescore-musescore-a36e598

oMrSmith:
+1

Tejkh1:
See PR #16752

PR #16752, opened 2023-03-12
 Fix # 16371 : Ensure score is inside viewport when notation is changed

Tejkh1:
Resolves: #16371
I fix the issue in which we were getting empty window we we change to "continuous view(vertical)" after "page  view".

- [] I created a unit test or vtest to verify the changes I made (if applicable)

### DIFF
--- a/src/notation/view/abstractnotationpaintview.cpp
+++ b/src/notation/view/abstractnotationpaintview.cpp
@@ -239,6 +239,7 @@ void AbstractNotationPaintView::onLoadNotation(INotationPtr)
     m_notation->notationChanged().onNotify(this, [this, interaction]() {
         interaction->hideShadowNote();
         m_shadowNoteRect = RectF();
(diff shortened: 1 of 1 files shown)


Tejkh1:
@RomanPudashkin  Hello Sir, it's been a long time I have made this pull request, please let me know if I have to make some more changes, actually it feels really good when our  work gets merged. 🙂

cbjeukendrup:
@Tejkh1 We decided not to merge this fix, because `m_notation->notationChanged().onNotify` happens very often and we're afraid that this will negatively impact performance. The change does make sense, but the benefits do not outweigh the impact on performance. 
Inspired by your change, I've implemented a safer version here: #20985.
If it later turns out that your version does have important benefits, we will reopen and merge this PR. 
Thanks for your contribution!

Tejkh1:
Ah, No problem And welcome

PR #16753, opened 2023-03-12
Fix a translatable string that contained a trailing space

cbjeukendrup:


src/inspector/view/qml/MuseScore/Inspector/notation/markers/MarkerSettings.qml
```diff
@@ -46,7 +46,7 @@ Column {
 
     StyledTextLabel {
         width: parent.width
-        text: qsTrc("inspector", "Marker type: ") + markerTypeToString(root.model ? root.model.type : null)
+        text: qsTrc("inspector", "Marker type:") + " " + markerTypeToString(root.model ? root.model.type : null)
```
Jojo-Schmitz, line 49: Wouldn'tz that better be
`text: qsTrc("inspector", "Marker type:") + markerTypeToString(root.model ? " " + root.model.type : null)`
i.e. add the space only if needed?

Can't we use this?
`text: qsTrc("inspector", "Marker type: %1"),arg(markerTypeToString(root.model ? root.model.type : null))`

cbjeukendrup: The first version wouldn't work, because `root.model.type` is not a string but a `PropertyItem*`. 

But good point about using .arg and RTL languages. Maybe that's better indeed.

RomanPudashkin (approved):

PR #16763, opened 2023-03-13
Backend. Compile fix

Eism:


igorkorsukov (approved):

build/ci/linux/setup.sh
```diff
@@ -24,7 +24,7 @@
 echo "Setup Linux build environment"
 trap 'echo Setup failed; exit 1' ERR
 
-GCC_VERSION="10"
+GCC_VERSION=""
 while [[ "$#" -gt 0 ]]; do
     case $1 in
         -g|--gcc_version) GCC_VERSION="$2"; shift ;;
```
cbjeukendrup, line 30: Wouldn't you need to add a dash here
```suggestion
        -g|--gcc_version) GCC_VERSION="-$2"; shift ;;
```
so that the `--gcc_version` option still works? (Otherwise the user of that option would need to add the dash)

Or should we get rid of the option entirely, if we are not using it?

PR #16715, opened 2023-03-09
Fix #16506: add inverted engraving colors for inverted score

limse10:
Resolves: #16506

Add colors to engravings when "Invert score" is checked. Inverted colors are calculated according to algorithm given by @PepijndeMaat

Here are some examples of the result.

"Invert score" unchecked:
image

"Invert score" checked:
image

- [x] I signed the [CLA](https://musescore.org/en/cla)
- [x] The title of the PR describes the problem it addresses
- [x] Each commit's message describes its purpose and effects, and references the issue it resolves
- [x] If changes are extensive, there is a sequence of easily reviewable commits
- [x] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] There are no unnecessary changes
- [x] The code compiles and runs on my machine, preferably after each commit individually

PR #16830, opened 2023-03-14
add inverted noteBackgroundColor

limse10:
Resolves: https://musescore.org/en/node/346509

This PR is separated from #16715 , preliminary discussion about this issue can be found there.

noteBackgroundColor now queries the current foreground color, so note backgrounds will always match the current score color. 

image

image

### DIFF
--- a/src/engraving/internal/engravingconfiguration.cpp
+++ b/src/engraving/internal/engravingconfiguration.cpp
@@ -201,7 +201,7 @@ Color EngravingConfiguration::thumbnailBackgroundColor() const
 
 Color EngravingConfiguration::noteBackgroundColor() const
 {
-    return Color::WHITE;
+    return notationConfiguration()->foregroundColor();
 }
 
 double EngravingConfiguration::guiScaling() const
--- a/src/engraving/internal/engravingconfiguration.h
+++ b/src/engraving/internal/engravingconfiguration.h
@@ -27,6 +27,7 @@
 #include "modularity/ioc.h"
 #include "global/iglobalconfiguration.h"
 #include "ui/iuiconfiguration.h"
+#include "notation/inotationconfiguration.h"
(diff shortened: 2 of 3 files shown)


limse10:
For further discussion, I believe it would be important for the vtests to include app-based vtests, rather than just running vtests on exported/printed scores and engravings. This would improve visibility for app-based ui issues. For example, inverted score engraving checks, as well as to verify that the app-preview and the exported scores dont have any synchronization issues (since they use different whites).

Would this be desirable/possible?

Jojo-Schmitz:
Strange, when it was part of #16715, it caused vtests failures with some Tablature tests, it doesn't here though.
Caused be the additional code in note.cpp? IIRC that wasn't in previously.

cbjeukendrup:
Note that this does not work when the user has chosen to use a wallpaper image instead of a plain color. But it's a big improvement already at least!

Jojo-Schmitz:
Do wallpapers work in inverted mode at all? Should they?

limse10:
> Caused be the additional code in note.cpp? IIRC that wasn't in previously.

Yes, I added an extra check to use pure white if the score is being exported. If not exported scores will have off-white note backgrounds for tablatures

limse10:
As for wallpapers, I would believe they wouldn't even work in normal mode, because tablature engravings draw a rectangle around the number to simulate transparency and "erase" a bit of the bar lines. Solving this problem properly would be abit more involved I believe. I can try to look into it but do share any insights you may have.

Also, what are our thoughts regarding a more comprehensive suite of vtests (as mentioned in the first comment)?

Jojo-Schmitz:
Rebase needed

Jojo-Schmitz:
Rebased in #22486

RomanPudashkin:
Closing due to a long period of inactivity + @Jojo-Schmitz rebased it

Issue #15417, opened 2022-12-20
[MU4 Issue] Resolution settings not initially present when exporting as PDF 


Leonerd-04:
**Describe the bug**
The Resolution option isn't initially available when exporting as a PDF:
![No DPI settings](https://user-images.githubusercontent.com/87563929/208618135-90bf2df2-c626-4257-b39d-f7958d02ba28.png)

**Expected behavior**
The Resolution option should be visible, below the format selector. I was able to bring it back by selecting any other format and then selecting PDF again:
![DPI settings present](https://user-images.githubusercontent.com/87563929/208618449-2c4b7841-3d04-4871-92cf-e219f6e767d6.png)

**Platform information**
 - OS: KDE Neon on Ubuntu 22.04

DmitryArefiev:
Confirm on my side. This is **regression** from **MS3**. Thanks for logging!


https://user-images.githubusercontent.com/90187801/224036668-6c7c1130-550b-4d2d-a23c-3cd97bc5550b.mp4

DmitryArefiev:
Fixed in #16837

PR #16837, opened 2023-03-15
Fix musescore#15417: Show PDF resolution settings on export dialog opening

iwoithe:
Resolves: #15417

### DIFF
--- a/src/project/qml/MuseScore/Project/internal/Export/ExportOptionsView.qml
+++ b/src/project/qml/MuseScore/Project/internal/Export/ExportOptionsView.qml
@@ -35,6 +35,10 @@ Column {
 
     spacing: 12
 
+    Component.onCompleted: {
+        pageLoader.refresh()
+    }
+
     NavigationPanel {
         id: navPanel
         name: "ExportOptions"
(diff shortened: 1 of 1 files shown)


src/project/qml/MuseScore/Project/internal/Export/ExportOptionsView.qml
```diff
@@ -183,4 +187,6 @@ Column {
             exportModel.shouldDestinationFolderBeOpenedOnExport = !checked
         }
     }
+
+    Component.onCompleted: root.refreshPageLoader()
```
cbjeukendrup, line 191: Our convention is to write `Component.onCompleted` near the top of the file; in this case, I'd put it between `spacing: 12` (line 37) and `NavigationPanel {`.

Also, we prefer to use braces, to make clear that this is a function rather than a property value, like this:
```suggestion
    Component.onCompleted: {
        root.refreshPageLoader()
    }
```

src/project/qml/MuseScore/Project/internal/Export/ExportOptionsView.qml
```diff
@@ -122,21 +126,25 @@ Column {
         id: pageLoader
         width: parent.width
 
+        function refresh() {
+            if (!root.exportModel.selectedExportType.settingsPagePath) {
+                setSource("")
```
iwoithe, line 131: @cbjeukendrup

Should I prefix this with `pageLoader` for readability, or is the object `setSource()` acts on clear enough contextually?

```suggestion
                pageLoader.setSource("")
```

(also at line 140)

cbjeukendrup: It's fine like this! Thanks!

cbjeukendrup (approved):

DmitryArefiev:
Tested #15417 on Win10, Mac13 - **FIXED**

PR #16858, opened 2023-03-16
Fix deprecation warnings in vtests CI

Jojo-Schmitz:
See for example https://github.com/musescore/MuseScore/actions/runs/4435730188

Following advice at https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


Jojo-Schmitz:
Hmm, resulted in vtests not runnig at all, strange

igorkorsukov (approved):

Issue #14055, opened 2022-10-26
[MU4 Issue] Staff type drop-down is not closed after clicking out of it when another dialog was opened


abariska:
**To Reproduce**
1. Create a simple score
2. Open the Staff type drop-down in Instruments (inspector)
3. With opened drop-down Open any dialog (Staff/Part properties, Style etc.)
4. Close the dialog 
5. Try to close Staff type dropdown by clicking out of it

**Expected behavior**
Staff type drop-down is closed after clicking out of it 

**Screenshots**

https://user-images.githubusercontent.com/101250347/198012776-0d227bef-8e97-47da-9b14-21f1e9f20483.mp4

abariska:
🟢 **FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

abariska:
🔴 Reproducible again after the PR https://github.com/musescore/MuseScore/pull/17268

Issue #14740, opened 2022-11-25
[MU4 Issue] Open menus fail to close when another dialog opens


rgreen5:
**Describe the bug**

When you click on a main menu/submenu, it persists until you click off it or press Escape. In MS3, if you had a main menu open you could not use a shortcut command – which was a safeguard.

However, in MS4, it is possible to open a menu, then open another dialog box (such as Styles) with a keyboard shortcut. The menu headings then persist across the newly-opened dialog. Not only that but the menu heading also persist across any other open program you switch to on your desktop.

**To Reproduce**

1. Create a shortcut to open the Styles dialog.
2. Click on Add > Text (main menu)
3. Now use the shortcut to open the Styles dialog
RESULT: The menu headings persist.

![menu_persists](https://user-images.githubusercontent.com/87246583/203972705-55960ca4-2b9e-4a31-8088-34d28968fb0b.png)

**Expected behavior**

Menu headings should not persist when another dialog is opened, and certainly should not persist across programs external to MuseScore.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223290405, revision: github-musescore-musescore-e57f39b

GabeS573:
I cannot reproduce this on Windows 11. Menus assigned to a shortcut correctly open in front of the Add>Text menu.

![image](https://user-images.githubusercontent.com/91428485/204107238-200bbc19-d785-44d4-9d6e-41f02d2aca11.png)

This is probably a Linux-specific problem.

rgreen5:
There's a similar issue with **Properties** panel buttons.

1. Click on a note in a score
2. Click **Appearance** in the **Properties** panel
3. Now use a keyboard shortcut to open the **Style** menu

ACTUAL RESULT: The **Appearance** tab remains open and in front of the **Style** menu.
EXPECTED RESULT: The **Appearance** tab should close.

abariska:
🟢 **FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

PR #16866, opened 2023-03-16
Close all tooltips, menus and popups when opening dialogs

Eism:
Resolves: #14055
Resolves: 
Resolves: 
Resolves:


RomanPudashkin (approved):

abariska:
@Eism There's a regression detected:
- https://github.com/musescore/MuseScore/issues/14740#issuecomment-1471830168 when using these steps dialog can't be opened by custom shortcut (except Preferences on macOS). Reproducible on all macOS, Windows and Linux

Eism:
@abariska In this case when the popup is opened, the text input field is in focus, and the component is waiting for user input, ignoring all shortcuts. You can press Esc after the popup is opened, and the text input field will lose focus, but the popup will still be open. If you try to open something using a shortcut, you will be able to do so, the popup will be closed.

abariska:
1. Shortcut is still don't work even after you lose focus by clicking on the drop-down body. 
2. I checked in another apps and active input field doesn't block dialogs when using shortcuts (Sibelius, Logic Pro)

abariska:
🟢 **TESTED** on macOS 12, Win 11, Ubuntu 20

Issue #15551, opened 2022-12-24
[MU4 Issue] "View | Fullscreen" does not appear as a toggle (Windows and Linux)


iwoithe:
**Describe the bug**
The `View | Fullscreen` menu item does not look like a toggle even though it acts as one.

About **macOS** [see comment](https://github.com/musescore/MuseScore/issues/15551#issuecomment-1711598339)

**To Reproduce**
1. Open MuseScore 4
2. Click `View | Fullscreen`
3. Open `View` again
4. A tick does not appear

https://user-images.githubusercontent.com/73422868/209420946-7a3b486f-68fa-41b4-b3e7-8bd69356da0b.mp4

**Expected behavior**

A tick to appear next to `Fullscreen`, like all of the togglable panels in the `View` menu.

**Platform information**

- Linux (Slackware 15, Xfce 4.16)

**Additional context**

- Build commit: d759f2983173fe513111d4d6ab2ce65334fd1c20

Tejkh1:
@cbjeukendrup Sir ,can you please assign it to me.

DmitryArefiev:
@Eism pointed me out that on Mac the behavior with Full Screen switching doesn't look as a bug (native macOS menu). 
It has different text for two states: **Enter Full Screen** and **Exit Full Screen**, so we can interpret as a toggle. 
Rewriting the bug to Win/Linux only (and closing)


![Screenshot 2023-09-08 at 14 56 33](https://github.com/musescore/MuseScore/assets/90187801/ba1b0f2c-fb3a-415a-bc37-1fdb514246d2)

![Screenshot 2023-09-08 at 14 56 45](https://github.com/musescore/MuseScore/assets/90187801/6eca35a3-c168-470e-9e23-5195035da582)

DmitryArefiev:
Fixed in #16870

PR #16870, opened 2023-03-16
Fix: #15551 "View | Fullscreen" appear as a toggle .

Tejkh1:
Resolves: #15551 


src/framework/ui/internal/mainwindow.cpp
```diff
@@ -87,3 +88,8 @@ QScreen* MainWindow::screen() const
 {
     return m_bridge ? m_bridge->screen() : nullptr;
 }
+
+mu::async::Notification MainWindow::isFullScreenChanged() const
+{
+    return m_isFullScreenChanged;
```
cbjeukendrup, line 94: This method just returns the notification of `MainWindow`, but `MainWindow` never sends this notification. As a result, nothing will happen. So, in `MainWindow::init`, you will need to subscribe to the notification from `m_bridge`, and when that notification is received, send the notification from `MainWindow`.

Bonus question: do you also understand why the following does ***not*** work?
```c++
mu::async::Notification MainWindow::isFullScreenChanged() const
 {
    return m_bridge ? m_bridge->isFullScreenChanged() : mu::async::Notification();
}
```

src/appshell/internal/applicationuiactions.cpp
```diff
@@ -252,6 +258,10 @@ bool ApplicationUiActions::actionEnabled(const UiAction& act) const
 
 bool ApplicationUiActions::actionChecked(const UiAction& act) const
 {
+    if (act.code == FULL_SCREEN_CODE) {
+        return mainWindow() ? mainWindow()->isFullScreen() : false;
```
cbjeukendrup, line 262: Now that `mainWindow()` exists _always_, you don't need the nullptr check anymore

Tejkh1: Done with all the changes, you can check

DmitryArefiev:
Tested #15551 on Win10, Mac13, LinuxUbuntu 22.04 - It's FIXED on Windows and Linux, but still **occurs** on **Mac**

@Tejkh1 Can you fix it in the current PR?

DmitryArefiev:
@cbjeukendrup OK. Let's merge it then. I'll modify the initial bug that it occurs only on Mac

PR #16872, opened 2023-03-17
Fix #16798: Allow beams to be joined over rests which are a quarter note or longer

### LINKED ISSUE #16798: Beams cannot be joined over rests which are a quarter or longer
### Your idea

To enable beams to be joined over rests of any duration.

### Problem to be solved

We can currently join beams over rests which are an 8th or shorter, but not over rests which are a quarter or longer:
image

In this example, applying 'join beam' to the quarter rest has no effect. This is an unfortunate limitation.

### Prior art

_No response_

### Additional context

_No response_

iwoithe:
Resolves: #16798

**Video**

https://user-images.githubusercontent.com/73422868/225829953-e63f552c-7424-403d-9762-520990474a95.mp4

**Checklist**

### DIFF
--- a/src/engraving/layout/layoutbeams.cpp
+++ b/src/engraving/layout/layoutbeams.cpp
@@ -402,7 +402,7 @@ void LayoutBeams::createBeams(Score* score, LayoutContext& lc, Measure* measure)
                 bm = BeamMode::NONE;
             }
 
-            if ((cr->durationType().type() <= DurationType::V_QUARTER) || (bm == BeamMode::NONE)) {
+            if (bm == BeamMode::NONE) {
                 bool removeBeam = true;
                 if (beam) {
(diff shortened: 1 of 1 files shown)


oktophonie:
Nice!
One inconsistency: if I set 'join beam' to this range of rests
image
I get this:
image

Whereas if I set 'join beam' to this range of rests (i.e. ending on a quarter-or-longer)
image
I get this:
image

I'm happy to say that it makes no sense to _start_ or _end_ a beam on a quarter-or-longer rest (though beams should be allowed to cross them), but in this case I'd expect the beam to at least go to the last 8th rest (i.e. like the same example).

Setting 'join beam' on the 8th here and 'start beam' on the preceding whole rest here
image
gives this, which seems correct (or at least logical):
image

PR #16882, opened 2023-03-17
scan_audio_plugins

RomanPudashkin:


src/app/app.cpp
```diff
@@ -102,7 +102,22 @@ int App::run(int argc, char** argv)
 
     QGuiApplication::styleHints()->setMousePressAndHoldInterval(250);
 
-    QApplication app(argc, argv);
+    // ====================================================
+    // Parse command line options
+    // ====================================================
+    CommandLineController commandLine;
+    commandLine.init();
+    commandLine.parse(argc, argv);
```
igorkorsukov, line 110: I believe that now this class needs to be renamed to just `CommandLineParser`, because it no longer performs an action (like a controller)

src/app/commandlinecontroller.h
```diff
@@ -25,37 +25,64 @@
...
-    INJECT(appshell, notation::INotationConfiguration, notationConfiguration)
-    INJECT(appshell, project::IProjectConfiguration, projectConfiguration)
-    INJECT(appshell, iex::guitarpro::IGuitarProConfiguration, guitarProConfiguration);
-
+    //! NOTE: This controller is created at the earliest stage of the application initialization
+    //! You should not inject anything into it
 public:
     CommandLineController() = default;
 
+    struct UiConfig {
```
igorkorsukov, line 39: It doesn't matter, but I would make one structure like this

```
    struct Options {
        struct {
            std::optional<int> physicalDotsPerInch;
        } ui;
        
        struct {
            std::optional<bool> fullMigration;
        } project;
    };
    
    Options m_options;
    
    
    m_options.ui.physicalDotsPerInch = 42;
    m_options.project.fullMigration = true;
```

src/framework/audio/internal/plugins/registeraudiopluginsscenario.cpp
```diff
@@ -0,0 +1,145 @@
...
+            continue;
+        }
+
+        AudioPluginInfo info;
+        info.meta.id = io::filename(pluginPath).toStdString();
+        info.path = pluginPath;
+        info.enabled = false;
+        info.errorCode = code;
+
+        Ret ret = knownPluginsRegister()->registerPlugin(info);
```
igorkorsukov, line 102: It seems to me that "successful" plugins should also be written here. Otherwise, it's not "those we know about", but something like "failure plugins"

igorkorsukov: looked more...
as I understand it, "successful" are register in a separate process (probe), and "not successful" are register in the main process - I believe that this can lead to misunderstandings.
I think that both should be registered in one place.
Apparently, it is better to do this in a separate process, i.e. there should be something like this:

// main proccess
```
...
const io::path_t& pluginPath = pluginPaths.at(i);
...
process()->execute(appPath, { "--audio-plugin-registre", pluginPathStr });
```

// separate process
```
    case framework::IApplication::RunMode::AudioPluginRegistre: {
        io::path_t pluginPath = commandLine.audioPluginPath();

        QMetaObject::invokeMethod(qApp, [pluginPath]() {
                LOGD() << "Scanning audio plugin: " << pluginPath;
                qApp->exit(0);
        QMetaObject::invokeMethod(qApp, [this, pluginPath]() {
                Ret ret = registerAudioPluginsScenario()->registerPlugin(pluginPath);
                if (!ret) {
                    LOGE() << ret.toString();
                }
                qApp->exit(ret.code());
            }, Qt::QueuedConnection);


mu::Ret RegisterAudioPluginsScenario::registerPlugin(const io::path_t& pluginPath)
{
    AudioPluginInfo info;
    RetVal<AudioResourceMeta> meta = reader->readMeta(pluginPath);
    if (meta.ret) {
       make success info...
    } else {
        LOGE() << meta.ret.toString();
        make fail info...
    } 

    Ret ret = knownPluginsRegister()->registerPlugin(info);
    return ret;
}
```

src/app/app.cpp
RomanPudashkin: renamed

src/app/commandlinecontroller.h
RomanPudashkin: Yes, looks much better, thanks!

src/framework/audio/internal/plugins/registeraudiopluginsscenario.cpp
```diff
@@ -0,0 +1,162 @@
...
+        io::paths_t paths = scanner->scanPlugins();
+
+        for (const io::path_t& path : paths) {
+            if (!knownPluginsRegister()->exists(path)) {
+                newPluginPaths.push_back(path);
+            }
+        }
+    }
+
+    startPluginsRegistration(newPluginPaths);
```
igorkorsukov, line 60: I think it's better to call it `processXXX`, because `start` means it's like it's an asynchronous operation, and there will be a finish somewhere.

igorkorsukov (approved):

PR #16905, opened 2023-03-18
Refactor implementation of IMainWindow interface

cbjeukendrup:
Before:
the MainWindowProvider class is created in QML, and then registered as the implementation of IMainWindow. This means that while the app is still launching, `mainWindow()` will return nullptr, making it impossible to start listening for notifications from it at the desired moment. (so we would create another initialisation step somewhere at some point where we are sure that the main window has been created blah blah...)

After:
the MainWindowProvider class is split into two parts:
- MainWindow: the implementation of the IMainWindow interface. It is always there, so that you can start using it even before the actual QML window is created.
- MainWindowHelper: created in QML, and acts as a bridge between C++ and QML. It connects the QML window to the `MainWindow` class, and performs tasks for the QML window that cannot be done in QML, such as platform specific things.

This is a prerequisite for #16870.

### DIFF
--- a/src/appshell/qml/AppWindow.qml
+++ b/src/appshell/qml/AppWindow.qml
@@ -54,13 +54,13 @@ ApplicationWindow {
         id: titleProvider
     }
 
-    MainWindowProvider {
-        id: windowProvider
+    MainWindowBridge {
+        id: bridge
 
         window: root
 
         //! NOTE These properties of QWindow (of which ApplicationWindow is derived)
-        //!      are not available in QML, so we access them via MainWindowProvider
+        //!      are not available in QML, so we access them via MainWindowBridge
         filePath: titleProvider.filePath
         fileModified: titleProvider.fileModified
     }
@@ -80,6 +80,6 @@ ApplicationWindow {
     }
 
     function showMinimizedWithSavePreviousState() {
-        windowProvider.showMinimizedWithSavePreviousState()
+        bridge.showMinimizedWithSavePreviousState()
     }
 }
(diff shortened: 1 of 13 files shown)


src/appshell/qml/AppWindow.qml
```diff
@@ -54,13 +54,13 @@ ApplicationWindow {
         id: titleProvider
     }
 
-    MainWindowProvider {
-        id: windowProvider
+    MainWindowHelper {
```
igorkorsukov, line 57: I propose to rename to `MainWindowQmlBridge` or just `MainWindowBridge`

cbjeukendrup: Done! I chose `MainWindowBridge`, because it's shorter and because it's a bidirectional bridge :)

src/framework/ui/imainwindow.h
```diff
@@ -28,13 +28,18 @@ class QWindow;
...
+class MainWindowBridge;
+
 class IMainWindow : MODULE_EXPORT_INTERFACE
 {
     INTERFACE_ID(IMainWindow)
 
 public:
     virtual ~IMainWindow() = default;
 
+    virtual void init(MainWindowBridge* helper) = 0;
```
igorkorsukov, line 40: helper -> bridge

src/framework/ui/tests/mocks/mainwindowmock.h
```diff
@@ -19,17 +19,19 @@
...
 #include <gmock/gmock.h>
 
 #include "framework/ui/imainwindow.h"
 
 namespace mu::ui {
-class MainWindowProviderMock : public IMainWindow
+class MainWindowMock : public IMainWindow
 {
 public:
+    MOCK_METHOD(void, init, (MainWindowBridge * helper), (override));
```
igorkorsukov, line 33: helper -> bridge

igorkorsukov (approved):

Issue #15736, opened 2023-01-06
[MU4 Issue] Saving parts with a dot in the name truncates at the wrong point


iainhallam:
**Describe the bug**
I name my files "Title (Composer, a. Arranger) [Instrumentation].ext", and when I exported the full score to a PDF, it correctly took that as the name. When I exported all the parts plus the full score as PDF, the new full score was named "Title (Composer, a.ext", and the parts were "Title (Composer, a-Part.ext". Clearly the code for exporting multiple parts takes a different execution path from the code for exporting a single full score, and believes that the first full stop is the end of the file name.

**To Reproduce**
1. Create a score named with another dot in the file name before the ".mscz" containing several parts
2. Export PDF to full score - the full name is respected
3. Export parts - the name is truncated

**Expected behavior**
The name including the dot should be preserved - only the last extension should be stripped off

**Platform information**
OS: Windows 10 Version 2009
Arch.: x86_64
MuseScore version (64-bit): 4.0.0-223472200
Revision: github-musescore-musescore-5485621

deswinteil:
can I try to work on this?

deswinteil:
Actually,  this the assignment statement in "exportprojectscenario.cpp" leads to the issue
![image](https://user-images.githubusercontent.com/109189977/226078494-ba24f22b-fd28-46ba-b756-cb0065dffa62.png)

![image](https://user-images.githubusercontent.com/109189977/226078468-64c666af-719d-4088-888d-10ab133abacf.png)
I will try to solve this.

deswinteil:
![image](https://user-images.githubusercontent.com/109189977/226085485-ce41559f-2cdb-43d4-9f4d-7e1085d1b20d.png)
Ok, Now I fully understand the reason for this error, it is because of the basename function, but I am not sure how to modify it 🤔

deswinteil:
![image](https://user-images.githubusercontent.com/109189977/226086421-69630cd0-3584-499d-85a2-11fcb7c70daa.png)
 I want to use **completeBasename()** to replace **basename()**,I think it should be able to solve this issue. @cbjeukendrup

Jojo-Schmitz:
IIRC a long time ago the same (or very similar) but got reported and resoloved in the very same way.
Might even have been me resolving it...

deswinteil:
> 
Hah, So my solution is corecct?

deswinteil:
[...]

PR #16906, opened 2023-03-18
fix [MU4 Issue] Saving parts with a dot in the name truncates at the …

deswinteil:
Resolves: #15736

### DIFF
--- a/src/converter/internal/convertercontroller.cpp
+++ b/src/converter/internal/convertercontroller.cpp
@@ -181,7 +181,8 @@ mu::Ret ConverterController::convertPageByPage(INotationWriterPtr writer, INotat
     TRACEFUNC;
 
     for (size_t i = 0; i < notation->elements()->pages().size(); i++) {
-        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::basename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
+        const QString filePath
+            = io::path_t(io::dirpath(out) + "/" + io::completeBasename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
 
         QFile file(filePath);
         if (!file.open(QFile::WriteOnly)) {
@@ -272,7 +273,7 @@ mu::Ret ConverterController::convertScorePartsToPngs(INotationWriterPtr writer, 
         excerpts.push_back(e->notation());
     }
 
-    io::path_t pngFilePath = io::dirpath(out) + "/" + io::path_t(io::basename(out) + "-excerpt.png");
+    io::path_t pngFilePath = io::dirpath(out) + "/" + io::path_t(io::completeBasename(out) + "-excerpt.png");
 
     for (size_t i = 0; i < excerpts.size(); i++) {
         Ret ret = convertPageByPage(writer, excerpts[i], pngFilePath);
--- a/src/autobot/internal/autobotconfiguration.cpp
+++ b/src/autobot/internal/autobotconfiguration.cpp
@@ -82,5 +82,5 @@ mu::io::path_t AutobotConfiguration::drawDataPath() const
 
 mu::io::path_t AutobotConfiguration::fileDrawDataPath(const io::path_t& filePath) const
 {
-    return drawDataPath() + "/" + io::basename(filePath) + ".json";
+    return drawDataPath() + "/" + io::completeBasename(filePath) + ".json";
 }
--- a/src/autobot/internal/autobotscriptsrepository.cpp
+++ b/src/autobot/internal/autobotscriptsrepository.cpp
@@ -51,7 +51,7 @@ Scripts AutobotScriptsRepository::scripts() const
         if (!engine.evaluate()) {
             LOGW() << "Bad script: " << p;
             s.type = ScriptType::Undefined;
-            s.title = io::basename(p).toQString();
+            s.title = io::completeBasename(p).toQString();
             scripts.push_back(std::move(s));
             continue;
         }
@@ -60,7 +60,7 @@ Scripts AutobotScriptsRepository::scripts() const
         TestCase tc(tcVal);
         if (!tc.isValid()) {
             s.type = ScriptType::Custom;
-            s.title = io::basename(p).toQString();
+            s.title = io::completeBasename(p).toQString();
             scripts.push_back(std::move(s));
             continue;
         }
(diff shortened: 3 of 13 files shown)


cbjeukendrup (approved):

cbjeukendrup:
@deswinteil I found that there is another very similar bug, namely when exporting scores on the command line: 
```
mscore -o my.score.png my.score.mscz
```
gives `my-1.png` and `my-2.png`, while I expected `my.score-1.png` and `my.score-2.png`. 

Maybe you could also fix that in this PR. 

Actually, I think there is hardly ever a reason to use `basename` rather than `completeBasename`. As far as I'm concerned, you could just replace all uses of `basename` with `completeBasename`, and maybe even remove `basename` completely to prevent future mistakes.

deswinteil:
@cbjeukendrup After modifying completeExportPath, I tried to export the png file in qtcreator. The file name is in line with my expectations (my.score-1.png, my.score-2.png). I don’t know if it is because of the command line?

src/converter/internal/convertercontroller.cpp
```diff
@@ -181,7 +181,7 @@ mu::Ret ConverterController::convertPageByPage(INotationWriterPtr writer, INotat
     TRACEFUNC;
 
     for (size_t i = 0; i < notation->elements()->pages().size(); i++) {
-        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::basename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
+        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::completeBasename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
```
cbjeukendrup, line 184: The code style check is now unhappy about this line; it has become too long. We need to insert a line break somewhere.

src/converter/internal/convertercontroller.cpp
```diff
@@ -181,7 +181,8 @@ mu::Ret ConverterController::convertPageByPage(INotationWriterPtr writer, INotat
     TRACEFUNC;
 
     for (size_t i = 0; i < notation->elements()->pages().size(); i++) {
-        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::basename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
+        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::completeBasename(out) + "-%1." \
```
cbjeukendrup, line 184: You don't need to add a backslash in C++ (that's why there are semicolons :) )

deswinteil: I just added a '\n'(entry) before, and it still failed. And I search on the Internet for what I should do when a code statement is too long, and that's the reason why I add '\' 😂

src/converter/internal/convertercontroller.cpp
```diff
@@ -181,7 +181,8 @@ mu::Ret ConverterController::convertPageByPage(INotationWriterPtr writer, INotat
     TRACEFUNC;
 
     for (size_t i = 0; i < notation->elements()->pages().size(); i++) {
-        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::basename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
+        const io::path_t filePath_temp= io::path_t(io::dirpath(out) + "/" + io::completeBasename(out) + "-%1." + io::suffix(out));
+        const QString filePath = filePath_temp.toQString().arg(i + 1);
```
cbjeukendrup, line 185: I don't understand what you're struggling with. You can add just a newline at any point in the middle of a C++ statement. For example, just before the `=`.

```suggestion
        const QString filePath 
            = io::path_t(io::dirpath(out) + "/" + io::completeBasename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
```

deswinteil: Ah, that's the point, I misunderstood the question, Thank you for answering me so patiently🥰

deswinteil: But after I modified it according to your suggestion, it's still wrong🤨

src/converter/internal/convertercontroller.cpp
```diff
@@ -181,7 +181,8 @@ mu::Ret ConverterController::convertPageByPage(INotationWriterPtr writer, INotat
     TRACEFUNC;
 
     for (size_t i = 0; i < notation->elements()->pages().size(); i++) {
-        const QString filePath = io::path_t(io::dirpath(out) + "/" + io::basename(out) + "-%1." + io::suffix(out)).toQString().arg(i + 1);
+       const QString filePath
```
cbjeukendrup, line 184: Ah, I see why: you accidentally removed a space in front of this line

cbjeukendrup: By the way, if you click on the "Details" link for the code style check, it tells you exactly what's wrong.

deswinteil: ok,I will pay attention to this

deswinteil: At least I resolve the format problem😁, but the Generate and Compare part indicates that"This PR appears to change some of the visual tests".😐I think maybe some `basename` can't be changed to `completeBasename` due to the reason that it will cause a difference in UI appearance?

cbjeukendrup: I've looked at the differences and I think they are unrelated to the PR. Could you please try rebasing (`git pull --rebase upstream master`)? I expect that this will fix it. 

(Those visual tests only compare the engraving output, not screenshots of the UI, by the way.)

deswinteil:
oops, it seems I wrongly click on the re-request button, forgive my fault😥

cbjeukendrup (approved):

DmitryArefiev:
@cbjeukendrup Do we still need a review from @igorkorsukov too?

cbjeukendrup:
I think that would be good indeed, just to be sure that I wasn't _too_ drastic with my advice to replace `basename` everywhere :)

igorkorsukov (approved):

Jojo-Schmitz:
Shouldn't those 8 commits get sqiashed? To at most 2 commits?

deswinteil:
> Shouldn't those 8 commits get sqashed? To at most 2 commits? Or are you going to do that at merge time?

sry, I will combine my commit record as soon as possible when I wake up today

deswinteil:
> Shouldn't those 8 commits get sqashed? To at most 2 commits? Or are you going to do that at merge time?

done

Jojo-Schmitz:
Well, the trick with squashing is to keep a meaningfull commit message ;-)

deswinteil:
> Well, the trick with squashing is to keep a meaningfull commit message ;-)

my negligence, I will definitely pay attention to my commit log and keep meaningful messages in the future. (Also, I'm not that familiar with the use of git, If there is anything wrong, Please feel free to point out my mistakes😄)

Issue #16841, opened 2023-03-15
Some files with rests removed from the sub voice are recognized as corrupted


bakajikara:
### Issue type

File corruption

### Steps to reproduce

1. input this
![image](https://user-images.githubusercontent.com/102702868/225331852-b2a7cd0c-48d8-4690-8bbd-24906a817750.png)

2. delete first two rests in voice 2
![image](https://user-images.githubusercontent.com/102702868/225331870-25e00976-2c2f-4a77-a0c4-6dca5ec86901.png)

3. save score
4. load score
5. file corrupted
![image](https://user-images.githubusercontent.com/102702868/225331899-481b8531-80c2-4d45-b4d8-00cd763b7fbe.png)

### MuseScore Version

Latest 4.1 Nightly build (OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.1.0-230740305, revision: github-musescore-musescore-05cb052)

### Regression

Yes, this used to work in a previous version of Musescore 4.x

### Operating system

Windows 11

### Additional context

- Does not occur in the March 14 nightly build and earlier.
- This seems to happen when 5 eighth rests are deleted.
- This also occurs with files saved in other builds.

cbjeukendrup:
@mike-spa I believe you worked on something with rests recently, might it be possible that this is related to that?

oktophonie:
(That's https://github.com/musescore/MuseScore/pull/16260)

mike-spa:
@cbjeukendrup No, the only work I did on rests is in the PR that Simon linked, and I don't think it is related (but we could try to revert it and see if indeed it is). 
What's happening here is that, when rests in the second voice are deleted, they are replaced by a "special" rest called "gap". It's treated as a rest internally, but it has no graphical output. The whole idea always seemed a bit silly to me, cause everything you can do with "gaps" you can do more easily (and more _correctly_) with invisible rests. Gaps are basically just "more invisible", and they cause a lot of different troubles, also for horizontal spacing. I suspect the corruption checks have some problem in interpreting these gaps (or the gaps themselves have problems). I didn't change anything related to gaps so far, except for details about how to handle them in spacing. I'd be glad to see them gone forever, tbh, unless they have important use cases that I'm just missing.

MarcSabatella:
[...]

PR #16913, opened 2023-03-18
Fix reading files where rests in non-first voices have been deleted

cbjeukendrup:
Resolves: #16841 

This was a regression in https://github.com/musescore/MuseScore/pull/16260, where the duration of the gap rest would be set to `d.fraction()` where `d` is some `TDuration`. This caused the gap rest to get the wrong duration, namely the duration that is closest to the desired duration and corresponds to a valid `TDuration`. However, for gap rests, the duration doesn't necessarily correspond to a valid `TDuration`; that is the reason that we can use just one to fill the whole gap. 

Summary: now we give the gap rest just the desired duration, rather than the closest duration that corresponds to a valid `TDuration`.

### DIFF
--- a/src/engraving/libmscore/check.cpp
+++ b/src/engraving/libmscore/check.cpp
@@ -270,14 +270,23 @@ void Measure::fillGap(const Fraction& pos, const Fraction& len, track_idx_t trac
          len.numerator(), len.denominator(),
          stretch.numerator(), stretch.denominator(),
          track);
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #16841 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #16927, opened 2023-03-19
Fix #16780: crash when adding instrument with two staves and then moving other instrument with two staves to the top

### LINKED ISSUE #16780: Crash on Add Instrument and Move to Top of Staff Stack
### Issue type

File corruption

### Bug description

Steps to reproduce:
1. Load attached score.
2. Activate the Instruments tab.
3. Add a piano.
4. Move the new piano to the top of the list of instruments.
5. Click [Ok]

Result (in version 4.0.2):
1. The new piano staff gets added, but with no lines.
2. Attempting to quit MU4 results in the dialog asking whether I want to save the changes. Whether I respond with [Save] or [Don't Save], MU4 crashes.
3. If I do save the score, attempting to re-open it (after restarting Musescore) results in a message that it is corrupted. Details of the corruption are shown in the second attachment.

Add Piano Corrupted Score

### MuseScore Version

4.0.2

### Regression

### Operating system

Windows 11

### Additional context

[Add Piano Crash Test.zip](https://github.com/musescore/MuseScore/files/10960982/Add.Piano.Crash.Test.zip)

cbjeukendrup:
Make the methods for inserting parts less confusing:
Before: the "index" where the part will be inserted means the staff index where the first staff of that part will be inserted

After: the "index" where the part will be inserted means the index where the part will be inserted in the list of parts

In 70% of the places, we were already using the method as if "after" was the case.

This helps to resolve https://github.com/musescore/MuseScore/issues/16780, because:
- in `NotationParts::setParts`, we get the new desired list of parts
- in `NotationParts::insertNewParts`, we add the new parts; each new part is added to the score's part list at the index where it is in the desired part list. This way, we can use that index to find back the new parts in `NotationParts::sortParts`.
- the problem is that to insert a part at the desired index, we needed to know the staff index for that part, or equivalently, the total number of staves of the parts that are currently in the score above where the new part is to be inserted. The root cause was that we were looking at the staves of the parts that are above the current part *in the desired part list* rather than *in the score's part list* (and that might lead to an invalid result, because even the existing parts in the desired list may be in a different order than in the score's list). Of course, it would be easy to change that, but I chose a more thorough solution, namely cleaning up the mess with part insert indices everywhere.

Resolves: #16780


RomanPudashkin (approved):

cbjeukendrup:
@DmitryArefiev It should be better now!

DmitryArefiev:
Yeah! It's good now!

DmitryArefiev:
Tested #16780 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #16932, opened 2023-03-20
Fix #15759: Disable "Save as default" button when multiple elements with different values selected

### LINKED ISSUE #15759: [MU4 Issue] Inspector: undefined behaviour when clicking "apply to style" with multiple elements with different values selected
**To Reproduce**
1. Create score
2. Enter notes with stems (or anything with a property that has an corresponding style setting)
3. Change stem thickness for one note to some value
4. Change stem thickness for another note to a different value
5. Select both stems
6. Try to click '...' > "Apply to style"
-> Because the current value is defined because it is different for the selected elements, an empty `PropertyValue` gets stored to style, resulting in a default stem width of zero.

**Expected behavior**
Probably the "Apply to style" button should just be disabled in this case? 

**Screenshots**

https://user-images.githubusercontent.com/48658420/211206209-76e5519f-4000-44ed-9a52-db108a87a6ac.mov

bakajikara:
Resolves: #15759

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/common/InspectorPropertyView.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/common/InspectorPropertyView.qml
@@ -142,7 +142,7 @@ Column {
                         result.push({ title: qsTrc("inspector", "Reset"), enabled: root.isModified, icon: IconCode.UNDO, id: "reset" })
 
                         if (root.isModified) {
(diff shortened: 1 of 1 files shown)


cbjeukendrup (approved):

DmitryArefiev:
Tested #15759 on Win10, Mac12 - **FIXED**

Issue #15648, opened 2022-12-29
[MU4 Issue] Staccato sign collides with fingering text


mercuree:
**Describe the bug**
Staccato sign (Add articulation: staccato) collides with fingering text (see screenshot)

**To Reproduce**
1. Add a note
2. Add staccato sign
3. Add fingering text

**Expected behavior**
Staccato sign and fingering text should not collide

**Screenshots**
![image](https://user-images.githubusercontent.com/919446/210009603-d85ad7a8-be88-4a75-b6bb-53c4411521ea.png)

**Platform information**
 - Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223472200, revision: github-musescore-musescore-5485621

PR #16972, opened 2023-03-22
Fingering vs articulation

mike-spa:
Resolves: #15648

The reason for the bug is quite trivial: we need to layout articulations before fingering, not the other way around.

I've also found the code around this area to be quite messy, so I've done some basic reorganization by extracting the fingering code (which has no point in being in LayoutSystem) into its own function LayourChord::layoutChordBaseFingering(), and made sure that things are called in the right order.

Before:
image

After:
image


oktophonie (approved):

abariska:
@mike-spa Seems like there's a regression in this PR. 
It's not possible to select a measure on the second and next pages. Works on the first page. I couldn't reproduce in few previous masters. Please, fix

https://user-images.githubusercontent.com/101250347/226992382-7e0cda32-62b5-4686-8253-cbbeeeb2893d.mp4

abariska:
It can't reproduce in the nightly build from  date 21/03

mike-spa:
That regression is caused by #16078. I'll work on it now

PR #16974, opened 2023-03-22
Fix MSVC compiler warnings

Jojo-Schmitz:

### DIFF
--- a/src/engraving/libmscore/tremolo.cpp
+++ b/src/engraving/libmscore/tremolo.cpp
@@ -503,8 +503,6 @@ void Tremolo::layoutTwoNotesTremolo(double x, double y, double h, double spatium
     UNUSED(y);
     UNUSED(h);
     UNUSED(spatium);
-    const bool defaultStyle = (!customStyleApplicable()) || (_style == TremoloStyle::DEFAULT);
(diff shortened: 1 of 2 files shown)


Jojo-Schmitz:
There's more to come...

src/engraving/libmscore/tremolo.cpp
```diff
@@ -499,7 +499,10 @@ void Tremolo::layoutOneNoteTremolo(double x, double y, double h, double spatium)
 
 void Tremolo::layoutTwoNotesTremolo(double x, double y, double h, double spatium)
 {
+    UNUSED(x);
+    UNUSED(y);
     UNUSED(h);
+    UNUSED(spatium);
```
Jojo-Schmitz, line 505: Keeping the arguments for symmetry with `layoutOneNotesTremolo()`

src/engraving/libmscore/tremolo.cpp
```diff
@@ -584,7 +585,7 @@ void Tremolo::layoutTwoNotesTremolo(double x, double y, double h, double spatium
         if (cr->isChord()) {
             Chord* chord = toChord(cr);
             staffIdx = chord->vStaffIdx();
-            int i = chord->staffMove();
+            //int i = chord->staffMove();
```
Jojo-Schmitz, line 588: Just commenting out because of the next 2 lines

Issue #14654, opened 2022-11-22
[MU4 Issue] Rhythm only mode: not possible to specify duration of rests correctly


rgreen5:
**Describe the bug**

In MS3 **Rhythm** mode, pressing a duration icon/shortcut enters a note, BUT pressing 0 (zero) has NO effect until a duration icon/shortcut is subsequently pressed, after which a rest of the corresponding duration is entered onto the staff.

However, in MS4 **Rhythm only** mode, 0 (zero) _immediately_ enters a rest. This makes it impossible to specify a duration for the rest.

**Expected behavior**

1. Clicking a duration icon or shortcut enters a note on the middle line of the staff of the same duration.
2. Pressing a . (period/full stop) makes subsequent durations dotted/undotted until pressed again.
3. Pressing a zero has no effect until a duration icon/shortcut is subsequently pressed; then a rest of the same duration is entered.

**Platform information**

OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223260405, revision: github-musescore-musescore-13f3d88

bkunda:
I can confirm I have just experienced this same problem. This is a regression from MS3 and pretty much makes "rhythm" mode unusable, because the duration of rests can't be selected before they are entered.

Tantacrul:
I'm keeping this in Post Release because we don't have time to be adding more stuff now.

oMrSmith:
> This is a regression from MS3 and pretty much makes "rhythm" mode unusable

I agree, but it's not the only problem with this mode. Even in MU3 you cannot specify the length of a following tied note.
While in rhythm-only input mode, ties would have to work _backwards_ in order for that to function. In MU4 I think you actually can specify the length of rests though, you just have to click on the rest-symbol in order to toggle rests to "on".

MarcSabatella:
The above PR address the rest shortcut issue.  It doesn't touch on the subject of ties, or change anything else about rhythm mode.

MarcSabatella:
BTW, I can confirm that only the shortcut was broken - the toolbar button does work.  Another (lame) workaround, that also works for rests on tab staves: enter the note then hit Delete.  And then hit the Right cursor key.

Issue #11736, opened 2022-05-24
[MU4 Issue] Tablature: missing methods of entering rests


MarcSabatella:
The usual shortcut to enter a rest is 0, but in tablature, this enters fret 0 instead.  So in MU3, an alternate shortcut of ";" was provided for rests in tablature input mode.  Either the command is not implemented or the shortcut is missing in MU4.

BTW, right-click is also supposed to enter a rest.  This doesn't work either on tab staves.  This much is not a regression - it was broken in MU3 also.  The lack of the keyboard method is the much more serious issue, since tab is almost always entered by keyboard anyhow.

Workaround is to first enter a note then delete it.

MarcSabatella:
Well, the workaround I gave originally (add note then immediately delete) still works

MarcSabatella:
I have to admit to being completely shocked we haven't been overrun with reports of this; I'd have assumed this would be one of the top two or three complaints we'd be seeing.  Still, it's an embarrassing omission I'd love to see us fix before people *do* notice :-).

I have to imagine it's a simple fix, but I'm not exactly sure where the breakdown is.  It seems there is no handler for the "rest-TAB" action, and that's presumably part of the problem, but I'm not sure that explains anything beyond the broken keyboard shortcut, not the broken right-click or broken toolbar button.

HemantAntony:
@MarcSabatella This is no longer reproducible. Can you check again?

MarcSabatella:
Pressing 0 definitely does not create a rest on a tablature staff - and if it did, that would
be an even more serious bug, because pressing 0 needs to enter a 0 :-).  Luckily, 0 works correctly in both 4.0.1 and in a current nightly.  Only the rest-TAB command is broken, as far as keyboard entry goes.

So yes, the rest-TAB command (shortcut semicolon) needs to work in tablature, and so should right-click and also click while in “rest mode” (after pressing the rest button on the toolbar).

I see rest-TAB is listed in the shortcut file but there is no handler for it that I can see.  So that much might be just a one line change.  But I’m not sure where the click handling happens these days.

cbjeukendrup:
[...]

PR #16978, opened 2023-03-22
fix rest commands for tab staves & rhythm mode

MarcSabatella:
Resolves: #11736
Resolves: #14654

The rest shortcuts are not functioning correctly on tab staves or in rhythm input mode.  The code is present but not hooked up properly.  This addresses those issues.

### DIFF
--- a/src/engraving/libmscore/noteentry.cpp
+++ b/src/engraving/libmscore/noteentry.cpp
@@ -92,7 +92,6 @@ NoteVal Score::noteValForPosition(Position pos, AccidentalType at, bool& error)
     }
     case StaffGroup::TAB: {
         if (_is.rest()) {
-            error = true;
             return nval;
         }
         stringData = instr->stringData();
--- a/src/notation/internal/notationactioncontroller.cpp
+++ b/src/notation/internal/notationactioncontroller.cpp
@@ -472,6 +472,7 @@ void NotationActionController::init()
     registerTabPadNoteAction("pad-note-256-TAB", Pad::NOTE256);
     registerTabPadNoteAction("pad-note-512-TAB", Pad::NOTE512);
     registerTabPadNoteAction("pad-note-1024-TAB", Pad::NOTE1024);
+    registerAction("rest-TAB", &Interaction::putRestToSelection);
 
     for (int i = 0; i < MAX_FRET; ++i) {
         registerAction("fret-" + std::to_string(i), [i, this]() { addFret(i); }, &Controller::isTablatureStaff);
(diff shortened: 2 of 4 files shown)


MarcSabatella:
I'd love to have this reviewed and hopefully merged - these are among the few remaining commands that just plain don't work.

MarcSabatella:
@bkunda Now that you're the head honcho :-), I would like to make sure this is on your radar.

src/engraving/libmscore/noteentry.cpp
```diff
@@ -92,7 +92,7 @@ NoteVal Score::noteValForPosition(Position pos, AccidentalType at, bool& error)
     }
     case StaffGroup::TAB: {
         if (_is.rest()) {
-            error = true;
+            //error = true;
```
Eism, line 95: is this still needed?

MarcSabatella: This was my question above :-).

As far as I can tell, the only purpose of this line is to prevent anyone from entering rests via mouse on tab staves,  It's been this way basically forever, so mouse entry of rests on tab has not worked in previous releases either.  MU3 even disabled the rest button on the toolbar for tab staves - they *really* didn't want you entering rests onto tab staff by mouse.  But I can't imagine why.

In MU4, the rest button on the toolbar is not disabled, so entry of rests *would* work, if not for this line arbitrarily forbidding it.  Removing this line allows mouse entry of rests to work - both using the rest button on the toolbar, and using right-click - and has no ill effects that I can see.

So I'm happy to simply remove the line rather than comment it out, but I wanted to get feedback on that first.

MarcSabatella (commented):
Somehow I marked my previous comments as needing review and they didn't get posted?  Trying again...

MarcSabatella:
@DmitryArefiev I'm still not sure what happened with my original comments not showing up at first, but I *think* I have resolved that and provided all necessary information for further review / testing.  Please let me know if there is anything else I need to do to make sure tablature users can finally enter rests!

zacjansheski:
This is working well on MacOS, re-requested code review @Eism

src/engraving/libmscore/noteentry.cpp
RomanPudashkin: > So I'm happy to simply remove the line

Yes, please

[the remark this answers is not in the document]
MarcSabatella: If you mean the error when pitch < - or when head is invalid, those are actual errors, so no change needed there.  For rests, percussion simply does a "break" to get out of the switch, then returns nval.  Returning nval right away here for tab is the same thing, just marginally more efficient, and made a smaller change.

I'll update the PR to simply delete this line within the hour.

RomanPudashkin (approved):

Issue #16900, opened 2023-03-18
let long dynamic text not collide with barlines


Vykoiazon:
### Your idea

sometimes default placement of dynamic text will overlap barlines so it could be changed to not overlap and have more space

### Problem to be solved

![image](https://user-images.githubusercontent.com/120861729/226082069-79f6f221-294b-4d99-9e9f-d5444f4d4f0a.png)

the default placement could avoid the barline so it does not have to be manually moved

PR #17001, opened 2023-03-24
A deep overhaul of dynamics and expression

mike-spa:
Resolves: #16900 (among many other things)

Summary:

1. Don't automatically flag as "other" dynamics with text. As long as the string contains the substring identifying the dynamic type, we should retain the type. **NOTE: this also resolves the issue that dynamics with text don't have playback.**

2. Keep dynamics aligned properly (i.e. dynamic under the notehead) when custom text is added. This is now the default, with an option to revert to old behaviour.

3. Offset dynamis laterally to avoide collision with barlines (with toggle/untoggle option).

In general, dynamics must be detached from text settings. Specifically:

4. Dynamics symbols now have the possibility to override the score music font (independently of text font)

5. Dynamics symbols can have custom size (independent from text size) expressed as a percentage of the default.

6. Dynamics text (i.e. text typed into a dynamic) must be treated as expression text, and it now takes all its defaults from there. The Style -> Text Styles -> Dynamics page has been removed.

7. Expressions have a new dedicated class of their own. "Old" expressions, i.e. staff text with an "expression" style, are mapped into the new expression item preserving the old layout.

8. Expressions entered on the same chord as a dynamic mark will automatically align to the dynamic (with toggle/untoggle option).

9. An expression + dynamics snapped pair is dragged together.

10. Long awaited quality-of-life improvement: when dropping (or clicking and entering) a new dynamic onto an existing one, it will replace the old one with the new one (instead of having to delete and re-enter to change a dynamic).

11. A new inspector panel for Dynamics, containing all the relevant applicable settings. Some of these settings are moved here from the Text inspector panel, mainly the Above - Below and the Frame Settings.

12. The Text inspector panel is not shown when a "pure" dynamic is selected. It is only shown when a dynamic+custom text is selected, and the options only refer to the text part.

13. New options in the Style -> Dynamics submenu.

14. A new inspector panel for Expressions, containing (for now) only the "Snap to dynamics" option.

https://user-images.githubusercontent.com/93707756/235150253-dd91e546-d551-4c98-b1cf-c924b1f7aff2.mp4

https://user-images.githubusercontent.com/93707756/235151377-a21f6d0a-61e7-41c8-b5fd-c87d770203a7.mp4


src/inspector/models/notation/dynamics/dynamicsettingsmodel.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "translation.h"
+
+using namespace mu::inspector;
+
+DynamicsSettingsModel::DynamicsSettingsModel(QObject* parent, IElementRepositoryService* repository)
+    : AbstractInspectorModel(parent, repository)
+{
+    setModelType(InspectorModelType::TYPE_DYNAMIC);
+    setTitle(qtrc("inspector ", "Dynamics \
+    "));
```
cbjeukendrup, line 33: This string looks wrong:
```suggestion
    setTitle(qtrc("inspector ", "Dynamics"));
```

src/inspector/models/notation/expressions/expressionsettingsmodel.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "translation.h"
+
+using namespace mu::inspector;
+
+ExpressionSettingsModel::ExpressionSettingsModel(QObject* parent, IElementRepositoryService* repository)
+    : AbstractInspectorModel(parent, repository)
+{
+    setModelType(InspectorModelType::TYPE_EXPRESSION);
+    setTitle(qtrc("inspector ", "Expression \
+    "));
```
cbjeukendrup, line 12: Same here
```suggestion
    setTitle(qtrc("inspector ", "Expression"));
```

oktophonie (approved):

RomanPudashkin (approved):

PR #17029, opened 2023-03-27
Fix #17004: Instrument name is distinct from long instrument name

### LINKED ISSUE #17004: Instrument Field is the same as Part Full Name
### Issue type

Other type of issue

### Bug description

Currently the Instrument Field simply displays the Full Name on the part.  This can be confusing and also represents a regression from behavior in MuseScore 3. Instead, "Instrument" should be fixed to what the instrument you have selected from the menu.

### Steps to reproduce

1.  In the Staff/Part Properties, change the part full name.
2. Save.  
3. Reopen the same Staff/Part Properties. The instrument name will now be the part full name.

_No response_

### MuseScore Version

OS: macOS 12.4, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230651546, revision: github-musescore-musescore-dbe7c6d

### Regression

### Operating system

MacOS 12.4

### Additional context

As a MuseScore user, I expect that what I select from this menu on a score:

Screen Shot 2023-03-24 at 11 29 00 AM

Would match the Instrument field shown here:

Screen Shot 2023-03-24 at 11 31 35 AM

Currently the Instrument Field simply displays the Full Name on the part.  This can be confusing and also represents a regression from behavior in MuseScore 3.  Instead, "Instrument" should be fixed to what the instrument you have selected from the menu.

Previously in MuseScore 3, if you added a Violin to a Score and then changed its Full Name to "Bassoon", the Instrument would still show as "Violin" as you have not changed instruments, only what the part name is (as seen here).  Once the Instrument field was set, the only way to change it for the staff was to select the "Change instrument" button.

Screen Shot 2023-03-24 at 11 51 28 AM

In MuseScore 4, you can now give the appearance of replacing the instrument by changing the part name text.

Screen Shot 2023-03-24 at 11 54 20 AM

Changing any variation of the part name, track name or even playback sound should not effect the Instrument Field.

To be clear, also I would also remove any ability for a number to show up in the Staff/Part Properties next to instrument as part of this update.  While a part name might be Violin 1, the Instrument selected was either Violin or Violins (section) as those are the two options when adding instruments, so one of those two is what should be shown here.  The defaults for _**part**_ naming are great, and I hope it is not misconstrued I'm suggesting otherwise

Any MusicXML import should also follow the same principles when it comes to populating the Instrument field.

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/17004

### DIFF
--- a/src/notation/view/widgets/editstaff.cpp
+++ b/src/notation/view/widgets/editstaff.cpp
@@ -213,7 +213,12 @@ void EditStaff::updateInstrument()
 
     longName->setPlainText(m_instrument.nameAsPlainText());
     shortName->setPlainText(m_instrument.abbreviatureAsPlainText());
-    instrumentName->setText(m_instrument.nameAsPlainText());
+    const InstrumentTemplate* templ = mu::engraving::searchTemplate(m_instrument.id());
+    if (templ) {
+        instrumentName->setText(formatInstrumentTitle(templ->trackName, templ->trait));
+    } else {
+        instrumentName->setText(qtrc("notation/editstaff", "Unknown"));
+    }
 
     m_minPitchA = m_instrument.minPitchA();
     m_maxPitchA = m_instrument.maxPitchA();


src/notation/view/widgets/editstaff.cpp
```diff
@@ -213,7 +213,7 @@ void EditStaff::updateInstrument()
 
     longName->setPlainText(m_instrument.nameAsPlainText());
     shortName->setPlainText(m_instrument.abbreviatureAsPlainText());
-    instrumentName->setText(m_instrument.nameAsPlainText());
+    instrumentName->setText(m_instrument.trackName());
```
shoogle, line 216: This appears to _almost_ work. You'll see the problem if you add a transposing instrument (actually two of them).

image

We should see "Instrument: Clarinet in C" but instead we only see "Instrument: Clarinet".

Now the thing is, `trackName()` should include the transposition, so the fact it doesn't is a bug. However, it should also include the instrument number (in this case "2"), which means `trackName()` isn't the right tool to be using here.

But the bigger problem is that `trackName()` is an MSCX string rather than an internal string.

shoogle (changes requested):
There's a problem with this solution: you are still relying on names stored in the MSCX file, which might have been edited by the user.

Instead, you need to use MuseScore's internal instrument names. The only thing you should trust from the MSCX is the instrument ID (the MuseScore one, not the MusicXML one). You can use the instrument ID to look up the default name of the instrument.

I think you can use [this function](https://github.com/musescore/MuseScore/blob/78392c2cfa67cd13663588e26a58bf6261b14c6a/src/notation/internal/instrumentsrepository.h#L42) to get an "instrument template" from the instrument ID, then use [this function](https://github.com/musescore/MuseScore/blob/90374217d1d6e1c389fb404263b62d1958a979fa/src/notation/notationtypes.h#L355-L375) to work out how that instrument appears in the New Score Wizard.

Something like this:

```C++
InstrumentTemplate templ = InstrumentsRepository::instrumentTemplate(id);
String displayName = formatInstrumentTitle(templ.trackName, templ.trait);
```

Note that an instrument _template's_ `trackName` is not the same as an _instrument's_ `trackName()` (confusing I know!).

Here's how you can tell if you were successful:

1. Create a score with 2 clarinets in C.
2. Save it.
3. Change MuseScore's language to "Deutsch" (German) and relaunch the program.
4. Open the file you saved.

In the 2nd clarinet's Staff/Part properties dialog, you should see:

- __Instrument:__ Klarinette in C
- __Langer Instrumentenname:  (long instrument name)__
    - Clarinet in C 2

Here the instrument title is translated because it is an internal string, whereas the long instrument name is not translated because it was loaded from the MSCX.

src/notation/view/widgets/editstaff.cpp
asattely: It took me a sec to figure out how to access the templates, since the repository functions aren't static, but I think I've done it! Doing the steps you specified for the German instrument names now produces the correct localized strings. If you could give this one a look-over one more time I'd be very grateful!

src/notation/view/widgets/editstaff.cpp
```diff
@@ -213,7 +213,12 @@ void EditStaff::updateInstrument()
 
     longName->setPlainText(m_instrument.nameAsPlainText());
     shortName->setPlainText(m_instrument.abbreviatureAsPlainText());
-    instrumentName->setText(m_instrument.nameAsPlainText());
+    const InstrumentTemplate* templ = mu::engraving::searchTemplate(m_instrument.id());
+    if (templ && !templ->musicXMLid.isEmpty()) {
```
shoogle, line 217: Why check for an empty MusicXML ID? Our instrument data (`trackName`, `trait`, etc.) is independent of MusicXML so the state of the MusicXML ID shouldn't matter.

src/notation/view/widgets/editstaff.cpp
```diff
@@ -213,7 +213,12 @@ void EditStaff::updateInstrument()
...
+        instrumentName->setText(formatInstrumentTitle(templ->trackName, templ->trait));
+    } else {
+        instrumentName->setText(m_instrument.trackName());
```
shoogle, line 220: I think it would probably best to display "Unknown" (translated string) when there's no template. The goal of this field is to tell the user what instrument MuseScore *thinks* is there, which could affect certain behaviours like playback, engraving, and pitch range warnings. If MuseScore doesn't know what the instrument is then the user needs to be told this so that they can fix it by choosing one from the dialog.

Once implemented, we won't have a translation for "Unknown", but you'll be able to test you implemented it properly by downloading the CI build and changing the language to "Placeholder translations". If you see `tr<<Unknown>>` in the program then it means translations will work once we have translations for it.
image

src/notation/view/widgets/editstaff.cpp
```diff
@@ -213,7 +213,12 @@ void EditStaff::updateInstrument()
 
     longName->setPlainText(m_instrument.nameAsPlainText());
     shortName->setPlainText(m_instrument.abbreviatureAsPlainText());
-    instrumentName->setText(m_instrument.nameAsPlainText());
+    const InstrumentTemplate* templ = mu::engraving::searchTemplate(m_instrument.id());
+    if (templ) {
+        instrumentName->setText(formatInstrumentTitle(templ->trackName, templ->trait));
+    } else {
+        instrumentName->setText(qtrc("EditStaffBase", "Unknown"));
```
shoogle, line 220: Judging by nearby files, I think the `qtrc` part of this line should be `qtrc("notation/editstaff", "Unknown")`.

The first argument to `qtrc` is a context string used for two purposes:

1. Give translators a clue about where the string appears in the program.
2. Enable the second argument (i.e. the main string) to be translated differently in different areas of the program.

Class names are often used for the context string but we are using source file paths for some reason.

asattely: Done and done!

shoogle (approved):
Works great, thanks!

PR #17033, opened 2023-03-27
Fix #16497: Include placement attributes for non-string harmonics in <technical> elements

### LINKED ISSUE #16497: [MU4 Issue] MusicXML export ignores articulation placement
**Describe the bug**
The MuseScore4 MusicXML exporter ignores the "Placement" setting of articulations and technical notations.

**To Reproduce**
1. Download and unzip the [example .mscz file](https://github.com/musescore/MuseScore/files/10808803/placement-bug.mscz.zip)
2. Open it in MuseScore and verify that it looks like: 
image
3. Export it in MusicXML format, uncompressed
4. Open the .musicxml file in MuseScore and see that the articulations and technical notations are not positioned correctly: 
image
5. Open the .musicxml file in a text or code editor and scroll to the bottom to see the `<note>` tags. Each note will have one of the two following tags:
```xml
<accent/>

<open-string/>
```

**Expected behavior**
The xml should contain the `placement` property for articulations and technical notations, i.e.
```xml
<accent placement="above"/>

<open-string placement="below"/>
```
MuseScore assumes a value of "above" if the property isn't present, but other software assumes "below", so MuseScore should always specify the placement.

If appropriate, these tags should also use other placement properties, such as `default-x`, `relative-y`, etc.

**Platform information**
OS: macOS 12.6, Arch.: x86_64, MuseScore version (64-bit): 4.0.1-230121751, revision: github-musescore-musescore-9b70a8c

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/16497

For some reason the only elements inside `<technical>` tags that saved placement information was string harmonics. The circle articulations in the file included in the issue are brass markings. This fix saves placement information for any technical articulation in the same way as normal articulations, regardless of symbol.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/exportxml.cpp
+++ b/src/importexport/musicxml/internal/musicxml/exportxml.cpp
@@ -3124,8 +3124,11 @@ void ExportMusicXml::chordAttributes(Chord* chord, Notations& notations, Technic
                 _xml.startElementRaw(mxmlTechn + attr);
                 _xml.tag("natural");
                 _xml.endElement();
-            } else {   // TODO: check additional modifier (attr) for other symbols
-                _xml.tagRaw(mxmlTechn);
+            } else {
+                if (placement != "") {
+                    attr += QString(" placement=\"%1\"").arg(placement);
+                }
+                _xml.tagRaw(mxmlTechn + attr);
             }
         }
     }


RobFog:
cc @lvinken

RomanPudashkin (approved):

PR #17035, opened 2023-03-27
Fix #16135: Allow <print-object> tags to apply to measures

### LINKED ISSUE #16135: [MU4 Issue] MusicXML import hides instruments incorrectly
**Describe the bug**
Attempting to import MusicXML that has been exported from something else, such as Finale Notepad, results in missing instruments in the score (the instruments are there as evidenced by the mixer and the instrument list, they just don't appear in the score).

**To Reproduce**
1. Export something from another program as MusicXML
2. Import it into MuseScore - instruments are missing
3. Examine instrument list panel you see the instruments exist but are hidden

A sample MusicXML file that fails to import:

[Razzle Dazzle (full).xml.gz](https://github.com/musescore/MuseScore/files/10535456/Razzle.Dazzle.full.xml.gz)

There should be full SATB+Piano. Instead you get just the piano in the score, but the full SATB+Piano in the mixer.

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/16135

Here's the plan:
1) If there is a print-object=no attribute on a measure, and there is no print-spacing=yes, then the most likely thing that the exporter was trying to convey was hidden empty staves. For that reason if a measure contains print-object=no, we simply set the style setting that hides empty staves. This may produce different results from the exporting program, based on how wide the page is and where the layout breaks fall--Dolet for example adds print-object attributes at the beginning of systems that have hidden staves, but doesn't explicitly mention system breaks, so it would be really difficult to try to reverse engineer where all of the system breaks are by investigating print-object attributes. That said this solves the problem in the issue above.

2) If the print-object=no attribute is accompanied by print-spacing=yes, this is a cutaway staff. There is very little information regarding whether each individual cutout measure should have these attributes, or if they apply until the next `print-object="yes"` attribute. Either way, there is a comment block that will apply this logic when uncommented, which I am holding off on until we have the exporter conforming to it as well. As it stands, it will only make measures that have these explicit attributes cutaway.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -2161,6 +2161,21 @@ void MusicXMLParserPass2::measure(const QString& partId, const Fraction time)
     measure->setRepeatStart(false);
     measure->setRepeatEnd(false);
 
+    /* TODO: for cutaway measures, i believe we can expect the staff to continue to be cutaway until another
+    * print-object="yes" attribute is found. Here is the code that does that, though I don't want to actually commit this until
+    * we have the exporter dealing with this sort of stuff as well.
+    *
+    * When print-object="yes" is encountered, the measure will explicitly be set to visible (see MusicXMLParserPass2::staffDetails)
+
+    MeasureBase* prevBase = measure->prev();
+    if (prevBase) {
+        Part* part = _pass1.getPart(partId);
+        staff_idx_t staffIdx = _score->staffIdx(part);
+        if (!toMeasure(prevBase)->visible(staffIdx)) {
+            measure->setStaffVisible(staffIdx, false);
+        }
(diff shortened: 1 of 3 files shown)


oktophonie (approved):

lvinken:
Please verify this works correctly with the test files I uploaded to https://github.com/musescore/MuseScore/issues/16135 a few minutes ago.

asattely:
> Please verify this works correctly with the test files I uploaded to #16135 a few minutes ago.

This PR correctly identifies all of those files as having hidden empty staves. In those xml files there's a `<print new-system="yes"/>` in each measure that MuseScore doesn't respect (so there is no system break between measures), but if you add the system breaks in manually after import, the result identical.

asattely:
I've rebased it, as well

RomanPudashkin (approved):

RomanPudashkin:
@asattely please rebase it

Issue #17038, opened 2023-03-27
The macOS version of MuseScore should use the system accent color rather than having a separate setting


ryan-roche:
### Your idea

MuseScore 4 gives the option to select an accent color to customize the program beyond the default blue. macOS has a system-level accent color selection, and it would be more intuitive for the Mac version of MuseScore to use _that_ rather than having its own, separate setting. It feels particularly redundant to have the selection in MuseScore, given that it has the exact same set of options as the macOS system selection.

### Problem to be solved

The redundancy of having two accent color selectors with the same choices for macOS users- the system level accent color and the accent color setting within MuseScore.

### Additional context

macOS 11 added a "multicolor" option for the system accent color, which defers the selection of the accent color to individual programs. I propose that this be handled by having MuseScore use some hard-coded default value (perhaps purple or blue, to match the app icon).

There is a field in the DevTools called AppleAccentColor which contains the int representation of the system accent color. I did some tinkering and I noted the following behaviour:
- Changing the system theme while MuseScore was running did nothing
- Changing the system theme, then restarting MuseScore did nothing
- Changing the system theme, then restarting MuseScore after unticking `application/hasCompletedFirstLaunchSetup` did nothing
- Across all of these experiments, the value remained at 4, which (as of macOS 11) corresponds to the "blue" option

The presence of this variable in the DevTools menu implies there was at least some work done to try and implement this, and I'd like to see if we can try to finish adding that functionality. Even if it can't update the moment that the user changes the accent color (the signal for an accent color change seems to be sent through a Swift-specific notification, and I'm not too sure how easy/hard it would be to call Swift from C++) having the macOS version of MuseScore check to make sure the matching accent color is selected seems like a good enough compromise, with real-time updates being a potential addition later on.

cbjeukendrup:
> The presence of this variable in the DevTools menu implies there was at least some work done to try and implement this

This is not the case. This just reveals that Qt's settings framework uses Apple's UserDefaults system under the hood, which is also used by macOS's own settings, which is why those macOS-related settings appear there in DevTools (which basically pulls all key/value pairs from Qt's `QSettings` thing).

C++ interop is being built into Swift right now, including calling Swift from C++, but I don't know if this has already been part of any released version of Swift, and it would be a lot of work to set up with CMake etc. Usually the easier way is to use plain old Objective-C++ (which is a bridge between C++ and Objective-C).

ryan-roche:
[...]

PR #17019, opened 2023-03-26
Fixed truncation of "Musescore version" text in the Project Propertie…

ryan-roche:
…s panel

Resolves: #15411 

- [X] I signed the [CLA](https://musescore.org/en/cla)
- [X] The title of the PR describes the problem it addresses
- [X] Each commit's message describes its purpose and effects, and references the issue it resolves
- [X] If changes are extensive, there is a sequence of easily reviewable commits
- [X] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [X] There are no unnecessary changes
- [X] The code compiles and runs on my machine, preferably after each commit individually
- [X] I created a unit test or vtest to verify the changes I made (if applicable)

PR #17056, opened 2023-03-29
Added functionality for macOS accent color setting

ryan-roche:
Resolves: #17038 

MuseScore has an accent color setting, and macOS has a system-level accent color setting. I implemented the necessary functionality for MuseScore- if running on macOS and `followSystemTheme` is enabled- to check for and respond to changes in the macOS accent color setting.

The behaviour is similar to the Dark/Light mode setting.
- If the `followSystemTheme` setting is enabled, it will check & update the value at startup (in case the user changed it while MuseScore was closed)
- If the `followSystemTheme` setting is enabled, it will recognize and respond to the NSNotification sent by macOS when the user changes their macOS accent color setting, updating the MuseScore setting accordingly.
- If the user manually selects an accent color in the preferences panel, `followSystemTheme` will be disabled
- If the user enables `followSystemTheme` using the checkbox in the preferences panel, it will immediately synchronize the MuseScore accent color setting with the macOS setting

I've attached a screencap demonstrating my implementation:
https://user-images.githubusercontent.com/48164514/228424218-f4cf4357-025a-4307-a140-51aaa82df295.mov

- [X] I created a unit test or vtest to verify the changes I made (if applicable)


src/framework/ui/internal/uiconfiguration.cpp
```diff
@@ -315,6 +330,23 @@ void UiConfiguration::synchThemeWithSystemIfNecessary()
     doSetIsDarkMode(platformTheme()->isSystemThemeDark());
 }
 
+void UiConfiguration::synchAccentColorWithSystemIfNecessary()
+{
+    if (QSysInfo::productType() != "osx" || !m_isFollowSystemTheme.val) {
```
cbjeukendrup, line 335: This check will be performed at runtime, but OS checks are better done at compile time. You can do that using the following:
```c++
#ifdef Q_OS_MAC
    // do macOS specific stuff
#endif
```
This will also fix the unit tests, because currently you are trying to use `MacOSPlatformTheme::getAccentColorIndex` on all platforms, while it only exists on macOS.

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -276,6 +277,12 @@ void AppearancePreferencesModel::setCurrentAccentColorIndex(int index)
 
     QColor color = accentColors()[index];
     uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::ACCENT_COLOR, Val(color));
+
+    // if on macOS, manually setting an accent color in this menu should disable followSystemTheme
+    LOGD() << "accent color manually changed";
+    if (QSysInfo::productType() == "osx") {
```
cbjeukendrup, line 283: (Same here about using #ifdef instead of a runtime check)

ryan-roche:
So just wrap the code in an `#ifdef` structure? Thanks for the suggestion- I'd only seen `#ifdef` used on preprocessor commands, so I assumed you could only do it on those.

ryan-roche:
That might be why the general tests are failing, too. They're likely not running on a macOS environment!

shoogle:
@ryan-roche, please can you drop the commits that are not yours? Easiest way to do this:

```Bash
git checkout macos-accent-color
git branch macos-accent-color-backup  # create backup copy of this branch (but stay on original)
git reset --hard 8bff8a57a3  # drop all commits after 8bff8a57a3  FINISHED IMPLEMENTATION
```

If you need any of the commits after `8bff8a57a3  FINISHED IMPLEMENTATION`, you can get them with:

```Bash
git cherry-pick 9276bd9749  # attempted to fix styling
git cherry-pick a8df7c8cda  # wrapped macOS-specific code in an #ifdef
git cherry-pick 5d518d5f28  # fixed code styling
```

But the point is:

- We don't want any merge commits (so avoid `8f58b9ad4c  Merge branch 'musescore:master' into macos-accent-color`).
- We don't want any commits that are already on master (so avoid all commits where you are not the author).

Once you have done all that, run this command:

```Bash
git log --oneline master..  # show commits on current branch that are not on master (--oneline is optional)
```

If this shows all the commits you need, and none of the ones you don't need, go ahead and update the PR:

```Bash
git push -f origin macos-accent-color  # force push necessary because you changed the commit history
```

ryan-roche:
That should have fixed it.

src/framework/ui/internal/uiconfiguration.h
```diff
@@ -118,6 +118,7 @@ class UiConfiguration : public IUiConfiguration, public async::Asyncable
 
     void updateSystemThemeListeningStatus();
     void synchThemeWithSystemIfNecessary();
+    void synchAccentColorWithSystemIfNecessary();
```
iwoithe, line 121: Two little things I've found:

The shortened form of  `synchronize` is `sync` (not `synch`).

src/project/qml/MuseScore/Project/internal/Properties/ProjectPropertiesFileInfoPanel.qml
```diff
@@ -78,7 +78,7 @@ ColumnLayout {
         spacing: 8
 
         PropertyItem {
-            propertyNameWidth: root.propertyNameWidth
+            propertyNameWidth: t_metrics.tightBoundingRect.width
```
iwoithe, line 81: Isn't this change part of #17019?

ryan-roche: oops- you're totally right. Not sure how that ended up in this branch- still learning git branch mechanics. Thanks for letting me know. I'll try & see if I can get rid of that in this branch

src/framework/ui/internal/uiconfiguration.h
ryan-roche: I spelled it as `synch` to match the spelling of the existing `synchThemeWithSystemIfNecessary` method name

ryan-roche:
Reverted the changes made by the first commit, as those were from a different branch and are not supposed to be in this one.

src/framework/ui/internal/uiconfiguration.h
iwoithe: @ryan-roche I'll leave this to a team member to decide. 🙂

src/appshell/view/preferences/appearancepreferencesmodel.cpp
```diff
@@ -276,6 +277,12 @@ void AppearancePreferencesModel::setCurrentAccentColorIndex(int index)
 
     QColor color = accentColors()[index];
     uiConfiguration()->setCurrentThemeStyleValue(ThemeStyleKey::ACCENT_COLOR, Val(color));
+
+    // if on macOS, manually setting an accent color in this menu should disable followSystemTheme
```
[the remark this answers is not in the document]
iwoithe: Adding to that, don't change [this](https://github.com/musescore/MuseScore/pull/17056/files#diff-627e4d4d37dc89cfb8aa4a24277aec3ad76b0ea0384f134494ad85b0970bd6c1R248) line though as macOS is the correct capitlisation of the name.

iwoithe (commented):
Two things to mention:
1. I have never used Objective-C++ before (not a macOS developer), so I can't really comment on much of the code or test this PR myself, and
2. I'm not part of the team, so an approval from me means virtually nothing, hence why I'm only commenting on the PR

 I've made a couple comments, relating to codestyle or comments. However, please take what I say with a grain of salt. 🧂

And if a team member overrides me, listen to them. 🙂

avvvvve:
Thanks for your contribution, but we're going to close this for now because the 'Follow system theme' checkbox is only meant to follow whether the system is in light or dark mode, not accent colors. It might make sense to do the latter if MuseScore used the _actual_ Mac system accent colors, but we use our own custom set of colors to ensure proper contrast.

PR #17083, opened 2023-03-31
fix #16764 Font point size fields in Properties panel only show/accept whole numbers

### LINKED ISSUE #16764: Font point size fields in Properties panel only show/accept whole numbers
### Bug description

After typing font size of 8.5 for example, it is changed to 9.

### MuseScore Version

Version: latest nightly, March 13th 

### Regression

I don't know

### Operating system

Windows 11

deswinteil:
Resolves: #16764 Font point size fields in Properties panel only show/accept whole numbers

### DIFF
--- a/src/inspector/models/text/textsettingsmodel.cpp
+++ b/src/inspector/models/text/textsettingsmodel.cpp
@@ -111,8 +111,8 @@ void TextSettingsModel::loadProperties()
     m_fontStyle->setIsEnabled(true);
 
     loadPropertyItem(m_fontSize, [](const QVariant& elementPropertyValue) -> QVariant {
-        return elementPropertyValue.toInt() == mu::engraving::TextBase::UNDEFINED_FONT_SIZE
-               ? QVariant() : elementPropertyValue.toInt();
+        return elementPropertyValue.toDouble() == mu::engraving::TextBase::UNDEFINED_FONT_SIZE
+               ? QVariant() : elementPropertyValue.toDouble();
     });
 
     m_fontSize->setIsEnabled(true);
--- a/src/inspector/models/text/textsettingsmodel.cpp
+++ b/src/inspector/models/text/textsettingsmodel.cpp
@@ -31,6 +31,7 @@
 
 #include "translation.h"
 #include "log.h"
+#include "realfn.h"
 
 using namespace mu::inspector;
 using namespace mu::engraving;
@@ -111,7 +112,7 @@ void TextSettingsModel::loadProperties()
     m_fontStyle->setIsEnabled(true);
 
     loadPropertyItem(m_fontSize, [](const QVariant& elementPropertyValue) -> QVariant {
-        return elementPropertyValue.toDouble() == mu::engraving::TextBase::UNDEFINED_FONT_SIZE
+        return RealIsEqual(elementPropertyValue.toDouble(), mu::engraving::TextBase::UNDEFINED_FONT_SIZE)
                ? QVariant() : elementPropertyValue.toDouble();
     });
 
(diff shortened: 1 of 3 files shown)


deswinteil:
After my test, this can indeed solve the problem, but I don't know if this modification will cause other problems.

cbjeukendrup:
@deswinteil Have you signed the CLA? (https://musescore.org/en/cla)
This is a requirement; if you don't sign it, we are not allowed to look at your code.

deswinteil:
@cbjeukendrup I just signed, in fact, this is the first time I know I need to sign this😣

cbjeukendrup:
Usually when you create a PR (when writing the title and description on GitHub), the default PR description contains a checklist where the CLA is mentioned.

deswinteil:
@cbjeukendrup This is indeed the case, but I saw that other people's PRs did not have this part at the beginning, so I mistakenly thought that these were not important, my fault😞I'll include that in my future PR

cbjeukendrup:
Regarding this comment:

> After my test, this can indeed solve the problem, but I don't know if this modification will cause other problems.
> Maybe we should create a new loadPropertyItem function to serve the text on the score alone🧐

I don't think these changes can cause many other problems, because the code you have changed only affects the `m_fontSize` property. 

(What happens in this code, is that `loadPropertyItem` is called with a "lambda function" (i.e. an anonymous function that's written inline) as an argument. You are only modifying this lambda function (not `loadPropertyItem` itself), and this lambda function is only used for the `m_fontSize` property.)

deswinteil:
Ah, I see. lambda function uses [] to define the function name, equivalent to a 'local' function. This is a c++11 feature that I rarely use
I at first thought it might affect other text that shouldn't be decimals.🤣

cbjeukendrup (commented):
It is correct to use `toDouble` here, since in the classes for text elements the font size is already represented by a double. However, `UNDEFINED_FONT_SIZE` is still an int, so that should actually be converted to a double too.

Another thing: when comparing doubles with `==`, a rounding error (that is inherent to floats and doubles) might occur, so that the comparison would return false while true is expected. Therefore, in `realfn.h`, we have defined the function `RealIsEqual()`, which compares two doubles with a tiny amount of tolerance, to prevent the rounding errors. It would be better to use that function.

cbjeukendrup:
One more comment: the PR contains a merge commit, namely "[Merge branch 'dev' of github.com:deswinteil/MuseScore into dev](https://github.com/musescore/MuseScore/pull/17083/commits/dce73da397e7c7c0732906f29d282becc8325fba)". The preferred way to update a branch for a pull request is rebasing: the command you need is 
```git
git pull --rebase upstream master
```

deswinteil:
Ok, I will submit a new commit tomorrow

src/inspector/models/text/textsettingsmodel.cpp
```diff
@@ -111,8 +112,8 @@ void TextSettingsModel::loadProperties()
     m_fontStyle->setIsEnabled(true);
 
     loadPropertyItem(m_fontSize, [](const QVariant& elementPropertyValue) -> QVariant {
-        return elementPropertyValue.toInt() == mu::engraving::TextBase::UNDEFINED_FONT_SIZE
-               ? QVariant() : elementPropertyValue.toInt();
+        return RealIsEqual(elementPropertyValue.toDouble(),double(mu::engraving::TextBase::UNDEFINED_FONT_SIZE))
```
cbjeukendrup, line 115: I meant changing the _definition_ of `UNDEFINED_FONT_SIZE` to double.

deswinteil: OK

RomanPudashkin (approved):

DmitryArefiev:
Tested #16764 on Win10 - **FIXED**

PR #17105, opened 2023-04-01
Create a different latest nightly symlink for each branch

shoogle:
Affects [nightly builds](https://musescore.org/en/nightly-builds) uploaded to the OSUOSL server.

Currently we create a single 'latest' symlink on each platform. E.g. on macOS:

    MuseScoreNightly-latest-x86_64.dmg

Now we will create one for each branch:

    MuseScoreNightly-latest-master-x86_64.dmg
    MuseScoreNightly-latest-4.0.2-x86_64.dmg

This helps to avoid confusion, and also avoids the race condition where the 'latest' symlink sometimes points to a master nightly and sometimes to a 4.0.2 nightly depending on the time of day.

On Linux, for similar reasons, a separate zsync automatic updates file will be created for each branch.

The current "branchless" symlink and zync files will remain unmodified on the server until they are eventually deleted by the same code that is used to delete old nightly files on each upload.

### DIFF
--- a/build/ci/linux/package.sh
+++ b/build/ci/linux/package.sh
@@ -86,7 +86,7 @@ if [ "$PACKTYPE" == "appimage" ]; then
     # https://github.com/AppImage/AppImageSpec/blob/master/draft.md#update-information
     case "${BUILD_MODE}" in
     "stable_build")  export UPDATE_INFORMATION="gh-releases-zsync|musescore|MuseScore|latest|MuseScore-*${PACKARCH}.AppImage.zsync";;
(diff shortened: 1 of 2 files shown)


Jojo-Schmitz:
Why still building 4.0.2 in the first place?

shoogle:
@Jojo-Schmitz, you're right, we should disable the 4.0.2 build. However, the changes in this PR are still necessary as there are bound to be builds on other branches in the future.

igorkorsukov (approved):

PR #17108, opened 2023-04-02
fixes #17107, add traitName fields to lutes to avoid confusion in Spanish

### LINKED ISSUE #17107: Unable to select the correct laúd variant in Spanish language
### Issue type

UI bug

### Bug description

When selecting Laúd instrument in Spanish, there is a drop-down with two empty elements, making very difficult to select the appropriate instrument.

It does not happens in english since the two options are labeled as 'Lute' for the classic lute, and 'Laúd' for the spanish (aka modern) lute, but in Spanish Lute is translated as Laúd, so both instruments are labeled the same in the UI.

The workaround is to select any of the empty options in the dropdown and, after adding the instrument to the system, verify that the tuning of the staff is the correct one (A1 for the Classic lute, or G#2 for the Spanish laúd).

### Steps to reproduce

1. open musescore
2. go to preferences and select 'Spanish' language
3. restart musescore
4. create a new score, and in the dialog filter by 'Laúd'
5. You can see the two empty options drop-down menu

**Selecting Lute or Laúd in English**
lute-en

**Selecting Laúd in Spanish**

lute-es

### MuseScore Version

OS: macOS 13.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230651546, revision: github-musescore-musescore-dbe7c6d

### Regression

Choose option...

### Operating system

macOS

### Additional context

_No response_

manolo:
Resolves: #17107

Adds labels to the empty dropdown options when selecting 'Laúd' instrument in Spanish, as it is shown in the gif screenshot:

lute-es-fixed

### DIFF
--- a/share/instruments/instruments.xml
+++ b/share/instruments/instruments.xml
@@ -12373,6 +12373,7 @@
                   <trackName>Lute</trackName>
                   <longName>Lute</longName>
                   <shortName>Lt.</shortName>
+                  <traitName type="variant">Classic</traitName>
                   <description>6-course lute (staff notation).</description>
                   <musicXMLid>pluck.lute</musicXMLid>
                   <StringData>
@@ -13170,6 +13171,7 @@
                   <trackName>Laúd</trackName>
                   <longName>Laúd</longName>
                   <shortName>Laúd</shortName>
+                  <traitName type="variant">Spanish</traitName>
                   <description>Spanish lute similar to the bandurria. (Staff notation).</description>
                   <musicXMLid>pluck.laud</musicXMLid>
                   <StringData>


shoogle (commented):
Thanks for identifying the problem and providing a solution! However, the `traitName` "type" attribute has a special meaning in the C++ code, and it may only take one of [these values](https://github.com/musescore/MuseScore/wiki/Translating-instrument-names#trait-type).

I think it would be better to fix this problem in the translation, by providing a different `trackName` for "Lute" compared to "Laúd" to ensure they appear as different instruments in the New Score Wizard (NSW).

You can translate the `longName` as "Laúd" in both cases if you want the instruments to look the same in the score.


For example:

### Lute

Component | English | *Spanish*
:---|:---|:---
`trackName` (NSW) | Lute | *Laúd*
`longName` (score) | Lute | *Laúd*

### Laúd

Component | English | *Spanish*
:---|:---|:---
`trackName` (NSW) | Laúd | *Laúd (español)*
`longName` (score) | Laúd | *Laúd*

Or you could use something like *Laúd (folclórico)* as the `trackName` if you prefer.


(Sorry if my suggestions don't make sense in Spanish).

Jojo-Schmitz:
Brings up another issue; Shouldn't the (English) "Lauds" rather be "Laúds"

Jojo-Schmitz:
OK, I've fixed the translations (es, en_US, en_GB, de), as per this conversation

manolo:
@Jojo-Schmitz so what is your change, is there a PR?

Jojo-Schmitz:
No, just update your translations and see

manolo:
Fantastic, translation is a much better solution.

By the way, regarding which family should group these Spanish (also popular in Latin America) instruments, I am not sure if they should be named 'laudes' or 'bandurrias', actually, Spanish Lute (aka Laúd) and Bandurria, among other variants (actually 5 instruments in total) share the same family, although bandurria and laúd are the most used.
I would send a PR soon with the missing instruments.

The thing is that, there is no consensus, and some authors group them as bandurrias, and others as laudes, thus both are valid. 
My opinion is that grouping as bandurrias would cause less confusion, because we use Laúd for the classical (baroque and renaissance) lute family.

So I would say:

bandurrias Family
- Bandurria Sopranino (bandurrín)
- Bandurria Soprano(bandurria)
- Bandurria Contralto (Laudín)
- Bandurria Tenor (Laúd) 
- Bandurria Barítono (Laudón)

shoogle:
Great, I'm glad it's working for you!

When it comes to instrument families, @oktophonie is the person to ask, but it's worth noting that our instrument families exist for the sole purpose of making the notation look right (e.g. in terms of instrument order, staff bracketing, sharing staves, etc.). We're not aiming to follow any existing system of classification, as these tend to be based on the instrument's physical characteristics rather than on how it is notated.

Closing for now since we decided to go with a different solution.

Jojo-Schmitz:
@shoogle are you taking care of that Lauds -> Laúds change?

Issue #17037, opened 2023-03-27
Flats and Text Fonts in MusicXML Import


smorell-1:
### Your idea

When it comes to importing MusicXML, our goal is to make seamless transitions into MuseScore.  MuseScore 4 does not do very well capturing the finer points of how instrument names are presented when it comes to flat symbols, and should use Edwin more often to make for better looking imports when MusicXML files are encoded to use certain default text styles.

### Problem to be solved

When exporting from other notation software, it is common for parts like a B-Flat Clarinet to presented like this in the part names:

B♭ Clarinet
B♭ Clarinet 1
Clarinet in B♭ 1

This formatting is generally ignored in MusicXML exports and it commonly converted to "Bb".

Finale, employs a method to try to tell other notation products that may be importing it to use the flat symbol by including a _display_ element also as seen here:

<score-part id="P6">
      <part-name>Eb Clarinet</part-name>
      <part-name-display>
        <display-text>E</display-text>
        <accidental-text>flat</accidental-text>
        <display-text> Clarinet</display-text>
      </part-name-display>
      <part-abbreviation>Eb Cl.</part-abbreviation>
      <part-abbreviation-display>
        <display-text>E</display-text>
        <accidental-text>flat</accidental-text>
        <display-text> Cl.</display-text>
      </part-abbreviation-display>

This approach tries to incorporate a key element (the flat sign) that reflects the very different approach than MuseScore for how to present a flat sign in an instrument name.  For MuseScore's purposes, it makes sense to continue simply using the part name as the main driver, but to present certain text more elegantly for the two most common flats used in Instrument Names, B♭ and E♭.

"Bb" when separated by space should be imported as "B♭"
"Eb" when separated by space should be imported as "E♭"

This approach is used when adding new parts in MuseScore so keeps a consistent approach.

In a very related topic, MuseScore sets certain default fonts based on the element <word-font>.   While this is a good approach when a user goes out of their way in other software to change these defaults, the user more commonly does not set anything on purpose or changes the font with a more localized approach.  

The problem is that MuseScore is using this default information and importing scores using ugly defaults that highlight the origins of the file and not the beautiful default font that was created a couple years ago: Edwin.   Using Times New Roman as your default font is not indicative of someone who cares about a certain look to the score they are trying to retain in MuseScore.

Do not use these as default text fonts when presented and instead use Edwin:

Plantin MT Std - Previous Sibelius default
Palatino - Current Sibelius default
Times New Roman - Finale default
[...]

PR #17118, opened 2023-04-03
Implement #17037: Replacement of 'b' and '#' with '♭' and '♯' for staff names imported from mxml

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/17037

This replaces b and # with their unicode equivalents during musicxml import iff:

- It is a capital or lowercase letter A-G, OR a valid note name in Italian or French followed by one of those two characters
- That note name is its own token (i.e. surrounded by whitespace or at the end or beginning of the name)

As is pointed out in a comment on the original issue, there is an import option available to replace (or not replace) fonts. This PR does not override that setting, and doesn't touch font stuff at all.

### DIFF
--- a/src/engraving/libmscore/instrument.cpp
+++ b/src/engraving/libmscore/instrument.cpp
@@ -150,18 +150,6 @@ StaffName::StaffName(const String& xmlText, int pos)
     TextBase::validateText(_name); // enforce HTML encoding
 }
 
-void StaffName::xmlReplaceAccidentals()
-{
-    // instrument names in some musicXML use 'b' instead of '♭' for flat, so we need to convert them.
-    // i don't know if there is a valid reason for # to be in a staff name, but we convert it too just in case.
-
-    _name.replace(std::regex(R"(((?:^|\s)([A-Ga-g]|[Uu][Tt]|[Dd][Oo]|[Rr][EeÉé]|[MmSsTt][Ii]|[FfLl][Aa]|[Ss][Oo][Ll]))b(?=\s|$))"),
-                  String::fromStdString(R"($1♭)"));
-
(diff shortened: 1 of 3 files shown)


asattely:
Updated to deal with transpositions in non-english languages, such as Si♭

oktophonie (approved):

src/engraving/libmscore/instrument.cpp
```diff
@@ -150,6 +150,18 @@ StaffName::StaffName(const String& xmlText, int pos)
     TextBase::validateText(_name); // enforce HTML encoding
 }
 
+void StaffName::xmlReplaceAccidentals()
```
RomanPudashkin, line 153: This code should be inside importmxml.cpp (since it's a musicxml-related code)

asattely: So, it should still be part of `StaffName` but it should be defined in `importmxml.cpp`? Should I move the `Instrument` method below it as well?

RomanPudashkin: No, it should not be part of StaffName. You can declare it directly in importmxml.cpp as a static function and pass StaffName as an argument

src/engraving/libmscore/instrument.h
```diff
@@ -425,6 +425,7 @@ class Instrument
     void setIsPrimary(bool isPrimary);
 
     void updateInstrumentId();
+    void updateNamesForAccidentals();
```
RomanPudashkin, line 428: Please remove this line (it is not needed)

src/importexport/musicxml/internal/musicxml/importmxml.cpp
```diff
@@ -53,6 +53,34 @@ static int musicXMLImportErrorDialog(QString text, QString detailedText)
     return errorDialog.exec();
 }
 
+void updateNamesForAccidentals(Instrument* inst)
```
RomanPudashkin, line 56: **static** void updateNamesForAccidentals(Instrument* inst)

PR #17126, opened 2023-04-03
Fix #15537: Vertical formatting to avoid text overflow on score export dialog

### LINKED ISSUE #15537: [MU4 Issue] Text impossible to read on export audio files
**Describe the bug**
I was exporting a score in MP3 format but one of the options isn't possible to read, the text is too big for the space.

**To Reproduce**
 - I'm using musescore in portuguese
1. Go to "Publish"
2. Click on 'Exportar'
3. change the format to export to "MP3 audio"
4. See error on "Taxa de am..."

**Expected behavior**
It should be possible to read all the text in this field

**Screenshots**
image

**Platform information**
 - OS: Windowns
 - lang: portuguese

hodori0719:
Resolves: #15537

Changes the formatting for {label / dropdown menu} on the score export dialog from horizontal to vertical. Resolves issue of unreadable overflowing label text in some localizations of Musescore. 

Design as specified by @jessjwilliamson 

### DIFF
--- a/src/project/qml/MuseScore/Project/internal/Export/ExportOptionItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/Export/ExportOptionItem.qml
@@ -27,7 +27,6 @@ import MuseScore.UiComponents 1.0
 ColumnLayout {
     id: root
 
-    property int labelWidth: 72
     property alias text: label.text
 
     width: Math.min(implicitWidth, parent.width)
@@ -35,7 +34,7 @@ ColumnLayout {
 
     StyledTextLabel {
         id: label
-        Layout.preferredWidth: root.labelWidth
+        Layout.fillWidth: true
         horizontalAlignment: Text.AlignLeft
     }
 }
--- a/src/project/qml/MuseScore/Project/internal/Export/ExportOptionItem.qml
+++ b/src/project/qml/MuseScore/Project/internal/Export/ExportOptionItem.qml
@@ -24,14 +24,14 @@ import QtQuick.Layouts 1.15
 
 import MuseScore.UiComponents 1.0
 
-RowLayout {
+ColumnLayout {
     id: root
 
     property int labelWidth: 72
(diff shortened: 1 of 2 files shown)


src/project/qml/MuseScore/Project/internal/Export/ExportOptionItem.qml
```diff
@@ -24,14 +24,14 @@ import QtQuick.Layouts 1.15
...
 
     property int labelWidth: 72
     property alias text: label.text
 
     width: Math.min(implicitWidth, parent.width)
-    spacing: 12
+    spacing: 8
 
     StyledTextLabel {
         id: label
```
cbjeukendrup, line 37: On the line below this one\*, I think `Layout.preferredwidth: root.labelWidth` should be replaced with `Layout.fillWidth: true`, since now the label may fill the whole width of the dialog if necessary. After that, it seems like `property int labelWidth` can be removed altogether. 

\* I'm posting the comment here because GitHub doesn't let me comment on the line I mean because that line is too far away from the modified lines...

hodori0719: Good point, thank you!

cbjeukendrup (approved):

cbjeukendrup:
@jessjwilliamson Would be great if you could give this a quick test!

DmitryArefiev:
Tested # on Win10, Mac13 - **FIXED** All looks according to https://github.com/musescore/MuseScore/issues/15537#issuecomment-1469987429

jessjwilliamson:
Hi @hodori0719 ! Thank you so much for your work! Everything looks great, just one **small** fix needs to be done :) The Braille option has some empty space here, if it's possible to move the checkbox area up slightly, to look like the design I've showed attached then that would be great! Makes it look consistent with every other option looks too. Thank you! 🔥
image
image

PR #17131, opened 2023-04-04
Fix #17068 Add new function for visibility of staff without unhiding …

### LINKED ISSUE #17068: [Instruments panel] Staves cannot be unhidden without first unhiding the associated instrument
### Issue type

Other type of issue

### Bug description

@bkunda  My expectation for this case is that the user is able to make visible in parts one (or more if available) of instrument staves, but it could be discussed if you don't agree.

@abariska my proposed solution is in the [comment below](https://github.com/musescore/MuseScore/issues/17068#issuecomment-1493335028).

Could be linked with https://github.com/musescore/MuseScore/issues/17066 and https://github.com/musescore/MuseScore/issues/17067

### Steps to reproduce

1. Open any score with 2+ instruments
2. Open a part
3. Make visible and invisible back one hidden instrument (if https://github.com/musescore/MuseScore/issues/17066 not fixed yet)
4. Expand the instrument drop-down
5. Try to make visible specific staves

https://user-images.githubusercontent.com/101250347/228572815-43918737-2a37-49d1-85bc-4b975681d723.mp4

### MuseScore Version

MuseScore 4.0.2 

### Regression

No.

### Operating system

macOS 13

### Additional context

_No response_

vashutangri:
Resolves: #17068 

Added new function setIsVisibleFromStaffTreeItem to allow staves to be unhidden without unhiding the associated instrument first. 

### DIFF
--- a/src/instrumentsscene/view/abstractinstrumentspaneltreeitem.cpp
+++ b/src/instrumentsscene/view/abstractinstrumentspaneltreeitem.cpp
@@ -244,7 +244,7 @@ void AbstractInstrumentsPanelTreeItem::setTitle(QString title)
     emit titleChanged(m_title);
 }
 
-void AbstractInstrumentsPanelTreeItem::setIsVisible(bool isVisible)
+void AbstractInstrumentsPanelTreeItem::setIsVisible(bool isVisible, bool setChildren)
 {
     if (m_isVisible == isVisible) {
         return;
@@ -253,8 +253,10 @@ void AbstractInstrumentsPanelTreeItem::setIsVisible(bool isVisible)
     m_isVisible = isVisible;
     emit isVisibleChanged(isVisible);
 
-    for (auto child : m_children) {
-        child->setIsVisible(isVisible);
+    if (setChildren) {
+        for (auto child : m_children) {
+            child->setIsVisible(isVisible);
+        }
     }
 }
 
--- a/src/instrumentsscene/view/stafftreeitem.cpp
+++ b/src/instrumentsscene/view/stafftreeitem.cpp
@@ -33,6 +33,9 @@ StaffTreeItem::StaffTreeItem(IMasterNotationPtr masterNotation, INotationPtr not
         if (!m_isInited) {
             return;
         }
+        if (isVisible && !this->parentItem()->isVisible()) {
+            this->parentItem()->setIsVisible(true, false);
+        }
 
         this->notation()->parts()->setStaffVisible(id(), isVisible);
     });
--- a/src/instrumentsscene/view/abstractinstrumentspaneltreeitem.h
+++ b/src/instrumentsscene/view/abstractinstrumentspaneltreeitem.h
@@ -86,7 +86,7 @@ public:
 public slots:
     void setType(InstrumentsTreeItemType::ItemType type);
     void setTitle(QString title);
-    void setIsVisible(bool isVisible);
+    void setIsVisible(bool isVisible, bool setChildren = true);
     void setId(const ID& id);
     void setIsExpandable(bool expandable);
     void setIsEditable(bool editable);


src/instrumentsscene/view/abstractinstrumentspaneltreeitem.cpp
```diff
@@ -258,6 +258,12 @@ void AbstractInstrumentsPanelTreeItem::setIsVisible(bool isVisible)
     }
 }
 
+void AbstractInstrumentsPanelTreeItem::setIsVisibleFromStaffTreeItem(bool isVisible)
+{
+    m_isVisible = isVisible;
```
Jojo-Schmitz, line 263: As requested by a user on Discord, better add:
```
    if (m_isVisible == isVisible) {
        return;
    }
```
to avoid unneeded signaling

vashutangri: Added to the function

src/instrumentsscene/view/abstractinstrumentspaneltreeitem.cpp
```diff
@@ -258,6 +258,15 @@ void AbstractInstrumentsPanelTreeItem::setIsVisible(bool isVisible)
     }
 }
 
+void AbstractInstrumentsPanelTreeItem::setIsVisibleFromStaffTreeItem(bool isVisible)
+{
+    if(m_isVisible == isVisible) {
```
Jojo-Schmitz, line 263: There's a space missing ;-)

src/instrumentsscene/view/abstractinstrumentspaneltreeitem.cpp
```diff
@@ -258,6 +258,15 @@ void AbstractInstrumentsPanelTreeItem::setIsVisible(bool isVisible)
...
 
+void AbstractInstrumentsPanelTreeItem::setIsVisibleFromStaffTreeItem(bool isVisible)
+{
+    if (m_isVisible == isVisible) {
+        return;
+    }
+    m_isVisible = isVisible;
+    emit isVisibleChanged(isVisible);
+}
+
```
cbjeukendrup, line 269: This new method is now almost completely a copy of the normal `setIsVisible` method. It would be better to merge the methods into one, and add an extra `bool setChildren` parameter, that will determine whether children items should be made visible or invisible too. 

Because `setIsVisible` is used as a setter for a Q_PROPERTY, the new parameter will need to have a default value of `true`.

vashutangri: Okay, I'll do that. Just allow me a little time :D

src/instrumentsscene/view/stafftreeitem.cpp
```diff
@@ -29,11 +29,14 @@ using namespace mu::notation;
 StaffTreeItem::StaffTreeItem(IMasterNotationPtr masterNotation, INotationPtr notation, QObject* parent)
     : AbstractInstrumentsPanelTreeItem(InstrumentsTreeItemType::ItemType::STAFF, masterNotation, notation, parent)
 {
-    connect(this, &AbstractInstrumentsPanelTreeItem::isVisibleChanged, [this](bool isVisible) {
+    connect(this, &AbstractInstrumentsPanelTreeItem::isVisibleChanged, [this](bool isVisible)
+    {
```
cbjeukendrup, line 33: This change seems unnecessary

src/instrumentsscene/view/stafftreeitem.cpp
```diff
@@ -29,11 +29,14 @@ using namespace mu::notation;
...
         if (!m_isInited) {
             return;
         }
-
+        if (isVisible && !this->parentItem()->isVisible()) {
+            this->parentItem()->setIsVisible(true, false);
+        }
         this->notation()->parts()->setStaffVisible(id(), isVisible);
     });
```
cbjeukendrup, line 41: It would be great to add some newlines here for readability:
```suggestion

        if (isVisible && !this->parentItem()->isVisible()) {
            this->parentItem()->setIsVisible(true, false);
        }

        this->notation()->parts()->setStaffVisible(id(), isVisible);
    });
```

cbjeukendrup (commented):
It's almost perfect now! I have only two stylistic remarks. (Yes, it's a bit nitpicky, but good code style is important to keep the code nice to work with.)

src/instrumentsscene/view/stafftreeitem.cpp
vashutangri: Hey, thank you for the suggestion. Couldn't get back before due to some college work :( I've made the changes :D

cbjeukendrup (approved):

abariska:
🟢 **TESTED**

PR #17138, opened 2023-04-04
Add Liblouis v3.24.0 to src/braille/thirdparty

shoogle:
Liblouis is an open source braille text library licensed under LGPL-2.1-or-later. See https://github.com/liblouis/liblouis.

Liblouis converts written text to and from contracted or uncontracted braille in various languages. It does this based on braille tables in the 'tables' directory, which are plain text files that must be shipped with the compiled executable, similarly to MuseScore's translation files and instrument resources.

We have included some extra braille tables in `src/braille/tables`. These provide new capabilities to liblouis, such as the ability to convert musical text and certain musical symbols.

---

### Method to add/update Liblouis:

1. Visit https://github.com/liblouis/liblouis/releases.
2. Pick a release (we used [__v3.24.0__](https://github.com/liblouis/liblouis/releases/tag/v3.24.0)).
3. Download the [__Source code (zip)__](https://github.com/liblouis/liblouis/archive/refs/tags/v3.24.0.zip) generated by GitHub.
4. Extract into `src/braille/thirdparty/liblouis`.
5. Delete all subdirectories except the ones we actually need:
    - `src/braille/thirdparty/liblouis/liblouis`
    - `src/braille/thirdparty/liblouis/tables`.
6. Apply any patches needed to make it compile in our build environment.
    - Keep the patches as separate commits so we can see which third party code was changed.

---

Liblouis is needed for the Live Braille features implemented in PR #16699. We're adding the library first in a separate PR to make the Live Braille PR easier to review.


shoogle:
I temporarily enabled the module in order to check the build was successful on all platforms. Here's proof:



Workflow runs for:
- [Linux](https://github.com/musescore/MuseScore/actions/runs/4613081436/jobs/8154704574?pr=17138)
- [MacOS](https://github.com/musescore/MuseScore/actions/runs/4613081429/jobs/8154704127?pr=17138)
- [Windows](https://github.com/musescore/MuseScore/actions/runs/4613081428/jobs/8154704428?pr=17138)

I will now drop the last two commits, which will disable the module. (There is no reason to enable it while it is unused.)

CMakeLists.txt
```diff
@@ -185,6 +186,10 @@ add_subdirectory(share)
 add_subdirectory(src/framework/global) # should be first to work pch
 add_subdirectory(src)
 
+if (MUE_BUILD_LIVEBRAILLE_MODULE)
+    add_subdirectory(thirdparty/liblouis)
+endif(MUE_BUILD_LIVEBRAILLE_MODULE)
```
igorkorsukov, line 191: It's better to add this 3rd party library in the same way as others are added - they are added by the module that uses it.
This has advantages:
* We can see exactly who is using a third-party library
* When we disable the target module, its dependencies are also automatically disabled

shoogle: That makes sense, but the module is not going to be part of this PR. @RomanPudashkin requested that the library go in first to make PR #16699 easier to review.

I could create a CMakeLists.txt at the location where the module will go, but I suspect that PR #16699 has put the files in the wrong places.

Where is the right place for the module?

- src/livebraille
- src/framework/livebraille
- src/notation/livebraille
- Somewhere else?

FYI, here's what the module will eventually do:

1. If the user selects an element in the score, the element's parent measure will be displayed as Braille in the viewer panel (this is implemented PR [#16699](https://github.com/musescore/MuseScore/pull/16699)).
2. Within the Braille string, the characters corresponding to the selected score element will be highlighted.
3. If the user clicks on a different Braille character, the corresponding element will become selected in the score.
4. If the user types some Braille into the viewer using some special keyboard shortcuts to represent Braille dots, this will be turned into ordinary notation in the score.

Also, the new module's code will be used to export Braille files instead of the current Braille export code. (I guess there would still be an export module but internally it would just call methods from the new module do the conversion.)

shoogle: @igorkorsukov, if you're not sure where the module should go then it might be best to merge this as it is so that PR [#16699 ](https://github.com/musescore/MuseScore/pull/16699) can be rebased and reviewed. I can always move these lines into the module as part of that PR once somebody has reviewed it and determined the proper place to put the module.

igorkorsukov: @shoogle thanks for the clarifications  

I think this new module should be located in `src/livebraille`, like `palette`, `inspector`...
I would prefer this structure: 
```
src/livebraille/
                        internal/
                                      ...
                        thirdparty/ 
                                          .untidy
                                          liblouis/
                                                        ...
                        livebraillemodule.cpp
                        livebraillemodule.h
                        ...
```

I would prefer third-party libraries to also located into modules that require them. Previese to this, this was not done due to the fact that it was difficult to exclude them from code style checks. Now it's easy to do, and I plan to move them...

CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ set(VST3_SDK_PATH "" CACHE PATH "Path to VST3_SDK. SDK version >= ${VST3_SDK_VER
 # Modules (alphabetical order please)
 option(MUE_BUILD_APPSHELL_MODULE "Build appshell module" ON)
 option(MUE_BUILD_AUTOBOT_MODULE "Build autobot module" ON)
+option(MUE_BUILD_BRAILLE_MODULE "Build braille module (not yet implemented)" ON)
```
igorkorsukov, line 75: please add disabling this module for VTest configuration https://github.com/musescore/MuseScore/blob/master/SetupConfigure.cmake#L109

src/braille/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+set(MODULE braille) # NOT YET IMPLEMENTED
+
+add_subdirectory(tables)
```
igorkorsukov, line 3: Well done!

CMakeLists.txt
shoogle: I decided to name the module `braille` rather than `livebraille` because the code will also be used by the Braille Export module ([`iex_braille`](https://github.com/musescore/MuseScore/blob/4d8711c157170ad1d47696f9a39c6cd4d4401d33/src/importexport/braille/CMakeLists.txt#L21)).

CMakeLists.txt
shoogle: Done!

RomanPudashkin (approved):

shoogle:
@igorkorsukov, @RomanPudashkin, thanks for your help and the merge!

PR #17143, opened 2023-04-05
Disable video export code commandline options

Jojo-Schmitz:
if and when disabled for the build

Came up in https://musescore.org/en/node/347329#comment-1184603 ff.

### DIFF
--- a/src/app/commandlineparser.cpp
+++ b/src/app/commandlineparser.cpp
@@ -96,6 +96,7 @@ void CommandLineParser::init()
     m_parser.addOption(QCommandLineOption({ "S", "style" }, "Load style file", "style"));
 
     // Video export
+#ifdef MUE_BUILD_VIDEOEXPORT_MODULE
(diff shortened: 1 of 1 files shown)


src/app/commandlineparser.cpp
```diff
@@ -108,6 +109,7 @@ void CommandLineParser::init()
     m_parser.addOption(QCommandLineOption("fps", "Frame per second [60, 30, 24]", "24"));
     m_parser.addOption(QCommandLineOption("ls", "Pause before playback in seconds (3.0)", "3.0"));
     m_parser.addOption(QCommandLineOption("ts", "Pause before end of video in seconds (3.0)", "3.0"));
+#endif
```
igorkorsukov, line 112: I believe that it makes no sense to remove all the code (for what? to speed up the compilation of these few lines?) I agree that it is worth removing only the addition of command options

Jojo-Schmitz: OK, I'm fine with that too.

igorkorsukov (approved):

Issue #17162, opened 2023-04-07
[WEB] Some issues with parts in PDF on musescore.com


DmitryArefiev:
### Steps to reproduce

**BUG 1**: Some parts in PDF are empty when downloaded from musescore.com, if they were not opened as tabs in MS while publishing 

1. Create a score (or open [publish2.zip](https://github.com/musescore/MuseScore/files/11179346/publish2.zip) )
2. Open some parts but not all
3. Publish
4. Go to musescore.com>your score>Download>PDF
5. Select all parts and download - some PDF of parts are empty

**BUG 2**: Custom parts are not available in musescore.com>your score>Download>PDF
1. Open [Check PDF publish (with custom parts).zip](https://github.com/musescore/MuseScore/files/11179371/Check.PDF.publish.with.custom.parts.zip)
2. Publish
4. Go to musescore.com>your score>Download>PDF - **the custom parts are not present**

### Screenshots/Screen recordings

**BUG 1**

https://user-images.githubusercontent.com/90187801/230620684-3f136d46-5f3f-4113-8f50-b94e1ec07494.mp4

https://user-images.githubusercontent.com/90187801/230620835-2f6383b2-7b33-4a4f-aa9d-a5d6a6b509ef.mp4

--

**BUG2**
![Capture](https://user-images.githubusercontent.com/90187801/230621198-da2873da-dcf4-45a7-ba7d-825265066aa1.PNG)

### MuseScore Version

4.1

### Regression

No.

### Operating system

Windows10

bkunda:
@Eism why was this reopened?

DmitryArefiev:
@bkunda I've checked that in latest master
**BUG 1:** Some parts in PDF are empty when downloaded from musescore.com, if they were not opened as tabs in MS while publishing **is FIXED**

But **BUG 2:** Custom parts are not available in musescore.com>your score>Download>PDF **STILL OCCURS**

![Capture](https://github.com/musescore/MuseScore/assets/90187801/6fe57e97-79e9-42d2-917c-0388d2c82ec3)

PR #17166, opened 2023-04-07
Backend pdfs fix

Eism:
Resolves: #17162

The problem with empty pdfs appeared after these changes 799e3eff72282ef67fc892461dc312d7dde398ef

### DIFF
--- a/src/converter/internal/compat/backendapi.cpp
+++ b/src/converter/internal/compat/backendapi.cpp
@@ -217,15 +217,18 @@ RetVal<project::INotationProjectPtr> BackendApi::openProject(const io::path_t& p
         return make_ret(Ret::Code::InternalError);
     }
 
-    INotationPtr notation = notationProject->masterNotation()->notation();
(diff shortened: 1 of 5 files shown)


src/converter/internal/compat/backendapi.cpp
```diff
@@ -738,6 +748,34 @@ Ret BackendApi::applyTranspose(const INotationPtr notation, const std::string& o
...
+        score->setLayoutAll();
+        score->update();
+    }
+}
+
+void BackendApi::initPotentialExcerpts(notation::IMasterNotationPtr masterNotation)
+{
+    ExcerptNotationList potentialExcerpts = masterNotation->potentialExcerpts();
+
+    masterNotation->addExcerpts(potentialExcerpts);
```
RomanPudashkin, line 775: You can use MasterNotation::initExcerpt instead of adding these excerpts to the main score

RomanPudashkin (approved):

PR #17179, opened 2023-04-09
Handle more element types in TRead::readItem

shoogle:
Without these additions, the latest `master` reaches the UNREACHABLE statement when loading Palettes and fails to launch for me on Windows.


igorkorsukov:
Thank you very much, I just wanted to do this now, it's good that I looked here

Jojo-Schmitz:
Strange, it starts for me even without that PR

igorkorsukov:
I think it depends on the showed palettes

PR #17188, opened 2023-04-10
[engraving] Added accidental, beam, chord... write. Step 10

igorkorsukov:


src/engraving/layout/layoutchords.cpp
```diff
@@ -1499,7 +1499,7 @@ void LayoutChords::resolveRestVSRest(std::vector<Rest*>& rests, Score* score, Se
     int lines = staff->lines(tick);
     const double minRestToRestClearance = 0.35 * spatium;
 
-    for (int i = 0; i < rests.size() - 1; ++i) {
+    for (size_t i = 0; i < rests.size() - 1; ++i) {
```
RomanPudashkin, line 1502: We should check if this list is empty

RomanPudashkin (approved):

src/engraving/layout/layoutchords.cpp
igorkorsukov: fixed

Issue #16466, opened 2023-02-21
[MU4 Task] Aux Sends implementation


jessjwilliamson:
Updated and new UX on Aux Sends

<img width="1304" alt="image" src="https://user-images.githubusercontent.com/71656189/220331645-cdd5db41-df30-4efa-9af0-66a481134ec7.png">

Link to Figma page (Aux sends): https://www.figma.com/file/tn0dH2CQRbh53oYJI4Hquv/Mixer?node-id=6%3A35

Tantacrul:
Thanks Jess! For anyone interested, we'll have a new reverb effect packaged with MS4. This will massively boost performance for both Muse Sounds and MS Basic. It'll also let people specify the reverb amount they want per instrument.

And of course, all the other benefits of aux channels.

PR #17189, opened 2023-04-10
Built-in reverb processor

vpereverzev:
Part of https://github.com/musescore/MuseScore/issues/16466


cbjeukendrup:
The Muse Reverb has no GUI (yet). (And it is not a VST, but something built-in.)

PR #17206, opened 2023-04-11
aux_channels

RomanPudashkin:
Part of https://github.com/musescore/MuseScore/issues/16466

Also see: https://github.com/musescore/MuseScore/pull/17189

### DIFF
--- a/src/framework/audio/audiotypes.h
+++ b/src/framework/audio/audiotypes.h
@@ -45,6 +45,7 @@ using volume_db_t = float;
 using volume_dbfs_t = float;
 using gain_t = float;
 using balance_t = float;
+using audio_signal_amount_t = float;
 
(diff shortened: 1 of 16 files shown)


src/framework/audio/audiotypes.h
```diff
@@ -45,6 +45,7 @@ using volume_db_t = float;
 using volume_dbfs_t = float;
 using gain_t = float;
 using balance_t = float;
+using audio_signal_amount_t = float;
```
vpereverzev, line 48: use gain_t instead, technically it's really what we're using there

vpereverzev (approved):

src/framework/audio/audiotypes.h
RomanPudashkin: I'll fix this in the next PR

PR #17231, opened 2023-04-13
Slur tie collision avoidance

mike-spa:
Resolves: #15805 

Wasn't a bug but a missing feature. I had done another PR before this which also included collision avoidance with tuplets, but we realized it was too soon for that. Here I just included slur VS tie.


oktophonie (approved):

RomanPudashkin:
@mike-spa please rebase it

Issue #17241, opened 2023-04-14
MuseScore 4 crashes when opening scores from OpenScore Lieder Corpus


apacha:
### Issue type

Crash or freeze

### Bug description

When trying to open a few songs from the OpenScore lieder corpus in MS4, the application reliably crashes, see https://github.com/OpenScore/Lieder/issues/4 for a full list.

### Steps to reproduce

Try to open the file like this https://github.com/OpenScore/Lieder/files/11219372/09_Der_Spinnerin_Nachtlied.zip with MuseScore 4 and the application crashes.

### MuseScore Version

MuseScore version (64-bit): 4.0.2-230651546, revision: github-musescore-musescore-dbe7c6d

### Regression

Yes, this used to work in Musescore 3.x and now is broken

### Operating system

macOS 13.3, Arch.: x86_64

DmitryArefiev:
Fixed in #17242

PR #17242, opened 2023-04-14
Fix crash when opening a certain 206 score (with nested frames?)

cbjeukendrup:
`b->system()` is of course not yet initialized while reading. In MS3.6.2, this was no problem because elements could have nullptr as their parent, but in MS4 the Factory methods expect a non-null parent.

Resolves: #17241


igorkorsukov (approved):

DmitryArefiev:
Tested #17241 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #15760, opened 2023-01-08
[MU4 Issue] Guitar Tablature and Change Instrument crashes Musescore 4


ericglam:
**Describe the bug**
Guitar Tablature and Change Instrument crashes Musescore 4. 

**To Reproduce**
1. Create a new score from template: solo guitar, normal staff + tablature
2. Enter some notes so you have more than one measure
3. Click on a note or rest in the second measure
4.  Change instrument from Guitar to Harp ( or any instrument that is not a guitar)
5. Musescore 4 dies a quick death

**Expected behavior**
I Expect Musescore not to crash. Even though I have a linked tablature staff, it should stay just like it was even though the instrument is not a guitar. It works that way in Musescore 3- OS: Windows 10 (10.0), Arch.: x86_64, MuseScore version (64-bit): 3.6.2.548021803, revision: github-musescore-musescore-3224f34.  

 (I use the Harp when I want the Guitar to play back with harmonics as a workaround. Harmonics really could be a channel like Palm Mute as most acoustical guitarists use harmonics). Yes, I know channels are not completely implemented yet in Musescore 4 which was why I was trying the change Instrument to get this effect in playback.  (Channels are not very intuitive and seem to be confusing to most Musescore users - a robust Change Instrument could be much more intuitive and is more in keeping with the feel of Musescore and easier to implement)

**Platform information**
OS: Windows 10 Version 2009, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-223472200, revision: github-musescore-musescore-5485621

**Additional context**
It seems to be the tablature staff causing this issue.  Another way to cause this crash is to  get rid of the tablature staff and then just link a standard staff .  Then add an instrument change from Guitar to Harp on the second measure.  So far so good. All you have to do now is change the linked staff to a linked tablature staff and the score will freeze for a second and Musescore 4 will crash.

DmitryArefiev:
Fixed in #17245

PR #17245, opened 2023-04-14
Prevent crash when number of strings for instrument is decreased

cbjeukendrup:
when changing to an instrument with less strings, or when deleting strings in Staff/Part properties > Edit string data

(also some very basic code quality improvements)

Resolves: #15760


RomanPudashkin (approved):

DmitryArefiev:
Tested #15760 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #17247, opened 2023-04-14
Can't run MuseScore CLI in headless mode


j-tai:
### Issue type

Crash or freeze

### Bug description

In MuseScore 3, we could run the MuseScore command-line interface without a display server (for example, on a headless server) by passing the option `-platform offscreen`. For example:
```console
$ mscore -platform offscreen -o output.mid input.mscz
```
This no longer works in MuseScore 4, with or without that option. It fails to start in both cases:
```console
$ mscore -v
QEventLoop: Cannot be used without QApplication
QEventLoop: Cannot be used without QApplication
qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, xcb.

$ mscore -platform offscreen -v
QEventLoop: Cannot be used without QApplication
QEventLoop: Cannot be used without QApplication

(mscore:2): Gtk-WARNING **: 11:47:02.947: cannot open display:
```
Expected behavior: the above two commands should print the MuseScore version (because of `-v` flag).

### Steps to reproduce

On a Linux machine:

```console
$ unset DISPLAY  # remove the DISPLAY env var so MuseScore can't connect to a display server
$ mscore3 -platform offscreen -v
```

### MuseScore Version

4.0.2 (Flatpak)

### Regression

Yes, this used to work in Musescore 3.x and now is broken

### Operating system

Ubuntu 22.04.2 LTS

Jojo-Schmitz:
The Flatpak version is not supported here, so does the same issue exist with the AppImages?

j-tai:
> The Flatpak version is not supported here, so does the same issue exist with the AppImages?

Ah got it. The same issue occurs with the AppImage downloaded from musescore.org, with a slightly different error message:

```console
$ ./MuseScore-4.0.2.230651545-x86_64.AppImage -platform offscreen -v
/usr/lib/x86_64-linux-gnu/pipewire-0.3/jack/libjack.so.0
/lib/x86_64-linux-gnu/libnss3.so
QEventLoop: Cannot be used without QApplication
QEventLoop: Cannot be used without QApplication
[...]

PR #17253, opened 2023-04-16
Try fix running with `offscreen` platform

cbjeukendrup:
Resolves: #17247


Eism (approved):

Eism:
Oh, in the issue I see that the solution does not work, so my approval can be ignored

cbjeukendrup:
The solution itself does work (see https://github.com/musescore/MuseScore/issues/17247#issuecomment-1510437936), but in the latest nightly builds, there is another separate problem with command line arguments. Therefore it does not _seem_ to work if you use the command line option rather than the environment variable. I could fix the command line issue separately or add the fix to this PR.

PR #17256, opened 2023-04-17
Fix #16376: Automatically add ties between intervals on top of tied notes

### LINKED ISSUE #16376: [MU4 Issue] Adding Intervals on top of tied notes should apply a tie
https://user-images.githubusercontent.com/78870770/219055409-5316de6f-8dcb-4099-880a-5d04f5d4c9a5.mp4

hodori0719:
Resolves: #16376 

Adding an interval (3rd) now automatically applies ties to the new notes if the highest selected notes are tied.


src/engraving/libmscore/cmd.cpp
```diff
@@ -655,9 +655,12 @@ void Score::expandVoice()
 
 void Score::addInterval(int val, const std::vector<Note*>& nl)
 {
+    std::vector<Note*> notes;
```
cbjeukendrup, line 658: Even though the existing code is not really doing a great job at this, it would be good to give the variables slightly clearer names. For example, I would rename this one to `newNotes`, and you could rename `nl` to `notes`. Generally, in this method, it would be nice to consistently call the existing notes `oldNote` and the newly created ones `newNote`.

src/engraving/libmscore/cmd.cpp
```diff
@@ -728,6 +731,23 @@ void Score::addInterval(int val, const std::vector<Note*>& nl)
 
         select(note, SelectType::SINGLE, 0);
     }
+    unsigned i = 0;
+    for (Note* on : nl) {
+        if (on->tieFor()) {
+            auto it = find(nl.begin(), nl.end(), on->tieFor()->endNote());
+            if (it != nl.end()) {
+                unsigned index = it - nl.begin();
```
cbjeukendrup, line 739: Note that it is not necessary to use this `index` variable here. If you have an iterator, you can just dereference that iterator:
```c++
    Note* oldEndNote = *it;
```
But in this case, that is not even necessary either, because you already have a pointer to that note: it's just `on->tieFor()->endNote()`.

src/engraving/libmscore/cmd.cpp
```diff
@@ -728,6 +731,23 @@ void Score::addInterval(int val, const std::vector<Note*>& nl)
...
+                Tie* tie = Factory::createTie(this->dummy());
+                tie->setStartNote(notes[i]);
+                tie->setEndNote(notes[index]);
+                tie->setTrack(notes[i]->track());
+                tie->setTick(notes[i]->chord()->segment()->tick());
+                tie->setTicks(notes[index]->chord()->segment()->tick() - notes[i]->chord()->segment()->tick());
```
cbjeukendrup, line 745: I believe these lines can be done slightly simpler:
```suggestion
                tie->setTick(notes[i]->tick());
                tie->setTick2(notes[index]->tick());
```

src/engraving/libmscore/cmd.cpp
```diff
@@ -728,6 +731,23 @@ void Score::addInterval(int val, const std::vector<Note*>& nl)
 
         select(note, SelectType::SINGLE, 0);
     }
+    unsigned i = 0;
+    for (Note* on : nl) {
+        if (on->tieFor()) {
+            auto it = find(nl.begin(), nl.end(), on->tieFor()->endNote());
+            if (it != nl.end()) {
```
cbjeukendrup, line 738: Actually, the only thing you want to know here is whether the end note of the tie was included in the list of selected notes. We have a small convenience method for that in `containers.h`:
```suggestion
            if (mu::contains(nl, on->tieFor()->endNote()) {
```

src/engraving/libmscore/cmd.cpp
hodori0719: Thank you for the feedback, I made the other changes you suggested.
For this one, I'm using the indices of the tied notes in the original list to find the corresponding notes in `newNotes` so I can tie them together, which is why I am using find rather than contains.

I considered getting rid of the vector entirely and using `on->chord()->upNote()` and `on->tieFor()->endNote()->chord()->upNote()` instead, of course changed to `downNote` if the interval is negative, but this does not work if notes in the middle of a chord are selected individually. Please correct me if I'm misunderstanding your feedback.

cbjeukendrup: Ah, you're right. You need `index` to retrieve the correct note from the other list. Looks good now!

RomanPudashkin (approved):

RomanPudashkin:
@hodori0719 please rebase this PR. Thanks!

DmitryArefiev:
@hodori0719 After rebase please assign it to me for testing. Thanks!

hodori0719:
> @hodori0719 After rebase please assign it to me for testing. Thanks!

Sorry for the delay, if you could test this now. Thank you!

cbjeukendrup:
@hodori0719 Sorry, but due to very recent changes there are conflicts _again_... so another rebase is needed!

DmitryArefiev:
@hodori0719 Hi! I tried **master build** and it already works like in #16376 expected result. Or I just missing some details..


https://github.com/musescore/MuseScore/assets/90187801/a5f23679-152c-4b22-ad6b-3b1a990dc0de

cbjeukendrup:
@DmitryArefiev @hodori0719 It looks like this feature got already implemented by commit 01d93a2eb98fdfadb783f5b360cbc9d6b5b0574a, in the meantime.

**However**, when you make a range selection containing multiple tied chords, the results are very strange in the **current master branch** and **current state of this PR**:
Scherm­afbeelding 2023-07-04 om 21 45 18

But in the **original version** of this PR, i.e. commit 01d93a2eb98fdfadb783f5b360cbc9d6b5b0574a, it worked fine: 
Scherm­afbeelding 2023-07-04 om 22 08 25

@hodori0719 It would be great if you could look into this! Perhaps you need to partially revert the changes from 01d93a2eb98fdfadb783f5b360cbc9d6b5b0574a, and replace them by your original changes.

hodori0719:
Looks like this was implemented in another PR and the bug was eventually fixed as far as I can tell. Closed

Issue #17099, opened 2023-03-31
Non-Gradual Tempo Modifiers in MusicXML imports


smorell-1:
### Your idea

Currently, when importing MusicXML, tempo modifiers like "a tempo", "Tempo 1" and "Più mosso" import as staff text attached to the top staff in the score. I would like to improve logic to more elegantly address importing these elements as tempo text with BPM defaults that make sense based on the existing tempos these types of instructions correspond with.

### Problem to be solved

After importing MusicXML that includes text like "a tempo" or "tempo 1", users must delete the staff text created in the import, add new tempo text and then change the text and BPM information in that tempo text to reflect the actual desired behavior.  While users are always invited to tweak this information as they see fit, MuseScore should provide a better default from a MusicXML import to streamline this workflow.  Along with other proposed updates, we can import MusicXML and play it back as expected even in situations like this one.

<img width="538" alt="Screen Shot 2023-03-31 at 12 15 39 PM" src="https://user-images.githubusercontent.com/128835319/229209840-2206c1ca-71e1-40f5-853c-92777e684ee9.png">

Attached here is a flowchart for details on how to best apply an appropriate tempo text and corresponding BPM based on the text found.

[Logic Path for Non-Gradual Tempo Modifiers.pdf](https://github.com/musescore/MuseScore/files/11125249/Logic.Path.for.Non-Gradual.Tempo.Modifiers.pdf)

Here are all the text variations we can use to trigger these tempo text which should not be case sensitive.  % changes are outlined on appropriate items.

### Steady Tempo Instructions Based on Existing Tempos (Non-Gradual)

**A tempo:** return to the base tempo after an adjustment
Listed as: "a tempo"

**Tempo 1:** Go back to the tempo at the very beginning of the piece.
Listed as: "tempo 1", "tempo i", "tempo 1o", "tempo io", "tempo 1°", "tempo i°", "primo tempo" "Au mouvement" or "tempo primo"

**Mosso:** movement; more lively, but not so much as più mosso
Listed as: "mosso". Behavior: +5% of current tempo

**Meno Mosso:** less motion, a little slower
Listed as: "meno mosso" Behavior: -10% of current tempo

**Piu mosso:** more motion; a little faster
Listed as: "piu mosso" or "più mosso" Behavior: +10% of current tempo

**Ritenuto:** immeditately (not gradually) slightly slower
Listed as: "riten.", "riten" or "ritenuto" Behavior: -10% of current tempo

[...]

Issue #17081, opened 2023-03-30
Gradually Accelerating Tempo Modifiers in MusicXML imports


smorell-1:
### Your idea

Currently, when importing MusicXML elements like "accel." or "stringendo" import as staff text attached to the top staff in the score. I would like to connect new accel. tool launched in MuseScore 4 (available in the tempo palette) to more elegantly address importing these elements.

### Problem to be solved

After importing MusicXML using "accel." or "stringendo", users must delete the staff text created in the import and add a accel. line to replace this behavior. The user must then identify the best location to drag the end of the line to decide where a logical end to the accel would be. This update seeks to streamline this behavior with smart defaults that leverage this great new tool (as seen here).

<img width="665" alt="Screen Shot 2023-03-30 at 1 51 48 PM" src="https://user-images.githubusercontent.com/128835319/228961642-59b66429-4a37-4d9d-a843-e2766a256a33.png">

Attached here is a flowchart for how/when to apply to an accel. line along with logic on where to set the end of the line, what % increase should be included, and when to display the line.

[Logic Path for Gradually Accelerating Tempo Modifiers.pdf](https://github.com/musescore/MuseScore/files/11116078/Logic.Path.for.Gradually.Accelerating.Tempo.Modifiers.pdf)

It is important to note that other updates to the MusicXML import process will need to be made for this to logic to be fully realized. This includes converting other text like "Allegro" to tempo text with a default BPM prior to applying this logic (requested in https://github.com/musescore/MuseScore/issues/17052), updates to gradually slowing tempo modifiers (requested in https://github.com/musescore/MuseScore/issues/17052) and properly creating jump elements for DC/DS form elements from MusicXML imports which will be handled in the future.

### Gradual Accelerating Terms

Here is the list of triggering text used in the gradual acceleration logic.

**Accelerando:** Gradually increase the tempo
Listed as: "accel.", "accel" or "accelerando"

**Stringendo:** gradually speeding up
Listed as: "stringendo"

Please see the list of Gradually Slowing Terms and Steady Tempo Instructions Based on Existing Tempos (Non-Gradual) in https://github.com/musescore/MuseScore/issues/17052 for text that triggers and end point as defined in the attached flowchart in this issue.

### Additional context

Here is a MusicXML file that will test a variety of situations.

[Test Accelerating Tempo Modifiers.zip](https://github.com/musescore/MuseScore/files/11116131/Test.Accelerating.Tempo.Modifiers.zip)

[...]

Issue #17069, opened 2023-03-29
Gradually Slowing Tempo Modifiers in MusicXML imports


smorell-1:
### Your idea

Currently, when importing MusicXML elements like "rit." or "rall." import as staff text attached to the top staff in the score.  I would like to connect new rit. tool launched in MuseScore 4 (available in the tempo palette) to more elegantly address importing these elements.

### Problem to be solved

After importing MusicXML using "rit." or "rall.", users must delete the staff text created in the import and add a rit. line to replace this behavior.  The user must then identify the best location to drag the end of the line to decide where a logical end to the rit would be.  This update seeks to streamline this behavior with smart defaults that leverage this great new tool (as seen here).

<img width="680" alt="Screen Shot 2023-03-29 at 1 43 10 PM" src="https://user-images.githubusercontent.com/128835319/228662648-6e215fef-7a7b-460b-b135-119f0f6e17e1.png">

Attached here is a flowchart for how/when to apply to a rit. line along with logic on where to set the end of the line, when to display the line, etc.

[Logic Path for Gradually Slowing Tempo Modifiers.pdf](https://github.com/musescore/MuseScore/files/11105328/Logic.Path.for.Gradually.Slowing.Tempo.Modifiers.pdf)

It is important to note that other updates to the MusicXML import process will need to be made for this to logic to be fully realized.  This includes converting other text like "Allegro" to tempo text with a default BPM prior to applying this logic (requested in [this issue](https://github.com/musescore/MuseScore/issues/17052)) and properly creating jump elements for DC/DS form elements from MusicXML imports which will be handled in the future.

This flowchart also references several types of tempo modifiers as triggering the end of the line to indicate the duration of time we are applying the slow down (all terms are handled the same way, but the text will display as whatever the text reads of course).  Please note we are covering how to treat Gradually Slowing Terms only, but accelerating and steady terms are referenced in the end point logic for slowing term, so they are listed here as triggering text for where to place the end of a line.  The treatment of accelerating and steady tempo elements will be handled in later issues.

There are also a variety of ways these terms are expressed and considerations for how they are performed.  The terms "molto", "poco" and "poco a poco" are often mixed with these terms as well and should be allowed in any variation and like all other terms is not case sensitive.  "Accel." "Molto accel" "poco A Poco Rit." or "accel." for example should all be picked up as triggering text to be clear noting that it is not case sensitive.

[...]

Issue #17052, opened 2023-03-28
Update to Tempo Markings from Palette and in MusicXML Imports


smorell-1:
### Your idea

Currently, the ready-made tempo texts in MuseScore are fixed to quarter notes being the beat.  Ideally these tempos would account for time signatures and update their tempos accordingly.  As an expansion of this idea, it would be great to better map MusicXML imports to change staff text to appropriate tempo texts while properly accounting for the time signatures.

### Problem to be solved

Currently, tempo text added from the pallettes do not account for time signature.  We could more effectively add these tempo texts if MuseScore accounted for time signatures.  For example, if you add a tempo marking in 4/4 as seen here

<img width="847" alt="Screen Shot 2023-03-28 at 1 17 41 PM" src="https://user-images.githubusercontent.com/128835319/228357396-bb531510-5950-4548-b830-79cedf688db2.png">

The tempo is added at 200BPM which is actually fixed at quarter note=200.  When the same tempo is added to a cut time measure, it stays at quarter note=200.  Because the beat is interpreted as half notes, this will not sound like Prestissimo requiring an update from the user.

<img width="885" alt="Screen Shot 2023-03-28 at 1 18 07 PM" src="https://user-images.githubusercontent.com/128835319/228358189-5db71602-7b5c-4cc2-8b35-ce87cd6c7558.png">

I propose two updates:

1.  Keeping the same list of 13 Tempos currently offered in MuseScore in the Tempo pallette but automating adjustment of their speed depending on the time signature it is added.
     a.  Changes will be applied to any /2 time sig, cut time, 3/8, 6/8, 9/8 and 12/8 according to the attached spreadsheet.

2.  Updating MusicXML imports to look for commonly used tempo texts and automatically creating tempo texts (versus staff text) set at accurate tempos that also account for time signature.

Attached is a spreadsheet detailing exactly the tempos that should be set.  I have not changed any of the current default BPM's in MuseScore to be clear just expanded on how they should be used in different time signatures.

[Tempos.xlsx](https://github.com/musescore/MuseScore/files/11093772/Tempos.xlsx)

PR #17257, opened 2023-04-17
[MXML Import] Interpret common tempo markings as musescore-native tempo objects

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/16797
Resolves: https://github.com/musescore/MuseScore/issues/17052
Resolves: https://github.com/musescore/MuseScore/issues/17069
Resolves: https://github.com/musescore/MuseScore/issues/17081
Resolves: https://github.com/musescore/MuseScore/issues/17099

- Recognized tempo text is added as a Tempo Text object rather than Staff Text
- Recognized gradual tempo change text is added as GradualTempoChange
- Gradual tempo changes' length is guessed based on context clues (i.e. proceeding tempo markings)
- The tempo lines' factor can stretch, if appropriate (so Largo -> accel. -> Presto will produce a smooth gradient between the two tempos)
- Relative tempos also work, with piu mosso for example adding 10% of the current tempo.
- Interpret dotted line with gradual tempo change text as an explicit indication of the length of the line, such as this combo generated by the Dolet plugin for Sibelius: [sib rit with line.zip](https://github.com/musescore/MuseScore/files/11254068/sib.rit.with.line.zip)

This PR also fixes an issue that was sorta created in #7171, an example of which can be found in the comments in the PR after it was merged. Now, instead of only adding tempo if no tempo has been added to that tick, it now only refuses to add if there is a tempo on that tick _and_ the text is identical.

### DIFF
--- a/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
+++ b/src/importexport/musicxml/internal/musicxml/importmxmlpass2.cpp
@@ -1342,7 +1342,7 @@ static void handleSpannerStop(SLine* cur_sp, track_idx_t track2, const Fraction&
 //---------------------------------------------------------
 
 MusicXMLParserPass2::MusicXMLParserPass2(Score* score, MusicXMLParserPass1& pass1, MxmlLogger* logger)
-    : _divs(0), _score(score), _pass1(pass1), _logger(logger)
+    : _divs(0), _score(score), _pass1(pass1), _logger(logger), _tempoLineManager(score)
 {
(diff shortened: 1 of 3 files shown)


oktophonie (approved):

bkunda:
@RomanPudashkin, could you review this at your convenience? We need to assess if it requires major changes, minor adjustments, or if it adequately addresses the issues it aims to resolve. Feel free to delegate the review if needed. Thanks!

Issue #16771, opened 2023-03-13
Vertical offset of beams affects notehead position


bkunda:
### Issue type

Engraving bug

### Bug description

Changing the vertical offset of beams in **Properties** > **Appearance** also affects the vertical offset of attached stems and noteheads (but, somewhat amusingly, not ledger lines).

### Steps to reproduce

1. Select a beam
2. Go to **Properties** > **Appearance**
3. Change vertical offset value
4. Observe how stems and noteheads attached to selected beam move together with the beam itself

### Screenshots/Screen recordings

https://user-images.githubusercontent.com/86290556/224749805-bf053acf-03d0-4020-b994-a34666368891.mov

### Expected behaviour

The control for vertical offset should be disabled when a beam is selected.

### MuseScore Version

MuseScore version (64-bit): 4.0.2-230651546, revision: github-musescore-musescore-dbe7c6d

### Regression

I don't know

### Operating system

macOS 13.1

### Additional context

Perhaps also relevant to https://github.com/musescore/MuseScore/issues/16663

oktophonie:
Question is, what _should_ the offset do?
In MU3 it had no effect, so probably it shouldn't in MU4 either; indeed, the offset should probably be disabled.
To actually change the vertical position of the beam, there are separate controls for each endpoint.

bkunda:
> Question is, what _should_ the offset do? In MU3 it had no effect, so probably it shouldn't in MU4 either; indeed, the offset should probably be disabled. To actually change the vertical position of the beam, there are separate controls for each endpoint.

I did actually have recourse to want to move the entire beam uniformly (i.e. maintaining its slope by not moving the separate endpoints), which is how I came across this issue. 

I wonder if it could work to have the vertical offset affect only the selected grip handle? This would allow the user to control the vertical offset for each grip handle separately (which they can already do, just not via the properties panel).

oktophonie:
The left/right handle positions can be set in Properties here (hidden beneath 'Show more'):
![image](https://user-images.githubusercontent.com/20604032/225601387-862049c4-606c-44bd-976d-a0ebf6d1806c.png)

bkunda:
OK! That answers my question then (the padlock achieves exactly what I wanted to achieve, which is the uniform adjustment of the beam height).

So then I am in agreement with you @oktophonie, in that the vertical offset option should be disabled in the **Appearance** popup. 

 I'll add this as "Expected behaviour" to this issue description.

PR #17263, opened 2023-04-17
Don't show vertical offset property for beams

mike-spa:
Resolves: #16771

### DIFF
--- a/src/inspector/view/qml/MuseScore/Inspector/general/appearance/AppearanceSettings.qml
+++ b/src/inspector/view/qml/MuseScore/Inspector/general/appearance/AppearanceSettings.qml
@@ -70,6 +70,7 @@ Column {
 
         offset: model ? model.offset : null
         isSnappedToGrid: model ? model.isSnappedToGrid : null
+        isVerticalOffsetAvailable: model ? model.isVerticalOffsetAvailable : false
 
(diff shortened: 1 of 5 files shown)


src/inspector/view/qml/MuseScore/Inspector/general/appearance/AppearanceSettings.qml
```diff
@@ -70,6 +70,7 @@ Column {
 
         offset: model ? model.offset : null
         isSnappedToGrid: model ? model.isSnappedToGrid : null
+        isVerticalOffsetAvailable: model ? model.isVerticalOffsetAvailable : null
```
cbjeukendrup, line 73: It would be nicer to use `false` instead of `null` here, since this is about booleans rather than objects:
```suggestion
        isVerticalOffsetAvailable: model ? model.isVerticalOffsetAvailable : false
```
(functionally it won't make any difference because JavaScript doesn't care very much about types, but semantically it is cleaner)

cbjeukendrup:
I noticed that if you select a beam together with some other object, the offset of the beam can still be changed: 

https://user-images.githubusercontent.com/48658420/232544670-9d4ccc3f-e978-47cf-88db-c79f58669338.mov

RomanPudashkin (approved):

PR #17265, opened 2023-04-17
Reset HitElementContext when deleting element

### LINKED ISSUE #13590: [MU4 Issue] Adding "Title" (etc.) text from menu doesn't create a title frame if there are no frames in the score
Edit @Tantacrul:

I've created a video to demonstrate
https://www.dropbox.com/s/7rkgsg2plgy7e80/Can%27t_Add_Title_to_Score_01.mov?dl=0

---

**Describe the bug**

If there are NO frames in the current score, when you attempt to add "Title" text (etc.) from the Add menu, NO frame is created.

**Expected behavior**

As in MS3. A title frame is created at the beginning of the score.

**Platform information**

 OS: Linux Mint 20.1, Arch.: x86_64, MuseScore version (64-bit): 4.0.0-3171545449, revision: github-musescore-musescore-3d92b93

cbjeukendrup:
In particular, this fixes #13590, where the problem is that the just deleted frame is still stored in the HitElementContext. Therefore, any newly created text object will be added to that deleted frame instead of to a newly created frame in the score.

Resolves: #13590

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -3697,6 +3697,7 @@ void NotationInteraction::deleteSelection()
     }
 
     apply();
+    resetHitElementContext();
 }
 
 void NotationInteraction::flipSelection()


RomanPudashkin (approved):

DmitryArefiev:
Tested #13590 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

Issue #17229, opened 2023-04-13
musescore 4 rolls grace note chords


janmusija:
### Issue type

General playback bug

### Bug description

Instead of articulating the notes in grace note chords simultaneously, MS4 articulates them as if they were rolled, or perhaps as if indicated to be played one by one from lowest on the staff to highest on the staff. This does not depend on whether you use the basic soundfonts or muse sounds.

### Steps to reproduce

1. Create a score. Instrument choice doesn't seem to matter but this issue definitely occurs on piano 
2. Write a note anywhere in the score
3. Add a grace note to this note. Doesn't matter what kind.
4. Make this grace note into a chord by adding intervals on top. Doesn't matter how many, though the effect is more obvious with more notes:
5. Hit "play"; it should sound rolled
6. Export the audio; it will again sound rolled.

### Screenshots/Screen recordings

<img width="260" alt="image" src="https://user-images.githubusercontent.com/110359879/231617381-b8441597-897b-4557-80f7-89a91524beb0.png">

(audio export from this image):

https://user-images.githubusercontent.com/110359879/231618685-0bb43eb8-758e-424d-8393-688c6a374ad5.mp4

### MuseScore Version

MS 4.0.2-230651546

### Regression

Yes, this used to work in Musescore 3.x and now is broken

### Operating system

macOS 12.5 (x84)

### Additional context

Works fine in my installation of MS3, which is version 3.6.2.548020600.

DmitryArefiev:
Fixed in #17266

PR #17266, opened 2023-04-17
Render grace chords as chords rather than rolled

cbjeukendrup:
Resolves: #17229

Also: 
Make sure that the played duration of grace notes is proportional to the notated duration, when a chord has multiple grace notes of different lengths. I.e. consider the following example:
Scherm­afbeelding 2023-04-17 om 20 23 34
Before, both measures would sound the same.
After, in the first measure you hear the appropriate rhythm.

The reason that I did that, is that in the code there were some unused methods relating to grace note duration, which could mean two things: either it was decided that all grace notes should play the same length and someone forgot to delete those methods, or there was no such decision but someone broke it accidentally. Since the designers are not aware of such decision, I assume the latter is the case.


RomanPudashkin (approved):

DmitryArefiev:
Tested #17229 on Win10, Mac13, LinuxUbuntu 22.04 - **FIXED**

Issue #17103, opened 2023-04-01
Doesn't seem to load soundfonts from symbolic links


itaranto:
### Issue type

Other type of issue

### Bug description

On Linux, If I place a symbolic link in `~/Documents/Musescore4/SoundFonts/` it doesn't loads in MS4 while it used to work this way in MS3.

This is handy since I don't have to copy a whole soundfont file in order to use it for MS4.

### Steps to reproduce

1. Create a symbolic link in `~/Documents/Musescore4/SoundFonts/` pointing to an SF2/SF3 file.
2. Open MS4, go to the Mixer and look for recognized soundfonts.

**Expected:**
The new soundfont appears in the Mixer's menu.

**Actual:**
The new soundfont doesn't appear in the Mixer's menu.

### MuseScore Version

4.0.2

### Regression

Yes, this used to work in Musescore 3.x and now is broken

### Operating system

Arch Linux

shoogle:
We should fix this, but as a workaround you could try making the entire `MuseScore4/SoundFonts` directory a symlink to the equivalent MuseScore 3 directory, so that you can use all the same sounds (except SFZs).

cbjeukendrup:
> as a workaround you could try making the entire `MuseScore4/SoundFonts` directory a symlink to the equivalent MuseScore 3 directory, so that you can use all the same sounds (except SFZs).

But then you could as well directly select the `MuseScore3/SoundFonts` directory in Preferences > General > Folders, right?

itaranto:
> > as a workaround you could try making the entire `MuseScore4/SoundFonts` directory a symlink to the equivalent MuseScore 3 directory, so that you can use all the same sounds (except SFZs).
> 
> But then you could as well directly select the `MuseScore3/SoundFonts` directory in Preferences > General > Folders, right?

Nope, I've just tried and it's the same, the app doesn't seem to be following symbolic links.

itaranto:
Also, linking the entire `MuseScore4/SoundFonts` to `MuseScore3/SoundFonts` doesn't work either.

MarcSabatella:
I think the suggestion of changing Preferences was to do that *instead* of using a symbolic link - just tell MuseScore 4 to looking in Documents/MuseScore3/SounfFonts directly, no links involved.  But, a downside of changing the preferences is not surviving revert to factory settings, which I end up doing more often than the average user probably for testing purposes.  So I prefer the symbolic link approach instead personally.

[...]

PR #17267, opened 2023-04-17
Add support for soundfonts as symbolic links

flxkrmr:
Resolves: #17103

Should fix the issue, but I could only test it on Windows. Any input is welcome :) 

I did not yet write a unit test for this, as I think it would not be much use in this code, but I can try to write one if you want to.


Jojo-Schmitz:
Exactly my point in the issue. As said there in MU3 the only place where symlinks were not permitted was for extension (MDL and the HQ soundfont, but not soundfonts in general)

shoogle:
@cbjeukendrup, yes, I don't think we need to worry about that situation. Like you said, it's no different to renaming the original file.

I think the issues specific to symlinks are:

1. **Unresolvable link:** Target file does not exist, or we lack permissions for it.
2. **Infinite loop:** Target file is another symlink pointing (possible via more symlinks) back to the original symlink.
3. **Infinite search:** A subfolder of the target directory is a link (possible via more symlinks) back to a parent directory of the original symlink.

But I don't think it's our job to worry about these cases. If the user has created a symlink then we should assume they know what they are doing.

cbjeukendrup:
... plus that I think that Qt's scan functions have some protection against these problems. 

So indeed, I think it's safe to allow symlinks everywhere in principle, and only disable it in specific places if it ever turns out to cause problems.

flxkrmr:
Thanks for the input :) I changed it to work globally with symbolic links and tested it with Soundfonts, Templates and Scores so far, which seem to work fine.

igorkorsukov (approved):

Issue #17246, opened 2023-04-14
export to PDF crashes (Linux only)


gucio321:
### Issue type

Crash or freeze

### Bug description

When trying to export to PDF, everything crashes

### Steps to reproduce

- open any file
- click file -> Export
- click Export
- after a few seconds everything crashes with "core dumped"

### Screenshots/Screen recordings

<details><summary>logs</summary>
<p>

```
QML debugging is enabled. Only use this in a safe environment.
QSocketNotifier: Can only be used with threads started with QThread
20:25:34.639 | INFO  | main_thread | GlobalModule | onPreInit: log path: /home/mszeptuch/.local/share/MuseScore/MuseScore4Development/logs/MuseScore_230414_202534.log
20:25:34.639 | INFO  | main_thread | GlobalModule | onPreInit: === Started MuseScore 4.1.0-dev, build number  ===
20:25:34.660 | DEBUG | main_thread | DiagnosticsModule | onInit: crash server url: 
20:25:34.665 | INFO  | main_thread | DiagnosticsModule | onInit: success start crash handler
20:25:34.681 | DEBUG | main_thread | LinuxAudioDriver | open: Connected to default
20:25:34.681 | ERROR | main_thread | MuseSamplerLibHandler | MuseSamplerLibHandler: Unable to open MuseSampler library, path: /home/mszeptuch/.local/share/MuseSampler/lib/libMuseSamplerCoreLib.so
20:25:34.681 | ERROR | main_thread | MuseSamplerResolver | checkLibrary: Incompatible MuseSampler library; ignoring
20:25:34.681 | ERROR | main_thread | MuseSamplerLibHandler | MuseSamplerLibHandler: Unable to open MuseSampler library, path: libMuseSamplerCoreLib.so
20:25:34.681 | ERROR | main_thread | MuseSamplerResolver | checkLibrary: Incompatible MuseSampler library; ignoring
20:25:34.683 | DEBUG | main_thread | ShortcutsRegister | expandStandardKeys: removed 5 shortcut, because they are not bound to standard key
20:25:34.683 | DEBUG | main_thread | ShortcutsRegister | expandStandardKeys: added 3 shortcut, because they are alternative shortcuts for the given standard keys
20:25:34.685 | DEBUG | main_thread | ShortcutsRegister | expandStandardKeys: removed 176 shortcut, because they are not bound to standard key
20:25:35.074 | WARN  | main_thread | IpcSocket  | connect: failed connect to server
[...]

Issue #17224, opened 2023-04-12
Crash when deleting a custom part (one score)


DmitryArefiev:
### Steps to reproduce

1. Open [Smoke test 7 del.zip](https://github.com/musescore/MuseScore/files/11213156/Smoke.test.7.del.zip)
2. Open Parts
3. Delete **Contrabasess (copy)** part

https://user-images.githubusercontent.com/90187801/231507704-665773f2-260c-49f3-a148-8b660ff82c3e.mp4

### MuseScore Version

4.1 (MuseScoreNightly-231020304-master-3837973-x86_64)

### Regression

Yes, this used to work in a previous version of Musescore 4.x

### Operating system

Windows10

### Additional context

Recent changes. Can be related to #17220

**Update: These changes are causing the crashes**: https://github.com/musescore/MuseScore/pull/16866

DmitryArefiev:
I think the problem related to dialog appearing. Because the same crash happens when creating a new workspace:

https://user-images.githubusercontent.com/90187801/231517385-e9df3bdc-44af-4147-83e7-de0082ad000a.mp4

DmitryArefiev:
Fixed in #17268

Issue #17220, opened 2023-04-12
Crash after Revert to factory settings


DmitryArefiev:
### Issue type

Crash or freeze

### Steps to reproduce

1. Open MS
2. Revert to factory settings
3. Click Next or X in Welcome dialog

https://user-images.githubusercontent.com/90187801/231456269-7f48fcba-1b68-47ed-a69c-3aa988a0c3e4.mp4

### MuseScore Version

4.1 (MuseScoreNightly-231020304-master-3837973-x86_64)

### Regression

Yes, this used to work in a previous version of Musescore 4.x

### Operating system

Windows10/Mac13

### Additional context

The latest master build

DmitryArefiev:
Fixed in #17268

PR #17268, opened 2023-04-17
Fix crash when current URI changes while DialogView is open

cbjeukendrup:
We should not attempt to close dialogs when the current URI changes or when the user clicks outside; only popups.

Resolves: #17246
Resolves: https://github.com/musescore/MuseScore/issues/17224
Resolves: #17251


RomanPudashkin (approved):

DmitryArefiev:
Tested #17246 #17224 #17220 #17251 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #17269, opened 2023-04-18
Crescendo lines overlap + Chord text frame

Jojo-Schmitz:
* Adjust bounding box of chord symbols if they have frames. Avoids collisions between and with them.
* Draw hairpins as polygons, except when crossing system breaks or on non-solid lines. Also using `BevelJoin` rather than `MiterJoin`.

See discussions and images in https://github.com/Jojo-Schmitz/MuseScore/issues/98, and the PRs https://github.com/Jojo-Schmitz/MuseScore/pull/99 and https://github.com/Jojo-Schmitz/MuseScore/pull/100, all for 3.x, but same/similar issues (and fixes) apply to master.

Hairpins look like this, before the change (and even worse in 3.x):
image
Reason being that they are drawn as individual lines rather than as a polygon. Looks even worse at larger angles.
And like this after ther change (using `RoundJoin`, which looks the same at any angle):
image
And like this using `BevelJoin`:
image
And like this when dashed (before and after the changes of this PR:
image

Or, in motion (using `RoundJoin`):

https://user-images.githubusercontent.com/7139517/232326404-324b162b-8ba1-4189-b94d-0585411133c1.mp4

And in motion using `MiterJoin` (looks different depending on angle):

https://user-images.githubusercontent.com/7139517/232324389-dc63e322-f8bb-4fc2-a2b1-50cdbdc4f0d0.mp4

Might be something to make available as a preference or style setting.

@oktophonie , what do you think?


Jojo-Schmitz:
The vtest failures unsurprisingly only affect the tips of hairpins and are entirlely expected.

Jojo-Schmitz:
@oktophonie , what do you think?

cbjeukendrup:
I believe the correct way is to draw them as sheared rectangles, so that the ends are always vertical. At least that's what Dorico does, and it looks quite good to me. The two disadvantages of that approach are: 1. it is maybe complicated 2. if you don't apply some correction, the lines become visually thinner when the angle is extremely steep, but if you do apply correction, then the ends will become thicker when the angle is steep.

Jojo-Schmitz:
Hmm, that might be achiend woth using `BevelJoin`, see https://doc.qt.io/qt-6/qt.html#PenJoinStyle-enum

cbjeukendrup:
Yeah, bevel join would work for the tip where the lines intersect, but not at the other side.

Jojo-Schmitz:
I can't fix the open ends (would require an entirely different method, drawing filled rectangles rather than lines, as far as I can tell, and would probably apply to other lines too, all that can a) be diagonal and b) don't start/end with a hook), but I can haripins change to use `BevelJoin`

Jojo-Schmitz:
What about this one?

Jojo-Schmitz:
it needs a vtest label ;-)

src/engraving/libmscore/textlinebase.cpp
```diff
@@ -149,9 +149,14 @@ void TextLineBaseSegment::draw(mu::draw::Painter* painter) const
             pen.setDashPattern({ dash, gap });
         }
 
+        // or RoundJoin, or leave at the default of MiterJoin? Make it a preference/style setting?
```
cbjeukendrup, line 152: I think we don't need this comment

cbjeukendrup (commented):
It would be good to add some VTests for this:
- At least one where we can see the effect of the chord symbol frame bounding box fix
- And maybe one with a very (unrealistically) thick hairpin (or maybe just a very big spatium works too) so that we can see the effect on hairpins clearly.

Jojo-Schmitz:
@worldwideweary can you come up with a VTest for the chord symbol frame bounding box fix?

src/engraving/libmscore/textlinebase.cpp
Jojo-Schmitz: Of course, it is currently there to trigger discission about what to use and whether to make it a preference setting.
I'll remove and keep as `BevelJoin`
I'll try to create a VTest for this too.

Jojo-Schmitz: done

Jojo-Schmitz:
VTests added

cbjeukendrup (approved):

cbjeukendrup:
@oktophonie Could you please test this? In my opinion, it's definitely an improvement of the situation, so we can merge it. But note that the hairpins are still not perfect yet: at the "open" end of the hairpin, the line ends are still not vertical (but perpendicular to the direction of the line). Fixing this would require a different, more complicated technique, that we can implement at some point in the future. 
And for dashed hairpins, the fix of this PR cannot be applied, because they need to be drawn as separate line pieces for other reasons (namely to make sure that the dashes of the upper and lower line are aligned vertically). So those remain unchanged for now. 
So, the scope of this improvement is limited, but it's definitely an improvement.

Jojo-Schmitz:
BTW: thanks for insisting on a vtest, it did reveal a problem on the corresponding PR for 3.x ;-)

oktophonie (approved):

Issue #17276, opened 2023-04-19
qt5-declarative 5.15.9+kde+r23-1 Bug MuseScore


ivanmelillo:
### Issue type

Other type of issue

### Bug description

With the new update of qt5-declarative 5.15.9+kde+r23-1 MuseScore cannot display the score editor. I'm an Arch user and had to downgrade to qt5-declarative 5.15.9+kde+r22-1

### Steps to reproduce

Update system pacman -Syu, y qt5-declarative 5.15.9+kde+r23-1 and open musescore

### Screenshots/Screen recordings

![Screenshot_20230418_215948](https://user-images.githubusercontent.com/67931709/232938029-fd715b99-e32b-4762-a93b-54b76da05314.png)

### MuseScore Version

4.0.2

### Regression

I don't know

### Operating system

ArchLinux

### Additional context

The bug is due to the qt5-declarative 5.15.9+kde+r23-1 update. With Downgrade MuseScore works correctly

Jojo-Schmitz:
As far as I know the AppImage used Qt 5.15.2, the last public available package.
I don't think we're supporting any distribution build. Or is this happening with the AppImage?

Jojo-Schmitz:
Then report this to the maintainers. Here we're only responsible for the AppImage

ivanmelillo:
In appimage MuseScore works correctly. See the test I did!

https://user-images.githubusercontent.com/67931709/233144524-abb43ea7-75a8-4f30-8b9e-79af1ca2589f.mp4

PR #17285, opened 2023-04-19
Set parentItem for pages in DockWindow

easyteacher:
Resolves:  #17276

After https://invent.kde.org/qt/qt/qtdeclarative/-/merge_requests/40 a QQuickItem without a parentItem is always invisible. While this avoids some crashes in plasmashell, it breaks some applications like MuseScore where pages don't have a parent item.

This is also important for Qt6 porting in the future.

### DIFF
--- a/src/appshell/view/dockwindow/dockwindow.cpp
+++ b/src/appshell/view/dockwindow/dockwindow.cpp
@@ -619,6 +619,7 @@ void DockWindow::initDocks(DockPageView* page)
     }
 
     if (page) {
+        page->setParentItem(this);
         page->init();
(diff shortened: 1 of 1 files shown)


Gobbel2000:
Thanks. I tested it and it does indeed solve the linked issue.

igorkorsukov (approved):

cbjeukendrup:
Oh! This also fixes the problems that I encountered after updating to Qt 6.5 on [my experimental Qt 6 branch](https://github.com/musescore/MuseScore/pull/10108)! Thanks a lot!

PR #17300, opened 2023-04-20
[MusicXML] Export turn with slash

rettinghaus:
This updates MusicXML export to correctly export turn `ornamentTurnSlash`.


oktophonie:
Can this be extended to also handle import? At the moment we import the slashed turn as a reversed turn (Finale and Sibelius too) which isn't the worst thing in the world, but it would be nice if we could import it accurately as well.

oktophonie (approved):

cbjeukendrup:
I wonder whether we should maybe add some tests for this (see `musicxml_tests.cpp`)

Issue #15108, opened 2022-12-11
[MU4 Issue] Instant crash when pressing Ctrl+A on an old musescore 2 file (with workaround).


bambancico:
**To Reproduce**
1. Load an old file [Musescore website](https://musescore.com/emperoni/gran-jota-de-concierto), [my copy](https://github.com/musescore/MuseScore/files/10202299/Files.zip) (I reproduced the error with original file again to ensure it is not corrupt).
2. Press Ok when shown the "it may be shown differently" prompt
3. Press Ctrl+A
4. Crash

**Expected behavior**
It outright crashes, but it doesn't when you select the whole score with the mouse.

**Workaround**
When saved in Mu4 AND reopened it no longer crashes.

**Platform information**
 - Windows 11 (22H2), AMD 5600u

**Additional context**
I cannot troubleshooot further since I don't have access to Mu2. [This](https://musescore.com/emperoni/gran-jota-de-concierto) has been, so far, the only file that's given me trouble. I am attaching the [original and the "resaved" files](https://github.com/musescore/MuseScore/files/10202299/Files.zip).

oktophonie:
Can confirm this crash still occurs in 4.0.1 (with https://musescore.com/emperoni/gran-jota-de-concierto)

Jojo-Schmitz:
3.6.2 crashes on thos too.

DmitryArefiev:
Fixed in #17302

PR #17302, opened 2023-04-20
Fix crash when Ctrl+A in 206 score with Glissandos without end element

cbjeukendrup:
Resolves: #15108

The crash was because apparently we expect Glissandos to always have an end element. In the 206 score where this happened, there were apparently Glissandos that did not have an end element. Those are removed; that is, they _were_ already removed for spanners starting on a normal note, but now I'll also remove them when starting on a _grace_ note.

As a next step, we could try to reconnect those Glissandos by trying to guess their end element. But I haven't done that yet in this PR.

(Maybe relevant info: if you'd open the crashing score in MS2.3.2, then those Glissandos that would be deleted by MS4 would not be visible in MS2.3.2 either.)

### DIFF
--- a/src/engraving/libmscore/score.cpp
+++ b/src/engraving/libmscore/score.cpp
@@ -5708,6 +5708,18 @@ void Score::connectTies(bool silent)
                     }
                 }
             }
+            for (Chord* gc : c->graceNotes()) {
+                for (Note* n : gc->notes()) {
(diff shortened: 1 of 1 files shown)


RomanPudashkin (approved):

DmitryArefiev:
Tested #15108 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #17308, opened 2023-04-21
Multiple fixes to MIDI output

Arakmar:
First, this PR is fixing the MIDI output due to a wrong velocity value (incorrect conversion to MIDI 2.0)
Then, it tries to mute the metronome for the MIDI output. (not very elegant for now)

It's still a work in progress and my ultimate goal is to be able to handle per channel mute state for the midi output.


RomanPudashkin:
This is actually not an elegant solution, unfortunately. However, I have some ideas on how to improve MIDI output. But thanks for your attempt and pointing out the problems!

Issue #14434, opened 2022-11-09
[MU4 Issue] Save to cloud: no "unlisted" visibility option


MarcSabatella:
**Describe the bug**
There is no unlisted option for "publish to musescore.com", and existing unlisted scores get downgraded to private on update.

**To Reproduce**
1. Create a score
2. File / Publish to MuseScore.com
3. See error: no unlisted option.  And when updating an existing unlisted score, it gets converted to private, breaking secret links.

**Expected behavior**
Unlisted should be an option for online scores

**Platform information**
 - OS: Linux

**Additional context**
When upload is complete, you have the opportunity to change to unlisted.  But, especially for larger scores that use Muse Sounds and might take the better part of an hour to upload, you might not be in a position to make the change when the upload completes.  Thus you can be left with an inaccessible score.  Also, the license gets changed in the process, from whatever you've set it to (e.g., "Attribution") to "None".

Tantacrul:
We did this intentionally to keep the saving options simpler in MS4. However, we may undo this decision if we find it is causing any issues.

Worth mentioning that we're in the middle of making alterations to Private scores so that a user can invite people to look at them and comment on them.

Tantacrul:
I'm going to close this as a won't fix for now. If we find there are issues users are raising, it'll wind back up here anyway. For now, it's not actionable.

Jojo-Schmitz:
I need my scores being unlisted.

Daniel-R15:
Those of us who create practice files absolutely do need the "Unlisted" option when publishing for a closed choral group.  "Private" doesn't allow members to see the scores, and "Public" breaks copyright.

Please reconsider and realise that "Unlisted" is essential to our workflow.

Jojo-Schmitz:
Came up yesterday in https://musescore.org/en/node/344336

bkunda:
> And when updating an existing unlisted score, it gets converted to private, breaking secret links.

Just to clarify, we did actually address this issue late last year and I can confirm (having just tested it) that the unlisted status of a score is preserved online even when saving changes in the desktop app. 

Certainly I can appreciate the hassle of not being able to set the status as unlisted from within the desktop app in the first place, but I just wanted to be clear that you can still work on an unlisted score and maintain its visibility status as such (unless someone has found a bug with this that I haven't been able to reproduce!).

Jojo-Schmitz:
No, there's no such but, to my knowledge

murimk:
[...]

PR #17315, opened 2023-04-22
Add unlisted to cloud saving options

flxkrmr:
Resolves: #14434

Adds the option to save project as unlisted in the cloud.

I left the "force audio saving" logic untouched, which I think means that it is the same as for private projects. If this needs to be different, let me know

### DIFF
--- a/src/project/internal/saveprojectscenario.h
+++ b/src/project/internal/saveprojectscenario.h
@@ -80,8 +80,9 @@ class QMLCloudVisibility
 
 public:
     enum CloudVisibility {
-        Private = int(cloud::Visibility::Private),
-        Public = int(cloud::Visibility::Public)
+        Public = int(cloud::Visibility::Public),
+        Unlisted = int(cloud::Visibility::Unlisted),
+        Private = int(cloud::Visibility::Private)
     };
     Q_ENUM(CloudVisibility);
 };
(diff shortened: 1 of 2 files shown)


Jojo-Schmitz:
Not really sure, but I'd rather have "Unlisted" between "Private" and "Public" than after "Public"
(and that merge commit doesn't belong theer, to rebase instead use `git pull --rebase upstream master; git push -f`

shoogle (commented):
Ideally the options should be specified in this order to match the website: Public, Unlisted, Private.

MueScore.com Score Manager | MuseScore.com Edit Score page
:---:|:---:
image | image

You can change this in the `CloudVisibility` enum and in the dialog itself.

The default visibility should be __Private__ for *File > Save to cloud* and __Public__ for *File > Publish to MuseScore.com*.

flxkrmr:
The order does make sense, I changed it. I also rebased everything to one commit because I think we don't really need a reorder commit :) However, I just realized in my german gui, unlisted is the only option that is not translated. Should I add a german translation right away and if so, can you point me in a direction how this works?

shoogle:
@flxkrmr, you don't need to provide a translation. You just need to mark the source string for translation (which you have done by putting it in the `qsTrc()` function), which will expose it to translators on [Transifex](https://app.transifex.com/musescore/musescore/dashboard/) after this is merged.

However, I have requested the options go in the opposite order to what @Jojo-Schmitz suggested, so please swap them round and double check that the defaults are correct. Sorry for the trouble!

flxkrmr:
Thanks for the info.

No problem, I hope the order is correct now :) The default values should also still be correct.

shoogle (approved):
Thanks for making those changes! I've tested and it's working great!

Issue #17000, opened 2023-03-24
"Swap with clipboard" command does not work and breaks range selection


MarcSabatella:
### Issue type

UI bug

### Bug description

The swap with clipboard command has broken badly between 4.0.1 and 4.0.2.  It correctly pastes the current clipboard contents, but rather than load the clipboard with the contents fo the destination, it appear to corrupt the clipboard in a strange way, where after attempting to paste it, nothing appears to happen, but you then become unable to make range selections until you save/reload.

### Steps to reproduce

1. new score, Treble Clef template, 4/4
2. enter quarter notes C C C C | D D D D into the first two measures
3. Select the first measure
4. Ctrl+C to copy
5. Select the second measure
6. Ctrl+Shift+X to swap with clipboard
7. Select the third (empty) measure
8. Ctrl+V to paste

Result: nothing is pasted, which is bad enough, but worse, you are now completely unable to make a range selection of any kind.  Clicking a measure, Click/Shift+click, Shift+Right, Shift+drag, etc - range selections just cannot be made anymore.  Save/reload fixes the inability to select, but the swap operation itself will still have failed.

### Screenshots/Screen recordings

[Screen recording 2023-03-24 8.54.24 AM.webm](https://user-images.githubusercontent.com/1799936/227560857-025fafd7-eaae-4a32-98e0-a089b654d827.webm)

### MuseScore Version

OS: Debian GNU/Linux 11 (bullseye), Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230651545, revision: github-musescore-musescore-dbe7c6d

### Regression

Yes, this used to work in a previous version of Musescore 4.x

### Operating system

Debian 11

### Additional context

This worked in 4.0.1, but is broken in 4.0.2.  It remains broken in a recent build from master.

DmitryArefiev:
Fixed in #17332

PR #17332, opened 2023-04-23
Fix swap with clipboard

cbjeukendrup:
Resolves: #17000

The last commit fixes the root cause of why "Swap with clipboard" itself didn't work, but the other two commits fix the other problem that became apparent because of this bug, namely the inability to create range selections after pasting a zero-length range.


RomanPudashkin (approved):

DmitryArefiev:
Tested #17000 on Win10, Mac12, LinuxUbuntu 22.04 - **FIXED**

PR #17343, opened 2023-04-24
Change pitchToString to show flated notes too

mikekirin:
Changed pitchToString, that it shows not only sharp notes, but flat notes as well.

### DIFF
--- a/src/framework/global/utils.cpp
+++ b/src/framework/global/utils.cpp
@@ -25,7 +25,7 @@
 
 using namespace mu;
 
-static constexpr const char* noteNames[] = {
+static constexpr const char* sharpNotes[] = {
     QT_TRANSLATE_NOOP("global", "C"),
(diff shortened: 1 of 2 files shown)


alexpavlov96 (approved):

handrok (approved):

Jojo-Schmitz:
All nice and good, but so far this isn't called anywhere in a way that would show flats, or am I missing something?

mikekirin:
> All nice and good, but so far this isn't called anywhere in a way that would show flats, or am I missing something?

This is a mu part of fix for xtz.

Jojo-Schmitz:
Ah, OK. So a new mobile app is in the works?

Jojo-Schmitz:
It probably translation-wise benefits from #16172, in that it uses the same strings in the same context.

Issue #16802, opened 2023-03-14
[FEATURE] Audio.com upload integration


bkunda:
This task is to establish a flow in MuseScore for uploading audio directly to audio.com.

This issue will be replaced with a design specification soon.

Tantacrul:
First pass design is done:
https://www.figma.com/file/IZ7TTcjlzgsN1mLbMfSkkL/MS4%3A-Upload-to-Audio.com?node-id=14%3A5817&t=kiBvvKAL6KTvNRGv-1

Please contact me in order to get a full explanation before proceeding with development.

PR #17346, opened 2023-04-24
Audio.com upload integration

Eism:
Resolves: #16802

image


src/cloud/internal/cloudsregister.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef MU_CLOUD_CLOUDSREGISTER_H
+#define MU_CLOUD_CLOUDSREGISTER_H
+
+#include <unordered_map>
+
+#include "../icloudsregister.h"
+
+namespace mu::cloud {
+class CloudsRegister : public ICloudsRegister
```
RomanPudashkin, line 31: This register looks strange. It uses the term "cloud" but provides IAuthorizationServicePtr

Tantacrul:
There's one other silly bug that's just a little bit worse now that there are two sign ins

If you sign in to either option and then hit the **sign out** option...
image

It takes you out of the **Accounts** section and moves you into the 'Scores' section instead. It should just stay in the Accounts section.

Tantacrul:
I've found another thing that needs correcting:

In the image below: we should change this copy to: 

**Login or create a new account on audio.com to share your music.**

image

---

The active button should say **Share** not **Publish**

image

src/project/internal/projectuiactions.cpp
```diff
@@ -86,6 +86,13 @@ const UiActionList ProjectUiActions::m_actions = {
              TranslatableString("action", "Publish to MuseScore.com…"),
              IconCode::Code::CLOUD_FILE
              ),
+    UiAction("file-publish-audio",
+             mu::context::UiCtxAny,
+             mu::context::CTX_ANY,
+             TranslatableString("action", "Publish to A&udio.com…"),
```
cbjeukendrup, line 92: We must be careful about `&u` in this string. We also use it in `"Save to clo&ud…"`, so the `"Publish to A&udio.com…"` item cannot be in the same menu as the `"Save to clo&ud…"` item. Currently it looks like it isn't, so there is no problem, but we should just keep it in mind.

DmitryArefiev:
Tested #16802 on Win10, Mac13, LinuxUbuntu 22.04 - **PASS**

I only didn't ability to test 2 dialogs for now: **Outdated version of MuseScore** and **Audio exceeds file size limit** 

Found some related bugs #17543 #17540

src/cloud/qml/MuseScore/Cloud/AccountInfoButton.qml
```diff
@@ -30,46 +30,32 @@ import "internal"
...
-    property string userName: accountModel.accountInfo.userName
-    property string avatarUrl: accountModel.accountInfo.avatarUrl
+    property var cloudInfo: Boolean(cloudsModel.userAuthorized) ? cloudsModel.firstAuthorizedCloudInfo() : null
 
     signal userAuthorizedChanged()
 
     orientation: Qt.Horizontal
 
-    spacing: Boolean(avatarUrl) ? 18 : 30
+    spacing: 30 - 32/4 // todo
```
RomanPudashkin, line 39: TODO?

src/cloud/qml/MuseScore/Cloud/internal/CloudItem.qml
```diff
@@ -0,0 +1,163 @@
...
+    property var userAvatarUrl: null
+    property var userCollectionUrl: null
+
+    signal signInRequested()
+    signal signOutRequested()
+    signal createAccountRequested()
+
+    property NavigationControl navigation: NavigationControl {
+        accessible.role: MUAccessible.ListItem
+        //        accessible.name: root.name
```
RomanPudashkin, line 46: Should be removed/fixed

src/cloud/audiocom/audiocomcloudservice.cpp
```diff
@@ -0,0 +1,353 @@
...
+cloud::CloudInfo AudioComCloudService::cloudInfo() const
+{
+    return {
+        AUDIO_COM_CLOUD_CODE,
+        AUDIOCOM_CLOUD_TITLE,
+        AUDIOCOM_CLOUD_URL
+    };
+}
+
+cloud::AbstractCloudService::ServerConfig AudioComCloudService::serverConfig()
```
RomanPudashkin, line 73: I would suggest not using the term "config" here. It is not a configuration, but rather "information. Also, the method should be const

src/cloud/audiocom/audiocomcloudservice.cpp
```diff
@@ -0,0 +1,353 @@
...
+{
+    TRACEFUNC;
+
+    QUrl url = QUrl(audioInfo.value("url").toString());
+    QUrl success = QUrl(audioInfo.value("success").toString());
+    QUrl fail = QUrl(audioInfo.value("fail").toString());
+    QUrl progress = QUrl(audioInfo.value("progress").toString());
+
+    if (!url.isValid() || !success.isValid() || !fail.isValid() || !progress.isValid()) {
+        return make_ret(Err::NetworkError);
```
RomanPudashkin, line 243: Ideally, we should use a more specific error code here

src/cloud/internal/abstractcloudservice.h
```diff
@@ -67,15 +67,51 @@ class CloudService : public QObject, public IAuthorizationService, public ICloud
...
-                                       const QUrl& sourceUrl = QUrl()) override;
-    framework::ProgressPtr uploadAudio(QIODevice& audioData, const QString& audioFormat, const QUrl& sourceUrl) override;
-
-    RetVal<ScoreInfo> downloadScoreInfo(const QUrl& sourceUrl) override;
-
 private slots:
     void onUserAuthorized();
 
+protected:
+    struct ServerConfig {
```
RomanPudashkin, line 74: Same issue: https://github.com/musescore/MuseScore/pull/17346#discussion_r1194749156

It's not a configuration

src/cloud/CMakeLists.txt
```diff
@@ -28,18 +28,28 @@ set(MODULE_SRC
...
     ${CMAKE_CURRENT_LIST_DIR}/internal/oauthhttpserverreplyhandler.cpp
+
+    ${CMAKE_CURRENT_LIST_DIR}/internal/abstractcloudservice.h
+    ${CMAKE_CURRENT_LIST_DIR}/internal/abstractcloudservice.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/musescorecom/imusescorecomcloudservice.h
+    ${CMAKE_CURRENT_LIST_DIR}/musescorecom/musescorecomcloudservice.h
+    ${CMAKE_CURRENT_LIST_DIR}/musescorecom/musescorecomcloudservice.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/audiocom/iaudiocomcloudservice.h
+    ${CMAKE_CURRENT_LIST_DIR}/audiocom/audiocomcloudservice.h
+    ${CMAKE_CURRENT_LIST_DIR}/audiocom/audiocomcloudservice.cpp
```
RomanPudashkin, line 49: I don't insist, but audiocomservice and musescorecomservice (without the word "cloud") look simpler

src/cloud/musescorecom/imusescorecomcloudservice.h
```diff
@@ -19,33 +19,36 @@
...
 {
-    INTERFACE_ID(ICloudProjectsService)
+    INTERFACE_ID(IMuseScoreComCloudService)
 
 public:
-    virtual ~ICloudProjectsService() = default;
+    virtual ~IMuseScoreComCloudService() = default;
 
-    virtual framework::ProgressPtr uploadScore(QIODevice& scoreData, const QString& title, Visibility visibility = Visibility::Private,
+    virtual QUrl scoreManagerUrl() const = 0;
```
RomanPudashkin, line 43: I would suggest keeping this method in ICloudConfiguration

src/cloud/audiocom/audiocomcloudservice.cpp
Eism: Disagree, it is a configuration for servers

src/cloud/musescorecom/imusescorecomcloudservice.h
Eism: Let's leave it as it is now. I like this method better in service

PR #17351, opened 2023-04-25
Fix gpconverter adds extra points for the same bend

mikekirin:


src/importexport/guitarpro/internal/gtp/gpconverter.cpp
```diff
@@ -1947,7 +1951,7 @@ void GPConverter::addBend(const GPNote* gpnote, Note* note)
         if (gpBend->middleOffset1 == -1 && gpBend->middleOffset2 == -1 && gpBend->middleValue != -1) {
             //!@NOTE It seems when middle point is places exactly in the middle
             //!of bend  GP6 stores this value equal -1
-            if (gpBend->destinationOffset > 50 || gpBend->destinationOffset == -1) {
```
mikekirin, line 1950: When the destinationOffset is equal to -1 we adding an extra point for the same bend in the next if statement.

alexpavlov96 (approved):

alexpavlov96 (approved):

PR #17361, opened 2023-04-26
Use `DO_ASSERT()` instead of `assert()` thoughout tread.cpp and twrite.cpp

Jojo-Schmitz:


src/engraving/rw/400/twrite.cpp
```diff
@@ -198,6 +198,9 @@ class WriteVisitor : public rtti::Visitor<WriteVisitor>
 void TWrite::writeItem(const EngravingItem* item, XmlWriter& xml, WriteContext& ctx)
 {
     bool found = WriteVisitor::visit(WriteTypes {}, item, xml, ctx);
+#ifdef NDEBUG // avoid compiler warning
+    UNUSED(found);
+#endif
     DO_ASSERT(found);
```
igorkorsukov, line 204: Thanks, but I already fixed it, I improved the `DO_ASSERT` macro, 
 (the fix will be in my next PR)

igorkorsukov: https://github.com/musescore/MuseScore/pull/17365/commits/9f33cede7c3a4da5f4ff7fc1c2440c895e57266c

igorkorsukov: > (Wouldn't it be wise to have that UNUSED() inside these Q_ASSERT() macros?)

Exactly!

Jojo-Schmitz: I tried the same, but got tons of compiler errors (when `cond` is something like `value = function()`)

igorkorsukov: I think `value = function()` should not be, maybe just `function()`

Jojo-Schmitz: Ah, it is gone, guess I didn't do a `make clean` before.

igorkorsukov (approved):

PR #17365, opened 2023-04-26
[engraving] Moved layout Accidental, Ambitus... Step 1

igorkorsukov:
Reasons:
1. We are considering the need to change the layout algorithm. In the new algorithm, it is possible that there will be a layout not of tree branches from parent to child, but, as it were, in layers, first the smallest elements, then larger and larger.
2. We would like to be able to draw the same elements in different layouts (for example, draw a fragment in a separate panel). i.e. the domain elements are the same, but the layout is different.
3. We would like to separate the layout into a separate subsystem in order to be able to effectively modify it, simplify it, reduce side effects, separate it from other subsystems, and so on.
4. We would like to be able to use different versions of the layout depending on the conditions (or just different versions, during the testing and stabilization phase). i.e. we want easy switching between layouts.
5. We would like to be able to unit test the layout more simply, so that it is easy to supply input parameters and check the output.

For these reasons, I started (rather even continued) moving the layout into separate classes.
Now the first phase is to just remove the layout from the objects. Then the phase of changing algorithms and implementation will begin, so the result may not look like it looks now (maybe in the future there will not be this class TLayout)

Just in case, we do not intend to take out all subsystems from objects (for example, we will not take out draw separately). We take out only read/write to simplify support for backward compatibility and make it easier to export (so that export and native write are not fundamentally different). And we take out layout for the above reasons.


RomanPudashkin (approved):

src/framework/global/log.h
```diff
@@ -31,6 +31,7 @@
 #define DO_ASSERT_X(cond, msg) \
     LOGE() << "\"ASSERT FAILED!\": " << msg << ", file: " << __FILE__ << ", line: " << __LINE__; \
     assert(cond); \
+    (void)(cond); \    // avoid compiler warning
```
cbjeukendrup: In a release build, doesn't this mean that the condition is unnecessarily evaluated? 

If you have `assert(someSimpleBooleanVariable)`, it's fine. But if you have `assert(evaluate(some, complicated[expression]))`, then I'm afraid that it means that `evaluate(some, complicated[expression])` will be evaluated unnecessary. As far as I know, `(void)` only discards the result, but does not per definition mean that the expression won't be evaluated. 

As a solution, I propose to make two macros: 
- `DO_ASSERT` which is simply `assert`, so that it can be used for complicated expressions and does completely nothing in release builds
- `ASSERT_OR_UNUSED` which is useful for asserting the value of a variable that would otherwise be unused.

igorkorsukov: I made the first implementation a little careless
Changed the implementation:
* Now the assertion message will be logged in the release as well. Of course, the expression will also be evaluated in the release, but I think the assertion triggering is a mistake, we should not hide it.
* If we need to check something only in debug, I added an explicit macro `DO_ASSERT_IF_DEBUG`

What do you think of this approach?

cbjeukendrup: That makes sense to me. Seeing the failure message in release builds may help understanding crashes from log files submitted by users. 

However, of course, we must make sure that for every case we make a wise decision about whether to use `DO_ASSERT` or `DO_ASSERT_ON_DEBUG`; for cheap assertions that check a critical error condition, we should use `DO_ASSERT`, but for more expensive assertions that are actually just "sanity checks", we should consider using `DO_ASSERT_ON_DEBUG` instead.

PR #17369, opened 2023-04-27
fixed #16104: No indication when error occurs while saving a file

### LINKED ISSUE #16104: [MU4 Issue] No indication when error occurs while saving a file
**Describe the bug**
Users have reported that sometimes nothing happens when saving, nor when clicking "Save" in the "Do you want to save" dialog. The dialog just disappears, but the score stays unsaved, and MuseScore doesn't close.

**Proposed solution**
We should show a dialog that saving failed, preferably with detailed information about why, so that the user can do something about it.

Came up in:
- Discord (https://discord.com/channels/818804595450445834/818804809784229898/1068615895938773013)
- https://musescore.org/en/node/340287

Eism:
Resolves: #16104

### DIFF
--- a/src/framework/global/io/iodevice.h
+++ b/src/framework/global/io/iodevice.h
@@ -24,6 +24,8 @@
 #define MU_IO_IODEVICE_H
 
 #include <cstdint>
+#include <string>
+
 #include "types/bytearray.h"
 
 namespace mu::io {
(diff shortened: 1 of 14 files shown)


src/framework/global/io/iodevice.h
```diff
@@ -75,12 +81,17 @@ class IODevice
...
+    void setError(int error, const std::string& errorString);
+
 private:
 
     const uint8_t* cdataOffsetted() const;
 
     OpenMode m_mode = OpenMode::Unknown;
     size_t m_pos = 0;
+
+    int m_error = -1;
```
igorkorsukov, line 93: I think, better use '0' for indicate no error, like Ret::Ok

igorkorsukov (approved):

PR #17237, opened 2023-04-13
Clefs and key signatures positioning improvement

mike-spa:
Resolves: #16814 
Resolves: #15311
Resolves: #16908

Issue #15705, opened 2023-01-04
[MU4 Task] Changes to consider when updating the file format


cbjeukendrup:
We don't get the chance to make breaking changes to the file format _that_ often, so when we do get such a chance, we should fully utilise it. Therefore I thought it would be good to build a list together of things we would like to consider changing. 

- [ ] Rename the `instrumentId` property to `musicXmlId`, which is far more accurate
- [ ] Maybe things necessary for "not-yet-fully-created" parts, see [#15546  (comment)](https://github.com/musescore/MuseScore/issues/15546#issuecomment-1368498512)
- [ ] ...

cbjeukendrup:
@igorkorsukov @shoogle @mike-spa I thought I'd quickly mention you, so that you know about the existence of this list, since you doubtlessly have ideas about this too.

cbjeukendrup:
Assigning this to the 4.1 project, since 4.1 will be the first chance to make changes to the file format.

shoogle:
See also https://musescore.org/en/node/296654 (you need to be logged in to MuseScore.org to see the "What links here" list).

Jojo-Schmitz:
PR #17001 and #17237 -> #17402

lgvr123:
When the file format changes, what about changing the extension ?

In GuitarPro, when they change the file format and broke the ascendant compatibility, they change the extension name. E.g. Guitar Pro 4 => `.gp4` , Guitar Pro 6 => `.gpx`, Guitar Pro and beyond `.gp`

Furthermore, they include and export mechanism to the previous formats.

Working with MU with other people is a hell. Between the ones using MU4.x and the ones sticking to MU3.6 (and there are numerous), the collaboration is complicated. Especially, when the file extension doesn't tell if the file has been saved in MU3.6 format or MU4.x format.

Jojo-Schmitz:
My Mu3.7 can now read Mu4 scores pretty well...

shoogle:
> When the file format changes, what about changing the extension?

FYI, we [no longer preserve compatibility between minor releases](https://github.com/musescore/MuseScore/issues/18083#issuecomment-1742204473).

> The consequence, in our music group, is that everyone must keep using MU3.6.

You would have to do this anyway, at least for your shared scores. Changing the file extension wouldn't allow MS3 to open MS4 scores.

> they include and export mechanism to the previous formats.

[...]

PR #17402, opened 2023-04-29
Increasing file format tag to "4.10"

mike-spa:
This is necessary because both PR #17001 and PR #17237 introduce new items and properties which can't be read by older versions.


Jojo-Schmitz:
Gives a chance to merge #12133 too

Jojo-Schmitz:
@cbjeukendrup said to have a list of things to consider when changing the file format...

src/engraving/libmscore/mscore.h
```diff
@@ -78,6 +78,11 @@ static constexpr int MSCVERSION = 400;
 //       - The style is stored in a separate file (inside mscz)
 //       - The ChordList is stored in a separate file (inside mscz)
 
+//    4.10 (Version 4.1)
+//       - New "Expression" item
+//       - A bunch of new options for dynamics
+//       - Clefs carry a "header" tag (istead of trying to guess it from context)
```
Jojo-Schmitz, line 84: typo: 'istead' -> 'instead'

igorkorsukov:
@mike-spa  I think we should make new read implementation for format 4.1:
* Make copy of dir 400
* Rename to 410
* Rename namespace `mu::engraving::rw400` to `mu::engraving::rw410`
* Rename class `Read400` to `Read410`
* Add make reader in https://github.com/musescore/MuseScore/blob/master/src/engraving/rw/mscloader.cpp#L53
* And so on 

After, at 410 we will start removing the old unnecessary code for supporting old formats (it will remain at 400 and below) and we will add new code that we want to change https://github.com/musescore/MuseScore/issues/15705

PR #17403, opened 2023-04-29
Fix MSVC compiler warnings

Jojo-Schmitz:
* reg. conversion from 'size_t' to 'int', possible loss of data (C4267)
* reg. declaration hides class member (C4458)
* reg. declaration hides function parameter (C4457)

### DIFF
--- a/src/notation/view/widgets/editstyle.cpp
+++ b/src/notation/view/widgets/editstyle.cpp
@@ -2082,8 +2082,8 @@ void EditStyle::valueChanged(int i)
     PropertyValue val  = getValue(idx);
     bool setValue = false;
     if (idx == StyleId::MusicalSymbolFont) {
-        bool dynamicsOverrideFont = getValue(StyleId::dynamicsOverrideFont).toBool();
-        if (!dynamicsOverrideFont) {
+        bool overrideDynamicsFont = getValue(StyleId::dynamicsOverrideFont).toBool();
+        if (!overrideDynamicsFont) {
             setStyleValue(StyleId::dynamicsFont, val); // Match dynamics font
         }
         if (optimizeStyleCheckbox->isChecked()) {
--- a/src/engraving/layout/tlayout.cpp
+++ b/src/engraving/layout/tlayout.cpp
@@ -1398,8 +1398,8 @@ void TLayout::layout(Expression* item, LayoutContext&)
         track_idx_t startTrack = track2staff(item->staffIdx());
         track_idx_t endTrack = startTrack + VOICES;
         for (track_idx_t track = startTrack; track < endTrack; ++track) {
-            EngravingItem* item = segment->elementAt(track);
-            if (item && item->isChord()) {
+            EngravingItem* engravingItem = segment->elementAt(track);
+            if (engravingItem && engravingItem->isChord()) {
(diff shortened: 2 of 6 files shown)


src/notation/view/widgets/editstyle.cpp
```diff
@@ -862,8 +862,8 @@ EditStyle::EditStyle(QWidget* parent)
             connect(checkBox, &QCheckBox::stateChanged, setSignalMapper, mapFunction);
         } else if (auto button = qobject_cast<QAbstractButton*>(sw.widget)) {
             connect(button, &QAbstractButton::toggled, setSignalMapper, mapFunction);
-        } else if (auto groupBox = qobject_cast<QGroupBox*>(sw.widget)) {
-            connect(groupBox, &QGroupBox::toggled, setSignalMapper, mapFunction);
+        } else if (auto myGroupBox = qobject_cast<QGroupBox*>(sw.widget)) {
```
alexpavlov96, line 865: why is it needed? didn't see the same name above

src/engraving/compat/midi/event.h
```diff
@@ -102,8 +102,8 @@ class NPlayEvent : public PlayEvent
 
     const Note* _note = nullptr;
     const Harmony* _harmony = nullptr;
-    int _origin = -1;
-    int _discard = 0;
+    size_t _origin = (size_t)-1;
```
alexpavlov96, line 105: (size_t)-1 is safe, but it's rather std::numeric_limits<size_t>::max()
but then we have 'getOriginatingStaff() + 1' and it's confusing, even though will give us zero
maybe leave int everywhere?

src/notation/view/widgets/editstyle.cpp
Jojo-Schmitz: It is there, somewhere, check the output of other Windowes builds:
```
2023-05-09T13:35:01.2361479Z D:/a/MuseScore/MuseScore/src/notation/view/widgets/editstyle.cpp(865): warning C4458: declaration of 'groupBox' hides class member
2023-05-09T13:35:01.4484152Z D:\a\MuseScore\MuseScore\build.release\src\notation\ui_editstyle.h(1017): note: see declaration of 'Ui_EditStyleBase::groupBox'
```

src/engraving/compat/midi/event.h
Jojo-Schmitz: It was `int` everywhere. And lead to a warning, I beliebe it is this'
```
2023-05-09T13:31:26.9125539Z D:/a/MuseScore/MuseScore/src/importexport/midi/internal/midiexport/exportmidi.cpp(317): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data
```

src/notation/view/widgets/editstyle.cpp
Jojo-Schmitz: Maybe we can fix it there, will have a look.

src/engraving/compat/midi/event.h
Jojo-Schmitz: But we can fix it with a single cast in exportmidi.cpp
I'll do that

RomanPudashkin (approved):

Issue #17387, opened 2023-04-28
Crash after deleting several instruments (one score)


DmitryArefiev:
### Issue type

Crash or freeze

### Steps to reproduce

1. Open [Smoke Test 8a.zip](https://github.com/musescore/MuseScore/files/11353909/Smoke.Test.8a.zip)
2. Press **I** to open Add or remove instruments dialog
3. Delete **Contrabasses** and **Violonchellos** and click OK

### Screenshots/Screen recordings

https://user-images.githubusercontent.com/90187801/235160774-d3cd8844-b9c7-4acb-90a5-71d2b6cb6280.mp4

### MuseScore Version

4.1

### Regression

Yes, this used to work in a previous version of **MuseScore 4.0.2**

### Operating system

Windows 10

DmitryArefiev:
Can't recreate from scratch for now

cbjeukendrup:
I've also seen this crash a few times. One way to reproduce it is very simple: just create a score from the Symphony Orchestra template, and then delete the Bass Drum using the Add/Remove Instruments dialog.

`git bisect` indicates that this was caused by commit 117e457ea09c4ead1a81479c7f37d5eb5ec5e8ee. Somehow, this causes the method `MasterScore::removeDeletedMidiMapping()` to malfunction; the InstrChannel is correctly moved within the `_midiMapping` vector, but the channels of the `InstrChannels` of the `part->instrument()` are not updated, which causes the crash later in `MasterScore::updateMidiMapping()`, because of reading out of bounds. 

That's all I know so far. I don't quite see why this is the case, and can't spot any mistakes in `TRead::read(InstrChannel*, …)` nor `TWrite::write(InstrChannel*, …)` either. @igorkorsukov Could you maybe also take a look? Maybe you do see what's wrong...

cbjeukendrup:
I found the problem, is was really quite subtle :)
https://github.com/musescore/MuseScore/pull/17405

DmitryArefiev:
Fixed in #17405

PR #17405, opened 2023-04-29
Fix that channels in `_midiMapping` got out of sync with channels in `Parts`

cbjeukendrup:
The MasterScore stores `InstrChannel`s in its `_midiMapping` field, and the `Instrument`s stored by the `Part`s also store `InstrChannel`s. The data of these `InstrChannel`s is synchronised from the `_midiMapping`'s `InstrChannel`s to the `Part`s' `InstrChannel`s.

In the old reading code for `InstrChannel`, the member fields would be set directly. But in the new code, we have to call the `set…` methods. But these methods normally emit the notification to sync with the other `InstrChannel`. To emulate the behaviour of the old code, we temporarily disable this notification. But we re-enable it too late, namely at the end of the method. This means that when `MasterScore::addMidiMapping` is called, which copies the `InstrChannel`, the notification is still disabled, so in the copy it will be disabled too, and will never be enabled again. And the synchronisation is supposed to go from the copy to the "original", so because the copy will never emit any notification, the synchronisation won't happen. This causes things to get out of sync, which causes a crash much later, as described in #17387. This crash happened in `int MasterScore::updateMidiMapping()`, due to reading out of bounds.

Resolves: #17387

### DIFF
--- a/src/engraving/rw/400/tread.cpp
+++ b/src/engraving/rw/400/tread.cpp
@@ -21,8 +21,6 @@
  */
 #include "tread.h"
 
-#include "global/defer.h"
-
 #include "../../types/typesconv.h"
 #include "../../types/symnames.h"
 #include "../../infrastructure/rtti.h"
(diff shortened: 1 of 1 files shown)


igorkorsukov (approved):

cbjeukendrup:
@DmitryArefiev I investigated and found that this crash is technically not really related to the one that this PR is fixing, but it will be fixed by https://github.com/musescore/MuseScore/pull/17409. So the best way might be to merge this PR, and then I rebase https://github.com/musescore/MuseScore/pull/17409 on top of it, and then we can test if the crash is indeed fixed by the combination of the PRs.

DmitryArefiev:
@cbjeukendrup Yeah, agree. Let's merge it!

Issue #17392, opened 2023-04-28
Crash when deleting staff from instrument


sammik:
### Issue type

Crash or freeze

### Bug description

When staff is added to instrument and than deleted, MU crashed. If there were at least two instruments.

### Steps to reproduce

1. score with two instruments
2. add staff to second instrument
3. remove staff from second isntrument
4. crash

### Screenshots/Screen recordings

https://user-images.githubusercontent.com/1646034/235218629-3d566db0-f18a-44c8-a908-c34c0c75d5b2.mp4

### MuseScore Version

4.1 master

### Regression

Yes, this used to work in a previous version of MuseScore 4.x

### Operating system

Ubuntu 22.04

### Additional context

may be related #17387

cbjeukendrup:
It turns out that this will be fixed by #17409!

DmitryArefiev:
Fixed in #17409

Issue #17390, opened 2023-04-28
Crash when deleting instruments in a certain order from a score with a certain list of instruments


cbjeukendrup:
1. Create score from the String Quartet template
2. Add Flute
3. Delete Flute (using Instruments panel)
4. Select the other instruments in the Instruments panel
5. Delete them
-> crash, related to brackets

Could happen in other situations too. This is just a reliable way to reproduce.

DmitryArefiev:
Fixed in #17409

Jojo-Schmitz:
I guess this is was a regression vs. 3.x?

cbjeukendrup:
Yeah, in a way... if only because MS3 does not have the UI to reproduce it.

PR #17409, opened 2023-04-29
Fix crash when removing instruments in some cases, caused by brackets

cbjeukendrup:
Resolves: #17390
Resolves: #17392

See commit messages for explanation

### DIFF
--- a/src/engraving/libmscore/edit.cpp
+++ b/src/engraving/libmscore/edit.cpp
@@ -6299,15 +6299,29 @@ void Score::undoAddBracket(Staff* staff, size_t level, BracketType type, size_t 
 {
     staff_idx_t startStaffIdx = staff->idx();
     staff_idx_t totStaves = nstaves();
+
     // Make sure this brackets won't overlap with others sharing same column.
     // If overlaps are found, move the other brackets outwards (i.e. increase column).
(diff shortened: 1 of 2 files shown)


cbjeukendrup:
@DmitryArefiev I rebased it and now it should indeed fix https://github.com/musescore/MuseScore/pull/17405#issuecomment-1531359555 too.

DmitryArefiev:
@cbjeukendrup All mentioned crashes are fixed. 

But I found **another one** with Undo+Redo and additional staves. Can you check it please?



https://user-images.githubusercontent.com/90187801/236423699-21bcb016-b4f1-4725-b8ba-7690a5a7c89e.mp4

cbjeukendrup:
@DmitryArefiev That one I haven't seen before. It seems unrelated to the earlier crashes, but I will investigate a bit more.

cbjeukendrup:
@DmitryArefiev I investigated more, and it turns out that this crash is unrelated to the crash this PR is solving, and it seems a bit more difficult to solve. So I think it is better to do it separately. I've already created an issue for it: #17494

DmitryArefiev:
@cbjeukendrup Cool. Thanks!

PR #17410, opened 2023-04-29
Fix #10403: Moving same-beat notes between strings

### LINKED ISSUE #10403: [MU4 Issue] Issues when moving same-beat notes between strings in tablature
**Describe the bug**

When working on Guitar tablature, moving multiple notes on the same beat between strings doesn't work as expected. 
It either fails, or causes weird changes.

**To Reproduce**
1. Create a new score with guitar tablature
2. Add 3 notes on the same beat
3. Select them
4. Press `Ctrl + Down` to move them to a lower string
5. Press `Ctrl + Up` to move them to the original position

**Expected behavior**
After moving down one string, and then back up, the notes are in their original state.

**Screenshots**
2. Add 3 notes on the same beat
    image
3. Select them
    image
4. Press `Ctrl + Down` to move them to a lower string
    image
5. Press `Ctrl + Up` to move them to the original position
    image

**Platform information**
 - OS: Windows 10
 - MuseScoreNightly-202201290418-master-b8de8be-x86_64
 - Reproduced on MuseScore 3.6 as well

**Additional context**
It seems that note selection type & order matters here.

If we select using `Shift`, we get one behavior. If we select using `Ctrl`, we get another.
When selecting using `Ctrl`, it seems that note selection order matters as well.

If I select notes from top to bottom, I get issues when moving them down.
If I select notes from bottom to top, I get issues moving them up, but moving down works well.

My assumption is that the notes are moved in sequence, and not as a group. As a result - the order in which they are moved is significant, as notes may block the movement of other notes.

tmr232:
Sort the notes by string, so that moved notes don't block each other.

Resolves: #10403 

Moving same-beat notes between strings currently depends on selection order.
We first try to move the notes that are selected first, resulting in conflicts when moving chords.

By sorting the notes by string, we ensure that the notes of a chord won't block each other if a move is possible.

### DIFF
--- a/src/engraving/libmscore/cmd.cpp
+++ b/src/engraving/libmscore/cmd.cpp
@@ -1646,6 +1646,14 @@ void Score::upDown(bool up, UpDownMode mode)
 {
     std::list<Note*> el = selection().uniqueNotes();
 
+    el.sort([up](Note* a, Note* b) {
+        if (up) {
+            return a->string() < b->string();
+        } else {
+            return a->string() > b->string();
+        }
+    });
+
     for (Note* oNote : el) {
         Fraction tick     = oNote->chord()->tick();
         Staff* staff = oNote->staff();


src/engraving/libmscore/cmd.cpp
```diff
@@ -26,7 +26,8 @@
 */
 
 #include <assert.h>
-
+#include <algorithm>
+#include <iostream>
 #include "translation.h"
```
cbjeukendrup, line 31: It looks like `<iostream>` is not necessary. Also, from a code-stylistic point of view, it is preferable to keep a newline between #includes from the standard library and our own #includes.
```suggestion
#include <algorithm>

#include "translation.h"
```

cbjeukendrup (commented):
Thanks for your contribution! Looks like a good solution. I only have a comment about some details:

src/engraving/libmscore/cmd.cpp
tmr232: Oh, thanks!
I missed that when removing my debug prints.

tmr232: Pushed an amended commit.
Should I mark this as resolved, or let you do that?

src/engraving/libmscore/cmd.cpp
```diff
@@ -26,6 +26,7 @@
 */
 
 #include <assert.h>
+#include <algorithm>
```
cbjeukendrup, line 29: I just realised that you don't need `<algorithm>` either, since the `sort` you are using is a member of `std::list`. Sorry, I should have spotted that earlier...

src/engraving/libmscore/cmd.cpp
cbjeukendrup: Regarding marking conversations as solved, feel free to do that when you're confident you've made the correct change. 

For debug logging, you could also use our `LOGD() << …` (or `LOGI` or `LOGW` or `LOGE`). Sometimes that's more convenient, because it also prints the name of the method that's being called and it can handle some custom types like `mu::String` etc.

src/engraving/libmscore/cmd.cpp
tmr232: You're quite right!
I'll fix that.

cbjeukendrup (approved):
Looks perfect now! Thanks!

DmitryArefiev:
Tested #10403 on Win10, Mac13, LinuxUbuntu 22.04 - **FIXED**

Thanks!

Issue #17352, opened 2023-04-25
Dropdown called from Mixer or Palette can't be closed by clicking out of it


abariska:
### Issue type

Other type of issue

### Bug description

Probably, the issue appeared after the fix https://github.com/musescore/MuseScore/pull/17268

### Steps to reproduce

1. Open any score
2. Open Mixer window
3. Tap a plugin cell to call dropdown with plugin list
4. Click out of it (e.g. any note or score sheet)

### Screenshots/Screen recordings

https://user-images.githubusercontent.com/101250347/234283458-211177fc-4ed5-44f2-a1c6-d1f194fb6d24.mp4

### MuseScore Version

MuseScore 4.1

### Regression

Yes, this used to work in a previous version of MuseScore 4.x

### Operating system

macOS 13

fmndantas:
Hi @abariska. Can I try to solve this issue?

abariska:
@fmndantas it's a regression after another fixes, so better if @Eism and @cbjeukendrup will check it first

DmitryArefiev:
The same with Properties:


https://user-images.githubusercontent.com/90187801/234847873-9a977711-3185-413a-966f-62ac8ed56af0.mp4

(will add to the summary)

abariska:
🟢 **FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

Issue #15481, opened 2022-12-21
Dropdown in settings going behind menu


MyCatIsAChonk:
**Describe the bug**
A short and clear description of what the bug is.
The dropdown menu in Preferences > Import > MIDI is going behind the menu, as shown in the screenshot. I'm using MacOS Ventura 13.0.1.

**To Reproduce**
1. Go to 'Preferences'
2. Click on 'Import'
3. Scroll down to 'MIDI'
4. Select the dropdown menu
5. See error (at least, on MacOS Ventura 13.0.1; not sure if this is the case on other OS's)

**Expected behavior**
I expect the dropdown to come in front of the menu so I can click one of the options.

**Screenshots**
![image](https://user-images.githubusercontent.com/121187295/208960740-68d30dce-99b3-4b11-aa44-7f2dee42a602.png)

**Platform information**
 - OS: MacOS Ventura 13.0.1

**Additional context**
N/A

Eism:
@MyCatIsAChonk 
I can't reproduce it on macOS Ventura 13. Need more info

Are you using a multi-monitor setup?

MyCatIsAChonk:
Yes, I have it pulled up on an alternate monitor connected to my actual computer. The monitor is a Dell S2721HN.

abariska:
🟢 **FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

abariska:
🔴 Reproducible again after the PR https://github.com/musescore/MuseScore/pull/17268

abariska:
🟢 **FIXED**
Tested on macOS 12, Win 11, Ubuntu 20

PR #17413, opened 2023-04-29
Init popup close controller *properly*

cbjeukendrup:
In #17268, we delayed the initialisation of the PopupView's CloseController until *after* PopupView's constructor, so that we can check if it is a popup or a dialog, and only use the CloseController when it is a popup.

This revealed that when initialising the CloseController, we were forgetting to set some of its properties.

Why it *did* work before that PR? That was because we *do* update the properties of the CloseController at the moment that the properties of the PopupView are updated. That happens very early, so acted as a replacement for setting all properties correctly during initialisation. But after the PR, the CloseController is not yet initialised at that moment, so the properties would not be set at all.

Resolves: #15481 
Resolves: #17352

### DIFF
--- a/src/framework/uicomponents/view/popupview.cpp
+++ b/src/framework/uicomponents/view/popupview.cpp
@@ -159,6 +159,11 @@ void PopupView::initCloseController()
 
     m_closeController->init();
 
+    m_closeController->setParentItem(parentItem());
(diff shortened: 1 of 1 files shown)


igorkorsukov (approved):

abariska:
Please, merge to master

Issue #17349, opened 2023-04-25
Corruption when Ctrl-deleting measure containing global time signature after measure with local time signature


cbjeukendrup:
### Issue type

File corruption

### Steps to reproduce

1. Create score with multiple staves in 4/4
2. Add local  3/4 time signature in measure 2
3. Add global 4/4 time signature in measure 3 (this must match the initial time signature)
4. Select measure 3
5. Ctrl+Delete
-> corruption: the contents of the measures after the deleted measures is still in 4/4, but it should be 3/4 in the top staff, or a new 4/4 time signature should be created in that staff

https://user-images.githubusercontent.com/48658420/234137076-5b937d9c-c4d7-4fe7-b61c-569e60dc8b0d.mov

### MuseScore Version

master

### Regression

No.

### Operating system

n/a

Jojo-Schmitz:
FWIW: same issue in Mu3

cbjeukendrup:
@DmitryArefiev I'm not really sure...

But the motivation for the current behaviour is: the user tries to delete the measure containing the time signature, so in principle we delete the time signature too, _unless_ a time signature is necessary because the following measures are in a different time signature. 
In your example, the time signature on the lower staff is not necessary, so it is deleted together with the measure. Musically, this is perfectly fine. 

This behaviour might be justified by more complex cases. For example, when the user tries to delete this passage:
<img width="722" alt="Scherm­afbeelding 2023-05-09 om 19 08 46" src="https://github.com/musescore/MuseScore/assets/48658420/80cebfb9-30dd-4d02-bb85-28ecd525ab4e">
Then you only want a 4/4 time sig in the upper staff, because of the local 3/4 time sig that will be preserved. But you definitely don't want any other time signature to be preserved on the bottom staff, since there is simply no reason for it. The original time signature was there because it was necessary, but after deleting those measures, no time sig is necessary anymore.

DmitryArefiev:
> In your example, the time signature on the lower staff is not necessary, so it is deleted together with the measure. Musically, this is perfectly fine.

@cbjeukendrup My only argue to have 4/4 on all staves is to make music more convenient for reading by musician because there were some TS changes before that global TS again. But it's not hard to add TS to all staves manually, so I'm fine to have how it is now. Let's merge it.

DmitryArefiev:
Fixed in #17418

PR #17418, opened 2023-04-30
Fix corruption when deleting measure containing time sig after measure containing local time sig

cbjeukendrup:
Resolves: #17349


RomanPudashkin (approved):

DmitryArefiev:
Tested #17349 on Win10, Mac13, LinuxUbuntu 22.04 - **FIXED**

Issue #13214, opened 2022-09-06
[MU4 Issue] Chord symbols do not render in MusicXML files


MarcSabatella:
**Describe the bug**
When loading a MusicXML file, any chord symbols present or loaded after do not render correctly - the roots are missing (so you see just "7" instead of "G7")

**To Reproduce**
1. Open any MusicXML file
2. If it has chord symbols, look at them
3. If not, try entering a new chord symbol: select first note, Ctrl+K, type "G7", hit Esc
4. See error: The chord symbols are missing their roots  "7" instead of "G7" etc.

**Expected behavior**
Chord symbols should display with their roots

**Screenshots**
![image](https://user-images.githubusercontent.com/1799936/188709321-2d605852-2fce-4736-accf-dd213f0649e7.png)

**Platform information**
 - OS: Linux

**Additional context**
The problem is that the chord description file that would normally be loaded as part of the basic score reading process seems to be skipped for MusicXML files.  The workaround is to go to Format / Style / Chord Symbols, change the appearance from Standard to Jazz and then back.  This forces the style file to be loaded.

This was originally reported in https://musescore.org/en/node/334422 but I just ran into it again and realized it was never reported here on GitHub.

lvinken:
Issue does not reproduce for me using commit 56b4855 on x86_64 Ubuntu 22.04.

edellaq:
On OSX it happens too.

My setup:
iMac (Retina 4K, 21.5-inch, Late 2015), 3,1 GHz Intel Core i5 quad-core
8 GB 1867 MHz DDR3, Intel Iris Pro Graphics 6200 1536 MB
OSX 12.6.2 (Monterey)

MarcSabatella:
@Tantacrul this is coming up quite often and is presumably a very simple fix.

@lvinken - I don't think the issue is likely to be in the MusicXML code specifically but in how the initialization of the score style is handled. Somewhere there should be a call to function that loads a chord description file during import.  Unfortunately I don't really know how it is supposed to work anymore.

I've edited the description to include a MusicXML file that reproduces the problem.

MarcSabatella:
That too; description updated.

MarcSabatella:
As per newly added comments, this also affect MIDI.  Wherever the code to initiatialize the chordlist moved to, it isn't happening for imported scores.

lvinken:
Will have a look at the MusicXML import code, but as far as I recall it does not explicitly load a score style.

[...]

PR #10113, opened 2021-12-16
[MU4] fix #10103: Chord symbols are recognized (parsed) properly only in a just created scores

RomanPudashkin:
Resolves: https://github.com/musescore/MuseScore/issues/10103

Issue #10103, opened 2021-12-15
[MU4 Issue] Chord symbols are recognized (parsed) properly only in a just created scores 


DmitryArefiev:
**To Reproduce**
1. Create a score
2. Add some notes
3. Save and close
4. Reopen
5. Select note and enter chord input mode
6. Type **am7** and hit Space - you see **m7** instead of Am7 

https://user-images.githubusercontent.com/90187801/146188507-31649f9d-ee89-439c-a47f-7f195b636184.mp4

**Expected behavior**

**Screenshots**

**Desktop:**
 - OS: all

RomanPudashkin:
I'm working on it

DmitryArefiev:
Tested on Win10, Mac12, LinuxMint 20.2 - FIXED

PR #17427, opened 2023-05-01
[MusicXML] Apply fix for #10103 to all non-native imported files

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/13214

There was an issue (https://github.com/musescore/MuseScore/issues/10103) about chord symbols being mangled in version 4.0 of MuseScore. It was fixed by PR https://github.com/musescore/MuseScore/pull/10113 by adding an explicit call to `score->checkChordList()` for 4.0 scores. The issue at hand appears to be the exact issue that the PR fixed, so this PR follows in its footsteps by adding an explicit call to `checkChordList` in all cases of XML import.

### DIFF
--- a/src/importexport/bww/internal/bww/importbww.cpp
+++ b/src/importexport/bww/internal/bww/importbww.cpp
@@ -554,7 +554,7 @@ Err importBww(MasterScore* score, const QString& path)
     if (!fp.open(QIODevice::ReadOnly)) {
         return engraving::Err::FileOpenError;
     }
-
+    score->checkChordList();
     Part* part = new Part(score);
     score->appendPart(part);
     Staff* staff = Factory::createStaff(part);
--- a/src/importexport/capella/internal/capella.cpp
+++ b/src/importexport/capella/internal/capella.cpp
@@ -2826,6 +2826,7 @@ Err importCapella(MasterScore* score, const QString& name)
         return Err::NoError;
     }
     fp.close();
+    score->checkChordList(); // make sure chordList is populated so we can import harmony objects
     convertCapella(score, &cf, false);
     return Err::NoError;
 }
--- a/src/importexport/guitarpro/internal/importgtp.cpp
+++ b/src/importexport/guitarpro/internal/importgtp.cpp
@@ -2949,7 +2949,6 @@ static Err importScore(MasterScore* score, mu::io::IODevice* io)
     }
 
     score->loadStyle(u":/engraving/styles/gp-style.mss");
-
     score->checkChordList();
     io->seek(0);
     char header[5];
--- a/src/importexport/midi/internal/midiimport/importmidi.cpp
+++ b/src/importexport/midi/internal/midiimport/importmidi.cpp
@@ -1282,7 +1282,7 @@ Err importMidi(MasterScore* score, const QString& name)
         loadMidiData(mf);
         opers.setMidiFileData(name, mf);
     }
-
+    score->checkChordList();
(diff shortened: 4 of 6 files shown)


cbjeukendrup:
I believe the same applies to MIDI import and potentially other import formats. Would it be a good idea to fix those in this PR as well?

asattely:
I'll investigate that now! I admit my knowledge of all of the types we can import is lacking

Jojo-Schmitz:
Well, handbook to the rescire ;-)  https://musescore.org/en/handbook/4/full-table-contents#Import-file-formats

cbjeukendrup:
All importers are derived from the `INotationReader` class, so that might help as well.

asattely:
How's this? If an importer loads some kind of custom chordlist it'll overwrite it, but loading the default chordlist here ensures that every importer has a valid chordlist it can use.

Jojo-Schmitz:
And remove it from other places, like https://github.com/musescore/MuseScore/pull/10113/files#diff-3fb520990974dcda5b381307834f2710da86829ff30a8c41cabd84918a200cdeR179 ?

asattely:
> And remove it from other places, like https://github.com/musescore/MuseScore/pull/10113/files#diff-3fb520990974dcda5b381307834f2710da86829ff30a8c41cabd84918a200cdeR179 ?

The change I made does not apply to mscx files; it's only hit when importing non-native file types. So we need both!

cbjeukendrup:
Ha, just wanted to say the same :)

But I'm not sure whether `NotationProject::doImport` is the best place for this. For example, the unit tests (and VTests too, I believe) are not aware of the existence of `NotationProject`. On the other hand, we also call `score->checkChordList();` in places like `MasterNotation::setupNewScore`, of which the unit tests are not aware either...

Maybe we should create a general method somewhere that finalizes the score after it has been created or read, and call that method in all places where we such thing needs to be done? (So probably, that will still need to be inside each individual importer, so that it will also be called by the unit tests.)

Or maybe we should even call it in the constructor of `MasterScore`? We do call it in the constructor of `Score` that is used to create part scores, but we don't call it in the default `Score()` constructor nor in the `MasterScore` constructor, which seems odd to me.

asattely:
Using the constructor of `MasterScore` seems reasonable. It can't be set as a final step, because the importers use the chordList in the process of importing harmony objects. If I need to add it to every importer individually I can do that as well. I'll play around with it and see what I can come up with

cbjeukendrup:
(Somehow I have a feeling that I've already posted this comment before, but it looks like I didn't, so here we go:)

I'm still not entirely sure if this is the best solution, since it is so scattered. I've also looked at the 3.x code (where `checkChordList` was still a method of `MStyle`), and it turns out that originally we _did_ call `checkChordList` always _before_ importing any format, namely here: https://github.com/musescore/MuseScore/blob/2513676e512d29d554cb6c4d37d3efaf53ea2c5b/mscore/file.cpp#L2340-L2349
(in this piece of code, `score->style().load()` also calls `checkChordList`).

In MuseScore 4, it was simply not added back in an equivalent place.

With this in mind, I would suggest reconsidering to place it for example in `NotationProject::doImport` before the call to `scoreReader->read`.

MarcSabatella:
I seem to recall you saying that too, and I *know* I made a similar comment in the issue.  Absolutely, each importer should not have to deal with this.

Long term, it would be possible to largely eliminate a ton of the code involved with chord list management (much of which is due to legacy from MU1).  All the more reason not to increase the dependencies on the current system by adding calls more places than necessary.

Issue #17428, opened 2023-05-01
Irregular measure not marked when followed by start repeat sign


henkdegroot:
### Bug description

The irregular measure marker (- or +) is not shown when the next measure starts with a repeat sign.

![irregular-measure-length](https://user-images.githubusercontent.com/13550012/235522029-f2f1c27a-8e7a-4382-95f5-02f34dacffa2.gif)

### Steps to reproduce

Create a score
Change the length of a measure
Place the repeat sign in the next measure

### Screenshots/Screen recordings

See bug description

### MuseScore Version

4.0.2

### Regression

No.

### Operating system

Windows 11

Jojo-Schmitz:
More a UI bug than an engraving bug. Interesting that is is in Mu3 too and has gone undetected so long

Jojo-Schmitz:
Please assign to me, I have a fix almost ready

PR #17429, opened 2023-05-01
Fix #291847 and GH#17428:  Irregular measure not marked when followed by start repeat barline

Jojo-Schmitz:
Resolves: #17428 and https://musescore.org/en/node/291847

Same issue and fix in/for Mu3


Jojo-Schmitz:
Hmm, for an end/start repeat barline it'd write that irregular marking twice.
Better twice than none?

cbjeukendrup:
It's even a bit more confusing when a start repeat barline is preceded by a system break:
Scherm­afbeelding 2023-05-01 om 23 15 18

There's probably a way to find out if the previous measure has an end barline (in which case we shouldn't show the sign on the start barline of the next measure). Maybe check if `Measure::endBarLine()` returns `nullptr` or not?

Jojo-Schmitz:
Check it where exactly and how? I tried out all sorts of combinations but none worked

Jojo-Schmitz:
I got the first issue solved, but not the second, the one with start repesat barlines after a system break
Any hints?

Jojo-Schmitz:
Hopelessly outdated

Jojo-Schmitz:
Hmm, OK, rebased.

Jojo-Schmitz:
Rebased

Jojo-Schmitz:
@miiizen thanks for trying to fix my attempt to fix this issue ;-)
However: it shows the marking twice now, at the start and the end of the measure:
image

Jojo-Schmitz:
That looks OK now (to me), thanks

cbjeukendrup:
@zacjansheski This PR should fix whether the `+` or `-` sign for irregular measures appears. ("Irregular measures" being measures where the actual duration has been changed via Measure Properties.) It should especially be tested with multiple barline types, esp. repeat barlines, and in the presence of a (natural or manual) system break or not.

See the grey `+` sign at the end of the first system in this image:
Scherm­afbeelding 2025-11-12 om 02 42 38

zacjansheski:
Tested on MacOS 15, Windows 11, Ubuntu 22.04.3. Approved
https://github.com/musescore/MuseScore/issues/17428 FIXED

PR #17446, opened 2023-05-03
Fixed problems when compiling clang and use asan 

igorkorsukov:


src/engraving/libmscore/score.h
```diff
@@ -785,7 +785,7 @@ class Score : public EngravingObject
 
     void changeSelectedNotesVoice(voice_idx_t);
 
-    const std::vector<Part*>& parts() const { return _parts; }
+    const std::vector<Part*>& parts() const;
```
igorkorsukov, line 788: There was a very strange crash if we use this code after compiling its clang and cpp20

build/cmake/SetupBuildEnvironment.cmake
```diff
@@ -67,10 +69,14 @@ elseif(CC_IS_CLANG)
...
-        string(APPEND CMAKE_CXX_FLAGS_DEBUG " -fsanitize=address -fno-omit-frame-pointer")
+        add_compile_options("-fsanitize=address")
+        add_compile_options("-fno-omit-frame-pointer")
+        link_libraries("-fsanitize=address")
     endif()
 
-    if (BUILD_IS_DEBUG)
+    # On MacOS with clang there are problems with debugging
+    # - the value of the std::u16string is not visible.
+    if (BUILD_IS_DEBUG AND MUE_ENABLE_STRING_DEBUG_HACK)
```
igorkorsukov, line 79: Asan swears at this hack, the true reasons have not been identified, so I added the ability to turn it off

cbjeukendrup: What surprises me, is that I've never had any problems with this (and I use Asan almost always). Or do these problems only occur when using C++20?

igorkorsukov: I tried, with с++17 there is also a problem
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==68468==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000000000 bp 0x000000000000 sp 0x7ffd8d1fbcb8 T0)
==68468==Hint: pc points to the zero page.
==68468==The signal is caused by a READ memory access.
==68468==Hint: address points to the zero page.
    #0 0x0  (<unknown module>)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (<unknown module>) 
==68468==ABORTING
```

cbjeukendrup (approved):

build/cmake/SetupBuildEnvironment.cmake
cbjeukendrup: Strange... Anyway, adding an option to disable it won't harm anyway, so let's do that.

Issue #15866, opened 2023-01-13
[MU4 Issue] Drumset selection playback, wrong instruments


MekaSergal:
**Describe the bug**
When picking which part of the drumset to play at the bottom of the page, the clicked drum part makes the sound of a different one.

Edit from @bkunda: the conditions for incorrect playback are that a transposing instrument needs to be in the score _above_ the drumset stave, and "concert pitch" needs to be toggled OFF. Note it only affects note input playback, not regular playback.  

**To Reproduce**
1. Create score with alto saxophone and drumset.
2. Change saxophone clef to "Treble clef 8va alta"
3. Using the default note input tool, double click on a measure of the drumset to bring up the note selection menu at the bottom of the page.
4. Click on any of the drum sounds and see (hear) that they don't match.

**Expected behavior**
Drum part clicked makes corresponding drum sound 

**Platform information**
 - OS: Windows 10

Found while making this:
[Sister Marian.zip](https://github.com/musescore/MuseScore/files/10416097/Sister.Marian.zip)

MarcSabatella:
Note: the issue seems to be caused by the transposing instrument at the top of the score.  Somehow it is causing the playback of the drum palette to also be affected.  If you switch to concert pitch, save, and reload, the problem goes away.

MarcSabatella:
There was a related by in an earlier 3.x release but fixed for 3.5 - not on behavior for preview specifically, but for double-click.  See https://musescore.org/en/node/291522.  This was fixed for 3.5 in https://github.com/musescore/MuseScore/pull/6094/.  Not sure if this is related or not, but it could be useful info.  Since this is coming up pretty often and it seems likely to be a small fix, I'll continue looking into it.

MarcSabatella:
So far no luck figuring out where the problem is.  First couple of things I tried didn’t work.  I’d love someone else to take a look - I suspect with a working g build environment & debugger it would be pretty to track down.  I’m thinking that code adjusting TPC on read is probably not the issue, it’s probably more in the code to play the selected element and it’s using the staff index 0 but grabbing transposition info from the staff 0 of the actual score instead of the palette score.

MarcSabatella:
I see this is on the list for 4.x but not 4.1.  It is actually one of the most commonly-reported bugs still outstanding.  @sammik  does this seem like something you might be able to take a look at, or at least help me track down (see the info I posted above)?

sammik:
[...]

PR #17453, opened 2023-05-03
fix drum palette / concert pitch bug (DRAFT)

MarcSabatella:
Resolves: #15866

A similar issue in 3.x was fixed in a way that seems incomplete.  This is a trial PR to see if the underlying issue is actually the same by disabling the relevant code, which was needed only to auto-fix certain corrupt scores.  If this is indeed the culprit, then a more complete fix can be developed.

- [ ] I signed the [CLA](https://musescore.org/en/cla)
- [ ] The title of the PR describes the problem it addresses
- [ ] Each commit's message describes its purpose and effects, and references the issue it resolves
- [ ] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [ ] There are no unnecessary changes
- [ ] The code compiles and runs on my machine, preferably after each commit individually

### DIFF
--- a/src/engraving/libmscore/note.cpp
+++ b/src/engraving/libmscore/note.cpp
@@ -1524,7 +1524,7 @@ void Note::setupAfterRead(const Fraction& ctxTick, bool pasteMode)
     // but since there are other causes of tpc corruption (eg, https://musescore.org/en/node/74746)
     // including perhaps some we don't know about yet,
     // we will attempt to fix some problems here regardless of version
-
+#if 0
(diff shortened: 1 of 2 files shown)


MarcSabatella:
Hmm, seem this isn't it.  But I'll keep investigating.

MarcSabatella:
At this point I'd love it if someone else could take over.  My guess is that the code to playback the selected note is using the staff index of 0 for the palette note and interpreting it relative to the actual score instead when getting the transposition info.  But I can't figure out where that is actually happening.  Would presumably be simple for someone working in a debugger...

PR #17462, opened 2023-05-04
Converter. Fixed backslashes in path for Windows

Eism:
If you invoke MuseScore with paths containing backslashes, the file writing in mscz format may be incorrect (see screenshot). 
`.\MuseScore 4.exe' C:\Users\elnur\Downloads\River_Flows_In_You.mxl -o C:\Users\elnur\Downloads\River_Flows_In_You.mscz`

This is because the paths contain backslashes and some algorithms interpret this as escaping special characters. (in our case it is QFileInfo::completeBaseName())

Solution: correct the paths when parsing the command line.

image

### DIFF
--- a/src/app/commandlineparser.cpp
+++ b/src/app/commandlineparser.cpp
@@ -21,6 +21,7 @@
  */
 #include "commandlineparser.h"
 
+#include "global/io/dir.h"
 #include "global/muversion.h"
 
 #include "log.h"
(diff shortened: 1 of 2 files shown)


igorkorsukov (approved):

DmitryArefiev:
Tested #17463 on Win10 - **FIXED**

PR #17464, opened 2023-05-04
fixed #16710: Command-line export of .mpos and .spos files is broken in MuseScore 4

### LINKED ISSUE #16710: Command-line export of .mpos and .spos files is broken in MuseScore 4
### Issue type

Other type of issue

### Bug description

Up to (and including) MuseScore 3.6.2, it was possible to use the command-line to export various files from a score, using the -o option (midi, wav, mp3, png, ...), following this documentation of the MuseScore command line options:  https://musescore.org/en/handbook/3/command-line-options

Most of these types of export still work, but the export of measure and segment position files (.mpos and .spos files) however does NOT work any longer in MusecScore 4.0..0 or 4.0.1.

### Steps to reproduce

1. Preamble: I did this on Windows 10 Pro, but it most probably is the same on macOS and Linux as well. MuseScore 3.6.2 and MuseScore 4.0.1 are installed and work properly.
2. Use the attached test score as an example: [La_Llorona.mscz.zip](https://github.com/musescore/MuseScore/files/10922176/La_Llorona.mscz.zip) (note: .zip was added for upload purposes, so you'll need to remove that extension again; the original file can also be obtained from: https://musescore.com/user/29343852/scores/6546801 )
3. Open a command prompt and first try the export with MuseScore 4.0.0 or 4.0.1:
    ```
    "C:\Program Files\MuseScore 4\bin\MuseScore4.exe" -o La_Llorona.mpos La_Llorona.mscz
    "C:\Program Files\MuseScore 4\bin\MuseScore4.exe" -o La_Llorona.spos La_Llorona.mscz
    ```
    Result: you do NOT get any output at all (.mpos and .spos files are NOT generated)..
4. Then try the export with MuseScore 3.6.2:
    ```
    "C:\Program Files\MuseScore 3\bin\MuseScore3.exe" -o La_Llorona.mpos La_Llorona.mscz
    "C:\Program Files\MuseScore 3\bin\MuseScore3.exe" -o La_Llorona.spos La_Llorona.mscz
    ```
    Result: you DO get the La_Llorona.mpos and La_Llorona.spos files as expected.

_No response_

### MuseScore Version

4.0.1

### Regression

### Operating system

Windows 10 Pro 64bit (latest updates)

### Additional context

**Original code**

The original code for this functionality (in branch 3.6.2) was here:
[https://github.com/musescore/MuseScore/blob/3.6.2/mscore/musescore.cpp#…](https://github.com/musescore/MuseScore/blob/3.6.2/mscore/musescore.cpp#L3715)
which then called into this code:
[https://github.com/musescore/MuseScore/blob/3.6.2/mscore/savePositions…](https://github.com/musescore/MuseScore/blob/3.6.2/mscore/savePositions.cpp#L51)

**Related info outside of GitHub**

This issue had already been discussed by another user ("Thibaultj") of this functionality here, while MuseScore 4 development was still in its early stages, but it never got fixed before the first 4.x release:
https://musescore.org/en/node/315219

It was also briefly discussed on the musescore.org issue forums, but I think it's best to have this stored in GitHub as a bug:
https://musescore.org/en/comment/1178056
Note that on that page, a PR is mentioned ( https://github.com/musescore/MuseScore/pull/13153 ), but from what I could see in that PR, mainly an extension name change from mpos / spos to mposXML / sposXML was committed in there. After I saw this, I also tried using the same command line as above, but using these other extensions, but that also did not work for me.:
```
"C:\Program Files\MuseScore 4\bin\MuseScore4.exe" -o La_Llorona.mposXML La_Llorona.mscz
"C:\Program Files\MuseScore 4\bin\MuseScore4.exe" -o La_Llorona.sposXML La_Llorona.mscz
```
So, that also seems to indicate that it's more than just the extension strings that changed.

**Use case**

See also the original report by user "Thibaultj" mentioned above.

In our case, we use these .mpos and .spos files in combination with the exported .png and .wav files to enable automatic score following on MuseScore files from live music performances in our desktop application "MusicFollower" (see SampleSumo website if you want more info, or ask me).

Not having the mappings of event IDs to locations on the images (which is what these files represent) basically makes it impossible to do our score following any longer (at least not with MuseScore 4), so for us this is a "breaking" change.

**Request for fix / alternative?** 

Can this useful functionality be restored please?
Or was this functionality perhaps transformed into something else (with the big changes for version 4), and can it be obtained in another way now from the command line?
I assume this kind of mapping data is also used for the "cursor" animations of the web site playback graphics?

Eism:
Resolves: #16710

### DIFF
--- a/src/converter/internal/compat/backendapi.cpp
+++ b/src/converter/internal/compat/backendapi.cpp
@@ -51,8 +51,10 @@ using namespace mu::io;
 
 static const std::string PNG_WRITER_NAME = "png";
 static const std::string SVG_WRITER_NAME = "svg";
-static const std::string SEGMENTS_POSITIONS_WRITER_NAME = "sposXML";
-static const std::string MEASURES_POSITIONS_WRITER_NAME = "mposXML";
+static const std::string SEGMENTS_POSITIONS_WRITER_NAME = "spos";
+static const std::string SEGMENTS_POSITIONS_TAG_NAME = "sposXML";
+static const std::string MEASURES_POSITIONS_WRITER_NAME = "mpos";
+static const std::string MEASURES_POSITIONS_TAG_NAME = "mposXML";
 static const std::string PDF_WRITER_NAME = "pdf";
 static const std::string MIDI_WRITER_NAME = "midi";
 static const std::string MUSICXML_WRITER_NAME = "mxl";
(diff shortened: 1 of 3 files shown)


igorkorsukov (approved):

Jojo-Schmitz:
See also my #13153

Eism:
@Jojo-Schmitz 
Oh, sorry. I forgot about your PR.

For mpos and spos, my solution is better, since it does not require changes to the backend.
As for midi - no problem. { "mid", "midi", "kar" } registered for the writer

DmitryArefiev:
Tested #16710 on Win10 - **FIXED**

PR #17466, opened 2023-05-04
Overlaps calculation fix for harm marks and vibrato segs

mikekirin:


src/engraving/layout/layoutsystem.cpp
```diff
@@ -1366,7 +1366,9 @@ void LayoutSystem::processLines(System* system, std::vector<Spanner*> lines, boo
                 && ss->visible()
                 && prevSegment->isHarmonicMarkSegment()
                 && ss->isVibratoSegment()
-                && prevSegment->x() == ss->x()) {
+                // for some zoom sizes when the line crosses the bar line x not always the same
+                // but the difference is around 0.00000... so we have to round the value
+                && (int)prevSegment->x() == (int)ss->x()) {
```
cbjeukendrup, line 1371: Usually, we use `RealIsEqual` from `realfn.h` to handle rounding errors.

Eism (approved):

Issue #16957, opened 2023-03-21
[MU4 Issue] "Update existing score" option missing from "Publish to musescore.com"


murimk:
### Issue type

Cloud saving/loading issue

### Bug description

In M3, when uploading to musescore.com, you have the option of selecting whether to update existing score or not. Often I choose not to, as I work from one document to make many different files. This is very useful for my purposes, making lots of one or two bar snippets (examples) that I later embed on a website. There is no need to keep files like these around, so I save them to the cloud directly. There's also custom page settings for different files, so setting up new files or making templates is not a good workaround.

In M4, there's no such option, and any file generated from the same "mother" file is automatically updated when uploaded. This completely break my workflow, and I'd love to see the option of updating existing score to be reintroduced.

### Steps to reproduce

1. ;Make a short score
2. Upload to musescore.com
3. Make change to score
4. Upload to musescore.com
5. See that there's no option to select whether you want to update your file or not
6. See on musescore.com that your file has been updated, and possibly renamed (if you chose to), but in any case your original file is gone.

### MuseScore Version

MuseScore version (64-bit): 4.0.2-230651553, revision: github-musescore-musescore-dbe7c6d

### Regression

Yes, this used to work in Musescore 3.x and now is broken

### Operating system

OS: Windows 10 Version 2009, Arch.: x86_64,

XiaoMigros:
Additionally, when updating an existing score, the title to be uploaded is always sourced from the file name rather than the title under which the score is already published.

shoogle:
Here's an updated design for the _Publish to MuseScore.com_ dialog:

![image](https://user-images.githubusercontent.com/11011881/235878650-8cefd2e2-5977-4417-9b10-587eedec3b69.png)

This adds radios buttons with options to:

- Replace the existing [online score](https://musescore.com/user/37221589/scores/8455808)
- Publish as new online score

In the "replace" option, the words "online score" will be a link that points to the score page on MuseScore.com.

The options will only displayed when there is an existing score that could be replaced. If there is no online score, or it is on a different user's account, then the radio buttons are not displayed and the default action is to publish as a new online score. This keeps the dialog simple for first time users.

When there is an existing online score that could be replaced, the title and visibility displayed in this dialog will be taken from the online score. MuseScore must query the website to get the latest title and visibility every time the dialog is displayed.

XiaoMigros:
After using the 'Publish' feature some more I've encountered another issue:

[...]

PR #17471, opened 2023-05-04
Publish: add option to choose whether to replace existing score or upload as new score

cbjeukendrup:
Resolves: #16957

### DIFF
--- a/src/project/internal/isaveprojectscenario.h
+++ b/src/project/internal/isaveprojectscenario.h
@@ -42,6 +42,8 @@ public:
     virtual RetVal<CloudProjectInfo> askPublishLocation(INotationProjectPtr project) const = 0;
 
     virtual bool warnBeforeSavingToExistingPubliclyVisibleCloudProject() const = 0;
+
+    virtual void showCloudSaveError(const Ret& ret, bool publishMode, bool alreadyAttempted) const = 0;
 };
 }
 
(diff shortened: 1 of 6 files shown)


shoogle (changes requested):
The dialog looks good, but before showing it you need to query the website to get the latest title and visibility of the online score, then display those values in the dialog.

<details><summary><b>Why?</b></summary>

1. The user might have changed the title or visibility online since the dialog was last opened.
    - By default, we want to preserve the latest online values.
2. The user might be about to replace the wrong online score by mistake.
    - For example, let's say the user copies their local Movement 1 file to use as a template for Movement 2. Unless the user remembers to delete the source URL in _File > Project Properties_, both movements will be linked to the same online score. When the user opens the _Publish to MuseScore.com_ dialog in Movement 2, we want to show the current title of the online score (i.e. "Movement 1") so that the user can see that they are about to replace the wrong online score, and they will know to click "Publish as new" instead.
 
</details>

If something prevents you from fetching the title and visibility then you'll need to show an appropriate message.
Logged out | No internet | Other error
:---:|:---:|:---:
image | image | image

It's possible the source URL is invalid or points to a score on a different user's account. In this situation we shouldn't see an error. Instead, the _Publish to MuseScore.com_ dialog should appear without the option to replace the existing score.

RomanPudashkin (approved):

RomanPudashkin (approved):

Issue #17454, opened 2023-05-03
MuseScore 4.0.2 won't start 


SteppenHerring:
I realise that there have been a bunch of these. So here's one more. 

Setup: Win 10 build 19044.2846. Intel 6800K. 64GB RAM. Lots of drives (NVME, SSD and HDD). Audio: Focusrite Scarlett 18i20 (2nd gen, USB). MuseScore splash screen then the process dies. Killing any process starting with "Muse" does not help.

If I change to a different audio output, MS4 starts and runs fine. I can't listen to the output though.

If I start MS4 and then change to the Focusrite, it hangs for a bit, then dies. The Focusrite is my only audio option that is connected to anything I can listen to.

Things tried:

- Rebooting
- Uninstall and reinstall
- Reinstall standalone MuseScore 4
- Uninstall all sounds
- Drinking
- Swearing

oktophonie:
@cbjeukendrup does this sound the same as https://github.com/musescore/MuseScore/issues/15235 to you?

cbjeukendrup:
Yes, most likely it is the same. So let's close this as a duplicate for now, but let's check it again when we find a solution for #15235.

Issue #17273, opened 2023-04-18
The program crashes every time the monitor sleeps


Auase:
### Issue type

Crash or freeze

### Steps to reproduce

The program crashes every time the monitor sleeps.

### Screenshots/Screen recordings

![Uploading 20230419004743.png…]()

### MuseScore Version

4.0.2

### Regression

I don't know

### Operating system

win11

DmitryArefiev:
**Can't reproduce** it on my side (Windows11)

@Auase Can you provide more info please? (Environment, monitor or computer sleep, does it occur with MS3 and other apps)

Auase:
@DmitryArefiev 
[MuseScore_230505_171455.log](https://github.com/musescore/MuseScore/files/11404867/MuseScore_230505_171455.log)
[MuseScore_230505_171820.log](https://github.com/musescore/MuseScore/files/11404868/MuseScore_230505_171820.log)
------------------------------------------
The following is a copy of the system error report file：
（Due to format limitations in uploading files, i have to compress them）
[WER2F1E.tmp.zip](https://github.com/musescore/MuseScore/files/11404900/WER2F1E.tmp.zip)

Auase:
![bug](https://user-images.githubusercontent.com/56365735/236423728-eff4a1cc-901f-4e0d-a755-7e3ef29e1fd3.png)

DmitryArefiev:
Thanks! 

Can you attach MS diagnostic files please (Menu bar>Diagnostic>Save diagnostic files)?

Also, can you answer some questions:
What is your environment (Laptop or Workstation, external monitor, external sound card, CPU, memory amount)?
Do you have the same crash with another apps?
MS crashes only when monitor sleeps or computer sleeps?
Does it crash even after several minutes of sleep?

Auase:
@DmitryArefiev 
Hello, only MuseScore crashes during screen sleep. Here is the relevant information:
[Diagnostic files.zip](https://github.com/musescore/MuseScore/files/11406326/Diagnostic.files.zip)
![20230505201912](https://user-images.githubusercontent.com/56365735/236460629-15b9cb8e-b575-433f-bf18-c4dc87064ea4.png)
![20230505202006](https://user-images.githubusercontent.com/56365735/236460648-068ec69c-23f5-462d-a217-76892a179438.png)
![20230505202108](https://user-images.githubusercontent.com/56365735/236460676-6ef3c62c-3530-4c38-84ab-bf8e888e5b6b.png)
[...]

Issue #17089, opened 2023-03-31
Musescore 4 not start (Sorry by my English), I´ve been written text in Spanish because I think is better for your own translation


Alberthoven2:
### Issue type

Crash or freeze

### Bug description

Hola.

Os escribo porque el musescore 4 me está dando miles de problemas. No consigo que se ejecute con normalidad. Desde el primer día que lo instalé solo he conseguido abrirlo 2 veces. El resto es imposible. Se me abre el logo del programa, se cierra y ya está. No aparece nada nunca. He probado a desinstalar el antivirus, desinstalar e istalar de nuevo musescore, revertir el sistema a un estado anterior, etc... y no hay manera de hacerlo funcionar.

También tengo preoblemas con el correo electrónico de confirmación de cuenta. Es este correo, pero por alguna razón no llega nunca, ya lo he pedido 4 o 5 veces y siempre es lo mismo.

Necesito vuestra ayuda porque no entiendo que puede estar pasando y esto ya me empieza a cansar mucho.
Por favor. Ayudadme a solucionar esto lo antes posible. Muchísimas gracias.

### Steps to reproduce

Try to open the program

### MuseScore Version

4.02

### Regression

Yes, this used to work in a previous version of Musescore 4.x

### Operating system

Windows 10 Pro

### Additional context

[DxDiag.txt](https://github.com/musescore/MuseScore/files/11123206/DxDiag.txt)
[Report.txt](https://github.com/musescore/MuseScore/files/11123230/Report.txt)

GabeS573:
Translation: (google translate + editing)

> Hello.
> 
> I am writing to you because MuseScore 4 is giving me thousands of problems. I can't get it to run normally. Since the first day I installed it I have only managed to open it 2 times. The rest is impossible. The logo of the program opens, closes and that's it. Nothing ever appears. I have tried uninstalling the antivirus, uninstalling and installing MuseScore again, reverting the system to a previous state, etc... and there is no way to make it work.
> 
> I am also having problems with the account confirmation email. It's this email, but for some reason it never arrives. I've already asked for it 4 or 5 times and it's always the same.
> 
> I need your help because I don't understand what could be happening and this is starting to tire me a lot.
> Please. Please help me fix this as soon as possible. Thanks a lot.

@Alberthoven2 

I'm not sure what the account confirmation email is referencing, but if it's related to the musescore.com website, you need to email their support team separately. This repository is for reporting issues with the desktop software specifically.

[...]

Issue #16950, opened 2023-03-20
Musescore 4 will not open unless my sound device is NOT Astro A50 Game/Voice (the two audio outputs from my headphones).


Oscartwest:
### Issue type

Crash or freeze

### Bug description

I did some research and looked at some form posts when I found this one.
https://musescore.org/en/node/338271#comment-1162366
This comment provides two builds which both solve my issue. The comment thread says the fix was implemented, so I am confused as to why this is happening.

### Steps to reproduce

1. Have one of the audio devices from my headphones selected.
2. Open Musescore...
3. Crash!

### MuseScore Version

4.0.2

### Regression

Yes, this used to work in Musescore 3.x and now is broken

### Operating system

Windows 11 Pro

### Additional context

OS: Windows 11 Pro
Headphones: Astro A50 Gen 3 (The Astro A50s have two different audio outputs)

2023-03-20T16:32:11.159 | INFO  | main_thread | GlobalModule | onPreInit: log path: C:/Users/tyler/AppData/Local/MuseScore/MuseScore4/logs/MuseScore_230320_163211.log
2023-03-20T16:32:11.159 | INFO  | main_thread | GlobalModule | onPreInit: === Started MuseScore 4.0.2, build number 230651553 ===
2023-03-20T16:32:11.332 | INFO  | main_thread | DiagnosticsModule | onInit: success start crash handler
2023-03-20T16:32:11.333 | WARN  | main_thread | Qt         | QIODevice::read (QFile, "C:\Users\tyler\AppData\Local\MuseScore\MuseScore4\shortcuts.xml"): device not open
2023-03-20T16:32:11.334 | WARN  | main_thread | Qt         | QIODevice::read (QFile, "C:\Users\tyler\AppData\Local\MuseScore\MuseScore4\midi_mappings.xml"): device not open
2023-03-20T16:32:11.371 | INFO  | main_thread | WasapiAudioDriver | open: WASAPI: trying to open the audio end-point with the following sample rate - 44100
2023-03-20T16:32:11.371 | INFO  | main_thread | WasapiAudioDriver | open: WASAPI: trying to open the audio end-point with the following samples per channel number - 1024
2023-03-20T16:32:11.377 | INFO  | 8036 | WasapiAudioClient | configureDeviceInternal: WASAPI: Settings device client properties
2023-03-20T16:32:11.377 | INFO  | 8036 | WasapiAudioClient | configureDeviceInternal: WASAPI: Getting device mix format
[...]

PR #17485, opened 2023-05-07
Fix the crash on launch on Windows

cbjeukendrup:
Resolves: #15235
Resolves: https://github.com/musescore/MuseScore/issues/17273
Resolves: https://github.com/musescore/MuseScore/issues/17454
Resolves: https://github.com/musescore/MuseScore/issues/17089
Resolves: https://github.com/musescore/MuseScore/issues/16947
Resolves: https://github.com/musescore/MuseScore/issues/16309
Resolves: #16950

Since the crash only occurred in builds with an installer, here is a build with an installer from this branch: https://github.com/musescore/MuseScore/actions/runs/4903046436 (not the portable but the other one)


cbjeukendrup:
In the wasapi driver code, we work a lot with exceptions that are thrown and caught. But in builds with an installer, exceptions were never caught, even when they are properly wrapped in a try/catch block. This was not limited to exceptions from the wasapi driver, but experiments showed that it also happened with for example a std::runtime_exception. So somehow exception handling was broken. 

Since it seemed so strange that this was caused by the packaging, we (=me and two very helpful community members) looked at the build process, but really everything was identical between builds with and without installers. 

I also tried first packaging as a MSI installer and then packaging the same build as a zip file too, both in the same GitHub Actions run. The version from the zip file worked, so it was really not that the build itself was wrong. So the problem had to be in the packaging. I feared that WIX somehow broke the build when packing it into the MSI, so I even started looking at the source code of WIX, but decided to spend not too much time on that because I wasn't even really sure that it was the culprit. 

Then I tried to copy more things into the artifact; not only the WIX logs, but also other intermediate files. I found that in these intermediate files, it was broken, so the problem was _not_ the packaging into the MSI. 

So I scrutinised the packaging scripts once more, and then my attention was caught by that line about stripping. I removed it, and that did the job. 

Those NSIS files that I removed are unrelated, but they are unused so I cleaned them up.

igorkorsukov (approved):

cbjeukendrup:
@DmitryArefiev I'm not sure how much you can test about this PR, since IIRC you could not reproduce the crash initially. But I still wanted to inform you :)
In #15235 and #17273, some users who did experience the crash before, have reported that it seems to work well for them.

DmitryArefiev:
@cbjeukendrup It's great you found a solution! I tried an installer from https://github.com/musescore/MuseScore/actions/runs/4903046436 - all good on my side. But it's a TEST version. Let's merge it to master, then create a release version of installer (stable build mode) and ask reporters of related bugs to check it again.

DmitryArefiev:
@cbjeukendrup Oh I see, that makes sense. But on the other hand they can have a build without crash in a regular location :) 
OK let's leave as it is now and create a stable build if users with crash will ask it.

Issue #15332, opened 2022-12-18
[FEATURE] Trills, mordents & Turns should have additional settings in the properties panel to allow different intervals


oMrSmith:
- There could be a palette to pick a desired accidental,
which will then automatically be placed on top of the trill
- and a switch to toggle between diatonic playback and "custom interval".

Tantacrul:
Design for this is ready **[here:](https://www.figma.com/file/7Hr6wLYlNxOs05vQQ69AeA/Ornament-Playback-With-Accidentals?node-id=1%3A2&t=J3AWoDarc2MRDx48-1)**

It has a few UI / engraving and playback tasks associated with it.

oMrSmith:
I just took a first glance and say: Wow, that looks like a very good and sophisticated solution.

hstanekovic:
The option name "Automatic (diatonic)" sounds misleading to me. The document says that this option takes into account the key as well as any prior accidentals within the same bar. 
"Diatonic" suggests to me that only the key will be taken into account. Why? Because accidentals in bars are normally used to produce altered (chromatic) tones from diatonic tones (which are determined by the key). Thus I think the better name for the option would be simply "Automatic".

PR #17515, opened 2023-05-09
Ornaments

mike-spa:
Resolves: #15332

A complete rework of ornaments and their options. Now ornaments can have customizable above/below intervals as appropriate, and can display accidentals when necessary, as well as displaying a cue note for non-standard trills.

Codewise, ornaments currently don't have any dedicated class, they are just treated as articulation. This PR introduces the Ornament class as an extension of the Articulation class. Trill-lines (which are internally represented by the Trill class) need to remain a separate type, but they now also wrap an Ornament, which makes them able to access all the same options.

This PR doesn't deal with playback, but it computes the implicit notes above/below as appropriate, which are the only thing needed for correct playback of the ornament and are accessible as `Ornament::noteAbove()` and `Ornament::noteBelow()`.

Ready for @Tantacrul and @oktophonie to test the design and engraving sides.


Tantacrul:
I have a few thoughts. I'll post each as an individual comment, because I'll likely find more over time. I'll also need to think about the UX here because for sure the existing UX doesn't feel right. For now, I'll stick to the obvious issues:

Here, I have selected the note in parenthesis generated next to the trill. In the Properties panel, I now have a lot of options that don't do anything. I would suggest hiding (or failing that, disabling) anything that won't work.  So, the **Stem** and **Beam** tabs in their entirety - and everything in the **Head** tab, apart from the Notehead type, Hide Notehead checkbox and Notehead offset setting (in 'more'). We could also simply remove all these options completely. I think that would be OK too in this instance.

image

Tantacrul:
This next problem may require handing over to another developer once the engraving part is right.

In essence: none of the choices made in the Properties panel affect playback, like they should. I'm guessing you are going to leave playback stuff for now @mike-spa ?

If so, I can move on to other things. Just to make the point that the feature won't be complete until the playback is working perfectly.

Tantacrul:
I can't get any visual results when I make changes to trill lines. Is this still in development?

image

Tantacrul:
Last one for tonight. (Although more are coming)

[Here's the file ](https://www.dropbox.com/s/8zswigyvnbrjj8u/ORNAMENT_TEST.mscz?dl=0)I'm using.

Here's a video of the strangeness, which I think might be related to dotted notes. Not 100% sure!

https://github.com/musescore/MuseScore/assets/21022311/eee2907d-a206-4887-bfbd-0d7e044a3cd7

Tantacrul:
Found a consistent crash. [Link to file.](https://www.dropbox.com/s/tyb744jbxjtnbsp/02_Crash_Turn.mscz?dl=0)

https://github.com/musescore/MuseScore/assets/21022311/66025532-3af4-4932-a790-dd3ac1afe4f7

oktophonie (approved):

src/inspector/models/notation/ornaments/ornamentsettingsmodel.h
```diff
@@ -40,12 +53,57 @@ class OrnamentSettingsModel : public AbstractInspectorModel
...
+
+signals:
+    void isIntervalAboveAvailableChanged(bool available);
+    void isIntervalBelowAvailableChanged(bool available);
+    void isFullIntervalChoiceAvailableChanged(bool available);
+    void isPerfectStepChanged(bool perfect);
 
 private:
-    PropertyItem* m_performanceType = nullptr;
     PropertyItem* m_placement = nullptr;
```
RomanPudashkin, line 82: All private members should be declared after private methods

src/inspector/models/notation/ornaments/ornamentsettingsmodel.h
```diff
@@ -40,12 +53,57 @@ class OrnamentSettingsModel : public AbstractInspectorModel
...
+    PropertyItem* m_showAccidental = nullptr;
+    PropertyItem* m_startOnUpperNote = nullptr;
+
+private:
+    void updateIsIntervalAboveAvailable();
+    void updateIsIntervalBelowAvailable();
+    void updateIsFullIntervalChoiceAvailable();
+    void updateIsPerfectStep();
+
+private slots:
```
RomanPudashkin, line 104: As far I can tell, they are just private methods, not private slots

src/engraving/types/types.h
```diff
@@ -609,6 +610,53 @@ enum class ArpeggioType : unsigned char {
...
+    AUGMENTED,
+    MAJOR,
+    PERFECT,
+    MINOR,
+    DIMINISHED
+};
+
+struct OrnamentInterval
+{
+    IntervalStep m_step = IntervalStep::SECOND;
```
RomanPudashkin, line 635: Please rename them to `step` and `type`

src/inspector/models/notation/ornaments/ornamentsettingsmodel.cpp
```diff
@@ -39,46 +48,361 @@ OrnamentSettingsModel::OrnamentSettingsModel(QObject* parent, IElementRepository
...
+        }
+    }
+
+    setIsFullIntervalChoiceAvailable(available);
+}
+
+void OrnamentSettingsModel::updateIsPerfectStep()
+{
+    int intervalStepValue = !m_intervalStep->isUndefined() ? m_intervalStep->value().toInt() : -1;
+    bool perfect = intervalStepValue == static_cast<int>(IntervalStep::UNISON)
```
RomanPudashkin, line 257: It is better to use sets when you need to compare a lot of types. In this case, the code will look like this:

```
std::unordered_set<IntervalStep> perfectStepTypes {
    IntervalStep::UNISON,
    IntervalStep::FOURTH,
    ...
};

setIsPerfectStep(mu::contains(perfectStepTypes, intervalStepValue));
```

src/inspector/models/notation/notes/noteheads/noteheadsettingsmodel.cpp
```diff
@@ -207,3 +213,24 @@ QVariantList NoteheadSettingsModel::possibleHeadSystemTypes() const
...
+        }
+    }
+    setIsTrillCueNote(isTrillCueNote);
+}
+
+void NoteheadSettingsModel::setIsTrillCueNote(bool v)
+{
+    if (v == m_isTrillCueNote) {
+        return;
+    }
```
RomanPudashkin, line 233: Please add an empty line after }

src/engraving/types/types.h
```diff
@@ -609,6 +610,53 @@ enum class ArpeggioType : unsigned char {
...
+    IntervalType m_type = IntervalType::AUTO;
+
+    OrnamentInterval() = default;
+    OrnamentInterval(IntervalStep s, IntervalType t)
+        : m_step(s), m_type(t) {}
+
+    inline bool operator ==(const OrnamentInterval& interval) const { return m_step == interval.m_step && m_type == interval.m_type; }
+    inline bool operator !=(const OrnamentInterval& interval) const { return !operator ==(interval); }
+
+    bool isPerfect() const
```
RomanPudashkin, line 645: Perhaps you should add a static function and reuse it here:
https://github.com/musescore/MuseScore/pull/17515/files#r1196125739

src/engraving/libmscore/trill.h
```diff
@@ -84,8 +81,10 @@ class Trill final : public SLine
 private:
     TrillType _trillType;
     Accidental* _accidental;
+    Chord* _cueNoteChord;
```
RomanPudashkin, line 84: = nullptr
 
 You should assign a default value for all the private members

src/engraving/libmscore/ornament.h
```diff
@@ -0,0 +1,74 @@
...
+    OBJECT_ALLOCATOR(engraving, Ornament)
+    DECLARE_CLASSOF(ElementType::ORNAMENT)
+
+private:
+    M_PROPERTY(OrnamentInterval, intervalAbove, setIntervalAbove)
+    M_PROPERTY(OrnamentInterval, intervalBelow, setIntervalBelow)
+    M_PROPERTY(OrnamentShowAccidental, showAccidental, setShowAccidental)
+    M_PROPERTY(bool, startOnUpperNote, setStartOnUpperNote)
+
+    std::array<Note*, 2> _notesAboveAndBelow { nullptr, nullptr }; // [0] = above, [1] = below
```
RomanPudashkin, line 20: See https://github.com/musescore/MuseScore/pull/17515#discussion_r1196106855

src/engraving/libmscore/ornament.h
```diff
@@ -0,0 +1,74 @@
+#ifndef ORNAMENT_H
```
RomanPudashkin, line 1: You missed the license for both .h/.cpp files

Also, this define should be MU_ENGRAVING_ORNAMENT_H

src/engraving/libmscore/engravingobject.h
```diff
@@ -504,6 +505,11 @@ class EngravingObject
     {
         return isStaffText() || isSystemText() || isTripletFeel() || isPlayTechAnnotation();
     }
+
+    bool isArticulation() const
```
RomanPudashkin, line 509: It is better not to add this function. Use this comparison directly in SymbolsMetaParser

src/engraving/libmscore/edit.cpp
```diff
@@ -5496,12 +5504,35 @@ static void undoChangeRestVisibility(Rest* rest, bool visible)
...
+    }
+}
+
 void Score::undoChangeVisible(EngravingItem* item, bool visible)
 {
     if (item->isNote()) {
         undoChangeNoteVisibility(toNote(item), visible);
     } else if (item->isRest()) {
         undoChangeRestVisibility(toRest(item), visible);
+    } else if (item->isOrnament()) {
```
RomanPudashkin, line 5528: I expect to see new unit-tests for all these new cases

See changevisibility_tests.cpp

src/engraving/libmscore/articulation.cpp
```diff
@@ -685,34 +645,38 @@ void Articulation::doAutoplace()
...
-        } else {
-            sk.add(r.x(), r.top(), r.width());
-            d = ss->skyline().south().minDistance(sk);
-        }
 
-        if (d > -md) {
-            double yd = d + md;
+        Shape thisShape = shape();
+        for (ShapeElement& shapeEl : thisShape) {
+            RectF r = shapeEl.translated(chordRest()->pos() + m->pos() + s->pos() + pos());
```
RomanPudashkin, line 651: This should be called outside of the for loop

`chordRest()->pos() + m->pos() + s->pos() + pos()`

src/engraving/libmscore/articulation.cpp
```diff
@@ -685,34 +645,38 @@ void Articulation::doAutoplace()
...
+
+            double d;
```
RomanPudashkin, line 653: double  d = 0.0;

src/engraving/libmscore/accidental.cpp
```diff
@@ -327,6 +332,19 @@ void Accidental::setSubtype(const AsciiStringView& tag)
     setAccidentalType(name2subtype(tag));
 }
 
+//---------------------------------------------------------
+//   layout
```
RomanPudashkin, line 336: Please remove this comment

DmitryArefiev:
Tested on **Win10** how big scores with ornaments (from master) open in PR's build - **didn't find a difference in time**

src/inspector/models/general/generalsettingsmodel.cpp
```diff
@@ -58,6 +58,9 @@ void GeneralSettingsModel::requestElements()
     QSet<EngravingItem*> elementsForIsSmallProperty;
 
     for (EngravingItem* element : m_elementList) {
+        if (element->isNote() && toNote(element)->isTrillCueNote()) {
```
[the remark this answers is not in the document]
mike-spa: That's because the trill cue note is small by definition, so the isSmall property can't be edited.

cueSizeExplanation

I've added a comment.

src/engraving/libmscore/engravingobject.h
mike-spa: I'm not sure I understand this. This function is the standard type-id check that all `EngravingObject`s have. The only difference is that now `Articulation` also has the sub-class `Ornament`, so this check should return true also for Ornament. Anyway, to avoid confusion, I have now returned `isArticulation()` to the way it was before (implemented via template [here](https://github.com/musescore/MuseScore/blob/master/src/engraving/libmscore/engravingobject.h#L316)) and renamed this function to `isArticulationFamily()`. This is now identical to the other cases of types/subtypes, for example `isRestFamily()` etc. Hope this is better

src/engraving/libmscore/edit.cpp
RomanPudashkin: @mike-spa please add the tests to your next PR. Thanks!

Issue #16874, opened 2023-03-17
Tuplet brtack


oktophonie:
### Issue type

Crash or freeze

### Bug description

When a tuplet is created and it contains only rests, 'Auto' bracket direction results in the bracket being drawn above the stave, which is fine when there is only a single voice.

It appears that a bracket for an empty tuplet is being drawn above the stave even when in a downstemmed voice (2 or 4), when it should instead be drawn below.

### Steps to reproduce

1. Enter note input mode
2. Switch to 2nd voice
3. Create a tuplet, e.g. with Ctrl+3

The tuplet is created fine, but the bracket is drawn _above_ the stave when it should be below:
![image](https://user-images.githubusercontent.com/20604032/225870989-e0e634bf-ed20-414e-bc54-4286a9f76f09.png)
As soon as you enter any notes it flips to the right side:
![image](https://user-images.githubusercontent.com/20604032/225871089-a827b72c-8d44-4d94-a563-1a0d805f4aae.png)

### MuseScore Version

OS: Artix Linux, Arch.: x86_64, MuseScore version (64-bit): 4.0.2-230651545, revision: github-musescore-musescore-dbe7c6d

### Regression

No.

### Operating system

Linux, Windows 11

Jojo-Schmitz:
What is a 'brtack'?

PR #17520, opened 2023-05-10
Enhance tuplet direction detection based on voices

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/16874

Empty tuplets would always default to 'up' as their auto direction, even when in voices 2 and 4. Now voices 2 and 4 empty tuplets face downwards.

I also noticed a TODO comment in there for nested tuplets, which I also replaced with some simple logic to make sure that the outer tuplet follows the direction of the inner ones (which are laid out first).


oktophonie (approved):

src/engraving/layout/tupletlayout.cpp
```diff
@@ -92,7 +92,7 @@ void TupletLayout::layout(Tuplet* item, LayoutContext&)
     // find out main direction
     //
     if (item->_direction == DirectionV::AUTO) {
-        int up = 1;
+        int up = 0;
```
RomanPudashkin, line 95: any new vtests?

RomanPudashkin (approved):

PR #17527, opened 2023-05-10
Fix pasting on Linux

Jojo-Schmitz:
See also https://doc.qt.io/qt-5/qclipboard.html#Mode-enum

Comments and esp. test reports welcome ;-)

### DIFF
--- a/src/notation/internal/notationinteraction.cpp
+++ b/src/notation/internal/notationinteraction.cpp
@@ -3032,12 +3032,7 @@ void NotationInteraction::changeTextCursorPosition(const PointF& newCursorPos)
 
     textEl->mousePress(m_editData);
     if (m_editData.buttons == mu::engraving::MiddleButton) {
-        #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
-        QClipboard::Mode mode = QClipboard::Clipboard;
-        #else
-        QClipboard::Mode mode = QClipboard::Selection;
-        #endif
-        QString txt = QGuiApplication::clipboard()->text(mode);
+        QString txt = QGuiApplication::clipboard()->text();
         textEl->paste(m_editData, txt);
     }
 
@@ -3552,12 +3547,7 @@ void NotationInteraction::copySelection()
         m_editData.element->editCopy(m_editData);
         mu::engraving::TextEditData* ted = static_cast<mu::engraving::TextEditData*>(m_editData.getData(m_editData.element).get());
         if (!ted->selectedText.isEmpty()) {
-            #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
-            QClipboard::Mode mode = QClipboard::Clipboard;
-            #else
-            QClipboard::Mode mode = QClipboard::Selection;
-            #endif
-            QGuiApplication::clipboard()->setText(ted->selectedText, mode);
+            QGuiApplication::clipboard()->setText(ted->selectedText);
         }
     } else {
         QMimeData* mimeData = selection()->mimeData();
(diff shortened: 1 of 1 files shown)


MarcSabatella:
This works for me, pasting now works with the system clipboard for me.  No ill effect I can see, but again, I have no idea why that ifdef was there originally.

RomanPudashkin:
> This works for me, pasting now works with the system clipboard for me. No ill effect I can see, but again, I have no idea why that ifdef was there originally.

It was added here: https://github.com/musescore/MuseScore/pull/10754

MarcSabatella:
Interesting, thanks for the link!  For me the bug this was supposed to fix does not exist - not in 4.0.2, not in master, and not with this PR.  Paste from within the score to within the score always works.  This PR however makes it so copy/paste with the system clipboard *also* works.

But, it looks like @DmitryArefiev was the one reporting the original problem in https://github.com/musescore/MuseScore/issues/10742 - does this PR break it again for him?  And if so, I wonder which experience is more typical for other Linux users?

Jojo-Schmitz:
That's just one of the three places

MarcSabatella:
Ooh, good catch!  So the problem, then, was that the copy and the paste were using different clipboards, which of course is a problem.  But then the question is, how did that come about?  It seems to have happened here: https://github.com/musescore/MuseScore/pull/8501.  I guess the fact that it was fdone in two place but not the third was just an overishgt, but it still doesn't answer the question of why it was done at all.

I guess the idea probably came from Lyrics::paste(), which was doing that already for some reason (?), and that then made it into other parts of the code.

The original ifdef in Lyrics::paste() has been there "forever", so hard to say why.

Jojo-Schmitz:
Ah yes, that got to be it, the code from lyrics paste got moves into 2 places, then the 3rd one got added later to match.
To me that means we can loose that platform dependency entirely, unless and until someone on Linus see it failing.

DmitryArefiev:
I re-tested #10742 on my LinuxUbuntu22.04.2 LTS - **all fine**. 

@MarcSabatella So if the current PR fixes your problem we can merge it after CR.

MarcSabatella:
This does indeed fix pasting on Linux - I can copy text from this window, for example, and paste into staff text or title in a score, which wasn't possible in 4.0.2 (but was in MU3).

So yes, this will definitely be most welcome for me.

src/notation/internal/notationinteraction.cpp
```diff
@@ -3025,11 +3025,7 @@ void NotationInteraction::changeTextCursorPosition(const PointF& newCursorPos)
 
     textEl->mousePress(m_editData);
     if (m_editData.buttons == mu::engraving::MiddleButton) {
-        #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
         QClipboard::Mode mode = QClipboard::Clipboard;
-        #else
-        QClipboard::Mode mode = QClipboard::Selection;
-        #endif
         QString txt = QGuiApplication::clipboard()->text(mode);
```
cbjeukendrup, line 3029: We don't need the `mode` variable anymore at all: `QClipboard::Clipboard` is already the default mode for `QClipboard::text()`. So we can simply call `QGuiApplication::clipboard()->text()`.

Jojo-Schmitz: Will so, tomorrow.

Jojo-Schmitz: I love fixing bug by removing code ;-)

Jojo-Schmitz:
> So if the current PR fixes your problem we can merge it after CR.

_after CR_ ?

RomanPudashkin (approved):

cbjeukendrup:
@Jojo-Schmitz CR = code review 🙂

Jojo-Schmitz:
Ah, I see. in my previous job it stood for Change Request

Issue #15028, opened 2022-12-07
[MU4 Issue] add interval commands (Alt+number) do not play sound


MarcSabatella:
**Describe the bug**
When using Alt+number commands to add a note a given interval above a given note, it does not play

**To Reproduce**
1. Add a note to an empty measure
2. Press Alt+3 to add another note a third above
3. Hear error: the note does not play

**Expected behavior**
Note should play

**Platform information**
 - OS: Linux, tested using RC1

**Additional context**
Similar bugs were fixed for other commands not too long ago, but I guess I missed this one.  See https://github.com/musescore/MuseScore/pull/14347 for the previous fix.  A similar approach would presumably work for the command to add intervals.  The processing of this command is a bit confusing to me, but I guess it should happen in addIntervalToSelectedNotes.  I'll let someone else worry about trying it out and submitting the PR though.

DmitryArefiev:
Thanks for logging!



https://user-images.githubusercontent.com/90187801/209936618-d902bd2c-235f-4f19-82e7-607e8dd21fe2.mp4

MarcSabatella:
This is a significant accessibility issue as blind users especially rely on audible feedback.  Since it seems I was pretty sure where the fix should go, I'll go ahead and make a PR soon.

DmitryArefiev:
Fixed in #17532 

After #17532 the behavior a little bit changed, see https://github.com/musescore/MuseScore/pull/17532#issue-1706353469

PR #17532, opened 2023-05-11
fix playback on add interval

MarcSabatella:
Resolves: #15028

Adds PlayMode::PlayChord to the handler for the add interval commands, so the full chord plays when adding a note via Alt+number.  Right now nothing is played, which is definitely a problem.  Note that MU3 only played the added note rather than the full chord, but Shift+letter plays the full chord.  I can't see a good reason for the difference so I'm setting PlayChord here.


cbjeukendrup (approved):

DmitryArefiev:
Tested #15028 on Win10, Mac13, LinuxUbuntu 22.04 - **FIXED** 

Thanks for fixing!

PR #17551, opened 2023-05-14
Fix #17542: Message user when attempting to delete initial key signature

### LINKED ISSUE #17542: Message user when attempting to delete initial key signature
### Your idea

Per https://github.com/musescore/MuseScore/pull/15503 it was resolved to disable deletion of the first key signature. This task ensures the user is appropriately messaged when attempting to do so to avoid likely confusion. 

### Problem to be solved

As above

### Prior art

key-sig

### Additional context

Marking as a priority to ensure it gets done for 4.1 – this really completes the feature implemented as part of #15503.

iwoithe:
Resolves: #17542


oktophonie:
This seems to trigger the popup once per stave - i.e. trying to delete the initial key signature in a wind quintet score, I had to click 'OK' five times.

iwoithe:
@oktophonie I have fixed the multiple staves issue (when you are able to test again)

oktophonie:
Thanks!
There seems to be another small window popping up before the dialog appears:

https://github.com/musescore/MuseScore/assets/20604032/75e64a6a-673e-4bed-ac62-d5a99b84d3f3

It's too fast for me to see what it says. What's the deal with that? Actually more obvious in Linux as, with my windows manager, the extraneous popup appears in a different place:

https://github.com/musescore/MuseScore/assets/20604032/a66e0f8c-9ece-4a54-90cd-12cf2796879b

Finally: it makes sense that the key signature become deselected after closing the dialog; but then the first rest after the key signature on the top stave becomes selected. Is that the desired behaviour? (I'm not saying it isn't - but I recall this question coming up before and I think there are some accessibility implications.)

oktophonie:
@iwoithe Any progress here? It would be excellent to have this finished and merged so that @sammik can leverage this also for his fix to https://github.com/musescore/MuseScore/issues/17626

iwoithe:
> There seems to be another small window popping up before the dialog appears: [...] It's too fast for me to see what it says. What's the deal with that? Actually more obvious in Linux as, with my window manager at least, the extraneous popup appears in a different place:

Not entirely sure what's going wrong here. Will look into it further. :confused:

> Finally: it makes sense that the key signature become deselected after closing the dialog; but then the first rest after the key signature on the top stave becomes selected. Is that the desired behaviour? (I'm not saying it isn't - but I recall this question coming up before and I think there are some accessibility implications.)

By default, deleting a range of objects defaults to selecting a single object. When only attempting to delete the initial key signature, it makes sense to select the initial key signature again. But what if a multi-selection is made, e.g. a key signature and a few notes? What should be selected after the notes are deleted and the popup is closed?

iwoithe:
@oktophonie I've updated the selection behaviour to the following:

https://github.com/musescore/MuseScore/assets/73422868/7418d5e3-2b79-492a-b100-7ece8639a071

src/engraving/libmscore/edit.cpp
```diff
@@ -3478,6 +3490,9 @@ void Score::cmdDeleteSelection()
             select(cr, SelectType::SINGLE);
         }
     }
+    if (keySig) {
+        select(keySig, SelectType::SINGLE);
+    }
```
iwoithe, line 3495: I must admit I'm not entirely sure if the key signature selection should override all other selections. :thinking:

I didn't find it a problem in the tests I did, but let me know if you'd like this fine-tuned further.

src/engraving/infrastructure/messagebox.cpp
```diff
@@ -25,7 +25,20 @@
 
 using namespace mu::engraving;
 
-MessageBox::Button MessageBox::warning(const std::string& title, const std::string& text, const std::set<Button>& buttons)
+MessageBox::Button MessageBox::warning(const std::string& title, const std::string& text, const std::set<Button>& buttons,
+                                       bool withIcon, bool withDontShowAgainCheckBox)
+{
+    return MessageBox::openDialog("WARNING", title, text, buttons, withIcon, withDontShowAgainCheckBox);
```
cbjeukendrup, line 31: It would be better to create a private enum in MessageBox and use the values from that enum instead of strings. That's more efficient and less error-prone.

cbjeukendrup:
> > @oktophonie After much toing and froing, I've updated the selection behaviour to the following.
> 
> I think, in score with multiple instruments, all key signatures in first measure should stay selected. Now only bottom signature is selected after dialog.
> 
> But better to wait for authority, about this design things.

I agree with this. See the advice I gave in https://github.com/musescore/MuseScore/pull/15363#pullrequestreview-1322471128 for a possible solution to a similar issue but for time signatures.

bkunda:
Just tested this on macOS. Generally the solution feels very nice (from a UX POV). I'm really glad we're implementing this.

Still not working:
- "Don't show again" checkbox (dialog continues to appear when attempting delete)
- Although there is no visible selection on the score, dialog can again be triggered by pressing `del` after closing the dialog:

https://github.com/musescore/MuseScore/assets/86290556/46f50dbb-4fc2-468f-a419-c2fe3db92bfc

bkunda (changes requested):
See comments: https://github.com/musescore/MuseScore/pull/17551#issuecomment-1594712700

PR #17568, opened 2023-05-15
Fix compiler warnings and non-unity build

cbjeukendrup:
When working outside the engraving module, it is convenient to have non-unity build working, to be able to use Qt Creator's code model (completion, navigation etc) (it doesn't work well with unity builds)


src/framework/vst/internal/vstpluginprovider.cpp
```diff
@@ -43,6 +43,10 @@ VstPluginProvider::VstPluginProvider(const PluginFactory& factory, const ClassIn
     m_impl = std::make_unique<Impl>(factory, info);
 }
 
+VstPluginProvider::~VstPluginProvider()
+{
```
cbjeukendrup, line 47: This is necessary because of the destruction of `std::unique_ptr<…> m_impl`, which requires the `Impl` to be a complete type, which is only the case in .cpp and not in .h.

RomanPudashkin (approved):

Issue #16918, opened 2023-03-19
Load thumbnails and metadata of recent scores asynchronously and only when necessary


cbjeukendrup:
Launching MuseScore, opening a score, and saving a score are very slow. And that's not because the operation itself is slow; it is because MuseScore has to reload the metadata and thumbnail (from your hard disk, or network drive, [or your non-responding OneDrive](https://github.com/musescore/MuseScore/issues/13161)) for every score in the "recent scores" list (which may contain an [infinite number of entries](https://github.com/musescore/MuseScore/issues/13799)), every single time when a score is added, moved or removed from the list. Even worse: it does this on the main thread, freezing the whole app for many seconds (in my case even 20 secs).

So I propose a few changes that will make a big difference:
- load the data in a background thread, so that the UI can move on. When the thumbnail for a score is not yet loaded but the home page is already displayed, just display a loading indicator in the spot of that thumbnail (or something else, or nothing).
- only load thumbnails for scores where needed. For example, the list may be very long, but until the user scrolls down, only the topmost few scores are visible, so we don't need to load a thumbnail yet for the others.
- don't reload _all_ thumbnails just when a score gets added/moved/removed, but cache them. (However, need to make sure that we do reload them when a score has been modified, otherwise an outdated thumbnail is displayed.)

DmitryArefiev:
Implemented in #17573

Issue #13161, opened 2022-09-04
[MU4 Issue] App fails to start up if recently used file can't be read


wizofaus:
**Describe the bug**
If a device or network issue means a recently used file can't be read from (in my example, OneDrive had gone into "paused" state and no files from my OneDrive folder could be read), MuseScore 4 will fail to start up at all, just showing an empty frame.

**To Reproduce**
1. Start MS, open/create a score and save to a location such as OneDrive where network reads can stall/timeout
2. Shutdown the app, change OneDrive or network settings such that the file just saved to can't be read from 
3. Restart MS
4. Application never gets past splash screen, or shows empty frame

**Expected behavior**
Attempts to read from recently used files on start up should not be blocking the UI.

**Screenshots**
![image](https://user-images.githubusercontent.com/5364225/188297458-e4e29c19-25dd-410b-9107-155854278d1b.png)

**Platform information**
 - Windows 11

**Additional context**

DmitryArefiev:
Should be fixed by #17573

PR #17573, opened 2023-05-15
Load score thumbnails asynchronously

cbjeukendrup:
Resolves: #16918 
Resolves: https://github.com/musescore/MuseScore/issues/13161


Jojo-Schmitz:
If you'd rebase it won't contain #17568 anymore ;-)

src/project/projecttypes.h
```diff
@@ -210,6 +210,14 @@ struct Template
 
 using Templates = QList<Template>;
 
+//struct RecentFile {
+//    io::path_t path;
+//    QString displayName;
+//};
+
+using RecentFile = io::path_t;
+using RecentFilesList = std::vector<RecentFile>;
```
cbjeukendrup, line 219: Spoiler alert: the plot unravels in (currently still only the comment in) https://github.com/musescore/MuseScore/pull/17764

src/project/internal/recentprojectscontroller.cpp
```diff
@@ -0,0 +1,214 @@
...
+        m_reloadingBlocked = false;
+    };
+
+    QJsonArray jsonArray;
+    for (const RecentFile& file : m_recentProjectsList) {
+        jsonArray << file.toQString();
+    }
+
+    QJsonDocument jsonDoc(jsonArray);
+    configuration()->setRawRecentProjectData(ByteArray::fromQByteArrayNoCopy(jsonDoc.toJson(QJsonDocument::Compact)));
```
RomanPudashkin, line 188: What do you think about saving this info to a separate json file instead of writing it to settings?

cbjeukendrup: That sounds like a good idea actually, I don't see why not. I'll try that out.

src/project/view/recentprojectsmodel.cpp
```diff
@@ -125,21 +123,21 @@ void RecentProjectsModel::updateRecentScores(const ProjectMetaList& recentProjec
...
+        obj[NAME_KEY] = io::filename(file, isSuffixInteresting).toQString();
+        obj[PATH_KEY] = file.toQString();
         obj[SUFFIX_KEY] = QString::fromStdString(suffix);
-        obj[IS_CLOUD_KEY] = configuration()->isCloudProject(meta.filePath);
+        obj[IS_CLOUD_KEY] = configuration()->isCloudProject(file);
 
-        if (!meta.thumbnail.isNull()) {
-            obj[THUMBNAIL_KEY] = !meta.thumbnail.isNull() ? meta.thumbnail : QVariant();
-        }
+//        if (!meta.thumbnail.isNull()) {
```
RomanPudashkin, line 136: This code should be fixed/removed

src/project/view/scorethumbnailloader.cpp
```diff
@@ -0,0 +1,72 @@
...
+        return;
+    }
+
+    m_scorePath = scorePath;
+    emit scorePathChanged();
+
+    loadThumbnail();
+}
+
+bool ScoreThumbnailLoader::isValid() const
```
RomanPudashkin, line 48: I would suggest renaming it to isThumbnailValid (because it is a thumbnail property, not a loader property)

src/project/internal/recentprojectscontroller.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+        async::Async::call(nullptr, [this, newList]() {
+            saveRecentProjectsList();
+        });
+    }
+}
+
+void RecentProjectsController::setRecentProjectsList(const RecentFilesList& list, bool save)
+{
+    m_recentProjectsList = list;
```
RomanPudashkin, line 160: It is better to check if this list has actually changed

src/project/internal/recentprojectscontroller.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+    {
+        std::lock_guard lock(m_thumbnailCacheMutex);
+        m_thumbnailCache.clear();
```
RomanPudashkin, line 164: Could you please explain why we need this? So the cache is cleared every time when the user opens a score. Is my assumption correct?

cbjeukendrup: This is indeed what would happen. This matches the old behaviour of regenerating all thumbnails every time when something changes in the list of recent files. This ensures that the user will see up-to-date thumbnails every time after saving a file in some instance. However, we could make this more sophisticated, for example by not invalidating the whole cache, but checking the "last modified" date of the file in question. I'll investigate.

src/project/view/scorethumbnailloader.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    return m_thumbnail;
+}
+
+void ScoreThumbnailLoader::loadThumbnail()
+{
+    recentProjectsController()->thumbnail(m_scorePath)
+    .onResolve(this, [this](const QPixmap& thumbnail) {
+        setThumbnail(thumbnail);
+    }).onReject(this, [this](int, const std::string&) {
```
RomanPudashkin, line 63: Please don't ignore the error

src/project/internal/recentfilescontroller.cpp
```diff
@@ -0,0 +1,263 @@
...
+            }
+        });
+
+        return Promise<QPixmap>::Result::unchecked();
+    }, Promise<QPixmap>::AsynchronyType::ProvidedByBody);
+}
+
+void RecentFilesController::cleanUpThumbnailCache(const RecentFilesList& files)
+{
+    QtConcurrent::run([this, files] {
```
RomanPudashkin, line 245: Why is QtConcurrent needed here?

cbjeukendrup: In order not to block the UI waiting for the mutex. If we would do that on the main thread, then we would first block the UI, waiting for all concurrent threads, until we can acquire the mutex lock, and then we can finally do the operation. 

Of course, when using QtConcurrent, the correct order might not be guaranteed (I can't find any promises about this in the documentation). But that does not matter much:
- if a thumbnail is inserted before cleaning, but it should have happened after, then it means that the thumbnail will have to be recomputed the next time it is requested. Not a big problem. In the majority of the cases, the cache does still do its job properly.
- if a thumbnail is inserted after cleaning, but it should have happened before, then the thumbnail will stay in the cache until the next cleanup. So one thumbnail remains in the cache a little longer. Not a big problem either. The only reason that we are doing these cleanups at all is to reduce useless memory usage, but one thumbnail more or less does not make a big difference.

DmitryArefiev:
Tested on Win10/Mac13

@cbjeukendrup I found some issues:

1. File > Open recent: score list **doesn't updated** after opening scores (**works in master**)



https://github.com/musescore/MuseScore/assets/90187801/2f2c96dc-d56a-465c-8e78-6349b1540858

2. File > Open recent > Clear recent files **doesn't clear score list right now** (**works in master**)


https://github.com/musescore/MuseScore/assets/90187801/74ef405a-e7ae-4ed4-837b-09aa34394987

3. Score list between master build and PR's build **loses sync between them** after changes (but works on first run of PR's build)


https://github.com/musescore/MuseScore/assets/90187801/4a2005a0-bd2d-4190-ad73-b235bf1c7e62

cbjeukendrup:
@DmitryArefiev Thanks! 1 and 2 should be better now. 3 is a necessary consequence of @RomanPudashkin's suggestion to store the recent files list in a different way from previously.

DmitryArefiev:
@cbjeukendrup Got it, thanks! Now all looks good.

Tested basic functionality with recent score feature on Win10, Mac13 - **PASS**

RomanPudashkin (approved):

PR #17575, opened 2023-05-16
Fix #17548 Allow melisma to be deleted on chordrest duration change

### LINKED ISSUE #17548: Adjust hyphens/extensions when changing note values
### Your idea

I noticed that _sometimes_ lyric syllable alignment changes when note values change. For example, if you have two quarter notes where the second is only a hyphen
image 
and you change those to a single half note, the syllable returns to center alignment. 
image
However, even if the notes that follow have new syllables, the hyphen remains. 

Ideally, Musescore should be able to recognize that there are no notes in between those explicit syllables and therefore the hyphen is erroneous and should be discarded. The same should apply to word extensions.

### Problem to be solved

This would enable the user to change note values and have the lyrics respond more intelligently.

### Prior art

_No response_

### Additional context

_No response_

asattely:
Resolves: https://github.com/musescore/MuseScore/issues/17548

When a chord's duration is changed, and that chord ends a melisma, the melisma was being lengthened unnecessarily. I've implemented a limit to the length of the melisma such that it will never be longer than the original.


oktophonie (approved):

src/engraving/libmscore/textedit.cpp
```diff
@@ -222,6 +222,26 @@ void TextBase::endEdit(EditData& ed)
     } else {
         triggerLayout();
     }
+    if (isLyrics()) {
```
RomanPudashkin, line 225: Why is this code here (in the base class) and not inside Lyrics?

asattely: The idea behind this section is that when a lyric is finished editing, it checks to see if there is a previous lyric with a line that extends through the current lyric's tick, and if so, resizes the previous lyric's line (or removes it) so that there is no overlap.

RomanPudashkin: > We could move this, but I'm not sure the best place to do it

This code looks like it should be inside the Lyrics class (since it's a lyrics-specific logic and TextBase is the base class for all text elements)

asattely: I still elected to call it from the base class, because the base class `endEdit` finishes the undo command, and it also has some early return logic that I felt would be a giant waste to copy and paste into the lyrics class. If this is still not good enough, I can move it into the lyrics class and replicate the early return checks from the base class.

Issue #16625, opened 2023-03-02
Frames: "Margin" should be renamed "Padding"


rgreen5:
### Your idea

In CSS (and WPs such as LibreOffice) "Margin" is used to describe the space _around_ an element and "Padding" that _within_ the element.

In MS frames "Left margin" and "Right margin" are currently the values of space _inside_ the frame and therefore should actually be named "Left Padding" and "Right Padding".

### Problem to be solved

It would make the MS terminology consistent with its usage elsewhere on the internet.

pratyaksh1610:
Can I work on this ?

cbjeukendrup:
Yes, feel free! 

(@bkunda could you please check if you agree with this idea? To me it would make sense, since "padding" seems to be used _everywhere_ to describe the space between a border and the content.)

pratyaksh1610:
I will start working on this once @bkunda approves this.

PR #16654, opened 2023-03-03
Reword "margin" to "padding" in frames

pratyaksh1610:
Resolves: #16625 

Reword margin to padding in frames.

- [x] I signed the [CLA](https://musescore.org/en/cla)
- [x] The title of the PR describes the problem it addresses
- [x] Each commit's message describes its purpose and effects, and references the issue it resolves
- [x] If changes are extensive, there is a sequence of easily reviewable commits
- [x] The code in the PR follows [the coding rules](https://github.com/musescore/MuseScore/wiki/CodeGuidelines)
- [x] There are no unnecessary changes
- [x] The code compiles and runs on my machine, preferably after each commit individually
- [x] I created a unit test or vtest to verify the changes I made (if applicable)

PR #17582, opened 2023-05-16
Fix GH#16625: Rename margin to padding

Jojo-Schmitz:
picking up where #16654 left

Resolves: #16625


bkunda:
@RomanPudashkin I've tested this in macOS.
🟢 FIXED

cbjeukendrup (commented):
See my reply at https://github.com/musescore/MuseScore/pull/16654#issuecomment-1552199374

Jojo-Schmitz:
If there are still other places, i didn't find them.
Will have another look, but after the weekend.

Jojo-Schmitz:
You mean those?

cbjeukendrup:
Yep, and thanks for spotting those in editstyle.ui too!

cbjeukendrup (approved):

cbjeukendrup:
@bkunda Now the similar properties for frames for Text objects have been renamed too: go to Properties > Text > More, select the rectangular frame option, and see that the "margin" property has been renamed to "padding". Do you want to re-test this?

Jojo-Schmitz:
As this PR changes strings, I'd rather get it merged soon...

bkunda:
> @bkunda Now the similar properties for frames for Text objects have been renamed too: go to Properties > Text > More, select the rectangular frame option, and see that the "margin" property has been renamed to "padding". Do you want to re-test this?

Tested in macOS. I'm happy with this change.
Thanks!

PR #17583, opened 2023-05-16
fixed #17565: Minor fixes for harp pedalling widget

### LINKED ISSUE #17565: Minor fixes for harp pedalling widget
### Issue type

UI bug

### Bug description

There are a few issues that should be addressed to polish-up the [recently implemented/merged harp pedal diagram feature](https://github.com/musescore/MuseScore/pull/12597):

#### Widget placement
The popup widget should always appear **above** the selected diagram (while also clearing the stave, as it currently does), _unless_ there is insufficient room at the top of the user's screen, in which case it should appear below the stave. 

In effect, this means the widget should be allowed to overlap the upper UI (note input toolbar, etc.), like so:

Screenshot 2023-05-15 at 2 48 31 pm

#### Key commands
These common key commands don't work on a selected harp pedal diagram:
- Delete (only works after scrolling the score to a different position)
- Copy/Paste (doesn't work at all)

https://github.com/musescore/MuseScore/assets/86290556/56201d79-dcf2-4812-a58b-a383001bae2d

In addition, the undo command reverts changes made to the diagram, but doesn't effect any visible change to the popup widget itself (the widget has to be closed and re-opened for the changes to appear).

https://github.com/musescore/MuseScore/assets/86290556/3514162d-bf81-409c-805d-834f2b54377c

_Expected behaviour_
It should be possible to **delete**, **copy/paste** harp pedal diagrams to different notes/rests.
It should also be possible to see the radio buttons within the UI of the widget itself revert to previous positions when the **undo** command is used 

 

### Steps to reproduce

See above videos

_No response_

### MuseScore Version

OS: macOS 13.1, Arch.: x86_64, MuseScore version (64-bit): 4.1.0-231350302, revision: github-musescore-musescore-494477a

### Regression

No.

### Operating system

macOS 13.1

### Additional context

@miiizen @Tantacrul FYI
Assigning P1 because this needs to be fixed in order to properly complete implementation of this feature

Eism:
Resolves: #17565

### DIFF
--- a/src/notation/view/abstractnotationpaintview.cpp
+++ b/src/notation/view/abstractnotationpaintview.cpp
@@ -81,6 +81,8 @@ AbstractNotationPaintView::~AbstractNotationPaintView()
         m_notation->accessibility()->setMapToScreenFunc(nullptr);
         m_notation->interaction()->setGetViewRectFunc(nullptr);
     }
+
+    m_previousSelectedElement = nullptr;
 }
 
 void AbstractNotationPaintView::load()
@@ -242,6 +244,14 @@ void AbstractNotationPaintView::onLoadNotation(INotationPtr)
 
     interaction->selectionChanged().onNotify(this, [this]() {
         redraw();
+
+        EngravingItem* selectedElement = notationInteraction()->selection()->element();
+        if (selectedElement != m_previousSelectedElement) {
+            hideElementPopup();
+            hideContextMenu();
+        }
+
+        m_previousSelectedElement = selectedElement;
     });
 
     interaction->showItemRequested().onReceive(this, [this](const INotationInteraction::ShowItemRequest& request) {
--- a/src/framework/ui/view/navigationsection.cpp
+++ b/src/framework/ui/view/navigationsection.cpp
@@ -73,20 +73,25 @@ bool NavigationSection::enabled() const
         return false;
     }
 
+    bool enbl = true;
+
     QWindow* sectionWindow = window();
     QWindow* topWindow = application()->focusWindow();
 
     if (sectionWindow && (topWindow != sectionWindow)) {
-        return type() == INavigationSection::Type::Exclusive;
+        enbl = type() == INavigationSection::Type::Exclusive;
     }
 
-    bool enbl = false;
-    for (INavigationPanel* p : m_panels) {
-        if (p->enabled()) {
-            enbl = true;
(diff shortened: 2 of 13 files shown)


src/notation/view/abstractnotationpaintview.cpp
```diff
@@ -81,6 +81,9 @@ AbstractNotationPaintView::~AbstractNotationPaintView()
         m_notation->accessibility()->setMapToScreenFunc(nullptr);
         m_notation->interaction()->setGetViewRectFunc(nullptr);
     }
+
+    m_previousSelectedElement = nullptr;
+    delete m_previousSelectedElement;
```
cbjeukendrup, line 86: This `delete` looks like it should not be there! (because 1. it deletes a nullptr and 2. we should not delete elements because that's handled by the Engraving module and the UndoStack)

DmitryArefiev:
Tested #17565 on Win10 - **seems to be fine**. @bkunda Can you try on Mac please? (if I missed something)

RomanPudashkin (approved):

DmitryArefiev:
@Eism I've tested #17565 more deeply on all platforms and found 2 bugs:

1. On **Mac**, **Linux** it allows less space above to show the diagram

See **Mac**

https://github.com/musescore/MuseScore/assets/90187801/231c71e7-c18e-4316-b90f-10c2164cdc26


and **Win**


https://github.com/musescore/MuseScore/assets/90187801/9d7937ae-a673-45df-b102-c5f311b10431

Eism:
@DmitryArefiev 
> On Mac, Linux it allows less space above to show the diagram

It's problematic to fix this in the current solution. We would spend more time than the benefits it would bring. But if you disagree I'll fix it

>Diagram popup doesn't close when moving MS window

Fixed

DmitryArefiev:
> It's problematic to fix this in the current solution. We would spend more time than the benefits it would bring. But if you disagree I'll fix it

Thanks! I'm fine with that.

FYI @bkunda See my https://github.com/musescore/MuseScore/pull/17583#issuecomment-1554338948

bkunda:
I too am fine with this. It's a lot better than where we started!
Thanks @Eism!

DmitryArefiev:
Re-tested #17565 + diagram popup issue on Win10, Mac13 - **FIXED**

